PR #54547, opened 2025-02-15
Make `rails new --quiet` fully quiet

### LINKED ISSUE #54546: `rails new --quiet` is not quiet
### Steps to reproduce

```
rails new myapp --quiet
```

### Expected behavior

There should be no output, as per `rails new --help`:

```
  -q, [--quiet], [--no-quiet], [--skip-quiet]        # Suppress status output
```

### Actual behavior

There is an output for some commands that run after the main app generation. By default these commands are `importmap:install` and `turbo:install stimulus:install`.

Full output:

```
       apply  /Users/jerome/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/importmap-rails-2.1.0/lib/install/install.rb
  Add Importmap include tags in application layout
      insert    app/views/layouts/application.html.erb
  Create application.js module as entrypoint
      create    app/javascript/application.js
  Use vendor/javascript for downloaded pins
      create    vendor/javascript
      create    vendor/javascript/.keep
  Configure importmap paths in config/importmap.rb
      create    config/importmap.rb
  Copying binstub
      create    bin/importmap
         run  bundle install --quiet
       apply  /Users/jerome/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/turbo-rails-2.0.11/lib/install/turbo_with_importmap.rb
  Import Turbo
      append    app/javascript/application.js
  Pin Turbo
      append    config/importmap.rb
         run  bundle install --quiet
       apply  /Users/jerome/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/stimulus-rails-1.3.4/lib/install/stimulus_with_importmap.rb
  Create controllers directory
      create    app/javascript/controllers
      create    app/javascript/controllers/index.js
      create    app/javascript/controllers/application.js
      create    app/javascript/controllers/hello_controller.js
  Import Stimulus controllers
      append    app/javascript/application.js
  Pin Stimulus
  Appending: pin "@hotwired/stimulus", to: "stimulus.min.js"
      append    config/importmap.rb
  Appending: pin "@hotwired/stimulus-loading", to: "stimulus-loading.js"
      append    config/importmap.rb
  Pin all controllers
  Appending: pin_all_from "app/javascript/controllers", under: "controllers"
      append    config/importmap.rb
         run  bundle install --quiet
      create  config/cache.yml
      create  db/cache_schema.rb
        gsub  config/environments/production.rb
      create  config/queue.yml
      create  config/recurring.yml
      create  db/queue_schema.rb
      create  bin/jobs
        gsub  config/environments/production.rb
      create  db/cable_schema.rb
       force  config/cable.yml
```

### System configuration

**Rails version**: 8.0.1

**Ruby version**: 3.4.1

jeromedalbert:
Fixes #54546.

### Motivation / Background

This Pull Request has been created because `rails new myapp --quiet` is not fully quiet when it calls Rails commands like `importmap:install` or `turbo:install stimulus:install`.

### Detail

This Pull Request overrides `rails_command` so any of the Rails commands executed in `Rails::Generators::AppBase` are quiet if the `--quiet` option is passed.

### Additional information

An alternative solution is to change `Rails::Generators::Actions` [`#rails_command`](https://github.com/rails/rails/blob/8e504b017a68981076746c2d19fd23c788e4293e/railties/lib/rails/generators/actions.rb#L391) or [`#execute_command`](https://github.com/rails/rails/blob/8e504b017a68981076746c2d19fd23c788e4293e/railties/lib/rails/generators/actions.rb#L460) to automatically set `capture` to `true` if `options[:quiet]` is present. But these are public API methods and I am not sure it makes sense to change these since the `quiet` option seems internal to the rails app generator itself.

### Checklist

### DIFF
--- a/railties/test/generators/actions_test.rb
+++ b/railties/test/generators/actions_test.rb
@@ -569,6 +569,13 @@ class ActionsTest < Rails::Generators::TestCase
     assert_match(/1234567890/, error.message)
   end
 
+  test "rails_command with quiet option" do
+    generator(default_arguments, quiet: true)
+    assert_runs "rails new myapp", capture: true do
+      action :rails_command, "new myapp"
+    end
+  end
+
   test "route should add route" do
     run_generator
     route_commands = ['get "foo"', 'get "bar"', 'get "baz"']
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -657,6 +657,11 @@ module Rails
         end
       end
 
+      def rails_command(command, command_options = {})
+        command_options[:capture] = true if options[:quiet]
+        super
+      end
+
       def bundle_install?
         !(options[:skip_bundle] || options[:pretend])
       end


railties/test/generators/app_generator_test.rb
```diff
@@ -762,7 +762,7 @@ def test_pretend_option
   end
 
   def test_quiet_option
-    output = run_generator [File.join(destination_root, "myapp"), "--quiet"]
+    output = run_generator [File.join(destination_root, "myapp"), "--quiet", "--no-skip-bundle"]
```
zzak, line 765: Not skipping bundler is very slow, is there another way to test this?
https://buildkite.com/rails/rails/builds/116660#01950c25-dfd8-4e24-a1c2-0ffab2fc1ebd/2249-2253

jeromedalbert: Oh, that is a very slow test indeed, thanks for pointing that out! There is another way to test this, but it is hacky.

Rails commands that are currently noisy, like `turbo:install stimulus:install`, need bundler not to be skipped, because they check for the [`bundle_install?`](https://github.com/rails/rails/blob/976e5180a00de26db65b08ac1fd7ab6b679cef9f/railties/lib/rails/generators/app_base.rb#L672-L674) method like so:

https://github.com/rails/rails/blob/976e5180a00de26db65b08ac1fd7ab6b679cef9f/railties/lib/rails/generators/app_base.rb#L719-L723

My workaround is to stub `bundle_install?` to return true, but only for the `run_hotwire` method, so only one rails command is run instead of multiple ones, and no `bundle install` command is run.

I have force-pushed the new test. It is a hack for sure, but it runs in less than 2 seconds on my local, compared to the old one which took 17 seconds. Another alternative is to not test this PR, which doesn't seem ideal, but I can do it if needed.

jeromedalbert: @zzak My new test now takes 3.1 seconds instead of 29.6, hopefully that is good enough:

https://buildkite.com/rails/rails/builds/117119#01955821-eb17-43fb-99b7-7dbad6db1b14/1211-1219

railties/test/generators/app_generator_test.rb
```diff
@@ -776,6 +776,17 @@ def test_quiet_option
...
   end
 
+  def test_quiet_option_quiets_rails_commands
+    generator = generator_class.new(
+      [File.join(destination_root, "myapp")], ["--quiet"], destination_root: destination_root
+    )
+
+    generator.stub :bundle_install?, -> { caller.grep(/run_hotwire/).any? } do
+      output = capture(:stdout) { generator.invoke_all }
+      assert_empty output
```
zzak, line 786:
```suggestion
      output = capture(:stdout) { run_generator }
      assert_empty output
      assert_empty @bundle_commands
```

How about this?

railties/test/generators/app_generator_test.rb
```diff
@@ -776,6 +776,17 @@ def test_quiet_option
     assert_empty output
   end
 
+  def test_quiet_option_quiets_rails_commands
+    generator = generator_class.new(
+      [File.join(destination_root, "myapp")], ["--quiet"], destination_root: destination_root
+    )
+
+    generator.stub :bundle_install?, -> { caller.grep(/run_hotwire/).any? } do
```
zzak, line 784: Is `run_hotwire` a hard dependency here or is there another native command we can look for?
My concern is just that the name may change or be removed and this test will fail, which is unrelated.

jeromedalbert: The only 3 methods that run rails commands by default if `bundle_install?` is true are: `run_hotwire`, `run_solid`, and `run_javascript` (the first 2 being faster). 

I can only choose among them, they don't have a common caller besides being all called by Thor.

I just commit `{ caller.grep(/run_(hotwire|solid)/).any? }` to lower the future chances of failure, but  they are still non zero chances.

railties/test/generators/app_generator_test.rb
jeromedalbert: I don't think that works, doing the following:

```suggestion
      output = capture(:stdout) { run_generator }
      assert_empty output
      assert_nil @bundle_commands
```

makes the test pass, but it also passes if I revert my `app_base.rb` code change, making this a test that passes all the time no matter what. 😬

zzak: I mean, the fact that we need to stub `bundle_install?` at all is a smell to me.

I would rather we proxy a generator that create some noise in a command that we can explicitly test for.

There maybe a day when `run_hotwire` etc are removed, or could break in a separate PR and this test failing will be unrelated.

Sorry I don't have any concrete suggestions at the moment. I will try to come up with one though.

railties/test/generators/actions_test.rb
```diff
@@ -569,6 +569,13 @@ def test_initializer_should_write_date_to_file_with_block_in_config_initializers
     assert_match(/1234567890/, error.message)
   end
 
+  test "rails_command with quiet option" do
+    generator(default_arguments, quiet: true)
+    assert_runs "rails new myapp", capture: true do
+      action :rails_command, "new myapp"
+    end
+  end
```
jeromedalbert, line 577: Wow, thanks for remembering about this and finding a nicer and faster way to unit test this Zak! 🙇

I took the liberty to change your last commit from `"about"` to `"new myapp"`, which I think better reflects the use case I was trying to fix. Also removed the `run_generator` call, inspired by the terseness of the test on line 510, and it still worked. But I can roll back if you prefer the way it looked before.

zzak: No worries, sorry I didn't have time to write back and explain, have been under the weather. 😭

PR #54550, opened 2025-02-16
Fix `#recognize` inconsistency with `#serve`

### LINKED ISSUE #31152: Journey router adds single slash to request.script_name after mounting Engine at '/'
### Steps to reproduce

Given a Rails `routes.rb` file that references an Engine: 

```ruby
MyApp::Application.routes.draw do
  get '*path',
    to: 'before_engine_requests#process',
    as: 'before_engine_requests',
    constraints: lambda { |request|
      # everything is fine
      request.fullpath == '/before_engine_request'
    }

  mount Spree::Core::Engine, at: '/'

  get '*path',
    to: 'after_engine_requests#process',
    as: 'after_engine_requests',
    constraints: lambda { |request|
      # This will never match as at this point, the router prepends a second '/ '
      request.fullpath == '/before_engine_request' 
    }
end
``` 

And a routing spec that references the second `get` block: 

```ruby
  it 'lets the redirects controller handle it' do
    expect(get: '/after_engine_request').to route_to(
      controller: 'after_engine_requests', action: 'process', path: 'after_engine_request'
    )
  end
```

This spec will fail although the `get` constraint in the first example works fine. 

### Expected behavior

`request.fullpath` inside the route constraint lambda should evaluate to `/before_engine_request`

### Actual behavior

`request.fullpath` inside the second route constraint lambda evaluates to `//before_engine_request` (with two slashes in front).

### Probable cause

After the engine is mounted, the request's `SCRIPT_NAME` environment key is set to '/'. 

This only happens in `ActionDispatch::Journey::Router#recognize`: https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/journey/router.rb#L70

In `ActionDispatch::Journey::Router#call`, it has been fixed in https://github.com/rails/journey/commit/61ddc45c07c2b946a725bbdca8ef8 (ages ago), referencing Rails issues

rails/rails#2131
rails/rails#2230
rails/rails#2579
rails/rails#5342

Also, the IETF says SCRIPT_NAME should be an empty string (although the wording here is not entirely clear): https://tools.ietf.org/html/rfc3875#section-4.1.13

### System configuration
**Rails version**: 4.2.10 (and newer versions)

**Ruby version**: 2.3.5

skipkayhil:
### Motivation / Background

When routing a request, `#find_routes` will iterate through potential routes for the request and yield them to the passed block. `#serve` and `#recognize` both used `#find_routes`, but ended up handling things slightly differently. Both methods would modify the `script_name` and `path_info` of the request if the current route is not anchored. However, only `#serve` used `#chomp` and would restore the request's original `script_name` and `path_info` after each iteration. This inconsistency can cause issues as `#recognize` can end up not matching them even though `#serve` would.

### Detail

This commit fixes the issue by unifying more of the implementations of `#serve` and `#recognize`. Now, `#find_routes` implements the mutation/restoration of request params for both `#serve` and `#recognize` so that their behavior is consistent and request mutation isn't leaked between iterations.

One test had to be changed because it depended on the request mutation leaking outside of `#recognize`, so it was updated to make assertions during `#recognize` instead.

### Additional information

Fixes #31152

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/journey/router.rb
+++ b/actionpack/lib/action_dispatch/journey/router.rb
@@ -27,7 +27,7 @@ module ActionDispatch
       end
 
       def serve(req)
-        find_routes(req) do |route, parameters|
+        recognize(req) do |route, parameters|
           req.path_parameters = parameters
           req.route = route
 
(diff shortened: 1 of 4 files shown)


skipkayhil:
@byroot would you mind reviewing this one as well 🙇

actionpack/lib/action_dispatch/journey/router.rb
```diff
@@ -27,45 +27,20 @@ def eager_load!
...
-            rails_req.script_name = match.to_s
-            rails_req.path_info   = match.post_match
-            rails_req.path_info   = "/" + rails_req.path_info unless rails_req.path_info.start_with? "/"
-          end
-
-          yield(route, parameters)
-        end
+      def recognize(rails_req, &block)
+        find_routes(rails_req, &block)
       end
```
byroot, line 44: Why not just use an alias? Is the method even still needed?

skipkayhil: I wanted to keep the diff clean, but I agree that an alias (or inlining everything into `recognize`) makes sense to me.

I'll update with another commit

PR #54552, opened 2025-02-16
Fix generating `database.yml` when skipping Kamal

### LINKED ISSUE #54549: database.yml generated by `rails new --skip-kamal` is missing required cable and queue databases
### Steps to reproduce

Generate a Rails 8.0.1 app using `--skip-kamal`.

```
rails new blog --no-rc --skip-kamal
```

The generated `config/database.yml` has a `cache` database defined, but is missing the `cable` and `queue` databases that are required for the app to boot in production.

```yaml
production:
  primary:
    <<: *default
    # database: path/to/persistent/storage/production.sqlite3
  cache:
    <<: *default
    # database: path/to/persistent/storage/production_cache.sqlite3
    migrations_paths: db/cache_migrate
```

As a result, the app cannot be started in the production environment. An `ActiveRecord::AdapterNotSpecified` exception is raised on boot:

> The `cable` database is not configured for the `production` environment. (ActiveRecord::AdapterNotSpecified)

```
$ RAILS_ENV=production bin/rails console
/Users/mbrictson/.rbenv/versions/3.4.2/lib/ruby/gems/3.4.0/gems/activerecord-8.0.1/lib/active_record/database_configurations.rb:229:in 'ActiveRecord::DatabaseConfigurations#resolve_symbol_connection': The `cable` database is not configured for the `production` environment. (ActiveRecord::AdapterNotSpecified)

  Available database configurations are:

  default
development
test
production: primary, cache
```

### Expected behavior

Expected: the `config/database.yml` file generated by `rails new --skip-kamal` should contain production entries for `cable` and `queue` databases, just like it does for `cache`.

### Actual behavior

The required production database configurations are missing.

### System configuration

**Rails version**: Rails 8.0.1

**Ruby version**: ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin24]

fatkodima:
Fixes #54549.

### DIFF
--- a/railties/test/generators/app_generator_test.rb
+++ b/railties/test/generators/app_generator_test.rb
@@ -698,6 +698,16 @@ class AppGeneratorTest < Rails::Generators::TestCase
     assert_no_file ".kamal/secrets"
   end
 
+  def test_database_yml_skip_kamal
+    run_generator [destination_root, "--skip-kamal"]
+
+    assert_file("config/database.yml") do |content|
+      assert_match("db/queue_migrate", content)
+      assert_match("db/cache_migrate", content)
+      assert_match("db/cable_migrate", content)
+    end
+  end
+
   def test_inclusion_of_kamal_storage_volume
     generator [destination_root]
     run_generator_instance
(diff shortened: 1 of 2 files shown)


railties/test/generators/app_generator_test.rb
```diff
@@ -698,6 +698,17 @@ def test_kamal_files_are_skipped_if_required
     assert_no_file ".kamal/secrets"
   end
 
+  def test_database_yml_skip_kamal
+    # Regression test for #54549
```
skipkayhil, line 702: I think git history will usually indicate why a change was made so I'm not sure the comment is necessary

skipkayhil (approved):

PR #53705, opened 2024-11-22
Use `secret_key_base` from ENV or credentials when present locally

p8:
### Motivation / Background

Since 0c76f17f2dbf0d7ad90c890e6f334743cacce41f  a `tmp/local_secret.txt` is always used for `Rails.config.secret_key_base` in the test and development environments. Previously, `ENV["SECRET_KEY_BASE"]` or `Rails.application.credentials.secret_key_base` were used if present.

As the `secret_key_base` is used to generate signed IDs, it breaks Action Text attachments in development if they previously used the `secret_key_base` from ENV var or credentials. Accidentally deleting `tmp/local_secret.txt` will also break all local Action Text attachments.

The race condition in generating the local secret file, as reported in https://github.com/rails/rails/issues/53661, should be avoidable by setting `ENV["SECRET_KEY_BASE"]` or `Rails.application.credentials.secret_key_base`, but this currently doesn't work.

Fixes #53661 
Fixes #54277

### Details

When `ENV["SECRET_KEY_BASE"]` or `Rails.application.credentials.secret_key_base` is set for test or development, it should be used for the `Rails.config.secret_key_base`, instead of generating a `tmp/local_secret.txt`.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54554, opened 2025-02-17
Locally, prefer tmp/local_secret.txt over config/credentials.yml.enc

brendon:
### Motivation / Background

Reverts #53705, which restored broken behaviour begun in Rails 7.1 (but fixed again in 7.2 and 8.0) that would load secret_key_base from config/credentials.yml.enc instead of tmp/local_secret.txt.

Traditionally `tmp/local_secret.txt` was always used in local environments over `config/credentials.yml.enc`.

This PR provides a middle ground where if an explicit local environment credential file is used (e.g. `config/credentials/development.yml.enc`), and `secret_key_base` is provided in that file, then that will be used over `tmp/local_secret.txt`.

### Additional information

I had trouble coming up with a test for the case where `config/credentials/development.yml.enc` exists but there is no `secret_key_base` in it, just because of how these tests are written. None of them are actually checking the content of the credentials file as far as I could tell.

Concerns: @p8, @rafaelfranca

### DIFF
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -510,18 +510,19 @@ module Rails
 
       def secret_key_base
         @secret_key_base || begin
-          self.secret_key_base = if ENV["SECRET_KEY_BASE_DUMMY"]
+          self.secret_key_base = if ENV["SECRET_KEY_BASE"] && !ENV["SECRET_KEY_BASE_DUMMY"]
+            ENV["SECRET_KEY_BASE"]
+          elsif generate_local_secret?
             generate_local_secret
           else
-            ENV["SECRET_KEY_BASE"] ||
-              Rails.application.credentials.secret_key_base ||
+            Rails.application.credentials.secret_key_base ||
               (Rails.env.local? && generate_local_secret)
           end
         end
       end
 
       def secret_key_base=(new_secret_key_base)
-        if new_secret_key_base.nil? && Rails.env.local?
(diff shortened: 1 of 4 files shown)


brendon:
I'm hoping this gets looked at before 8.0.2 is released so that there isn't a release that changes people's development secret without them understanding why.

railties/lib/rails/application/configuration.rb
```diff
@@ -649,6 +650,11 @@ def generate_local_secret
 
           File.binread(key_file)
         end
+
+        def generate_local_secret?
+          Rails.env.local? && !root.join("config/credentials/#{Rails.env}.yml.enc").exist? ||
+            ENV["SECRET_KEY_BASE_DUMMY"]
```
p8, line 656: `secret_key_base` might not be defined in credentials:
```suggestion
          Rails.env.local? && !Rails.application.credentials.secret_key_base ||
            ENV["SECRET_KEY_BASE_DUMMY"]
```

Also, this will generate a local secret in production if `SECRET_KEY_BASE_DUMMY` is set:
```
false && false || true
=> true
```

```suggestion
          Rails.env.local? &&
            (!Rails.application.credentials.secret_key_base ||
             ENV["SECRET_KEY_BASE_DUMMY"])
```

rafaelfranca: Agree. Can someone update this patch so it cover this case as well?

rafaelfranca (changes requested):

PR #54559, opened 2025-02-17
Fix rails guides navigation [ci skip]

luciagirasoles:
### Motivation / Background

This Pull Request has been created because there is a bug on Rails guides navigation. When you try to use the 'Guides Index' dropdown it no longer functions after using Back button.

In the HTML layout, there's a button with id `guides-menu-button` that controls a div with id guides. The div is initially hidden with `style="display: none;"`.

The JavaScript error occurs because there's a mismatch between how the button's `aria-controls` attribute is being handled. Looking at the JavaScript code:

1. The code first sets `aria-controls` attribute based on a `data-aria-controls` attribute
2. Then it removes the `data-aria-controls` attribute
3. When trying to find the element later, it's using `getAttribute("aria-controls")` but something is going wrong with this process.

### Detail
I've made the following changes to fix the issue:

1. Removed the code that was trying to get values from `data-aria-controls` and `data-aria-expanded` attributes
2. Directly set the `aria-controls` attribute to 'guides' since we know that's the ID of the target element
3. Set the initial `aria-expanded state` to `'false'` since the menu starts hidden
4. Simplified the code to directly get the guides element by its ID

### Additional information 
To replicate the bug:
1. Start on the Docs index page: https://guides.rubyonrails.org/
2. Use the 'Guides Index' drop down menu in the top right to navigate to any other guide.
3. Hit back button in browser
4. BUG: when you try to use the 'Guides Index' dropdown it no longer functions

### Checklist

### DIFF
--- a/guides/assets/javascripts/guides.js
+++ b/guides/assets/javascripts/guides.js
@@ -80,15 +80,12 @@
     var guidesMenuButton = document.getElementById("guides-menu-button");
 
     // The link is now acting as a button (but still allows for open in new tab).
-    guidesMenuButton.setAttribute('role', 'button')
-    guidesMenuButton.setAttribute('aria-controls', guidesMenuButton.getAttribute('data-aria-controls'));
-    guidesMenuButton.setAttribute('aria-expanded', guidesMenuButton.getAttribute('data-aria-expanded'));
-    guidesMenuButton.removeAttribute('data-aria-controls');
-    guidesMenuButton.removeAttribute('data-aria-expanded');
+    guidesMenuButton.setAttribute('role', 'button');
+    // Set aria-controls to point to the guides element
+    guidesMenuButton.setAttribute('aria-controls', 'guides');
+    guidesMenuButton.setAttribute('aria-expanded', 'false');
 
-    var guides = document.getElementById(
-      guidesMenuButton.getAttribute("aria-controls")
-    );
+    var guides = document.getElementById('guides');
 
     var toggleGuidesMenu = function () {
       var nextExpanded = guidesMenuButton.getAttribute("aria-expanded") === "false";


guides/assets/javascripts/guides.js
```diff
@@ -80,15 +80,12 @@
...
-    guidesMenuButton.setAttribute('aria-expanded', guidesMenuButton.getAttribute('data-aria-expanded'));
-    guidesMenuButton.removeAttribute('data-aria-controls');
-    guidesMenuButton.removeAttribute('data-aria-expanded');
-
-    var guides = document.getElementById(
-      guidesMenuButton.getAttribute("aria-controls")
-    );
+    guidesMenuButton.setAttribute('role', 'button');
+    // Set aria-controls to point to the guides element
+    guidesMenuButton.setAttribute('aria-controls', 'guides');
```
Edouard-chin, line 85: Do you maybe know why we use a dataset in the html rather than using the `aria-controls` attribute ? https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/guides/source/layout.html.erb#L81

Edouard-chin (commented):
> When trying to find the element later, it's using getAttribute("aria-controls") but something is going wrong with this process.

It is because of the Turbo feature. When you click on the back button, the page displayed comes from the cache which was slightly modified by our script the first time it's loaded https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/guides/assets/javascripts/guides.js#L86-L87.

guides/assets/javascripts/guides.js
luciagirasoles: They were added here #51499 . You can see in the lines above that when it is a button it uses the attribute without `data-`, but when it is an `<a>` it does use it. I would change that link to a button to give it a semantic correction, but I think it does not add value to solve this bug.

Edouard-chin: I see. I was asking because if we change the html and use the right attribute from the start, then we don't even need this javascript code (and so we fix the bug). But maybe I'm missing the point of this javascript.

```diff
diff --git a/guides/source/layout.html.erb b/guides/source/layout.html.erb
index 8e4849a1be..a62f7286c0 100644
--- a/guides/source/layout.html.erb
+++ b/guides/source/layout.html.erb
@@ -78,7 +78,7 @@
         <ul class="nav">
           <li><a class="nav-item" id="home_nav" href="https://rubyonrails.org/">Home</a></li>
           <li class="guides-index guides-index-large">
-            <a href="index.html" id="guides-menu-button" data-aria-controls="guides" data-aria-expanded="false" class="guides-index-item nav-item">Guides Index</a>
+            <a href="index.html" id="guides-menu-button" aria-controls="guides" aria-expanded="false" role="button" class="guides-index-item nav-item">Guides Index</a>
             <div id="guides" class="clearfix" style="display: none;">
               <hr />
               <dl class="guides-section-container">
(END)
```

Edouard-chin:
Sorry that you were not added as co-author, but it seems like someone else fixed the issue in https://github.com/rails/rails/pull/54934

We can probably close this PR now

keegankb93:
> Sorry that you were not added as co-author, but it seems like someone else fixed the issue in #54934
> 
> We can probably close this PR now

I apologize. I did a search for open issues on 'guides', but for some reason this didn't show up, so I figured there was no active working PR.

AmandaPerino:
Hi all! This was a bounty requested by the Rails Foundation, and it's great that it was addressed in #54934 but I am still seeing the same issue in Chrome.

Edouard-chin:
The changes are now live on the [edge guides](https://edgeguides.rubyonrails.org/). For them to take effect on the [8.0 guide](https://guides.rubyonrails.org/v8.0/index.html), we'll need to wait for a release in the 8.0 serie

AmandaPerino:
Thanks @Edouard-chin & @zzak! I've reached out to @luciagirasoles; she is currently ooo so I'll close this on her behalf.

PR #54561, opened 2025-02-18
Fix `update_all` producing a broken query when using more than one join on the target:

### LINKED ISSUE #54560: Updates in joins crash in some cases that use a `through` association
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails" # passes

  # edge fails. this includes https://github.com/rails/rails/pull/54530 and https://github.com/rails/rails/pull/53950

  # commit before https://github.com/rails/rails/pull/53950, passes
  # gem "rails", github: "rails/rails", ref: "4e5a13a55b5960c9a35afdcecd1120bc926d377c"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

  end

    t.integer :post_id
  end

  create_table :likes, force: true do |t|
    t.integer :comment_id
    t.integer :reply_id
    t.boolean :seen, default: false
  end

  create_table :replies, force: true do |t|
  end
end

  has_many :comments
  has_many :likes, through: :comments
end

  belongs_to :post
  has_many :likes
end

class Like < ActiveRecord::Base
  belongs_to :comment
  belongs_to :reply, optional: true
  scope :for_post, lambda { |post| joins(:comment).where(comment: { post_id: post.id }) }
end

class Reply < ActiveRecord::Base
end

  def test_association_stuff
    post = Post.create!
    comment = post.comments.create!
    like = comment.likes.create!

    post.likes.update_all(seen: true) # fine
    post.likes.left_joins(:reply).update_all(seen: true) # 💥 - workaround is to not use the through: association
    post.likes.for_post(post).update_all(seen: true) # 💥 - fixed by removing the double-join
  end
end

```

### Expected behavior

Tests should complete without crashing.

### Actual behavior

Two of the updates fail. Here are the errors:

```
BugTest#test_association_stuff:
ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: likes.reply_id:
UPDATE "likes" SET "seen" = ? FROM "comments" LEFT OUTER JOIN "replies" ON "replies"."id" = "likes"."reply_id"  WHERE "likes"."comment_id" = "comments"."id" AND "comments"."post_id" = ?
```

```
BugTest#test_association_stuff:
ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: likes.comment_id:
UPDATE "likes" SET "seen" = ? FROM "comments" INNER JOIN "comments" "comment" ON "comment"."id" = "likes"."comment_id"  WHERE "likes"."comment_id" = "comments"."id" AND "comments"."post_id" = ? AND "comment"."post_id" = ?
```

Our app, which runs on postgres, had a similar issue come up when trying to update to edge Rails.

This issue was introduced by https://github.com/rails/rails/pull/53950 and was not fixed by https://github.com/rails/rails/pull/54530

### System configuration

**Rails version**:  edge

**Ruby version**:  3

Edouard-chin:
- Fix #54560
- `update_all` would produce a broken query when using more than one JOIN on the target.

```ruby
Developer.join(:computer, :mentor).update_all(ready: true)
```

### Before

```sql
UPDATE developers SET ready = true FROM computers JOIN mentors on mentors.id = developers.mentor_id WHERE developer.computer_id = computer.id; # We can't make a JOIN on the developer table, this should in the FROM clause.
# We can't make a join on the target, that's what the FROM clause is used for.
```

### After

```sql
UPDATE developers SET ready = true FROM computers, mentors WHERE developer.computer_id = computer.id AND developers.mentor_id = mentors.id;
```

### Checklist

### DIFF
--- a/activerecord/lib/arel/visitors/sqlite.rb
+++ b/activerecord/lib/arel/visitors/sqlite.rb
@@ -64,7 +64,7 @@ module Arel # :nodoc: all
             # The SQLite3 dialect isn't flexible enough to allow anything other than a inner join
             # for the first join:
             #   UPDATE table SET .. FROM joined_table WHERE ...
-            (o.relation.right.first.is_a?(Arel::Nodes::InnerJoin))
+            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
             o
           else
             super
--- a/activerecord/lib/arel/visitors/sqlite.rb
+++ b/activerecord/lib/arel/visitors/sqlite.rb
@@ -30,7 +30,12 @@ module Arel # :nodoc: all
 
             collector << " FROM "
             first_join, *remaining_joins = o.relation.right
-            visit first_join.left, collector
+            from_items = remaining_joins.extract! do |join|
+              join.right.expr.right.relation == o.relation.left
+            end
+
+            from_where = [first_join.left] + from_items.map(&:left)
+            collect_nodes_for from_where, collector, " ", ", "
 
             if remaining_joins && !remaining_joins.empty?
               collector << " "
@@ -40,7 +45,8 @@ module Arel # :nodoc: all
               end
             end
 
-            collect_nodes_for [first_join.right.expr] + o.wheres, collector, " WHERE ", " AND "
+            from_where = [first_join.right.expr] + from_items.map { |i| i.right.expr }
+            collect_nodes_for from_where + o.wheres, collector, " WHERE ", " AND "
           else
             collector = visit o.relation, collector
             collect_nodes_for o.values, collector, " SET "
--- a/activerecord/lib/arel/visitors/postgresql.rb
+++ b/activerecord/lib/arel/visitors/postgresql.rb
@@ -64,7 +64,7 @@ module Arel # :nodoc: all
             # The PostgreSQL dialect isn't flexible enough to allow anything other than a inner join
             # for the first join:
             #   UPDATE table SET .. FROM joined_table WHERE ...
-            (o.relation.right.first.is_a?(Arel::Nodes::InnerJoin))
+            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
             o
           else
             super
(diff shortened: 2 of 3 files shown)


ghiculescu:
This fixes the crashes in our test suite that led to https://github.com/rails/rails/issues/54560

But at least one of the tests that was crashing, is now failing. The relevant `update_all` is updating 0 rows. You can simulate this in the initial test script pointed to this commit:

```ruby
    assert_equal 1, post.likes.update_all(seen: true) # fine
    assert_equal 1, post.likes.left_joins(:reply).update_all(seen: true) # 💥
    assert_equal 1, post.likes.for_post(post).update_all(seen: true) # fine
```

It looks like the left outer join has been replaced with an inner join?

image
image

byroot:
Yeah, I'm tempted to just give up on this. But yeah, let's give it one more try. Let me know once you handle the left join.

Edouard-chin:
> let's give it one more try.

🤞 

I pushed a commit with the change related to the left_join issue with some explanation.

activerecord/lib/arel/visitors/sqlite.rb
```diff
@@ -58,7 +64,7 @@ def prepare_update_statement(o)
             # The SQLite3 dialect isn't flexible enough to allow anything other than a inner join
             # for the first join:
             #   UPDATE table SET .. FROM joined_table WHERE ...
-            (o.relation.right.first.is_a?(Arel::Nodes::InnerJoin))
+            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
```
byroot, line 67: I'm not sure I understand that `||` condition, could you explain what this is trying to match for?

Edouard-chin: When we have a join on the target (the table we are updating), the joined table has to be in the `FROM` clause for the query to be valid. This is what the first commit in this PR fixes.

By introducing this fix, we now hit the flexibility issue you previously mentioned when a left outer join is used. 
What this `||` checks, is whether there is any left joins on the target which we have to workaround and resort to making a subselect query instead.

This check is only required on left joins on the target (as we have to translate that join into a FROM and we hit the flexibility issue). Issuing a left outer join on another table is fine since we don't have to translate it for the query to be syntax valid.

byroot: > What this `||` checks, is whether there is any left joins on the target which we have to workaround and resort to making a subselect query instead.

That is where I'm confused, this `if` condition is for case where we don't have to fallback to a a subquery. So why `||`, wasn't this meant to be `&&`? (Sorry, I'm not the best with Arel API)

PR #54562, opened 2025-02-18
Update getting_started.md [ci skip]

dsichkar:
Entering just IP address in the browser gives 404 page not found


fatkodima:
It works as expected.

PR #54570, opened 2025-02-19
Speed up `password_digest` in fixtures templates

Uaitt:
When you launch one of the following commands:
 - `bin/rails g model user name password:digest`
 - `bin/rails g authentication`

you end up with fixtures that create a `password_digest` with `BCrypt::Password.create("...")`.

Since those digests will only be used in tests, it might be a good idea to specify the minimum allowed bcrypt cost so that the digests will take less time to be computed, and as a result tests will run faster.


byroot:
I think a better solution is to set `BCrypt::Engine.cost = BCrypt::Engine::MIN_COST` in test.

byroot:
Looks like we've been doing so for 13 years: https://github.com/rails/rails/pull/8216

Uaitt:
@byroot my bad, I completely forgot to even think about that possibility :)

Sorry for the unnecessary noise

PR #54573, opened 2025-02-19
Make filters in `rails routes` work on routes from Engines

dennispaagman:
### Motivation / Background

I noticed that the filters that you can give as arguments to`rails routes`:

https://github.com/rails/rails/blob/972a52d127ae90915fe68a6802c3213e28d80d45/railties/lib/rails/commands/routes/routes_command.rb#L8-L9

Do not actually filter routes that come from engines and gives false messages that no routes are found.

### Example

I have an app with the [Avo gem](https://github.com/avo-hq/avo) as engine, which has some routes, for example these:

```
Routes for Avo::Engine:
media_library_index GET    /media-library(.:format)     avo/media_library#index
      media_library GET    /media-library/:id(.:format) avo/media_library#show
                    PATCH  /media-library/:id(.:format) avo/media_library#update
                    PUT    /media-library/:id(.:format) avo/media_library#update
                    DELETE /media-library/:id(.:format) avo/media_library#destroy
       attach_media GET    /attach-media(.:format)      avo/media_library#attach
```

but when I apply a filter they are not showing up.

```
❯ bin/rails routes -c avo/media_library
No routes were found for this controller.
For more information about routes, see the Rails guide: https://guides.rubyonrails.org/routing.html.
```

This PR fixes that by making sure all engines routes are parsed/loaded before applying the filters to them as well.

With this fix the filter works as intended:

```
❯ bin/rails routes -c avo/media_library
Routes for Avo::Engine:
             Prefix Verb   URI Pattern                  Controller#Action
media_library_index GET    /media-library(.:format)     avo/media_library#index
      media_library GET    /media-library/:id(.:format) avo/media_library#show
                    PATCH  /media-library/:id(.:format) avo/media_library#update
                    PUT    /media-library/:id(.:format) avo/media_library#update
                    DELETE /media-library/:id(.:format) avo/media_library#destroy
       attach_media GET    /attach-media(.:format)      avo/media_library#attach
```

### Additional information

I also made a slight change in the format itself, it now adds the headers for each engine individually, as they would not show up if only routes from an engine match and it also felt nice to have them in all situations.

It also removes a couple of redundant collect/map calls by only mapping the routes to `RouteWrapper` once and only calling `collect_routes` once.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Make filters work on Engine routes when using `rails routes`.
+
+    *Dennis Paagman*
+
 *   Add headers to engine routes inspection command
 
     *Petrik de Heus*
--- a/actionpack/lib/action_dispatch/routing/inspector.rb
+++ b/actionpack/lib/action_dispatch/routing/inspector.rb
@@ -72,26 +72,29 @@ module ActionDispatch
     # not use this class.
     class RoutesInspector # :nodoc:
       def initialize(routes)
-        @engines = {}
-        @routes = routes
+        @routes = wrap_routes(routes)
       end
 
       def format(formatter, filter = {})
-        routes_to_display = filter_routes(normalize_filter(filter))
-        routes = collect_routes(routes_to_display)
-        if routes.none?
-          formatter.no_routes(collect_routes(@routes), filter)
+        normalized_filter = normalize_filter(filter)
+        filtered_routes = filter_routes(@routes, normalized_filter)
+
+        if filtered_routes.values.all?(&:empty?)
+          formatter.no_routes(@routes, filter)
+
           return formatter.result
         end
 
-        formatter.header routes
-        formatter.section routes
+        filtered_routes.each do |name, routes|
+          next if normalized_filter && routes.none?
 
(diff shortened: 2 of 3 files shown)


gmcgibbon (commented):
Thanks for your patience on this. I think the original routing inspector has a few engine-specific issues I'd like to address while we're here:
- Engine route unfurling is lazy and filtering can affect the engines that are displayed
- Messaging when an application has app routes and multiple engines is a little confusing, and we should show consistent messaging per route set
- In the current implementation, what a "route" is is hard to follow (wrapped or not wrapped)

This is a bit complicated, so I've gone ahead and addressed all of this and the filtering bug here: https://github.com/rails/rails/compare/main...Shopify:rails:engine_routes_filter

If it is ok with you, I'd like to proceed with this patch, and give you credit since I used your patch as inspiration.

dennispaagman:
@gmcgibbon yes of course, please go ahead! Thanks for having a look and making it even btter, shall I close this PR or do you want to wait until yours is up/ready?

gmcgibbon:
I opened https://github.com/rails/rails/pull/55752. It will close this once merged. Thanks!

PR #53974, opened 2024-12-16
[Fixes #53758] Raise `ActiveRecord::ReadOnlyError` on lock query while preventing writes

joshuay03:
Raise `ActiveRecord::ReadOnlyError` when attempting a locking query in readonly mode i.e. `prevent_writes: true`. The locks covered by this logic are database/adapter-specific and ideally include only those that the database itself would raise an error for if attempted in a readonly session, such as on a replica.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54580, opened 2025-02-20
[Fix #53758] Raise `ActiveRecord::ReadOnlyError` when pessimistically locking with a readonly role

### LINKED ISSUE #53758: Error should be raised when pessimistically locking with reader role
### Steps to reproduce

In our development environment, we use the same PostgreSQL database for both the writer and reader roles. While testing some automatic role-switching middleware and relying on our test suite to detect writes in read-only mode, we encountered a hanging test. We traced the issue to a pessimistic lock being acquired using with_lock (`SELECT ... FOR UPDATE`) while using the reader role, followed by switching to the writer role to update the same row. This resulted in a deadlock, as the `UPDATE` transaction waited on the `SELECT ... FOR UPDATE` lock, and vice versa.

From my understanding, PostgreSQL would have raised an error (`ERROR:  cannot execute SELECT FOR UPDATE in a read-only transaction`) if this had been attempted in a read-only role, as it would in our production environment. Therefore, I believe Rails could potentially raise an error in this scenario as well (maybe with differing behavior based on the adapter).

Please let me know if a PR to raise an error in this case would be accepted and I can send one 👍🏽

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "pg", "~> 1.1"
end

require "minitest/autorun"

# Note: This script requires a local PostgresQL database `primary`
config = YAML.load(<<~YAML)
  development:
    primary:
      adapter: postgresql
      database: primary
    replica:
      adapter: postgresql
      database: primary
      replica: true
YAML
ActiveRecord::Base.configurations = config

ActiveRecord::Base.establish_connection :primary

  create_table :posts, force: :cascade do |t|
    t.text :title
  end
end

  primary_abstract_class

  connects_to database: { writing: :primary, reading: :replica }
end

end

  def test_pessimistic_lock_with_database_switch
    post = Post.create!(title: "original")
    ApplicationRecord.connected_to(role: :reading) do
      post.with_lock do
        ApplicationRecord.connected_to(role: :writing) do
          post.update!(title: "updated")
        end
      end
    end
  end
end
```

### Expected behavior

An error should be raised telling me that a lock cannot be obtained when in readonly mode, similar to: https://github.com/rails/rails/blob/07bf01aba4904ff2f9eb5cfd11b33c5d46db26a8/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L208

### Actual behavior

It hangs indefinitely.

### System configuration

**Rails version**: Edge

**Ruby version**: 3.4.0preview2

joshuay03:
### Motivation / Background

An alternative approach to https://github.com/rails/rails/pull/53974 which fixes https://github.com/rails/rails/issues/53758.

### Detail

Adds a condition in `ActiveRecord::Locking::Pessimistic#lock!` (also used by `#with_lock`) and `ActiveRecord::Relaton#exec_queries` to check if we're currently in the context of a readonly role (`prevent_writes: true`) and raise an error if so.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/locking/pessimistic.rb
+++ b/activerecord/lib/active_record/locking/pessimistic.rb
@@ -67,6 +67,10 @@ module ActiveRecord
       # or pass true for "FOR UPDATE" (the default, an exclusive row lock). Returns
       # the locked record.
       def lock!(lock = true)
+        if self.class.current_preventing_writes
+          raise ActiveRecord::ReadOnlyError, "Lock query attempted while in readonly mode"
+        end
+
         if persisted?
           if has_changes_to_save?
             raise(<<-MSG.squish)
@@ -79,6 +83,7 @@ module ActiveRecord
 
           reload(lock: lock)
         end
+
         self
       end
 
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Raise `ActiveRecord::ReadOnlyError` when pessimistically locking with a readonly role.
+
+    *Joshua Young*
+
 *   Fix using the `SQLite3Adapter`'s `dbconsole` method outside of a Rails application.
 
     *Hartley McGuire*
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -1413,6 +1413,10 @@ module ActiveRecord
       end
 
       def exec_queries(&block)
+        if lock_value && model.current_preventing_writes
+          raise ActiveRecord::ReadOnlyError, "Lock query attempted while in readonly mode"
+        end
+
         skip_query_cache_if_necessary do
           rows = if scheduled?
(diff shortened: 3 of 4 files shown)


activerecord/lib/active_record/locking/pessimistic.rb
```diff
@@ -67,6 +67,10 @@ module Pessimistic
       # or pass true for "FOR UPDATE" (the default, an exclusive row lock). Returns
       # the locked record.
       def lock!(lock = true)
+        if self.class.current_preventing_writes
+          raise ActiveRecord::ReadOnlyError, "Lock query attempted while in readonly mode"
```
joshuay03, line 71: This is the matching write query error for reference: https://github.com/rails/rails/blob/ffc8c47bee784ed635899765cdb8a52ff961b6f2/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L195

PR #54583, opened 2025-02-21
Do cheap/happy check first in `verified_request?`

skipkayhil:
Since Rails 5.2, Action Controller enables `protect_from_forgery` by default. `verified_request?` is called in a `before_action` to perform a series of checks to ensure the request should proceed.

This commit reorders the checks so that the cheaper ones happen first, which allows the method to return more quickly in cases that the request does not need to be protected (GET/HEAD requests).

In the `r10k` benchmark, `verified_request?` shows up as ~2% in a profile before this change and does not show up afterwards.


byroot (approved):

byroot:
No, just me remembering that you don't have commit rights.

PR #54585, opened 2025-02-21
Deserialize column defaults

### LINKED ISSUE #54556: ActiveModel attribute with `default` requires to handle a `String` during deserialization
### Steps to reproduce

Here's a [gist](https://gist.github.com/Vagab/088bdd5d36e873dd8a13aaab19c9b75c)
```ruby
    t.integer :field, default: 0
  end
end

class FieldType < ActiveModel::Type::Value
  VALUES = { 0 => "zero" }

  def type
    :field_type
  end

  def deserialize(value)
    VALUES.fetch(value) unless value.nil?
  end

  def serialize(value)
    VALUES.key(value)
  end
end

  attribute :field, FieldType.new
end

  def test_association_stuff
    post = Post.create!

    assert post.field == "zero"
  end
end
```

### Expected behavior

I would expect that the value I receive in `deserialize` is an `Integer`, and the script runs fine

### Actual behavior

The value I receive in `deserialize` is a `String`, and the script fails on `VALUES.fetch`.
As far as I understand this is "intentional", from what I gathered playing with it it seems that all the values that an attribute gets from the db are internally represented as strings. I think I would expect those to be cast to the column type as specified in the database before passed into `deserialize`. But I'm not sure if that's a viable option. In any case I think this is an unexpected behaviour, at least for me, given that it works fine if the column does not have a default.

### System configuration

**Rails version**:
Rails 8.0.1

**Ruby version**: 
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin24]

byroot:
Fix: https://github.com/rails/rails/issues/54556

When getting an integer value from the database, all adapters will create return a real integer. But when we initialize an attribute from the column defaults, the attribute is given a string, that's inconsistent.

### DIFF
--- a/activerecord/lib/active_record/attributes.rb
+++ b/activerecord/lib/active_record/attributes.rb
@@ -244,7 +244,10 @@ module ActiveRecord
           # TODO: Remove the need for a connection after we release 8.1.
           attributes_hash = with_connection do |connection|
             columns_hash.transform_values do |column|
-              ActiveModel::Attribute.from_database(column.name, column.default, type_for_column(connection, column))
+              type = type_for_column(connection, column)
+              default = column.default
+              default = type.deserialize(default) unless default.nil?
+              ActiveModel::Attribute.from_database(column.name, default, type)
             end
           end
 
--- a/activerecord/test/cases/adapters/postgresql/geometric_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/geometric_test.rb
@@ -131,11 +131,11 @@ class PostgresqlPointTest < ActiveRecord::PostgreSQLTestCase
   end
 
   def test_legacy_default
-    assert_equal [12.2, 13.3], PostgresqlPoint.column_defaults["legacy_y"]
-    assert_equal [12.2, 13.3], PostgresqlPoint.new.legacy_y
+    assert_equal ActiveRecord::Point.new(12.2, 13.3), PostgresqlPoint.column_defaults["legacy_y"]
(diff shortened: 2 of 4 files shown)


byroot:
I believe this change is positive, but I fear it may break existing apps. @matthewd @rafaelfranca any opinions?

activerecord/lib/active_record/attributes.rb
```diff
@@ -244,7 +244,10 @@ def _default_attributes # :nodoc:
           # TODO: Remove the need for a connection after we release 8.1.
           attributes_hash = with_connection do |connection|
             columns_hash.transform_values do |column|
-              ActiveModel::Attribute.from_database(column.name, column.default, type_for_column(connection, column))
+              type = type_for_column(connection, column)
+              default = column.default
+              default = type.deserialize(default) unless default.nil?
+              ActiveModel::Attribute.from_database(column.name, default, type)
```
Edouard-chin, line 250: Would it make sense to deserialize the column default when we build the column instead, so that `Post.columns#default` also returns the deserialized default ?

robinator:
@byroot

This PR change is causing a bug in our code when we use `store_accessor` on its own to store attributes in a jsonb column.  eg
```ruby
  store_accessor :configuration, %i[ host port ]
```
What's weird is that once a record is created, subsequent instantiations of new records have a cached version of `configuration` set.  This bug disappears if we just call:
```
store :configuration, accessors: %i[ host port ]
```
directly.  I'm still trying to find the exact issue but thought it'd be helpful to flag in case someone else sees as well.

byroot:
Thanks. Please open a dedicated issue once you have a repro.

PR #42312, opened 2021-05-28
Expose BatchEnumerator attributes

etiennebarrie:
`BatchEnumerator` is a simple object storing a few parameters coming from `ActiveRecord::Batches#in_batches` when it's called without a block.

Sometimes, it's useful to access those values, in our case, a library receiving a `BatchEnumerator` from application code.

Co-authored-by: Adrianna Chang <adrianna.chang@shopify.com> @adrianna-chang-shopify 

### Summary

Adds the attribute readers, and a method to return `batch_size` from the `@of` instance variable.

cc @byroot @rafaelfranca

PR #54587, opened 2025-02-21
Expose all `BatchEnumerator` attributes

fatkodima:
Follow up to #42312.

There are already some attributes exposed by `BatchEnumerator`. In my gem, I currently have a need to be able to access all of the attributes from `BatchEnumerator` to do some more logic based on them.

* `cursor`
* `order`
* `use_ranges`

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Expose `cursor`, `order` and `use_ranges` attributes for `BatchEnumerator`
+
+    *fatkodima*
+
 *   Fix `ActiveRecord::QueryMethods#in_order_of` when passing an out-of-range Integer
 
     To match the behavior of the `Enumerable` version, `in_order_of` now ignores an out-of-range Integer.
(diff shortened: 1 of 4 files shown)


zzak:
> In my gem, I currently have a need to be able to access all of the attributes from `BatchEnumerator` to do some more logic based on them.
>
> * `cursor`
> * `order`
> * `use_ranges`

I think typically any "gem" is not sufficient use-case to expose API, but if there is a case inside Rails is a good example.
But it was merged previously, so maybe.

Can you point to a particular piece of code that could benefit from these being public?

fatkodima:
Changes are not released yet, only in a branch.

Monkey patch https://github.com/fatkodima/online_migrations/blob/reimplement-bg-data-migrations/lib/online_migrations/active_record_batch_enumerator.rb and usage https://github.com/fatkodima/online_migrations/blob/392b15cd5e7973ab0f7b9013dd71d0aa7ba46903/lib/online_migrations/background_data_migrations/migration_job.rb#L99-L100

byroot (approved):

PR #54589, opened 2025-02-21
[Experiment] Try to modify all tests making use of a dynamic controller and action segment route

Edouard-chin:
Sorry for the draft PR, I'm mostly trying to help move the needle towards removing a almost 10 year old deprecation introduced in 6520ea5f7e2 where using a `:controller/:action` route leads to a bunch of issues and security problem.

The Rails codebase has a lot of such routes which makes it difficult to remove that deprecation until we fix all tests.

------------

There is basically around 1500 tests that relies on this feature.
Around 1000 of them uses the same route set which is what this PR address.
The remaining 500 ish tests uses different a route set (they do `routes.draw` directly in the test or somewhere in the test suite setup).

I made multiple commits to help understand what's going on and also left the code snippet in one commit to highlight how it was done.

-------------

This PR isn't really meant to be merged as is, but mainly used as a starting point. Not sure how we'd want to proceed , but I'm happy to help put the effort :).


Edouard-chin:

PR #54594, opened 2025-02-22
Add column types to `ActiveRecord::Result` for SQLite3

ankane:
### Motivation / Background

This allows `ActiveRecord::Result#cast_values` to cast dates and times with SQLite like it does with PostgreSQL. This is useful for queries that use `ActiveRecord::Base.lease_connection.select_all`. Column types can be referenced by column name or position like with PostgreSQL.

Relevant PostgreSQL adapter code: https://github.com/rails/rails/blob/v8.0.1/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb#L177-L184

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
@@ -87,11 +87,7 @@ module ActiveRecord
                 stmt.step
                 ActiveRecord::Result.empty
               else
-                types = {}
-                stmt.columns.zip(stmt.types).each_with_index do |(c, t), i|
-                  types[c] = types[i] = type_map.lookup(t)
-                end
-                ActiveRecord::Result.new(stmt.columns, stmt.to_a, types.freeze)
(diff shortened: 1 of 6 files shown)


activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
```diff
@@ -100,7 +104,11 @@ def perform_query(raw_connection, sql, binds, type_casted_binds, prepare:, notif
                   stmt.step
                   ActiveRecord::Result.empty
                 else
-                  ActiveRecord::Result.new(stmt.columns, stmt.to_a)
+                  types = {}
+                  stmt.columns.zip(stmt.types).each_with_index do |(c, t), i|
+                    types[c] = types[i] = type_map.lookup(t)
+                  end
+                  ActiveRecord::Result.new(stmt.columns, stmt.to_a, types.freeze)
```
byroot, line 111: I get that you copied what was done for PG, but this implementation isn't great performance wise. I probably never noticed it because I never profile with pg.

I'll experiment a few things.

ankane:
Thanks @byroot!

PR #52467, opened 2024-07-31
Rename `text_area` to `textarea` and `rich_text_area` to `rich_textarea`

seanpdoyle:
### Motivation / Background

Follow-up to [#52432][]
Related to [#52430][]

Helpers like `text_area_tag` construct [textarea][] elements. This commit renames all view helper and form builder methods to replace occurrences of `text_area` with `textarea`.

### Detail

In the same style, this commit also renames the `rich_text`-prefixed helpers to utilize a `textarea` suffix instead of the existing `text_area` suffix.

To preserve backwards compatibility, this commit defines aliases for the existing `text_area` format.

[#52432]: https://github.com/rails/rails/pull/52432
[#52430]: https://github.com/rails/rails/issues/52430
[textarea]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54597, opened 2025-02-23
[ci skip] Fix incorrect `rich_textarea` method in Rails Guides

afzalrahman11:
Fixed incorrect method `rich_textarea` as it should be `rich_text_area`.


byroot:
Thanks for your PR, but this isn't a mistake: https://github.com/rails/rails/pull/52467

PR #54598, opened 2025-02-23
Fix `update_all` when right expression is an instance of Arel::Nodes::Equality

chaadow:
Followup to #54561 

in https://github.com/rails/rails/pull/54561 it was assumed that the right expression would always respond to `#relation`. However in some instances the right expression returns an `Arel::Nodes::Equality` instance which does not respond to `relation`

Not sure if this is the right fix @Edouard-chin @byroot. But it did fix my query on a production app. I'll try to reproduce with a failing test. ( sorry I'm still not well used to testing the `activerecord` gem )

image

### Checklist

### DIFF
--- a/activerecord/lib/arel/visitors/postgresql.rb
+++ b/activerecord/lib/arel/visitors/postgresql.rb
@@ -30,7 +30,7 @@ module Arel # :nodoc: all
             collector << " FROM "
             first_join, *remaining_joins = o.relation.right
             from_items = remaining_joins.extract! do |join|
-              join.right.expr.right.relation == o.relation.left
+              join.right.expr.right.try(:relation) == o.relation.left
             end
 
             from_where = [first_join.left] + from_items.map(&:left)
@@ -64,7 +64,7 @@ module Arel # :nodoc: all
             # The PostgreSQL dialect isn't flexible enough to allow anything other than a inner join
             # for the first join:
             #   UPDATE table SET .. FROM joined_table WHERE ...
-            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
+            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.try(:relation) != o.relation.left })
             o
           else
             super
--- a/activerecord/test/cases/relation/update_all_test.rb
+++ b/activerecord/test/cases/relation/update_all_test.rb
@@ -179,6 +179,14 @@ class UpdateAllTest < ActiveRecord::TestCase
     assert_equal pets.count, pets.update_all("name = toys.name")
   end
 
+
+  def test_update_all_with_multiple_condition_join
+    relation = Toy.left_joins(:sponsors)
+
+    assert_equal(3, relation.count)
+    assert_equal(3, relation.update_all(name: "sponsor-toy"))
+  end
+
   def test_update_all_with_left_outer_joins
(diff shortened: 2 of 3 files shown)


activerecord/lib/arel/visitors/postgresql.rb
```diff
@@ -64,7 +64,7 @@ def prepare_update_statement(o)
             # The PostgreSQL dialect isn't flexible enough to allow anything other than a inner join
             # for the first join:
             #   UPDATE table SET .. FROM joined_table WHERE ...
-            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
+            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.try(:relation) != o.relation.left })
```
byroot, line 67: `try` is awful for performance, we shouldn't use that. We should check the type of the node instead.

activerecord/test/cases/relation/update_all_test.rb
```diff
@@ -171,6 +173,12 @@ def test_update_all_with_left_joins
     assert_equal pets.count, pets.update_all(name: "Bob")
   end
 
+  def test_update_all_with_scoped_has_many
```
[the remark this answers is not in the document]
chaadow: I mean those `Arel::Nodes::Grouping` or `Arel::Nodes::Equality` do not respond to `relation` but I'm afraid `Arel::Nodes::Attribute` is not enough to whitelist because that would lead to a weird conditional : 

```ruby
    (o.relation.right.all? do |join|
                right_join_right_expr = join.right.expr.right
                binding.irb

                join.is_a?(Arel::Nodes::InnerJoin) || (right_join_right_expr.is_a?(Arel::Attributes::Attribute) ? right_join_right_expr.relation != o.relation.left : true)
              end)
```

If I don't return `true` when the right expression (`right_join_right_expr`) is not an arel attribute, my test will fail.

activerecord/test/models/pet.rb
```diff
@@ -6,6 +6,7 @@ class Pet < ActiveRecord::Base
   self.primary_key = :pet_id
   belongs_to :owner, touch: true
   has_many :toys
+  has_many :custom_toys, -> { joins(:sponsors).merge(Sponsor.left_joins(:sponsor_club).where(sponsor_club: { name: "123" }).or(Sponsor.left_joins(:sponsor_club).where(sponsor_club: { name: "" }))) }, class_name: "Toy"
```
chaadow, line 9: This almost perfectly replicates the custom join that I have in my application.

byroot:
> I don't know if it's worth going the rabbit hole and whether we should fallback to a subselect when we detect a more complex join

If we're to try to continue, I think we should adopt an allow list approach, only optimize if we detect a simple case we know about.

chaadow:
@Edouard-chin thanks for the explanation. I understand what you say, but all the tests pass ( existing ones + mine) do you think you can push over my PR with a failing test so that it can help me better implement your theory. thanks!

I'll also try to reproduce an error, but it's weird that with the "respond_to? / try" approach all pass 

@byroot I'll make an attempt later this week or on the week end with your allow list approach

Edouard-chin:
I pushed a failing test. This can be reproduced either when joining on a table with STI, or a table that has default scope. (The test I added uses a join with a STI table).

chaadow:
@Edouard-chin @byroot I confirm #54722  has fixed the issue in my PR. Closing.

PR #51124, opened 2024-02-19
Map enums using array to the correct column type

lazaronixon:
### Motivation / Background

Previously enums without keywords were always mapped to integers and indexed by position, now it is converted based on the column type, it works for strings and database enums as well.

```ruby
# Before
enum status: { proposed: "proposed", written: "written", published: "published" }

# After
enum status: [:proposed, :written, :published]
```

### Detail

This Pull Request closes #47278

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54605, opened 2025-02-24
Map enums using array to the correct column type

### LINKED ISSUE #52604: Map enums using array to the correct column type
Previously enums without keywords were always mapped to integers and indexed by position, now it is converted based on the column type, it works for strings and database enums as well.

```ruby
# Before
enum status: { proposed: "proposed", written: "written", published: "published" }

# After
enum status: [:proposed, :written, :published]
```

We can't connect to the database when evaluating the `enum` call, so we would need to delay the database type inflection possibly to the first usage of the `enum`.

Related to #51124
Related to #47278

felixefelip:
### Motivation / Background
This Pull Request closes https://github.com/rails/rails/issues/52604

Previously enums without keywords were always mapped to integers and indexed by position, now it is converted based on the column type, it works for strings and database enums as well.

```ruby
# Before
enum status: { proposed: "proposed", written: "written", published: "published" }

# After
enum status: [:proposed, :written, :published]
```

### Detail
- I create the method `define_enum_values` to do the database type inflection and map the enum values only when the database is connected.
- In Enum inherited method, I had to update the `klass` of `EnumType` and set the `decorate_attribute` because it was just being executed in superclass https://github.com/felixefelip/rails/blob/7844162de0ac3cf75c21d346c2c7d5f67dd24208/activerecord/lib/active_record/enum.rb#L332

### Additional information
- Some things from this PR https://github.com/rails/rails/pull/51124 were reused, such as tests, CHANGELOG and the `enum_typed_pairs` method.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/enum.rb
+++ b/activerecord/lib/active_record/enum.rb
@@ -164,15 +164,16 @@ module ActiveRecord
   #   conversation.status = :unknown # 'unknown' is not a valid status (ArgumentError)
   module Enum
     def self.extended(base) # :nodoc:
-      base.class_attribute(:defined_enums, instance_writer: false, default: {})
+      base.class_attribute(:defined_enums_type, instance_writer: false, default: {})
     end
 
     class EnumType < Type::Value # :nodoc:
       delegate :type, to: :subtype
 
-      def initialize(name, mapping, subtype, raise_on_invalid_values: true)
+      def initialize(klass, name, initial_mapping, subtype, raise_on_invalid_values: true)
+        @klass = klass
         @name = name
-        @mapping = mapping
+        @initial_mapping = initial_mapping
         @subtype = subtype
         @_raise_on_invalid_values = raise_on_invalid_values
       end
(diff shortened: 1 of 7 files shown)


felixefelip:
Closing to fix a failed test.

activerecord/lib/active_record/enum.rb
```diff
@@ -207,46 +208,92 @@ def assert_valid_value(value)
...
+      def mapping
+        @mapping ||= define_enum_values(name, subtype, initial_mapping)
+      end
+
+      def subtype
+        @subtype ||= begin
+          klass.load_schema
+          @subtype
+        end
+      end
```
felixefelip, line 220: As the subtype is only assigned when the database is loaded in `decorate_attributes`. Sometimes the mapping can be called with the schema not loaded. The cases are: scopes  and when `defined_enums_type` is [called in fixtures](https://github.com/rails/rails/blob/f0dcf888efec6b4f0a4663a31591e5bf61cb21e9/activerecord/lib/active_record/fixture_set/table_row.rb#L141) and [uniqueness validator](https://github.com/rails/rails/blob/f0dcf888efec6b4f0a4663a31591e5bf61cb21e9/activerecord/lib/active_record/validations/uniqueness.rb#L22)

felixefelip:
I had a problem in a rebase that causes this PR to have a lot of other changed files, so the bot added a lot of tags and I don't have access to remove them.

MatheusRich:
@felixefelip please squash your commits.

activerecord/lib/active_record/enum.rb
```diff
@@ -207,46 +208,92 @@ def assert_valid_value(value)
...
 
       private
-        attr_reader :name, :mapping
+        attr_reader :initial_mapping
+        attr_writer :mapping
+
+        def define_enum_values(name, subtype, initial_mapping)
+          pairs = initial_mapping.respond_to?(:each_pair) ? initial_mapping.each_pair : enum_typed_pairs(name, initial_mapping, subtype)
+
+          # statuses = { }
```
MatheusRich, line 233:
```suggestion
```

felixefelip: I just moved this part [from here](https://github.com/rails/rails/blob/0ed3cd46d0038f8108ed92edb308c81fa6cc3730/activerecord/lib/active_record/enum.rb#L226)

But this comment seems really not to be useful

felixefelip:
> @felixefelip please squash your commits.

Done!

MatheusRich:
@rafaelfranca thoughts on this one?

Issue #54591, opened 2025-02-22
Inconsistent finder method behaviour without primary key and implicit_order_column


joshuay03:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails", branch: "main"

  gem "sqlite3"
end

require "active_record/railtie"
require "minitest/autorun"

# This connection will do for database-independent bug reports.
ENV["DATABASE_URL"] = "sqlite3::memory:"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.eager_load = false
  config.logger = Logger.new($stdout)
  config.secret_key_base = "secret_key_base"

  config.active_record.encryption.primary_key = "primary_key"
  config.active_record.encryption.deterministic_key = "deterministic_key"
  config.active_record.encryption.key_derivation_salt = "key_derivation_salt"
end
Rails.application.initialize!

ActiveRecord::Schema.define do
  create_table :posts, force: true, id: false do |t|
    t.integer :order_column, null: false
  end
end

class Post < ActiveRecord::Base
  # self.implicit_order_column = "order_column"
end

class BugTest < ActiveSupport::TestCase
  def test_first_and_last_without_primary_key_and_order_column
    Post.create! order_column: 3
    Post.create! order_column: 1
    Post.create! order_column: 2

    assert_nothing_raised { Post.first } # ✅
    assert_nothing_raised { Post.last } # ❌
  end
end

```

### Expected behavior

Both `first` and `last` should either return a result without ordering or raise an error.

### Actual behavior

`first` returns a result without ordering, `last` raises:

```
ActiveRecord::IrreversibleOrderError: Relation has no current order and table has no primary key to be used as default order
```

### System configuration

**Rails version**: Edge

**Ruby version**: 3.4.2

joshuay03:
This happens because `last` relies on `reverse_order!` [here](https://github.com/rails/rails/blob/d57c5073a63851f887668e8384f963b99d4e0d62/activerecord/lib/active_record/relation/finder_methods.rb#L206). Let me know what the expected behaviour should be and I'd be happy to send a PR.

byroot:
I'd say the correct behavior would be for `first` to raise if there isn't an implicit order.

Because right now it relies on whatever order the database returns which is often somewhat random and is a recipe for bugs.

[...]

PR #54607, opened 2025-02-24
Ensure `#reverse_order` respects `implicit_order_column`

joshuay03:
### Motivation / Background

Addresses https://github.com/rails/rails/issues/54591#issuecomment-2677869549

### Detail

Updates `ActiveRecord::QueryMethods#reverse_sql_order` ~to utilize `_order_columns` similar to `ActiveRecord::FinderMethods#ordered_relation`~ to account for `implicit_order_column`. This ensures that:
1. ~`#reverse_sql_order` is now consistent with `#ordered_relation`.~
2. `#reverse_order` now respects `implicit_order_column` ~as well as `query_constraints`~, and not just `primary_key`.
3. ~We can now simply use `#reverse_order` in `#last` instead of having to order and then reverse the order (this has comprehensive test coverage)~

Edits were made after https://github.com/rails/rails/pull/54607#issuecomment-2678291542

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/finder_methods.rb
+++ b/activerecord/lib/active_record/relation/finder_methods.rb
@@ -638,7 +638,7 @@ module ActiveRecord
       end
 
       def ordered_relation
-        if order_values.empty? && (model.implicit_order_column || !model.query_constraints_list.nil? || primary_key)
+        if order_values.empty? && !_order_columns.empty?
           order(_order_columns.map { |column| table[column].asc })
         else
           self
--- a/activerecord/test/cases/relations_test.rb
+++ b/activerecord/test/cases/relations_test.rb
@@ -370,6 +370,19 @@ class RelationTest < ActiveRecord::TestCase
     end
   end
 
+  def test_default_reverse_order_on_table_without_primary_key_with_implicit_order_column
+    ordered_edge = Class.new(Edge) do
+      self.implicit_order_column = "source_id"
+    end
+
+    ordered_edge.delete_all
+
+    edge_1 = ordered_edge.create!(source_id: 1, sink_id: 2)
+    edge_2 = ordered_edge.create!(source_id: 2, sink_id: 3)
+    assert_equal edge_1.source_id, ordered_edge.all.first.source_id
+    assert_equal edge_2.source_id, ordered_edge.all.reverse_order.first.source_id
+  end
+
   def test_order_with_hash_and_symbol_generates_the_same_sql
     assert_equal Topic.order(:id).to_sql, Topic.order(id: :asc).to_sql
   end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Respect `implicit_order_column` in `ActiveRecord::Relation#reverse_order`.
+
+    *Joshua Young*
+
 *   Add column types to `ActiveRecord::Result` for SQLite3.
 
     *Andrew Kane*
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -2009,9 +2009,12 @@ module ActiveRecord
 
       def reverse_sql_order(order_query)
         if order_query.empty?
-          return [table[primary_key].desc] if primary_key
+          if !_reverse_order_columns.empty?
+            return _reverse_order_columns.map { |column| table[column].desc }
+          end
+
           raise IrreversibleOrderError,
-            "Relation has no current order and table has no primary key to be used as default order"
+            "Relation has no current order and table has no order columns to be used as default order"
         end
 
         order_query.flat_map do |o|
@@ -2034,6 +2037,13 @@ module ActiveRecord
             o
           end
         end
+      end
+
+      def _reverse_order_columns
+        roc = []
+        roc << model.implicit_order_column if model.implicit_order_column
+        roc << model.primary_key if model.primary_key
+        roc.flatten.uniq.compact
       end
 
       def does_not_support_reverse?(order)


activerecord/test/cases/finder_test.rb
```diff
@@ -1883,6 +1883,15 @@ def test_finder_with_offset_string
     end
   end
 
+  test "#reverse_order for a model with composite query constraints" do
+    quoted_type = Regexp.escape(quote_table_name("clothing_items.clothing_type"))
+    quoted_color = Regexp.escape(quote_table_name("clothing_items.color"))
+
+    assert_queries_match(/ORDER BY #{quoted_type} DESC, #{quoted_color} DESC LIMIT/i) do
```
joshuay03, line 1890: Without patch:
Screenshot 2025-02-24 at 8 15 16 pm

The expected query is the same as `ClothingItem.last` tested above.

activerecord/test/cases/relations_test.rb
```diff
@@ -370,6 +370,18 @@ def test_default_reverse_order_on_table_without_primary_key
...
   end
 
+  def test_default_reverse_order_on_table_without_primary_key_with_implicit_order_column
+    implicit_order_column_before = Edge.implicit_order_column
+    Edge.implicit_order_column = "source_id"
+
+    edge_1 = Edge.create!(source_id: 1, sink_id: 2)
+    edge_2 = Edge.create!(source_id: 2, sink_id: 3)
+    assert_equal edge_1.source_id, Edge.all.first.source_id
+    assert_equal edge_2.source_id, Edge.all.reverse_order.first.source_id
```
joshuay03, line 380: Without patch:
Screenshot 2025-02-24 at 8 16 47 pm

activerecord/test/cases/finder_test.rb
byroot: Rather than to assert on the exact query, wouldn't it be preferable to assert on the records order?

e.g.

```ruby
assert_equal ClothingItem.pluck(:id).reverse, ClothingItem.reverse_order.pluck(:id)
```

joshuay03: Can do. Actually this test isn't appropriate here cause `#reverse_order` is not a finder method... Will address.

activerecord/test/cases/relations_test.rb
```diff
@@ -370,6 +370,20 @@ def test_default_reverse_order_on_table_without_primary_key
...
+
+    Edge.delete_all
+
+    edge_1 = Edge.create!(source_id: 1, sink_id: 2)
+    edge_2 = Edge.create!(source_id: 2, sink_id: 3)
+    assert_equal edge_1.source_id, Edge.all.first.source_id
+    assert_equal edge_2.source_id, Edge.all.reverse_order.first.source_id
+  ensure
+    Edge.implicit_order_column = implicit_order_column_before
+  end
```
byroot, line 385: Rather than to mutate the model, you can do an anonymous subclass of it:

```suggestion
  def test_default_reverse_order_on_table_without_primary_key_with_implicit_order_column
    ordered_edge = Class.new(Edge) do
      self.implicit_order_column = "source_id"
    end

    ordered_edge.delete_all

    edge_1 = ordered_edge.create!(source_id: 1, sink_id: 2)
    edge_2 = ordered_edge.create!(source_id: 2, sink_id: 3)
    assert_equal edge_1.source_id, ordered_edge.all.first.source_id
    assert_equal edge_2.source_id, ordered_edge.all.reverse_order.first.source_id
  end
```

activerecord/lib/active_record/relation/finder_methods.rb
```diff
@@ -638,7 +637,7 @@ def find_last(limit)
       end
 
       def ordered_relation
-        if order_values.empty? && (model.implicit_order_column || !model.query_constraints_list.nil? || primary_key)
+        if order_values.empty? && _order_columns.any?
```
byroot, line 640: personal pet peeve. `any?` isn't the inverse of `empty?`, it ignores falsy members.

Not a huge deal here, but IMO `!_order_columns.empty?` is clearer.

activerecord/test/cases/relations_test.rb
```diff
@@ -370,6 +371,29 @@ def test_default_reverse_order_on_table_without_primary_key
...
+
+  def test_default_reverse_order_on_table_with_query_constraints
+    ClothingItem.delete_all
+
+    ClothingItem.create!(clothing_type: "shirt")
+    ClothingItem.create!(clothing_type: "pant")
+    ClothingItem.create!(clothing_type: "tie")
+
+    assert_equal ["tie", "shirt", "pant"], ClothingItem.all.reverse_order.pluck(:clothing_type)
+  end
```
joshuay03, line 395: Without patch:
Screenshot 2025-02-24 at 9 35 56 pm

activerecord/test/cases/finder_test.rb
joshuay03: Addressed 👍🏽

activerecord/test/cases/relations_test.rb
joshuay03: Addressed 👍🏽

activerecord/lib/active_record/relation/finder_methods.rb
joshuay03: I always forget about that caveat. Addressed 👍🏽

activerecord/test/cases/relations_test.rb
```diff
@@ -370,6 +371,29 @@ def test_default_reverse_order_on_table_without_primary_key
...
+
+  def test_default_reverse_order_on_table_with_query_constraints
+    ClothingItem.delete_all
+
+    ClothingItem.create!(clothing_type: "shirt")
+    ClothingItem.create!(clothing_type: "pant")
+    ClothingItem.create!(clothing_type: "tie")
+
+    assert_equal ["tie", "shirt", "pant"], ClothingItem.all.reverse_order.pluck(:clothing_type)
```
byroot, line 394:
```suggestion
    assert_equal ClothingItem.all.pluck(:clothing_type).reverse, ClothingItem.all.reverse_order.pluck(:clothing_type)
```
So that it doesn't break if someone adds a fixture.

joshuay03: Addressed but I also sorted before reversing, cause the expectation query won't be ordered.

joshuay03:
Hmm looks like I've misunderstood how `query_constraints` are meant to be used... It seems they're only meant for `#first` and `#last`, so `#reverse_order` shouldn't be respecting them by default: https://github.com/rails/rails/blob/7845576df0fb68ff38f9ca754fca457a8e6758a3/activerecord/lib/active_record/persistence.rb#L183

joshuay03:
Updated to only account for `implicit_order_column`.

PR #54608, opened 2025-02-24
[Fix #54591] Raise an error in order dependent finder methods when the model has no order columns

### LINKED ISSUE #54591: Inconsistent finder method behaviour without primary key and implicit_order_column

joshuay03:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/54591
Follow-up to https://github.com/rails/rails/pull/54607

### Detail

1. Adds a new error `ActiveRecord::MissingRequiredOrderError` which gets raised in `ActiveRecord::FinderMethods#ordered_relation` (used by `#last` and and all `find_nth*` methods) when there are no `order_values` and no order columns to be used for the default order.
2. Updates a similar case where `ActiveRecord::IrreversibleOrderError` is raised in `ActiveRecord::QueryMethods#reverse_sql_order` to have a message consistent with this new error.

cc @byroot

### Checklist

### DIFF
--- a/activerecord/lib/active_record/errors.rb
+++ b/activerecord/lib/active_record/errors.rb
@@ -552,6 +552,11 @@ module ActiveRecord
   class Deadlocked < TransactionRollbackError
   end
 
+  # MissingRequiredOrderError is raised when a relation requires ordering but
+  # lacks any +order+ values in scope or any model order columns to use.
+  class MissingRequiredOrderError < ActiveRecordError
+  end
+
   # IrreversibleOrderError is raised when a relation's order is too complex for
   # +reverse_order+ to automatically reverse.
   class IrreversibleOrderError < ActiveRecordError
--- a/activerecord/lib/active_record/relation/finder_methods.rb
+++ b/activerecord/lib/active_record/relation/finder_methods.rb
@@ -638,8 +638,27 @@ module ActiveRecord
       end
 
       def ordered_relation
-        if order_values.empty? && !_order_columns.empty?
-          order(_order_columns.map { |column| table[column].asc })
+        if order_values.empty?
+          if !_order_columns.empty?
+            return order(_order_columns.map { |column| table[column].asc })
+          end
+
+          if ActiveRecord.raise_on_missing_required_finder_order_columns
+            raise MissingRequiredOrderError, <<~MSG.squish
+              Relation has no order values, and #{model} has no order columns to use as a default.
+              Set at least one of `implicit_order_column`, `query_constraints` or `primary_key` on
+              the model when no `order `is specified on the relation.
+            MSG
+          else
+            ActiveRecord.deprecator.warn(<<~MSG)
+              Calling order dependent finder methods (e.g. `#first`, `#second`) without `order` values on the relation,
+              and on a model (#{model}) that does not have any order columns (`implicit_order_column`, `query_constraints`,
+              or `primary_key`) to fall back on is deprecated and will raise `ActiveRecord::MissingRequiredOrderError`
+              in Rails 8.2.
+            MSG
+
+            self
+          end
         else
           self
         end
--- a/activerecord/test/cases/finder_test.rb
+++ b/activerecord/test/cases/finder_test.rb
@@ -26,6 +26,7 @@ require "models/subscriber"
 require "models/non_primary_key"
 require "models/clothing_item"
 require "models/cpk"
+require "models/edge"
 require "support/stubs/strong_parameters"
 require "support/async_helper"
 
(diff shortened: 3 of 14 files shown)


activerecord/lib/active_record/errors.rb
```diff
@@ -552,6 +552,11 @@ class SerializationFailure < TransactionRollbackError
...
   end
 
+  # OrderError is raised when a relation needs to be ordered but does not have any
+  # order columns to use.
+  class OrderError < ActiveRecordError
+  end
+
   # IrreversibleOrderError is raised when a relation's order is too complex for
   # +reverse_order+ to automatically reverse.
   class IrreversibleOrderError < ActiveRecordError
```
skipkayhil, line 562: I like how `IrreversibleOrderError` is self-documenting; it has a sentence describing it but you can generally tell what it does without having to look it up. Should `OrderError` be something like `MissingRequiredOrderError` so that its more self-documenting?

activerecord/lib/active_record/relation/finder_methods.rb
```diff
@@ -638,8 +638,15 @@ def find_last(limit)
...
-        if order_values.empty? && !_order_columns.empty?
-          order(_order_columns.map { |column| table[column].asc })
+        if order_values.empty?
+          if !_order_columns.empty?
+            return order(_order_columns.map { |column| table[column].asc })
+          end
+
+          raise OrderError, <<~MSG.squish
+            Relation has no current order, and #{model} has no order columns to use as a default.
+            At least one of `implicit_order_column`, `query_constraints`, or `primary_key` must be set on the model.
```
skipkayhil, line 648: I think `must be set` isn't quite correct, because you could also specify an `order` on the relation? (same concern with the change to the other error)

joshuay03: > because you could also specify an order on the relation?

Yes that's what this first bit is alluding to:

> Relation has no current order,

But I can make it more clear by specifying that in the suggestion as well 👍🏽

activerecord/lib/active_record/errors.rb
joshuay03: I like that, will action 👍🏽

joshuay03: Done!

activerecord/lib/active_record/relation/finder_methods.rb
joshuay03: Done, let me know if it's clearer now.

activerecord/test/cases/finder_test.rb
```diff
@@ -1055,10 +1056,93 @@ def test_last_on_loaded_relation_should_not_use_sql
...
+    assert_nil Edge.implicit_order_column
+    assert_nil Edge.query_constraints_list
+    error = assert_raises(ActiveRecord::MissingRequiredOrderError) do
+      Edge.all.last
+    end
+    assert_equal <<~MSG.squish, error.message
+      Relation has no current order, and Edge has no order columns to use as a default.
+      Set at least one of `implicit_order_column`, `query_constraints` or `primary_key` on
+      the model when no `order `is specified on the relation.
+    MSG
```
byroot, line 1111: I'm not sure it's worth matching the whole message. Something like `assert_match "Relation has no current order"` is enough and less things to change if we decide to reword the error message.

byroot (approved):
I like this a lot. I'll wait a couple days to see if someone opposes though.

byroot:
I wonder if this should be considered breaking enough to warrant a framework default and a deprecation.

Which is debatable because this is solving a bug, but prior you'd get a somewhat unpredictable behavior but "working" behavior, now you'd get a straight error.

activerecord/test/cases/finder_test.rb
joshuay03: Addressed.

joshuay03:
> > I wonder if this should be considered breaking enough to warrant a framework default and a deprecation.
> > Which is debatable because this is solving a bug, but prior you'd get a somewhat unpredictable behavior but "working" behavior, now you'd get a straight error.
> 
> Hmm, yeah, that's a good point, I can see the argument for a deprecation notice. I will action soon 👍🏽

Action-ed! Hopefully I've done all the things right...

byroot:
Uh, just realized this fell through the cracks, I'll try to rebase it.

activerecord/lib/active_record/relation/finder_methods.rb
```diff
@@ -638,8 +638,27 @@ def find_last(limit)
...
-        if order_values.empty? && !_order_columns.empty?
-          order(_order_columns.map { |column| table[column].asc })
+        if order_values.empty?
+          if !_order_columns.empty?
+            return order(_order_columns.map { |column| table[column].asc })
+          end
+
+          if ActiveRecord.raise_on_missing_required_finder_order_columns
+            raise MissingRequiredOrderError, <<~MSG.squish
+              Relation has no order values, and #{model} has no order columns to use as a default.
```
byroot, line 648:
```suggestion
              Relation has no order values, and #{model.name} has no order columns to use as a default.
```

`model` is a AR class here right? I think we should just use the name, because otherwise `AR::Base.inspect` can be super long and hard to read.

byroot: Nevermind, `AR::Base.to_s` no longer include columns.

joshuay03:
Thank you!

PR #54616, opened 2025-02-25
[ci skip] docs: Fix form_with model binding example in guides

porporpor020202:
### Motivation / Background

The current form_helpers guide shows an incorrect example where a form for an existing record displays "Create Book" as the submit button text. This is misleading because Rails automatically sets the submit button text to "Update [Model]" for existing records.

In the example, `@book = Book.find(42)` is used to fetch an existing record, so the submit button should show "Update Book" instead of "Create Book".

### Detail

This Pull Request updates the form_helpers guide to show the correct submit button text ("Update Book") in the HTML output example when working with an existing record. This change makes the documentation more accurate and helps prevent confusion for developers learning Rails form helpers.

### Additional information

When using `form_with` with a model, Rails determines the submit button text based on whether the record is persisted:
- For new records: "Create [Model]"
- For existing records: "Update [Model]"

The example in the guide uses `Book.find(42)`, which returns a persisted record, so the submit button text should be "Update Book".

### Checklist

* [x] No tests needed as this is a documentation-only change.
* [x] No CHANGELOG update needed as this is a documentation correction.

### DIFF
--- a/guides/source/form_helpers.md
+++ b/guides/source/form_helpers.md
@@ -282,7 +282,7 @@ It will generate this HTML:
     <label for="book_author">Author</label>
     <input type="text" name="book[author]" id="book_author">
   </div>
-  <input type="submit" name="commit" value="Create Book" data-disable-with="Create Book">
+  <input type="submit" name="commit" value="Update Book" data-disable-with="Update Book">
 </form>
 ```
 
@@ -290,7 +290,7 @@ Some important things to notice when using `form_with` with a model object:
 
 * The form `action` is automatically filled with an appropriate value, `action="/books"`. If you were updating a book, it would be `action="/books/42"`.
 * The form field names are scoped with `book[...]`. This means that `params[:book]` will be a hash containing all these field's values. You can read more about the significance of input names in chapter [Form Input Naming Conventions and Params Hash](#form-input-naming-conventions-and-params-hash) of this guide.
-* The submit button is automatically given an appropriate text value, "Create Book" in this case.
+* The submit button is automatically given an appropriate text value, "Update Book" in this case.
 
 TIP: Typically your form inputs will mirror model attributes. However, they don't have to. If there is other information you need you can include a field in your form and access it via `params[:book][:my_non_attribute_input]`.
 
--- a/guides/source/form_helpers.md
+++ b/guides/source/form_helpers.md
@@ -288,7 +288,7 @@ It will generate this HTML:
 
 Some important things to notice when using `form_with` with a model object:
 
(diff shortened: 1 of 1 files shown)


guides/source/form_helpers.md
```diff
@@ -282,15 +282,15 @@ It will generate this HTML:
...
     <input type="text" name="book[author]" id="book_author">
   </div>
-  <input type="submit" name="commit" value="Create Book" data-disable-with="Create Book">
+  <input type="submit" name="commit" value="Update Book" data-disable-with="Update Book">
 </form>
 ```
 
 Some important things to notice when using `form_with` with a model object:
 
 * The form `action` is automatically filled with an appropriate value, `action="/books"`. If you were updating a book, it would be `action="/books/42"`.
```
skipkayhil, line 291: Looks like this line should be updated as well? (And the `action=` on the form itself)

porporpor020202: Should we swap the examples? From:

> The form `action` is automatically filled with an appropriate value, `action="/books"`. If you were updating a book, it would be `action="/books/42"`.

To:

> The form `action` is automatically filled with an appropriate value, `action="/books/42"`. If you were creating a book, it would be `action="/books"`.

porporpor020202: done!

skipkayhil: > (And the action= on the form itself)

Does the form still match the updated documentation?

porporpor020202: "I'm sorry. My code understanding is still limited, so it took me a while to understand what you meant. May I ask for your help? I initially thought you were asking if I had also fixed the actual Rails code, but that's not it. I don't know how to fix the actual code. I only fixed the documentation.

Screenshot 2025-02-27 at 4 34 51 PM

I'm pointing out that there's something strange in the part shown in the image. Clearly, @book is found using Book.find(42), but the HTML that's actually generated says 'Create Book'. This seems inconsistent because when we're finding an existing book with ID 42, the button should say 'Update Book' instead of 'Create Book'."

<URL> - creating form with model object
https://guides.rubyonrails.org/form_helpers.html

porporpor020202: And I'll squash my commits!

porporpor020202: Oh, I see. Thank you for your answer! I didn't catch the other inconsistency, so thanks for paying attention to that. I also think that changing @book = Book.find(42) to @book = Book.new looks good, but what do you think? Would it be okay if I make that change, squash the commits, and push it?

PR #54615, opened 2025-02-25
Enable `Minitest/AssertNil` cop to address `DEPRECATED: Use assert_nil if expecting nil` warning

yahonda:
### Motivation / Background

This commit addresses the `DEPRECATED: Use assert_nil if expecting nil` warning at https://buildkite.com/rails/rails/builds/116902#01953bd2-7999-468a-bcd6-5e9b68964fa6/1270-1276

### Detail

```ruby
$ cd railties
$ RAILS_STRICT_WARNINGS=1 bin/test test/application/configuration_test.rb:4031
Run options: --seed 1261

DEPRECATED: Use assert_nil if expecting nil from /home/yahonda/src/github.com/rails/rails/railties/test/application/configuration_test.rb:4031. This will fail in Minitest 6.
E

Error:
ApplicationTests::ConfigurationTest#test_ActiveStorage.variant_processor_uses_mini_magick_without_Rails_7_defaults:
ActiveSupport::RaiseWarnings::WarningError: DEPRECATED: Use assert_nil if expecting nil from /home/yahonda/src/github.com/rails/rails/railties/test/application/configuration_test.rb:4031. This will fail in Minitest 6.

    /home/yahonda/src/github.com/rails/rails/activesupport/lib/active_support/testing/strict_warnings.rb:38:in 'ActiveSupport::RaiseWarnings#warn'

bin/test test/application/configuration_test.rb:4026

Finished in 1.568723s, 0.6375 runs/s, 0.6375 assertions/s.
1 runs, 1 assertions, 0 failures, 1 errors, 0 skips
$
```

### Additional information
Refer to https://github.com/minitest/minitest/blob/master/History.rdoc#5230--2024-05-15
https://github.com/minitest/minitest/commit/f0f17b99

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54617, opened 2025-02-25
Move StrictWarnings to the tools directory

byroot:
Followup: https://github.com/rails/rails/pull/46189

Not sure why I put it in Active Support lib directory, that makes little sense, as it's shipped with the gem but isn't something usable outside of rails itself.


zzak:
Shouldn't CI fail if #54615 is not merged yet?

byroot:
Yes and no. Railties are a bit annoying because they don't have one common shared file, so https://github.com/rails/rails/pull/54615 shows that some railties tests are missing strict warnings.

byroot:
> Worst case we could inject a global require here?

Something like that yes.

PR #54620, opened 2025-02-25
Ensure all railties tests require strict_warnings

zzak:
Re: #54617

### DIFF
--- a/railties/Rakefile
+++ b/railties/Rakefile
@@ -116,6 +116,7 @@ namespace :test do
           ARGV.clear.concat test_options
           Rake.application = nil
 
+          require "../tools/strict_warnings"
           load file
         }
       else
--- a/railties/test/abstract_unit.rb
+++ b/railties/test/abstract_unit.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require_relative "../../tools/strict_warnings"
-
 ENV["RAILS_ENV"] ||= "test"
 
 require "stringio"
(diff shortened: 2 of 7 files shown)


zzak:
Hmm looks like it still doesn't fail the test:
https://buildkite.com/rails/rails/builds/116917#01953d1f-27a7-4d40-ba82-1e5f2fc22557

zzak:
Ok it should fail with e179f6d, it's kind of a hack but a neat way I think for us to identify warnings that occur in our tests that don't have the full path in the message. Since it's extremely unlikely that a path in the message from a gem will match a test file in our current directory. Not impossible though, I guess.

yahonda:
In the meantime I’m keeping #54615 open, not merged, to make this pull request easier to test.

byroot:
Look like it's now failing as expected.

zzak:
I guess we don't need e179f6d if the minitest PR is fixed :thinking: I guess it's useful if other gems put the caller inside the message -- which hasn't happened outside this case that I am aware of so probably don't need it.

byroot:
I'm fine with merging like that, just need to integrate @yahonda 's fix.

yahonda:
I am also fine with keeping https://github.com/rails/rails/commit/e179f6d637aa58529ffededeea616d0b537214d6 . because not all of `warn` usage is written as https://github.com/minitest/minitest/pull/1028 .

yahonda:
I'm merging https://github.com/rails/rails/pull/54615 because we have validated this pull request works fine.

PR #54625, opened 2025-02-25
Add column types to `ActiveRecord::Result` for the mysql2 adapter

ankane:
### Motivation / Background

Applies #54594 to the mysql2 adapter. This is useful for casting boolean columns.

I did not see a way to get the SQL column types with Trilogy.

### Checklist


ankane:
It looks like the `mysql2` gem doesn't include precision in the `field_types` method for `datetime` fields. Will follow up there.

PR #54627, opened 2025-02-26
Only create one config per AV::Base

skipkayhil:
Currently, AV::Base's config is always initialized to a blank InheritableOptions, and then later config is overridden with the given controller's config if it responds to config (which it will most of the time since AbstractController::Base has a config).

This commit moves the blank config instantiation into the ControllerHelper so that only a single InheritableOptions is ever instantiated for an AV::Base.

### DIFF
--- a/actionview/lib/action_view/base.rb
+++ b/actionview/lib/action_view/base.rb
@@ -242,8 +242,6 @@ module ActionView # :nodoc:
     # :startdoc:
 
     def initialize(lookup_context, assigns, controller) # :nodoc:
-      @_config = ActiveSupport::InheritableOptions.new
-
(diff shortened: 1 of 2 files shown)


skipkayhil:
It seems like `config` is unused internally, I've only found two tests that fail with it removed:

https://github.com/rails/rails/blob/1a820f3cc489a49cb146a9ad707522534354eaa3/actionview/test/template/test_case_test.rb#L68-L70

https://github.com/rails/rails/blob/1a820f3cc489a49cb146a9ad707522534354eaa3/actionpack/test/controller/helper_test.rb#L257-L261

Maybe it could just be deprecated (or removed since its `attr_internal`)?

PR #54628, opened 2025-02-26
Modify the transactional callbacks order to match the SQL sequence: 

### LINKED ISSUE #54623: Bug Report: `after_commit` callbacks Incorrectly ordered when using `accepts_nested_attributes_for`
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails"

  gem "sqlite3"
end

require "active_support/all"
require "minitest/autorun"

  config.eager_load = false

end

Rails.application.initialize!

  end

    t.integer :parent_id
  end

  create_table :events, force: true do |t|
    t.string :name
    t.string :record_type
    t.integer :record_id
    t.timestamps
  end
end

  has_many :children
  has_many :events, as: :record

  after_commit :log_event, on: :create

  def log_event
    ActiveSupport::Notifications.instrument "parent.created", { record_type: self.class.name, record_id: self.id }
  end
end

class Child < ActiveRecord::Base
  belongs_to :parent
  has_many :events, as: :record

  accepts_nested_attributes_for :parent

  after_commit :log_event, on: :create

  def log_event
    ActiveSupport::Notifications.instrument "child.created", { record_type: self.class.name, record_id: self.id }
  end
end

class Event < ActiveRecord::Base
  belongs_to :record, polymorphic: true
end

ActiveSupport::Notifications.subscribe(/(child|parent)\.created/) do |event|
  Event.create!(name: event.name, **event.payload)
end

  def test_when_objects_are_created_independently
    parent = Parent.create!
    assert_equal 1, parent.events.count

    child = Child.create!(parent: parent)
    assert_equal 1, child.events.count

    parent_event = parent.events.first
    child_event = child.events.first


    assert_equal parent_event.name, 'parent.created'
    assert_equal child_event.name, 'child.created'
    assert parent_event.created_at < child_event.created_at
  end

  def test_when_objects_are_created_using_nested_attributes

    child = Child.create!(parent_attributes: {})
    parent = child.parent

    assert_equal 1, parent.events.count
    assert_equal 1, child.events.count

    parent_event = parent.events.first
    child_event = child.events.first

    assert_equal parent_event.name, 'parent.created'
    assert_equal child_event.name, 'child.created'
    assert parent_event.created_at < child_event.created_at # will fail
  end
end
```

### Expected behavior

When using `accepts_nested_attributes_for` to create parent and child records, the `after_commit` callback on the parent record should fire before the callback on the child record. This ensures the correct order of associated `Event` records, where the `"parent.created"` event is consistently sent before the `"child.created"` event.
This behavior is especially important when the `Event` class is used to send messages to an external event store. If the events are out of order, consumers of the event store may process child-related events (e.g., `"child.created"`) before the parent-related events (e.g., `"parent.created"`), leading to inconsistencies or errors.

### Actual behavior

When using `accepts_nested_attributes_for`, the `after_commit` callback for the child record executes before the parent record’s callback. As a result:
- The `"child.created"` event is logged before the `"parent.created"` event in the `Event` table.
- For systems relying on event ordering (e.g., an external event store), this causes issues because events might be consumed out of order, resulting in invalid or incomplete data processing.

For example, the `"parent.created"` event is expected to arrive before the `"child.created"` event in the event store. However, due to this behavior, the child’s event will be processed first.

### System configuration

**Rails version**: main (latest branch) (but I've seen this issue in older versions like 7.2)
**Ruby version**: 3.3.4

Edouard-chin:
### Motivation / Background

Fix https://github.com/rails/rails/issues/54623

### Detail

- ### Problem

  Previously, transactional callbacks such as `before_commit`, `after_commit` and their `rollbacks`
  counterpart, were running in the same order as when persistence methods such as `save` gets called.

  Once opted-in to the new behaviour, transaction callbacks will run in the same order as the SQL
  statements execution sequence.

  ```ruby
    after_commit -> { puts "Hello" }
  end

  class Comment < ApplicationRecord
    belongs_to :post, autosave: true

    after_commit -> { puts "World" }
  end

  Comment.create

  # The resulting SQL would look like: (Note that this doesn't change)
  # BEGIN TRANSACTION
  #   INSERT_INTO posts ...
  #   INSERT_INTO comments ...
  # COMMIT

  # Previously, the callback order was [Comment, Post]. The output would be "World Hello".
  # The new behaviour matches the SQL execution statements, and the callback order is [Post, Comment],
  # the output is "Hello World".
  ```

  This new behaviour is disabled by default. To opt-in, set the
  `config.active_record.run_transaction_callbacks_in_same_sequence = true`
  in your application's configuration.

  ### Solution

  Set a timestamp on the record, higher up the stack, just before we
  run the SQL statement. Before running the transactional callbacks,
  order the records by this timestamp.

### Additional details

Although I believe that this behaviour is the correct one, I added this new behaviour behind a flag because I was worried this could break existing apps.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,41 @@
+*   Modify the transactional callbacks calling order to match the same execution sequence as the SQL statements.
+
+    Previously, transactional callbacks such as `before_commit`, `after_commit` and their `rollbacks`
+    counterpart, were running in the same order as persistence methods such as `save` gets called.
+
+    Once opted-in to the new behaviour, transaction callbacks will run in the same order as the SQL
+    statements execution sequence.
+
+    ```ruby
+    class Post < ApplicationRecord
+      after_commit -> { puts "Hello" }
+    end
+
+    class Comment < ApplicationRecord
+      belongs_to :post, autosave: true
+
+      after_commit -> { puts "World" }
+    end
+
+    Comment.create
+
+    # The resulting SQL would look like: (Note that this doesn't change)
+    # BEGIN TRANSACTION
+    #   INSERT_INTO posts ...
+    #   INSERT_INTO comments ...
+    # COMMIT
+
+    # Previously, the callback order was [Comment, Post]. The output would be "World Hello".
+    # The new behaviour matches the SQL execution statements, and the callback order is [Post, Comment],
+    # the output is "Hello World".
+    ```
+
+    This new behaviour is disabled by default. To opt-in, set the
+    `config.active_record.run_transaction_callbacks_in_same_sequence = true` in your application's
+    configuration.
+
+    *Edouard Chin*
+
 *   Support disabling indexes for MySQL v8.0.0+ and MariaDB v10.6.0+
 
     MySQL 8.0.0 added an option to disable indexes from being used by the query
--- a/activerecord/lib/active_record.rb
+++ b/activerecord/lib/active_record.rb
@@ -354,6 +354,9 @@ module ActiveRecord
   singleton_class.attr_accessor :run_after_transaction_callbacks_in_order_defined
   self.run_after_transaction_callbacks_in_order_defined = false
 
+  singleton_class.attr_accessor :run_transaction_callbacks_in_same_sequence
+  self.run_transaction_callbacks_in_same_sequence = false
+
   singleton_class.attr_accessor :application_record_class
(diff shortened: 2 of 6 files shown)


dlupu:
@Edouard-chin thank you very much for suggesting a fix so quickly !

Edouard-chin:

PR #54631, opened 2025-02-26
Improve devcontainer volume mount to only include the app folder

elalemanyo:
**Motivation / Background**

Currently, the `compose.yaml.tt` template mounts the entire parent directory (../..) into the devcontainer workspace. This can lead to unnecessary files being included in the container, which is not ideal.

This PR updates the volume mount path to only include the application folder `../../<%= options[:app_name] %>`. This ensures a cleaner and more isolated workspace inside the devcontainer.

**Detail**

- Updated `railties/lib/rails/generators/rails/devcontainer/templates/devcontainer/compose.yaml.tt`
- Changed the volume mount from `../..:/workspaces:cached` to `../../<%= options[:app_name] %>:/workspaces/<%= options[:app_name] %>:cached`
- This makes sure only the application directory is mounted instead of the entire parent directory

**Additional Information**

This change aligns with best practices for mounting project-specific directories in devcontainers. It prevents mounting unnecessary files and ensures a more focused development environment.

**Checklist**

- This Pull Request is related to one change.
- Commit message has a detailed description of what changed and why.
- Tests have been updated to reflect the volume mount change.
- No behavior change requiring a CHANGELOG update.

### DIFF
--- a/railties/lib/rails/generators/rails/devcontainer/templates/devcontainer/compose.yaml.tt
+++ b/railties/lib/rails/generators/rails/devcontainer/templates/devcontainer/compose.yaml.tt
@@ -7,7 +7,7 @@ services:
       dockerfile: .devcontainer/Dockerfile
 
     volumes:
-    - ../..:/workspaces:cached
+    - ../../<%= options[:app_name] %>:/workspaces/<%= options[:app_name] %>:cached
 
     # Overrides default command so things don't shut down after the process ends.
     command: sleep infinity
--- a/railties/test/generators/app_generator_test.rb
+++ b/railties/test/generators/app_generator_test.rb
@@ -1400,7 +1400,7 @@ class AppGeneratorTest < Rails::Generators::TestCase
           "context" => "..",
           "dockerfile" => ".devcontainer/Dockerfile"
         },
-        "volumes" => ["../..:/workspaces:cached"],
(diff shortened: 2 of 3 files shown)


rafaelfranca:
> This change aligns with best practices for mounting project-specific directories in devcontainers.

Where are those best practices defined?

The only place I could think about is in the official devcontainer templates, and I don't see this being a problem there https://github.com/search?q=repo%3Adevcontainers%2Ftemplates%20-%20..%2F..%3A%2Fworkspaces%3Acached&type=code

rafaelfranca:
Anyway, this is a better default.

GarrisonD:
Don't know why, but it broke my dev container after upgrading Ruby on Rails to `8.1.1`

Had to revert your change locally:

```diff
    volumes:
-      - ../../my-project:/workspaces/my-project:cached
+      - ../..:/workspaces:cached
```

@elalemanyo How did you test it?

elalemanyo:
@GarrisonD 😬 I tried it with all the projects I work on. Then, for the pull request, I did it with a new version of a rails app.
I think the case where this can cause problems is when the folder has a different name than the project. In my case, I experienced this after forking a repository and changing its name. As a result, the project name does not match the folder name. Could this be your case? I apologize if I have caused any trouble. Please let me know how I can help.

PR #54632, opened 2025-02-26
[ci skip] Clarify nested transaction and exception behavior

dersam:
### Motivation / Background

The documentation on [nested transactions and rollbacks](https://api.rubyonrails.org/v8.0/classes/ActiveRecord/Transactions/ClassMethods.html) can be confusing. The situation described in the documentation is specific to `ActiveRecord::Rollback` - however I've encountered many developers that thought this was the behavior for any raised exception in a nested transaction.

### Detail

I've updated the docs to call out that this applies only to ActiveRecord::Rollback. Any other exception will be passed on, not hidden.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/transactions.rb
+++ b/activerecord/lib/active_record/transactions.rb
@@ -136,7 +136,7 @@ module ActiveRecord
     #
     # #transaction calls can be nested. By default, this makes all database
     # statements in the nested transaction block become part of the parent
-    # transaction. For example, the following behavior may be surprising:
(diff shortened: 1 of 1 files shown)


Edouard-chin:
Thanks for opening the Pull Request. This is already documented in the same file, just a few lines before.

https://github.com/rails/rails/blob/d8f68f9d558be9420d9432dcfcb5b80a2b3cb161/activerecord/lib/active_record/transactions.rb#L98-L104

I don't think it's worth repeating tnis a few lines after.

PR #54635, opened 2025-02-26
Allow params with nested brackets to be parsed similar to how it worked on Rails 7

### LINKED ISSUE #54334: Field name generated incorrectly
I have a `Subscription` model that has these lines:

```ruby
has_many :included_prices
accepts_nested_attributes_for :included_prices
```

I have a form that looks like this:

```erb
<%= form_for @subscription do |form| %>
  <%= form.hidden_field "included_prices_attributes[0][price_id]",
    value: price.id
  %>
  <%= form.submit %>
<% end %>
```

The hidden field is being generated with an improperly formatted name like this:

```html
<input
  value="solo_monthly_2023"
  autocomplete="off"
  type="hidden"
  name="billing_subscription[included_prices_attributes[0][price_id]]" 
  id="billing_subscription_included_prices_attributes[0][price_id]"
>
```

The name is: `billing_subscription[included_prices_attributes[0][price_id]]`
But it should be: `billing_subscription[included_prices_attributes][0][price_id]`

The square brackets are placed incorrectly.

I can work around it by explicitly setting a `name` attribute on the hidden field like this:

```erb
<%= form_for @subscription do |form| %>
  <%= form.hidden_field "included_prices_attributes[0][price_id]",
    name: "billing_subscription[included_prices_attributes][0][price_id]",
    value: price.id
  %>
  <%= form.submit %>
<% end %>
```

This form used to generate the correct markup. I recently updated my app from Rails `7.2.2.1` to `8.0.1`. With Rails `7.2.2.1` the markup was generated with the square brackets in the right place.

Edouard-chin:
### Motivation / Background

- Fix #54334

### Detail

- ### Problem

  Sending such param `subscription[prices_attributes[0][price_id]]=3`, where there is nested brackets would parse to this hash on Rails 7.2.2 and Rack 2:

  ```ruby
  {
    "subscription" => {
      "prices_attributes" => {
        "0" => {
          "price_id" => "3"
        }
      }
    }
  }
  ```

  On Rail 8.0, *with* Rack 2, the params are no longer parsed the same way and the result is:

  ```ruby
  {
    "subscription" => {
      "prices_attributes[0" => {
        "price_id" => {
          "[" => "456",
        }
      }
    }
  }
  ```

  ### Context

  On Rack 2, the content inside brackets is [extracted using a regex](https://github.com/rack/rack/blob/aa5a0f532aac7a57e4bc7e857eca1c38229f7b30/lib/rack/query_parser.rb#L90), whereas our parser uses a simpler solution and extract the content up until a "]" token is found, without taking in consideration that there could be a nested brack opening.

  ### Solution

  Tweaked the param builder to check if we have a nested bracket opening and trigger a deprecation message, while continuing to parse as it used to work on Rails 7.

  ### Rack 3

  On Rack 3, the params no longer parses as it used to on Rack 2. And the result is similar to what the param builder parses. 

### Checklist

### DIFF
--- a/actionpack/test/dispatch/param_builder_test.rb
+++ b/actionpack/test/dispatch/param_builder_test.rb
@@ -33,6 +33,28 @@ class ParamBuilderTest < ActiveSupport::TestCase
         assert_equal({ "foo" => { "bar" => "baz" } }, result)
       end
     end
+
+    test "(rack 2) when nested brackets are used in parameters" do
+      assert_deprecated(ActionDispatch.deprecator) do
+        query_string = "subscription[included_prices_attributes[0][price_id[1][name]]]=3"
+
+        result = ActionDispatch::ParamBuilder.from_query_string(query_string)
+        expected = {
+          "subscription" => {
+            "included_prices_attributes" => {
+              "0" => {
+                "price_id" => {
+                  "1" => {
+                    "name" => "3"
+                  }
+                }
+              }
+            }
+          }
+        }
+        assert_equal(expected, result)
+      end
+    end
   else
     test "(rack 3) defaults to retaining leading bracket" do
       result = ActionDispatch::ParamBuilder.from_query_string("[foo]=bar")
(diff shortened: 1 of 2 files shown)


actionpack/test/dispatch/param_builder_test.rb
```diff
@@ -33,6 +33,28 @@ class ParamBuilderTest < ActiveSupport::TestCase
         assert_equal({ "foo" => { "bar" => "baz" } }, result)
       end
     end
+
+    test "(rack 2) when nested brackets are used in parameters" do
+      assert_deprecated(ActionDispatch.deprecator) do
```
zzak, line 38: Just a knit but I like to match on the message to avoid collisions with other deprecations

alecslupu:
Hello! 
I am facing the same issue, and i confirm the fix works. Is there any chance to retake this ?  Or migrating it back to rails 8.0 ?

Edouard-chin:

PR #54638, opened 2025-02-27
[ci skip] Update getting_started.md

fuzzyalej:
Changed parameter to button_to in order to properly create a DELETE action using Turbo (not working as it is right now in the guide)


Edouard-chin:
Thank you for the pull request but the documention is correct. The `turbo_method` is only useful to modify the request type when you are not using a form. The `button_to` helper creates a form.

PR #53105, opened 2024-09-30
BroadcastLogger should support TaggedLogging correctly

armstrjare:
[Fixes #49745 #52876]
[Related #51883 #49771 #53093]

NB: I have a related pull request in #53093  that refactors BroadcastLogger to ensure that
block forms of delegated methods also only execute the block once. 

This commit extends BroadcastLogger when TaggedLogging is loaded in
order to ensure that BroadcastLogger#tagged behaves as expected when
there are multiple loggers being broadcast to.

TaggedLogging provides two related, but distinct interfaces:

1. Calling `logger.tagged(*tags)` without a block returns a new Logger
   with the provided tags pushed to its tag stack.

2. Calling `logger.tagged(*tags) { |logger| ... }` with a block pushes
   the tags to the existing Logger and yields the logger to block.

Previously, BroadcastLogger would only function as expected if there
was just one Logger being broadcast to. When there were multiple
loggers: when calling `tagged` with a block, it would yield a block
multiple times, once for each logger; when called without a block, it
would call `tagged` on each logger and return an Array of the results.

This inconsistent behaviour has caused issues such as those referenced
above. In development environments in particular, logger configuration
can often result in a BroadcastLogger with two loggers, since the
`bin/rails server` command will always broadcast to STDOUT, resulting
in confusing behaviour.

This commit provides a `BroadcastLogger#tagged` implementation that,
for the non-block form returns a new BroadcastLogger broadcasting to
the new tagged Loggers, and for the block-form, it 'wraps' the
user-provided block within nested calls to `TaggedLogging#logged`.
The user-provided block is only executed in the innermost call.

For example:

  ```ruby
  # BroadcastLogger with two Loggers being broadcasted to
  broadcaster.tagged("FOO") { |logger|
    ...
  }
  ```

Would execute similarly to:

  ```ruby
  broadcasts[0].tagged("FOO") {
    broadcasts[1].tagged("FOO") {
      yield(broadcaster)
    }
  }
  ```

PR #53093, opened 2024-09-29
Fix ActiveSupport::BroadcastLogger from executing a block argument for each logger (tagged, info, etc.)

armstrjare:
[Fixes #49745 #52876]
[Related #51883 #49771 #53105]

NB: I have a related pull request in #53105 that extends
BroadcastLogger to support TaggedLogging correctly. Both could be
merged independently of each other.

This commit refactors the implementation of dispatching method calls
to broadcasted loggers. The updated implementation ensures that any
block is only executed once. The first logger would recieve the block
as per normal, but subsequent loggers would only yield the result of
the initial execution.

The updated implementation of `dispatch` opened up an opportunity to
refactor the way each Logger method is delegatated to broadcasted
loggers - simplifying the delegator definitions.

Prior to these changes, BroadcastLoggers would iterate each broadcast
and re-execute the user provided block for each.

The consumer of any Logger would reasonably expect than when calling a
method with a block, that the block would only execute a single time.
That is, the fact that a Logger is a BroadcastLogger should be
irrelevant to consumer.

The most common example of this is when using
ActiveSupport::TaggedLogging and wrapping behaviour in a
`tagged(*tags) { }` block. But this also applies when using the block
form `info`, `warn` etc. If a BroadcastLogger is used, and there are
multiple loggers being broadcast to, then calling one of these methods
with a block would result in the block being executed multiple times.

For example:

```ruby
  broadcasts = ActiveSupport::BroadcastLogger.new(
    *Array.new(2) { ActiveSupport::Logger.new(STDOUT) }
  )
  number = 0
  broadcasts.info {
    number += 1
    "Updated number to #{number}"
  }
  # Expected:
  # Updated number to 1
  # Updated number to 1
  #
  # Actual:
  # Updated number to 1
  # Updated number to 2
```

After these changes, the behaviour of BroadcastLogger reflects the
expected behaviour above.

Issue #49757, opened 2023-10-24
BroadcastLogger#tagged without a block returns an array instead of a logger


mroach:
### Steps to reproduce

1. Use `BroadcastLogger#broadcast_to(logger)` to add a new logger
2. Use `BroadcastLogger#tagged("MyTag")` without a block to return a new tagged logger
3. Try to use that new tagged logger

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem "pry"
end

require "active_support"
require "active_support/core_ext/object/blank"
require "minitest/autorun"

class BugTest < Minitest::Test
  def test_broadcast_to_multiple_tagged_loggers
    iologger1 = Logger.new($stdout)
    iologger1 = ActiveSupport::TaggedLogging.new(iologger1)

    logger = ActiveSupport::BroadcastLogger.new(iologger1)

    # This works fine
    logger.tagged("Test1").info("OK")

    iologger2 = Logger.new($stderr)
    iologger2 = ActiveSupport::TaggedLogging.new(iologger2)

    logger.broadcast_to(iologger2)

    # This works fine
    logger.tagged("Test1") { _1.info "OK" }

    # Now we have a problem. `logger.tagged` returns an Array
    # NoMethodError: undefined method `info' for [#<Logger
    #
    # https://github.com/rails/rails/blob/ca5132b78a44677f708045905dc1b092a02f6b30/activesupport/lib/active_support/broadcast_logger.rb#L242
    logger.tagged("Test2").info("OK")

    # We'll never get here, but this fails
    assert logger.is_a?(ActiveSupport::BroadcastLogger)
  end
end

```

### Expected behavior

Using `ActiveSupport::BroadcastLogger#tagged` without a block returns a tagged logger that can be used for logging as normal.

### Actual behavior

When broadcasting to more than one logger that supports tagging, an `Array` is returned rather than an object that you can log to directly.

The specific use case where this came up: in some tests, I want to capture logs so I can test them. Previously I was replacing `Rails.logger` entirely, but now I tried adding and removing a broadcast target and I noticed this didn't work.

This appears to be caused by the `method_missing` implementation:
https://github.com/rails/rails/blob/ca5132b78a44677f708045905dc1b092a02f6b30/activesupport/lib/active_support/broadcast_logger.rb#L242

[...]

Issue #49745, opened 2023-10-23
`Rails.logger.tagged` will execute block as many times as the number of tagged loggers in the broadcast


davidstosik:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem "debug"
end

require "active_support"
require "active_support/core_ext/object/blank"
require "minitest/autorun"

module ActiveSupport
  class BugTest < Minitest::Test
    def test_tagged_broadcast
      logger1_io = StringIO.new
      logger2_io = StringIO.new

      logger = BroadcastLogger.new(TaggedLogging.new(Logger.new(logger1_io)))
      logger.broadcast_to(TaggedLogging.new(Logger.new(logger2_io)))

      run_counter = 0

      logger.tagged("TEST") do
        logger.info("Doing something")
        run_counter += 1
      end

      # require "debug";binding.b
      assert_equal 1, run_counter                                  # Actual: 2
      assert_equal "[TEST] Doing something\n", logger1_io.string   # Actual: "[TEST] Doing something\nDoing something\n"
      assert_equal "[TEST] Doing something\n", logger2_io.string   # Actual: "Doing something\n[TEST] Doing something\n"
    end
  end
end
```

### Expected behavior

When I do `Rails.logger.tagged("MY TAG") { do_something }`, I expect `do_something` to be called once, regardless of how many loggers Rails is broadcasting to.

I also expect my log to be tagged and displayed in each logger. (I should see "[TEST] Doing something" once in the first log sink, and once in the second log sink.)

I don't know how common this is, but I've seen many applications (starting with Shopify), passing rather long blocks to `#tagged`, expecting all logs produced by the code in that block to be tagged. This becomes a problem if we for example do database operations in that block.

### Actual behavior

The block passed to `tagged` is run as many times as there are loggers broadcasted to.
The log messages get multiplied too (obviously), and only one for each logger is tagged.

### System configuration

**Rails version**: `HEAD`
**Ruby version**: 3.2.1

---

Still working on a graceful but quick solution to handle this. Suggestions welcome!

davidstosik:
Here's an idea: implement a "recursive" `tagged` method in `BroadcastLogger`:

```rb
module ActiveSupport
  class BroadcastLogger
    # ...

    def tagged(*tags, &block)
[...]

PR #54640, opened 2025-02-27
ActiveJob: Improve compatibility with Zeitwerk

pboling:
Following are removed because `active_support` autoloads them:

    require "active_support/tagged_logging"
    require "active_support/logger"

In accordance with Rails Guides' documentation on how to load active_support, and in compliance with patterns (the good ones, not the bad ones) everywhere else in Rails code.

- See: https://guides.rubyonrails.org/active_support_core_extensions.html

### Motivation / Background

According to documentation [here on Rails Guides](https://guides.rubyonrails.org/active_support_core_extensions.html)
libraries that are not `active_support` that depend on parts of `active_support` should first do:

```
require "active_support"
```

and then whatever specific part they need.

Or, if they need the whole thing they should instead do:
```
require "active_support/all"
```

They should never drop in like a bird of prey and pluck out some internal part of `active_support` without first doing:

```
require "active_support"
```

Unfortunately this is what [active_job does](https://github.com/rails/rails/blob/main/activejob/lib/active_job/logging.rb#L3-L4), and it breaks some libraries I am working on.

Currently the top of `active_job/logging.rb` has:

```
require "active_support/tagged_logging"
require "active_support/logger"
```

It should be changed to:

```
require "active_support"
```

Since TaggedLogging and Logger will be loaded by `autoload`.

This affects every version of Rails that has active_job, tested with 7.2 and 8.0.

Tellingly, it is the only violation of the pattern set forth in the Rails Guides in the entire Rails code base, which is a strong indicator that it should be fixed.

### Detail

This Pull Request changes `activejob/lib/active_job/logging.rb`

### Checklist

### DIFF
--- a/activejob/lib/active_job/logging.rb
+++ b/activejob/lib/active_job/logging.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-require "active_support/tagged_logging"
-require "active_support/logger"
+require "active_support"
 
 module ActiveJob
   module Logging


pboling:
The buildkite/rails failure does not appear to be related in any way to this change.  Perhaps the test ENV is flaky?

activejob/lib/active_job/logging.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-require "active_support/tagged_logging"
-require "active_support/logger"
+require "active_support"
```
skipkayhil, line 3: `require "active_support"` already happens in the Active Job entrypoint, so it should not be added here.

https://github.com/rails/rails/blob/800976975253be2912d09a80757ee70a2bb1e984/activejob/lib/active_job.rb#L26

However, you are right that these requires can be removed because they reference an autoloaded constant (`TaggedLogging`) and an already required file (`active_support.rb` requires `active_support/logger`)

pboling: I agree.  It seems like there are historical patterns around the codebase of Rails which end up getting copied and pasted into other code, even though the paradigm has been dead for a long time.

I've pushed the branch again with the removal of that line as well.

pboling:
These are far from the only lines that exhibit requires that would throw a wrench into Zeitwerk, but since they're pre-existing I think removing them one file at a time is fine.

pboling:
The issue is that by hardcoding those requires it preempts Zeitwerk, and thus prevents other things that are very nice to have.

Rails currently only pretends to use Zeitwerk.

rafaelfranca:
Can you give me an example why it preempts Zeitwerk? I can't see how.

pboling:
The classes can't be autoloaded if they have been loaded by a require.

The beauty of autoload is that I can define the class in another gem, keeping the same API, and essentially swap out internals as needed.

These hard coded requires prevent me from doing that.

pboling:
I'm writing an extraction of the Rails logging toolchain where I can test and fix issues in the standard Rails logging toolchain.  Tagged Logging with Broadcasting has been fundamentally broken for a long time:
  
- https://github.com/rails/rails/issues/49757
- https://github.com/rails/rails/issues/49745
- https://github.com/rails/rails/issues/46084
- https://github.com/rails/rails/issues/44668

So I created these:

- Enhanced [activesupport-logger](https://github.com/pboling/activesupport-logger) which was ripped from Rails v8.0
- Enhanced [activesupport-broadcast_logger](https://github.com/pboling/activesupport-broadcast_logger) which was ripped from Rails v8.0, and https://github.com/rails/rails/pull/53093
- Enhanced [activesupport-tagged_logging](https://github.com/pboling/activesupport-tagged_logging) which was ripped from Rails v8.0, and https://github.com/rails/rails/pull/53105

As test beds for proposed fixes.  I also added back support for all versions of Rails back to 5.2, so they should be able to unify logging config (and crucially output) across a deployment of many discrete Rails apps running myriad versions of Rails.

It's working, except we still haven't migrated from Sidekiq job classes to ActiveJob, and those two lines are preventing us from switching.

PS. If you're thinking "why would anyone..." I can relate.  In this particular industry (FDA regulated-and-audited medical devices) being on a "fork of Rails" would require much more validation than being on a standard Rails with a "gem fork of the Rails' logger".

rafaelfranca:
Rails has different requirements of Rails applications. For example, you don't want to eager load the postgresql adapter in an application using mysql. Using zeitlwerk, would require us to white custom logic for this kind of requirements, so using zeitwerk in some of the frameworks doesn't bring much advantages.



> The classes can't be autoloaded if they have been loaded by a require.
> 
> The beauty of autoload is that I can define the class in another gem, keeping the same API, and essentially swap out internals as needed.
> 
> These hard coded requires prevent me from doing that.

This has nothing to do with Zeitwerk. What you are describing is that you want a way to monkey patch Rails by using autoload. All I can say is that isn't supported and never will.

rafaelfranca:
That being said. I makes me sad that you need to create all those gems. I'll make sure those issues are fixed, and I hope that helps you as well.

pboling:
I'd love to see movement on the logging bugs!

Regarding autoload, the dream is the one inherited from the merb 3 merge.  We could have a truly modular Rails if more parts of Rails relied exclusively on autoload.  I think that would be a pretty good thing, and enable some really interesting abilities.

PR #54484, opened 2025-02-10
Optimize JSON escaping

etiennebarrie:
Instead of using multiple gsub! calls, this uses binary encoding to do the regexp gsub with map, then forces encoding back to UTF-8.

This uses the same approach as before ebe0c4066b5b2a4b0e08220b04b33243b54d8e34, with binary encoding, which performs better.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-ips'
  gem 'rails', path: '.'
  gem "json", path: "../../ruby/json"
  gem "vernier"
end

require "active_support"
require "active_support/json/encoding"

obj = JSON.load_file("../../ruby/json/benchmark/data/twitter.json")

Benchmark.ips do |x|
  x.report "obj.to_json" do
    obj.to_json
  end
end

obj = "a"*obj.to_json.bytesize
Benchmark.ips do |x|
  x.report "obj.to_json nothing to escape" do
    obj.to_json
  end
end
```

Before:
```
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
         obj.to_json    28.000 i/100ms
Calculating -------------------------------------
         obj.to_json    290.133 (± 2.4%) i/s    (3.45 ms/i) -      1.456k in   5.020937s
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
obj.to_json nothing to escape
                       100.000 i/100ms
Calculating -------------------------------------
obj.to_json nothing to escape
                          1.027k (± 3.7%) i/s  (973.57 μs/i) -      5.200k in   5.069529s
```

After:
```
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
         obj.to_json    61.000 i/100ms
Calculating -------------------------------------
         obj.to_json    610.559 (± 4.6%) i/s    (1.64 ms/i) -      3.050k in   5.007098s
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
obj.to_json nothing to escape
                        97.000 i/100ms
Calculating -------------------------------------
obj.to_json nothing to escape
[...]

PR #54643, opened 2025-02-27
Don't always escape JSON when calling `render json:`

etiennebarrie:
When `render json:` is used in a controller, and there's no callback option (used for JSONP), the resulting JSON document doesn't need to be HTML-safe (no need to escape HTML entities) or embeddable into JavaScript (no need to escape U+2028 and U+2029).

This both saves a costly operation (the gsub optimized in #54484) and renders cleaner JSON.

This may break some apps or make them vulnerable so I think we'll need a new framework default. Also the name `html_safe` isn't great, by it we mean both JSON that can be included in a script tag in HTML and JSON that can be interpreted as JS (e.g. in a script tag or in a JSONP response). Maybe `browser_safe`?

cc @byroot

### DIFF
--- a/actionpack/lib/action_controller/railtie.rb
+++ b/actionpack/lib/action_controller/railtie.rb
@@ -133,5 +133,18 @@ module ActionController
         ActionController::TestCase.executor_around_each_request = app.config.active_support.executor_around_test_case
       end
     end
+
+    initializer "action_controller.escape_json_responses_deprecated_warning" do
+      config.after_initialize do
+        ActiveSupport.on_load(:action_controller) do
+          if ActionController::Base.escape_json_responses
+            ActionController.deprecator.warn(<<~MSG.squish)
+              Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
+              Set it to `false` or use `config.load_defaults(8.1)`.
+            MSG
+          end
+        end
+      end
+    end
   end
 end
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,7 +1,7 @@
 *   The JSON renderer doesn't escape HTML entities or Unicode line separators anymore.
 
-    Using `render json:` will no longer escape a few characters that can cause errors when the resulting JSON is
-    embedded in JavaScript, or vulnerabilities when the resulting JSON is embedded in HTML.
+    Using `render json:` will no longer escape `<`, `>`, `&`, `U+2028` and `U+2029` characters that can cause errors
+    when the resulting JSON is embedded in JavaScript, or vulnerabilities when the resulting JSON is embedded in HTML.
 
     Since the renderer is used to return a JSON document as `application/json`, it's typically not necessary to escape
     those characters, and it improves performance.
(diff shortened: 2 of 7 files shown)


byroot:
I just realized, another place where this would be useful is the Active Record JSON type: https://github.com/rails/rails/blob/3eb5f34ec33d0d20b23050f451ed18611996aced/activerecord/lib/active_record/type/json.rb

byroot:
> Also the name `html_safe` isn't great

Maybe just `escape: false`?

byroot:
I rebased the PR on main now that https://github.com/rails/rails/pull/54748 was merged.

The option is now `escape: false`.

Still unsure if we should add a new framework default or not.

etiennebarrie:
> Still unsure if we should add a new framework default or not.

Yeah it doesn't make a lot of sense as a long-term configuration, but for migrating apps I think it's pretty important, the alternative being to add `escape: true` to all `render json:` calls.

zzak:
If we're not sure how this will badly this will break existing applications:

* We should rollout with the existing behavior
* Set the new behavior with `load_defaults(8.1)`

https://github.com/rails/rails/blob/951ea8f75d9bf6d2999df2f0497347976dce9476/railties/lib/rails/application/configuration.rb#L351-L357

If it's not intended to stay long-term then we need a deprecation cycle for the flag.

This sounds like a lot of ceremony, but if there is potential to break applications, that is the usual approach.

byroot:
> Yeah it doesn't make a lot of sense as a long-term configuration,

Indeed. So I think we should deprecate it immediately. I'm trying to remember how best to do that.

bquorning:
This new setting seems kind of related to the existing setting `config.active_support.escape_html_entities_in_json`, but I don’t see it mentioned in the PR. Is there an overlap between the two settings?

byroot:
> Is there an overlap between the two settings?

Yes and no. The setting you point out disable some of these escapes, but globally, which I wish we could do, but there is probably too many people relying on those escapes for security, and we  basically can never change it.

Here the effect is only when you do `render json: something`. In that case it's much easier to assume you don't need that extra escaping.

actionpack/lib/action_controller/railtie.rb
```diff
@@ -133,5 +133,18 @@ class Railtie < Rails::Railtie # :nodoc:
...
       end
     end
+
+    initializer "action_controller.escape_json_responses_deprecated_warning" do
+      config.after_initialize do
+        ActiveSupport.on_load(:action_controller) do
+          if ActionController::Base.escape_json_responses
+            ActionController.deprecator.warn(<<~MSG.squish)
+              Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
+              Set it to `false` or use `config.load_defaults(8.1)`.
```
ghiculescu, line 143: It seems odd to recommend `load_defaults` here. Changing that will influence a lot of other things, not just this setting. We don't normally recommend this in deprecation warnings, is there a reason that it's done here?

ghiculescu: @byroot ^

ghiculescu: nvm commented heree https://github.com/rails/rails/pull/54840

PR #54646, opened 2025-02-28
Fix CurrentAttributes not setting an attribute's default:

### LINKED ISSUE #54636: Default value for ActiveSupport::CurrentAttributes attribute not set on first request after Zeitwerk reload
### Steps to reproduce

I've had a go at writing a failing reproduction but this issue relates to code reloading in development mode and I couldn't seem to emulate it in a test. Here's how you'd manually emulate it:

```
  attribute :has_permission_to, default: {}
end
```

Anywhere really, but in my case this is a method in a model:
```
def has_permission_to(access_type)
  Current.has_permission_to.fetch access_type do |key|
      Current.has_permission_to[key] = some_calculation
  end
end
```

### Expected behaviour

Saving code anywhere in the application (triggering a Zeitwerk reload) should reload Current and reinitialise the default value of `Current.has_permission_to` which is an empty hash.

### Actual behaviour

The first request after a code save (and Zeitwerk reload) where that request accesses the model method `has_permission_to(access_type)` triggers `undefined method 'fetch' for nil` indicating that the default value for `Current.has_permission_to` hasn't been set yet. Reloading the page succeeds and I'm not sure why.

### System configuration

**Rails version**:  `8.0.1`

**Ruby version**: `3.3.7`

Edouard-chin:
### Motivation / Background

- Fix #54636

### Detail

- ### Problem

  When calling an attribute that has a default, `nil` would be returned instead.

  ### Context

  CurrentAttributes uses a cache to store each CurrentAttributes subclasses instance. When the class method `instance` gets called, we instantiate the class and populate the cache. The defaults of each attributes are set when the class is instantiated.

  This basically means that once we instantiate the class, all attribute having a default that are later defined will be ignored.

  ### Details

  CurrentAttributes has a `method_added` hook which also call the `instance` class method. So depending on how your class is setup (e.g. you define a method and then define an attribute with a default), you'd hit the bug.

  This is even more sneaky when using a `delgate_to`:

  ```ruby
  class Current < CurrentAttributes
    attribute :user_session
    delegate :user, to: :user_session # Ends up triggering the `method_added` hook

    attribute :foo, default: 0 # Default would be ignored.
  end
  ```

  ### Solution

  Remove the instance from the cache each time we define an attribute.

  ### Alternative

  We could modify the `method_added` hook instead but we'd still be prone to this issue anytime `respond_to_missing` or the `method_missing` hooks are called as they also trigger the bug.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/current_attributes.rb
+++ b/activesupport/lib/active_support/current_attributes.rb
@@ -187,7 +187,7 @@ module ActiveSupport
           super
           return if name == :initialize
           return unless public_method_defined?(name)
-          return if respond_to?(name, true)
+          return if singleton_class.method_defined?(name) || singleton_class.private_method_defined?(name)
           Delegation.generate(singleton_class, [name], to: :instance, as: self, nilable: false)
         end
     end
--- a/activesupport/test/current_attributes_test.rb
+++ b/activesupport/test/current_attributes_test.rb
@@ -262,4 +262,18 @@ class CurrentAttributesTest < ActiveSupport::TestCase
       end
     end
   end
+
+  test "method_added hook doesn't reach the instance. Fix for #54646" do
+    current = Class.new(ActiveSupport::CurrentAttributes) do
+      def self.name
+        "MyCurrent"
+      end
+
+      def foo; end # Sets the cache because of a `method_added` hook
+
+      attribute :bar, default: {}
+    end
+
+    assert_instance_of(Hash, current.bar)
+  end
 end


activesupport/lib/active_support/current_attributes.rb
```diff
@@ -137,6 +137,7 @@ def attribute(*names, default: NOT_SET)
         Delegation.generate(singleton_class, names, to: :instance, nilable: false, signature: "")
         Delegation.generate(singleton_class, names.map { |n| "#{n}=" }, to: :instance, nilable: false, signature: "value")
 
+        current_instances.delete(current_instances_key)
```
byroot, line 140: I think it would be better if we avoided the call to `instance`, we can probably do it by not using `respond_to?` but `singleton_class.method_defined?`.

byroot:
Ah, I see you considered it:

> * We could modify the `method_added` hook instead but we'd still be prone to this issue anytime `respond_to_missing` or the `method_missing` hooks are called as they also trigger the bug.

I kinda fear the delete might cause the state to be lost at runtime though.

byroot:
I pushed an extra commit, let me know what you think.

Edouard-chin:
Thanks for pushing. I'll delegate the decision to you but I thought about this solution and was worried this would break classes using metaprogramming and where `public_method_defined` would return false but `respond_to?` on the instand would return true.

byroot:
> where `public_method_defined` would return false but `respond_to?` on the instance would return true.

Right, I think you got a point here, but I think that means `method_added` shouldn't call `respond_to?`

byroot:
I changed my commit a little bit.

Edouard-chin:
Yeah that make sense. Regarding my initial concern in the PR

> We could modify the method_added hook instead but we'd still be prone to this issue anytime respond_to_missing or the method_missing hooks

I can't really think of a legitimate way a user would be able to trigger the issue, it's possible but very unlikely.

I squashed the commits and modified the uncecessary changes in the tests. I added you as co-author. Thanks man

george-ma:
Hey @byroot and @Edouard-chin, this inadvertently introduced some breaking behaviour with `sorbet-runtime` when a type signature is defined.

If you have something like 

```ruby
class CurrentRequest < ActiveSupport::CurrentAttributes
     attribute :request

    sig { returns(T.untyped) }
    def request_id
      request&.request_id
    end
end
```

and then somewhere else call

```ruby
 self.request_id = CurrentRequest.request_id
 `sorbet-runtime` will error with `RuntimeError: The declaration for `request_id` is missing parameter(s): &`

**Comment by Edouard-chin** [comment]
👋 

Sorry for my ignorance as I have never used sorbet. If you declare a method with an anonymous block forwarding such as:

```ruby
def foo(&)
  bar(&)
end
```

Does sorbet support specifying a signature for the anonymous block parameter?

---------------

I'm also not entirely sure I understand why this issue arise with this patch, the delegation code with the anonymous block passing has been there for a while now (since https://github.com/rails/rails/commit/ddc32f5a47f70a02ea429ef9cce9cbf03e15be4f).

george-ma:
👋🏻 Heya, after getting some second thoughts I believe this is a problem with tapioca's RBI generation, this PR should be ok- thank you Edouard for following up 🙏🏻

KaanOzkan:
Hi Edouard 😄 ❤️  it's been a while.

> I'm also not entirely sure I understand why this issue arise with this patch, the delegation code with the anonymous block passing has been there for a while now (since https://github.com/rails/rails/commit/ddc32f5a47f70a02ea429ef9cce9cbf03e15be4f).

Looks like previously instance methods were covered by `respond_to?` and returning early, however, with this change we continue execution and append a `&` parameter to the method. So it's not directly related to this PR.

> Does sorbet support specifying a signature for the anonymous block parameter?

It doesn't support it but there is a workaround available. It's still not ideal for the user as first they need to observe the runtime error and then know they should change the sorbet signature (They can use `T::Sig::WithoutRuntime.sig` for future reference).

Since the usability isn't great I wanted to see if there is an alternative way of achieving this PR or see if `delegate` can define  a block argument only when the original method has a block argument 🙂

byroot:
> if `delegate` can define a block argument only when the original method has a block argument 🙂

That's unfortunately not possible because:

```ruby
def foo
  yield
end
```

Takes a block, but it's not part of it's signature. So we have no choice but to always forward blocks.

That being said, all this optimization was for previous versions of Ruby, with Aaron's optimiation in 3.4 we can very likely just use `...` delegation.

Edouard-chin:
Hey @KaanOzkan ❤️ !!

Thanks for the explanation. I didn't understand why this patch would suddenly trigger the sorbet runtime error but I got it now, and it's legit! I opened https://github.com/rails/rails/pull/54746 to explain, but I'm not sure if the current behaviour is actually the correct one.

This doesn't fix the root cause of the sorbet issue, and you may re-encounter it at some point, but we can try Jean's suggestion in a different patch.

PR #54647, opened 2025-02-28
Modify the Test Runner to allow running test at root:

Edouard-chin:
### Motivation / Background

I'd like to be able to do `bin/rails my_test.rb`. But I can't, as `my_test.rb` is not considered a path by the runner, so Rails ends up running the entire test suite.

This is also confusing, when I run `bin/rails test unexisting.rb` I don't get a load error, and the whole suite runs.

### Solution

Don't make the assumptions about the `/` character to consider the arg as a path. Instead, skip over any arg that is preceded by a flag, and consider all others to be a path.

### Caveat

(Not really one, since you could never do that.)
Running a file named `--file` or `-file` is not possible.
But this isn't anything new, Minitest will trip over when running the OptionParser and see that this "flag" isn't defined.

### Bonus point

You can now run a test that has a slash in its name. `bin/rails test -n foo\/bar`

### Checklist

### DIFF
--- a/railties/lib/rails/test_unit/runner.rb
+++ b/railties/lib/rails/test_unit/runner.rb
@@ -91,12 +91,22 @@ module Rails
 
         private
           def extract_filters(argv)
+            previous_arg_was_a_flag = false
             # Extract absolute and relative paths but skip -n /.*/ regexp filters.
             argv.filter_map do |path|
-              next unless path_argument?(path)
+              current_arg_is_a_flag = /^-{1,2}[a-zA-Z0-9\-_.]+=?.*\Z/.match?(path)
+
+              if previous_arg_was_a_flag && !current_arg_is_a_flag
+                # Handle the case where a flag is followed by another flag (e.g. --fail-fast --seed ...)
+                previous_arg_was_a_flag = false
+                next
+              end
 
               path = path.tr("\\", "/")
               case
+              when current_arg_is_a_flag
+                previous_arg_was_a_flag = true unless path.include?("=") # Handle the case when "--foo=bar" is used.
+                next
               when /(:\d+(-\d+)?)+$/.match?(path)
                 file, *lines = path.split(":")
                 filters << [ file, lines ]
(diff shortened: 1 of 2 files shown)


railties/lib/rails/test_unit/runner.rb
```diff
@@ -91,12 +91,22 @@ def compose_filter(runnable, filter)
...
             # Extract absolute and relative paths but skip -n /.*/ regexp filters.
             argv.filter_map do |path|
-              next unless path_argument?(path)
+              current_arg_is_a_flag = /^-{1,2}[a-zA-Z0-9\-_.]+=?.*\Z/.match?(path)
+
+              if previous_arg_was_a_flag && !current_arg_is_a_flag
+                # Handle the case where a flag is followed by another flag (e.g. --fail-fast --seed ...)
+                previous_arg_was_a_flag = false
+                next
+              end
```
byroot, line 103: I get the need, but this code seems a bit fragile to me. Rather than all this extra logic, have you tried to improve `path_argument?` ?

I think it could ultimately use `File.exist?` to figure out that something is a path.

Edouard-chin: I dug a bit more to check what less brittle alternative we can have, but I couldn't find any that would fix all current use cases and the missing ones I described in this PR.

> I think it could ultimately use File.exist? to figure out that something is a path.

If we were to do this, we still would need to determine whether we need to require the file (even in the event that `File.exist?` returns false). If we don't, we'd just skip over any un-existing files and running `bin/rails test test/unexisting_file.rb` would just run the entire test suite (instead of raising a LoadError).
So ultimately we'd have the same logic from this patch where we'd have to check whether the arg is an option from a flag.

Also take in consideration a random minitest plugin that accepts an option, we'd have no way to determine whether the arg is part of an option:

`bin/rails test test/test_file.rb --load-fixtures test/fixtures/foo.rb`
# puts ARGV #=> ['test/test_file.rb', '--load-fixtures', 'test/fixtures/foo.rb']

-----

I think the ideal solution would be to use the `OptionParser`, parse all options and return only the arguments. Rafael had a similar ish use case in https://github.com/ruby/optparse/issues/38, but the implemented solution doesn't fully work (it doesn't proxy short flags such as `-t`, and stops the parsing at the first unknown long flag).

PR #54655, opened 2025-02-28
Raise `DoubleRenderError` on `head` after rendering

viralpraxis:
Currently, calling `head` to set the response code silently removes a previously set response body [0]. This can lead to unnecessary and potentially expensive rendering, as shown in the following snippet:

```ruby
def index
  render json: User.all.as_json # expensive rendering
  head :ok # Overwrites `response_body` to ``
end
```

This behavior is inconsistent with calling `render` twice or calling `redirect_to` after `render`: both of which raise a `DoubleRenderError` [1, 2].

Since this change could break existing endpoints when upgrading Rails, it might make sense to introduce a configuration option would allow developers to choose whether this scenario raises an exception or a warning.

[0] https://github.com/rails/rails/blob/b721c62de11c2c29fa2b8ef51b00d95dce163afc/actionpack/lib/action_controller/metal/head.rb#L50
[1] https://github.com/rails/rails/blob/b721c62de11c2c29fa2b8ef51b00d95dce163afc/actionpack/lib/action_controller/metal/redirecting.rb#L103-L105
[2] https://github.com/rails/rails/blob/b721c62de11c2c29fa2b8ef51b00d95dce163afc/actionpack/lib/action_controller/metal/rendering.rb#L171-L174

### Motivation / Background

### Detail

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Raise `AbstractController::DoubleRenderError` if `head` is called after rendering.
+
+    After this change, invoking `head` will lead to an error if response body is already set:
+
+    ```ruby
+    class PostController < ApplicationController
+      def index
+        render locals: {}
+        head :ok
+      end
+    end
+    ```
+
+    *Iaroslav Kurbatov*
+
 *   The Cookie Serializer can now serialize an Active Support SafeBuffer when using message pack.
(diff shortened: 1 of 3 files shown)


zzak:
It seems obvious that this shouldn't be supported but if someone is incidently depending on it, maybe difficult to deal with

PR #54656, opened 2025-02-28
Add `rails_app` helper for initializing a Rails app inline.

p8:
### Motivation / Background

Initializing a Rails application in a single script currently requires setting multiple application configurations and calling `initialize!` on the application.
It is easy to miss configuration.

For example in the bug report templates we currently need to do the following:

```ruby
  config.eager_load = false
  config.hosts << "example.org"
end
Rails.application.initialize!
```

By wrapping this logic into a helper method, similar to `bundler/inline`, we can simplify this:

```ruby
require "rails/inline"

rails_app
```

Now we can create an application and for example start the console:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

end

require "rails/inline"

rails_app

rails :console
```

### Additional information

~~Setting `config.root` through a block seems required for now, as we need to pass the scripts location to the application.~~

Not all commands are supported yet. For example `rails :server` does not work yet. 

### Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -750,4 +750,4 @@ DEPENDENCIES
   websocket-client-simple
 
 BUNDLED WITH
-   2.5.16
+   2.6.5
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Add `rails_app` helper for initializing a Rails app inline.
+
+    ```ruby
+      require "bundler/inline"
+
+      gemfile(true) do
+        source "https://rubygems.org"
+
+        gem "rails", github: "rails/rails", branch: "main"
+      end
+
+      require "rails/inline"
+
+      rails_app do |config|
+        config.root = __dir__
+      end
+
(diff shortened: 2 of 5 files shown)


rafaelfranca:
I'm not saying we will accept this, but can we get the `root` from `Dir.pwd`?

I'm personally still in doubt if we should expose this since I don't know the use case that this could be enabled. Our bug report scripts aren't reason enough for adding things like this that we have to maintain. We need some use case that isn't possible today that this could unlock.

p8:
`Dir.pwd` works, thanks!

The main reason I'd like to use this for the ActionText bug report template.
It's currently not easy to create a single script that starts a webserver so we can test the integration of ActionText with Trix.
Instead a repo with a new rails app has to be created.
That's still for the bug report templates, though.

Another use case I can see is replacing the dummy application that is generated for testing engines with a singe file.

p8:
As the config that is encapsulated is only useful for bug reports, I don't think this is useful in other use cases.

PR #54662, opened 2025-03-01
 Support hash options for YJIT configuration

a5-stable:
### Motivation / Background
Currently, Rails allows enabling YJIT using [config.yjit = true](https://guides.rubyonrails.org/configuring.html#config-yjit), but it does not support configuring YJIT options through this setting.
While options such as stats and log can be enabled by setting RUBYOPT="--yjit-log", Rails does not provide a way to configure these options directly via config.yjit.
[YJIT itself ](https://github.com/ruby/ruby/blob/db4ea95219f045b8e855afae0b6d28027ec8d3eb/yjit.rb#L47)allows enabling these options through `RubyVM::YJIT.enable(stats: false, log: false)`, so it would be more convenient to manage them within the Rails configuration layer.

### Detail
This PR enhances `config.yjit` to accept a hash, allowing users to specify YJIT options.
```
config.yjit = true  # Enable YJIT with default settings
config.yjit = { stats: true }  # Enable YJIT with custom options (newly supported)
config.yjit = false  # Disable YJIT
```

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -641,6 +641,7 @@ See [Custom Configuration](#custom-configuration)
 
 Enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. If you are
 deploying to a memory constrained environment you may want to set this to `false`.
+Additionally, you can pass a hash to configure YJIT options such as `{ stats: true }`.
 
 | Starting with version | The default value is |
 | --------------------- | -------------------- |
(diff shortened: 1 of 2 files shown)


guides/source/configuring.md
```diff
@@ -648,6 +648,16 @@ deploying to a memory constrained environment you may want to set this to `false
...
 | 8.1                   | `!Rails.env.local?`  |
 
+#### `config.yjit_options`
+
+Allows you to set options for YJIT. The default value is an empty hash.
+When `config.yjit` is `false`, this setting will be ignored.
+
+```ruby
+config.yjit = true
+config.yjit_options = { stats: true }
```
byroot, line 658: I think we could simply allow:

```suggestion
config.yjit = { stats: true }
```

a5-stable: Thank you! As you pointed out, I simplified it by allowing a hash.

PR #43598, opened 2021-11-05
Extract ActiveSupport::ExecutionContext out of ActiveRecord::QueryLogs

casperisfine:
I'm working on a standardized error reporting interface (https://github.com/rails/rails/issues/43472) and it has the same need for a `context` than Active Record's query logs.

Just like query logs, when reporting an error you want to know what the current controller or job is, etc.

So by extracting it we can allow both API to use the same store.

### API design

`ActiveSupport::ExecutionContext..` is quite the mouth full, I think a shortcut such as `Rails.context` could make sense.

PR #54663, opened 2025-03-01
Reset execution context in jobs and controllers

BlakeWilliams:
### Motivation / Background

At GitHub we're spiking out tooling that identifies queries in CI, highlighting them to the PR author, linting for index usage, etc. While building this functionality we discovered that the query logs attributed a significant number of queries to a common job that's run inline in our test suite. After some investigation, we discovered that the cause was due to `:job` (and `:controller`) context values not being cleared after the job or controller had completed execution.

The job in question is commonly run in our test suite and we saw a ~4x reduction in the number of queries attributed to it after a small monkey patch to clear `ActiveSupport::ExecutionContext[:job]` post perform.

The accuracy of this tag is valuable in this testing context since it's highlighting a lot of potential optimizations we can make to our test suite and avoids potential confusion why when queries are attributed incorrectly. This can also impact production in some circumstances, like mistakenly attributing queries to the wrong source (e.g. running a `perform_now` could incorrectly begin attributing queries to that job instead its parent job, or queries made in a controller post-perform).

### Detail

Currently `ActiveSupport::ExecutionContext#[]` is used to set the `:job` and `:controller` attributes, which does not reset the values once the job or controller have completed execution. This can result in that context leaking, resulting in query logs that are misattributed.

This leaky behavior can be seen in production (e.g. performing a job via `perform_now`) but is much more noticeable in the test environment where it's not uncommon to run some combination of jobs/controller actions in a single test.

This updates both uses of `ActiveSupport::ExecutionContext#[]` to avoid leaking the context and generating inaccurate query logs.

### Additional information

I wasn't 100% positive if there was a better place to put these tests but thought they'd be valuable as small/cheap regression tests.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/instrumentation.rb
+++ b/actionpack/lib/action_controller/metal/instrumentation.rb
@@ -57,31 +57,31 @@ module ActionController
 
     private
       def process_action(*)
-        ActiveSupport::ExecutionContext[:controller] = self
+        ActiveSupport::ExecutionContext.set(controller: self) do
+          raw_payload = {
+            controller: self.class.name,
+            action: action_name,
+            request: request,
+            params: request.filtered_parameters,
+            headers: request.headers,
+            format: request.format.ref,
+            method: request.request_method,
+            path: request.filtered_path
+          }
 
-        raw_payload = {
-          controller: self.class.name,
-          action: action_name,
-          request: request,
-          params: request.filtered_parameters,
-          headers: request.headers,
-          format: request.format.ref,
-          method: request.request_method,
-          path: request.filtered_path
-        }
+          ActiveSupport::Notifications.instrument("start_processing.action_controller", raw_payload)
 
-        ActiveSupport::Notifications.instrument("start_processing.action_controller", raw_payload)
-
-        ActiveSupport::Notifications.instrument("process_action.action_controller", raw_payload) do |payload|
-          result = super
-          payload[:response] = response
-          payload[:status]   = response.status
-          result
-        rescue => error
-          payload[:status] = ActionDispatch::ExceptionWrapper.status_code_for_exception(error.class.name)
-          raise
-        ensure
-          append_info_to_payload(payload)
+          ActiveSupport::Notifications.instrument("process_action.action_controller", raw_payload) do |payload|
+            result = super
+            payload[:response] = response
+            payload[:status]   = response.status
+            result
+          rescue => error
+            payload[:status] = ActionDispatch::ExceptionWrapper.status_code_for_exception(error.class.name)
+            raise
+          ensure
+            append_info_to_payload(payload)
+          end
         end
       end
 
--- a/actionpack/test/controller/base_test.rb
+++ b/actionpack/test/controller/base_test.rb
@@ -358,3 +358,13 @@ class BaseTest < ActiveSupport::TestCase
     )
   end
 end
+
+class BaseInstrumentationTest < ActionController::TestCase
+  tests SimpleController
+
+  def test_clears_controller_execution_context
+    get :hello
+
+    assert_nil ActiveSupport::ExecutionContext.to_h[:controller]
+  end
+end
--- a/activejob/lib/active_job/execution.rb
+++ b/activejob/lib/active_job/execution.rb
@@ -63,9 +63,10 @@ module ActiveJob
 
     private
       def _perform_job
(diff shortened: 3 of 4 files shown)


byroot:
Unfortunately I don't think that's exactly the proper fix, because there's some expectation that `:controller` or `:job` remains in the context so that's it's still available when you are returning from the middleware chain.

But of course in test that's a bit different.

BlakeWilliams:
> Unfortunately I don't think that's exactly the proper fix, because there's some expectation that `:controller` or `:job` remains in the context so that's it's still available when you are returning from the middleware chain.

I think for `:controller` that makes _a bit_ more sense, although the behavior is still likely surprising to folks since you're only able to attribute the latter half of the middleware and completely miss out on attributing the initial half, which is (at least anecdotally) much more likely to run queries. So as-is consumers of this API get query logs that _seemingly_ attribute queries to middleware with a significant and hidden gap in coverage.

In at least in certain job scenarios, the existing behavior leads to plain incorrect results. I wrote up a [quick repro script to highlight that](https://gist.github.com/BlakeWilliams/b2f3a5f84e0eeb38b9772dd854bfe90d):

```
I, [2025-03-02T10:36:37.267797 #98625]  INFO -- : -------------------------------
I, [2025-03-02T10:36:37.267858 #98625]  INFO -- : BugTest: test_query_association
I, [2025-03-02T10:36:37.267865 #98625]  INFO -- : -------------------------------
I, [2025-03-02T10:36:37.268202 #98625]  INFO -- : Performing OuterJob (Job ID: bc1ca609-45f8-4865-a77c-7c5a25710a88) from Async(default)
D, [2025-03-02T10:36:37.272491 #98625] DEBUG -- :   Post Load (0.0ms)  SELECT "posts".* FROM "posts" ORDER BY "posts"."id" ASC LIMIT 1 /*application='TestApp',job='OuterJob'*/
I, [2025-03-02T10:36:37.272641 #98625]  INFO -- : Performing InnerJob (Job ID: 854efa1e-4fd3-4500-96ab-98e219fb7194) from Async(default)
Hello world
I, [2025-03-02T10:36:37.272711 #98625]  INFO -- : Performed InnerJob (Job ID: 854efa1e-4fd3-4500-96ab-98e219fb7194) from Async(default) in 0.1ms
D, [2025-03-02T10:36:37.272886 #98625] DEBUG -- :   Post Load (0.0ms)  SELECT "posts".* FROM "posts" ORDER BY "posts"."id" ASC LIMIT 1 OFFSET 1 /*application='TestApp',job='InnerJob'*/
I, [2025-03-02T10:36:37.272913 #98625]  INFO -- : Performed OuterJob (Job ID: bc1ca609-45f8-4865-a77c-7c5a25710a88) from Async(default) in 4.82ms
```

Both queries occur in the `OuterJob`, but after performing the `InnerJob` all subsequent queries are associated with it. This is incorrect attribution in both test, and prod environments. We'll also see misattribution in controllers too, with queries being associated to the job after the job has already been performed.

As-is though, I'd argue that the existing behavior is often misleading, if not incorrect. I think there's a strong argument for changing the job behavior given how much more common it would be to run into it leaking query tags.

> Unfortunately I don't think that's exactly the proper fix

Either way, I would definitely be interested in what you imagine the proper fix might be because I'd love to fix this behavior  upstream vs just in our application(s).

byroot:
> although the behavior is still likely surprising to folks since you're only able to attribute the latter half of the middleware and completely miss out on attributing the initial half,

Yes, but it's not like you can identify the controller before going through the router, so it's a bit innevitable.

> I would definitely be interested in what you imagine the proper fix might be

I haven't put much thoughts into it yet, I'm just saying that for the case of controllers at least, it's expected that the context remain. If we want to clear it in test environment, I think it's more the job of the test helpers to do it.

For jobs specifically, it might make sense to restore it right away.

byroot:
> I'd be happy to split the controller and job changes out into separate PRs if that's helpful.

Sure, let's go with that.

byroot:
> I believe that intent changed at some point since it looks like it [was behaving as desired before it was merged with (or into?) `ActiveSupport::ExecutionContext`](https://github.com/rails/rails/blob/f5a7d2e266a8d0deb9257a804b2909c72345ed7a/actionpack/lib/action_controller/metal/query_tags.rb#L7-L14).

Here's the context: https://github.com/rails/rails/pull/43598#discussion_r743720964

PR #53950, opened 2024-12-14
Support joins in `update_all` for Postgresql and SQlite

byroot:
Fix: https://github.com/rails/rails/issues/13496

Both support UPDATE with JOIN but in a weird way. The first join has to be turned into a `FROM + WHERE`.

Also they doesn't support UPDATE with LIMIT/OFFSET nor with GROUP BY, so in such case there's really nothing we can do.

PR #54664, opened 2025-03-01
Add Relation method for updating many records with different values

brunodccarvalho:
### Motivation / Background

Rails lacks an abstraction to update many records in a table while setting different values for each one. Workarounds include using `Relation.upsert_all` with conflicting primary keys or multiple queries to update the records one by one.

This adds a method `Relation.uniform_update_all` which implements a limited version of this abstraction. The interface resembles `upsert_all` and does not instantiate records. The update query generated uses the pattern of joining with a handwritten [VALUES table](https://www.postgresql.org/docs/current/sql-values.html) that holds the row matching condition values and the assignment values. The construct is supported on all the engines with modest variation in syntax to account for typing and column aliasing.

### Detail

A simple example of the construct with a join on the primary key (postgres syntax):
```sql
UPDATE "books" SET "name" = "data"."column2"
  FROM (VALUES (1, "Reword"), (2, "Peopleware")) AS "data" ("column1", "column2")
  WHERE "books".id = "data"."column1";
```

It would be generated from this code:
```rb
# [[conditions, updates], ...] syntax
Book.uniform_update_all([
  [{ id: 1 }, { name: "Reword" }],
  [{ id: 2 }, { name: "Peopleware" }]
])

# shorthand syntax when conditions is the primary key
Book.uniform_update_all({
  1 => { name: "Reword" },
  2 => { name: "Peopleware" },
})
```

All rows of the `VALUES` table have identical schema, so we require that the conditions and assignments objects all have the same set of keys, just like `Relation.upsert_all` does.

Another example, here the conditions are not the primary key:
```rb
# one-shot migration to combine two columns into one
Book.uniform_update_all [
  [{ written: false, published: false }, { status: :proposed }],
  [{ written: true,  published: false }, { status: :written }],
  [{ written: true,  published: true },  { status: :published }]
]
```

An example of the construct from the postgres docs linked above, which this abstraction cannot assemble (non-trivial assignment and join):
```sql
UPDATE employees SET salary = salary * v.increase
  FROM (VALUES(1, 200000, 1.2), (2, 400000, 1.4)) AS v (depno, target, increase)
  WHERE employees.depno = v.depno AND employees.sales >= v.target;
```

### Additional information

Running `EXPLAIN` on all engines suggests they do run these queries as one would expect, using the correct index based on the join condition, but I have no benchmarks yet.

This was dependent on #53950 for postgres and sqlite, with the same caveats noted there. Since the derived table is mentioned in the `SET` clause, the query cannot have its join fully converted to a subquery, which is currently done when there are `order`, `limit`, `offset`, `group` or `having` clauses in the relation.

Bonus: the constraint that the assignment objects need to have the same keys can be lifted by introducing boolean columns to flag blank entries (or inplace `NULL` if none of the assignment values is `NULL`), this is not implemented:

```sql
UPDATE "products"
SET "price" = (CASE WHEN "data"."update_price" THEN "data"."price" ELSE "products"."price" END),
    "name"  = (CASE WHEN "data"."update_name"  THEN "data"."name"  ELSE "products"."name" END)
FROM (VALUES (1, 'Electronics', 199.99, NULL, TRUE, FALSE), (2, 'Books', NULL, 'New Book', FALSE, TRUE))
    AS "data" ("category_id", "type", "price", "name", "update_price", "update_name")
WHERE "products"."category_id" = "data"."category_id" AND "products"."type" = "data"."type";
```

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
@@ -559,6 +559,14 @@ module ActiveRecord
         false
       end
 
+      def supports_derived_values_table?
+        true
+      end
+
+      def supports_derived_table_column_aliases?
+        false
+      end
+
       def return_value_after_insert?(column) # :nodoc:
         column.auto_populated?
       end
--- a/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
@@ -188,6 +188,14 @@ module ActiveRecord
         end
       end
 
+      def supports_derived_values_table?
+        mariadb? || database_version >= "8.0.0"
+      end
+
+      def supports_derived_table_column_aliases?
+        !mariadb? || database_version >= "11.7.0"
+      end
+
       def get_advisory_lock(lock_name, timeout = 0) # :nodoc:
         query_value("SELECT GET_LOCK(#{quote(lock_name.to_s)}, #{timeout})") == 1
       end
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -288,6 +288,10 @@ module ActiveRecord
         database_version >= 10_00_00 # >= 10.0
       end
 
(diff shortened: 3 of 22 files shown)


brunodccarvalho:
Published as a gem [activerecord-updateinbulk](https://github.com/brunodccarvalho/activerecord-updateinbulk)

PR #50427, opened 2023-12-22
[ActiveStorage] Replace test/dummy with railtie-like test harness

zzak:
## Motivation

The dummy apps were last updated in #47259 and previously in #42286.

There is an interest from inside the core team to get rid of them.

> In my option, It doesn't make sense to upgrade dummy apps inside the framework every time we change the framework itself. Probably those dummy apps get outdated daily, and it is very likely we will change Rails in a way that would break those frameworks but the test will not catch because the dummy app is outdated.
> 
> If you want to work in that direction, it would be great.

PR #54667, opened 2025-03-02
Remove test/dummy from ActiveStorage

zzak:
Replaces #50427

The goal here is to remove the committed Rails app that lives in
test/dummy used for initializing Rails and thus ActiveStorage.

Instead, replace it with a test setup only required to load the parts it
is testing.

By using the Rails namespace as a mock for the application for two main
reasons:

1. House the routes, but that should be fixed in a
   future PR that makes them modular
2. Some migrations depend on Rails config to determine the primary key
   type, there is no public API to get that info since it's applied
   directly to the migration generator during initialization

The plan would be to refactor any dependencies, both in test and
library, that depend on the Rails namespace to use an appropriate
framework API or introduce one if necessary.

Alternatives considered were _actually_ initializing a Rails app, or
generating one dynamically but that is what Railties does. And we should
be able to test most-if-not-all of the engine without Rails, like we do
with ActionController, ActiveJob, etc.


activestorage/test/test_helper.rb
```diff
@@ -2,19 +2,186 @@
...
 require "active_support/current_attributes/test_helper"
 require "active_record/testing/query_assertions"
 
+require "zeitwerk"
+loader = Zeitwerk::Loader.new
+loader.push_dir("app/controllers")
+loader.push_dir("app/controllers/concerns")
+loader.push_dir("app/jobs")
+loader.push_dir("app/models")
+loader.setup
```
rafaelfranca, line 18: What is this for?

zzak: We setup an autoloader so the engine's classes are in the path, otherwise we get:

```
test/test_helper.rb:150:in '<top (required)>': uninitialized constant ActiveStorage::Blob (NameError)

ActiveStorage::Blob.services = ActiveStorage::Service::Registry.new(SERVICE_CONFIGURATIONS.m...
             ^^^^^^
        from ruby/3.5.0+4/bundled_gems.rb:55:in 'Kernel.require'
        from ruby/3.5.0+4/bundled_gems.rb:55:in 'block (2 levels) in Kernel#replace_require'
        from test/analyzer/audio_analyzer_test.rb:3:in '<top (required)>'
```

PR #54669, opened 2025-03-02
Remove test/dummy from ActionText

zzak:
Replaces #50428

This is the same approach taken for ActiveStorage. (#54667)

Splitting them up into separate PRs for review.


zzak:
:red_circle: This failure is legit and I'm looking into it:
https://buildkite.com/rails/rails/builds/117127#019558f4-63a6-4d63-a723-33e9dc17451d/1281-1298

railties/test/application/action_text/system_test_helper_test.rb
```diff
@@ -0,0 +1,62 @@
...
+
+require "isolation/abstract_unit"
+require "application/action_text_integration_test_helper"
+
+module ApplicationTests
+  class ActionTextSystemTestHelperTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    # See rails/rails#49527 and rails/buildkite-config#33
+    unless ENV["BUILDKITE"]
```
zzak, line 11: This fixes the failing railties tests, because chrome is not available.

There is a PR for it, but doesn't target the `railties` service, so that should be fixed but can happen in a separate PR.

PR #53797, opened 2024-12-01
Introduce versions formatter for the schema dumper

fatkodima:
Follow up to #44363 (cc @ghiculescu).

Previously, the versions were sorted in decreasing order and new added to the top of the list. Within large-enough teams, this can cause many merge conflicts near the top of this list. Within my company, we have lots of developers pushing lots of schema changes and half of the time someone merges a PR, lots of existing PRs now have conflicts with the master.

Now, the versions are sorted by their reverse representations (`20241201183002` -> `20038110214202`), in decreasing (can be increasing, does not matter, but to be consistent with the previous implementation) order. This will greatly reduce the likelihood of merge conflicts as now new versions can be inserted at any place in the existing list. I did a simple calculation and while before we had a likelihood of conflict within a list of, say 5 top elements (20%), now - within a whole list of 600+ items (0.1%).

I also added a comment with reversed version representation for each version within dumped `structure.sql` file for people to more easily solve the merge conflicts. But we can remove this and add a comment at the top of the `INSERT` describing the algorithm.

For existing apps, there will be a large diff the next time `structure.sql` is generated.

PR #54670, opened 2025-03-02
Add sql_filesystem_versions schema format

rwojsznis:
### Motivation / Background

Introduce new schema_version format which piggybacks on top of `sql` format, but instead of dumping `schema_information` as one big SQL statement it keeps versioning on the filesystem. This minimizes the risk of merge conflicts on that statement which still might occur despite attempts done in #53797

I briefly [suggested proposal of this feature here](https://discuss.rubyonrails.org/t/feature-proposal-database-schema-version-handling-outside-of-structure-sql/88615), despite lack of feedback I have decided to deep dive into ActiveRecord's internal. Please do check linked thread for more context and references.

### Detail

Spoiler alert: mostly I have no idea what I'm doing. I found finding a clear boundaries, naming things and sticking to existing patterns extremely difficult - mostly due to lack of context. A lot of things here just feels wrong and I expect this PR is not merge-able in current form, but it should at least present a working proof of concept.

### Additional information

Here you can find a [sample repository](https://github.com/rwojsznis/rails-test-app) and I have [recorded a youtube video](https://www.youtube.com/watch?v=hB6kdCPybGg) which a brief demonstration of this change.

### Checklist


activerecord/lib/active_record/migration/sql_filesystem_versions.rb
```diff
@@ -0,0 +1,106 @@
...
+require "digest"
+require "tempfile"
+require "fileutils"
+
+module ActiveRecord
+  class Migration
+    # Handles schema dumping and loading operations for the :sql_filesystem_versions schema format.
+    # This format keeps the schema in SQL format, but stores migration versions in separate files
+    # to minimize conflicts when merging branches.
+    class SqlFilesystemVersions
```
rwojsznis, line 12: I had an existential crisis naming this, I _badly_ waned to escape `DatabaseTask` - as adding tests there was a bit challenging, file have almost 2k lines already; ended up in `ActiveRecord::Migration` which seems bit _off_

activerecord/lib/active_record/tasks/database_tasks.rb
```diff
@@ -385,6 +385,9 @@ def load_schema(db_config, format = ActiveRecord.schema_format, file = nil) # :n
           load(file)
         when :sql
           structure_load(db_config, file)
+        when :sql_filesystem_versions
+          migration_operation = ActiveRecord::Migration::SqlFilesystemVersions.new(db_config, migration_connection_pool, self)
```
rwojsznis, line 389: as per comment above - I wanted to somehow break out of this namespace to unit test this thing without baggage of `DatabaseTasks` module; any better ideas? 🤔

guides/source/configuring.md
```diff
@@ -641,7 +641,6 @@ See [Custom Configuration](#custom-configuration)
 
 Enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. If you are
 deploying to a memory constrained environment you may want to set this to `false`.
-Additionally, you can pass a hash to configure YJIT options such as `{ stats: true }`.
```
rwojsznis, line 644: mhmm, not sure what happened here, will rebase (also `railties/lib/rails/application/finisher.rb`)

activerecord/test/cases/migration/sql_filesystem_versions_test.rb
```diff
@@ -0,0 +1,135 @@
...
+        end
+
+        attr_reader :loaded_content
+      end
+
+      def setup
+        @connection_pool = ActiveRecord::Base.connection_pool
+        connection = ActiveRecord::Base.connection
+
+        # TODO: what is the proper way of dealing with lack of current_database in sqlite adapter?
```
rwojsznis, line 37: yeah, that needs to be sorted out 🆘

railties/test/application/rake/dbs_test.rb
```diff
@@ -461,6 +461,178 @@ def db_schema_sql_dump
...
+        Dir.chdir(app_path) do
+          rails "generate", "model", "book", "title:string"
+          rails "db:migrate"
+
+          assert_not File.exist?("db/schema.rb")
+          assert File.exist?("db/structure.sql")
+
+          schema_migrations = Dir["db/schema_migrations/*"].map { |f| File.basename(f) }
+          assert_equal 1, schema_migrations.size
+          assert_match(/\A\d{14}\z/, migration_files.first)
```
rwojsznis, line 477: whops, some leftovers after rename (I rewrote this whole PR like 4 times 😅 )

will resolve shortly

fatkodima:
Note about #53797: I introduced these changes via a monkey patch in my company almost after the PR was merged. 2 months later, with tens of developers constantly committing migrations I saw the conflict related to versions dumped information only once, throughout the whole company. While before there were conflicts almost every day.

I am sure @ghiculescu saw similar results. 

So I am 100% sure that the changes in this PR are not needed and it can be closed.

rwojsznis:
While I appreciate the mentioned improvement, it seems it doesn't fully solve the problem 🤔 

I think it really depends on your git flow; with shared non-production branches - having a shared `staging` or `testing` environment+branch seems like a common pattern across small to medium companies according to, well, _my_ experience 🤷 

[You can clone sample repo](https://github.com/rwojsznis/rails-test-app) and try it out.

```
> git checkout testing
Switched to branch 'testing'
Your branch is up to date with 'origin/testing'.
>  git merge feature-in-the-works
Updating f0810ab..7cfe9ac
Fast-forward
 db/animals_structure.sql                  | 2 +-
 db/migrate/2021050100000_wip_migration.rb | 5 +++++
 db/structure.sql                          | 1 +
 3 files changed, 7 insertions(+), 1 deletion(-)
 create mode 100644 db/migrate/2021050100000_wip_migration.rb
>  git merge some-other-feature
Auto-merging db/structure.sql
CONFLICT (content): Merge conflict in db/structure.sql
Recorded preimage for 'db/structure.sql'
Automatic merge failed; fix conflicts and then commit the result.
```

(I will try to prep some more use cases after 8th of March)

fatkodima:
You have only 2 migrations in the linked repo, so of course there will be conflicts at the beginning. The more migrations you have, the less conflicts you will have. With only 10 migrations you already have only a 10% chance to have a conflict.

That PR is not meant to fully solve the problem, but is solves by 95% almost without any changes to the framework and to the development workflow, so is good enough.

marzdrel:
I've not seen the previously merged PR before https://github.com/rails/rails/pull/53797. I just took a look while reading this idea here. I share the sentiment of some of the commenters and dislike the reversed-timestamp solution.

The https://github.com/rails/rails/pull/53797 seems to solve the problem in many cases, at least for apps with lots of migrations, but does that at the cost of basically losing the readability of the timestamp. You could just generate a random number at this point. I am not a fan of the proposed `proc` as well, since that kind of defeats the purpose of convention over configuration and might introduce confusion while moving between projects.

Solution proposed in this PR seems pretty good, as it seems to address the problem as whole, without any remaining edge-cases. It also doesn't seem to require any changes to values in existing tables, keeps the format familiar, keeps the information carried by the timestamp.

> _So I am 100% sure that the changes in this PR are not needed and it can be closed._

@fatkodima You cannot be 100% sure of anything in this game. 😅 

Seeing you are the author of the https://github.com/rails/rails/pull/53797 i feel your opinion is somewhat biased. I would suggest we should let other, less involved people compare both solutions and state their opinions. 

PS. btw: Does the https://github.com/rails/rails/pull/53797 addresses the problem of changing existing values in database? You mention a large diff, but is there something preventing the migration from being re-executed on first run? This solution here doesn't seem to have this problem, as the values remain unchanged.

fatkodima:
Maybe the original PR's description is stale with the implementation or not well written, but timestamp values are preserved, only the `INSERT` ordering in the `structure.sql` file can be changed via the configuration. Try to look at the implementation.

marzdrel:
> Maybe the original PR's description is stale with the implementation or not well written, but timestamp values are preserved, only the `INSERT` ordering in the `structure.sql` file can be changed via the configuration. Try to look at the implementation.

I've tested the https://github.com/rails/rails/pull/53797 in the meantime and realized it works a bit differently. It doesn't change anything regarding the version number just uses provided logic to determine the order in the `structure.sql`. 

To this day we struggle with people sorting the migrations in a wrong order while resolving the conflict (since they need to do reverse order after 7.1 change). I am not sure what will be the outcome of the order which is way harder to be reformatted by the developer. 

Still, I would reconsider the solution from this PR, as the sole fact that it removes this insert from `structure.sql` and resolves the problem entirely is still somewhat still appealing to me.

PS. The [linked video](https://www.youtube.com/watch?v=hB6kdCPybGg) mentions that this solution was introduced in GitLab a long time ago. I haven't worked with GitLab. Seeing they have approximately 3300 migrations and indeed [use the files to register migrations](https://gitlab.com/gitlab-org/gitlab/-/tree/master/db/schema_migrations), I would love to hear how they found that solution worked for them.

ghiculescu:
> To this day we struggle with people sorting the migrations in a wrong order while resolving the conflict

We saw this too, until we told them to stop manually sorting them. Just run `db:migrate` again and it'll fix itself up. Or leave them out of order - the order doesn't matter at all.

> I am sure @ghiculescu saw similar results.

Yes - very happy with that PR.

My opinion on this PR is a bit icky at the idea of thousands of empty new files being added to the repo.

byroot:
I agree with @fatkodima that this isn't a good solution. While you can prune old migrations (and it's even recommended), pruning old migration versions is more tricky. So this format would lead to tons of files in the repo, which isn't good UX.

Thanks for the PR though.

PR #54683, opened 2025-03-03
Don't deserialize mutable defaults

### LINKED ISSUE #54681: Bug: prior attributes are being cached when using store_accessor only
We noticed a regression in our codebase after updating rails to `7845576df0fb68ff38f9ca754fca457a8e6758a3` ([offending PR](https://github.com/rails/rails/pull/54585).  Basically, we use `store_accessor` directly to store configuration details on a record.  After updating, the configuration hash (store) is defaulting to include previously saved data for a different record.  We are using `pg` but I can reproduce the error in `sqlite` as well below:

### Steps to reproduce
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.json :extra_data, default: {}, null: false
  end
end

  # store :extra_data, accessors: [:fav_color, :homepage], coder: JSON
  store_accessor :extra_data, :email
end

  def test_of_default_will_fail
    user = User.create(email: 'test@rails.org')

    new_user = User.new
    assert_nil new_user.email
  end
end
```
### Output
```
code ~ ruby robtest.rb
Fetching https://github.com/rails/rails.git
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
-- create_table(:users, {:force=>true})
D, [2025-03-03T13:58:07.765954 #23805] DEBUG -- :    (1.8ms)  DROP TABLE IF EXISTS "users"
D, [2025-03-03T13:58:07.766144 #23805] DEBUG -- :    (0.1ms)  CREATE TABLE "users" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "extra_data" json DEFAULT '{}' NOT NULL)
   -> 0.0048s
D, [2025-03-03T13:58:07.767016 #23805] DEBUG -- :    (0.0ms)  CREATE TABLE "schema_migrations" ("version" varchar NOT NULL PRIMARY KEY)
D, [2025-03-03T13:58:07.768186 #23805] DEBUG -- :    (0.0ms)  CREATE TABLE "ar_internal_metadata" ("key" varchar NOT NULL PRIMARY KEY, "value" varchar, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL)
D, [2025-03-03T13:58:07.782212 #23805] DEBUG -- :   ActiveRecord::InternalMetadata Load (1.9ms)  SELECT * FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = ? ORDER BY "ar_internal_metadata"."key" ASC LIMIT 1  [[nil, "environment"]]
D, [2025-03-03T13:58:07.784279 #23805] DEBUG -- :   ActiveRecord::InternalMetadata Create (0.0ms)  INSERT INTO "ar_internal_metadata" ("key", "value", "created_at", "updated_at") VALUES ('environment', 'development', '2025-03-03 18:58:07.782239', '2025-03-03 18:58:07.782240') RETURNING "key"
Run options: --seed 45293

# Running:

D, [2025-03-03T13:58:07.829787 #23805] DEBUG -- :   TRANSACTION (0.0ms)  BEGIN immediate TRANSACTION
D, [2025-03-03T13:58:07.829884 #23805] DEBUG -- :   User Create (0.1ms)  INSERT INTO "users" DEFAULT VALUES RETURNING "id"
D, [2025-03-03T13:58:07.829966 #23805] DEBUG -- :   TRANSACTION (0.0ms)  COMMIT TRANSACTION
F

Finished in 0.003214s, 311.1388 runs/s, 311.1388 assertions/s.

  1) Failure:
BugTest#test_of_default_will_fail [robtest.rb:38]:
Expected "test@rails.org" to be nil.

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```
### Expected behavior
A new `User` should have nil default values in the test above.

### Actual behavior

Instead, the prior email value is now defaulted to all new records.

### System configuration

**Rails version**: 
`gem 'rails', github: 'rails/rails', branch: 'main', ref: '7845576df0fb68ff38f9ca754fca457a8e6758a3'`

**Ruby version**: 

`ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin23]`

byroot:
Fix: https://github.com/rails/rails/issues/54681
Followup: https://github.com/rails/rails/pull/54585

We can't deserialize mutable types because otherwise mutations would leak, or we'd need to deep dup them. It's more efficient to keep them serialized in default.

However for immutable types like Integer, it's both faster and more correct to eagerly deserialize them.

### DIFF
--- a/activerecord/lib/active_record/attributes.rb
+++ b/activerecord/lib/active_record/attributes.rb
@@ -244,10 +244,7 @@ module ActiveRecord
           # TODO: Remove the need for a connection after we release 8.1.
           attributes_hash = with_connection do |connection|
             columns_hash.transform_values do |column|
-              type = type_for_column(connection, column)
-              default = column.default
-              default = type.deserialize(default) unless default.nil?
-              ActiveModel::Attribute.from_database(column.name, default, type)
+              ActiveModel::Attribute.from_database(column.name, column.default, type_for_column(connection, column))
             end
           end
 
--- a/activerecord/lib/active_record/connection_adapters/column.rb
+++ b/activerecord/lib/active_record/connection_adapters/column.rb
@@ -22,7 +22,7 @@ module ActiveRecord
         @cast_type = cast_type
         @sql_type_metadata = sql_type_metadata
         @null = null
-        @default = default
+        @default = default.nil? || cast_type.mutable? ? default : cast_type.deserialize(default)
         @default_function = default_function
         @collation = collation
         @comment = comment
@@ -117,7 +117,7 @@ module ActiveRecord
         def deduplicated
           @name = -name
           @sql_type_metadata = sql_type_metadata.deduplicate if sql_type_metadata
-          @default = -default if default
+          @default = -default if String === default
           @default_function = -default_function if default_function
           @collation = -collation if collation
           @comment = -comment if comment
--- a/activerecord/test/cases/adapters/postgresql/geometric_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/geometric_test.rb
@@ -131,11 +131,11 @@ class PostgresqlPointTest < ActiveRecord::PostgreSQLTestCase
   end
 
   def test_legacy_default
-    assert_equal ActiveRecord::Point.new(12.2, 13.3), PostgresqlPoint.column_defaults["legacy_y"]
-    assert_equal ActiveRecord::Point.new(12.2, 13.3), PostgresqlPoint.new.legacy_y
+    assert_equal [12.2, 13.3], PostgresqlPoint.column_defaults["legacy_y"]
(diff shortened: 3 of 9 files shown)


byroot:
> are custom types

Note that this was an issue before when I was dealing with Active Model attributes, but here I use the `cast_type`, which I think is never custom?

ghiculescu:
@byroot is this a change to public API? Given the change to tests required in https://github.com/Shopify/ruby-lsp-rails/pull/589 etc.

byroot:
I personally see it as a bug fix, as evidenced by the surprise in that thread, but it's always debatable of course.

PR #52369, opened 2024-07-19
Session generator - NoMethodError on sign out

yshmarov:
When an authenticated user clicks `button_to 'Sign out', session_path, method: :delete>, he gets an error:

Screenshot 2024-07-19 at 21 39 54

This will fix it.

PR #54686, opened 2025-03-04
Generate mailer files in auth generator only if ActionMailer is used

ramimassoud:
### Motivation / Background

This Pull Request has been created because if the authentication generator is run in an app created with `--skip-action-mailer`, the app will be silently broken. When deployed to an environment where `eager_load` is true, the app will fail to start as the generated `PasswordMailer` inherits from a nonexistent `ApplicationMailer`.

Fixes #54501 

### Detail

This Pull Request changes the authentication generator to skip all of the mailer-related files if `ActionMailer` is not defined. Also leave `PasswordsController#create` as blank to be implemented by the user.

### Additional information

In my use case, a simple login was enough, and I created user(s) manually in a console. I do not need a password reset flow. An alternative approach would be to require an app to have `ActionMailer` to run the generator.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/authentication/templates/app/controllers/passwords_controller.rb.tt
+++ b/railties/lib/rails/generators/rails/authentication/templates/app/controllers/passwords_controller.rb.tt
@@ -1,10 +1,13 @@
 class PasswordsController < ApplicationController
   allow_unauthenticated_access
   before_action :set_user_by_token, only: %i[ edit update ]
+  <% if defined?(ActionMailer::Railtie) -%>
   rate_limit to: 10, within: 3.minutes, only: :create, with: -> { redirect_to new_password_path, alert: "Try again later." }
+  <% end -%>
 
   def new
   end
+  <% if defined?(ActionMailer::Railtie) -%>
 
   def create
     if user = User.find_by(email_address: params[:email_address])
@@ -13,6 +16,7 @@ class PasswordsController < ApplicationController
 
     redirect_to new_session_path, notice: "Password reset instructions sent (if user with that email address exists)."
   end
+  <% end -%>
 
   def edit
   end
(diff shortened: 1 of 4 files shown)


railties/lib/rails/generators/rails/authentication/templates/app/controllers/passwords_controller.rb.tt
```diff
@@ -7,11 +7,13 @@ class PasswordsController < ApplicationController
   end
 
   def create
+  <% if defined?(ActionMailer::Railtie) -%>
```
byroot, line 10: Shouldn't this action just not exist if ActionMailer is skipped? It's basically a noop.

The route would probably need to be removed to?

ramimassoud: Ah yes, of course! I'll take a crack at adjusting the routes and removing `#create` and `#new`.

One thing I noticed is the way the routes are set up would also generate `#show`, `#index`, and `#destory`. Would it make sense to always have an `only:` in there? Then if the mailer is defined it would be `only: [:new, :create, :edit, :update]`, otherwise `only: [:edit, :update]`

https://github.com/rails/rails/blob/7cb69a8fd004c94b9266f2fdb2a833d15c09192b/railties/lib/rails/generators/rails/authentication/authentication_generator.rb#L40-L43

skipkayhil: I think adding `only:` to generated routes would not be accepted: https://github.com/rails/rails/pull/52369#discussion_r1699079086

ramimassoud: @skipkayhil Thank you for the heads up! 🙏 

@byroot I think I addressed the feedback fully. Let me know if you think any more changes would be appropriate here.

PR #54692, opened 2025-03-05
RuboCop: permit modern Ruby syntax in new apps

jeremy:
Generate apps with `.rubocop.yml` that permits modern Ruby syntax.

Specifies `TargetRubyVersion` to target the current Ruby version rather than the conservative Ruby 2.7 default.


railties/lib/rails/generators/rails/plugin/templates/rubocop.yml.tt
```diff
@@ -1,6 +1,9 @@
 # Omakase Ruby styling for Rails
 inherit_gem: { rubocop-rails-omakase: rubocop.yml }
 
+AllCops:
+  TargetRubyVersion: <%= ENV["RBENV_VERSION"] || ENV["rvm_ruby_string"] || RUBY_ENGINE_VERSION %>
```
skipkayhil, line 5: Should we add this as `required_ruby_version` in the plugin's gemspec since Rubocop may suggest using newer syntax?

railties/lib/rails/generators/rails/app/templates/rubocop.yml.tt
```diff
@@ -1,6 +1,9 @@
 # Omakase Ruby styling for Rails
 inherit_gem: { rubocop-rails-omakase: rubocop.yml }
 
+AllCops:
+  TargetRubyVersion: <%= ENV["RBENV_VERSION"] || ENV["rvm_ruby_string"] || RUBY_ENGINE_VERSION %>
```
skipkayhil, line 5: Rubocop should be using the value in `.ruby-version` by default, which we already generate for apps: https://github.com/rails/rails/blob/d8f68f9d558be9420d9432dcfcb5b80a2b3cb161/railties/lib/rails/generators/rails/app/templates/ruby-version.tt

jeremy: Indeed it should! I'll go back and double-check why I was seeing 2.7 syntax enforcement.

jeremy: Yep, reverting. It was due to not having a `.ruby-version` in generated plugins' `test/dummy`. See `Rails::PluginBuilder#test_dummy_clean`.

PR #54693, opened 2025-03-05
Structured CI with bin/ci

jeremy:
Introduce `bin/ci` to standardize CI workflows based on a new DSL for declaring workflow steps in `config/ci.rb`.

`bin/ci` runs your all tests, linters, and security scanners. And it optionally signs off on your work by giving your PR a green status.

This is your final check after completing a major chunk of work, before you merge your PR, etc. For everyday dev/test, use targeted `rails t …`.

Use in tandem with [gh-signoff](https://github.com/basecamp/gh-signoff) to set a green PR status when `bin/ci` completes successfully.

Example of a `config/ci.rb` definition that can be run with `bin/ci` locally or in the cloud:

```ruby
CI.run do
  step "Setup", "bin/setup --skip-server"

  step "Style: Ruby", "bin/rubocop"

  step "Security: Brakeman code analysis", "bin/brakeman", "--quiet", "--no-pager", "--exit-on-warn", "--exit-on-error"
  step "Security: Importmap vulnerability audit", "bin/importmap", "audit"

  step "Tests: Rails", "bin/rails", "test", "test:system"
  step "Tests: Seeds", "env RAILS_ENV=test bin/rails db:seed:replant"

  if success?
    step "Signoff: All systems go. Ready for merge and deploy.", "gh signoff"
  else
    failure "Signoff: Failed. Do not merge or deploy.", "Fix the issues and try again."
  end
end
```

Example output:

```
 ./bin/ci
Continuous Integration
Running tests, style checks, and security audits
                                                                                                                                                                                                            
                                                                                                                                                                                                            
Setup
bin/setup --skip-server
                                                                                                                                                                                                            
== Installing ruby ==
mise all runtimes are installed
                                                                                                                                                                                                            
== Installing dependencies ==
The Gemfile's dependencies are satisfied
                                                                                                                                                                                                            
== Preparing database ==
                                                                                                                                                                                                            
== Removing old logs and tempfiles ==
                                                                                                                                                                                                            
== Now start developing with bin/dev
                                                                                                                                                                                                            
✅ Setup passed in 3.33s
                                                                                                                                                                                                            
                                                                                                                                                                                                            
Style: Ruby
bin/rubocop
                                                                                                                                                                                                            
Inspecting 563 files
...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
                                                                                                                                                                                                            
563 files inspected, no offenses detected
                                                                                                                                                                                                            
✅ Style: Ruby passed in 1.07s
                                                                                                                                                                                                            
                                                                                                                                                                                                            
Security: Brakeman code analysis
bin/brakeman --quiet --no-pager --exit-on-warn --exit-on-error
                                                                                                                                                                                                            
                                                                                                                                                                                                            
== Brakeman Report ==
                                                                                                                                                                                                            
Application Path: /home/dhh/Work/basecamp/know_it_all
Rails Version: 8.1.0.alpha
Brakeman Version: 6.2.2
Scan Date: 2025-03-06 12:16:47 +0100
Duration: 1.273865987 seconds
Checks Run: BasicAuth, BasicAuthTimingAttack, CSRFTokenForgeryCVE, ContentTag, CookieSerialization, CreateWith, CrossSiteScripting, DefaultRoutes, Deserialize, DetailedExceptions, DigestDoS, DynamicFinders, EOLRails, EOLRuby, EscapeFunction, Evaluation, Execute, FileAccess, FileDisclosure, FilterSkipping, ForgerySetting, HeaderDoS, I18nXSS, JRubyXML, JSONEncoding, JSONEntityEscape, JSONParsing, LinkTo, LinkToHref, MailTo, MassAssignment, MimeTypeDoS, ModelAttrAccessible, ModelAttributes, ModelSerialize, NestedAttributes, NestedAttributesBypass, NumberToCurrency, PageCachingCVE, Pathname, PermitAttributes, QuoteTableName, Ransack, Redirect, RegexDoS, Render, RenderDoS, RenderInline, ResponseSplitting, RouteDoS, SQL, SQLCVEs, SSLVerify, SafeBufferManipulation, SanitizeConfigCve, SanitizeMethods, SelectTag, SelectVulnerability, Send, SendFile, SessionManipulation, SessionSettings, SimpleFormat, SingleQuotes, SkipBeforeFilter, SprocketsPathTraversal, StripTags, SymbolDoSCVE, TemplateInjection, TranslateBug, UnsafeReflection, UnsafeReflectionMethods, ValidationRegex, VerbConfusion, WeakRSAKey, WithoutProtection, XMLDoS, YAMLParsing
                                                                                                                                                                                                            
== Overview ==
                                                                                                                                                                                                            
Controllers: 58
Models: 63
Templates: 181
Errors: 0
Security Warnings: 0
Ignored Warnings: 3
                                                                                                                                                                                                            
== Warning Types ==
                                                                                                                                                                                                            
                                                                                                                                                                                                            
No warnings found
                                                                                                                                                                                                            
                                                                                                                                                                                                            
✅ Security: Brakeman code analysis passed in 1.61s
                                                                                                                                                                                                            
                                                                                                                                                                                                            
Security: Importmap vulnerability audit
bin/importmap audit
                                                                                                                                                                                                            
No vulnerable packages found
                                                                                                                                                                                                            
✅ Security: Importmap vulnerability audit passed in 1.40s
                                                                                                                                                                                                            
                                                                                                                                                                                                            
Tests: Rails
bin/rails test test:system
                                                                                                                                                                                                            
Running 618 tests in parallel using 32 processes
Run options: --seed 52099
                                                                                                                                                                                                            
# Running:
                                                                                                                                                                                                            
..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
                                                                                                                                                                                                            
Finished in 4.164330s, 148.4032 runs/s, 591.2115 assertions/s.
618 runs, 2462 assertions, 0 failures, 0 errors, 0 skips
                                                                                                                                                                                                            
✅ Tests: Rails passed in 6.64s
                                                                                                                                                                                                                                                                                                                                                                                                                     
✅ Continuous Integration passed in 14.42s
```


railties/lib/rails/generators/rails/app/templates/bin/ci.tt
```diff
@@ -0,0 +1,77 @@
...
+<% end -%>
+<% unless options.skip_docker? -%>
+  step "Style: Dockerfile", "docker run --rm -i -v #{ENV['PWD']}/.hadolint.yaml:/hadolint.yaml -v #{ENV['PWD']}/Dockerfile:/Dockerfile hadolint/hadolint hadolint --config /hadolint.yaml /Dockerfile"
+<% end -%>
+
+<% unless options.skip_brakeman? -%>
+  step "Security: Brakeman code analysis", "bin/brakeman --quiet --no-pager --exit-on-warn --exit-on-error"
+<% end -%>
+<% if using_node? -%>
+  step "Security: Yarn vulnerability audit", "yarn audit"
```
brunoprietog, line 68: We could use `bin/impormap audit` as well.

railties/lib/rails/generators/rails/app/templates/bin/ci.tt
```diff
@@ -0,0 +1,77 @@
...
+  error: "\033[1;31m",    # Red
+  success: "\033[1;32m"   # Green
+}
+
+
+echo :banner, "🚀 Local CI"
+echo :subtitle, "Running tests, style checks, and security audits"
+
+report "Local CI" do
+  step "Setup", "bin/setup"
```
brunoprietog, line 55: The default Rails setup script starts the application at the end of the script execution. I'd be in favor of not doing that so that this can run cleanly.

dhh:
@byroot Next step would be to have the default GitHub actions run this command instead of their own workflow setup. But I don't agree that the point of CI is to run things in a cloud setup. For me, it's to build confidence that your deploy is going to be a success, that your code base stays neatly linted, and that there are no pending security issues.

Fine for a team to make the trade-off that they want to setup a cloud machinery to do that, but at least just as fine to decide that running things locally is the way to go.

The problem you point out about different versions of postgresql is something we should actually tackle next. At 37signals, we run such dependencies in Docker exactly for that reason. And you can do that, while still having your code run a local Ruby, and get the best of all worlds (speed, version locking, and high confidence of a successful deploy).

byroot:
> I don't agree that the point of CI is to run things in a cloud setup. For me

I'm not saying it's not all of these things too. And note that I'm not saying it should be in a cloud setup. I'm saying it should be an environment you are confident is as close to production as reasonably possible. Same datastore version, same ruby version, same operating system (yes there are numerous difference between macOS and Linux) etc etc. Which you can perfectly achieve locally with some specific tools (e.g. docker), but we can't really assume that.

I'm also not saying this script shouldn't exist. If teams are OK with that tradeoff, good for them. All I'm saying is that we should make it clear what the tradeoff is.

activesupport/lib/active_support/continuous_integration.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def initialize
+    end
+
+    def step(title, command)
+      echo :title, "\n\n#{title}"
+      echo :subtitle, "#{command}\n"
+
+      report(title) { system command }
```
byroot, line 22:
```suggestion
      report(title) { system { "CI" => "true" }, command }
```

If the goal is to behave like a CI, setting this env variable is preferable as some tools will behave differently, including Rails' default environment:  https://github.com/rails/rails/blob/c218fb6cd2bdd67771aac5bfce8097e43c05a26c/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt#L16

byroot:
> Next step would be to have the default GitHub actions run this command instead of their own workflow setup.

The issue with that is that you can't easily parallelize such a script. With most CI systems, including GitHub Actions, you can define the linter and tests each in their own job and they run in parallel. 

So if you do what you suggests, you make GitHub Actions worse.

dhh:
Got a plan cooking for parallelizing steps as well! Although of course possible that whatever instance GH gives you won't have more than one vCPU, so possible that doesn't help. But let's have a look.

dhh:
Yeah, there is a weighing here of not declaring the CI steps in multiple locations, having a CI default that's service agnostic, and then providing something that works well for most people out of the box.

Going to experiment with what a default GH workflow that just calls bin/ci looks like. But if it doesn't feel right, I'm not religious about it. I would, however, prefer to have a service-agnostic CI setup.

byroot:
Yep. The somewhat obvious thing to do to satisfy all your requirements is to make your `CI` class not directly execute, but store the metadata of what each task is, so that you can then hook it into whatever CI system to be `bin/ci sub-task`.

The downside being that it obviously complexify that CI class.

.rubocop.yml
```diff
@@ -387,9 +387,6 @@ Performance/StringInclude:
 Minitest/AssertNil:
   Enabled: true
 
-Minitest/AssertPredicate:
-  Enabled: true
```
byroot, line 391: The reason for this rule is that it gives you much better error message in case of assertion failures, making debugging much easier:

e.g. 

```
Failure:
ContinuousIntegrationTest#test_report_with_only_successful_steps_combined_gives_success [test/continuous_integration_test.rb:30]:
Expected true to be nil or false
```

vs

```
Failure:
ContinuousIntegrationTest#test_report_with_only_successful_steps_combined_gives_success [test/continuous_integration_test.rb:30]:
Expected #<ActiveSupport::ContinuousIntegration:0x000000011e281b20 @results=[true, true]> to not be success?.
```

activesupport/lib/active_support/continuous_integration.rb
```diff
@@ -0,0 +1,145 @@
...
+    def self.run(title = "Continuous Integration", subtitle = "Running tests, style checks, and security audits", &block)
+      new.tap do |ci|
+        ENV["CI"] = "true"
+        ci.heading title, subtitle, padding: false
+        ci.report(title, &block)
+        abort unless ci.success?
+      end
+    end
+
+    def initialize(&block)
```
sanjioh, line 64: Hi, here `block` is not used, maybe it’s worth removing it from the signature.

.rubocop.yml
dhh: I don't think that carries the weight for how ugly that method is. To fix what's wrong, you still have to look at the line, and when you do, you'll see what the predicate is. Appreciate that the output is marginally better, but to me that doesn't outweigh the aesthetic regression.

activesupport/lib/active_support/continuous_integration.rb
dhh: Good catch. Removed in f98c0188ae2e.

railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
```diff
@@ -0,0 +1,33 @@
...
+<% if options[:javascript] == "importmap" && !options[:api] && !options[:skip_javascript] -%>
+  step "Security: Importmap vulnerability audit", "bin/importmap audit"
+<% end -%>
+<% unless options.skip_brakeman? -%>
+  step "Security: Brakeman code analysis", "bin/brakeman --quiet --no-pager --exit-on-warn --exit-on-error"
+<% end -%>
+<% if options[:api] || options[:skip_system_test] -%>
+  step "Tests: Rails", "bin/rails test"
+<% else %>
+  step "Tests: Rails", "bin/rails test test:system"
```
kylekeesling, line 22: it appears this might be incorrect - running it as is, Rails attempts to test a file named 'test:system' rather than to run both normal tests and system tests, resulting in the following error:

> /.rbenv/versions/3.4.2/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': cannot load such file -- /path/to/rails-project/test:system (LoadError)

Does it make sense to create separate steps for each or to just concat the call? (`rails test && rails test:system`)

byroot: > cannot load such file -- /path/to/rails-project/test:system (LoadError)

This might be a fallout from https://github.com/rails/rails/pull/54647 (FYI: @Edouard-chin )

kylekeesling: I also just realized there is a command to run all tests `bin/rails test:all' maybe that's the best option?

dhh: No, I think test:all mixes all tests together. We want to run all the regular tests first, then the system tests. But we can also just split it into two separate steps?

Edouard-chin: We reverted https://github.com/rails/rails/pull/54647. I think it's fine to leave the `bin/rails test test:system` as is. I'll try to find another solution for #54647.

Sorry for the issue

rmacklin: It still might be better to have separate steps if we're thinking about the proposal in https://github.com/rails/rails/pull/54693#issuecomment-2703598874:

> "bin/ci --step rubocop" and "bin/ci --step test" being separate and called from a GH workflow file would be pretty neat

Reason being: folks could use a lighter weight image (with no browsers installed) to run non-system tests.

Edouard-chin: > We reverted https://github.com/rails/rails/pull/54647. I think it's fine to leave the bin/rails test test:system

Actually scratch that. The behaviour of `bin/rails test test:system` just runs the the `test` command and completely ignore `test:system`. Probably best to run it in two steps as DHH mentioned

MiroslavCsonka:
For anyone looking to actually skip their Github build (after local signoff), checkout [this gist](https://gist.github.com/MiroslavCsonka/73dc6e988a9883a5e7e357ec48daa00f) as a starting point.

PR #54695, opened 2025-03-05
Add a default bin/bundle-audit configuration

dhh:
Finding known security issues in the application's Gemfile should be part of every CI flow.


jeremy (approved):

jeremy (commented):
Would include in the GitHub CI action as well.

byroot:
> Would include in the GitHub CI action as well.

It's debatable, things like bundle-audit that can start failing without any change on your side aren't ideal as part of CI.

Should you know that you have an unfixed CVE? Yes absolutely.

Does it help in anyway to prevent you from merging anything else until this is resolved? I don't think so.

dhh:
I think the build should break until you acknowledge your response. Which is either mitigation or knowingly ignoring it (which is why this expose a config/bundler-audit.yml file to do that). Shipping the app with a known, unaddressed security issue is imo same as shipping it with failing tests.

byroot:
> Shipping the app with a known, unaddressed security issue is imo same as shipping it with failing tests.

I don't think so. The security issue is pre-existing to you shipping. Whether you ship or not it's there.

dhh:
But the point of this is to be forced to deal with the situation explicitly. Just like a failing tests may or may not be something you actually want to address. You can either fix the test or you can skip it. But CI is there to ensure you did either consciously.

byroot:
I totally get your point, but the point of CI is to tell me if there is something wrong with changes I made in my PR, not with the overall system.

If I have 5 devs each working on their feature concurrently, failing all their builds doesn't make sense.

The ideal workflow is to receive an async notification that action should be taken immediately.

Now I get that, for a fresh Rails app, GItHub Actions is kind of the only system you can assume, so there isn't any better places to put that, and perhaps it's better than nothing. I'm just saying ideally this isn't what a CI is supposed to do.

Also, GitHub tells you about CVEs it can detect, I'm not super up to date on that, but I wonder if `bundle-audit` adds anything over what GitHub already does for you.

dhh:
I don't want to tie the entire Rails stance on CI to GitHub features. That's the other reason that we're pursuing bin/ci. It's already a bit of an uneasy thing to have a specific vendor file in a default Rails installation, like we do with the GH CI workflow file. Not interested in pushing that further.

dhh:
Rails should have a complete answer to all elements of CI in the box, without requiring any specific cloud vendor. If you want to host your stuff on your own git setup, that should be fine. Gitlab should be fine. GitHub should be fine. I think we can have our cake and eat it too, though. Push the bulk of the work towards bin/ci, and then also have a default setup that makes it easy to use this on GitHub.

byroot:
> Rails should have a complete answer to all elements of CI in the box

Well, my point is that this isn't a CI task. But that's fine, it's easy enough to remove.

n-rodriguez:
@dhh very nice addition! But `bundler-audit` is licensed under GPL v3 which can be an issue for some users (like me).

```sh
Dependencies that need approval:
bundler-audit, 0.9.2, GPL-3.0-or-later
```

Thank you!

skipkayhil:
> @dhh very nice addition! But `bundler-audit` is licensed under GPL v3 which can be an issue for some users (like me).
> 
> ```shell
> Dependencies that need approval:
> bundler-audit, 0.9.2, GPL-3.0-or-later
> ```
> 
> Thank you!

Is this an issue because it's in your gemfile? Maybe `bin/bundler-audit` could call `gem exec bundler-audit` if so

n-rodriguez:
> Is this an issue because it's in your gemfile?

Yes. Note that I'm not a lawyer and it's a very old/sensitive/complex question. From what I read it should not be an issue (https://github.com/rubysec/bundler-audit/issues/81#issuecomment-51815922) since it's concerning a SaaS application.

So I don't know 😕

dhh:
Where is that approval notice coming? There's no problem using a GPL tool in a commercial project. But if you modify it, you have to release those mods. And if you're in doubt on distribution, you can just remove it from the dist bundle.

n-rodriguez:
> Where is that approval notice coming?

A CI job. 

> There's no problem using a GPL tool in a commercial project. But if you modify it, you have to release those mods. And if you're in doubt on distribution, you can just remove it from the dist bundle.

Thank you!

jgrau:
FWIW and somewhat related to the comments by @byroot: we run `bundle-audit` nightly as well as on pull requests - but only if those pull requests have changes in `Gemfile.lock`:

Example:

```yaml
# .github/workflows/bundle-audit.yml

on:
  schedule:
    - cron: '0 0 * * *'

  pull_request:
    paths:
      - .github/workflows/bundle-audit.yml # run the workflow if the workflow definition changes
      - Gemfile.lock

...
```

1. By running it nightly we get notified if there's a vulnerability in our current packages even if we don't make any changes/prs.
2. By only running on Gemfile changes we only fail the PRs that are introducing a known vulnerability (that's not 100% true as a PR could have a change in the Gemfile which makes bundle-audit run and report on a CVE unrelated to the change but still...)

ghiculescu:
@jgrau for the nightly runs, where do you output the results to if there's issues found?

jgrau:
> @jgrau for the nightly runs, where do you output the results to if there's issues found?

I'll get an email notification when the build fails and the actual output from bundle-audit can be found in the github action run :)

julienanne:
> Since this is in CI, shouldn't the audit call use `--update` to update the advisory DB?

I agree, this is problematic. For example, in my Rails project, `bin/bundler-audit` doesn’t update the advisory DB, so no errors are returned. I added the advisory DB update in `config/ci.rb`, and that fixed the issue. I think it would be better if `bin/bundler-audit` updated it by default, right?
bundler-audit-rails-issue

If it is relevant, should I open a PR or an issue, or should this be reported as a security disclosure on https://hackerone.com/rails?type=team ?

byroot:
> If it is relevant, should I open a PR or an issue, or should this be reported as a security disclosure on https://hackerone.com/rails?type=team ?

Just a PR is fine. This is not a security issue (and if it was you would have just made it public anyway, so no need for HackerOne 😄 )

julienanne:
> > If it is relevant, should I open a PR or an issue, or should this be reported as a security disclosure on https://hackerone.com/rails?type=team ?
> 
> Just a PR is fine. This is not a security issue (and if it was you would have just made it public anyway, so no need for HackerOne 😄 )

Thanks for your reply and yes it is not a good idea to ask that here, sorry.

The PR https://github.com/rails/rails/pull/55913

cc @justinko

PR #54696, opened 2025-03-05
[ci skip]Fix the incorrect version where `allow_browser` was added in the guide

willnet:
`allow_browser` was added in Rails 7.2 and is set by default in `ApplicationController` when running `rails new`.

ref: https://github.com/rails/rails/commit/e3da4fc53d5da7be1d5b93d5659e77c8daddccea#diff-f355d10a96623990e497d37948436731adb5dc46d9f447d00f1a2590811ac606R4


Edouard-chin:
Thank you!

PR #54698, opened 2025-03-05
perf: prevent unnecesary query on schema_search

jvlara:
### Motivation / Background

While working on a project, I noticed that the PostgreSQL adapter executes an unnecessary query when setting schema_search_path to the same schema that the application is already using.

### Details

This PR optimizes the schema_search_path setter by skipping the query if the new schema matches the current one. This avoids redundant database calls and improves efficiency.
### Checklist

* [x] Tests are added or updated if you fix a bug or add a feature. This is not a bug, only a micro optimization

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -280,6 +280,7 @@ module ActiveRecord
         #
         # This should be not be called manually but set in database.yml.
         def schema_search_path=(schema_csv)
+          return if schema_csv == @schema_search_path
           if schema_csv
             internal_execute("SET search_path TO #{schema_csv}")
             @schema_search_path = schema_csv


activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -224,6 +224,7 @@ def drop_schema(schema_name, **options)
         #
         # This should be not be called manually but set in database.yml.
         def schema_search_path=(schema_csv)
+          return if schema_csv == schema_search_path
```
byroot, line 227: Wouldn't it make more sense to compare with `@schema_search_path` to avoid the `SHOW` query as well?

jvlara: I don't think so. It's memoized, so the `SHOW` query won't appear if the variable is already filled

byroot: Yes, but if you are setting `search_schema_cache = ` *before* `search_schema_cache` was ever accessed, you'll cause an extra useless query.

jvlara: Ah, now I understand. You're right! I'll fix it. Thanks!

PR #54699, opened 2025-03-06
Fix `ActiveRecord::Result#dup` consistency

byroot:
`@column_types = column_types.dup` is incorrect since https://github.com/rails/rails/pull/54594 because the `column_types` method is no longer a simple attribute reader of `@column_type`.

### DIFF
--- a/activerecord/lib/active_record/result.rb
+++ b/activerecord/lib/active_record/result.rb
@@ -208,7 +208,6 @@ module ActiveRecord
 
     def initialize_copy(other)
       @rows = rows.dup
-      @column_types = column_types.dup
       @hash_rows    = nil
     end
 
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/result.rb
```diff
@@ -239,6 +238,11 @@ def indexed_rows # :nodoc:
       end
     end
 
+    protected
+      def raw_column_types
```
kamipo, line 242: Where is this method intended to be called from?

matthewd, line 242: Leftover from a previous iteration?

byroot: Yes, -_-.

Sorry, I haven't finished my coffee, should have taken the time to review my diff.

byroot: Fixed in 7c7500f496d8a5d11d0b6e610e1b254ec34edf5a, thanks both for catching that.

PR #54702, opened 2025-03-06
Fix `ActionMailer::Base.default` docs [skip-ci]

### LINKED ISSUE #54700: `config.action_mailer.default` doesn't work
The documentation of ActionMailer::Base.default [says](https://github.com/rails/rails/blob/v8.0.1/actionmailer/lib/action_mailer/base.rb#L577-L582):

```ruby
# Sets the defaults through app configuration:
#     config.action_mailer.default(from: "no-reply@example.org")
# Aliased by ::default_options=
```

So I think I should be able to write that line in a Rails::Application subclass. But when I do it has no effect.

On the other hand, `config.action_mailer.default_options = { from: "no-reply@example.org" }` works as expected.

I suspect the ActionMailer railtie is responsible.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  #gem "rails", github: "rails/rails", branch: "main"
end

require "action_mailer/railtie"

  config.action_mailer.default(from: "no-reply@example.org") # BROKEN
  #config.action_mailer.default_options = { from: "no-reply@example.org" } # WORKS
end

Rails.application.initialize!

puts "ActionMailer::Base.default_params[:from] = #{ActionMailer::Base.default_params[:from].inspect}"
mail = ActionMailer::Base.mail(to: "test@example.com", subject: "Test", body: "Test")
puts "mail.from = #{mail.from.inspect}"
```

### Expected behavior
```
ActionMailer::Base.default_params[:from] = "no-reply@example.org"
mail.from = ["no-reply@example.org"]
```

### Actual behavior
```
ActionMailer::Base.default_params[:from] = nil
mail.from = nil
```

Uncommenting the line marked "WORKS" produces the expected output.

### System configuration

**Rails version**: 8.0.1, 8.1.0.alpha; probably affects all older versions too

**Ruby version**: 3.4.2

fatkodima:
Fixes #54700.

`config.action_mailer.default(..)` example is wrong, because `action_mailer` is an instance of `OrderedOptions` and this is calling a `Hash#default(..)` method, which does not make sense.

### DIFF
--- a/actionmailer/lib/action_mailer/base.rb
+++ b/actionmailer/lib/action_mailer/base.rb
@@ -574,18 +574,13 @@ module ActionMailer
       attr_writer :mailer_name
       alias :controller_path :mailer_name
 
-      # Sets the defaults through app configuration:
+      # Allows to set defaults through app configuration:
       #
-      #     config.action_mailer.default(from: "no-reply@example.org")
-      #
-      # Aliased by ::default_options=
+      #    config.action_mailer.default_options = { from: "no-reply@example.org" }
       def default(value = nil)
         self.default_params = default_params.merge(value).freeze if value
         default_params
       end
-      # Allows to set defaults through app configuration:
-      #
-      #    config.action_mailer.default_options = { from: "no-reply@example.org" }
       alias :default_options= :default
 
       # Wraps an email delivery inside of ActiveSupport::Notifications instrumentation.


Edouard-chin (approved):

skipkayhil (approved):

fatkodima:
@skipkayhil I wonder if such changes worth backporting?

skipkayhil:
> @skipkayhil I wonder if such changes worth backporting?

I think backporting is a good idea since the current text is just very wrong

PR #54703, opened 2025-03-06
GitHub Actions improvements

byroot:
Fixing a few GitHub Actions pet peeves for the auto-generated CI config.

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
@@ -134,12 +134,81 @@ jobs:
           <%- end -%>
           # RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
           # REDIS_URL: redis://localhost:6379/0
-        <%- if options[:api] || options[:skip_system_test] -%>
         run: bin/rails db:test:prepare test
-        <%- else -%>
+  <%- unless options[:api] || options[:skip_system_test] -%>
+
+  system-test:
+    runs-on: ubuntu-latest
+
+    <%- if options[:database] == "sqlite3" -%>
+    # services:
+    #  redis:
+    #    image: valkey/valkey:8
+    #    ports:
+    #      - 6379:6379
+    #    options: --health-cmd "redis-cli ping" --health-interval 10s --health-timeout 5s --health-retries 5
+    <%- else -%>
+    services:
+      <%- if options[:database] == "mysql" || options[:database] == "trilogy" -%>
+      mysql:
+        image: mysql
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: true
+        ports:
+          - 3306:3306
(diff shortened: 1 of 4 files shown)


railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
```diff
@@ -9,6 +9,8 @@ jobs:
 <%- unless skip_brakeman? -%>
   scan_ruby:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
byroot, line 13: Hum, actually fail fast only applies to matrixes, so I guess it's only really an annoyance for gems. I guess this isn't needed.

byroot:
I also wanted to setup persistent cache for rubocop, as when a project grows a bit it makes a major speed difference, but I can't seem to make `actions/cache` work...

byroot:
Why is rubocop complaining on my PR, I haven't touched it???

Earlopain:
> Why is rubocop complaining on my PR, I haven't touched it???

https://github.com/rubocop/rubocop/issues/13950 Apologies, will be fixed in the next release. The hash used an array as the cache key, and I suspect ruby sometimes rehashes it (not sure what else would be the cause otherwise). When that happens, you get this error. I guess some code shifted around that made this more likely now, since this bug has been introduced more than a year ago already.

byroot:
Ah thank you for the heads up.

byroot:
I thought I was taking crazy pills for a minute.

byroot:
> I also wanted to setup persistent cache for rubocop

If someone reads this and is interested, here's what I tried: https://github.com/byroot/test_app/pull/1/commits/9c9f449a0f1871b1a5d42d70b92bf06c4c78f83b

For some reason it appear to work, but then on any subsequent build the save step soft fail with:

```
Failed to save: Unable to reserve cache with key rubocop-v2, another job may be creating this cache.
```

But perhaps I don't get how GitHub caches are supposed to work.

Earlopain:
I'm not quite sure if a cache for rubocop is easily possible in (GitHub) CI. The cache, once uploaded, is immutable for a given cache key.

Usually the cache key contains exactly by what it can change. For `setup-ruby`, that would among other things be the gemfile, so if any gem changes _all_ gems are installed again on new runs. For RuboCop you'd need to scope it by all ruby files, which pretty much defeats the purpose.

Earlopain:
A more simpler approach would be to only run it on files that actually changed ala https://robertfaldo.medium.com/commands-to-run-rubocop-and-specs-you-changed-in-your-branch-e6d2f2e4110b, with the obvious caviat that it doesn't run on all files anymore.

byroot:
> The cache, once uploaded, is immutable for a given cache key.

Ah, I see. Well that a pretty useless cache primitive... I mean, it's useful for some cases, but very limited.

Earlopain:
Yeah, it's very hard to get right. I don't think I've ever manually set it up myself. Artifacts exists but their main purpose is to share data between jobs of the same workflow. Anything outside of that is also very janky.

PR #54705, opened 2025-03-06
Don't rebuild the middleware stack when using `with_routing`

### LINKED ISSUE #54595: Race condition in with_routing integration test helper
### Steps to reproduce

Hoping to get some guidance on making a consistent reproducible example. The error I'm observing (`No Failure App provided`) is occurring during `bin/rails test` and originates in the [warden gem](https://github.com/wardencommunity/warden/blob/67f5ba6baaa7564ec79afef02cf3a4d0f7d312e5/lib/warden/manager.rb#L145). This only occurs about 20% of the time, i.e., running `bin/rails test` will complete without errors 4/5 times. The error only occurs when using the `with_routing` helper, e.g., 

```ruby
class ErrorPagesConcernTest < ActionDispatch::IntegrationTest
  with_routing do |routes|
    routes.draw do
      %w[action_not_found invalid_authenticity_token name_error no_method_error record_not_found routing_error standard_error].each do |action|
        get "/default/#{action}", to: "error_pages_test##{action}"
        get "/rescued/#{action}", to: "rescued_error_pages_test##{action}"
      end
    end
  end
  
  ...
end
```

However, when I remove the `with_routing` helper and put the testing routes directly in `routes.rb`, the test suite runs without error 100% of the time.

### Expected behavior

The results of a test suit should be consistent when using the `with_routing` helper.

### Actual behavior

Using the `with_routing` helper causes inconsistent results in tests.

### System configuration

**Rails version**: 8.0.1

**Ruby version**: 3.2.6

Edouard-chin:
### Motivation / Background

Fix #54595

#### Problem

When using the `with_routing` helper method in an integration test, we rebuild the middleware stack so that the entrypoint of our new stack is the new RouteSet object.

By rebuilding the middleware stack from scratch, we may end un-configuring a middleware from an application. We also break the contract of the Rails initialization process.

For instance, if you have a middleware that needs to be configured, after the Rails routes are loaded, this is what it would look like:

```ruby
# config/application.rb

warden_config = nil

config.middleware.use Warden::Manager do |config|
  # The Warden middleware yields its config when the middleware is
  # instantiated. We can't configure warden yet as the routes
  # are not yet loaded.

  warden_config = config
end

initializer :configure_warden do
  ActiveSupport.on_load(:after_routes_loaded) do
    # The routes are now loaded, we can configure warden
  end
end
```

This snippet follows the order of the documented Rails initialization process where first the middleware stack is built and then the routes gets loaded.

--------------------

By rebuilding the middleware stack, we unconfigure what was done in the initializer.

### Solution

Since rebuilding the middleware stack has some caveats, my idea is to modify the middleware stack original entry point and redefine `call` so that it calls our new RouteSet middleware.
We have a reference to the original RouteSet middleware thanks to the `application.routes` method.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/testing/assertions/routing.rb
+++ b/actionpack/lib/action_dispatch/testing/assertions/routing.rb
@@ -5,6 +5,7 @@
 require "uri"
 require "active_support/core_ext/hash/indifferent_access"
 require "active_support/core_ext/string/access"
+require "active_support/core_ext/module/redefine_method"
 require "action_controller/metal/exceptions"
 
 module ActionDispatch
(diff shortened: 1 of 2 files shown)


quixoten:
Verified my test suite passes against this branch (and devise 4.9.4)

Edouard-chin:
Thanks for reviewing! Also thank you @quixoten for helping debug the issue !

PR #54710, opened 2025-03-07
[7.1] Disconnect if `configure_connection` failed

byroot:
Fix: https://github.com/rails/rails/issues/51780
Backport: https://github.com/rails/rails/pull/54711

There are a number of scenarios in which one of the queries performed as part of `configure_connection` fails, and leave the `Adapter` in an half initialized state.

Up to Rails 7.0, this could likely happen during a call to `reconnect!`, so relatively rare. For the more classic initial connection, the `Adapter` instance would only be added to the pool on a successful connect.

But in 7.1 I made the connection fully lazy, meaning the `Adapter` instance is now added to the pool before attempting to connect.

This made this bug much more likely.

The solution is to disconnect if `configure_connection` raises an error. This way on a retry we'll reconnect and reconfigure from scratch.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix error handling during connection configuration.
+
+    Active Record wasn't properly handling errors during the connection configuration phase.
+    This could lead to a partially configured connection being used, resulting in various exceptions,
+    the most common being with the PostgreSQLAdapter raising `undefined method `key?' for nil`
+    or `TypeError: wrong argument type nil (expected PG::TypeMap)`.
+
+    *Jean Boussier*
+
 *   Fix a race condition in `ActiveRecord::Base#method_missing` when lazily defining attributes.
 
     If multiple thread were concurrently triggering attribute definition on the same model,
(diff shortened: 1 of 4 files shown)


matthewd (approved):

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+*   Fix error handling during connection configuration.
+
+    Active Record wasn't properly handling errors during the connection configuration phase.
+    This could lead to partially configured connection being used, resulting is various exception,
```
joshuay03, line 4:
```suggestion
    This could lead to a partially configured connection being used, resulting in various exceptions,
```

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+    the most common being with the PostgresAdapter raising `undefined method `key?' for nil`
```
joshuay03, line 5:
```suggestion
    the most common being with the PostgreSQLAdapter raising `undefined method `key?' for nil`
```

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+    of `TypeError: wrong argument type nil (expected PG::TypeMap)`.
```
joshuay03, line 6:
```suggestion
    or `TypeError: wrong argument type nil (expected PG::TypeMap)`.
```

I think this is what you meant here?

joshuay03 (commented):
Thank you, just some typos in the change-log entry.

PR #54713, opened 2025-03-07
Handle libpq db version 0 as connection failure in `PostgreSQLAdapter`

### LINKED ISSUE #54712: `PostgreSQLAdapter` caches database version of 0 returned by libpq
### Steps to reproduce

During a recent database fail-over we saw many errors like this:
```
RuntimeError: Your version of PostgreSQL (0) is too old. Active Record supports PostgreSQL >= 9.3.
```

We tracked it down to libpq returning 0 when asked to query the database version and when connection was bad:
- https://github.com/ged/ruby-pg/blob/62508cc14189653153c205a0c338066b90399834/ext/pg_connection.c#L869-L873
- https://github.com/postgres/postgres/blob/7f24c0274385ea3d4c797cc2fd60d9a55f3a8d74/src/interfaces/libpq/fe-connect.c#L7542-L7550

This then gets compared against the minimum version supported by `PostgreSQLAdapter`, and if incompatible an error is raised here: https://github.com/rails/rails/blob/3720ca0e29d2afaafea18258c7c41fcf909929d0/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L665-L669

This gets called at some point during `#configure_connection`.

Despite having had database fail-overs before, we've never seen this error until we upgraded to 7.2. I suspect it has something to do with https://github.com/rails/rails/pull/49415, possibly also in combination with https://github.com/rails/rails/pull/44570.

This error is fine as is, the issue is that the version of 0 gets cached and reused on subsequent `#configure_connection` calls i.e. during re-connections, which means the connection will never recover even if the database recovers and raw connections are successful: https://github.com/rails/rails/blob/3720ca0e29d2afaafea18258c7c41fcf909929d0/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L30-L32

I don't have a reproduction script, but I've been able to replicate with a test which I've paired with a patch: https://github.com/rails/rails/pull/54713

### Expected behavior

Following the error, re-connections to a healthy database with a valid version should succeed.

### Actual behavior

Following the error, re-connections to a healthy database with a valid version continue to raise the same error.

### System configuration

**Rails version**: 7.2.2.1

**Ruby version**: 3..3.7

joshuay03:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/54712

### Detail

In `PostgreSQLAdapter`, if the database version returned by pg/libpq is 0, we now raise `ActiveRecord::ConnectionFailed` which:
1. more accurately tells the user that a connection failure has happened instead of a version incompatibility
2. ensures the version isn't cached so that subsequent re-connection attempts may retrieve the correct version with a successful connection

**Note:** I think this should also be back-ported to 7.2.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -634,7 +634,11 @@ module ActiveRecord
       # Returns the version of the connected PostgreSQL server.
       def get_database_version # :nodoc:
         with_raw_connection do |conn|
-          conn.server_version
+          version = conn.server_version
+          if version == 0
+            raise ActiveRecord::ConnectionFailed, "Could not determine PostgreSQL version"
+          end
+          version
         end
       end
       alias :postgresql_version :database_version
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -95,6 +95,26 @@ module ActiveRecord
         end
       end
 
+      def test_reconnect_after_bad_connection_on_check_version
+        db_config = ActiveRecord::Base.configurations.configs_for(env_name: "arunit", name: "primary")
+        connection = ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.new(db_config.configuration_hash.merge(connection_retries: 0))
+        connection.connect!
+
+        # mimic a connection that hasn't checked and cached the server version yet i.e. without a raw_connection
+        connection.pool.instance_variable_set(:@server_version, nil)
+        connection.raw_connection.stub(:server_version, 0) do
+          error = assert_raises ActiveRecord::ConnectionFailed do
+            connection.reconnect!
+          end
+          assert_equal "Could not determine PostgreSQL version", error.message
+        end
+
+        # can reconnect after a bad connection
+        assert_nothing_raised do
+          connection.reconnect!
+        end
+      end
+
       def test_database_exists_returns_false_when_the_database_does_not_exist
         config = { database: "non_extant_database", adapter: "postgresql" }
         assert_not ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.database_exists?(config),


activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -95,6 +95,26 @@ def test_bad_connection_to_postgres_database
...
+          error = assert_raises ActiveRecord::ConnectionFailed do
+            connection.reconnect!
+          end
+          assert_equal "Could not determine PostgreSQL version", error.message
+        end
+
+        # can reconnect after a bad connection
+        assert_nothing_raised do
+          connection.reconnect!
+        end
```
joshuay03, line 115: Without the patch, if I update the assertion above to expect [this](https://github.com/rails/rails/blob/049d1af38d19968fc8e8648989b0fb8aae2caa71/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L665-L669) `RuntimeError` and the respective message, this assertion fails with:
Screenshot 2025-03-08 at 12 08 10 am

i.e. it cannot successfully reconnect.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -634,7 +634,11 @@ def use_insert_returning?
       # Returns the version of the connected PostgreSQL server.
       def get_database_version # :nodoc:
         with_raw_connection do |conn|
-          conn.server_version
+          version = conn.server_version
+          if version == 0
+            raise ActiveRecord::ConnectionFailed, "Could not determine PostgreSQL version"
```
joshuay03, line 639: I wasn't sure what the appropriate error here was. I went with this over `ActiveRecord::ConnectionNotEstablished` only because of this snippet of code indicating the former is used for libpq related errors: https://github.com/rails/rails/blob/049d1af38d19968fc8e8648989b0fb8aae2caa71/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L804-L814

Note that libpq may returr 0 either of there is no connection or if an existing connection is bad: https://github.com/postgres/postgres/blob/7f24c0274385ea3d4c797cc2fd60d9a55f3a8d74/src/interfaces/libpq/fe-connect.c#L7542-L7550

PR #54717, opened 2025-03-08
Make auth generator session controller test pass by default

jeromedalbert:
The authentication generator [session controller test](https://github.com/rails/rails/blob/182503ca27d4b940da925fb7374b69c4bcebd322/railties/lib/rails/generators/test_unit/authentication/templates/test/controllers/sessions_controller_test.rb) fails by default on new Rails apps, because they do not have a root route set up.

It is arguable whether the session controller test should pass by default on a newly created Rails app. If it should, some version of this PR could achieve that goal (FYI the password controller test passes by default). If it shouldn't, this PR can be closed.

### DIFF
--- a/railties/lib/rails/generators/test_unit/authentication/authentication_generator.rb
+++ b/railties/lib/rails/generators/test_unit/authentication/authentication_generator.rb
@@ -14,6 +14,14 @@ module TestUnit # :nodoc:
         template "test/controllers/sessions_controller_test.rb"
         template "test/controllers/passwords_controller_test.rb"
       end
+
+      def ensure_root_route
+        in_root do
+          if !File.read("config/routes.rb").match?(/^ *root /)
+            route 'root "fill-me"'
+          end
+        end
+      end
     end
   end
 end
--- a/railties/test/generators/authentication_generator_test.rb
+++ b/railties/test/generators/authentication_generator_test.rb
@@ -53,6 +53,7 @@ class AuthenticationGeneratorTest < Rails::Generators::TestCase
 
     assert_file "config/routes.rb" do |content|
       assert_match(/resource :session/, content)
+      assert_match(/root "fill-me"/, content)
     end
 
     assert_includes @rails_commands, "generate migration CreateUsers email_address:string!:uniq password_digest:string! --force"


railties/lib/rails/generators/test_unit/authentication/authentication_generator.rb
```diff
@@ -14,6 +14,14 @@ def create_controller_test_files
         template "test/controllers/sessions_controller_test.rb"
         template "test/controllers/passwords_controller_test.rb"
       end
+
+      def ensure_root_route
```
Edouard-chin, line 18: Should we modify the authentication generator instead? Perform a check whether `root_path` is defined, otherwise redirect to "/".

jeromedalbert: I had a go at your suggestion but I am not sure the resulting code would look good.

Not only does the generated `Authentication` module check the `root_url`:

https://github.com/rails/rails/blob/2318163b4b9e9604b557057c0465e2a5ef162401/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt#L37-L39

But the generated `SessionsControllerTest` also expects a `root_path`:

https://github.com/rails/rails/blob/2318163b4b9e9604b557057c0465e2a5ef162401/railties/lib/rails/generators/test_unit/authentication/templates/test/controllers/sessions_controller_test.rb#L14

So if I were to change `Authentication` to something like this:

```ruby
def after_authentication_url
  session.delete(:return_to_after_authenticating) || try(:root_url) || "/"
end
```

The generated `SessionsControllerTest` would need to be changed to:

```ruby
  assert_redirected_to try(:root_path) || "/"
```

Which may not look esthetically the best. Alternatively, Rails internals could be changed so that `root_path` would return "/" if it has never been defined in the routes, but that seems like too big of a user-facing change.

So to me it seems that it's probably best to either:

- explicitly define `root_path`, like what this PR is attempting to do
- be okay with `SessionsControllerTest` failing by default on new Rails apps, with the expectation that users need to manually set it themselves if they want their `SessionsControllerTest` to pass. Which would be a totally fine option as it is the current expectation so far.

jeromedalbert: Good point!

Incidentally I feel like even if it is decided that the `SessionsControllerTest` test is expected to fail, a separate PR could be made to the Rails codebase to actually run the generated authentication tests (and do whatever setup is needed to make them pass), just for extra test coverage in the Rails codebase.

PR #54720, opened 2025-03-08
add note about password validation of the authentication generator [ci-skip]

gregmolnar:
### Motivation / Background

This Pull Request has been created because it would be good to warn people about the fact that they authentication generator doesn't add any password length or complexity validation.

### DIFF
--- a/guides/source/security.md
+++ b/guides/source/security.md
@@ -182,6 +182,12 @@ class User < ApplicationRecord
 end
 ```
 
+NOTE: `has_secure_password` adds the following validations automatically:<br/><br/>
+- Password must be present on creation<br/>
+- Password length should be less than or equal to 72 bytes<br/>
+- Confirmation of password (using a XXX_confirmation attribute)<br/><br/>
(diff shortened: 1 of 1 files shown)


zzak:
Makes sense to me but the API docs also mention the following:

> The following validations are added automatically:
>
>  *  Password must be present on creation
>  *  Password length should be less than or equal to 72 bytes
>  *  Confirmation of password (using a XXX_confirmation attribute)

So maybe we should mention those here as well?

/cc @p8 who has opinions about this stuff too

gregmolnar:
@zzak good point, I changed it to mention those validations.

p8:
Looks good to me!

gregmolnar:
Thanks @p8! It fails the linter check due to not having lines around the list items. But since it is in a note block, I am not sure how to do that without breaking out of it. Is that possible anyhow?

zzak:
It doesn't have to be in a `NOTE` blob, that would fix it.

Screenshot from 2025-06-30 13-13-09

PR #54721, opened 2025-03-08
Fix association inverse in combination with composite foreign key / primary key

arthurschreiber:
### Motivation / Background

It looks like association inversion does not work if the association uses a composite foreign key / primary key. This adds a test case that shows this is not working, and fixes the foreign key checking logic to make it work as expected.

/cc @bensheldon @matthewd 

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -409,11 +409,13 @@ module ActiveRecord
         end
 
         def matches_foreign_key?(record)
-          if foreign_key_for?(record)
-            record.read_attribute(reflection.foreign_key) == owner.id ||
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
-          else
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/associations/association.rb
```diff
@@ -409,11 +409,13 @@ def inversable?(record)
         end
 
         def matches_foreign_key?(record)
-          if foreign_key_for?(record)
-            record.read_attribute(reflection.foreign_key) == owner.id ||
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
```
arthurschreiber, line 414: I'm really not sure what this logic here was trying to achieve, but it for sure didn't seem right. Either it was missing test cases, or it was never utilized. Either way, all tests pass with the reworked logic I added.

bensheldon: Is it possible to maintain something similar so that it avoids allocating arrays in the case of a single pkey?

PR #54722, opened 2025-03-08
Use a self-join for UPDATE with outer joins on PostgreSQL and SQLite

brunodccarvalho:
In PostgreSQL and SQLite `UPDATE` when an `OUTER JOIN` references the updated table in the `ON` clause, the join condition cannot be moved to the `WHERE` clause without breaking the query, so the current implementation is forced to use a subquery. We can support this with a more aggressive manipulation: we duplicate the updated table into the `FROM` clause with a self-join on the primary key. So, transform

```sql
UPDATE "products" SET ...
LEFT JOIN "categories" ON "products"."category_id" = "categories"."id"
  ... other joins ...
WHERE ...
```

into

```sql
UPDATE "products" AS "alias" SET ...
FROM "products"
LEFT JOIN "categories" ON "products"."category_id" = "categories"."id"
  ... other joins ...
WHERE "alias"."id" = "products"."id" AND ...
```

The table referenced by `"products"."?"` changes with this transformation. Top-level `FULL|RIGHT OUTER JOIN` are converted to `LEFT|INNER JOIN`. This should be fine given the update context.

On PostgreSQL this appends an index scan step to the query plan, just like a subquery.

Unfortunately it introduces a possible ambiguity: when the `SET` or `WHERE` references an unqualified column of the target table, as there are now two candidate tables. This shouldn't happen with AR generated nodes but it will affect users with handcrafted nodes.

### DIFF
--- a/activerecord/lib/arel/visitors/postgresql.rb
+++ b/activerecord/lib/arel/visitors/postgresql.rb
@@ -12,46 +12,21 @@ module Arel # :nodoc: all
 
           # UPDATE with JOIN is in the form of:
           #
-          #   UPDATE t1
+          #   UPDATE t1 AS __active_record_update_alias
           #   SET ..
-          #   FROM t2
-          #   WHERE t1.join_id = t2.join_id
-          #
-          # Or if more than one join is present:
-          #
-          #   UPDATE t1
-          #   SET ..
-          #   FROM t2
-          #   JOIN t3 ON t2.join_id = t3.join_id
-          #   WHERE t1.join_id = t2.join_id
+          #   FROM t1 JOIN t2 ON t2.join_id = t1.join_id ..
+          #   WHERE t1.id = __active_record_update_alias.id AND ..
           if has_join_sources?(o)
-            visit o.relation.left, collector
+            collector = visit o.relation.left, collector
             collect_nodes_for o.values, collector, " SET "
             collector << " FROM "
-            first_join, *remaining_joins = o.relation.right
-            from_items = remaining_joins.extract! do |join|
-              join.right.expr.right.relation == o.relation.left
-            end
-
-            from_where = [first_join.left] + from_items.map(&:left)
-            collect_nodes_for from_where, collector, " ", ", "
-
-            if remaining_joins && !remaining_joins.empty?
-              collector << " "
-              remaining_joins.each do |join|
-                visit join, collector
-                collector << " "
-              end
-            end
-
-            from_where = [first_join.right.expr] + from_items.map { |i| i.right.expr }
-            collect_nodes_for from_where + o.wheres, collector, " WHERE ", " AND "
+            collector = inject_join o.relation.right, collector, " "
           else
             collector = visit o.relation, collector
             collect_nodes_for o.values, collector, " SET "
-            collect_nodes_for o.wheres, collector, " WHERE ", " AND "
           end
 
+          collect_nodes_for o.wheres, collector, " WHERE ", " AND "
           collect_nodes_for o.orders, collector, " ORDER BY "
           maybe_visit o.limit, collector
         end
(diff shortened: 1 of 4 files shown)


activerecord/lib/arel/visitors/postgresql.rb
```diff
@@ -133,6 +134,12 @@ def visit_Arel_Nodes_Lateral(o, collector)
           grouping_parentheses o.expr, collector
         end
 
+        def visit_Arel_Nodes_InnerJoin(o, collector)
```
Edouard-chin, line 137: Would it make sense to create a new `CrossJoin` node instead of modifying the join type dynamically?

----

Also, I suppose this isn't necessary on sqlite as it creates the right join based on whether there is a ON condition ?

Edouard-chin (commented):
This is a very interesting solution!

> I wonder if this fixes https://github.com/rails/rails/pull/54598. cc @Edouard-chin @chaadow

It does!

activerecord/lib/arel/visitors/postgresql.rb
```diff
@@ -60,12 +43,30 @@ def visit_Arel_Nodes_UpdateStatement(o, collector)
...
-            (o.relation.right.all? { |join| join.is_a?(Arel::Nodes::InnerJoin) || join.right.expr.right.relation != o.relation.left })
-            o
+          return super unless has_join_sources?(o) &&
+            !has_limit_or_offset_or_orders?(o) && !has_group_by_and_having?(o)
+
+          # When all join source nodes are {INNER | CROSS} JOIN nodes, move the
+          # join conditions to the WHERE clause so they can reference the
+          # updated table. Otherwise, alias the updated table, duplicate it into
+          # the FROM clause, and add a self-join (which requires the primary key)
+          if o.relation.right.all?(Arel::Nodes::InnerJoin)
```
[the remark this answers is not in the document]
brunodccarvalho: I get the impression the self-join was slower from the query plan and a little experimentation on postgres.
With some help made up a simple [benchmark](https://pastebin.com/gXRmvuRm), it appears to be slower and on par with the subquery.

activerecord/lib/arel/visitors/postgresql.rb
brunodccarvalho: From the docs only postgresql rejects the INNER JOIN without ON, and sqlite adds [some extra semantics on top](https://www.sqlite.org/lang_select.html#special_handling_of_cross_join_) so "CROSS JOIN" there should actually be avoided.

About a new class I don't know what is better

activerecord/lib/arel/visitors/postgresql.rb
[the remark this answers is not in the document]
byroot: Kinda reduce risks as well. We found many regressions since I attempted to lift that limitation, so I'm more confortable with scoping the transformations to known cases.

activerecord/lib/arel/visitors/postgresql.rb
byroot: > About a new class I don't know what is better

Mutating the AST prior to calling the visitor is how it's supposed to work. Now sometimes it's a bit inconvenient and it turns out simpler to do what you've done.

I'd say give it a try, but feel free to push back if that sounds too messy to you.

activerecord/lib/arel/visitors/postgresql.rb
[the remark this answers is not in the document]
brunodccarvalho: Does look cleaner with just the self-join case. I'll leave it to someone else to haggle over the performance if they wish :stuck_out_tongue:

activerecord/lib/arel/visitors/postgresql.rb
brunodccarvalho: This doesn't matter much anymore with the special case for inner joins removed. But it seems like a cross join node would not do much: it sounds correct for `join.is_a?(InnerJoin)` to match cross joins, implying the `o.right` dummy nil shouldn't go away. And in sqlite3 we'd want to output `INNER JOIN` anyway, as explained in the docs, so the sql visitors would just trade on special case for another.

PR #53835, opened 2024-12-04
Add a config for automatically including `nonce` in `javascript_tag`, `javascript_include_tag` and `stylesheet_link_tag`

francktrouillez:
### Motivation / Background

When using `nonce` in context of [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), the `nonce` attribute can currently not be added automatically to the tags used in the views.

This means that the `nonce` attribute must be added manually to each tag (`javascript_tag`, `javascript_include_tag` and `stylesheet_link_tag`) that requires it.

From my understanding, there's currently no way to automatically add those nonces to the tags.

The current solutions to this I could identify are:
- Overriding the `javascript_include_tag` and `stylesheet_link_tag` helpers to explicitly add the `nonce` attribute.
- Adding manually `nonce: true` to each tag, even though this means tags in views coming straight from third-parties will not benefit from it, and might be blocked by the CSP, even though my application should trust them.

Ideally, I would like to have access to a configuration option to automatically enable the `nonce` attribute on the tags that require it.

I don't know if there's a better way to handle this, or if there's a specific reason why we wouldn't want to.

### Detail

This pull request main change is the addition of a new configuration option `content_security_policy_nonce_auto` that allows to automatically include a `nonce` to tags affected by the directives specified in the config option `content_security_policy_nonce_directives`. In order to keep the current behavior, the new configuration option is set to `false` by default.

### Additional information

This would allow developers to easily enable the `nonce` attribute for the tags that require it, without having to manually add it to each tag. This also means that gems that are trusted and used by the application would automatically benefit from the `nonce` attribute, even if the views are internal to the third-party.

Once again, I don't know if there's a better way to handle this, or if there's a specific reason why we wouldn't want to allow such behavior.

I'm open to any feedback or suggestions on how to improve this pull request.

Thank you for your time and consideration!

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54724, opened 2025-03-09
Make `nonce: false` remove the nonce attribute for `javascript_tag`, `javascript_include_tag` and `stylesheet_link_tag`

francktrouillez:
### Motivation / Background

Currently, when using `javascript_tag`, `javascript_include_tag` and `stylesheet_link_tag`, adding `nonce: false` option adds `nonce="false"` to the generated tag, instead of not adding the `nonce` attribute at all.

Since `nonce: true` option adds the `nonce` attribute, with an actual nonce as value, I would expect that `nonce: false` would not add the `nonce` attribute at all.

Currently, using `nonce: false` technically says that the tag has a nonce set to `"false"`, which I believe is not the intended behavior.

Let me know if I'm missing something here.

### Detail

This PR changes the behavior of `nonce: false` to not add the `nonce` attribute at all.

**It changes current behavior**, but I don't expect this to break any production code, as I strongly believe no one relies on a nonce with value `nonce="false"`, and even if they do, I don't think that it should be a practice Rails promotes.

Let me know your thoughts on this.

### Additional information

This was spotted when working on #53835.

Current solution to mitigate this issue is to set `nonce: nil` instead, but I think that `nonce: false` should do the opposite of `nonce: true`, which is not adding a `nonce`.

Please let me know if you have any questions or concerns.

Thank you for your time and consideration!

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Make `nonce: false` remove the nonce attribute from `javascript_tag`, `javascript_include_tag`, and `stylesheet_link_tag`.
+
+    *francktrouillez*
+
 *   Add `dom_target` helper to create `dom_id`-like strings from an unlimited
     number of objects.
 
--- a/actionview/lib/action_view/helpers/asset_tag_helper.rb
+++ b/actionview/lib/action_view/helpers/asset_tag_helper.rb
@@ -139,6 +139,8 @@ module ActionView
           }.merge!(options)
           if tag_options["nonce"] == true || (!tag_options.key?("nonce") && auto_include_nonce_for_scripts)
             tag_options["nonce"] = content_security_policy_nonce
+          elsif tag_options["nonce"] == false
+            tag_options.delete("nonce")
           end
           content_tag("script", "", tag_options)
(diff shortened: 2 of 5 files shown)


actionview/CHANGELOG.md
```diff
@@ -51,4 +51,8 @@
 
     *francktrouillez*
 
+*   Make `nonce: false` remove the nonce attribute from `javascript_tag`, `javascript_include_tag`, and `stylesheet_link_tag`.
+
+    *francktrouillez*
+
```
francktrouillez, line 57: :information_source: Since it changes behavior, I've added a line to the changelog. However, the bug is quite small and niche, so we can remove this change if you consider it irrelevant.

Let me know if you want me to make the change.

actionview/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
```
francktrouillez, line 4: :information_source: Since it changes behavior, I've added a line to the changelog. However, the bug is quite small and niche, so we can remove this change if you consider it irrelevant.

Let me know if you want me to make the change.

PR #54730, opened 2025-03-10
The skip_brakeman option should not skip bundle_audit

coorasse:
### Motivation / Background
While I was working on a skip_bundle_audit option I noticed that skip_brakeman skips also bundle_audit at the moment.

### Detail
This PR fixes the behavior so skip_brakeman actually skips only brakeman.


zzak:
@coorasse Thanks for your PR and sorry it hasn't received any feedback so far.

I think because we also have bundle-audit here the correct thing would be to combine them, so you don't end up with a `scan_ruby` job that does nothing. You could use `unless skip_bundle_audit? || skip_brakeman?`, I think.

Issue #32888, opened 2018-05-15
ActiveRecord doesn't correctly handle Infinity values for decimal columns with a scale


gilmoreorless:
In our codebase we use some models with decimal columns and validation for numericality and uniqueness. If the string `'Infinity'` is set as the value for a decimal field, the model's save action fails with a database exception, as it's passing the unquoted value `Infinity` straight to the database.

After digging into a root cause, I found it only happens when the decimal column has a scale parameter. ActiveRecord uses the `Decimal` type class for this case instead of `DecimalWithoutScale`. These two classes have different behaviour when handling `BigDecimal` values.

### Steps to reproduce

<details><summary>Full script to reproduce the bug:</summary>

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "activerecord", "5.2.0"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

ActiveRecord::Base.establish_connection({
  adapter: "sqlite3",
  database: ":memory:",
  prepared_statements: false,
})
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :valids, force: true do |t|
    t.decimal :value, precision: 10
  end

  create_table :invalids, force: true do |t|
    t.decimal :value, precision: 10, scale: 1
  end
end

class Valid < ActiveRecord::Base
  validates :value, presence: true, numericality: true, uniqueness: true
end

class Invalid < ActiveRecord::Base
  # NOTE: Removing the uniqueness check here makes the validation work correctly
  validates :value, presence: true, numericality: true, uniqueness: true
end

class BugTest < Minitest::Test
  def test_infinity_no_scale
    valid = Valid.new(value: 'Infinity')
    saved = valid.save
    assert !saved
    assert_equal valid.errors[:value], ["is not a number"]
  end

  def test_infinity_with_scale
    invalid = Invalid.new(value: 'Infinity')
    # BUG: This save will fail with an exception:
[...]

Issue #21262, opened 2015-08-17
BigDecimal::NaN not quoted in SQL query


Hugo-Hache:
With any model with a decimal column, if you try to query for the rows with a NaN value you get the following error:

<pre>
Model.where(column: BigDecimal::NAN)

PG::UndefinedColumn: ERROR:  column "nan" does not exist
LINE 1: ... "model"."column" = NaN
</pre>

The problem is the missing quotes around NaN, as <pre>Model.where(" column = 'NaN' ")</pre> works just fine.

By browsing the source code I presume there is a missing BigDecimal case in this class, https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb#L83, but being not aware of the AR internals I was not confident enough to propose a patch.

robin850:
Hi @Hugo-Hache,

Thanks for reporting! Yes, this is the right place to edit. Would you mind to open a pull request with the patch, a changelog entry and a test ?

BenEddy:
[Here](https://github.com/BenEddy/rails/commit/b1421b8ce8786503090f89cff09945c897969228) is a patch per @Hugo-Hache's suggestion. 

I can open a PR and apply it to this issue, but quick question: Is there a reason NaN and infinity handling doesn't happen in the [abstract implementation](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L155)? As far as I can tell `'NaN'` and `'Infinity'` are supported by SQLite, pg and mysql.

idler921:
Same problem for mysql

ActiveRecord::StatementInvalid: Mysql2::Error: Unknown column 'NaN' in 'field list': UPDATE `t` SET `f` = NaN WHERE `t`.id = 2

        from /usr/local/rvm/gems/ruby-2.2.1/gems/mysql2-0.4.2/lib/mysql2/client.rb:107:in `_query'
        from /usr/local/rvm/gems/ruby-2.2.1/gems/mysql2-0.4.2/lib/mysql2/client.rb:107:in `block in query'
        from /usr/local/rvm/gems/ruby-2.2.1/gems/mysql2-0.4.2/lib/mysql2/client.rb:106:in `handle_interrupt'
        from /usr/local/rvm/gems/ruby-2.2.1/gems/mysql2-0.4.2/lib/mysql2/client.rb:106:in `query'
[...]

PR #54735, opened 2025-03-11
fix: sqlite3 adapter should quote Infinity and NaN

flavorjones:
### Motivation / Background

When literal SQL strings are built by the sqlite3 adapter, the values "Infinity", "-Infinity", and "NaN" must be quoted. (This isn't a problem if the value is provided as a parameter to a prepared statement.)

For example, if `.upsert` is passed one of these values, the exception raised is:

```
>> Shirt.upsert({id: 1, size: Float::INFINITY})

sqlite3-2.6.0-x86_64-linux-gnu/lib/sqlite3/statement.rb:36:in 'SQLite3::Statement#prepare': no such column: Infinity: (SQLite3::SQLException)
INSERT INTO "shirts" ("id","size","created_at","updated_at") VALUES (1, Infinity, STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW'), STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')) ON CONFLICT ("id") DO UPDATE SET updated_at=(CASE WHEN ("size" IS excluded."size") THEN "shirts".updated_at ELSE STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW') END),"size"=excluded."size" RETURNING "id"
                                                                        ^
from sqlite3-2.6.0-x86_64-linux-gnu/lib/sqlite3/statement.rb:36:in 'SQLite3::Statement#initialize'
from sqlite3-2.6.0-x86_64-linux-gnu/lib/sqlite3/database.rb:216:in 'Class#new'
from sqlite3-2.6.0-x86_64-linux-gnu/lib/sqlite3/database.rb:216:in 'SQLite3::Database#prepare'
from rails/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb:94:in 'ActiveRecord::ConnectionAdapters::SQLite3::DatabaseStatements#perform_query'
...
from rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:558:in 'block in ActiveRecord::ConnectionAdapters::DatabaseStatements#raw_execute'
...
from rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:557:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#raw_execute'
from rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:601:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#internal_execute'
from rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:550:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#internal_exec_query'
from rails/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb:180:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#exec_insert_all'
from rails/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb:27:in 'ActiveRecord::ConnectionAdapters::AbstractAdapter#exec_insert_all'
from rails/activerecord/lib/active_record/insert_all.rb:54:in 'ActiveRecord::InsertAll#execute'
from rails/activerecord/lib/active_record/insert_all.rb:13:in 'block in ActiveRecord::InsertAll.execute'
...
```

Quoting these values does what's intended.

### Detail

This is similar to the fix applied to the postgresql adapter in #3713.

### Additional information

1. This may be a partial fix for #32888.
2. Based on comments in #21262, it's possible that these values should be quoted in the [abstract Quoting module](https://github.com/flavorjones/rails/blob/0085043a6e26589c43793a8d6f8af2af801c8f5e/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L72), but I have not investigated the behavior of MySQL2 or Trilogy to determine if this same problem exists there.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3/quoting.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3/quoting.rb
@@ -50,6 +50,19 @@ module ActiveRecord
           end
         end
 
+        def quote(value) # :nodoc:
+          case value
+          when Numeric
+            if value.finite?
+              super
+            else
+              "'#{value}'"
+            end
+          else
+            super
+          end
+        end
+
         def quote_string(s)
           ::SQLite3::Database.quote(s)
         end
--- a/activerecord/test/cases/adapters/sqlite3/quoting_test.rb
+++ b/activerecord/test/cases/adapters/sqlite3/quoting_test.rb
@@ -95,4 +95,16 @@ class SQLite3QuotingTest < ActiveRecord::SQLite3TestCase
       end
     end
   end
+
+  def test_quote_numeric_infinity
+    assert_equal "'Infinity'", @conn.quote(Float::INFINITY)
+    assert_equal "'-Infinity'", @conn.quote(-Float::INFINITY)
+    assert_equal "'Infinity'", @conn.quote(BigDecimal(Float::INFINITY))
+    assert_equal "'-Infinity'", @conn.quote(BigDecimal(-Float::INFINITY))
+  end
+
+  def test_quote_float_nan
+    assert_equal "'NaN'", @conn.quote(Float::NAN)
+    assert_equal "'NaN'", @conn.quote(BigDecimal(Float::NAN))
+  end
 end


flavorjones:
If it's useful, here's a self-contained repro of the issue I'm trying to solve:

```ruby
#!/usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "rails"
  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

    t.float :importance
  end
end

end

  def test_infinity
    post = Post.create!(importance: 1.0)

    Post.upsert({ id: post.id, importance: Float::INFINITY })
    post.reload
    assert_equal(Float::INFINITY, post.importance)
  end
end
```

PR #54736, opened 2025-03-11
Revert "Modify the Test Runner to allow running test at root:"

byroot:
Reverts rails/rails#54647
Fix: https://github.com/rails/rails/pull/54693#pullrequestreview-2675820144

This breaks the fairly common `bin/rails test test:system`.

### DIFF
--- a/railties/lib/rails/test_unit/runner.rb
+++ b/railties/lib/rails/test_unit/runner.rb
@@ -91,22 +91,12 @@ module Rails
 
         private
           def extract_filters(argv)
-            previous_arg_was_a_flag = false
             # Extract absolute and relative paths but skip -n /.*/ regexp filters.
             argv.filter_map do |path|
-              current_arg_is_a_flag = /^-{1,2}[a-zA-Z0-9\-_.]+=?.*\Z/.match?(path)
-
-              if previous_arg_was_a_flag && !current_arg_is_a_flag
-                # Handle the case where a flag is followed by another flag (e.g. --fail-fast --seed ...)
-                previous_arg_was_a_flag = false
-                next
-              end
+              next unless path_argument?(path)
 
               path = path.tr("\\", "/")
               case
-              when current_arg_is_a_flag
-                previous_arg_was_a_flag = true unless path.include?("=") # Handle the case when "--foo=bar" is used.
-                next
               when /(:\d+(-\d+)?)+$/.match?(path)
                 file, *lines = path.split(":")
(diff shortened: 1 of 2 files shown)


Edouard-chin:
Arf. Chaining rails command is indeed something I had not anticipated 😓

Edouard-chin:
I justed created a new app on 8.0.1 and running `bin/rails test test:system` doesn't raise any error, but it also doesn't run the `test:system` at all 😂 .

WDYT @byroot, I think the behaviour of failing loudly is better, no?

byroot:
Yeah probably. Given this mistake has creeped up. 

e.g. prior to https://github.com/rails/rails/pull/54703, the same mistake was present in the default GitHub action. 

So yeah, might as well error out loudly.

byroot:
Or perhaps we should try to make it work the way mulptiple people expected it to?

Edouard-chin:
I'll try to give it a go tomorrow 👍

matthewd:
FWIW, I think I'd prefer to keep it breaking... I think the do-both expectation is an echo of `rake` behaviour, and IMO we're better able to be clear about subcommand behaviour [and option/flag interpretation] if we keep our proper commands more strict/structured.

(Much as I don't love the implementation in #54647, for all the reasons already discussed there.)

Edouard-chin:
I looked at `bin/rails test test:system` and I'd agree with @matthewd.

While we could find a way to determine if any arg from ARGV is another command and try to run them all within a thor group, we'd have a lot of issues with options ambiguity.

For instance, what should happen if you do `bin/rails test test`, where the second `test` argument is a command. Is the argument part of the first test command, or a subcommand.

If we support `bin/rails test test:system` (and other `test:unit`, `test:models`... alternative), I suppose we'd need to support all commands to run altogether (`bin/rails test server console`).

At the end I figured that this is a different type of problems that I originally tried to address, and would require some decisions on your side before we implement it. I ended up opening #54741 as an alternative solution to what the one we reverted.

Edouard-chin:
Hey @KaanOzkan ❤️ !!

Thanks for the explanation. I didn't understand why this patch would suddenly trigger the sorbet runtime error but I got it now, and it's legit! I opened https://github.com/rails/rails/pull/54746 to explain, but I'm not sure if the current behaviour is actually the correct one.

This doesn't fix the root cause of the sorbet issue, and you may re-encounter it at some point, but we can try Jean's suggestion in a different patch.

PR #54738, opened 2025-03-12
AbstractAdapter#attempt_configure_connection: handle Timeout.timeout

### LINKED ISSUE #51780: Possible race condition in postgresql type_map initialization
### Steps to reproduce
Since upgrading from Rails 7.0 to 7.1 we've started to see the following exception:

```
# activerecord (7.1.3.2) lib/active_record/connection_adapters/postgresql_adapter.rb in get_oid_type
undefined method `key?' for nil:NilClass (NoMethodError)
          if !type_map.key?(oid)
                      ^^^^^
```

I haven't been able to replicate the behavior outside of production where it occurs when the application briefly loses its connection to the postgres database while under load. This is a multi-threaded environment. The server never recovers after entering this state. 

I know there was a change to type_map initialization at https://github.com/rails/rails/commit/dc4420c564565affadcd0d68db1af7ce2183bea4, and it looks to me like it's possible there's a race condition here. I'll update this issue with steps to replicate if I can manage to trigger this outside of production.

### Expected behavior
The application should recover from a brief database outage and reconnect.

### Actual behavior
The server is never able to reconnect. 

### System configuration
**Rails version**: 7.1.3.2

**Ruby version**: 3.2.4

byroot:
Fix: https://github.com/rails/rails/issues/51780
Followup: https://github.com/rails/rails/pull/54711

If `configure_connection` fails because of `Timeout.timeout`, the raised exception is `Timeout::ExitError` which directly inherits `Exception`, so we must rescue `Exception`.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
@@ -1221,7 +1221,7 @@ module ActiveRecord
 
         def attempt_configure_connection
           configure_connection
-        rescue
+        rescue Exception # Need to handle things such as Timeout::ExitException
           disconnect!
           raise
         end
--- a/activerecord/test/cases/adapter_test.rb
+++ b/activerecord/test/cases/adapter_test.rb
@@ -894,6 +894,29 @@ module ActiveRecord
         connection&.disconnect!
       end
 
+      test "disconnect and recover on #configure_connection timeout" do
+        connection = ActiveRecord::Base.connection_pool.send(:new_connection)
+
+        slow = [5]
+        connection.singleton_class.define_method(:configure_connection) do
+          if duration = slow.pop
+            sleep duration
+          end
+          super()
+        end
+
+        assert_raises Timeout::Error do
+          Timeout.timeout(0.2) do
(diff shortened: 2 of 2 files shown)


activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
```diff
@@ -1221,7 +1221,7 @@ def configure_connection
 
         def attempt_configure_connection
           configure_connection
-        rescue
+        rescue Exception # Need to handle things such as Timeout::ExitException
```
matthewd, line 1224: Should we use `ensure`, to cover the version range where Timeout used `throw`?

byroot: Hum, you have a point. I haven't tested with older versions of timeout.

byroot: Actually since https://github.com/rails/rails/commit/5fbaa524b99 (7.1) we do require `timeout >= 0.4`.

So I guess we're fine. It's a bit weird we depend on timeout even though we don't need it though.

PR #54741, opened 2025-03-12
Let minitest parse ARGV:

Edouard-chin:
### Problem

See #54647 for the original reason this needs to be fixed.

### Context

[The previous PR](https://github.com/rails/rails/pull/54647) was reverted as we thought it broke running `bin/rails test test:system` where in fact this command was never a valid command, but it was previously silently ignoring the `test:system` arg (anything not considered a path is discarded).
It was in some way a better behaviour as it was at least failing loudly.

Ultimately, #54647 implementation was not broken but the solution wasn't elegant and didn't reach any consensus, so I figured let's try a new approach.

----------------

Whether `bin/rails test test:system` should be supported (as discussed [here](https://github.com/rails/rails/pull/54736#issuecomment-2716156252)), is a different problem that is out of scope of this patch.

----------------

### Solution

Basically the issue is that it's impossible for the Rails test command to know which arguments should be parsed and which should be proxied to minitest. And what it tries to achieve is providing some sort of CLI that minitest lacks.

This implementation relies on letting minitest parse all options **and then** we load the files. Previously it was the opposite, where the rails runner would try to guess what files needs to be required, and then let minitest parse all options. By modifying the order, this allow to provide a "catch-all non-flag arguments" to minitest option parser.

--------------------

Now that we delegate the parsing of ARGV to minitest, I had to workaround two issues:

  1) Running any test subcommand such as `bin/rails test:system` is the equivalent of `bin/rails test test/system`, but in the former, ARGV is empty and the "test/system" string is passed as a ruby parameter to the main test command.

     So we need to mutate ARGV. **But** mutating argv can only be done inside a `at_exit` hook, because any mutation would be rolled back when the command exists. This happens before minitest
has run any test (at process exit).
     https://github.com/rails/rails/blob/f575fca24a72cf0631c59ed797c575392fbbc527/railties/lib/rails/command.rb#L140-L146

2) Running a test **without** the rails command: `ruby test/my_test.rb` would end up loading the `my_test.rb` file twice.
First, because of `ruby`, and then because of our minitest plugin. So we need to let our minitest plugin know whether loading file is required (e.g. did the user end up using the CLI).

cc/ @byroot @matthewd

### DIFF
--- a/railties/lib/minitest/rails_plugin.rb
+++ b/railties/lib/minitest/rails_plugin.rb
@@ -127,6 +127,11 @@ module Minitest
       options[:profile] = count
     end
 
+    opts.on(/^[^-]/) do |test_file|
+      options[:test_files] ||= []
+      options[:test_files] << test_file
+    end
+
     options[:color] = true
     options[:output_inline] = true
   end
@@ -136,6 +141,10 @@ module Minitest
   def self.plugin_rails_init(options)
     # Don't mess with Minitest unless RAILS_ENV is set
     return unless ENV["RAILS_ENV"] || ENV["RAILS_MINITEST_PLUGIN"]
+
+    if ::Rails::TestUnit::Runner.load_test_files
+      ::Rails::TestUnit::Runner.load_tests(options.fetch(:test_files, []))
+    end
 
     unless options[:full_backtrace]
       # Plugin can run without Rails loaded, check before filtering.
--- a/railties/lib/rails/test_unit/runner.rb
+++ b/railties/lib/rails/test_unit/runner.rb
@@ -26,6 +26,7 @@ module Rails
       TEST_FOLDERS = [:models, :helpers, :channels, :controllers, :mailers, :integration, :jobs, :mailboxes]
       PATH_ARGUMENT_PATTERN = %r"^(?!/.+/$)[.\w]*[/\\]"
       mattr_reader :filters, default: []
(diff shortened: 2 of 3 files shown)


byroot:
Looks good but we should also probably fix the `bin/rails test test:system` occurrences.

Edouard-chin:
Thanks Jean !

ron-shinall:
@Edouard-chin we have a system test helper that checks for an argument ("head") to conditionally disable headless mode for observing tests. We use it in the following manner:
`bin/rails test test/system/some_test.rb head`
or
`bin/rails test:system head`

However, now "head" is interpreted as the name or path to a test and it cannot be used. Do you have any suggestions for passing arguments in this manner?

Edouard-chin:
👋 @ron-shinall , I'd recommend using an environment variable instead. `HEADLESS=true bin/rails test test/system/some_test.rb`

Edouard-chin:
👋 

Ah well, that's unfortunate 😞 . I haven't had the chance to dig much yet but it seems to be  to a order problem. Now that the test gets loaded with the `rails` plugin (which is almost at then in the alphabetical order), any plugins that run prior and requires tests to be loaded may malfunction.

I'll try to think of a solution. It could be possible to just rename the `rails.rb` file plugin for `000_rails.rb` so that minitest runs it first (similar to Rails initializers), but I'll have to dig a bit more to see if there is no other alternatives.

PR #54742, opened 2025-03-13
Split `bin/rails test test:system` into two steps:

Edouard-chin:
`bin/rails test test:system` is not a thing. The `test:system` part is silently discarded.

This should be loudly failing with the change I'm trying to add in https://github.com/rails/rails/pull/54741, and there are discussions in #54736 whether running multiple commands à la rake should be supported.

But in the meantime, modifying this into two steps would ensure your system tests actually run.


etiennebarrie:
`bin/rails test:all` could have worked too.

Edouard-chin:
👋 Yeah that's true , but DHH wanted to run them in a different process (test:all shuffle all tests and run them all in a single process). https://github.com/rails/rails/pull/54693#discussion_r1990024697

PR #54746, opened 2025-03-13
Add tests ensuring CurrentAttributes eagerly define class methods

Edouard-chin:
> [!NOTE]
> This patch is all about whether the previous behaviour was a bug or the expected behaviour. I honestly don't know. If this is a legit bug and the current behaviour is the correct one, feel free to close this PR!

- ### Context

  In #54646, I fixed a bug. The patch introduced a behaviour change which was caught by sorbet typechecking.

  ------------------------

  ```ruby
    class Current < ActiveSupport::Current
      def foo
      end
    end
  ```

  Previously:

  ```ruby
  Current.method(:foo).source_location => nil
  ```

  After:

  ```ruby
   Current.method(:foo).source_location => [lib/active_support/current_attributes.rb", 191]
  ```

  --------------------

  Basically, before, we were never creating a delegation method to the class, and instead relied on `method_missing` entirely.

  ### Problem

  There is no problem per se, but this method_added hook used to do nothing. And the previous patch now creates a delegation on the class.

  This is because when a method is added, `respond_to?` on the instance will always return `true` since we have just defined it. When calling `attribute :foo`, a method on the **singleton class** gets created, but doesn't not trigger our method_added hook defined on the **class**.

### DIFF
--- a/activesupport/lib/active_support/current_attributes.rb
+++ b/activesupport/lib/active_support/current_attributes.rb
@@ -117,6 +117,9 @@ module ActiveSupport
           raise ArgumentError, "Restricted attribute names: #{invalid_attribute_names.join(", ")}"
         end
 
+        Delegation.generate(singleton_class, names, to: :instance, nilable: false, signature: "")
+        Delegation.generate(singleton_class, names.map { |n| "#{n}=" }, to: :instance, nilable: false, signature: "value")
+
         ActiveSupport::CodeGenerator.batch(generated_attribute_methods, __FILE__, __LINE__) do |owner|
           names.each do |name|
             owner.define_cached_method(name, namespace: :current_attributes) do |batch|
@@ -133,9 +136,6 @@ module ActiveSupport
             end
           end
         end
-
-        Delegation.generate(singleton_class, names, to: :instance, nilable: false, signature: "")
-        Delegation.generate(singleton_class, names.map { |n| "#{n}=" }, to: :instance, nilable: false, signature: "value")
 
         self.defaults = defaults.merge(names.index_with { default })
       end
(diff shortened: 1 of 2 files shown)


byroot:
Yeah, so:

```ruby
  test "instance deelgators are eagerly defined" do
    current = Class.new(ActiveSupport::CurrentAttributes) do
      def self.name
        "MyCurrent"
      end

      def regular
        :regular
      end

      attribute :attr, default: :att
    end

    assert current.singleton_class.method_defined?(:attr)
    assert current.singleton_class.method_defined?(:regular)
  end
```

Fails on the last assert. So I guess I see what you mean now. Yes the attribute delegation is already done explicitly, but the method added was to handle regular methods.

I'll see if I can fix the issue you spotted without breaking the regular method delegation.

byroot:
I added some tests to try understand what this solves, but I can't see to see what you reported.

Edouard-chin:
Sorry if this wasn't clear and thanks for pushing a test.
If you run the test you added on 8.0.1, you'll see that it fails.

> This patch is all about whether the previous behaviour was a bug or the expected behaviour. I honestly don't know. 

What i meant is that this patch doesn't fix anything currently broken. Rather it rolls back to the previous behaviour of not eagerly defining a class method and entirely rely on method_missig while still fixing the default bug thing from #54646. 

I wasn't sure whether the behaviour of 8.0.1 was the expected one, but as Matthew pointed out, we are now doing the more efficient thing. 

--------------

Merging this PR with the tests you added would be a good idea.

zzak:
Should this be backported then?

zzak:
Just in case, I opened backports for 8.0 (#54755) and 7.2 (#54756), but feel free to close them if I am misunderstanding this is not a bug fix :bow:

PR #54749, opened 2025-03-13
[8.0] Fix default GitHub Actions config

byroot:
Ref: https://github.com/rails/rails/pull/54742

`bin/rails test test:system` doesn't actually run system tests like `rake` would.

We need to invoke `bin/rails` twice.

cc @Edouard-chin


Edouard-chin:
If this wasn't even confusing enough, `bin/rails db:test:prepare test test:system` works correctly ...

This is because the `bin/rails` command detects the first `db:test:prepare` and see that it's not a thor command, and so fallback to running it in a rake process. When the process exists rake continues parsing the remaing argv and surprise surprise, `test` and `test:system` are also defined are rake tasks https://github.com/rails/rails/blob/8e9d63d04bd84d40e958950eea3586bc4fa2065b/railties/lib/rails/test_unit/testing.rake

PR #54753, opened 2025-03-13
Add support for regexp query conditions

jenshenny:
### Motivation / Background

There currently is not a way to use regular expression operators in the Active Record query interface. At Shopify, we're looking to have queries to be db-agnostic to use different DBMSs so raw sql operators wouldn't work. You can write a workable query with Arel but Arel is private API which shouldn't be used.

```ruby
comments = Comment.arel_table
Comment.where(comments[:body].matches_regexp("Thank you( again)? for the welcome"))
```

### Detail

After discussion, @skipkayhil suggested we could expose regex match behaviour as a hash condition.

```ruby
Book.where(title: /The/)
```

This was done by adding a predicate builder handler to handle `Regexp`s. `Regexp` supports options that would modify how the regex should behave that I chose not to support for now at least https://docs.ruby-lang.org/en/3.4/Regexp.html#class-Regexp-label-Modes. There isn't any verification that the regex will be compatible with the database either.

I also added regexp match support for SQLite for this feature to be available for all adapters. SQLite uses `REGEXP` operator, but it doesn't have it implemented by default. I think we should still have it implemented if the database has it supported. I tested it by loading a SQLite extension that implements the `regexp()` user function. Without it, a no such function error is raised `ActiveRecord::StatementInvalid: SQLite3::SQLException: no such function: REGEXP:`.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Support Regexp values in query hash conditions.
+
+    Active Record now supports accepting a Regexp without modifiers in a
+    hash condition. This will generate SQL using per-database adapter
+    regular expression operators.
+
+    ```ruby
+      Book.where(title: /The/)
+    ```
+
+    *Jenny Shen*
+
 *   Allow bypassing primary key/constraint addition in `implicit_order_column`
 
     When specifying multiple columns in an array for `implicit_order_column`, adding
--- a/activerecord/lib/active_record/relation/predicate_builder.rb
+++ b/activerecord/lib/active_record/relation/predicate_builder.rb
@@ -5,6 +5,7 @@ module ActiveRecord
     require "active_record/relation/predicate_builder/array_handler"
     require "active_record/relation/predicate_builder/basic_object_handler"
     require "active_record/relation/predicate_builder/range_handler"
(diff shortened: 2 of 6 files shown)


matthewd:
I don't think it's realistically possible for us to provide this API :confused:

I see that you've handled the possibility of the supplied regex having option flags set, but as you note, the overarching problem is that regex dialects vary substantially: to accept a Ruby Regexp and offer to send it to the database, we'd need to implement a full parse+rewrite for each DB adapter... or at least a parse that would then assert it only used some super-minimal allowlist of believed-universal regex operators. I don't think it's necessarily impossible for such a thing to exist, but it feels like a lot for Rails to take on.

Given that, I think this is a case where we're better off promoting the Arel API as the official answer: its moderately increased clunkiness hopefully serves as a nudge that the caller needs to stop and understand that they're supplying a string containing a SQL-dialect-specific regex, _not_ a Ruby one, and it's up to them to be sure it'll have the right meaning in their target DB(s).

> Arel is private API which shouldn't be used

If you're up for it, this seems like a good reason to finally make some movement on fixing that: have a look at which parts of Arel API are useful for this sort of operation, and get them minimally documented (in the not-`:nodoc:` sense... many may not necessarily need actual words).

PR #54754, opened 2025-03-13
Add RuboCop cache to GHA workflow templates

lovro-bikic:
Opening this PR as follow-up to https://github.com/rails/rails/pull/54703#issuecomment-2703802664 (cc @byroot, @Earlopain).

Adds RuboCop cache restoration to RuboCop job in GitHub Actions app and plugin workflow templates. The aim is to speed up RuboCop checks by sharing the cache between workflow runs, when possible. This speed-up is most evident as your codebase grows (personal example: codebase I'm working on takes 8 minutes to scan for RuboCop offenses without cache, and about 40 seconds with cache on subsequent runs).

The reasoning behind this implementation is explained in detail [here](https://lovro-bikic.github.io/github-actions-rubocop-workflow/) ([Wayback](https://web.archive.org/web/20250313210604/https://lovro-bikic.github.io/github-actions-rubocop-workflow/) link, for posterity). Recapping important points:
- Cache is invalidated when either GHA runner OS, Ruby version, any `.rubocop.yml` file in the codebase or `Gemfile.lock` change. This closely follows RuboCop's [cache validity](https://docs.rubocop.org/rubocop/usage/caching.html#cache-validity) rules and [cache action practice](https://github.com/actions/cache) to invalidate cache when runner OS changes
- On the default branch (master/main), a new cache entry is created after each workflow run. This ensures up-to-date RuboCop cache. It's implemented this way because [GHA cache is immutable and the only way to "update" it is to create a new entry](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache)
- On branches other than the default, cache is restored from the base branch (if available) and saved for that branch after the RuboCop run finishes as a new entry. These branches have a **single** cache (unless invalidation happens; see first bullet)

  The choice to have a single cache for these branches is a trade-off between using more cache storage and having slightly slower RuboCop checks if later commits change many source files (which would invalidate RuboCop's cache for those files and create new ones). On the default branch, creating a new cache entry for each commit is necessary to keep it up-to-date, but on other branches I don't think it's cost-effective to cache every commit (especially if not much has changed between commits)

Old cache entries are [deleted after not being used for at least 7 days](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). Users are also free to create [custom workflows to delete cache entries](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries) sooner.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
@@ -48,6 +48,8 @@ jobs:
 <%- unless skip_rubocop? -%>
   lint:
     runs-on: ubuntu-latest
+    env:
+      RUBOCOP_CACHE_ROOT: tmp/rubocop
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -57,6 +59,16 @@ jobs:
         with:
           ruby-version: .ruby-version
           bundler-cache: true
+
+      - name: Prepare RuboCop cache
+        uses: actions/cache@v4
+        env:
+          DEPENDENCIES_HASH: ${{ hashFiles('.ruby-version', '**/.rubocop.yml', 'Gemfile.lock') }}
+        with:
+          path: ${{ env.RUBOCOP_CACHE_ROOT }}
+          key: rubocop-${{ runner.os }}-${{ env.DEPENDENCIES_HASH }}-${{ github.ref_name == github.event.repository.default_branch && github.run_id || 'default' }}
+          restore-keys: |
+            rubocop-${{ runner.os }}-${{ env.DEPENDENCIES_HASH }}-
 
       - name: Lint code for consistent style
         run: bin/rubocop -f github
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add RuboCop cache restoration to RuboCop job in GitHub Actions workflow templates.
+
+    *Lovro Bikić*
+
 *   Skip generating mailer-related files in authentication generator if the application does
     not use ActionMailer
 
--- a/railties/lib/rails/generators/rails/plugin/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/plugin/templates/github/ci.yml.tt
@@ -9,6 +9,9 @@ jobs:
 <%- unless skip_rubocop? -%>
   lint:
     runs-on: ubuntu-latest
+    env:
+      RUBY_VERSION: <%= ENV["RBENV_VERSION"] || ENV["rvm_ruby_string"] || "#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}" %>
+      RUBOCOP_CACHE_ROOT: tmp/rubocop
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -16,8 +19,18 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: <%= ENV["RBENV_VERSION"] || ENV["rvm_ruby_string"] || "#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}" %>
+          ruby-version: ${{ env.RUBY_VERSION }}
           bundler-cache: true
+
+      - name: Prepare RuboCop cache
+        uses: actions/cache@v4
+        env:
+          DEPENDENCIES_HASH: ${{ hashFiles('**/.rubocop.yml', 'Gemfile.lock') }}
+        with:
+          path: ${{ env.RUBOCOP_CACHE_ROOT }}
+          key: rubocop-${{ runner.os }}-${{ env.RUBY_VERSION }}-${{ env.DEPENDENCIES_HASH }}-${{ github.ref_name == github.event.repository.default_branch && github.run_id || 'default' }}
+          restore-keys: |
+            rubocop-${{ runner.os }}-${{ env.RUBY_VERSION }}-${{ env.DEPENDENCIES_HASH }}-
 
       - name: Lint code for consistent style
         run: bin/rubocop -f github


byroot:
> * The choice to have a single cache for these branches is a trade-off between using more cache storage and having slightly slower RuboCop checks if later commits change many source files (which would invalidate RuboCop's cache for those files and create new ones).

Right, I'm almost tempted to say that branches shouldn't store their cache at all?

The idea being that you'd basically only re-parse files your branch has modified, which is already very good. I think it might even simplify the condition a bit? What do you think?

lovro-bikic:
> I'm almost tempted to say that branches shouldn't store their cache at all?

I think it depends on what the branch changes. If it's only source files, default branch's latest cache is good enough because you only pay the price for modified files. But if you change something which invalidates cache (e.g., change RuboCop version), RuboCop will have to scan everything. If you don't save this new cache on the branch, later commits (if there are any) will again have to scan everything.

For the former case (only modified source files), a new cache entry will mostly be a duplication of default branch's latest cache, so it's not that useful. For the latter case, it's going to be a fresh RuboCop cache which is useful if you push more commits to the branch afterwards.

Let's say that middle ground is to pull the latest cache from the default branch if available and otherwise create a single entry for the branch. We have a technical problem here because of cache immutability which forces us to add a dynamic component to the key such as workflow run ID, and I don't know how to look up the latest key without extra steps in the workflow.

I came up with this (overly complicated, but I don't know how to make it simpler) approach which uses `gh` CLI. This example _should_ work:
```yml
- name: Find workflow run ID for most recent RuboCop cache
  env:
    DEPENDENCIES_HASH: ${{ hashFiles('.ruby-version', '**/.rubocop.yml', 'Gemfile.lock') }}
  run: |
    LAST_WORKFLOW_RUN=$(gh cache list --ref refs/heads/master --sort created_at --order desc --key rubocop-cache- --json key --jq '.[].key' | grep $DEPENDENCIES_HASH | head -n 1 | rev | cut -d'-' -f1 | rev)
    echo "LAST_WORKFLOW_RUN=$LAST_WORKFLOW_RUN" >> "$GITHUB_ENV"
```

It lists RuboCop caches on the default branch (technically, master in this example) and finds the most recent one with a matching DEPENDENCIES_HASH. Then it takes its workflow run ID (`rev | cut -d'-' -f1 | rev`). 

`Prepare RuboCop cache` step stays mostly the same except for the `key` property which would be:
```
rubocop-cache-
${{ runner.os }}-
${{ env.DEPENDENCIES_HASH }}-
${{ github.ref_name == github.event.repository.default_branch && github.run_id || env.LAST_WORKFLOW_RUN || 'default' }}
```
meaning default branch always gets a new cache, and other branches either pull that cache or create their own cache as a fallback.

This would have to be tested in production, I don't know how this setup would function in practice.

But I think all this just goes to show that cache immutability is such a pain to work with. I'm not sure if this extra setup is worth the trouble versus just having an additional cache entry.

byroot:
> I think it depends on what the branch changes.

Indeed, but it only work of the important thing is changed in the first commit. Hence why at that point I think we might as well minimize the cache entries.

> I came up with this (overly complicated, but I don't know how to make it simpler) approach

Sorry, as mentioned previously, I'm really unfamiliar with GHA (I'm really used to a different CI system), but can't we just skip the cache saving if we're not on the default branch?

railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
```diff
@@ -58,6 +58,16 @@ jobs:
           ruby-version: .ruby-version
           bundler-cache: true
 
+      - name: Prepare RuboCop cache
+        uses: actions/cache@v4
+        env:
+          DEPENDENCIES_HASH: ${{ hashFiles('.ruby-version', '**/.rubocop.yml', 'Gemfile.lock') }}
+        with:
+          path: ~/.cache/rubocop_cache/
```
simi, line 66: Usually I do use custom path like `/tmp/rubocop` and call `bin/rubocop` with `--cache-root /tmp/rubocop` to keep it bulletproof for future changes of default cache location.

lovro-bikic:
> Sorry, as mentioned previously, I'm really unfamiliar with GHA (I'm really used to a different CI system), but can't we just skip the cache saving if we're not on the default branch?

We can, but we can't use `actions/cache` action then. We'd have to use separate actions `actions/cache/restore` before running RuboCop and `actions/cache/save` after running RuboCop. We can add an `if` condition to `actions/cache/save` to not save if we're not on the default branch. Take a look at [this example](https://github.com/actions/cache?tab=readme-ov-file#using-a-combination-of-restore-and-save-actions).

byroot:
Yeah, that's pretty much what I had in mind.

Another possibility is to have:

  - `uses: actions/cache@v4 if main_branch?`
  - `uses: actions/cache/restore@v4 unless main_branch?`

railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
lovro-bikic: That would make it more resistant to possible cache location changes, yes. I do this on some projects, but I use [`RUBOCOP_CACHE_ROOT`](https://docs.rubocop.org/rubocop/usage/caching.html#cache-path) env variable which I set up on job level so it's not duplicated, e.g. (truncated example):
```yml
jobs:
  rubocop:
    env:
      RUBOCOP_CACHE_ROOT: tmp/rubocop
    steps:
      - uses: actions/cache@v4
        path: ${{ env.RUBOCOP_CACHE_ROOT }}
      - name: RuboCop
        run: bundle exec rubocop
```

lovro-bikic: I just remembered that users can also override cache directory in their `.rubocop.yml` (`AllCops.CacheRootDirectory` property), so we definitely have to set something like this to make sure cache is in the right location on CI. I'll update the PR.

byroot:
@Earlopain sorry, I'm not sure to understand what you are advocating for? Do you prefer the current state of this PR?

lovro-bikic:
> The cache of a rails rubocop run is ~20mb 

Is this before or after compression (`actions/cache` compresses the)? 20 MB sounds like a lot, so I'm going to assume it's before. `actions/cache` compresses the cache before saving, so storage-wise it shouldn't take much space.

To give some data points, my Rails app has ~27k files scanned by RuboCop and GH reports that a cache entry takes around 1.1 MB. I see that Rails codebase has around ~3.4k files, so the demand on storage should be much lower.

Earlopain:
> Is this before or after compression (actions/cache compresses the)? 20 MB sounds like a lot, so I'm going to assume it's before. 

Oops, yeah. It's before. I just measured it locally. I guess they compress rather well, considering it's mostly files with `[]` inside them.

simi:
> Yeah, I see your point, I'm also annoyed at how complex GHA config files get. Alright, I think I'm OK with the current behavior.

@byroot Do you mean how complex the GHA config template gets or GHA config file in general?

byroot:
> GHA config file in general?

That one. I don't mind the template being complex.

It's one thing to generate files for users, but then they'll likely need to modify it at some point, hence need to understand it. So complex setups are annoying, because we're dumping complexity on the user.

This can be in part counteracted with some comments, but still.

railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
lovro-bikic: Updated.

lovro-bikic:
@byroot I'm done with the latest changes. The two changes I made are adding `RUBOCOP_CACHE_ROOT` (per https://github.com/rails/rails/pull/54754#discussion_r1995316805) and simplifying the cache key prefix from `rubocop-cache-` to just `rubocop-` (`cache` felt redundant to me).

I tested this by creating a new Rails app and copying the generated workflow to my project. It ran as expected, so I'm satisfied on my end.

chaadow:
@lovro-bikic Thanks for your article and your contribution. In my case I don't run the CI ( and precisely the rubocop job ) on the main branch.

And I was having 10 pull requests that only changed HTML files, but still, everytime i open a PR, the cache has to be created from scratch the first time
And that is because sibling pull requests that target the same branch ( in my case 'main' branch), **cannot share the same cache because of isolation** even though they have the same key (`rubocop-Linux-ba55c373bf168863bcaeb0a8fba7700b26aab1468ee15ae112765dd356318f98-default` in my case)

Here is [the relevant section](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

I think it's important to note that you need to have : 
```
on:
  pull_request:
  push:
    branches: [ main ]
```

In order to have the cache  be shared between feature branches. I don't know if this needs to be added as a comment to the CI. because for my case, we deleted the : 
```
  push:
    branches: [ main ]
```
because we didn't think it was necessary, before reading that section. What are your thoughts?

chaadow:
@lovro-bikic Actually the proper tip I got it from here : https://github.com/actions/cache/blob/main/tips-and-workarounds.md#use-cache-across-feature-branches

> a good way to achieve this is to ensure that the default branch has a cache

lovro-bikic:
> In my case I don't run the CI ( and precisely the rubocop job ) on the main branch.

Is there a technical reason for this? Personally, I'd always want to run CI on master/main because it's the branch I'm deploying to production. I've had cases in the past where CI would break after a merge, or because of a flaky spec, or some other issue, so knowing your default branch is green seems to me important.

> I don't know if this needs to be added as a comment to the CI.

This is not my call (perhaps @byroot can weigh in?), but the way I see it, any change to the template workflow could break something, and adding a comment for every technicality could clutter it. Also, removing CI for the default branch seems like an edge case, so documenting only for that might not bring much value overall. But as I said, not my call.

Hope this helps!

PR #54755, opened 2025-03-14
[8-0-stable] Ensure CurrentAttributes eagerly define class methods

zzak:
Backports #54746 to `8-0-stable`

/cc @Edouard-chin @byroot


Edouard-chin:
Thanks @zzak !

PR #54759, opened 2025-03-14
Add a run hook to `ActiveRecord::Migration`

percysnoodle:
### Motivation / Background

In order to support our deployment process, we'd like to include a module in `ActiveRecord::Migration`.  We don't want to load the class unnecessarily.  `ActiveSupport` provides a method to run code when a class is loaded, but it depends on a hook having been defined for the class in question.  

### Detail

This commit adds such a hook when the Migration class, and its support classes in the same file, have been loaded.

### Additional information

The only test I could find involving load hooks was `test_inclusion_runs_active_record_fixtures_load_hook`.  If there's a way to test this change, please let me know!

### Checklist

* [x] Tests are added or updated if you fix a bug or add a feature. (checked, but see above)

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add a load hook for `ActiveRecord::Migration`
+
+    *Simon Booth*
+
 *   Allow bypassing primary key/constraint addition in `implicit_order_column`
 
     When specifying multiple columns in an array for `implicit_order_column`, adding
--- a/activerecord/lib/active_record/migration.rb
+++ b/activerecord/lib/active_record/migration.rb
@@ -1627,4 +1627,6 @@ module ActiveRecord
         MIGRATOR_SALT * db_name_hash
       end
   end
+
+  ActiveSupport.run_load_hooks(:active_record_migration, Migration)
 end


fatkodima:
> In order to support our deployment process

Explain more.

> We don't want to load the class unnecessarily

Why? What is the problem?

percysnoodle:
>> In order to support our deployment process
>
> Explain more.

Most recently, I was trying to add a reusable migration step which we could use to execute
```sql
SET FOREIGN_KEY_CHECKS=0;
```
at the beginning of a block, then
```sql
SET FOREIGN_KEY_CHECKS=1;
```
at the end regardless of the direction of the migration.  I tried including a module in `Mysql2Adapter`, since there's already a run hook defined for that and the migration delegates `execute` to that class, but it doesn't have access to the `reversible` method.

Perhaps "deployment process" was the wrong phrase - maybe I should have said "large database"?

>> We don't want to load the class unnecessarily
>
> Why? What is the problem?

Well, we could just include our module into `ActiveRecord::Migration` unconditionally, but that would mean that we loaded that class even when we weren't going to use it.  Loading a class we're not going to use adds time and memory consumption.  Maybe not a lot, but it adds up.  Isn't that why `ActiveSupport::on_load` exists?

fatkodima:
> Well, we could just include our module into ActiveRecord::Migration unconditionally

This is the simplest and correct solution. 

> Loading a class we're not going to use adds time and memory consumption. 

:) You wouldn't really notice a difference measured in nanoseconds or kilobytes. There are so many classes in rails apps, that loading one or ten more will not make a difference.

percysnoodle:
@fatkodima oh well, thank you for your consideration.

Out of interest, do you happen to know what's different about `ActiveRecord::ConnectionAdapters::Mysql2Adapter`, that it would get a load hook?  It'd be good to understand the rules about when things do or don't so I don't waste time raising more PRs like this.

fatkodima:
It has more code to load and the main reason is for gem developers to know when the project uses mysql and add mysql specific code. I don't think that currently any more code needs this on load hook.

PR #54766, opened 2025-03-15
Add `http_token_authenticate_with`

c960657:
### Motivation 
Just like `http_basic_authenticate_with` offers a [quick way to add HTTP Basic password protection](https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Basic.html) to a controller, I suggest introducing `http_token_authenticate_with` for token authentication.

This is useful for protecting an API with a fixed shared secret, e.g. when two internal systems are talking to each other. To prevent hard-coding credentials, the token could be set via an environment variable.

Examples:
`http_basic_authenticate_with name: "john", password: "secret"`

`http_token_authenticate_with token: "secret-token"`

### Checklist


actionpack/lib/action_controller/metal/http_authentication.rb
```diff
@@ -429,14 +418,32 @@ module Token
...
 
       module ControllerMethods
+        extend ActiveSupport::Concern
+
+        module ClassMethods
+          # Enables HTTP Token authentication.
+          #
+          # See ActionController::HttpAuthentication::Token for example usage.
+          def http_token_authenticate_with(token:, realm: nil, **options)
+            raise ArgumentError, "Expected token: to be a String, got #{name.class}" unless name.is_a?(String)
```
dwightwatson, line 428: This should be `token`, right?

```suggestion
            raise ArgumentError, "Expected token: to be a String, got #{token.class}" unless token.is_a?(String)
```

PR #54774, opened 2025-03-17
gitignore all key files in config/credentials

gregmolnar:
As mentioned in https://github.com/rails/rails/discussions/54773, it would be a good idea to gitignore all key files with a wildcard. That's how it is done in the dockerignore file too.

### DIFF
--- a/railties/lib/rails/generators/rails/encryption_key_file/encryption_key_file_generator.rb
+++ b/railties/lib/rails/generators/rails/encryption_key_file/encryption_key_file_generator.rb
@@ -21,18 +21,20 @@ module Rails
 
           log ""
           add_key_file_silently(key_path, key)
+          ensure_key_files_are_ignored(key_path)
           log ""
         end
       end
 
       def add_key_file_silently(key_path, key = nil)
         create_file key_path, key || ActiveSupport::EncryptedFile.generate_key, perm: 0600
+        ensure_key_files_are_ignored_silently(key_path)
       end
 
-      def ignore_key_file(key_path, ignore: key_ignore(key_path))
+      def ensure_key_files_are_ignored(key_path, ignore: key_ignore(key_path))
         if File.exist?(".gitignore")
           unless File.read(".gitignore").include?(ignore)
-            log "Ignoring #{key_path} so it won't end up in Git history:"
+            log "Ignoring #{ignore} so it won't end up in Git history:"
             log ""
             append_to_file ".gitignore", ignore
             log ""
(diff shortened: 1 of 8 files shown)


zzak:
Could you add a test please?

gregmolnar:
@zzak I added the test.

railties/lib/rails/generators/rails/app/templates/gitignore.tt
```diff
@@ -35,3 +35,4 @@
 
 /public/assets
 <% end -%>
+/config/credentials/*.key
```
zzak, line 38: Rather than adding it here, why don't we generate it like we do with the master key?

https://github.com/rails/rails/blob/951ea8f75d9bf6d2999df2f0497347976dce9476/railties/lib/rails/generators/rails/app/app_generator.rb#L191-L205

https://github.com/rails/rails/blob/951ea8f75d9bf6d2999df2f0497347976dce9476/railties/lib/rails/generators/rails/master_key/master_key_generator.rb#L35-L41

gregmolnar: I changed it to use that. It took slightly more code, but I made it in a separate commit so I can revert if the previous version is preferred.

zzak:
Wouldn't it make sense to use `Rails::Generators::CredentialsGenerator` instead? I thought there were separate commands for each key we are generating so it makes sense to update the gitignore there.

gregmolnar:
The actual keygeneration happens in `Rails::Generators::EncryptionKeyFileGenerator` and you are right, that's  a much better place for this, because it will add the new gitignore entry even for old apps when they generate a new key, ie they introduce a new environment.

gregmolnar:
@zzak I changed this quite a bit. I moved the gitignore part into Rails::Generators::EncryptionKeyFileGenerator and calling it from there every time a key is generated. This made obsolete the calls to igoring the keys from the various parts in the codebase, so I removed those.

railties/lib/rails/generators/rails/encryption_key_file/encryption_key_file_generator.rb
```diff
@@ -44,13 +46,18 @@ def ignore_key_file(key_path, ignore: key_ignore(key_path))
...
+            append_to_file ".gitignore", ignore
+          end
+        end
       end
 
       private
         def key_ignore(key_path)
-          [ "", "/#{key_path}", "" ].join("\n")
+          key_path = Pathname.new(key_path) unless key_path.is_a?(Pathname)
+          [ "", "# Ignore key files for decrypting credentials and more.", "/#{key_path.dirname.join("*.key")}", "" ].join("\n")
```
zzak, line 60:
```suggestion
          <<~IGNORE.squish
            
            # Ignore key files for decrypting credentials and more.
            /#{key_path.dirname.join("*.key")}
            
          IGNORE
```

You might want to inspect the output to make sure indents, etc line up.

gregmolnar:
Are there any changes I should make here to get this merged?

zzak:
Could you add a changelog?

gregmolnar:
Sorry, I missed the changelog comment somehow. Thanks for adding it @zzak!

PR #54775, opened 2025-03-17
Add an easy way to translate your enum invalid value

Asiaticous:
### Motivation / Background

When an application supports multiple languages, one issue that arises is the initialization of a class with enums. If the application populates enum fields in bulk, Rails immediately validates the passed values and raises an ArgumentError if they are invalid. However, the error message is always fixed, and the enum name appears exactly as defined in the code—including underscores—rather than being properly translated.

### Detail

This change introduces custom translations for enum values and includes the class name where the enum is defined, improving clarity and localization.

### DIFF
--- a/activerecord/lib/active_record/enum.rb
+++ b/activerecord/lib/active_record/enum.rb
@@ -161,7 +161,7 @@ module ActiveRecord
   #
   #   conversation = Conversation.new
   #
-  #   conversation.status = :unknown # 'unknown' is not a valid status (ArgumentError)
+  #   conversation.status = :unknown # 'unknown' is not a valid Status (ArgumentError)
   module Enum
     def self.extended(base) # :nodoc:
       base.class_attribute(:defined_enums, instance_writer: false, default: {})
@@ -170,10 +170,11 @@ module ActiveRecord
     class EnumType < Type::Value # :nodoc:
       delegate :type, to: :subtype
 
-      def initialize(name, mapping, subtype, raise_on_invalid_values: true)
+      def initialize(name, mapping, subtype, klass, raise_on_invalid_values: true)
         @name = name
         @mapping = mapping
         @subtype = subtype
+        @klass = klass
         @_raise_on_invalid_values = raise_on_invalid_values
       end
 
@@ -203,14 +204,15 @@ module ActiveRecord
         return unless @_raise_on_invalid_values
 
         unless value.blank? || mapping.has_key?(value) || mapping.has_value?(value)
-          raise ArgumentError, "'#{value}' is not a valid #{name}"
+          field_name = klass.human_attribute_name(name)
+          raise ArgumentError, I18n.t("errors.messages.invalid_enum", value: value, name: field_name)
         end
       end
 
       attr_reader :subtype
 
       private
-        attr_reader :name, :mapping
+        attr_reader :name, :mapping, :klass
     end
 
     def enum(name, values = nil, **options)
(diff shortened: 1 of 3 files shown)


Edouard-chin:
Thanks for the Pull Request. I don't think translating exception message matters, those exceptions are not supposed to be seen by users of your application. 
If you'd like to provide a translated message to a user, when an invalid value is provided, you can already do so using the `validate` option when defining your enum.

https://github.com/rails/rails/blob/951ea8f75d9bf6d2999df2f0497347976dce9476/activerecord/lib/active_record/enum.rb#L122

Asiaticous:
> Thanks for the Pull Request. I don't think translating exception message matters, those exceptions are not supposed to be seen by users of your application. 
> 
> If you'd like to provide a translated message to a user, when an invalid value is provided, you can already do so using the `validate` option when defining your enum.
> 
> 
> 
> https://github.com/rails/rails/blob/951ea8f75d9bf6d2999df2f0497347976dce9476/activerecord/lib/active_record/enum.rb#L122

Thanks for the response! I understand that exception messages aren’t meant for users, but my PR is focused on improving the translation of validation errors when an invalid enum value is assigned.

The issue arises because the validation runs before saving, even when simply assigning an incorrect enum value. In cases like massive imports or API endpoints, I often need to initialize an instance with user-provided data and validate it before saving. Since the validation is correct, it would be useful to have a way to customize the error message.

Without this change, if I have multiple classes using enums, I’d need to implement a custom validation and translation for each one.

Asiaticous:
> Sorry, I don't fully follow what issue this is causing you. Just to make sure we are on the same page:
> 
> > but my PR is focused on centralizing the translation of validation errors [...] I often need to initialize an instance with user-provided data and validate it before saving
> 
> Do you need to display the error message raised by the exception to the users of your application?
> 
>     1. If no, why do you need to translate the error message?
> 
>     2. If yes, why don't you use the built-in validation feature when declaring an enum?
> 
> 
> ```ruby
> class Shop < ApplicationRecord
>   enum :state, [ :opened, :closed ], validate: { message: :"invalid_shop_state" }
> end
> ```
> 
> ```yaml
> en:
>   activerecord:
>     errors:
>       models:
>         shop:
>           attributes:
>             state:
>               invalid_shop_state: "can't be set to '%{value}'"
> ```
> 
> ```ruby
> shop = Shop.new(state: :renovating)
> puts shop.valid? #=> false
> 
> puts shop.errors.full_messages #=> "State can't be set to 'renovating'"
> ```

Many applications, including mine, handle massive imports where users upload large datasets via Excel or API. This often involves multiple models like `Shop`, `Customer`, and `Product`, each requiring the same `enum` validation:  

```ruby
enum :state, [:opened, :closed], validate: { message: :"invalid_shop_state" }
```

Since the validation message is always the same, using `assert_valid_value` directly is more efficient and reduces the risk of human errors from duplicating invalid_class_enum across multiple YAML files. This approach keeps the code DRY while maintaining clear validation behavior.

It can also be used with:
```ruby
enum :status, [ :active, :archived ], validate: true
```
Additionally, we can add the default error message so that:
```ruby
instance.errors.full_messages 
```
returns the same message as assert_valid_value.

Edouard-chin:
Thanks for the explanation, I understand your use case now 😅. I'll let the core team decide whether this is a feature worth adding, but IMHO I doubt so. 

> [...] reduces the risk of human errors from duplicating invalid_class_enum across multiple YAML files

In case you'd like to avoid having copy pasting translation, active model looks up the translation in multiple scopes, which will apply to all your models.

```yml
en:
  errors:
    messages:
      invalid_state: "blabla"
```

PR #54783, opened 2025-03-19
[ci skip] Remove warning and recommendation from Parameters#require

### LINKED ISSUE #54782: Documentation for ActionController::Parameters#require has incorrect warning
### Steps to reproduce

> This method is not recommended for fetching terminal values because it does not permit the values. For example, this can cause problems:

```
# CAREFUL
params = ActionController::Parameters.new(person: { name: "Finn" })
name = params.require(:person).require(:name) # CAREFUL
```

### Expected behavior

Some error

### Actual behavior

```
> Rails.version
=> "8.0.2"
> RUBY_VERSION
=> "3.2.6"
> params = ActionController::Parameters.new(person: { name: "Finn" })
> params.require(:person).require(:name)
=> "Finn"
```

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.2.6

quasimik:


quasimik:
closes #54782

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -503,19 +503,6 @@ def permit!
     #     user_params, profile_params = params.require([:user, :profile])
     #     # ActionController::ParameterMissing: param is missing or the value is empty or invalid: user
     #
-    # This method is not recommended for fetching terminal values because it does
```
Edouard-chin, line 506: If I'm not mistaken, the example is confusing/incorrect but it's still worth the note, and we should instead modify the example.

I think "fetching terminal values" was meant for when you need to fetch a non-scalar value, which ends up being not permitted:

```ruby
params = ActionController::Parameters.new(animals: { cat: { age: 7 } })

# If you need to get the values for the cat:
params.require(:animals).require(:cat)
=> #<ActionController::Parameters {"age"=>7} permitted: false> # Careful

# The recommendation 

post-css(dev)> params.expect(animals: { cat: :age }).require(:cat)
=> #<ActionController::Parameters {"age"=>7} permitted: true>
```

PR #47310, opened 2023-02-08
Fix regression in ActionDispatch::Routing::RouteSet#recognize_path

olefriis:
### Motivation / Background

#47262 includes a subtle change to `ActionDispatch::Routing::RouteSet#recognize_path`: If given `nil`, it will now fail like

```
NoMethodError: undefined method `include?' for nil:NilClass
    /workspaces/rails/actionpack/lib/action_dispatch/routing/route_set.rb:887:in `recognize_path'
```

instead of providing the root path.

### Detail

This Pull Request brings back the previous behaviour, so that `recognize_path(nil)` gives the root path.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] ~CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.~

PR #54784, opened 2025-03-19
Fix ActionDispatch::Routing::RouteSet#recognize_path regression against symbols and give clear error for other types

araishikeiwai:
### Motivation / Background

Similar to #47310, I noticed a regression in `ActionDispatch::Routing::RouteSet#recognize_path` against symbols that was introduced in #47262. Rails 7.0 accepted symbols in this method.

Rails 7.0:

```ruby
[1] pry(main)> Rails.application.routes.recognize_path(:accounts)
=> {:controller=>"accounts", :action=>"index"}
[2] pry(main)> Rails.application.routes.recognize_path({ controller: 'test' })
TypeError: no implicit conversion of Hash into String
from /Users/araishikeiwai/.rbenv/versions/3.2.7/lib/ruby/gems/3.2.0/gems/actionpack-7.0.8.5/lib/action_dispatch/routing/route_set.rb:857:in `match?'
```

vs 7.1 and above:

```ruby
[1] pry(main)> Rails.application.routes.recognize_path(:accounts)
NoMethodError: undefined method `include?' for :accounts:Symbol
from /Users/araishikeiwai/.rbenv/versions/3.2.7/lib/ruby/gems/3.2.0/gems/actionpack-7.1.5.1/lib/action_dispatch/routing/route_set.rb:887:in `recognize_path'
[2] pry(main)> Rails.application.routes.recognize_path({ controller: 'test' })
NoMethodError: undefined method `encoding' for {:controller=>"test"}:Hash
from /Users/araishikeiwai/.rbenv/versions/3.2.7/lib/ruby/gems/3.2.0/gems/actionpack-7.1.5.1/lib/action_dispatch/journey/router/utils.rb:19:in `normalize_path'
```

### Detail

This Pull Request keeps the method behaviour similar to previous version when given a symbol, and gives a clear indication of the error when given anything else

### Additional information

Alternative solution would be to outright reject symbols and raise `TypeError`

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/routing/route_set.rb
+++ b/actionpack/lib/action_dispatch/routing/route_set.rb
@@ -910,7 +910,10 @@ module ActionDispatch
 
       def recognize_path(path, environment = {})
         method = (environment[:method] || "GET").to_s.upcase
-        path = Journey::Router::Utils.normalize_path(path) unless path&.include?("://")
+
+        raise TypeError, "path must be a String or Symbol" if path && !(path.is_a?(String) || path.is_a?(Symbol))
+
+        path = Journey::Router::Utils.normalize_path(path) unless path&.to_s&.include?("://")
         extras = environment[:extras] || {}
 
         begin
--- a/actionpack/test/controller/routing_test.rb
+++ b/actionpack/test/controller/routing_test.rb
@@ -2125,6 +2125,7 @@ class RackMountIntegrationTests < ActiveSupport::TestCase
 
     assert_equal({ controller: "posts", action: "ping" }, @routes.recognize_path("/posts/ping", method: :get))
     assert_equal({ controller: "posts", action: "index" }, @routes.recognize_path("/posts", method: :get))
+    assert_equal({ controller: "posts", action: "index" }, @routes.recognize_path(:posts, method: :get))
     assert_equal({ controller: "posts", action: "index" }, @routes.recognize_path("/posts/index", method: :get))
     assert_equal({ controller: "posts", action: "show" }, @routes.recognize_path("/posts/show", method: :get))
     assert_equal({ controller: "posts", action: "show", id: "1" }, @routes.recognize_path("/posts/show/1", method: :get))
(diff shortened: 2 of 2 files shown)


actionpack/lib/action_dispatch/routing/route_set.rb
```diff
@@ -910,7 +910,10 @@ def call(env)
 
       def recognize_path(path, environment = {})
         method = (environment[:method] || "GET").to_s.upcase
-        path = Journey::Router::Utils.normalize_path(path) unless path&.include?("://")
+
+        raise TypeError, "path must be a String or Symbol" if path && !(path.is_a?(String) || path.is_a?(Symbol))
```
zzak, line 914: I think if we want to introduce an error for this path, it must be deprecated first.

That is just my feeling though, I haven't looked deeper into the issue.

skipkayhil (commented):
Since this method is private API I'm not sure it makes sense to add a type check to it. How are you currently getting a symbol in here?

araishikeiwai:
@skipkayhil we have some API that relies on this method that has breaking tests upgrading to 7.1. I can typecheck in there before calling this method of course but I also noticed the difference in behaviour against symbols which is  a regression and not mentioned in the changelog.

araishikeiwai:
Also I'm not sure what you meant by private API but it's available from Rails routes

```ruby
 Rails.application.routes.method(:recognize_path)
=> #<Method: ActionDispatch::Routing::RouteSet#recognize_path(path, environment=...) /Users/araishikeiwai/.rbenv/versions/3.2.7/lib/ruby/gems/3.2.0/gems/actionpack-7.1.5.1/lib/action_dispatch/routing/route_set.rb:885
```

skipkayhil:
> Also I'm not sure what you meant by private API but it's available from Rails routes

If the method isn't documented in https://api.rubyonrails.org/, then it's private to the framework and can be changed at any time without deprecation/changelog.

araishikeiwai:
That's fair, but the fact remains that it's still a widely used method

Screenshot 2025-03-25 at 14 41 53

If we're not going to add type check I was hoping we could still fix the change in behaviour somehow

PR #54787, opened 2025-03-20
Ignore .rubocop.yml when building Docker image

delphaber:
### Motivation / Background

Since rubocop gem is required in development and test environments only, we can ignore its configuration file when building the Docker image for deployment.

### Detail

I was not 100% sure about where to place the new ignore line, but I think the "# Ignore development files" section was a good one.

### Additional information

Since users can skip rubocop installation, the template can react to this.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/dockerignore.tt
+++ b/railties/lib/rails/generators/rails/app/templates/dockerignore.tt
@@ -55,6 +55,9 @@
 <% end -%>
 # Ignore development files
 /.devcontainer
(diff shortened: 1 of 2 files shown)


matthewd:
Unlike the potentially-huge other things that are being explicitly ignored there, I don't think there's any particular benefit to omitting one generally-small text file?

byroot:
I agree with @matthewd, there will always be some development files "leaking" in the production container but unless that siginificantly increase the image size it's not really worth bothering.

PR #54788, opened 2025-03-20
Refactor IsolatedExecutionState and ExecutionContext to be classes

adrianna-chang-shopify:
### Motivation / Background

This PR refactors `ActiveSupport::IsolatedExecutionState` and `ActiveSupport::ExecutionContext` to be classes, allowing independent execution contexts to exist. At Shopify, we'd like to leverage `ExecutionContext` to store [Error Reporter metadata](https://github.com/rails/rails/blob/3235827585d87661942c91bc81f64f56d710f0b2/activesupport/lib/active_support/error_reporter.rb#L201), but we want the context object to use Fiber storage under the hood so that metadata is inherited by child threads / fibers without clobbering context in the main thread. 

Originally, this work began with changes to support a `:fiber_storage` option on the `IsolatedExecutionState` singleton, but we realized this created issues with other aspects of the framework that rely on `IsolatedExecutionState`, such as DB connections and `CurrentAttributes`. @tenderlove and I settled on first refactoring the execution context modules to allow independent execution states to exist; this will enable us to configure the behaviour for the Error Reporter's execution context without impacting the global isolation level.

Future PRs will add support for `:fiber_storage` isolation, and allow the Error Reporter to instantiate its own execution context.

### Detail

We maintain the singleton interface and delegate to an instance, so these changes are fully backwards compatible. `IsolatedExecutionState` now has two classes, one for thread-specific isolation, and the other for fiber-specific isolation. We instantiate the appropriate one based on the isolation level.

### Checklist

* [ ] Tests are added or updated if you fix a bug or add a feature. (N/A)
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. (N/A)

### DIFF
--- a/activesupport/lib/active_support/isolated_execution_state.rb
+++ b/activesupport/lib/active_support/isolated_execution_state.rb
@@ -1,14 +1,11 @@
 # frozen_string_literal: true
 
 module ActiveSupport
-  module IsolatedExecutionState # :nodoc:
-    @isolation_level = nil
-
-    Thread.attr_accessor :active_support_execution_state
-    Fiber.attr_accessor :active_support_execution_state
+  class IsolatedExecutionState # :nodoc:
+    @isolation_level = :thread
 
     class << self
-      attr_reader :isolation_level, :scope
+      attr_reader :isolation_level
 
       def isolation_level=(level)
         return if level == @isolation_level
@@ -17,56 +14,77 @@ module ActiveSupport
           raise ArgumentError, "isolation_level must be `:thread` or `:fiber`, got: `#{level.inspect}`"
         end
 
-        clear if @isolation_level
-
-        @scope =
-          case level
-          when :thread; Thread
-          when :fiber; Fiber
-          end
-
+        clear if @instance
         @isolation_level = level
+        @instance = nil
       end
 
-      def unique_id
-        self[:__id__] ||= Object.new
-      end
-
-      def [](key)
-        if state = @scope.current.active_support_execution_state
-          state[key]
+      def instance
+        @instance ||= case isolation_level
+        when :thread; ThreadIsolatedExecutionState.new
+        when :fiber; FiberIsolatedExecutionState.new
         end
       end
 
-      def []=(key, value)
-        state = (@scope.current.active_support_execution_state ||= {})
-        state[key] = value
-      end
+      delegate :[], :[]=, :key?, :delete, :clear, :context, :share_with, :scope, to: :instance
+    end
 
-      def key?(key)
-        @scope.current.active_support_execution_state&.key?(key)
-      end
-
-      def delete(key)
-        @scope.current.active_support_execution_state&.delete(key)
-      end
-
-      def clear
-        @scope.current.active_support_execution_state&.clear
-      end
-
-      def context
-        scope.current
-      end
-
-      def share_with(other)
-        # Action Controller streaming spawns a new thread and copy thread locals.
-        # We do the same here for backward compatibility, but this is very much a hack
-        # and streaming should be rethought.
-        context.active_support_execution_state = other.active_support_execution_state.dup
+    def [](key)
+      if (state = context.active_support_execution_state)
+        state[key]
       end
     end
 
-    self.isolation_level = :thread
+    def []=(key, value)
+      state = (context.active_support_execution_state ||= {})
+      state[key] = value
+    end
+
+    def key?(key)
+      context.active_support_execution_state&.key?(key)
+    end
+
+    def delete(key)
+      context.active_support_execution_state&.delete(key)
+    end
+
+    def clear
+      context.active_support_execution_state&.clear
+    end
+
+    def context
+      scope.current
+    end
+
+    def share_with(other)
+      # Action Controller streaming spawns a new thread and copy thread locals.
+      # We do the same here for backward compatibility, but this is very much a hack
+      # and streaming should be rethought.
+      context.active_support_execution_state = other.active_support_execution_state.dup
+    end
+  end
+
+  class ThreadIsolatedExecutionState < IsolatedExecutionState
+    Thread.attr_accessor :active_support_execution_state
+
+    def scope
+      Thread
+    end
+
+    def isolation_level
+      :thread
+    end
+  end
+
+  class FiberIsolatedExecutionState < IsolatedExecutionState
+    Fiber.attr_accessor :active_support_execution_state
+
+    def scope
+      Fiber
+    end
+
+    def isolation_level
+      :fiber
+    end
   end
 end
--- a/activesupport/lib/active_support/execution_context.rb
+++ b/activesupport/lib/active_support/execution_context.rb
@@ -1,53 +1,63 @@
 # frozen_string_literal: true
 
 module ActiveSupport
-  module ExecutionContext # :nodoc:
-    @after_change_callbacks = []
+  class ExecutionContext # :nodoc:
     class << self
-      def after_change(&block)
-        @after_change_callbacks << block
+      def instance
+        @instance ||= new(IsolatedExecutionState)
       end
 
-      # Updates the execution context. If a block is given, it resets the provided keys to their
-      # previous value once the block exits.
-      def set(**options)
-        options.symbolize_keys!
-        keys = options.keys
-
-        store = self.store
-
-        previous_context = keys.zip(store.values_at(*keys)).to_h
-
-        store.merge!(options)
-        @after_change_callbacks.each(&:call)
-
-        if block_given?
-          begin
-            yield
-          ensure
-            store.merge!(previous_context)
-            @after_change_callbacks.each(&:call)
-          end
-        end
-      end
-
-      def []=(key, value)
-        store[key.to_sym] = value
-        @after_change_callbacks.each(&:call)
-      end
-
-      def to_h
-        store.dup
-      end
-
-      def clear
-        store.clear
-      end
-
-      private
-        def store
-          IsolatedExecutionState[:active_support_execution_context] ||= {}
-        end
+      delegate :after_change, :set, :to_h, :clear, :[]=, to: :instance
     end
+
+    def initialize(isolated_execution_state)
+      @isolated_execution_state = isolated_execution_state
+      @after_change_callbacks = []
+    end
+
+    def after_change(&block)
+      @after_change_callbacks << block
+    end
+
+    def set(**options)
+      options.symbolize_keys!
+      keys = options.keys
+
+      previous_context = keys.zip(store.values_at(*keys)).to_h
+
+      store.merge!(options)
+      notify_callbacks
+
+      if block_given?
+        begin
+          yield
+        ensure
+          store.merge!(previous_context)
+          notify_callbacks
+        end
+      end
+    end
+
+    def []=(key, value)
+      store[key.to_sym] = value
+      notify_callbacks
+    end
+
+    def to_h
+      store.dup
+    end
+
+    def clear
+      store.clear
+    end
+
+    private
+      def notify_callbacks
+        @after_change_callbacks.each(&:call)
+      end
+
+      def store
+        @isolated_execution_state[:active_support_execution_context] ||= {}
+      end
   end
 end


tenderlove (approved):

activesupport/lib/active_support/isolated_execution_state.rb
```diff
@@ -1,14 +1,11 @@
 # frozen_string_literal: true
 
 module ActiveSupport
-  module IsolatedExecutionState # :nodoc:
-    @isolation_level = nil
-
-    Thread.attr_accessor :active_support_execution_state
-    Fiber.attr_accessor :active_support_execution_state
+  class IsolatedExecutionState # :nodoc:
```
jhawthorn, line 4: Non-blocking, but would it make sense to make this the instance, ie...

```
class IsolatedExecutionStateContainer # not sure about naming 😅
   ...
end

IsolatedExecutionState = IsolatedExecutionStateContainer.new
```

That way we'd avoid the extra method calls from delegation on this heavily used path

activesupport/lib/active_support/isolated_execution_state.rb
```diff
@@ -17,56 +14,77 @@ def isolation_level=(level)
...
-        self[:__id__] ||= Object.new
-      end
-
-      def [](key)
-        if state = @scope.current.active_support_execution_state
-          state[key]
+      def instance
+        @instance ||= case isolation_level
+        when :thread; ThreadIsolatedExecutionState.new
+        when :fiber; FiberIsolatedExecutionState.new
```
byroot, line 25:
```suggestion
        when :thread then ThreadIsolatedExecutionState.new
        when :fiber then FiberIsolatedExecutionState.new
```

byroot (changes requested):
I'm afraid I don't quite understand where you are trying to get at. And I understand that this is an initial refactoring for another PR that's supposed to be opened after, but the problem is that IsolatedExecutionState is a very huge hotspot, so would like to understand the benefits before adding some overhead.

adrianna-chang-shopify:
👋 Hey @byroot ! Yeah I'm happy to provide more context here and work through our use case with you.

Here's a branch on the Shopify fork with the full set of changes: https://github.com/rails/rails/compare/main...Shopify:rails:ac-isolated-ex-state-fiber-storage?expand=1 

I mentioned the error reporter being our use case, but we're actually building this out to accommodate setting metadata on the new structured event reporter that's being built. We'd like to follow suit with what the `ErrorReporter` is doing and leverage `#set_context`.

Here's an example of the behaviour we'd like to see:
```ruby
reporter = Rails.error 
reporter.set_context(foo: "bar")

# An enumerator, async query, etc.
fiber = Fiber.new do
  reporter.error(SomeError.new) # We want the context from the parent thread here

  reporter.set_context(foo: "baz") # Update the context
end
fiber.resume

reporter.error(SomeError.new) # context should still be <foo: "bar">
```

So notably:
- Child fibers / threads inherit context from the parent
- Context in the main thread is not mutated by children

Internally at Shopify, we have a custom singleton that is using fiber storage to store context that we incorporate into logs that are emitted (things like `shop_id`, `request_id`), etc. My goal is to eventually be able to leverage the `ExecutionContext` for storing this metadata, but using a thread or fiber accessor doesn't give us the behaviour we want.

Reiterating from the PR description:
> Originally, this work began with changes to support a :fiber_storage option on the IsolatedExecutionState singleton, but we realized this created issues with other aspects of the framework that rely on IsolatedExecutionState, such as DB connections and CurrentAttributes.

Simply adding support for fiber storage isn't enough, because we don't want to change behaviour around e.g. DB connections and the locality there. So this first change would allow us to specify a dedicated `ExecutionContext` for the error / event reporter, with the `isolation_level` configured to use fiber storage, without impacting the rest of the app.

byroot:
Ok, I suppose I somewhat understand what you are trying to do. So I would understand why you'd change the `ExecutionContext`, but the `IsolatedExceptionState` should never be "inherited" as you noticed.

Perhaps there are some keys in it you'd want inherited, but then it probably means another "store" should be introduced to distinguish between what is fully private and what is "inheritable". But then you open the huge can of worm of "inheriting" mutable state. 

I also don't think relying on Fiber.storage is correct, because while it somewhat works when you spawn a new fiber, it doesn't help at all when spawning a thread or pushing work on a thread pool.

adrianna-chang-shopify:
> Perhaps there are some keys in it you'd want inherited, but then it probably means another "store" should be introduced to distinguish between what is fully private and what is "inheritable". 

Is there an alternative here where `ExecutionContext` could be configured to use a different store (not `IsolatedExecutionState`) that would allow inheritable state (either by means of Fiber storage or otherwise)..? I acknowledge that our use case is specific to the error / event reporter wanting to leverage `ExecutionContext` and have this specific behaviour, and that "fiber storage" as an option on `IsolatedExecutionState` doesn't really make sense in other contexts.

> But then you open the huge can of worm of "inheriting" mutable state.

Yeah, as you may have noticed from the branch I linked, we were getting around that by `dup`ing the underlying store every time we write to it (this is the approach we're taking with the context object in Shopify as well 😅 )

> I also don't think relying on Fiber.storage is correct, because while it somewhat works when you spawn a new fiber, it doesn't help at all when spawning a thread 

Can you elaborate? If context is pushed into `Fiber.storage` from a parent thread, and a child thread is spawned, that context will be inherited, no? I acknowledge your point about the thread pool though, async queries was a bad example (and not actually a use case we care about AFAIK). What we care about is a thread (or Fiber) being spun up in the context of the main thread serving the request, and wanting any context about the request id, shop id, etc. to propagate to logs (or errors) reported from within that thread / fiber.

adrianna-chang-shopify:
> Is there an alternative here where ExecutionContext could be configured to use a different store (not IsolatedExecutionState) that would allow inheritable state

Okay, so to this point, I see a couple of different alternatives:
1) We could allow independent `ExecutionContext`s, but drop all the changes to `IsolatedExecutionState`. Instead, we'd allow injecting our own store, which could have the `Fiber.storage` behaviour we want. This way, Rails doesn't take on the risk of allowing inheritable, mutable state (it remains on us to manage).
2) We build something for metadata tracking on the event reporter that doesn't rely on `ExecutionContext`. It either uses Fiber storage or we allow it to be pluggable with our own thing.

activesupport/lib/active_support/isolated_execution_state.rb
```diff
@@ -17,56 +14,77 @@ def isolation_level=(level)
...
         end
       end
 
-      def []=(key, value)
-        state = (@scope.current.active_support_execution_state ||= {})
-        state[key] = value
-      end
+      delegate :[], :[]=, :key?, :delete, :clear, :context, :share_with, :scope, to: :instance
```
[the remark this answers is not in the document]
tenderlove: I think if we do like @jhawthorn said it should be fine. We can eliminate the delegate and still remove the singleton.

tenderlove:
> I also don't think relying on Fiber.storage is correct, because while it somewhat works when you spawn a new fiber, it doesn't help at all when spawning a thread or pushing work on a thread pool.

Correct, that's why we want to remove the singleton.  The use case we have requires an "isolated environment" that is different from what a normal webserver would want.  In the client library, we want to use the same API that `IsolatedExecutionState` offers, but have a different backing isolation "scheme".

> Ok, I suppose I somewhat understand what you are trying to do. So I would understand why you'd change the `ExecutionContext`, but the `IsolatedExceptionState` should never be "inherited" as you noticed.
> 
> Perhaps there are some keys in it you'd want inherited, but then it probably means another "store" should be introduced to distinguish between what is fully private and what is "inheritable". But then you open the huge can of worm of "inheriting" mutable state.

I'm not really following this.  Pulling out a singleton so that other code can choose a different strategy seems fine?

matthewd:
> the same API that IsolatedExecutionState offers

Stripped of "'global' apparent-singleton that's actually scoped to the current AS::Executor block", I feel like that's basically "a limited-API Hash"? :confused:


> we don't want to change behaviour around e.g. DB connections and the locality there

I also feel like there's a more nuanced conversation to be had here. _Shouldn't_ an enumerator use the same DB connection as its caller, ideally? Wanting that is one of the biggest reasons that we default to thread isolation outside of fiber-based servers.

We knew that a pure, hard-line, delineation on fiber boundaries would have potential to be bumpy [because server-level "strong scope" fibers naturally intermix with indirectly user-initiated "soft scopes" like enumerators], and expected that we'd be able to evolve that as people started using it in anger (and fiber storage became viable as a tool in the toolbox).

byroot:
> _Shouldn't_ an enumerator use the same DB connection as its caller, ideally?

Yes, and that was the primary motivation for the `IsolatedExecutionState` refactoring. Move all this state to true Thread Local while lots of it used to be fiber local (ad hoc `Thread.current` usage).

But if your primary execution primitive is fibers, it become much less clear. And even if you use threads but spawn a fiber, you might want this fiber to be allowed to checkout its own connection to query the DB concurrently.

That's why IMO it's not as simple as this PR make it out to be. I moved everything into `IsolatedExecutionState` as a first step, but if we're talking about sharing (or inheriting) some of that state, I think we need to review everything that is in it and figure out what should be inherited/shared and what should remain private.

tenderlove:
> if we're talking about sharing (or inheriting) some of that state

I don't think that we are talking about this.  AFAIU, the motivation behind this is to enable different code paths to use different isolation contexts outside of the webserver's concurrency model.  Right now we are tightly coupled with the webserver's concurrency model: anything that wants to use global state must adhere to whatever the webserver chose. We've found that's good for database connections (to be very clear I don't want to change that behavior), but it might not be good for other data (e.g. error reporting metadata).

Rather than invent bespoke "isolation context" classes, I think it would be good if we could instantiate new ones that fit the needs of the particular data.

byroot:
> I think it would be good if we could instantiate new ones that fit the needs of the particular data.

I guess I don't understand what you guys are trying to do. Any chance to see a demo script of some sort?

Is the goal to spawn a fiber with its own context from a threaded server?

> I think it would be good if we could instantiate new ones that fit the needs of the particular data.

I don't mind being able to instantiate them, the problem is that ultimately the instance is accessed via a global API, so it doesn't really matter if more than one exist.

I'm just trying to say I don't understand the use case you are trying to enable.

matthewd:
> We've found that's good for database connections (to be very clear I don't want to change that behavior)

I'm saying that we _do_ want to change that behaviour. It's reasonable in Thread mode, but current behaviour in Fiber mode is known to be mediocre. I think it's very possible that we could get a best-of-all-worlds behaviour by [carefully] using Fiber storage. But for that future to be viable, I think we need to retain the hard-won unification of existing state storage.

> Right now we are tightly coupled with the webserver's concurrency model: anything that wants to use global state must adhere to whatever the webserver chose

To me that's the whole of the intention/point of AS::IES: it exists to encapsulate the server's concurrency model. Separated from that, it's just a hash. (Which does not, AFAIK, do anything that deals with inheriting data from outer concurrency boundaries. In the context of the server's concurrency model, I don't think that would make sense.)

Until very recently, we were handling each piece of state separately, in the place it was used... if we need to introduce something of a different shape, I feel like it would be clearer to start it off like that, and then consider moving it towards IES afterwards.

matthewd:
Yeah... I _think_ that the natural inheritance provided by fiber storage, along with manual handling at more "managed" execution boundaries like thread/fiber pools, _might_ allow a future where IES doesn't need to be Thread-vs-Fiber configurable at all to achieve its current goals -- and give equal-to-now behaviour for Threads and better-than-now behaviour for Fibers in the process.

But my view is that that path requires it to maintain the encapsulation of being "a contextually-scoped state bucket you don't have to think about", at minimum, and I suspect even growing a deeper understanding of the scoping needs of the values it carries (inheritance by reference or by copy, or just flushing, etc). 

If we can get it there, then I think IES-proper [the _same_ global state bucket used for AR etc] could end up working great for the error context stuff you're talking about. But ISTM it's likely easier to refactor from a small and specific hand-rolled solution (as we did with previous IES unification) than to have a more fully generic implementation we're trying to fold back in. 🤔 

> So you're suggesting we pivot instead to building what we need for the error reporter, with the inheritance we want, rather than trying to make this fit into ExecutionContext. ie.
>
>> We build something for metadata tracking on the event reporter that doesn't rely on ExecutionContext. It either uses Fiber storage or we allow it to be pluggable with our own thing.

Acknowledging that it would be with a broadly-agreed "for now" -- yeah.

adrianna-chang-shopify:
Going to close this, we'll move forward with a more targeted approach for our context needs. Thanks for the feedback + discussion, everyone! ❤️

PR #35531, opened 2019-03-08
Update documentation on upsert_all so that it is correct for Postgres

boblail:
addresses: #35519

### Summary

In short, MySQL and Sqlite3 allow a record to be both inserted _and_ replaced in the same operation. Postgres (and the SQL-2003 rules for MERGE) do not.

Postgres's rationale seems to be that the operation would be nondeterministic.

I think it's OK for Rails users to have a different experience with this feature depending on their database; but I think you should be able to follow the examples in the docs on any database.

PR #54790, opened 2025-03-20
Test if `insert_all` / `upsert_all` apply correct values when given duplicate identifiers in same call

andyundso:
### Motivation / Background

I recently implemented `insert_all` and `upsert_all` support for the [Microsoft SQL Server adapter](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1312). MSSQL does not support something like `ON DUPLICATE UPDATE`, so I had to resort to `MERGE`, which essentially joins your new data together with the existing data in the table, then you give it a `CASE` statement what should happen if the entry was found in the database or not.

The current implementation passes all of Rails' tests, but I am unsure if our ordering of the source values is correct.

### Detail

This PR adds two additional tests. The first one checks that `insert_all` only applies the first value for a column that is not an identifier. For `upsert_all`, it is checked that only the last given value is applied.

### Additional information

The SQL server adapter uses the test cases by ActiveRecord to verify its functionality. While this PR only really tests if the `ON UPDATE` clause is set correctly on the adapters supported directly by Rails, it will greatly help us in the SQL adapter to see if we do our `MERGE` statement correctly.

### Checklist

### DIFF
--- a/activerecord/test/cases/insert_all_test.rb
+++ b/activerecord/test/cases/insert_all_test.rb
@@ -185,6 +185,16 @@ class InsertAllTest < ActiveRecord::TestCase
     end
   end
 
+  def test_insert_all_only_applies_last_value_when_given_duplicate_identifiers
+    skip unless supports_insert_on_duplicate_skip?
+
+    Book.insert_all [
+      { id: 111, name: "expected_new_name" },
+      { id: 111, name: "unexpected_new_name" }
+    ]
+    assert_equal "expected_new_name", Book.find(111).name
+  end
+
   def test_skip_duplicates_strategy_does_not_secretly_upsert
     skip unless supports_insert_on_duplicate_skip?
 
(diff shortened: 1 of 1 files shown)


andyundso:
interesting, so the new tests fail on PostgreSQL, but pass on sqlite and MySQL.

After looking some more into the source code, I would have expected that sqlite should also fail, since sqlite and PostgreSQL both support `supports_insert_on_conflict?`. not quite sure how to resolve this situation.

Edouard-chin:
Thanks for the PR. You'll have to skip the `upsert_all` test for postgres as it's not supported. You have a better explanation of why on this PR: https://github.com/rails/rails/pull/35531

andyundso:
Hi @Edouard-chin, many thanks for the feedback. I now adjusted the `upsert_all` test to be skipped when running PostgreSQL.

PR #54796, opened 2025-03-21
Allow having a default of `false` for `support_unencrypted_data`

jordan-brough:
Adds a new configuration option: `ActiveRecord::Encryption.config.support_unencrypted_data_default_value`.

Previously, if you had `ActiveRecord::Encryption.config.support_unencrypted_data` enabled at all, you had to have a default of `support_unencrypted_data=true` for all encrypted attributes in the app.

When transitioning existing attributes over to encrypted attributes, it's helpful to be able to allow unencrypted attributes where needed, but still have a default of `support_unencrypted_data=false`.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Add `support_unencrypted_data_default_value` configuration option for ActiveRecord::Encryption
+
+    Adds a new configuration option that allows changing the default behavior for model-level
+    `support_unencrypted_data` configuration when `ActiveRecord::Encryption.config.support_unencrypted_data`
+    is true. When global `support_unencrypted_data` is false, unencrypted data is still never
+    supported, regardless of default value or model-level settings.
+
+    Previously, if you had `ActiveRecord::Encryption.config.support_unencrypted_data` enabled
+    at all, you had to have a default of `support_unencrypted_data=true` for all encrypted
+    attributes in the app.
+
+    *Jordan Brough*
+
 *   Allow bypassing primary key/constraint addition in `implicit_order_column`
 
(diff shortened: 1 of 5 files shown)


matthewd:
I see the goal here, but I find the two similarly-named booleans confusing -- and I think that flows through to the documentation.

What do you think about `config.support_unencrypted_data = :optional`? (or something? I don't feel like I've come up with the right word yet)

jordan-brough:
Yup that makes sense.

What I think would be ideal is:
* Get rid of `ActiveRecord::Encryption.config.support_unencrypted_data`
* Default encrypted attributes to `support_unencrypted_data: false`
* Allow setting `support_unencrypted_data: true` when needed

Then we'd be down to a single config instead of two (global and model-level).
That'd be a breaking change of course.

Is there a scenario I'm not thinking of where we need the global setting?

jordan-brough:
@matthewd, or anyone: Any thoughts on instead just removing `ActiveRecord::Encryption.config.support_unencrypted_data`?

I think the changelog for that would read like this:

*   Deprecate `ActiveRecord::Encryption.config.support_unencrypted_data`

    By default, encrypted attributes will require encrypted data.
    To permit unencrypted data for an attribute, do the following:
    
    ```ruby
    encrypts :your_attribute, support_unencrypted_data: true
    ```

    Previously, to allow unencrypted data you would set `ActiveRecord::Encryption.config.support_unencrypted_data`
    to `true` and then pass `support_unencrypted_data: false` to all attributes that *shouldn't*  allow
    unencrypted data.

I'm happy to update this PR with that, or to open a separate PR for comparison, if it looks OK.

PR #54797, opened 2025-03-21
Add DSL to enable easier configuration of Reporting-Endpoints HTTP he…

fresh-eggs:
### Motivation / Background

The [Reporting-Endpoints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) response header provides browsers with a list of endpoints that various security headers and features can reference. It instructs them where to send structured reports in accordance with the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API).

From my testing, support for [Reporting-Endpoints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) appears present in Chrome (fully supported), Safari (supports referencing endpoint names in directives but doesn't perform any report batching) and Firefox (behind experimental flag).

The following is a subset of the response headers that are able to leverage [Reporting-Endpoints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) in their `report-to` directives: 
- [Content-Security-Policy (level 3)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/report-to)
- [Cross Origin Opener Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy)
- [Cross Origin Embedder Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy)

### Detail

The DSL makes it possible for application owners to set an application wide set of endpoints and override it per-request:

```ruby
# config/initializers/reporting_endpoints.rb
Rails.application.config.reporting_endpoints do |reporting_endpoints|
  reporting_endpoints.endpoints = {
    "csp-reports": "/csp-reports",
    "other-reports": "https://reporting.endpoint.com"
  }
end
```

```ruby
class MyController < ApplicationController
  reporting_endpoints(only: :index) do |config|
    config.endpoints = {
      "csp-reports": "/csp-reports",
      "other": "https://reporting.endpoint.com"
    }
  end
end
```

This adds first class support for the [Reporting-Endpoints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) header. Since there is no current API that allows hosts to modify the contents of the Report body sent by browsers, most configuration done in the wild relies on URL parameters. Per-request configuration is useful in those cases.

Example from Google Cloud Admin console response headers:
```
reporting-endpoints:    default="/_/CloudConsoleWeb/web-reports?context=<some_base64_encoded_set_of_useful_request_specific_metadata>
```

### Checklist


matthewd:
I don't remember the details, but I think we recently felt this lesson on adding a custom CSP-definition API so early:

> For a technology marked experimental, one or more of the following conditions apply:
> * Its defining specification is likely to change significantly in backwards-incompatible ways (i.e., it may break existing code that relies on the feature).

fresh-eggs:
Yeah I worry about that here also. Would it be more acceptable if we didn't include it in the default middleware stack ?

matthewd:
I'm not sure that helps much.. we still potentially end up with an API that represents a future-wrong version of the header, and have to manage that over a far longer timeframe than browsers. :confused:

> default middleware stack

On that note, if we do add this (or, eventually, _when_), I think the middleware should definitely be shared with CSP. I _suspect_ we might want to do something share-ful with the controller API too. In isolation this looks like a lot of over-build for a flat K/V list (which I'm also feeling hard to convince would often vary across controllers/actions?) -- we have plenty of other headers that we set using simpler technology.

fresh-eggs:
The intent was to avoid coupling this header too closely to CSP since it expects to serve a variety of response headers. CSP `report-to` directive being only one of many.

Open to moving it elsewhere though!

fresh-eggs:
> which I'm also feeling hard to convince would often vary across controllers/actions?

The specific use case in mind is being able to add controller / action specific metadata to the URL parameters for the reporting endpoints.

At this time, user-agents are in control of the body of reports sent. To my knowledge there are no browser APIs available that allow servers to include custom metadata. This is regularly accomplished in the wild by including URL parameters for the reporting endpoints served with a document.

It would be useful to enable action specific URL parameters to denote documents with differing types of compliance requirements (ex: `?pci=true`) or append request specific metadata to further enable debugging.

yoavweiss:
Hey! I just wanted to jump in and say that MDN's characterization of Reporting-Endpoints seems off. The feature is [supported](https://wpt.fyi/results/reporting/cross-origin-reports-isolated.https.sub.html?label=experimental&label=master&aligned) in both Chromium and Safari. I [filed an issue](https://github.com/mdn/browser-compat-data/issues/26361) on the relevant repo.

PR #54803, opened 2025-03-23
Add json_key option to customize attribute names in serialization

heka1024:
### Motivation / Background

ActiveModel's serialization currently doesn't provide a built-in way to customize JSON key names. When working with external APIs or JavaScript frontends that use camelCase naming conventions, developers must manually transform attribute names or implement custom serializers.

Fixes the need for custom serializers or manual transformations when Rails models use snake_case but need to produce camelCase JSON.

### Detail

This Pull Request adds a new `json_key` option to the `attribute` method in ActiveModel. This allows developers to specify a custom key name used during JSON serialization while keeping Ruby-style snake_case attribute names in their models.

Example usage:
```ruby
class Person
  include ActiveModel::Serializers::JSON
  
  attribute :first_name, json_key: "firstName"
end

person = Person.new(first_name: "John")
person.as_json # => {"firstName"=>"John"} instead of {"first_name"=>"John"}
```

### Checklist

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,19 @@
+*   Add `json_key` option for attributes to customize serialization key names.
+
+    ```ruby
+    class Person
+      include ActiveModel::Serializers::JSON
+      include ActiveModel::Attributes
+
+      attribute :first_name, json_key: "firstName"
+    end
+
+    person = Person.new(first_name: "John")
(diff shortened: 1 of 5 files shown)


byroot:
I'd rather not complexify Active Model serialization.

The model isn't the right place to do this sort of things anyway, as generally you have more than one representation for a model, it's much better to go with a presentation layer. e.g. libraries like https://rubygems.org/gems/alba etc.

PR #54809, opened 2025-03-25
Introduce `capture_error_reports`

andrewn617:
### Motivation / Background

I want to use `assert_error_reported` to test that an error is raised a certain number of times. For example, when testing retry logic for an external dependency.

### Detail

~I introduced a `times` option for this assertion. If it is supplied, we assert that there are an equal number of reports and then return them in an array.~

I introduced `capture_error_reports` which captures all the reports from that passed block that match the expected exception. Then you can do assertions on those reports.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/testing/error_reporter_assertions.rb
+++ b/activesupport/lib/active_support/testing/error_reporter_assertions.rb
@@ -102,6 +102,23 @@ module ActiveSupport
           assert(false, message)
         end
       end
+
+      # Captures reported errors from within the block that match the given
+      # error class.
+      #
+      #   reports = capture_error_reports(IOError) do
+      #     Rails.error.report(IOError.new("Oops"))
+      #     Rails.error.report(IOError.new("Oh no"))
+      #     Rails.error.report(StandardError.new)
+      #   end
+      #
+      #   assert_equal 2, reports.size
+      #   assert_equal "Oops", reports.first.error.message
+      #   assert_equal "Oh no", reports.last.error.message
+      def capture_error_reports(error_class = StandardError, &block)
+        reports = ErrorCollector.record(&block)
+        reports.select { |r| error_class === r.error }
+      end
     end
   end
 end
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,22 @@
+*   Introduce ActiveSupport::Testing::ErrorReporterAssertions#capture_error_reports
+
+    Captures all reported errors from within the block that match the given
+    error class.
+
+    ```ruby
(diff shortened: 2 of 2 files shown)


byroot:
Seems a bit too ad hoc to me. I think it would be better to give a helper that returns you all reported errors, and then you can do all the assertions you want on it.

activesupport/lib/active_support/testing/error_reporter_assertions.rb
```diff
@@ -102,6 +102,26 @@ def assert_error_reported(error_class = StandardError, &block)
...
+      #     Rails.error.report(IOError.new("Oh no"))
+      #     Rails.error.report(StandardError.new)
+      #   end
+      #
+      #   assert_equal 2, reports.size
+      #   assert_equal "Oops", reports.first.error.message
+      #   assert_equal "Oh no", reports.last.error.message
+      def capture_error_reports(error_class = StandardError, &block)
+        reports = ErrorCollector.record do
+          _assert_nothing_raised_or_warn("capture_error_reports", &block)
```
andrewn617, line 120: @byroot Wasn't sure if I should keep or get rid of this since this is not really an assertion anymore. Do you have a preference?

activesupport/lib/active_support/testing/error_reporter_assertions.rb
```diff
@@ -102,6 +102,26 @@ def assert_error_reported(error_class = StandardError, &block)
...
+      #
+      #   assert_equal 2, reports.size
+      #   assert_equal "Oops", reports.first.error.message
+      #   assert_equal "Oh no", reports.last.error.message
+      def capture_error_reports(error_class = StandardError)
+        reports = ErrorCollector.record do
+          yield
+        end
+
+        reports.select { |r| error_class === r.error }
```
byroot, line 123: protip: 
```suggestion
        reports.grep(error_class)
```

activesupport/lib/active_support/testing/error_reporter_assertions.rb
```diff
@@ -102,6 +102,26 @@ def assert_error_reported(error_class = StandardError, &block)
...
+      #     Rails.error.report(StandardError.new)
+      #   end
+      #
+      #   assert_equal 2, reports.size
+      #   assert_equal "Oops", reports.first.error.message
+      #   assert_equal "Oh no", reports.last.error.message
+      def capture_error_reports(error_class = StandardError)
+        reports = ErrorCollector.record do
+          yield
+        end
```
byroot, line 121:
```suggestion
      def capture_error_reports(error_class = StandardError, &block)
        reports = ErrorCollector.record(&block)
```

PR #54810, opened 2025-03-25
Read generator options properly

brunoprietog:
### Motivation / Background

This Pull Request has been created because #54021 is not working.

### Detail

This Pull Request changes the way of reading the app generator options.

### Additional information

cc @dhh

### Checklist

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -494,7 +494,7 @@ module Rails
       def javascript_gemfile_entry
         return if options[:skip_javascript]
 
-        if options[:javascript] == "importmap"
+        if using_importmap?
           GemfileEntry.floats "importmap-rails", "Use JavaScript with ESM import maps [https://github.com/rails/importmap-rails]"
         else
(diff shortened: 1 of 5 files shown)


railties/lib/rails/generators/app_base.rb
```diff
@@ -388,6 +388,10 @@ def skip_asset_pipeline? # :doc:
         options[:skip_asset_pipeline]
       end
 
+      def javascript?
+        !options.skip_javascript? && !options.api?
+      end
```
skipkayhil, line 393: if `options.api?` is true then `options.skip_javascript?` should also be true

https://github.com/rails/rails/blob/9f39c019243138d73bb265ec32da9aee26b2c18f/railties/lib/rails/generators/rails/app/app_generator.rb#L333-L336

Can you also add a test in `app_generator_test.rb`?

brunoprietog: True, good catch. I've removed that verification. I'll add a test soon

eileencodes:
Does this need a backport?

brunoprietog:
Thanks for merging! No, This was only in main yet.

PR #54811, opened 2025-03-25
Add multi db support for ActiveStorage

rubyleth:
### Motivation / Background

ActiveStorage currently does not handle multi db support, as mentioned in this [discussion](https://discuss.rubyonrails.org/t/activestorage-multi-db-support/80123/2).

This Pull Request has been created because I work a on project using three databases on the same application. Two of those databases use ActiveStorage and I need therefore this feature.

### Detail

- I moved most of the app folders and files in the lib folder since most of the classes are now created dynamically through meta-programming ;
- I left the whole thing as an engine to properly and easily handle migrations and javascript related files ;
- I added a new initializer which is run after the app has been initialized to create the blob, attachment and variant_record needed models, controllers and jobs based on app config (which provides the db name and connection class for each case) ;
- since the dummy test application is run in two different configurations for single or multi db, I created two separated rake tasks to be run separately for testing ;
- the services are only tested in the single_db tests since they do not directly involve named blob, attachment or variant_record classes ;
- I tested my solution on local apps using single db and multi db, both worked.

### Additional information

I made the choice to make the old blob/attachment/variant classes as modules included in the dynamically generated classes. I could also have used inheritance but I find it more readable and flexible this way.

One test keeps failing:
- `eager_loading_has_many_attached_records` for `variant_record_test`, 12 requests are performed instead of 10, I cannot manage to fix it.

The rotating image related test in vips_test.rb was already failing on main branch.

I did not change the CHANGELOG yet since I wanted to have a feedback on the philosophy I followed here first.
If we agree on the solution I used here, I will also adapt test failing in other rails gems since they break on CI.

### Checklist


eileencodes:
> Do you have an idea of the time it will take for the whole process of reviewing/testing/adapting ?

I can't really give an estimate on this, since I have other responsibilities on my plate and this is a large change. But it would speed things up if you can make an app using active storage before this change, then "upgrade" it to this branch and implement multiple databases.

thierry-tranchant:
> > Do you have an idea of the time it will take for the whole process of reviewing/testing/adapting ?
> 
> I can't really give an estimate on this, since I have other responsibilities on my plate and this is a large change. But it would speed things up if you can make an app using active storage before this change, then "upgrade" it to this branch and implement multiple databases.

Alright I'm on it, I will provide you with a testing app asap !

rubyleth:
@eileencodes we did not have enough time to finish this work regarding our last business priorities so we had to proceed differently and we don't need this multi-db adaptation anymore
As it's no longer my company's priority, I'm afraid I won't have time to properly conclude this PR, and my free time is already dedicated to other code topics at the moment :/

kapso:
@rubyleth thanks for your work on this, is this going to be part of 8.2? and if thats the case, when is 8.2 expected?

Issue #54805, opened 2025-03-24
`structure_load` only appends `extra_flags`, does not allow to prepend them


AliceLoeser:
### Steps to reproduce

When using `structure_load_flags`, if the flags contain the `--set` command to change run-time configuration parameters, those will not be taken into account by `psql` if they are passed after the `--file` parameter. With the way `structure_load` is written, all `extra_args` are prepended, there is no option to append them, making it impossible to update an extension setting.

In our case we need to change a specific postgres extension setting to make the load command work, which forces us to monkey patch the `structure_load` like so:

```
  def structure_load(filename, extra_flags)
    # The change is this monkey patch is only to prepend extra_flags instead of appending them
    args = ['--set', 'ON_ERROR_STOP=1', '--quiet', '--no-psqlrc', '--output', File::NULL, '--file', filename]
    args = Array(extra_flags) + args if extra_flags
    args << db_config.database
    run_cmd('psql', args, 'loading')
  end
```

### Expected behavior

I can chose the order of `extra_flags` relative to the `--file` parameter. 

### Actual behavior

`extra_flags` are always set after `--file`, and will not always be taken into account by the `psql` command.

### System configuration

**Rails version**: Rails 7.2.2.1

**Ruby version**: Ruby 3.3.7

### Options to fix this behavior

1. Update `structure_load` to prepend the `extra_flags`: I believe this could cause the opposite of my current issue to other users, I wouldn't recommend it
2. Add a new argument to `structure_load`, with default set to nil (no breaking change), that would be `extra_flags_prepended` or something like that, and would be set the same way `structure_load_flags` is 
3. Change the structure of `structure_load`'s `extra_flags`: now instead of being an array/a hash with array value, it would be a 2-dimensional array, `[[flags_prepended], [flags_appended]]`. Breaking change but an easy fix.

I'll be happy to write the fix, I'd like to know what would be the best option to do so beforehand.

AliceLoeser:
Hello @Edouard-chin I don't doubt that for some flags it works properly with `--set`. 
In this specific case I am updating a custom postgres extension setting to disable it, like so: 

```
ActiveRecord::Tasks::DatabaseTasks.structure_load_flags = ['-c', 'SET my_extension.level=off;'].freeze
```

And it won't be taken into account by the `psql` command if the flag is set before the `--file` argument.

I've also tested to launch the command manually in my terminal with a simple `select`: 
```
[...]

PR #54813, opened 2025-03-25
With postgres adapter, prepend `structure_load_flags` instead of appending them.

AliceLoeser:
When using postgres adapter and the `structure_load_flags` options, the extra flags were appended instead of prepended to the default ones, causing the `psql` command to ignore some of the extra flags.

Now, the default arguments `args` are appended to the `extra_flags` instead of the opposite.

Example of `extra_flags` that were ignored by postgres:

    ActiveRecord::Tasks::DatabaseTasks.structure_load_flags = ['-c', 'SET my_extension.level=off;'].freeze

Now, they will be taken into account.

Resolves issue #54805. 

### Motivation / Background

This Pull Request has been created because of the following issue: https://github.com/rails/rails/issues/54805

### Detail

This Pull Request changes how `structure_load` uses the `extra_flags` argument, prepending it instead of appending it (for postgres only). 

### Additional information

Alternative solutions considered:

1. Add a new argument to structure_load, with default set to nil (no breaking change), that would be extra_flags_prepended or something like that, and would be set the same way structure_load_flags is
2. Change the structure of structure_load's extra_flags: now instead of being an array/a hash with array value, it would be a 2-dimensional array, [[flags_prepended], [flags_appended]]. Breaking change but an easy fix.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/tasks/postgresql_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/postgresql_database_tasks.rb
@@ -78,8 +78,9 @@ module ActiveRecord
       end
 
       def structure_load(filename, extra_flags)
-        args = ["--set", ON_ERROR_STOP_1, "--quiet", "--no-psqlrc", "--output", File::NULL, "--file", filename]
+        args = ["--set", ON_ERROR_STOP_1, "--quiet", "--no-psqlrc", "--output", File::NULL]
         args.concat(Array(extra_flags)) if extra_flags
+        args.concat(["--file", filename])
         args << db_config.database
         run_cmd("psql", args, "loading")
       end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Prepend `extra_flags` in postgres' `structure_load`
+
+    When specifying `structure_load_flags` with a postgres adapter, the flags
+    were appended to the default flags, instead of prepended.
+    This caused issues with flags not being taken into account by postgres.
+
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/tasks/postgresql_database_tasks.rb
```diff
@@ -79,7 +79,7 @@ def structure_dump(filename, extra_flags)
 
       def structure_load(filename, extra_flags)
         args = ["--set", ON_ERROR_STOP_1, "--quiet", "--no-psqlrc", "--output", File::NULL, "--file", filename]
-        args.concat(Array(extra_flags)) if extra_flags
+        args = Array(extra_flags).concat(args) if extra_flags
```
p8, line 82: Instead of prepending the `extra_flags` afterwards, maybe start with the `extra_flags` and append the default flags?
```ruby
        args = Array(extra_flags)
        args.concat(["--set", ON_ERROR_STOP_1, "--quiet", "--no-psqlrc", "--output", File::NULL, "--file", filename])
```
Or:
```ruby
        args = Array(extra_flags) + ["--set", ON_ERROR_STOP_1, "--quiet", "--no-psqlrc", "--output", File::NULL, "--file", filename]
```

AliceLoeser: I like the last option best. Will fix it today 👍

AliceLoeser: @p8 All good! I tested it locally on my problematic PR as well, works like a charm.

PR #54821, opened 2025-03-27
Add docs for reading values written by `Rails.cache.increment`

ghiculescu:
This corresponds to the docs for writing a value, where `raw: true` is required by some cache stores.


byroot:
They'll be strings though, no?

byroot:
Making me think, a `read_counter` (or something) method might be better.

ghiculescu:
@byroot yeah I like that, wdyt: https://github.com/rails/rails/pull/54855

PR #54829, opened 2025-03-28
Allow using aliases for unions in `from` clause

fatkodima:
I am optimizing one query in my project and rewrote it using `UNION` instead of `OR`.
 
The new query was like this: `SomeModel.from(union, :some_models)`. It produced an error `missing FROM-clause entry for table "some_models"` in PostgreSQL. The problem is that `#from` happily accepts unions, but silently ignores passed aliases.

### DIFF
--- a/activerecord/test/cases/relations_test.rb
+++ b/activerecord/test/cases/relations_test.rb
@@ -259,6 +259,18 @@ class RelationTest < ActiveRecord::TestCase
     assert_equal relation.to_a, Comment.select("a.*").from(relation, :a).to_a
   end
 
+  unless current_adapter?(:SQLite3Adapter)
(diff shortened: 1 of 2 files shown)


activerecord/test/cases/relations_test.rb
```diff
@@ -256,6 +256,18 @@ def test_finding_with_subquery_with_eager_loading_in_from
     assert_equal relation.to_a, Comment.select("a.*").from(relation, :a).to_a
   end
 
+  unless current_adapter?(:SQLite3Adapter)
```
fatkodima, line 259: Using unions currently produces invalid queries for sqlite - it wraps them in extra parentheses.

PR #54836, opened 2025-03-30
[ActiveRecord] Introduce with_default_isolation_level

kirs:
When migrating a larger application to a new database isolation level, you might want to start enforcing default isolation level per area (or per a base controller) in your app.

Having something like `#with_default_isolation_level` that takes a new level and a block could drastically help with such migrations.

Example:

```ruby
class ApiV2Controller < ApplicationController
  around_action :set_isolation_level

  def set_isolation_level
    Product.with_default_isolation_level(:read_committed) do
      yield
    end
  end
end

# forces all controllers that subclass from ApiV2Controller to start getting new isolation level
```

I've opted to store the default on the connection object rather than a thread/fiber variable, to make this work nicely with multi-database apps.

cc @byroot @rafaelfranca

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
@@ -620,6 +620,7 @@ module ActiveRecord
       end
 
       def within_new_transaction(isolation: nil, joinable: true)
+        isolation ||= @connection.pool.default_isolation_level
         @connection.lock.synchronize do
           transaction = begin_transaction(isolation: isolation, joinable: joinable)
           begin
--- a/activerecord/lib/active_record/transactions.rb
+++ b/activerecord/lib/active_record/transactions.rb
@@ -234,6 +234,25 @@ module ActiveRecord
         end
       end
 
+      # Makes all transactions initiated within the block use the isolation level
+      # that you set as the default. Useful for gradually migrating apps onto new isolation level.
+      def with_default_isolation_level(isolation_level, &block)
+        if current_transaction.open?
+          raise ActiveRecord::TransactionIsolationError, "cannot set default isolation level while transaction is open"
+        end
+
+        old_level = connection_pool.default_isolation_level
+        connection_pool.default_isolation_level = isolation_level
+        yield
+      ensure
+        connection_pool.default_isolation_level = old_level
+      end
+
+      # Returns the default isolation level for the connection pool, set earlier by #with_default_isolation_level.
+      def default_isolation_level
+        connection_pool.default_isolation_level
+      end
+
       # Returns a representation of the current transaction state,
       # which can be a top level transaction, a savepoint, or the absence of a transaction.
       #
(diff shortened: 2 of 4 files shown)


activerecord/lib/active_record/transactions.rb
```diff
@@ -234,6 +234,12 @@ def transaction(**options, &block)
         end
       end
 
+      def with_default_isolation_level(isolation_level, &block)
+        with_connection do |connection|
```
byroot, line 238: Eagerly checking out the connection isn't ideal. I think we should try not to do that.

Ultimately we only need to know what the default isolation is when starting the transaction.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -365,6 +365,22 @@ def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
         # rollbacks are silently swallowed
       end
 
+      attr_reader :default_isolation_level
```
byroot, line 368: I think the proper place to store this is on the DbConfig, with an `IsolatedExcutionState` override.

byroot: > Do you think we could store it like `IsolatedExecutionState["activerecord_default_isolation_level_#{connection_specification_name}"]` or so?

That's what I think we should do, yes.

kirs: @byroot I've started implementing it via `connection_specification_name` but then I've realized that for some reason it points to the same thing for two models that use different connection:

```
# uses sharded connection
shopify(dev):002> Shop.connection_specification_name
=> "ActiveRecord::Base"
# uses unsharded
shopify(dev):004> ShopRoute.connection_specification_name
=> "ActiveRecord::Base"
```

Which makes me think that I should be using `connection_pool.db_config.name` as that reference instead.

```
shopify(dev):004> Shop.connection_pool.db_config.name
=> "shard_0"
shopify(dev):005> ShopRoute.connection_pool.db_config.name
=> "primary"
```

byroot (approved):

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -620,6 +620,7 @@ def rollback_transaction(transaction = nil)
       end
 
       def within_new_transaction(isolation: nil, joinable: true)
+        isolation ||= @connection.pool.default_isolation_level
```
chaadow, line 623: I think to properly test this line ( specifically `||=` ) we should test this case : 
```ruby
 Tag.with_default_isolation_level(:read_committed) do
          assert_equal :read_committed, Tag.default_isolation_level
          Tag.transaction(isolation: :repeatable_read) do
            Tag.create!(name: "jon")
          end
        end
```
and that in the AS event it should match `BEGIN ISOLATION LEVEL REPEATABLE READ`

@byroot @kirs if you think this can be valuable, I can quickly create a PR for it

vipulnsward:
Needs changelog?

PR #54840, opened 2025-03-31
Deprecate ActionController.escape_json_responses= at the method

zzak:
Follow up to #54643

Instead of emitting the deprecation in an initializer, we should emit the deprecation when calling the writer method.

In order to reduce warnings when users are updating their apps to use the new default `true`, we check the value before emitting.

This includes the class attribute method as well.

In order to set the value for the internal attribute, we need to find out the name of the internal setter[^1].

In the future, it would be nice if `class_attribute` and `mattr_*`, etc had a way to easily deprecate parts of the methods being defined, or provide a callback (so we can emit warnings based on the value, etc).

/cc @byroot @etiennebarrie 

[^1]: https://github.com/rails/rails/blob/1c094d762d25805cc2d110ee2f7b54b33352e293/activesupport/lib/active_support/core_ext/class/attribute.rb#L95-L102

### DIFF
--- a/actionpack/lib/action_controller/metal/renderers.rb
+++ b/actionpack/lib/action_controller/metal/renderers.rb
@@ -27,9 +27,23 @@ module ActionController
     # Default values are `:json`, `:js`, `:xml`.
     RENDERERS = Set.new
 
+    module DeprecatedEscapeJsonResponses # :nodoc:
+      def escape_json_responses=(value)
+        if value
+          ActionController.deprecator.warn(<<~MSG.squish)
+            Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
+            Set it to `false`, or remove the config.
+          MSG
+        end
+        super
+      end
+    end
+
     included do
       class_attribute :_renderers, default: Set.new.freeze
-      class_attribute :escape_json_responses, instance_accessor: false, default: true
+      class_attribute :escape_json_responses, instance_writer: false, instance_accessor: false, default: true
+
+      singleton_class.prepend DeprecatedEscapeJsonResponses
     end
 
     # Used in ActionController::Base and ActionController::API to include all
--- a/actionpack/lib/action_controller/railtie.rb
+++ b/actionpack/lib/action_controller/railtie.rb
@@ -133,18 +133,5 @@ module ActionController
         ActionController::TestCase.executor_around_each_request = app.config.active_support.executor_around_test_case
       end
     end
-
-    initializer "action_controller.escape_json_responses_deprecated_warning" do
-      config.after_initialize do
-        ActiveSupport.on_load(:action_controller) do
-          if ActionController::Base.escape_json_responses
-            ActionController.deprecator.warn(<<~MSG.squish)
-              Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
-              Set it to `false` or use `config.load_defaults(8.1)`.
-            MSG
-          end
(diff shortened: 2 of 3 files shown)


actionpack/lib/action_controller/metal/renderers.rb
```diff
@@ -30,6 +30,14 @@ module Renderers
     included do
       class_attribute :_renderers, default: Set.new.freeze
       class_attribute :escape_json_responses, instance_accessor: false, default: true
+
+      redefine_method :escape_json_responses= do |value|
+        ActionController.deprecator.warn(<<~MSG.squish)
+          Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
```
casperisfine, line 36: I think you meant to check if `value` is true?

zzak: Good catch, fixed in [a12a15e](https://github.com/rails/rails/pull/54840/commits/a12a15ef90faf4a96f24bb9bca2ad89204267e82).

actionpack/lib/action_controller/metal/renderers.rb
```diff
@@ -30,6 +30,30 @@ module Renderers
     included do
       class_attribute :_renderers, default: Set.new.freeze
       class_attribute :escape_json_responses, instance_accessor: false, default: true
+
+      redefine_method :escape_json_responses= do |value|
+        if value
+          ActionController.deprecator.warn(<<~MSG.squish)
+            Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
+            Set it to `false` or use `config.load_defaults(8.1)`.
```
ghiculescu, line 38: wdyt about this? https://github.com/rails/rails/pull/54643#discussion_r2022206660

```suggestion
            Set it to `false`, or remove the config.
```

railties/lib/rails/application/configuration.rb
```diff
@@ -357,7 +357,9 @@ def load_defaults(target_version)
           self.yjit = !Rails.env.local?
 
           if respond_to?(:action_controller)
-            action_controller.escape_json_responses = false
+            ActionController.deprecator.silence do
+              action_controller.escape_json_responses = false
+            end
```
byroot, line 362: I don't think this is correct, because here `action_controller.escape_json_responses =` isn't the method you defined, it's an `OrderedOption`, the value will be forwarded to `ActionController:Base` from an initializer.

byroot: Also `false` doesn't emit a warning anyway.

actionpack/lib/action_controller/metal/renderers.rb
```diff
@@ -30,6 +30,30 @@ module Renderers
...
+        namespaced_name = :"__class_attr_escape_json_responses"
+        self.class.send(:"#{namespaced_name}=", value)
+      end
+
+      class << self
+        redefine_method :escape_json_responses= do |value|
+          if value
+            ActionController.deprecator.warn(<<~MSG.squish)
+              Setting action_controller.escape_json_responses = true is deprecated and will have no effect in Rails 8.2.
+              Set it to `false` or use `config.load_defaults(8.1)`.
```
ghiculescu, line 50:
```suggestion
              Set it to `false` or remove the config.
```

railties/lib/rails/application/configuration.rb
byroot: This one wasn't addressed.

Issue #53727, opened 2024-11-24
test_mysql using MariaDB raises `Mysql2::Error: SAVEPOINT active_record_1 does not exist (ActiveRecord::StatementInvalid)`


yahonda:
Managed to reproduce CI failure against MariaDB https://buildkite.com/rails/rails/builds/114040#0193581b-e8af-4324-9429-42f56cea8085

### Steps to reproduce

Install MariaDB 11.7.1.
```sql
$ mysql -uroot
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 379
Server version: 11.7.1-MariaDB MariaDB Server

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> select version();
+----------------+
| version()      |
+----------------+
| 11.7.1-MariaDB |
+----------------+
1 row in set (0.001 sec)

MariaDB [(none)]>
```
```ruby
cd activerecord
ARCONN=mysql2 bin/test test/cases/relations_test.rb -n test_multiple_find_or_create_by_within_transactions
```

### Expected behavior
It should pass.

### Actual behavior
```ruby
$ ARCONN=mysql2 bin/test test/cases/relations_test.rb -n test_multiple_find_or_create_by_within_transactions
Using mysql2
Run options: -n test_multiple_find_or_create_by_within_transactions --seed 36869

# Running:

#<Thread:0x00007f2ee63a7b38 /home/yahonda/src/github.com/rails/rails/activerecord/test/cases/relations_test.rb:2497 run> terminated with exception (report_on_exception is true):
/home/yahonda/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/mysql2-0.5.6/lib/mysql2/client.rb:151:in `_query': Mysql2::Error: SAVEPOINT active_record_1 does not exist (ActiveRecord::StatementInvalid)
	from /home/yahonda/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/mysql2-0.5.6/lib/mysql2/client.rb:151:in `block in query'
	from /home/yahonda/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/mysql2-0.5.6/lib/mysql2/client.rb:150:in `handle_interrupt'
[...]

PR #54842, opened 2025-03-31
Use MySQL instead of MariaDB in devcontainer

yahonda:
### Motivation / Background

This pull request uses MySQL container instead of MariaDB one in devcontainer

### Detail

MariaDB was likely chosen because Debian "bookworm" sets it as the default MySQL-compatible database. However, recent changes in MariaDB introduced incompatibilities with MySQL—for example, #53727. This makes MariaDB less suitable as the default for developing the mysql2 and trilogy adapters.

Now that the database runs as a separate container, we are no longer bound by the base OS default. Using the official MySQL container offers a development environment that more closely aligns with the behavior of MySQL itself.

### Additional information
References:
https://lists.debian.org/debian-devel-announce/2016/09/msg00000.html
https://github.com/rails/rails/pull/48891

### Checklist

### DIFF
--- a/.devcontainer/compose.yaml
+++ b/.devcontainer/compose.yaml
@@ -12,7 +12,7 @@ services:
 
     depends_on:
       - postgres
-      - mariadb
+      - mysql
       - redis
       - memcached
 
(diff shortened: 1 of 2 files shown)


zzak:
Isn't this kind of breaking change for people who use the devcontainer? I don't use it so I'm not really sure.

yahonda:
Thanks for the comment.

This pull request is based on my recent experience using Devcontainer for Rails framework development.

This change is specifically intended for working on the Rails framework itself, not for developing Rails applications with Devcontainer. For example, Active Record’s activerecord_unittest database is frequently recreated during test runs, so switching the database engine shouldn’t be a major concern.

PR #54843, opened 2025-03-31
Include cookie name in length calculation

### LINKED ISSUE #54838: ActionDispatch::Cookies::CookieOverflow not raised on cookie overflow
It is possible to generate a session cookie that is silently dropped by the browser because it is too large, but which does not trigger rails's overflow check.

I suspect that this is because rails performs the check on the length of the value of the cookie only, but browsers check the sum of the length of the cookie's name, value, and attributes ( as per [section 6.1 of rfc6265](https://datatracker.ietf.org/doc/html/rfc6265#section-6.1) )

### Steps to reproduce

Put something like this in a rails app

```
def reproduce 
    reset_session
    session[:a] = "a" * 2000 # you will need to adjust this number
end
after_action :reproduce
```

Increase the number until a request to the app generates a  ActionDispatch::Cookies::CookieOverflow

Now slightly decrease the number again until no ActionDispatch::Cookies::CookieOverflow is raised

### Expected behavior

Since the overflow check has passed the browser should accept the latest session cookie

### Actual behavior

The browser will reject the cookie as too large. Any changes made to the session in the last request will have been lost.

In Chrome for example you can check whether the cookie was rejected by opening dev tools, going to the network tab and checking the Set-Cookie response header for a warning icon:

Image

### System configuration

**Rails version**: 
 7.1.3.4

**Ruby version**: 

3.3.0

skipkayhil:
### Detail

Fix #54838 

Browsers add the length of the name and content when validating that a cookie is under 4kb, so this commit updates Rails to do the same.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Include cookie name when calculating maximum allowed size.
+
+    *Hartley McGuire*
+
 *   Implement `must-understand` directive according to RFC 9111.
 
     The `must-understand` directive indicates that a cache must understand the semantics of the response status code, or discard the response. This directive is enforced to be used only with `no-store` to ensure proper cache behavior.
--- a/actionpack/lib/action_dispatch/middleware/cookies.rb
+++ b/actionpack/lib/action_dispatch/middleware/cookies.rb
@@ -610,8 +610,10 @@ module ActionDispatch
         end
 
         def check_for_overflow!(name, options)
-          if options[:value].bytesize > MAX_COOKIE_SIZE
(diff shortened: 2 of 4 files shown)


byroot:
I think this is still not enough, because it's still missing the attributes (e.g. expiration, path, etc)

skipkayhil:
> I think this is still not enough, because it's still missing the attributes (e.g. expiration, path, etc)

I tested in Firefox 138 and Edge 134 and both accepted `cookies[:a] = "a" * 4095` but rejected `cookies[:a] = "a" * 4096`. I'm happy to include the options in the calculation since that's what the RFC suggests, but for how simple this change is I'm not sure it's worth doing.

PR #54844, opened 2025-03-31
Extract system test setup to system_helper

skipkayhil:
Defining these classes in abstract_unit makes running all Action Pack tests slower because they have to download/setup browsers for system tests (even if the tests being run never end up using the system test classes).

This commit moves the system test class definitions into a helper which is only included in the two test files that use them. This will make running non-system Action Pack test files much much faster.

### DIFF
--- a/actionpack/test/abstract_unit.rb
+++ b/actionpack/test/abstract_unit.rb
@@ -515,20 +515,4 @@ module HeadersAssertions
   end
 end
 
-class DrivenByRackTest < ActionDispatch::SystemTestCase
-  driven_by :rack_test
-end
-
-class DrivenBySeleniumWithChrome < ActionDispatch::SystemTestCase
-  driven_by :selenium, using: :chrome
-end
-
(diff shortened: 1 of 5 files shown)


yahonda:
> This will make running non-system Action Pack test files much much faster.

Hi, Do you have any benchmarks?

skipkayhil:
On main:

```
❯ time bin/test test/dispatch/cookies_test.rb
bin/test test/dispatch/cookies_test.rb  1.37s user 0.49s system 108% cpu 1.719 total
```

This branch
```
❯ time bin/test test/dispatch/cookies_test.rb
bin/test test/dispatch/cookies_test.rb  1.20s user 0.37s system 119% cpu 1.317 total
```

Not super scientific but these times were pretty consistent for me when switching between branches. The real speedup is when a driver is downloaded: my first test run took 14 seconds to download a driver even though `cookies_test` doesn't need one.

yahonda:
That makes sense. 

> The real speedup is when a driver is downloaded: my first test run took 14 seconds to download a driver even though cookies_test doesn't need one.

PR #54845, opened 2025-03-31
Fix `create_or_find_by` not rolling back a transaction:

### LINKED ISSUE #54830: find_or_create_by in Rails 7.1 can persist invalid object
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
end

require "active_record"
require "logger"
require "minitest/autorun"

# activate to see sql statement

  create_table :tables
end

class Table < ActiveRecord::Base
  after_save do
    raise_validation_error
  end
end

  def test_find_or_create_by
    assert_equal false, Table.create.persisted?
    assert_equal false, Table.find_or_create_by({}).persisted?
  end
end
```

### Expected behavior

both 'create' and 'find_or_create_by' should not persist the object

### Actual behavior

'find_or_create_by' persist the object which was not the case in Rails 7.0. Looks like it's a very specific case where a RecordInvalid is raised in a after_.. callback. 'create' is now run in a subtransaction that is not rollback

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.3.4

Edouard-chin:
### Motivation / Background

Fix #54830 (This indirectly fix the issue)

### Problem

The `create_or_find_by` method doesn't allow rolling back a transaction. This is a different behaviour `create`. The reason is because we are in the double transaction edge case and the outer transaction doesn't see the rollback.

### Details

```ruby
  after_save do
    raise ActiveRecord::Rollback
  end
end

User.create(name: "John") #=> Correctly rolled back
User.find_or_create_by(name: "John") #=> Correctly rolled back on Rails 7.0, stopped rolling back since commit 023a3eb3c046091a5d52027393a6d29d0576da01
User.create_or_find_by(name: "John") #=> Does not roll back
```

### Solution

We need to be able to know whether the inner transaction succeeded, which is not possible using `.create`, as this method always returns the record (rather than the status of the transaction). https://github.com/rails/rails/blob/1eac7f27748ffa43c5ce91c036c928fc22d4c597/activerecord/lib/active_record/persistence.rb#L39

Using `new` followed by a save is equivalent to `create`, but with `save` we get the transaction return status and we can properly rollback the outer transaction.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -272,7 +272,12 @@ module ActiveRecord
     # such situation.
     def create_or_find_by(attributes, &block)
       with_connection do |connection|
-        transaction(requires_new: true) { create(attributes, &block) }
+        record = nil
+        transaction(requires_new: true) do
+          record = new(attributes, &block)
+          record.save || raise(ActiveRecord::Rollback)
+        end
+        record
       rescue ActiveRecord::RecordNotUnique
         if connection.transaction_open?
           where(attributes).lock.find_by!(attributes)
@@ -287,7 +292,12 @@ module ActiveRecord
     # is raised if the created record is invalid.
     def create_or_find_by!(attributes, &block)
       with_connection do |connection|
-        transaction(requires_new: true) { create!(attributes, &block) }
+        record = nil
+        transaction(requires_new: true) do
+          record = new(attributes, &block)
+          record.save! || raise(ActiveRecord::Rollback)
+        end
+        record
       rescue ActiveRecord::RecordNotUnique
         if connection.transaction_open?
           where(attributes).lock.find_by!(attributes)
--- a/activerecord/test/cases/relations_test.rb
+++ b/activerecord/test/cases/relations_test.rb
@@ -1593,6 +1593,24 @@ class RelationTest < ActiveRecord::TestCase
     assert_not_equal subscriber, Subscriber.create_or_find_by(nick: "cat")
   end
 
+  def test_create_or_find_by_rollbacks_a_transaction
+    assert_no_difference(-> { Car.count }) do
+      car = BrokenCar.create_or_find_by(name: "Civic")
+
+      assert_instance_of(BrokenCar, car)
+      assert_not_predicate(car, :persisted?)
+    end
+  end
+
+  def test_create_or_find_by_bang_rollbacks_a_transaction
+    assert_no_difference(-> { Car.count }) do
+      car = BrokenCar.create_or_find_by!(name: "Civic")
+
+      assert_instance_of(BrokenCar, car)
+      assert_not_predicate(car, :persisted?)
+    end
+  end
+
   def test_create_or_find_by_with_block
     assert_nil Subscriber.find_by(nick: "bob")
 
--- a/activerecord/test/models/car.rb
+++ b/activerecord/test/models/car.rb
@@ -35,3 +35,9 @@ end
 class FastCar < Car
   default_scope { order("name desc") }
 end
+
+class BrokenCar < Car
+  after_save do
+    raise ActiveRecord::Rollback
+  end
+end


activerecord/lib/active_record/relation.rb
```diff
@@ -287,7 +292,12 @@ def create_or_find_by(attributes, &block)
     # is raised if the created record is invalid.
     def create_or_find_by!(attributes, &block)
       with_connection do |connection|
-        transaction(requires_new: true) { create!(attributes, &block) }
+        record = nil
+        transaction(requires_new: true) do
+          record = new(attributes, &block)
+          record.save! || raise(ActiveRecord::Rollback)
```
alexandre184, line 298: Oh thanks, that was quick.
`save!` will normally never return `false`, it raises an exception in that case leaving `|| raise(ActiveRecord::Rollback)` unreachable. You could probably leave `create!` here

Edouard-chin: Thanks for having a look. 

> save! will normally never return false [...] You could probably leave create! here

Unfortunately we have to modify both (bang and non-bang). In a double transaction, if a Rollback is raised in the after_save callback, both `save` and `save!` returns nil.

The `yield` here calls `save!` super method, which ends up raising (because of the after_save callback),
https://github.com/rails/rails/blob/b132744e22de64eb530d6ac407fa4b6b4a926143/activerecord/lib/active_record/transactions.rb#L417

The exception is rescued and swallowed by the transaction, https://github.com/rails/rails/blob/b132744e22de64eb530d6ac407fa4b6b4a926143/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L364

and finally, the status of `save` is returned (which is nil, since the ActiveRecoord::Rollback exception stoped the execution)

https://github.com/rails/rails/blob/b132744e22de64eb530d6ac407fa4b6b4a926143/activerecord/lib/active_record/transactions.rb#L420 

-----------------

Ultimately, we never get a change to enter the other part of the `||` condition 
https://github.com/rails/rails/blob/b132744e22de64eb530d6ac407fa4b6b4a926143/activerecord/lib/active_record/persistence.rb#L424C46-L424C62

activerecord/test/cases/relations_test.rb
```diff
@@ -1593,6 +1593,18 @@ def test_create_or_find_by
     assert_not_equal subscriber, Subscriber.create_or_find_by(nick: "cat")
   end
 
+  def test_create_or_find_by_rollbacks_a_transaction
+    assert_no_difference(-> { Car.count }) do
+      BrokenCar.create_or_find_by(name: "Civic")
```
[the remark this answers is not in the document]
Edouard-chin: 👍 Done

activerecord/test/cases/relations_test.rb
```diff
@@ -1593,6 +1593,24 @@ def test_create_or_find_by
     assert_not_equal subscriber, Subscriber.create_or_find_by(nick: "cat")
   end
 
+  def test_create_or_find_by_rollbacks_a_transaction
+    assert_no_difference(-> { Car.count }) do
+      car = BrokenCar.create_or_find_by(name: "Civic")
+
+      assert_instance_of(BrokenCar, car)
+      assert_not_predicate(car, :persisted?)
```
byroot, line 1601: This is weird to me. I expect `create_or_find_by` to give me a persisted record.

Edouard-chin: I read `create_or_find_by` similar to `create || find_by` where `create` also wouldn't return a persisted record.
Not sure how we can return a persisted record in all cases if a Rollback is requested.

Edouard-chin: Sorry, I tried to wrap my head a few times around your comments but unfortunately I’m still not following :sob:

> This is weird to me. I expect create_or_find_by to give me a persisted record.

I think I’m misreading this, which lead to the comment you didn’t get. So I’ll start from the beginning.

--------------

Would you expect `create_or_find_by` to **always** return a persisted record? I suppose no, since you mention in the next comment “unless of course validation fail or **something**”.

This is where I’m confused, we are in the **"something"** case, which is: a `ActiveRecord::Rollback` error is raised by the application. Since this patch is about taking in consideration the raised Rollback error and preventing the transaction from being committed , why should the record be persisted? 

> `create` should return a persisted record

Just to make sure we are on the same page, `User.create` doesn’t return a persisted record when a Rollback is raised.

<details>
<summary> 🐛 Script </summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.string :title
  end
end

  after_save do
    raise ActiveRecord::Rollback
  end
end

  def test_create
    post = Post.create(title: ".")

    assert_instance_of(Post, post) #=> succeeds
    assert(post.persisted?) #=> fails
  end
end

```
</details>

What I think I’m getting from your suggestion is either:

- `create_or_find_by` should fallback to `find_by` if the creation fails. (It's the case now but only when the creation fails due to a record already existing)
- The current behaviour is correct and we should silently ignore a Rollback error.

Edouard-chin: Ah! I get you now 😅 , thanks!

Just so that you have the full cards in hands to decide whether this should be merged as is:

### Current behaviour

1. If the record fails the creation because of a validation error, a non persisted record is returned. Documented here https://github.com/rails/rails/blob/b97a7625970c74f2273211ccb17046049f409110/activerecord/lib/active_record/relation.rb#L270-L272
2. If the record fails the creation because of a uniquneness validation error, a non persisted record will be returned. This is documented: https://github.com/rails/rails/blob/b97a7625970c74f2273211ccb17046049f409110/activerecord/lib/active_record/relation.rb#L266-L267
3. If a rollback is raised during creation, the rollback is silently ignored.

### With this patch

1. Same as before
2. Same as before
3. The rollback is taken in consideration


Thanks for your help and get better soon !

byroot: Ok, I get it now, I kinda missed the rollback in `after_save` part.

byroot:
Backported to 8.0 and 7.2

PR #42744, opened 2021-07-09
Make vips the default variant processor for new apps

brenogazzola:
### Summary
Changes the default variant processor from `:mini_magick` to `:vips`.

### Other Information
For the full, paragraphs long, explanation about my experiences with both and why I believe vips is the better choice, check my [Rails Discussion post](https://discuss.rubyonrails.org/t/make-vips-the-recommended-default-variant-processor-for-active-storage/78368).

**The abridged version:**
I've been running active storage in production since version 5.2 for image galleries. I've noticed that the combination of Active Storage creating variants through an HTTP request in the web server, ImageMagick high resource needs, and servers with low puma concurrency and memory (eg: Heroku), cause increased response times for the entire application.

It took me a long time to understand all this and that using Vips can mitigate the problem. As a form of "conceptual compression", I propose that Rails 7 make vips the default variant processor.

For reference, I've [prepared a gist](https://gist.github.com/brenogazzola/a4369965a1da426d50f11d080fe2e563) with all the steps I had to go through to migrate from the original mini_magick macros and ImageMagick to the image_processing macros and libvips.

EDIT: Another user in the forum has raised another point: Security. The list of known CVEs for vips is considerably smaller than ImageMagick.

PR #54850, opened 2025-04-01
[ci skip] docs(activestorage): Update default variant processor

mrhead:
Vips has been the new default since https://github.com/rails/rails/pull/42744.


Edouard-chin:
Thanks

PR #54851, opened 2025-04-01
Laying some ground work to remove AR::Base from the modules

kwstannard:
### Motivation / Background

This Pull Request has been created because I was looking into giving engines more complete control of their database connections. A consistent foil is that everything goes through ActiveRecord::Base. To make the hard change easy, the first thing to do is to fix the cohesion of the AR modules.

### Detail

This Pull Request moves all the methods that the connection handling module was calling that were located in the core module. This will not result in functionality changes to ActiveRecord::Base, but will increase cohesion in the Core and ConnectionHandling modules.

### Why this is okay to merge (edited in)

The current state of Core and ConnectionHandling modules is that it is impossible to use ConnectionHandling without Core and also, those specific connection handling methods in core cannot be used without including ConnectionHandling. The modules are incoherent.

Given that Core and ConnectionHandling must always be used together currently, then there should be no danger in increasing the coherence of the connection handling code by putting it in ConnectionHandling.

Thus, the only outcome of this PR is that it will become possible to utilize ConnectionHandling without Core.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_handler.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_handler.rb
@@ -73,9 +73,12 @@ module ActiveRecord
         end
       end
 
+      attr_accessor :default_role
+
       def initialize
         # These caches are keyed by pool_config.connection_name (PoolConfig#connection_name).
         @connection_name_to_pool_manager = Concurrent::Map.new(initial_capacity: 2)
+        @default_role = ActiveRecord.writing_role
       end
 
       def prevent_writes # :nodoc:
--- a/activerecord/lib/active_record/connection_handling.rb
+++ b/activerecord/lib/active_record/connection_handling.rb
@@ -6,6 +6,20 @@ module ActiveRecord
     RAILS_ENV   = -> { (Rails.env if defined?(Rails.env)) || ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence }
     DEFAULT_ENV = -> { RAILS_ENV.call || "default_env" }
 
+    delegate(*%i[
+             default_role
+             ], to: :connection_handler)
+
+    def self.extended(base)
+      base.class_attribute :default_connection_handler, instance_writer: false
+      base.default_connection_handler = ConnectionAdapters::ConnectionHandler.new
+
+      base.class_attribute :default_shard, instance_writer: false, default: :default
+      base.class_attribute :shard_selector, instance_accessor: false, default: nil
+
+      base.include InstanceMethods
+    end
+
     # Establishes the connection to the database. Accepts a hash as input where
     # the <tt>:adapter</tt> key must be specified with the name of a database adapter (in lower-case)
     # example for regular databases (MySQL, PostgreSQL, etc):
@@ -254,6 +268,24 @@ module ActiveRecord
       current_role == role.to_sym && current_shard == shard.to_sym
     end
 
+    # Returns the symbol representing the current connected role.
+    #
+    #   ActiveRecord::Base.connected_to(role: :writing) do
+    #     ActiveRecord::Base.current_role #=> :writing
+    #   end
+    #
+    #   ActiveRecord::Base.connected_to(role: :reading) do
+    #     ActiveRecord::Base.current_role #=> :reading
+    #   end
+    def current_role
+      connected_to_stack.reverse_each do |hash|
+        return hash[:role] if hash[:role] && hash[:klasses].include?(Base)
+        return hash[:role] if hash[:role] && hash[:klasses].include?(connection_class_for_self)
+      end
+
+      default_role
+    end
+
     # Clears the query cache for all connections associated with the current thread.
     def clear_query_caches_for_current_thread
       connection_handler.each_connection_pool do |pool|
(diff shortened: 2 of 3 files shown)


eileencodes:
Thanks for the PR @kwstannard 

I don't think we should move these methods, the change feels cosmetic to me. Lots of parts of Active Record are spread across files so having all the connection code in one file isn't a compelling argument to make this change.

kwstannard:
Hi Eileen! I love your work.

If you are wondering where I personally am going, my thought is that I want to build Rails applications using nothing but an empty Rails app and plug-and-play Rails Engines that only need to be added to the Gemfile and the Railtie initialization system would handle everything. This would allow things like an engine that gets deployed individually in production, but could be added to the 'test' and 'development' groups in another app's gemfile such that the main application could run "integration" tests without needing a full deployment of the remote engine.

Based on my research, the primary thing standing in the way of this is that the connection code and the migration code still assume the ActiveRecord::Base class is the only controller of database stuff. If we could change this so that that code can utilize *any* descendant of ActiveRecord::Base, then the engines could connect and migrate using their own "base" class.

I also have a hunch this would open the doors to other cool patterns as well.

Anyway, I was wasting significant time jumping between many files, so you know, "don't make a difficult change, make the change easy, then make an easy change."

Regards,
Wolf

eileencodes:
> Based on my research, the primary thing standing in the way of this is that the connection code and the migration code still assume the ActiveRecord::Base class is the only controller of database stuff

But these methods are not what's standing in the way of fixing the issue of migrations and rake tasks relying on `ActiveRecord::Base`, it's the fact the app isn't eager loaded until later, so Rails isn't aware of the other connection classes. This is a super non-trivial problem to fix.

> I also have a hunch this would open the doors to other cool patterns as well.

Cool patterns for what though? The database handling should be boring and mostly hidden. Almost no one should have to care about how the internals work.

kwstannard:
> The database handling should be boring and mostly hidden. Almost no one should have to care about how the internals work.

Yes, this is the problem I am hoping to solve. Currently it is impossible to have engines databases wholly encapsulated without knowing how the internals work. I want to make it so an engine can own a database and the only setup necessary is to add it to the Gemfile.

> Rails isn't aware of the other connection classes. This is a super non-trivial problem to fix.

The Rails application could avoid the need to know about other connection classes. What I am leaning towards is that each engine with a database configuration could own its own connection via an "EngineRecord" that will have a class instance variable "configuration" and also the engine has its own subclass of DatabaseTasks. Then, when you run `rails db:whatever`, each engine would append its own `whatever` task onto the rake tasks such that it will run.

I have some prototypes I could show you this sounds interesting.

eileencodes:
I agree that engines aren't where they need to be overall, not just in the database handling. But I'm concerned what you're proposing only fixes engines, while making connection handling more complex, rather than a path towards no reliance on AR Base in the rake tasks.

I'm against moving code around without understanding where you're going, so if you have prototypes, I'm happy to take a look at them.

kwstannard:
Here is some initial initial work towards the goal done within activerecord. Sorry the commit isn't more organized: https://github.com/rails/rails/commit/21d674463c8ec12f06124c6970434c3275892ab6

Here is a slightly later repo where I hack a bunch of stuff in an engine instance to make it encapsulated. https://github.com/kwstannard/test-repo/tree/main/modules/database

kwstannard:
@eileencodes Got some more work in that is hopefully going to make this make more sense. Databases are getting created and migrated with `db:prepare`. Note: I am planning on deleting everything but the tests and redoing it with a pair, this is more of a spike.

https://github.com/rails/rails/compare/main...kwstannard:rails:engine-only?expand=1

Tests showing an example application with two rails engines:

https://github.com/rails/rails/compare/main...kwstannard:rails:engine-only?expand=1#diff-a23082989c602d080f1e308e79bea56300e6889223becc5aa79a7b7f63aab3f0

Note here that now that connection_handling contains all connection related code and is fully cohesive, we can use it else-where and cut duplicate code.

https://github.com/rails/rails/compare/main...kwstannard:rails:engine-only?expand=1#diff-c7b2018646f254d00541db2d6cdb3b02b64ac8cc7a7dc2fb0f1b67e9c8cb7ff8
https://github.com/rails/rails/compare/main...kwstannard:rails:engine-only?expand=1#diff-cb5c658a1aa7877862198142775880b76b1d327ade6fd39f7c29a080535de6b7

eileencodes:
Thanks for sharing your work. I'm not sure the path you're going down is going to work - at least not within the way Rails needs it to.

For one there's a bunch of public APIs getting changed / removed, which can't happen without a deprecation cycle. Anything without a `nodoc` or under the private namespace is public, regardless of whether there's accompanying documentation. 

Second this appears to entirely break configurations and connections for applications - I can't create a db in a new app, nor do the active record tests boot. I get that this PR is in progress and not done, but I'm wary of major changes to connection management without a clear plan where we're going.

Engines aren't the default in Rails, applications are, so they take precedent. Connection management isn't a simple part of the codebase and it's really easy to break applications if you're not aware of the edge cases.

kwstannard:
> rather than a path towards no reliance on AR Base in the rake tasks.

let me specifically note that the above branch removes references to Base

kwstannard:
Sorry, I just saw your new comment.

I agree with you about maintaining interfaces. The plan is that during the rework I will re-implement everything through ConnectionHandling methods.

kwstannard:
To get back to this specific PR, can we agree that moving the connection handling methods into connection_handling will allow us to more easily utilize connection_handling in other places like DatabaseTasks in place of ActiveRecord::Base, which it sounds like you already believe is a good idea?

kwstannard:
Updated the title and description to make this more compelling.

matthewd:
To me this appears to further conflate two currently (more-or-less, though not perfectly) distinct responsibilities: the heritable class-scoped behaviour in ConnectionHandling, and the global / caller-scoped state in Core.

kwstannard:
Okay, I think my brain is getting it. Let me try to explain this back: Core module is for global state and caller state. ConnectionHandling is class scoped behavior. You are saying that methods that I moved are state related and should be in Core?

kwstannard:
Can someone hop on a call with me so that we can address our disconnect here in a reasonable amount of time?

PR #50946, opened 2024-02-02
Fix error raised when running `db:schema:load:#{name}` and test DB config is missing

ricardotk002:
### Motivation / Background

Fixes #50672

### Detail

The task `db:schema:load:#{name}` depends on `db:test:purge:#{name}` since adb64db which requires the DB config to be present, otherwise an error is raised.

```
$ bin/rails db:schema:load:one --trace
** Invoke db:schema:load:one (first_time)
** Invoke db:test:purge:one (first_time)
** Invoke db:load_config (first_time)
** Invoke environment (first_time)
** Execute environment
** Execute db:load_config
** Invoke db:check_protected_environments (first_time) ** Invoke db:load_config
** Execute db:check_protected_environments
** Execute db:test:purge:one
** Execute db:schema:load:one
```

This commit fixes the problem by returning early when the DB config is indeed missing.

### Additional information

The task `db:schema:load` may need to be updated to depend on `db:test:purge` too, to keep it consistent with `db:schema:load:#{name}`.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54853, opened 2025-04-01
Avoid purging the test database when loading a schema:

### LINKED ISSUE #50672: db:schema:load:<name> purges test database
Commit [adb64db43d](https://github.com/rails/rails/commit/adb64db43dba5cad1007ce57e25df67f8b3e91f8#diff-4a42b5efdab725af0936f94b4a84dcf6a8be8738e28ed265e0a3e74f275f5048R491) changed task `db:schema:load:<name>` to invoke `db:test:purge:<name>`.

Because of that Rails 7.1 breaks my workflow for initializing a local environment:

- create database with `db:create`
- run custom SQL to create Postgres extensions, schemas and triggers
- load schema with `db:schema:load`

Strangely, `db:schema:load` keeps the old behavior, and I could not find the explanation for the change in the commit message or in related Github issues.

Also, it _requires_ a test database to be configured in `database.yml`; otherwise the task fails:

```
$ RAILS_ENV=development bin/rails db:schema:load:primary
bin/rails aborted!
TypeError: Invalid type for configuration. Expected Symbol, String, or Hash. Got nil (TypeError)

        raise TypeError, "Invalid type for configuration. Expected Symbol, String, or Hash. Got #{config.inspect}"
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Tasks: TOP => db:schema:load:primary => db:test:purge:primary
(See full trace by running task with --trace)
```

Can someone confirm the change was intentional?

Edouard-chin:
### Motivation / Background

Fix #50672
Closes #50946

#### Problem

In adb64db, we added a change to purge the test database whenever the `load:schema:<name>` was run. This change meant that in order for the schema to load:
1) the test database configuration needs to be setup
2) can be connected to.

This creates some issues for users if either of both conditions aren't met. (For 1. it can be argued that it's a user error, but for 2., it's likely that a production environment can't connect to the test database).

#### Details

I dug to understand why loading a specific schema requires purging the test database (whereas loading all schema doesn't), but couldn't really see anything obvious, so I concluded that it may just be some overly defensive code.

#### Solution

Whenever we run a test, Rails maintains all test databases schema, and if a schema changes, the equivalent of a purge will be executed on all test databases. https://github.com/rails/rails/blob/b132744e22de64eb530d6ac407fa4b6b4a926143/railties/lib/rails/testing/maintain_test_schema.rb#L5

I couldn't really think of a negative behaviour difference between purging the test database eagerly when we load the schema and lazily when we run the tests, so I simply removed the `purge` rake task dependency on `schema:load:<name>`.

### Checklist

### DIFF
--- a/railties/test/application/rake/multi_dbs_test.rb
+++ b/railties/test/application/rake/multi_dbs_test.rb
@@ -510,6 +510,45 @@ module ApplicationTests
         end
       end
 
+      test "db:schema:load:name doesn't purge the test database. The test schema is maintained when running tests" do
+        require "#{app_path}/config/environment"
+        Dir.chdir(app_path) do
+          generate_models_for_animals
+
+          File.open("test/models/dog_test.rb", "w") do |file|
+            file.write(<<~EOS)
+              require "test_helper"
+
+              class DogTest < ActiveSupport::TestCase
+                test "Dog name type" do
+                  puts Dog.type_for_attribute(:name).type
+                end
+              end
+            EOS
+          end
+
+          rails("db:migrate:primary", "db:migrate:animals")
+
+          development_runner_output = rails("runner", "puts Dog.type_for_attribute(:name).type")
+          assert_match(/string/, development_runner_output)
+
+          test_output = rails("test", "test/models/dog_test.rb")
+          assert_match(/string/, test_output)
+
+          # Simulate a schema change
+          content = File.read("db/animals_schema.rb")
+          content.gsub!(/t\.string "name"/, "t.text \"name\"")
+          File.write("db/animals_schema.rb", content)
+
+          rails("db:schema:load:animals")
+          development_runner_output = rails("runner", "puts Dog.type_for_attribute(:name).type")
+          assert_match(/text/, development_runner_output)
+
+          test_output = rails("test", "test/models/dog_test.rb")
+          assert_match(/text/, test_output)
+        end
+      end
+
       test "db:migrate respects timestamp ordering across databases" do
         require "#{app_path}/config/environment"
         app_file "db/migrate/01_one_migration.rb", <<-MIGRATION
--- a/activerecord/lib/active_record/railties/databases.rake
+++ b/activerecord/lib/active_record/railties/databases.rake
@@ -474,7 +474,7 @@ db_namespace = namespace :db do
     namespace :load do
       ActiveRecord::Tasks::DatabaseTasks.for_each(databases) do |name|
         desc "Load a database schema file (either db/schema.rb or db/structure.sql, depending on configuration) into the #{name} database"
-        task name => "db:test:purge:#{name}" do
+        task name => [:load_config, :check_protected_environments] do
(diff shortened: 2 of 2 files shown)


railties/test/application/rake/multi_dbs_test.rb
```diff
@@ -510,6 +510,45 @@ def generate_models_for_animals
         end
       end
 
+      test "db:schema:load:name doesn't purge the test database. The test schema is maintained when running tests" do
```
ghiculescu, line 513: So with this change, what do you need to do to update the test database?

Edouard-chin: In most scenarios you won't have to do anything manually to update the test schema, it will be taken care of by Rails whenever a test is ran.

One of the only reason I can think of where you may need to update it manually (by running the `bin/rails db:test:prepare command`) is if you run a console in the test environment `RAILS_ENV=test bin/rails c`, just after the schema is updated. This I think is very edge case and I don't think would cause issues.

It's worth noting that loading all schemas `db:schema:load` also doesn't purge the tests database and AFAIK hasn't caused any issues for users.

ghiculescu: > you won't have to do anything manually to update the test schema, it will be taken care of by Rails whenever a test is ran.

But isn't this what this test is showing? That the test schema is not updated between test runs?

(It's very likely I'm not understanding it!)

Edouard-chin: > That the test schema is not updated between test runs?

The opposite 😅 .

The schema is modified here https://github.com/rails/rails/blob/ef1ccde0d8825a5d0dc784e3185b206d5bd8d56c/railties/test/application/rake/multi_dbs_test.rb#L539-L541

and we check that running the test once more keep the test schema up to date 

https://github.com/rails/rails/blob/ef1ccde0d8825a5d0dc784e3185b206d5bd8d56c/railties/test/application/rake/multi_dbs_test.rb#L547-L548

----

I suppose the test name is confusing?

ghiculescu: Well it clearly confused me, but now that you've explained it to me and I've read it again, I am pretty sure you are right and the test is fine. 

Sorry for the confusion!

alexcwatt:
@Edouard-chin thanks for working on this! I stumbled across this PR while trying to get solid queue working in production; `bin/rails db:schema:load:queue` didn't work, and I had to use this workaround: https://github.com/rails/solid_queue/issues/365#issuecomment-2548639056

It seems your PR might fix this - hope it can be merged if so 😄

Edouard-chin:
Thanks Jean ❤️

tombeynon:
Thank you for fixing this - does anyone know if this will be backported to rails 7.1?

byroot:
7.1 is in security only maintenance: https://rubyonrails.org/maintenance

PR #54854, opened 2025-04-01
Validate size of all cookies

skipkayhil:
### Motivation / Background

Previously, cookie size validation was only performed for signed/encrypted cookies, but not plaintext ones.

### Detail

This commit refactors the base CookieJar to perform cookie size validation so that even non-{signed,encrypted} cookies have their size validated.

### Checklist

### DIFF
--- a/actionpack/test/dispatch/cookies_test.rb
+++ b/actionpack/test/dispatch/cookies_test.rb
@@ -49,17 +49,17 @@ class CookieJarTest < ActiveSupport::TestCase
   end
 
   def test_each
-    request.cookie_jar["foo"] = :bar
+    request.cookie_jar["foo"] = "bar"
     list = []
     request.cookie_jar.each do |k, v|
       list << [k, v]
     end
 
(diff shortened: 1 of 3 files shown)


actionpack/test/dispatch/cookies_test.rb
```diff
@@ -49,17 +49,17 @@ def test_fetch_type_error
   end
 
   def test_each
-    request.cookie_jar["foo"] = :bar
+    request.cookie_jar["foo"] = "bar"
```
skipkayhil, line 52: This is a slight behavior change from before: previously the signed/encrypted jars would cast values to a string so that `check_for_overflow!` can call `bytesize`, but the new implementation requires all values to be Strings so that `bytesize` can be called.

I couldn't find any documentation indicating that a symbol or other object in a cookie value was supported, only the following in `ActionController::Cookies`:

```ruby
# Cookie values are String-based. Other data types need to be serialized.
cookies[:lat_lon] = JSON.generate([47.68, -122.37])
```

PR #54855, opened 2025-04-02
Add `Cache#read_counter` and `Cache#write_counter`

ghiculescu:
ref: https://github.com/rails/rails/pull/54821#issuecomment-2768469886

This PR introduces 2 new methods on `Rails.cache`: `read_counter` and `write_counter`.

The idea is that you should use them to read values that are being `increment`ed / `decrement`ed. This is to avoid the code smell of having to pass `raw: true` around on `read`/`write` calls for counters (which is only necessary with some cache implementations).

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Add `Cache#read_counter` and `Cache#write_counter`
+
+    ```ruby
+    Rails.cache.write_counter("foo", 1)
+    Rails.cache.read_counter("foo") # => 1
+    Rails.cache.increment("foo")
+    Rails.cache.read_counter("foo") # => 2
+    ```
+
+    *Alex Ghiculescu*
+
 *   Introduce ActiveSupport::Testing::ErrorReporterAssertions#capture_error_reports
 
     Captures all reported errors from within the block that match the given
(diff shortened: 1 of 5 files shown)


activesupport/test/cache/behaviors/cache_increment_decrement_behavior.rb
```diff
@@ -31,6 +31,18 @@ def test_decrement
...
+  def test_read_counter_and_write_counter
+    key = SecureRandom.uuid
+    @cache.write_counter(key, 1)
+    assert_equal 1, @cache.read(key, raw: true).to_i
+
+    assert_equal 1, @cache.read_counter(key)
+    assert_equal 2, @cache.increment(key)
+    assert_equal 2, @cache.read_counter(key)
+
+    assert_nil @cache.read_counter(SecureRandom.alphanumeric)
```
camallen, line 43: won't this be 0 via `nil.to_i` ?

ghiculescu: Yeah I just pushed a fix, now it's `nil&.to_i`.

ghiculescu:
Red test seems unrelated.

chaadow:
Nice!

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+*   Add `Cache#read_counter` and `Cache#write_counter`
+
+    ```ruby
+    Rails.cache.write_counter("foo", 1)
+    Rails.cache.read_counter("foo") # => 1
+    Rails.cache.increment("foo")
```
zzak, line 6: Doesn't `increment` also return the new value, or keeping that out for readability?

ghiculescu: Yes, and yes.
But happy to change if you think it's confusing.

PR #54856, opened 2025-04-02
fix(devcontainer): support DB switching when containerEnv is not defined

vsppedro:
### Motivation / Background

This Pull Request fixes an issue with the `bin/rails db:system:change --to=postgresql` command when environment variables are not defined using containerEnv in the devcontainer configuration.

### Detail

When running the command:

```bash
bin/rails db:system:change --to=postgresql
```

Rails attempts to update the devcontainer.json file and assumes container_env is always present. If that's not the case, it raises the following error:

```
/usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/generators/rails/db/system/change/change_generator.rb:151:in `update_devcontainer_db_host': undefined method `[]' for nil (NoMethodError)

              if container_env["DB_HOST"]
                              ^^^^^^^^^^^
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/generators/rails/db/system/change/change_generator.rb:116:in `edit_devcontainer_json'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/generators/rails/db/system/change/change_generator.rb:58:in `edit_devcontainer_files'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/command.rb:28:in `run'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:127:in `invoke_command'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:134:in `block in invoke_all'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:134:in `each'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:134:in `map'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:134:in `invoke_all'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/group.rb:243:in `dispatch'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/base.rb:584:in `start'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/commands/db/system/change/change_command.rb:20:in `perform'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/command.rb:28:in `run'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor/invocation.rb:127:in `invoke_command'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/command/base.rb:178:in `invoke_command'
        from /usr/local/bundle/ruby/3.3.0/gems/thor-1.3.2/lib/thor.rb:538:in `dispatch'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/command/base.rb:73:in `perform'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/command.rb:65:in `block in invoke'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/command.rb:143:in `with_argv'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/command.rb:63:in `invoke'
        from /usr/local/bundle/ruby/3.3.0/gems/railties-8.0.2/lib/rails/commands.rb:18:in `<main>'
        from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:69:in `require'
        from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:69:in `block (2 levels) in replace_require'
        from /usr/local/bundle/ruby/3.3.0/gems/bootsnap-1.18.4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
        from bin/rails:4:in `<main>'
```

This PR adds a check to skip modifications to containerEnv if it's not defined. 

### Checklist

### DIFF
--- a/railties/test/fixtures/.devcontainer/no_env/devcontainer.json
+++ b/railties/test/fixtures/.devcontainer/no_env/devcontainer.json
@@ -1,0 +1,25 @@
+// For format details, see https://containers.dev/implementors/json_reference/.
+// For config options, see the README at: https://github.com/devcontainers/templates/tree/main/src/ruby
+{
+  "name": "tmp",
+  "dockerComposeFile": "compose.yaml",
+  "service": "rails-app",
+  "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {}
+  },
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  // "forwardPorts": [],
+
+  // Configure tool-specific properties.
+  // "customizations": {},
+
+  // Uncomment to connect as root instead. More info: https://containers.dev/implementors/json_reference/#remoteUser.
+  // "remoteUser": "root",
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "bin/setup --skip-server"
+}
--- a/railties/lib/rails/generators/rails/db/system/change/change_generator.rb
+++ b/railties/lib/rails/generators/rails/db/system/change/change_generator.rb
@@ -145,7 +145,7 @@ module Rails
             end
 
             def update_devcontainer_db_host
-              container_env = devcontainer_json["containerEnv"]
+              container_env = devcontainer_json["containerEnv"] || {}
               db_name = database.name
 
               if container_env["DB_HOST"]
@@ -162,7 +162,13 @@ module Rails
 
               new_json = JSON.pretty_generate(container_env, indent: "  ", object_nl: "\n  ")
 
-              gsub_file(".devcontainer/devcontainer.json", /("containerEnv"\s*:\s*)(.|\n)*?(^\s{2}})/, "\\1#{new_json}")
+              if devcontainer_json["containerEnv"].present?
+                gsub_file(".devcontainer/devcontainer.json", /("containerEnv"\s*:\s*)(.|\n)*?(^\s{2}})/, "\\1#{new_json}")
+              else
+                gsub_file ".devcontainer/devcontainer.json", /(\s+\S+":.*)(?=\n\s*})/m do |match|
+                  "#{match},\n\n  \"containerEnv\": #{new_json}"
+                end
+              end
             end
 
             def update_devcontainer_db_feature
--- a/railties/test/generators/db_system_change_generator_test.rb
+++ b/railties/test/generators/db_system_change_generator_test.rb
@@ -218,6 +218,16 @@ module Rails
               assert_not_includes compose_config.keys, "volumes"
             end
           end
+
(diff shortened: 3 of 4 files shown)


andrewn617:
Thanks for the report and the PR, but I am not sure this fixes the issue. We need to style write the `DB_HOST` for the new database type (eg postgres), even if there was previously no env variables defined in container env. Can you update your implementation to do that, and write a test?

andrewn617:
When you generate a devcontainer with the app generator the `DB_HOST` variable is there in `devcontainer.json`. You are right we can't predict every place that the user may have put env variables. This generator is a best effort and getting a work dev container when doing a db change, so it hews closely to where we put env variables. I don't think there is any issue if the variable ends up being set twice, but if we don't set it at all then many users will probably not have it and thus not have a working dev container after running the command. So I think it is important that we add it.

railties/test/fixtures/.devcontainer/devcontainer-no-env.json
```diff
@@ -0,0 +1,25 @@
...
+
+  // Configure tool-specific properties.
+  // "customizations": {},
+
+  // Uncomment to connect as root instead. More info: https://containers.dev/implementors/json_reference/#remoteUser.
+  // "remoteUser": "root",
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "bin/setup --skip-server"
+}
```
vsppedro, line 25: Hmm, I need to give this some more thought

railties/test/fixtures/.devcontainer/no_env/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+
+  // Configure tool-specific properties.
+  // "customizations": {},
+
+  // Uncomment to connect as root instead. More info: https://containers.dev/implementors/json_reference/#remoteUser.
+  // "remoteUser": "root",
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "bin/setup --skip-server"
+}
```
vsppedro, line 25: To be honest, I’m not a fan of adding a folder just for this file. However, if I don’t, and instead try to use a differently named devcontainer file like devcontainer-no-env.json, the comments end up being marked as errors in both GitHub and VSCode.

andrewn617: I think it's ok. Thanks for adding the test.

andrewn617 (approved):

PR #54863, opened 2025-04-03
Add missing test case for default_isolation_level

chaadow:
Related to Pull Request: #54836

cc @byroot @kirs


kirs (approved):
Thanks for adding this one.

PR #54380, opened 2025-01-27
Load the routes in the console when calling the `app` IRB helper

Edouard-chin:
### Motivation / Background

Load the routes in the console when calling  the `app` IRB helper:

- When running a console, the routes are not loaded, which results in not being able to see what route helpers are available.

```shell
$ bin/rails c
$ catana(dev)> app.methods.grep(/_(path|url)/) # []
$ catana(dev)> app.foo_path # Works. The routes are now loaded.
$ catana(dev)> app.methods.grep(/_(path|url)/) # [:foo_path]
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54865, opened 2025-04-04
Load routes when using Rails runner command

### LINKED ISSUE #54065: named_routes.each never called after upgrade to Rails 8
### Steps to reproduce

In a rails app with routes

```
echo 'puts Rails.application.routes.named_routes.to_a.size' | bin/rails runner -
```

Before upgrading to rails 8, this prints a number greater than zero. 

Afterwards, this prints zero.

I couldn't trigger this using the standard executable testcase template.

### Expected behavior

Iterating over the route list (eg to generate ajax helpers) works.

### Actual behavior

No routes are found until you call a method like `recognize_path` which forces the `LazyRouteSet` to be loaded.

### System configuration

**Rails version**:

8.0.1

**Ruby version**:

3.4.1

zzak:
Fixes: #54065
Inspired by: #54380

/cc @Edouard-chin @byroot

### DIFF
--- a/railties/lib/rails/commands/runner/runner_command.rb
+++ b/railties/lib/rails/commands/runner/runner_command.rb
@@ -28,9 +28,7 @@ module Rails
         end
 
         boot_application!
-        app = Rails.application
-        app.load_runner
-        app.reload_routes_unless_loaded
+        Rails.application.load_runner
 
         ARGV.replace(command_argv)
 
--- a/railties/lib/rails/commands/runner/runner_command.rb
+++ b/railties/lib/rails/commands/runner/runner_command.rb
@@ -28,7 +28,9 @@ module Rails
         end
 
         boot_application!
-        Rails.application.load_runner
+        app = Rails.application
+        app.load_runner
+        app.reload_routes_unless_loaded
 
         ARGV.replace(command_argv)
 
--- a/railties/lib/rails/engine/lazy_route_set.rb
+++ b/railties/lib/rails/engine/lazy_route_set.rb
@@ -12,6 +12,11 @@ module Rails
           Rails.application&.reload_routes_unless_loaded
           super
         end
+
+        def routes
+          Rails.application&.reload_routes_unless_loaded
+          super
+        end
       end
 
       module ProxyUrlHelpers
@@ -86,6 +91,16 @@ module Rails
       def routes
         Rails.application&.reload_routes_unless_loaded
         super
+      end
+
+      def collection
+        named_routes.names.each_with_object({}) do |name, collection|
+          path_helper = "#{name}_path".to_sym
+          url_helper = "#{name}_url".to_sym
+
+          collection[path_helper] = url_helpers.method(path_helper)
+          collection[url_helper] = url_helpers.method(url_helper)
+        end
       end
 
       private
--- a/railties/test/application/runner_test.rb
+++ b/railties/test/application/runner_test.rb
@@ -37,6 +37,16 @@ module ApplicationTests
       assert_match "42", rails("runner", "puts User.count")
     end
 
+    def test_app_routes_are_loaded
+      app_file "config/routes.rb", <<-RUBY
+        Rails.application.routes.draw do
+          get 'foo', to: 'foo#index'
+        end
+      RUBY
+
+      assert_match "[:foo_path]", rails("runner", "puts Rails.application.routes.url_helpers.methods.grep(/foo_path/).inspect")
+    end
+
     def test_should_set_argv_when_running_code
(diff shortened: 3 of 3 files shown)


byroot:
Seems to narrow of a fix. The right behavior should be that calling `named_routes` should trigger routes loading no matter what the context is.

Edouard-chin:
Yeah unfortunately I don't this fix the issue mentioned. (The snippet in the issue was given for example purposes but the real use case is to expose all named routes to the front-end so that javascript code can make use of them).

Loading the routes as soon as `named_routes` is called is going to be tricky as it will defeat the purpose of the Lazy route feature. There are many instance where `named_routes` get called internally but it shouldn't trigger loading routes, e.g. [here](https://github.com/rails/rails/blob/6b5c91d2af47c976b57d8a90d372b8ec8fb7c7bd/railties/lib/rails/engine/lazy_route_set.rb#L52-L53)

Worth to note that the `named_routes` is nodoc and I don't think users really need the whole object, they seem to just need a hash of the route name and the url. Maybe we could provide a method as a public API instead and override it in the LazyRoute class, similar to what's already existing for another method:

https://github.com/rails/rails/blob/6b5c91d2af47c976b57d8a90d372b8ec8fb7c7bd/railties/lib/rails/engine/lazy_route_set.rb#L10-L15

zzak:
>  The right behavior should be that calling `named_routes` should trigger routes loading no matter what the context is.

If we do it like that, then as Edouard said will cause the routes to always be loaded. I think this backtrace confirms it, I just raised for the caller in some random test:

```
railties/lib/rails/engine/lazy_route_set.rb:97:in 'Rails::Engine::LazyRouteSet#named_routes': called (RuntimeError)
  from railties/lib/rails/engine/lazy_route_set.rb:57:in 'Rails::Engine::LazyRouteSet#initialize'
  from actionpack/lib/action_dispatch/routing/route_set.rb:381:in 'Class#new'
  from actionpack/lib/action_dispatch/routing/route_set.rb:381:in 'ActionDispatch::Routing::RouteSet.new_with_config'
  from railties/lib/rails/engine.rb:545:in 'Rails::Engine#routes'
  from railties/lib/rails/engine.rb:597:in 'block in <class:Engine>'
  from railties/lib/rails/initializable.rb:24:in 'BasicObject#instance_exec'
  from railties/lib/rails/initializable.rb:24:in 'Rails::Initializable::Initializer#run'
  from railties/lib/rails/initializable.rb:103:in 'block in Rails::Initializable#run_initializers'
  from railties/lib/rails/initializable.rb:59:in 'Array#each'
  from railties/lib/rails/initializable.rb:59:in 'Rails::Initializable::Collection#each'
  from railties/lib/rails/initializable.rb:102:in 'Rails::Initializable#run_initializers'
  from railties/lib/rails/application.rb:442:in 'Rails::Application#initialize!'
  from railties/test/fixtures/tmp/config/environment.rb:5:in '<top (required)>'
  from railties/lib/rails/application.rb:418:in 'Rails::Application#require_environment!'
  from railties/lib/rails/command/actions.rb:20:in 'Rails::Command::Actions#boot_application!'
  from railties/lib/rails/commands/generate/generate_command.rb:21:in 'Rails::Command::GenerateCommand#perform'
```

So I think it is a little more nuanced than making `named_routes` eagerly load the routes.

> However if named_routes was supposed to be public (it's not documented but could have been overlooked due to named_routes being an attr reader) then Rails may need to redefine reader and call reload_routes_unless_loaded to ensure routes are loaded

This [comment](https://github.com/rails/rails/issues/54065#issuecomment-2564382449) kind of goes along with what Edouard suggested to provide a new public method instead that will load the routes and return an object that users can call.

I think _they don't really need the whole object_ is a good point, so in this commit I'm just returning a hash with the helper method name and method as the value.

https://github.com/rails/rails/pull/54865/commits/bf7a4156766497c48b77821eadc99d25fddde02d

I'm only suggesting this as an idea for discussion, not this should be merged as is. But handing back the user an instance of `NamedRouteCollection` which should be private seems bad.

Interestingly [AD::Routing::RouteSet::NamedRouteCollection](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Routing/RouteSet/NamedRouteCollection.html) and [Rails::Engine::LazyRouteSet, NamedRouteCollection](https://edgeapi.rubyonrails.org/classes/Rails/Engine/LazyRouteSet/NamedRouteCollection.html) are incorrectly being exposed in the docs.

byroot:
> we should deprecate `named_routes` and introduce another method that is public and eager load the routes.

Or make it public, but have a named argument to toggle eagerloading `named_routes(eager_load: false)`.

PR #54869, opened 2025-04-04
Fix check for remote databases when host is a socket path

9999years:
### Motivation / Background

We're configuring our database by setting `PGHOST` to a local Unix socket path.

Unfortunately, this causes Rails to think that the database is on a remote host:

    This task only modifies local databases. example-database is on a remote host.

Let's fix this by checking if the host is a path to a file.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/tasks/database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/database_tasks.rb
@@ -617,7 +617,7 @@ module ActiveRecord
 
         def local_database?(db_config)
           host = db_config.host
-          host.blank? || LOCAL_HOSTS.include?(host)
+          host.blank? || LOCAL_HOSTS.include?(host) || File.exist?(host)
         end
 
         def schema_sha1(file)
--- a/activerecord/test/cases/tasks/database_tasks_test.rb
+++ b/activerecord/test/cases/tasks/database_tasks_test.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "tempfile"
 require "cases/helper"
 require "active_record/tasks/database_tasks"
 require "models/course"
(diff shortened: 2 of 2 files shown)


yahonda:
Thanks for opening this pull request! Could you share your database.yml?

From what I understand, the value assigned to the PGHOST environment variable should be the directory specified by unix_socket_directories. 
In this case, /tmp not a specific socket file path like /tmp/.s.PGSQL.5432.

```sql
% psql -d postgres
psql (16.8 (Homebrew))
Type "help" for help.

postgres=# SHOW unix_socket_directories;
 unix_socket_directories
-------------------------
 /tmp
(1 row)
```

The current implementation and test seem to check for the existence of what looks like a socket file path, which might unintentionally lead to false positives depending on the environment.

9999years:
> From what I understand, the value assigned to the PGHOST environment variable should be the directory specified by unix_socket_directories.

That's correct — directories also count as files for the purposes of `File.exists?`.

> The current implementation and test seem to check for the existence of what looks like a socket file path, which might unintentionally lead to false positives depending on the environment.

Could you explain how that might lead to false positives? I'm not sure how using a file that exists locally as a host could actually be a remote host.

yahonda:
Thanks for the patch.

Your use case is already supported without any code change. The path is:

1. If `host:` is omitted from `database.yml`, `db_config.host` is `nil`, so the existing `host.blank?` branch in `local_database?` returns `true`.
2. Active Record's PostgreSQL adapter then calls `@config.compact` (`postgresql_adapter.rb:323`) which strips `nil` values, so `PG.connect` receives no `host` kwarg.
3. libpq falls back to the `PGHOST` environment variable per its [documented behavior](https://www.postgresql.org/docs/current/libpq-envars.html), which is exactly the Unix socket directory you want to point at.

Concretely:

```yaml
# config/database.yml
development:
  adapter: postgresql
  database: myapp_dev
  # no host: line
```

```sh
export PGHOST=/tmp   # or /var/run/postgresql, $PWD/.devenv/state/postgres, etc.
bin/rails db:create db:migrate
```

One side note on the proposed check: `File.exist?(host)` resolves relative to the current working directory, so `host: db` in a Rails project root would match the `db/` directory and flip a remote database to "local" — another reason to prefer the configuration route here.

Closing in favor of the workaround above. Thanks again for raising the use case.

PR #50475, opened 2023-12-28
support dumping PostgreSQL inheritance & partitioning options to `schema.rb`

waymondo:
### Motivation / Background & Detail

[`create_table`](https://api.rubyonrails.org/v7.1.2/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-create_table) provides an `options:` key that allows you to append options to your table definition. One of the things you can do this way with PostgreSQL 10+ is use [native partitioning](https://www.postgresql.org/docs/current/ddl-partitioning.html) to declare partitioning definitions:

```
create_table :events, id: false, options: "PARTITION BY LIST (account_id)" do |t|
  t.bigint :account_id, null: false
  t.integer :kind, null: false
  t.datetime :occurred_at, null: false
end
```

Adding this to a migration and running it will properly create the partitioned table in development, but the dumped schema will not persist the specified partitioning options. Therefore, if you run your test suite which loads from `schema.rb`, your test database's table won't be properly partitioned.

This PR extends the PostgeSQL adapter's schema statements so that it can dump this partitioning definition as defined in the relevant system tables. 

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54872, opened 2025-04-05
Fix dumping declaratively partitioned tables in PostgreSQL

### LINKED ISSUE #54841: pg partitioned table schema dump is invalid for sub-tables
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  gem "pg"
end

ENV['DATABASE_URL'] ||= 'postgresql:test'

require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

end

  enum :kind, [:a, :b].index_with(&:to_s)
end

class CreatePartitionedPost < ActiveRecord::Migration::Current
  def up
    create_table :posts, id: false, force: true, options: 'partition by list (kind)' do |t|
      t.string :title
      t.string :kind
    end

    create_table :posts_a, id: false, options: "partition of posts for values in ('a')"
    create_table :posts_b, id: false, options: "partition of posts for values in ('b')"
  end

  def down
    drop_table :posts, cascade: true, if_exists: true
  end
end

require 'rake'
load 'active_record/railties/databases.rake'

task :environment do; end

# cleanup
CreatePartitionedPost.migrate(:down)
# create tables
CreatePartitionedPost.migrate(:up)
# child tables have "inherits(parent)" instead of "partition of (parent) for values in ('value')"
Rake::Task['db:schema:dump'].invoke
Post.reset_column_information

  def test_schema_load
    # ActiveRecord::StatementInvalid: PG::WrongObjectType: ERROR:  cannot inherit from partitioned table "posts"
    Rake::Task['db:schema:load'].invoke
  end
end
```

### Expected behavior

at least: 
`rails db:schema:load` should not raise. which means, not including inherits(parent) options for partitioned child tables.

better:
schema should not define `inherits(parent_table)` for child tables and instead attach them after table definition.

```ruby
ActiveRecord::Schema[8.0].define(version: 0) do
  # These are extensions that must be enabled in order to support this database
  enable_extension "pg_catalog.plpgsql"

  create_table "posts", id: false, options: "PARTITION BY LIST (kind)", force: :cascade do |t|
    t.string "title"
    t.string "kind"
  end

  create_table "posts_a", id: false do |t|
    t.string "title"
    t.string "kind"
  end
  # we probably lack corresponding ar migration command
  execute "ALTER TABLE ONLY posts ATTACH PARTITION posts_a FOR VALUES IN ('a');"

  create_table "posts_b", id: false do |t|
    t.string "title"
    t.string "kind"
  end
  # we probably lack corresponding ar migration command
  execute "ALTER TABLE ONLY posts ATTACH PARTITION posts_b FOR VALUES IN ('b');"
end
```

### Actual behavior

schema is invalid for children tables and exception is raised when we try to load it.

`ActiveRecord::StatementInvalid: PG::WrongObjectType: ERROR:  cannot inherit from partitioned table "posts"`

```ruby
ActiveRecord::Schema[8.0].define(version: 0) do
  # These are extensions that must be enabled in order to support this database
  enable_extension "pg_catalog.plpgsql"

  create_table "posts", id: false, options: "PARTITION BY LIST (kind)", force: :cascade do |t|
    t.string "title"
    t.string "kind"
  end

  create_table "posts_a", id: false, options: "INHERITS (posts)", force: :cascade do |t|
    t.string "title"
    t.string "kind"
  end

  create_table "posts_b", id: false, options: "INHERITS (posts)", force: :cascade do |t|
    t.string "title"
    t.string "kind"
  end
end
```

### System configuration

**Rails version**:  latest stable
**Ruby version**:  latest stable

fatkodima:
Fixes #54841.

The problem was that tables defined with declarative partitioning (https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE) weren't correctly dumped. 

In the original PR (https://github.com/rails/rails/pull/50475) it was always assumed that tables are defined with inheritance based partitioning (https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-USING-INHERITANCE).

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
@@ -39,6 +39,17 @@ module ActiveRecord
             end
           end
 
+          def table(table, stream)
+            super
+
+            partition_bound_definition = @connection.table_partition_bound_definition(table)
+            if partition_bound_definition
+              partition_table = @connection.inherited_table_names(table).first
+
+              stream.puts "  execute \"ALTER TABLE ONLY #{partition_table} ATTACH PARTITION #{table} #{partition_bound_definition}\""
+            end
+          end
+
           def exclusion_constraints_in_create(table, stream)
             if (exclusion_constraints = @connection.exclusion_constraints(table)).any?
               exclusion_constraint_statements = exclusion_constraints.map do |exclusion_constraint|
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -162,11 +162,14 @@ module ActiveRecord
 
           options[:comment] = comment if comment
 
+          partition_bound_definition = table_partition_bound_definition(table_name)
           inherited_table_names = inherited_table_names(table_name).presence
 
-          options[:options] = "INHERITS (#{inherited_table_names.join(", ")})" if inherited_table_names
-
-          if !options[:options] && supports_native_partitioning?
+          # Can't use :options for declarative partitioning. Partition will be attached
+          # via a separate 'ALTER TABLE ... ATTACH PARTITION ...' call.
+          if inherited_table_names && !partition_bound_definition
+            options[:options] = "INHERITS (#{inherited_table_names.join(", ")})"
+          elsif supports_native_partitioning?
             partition_definition = table_partition_definition(table_name)
 
             options[:options] = "PARTITION BY #{partition_definition}" if partition_definition
@@ -190,7 +193,7 @@ module ActiveRecord
           end
         end
 
-        # Returns the partition definition of a given table
+        # Returns the partitioned table definition of a given table
         def table_partition_definition(table_name) # :nodoc:
           scope = quoted_scope(table_name, type: "BASE TABLE")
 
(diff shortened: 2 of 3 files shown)


balbesina:
this solution does not consider child tables' columns might differ from main table definition: indexes, constraints and default values.
[see](https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE) doc.

> Although all partitions must have the same columns as their partitioned parent, partitions may have their own indexes, constraints and default values, distinct from those of other partitions

that's why in the issue #54841 i mentioned attaching partition after child table is defined. this is the way pg_dump solves the problem.

balbesina:
i mean, when we only define table with options: `partiotion of parent for values in ('value')` child tables will lack indexes or default values if they have them different from parent table.

if we create tables like
```ruby
# migration
def up
  create_table :parent, id: false, options: 'partition by list(type)' do |t|
    t.string :type, null: false
    t.string :status, default: 'X', null: false
    t.timestamps
  end

  create_table :parent_a, id: false, options: "partition of parent for values in ('a')"
  create_table :parent_b, id: false, options: "partition of parent for values in ('b')"

  # index for child table only
  add_index :parent_b, :status, name: 'index_parent_b_on_status'
  # change default value for child table only
  change_column_default :parent_b, :status, 'Y'
end
```

this pr will dump tables similar to:
```ruby
create_table :parent, id: false, options: 'partition by list(type)' do |t|
  t.string :type, null: false
  t.string :status, default: 'X', null: false
end

create_table :parent_a, id: false, options: "partition of parent for values in ('a')"

create_table :parent_b, id: false, options: "partition of parent for values in ('b')"
```
and now we will miss index and default value for `status` column in table_b.

fatkodima:
Agreed. Using a separate `alter table` is the only way I can see currently how it can be done. Switched the PR to use it.

rajnishmitu:
Any updates on this? Still seeing the schema dump issue with partitioned tables in Rails 8.0.3.

PR #49930, opened 2023-11-05
Fix sqlite3 dropping database with bad permissions test

fatkodima:
Fixes #49928.

It fails here https://github.com/rails/rails/blob/febd21da3438d190f0c6ce296a114baee983d1e1/activerecord/lib/active_record/railties/databases.rake#L19 when trying to create a new connection down the road because of this recent change https://github.com/rails/rails/blob/febd21da3438d190f0c6ce296a114baee983d1e1/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L762 which tries to create WAL files.

We can ignore checking protected environments by setting an ENV variable, see https://github.com/rails/rails/blob/febd21da3438d190f0c6ce296a114baee983d1e1/activerecord/lib/active_record/tasks/database_tasks.rb#L65-L71

PR #54879, opened 2025-04-07
Add ActiveRecord::Tasks::AbstractTasks for subclassing per adapter behavior

### LINKED ISSUE #49928: Test fails expecting "Couldn't drop" to match output from rails("db:drop")
### Steps to reproduce
cd railties
bin/test test/application/rake/dbs_test.rb:271
on line 274-5: the test is expecting the output from the rails("db:drop") command to contain the string "Couldn't drop", but the output is 
bin/rails aborted!
ActiveRecord::StatementInvalid: SQLite3::ReadOnlyException: attempt to write a readonly database

### System configuration
Apple M2 Pro
Macos 13.6

**Rails version**: 7.2.0

**Ruby version**: 3.2.2

zzak:
The main motivation was to be able to override `check_current_protected_environment!` in the case of SQLite3 which raises an error on read-only databases in Linux.

As an alternative to #49930, which aims to fix #49928.

There might be a little more refactoring in there, but I wanted to open a PR to get feedback and see if anything fails in CI.

### DIFF
--- a/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
@@ -2,11 +2,7 @@
 
 module ActiveRecord
   module Tasks # :nodoc:
-    class SQLiteDatabaseTasks # :nodoc:
-      def self.using_database_configurations?
-        true
-      end
-
+    class SQLiteDatabaseTasks < AbstractTasks # :nodoc:
       def initialize(db_config, root = ActiveRecord::Tasks::DatabaseTasks.root)
         @db_config = db_config
         @root = root
@@ -37,10 +33,6 @@ module ActiveRecord
         connection.reconnect!
       end
 
-      def charset
-        connection.encoding
-      end
-
       def structure_dump(filename, extra_flags)
         args = []
         args.concat(Array(extra_flags)) if extra_flags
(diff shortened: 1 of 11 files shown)


activerecord/lib/active_record/tasks/adapter_tasks.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module Tasks # :nodoc:
+    class AdapterTasks # :nodoc:
```
byroot: This probably should be `AbstractTasks` or something?

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
```diff
@@ -54,36 +46,32 @@ def structure_dump(filename, extra_flags)
...
-      private
-        attr_reader :db_config, :root
-
-        def connection
-          ActiveRecord::Base.lease_connection
+      def check_current_protected_environment!(db_config, migration_class)
+        super
+      rescue ActiveRecord::StatementInvalid => e
+        case e.message
+        when /SQLite3::ReadOnlyException/
```
byroot: Have you try matching `e.cause` instead? Regexp on error messages isn't the end of the world, but is best avoided.

byroot (commented):
Generally makes sense to me.

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
zzak: Good catch (pun intended), fixed in fa89566972

PR #54886, opened 2025-04-08
Add fixture-level support for `json` column types

stevepolitodesign:
### Motivation / Background

It's not clear how best to set a JSON-backed column in a Fixture, so I figured providing a sensible default would help people get started.

Since `json` is a [supported type][1], I figured we should support it in a Fixture.

[1]: https://api.rubyonrails.org/classes/ActiveRecord/Type/Json.html

### Checklist

### DIFF
--- a/railties/lib/rails/generators/generated_attribute.rb
+++ b/railties/lib/rails/generators/generated_attribute.rb
@@ -148,6 +148,7 @@ module Rails
                      when :references, :belongs_to,
                           :attachment, :attachments,
                           :rich_text                   then nil
+                     when :json                        then {}
                      else
                        ""
         end
--- a/railties/lib/rails/generators/generated_attribute.rb
+++ b/railties/lib/rails/generators/generated_attribute.rb
@@ -148,7 +148,7 @@ module Rails
                      when :references, :belongs_to,
                           :attachment, :attachments,
                           :rich_text                   then nil
-                     when :json                        then {}
+                     when :json                        then '{"key": "value"}'
                      else
                        ""
         end
--- a/railties/test/generators/generated_attribute_test.rb
+++ b/railties/test/generators/generated_attribute_test.rb
@@ -137,7 +137,7 @@ class GeneratedAttributeTest < Rails::Generators::TestCase
   end
 
   def test_default_value_is_empty_hash
-    assert_field_default_value :json, {}
+    assert_field_default_value :json, '{"key": "value"}'
   end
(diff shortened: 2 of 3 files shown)


railties/lib/rails/generators/generated_attribute.rb
```diff
@@ -148,6 +148,7 @@ def default
                      when :references, :belongs_to,
                           :attachment, :attachments,
                           :rich_text                   then nil
+                     when :json                        then {}
```
stevepolitodesign, line 151: This might require more thought. We really need to return a String, not a Hash.

```suggestion
                     when :json                        then "{\"key\": \"value\"}"
```

But that's pretty cumbersome.

railties/test/generators/generated_attribute_test.rb
```diff
@@ -136,6 +136,10 @@ def test_default_value_is_empty_string
     end
   end
 
+  def test_default_value_is_empty_hash
```
neilvcarvalho, line 139: The test name references an empty hash, as it looks like it used to be.

Seeing the other tests, it seems like the pattern for the name is something like:
```suggestion
  def test_default_value_is_json
```

railties/lib/rails/generators/generated_attribute.rb
```diff
@@ -148,6 +148,7 @@ def default
                      when :references, :belongs_to,
                           :attachment, :attachments,
                           :rich_text                   then nil
+                     when :json                        then '{"key": "value"}'
```
neilvcarvalho, line 151: Is `jsonb` included in `json`? Given that `references` and `belongs_to` are both mentioned, and `belongs_to` is an alias of `references`, my guess is that it's not.

stevepolitodesign: Unfortunately, `jsonb` is not a [valid type](https://github.com/rails/rails/blob/c3c33947160e473a3d41555f8cf3aee7f3a48c76/railties/lib/rails/generators/generated_attribute.rb#L73-L77).

stevepolitodesign: Although, now that I'm looking at that method, I'm realizing it's dynamic based on the current database adapter. I think this test suite just happens to support `json`.

stevepolitodesign: So, maybe we could support `jsonb` too.

PR #42511, opened 2021-06-16
Fix capybara driver name conflict

cjlarose:
Fixes #42502

Related work: #28144, 7c9af60e5c26941dc9ec6a263f5fc5fe10050cba, #32468

Issue #32468, opened 2018-04-05
[SystemTest] driven_by per test with different browser capabilities


Edouard-chin:
Drafting this issue in case anyone wants to work on it, otherwise when I find time I will tackle it

### Problem

If you have multiple tests that uses the same driver, but each test suite needs to test with different browser capabilities then SystemTest will keep overriding Capybara default driver.
Example:

```ruby
  class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
    driven_by :selenium, using: :chrome, options: { args: ['--disable-translate'] }
  end

  class RegularChromeTest < ApplicationSystemTestCase
    test 'my test works on chrome' do
      ...
    end
  end

  class TestOnIphoneFour < ApplicationSystemTestCase
    driven_by :selenium, using: :chrome, options: { emulation: { device_name: 'iPhone 4'} }

    test 'my test works on iphone 4' do
      ...
    end
  end
```

<details>
  <summary> Disregard this explanation, this is not the correct one </summary>
The reason is because system test will keep overriding the Capybara driver with the same name `selenium`
https://github.com/rails/rails/blob/d514ce9199fac58e569482dc901e53d0526abdf7/actionpack/lib/action_dispatch/system_testing/driver.rb#L25
</details>

-------------------

I think one solution would be to decouple the driver name (what is currently passed to the `using` keyword) and make use of it only internally to Rails System test, then add an extra option to the `driven_by` method that will be used to register the driver on capybara.

Finally, inside the before_setup method, we would redefine the Capybara current's driver.
```ruby
def before_setup
  # ... old code
  Capybara.current_driver = self.class.driver.driver_name
end
```

Edouard-chin:
The explanation I provided on why this is happening is incorrect, however I confirm that a call to `driven_by` using the same driver's name will leak to other test class.

My previous explanation is incorrect because when **each** test is run the driver is getting properly reset https://github.com/rails/rails/blob/d514ce9199fac58e569482dc901e53d0526abdf7/actionpack/lib/action_dispatch/system_test_case.rb#L104

The root of the problem comes from the fact that SystemTest reuses the same name to register a driver(1 of selenium/webkit/poltergeist) each time `driven_by` is called.


Capybara has the concept of a "session", which to summarize things up is a browser window.
When test runs the same session (or the same browser window) will be reused for each test that needs the same driver. In the example below, because both tests uses the same driver, a single session will be used (you'll only see one browser window getting open)
```ruby
class Test < SystemTest
[...]

PR #54887, opened 2025-04-08
Pass explicit driver name as top-level argument

ngan:
https://github.com/rails/rails/pull/42511 implemented the ability to pass an explicit `name` for the capybara driver. But the implementation slightly differs from this comment (by the author): https://github.com/rails/rails/issues/32468#issuecomment-946056251

This PR changes it so that `name` is passed as a top-level argument instead of a driver option.
**Before**
```ruby
driven_by :selenium,
          using: :chrome,
          options: {
            name: 'chrome_with_translate_disabled',
            args: ['--disable-translate']
          }
```

**After**
```ruby
driven_by :selenium,
          using: :chrome,
          name: 'chrome_with_translate_disabled'
          options: {
            args: ['--disable-translate']
          }
```

### DIFF
--- a/actionpack/lib/action_dispatch/system_test_case.rb
+++ b/actionpack/lib/action_dispatch/system_test_case.rb
@@ -155,8 +155,8 @@ module ActionDispatch
     #     driven_by :selenium, using: :firefox
     #
     #     driven_by :selenium, using: :headless_firefox
-    def self.driven_by(driver, using: :chrome, screen_size: [1400, 1400], options: {}, &capabilities)
-      driver_options = { using: using, screen_size: screen_size, options: options }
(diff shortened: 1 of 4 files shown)


zzak:
The thing is `new` here is not public api, so any deprecation would need to happen on `driven_by`. I'm really sure on the context though.

ngan:
Hey everyone I just put up a PR mostly to start a conversation. It does look cosmetic, but imo feel more “correct.” Having the `name` argument in the `options` makes it look like it’s a configuration of the Capybara Driver (ie Selenium, Cuprite, Playeright). Instead, the name is really meant to represent the configuration name to be used for Capybara itself.

> so any deprecation would need to happen on driven_by.

Agree! I can get that fixed if we decide to move forward.

PR #54891, opened 2025-04-08
Introduce `config.active_record.strict_safe_http_methods`

flavorjones:
### Motivation / Background

Well-behaved web applications should not mutate database state in a ["safe" HTTP request](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) (e.g., a GET). There are a few benefits to following this guidance:

- Browsers can safely pre-fetch pages via GET for performance benefits
- Applications reap the benefits of Rails's CSRF protection by reserving mutations to POST et al
- Applications can use a proxy to route GET requests to a read-only replica, with all the benefits of geo locality and load balancing

Rails doesn't enforce this property of "safe" HTTP verbs today, providing little help and no safeguards for developers who want their application to follow the safe request semantics.

This PR introduces a configuration option that allows developers to enforce strict "safe" requests. When enabled, an exception is raised if database writes are attempted while serving a GET/HEAD/OPTIONS request.

### Detail

This PR introduces new middleware, `ActiveRecord::Middleware::StrictSafeHTTPMethods`, that wraps the request handling with a `ActiveRecord::Base.while_preventing_writes` block if the request is a `GET`, `HEAD`, or `OPTIONS`.

By setting `config.active_record.strict_safe_http_methods` to `true`, this middleware is added.

The config option defaults to `false` for now, but the intention is for a future Rails release to default to `true`.

When enabled during development, a readable error page is displayed showing the `ActiveRecord::ReadOnlyError` exception.

When enabled in production, will return a 500 status code.

### Additional information

- https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP

### Checklist

### DIFF
--- a/activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
+++ b/activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# :markup: markdown
+
+module ActiveRecord
+  module Middleware
+    # # Strict Safe HTTP Methods \Middleware
+    #
+    # This middleware prevents database writes while serving a "safe" HTTP request (`GET`, `HEAD`, or
+    # `OPTIONS`) by raising an `ActiveRecord::ReadOnlyError` exception if a database write is
+    # attempted.
+    #
+    # Disabled by default, this middleware can be added by setting
+    # `config.active_record.strict_safe_http_methods` to `true`.
+    class StrictSafeHTTPMethods
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        request = ActionDispatch::Request.new(env)
+
+        if request.get? || request.head? || request.options?
+          ActiveRecord::Base.while_preventing_writes do
+            @app.call(env)
+          end
+        else
+          @app.call(env)
+        end
+      end
+    end
+  end
+end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Introduce `config.active_record.strict_safe_http_methods` to prevent database writes while serving a "safe" HTTP request (`GET`, `HEAD`, or `OPTIONS`).
+
+    *Mike Dalessio*
+
 *   Encryption now supports `support_unencrypted_data: true` being set per-attribute.
 
     Previously this only worked if `ActiveRecord::Encryption.config.support_unencrypted_data == true`.
--- a/activerecord/lib/active_record.rb
+++ b/activerecord/lib/active_record.rb
@@ -167,6 +167,7 @@ module ActiveRecord
 
     autoload :DatabaseSelector
     autoload :ShardSelector
+    autoload :StrictSafeHTTPMethods
   end
 
   module Tasks
--- a/activerecord/lib/active_record/railtie.rb
+++ b/activerecord/lib/active_record/railtie.rb
@@ -39,6 +39,7 @@ module ActiveRecord
     config.active_record.belongs_to_required_validates_foreign_key = true
     config.active_record.generate_secure_token_on = :create
     config.active_record.use_legacy_signed_id_verifier = :generate_and_verify
+    config.active_record.strict_safe_http_methods = false
 
     config.active_record.queues = ActiveSupport::InheritableOptions.new
 
@@ -235,6 +236,7 @@ To keep using the current cache store, you can turn off cache versioning entirel
           :use_schema_cache_dump,
           :postgresql_adapter_decode_dates,
           :use_legacy_signed_id_verifier,
+          :strict_safe_http_methods,
         )
 
         configs_used_in_other_initializers.each do |k, v|
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -1817,6 +1817,12 @@ config.active_record.protocol_adapters.mysql = "trilogy"
 
 If no mapping is found, the protocol is used as the adapter name.
 
+#### `config.active_record.strict_safe_http_methods`
+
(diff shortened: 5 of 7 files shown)


activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
```diff
@@ -0,0 +1,33 @@
...
+    # `config.active_record.strict_safe_http_methods` to `true`.
+    class StrictSafeHTTPMethods
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        request = ActionDispatch::Request.new(env)
+
+        if request.get? || request.head? || request.options?
```
zzak, line 23: The [spec](https://httpwg.org/specs/rfc9110.html#safe.methods) also mentions `TRACE`. Should it be included or should the set of safe methods be configurable?

activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
```diff
@@ -0,0 +1,33 @@
...
+          ActiveRecord::Base.while_preventing_writes do
```
byroot, line 24: This doesn't handle multi-database, IIRC.

activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
```diff
@@ -0,0 +1,33 @@
...
+    #
+    # Disabled by default, this middleware can be added by setting
+    # `config.active_record.strict_safe_http_methods` to `true`.
+    class StrictSafeHTTPMethods
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        request = ActionDispatch::Request.new(env)
```
byroot, line 21: I'm not convinced this middleware should be in Active Record. 

Logic that depends on two frameworks are always a bit tricky, but I think it would makes more sense to be in `actionpack`, and would need to check for Active Record presence. A somewhat similar case is how Active Job defer push to after commit if Active Record is present.

byroot (commented):
> Well-behaved web applications should not mutate database state in a ["safe" HTTP request](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) (e.g., a GET).

This is a bit debatable. Overall yes, but it's perfectly legitimate to update the db, e.g. with a `.touch(:last_seen_at)`. Or if you are using Active Record for sessions, update some `last_active_at` property.

But I do get that you may want to heavily rely on read replicas and might want to enforce this in your app, I'm not not sure if backing this in Rails make it any easier though.

Also now that SolidCache was made the default, that's yet another potential for DB writes in a GET request.

activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
flavorjones: I don't think the routing layer supports `TRACE`, for example

https://github.com/flavorjones/rails/blob/91b0e30e146d0af8d12c8d23053167efa7ef45c7/actionpack/lib/action_dispatch/routing.rb#L260 only specifies

```
    HTTP_METHODS = [:get, :head, :post, :patch, :put, :delete, :options] # :nodoc:
```

flavorjones:
> I think this is what DatabaseSelector already does when enabled?

Yes, DatabaseSelector does set `preventing_writes` but the Resolver is designed for use with replicas that are declared in the database configuration. The use case I'm really addressing here is a server that is connected to a read-only sqlite replica.

Though, I guess what you mean is that I should consider introducing a new "resolver" instead of an entirely new middleware class. I'll give that a try.

flavorjones:
> it's perfectly legitimate to update the db, e.g. with a .touch(:last_seen_at). Or if you are using Active Record for sessions, update some last_active_at property.

@byroot Sure, everyone will have an opinion. But I think that for the use case of read-replica servers fronted by a geo-local proxy, even those writes are unwanted and I would like safeguards to prevent code from being introduced that breaks the "safe" semantics.

> This doesn't handle multi-database, IIRC. ... Also now that SolidCache was made the default, that's yet another potential for DB writes in a GET request

Yes, this is definitely something not addressed in this PR. I'm going to do some testing with a real app, and likely this design will need to change to allow specifying connection classes (and potentially, the complex way SolidCache implements its connection handling may require further changes to either that gem or to Rails to make this work).

activerecord/lib/active_record/middleware/strict_safe_http_methods.rb
flavorjones: When I started drafting this change, I put it in Action Pack, but changed my mind after seeing how similar the ShardSelector and DatabaseSelector middlewares are (see [`default_middlware_stack.rb` in particular](https://github.com/rails/rails/blob/1d1790b04d3930d65c2436bbe1d381af821d0ef4/railties/lib/rails/application/default_middleware_stack.rb#L95-L107)). That said, I don't have a strong opinion and will try moving it back to see how it feels.

flavorjones:
Thanks, everybody, for the quick reviews.

I think the desired behavior is complex enough, and this particular approach is still naïve enough, that I might close this for now and work out a more robust approach in a gem that I can put through its paces in some real applications.

I'll let that idea, and this PR, bake for a few days before doing so.

matthewd:
I imagine even just setting `ActiveRecord.reading_role` to `:primary` might be sufficient to get the default DatabaseSelector / Resolver there in the "I actually only have one database" case? 🤔 

https://github.com/rails/rails/blob/1d1790b04d3930d65c2436bbe1d381af821d0ef4/activerecord/lib/active_record/middleware/database_selector/resolver.rb#L63

(modulo different opinions about read-only verbs, but I think that's probably fine to refine -- and maybe also delegate to AD::Request)

PR #54892, opened 2025-04-08
Allow allocated Active Records to lookup associations

gmcgibbon:
Previously, the association cache isn't setup on allocated record objects, so association lookups will crash. Test frameworks like mocha use allocate to check for stubbable instance methods, which can trigger an association lookup (like [this one](https://github.com/rails/rails/blob/c3c33947160e473a3d41555f8cf3aee7f3a48c76/activestorage/app/models/active_storage/attachment.rb#L33) in Rails).

### Motivation / Background

This Pull Request has been created because allocated Active Record object can't lookup associations. I think this is incorrect.

### Detail

This Pull Request changes the association cache to be lazily initialized so that lookups can happen.

### Additional information

We might want to `init_internals` instead on allocated record objects. But, that would require using `send`, or making that method public. I think this makes more sense, unless we can think of a reason the other ivars need to be there too.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Allow allocated Active Records to lookup associations.
+
+    Previously, the association cache isn't setup on allocated record objects, so association
+    lookups will crash. Test frameworks like mocha use allocate to check for stubbable instance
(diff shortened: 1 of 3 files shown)


rodrigotoledo (commented):
But if the key does not exist?

PR #54372, opened 2025-01-27
[ci skip] Clarify the doc for creating links:

Edouard-chin:
- We mention how `*_path` works with various examples and suddenly use a code snippet that don't use the `*_path` method.

  Modified the doc to use `product_path(product.id)`, to make it explicit. While it's longer, it's clearer, especially since we are in the "getting started guide"

  I don't think mentioning how `link_to(product)` works is a good idea in this guide, the `link_to` accepts a crazy amount of different values with many heuristics to determine the url to generate. This is documented in the action view guide and in the API. Let's keep it simple in the Getting Started guide.

  Fix #54162

PR #54897, opened 2025-04-09
Improve explanation of path/link_to helpers in Getting Started guide

SimBed:
I followed through the **Getting Started with Rails** guide and found a few things unclear, in particular section [10.4. Showing Individual Products](https://guides.rubyonrails.org/getting_started.html#showing-individual-products) describing url and link_to helpers, so tried to improve it.

This is my first contribution, so very happy to receive feedback if I have not followed the protocols correctly.

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -1093,15 +1093,22 @@ domain.
 
 `_url` returns a full URL including the protocol, host, and port.
 
-URL helpers are useful for rendering emails that will be viewed outside of the
+URL generating helpers are useful for rendering emails that will be viewed outside of the
 browser.
 
-Combined with the `link_to` helper, we can generate anchor tags and use the URL
-helper to do this cleanly in Ruby. `link_to` accepts the display content for the
-link (`product.name`) and the path or URL to link to for the `href` attribute
-(`product`).
+We can use the `link_to` helper method together with the path or URL generating helper methods to create very clean anchor tags in Ruby. `link_to` takes an argument for the link's display content and an argument for the link's `href` attribute (the path or URL to link to). We could therefore create a link in Ruby like so:
 
-Let's refactor this to use these helpers:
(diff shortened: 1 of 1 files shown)


p8:
Thanks for the PR @SimBed !
Not explaining all the details and magic of link_to was specifically changed in #54925 to keep things simple in a getting started guide.

SimBed:
Could you explain please @p8? [#54925](https://github.com/rails/rails/pull/54925) doesn't seems to relate to the docs and doesn't mention link_to? Am i missing something?

Possibly in an effort to avoid explaining _link_to_ too much, the section has become unclear (baffling to a beginner I would think) and inconsistent with the approach to _redirect_to_ further on in the guide.

Edouard-chin:
👋 The change we made to simplify the `link_to` was https://github.com/rails/rails/pull/54372 

If we mention the shorter version `link_to('view', product)` as example we would have to go the extra mile of why this generates a link to view the product as opposed to viewing all products (which would be the case in the event that you pass a un-persisted product to `link_to`).

As mentioned in #54372, the `url_for` method (which what uses `link_to` internally), accepts a wide variety of objects (from Active Record model, to regular string, hashes, PORO) and it feels like a cognitive overload to explain that in a getting started guide.

SimBed:
Thanks for the link @Edouard-chin . Perhaps the title I used is unfortunate. I am not advocating for a detailed description of link_to, just a tidying up of the existing wording, so it is easier to follow (and consistent with the approach to redirect_to).

p8:
> Could you explain please @p8? [#54925](https://github.com/rails/rails/pull/54925) doesn't seems to relate to the docs and doesn't mention link_to? Am i missing something?

Sorry, posted the wrong PR. I meant https://github.com/rails/rails/pull/54372 as @Edouard-chin mentioned.

PR #54905, opened 2025-04-11
Add method to find Rails tz by IANA identifier

babilonczyk:
This commit introduces a method that allows mapping an IANA time zone identifier to a corresponding Rails time zone, even if the identifier is a linked (non-canonical) one.

Rails’ MAPPING only includes a subset of IANA canonical identifiers. Some identifiers—like US/Pacific—are valid IANA time zones but are not directly included in MAPPING. However, they often point to canonical zones that are mapped by Rails (e.g., US/Pacific links to America/Los_Angeles).

This method is useful when handling user input based on IANA time zones, allowing consistent mapping to Rails time zones regardless of whether the identifier is canonical or linked.

To ensure consistent behavior across platforms, this commit also makes tzinfo-data available on all platforms (not just Windows or JRuby). Without it, platforms like macOS and Linux may treat linked time zones as distinct, making canonical resolution unreliable. For example:

    TZInfo::Timezone.get("US/Pacific").canonical_identifier
    # => "US/Pacific" on macOS/Linux (without tzinfo-data)
    # => "America/Los_Angeles" with tzinfo-data

### Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -163,7 +163,7 @@ platforms :ruby, :windows do
   end
 end
 
-gem "tzinfo-data", platforms: [:windows, :jruby]
+gem "tzinfo-data", "~> 1.2025", ">= 1.2025.2"
 gem "wdm", ">= 0.1.0", platforms: [:windows]
 
 gem "launchy"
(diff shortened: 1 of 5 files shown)


babilonczyk:
[buildkite/rails](https://buildkite.com/rails/rails/builds/117988) fails on `actioncable integration (3.3)`. I don't think its directly related to any of the changes. Can I as any moderator to rerun the check?

PR #54906, opened 2025-04-11
Clarify Active Storage setup for importmap-rails [skip ci]

shouichi:


guides/source/active_storage_overview.md
```diff
@@ -1057,13 +1057,19 @@ directly from the client to the cloud.
...
     import * as ActiveStorage from "@rails/activestorage"
     ActiveStorage.start()
     ```
 
+    Only for importmap-rails users: pin activestorage.
+
+    ```ruby
+    pin "@rails/activestorage", to: "activestorage.esm.js"
+    ```
+
```
Edouard-chin, line 1072: 👍 It's correct.

However, I find now a bit confusing that we previously only mentioned how to import the activestorage package, and are now mixing it with "how to add it to your dependency". 

IMHO, we'd be better of copy/pasting the first bullet point of the documentation from Active Storage README, which is more comprehensive.

https://github.com/rails/rails/blob/59a64cd0f5339b78aadc1aa9bd86a5b09c478dc2/activestorage/README.md#L145

shouichi: Thanks for the feedback. We may need to add a new section for importmap-rails. Does that sound good to you?

shouichi: @Edouard-chin Copied the "direct upload" section from active storage README. Also copied the "CORS step" of the guide to active storage README so that they are complete and in sync.

Edouard-chin:
Thanks, I re-added a couple things that I felt was worth mentioning in the guides.

Issue #52858, opened 2024-09-10
Calling connection.active? after establish_connection returns false since AR 7.2.0


lvonk:
There is a difference in result in calling `ActiveRecord::Base.connection.active?` since 7.2 release after calling `ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "dev_db")`. 
Prior to 7.2 `active?` would return `true` for a postgresql database, but since 7.2 it returns `false`. In both scenarios the connection can be used right after. 
I read through the changelogs but couldn't find a change that caused this, but perhaps I missed something. Did the meaning of `active?` change or is connection setup more lazy then before? I any case the doc of the method doesn't seem to reflect reality anymore imho.

### Steps to reproduce
```ruby
# AR 7.1
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activerecord", "= 7.1.4"
  gem "pg"
end

require "active_record"

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "dev_db")
puts ActiveRecord::Base.connection.active?
# true
```

```ruby
# AR 7.2
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activerecord", "= 7.2.0"
  gem "pg"
end

require "active_record"

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "dev_db")
puts ActiveRecord::Base.connection.active?
# false
```
### Expected behavior
That `active?` returns `true` in AR 7.2 as the connection is ready to use.

### Actual behavior
`active?` returns `false` even though the connection can be used right after.

### System configuration
**Rails version**: NA

**Ruby version**: 3.2.2

**Postgres version**: 16.4

fatkodima:
This was changed in https://github.com/rails/rails/pull/49378 (cc @byroot)

lvonk:
Thanks for clarifying @fatkodima

byroot:
Yes, the connection is now fully lazy. Of course that makes the `establish_connection` name weird, but it has always been a bit of a lie. It doesn't establish a connection, just configure it. It being active was always just a side effect.

> I any case the doc of the method doesn't seem to reflect reality anymore imho.

You mean `#active?` ?

lvonk:
> You mean `#active?` ?

Yeah I meant `#active?`. But as you explained it, the docs do probably makes sense.

byroot:
Sounds good. If you want force the connection for some reason, you can use `#verify!`.

PR #54908, opened 2025-04-11
Add ActiveRecord::Base.establish_connection behavior change to Rails 7.2 release notes [ci skip]

sebaherrera07:
### Motivation / Background

While upgrading from Rails 7.1 to Rails 7.2, one issue we encountered—and that was not very straightforward to find—was that things broke because we assumed that `ActiveRecord::Base.establish_connection` would make the connection active.

Then we found this issue https://github.com/rails/rails/issues/52858 which helped us finally solve the issue, by calling `ActiveRecord::Base.connection.verify!` instead.

If this change had been part of the Rails 7.2 release notes, it would have saved us some time.

### Detail

This Pull Request adds this change as part of `ActiveRecord` Notable changes for `guides/source/7_2_release_notes.md`

### Checklist

### DIFF
--- a/guides/source/7_2_release_notes.md
+++ b/guides/source/7_2_release_notes.md
@@ -478,6 +478,8 @@ Please refer to the [Changelog][active-record] for detailed changes.
 
 ### Notable changes
 
+*   `ActiveRecord::Base.establish_connection` no longer sets `ActiveRecord::Base.connection.active?` to `true`. If you need this behavior, you can use `ActiveRecord::Base.connection.verify!` instead.
+
 Active Storage
 --------------
 


akhilgkrishnan (approved):
@sebaherrera07 Thanks for changes 👍

zzak:
This should be a changelog, not a notable change in the release notes. Was there not a changelog already?

sebaherrera07:
> This should be a changelog, not a notable change in the release notes. Was there not a changelog already?

I couldn't find anything here https://github.com/rails/rails/blob/7-2-stable/activerecord/CHANGELOG.md @zzak 

What's your suggestion now? Is it more likely that people will see that CHANGELOG instead of the Rails Guides? Or maybe the Rails Guides should have a link to the CHANGELOG (I think they do not have right now, [see link](https://guides.rubyonrails.org/7_2_release_notes.html))?

byroot:
I think it makes sense to have it there, even though I must admit I'm not super knowledgeable about how the doc and changelog etc are supposed to be lined out.

zzak:
Yeah I would go back on that and say Notable Changes is fine, but also a changelog would have been acceptable at the very least.

PR #54915, opened 2025-04-13
Allow to configure maximum cache key sizes

fatkodima:
When analyzing my project's cache store usage (we use redis), I noticed that 30% of the whole cache was used only by keys. I calculated, if we enforce the limit of 64 bytes, then the whole cache will become 15% smaller. There are some pretty large keys (in the order of hundreds of bytes). It will be tedious to ask people to fix their keys in all the places. Or introduce a monkey patch similar to this PR.

The approach used is similar to foreign keys naming scheme - truncate the key and append some digest of the rest to it.

Fortunately, in our codebase all the keys are prefixed with some unique identifiers that will help to find in the codebase where the key is set, so this approach will work perfectly.

<strike>It makes sense for me to only have this config for redis and memcache to reduce their memory usage, so I added this configuration only for them.</strike>

Example usage:
```ruby
# config/application.rb
config.cache_store = :redis_cache_store, { max_key_size: 64 }
```

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,17 @@
+*   Allow to configure maximum cache key sizes
+
+    When the key exceeds the configured limit (250 bytes by default), it will be truncated and
+    the digest of the rest of the key appended to it.
+
+    Note that previously `ActiveSupport::Cache::RedisCacheStore` allowed up to 1kb cache keys before
+    truncation, which is now reduced to 250 bytes.
+
+    ```ruby
+    config.cache_store = :redis_cache_store, { max_key_size: 64 }
+    ```
+
+    *fatkodima*
+
 *   Use `UNLINK` command instead of `DEL` in `ActiveSupport::Cache::RedisCacheStore` for non-blocking deletion.
 
     *Aron Roh*
--- a/activesupport/lib/active_support/cache.rb
+++ b/activesupport/lib/active_support/cache.rb
@@ -36,6 +36,7 @@ module ActiveSupport
       :serializer,
       :skip_nil,
       :raw,
+      :max_key_size,
     ]
 
     # Mapping of canonical option names to aliases that a store will recognize.
@@ -190,6 +191,9 @@ module ActiveSupport
       # Default +ConnectionPool+ options
       DEFAULT_POOL_OPTIONS = { size: 5, timeout: 5 }.freeze
 
+      # Keys are truncated with the Active Support digest if they exceed the limit.
+      MAX_KEY_SIZE = 250
+
       cattr_accessor :logger, instance_writer: true
       cattr_accessor :raise_on_invalid_cache_expiration_time, default: false
 
@@ -298,6 +302,9 @@ module ActiveSupport
 
         @options[:compress] = true unless @options.key?(:compress)
         @options[:compress_threshold] ||= DEFAULT_COMPRESS_LIMIT
+
+        @max_key_size = @options.delete(:max_key_size)
+        @max_key_size = MAX_KEY_SIZE if @max_key_size.nil? # allow 'false' as a value
 
         @coder = @options.delete(:coder) do
           legacy_serializer = Cache.format_version < 7.1 && !@options[:serializer]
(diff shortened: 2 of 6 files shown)


byroot:
Yes, large keys with redis is a pretty big footgun. Memcached has a max key size, but Redis don't, so you can use multi-gb keys.

But there again we should ensure all stores have that feature.

fatkodima:
Redis actually has a "limit" - 512mb. Currently, `MemCacheStore` enforces 250 bytes limit and `RedisCacheStore` enforces 1024 bytes limit. This PR only allows to configure these values, instead of using hardcoded values. `FileStore` also does its own magic.

> But there again we should ensure all stores have that feature.

By "feature" you mean ability to configure max cache sizes and truncate longer than these? For other stores, I think adding the same is only for consistency?

byroot:
> Redis actually has a "limit" - 512mb.

Oh, that's new I guess. That still is ridiculously large though.

> MemCacheStore enforces 250 bytes

Yes, because that's the actual Memcached server limit. 

> RedisCacheStore enforces 1024 bytes limit. 

I guess it's justified to not break some fringe use case that may rely on fairly large keys not being modified, but seems overly large to me. I think we could go with 250 like memcached.

> By "feature" you mean ability to configure max cache sizes and truncate longer than these? 

Yes.

> For other stores, I think adding the same is only for consistency?

Mostly, but It can still be useful for memory store or file store. It's not rare for memory store to be used as a stand-in in test or other environment, so it's good if it behave somewhat similarly to more production suitable stores.

It also kinda ensure the logic in in AS::Cache itself and not in all individual backends.

fatkodima:
Changed to truncate on 250 bytes by default for all stores.

I removed a few unused constants and an `attr_reader`, not sure if a good idea.

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+*   Allow to configure maximum cache key sizes
+
+    When the key exceeds the configured limit, it will be truncated and the digest of the rest
+    of the key appended to it.
```
byroot, line 4: We probably should note here that the redis behavior changed a tiny bit.

fatkodima:
Updated changelog and allow now `max_key_size: false` (but still enforcing a 250 bytes limit for memcached).

activesupport/lib/active_support/cache.rb
```diff
@@ -299,6 +303,9 @@ def initialize(options = nil)
         @options[:compress] = true unless @options.key?(:compress)
         @options[:compress_threshold] ||= DEFAULT_COMPRESS_LIMIT
 
+        @max_key_size = @options.delete(:max_key_size)
+        @max_key_size ||= MAX_KEY_SIZE if @max_key_size.nil? # allow 'false' as a value
```
byroot, line 307:
```suggestion
        @max_key_size = MAX_KEY_SIZE if @max_key_size.nil? # allow 'false' as a value
```

activesupport/lib/active_support/cache.rb
```diff
@@ -955,16 +962,31 @@ def validate_options(options)
...
           raise(ArgumentError, "key cannot be blank") if !str_key || str_key.empty?
 
           namespace_key str_key, options
         end
 
+        def truncate_key(key)
+          if key && @max_key_size && key.bytesize > @max_key_size
+            suffix = ":hash:#{ActiveSupport::Digest.hexdigest(key)}"
+            truncate_at = @max_key_size - suffix.bytesize
+            "#{key.byteslice(0, truncate_at)}#{suffix}"
```
byroot, line 984: One small concern here is that we could end up with invalid encoding if we truncate an utf8 string in the middle of a multi-byte character.

I need to test a few things as to how to better handle this.

byroot, line 984:
```suggestion
            key = key.byteslice(0, truncate_at)
            key.scrub!("")
            "#{key}#{suffix}"
```

If you need a test string, I did this:

```ruby
key = "a" * 25 + "€" + "b" * 50
```

byroot (approved):

activesupport/test/cache/stores/mem_cache_store_test.rb
```diff
@@ -241,6 +241,17 @@ def test_dalli_cache_nils
     assert_nil cache.fetch("not_nil_foo") { "bar" }
   end
 
+  def test_max_key_size
+    cache = lookup_store(namespace: nil, max_key_size: 64)
+    key = "foobar" * 20
+    truncated_key = "foobarfoobarfoobarfoobarfo:hash:f5e10320532914e3769e5c842bc3b64e"
```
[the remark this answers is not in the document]
fatkodima: no

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+*   Allow to configure maximum cache key sizes
+
+    When the key exceeds the configured limit (250 bytes by default), it will be truncated and
+    the digest of the rest of the key appended to it.
+
+    Note that previously `ActiveSupport::Cache::RedisCacheStore` allowed up to 1kb cache keys before
+    truncation, which is now reduced to 250 bytes.
```
jeremy, line 7: Ideally this new default would only apply to apps that opt in to 8.1 defaults. Then apps can carefully upgrade without inadvertently invalidating their caches.

byroot: My thinking was that keys larger than 250 bytes are rare enough that this isn't necessary. But if you think that's not the case, yes we can introduce another framework default.

PR #52353, opened 2024-07-17
Defer route drawing to the first request, or when url_helpers called

gmcgibbon:
### Motivation / Background

Redo of https://github.com/rails/rails/pull/51614, https://github.com/rails/rails/pull/52012

Executes the first routes reload when the route set is first used to serve a request, generate a url, or otherwise. Previously, this was executed unconditionally on boot, which can slow down boot time unnecessarily for larger apps with lots of routes.

This Pull Request has been created to reduce route draw time for larger apps by deferring it until necessary. 

### Detail

This Pull Request changes previous attempts were reverted. In this one, we're depending entirely on the route set class' methods to load routes instead of littering the codebase with manual reload calls. This should help prevent loading edge cases with the lazy route set. We're also swapping out the lazy route set for the normal one in eager-loaded environments to reduce impact on production.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54916, opened 2025-04-13
Handle when `routes` object is explicitly called inside `config/application.rb`

chaadow:
Related to https://github.com/rails/rails/pull/52353 cc @gmcgibbon 
More context here: https://github.com/rails/rails/pull/52353/files#r2041134698

## Motivation
I was investigating in my application why routes were loaded very early on. and I couldn't figure out why `LazyRouteSet` was never set on `config/route_set_class` then I realized that it was because of an explicit mention of `self.routes` inside of my `config/application.rb`

https://github.com/rails/rails/blob/b97917da8a801bf5360c3a9da913415bc8582735/railties/lib/rails/engine.rb#L544-L548

When it's first called, `config.route_set_class` is set to `ActionDispatch::Routing::RouteSet`
https://github.com/rails/rails/blob/b97917da8a801bf5360c3a9da913415bc8582735/railties/lib/rails/engine/configuration.rb#L47

The issue is that this initializer becomes useless, because it's already too late since `routes` was already initialized AND **memoized**
https://github.com/rails/rails/blob/b97917da8a801bf5360c3a9da913415bc8582735/railties/lib/rails/engine.rb#L590-L592

## Detail

I've created a failing test. one alternative is to wrap the explicit calls to `routes` in a `after_initialize`
```ruby
    config.after_initialize do
      config.exceptions_app = routes
    end
```
And update the docs ( https://guides.rubyonrails.org/configuring.html#config-exceptions-app ) accordingly

Or try to find a solution so that we duplicate the existing routes object inside the initializer

### DIFF
--- a/railties/lib/rails/engine.rb
+++ b/railties/lib/rails/engine.rb
@@ -588,7 +588,22 @@ module Rails
     end
 
     initializer :make_routes_lazy, before: :bootstrap_hook do |app|
-      config.route_set_class = LazyRouteSet if Rails.env.local?
+      if Rails.env.local?
+        config.route_set_class = LazyRouteSet
+
+        # If routes already exists we duplicate it
+        if @routes
+          appends = @routes.instance_variable_get(:@append).dup
+          prepends = @routes.instance_variable_get(:@prepend).dup
+          default_url_options = @routes.default_url_options.dup
+
+          @routes = config.route_set_class.new_with_config(config)
+
+          @routes.instance_variable_set(:@append, appends)
+          @routes.instance_variable_set(:@prepend, prepends)
+          @routes.default_url_options = default_url_options
+        end
+      end
     end
 
     initializer :add_routing_paths do |app|
--- a/railties/lib/rails/engine.rb
+++ b/railties/lib/rails/engine.rb
@@ -587,25 +587,6 @@ module Rails
       config.eager_load_paths.freeze
     end
 
-    initializer :make_routes_lazy, before: :bootstrap_hook do |app|
-      if Rails.env.local?
-        config.route_set_class = LazyRouteSet
-
-        # If routes already exists we duplicate it
-        if @routes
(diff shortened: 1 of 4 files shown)


chaadow:
I managed to find a solution and added an independent test, so that exisiting applications using
```ruby
    config.exceptions_app = routes
```

Still get lazy loading routes for free without any modification 👍🏼

railties/lib/rails/engine.rb
```diff
@@ -587,10 +587,6 @@ def load_seed
       config.eager_load_paths.freeze
     end
```
chaadow, line 589: I have to remove this intitializer as it is useless when `routes` is accessed directly

check the first commit for an attempt to keep it : https://github.com/rails/rails/pull/54916/commits/3346b19d0025a2048f975e7eec9ac9eac74ff908

I've tried the first commit on my rails app, and setting `exceptions_app` early on, or even in a `after_initialize` does not work

railties/lib/rails/engine/configuration.rb
```diff
@@ -44,7 +44,7 @@ def initialize(root = nil)
         @generators = app_generators.dup
         @middleware = Rails::Configuration::MiddlewareStackProxy.new
         @javascript_path = "javascript"
-        @route_set_class = ActionDispatch::Routing::RouteSet
+        @route_set_class = Rails.env.local? ? Rails::Engine::LazyRouteSet : ActionDispatch::Routing::RouteSet
```
chaadow, line 47: this is the safest way as when `routes` gets accessed outisde of the initialization process, `@route_set_class` will be properly set

gmcgibbon: The problem is we have specific use-cases (mostly tests) that draw routes early and expect the routes to be there without initialization. In this case, we can't use lazy routes because the initialization process is never fully completed. I think we need to find a different way to approach exceptions app setting.

PR #54917, opened 2025-04-13
Fix configuring `RedisCacheStore` with `raw: true`

fatkodima:
It is currently not possible to configure redis cache store with `raw: true` (works for memcache store):

```ruby
c = ActiveSupport::Cache.lookup_store(:redis_cache_store, raw: true)
c.read("foo") # => raises 'unknown keyword: :raw (ArgumentError)'
```

### DIFF
--- a/activesupport/test/cache/behaviors/cache_store_behavior.rb
+++ b/activesupport/test/cache/behaviors/cache_store_behavior.rb
@@ -742,6 +742,12 @@ module CacheStoreBehavior
     end
   end
 
+  def test_configuring_store_with_raw
+    cache = lookup_store(raw: true)
+    cache.write("foo", "bar")
+    assert_equal "bar", cache.read("foo")
+  end
+
   private
     def with_raise_on_invalid_cache_expiration_time(new_value, &block)
       old_value = ActiveSupport::Cache::Store.raise_on_invalid_cache_expiration_time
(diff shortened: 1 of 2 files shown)


activesupport/test/cache/stores/redis_cache_store_test.rb
```diff
@@ -507,6 +507,11 @@ class ClearTest < StoreTest
   end
 
   class RawTest < StoreTest
+    test "configuring store with 'raw: true'" do
+      cache = lookup_store(raw: true)
+      assert_equal("bar", cache.fetch("foo") { "bar" })
+    end
```
byroot, line 513: Should we put this in the generic store behavior test?

fatkodima: yes, updated.

activesupport/test/cache/behaviors/cache_store_behavior.rb
```diff
@@ -742,6 +742,11 @@ def test_setting_options_in_fetch_block_does_not_change_cache_options
     end
   end
 
+  def test_configuring_store_with_raw
+    cache = lookup_store(raw: true)
+    assert_equal("bar", cache.fetch("foo") { "bar" })
+  end
```
byroot, line 748: Not sure this is a good enough test.

Perhaps something like:

```ruby
cache.increment("foo")
assert_equal "1", cache.read("foo")
```
? This way we test the read was implictly raw.

byroot (approved):

PR #54918, opened 2025-04-13
Update active_record_encryption.md [ci skip]

pardeyke:
This might be obvious, but as a beginner it took me some time to get my head around why I was getting decryption errors.

### DIFF
--- a/guides/source/active_record_encryption.md
+++ b/guides/source/active_record_encryption.md
@@ -138,6 +138,7 @@ config.active_record.encryption.encrypt_fixtures = true
 ```
 
 When enabled, all the encryptable attributes will be encrypted according to the encryption settings defined in the model.
+If you are using `rails db:fixtures:load` in development, make sure to add the line above to `development.rb` as well.
 
 #### Action Text Fixtures
 


guides/source/active_record_encryption.md
```diff
@@ -138,6 +138,7 @@ config.active_record.encryption.encrypt_fixtures = true
 ```
 
 When enabled, all the encryptable attributes will be encrypted according to the encryption settings defined in the model.
+If you are using `rails db:fixtures:load` in development, make sure to add the line above to `development.rb` as well.
```
akhilgkrishnan, line 141: @pardeyke Not sure we need it here, can you explain what was the error that you're getting.

pardeyke: I use `rails db:fixtures:load` in my dev environment to not work on an empty app. I have multiple models with encrypted attributes. So when `rails db:fixtures:load` inserts the data in the database the are not encrypted unless you specify `config.active_record.encryption.encrypt_fixtures = true` in `development.rb`. If you then try to call a model with an encrypted attribute it can't be decrypted, because it was not encrypted in the first place and you get an decryption-error.

PR #54922, opened 2025-04-14
Fix `find_or_create_by!` not checking owner persistence:

### LINKED ISSUE #54920: `create_or_find_by` creates record when parent is not persisted
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports

  create_table :authors do |t|
    t.string :name
    t.timestamps
  end

  create_table :books do |t|
    t.string :name
    t.references :author
    t.timestamps
  end
end

  has_many :books
end

class Book < ActiveRecord::Base
  belongs_to :author
end

  def test_find_or_create_by_with_destroyed_owner
    author = Author.create!(name: "Test Author")
    author.destroy!

    assert_raises(ActiveRecord::RecordNotSaved) do
      author.books.create_or_find_by!(name: "Test Book")
    end
  end
end
```

### Expected behavior

A `ActiveRecord::RecordNotSaved` is raised with `You cannot call create unless the parent is saved`. I'd expect `create_or_find_by` to act like `create` and raise this error.

### Actual behavior

The new record is created. 

### System configuration

**Rails version**: 
Rails main since https://github.com/rails/rails/pull/54845

cc: @Edouard-chin

Edouard-chin:
### Motivation / Background

Fix #54920

#### Problem

In #54845, I modified a call to `Record#create` for `Record#new` followed by `Record#save`. I didn't seen that the behaviour between both is different on associations.

  https://github.com/rails/rails/blob/b97917da8a801bf5360c3a9da913415bc8582735/activerecord/lib/active_record/associations/collection_association.rb#L354-L356

```ruby
  car = Car.new
  car.wheels.create # Check if `car` is persisted before creating a wheel.

  wheel = car.wheels.new
  wheel.save # Doesn't check if `car` is persisted
```

#### Solution

The original reason why I modified `#create` is because it doesn't allow to get the result of the transaction status within a nested transaction. The documentation explains it and mention to raise a rollback so that the outer transaction sees it but it misses an important detail: 

**How to know whether something happened in the inner transaction since the Rollback error is silently swallowed.**
  https://github.com/rails/rails/blob/b97917da8a801bf5360c3a9da913415bc8582735/activerecord/lib/active_record/transactions.rb#L159-L165

The solution in this patch is to provide a flag on the record to know whether the last transaction was succesful.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -274,8 +274,8 @@ module ActiveRecord
       with_connection do |connection|
         record = nil
         transaction(requires_new: true) do
-          record = new(attributes, &block)
-          record.save || raise(ActiveRecord::Rollback)
+          record = create(attributes, &block)
+          record._last_transaction_return_status || raise(ActiveRecord::Rollback)
         end
         record
       rescue ActiveRecord::RecordNotUnique
@@ -294,8 +294,8 @@ module ActiveRecord
       with_connection do |connection|
         record = nil
         transaction(requires_new: true) do
-          record = new(attributes, &block)
-          record.save! || raise(ActiveRecord::Rollback)
+          record = create!(attributes, &block)
+          record._last_transaction_return_status || raise(ActiveRecord::Rollback)
         end
         record
       rescue ActiveRecord::RecordNotUnique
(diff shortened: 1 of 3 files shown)


jenshenny:
Thanks @Edouard-chin for putting up a fix so quickly! Tested this patch it solves all of the `find_or_create_by!` errors on my end 👍

PR #54923, opened 2025-04-14
Improve leap years counting performance in distance_of_time_in_words

Stazer:
### Motivation / Background

The `distance_of_time_in_words` function can lead to a denial of service if the given `from_time` and `to_time` arguments are far apart. I replaced the concerning code with a constant calculation.

### Detail

The `distance_of_time_in_words` function counts the leap years by using a range and the `count` function:

```
  def distance_of_time_in_words(from_time, to_time = 0, options = {})
    # ...    
    leap_years = (from_year > to_year) ? 0 : (from_year..to_year).count { |x| Date.leap?(x) }
    # ...    
```

This code can lead to a denial of service if `from_year` and `to_year` are far apart, as it iterates over a range generated from the difference of the two arguments, effectively blocking execution. I replaced the code with the following, which calculates the leap years in constant time:

```
    leap_years_up_to = ->(year) { (year / 4) - (year / 100) + (year / 400) }
    leap_years = (from_year > to_year) ? 0 : leap_years_up_to.call(to_year) - leap_years_up_to.call(from_year - 1)
```

This particular case may seem irrelevant at first, but I discovered this problem in a project of mine. Users who are able to set a timestamp can easily trigger this problem when the timestamp is passed somewhere to the `distance_of_time_in_words` function.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/date_helper.rb
+++ b/actionview/lib/action_view/helpers/date_helper.rb
@@ -136,8 +136,15 @@ module ActionView
             from_year += 1 if from_time.month >= 3
             to_year = to_time.year
             to_year -= 1 if to_time.month < 3
-            leap_years = (from_year > to_year) ? 0 : (from_year..to_year).count { |x| Date.leap?(x) }
+
+            leap_years = if from_year > to_year
+              0
+            else
+              fyear = from_year - 1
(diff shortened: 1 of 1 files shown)


byroot:
Nice trick. I just changed the code a bit not to rely on a lambda.

PR #54924, opened 2025-04-15
Fix `config.active_storage.touch_attachment_records` to work with eager loading

### LINKED ISSUE #54921: `touch_attachment_records` doesn't play nice with `eager_load`
I've recently been trying to reduce the amount of unnecessary `updated_at` updates when uploading using `ActiveStorage`, and found the amazing option `touch_attachment_records` introduced here: https://github.com/rails/rails/pull/49723

And it seemed to do exactly what it advertised, except only after a `reload!`.
It is because `ActiveStorage::Attachment` runs before `ActiveStorage::Engine` when `eager_load` is `true` -- and `eager_load` is very often `true`.

It is quite easy reproduce thanks to the example scripts 🙏 

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.root = __dir__
  config.hosts << "example.org"

  config.eager_load = true
  
  Rails.logger = config.logger

  config.active_storage.touch_attachment_records = false
    local: {
      root: Dir.tmpdir,
      service: "Disk"
    }
  }
end

Rails.application.initialize!

    t.timestamps null: false
  end
end

  has_one_attached :profile
end

  def test_touch_attachment_records
    user = User.create!(
      profile: {
        content_type: "text/plain",
        filename: "dummy.txt",
        io: ::StringIO.new("dummy"),
      }
    )
    assert_equal(ActiveStorage::Attachment.reflect_on_association(:record).options[:touch], false)
  end

  def test_updated_at
    user = User.create!
    updated_at = user.updated_at
    sleep(0.5)
    user.profile.attach(
      content_type: "text/plain",
      filename: "dummy.txt",
      io: ::StringIO.new("dummy"),
    )

    assert_equal(user.updated_at, updated_at)
  end
end
```

### Expected/Actual behavior

I expect both tests to pass when `eager_load` is either true or false.
But it passes only when `eager_load` is false.

If this is expected, are there any docs that point to a Rails way to reload specific models in `after_initialize`?

### System configuration

**Rails version**: Rails 7.2.2.1

**Ruby version**: `ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin24]`

fatkodima:
Fixes #54921.
Follow up to #49723.

@bendillinger can you confirm that this fixes the problem?

### DIFF
--- a/activestorage/lib/active_storage/engine.rb
+++ b/activestorage/lib/active_storage/engine.rb
@@ -82,6 +82,10 @@ module ActiveStorage
     end
 
     initializer "active_storage.configs" do
+      config.before_initialize do |app|
+        ActiveStorage.touch_attachment_records = app.config.active_storage.touch_attachment_records != false
+      end
+
       config.after_initialize do |app|
         ActiveStorage.logger            = app.config.active_storage.logger || Rails.logger
         ActiveStorage.variant_processor = app.config.active_storage.variant_processor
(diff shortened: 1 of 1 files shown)


bendillinger:
Works perfectly in the reproduction and in our application 🎉 
I learned a lot, thanks!

Issue #51228, opened 2024-03-01
It is not possible to handle "not a valid MIME type" error using the exceptions_app = self.routes when the value of CONTENT_TYPE header is invalid


ayuto-takasaki-LITALICO:
### Steps to reproduce

almost same as
https://github.com/rails/rails/issues/43094

1. Generate a new application:

```
rails new InvalidMimeTypeApp --skip-action-mailer --skip-action-mailbox --skip-action-text --skip-active-storage --skip-action-cable --skip-sprockets --skip-spring --skip-listen --skip-javascript --skip-turbolinks --skip-jbuilder --skip-test --skip-system-test --skip-bootsnap
```

2. Apply the following diff:

```
diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
index 09705d1..1b28aee 100644
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,2 +1,13 @@
 class ApplicationController < ActionController::Base
+    def about
+        render '/about'
+    end
+
+    def not_found
+        render json: { error: "some error message" }, status: :not_found
+    end
+    
+    def unsupported
+        render json: { error: "some error message" }, status: :unsupported_media_type
+    end
 end
diff --git a/config/application.rb b/config/application.rb
index e28b261..d960e3c 100644
--- a/config/application.rb
+++ b/config/application.rb
@@ -38,5 +38,7 @@ module InvalidMimeTypeApp
 
     # Don't generate system test files.
     config.generators.system_tests = nil
+
+    config.exceptions_app = self.routes
   end
 end
diff --git a/config/environments/development.rb b/config/environments/development.rb
index ff757b7..5b3fdfb 100644
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -12,7 +12,7 @@ Rails.application.configure do
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
 
   # Enable server timing
   config.server_timing = true
diff --git a/config/routes.rb b/config/routes.rb
index a125ef0..36308ca 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,4 +7,8 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "posts#index"
+
+  get "/about", to: "application#about"
+  match "/404", to: "application#not_found", via: :all
+  match "/415", to: "application#unsupported", via: :all
[...]

Issue #43094, opened 2021-08-25
It is not possible to handle "not a valid MIME type" error using the `exceptions_app = self.routes`


smt116:
### Steps to reproduce

1. Generate a new application:

    ```
    rails new InvalidMimeTypeApp --skip-action-mailer --skip-action-mailbox --skip-action-text --skip-active-storage --skip-action-cable --skip-sprockets --skip-spring --skip-listen --skip-javascript --skip-turbolinks --skip-jbuilder --skip-test --skip-system-test --skip-bootsnap
    ```

2. Apply the following diff:

    ```diff
    diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
    index 09705d1..f284d2b 100644
    --- a/app/controllers/application_controller.rb
    +++ b/app/controllers/application_controller.rb
    @@ -1,2 +1,9 @@
     class ApplicationController < ActionController::Base
    +  def about
    +    render "/about"
    +  end
    +
    +  def not_acceptable
    +    render json: { error: "some error message" }, status: :not_acceptable
    +  end
     end
    diff --git a/app/views/about.html.erb b/app/views/about.html.erb
    new file mode 100644
    index 0000000..345e6ae
    --- /dev/null
    +++ b/app/views/about.html.erb
    @@ -0,0 +1 @@
    +Test
    diff --git a/config/application.rb b/config/application.rb
    index d03cb9d..3627de7 100644
    --- a/config/application.rb
    +++ b/config/application.rb
    @@ -34,5 +34,7 @@ module InvalidMimeTypeApp
    
         # Don't generate system test files.
         config.generators.system_tests = nil
    +
    +    config.exceptions_app = self.routes
       end
     end
    diff --git a/config/environments/development.rb b/config/environments/development.rb
    index 6996a45..fc9d98f 100644
    --- a/config/environments/development.rb
    +++ b/config/environments/development.rb
    @@ -12,7 +12,7 @@ Rails.application.configure do
       config.eager_load = false
    
       # Show full error reports.
    -  config.consider_all_requests_local = true
    +  config.consider_all_requests_local = false
    
       # Enable/disable caching. By default caching is disabled.
       # Run rails dev:cache to toggle caching.
    diff --git a/config/routes.rb b/config/routes.rb
    index c06383a..43741dc 100644
    --- a/config/routes.rb
    +++ b/config/routes.rb
    @@ -1,3 +1,6 @@
     Rails.application.routes.draw do
[...]

PR #54925, opened 2025-04-15
Fallback to application/octet-stream content-type when a invalid multipart request is received:

### LINKED ISSUE #54096: Fix Invalid Parameter Error Handling: Raise More Specific Exception in POST Method
### Steps to reproduce
To reproduce this issue, send a malicious or exploratory request targeting an endpoint configured to accept multipart uploads. For example, use the following curl command:

```bash
$ curl --request POST 'http://localhost:3000/any_path' \
  --header 'Content-Type:multipart/form-data; boundary=----WebKitFormBoundaryxQm6bAVJJEu1FxWw' \
  --data '---invalid-boundary'
```
The Content-Type header specifies a boundary, but the request body contains invalid data, leading to the error.
  
### Expected behavior
The invalid request should be handled properly, and a 400 Bad Request error should be routed to the appropriate handler, such as:

```ruby 
match "/400", to: "errors#bad_request", via: :all, as: "bad_request"
```

### Actual behavior
The error is intercepted by the Rack middleware and forwarded to the Controller, but the exception raised `(ActionController::BadRequest: Invalid request parameters: Rack::Multipart::EmptyContentError) `occurs before it hits the controller cycle. The standard `rescue_from` mechanism does not catch the error, and the request does not reach the` /400` route handler.

### System configuration

**Rails version**: Rails 8.0.1
**Ruby version**: 3.3.5

### Steps to reproduce
The issue occurs in the `filtered_parameters` method in `ActionDispatch::Http::FilterParameters`, which is designed to handle parameter parsing errors:
```ruby
def filtered_parameters
  @filtered_parameters ||= parameter_filter.filter(parameters)
rescue ActionDispatch::Http::Parameters::ParseError
  @filtered_parameters = {}
end
```
However, the error raised is an `ActionController::BadRequest`, originating from the `ActionDispatch::Request#POST` method:

```ruby
rescue ActionDispatch::ParamsParser::ParseError, EOFError => e
  raise ActionController::BadRequest.new("Invalid request parameters: #{e.message}")  
```
Because ActionController::BadRequest is not caught by filtered_parameters, the request flow is interrupted before it can reach the configured error handling route.

### Proposed Solution
I suggest modifying the `ActionDispatch::Request#POST` method to raise a more specific exception, such as `ActionDispatch::Http::Parameters::ParseError`, instead of `ActionController::BadRequest`:
```ruby
raise ActionDispatch::Http::Parameters::ParseError.new("Invalid request parameters: #{e.message}")
```
### Community Question

Does this change make sense within the Rails architecture? Would it be appropriate to propose a pull request to implement this change, improving the handling of invalid parameters?

Thank you for your attention and feedback!

Edouard-chin:
### Motivation / Background

Fix #54096

#### Problem

When an application uses the `routes` object as its exceptions app, then it can't rescue invalid multipart requests, resulting in a 500 instead of gracefully show a 400.

```ruby
# config/application.rb
config.exceptions_app = routes

# Browser request
post "/send_photo", { "CONTENT_TYPE" => "multipart/form-data; boundary=..." }
# The 500 page is displayed to the user as there is no parts in the body.
```

#### Details

When sending a `multipart/form-data` request, the [RFC](https://tools.ietf.org/html/rfc2046#section-5.1) mentions that the body must contain one or more valid body parts.
If the body contains invalid parts or no parts at all, Rack raises a EOFError.

Rails rescue that error and raises a `BadRequest` instead.
https://github.com/rails/rails/blob/73761aae49cbefd4d825a7071f0a2fd215ad9740/actionpack/lib/action_dispatch/http/request.rb#L437-L438

When Rails calls the `routes` exceptions app, the same codepath is called again and the exception is raised again, this time not caught, resulting in a 500.

I originally figured that this should be handled by the application and create a wrapper around `routes` to rescue this, but we fixed 2 similar issues: #43094 and #51228.

#### Solution

Fallback to a text/html request.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/show_exceptions.rb
+++ b/actionpack/lib/action_dispatch/middleware/show_exceptions.rb
@@ -50,6 +50,7 @@ module ActionDispatch
         request.set_header "action_dispatch.original_path", request.path_info
         request.set_header "action_dispatch.original_request_method", request.raw_request_method
         fallback_to_html_format_if_invalid_mime_type(request)
+        fallback_to_octet_stream_when_invalid_multipart(request, wrapper)
         request.path_info = "/#{status}"
         request.request_method = "GET"
         response = @exceptions_app.call(request.env)
@@ -78,6 +79,15 @@ module ActionDispatch
         rescue ActionDispatch::Http::MimeNegotiation::InvalidType
           request.set_header "HTTP_ACCEPT", "text/html"
         end
+      end
+
+      def fallback_to_octet_stream_when_invalid_multipart(request, wrapper)
+        rack_error = wrapper.wrapped_causes.find { |wrapped| wrapped.exception.is_a?(::EOFError) }
+        return if rack_error.nil? || !request.form_data?
+
+        request.set_header "CONTENT_TYPE", "application/octet-stream"
+        # Required for Rack 3
+        request.delete_header(Rack::RACK_REQUEST_FORM_ERROR) if defined?(Rack::RACK_REQUEST_FORM_ERROR)
       end
 
       def pass_response(status)
--- a/railties/test/application/middleware/exceptions_test.rb
+++ b/railties/test/application/middleware/exceptions_test.rb
@@ -108,6 +108,35 @@ module ApplicationTests
       assert_not_equal "rendering index!", last_response.body
     end
 
+    test "renders when a multipart form is invalid and routes are used as the custom exceptions app" do
+      controller :foo, <<-RUBY
+        class FooController < ActionController::Base
+          def index
+            render plain: "rendering index!"
+          end
+
+          def bad_request
+            render(plain: "Oh no!", status: :bad_request)
+          end
+        end
+      RUBY
+
+      app_file "config/routes.rb", <<-RUBY
+        Rails.application.routes.draw do
+          post "/foo", to: "foo#index"
+          match "/400", to: "foo#bad_request", via: :all
+        end
+      RUBY
+
+      add_to_config "config.exceptions_app = self.routes"
+      add_to_config "config.action_dispatch.show_exceptions = :all"
+      add_to_config "config.consider_all_requests_local = false"
+
+      post "/foo", {}, { "HTTPS" => "on", "CONTENT_TYPE" => "multipart/form-data; boundary=----WebKitFormBoundaryxQm6bAVJJEu1FxWw", input: "---invalid" }
+      assert_equal 400, last_response.status
+      assert_equal "Oh no!", last_response.body
+    end
+
     test "uses custom exceptions app" do
       add_to_config <<-RUBY
         config.exceptions_app = lambda do |env|
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Sending a invalid multipart request will no longer present a 500 error to the user.
+
+    For applications using `Rails.application.routes` as their `exceptions_app`,
+    an invalid `multipart/form-data` request would present the 500 error page,
+    instead of routing it to the "/400" endpoint.
+
+    *Edouard Chin*
+
 *   Include cookie name when calculating maximum allowed size.
 
     *Hartley McGuire*


Edouard-chin:
Looks like I need to handle things differently on Rack 2. I'll fix it.

actionpack/lib/action_dispatch/middleware/show_exceptions.rb
```diff
@@ -80,6 +81,14 @@ def fallback_to_html_format_if_invalid_mime_type(request)
         end
       end
 
+      def fallback_to_html_when_invalid_multipart(request)
+        rack_error = request.get_header(Rack::RACK_REQUEST_FORM_ERROR)
+        return unless rack_error.is_a?(::EOFError)
+
+        request.delete_header(Rack::RACK_REQUEST_FORM_ERROR)
+        request.set_header "CONTENT_TYPE", "text/html"
```
byroot, line 89: `text/html` doesn't make much sense as a request `Content-Type`. 

If we need some fallback value, I think `application/octet-stream` would make more sense, no?

actionpack/lib/action_dispatch/middleware/show_exceptions.rb
```diff
@@ -80,6 +81,15 @@ def fallback_to_html_format_if_invalid_mime_type(request)
         end
       end
 
+      def fallback_to_html_when_invalid_multipart(request, wrapper)
+        rack_error = wrapper.wrapped_causes.find { |wrapped| wrapped.exception.is_a?(::EOFError) }
+        return if rack_error.nil? || !request.form_data?
+
+        request.set_header "CONTENT_TYPE", "application/octet-stream"
+        # Required for Rack 3
+        request.delete_header(Rack::RACK_REQUEST_FORM_ERROR) if defined?(Rack::RACK_REQUEST_FORM_ERROR)
```
Edouard-chin, line 90: https://github.com/rack/rack/blob/d73095f6c03007ceb7ccb4f76ebb785899c0dcde/lib/rack/request.rb#L502-L503

Rack 3 re-raise if this header is set.

actionpack/lib/action_dispatch/middleware/show_exceptions.rb
Edouard-chin: Thanks, yeah good point.

railties/test/application/middleware/exceptions_test.rb
```diff
@@ -108,6 +108,35 @@ def not_acceptable
...
+      RUBY
+
+      app_file "config/routes.rb", <<-RUBY
+        Rails.application.routes.draw do
+          post "/foo", to: "foo#index"
+          match "/400", to: "foo#bad_request", via: :all
+        end
+      RUBY
+
+      add_to_config "config.exceptions_app = self.routes"
```
byroot, line 131: I'm not sure I understand the point of setting routes as the `exceptions_app`. Not saying we shouldn't support it, but I'd like to understand the intent/benefits.

Edouard-chin: It's an old trick to allow having finer control on the error page displayed to a user.

By default, Rails just render a **static** template in `public/<error_code>.html` page, which an application can customise as they want.
You don't have access to all the goodies Action View provides in those templates (no helper, no access to any variables, basically no erb evaluation).

A few examples of what you can't do:

- You can't display the request id to the user (for communication with a support team to help find the request in the logs)
- Displaying a different error page conditionally wasn't possible (e.g if you wanted to A/B test)
- Before the `ErrorReporter` existed, you wouldn't have a way to send analytics such as StatsD event when an exception occurred.

-----

By setting the exceptions app to routes, you re-route the error and have access to a new request lifecycle and solves all the limitation from above and more.

> Not saying we shouldn't support it,

It's kind of a blury line. The `exceptions_app` accepts any Rack App, so it's already supported. Now comes the reverse question whether the RouteSet supports being the exception app (and whether this issue should be fixed by Rails).

There is no official documentation that it's supported, and the only occurrence of such mention was about creating a wrapper to workaround this type of issues https://github.com/rails/rails/blob/eb9ab54176e6bb0e23afb4083ab42ed005fd5d7d/guides/source/configuring.md?plain=1#L382

---

Feel free to push back, as I originally had the same questions but ultimately felt that https://github.com/rails/rails/issues/43094 and https://github.com/rails/rails/issues/51228 opened the door for official support though.

byroot: Yeah, I feel like `exception_app` is meant to be a simple and barebone app, as when you are recovering from errors, you probably should try to not do too much.

Edouard-chin: > I feel like exception_app is meant to be a simple and barebone app

Right, I'd agree with that too. Maybe in the future we could provide a more useful default exception app than just rendering a static template (and fix those kind of issues there).
I think the current default exception app is only useful when an app is at the "hello world" stage (e.g. adding a stylesheet through the asset pipeline and styling the page with tailwind is not even possible).

PR #54930, opened 2025-04-16
Add referential integrity note in AR::FixtureSet docs [ci skip]

danimashu:
### Motivation / Background

This PR adds a note into the `ActiveRecord::FixtureSet` API docs to give more visibility to the mechanism Rails uses when deleting/inserting fixtures before each test. Rails disables referential integrity constraints for it, so the data can be loaded in different order than required by their FKs dependencies. This Rails mechanism is so natural once you're aware about it, but it can catch some developers off-guard — it could lead to "non-integral" data in the test suite.

For example, in PostgreSQL, users who implemented custom checks via DB triggers can find their business rules bypassed during fixture loading ([Rails executes `ALTER TABLE ... DISABLE TRIGGER ALL`](https://github.com/rails/rails/blob/1ac40035478cba2e21d698ffa57b43355f5bed7b/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb#L7-L39)). Rails gracefully [verifies FK constraints later on](https://github.com/rails/rails/blob/1ac40035478cba2e21d698ffa57b43355f5bed7b/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb#L41-L65), if the config [`config.active_record.verify_foreign_keys_for_fixtures`](https://guides.rubyonrails.org/configuring.html#config-active-record-verify-foreign-keys-for-fixtures) is enabled, but other triggers would remain bypassed.

In MySQL, the functionality to verify FKs via `config.active_record.verify_foreign_keys_for_fixtures` is pending to be implemented. So the test DB could end up with non integral FKs.

### Additional information

- Note that the guides have [a mention about this particularity](https://github.com/rails/rails/blob/1ac40035478cba2e21d698ffa57b43355f5bed7b/guides/source/testing.md?plain=1#L761-L767), probably that's fine enough? What's not mentioned anywhere afaik is the existence of `config.active_record.verify_foreign_keys_for_fixtures`, except in [the config docs themselves](https://guides.rubyonrails.org/configuring.html#config-active-record-verify-foreign-keys-for-fixtures). Perhaps I should add that to my doc addition?
- Instead of just this note, a `== Referential Integrity` section at the bottom of the `ActiveRecord::FixtureSet` class documentation would be an option also. But perhaps it's too much noise. All the docs are very positive-oriented, and somehow this referential integrity topic is slightly be-careful-oriented, which I don't love. That's why I opted for just a very subtle awareness-oriented note: "Rails does this". That should ring a bell to those using PG triggers for example.

### DIFF
--- a/activerecord/lib/active_record/fixtures.rb
+++ b/activerecord/lib/active_record/fixtures.rb
@@ -80,6 +80,8 @@ module ActiveRecord
   #
   # The testing environment will automatically load all the fixtures into the database before each
   # test. To ensure consistent data, the environment deletes the fixtures before running the load.
+  # During this deletion and insertion of fixtures before each test, the environment temporarily
+  # disables the database's referential integrity constraints.
   #
   # In addition to being available in the database, the fixture's data may also be accessed by
   # using a special dynamic method, which has the same name as the model.


guides/source/testing.md
```diff
@@ -758,13 +758,17 @@ default. Loading involves three steps:
...
 
-TIP: In order to remove existing data from the database, Rails tries to disable
-referential integrity triggers (like foreign keys and check constraints). If you
-are getting permission errors on running tests, make sure the database user has
-the permission to disable these triggers in the testing environment. (In
-PostgreSQL, only superusers can disable all triggers. Read more about
-[permissions in the PostgreSQL
-docs](https://www.postgresql.org/docs/current/sql-altertable.html)).
+TIP: In order to delete and insert fixtures' data into the tables, Rails
+disables referential integrity triggers (like foreign keys and check
```
zzak, line 762: I don't think this change is necessary

danimashu: thx for the feedback @zzak! I was trying to be more explicit clarifying that Rails disables triggers not just for deleting existing data, but also for inserting it

zzak: This is just my feeling, but trying to cover all of the corner cases in this section is a bit much. We should just focus on introducing the core concept of fixtures in the context of Rails test strategy here.

danimashu: I totally see your point, appreciate it! Do you think it'd be helpful to add some clarification about this fixtures/referential-integrity topic somewhere in the API docs instead? Potentially in [activerecord/lib/active_record/fixtures.rb](
https://github.com/rails/rails/blob/17cde11cd13cf07a7d6c712cb1593881496abacb/activerecord/lib/active_record/fixtures.rb#L13-L526).

danimashu:
heya @zzak I'd really appreciate your eyes on this if you get a chance 🙏 Moved the referential integrity note to API docs - helps developers reckon with fixtures bypassing database triggers/constraints.

zzak:
Sorry for not checking back sooner.

To be honest, I'm not really sure what the benefit of documenting this is.

Could you add some detail as to why? And any additional information you can dig up about that behavior will help reviewers.

danimashu:
@zzak no worries at all, thanks so much for looking into this! I've added more context to [the op](https://github.com/rails/rails/pull/54930#issue-2999260813), would love your thoughts if you get a chance. If you feel this adds more noise than value, let's totally close the PR. I hope it could help some ppl increasing awareness about this mechanism, but it's somehow a corner-case for the understanding of fixtures.

zzak:
Thanks that is a good bit of information, and also #17982 has some more detail why the original comment was added.

I'd like to get feedback from other folks, especially ones who are working on and reviewing docs.

Are you sure `FixtureSet` is the one responsible for disabling the referenctial integrity checks?

For example, I found it to be part of the [test database setup](https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/activerecord/lib/active_record/test_databases.rb#L25), which calls [reconstruct_from_schema](https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/activerecord/lib/active_record/tasks/database_tasks.rb#L414-L421), which calls [truncate_tables](https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L225-L230).

I didn't look that deeply into it however.

danimashu:
Brilliant stuff, thanks for all those finds! 🙌 Basically this [disable_referential_integrity](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L642-L644) functionality is reached mainly for two use cases:
- **Truncation**: This is the code you're referencing. `FixtureSet` is really not the responsible for this case. This truncation is used when setting up the test DB. The calls start with [`create_and_load_fixtures`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/test_databases.rb#L13-L17) — also used in [`truncate_all`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/railties/databases.rake#L78-L81) rake task.
- **Deletion/Insertion**: This is the case `FixtureSet` has a lot of responsibility for. See 
[`FixtureSet.create_fixtures`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/fixtures.rb#L597-L616) which calls [`FixtureSet.read_and_insert`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/fixtures.rb#L648-L665) then [`FixtureSet.insert`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/fixtures.rb#L667-L696) to finally use [`insert_fixtures_set`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L489-L499)'s `DELETE`s & `INSERT`s statements. This `FixtureSet.create_fixtures` feels core method of the class, used directly in [`db:fixtures:load`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/railties/databases.rake#L431) rake task, and it's also reached after a couple of calls from the [`before_setup`](https://github.com/rails/rails/blob/c5fc494e39a42b53fbe37ddc6be99c41417b72fd/activerecord/lib/active_record/test_fixtures.rb#L10) hook.

zzak (approved):

Issue #38244, opened 2020-01-16
Unable to use ActionCable on GCP Memorystore


archonic:
### Steps to reproduce
GCP Memory store has the `client` redis command disabled. ActionCable attempts to use this command. If the Redis connection is configurable, this issue is avoidable.

This comment explains the workaround for Sidekiq:
https://github.com/mperham/sidekiq/issues/3518#issuecomment-390898217

Setting `id: nil` in cable.yml will stringify `nil`. I'm not sure if there's another way to configure the Redis connection for ActionCable.

### Expected behavior
ActionCable should be able to work with GCP Memorystore using the redis adapter.

### Actual behavior
Doesn't work. Here's my production log:
```
Registered connection (Z2lkOi8vY29sbGFib3Jhc2UvVXNlci8x)
#<Thread:0x000056542e52e760@/usr/local/bundle/gems/actioncable-6.0.2.1/lib/action_cable/subscription_adapter/redis.rb:151 run> terminated with exception (report_on_exception is true):
/usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:124:in `call': ERR unknown command `client`, with args beginning with: `setname`, `nil`,  (Redis::CommandError)
	from /usr/local/bundle/gems/newrelic_rpm-6.5.0.357/lib/new_relic/agent/instrumentation/redis.rb:74:in `call'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:107:in `block in connect'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:296:in `with_reconnect'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:103:in `connect'
	from /usr/local/bundle/gems/newrelic_rpm-6.5.0.357/lib/new_relic/agent/instrumentation/redis.rb:120:in `connect'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:279:in `with_socket_timeout'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/client.rb:136:in `call_loop'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/subscribe.rb:43:in `subscription'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis/subscribe.rb:12:in `subscribe'
	from /usr/local/bundle/gems/redis-4.0.3/lib/redis.rb:3010:in `_subscription'
[...]

PR #54931, opened 2025-04-16
Action Cable: Allow setting nil as subscription connection identifier for Redis

bubiche:
[workerB](https://moontower.gitstream.cm/v2/badge/collaboration-page?magicLinkId=f3nHQXY)

### Motivation / Background

I ran into the same issue as https://github.com/rails/rails/issues/38244. Google Cloud Memorystore and some other platforms block the command `CLIENT SETNAME` in redis so I need to set `id` to `nil`. I think it would be nice to be able to set `id: nil` in the config file instead of overwriting the redis connections factory as suggested [here](https://github.com/rails/rails/issues/38244#issuecomment-575454444) since people won't have to check if their overwritten method needs to be updated whenever they upgrade rails.

### Detail

- Check if `@server.config.cable[:id]` is already `nil` and don't fallback with `||=`
- Setting `identifier` to `nil` won't work for postgres since `pg_conn.escape_identifier(identifier)` can't convert `nil` to a string so this change is only for Redis

### Checklist

### DIFF
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,2 +1,5 @@
+*   Allow setting nil as subscription connection identifier for Redis.
+
+    *Nguyen Nguyen*
 
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/actioncable/CHANGELOG.md) for previous changes.
--- a/actioncable/lib/action_cable/subscription_adapter/base.rb
+++ b/actioncable/lib/action_cable/subscription_adapter/base.rb
@@ -29,7 +29,8 @@ module ActionCable
       end
 
       def identifier
-        @server.config.cable[:id] ||= "ActionCable-PID-#{$$}"
+        @server.config.cable[:id] = "ActionCable-PID-#{$$}" unless @server.config.cable.key?(:id)
+        @server.config.cable[:id]
       end
     end
   end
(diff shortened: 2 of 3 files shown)


byroot:
> Google Cloud Memorystore and some other platforms block the command CLIENT SETNAME in redis so I need to set id to nil

Urk, this is really a shitty behavior.

I think I'm fine with this feature, but I think the implementation is more complicated than it need to be. I don't think we need that `allow_explicit_nil_identifier?` definition.

Just change `@server.config.cable[:id] ||= "ActionCable-PID-#{$$}"` for:

```ruby
@server.config.cable[:id] = "ActionCable-PID-#{$$}" unless @server.config.cable.key?(:id)
```

And that should be enough.

bubiche:
> > Google Cloud Memorystore and some other platforms block the command CLIENT SETNAME in redis so I need to set id to nil
> 
> Urk, this is really a shitty behavior.
> 
> I think I'm fine with this feature, but I think the implementation is more complicated than it need to be. I don't think we need that `allow_explicit_nil_identifier?` definition.
> 
> Just change `@server.config.cable[:id] ||= "ActionCable-PID-#{$$}"` for:
> 
> ```ruby
> @server.config.cable[:id] = "ActionCable-PID-#{$$}" unless @server.config.cable.key?(:id)
> ```
> 
> And that should be enough.

Thank you, I did it at first too but realized if someone set `id: nil` for postgres then the postgres connection won't work since [this line](https://github.com/rails/rails/blob/26bc62a795a2627026af7574e350c2d29eda54ef/actioncable/lib/action_cable/subscription_adapter/postgresql.rb#L45) expects a string. I guess it's not that big of a gotcha.

PR #54938, opened 2025-04-18
Add assert_in_body/assert_not_in_body

dhh:
As a simple encapsulation of checking a response body for a piece of text without having to go through a heavy-duty DOM operation.

### DIFF
--- a/actionpack/lib/action_dispatch/testing/assertions/response.rb
+++ b/actionpack/lib/action_dispatch/testing/assertions/response.rb
@@ -71,6 +71,20 @@ module ActionDispatch
         assert_operator redirect_expected, :===, redirect_is, message
       end
 
+      # Asserts that the given +text+ is present somewhere in the response body.
+      #
+      #     assert_in_body fixture(:name).description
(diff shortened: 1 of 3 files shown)


actionpack/lib/action_dispatch/testing/assertions/response.rb
```diff
@@ -71,6 +71,20 @@ def assert_redirected_to(url_options = {}, options = {}, message = nil)
         assert_operator redirect_expected, :===, redirect_is, message
       end
 
+      # Asserts that the given +text+ is present somewhere in the response body.
+      #
+      #     assert_in_body fixture(:name).description
+      def assert_in_body(text)
+        assert_match(/#{text}/, @response.body)
```
Edouard-chin, line 78: Maybe we can escape the Regexp? I can see people using something like `assert_in_body("Name (optional)")` which wouldn't match because of the unescaped parenthesis.

dhh: Good idea 👍

PR #54942, opened 2025-04-18
Allow `direct` and `resolve` in engine routes at the root level

### LINKED ISSUE #31325: Can't resolve image into URL when using ActiveStorage in mountable Engine
### Steps to reproduce

- Create a new 5.2 app and setup ActiveStorage using a simple form as described [in this blogpost](http://blog.jasonkim.ca/blog/2017/11/13/Using-ActiveStorage-Today/)
- Create a new mountable engine and move the `PostsController` and views into the engine's namespace.
- Add the newly created engine to your app's Gemfile and mount the engine 

### Expected behavior
`url_for` should still work with ActiveStorage models.

### Actual behavior
ArgumentError when using `url_for` or `image_tag` after form submission.
`Can't resolve image into URL: undefined method `to_model' for #<ActiveStorage::Variant`

### System configuration
**Rails version**:
5.2.0.beta2

**Ruby version**:
2.4.2

pixeltrix:
### Motivation / Background

The detection of being at the root level of an engine's routeset was failing because an isolated engine would always have a default scope for the module and would call the `with_default_scope` method causing it to be always one level down. 

### Detail

This pull request fixes the problem by initializing the root scope with the default scope so there's no need to call the `with_default_scope` method.

### Additional information

This is the first stage in an attempt to fix #31325. The complete fix would require the `main_app` mounted helper to be present in the context of the block passed to `resolve` so an engine author could do this:

``` ruby
ShopApp::Engine.routes.draw do
  resolve("ActiveStorage::Blob") do |blob, options|
    main_app.route_for(ActiveStorage.resolve_model_to_route, blob, options)
  end
end
```

However that is currently not the case and it should be done in a different PR since this change is worthwhile outside of the context of the above issue.

### Checklist

* [x] This Pull Request is related to one change
* [x] Commit message has a detailed description of what changed and why
* [x] Tests are added or updated
* [x] CHANGELOG files are updated for the changed libraries

### DIFF
--- a/actionpack/lib/action_dispatch/routing/mapper.rb
+++ b/actionpack/lib/action_dispatch/routing/mapper.rb
@@ -2528,11 +2528,15 @@ module ActionDispatch
 
       DEFAULT = Object.new # :nodoc:
 
-      def initialize(set) # :nodoc:
+      def initialize(set, scope = nil) # :nodoc:
         @set = set
         @draw_paths = set.draw_paths
-        @scope = Scope.new(path_names: @set.resources_path_names)
         @concerns = {}
+
+        hash = { path_names: @set.resources_path_names }
+        hash.merge!(scope) unless scope.nil?
+
+        @scope = Scope.new(hash)
       end
 
       include Base
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Allow `direct` and `resolve` in engine routes at the root level
+
+    The detection of being at the root level of an engine's routeset was failing because an isolated engine
+    would always have a default scope for the module and would call the `with_default_scope` method causing
+    it to be always one level down. Fix the problem by initializing the root scope with the default scope so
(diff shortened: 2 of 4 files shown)


actionpack/lib/action_dispatch/routing/mapper.rb
```diff
@@ -2528,11 +2528,15 @@ def each
...
 
-      def initialize(set) # :nodoc:
+      def initialize(set, scope = nil) # :nodoc:
         @set = set
         @draw_paths = set.draw_paths
-        @scope = Scope.new(path_names: @set.resources_path_names)
         @concerns = {}
+
+        hash = { path_names: @set.resources_path_names }
+        hash.merge!(scope) unless scope.nil?
```
pixeltrix, line 2537: This is the bit that concerns me - all the examples I can find of `default_scope` being defined in an app are simple hashes but there may be something else in private repos. If you feel it's necessary we can extract the body of the `scope` method to a separate`build_scope` method and then use that here instead of just merging the hash.

PR #54944, opened 2025-04-18
fix show exception backtrace if backtrace_locations is empty

ermolaev:
### Motivation / Background

Some gems (or code in the app) in a `rescue` block create new errors, passing the backtrace, but the `backtrace_locations` will be empty.

This makes it impossible to understand where the error occurred, increasing debugging difficulties.

Currently, for some errors, only the error name is shown without a backtrace.

#### some problems gems
[redis-rb 1](https://github.com/redis/redis-rb/blob/317a0204d8361927757117845f825b42c26fac14/lib/redis/client.rb#L35)
[redis-rb 2](https://github.com/redis/redis-rb/blob/317a0204d8361927757117845f825b42c26fac14/cluster/lib/redis/cluster/client.rb#L32)

[redis-client 1](https://github.com/redis-rb/redis-client/blob/0acc9184f066ec3ebbe3335a922a7ff35a5e2152/lib/redis_client/ruby_connection.rb#L159)
[redis-client 2](https://github.com/redis-rb/redis-client/blob/0acc9184f066ec3ebbe3335a922a7ff35a5e2152/lib/redis_client.rb#L808)

#### example for emulate nil backtrace_locations

```ruby

require 'json'

NewError = Class.new(StandardError)

def parse_payload(text)
  JSON.parse(text)
rescue JSON::ParserError => ex
  raise NewError, "error", ex.backtrace
  # only ruby 3.4 support backtrace_locations as 3-th arg
  # raise NewError, "error", ex.backtrace_locations
end

begin
  parse_payload('{"wrong: "json"')
rescue => ex
  p ex.backtrace.size # => 4
  p ex.backtrace_locations&.size # => nil
end
```

https://docs.ruby-lang.org/en/3.4/Exception.html#method-i-set_backtrace

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
+++ b/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
@@ -52,7 +52,7 @@ module ActionDispatch
       @exception_class_name = exception.class.name
       @wrapped_causes = wrapped_causes_for(exception, backtrace_cleaner)
       @exception = exception
-      if exception.is_a?(SyntaxError)
+      if exception.is_a?(SyntaxError) || !exception.backtrace_locations
         @exception = ActiveSupport::SyntaxErrorProxy.new(exception)
       end
       @backtrace = build_backtrace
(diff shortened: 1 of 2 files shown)


byroot:
Looks good to me, but it would be valuable to add a test for it.

ermolaev:
closed this PR, because rails 8.0 show `exception_trace` for **all causes** unhandled exceptions

https://github.com/rails/rails/commit/f47ef36c4d0edf6965ec2c3069bc48b79ebe4245

Thanks :)

PR #54946, opened 2025-04-18
Extract `Or#or` from `WhereClause`

skipkayhil:
This code was added to `WhereClause` when `Or` was [changed][1] from `Binary` to `Nary`.

By defining `Or#or`, we can simplify `WhereClause`.

Since `Grouping` nodes are unwrapped, this also requires defining `#or`
on `SqlLiteral` (since it's a `String` and not a `Node`).

[1]: https://github.com/rails/rails/commit/bfcc13ab7cd9837bb60d9e5da4a2f9a7c371f9d0


byroot:
The CI failures look legit.

PR #54949, opened 2025-04-20
Avoid unnecessary splat allocations

jeremyevans:
In these 5 calls, an array is allocated for the splat, as the Ruby compiler cannot determine that the keyword/block expressions will not have side effects that will affect the splat. By assigning the related expression to a local variable and using the local variable, you can avoid the allocation for the splat.

CC @tenderlove (not even tens!)

### Motivation / Background

This is just a micro-optimization to avoid allocations and thus improve performance.  All behavior should be the same.

### Detail

These were the issue flagged by the performance warning in Ruby feature 21274 (not yet merged into Ruby itself): https://bugs.ruby-lang.org/issues/21274

### Checklist

* [X] Tests are added or updated if you fix a bug or add a feature (not applicable).
* [X] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included (not applicable).

### DIFF
--- a/actionpack/lib/action_dispatch/routing/mapper.rb
+++ b/actionpack/lib/action_dispatch/routing/mapper.rb
@@ -897,7 +897,8 @@ module ActionDispatch
             assign_deprecated_options(deprecated_options, mapping, :connect)
           end
 
-          match(*path_or_actions, as:, to:, controller:, action:, on:, defaults:, constraints:, anchor:, format:, path:, internal:, **mapping, via: [:get, :connect], &block)
+          via = [:get, :connect]
+          match(*path_or_actions, as:, to:, controller:, action:, on:, defaults:, constraints:, anchor:, format:, path:, internal:, **mapping, via:, &block)
           self
         end
       end
--- a/actionpack/lib/action_dispatch/middleware/stack.rb
+++ b/actionpack/lib/action_dispatch/middleware/stack.rb
@@ -41,6 +41,7 @@ module ActionDispatch
       end
 
       def build(app)
+        block = self.block
         klass.new(app, *args, &block)
       end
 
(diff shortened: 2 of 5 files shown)


actionpack/lib/action_dispatch/routing/mapper.rb
```diff
@@ -897,7 +897,8 @@ def connect(*path_or_actions, as: DEFAULT, to: nil, controller: nil, action: nil
             assign_deprecated_options(deprecated_options, mapping, :connect)
           end
 
-          match(*path_or_actions, as:, to:, controller:, action:, on:, defaults:, constraints:, anchor:, format:, path:, internal:, **mapping, via: [:get, :connect], &block)
+          via = [:get, :connect]
```
byroot, line 900: This one is a bit surprising. The array being a literal I'd expect it not to be subject to that problem.

byroot:
This is interesting, but all of these are rather cold paths, so I'm not sure it's really worth it.

actionpack/lib/action_dispatch/routing/mapper.rb
jeremyevans: This is probably possible to fix in Ruby's compilers.  You'd need to inspect all subnodes of NODE_LIST/ArrayNode and check if all of them are also safe.  Those nodes are not currently handled, so the default is to treat them as unsafe.

jeremyevans: Submitted a pull request to Ruby to avoid array allocation for literal arrays if all elements of the array are safe (which they are in this case): https://github.com/ruby/ruby/pull/13673.  With the optimization, you would not need to create a local variable to avoid the allocation.

PR #54951, opened 2025-04-20
Enable passing retryable `SqlLiteral`s to `#where`

skipkayhil:
### Motivation / Background

Previously, retryable SqlLiterals passed to `#where` would lose their retryability because both `#build_where_clause` and `WhereClause` would wrap them in non-retryable `SqlLiteral`s.

### Detail

To fix this problem, this commit updates `#build_where_clause` to check for `SqlLiterals` and updates `WhereClause` to assume that any predicate passed in will already be wrapped.

### Additional information

There were only a few places that passed Strings to `WhereClause` (`"1=0"` in `PredicateBuilder` and sanitized sql strings in `#build_where_clause`) that needed to be updated to use `Arel.sql`. The `WhereClause` tests also had Strings to wrap but they are really unit tests and aren't representative of what can be done with the public API

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1624,11 +1624,7 @@ module ActiveRecord
         case opts
         when String
           if rest.empty?
-            if Arel.arel_node?(opts)
-              parts = [opts]
-            else
-              parts = [Arel.sql(opts)]
-            end
+            parts = [Arel.sql(opts)]
           elsif rest.first.is_a?(Hash) && /:\w+/.match?(opts)
             parts = [build_named_bound_sql_literal(opts, rest.first)]
           elsif opts.include?("?")
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1624,13 +1624,17 @@ module ActiveRecord
         case opts
         when String
           if rest.empty?
-            parts = [Arel.sql(opts)]
+            if Arel.arel_node?(opts)
+              parts = [opts]
+            else
+              parts = [Arel.sql(opts)]
+            end
           elsif rest.first.is_a?(Hash) && /:\w+/.match?(opts)
             parts = [build_named_bound_sql_literal(opts, rest.first)]
           elsif opts.include?("?")
             parts = [build_bound_sql_literal(opts, rest)]
           else
-            parts = [model.sanitize_sql(rest.empty? ? opts : [opts, *rest])]
+            parts = [Arel.sql(model.sanitize_sql([opts, *rest]))]
           end
         when Hash
           opts = opts.transform_keys do |key|
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Enable passing retryable SqlLiterals to `#where`.
+
+    *Hartley McGuire*
+
 *   Set default for primary keys in `insert_all`/`upsert_all`.
 
     Previously in Postgres, updating and inserting new records in one upsert wasn't possible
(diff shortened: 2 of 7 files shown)


activerecord/lib/active_record/relation/query_methods.rb
```diff
@@ -1620,13 +1620,17 @@ def build_where_clause(opts, rest = []) # :nodoc:
...
+            else
+              parts = [Arel.sql(opts)]
+            end
           elsif rest.first.is_a?(Hash) && /:\w+/.match?(opts)
             parts = [build_named_bound_sql_literal(opts, rest.first)]
           elsif opts.include?("?")
             parts = [build_bound_sql_literal(opts, rest)]
           else
-            parts = [model.sanitize_sql(rest.empty? ? opts : [opts, *rest])]
+            parts = [Arel.sql(model.sanitize_sql([opts, *rest]))]
```
skipkayhil, line 1633: The `rest.empty?` check appears to have been dead code since 8e6a5deca626c944933211d7977f4e7e76869e74

activerecord/lib/active_record/relation/query_methods.rb
```diff
@@ -1620,13 +1620,17 @@ def build_where_clause(opts, rest = []) # :nodoc:
         case opts
         when String
           if rest.empty?
-            parts = [Arel.sql(opts)]
+            if Arel.arel_node?(opts)
+              parts = [opts]
+            else
+              parts = [Arel.sql(opts)]
+            end
```
byroot, line 1627: Shouldn't we just do that in `Arel.sql`? Right now it assumes it always receives a `String`, hence always wrap it, but we could make it return arel nodes unchanged.

skipkayhil: I tried searching for usages of `Arel.sql` that would pass a `SqlLiteral`, but it seems like there aren't many (any?) others than this one (I used `❯ rg '[^rf]\.sql\([^"%]'`).

I think it could make sense to do if the pattern was more common, but it seems like 90+% of the time a string is passed anyways.

skipkayhil: For a more definitive answer, I ran `ARCONN=sqlite3 bin/test` with this patch:

```
diff --git a/activerecord/lib/arel.rb b/activerecord/lib/arel.rb
index 738e80df35..cf82e60656 100644
--- a/activerecord/lib/arel.rb
+++ b/activerecord/lib/arel.rb
@@ -50,6 +50,7 @@ module Arel
   # Use this option only if the SQL is idempotent, as it could be executed
   # more than once.
   def self.sql(sql_string, *positional_binds, retryable: false, **named_binds)
+    raise if Nodes::SqlLiteral === sql_string
     if positional_binds.empty? && named_binds.empty?
       Arel::Nodes::SqlLiteral.new(sql_string, retryable: retryable)
     else
```

It fails on main but passes on this branch, so the two places touched by this PR are the only places we pass a `SqlLiteral` to `Arel.sql`.

byroot: I guess my point was more for the future, and for general API design. e.g. `Array([]) # => []`.

matthewd: I feel a bit non-specifically wary of defining `Arel.sql` to pass-through all Nodes (while conceding that's functionally what we do in several call-sites already)... but regardless of that, I do agree it seems reasonable/sensible for it to pass-through its own results rather than re-wrapping them.

skipkayhil: Updated to pass through `SqlLiterals` 👍 I agree passing through other Nodes probably doesn't make sense (and I think `Arel.sql` doesn't even accept other nodes at the moment since they can't be coerced to String).

summera:
@skipkayhil thanks for your awesome work on this! Will `BoundSqlLiteral`s also be supported soon?

skipkayhil:
> @skipkayhil thanks for your awesome work on this! Will `BoundSqlLiteral`s also be supported soon?

There's a linked PR just above your comment, but I can't speak to if/when it'll be merged

byroot:
Well, you're a committer now...

PR #54959, opened 2025-04-23
Add a load hook for `ActiveRecord::DatabaseConfigurations`

flavorjones:
### Motivation / Background

Summary: Rails does not offer a way to use an initializer to register a database config handler before database tasks are defined.

I'm using `ActiveRecord::DatabaseConfigurations.register_db_config_handler` in an engine to extend the default database configuration. One of the things the custom config class does is to conditionally override `#database_tasks?` to `false` so that the default database tasks are not defined for that database.

I would like to be able to call `.register_db_config_handler` from a railtie initializer, something like:

``` ruby
class Railtie < ::Rails::Railtie
  config.before_configuration do
      ActiveRecord::DatabaseConfigurations.register_db_config_handler { ... }
    end
  end
end
```

However, the default rake tasks in `active_record/railties/databases.rake` are loaded before the Active Record load hook is invoked. As a result, the database tasks that I want to omit are nevertheless created for the database.

Worth mentioning that `ActiveRecord::DatabaseConfigurations` seems to generally get autoloaded transitively by [this line](https://github.com/rails/rails/blob/556881c9a7846fd379e003ec3187c6d5bef5a832/activerecord/lib/active_record/railties/databases.rake#L7) when `ActiveRecord::Tasks::DatabaseTasks` gets resolved:

``` ruby
databases = ActiveRecord::Tasks::DatabaseTasks.setup_initial_database_yaml
```

### Detail

This PR adds a load hook to `activerecord/lib/active_record/database_configurations.rb`.

Using that load hook, a db config handler may be registered in a railtie before rake tasks are generated:

``` ruby
class Railtie < ::Rails::Railtie
  config.before_configuration do
    ActiveSupport.on_load(:active_record_database_configurations) do
      ActiveRecord::DatabaseConfigurations.register_db_config_handler { ... }
    end
  end
end
```

### Alternative approaches

One alternative, which I'm using now, is to explicitly require `database_configurations` and immediately register the db config handler, something like:

```ruby
# in the railtie
require "active_record/database_configurations"
ActiveRecord::DatabaseConfigurations.register_db_config_handler { ... }
```

and maybe this is OK, but I'd prefer to use a load hook.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/database_configurations.rb
+++ b/activerecord/lib/active_record/database_configurations.rb
@@ -36,9 +36,11 @@ module ActiveRecord
     # to respond to `sharded?`. To implement this define the following in an
     # initializer:
     #
-    #   ActiveRecord::DatabaseConfigurations.register_db_config_handler do |env_name, name, url, config|
-    #     next unless config.key?(:vitess)
-    #     VitessConfig.new(env_name, name, config)
+    #   ActiveSupport.on_load(:active_record_database_configurations) do
+    #     ActiveRecord::DatabaseConfigurations.register_db_config_handler do |env_name, name, url, config|
+    #       next unless config.key?(:vitess)
+    #       VitessConfig.new(env_name, name, config)
+    #     end
     #   end
     #
     # Note: applications must handle the condition in which custom config should be
@@ -306,4 +308,6 @@ module ActiveRecord
         url
       end
   end
+
+  ActiveSupport.run_load_hooks(:active_record_database_configurations, DatabaseConfigurations)
 end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add a load hook `active_record_database_configurations` for `ActiveRecord::DatabaseConfigurations`
+
+    *Mike Dalessio*
+
 *   Use `TRUE` and `FALSE` for SQLite queries with boolean columns.
 
     *Hartley McGuire*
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/database_configurations.rb
```diff
@@ -36,9 +36,11 @@ class InvalidConfigurationError < StandardError; end
...
     #
-    #   ActiveRecord::DatabaseConfigurations.register_db_config_handler do |env_name, name, url, config|
-    #     next unless config.key?(:vitess)
-    #     VitessConfig.new(env_name, name, config)
+    #   ActiveSupport.on_load(:active_record_database_configurations) do
+    #     ActiveRecord::DatabaseConfigurations.register_db_config_handler do |env_name, name, url, config|
+    #       next unless config.key?(:vitess)
+    #       VitessConfig.new(env_name, name, config)
+    #     end
     #   end
```
seanpdoyle, line 44: Could this code example demonstrate the fact that the hook executes within the context of the `ActiveRecord::DatabaseConfigurations` class?

```diff
    #   ActiveSupport.on_load(:active_record_database_configurations) do
-   #     ActiveRecord::DatabaseConfigurations.register_db_config_handler do |env_name, name, url, config|
+   #     register_db_config_handler do |env_name, name, url, config|
    #       next unless config.key?(:vitess)
    #       VitessConfig.new(env_name, name, config)
    #     end
```

flavorjones: Ooh, I do like that better, if I'm in this code again I will update the comment.

PR #54960, opened 2025-04-23
Sort schema cache columns and indexes per table when dumping

flavorjones:
### Motivation / Background

See https://github.com/rails/rails/issues/42717 for the original description of the problem, which is unstable sort order of tables and columns in the schema cache file.

The pull request https://github.com/rails/rails/pull/48824 partially solved this problem by sorting the tables by name, but the columns and indexes in each table may still appear in different order depending on how migrations were applied on a developer workstation.

### Detail

This extends the solution introduced in https://github.com/rails/rails/pull/48824 (3c7f48b8) by additionally sorting each table's columns and indexes by name.

### Additional information

See e403adf8 for prior art in sorting the columns in schema dumps.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/schema_cache.rb
+++ b/activerecord/lib/active_record/connection_adapters/schema_cache.rb
@@ -271,10 +271,10 @@ module ActiveRecord
       end
 
       def encode_with(coder) # :nodoc:
-        coder["columns"]          = @columns.sort.to_h
+        coder["columns"]          = @columns.sort.to_h.transform_values { _1.sort_by(&:name) }
         coder["primary_keys"]     = @primary_keys.sort.to_h
         coder["data_sources"]     = @data_sources.sort.to_h
-        coder["indexes"]          = @indexes.sort.to_h
+        coder["indexes"]          = @indexes.sort.to_h.transform_values { _1.sort_by(&:name) }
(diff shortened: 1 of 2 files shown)


byroot:
CI failure is legit

flavorjones:
@byroot good to go now.

activerecord/lib/active_record/connection_adapters/schema_cache.rb
```diff
@@ -271,10 +271,10 @@ def initialize_dup(other) # :nodoc:
       end
 
       def encode_with(coder) # :nodoc:
-        coder["columns"]          = @columns.sort.to_h
+        coder["columns"]          = @columns.sort.to_h.transform_values { _1.sort_by(&:name) }
```
byroot, line 274: Actually this turns out to be a problem for code that depends on column ordering (debatable whether it's a good idea, but a problem regardless).

flavorjones: @byroot Interesting. Want me to add a config to turn it off (to opt out)?

flavorjones: See https://github.com/rails/rails/pull/56566

PR #54961, opened 2025-04-23
Improve the default exceptions app:

Edouard-chin:
### Motivation / Background

I'd like to make it easier to do simple things on error pages, such as:

* Customising them with tailwind, or allow to add stylesheets
* Add dynamic rendering to allow for displaying links
* Display images

### Detail

The Rails default exception app is very limited in term of features because it can only render static page. While the recent redesign of the error pages are super nice, branding error pages is a feature that should be easy to do, but unfortunately in Rails, it's not.

There is 3 ways to solve this problem, but none are easy:

1) Set the `exceptions_app = routes`
2) Add a bunch of `rescue_from` in your ApplicationController. This effectively bypass the exception app.
3) Create your own exception app from scratch.

-----------------

Here are all the issues with those 3 approach:

1) Setting the `exceptions_app` to `routes` is overkilled, as the error is re-routed and the entire middleware stack is reperformed.
Some exceptions that originally triggered the exception app, will resurface (as the same codepath will get called). See [#43094](43094) [#51228](51228) [#54925](51228).
This solution (while the most popular) was also never officialy documented.
2) It works okay ish, but you may miss some exceptions to rescue, it also bloats your controller with rescue_from clause. An example of a real app using this approach is Mastodon.
3) This is the official way, works great but you have to implement the rendering logic from scratch.

Worth remembering that we have to do all of this to simply display an image :( .

### Additional information

I'd like to provide a stripped down controller with minimum functionalities to be able to do all of the above without writing any line of code.
I went with a Metal controller and added the necessary rendering concerns.

All previous behaviours were conserved. It's also backward compatible and will allow rendering existing and static `404.html` page.

This PR is big enough and if this feature is accepted, I'd like to rename the 404.html files generated by Rails, and maybe move them from the public folder? Not sure if those shouldn't belong to the views.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   All error pages (404, 422 ...) can now be rendered dynamically.
+
+    Rendering error pages to display dynamic content required to modify the
+    default Rails `exceptions_app`. It was not possible to style the pages
+    with tailwind or have layouts.
+
+    This is now possible by renaming the `404.html` to `404.html.erb` and
+    modifying the templates as any other views.
+
+    *Edouard Chin*
+
 *   Add assert_in_body/assert_not_in_body as the simplest way to check if a piece of text is in the response body.
 
     *DHH*
--- a/actionpack/lib/action_dispatch.rb
+++ b/actionpack/lib/action_dispatch.rb
@@ -88,6 +88,7 @@ module ActionDispatch
   end
 
   autoload :Constants
+  autoload :ExceptionsController, "action_dispatch/middleware/public_exceptions"
   autoload :Journey
   autoload :MiddlewareStack, "action_dispatch/middleware/stack"
   autoload :Routing
--- a/actionpack/lib/action_dispatch/middleware/public_exceptions.rb
+++ b/actionpack/lib/action_dispatch/middleware/public_exceptions.rb
@@ -2,7 +2,117 @@
 
 # :markup: markdown
 
+require "action_view"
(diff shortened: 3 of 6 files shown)


byroot:
I haven't reviewed your PR, but:

> The Rails default exception app is very limited in term of features because it can only render static page.

Note that IMO this is a feature. There's nothing worse than failing to render an error page. I get there are cases where you absolutely need a dynamic error page, but ideally you only use a static, self contained (not even asset links, all inline CSS) html page.

Edouard-chin:
I agree with you on theory. But it’s a hard sell in practice to tell users to render static pages with the argument that it may break, especially if there are no custom logic and you just need rendering capabilities.

Without rendering, this means: no links (or break as soon as a route gets changed), no assets pipeline (no images, javascript, icons, font…), no access to any ui components (good luck having uniformed error pages, or tweaking colour scheme) and a ton of duplication making it painful for future layout changes. If the Rails guideline is to say: “render static pages”, then users will need to start hardcoding link to resources, and ultimately you have a higher chance to display broken pages as time pass and resources are no longer available. Not sure what’s worth , a page that renders but is aesthetically awful or a page that doesn’t render.

Not sure if this is an argument the team care about when making decision, but I couldn’t find any framework (from the major ones) that doesn’t offer dynamic rendering of error pages. All popular Rails shop, open source or not also use rendering.

k0va1:
any updates here guys?

Edouard-chin:

PR #54962, opened 2025-04-24
Set default for primary keys in `insert_all`/`upsert_all`

jenshenny:
### Motivation / Background

In MySQL/SQLite, when a null value for a primary key gets passed into an insert, the auto increment will automatically set the value. However, in Postgres, a null value for a primary key will raise a `PG::NotNullViolation`. This limits the ability to update and insert new records in one bulk upsert.

```ruby
Book.upsert_all [
  { id: 1, name: "New edition" },
  { id: nil, name: "New edition 2" },
  { id: nil, name: "New edition 3" },
]
```

### Detail

When determining the value to insert, I chose to set the value to `default_insert_value` if the attribute was the primary key and the value is `nil`. This would translate to `Arel.sql("DEFAULT")` for Postgres but remain `nil` for some adapters like MySQL as primary key columns are auto increment columns.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Set default for primary keys in `insert_all`/`upsert_all`.
+
+    Previously in Postgres, updating and inserting new records in one upsert wasn't possible
+    due to null primary key values. `nil` primary key values passed into `insert_all`/`upsert_all`
+    are now implicitly set to the default insert value specified by adapter.
+
+    *Jenny Shen*
+
 *   Add a load hook `active_record_database_configurations` for `ActiveRecord::DatabaseConfigurations`
 
     *Mike Dalessio*
--- a/activerecord/lib/active_record/connection_adapters/mysql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/mysql/database_statements.rb
@@ -45,14 +45,14 @@ module ActiveRecord
           end
         end
 
+        def default_insert_value(column) # :nodoc:
+          super unless column.auto_increment?
+        end
+
         private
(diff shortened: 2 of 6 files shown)


activerecord/lib/active_record/connection_adapters/mysql/database_statements.rb
```diff
@@ -45,16 +45,16 @@ def build_explain_clause(options = [])
           end
         end
 
+        def default_insert_value(column)
```
eileencodes, line 48:
```suggestion
        def default_insert_value(column) # :nodoc:
```

activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
```diff
@@ -61,6 +61,14 @@ def reset_isolation_level # :nodoc:
           @previous_read_uncommitted = nil
         end
 
+        def default_insert_value(column)
```
eileencodes, line 64:
```suggestion
        def default_insert_value(column) # :nodoc:
```

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+*   Set default for primary keys in `insert_all`/`upsert_all`
```
eileencodes, line 1:
```suggestion
*   Set default for primary keys in `insert_all`/`upsert_all`.
```

eileencodes (changes requested):
We don't want to expose any of the `default_insert_value` methods so the non-abstract ones should also have `nodoc` added. Otherwise I think this looks good.

eileencodes (approved):

PR #54967, opened 2025-04-24
[actionview] Check choice type before calling .second to detect grouped_choices in Select helper

### LINKED ISSUE #54966: ActionView::Helpers::Tags::Select's group_choices detection can unknowingly end up calling the `second` method
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "8.0.2"

  gem "sqlite3"
end

require "action_view/railtie"
require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end
end

end

  def test_stuff

    render inline: <<~ERB, locals: { options: [Post] }
      <%= select(nil, "class", options) %>
    ERB

    assert_equal "Post", Post.to_s
    assert_equal <<~HTML, rendered.to_s
      <select name="class" id="class"><option value="Post">Post</option></select>
    HTML
  end
end
```

### Expected behavior

Since I'm passing a unidimensional array, I would expect:
1. The helper to not detect that I'm passing group_choices
2. To render my option of Class type using `to_s`

See the test above

### Actual behavior

[`group_choices?`](https://github.com/rails/rails/blob/22a61df687a5408a37ed0b9dcf73c08b250dc9be/actionview/lib/action_view/helpers/tags/select.rb#L39-L40) ends up calling [`Post.second`](https://github.com/rails/rails/blob/22a61df687a5408a37ed0b9dcf73c08b250dc9be/activerecord/lib/active_record/querying.rb#L7), triggering a DB load, resulting in a nil object:

```
I, [2025-04-24T09:23:50.508624 #88242]  INFO -- : -------------------
I, [2025-04-24T09:23:50.508650 #88242]  INFO -- : BugTest: test_stuff
I, [2025-04-24T09:23:50.508658 #88242]  INFO -- : -------------------
D, [2025-04-24T09:23:50.516487 #88242] DEBUG -- :   Rendering inline template
D, [2025-04-24T09:23:50.523872 #88242] DEBUG -- :   [1m[36mPost Load (0.1ms)[0m  [1m[34mSELECT "posts".* FROM "posts" ORDER BY "posts"."id" ASC LIMIT ? OFFSET ?[0m  [["LIMIT", 1], ["OFFSET", 1]]
D, [2025-04-24T09:23:50.526037 #88242] DEBUG -- :   [1m[36mPost Load (0.1ms)[0m  [1m[34mSELECT "posts".* FROM "posts" ORDER BY "posts"."id" ASC LIMIT ?[0m  [["LIMIT", 1]]
D, [2025-04-24T09:23:50.527187 #88242] DEBUG -- :   [1m[36mPost Load (0.0ms)[0m  [1m[34mSELECT "posts".* FROM "posts" ORDER BY "posts"."id" DESC LIMIT ?[0m  [["LIMIT", 1]]
I, [2025-04-24T09:23:50.527786 #88242]  INFO -- :   Rendered inline template (Duration: 10.8ms | GC: 0.0ms)

  1) Failure:
BugTest#test_stuff [bug.rb:53]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
-"<select name=\"class\" id=\"class\"><option value=\"Post\">Post</option></select>
+"<select name=\"class\" id=\"class\"><option selected=\"selected\" value=\"\"></option></select>
 "
```

If the table isn't defined, it's even worse, it crashes with `no such table: posts:`

I understand that calling the form helper with a Class may not be supported, but I don't think the form helper should be calling `.second` without checking the type first.

### System configuration

**Rails version**: 
8.0.2

**Ruby version**: 
3.3.4

lavoiesl:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/54966

### Detail

This Pull Request changes `ActionView::Helpers::Tags::Select#grouped_choices?` to do type checking instead of duck typing.

This avoids calling `second` on arbitrary objects.

In the issue above, it's calling `.second` on an `ActiveRecord` class, triggering a database call.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/tags/select.rb
+++ b/actionview/lib/action_view/helpers/tags/select.rb
@@ -37,7 +37,12 @@ module ActionView
           #   [nil, []]
           #   { nil => [] }
           def grouped_choices?
-            !@choices.blank? && @choices.first.respond_to?(:second) && Array === @choices.first.second
+            return false if @choices.blank?
+
+            first_choice = @choices.first
+            return false unless first_choice.is_a?(Enumerable)
+
+            first_choice.second.is_a?(Array)
           end
       end
     end
--- a/actionview/test/template/form_options_helper_test.rb
+++ b/actionview/test/template/form_options_helper_test.rb
@@ -537,6 +537,13 @@ class FormOptionsHelperTest < ActionView::TestCase
     )
   end
 
+  def test_select_with_class
+    assert_dom_equal(
+      "<select name=\"post[class]\" id=\"post_class\"><option value=\"Struct::Post\">Struct::Post</option></select>",
+      select(:post, :class, [Post])
+    )
+  end
+
   def test_select_without_multiple
     assert_dom_equal(
       "<select id=\"post_category\" name=\"post[category]\"></select>",


lavoiesl:
The CI failure seems unrelated to this PR

actionview/lib/action_view/helpers/tags/select.rb
```diff
@@ -37,7 +37,12 @@ def render
           #   [nil, []]
           #   { nil => [] }
           def grouped_choices?
-            !@choices.blank? && @choices.first.respond_to?(:second) && Array === @choices.first.second
+            return false if @choices.blank?
+
+            first_choice = @choices.first
+            return false unless first_choice.is_a?(Array) || first_choice.is_a?(Hash)
```
Edouard-chin, line 43: Maybe we can check if it's an Enumerable? It seems that this was the intended behaviour according to https://github.com/rails/rails/commit/efe62b71d52c40849ead8855fb4f42fae91d4c54

lavoiesl: Good point.

Edouard-chin: I think I'm wrong actually, I edited my comment.

lavoiesl: Would we want to allow grouped_choices for an Enumerable of Array?

I _think_ so?

Edouard-chin: > Would we want to allow grouped_choices for an array of Enumerable?

Ah yes, that would make sense to me

PR #54970, opened 2025-04-24
Expose ActiveSupport::Cache::MemCacheStore client

PatrickTulskie:
### Motivation / Background

My team maintains several Rails monoliths where we have begun to leverage [Pitchfork's reforking](https://github.com/Shopify/pitchfork/blob/master/docs/REFORKING.md). One issue with that is that when you refork and you are using the native Rails MemCacheStore, the underlying Dalli connection actually [closes and then raises an exception](https://github.com/petergoldstein/dalli/blob/1eb5efd0e162f33d3ef7349623ffc91c25c1ca3e/lib/dalli/protocol/connection_manager.rb#L221-L228) when a fork was detected. The recommended course of action is to call `.close` on the connection when forking (before making a call to the cache), but in recent versions of Rails, the underlying client is not accessible.

### Detail

This pull request introduces a new `client` method to the `ActiveSupport::Cache::MemCacheStore` class, providing direct access to the underlying Dalli client or connection pool. This change enhances flexibility for advanced operations not exposed by the standard cache store interface. Additionally, a corresponding test has been added to ensure proper functionality.

### Additional information

In some applications, we simply call `Rails.cache.instance_variable_get(:@data).close` in the `after_mold_fork` callback  and that seems to work as intended. This change would allow us and others to replace that with `Rails.cache.client.close` though which feels like less of a hack.

I had also considered just adding a close method to the cache store itself which just proxies that on to the `@data` client, but after some thought I saw a situation where that would lead to other people just adding lots of extra proxy methods to the `MemCacheStore`.

As far as naming, I just called it `client` because that's what it is, but I'm happy to change it to `connection` like some other libraries use for similar patterns. I could also call it `dalli` like it was called in Rails until 7+ for those who crave nostalgia but that introduces a loose coupling to the name of the client.

Lastly, there are a bunch of other tests that use `cache.instance_variable_get(:@data)` that I did not touch in order to keep this diff simple. If so desired, I can go through and update those to use the new `client` method as well.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add `client` method to `ActiveSupport::Cache::MemCacheStore` to provide access to the underlying Dalli client.
+
+    ```ruby
+    Rails.cache.client.close if Rails.cache.is_a?(ActiveSupport::Cache::MemCacheStore)
+    ```
+
+    *Patrick Tulskie*
+
 *   Allow to configure maximum cache key sizes
 
     When the key exceeds the configured limit (250 bytes by default), it will be truncated and
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -98,6 +98,24 @@ module ActiveSupport
         "#<#{self.class} options=#{options.inspect} mem_cache=#{instance.inspect}>"
       end
 
+      # Returns the underlying Dalli client instance.
+      #
+      # This provides direct access to the Dalli client or connection pool to perform operations
+      # not exposed by the cache store interface.
+      #
+      # For example, to close the connection before forking:
+      #
+      #   Rails.cache.client.close if Rails.cache.is_a?(ActiveSupport::Cache::MemCacheStore)
+
+      # To perform operations on a pooled connection:
+      #
+      #   Rails.cache.client.with do |c|
+      #     c.close
+      #   end
+      def client
+        @data
+      end
+
       ##
       # :method: write
       # :call-seq: write(name, value, options = nil)
--- a/activesupport/test/cache/stores/mem_cache_store_test.rb
+++ b/activesupport/test/cache/stores/mem_cache_store_test.rb
@@ -387,6 +387,18 @@ class MemCacheStoreTest < ActiveSupport::TestCase
     assert_equal 5, pool.instance_variable_get(:@timeout)
   end
 
+  def test_client_returns_direct_dalli_client_when_not_pooled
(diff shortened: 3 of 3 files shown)


PatrickTulskie:
> I think we could instead eagerly close the connections on after fork.
> 
> As in instead of exposing `#client`, we'd add `ActiveSupport::Cache::Store#close` or something, as a signal connections should be abandoned.

I figured this PR was minimally invasive and had some side benefits. Then again connections persisting after fork are really the bane of my existence and I'd rather fix them at the source if possible.

We had looked into why Dalli was raising this exception and found this issue here: https://github.com/petergoldstein/dalli/issues/997 so it seems like there isn't a lot of interest in actually changing this behavior. That said, I have some actual work I'm procrastinating on so maybe I'll put together a PR for Dalli today that addresses this. This has the additional benefit of not requiring a Rails update to remove our hack.

Failing that, I see your point about adding a `ActiveSupport::Cache::Store#close` (which I also considered) but from what I'm seeing, it looks like only Dalli has this issue since it seems you already addressed it (https://github.com/redis/redis-rb/commit/389d8b637713346e9052f4a5c2f1e0cb6a4e8add) in `redis-rb`. For all of the other store adapters, it would just be a NOOP.

I don't love adding automatic reconnection-after-fork handling to Rails since I don't think it's a concern of Rails and it adds overhead in non-forking environments, but it could be a good stop gap as well.

byroot:
> For all of the other store adapters, it would just be a NOOP.

Yes, but that's not a problem for me. Also there are gems that extend Store and may need it.

> I don't love adding automatic reconnection-after-fork handling to Rails since I don't think it's a concern of Rails

We do it for Active Record, so it's very much a concern of rails.

In any case, I much prefer exposing a generic `#close` method, which you can blindly call regardless of which implementation you use, than to expose a `#client` that is memcached specific, or return widely different things.

Especially since as shown in your PR, `#client` may be a Dalli instance or a ConnectionPool, so it's a very leaky API.

PatrickTulskie:
@byroot hey just a heads up, I put some work in getting this solved in Dalli directly https://github.com/petergoldstein/dalli/pull/1036 - seems like jruby is putting up a fight in CI, but I'm not sure if it's actually related to my change or not. In any case I gotta dig into that one a bit more. This would be ideal though because then I can just update Dalli on some Rails 7.x apps and get the fix without needing a major Rails upgrade.

In parallel I started work on just making a close method on ActiveSupport::Cache::Store which I hope to wrap up work on this week. Code is done for the simple stuff, just trying to figure out testing to ensure that I didn't whack the connection pool.

In any case, thanks for your help and the gut check here. I'll close this out for now and reference it in that other PR when I get it opened.

PR #54984, opened 2025-04-28
Implement ability to opt out of parallel database hooks

eileencodes:
Preivously, you could append your own hooks to parallel tests, but there was no way to skip the Rails database parallelization hooks. There may be cases where you want to implement your own database handling, so this allows you to implement all of that in a `parallelize_setup` block without having Rails create any databases for you.

To skip Rails database parallelization when running parallel tests you can set it in the `parallelize` caller:

```ruby
parallelize(workers: 10, parallelize_databases: false)
```

Or in your application config:

```ruby
config.active_support.parallelize_test_databases = false
```

This is set to true by default.

Note: if you do not create databases for the processes, it is possible for your test suite to deadlock. If you're seeing frequent deadlocks, you may need to create more databases for the processes to use or adjust your tests/code so it's less likely to cause database contention.

### DIFF
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -5,11 +5,15 @@ require "active_support/testing/parallelization"
 module ActiveRecord
   module TestDatabases # :nodoc:
     ActiveSupport::Testing::Parallelization.before_fork_hook do
-      ActiveRecord::Base.connection_handler.clear_all_connections!
+      if ActiveSupport.parallelize_test_databases
+        ActiveRecord::Base.connection_handler.clear_all_connections!
+      end
     end
 
     ActiveSupport::Testing::Parallelization.after_fork_hook do |i|
-      create_and_load_schema(i, env_name: ActiveRecord::ConnectionHandling::DEFAULT_ENV.call)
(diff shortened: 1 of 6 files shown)


activesupport/lib/active_support/testing/parallelize_executor.rb
```diff
@@ -35,6 +36,7 @@ def parallel_executor
         def build_parallel_executor
           case parallelize_with
           when :processes
+            ActiveSupport.parallelize_test_databases = parallelize_databases
```
etiennebarrie, line 39: It does not seem necessary for `ParallelizeExecutor` to know about this concern, couldn't the flag could be set directly from `parallelize`?

eileencodes: I didn't do that because the hooks aren't available or used in the threaded parallelizer so setting it in `parallelize` would set it for both, but only be used by the process. I guess it's kind of odd it defaults to true then, but better than having it have it change when set and then do nothing.

eileencodes: Changed my mind and moved it to `parallelize`.

PR #54993, opened 2025-04-30
[ci skip] Fix typos in the Multiple Databases with Active Record guide

duffuniverse:
### Detail

This Pull Request fixes several typos in the 'Multiple Databases with Active Record' guide.

### DIFF
--- a/guides/source/active_record_multiple_databases.md
+++ b/guides/source/active_record_multiple_databases.md
@@ -24,7 +24,7 @@ At this time the following features are supported:
 * Multiple writer databases and a replica for each
 * Automatic connection switching for the model you're working with
 * Automatic swapping between the writer and replica depending on the HTTP verb and recent writes
-* Rails tasks for creating, dropping, migrating, and interacting with multiple databases
+* Rails tasks for creating, dropping, migrating, and interacting with the multiple databases
 
 The following features are not (yet) supported:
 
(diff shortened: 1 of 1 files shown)


guides/source/active_record_multiple_databases.md
```diff
@@ -24,7 +24,7 @@ At this time the following features are supported:
 * Multiple writer databases and a replica for each
 * Automatic connection switching for the model you're working with
 * Automatic swapping between the writer and replica depending on the HTTP verb and recent writes
-* Rails tasks for creating, dropping, migrating, and interacting with the multiple databases
+* Rails tasks for creating, dropping, migrating, and interacting with multiple databases
```
eileencodes, line 27: This was correct before.

guides/source/active_record_multiple_databases.md
```diff
@@ -166,8 +166,8 @@ config.active_record.reading_role = :readonly
 ```
 
 It's important to connect to your database in a single model and then inherit from that model
-for the tables rather than connect multiple individual models to the same database. Database
-clients have a limit to the number of open connections there can be, and if you do this, it will
+for the tables rather than connecting multiple individual models to the same database. Database
+clients have a limit to the number of open connections that can be, and if you do this, it will
```
eileencodes, line 170: These changes also aren't correct.

eileencodes (changes requested):
Thanks for the PR. 2 of the changes you're proposing aren't correct, can you undo them?

eileencodes (approved):

PR #54994, opened 2025-04-30
Fix markdown formatting error at AR 7.2-stable changelog [ci skip]

brkn:
### Old version

This could lead to a partially configured connection being used, resulting in various  the most common being with the PostgreSQLAdapter raising `undefined method `key?' for nil` or `TypeError: wrong argument type nil (expected PG::TypeMap)`.

### Fixed version

This could lead to a partially configured connection being used, resulting in various  the most common being with the PostgreSQLAdapter raising `undefined method 'key?' for nil` or `TypeError: wrong argument type nil (expected PG::TypeMap)`.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -11,7 +11,7 @@
     Use of `Timeout.timeout` could result in improperly initialized database connection.
 
     This could lead to a partially configured connection being used, resulting in various exceptions,
-    the most common being with the PostgreSQLAdapter raising `undefined method `key?' for nil`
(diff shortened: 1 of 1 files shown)


eileencodes:
This needs to be fixed on the first version that has this entry (which is [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/activerecord/CHANGELOG.md)) and then backported to 7-2-stable. Can you open a new PR pointing at 8-0-stable?

brkn:
@eileencodes recreated at #54995

PR #54995, opened 2025-04-30
 Fix markdown formatting error at AR 8.0-stable changelog [ci skip]

brkn:
### Old version

This could lead to a partially configured connection being used, resulting in various  the most common being with the PostgreSQLAdapter raising `undefined method `key?' for nil` or `TypeError: wrong argument type nil (expected PG::TypeMap)`.

### Fixed version

This could lead to a partially configured connection being used, resulting in various  the most common being with the PostgreSQLAdapter raising `undefined method 'key?' for nil` or `TypeError: wrong argument type nil (expected PG::TypeMap)`.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -11,7 +11,7 @@
     Use of `Timeout.timeout` could result in improperly initialized database connection.
 
     This could lead to a partially configured connection being used, resulting in various exceptions,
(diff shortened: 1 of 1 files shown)


eileencodes:
Thanks, cherry-picked into 7-2-stable at 4053377d723ea957ae8f3f1e318501d47d2e8e97

PR #54996, opened 2025-04-30
Prevent (de)serializing values multiple times in MemCacheStore

nickamorim:
### Motivation / Background

This Pull Request has been created because it seems that we are currently (de)serializing values multiple times since Dalli's serialization layer is not disabled. By default, [Dalli uses Marshal](https://github.com/petergoldstein/dalli/blob/main/lib/dalli/client.rb#L40) unless the [`raw: true` request option](https://github.com/petergoldstein/dalli/blob/e5378846becf6382e6ca38b24ac89ff2ec88e8df/lib/dalli/protocol/value_serializer.rb#L30) is sent on every storage request.

Here is a simple reproduction:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org/"

  gem "activesupport"
  gem "dalli"
  gem "msgpack"
end

require "active_support"

C = ActiveSupport::Cache.lookup_store(:mem_cache_store, serializer: :message_pack)
C.write("foo", "bar")

puts "cache: #{C.inspect}"
```

The important part of the output is: `@value_serializer=#<Dalli::Protocol::ValueSerializer:0x0000000133373780 @serialization_options={serializer: Marshal}>`. This means ActiveSupport::Cache will use MessagePack to (de)serialize the payload and then Dalli will marshal it.

### Detail

Similar to how we disable Dalli's compression layer, the change I am proposing will _effectively_ disable Dalli's serialization by configuring it with a null/no-op serializer. An alternative approach would be to pass in the `raw: true` request option but I've found that it does not work well with some coder configurations when determining when to deserialize a payload https://github.com/petergoldstein/dalli/blob/e5378846becf6382e6ca38b24ac89ff2ec88e8df/lib/dalli/protocol/value_serializer.rb#L37.
 
### Additional information

A couple (3) of `MemCacheStoreTest` tests are failing because they relied on Dalli's marshal behaviour for non-string values ([example](https://github.com/rails/rails/blob/0367cbf265ff25546230d2f668aa887e09995509/activesupport/test/cache/stores/mem_cache_store_test.rb#L339-L345)). I'm not sure if this was explicitly intentional but it doesn't seem correct. 

This may be a breaking change since now payloads will be marshaled when attempting to deserialize them via ActiveSupport's `coder` / `serializer`. Perhaps when reading an entry from the cache, we will need to unmarshal it first or rescue deserialization errors and attempt to unmarshal it before returning an error? I'm curious to hear others' thoughts.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -43,6 +43,16 @@ module ActiveSupport
 
       ESCAPE_KEY_CHARS = /[\x00-\x20%\x7F-\xFF]/n
 
+      class NullSerializer
+        def self.dump(value)
+          value
+        end
+
+        def self.load(value)
+          value
+        end
+      end
+
       # Creates a new Dalli::Client instance with specified addresses and options.
       # If no addresses are provided, we give nil to Dalli::Client, so it uses its fallbacks:
       # - ENV["MEMCACHE_SERVERS"] (if defined)
@@ -89,7 +99,9 @@ module ActiveSupport
         @mem_cache_options = options.dup
         # The value "compress: false" prevents duplicate compression within Dalli.
         @mem_cache_options[:compress] = false
-        (OVERRIDDEN_OPTIONS - %i(compress)).each { |name| @mem_cache_options.delete(name) }
+        # The value "serializer: NullSerializer" prevents duplicate serialization within Dalli.
+        @mem_cache_options[:serializer] = NullSerializer
+        (OVERRIDDEN_OPTIONS - %i(compress) - %i(serializer)).each { |name| @mem_cache_options.delete(name) }
         @data = self.class.build_mem_cache(*(addresses + [@mem_cache_options]))
       end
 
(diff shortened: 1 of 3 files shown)


nickamorim:
@byroot I'm curious to get your thoughts on this PR. My main concern is that it isn't backwards compatible and, if shipped, it will result in some (de)serialization errors. I can work on addressing that if you think this is worth pursuing.

byroot:
I'm pretty sure we ran into this already in the past and is now the second time we're trying to fix it.

Unfortunately breaking all the caches is a bit of a problem as it makes upgrading tricky.

I think this would be better handled on the dalli side, given dalli has the opportunity to not serialize strings and use memcached tags to allow for a seemless upgrade.

Same thing we do in `memcached_store`: https://github.com/Shopify/memcached_store/blob/f8d2ef77e5b21d89e749649f1b4362363688951a/lib/active_support/cache/memcached_store.rb#L33-L36

But it should be done a bit better, if you skip string serialization you have to make sure you restore strings with the same encoding, so similar to `paquito`'s StringBypass: https://github.com/Shopify/paquito/blob/34b4cd31bc4a43120403726226f8d0a86809a5b9/lib/paquito/single_byte_prefix_version_with_string_bypass.rb#L15-L23

And to allow for a seemless upgrade, that new behavior should be opt-in, so that you can first upgrade dalli without using that bypass, and then in a followup you turn the option on.

byroot:
Closing in favor of: https://github.com/petergoldstein/dalli/pull/1041. We can reopen if somehow the dalli feature doesn't make it in or something.

PR #54998, opened 2025-05-01
Fix optimistic locking to always use original lock_version for checks

zakariaf:
### Motivation / Background

This Pull Request addresses a subtle bug in ActiveRecord's optimistic locking mechanism (`ActiveRecord::Locking::Optimistic`). The current implementation incorrectly determines the `lock_version` value used in the `WHERE` clause for `UPDATE` and `DELETE` operations when the `lock_version` attribute has been manually modified in memory before saving.

Specifically, if `will_save_change_to_attribute?(:lock_version)` is true (meaning the in-memory value differs from the original database value), the code uses the *new* in-memory value (`value_for_database`) for the `WHERE` clause check (e.g., `WHERE lock_version = <new_value>`).

This violates the fundamental principle of optimistic locking, which requires comparing against the version of the record *as it was when loaded or last saved* to detect concurrent modifications. This flaw can lead to stale writes succeeding silently when they should raise an `ActiveRecord::StaleObjectError`.

### Detail

This Pull Request modifies the `_lock_value_for_database` method within `ActiveRecord::Locking::Optimistic` to always return the original value loaded from the database (`original_value_for_database`). This ensures that the `WHERE` clause used for optimistic locking checks correctly compares against the version last seen in the database, regardless of any manual in-memory modifications to the attribute before the save attempt.

Additionally:
*   New test cases have been added to `test/cases/locking_test.rb` to specifically verify the fixed scenario for both update (`save!`) and `destroy!` operations where a manual `lock_version` assignment previously bypassed the stale check during a *concurrent* update.
*   The existing test `test_explicit_update_lock_column_raise_error` has been modified (and renamed to `test_update_succeeds_with_manually_assigned_lock_version_if_record_is_not_actually_stale`). This test was previously passing *because* of the bug (it expected a `StaleObjectError` which was raised accidentally due to the `WHERE` clause failing with the wrong value). The modified test now correctly asserts that a save should *succeed* when `lock_version` is manually assigned but no concurrent update has actually made the record stale.

### Additional information

This change addresses an unintended consequence related to the logic discussed in #41082, ensuring optimistic locking is robust even in this specific edge case. The core behavior for standard optimistic locking scenarios remains unchanged and continues to pass existing tests. The modification of the existing test confirms the correct behavior under the new logic.

### Checklist

* [X] Commit message has a detailed description of what changed and why.
* [X] Tests are added or updated if you fix a bug or add a feature. *(New tests added, one existing test modified)*
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. *(This is a bug fix for an edge case, likely not requiring a CHANGELOG entry)*

### DIFF
--- a/activerecord/lib/active_record/locking/optimistic.rb
+++ b/activerecord/lib/active_record/locking/optimistic.rb
@@ -139,11 +139,10 @@ module ActiveRecord
         end
 
         def _lock_value_for_database(locking_column)
-          if will_save_change_to_attribute?(locking_column)
-            @attributes[locking_column].value_for_database
-          else
-            @attributes[locking_column].original_value_for_database
-          end
+          # Always use the original value read from the database for the optimistic lock check.
+          # This ensures that even if the lock_version attribute is manually assigned a new
+          # value in memory, the WHERE clause correctly checks against the version last seen.
+          @attributes[locking_column].original_value_for_database
         end
 
         def _clear_locking_column
--- a/activerecord/test/cases/locking_test.rb
+++ b/activerecord/test/cases/locking_test.rb
@@ -238,17 +238,86 @@ class OptimisticLockingTest < ActiveRecord::TestCase
     end
   end
 
-  def test_explicit_update_lock_column_raise_error
+  test "update succeeds with manually assigned lock_version if record is not actually stale" do
     person = Person.find(1)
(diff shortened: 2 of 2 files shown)


matthewd:
Yeah, the current behaviour is intended and documented, per #41049

PR #55000, opened 2025-05-01
Rescue connection related errors in MemCacheStore#read_multi_entries

aledalgrande:
### Motivation / Background

This Pull Request has been created because Rails is not managing network blips when reading multiple values from memcached.

### Detail

This Pull Request changes `MemCacheStore` to swallow various connection exceptions when reading multiple values. The class already does this for every other read and write method, `read_multi_entries` was the only one not implementing the behavior.

### Additional information

For further validation, see how `RedisCacheStore` implements `failsafe` (equivalent to `rescue_error_with` in `MemCacheStore`) for all read and write methods, including `read_multi_entries`: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/cache/redis_cache_store.rb#L356

### Checklist

### DIFF
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -219,26 +219,24 @@ module ActiveSupport
         def read_multi_entries(names, **options)
           keys_to_names = names.index_by { |name| normalize_key(name, options) }
 
-          raw_values = begin
-            @data.with { |c| c.get_multi(keys_to_names.keys) }
-          rescue Dalli::UnmarshalError
-            {}
-          end
+          rescue_error_with({}) do
+            raw_values = @data.with { |c| c.get_multi(keys_to_names.keys) }
 
-          values = {}
+            values = {}
 
-          raw_values.each do |key, value|
-            entry = deserialize_entry(value, raw: options[:raw])
+            raw_values.each do |key, value|
+              entry = deserialize_entry(value, raw: options[:raw])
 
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/cache/mem_cache_store.rb
```diff
@@ -219,26 +219,28 @@ def write_serialized_entry(key, payload, **options)
...
 
-          raw_values = begin
-            @data.with { |c| c.get_multi(keys_to_names.keys) }
-          rescue Dalli::UnmarshalError
-            {}
-          end
+          rescue_error_with({}) do
+            raw_values = begin
+              @data.with { |c| c.get_multi(keys_to_names.keys) }
+            rescue Dalli::UnmarshalError
```
skipkayhil, line 225: Can the `UnmarshalError` rescue be removed since it should be caught by the outer block rescuing `Dalli::DalliError`?

skipkayhil (commented):
Test failures look real (ignore the Rubocop one though, that just looks broken)

activesupport/lib/active_support/cache/mem_cache_store.rb
aledalgrande: Thanks sorry for the failing tests, I'm having an issue with my local setup and thought they actually passed.

Yeah I can remove that, since we're now rescuing a superclass, good catch!

skipkayhil (approved):

byroot:
Good catch. Thank you!

byroot:
Backported to 8.0 and 7.2

PR #55002, opened 2025-05-02
Enforce strict loading violations for associations where joins are disabled

dewski:
### Motivation / Background

When a model has been marked for strict loading (via [query methods](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-strict_loading) or on the [model instance itself](https://api.rubyonrails.org/classes/ActiveRecord/Core.html#method-i-strict_loading-21)) the strict loading requirement is not enforced when joins are disabled on an association.

### Detail

This Pull Request changes where strict loading is enforced in all associations to ensure associations are eager loaded as expected. Currently, on the `main` branch strict loading is enforced in the `ActiveRecord::Associations::Association#find_target` method:

https://github.com/rails/rails/blob/e0bfbb2da5c7bd550d3576a38bcdd6fe00d56dfd/activerecord/lib/active_record/associations/association.rb#L248-L251

Strict loading is enforced as expected until `find_target` is overridden by a subclass. `find_target` is overridden in both `ActiveRecord::Associations::HasManyThroughAssociation` and `ActiveRecord::Associations::SingularAssociation` classes to return early when joins are disabled. `ActiveRecord::Associations::SingularAssociation` in particular is used for both `belongs_to` and `has_one` associations:

https://github.com/rails/rails/blob/e0bfbb2da5c7bd550d3576a38bcdd6fe00d56dfd/activerecord/lib/active_record/associations/has_many_through_association.rb#L225-L230

https://github.com/rails/rails/blob/e0bfbb2da5c7bd550d3576a38bcdd6fe00d56dfd/activerecord/lib/active_record/associations/singular_association.rb#L47-L57

### Additional information

It is desirable when a model is marked for strict loading that even associations with `disable_joins` set to `true` behave similar to when `disable_joins` is set to false.

Developers can still eager load those associations to avoid the strict loading violation. Without this change, strict loading could be bypassed causing additional queries to be performed unexpectedly.

For example:

```ruby
  has_many :posts
  has_many :comments_with_order, -> { ordered_by_post_id }, through: :posts, source: :comments
  has_many :no_joins_comments, through: :posts, disable_joins: true, source: :comments
end

author = Author.strict_loading.first
author.strict_loading!
author.comments_with_order.to_a # => raises ActiveRecord::StrictLoadingViolationError
# Executes:
# SELECT "posts"."id" FROM "posts" WHERE "posts"."author_id" = $1
# SELECT "comments".* FROM "comments" WHERE "comments"."post_id" IN ($1, $2, $3, $4, $5)
author.no_joins_comments.to_a
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   `strict_loading` is enforced on associations configured with joins disabled.
+
+    *Garrett Bjerkhoel*
+
 *   `:class_name` is now invalid in polymorphic `belongs_to` associations.
 
     Reason is `:class_name` does not make sense in those associations because
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -187,7 +187,13 @@ module ActiveRecord
       # ActiveRecord::RecordNotFound is rescued within the method, and it is
       # not reraised. The proxy is \reset and +nil+ is the return value.
       def load_target
-        @target = find_target(async: false) if (@stale_state && stale_target?) || find_target?
+        if (@stale_state && stale_target?) || find_target?
+          if violates_strict_loading?
+            Base.strict_loading_violation!(owner: owner.class, reflection: reflection)
+          end
+
+          @target = find_target(async: false)
+        end
 
         loaded! unless loaded?
         target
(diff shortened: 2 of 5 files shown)


adityapandit17:
💯 We might need to perform a rebase to see if that resolves the CI issues.

Also, is it possible to test the behavior when joins are disabled, and we attempt to load associated data from a different database cluster—similar to the setup we have at GitHub?

adityapandit17 (commented):
💯 We might need to perform a rebase to see if that resolves the CI issues.

Also, is it possible to test the behavior when joins are disabled, and we attempt to load associated data from a different database cluster—similar to the setup we have at GitHub?

dewski:
> Also, is it possible to test the behavior when joins are disabled, and we attempt to load associated data from a different database cluster—similar to the setup we have at GitHub?

The strict loading enforcement is performed before any association is loaded in `ActiveRecord::Associations::Association#load_target` which should be applied independent of the connection. Is there a specific example you are thinking of?

PR #55003, opened 2025-05-02
`ActiveSupport::ErrorReporter#report` assigns a cause to unraised exceptions

amomchilov:
### Motivation / Background

`#report` has an undocumented side-effect of setting the `cause` on a never-raised exception.

### Detail

This Pull Request adds a test to document this behaviour.

The fact that this sets the cause is desirable, because it makes it possible to report a newly-constructed error. This can be useful for rewrapping a genetic error into a more domain-specific error with a more contextual message.

```ruby
begin
  do_something
rescue ActiveRecord::RecordNotFound
  Rails.report.unexpected(SomeMoreSpecificException.new("A domain-specific explanation of what went wrong"))
  return
end
```

If `#report` didn't set the cause, then the link to the `ActiveRecord::RecordNotFound` would be lost. And there's no way for the user to set it themselves, without doing the same `raise`+`rescue` trick that we use. It's good that the framework does it, so they don't have to.

### Additional information

See also: #52684

### Checklist

### DIFF
--- a/activesupport/lib/active_support/error_reporter.rb
+++ b/activesupport/lib/active_support/error_reporter.rb
@@ -276,14 +276,12 @@ module ActiveSupport
 
     private
       def ensure_backtrace(error)
-        return if error.frozen? # re-raising won't add a backtrace
+        return if error.frozen? # re-raising won't add a backtrace or set the cause
         return unless error.backtrace.nil?
 
         begin
-          # We could use Exception#set_backtrace, but until Ruby 3.4
-          # it only support setting `Exception#backtrace` and not
-          # `Exception#backtrace_locations`. So raising the exception
-          # is a good way to build a real backtrace.
+          # As of Ruby 3.4, we could use Exception#set_backtrace to set the backtrace,
+          # but there's nothing like Exception#set_cause. Raising+rescuing is the only way to set the cause.
           raise error
         rescue error.class => error
         end
--- a/activesupport/test/error_reporter_test.rb
+++ b/activesupport/test/error_reporter_test.rb
@@ -172,6 +172,17 @@ class ErrorReporterTest < ActiveSupport::TestCase
     assert_equal __FILE__, error.backtrace_locations.first.path
   end
 
+  test "#report assigns a cause if it's missing" do
+    raise "the original cause"
+  rescue => cause
+    new_error = StandardError.new("A new error that should wrap the StandardError")
+    assert_nil new_error.cause
+
+    @reporter.report(new_error)
+
+    assert_same cause, new_error.cause
+  end
+
   test "#record passes through the return value" do
     result = @reporter.record do
       2 + 2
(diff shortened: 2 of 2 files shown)


activesupport/lib/active_support/error_reporter.rb
```diff
@@ -276,7 +276,7 @@ def report(error, handled: true, severity: handled ? :warning : :error, context:
 
     private
       def ensure_backtrace(error)
-        return if error.frozen? # re-raising won't add a backtrace
+        return if error.frozen? # re-raising won't add a backtrace or set the cause
```
amomchilov, line 279: Should I update the comment on lines 283-286?

There's no way to set the cause besides the `raise`+`rescue` trick, so Ruby 3.4's `set_backtrace` trick won't help us here.

activesupport/test/error_reporter_test.rb
```diff
@@ -166,9 +166,21 @@ class ErrorReporterTest < ActiveSupport::TestCase
     assert_nil error.backtrace
     assert_nil error.backtrace_locations
 
-    assert_nil @reporter.report(error)
-    assert_not_predicate error.backtrace, :empty?
-    assert_not_predicate error.backtrace_locations, :empty?
```
amomchilov, line 171: These tests were too lax, and don't cover the backtrace shifting logic.

It's wrong when running in Spring, but that's a Spring bug (https://github.com/rails/spring/issues/739)

amomchilov (commented):
cc @byroot

activesupport/lib/active_support/error_reporter.rb
byroot: Yes, you can mostly remove it I think.

activesupport/lib/active_support/error_reporter.rb
```diff
@@ -276,14 +276,12 @@ def report(error, handled: true, severity: handled ? :warning : :error, context:
...
 
         begin
-          # We could use Exception#set_backtrace, but until Ruby 3.4
-          # it only support setting `Exception#backtrace` and not
-          # `Exception#backtrace_locations`. So raising the exception
-          # is a good way to build a real backtrace.
+          # As of Ruby 3.4, we could use Exception#set_backtrace to set the backtrace,
+          # but there's nothing like Exception#set_cause. Raising+rescuing is the only way to set the cause.
           raise error
         rescue error.class => error
```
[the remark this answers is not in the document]
byroot: In case `error.exception` is implemented and craps out.

https://docs.ruby-lang.org/en/master/Exception.html#method-c-exception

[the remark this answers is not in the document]
amomchilov: AH yeah, somebody could call e.g. `Rails.error.report(StandardError)` instead of `Rails.error.report(StandardError.new)`, and `exception` would be called in the latter case.

Cursed.

PR #51426, opened 2024-03-26
Automatically delay Active Job enqueues to after commit

casperisfine:
Ref: https://github.com/rails/rails/pull/26103
Ref: https://github.com/rails/rails/pull/51474

A fairly common mistake with Rails is to enqueue a job from inside a transaction, with a record as argumemnt, which then lead to a `RecordNotFound` error when picked up by the queue.

This is even one of the arguments advanced for job runners backed by the database such as `solid_queue`, `delayed_job` or `good_job`. 

But relying on this is undesirable in my opinion as it makes the Active Job abstraction leaky, and if in the future you need to migrate to another backend or even just move the queue to a separate database, you may experience a lot of race conditions of the sort.

To resolve this problem globally, we can make Active Job optionally transaction aware, and automatically defer job queueing to `after_commit`.

PR #51525, opened 2024-04-09
Fix `ActiveJob::EnqueueAfterTransactionCommit` API

casperisfine:
Fix: https://github.com/rails/rails/pull/51426#issuecomment-2042611790

`perform_later` is supposed to return the Job instance on success, and `false` on error.

When the `enqueue` is automatically delayed, it's of course impossible to predict if the actual queueing will succeed, but for backward compatibility reasons, it's best to assume it will.

If necessary, you can hold onto the job instance and check for `#successfully_enqueued?` after the transaction has completed.

cc @matthewd @bensheldon

PR #55004, opened 2025-05-02
Defer ActiveJob enqueue callbacks until after commit when `enqueue_after_transaction_commit` enabled

wroever:
### Motivation / Background

At work we use an `around_enqueue` callback to handle lapses in connectivity to Sidekiq by temporarily storing job payloads we’re unable to deliver to Sidekiq in our database (so that we can manually enqueue them later, if needed).

The code looks roughly like this:
```ruby
class ApplicationJob < ActiveJob::Base

  around_enqueue do |job, block|
    block.call
  rescue RedisClient::Error => e
    FailedEnqueueAttempt.create!(
      job: job,
      error_message: e.detailed_message
    )
    # Log/report error, etc.
  end

  # ...
end
```

I recently noticed this wasn’t working as I expected with the `enqueue_after_transaction_commit` setting enabled. Looking at the [ActiveJob code](https://github.com/rails/rails/blob/v7.2.2.1/activejob/lib/active_job/enqueuing.rb#L117), I noticed that when `enqueue_after_transaction_commit` is enabled, `enqueue` callbacks fire at the time `perform_later` is called, as opposed to the time at which the job is later enqueued via the queue adapter.

While I see that this has always been the behavior since `enqueue_after_transaction_commit` was first added (#51426), I think this is likely not the desired behavior when the setting is enabled. My understanding is that `enqueue` callbacks are designed to instrument the enqueueing interaction with the queue adapter. Looking around at other usage examples ([example](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/instrumentation/active_job.rb#L18), [example](https://github.com/NREL/OpenStudio-server/blob/v3.9.0-rc3/server/app/jobs/resque_jobs/run_simulate_data_point.rb#L11)), it seems likely to me that the current behavior might catch other folks by surprise too.

This is my attempt at a Rails contribution so please do guide me if I've missed something here - thanks! 🙏

### Checklist

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Defer invocation of ActiveJob enqueue callbacks until after commit when
+    `enqueue_after_transaction_commit` is enabled.
+
+    *Will Roever*
+
 *   Add `report:` option to `ActiveJob::Base#retry_on` and `#discard_on`
 
     When the `report:` option is passed, errors will be reported to the error reporter
--- a/activejob/lib/active_job/enqueuing.rb
+++ b/activejob/lib/active_job/enqueuing.rb
@@ -113,9 +113,7 @@ module ActiveJob
       set(options)
       self.successfully_enqueued = false
 
-      run_callbacks :enqueue do
-        raw_enqueue
-      end
+      raw_enqueue
 
       if successfully_enqueued?
         self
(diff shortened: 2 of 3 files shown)


bensheldon:
Conceptually this looks correct to me 👍

Historically, enqueue_after_transaction_commit was implemented using around_enqueue and when it was refactored to instead patch raw_enqueue the callback wasn't considered (I say this for myself as I remember reviewing it): https://github.com/rails/rails/pull/51525

PR #55007, opened 2025-05-03
Make the executor hooks in AR::QueryCache private

fxn:
API docs [show](https://api.rubyonrails.org/v8.0/classes/ActiveRecord/QueryCache.html#method-c-run) the executor hooks in `ActiveRecord::QueryCache` and their installer. These should be private, they are not supposed to be callable or overridden by external code.

The patch also groups the hooks to ease making sense of this class, [just like the connection pool does](https://github.com/rails/rails/blob/e0bfbb2da5c7bd550d3576a38bcdd6fe00d56dfd/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L218).

This grouping is also coherent with `ActiveRecord::QueryCache::ClassMethods`, though, if you ask me, that extends `ActiveRecord::Base` and I personally would _not_ group those methods.

Technically, this is a breaking change, but this stuff seems too internal to me. However, happy to reconsider if someone sees it differently.

There are also some documentation tweaks since I was on it.

### DIFF
--- a/activerecord/lib/active_record/query_cache.rb
+++ b/activerecord/lib/active_record/query_cache.rb
@@ -3,6 +3,7 @@
 module ActiveRecord
   # = Active Record Query Cache
   class QueryCache
+    # ActiveRecord::Base extends this module, so these methods are available in models.
     module ClassMethods
       # Enable the query cache within the block if Active Record is configured.
       # If it's not, it will execute the given block.
@@ -20,11 +21,15 @@ module ActiveRecord
         end
       end
 
-      # Disable the query cache within the block if Active Record is configured.
-      # If it's not, it will execute the given block.
+      # Runs the block with the query cache disabled.
       #
-      # Set <tt>dirties: false</tt> to prevent query caches on all connections from being cleared by write operations.
-      # (By default, write operations dirty all connections' query caches in case they are replicas whose cache would now be outdated.)
+      # If the query cache was enabled before the block was executed, it is
+      # enabled again after it.
+      #
+      # Set <tt>dirties: false</tt> to prevent query caches on all connections
+      # from being cleared by write operations. (By default, write operations
+      # dirty all connections' query caches in case they are replicas whose
+      # cache would now be outdated.)
       def uncached(dirties: true, &block)
         if connected? || !configurations.empty?
           connection_pool.disable_query_cache(dirties: dirties, &block)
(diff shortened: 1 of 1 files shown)


activerecord/lib/active_record/query_cache.rb
```diff
@@ -3,6 +3,7 @@
 module ActiveRecord
   # = Active Record Query Cache
   class QueryCache
+    # ActiveRecord::Base extends this module, so these methods are available in models.
```
zzak, line 6: I wish there was a way to see what classes are extended in the API docs, so we don't have to call this out each time. And for a way to see methods on the receiving class as if they were defined directly.

fxn: @zzak 100%, a should be able to discover the API of `ActiveRecord::Base` just by browsing.

Furthermore, the module should not be listed in the docs. The way the code is organized does not belong to the public contract. The method could be defined directly, could be injected via a module, ..., that aspect does not belong to the public interface and we should be able to refactor these modules at will.

fxn: @zzak an idea would be to have a new directive, something like

```ruby
class C
  include M # :doc:inject
end
```

which would also nodoc `M`, perhaps.

For simple use cases, that would work.

It would not suffice for the use case in which the implementation is spread across multiple ancestors, like what happens with `ActiveRecrod::Base#save`.

zzak: > is spread across multiple ancestors

Yeah, and also if we show every included method the base objects API docs could get unruly.

One hacky way to do it is nodoc the module then do an explicit `:doc: whatever-method` on the target class, but you lose on the source reference (toggle to "show source").

It's that or we make it easier to to find the extending modules, but sounds like you're rather keep those private?

For example, we already show a list of the included (inherited) modules, it should be similar to get a list of extended modules.

Otherwise yeah, we have to introduce a new directive to RDoc.

fxn: > but sounds like you're rather keep those private?

If `A::B::C` is a module that is published, renaming it, deleting it, etc. is a breaking change. And that is unnecessary, because in the use cases I have in mind (like this one), ancestor chains are 100% internal. The only thing the user has to be told is that the method exists in the containing, user-facing object.

fxn: Let me show a concrete example in Zeitwerk.

In order to configure your root directories, you invoke `push_dir`:

```ruby
loader.push_dir(...)
```

Where is `push_dir` implemented? That is not a contract we have to stablish, we only need to define that loaders respond to that method, and it is my job to respect that contract through releases. (It is indeed defined in an internal mixin right now.)

Zeitwerk is not documented via RDoc, it is all in the README, where I have full control. But you see what I mean.

PR #55008, opened 2025-05-04
Add support for Cache-Control: only-if-cached directive

egg528:
### Motivation / Background

Add support for the `Cache-Control: only-if-cached` directive as specified in RFC 9111. This allows Rails applications to properly parse and detect this directive in incoming requests.

### Detail

This PR adds a `request.only_if_cached?` method to detect when a request includes the `only-if-cached` directive, with appropriate tests and documentation.

### Additional information

RFC reference: https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.7

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/http/cache.rb
+++ b/actionpack/lib/action_dispatch/http/cache.rb
@@ -8,6 +8,9 @@ module ActionDispatch
       module Request
         HTTP_IF_MODIFIED_SINCE = "HTTP_IF_MODIFIED_SINCE"
         HTTP_IF_NONE_MATCH     = "HTTP_IF_NONE_MATCH"
+        HTTP_CACHE_CONTROL    = "HTTP_CACHE_CONTROL"
+
+        ONLY_IF_CACHED = "only-if-cached"
 
         mattr_accessor :strict_freshness, default: false
 
@@ -34,6 +37,14 @@ module ActionDispatch
             validators = if_none_match_etags
             validators.include?(etag) || validators.include?("*")
           end
+        end
+
+        # Returns true when the request includes the `Cache-Control: only-if-cached` directive.
+        # If no suitable cached response exists, recommends responding with `504 Gateway Timeout`.
+        # Reference: https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.7
+        def only_if_cached?
+          cache_control = get_header(HTTP_CACHE_CONTROL)
+          cache_control && cache_control.split(",").any? { |d| d.strip == ONLY_IF_CACHED }
         end
 
         # Check response freshness (`Last-Modified` and `ETag`) against request
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -140,4 +140,26 @@
 
     *Petrik de Heus*
 
+*   Add support for the `Cache-Control: only-if-cached` directive according to RFC 9111.
+
+    Provides a `request.only_if_cached?` method to detect when a request includes this directive.
+    According to RFC 9111, a 504 Gateway Timeout should be returned if no suitable cached response exists.
+
+    ```ruby
+    def show
+      if request.only_if_cached?
+        @article = Rails.cache.read("article/#{params[:id]}")
+        return head(:gateway_timeout) if @article.nil?
+
+        render :show
+        return
+      end
+
+      @article = Article.find(params[:id])
+      render :show
+    end
+    ```
+
+    *egg528*
+
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/actionpack/CHANGELOG.md) for previous changes.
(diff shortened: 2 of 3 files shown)


actionpack/lib/action_dispatch/http/cache.rb
```diff
@@ -8,6 +8,9 @@ module Cache
       module Request
         HTTP_IF_MODIFIED_SINCE = "HTTP_IF_MODIFIED_SINCE"
         HTTP_IF_NONE_MATCH     = "HTTP_IF_NONE_MATCH"
+        HTTP_CACHE_CONTROL    = "HTTP_CACHE_CONTROL"
```
byroot, line 11: nitpick: Using constants like this is a thing that predates frozen string literals, it isn't really preferable anymore today.

actionpack/lib/action_dispatch/http/cache.rb
```diff
@@ -36,6 +39,14 @@ def etag_matches?(etag)
           end
         end
 
+        # Returns true when the request includes the `Cache-Control: only-if-cached` directive.
+        # If no suitable cached response exists, recommends responding with `504 Gateway Timeout`.
+        # Reference: https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.7
+        def only_if_cached?
+          cache_control = get_header(HTTP_CACHE_CONTROL)
+          cache_control && cache_control.split(",").any? { |d| d.strip == ONLY_IF_CACHED }
+        end
```
byroot, line 48: I wonder if we should instead expose a `#cache_control` method with a bunch of methods on it:

```ruby
request.cache_control.only_if_cached?
request.cache_control.no_store?
request.cache_control.max_age
```

egg528: Thanks for the suggestion! I completely agree this is a better approach.  

However, #cache_control is already dynamically defined in ActionDispatch::Request as a method that returns the cache-control header value as a string :sad: For backward compatibility reasons, we can't use this method name.   

Instead, what do you think about implementing it this way?

```ruby
module ActionDispatch
  module Http
    module Cache
      module Request
        ...

        # Returns a CacheControl object that provides access to request cache control directives.
        def cache_control_directives
          @cache_control_directives ||= CacheControlDirectives.new(get_header("HTTP_CACHE_CONTROL"))
        end
        
        # Represents the HTTP Cache-Control header for requests,
        # providing methods to access various cache control directives
        # Reference: https://www.rfc-editor.org/rfc/rfc9111.html#name-request-directives
        class CacheControlDirectives
          # Value directives: directives that take a numeric value
          VALUE_DIRECTIVES = %w[max-age max-stale min-fresh stale-if-error]
          
          # Boolean directives: directives that don't take a value
          BOOLEAN_DIRECTIVES = %w[no-cache no-store no-transform only-if-cached]
          
          def initialize(header_value)
            @directives = parse_directives(header_value)
          end
          
          # Returns true if the only-if-cached directive is present.
          # This directive indicates that the client only wishes to obtain a
          # stored response. If a valid stored response is not available,
          # the server should respond with a 504 (Gateway Timeout) status.
          def only_if_cached?
            @directives.key?("only-if-cached")
          end
          
          # Returns true if the no-cache directive is present.
          # This directive indicates that a cache must not use the response
          # to satisfy subsequent requests without successful validation on the origin server.
          def no_cache?
            @directives.key?("no-cache")
          end
          
          # Returns true if the no-store directive is present.
          # This directive indicates that a cache must not store any part of the
          # request or response.
          def no_store?
            @directives.key?("no-store")
          end
          
          # Returns true if the no-transform directive is present.
          # This directive indicates that a cache or proxy must not transform the payload.
          def no_transform?
            @directives.key?("no-transform")
          end
          
          # Returns the value of the max-age directive.
          # This directive indicates that the client is willing to accept a response
          # whose age is no greater than the specified number of seconds.
          def max_age
            @directives["max-age"]
          end
          
          # Returns the value of the max-stale directive.
          # This directive indicates that the client is willing to accept a response
          # that has exceeded its freshness lifetime by no more than the specified number of seconds.
          def max_stale
            @directives["max-stale"]
          end
          
          # Returns the value of the min-fresh directive.
          # This directive indicates that the client is willing to accept a response
          # whose freshness lifetime is no less than its current age plus the specified time in seconds.
          def min_fresh
            @directives["min-fresh"]
          end
          
          # Returns the value of the stale-if-error directive.
          # This directive indicates that the client is willing to accept a stale response
          # if the check for a fresh one fails with an error for the specified number of seconds.
          def stale_if_error
            @directives["stale-if-error"]
          end
          
          private
          
          def parse_directives(header_value)
            return {} unless header_value
          
            header_value.delete(" ").downcase.split(',').each_with_object({}) do |directive, result|
              name, value = directive.split('=', 2)
              
              if value && VALUE_DIRECTIVES.include?(name)
                result[name] = value.to_i
              elsif BOOLEAN_DIRECTIVES.include?(name)
                result[name] = true
              end
            end
          end
        end
      end
    end
  end
end

```

egg528:
@byroot 

I've submitted a new [PR](https://github.com/rails/rails/pull/55033) that incorporates all the suggestions. I'll close this PR now! Thanks for your helpful feedback and guidance throughout the process.

egg528:
@byroot

I've submitted a new https://github.com/rails/rails/pull/55033 that incorporates all the suggestions. I'll close this PR now! Thanks for your helpful feedback and guidance throughout the process.

PR #55009, opened 2025-05-04
Add UniquenessAmongValidator to validate attribute uniqueness within a model

zakariaf:
### Motivation / Background
This Pull Request has been created because there is currently no built-in way to validate that multiple attributes within the same model have unique values. While ActiveRecord provides `validates_uniqueness_of` for database-level uniqueness, there is no equivalent for validating uniqueness among attributes in the same model instance.

This validator is useful for various scenarios including:
- Ensuring that a user's primary email differs from their secondary email
- Validating that contact information fields (e.g., home phone, work phone, mobile) contain distinct values
- Checking that tag fields or categorization attributes don't contain duplicates
- Verifying that shipping and billing addresses are different when required
- Ensuring internationalized content fields don't duplicate content across attributes

### Detail
This Pull Request adds:
1. A new `UniquenessAmongValidator` class to validate that an attribute's value is distinct from other specified attributes
2. A `validates_uniqueness_of_among` helper method to easily apply this validation
3. Support for standard validation options (allow_nil, allow_blank, case_sensitive, etc.)
4. Comprehensive handling of different data types:
   - String comparison with optional case insensitivity
   - Array comparison regardless of element order
   - Hash comparison regardless of key order
   - Support for nested data structures
5. An optional `compare_hash_keys_as_strings` setting to treat string and symbol hash keys as equivalent
6. Comprehensive tests to verify functionality with various data types and edge cases

Example usage:

```ruby
  # Ensure primary and secondary emails are different (case insensitive)
  validates_uniqueness_of_among :primary_email, :secondary_email, case_sensitive: false
  
  # Ensure phone numbers are unique
  validates_uniqueness_of_among :phone_numbers, :emergency_contact_phone
  
  # Ensure translated profile fields don't duplicate content across attributes
  validates_uniqueness_of_among :bio, :tagline, :headline, compare_hash_keys_as_strings: true
end
```

### Additional information
The implementation supports various data types including strings, arrays, hashes, and primitive values. For string comparisons, it respects the `case_sensitive` option similar to other validators. For array and hash comparisons, it checks if they contain the same elements regardless of order.

For hashes, the validator can compare keys as strings with the `compare_hash_keys_as_strings` option, which is useful for internationalized content where keys might be strings in one attribute and symbols in another.

### Checklist
- [x] Tests are added for the new validator, covering various cases and edge scenarios.
- [x] CHANGELOG updated to reflect the new functionality.

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Add `UniquenessAmongValidator` to validate attribute uniqueness within a model.
+
+    ```ruby
+    class User < ApplicationRecord
+      # Validates that primary and secondary emails are different
+      validates_uniqueness_of_among :primary_email, :secondary_email
+
+      # Validates that all phone numbers are unique
+      validates_uniqueness_of_among :home_phone, :work_phone, :mobile_phone,
+                                    message: "must be different from other phone numbers"
+
+      # Validates that all tag sets are unique
+      validates_uniqueness_of_among :primary_tags, :secondary_tags
+
+      # Validates that metadata fields are unique, treating string and symbol keys as equivalent
+      validates_uniqueness_of_among :primary_metadata, :secondary_metadata,
+                                    compare_hash_keys_as_strings: true
+    end
+    ```
+
+    *Zakaria Fatahi*
+
 *   Add `except_on:` option for validation callbacks.
 
     *Ben Sheldon*
--- a/activemodel/lib/active_model/locale/en.yml
+++ b/activemodel/lib/active_model/locale/en.yml
@@ -36,3 +36,4 @@ en:
       in: "must be in %{count}"
       odd: "must be odd"
       even: "must be even"
+      taken_among: "has already been taken among %{attributes}"
--- a/activemodel/lib/active_model/validations/uniqueness_among.rb
+++ b/activemodel/lib/active_model/validations/uniqueness_among.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+module ActiveModel
+  module Validations
+    class UniquenessAmongValidator < EachValidator # :nodoc:
+      DEFAULTS = { case_sensitive: true }.freeze
+
+      def initialize(options)
+        super(DEFAULTS.merge(options))
+      end
+
+      def validate_each(record, attribute, value)
+        return if value.blank? && options[:allow_blank]
+        return if value.nil? && options[:allow_nil]
+
+        other_attributes = Array(options[:in] || options[:within] || attributes) - [attribute]
+        return if other_attributes.empty?
+
+        duplicates = []
+        other_attributes.each do |attr|
+          other_value = record.public_send(attr)
+          next if other_value.nil? && options[:allow_nil]
+          next if other_value.blank? && options[:allow_blank]
+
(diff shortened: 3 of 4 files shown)


fatkodima:
I am currently thinking that this is rarely used and not worth the complexity and better for people to implement it by custom validations in their applications. 
Can you try to find examples amongst the OSS rails apps (https://github.com/asyraffff/Open-Source-Ruby-and-Rails-Apps) where they will benefit from this validator?

zakariaf:
@fatkodima This was something that was used in our company's projects a few times. but in these open source apps, finding direct existing implementations is challenging in other apps, but I did find a strong example and numerous clear potential use cases that highlight the need for this type of validation

- OneBody::Family: This model has `barcode_id` and `alternate_barcode_id`. It currently uses a custom validation method (validates_each) to ensure these two fields are not identical within the same record (among other checks):
https://github.com/seven1m/onebody/blob/0c109cc676853c8e95397dd59393d59ed7dd4976/app/models/family.rb#L33

- Mastodon::AccountMigration: https://github.com/mastodon/mastodon/blob/89cafb01b47bee18db62da36a372a49d93b4a909/app/models/account_migration.rb#L72

And this strong potential use cases. it's like a pattern, attribute, alternate_attribute:

- Code.org::User: Has `email`, `parent_email`
- Code.org::Pd::Application::TeacherApplication: Has `email`, `alternate_email`

- Mastodon::User: Has `email`, `unconfirmed_email`

- TimeOverflow::User: Has `phone` and `alt_phone`
- Monocle::User: Has `twitter` and `github` or any other places that we want to get the social links

- Discourse::EmailChangeRequest: old_email, new_email, old_email_token_id, new_email_token_id
- Discourse::Topic: featured_user1_id, featured_user2_id, featured_user3_id, featured_user4_id
- Discourse::MovedPost: old_topic_id, new_topic_id, and old_post_id, new_post_id

- OneBody::Site: host, secondary_host
- OneBody::Person: email,  alternate_email

fatkodima:
Thanks for examples. I see most of the examples have 2 attributes in the validation and I think this is the most useful and popular pattern. And they compare simple literals, not composites like arrays hashes. The implementation can be simplified and the validation's name changed to something like `validates_different` or something like this (but I don't know how to name it better to be clear in both types like `validates_something` vs `validates ..., something: true`).

fatkodima:
Em, just recall: isn't this is already implemented by `validates :email, comparison: { other_than: :old_email }` (https://guides.rubyonrails.org/active_record_validations.html#comparison) ?

zakariaf:
oh dear, you’re absolutely right—they’re totally the same. ComparisonValidator is missing some capabilities (array, hash, case sensitivity, etc.), but yeah, they do the same. I was fully focused on ActiveModel in-depth last month to understand and rewrite it, find missing parts, and figure out how to improve, and I totally forgot about the existing things in the our parts. I’ll close this PR and extend ComparisonValidator to support these new capabilities.

fatkodima:
You spend a good amount of time on this and other ones, so I recommend discussing decent changes before opening PRs (to potentially not waste time) in the forum https://discuss.rubyonrails.org. I am not sure the proposed changes will still be needed, so recommend discussing first for this one for sure.

PR #51421, opened 2024-03-25
Add Active Model `Collection` attribute type

rafaelfranca:
### Summary

One case that is not yet covered by the current standard types is for attributes that represent a collection of multiple values. For example, one might need to represent a subset of strings as a list in a single attribute; or even an entire group of associated models that are all embedded in a named collection. 

This PR aims to implement a new Collection type.
It subclasses the base Value class and overrides the necessary methods so casting and serialization work for an array of objects that are themselves valid for one of the types in the Registry. This ensures that each element of the collection can be cast and serialized individually by delegating the operation to the appropriate type instance.

Collection serializes itself as a JSON array on the database. For casting, it attempts to convert values to a Ruby array.

Methods:

- `type`: returns the :collection symbol.
- `serializable?`: checks if the given value is an array containing only elements in which are all serializable for the collection type. It will instantiate the type passed during its own initialization and call serializable? for each element.
- `serialize`: returns a JSON string of an array with all elements serialized by the collection type.
- `deserialize`: expects a JSON array string and converts it into a Ruby array with each element deserialized by the collection type.
- `cast`: converts the given array into a normalized collection of cast values. It will map the given array to a new one in which each element is cast into a new value by the collection type.
- `changed_in_place?`: Accepts raw serialized value, deserializes it and checks if any of the elements of the collections were mutated according to the collection type method of the same name.
- `assert_valid_value`: checks if all elements of the given array are accepted by the collection type method of the same name.

### Other Information

This change also introduces additional `Value#valid_value?` method that is supposed to separate the assertion, which raises an exception from a boolean check that can be reused in other places. This way new types won't have to override the whole `assert_valid_value` method if they want to keep the default error message.

Let me know if this change should be implemented in a separate PR. I decided to keep it in this PR for better readability of the change but definitely willing to extract it and have a discussion is a different PR. Thanks!

### Documentation

Currently only the class itself is covered with documentation. But we have plans to cover public methods as well.
Either in this PR, in a separate one or perhaps as part of the https://github.com/rails/rails/pull/44306

Original PR #44324

PR #55012, opened 2025-05-04
Add ActiveModel::Type::Array for array attribute support

zakariaf:
### Motivation / Background

This Pull Request has been created because ActiveModel currently lacks a built-in type for handling array-like data. While ActiveRecord has array support for PostgreSQL, ActiveModel users have no standard way to define array attributes with proper type casting. This addition brings a consistent way to handle array data in any model that includes ActiveModel::Attributes.

### Detail

This Pull Request adds a new type class, `ActiveModel::Type::Array`, that allows defining array attributes with optional element type casting. The implementation supports:

- Creating array attributes with `attribute :tags, :array`
- Element type casting via the `:of` option: `attribute :counts, :array, of: :integer`
- Passing type options to the element type: `attribute :ratings, :array, of: :decimal, precision: 2`
- JSON string assignment that gets automatically parsed into arrays
- Proper mutation detection for dirty tracking
- Full compatibility with the existing type system

This makes it easier to work with array data in any model using ActiveModel::Attributes, providing a clean interface for collection data that would otherwise require manual serialization or custom types.

### Additional information

The implementation follows the pattern of existing ActiveModel types and integrates with the type system architecture. It supports all registered types as element types, making it very flexible and composable.

Unit tests have been added to verify the basic functionality and dirty tracking capabilities, but additional integration tests with real-world scenarios would be beneficial in a future iteration.

### Checklist

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,27 @@
+*   Add `ActiveModel::Type::Array` for handling array attributes.
+
+    This new type allows model attributes to store arrays with optional element type casting.
+
+    ```ruby
+    class Person
+      include ActiveModel::Attributes
+
+      attribute :tags, :array
+      attribute :counts, :array, of: :integer
+      attribute :ratings, :array, of: :decimal, precision: 2
+    end
+
+    person = Person.new
+    person.tags = ["ruby", "rails"]
+    person.counts = ["1", "2", "3"] # => [1, 2, 3]
+    person.ratings = ["3.5", "4.0"] # => [3.5, 4.0]
+
+    # JSON string assignment is also supported
+    person.counts = "[4, 5, 6]" # => [4, 5, 6]
+    ```
+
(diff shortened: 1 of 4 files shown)


fatkodima:
Isn't something like `attribute :tags, :string, array: true` already working? (https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute)

zakariaf:
@fatkodima The `array: true` option mentioned in the documentation is specific to PostgreSQL in **ActiveRecord**, not a general **ActiveModel** feature. 

`array: true`  tells ActiveRecord to use PostgreSQL array types, but `Type::Array` is just a proper type implementation in the type system that works without a database.
It's just like filling a gap for ActiveModel users who don't use ActiveRecord or PostgreSQL but still need array attribute support with type casting.

rafaelfranca:
Duplicated of #51421

PR #41043, opened 2021-01-07
Change the default digest for new apps to SHA256

dbussink:
### Summary

As mentioned in https://github.com/rails/rails/pull/40770#issuecomment-748347066 we should default to SHA256 where SHA1 is used today. This switches over the ActiveSupport::Digest to use SHA256 for new applications.

It also updates the constants to always refer to and use the OpenSSL constants as well, as also discussed in that PR.

PR #29716, opened 2017-07-08
Add Key Rotation to MessageEncryptor and MessageVerifier and simplify the Cookies middleware

mjc-gh:
### Summary

This PR introduces `ActiveSupport::KeyRotator` which wraps `KeyGenerator` as well as `MessageEncryptor` and `MessageVerifier` and provides an interface for easily rotating between different encryption ciphers or message digests, salts, and secrets.

This PR also simplifies the cookies middleware and removes several of the internal legacy middlewares. Upgrading both signed and encrypted legacy cookies is now handled by rotation capable middlewares which each use a `KeyRotator` instance. Is also introduces a new interface for configuring the middleware.

### Other Information

This feature spawned from discussions had in the Rails’ basecamp.

I still need to update the configuring and security markdown guides with details of this change. I plan on adding a new section the security guide detailing how secrets and salts can be rotated. I also want to add more rotation tests to `railties/test/application/middleware/cookies_test.rb`.

Any and all feedback is welcomed!

/cc @matthewd @kaspth

PR #55017, opened 2025-05-05
Default to SHA-256 for cookie signing

### LINKED ISSUE #54926: Signed cookies use SHA-1 by default
### Steps to reproduce

In a new rails app, create a signed cookie with `cookies.signed[:foo] = 'bar'`.

### Expected behavior

Signed cookies use a secure-by-default signing method such as SHA-256.

### Actual behavior

In the controller if you then call `cookies.send :signed_cookie_digest`, you get "SHA1".

https://guides.rubyonrails.org/configuring.html#config-action-dispatch-signed-cookie-digest does document that the default for signed_cookie_digest is "SHA1", though the upgrade guide for Rails 7.0 implied this was changing to SHA-256:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#digest-class-for-activesupport-digest-changing-to-sha256

and the active support digest class did in fact change to SHA-256:
https://guides.rubyonrails.org/configuring.html#config-active-support-hash-digest-class

but not the cookie digest class it seems.

### System configuration

**Rails version**: 8.0.1

**Ruby version**:  3.3.5, I suspect any version would work.

jg23497:
### Motivation / Background

Fixes #54926. This pull request has been created so that we use the SHA256 algorithm by default for signed cookies, moving away from SHA1.

Unless overriding `config.action_dispatch.signed_cookie_digest`, Rails currently defaults to SHA1 for the generation of the signed cookie HMAC.

As the [issue](https://github.com/rails/rails/issues/54926) implies, use of the SHA1 algorithm is [no longer recommended](https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm), even if there is currently no known attack vector specifically for HMACs. Rails is already migrating to SHA256 in related areas.

### Detail

This pull request changes the default value of `signed_cookie_digest`.

**Note**: Rails applications relying on the previous SHA1 default will no longer read these SHA1-signed cookies unless a cookie rotation is configured:

```
Rails.application.config.action_dispatch.cookies_rotations.tap do |cookies|
  cookies.rotate :signed, digest: "SHA1"
end
```

This rotation allows the upgrading applications to read SHA1 signed cookies created prior to this change without breaking sessions or user data.

The rotation can be removed once no clients are sending cookies signed using the previous digest (e.g. the session lifetimes or cookie expiration times have passed). 

#### Example

Prior to this change, imagine we have the following action method where we set a signed cookie:

```ruby
def index
  cookies.signed[:foo] = { value: 'bar' }
end
```

Cookie value:

```
eyJfcmFpbHMiOnsibWVzc2FnZSI6IkltSmhjaUk9IiwiZXhwIjpudWxsLCJwdXIiOiJjb29raWUuZm9vIn19--a1d2517adb63df049f06fe652040f18b10555ebd
```

Base64 decoding the first part, we see: `{"_rails":{"message":"ImJhciI=","exp":null,"pur":"cookie.foo"}}`. The SHA1 HMAC follows: `a1d2517adb63df049f06fe652040f18b10555ebd`.

After applying this PR's change, restarting the server and attempting to read the cookie using another action method:

```ruby
def foo
  message = cookies.signed[:foo]
  render plain: "#{message || "Unable to read cookie"}\n"
end
```

We see `Unable to read cookie` because the client presented a SHA1-signed cookie, but we're now using SHA256.

We can apply the rotation to gracefully handle this scenario:

```ruby
# config/initializers/cookie_rotations.rb
Rails.application.config.action_dispatch.cookies_rotations.tap do |cookies|
  cookies.rotate :signed, digest: "SHA1"
end
```

The action method now renders `bar` and the cookie's value is updated, including the new SHA256 HMAC:

```
eyJfcmFpbHMiOnsibWVzc2FnZSI6IkltSmhjaUk9IiwiZXhwIjpudWxsLCJwdXIiOiJjb29raWUuZm9vIn19--80d33014a7c363278ff169d809a8740399f0fbaec9d6e944cd6dfed89581f83c
```

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -61,6 +61,7 @@ Below are the default values associated with each target version. In cases of co
 #### Default Values for Target Version 8.1
 
 - [`config.action_controller.escape_json_responses`](#config-action-controller-escape-json-responses): `false`
+- [`config.action_dispatch.signed_cookie_digest`](#config-action-dispatch-signed-cookie-digest): `"SHA256"`
 - [`config.yjit`](#config-yjit): `!Rails.env.local?`
 
 #### Default Values for Target Version 8.0
@@ -151,6 +152,7 @@ Below are the default values associated with each target version. In cases of co
 #### Default Values for Target Version 5.2
 
 - [`config.action_controller.default_protect_from_forgery`](#config-action-controller-default-protect-from-forgery): `true`
+- [`config.action_dispatch.signed_cookie_digest`](#config-action-dispatch-signed-cookie-digest): `"SHA1"`
 - [`config.action_dispatch.use_authenticated_cookie_encryption`](#config-action-dispatch-use-authenticated-cookie-encryption): `true`
 - [`config.action_view.form_with_generates_ids`](#config-action-view-form-with-generates-ids): `true`
 - [`config.active_record.cache_versioning`](#config-active-record-cache-versioning): `true`
@@ -2067,7 +2069,7 @@ Sets the cipher to be used for encrypted cookies. This defaults to
 
 #### `config.action_dispatch.signed_cookie_digest`
 
-Sets the digest to be used for signed cookies. This defaults to `"SHA1"`.
+Sets the digest to be used for signed cookies. This defaults to `"SHA256"`.
 
 #### `config.action_dispatch.cookies_rotations`
 
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -152,7 +152,6 @@ Below are the default values associated with each target version. In cases of co
 #### Default Values for Target Version 5.2
 
 - [`config.action_controller.default_protect_from_forgery`](#config-action-controller-default-protect-from-forgery): `true`
-- [`config.action_dispatch.signed_cookie_digest`](#config-action-dispatch-signed-cookie-digest): `"SHA1"`
 - [`config.action_dispatch.use_authenticated_cookie_encryption`](#config-action-dispatch-use-authenticated-cookie-encryption): `true`
 - [`config.action_view.form_with_generates_ids`](#config-action-view-form-with-generates-ids): `true`
 - [`config.active_record.cache_versioning`](#config-active-record-cache-versioning): `true`
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -145,6 +145,7 @@ module Rails
 
           if respond_to?(:action_dispatch)
             action_dispatch.use_authenticated_cookie_encryption = true
+            action_dispatch.signed_cookie_digest = "SHA1"
           end
 
           if respond_to?(:active_support)
@@ -358,6 +359,10 @@ module Rails
 
           if respond_to?(:action_controller)
             action_controller.escape_json_responses = false
+          end
+
+          if respond_to?(:action_dispatch)
+            action_dispatch.signed_cookie_digest = "SHA256"
           end
         else
           raise "Unknown version #{target_version.to_s.inspect}"
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -145,7 +145,6 @@ module Rails
 
           if respond_to?(:action_dispatch)
             action_dispatch.use_authenticated_cookie_encryption = true
-            action_dispatch.signed_cookie_digest = "SHA1"
           end
 
           if respond_to?(:active_support)
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Default to SHA256 for the signed cookie digest, replacing the previous SHA1 default.
+
+    To ensure the legacy SHA1-signed cookies can still be read, Rails applications relying on the
+    previous default should define a signed cookie rotation until all these cookies have expired
+    or been replaced. For example:
+
+    ```ruby
+    # config/initializers/cookie_rotations.rb
+    Rails.application.config.action_dispatch.cookies_rotations.tap do |cookies|
+      cookies.rotate :signed, digest: "SHA1"
+    end
+    ```
+
+    *Jan Gulan*
+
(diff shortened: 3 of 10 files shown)


jg23497:
Reviewers, this is my first proposed contribution to rails/rails and so I am not sure whether this change warrants a CHANGELOG entry — please let me know. My primary concern was ensuring that the note on cookie rotations is included in the upcoming release notes.

guides/source/configuring.md
```diff
@@ -2067,7 +2067,7 @@ Sets the cipher to be used for encrypted cookies. This defaults to
 
 #### `config.action_dispatch.signed_cookie_digest`
 
-Sets the digest to be used for signed cookies. This defaults to `"SHA1"`.
+Sets the digest to be used for signed cookies. This defaults to `"SHA256"`.
```
Earlopain, line 2070: The default should depend on `load_defaults` so that upgrading/legacy apps continue to work. Previous PRs for similar things should give a good template you can take inspiration from: https://github.com/rails/rails/pull/41043

This probably also needs some update https://github.com/rails/rails/blob/96a2f03a2a82e91ab4438b625ba47bd8ed917b45/guides/source/security.md#L333-L347

There are also some tests like https://github.com/rails/rails/blob/96a2f03a2a82e91ab4438b625ba47bd8ed917b45/actionpack/test/dispatch/cookies_test.rb#L630-L631 that don't make much sense anymore when the default is already SHA256.

Earlopain (commented):
👍 I have one concern which I'm fairly sure will need to be adressed. A changelog entry seems fine, but you should also add this to the upgrading rails guide, including the cookie rotator for the (probably many) people that need it.

Earlopain:
If you rebase this PR against main, most (all?) CI failures should be resolved.

KelseyDH:
I'm upgrading an application to Rails 7.1 right now and was about to delete our SHA256 cookie rotator.  I assume with this bug that I should I also set in `application.rb`:

```ruby
  config.action_dispatch.signed_cookie_digest = "SHA256"
```

To avoid this problem, until a bug fix gets merged into a new Rails release?

guides/source/configuring.md
jg23497: Thanks for the pointers!

I have done as you suggested with `load_defaults`. I notice that the `signed_cookie_digest` configuration was [added in Rails 5.2](https://github.com/rails/rails/pull/29716) and so have added the previous "SHA1" default to the method's "5.2" switch clause.

I will revisit cookies_test.rb soon to check the redundant tests, hopefully tomorrow.

jg23497:
> I'm upgrading an application to Rails 7.1 right now and was about to delete our [SHA1 to SHA256 cookie rotator from the Rails upgrade guides](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-change-requires-a-cookie-rotator). (As part of enabling `config.active_record.encryption.support_sha1_for_non_deterministic_encryption = false`)
> 
> I assume with this bug that I should also set:
> 
> ```ruby
>   # config/application.rb
>   
>   config.action_dispatch.signed_cookie_digest = "SHA256"
> ```
> 
> to avoid this problem, until this bug fix gets merged into a new Rails release?

I would agree that SHA256 is preferable here given the weakness of SHA1, which is generally deprecated. However, I understand that there is no known attack vector for SHA1 HMACs specifically and so no need for immediate concern.

jg23497:
> I am by no means an expert on hashing algorithms but I have done some reading and have found that the SHA-2 can be vulnerable to length extension attacks but SHA-3 is not. Should the hashing algorithm be set to SHA-3 by default ?

Good point that we may want to consider a different default. SHA256 would at least align with the other Rails defaults (e.g. ActiveRecord's hash digest class) and so would be a good contender. 

As far as I understand, length extension attacks do not apply to SHA256 HMACs because the attacker would also need to have knowledge of the key — this is not a regular hash that considers only the message/value itself.

guides/source/configuring.md
jg23497: I have updated the relevant unit tests.

I can also confirm that `config.load_defaults 8.1` produces a SHA256 signed cookie digest and `config.load_defaults 8.0` produces SHA1.

p8 (changes requested):

railties/lib/rails/application/configuration.rb
```diff
@@ -359,6 +360,10 @@ def load_defaults(target_version)
           if respond_to?(:action_controller)
             action_controller.escape_json_responses = false
           end
+
+          if respond_to?(:action_dispatch)
+            action_dispatch.signed_cookie_digest = "SHA256"
+          end
```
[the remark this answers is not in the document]
jg23497: Hi @p8. That approach will simplify the changes here -  I have done as you suggest.

There is no functional difference as you point out, but my slight concern with this approach is that the `signed_cookie_digest` method now/still implies that "SHA1" is the _current_ default signed cookie digest. However, that is no longer true now that SHA256 is the default and that SHA1 is deprecated, becoming part of the legacy configuration. In other words, we're falling back to a digest that we actually no longer recommend.

Happy to take guidance here, and the above may not be a concern, but sharing just in case.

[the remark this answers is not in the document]
p8: We try to not break applications when they get upgraded. The `frameworks_defaults` file is used to allow developers to upgrade to breaking changes on their own choosing. So if someone upgrades Rails but does not yet load the new defaults, nothing should change.

Adding a setting to older Rails version defaults could break things unexpectedly.
For example, someone might define the cookie digest before loading the defaults:
```ruby
config.action_dispatch.signed_cookie_digest = "SHA256"
config.load_defaults 8.0 # this would set the digest back to SHA1 if the defaults set it to SHA1
```

bdewater:
> I am by no means an expert on hashing algorithms but I have done some reading and have found that the SHA-2 can be vulnerable to length extension attacks but SHA-3 is not. Should the hashing algorithm be set to SHA-3 by default ?

No, HMAC is not vulnerable to this. See https://crypto.stackexchange.com/questions/3349/purpose-of-outer-key-in-hmac

PR #55018, opened 2025-05-05
Preserve enum methods when using becomes! on STI subclasses

### LINKED ISSUE #55001: Using `become!` on an STI model sets enum attribute to nil
### Steps to reproduce

```ruby
require 'bundler/inline'


gemfile(true) do
  source 'https://rubygems.org'

  gem 'rails'

  gem 'sqlite3'
end

require 'active_record/railtie'
require 'minitest/autorun'

ENV['DATABASE_URL'] = 'sqlite3::memory:'

  config.eager_load = false

  config.active_record.encryption.primary_key = 'primary_key'
  config.active_record.encryption.deterministic_key = 'deterministic_key'
  config.active_record.encryption.key_derivation_salt = 'key_derivation_salt'
end
Rails.application.initialize!

  create_table :foos, force: true do |t|
    t.string :state
    t.string :type
  end
end

class Foo < ActiveRecord::Base
end

class Bar < Foo
  enum(:state, { draft: 'draft' }, validate: true) # changing this to false will trigger an error
end

class Baz < Foo
  enum(:state, { published: 'published' })
end


  def test_enum
    bar = Bar.create!(state: 'draft')
    baz = bar.becomes!(Baz)
    baz.published!

    assert(baz.published?) # state is nil
  end
end
```

### Expected behavior

`Baz#state` should return "published" after doing `baz.published!`

### Actual behavior

`Baz#state` is returning nil

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.3.5

Yuhi-Sato:
### Motivation / Background

Fixes [rails#55001](https://github.com/rails/rails/issues/55001)

The `becomes!` method fails to preserve enum-based behavior when converting between STI (Single Table Inheritance) subclasses that define different `enum` values for the same attribute.

This is particularly problematic because `enum` defines predicate (`#published?`) and mutator (`#published!`) methods specific to each class. When switching classes via `becomes!`, the internal type mapping (`EnumType`) from the original class is retained, leading to incorrect validation and method behavior.

### Detail

This change ensures that when performing `becomes!`, each attribute is rebound to the destination class’s attribute type using `with_type`. This guarantees that enum accessors work correctly in the destination class context.

A test case using `LiveParrot` and `DeadParrot` subclasses of `Parrot` (each declaring their own enum `:breed`) demonstrates that the fix allows safe conversion between STI subclasses while maintaining proper enum behavior.

Before this change:

```ruby
    parrot = LiveParrot.create!(name: "Scipio", breed: "african")
    dead_parrot = parrot.becomes!(DeadParrot)
    dead_parrot.asian! # => raises ArgumentError
```

This failed due to the original enum mapping from `LiveParrot` being retained.

### Additional information

- The error is raised by `EnumType#assert_valid_value` when the new enum value is not recognized by the old enum mapping.
- This fix ensures consistency with developer expectations when using enums in STI environments.
- To avoid decryption errors for encrypted attributes
  (`ActiveRecord::Encryption::EncryptedAttributeType`), such attributes are
  explicitly skipped when rebinding types.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/test_case.rb
+++ b/activesupport/lib/active_support/test_case.rb
@@ -84,7 +84,7 @@ module ActiveSupport
       # writing your own implementation, you can set +parallelize_databases+, or configure it
       # via +config.active_support.parallelize_test_databases+.
       #
-      #   parallelize(workers: :number_of_processes, parallelize_databases: false)
+      #   parallelize(workers: :number_of_processors, parallelize_databases: false)
       #
       # Note that your test suite may deadlock if you attempt to use only one database
       # with multiple processes.


activerecord/test/models/mikan.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class Mikan < ActiveRecord::Base
+end
+
+class Unshu < Mikan
```
yuporon, line 6:
```suggestion
class Unshiu < Mikan
```
nit: 
I think the Romanization of Citrus unshiu is “Unshiu”, not “Unshu”🍊
ref) https://en.wikipedia.org/wiki/Citrus_unshiu

activerecord/test/schema/schema.rb
```diff
@@ -812,6 +812,11 @@
     t.datetime :updated_at
   end
 
+  create_table :mikans, force: true do |t|
+    t.string :state
+    t.string :type
+  end
```
eileencodes, line 818: Adding new tables is a perf hit to the database so we prefer to only do it if there's not already a model that can't be used. There's a ton of models with enums in the test models already, can you reuse one of those instead of adding new ones?

eileencodes (changes requested):
Thanks for the PR. I added a request for changes in the tests. This also needs a changelog entry.

activerecord/test/schema/schema.rb
Yuhi-Sato: Thanks for the reply. I have modified it to use the existing test model as you suggested.

eileencodes (approved):
Can you fix the merge conflicts? Thanks!

activerecord/CHANGELOG.md
```diff
@@ -2,6 +2,35 @@
...
+      enum :state, { published: "published" }
+    end
+
+    bar = Bar.create!(state: "draft")
+    baz = bar.becomes!(Baz)
+    baz.published! # Previously: raises ArgumentError
+    ```
+
+    Fixes #55001.
+
```
zzak, line 31:
```suggestion
```

We don't usually mention which issues are affected in the changelog

activerecord/CHANGELOG.md
```diff
@@ -2,6 +2,35 @@
 
     *Hartley McGuire*
```
zzak, line 4: Usually we ask that changelog entries go at the top of the file, even it is prone to merge conflicts :pray:

Yuhi-Sato: Sorry, I accidentally bundled all of my changes into your commit. I'll fix it as soon as possible. 🙇‍♂️🙇‍♂️

PR #55020, opened 2025-05-05
[RF-DOCS] Rails Application Template Guide - merge with Rails Generators Guide [ci-skip]

bhumi1102:
### Motivation/Background

The goal of this PR is to update the Rails Application Template Guide. In the process of updating this guide, it made sense to merge its content into the Generators Guide, which already had a section about Templates. The scope of this PR is the Rails Application Template sections (the rest of the existing Generators Guide will be updated later on).

### Testing

- [X]  Run guides:generate
- [X]  Run guides:lint


pandalowry:
Very needed update!

guides/source/generators.md
```diff
@@ -475,35 +496,255 @@ after_bundle do
...
-`-m` option to the `rails new` command:
+The above `template.rb` uses helper methods such as `after_bundle` and `rails_command`. You can see a list of available helper methods that templates can use in the [generator helper methods](#generator-helper-methods) section. The following sections explains some of them with examples.
 
-```bash
-$ rails new my_cool_app -m path/to/template.rb
+Rails Template API
+------------------
+
+The Rails template Ruby scripts have access to several helper methods using a [DSL](https://en.wikipedia.org/wiki/Domain-specific_language) (Domain Specific Language).
+
```
pandalowry, line 510: Details on which methods are available for use in templates will be below, but you can look up the capabilities directly in the [Rails API documentation] (https://api.rubyonrails.org/classes/Rails/Generators/Actions.html)
Need more links, or description about fact: not all available methods linked in api. Class hierarchy.

guides/source/generators.md
```diff
@@ -449,12 +446,36 @@ $ bin/rails generate scaffold Comment body:text
       create      app/views/comments/show.json.jbuilder
 ```
 
-Application Templates
----------------------
+Creating Application Templates
+------------------------------
+
+Application templates are a little different from generators. While generators add files to an existing Rails application (models, views, etc.), templates are used to automate the setup of a new Rails application. Templates are written as a Ruby script.
+
```
pandalowry, line 453: if you spend a lot of time on individual actions manually after you have created an application, perhaps you need a template that will perform these actions automatically when creating an application

pandalowry (commented):
two points, where imho need more explain and links. But, I prepare PR, not comments.

p8:
@bhumi1102 The "Rails Template API" section has a lot of overlap with the "Generator Helper Methods" section. They basically mention the same methods, although the "Rails Template API" section is a lot more detailed.
As the "Creating Application Templates" section mentions these helper methods maybe merge "Rails Template API" into "Generator Helper Methods" and put it before "Creating Application Templates"?
- ...
- "Overriding Rails Generators" 
- "Generator Helper Methods" (with most content from "Rails Template API")
- "Creating Application Templates"
- "Testing Generators"

guides/source/generators.md
```diff
@@ -449,12 +446,44 @@ $ bin/rails generate scaffold Comment body:text
...
+```
+
+Templates don't have to be stored locally, you can also specify a URL instead of
+a path:
+
+```bash
+$ rails new blog -m http://example.com/template.rb
+$ bin/rails app:template LOCATION=http://example.com/template.rb
+```
+
```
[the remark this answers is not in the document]
bhumi1102: Good call!

guides/source/generators.md
```diff
@@ -449,12 +446,44 @@ $ bin/rails generate scaffold Comment body:text
...
+Application templates are a little different from generators. While generators
+add files to an existing Rails application (models, views, etc.), templates are
+used to automate the setup of a new Rails application. Templates are written as
+a Ruby script.
+
+Let's see how to use a template while creating a new Rails application and also
+an example of a template Ruby script.
+
+### Usage
```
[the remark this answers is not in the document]
bhumi1102: Agreed! Updated the flow to define the template first and then use it and reworded some things.

bhumi1102:
> On the other hand the methods used in "Creating Application Templates" are clearly explained 🤔 so maybe the hlper methods can still be mentioned after "Creating Application Templates"?

Hi @p8 - I noticed the overlaps between those two sections...thought maybe it would be nice for readers to have a list of methods so left it. But I have revamped it now - combined the two sections, added new examples of how to use more helpers, added links to api docs for all methods referenced and links to the overall Thor rubydocs for readers to get more details. I think it's better. 

Lastly, I do like showing an example of how to create and use a template first, before going into the nitty gritty for listing available methods. So kept that ordering. Let me know what you think 😄

guides/source/generators.md
```diff
@@ -475,58 +482,316 @@ after_bundle do
...
+following sections shows how to use more of these methods with examples.
+
+Rails Template API
+------------------
+
+The template Ruby scripts have access to several helper methods using a
+[DSL](https://en.wikipedia.org/wiki/Domain-specific_language) (Domain Specific
+Language). These methods are part of the Rails Template API and you can find
+more details at [`Thor::Actions`][] and [`Rails::Generators::Actions`][] API
+documentation.
```
p8, line 525: Generators and Application Templates have access to the same methods (the application template script is basically run as an inline generator).
So  this section should probably make clear these methods are available to both?
Also as these methods are defined in Rails::Generators::Actions refer to the generators API instead?
Maybe something like:
```suggestion
Rails Generators API
--------------------

Generators and the template Ruby scripts have access to several helper methods using a
[DSL](https://en.wikipedia.org/wiki/Domain-specific_language) (Domain Specific
Language). These methods are part of the Rails Generators API and you can find
more details at [`Thor::Actions`][] and [`Rails::Generators::Actions`][] API
documentation.
```

Or maybe even refer to them as actions in the title (like the namespace)?
```suggestion
Rails Generators Actions
------------------------
....
```

p8:
@bhumi1102 This new section looks great! I've left one last comment.

PR #53790, opened 2024-11-30
Use `_N` as a parallel tests databases suffixes

fatkodima:
Fixes #53780 (see there for a discussion).

We can even skip using `-` or `_` and just use a `db_name0`, `db_name1` naming pattern etc (like in https://github.com/grosser/parallel_tests).

PR #55023, opened 2025-05-06
Cleanup parallel test dbs after run

zzak:
This is an alternative to possibly teaching `bin/rails db:drop` about parallel dbs.

For consequtive re-runs with an up-to-date schema, this does add some overhead as each db will have to be made from scratch.

See: #53790

I will investigate changing `db:drop` to check for parallel dbs, but I think this is a safer and less complicated alternative if anyone is interested that trade-off.

### DIFF
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -16,6 +16,12 @@ module ActiveRecord
       end
     end
 
+    ActiveSupport::Testing::Parallelization.run_cleanup_hook do |i|
+      if ActiveSupport.parallelize_test_databases
+        cleanup_test_databases(id: i, env_name: ActiveRecord::ConnectionHandling::DEFAULT_ENV.call)
+      end
+    end
+
     def self.create_and_load_schema(i, env_name:)
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
@@ -27,6 +33,14 @@ module ActiveRecord
     ensure
       ActiveRecord::Base.establish_connection
       ENV["VERBOSE"] = old
+    end
+
+    def self.cleanup_test_databases(id:, env_name:)
+      ActiveRecord::Base.configurations.configs_for(env_name: env_name).each do |db_config|
+        db_config._database = "#{db_config.database}_#{id}"
+
+        ActiveRecord::Tasks::DatabaseTasks.drop(db_config)
+      end
     end
   end
 end
--- a/activerecord/test/cases/test_databases_test.rb
+++ b/activerecord/test/cases/test_databases_test.rb
@@ -101,5 +101,30 @@ class TestDatabasesTest < ActiveRecord::TestCase
       ActiveRecord::Base.establish_connection(:arunit)
       ENV["RAILS_ENV"] = previous_env
     end
+
+    def test_databases_are_cleaned_up_after_run
+      previous_env, ENV["RAILS_ENV"] = ENV["RAILS_ENV"], "arunit"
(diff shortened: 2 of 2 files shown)


eileencodes:
Why do the databases need to be dropped between runs? The overhead would be a lot in a large app like Shopify or GitHub.

zzak:
Thanks for taking a look.

I agree with you, but I found a couple of cases where teaching drop maybe difficult:

* Where we have fewer tests than the threshold
* When [parallelize](https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize) is called directly in the test case or helper

Came up with a naive approach but cases where weird things are happening, I guess we could just swallow the exception?

```diff
--- a/activerecord/lib/active_record/tasks/database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/database_tasks.rb
@@ -224,7 +224,24 @@ def drop_all
       end

       def drop_current(environment = env)
-        each_current_configuration(environment) { |db_config| drop(db_config) }
+        each_current_configuration(environment) do |db_config|
+          drop(db_config)
+
+          if ActiveSupport.parallelize_test_databases && db_config.env_name == "test"
+            case
+            when ENV["PARALLEL_WORKERS"]
+              workers = ENV["PARALLEL_WORKERS"].to_i
+            else
+              workers = (Concurrent.available_processor_count || Concurrent.processor_count).floor
+            end
+
+            workers.times do |i|
+              new_config = db_config.dup
+              new_config._database = "#{db_config.database}_#{i}"
+              drop(new_config)
+            end
+          end
+        end
```

eileencodes:
> Where we have fewer tests than the threshold

but then the tests wouldn't be parallelized unless someone was using `PARALLEL_WORKERS` which I don't think is _that_ common. I mostly have it in there as a debug tool.

> When [parallelize](https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize) is called directly in the test case or helper

It doesn't parallelize unless it's called in the test helper, so that's required.

I don't think calling it in the test case is that common. That kind of defeats the purpose of the feature. Why would you parallelize only one test case file rather than the whole suite? I think we don't need to worry about edge cases here, if your app is doing an odd thing you can write your own drop hook.

Other than "it's annoying" is it causing any problems to leave the dbs behind?

zzak:
> unless someone was using `PARALLEL_WORKERS`

It didn't occur to me that was only useful for debugging, so I think we can leave it out.

> I think we don't need to worry about edge cases here, if your app is doing an odd thing you can write your own drop hook.

:100: I think we can just swallow the error if we try to drop a db that doesn't exist, there's no point in reporting it to the user.

> Other than "it's annoying" is it causing any problems to leave the dbs behind?

I tried to come up with a few examples, but they're probably not very good:

* If you're using sqlite and you end up with a ton of random files in your storage dir
* Depending on the size of your test database even it could impact disk space
* I guess if you have tons of databases on your system you could run into file descriptor limits

All of those are depending on the size of your database (after cleanup) and how many cores you have.

I guess the motivation is to solve the surprise of "I just dropped my database, why are there still 16 sqlite files in my project folder?". I don't have a better answer than that.

If we wanted to do it though, I think we just use the CPU count and try to drop them or move on?

https://github.com/rails/rails/blob/a3b37ff166a925f7e069c9068f7ee92808da6e82/activesupport/lib/active_support.rb#L107

This was also my guess at finding out "are they using parallel test databases or not" but I'm not so sure on that, tbh. It is added to the test_helper by default for new apps, fwiw.

Anyway, I will close this since we want to keep the parallel dbs to save time in subsequent runs and that is more valuable than potentially leaving behind garbage when they decide to drop their local database.

PR #55029, opened 2025-05-07
Avoid DISTINCT on COUNT queries when possible

### LINKED ISSUE #55027: Counts with eager loaded associations include unnecessary DISTINCT in certain scenarios
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
  gem "mocha"
end

require "minitest/autorun"
require "mocha/minitest"

  config.eager_load = false

end
Rails.application.initialize!

  end

    t.integer :post_id
  end
end

  has_many :comments
end

  belongs_to :post
end

  def test_expected_count_query
    post = Post.create!
    2.times { post.comments.create! }

    Rails.logger.expects(:debug).with(includes('SELECT COUNT(DISTINCT')).once
    Rails.logger.expects(:debug).with(includes('SELECT COUNT(*')).once

    assert_equal 1, Post.eager_load(:comments).count
    assert_equal 2, Comment.eager_load(:post).count
  end
end
```

### Expected behavior

This isn't a bug so much as an inefficiency in `.count` queries. When running left-outer-joining on a `belongs_to` relationship, ActiveRecord changes `COUNT(*)` to `COUNT(DISTINCT "id")`, but this DISTINCT is unnecessary because the join will never include duplicate IDs if left-outer-joining from the child to the parent in a one-to-many relationship.

This is normally fairly harmless, but when running counts against large tables, this can have significant performance implications.

This query doesn't need a DISTINCT because the left-outer-join will never include duplicate Commit IDs because of the one-to-many relationship:

```ruby
Comment.eager_load(:post).count
# => SELECT COUNT(*) FROM "comments" LEFT OUTER JOIN "posts" ON "posts"."id" = "comments"."post_id"
```

### Actual behavior

The query includes an unnecessary DISTINCT (due to Post's one-to-many relationship with Comment):

```ruby
Comment.eager_load(:post).count
# => SELECT COUNT(DISTINCT "comments"."id") FROM "comments" LEFT OUTER JOIN "posts" ON "posts"."id" = "comments"."post_id"
```

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.4.3

chrisyuska:
### Motivation / Background

Fixes #55027

When calling `.count` on an eager-loaded query where the base table `belongs_to` the eager-loaded table, AR switches `COUNT(*)` to a `COUNT(DISTINCT "id")`.  However, this is unnecessary as there will never be duplicates in a `belongs_to` relationship like this.

This isn't normally an issue, but when running COUNT queries on larger tables (at least in Postgres), I've seen this turn ~100ms queries into ~30s queries.

This Pull Request has been created because a very common scenario is paginating data that might have an `order` clause on an included table, which is then paginated with gems like Kaminari that will call `count` on the relation without additional reworking of the query.  e.g.:

```ruby
Comment.includes(:post).order(posts: { title: :asc }, comments: { created_at: :desc }).count
# => SELECT COUNT(DISTINCT "comments"."id") FROM "comments" LEFT OUTER JOIN "posts" ON "posts"."id" = "comments"."post_id"
```

### Detail

This Pull Request changes ActiveRecord's Calculation logic around COUNT queries to only add a `DISTINCT` column when any eager-loaded associations are not belongs_to associations.

### Additional information

One workaround we've used is to avoid the eager-load and just change the query to:

```ruby
Comment.left_outer_joins(:post).order(posts: { title: :asc }, comments: { created_at: :desc }).count
```

This works fine, but we lose the eager-loaded `post` association in doing so.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/calculations.rb
+++ b/activerecord/lib/active_record/relation/calculations.rb
@@ -232,7 +232,7 @@ module ActiveRecord
         relation = apply_join_dependency
 
         if operation == "count"
-          unless distinct_value || distinct_select?(column_name || select_for_count)
+          if distinct_count?(relation, column_name)
             relation.distinct!
             relation.select_values = Array(model.primary_key || table[Arel.star])
           end
@@ -614,6 +614,17 @@ module ActiveRecord
           return type if type
         end
         nil
+      end
+
+      def distinct_count?(relation, column_name)
+        return false if distinct_value || distinct_select?(column_name || select_for_count)
+        return true if limit_value || relation.limit_value
+
+        joined_fields = references_values | eager_load_values
+        joined_fields.any? do |field|
+          association  = model._reflect_on_association(field)
+          !association&.belongs_to?
+        end
       end
 
       def type_cast_pluck_values(result, columns)
--- a/activerecord/test/cases/calculations_test.rb
+++ b/activerecord/test/cases/calculations_test.rb
@@ -379,6 +379,24 @@ class CalculationsTest < ActiveRecord::TestCase
     assert_queries_count(1) { assert_equal 11, posts.count(:all) }
   end
 
+  def test_count_with_eager_loading_and_custom_order_from_child_association
+    comments = Comment.includes(:post).order("posts.id")
+    assert_queries_count(1) { assert_equal 12, comments.count }
+    assert_queries_count(1) { assert_equal 12, comments.count(:all) }
+  end
+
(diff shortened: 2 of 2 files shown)


chrisyuska:
I suspect this PR will need some work before it can be merged.  Totally open to alternative approaches or just closing it too if the complexity trade-off isn't worth it.

PR #55033, opened 2025-05-08
Add support for Cache-Control request directives

egg528:
### Motivation / Background

This Pull Request has been created because the HTTP Cache-Control request directives were not fully supported in Rails. According to RFC 9111, HTTP clients can use various cache control directives in their requests to control how caching is handled. This implementation adds comprehensive support for these directives, with special attention to the `max-stale` directive which can be used with or without a value.

### Detail

This Pull Request changes the following:

1. Adds a new `CacheControlDirectives` class that parses and provides access to HTTP Cache-Control request directives
2. Implements support for boolean directives (`no-cache`, `no-store`, `no-transform`, `only-if-cached`)
3. Implements support for value directives (`max-age`, `min-fresh`, `stale-if-error`)
4. Provides special handling for the `max-stale` directive which can appear both with a value (e.g., `max-stale=300`) or without a value (indicating unlimited staleness)
5. Adds comprehensive test coverage for all supported directives
6. Updates the CHANGELOG to document the new functionality

This enhancement allows Rails applications to better integrate with HTTP caching mechanisms and provides more control over how requests interact with cached resources.

### Additional information

- The implementation follows RFC 9111 (HTTP Caching): https://www.rfc-editor.org/rfc/rfc9111.html#name-cache-control

- The `max-stale` directive required special handling as it can appear without a value, in which case it indicates the client is willing to accept a response no matter how stale it is.
- This PR is a comprehensive implementation that incorporates feedback and suggestions from the previous PR #55008. (@byroot )

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,42 @@
+*   Add comprehensive support for HTTP Cache-Control request directives according to RFC 9111.
+
+    Provides a `request.cache_control_directives` object that gives access to request cache directives:
+
+    ```ruby
+    # Boolean directives
+    request.cache_control_directives.only_if_cached?  # => true/false
+    request.cache_control_directives.no_cache?        # => true/false
+    request.cache_control_directives.no_store?        # => true/false
+    request.cache_control_directives.no_transform?    # => true/false
+
+    # Value directives
+    request.cache_control_directives.max_age          # => integer or nil
+    request.cache_control_directives.max_stale        # => integer or nil (or true for valueless max-stale)
+    request.cache_control_directives.min_fresh        # => integer or nil
+    request.cache_control_directives.stale_if_error   # => integer or nil
+
+    # Special helpers for max-stale
+    request.cache_control_directives.max_stale?         # => true if max-stale present (with or without value)
+    request.cache_control_directives.max_stale_unlimited? # => true only for valueless max-stale
+    ```
+
+    Example usage:
+
+    ```ruby
+    def show
+      if request.cache_control_directives.only_if_cached?
+        @article = Article.find_cached(params[:id])
+        return head(:gateway_timeout) if @article.nil?
+      else
+        @article = Article.find(params[:id])
+      end
+
+      render :show
+    end
+    ```
+
+    *egg528*
+
 *   Add assert_in_body/assert_not_in_body as the simplest way to check if a piece of text is in the response body.
 
     *DHH*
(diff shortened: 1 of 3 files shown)


actionpack/CHANGELOG.md
```diff
@@ -140,4 +140,42 @@
...
+    ```ruby
+    # Boolean directives
+    request.cache_control_directives.only_if_cached?  # => true/false
+    request.cache_control_directives.no_cache?        # => true/false
+    request.cache_control_directives.no_store?        # => true/false
+    request.cache_control_directives.no_transform?    # => true/false
+
+    # Value directives
+    request.cache_control_directives.max_age          # => integer or nil
+    request.cache_control_directives.max_stale        # => integer or nil (or true for valueless max-stale)
```
byroot, line 156: This is a bit of a problem, users are likely to assume `max_stale` if truthy is an Integer and not handle `true` correct.

actionpack/lib/action_dispatch/http/cache.rb
```diff
@@ -63,6 +60,108 @@ def fresh?(response)
...
+                case
+                when SPECIAL_DIRECTIVES.include?(directive_name)
+                  directives[directive_name] = directive_value ? directive_value.to_i : true
+                when directive_value && VALUE_DIRECTIVES.include?(directive_name)
+                  directives[directive_name] = directive_value.to_i
+                when BOOLEAN_DIRECTIVES.include?(directive_name)
+                  directives[directive_name] = true
+                end
+              end
+            end
```
byroot, line 163: Rather than to decorate a Hash, I think it would be better to store individual values in instance variables.

actionpack/CHANGELOG.md
byroot: Nevermind, I see `max_stale_unlimited?`. I suppose that is acceptable

byroot (commented):
The interface look good to me. But I'd rather have an implementation that doesn't use a Hash.

actionpack/lib/action_dispatch/http/cache.rb
egg528: Thanks for suggesting such a good improvement 👍

egg528:
@byroot 
Thanks for suggesting such a good improvement 👍

actionpack/lib/action_dispatch/http/cache.rb
egg528: Thanks for suggesting such a good improvement 👍

PR #55038, opened 2025-05-09
Allow prepending error middleware

dersam:
ActiveSupport::ErrorReporter#add_middleware now receives an `insert` kwarg. If this is set to `:prepend`, the middleware will be added to the top of the stack, rather than the bottom. The default behavior remains to append middleware.

This provides better control over middleware provided automatically vs context dependent middleware in an app.

I chose to pass symbols to set the behavior instead of a boolean to allow for additional behaviors in the future.


dersam:
Ignore this. Looking at the [Rails middleware stack](https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack), this diverges from that design, which this system is explicitly trying to follow. Closing.

Issue #46685, opened 2022-12-09
Changes in behaviour with `#dup` and `#concat`


adrianna-chang-shopify:
👋 Hi folks! 

This is not necessarily a bug, but a change in behaviour introduced by https://github.com/rails/rails/commit/2db098c177d51f5bd6ecad57a545da5a5e0644c0 that is impacting Shopify's Core monolith.

Some background: There's a pattern being used in a couple places in our Rails monolith where associations are added to an already-persisted model without making changes to the database. This is so that complex calculations involving orders, sales taxes, etc. can be performed without persisting those associations to the database.

### Steps to reproduce
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", ref: "c6b227b95b631ead7ffd5c8984b16b0fbe5b3ca4" # Commit before Alex's changes
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
  end

  create_table :comments, force: true do |t|
    t.integer :post_id
  end
end

class Post < ActiveRecord::Base
  has_many :comments
end

class Comment < ActiveRecord::Base
  belongs_to :post
end

class BugTest < Minitest::Test
  def test_association_stuff
    post = Post.create
    post.comments << Comment.new

    assert_equal 1, post.comments.count

    # Add associations to already-persisted post without making changes to the database
    another_post = post.dup
    another_post.id = post.id

    assert_equal 1, another_post.comments.size

    another_post.comments << Comment.new
    assert_equal 2, another_post.comments.size
  end
end
```
(Test passes)

When run with `  gem "rails", github: "rails/rails", branch: "main"`:
```ruby
Failure:
BugTest#test_association_stuff [scripts/46392_strict_loading_has_many.rb:55]:
Expected: 2
  Actual: 1
````

### Expected behavior
As shown in the script, this pattern of duping an Active Record, assigning the id, and then adding records to the collection **previously resulted in the persisted record's collection being preserved on the duped record**.

### Actual behavior
[...]

Issue #42097, opened 2021-04-28
`assign_attributes` persists changes to join table relationships for already persisted records


thewatts:
### Steps to reproduce
Test to reproduce:

Copy the below code into a file, and run `ruby {filename}.rb`

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :movies, force: true do |t|
  end

  create_table :characters, force: true do |t|
  end

  create_table :movie_characters, force: true do |t|
    t.integer :movie_id
    t.integer :character_id
  end
end

class Movie < ActiveRecord::Base
  has_many :movie_characters
  has_many :characters, through: :movie_characters
end

class Character < ActiveRecord::Base
  has_many :movie_characters
  has_many :movies, through: :movie_characters
end

class MovieCharacter < ActiveRecord::Base
  belongs_to :movie
  belongs_to :character
end

class BugTest < Minitest::Test
  def test_association_stuff
    # initialize
    hobbit = Movie.new

    luke = Character.create!
    frodo = Character.create!

    assert_equal 0, hobbit.movie_characters.count

    # assigning in-memory an incorrect character, Luke, to The Hobbit
    hobbit.assign_attributes(character_ids: [luke.id])
    assert_equal 0, hobbit.movie_characters.where(character_id: luke.id).count

    # persisting with incorrect character
    hobbit.save
    assert_equal 1, hobbit.movie_characters.where(character_id: luke.id).count

    # replacing in-memory with correct character, Frodo
    hobbit.assign_attributes(character_ids: [frodo.id])

    # Luke should still be associated with The Hobbit, since we haven't persisted yet
    assert_equal 1, hobbit.movie_characters.where(character_id: luke.id).count
    # Frodo should not be associated with the hobbit yet
    assert_equal 0, hobbit.movie_characters.where(character_id: frodo.id).count

    # persisting with correct character
    hobbit.save

    # Luke should no longer be associated with The Hobbit
[...]

Issue #17368, opened 2014-10-23
Updating associations with assign_attributes should not persist


runephilosof:
assign_attributes should not persist the attributes.
But it does for associations.
```ruby
class Firm
  has_many :clients
  validates :clients, presence: true
end
class Client
  belongs_to :firm
end
firm = Firm.new
firm.clients.build
firm.save
firm.assign_attributes({clients: []})
saved = firm.save
```
saved becomes false, but it has already been  persisted in the call to assign_attributes.

runephilosof:
Actually, it would be nice it this could be changed in rails 5.
I would like the assignment of a `has_one, has_many, has_and_belongs_to_many` to not save automatically be default, but instead do the same as the `collection.build` method.
This would be way more intuitive.

rafaelfranca:
We don't take feature request on the issue tracker. We recommend you to try to implement the feature and send us a pull request or if you are in doubt if the feature will be accepts ask in Rails Core mailing list for feedback.

JasonBarnabe:
I've posted my proposal for this at https://groups.google.com/forum/?fromgroups#!topic/rubyonrails-core/-BP3yigVvII

samiggapps:
I am having the same issue any updates on this??

remofritzsche:
I'd also be very interested in a fix. I think the default behavior of Rails is incredibly dangerous and completely counter-intuitive.

### Example

Consider the following scenario:

```ruby
class User
  has_many :groups
  validates :name, presence: true
end

class UsersController
  def update
    User.transaction do
      @user = User.find(params[:id])

      if @user.update_attributes(params.require(:user).permit!)
        redirect_to @user
      else
        render action: :edit
      end
    end
  end
end
```

```haml
-# Example with simple_form to simplify example
= simple_form_for user do |f|
    = f.association :groups, as: :check_boxes
```

For this example, we submit the following params:

```ruby
{ user: { name: '', group_ids: [1, 2, 3] } }
```

### Expected behavior

As we send an empty name, the validation is failing. Therefore, `update_attributes` returns `false`, nothing is persisted and the user is presented with the form again, showing the validation errors.

### Actual behavior

As expected, the validation fails. The very dangerous thing though is that **the group associations are saved anyway**, as the Rails-generated `_ids`-setter (and thus `assign_attributes` and `update_attributes`) automatically save the association. 

### Temporary workaround

[...]

PR #55041, opened 2025-05-09
Add option to defer persisting additions/removals on collection associations. Add `marked_for_deletion`.

WelpThatWorked:
### Motivation / Background

Presently, any additions or removals of existing records to a collection association are persisted to the database immediately, rather than waiting for save to be called manually. This can be unexpected or undesirable in cases such as generating a preview or looking back at what changes just occurred. While there exist some workarounds, they seem a bit cumbersome, and there have occasionally been requests for it to not be persisted immediately, so I thought having a toggle on that behavior might be more convenient.

### Detail

- Add the `:defer` option for collection associations with autosave enabled. Setting this to true will defer the persisting of additions/removals until the parent (or child, in the case of has_many) is saved.
    - Applies to methods using `concat()` or `replace()`. Using `collection.destroy(record)` or `collection.delete(record)` still persists immediately, since `mark_for_deletion`/`destruction` is simultaneously available.
    - Removals are flagged for deletion, while additions are autosaved due to changes in the record (`has_many`) or changes in the through record (`habtm`, `has_many :through`)
    - Reloading a `habtm` or `has_many :through` association now resets the internal through association to keep the records listed on both targets in sync.
- Add `marked_for_deletion` to Autosave Associations, a mirror of `marked_for_destruction` that calls `delete` instead of `destroy`. 
    - Add `marked_for_removal?` which checks if either flag is set. 
    - If both flags are set, destruction takes precedence.
    - I'm concerned the wording of some of the documentation in [autosave_association.rb](https://github.com/rails/rails/compare/main...WelpThatWorked:rails:defer-autosave-collection?expand=1#diff-d88c50e3eec9509f3060d7847e8d6f3d82ad951b9f586698241d324eafc59cfd) might be a bit messy where the two overlap, but I'm unsure.
- Add change-tracking methods `<association>_added_records` and `<association>_previously_added_records`
    - I'm not sure this is worthwhile, as it might be more flexibly accomplished by the use of added/removed callback, and implementing similar methods for removals seems it would require keeping a copy of deleted/destroyed records in memory regardless of if the function would be used.
    - Not sure about the name.

### Additional information

Other previous related discussion:
#17368
#42097
#46685
[Rails Forum](https://discuss.rubyonrails.org/t/using-accepts-nested-attributes-for-with-assign-attributes-means-immediate-saving-of-associated-model/70525)

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add option to defer persisting added/removed records on autosave collection associations. Add `marked_for_deletion`, a mirror of `marked_for_destruction` that calls `delete` instead of `destroy`.
+
+    *WelpThatWorked*
+
 *   `:class_name` is now invalid in polymorphic `belongs_to` associations.
 
     Reason is `:class_name` does not make sense in those associations because
--- a/activerecord/lib/active_record/associations.rb
+++ b/activerecord/lib/active_record/associations.rb
@@ -184,6 +184,8 @@ module ActiveRecord
       #   others.distinct                   |   X   |    X     |    X
       #   others.reset                      |   X   |    X     |    X
       #   others.reload                     |   X   |    X     |    X
+      #   others_added_records              |   X   |    X     |    X
+      #   others_previously_added_records   |   X   |    X     |    X
       #
       # === Overriding generated methods
       #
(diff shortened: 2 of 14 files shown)


scriptforfivem:
Great implementation of this feature. I particularly like the approach to error handling.

eyestoreoye:
The performance optimizations here are quite clever. Nice engineering work!

PR #55047, opened 2025-05-11
Highlight current index link in Guides.

ahmedsayedabdelsalam:
### Motivation / Background

When reading guides section by section user confused where to go next and what is the next guide.

### Detail

This Pull Request changes the the guides-section anchors to add `.active` class if `@page == document['url']` also add some styles for the .active link

### Additional information

image

### Checklist

### DIFF
--- a/guides/assets/stylesrc/_main.scss
+++ b/guides/assets/stylesrc/_main.scss
@@ -792,6 +792,50 @@ body.guide {
             color: $rf-brand !important;
             text-decoration: underline !important;
           }
+
+          a.active {
+            position: relative;
+            display: block;
+
+            &::before { // red square
+              @include media('>desktop') {
+                content: '';
+                display: block;
+              }
+              position: absolute;
+              height: 8px;
+              width: 8px;
+              top: 4px;
+              left: -1.1em;
+              border-radius: 3px;
+              background-color: $rf-brand;
+
+              :where(html[dir="rtl"]) & {
+                right: -1.1em;
+                left: unset;
+              }
+            }
+
+            &::after { // highlight
+              content: '';
+              position: absolute;
+              display: block;
+              height: calc(100% + 10px);
+              @include media('>desktop') {
+                width: 100%;
+              }
+              top: 50%;
+              transform: translateY(-50%);
+              left: -1.5em;
+              border-radius: 8px;
+              background-color: $stop-bkgnd;
+
+              :where(html[dir="rtl"]) & {
+                right: -1.5em;
+                left: unset;
+              }
+            }
+          }
         }
       }
     }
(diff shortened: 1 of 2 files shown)


guides/assets/stylesrc/_main.scss
```diff
@@ -792,6 +792,50 @@ body.guide {
             color: $rf-brand !important;
             text-decoration: underline !important;
           }
+
+          a.active {
+            position: relative;
```
adityapandit17, line 797: please arrange them alphabetically

guides/source/layout.html.erb
```diff
@@ -86,7 +86,7 @@
                   <div class="guides-section">
                     <dt><%= section['name'] %></dt>
                     <% finished_documents(section['documents']).each do |document| %>
-                    <dd><a href="<%= document['url'] %>"><%= document['name'] %></a></dd>
+                    <dd><a href="<%= document['url'] %>" <% if @path == document['url'] %> class="active" <% end %>><%= document['name'] %></a></dd>
```
adityapandit17, line 89: Could be a better way for writing this 
```suggestion
                    <dd><a href="<%= document['url'] %>" class="<%= 'active' if @path == document['url'] %>"><%= document['name'] %></a></dd>
```

ahmedsayedabdelsalam: this will leave class="" for non active links

guides/assets/stylesrc/_main.scss
ahmedsayedabdelsalam: i copied the same style used for sidebar navigation section and changed some values.

https://github.com/rails/rails/blob/e43991b3ea26186633b74c1ca03396926d07c613/guides/assets/stylesrc/_main.scss#L1013

guides/source/layout.html.erb
MatheusRich: thoughts on using [aria-current="page"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) instead of a class?

ahmedsayedabdelsalam: what is the advantages over the class?

MatheusRich: It helps with accessibility, because it tells assistive technologies (like screen readers) which link represents the current page.

It's more semantically correct than a one-off `.current` or `.active` class. The implementation should stay mostly the same.

ahmedsayedabdelsalam: thank you for the details. i agree with you
https://github.com/ahmedsayedabdelsalam/rails/pull/1
if you approve it i'll merge it with this branch

PR #55053, opened 2025-05-12
Move ActionDispatch::MiddlewareStack to activesupport

dersam:
### Motivation / Background

In `ActiveSupport::ErrorReporter`, we've implemented a middleware pattern for modifying the error context loosely based on Rack. While using it in our app, it's become clear we need more granular control over the ordering of middlewares.

Specifically, we want to be able to insert middlewares before/after other middleware so dependencies are clearly declared. 

I started looking at how to implement this when I realized Rails already has this as part of `ActionDispatch::MiddlewareStack`. This class has no dependencies on anything else in `actionpack`, and the error reporter could simply reuse it instead of reimplementing the same set of features (as could other gems that implement Rack-like middleware patterns).

If the move of `ActionDispatch::MiddlewareStack` is accepted, `ActiveSupport::ErrorReporter` will need to be updated in a follow up to conform to the same API (which I will also build). `ErrorReporter` currently expects to receive instances of middleware to use directly, unlike the `ActionDispatch::MiddlewareStack` which expects a class and to instantiate on demand. The error context middleware is currently only on edge rails, so this is a good time to modify it to avoid backwards compatibility issues.

### Detail

This pull request only moves `ActionDispatch::MiddlewareStack` to `activesupport`. The class in `action_dispatch` remains for backwards compatibility, and to provide instrumentation.

### Checklist


byroot:
> `ActiveSupport::ErrorReporter` will need to be updated in a follow up to conform to the same API (which I will also build).

I'd rather not. I really don't like the mutable middleware interface of Rack (it's a big footgun) and I specifically didn't want the error reporter middleware stack to be like that.

dersam:
> > `ActiveSupport::ErrorReporter` will need to be updated in a follow up to conform to the same API (which I will also build).
> 
> I'd rather not. I really don't like the mutable middleware interface of Rack (it's a big footgun) and I specifically didn't want the error reporter middleware stack to be like that.

That's fine with me - we did some hacking to try out what it would be like in the ErrorReporter and it's definitely a more difficult interface (or requires substantial wrapping to maintain the same interface the ErrorReporter has today).

I do think we need additional control over where middleware are on the stack, as that makes it easier for gems to define known dependencies and for the Rails app to reorganize them if necessary. This is a scenario we're already encountering in our apps, as we have a gem that defines common error handling that sometimes needs to be partially overridden or supplemented by the app using it.

What do you think of adding a similar set of methods as `ActionDispatch::MiddlewareStack`, but implemented specifically for the `ErrorReporter` with nothing shared from the ActionDispatch/Rack version? I'm thinking we would want the following actions available:

- `insert_before/after`
- `swap`
- `delete`
- `move`

PR #55055, opened 2025-05-12
Add config option to report `rescue_responses`

dari-us:
### Motivation / Background

This Pull Request has been created because `Rails.error.subscribe` does not report `rescue_responses`. It adds `config.action_dispatch.report_rescued_responses` config option to enable it.

### Detail

This Pull Request changes middleware `ActionDispatch::ShowExceptions` to set header `action_dispatch.report_exception` to `!wrapper.rescue_response? || request.get_header("action_dispatch.report_rescue_responses")`.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/show_exceptions.rb
+++ b/actionpack/lib/action_dispatch/middleware/show_exceptions.rb
@@ -35,7 +35,7 @@ module ActionDispatch
       backtrace_cleaner = request.get_header("action_dispatch.backtrace_cleaner")
       wrapper = ExceptionWrapper.new(backtrace_cleaner, exception)
       request.set_header "action_dispatch.exception", wrapper.unwrapped_exception
-      request.set_header "action_dispatch.report_exception", !wrapper.rescue_response?
+      request.set_header "action_dispatch.report_exception", !wrapper.rescue_response? || request.get_header("action_dispatch.report_rescued_responses")
 
       if wrapper.show?(request)
         render_exception(request.dup, wrapper)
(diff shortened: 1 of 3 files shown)


byroot:
I really don't think you want this. It would mean every 404 and other client errors would be reported.

dari-us:
I really do want this. Up until now, I was using a custom middleware that does exactly that.
All I want to do is migrate from using a custom middleware to `Rails.error`, but I can't because these kinds of errors get swallowed.
To clarify, I am not using this for public apps, it is used for IP-restricted environments.

byroot:
> I really do want this. Up until now, I was using a custom middleware that does exactly that.

Then you could modify your middleware to report the error to `Rails.error`, Rails will take care of not reporting these exceptions twice.

I know it's an extra middleware, but it will just be 5 lines of code or so, which I think isn't too bad compared to adding a feature in Rails that very very few people will want, and bonus point you get to use it right now instead of whenever you get to upgrade to Rails 8.1.

dari-us:
Ok, I see your point there and the solution is fine (is it save to remove the `ActionDispatch::Executor` middleware then?), however, the fact it does not report these errors took me by surprise, because it is not explicitly stated in the docs.

https://guides.rubyonrails.org/error_reporting.html

> Rails wraps all executions (such as HTTP requests, [jobs](https://guides.rubyonrails.org/active_job_basics.html), and [rails runner](https://guides.rubyonrails.org/command_line.html#bin-rails-runner) invocations) in the error reporter, so any unhandled errors raised in your app will automatically be reported to your error-reporting service via their subscribers.

Maybe add something like this to clarify: "for HTTP requests, errors present in `ActionDispatch::ExceptionWrapper.rescue_responses` are not reported."

byroot:
> is it save to remove the `ActionDispatch::Executor` middleware then?

No, it serves many other purposes. 

>  the fact it does not report these errors took me by surprise, because it is not explicitly stated in the docs

Most existing error reporting libraries don't either. In Rails it is assumed that not all uncaught exceptions are server errors, some are directly translated into client errors, that is the purpose `rescue_from` serves, declaring that a specific exception isn't a server error.

The error reporter purpose is to report internal server errors (500).

> Maybe add something like this to clarify

Sure PR welcome.

dari-us:
Ok, thanks for weighing in, closing this as I agree most people don't need this.

PR #55058, opened 2025-05-13
Drop vendored Trix files in favor of the `action_text-trix` gem

### LINKED ISSUE #54148: Extract or remove Trix
Once in a while Trix may have [security issues](https://github.com/basecamp/trix/security/advisories/GHSA-qjqp-xr96-cj99) and require a release.  Since Rails embeds Trix, we have to do a release as well.

Can we remove Trix and let people install it via source maps, or extract Trix to a gem so that when there is a security issue we don't need to release all of Rails?

There might be another solution as well, but I'm not sure what.

flavorjones:
### Motivation / Background

Currently Trix is vendored in Rails, meaning that when bug fixes or security patches are released, application developers cannot easily update to the latest Trix without waiting for a Rails release.

This situation was originally described in #54148 after a Rails security team discussion.

### Detail

Drop vendored Trix files in favor of the [`action_text-trix` gem](https://rubygems.org/gems/action_text-trix), which itself vendors just the Trix assets. This gem will be maintained by the Trix maintainers at 37signals.

Closes #54148

### Checklist


ghiculescu:
Just to confirm - if you're installing trix via the npm/yarn package, this will have no affect, right? It's just if you were using the vendored library previously.

flavorjones:
@ghiculescu That's correct. There should be no functional change here, it's just an extraction.

PR #53666, opened 2024-11-18
Allow per-database schema format

tsvallender:
### Motivation / Background

This PR allows applications with multiple databases to set a `schema_format` independently for each one.

I've personally found a desire for this when converting a legacy application to use the standard Rails methods one piece at a time, but it would also be useful when there is a need to move a database to use `structure.sql` (for some specific functionality) but no desire to do this for _all_ databases.

### Detail

The `schema_format` can now be set to `ruby` or `sql` in the database configuration. If this does not exist, there is no change in behaviour and the application-wide configuration is used, or the environment variable if present and appropriate.

### Checklist

Before submitting the PR make sure the following are checked:

* [X] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [X] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [X] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55059, opened 2025-05-13
Respect SCHEMA_FORMAT in db:schema:load

schmijos:
Considering how we go forward in Rails >= 8 (see #53666), this has been an oversight when introducing the `SCHEMA_FORMAT` environment variable. This bugfix now respects it as described in the task description and should only be applied to Rails 7.2.

### DIFF
--- a/activerecord/lib/active_record/railties/databases.rake
+++ b/activerecord/lib/active_record/railties/databases.rake
@@ -473,7 +473,8 @@ db_namespace = namespace :db do
 
     desc "Load a database schema file (either db/schema.rb or db/structure.sql, depending on `ENV['SCHEMA_FORMAT']` or `config.active_record.schema_format`) into the database"
     task load: [:load_config, :check_protected_environments] do
-      ActiveRecord::Tasks::DatabaseTasks.load_schema_current(ActiveRecord.schema_format, ENV["SCHEMA"])
+      schema_format = ENV.fetch("SCHEMA_FORMAT", ActiveRecord.schema_format).to_sym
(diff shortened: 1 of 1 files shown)


salzig:
btw, this bug is also present in 8.1.0.beta1 https://github.com/rails/rails/blob/v8.1.0.beta1/activerecord/lib/active_record/railties/databases.rake#L469, in 8.0.2.1 https://github.com/rails/rails/blob/v8.0.2.1/activerecord/lib/active_record/railties/databases.rake#L461, and annoyingly in 7.2.2.2 https://github.com/rails/rails/blob/v7.2.2.2/activerecord/lib/active_record/railties/databases.rake#L476

skipkayhil:
> btw, this bug is also present in 8.1.0.beta1 https://github.com/rails/rails/blob/v8.1.0.beta1/activerecord/lib/active_record/railties/databases.rake#L469, in 8.0.2.1 https://github.com/rails/rails/blob/v8.0.2.1/activerecord/lib/active_record/railties/databases.rake#L461, and annoyingly in 7.2.2.2 https://github.com/rails/rails/blob/v7.2.2.2/activerecord/lib/active_record/railties/databases.rake#L476

Can you open a PR to fix it?

salzig:
@skipkayhil Rails 7.2 just needs a new release, as it was fixed in 7-2-stable, but not released yet (?).

skipkayhil:
> @skipkayhil Rails 7.2 just needs a new release, as it was fixed in 7-2-stable, but not released yet (?).

Sure, but you mentioned other versions that don't have the fix. You can also point your app at the `7-2-stable` branch in the meantime

salzig:
@skipkayhil Like this? https://github.com/rails/rails/pull/55698

PR #55060, opened 2025-05-13
Add `affected_rows` to `ActiveRecord::Result`

jenshenny:
### Motivation / Background

For database adapters that do not support insert `RETURNING`, the result for `insert_all` and `upsert_all` does not provide any information on the rows that has been affected. Adding `affected_rows` to the result allows this information to be accessed similar to the return value of other bulk operations (like `delete_all` and `update_all`).

### Detail

I added an `affected_rows` reader for the result. Database specific adapters would provide the value as a keyword arg when the result is initialized (often in `cast_result`).

### Additional information

One note to point out is that different adapters have different behaviour for affected rows in a `SELECT` statement.

#### MySQL `@affected_rows_before_warnings` / Trilogy `result.affected_rows` 
 `nil`  eg. https://github.com/trilogy-libraries/trilogy/commit/ddbefb06e5c6bd8c15b0b02edc2a8a2a309b1a76

#### Postgres `result.cmd_tuples` 
Number of rows returned in the `SELECT` https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-PQCMDTUPLES

#### SQLite `raw_connection.changes` 
Number of rows affected by the last `UPDATE`/`DELETE`/`INSERT` https://www.sqlite.org/c3ref/changes.html

I think MySQL / Postgres is fine since the value relates to the query itself but exposing this value for SQLite might not be the best idea. Though the same `affected_rows` is shared in `sql.active_record` notifications. https://github.com/rails/rails/commit/a029da566f0b9716d90aa670edf04f90b8d72b89

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
@@ -88,35 +88,31 @@ module ActiveRecord
 
             if batch
               raw_connection.execute_batch2(sql)
-            elsif prepare
-              stmt = @statements[sql] ||= raw_connection.prepare(sql)
-              stmt.reset!
-              stmt.bind_params(type_casted_binds)
-
-              result = if stmt.column_count.zero? # No return
-                stmt.step
-                ActiveRecord::Result.empty
-              else
-                ActiveRecord::Result.new(stmt.columns, stmt.to_a, stmt.types.map { |t| type_map.lookup(t) })
-              end
             else
-              # Don't cache statements if they are not prepared.
-              stmt = raw_connection.prepare(sql)
+              stmt = if prepare
+                @statements[sql] ||= raw_connection.prepare(sql)
+                @statements[sql].reset!
+              else
+                # Don't cache statements if they are not prepared.
+                raw_connection.prepare(sql)
+              end
               begin
                 unless binds.nil? || binds.empty?
                   stmt.bind_params(type_casted_binds)
                 end
                 result = if stmt.column_count.zero? # No return
                   stmt.step
-                  ActiveRecord::Result.empty
+                  @affected_rows = raw_connection.total_changes - total_changes_before_query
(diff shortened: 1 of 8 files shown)


activerecord/lib/active_record/result.rb
```diff
@@ -89,17 +89,17 @@ def to_h
       alias_method :to_hash, :to_h
     end
 
-    attr_reader :columns, :rows
+    attr_reader :columns, :rows, :affected_rows
```
eileencodes, line 92: The documentation at the top of this file needs to be updated to include affected_rows.

jenshenny: Added some docs for `affected_rows` 👍

activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
jenshenny, line 1: I rebased off https://github.com/rails/rails/pull/55065 and refactored `perform_query` here to limit the number of `ActiveRecord::Result.new` and `raw_connection.total_changes - total_changes_before_query`.

activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
```diff
@@ -88,35 +88,31 @@ def perform_query(raw_connection, sql, binds, type_casted_binds, prepare:, notif
...
+                  ActiveRecord::Result.empty(affected_rows: @affected_rows)
                 else
-                  ActiveRecord::Result.new(stmt.columns, stmt.to_a, stmt.types.map { |t| type_map.lookup(t) })
+                  rows = stmt.to_a
+                  @affected_rows = raw_connection.total_changes - total_changes_before_query
+                  ActiveRecord::Result.new(stmt.columns, rows, stmt.types.map { |t| type_map.lookup(t) }, affected_rows: @affected_rows)
                 end
               ensure
-                stmt.close
+                stmt.close unless prepare
```
zzak, line 113: :see_no_evil: Everytime this code changes

PR #55065, opened 2025-05-14
Fix `affected_rows` for SQLite `sql.active_record` notifications

jenshenny:
### Motivation / Background
From https://github.com/rails/rails/pull/55060, it was noticed that `SQLite3::Database.changes` only reflects the number of rows affected by the last `DELETE`/`UPDATE`/`INSERT`, and the number would not change for any other query ([ref](https://www.sqlite.org/c3ref/changes.html)). 

> Executing any other type of SQL statement does not modify the value returned by these functions.

Therefore, `SELECT` statements would have the affected row count of a previous query for `sql.active_record` notifications.

### Detail

@skipkayhil suggested that instead of using `SQLite3::Database.changes`, we should calculate the `affected_rows` by subtracting the `SQLite3::Database.total_changes` before the query was executed from afterwards. This would result `affected_rows` = 0 for `SELECT` statements now.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3/database_statements.rb
@@ -84,6 +84,8 @@ module ActiveRecord
           end
 
           def perform_query(raw_connection, sql, binds, type_casted_binds, prepare:, notification_payload:, batch: false)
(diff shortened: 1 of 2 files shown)


skipkayhil (approved):
Test failures look unrelated

zzak:
Clever!

I guess we don't care if the count only comes from the current connection?

PR #55068, opened 2025-05-15
Correct ActiveStorage guide phrasing

tekin:
1. A tiny correction to the ActiveStorage guide: it's the attachment that is passed to `url_for`, not the blob.


Edouard-chin:
Agreed that the example is a bit confusing and it's the attachment that get passed as argument. But it is also possible to generate a permanent url by passing the blob, so maybe we can mention that both works?

Edouard-chin:
Thank you!

PR #55072, opened 2025-05-15
Provide more control for error context middleware in ErrorReporter

dersam:
### Motivation / Background

Error context middleware in `ActiveSupport::ErrorReporter` can currently only be added. This PR adds simple `insert`/`swap`/`delete` capabilities, inspired from `ActionDispatch::MiddlewareStack`. (Additional context in [this PR](https://github.com/rails/rails/pull/55053) - the features were reimplemented instead of reusing the same class to preserve the simpler middleware interface).

This will allow error context middleware to be reorganized by the Rails app, which is especially useful if the middleware is being provided from a gem Railtie.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Provide more control for error context middleware in ErrorReporter.
+
+    The error context middleware can now insert, swap, and remove middleware
+    based on integer indexes and relative to other middleware in stack. This simplifies
+    managing apps with middleware from multiple sources (gems or the app).
+
+    *Sam Schmidt*
+
 *   Add public API for `before_fork_hook` in parallel testing.
 
     Introduces a public API for calling the before fork hooks implemented by parallel testing.
--- a/activesupport/lib/active_support/error_reporter.rb
+++ b/activesupport/lib/active_support/error_reporter.rb
@@ -29,6 +29,7 @@ module ActiveSupport
     DEFAULT_RESCUE = [StandardError].freeze
 
     attr_accessor :logger, :debug_mode
+    attr_reader :context_middlewares
 
     UnexpectedError = Class.new(Exception)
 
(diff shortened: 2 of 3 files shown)


zzak:
I wonder if the API should more like the middleware stack in AD:
https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/stack.rb

Things like `index_of` vs. `assert_index` which returns the index in the error message, `insert_at` vs. `insert` (and `insert_before` is just an alias), and the arguments etc.

Consistency is an important part of API design, even these things are not directly related -- I almost wonder if ActiveSupport should provide the backing for this API (e.g. `AS::Stack` or similar) to give confidence that when you're dealing with a [middleware] stack they behave similarly as one would expect.

dersam:
> I wonder if the API should more like the middleware stack in AD: https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/stack.rb
> 
> Things like `index_of` vs. `assert_index` which returns the index in the error message, `insert_at` vs. `insert` (and `insert_before` is just an alias), and the arguments etc.
> 
> Consistency is an important part of API design, even these things are not directly related -- I almost wonder if ActiveSupport should provide the backing for this API (e.g. `AS::Stack` or similar) to give confidence that when you're dealing with a [middleware] stack they behave similarly as one would expect.

Based on @byroot's [comment here](https://github.com/rails/rails/pull/55053#issuecomment-2875352043), the `ActiveSupport::ErrorReporter` middleware stack shouldn't behave like the one in `ActionDispatch`. They're two different approaches to middleware for different situations.

In that other PR, I moved the middleware stack into AS in preparation to make both AD and ErrorReporter use the same base class. However, this ends up complicating the ErrorReporter's stack: AD receives class names and instantiates them on demand while currying them into each other. The ErrorReporter receives instances and stores them in a simple array, which it executes in order, passing the context hash to each one.

I agree with matching the method names and aliases to aid understanding, but having them inherit from the same parent doesn't make sense to me given the design goals with ErrorReporter's version.

PR #55074, opened 2025-05-16
Preserve enum methods when using becomes! on STI subclasses

### LINKED ISSUE #55001: Using `become!` on an STI model sets enum attribute to nil

Yuhi-Sato:
### Motivation / Background

Fixes [rails#55001](https://github.com/rails/rails/issues/55001)

The `becomes!` method fails to preserve enum-based behavior when converting between STI (Single Table Inheritance) subclasses that define different `enum` values for the same attribute.

This is particularly problematic because `enum` defines predicate (`#published?`) and mutator (`#published!`) methods specific to each class. When switching classes via `becomes!`, the internal type mapping (`EnumType`) from the original class is retained, leading to incorrect validation and method behavior.

### Detail

This change ensures that when performing `becomes!`, each attribute is rebound to the destination class’s attribute type using `with_type`. This guarantees that enum accessors work correctly in the destination class context.

A test case using `LiveParrot` and `DeadParrot` subclasses of `Parrot` (each declaring their own enum `:breed`) demonstrates that the fix allows safe conversion between STI subclasses while maintaining proper enum behavior.

Before this change:

```ruby
    parrot = LiveParrot.create!(name: "Scipio", breed: "african")
    dead_parrot = parrot.becomes!(DeadParrot)
    dead_parrot.asian! # => raises ArgumentError
```

This failed due to the original enum mapping from `LiveParrot` being retained.

### Additional information

- The error is raised by `EnumType#assert_valid_value` when the new enum value is not recognized by the old enum mapping.
- This fix ensures consistency with developer expectations when using enums in STI environments.
- To avoid decryption errors for encrypted attributes
  (`ActiveRecord::Encryption::EncryptedAttributeType`), such attributes are
  explicitly skipped when rebinding types.
- This PR is a re-submission of [#55018](https://github.com/rails/rails/pull/55018), which was automatically closed after force-pushing to clean up the commit history. The content is unchanged; only the history was rewritten for clarity.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/persistence.rb
+++ b/activerecord/lib/active_record/persistence.rb
@@ -495,6 +495,14 @@ module ActiveRecord
         becoming.instance_variable_set(:@previously_new_record, previously_new_record?)
         becoming.instance_variable_set(:@destroyed, destroyed?)
         becoming.errors.copy!(errors)
+
+        klass.attribute_types.each do |name, type|
+          next if type.is_a?(ActiveRecord::Encryption::EncryptedAttributeType)
+
+          if (attribute = @attributes[name])
+            @attributes[name] = attribute.with_type(type)
+          end
+        end
       end
 
       became
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,30 @@
+*   Fix `becomes!` to correctly rebind enum attributes when converting
+    between STI subclasses that define different enum mappings for the same column.
+
+    Previously, enum methods such as `#published!` or `#published?` could raise
+    `ArgumentError` or return incorrect values after calling `becomes!`,
+     because the enum type was not properly updated to the new subclass.
+
+    Example:
+
+    ```ruby
+    class Foo < ActiveRecord::Base; end
+
+    class Bar < Foo
+      enum :state, { draft: "draft" }
+    end
+
+    class Baz < Foo
+      enum :state, { published: "published" }
+    end
+
+    bar = Bar.create!(state: "draft")
+    baz = bar.becomes!(Baz)
+    baz.published! # Previously: raises ArgumentError
+    ```
+
+    *Yuhi Sato*
+
 *   `:class_name` is now invalid in polymorphic `belongs_to` associations.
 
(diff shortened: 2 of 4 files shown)


activerecord/lib/active_record/persistence.rb
```diff
@@ -495,6 +495,14 @@ def becomes(klass)
         becoming.instance_variable_set(:@previously_new_record, previously_new_record?)
         becoming.instance_variable_set(:@destroyed, destroyed?)
         becoming.errors.copy!(errors)
+
+        klass.attribute_types.each do |name, type|
+          next if type.is_a?(ActiveRecord::Encryption::EncryptedAttributeType)
```
zzak, line 500: Can you add a test for this?

Yuhi-Sato: @zzak 
I added a test for `ActiveRecord::Encryption::EncryptedAttributeType`.

PR #55081, opened 2025-05-19
[ci skip] Clarify ActiveRecord callbacks guide

kirosmonad:
### Motivation / Background

This Pull Request has been created because while I was reviewing the docs I noticed that this piece of the docs could be improved to be clearer and easier to follow.

### Detail

This Pull Request changes slightly adjust the sentence without changing the intended message. I believe is easier to follow if we make clear that `after_destroy` is added to `Articles` and `User` deletion will cascade to triggering the callback.

### Checklist

### DIFF
--- a/guides/source/active_record_callbacks.md
+++ b/guides/source/active_record_callbacks.md
@@ -1042,8 +1042,8 @@ through the model associations whereby life cycle events can cascade on
 associations and fire callbacks.
 
 Suppose an example where a user has many articles. A user's articles should be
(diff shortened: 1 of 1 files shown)


Edouard-chin:
Thank you for the Pull Request. I find the correct documentation comprehensible and the new changes rather confusing. I read `Let's trigger [...]` as "having to manually trigger something", whereas the intent is to define a piece of code and let Rails automatically call it whenever needed.

Going to close as I don't think it improves the documentation

PR #55082, opened 2025-05-19
Add  jsonb support for SQLite.

dorzey:
### Motivation / Background

This Pull Request has been created because I want to be able to take advantage of `jsonb` support in postgres, but be able to use sqlite when testing

### Detail

This Pull Request changes the ActiveRecord connection adapter for sqlite

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
@@ -106,6 +106,7 @@ module ActiveRecord
         binary:       { name: "blob" },
         boolean:      { name: "boolean" },
         json:         { name: "json" },
+        jsonb:        { name: "json" },
       }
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
```diff
@@ -106,6 +106,7 @@ def native_database_types # :nodoc:
         binary:       { name: "blob" },
         boolean:      { name: "boolean" },
         json:         { name: "json" },
+        jsonb:        { name: "json" },
```
skipkayhil, line 109: Reading through https://github.com/sparklemotion/sqlite3-ruby/discussions/577, it seems like the correct column type for "jsonb" would be "blob", otherwise we're just aliasing `jsonb` to regular `json` right?

PR #55085, opened 2025-05-20
Avoid a module named `Namespace` for Ruby 3.5

Earlopain:
This will be a feature in Ruby 3.5 and fails:
> Namespace is not a module (TypeError)

Other places in rails already use `Namespaced` for this purpose.

Also reported upstream: https://bugs.ruby-lang.org/issues/21341

cc @yahonda


yahonda:
While this change is easy for Rails point of view because we are using this `module Namespace` inside the test only. However, I'd like to see if https://bugs.ruby-lang.org/issues/21341 is answered by Ruby committers.

PR #55086, opened 2025-05-20
Fix `remove_foreign_key` when multiple columns are found

### LINKED ISSUE #55078: `remove_foreign_key` removes an unexpected key when there is another key for the same column.
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  create_table :users do |t|
    t.timestamps
  end

  create_table :books do |t|
    t.references :user, foreign_key: true, type: :bigint
    t.references :author, foreign_key: { to_table: :users }, type: :bigint
    t.timestamps
  end
end


ActiveSupport::TestCase.test_order = :sorted
  # To check the table definition
  def test_definition
    definition = ActiveRecord::Base.connection.execute("select * from sqlite_master;").detect {|obj| obj['tbl_name'] == 'books' }['sql']

    assert_equal <<~SQL.strip, definition
      CREATE TABLE "books" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "user_id" bigint, "author_id" bigint, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL, CONSTRAINT "fk_rails_bc582ddd02"
      FOREIGN KEY ("user_id")
        REFERENCES "users" ("id")
      , CONSTRAINT "fk_rails_53d51ce16a"
      FOREIGN KEY ("author_id")
        REFERENCES "users" ("id")
      )
    SQL
  end

  def test_remove_foreign_key
      # If you replace the following two lines, the test will pass.
      remove_foreign_key :books, :users
      remove_foreign_key :books, :users, column: :author_id
    end

    definition = ActiveRecord::Base.connection.execute("select * from sqlite_master;").detect {|obj| obj['tbl_name'] == 'books' }['sql']

    assert_equal <<~SQL.strip, definition
      CREATE TABLE "books" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "user_id" bigint, "author_id" bigint, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL)
    SQL
  end
end
```

### Expected behavior

It removes each foreign keys.

### Actual behavior

It raises the error.
And it removed foreign key for `author_id` in `remove_foreign_key :books, :users`.

```
-- remove_foreign_key(:books, :users, {column: :author_id})
E

Finished in 0.034989s, 57.1608 runs/s, 28.5804 assertions/s.

  1) Error:
BugTest#test_remove_foreign_key:
ArgumentError: Table 'books' has no foreign key for users
```

### System configuration

**Rails version**: 
8.0.2

**Ruby version**: 
3.4.1

zzak:
Fixes #55078

This PR has been updated to fix the behavior using 8.1 migration for backwards compatibilty.

/cc @kei-s

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -1815,7 +1815,20 @@ module ActiveRecord
 
         def foreign_key_for(from_table, **options)
           return unless use_foreign_keys?
-          foreign_keys(from_table).detect { |fk| fk.defined_for?(**options) }
+
+          keys = foreign_keys(from_table)
+
+          if options[:_skip_column_match]
+            return keys.find { |fk| fk.defined_for?(**options) }
+          end
+
+          if options[:column].nil?
+            default_column = foreign_key_column_for(options[:to_table], "id")
+            matches = keys.select { |fk| fk.column == default_column }
+            keys = matches if matches.any?
+          end
+
+          keys.find { |fk| fk.defined_for?(**options) }
         end
 
         def foreign_key_for!(from_table, to_table: nil, **options)
--- a/activerecord/test/cases/migration/compatibility_test.rb
+++ b/activerecord/test/cases/migration/compatibility_test.rb
@@ -664,6 +664,60 @@ module ActiveRecord
         end
       end
 
+      def test_remove_foreign_key_on_8_1
+        connection.create_table(:sub_testings) do |t|
+          t.references :testing, foreign_key: true, type: :bigint
+          t.references :experiment, foreign_key: { to_table: :testings }, type: :bigint
+        end
+
+        migration = Class.new(ActiveRecord::Migration[8.1]) do
+          def up
+            change_table(:sub_testings) do |t|
+              t.remove_foreign_key :testings
+              t.remove_foreign_key :testings, column: :experiment_id
+            end
+          end
+        end
+
+        ActiveRecord::Migrator.new(:up, [migration], @schema_migration, @internal_metadata).migrate
+
+        foreign_keys = @connection.foreign_keys("sub_testings")
+        assert_equal 0, foreign_keys.size
+      ensure
+        connection.drop_table(:sub_testings, if_exists: true)
+        ActiveRecord::Base.clear_cache!
+      end
+
+      def test_remove_foreign_key_on_8_0
+        connection.create_table(:sub_testings) do |t|
+          t.references :testing, foreign_key: true, type: :bigint
+          t.references :experiment, foreign_key: { to_table: :testings }, type: :bigint
+        end
+
+        migration = Class.new(ActiveRecord::Migration[8.0]) do
+          def up
+            change_table(:sub_testings) do |t|
+              t.remove_foreign_key :testings
+              t.remove_foreign_key :testings, column: :experiment_id
+            end
+          end
+        end
+
+        assert_raise(StandardError, match: /Table 'sub_testings' has no foreign key for testings$/) {
+          ActiveRecord::Migrator.new(:up, [migration], @schema_migration, @internal_metadata).migrate
+        }
+
+        foreign_keys = @connection.foreign_keys("sub_testings")
+        if current_adapter?(:PostgreSQLAdapter, :SQLite3Adapter)
+          assert_equal 2, foreign_keys.size
+        else
+          assert_equal 1, foreign_keys.size
+        end
+      ensure
+        connection.drop_table(:sub_testings, if_exists: true)
+        ActiveRecord::Base.clear_cache!
+      end
+
       private
         def precision_implicit_default
           if current_adapter?(:Mysql2Adapter, :TrilogyAdapter)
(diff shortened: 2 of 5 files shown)


fatkodima:
I wonder if we should make older migrations backwards compatible and how likely is the following case, because previously this worked (tables are from the issue)
```ruby
remove_foreign_key :books, :users # removes fkey on author_id
remove_foreign_key :books, :users, column: :user_id
```

and now will not work
```ruby
remove_foreign_key :books, :users # removes fkey on user_id
remove_foreign_key :books, :users, column: :user_id # raises because there is no foreign key
```

zzak:
@fatkodima Thanks for taking a look!

> remove_foreign_key :books, :users # removes fkey on author_id

I feel this is a bug and it should raise, but also that is kind of a breaking change.
My idea for `Compatibility` is that it's there for API changes to migration methods, or handling new options in a backwards compatible way, does that align with your understanding?

activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
```diff
@@ -1815,7 +1815,23 @@ def foreign_key_name(table_name, options)
...
+          if options[:column]
+            expected_columns = Array(options[:column]).map(&:to_s)
+            keys.select! { |fk| Array(fk.options[:column]).map(&:to_s) == expected_columns }
+          elsif options[:to_table] && !options.key?(:primary_key) && keys.size > 1
+            default_column = foreign_key_column_for(options[:to_table], "id").to_s
+            match = keys.find { |fk| Array(fk.options[:column]).map(&:to_s) == [default_column] }
+            keys = [match].compact if match
+          end
+
+          keys.detect { |fk| fk.defined_for?(**options) }
```
fatkodima, line 1834: Haven't tested, but something like this is shorter and should work too
```suggestion
          keys = foreign_keys(from_table)
          
          if keys.size == 1 || options[:_skip_column_match]
            keys.find { |fk| fk.defined_for?(**options) }
          elsif options[:column]
            expected_columns = Array(options[:column]).map(&:to_s)
            keys = keys.select { |fk| Array(fk.column) == expected_columns }
            keys.find { |fk| fk.defined_for?(**options) }
          else
            default_column = foreign_key_column_for(options[:to_table], "id")
            keys.find { |fk| fk.column == default_column }
          end
```

zzak: That works, thanks for tidying it up!
Fixed in 02bcd2e

fatkodima (approved):

zzak:
@fatkodima I think this is ready for another review. Couple changes:

* Reverted the change to sqlite3, test coverage was a concern
* Added TableDefinition with tests (for change_table,etc apis)
* Left backwards compat path at the top to separate new logic
* Simplified branching instead of checking size for composite keys

We can try to circle back and fix sqlite, but I didn't want to hold this up anymore.

Thanks!

activerecord/test/cases/migration/compatibility_test.rb
```diff
@@ -664,6 +664,166 @@ def up
         end
       end
 
+      if current_adapter?(:SQLite3Adapter)
+        def test_remove_foreign_key_on_8_1
```
fatkodima: This test here (and in other places) looks for me the same as the following (except `def migrate(x)`), I would leave only one of them?

activerecord/test/cases/migration/compatibility_test.rb
```diff
@@ -664,6 +664,166 @@ def up
         end
       end
 
+      if current_adapter?(:SQLite3Adapter)
```
fatkodima: If you will fix for sqlite too, I would suggest instead of these `if`s and removing blocks of code in the future PRs, and calling `skip if sqlite adapter` inline in tests. Then only a single line will be removed.

fatkodima (commented):
I didn't get why you decided to not fix sqlite in this PR? I think it will require only a few additional test cases and the PR will be of the same size as currently?

If you don't have time etc, let me know and I will try to implement it in this PR.

activerecord/test/cases/migration/foreign_key_test.rb
```diff
@@ -771,6 +787,38 @@ def test_add_foreign_key_with_suffix
           ActiveRecord::Base.table_name_suffix = nil
         end
 
+        def test_remove_foreign_key_with_prefix
```
zzak, line 790: Before this change, this test would fail like this only for sqlite:

```
ActiveRecord::Migration::ForeignKeyTest#test_remove_foreign_key_with_suffix:
ArgumentError: Table 'classes_s' has no foreign key for {}
    lib/active_record/connection_adapters/sqlite3/schema_statements.rb:81:in 'ActiveRecord::ConnectionAdapters::SQLite3::SchemaStatements#remove_foreign_key'
    test/cases/migration/foreign_key_test.rb:813:in 'ActiveRecord::Migration::ForeignKeyTest#test_remove_foreign_key_with_suffix'
```

zzak:
@fatkodima Sorry had a few things come up, I've added back the sqlite part with some tests and one other tiny refactor.

I thought about possibly moving the check into here, but didn't like how it came out:
https://github.com/rails/rails/blob/5a7a59ab4a9ac34b8cadf9d8ce7b554da50c5e43/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L165-L171

fatkodima (approved):

PR #55089, opened 2025-05-20
:class_name should be invalid in polymorphic belongs_to

fxn:
## Rationale

_(Preliminary cleanup for a forthcoming patch that will allow users to declare the targets of polymorphic `belongs_to`.)_

In polymorphic `belongs_to`, you can customize `:foreign_type`, but setting `:class_name` makes no sense, since the class name of target records is determined at runtime from the value in the type column.

## The patch

We take `:class_name` out of the generic valid options in the parent class of association builders, and move it down to the subclasses that support it.

## Deprecating the option comes next

Interestingly, our own test suite had a few of these! That was doing nothing!

This is also telling us we should warn users. I'll followup with a deprecation in `8-0-stable`.

### DIFF
--- a/activerecord/test/models/cpk/comment.rb
+++ b/activerecord/test/models/cpk/comment.rb
@@ -3,7 +3,7 @@
 module Cpk
   class Comment < ActiveRecord::Base
     self.table_name = :cpk_comments
-    belongs_to :commentable, class_name: "Cpk::Post", foreign_key: %i[commentable_title commentable_author], polymorphic: true
-    belongs_to :post, class_name: "Cpk::Post", foreign_key: %i[commentable_title commentable_author]
+    belongs_to :commentable, foreign_key: %i[commentable_title commentable_author], polymorphic: true
+    belongs_to :post, foreign_key: %i[commentable_title commentable_author]
   end
 end
--- a/activerecord/lib/active_record/associations/builder/association.rb
+++ b/activerecord/lib/active_record/associations/builder/association.rb
@@ -19,7 +19,7 @@ module ActiveRecord::Associations::Builder # :nodoc:
     self.extensions = []
 
     VALID_OPTIONS = [
-      :class_name, :anonymous_class, :primary_key, :foreign_key, :dependent, :validate, :inverse_of, :strict_loading, :query_constraints
+      :anonymous_class, :primary_key, :foreign_key, :dependent, :validate, :inverse_of, :strict_loading, :query_constraints
     ].freeze # :nodoc:
 
     def self.build(model, name, scope, options, &block)
--- a/activerecord/lib/active_record/associations/builder/belongs_to.rb
+++ b/activerecord/lib/active_record/associations/builder/belongs_to.rb
@@ -8,8 +8,9 @@ module ActiveRecord::Associations::Builder # :nodoc:
 
     def self.valid_options(options)
       valid = super + [:polymorphic, :counter_cache, :optional, :default]
-      valid += [:foreign_type] if options[:polymorphic]
-      valid += [:ensuring_owner_was] if options[:dependent] == :destroy_async
+      valid << :class_name unless options[:polymorphic]
+      valid << :foreign_type if options[:polymorphic]
+      valid << :ensuring_owner_was if options[:dependent] == :destroy_async
       valid
     end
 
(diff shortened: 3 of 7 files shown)


activerecord/test/models/cpk/comment.rb
```diff
@@ -3,7 +3,7 @@
 module Cpk
   class Comment < ActiveRecord::Base
     self.table_name = :cpk_comments
-    belongs_to :commentable, class_name: "Cpk::Post", foreign_key: %i[commentable_title commentable_author], polymorphic: true
-    belongs_to :post, class_name: "Cpk::Post", foreign_key: %i[commentable_title commentable_author]
+    belongs_to :commentable, foreign_key: %i[commentable_title commentable_author], polymorphic: true
+    belongs_to :post, foreign_key: %i[commentable_title commentable_author]
```
skipkayhil, line 7: Hey @fxn, I was curious why `class_name` was removed from `belong_to :post` since it doesn't have `polymorphic: true` (I tried locally and it didn't raise or warn when present).

fxn: @skipkayhil Right! I just restored it in https://github.com/rails/rails/commit/612666219b8c44770479c366d9efc936269033ee. Thanks!

alexevanczuk:
@fxn 

Hey Xavier, long time no chat. I discovered this PR recently which was a breaking change for us.

We depend on `class_name` as a type hint for sorbet for polymorphic associations. The use case is, we have an RBI generator which, if a polymorphic association defines a class name, it uses that as a type hint to determine the signature that should be associated with that polymorphic association and the various helpers it generates.

Therefore, this test broke this functionality and we have to find a much less ergonomic way to do it (a separate method which is read and sits next to the association definition by convention).

Was it necessary technically to remove this or was this a philosophical choice based on the fact that polymorphic classes have class names determined at runtime?

fxn:
Hi @alexevanczuk!

As you see in the patch, no association code is touched. This by itself tells us the option was ignored by Active Record. And that is because it is useless for polymorphic associations. That is the point. Allowing this option was a mistake, it was an oversight in the way options are setup that allowed it to sneak in.

Most associations have a deterministic class name associated, either derivable by convention, or explicit. But polymorphic ones do not.

Polymorphic associations are inherently dynamic. They are not FK constraint-friendly. They are open by design. You define something as being taggable, say, and it is the rest of the application that decides what to tag. The tag model itself does not care about that.

One possible way to address this in Sorbet is to require that users define all taggables using `as: :taggable` associations on the other side. You collect them, and build a union type.

You need to ask users for that, because on paper there is no requirement to define such associations. In general, a `belongs_to` is not required to have a corresponding `has_many` or `has_one`. Even more, it is OK to set/read the `_type` and `_id` attributes directly, bypassing the association API.

But telling users "you need to define the corresponding associations in all taggables" is reasonable, I think. That is normally already happening, and giving Sorbet a bit of help so that it can do its job is normal anyway.

alexevanczuk:
Hey thanks @fxn for your prompt reply. I understand that you removed it because it wasn't used... mostly I'm just looking for an alternative.

Note that the idea of having the corresponding association define the type doesn't quite meet the need, because the idea is that we are defining polymorphic associations where the client can only use it if the `taggable` implements the `Taggable` sorbet interface. So having the polymorphic association from the `Tag` model dictate that "the class name must be a `Taggable`" is a way for the tag service to enforce that all users implement that interface.

I might be misunderstanding what you're saying though! Thanks for engaging with this discussion 🙏🏼 

Maybe there's another property, like `as`, we should be using instead?

fxn:
@alexevanczuk yes! You have it in the `belongs_to`. For example:

```ruby
class Picture < ApplicationRecord
  belongs_to :imageable, polymorphic: true
end

class Employee < ApplicationRecord
  has_many :pictures, as: :imageable
end
```

See the match?

daicoden:
how do we build the type signature from picture? - We were using class_name.

daicoden:
Maybe more details, we had things like this:

```
class Picture < ApplicationRecord
  belongs_to :imageable, polymorphic: true, class_name: Imageable
end

class Employee < ApplicationRecord
  include Imageable
end

class Teacher < ApplicationRecord
  include Imageable
end

module Imageable
  extend ActiveSupport::Concern

  included do
       has_many :pictures, as: Imageable
  end
end
```

fxn:
@daicoden it is midnight here, if this is blocking you, please define a decoration in `ApplicationRecord`:

```ruby
  def self.belongs_to(...)
    # delete class_name from options
    # call super
  end
end
```

until Sorbet figures this out.

daicoden:
roger that! I think that's what @alexevanczuk did - if we can do the monkey patch long term I think that will work, this is key to some dependency inversion work we are doing.

PR #55090, opened 2025-05-20
[ci skip] docs: pluck silently ignore select statement

daffo:
### Motivation / Background

based on [this test](https://github.com/rails/rails/blob/main/activerecord/test/cases/calculations_test.rb#L1277) (from [this PR](https://github.com/rails/rails/pull/8176) - fixing [this issue](https://github.com/rails/rails/issues/7551)) it is expected, but it might be obscure and possibly dangerous behavior.

QUESTION: in [pluck API doc](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/calculations.rb#L286) there's an explicit example if you happen to require advanced select statements, is it worth to add a mention there? imo no, but please advise.

### Additional information

had a presentation to ~100 people in my company and most (almost totality) were surprised/confused, this is why I'm trying to add it to the official pluck guide

### DIFF
--- a/activerecord/lib/active_record/relation/calculations.rb
+++ b/activerecord/lib/active_record/relation/calculations.rb
@@ -287,6 +287,11 @@ module ActiveRecord
     #   # SELECT DATEDIFF(updated_at, created_at) FROM people
     #   # => ['0', '27761', '173']
     #
+    # Be aware that #pluck ignores any previous select clauses
+    #
+    #   Person.select(:name).pluck(:id)
+    #   # SELECT people.id FROM people
+    #
     # See also #ids.
     def pluck(*column_names)
       if @none


guides/source/active_record_querying.md
```diff
@@ -2427,6 +2427,26 @@ with:
...
+irb> Customer.select("DISTINCT id").pluck(:id)
+SELECT "customers"."id" FROM customers
+```
+
+If you need to use raw SQL, this is how you can do it:
+
+```irb
+irb> Customer.pluck(Aler.sql("DISTINCT id"))
+SELECT DISTINCT id FROM customers
+```
```
fatkodima, line 2448: This also needs to be placed under the `#pluck` section.

```suggestion
NOTE: Be aware that `pluck` ignores any previous `select` clauses:

```irb
irb> Customer.select(:email).pluck(:id)
SELECT "customers"."id" FROM customers
```

If you need to use raw SQL, this is how you can do it:

```irb
irb> Customer.pluck(Arel.sql("DISTINCT id"))
SELECT DISTINCT id FROM customers
```
```

Comment (daffo) [reply guides/source/active_record_querying.md]: this is definitely the wrong section, my bad 😕 
https://github.com/rails/rails/pull/55090/commits/b43a9d62c9c2a07aad7c8c39987cb3a7ac19d5c2

File: guides/source/active_record_querying.md [hunk 2]
```diff
@@ -2408,6 +2408,20 @@ One way to avoid this is to `unscope` the includes:
...
+
```
skipkayhil, line 2424: One of the examples in the [API docs for pluck](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-pluck) demonstrates using `Arel.sql`. Sinces the guides are more of an entrypoint than the API docs, I'm thinking we may want to only document simpler cases in the guide and leave more complex cases in the API docs.

daffo: gotcha

do you think is worth having the note/warning directly in the API docs rather than here in the guide?
just above the Arel.sql example perhaps? or below, under all the examples section?

skipkayhil: I think my opinion would be yes, even the `select(:column).pluck(:another)` edge case probably is more suited for API docs

daffo: https://github.com/rails/rails/pull/55090/commits/c9a2a70b4e14ebc73355dd45a2e65fb43246d092

skipkayhil (approved):
Can you squash your commits into one? I'll hold off merging to hear fatkodima's opinion, but it looks good to me now

PR #51464, opened 2024-04-02
[Fix #51463] Raise an error when invalid `:only` or `:except` options are given to `#resource` or `#resources`

joshuay03:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/51463.

### Detail

Raises an `ArgumentError` when a `ActionDispatch::Routing::Mapper::Resources::Resource` is initalized with `:only` or `:except` options that aren't default (index, create, new, show, update, and destroy) actions .

### Additional information

See my comment below as to why this seems like more changes than it needs to be.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55091, opened 2025-05-20
Deprecate :class_name for polymorphic belongs_to

fxn:
This is a deprecation followup for #55089.

The option `:class_name` is invalid for polymorphic `belongs_to` in the `main` branch. This patch adds a deprecation to `8-0-stable`.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Passing `:class_name` to a polymorphic `belongs_to` is deprecated and it
+    will be removed in Rails 8.1.
+
+    The class name of associated records is stored in the type column. If you
+    have this option set, please just delete it.
+
+    *Xavier Noria*
+
 *   `create_or_find_by` will now correctly rollback a transaction.
 
     When using `create_or_find_by`, raising a ActiveRecord::Rollback error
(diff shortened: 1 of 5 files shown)


fxn:
Once a stable release is out, its patch releases should not remove deprecated behavior, and should not make breaking changes.

But they can add deprecations, and that is why the upgrading guide says to upgrade to the **last patch release**, then fix deprecations.

This is the way to be able to make a breaking change in the next minor/major. Otherwise, the deprecation cycle would be too long.

Earlopain:
From the guide:

> Patch versions only include bug fixes

The reason to upgrade to the latest patch version during upgrade to me is there because historically already existing deprecations had bugs where they were either to eager or not eager enough.

I'm very sure that if I would have proposed this PR then it would have been promptly rejected. But I understand that you are rails core, so 🤷

fxn:
@Earlopain I'll double-check with the team.

Earlopain:
How official 😆. I think your change on main doesn't even need a deprecation. https://github.com/rails/rails/pull/51464 for example was also just done without one.

fxn:
@Earlopain you were right, thanks for raising this, I'll redo the patches.

PR #50542, opened 2024-01-02
Set `lang` attribute in application layout

stevepolitodesign:
### Motivation / Background

Leverage [I18n][] to set the [lang][] attribute on the `<html>` tag in the application layout in an effort to improve document semantics.

[I18n]: https://guides.rubyonrails.org/i18n.html#the-public-i18n-api
[lang]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang

### Additional information

Per the [guides](https://guides.rubyonrails.org/i18n.html#configure-the-i18n-module):

> The I18n library will use English as a default locale, i.e. if a different locale is not set, :en will be used for looking up translations.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55101, opened 2025-05-22
feat: set html lang attribute dynamically based on locale

ibramsterdam:
### Motivation / Background
This Pull Request was created because a new Rails app does not follow the WCAG 2.1 rules. Specifically: "The default language of each web page cannot be found by a computer program". [See WCAG 2.1 - Language of Page](https://www.w3.org/TR/WCAG21/#language-of-page).

### Detail

This Pull Request changes `application.html.erb.tt` so that the `<html>` tag uses `lang="<%= I18n.locale %>"`.

### Additional information
The EU asks more and more every year that websites follow the WCAG 2.1 standard ([see more](https://digital-strategy.ec.europa.eu/policies/web-accessibility-directive-standards-and-harmonisation)). I don’t know the rules in other countries, but I know that in The Netherlands for example, government websites and most universities (maybe all) must use software that meets these rules, even if the software comes from other companies.

Including the `lang` tag also helps make the page better for SEO [according to Google](https://developers.google.com/search/docs/specialty/international/localized-versions#language). It is also a best practice according to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang#best_practices). It also helps new developers like me learn to do things the right way! :-)

### Checklist

### DIFF
--- a/guides/source/action_view_overview.md
+++ b/guides/source/action_view_overview.md
@@ -749,7 +749,7 @@ Here is an example of a simple layout in `application.html.erb` file:
 
 ```html+erb
 <!DOCTYPE html>
-<html>
+<html lang="<%= I18n.locale %>">
 <head>
   <title><%= "Your Rails App" %></title>
   <%= csrf_meta_tags %>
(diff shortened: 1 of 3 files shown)


matthewd:
https://github.com/rails/rails/pull/50542

ibramsterdam:
Whoops, I completely missed that this was a duplicate and already rejected. Thanks for taking a look though ☺️

PR #55109, opened 2025-05-23
Set cache `max-age` values consistently in configs

grantbdev:
### Motivation / Background

Set cache `max-age` values consistently in configs so there is no difference in the requirements for the environment config files and it is easier to compare different environments.

At one point it was accepted that these values should use ActiveSupport methods to make the value self-explanatory: https://github.com/rails/rails/commit/33f116f22cd46c26ecc60e7f9f9d2bd0691b9fbb

However, more recently this was reverted in the `test` environment with the rationale that it wasn't worth the ActiveSupport require: https://github.com/rails/rails/commit/44548856c4c50e23bae0d007d90bfaf95b87d1d7

This commit updates the other environment config files to match `test` as I don't see a reason why they should be different.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -1,5 +1,3 @@
-require "active_support/core_ext/integer/time"
-
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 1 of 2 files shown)


railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -18,7 +16,7 @@ Rails.application.configure do
   <%- end -%>
 
   # Cache assets for far-future expiry since they are all digest stamped.
-  config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
+  config.public_file_server.headers = { "cache-control" => "public, max-age=3600" }
```
Earlopain, line 19: This should be one year, not one hour

grantbdev: Oops, thanks!

PR #55107, opened 2025-05-23
Document through + polymorphic

fxn:
Documents that polymorphic associations are not allowed as through associations.

If you try to set a polymorphic as through, Active Record [raises](https://github.com/rails/rails/blob/6e4a09efdbb43fb965b3f9a414ee2f4877ce07f0/activerecord/lib/active_record/reflection.rb#L1149-L1155) when the reflection is validated. Those exceptions are undocumented. I think we do not need to make them public, at least without a use case. Stating that is an error condition may suffice.

PR #55111, opened 2025-05-24
Improve docs through + polymorphic

fxn:
Proposal to address feedback in https://github.com/rails/rails/pull/55107#issuecomment-2906259301.

(I am not sure if we have established notation for model + association, I picked one that seemed intuitive.)

### DIFF
--- a/guides/source/association_basics.md
+++ b/guides/source/association_basics.md
@@ -979,6 +979,9 @@ end
 allowing instances of one model (Physician) to be associated with multiple
 instances of another model (Patient) through a third "join" model (Appointment).
 
+We call `Physician.appointments` and `Appointment.patient` the _through_ and
+_source_ associations of `Physician.patients`, respectively.
+
 ![has_many :through Association
 Diagram](images/association_basics/has_many_through.png)
 
@@ -1013,7 +1016,8 @@ created with `physician_id` and `patient_id` columns, establishing the
 many-to-many relationship between `physicians` and `patients`.
 
 INFO: The through association can be any type of association, including other
-through associations, but it cannot be a polymorphic association.
+through associations, but it cannot be [polymorphic](#polymorphic-associations).
+Source associations can be polymorphic as long as you provide a source type.
 
 You could also consider using a [composite primary
 key](active_record_composite_primary_keys.html) for the join table in the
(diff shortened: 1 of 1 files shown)


vfonic:
Thanks for looking into this!

> We call `Physician.appointments` and `Appointment.patient` the _through_ and _source_ associations of `Physician.patients`, respectively.

> The through association can be any type of association, including other through associations, but it cannot be a polymorphic association.

I've been trying to come up with an example where `Physician.appointments` is polymorphic. It's `has_many :appointments`, while polymorphic association is defined on `belongs_to :x, polymorphic: true`.

The only way I see it possible is in the way I described in the other comment:
`has_many :professionals, through: :appointments # belongs_to :professional, polymorphic: true`

This is where I get: `ActiveRecord::HasManyThroughAssociationPolymorphicSourceError` as expected.

```rb
Patient => has_many :appointments
Appointment => belongs_to :professional, polymorphic: true
Patient => has_many :professionals, through: :appointments # error
```

This seems incorrect:

> We call `Physician.appointments` the _through_ association
> The through association (...) cannot be a polymorphic association.

It's the `has_many :professionals` part of `has_many :professionals, through: :appointments`, not the `through: :appointments`.

Do you think it would be useful to add an example to the docs when polymorphic is not allowed and how to use source to circumvent that?

```rb
class Patient < ApplicationRecord
  has_many :appointments
  # has_many :professionals, through: :appointments # belongs_to :professional, polymorphic: true # error
  has_many :physicians, through: :appointments, source: :professional, source_type: "Physician"
  has_many :life_coaches, through: :appointments, source: :professional, source_type: "LifeCoach"
  has_many :chefs, through: :appointments, source: :professional, source_type: "Chef"
end
```

fxn:
> I've been trying to come up with an example where Physician.appointments is polymorphic.

That is not possible, [it is an error condition in through associations](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/reflection.rb#L1149-L1155).

That is what the patch is documenting.

The through association cannot be polymorphic, and the source association can be polymorphic as long as you provide a source type.

I believe there is a vocabulary mismatch somewhere, let's see if we find out where!

fxn:
Need to do more edits in these files, I'm going to merge but we can continue the discussion.

vfonic:
Got it! I think the best way would be to have this example you posted (or similar) in the docs:

```rb
  belongs_to :taggable, polymorphic: true
  has_many :whatever, through: :taggable
end
```

PR #51308, opened 2024-03-12
Do not build View watcher until the first `updated?` check

nvasilevski:
Partially reworks https://github.com/rails/rails/pull/47347

Currently initialization of every Rails::Engine leads to the creation of a new View watcher when the engine prepends its paths.
https://github.com/rails/rails/blob/1555d9a5579b19acf7b9e14563b460b91ab83ddc/railties/lib/rails/engine.rb#L614
This contributes to the time it takes to perform the first cold request on a lazy loaded application.

`prepend_view_path` -> `_build_view_paths` -> `cast_file_system_resolvers` -> `file_system_resolver_hooks.each(&:call)` -> `rebuild_watcher` 

This change delays the initialization of the View watcher until the first `updated?` check is performed which leads to only one initialization of the watcher.

### Benchmark

It's not perfect but I believe it reliably shows the behavior and impact of unnecessary watcher rebuilds:

<details>
  <summary>Benchmark</summary>

  ```ruby
  require "bundler/inline"

  gemfile(true) do
    source "https://rubygems.org"

    gem "rails", path: "."
    gem "listen"
  end

  require "action_controller/railtie"

  class TestApp < Rails::Application
    config.root = __dir__
    config.hosts << "example.org"
    config.secret_key_base = "secret_key_base"
    config.file_watcher = ActiveSupport::EventedFileUpdateChecker

    config.logger = Logger.new($stdout)
    Rails.logger  = config.logger

    routes.draw do
      get "/" => "test#index"
    end
  end

  TestApp.initialize!

  class TestController < ActionController::Base
    include Rails.application.routes.url_helpers

    def index
      render plain: "Home"
    end
  end

  require "minitest/autorun"
  require "rack/test"

  class BugTest < Minitest::Test
    include Rack::Test::Methods

    def test_returns_success
      view_reloader = TestApp.reloaders.find { |reloader| reloader.is_a?(ActionView::CacheExpiry::ViewReloader) }
      result = Benchmark.ms do |x|
        20.times do |i|
          TestController.prepend_view_path "test#{i}"
        end
        view_reloader.updated?
      end
      puts "Time: #{result.round(2)} ms"
    end

    private
      def app
        Rails.application
      end
  end
  ```
  
</details>

Results:

`main`:  `Time: 17.98 ms`
this branch: `Time: 2.93 ms`

In our modular monolith we have about ~100 engines that prepend view paths causing the watcher to be rebuilt with a growing directories list and this change results in us cutting ~8s from the ~16s cold request   

[...]

PR #55114, opened 2025-05-24
[ActiveSupport] Fix FileUpdateChecker false positives with travel_to in test environments

### LINKED ISSUE #55112: Classes are reloaded during request specs since 7.2
It appears that due to changes in #51308, `ActiveSupport::FileUpdateChecker` incorrectly compares file update times in test environments that use `travel_to` to manipulate time. This causes the application to reload even when files have not actually changed.

This issue seems to arise from a change in the initialization timing of `ActionView::CacheExpiry::ViewReloader` introduced in #51308.

For example, in a request spec, if the first test runs at a past time and the next test runs at the current time, a reload occurs in the middle of the request. This can lead to issues like DB connection disconnection and test data rollback.

A similar issue was discussed in an rspec-rails issue (https://github.com/rspec/rspec-rails/issues/2782), which was closed by setting `enable_reloading = false`. However, our team uses Spring, so we would like to keep `enable_reloading` set to `true`.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "8.0.2"
end

require "minitest/autorun"
require "active_support/core_ext/integer/time"
require "active_support/test_case"
require "active_support/file_update_checker"
require "fileutils"
require "tmpdir"

  include ActiveSupport::Testing::TimeHelpers

  def setup
    @tmpdir = Dir.mktmpdir
  end

  def teardown
    FileUtils.remove_entry_secure @tmpdir
  end

  def test_file_update_checker
    file_to_watch = File.join(@tmpdir, "watched_file.txt")
    FileUtils.touch(file_to_watch)

    checker = nil

    travel_to 1.hour.ago do
      checker = ActiveSupport::FileUpdateChecker.new([file_to_watch]) { nil }
      # `checker(@last_update_at)` #=> 1970-01-01 00:00:00 UTC
      assert_not checker.updated?, "should return false immediately after initialization"
    end

    travel_to 1.hour.since do
      assert_not checker.updated?, "should return false when the file has not been updated"
    end
  end
end
```

### Expected behavior

The tests in the reproduction code should pass.

### Actual behavior

The second `assert_not` in the reproduction code fails.

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.4.3

drymar:
### Motivation / Background

This Pull Request fixes a bug where `ActiveSupport::FileUpdateChecker#updated?` may return `true` when using `travel_to` in test environments, even if no files were changed.

This happens because `@last_update_at` is initialized with a frozen time (from `travel_to`), while `File.mtime` reflects real system time. When compared, this results in false positives and class reloading during tests, which can lead to failures like broken DB connections or transaction rollback.

Fixes #55112

---

### Detail

This Pull Request changes the initialization logic of `FileUpdateChecker` to improve robustness in test environments:

- Introduces `@init_time` to store the real system time at initialization.
- In `max_mtime`, adds logic to detect test time manipulation via `travel_to` by introducing a constant `TIME_TRAVEL_THRESHOLD = 60` (seconds).
- If significant time travel is detected, files modified after `@init_time` are ignored when computing `updated?`, preventing false positives.
- This logic is scoped to test-only conditions and does not affect production behavior.

---

### Additional information

- All existing tests pass
- Reproduction test based on issue steps has been added
- The fix maintains compatibility with future `mtime` edge-cases (`test_should_be_robust_to_handle_files_with_wrong_modified_time` still passes)
- No CHANGELOG update required as this is a minor bug fix scoped to test behavior

---

### Checklist

* [x] Commit message has a detailed description and references issue #55112.
* [x] Tests are added to reproduce the problem and now pass.
* [x] CHANGELOG not updated because this is a test-related fix with no user-visible behavior change.

### DIFF
--- a/activesupport/lib/active_support/file_update_checker.rb
+++ b/activesupport/lib/active_support/file_update_checker.rb
@@ -33,6 +33,7 @@ module ActiveSupport
   #     i18n_reloader.execute_if_updated
   #   end
   class FileUpdateChecker
+    TIME_TRAVEL_THRESHOLD = 60 # seconds
     # It accepts two parameters on initialization. The first is an array
     # of files and the second is an optional hash of directories. The hash must
     # have directories as keys and the value is an array of extensions to be
@@ -49,6 +50,7 @@ module ActiveSupport
       @files = files.freeze
       @globs = compile_glob(dirs)
       @block = block
+      @init_time = Time.now
 
       @watched    = nil
       @updated_at = nil
@@ -131,7 +133,14 @@ module ActiveSupport
         paths.each do |path|
           mtime = File.mtime(path)
 
+          # Ignore files with mtime in the future relative to the current time
           next if time_now.compare_without_coercion(mtime) < 0
+
+          # Handle time travel in tests: if there's a significant time difference
+          # between init and current time, ignore files modified after initialization
+          # to prevent false positives from test time manipulation
+          time_diff = (time_now - @init_time).abs
+          next if time_diff > TIME_TRAVEL_THRESHOLD && @init_time.compare_without_coercion(mtime) < 0
 
           if max_mtime.nil? || max_mtime.compare_without_coercion(mtime) < 0
             max_mtime = mtime
--- a/activesupport/test/file_update_checker_test.rb
+++ b/activesupport/test/file_update_checker_test.rb
@@ -2,9 +2,11 @@
 
 require_relative "abstract_unit"
 require_relative "file_update_checker_shared_tests"
+require "active_support/core_ext/integer/time"
 
 class FileUpdateCheckerTest < ActiveSupport::TestCase
   include FileUpdateCheckerSharedTests
+  include ActiveSupport::Testing::TimeHelpers
 
   def new_checker(files = [], dirs = {}, &block)
     ActiveSupport::FileUpdateChecker.new(files, dirs, &block)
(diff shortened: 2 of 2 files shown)


drymar:
Hi! All tests related to ActiveSupport are passing locally.
The `buildkite/rails` failure seems to be unrelated (in `activerecord postgresql:test`), and my changes do not touch AR or database behavior.

Happy to rebase or adjust if needed.

drymar:
Hi! Noticed this issue arose after changes in #51308, where `FileUpdateChecker#updated?` started being called during test setups that use `travel_to`.

This PR adds logic to avoid false positives by distinguishing real `mtime` drift from mocked time.
Tagging @fxn as the author of the related change — happy to hear your thoughts or suggestions 🙇

All tests pass and a reproduction case is included. Thanks for reviewing!

nvasilevski:
Unfortunately I don't think this is the right fix, I explained my thinking here https://github.com/rails/rails/issues/55112#issuecomment-3053261968

But TLDR: This PR make it look like there is an issue with `FileUpdateChecker` when running tests while in reality it shouldn't be doing any watching in the process that runs tests in the first place

rafaelfranca:
Fixed by https://github.com/rails/rails/pull/55710

PR #55115, opened 2025-05-24
Memoize successful reflection cache validation

fxn:
Association reflections are objects that store metadata about an association.

Once an association has been defined, it is all set, the association cannot be mutated, and the reflection should not be mutated either (though, technically, Active Record does not freeze the related state, something I might look at later).

Associations are lazy, to delay accessing models as much as possible. In particular, the validity of the association is not checked when the macro is processed, but later on when/if the association is used in one particular record, for every record. That is why association initializers have `reflection.check_validity!` calls ([here](https://github.com/rails/rails/blob/ca372893014050ace4c7f58eade51772ab810fa6/activerecord/lib/active_record/associations/association.rb#L42)).

But this is overkill, because that means that for every instantiated record and every association accessed in that record, the corresponding reflection is validated again, and again. However, one time should suffice!

With everything being so lazy, I don't quite see some class-level spot in which to validate the reflections. But, at least, we can memoize the reflection validation if it was successful. This is what this patch does.

I benchmarked

```ruby
Author.new.posts
```

and the speedup is about 6%.

### DIFF
--- a/activerecord/lib/active_record/reflection.rb
+++ b/activerecord/lib/active_record/reflection.rb
@@ -520,6 +520,8 @@ module ActiveRecord
 
       def initialize(name, scope, options, active_record)
         super
+
+        @validated = false
         @type = -(options[:foreign_type]&.to_s || "#{options[:as]}_type") if options[:as]
         @foreign_type = -(options[:foreign_type]&.to_s || "#{name}_type") if options[:polymorphic]
         @join_table = nil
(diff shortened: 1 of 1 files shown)


byroot (approved):
Works for me. 

The laziness of AR is indeed painful for this sort of stuff, but it might be possible to validate all association after eager load (some users might discover they have invalid associations that are dead code though).

fxn:
@byroot awesome, let's merge then.

One question that makes me think is that `reflect_on_association` can return invalid reflections to the user, if the user inspects them before the association is used. That does not seem right to me, perhaps we could leverage this memoization to verify there too.

PR #55121, opened 2025-05-27
Add `only_columns` to activerecord

Glyde02:
### Motivation / Background

This Pull Request has been created because Rails support set only ignored_column to the models. But when you use the same database model in different projects, there is issue with deleting column.
Our way for deleting:
- delete all usages
- add column and column+to_remove into ignored list
- rename column in db
- drop column in db, clean code

For this way needed to create project_count*2 the same pull requests. Allowed columns can help avoid this routine, because not all projects use all columns.

### Detail

This Pull Request changes model_schema. It's added new methods for get and set allowed_columns. Also added checks for raising error if you use both methods at the same time.
Allowed columns, also based on ignored_columns list, like negative of them. 

### Additional information

Open to suggestions for a list of default columns. Are there any other fields that should go there, if there are any in the base.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/model_schema.rb
+++ b/activerecord/lib/active_record/model_schema.rb
@@ -181,6 +181,7 @@ module ActiveRecord
       self.protected_environments = ["production"]
 
       self.ignored_columns = [].freeze
+      self.only_columns = [].freeze
 
       delegate :type_for_attribute, :column_for_attribute, to: :class
 
@@ -334,6 +335,12 @@ module ActiveRecord
         @ignored_columns || superclass.ignored_columns
       end
 
+      # The list of columns names the model should allow. Only columns are used to define
+      # attribute accessors, and are referenced in SQL queries.
+      def only_columns
+        @only_columns || superclass.only_columns
+      end
+
       # Sets the columns names the model should ignore. Ignored columns won't have attribute
       # accessors defined, and won't be referenced in SQL queries.
       #
@@ -366,8 +373,15 @@ module ActiveRecord
       #   user = Project.create!(name: "First Project")
       #   user.category # => raises NoMethodError
       def ignored_columns=(columns)
+        check_model_columns(@only_columns.present?)
         reload_schema_from_cache
         @ignored_columns = columns.map(&:to_s).freeze
+      end
+
+      def only_columns=(columns)
+        check_model_columns(@ignored_columns.present?)
+        reload_schema_from_cache
+        @only_columns = columns.map(&:to_s).freeze
       end
 
       def sequence_name
@@ -579,6 +593,7 @@ module ActiveRecord
           child_class.reload_schema_from_cache(false)
           child_class.class_eval do
             @ignored_columns = nil
+            @only_columns = nil
           end
         end
 
@@ -592,7 +607,11 @@ module ActiveRecord
           end
 
           columns_hash = schema_cache.columns_hash(table_name)
-          columns_hash = columns_hash.except(*ignored_columns) unless ignored_columns.empty?
+          if only_columns.present?
+            columns_hash = columns_hash.slice(*only_columns)
+          elsif ignored_columns.present?
+            columns_hash = columns_hash.except(*ignored_columns)
+          end
           @columns_hash = columns_hash.freeze
 
           _default_attributes # Precompute to cache DB-dependent attribute types
@@ -630,6 +649,10 @@ module ActiveRecord
           end
 
           type
+        end
+
+        def check_model_columns(columns_present)
+          raise ArgumentError, "You can not use both only_columns and ignored_columns in the same model." if columns_present
         end
     end
   end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Add `ActiveRecord::Base.only_columns`
+
+    Similar in use case to `ignored_columns` but listing columns to consider rather than the ones
+    to ignore.
+
+    Can be useful when working with a legacy or shared database schema, or to make safe schema change
+    in two deploys rather than three.
+
+    *Anton Kandratski*
+
 *   Use `PG::Connection#close_prepared` (protocol level Close) to deallocate
     prepared statements when available.
 
--- a/activerecord/test/cases/base_test.rb
+++ b/activerecord/test/cases/base_test.rb
@@ -1833,6 +1833,31 @@ class BasicsTest < ActiveRecord::TestCase
     assert_respond_to SymbolIgnoredDeveloper.new, :last_name?
   end
 
+  test "permitted columns have attribute methods" do
+    assert_respond_to OnlyColumnsDeveloper.new, OnlyColumnsDeveloper.primary_key
+    assert_respond_to OnlyColumnsDeveloper.new, :name
+    assert_respond_to OnlyColumnsDeveloper.new, :name=
+    assert_respond_to OnlyColumnsDeveloper.new, :name?
+    assert_respond_to OnlyColumnsDeveloper.new, :salary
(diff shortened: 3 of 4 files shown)


fatkodima:
> But when you use the same database model in different projects

Can you describe this in details.

I think for each project you can configure `ignore_column` to ignore columns it is not using, and then when removing a column - do the work only in one project.

Glyde02:
> > But when you use the same database model in different projects
> 
> Can you describe this in details.
> 
> I think for each project you can configure `ignore_columns` to ignore columns it is not using from the beginning, and then when removing a column - do the work only in one project.

I can give example. In DB we have model User
```sql
users (
  id,
  name,
  email,
  phone,
  address,
  admin_notes,
  created_at,
  updated_at
)
```

And for example we have different projects for User, Admin, HR etc. Then we can configure only needed columns for each project

HR:
```
  def self.allowed_columns
    %w[ id name phone address ]
  end
end
```

Admin:
```
  def self.allowed_columns
    %w[ id name email phone address admin_notes created_at updated_at ]
  end
end
```

With `ignored_columns` we need to update each projects after each creating/editing column in db.

## ✅ Safe column removal with `allowed_columns`

`allowed_columns` also helps avoid some steps with safely removing column. 
Steps:

1. Delete all usages of the column in the code.
2. Remove the column from the `allowed_columns` list.
3. Rename the column in the database by adding a `_to_remove` suffix (e.g., `old_field` → `old_field_to_remove`). It's can help to track using this column cross all projects.
4. Drop the column from the database.

n-rodriguez:
You could also implement this in your own gem (public or private)

mikhliuk-k:
> You could also implement this in your own gem (public or private)

Yes, we already did that — battle-proven in our project. Now we've decided it might be useful to have this out of the box. And IMO, it fits nicely alongside the `ignored_columns` ;)

YegorZdanovich:
hi @fatkodima why so passive aggressive? If you don't have such problems that's ok and thanks for your feedback,  but in real world there are many applications which share same database, with huge legacy logic, and that feature helps to identify all dependencies in explicit way.

we are not first with such request, there some ideas 6 years ago https://discuss.rubyonrails.org/t/activerecord-blocklist-ignored-columns-allowlist-feature-idea/73667

byroot:
> why so passive aggressive?

I don't think Dima is being aggressive in any way. He's just providing his technical opinion in a, perhaps terse, but honest way.

That said, I do see some value in the proposed feature. It can make sense in some contexts to want to work with an allow-list rather than deny-list. It also saves you one "roundtrip" when you want to safely introduce a new column or remove one.

I think the implementation however would need some work:

  - I'm not sure `DEFAULT_ALLOWED_COLUMNS` make sense. Ok `updated_at` etc are incredibly common, but what if you want to remove them from a model?
  - I don't think `allowed_columns` is a good name. I get it comes from `allow-list`, but it really reads weird and isn't quite descriptive enough. I don't have any evident proposal though, perhaps `only_columns=` ? Or even just `columns=`?
  - Setting this value shouldn't call into `AR::Base.column_names` because that trigger loading the schema, which is a big no no. Any kind of filtering need to be delayed until the schema is loaded.

So if you are willing to put the work to get this feature to the required quality level, I'm open to review and merge it. However I don't have much time to give very detailed directions these days.

simi:
> Why do you have 10 different projects sharing the same database and the same table?

That is common use case, I have seen it often. Sometimes it is crucial for performance to access DB directly.

Glyde02:
> if you are willing to put the work to get this feature to the required quality level

Yes. I think it's a useful feature, so I'm willing to make fixes to merge it.
In recent commits, I've renamed the `allowed_columns` to `permitted_columns`, like params method in ActionController.  And I also separated the list of permitted_columns from the ignored_columns. And I have one question. What logic should we follow if we're trying to get a second list.

```
class Developer
  self.permitted_columns = %w[name salary]
  ...
end
```

`Developer.permitted_columns => ["name", "salary"]`
`Developer.ignored_columns => ?`

And the same issue when we define ignored_columns and call permitted_columns.

I propose to return `columns - permitted_columns` for first example, but only if schema is loaded. Or is it ok, to make load_schema there?

byroot:
> I've renamed the `allowed_columns` to `permitted_columns`

Why not just `columns`? Or `only_columns`? I don't have a very strong opinion, but `permitted` seem a bit weird to be.

> I propose to return columns - permitted_columns for first example, but only if schema is loaded.

I don't think it's a good idea. It's better if these two methods are simple accessors without any specific logic.

Glyde02:
> Why not just `columns`?

We have already method `columns` in model_schema and I can not change it. [link](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/model_schema.rb#L434-L436)

> Or `only_columns`?

I don't think this name fully captures its essence. Only columns for what? For loading columns from the database, for availability them in the model, or for some kind of availability to change. The user will not be able to guess so easily what exactly only means.
In comparison permitted is already known in the API, and in fact performs almost a similar function. And the word permit itself is like only + allow.

> It's better if these two methods are simple accessors

Maybe. I don't know. But I'm confused about one problem (though maybe it's not a problem). Now without looking at the code it is enough to call Model.ignored_columns to find out whether all columns from the database are used or not. I don't know if there are scenarios when you need to keep track of this in the project. 
If these two lists will be independent of each other, then checking of columns usage restriction will have to be looked at by two methods `Model.ignored_columns + Model.allowed_columns`
Initially in logic we have `Model.columns - Model.allowed_columns = Model.ignored_columns` but without dependent lists methods from activerecord will return incorrect results.

Glyde02:
@byroot 
Have you any updates?

byroot:
Sorry, as I warned, I've been busy lately. Let me have another look.

activerecord/lib/active_record/model_schema.rb
```diff
@@ -181,6 +181,7 @@ module ModelSchema
       self.protected_environments = ["production"]
 
       self.ignored_columns = [].freeze
+      self.permitted_columns = [].freeze
```
byroot: I still don't think `permitted` is a good prefix. If we absolutely need a prefix than `only_` could work.

activerecord/lib/active_record/model_schema.rb
```diff
@@ -331,7 +332,21 @@ def real_inheritance_column=(value) # :nodoc:
       # The list of columns names the model should ignore. Ignored columns won't have attribute
       # accessors defined, and won't be referenced in SQL queries.
       def ignored_columns
-        @ignored_columns || superclass.ignored_columns
+        return @ignored_columns if @ignored_columns
+        return superclass.ignored_columns if superclass.respond_to?(:ignored_columns)
+        return column_names - @permitted_columns if schema_loaded? && @permitted_columns.present?
```
byroot: We should avoid having different behavior based on whether the schema is loaded or not. I don't think this is necessary at all.

activerecord/lib/active_record/model_schema.rb
```diff
@@ -331,7 +332,21 @@ def real_inheritance_column=(value) # :nodoc:
...
+
+        []
+      end
+
+      # The list of columns names the model should allow. Permitted columns are used to define
+      # attribute accessors, and are referenced in SQL queries.
+      def permitted_columns
+        return @permitted_columns if @permitted_columns
+        return superclass.permitted_columns if superclass.respond_to?(:permitted_columns)
+        return column_names - @ignored_columns if schema_loaded? && @ignored_columns.present?
```
byroot: Same here. Let's never consider the schema.

Glyde02:
Hi @byroot 
I renamed this feature to only_columns and removed considering to the schema

vertden:
@byroot any chance you could review?

byroot:
uh, somehow I was convinced I had merged this.

PR #53708, opened 2024-11-22
Add login_as(user) testing helper when generating authentication

gobijan:
### Motivation / Background

This Pull Request has been created because the generated authentication logic does not offer a way to test when a user is authenticated. There is a discussion going in here I initialized: #53207
So this PR adds a `SessionTestHelper` module with `sign_in_as(user)` and `sign_out` test helpers when
running `rails g authentication`. This simplifies authentication in integration tests.

### Detail

This Pull Request changes the generator for authentication to add described test helpers.

### Additional information

There were a few discussions how to do it best and for now I settle with the approach from the PR.

### Checklist

Before submitting the PR make sure the following are checked:

* [X] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [X] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [X] Tests are added or updated if you fix a bug or add a feature.
* [X] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55124, opened 2025-05-27
Add System test support in session test helper

excid3:
### Motivation / Background

The new SessionTestHelper from #53708 is accessible in system tests but does not work since there is no direct access to cookies. 

Users will find this confusing, so this PR updates the method to also work for system tests.

### Detail

This uses Lewis Buckley's approach for setting cookies directly in the browser.
https://world.hey.com/lewis/faster-rails-system-tests-f01df53b

### Additional information

There might be a cleaner way of doing this than introducing a conditional if anyone has improvements to suggest there.

The visit to a URL (in this case `new_session_url` from the generator) is required so the browser is on the correct domain. Chrome defaults to opening a tab at `data:,` which doesn't allow setting the cookie on the other domain. We can skip this if the browser is already loaded.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/authentication/templates/test/test_helpers/session_test_helper.rb.tt
+++ b/railties/lib/rails/generators/rails/authentication/templates/test/test_helpers/session_test_helper.rb.tt
@@ -4,7 +4,12 @@ module SessionTestHelper
 
     ActionDispatch::TestRequest.create.cookie_jar.tap do |cookie_jar|
       cookie_jar.signed[:session_id] = Current.session.id
-      cookies[:session_id] = cookie_jar[:session_id]
+      if defined?(ApplicationSystemTestCase) && is_a?(ApplicationSystemTestCase)
+        visit new_session_url unless page.current_url.start_with?("http")
+        page.driver.browser.manage.add_cookie(name: :session_id, value: cookie_jar[:session_id], sameSite: :Lax, httpOnly: true)
+      else
+        cookies[:session_id] = cookie_jar[:session_id]
+      end
     end
   end
 


p8:
I was curious how Devise handles this.
Devise [calls](https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/test/integration_helpers.rb#L40) `login_as`, which is a Warden helper.
Warden adds a callback that logs the user in on the next request:
https://github.com/wardencommunity/warden/blob/67f5ba6baaa7564ec79afef02cf3a4d0f7d312e5/lib/warden/test/helpers.rb#L18-L23
This is a nice way to avoid the extra `visit` call (and avoid calling the `driver` and `cookie_jar`), but it's probably difficult to implement this in Rails.

railties/lib/rails/generators/rails/authentication/templates/test/test_helpers/session_test_helper.rb.tt
```diff
@@ -4,7 +4,12 @@ module SessionTestHelper
 
     ActionDispatch::TestRequest.create.cookie_jar.tap do |cookie_jar|
       cookie_jar.signed[:session_id] = Current.session.id
-      cookies[:session_id] = cookie_jar[:session_id]
+      if is_a? ApplicationSystemTestCase
+        visit new_session_url unless page.current_url.start_with?("http")
+        page.driver.browser.manage.add_cookie(name: :session_id, value: cookie_jar[:session_id], sameSite: :Lax, httpOnly: true)
```
p8, line 9: The `sameSite` should probably be set to what is configured for ActionDispatch:
https://edgeguides.rubyonrails.org/configuring.html#config-action-dispatch-cookies-same-site-protection
Although that could be a Proc as well 🙁.

excid3: There's no `request` accessible here to use that unfortunately.

railties/lib/rails/generators/rails/authentication/templates/test/test_helpers/session_test_helper.rb.tt
```diff
@@ -4,7 +4,12 @@ module SessionTestHelper
 
     ActionDispatch::TestRequest.create.cookie_jar.tap do |cookie_jar|
       cookie_jar.signed[:session_id] = Current.session.id
-      cookies[:session_id] = cookie_jar[:session_id]
+      if is_a? ApplicationSystemTestCase
```
t27duck: `ApplicationSystemTestCase` is only defined in `test/application_system_test_case.rb` which is required in system tests only. 

When I try to run only controller and integration tests locally with this change, the test suite fails with uninitialized constant errors.

rafaelfranca:
With system tests being now removed by default, I think we should avoid adding this helper. Thanks

PR #55125, opened 2025-05-28
Freeze instance of current attributes

janko:
### Motivation / Background

We used additional memoization inside our current attributes subclass, setting a custom instance variable, which caused the instance variable to persist across requests instead of being reset, resulting in serious issues in our app.

### Detail

This prevents setting custom instance variables inside current attributes, which would necessarily lead to global state being leaked across requests. It does so by freezing the instance.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -201,4 +201,6 @@
 
     *Jonathan del Strother*
 
+*   Freeze instances of `ActiveSupport::CurrentAttributes` to prevent assignment of custom instance variables, which would cause global state to be leaked across requests.
+
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/activesupport/CHANGELOG.md) for previous changes.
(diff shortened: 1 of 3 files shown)


byroot:
This is definitely a footgun that deserve to be addressed. I'm not sure freezing is really the best way though, and more importantly, I highly suspect it would break lots of apps.

I don't have the time to investigate deeper right now, but we should look into deprecating the `reset` interface (on the instance at least) and always starting fresh from a new object.

Trying to re-use objects like this is a bad idea.

janko:
> Trying to re-use objects like this is a bad idea.

Yeah, I was surprised to see the current attributes object being reused across requests. If it shouldn't carry any persistent state across requests, then we should be able to throw it away after the request.

I guess the reset hooks allow you to store additional state that you can reset between requests, but I think many people will not realize they have to reset it.

janko:
Appreciate the quick fix! ❤️

PR #55127, opened 2025-05-29
Active Job Continuations

djmb:
Continuations provide a mechanism for interrupting and resuming jobs. This allows long running jobs to make progress across application restarts.

Jobs should include the `ActiveJob::Continuable` module to enable continuations. Continuable jobs are automatically retried when interrupted.

Use the `step` method to define the steps in your job. Steps can use an optional cursor to track progress in the step.

Steps are executed as soon as they are encountered. If a job is interrupted, previously completed steps will be skipped. If a step is in progress, it will be resumed with the last recorded cursor.

Code that is not part of a step will be executed on each job execution.

You can pass a block or a method name to the step method. The block will be called with the step object as an argument. Methods can either take no arguments or a single argument for the step object.

```ruby
class ProcessImportJob < ApplicationJob

  def perform(import_id)
    # This always runs, even if the job is resumed.
    @import = Import.find(import_id)

    step :validate do
      @import.validate!
    end

    step :process_records do |step|
      @import.records.find_each(start: step.cursor)
        record.process
        step.advance! from: record.id
      end
    end

    step :reprocess_records
    step :finalize
  end

  def reprocess_records(step)
    @import.records.find_each(start: step.cursor)
      record.reprocess
      step.advance! from: record.id
    end
  end

  def finalize
    @import.finalize!
  end
end
```

**Cursors**

Cursors are used to track progress within a step. The cursor can be any object that is serializable as an argument to
`ActiveJob::Base.serialize`. It defaults to `nil`.

When a step is resumed, the last cursor value is restored. The code in the step is responsible for using the cursor to continue from the right point.

`set!` sets the cursor to a specific value.

```ruby
step :iterate_items do |step|
  items[step.cursor..].each do |item|
    process(item)
    step.set! (step.cursor || 0) + 1
  end
end
```

A starting value for the cursor can be set when defining the step:

```ruby
step :iterate_items, start: 0 do |step|
  items[step.cursor..].each do |item|
    process(item)
    step.set! step.cursor + 1
  end
end
```

The cursor can be advanced with `advance!`. This calls `succ` on the current cursor value. It raises an
`ActiveJob::Continuation::UnadvanceableCursorError` if the cursor does not implement `succ`.

```ruby
step :iterate_items, start: 0 do |step|
  items[step.cursor..].each do |item|
    process(item)
    step.advance!
  end
end
```

You can optionally pass a `from` argument to `advance!`. This is useful when iterating over a collection of records where IDs may not be contiguous.

```ruby
step :process_records do |step|
  import.records.find_each(start: step.cursor)
    record.process
    step.advance! from: record.id
  end
end
```

You can use an array to iterate over nested records:

```ruby
step :process_nested_records, start: [ 0, 0 ] do |step|
  Account.find_each(start: step.cursor[0]) do |account|
    account.records.find_each(start: step.cursor[1]) do |record|
      record.process
      step.set! [ account.id, record.id + 1 ]
    end
    step.set! [ account.id + 1, 0 ]
  end
end
```

Setting or advancing the cursor creates a checkpoint. You can also create a checkpoint manually by calling the `checkpoint!` method on the step. This is useful if you want to allow interruptions, but don't need to update the cursor.

```ruby
step :destroy_records do |step|
  import.records.find_each do |record|
    record.destroy!
    step.checkpoint!
  end
end
```

**Checkpoints**

A checkpoint is where a job can be interrupted. At a checkpoint the job will call `queue_adapter.stopping?`. If it returns true, the job will raise an `ActiveJob::Continuation::Interrupt` exception.

There is an automatic checkpoint at the end of each step. Within a step calling one is created when calling `set!`, `advance!` or `checkpoint!`.

Jobs are not automatically interrupted when the queue adapter is marked as stopping - they will continue to run either until the next checkpoint, or when the process is stopped.

This is to allow jobs to be interrupted at a safe point, but it also means that the jobs should checkpoint more frequently than the shutdown timeout to ensure a graceful restart.

When interrupted, the job will automatically retry with the progress serialized in the job data under the `continuation` key.

The serialized progress contains:
- a list of the completed steps
- the current step and its cursor value (if one is in progress)

**Errors**

If a job raises an error and is not retried via ActiveJob, it will be passed back to the queue adapter and any progress in this execution will be lost.

To mitigate this, the job will automatically retried if it raises an error after it has made progress. Making progress is defined as having completed a step or advanced the cursor within the current step.

**Queue Adapter support**

Active Job Continuations call the `stopping?` method on the queue adapter to check if we are in the shutdown phase. By default this will return false, so the adapters will need to be updated to implement this method.

This implements the `stopping?` method in the Test and Sidekiq adapters.

It would be possible to add support to Delayed Job via a plugin, but it would probably be better to add a new lifecycle callback to DJ for when it is shutting down.

Resque also will require a new hook before it can be supported.

Solid Queue's adapter is not part of Rails, but support can be added there via the `on_worker_stop` hook.

**Inspiration**

This took a lot inspiration from Shopify's [job-iteration](https://github.com/Shopify/job-iteration) gem.

The main differences are:
- Continuations are Active Job only, so they don't provide the custom enumerators that job-iteration does.
- They allow multi-step flows
- They don't intercept the perform method
- Continuations are a sharp knife - you need to manually checkpoint and update the cursor. But you could build a job-iteration-like API on top of them.

**Future work**

It would be a good exercise to see if the job-iteration gem could be adapted to run on top of Active Job Continuations to highlight any missing features - we'd want to add things like max iteration time, max job runtime and forcing a job to stop.

Another thing to consider is a mechanism for checking whether it is safe to call a checkpoint. Ideally you wouldn't allow them within database transactions as they'll cause a rollback. We can maybe inject checkpoint safety handlers and add a default one that checks whether we are in any active transactions.

### DIFF
--- a/activejob/lib/active_job/continuable.rb
+++ b/activejob/lib/active_job/continuable.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+module ActiveJob
+  # = Active Job Continuable
+  #
+  # Mix ActiveJob::Continuable into your job to enable continuations.
+  #
+  # See +ActiveJob::Continuation+ for usage. # The Continuable module provides the ability to track the progress of your jobs,
+  # and continue from where they left off if interrupted.
+  #
+  module Continuable
+    extend ActiveSupport::Concern
+
+    CONTINUATION_KEY = "continuation"
+
+    included do
+      retry_on Continuation::Interrupt, attempts: :unlimited
+      retry_on Continuation::AfterAdvancingError, attempts: :unlimited
+
+      around_perform :continue
+    end
+
+    def step(step_name, start: nil, &block)
+      continuation.step(step_name, start: start) do |step|
+        if block_given?
+          block.call(step)
+        else
+          step_method = method(step_name)
+
+          raise ArgumentError, "Step method '#{step_name}' must accept 0 or 1 arguments" if step_method.arity > 1
+
+          if step_method.parameters.any? { |type, name| type == :key || type == :keyreq }
+            raise ArgumentError, "Step method '#{step_name}' must not accept keyword arguments"
+          end
+
+          step_method.arity == 0 ? step_method.call : step_method.call(step)
+        end
+      end
+    end
+
+    def serialize
+      super.merge(CONTINUATION_KEY => continuation.to_h)
+    end
+
+    def deserialize(job_data)
+      super
+      @continuation = Continuation.new(self, job_data.fetch(CONTINUATION_KEY, {}))
+    end
+
+    private
+      def continuation
+        @continuation ||= Continuation.new(self, {})
+      end
+
+      def continue(&block)
+        continuation.continue(&block)
+      end
+  end
+end
--- a/activejob/lib/active_job/continuation.rb
+++ b/activejob/lib/active_job/continuation.rb
@@ -1,0 +1,321 @@
+# frozen_string_literal: true
+
+require "active_support/core_ext/numeric/time"
+require "active_job/continuable"
+
+module ActiveJob
+  # = Active Job \Continuation
+  #
+  # Continuations provide a mechanism for interrupting and resuming jobs. This allows
+  # long-running jobs to make progress across application restarts.
+  #
+  # Jobs should include the ActiveJob::Continuable module to enable continuations.
+  # \Continuable jobs are automatically retried when interrupted.
+  #
+  # Use the +step+ method to define the steps in your job. Steps can use an optional
+  # cursor to track progress in the step.
+  #
+  # Steps are executed as soon as they are encountered. If a job is interrupted, previously
+  # completed steps will be skipped. If a step is in progress, it will be resumed
+  # with the last recorded cursor.
+  #
+  # Code that is not part of a step will be executed on each job run.
+  #
+  # You can pass a block or a method name to the step method. The block will be called with
+  # the step object as an argument. Methods can either take no arguments or a single argument
+  # for the step object.
+  #
+  #   class ProcessImportJob < ApplicationJob
+  #     include ActiveJob::Continuable
+  #
+  #     def perform(import_id)
+  #       # This always runs, even if the job is resumed.
+  #       @import = Import.find(import_id)
+  #
+  #       step :validate do
+  #         @import.validate!
+  #       end
+  #
+  #       step(:process_records) do |step|
+  #         @import.records.find_each(start: step.cursor)
+  #           record.process
+  #           step.advance! from: record.id
+  #         end
+  #       end
+  #
+  #       step :reprocess_records
+  #       step :finalize
+  #     end
+  #
+  #     def reprocess_records(step)
+  #       @import.records.find_each(start: step.cursor)
+  #         record.reprocess
+  #         step.advance! from: record.id
+  #       end
+  #     end
+  #
+  #     def finalize
+  #       @import.finalize!
+  #     end
+  #   end
+  #
+  # === Cursors
+  #
+  # Cursors are used to track progress within a step. The cursor can be any object that is
+  # serializable as an argument to +ActiveJob::Base.serialize+. It defaults to +nil+.
+  #
+  # When a step is resumed, the last cursor value is restored. The code in the step is responsible
+  # for using the cursor to continue from the right point.
+  #
+  # +set!+ sets the cursor to a specific value.
+  #
+  #   step :iterate_items do |step|
+  #     items[step.cursor..].each do |item|
+  #       process(item)
+  #       step.set! (step.cursor || 0) + 1
+  #     end
+  #   end
+  #
+  # An starting value for the cursor can be set when defining the step:
+  #
+  #   step :iterate_items, start: 0 do |step|
+  #     items[step.cursor..].each do |item|
+  #       process(item)
+  #       step.set! step.cursor + 1
+  #     end
+  #   end
+  #
+  # The cursor can be advanced with +advance!+. This calls +succ+ on the current cursor value.
+  # It raises an ActiveJob::Continuation::UnadvanceableCursorError if the cursor does not implement +succ+.
+  #
+  #   step :iterate_items, start: 0 do |step|
+  #     items[step.cursor..].each do |item|
+  #       process(item)
+  #       step.advance!
+  #     end
+  #   end
+  #
+  # You can optionally pass a +from+ argument to +advance!+. This is useful when iterating
+  # over a collection of records where IDs may not be contiguous.
+  #
+  #   step :process_records do |step|
+  #     import.records.find_each(start: step.cursor)
+  #       record.process
+  #       step.advance! from: record.id
+  #     end
+  #   end
+  #
+  # You can use an array to iterate over nested records:
+  #
+  #   step :process_nested_records, start: [ 0, 0 ] do |step|
+  #     Account.find_each(start: step.cursor[0]) do |account|
+  #       account.records.find_each(start: step.cursor[1]) do |record|
+  #         record.process
+  #         step.set! [ account.id, record.id + 1 ]
+  #       end
+  #       step.set! [ account.id + 1, 0 ]
+  #     end
+  #   end
+  #
+  # Setting or advancing the cursor creates a checkpoint. You can also create a checkpoint
+  # manually by calling the +checkpoint!+ method on the step. This is useful if you want to
+  # allow interruptions, but don't need to update the cursor.
+  #
+  #   step :destroy_records do |step|
+  #     import.records.find_each do |record|
+  #       record.destroy!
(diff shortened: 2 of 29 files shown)


dhh:
Fantastic work, @djmb!

activejob/lib/active_job/continuable.rb
```diff
@@ -0,0 +1,59 @@
...
+        else
+          step_method = method(step_name)
+
+          raise ArgumentError, "Step method '#{step_name}' must accept 0 or 1 arguments" if step_method.arity > 1
+
+          if step_method.parameters.any? { |type, name| type == :key || type == :keyreq }
+            raise ArgumentError, "Step method '#{step_name}' must not accept keyword arguments"
+          end
+
+          step_method.arity == 0 ? step_method.call : step_method.call(step)
```
byroot, line 36: Is making the argument optional really this necessary? Allocating a `Method` instance and then its parameters isn't cheap.

It's actually very expensive once you compare it to a simple method call.

activejob/lib/active_job/continuable.rb
```diff
@@ -0,0 +1,59 @@
...
+  #
+  # Mix ActiveJob::Continuable into your job to enable continuations.
+  #
+  # See +ActiveJob::Continuation+ for usage. # The Continuable module provides the ability to track the progress of your jobs,
+  # and continue from where they left off if interrupted.
+  #
+  module Continuable
+    extend ActiveSupport::Concern
+
+    CONTINUATION_KEY = "continuation"
```
byroot, line 14: In the past, the keys used by Active Job itself were prefixed with `_aj_`, e.g.:

```ruby
      GLOBALID_KEY = "_aj_globalid"
      # :nodoc:
      SYMBOL_KEYS_KEY = "_aj_symbol_keys"
      # :nodoc:
      RUBY2_KEYWORDS_KEY = "_aj_ruby2_keywords"
```

We probably should do the same here.

activejob/lib/active_job/continuation.rb
```diff
@@ -0,0 +1,321 @@
...
+  #
+  class Continuation
+    extend ActiveSupport::Autoload
+
+    autoload :Step
+
+    # Raised when a job is interrupted, allowing Active Job to requeue it.
+    # This inherits from +Exception+ rather than +StandardError+, so it's not
+    # caught by normal exception handling.
+    class Interrupt < Exception; end
```
byroot, line 169: Have you considered using `throw/catch` instead?

activejob/lib/active_job/continuation.rb
```diff
@@ -0,0 +1,321 @@
...
+        instrument :step_interrupted, step: step
+        raise
+      end
+
+      def instrument_job(event)
+        instrument event, description: description, completed_steps: completed, current_step: current
+      end
+
+      def instrument(event, payload = {})
+        job.send(:instrument, event, **payload)
```
byroot, line 318: It would be best to avoid `send` here, as it not good for JIT and such. The method can be public but marked as `# :nodoc:`.

activejob/lib/active_job/continuation/step.rb
```diff
@@ -0,0 +1,77 @@
...
+        checkpoint!
+      end
+
+      # Advance the cursor from the current or supplied value
+      #
+      # The cursor will be advanced by calling the +succ+ method on the cursor.
+      # An UnadvanceableCursorError error will be raised if the cursor does not implement +succ+.
+      def advance!(from: nil)
+        from = cursor if from.nil?
+        raise UnadvanceableCursorError, "Cursor class '#{from.class}' does not implement succ, " unless from.respond_to?(:succ)
```
byroot, line 51: Rather than to check `respond_to?` on every single call, it would be preferable to handle `NoMethodError`.

activejob/test/jobs/continuable_duplicate_step_job.rb
```diff
@@ -0,0 +1,12 @@
...
+class ContinuableDuplicateStepJob < ActiveJob::Base
+  include ActiveJob::Continuable
+
+  def perform
+    step :duplicate do |step|
+    end
+    step :duplicate do |step|
+    end
+  end
+end
```
byroot, line 12: All these "fixture" jobs could be defined inside the test classes themselves.

ghiculescu:
@djmb will you be working on a solid queue PR or is that open to the community? It would be good to have this supported in the default Rails setup.

activejob/lib/active_job/continuable.rb
ghiculescu: Now that this is live is it too late to change it?

simi: @ghiculescu it wasn't released yet.

activejob/lib/active_job/continuable.rb
djmb: Given that a job that is using continuations would be expected to be long running and to have a reasonably small number of steps, I wouldn't expect the cost of this to be significant.

activejob/lib/active_job/continuation.rb
djmb: Oh yeah that's a much better way to do it 👍

activejob/test/jobs/continuable_duplicate_step_job.rb
djmb: There are a couple of test jobs defined in their classes, but the pattern here is separate files even though most jobs are only used by one test case. But either way works for me if you'd prefer I moved them.

djmb (commented):
Thanks for the review @byroot! I'll raise another PR to address your points.

activejob/lib/active_job/continuable.rb
byroot: Right. This isn't a hill I'm gonna die on, and indeed in this specific case it is probably fine.

I just tend to question more and more how helpful this type of DSL heavy API design really is. Aside from the performance consideration it makes the code harder to grep for / statically analyze for not a huge gain in term of readability:

e.g.
```ruby
step :finalize
```

Is only marginally nicer than:

```ruby
step :finalize { finalize }
```

But:

  - Will be qualified of "magic".
  - Will defeat JITs.
  - Will dead code analysis harder.
  - Requires you to know the continuation API well to understand which method is called. It's not immediately obvious to the uninitiated that `step :foo` does call `foo`.

But again, if you're set on it, I'm not gonna argue you forever on this.

activejob/test/jobs/continuable_duplicate_step_job.rb
byroot: Thanks. IMHO it's much easier to understand tests when the job they rely on is just above.

activejob/lib/active_job/continuable.rb
djmb: I take your points and just allowing the block format will make the implementation and documentation simpler.

There's not much in `step :finalize` vs `step :finalize { finalize }`, but when you have say 5 or more steps the repetition of the block format does start to grate a bit.

What do you think @dhh?

activejob/lib/active_job/continuable.rb
djmb: The `_aj_` prefix is used as part of argument serialization and not for [job data keys](https://github.com/rails/rails/blob/c7c72ee6d39667014e7413c1430a0776b52801b2/activejob/lib/active_job/core.rb#L119-L130), so I think the unprefixed name makes more sense here.

byroot: My bad. I mistakenly assumed that keys was in the arguments. If it's at the top-level, there shouldn't be risks of conflict indeed.

activejob/lib/active_job/continuation.rb
djmb: Ok, I've been looking at this and I'm concerned about the fact that `throw/catch` could do unexpected things. It will execute all the ensure blocks, but skip any attempts at cleanup via rescue blocks.

For example if we were inside a database transaction Rails would [commit that transaction because it doesn't see an exception](https://github.com/rails/rails/blob/b12f58620610f8247f897e7f396bb715bae93198/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L625-L649).

Throwing an exception has the drawback that the calling code might rescue and swallow it, but I think that maybe the lesser of two evils.

(We should actually disallow attempting to interrupt jobs inside database transactions and I have a plan to add that soon but that's a separate issue I think).

byroot: Thanks for taking a look.

djmb:
@ghiculescu - yes I'll be adding a PR for SolidQueue soon. I have some additional changes coming soon to add configuration and safety checks (e.g. ensuring that you don't interrupt database transactions) so I want to get those out first.

mech:
Is this doing the same thing as what https://github.com/fractaledmind/acidic_job is doing?

PR #55134, opened 2025-06-01
Move `prevent_writes` defaulting to params of role switching methods

joshuay03:
### Motivation / Background

In all public connection role switching methods, `prevent_writes` is set to `true` if the given role is the `reading` role. This 'defaulting' (overriding really) logic lives in the bodies of these methods (and in one case a private method - `#with_role_and_shard`), and isn't consistently documented.

I believe it's better to have this defaulting logic live in the params so that looking up the method signature documentation makes this obvious at first glance, and that instead of overriding the argument we should be raising an `ArgumentError` instead - it doesn't make sense to call `connected_to(role: :reading, prevent_writes: false)`, and with the current logic `prevent_writes` will be overridden to `true` anyway.

This PR implements these changes and also ensures that the defaulting logic is consistently documented in all of these methods.

### Additional information

Also includes some minor corrections to the method docs to differentiate between when a role argument is required and a shard argument is optional, and when either a role or shard or both can be provided ('and/or').

### Checklist

### DIFF
--- a/activerecord/test/cases/base_test.rb
+++ b/activerecord/test/cases/base_test.rb
@@ -1994,7 +1994,7 @@ class BasicsTest < ActiveRecord::TestCase
     assert_equal "`connects_to` can only be called on ActiveRecord::Base or abstract classes", error.message
   end
 
-  test "cannot call connected_to with role and shard on non-abstract classes" do
+  test "cannot call #connected_to with role and shard on non-abstract classes" do
     error = assert_raises(NotImplementedError) do
       Bird.connected_to(role: :reading, shard: :default) { }
     end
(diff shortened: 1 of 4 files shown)


activerecord/test/cases/base_test.rb
```diff
@@ -1978,24 +2016,35 @@ def test_protected_environments_are_stored_as_an_array_of_string
...
+
+    assert_equal "cannot set `prevent_writes` to false when `role` is `reading`.", error.message
+  end
+
   test "#connected_to_many with a single argument for classes" do
     ActiveRecord::Base.connected_to_many(SecondAbstractClass, role: :reading) do
-      assert SecondAbstractClass.current_preventing_writes
-      assert_not ActiveRecord::Base.current_preventing_writes
+      assert SecondAbstractClass.connected_to?(role: :reading)
+      assert_not ActiveRecord::Base.connected_to?(role: :reading)
```
joshuay03, line 2116: There are other tests for this method that test `current_preventing_writes`, what this should probably be testing is the connected role given that it's the provided argument. Same applies for the next test.

eileencodes:
I feel like there's a reason I didn't do this originally but I can't remember why.

Ultimately `prevent_writes` is more for testing or finding where you have writes before switching your queries to a replica. If you have a real replica, `prevent_writes` just catches the error earlier, it won't let you write to a db setup as a replica.

joshuay03:
> I feel like there's a reason I didn't do this originally but I can't remember why.

Ah, perhaps there's a valid reason to override instead. If there isn't, however, do you feel strongly against raising an error?

Essentially, disabling `prevent_writes` for the reader role is disallowed behavior; we're just being explicit about it now (and perhaps doing some users a favor if they've made an incorrect assumption at such a call site).

PR #48861, opened 2023-08-01
Fixes #47909 when using order or unscope

paulreece:
This fix corrects the logic in both the associated and missing methods. If the reflection.options hash has a key/value pair for :class_name it will use the association. This fixes [48651](https://github.com/rails/rails/issues/48651) which was not returning the correct value when using an enum association and querying for a record that had one key of the enum but was missing another. This also fixes [44964](https://github.com/rails/rails/issues/44964) where ActiveRecord was not properly aliasing self-referencing relations. If the reflection.options doesn’t contain the key/value pair then it will use the reflection.table_name. This fixes  [47909](https://github.com/rails/rails/issues/47909) in which a user was trying to find records that were either missing a relation or were missing a relation that was ordered or unscoped or were missing a relation that was using extends in the query which resulted in an ActiveRecord Exception. It also allows for using extends in any of these capacities.

* [ X] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [X ] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ X] Tests are added or updated if you fix a bug or add a feature.
* [X ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #48738, opened 2023-07-14
Restores functionality to the missing method when using enums and fixes #48651

paulreece:
### Motivation / Background

This Pull Request has been created because of the bug raised in issue [48651](https://github.com/rails/rails/issues/48651). 

Say you have the following models in Rails 7.0.6, and have used an enum in one and aliases in the other.

```Ruby
class Publisher < ApplicationRecord
  belongs_to :listing

  enum name: {
    foo: 0,
    bar: 1
  }
end

class Listing < ApplicationRecord
  has_many :publishers, dependent: :destroy
  has_one :foo_listing, -> { foo }, class_name: Publisher.name
  has_one :bar_listing, -> { bar }, class_name: Publisher.name
end
```

If you have an instance of Listing that has a foo publisher but not a bar publisher and call:

 ```Ruby 
Listing.joins(:foo_listing).where.missing(:bar_listing)
```

It returns nothing whereas one would expect it to return the Listing you created that only has a `:foo_listing`

Please see bug fix if more detail is needed linked above. 

### Detail

This Pull Request changes the missing method in `active_record/relation/query_methods.rb`

Adds another conditional check in the missing query method  in ActiveRecord. This extra check makes sure that the values in `:joins` are different than the values in `:left_outer_joins`.  When they are it is likely that you are using an enum.  I do this checking first for the existence of values in `:joins` and `:left_outer_joins`.  Then I compare the size of the intersection of the two Arrays to the size of the `:joins` Array. If they are different then it implements the scope method excluding the desired enum alias.  The `size` method was used to keep this performant.  This passes all checks in ActiveRecord, including those from [47940](https://github.com/rails/rails/issues/47909) and [45015](https://github.com/rails/rails/pull/45015).

### Checklist

* [ X] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [ X] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ X] Tests are added or updated if you fix a bug or add a feature.
* [ X] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55138, opened 2025-06-02
Fix has_many, through joins for self references with `.missing`

alexcameron89:
### Motivation / Background

This Pull Request has been created in order to address https://github.com/rails/rails/issues/55130

When using a model with a `has_many :through` association that references itself, `where.missing` does not properly handle table aliasing, causing SQL ambiguity errors or incorrect results.

```ruby
class Person < ActiveRecord::Base
  has_many :leadership_roles, foreign_key: "boss_id"
  has_many :subordinates, through: :leadership_roles

  has_many :subordinate_roles, foreign_key: "subordinate_id", class_name: "LeadershipRole"
  has_many :bosses, through: :subordinate_roles
end

class LeadershipRole < ActiveRecord::Base
  belongs_to :subordinate, class_name: "Person"
  belongs_to :boss, class_name: "Person"
end

Person.where.missing(:subordinates)
#   SELECT COUNT(*)
#   FROM "people"
#     LEFT OUTER JOIN "leadership_roles" ON "leadership_roles"."boss_id" = "people"."id"
#     LEFT OUTER JOIN "people" AS "subordinates_people" ON "subordinates_people"."id" = "leadership_roles"."subordinate_id"
#   WHERE "people"."id" IS NULL
#   -- expect to be WHERE "subordinates_people"."id" IS NULL
```

The issue occurs because when the scope's table name matches the association's table name (self-referential relationships), the method incorrectly uses the table name instead of the association alias in the WHERE clause.

A minimal reproduction script is available below.

### Additional information

Previously, the `.missing` method performed a check to see if the table name of `@scope`was the same as the `reflection`, but it was replaced with a few alternatives in
https://github.com/rails/rails/pull/48738/files and https://github.com/rails/rails/pull/48861

After reviewing the test cases from both PRs and related issues, adding this check back introduces no regressions while fixing the self-referential association case in https://github.com/rails/rails/issues/55130.

### Minimal Reproduction Script

```ruby

# Github issue: https://github.com/rails/rails/issues/55130
# Link: https://github.com/rails/rails/raw/refs/heads/main/guides/bug_report_templates/active_record.rb
# ================================================================================
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  create_table :people, force: true do |t|
    t.string :first_name
  end

  create_table :leadership_roles, force: true do |t|
    t.integer :subordinate_id
    t.integer :boss_id
  end
end

class Person < ActiveRecord::Base
  has_many :leadership_roles, foreign_key: "boss_id"
  has_many :subordinates, through: :leadership_roles

  has_many :subordinate_roles, foreign_key: "subordinate_id", class_name: "LeadershipRole"
  has_many :bosses, through: :subordinate_roles
end

class LeadershipRole < ActiveRecord::Base
  belongs_to :subordinate, class_name: "Person"
  belongs_to :boss, class_name: "Person"
end

  # ACTUAL QUERY:
  #   SELECT COUNT(*)
  #   FROM "people"
  #     LEFT OUTER JOIN "leadership_roles" ON "leadership_roles"."boss_id" = "people"."id"
  #     LEFT OUTER JOIN "people" AS "subordinates_people" ON "subordinates_people"."id" = "leadership_roles"."subordinate_id"
  #   WHERE "people"."id" IS NULL -- expect to be WHERE "subordinates_people"."id" IS NULL
  def test_missing_with_self_has_many
    Person.create!(first_name: 'test-first-name')

    assert_equal 1, Person.where.missing(:subordinates).count
  end
end
```

### DIFF
--- a/activerecord/test/cases/relation/where_chain_test.rb
+++ b/activerecord/test/cases/relation/where_chain_test.rb
@@ -202,6 +202,10 @@ module ActiveRecord
       assert_predicate Cpk::Book.where.missing(:author), :any?
     end
 
+    def test_missing_with_self_reference_has_many
+      assert_equal 1, Author.where(id: 2).where.missing(:favorite_authors).count
+    end
+
     def test_not_inverts_where_clause
       relation = Post.where.not(title: "hello")
       expected_where_clause = Post.where(title: "hello").where_clause.invert
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -126,7 +126,7 @@ module ActiveRecord
           reflection = scope_association_reflection(association)
           @scope.left_outer_joins!(association)
           association_conditions = Array(reflection.association_primary_key).index_with(nil)
-          if reflection.options[:class_name]
+          if reflection.options[:class_name] || @scope.table_name == reflection.table_name
             @scope.where!(association => association_conditions)
           else
             @scope.where!(reflection.table_name => association_conditions)


activerecord/test/models/person.rb
```diff
@@ -11,6 +11,7 @@ class Person < ActiveRecord::Base
                                     through: :readers, source: :post
 
   has_many :friendships, foreign_key: "friend_id"
+  has_many :friends, through: :friendships
   # friends_too exists to test a bug, and probably shouldn't be used elsewhere
   has_many :friends_too, foreign_key: "friend_id", class_name: "Friendship"
   has_many :followers, through: :friendships
```
skipkayhil: Could `followers` be used so that we aren't adding a relationship that's used for a single test?

alexcameron89: Ahh, yeah I meant to remove this change since I ended up finding an existing relationship with `Author.where.missing(:favorite_authors)`.

~I'll make that change real quick.~
Done in https://github.com/rails/rails/pull/55138/commits/3a512bd563f1fd3c5ea1662fe469be53ec9482b8 👍

activerecord/test/cases/relation/where_chain_test.rb
```diff
@@ -202,6 +202,10 @@ def test_missing_with_composite_primary_key
       assert_predicate Cpk::Book.where.missing(:author), :any?
     end
 
+    def test_missing_with_self_reference_has_many
+      assert_equal 1, Author.where(id: 2).where.missing(:favorite_authors).count
+    end
+
```
shivabhusal, line 208: Nice work! 
Should we also mention this fix in the changelog?

One more thought - rather than only checking count, it might be safer to create fresh records and assert the returned record IDs. I think counts can sometimes be deceptive.

PR #55139, opened 2025-06-02
Always fully clear current attributes

byroot:
Attempting to reset instances is a recipe for causing state leak.

Regular instance variables in `Current` subclasses aren't cleared, so unless you remember to undefine them in the `reset` callback they will leak across requests.

We could of course document that, but it's just too much of a footgun, and I see no benefit of trying to re-use that object instance.

It's much safer to just always drop it and start fresh.

FYI: @janko 

Closes: https://github.com/rails/rails/pull/55125/


franzliedke:
I can't give this enough positive emojis! 💯

PR #55140, opened 2025-06-02
Load tests prior to when Minitest load plugins:

Edouard-chin:
### Motivation / Background

Fix https://github.com/rails/rails/pull/54741#issuecomment-2922863186

#### Problem

Custom minitest plugins that require tests to be loaded (such as minitest-focus) may malfunction.
This is due to a behaviour change introduced in #54741.

#### Context

Rails now load the test thanks to a custom plugin, but this creates a caveat: Any plugins running prior than ours may need the tests to be loaded.

The order of thing in a normal minitest setup is:

1) Test files are loaded
2) Minitest load plugins
3) Minitest initialize plugins

Before this patch it was:

1) Minitest load plugins
2) Minitest initialize plugins (This is where some plugins may require tests to be loaded)
3) Our plugin loads the test files

The plugins are loaded/initialized by Minitest in a random order.

#### Solution

I added a OptionsParser callback which get called at the very end of the parsing process and is responsible to load the test files. The order of thing is now:

1) Minitest load plugins
2) Test files are loaded
3) Minitest initialize the plugins

### Checklist

cc/ @issyl0

### DIFF
--- a/railties/lib/minitest/rails_plugin.rb
+++ b/railties/lib/minitest/rails_plugin.rb
@@ -134,6 +134,12 @@ module Minitest
 
     options[:color] = true
     options[:output_inline] = true
+
+    opts.on do
+      if ::Rails::TestUnit::Runner.load_test_files
+        ::Rails::TestUnit::Runner.load_tests(options.fetch(:test_files, []))
+      end
+    end
   end
 
   # Owes great inspiration to test runner trailblazers like RSpec,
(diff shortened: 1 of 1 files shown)


byroot:
Can @issyl0 confirms whether this fixes their issue?

issyl0:
@byroot I'll check today! Thanks for the ping, and thanks @Edouard-chin for trying to fix it!

issyl0:
It works! We can use `focus` again in our large app ([and in my small repro app](https://github.com/issyl0/rails-54741-focus-repro/commit/8b3dd657c66a2b69f74dbf8185f2de257a72725f)). Thank you!

Issue #54444, opened 2025-02-05
ActiveJob's `retry_job` causes incorrect instrumentation when passed in parameters


ducmtran:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"
end

require "active_job/railtie"
require "minitest/autorun"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.eager_load = false
  config.secret_key_base = "secret_key_base"
  config.active_job.queue_adapter = :test

  config.logger = Logger.new($stdout)
end
Rails.application.initialize!

class FooJob < ActiveJob::Base
  queue_as :high

  def perform(val:)
    retry_job(queue: :low) if queue_name.to_s == 'high'
  end
end

ActiveSupport::Notifications.monotonic_subscribe('perform.active_job') do |event|
  job = event.payload[:job]
  p "queue_name: #{job.queue_name}"
end

class FooJobTest < ActiveJob::TestCase
  def test_stuff
    FooJob.perform_later(val: 1)
    perform_enqueued_jobs
    perform_enqueued_jobs
  end
end
```

```
  INFO -- : Enqueued FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) to Test(high) with arguments: {:val=>1}
  INFO -- : Performing FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) from Test(high) enqueued at 2025-02-05T23:45:25.162898000Z with arguments: {:val=>1}
  INFO -- : Enqueued FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) to Test(low) with arguments: {:val=>1}
  INFO -- : Retrying FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) after 1 attempts in 0 seconds.
"queue_name: low"
  INFO -- : Performed FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) from Test(low) in 1.44ms
  INFO -- : Performing FooJob (Job ID: fa710ac3-5a45-4e85-b957-ff05876626aa) from Test(low) enqueued at 2025-02-05T23:45:25.164454000Z with arguments: {:val=>1}
"queue_name: low"
[...]

PR #55141, opened 2025-06-02
Keep the original job object when using `retry_job`

### LINKED ISSUE #55116: Error in `before_enqueue` callback does not persist job state
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "mocha"
end

require "active_job/railtie"
require "minitest/autorun"
require "mocha/minitest" 

  config.eager_load = false

end
Rails.application.initialize!

class FooJob < ActiveJob::Base
  retry_on StandardError, queue: :default, attempts: 10

  before_enqueue do
    puts "before_enqueue result: #{@result}"
    puts "before_enqueue count: #{@count}"
    if @result.present?
      @result = ""
      save_progress
    end
  rescue
    @count = 0
    raise
  end

  def perform
    @result ||= ""
    @count ||= 0
    (3-@count).times do |i|
      @result += "x "
      puts "perform result: #{@result}"
      interrupt if @count == 1
      @count += 1
    end
  end

  def save_progress
    puts "saving!"
  end

  def interrupt
    puts "interrupt"
    retry_job
  end

  def deserialize(job_data)
    @count = job_data["count"]
    @result = job_data["result"]
    super
  end

  def serialize
    super.merge("count" => @count, "result" => @result)
  end
end

class FooJobTest < ActiveJob::TestCase
  def test_job_state_is_preserved_during_retry
    FooJob.any_instance.stubs(:save_progress).raises(StandardError, 'failed to save progress')

    assert_nothing_raised do
      FooJob.perform_later
      perform_enqueued_jobs
    end
  end
end
```

### Expected behavior

When job gets "interrupted" and re-enqueued at count = 1, the `before_enqueue` block will reset the count to 0 when an exception occurs.  

```
I, [2025-05-24T21:55:16.770459 #88347]  INFO -- : ----------------------------------------------------
I, [2025-05-24T21:55:16.770492 #88347]  INFO -- : FooJobTest: test_job_state_is_preserved_during_retry
I, [2025-05-24T21:55:16.770498 #88347]  INFO -- : ----------------------------------------------------
before_enqueue result: 
before_enqueue count: 
I, [2025-05-24T21:55:16.771251 #88347]  INFO -- : Enqueued FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) to Test(default)
I, [2025-05-24T21:55:16.772177 #88347]  INFO -- : Performing FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) from Test(default) enqueued at 2025-05-25T01:55:16.771216000Z
perform result: x 
perform result: x x 
interrupt
before_enqueue result: x x 
before_enqueue count: 1
E, [2025-05-24T21:55:16.772340 #88347] ERROR -- : Failed enqueuing FooJob to Test(default): StandardError (failed to save progress)
I, [2025-05-24T21:55:16.772355 #88347]  INFO -- : Retrying FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) after 1 attempts in 0 seconds.
before_enqueue result: 
before_enqueue count: 0
I, [2025-05-24T21:55:16.772495 #88347]  INFO -- : Enqueued FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) to Test(default) at 2025-05-25 01:55:20 UTC
I, [2025-05-24T21:55:16.772508 #88347]  INFO -- : Retrying FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) after 1 attempts in 3 seconds, due to a StandardError (failed to save progress).
I, [2025-05-24T21:55:16.772524 #88347]  INFO -- : Performed FooJob (Job ID: f64ef6c8-54ab-41dd-a5fd-7d69693396ba) from Test(default) in 0.36ms
```

### Actual behavior

`before_enqueue` does not reset the count to 0.

```
before_enqueue result: 
before_enqueue count: 
I, [2025-05-24T22:10:24.671247 #6354]  INFO -- : Enqueued FooJob (Job ID: 813471e0-97b4-4e32-b499-ca5434c3a036) to Test(default)
I, [2025-05-24T22:10:24.672472 #6354]  INFO -- : Performing FooJob (Job ID: 813471e0-97b4-4e32-b499-ca5434c3a036) from Test(default) enqueued at 2025-05-25T02:10:24.671218000Z
perform result: x 
perform result: x x 
interrupt
before_enqueue result: x x 
before_enqueue count: 1
E, [2025-05-24T22:10:24.672622 #6354] ERROR -- : Failed enqueuing FooJob to Test(default): StandardError (failed to save progress)
I, [2025-05-24T22:10:24.672636 #6354]  INFO -- : Retrying FooJob (Job ID: 813471e0-97b4-4e32-b499-ca5434c3a036) after 1 attempts in 0 seconds.
before_enqueue result: x x 
before_enqueue count: 1
E, [2025-05-24T22:10:24.672780 #6354] ERROR -- : Failed enqueuing FooJob to Test(default): StandardError (failed to save progress)
```

This is because now in `retry_job`, we first dup the job and enqueue the duped job. When an error is raised in `before_enqueue`, any modifications to the job in the `before_enqueue` will be lost since the `retry_job` after will dup the previous job instance and not the duped job instance.

### System configuration

Rails main, after https://github.com/rails/rails/pull/54447

Note: the interrupting feature here is simulating what the [job-iteration](https://github.com/Shopify/job-iteration) gem does, it allows the job to be interrupted and retried.

cc: @fatkodima

etiennebarrie:
Fix https://github.com/rails/rails/issues/55116
Follow-up to https://github.com/rails/rails/pull/54447
cc @jenshenny @fatkodima @byroot 

When the job is retried, we want to preserve the original `scheduled_at`, `queue_name` and `priority`. Instead of using `dup`, this copies and restores the necessary attributes.

---

As described in the issue, previously the same job object was used when enqueuing a retry, which allowed to mutate the state of the object from `before_enqueue` and then raise an error which was rescued with `retry_on`, which was enqueuing another retry. During that second retry's `before_enqueue` callback, the changes from the first callback would be available.

By switching to another instance with the call to `dup`, we lost that ability. As far as I can tell, based on #54444, what we want is the attributes changed by `set` to be restored.

In terms of implementation, this adds some coupling between `retry_job` and `set` (knowledge of the changed attributes).
An alternative would be to introduce a new API, maybe `set(**options, &block)` which would restore the attributes. Instead of calling `enqueue options`, we would call `set(options) { enqueue }`, and the attributes would be restored by `set` after the block is called.

### DIFF
--- a/activejob/lib/active_job/exceptions.rb
+++ b/activejob/lib/active_job/exceptions.rb
@@ -157,8 +157,10 @@ module ActiveJob
     #  end
     def retry_job(options = {})
       instrument :enqueue_retry, options.slice(:error, :wait) do
-        job = dup
-        job.enqueue options
+        scheduled_at, queue_name, priority = self.scheduled_at, self.queue_name, self.priority
+        enqueue options
+      ensure
+        self.scheduled_at, self.queue_name, self.priority = scheduled_at, queue_name, priority
       end
     end
 
--- a/activejob/test/cases/exceptions_test.rb
+++ b/activejob/test/cases/exceptions_test.rb
@@ -2,6 +2,7 @@
 
 require "helper"
 require "jobs/retry_job"
+require "jobs/retries_job"
 require "jobs/after_discard_retry_job"
 require "models/person"
 require "minitest/mock"
@@ -341,6 +342,13 @@ class ExceptionsTest < ActiveSupport::TestCase
       end
 
       assert_equal ["Raised DefaultsError for the 5th time"], JobBuffer.values
+    end
+
+    test "retrying a job when before_enqueue raised uses the same job object" do
+      job = RetriesJob.new
+      assert_nothing_raised do
+        job.enqueue
+      end
     end
 
     test "retrying a job reports error when report: true" do
--- a/activejob/test/jobs/retries_job.rb
+++ b/activejob/test/jobs/retries_job.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class BeforeEnqueueError < StandardError; end
+
+class RetriesJob < ActiveJob::Base
+  attr_accessor :raise_before_enqueue
+
+  # The job fails in before_enqueue the first time it retries itself.
+  before_perform do
+    self.raise_before_enqueue = true
+  end
+
+  # The job fails once to enqueue/retry itself, then succeeds.
+  before_enqueue do
+    raise BeforeEnqueueError if raise_before_enqueue
+  ensure
+    @raise_before_enqueue = false
+  end
+
+  # The job retries on BeforeEnqueueError errors.
+  retry_on BeforeEnqueueError
+
+  def perform
+    retry_job if executions <= 1
+  end
+end


byroot:
So this is essentially a revert?

etiennebarrie:
> So this is essentially a revert?

No the test added in #54447 and the fixed behavior around logging is still preserved. This fixes a new issue with that implementation, adding a test that used to pass before #54447.

peter-evans:
This change fixes the issue for us at GitHub. 👍

byroot:
> No the test added in #54447 and the fixed behavior around logging is still preserved.

Right.

Not sure why, but I guess I like the idea of `retry` starting with a clone, but if the original issue is still fixed, no reason to be too attached to it.

PR #55147, opened 2025-06-03
Pin GitHub Actions versions

hlascelles:
After recent [supply chain attacks](https://www.bleepingcomputer.com/news/security/github-action-hack-likely-led-to-another-in-cascading-supply-chain-attack/) on widely used GitHub Actions, it is now [security best practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions) to pin GitHub Actions versions in workflow files.

This PR makes that change, using the latest versions that the current major versions in code imply.

### Checklist

### DIFF
--- a/.github/workflows/devcontainer-shellcheck.yml
+++ b/.github/workflows/devcontainer-shellcheck.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout (GitHub)
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@v4
 
       - name: Lint Devcontainer Scripts
         run: |
--- a/.github/workflows/devcontainer-shellcheck.yml
+++ b/.github/workflows/devcontainer-shellcheck.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout (GitHub)
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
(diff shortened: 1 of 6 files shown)


byroot:
> it is now [security best practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions) to pin GitHub Actions versions in workflow files.

I entirely agree for third party actions, but not for the ones provided by GitHub itself. I know one could argue that it's possible the official actions repos could be compromised without the entirety of GitHub Actions as a platform being compromised, but still.

Anyway, thanks for your PR.

hlascelles:
OK, understood... Some are GitHub, some are "big hitters" (rubygems). One other is smaller, `zzak/action-discord` (who may be a Ruby or Rails employee I think!). I can reduce the PR to third parties, or just the small one...

More generally wouldn't we want to pin versions for stability reasons, to stop builds from suddenly failing "for no reason" if a third party does an action update deploy?

I understand that is also a more general question for the whole community!

byroot:
> I can reduce the PR to third parties, or just the small one...

Yes, that would be acceptable. I consider `github/*` and `ruby/*` to be highly trusted and frequently updated, so I don't want them pinned.

For others 👍 .

> to stop builds from suddenly failing "for no reason" if a third party does an action update deploy?

That's not really something that ever happens. Overall it's much less trouble than having to merge very frequent dependabot PRs to keep actions in sync. Especially for `ruby/` as it's quite frequently updated to deal with various ecosystem changes.

hlascelles:
Hi @byroot. I've pushed changes to the branch to resolve this, and only pin one SHA. However, since this PR is closed the new commits do not show up. 

Do you want to reopen this PR (since the changes are on the same branch?), rather than me opening a new PR? I can squash the commits if happy. Thanks!

PR #55148, opened 2025-06-03
[#55146] Fix tests for Sidekiq integration queueing.

ruyrocha:
### Motivation / Background

This Pull Request has been created because of #55146.

### Detail

This Pull Request changes Sidekiq integration queueing test, as the wrapper class has changed in Sidekiq v8.

### Checklist

### DIFF
--- a/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -18,21 +18,15 @@ module ActiveJob
     #
     #   Rails.application.config.active_job.queue_adapter = :sidekiq
     class SidekiqAdapter < AbstractAdapter
-      def initialize(*) # :nodoc:
-        @stopping = false
+      @stopping = false
 
-        Sidekiq.configure_server do |config|
-          config.on(:quiet) { @stopping = true }
-        end
+      callback = -> { @stopping = true }
 
-        Sidekiq.configure_client do |config|
-          config.on(:quiet) { @stopping = true }
-        end
-      end
+      Sidekiq.configure_client { |config| config.on(:quiet, &callback) }
+      Sidekiq.configure_server { |config| config.on(:quiet, &callback) }
 
-      delegate :enqueue, :enqueue_at, :enqueue_all, to: Sidekiq::ActiveJob::QueueAdapters::SidekiqAdapter
-
-      def enqueue_after_transaction_commit? = true
+      delegate :enqueue, :enqueue_at, :enqueue_all, :stopping?,
+        :enqueue_after_transaction_commit?, to: Sidekiq::ActiveJob::QueueAdapters::SidekiqAdapter
 
       JobWrapper = Sidekiq::ActiveJob::Wrapper
     end
--- a/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -30,58 +30,11 @@ module ActiveJob
         end
       end
 
-      def enqueue(job) # :nodoc:
-        job.provider_job_id = JobWrapper.set(
-          wrapped: job.class,
-          queue: job.queue_name
-        ).perform_async(job.serialize)
-      end
+      delegate :enqueue, :enqueue_at, :enqueue_all, to: Sidekiq::ActiveJob::QueueAdapters::SidekiqAdapter
 
-      def enqueue_at(job, timestamp) # :nodoc:
-        job.provider_job_id = JobWrapper.set(
-          wrapped: job.class,
-          queue: job.queue_name,
-        ).perform_at(timestamp, job.serialize)
-      end
+      def enqueue_after_transaction_commit? = true
 
-      def enqueue_all(jobs) # :nodoc:
-        enqueued_count = 0
-        jobs.group_by(&:class).each do |job_class, same_class_jobs|
-          same_class_jobs.group_by(&:queue_name).each do |queue, same_class_and_queue_jobs|
-            immediate_jobs, scheduled_jobs = same_class_and_queue_jobs.partition { |job| job.scheduled_at.nil? }
-
-            if immediate_jobs.any?
-              jids = Sidekiq::Client.push_bulk(
-                "class" => JobWrapper,
-                "wrapped" => job_class,
-                "queue" => queue,
-                "args" => immediate_jobs.map { |job| [job.serialize] },
-              )
-              enqueued_count += jids.compact.size
-            end
-
-            if scheduled_jobs.any?
-              jids = Sidekiq::Client.push_bulk(
-                "class" => JobWrapper,
-                "wrapped" => job_class,
-                "queue" => queue,
-                "args" => scheduled_jobs.map { |job| [job.serialize] },
-                "at" => scheduled_jobs.map { |job| job.scheduled_at&.to_f }
-              )
-              enqueued_count += jids.compact.size
-            end
-          end
-        end
-        enqueued_count
-      end
-
-      class JobWrapper # :nodoc:
-        include Sidekiq::Worker
-
-        def perform(job_data)
-          Base.execute job_data.merge("provider_job_id" => jid)
-        end
-      end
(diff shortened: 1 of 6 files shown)


activejob/test/integration/queuing_test.rb
```diff
@@ -48,17 +56,15 @@ class QueuingTest < ActiveSupport::TestCase
     end
 
     test "should interrupt jobs" do
-      ContinuableTestJob.perform_later @id
-      wait_for_jobs_to_finish_for(1.seconds)
+      ContinuableTestJob.perform_later(@id)
+      assert wait_until(timeout: 1) { continuable_job_started }
```
djmb: This check needs to happen after the job manager is stopped - we are confirming that the job was interrupted while in progress. The tests pass now, but that's because they are not testing the right behaviour any more.

The test fails sometimes because depending on the order of paths in the load path sometimes we run it with the Sidekiq adapter from Rails which implements the new `stopping?` method and sometimes with the adapter from the Sidekiq repo which doesn't.

I think we need to add a check that the Sidekiq adapter has its own implementation of the `stopping?` method so that we don't bother with the test if we pull in the version from the Sidekiq repo.

Then once the sidekiq repo is updated we'll automatically run the test on it.

Something like:
```ruby
  if ActiveJob::QueueAdapters::SidekiqAdapter.public_instance_methods(false).include?(:stopping?)
    test "should interrupt jobs" do
      ContinuableTestJob.perform_later @id
      wait_for_jobs_to_finish_for(1.seconds)

      jobs_manager.stop_workers
      wait_for_jobs_to_finish_for(1.seconds)
      assert_not job_executed
      assert continuable_job_started

      jobs_manager.start_workers
      wait_for_jobs_to_finish_for(10.seconds)
      assert job_executed
    end
  end
```

ruyrocha: This is what happens if I run only that test (caught while monitoring Redis):

```
1749217017.142675 [1 192.168.100.14:33274] "zadd" "retry" "1749217033.1421435" "{\"retry\":true,\"queue\":\"integration_tests\",\"wrapped\":\"ContinuableTestJob\",\"args\":[{\"job_class\":\"ContinuableTestJob\",\"job_id\":\"b7a36603-7c87-4ab0-9c1e-d28a976cb5f7\",\"provider_job_id\":\"277264cc70c4561e6a04577d\",\"queue_name\":\"integration_tests\",\"priority\":null,\"arguments\":[\"AJ-2e20db9d-fd82-4675-91bb-84dcb8618107\"],\"executions\":1,\"exception_executions\":{\"[ActiveJob::Continuation::AfterAdvancingError]\":1},\"locale\":\"en\",\"timezone\":\"UTC\",\"enqueued_at\":\"2025-06-06T13:36:49.313127266Z\",\"scheduled_at\":\"2025-06-06T13:36:52.584316272Z\",\"continuation\":{\"completed\":[\"step_one\"]}}],\"class\":\"Sidekiq::ActiveJob::Wrapper\",\"jid\":\"23b527b88dd7277df0106a29\",\"created_at\":1749217009313,\"enqueued_at\":1749217016871,\"error_message\":\"undefined method 'stopping?' for an instance of ActiveJob::QueueAdapters::SidekiqAdapter\",\"error_class\":\"NoMethodError\",\"failed_at\":1749217017141,\"retry_count\":0}"
```

So everything should be set once Sidekiq is updated with [this PR](https://github.com/sidekiq/sidekiq/pull/6727), right?

This other test started to fail (again) after check for `stopping?` was introduced conditionally on `should interrupt jobs` test: https://github.com/rails/rails/blob/4b7fb1d14b954d2348d2065ff955466479509656/activejob/test/integration/queuing_test.rb#L34-L41

This is the new failure running tests against sidekiq's main branch:
```
1749219039.638569 [1 192.168.100.14:44028] "zadd" "retry" "1749219059.638016" "{\"retry\":true,\"queue\":\"integration_tests\",\"wrapped\":\"ContinuableTestJob\",\"args\":[{\"job_class\":\"ContinuableTestJob\",\"job_id\":\"b7a36603-7c87-4ab0-9c1e-d28a976cb5f7\",\"provider_job_id\":\"23b527b88dd7277df0106a29\",\"queue_name\":\"integration_tests\",\"priority\":null,\"arguments\":[\"AJ-2e20db9d-fd82-4675-91bb-84dcb8618107\"],\"executions\":2,\"exception_executions\":{\"[ActiveJob::Continuation::AfterAdvancingError]\":2},\"locale\":\"en\",\"timezone\":\"UTC\",\"enqueued_at\":\"2025-06-06T14:09:58.403628846Z\",\"scheduled_at\":\"2025-06-06T14:10:01.744841301Z\",\"continuation\":{\"completed\":[\"step_one\",\"step_two\"],\"current\":[\"step_three\",null]}}],\"class\":\"Sidekiq::ActiveJob::Wrapper\",\"jid\":\"6a843da7ab641a00af9e37fe\",\"created_at\":1749218998404,\"enqueued_at\":1749219037622,\"error_message\":\"No such file or directory @ rb_file_s_rename - (/tmp/d20250606-111254-647etu/dummy/tmp/AJ-2e20db9d-fd82-4675-91bb-84dcb8618107.started, /tmp/d20250606-111254-647etu/dummy/tmp/AJ-2e20db9d-fd82-4675-91bb-84dcb8618107)\",\"error_class\":\"Errno::ENOENT\",\"failed_at\":1749219002497,\"retry_count\":1,\"retried_at\":1749219039637}"
```

Also, the results are still different when I run all integration tests opposed to running a single one.

It's weird, but I had to manually declare `stopping?` on Sidekiq:
image

djmb: [Rails](https://github.com/rails/rails/blob/main/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L20) and [Sidekiq](https://github.com/sidekiq/sidekiq/blob/main/lib/active_job/queue_adapters/sidekiq_adapter.rb) both have an implementation of the adapter which one is used by the tests is not consistent. It depends on the order of the active_job and sidekiq libs in the load path. 

I'm finding that activejob is first when running `bundle exec rake` and sidekiq when running `bin/test`.

This test only works right now with the Rails version of the adapter. And that sidekiq PR doesn't implement the `stopping?` method, it just ensures that we fall back to the version on the abstract adapter the always returns false.

We need to update the Sidekiq version of the adapter but I want to give it a little while in case we want to change the API.

Sounds also like some state is leaking between the integration tests - have you tried different seeds to set if you can determine a load order that triggers them?

zzak:
We shouldn't have to change the adapter on our end, that file is there because Rails supports more versions than just Sidekiq 8.whatever, and will be deprecated in a future version once everyone has time to update their sidekiq.

If you're using a newer version of Sidekiq which changes the JobWrapper name, and writes that new name to Redis, we shouldn't get in the way.

activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -18,70 +18,17 @@ module QueueAdapters
...
-      end
-
-      class JobWrapper # :nodoc:
-        include Sidekiq::Worker
-
-        def perform(job_data)
-          Base.execute job_data.merge("provider_job_id" => jid)
-        end
-      end
+      JobWrapper = Sidekiq::ActiveJob::Wrapper
```
zzak: This constant was only just introduced in sidekiq/sidekiq#6430, which seems to have made it in sidekiq [v7.3.3](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#733).

But Rails aims to maintain compatibility with older versions of Sidekiq, unless there is a good reason to break that guarantee and update the dependency requirement on the gem:
https://github.com/rails/rails/blob/5a7a59ab4a9ac34b8cadf9d8ce7b554da50c5e43/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L3

However, because we plan to deprecate and remove this adapter after giving users time to upgrade to a version of Sidekiq that contains the upstream adapter, this change is not acceptable.

ruyrocha: Got it, @zzak! I just pinged you with my 2 cents on #55146 for better visibility.

zzak:
This was fixed upstream, there is no need for Rails to modify the adapter now.

PR #55151, opened 2025-06-04
Follow up on Active Job Continuations PR feedback

djmb:
Addressing feedback from @byroot on #55127 

- Make `ActiveJob::Instrumentation#instrument` public with :nodoc: to avoid send
- Rescue `NoMethodError` to avoid `responds_to?` calls in `Step#advance!`
- Inline test jobs into the continuation test case


byroot:
Thank you!

PR #53623, opened 2024-11-13
Prevent Active Storage Blob from autosaving Attachments:

Edouard-chin:
Fix: #50800,
Fix: #53451

### Motivation / Background

This commit is meant to fix a behaviour change introduced in #50800, which causes issues reported in #53451 and #53452.

### Detail

There are a lot of codepaths and callbacks happening and the issue is subtle which makes things a bit tricky to explain, but in short, adding the `inverse_of` on the `ActiveStorage::Attachment <-> Blob` relation added behavioural difference on how the Attachement get saved.

### Problem

  By adding the `ActiveStorage::Attachment <-> Blob` inverse_of in #50800, a Blob is now aware of its associated Attachment and this creates a problem: **Order of association save callbacks**

  ```ruby
  class User < ApplicationRecord
    has_one_attached :avatar

    blob = create_active_storage_blob("avatar.png")
    User.create!(avatar: blob)
  end
  ```

  ----------------------

  Before #50800, this was the order of how the records were persisted:

  
  1. Blob is created
  2. User is created
  3. Blob is updated through the User<->Blob autosave association (Active storage [modifies some metadata](https://github.com/rails/rails/blob/ad9bc2a51d5980d1c0990db97e552015410778f0/activestorage/lib/active_storage/attached/changes/create_one.rb#L12) on the Blob when creating an attachment)
  4. Attachment is created through the User<->Attachment autosave association
  
  ----------------------

  After #50800, this is the order:

  
  1-3. Similar as above
  4.   Attachment is created through the ~User~Blob<->Attachment autosave association
  
___________________________________
  The reasons this creates the issues in #53451 and #53452 are different so I'll explain them separately.

  #### Problem in #53451

  Because an Attachment now gets created through the Blob association in a [after_update](https://github.com/rails/rails/blob/ad9bc2a51d5980d1c0990db97e552015410778f0/activerecord/lib/active_record/autosave_association.rb#L198) callback that **runs prior** to the [`Blob#touch_attachments`](https://github.com/rails/rails/blob/ad9bc2a51d5980d1c0990db97e552015410778f0/activestorage/app/models/active_storage/blob.rb#L44) after_update callback, the Blob will now `touch` associated attachments which by repercusion ends up loading the record (e.g the User in the example above) and touching it.

[...]

PR #55156, opened 2025-06-04
Fix crash when trying to attach a Blob to a Record:

### LINKED ISSUE #55144: Disabling ActiveStorage.touch_attachment_records causes attaching blobs to raise ActiveRecord::ActiveRecordError (Cannot touch on a new record)
After #53623, attaching a blob to a record raises
```ActiveRecord::ActiveRecordError: Cannot touch on a new or destroyed record object. Consider using persisted?, new_record?, or destroyed? before touching.``` if ```ActiveStorage.touch_attachment_records``` is false.

Reverting the commit (removing **autosave: false** from **has_many :attachments**) or setting **ActiveStorage.touch_attachment_records = true** fixes it.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
    source "https://rubygems.org"

    gem "rails", "~> 8.0.2"
    #gem "rails", github: "rails/rails", branch: "main"

    gem "sqlite3"
end

require "minitest/autorun"

    config.root = __dir__
    config.hosts << "example.org"
    config.eager_load = false

    Rails.logger  = config.logger

        local: {
                root: Dir.tmpdir,
                service: "Disk"
               }
    }

    # setting it to true (the default), fixes it
    config.active_storage.touch_attachment_records = false

end
Rails.application.initialize!

end

    has_one_attached :tmpfile
    has_many_attached :files
end

    def test_attachment_create_on_has_one
        data = "Something else entirely!"
        io = StringIO.new(data)
        blob = ActiveStorage::Blob.create_before_direct_upload! filename: "tmp.txt", byte_size: data.size, checksum: OpenSSL::Digest::MD5.base64digest(data)
        blob.upload(io)

        user = User.create!(
            tmpfile: blob.signed_id
        )

        assert(user.tmpfile.attached?)
    end

    def test_attachment_create_on_has_one_with_attach
        data = "Something else entirely!"
        io = StringIO.new(data)
        blob = ActiveStorage::Blob.create_before_direct_upload! filename: "tmp2.txt", byte_size: data.size, checksum: OpenSSL::Digest::MD5.base64digest(data)
        blob.upload(io)

        user = User.create!

        assert_nothing_raised {
            user.tmpfile.attach(blob)
        }

        assert(user.tmpfile.attached?)
    end

    def test_attachment_create_on_has_many
        data = "Something else entirely!"
        io = StringIO.new(data)
        blob = ActiveStorage::Blob.create_before_direct_upload! filename: "test.txt", byte_size: data.size, checksum: OpenSSL::Digest::MD5.base64digest(data)
        blob.upload(io)

        user = User.create!(
            files: blob.signed_id
        )

        assert(user.files.attached?)
        assert_equal(1, user.files.count)
    end
end
```

### Expected behavior

Record to be saved and blob to be attached, with corresponding ```Attachment``` being auto created, irrespective of touching the parent record.

### Actual behavior

```ActiveRecord::ActiveRecordError``` is raised. It seems ```Attachment``` for the blob isn't being saved, hence the ```:touch_attachments``` callback tries to touch non-persisted instances.

### System configuration

**Rails version**: 8.0.2 and ```main```

**Ruby version**: 3.3.7

### Notes

I think **ActiveStorage.touch_attachment_records** wasn't supposed to affect this, and only apply to whether touch the record's/parent's model or not? Within the cited PR, responsibilities of the ```Attachment``` creation seems to be changed, but not sure how it applies here.

Edouard-chin:
### Motivation / Background

- Fix #55144

#### Problem

Setting `ActiveStorage.touch_attachment_records` to `false` will make the application crash if it uses the Active Storage direct upload feature.

#### Context

When an existing Blob is attached to a Record, Active Storage tries to `touch` its Attachment, but the Attachment is not yet persisted, resulting in an exception.

Before #53623, a Blob was responsible to autosave its Attachment, which would result in:

1) Creating a Blob
2) Autosaving its Attachment
3) Touching the just saved Attachment

We were basically making an unnecessary SQL query in 3).

This problem only happens when `ActiveStorage.touch_attachment_records` is set to false. When it's set to true (the default), the bug doesn't reproduce because of [this line](https://github.com/rails/rails/blob/fc55e5b812f75b70bc10c99c6a81d1bff5412438/activestorage/app/models/active_storage/blob.rb#L366) which returns an empty Relation (an Attachment doesn't yet have a Record at this stage)

#### Solution

Now that a Blob no longer autosave its Attachment, we need to manually check whether the Attachment is persisted before trying to `touch` it.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Prevent `ActiveStorage.touch_attachment_records = false` from crashing the attachment of a Blob.
+
+    When `ActiveStorage.touch_attachment_records` was set to `false`, attaching a existing Blob to a Record
+    would raise an error. This is now fixed.
+
+    *Edouard Chin*
+
 *   Parallelize `exist?` checks and uploads in `MirrorService#mirror` using
     the existing internal thread pool. With N mirrors, wall time drops from
     O(N) to O(1) network round-trips for both phases.
--- a/activestorage/app/models/active_storage/blob.rb
+++ b/activestorage/app/models/active_storage/blob.rb
@@ -433,7 +433,7 @@ class ActiveStorage::Blob < ActiveStorage::Record
             relation
           end
         end.each do |attachment|
-          attachment.touch
+          attachment.touch unless attachment.new_record?
         end
       end
     end
--- a/activestorage/test/models/attachment_test.rb
+++ b/activestorage/test/models/attachment_test.rb
@@ -53,6 +53,23 @@ class ActiveStorage::AttachmentTest < ActiveSupport::TestCase
     assert_equal(1, user.callback_counter)
   end
 
+  test "attaching a blob to a record works when ActiveStorage.touch_attachment_records is false" do
+    ActiveStorage.with(touch_attachment_records: false) do
+      data = "Something else entirely!"
+      io = StringIO.new(data)
+      blob = create_blob_before_direct_upload byte_size: data.size, checksum: ActiveStorage.checksum_implementation.base64digest(data)
+      blob.upload(io)
+
+      assert_nothing_raised do
+        User.create!(
+          name: "Roger",
+          avatar: blob.signed_id,
+          record_callbacks: true,
+        )
+      end
+    end
+  end
+
   test "attaching a record doesn't reset the previously_new_record flag" do
     @user.highlights.attach(io: ::StringIO.new("dummy"), filename: "dummy.txt")
 


soberstadt:
We ran into this bug, and this change is exactly what I was going to suggest before we found this PR. Thanks for the fix

soberstadt:
Is there anything that can be done to get this PR moving forward again? This is still an issue for us that we would like to have fixed in a release. Thanks again for proposing a fix for this!

PR #55159, opened 2025-06-05
Fix ActiveStorage filename sanitization for consecutive spaces

### LINKED ISSUE #55158: ActiveStorage S3 signature mismatch with filenames containing consecutive spaces
### Expected Behavior

File should upload successfully regardless of consecutive spaces in filename.

### Actual Behavior

Upload fails with:
The request signature we calculated does not match the signature you provided. Check your key and signing method. (Aws::S3::Errors::SignatureDoesNotMatch)

## Steps to Reproduce

Note: We experienced the error with a bucket on [Exoscale SOS](https://www.exoscale.com/object-storage/) which seems to use Ceph to host it. The issue does not occur using [Minio](min.io). Don't have an AWS account to test with original S3.

1. Configure ActiveStorage with S3 service
2. Upload a file with a filename containing consecutive spaces (e.g., "my  file.txt")
3. Observe Aws::S3::Errors::SignatureDoesNotMatch error

###  Code Example
```ruby
  has_one_attached :document
end

# This fails with signature mismatch
user = User.new
user.document.attach(
  io: File.open("path/to/file.txt"),
  filename: "my  file.txt",  # Note the double space
  content_type: "text/plain"
)
user.save!  # Raises Aws::S3::Errors::SignatureDoesNotMatch
```

### Root Cause Analysis

The issue occurs in the upload path when ActiveStorage generates the Content-Disposition header. The ActiveStorage::Filename#sanitized method doesn't normalize consecutive spaces, leading to inconsistent
encoding. It already does handle many other cases like trailing whitespaces, but not consecutive spaces.

### Proposed Solution

Normalize consecutive whitespace in ActiveStorage::Filename#sanitized` to also convert multiple spaces to single spaces:

```ruby
def sanitized
  @filename.encode(Encoding::UTF_8, invalid: :replace, undef: :replace, replace: "�")
           .strip
           .squeeze(" ")  # Normalize consecutive whitespace
           .tr("\u{202E}%$|:;/<>?*\"\t\r\n\\", "-")
end
```

## System configuration
- Rails Version: 8.1.0.alpha
- Ruby Version: 3.4.1
- Storage Service: S3

huerlisi:
Fixes S3 signature mismatch errors when filenames contain consecutive spaces. The issue occurred because Content-Disposition headers had inconsistent encoding between ASCII and UTF-8 parts when consecutive spaces were present.

Changes:
- Add squeeze(" ") in ActiveStorage::Filename#sanitized to normalize consecutive spaces into single spaces while preserving other sanitization
- Add test coverage for consecutive space normalization

This maintains existing behavior for other whitespace characters (tabs, newlines) which are still converted to dashes as before.

Fixes #55158

### DIFF
--- a/activestorage/app/models/active_storage/filename.rb
+++ b/activestorage/app/models/active_storage/filename.rb
@@ -57,7 +57,7 @@ class ActiveStorage::Filename
   #
   # Characters considered unsafe for storage (e.g. \, $, and the RTL override character) are replaced with a dash.
   def sanitized
-    @filename.encode(Encoding::UTF_8, invalid: :replace, undef: :replace, replace: "�").strip.tr("\u{202E}%$|:;/<>?*\"\t\r\n\\", "-")
+    @filename.encode(Encoding::UTF_8, invalid: :replace, undef: :replace, replace: "�").strip.squeeze(" ").tr("\u{202E}%$|:;/<>?*\"\t\r\n\\", "-")
   end
 
   # Returns the sanitized version of the filename.
--- a/activestorage/test/models/filename_test.rb
+++ b/activestorage/test/models/filename_test.rb
@@ -46,6 +46,13 @@ class ActiveStorage::FilenameTest < ActiveSupport::TestCase
     assert_equal "evil-fdp.exe", ActiveStorage::Filename.new("evil\u{202E}fdp.exe").sanitized
   end
 
+  test "sanitize normalizes consecutive spaces" do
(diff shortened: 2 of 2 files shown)


jeremy:
There's no error to fix; spaces should be faithfully preserved as given rather than removed to mask a backend bug. Would fix in Exoscale SOS; in the meantime, can normalize filenames within the app that's using that backend.

PR #55160, opened 2025-06-05
Docs: Add link to Turbo Streams docs in ActionCable guide [ci skip]

jslag:
I've been learning about Turbo Streams recently, and noted a few places in the docs for potential improvements. 

One is that the first place searches like "rails websockets" lead to is the ActionCable guide, which doesn't currently make any references to Hotwire.

If this proposed change seems in the ballpark but not quite right, I'm happy to do additional wordsmithing.

### DIFF
--- a/guides/source/action_cable_overview.md
+++ b/guides/source/action_cable_overview.md
@@ -27,6 +27,13 @@ client-side JavaScript framework and a server-side Ruby framework. You have
 access to your entire domain model written with Active Record or your ORM of
 choice.
 
+Turbo Streams, part of [Hotwire](https://hotwired.dev/), provides a higher-level
+abstraction of many common ActionCable uses, such as delivering real-time UI
+updates. Depending on an app's needs, it may not be necessary to work directly
+with ActionCable. Learn more in the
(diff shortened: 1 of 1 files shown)


zzak:
@jslag Thanks for your PR, I'm not sure this is how we want to introduce or describe the relationship between turbo and action cable in this guide.

jslag:
@zzak do you think there's any value in mentioning Turbo Streams some way or another in this doc? I'm happy to reword or rework if so. I do think that something along these lines would have helped clarify the relationship between turbo & action cable for me, so probably there are other people who could benefit as well.

zzak:
Sorry I don't have a good answer for you, I think these type of changes to the docs are being handled by the Rails Foundation.

I will make a note for them when they get to this guide, currently it's listed under "needs auditing" and there is a placeholder that includes the "Working with JavaScript in Rails". 

This information could go in either place, but it's not a strong feeling.

As an alternative to this PR, how about sharing your experience and knowledge on [discuss](https://discuss.rubyonrails.org/)? That way other developers may be able to engage in that discussion and can become a source of inspiration for future doc improvements.

PR #55165, opened 2025-06-06
Prevent Action View digestor from tracking random "render":

### LINKED ISSUE #55161: When rendering a partial, any mention of the string "render" inside a `cache` block results in the logged error message "Couldn't find template for digesting: strings/string"
### Steps to reproduce

Create a file `_partial.html.erb` with the following contents alongside the minimal reproduction shown below:

```erb
<%# _partial.html.erb %>

<% cache "key" do %>
  <div class="render flex" data-action="controller#render other-controller#click">
    Hello, world!
  </div>
<% end %>
```

Minimal reproduction:

```
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
end

require "action_view/railtie"
require "minitest/autorun"

puts "#{Rails::VERSION::STRING.to_f}"

  config.eager_load = false
  config.action_controller.perform_caching = true
  config.action_controller.enable_fragment_cache_logging = true
  config.cache_store = :null_store
  config.paths["app/views"] << File.expand_path(".")
end
Rails.application.initialize!

  # this render works fine
  def test_regular_erb_render
    render inline: <<~ERB
      <% cache "key" do %>
        <div class="render flex" data-action="controller#render other-controller#click">
          Hello, world!
        </div>
      <% end %>
    ERB

    assert true
  end

  # rendering the same code, but as nested inline render works fine
  def test_nested_inline_render
    nested_partial = <<~ERB
      <% cache "key" do %>
        <div class="render flex" data-action="controller#render other-controller#click">
          Hello, world!
        </div>
      <% end %>
    ERB

    render inline: <<~ERB, locals: { nested_partial: nested_partial }
      <%= render inline: nested_partial %>
    ERB

    assert true
  end

  # rendering the same code, but as a partial on the file system fails
  def test_file_partial
    render inline: <<~ERB
      <%= render partial: "partial" %>
    ERB

    assert false, "Check for 'Couldn't find template for digesting: flexes/flex' in log output ↑"
  end
end
```

### Expected behavior

Everything should render fine, cache is written, no error messages output.

### Actual behavior

When running the third test `test_file_partial` the logs will show an error message:

```
Couldn't find template for digesting: flexes/flex
Couldn't find template for digesting: others/other
```

If the string `render` appears anywhere in the partial, anything following will be assumed to be a template name (ie. `flex` in the above example), and Rails will attempt to digest it for the cache key.

It also occurs even if another symbol separates the two, like in the Stimulus action `turbo:before-stream-render@document->leaderboard#render` the error message will read "Couldn't find template for digesting: documents/document".  This is how I actually found the bug originally—removing the Stimulus action string in my code made the error go away.

This only occurs inside of a `cache` block, and only if the template that contains it is nested at least one level deep in partial rendering, and that partial is on the file system (as seen above it wasn't an issue when rendering an inline partial).

### System configuration

**Rails version**: 8.0

**Ruby version**: 3.4.2

Edouard-chin:
### Motivation / Background

Fix #55161

#### Problem

Caching a partial that contains the word "render" (literally, not the `render` method), would create a false positive in AV's digestor.
This results in a rather confusing log message.

```html+erb
<%# _partial.html.erb %>

<% cache "key" do %>
  <div class="render flex"></div>
<% end %>
```

The log error: "Couldn't find template for digesting: flexes/flex"

#### Solution

Tweak the regex and check for surrounding erb tags.

### Checklist

### DIFF
--- a/actionview/lib/action_view/dependency_tracker/erb_tracker.rb
+++ b/actionview/lib/action_view/dependency_tracker/erb_tracker.rb
@@ -91,7 +91,7 @@ module ActionView
 
         def render_dependencies
           dependencies = []
-          render_calls = source.split(/\brender\b/).drop(1)
+          render_calls = source.scan(/<%(?:(?:(?!<%).)*?\brender\b((?:(?!%>).)*?))%>/m).flatten
 
           render_calls.each do |arguments|
             add_dependencies(dependencies, arguments, LAYOUT_DEPENDENCY)
--- a/actionview/test/template/dependency_tracker_test.rb
+++ b/actionview/test/template/dependency_tracker_test.rb
@@ -118,6 +118,29 @@ module SharedTrackerTests
     assert_equal [], tracker.dependencies
   end
 
+  def test_finds_no_dependency_when_render_is_not_a_ruby_call
+    template = FakeTemplate.new("<div class='render foo'>", :erb)
+    tracker = make_tracker("resources/_resource", template)
+
+    assert_equal [], tracker.dependencies
+  end
+
+  def test_find_dependencies_and_respect_erb_tag_boundaries
+    template = FakeTemplate.new("<p>Hello</p> <% link_to abc %> <%= render 'single/quote' %>", :erb)
+    tracker = make_tracker("resources/_resource", template)
+
+    assert_equal ["single/quote"], tracker.dependencies
+  end
+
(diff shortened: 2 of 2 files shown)


cannikin:
That was quick! Thanks @Edouard-chin

Edouard-chin:
Thanks Aaron!

PR #51674, opened 2024-04-27
Add `Parameters#expect` to safely filter and require params

martinemde:
### Motivation / Background

I've been hunting around trying to fix the problem with the default, recommended way of handling parameters in Rails.

```ruby
user_params = params.require(:user).permit(:name, :age)
```

This is fine until someone using your app starts messing with the parameters and causing 500 errors by passing:

```
/path?user=string
```

This causes a NoMethodError because `permit` is called on `"string"`.

The recommendation is not the best way. Instead, the first statement should be written like this:

```ruby
user_params = params.permit(user: [:name, :age]).require(:user)
```

### Detail

In this PR, I propose a new method on ActionController::Parameters that safely handles params filtering in one method, cuts down on ignorable errors caused by params tampering, and improves the specificity of param types.

```ruby
params = ActionController::Parameters.new(user: { name: "Martin", age: 40 })
permitted = params.expect(user: %i[name age])
permitted.permitted?   # => true
permitted.has_key?(:name) # => true
permitted.has_key?(:age) # => true
```

In order to ensure consistent filtering of hashes vs arrays in params, `expect` introduces a new Array matching syntax using a double array `[[:attr]]`. (this example uses shorthand for all but the first example to make it more readable. The hashes returned are actually permitted ActionController::Parameter objects)

```ruby
# Array is expected with `permit`
ActionController::Parameters.new(comments: { text: "Hello" }).permit(comments: [:text]).require(:comments)
# => #<ActionController::Parameters {"text"=>"Hello"} permitted: true>
ActionController::Parameters.new(comments: [{ text: "Hello" }]).permit(comments: [:text]).require(:comments)
# => [{"text"=>"Hello"}]
Parameters.new(comments: { text: "Hello" }).permit(comments: [[:text]]).require(:comments)
# => param is missing or the value is empty: comments (ActionController::ParameterMissing)
Parameters.new(comments: [{ text: "Hello" }]).permit(comments: [[:text]]).require(:comments)
# => [{"text"=>"Hello"}]

# Array is expected with `expect`
Parameters.new(comments: { text: "Hello" }).expect(comments: [[:text]])
# => param is missing or the value is empty: comments (ActionController::ParameterMissing)
Parameters.new(comments: [{ text: "Hello" }]).expect(comments: [[:text]])
# => [{"text"=>"Hello"}]

# Hash is expected with `permit`
Parameters.new(user: { name: "Martin" }).permit(user: [:name]).require(:user)
# => {"name"=>"Martin"}
[...]

PR #55166, opened 2025-06-06
Use `Parameters#expect` in the templates for the authentication generator too

willnet:
The templates used in `rails g authentication` were still using `Parameters#permit`, while the scaffold templates were already updated to use `Parameters#expect` in https://github.com/rails/rails/pull/51674 . This change makes them consistent.

### DIFF
--- a/guides/source/security.md
+++ b/guides/source/security.md
@@ -190,7 +190,7 @@ in the database (e.g. password) for that user:
 ```ruby
 class SessionsController < ApplicationController
   def create
-    if user = User.authenticate_by(params.permit(:email_address, :password))
+    if user = User.authenticate_by(params.expect(:email_address, :password))
       start_new_session_for user
       redirect_to after_authentication_url
     else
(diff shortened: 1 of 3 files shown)


songjiz:
Here, `params.expect(:email_address, :password)` returns an array rather than permitted parameters.

```ruby
params = ActionController::Parameters.new({
  person: {
    name: "Francesco",
    age:  22,
    role: "admin"
  },
  email_address: "test@example.com",
  password: "secret"
})

params.expect(:email_address, :password)
=> ["test@example.com", "secret"]

params.expect(person: [ :name, :age ])
=> #<ActionController::Parameters {"name" => "Francesco", "age" => 22} permitted: true>

params.permit(:email_address, :password)
=> #<ActionController::Parameters {"email_address" => "test@example.com", "password" => "secret"} permitted: true>
```

willnet:
@songjiz Ah, I see. I was mistaken. It looks like we need to use `permit` here after all.

willnet:
@songjiz Thanks for pointing that out.

PR #50969, opened 2024-02-05
Support `:source_location` tag option for query log tags

fatkodima:
Follow up to #42240 (as discussed in https://github.com/rails/rails/pull/42240#issuecomment-1926446571).

We lost the ability to use the `:line` option with QueryLogs (https://github.com/rails/rails/pull/42240#issuecomment-898668464), while there is one in the marginalia. It was costly to use it before in the gem, but we made it fast (https://github.com/basecamp/marginalia/pull/138), so it is not *that* costly and people deliberately decide if they want to use it.

I would like to have the `:line` option (upd: decided to name it as `:source_location`), it is very much useful. Without it, if we have a slow db query log, for example, it is very hard to know from which part of the codebase the query was generated. Even having an `:action` component, you need to investigate the unfamiliar codebase to find out where the query was triggered.

cc @byroot

PR #55167, opened 2025-06-06
[ci skip] Add warning note to :source_location tag option

dizlv:
`:source_location` adds significant overhead to each SQL query issued from a Rails application. Although this behavior has been noted in the changelog, it is not immediately apparent when reading the official Rails application [configuration documentation](https://guides.rubyonrails.org/v8.0.2/configuring.html#config-active-record-query-log-tags).

### Motivation / Background

While profiling the application, we observed that a substantial amount of time was spent parsing thread stack frames. We discovered that this overhead is caused by the `:source_location` tag option. Eventually, we found a note [in the changelog](https://github.com/rails/rails/pull/50969/files#diff-353f4f101b1bb04938d761b6e29c88e0f147084e9e276854a77ae5d9a9a3b2a8R1), but it is not necessarily the most discoverable information. In fact, we might not have noticed this issue at all without addressing performance issues in production.

### Detail
This pull request adds a WARNING note to the configuration documentation to highlight this potential performance impact.

### Additional information
If the code path issues multiple SQL queries, the system will attempt to parse the source location for each query. In our case, the total time spent in this process accounts for approximately 20-30% of the overall execution time.

Screenshot 2025-06-06 at 16 32 31

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -1488,6 +1488,8 @@ Define an `Array` specifying the key/value tags to be inserted in an SQL comment
 `[ :application, :controller, :action, :job ]`. The available tags are: `:application`, `:controller`,
 `:namespaced_controller`, `:action`, `:job`, and `:source_location`.
 
+WARNING: Calculating the caller location via `:source_location` is a costly operation and should be used primarily in development (note, there is also a `config.active_record.verbose_query_logs` that serves the same purpose) or occasionally on production for debugging purposes.
+
 #### `config.active_record.query_log_tags_format`
 
 A `Symbol` specifying the formatter to use for tags. Valid values are `:sqlcommenter` and `:legacy`.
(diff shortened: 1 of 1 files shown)


guides/source/configuring.md
```diff
@@ -1488,6 +1488,8 @@ Define an `Array` specifying the key/value tags to be inserted in an SQL comment
 `[ :application, :controller, :action, :job ]`. The available tags are: `:application`, `:controller`,
 `:namespaced_controller`, `:action`, `:job`, and `:source_location`.
 
+WARNING: Calculating the caller location via `:source_location` is a costly operation and should be used primarily in development (note, there is also a `config.active_record.verbose_query_logs` that serves the same purpose) or occasionally on production for debugging purposes.
```
skipkayhil:
```suggestion
WARNING: Calculating the `:source_location` of a query can be slow, so you should consider its impact if using it in a production environment.
```

This is super subjective, but I think we can be more concise (more descriptive and less prescriptive). What do you think about this?

skipkayhil:
Thanks!

PR #55172, opened 2025-06-08
Alias `allow_nil` and `allow_blank` to `ignore_if_nil` and `ignore_if_blank`

jeromedalbert:
### Motivation / Background

This Pull Request has been created because some users find the names of the `allow_blank` and `allow_nil` validation options to be unintuitive.

I realized this at work when reviewing a PR that looked like this:

```ruby
class Fruit < ApplicationRecord
  validates :name, presence: true
  validates :name, inclusion: { in: %w[apple banana] }, if: -> { name.present? }
end
```

I mentioned that this could be a use case for `allow_blank`:

```diff
class Fruit < ApplicationRecord
  validates :name, presence: true
-  validates :name, inclusion: { in: %w[apple banana] }, if: -> { name.present? }
+  validates :name, inclusion: { in: %w[apple banana] }, allow_blank: true
end
```

My coworker did not have any strong feelings either way, but mentioned that `allow_blank` makes them think that the field can be blank from a wording perspective. It didn't feel very clear to them.

While the intended Rails meaning is "allow the validation to pass (aka ignore or skip it) if the field is blank", I think that my coworker had a point: this is not the most intuitive wording. I have been using `allow_blank` for years, so my brain is used to it and I do not think twice about it, but I could see how it could be confusing at first with a pair of fresh eyes.

A name like `ignore_validation_if_field_is_blank: true` would be extra clear albeit clunky, so more realistically we thought that a name like `ignore_if_blank` would be nice. Some other names could be `skip_if_blank` or `if_present`.

### Detail

This Pull Request adds `ignore_if_nil` and `ignore_if_blank` as aliases for the `allow_nil` and `allow_blank` validation options. If those aliases are present, they are internally converted to `allow_nil` and `allow_blank` keys.

While different people may have different opinions about this, there is precedent in the Rails codebase for more intuitive aliases, such as ActiveRecord's `t.references`/`t.belongs_to` and `add_reference`/`add_belongs_to`.

If the maintainers think that some form of this PR is worth merging, the implementation details could be tweaked, or different names like `skip_if_nil`/`skip_if_blank` or `if_present`/`if_not_nil` could be used based on preference.

### Checklist

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Add `:ignore_if_nil` and `:ignore_if_blank` validation aliases.
+
+    They are aliases for the `:allow_nil` and `:allow_blank` validation options.
+
+    *Jerome Dalbert*
+
 *   Add `except_on:` option for validation callbacks.
 
     *Ben Sheldon*
--- a/activemodel/lib/active_model/validations.rb
+++ b/activemodel/lib/active_model/validations.rb
@@ -75,7 +75,9 @@ module ActiveModel
       #   <tt>except_on: :custom_validation_context</tt> or
       #   <tt>except_on: [:create, :custom_validation_context]</tt>)
       # * <tt>:allow_nil</tt> - Skip validation if attribute is +nil+.
+      #   Also aliased as `:ignore_if_nil`.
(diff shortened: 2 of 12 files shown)


jeromedalbert:
It seems that there isn't much appetite for this change, which is totally fine. I'm going to close this PR just to clean things up. People can always comment or reference this later if needed!

PR #55174, opened 2025-06-09
Active Job Continuations 3

djmb:
Follow up to https://github.com/rails/rails/pull/55127 and https://github.com/rails/rails/pull/55151

- Instrument steps in a block so that runtime is recorded (matching `perform`/`perform_started`)
- Allow job resumption configuration - `max_resumptions`, `resume_options`, `resume_errors_after_advancing`
- Add Active Record Railtie to prevent checkpoints in database transactions
- Checkpoint before each step except the first, rather than after each step.
- Error if order of completed steps changes when re-running

### DIFF
--- a/activejob/lib/active_job/continuable.rb
+++ b/activejob/lib/active_job/continuable.rb
@@ -11,15 +11,26 @@ module ActiveJob
   module Continuable
     extend ActiveSupport::Concern
 
-    CONTINUATION_KEY = "continuation"
-
     included do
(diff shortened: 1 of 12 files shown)


fatkodima:
Please don't forget to squash commits when merging to make the git history clean.

fatkodima:
Also wondering why it was initially quickly merged without reviews (which will lead to less follow ups).

PR #55176, opened 2025-06-09
Add ability to change the transaction isolation for all pools within a block

eileencodes:
rails/rails#54836 introduced the ability to change the isolation for the current pool. If you're changing the default transaction isolation for an app with many dbs/shards you would need to loop through all pools and apply the isolation. If there are hundreds of connections, that might create a performance issue.

This change adds a new method
`ActiveRecord.with_transaction_isolation_level` which will change the transaction isolation for any pool accessed within the block. The method can be used in an around filter to update all transactions for an action.

I chose to call this from `with_transaction_returning_status` and `transaction` because called from within `connection.transaction` made it too difficult to tell if we were changing the isolation on an open transaction.

While implementing this I felt that the original method had a confusing name paired with the new one so I updated it to be more explicit that it's for only the currently accessed pool.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,30 @@
+*   Add ability to change transaction isolation for all pools within a block.
+
+    This functionality is useful if your application needs to change the database
+    transaction isolation for a request or action.
+
+    Calling `ActiveRecord.with_transaction_isolation_level(level) {}` in an around filter or
+    middleware will set the transaction isolation for all pools accessed within the block,
+    but not for the pools that aren't.
+
+    This works with explicit and implicit transactions:
+
+    ```ruby
+    ActiveRecord.with_transaction_isolation_level(:read_committed) do
+      Tag.transaction do # opens a transaction explicitly
+        Tag.create!
+      end
+    end
+    ```
+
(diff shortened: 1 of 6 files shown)


activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,28 @@
+*   Add ability to change transaction isolation for all pools within a block.
+
+    If your application needs to change transaction isolation for the duration of a
+    block, use `ActiveRecord::Base.with_transaction_isolation_level {}`. This will change
```
byroot: If this is fully global, perhaps it would be better to define it on `ActiveRecord` rather than `ActiveRecord::Base`.

This way it doesn't look like it's a per model or per db_config action.

eileencodes: Good point, I'll move it tomorrow.

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+*   Add ability to change transaction isolation for all pools within a block.
+
+    This functionaility is useful if your application needs to change the database
```
p8: Nitpick..
```suggestion
    This functionality is useful if your application needs to change the database
```

PR #55188, opened 2025-06-12
[RF-DOCS][ci skip] Update Active Record Encryption Guide

bhumi1102:
### Motivation

The purpose of this PR is to update the Active Record Encryption Guide. The Active Record Encryption guide is currently labeled work-in-progress (WIP). 

### Detail

* [X]  Update introduction and basic setup
- [X] Maybe rename "Supported Types" to "Serialized Attributes"?
   "Supported types" could apply to a lot of things (encryption types, column types?) which might be confusing.
- [X] Features is a bit of a grab bag of items.
   - Maybe move some of those sections to the "Basic Usage" section? For example "Supported Types", "Ignoring Case" and "Compression" describe options to the `encrypts` method and seem pretty common. Maybe this should also apply to "Action Text" and "Fixtures".
- [X] Maybe create a new section, something like "Migrating Existing Data" that includes "Support for Unencrypted Data" and "Support for Previous Encryption Schemes"?
- [X] I'm not sure "API" deserves its own section. "API" also isn't very descriptive. Some sections could probably be integrated under `Basic Usage` or maybe removed?
- [X] The "Configuration Options" section can probably be removed because these options should already be described in https://edgeguides.rubyonrails.org/configuring.html
Left the configuration options in for now. The Configuration guide has a couple but not all 14 of them. Maybe it makes sense to move them all there...
- [X] The "Encryption Contexts" section should probably be a top section instead.
- [X]  Maybe describe [with_encryption_context](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Encryption/Contexts.html#method-i-with_encryption_context) first in/before "Built-in Encryption Contexts"? This would make describing `without_encryption` a bit clearer as it [passes the NullEncryptor context](https://github.com/rails/rails/blob/747f85f200e7bb2c1a31b4e26e5a5655e2dc0cdc/activerecord/lib/active_record/encryption/contexts.rb#L49-L51) to `with_encryption_context`.
- [X]  The subsections of "Encryption Context" should probably be mimic "Key Management": "Built-in Encryption Contexts", "Custom Encryption Contexts", "Attribute Encryption Contexts".


guides/source/configuring.md
```diff
@@ -1747,40 +1747,63 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
 default, foreign key names starting with `fk_rails_` are not exported to the
 database schema dump. Defaults to `/^fk_rails_[0-9a-f]{10}$/`.
 
+#### `config.active_record.encryption.support_unencrypted_data`
+
+When true, unencrypted data can be read normally. When false, it will raise errors. Default: `false`.
```
p8: I think it's common to use backticks for all boolean values in this guide:
```suggestion
When `true`, unencrypted data can be read normally. When `false`, it will raise errors. Default: `false`.
```

guides/source/configuring.md
```diff
@@ -1747,40 +1747,63 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
...
+
+#### `config.active_record.encryption.extend_queries`
+
+When true, queries referencing deterministically encrypted attributes will be modified to include additional values if needed. Those additional values will be the clean version of the value (when `config.active_record.encryption.support_unencrypted_data` is true) and values encrypted with previous encryption schemes, if any (as provided with the `previous:` option). Default: `false`.
```
p8:
```suggestion
When `true`, queries referencing deterministically encrypted attributes will be modified to include additional values if needed. Those additional values will be the clean version of the value (when `config.active_record.encryption.support_unencrypted_data` is `true`) and values encrypted with previous encryption schemes, if any (as provided with the `previous:` option). Default: `false`.
```

guides/source/configuring.md
```diff
@@ -1747,40 +1747,63 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
...
+
+#### `config.active_record.encryption.encrypt_fixtures`
+
+When true, encryptable attributes in fixtures will be automatically encrypted when loaded. Default: `false`.
```
p8:
```suggestion
When `true`, encryptable attributes in fixtures will be automatically encrypted when loaded. Default: `false`.
```

guides/source/configuring.md
```diff
@@ -1747,40 +1747,63 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
...
+
+#### `config.active_record.encryption.store_key_references`
+
+When true, a reference to the encryption key is stored in the headers of the encrypted message. This makes for faster decryption when multiple keys are in use. Default: `false`.
```
p8:
```suggestion
When `true`, a reference to the encryption key is stored in the headers of the encrypted message. This makes for faster decryption when multiple keys are in use. Default: `false`.
```

guides/source/configuring.md
```diff
@@ -1747,40 +1747,63 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
...
+
 #### `config.active_record.encryption.add_to_filter_parameters`
 
-Enables automatic filtering of encrypted attributes on `inspect`.
+When true, encrypted attribute names are added automatically to [`config.filter_parameters`](#config-filter-parameters) and won't be shown in logs. Default: `true`.
```
p8:
```suggestion
When `true`, encrypted attribute names are added automatically to [`config.filter_parameters`](#config-filter-parameters) and won't be shown in logs. Default: `true`.
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
 * How to migrate unencrypted data.
 * How to make different encryption schemes coexist.
-* How to use the API.
-* How to configure the library and how to extend it.
+* More about advanced concepts like Encryption Contexts and Key Providers.
 
 --------------------------------------------------------------------------------
 
-Active Record supports application-level encryption. It works by declaring which attributes should be encrypted and seamlessly encrypting and decrypting them when necessary. The encryption layer sits between the database and the application. The application will access unencrypted data, but the database will store it encrypted.
+Active Record Encryption exists to protect sensitive information in your application, such as personally identifiable information (PII) about your users. Active Record supports application-level encryption by allowing you to declare which attributes should be encrypted. It enables transparent encryption and decryption of attributes when saving and retrieving data. The encryption layer sits between the application and the database.
```
p8: If the next section states that encryption sits at the application level, the last line of this paragraph should probably do the same?
```suggestion
Active Record Encryption exists to protect sensitive information in your application, such as personally identifiable information (PII) about your users. Active Record supports application-level encryption by allowing you to declare which attributes should be encrypted. It enables transparent encryption and decryption of attributes when saving and retrieving data. The encryption layer sits at the application level.
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
   encrypts :email, deterministic: true
 end
 
-Author.find_by_email("some@email.com") # You can query the model normally
+# You can only query by email if using deterministic encryption.
+Author.find_by_email("tolkien@email.com")
 ```
 
-The non-deterministic approach is recommended unless you need to query the data.
+The `deterministic:` option generates initialization vectors in a deterministic
```
p8: Nitpick. I think it's more common in the guides to put the colon at the beginning for options.
It also makes it more consisitent with line 123:
```suggestion
The `:deterministic` option generates initialization vectors in a deterministic
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
 
-### Action Text
+## Basic Usage
 
-You can encrypt Action Text attributes by passing `encrypted: true` in their declaration.
+### Ignoring Case
+
+You might need to ignore the case when querying deterministically encrypted data. Two approaches make accomplishing this easier:
+
+You can use the `:downcase` option when declaring the encrypted attribute to downcase the content before encryption occurs.
```
p8: The colon at the end of the second sentence feels somewhat out of place, if there isn't a list or the approaches aren't clearly mentioned as options.
Maybe something like:
```suggestion
You might need to ignore the case when querying deterministically encrypted data. Two approaches make accomplishing this easier.

Firstly, you can use the `:downcase` option when declaring the encrypted attribute to downcase the content before encryption occurs.
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
 ```
 
-NOTE: Passing individual encryption options to Action Text attributes is not supported yet. It will use non-deterministic encryption with the global encryption options configured.
-
-### Fixtures
-
-You can get Rails fixtures encrypted automatically by adding this option to your `test.rb`:
+When using `:downcase`, the original case is lost. In some situations, you might
+want to ignore the case only when querying while also storing the original case.
+For those situations, you can use the `:ignore_case` option. This requires you
```
p8: Maybe make it clear that this is the second approach?
Maybe something like:
```suggestion
For those situations, you can use the `:ignore_case` option, as the second approach. This requires you
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
+to add a new column named `original_<column_name>` to store the content with the
+case unchanged:
```
p8: This is a bit unclear to me, but I expect this also encrypts the original content?
```suggestion
to add a new column named `original_<column_name>` to store the encrypted content with the
case unchanged:
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
 
 ```ruby
-config.active_record.encryption.encrypt_fixtures = true
+class Label
+  encrypts :name, deterministic: true, ignore_case: true # the content with the original case will be stored in the column `original_name`
```
p8:
```suggestion
  encrypts :name, deterministic: true, ignore_case: true # the encrypted content with the original case will be stored in the column `original_name`
```

guides/source/active_record_encryption.md
```diff
@@ -3,151 +3,174 @@
...
 
-### Supported Types
+```ruby
+class Article < ApplicationRecord
+  encrypts :metadata, message_serializer: SomeCustomMessageSerializer.new
+end
+```
 
-`active_record.encryption` will serialize values using the underlying type before encrypting them, but, unless using a custom `message_serializer`, *they must be serializable as strings*. Structured types like `serialized` are supported out of the box.
+Attributes with structured types using the [`serialized`](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) method are supported with encryption as well. The `serialized` method is used when you have an attribute that needs to be saved to the database as a serialized object (using `YAML`, `JSON` or such), and retrieved by deserializing into the same object.
```
p8: Nitpick.
```suggestion
Attributes with structured types using the [`serialized`](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) method support encryption as well. The `serialized` method is used when you have an attribute that needs to be saved to the database as a serialized object (using `YAML`, `JSON` or such), and retrieved by deserializing into the same object.
```

...or maybe:
```suggestion
Attributes with structured types using the [`serialized`](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) method can be encrypted as well. The `serialized` method is used when you have an attribute that needs to be saved to the database as a serialized object (using `YAML`, `JSON` or such), and retrieved by deserializing into the same object.
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +186,122 @@ class Article < ApplicationRecord
...
-* When reading encrypted data, Active Record Encryption will try previous encryption schemes if the current scheme doesn't work.
-* When querying deterministic data, it will add ciphertexts using previous schemes so that queries work seamlessly with data encrypted with different schemes. You must set `config.active_record.encryption.extend_queries = true` to enable this.
-
-You can configure previous encryption schemes:
-
-* Globally
-* On a per-attribute basis
-
-#### Global Previous Encryption Schemes
+NOTE: If you want to ignore the case, make sure to use the `downcase` or `ignore_case` option in the `encrypts` declaration. Using the `case_sensitive` option in the validation won't work.
```
p8: It was a bit unclear to me this note is for uniqueness.
I think it's also clearer to use colons as a prefix for options here:
Maybe something like:
```suggestion
NOTE: If you want to ignore the case for uniqueness, make sure to use the `:downcase` or `:ignore_case` option in the `encrypts` declaration. Using the `:case_sensitive` option in the validation won't work.
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +186,122 @@ class Article < ApplicationRecord
 end
 ```
 
-### Ignoring Case
+### Unique Constraints with Encrypted Data
```
p8: I was expecting this section to describe database constraints.
Maybe avoid that term?
```suggestion
### Ensuring Uniqueness with Encrypted Data
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +186,122 @@ class Article < ApplicationRecord
...
-You can use the `:downcase` option when declaring the encrypted attribute to downcase the content before encryption occurs.
+#### Unique Validations
+
+In order to support unique validations, you'll need to enable extended queries. The default value for this configuration is false:
 
 ```ruby
-class Person
-  encrypts :email_address, deterministic: true, downcase: true
-end
+# config/applicaiton.rb
```
p8:
```suggestion
# config/application.rb
```

guides/source/active_record_encryption.md
bhumi1102: Yeah I guess so. OR I'm thinking about just deleting that last sentence, it doesn't add anything new (we already mention application-level earlier in this paragraph)

bhumi1102: Removed.

guides/source/active_record_encryption.md
bhumi1102: Agree. Updated that whole section to clarify what the two options are. Updated that specific intro sentence to:
"There are two options for achieving this - `:downcase` and `:ignore_case`."

guides/source/active_record_encryption.md
bhumi1102: Yes...I was thinking that too, otherwise it kinda defeats the purpose :)

guides/source/active_record_encryption.md
bhumi1102: updated manually

guides/source/active_record_encryption.md
p8: Nice! This section reads a lot better now.

guides/source/active_record_encryption.md
```diff
@@ -328,281 +408,368 @@ class User
...
-This can be handy if you want to protect encrypted data while still running arbitrary code against it (e.g. in a Rails console).
+## Configuration
+
+You can configure Active Record Encryption options in your `application.rb`
+(most common scenario) or in a specific environment config file
+`config/environments/<env name>.rb` if you want to set them on a per-environment
+basis. You can see the Active Record Encryption related configuration properties
+in the [Configuration
+Guide](configuring.html#config-active-record-encryption-support-unencrypted-data)
+
```
p8: I think we can remove this section now.
It's pretty generic and other guides don't have a configuration section either .
```suggestion
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +238,157 @@ class Article < ApplicationRecord
...
+### Ensuring Uniqueness with Encrypted Data
+
+Checking for uniqueness is only supported with deterministically encrypted data.
 
-You might need to ignore casing when querying deterministically encrypted data. Two approaches make accomplishing this easier:
+#### Unique Validations
 
-You can use the `:downcase` option when declaring the encrypted attribute to downcase the content before encryption occurs.
+In order to support unique validations, you'll need to enable extended queries.
+The default value for this configuration is false:
```
p8: I think it would be clearer if we can move the reasoning of using `extend_queries` on line 265 here, so there is less jumping of context?
Maybe something like:
```suggestion
In order to support unique validations, you'll need to enable extended queries. This makes sure we query for the downcased encrypted attribute (if used) and previous encryption schemes as well. The default value for this configuration is false:
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +238,157 @@ class Article < ApplicationRecord
...
+  validates :email_address, uniqueness: true
+  encrypts :email_address, deterministic: true, downcase: true
 end
 ```
 
-### Support for Unencrypted Data
-
-To ease migrations of unencrypted data, the library includes the option `config.active_record.encryption.support_unencrypted_data`. When set to `true`:
+The `extended_queries` configuration also allows combining encrypted and
+unencrypted data, and when configuring previous encryption schemes.
```
p8: I we move this into the opening paragraph, we can remove it here
```suggestion
```

p8 (approved):

fxn:
Hey, this guide is great!

I have done a pass. I don't know exactly which points belong to the patch, and which were already present, but I'll report it all anyway:

* _"Encrypting specific attributes at the application-level adds an additional security layer."_ —  "application-level" is not an adjective, I believe it should go without hyphen: "attributes at the application level".

* In addition to compression, Active Record Encryption encodes the values in the JSON document using Base64. Maybe worth saying somewhere so that readers understand things like `n7J0/ol+a7DRMeaE` in the samples.

* I wonder if the warning about encryption taking extra space makes sense in "Why Encrypt Data at the Application Level?". Could perhaps make more sense in another section? Perhaps in "Declare Encrypted Attributes"?

* After you execute `db:encryption:init`, you need to `bin/rails credentials:edit` to be able to paste the keys (there is no `config/credentials.yml`).

* You do not need `Rails.application.credentials` in `config/application.rb`, `credentials` suffices:

  ```ruby
  config.active_record.encryption.primary_key = credentials.dig(...)
  ```

  Just like you say `config` because it is shorter than `Rails.application.config`.

* I would say the 2nd level heading "Basic Usage" should go above "Declare Encrypted Attributes".

* Style: `article = Article.create title: "Encrypt it all!"` we use parentheses in method calls.

* When I created the article using SQLite, the printed SQL was

  ```sql
  INSERT INTO "articles" ("title", "created_at", "updated_at")
  VALUES ('{"p":"oq+RFYW8CucALxnJ6ccx","h":{"iv":"3nrJAIYcN1+YcGMQ","at":"JBsw7uB90yAyWbQ8E3krjg=="}}', '2025-09-11 10:58:28.364295', '2025-09-11 10:58:28.364295') RETURNING "id"
  ```

  which could be abbreviated to

  ```sql
  INSERT INTO "articles" ("title", "created_at", "updated_at")
  VALUES ('{"p":"oq+RFYW8CucALxnJ6ccx","h":{"iv":"3nrJAIYcN1+YcGMQ","at":"JBsw7uB90yAyWbQ8E3krjg=="}}', ...) RETURNING "id"
  ```

  or something.

  What do you think? Maybe looks bit cleaner, but it is a minor point, your call :).

* Look at the title database value, the reader may be puzzled. Shouldn't it be the ciphertext? What is all that?

  May I suggest giving an overview? Something like, "AR Encryption stores a JSON object that contains the ciphertext (compressed and encoded in Base64) in the "p" key ("p" stands for "payload"). Then, "h" stands for "headers", and it stores metadata needed to decrypt the value." I don't know, something like that.

* Maybe it could be interesting to show that the attribute appears as `[FILTERED]` in the console:

  ```
  > Article.find_by(title: 'foo')
  #<Article:0x000000012b034608
   id: 1,
   title: "[FILTERED]",
   created_at: "2025-09-11 11:03:01.089735000 +0000",
   updated_at: "2025-09-11 11:03:01.089735000 +0000"
  ```

* _"The `:deterministic` option generates initialization vectors in a deterministic way"_ — The reader may not know what is an initialization vector. If you explain the format, maybe you could observe that "p" and "iv" are the same for the same value? Even the same value set in different attributes or different models!

* What about elaborating a bit more about searches and determinism? For example, saying that when you set `deterministic: true`, Active Record is able to compute the corresponding and unique database value for that clear text (which is not just the ciphertext, as we just saw) and with that do a regular string equality comparison. Without determinism, such query condition is not possible because on every encryption you get a different JSON document.

* Typo in _"It is generated as an a function of the key and the plaintext content"_.

* _"In order to support unique validations, you'll need to enable extended queries. This makes sure we query for the downcased encrypted attribute (if used) and previous encryption schemes as well."_  — Why should it query for downcased encrypted attributes and not the exact value?

* I do not understand extended queries vs uniqueness. We saw above that you can specify `downcase: true` and it just works. Why do we need "extended queries"? Which queries are "extended"? What does that even mean? (Playing evil's advocate here :).

* _"One thing Rails does to help is that, by default, deterministic attributes will use the oldest available encryption scheme when multiple encryption schemes are configured."_  — At this point in the guide, the reader does not know what an encryption scheme is, or that it can configure more than one.

* _"Encrypted columns are configured"_ — The guide is using the term "Encrypted attributes". Indeed, that is what you configure, the model attribute (a model has _attributes_, a table has _columns_). Also, the column is not encrypted, if you think about it, it is a JSON document. What the feature encrypts, compresses, and encodes, is the attribute original value, which ends up in the "p" key. (There are several similar instances of "column".)

* Fixtures section: The reader may not realize fixtures loading is direct and wonder what is this all about. Perhaps it would be helpful to be more explicit. Something like: Fixtures can have these values as clear text in the YAML file (something useful to say, regardless), but Active Record would load them verbatim in the database. To have them encrypted, you need to set `config.active_record.encryption.encrypt_fixtures = true`.

* Typo: _"You can disable compression by setting `compress` option when encrypting attributes:"_

* Typo: _"Active Record Encryption is meant to be used declaratively, but there is also an API for debugging or advance use cases."_

fxn:
Once the feedback is addressed ... @jorgemanrubia (hey! :) there is nobody more qualified to review this guide than you, would you like to have a look at it?

guides/source/active_record_encryption.md
```diff
@@ -1,153 +1,227 @@
...
 ```
 
-But, under the hood, the executed SQL looks like this:
+However, in the Rails console, the executed SQL looks like this:
 
 ```sql
 INSERT INTO `articles` (`title`) VALUES ('{\"p\":\"n7J0/ol+a7DRMeaE\",\"h\":{\"iv\":\"DXZMDWUKfp3bg/Yu\",\"at\":\"X1/YjMHbHD4talgF9dt61A==\"}}')
 ```
 
-#### Important: About Storage and Column Size
```
jorgemanrubia, line 77: This section on storage concerns is very important. It's not intuitive, and the user needs to be well aware of it. Why removing it? I initially thought it had been moved but I can't find it.

guides/source/active_record_encryption.md
```diff
@@ -1,153 +1,227 @@
...
+These values can be stored by copying and pasting the generated values into your
+existing [Rails credentials](/security.html#custom-credentials). Then, you can
+set the credentials in a config file:
+
+```ruby
+# config/application.rb
+config.active_record.encryption.primary_key = Rails.application.credentials.dig(:active_record_encryption, :primary_key)
+config.active_record.encryption.deterministic_key = Rails.application.credentials.dig(:active_record_encryption, :deterministic_key)
+config.active_record.encryption.key_derivation_salt = Rails.application.credentials.dig(:active_record_encryption, :key_derivation_salt)
+```
```
jorgemanrubia: The system already picks the credentials described before, when they are set. Adding this additional example where you assign the same credentials manually can create confusion, and doesn't add much value IMO. The next example with env vars already clarifies that you can assign these values too.

jorgemanrubia (commented):
This looks fantastic @bhumi1102, thanks for working on this. I left a couple of comments on top of the great ones by @fxn. Regarding this:

> "In order to support unique validations, you'll need to enable extended queries. This makes sure we query for the downcased encrypted attribute (if used) and previous encryption schemes as well." — Why should it query for downcased encrypted attributes and not the exact value?
> 
> I do not understand extended queries vs uniqueness. We saw above that you can specify downcase: true and it just works. Why do we need "extended queries"? Which queries are "extended"? What does that even mean? (Playing evil's advocate here :).

I know you carried this over from the previous guide but I think that was misleading. The `extended_queries` is meant to let you query deterministically encrypted data when (1) you have encrypted and unencrypted data and (2) when you configure other encryption schemes to support old data. So, in general, uniqueness will work. In the cases where you need (1) or (2), then, for unique validations to work, you need to set this flag.

Notice that (2) will be extremely rare in practice. It was a need we had as we developed this and found ourselves with encrypted data and having to do big changes to the encryption system.

Thanks 🙏

bhumi1102:
Great! Thank you both for reviewing! Once I process the feedback, I'll update here.

fxn:
@jorgemanrubia to understand this limitation better: On paper, would it be possible to support key rotation for deterministic encryption generating `OR` clauses in searches? Or is there a fundamental reason that makes that impossible?

p8:
I'm not Jorge, but I think this is what the extended queries are used for:
With the following table:
| id | encrypted_name |
|- | - |
| 1 | xavier_deterministically_encrypted_with_old_key |
| 2 | xavier_deterministically_encrypted_with_new_key |
| 3 | jorge_deterministically_encrypted_with_new_key |

When searching for the name "xavier", you basically search with the following SQL:
```sql
where encrypted_name = 'xavier_deterministically_encrypted_with_old_key'
   or encrypted_name = 'xavier_deterministically_encrypted_with_new_key'
```

fxn:
Hmmm, if you try to set two deterministic keys, you get

```
ActiveRecord::Encryption::DeterministicKeyProvider#initialize': Deterministic encryption keys can't be rotated (ActiveRecord::Encryption::Errors::Configuration
```

as the guide says, it is not supported.

I'd like to understand if this is because support has not been implemented, or because it cannot be implemented.

fxn:
Another point about compression.

Compression logic can be located in [ActiveRecord::Encryption::Encryptor](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/encryption/encryptor.rb):

```ruby
def build_encrypted_message(clear_text, key_provider:, cipher_options:)
  key = key_provider.encryption_key

  clear_text, was_compressed = compress_if_worth_it(clear_text)
  cipher.encrypt(clear_text, key: key.secret, **cipher_options).tap do |message|
    message.headers.add(key.public_tags)
    message.headers.compressed = true if was_compressed
  end
end
```

Two observations:

1. Compression happens _before_ ecryption. Therefore, when we describe what the feature does or what we are looking at in the database, compression should be mentioned before encryption: "Compress (maybe), encrypt, encode values in Base64, serialize with JSON" (I believe I got the sequence right, assuming defaults).

Now, as the name suggests `compress_if_worth_it` compresses depending on some heuristic:

```ruby
def compress_if_worth_it(string)
  if compress? && string.bytesize > THRESHOLD_TO_JUSTIFY_COMPRESSION
    [compress(string), true]
  else
    [string, false]
  end
end
```

2. Therefore, instead of saying that "Active Record Encryption compresses encrypted payloads by default.", perhaps we could say that compression is _enabled_ by default. Then explain that it happens on individual clear texts depending on whether it is... worth it :).

If compression was applied, you can see it in the headers. See the `"c": true` in:

```json
{
  "p": "PS387lhXYzpsyjGQ4mw1",
  "h": {
    "iv": "wuftnVRYkPzXJpEE",
    "at": "V+596sl6LqTNFvJjLwe5Ug==",
    "i": "ZWVjNw==",
    "c": true
  }
}
```

jorgemanrubia:
@fxn the system for rotating keys is pretty simple and is based letting you add multiple keys, using all of them for trying to decrypt and the last one for encrypting. With deterministic encryption, you want that two identical clear texts result in the same ciphertexts when encrypted. Rotating keys would make this impossible. Maybe you are confused because the library support multiple encryption schemes and, with the extended queries system, you can query using old encryption schemes too. But this system wasn't designed with key rotation schemes in mind, although it could probably be used for that. In practice, I think wanting to rotate keys when you want data encrypted deterministically is unnatural.

jorgemanrubia:
> On paper, would it be possible to support key rotation for deterministic encryption generating OR clauses in searches? Or is there a fundamental reason that makes that impossible?

I think it would be possible in paper with the ¨previous encryption schemes" system, but it wouldn't be something I would advise doing. That system is intended to changes in your encryption approach. E.g: because you want to support data deterministically encrypted with other system, and you want to keep that working while you introduce a Active Record encryption.

fxn:
@jorgemanrubia hmmm, let me explain what I had in mind.

> With deterministic encryption, you want that two identical clear texts result in the same ciphertexts when encrypted.

I would say that with deterministic encryption, you want that two identical clear texts **encrypted with the same key** result in the same ciphertexts.

So, I was thinking that if `bank_account` is encrypted using a deterministic key for searches, you could still rotate. With rotation, now the column would have some values coming from one key, and others from the other one. If one bank account was encrypted with the old key, the message for the same bank account would be different today.

Encryption would need to consider the two possible JSONs when searching, looking for uniqueness, etc.

Does that make sense?

fxn:
Just to be clear, I don't mean to ask if that is possible with the current feature set, eh?

It is a mental exercise: I _believe_ that if the feature _was_ able to generate an `OR` when doing `Account.where(bank_account: bank_account)` (and similar use cases) that checks the JSON corresponding to the _n_ possible rotated deterministic keys, it would be possible. (But maybe I miss something.)

Today, this is very far away, because encryption/decryption and searches is based on AMo attributes.

fxn:
I am wondering this because it is a pity that the need for searches prevents key rotation, which is a must.

I am trying to understand the limitation, because it seems to me it is not a fundamental one, only a feature that has not been implemented. Do you agree?

jorgemanrubia:
Yes, the current implementation proves that this is technically possible. The previous encryption schemes system does this: use all the previous schemes for trying to decrypt and resolve queries, and only use one for encryption.

guides/source/active_record_encryption.md
bhumi1102: Restored

guides/source/active_record_encryption.md
bhumi1102: Got it, removed.

bhumi1102:
I have addressed each of the bullet points within @fxn's big comment above. Additionally, I've also updated compression section (enabled by default but applied if "worth it"). And finally, I've corrected the docs around extended queries (that it is not necessary for uniqueness).

CC: @p8

guides/source/active_record_encryption.md
```diff
@@ -1,92 +1,149 @@
...
+INSERT INTO "articles" ("title", "created_at", "updated_at")
+VALUES ('{"p":"oq+RFYW8CucALxnJ6ccx","h":{"iv":"3nrJAIYcN1+YcGMQ","at":"JBsw7uB90yAyWbQ8E3krjg=="}}', ...) RETURNING "id"
+```
+
+The value inserted is a JSON object that contains the encrypted value for the `title` attribute. More specifically, the JSON object stores two keys: `p` for payload and `h` for headers. The ciphertext, which is compressed and encoded in Base64, is stored as the payload. The `h` key stores metadata needed to decrypt the value. The `iv` value is the initialization vector and `at` is authentication tag (used to ensure the ciphertext has not been tampered with).
+
+When looking at the `Article` in the Rails console, the encrypted attribute `title` will also be filtered:
+
+```irb
+irb> Article.first
```
p8: As this is the Rails console we should probably show the console prompt instead.
```suggestion
my-app(dev)> Article.first
```

guides/source/active_record_encryption.md
```diff
@@ -95,59 +152,129 @@ Some examples:
...
-The non-deterministic approach is recommended unless you need to query the data.
+The `:deterministic` option generates initialization vectors in a deterministic
+way, meaning it will produce the same encrypted output given the same plaintext
+input value. This makes querying encrypted attributes possible by string
+equality comparison. For example, notice that the `p` and `iv` key in the JSON
+document have the same value when we create and when we query an Author's email:
 
-NOTE: In non-deterministic mode, Active Record uses AES-GCM with a 256-bits key and a random initialization vector. In deterministic mode, it also uses AES-GCM, but the initialization vector is generated as an HMAC-SHA-256 digest of the key and contents to encrypt.
+```irb
+irb> author = Author.create(name: "J.R.R. Tolkien", email: "tolkien@email.com")
```
p8:
```suggestion
my-app(dev)> author = Author.create(name: "J.R.R. Tolkien", email: "tolkien@email.com")
```

guides/source/active_record_encryption.md
```diff
@@ -95,59 +152,129 @@ Some examples:
...
+  TRANSACTION (0.1ms)  begin transaction
+  Author Create (0.4ms)  INSERT INTO "authors" ("name", "email", "created_at", "updated_at") VALUES (?, ?, ?, ?) RETURNING "id"  [["name", "J.R.R. Tolkien"], ["email", "{\"p\":\"8BAc8dGXqxksThLNmKmbWG8=\",\"h\":{\"iv\":\"NgqthINGlvoN+fhP\",\"at\":\"1uVTEDmQmPfpi1ULT9Nznw==\"}}"], ["created_at", "2025-09-19 18:08:40.104634"], ["updated_at", "2025-09-19 18:08:40.104634"]]
+  TRANSACTION (0.1ms)  commit transaction
 
-NOTE: You can disable deterministic encryption by omitting a `deterministic_key`.
+irb> Author.find_by_email("tolkien@email.com")
```
p8:
```suggestion
my-app(dev)> Author.find_by_email("tolkien@email.com")
```

guides/source/active_record_encryption.md
```diff
@@ -95,59 +152,129 @@ Some examples:
...
+    created_at: Fri, 19 Sep 2025 18:08:40.104634000 UTC +00:00,
+    updated_at: Fri, 19 Sep 2025 18:08:40.104634000 UTC +00:00>
+```
 
-## Features
+In the above example, the initialization vector, `iv`, has the value `"NgqthINGlvoN+fhP"` for the same string. Even if you use the same email string in a different model instance (or different attribute with deterministic encryption), it will map to the same `p` and `iv` values:
 
-### Action Text
+```irb
+irb> author2 = Author.create(name: "Different Author", email: "tolkien@email.com")
```
p8:
```suggestion
my-app(dev)> author2 = Author.create(name: "Different Author", email: "tolkien@email.com")
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +290,146 @@ class Article < ApplicationRecord
...
-Unique validations are supported normally as long as extended queries are enabled (`config.active_record.encryption.extend_queries = true`).
+In case you need to disable filtering of encrypted parameters, you can use the
+following configuration:
 
 ```ruby
-class Person
-  validates :email_address, uniqueness: true
-  encrypts :email_address, deterministic: true, downcase: true
-end
+# config/applicaiton.rb
```
p8:
```suggestion
# config/application.rb
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +290,146 @@ class Article < ApplicationRecord
...
 
-### Filtering Params Named as Encrypted Columns
+NOTE: Passing individual encryption options to Action Text attributes is not
+supported. It will use non-deterministic encryption with the global encryption
+options configured.
 
-By default, encrypted columns are configured to be [automatically filtered in Rails logs](configuring.html#config-filter-parameters). You can disable this behavior by adding the following to your `application.rb`:
+### Fixtures
+
+To allow your tests can use plain text values in the YAML fixture files for encrypted attributes, you can configure fixtures to be automatically encrypted by adding this configuration to your `test.rb` file:
```
p8: Showing the fullpath to `test.rb` might be useful for newcomers.
```suggestion
To allow your tests can use plain text values in the YAML fixture files for encrypted attributes, you can configure fixtures to be automatically encrypted by adding this configuration to your `config/environments/test.rb` file:
```

guides/source/active_record_encryption.md
```diff
@@ -163,152 +290,146 @@ class Article < ApplicationRecord
...
 
 ```ruby
-config.active_record.encryption.add_to_filter_parameters = false
+config.active_record.encryption.encrypt_fixtures = true
```
p8: Nitpick.. feel free to ignore.
Maybe make it clear this needs to be set in the `configure` block?
```suggestion
Rails.application.configure do
  ...
  config.active_record.encryption.encrypt_fixtures = true
  ...
end
```

guides/source/active_record_encryption.md
```diff
@@ -328,281 +449,358 @@ class User
...
 ```
 
-### Attribute-specific Keys
+Next, let's see how to configure previous encryption schemes.
+
+#### Global Previous Encryption Schemes
 
-You can configure a given key on a per-attribute basis with the `:key` option:
+You can add previous encryption schemes by adding them as a list of properties
+using the `previous` config property in your `application.rb`:
```
p8: Showing the fullpath to application.rb might be useful for newcomers.
```suggestion
using the `previous` config property in your `config/application.rb`:
```

guides/source/active_record_encryption.md
```diff
@@ -328,281 +449,358 @@ class User
...
+instead).
 
-```ruby
-article.ciphertext_for(:title)
-```
+### Built-In Encryption Context
 
-#### Check if the Attribute is Encrypted or Not
+The global encryption context is the one used by default and is configured with
+other configuration properties in your `application.rb` or environment config
```
p8:
```suggestion
other configuration properties in your `config/application.rb` or environment config
```

guides/source/active_record_encryption.md
```diff
@@ -328,281 +449,358 @@ class User
...
 
-The default encoding for attributes encrypted deterministically. You can disable forced encoding by setting this option to `nil`. It's `Encoding::UTF_8` by default.
+The `EnvelopeEncryptionKeyProvider` generates a random key for each data
+encryption operation. It stores the data-key with the data itself. Then, the
+data-key is also encrypted with a primary key defined in the credential
+`active_record.encryption.primary_key`.
 
-#### `config.active_record.encryption.hash_digest_class`
+You can configure Active Record to use this key provider by adding this to your
+`application.rb`:
```
p8:
```suggestion
`config/application.rb`:
```

guides/source/active_record_encryption.md
```diff
@@ -328,281 +449,358 @@ class User
...
-You can override encryption context params by passing them in the attribute declaration:
+You can configure a specific key for a given attribute using the `key` option:
 
 ```ruby
-class Attribute
-  encrypts :title, encryptor: MyAttributeEncryptor.new
+class Article < ApplicationRecord
+  encrypts :summary, key: "some secret key for article summaries"
 end
 ```
```
p8, line 809: We probably need to warn against hard coding secret keys here:
```suggestion

WARNING: Never hard code your secret keys in your application. Use an `ENV` variable or Rails' credentials instead.

```

And/or maybe also give a good example?
```
  encrypts :summary, key: ENV["SOME_SECRET_KEY_FOR_ARTICLE_SUMMARIES"]
```

bhumi1102: Good point. I like the option of going with the "good example" and leaving out the hard-coding + warning.

PR #55191, opened 2025-06-13
Omit charset for application/json Content-Types (IANA-compliant)

### LINKED ISSUE #55063: Rails JSON output includes useless `; charset=utf-8` appended to `content-type`
### Steps to reproduce

Follow the instructions at https://guides.rubyonrails.org/api_app.html to create a new API app.

For completeness, this was:

```shell
% ruby --version
ruby 3.2.7 (2025-02-04 revision 02ec315244) [x86_64-linux]

% gem install rails
...

% rails --version
Rails 8.0.2

% echo Following https://guides.rubyonrails.org/api_app.html

% rails new charsetbug --api

% cd charsetbug

% bin/rails g scaffold Group name:string

% bin/rails db:migrate

% bin/rails console
irb> Group.create(name: "Rails Founders")
irb> Group.create(name: "Rails Contributors")

% bin/rails server
```

Then from another terminal fetch the headers.

```shell
% curl -I http://localhost:3000/groups.json
HTTP/1.1 200 OK
x-frame-options: SAMEORIGIN
x-xss-protection: 0
x-content-type-options: nosniff
x-permitted-cross-domain-policies: none
referrer-policy: strict-origin-when-cross-origin
content-type: application/json; charset=utf-8
etag: W/"864f7877f0a843b38bd31492cc22a2e1"
cache-control: max-age=0, private, must-revalidate
x-request-id: ff8c016c-f54d-4a43-b31a-dd055d55d750
x-runtime: 0.005577
server-timing: start_processing.action_controller;dur=0.01, sql.active_record;dur=0.17, instantiation.active_record;dur=0.06, process_action.action_controller;dur=2.10
Content-Length: 0
```

### Expected behavior

The `content-type` header should be `application/json`.

See e.g., https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type#content-type_in_a_rest_api_using_json which includes a correct example.

### Actual behavior

Actual value: `application/json; charset=utf-8`

The `application/json` type doesn't have a charset, per https://www.iana.org/assignments/media-types/application/json ("No "charset" parameter is defined for this registration.")

Other JSON based types are similar, e.g., 

- https://www.iana.org/assignments/media-types/application/activity+json
- https://www.iana.org/assignments/media-types/application/jrd+json

While this is unlikely to introduce any real-world interoperability problems, it is wasteful.

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.2.7

rebase-master:
### Motivation / Background
This Pull Request has been created because Rails was appending `; charset=utf-8` to pure JSON responses (`Content-Type: application/json; charset=utf-8`), which contradicts the IANA registration for `application/json` (no charset parameter defined). [Nik Clayton](https://github.com/nikclayton) reported this in issue [#55063](https://github.com/rails/rails/issues/55063) and we need to align Rails’ behavior with the spec.

This pull request needs to be merged because:

- Bug fix: Corrects non-spec-compliant behavior in JSON APIs so that consumers see exactly Content-Type: application/json. Fixes #55063 
- Interoperability: Some HTTP frameworks and clients strictly parse Content-Type parameters; removing the charset prevents subtle bugs or middleware rejections.
- Consistency with MDN and IANA: Aligns Rails with widely-cited documentation (MDN) and the official media-type registration.
- User impact: Applications serving JSON via Rails APIs will immediately benefit and no longer need manual workarounds.

### Detail

This Pull Request changes:

- Overrides `ActionDispatch::Response#content_type=` so that when the MIME type is exactly `application/json`, Rails uses `set_content_type 'application/json', nil` instead of defaulting to UTF-8.
- Leaves all other content types unchanged, preserving existing charset logic.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -385,4 +385,8 @@
 
     *Petrik de Heus*
 
+*   Omit `charset=utf-8` for `Content-Type: application/json`, per IANA spec (#55063) – @rebase-master
+
+    *Mansoor Khan*
+
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/actionpack/CHANGELOG.md) for previous changes.
--- a/actionpack/lib/action_dispatch/http/response.rb
+++ b/actionpack/lib/action_dispatch/http/response.rb
@@ -301,8 +301,14 @@ module ActionDispatch # :nodoc:
       prev_header_info = parsed_content_type_header
       charset = new_header_info.charset || prev_header_info.charset
       charset ||= self.class.default_charset unless prev_header_info.mime_type
-      set_content_type new_header_info.mime_type, charset
+
+      if new_header_info.mime_type == "application/json"
+        set_content_type new_header_info.mime_type, nil
+      else
+        set_content_type new_header_info.mime_type, charset
+      end
     end
+
 
     # Content type of response.
     def content_type
@@ -327,7 +333,10 @@ module ActionDispatch # :nodoc:
     #     response.charset = nil      # => 'utf-8'
     def charset=(charset)
       content_type = parsed_content_type_header.mime_type
-      if false == charset
+
+      # omit charset for pure JSON per IANA spec
+      # https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type#content-type_in_a_rest_api_using_json
+      if false == charset || content_type == "application/json"
         set_content_type content_type, nil
       else
         set_content_type content_type, charset || self.class.default_charset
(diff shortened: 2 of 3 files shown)


rafaelfranca (changes requested):
I'm all for fixing the problem but adding conditional behavior based on the mime type is the wrong direction to go. `"application/json"` isn't the only mime type that we should omit the charset. I think the fix should be the opposite. By default we add no charset unless it is required

rebase-master:
> I'm all for fixing the problem but adding conditional behavior based on the mime type is the wrong direction to go. `"application/json"` isn't the only mime type that we should omit the charset. I think the fix should be the opposite. By default we add no charset unless it is required

@rafaelfranca when you say "...unless it is required", what factor should determine that?

rafaelfranca:
Good question. I'd try to check why the charset was added in the first place using the git history

PR #55192, opened 2025-06-14
Allow attribute definition without database connection

### LINKED ISSUE #52311: ActiveRecord `attribute :foo, :bar` requires an active connection
### Steps to reproduce
```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails", "7.1"
  gem "sqlite3", "~> 2.0"
end

require "active_record"
require "minitest/autorun"
require "logger"

# ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  # ActiveRecord::ConnectionNotDefined: No database connection defined.
  def test_attribute_requires_no_connection
    assert_nothing_raised do
      Class.new(ActiveRecord::Base) do
        attribute :foo, :interval
      end
    end
  end
end
```

### Expected behavior
It should succeed even when there is no active connection.

### Actual behavior
It fails when the connection is not established.

### Additional context

https://github.com/rails/rails/pull/41302 added a test to validate that it works without a connection, however I believe it is currently not checking the right thing: https://github.com/rails/rails/blob/4fa56814f18fd3da49c83931fa773caa727d8096/activerecord/test/cases/attributes_test.rb#L359-L365

I'd guess it goes through some other thing that requires a connection now instead.

GoodJob added `attribute :duration, :interval`  to work around a rails 6.1 deprecation warning which resulted in errors for `tapioca` since it evaluates all code to generate the rbi files. More context at https://github.com/bensheldon/good_job/pull/1412

### System configuration
**Rails version**: main

**Ruby version**: 3.3.5

Yuhi-Sato:
### Motivation / Background

PR #41302 "attribute should not require a connection is established" attempted to fix adapter_name_from to get the adapter name from the connection db config to not retrieve a connection. However, the original PR's test was insufficient and didn't properly verify that the connection was actually unavailable during attribute definition. 

This PR properly implements the fix by making adapter_name_from return nil
 when no connection is configured, and includes comprehensive tests that actually verify the behavior when connections are not established.

Fix #52311

### Detail

This Pull Request changes `ActiveRecord::Type.adapter_name_from` to handle missing database connections by returning `nil` instead of raising `ActiveRecord::ConnectionNotDefined`. This allows attributes to be defined on models even when no connection is established.

* Add rescue clause in adapter_name_from to catch ConnectionNotDefined
* Update attributes test to verify connection is not required
* Add comprehensive tests for adapter_name_from behavior

### Additional information

The original PR #41302 used `assert_not_called(ActiveRecord::Base, :lease_connection)` which only verified that lease_connection wasn't called, but didn't test the actual scenario where connection is unavailable. When we explicitly remove the connection in the original test, it fails, demonstrating that the changes in PR #41302 were not actually implementing the intended behavior of allowing attribute definition without a connection.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/type.rb
+++ b/activerecord/lib/active_record/type.rb
@@ -48,6 +48,8 @@ module ActiveRecord
 
       def adapter_name_from(model) # :nodoc:
         model.connection_db_config.adapter.to_sym
+      rescue ActiveRecord::ConnectionNotDefined
+        nil
       end
 
       private
--- a/activerecord/test/cases/attributes_test.rb
+++ b/activerecord/test/cases/attributes_test.rb
@@ -357,9 +357,11 @@ module ActiveRecord
     end
 
     test "attributes do not require a connection is established" do
-      assert_not_called(ActiveRecord::Base, :lease_connection) do
-        Class.new(OverloadedType) do
-          attribute :foo, :string
+      ActiveRecord::Base.stub(:connection_db_config, -> { raise ActiveRecord::ConnectionNotDefined }) do
+        assert_nothing_raised do
+          Class.new(OverloadedType) do
+            attribute :foo, :string
+          end
         end
       end
     end
--- a/activerecord/test/cases/type_test.rb
+++ b/activerecord/test/cases/type_test.rb
@@ -38,4 +38,18 @@ class TypeTest < ActiveRecord::TestCase
 
     assert_equal adapter_type.new, ActiveRecord::Type.lookup(:foo)
   end
+
+  test "adapter_name_from returns adapter symbol when connection exists" do
+    assert_equal ActiveRecord::Base.connection_db_config.adapter.to_sym,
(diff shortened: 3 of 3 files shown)


activerecord/lib/active_record/type.rb
```diff
@@ -48,6 +48,8 @@ def default_value # :nodoc:
 
       def adapter_name_from(model) # :nodoc:
         model.connection_db_config.adapter.to_sym
+      rescue ActiveRecord::ConnectionNotDefined
+        nil
```
skipkayhil, line 52: I don't think this change makes sense. `ConnectionNotDefined` means you don't have a database defined in `config/database.yml`, not that there isn't an active database connection.

Yuhi-Sato: Thank you for the review. As you pointed out in the issue, my approach was incorrect, so I’m closing this PR.

PR #55193, opened 2025-06-14
Add documentation for Active Job Continuations [ci skip]

satyakampandya:
### Motivation / Background

This Pull Request adds documentation for the new Active Job Continuations feature introduced in [rails/rails#55127](https://github.com/rails/rails/pull/55127).

### Detail

This Pull Request updates the Active Job Basics guide to include a new Job Continuations section.

### Checklist

### DIFF
--- a/guides/source/active_job_basics.md
+++ b/guides/source/active_job_basics.md
@@ -719,7 +719,7 @@ Each step runs sequentially. If the job is interrupted between steps, or within 
 step that uses a cursor, the job resumes from the last recorded position. This
 makes it easier to build long-running or multi-phase jobs that can safely pause
 and resume without losing progress.
-For more details, see [ActiveJob::Continuable](https://edgeapi.rubyonrails.org/classes/ActiveJob/Continuation.html).
+For more details, see [ActiveJob::Continuable](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
 
 Callbacks
 ---------
--- a/guides/source/active_job_basics.md
+++ b/guides/source/active_job_basics.md
@@ -719,7 +719,7 @@ Each step runs sequentially. If the job is interrupted between steps, or within 
 step that uses a cursor, the job resumes from the last recorded position. This
 makes it easier to build long-running or multi-phase jobs that can safely pause
 and resume without losing progress.
-For more details, see [ActiveJob::Continuable](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
+For more details, see [ActiveJob::Continuation](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
 
 Callbacks
 ---------
--- a/guides/source/active_job_basics.md
+++ b/guides/source/active_job_basics.md
@@ -673,13 +673,14 @@ number as more important.
 
 Job Continuations
 -----------------
+
 Jobs can be split into resumable steps using continuations. This is useful when
-a job may be interrupted - for example, during queue shutdown. When using 
-continuations, the job can resume from the last completed step, avoiding the 
+a job may be interrupted - for example, during queue shutdown. When using
+continuations, the job can resume from the last completed step, avoiding the
 need to restart from the beginning.
 
-To use continuations, include the `ActiveJob::Continuable` module. You can then 
-define each step using the `step` method inside the `perform` method. Each step can 
+To use continuations, include the `ActiveJob::Continuable` module. You can then
+define each step using the `step` method inside the `perform` method. Each step can
 be declared with a block or by referencing a method name.
 
 ```ruby
@@ -713,9 +714,10 @@ class ProcessImportJob < ApplicationJob
     end
 end
 ```
+
 Each step runs sequentially. If the job is interrupted between steps, or within a
-step that uses a cursor, the job resumes from the last recorded position. This 
-makes it easier to build long-running or multi-phase jobs that can safely pause 
+step that uses a cursor, the job resumes from the last recorded position. This
+makes it easier to build long-running or multi-phase jobs that can safely pause
 and resume without losing progress.
 For more details, see [ActiveJob::Continuable](https://edgeapi.rubyonrails.org/classes/ActiveJob/Continuation.html).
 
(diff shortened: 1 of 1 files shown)


guides/source/active_job_basics.md
```diff
@@ -671,6 +671,56 @@ number as more important.
...
+      @import.finalize
+    end
+end
+```
+
+Each step runs sequentially. If the job is interrupted between steps, or within a
+step that uses a cursor, the job resumes from the last recorded position. This
+makes it easier to build long-running or multi-phase jobs that can safely pause
+and resume without losing progress.
+For more details, see [ActiveJob::Continuable](https://edgeapi.rubyonrails.org/classes/ActiveJob/Continuation.html).
```
zzak:
```suggestion
For more details, see [ActiveJob::Continuable](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
```

Knit: we only link to `api.` in the guides source, there is a method for converting to `edge` when building the edge docs automatically.

satyakampandya: @zzak Thanks for the clarification! I've updated the link accordingly to api.rubyonrails.org. 🙌

guides/source/active_job_basics.md
```diff
@@ -671,6 +671,56 @@ number as more important.
...
+      end
+    end
+
+    # Step defined by referencing a method
+    step :finalize
+  end
+
+  private
+    def finalize
+      @import.finalize
```
MiroslavCsonka, line 713: I'm not sure I understand this completely, or if our setup (SolidQueue) is incorrect, but what would happen if the finalize method takes 10 minutes and a deployment interrupts it? Like a long SQL query or a huge file download. Would the scheduler just kill the job, not record any continuations, and it would just start from scratch?

It's my understanding that it's `step`, `.checkpoint!`, `.advance!`, and `.set!` that are responsible for checking whether or not we are in the process of shutting down, and if we are, they attach metadata to the job and reschedule it.

This means that if your SolidQueue shutdown_timeout config is 5 seconds (the default), the above methods have to be called at least every 5 seconds; otherwise, the whole progress is potentially lost, right?

guides/source/active_job_basics.md
```diff
@@ -671,6 +671,56 @@ number as more important.
...
+    end
+end
+```
+
+Each step runs sequentially. If the job is interrupted between steps, or within a
+step that uses a cursor, the job resumes from the last recorded position. This
+makes it easier to build long-running or multi-phase jobs that can safely pause
+and resume without losing progress.
+For more details, see [ActiveJob::Continuable](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
```
p8: You are linking to Continuation, so the label should reflect that:
```suggestion
For more details, see [ActiveJob::Continuation](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html).
```

satyakampandya: @p8 Good point - I’ve updated the label to match the ActiveJob::Continuation class name.

This is actually my first PR to Rails, so I really appreciate these suggestions - I’m still getting familiar with the conventions we follow for guides. Thanks a lot for your help!

guides/source/active_job_basics.md
satyakampandya: As far as I understand, yes - if the `finalize` method takes 10 minutes and gets interrupted by a deployment, and no continuation methods like `step`, `.checkpoint!`, `.advance!`, or `.set!` are called during that time, then the job might just get stopped without saving any progress. That would mean it could start over from the beginning, depending on how retries are set up.

So yeah, you're right, for long-running tasks, it seems important to call one of those methods regularly to make sure progress isn’t lost if the job shuts down mid-way.

This feature was added recently by @djmb in [PR #55127](https://github.com/rails/rails/pull/55127), so tagging him in case he’d like to chime in and clarify any edge cases or gotchas I might have missed. 🙏

djmb: Yes that's right - checkpoints need to be created by the job, so if you don't create them often enough then the progress could be lost.

Really I guess the answer is not to have single actions that take that long. It's a good constraint to work with. SQL can be optimized and downloads can maybe by split by using range requests so there's often a way. Long running things that can't be split up (say shelling out to ffmpeg or gzip or something like that) could be run in a separate thread.

But I'm sure there will be cases where that's not possible or desirable. Maybe we need a mechanism to deliberately interrupt the job to save progress before doing something slow. I'll have a think about that.

MiroslavCsonka: Thanks, gentlemen, for the clarification. That's the behaviour I've been seeing in our production job.

It takes 6 hours, but basically, it downloads tens of GB of zip files from S3, unwraps them, imports them into the database, and then recomputes some statistics on top of it. Everything, except the last stage, could be done in 5-second chunks, but the last bit is a problem since if that gets interrupted, it has to start from scratch (redownload, unzip, parse CSV, etc.). The recomputation is the problem since it's a single materialised view, which takes 90 minutes at the moment, so it's hard to say if we can do it within 5 seconds. We would need to improve the performance at least 1000 times 😅

I liked the idea of steps, so I came up with [this gist](https://gist.github.com/MiroslavCsonka/c13674099ce1366d2e774bc05dcd1fe1). It does a "step" (I called it stage so it doesn't clash with Rails), but the idea is to schedule another job to do the next stage.

djmb: Thanks for that pointer - @MiroslavCsonka. I think this is going to be a common problem, so inspired by your gist I've added a [PR for isolated steps](https://github.com/rails/rails/pull/55212).

MiroslavCsonka: Oh nice! Once the CI build is green, happy to upgrade our prod and test it out

satyakampandya:
@djmb I noticed that Rails 8.1 has been released with Active Job continuations, but I still don't see the related documentation updates in the edge guides on the Rails site. Just wanted to check if there's anything missing or if I might've overlooked something in this PR. I'm happy to make any adjustments if needed. Please let me know your thoughts when you get a chance. 
Thanks!

PR #55194, opened 2025-06-14
Enable configuring Action View's render tracker

skipkayhil:
### Motivation / Background

Ref #55161

In Rails 7, a new render tracking implementation was [added][1] to Rails: the RipperTracker. From that commit message:

> Using a parser finds dependencies that would otherwise be difficult to
> find with the regular expressions. It should also theoretically work
> with other template systems since it operates on the compiled template
> instead of the contents of the file.

It was initially made the default render tracker, but that change was later [backed out][2].

Since then, the Prism parser was released and made the default Ruby parser. A Prism render tracker implementation was also [added][3] to Rails in 7.2.

### Detail

Until now, there hasn't been a "blessed" way for applications to try out the new tracker implementations (you would have to go through private API). This commit adds configuration to not only make it easy for applications to use the new Ruby-parser-based trackers, but it also makes these new trackers the default for new applications on 8.1.

The new trackers are not only better at finding `render` calls, they are also much easier to maintain. When I [added][4] trailing interpolation tracking to the three trackers, the Prism tracker was by far the easiest to work with.

[1]: https://github.com/rails/rails/commit/897b9bf6e29a9aa7c04d3e4cb96a65665565154d
[2]: https://github.com/rails/rails/commit/41c0681fb7626d036005f8dec559a507ff143874
[3]: https://github.com/rails/rails/commit/d743a2b10fb50dc4702ba4e73955c4f94b3bf700
[4]: https://github.com/rails/rails/commit/13df150c06e37a7024baa655000ae2d7eff9fc0c

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Enable configuring the strategy for tracking dependencies between Action
+    View templates.
+
+    The existing `:regex` strategy is kept as the default, but with
+    `load_defaults 8.1` the strategy will be `:ruby` (using a real Ruby parser).
+
+    *Hartley McGuire*
+
 *   Introduce `relative_time_in_words` helper
 
     ```ruby
--- a/actionview/lib/action_view.rb
+++ b/actionview/lib/action_view.rb
@@ -91,6 +91,9 @@ module ActionView
   autoload :CacheExpiry
   autoload :TestCase
 
+  singleton_class.attr_accessor :render_tracker
+  self.render_tracker = :regex
+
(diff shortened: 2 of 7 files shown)


tenderlove:
This makes sense to me.  Is the plan that we would make the Ruby parser implementation default? I think that we should.

Also what about #55165? It seems like we should merge it as well.

tenderlove:
> Is the plan that we would make the Ruby parser implementation default?

🤦 I misread the PR. I see that indeed this makes the Ruby parser implementation default.  I think that's good, lets ship it.

PR #55195, opened 2025-06-14
[ci skip] Print the record count in the generator guide template override example

murjax:
### Motivation / Background

While following the "Creating and Customizing Rails Generators & Templates" guide on overriding a Rails generator template, I observed the record count in the template did not render as expected. After running the scaffold generator, the resulting `index.html.erb` file does not actually print the count, but only calls it.

### Detail

This Pull Request changes the guide's example to use `<%=` tags so the record count is actually printed.

### Checklist


skipkayhil:
Nice catch, thanks for contributing to Rails 🎉

PR #55196, opened 2025-06-15
Fix `include?` on strict-loaded HABTM association

### LINKED ISSUE #55183: Strict loading error raised incorrectly when checking if HABTM association includes an unpersisted instance
### Steps to reproduce
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end

  create_table :tags, force: true do |t|
  end

  create_join_table :posts, :tags, force: true do |t|
  end
end

  has_and_belongs_to_many :tags
end

  has_and_belongs_to_many :posts
end

  def test_habtm_strict_loading_include
    post = Post.create!
    tag = Tag.create!
    post.tags << tag

    assert Post.includes(:tags).first.tags.include?(tag)
    assert_not Post.includes(:tags).first.tags.include?(Tag.new)

    assert Post.includes(:tags).strict_loading.first.tags.include?(tag)
    assert_not Post.includes(:tags).strict_loading.first.tags.include?(Tag.new)
  end
end
```

### Expected behavior

Test passes.

### Actual behavior

The first three assertions succeed, but the last one (checking if an unpersisted model instance is a member of a pre-loaded association) raises a StrictLoadingViolationError. Sample output:
```
ActiveRecord::StrictLoadingViolationError: `Post` is marked for strict_loading. The Post::HABTM_Tags association named `:posts_tags` cannot be lazily loaded.
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/core.rb:257:in `strict_loading_violation!'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/association.rb:250:in `find_target'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_association.rb:274:in `load_target'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_proxy.rb:45:in `load_target'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_proxy.rb:1025:in `records'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/relation/delegation.rb:101:in `each'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/relation.rb:394:in `any?'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/relation.rb:394:in `any?'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_association.rb:510:in `include_in_memory?'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_association.rb:263:in `include?'
    /usr/local/rvm/gems/ruby-3.3.6/gems/activerecord-8.0.2/lib/active_record/associations/collection_proxy.rb:928:in `include?'
    habtm_strict_bug.rb:62:in `test_habtm_strict_loading_include'
```

This also occurs if the `has_and_belongs_to_many` is changed to a `has_many :through`. 

It's possible to work around this by adding additional associations to the `includes()` call, but it doesn't seem like that should be necessary.

### System configuration

**Rails version**: 8.0.2 (also occurs on `main` and goes back to 7.2 and before)

**Ruby version**: 3.3 in this case, but I don't think it matters much

skipkayhil:
### Motivation / Background

Fix #55183

Previously, trying to check whether a `new_record?` (`Post.new`) is `include?`d in a strict-loaded HABTM association would raise a `StrictLoadingViolationError` even when the HABTM association is already loaded.

### Detail

This commit fixes the issue by always using the `target` if the association is `loaded?` instead of querying the Through association (which is only needed when the association is dirty/has newly built records).

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/collection_association.rb
+++ b/activerecord/lib/active_record/associations/collection_association.rb
@@ -259,10 +259,10 @@ module ActiveRecord
         klass = reflection.klass
         return false unless record.is_a?(klass)
 
-        if record.new_record?
-          include_in_memory?(record)
-        elsif loaded?
+        if loaded?
           target.include?(record)
+        elsif record.new_record?
+          include_in_memory?(record)
         else
           record_id = klass.composite_primary_key? ? klass.primary_key.zip(record.id).to_h : record.id
           scope.exists?(record_id)
--- a/activerecord/test/cases/strict_loading_test.rb
+++ b/activerecord/test/cases/strict_loading_test.rb
@@ -636,6 +636,14 @@ class StrictLoadingTest < ActiveRecord::TestCase
     end
   end
 
+  def test_does_not_raise_when_checking_if_new_record_included_in_eager_loaded_habtm_relation
+    Developer.first.projects << Project.first
+
+    developer = Developer.includes(:projects).strict_loading.first
+
+    assert_nothing_raised { developer.projects.include?(Project.new) }
+  end
+
   def test_strict_loading_violation_raises_by_default
     assert_equal :raise, ActiveRecord.action_on_strict_loading_violation
 


maxfelsher:
Thanks for the fix! I confirmed that my test script from #55183 passes when running against your branch.

One thing that confused me when I was trying to debug this: it looks like for some reason adding `strict_loading` causes the "through" association to no longer be `loaded?`. That is, using the setup from the test script:

```ruby
Post.includes(:tags).first.association(:posts_tags).loaded?  # => true
Post.includes(:tags).strict_loading.first.association(:posts_tags).loaded?  # => false
```

I couldn't figure out what exactly was causing `strict_loading` to make the through association not be `loaded?`. Though with the change here, it seems like it might not be necessary to address that to fix the original issue.

skipkayhil:
> I couldn't figure out what exactly was causing `strict_loading` to make the through association not be `loaded?`.

Because I was curious, I tracked it down:

The queried records don't get set on the association because `@associate` here is `false`: https://github.com/rails/rails/blob/5a7a59ab4a9ac34b8cadf9d8ce7b554da50c5e43/activerecord/lib/active_record/associations/preloader/association.rb#L141-L143

And the difference is because `preload_scope.empty_scope?` here returns `false`: https://github.com/rails/rails/blob/5a7a59ab4a9ac34b8cadf9d8ce7b554da50c5e43/activerecord/lib/active_record/associations/preloader/association.rb#L110

Because `strict_loading` gets stored in the `Relation`'s `@values`, making it not match the "default" `values` of a `Relation`: https://github.com/rails/rails/blob/5a7a59ab4a9ac34b8cadf9d8ce7b554da50c5e43/activerecord/lib/active_record/relation.rb#L1318

eileencodes:
I backported this to 8-0-stable. 7-2-stable has a conflict, do you want to take a look @skipkayhil?

PR #50887, opened 2024-01-26
Add row_count field to sql.active_record notification

marvin-bitterlich:
### Motivation / Background

In Intercom we are working on detecting queries with large result sets, as Vitess based databases (like in our case Planetscale) block queries that return more than (by default) 100k result set rows.

We have an internal patch for this that works well for us, but thought to contribute this upstream, so others can benefit

### Detail

This Pull Request changes `{*}Adapter.log`, which emits `sql.active_record` notifications.

Previously the instrument call only logs information that goes into a request, but allows the caller inside the block to modify the payload. We make use of this inside each adapter to add a new field to each `sql.active_record` notification with the number of result rows being returned in each query.

This allows consumers of this notification [(example)](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/active_record/events/sql.rb) to consume this metric. 

There is previous work in this regard with the [`RecordFetchWarning`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/record_fetch_warning.rb) module that can be included, which demonstrates the need to monitor queries with large result sets. In our case these log lines weren't easy to integrate into our observability stack, and having a numerical value attached to each sql trace allowed us to use the full power of our observability suite. 

This pr also includes three tests in the instrumentation test suite that verify this working for model instanciation queries, direct value queries like `pluck` and raw `ActiveRecord::Base.connection.execute(sql)` queries.

### Performance

The change asks the resultset how many elements it has. Expectation was that that should be a constant lookup, and benchmarks suggest this to be true:

I have tested this change with `examples/performance` on adapters `sqlite3`, `mysql2` and `postgresql` with no measurable difference in performance (less than 0.1% when standard deviation was 1%).

### Additional information

I originally approached this inside `abstract_adapter` but it turns out every adapters Result uses a different method, so I moved the call into each adapters respective usage of `log` 

[`RecordFetchWarning`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation/record_fetch_warning.rb) works by patching `exec_queries`, so it aggregates the result set size of multiple queries into one, which is less helpful for this case, as the limits imposed by Vitess are per query. Additionally, this does not support the case of executing raw SQL on the connection, which is something that this patch captures.

As for the field name on the payload `row_count` seemed like a good name as it follows the same pattern as `result_count` does in https://guides.rubyonrails.org/active_support_instrumentation.html#instantiation-active-record

### Checklist

Before submitting the PR make sure the following are checked:

[...]

PR #55200, opened 2025-06-17
Use ntuples to populate row_count instead of count for Postgres

jcalvert:
### Motivation / Background

`PG::Result#ntuples` ([doc](https://deveiate.org/code/pg/PG/Result.html#method-i-ntuples)) uses a libpq function for getting the number of rows returned from a query, whereas `#count` is provided by `Enumerable` and thus iterates through the entire result set. 

### Detail

I first noticed this because upgrading an application where #50887 was added resulted in a performance regression for a number of queries. For larger result sets, iterating through the result set via `#count` makes a noticeable performance difference, whereas `ntuples` will read the returned count directly from the result which is sent as part of the underlying protocol.

### Additional information

Benchmarking with 100k queries that retrieve 10 rows in my development instance this change was 18% faster in wall clock time and a whopping 44% less user CPU time. This also results in fewer `T_DATA` allocations in ObjectSpace to be garbage collected later. The performance of iterating through the results is linear time complexity so with more rows the difference is more pronounced 

```
ActiveRecord::Base.with_connection do |conn|
  Benchmark.ips do |x|
    x.report("count") do
      conn.raw_execute("SELECT * FROM schema_migrations", "SQL")
    end
    x.hold! "count_ntuple" # apply patch here
    x.report("ntuple") do
      conn.raw_execute("SELECT * FROM schema_migrations", "SQL")
    end
  end
end
```

With a 1000 rows gives me 

```
             ntuples   487.000  i/100ms
             ntuples      4.921k (± 4.2%) i/s -     24.837k in   5.058571s

Comparison:
             ntuples:     4920.6 i/s
               count:     2309.9 i/s - 2.13x  slower
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Use ntuples to populate row_count instead of count for Postgres
+
+    *Jonathan Calvert*
+
 *   Fix checking whether an unpersisted record is `include?`d in a strictly
     loaded `has_and_belongs_to_many` association.
 
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -165,7 +165,7 @@ module ActiveRecord
             verified!
 
             notification_payload[:affected_rows] = result.cmd_tuples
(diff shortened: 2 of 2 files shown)


skipkayhil (approved):
Great find (and great first contribution)!

PR #50608, opened 2024-01-05
allow dom_id method to accept an array of records_or_classes

jwilsjustin:
# Context

See also: https://github.com/hotwired/turbo-rails/pull/476

[hotwired/turbo-rails/pull/476](https://github.com/hotwired/turbo-rails/pull/476) removed the ability to pass multiple ids/models to the `turbo_frame_tag` helper method. In that PR it was suggested by @skipkayhil that this ability be passed down to the `dom_id` method (which is a part of rails core) method rather than the `turbo_frame_tag` helper (which is a part of turbo-rails).

ref: https://discuss.rubyonrails.org/t/allow-dom-id-method-to-accept-multiple-ids-models/84408

# Changes

If the first argument of the `dom_id` is an array, this change will loop through each value and form a singular dom ID, then join the parts together. This change will allow turbo frame tags to accept multiple models (similar to the `cache` helper from ActionView).

```
<%= turbo_frame_tag [customer, field] do %>
  <!-- code -->
<% end %>
```

***

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #44081, opened 2022-01-05
Extend `dom_id` and `dom_class` to accept var-args

seanpdoyle:
### Summary

Changes `dom_id` and `dom_class` to accept a variable set of prefixes:

```ruby
dom_class @post, :first_prefix, :second_prefix
  # => "first_prefix_second_prefix_post"
dom_class @post, [:first_prefix, :second_prefix]
  # => "first_prefix_second_prefix_post"
dom_id @post, :first_prefix, :second_prefix
  # => "first_prefix_second_prefix_post_123"
dom_id @post, [:first_prefix, :second_prefix]
  # => "first_prefix_second_prefix_post_123"
```

These changes align with the [FormBuilder#field_id][] interface, which
accepts multiple `suffixes` arguments.

[FormBuilder#field_id]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html#method-i-field_id

PR #55204, opened 2025-06-17
Add `dom_target` helper to create `dom_id`-like strings from an unlimited number of objects

bensheldon:
### Motivation / Background

This Pull Request has been created because it's increasingly necessary to create complex DOM identifiers and the existing `dom_id` is limited.

This PR creates a new method `ActionView::RecordIdentifier#dom_target` that can create a unique identifier from an unlimited number of objects, which differentiates it from `dom_id` which only takes a single object and an optional string prefix.

```ruby
dom_target(Post.find(45))                  # => "post_45"
dom_target(Post.find(45), :edit)           # => "post_45_edit"
dom_target(Post.find(45), :edit, :special) # => "post_45_edit_special"
dom_target(Post.find(45), Comment.find(1)) # => "post_45_comment_1"
```

I chose the name of `target` because I found myself wanting this for generating `turbo-rails` Turbo Broadcast targets. `dom_target` is inspired by the structure of `turbo_stream_from(*streamables)`, which concatenates an unlimited number of objects into a single identifier.

Adding an new helper method is proposed as an alternative to adding new behavior to existing `dom_id` or `dom_class` helper methods, which had questions/objections. Related:

- #44081
- #50608
- https://github.com/hotwired/turbo-rails/issues/543

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Add `dom_target` helper to create `dom_id`-like strings from an unlimited
+    number of objects.
+
+    *Ben Sheldon*
+
 *   Respect `html_options[:form]` when `collection_checkboxes` generates the
     hidden `<input>`.
 
--- a/actionview/lib/action_view/record_identifier.rb
+++ b/actionview/lib/action_view/record_identifier.rb
@@ -101,6 +101,27 @@ module ActionView
       end
     end
 
+    # The DOM target convention is to concatenate any number of parameters into a string.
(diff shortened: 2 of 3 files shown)


bensheldon:
I can fix up the changelog.

I want to change one behavior I noticed that I want to fix before it's baked:

```ruby
dom_target(User) #=> "user" 
dom_target(User.new) #=> "new_user"
```

Edit: ✅

byroot:
I did a small change to use a `case/when`, other than that, seems sensible to me.

PR #55135, opened 2025-06-01
[RF-DOCS] Move PostgreSQL Guide into the API and other guides

Ridhwana:
:link: Preview:
https://ridhwana-ridhwana-postgresql.rails-docs-preview.pages.dev

### Motivation / Background

[Active Record and PostgreSQL](https://guides.rubyonrails.org/active_record_postgresql.html) is a WIP guide. 

However, we want to remove database-specific guides like this one, because Active Record should be database-agnostic. Hence, I moved DataTypes and the other parts of the PostgreSQL guide to the API docs so that we can remove the PostgreSQL Guide.

### Detail

This Pull Request changes:

- [x] Datatypes moved to the [API docs](https://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQL/ColumnMethods.html) and we fill out the details for other methods in this doc.
- [x] Structure Dumps are merged into https://edgeguides.rubyonrails.org/active_record_migrations.html#schema-dumping-and-you
- [x] Added some additional guidance on schema.rb vs structure.sql in  the migration guide.
- [x] Full Text Search has been removed.
- [x] Composite Types and Database View has moved to Active Record Migrations 
- [x] Added Buffers explanation to the Active Record Querying
- [x] Deferrable Foreign Keys are added to Association Basics
- [x] Virtual Columns and Arrays need to be decided upon.

### Checklist

Before submitting the PR make sure the following are checked:

* [ ] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [ ] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55207, opened 2025-06-18
Remove documentation for non existing ColumnMethods#column [ci skip]

p8:
### Motivation / Background

The `column` method is [defined](https://github.com/rails/rails/blob/1921f0fef60983d6423b937884de5e83b2619960/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L495) (and documented) on `TableDefinition` not on `ColumnMethods`.
So there is no need to document it on `ColumnMethods`.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb
@@ -328,17 +328,6 @@ module ActiveRecord
         column(name, type, **options, primary_key: true)
       end
 
-      ##
-      # :method: column
-      # :call-seq: column(name, type, **options)
-      #
-      # Appends a column or columns of a specified type.
-      #
-      #  t.string(:goat)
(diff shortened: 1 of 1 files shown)


Edouard-chin (commented):
Make sense to me !

Side note (maybe we can improve it in another PR?), what do you think about documenting all column methods dynamically generated?
Adapters already does that for specific columns but for the core ones (e.g. boolean, integer...) it doesn't show up in the doc.

https://github.com/rails/rails/blob/1921f0fef60983d6423b937884de5e83b2619960/activerecord/lib/active_record/connection_adapters/postgresql/schema_definitions.rb#L141-L171

p8:
> Side note (maybe we can improve it in another PR?), what do you think about documenting all column methods dynamically generated?

Yes, good point. They are getting documented here (which is where I noticed the documentation of column):
https://github.com/rails/rails/pull/55135

Edouard-chin (approved):

PR #51078, opened 2024-02-14
Fix override existing join types in the query in the `where.associated` method

saleh-alhaddad:
**Motivation / Background**

Currently, the `where.associated` method in Rails only allows filtering associated records with an `inner join`, which can potentially override existing join types in the query. The fix ensures that the association is joined using the appropriate join type (either inner join or left outer join) based on the existing joins in the scope. This prevents unintentional overrides of existing join types and ensures consistency in the generated SQL queries.

Example:

```ruby
 User.left_outer_joins(:orders).where.associated(:orders).count
```
Old query
```sql
SELECT COUNT(*) FROM "users" INNER JOIN "orders" ON "orders"."user_id" = "users"."id" WHERE "orders"."id" IS NOT NULL
```
after fixing the issue, the query:
New Query
```sql
SELECT COUNT(*) FROM "users" LEFT OUTER JOIN "orders" ON "orders"."user_id" = "users"."id" WHERE "orders"."id" IS NOT NULL
```

This enhancement empowers developers with greater flexibility in querying associated data, facilitating scenarios such as:

- Retrieving posts with authors, even if some posts lack an assigned author (utilizing left_joins).
- Obtaining comments associated with specific users, encompassing orphaned comments without a user (employing multiple joins with varied types).

**Detail**

The pivotal improvement in this Pull Request is the inclusion of a check ensuring that associations are joined using the appropriate join type (either inner join or left outer join) based on existing joins within the scope. This safeguards against unintentional overrides of existing join types and fosters consistency in the generated SQL queries.

**Additional Information**

This refinement is informed by common requirements articulated by Rails developers in various forums and discussions. Analogous functionality is prevalent in certain popular ORM extensions, underscoring its utility. The implementation is crafted to preserve clarity and coherence within the existing codebase.

Note: This is a draft description and should be adapted to your specific implementation details and the project's context.

FYI: After merging this PR, I will check the `missing` method and fix this issue in a separate PR alternative same PR.

**Performance:**

- Benchmark Before the fix:

```
Benchmark.ips do |x| 
  x.report('joins') { User.joins(:comments, :orders).where.associated(:orders) }
  x.report('left_joins') { User.left_joins(:comments, :orders).where.associated(:orders) }
  x.report('left_outer_joins') { User.left_outer_joins(:comments, :orders).where.associated(:orders) }
  x.report('with out joins') { User.where.associated(:orders) }
end
```

```info
Warming up --------------------------------------
  joins                        3.260k i/100ms
[...]

PR #55210, opened 2025-06-19
[7-2-stable] Clean up changelog for where.associated scope bug

zzak:
Source: #51078

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -862,19 +862,15 @@
 
     *Jason Nochlin*
 
-*   The fix ensures that the association is joined using the appropriate join type
-    (either inner join or left outer join) based on the existing joins in the scope.
-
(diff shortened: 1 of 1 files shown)


activerecord/CHANGELOG.md
```diff
@@ -862,20 +862,17 @@
...
     Example:
 
-
-
     ```ruby
     # `associated` will use `LEFT JOIN` instead of using `JOIN`
     Post.left_joins(:author).where.associated(:author)
     ```
 
+    Previously, the `LEFT OUTER JOIN` would be lost and converted to an `INNER JOIN`.
```
p8, line 873: This line is very similar to the comment on line 870. Maybe combine them?

zzak: Whoops, yeah I just removed the comment because it was confusing.

Thanks for checking!

PR #55211, opened 2025-06-19
[Fix #55206] Granular connection switching does not respect primary abstract class

### LINKED ISSUE #55206: Confusing transaction behaviour with global and granular connection switching
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails"

  gem "pg", "~> 1.1"
end

require "minitest/autorun"

ENV["DATABASE_URL"] = "postgresql:postgres"

  config.eager_load = false

end
Rails.application.initialize!

  end

    t.integer :post_id
  end
end

  self.abstract_class = true

  connects_to database: { writing: :primary, reading: :primary }
end

class SecondaryRecord < ApplicationRecord
  self.abstract_class = true

  connects_to database: { writing: :primary, reading: :primary }
end

end

  def test_association_stuff
    ActiveRecord::Base.connected_to(role: :reading) do
      ApplicationRecord.connected_to(role: :writing) do
        ActiveRecord::Base.transaction do # t1 is created using a ActiveRecord::Base reading connection
          Post.create! # t2 is created using a ApplicationRecord writing connection

          raise ActiveRecord::Rollback # t1 is rolled back, t2 is not
        end
      end
    end

    assert_equal 0, Post.count # ❌
  end
end
```

This doesn't seem that surprising when you know what's going on, but the way we ran into this was:
1. Someone wrote this code (possibly before we had multiple dbs), and it worked fine:

   ```ruby
      Post.create! 

      raise ActiveRecord::Rollback # just to mimic a failed insert
    end
   ```
2. In certain contexts, we want to use the reader role for all connections, so we introduce:

    ```ruby
    ActiveRecord::Base.connected_to(role: :reading)
    ```
3. We identify that this write should be allowed even in this context, so we only switch to the writing connection granularly using model's abstract parent:

    ```ruby
    ActiveRecord::Base.connected_to(role: :reading)
      ApplicationRecord.connected_to(role: :writing) do
    ```

This ends up causing the `Post.create!` to no longer use an `ActiveRecord::Base` connection object but a `ApplicationRecord` connection object, resulting in a nested transaction being created.

### Expected behavior

(Maybe) `Post.create!` should raise an error stating that creating a nested (in this case, implicit) transaction within a 'global' transaction isn't allowed.

### Actual behavior

What seems like an insert that should roll back, doesn't. Keep in mind that the connection switching blocks may be in different files e.g. middleware, and this might not be super obvious when reading just the transaction block.

### System configuration

**Rails version**:  8.0.2, edge

**Ruby version**: 3.4.4

joshuay03:
### Motivation / Background

Closes https://github.com/rails/rails/issues/55206

### Detail

Updates `ActiveRecord::Core.current_{role,shard,preventing_writes}` and `ActiveRecord::Core.preventing_writes?` to check if the `connected_to_stack` entry being evaluated has a `:klasses` entry that is the primary abstract class if the receiver is `ActiveRecord::Base`, as they should be sharing the same connection.

### Checklist

### DIFF
--- a/activerecord/test/cases/connection_adapters/connection_swapping_nested_test.rb
+++ b/activerecord/test/cases/connection_adapters/connection_swapping_nested_test.rb
@@ -488,6 +488,34 @@ module ActiveRecord
           Object.send(:remove_const, :ReloadedRecord)
           ActiveRecord::Base.establish_connection :arunit
         end
+
+        def test_base_role_and_preventing_writes_with_primary_abstract_class
+          Object.const_set(:ApplicationRecord, ApplicationRecord)
+
+          assert_predicate ApplicationRecord, :application_record_class?
+          assert_predicate ApplicationRecord, :primary_class?
+          assert_predicate ApplicationRecord, :abstract_class?
+
+          ApplicationRecord.connects_to(database: { writing: :arunit, reading: :arunit })
+
+          # Switch everything to reading
+          ActiveRecord::Base.connected_to(role: :reading) do
+            # Switch primary abstract class, and therefore ActiveRecord::Base, to writing
+            ApplicationRecord.connected_to(role: :writing) do
+              assert_equal :writing, ActiveRecord::Base.current_role
+              assert_equal :writing, ApplicationRecord.current_role
+              assert_not ActiveRecord::Base.current_preventing_writes
+              assert_not ApplicationRecord.current_preventing_writes
+              assert_not ActiveRecord::Base.preventing_writes?(ActiveRecord::Base.connection_pool.connection_descriptor.name)
+              assert_not ActiveRecord::Base.preventing_writes?(ApplicationRecord.connection_pool.connection_descriptor.name)
+            end
+          end
+        ensure
+          ApplicationRecord.remove_connection
+          ActiveRecord.application_record_class = nil
+          Object.send(:remove_const, :ApplicationRecord)
+          ActiveRecord::Base.establish_connection :arunit
+        end
       end
     end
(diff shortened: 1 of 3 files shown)


activerecord/test/cases/connection_adapters/connection_swapping_nested_test.rb
```diff
@@ -488,6 +488,30 @@ def test_prevent_writes_handles_class_reloading
...
+          assert_predicate ApplicationRecord, :primary_class?
+          assert_predicate ApplicationRecord, :abstract_class?
+
+          ApplicationRecord.connects_to(database: { writing: :arunit, reading: :arunit })
+
+          # Switch everything to reading
+          ActiveRecord::Base.connected_to(role: :reading) do
+            # Switch primary abstract class, and therefore ActiveRecord::Base, to writing
+            ApplicationRecord.connected_to(role: :writing) do
+              assert_equal :writing, ActiveRecord::Base.current_role
```
joshuay03, line 505: Without patch:

Screenshot 2025-06-19 at 8 33 15 pm

joshuay03: I've also verified that my script in https://github.com/rails/rails/issues/55206 passes against this branch.

zzak:
It's possible someone is depending on this behavior so it might need a flag, but let's wait to see what core thinks. :bow:

PR #55214, opened 2025-06-20
Enable `debug` `no_repeat` config in generated app

joshuay03:
### Motivation / Background

Addresses https://github.com/ruby/debug/issues/1098#issuecomment-2394771899

### Detail

https://github.com/ruby/debug/pull/1140 introduced a new `no_repeat` config option (see https://github.com/ruby/debug/issues/1098 for context) which is included in the [latest release](https://github.com/ruby/debug/releases/tag/v1.11.0). `debug` [isn't pinned to a specific version in the generated Gemfile](https://github.com/rails/rails/blob/aa66aece44f191e6330c04ce4942c3edf31d85d7/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L56), so we can start using this in new apps.

### Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,7 +198,7 @@ GEM
       railties (>= 6.0.0)
       sass-embedded (~> 1.63)
     date (3.4.1)
-    debug (1.10.0)
+    debug (1.11.0)
       irb (~> 1.10)
       reline (>= 0.3.8)
     declarative (0.0.20)
--- a/railties/lib/rails/generators/rails/app/templates/config/initializers/debug.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/initializers/debug.rb.tt
@@ -1,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+if Rails.env.local?
+  require "debug/config"
+
+  # See https://github.com/ruby/debug#configuration-list for more options.
+  DEBUGGER__::CONFIG[:no_repeat] = true
+end
--- a/railties/lib/rails/generators/rails/plugin/templates/Gemfile.tt
+++ b/railties/lib/rails/generators/rails/plugin/templates/Gemfile.tt
@@ -14,8 +14,8 @@ gem "rubocop-rails-omakase", require: false
 <%- end -%>
 <% if RUBY_ENGINE == "ruby" -%>
 
-# Start debugger with binding.b [https://github.com/ruby/debug]
-# gem "debug", ">= 1.0.0"
+# See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
+gem "debug", platforms: %i[ mri windows ], require: "debug/prelude"
 <% end -%>
 <% if RUBY_PLATFORM.match?(/mingw|mswin|java/) -%>
 
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Generate an initializer that configures `debug` and enables the `no_repeat` option when creating new applications.
+
+    *Joshua Young*
+
 *   Remove unnecessary `ruby-version` input from `ruby/setup-ruby`
 
     *TangRufus*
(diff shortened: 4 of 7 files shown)


joshuay03:
Hmm, I tested this out with a new app and it worked as expected, but CI disagrees.

railties/lib/rails/generators/rails/plugin/templates/Gemfile.tt
```diff
@@ -15,7 +15,7 @@ gem "rubocop-rails-omakase", require: false
 <% if RUBY_ENGINE == "ruby" -%>
 
 # Start debugger with binding.b [https://github.com/ruby/debug]
-# gem "debug", ">= 1.0.0"
+gem "debug", ">= 1.0.0"
```
joshuay03, line 18: This is the only commented out gem in plugins, and [it isn't commented out in apps](https://github.com/rails/rails/blob/aa66aece44f191e6330c04ce4942c3edf31d85d7/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L56), so I opted to un-comment it here rather than have a separate conditional initializer just for plugins that does something hacky like check the loaded gems first.

railties/lib/rails/generators/rails/app/templates/config/initializers/debug.rb.tt
joshuay03, line 1: I think this is the easiest way to configure `debug` without messing around with environment variables (ref: https://github.com/ruby/debug/issues/1098).

Gemfile.lock
```diff
@@ -198,7 +198,7 @@ GEM
       railties (>= 6.0.0)
       sass-embedded (~> 1.63)
     date (3.4.1)
-    debug (1.10.0)
+    debug (1.11.0)
```
joshuay03, line 201: > Hmm, I tested this out with a new app and it worked as expected, but CI disagrees.

It seems some of the railties tests are picking up the framework Gemfile.lock, so we need to bump this to get CI green.

railties/lib/rails/generators/rails/app/templates/config/initializers/debug.rb.tt
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+if Rails.env.local?
+  require "debug"
+
+  # See https://github.com/ruby/debug#configuration-list for more options.
+  DEBUGGER__::CONFIG[:no_repeat] = true
```
joshuay03, line 7: I think this is the easiest way to configure `debug` without messing around with environment variables (ref: https://github.com/ruby/debug/issues/1098).

railties/lib/rails/generators/rails/app/templates/config/initializers/debug.rb.tt
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+if Rails.env.local?
+  require "debug"
```
eugeneius: This will eagerly activate the debugger on boot, which we don't want: https://github.com/rails/rails/pull/51692

joshuay03: Thanks! I've updated to require just `debug/config` which is more than enough to do what we need.

railties/lib/rails/generators/rails/plugin/templates/Gemfile.tt
```diff
@@ -14,8 +14,8 @@ gem "rubocop-rails-omakase", require: false
 <%- end -%>
 <% if RUBY_ENGINE == "ruby" -%>
 
-# Start debugger with binding.b [https://github.com/ruby/debug]
-# gem "debug", ">= 1.0.0"
+# See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
+gem "debug", platforms: %i[ mri windows ], require: "debug/prelude"
```
joshuay03, line 18: This is the only commented out gem in plugins, and [it isn't commented out in apps](https://github.com/rails/rails/blob/aa66aece44f191e6330c04ce4942c3edf31d85d7/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L55-L56), so I opted to un-comment it here rather than have a separate conditional initializer just for plugins that does something hacky like check if the gem is loaded first.

PR #55217, opened 2025-06-20
Respect users configured `IRB.conf[:IRB_NAME]`

jclusso:
### Details

If a user has configured `IRB_NAME` in their IRB configuration, Rails is overriding it. This PR makes it so the `IRB_NAME` is only set if it is not the default, similar to the way the `PROMPT_MODE` is set.

### Example

In my projects `.irbrc` I had the following. I had to dig into the Rails code to figure out why this no longer worked after upgrading to the latest version because of the override.

```ruby
project_name = colorize(File.basename(Dir.pwd), :blue)
if Rails.env.development?
  IRB.conf[:IRB_NAME] = project_name
else
  environment = case Rails.env
    when 'staging' then colorize('STAG', :yellow)
    when 'production' then colorize('PROD', :red)
    when 'test' then colorize('TEST', :yellow)
  end
  IRB.conf[:IRB_NAME] = "#{project_name}[#{environment}]"
end
IRB.conf[:PROMPT][:RAILS] = {
  PROMPT_I: '%N(%m):%n> ',
  PROMPT_S: '%N(%m):%n%l ',
  PROMPT_C: '%N(%m):%n* ',
  RETURN: "=> %s\n",
  AUTO_INDENT: true
}
IRB.conf[:PROMPT_MODE] = :RAILS
```

### Checklist

### DIFF
--- a/railties/lib/rails/commands/console/irb_console.rb
+++ b/railties/lib/rails/commands/console/irb_console.rb
@@ -87,7 +87,8 @@ module Rails
 
         env = colorized_env
         prompt_prefix = "%N(#{env})"
-        IRB.conf[:IRB_NAME] = @app.name
+        # Respect user's configured irb name.
+        IRB.conf[:IRB_NAME] = @app.name if IRB.conf[:IRB_NAME] == "irb"
 
         IRB.conf[:PROMPT][:RAILS_PROMPT] = {
           PROMPT_I: "#{prompt_prefix}:%03n> ",
--- a/railties/test/application/console_test.rb
+++ b/railties/test/application/console_test.rb
@@ -242,6 +242,21 @@ class FullStackConsoleTest < ActiveSupport::TestCase
     write_prompt "User.new.respond_to?(:age)", "=> true"
   end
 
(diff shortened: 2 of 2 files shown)


railties/lib/rails/commands/console/irb_console.rb
```diff
@@ -87,7 +87,8 @@ def start
 
         env = colorized_env
         prompt_prefix = "%N(#{env})"
-        IRB.conf[:IRB_NAME] = @app.name
+        # Respect user's configured irb name.
+        IRB.conf[:IRB_NAME] = @app.name if IRB.conf[:IRB_NAME] == "irb"
```
zzak, line 91: This seems likely to break if irb ever changes the default name.

I'm not sure we need to respect their setting here either, as the goal seems to be to put the Application name front and center to avoid confusion.

While this change could re-introduce that ambiguity.

jclusso: Seems unlikely that they'd change the default name.

That being said, this is as prone to issues as the PROMPT_MODE config option.

PR #51045, opened 2024-02-11
[Fix #51044] Deprecate providing a non-boolean argument to `#distinct`

joshuay03:
### Motivation / Background

Fixes #51044

### Detail

Updates the documented `ActiveRecord::QueryMethods#distinct` and undocumented `ActiveRecord::QueryMethods#distinct!` methods to ~raise an `ArgumentError`~ log a deprecation warning if given a non-boolean argument.

### Additional information

None.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55223, opened 2025-06-22
[Fix #51044] Deprecate boolean flag arguments in query methods

### LINKED ISSUE #51044: Relation `#distinct` doesn't raise an error for invalid arguments
### Steps to reproduce

[`#distinct`](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-distinct) only takes in a single `value` argument that defaults to `true`. This determines whether or not the distinct clause is applied to the relation. However, the argument isn't validated and is always treated as truthy or falsey (e.g. [in this condition](https://github.com/rails/rails/blob/0f9aaa5ca9b8421ebf42b3d7720632d4c1cae5fc/activerecord/lib/active_record/relation.rb#L474)). This means that a non-boolean argument can be incorrectly passed in with certain expectations (e.g. a column that's not the primary key expecting the query to result in a `DISTINCT ON`), but the `DISTINCT` clause will always only either apply or not apply on the base table's primary key. I've had to explain that passing (a truthy) argument this way doesn't change the outcome of the clause to junior engineers a couple of times, so I feel like this is something that should be validated. The script below is a good example of how one could misunderstand this API.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

  create_table :authors, force: true do |t|
  end

    t.integer :author_id
  end
end

  has_many :posts
end

  belongs_to :author
end

  def test_distinct_with_non_boolean_arguments
    author = Author.create!
    post_1 = Post.create!(author: author)
    post_2 = Post.create!(author: author)

    # Passes as expected
    assert_equal 1, Post.where(id: [post_1.id, post_2.id]).select("DISTINCT author_id").count

    # Fails with 'expected 1, got 2' which is correct, but it should raise an ArgumentError
    # since the `distinct` argument is invalid and the DISTINCT clause is on id and not author_id
    assert_equal 1, Post.where(id: [post_1.id, post_2.id]).distinct(:author_id).count
  end
end

```

### Expected behavior

An `ArgumentError` should be raised

### Actual behavior

No error is raised, argument is treated the same as `true` if truthy, and `false` if falsey

### System configuration
**Rails version**: 7.1.3 (latest), main

**Ruby version**: 3.3.0

joshuay03:
### Motivation / Background

Closes https://github.com/rails/rails/issues/51044
Supersedes https://github.com/rails/rails/pull/51045
Addresses https://github.com/rails/rails/pull/51045#issuecomment-2214613080

cc @fatkodima @nvasilevski

### Detail

Deprecates providing boolean and `nil` (where explicitly supported) arguments to the following relation query methods:
- `#lock`
- `#readonly`
- `#strict_loading`
- `#distinct`

By extension, the changes to `#lock` will also affect the following model methods:
- `#lock!`
- `#with_lock`

In the case of `true` or `nil`, all that is needed to address the deprecation is omit the argument. In the case of `false`, I've introduced new `#unscope_<query_method>` methods which users can migrate to instead.

### Additional information

Some or all of these methods may have unintentionally accepted truthy and falsy values as arguments, which creates ambiguity—one of the key motivations behind these changes. Considering none of these behaviours have been documented or tested against, neither the deprecation message nor the changelog entry refer to them.

I've only changed public APIs in these changes. All these methods call out to undocumented relation query methods (e.g. `#lock!`) that actually mutate the relation, and those in turn may be relying on similar arel methods, both of which may and will continue to handle truthy/falsy arguments.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1301,16 +1301,32 @@ module ActiveRecord
       self
     end
 
-    # Specifies locking settings (default to +true+). For more information
+    # Specifies locking settings (defaults to `"FOR UPDATE"`). For more information
     # on locking, please see ActiveRecord::Locking.
-    def lock(locks = true)
-      spawn.lock!(locks)
+    def lock(lock = (lock_not_given = "FOR UPDATE"))
+      unless lock_not_given || lock.is_a?(String)
+        ActiveRecord.deprecator.warn(<<~MSG)
+          Passing a non-`String` argument to `lock` is deprecated and will be removed in Rails 8.2.
+          Use `lock` without arguments, or with a custom lock statement to specify the lock type,
+          and `unscope_lock` to unspecify the lock.
+        MSG
+      end
+
+      spawn.lock!(lock)
     end
 
-    def lock!(locks = true) # :nodoc:
-      case locks
-      when String, TrueClass, NilClass
-        self.lock_value = locks || true
+    # Unspecifies locking settings. For more information on locking, please see
+    # ActiveRecord::Locking.
+    def unscope_lock
+      spawn.lock!(false)
+    end
+
+    def lock!(lock = "FOR UPDATE") # :nodoc:
+      case lock
+      when String
+        self.lock_value = lock
+      when TrueClass, NilClass
+        self.lock_value = "FOR UPDATE"
       else
         self.lock_value = false
       end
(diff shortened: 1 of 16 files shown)


activerecord/lib/active_record/relation/query_methods.rb
```diff
@@ -1233,16 +1233,32 @@ def offset!(value) # :nodoc:
       self
     end
 
-    # Specifies locking settings (default to +true+). For more information
+    # Specifies locking settings (defaults to `"FOR UPDATE"`). For more information
     # on locking, please see ActiveRecord::Locking.
-    def lock(locks = true)
-      spawn.lock!(locks)
+    def lock(lock = (lock_not_given = "FOR UPDATE"))
```
joshuay03, line 1306: Ref for default: https://github.com/rails/rails/blob/276b655f3606cf81b26464d135db9f24bacc1f0f/activerecord/lib/arel/select_manager.rb#L52-L63

activerecord/lib/arel/nodes/casted.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "active_model/attribute"
```
joshuay03, line 3: Necessary to run this suite on its own.

activerecord/test/cases/locking_test.rb
```diff
@@ -821,6 +821,20 @@ def test_no_locks_no_wait
...
+      assert_not_deprecated(ActiveRecord.deprecator) { person.lock!("FOR UPDATE") }
+      assert_not_deprecated(ActiveRecord.deprecator) { person.lock!(Arel.sql("FOR UPDATE")) }
+    end
+
+    def test_with_lock_with_true_argument_deprecation
+      person = Person.find 1
+      assert_deprecated(ActiveRecord.deprecator) { person.with_lock(true) { } }
+      assert_not_deprecated(ActiveRecord.deprecator) { person.with_lock("FOR UPDATE") { } }
+      assert_not_deprecated(ActiveRecord.deprecator) { person.with_lock(Arel.sql("FOR UPDATE")) { } }
+    end
```
joshuay03, line 896: There's no need to test for `false` or `nil` here unlike with the relation `#lock` since the method is only called with truthy values, which makes sense cause calling `record.lock!(false)` would be very unusual...

Ref: https://github.com/rails/rails/blob/276b655f3606cf81b26464d135db9f24bacc1f0f/activerecord/lib/active_record/persistence.rb#L838-L839

PR #55226, opened 2025-06-23
Add note to Migrations guide for migrations_paths option [ci skip]

zzak:
* ~Adds link to Configuring guide for more~
* Add note to Migrations guide for `migrations_paths` option
* ~Adds descriptions for undocumented public getters on HashConfig~

**Edit**: I backtracked on using the API docs to provide a reference of available config/database.yml flags

### DIFF
--- a/guides/source/active_record_migrations.md
+++ b/guides/source/active_record_migrations.md
@@ -112,6 +112,9 @@ another application or generating a file yourself, be aware of its position in
 the order. You can read more about how the timestamps are used in the [Rails
 Migration Version Control section](#rails-migration-version-control).
 
+NOTE: You can override the directory that migrations are stored in by setting the
+`migrations_paths` option in your `config/database.yml`.
+
 When generating a migration, Active Record automatically prepends the current
 timestamp to the file name of the migration. For example, running the command
 below will create an empty migration file whereby the filename is made up of a


guides/source/active_record_migrations.md
```diff
@@ -112,6 +112,9 @@ another application or generating a file yourself, be aware of its position in
 the order. You can read more about how the timestamps are used in the [Rails
 Migration Version Control section](#rails-migration-version-control).
 
+NOTE: You can override the directory migrations are stored by setting the
```
alexcameron89: This sentence was a little confusing the first time that I read it.
```suggestion
NOTE: You can override the directory that migrations are stored in by setting the
```

activerecord/lib/active_record/database_configurations/hash_config.rb
```diff
@@ -69,26 +71,33 @@ def _database=(database) # :nodoc:
...
       def query_cache
         configuration_hash[:query_cache]
       end
 
+      # The maximum number of connections that can be queued up waiting for a query.
       def max_queue
         max_threads * 4
       end
 
+      # The number of seconds to wait for a connection to be checked out from the pool.
```
alexcameron89:
```suggestion
      # The number of seconds to wait for a connection to be checked out from the pool before it is timed out.
```

alexcameron89 (commented):
I left a couple of suggestions, but overall I like these additions and I learned something new. 😄

zzak:
It is not _all of them_, so maybe misleading.

Many options only exist on the adapter, but is hard to differentiate from config methods, for example:
https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L211

Maybe there is a way to phrase it that includes some, but not limited to, API available in: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/AbstractAdapter.html

eileencodes:
Methods on hash config are the ones Rails internals need to access so we don't have to access the hash in the code (ie we can do `db_config.min_threads` instead of `db_config.configuration_hash[:min_threads]` . I can't think of cases where apps need to call into these methods. It's fine to document them since they aren't likely to change much, but I thought they were nodoc'd.

zzak:
I think my idea was to use this as a way to document "what config options are available", not the official "what methods can I call" use of API docs -- so this maybe misplaced, and the proper place to document is the guides.

The current home for some of them, is buried among a billion other flags: https://guides.rubyonrails.org/configuring.html#configuring-a-database

Anyway, I rebased those changes because it's not right and now just adding the info about `migrations_paths` to the Migrations guide here. :bow:

PR #55227, opened 2025-06-23
Fix 55215 test failure

fabricerenard12:
### Motivation / Background

Fixes #55215 

### Detail

This Pull Request adds `require "action_controller"` to [railties/lib/rails/health_controller.rb](https://github.com/rails/rails/pull/55227/files#diff-dfd63e49a9b67adb3b3ada9ac7e38ad7f29945b3f3db66eab4a663e98ac96ab1), guaranteeing ActionController::Base is loaded before Rails::HealthController is defined and eliminating the NameError seen in #55215.

### Checklist

### DIFF
--- a/railties/lib/rails/health_controller.rb
+++ b/railties/lib/rails/health_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "action_controller"
+
 module Rails
   # Built-in Health Check Endpoint
   #
--- a/railties/lib/rails/application_controller.rb
+++ b/railties/lib/rails/application_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "action_controller"
+
 class Rails::ApplicationController < ActionController::Base # :nodoc:
   prepend_view_path File.expand_path("templates", __dir__)
   layout "application"


skipkayhil:
👋 hey @fabricerenard12, thanks for taking a stab at this! It looks like there's a real failure in CI, could you take a look?

railties/lib/rails/health_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "action_controller/railtie"
```
Edouard-chin: I think the right fix is to require "action_controller` only so that all autoloads' are setup.
Requiring the railtie should be opt-in. In a normal Rails application this would not matter much, but in standalone scripts like Rails bug report templates (e.g. https://github.com/rails/rails/blob/05f0b1614d82424a7507caaec5cf701d3a8c2b21/guides/bug_report_templates/active_job.rb) we want to be able to choose which railtie to load.

Edouard-chin (commented):
We we'll also need to modify the `Rails::ApplicationController` with the same fix https://github.com/rails/rails/blob/05f0b1614d82424a7507caaec5cf701d3a8c2b21/railties/lib/rails/application_controller.rb

It's the parent class of the PwaController which gets also eagerloaded. (it works right now without issue but if we end up swapping these two lines in the future it would fail the same way) https://github.com/rails/rails/blob/05f0b1614d82424a7507caaec5cf701d3a8c2b21/railties/lib/rails.rb#L34-L37)

Thanks !

Edouard-chin (approved):
👍 ! Someone from the team with commit privilege will review and merge when they get the chance.

Thanks for your contribution !

PR #25289, opened 2016-06-05
HTML tag dynamic proxy - modern syntax for tag helpers, related to issue #25195

marekkirejczyk:
### Prototype for new tag helper syntax

This is prototype implementation for @dhh's proposal in issue #25195 for HTML tag dynamic proxy.
The purpose of this PR is to discuss and gather feedback rather then being merged.

Example usage:

```
# To produce <span class="bookmark"></span> before:
content_tag :span, nil, class: 'bookmark'
# After (because we will keep an index of tags that need closing, like <span>, vs those that do not, like <br>)
tag.span class: 'bookmark'

# To produce <div id="post_1"> before:
content_tag :div, post.title, id: dom_id(post)  
# After:
tag.div post.title, id: dom_id(post)

# To produce <br> before:
tag :br, nil, true
# After (becomes like span, we know this tag doesn't need closing)
tag.br

# Would be nice to support nesting too, so to produce <div id="header"><span>hello</span></div>, you'd:
tag.div(id: 'header') { |tag| tag.span 'hello' }
```

PR #55228, opened 2025-06-23
Modernize tag helper examples to use `tag.<tag name>` syntax [ci skip]

sato11:
### Motivation / Background
This Pull Request has been created because the documentation of Action View Tag Helper's `#tag` contains a set of examples but they are formatted using what the same documentation considers as legacy syntax[^1]

[^1]: https://api.rubyonrails.org/v8.0.2/classes/ActionView/Helpers/TagHelper.html#method-i-tag

### Detail
This Pull Request replaces legacy `tag("element")` examples with modern `tag.<tag name>` and highlights HTML5-compliant formatting output in which trailing slashes are unnecessary.

### Additional information
The new syntax was added at a65a3bde0be. It seems it was planned to deprecate the legacy syntax but apparently the deprecation has not been commenced. I am not sure if I would like to see the deprecation resumed after years, but separately I guess it would be great to show modern syntax for developers to learn a better practice from examples.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/tag_helper.rb
+++ b/actionview/lib/action_view/helpers/tag_helper.rb
@@ -453,29 +453,26 @@ module ActionView
       #
       # ==== Examples
       #
-      #   tag("br")
-      #   # => <br />
-      #
-      #   tag("br", nil, true)
+      #   tag.br
       #   # => <br>
       #
-      #   tag("input", type: 'text', disabled: true)
-      #   # => <input type="text" disabled="disabled" />
+      #   tag.input(type: "text", disabled: true)
+      #   # => <input type="text" disabled="disabled">
       #
-      #   tag("input", type: 'text', class: ["strong", "highlight"])
-      #   # => <input class="strong highlight" type="text" />
+      #   tag.input(type: "text", class: ["strong", "highlight"])
+      #   # => <input type="text" class="strong highlight">
       #
-      #   tag("img", src: "open & shut.png")
-      #   # => <img src="open &amp; shut.png" />
+      #   tag.img(src: "open & shut.png")
+      #   # => <img src="open &amp; shut.png">
       #
-      #   tag("img", { src: "open &amp; shut.png" }, false, false)
-      #   # => <img src="open &amp; shut.png" />
+      #   tag.img(src: "open &amp; shut.png", escape: false)
+      #   # => <img src="open &amp; shut.png">
       #
-      #   tag("div", data: { name: 'Stephen', city_state: %w(Chicago IL) })
-      #   # => <div data-name="Stephen" data-city-state="[&quot;Chicago&quot;,&quot;IL&quot;]" />
+      #   tag.div(data: { name: "Stephen", city_state: %w(Chicago IL) })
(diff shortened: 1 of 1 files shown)


Edouard-chin (commented):
Do we need to modify the documentation considering that the doc is already split into: "New syntax" and "Legacy Syntax"

https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/actionview/lib/action_view/helpers/tag_helper.rb#L337
https://github.com/rails/rails/blob/c5bb138e43390a191ddb7aa4e0f46e7af8563dcc/actionview/lib/action_view/helpers/tag_helper.rb#L432

Maybe "Legacy Syntax" documentation can be entirely removed once it's no longer supported, but for now it's still worth having.

sato11:
Thanks for the comment. I'm afraid I do not fully grasp what actions I might or should take, but let me assume that you are inclined to be against the change so that I can respond.

I think the point is that ordinary people including me primarily learn from examples and then gradually deepen the contextual understanding. In this particular case, I believed for a long time that the two syntax were simply interchangeable until I recently happened to read through the whole article and proved it wrong. To be honest I did not even know what XHTML was and if it's good or meaningless to keep its compatibility until I further research and knew that it had been retired as of 2018.

I wished I had learned better, and thought making the examples stick with the new syntax could have saved me from the confusion. I know it is not adding or removing anything in terms of coverage but giving a proper emphasis on one paradigm over the other makes the intention clearer, doesn't it?

That said, if this does not satisfy the requirement of a meaningful change, it's just okay as I think I have learned a lot already, so let me know. On the other hand, it would be my pleasure if I can feed back what I have been given from this 🙂

skipkayhil:
I believe what Edouard is trying to point out is that the `Examples` header is nested below the `Legacy Syntax` header, implying that the examples should use the legacy syntax instead of the new one. If you're interested in more example for the new syntax, I think that may have merit (although there are a lot of existing example for the new syntax already, just spread out more), but it should go above the `Legacy Syntax` header.

sato11:
Ah it makes sense to me, and with it explained the structure before the change does look more coherent. Thank you for the clarification! And since I'm not really interested with expanding the examples I'll be closing this later 🙏

sato11:
May I ask if you think it is a good idea to actually move on to deprecate the legacy format? It looks like it is not going to be an easy path but I wonder if some of you in the team might encourage or discourage it 🤔

skipkayhil:
Some [context](https://github.com/rails/rails/pull/25289#issuecomment-223897212)

> primarily because I don't see a compelling reason to hard deprecate the existing usage of tag and content_tag. There's so much code written with that already that wouldn't be materially better off by being rewritten with the proxy tag.

I don't think a deprecation would be worthwhile. There isn't much (if any) maintenance cost, and there are valid reasons to use the old version (if the tag name is dynamic, ex. `tag(h1_or_h2)`).

Edouard-chin:
> I'm afraid I do not fully grasp what actions I might or should take,

Sorry that my comment was not clear 😅 , @skipkayhil's clarification is what I meant (thanks!).
I also don't think a deprecation would be worthwhile.

sato11:
Okay, leaving them untouched makes sense to me as well. Thank you both for the teachings! 🙏

PR #55232, opened 2025-06-23
`has_secure_password`: fix password validation.

### LINKED ISSUE #55225: [`has_secure_password`] Password confirmation validation skipped if password is whitespace-only
Hi all,

this is actually a follow-up to #35444, which was opened many years ago but unfortunately got stale and then closed.

The password confirmation validation added by `has_secure_password` is skipped if the password string is a sequence of whitespace characters, regardless of the string provided as confirmation, which is ignored with no error messages. This happens both in the latest Rails (currently 8.0.2) and in `main`.

As far as I can tell Rails is committed to support whitespace-only passwords (#25373, #16412), and since they trigger actual changes to the password digest I think it would make sense to align the validation logic to this use case (IMHO it would just make a better default).

I've already taken a quick look at the code; I wasn't able to find any tests documenting the current behavior is intended. But I did find a possible fix, so I volunteer to open a PR should the core team approve the change.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
  gem "bcrypt", "~> 3.1.7"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.string :password_digest, null: false
  end
end

  has_secure_password
end

  test "create new user with whitespace password and wrong password confirmation" do
    user = User.new(password: " ", password_confirmation: "anything")

    assert_not user.valid?(:create), "user should not be valid"
  end

  test "update existing user with whitespace password and wrong password confirmation" do
    user = User.create!(password: "password", password_confirmation: "password")

    user.password = " "
    user.password_confirmation = "anything"

    assert_not user.valid?(:update), "user should not be valid"
  end
end
```

### Expected behavior

The validation should fail if the password and the password confirmation strings are different, no matter if the password is a sequence of whitespace-only characters.

### Actual behavior

If the password is a sequence of whitespace-only characters, the validation succeeds regardless of the contents of the confirmation string.

### System configuration

**Rails version**: 8.0.2, `main@1ac4003547`

**Ruby version**: 3.4.4

callmesangio:
Previously, the password confirmation validation added by `has_secure_password` was skipped if the password string was a sequence of whitespace characters, regardless of the string provided as confirmation, which was ignored with no error messages.

This change fixes that behavior, so that the confirmation validation runs consistently, independently of the password string content.

Fixes #55225

### Motivation / Background

This Pull Request has been created because it fixes a bug in `has_secure_password` validations.

### Detail

This Pull Request changes how the password confirmation validation added by default by `has_secure_password` works.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/secure_password.rb
+++ b/activemodel/lib/active_model/secure_password.rb
@@ -155,7 +155,7 @@ module ActiveModel
             end
           end
 
-          validates_confirmation_of attribute, allow_blank: true
+          validates_confirmation_of attribute, allow_nil: true
         end
 
         # Only generate tokens for records that are capable of doing so (Active Records, not vanilla Active Models)
--- a/activemodel/test/cases/secure_password_test.rb
+++ b/activemodel/test/cases/secure_password_test.rb
@@ -104,6 +104,14 @@ class SecurePasswordTest < ActiveModel::TestCase
     assert_equal ["doesn't match Password"], @user.errors[:password_confirmation]
   end
 
+  test "create a new user with validation, a spaces only password, and an incorrect password confirmation" do
+    @user.password = " "
+    @user.password_confirmation = "something else"
+    assert_not @user.valid?(:create), "user should be invalid"
+    assert_equal 1, @user.errors.count
+    assert_equal ["doesn't match Password"], @user.errors[:password_confirmation]
+  end
+
   test "resetting password to nil clears the password cache" do
     @user.password = "password"
     @user.password = nil
@@ -173,6 +181,14 @@ class SecurePasswordTest < ActiveModel::TestCase
 
   test "updating an existing user with validation and an incorrect password confirmation" do
     @existing_user.password = "password"
+    @existing_user.password_confirmation = "something else"
+    assert_not @existing_user.valid?(:update), "user should be invalid"
+    assert_equal 1, @existing_user.errors.count
+    assert_equal ["doesn't match Password"], @existing_user.errors[:password_confirmation]
+  end
+
+  test "updating an existing user with validation, a spaces only password, and an incorrect password confirmation" do
+    @existing_user.password = " "
     @existing_user.password_confirmation = "something else"
     assert_not @existing_user.valid?(:update), "user should be invalid"
     assert_equal 1, @existing_user.errors.count


activemodel/lib/active_model/secure_password.rb
```diff
@@ -155,7 +155,7 @@ def has_secure_password(attribute = :password, validations: true, reset_token: t
             end
           end
 
-          validates_confirmation_of attribute, allow_blank: true
+          validates_confirmation_of attribute, allow_nil: true
```
callmesangio, line 158: This change assumes that, given the current implementation of the writer method `password=`, the corresponding reader method `password` can only actually return `nil` or a string that is not `empty?` (whitespace-only or not). Therefore, by skipping the confirmation validation only when `password` is `nil`, we get to validate every non-empty string, which is what we want.

activemodel/test/cases/secure_password_test.rb
```diff
@@ -104,6 +104,14 @@ class SecurePasswordTest < ActiveModel::TestCase
     assert_equal ["doesn't match Password"], @user.errors[:password_confirmation]
   end
 
+  test "create a new user with validation, a spaces only password, and an incorrect password confirmation" do
+    @user.password = " " * 72
```
MatheusRich: any particular reason for 72 here?

simi: Isn't that BCrypt limit? https://www.ory.sh/docs/troubleshooting/bcrypt-secret-length

callmesangio: @MatheusRich No technical reason, just consistency: I followed what looked like a convention applied by other test methods of the class.

But yeah, as @simi says, 72 (bytes, not characters) is the limit beyond which bcrypt implementations usually truncate the plaintext.

MatheusRich: Gotcha. I was just wondering if it meant something relevant for this test. I wonder if it makes sense to just use a single space, as that seems to be enough for this test.

callmesangio: @MatheusRich I made the change to have a single space string, and updated the commit accordingly, see https://github.com/rails/rails/compare/8de663fde4e813aef4c7593a55c216818bd81a6c..20e28b900dad008784c36f08742b426c0e7b6ceb

MatheusRich:
I'm unsure if this is considered a bugfix, so IDK if we need a changelog entry for it. I'll let someone else chime in.

callmesangio:
Hello there 👋🏻 
Is there anything I can do to push this forward?
Thanks!

activemodel/lib/active_model/secure_password.rb
seanpdoyle: Is the condition of a "blank" value here related to the fact that a `<form>` element submitted with an empty `<input type="password">` will be transformed into an empty `""` by a controller's `params` helper?  When would an empty `<input>` result in a `nil` value?

callmesangio: To my knowledge, an empty form `<input>` should not result in a `nil` value, but nonetheless, `nil` as a `password` value (e.g. assigned by other means) is explicitly supported, and has the effect of setting `password_digest` to `nil` as well, see https://github.com/rails/rails/blob/main/activemodel/lib/active_model/secure_password.rb#L187. We want to skip confirmation in that case alone.
Plus, I tried to remove the `allow_` condition altogether, but that resulted in test failures.
Let me know if I answered your question!

byroot: Makes sense to me. If there was a confirmation field in the form, you'll get `""`, if the field is being changed from another context, it just won't be there at all (`nil`).

byroot:
> I'm unsure if this is considered a bugfix,

I'd say so. But it's easier to add the changelog during backport anyways.

byroot:
Backported to 8-0-stable and 7-2-stable.

PR #55235, opened 2025-06-23
Deprecated Associations

fxn:
## Overview

This patch adds the ability to deprecate associations:

```ruby
has_many :posts, deprecated: true
```

With that, Active Record will report any usage of the `posts` association.

## Use cases

The main use case for this feature is deleting an association. By deprecating first, you can identify and eliminate current usage before the actual removal.

Even if the test suite reports no usage, you can still monitor production in case there is mocked usage or lack of coverage for whatever reason.

The `:notify` mode is particularly flexible. You can use it to register usage in a structured manner, thanks to the objects in the payload, and can also be used in creative ways. Like, enabling the flag dynamically for all associations and introspecting usage in CI, which can help detect unused associations in your code base. Subscribers may have arbitrary logic, allow lists per team, bubble up things only once, create CI artifacts, etc.

## Configuration

The railtie defines a new configuration endpoint:

```
config.active_record.deprecated_associations_options
```

If present (it is optional), this has to be a hash with keys `:mode` and/or `:backtrace`.

The default mode is `:warn`, and by default backtraces are disabled.

As usual, that maps to `ActiveRecord.deprecated_associations_options` for usage outside Rails.

### Mode

In `:warn` mode, usage issues a warning that includes the application-level place where the access happened, if any. This is the default mode.

In `:raise` mode, usage raises an `ActiveRecord::DeprecatedAssociationError` with a similar message and a clean backtrace in the exception object.

In `:notify` mode, usage publishes a `deprecated_association.active_record` Active Support notification. The payload has the association reflection (`:reflection`), the application-level location (`:location`) where the access happened (a `Thread::Backtrace::Location` object, or `nil`), and a deprecation message (`:message`).

### Backtrace

If `:backtrace` is true, warnings include a clean backtrace in the message, and notifications have a `:backtrace` key in the payload with an array of clean `Thread::Backtrace::Location` objects.

Backtraces may help gathering context about some reported usage. For example, if the application-level spot reads

```
model.save(params)
```

and that is deep in the model layer and may be reached from 10 different places, perhaps you need help tracing the call up to see which one of them is setting `params` such that the association is used.

### Logger and Backtrace Cleaner

Warnings are issued with `ActiveRecord::Base.logger`, which is the same as `Rails.logger` by default.

Application-level frame/location and backtraces are computed with `ActiveRecord::LogSubscriber.backtrace_cleaner`, which is the same as `Rails.backtrace_cleaner` by default.

### Associations API

The entire associations API is guarded, think `Author#posts`, `Author#posts=`, `Author#post_ids`, etc.

### Nested Through Associations

By following throughs and sources, a through association may be chained arbitrarily deep.

Say a through association has other 7 nested ones, and let's assume 2 of them are deprecated. Every usage of the root through association reports those 2, since they are used, even if only to generate JOINs.

### Nested Attributes

If there is also

```ruby
accepts_nested_attributes_for :posts
```

Any assignment via `posts_attributes=` is reported.

### Queries

For example, in queries like

```ruby
Author.includes(:posts).where(...)
```

### Side Effects

An association may have side effects.

For example, if a deprecated `has_many` has `dependent: :nullify`, FKs become `NULL` whenever the owner is destroyed, and that would stop happening if you remove the association (you remove the association, not necessarily the database table or other associations pointing to that table). Because of that, a warning is issued if the owner is destroyed, and the context explains why (see next section).

## Deprecation Context

In order to help users understand the reported deprecated usage, each deprecation message has additional context.

Some examples:

* *The association Author#posts is deprecated, the method post_ids was invoked*
* *The association Appointment#patients is deprecated, referenced as nested association of the through Physician#patients*
* *The association Author#posts is deprecated, referenced in query to preload records*
* *The association Author#posts is deprecated, :dependent has a side effect here*
* ...

### Association Builders

* Association builders accept a new `:deprecated` option.
* They add guards to the generated APIs.
* Guards go via `association.reflection` to have the reflection validated.
* They may add callbacks to cover implicit usage via side effects.

### Association Reflections

* Association reflections respond to `deprecated?`.
* Through associations respond to `deprecated_nested_reflections`, which is a cached array of nested deprecated association reflections (possibly empty).

### Queries

By guarding in the precise places when preloading or joining, I _believe_ all is covered. Please let me know if something is overlooked.

### ActiveRecord::Associations::Deprecation

The core of the feature is centralized in the new private module `ActiveRecord::Associations::Deprecation`

### Testing

The `models/dats` directory has stripped-down versions of existing fixture models. I tried first to add new deprecated associations in the existing models, but became too complicated due to side effects. With these ad-hoc models we have much more control.

## Documentation Comes Next

Once feedback is addressed, I'd like to merge and followup with documentation in API and guides.

## Thanks

I wrote a simpler version of this feature for @gusto, we are collaborating together. This was initially a private monkey patch to Active Record that is helping them identify and refactor associations that cross certain boundaries.

After seeing its potential, @gusto decided to contribute this work back. This has been also done in the context of our collaboration.

Would like to thank @gusto for this.


fatkodima:
I haven't looked at the implementation, but won't something like this work?
```ruby
Rails.application.deprecators[:association_deprecator] = ActiveSupport::Deprecation.new(...)

has_many :posts
deprecate :posts, deprecator: Rails.application.deprecators[:association_deprecator]
```

But it will not work for indirect accessing of the association, like `joins(:association)` etc.

fxn:
@fatkodima using a deprecator is something that was considered, not with that "deprecate method" approach, but somehow.

But I was not fully convinced, because deprecators seem to be designed nowadays for libraries to deprecate _their_ stuff, these are application-level entities, no gem is involved, there is no versioning, etc.

But open to tweak this aspect of the feature if a different way to report usage may be better.

fatkodima:
We started to use deprecators in our app. But do not have enough information yet to share how it's going.

fxn:
@fatkodima in that regard, if you'd like to read the patch, `ActiveRecord::Associations::Deprecation.report` is the most relevant method. The rest converge to that one.

So, you'll see the meat of the feature is the custom logic related to associations and their semantics or usage patterns. This is not just deprecating a method or a constant, you know, it works at a different level.

Then, `report` needs a logger and a backtrace cleaner, besides the ability to send AS notifications, which we can take for granted.

Raising would not make a lot of sense, because a single spot may trigger multiple notifications and this is designed so you get individual information about every offense. This happens with through associations and with queries that refer a tree of associations.

Could `report` be abstracted away using your application-level deprecations? How do they look like?

fxn:
> Raising would not make a lot of sense

After sleeping on this, I have thought that there might be use cases for raising. Perhaps you want to warn in `development` and `test` modes, unless you are in CI, where you don't want to allow any deprecated usage.

I have added support for `:raise`.

fxn:
Added a `:backtrace` flag which adds clean backtraces (either text, or location objects). Details in the rewritten PR description.

fxn:
I have iterated this one a lot, going to open a clean PR.

PR #55240, opened 2025-06-25
Make sure that `CurrentAttributes.set` works on the current instance

davidstosik:
### Motivation / Background

If the block passed to `CurrentAttributes.set` happens to clear all current attributes (which happens for example when performing jobs), then `.set` is unable to restore attribute values on the proper singleton instance of the `CurrentAttributes` class.
See the added test for details.

The problem has always existed, but was emphasized by PR #55139, causing tests that perform enqueued jobs in a `Current.set` block to reset current attributes to their defaults when exiting the block, instead of the values that were set previously in the test.

#### Before
`CurrentAttributes.set` delegates to the singleton instance, eventually calling `Object#with` on the current instance.
This can be a problem when the block clears the current instances.

#### After
`CurrentAttributes.set` does not delegate. Instead, it calls `Object#with` directly, ensuring it always acts on the right instance.

### Additional information

I also wrote a single-file executable test case to check behavior on different versions of Rails and to illustrate a closer-to-real-life scenario using jobs.
Using jobs, this actually demonstrates a regression between Rails 8.0.2 and HEAD.

<details><summary>You'll find the code here.</summary>

```rb
#!/usr/bin/env ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  if ENV["RAILS_VERSION"]
    gem "rails", ENV["RAILS_VERSION"]
  else
  end
end

require "active_job/railtie"
require "active_support/railtie"
require "minitest/autorun"

puts "🎯 Rails version: #{Rails::VERSION::STRING}"

  config.eager_load = false

end
Rails.application.initialize!

  if Rails::VERSION::STRING.to_f >= 7.2
    attribute :string_value, default: "default value set in Current"
  else
    attribute :string_value
  end
end

class MyJob < ActiveJob::Base
  def perform
    Current.string_value = "value set in MyJob#perform"
  end
end

class MyJobTest < ActiveJob::TestCase
  def test_1
    Current.string_value = "value set in test_1" 

    Current.set(string_value: "temporary value for the block") do
      perform_enqueued_jobs do
	MyJob.perform_later
      end
    end

    assert_equal(
      "value set in test_1",
      Current.string_value,
      "Coming out of `ActiveSupport::CurrentAttributes.set`, the attributes should be restored to the values they were before the block"
    )
  end
end
```

</details> 

<details><summary>And how to run it, and its output, here.</summary>

```sh-session
$ RAILS_VERSION=8.0.2 ruby test_current_attributes_job_set_clear.rb
...
🎯 Rails version: 8.0.2
...
1 runs, 2 assertions, 0 failures, 0 errors, 0 skips

$ ruby test_current_attributes_job_set_clear.rb
...
│🎯 Rails version: 8.1.0.alpha
...
  1) Failure:
MyJobTest#test_1 [test_current_attributes_job_set_clear.rb:56]:
Coming out of `ActiveSupport::CurrentAttributes.set`, the attributes should be restored to the values they were before the block.
--- expected
+++ actual
@@ -1 +1 @@
-"value set in test_1"
+"default value set in Current"


```

</details> 

### Checklist

* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. **Minor bug fixes** and documentation changes should not be included.

### DIFF
--- a/activesupport/lib/active_support/current_attributes.rb
+++ b/activesupport/lib/active_support/current_attributes.rb
@@ -151,7 +151,21 @@ module ActiveSupport
       end
       alias_method :after_reset, :resets
 
-      delegate :set, :reset, to: :instance
+      delegate :reset, to: :instance
+
+      # Expose one or more attributes within a block. Old values are returned after the block concludes.
+      # Example demonstrating the common use of needing to set Current attributes outside the request-cycle:
+      #
+      #   class Chat::PublicationJob < ApplicationJob
+      #     def perform(attributes, room_number, creator)
+      #       Current.set(person: creator) do
+      #         Chat::Publisher.publish(attributes: attributes, room_number: room_number)
+      #       end
+      #     end
+      #   end
+      def set(attributes, &block)
+        with(**attributes, &block)
+      end
 
       def clear_all # :nodoc:
         current_instances.each_value(&:reset)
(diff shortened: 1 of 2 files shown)


jenshenny:
The CI failures look unrelated

byroot:
I see how that fixes your very specific issue, but as you noticed this really isn't a clean fix for the root cause, and the behavior you suggest is even weirder. 

I think we really need to ding into nested executors and CurrentAttributes interaction. This has been known to be not working great for a long time, and I meant to figure something for the past 2 or 3 years but never got around to do it.

Nested executors aren't supposed to happen in production environment, but are fairly common in test. We need some proper `push / pop` semantic.

byroot:
Alright, I tried to address the root cause in https://github.com/rails/rails/pull/55247, could you give it a try?

jenshenny:
@byroot I tried https://github.com/rails/rails/pull/55247 and it fixes the issue 👍

PR #43734, opened 2021-11-26
Wrap `ActionController::TestCase` with Rails executor

ghiculescu:
Currently when you run an integration test (`ActionDispatch::IntegrationTest`), current attributes etc are reset at the end of each request, even if a test has multiple requests in it. This was refactored in https://github.com/rails/rails/pull/43550 but the functionality has been there for a long time.

The same is *not* true if you use `ActionController::TestCase`. Since https://github.com/rails/rails/pull/43550 attributes are reset at the end of a **test**, but if a test makes more than one request, current attributes will persist between requests.

This PR calls `Executor#wrap` around each request in a test if `config.active_support.executor_around_test_case` is enabled. It is by default in Rails 7, but not in older Rails.

PR #43550, opened 2021-10-27
Call Executor#wrap around each test

casperisfine:
Fix: https://github.com/rails/rails/pull/43546

We do want to run the executor callbacks because they do cleanup various global state at the end of an unit of work (`CurrentAttributes`, `QueryLogs`, `QueryCache`, etc).

However `Executor#wrap` has a safety to prevent being called recursively.

So we can't use it, otherwise all the integration test helpers such as `get`, `post` etc, won't properly cleanup after themselves.

So we introduce an alternative `perform` method, which is just like `wrap`, but don't prevent re-entrency.

cc @manuelpuyol @camertron, I tested this branch against https://github.com/github/view_component, I believe it fixes the bug you reported.

PR #55247, opened 2025-06-26
Allow for nested ExecutionContext in test

### LINKED ISSUE #49227: `CurrentAttribute`s are cleared when a job gets executed inline in tests
### Steps to reproduce

I have a test case like this

```rb
require "test_helper"

  attribute :user
end

class DummyJob < ActiveJob::Base
  def perform
  end
end

class CurrentAttributeTest < ActiveSupport::TestCase
  test "CurrentAttributes" do
    Current.user = "test"

    assert_equal("test", Current.user, "FIRST")
    DummyJob.perform_later
    assert_equal("test", Current.user, "SECOND") # 💣
  end
end
```

Using the inline queue adapter the second assertion failed. I tried making a self-contained test script but couldn't manage, the attributes only seem to be reset when I run it in a full-blown rails app. I have created a sample app here https://github.com/Earlopain/rails-current-attributes, you can just run rake test there.

I found issue #37526 and pr #37568 which seem to be about the same thing and it looks like they were closed as fixed/completed. Don't know what to make of that, just wanted to document.

### Expected behavior
CurrentAttributes are preserved.

### Actual behavior
CurrentAttributes are cleared.

### System configuration
**Rails version**: main

**Ruby version**: 3.2.2

byroot:
`ActiveSupport::CurrentAttribute` lifetime in test always have been a bit wonky.

This was addressed in the past by https://github.com/rails/rails/pull/43734 and then https://github.com/rails/rails/pull/43550 but a problem remained that various states would leak from the test case into the controller or job being tested, and back.

Ideally when executing a controller or job from a test case, we want to "save and restore" current attributes and the execution context.

This is what this commit does, by implementing `push` and `pop` logic for `ActiveSupport::ExecutionContext`.

For extra safety, since such use case isn't supposed to ever happen in production, this behavior is test only and in production we keep dropping all the state.

Fix: https://github.com/rails/rails/pull/55240
Fix: https://github.com/rails/rails/issues/49227

### DIFF
--- a/activestorage/test/test_helper.rb
+++ b/activestorage/test/test_helper.rb
@@ -12,6 +12,7 @@ require "active_support/core_ext/object/try"
 require "active_support/testing/autorun"
 require "image_processing/mini_magick"
 
+require "active_support/current_attributes/test_helper"
 require "active_record/testing/query_assertions"
 
 require "active_job"
@@ -25,6 +26,7 @@ ActiveStorage::FixtureSet.file_fixture_path = File.expand_path("fixtures/files",
 class ActiveSupport::TestCase
   self.file_fixture_path = ActiveStorage::FixtureSet.file_fixture_path
 
+  include ActiveSupport::CurrentAttributes::TestHelper
   include ActiveRecord::TestFixtures
   include ActiveRecord::Assertions::QueryAssertions
 
@@ -32,10 +34,6 @@ class ActiveSupport::TestCase
 
   setup do
     ActiveStorage::Current.url_options = { protocol: "https://", host: "example.com", port: nil }
-  end
-
-  teardown do
-    ActiveStorage::Current.reset
   end
 
   private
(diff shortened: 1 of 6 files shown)


Earlopain:
❤️ This seems relevant to https://github.com/rails/rails/issues/49227 and indeed when pointing to this branch the test case there passes. Thanks for working on this.

byroot:
Ah thanks, I knew we had an issue open for that, but couldn't find it.

such:
Hello,
It seems this commit broke my setup.
I'm running my test suite in spring locally and whenever I change a file and re-run a test, `ActiveSupport::ExecutionContext.to_h` starts returning `nil` which breaks a lot of things.

This is because `ActiveSupport::ExecutionContext.pop` is now called in `app.executor.to_complete` and the stack is empty.

Any chance you can help me figure out what the best fix is?

byroot:
It would be helpful if you could provide a reproduction (e.g. a fresh new app in a test repository).

alpaca-tc:
@such @byroot Fixed an issue where `ActiveSupport::ExecutionContext.to_h` would sometimes return `nil` 😄 
https://github.com/rails/rails/pull/56889

PR #55248, opened 2025-06-26
Set `ActiveStorage::Current.url_options[:script_name]` properly before Active Storage controller actions

flavorjones:
### Motivation / Background

When setting `Rack::Request#script_name` in middleware (for example, in a multi-tenant application), the Rails Active Storage Disk Service does not generate correct URLs.

### Detail

This pull request sets `ActiveStorage::Current.url_options[:script_name]` to the value of the request's script_name, so that Active Storage controllers will generate correct URLs.

### Additional information

I've also backfilled tests for the exissting ActiveStorage::SetCurrent behavior (which did not have test coverage other than indirect `:host` coverage).

### Checklist

### DIFF
--- a/activestorage/app/controllers/concerns/active_storage/set_current.rb
+++ b/activestorage/app/controllers/concerns/active_storage/set_current.rb
@@ -9,7 +9,12 @@ module ActiveStorage::SetCurrent
 
   included do
     before_action do
-      ActiveStorage::Current.url_options = { protocol: request.protocol, host: request.host, port: request.port }
+      ActiveStorage::Current.url_options = {
+        protocol: request.protocol,
+        host: request.host,
+        port: request.port,
+        script_name: request.script_name
+      }
     end
   end
 end
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   `ActiveStorage::Current.url_options[:script_name]` is set before Active Storage controller actions.
+
+    *Mike Dalessio*
+
 *   Direct upload progress accounts for server processing time.
 
     *Jeremy Daer*
--- a/activestorage/test/controllers/base_controller_test.rb
+++ b/activestorage/test/controllers/base_controller_test.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ActiveStorage::BaseControllerTest
+  class PlainController < ActiveStorage::BaseController
+    def index
+      self.response_body = ActiveStorage::Current.url_options.to_yaml
+    end
+  end
+
+  class TestUrlOptions < ActionDispatch::IntegrationTest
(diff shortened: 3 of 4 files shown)


activestorage/app/controllers/concerns/active_storage/set_current.rb
```diff
@@ -9,7 +9,12 @@ module ActiveStorage::SetCurrent
 
   included do
     before_action do
-      ActiveStorage::Current.url_options = { protocol: request.protocol, host: request.host, port: request.port }
```
Edouard-chin, line 12: I'm most likely missing something but do you maybe know why we need this rather than relying on the `UrlFor` module https://github.com/rails/rails/blob/99f918b15de469b8648f30231e0788c06598c294/actionpack/lib/action_controller/metal/url_for.rb#L37 that does exactly the same. I tried digging into the git history but couldn't see the reason.

Since we are in the context of a controller we should be able to do `ActiveStorage::Current.url_options = url_options` which would also solve the script name issue as it takes care of it as well.

flavorjones: @Edouard-chin This is a great question, thanks. It certainly _looks_ like an opportunity to simplify things, so let me make that change.

flavorjones: Hmm. What's interesting is that, in the Active Storage controllers, `UrlFor.url_options` may _not_ include `script_name` in the returned hash.

In my case, I have middleware setting `engine_script_name` as well as `script_name`, but because `engine_script_name` is set on the main application RouteSet, any other engine (like Active Storage) is not going to use it. So I would have to patch `UrlFor` with something like this for it to work:

```diff
--- a/actionpack/lib/action_controller/metal/url_for.rb
+++ b/actionpack/lib/action_controller/metal/url_for.rb
@@ -39,6 +39,7 @@ def url_options
         host: request.host,
         port: request.optional_port,
         protocol: request.protocol,
+        script_name: request.script_name.presence,
         _recall: request.path_parameters
       }.merge!(super).freeze
```

Which made me realize that this PR is being naïve in assuming that a request's script name should take effect in every engine. In my app, I want Active Storage to use it, but it's overly presumptuous to assume that every app setting script name will want this behavior for every engine; it's probably even presumptuous to assume every app will want it just for Active Storage.

TL;DR I think this PR is not right, and I'm going to close it. I should be able to inject the behavior I want into Active Storage controllers with a `before_action`, which is good enough unless/until the right shearing layer becomes apparent.

Edouard-chin: Thanks Mike for digging and sorry I made you go through the rabbit hole 😅 

> [...] which is good enough unless/until the right shearing layer becomes

That make sense to me as well. I'd be happy to dig myself when I have a bit more time on the reason `url_options` may not return `script_name` (I was assuming it would if passed as a rack environment variable).

PR #53067, opened 2024-09-27
Include the immutable directive in the cache-control header

guillaumebriday:
### Motivation / Background

As explained in the Mozilla documentation:

> When a user reloads the browser, the browser will send conditional requests for validating to the origin server. But it's not necessary to revalidate those kinds of static resources even when a user reloads the browser, because they're never modified. immutable tells a cache that the response is immutable while it's [fresh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#fresh_and_stale_based_on_age) and avoids those kinds of unnecessary conditional requests to the server.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
and https://www.keycdn.com/blog/cache-control-immutable

I believe it's an effective and straightforward approach to improve performance. 

For instance, GitHub is already using it in production.

Following @dhh's PR https://github.com/rails/rails/pull/52883

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55249, opened 2025-06-26
[Fix #53683] Reduce cache time for non-asset files in public dir

### LINKED ISSUE #53683: robot.txt, favicons cached for 1 year by default in production
**Summary**: Non-digested assets (including those _outside_ `/public/assets`) default to the same 1 year cache period in cache-control as digested assets. Since non-digested assets are mutable, a 1 year cache period is inappropriate.

### Background

The built-in public file server's current default cache-control specifies a 1 year cache period in `production.rb`. While this is fine for digested assets, it also applies to non-digested assets such as robots.txt, favicons, and anything else in `/public`, even if not in `/public/assets`.

https://github.com/rails/rails/blob/a8709e6ea26eca73a652af4fdd0a9f7db5352af4/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L20-L21

The comment notes that "they are all digest stamped", but indeed this is not the case.

To make Rails friendly to folks who are not well versed in HTTP caching semantics, I believe the default period in production.rb should be radically shorter. I would suggest 1 day maximum, and perhaps just 6 to 12 hours. This would be substantially more forgiving to mistakes than being stuck with something cached in a remote system for days or weeks.

Favicons in particular can be cached for extended periods. Crawlers are unlikely to honor a cached robots.txt for a full year, but are likely to honor the cache period for several days into weeks.

Changing this would affect digested assets too, but I suggest that there's little practical difference in overhead for the server between 1 day and 1 year. I'd suggest dropping the default period in development and test as well, to match, but those aren't particularly critical in practice.

Would there be any openness to such a change? I'm happy to create a PR if so (or someone else can--either way).

### Steps to reproduce

In production: 
```bash
curl -I http://some-production-app.com/robots.txt
```

In development:
```bash
# Enable caching
rails dev:cache
curl -I http://localhost:3000/icon.svg
```
Note that the default in development is 2 days, so that's what will be returned. 

### Expected behavior

The default cache period for non-digested static files should be forgiving to changes and mistakes.

I'd suggest a default cache period of something between 6 and 24 hours.

### Actual behavior

Max age is 1 year for a non-digested URL. Note that both `robots.txt` and `icon.svg` are default, stock files that are generated by a newly generated Rails app.

Production:
```bash
$ curl -i http://some-production-app.com/robots.txt
HTTP/1.1 200 OK
Cache-Control: public, max-age=31556952
Content-Length: 99
Content-Type: text/plain
Last-Modified: Tue, 19 Nov 2024 21:53:41 GMT
Date: Wed, 20 Nov 2024 18:08:41 GMT
```

Development (with `rails dev:cache`):
```bash
$ curl -I http://localhost:3000/icon.svg
HTTP/1.1 200 OK
last-modified: Wed, 20 Nov 2024 01:56:36 GMT
content-type: image/svg+xml
cache-control: public, max-age=172800
Content-Length: 122
```

### System configuration
**Rails version**: 8.0.0

**Ruby version**: 3.3.6

thisismydesign:
### Motivation / Background

This Pull Request is a fix for https://github.com/rails/rails/issues/53683

### Detail

This Pull Request changes the default caching to only cache digest stamped assets for a 1 year, other public files for 1 hour only.

This is a bad default and should be fixed asap. If this is not the way, consider removing the cache altogether.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -17,8 +17,17 @@ Rails.application.configure do
   config.action_controller.perform_caching = true
   <%- end -%>
 
-  # Cache assets for far-future expiry since they are all digest stamped.
-  config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
+  # Cache digest stamped assets for far-future expiry.
+  # Short cache for others: robots.txt, sitemap.xml, 404.html, etc.
+  config.public_file_server.headers = {
+    "cache-control" => lambda do |path, _|
+      if path.start_with?("/assets/")
+        "public, max-age=#{1.year.to_i}"
+      else
+        "public, max-age=#{1.minute.to_i}"
+      end
+    end
+  }
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.asset_host = "http://assets.example.com"
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -22,9 +22,12 @@ Rails.application.configure do
   config.public_file_server.headers = {
     "cache-control" => lambda do |path, _|
       if path.start_with?("/assets/")
-        "public, max-age=#{1.year.to_i}"
+        # Files in /assets/ are expected to be fully immutable.
+        # If the content change the URL too.
+        "public, immutable, max-age=#{1.year.to_i}"
       else
-        "public, max-age=#{1.minute.to_i}"
+        # For anything else we cache for 1 minute.
+        "public, max-age=#{1.minute.to_i}, stale-while-revalidate=#{5.minutes.to_i}"
       end
     end
   }


railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -17,8 +17,17 @@ Rails.application.configure do
   config.action_controller.perform_caching = true
   <%- end -%>
 
-  # Cache assets for far-future expiry since they are all digest stamped.
-  config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
+  # Cache digest stamped assets for far-future expiry.
+  # Short cache for others: robots.txt, sitemap.xml, 404.html, etc.
+  config.public_file_server.headers = {
+    'cache-control' => lambda do |path, _|
+      if path.match?(/\A\/assets\/.*-[a-f0-9]{8,}[a-f0-9.]*$/)
```
byroot:
```suggestion
      if path.start_with?("/assets/")
```

I think just checking the directory is enough. If you put non-immutable assets in `/assets/` that's on you.

railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -17,8 +17,17 @@ Rails.application.configure do
...
+        "public, max-age=#{1.year.to_i}"
+      else
+        "public, max-age=#{1.hour.to_i}"
```
byroot:
```suggestion
        "public, max-age=#{1.hour.to_i}"
```

You can go even lower. 1 or 5 minutes just to avoid thundering herds should be plenty.

thisismydesign:
Thanks @byroot, will update the pr

railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
thisismydesign: What about `public/assets/.manifest.json`? Is there a downside to sticking with a more precise match? Seems like the safer option.

byroot: What's the use case for downloading it?

byroot: > Is there a downside to sticking with a more precise match?

Yes, the digest implementation may change, and the regexp stop matching.

thisismydesign: You're right, updated.

railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -17,8 +17,20 @@ Rails.application.configure do
...
+  # Short cache for others: robots.txt, sitemap.xml, 404.html, etc.
+  config.public_file_server.headers = {
+    'cache-control' => lambda do |path, _|
+      if path.start_with?("/assets/")
+        # Files in /assets/ are expected to be fully immutable.
+        # If the content change the URL too.
+        "public, immutable, max-age=#{1.year.to_i}"
+      else
+        # For anything else we cache for 1 minute.
+        "public, max-age=#{1.minute.to_i}, stale-while-revalidate=#{5.minutes.to_i}"
```
byroot, line 30: I improved the defaults with `immutable` and `stale-while-revalidate`.

zzak:
There is also #53067, just FYI.

jdelStrother:
If it's not a dumb question... this change does work for you, right?
I'm not aware of being able to set Rack headers to procs, and when I try it I get response headers looking like this - 

```
< HTTP/1.1 200 OK
< last-modified: Wed, 18 Aug 2021 12:07:09 GMT
< content-type: text/plain
< cache-control: #<Proc:0x0000000151174718 /Users/jon/Developer/web/config/environments/development.rb:36 (lambda)>
< Content-Length: 14
```

if it works for you I'll dig into it some more, maybe I've got something weird going on locally.

byroot:
@jdelStrother please open a new issue. I suspect it is an issue on your side, but if it isn’t it’s a pretty bad bug so this shouldn’t be overlooked.

jdelStrother:
@byroot ok, done - https://github.com/rails/rails/issues/55617

byroot:
This callable interface indeed doesn't exist (my failure for not catching it) and isn't easy to add, at least in time for the 8.1 release.

Hence I will revert this for now.

PR #55250, opened 2025-06-26
`action_cable_meta_tag` uses `Rack::Request#script_name` to build the mount path

flavorjones:
### Motivation / Background

When setting `Rack::Request#script_name` in middleware (for example, in a multi-tenant application), the Action Cable url set in the meta tag is incorrect.

### Detail

This PR adds a name to the Action Cable server route, so that we can use ordinary URL helpers to generate the path correctly. The `action_cable_meta_tag` helper now uses that URL helper.

### Additional information

I couldn't find any existing direct test coverage for the contents of this meta tag, so I have not added test coverage for this new behavior.

### Checklist

### DIFF
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   `action_cable_meta_tag` uses `Rack::Request#script_name` to generate the Action Cable mount path.
+
+    The Action Cable server route, if it is configured via `config.action_cable.mount_path`, now has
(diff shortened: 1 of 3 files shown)


skipkayhil (approved):

flavorjones:
I'm going to close this; after digging in a bit more on related PR https://github.com/rails/rails/pull/55248 I've realized that not every app is going to want `script_name` to take effect for Action Cable.

I can add this behavior into my app with a custom meta tag helper, so I'll just go ahead and do that instead.

PR #33756, opened 2018-08-30
Add `inspection_masks` to make values of sensitive database columns w…

piecehealth:
…on't be exposed while call `#inspect`.

* Why:
Some sensitive data will be exposed in log accidentally, e.g.

```ruby
account = Account.find params[:id]
payload = { account: @account }
logger.info "payload will be #{ payload }"
```

All the information of `account` will be exposed in log.

* Solution:
Add a class attribute `inspection_masks` to specify which values of columns shouldn't be exposed.
```
class Account < ActiveRecord::Base
  self.inspection_masks = [:credit_card_number]
end

Account.last.insepct # => #<Account id: 123, credit_card_number: [FILTERED] ...>
```

### Summary

Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

If you are updating any of the CHANGELOG files or are asked to update the
CHANGELOG files by reviewers, please add the CHANGELOG entry at the top of the file.

Finally, if your pull request affects documentation or any non-code
changes, guidelines for those changes are [available
here](http://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#contributing-to-the-rails-documentation)

Thanks for contributing to Rails!

PR #55251, opened 2025-06-26
filter sensitive attributes in logs

that-jill:
### Motivation / Background

We have two ways to configure things that get filtered in logs: `filter_parameters` and `filter_attributes`.  The [latter was added in 2018](https://github.com/rails/rails/pull/33756) and ensures that things added to `filter_parameters` are also added to `filter attributes`. However, the opposite is not true, which makes for some weird behavior right out of the box: 

Take this example with a model called `CreditCard` with two attributes: `digits` and `cvv`. 

`cvv` is added to `filter_parameters` by default:

Screenshot 2025-06-26 at 5 20 03 PM

`digits` is added to `filter_attributes`: 

Screenshot 2025-06-26 at 5 21 05 PM

Which results in logs that look like this:

Screenshot 2025-07-03 at 11 19 03 AM

cvv gets filtered in both places because `filter_parameters` also adds it to `filter_attributes`. However, the opposite is not true, despite the fact that it's typical for parameters that are passed into a controller to match attributes on the model. Let's make sure we can define what needs filtering in one place and ensure it's filtered in every place.

### Detail

This PR aims to ensure that if you've added something to `filter_attributes`, it is filtered by default in the logs. 

### Additional information

I looked at several places to put this code. I don't love where it is now, but we do have a circular loading problem because the AR classes get loaded well after filter_parameters are precompiled. 

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Attributes filtered by `filter_attributes` will now also be filtered by `filter_parameters`
+    so sensitive information is not leaked.
+
+    *Jill Klang*
+
 *   Add `connection.current_transaction.isolation` API to check current transaction's isolation level.
 
     Returns the isolation level if it was explicitly set via the `isolation:` parameter
--- a/activerecord/lib/active_record/filter_attribute_handler.rb
+++ b/activerecord/lib/active_record/filter_attribute_handler.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  class FilterAttributeHandler # :nodoc:
+    class << self
+      def on_sensitive_attribute_declared(&block)
+        @sensitive_attribute_declaration_listeners ||= Concurrent::Array.new
+        @sensitive_attribute_declaration_listeners << block
+      end
+
+      def sensitive_attribute_was_declared(klass, list)
+        @sensitive_attribute_declaration_listeners&.each do |block|
+          block.call(klass, list)
+        end
+      end
+    end
+
+    def initialize(app)
+      @app = app
+      @attributes_by_class = Concurrent::Map.new
+      @collecting = true
+    end
+
+    def enable
+      install_collecting_hook
+
+      apply_collected_attributes
+      @collecting = false
+    end
+
+    private
+      attr_reader :app
+
+      def install_collecting_hook
+        self.class.on_sensitive_attribute_declared do |klass, list|
+          attribute_was_declared(klass, list)
+        end
+      end
+
+      def attribute_was_declared(klass, list)
+        if collecting?
+          collect_for_later(klass, list)
+        else
+          apply_filter(klass, list)
+        end
+      end
+
+      def apply_collected_attributes
+        @attributes_by_class.each do |klass, list|
+          apply_filter(klass, list)
+        end
+      end
+
+      def collecting?
+        @collecting
+      end
+
+      def collect_for_later(klass, list)
+        @attributes_by_class[klass] ||= Concurrent::Array.new
+        @attributes_by_class[klass] += list
+      end
+
+      def apply_filter(klass, list)
+        list.each do |attribute|
+          next if klass.abstract_class?
+
+          klass_name = klass.name ? klass.model_name.element : nil
+          filter = [klass_name, attribute.to_s].compact.join(".")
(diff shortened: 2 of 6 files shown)


that-jill:
Will work on the tests and see if I can find a cleaner way to do this -- maybe something like https://github.com/rails/rails/blob/main/activerecord/lib/active_record/encryption/auto_filtered_parameters.rb ? Thanks to @skipkayhil for pointing that out

activerecord/lib/active_record/filter_attribute_handler.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  class FilterAttributeHandler # :nodoc:
+    class << self
+      def on_sensitive_attribute_declared(&block)
+        @encrypted_attribute_declaration_listeners ||= Concurrent::Array.new
```
eileencodes: This isn't dependent on the encrypted attributes right? So I think these should be renamed to something like `sensitive_attribute_declaration_listeners`.

activerecord/lib/active_record/filter_attribute_handler.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+      end
+    end
+
+    def initialize(app)
+      @app = app
+      @attributes_by_class = Concurrent::Map.new
+      @collecting = true
+
+      install_collecting_hook
```
eileencodes: I think `install_collecting_hook` should be in `#enable`, but that's a person preference and non-blocking. it's a littl cleaner to have the initializer do inits and the `enable` call do the work.

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+*   Ensure anything added to `filter_attributes` also gets added to `filter_parameters`
+    so sensitive information is not leaked.
```
eileencodes:
```suggestion
*   Attributes filtered by `filter_attributes` will now also be filtered by `filter_parameters`
    so sensitive information is not leaked.
```

eileencodes (commented):
This is looking great, just a few comments on minor changes and we'll be good to merge.

activerecord/lib/active_record/filter_attribute_handler.rb
```diff
@@ -0,0 +1,71 @@
...
+      def collect_for_later(klass, list)
+        @attributes_by_class[klass] ||= Concurrent::Array.new
+        @attributes_by_class[klass] += list
+      end
+
+      def apply_filter(klass, list)
+        list.each do |attribute|
+          klass_name = (klass.name && klass < ActiveRecord::Base) ? klass.model_name.element : nil
+          filter = [klass_name, attribute.to_s].compact.join(".")
+          app.config.filter_parameters << filter unless app.config.filter_parameters.include?(filter)
```
Edouard-chin, line 69: Do you think it's worth re-compiling the filters?

The filters are precompiled and memoized on the first request, models that get loaded after the memoization happen (this is mostly in development when things are not eagerloaded) will not be precompiled. This may result in having an increasing number of filters to iterate through each time a new model is loaded.

https://github.com/rails/rails/blob/3954e4681de1013d1dc2efc166fb765be62650bd/railties/lib/rails/application.rb#L320-L321
https://github.com/rails/rails/blob/3954e4681de1013d1dc2efc166fb765be62650bd/railties/lib/rails/application.rb#L654-L659

that-jill:
Thanks for all the feedback. I was hoping to push some updates, but I wasn't able to during RailsConf and now I'm afk for a bit. I'll update things when I'm back in the office.

zzak:
For the test, I wonder if it's worth doing something like query logs tests in railties?
https://github.com/rails/rails/blob/main/railties/test/application/query_logs_test.rb

activerecord/lib/active_record/filter_attribute_handler.rb
that-jill: I believe this already works because the memoized object isn't replaced, the code just adds onto the pre-existing array.

It's reflected in the tests I wrote as well, but I tried adding a new attribute filter (`credit_card.digits`) in a test app and it worked without having to completely reload:

Screenshot 2025-08-14 at 4 25 03 PM

gmcgibbon (approved):

gmcgibbon:
I think the tests are good as-is. We can add a more complicated integration test if there's any regressions. Thanks @that-jill!

PR #55252, opened 2025-06-27
Moved LD_PRELOAD variable initialization to the Dockerfile from the entrypoint

igor-alexandrov:
### Motivation / Background

This Pull Request has been created because the `LD_PRELOAD` variable was initialized only for the shell used to start the process from the `docker-entrypoint` file. 

### Detail

This Pull Request changes the way `LD_PRELOAD` variable is initialized.

Libvips and jemalloc are both installed out of the box in Rails 8. However, libvips does not know that jemalloc is installed. `LD_PRELOAD` variable was only set in the `docker-entrypoint`script for the shell session that starts the Rails server, and does not get propagated to the list of environment variables of the container.

The official libvips guide suggests using jemalloc if you are on glibc-based distro: https://libvips.org/API/current/developer-checklist.html#linux-memory-allocator. 

The solution is to move the `LD_PRELOAD` variable initialization to the Dockerfile, so that it is available to all processes in the container.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt
@@ -17,13 +17,15 @@ WORKDIR /rails
 # Install base packages
 RUN apt-get update -qq && \
     apt-get install --no-install-recommends -y <%= dockerfile_base_packages.join(" ") %> && \
+    ln -s /usr/lib/$(uname -m)-linux-gnu/libjemalloc.so.2 /usr/local/lib/libjemalloc.so && \
     rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
-# Set production environment
+# Set production environment variables and enable jemalloc for reduced memory usage and latency.
 ENV RAILS_ENV="production" \
     BUNDLE_DEPLOYMENT="1" \
     BUNDLE_PATH="/usr/local/bundle" \
-    BUNDLE_WITHOUT="development"
+    BUNDLE_WITHOUT="development" \
+    LD_PRELOAD=/usr/local/lib/libjemalloc.so
 
 # Throw-away build stage to reduce size of final image
 FROM base AS build
--- a/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt
+++ b/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt
@@ -1,11 +1,5 @@
 #!/bin/bash -e
 
-# Enable jemalloc for reduced memory usage and latency.
-if [ -z "${LD_PRELOAD+x}" ]; then
-    LD_PRELOAD=$(find /usr/lib -name libjemalloc.so.2 -print -quit)
-    export LD_PRELOAD
-fi
-
 <% unless skip_active_record? -%>
 # If running the rails server then create or migrate existing database
 if [ "${@: -2:1}" == "./bin/rails" ] && [ "${@: -1:1}" == "server" ]; then


flavorjones:
@igor-alexandrov Nice improvement!

I would like to understand whether this is a problem that needs to be fixed in existing deployments. Kamal runs all commands through the entrypoint, so any long-running processes in a Kamal-deployed app _should_ pick up `LD_PRELOAD` (I tested this just to make sure, by looking at `/proc/${pid}/maps` to see if jemalloc was loaded (it was)).

So: can you say more about your setup, to help me understand your use case and why you're overriding the default entrypoint? I'm trying to determine if there's a general problem here with everyone's deployments that they should fix.

igor-alexandrov:
The problem is not in processes that are managed with Kamal, but in the `libvips`.

Below are the results of the `ldd` command without and with the `LD_PRELOAD` variable.

**Before**
```
rails@172:~$ ldd /usr/lib/x86_64-linux-gnu/libvips.so.42
        linux-vdso.so.1 (0x00007fff1f539000)
        libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007c8dd9c99000)
        libgio-2.0.so.0 => /lib/x86_64-linux-gnu/libgio-2.0.so.0 (0x00007c8dd9ab9000)
        libgobject-2.0.so.0 => /lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007c8dd9a5a000)
        libgmodule-2.0.so.0 => /lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007c8dd9a54000)
        libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007c8dd9a29000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007c8dd9a08000)
```

**After**
```
rails@172:~$ ldd /usr/lib/x86_64-linux-gnu/libvips.so.42
        linux-vdso.so.1 (0x00007fff8e293000)
        /usr/local/lib/libjemalloc.so (0x00007748ef0b2000)
        libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007748eef76000)
        libgio-2.0.so.0 => /lib/x86_64-linux-gnu/libgio-2.0.so.0 (0x00007748eed96000)
        libgobject-2.0.so.0 => /lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007748eed37000)
        libgmodule-2.0.so.0 => /lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007748eed31000)
        libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007748eed04000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007748eece5000)
```

The last listing shows `jemalloc` has been linked `/usr/local/lib/libjemalloc.so (0x00007748ef0b2000)`, which makes all the difference.

flavorjones:
I totally understand that you want libvips to use jemalloc, and that setting LD_PRELOAD is needed to do that.

What I'm asking is: what are you doing in the container that you're not going through the entrypoint which sets LD_PRELOAD? Again, I'm just trying to better understand whether there's a problem that widely affects existing Rails deployments.

flavorjones:
I'll note that any commands run via `kamal app exec` do _not_ currently go through the entrypoint, and so I'm :+1: on this change to address that use case.

igor-alexandrov:
Yes, the problem with the current implementation of the `docker-entrypoint` is that `LD_PRELOAD` will only be available for the processes that are spawned by this entrypoint and won't be available for others, e.g. for the `env` command results.

igor-alexandrov:
To conclude: because jemalloc and similar alternative allocators can improve performance for all processes in the container, not just Rails, they should be initialized at container startup instead of during Rails application startup.

byroot:
Same as @flavorjones, I'd like to understand what you are doing. Are you invoking things with `--entrypoint` hence are bypassing the default entrypoint?

I really don't mind this change, but I don't think it's really a bug so not worth backporting I think.

PR #55253, opened 2025-06-27
Do not clear `ExecutionContext` for nestable contexts

jenshenny:
Follow up to https://github.com/rails/rails/pull/55247

### Motivation / Background

From https://github.com/rails/rails/commit/d43dc9af18766f91eb450e1c6b8c8c5e64f925e6, a push and pop logic was implemented for `ExecutionContext`. 

However, in Shopify, there is an edge case where a test calls `Rails.application.reloader.reload!`. This causes the `ExecutionContext` to clear and the stack would be corrupted. I think we should skip this clear for nestable contexts since the pop after the reload is complete would set the context to the correct state.

### Bug Reproduction script

<details>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  CLOUDSMITH = "https://pkgs.shopify.io/basic/gems/ruby"

  # gem "rails"
  gem "job-iteration"
end

require "active_job/railtie"
require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  attribute :string_value, default: "default value set in Current"
end

class ReloadTest < ActiveJob::TestCase
  def test_1
    Current.string_value = "value set in test_1" 

    Rails.application.reloader.reload!

    assert_equal(
      "value set in test_1",
      Current.string_value,
      "Coming out of `ActiveSupport::CurrentAttributes.set`, the attributes should be restored to the values they were before the block"
    )
  end
end
```

### Expected
`Current.string_value` is `value set in test_1`

### Actual 
`NoMethodError: undefined method '[]' for nil`

</details>

cc: @byroot 

### Checklist

### DIFF
--- a/activesupport/lib/active_support/execution_context.rb
+++ b/activesupport/lib/active_support/execution_context.rb
@@ -94,7 +94,7 @@ module ActiveSupport
       end
 
       def clear
-        IsolatedExecutionState[:active_support_execution_context] = nil
+        IsolatedExecutionState[:active_support_execution_context] = nil unless @nestable
       end
 
       def current_attributes_instances
--- a/activesupport/test/current_attributes_test.rb
+++ b/activesupport/test/current_attributes_test.rb
@@ -309,7 +309,24 @@ class CurrentAttributesTest < ActiveSupport::TestCase
   test "set and restore attributes when re-entering the executor" do
     ActiveSupport::ExecutionContext.with(nestable: true) do
       # simulate executor hooks from active_support/railtie.rb
-      executor = Class.new(ActiveSupport::Executor)
+      reloader = Class.new(ActiveSupport::Reloader)
+      executor = reloader.executor
+
+      reloader.to_run do |instance|
+        ActiveSupport::ExecutionContext.push
+        instance.class_unload!
+      end
+
+      reloader.to_complete do |instance|
+        ActiveSupport::CurrentAttributes.clear_all
+        ActiveSupport::ExecutionContext.pop
+      end
+
+      reloader.before_class_unload do
+        ActiveSupport::CurrentAttributes.clear_all
+        ActiveSupport::ExecutionContext.clear
+      end
+
       executor.to_run do
         ActiveSupport::ExecutionContext.push
       end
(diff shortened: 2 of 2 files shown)


activesupport/lib/active_support/execution_context.rb
```diff
@@ -94,7 +94,7 @@ def pop
       end
 
       def clear
-        IsolatedExecutionState[:active_support_execution_context] = nil
+        IsolatedExecutionState[:active_support_execution_context] = nil unless @nestable
```
byroot, line 97: I dounbt this is the right fix. If code is reloaded everything should be cleared.

jenshenny: Yeah I'm not super confident on this fix. However, this is the only way I could think of to keep the existing behaviour. My interpretation is that a "push" is equivalent to a "clear" in tests / nestable. Before we reload, we push a record to a stack, the context is empty for the reload, and when the reload is finished, the state is reset to the outer state. 

```ruby
Current.string_value = "value set in test_1" 

Rails.application.reloader.reload!

# Current.string_value is "value set in test_1" 
```
Maybe that behaviour is incorrect and `Current.string_value` should be `nil` after the reload? Then we can make it so any remaining pops would return an empty hash after the clear.

byroot: `reload` should clear absolutely everything, but it also doesn't be called in a nested way. If that's the case I may have missed a legitimate reason why it would.

But from my perspective, `.reload` should only happen in dev anyway. The `test` environment has reloadable classes off by default.

So I'd need to know more about why the reloader is being called to take a decision here.

Schwad: I'm going to try and take a look at this tomorrow morning.

@byroot to confirm is your opinion that [the repro is the correct behaviour](https://github.com/rails/rails/pull/55253#:~:text=%23%20frozen_string_literal%3A%20true)? Otherwise I will look into this fix and think of something else. And probably build a bit more context

byroot: > is your opinion that [the repro is the correct behaviour](https://github.com/rails/rails/pull/55253#:~:text=%23%20frozen_string_literal%3A%20true)?

My opinion is mostly that I don't understand in which condition `reload` could be called nested inside a `push`.

So yes, the repro work as intended, but mostly because It's not meant to happen.

byroot:
I believe this has been addressed.

PR #55257, opened 2025-06-27
query_logs.rb: Doc source_location being expensive

henrik:
The changelog in https://github.com/rails/rails/releases/tag/v7.2.0 mentions this, but it was not mentioned in these docs.

It seems an easy mistake to make.

### DIFF
--- a/activerecord/lib/active_record/query_logs.rb
+++ b/activerecord/lib/active_record/query_logs.rb
@@ -29,6 +29,7 @@ module ActiveRecord
   # * +source_location+
   #
   # WARNING: Calculating the +source_location+ of a query can be slow, so you should consider its impact if using it in a production environment.
+  #
   # Also see {config.active_record.verbose_query_logs}[https://guides.rubyonrails.org/debugging_rails_applications.html#verbose-query-logs].
   #
   # Action Controller adds default tags when loaded:
--- a/activerecord/lib/active_record/query_logs.rb
+++ b/activerecord/lib/active_record/query_logs.rb
@@ -28,6 +28,9 @@ module ActiveRecord
   # * +database+
   # * +source_location+
   #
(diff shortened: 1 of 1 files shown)


henrik:
cc @fatkodima

fatkodima:
There is already a warning for this https://github.com/rails/rails/pull/55167.

henrik:
Thank you, @fatkodima! I didn't spot that warning since I was looking at the `QueryLogs` docs, so I think it would be good to have this in both places.

I've now modified this PR to use the same phrasing.

For the record, my previous phrasing was:

> Note that +source_location+ is costly and should be used carefully, typically only to gather data for a short period.

I think I prefer the guide's phrasing (which this PR now uses):

> WARNING: Calculating the +source_location+ of a query can be slow, so you should consider its impact if using it in a production environment.

It seems better to just point out the risk like this and then it's up to the user if they want to use it for a short period, or take the perf hit with eyes open, or use it in staging or whatever.

activerecord/lib/active_record/query_logs.rb
```diff
@@ -28,6 +28,9 @@ module ActiveRecord
   # * +database+
   # * +source_location+
   #
+  # WARNING: Calculating the +source_location+ of a query can be slow, so you should consider its impact if using it in a production environment.
+  # Also see {config.active_record.verbose_query_logs}[https://guides.rubyonrails.org/debugging_rails_applications.html#verbose-query-logs].
```
fatkodima:
```suggestion
  # WARNING: Calculating the +source_location+ of a query can be slow, so you should consider its impact if using it in a production environment.
  #
  # Also see {config.active_record.verbose_query_logs}[https://guides.rubyonrails.org/debugging_rails_applications.html#verbose-query-logs].
```

henrik:
Thank you, @fatkodima 😊

PR #55258, opened 2025-06-27
Fix association preloading when the foreign key is provided via `alias_attribute`

georgeu2000:
### Motivation / Background

When an association’s foreign key is exposed only through `alias_attribute`, **ActiveRecord** builds the eager-load query with the _alias_ instead of the physical column.  
The resulting `WHERE` clause compares the primary key against **NULL**, so no rows are returned—eager-loaded collections come back empty.

*Fixes #55231*

### Detail

* **Bug fix** – `AssociationScope#last_chain_scope` now resolves aliased foreign keys via `owner.class.attribute_aliases` before reading the owner attribute.
* **Regression test** – `test_preloading_with_alias_attribute_foreign_key` reproduces the failure on `main` and passes with the patch.

There are no public-API changes; the fix is internal and completely covered by tests.

### Additional information

* Full `activerecord` suite passes (`1497 runs, 0 failures`) on SQLite, PostgreSQL, and MySQL.
* No measurable impact on query plans or load times (micro-benchmark in issue thread).

### Checklist

- ✅  This Pull Request is limited to a single, well-scoped change.  
- ✅  Commit message explains **what** changed and **why**, and references the issue (`Fixes #55231`).  
- ✅  Tests are added to reproduce the bug and now pass with the fix.  
- ✅  CHANGELOG entry added for **activerecord** under “Bug Fixes”.

### DIFF
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -263,11 +263,23 @@ module ActiveRecord
             @key_conversion_required
           end
 
-          def derive_key(owner, key)
+          # `key` can be the real column *or* an `alias_attribute`.  Because
+          # `_read_attribute` bypasses alias resolution we translate any alias
+          # into the underlying column name before reading.
+          def derive_key(record, key)
             if key.is_a?(Array)
-              key.map { |k| convert_key(owner._read_attribute(k)) }
+              key.map { |k| convert_key(record._read_attribute(resolve_alias(record, k))) }
             else
-              convert_key(owner._read_attribute(key))
+              convert_key(record._read_attribute(resolve_alias(record, key)))
+            end
+          end
+
+          def resolve_alias(record, attr_name)
+            klass = record.class
+            if klass.attribute_alias?(attr_name.to_s)
+              klass.attribute_aliases[attr_name.to_s]
+            else
+              attr_name
             end
           end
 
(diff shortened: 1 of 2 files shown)


tylerwood-io:
The root cause is correct — \`_read_attribute\` bypasses alias resolution and reads the raw attribute store by column name, so when the FK is declared as an alias, the preloader reads nil and the association comes back empty.

Two observations:

1. The test sets both the alias and the real column: \`AliasedComment.create!(body: "Comment 1", parent_id: post.id, post_id: post.id)\`. This validates the preloader fix but doesn't fully exercise the alias-only path. Using just \`parent_id: post.id\` (without the explicit \`post_id:\`) would be a stronger assertion that the alias resolution works end-to-end through both the write and read paths.

2. \`resolve_alias\` is called per record per key during \`derive_key\`. For the common case (no aliases), this adds an \`attribute_alias?\` hash lookup per iteration. It's cheap, but since all records in a preloader batch share the same class, the alias resolution could be memoized once at the preloader level rather than re-checked per record. Something like:

\`\`\`ruby
def resolved_key(key)
  @resolved_keys ||= {}
  @resolved_keys[key] ||= begin
    if @klass.attribute_alias?(key.to_s)
      @klass.attribute_aliases[key.to_s]
    else
      key
    end
  end
end
\`\`\`

Neither of these is a blocker — the fix is correct as-is. Also, this likely warrants a CHANGELOG entry since it changes observable behavior (preloading with aliased FKs now works).

PR #55179, opened 2025-06-10
[RF-DOCS][ci skip] Update Threading and Code Execution Guide

OughtPuts:
### Motivation / Background

This PR has been created to update the Threading and Code Execution Guide as part of the Rails Documentation project.

### Detail

The following changes have been made:

- Phrasing and language choices have been updated for improved clarity and simplicity.
- The structure of the guide has been slightly altered to improve flow
- Information about asynchronous Active Record has been added. 
- The 'work in progress' label has been removed.
- Examples of wrapped application code have been added.
- Information on Isolated Execution State has been added.
- A section on `CurrentAttributes` has been added.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55262, opened 2025-06-30
Revert "[RF-DOCS][ci skip] Update Threading and Code Execution Guide (#55179)"

p8:
The threading guide rewrite "seems to be less accurate, and at points misleading".

This reverts commit 8ca0f945808c8daf604c3fd9ea278137f86a219d.


zzak:
How about reverting for now?

This way there is no pressure to get a review immediately.

p8:
@zzak Sure. I've reverted the commit. I'll create a new PR for the original PR, after this gets merged.

PR #50505, opened 2023-12-31
Add allow_browser to set minimum versions for your application

dhh:
Specify the browser versions that will be allowed to access all actions (or some, as limited by <tt>only:</tt> or <tt>except:</tt>).

Only browsers matched in the hash or named set passed to <tt>versions:</tt> will be blocked if they're below the versions specified.

This means that all other browsers, as well as agents that aren't reporting a user-agent header, will be allowed access.

A browser that's blocked will by default be served the file in public/426.html with a HTTP status code of "426 Upgrade Required".

In addition to specifically named browser versions, you can also pass <tt>:modern</tt> as the set to restrict support to browsers natively supporting webp images, web push, badges, import maps, CSS nesting, and CSS :has. This includes Safari 17.2+, Chrome 119+, Firefox 121+, Opera 104+.

You can use https://caniuse.com to check for browser versions supporting the features you use.

You can use <tt>ActiveSupport::Notifications</tt> to subscribe to events of browsers being blocked using the <tt>browser_block.action_controller</tt> event name.

Examples:

```ruby
class ApplicationController < ActionController::Base
  # Allow only browsers natively supporting webp images, web push, badges, import maps, CSS nesting + :has 
  allow_browser versions: :modern
end

class ApplicationController < ActionController::Base
  # All versions of Chrome and Opera will be allowed, but no versions of "internet explorer" (ie). Safari needs to be 16.4+ and Firefox 121+.
  allow_browser versions: { safari: 16.4, firefox: 121, ie: false }
end

class MessagesController < ApplicationController
  # In addition to the browsers blocked by ApplicationController, also block Opera below 104 and Chrome below 119 for the show action.
  allow_browser versions: { opera: 104, chrome: 119 }, only: :show
end
```

PR #55263, opened 2025-06-30
Rails New: Only add browser restrictions when using importmap

moonglum:
### Motivation / Background

When you generate a new Rails application right now, it will block older browsers. This behavior makes sense if you are using [`importmap-rails`](https://github.com/rails/importmap-rails), but if you for example use jsbundling-rails, it does not. This PR suggests only adding this blocker when you are using importmap, and not for all Rails applications.

### Detail

This Pull Request changes the generator for new Rails applications to only add:

```ruby
# Only allow modern browsers supporting webp images, web push, badges, import maps, CSS nesting, and CSS :has.
allow_browser versions: :modern
```

if the application is generated with `--javascript=importmap`. This is the default, so this PR does not change anything for people that do not override the JavaScript option.

If you, for example, set `--javascript=esbuild`, you don't need to block older browsers.

### Checklist

* [ ] ~~Tests are added or updated if you fix a bug or add a feature.~~ _There are no tests for the generators at this level from what I can see._

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -104,4 +104,8 @@
 
     *Petrik de Heus*
 
+*   Only add browser restrictions for a new Rails app when using importmap.
+
+    *Lucas Dohmen*
+
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/railties/CHANGELOG.md) for previous changes.
--- a/railties/lib/rails/generators/rails/app/templates/app/controllers/application_controller.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/app/controllers/application_controller.rb.tt
@@ -1,11 +1,9 @@
 class ApplicationController < ActionController::<%= options.api? ? "API" : "Base" %>
-<%- unless options.api? -%>
+<%- if using_importmap? -%>
   # Only allow modern browsers supporting webp images, web push, badges, import maps, CSS nesting, and CSS :has.
   allow_browser versions: :modern
-<%- if using_importmap? -%>
 
   # Changes to the importmap will invalidate the etag for HTML responses
   stale_when_importmap_changes
-<% end -%>
 <% end -%>
 end


moonglum:
I don't think the error in the docs-preview step is related to my changes :thinking: 

```
docker: Error response from daemon: manifest for 973266071021.dkr.ecr.us-east-1.amazonaws.com/builds:ruby-3-3-br-main not found: manifest unknown: Requested image not found.
```

byroot:
I'll merge this for now, but I think even better than a check during `rails new` we could move this to be a before_filter in `import-map` rails.

brunoprietog:
In my opinion this is not only about Importmaps, but about CSS as well. Not only old browsers are blocked by using Importmaps, I would say that even that is the least of it, as it can work perfectly well with a polyfill.

byroot:
I guess it's just that rejecting browsers based on User Agent string really rubs me the wrong way.

brunoprietog:
What I mean is that in order not to add that line, the conditions should be that you are using jsbundling and cssbundling, not that you are not using importmaps.

moonglum:
Thanks for the merge, @byroot :bow:

moonglum:
@brunoprietog  If the importmap approach works without blocking browsers, then this line should be removed entirely in my opinion :wink: Blocking browsers based on user agents is a huge anti pattern.

brunoprietog:
Importmaps natively does require modern browsers, what I am trying to say is that it is easier to escape that problem by adding a polyfill. For CSS features, on the other hand, it's not as fast as that, unless you use a preprocessor. Therefore, the reason for that line is not just because of importmaps. Even if you don't use importmaps but want to use modern CSS features, you still need that line, unless you opt to use something like esbuild via cssbundling. That's why I think we should only remove that line when both conditions are met, i.e. jsbundling and cssbundling are being used.

moonglum:
When you're talking about "polyfilling" modern CSS features, I guess you are talking about using PostCSS (not esbuild), right? There are many ways to get PostCSS into the project, and cssbundling is just one of them. I, for example, use PostCSS, but don't use cssbundling. So this "detection" of using PostCSS does not work. You could also use LightningCSS instead of PostCSS which is not supported by cssbundling.

Additionally, you can write modern CSS, but keep progressive enhancement in mind, then you don't need to block browsers as well.

I don't think adding this line for every Rails app makes sense. People who do not want to make their website work in "older" browsers can still add the line back in. From what I know, a new default Rails app using `importmaps-rails` will plain out not work in "non modern browsers", so adding it for those apps makes sense. But it would be the better solution to fix it in importmaps-rails, and not by adding it to new Rails apps.

brunoprietog:
Yes, you're right, I said wrong about esbuild.

Well, I don't think we'll reach a consensus, that's just my opinion. The idea of that line is to encourage the use of modern features to make #nobuild possible. Importmap is part of the story, but not all of it, as that would not be feasible at scale with vanilla CSS only. There's a reason people ended up using processors like sass or PostCSS anyway. Certainly, there may be a better approach to encourage or improve this.

ghiculescu:
FYI @dhh intent with this feature was that it is based on CSS support, not just JS: https://github.com/rails/rails/pull/50505#issuecomment-1894694783

dhh:
Yeah, this is just as much about CSS features as it is about JS. It's about shipping with a default setup where #nobuild is the reality with modern JS and modern CSS, without having to deal with old, broken versions of browsers that'll prevent that from working.

The reason it's exposed directly in the generated application_controller.rb is exactly such that people very easily can turn this off, if they either don't need modern JS/CSS or are using polyfills. That's totally fine. But the default for Rails is #nobuild out-of-the-box, and that's why this is there. 

So we'll have to revert this.

dhh:
Also, note that we're not blocking unknown browsers. Only #nobuild-incompatible versions of known browsers. I get that there's a purist take here that still dislikes that, and that's fine. This doesn't have to do everything for everyone. But we're actually using this feature in both Basecamp and HEY and ONCE/Campfire and all new products we're building. It's not speculative.

moonglum:
This is a bit too hand-wavy for me. Can you explain why you need the browser blocker for “no build”? I understand why you need it for import maps, but I don’t understand beyond that.

dhh:
It's in comment. Access to nested CSS and :has is essential for designs teams to be able to drop preprocessors like SCSS and go #nobuild for CSS. All this is extracted from how we actually managed to do it at 37s for major applications like HEY.

Either way, this is exposed so that you may easily change it in a new app. But the default is important to me. Rails Is a #nobuild Framework by default. So let's get this reverted.

moonglum:
> Access to nested CSS and :has is essential for designs teams to be able to drop preprocessors like SCSS and go #nobuild for CSS.

If that's essential for your design team at 37signals, then that's great. But it is of course an entirely arbitrary line that those two features are essential for Rails apps by default. Only 26% in the [State of CSS 2024 survey](https://2024.stateofcss.com/en-US/features) used nesting and have a positive sentiment. This doesn't even mean that those CSS developers see it as essential, just that they see it as positive. `:has` is slightly higher here with 36%, but even calling `:has` essential is a big stretch.

> But the default is important to me. Rails Is a #nobuild Framework by default.

This PR does not change that at all. The Rails default is not changed in any way, a `rails new` app will still have this line in there. This will only affect applications that opt out of the default `--javascript=importmap`. If my understandig of "no build" is correct (asset hashing and rewriting all files to use those hashes is okay, everything else is not), the apps that opt out of that default are not "no build" anyway:

* Either they go with bun, webpack, esbuild or rollup - then they are clearly "build".
* or they go with `--skip-js`. Then they are broken by default due to propshaft only adjusting the import statements in CSS files, and not in JS files. As it still fingerprints the JS files, all imports break. This is documented behavior and not a bug.

Therefore, in this PR, the generated code is only adjusted for Rails apps configured to not be "no build".

dhh:
I appreciate that, but I find that more confusing than just having baseline default that can be easily changed and is transparently exposed. Anyway, I'm thankful you took the time to do this PR. It was a fair proposal. I happen to disagree that this is the way to go based on the arguments already presented. All defaults are to some degree "arbitrary", and that's why I consider it a primary duty to continue to drive those decisions on the 22nd year of my work on this framework 😄

PR #55277, opened 2025-07-01
Add quotations for consistency with other envs

morgoth:
Follow up to https://github.com/rails/rails/pull/55252


byroot:
That's basically cosmetic, but OK.

morgoth:
Yes, cosmetic, but gives consistency in the generated file, so I guess it's worth it.
Thanks.

PR #55281, opened 2025-07-01
Add `require "rails"` to generated engines

Earlopain:
### Motivation / Background

Normally, gems get loaded in `application.rb` after `require "rails"` has already been run, so there is not much of an issue.

However, trying to require the gem in isolation will fail since the all sorts of things have not been set up yet (like the `Rails` constant).

### Additional information

I encountered this with `good_job` and `bundle console` but I don't think it's a very rare issue. `solid_queue` for example runs into the same thing (and also other problems).

If `rails` comes first in the gemfile, it works. If it comes after, it doesn't. Additionally, if you cherry-pick your frameworks and don't pull in all of rails, the gem order doesn't seem to matter at all.

cc @bensheldon 

### Checklist

### DIFF
--- a/guides/source/engines.md
+++ b/guides/source/engines.md
@@ -104,6 +104,8 @@ that provides the following:
     standard Rails application's `config/application.rb` file:
 
     ```ruby
+    require "rails"
+
     module Blorgh
       class Engine < ::Rails::Engine
       end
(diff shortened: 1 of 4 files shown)


rafaelfranca (changes requested):
I like the idea, but we should not be using Rails code in plugin code outside of the Railtie. Can you move the require to the railtie/engine file?

If you have a Rails plugin and you need to interface with `Rails` and constants defined in the `railties` gem, it should be done in the Railtie of your gem.

Earlopain:
That sounds fine to me, I moved it over there. Thanks for the review.

PR #55282, opened 2025-07-01
Bump required PostgreSQL version to 10.0

yahonda:
This pull request is based on the discussion at https://discuss.rubyonrails.org/t/proposal-bump-the-minimum-supported-postgresql-version-from-9-3-to-10/89422

### Motivation / Background

This pull request bumps the required PostgreSQL version to 10.0.

### Detail

PostgreSQL 18 (scheduled for release around September/October 2025) changes
the format of the cancel request key, which causes the following test to fail on Rails main.

This change was introduced in PostgreSQL core via https://github.com/postgres/postgres/commit/a460251f0a1ac987f0225203ff9593704da0b1a9

The issue is already addressed in https://github.com/ged/ruby-pg/pull/614 and the fix is included in the upcoming pg gem version v1.6.0 (currently released as v1.6.0.rc1).

The pg gem v1.6 raises its minimum required PostgreSQL version to 10 via https://github.com/ged/ruby-pg/pull/606

Rails main currently claims support for PostgreSQL 9.3+, which is incompatible with PostgreSQL 18 unless we raise the minimum supported version to 10.

I propose updating Rails main to require PostgreSQL 10 or newer, to ensure compatibility with PostgreSQL 18 and future releases.

### Additional information

- PostgreSQL 18 Beta 1

```sql
$ psql -d postgres
psql (18beta1)
Type "help" for help.

postgres=# select version();
                              version
--------------------------------------------------------------------
 PostgreSQL 18beta1 on x86_64-linux, compiled by gcc-15.1.1, 64-bit
(1 row)

postgres=#
```

- With pg 1.5.9 ActiveRecord::PostgresqlTransactionTest#test_raises_Interrupt_when_canceling_statement_via_interrupt fails
```ruby
$ bundle info pg
  * pg (1.5.9)
    Summary: Pg is the Ruby interface to the PostgreSQL RDBMS
    Homepage: https://github.com/ged/ruby-pg
    Documentation: http://deveiate.org/code/pg
    Source Code: https://github.com/ged/ruby-pg
    Changelog: https://github.com/ged/ruby-pg/blob/master/History.md
    Path: /home/yahonda/.local/share/mise/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/gems/pg-1.5.9
    Reverse Dependencies:
        queue_classic (4.0.0) depends on pg (>= 1.1, < 2.0)
$
$ ARCONN=postgresql bin/test test/cases/adapters/postgresql/transaction_test.rb -n test_raises_Interrupt_when_canceling_statement_via_interrupt
Using postgresql
Run options: -n test_raises_Interrupt_when_canceling_statement_via_interrupt --seed 17678

F

Failure:
ActiveRecord::PostgresqlTransactionTest#test_raises_Interrupt_when_canceling_statement_via_interrupt [test/cases/adapters/postgresql/transaction_test.rb:199]:
Expected 10.014372003 to be < 5.

bin/test test/cases/adapters/postgresql/transaction_test.rb:183

Finished in 10.067954s, 0.0993 runs/s, 0.1987 assertions/s.
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Bump the minimum PostgreSQL version to 10.0.
+
+    As part of this change, `supports_pgcrypto_uuid?` is deprecated because
+    `pgcrypto` provides `gen_random_uuid()` since PostgreSQL 9.4, which is
+    below the new 10.0 minimum.
+
+    *Yasuo Honda*
+
 *   Let `add_column` raise `ArgumentError` if `:null` is set to a true value
     when defining a primary key.
 
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -508,11 +508,7 @@ module ActiveRecord
 
             def select_min_column_value_sql(sequence)
               quoted_sequence = @adapter.quote_table_name(sequence)
-              if @adapter.database_version >= 10_00_00
-                "SELECT seqmin FROM pg_sequence WHERE seqrelid = #{@adapter.quote(quoted_sequence)}::regclass"
-              else
-                "SELECT min_value FROM #{quoted_sequence}"
-              end
+              "SELECT seqmin FROM pg_sequence WHERE seqrelid = #{@adapter.quote(quoted_sequence)}::regclass"
             end
 
             def reset_sequence_sql(sequence, max_value, min_value)
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -284,7 +284,7 @@ module ActiveRecord
       end
 
       def supports_insert_on_conflict?
-        database_version >= 9_05_00 # >= 9.5
+        true
       end
       alias supports_insert_on_duplicate_skip? supports_insert_on_conflict?
       alias supports_insert_on_duplicate_update? supports_insert_on_conflict?
@@ -295,7 +295,7 @@ module ActiveRecord
       end
 
       def supports_identity_columns? # :nodoc:
-        database_version >= 10_00_00 # >= 10.0
+        true
       end
 
       def supports_nulls_not_distinct?
@@ -303,7 +303,7 @@ module ActiveRecord
       end
 
       def supports_native_partitioning? # :nodoc:
-        database_version >= 10_00_00 # >= 10.0
+        true
       end
 
       if PG::Connection.method_defined?(:close_prepared) # pg 1.6.0 & libpq 17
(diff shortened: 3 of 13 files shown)


yahonda:
https://github.com/rails/rails/pull/55282/commits/69df7ae6d486042b634ca269fc8c4f5f282adc66 should be reverted before merging to the main. This is necessary until pg gem 1.6.0 is released. Refer to https://github.com/ged/ruby-pg/issues/639

matthewd:
Agree bumping our minimum on the server version is reasonable (thank you for the analysis on that ❤️). 

I am less sure about bumping our minimum on the `pg` gem though 😅 -- I think it's _reasonably_ unlikely that an app would have some other, tighter, dependency locking it... but even so, I feel like our tooling around version mismatch on optional deps is Not Great: if someone does a `bundle update rails` (so pg doesn't get bumped at the same time), I think the user experience is mediocre. 

Should we consider still allowing older pg versions and just accepting that older-gem users will have a bad time if connecting to a new server? Should we consider detecting that situation and giving a custom error? Or maybe it's fine to require the new gem? (Maybe we can improve that error experience, if I'm right that it's currently bad?)

yahonda:
Thanks for the feedback.

I suggested raising the pg gem requirement because versions ≤ 1.5 are known to break cancel behavior with PostgreSQL 18. While those pg versions work fine with PostgreSQL 17 and earlier, I’d prefer not to silently allow a broken setup once PostgreSQL 18 is out.

That said, I understand that requiring a newer version—especially since 1.6 is still only a release candidate—might be too aggressive. One possible compromise could be to detect the issue at runtime and emit a warning if pg ≤ 1.5 is used with PostgreSQL 18+. We can’t determine the server version until after a connection is established, so the check would need to happen post-connection.

yahonda:
> One possible compromise could be to detect the issue at runtime and emit a warning if pg ≤ 1.5 is used with PostgreSQL 18+. 

I’ve added the warning—feedback welcome. I’ll squash the commits once we’ve reached a decision.

yahonda:
Discussed with @matthewd and we have agreed to split this pull request into targets:

1. PostgreSQL Minimum Version Bump
Bump the minimum version of PostgreSQL to 10 by this pull request, which should not require all users to bump the pg gem

2. Compatibility Warning Implementation
Emit a warning if pg gem < 1.6 users attempt to connect to PostgreSQL 18+ because pg gem < 1.6 PG::Connection#cancel is not compatible with PostgreSQL 18
Reference: https://github.com/rails/rails/pull/55368

yahonda:
Now I'm seeking feedback on bumping the minimum supported PostgreSQL version from 9.3 to 10. 

While PostgreSQL 10 is also already EOL (November 2022), this bump would allow us to remove compatibility code specifically for PostgreSQL 9.3 and below. 
With PostgreSQL 18 already released and 19 expected this fall, I'd like to clean up this older compatibility code.

matthewd (approved):
I suspect we can now reasonably go higher again —- sorry this has been sitting for so long 😞❤️

yahonda:
Thanks for the review and approval. Yes, even PostgreSQL 10 is already 10 years old, so we'll need to bump the minimum version again.

PR #55286, opened 2025-07-01
Allow `current_page?` to match against specific HTTP method(s) with a `method:` option

bensheldon:
### Motivation / Background

The `current_page?` helper [has historically](https://github.com/rails/rails/issues/17855#issuecomment-279234771) only matched on `GET` and `HEAD` requests. This creates challenges when rendering content around validation errors on POST/PUT/PATCH actions. 

This PR adds an optional `method:` option that allows passing an explicit http method (or methods) to the `current_page?` helper to match specific http verbs.

This makes it easier to render simple navigations like this:

```erb
<ul class="nav nav-tabs">
  <li class="nav-item"><%= link_to "All posts", posts_path, class: ["nav-link", { active: current_page?(posts_path) }] %></li>
  <li class="nav-item"><%= link_to "Create post", new_post_path, class: ["nav-link", { active: current_page?(new_post_path) || current_page?(posts_path, method: :post) }] %></li>
</ul>
```

This is necessary because both the `PostsController#index` and `PostsController#create` share the same URL path but are differentiated by their HTTP method.

### Checklist


actionview/lib/action_view/helpers/url_helper.rb
```diff
@@ -538,24 +538,49 @@ def mail_to(email_address, name = nil, html_options = {}, &block)
...
+          request.get? || request.head?
+        else
+          request_method = request.method.downcase.to_sym
+          if method.is_a?(Array)
+            method << :head if method.include?(:get)
+            method.include?(request_method)
+          else
+            method == request_method
+          end
+        end
```
p8: The extra nesting of this if statement could probably be removed?
```suggestion
        method_matches = if method == :get
          request.get? || request.head?
        elsif method.is_a?(Array)
          request_method = request.method.downcase.to_sym
          method << :head if method.include?(:get)
          method.include?(request_method)
        else
          method == request_method
        end
```

Or maybe even:
```suggestion
        method_matches = case method
        when :get
          request.get? || request.head?
        when Array
          request_method = request.method.downcase.to_sym
          method << :head if method.include?(:get)
          method.include?(request_method)
        else
          method == request_method
        end
```

p8: Argh, nevermind. `request_method` is used in the last case.

bensheldon: yeah, I couldn't find the prettiest logic myself 😞  I was wanting to preserve the existing/default as a fast-path and the `else` block is "new behavior".

bensheldon: ... and I'm trying to avoid creating an array unnecessarily as I _could_ coerce everything to `Array(method)` but I think I've been warned off that before 😁

byroot:
Maybe I'm dense, and this change is probably fine, but I don't understand in which case you'd need pagination helpers on a POST.

p8:
@byroot `current_page?` is unrelated to pagination. It returns true if "the current request URI was generated by the given options".
https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-current_page-3F

bensheldon:
@byroot The `current_page?` helper is [almost exclusively](https://github.com/search?type=code&q=path%3Aapp%2Fviews%2F**%2F*.erb+%22current_page%3F%22) used for adding an "active" state to HTML links in navigation elements within shared-partials and layouts. 

That could be pagination, but my targeted need is when the navigation includes "New Post" or "Edit Post" links that render a form that could be submitted and have validation errors. In those cases, there can still be a requirement for an "active" state on the navigation when rendering validation errors. That's not possible without a change like this PR.

byroot:
Yeah, sorry I had a bit of a brain fart.

I skimmed over the historical issues, seems like that behavior was almost immediately regretted and people wanted it reverted but nothing happened.

I need to find some clear headed time to see if there is an elegant solution to this.

byroot:
Ok, backtracking a bit here. I haven't done frontend work in a very long time, and even then I don't remember using `current_page?`, so I need to read in tea leaves somewhat.

Here's what the doc says:

> True if the current request URI was generated by the given options.

With examples such as:

```ruby
current_page?(action: 'process')
# => false

current_page?(action: 'checkout')
# => true

current_page?(controller: 'library', action: 'checkout')
# => false
```

My read on this is that this helper kinda made sense back in the `url_for` days, but modern rails is more about path helpers.

Which begs the question of whether this need of your wouldn't be better solved by a newer method (or just a newer argument pattern) that is more "path helpers" native.

e.g. your example:

```ruby
active: current_page?(new_post_path) || current_page?(posts_path, method: :post)
```

What initially rubbed me the wrong way here is that `posts_path, method: :post` was weird, because we're referencing the helper for a route with a method that don't match it. But at the same time, `posts#create` doesn't have a route helper, so... I guess that's the best we can do here?

In the end that method is still very much `url_for` alike, so supporting `method:` is probably fine.

I'll try to review the code more in detail though.

actionview/lib/action_view/helpers/url_helper.rb
```diff
@@ -539,24 +539,49 @@ def mail_to(email_address, name = nil, html_options = {}, &block)
...
                 "in a #request method"
         end
 
-        return false unless request.get? || request.head?
-
         options ||= options_as_kwargs
-        check_parameters ||= options.is_a?(Hash) && options.delete(:check_parameters)
+        if options.is_a?(Hash)
+          check_parameters ||= options.delete(:check_parameters)
+          method ||= options.delete(:method)
```
byroot: I don't think you can do `||=` here. it works for `check_parameters` because it defaults to `false`, bit `method` defaults to a truthy value.

actionview/lib/action_view/helpers/url_helper.rb
```diff
@@ -539,24 +539,49 @@ def mail_to(email_address, name = nil, html_options = {}, &block)
...
+        end
+
+        method_matches = if method == :get
+          request.get? || request.head?
+        else
+          request_method = request.method.downcase.to_sym
```
byroot: We can use `Request#method_symbol` here.

actionview/lib/action_view/helpers/url_helper.rb
```diff
@@ -539,24 +539,46 @@ def mail_to(email_address, name = nil, html_options = {}, &block)
       #   current_page?('http://www.example.com/shop/checkout?order=desc&page=1')
       #   # => true
       #
-      # Let's say we're in the <tt>http://www.example.com/products</tt> action with method POST in case of invalid product.
+      #   Different actions may share the same URL path but have a different HTTP method. Let's say we
```
byroot: I think the indentation is wrong here.

actionview/lib/action_view/helpers/url_helper.rb
```diff
@@ -539,24 +539,47 @@ def mail_to(email_address, name = nil, html_options = {}, &block)
...
                 "in a #request method"
         end
 
-        return false unless request.get? || request.head?
+        if options.is_a?(Hash)
+          check_parameters = options.delete(:check_parameters) { check_parameters }
+          method = options.delete(:method) { method }
+        else
+          options ||= options_as_kwargs
+        end
```
byroot, line 571: I tweaked the options parsing code a bit.

PR #46378, opened 2022-10-29
Add link to external editor to rescue view

pedz:
This is a somewhat trivial diff.  Currently, when an exception is thrown, a page is rendered and part of the page has the paths and line numbers from the exception stack.  The lines are clickable and when clicked, the upper section changes to show more context of the line in the original source code.

The proposed change is to add an additional link whose text is the Unicode link symbol after the existing link.  The more common glyph was rejected by Unicode.

The href for the link symbol comes from Ruby code that first checks, at render time, to see if ActionDispatch::DebugView#ext_editor is defined.  If it is, it is called passing the "frame".  Otherwise, ActionDispatch::DebugView#int_editor is called.

int_editor creates a URL using the `file:` scheme.  While most browsers won't follow the link, it will show to the developer what is possible.  My example of ext_editor creates a URL using the `emacs:` scheme.

I've added a tool tip to try and tell the developers how they can configure the URL to invoke their favorite editor.

I thought the approach of checking to see if a method is defined and calling it would be preferred to having the user redefine the definition of an existing method.

Another piece of work to do would be to add something like the sample I have to the templates so that config/initializers/debug_view.rb would be created when the Rails app is created.

Having both links, the currently existing link that shows the code snippets and the new link that opens the file in the external editor, gives the users the best of both worlds.  They can view in the browser what happened and then click to get them placed in their favorite editor at the line number for the issue.

The key here is "their favorite editor" can be customized to their liking via the ext_editor method.

### Motivation / Background

This Pull Request has been created because adding a link to the exception rescue page to the files and lines that are in the exception stack will help the developer edit the area that caused the fault.

### Detail

This Pull Request changes:
M	actionpack/lib/action_dispatch.rb
M	actionpack/lib/action_dispatch/middleware/debug_view.rb
M	actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
A     actionpack/lib/action_dispatch/locale/en.yml

### Additional information

My example for what a developer can add to config/initializers/debug_view.rb:
```ruby
# frozen_string_literal: true

module ActionDispatch
  class DebugView < ActionView::Base # :nodoc:
    def ext_editor(frame)
      frame[:trace].sub(/([^:]+):([0-9]+)(:.*)?/, 'emacs:///Users/pedz/Source/hatred/\\1#\\2')
    end
  end
end
```

PR #55295, opened 2025-07-04
Add support to open files in the source code editor from the crash page

igorkasyanchuk:
Fix: https://github.com/rails/rails/pull/46378

### Motivation / Background

My main motivation is to make life easier. I want to be able to open my Cursor directly from the browser.

### Detail

I want to make it a part of Rails. For backward compatibility, I created a gem: https://github.com/igorkasyanchuk/editor_opener

I added support for the most popular editors (copied some code from the `better_errors` gem).

The developer can set an editor using the ENV variable (ENV["EDITOR"]). If it's not set, no changes will be applied.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Allow to open source file with a crash from the browser.
+
+    *Igor Kasyanchuk*
+
 *   Always check query string keys for valid encoding just like values are checked.
 
     *Casper Smits*
--- a/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
+++ b/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
@@ -148,15 +148,20 @@ module ActionDispatch
       application_trace_with_ids = []
       framework_trace_with_ids = []
       full_trace_with_ids = []
+      application_traces = application_trace.map(&:to_s)
 
+      full_trace = backtrace_cleaner&.clean_locations(backtrace, :all).presence || backtrace
       full_trace.each_with_index do |trace, idx|
+        filtered_trace = backtrace_cleaner&.clean_frame(trace, :all) || trace
+
         trace_with_id = {
           exception_object_id: @exception.object_id,
           id: idx,
-          trace: trace
+          trace: trace,
+          filtered_trace: filtered_trace,
         }
 
-        if application_trace.include?(trace)
+        if application_traces.include?(filtered_trace.to_s)
           application_trace_with_ids << trace_with_id
         else
           framework_trace_with_ids << trace_with_id
@@ -197,7 +202,7 @@ module ActionDispatch
 
     def source_extracts
       backtrace.map do |trace|
-        extract_source(trace)
+        extract_source(trace).merge(trace: trace)
       end
     end
 
--- a/actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
+++ b/actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
@@ -13,13 +13,17 @@
   <% end %>
 
   <% traces.each do |name, trace| %>
-    <div id="<%= "#{name.gsub(/\s/, '-')}-#{error_index}" %>" style="display: <%= (name == trace_to_show) ? 'block' : 'none' %>;">
+    <div id="<%= "#{name.gsub(/\s/, '-')}-#{error_index}" %>" class="trace-container" style="display: <%= (name == trace_to_show) ? 'block' : 'none' %>;">
       <code class="traces">
         <% trace.each do |frame| %>
-          <a class="trace-frames trace-frames-<%= error_index %>" data-exception-object-id="<%= frame[:exception_object_id] %>" data-frame-id="<%= frame[:id] %>" href="#">
-            <%= frame[:trace] %>
-          </a>
-          <br>
+          <div class="trace">
+            <% file_url = editor_url(frame[:trace]) %>
+            <%= file_url && link_to("✏️", file_url, class: "edit-icon") %>
+            <a class="trace-frames trace-frames-<%= error_index %>" data-exception-object-id="<%= frame[:exception_object_id] %>" data-frame-id="<%= frame[:id] %>" href="#">
+              <%= frame[:trace] %>
+            </a>
+            <br>
+          </div>
         <% end %>
       </code>
     </div>
--- a/actionpack/test/dispatch/debug_exceptions_test.rb
+++ b/actionpack/test/dispatch/debug_exceptions_test.rb
@@ -868,6 +868,16 @@ class DebugExceptionsTest < ActionDispatch::IntegrationTest
     end
   end
 
+  test "shows the link to edit the file in the editor" do
+    @app = DevelopmentApp
+    ActiveSupport::Editor.stub(:current, ActiveSupport::Editor.find("atom")) do
+      get "/actionable_error"
+
+      assert_select "code a.edit-icon"
+      assert_includes body, "atom://core/open"
+    end
+  end
+
   test "shows a buttons for every action in an actionable error" do
     @app = DevelopmentApp
     Rails.stub :root, Pathname.new(".") do
--- a/actionpack/lib/action_dispatch/middleware/debug_view.rb
+++ b/actionpack/lib/action_dispatch/middleware/debug_view.rb
@@ -55,6 +55,17 @@ module ActionDispatch
       end
     end
 
+    def editor_url(location, line: nil)
+      if editor = ActiveSupport::Editor.current
+        line ||= location&.lineno
+        absolute_path = location&.absolute_path
+
+        if absolute_path && line && File.exist?(absolute_path)
+          editor.url_for(absolute_path, line)
+        end
+      end
+    end
+
     def protect_against_forgery?
       false
     end
--- a/actionpack/lib/action_dispatch/middleware/templates/rescues/_source.html.erb
+++ b/actionpack/lib/action_dispatch/middleware/templates/rescues/_source.html.erb
@@ -11,8 +11,9 @@
           <tr>
             <td>
               <pre class="line_numbers">
-                <% source_extract[:code].each_key do |line_number| %>
-<span><%= line_number -%></span>
+                <% source_extract[:code].each_key do |line| %>
+                  <% file_url = editor_url(source_extract[:trace], line: line) %>
+<span><%= link_to_if file_url, line, file_url -%></span>
                 <% end %>
               </pre>
             </td>
--- a/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
+++ b/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
@@ -54,6 +54,30 @@
       font-size: 11px;
     }
 
+    .trace-container {
+      margin-top: 10px;
+    }
+
+    code.traces .trace {
+      display: flex;
+      align-items: center;
+      gap: 2px;
+    }
+
+    .edit-icon {
+      width: 16px;
+      height: 16px;
+      display: flex;
+      font-size: 13px;
+      align-items: center;
+      justify-content: center;
+      text-decoration: none;
+    }
+
+    .edit-icon:hover {
+      scale: 1.05;
+    }
+
     .response-heading, .request-heading {
       margin-top: 30px;
     }
--- a/actionpack/test/dispatch/exception_wrapper_test.rb
+++ b/actionpack/test/dispatch/exception_wrapper_test.rb
@@ -56,10 +56,12 @@ module ActionDispatch
 
     test "#source_extracts fetches source fragments for every backtrace entry" do
       exception = begin index; rescue TestError => ex; ex; end
+
       wrapper = ExceptionWrapper.new(nil, TopErrorProxy.new(exception, 1))
+      trace = wrapper.source_extracts.first[:trace]
 
       assert_called_with(wrapper, :source_fragment, ["lib/file.rb", 42], returns: "foo") do
-        assert_equal [ code: "foo", line_number: 42 ], wrapper.source_extracts
+        assert_equal [ code: "foo", line_number: 42, trace: trace ], wrapper.source_extracts
       end
     end
 
@@ -69,9 +71,10 @@ module ActionDispatch
       exc = begin ms_index; rescue TestError => ex; ex; end
 
       wrapper = ExceptionWrapper.new(nil, TopErrorProxy.new(exc, 1))
+      trace = wrapper.source_extracts.first[:trace]
 
       assert_called_with(wrapper, :source_fragment, ["c:/path/to/rails/app/controller.rb", 27], returns: "nothing") do
-        assert_equal [ code: "nothing", line_number: 27 ], wrapper.source_extracts
+        assert_equal [ code: "nothing", line_number: 27, trace: trace ], wrapper.source_extracts
       end
     end
 
@@ -81,9 +84,10 @@ module ActionDispatch
       exc = begin invalid_ex; rescue TestError => ex; ex; end
 
       wrapper = ExceptionWrapper.new(nil, TopErrorProxy.new(exc, 1))
+      trace = wrapper.source_extracts.first[:trace]
 
       assert_called_with(wrapper, :source_fragment, ["invalid", 0], returns: "nothing") do
-        assert_equal [ code: "nothing", line_number: 0 ], wrapper.source_extracts
+        assert_equal [ code: "nothing", line_number: 0, trace: trace ], wrapper.source_extracts
       end
     end
 
@@ -95,9 +99,10 @@ module ActionDispatch
       exception = begin throw_syntax_error; rescue SyntaxError => ex; ex; end
 
       wrapper = ExceptionWrapper.new(nil, TopErrorProxy.new(exception, 1))
+      trace = wrapper.source_extracts.first[:trace]
 
       assert_called_with(wrapper, :source_fragment, ["lib/file.rb", 42], returns: "foo") do
-        assert_equal [ code: "foo", line_number: 42 ], wrapper.source_extracts
+        assert_equal [ code: "foo", line_number: 42, trace: trace ], wrapper.source_extracts
       end
     end
 
@@ -123,7 +128,7 @@ module ActionDispatch
         code[lineno + i] = line
       end
       code[lineno + 2] = ["        1", ".time", "\n"]
-      assert_equal({ code: code, line_number: lineno + 2 }, wrapper.source_extracts.first)
+      assert_equal({ code: code, line_number: lineno + 2, trace: wrapper.source_extracts.first[:trace] }, wrapper.source_extracts.first)
     end
 
     class_eval "def _app_views_tests_show_html_erb;
@@ -142,7 +147,8 @@ module ActionDispatch
 
       assert_equal [{
         code: { 1 => "translated @ _app_views_tests_show_html_erb:3" },
-        line_number: 1
+        line_number: 1,
+        trace: wrapper.source_extracts.first[:trace]
       }], wrapper.source_extracts
     end
 
@@ -168,17 +174,20 @@ module ActionDispatch
       extracts = wrapper.source_extracts
       assert_equal({
         code: { 1 => "translated @ _app_views_tests_nested_html_erb:5" },
-        line_number: 1
+        line_number: 1,
+        trace: extracts[0][:trace]
       }, extracts[0])
       # extracts[1] is Array#each (unreliable backtrace across rubies)
       assert_equal({
         code: { 1 => "translated @ _app_views_tests_nested_html_erb:4" },
-        line_number: 1
+        line_number: 1,
+        trace: extracts[2][:trace]
       }, extracts[2])
       # extracts[3] is Array#each (unreliable backtrace across rubies)
       assert_equal({
         code: { 1 => "translated @ _app_views_tests_nested_html_erb:3" },
-        line_number: 1
+        line_number: 1,
+        trace: extracts[4][:trace]
       }, extracts[4])
     end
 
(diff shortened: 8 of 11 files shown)


igorkasyanchuk:
I know my tests are not perfect, I'll need some help to make everything work properly, if there is a chance it will be merged.

byroot:
I don't have time to do a proper review just now, but I'm very much in favor of such feature, however the implementation will need quite a bit of work. I'll try find time to review / improve it in the next few days.

igorkasyanchuk:
@byroot, thanks, I look forward to hearing your feedback. I hope we can make it together.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,85 @@
...
+        editor.present?
+      end
+
+      def editor
+        @editor ||= ENV["EDITOR"].present? && KNOWN_EDITORS.find { |editor| editor[:symbols].include?(ENV["EDITOR"].to_sym) }
+      end
+    end
+
+    def initialize(trace, line_number: nil)
+      @trace = trace.to_s.strip
```
Edouard-chin: We should be able to largely simplify the `TraceToFileExtractor` by passing a `Thread::Backtrace::Location` object and using methods such as `lino` and `absolute_path`.

We already get a `Thread::Backtrace::Location` object when rendering sources, but for traces we'd have to do a bit of tweaking.

--------- 

https://github.com/rails/rails/blob/3954e4681de1013d1dc2efc166fb765be62650bd/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L152-L156

The `full_trace` returns an array of string after the backtrace has been through all filters and silencers, but we'd need to preserve the original `Thread::Backtrace::Location`. Lucky for us there is a `clean_locations` method that was [added last week](https://github.com/rails/rails/pull/55255) which should help us here. 
Something along the line of

```diff
diff --git a/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb b/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
index bfb67d8129b..fe558a68ded 100644
--- a/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
+++ b/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
@@ -149,14 +149,18 @@ def traces
       framework_trace_with_ids = []
       full_trace_with_ids = []

+      full_trace = backtrace_cleaner&.clean_locations(backtrace, :all) || backtrace
       full_trace.each_with_index do |trace, idx|
+        filtered_trace = backtrace_cleaner&.clean_frame(trace, :all) || trace
+
         trace_with_id = {
           exception_object_id: @exception.object_id,
           id: idx,
-          trace: trace
+          filtered_trace: filtered_trace,
+          trace: trace,
         }

-        if application_trace.include?(trace)
+        if application_trace.include?(filtered_trace)
           application_trace_with_ids << trace_with_id
         else
           framework_trace_with_ids << trace_with_id

```

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,85 @@
...
+      { symbols: [:cursor], sniff: /cursor/i, url: "cursor://file/%{file}:%{line}" },
+    ]
+
+    class << self
+      def open_in_editor?
+        editor.present?
+      end
+
+      def editor
+        @editor ||= ENV["EDITOR"].present? && KNOWN_EDITORS.find { |editor| editor[:symbols].include?(ENV["EDITOR"].to_sym) }
```
Edouard-chin: I think we should fallback to a `file://` scheme in the case no EDITOR is set or the editor isn't supported. This has a couple of advantages:

1. Some users may have a browser addon that scans for `file:///` handler and let user click to open native apps (e.g. [Local Filesystem Links](https://addons.mozilla.org/en-US/firefox/addon/local-filesystem-links/?src=search)
2. We get rid of the `open_in_editor?` check

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
```diff
@@ -13,13 +13,23 @@
...
+    <div id="<%= "#{name.gsub(/\s/, '-')}-#{error_index}" %>" class="trace-container" style="display: <%= (name == trace_to_show) ? 'block' : 'none' %>;">
       <code class="traces">
         <% trace.each do |frame| %>
-          <a class="trace-frames trace-frames-<%= error_index %>" data-exception-object-id="<%= frame[:exception_object_id] %>" data-frame-id="<%= frame[:id] %>" href="#">
-            <%= frame[:trace] %>
-          </a>
-          <br>
+          <div class="trace">
+            <% file_url = ActionDispatch::TraceToFileExtractor.open_in_editor? && ActionDispatch::TraceToFileExtractor.new(frame[:trace]).call %>
+            <% if file_url.present? %>
```
Edouard-chin: Minor suggestion: We could use [link_to_if](https://api.rubyonrails.org/v8.0.1/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to_if)

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,85 @@
...
+
+# :markup: markdown
+
+module ActionDispatch
+  # # Extract file and line from a trace and open it in the editor
+  #
+  # This is used to open the file in the editor when the user clicks on the edit icon. You have to set the `EDITOR` environment variable to the editor you want to use.
+  class TraceToFileExtractor
+    KNOWN_EDITORS = [
+      { symbols: [:atom], sniff: /atom/i, url: "atom://core/open/file?filename=%{file}&line=%{line}" },
```
Edouard-chin: It would be great to let user define their own scheme. Maybe introduce an env var `RAILS_FILE_LINK_FORMAT=brackets://open/file/...`. This would allow users working with unusual editor or even having custom url handlers.

Edouard-chin (commented):
Great addition !

I made a couple suggestions, which should make the implementation simpler.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
byroot: Yes, `backtrace_locations` is the way to go, just note that in rare case it may not exist. But in such case, it's fine to just not render links.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
byroot: We should also do like bundler, first check for `RAILS_EDITOR` and then fallback to `EDITOR`. It's not rare to have EDITOR point to a terminal editor like `vim` or `nano` but to use a GUI editor for code.

Also please try to refactor the `ENV` access out of this class, it should receive the content of the `ENV` not check it itself, this will also make your testing easier.

actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+*   Allow to open source file with a crash from the browser.
+
+    *Igor Kasyanchuk*
+
+
```
byroot:
```suggestion
```

✂️

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
byroot: I think an API to register editors is preferable, this allow to both define yours in an initializer, but also to just publish small gems to add a specific editor support.

I also think we might want to move this in Active Support, so it's more generic, and have a thin layer, or just use it in Action Pack. In the future it could be extended to also generate links in terminal (e.g. iTerm2 GNONE terminal and a few others support links).

byroot (commented):
Édouard already mentioned most of the issues.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
igorkasyanchuk: Thank you, that really simplifies the code.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
igorkasyanchuk: I've added `ENV["RAILS_FILE_LINK_FORMAT"]`, I hope this will be enough to allow defining custom editors.

as for moving code to a different place, or initializers, please tell if I should do it now or if this is something we can do later?

If now, what are the recommendations?

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
igorkasyanchuk: @Edouard-chin  I used file_url && link_to.
I think with `link_to_if` I need to set text as nil. So it will render nil, if condition is false.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
igorkasyanchuk: please review the latest code

actionpack/test/dispatch/trace_to_file_extractor_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    @original_editor = ENV["EDITOR"]
+    ENV["EDITOR"] = "atom"
+  end
+
+  teardown do
+    ENV["EDITOR"] = @original_editor
+  end
+
+  test "trace to file extractor with exception in app code" do
+    Rails.stub(:root, "/Users/john/projects/rails") do
```
Edouard-chin: No need to stub `root` anymore.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,47 @@
...
+      { symbols: [:vscode, :code], url: "vscode://file/%{file}:%{line}" },
+      { symbols: [:vscodium, :codium], url: "vscodium://file/%{file}:%{line}" },
+      { symbols: [:windsurf], url: "windsurf://file/%{file}:%{line}" },
+      { symbols: [:zed], url: "zed://file/%{file}:%{line}" },
+      { symbols: [:nova], url: "nova://open?path=%{file}&line=%{line}" },
+      { symbols: [:cursor], url: "cursor://file/%{file}:%{line}" },
+    ]
+
+    class << self
+      def call(trace, line_number: nil)
```
Edouard-chin: Do we have to pass the `line_number` now that we always have a `Backtrace::Location` object?

actionpack/test/dispatch/debug_exceptions_test.rb
```diff
@@ -149,6 +149,7 @@ def self.build_app(app, *args)
   RoutesApp = Struct.new(:routes).new(SharedTestRoutes)
   ProductionApp  = build_app(Boomer.new(false), RoutesApp)
   DevelopmentApp = build_app(Boomer.new(true), RoutesApp)
+  OpenInEditorApp = build_app(Boomer.new(true), RoutesApp)
```
Edouard-chin: No need for a new App, we can use the `DevelopmentApp` above.

Edouard-chin (commented):
Thanks for simplifying the implementation. It looks much better.

I gave this a quick review, but could you fix all the tests you introduced and try to have a green CI, so it's easier for us to review.

Thanks again for your work!

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
igorkasyanchuk: yes, this is to make this line clickable, otherwise, all will be pointed to the same location (location of the crash)

image

actionpack/test/dispatch/debug_exceptions_test.rb
igorkasyanchuk: glad that now you consider code as acceptable :) and now it's time to fix specs, thanks for the support!

actionpack/test/dispatch/trace_to_file_extractor_test.rb
igorkasyanchuk: done

igorkasyanchuk:
This is how the page looks now

image

@Edouard-chin @byroot any suggestions on how to fix tests? It's really beyond my understanding :( 

I tried to debug, but I'm not sure why "Application Trace" is empty.

image

igorkasyanchuk:
Actually, I was able to make some fixes, and fix the specs, but I had to introduce a new variable and change this check

`if application_traces.include?(filtered_trace.to_s)`

the reason why application_traces contains different objects than filtered_trace (even if `.to_s` value are equal)

please advice what to do

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
```diff
@@ -148,15 +148,20 @@ def traces
...
         trace_with_id = {
           exception_object_id: @exception.object_id,
           id: idx,
-          trace: trace
+          trace: trace,
+          filtered_trace: filtered_trace,
         }
 
-        if application_trace.include?(trace)
+        if application_traces.include?(filtered_trace.to_s)
```
byroot, line 164: Not a huge concern because it's an error page, but casting with `to_s` on both side is wasteful. It's much cheaper to compare two `Backtrace::Location` objects, than two long strings.

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
```diff
@@ -13,13 +13,21 @@
...
-          <a class="trace-frames trace-frames-<%= error_index %>" data-exception-object-id="<%= frame[:exception_object_id] %>" data-frame-id="<%= frame[:id] %>" href="#">
-            <%= frame[:trace] %>
-          </a>
-          <br>
+          <div class="trace">
+            <% file_url = ActionDispatch::TraceToFileExtractor.call(frame[:trace]) %>
+            <%= file_url && link_to(file_url, class: "edit-icon") do %>
+              <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="size-6">
+                <path stroke-linecap="round" stroke-linejoin="round" d="m16.862 4.487 1.687-1.688a1.875 1.875 0 1 1 2.652 2.652L10.582 16.07a4.5 4.5 0 0 1-1.897 1.13L6 18l.8-2.685a4.5 4.5 0 0 1 1.13-1.897l8.932-8.931Zm0 0L19.5 7.125M18 14v4.75A2.25 2.25 0 0 1 15.75 21H5.25A2.25 2.25 0 0 1 3 18.75V8.25A2.25 2.25 0 0 1 5.25 6H10" />
+              </svg>
```
byroot: I have to ask, where did you get that SVG? because there might be licensing concerns.

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
igorkasyanchuk: What exactly I can use for comparison? because they have different object_id's.

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
igorkasyanchuk: I took icon from https://heroicons.com/outline
as I see they use MIT license

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
igorkasyanchuk: @Edouard-chin do you have any other questions, or I can resolve the conversation

Edouard-chin: Ah ! Thanks for the explanation, I forgot we provided contextual lines.

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
Edouard-chin: > Actually, I was able to make some fixes, and fix the specs, but I had to introduce a new variable and change this check

I dug on this and I'm ok with this change although it's a bit hacky. Let's see Jean's opinion though. The real fix may be in the BacktraceSilencer class, but it's out of scope of this PR.

Basically the issue is the BacktraceSilencer class is inconsistent on what it returns and make thigs hard to work with in our case. When `BacktraceCleaner#clean` is called, it can return either an array of `Backtrace::Location` object or an array of `String`, this depends whether the Cleaner applies filters: https://github.com/rails/rails/blob/8f42c2b4a1a13a90035cbb9b67ec7b931265105e/activesupport/lib/active_support/backtrace_cleaner.rb#L210-L215

The second issue is that when we call `#clean_frame`, we get back a `String` instead of a `Backtrace::Location`, even though the frame didn't go through any filter. https://github.com/rails/rails/blob/8f42c2b4a1a13a90035cbb9b67ec7b931265105e/activesupport/lib/active_support/backtrace_cleaner.rb#L73-L74

Ultimately when we end up comparing apples with oranges.

Edouard-chin (commented):
Apart from Jean's suggestion regarding ENV access, I don't have any other comments. 

Thank you!!

igorkasyanchuk:
> Apart from Jean's suggestion regarding ENV access, I don't have any other comments.
> 
> Thank you!!

Which suggestion regarding ENV access do you mean? Now with ENV["RAILS_EDITOR"] and ENV["RAILS_FILE_LINK_FORMAT"] we have full flexibility. Or do I need to change something?

Edouard-chin:
This one : https://github.com/rails/rails/pull/55295#discussion_r2188122094

> Also please try to refactor the ENV access out of this class, it should receive the content of the ENV not check it itself, this will also make your testing easier.

I'll let Jean decide if this is something can iterate on in a future PR.

actionpack/lib/action_dispatch.rb
```diff
@@ -83,6 +83,7 @@ class MissingController < NameError
     autoload :RemoteIp
     autoload :ServerTiming
     autoload :ShowExceptions
+    autoload :TraceToFileExtractor
     autoload :SSL
     autoload :Static
```
zzak:
```suggestion
    autoload :SSL
    autoload :Static
    autoload :TraceToFileExtractor
```

Knit: to put the autoload at the end (in alphabetical order) also for a cleaner diff

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
zzak: I think there is no precedent for using external icons (even MIT-licensed) in the Rails source code, we should just use emoji IMO.

zzak:
Please [squash your commits](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#squashing-commits) too please. :bow:

thibaudgg:
Nice, I was looking at exactly how to configure this! 🙌🏻

Once merged, this PR could be marked as closed: https://github.com/rails/rails/pull/46378

actionpack/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
igorkasyanchuk: image

replaced icon with emoji

actionpack/lib/action_dispatch.rb
igorkasyanchuk: sorted

igorkasyanchuk:
@zzak PR is ready for a review, let me know if I need to do anything

actionpack/lib/action_dispatch.rb
```diff
@@ -66,25 +66,26 @@ class MissingController < NameError
...
     autoload :DebugLocks
     autoload :DebugView
     autoload :ExceptionWrapper
     autoload :Executor
     autoload :Flash
+    autoload :HostAuthorization
     autoload :PublicExceptions
     autoload :Reloader
     autoload :RemoteIp
+    autoload :RequestId
```
zzak: Please don't change autoloads unrelated to this PR. I only meant, that since you were adding `TraceToFileExtractor` it made sense to put at the end.

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
```diff
@@ -148,15 +148,20 @@ def traces
       application_trace_with_ids = []
       framework_trace_with_ids = []
       full_trace_with_ids = []
+      application_traces = application_trace.map(&:to_s)
 
+      full_trace = backtrace_cleaner&.clean_locations(backtrace, :all) || backtrace
```
zzak: If there are no locations, it will return an empty array, not nil:
https://edgeapi.rubyonrails.org/classes/ActiveSupport/BacktraceCleaner.html#method-i-clean_locations

actionpack/lib/action_dispatch.rb
igorkasyanchuk: fixed

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
igorkasyanchuk: fixed

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+# :markup: markdown
+
+module ActionDispatch
+  # # Extract file and line from a trace and open it in the editor
+  #
+  # This is used to open the file in the editor when the user clicks on the edit icon. You have to set the `EDITOR` environment variable to the editor you want to use.
+  class TraceToFileExtractor
```
zzak: Maybe leave this `:nodoc:` for now, otherwise making changes to the list of known editors becomes a breaking change

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
byroot: Nevermind, I remembered `Thread::Backtrace::Location` objects had a `==` method, but actually you are right they don't.

actionpack/lib/action_dispatch/middleware/trace_to_file_extractor.rb
```diff
@@ -0,0 +1,55 @@
...
+    KNOWN_EDITORS = [
+      { symbols: [:atom], url: "atom://core/open/file?filename=%{file}&line=%{line}" },
+      { symbols: [:cursor], url: "cursor://file/%{file}:%{line}" },
+      { symbols: [:emacs, :emacsclient], url: "emacs://open?url=file://%{file}&line=%{line}" },
+      { symbols: [:idea], url: "idea://open?file=%{file}&line=%{line}" },
+      { symbols: [:macvim, :mvim], url: "mvim://open?url=file://%{file_unencoded}&line=%{line}" },
```
byroot: What's `file_unencoded` ? Just a typo?

byroot:
Alright, I refactored the whole thing a bit, mostly moved the Editor knowledge in Active Support, with an API to register unknown editors (typically from an initializer). Also replaced the "service object" by a simple view helper.

I think it's good to go now.

elalemanyo:
@igorkasyanchuk @byroot  Thanks for adding this feature — it’s super useful!

Quick question: is there (or will there be) a way to make `RAILS_EDITOR` work when the Rails app is running inside a devcontainer/Docker?

Right now the editor link on the error page contains the container path (e.g. /workspaces/app/...), but VS Code on the host receives that URL and can’t find the file because the path only exists inside the container.

Is there any planned support for:

- translating container paths to host paths?
- emitting `vscode-remote://` URLs?
- allowing a custom path-mapper for `RAILS_EDITOR`?
- or any recommended way to use this feature with Docker/devcontainers?

Thanks again for the great work!

igorkasyanchuk:
maybe we can introduce something like host_app_path and use it instead of the local docker app path

you can try to experiment in my original gem https://github.com/igorkasyanchuk/editor_opener for older Rails versions, with this approach.  

and later prepare a PR into Rails :)

elalemanyo:
@igorkasyanchuk Thanks! I went ahead and introduced `RAILS_HOST_APP_PATH` as you suggested and opened a [pull request](https://github.com/rails/rails/pull/56263) with that change 🙂

PR #55297, opened 2025-07-05
[ci skip] Remove redundant package from installation command

anonychun:
### Detail

This pull request remove duplicate package from pacman installation command in development dependencies guide


p8:
Thanks @anonychun !

PR #55298, opened 2025-07-05
Fix time attribute dirty tracking with timezone conversions

### LINKED ISSUE #55118: time attributes falsely marked as changed due to date shift on timezone conversion
### Steps to reproduce

```ruby
# config/application.rb
config.time_zone = 'Tokyo'
config.active_record.default_timezone = :utc

# Migration
create_table :users do |t|
  t.time :created_at
end

# Model
class User < ApplicationRecord; end

# Console
user = User.create created_at: "08:00"
user.reload

user.created_at
# => 2000-01-02 08:00:00.000000000 JST +09:00

user.created_at = "08:00"

user.created_at
# => 2000-01-01 08:00:00.000000000 JST +09:00

user.changed_attribute_names_to_save
# => [:created_at]  ← unexpected!
```

Although the same `"08:00"` value is assigned again, Active Record considers the attribute as changed.

---

### Expected behavior

The `created_at` attribute should not be marked as changed when assigning the same time string (`"08:00"`) again.
This should hold true even when the application and database time zones differ, and internal `Time` objects have different date parts.

---

### Actual behavior

Assigning `"08:00"` again results in `:created_at` being marked as changed.

This happens because:

```ruby
user.created_at.to_s
# => "2000-01-01 08:00:00 +0900"

user.created_at.utc.to_s
# => "1999-12-31 23:00:00 UTC"
```

Due to the timezone conversion, the internal `Time` object has a different date, triggering false positives in dirty tracking.

---

### Proposal

I propose introducing a config option such as:

```ruby
config.active_record.use_fixed_date_for_time_attributes = true
```

When enabled, Active Record would normalize `time`-typed values to a fixed date (e.g., `2000-01-01`) internally.
This would eliminate unnecessary dirty tracking caused by timezone-induced date mismatches, while preserving timezone support and default behavior unless explicitly enabled.

Would the Rails team be open to such a feature?

---

### System configuration

**Rails version**: 7.x.x
**Ruby version**: 3.x.x
**Database**: PostgreSQL / MySQL (both reproduce the issue)
**OS**: macOS / Linux

prateekkish:
## Summary

This PR fixes an issue where time-only attributes are incorrectly marked as changed when setting the same time value due to timezone conversion causing date shifts.

## Problem

When using time-only attributes with timezone-aware settings, the internal date component can shift during timezone conversions (e.g., from 2000-01-01 to 2000-01-02), causing Rails to incorrectly detect the attribute as changed even when the time value remains the same.

### Before (Bug)
```ruby
# Configuration
Time.zone = "Tokyo"
ActiveRecord.default_timezone = :utc

user = User.create!(created_at: "14:30")
user.reload

user.created_at = "14:30"
user.changed_attribute_names_to_save
# => ["created_at"] # Incorrectly marked as changed!
```

### After (Fixed)
```ruby
# Configuration
config.active_record.use_fixed_date_for_time_attributes = true

user = User.create!(created_at: "14:30")
user.reload

user.created_at = "14:30"
user.changed_attribute_names_to_save
# => [] # Correctly not marked as changed
```

## Solution

This PR ensures time-only attributes maintain a fixed date of 2000-01-01 during timezone conversions. This prevents false positive dirty tracking.

The fix:

- Modifies `TimeZoneConverter` to maintain the fixed date
- Only affects time-only attributes, not datetime attributes

Fixes #55118

### DIFF
--- a/activerecord/test/cases/attribute_methods/time_zone_converter_test.rb
+++ b/activerecord/test/cases/attribute_methods/time_zone_converter_test.rb
@@ -2,6 +2,7 @@
 
 require "cases/helper"
 require "active_support/core_ext/enumerable"
+require "models/topic"
 
 module ActiveRecord
   module AttributeMethods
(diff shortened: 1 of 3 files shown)


byroot:
Is a configuration really needed? What's the reason not to want that behavior?

prateekkish:
@byroot This looks ready to me. Could you help with a review or tagging somebody who might?

activerecord/test/cases/attribute_methods/time_zone_converter_test.rb
```diff
@@ -15,6 +16,77 @@ def test_comparison_with_date_time_type
           assert_equal value, value_from_cache
           assert_not_equal value, "foo"
         end
+
+        def test_time_attributes_with_fixed_date_normalization
```
byroot, line 20: This test passes on the `main` branch, which suggest it's not correctly ensuring the bug is fixed.

byroot: Alright, I rewrote the test so it properly fails on `main`.

prateekkish:
Thanks a bunch @byroot 
Was away for the weekend so couldn't instantly reply to your review. 

Happy to see this merged.

canadakickass:
In which version of rails will this be available?

PR #53449, opened 2024-10-25
Rate Limiting: Allow to define shared rate limit across controllers.

arthurwozniak:
### Motivation / Background

This Pull Request has been created because we would like to define rate limits shared for certain controllers (e.g. for all API endpoints). At the moment, rate-limit is evaluated on controller-based level due to `controller_path` part of cache key and configuration is not flexible in this  case.

### Detail

In order to achieve cross-controller rate limits, I added option `:scope` arguments which may be passed to `rate_limit` method. It is expected to be symbol or string. 

If multiple controllers (or e.g. `ApplicationController`) defines rate-limit with `:scope` value provided, rate limit counter is incremented across controller actions, which allows user to define e.g. global rate limit to 100req/s.

```ruby
rate_limit to: 100, within: 1.minute, scope: :api_global
```

### Additional information

I am not happy from provided test case. Assigning `@controller` in `cross-controller rate limit` test feels kinda hackish, however I cannot find better solution. Feedback is really welcomed.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55299, opened 2025-07-06
Add `context:` option to `rate_limit` method

ArthurPV:
### Motivation / Background

A few weeks ago, I tried to use the `rate_limit` method to limit the number of requests on an API.

So I tried to implement the following code:

```ruby
# app/controllers/api_controller.rb
class ApiController < ApplicationController
  rate_limit to: 10, within: 3.minutes
end

# app/controllers/api/messages_controller.rb
class Api::MessagesController < ApiController
  def index; end
end

# app/controllers/api/users_controller.rb
class Api::UsersController < ApiController
  def index; end
end
```

The behavior I expected was that, for example, if someone exceeded the limit on one controller, the same person couldn't access the others. But that's not the behavior I got.

Instead, the behavior I got was that a person could exceed the limit on one controller, then continue to make requests on other controllers that inherit from `ApiController` (reproduction available [here](https://github.com/ArthurPV/reproduction-rails-rate-limit-counted-by-controller-path/blob/main/test/controllers/api_rate_limiting_test.rb)).

Then I went to look at the [source code](https://github.com/rails/rails/blob/b6963db965aa10ef9d63002f3a752cfcb20f0203/actionpack/lib/action_controller/metal/rate_limiting.rb#L62) and it seemed that there was no way to share a rate limit count between several controllers, since by default it was a separate count for each controller.

```ruby
    cache_key = ["rate-limit", controller_path, name, instance_exec(&by)].compact.join(":")
```

NOTE: Already available options such as `name` or `by` cannot help, due to the use of `controller_path`.

That's why I've opened this PR, so that I can add a new option called `context:` to the `rate_limit` method, to solve this problem.

### Details

The `context:` option lets you create a rate limit count shareable between several controllers.

### Additional information

Here's the [link](https://discuss.rubyonrails.org/t/proposal-add-purpose-option-to-rate-limit-method/89453) to the topic I opened on `https://discuss.rubyonrails.org`.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/rate_limiting.rb
+++ b/actionpack/lib/action_controller/metal/rate_limiting.rb
@@ -32,6 +32,9 @@ module ActionController # :nodoc:
       # If you want to use multiple rate limits per controller, you need to give each of
       # them an explicit name via the `name:` option.
       #
+      # If you want to have a shared rate limit count for several controllers at once,
+      # you need to use the `context:` option.
+      #
       # Examples:
       #
       #     class SessionsController < ApplicationController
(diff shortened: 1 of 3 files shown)


mwnciau:
Would like to see this merged. The only change I might suggest is to use `context` rather than `purpose` as that seems more intuitive to me (although may be confused with context in other places)

ArthurPV:
@mwnciau for my part, I have no objection to renaming the `purpose` option to `context`, if that's more intuitive. I'll try to update the PR soon.

ArthurPV:
@mwnciau I updated the PR. Thank you for the suggestion.

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -45,22 +48,30 @@ module ClassMethods
...
+        before_action -> { rate_limiting(to: to, within: within, by: by, with: with, store: store, name: name, context: context) }, **options
       end
     end
 
     private
-      def rate_limiting(to:, within:, by:, with:, store:, name:)
+      def rate_limiting(to:, within:, by:, with:, store:, name:, context:)
         by = instance_exec(&by)
-        cache_key = ["rate-limit", controller_path, name, by].compact.join(":")
+        cache_key = ["rate-limit", context || controller_path, name, by].compact.join(":")
```
MatheusRich: Could this be a default value of the argument, instead of the boolean or here?

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -45,22 +48,30 @@ module ClassMethods
...
+      #       # This controller is limited by a shared rate limit based on `api` context, defined in `APIController`.
       #     end
       #
       #     class SessionsController < ApplicationController
       #       rate_limit to: 3, within: 2.seconds, name: "short-term"
       #       rate_limit to: 10, within: 5.minutes, name: "long-term"
       #     end
-      def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { head :too_many_requests }, store: cache_store, name: nil, **options)
-        before_action -> { rate_limiting(to: to, within: within, by: by, with: with, store: store, name: name) }, **options
+      def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { head :too_many_requests }, store: cache_store, name: nil, context: nil, **options)
```
MatheusRich:
```suggestion
      def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { head :too_many_requests }, store: cache_store, name: nil, context: controller_path, **options)
```

ArthurPV: I applied your suggestion. It seems better than before. Thank you for the suggestion.

actionpack/lib/action_controller/metal/rate_limiting.rb
ArthurPV: https://github.com/rails/rails/pull/55299#discussion_r2252604769

MatheusRich:
I just notice this is essentially the same as

https://github.com/rails/rails/pull/53449

rafaelfranca:
Merged into #53449

PR #55301, opened 2025-07-06
Remove duplicate example in guides/testing

drjayvee:
The section "Testing Enqueued Emails" contained two duplicate examples. This removes one of them.

See https://edgeguides.rubyonrails.org/testing.html#testing-enqueued-emails

### DIFF
--- a/guides/source/testing.md
+++ b/guides/source/testing.md
@@ -2190,25 +2190,6 @@ As with the basic test case, we create the email and store the returned object
 in the `email` variable. The following examples include variations of passing
 arguments and/or parameters.
 
-This example will assert that the email has been enqueued with the correct
-arguments:
-
-```ruby
-require "test_helper"
-
-class UserMailerTest < ActionMailer::TestCase
-  test "invite" do
-    # Create the email and store it for further assertions
-    email = UserMailer.create_invite("me@example.com", "friend@example.com")
-
-    # Test that the email got enqueued with the correct arguments
-    assert_enqueued_email_with UserMailer, :create_invite, args: ["me@example.com", "friend@example.com"] do
-      email.deliver_later
-    end
-  end
-end
-```
-
 This example will assert that a mailer has been enqueued with the correct mailer
 method named arguments by passing a hash of the arguments as `args`:
 


Edouard-chin (commented):
Thanks for the PR.

Both examples are slightly similar but not duplicates. One shows email enqueued with arguments, the other with named arguments.

I think it's worth having both, more so when the doc has already been written 😅 .

drjayvee:
Hey Eduard,

Thanks for the review.

> One shows email enqueued with arguments, the other with named arguments.

Whoops, you're right! I hadn't noticed the difference between positional and named arguments.

The [Action Mailer guide](https://edgeguides.rubyonrails.org/action_mailer_basics.html#edit-the-mailer) uses the `params`-based style exclusively.

As a newcomer to Rails, this is quite confusing. One of these (`args`, named `args`, or `params`) is surely the preferred style.

Maybe it would be good to change the testing docs to use the `params` style first, maybe even using the `welcome_email` action used in the Mailer guide. The testing guide could then continue to describe the args style. Ideally, the Mailer guide would also show the `args` style.

I'd be happy to help, but I don't know which style is actually the most modern or idiomatic.

As for this very PR, I can either close it or update with a different change to the docs. Maybe it's best to open an issue to discuss the best approach? I'd appreciate any guidance.

Edouard-chin:
Sorry for the late response. 

> One of these (args, named args, or params) is surely the preferred style.

The difference between a Parameterized Mailer (the ones using params) and a "vanilla" one is explained in more detail in the API: https://api.rubyonrails.org/classes/ActionMailer/Parameterized.html
So in short, there is no preferred style, it just depends on your use case.

> Maybe it would be good to change the testing docs to use the params style first, maybe even using the welcome_email action used in the Mailer guide

Having the Parameterized example be the first make sense. Feel free to reopen this PR :)

drjayvee:
Hey Edouard,

Thanks for your reply and information. I'm looking forward to diving into the source and updating the docs!

Have a great weekend.

drjayvee:
I've finally taken another look at the docs.

While it's still confusing to read about parameters, `params`, arguments and named arguments, everything does add up on close reading.

Therefore, I'll close this PR (definitively this time 😇).

PR #54402, opened 2025-01-29
Allow ActiveStorage analyzers to be empty

alexandreruban:
Previously, setting `config.active_storage.analyzers` to an empty array would still add a default analyzer to `ActiveStorage.analyzers`.

However, some applications do not require file to be analyzed and prefer to use ActiveStorage without any analyzers.

This commit resolves the issue by ensuring that no default analyzer is added when `config.active_storage.analyzers` is set to an empty value.

Fix #54391

PR #45100, opened 2022-05-14
Load configured Active Storage plugins during boot

skipkayhil:
### Summary

Previously, the parts of Active Storage using ruby-vips, mini_magick, or
image_processing would try to load gems only when used. This strategy works
well because it allows apps that don't need these features to easily
ignore them and not have to depend on gems they don't need.

However, the downsides to this strategy are that it requires loading
code during requests and that it moves potential error messages into request
logs instead of those errors being immediately visible on boot.

This commit addresses these issues by restructing how the Vips and Image
Magick transformers/analyzers are organized so that they will be loaded
(if configured) on boot along with whatever dependencies they need.

For now, if :vips or :mini_magick is specified as the Active Storage
:variant_processor, not having the required gem in the Gemfile will print a
deprecation warning instead of erroring because it is likely many apps
are currently configured this way.

### Other Information

Looking for feedback on the approach, I will add tests/docs/changelog etc. after

Closes #44991 
cc @byroot because of our discussion in the other issue

PR #55303, opened 2025-07-07
Restore Active Storage config to disable variants and analyzers

### LINKED ISSUE #54152: [head] ActiveStorage shows deprecation warning about `image_processing` even without using it
### Steps to reproduce

* Create a new rails app, pointing to main
* Add the following config, set the log-level to `:warn`:
  ```rb
    config.logger = ActiveSupport::Logger.new($stdout)
    config.log_level = :warn
  ```
* Open the console

Caused by https://github.com/rails/rails/pull/45100

### Expected behavior

No deprecation warning

### Actual behavior

```
$ bin/rails c
Generating image variants require the image_processing gem. Please add `gem 'image_processing', '~> 1.2'` to your Gemfile.
Loading development environment (Rails 8.1.0.alpha)
testing(dev)>
```

My application doesn't generate variants, and the default rails template doesn't either, so this seems like a bug.

`load_defaults` sets the variant_processor to `:vips` by default, which is why it tries to load it: https://github.com/rails/rails/blob/aaa111f4c39b441b45d8b1b1408ee0052ed27d9e/railties/lib/rails/application/configuration.rb#L257

### System configuration
**Rails version**: [head](https://github.com/rails/rails/commit/aaa111f4c39b441b45d8b1b1408ee0052ed27d9e)

**Ruby version**: 3.4.1

### LINKED ISSUE #54391: In Active Storage config.active_storage.analyzers = [] is not obeyed
In our config we have:

```
config.active_storage.variant_processor = :mini_magick
config.active_storage.analyzers = []
```

This is because we don't need to do any analysis on the images, but we do want to generate variants.

Since https://github.com/rails/rails/pull/45100 was merged, `config.active_storage.analyzers = []` doesn't work to prevent analysis because [this initializer](https://github.com/rails/rails/pull/45100/files#diff-c83e57b202f71e5fdfb58f3df8bb3c8ad7886b7766fc5e5fefdd870fb04b249cR90-R105) now inserts analyzers if the variant processor is set.

cc @skipkayhil

zzak:
Fixes a couple of mistakes made in e978ef0 (#45100).

* Allow setting `config.active_storage.analyzers = []`
* Allow disabling `config.active_storage.variant_processor` to remove warning
* Ensure original default `ActiveStorage.analyzers` array is not broken (all available analyzers)
* Ensure original default `ActiveStorage.variant_processor` is not broken (`:mini_magick` until `load_defaults(7.0)`)

This change restores the configuration of analyzers back to 8.0, instead of trying to gracefully handle loading those constants if the associated gem is missing.

Due to that config being set at require time, this happens prior to initialization and we don't have access to a logger or deprecator to warn the user.

Since `image_processing` gem transitively depends on `ruby-vips` and `mini_magick`, you only need to install that to remove the warning.

However, if you're not using variant transformers, but still using the default analyzers you can remove the warning by including those gems or setting the analyzers config to an empty array.

We determine the variant transformer based on `config.active_storage.variant_processor` value, or nothing if you passed an invalid value (like `mmmmmini_magick`).

Passing an invalid value, should have the same existing behavior and not impact a working application but in the future we'd like to deprecate it.

Fixes: #54391
Alternative to: #54402

Fixes: #54152
Alternative to: #54166

/cc @alexandreruban @Earlopain @Edouard-chin @skipkayhil

### DIFF
--- a/activestorage/lib/active_storage/engine.rb
+++ b/activestorage/lib/active_storage/engine.rb
@@ -25,7 +25,7 @@ module ActiveStorage
 
     config.active_storage = ActiveSupport::OrderedOptions.new
     config.active_storage.previewers = [ ActiveStorage::Previewer::PopplerPDFPreviewer, ActiveStorage::Previewer::MuPDFPreviewer, ActiveStorage::Previewer::VideoPreviewer ]
-    config.active_storage.analyzers = [ ActiveStorage::Analyzer::VideoAnalyzer, ActiveStorage::Analyzer::AudioAnalyzer ]
+    config.active_storage.analyzers = [ ActiveStorage::Analyzer::ImageAnalyzer::Vips, ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick, ActiveStorage::Analyzer::VideoAnalyzer, ActiveStorage::Analyzer::AudioAnalyzer ]
     config.active_storage.paths = ActiveSupport::OrderedOptions.new
     config.active_storage.queues = ActiveSupport::InheritableOptions.new
     config.active_storage.precompile_assets = true
@@ -88,26 +88,20 @@ module ActiveStorage
 
       config.after_initialize do |app|
         ActiveStorage.logger            = app.config.active_storage.logger || Rails.logger
-        ActiveStorage.variant_processor = app.config.active_storage.variant_processor
+        ActiveStorage.variant_processor = app.config.active_storage.variant_processor || :mini_magick
         ActiveStorage.previewers        = app.config.active_storage.previewers || []
+        ActiveStorage.analyzers         = app.config.active_storage.analyzers || []
 
         begin
-          analyzer, transformer =
+          ActiveStorage.variant_transformer =
             case ActiveStorage.variant_processor
+            when :disabled
+              ActiveStorage::Transformers::NullTransformer
             when :vips
-              [
-                ActiveStorage::Analyzer::ImageAnalyzer::Vips,
-                ActiveStorage::Transformers::Vips
-              ]
+              ActiveStorage::Transformers::Vips
             when :mini_magick
-              [
-                ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick,
-                ActiveStorage::Transformers::ImageMagick
-              ]
+              ActiveStorage::Transformers::ImageMagick
             end
-
-          ActiveStorage.analyzers = [analyzer].compact.concat(app.config.active_storage.analyzers || [])
-          ActiveStorage.variant_transformer = transformer
         rescue LoadError => error
           case error.message
           when /libvips/
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Allow analyzers and variant transformer to be fully configurable
+
+    ```ruby
+    # ActiveStorage.analyzers can be set to an empty array:
+    config.active_storage.analyzers = []
+    # => ActiveStorage.analyzers = []
+
+    # or use custom analyzer:
+    config.active_storage.analyzers = [ CustomAnalyzer ]
+    # => ActiveStorage.analyzers = [ CustomAnalyzer ]
+    ```
+
+    If no configuration is provided, it will use the default analyzers.
+
+    You can also disable variant processor to remove warnings on startup about missing gems.
+
+    ```ruby
+    config.active_storage.variant_processor = :disabled
+    ```
+
+    *zzak*, *Alexandre Ruban*
+
 ## Rails 8.1.0.beta1 (September 04, 2025) ##
 
 *   Remove deprecated `:azure` storage service.
--- a/activestorage/lib/active_storage.rb
+++ b/activestorage/lib/active_storage.rb
@@ -377,6 +377,7 @@ module ActiveStorage
     extend ActiveSupport::Autoload
 
     autoload :Transformer
+    autoload :NullTransformer
     autoload :ImageProcessingTransformer
     autoload :Vips
     autoload :ImageMagick
--- a/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb
+++ b/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 begin
+  require "nokogiri"
+rescue LoadError
+  # Ensure nokogiri is loaded before vips, which also depends on libxml2.
+  # See Nokogiri RFC: Stop exporting symbols:
+  #   https://github.com/sparklemotion/nokogiri/discussions/2746
+end
+
+begin
   gem "ruby-vips"
   require "ruby-vips"
 rescue LoadError => error
--- a/activestorage/lib/active_storage/transformers/null_transformer.rb
+++ b/activestorage/lib/active_storage/transformers/null_transformer.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module ActiveStorage
+  module Transformers
+    class NullTransformer < Transformer # :nodoc:
+      private
+        def process(file, format:)
+          file
+        end
+    end
+  end
+end
--- a/activestorage/test/jobs/transform_job_test.rb
+++ b/activestorage/test/jobs/transform_job_test.rb
@@ -55,4 +55,22 @@ class ActiveStorage::TransformJobTest < ActiveJob::TestCase
       end
     end
   end
+
+  test "null transformer returns original file" do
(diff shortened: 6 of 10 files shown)


activestorage/lib/active_storage/engine.rb
```diff
@@ -88,26 +87,49 @@ class Engine < Rails::Engine # :nodoc:
 
       config.after_initialize do |app|
         ActiveStorage.logger            = app.config.active_storage.logger || Rails.logger
-        ActiveStorage.variant_processor = app.config.active_storage.variant_processor
+        ActiveStorage.variant_processor = app.config.active_storage.variant_processor || :mini_magick
         ActiveStorage.previewers        = app.config.active_storage.previewers || []
+        begin
+          ActiveStorage.analyzers = app.config.active_storage.analyzers ||
```
Earlopain: In the changelog you write:
> You can also disable variant processor to remove warnings on startup about missing gems.

But here the warnings are still shown if the gem is not present. Maybe something along the likes of this makes sense?

```rb
analyzers = []
begin
  analyzers += ActiveStorage::Analyzer::ImageAnalyzer::Vips
rescue LoadError
  # handle it
end
begin
  analyzers += ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick
rescue LoadError
    # handle it
end
if analyzers.empty?
  # warn that both gems are missing, no image analyzing will happen
end
# Side note: These two inform you about missing deps at runtime only. They need the `ffprobe` command, can't exactly eagerly require that.
analyzers.concat(ActiveStorage::Analyzer::VideoAnalyzer, ActiveStorage::Analyzer::AudioAnalyzer)
```

I use vips for analyzing and have the necessary gem in my gemfile, but since it tries to eager-load image-magick as well in the same block, it warns and fails to use vips which I think should just work.

Earlopain (commented):
I pretty much forgot about this already, thanks for checking it out again. `:disabled` is much nicer than relying on `nil` for the processor.

activestorage/lib/active_storage/engine.rb
zzak: The default (as [documented](https://edgeguides.rubyonrails.org/configuring.html#config-active-storage-analyzers)) is to include all analyzers, changing that would require a deprecation cycle.

You are welcome to set analyzers to exclude the one you don't use, and remove the warning.
For example, you can (as [documented](https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blob/Analyzable.html#method-i-analyze)) do something like `Rails.application.config.active_storage.analyzers.delete(ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick)` to remove it (all of this happens _after_initialize_).

If there's something we can put in the changelog that makes this more clear, suggestions are welcome.

Earlopain: If you want to remove this one, you can't reference it via the constant since that triggers autoload, which also tries to load the gem and then potentially raises. You need to completely override it and make sure to include the other default processors if you do so.

The code as documented actually doesn't work anymore. By the time the initializer runs, it is still `nil`. So it has to be wrapped in `after_initialize`, even if `image_processing` is installed. That might be fine (?), no clue if that is part of public contract.

----

I thought about it a bit, and there is actually no possible way that the ImageMagic  analyzer is ever used, unless the user explicitly removes vips from the analyzer list. Since both Vips and ImageMagic say `accept? = true`, it will always choose the one that is first in the list (vips in that case). Previously you could choose which one to use with `variant_processor`, that is now irrelevant for the check if it should be used or not.

I don't think that is the fault of your PR here but rather the previous one.

zzak: > The code as documented actually doesn't work anymore

Can you be more specific?

> There is actually no possible way that the ImageMagic analyzer is ever used

Again, I think I know what you mean but an example might help.

The goal here is definitely to try and stay as close to the expected behavior, without breaking any existing API contracts.

Thanks for your time reviewing this!

Earlopain: Sorry, yes. I can be more specific. Consider this test:

```rb
    test "ActiveStorage.analyzers in initializer" do
      app_file "config/initializers/add_analyzer.rb", <<-RUBY
        # Appending just a string for demonstration purposes
        Rails.application.config.active_storage.analyzers.append("foo") 
      RUBY

      app "development"

      assert ActiveStorage.analyzers.include?("foo"), ActiveStorage.analyzers.to_s
    end
```

It passes on current main but fails on this PR because `Rails.application.config.active_storage.analyzers` is nil by the time initializers run. I'm don't actually know how to add/delete from that array with this PR, since it is always left nil unless explicitly configured to be something else.

-----

As for the other point, I'm not so sure on how to demonstrate it. Which analyzer is used is invisible to the application. The common implementation is here: https://github.com/rails/rails/pull/45100/files#diff-86760508d8c4d325dd2a361cebb891b3b347d2dd890d3cd1fb641cb3c0fda322R20-R22
And it also used to do an additional check here and here (which is also documented in the config guide):
https://github.com/rails/rails/pull/45100/files#diff-bbfba04c81dcdc406986d3697ddc362cabcbd130b8888ed7752d074dac1ba8c6R12
https://github.com/rails/rails/pull/45100/files#diff-d44e964f5b1a9ae851af9f6f90df352432b6862b7442295cfb451c770c1457d2R13

Because the code that chooses the analyzer looks like this:
https://github.com/rails/rails/blob/0bd254ebf835e880d485897d23de667007caa47b/activestorage/app/models/active_storage/blob/analyzable.rb#L60

it will now always choose the first image analyzer in the list (when the file is an image that is). Hopefully that explanation makes sense.

zzak: Thanks for the test case! I understood what you mean.

I've reverted the behavior for handling analyzers back to original, Rails 8.0 method. So there shouldn't be any weirdness.

Could you try it?

Earlopain: Thanks, I believe that works fine. I don't see any trouble with that anymore.

It still always uses vips to analyze but I'm not sure how impactful that actually is (and it is also out of scope for this PR I think). I'll just make a separate issue for that later.

julianrubisch (commented):
I can confirm that this PR restores the ability to prepend custom analyzers to `ActiveStorage.analyzers` 🙌

activestorage/lib/active_storage/engine.rb
```diff
@@ -88,26 +88,26 @@ class Engine < Rails::Engine # :nodoc:
...
             when :mini_magick
-              [
-                ActiveStorage::Analyzer::ImageAnalyzer::ImageMagick,
-                ActiveStorage::Transformers::ImageMagick
-              ]
+              ActiveStorage::Transformers::ImageMagick
+            else
+              ActiveStorage.deprecator.warn(<<~WARN.squish)
+                ActiveStorage.variant_processor must be set to :vips, :mini_magick, or :disabled.
+                Passing #{ActiveStorage.variant_processor.inspect} is deprecated and will raise an exception in Rails 8.1.
```
zzak: I guess this part can wait, or we can split the PR. 🤔 
The goal was to help when people fumble the name, e.g. input `variant_processor = :minimagick` instead of `:mini_magick`.

zzak: Ok, I'd still like to get this in 8.1 so removed the deprecation we can do later.

Earlopain:
Something isn't quite right I think. I generated a new rails app with `--main` and opening the console fails with this error:

```sh
$ bin/rails c
/home/user/code/rails/railties/lib/rails/application/configuration.rb:178:in 'Rails::Application::Configuration#load_defaults': undefined method 'analysis=' for nil (NoMethodError)

            active_storage.queues.analysis = :active_storage_analysis
                                 ^^^^^^^^^^^
```

bisect lead me back to this. Can someone confirm?

Earlopain:
Oh, this happens when libvips is not installed. I forgot I uninstalled this to test something else.

jeromedalbert:
FYI my daily-run automated test broke 2 days ago with the same error as @Earlopain. This is an end-to-end test that generates a Rails app with `--main` followed by `rails generate rspec:install`, which errors out like so:

```
NoMethodError: undefined method 'analysis=' for nil (NoMethodError)

            active_storage.queues.analysis = :active_storage_analysis
```


My GH action uses the `ubuntu-latest` image. I can try @Earlopain's suggestion of adding `libvips` to my CI setup, but this may still be an issue regardless.

I.e. maybe users expect `bin/rails some_command` to work after running `rails new`. Meaning they expect `rails new` not to generate an app that appears to be broken.

zzak:
Thank you both for checking.

I can reproduce after uninstalling libvips with this:

```
bundle exec railties/exe/rails new ~/code/apps/IssueCheck_55303 --dev
```

Opening a PR.

PR #55302, opened 2025-07-06
Fix path for Product::Notifications concern in guide

dpankratz1:
### Motivation / Background

This Pull Request has been created because the current Rails guide suggests placing the Product::Notifications concern in app/models/product/notifications.rb, which can be misleading.

When creating a new Rails app with `rails new`, Rails generates an `app/models/concerns/` directory, which is the conventional location for model concerns using ActiveSupport::Concern. Updating the file path to `app/models/concerns/product/notifications.rb` follows this convention and avoids potential confusion.

This change keeps the namespacing as `Product::Notifications`, which remains correct and consistent with the example code.

### Detail

This Pull Request changes the file path in the documentation from:
```
app/models/product/notifications.rb
```
to:

```
app/models/concerns/product/notifications.rb
```
to match the folder structure generated by `rails new` and align with Rails' conventions for organizing concerns.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55304, opened 2025-07-07
Fix incorrect module include in Getting Started guide [ci skip]

Ikass:
The guide defines `module Product::Notifications` but shows `include Notifications`. This should be `include Product::Notifications` to match the module's namespaced definition.

Fixes getting_started.md line 2339.

### Motivation / Background

This Pull Request has been created because Section 16.4 of the Getting Started guide defines `module Product::Notifications` but then shows `include Notifications`, which would cause a `NameError`.

### Detail

This Pull Request fixes line 2339 in the Getting Started guide (Section 16.4 "Extracting a Concern"). The code example incorrectly shows `include Notifications` when it should be `include Product::Notifications` to match the module definition shown earlier.

### Additional information

Alternative would be to place the module in the standard `app/models/concerns/` directory and define as `module Notifications`.

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2336,7 +2336,7 @@ Notifications module.
 
 ```ruby#2
 class Product < ApplicationRecord
-  include Notifications
+  include Product::Notifications
 
   has_one_attached :featured_image
   has_rich_text :description


dpankratz1:
Hi @Ikass 👋 
when you're inside the Product class, the constant Notifications is automatically resolved to Product::Notifications. I've tried it and it works fine.

Checkout [Zeitwerk's Readme](https://github.com/fxn/zeitwerk?tab=readme-ov-file#explicit-namespaces): 
`The classes and modules from the namespace are already available in the body of the class or module defining it...`

p8:
Hmm, I can see this is confusing.
The concern is also tightly coupled to the product because it depends on `Product#inventory_count_previously_was?`, `Product#inventory_count` and `ProductMailer`, so it's not that reusable as is.

Ikass:
Hi @tsuruby Did not know that.

For a novice Rails dev like myself, this discrepancy was quite puzzling. If it is the accepted best practice - the Rails way of doing things to take advantage of 'Rails Magic' behind the scenes - then it would be beneficial to mention it. Perhaps in a Note?

Especially since a couple of paragraphs later the guide references the `Product::Notifications` concern again: "Instead, we can use Concerns to extract each feature out into a self-contained module like Product::Notifications which contains all the functionality for handling subscribers and how notifications are sent."

I realize it makes sense to make a Product class specific Concern, since the functionality is unlikely to me reused in other classes. But it was a big leap for me as a novice going through the tutorial and stopped me.

Perhaps we can add a Note or a paragraphs with something like - and please correct me if I am wrong in any of this -  "Normally we would put Concerns in the `app/models/concerns/ ` directory, so they can be used by all Models. But the notification functionality will only be used by the Product model, so let's put it in it's own directly to keep the codebase more organized."

dpankratz1:
I totally agree, it was also confusing to me when I read it. 
I was actually expecting the concern to be placed in `app/models/concerns/product/`
Then the namespacing would also be correct as it is in the code example.

I also opened a PR for that:
https://github.com/rails/rails/pull/55302

To me this looks confusing:
```
-models/
  --concerns/ <- empty folder ?
  --product.rb    <--- here I asked myself why the product model is not inside the product folder
  --product/
     --notifications <- concern
```

I would expect:
```
-models/
  --concerns/ 
      --product/
         --notifications    <--- concern for product
  --product.rb
```

guides/source/getting_started.md
```diff
@@ -2336,7 +2336,7 @@ Notifications module.
 
 ```ruby#2
-  include Notifications
+  include Product::Notifications
```
byroot, line 2339: Since it's inside `Product` the `Product::` is implicit, so it should work even without your change.

p8: Maybe just a comment for clarity?
```suggestion
  include Notifications # inside Product we can remove Product::
```

eileencodes: I don't think the comment is even necessary, this is how you write Ruby 🤷🏼‍♀️

MatheusRich: I'm torn about the comment. One one hand, it's not Rails' problem to teach people basic Ruby. On the other hand, the reality is that many folks pick up Ruby by using Rails, so our beginner tutorials are also teaching Ruby basics whether we want it or not.

eileencodes: I actually find the comment _more_ confusing. Unless I read this issue I wouldn't know what was meant by "we can remove Product::" because I wouldn't word it that way if I was trying to say namespacing isn't necessary when within a namespace.

MatheusRich: Sure. Is the wording the only problem, then?

```suggestion
  include Notifications # it's not necessary to prefix Product:: when inside that namespace
```

p8: Maybe it would be clearer for beginners if the `Notifications` module was defined with nesting:
```
  module Notifications
    ...
  end
end
```

Another option could be linking to the documentation of `Module`, but that currenlty doesn't explain nested "namespaces" 😞 .

p8: There is some [Zeitwerk magic ](https://github.com/fxn/zeitwerk?tab=readme-ov-file#explicit-namespaces)going on here.

For example the following plain Ruby:
```ruby
# hotel.rb
require_relative 'hotel/pricing'
class Hotel
  include Pricing
end

puts Hotel.new.price
```

```ruby
# hotel/pricing.rb
# With Zeitwerk Hotel would already available here
module Hotel::Pricing
  def price
    "$100.0"
  end
end
```

Running this in plain Ruby would raise an error:
```bash
$ ruby hotel.rb
/private/tmp/hotel/pricing.rb:1:in '<top (required)>': uninitialized constant Hotel (NameError)

module Hotel::Pricing
       ^^^^^
	from hotel.rb:1:in 'Kernel#require_relative'
	from hotel.rb:1:in '<main>'
```

manuel1618:
Just my 2ct because my PR got mentioned. Context: Beginner on rails / ruby here. 

I am fine putting the notifications inside a `product` folder and referencing it as `Notifications` inside the product.rb model. No reuse of the concern intended. It still breaks up a potentially big product.rb into smaller chunks. 

But then there is the sentence about code reuse of concerns -> immediately I asked myself how would that work across different models. That's what the `concerns` folder is for. So you could choose and see with the folder structure which concerns are potentially used by multiple models and which are just bound to one model.

guides/source/getting_started.md
rafaelfranca: Instead of a comment, I'd just explain it after the class definition.

PR #55308, opened 2025-07-08
Allow to log/notify/raise on path relative redirects

TobiasBales:
### Motivation / Background

The current situation, when either requiring redirects to other host (a lot of authentication flows) or for applications that, so far, have not enabled `raise_on_open_redirect`,
can lead to unexpected redirects since it allows redirects to different domains while calling `redirect_to` with a relative url.

E.g. with the app running on https://example.com, calling `redirect_to("foo")`, will redirect to "https://example.comfoo" which can potentially be exploited. `redirect_to("@dell.com") is even worse since it redirects to "https://example.com@dell.com" with `example.com`being interpreted as the user and dell.com as the domain.

Allowing redirects to other hosts is very common in the authentication real, especially when using OAuth and guarding more applications from patterns like this makes sense (at least to me).

This adds the option to log, notify or raise with the default for existing applications being log and the default for new applications being raise. Notify is added to allow existing applications to gather statistics about where this happens to analyze call sites individually, working on them and then being able to confidently turn the setting to raise.

### Detail

This introduces an config parameter `action_on_path_relative_redirects` which can be set to `:log` (default), `:notify` or `:raise` to allow migrating from the default (`:log`) to the safe option (`:raise`) without causing production issues.

### Checklist

### DIFF
--- a/actionpack/test/controller/redirect_test.rb
+++ b/actionpack/test/controller/redirect_test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "abstract_unit"
+require "active_support/log_subscriber/test_helper"
 
 class Workshop
   extend ActiveModel::Naming
@@ -152,6 +153,14 @@ class RedirectController < ActionController::Base
 
   def redirect_to_url_with_network_path_reference
     redirect_to "//www.rubyonrails.org/"
+  end
+
+  def redirect_to_path_relative_url
+    redirect_to "example.com"
+  end
+
+  def redirect_to_path_relative_url_starting_with_an_at
+    redirect_to "@example.com"
   end
 
   def redirect_to_existing_record
@@ -333,6 +342,18 @@ class RedirectTest < ActionController::TestCase
     get :redirect_to_url_with_complex_scheme
     assert_response :redirect
     assert_equal "x-test+scheme.complex:redirect", redirect_to_url
+  end
+
+  def test_redirect_to_path_relative_url
+    get :redirect_to_path_relative_url
+    assert_response :redirect
+    assert_equal "http://test.hostexample.com", redirect_to_url
+  end
+
+  def test_redirect_to_url_with_path_relative_url_starting_with_an_at
+    get :redirect_to_path_relative_url_starting_with_an_at
+    assert_response :redirect
+    assert_equal "http://test.host@example.com", redirect_to_url
   end
 
   def test_redirect_to_url_with_network_path_reference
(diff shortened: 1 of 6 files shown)


actionpack/test/controller/redirect_test.rb
```diff
@@ -335,6 +344,18 @@ def test_redirect_to_url_with_complex_scheme
     assert_equal "x-test+scheme.complex:redirect", redirect_to_url
   end
 
+  def test_redirect_to_url_with_relative_url_without_leading_slash
+    get :redirect_to_relative_url_without_leading_slash
+    assert_response :redirect
+    assert_equal "http://test.hostexmaple.com", redirect_to_url
```
TobiasBales: this is just documenting the unexpected status quo - not sure if we really should keep them

actionpack/test/controller/redirect_test.rb
```diff
@@ -335,6 +344,18 @@ def test_redirect_to_url_with_complex_scheme
...
+  end
+
+  def test_redirect_to_url_with_relative_url_starting_with_an_at
+    get :redirect_to_relative_url_starting_with_an_at
+    assert_response :redirect
+    assert_equal "http://test.host@example.com", redirect_to_url
```
TobiasBales, line 356: ditto

actionpack/test/controller/redirect_test.rb
```diff
@@ -620,6 +641,180 @@ def test_redirect_to_external_with_rescue
...
+    get :relative_url_redirect_with_status
+    assert_response :redirect
+    assert_equal "http://test.host/things/stuff", redirect_to_url
+    assert_empty events
+  ensure
+    ActiveSupport::Notifications.unsubscribe("unsafe_redirect.action_controller")
+    ActionController::Base.relative_redirects_without_leading_slash = old_config
+  end
+
+  def test_redirect_to_absolute_url_does_not_raise
```
TobiasBales, line 799: I am not sure we want the counter tests for log AND raise but I added them for now to make sure/prove that it does not break flows we expect to work. Happy to remove/update them though

tenderlove (commented):
This makes sense to me.  One question about the option name though:

`action_on_relative_redirects_without_leading_slash`

The definition of a relative redirect is one that doesn't have a leading slash, right?  "without_leading_slash" seems redundant, but maybe I'm missing something?

Regardless, I think this is a good feature.

TobiasBales:
> This makes sense to me. One question about the option name though:
> 
> `action_on_relative_redirects_without_leading_slash`
> 
> The definition of a relative redirect is one that doesn't have a leading slash, right? "without_leading_slash" seems redundant, but maybe I'm missing something?
> 
> Regardless, I think this is a good feature.

I might be wrong but my understanding (and google/gemini seems to back it up) are:
redirect_to("/foo"), redirect_to("foo") -> relative
redirect_to("///www.example.com/foo") -> protocol relative
redirect_to("https://www.example.com/foo") -> absolute

tenderlove:
> redirect_to("/foo")

I think this case is a "Relative URL, absolute path" example.

>  redirect_to("foo") -> relative

Where this one is relative URL, relative path. [Taken from wikipedia](https://en.wikipedia.org/wiki/HTTP_location).

Maybe `action_on_path_relative_redirect` would be more clear? wdyt?

TobiasBales:
> > redirect_to("/foo")
> 
> I think this case is a "Relative URL, absolute path" example.
> 
> > redirect_to("foo") -> relative
> 
> Where this one is relative URL, relative path. [Taken from wikipedia](https://en.wikipedia.org/wiki/HTTP_location).
> 
> Maybe `action_on_path_relative_redirect` would be more clear? wdyt?

today I learned - and yes, that is absolutely fine and more clear, will update

actionpack/test/controller/redirect_test.rb
tenderlove: I think it's fine.  We can always delete them later.

tenderlove (approved):

PR #26811, opened 2016-10-18
Adds proc support for :allow_blank and :allow_nil

gucki:
Allows to pass a proc to :allow_blank and :allow_nil for validators.

PR #55309, opened 2025-07-08
Add callable support to allow_nil and allow_blank

rjlynch:
Adds the ability to pass a callable to `allow_nil` and `allow_blank`. This allows handling scenarios where you want to conditionally allow an empty value to be saved but ensure other values are validated.

### Motivation / Background

We ran into a limitation with `allow_nil` while implementing a form as part of a multi step wizard.

We wanted to conditionally allow a `nil` value for an attribute but otherwise ensure the value was included in a collection, eg

```ruby
validates(
  :course_id,
  inclusion: {
    in: -> (form) { form.permitted_courses.map(&:id) }
  },
  allow_nil: :saving_progress?
)
```

While there are work arounds for this (duplicating validation or conditionally appending a nil to the collection), they're noisy, and reduce readability, eg
```ruby
validates(
  :course_id,
  inclusion: {
    in: -> form { form.permitted_courses.map(&:id) }
  },
  unless: :saving_progress?
)
validates(
  :course_id,
  inclusion: {
    in: -> (form) { form.permitted_courses.map(&:id) << nil } # note we use this list for the select and don't want nil as an option in the view
  },
  if: :saving_progress?
)
```

We created this pull request as we were surprised this behaviour wasn't already supported, usually with Rails everything just works as you'd expect!

### Detail
This Pull Request changes the `ActiveModel::Validator#validate` method to allow passing a callable (responds to `to_proc`) as an option for `allow_nil` and `allow_blank`

### Additional information
I saw this functionality was [previous proposed](https://github.com/rails/rails/pull/26811/) in 2016 but shot down due to being achievable with the `if` and `unless` options. Just close this PR if you still think this option isn't necessary, I appreciate you're very busy so no hard feelings!

Our [work around](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3918/commits/5963d89bda76bdfdd9964634f4a02e576ab32cb5) was to implement a custom inclusion validator with callable support for allow nil, but that approach isn't ideal as  `allow_nil: :some_callable` doesn't work as expected for other validators.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/validator.rb
+++ b/activemodel/lib/active_model/validator.rb
@@ -150,7 +150,7 @@ module ActiveModel
     def validate(record)
       attributes.each do |attribute|
         value = record.read_attribute_for_validation(attribute)
-        next if (value.nil? && options[:allow_nil]) || (value.blank? && options[:allow_blank])
+        next if (value.nil? && allow_nil?(record)) || (value.blank? && allow_blank?(record))
         value = prepare_value_for_validation(value, record, attribute)
         validate_each(record, attribute, value)
       end
@@ -171,6 +171,22 @@ module ActiveModel
     private
       def prepare_value_for_validation(value, record, attr_name)
         value
+      end
+
+      def allow_nil?(record)
+        if options[:allow_nil].respond_to?(:to_proc)
+          !!options[:allow_nil].to_proc.call(record)
+        else
+          !!options[:allow_nil]
+        end
+      end
+
+      def allow_blank?(record)
+        if options[:allow_blank].respond_to?(:to_proc)
+          !!options[:allow_blank].to_proc.call(record)
+        else
+          !!options[:allow_blank]
+        end
       end
   end
 
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,19 @@
+*   Allow passing method name or proc to `allow_nil` and `allow_blank`
+
+    ```ruby
+    class EnrollmentForm
+      include ActiveModel::Validations
+
+      attr_accessor :course
+
+      validates :course,
+                inclusion: { in: :open_courses },
+                allow_nil: :saving_progress?
+    end
+    ```
+
+    *Richard Lynch*
+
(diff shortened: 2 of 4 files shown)


activemodel/lib/active_model/validator.rb
```diff
@@ -172,6 +172,22 @@ def check_validity!
       def prepare_value_for_validation(value, record, attr_name)
         value
       end
+
+      def allow_nil?(record)
+        if options[:allow_nil].respond_to?(:to_proc)
+          !!options[:allow_nil].to_proc.call(record)
```
MatheusRich, line 178: We didn't use !! before. Should we add it now?

MatheusRich (commented):
LGTM

activemodel/lib/active_model/validator.rb
rjlynch: I think as it's a predicate method we want to make sure it always returns a boolean. Wont change the existing behaviour and avoids any surprises down the line

MatheusRich: Not all predicate methods return booleans. It is just expected to behave as truthy/falsey (return objects and nil, for example is valid). From the [Ruby docs](https://docs.ruby-lang.org/en/3.4/syntax/methods_rdoc.html#:~:text=Methods%20that%20end%20with%20a%20question%20mark%20by%20convention%20return%20boolean%2C%20but%20they%20may%20not%20always%20return%20just%20true%20or%20false.%20Often%2C%20they%20will%20return%20an%20object%20to%20indicate%20a%20true%20value%20(or%20%E2%80%9Ctruthy%E2%80%9D%20value).):

> Methods that end with a question mark by convention return boolean, but they may not always return just true or false. Often, they will return an object to indicate a true value (or “truthy” value).

rjlynch: Happy to remove the `!!` if you prefer 👍 , though I've been surprised by a predicate returning `nil` before (`Model.limit(n).loaded?` surprised me!)

guilleiguaran:
@rjlynch, sorry this got outdated once again. Can you rebase and fix the conflict?

rjlynch:
> @rjlynch, sorry this got outdated once again. Can you rebase and fix the conflict?

No problem, done 👍

PR #55313, opened 2025-07-09
Include current character length in index and table name length error messages

joshuay03:
Small QoL improvement.

Including both the current character length and the max character length makes it easier to immediately tell how many characters over the current index/table name is, without having to figure it out by pasting it into an editor, for example.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/core_ext/string/access"
+require "active_support/core_ext/string/filters"
 require "openssl"
 
 module ActiveRecord
@@ -1871,13 +1872,19 @@ module ActiveRecord
 
         def validate_index_length!(table_name, new_name, internal = false)
           if new_name.length > index_name_length
-            raise ArgumentError, "Index name '#{new_name}' on table '#{table_name}' is too long; the limit is #{index_name_length} characters"
+            raise ArgumentError, <<~MSG.squish
+              Index name '#{new_name}' on table '#{table_name}' is too long (#{new_name.length} characters); the limit
+              is #{index_name_length} characters
+            MSG
           end
         end
 
         def validate_table_length!(table_name)
           if table_name.length > table_name_length
-            raise ArgumentError, "Table name '#{table_name}' is too long; the limit is #{table_name_length} characters"
+            raise ArgumentError, <<~MSG.squish
+              Table name '#{table_name}' is too long (#{table_name.length} characters); the limit is
+              #{table_name_length} characters
+            MSG
           end
         end
 
--- a/activerecord/test/cases/migration/index_test.rb
+++ b/activerecord/test/cases/migration/index_test.rb
@@ -53,7 +53,7 @@ module ActiveRecord
         e = assert_raises(ArgumentError) {
           connection.rename_index(table_name, "old_idx", too_long_index_name)
         }
-        assert_match(/too long; the limit is #{connection.index_name_length} characters/, e.message)
+        assert_match(/too long \(#{too_long_index_name.length} characters\); the limit is #{connection.index_name_length} characters/, e.message)
 
         assert connection.index_name_exists?(table_name, "old_idx")
       end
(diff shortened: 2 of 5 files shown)


alexcameron89 (approved):
I like this, +1 to highly contextual error messages.

activerecord/test/cases/migration/index_test.rb
```diff
@@ -53,7 +53,7 @@ def test_rename_index_too_long
         e = assert_raises(ArgumentError) {
           connection.rename_index(table_name, "old_idx", too_long_index_name)
         }
-        assert_match(/too long; the limit is #{connection.index_name_length} characters/, e.message)
+        assert_match(/too long \(#{connection.index_name_length + 1} characters\); the limit is #{connection.index_name_length} characters/, e.message)
```
yahonda:
```suggestion
        assert_match(/too long \(#{too_long_index_name.length} characters\); the limit is #{connection.index_name_length} characters/, e.message)
```

activerecord/test/cases/migration/index_test.rb
```diff
@@ -75,7 +75,7 @@ def test_add_index_does_not_accept_too_long_index_names
         e = assert_raises(ArgumentError) {
           connection.add_index(table_name, "foo", name: too_long_index_name)
         }
-        assert_match(/too long; the limit is #{connection.index_name_length} characters/, e.message)
+        assert_match(/too long \(#{connection.index_name_length + 1} characters\); the limit is #{connection.index_name_length} characters/, e.message)
```
yahonda:
```suggestion
        assert_match(/too long \(#{too_long_index_name.length} characters\); the limit is #{connection.index_name_length} characters/, e.message)
```

activerecord/test/cases/migration/rename_table_test.rb
```diff
@@ -56,7 +56,7 @@ def test_rename_table_raises_for_long_table_names
         error = assert_raises(ArgumentError) do
           connection.rename_table :test_models, long_name
         end
-        assert_equal "Table name '#{long_name}' is too long; the limit is #{name_limit} characters", error.message
+        assert_equal "Table name '#{long_name}' is too long \(#{name_limit + 1} characters\); the limit is #{name_limit} characters", error.message
```
yahonda:
```suggestion
        assert_equal "Table name '#{long_name}' is too long \(#{long_name.length} characters\); the limit is #{name_limit} characters", error.message
```

activerecord/test/cases/migration_test.rb
```diff
@@ -194,7 +194,7 @@ def test_create_table_raises_for_long_table_names
     error = assert_raises(ArgumentError) do
       connection.create_table(long_name)
     end
-    assert_equal "Table name '#{long_name}' is too long; the limit is #{name_limit} characters", error.message
+    assert_equal "Table name '#{long_name}' is too long \(#{name_limit + 1} characters\); the limit is #{name_limit} characters", error.message
```
yahonda:
```suggestion
    assert_equal "Table name '#{long_name}' is too long \(#{long_name.length} characters\); the limit is #{name_limit} characters", error.message
```

yahonda (commented):
Would it make sense to update the test to reflect the actual length of the table or index name, like this?

activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
```diff
@@ -1871,13 +1872,18 @@ def check_constraint_for!(table_name, expression: nil, **options)
...
+              Index name '#{new_name}' on table '#{table_name}' is too long (#{new_name.length} characters); the limit
+              is #{index_name_length} characters
+            MSG
           end
         end
 
         def validate_table_length!(table_name)
           if table_name.length > table_name_length
-            raise ArgumentError, "Table name '#{table_name}' is too long; the limit is #{table_name_length} characters"
+            raise ArgumentError, <<~MSG.squish
```
yahonda, line 1884: I don’t think squish is needed here, since the message is already one line and <<~ should be enough by itself. That said, if you want to keep it consistent with validate_index_length! or anticipate that the message might grow longer in the future, leaving it as is would be totally fine.
```suggestion
            raise ArgumentError, <<~MSG
```

yahonda:
Would you mind squashing the commits if the suggestion looks good? I added a few commits while reviewing.

activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
joshuay03: Agree that it's not needed. I just wasn't paying attention and updated both together.

joshuay03: Actually, I think I meant to make it multi-line and forgot. I personally use 120 chars as a wrap guide.

joshuay03:
> Would you mind squashing the commits if the suggestion looks good? I added a few commits while reviewing.

Done 👍🏽 Thank you for the review

Issue #53464, opened 2024-10-27
`redirect_to` host verification fails with URLs that have a non-ASCII querystring param


ghiculescu:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails", branch: "main"
end

require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  config.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  routes.draw do
    get "/good" => "test#good"
    get "/bad" => "test#bad"
  end
end

class TestController < ActionController::Base
  include Rails.application.routes.url_helpers

  def good
    redirect_to "/query?name=alex", allow_other_host: false
  end
  def bad
    redirect_to "/query?name=fernández", allow_other_host: false
  end
end

require "minitest/autorun"
require "rack/test"

class BugTest < ActiveSupport::TestCase
  include Rack::Test::Methods

  def test_returns_success
    get "/good"
    assert last_response.redirect?
    get "/bad"
    assert last_response.redirect?
  end

  private
    def app
      Rails.application
    end
end

```

### Expected behavior
I expect both assertions to pass.

### Actual behavior
It raises:

```
ActionController::Redirecting::UnsafeRedirectError (Unsafe redirect to "http://example.org/query?name=fernández", pass allow_other_host: true to redirect anyway.):
```

The root cause is here: https://github.com/rails/rails/blob/e6429269fd5a8a7d728557f4e7d7f82c0ad1478c/actionpack/lib/action_controller/metal/redirecting.rb#L231

`URI(url.to_s)` raises with error `"URI must be ascii only"`. I guess this is because https://github.com/ruby/uri doesn't handle non-ASCII characters in the URI. Maybe that's valid in general but in this specific case, I think it should not break this, as long as the host is valid.

### System configuration
**Rails version**: main

**Ruby version**: 3.3

p8:
I can see the error is confusing but shouldn't this be fixed by encoding the user supplied data?
```ruby
name = URI.encode_uri_component "fernández"
redirect_to "/query?name=#{name}", allow_other_host: false
```

ghiculescu:
Hmmm, maybe. If that's what's expected, the error message could be less confusing.

PR #55319, opened 2025-07-10
Fix errors when querystring keys have invalid encoding 

### LINKED ISSUE #52114: Rails throws an error when sending invalid UTF-8 query param
### Steps to reproduce

Start a rails application and go to http://localhost:3000?%ADf

### Expected behavior

It should respond with `ActionController::BadRequest` This is the case when you send  a request like http://localhost:3000?test=%ADf

### Actual behavior

It throws an `ArgumentError` because the `parameter_filter` is crashing on the invalid UTF-8 char.
 
Screenshot 2024-06-13 at 11 01 53

### Solution

The fix seems easy enough. The `check_param_encoding` method on https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/request/utils.rb#L36 only checks the param **values** and not the **keys**.

Changing L36 from `params.each_value { |value| check_param_encoding(value) }` to `params.each { |value| check_param_encoding(value) }` fixes the problem for me.

### System configuration
**Rails version**: 7.1.3.4

**Ruby version**: 3.3.1

**Rack version**: 3.1.12

cmitz:
### Motivation / Background

We encountered a lot of errors when bots were firing malformed requests at our servers. In query strings, when bad **values** are supplied, these get checked and the request gets killed. However, when bad **keys** are supplied, there will be a 500 error. This pull request fixes #52114 (and replaces #52116)

A coleague of mine already attempted a fix in the other repo, but the tests are still failing (didn't fix the issue) so I took another shot at it. Shout out to Chris for pinpointing where in the code I had to look!

### Detail

This Pull Request add a little encoding check in the parsing of query strings. It is a small change but resolves our 500 errors.

### Additional information

Benchmark

```ruby
require "benchmark/ips"

require "action_dispatch"
require "active_support/all"

require_relative "param_builder"
require_relative "param_builder_original"

QUERY_STRING = "foo=bar"
QUERY_STRING_INVALID = "foo=%81E"

def check_param_encoding
  request_params = ActionDispatch::ParamBuilder.from_query_string(QUERY_STRING, encoding_template: nil)
  request_params.inspect
end

def check_param_encoding_original
  request_params = ActionDispatch::ParamBuilderOriginal.from_query_string(QUERY_STRING, encoding_template: nil)
  request_params.inspect
end


Benchmark.ips do |x|
  x.report("new") { check_param_encoding }
  x.report("original") { check_param_encoding_original }

  x.compare!
end
```

Result
```
$ bundle exec ruby benchmark.rb
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin24]
                 new    40.767k i/100ms
            original    41.113k i/100ms
                 new    408.540k (± 1.5%) i/s    (2.45 μs/i) -      2.079M in   5.090267s
            original    408.498k (± 3.2%) i/s    (2.45 μs/i) -      2.056M in   5.037672s

Comparison:
                 new:   408540.0 i/s
            original:   408498.2 i/s - same-ish: difference falls within error
```

and because I did not believe it, a second time the same benchmark:
```
$ bundle exec ruby benchmark.rb
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin24]
                 new    41.459k i/100ms
            original    41.624k i/100ms
                 new    411.373k (± 1.6%) i/s    (2.43 μs/i) -      2.073M in   5.040512s
            original    414.142k (± 1.3%) i/s    (2.41 μs/i) -      2.081M in   5.026239s

Comparison:
            original:   414142.1 i/s
                 new:   411373.2 i/s - same-ish: difference falls within error
```

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Always check query string keys for valid encoding just like values are checked.
+
+    *Casper Smits*
+
 *   Always return empty body for HEAD requests in `PublicExceptions` and
     `DebugExceptions`.
 
--- a/actionpack/lib/action_dispatch/http/param_builder.rb
+++ b/actionpack/lib/action_dispatch/http/param_builder.rb
@@ -111,6 +111,10 @@ module ActionDispatch
 
         return if k.empty?
 
+        unless k.valid_encoding?
+          raise InvalidParameterError, "Invalid encoding for parameter: #{k}"
+        end
+
         if depth == 0 && String === v
           # We have to wait until we've found the top part of the name,
           # because that's what the encoding template is configured with
--- a/actionpack/test/dispatch/request_test.rb
+++ b/actionpack/test/dispatch/request_test.rb
@@ -1121,6 +1121,11 @@ class RequestParameters < BaseRequestTest
     assert_raises(ActionController::BadRequest) { request.parameters }
   end
 
+  test "parameters key containing an invalid UTF8 character" do
+    request = stub_request("QUERY_STRING" => "%81E=bar")
+    assert_raises(ActionController::BadRequest) { request.parameters }
+  end
+
   test "parameters containing a deeply nested invalid UTF8 character" do
     request = stub_request("QUERY_STRING" => "foo[bar]=%81E")
     assert_raises(ActionController::BadRequest) { request.parameters }


zzak:
cc @ghiculescu I think this is kind of similar to #53464 you reported, well I am assuming that host verification error is not resolved but handling the request with invalid query string will now fail expectedly?

PR #55327, opened 2025-07-13
Include `author_addresses` fixtures when using `authors` fixtures

fatkodima:
This fixes an existing (https://buildkite.com/rails/rails/builds/119881#019804bb-8c58-4be4-83fb-38b3b9bea422) and potentially future flaky tests.

The flaky test case started appearing after merging https://github.com/rails/rails/pull/55285.

To reproduce, run
```
ARCONN=postgresql TESTOPTS="-v" bin/test --seed 20799 test/cases/fixtures_test.rb test/cases/associations/has_one_through_associations_test.rb -n "/^(?:DeprecatedHasOneThroughAssociationsTest#(?:test_the_source_association_is_deprecated)|FixturesWithForeignKeyViolationsTest#(?:test_does_not_raise_if_no_fk_violations|test_raises_fk_violations))$/"
```

### DIFF
--- a/activerecord/test/cases/associations/has_many_through_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_through_associations_test.rb
@@ -1728,7 +1728,7 @@ end
 class DeprecatedHasManyThroughAssociationsTest < ActiveRecord::TestCase
   include DeprecatedAssociationsTestHelpers
 
-  fixtures :authors
+  fixtures :authors, :author_addresses
 
   setup do
     @model = DATS::Author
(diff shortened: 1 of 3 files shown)


fxn:
@fatkodima did you investigate what made it pass or fail? what made it flaky?

fatkodima:
Inserting one records without others produced bad foreign keys records. Do not remember other details.

fxn:
OK. Makes me think, because those tests do not access the association, the suite passed, and I don't think one should by hand load all transitive dependencies, that would very impractical.

Perhaps the patch is correct and I do not quite see it, but wonder if there is some other suite that is truly missing something.

PR #55328, opened 2025-07-13
Clarification about instance-level methods defined by cattr_accessor

flavio-b:
### Motivation / Background

I found the explanation about instance-level methods defined by `cattr_accessor` a bit confusing and error-prone, especially the terminology of "changing" and "overriding" values.

### Detail

I rewrote a segment to make it explicit that class variables are in use, and the implication of changing their value.

I rewrote the example to something more neutral, and that briefly demonstrates the effect of the class variable.

[ci skip]

### DIFF
--- a/guides/source/active_support_core_extensions.md
+++ b/guides/source/active_support_core_extensions.md
@@ -977,19 +977,23 @@ class MysqlAdapter < AbstractAdapter
 end
 ```
 
-Instance methods are created as well for convenience, they are just proxies to the class attribute. So, instances can change the class attribute, but cannot override it as it happens with `class_attribute` (see above). For example given
+Instance methods are also created for convenience, but they are simply proxies to the internal value which is shared among the class. As a result, when an instance modifies the value, this affects the entire class hierarchy. This behavior is different than `class_attribute` (see above).
+
+For example:
 
 ```ruby
-module ActionView
-  class Base
-    cattr_accessor :field_error_proc, default: Proc.new {
-      # ...
-    }
-  end
+class Foo
+  cattr_accessor :bar
 end
-```
 
-we can access `field_error_proc` in views.
+instance = Foo.new
+
+Foo.bar = 1
+instance.bar # => 1
+
+instance.bar = 2
+Foo.bar # => 2
+```
 
 The generation of the reader instance method can be prevented by setting `:instance_reader` to `false` and the generation of the writer instance method can be prevented by setting `:instance_writer` to `false`. Generation of both methods can be prevented by setting `:instance_accessor` to `false`. In all cases, the value must be exactly `false` and not any false value.
 


guides/source/active_support_core_extensions.md
```diff
@@ -977,7 +977,7 @@ class MysqlAdapter < AbstractAdapter
 end
 ```
 
-Instance methods are created as well for convenience, they are just proxies to the class attribute. So, instances can change the class attribute, but cannot override it as it happens with `class_attribute` (see above). For example given
+Instance methods are also created for convenience, but they are simply proxies to the underlying class variable. As a result, when an instance sets the value, it modifies the shared class variable, which affects the entire class hierarchy. This is different behaviour when compared with `class_attribute` (see above). Here’s an example of the instance-level method in use:
```
skipkayhil: I think this is a good improvement, except for the last sentence. The code example is in the middle of a sentence 

> For example given ... we can access `field_error_proc` in views

but the new phrasing doesn't work 

> Here's an example of the instance-level method in use: ... we can access `field_error_proc` in views

flavio-b: Thank you! Come to think of it, perhaps the example should be removed altogether, because:

a) The concept of instance-level methods is mostly self-explanatory. 

b) I find the example distracting because it talks about Views, an unrelated topic, and it's not clear why one would use that technique instead of view helpers, or the myriad of other ways of having instance methods in views. 

c) It seems like `cattr_accessor` is not super encouraged these days because of class variables, so maybe there's no harm in removing the example.

zzak: Just my opinion but until we have officially (or somewhat-officially) deprecated something, removing the documentation seems undesireable. But I don't feel comfortable suggesting we rewrite the example here either.

I think the following examples are there to help the reader understand the usage, not that it is a recommended use-case?

It doesn't make sense to show people how something works by monkey-patching Rails, it'd be better to find a more realistic example which follows modern best practices -- but I realize that may be asking a lot of this PR.

flavio-b: Thank you for the input! I thought about other examples, but nothing concrete comes to mind that I would feel comfortable using `cattr_accessor` instance methods for, that `class_attribute` doesn't already cover in a much safer way.

I could add something neutral, like
```
class Foo
  cattr_accessor :bar
end

instance = Foo.new
Foo.bar = 1
instance.bar # => 1

instance.bar = 2
Foo.bar # => 2
```

but, that inadvertently could make it seems like a desirable effect! The text explanation is, IMO, sufficiently clear: Instance methods are created, and they're backed by a class variable and its known effects.

I kept out words like "beware", "dangerous", to avoid judgement and any suggestion of deprecation. It was only to shed light on the mechanism, remind a bit about class variable implications, and assume the reader will know the rest and decide.

Removing the current example would be simply to prevent confusion. When I saw it, I kept thinking: Why did they choose this example? Is this a scenario where class variables bring some unique advantages, despite the risks? Am I missing out by not using this?

zzak: I think neutral, non-specific examples are fine given they convey the point of this feature. They can decide how and when to use it as they wish.

flavio-b: Ok, I updated the example with the neutral version.

flavio-b:
Is there anything else that I can do to help move this PR forward?

guides/source/active_support_core_extensions.md
```diff
@@ -977,19 +977,21 @@ class MysqlAdapter < AbstractAdapter
 end
 ```
 
-Instance methods are created as well for convenience, they are just proxies to the class attribute. So, instances can change the class attribute, but cannot override it as it happens with `class_attribute` (see above). For example given
+Instance methods are also created for convenience, but they are simply proxies to the underlying class variable. As a result, when an instance sets the value, it modifies the shared class variable, which affects the entire class hierarchy. This is different behaviour when compared with `class_attribute` (see above). Example:
```
zzak:
```suggestion
Instance methods are also created for convenience, but they are simply proxies to the internal value which is shared among the class. As a result, when an instance modifies the value, this affects the entire class hierarchy. This behavior is different than `class_attribute` (see above).

For example:
```

zzak:
@flavio-b Could you [squash your commits](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#squashing-commits) please?

flavio-b:
@zzak, it's done.

zzak (approved):

skipkayhil (approved):

skipkayhil:
Thank you both!

PR #55335, opened 2025-07-14
Introduce `Object#absence`, complementary to `Object#presence`

p8:
### Motivation / Background

Sometimes you want to return a value when a variable is blank, or nil otherwise.

`Object#absence` returns `true` if the receiver is blank, otherwise `nil`. This is equivalent to:

```ruby
object.blank? ? true : nil
```

For example, when defining a Hash where:
* `:processed_at` should be set to `Time.current` if `processed` is `true`, otherwise set it to `nil`.
* `:process_at` should be set to `Time.current` if `processed` is `false`, otherwise set it to `nil`.

This could be defined as:
```ruby
  processed = false
  {
    processed_at: processed.present? ? Time.current : nil,
    process_at: processed.blank? ? Time.current : nil
  }
  # => { processed_at: nil, process_at: Mon, 14 Jul 2025 ... }
```

Using `presence` and `absence` this could be defined as:

```ruby
  {
    processed_at: processed.presence && Time.current,
    process_at: processed.absence && Time.current
  }
  # => { processed_at: nil, process_at: Mon, 14 Jul 2025 ... }
```

Another use-case would be setting a class on an HTML element if a variable is false/blank:

```erb
<div class="<%= enabled.blank? ? "disabled" : "" %>">...
```

With `absence` this could be changed to:

```erb
<div class="<%= enabled.absence && "disabled" %>">...
```

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,26 @@
+*   Introduce `Object#absence`, complementary to `Object#presence`.
+
+    Returns `true` if the receiver is blank, otherwise `nil`. This is
+    equivalent to:
+
+    ```ruby
+    object.blank? ? true : nil
+    ```
+
+    This can be useful when you want to return a value when a variable is blank,
+    or nil otherwise:
+
+    ```ruby
+    processed = false
+    {
+      processed_at: processed.presence && Time.current,
+      process_at: processed.absence && Time.current
+    }
+    # => { processed_at: nil, process_at: Mon, 14 Jul 2025 ... }
+    ```
+
+    *Petrik de Heus*
+
 *   Given an array of `Thread::Backtrace::Location` objects, the new method
     `ActiveSupport::BacktraceCleaner#clean_locations` returns an array with the
     clean ones:
--- a/activesupport/lib/active_support/core_ext/object/blank.rb
+++ b/activesupport/lib/active_support/core_ext/object/blank.rb
@@ -26,7 +26,29 @@ class Object
     !blank?
   end
 
-  # Returns the receiver if it's present otherwise returns +nil+.
+  # Returns +true+ if the receiver is blank, otherwise returns +nil+.
+  # <tt>object.absence</tt> is equivalent to
+  #
+  #    object.blank? ? true : nil
+  #
+  # For example, something like
+  #
+  #   {
+  #     process_at: params[:processed].blank? ? Time.current : nil
+  #   }
+  #
+  # becomes
+  #
+  #   {
+  #     process_at: params[:processed].absence && Time.current
+  #   }
+  #
+  # @return [true, nil]
+  def absence
+    true if blank?
+  end
+
+  # Returns the receiver if it's present, otherwise returns +nil+.
   # <tt>object.presence</tt> is equivalent to
   #
   #    object.present? ? object : nil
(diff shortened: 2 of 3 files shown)


MatheusRich:
I like the complement aspect of this, but for some reason my brain doesn't parse it as easily as `presence`. I keep trying to rewrite the expression you gave it with only presence:

```rb
{
  processed_at: processed.presence && Time.current,
  process_at: processed.presence || Time.current
}
```

(not 100% exactly the same, but you got the idea)

p8:
@MatheusRich In this case if `processed` is anything but `false` or `nil` it will set `process_at` to that value (instead of `nil`):
```ruby
processed = "1"
{
  processed_at: processed.presence && Time.current,
  process_at: processed.presence || Time.current
}
# => {processed_at: 2025-07-16..., process_at: "1"}

{
  processed_at: processed.presence && Time.current,
  process_at: processed.absence && Time.current
}
# => {processed_at: 2025-07-16..., process_at: nil }
```

p8:
`absence` would also allow chaining, although that might get somewhat difficult to understand...
```ruby
1.absence && 2.absence && 3 # => nil
nil.absence && 2.absence && 3 # => nil
nil.absence && nil.absence && 3 # => 3
```

p8:
Another example of chaining would be setting a class to "active" when checking multiple variables:
```erb
<div class="<%= hidden.absence && disabled.absence && "active" %>">...
```
This renders `<div class="active">` only when `hidden` and `disabled` are blank, otherwise `<div class="">`.

This would also allow combining it with `presence`:

```erb
<div class="<%= hidden.absence && enabled.presence && "active" %>">...
```
This renders `<div class="active">` only when `hidden` is blank and `enabled` is present, otherwise `<div class="">`.

This is the same as (which might be more readable?):
```erb
<div class="<%= (hidden.blank? && enabled.present?) ? "active" : nil %>">...
```

Having the inverse makes it easier to refactor when you want the opposite of what `presence` currently does,
just change `presence` to `absence`.

For example, if we have the following code and want to use the inverse of the `hidden` variable instead:
```erb
<div class="<%= hidden.presence && "hidden" %>">...
```

... we could use the new variable and rename `presence` to `absence`:

```erb
<div class="<%= visible.absence && "hidden" %>">...
```

zzak:
I don't know about adding another method to `Object` when there is already a similar method that performs the same task.

We have `Object#blank?` and `Object#present?` already, and `Object#presence` is meant to be a compliment to `present?` (afaict) -- is `abscence` intended to be the compliment to `blank?`?

My feeling is that stuff in Object should be there if matz doesn't want it in core, or it's specific to Rails.
I think in the past we started with proposing these types of features to ruby-core first, but not sure that is necessary or I may be misremembering.

p8:
@zzak Yes, `Object#absence` is intended to complement `blank?`.
I basically want a method that returns nil if the receiver is blank and true otherwise.

I first had `Object#blankness` but I didn't like that name.
We also use `absence` instead of `blankness` in validations:
https://edgeapi.rubyonrails.org/classes/ActiveModel/Validations/HelperMethods.html#method-i-validates_absence_of

rafaelfranca:
Thank you for the pull request but this doesn't reads as nice as `presence` mostly because you need to use `&&` instead of `||`

`foo.absence && "default"` looks like a boolean expression, not as a way to use a fallback.

PR #55336, opened 2025-07-15
[Fix #55209] Remove autocomplete="off" from hidden inputs in button_to

### LINKED ISSUE #55209: button_to and form_form helper produce invalid html
`<input type="hidden" name="_method" value="delete" autocomplete="off">`
`<input type="hidden" name="authenticity_token" value="ov9jo3L" autocomplete="off">`

W3C Validator says:
Error: An input element with a type attribute whose value is hidden must not have an autocomplete attribute whose value is on or off.

I believe the feature was added a long time ago to prevent some bugs in browsers.
#42610 

But now we should consider to remove it in favor of valid html

nkulway:
### Motivation / Background

This Pull Request removes the `autocomplete="off"` attribute from hidden form inputs generated by button_to and its helpers. This attribute is unnecessary on hidden inputs and caused invalid HTML.
Fixes #[55209](https://github.com/rails/rails/issues/55209)

### Detail

This Pull Request changes `button_to`, `token_tag`, and `method_tag` to stop adding `autocomplete: "off"` to hidden fields. Tests have been updated to reflect the new behavior.

### Additional information

This improves HTML standards compliance and avoids inserting redundant attributes.

### Checklist


zzak:
We probably have to make this an option we pass in or configurable as to not break existing applications.

actionview/lib/action_view/railtie.rb
```diff
@@ -15,6 +15,8 @@ class Railtie < Rails::Engine # :nodoc:
     config.action_view.apply_stylesheet_media_default = true
     config.action_view.prepend_content_exfiltration_prevention = false
 
+    config.action_view.remove_hidden_field_autocomplete = true
```
rafaelfranca: Can you document this new config in the configuring guide?

rafaelfranca (changes requested):

rafaelfranca (approved):

tagliala:
Hello,

This change introduced some test failures here: https://github.com/DavyJonesLocker/client_side_validations/issues/987

Our spec checks for the presence of the `autocomplete` attribute like this:

```rb
def autocomplete_attribute
  @autocomplete_attribute ||=
    if hidden_field_tag(:test).include?('autocomplete')
      'autocomplete="off"'
    else
      ''
    end
end
```

With Rails edge, `hidden_field_tag` correctly omits the `autocomplete` attribute. However, hidden fields generated by the following helpers still include it:
- `check_box`
- `select` (with `multiple`)
- `file_field`

These helpers use `tag` directly instead of `hidden_field_tag` and therefore don’t benefit from the recent change.

You can find the relevant code by searching for `/tag\("input".+autocomplete/`:
- https://github.com/rails/rails/blob/5a0ba202b9c855534a88057c04d741b62fc391bb/actionview/lib/action_view/helpers/tags/check_box.rb#L60
- https://github.com/rails/rails/blob/5a0ba202b9c855534a88057c04d741b62fc391bb/actionview/lib/action_view/helpers/tags/file_field.rb#L21
- https://github.com/rails/rails/blob/5a0ba202b9c855534a88057c04d741b62fc391bb/actionview/lib/action_view/helpers/tags/select_renderer.rb#L25

Would you prefer that I open a new issue for this, or continue the discussion here?

Thank you!

nkulway:
Hey @tagliala,

Thanks for catching this! You're absolutely right, this inconsistency needs to be addressed.
Could you please open a new issue for this? It would be helpful if you could include:
- specific helpers affected
- links to the relevant code
- maybe mention that it's a follow-up to this PR

I'd be happy to work on the fix for this, or if someone else feels like taking it on go for it! It should follow the same pattern as the original change by using the `remove_hidden_field_autocomplete` config.

Thanks for taking the time to test this thoroughly and report back!

tagliala:
Hi @nkulway, you're welcome. New issue opened at #55619

I will also add a spec for multiple files, which is missing

PR #55337, opened 2025-07-15
Update getting_started.md

manfredyau:
When I followed this tutorial, I found that it wasn't working and an error occurred:  Error performing ActionMailer::MailDeliveryJob (Job ID: 31109699-faaf-4d31-b83d-cf146353defe) from Async(default) in 22.71ms: ActionView::Template::Error (No route matches {:action=>"show", :controller=>"unsubscribes", :token=>"eyJfcmFpbHMiOnsiZGF0YSI6WzNdLCJwdXIiOiJTdWJzY3JpYmVyXG51bnN1YnNjcmliZVxuIn19--efad91101aeb598f03698e7494e2868b4e1356c5"}, missing required keys: [:id]):

To fix it, I've modified some code in routes.rb, and now it's working.

[ci skip]

### Motivation / Background

### Detail

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2364,7 +2364,7 @@ First, we need a route for unsubscribing that will be the URL we include in
 emails.
 
 ```ruby
-    get "/unsubscribes/:token", to: "unsubscribes#show", as: :unsubscribe
+    resources :unsubscribe, only: [ :show ], param: :token
 ```
 
 Active Record has a feature called `generates_token_for` that can generate
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2364,7 +2364,7 @@ First, we need a route for unsubscribing that will be the URL we include in
 emails.
 
 ```ruby
-  resource :unsubscribe, only: [ :show ]
+    get "/unsubscribes/:token", to: "unsubscribes#show", as: :unsubscribe
 ```
 
 Active Record has a feature called `generates_token_for` that can generate


manfredyau:
[ci skip]

guides/source/getting_started.md
```diff
@@ -2364,7 +2364,7 @@ First, we need a route for unsubscribing that will be the URL we include in
 emails.
 
 ```ruby
-  resource :unsubscribe, only: [ :show ]
+    get "/unsubscribes/:token", to: "unsubscribes#show", as: :unsubscribe
```
MatheusRich: I didn't follow the tutorial yet, but if this is indeed wrong, could we still stick with using `resources`?
```suggestion
    resources :unsubscribe, only: [ :show ], param: :token
```

manfredyau:
[ci skip]

excid3:
~~Good catch @manfredyau. I forgot the `param` option in the route. 👍~~

guides/source/getting_started.md
p8: Shouldn't this be `resource` instead of `resources`, as `id` isn't set?
```ruby
resource :unsubscribe, only: [ :show ], param: :token
```

p8: Ah nevermind, `token` replaces `id`.

excid3: @p8 oh yeah, it was written as singular on purpose.

A singular resource doesn't need a param, so this change isn't necessary as the URL is:

```
/unsubscribe?token=xyz
```

Maybe it's the mail template that is wrong instead.

excid3:
Re-reviewing thanks to @p8:

The guide is correct, this should be a _singular_ route `resource :unsubscribe` so it generates `/unsubscribe`

The error you pasted suggests that you used `resources` plural as it's requiring an ID to generate the URL in the mailer.

> ActionView::Template::Error (No route matches {:action=>"show", :controller=>"unsubscribes", :token=>"eyJfcmFpbHMiOnsiZGF0YSI6WzNdLCJwdXIiOiJTdWJzY3JpYmVyXG51bnN1YnNjcmliZVxuIn19--efad91101aeb598f03698e7494e2868b4e1356c5"}, missing required keys: [:id]):

The url in the mailer should reference the singular resource path without an ID required and using the `token` query param to pass it along.

```erb
<%= link_to "Unsubscribe", unsubscribe_url(token: params[:subscriber].generate_token_for(:unsubscribe)) %>
```

If you had the plural `resources :unsubscribes` route, it would raise this error.

So I don't think this should be merged since it's working as expected already.

MatheusRich:
@manfredyau Can you confirm the mailer route was wrong? If so, I'll cose this PR.

manfredyau:
> @manfredyau Can you confirm the mailer route was wrong? If so, I'll cose this PR.

Yes, I have tried to modify **resources** to **resource**, and it works well. That was my fault. Thanks, everyone.

p8:
Maybe we should make it clear that this route uses `resource` in this case?

PR #55338, opened 2025-07-15
Fix numericality validator skipping validation after normalization of invalid input

himanshukale36:
… input

### Motivation / Background

This PR fixes a bug in the `numericality` validator where invalid input (e.g., `"abc"`) could be incorrectly accepted when normalization is applied. Specifically, when Active Record typecasts `"abc"` to `0`, and a normalizer turns `0` into `nil`, the validator skips checking due to `allow_nil: true`.

Fixes #55326

### Detail

The `ActiveModel::EachValidator#validate` method now performs `allow_nil` and `allow_blank` checks **after normalization**, rather than before. This ensures that validations are always performed on the final, normalized value.

### Additional information

A regression test has been added to `activerecord/test/cases/validations/numericality_validation_test.rb` to verify that invalid input such as `"abc"` is rejected, even when normalization would otherwise convert it to `nil`.

### Checklist


eglitobias:
FYI: This bug got already fixed by https://github.com/rails/rails/pull/55309

PR #55339, opened 2025-07-15
Handle counter cache registration for associations with unloaded classes

mickeytgl:
### Motivation / Background

Fixes #55133 

This commit addresses an issue with ActiveRecord's counter cache feature that occurred when the associated class was loaded after the class defining the association. `belongs_to`.

Previously, for a counter cache to be correctly registered, the associated class needed to be loaded first. As pointed out in the issue this PR fixes: This may happen, for example, in scripts using Active Record directly without autoloading.

This commit introduces a that stores pending counter cache registrations. When a class is defined, this registry is checked, and any pending counter caches for that class are applied. `CounterCacheRegistry`

### Additional information

This PR's approach was heavily inspired on `ActiveRecord::Base.descendants` implementation since it also keeps track of records of classes and solve load-order issues. However, since `DescendantsTracker` is focused on inheritance, it wouldn't have worked in this case

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/builder/belongs_to.rb
+++ b/activerecord/lib/active_record/associations/builder/belongs_to.rb
@@ -38,7 +38,11 @@ module ActiveRecord::Associations::Builder # :nodoc:
       }
 
       klass = reflection.class_name.safe_constantize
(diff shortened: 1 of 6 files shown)


mickeytgl:
Hey @fxn , let me know if this approach makes sense.

PR #55343, opened 2025-07-16
Fix test suite compatibility with propshaft 1.2.0

byroot:
Ref: https://github.com/rails/propshaft/pull/238

The newer version has a different manifest schema.

cc @yahonda


rafaelfranca:
Thanks for fixing it! I should have tested Rails as well.

PR #44996, opened 2022-05-02
 Introduce `config/application/` directory

casperisfine:
Ref: https://github.com/rails/rails/pull/44980

Historically, configs were supposed to be entirely set from `application.rb` and the various `config/environments/` files.

But Rails itself broke this assumption by setting some configurations from initializer files. Actually all the default initializer files are setting configurations.

The problem with this is that it becomes very hard to know when a specific configuration is considered final and is applied. This has led to many configurations being specifically applied on `after_initialize` so that they can be defined in initializers (See https://github.com/rails/rails/pull/44980 as one of the many example).

This is bad for several reasons:

  - We need an ad hoc fix every time.
  - It's totally unclear for users when their config actually takes effect.
  - It cause some other initializers to need to be delayed as well.
  - If you get it wrong, your config might silently not be applied.

The idea here is two fold.

First we introduce `config/application` so that we better support the pattern of breaking down `application.rb` in smaller initialization files, which seems to be the original need here.

Then in a followup PR, we can warn or error out when the `config` object is mutated past the point where it no longer is applied. This can be implemented as a new framework default for 7.1, e.g. (`config.freeze = true`).

PR #44980, opened 2022-04-29
Allow setting some Active Support settings via initializers

ghiculescu:
Through investigating https://github.com/rails/rails/pull/43596#discussion_r860540140 I noticed a few settings can't be configured through initializers. This PR fixes it for these, to match how most other settings in the Active Support railtie work.

- `isolation_level`
- ~~`report_deprecations`~~
- `executor_around_test_case`

**Pros of this change**: you can now set these settings in initializers, such as the new framework defaults initializer created when upgrading Rails.

**Cons of this change**: the changes won't apply while application initializers are running. So for example you might still report deprecations during application initializers, even if they get silenced later.

**Should we merge this?** Yes, I think so - currently new framework defaults suggests configs that don't work. But in future it's probably worth looking more into the configuration architecture.

cc @byroot

PR #55347, opened 2025-07-16
Add load-hooks for Active Model autoloaded constants referenced in initializers

bensheldon:
### Motivation / Background

This Pull Request has been created because I noticed that there are 2 autoloaded constants referenced in the Active Model initializer that are not wrapped with load hooks. This means that the autoloaded constants will be loaded immediately during boot rather than deferred.

Additionally, I noticed that the Active Model configuration is mutated with a `delete` unnecessarily and I removed the `delete`. afaict, configuration is only necessarily mutated when configuration is mass-assigned against another interface. For example, Action View does a lot of `config.action_view.delete` in its Railtie because it later does this:

https://github.com/rails/rails/blob/5d8c81c2c0985dbb1c99d16b08e272b05a1dd176/actionview/lib/action_view/railtie.rb#L80-L86

(I say "afaict" because I'm pretty sure the `delete` pattern is not intended to signal consumption in the theme of #44996 ... though there are a few other seemingly unnecessary imo usages in the codebase... Happy to clean those up if given the signal) 

### Checklist

* ~~[ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.~~ didn't seem to warrant it

### DIFF
--- a/activemodel/lib/active_model/error.rb
+++ b/activemodel/lib/active_model/error.rb
@@ -204,4 +204,6 @@ module ActiveModel
         [@base, @attribute, @raw_type, @options.except(*CALLBACKS_OPTIONS)]
       end
   end
+
+  ActiveSupport.run_load_hooks(:active_model_error, Error)
 end
--- a/activemodel/lib/active_model/railtie.rb
+++ b/activemodel/lib/active_model/railtie.rb
@@ -14,11 +14,15 @@ module ActiveModel
     end
 
     initializer "active_model.secure_password" do
-      ActiveModel::SecurePassword.min_cost = Rails.env.test?
+      ActiveSupport.on_load(:active_model_secure_password) do
+        ActiveModel::SecurePassword.min_cost = Rails.env.test?
+      end
     end
 
-    initializer "active_model.i18n_customize_full_message" do
-      ActiveModel::Error.i18n_customize_full_message = config.active_model.delete(:i18n_customize_full_message) || false
+    initializer "active_model.i18n_customize_full_message" do |app|
+      ActiveSupport.on_load(:active_model_error) do
+        ActiveModel::Error.i18n_customize_full_message = app.config.active_model.i18n_customize_full_message || false
+      end
     end
   end
 end
--- a/activemodel/lib/active_model/secure_password.rb
+++ b/activemodel/lib/active_model/secure_password.rb
@@ -228,4 +228,6 @@ module ActiveModel
       end
     end
   end
+
+  ActiveSupport.run_load_hooks(:active_model_secure_password, SecurePassword)
 end


bensheldon:
Current test failures seem unrelated to this change 👍🏻

Edouard-chin:
👋 Thanks for the PR @bensheldon ! I have a question for my own curiosity.

> I noticed that there are 2 autoloaded constant [...] will be loaded immediately during boot rather than deferred.

I maybe misunderstood something regarding lazy load hooks, but for those constants isn't that fine to load them at boot time? They have a very small footprint and loading them at boot time doesn't interfere with Rails' booting process (like it's the case for example if an applications uses `ActionView::Base` too early at boot time).

There are many other initializers that make use of autoloaded constants, should those also have their own lazy hook?

bensheldon:
@Edouard-chin I'm happy to explain my reasoning more.

> They have a very small footprint and loading them at boot time doesn't interfere with Rails' booting process 

I think this is the crux of it. If they're so small that they don't impact boot time, then I don't think they should be autoloaded. Setting up a constant to autoload and then always/deliberately requiring it only a little later does take more time than simply requiring it in the first place.

> ...like it's the case for example if an applications uses `ActionView::Base` too early at boot time

I think it's (cognitively/socially/maintainably) easier to think of this strictly as 2 categories of constants, rather than 3:

- Constants that must be loaded during boot (not autoloaded).
- Constants that can be deferred from boot (autoloaded). 
- ~Constants that look like they have been deferred but are loaded during boot! (autoloaded but referenced during boot).~

(I'll share that I don't have a comprehensive opinion yet about why one would make an autoloadable constant eager-autoloaded or not, but we can probably set that aside)

The other direction this PR could have gone is to remove the autoload for these two constants and require them directly. But they _could_ be deferred with a load hook, so I went in that direction.

> There are many other initializers that make use of autoloaded constants, should those also have their own lazy hook?

_I_ think so. I'm happy to give you feedback on any PRs or work with you on it.

rafaelfranca:
I mean, those constants are always eager loaded in production, so they are loaded as part of the boot. I just accepted this PR because of the `SecurePassword` that it is not a problem to load, but, not every application uses it, so I just want to avoid to load it in development.

rafaelfranca:
Although, I just realized that `ActiveRecord::Base` loads it.

Edouard-chin:
Thank you for explaining your reasoning @bensheldon ! 

I was mostly asking because lazy load hooks tend to be public facing and Rails document which load hooks are available for an application https://github.com/rails/rails/blob/1fcd079e1dfb7be7792897fb3faad07026dd2c46/guides/source/configuring.md#L3828-L3830

I would agree a bit more with Rafael on why this is worth having (e.g. if `SecurePassword` is not always used), but admittedly don't think that the `ActiveModel:Error` should have a hook.

> If they're so small that they don't impact boot time, then I don't think they should be autoloaded.

Individually that's true but if we were to modify all autoloaded constants that have a small footprint and instead require them on the spot, this would I believe slow down boot time (for instance requiring `activesupport/all` takes a while, even though each individual constants have a small footprint).

rafaelfranca:
Ah. Those hooks I'd not document. They should not be public.

bensheldon:
Great discussion!

Active Support is maybe not a good example because I understand the autoloading to have a functional purpose because requiring `active_support/all` is optional in Rails framework:

https://github.com/rails/rails/blob/1921f0fef60983d6423b937884de5e83b2619960/railties/lib/rails/application/bootstrap.rb#L15-L17

The lazy load hook documentation says lazy load hooks exist for the express purpose of configuring autoloaded constants:

https://github.com/rails/rails/blob/af42a4443c2e2018f58022464aaf308cb4d36af1/activesupport/lib/active_support/lazy_load_hooks.rb#L7-L9

... hence my very blunt-headed: "autoloaded constant" + "configured in initializer" => "lazy load hook".

And in the same vein of functional consistency, I'm gonna go make that PR to remove the non-functional `config.delete`s 😁

bensheldon:
@Edouard-chin If you're still curious about this, I added [a test to smoke out](https://github.com/bensheldon/rails/pull/2) all of the autoloaded constants that get loaded during boot. It's interesting!! (for example, TIL that Action Cable uses Zeitwerk)

```
❯ OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES bin/test test/application/loading_test.rb:560
Run options: --seed 38266

# Running:

# LOTS OF DEPRECATION WARNINGS
E

Error:
LoadingTest#test_rails_framework_constants_marked_for_autoloading_are_not_loaded_during_initialization:
RuntimeError: The following constants were marked for autoloading but were loaded during initialization:
  AbstractController::Helpers
  ActionCable::Engine
  ActionController::Helpers
  ActionDispatch::ActionableExceptions
  ActionDispatch::Callbacks
  ActionDispatch::ContentSecurityPolicy
  ActionDispatch::Cookies
  ActionDispatch::DebugExceptions
  ActionDispatch::ExceptionWrapper
  ActionDispatch::Executor
  ActionDispatch::Flash
  ActionDispatch::HostAuthorization
  ActionDispatch::Http::Cache
  ActionDispatch::Http::Headers
  ActionDispatch::Http::MimeNegotiation
  ActionDispatch::Http::Parameters
  ActionDispatch::Http::URL
  ActionDispatch::MiddlewareStack
  ActionDispatch::ParamBuilder
  ActionDispatch::PublicExceptions
  ActionDispatch::QueryParser
  ActionDispatch::Reloader
  ActionDispatch::RemoteIp
  ActionDispatch::Request
  ActionDispatch::RequestId
  ActionDispatch::Routing::Mapper
  ActionDispatch::ServerTiming
  ActionDispatch::Session::AbstractSecureStore
  ActionDispatch::Session::AbstractStore
  ActionDispatch::Session::CookieStore
  ActionDispatch::ShowExceptions
  ActionDispatch::Static
  ActionText::Attachment
  ActionText::Attachments::Caching
  ActionText::Attachments::Minification
  ActionText::Attachments::TrixConversion
  ActionView::AbstractRenderer
  ActionView::CacheExpiry
  ActionView::PartialRenderer
  ActionView::PathRegistry
  ActionView::Template
  ActiveJob::Serializers
  ActiveModel::SecurePassword
  ActiveRecord::Associations
  ActiveRecord::AsynchronousQueriesTracker
  ActiveRecord::ConnectionAdapters
  ActiveRecord::Marshalling
  ActiveRecord::QueryCache
  ActiveRecord::QueryLogs
  ActiveRecord::Reflection
  ActiveRecord::RuntimeRegistry
  ActiveRecord::SecureToken
  ActiveStorage::Analyzer
  ActiveStorage::Attached
  ActiveStorage::FixtureSet
  ActiveStorage::Previewer
  ActiveStorage::Service
  ActiveStorage::Transformers::ImageProcessingTransformer
  ActiveStorage::Transformers::Transformer
  ActiveStorage::Transformers::Vips
  ActiveSupport::ArrayInquirer
  ActiveSupport::Cache
  ActiveSupport::Dependencies
  ActiveSupport::Digest
  ActiveSupport::EnvironmentInquirer
  ActiveSupport::ExecutionContext
  ActiveSupport::ExecutionWrapper
  ActiveSupport::Executor
  ActiveSupport::FileUpdateChecker
  ActiveSupport::Reloader
  ActiveSupport::StringInquirer
  ActiveSupport::TaggedLogging
  GlobalID::Verifier
  I18n::Backend::Base
  I18n::Backend::Simple
  I18n::Backend::Transliterator
  Module::Mime
  Module::SignedGlobalID
  Object::Mime
  Object::SignedGlobalID
  Rack::BodyProxy
  Rack::ConditionalGet
  Rack::ETag
  Rack::Files
  Rack::Head
  Rack::MethodOverride
  Rack::Runtime
  Rack::Sendfile
  Rack::TempfileReaper
  Rails::Application::Bootstrap
  Rails::Application::Configuration
  Rails::Application::DefaultMiddlewareStack
  Rails::Application::Finisher
  Rails::Application::LazyRouteSet
  Rails::Application::RoutesReloader
  Rails::Engine::Configuration
  Rails::Engine::LazyRouteSet
  Rails::Rack::Logger
    test/application/loading_test.rb:584:in 'block in <class:LoadingTest>'


bin/test test/application/loading_test.rb:560


```

Edouard-chin:
Thanks for writing this smoke test , it's indeed quite interesting ! I'm not too surprised about the result.

Just to add on your previous comment:

> The lazy load hook documentation says lazy load hooks exist for the express purpose of configuring autoloaded constants:

I agree with this, but maybe the phrasing of the documentation should be tweaked to mention that it's especially useful when loading specific contants incurs a performance penalty, at boot time.

I believe the load hook features was written for two purposes (happy to be corrected though!)

1. Speed up boot time
2. Hook on to constants that **may** be defined at a later time.

The constants surfaced from the smoke test don't seem to affect boot time significantly and will always be defined (i.e. as opposite to `SecureCompare` which only gets loaded if an application uses it or use Active Record). And for these reasons I don't *think* that we should add load hooks for them.

> The other direction this PR could have gone is to remove the autoload for these two constants and require them directly.

You are right that we could require them directly, but users using those gems without rails would have to require all of these files one by one instead of just "require 'active_model'` and having all autoloaded constants setup.

Nonetheless, thank you for the discussion, it's very interesting and it challenges my comprehension of autoloading and load hooks 😃.

bensheldon:
@Edouard-chin I think we're aligned on this. With a couple of qualifiers.

I don't expect that every one of these will require a load-hook. I imagine that a small number of the constants are being explicitly referenced and _could_ be load-hooked, and the majority of the other constants coming along for the ride. So changing a small number of them potentially would change them all.

I did some [messy benchmarking](https://gist.github.com/bensheldon/ea85f7f4836942c55fb362048f4202ed)---defining and loading 10k constants---and autoload+reference is about 4% slower than a `require`, 6% slower with Zeitwerk loaded/setup (not loaded via Zeitwerk, merely that Zeitwerk extends some of the codepaths and causes more work to happen).  

Also, I can't believe I failed to mention the functional problem with not load-hooking constant configuration in railties/engines 🤦🏻 : **the absence of a load hook prevents/breaks configuring with Rails.config in application initializers**. That's because engines/railties run before application initializers run. Deferring configuration via a load-hook allows the application initializers to add configuration before it's applied to the constant when the constant is loaded later.  

Edit to add: I'm not entirely sure where the current consensus is on config-in-initializers. [Nice summary here](https://github.com/rails/rails/pull/44980#issuecomment-1114171816) though [maybe superseded](https://github.com/rails/rails/pull/44996#issuecomment-1119589027) and [also](https://github.com/rails/rails/pull/44996#issuecomment-1152674200):

> A new Rails app should not have any problem on setting config on initializers. 

(but again, I imagine that not all these constants are configured, just a few of them)

> You are right that we could require them directly, but users using those gems without rails would have to require all of these files one by one instead of just "require 'active_model'` and having all autoloaded constants setup.

I'm imagining that, if explicitly required, they would be required in either:

- via the base component (e.g. a `require` in `lib/active_model.rb`), which is the conventional require for gems via bundler, or,
- via the railtie/engine (e.g. a `require` in `lib/active_model/railtie.rb`) which is loaded by the Rails framework in [`rails/all`](https://github.com/rails/rails/blob/main/railties/lib/rails/all.rb)

Anyways, fun stuff. I'm gonna do some more analysis on this and see if I can narrow down the causes 😁

PR #48682, opened 2023-07-06
Introduce `config.dom_testing_default_html_version` and use Rails::Dom::Testing to parse HTML in test helpers

flavorjones:
### Motivation / Background

Some early users of Rails edge reported test breakage due to the change in #48523 that updated test helpers to use Nokogiri's HTML5 parser.

After discussing with Rafael, he suggested that we introduce a Rails config parameter so users may opt into this change at their convenience, essentially wrapping a config param around the changes in #48523.

This PR takes that idea a step further and groups all the test helpers from `ActionView::TestCase`, `ActionDispatch::Assertions`, and everything in `Rails::Dom::Testing` under a single config param:

```
config.dom_testing_default_html_version
```

As a result, by setting this config to either `:html4` or `:html5`, a developer can control the HTML parser used by all the test helpers in Action View, Action Dispatch, and Rails::Dom::Testing; and some assertions support overriding that default for individual tests.

### Detail

- new config parameter `config.dom_testing_default_html_version`, which is injected into `Dom::Rails::Testing` via the railstie in https://github.com/rails/rails-dom-testing/pull/109
  - the value of that param is `:html5` in Rails 7.1, and `:html4` in earlier config versions
  - though, on JRuby, where Nokogiri::HTML5 is not supported, this will always be `:html4`
- Action View and Dispatch test helpers use `Rails::Dom::Testing.html_document` or `.html_document_fragment` to ensure the desired HTML parser is used
  - Update dependency on `rails-dom-testing` to `~> 2.2`

### Additional information

Note that integration testing across gems like this is challenging, but I have actually fired up a Rails app using this PR and https://github.com/rails/rails-dom-testing/pull/109 together, and it worked as expected.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55348, opened 2025-07-16
Extend `ActionMailer::TestCase` with multi-part assertions

seanpdoyle:
The Problem
---

Prior to this change, testing the content of multi-part Action
Mailer-generated `Mail` instances involved parsing content derived from
each part into the appropriate type. For example, tests might read
`mail.body.raw_source` from `text/plain` bodies into `String` instances
and parse `text/html` bodies into HTML documents with Nokogiri.

The Proposal
---

This commit defines `assert_part` and `assert_no_part` as assertion
methods on `ActionMailer::TestCase`.

The `assert_part` method iterates over each available part, and asserts
that the multi-part email includes a part that corresponds to the
`content_type` argument.

```ruby
test "assert against the HTML and text parts of the last delivered MyMailer.welcome mailer" do
  MyMailer.welcome("Hello, world").deliver_now

  assert_part :text do |text|
    assert_includes text, "Hello, world"
  end
  assert_part :html do |html|
    assert_dom html.root, "h1", "Hello, world"
  end
  assert_no_part :xml
end

test "assert against the HTML and text parts of a MyMailer.welcome mail instance" do
  mail = MyMailer.welcome("Hello, world")

  assert_part :text, mail do |text|
    assert_includes text, "Hello, world"
  end
  assert_part :html, mail do |html|
    assert_dom html.root, "h1", "Hello, world"
  end
  assert_no_part :xml, mail
end
```

Additional details
---

While HTML assertion support is already baked-into the gem's railtie,
calls to `assert_part(:html)` make the fully-formed HTML
document (a `Nokogiri::HTML4::Document` or `Nokogiri::HTML5::Document`
instance) available to the block. The method also accepts a `Mail` instance directly.

By comparison, the [assert_dom_email][] method (and its
`assert_select_email` alias) provided by `Rails::Dom::Testing` yields an
HTML fragment, and cannot assert against any `Mail` instance except for
the last delivery.

[Rails::Dom::Testing]: https://github.com/rails/rails-dom-testing
[assert_dom_email]: https://github.com/rails/rails-dom-testing/blob/v2.3.0/lib/rails/dom/testing/assertions/selector_assertions.rb#L287-L317

### Checklist

### DIFF
--- a/actionmailer/lib/action_mailer/test_case.rb
+++ b/actionmailer/lib/action_mailer/test_case.rb
@@ -38,6 +38,9 @@ module ActionMailer
       include Rails::Dom::Testing::Assertions::DomAssertions
 
       included do
+        class_attribute :_decoders, default: Hash.new(->(body) { body }).merge!(
+          Mime[:html] => ->(body) { Rails::Dom::Testing.html_document.parse(body) }
+        ).freeze # :nodoc:
         class_attribute :_mailer_class
         setup :initialize_test_deliveries
         setup :set_expected_mail
@@ -83,6 +86,53 @@ module ActionMailer
         IO.readlines(File.join(Rails.root, "test", "fixtures", self.class.mailer_class.name.underscore, action))
       end
 
+      # Assert that a Mail instance has a part matching the content type.
+      # If the Mail is multipart, extract and decode the appropriate part. Yield the decoded part to the block.
+      #
+      # By default, assert against the last delivered Mail.
+      #
+      #   UsersMailer.create(user).deliver_now
+      #   assert_part :text do |text|
+      #     assert_includes text, "Welcome, #{user.email}"
+      #   end
+      #   assert_part :html do |html|
+      #     assert_dom html.root, "h1", text: "Welcome, #{user.email}"
+      #   end
+      #
+      # Assert against a Mail instance when provided
+      #
+      #   mail = UsersMailer.create(user)
+      #   assert_part :text, mail do |text|
+      #     assert_includes text, "Welcome, #{user.email}"
+      #   end
+      #   assert_part :html, mail do |html|
+      #     assert_dom html.root, "h1", text: "Welcome, #{user.email}"
+      #   end
+      def assert_part(content_type, mail = last_delivered_mail!)
+        mime_type = Mime[content_type]
+        part = [*mail.parts, mail].find { |part| mime_type.match?(part.mime_type) }
+        decoder = _decoders[mime_type]
+
+        assert_not_nil part, "expected part matching #{mime_type} in #{mail.inspect}"
+
+        yield decoder.call(part.decoded) if block_given?
+      end
+
+      # Assert that a Mail instance does not have a part with a matching MIME type
+      #
+      # By default, assert against the last delivered Mail.
+      #
+      #   UsersMailer.create(user).deliver_now
+      #
+      #   assert_no_part :html
+      #   assert_no_part :text
+      def assert_no_part(content_type, mail = last_delivered_mail!)
+        mime_type = Mime[content_type]
+        part = [*mail.parts, mail].find { |part| mime_type.match?(part.mime_type) }
+
+        assert_nil part, "expected no part matching #{mime_type} in #{mail.inspect}"
+      end
+
       private
         def initialize_test_deliveries
           set_delivery_method :test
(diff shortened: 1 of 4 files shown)


actionmailer/lib/action_mailer/test_case.rb
```diff
@@ -119,6 +235,24 @@ def charset
...
+          end
+        end
+
+        def html_document_class
+          Rails::Dom::Testing.html_document
+        end
+
+        def document_root_element
+          html_document.root
+        end
```
seanpdoyle: @flavorjones I'd introduced these in the same style as https://github.com/rails/rails/pull/48523, https://github.com/rails/rails/pull/48682, and https://github.com/rails/rails-dom-testing/pull/109.

Is there a different or more modern pattern to use instead?

flavorjones: @seanpdoyle You may want to consider calling `Rails::Dom::Testing.html_document` directly rather than implement a `html_document_class` method, especially if you're explicitly wrapping rails-dom-testing anyway.

And `document_root_element` is the convention followed in other parts of the Rails testing frameworks; but does it make sense here where there may be multiple parts that are each an HTML document? I don't see this method being called or tested anywhere so I'm not sure what the intention is.

seanpdoyle: > I don't see this method being called or tested anywhere so I'm not sure what the intention is.

Your assessment is right, this change adheres to the convention for the sake of integrating with `rails-dom-testing`-based assertions.

> but does it make sense here where there may be multiple parts that are each an HTML document?

Is Action Mailer capable of sending more than one HTML part for any outbound `Mail` instance? The intent of these assertions is to simplify test coverage for emails that have a single `text/html` part, a single `text/plain` part, or a body with one of each.

seanpdoyle: @flavorjones In response to your comments, I've attempted to simplify the PR and minimize the number of methods it introduces. As an alternative to aiming to integrate with `rails-dom-testing` directly through the `html_document` and `document_root_element`, I've instead proposed an `assert_decoded_part(:html)` method that integrates with a `Mime[:html]`-specific decoder to yield the parsed document to the block. From there, integration with `rails-dom-testing` is optional. There's also an opportunity to reach for other assertions like Capybara:

```ruby
mail = MyMailer.welcome("Hello, world")

assert_decoded_part :html, mail do |html|
  # baseline rails-dom-testing integration
  assert_dom html.root, "h1", text: "Hello, world"

  # baseline Capybara integration
  Capybara.string(html).assert_css "h1", text: "Hello, world"
end
```

What do you think of those changes?

flavorjones (commented):
@seanpdoyle I really love this approach! And I think it would be a good addition to the Action Mailer test framework.

One aspect of the design I'm curious about: why not have `assert_part` yield decoded content by default?

To me encoded content seems like it would be the dominant use case, especially since it works so well with the rails-dom-testing assertions.

seanpdoyle:
> To me encoded content seems like it would be the dominant use case, especially since it works so well with the rails-dom-testing assertions.

That's great feedback! I've combined the `assert_decoded_part` with the `assert_part`, and reduced the surface area of the changes to only include `assert_part` and `assert_no_part`.

flavorjones (approved):
This API is *chefkiss*. Thank you for all the work that went into it! Two enthusiastic thumbs up. 👍👍

actionmailer/lib/action_mailer/test_case.rb
```diff
@@ -38,6 +38,9 @@ module Behavior
       include Rails::Dom::Testing::Assertions::DomAssertions
 
       included do
+        class_attribute :_decoders, default: Hash.new(:itself.to_proc).merge!(
```
rafaelfranca: What is `:itself.to_proc` here? What is `itself` in this context? The mailer test case?

seanpdoyle: The `:itself.to_proc` value here would return the value without doing additional decoding.

If the `:itself.to_proc` declaration is not explicit enough, maybe it could be declared like:

```ruby
# using Hash#[]=
class_attribute :_decoders, default: Hash.new(->(body) { body }) # :nodoc:
_decoders[Mime[:html]] = ->(body) { Rails::Dom::Testing.html_document.parse(body) }

# using Hash#merge!
class_attribute :_decoders, default: Hash.new(->(body) { body }).merge!(
  Mime[:html] => ->(body) { Rails::Dom::Testing.html_document.parse(body) }
) # :nodoc:

```

Is that an improvement?

rafaelfranca: I think the `merge!` option is an improvement. At least it is clear what it does.

PR #55350, opened 2025-07-17
Document AttributeMethods::Query [ci skip]

p8:
### Checklist

### DIFF
--- a/activerecord/lib/active_record/attribute_methods/query.rb
+++ b/activerecord/lib/active_record/attribute_methods/query.rb
@@ -3,6 +3,38 @@
 module ActiveRecord
   module AttributeMethods
     # = Active Record Attribute Methods \Query
+    #
+    # Adds query methods for attributes that return either +true+ or +false+
+    # depending on the attribute type and value.
+    #
+    # For Boolean attributes this will return +true+ if the value is present
+    # and return +false+ otherwise:
+    #
+    #   class Product < ActiveRecord::Base
+    #   end
+    #
+    #   product = Product.new(archived: false)
+    #   product.archived? # => false
+    #   product.archived = true
+    #   product.archived? # => true
+    #
+    # For Numeric attributes this will return +true+ if the value is a non-zero
+    # number and return +false+ otherwise:
+    #
+    #   product.inventory_count = 0
+    #   product.inventory_count? # => false
+    #   product.inventory_count = 1
+    #   product.inventory_count? # => true
+    #
+    # For other attributes it will return +true+ if the value is present
+    # and return +false+ otherwise:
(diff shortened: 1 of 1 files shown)


activerecord/lib/active_record/attribute_methods/query.rb
```diff
@@ -3,6 +3,38 @@
...
+    #
+    # For Numeric attributes this will return +true+ if the value is a non-zero
+    # number and return +false+ otherwise:
+    #
+    #   product.inventory_count = 0
+    #   product.inventory_count? # => false
+    #   product.inventory_count = 1
+    #   product.inventory_count? # => true
+    #
+    # For other attributes it will return +true+ if the value is present
```
MatheusRich, line 29: Could we add a link to #present? here?

MatheusRich (approved):

activerecord/lib/active_record/attribute_methods/query.rb
p8: We don't do this for blank either:

> An object is present if it’s not blank.

https://edgeapi.rubyonrails.org/classes/Object.html#method-i-present-3F

> Passing `:all_blank` instead of a Proc will create a proc that will reject a record where all the attributes are blank excluding any value for `_destroy`.

https://edgeapi.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html#method-i-accepts_nested_attributes_for

p8:
@MatheusRich I've also added documentation to the `query_attribute` method, similar to:
https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Read.html#method-i-read_attribute
and
https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Write.html#method-i-write_attribute

activerecord/lib/active_record/attribute_methods/query.rb
```diff
@@ -3,6 +3,38 @@
...
+    # For Boolean attributes this will return +true+ if the value is present
+    # and return +false+ otherwise:
+    #
+    #   class Product < ActiveRecord::Base
+    #   end
+    #
+    #   product = Product.new(archived: false)
+    #   product.archived? # => false
+    #   product.archived = true
+    #   product.archived? # => false
```
alexcameron89: Is this supposed to be `true` now?
```suggestion
    #   product.archived? # => true
```

p8: Yes, thank you!

MatheusRich (approved):

p8:
Thanks @MatheusRich @alexcameron89 !

PR #55365, opened 2025-07-18
Add `case_sensitive:` option to `ComparisonValidator`

ryantownsend:
### Motivation / Background

I often find I need to compare strings in validation for either equality or 'other than', but I'm comparing to case-insensitive database columns (`citext` in PostgreSQL), but the comparison validator is case-sensitive so I end up having to write my own validation method – introducing a `case_sensitive: false` option means this custom logic can be avoided.

### Detail

This Pull Request changes `ActiveModel::Validations::ComparisonValidator` to allow `:case_sensitive` to be supplied as an option. This option effectively defaults to true, which means the current behaviour is retained unless explicitly opting-in, making it a non-breaking change for existing users of the validator.

An `ArgumentError` is raised when the option is supplied for comparisons other than `:equal_to` or `:other_than`. Technically, the option to perform case-insensitive string comparison with `:greater_than`, `:less_than` etc would be possible, but I don't personally have a use-case for this (and I'd imagine that's exceptionally rare) – happy to implement this in the PR if this is desirable though.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/validations/comparison.rb
+++ b/activemodel/lib/active_model/validations/comparison.rb
@@ -14,6 +14,10 @@ module ActiveModel
           raise ArgumentError, "Expected one of :greater_than, :greater_than_or_equal_to, "\
           ":equal_to, :less_than, :less_than_or_equal_to, or :other_than option to be supplied."
         end
+
+        if options.key?(:case_sensitive) && options.keys.intersect?(COMPARE_CHECKS.keys.excluding(:equal_to, :other_than))
+          raise ArgumentError, ":case_sensitive can only be supplied with :equal_to and :other_than"
+        end
       end
 
       def validate_each(record, attr_name, value)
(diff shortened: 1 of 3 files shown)


activemodel/lib/active_model/validations/comparison.rb
```diff
@@ -22,6 +26,16 @@ def validate_each(record, attr_name, value)
...
+          elsif false == options[:case_sensitive] && option == :equal_to
+            unless value.casecmp?(option_value)
+              record.errors.add(attr_name, option, **error_options(value, option_value))
+            end
+            return
+          elsif false == options[:case_sensitive] && option == :other_than
+            if value.casecmp?(option_value)
+              record.errors.add(attr_name, option, **error_options(value, option_value))
+            end
+            return
```
ryantownsend, line 38: If an alternative approach would be preferred here, I'd welcome suggestions.

I don't _love_ this current code. It works and I think it's clear enough but for such a simple addition, it seems like more verbose than necessary.

The early-returns are necessary to avoid subsequent execution of the case-sensitive check on line 41 below.

MatheusRich:
I feel somewhat uneasy about this because it's very string related and this validator is more generic than that.

ryantownsend:
@MatheusRich I hear you, I don’t really know of a better option though.

Introducing a whole extra validator just for case insensitivity seems overkill.

Plus case insensitivity lives as an option on the uniqueness validator so it makes sense to be an option elsewhere too so it’s consistent.

Any suggestions?

ryantownsend:
@MatheusRich I’ve thought on it some more… one idea I do actually have is introducing a `String#case_insensitive` method, similar to `Hash#with_indifferent_access`, whereby non-strict equality checks are overridden such that if the object and the argument are both `String` objects and either are case-insensitive then it uses `casecmp?` instead 🤔

_Update: having looked into just how many methods could need overriding (e.g. `start_with?`, `gsub`, `replace`, `rindex`... the list goes on), this might be a significant undertaking to maintain compatibility over time... may not be a good idea for that reason alone_

MatheusRich:
> Plus case insensitivity lives as an option on the uniqueness validator so it makes sense to be an option elsewhere too so it’s consistent.

Good point. Leave it be.

PR #55366, opened 2025-07-18
Fix enum conflict message grammar.

### LINKED ISSUE #55364: Incorrect grammar in enum conflict message
The enum conflict message uses the wrong indefinite article when referring to instance methods (e.g. "**a** instance method" instead of "**an** instance method").

I'll open a PR to fix this in a moment.

### Steps to reproduce

```ruby
  enum :logger, [ :value ]  # raises `ArgumentError` with message:
                            # "...this will generate a instance method "logger"..."
end
```

### Expected behavior

The message should tell "an instance method".

### Actual behavior

The message tells "a instance method".

### System configuration

**Rails version**: `main`

**Ruby version**: 3.4.4

callmesangio:
Format the error message so that the correct article is used according to the method type (e.g. "an instance method", "a class method").

~Fallback to omitting the method type altogether if unknown, producing the generic wording "a method".~

Fixes #55364

### Motivation / Background

This Pull Request has been created because the grammar in the enum conflict message is slightly off.

### Detail

This Pull Request changes how the message is formatted to fix the grammar issue.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/enum.rb
+++ b/activerecord/lib/active_record/enum.rb
@@ -383,7 +383,7 @@ module ActiveRecord
 
       ENUM_CONFLICT_MESSAGE = \
         "You tried to define an enum named \"%{enum}\" on the model \"%{klass}\", but " \
-        "this will generate a %{type} method \"%{method}\", which is already defined " \
+        "this will generate %{type} method \"%{method}\", which is already defined " \
         "by %{source}."
       private_constant :ENUM_CONFLICT_MESSAGE
 
@@ -405,10 +405,18 @@ module ActiveRecord
         raise ArgumentError, ENUM_CONFLICT_MESSAGE % {
           enum: enum_name,
           klass: name,
-          type: type,
+          type: type_sentence(type),
           method: method_name,
           source: source
         }
+      end
+
+      def type_sentence(type)
+        case type
+        when "instance" then "an #{type}"
+        when "class"    then "a #{type}"
+        else "a"
+        end
       end
 
       def detect_negative_enum_conditions!(method_names)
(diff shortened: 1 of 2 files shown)


skipkayhil (approved):

callmesangio:
Hi @skipkayhil, thanks for the approval.

I was thinking that another (and maybe simpler) approach to this would be invoking `raise_conflict_error()` with the right article passed in the same string to the `type` kwarg. It would save the `type_sentence` method.

Any thoughts? Thanks!

byroot:
> invoking `raise_conflict_error()` with the right article passed in the same string to the `type` kwarg. It would save the `type_sentence` method.

Yes, please do that.

byroot:
Merged as 5afb3068f4a33128e2db5f1d2f503b9738459360 because I couldn't push in your fork.

callmesangio:
> Merged as [5afb306](https://github.com/rails/rails/commit/5afb3068f4a33128e2db5f1d2f503b9738459360) because I couldn't push in your fork.

Hi @byroot, I was going to work on this today, but thanks anyway for stepping in.

Just a couple of notes:

1. the commit message (specifically, the "fallback" part) does not reflect the actual change that has been merged, because it still refers to the first implementation;
2. along the same lines, the added test case made sense for the first proposal, less so for the merged one: since passing the right string is now `detect_enum_conflict!`'s responsibility, the coverage added by the test is currently incomplete. I think we should either a) assert all invocations of `raise_conflict_error` (which I think could be done by extending the regex in existing test cases, instead of adding a new one) or b) delete the new test altogether and go without (as it was the case before this change, since the added value is minimal).


I can surely volunteer for a follow-up PR to address number 2a or 2b, if you can provide a preferred direction.

byroot:
@callmesangio this isn't a big deal. It can stay as is.

PR #50507, opened 2023-12-31
Add Brakeman by default to new apps

vipulnsward:
Add Brakeman by default to all new apps

It can be skipped with the `--skip-brakeman` flag.

Closes #50501

PR #55375, opened 2025-07-21
Raise explicit error on null byte in SQLite string quoting

### LINKED ISSUE #55373: ActiveRecord does not escape NUL bytes for SQLite3
### Steps to reproduce

1. Try to insert something that contains a NUL byte in a string field, in a SQLite3 database
2. It fails with an "unrecognised token" error instead of either working or failing with an appropriate error

Example here: https://asciinema.org/a/QFuSdpD8LWlGLfnsVSgJtu5fD

Example steps are:

```
rails new jestest
cd jestest
rails generate model TestModel foo:string
rails db:migrate
rails console
TestModel.create!(foo:"jes\0test")
```

I tried to reopen #947 but I don't have permission so I've created a new ticket.

### Expected behavior

Either there should be a sensible exception thrown that tells you that it can't do the SQL statement because of a string containing a NUL byte (analogous to what happens if you try to insert bad utf8), or it should somehow manage to insert a string with a NUL byte in it without truncating it (unsure if that is possible with SQLite).

### Actual behavior

The program crashes with an error about an unrecognised token in the generated SQL statement.

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.4.5

himanshukale36:
### Motivation / Background

This Pull Request fixes a long-standing bug in the SQLite adapter where inserting strings containing a NUL (`\0`) byte would result in a confusing SQL error like:

SQLite3::SQLException: unrecognized token: "'jes"

The underlying issue was that SQLite does not support NUL bytes in quoted strings, and Rails did not validate this before generating SQL.

Fixes [#947](https://github.com/rails/rails/issues/947) (original issue that was closed)  
Closes [#55373](https://github.com/rails/rails/issues/55373) (new issue created)

---

### Detail

This PR updates `quote_string` in the SQLite3 adapter to raise an explicit `ArgumentError` if a string contains a null byte (`\x00`), avoiding a confusing syntax error at the database level.

A regression test was added in `activerecord/test/cases/adapters/sqlite3/quoting_test.rb` to ensure this behavior is maintained.

Example now raises:

```ruby
3.4.5 :001 > TestModel.create!(foo:"jes\0test")
(jestest):1:in '<main>': string contains null byte (ArgumentError)

            raise ArgumentError, "string contains null byte"
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Detail
This Pull Request changes the `quote_string` method in the SQLite3 adapter to explicitly raise an `ArgumentError` when a null byte (`\x00`) is present in a string. This prevents SQLite from attempting to parse an invalid query and raising a vague `unrecognized token` error.

A corresponding regression test has been added in `quoting_test.rb` to ensure this behavior remains enforced.

### Additional information
This behavior aligns with how Rails already handles other invalid string inputs, such as bad UTF-8 sequences. By failing fast with an `ArgumentError`, it improves developer experience and avoids low-level, cryptic SQLite errors.

This issue was previously discussed in [#947](https://github.com/rails/rails/issues/947), but could not be reopened. A new issue has been created and linked (see #50507). This change ensures better consistency across adapters in how invalid data is handled before reaching the database.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3/quoting.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3/quoting.rb
@@ -64,6 +64,9 @@ module ActiveRecord
         end
 
         def quote_string(s)
+          if s.include?("\x00")
+            raise ArgumentError, "string contains null byte"
+          end
           ::SQLite3::Database.quote(s)
         end
 
--- a/activerecord/test/cases/adapters/sqlite3/quoting_test.rb
+++ b/activerecord/test/cases/adapters/sqlite3/quoting_test.rb
@@ -14,6 +14,13 @@ class SQLite3QuotingTest < ActiveRecord::SQLite3TestCase
     assert_equal "''", @conn.quote_string("'")
   end
 
+  def test_quote_string_with_null_byte_raises
+    error = assert_raises(ArgumentError) do
+      @conn.quote_string("foo\x00bar")
+    end
+    assert_match(/null byte/, error.message)
+  end
+
   def test_quote_column_name
     [@conn, @conn.class].each do |adapter|
       assert_equal '"foo"', adapter.quote_column_name("foo")


activerecord/lib/active_record/connection_adapters/sqlite3/quoting.rb
```diff
@@ -64,6 +64,9 @@ def quote(value) # :nodoc:
         end
 
         def quote_string(s)
+          if s.include?("\x00")
+            raise ArgumentError, "string contains null byte"
+          end
           ::SQLite3::Database.quote(s)
```
byroot, line 70: Sounds more like something we should change in sqlite3-ruby. FYI @flavorjones

byroot: Also the sqlite3 documentation isn't that clear cut: https://www.sqlite.org/nulinstr.html

TL;DR; it says `NULL` in string columns somewhat works but has lots of short comings, however you can still send them in blobs, and I fear `quote_string` is used for blobs.

himanshukale36: Thanks for the feedback @byroot! 🙏

You're absolutely right — I did notice SQLite accepts `\0` in blobs, and the [SQLite NUL documentation](https://www.sqlite.org/nulinstr.html) does mention partial support in text fields, but with undefined behavior.

I added the check in `quote_string` specifically to prevent obscure `SQLite3::SQLException: unrecognized token` errors when a `\0` is used in a string column — which seems to be a surprisingly common developer trap ([#947](https://github.com/rails/rails/issues/947), [#55373](https://github.com/rails/rails/issues/55373)).

If you feel this would be better handled in `sqlite3-ruby`, I’m happy to raise the issue (or even PR) there. Would you prefer this check moved out of Rails entirely, or scoped only to strings in `ActiveRecord::Quoting`?

Thanks again for the guidance!

byroot: I think the first step is to explore if we can distinguish the blob quoting vs the string quoting. Once we have that, ideally this check is better done in sqlite3-ruby because it's in C and is already iterating over the whole string, so the overhead would be minimal, whereas `String#include?` is more costly, and also have to deal with encodings etc.

himanshukale36: Thanks @byroot — that makes sense.

I’ll dig into where `quote_string` is used to see if there’s a clean way to distinguish string quoting from blob quoting.

If it turns out blobs go through the same path, then I agree that this check is better suited for the `sqlite3-ruby` gem — and I’ll follow up with a PR or issue there since the C layer can handle the check with minimal overhead.

Appreciate the direction!

byroot: > explore if we can distinguish the blob quoting vs the string quoting.

So for blobs we call `quoted_binary`: https://github.com/rails/rails/blob/f49d7a0d5a67eccb8d7017c38ac157a5244a7526/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L82

And for SQLite3 we don't redefine it, it use the default implementation: https://github.com/rails/rails/blob/f49d7a0d5a67eccb8d7017c38ac157a5244a7526/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L196-L198

So we could indeed distinguish the two by defining `quoted_binary` in `Sqlite3Adapter`.

byroot: >  sqlite3-ruby because it's in C and is already iterating over the whole string, so the overhead would be minimal

Actually, I misremembered, it's implemented as a `String#gsub`: https://github.com/sparklemotion/sqlite3-ruby/blob/0cfccfc91c93afb11769dc7b830cb5d7ae2addcc/lib/sqlite3/database.rb#L114

himanshukale36: Just to clarify the behavior after testing:

For binary (blob) values containing null bytes like "jes\0test", SQLite correctly inserts the value. It uses X'...' quoting, and \x00 is valid here.

For string (text) values, inserting a string with a null byte fails with the following error:

```ruby
Invalid string value: contains null byte (\x00). Consider using a binary column if this is intentional.
```
So, I’m only adding the check in sqlite3-ruby to raise an error when a null byte is used in a regular string. Blob values are left as-is, since they are valid in SQLite and don't cause problems.

Let me know if you’d prefer to block blobs with null bytes too, but that doesn't seem necessary at this point.

himanshukale36: Regarding the suggestion to handle this in the sqlite3-ruby gem instead:

While it’s true that ideally null byte validation could be done deeper in sqlite3-ruby (and some of it is already handled in C during parameter binding), not all cases pass through those C layers. For example, ActiveRecord often uses raw SQL string quoting (e.g., for delete_all, where("foo LIKE ?", ...), etc.) which ends up calling the Ruby-level quote method.

This means that unless we patch both the Ruby quoting logic and the C-level bind_param logic, null bytes can still cause confusing SQLite token errors.

Therefore, handling this consistently at the Ruby level (within quote) ensures null bytes are proactively caught across all SQL construction paths, regardless of whether binding is used or not.

Of course, it would be great to have unified handling in sqlite3-ruby, and if desired, I can upstream a follow-up patch there too. But for Rails, this ensures safety immediately and prevents silent corruption or confusing runtime errors.

byroot: To clarify, `sqlite3-ruby` doesn't have the context to know if it's escaping for a string or a blob, at least not now.

So if done there (I'd wait on @flavorjones opinion first), it might need to take a parameter or something.

As for Rails, as mentioned we generally know if we're dealing with a string or a blob, so it's not a problem.

himanshukale36: Thanks for the clarification!

You're absolutely right — sqlite3-ruby currently lacks the context to distinguish between strings and blobs during quoting, so a fix there might indeed require adding an explicit parameter or changing the interface, which could be more invasive.

For now, handling this in Rails feels like the most practical and least disruptive solution since Rails does have the necessary type context, and this helps prevent obscure SQLite errors in real-world usage (e.g., during where, delete_all, etc.).

flavorjones:
I want to suggest that this is something we don't need to fix.

In the description above, you say:

> This behavior aligns with how Rails already handles other invalid string inputs, such as bad UTF-8 sequences.

but I don't think this is a totally accurate statement, because I don't see Rails doing any sort of explicit input encoding validation.

For example, if I inject invalid UTF-8 into the use case from #55373, Rails raises an exception, but that exception comes from the logging subsystem, not any sort of input validation:

```
$ rails new rails-encoding-error
$ cd rails-encoding-error
$ rails generate model TestModel foo:string
$ rails db:migrate
$ rails console
>> invalid_utf8_bytes = [0xE9].pack("C*")
>> invalid_encoded_string = invalid_utf8_bytes.force_encoding("UTF-8")
>> TestModel.create!(foo: invalid_encoded_string)

Could not log "sql.active_record" event. ArgumentError: invalid byte sequence in UTF-8
"gems/activerecord-8.0.2/lib/active_record/log_subscriber.rb:90:in 'Regexp#==='",
"gems/activerecord-8.0.2/lib/active_record/log_subscriber.rb:90:in 'ActiveRecord::LogSubscriber#sql_color'",
"gems/activerecord-8.0.2/lib/active_record/log_subscriber.rb:54:in 'ActiveRecord::LogSubscriber#sql'",
"gems/activesupport-8.0.2/lib/active_support/subscriber.rb:138:in 'ActiveSupport::Subscriber#call'",
"gems/activesupport-8.0.2/lib/active_support/log_subscriber.rb:147:in 'ActiveSupport::LogSubscriber#call'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:177:in 'block in ActiveSupport::Notifications::Fanout::EventObjectGroup#finish'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:24:in 'block in ActiveSupport::Notifications::FanoutIteration#iterate_guarding_exceptions'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:23:in 'Array#each'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:23:in 'ActiveSupport::Notifications::FanoutIteration#iterate_guarding_exceptions'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:121:in 'ActiveSupport::Notifications::Fanout::BaseGroup#each'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:176:in 'ActiveSupport::Notifications::Fanout::EventObjectGroup#finish'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:258:in 'block in ActiveSupport::Notifications::Fanout::Handle#finish_with_values'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:24:in 'block in ActiveSupport::Notifications::FanoutIteration#iterate_guarding_exceptions'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:23:in 'Array#each'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:23:in 'ActiveSupport::Notifications::FanoutIteration#iterate_guarding_exceptions'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:257:in 'ActiveSupport::Notifications::Fanout::Handle#finish_with_values'",
"gems/activesupport-8.0.2/lib/active_support/notifications/fanout.rb:250:in 'ActiveSupport::Notifications::Fanout::Handle#finish'",
"gems/activesupport-8.0.2/lib/active_support/notifications/instrumenter.rb:64:in 'ActiveSupport::Notifications::Instrumenter#instrument'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract_adapter.rb:1135:in 'ActiveRecord::ConnectionAdapters::AbstractAdapter#log'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:554:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#raw_execute'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:591:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#internal_execute'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:547:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#internal_exec_query'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:159:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#exec_insert'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:197:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#insert'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/query_cache.rb:27:in 'ActiveRecord::ConnectionAdapters::AbstractAdapter#insert'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:257:in 'ActiveRecord::Persistence::ClassMethods#_insert_record'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:926:in 'block in ActiveRecord::Persistence#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:412:in 'ActiveRecord::ConnectionAdapters::ConnectionPool#with_connection'",
"gems/activerecord-8.0.2/lib/active_record/connection_handling.rb:310:in 'ActiveRecord::ConnectionHandling#with_connection'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:923:in 'ActiveRecord::Persistence#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/counter_cache.rb:201:in 'ActiveRecord::CounterCache#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/locking/optimistic.rb:84:in 'ActiveRecord::Locking::Optimistic#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/encryption/encryptable_record.rb:184:in 'ActiveRecord::Encryption::EncryptableRecord#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/attribute_methods/dirty.rb:240:in 'ActiveRecord::AttributeMethods::Dirty#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/callbacks.rb:445:in 'block in ActiveRecord::Callbacks#_create_record'",
"gems/activesupport-8.0.2/lib/active_support/callbacks.rb:100:in 'ActiveSupport::Callbacks#run_callbacks'",
"gems/activesupport-8.0.2/lib/active_support/callbacks.rb:913:in 'ActiveRecord::Base#_run_create_callbacks'",
"gems/activerecord-8.0.2/lib/active_record/callbacks.rb:445:in 'ActiveRecord::Callbacks#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/timestamp.rb:116:in 'ActiveRecord::Timestamp#_create_record'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:894:in 'ActiveRecord::Persistence#create_or_update'",
"gems/activerecord-8.0.2/lib/active_record/callbacks.rb:441:in 'block in ActiveRecord::Callbacks#create_or_update'",
"gems/activesupport-8.0.2/lib/active_support/callbacks.rb:100:in 'ActiveSupport::Callbacks#run_callbacks'",
"gems/activesupport-8.0.2/lib/active_support/callbacks.rb:913:in 'ActiveRecord::Base#_run_save_callbacks'",
"gems/activerecord-8.0.2/lib/active_record/callbacks.rb:441:in 'ActiveRecord::Callbacks#create_or_update'",
"gems/activerecord-8.0.2/lib/active_record/timestamp.rb:127:in 'ActiveRecord::Timestamp#create_or_update'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:424:in 'ActiveRecord::Persistence#save!'",
"gems/activerecord-8.0.2/lib/active_record/validations.rb:54:in 'ActiveRecord::Validations#save!'",
"gems/activerecord-8.0.2/lib/active_record/transactions.rb:365:in 'block in ActiveRecord::Transactions#save!'",
"gems/activerecord-8.0.2/lib/active_record/transactions.rb:417:in 'block (2 levels) in ActiveRecord::Transactions#with_transaction_returning_status'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/transaction.rb:626:in 'block in ActiveRecord::ConnectionAdapters::TransactionManager#within_new_transaction'",
"gems/activesupport-8.0.2/lib/active_support/concurrency/null_lock.rb:9:in 'ActiveSupport::Concurrency::NullLock#synchronize'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/transaction.rb:623:in 'ActiveRecord::ConnectionAdapters::TransactionManager#within_new_transaction'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:367:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#within_new_transaction'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/database_statements.rb:359:in 'ActiveRecord::ConnectionAdapters::DatabaseStatements#transaction'",
"gems/activerecord-8.0.2/lib/active_record/transactions.rb:413:in 'block in ActiveRecord::Transactions#with_transaction_returning_status'",
"gems/activerecord-8.0.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:418:in 'ActiveRecord::ConnectionAdapters::ConnectionPool#with_connection'",
"gems/activerecord-8.0.2/lib/active_record/connection_handling.rb:310:in 'ActiveRecord::ConnectionHandling#with_connection'",
"gems/activerecord-8.0.2/lib/active_record/transactions.rb:409:in 'ActiveRecord::Transactions#with_transaction_returning_status'",
"gems/activerecord-8.0.2/lib/active_record/transactions.rb:365:in 'ActiveRecord::Transactions#save!'",
"gems/activerecord-8.0.2/lib/active_record/suppressor.rb:56:in 'ActiveRecord::Suppressor#save!'",
"gems/activerecord-8.0.2/lib/active_record/persistence.rb:55:in 'ActiveRecord::Persistence::ClassMethods#create!'",
```

In a comment above you say:

> which seems to be a surprisingly common developer trap (https://github.com/rails/rails/issues/947, https://github.com/rails/rails/issues/55373).

and honestly the fact that there have been 45,000 issues and PRs in the intervening time tells me the opposite -- this is _not_ a common problem, and iterating over every string to do input validation and check for NULL bytes may not have much benefit.

As a maintainer of the sqlite3-ruby gem, I can't recall any complaints about this -- and couldn't find any issues when I searched.

Obviously it's up to the Rails core team to decide whether this is worth fixing, but I wanted to at least suggest we consider not fixing it.

himanshukale36:
@byroot So what should we do now?

byroot:
Alright, let's leave it as is.

kyrylo:
I just ran into this bug myself. I know it’s marked as a “won’t fix”, but I still wanted to comment for historical context and to show that people do encounter it in the wild.

Tweet for reference: https://x.com/kyrylo/status/1991436143301415388

PR #55374, opened 2025-07-20
Fix key collision bug in HashWithIndifferentAccess#transform_keys

postpostmodern:
### Motivation / Background

This Pull Request has been created because `HashWithIndifferentAccess#transform_keys` contains a bug that can cause key collisions.

### Detail

This Pull Request changes the implementation of `HashWithIndifferentAccess#transform_keys` to transform all keys in one atomic action instead of altering each one per-iteration.

#### Example with Symbol Keys:

Where Hash will give you this:

```ruby
sym_hash = { a: "one", aa: "two", aaa: "three", aaaa: "four" }
sym_hash.transform_keys { |k| "#{k}a".to_sym }
  #=> {:aa=>"one", :aaa=>"two", :aaaa=>"three", :aaaaa=>"four"}
```

Whereas, HashWithIndifferentAccess was returning this:

```ruby
sym_hwia.transform_keys { |k| "#{k}a".to_sym }
  #=> {"aaaaa"=>"one"}
```

#### Example with Integer Keys:

Where Hash will give you this:

```ruby
int_hash = {2=>"A", 3=>"B", 4=>"C", 5=>"D"}
int_hash.transform_keys { |k| k + 1 }
  #=> {3=>"A", 4=>"B", 5=>"C", 6=>"D"}
```

Whereas, HashWithIndifferentAccess was returning this:

```ruby
int_hwia.transform_keys { |k| k + 1 }
  #=> {6=>"A"}
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55376, opened 2025-07-21
Fix HashWithIndifferentAccess#transform_keys! to avoid collisions

byroot:
`transform_keys!` can't directly mutate the receiver otherwise if the key transformation cause some collision it will lose some keys.

Fix: https://github.com/rails/rails/pull/55374

### DIFF
--- a/activesupport/lib/active_support/hash_with_indifferent_access.rb
+++ b/activesupport/lib/active_support/hash_with_indifferent_access.rb
@@ -354,21 +354,26 @@ module ActiveSupport
     NOT_GIVEN = Object.new # :nodoc:
 
     def transform_keys(hash = NOT_GIVEN, &block)
-      return to_enum(:transform_keys) if NOT_GIVEN.equal?(hash) && !block_given?
-      dup.tap { |h| h.transform_keys!(hash, &block) }
+      if NOT_GIVEN.equal?(hash)
+        if block_given?
+          self.class.new(super(&block))
+        else
+          to_enum(:transform_keys)
+        end
+      else
+        self.class.new(super)
+      end
     end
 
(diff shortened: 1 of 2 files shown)


hmcguire-shopify:
Edit: Actually, while this appears to have made `HWIA#transform_keys` behave like `Hash#transform_keys`, it broke `#transform_keys!` handling of `default`/`default_proc`... I'll work on a fix

---

FYI this changed behavior related to `default`:

```ruby
require "active_support"
require "active_support/core_ext/hash/indifferent_access"
require "minitest/autorun"

  def test_hash_transform_keys_keeps_default
    hash_with_default = Hash.new { |h, k| h[k] = :a }

    hash_with_default[:b] = :c

    transformed_hash = hash_with_default.transform_keys(&:to_s)

    assert_equal :c, transformed_hash["b"]
    assert_nil transformed_hash[:d]

    assert_nil transformed_hash.default_proc
  end

  def test_hwia_transform_keys_keeps_default
    hwia_with_default = Hash.new { |h, k| h[k] = :a }.with_indifferent_access

    hwia_with_default[:b] = :c

    transformed_hash = hwia_with_default.transform_keys(&:to_s)

    assert_equal :c, transformed_hash["b"]
    assert_nil transformed_hash[:d]

    assert_nil transformed_hash.default_proc
  end
end
```

Before this change, a `default`/`default_proc` is kept post-`transform_keys` because of the `dup`. Now `HWIA` behaves like `Hash`, which does not keep the `default_proc`/`default` post-`transform_keys`.

I think behaving like `Hash` is "more correct" but figured I'd point it out.

byroot:
Yes it's a mistake.

PR #53320, opened 2024-10-15
Load schema when running db:migrate on fresh dbs

mkasztelnik:
### Motivation / Background

Loading database schemas before `db:migrate` for a single database was fixed in #52830. However, this approach fails when multiple databases are defined. This fix addresses the issue by moving schema loading to occur before initiating database migrations.

### Detail

`db:migrate` has different logic when single or multiple databases are defined (https://github.com/rails/rails/blob/main/activerecord/lib/active_record/railties/databases.rake#L92-L102). When multiple DBs are defined, `migration` method is invoked only for the DBs with pending migrations. The fix moves DBs schema load for new DBs before DBs migrations are invoked.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55377, opened 2025-07-21
[8-0-stable] [Fix #55372] Allow per-database schema format in database config

### LINKED ISSUE #55372: `solid_*` gems were unable to migrate the database properly if `config.active_record.schema_format = :sql` set.
### Steps to reproduce

1. Set `config.active_record.schema_format = :sql` in `application.rb` since we have some stored procedure in our postgres database. Without setting this, the dumped `schema.rb` might be incorrent.
2. since Ruby on Rails 8.0.0.beta1, schema would be loaded first on fresh database. See [https://github.com/rails/rails/blob/8-0-stable/activerecord/CHANGELOG.md#rails-800beta1-september-26-2024](https://github.com/rails/rails/blob/8-0-stable/activerecord/CHANGELOG.md#rails-800beta1-september-26-2024)
3. When installing `solid_*` gems, the generated schema does not include `queue_structure.sql`, `cache_structure.sql`, `cable_structure.sql` files, the database was unable to get migrated properly.

### Expected behavior

The database schema should be created.

### Actual behavior

It leaves an empty database schema.

I don't really think generating the `sql` schema is possible, maybe generating all migrate files like how rails engine does would be a better idea comparing to a single schema file.

Generating 

### System configuration

**Rails version**: 

8.0

**Ruby version**: 

3.4.4

dsh0416:
### Motivation / Background

With the release of Rails 8.0.0.beta1 and onward, a significant change was introduced where the database schema is now loaded first on a fresh database (#53320).

#55372 shows a potential case that when an existing Rails app migrates from the `:ruby` schema format to `:sql` (often to leverage advanced DB features), and then introduces one of the `solid_*` gems, the generation of `queue_structure.sql`, `cable_structure.sql`, and `cache_structure.sql` files in development can either be silently skipped or produce empty files. As a result, these broken or empty schema files might be committed and pushed to the repository. When another team member or a CI/CD pipeline later attempts to set up the database on a new server from these files, critical tables and structures are missing or incomplete. This introduces a real risk of hidden production bugs, corrupted state, or failed deployments. The process of generating proper SQL schema files is database adapter dependent and cannot be robustly solved at the generator level.

#53666 introduces a crucial configuration, allowing schema_format to be set per database in `config/database.yml`. This lets users keep the `:ruby` format for the `solid_*` databases while using `:sql` for their app where needed. This is a simple, targeted, and backwards-compatible solution that prevents the regression scenario outright, with no effect on those not using the feature.

Normally, adding new configurations are treated as new features, and do not need to be backported. But for users having  existing workflows for apps relying on both `:sql` schema format and the `solid_*` gems, this could achieve a real, user-harmful regression.

While the patch is merged, it does not affect users not leveraging these gems or the new configuration, the backport restores the expected and sensible behavior for schema management when using `solid_*` gems in combination with the SQL schema format. This will foster stability and confidence for both existing and new users as they migrate to Rails 8.

### Detail

This Pull Request cherry picked from commit 3081fc9ddd69f54cace2d2533d9e98f44e344338 and 3f177286a7fc385920da28bf9e8d9298f683bab2, and modified to resolve conflicts.

### Checklist


dsh0416:
Wait, the patch causes extra regressions, fixing.

zzak:
We don't backport features to stable branches, or am I missing something?

dsh0416:
> We don't backport features to stable branches, or am I missing something?

You are correct, this is where I am also concerning about.

The patch is about to add a new configuration which is a new feature, but the change restores expected behavior and prevents harmful side effects—such as broken schema management with SQL formats and `solid_*` gems—that could otherwise disrupt teams upgrading to Rails 8. Since the patch is both backwards-compatible and only affects users who opt in to use this config, I am not sure if we could treat it as a regression fix than a new feature, justifying its backport to seamless migrations.

My current workaround to this case is to create a new rails project with `solid_*` gems, migrate it, set the `schema_format` globally to `:sql`, redump the schema, and then carefully copy the structure file back and modifies it to make it work with existing project. I am not sure if this is a proper workaround till Rails 8.1.

zzak:
> The patch is about to add a new configuration which is a new feature

But this is my point, we do not backport features, even backwards compatible ones. To avoid risk of introducing regressions inside a release.

You will have to wait until a release is made.

That is assuming someone on core decides otherwise, which is possible, I'm only trying to set realistic expectations based on my experience.

PR #55357, opened 2025-07-18
Rename PasswordsMailer to PasswordMailer

dwightwatson:
### Motivation / Background

The [`PasswordsMailer`](https://github.com/rails/rails/blob/e5e1ce7114387ff170242cdc7a8e2905e30ccccb/railties/lib/rails/generators/rails/authentication/templates/app/mailers/passwords_mailer.rb.tt) that is generated with the authentication scaffold is named plurally.

However, when generating a resource scaffold or [when reading the `ActionMailer` docs](https://guides.rubyonrails.org/action_mailer_basics.html#creating-a-mailer-and-views) the naming is singular.

### Detail

This pull request renames the mailer, preview and view files from plural to singular.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55378, opened 2025-07-21
Update mailers to be plural

dwightwatson:
This Pull Request has been created as an alternative to #55357 where I proposed the authentication generator should create `PasswordsMailer` instead of `PasswordMailer`. It was suggested from someone the Bruno at Basecamp that mailers should in fact be plural.

For as long as I've been doing Rails I've always used singular names for mailers, and in fact my team has all done the same too. [And I suspect it's simply due to the fact that the ActionMailer docs primarily use singular names](https://guides.rubyonrails.org/action_mailer_basics.html).

I'm proposing to update the guides to use plural mailers to better indicate that this should be the convention. It aligns better with plural controllers, helpers and views.

### Detail

This PR makes these changes:

* All references to `UserMailer` to become `UsersMailer`
* A reference to an `AdminMailer` becomes `AdminsMailer`
* A reference to a `NotifierMailer` becomes `NotifiersMailer`

### Checklist

### DIFF
--- a/guides/source/action_mailer_basics.md
+++ b/guides/source/action_mailer_basics.md
@@ -681,7 +681,7 @@ end
 ```
 
 The above will render the HTML part using the `my_layout.html.erb` file and the
-text part with the usual `welcome_email.text.erb` file.
+text part with the usual `users_mailer.text.erb` file.
 
 [`layout`]:
     https://api.rubyonrails.org/classes/ActionView/Layouts/ClassMethods.html#method-i-layout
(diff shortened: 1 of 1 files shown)


p8:
Hmm, I'm not sure about this.
The ActionMailer API also uses singular form: https://edgeapi.rubyonrails.org/classes/ActionMailer/Base.html
But maybe that needs to be updated as well?

rafaelfranca:
@dhh any options here? It is your design. I find that being in singular makes more sense.

dwightwatson:
I'll wait for guidance on whether to also update the `ActionMailer::Base` docs, or alternatively re-open #55357.

PR #55381, opened 2025-07-21
Add missing TypeScript definitions to DirectUpload in @rails/activestorage

KirbyBrown:
### Motivation / Background

This Pull Request has been created because the `@rails/activestorage` package does not currently include TypeScript definitions that reflect the actual runtime behavior of `DirectUpload`. 

For example, `DirectUpload` supports an optional `delegate` parameter, assigns `xhr` and `uploadRequest` dynamically, and allows access to upload progress events like `onprogress`. None of these are typed, which means TypeScript users either have to cast unsafely or cannot access features that clearly exist in the implementation.

Adding accurate definitions directly to the package improves developer experience and avoids the need for a separate `@types/activestorage` fallback.

### Detail

This Pull Request adds a new `activestorage.d.ts` file in `app/javascript/activestorage/` that describes the public and runtime-visible properties of the `DirectUpload` class. It also updates the `package.json` file to expose these types using the `types` field so that they are available when the package is consumed via npm.

These types are based directly on the current `direct_upload.js` implementation. Properties like `xhr` and `uploadRequest` are dynamically assigned during `create()`, and delegate hooks like `directUploadWillCreateBlobWithXHR` are supported but undocumented. This change brings the types in line with how the library actually behaves.

### Example usage with progress tracking

```ts
import { DirectUpload } from "@rails/activestorage";

const upload = new DirectUpload(file, "/rails/active_storage/direct_uploads");

upload.create((error, blob) => {
  if (error) {
    console.error("Upload failed", error);
  } else {
    console.log("Upload successful", blob?.signed_id);
  }
});

// Add progress listener after create() assigns xhr
// This PR also adds type support for the preferred pattern:
// using a delegate with directUploadWillStoreFileWithXHR to access xhr safely
if (upload.xhr?.upload) {
  upload.xhr.upload.onprogress = (event) => {
    if (event.lengthComputable) {
      const percent = (event.loaded / event.total) * 100;
      console.log(`Upload progress: ${percent.toFixed(1)}%`);
    }
  };
}
```

### Additional information

The activestorage.d.ts file will be copied to the src/ directory on publish via the existing prepublishOnly script. This means consumers of @rails/activestorage will automatically get these type definitions without needing to install a separate @types package.

These type definitions are already in use in production at [Langiri](https://langiri.com), a site for language learners that uses `DirectUpload` for applicant and contributor video uploads. We’ve verified both the delegate pattern and progress tracking through real usage.

### Checklist

* [] Tests are added or updated if you fix a bug or add a feature.
_N/A: this change adds TypeScript type definitions only, with no behavior changes._
* [] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.
_N/A: no behavior changes or user-facing API changes._

### DIFF
--- a/activestorage/app/javascript/activestorage/activestorage.d.ts
+++ b/activestorage/app/javascript/activestorage/activestorage.d.ts
@@ -1,0 +1,49 @@
+// activestorage/app/javascript/activestorage/activestorage.d.ts
+
+declare module "@rails/activestorage" {
+    export class DirectUpload {
+      constructor(file: File, url: string, delegate?: DirectUploadDelegate, customHeaders?: Record<string, string>);
+      create(callback: (error: Error | null, blob?: Blob) => void): void;
+  
+      // Core properties
+      id: number;
+      file: File;
+      url: string;
+      delegate?: DirectUploadDelegate;
+      customHeaders: Record<string, string>; // e.g., x-amz-* headers for direct S3 upload
+  
+      // Dynamically assigned after create() is called
+      xhr?: XMLHttpRequest;
+      uploadRequest?: XMLHttpRequest;
+  
+      // Optional hook assignable directly
+      directUploadWillCreateBlobWithXHR?: (xhr: XMLHttpRequest) => void;
+    }
+  
+    export class DirectUploadController {
+      constructor(input: HTMLInputElement, file: File);
+      start(callback: (error: Error | null) => void): void;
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thank you for the pull request but we don't have interest on maintaining typescript files.

PR #55387, opened 2025-07-22
Avoid dynamic encrypting in generated fixtures

cassiascheffer:
### Motivation / Background

The current fixture generation for password attributes uses dynamic BCrypt encryption with ERB tags (`<%= BCrypt::Password.create("secret") %>`), which causes fixtures to regenerate different encrypted values each time they're loaded. 

### Detail

This Pull Request changes the fixture generation template to:
- Generate the BCrypt password hash once at fixture creation time instead of dynamically during each test run
- Add a comment showing how the encrypted value was generated (`# Generated with BCrypt::Password.create("secret")`)
- Update the corresponding test to verify that the generated fixture contains a valid BCrypt hash

### Additional information

The change affects the Rails generator template for model fixtures when using `has_secure_password` or similar password digest attributes. This is a minor improvement to the developer experience and test performance.

### Checklist

Before submitting the PR, make sure the following are checked:

### DIFF
--- a/railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
+++ b/railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
@@ -4,7 +4,7 @@
 <%= name %>:
 <% attributes.each do |attribute| -%>
   <%- if attribute.password_digest? -%>
-  password_digest: <%%= BCrypt::Password.create("secret") %>
+  password_digest: <%= BCrypt::Password.create("secret") %>  # Generated with BCrypt::Password.create("secret")
   <%- elsif attribute.reference? -%>
   <%= yaml_key_value(attribute.column_name.delete_suffix("_id"), attribute.default || name) %>
   <%- elsif !attribute.virtual? -%>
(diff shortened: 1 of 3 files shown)


railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
```diff
@@ -4,7 +4,8 @@
 <%= name %>:
 <% attributes.each do |attribute| -%>
   <%- if attribute.password_digest? -%>
-  password_digest: <%%= BCrypt::Password.create("secret") %>
+  # Generated with BCrypt::Password.create("secret")
+  password_digest: <%= BCrypt::Password.create("secret") %>
```
byroot: You might need some escaping to not conflict with YAML here (not sure what's the range of characters used by bcrypt). A good way to do that is `.to_json`.

byroot: Actually nevermind, It seems like it's not using any of the characters that could cause issues.

railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
```diff
@@ -4,7 +4,7 @@
 <%= name %>:
 <% attributes.each do |attribute| -%>
   <%- if attribute.password_digest? -%>
-  password_digest: <%%= BCrypt::Password.create("secret") %>
+  password_digest: <%= BCrypt::Password.create("secret") %>  # Generated with BCrypt::Password.create("secret")
```
byroot, line 7: Nitpick, I moved the comment on the same line.

PR #55388, opened 2025-07-22
Respect `schema_search_path` on `rails dbconsole` for Postgres

sobrinho:
### Motivation / Background

This Pull Request has been created because I want `rails dbconsole` to respect the `schema_search_path` from `config/database.yml`.

### Detail

This Pull Request will add the `search_path` to `PGOPTIONS` when set before calling `psql`.

### Additional information

This will keep parity between the `psql` connection and the connection used by the rails app.

If you have a custom search path set to lets say `myschema`, the application will be running with that search path but `rails dbconsole` will open with search path as `$user,public` instead.

So to workaround you have to remember to call `SET search_path TO myschema` on the psql session to have the same connection configuration.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Respect `schema_search_path` on `rails dbconsole` for PostgreSQL.
+
+    *Gabriel Sobrinho*
+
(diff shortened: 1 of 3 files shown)


sobrinho:
@rafaelfranca this is an easy one to review :)

PR #55389, opened 2025-07-23
Log when rescue_from handles an exception

swebb:
### Motivation / Background

In development when a `before_action` renders or redirects this error message appears on the logs:

> Filter chain halted as :before_action rendered or redirected

I find this message very helpful. However, if a before_action raises an error that is handled by a rescue_from block, nothing is logged indicating an error occurred.

[Original proposal](https://discuss.rubyonrails.org/t/feature-proposal-log-when-before-action-raises-an-exception/89414)

### Detail

This PR adds a log message indicating an exception was rescued, and displays where it was originally raised. For example:

> rescue_from handled exception HomeController::MyError (Oh no!) -
>   /home/vscode/my-test-app/app/controllers/home_controller.rb:15:in
>   'HomeController#authenticate'

### Checklist

### DIFF
--- a/activesupport/lib/active_support/rescuable.rb
+++ b/activesupport/lib/active_support/rescuable.rb
@@ -92,6 +92,7 @@ module ActiveSupport
 
         if handler = handler_for_rescue(exception, object: object)
           handler.call exception
+          logger.info("rescue_from handled exception #{exception.class} (#{exception.message}) - #{exception.backtrace.first}")
           exception
         elsif exception
           if visited_exceptions.include?(exception.cause)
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/rescuable.rb
```diff
@@ -92,6 +92,7 @@ def rescue_with_handler(exception, object: self, visited_exceptions: [])
 
         if handler = handler_for_rescue(exception, object: object)
           handler.call exception
+          logger.info("rescue_from handled exception #{exception.class} (#{exception.message}) - #{exception.backtrace.first}")
```
byroot, line 95: I think the feature makes sense, but this isn't the proper way to implement it.

The log should be in `actionpack/lib/action_controller/log_subscriber.rb`, and here it should just fire an `ActiveSupport::Notification`.

You can look at `def halted_callback_hook` for inspiration.

swebb: Thanks @byroot. I'll check it out.

PR #55390, opened 2025-07-23
Introduce Hash#deep_compact to core_ext

amitleshed:
### Summary

Introduce `Hash#deep_compact` (non‑destructive) and `Hash#deep_compact!` (destructive) to recursively remove nil (and optionally blank) values from nested hashes and arrays.

Usage
```ruby
h = { a: 1, b: nil, c: { d: nil, e: 2 }, f: [1, nil, { g: nil, h: 3 }] }
h.deep_compact
# => { a: 1, c: { e: 2 }, f: [1, { h: 3 }] }

h.deep_compact(blank: true)
# removes blank? values (""/[], {}) as well
```

API
```ruby
Hash#deep_compact(blank: false)  -> new_hash
Hash#deep_compact!(blank: false) -> self
```

Motivation

- Common need when preparing JSON/API payloads or normalizing params.
- Rails ships Hash#compact, but not a recursive variant; everyone rewrites this.
- Aligns with other deep helpers (deep_transform_keys, deep_merge) in ActiveSupport.

### DIFF
--- a/activesupport/lib/active_support/core_ext/hash/deep_compact.rb
+++ b/activesupport/lib/active_support/core_ext/hash/deep_compact.rb
@@ -1,49 +1,46 @@
 class Hash
-  def deep_compact(blank: false)
-    dup.deep_compact!(blank: blank)
+  def deep_compact(remove_blank: false)
+    dup.deep_compact!(remove_blank: remove_blank)
   end
 
-  def deep_compact!(blank: false)
-    each_key do |k|
-      v = self[k]
+  def deep_compact!(remove_blank: false)
+    should_prune = lambda do |obj|
+      return true  if obj.nil?
+      return false unless remove_blank
 
-      case v
-      when Hash
-        v.deep_compact!(blank: blank)
-        delete(k) if blank ? v.blank? : v.nil?
-      when Array
-        v.map! do |elem|
-          if elem.is_a?(Hash)
-            elem.deep_compact!(blank: blank)
-            elem
-          elsif elem.is_a?(Array)
-            deep_compact_array(elem, blank)
-          else
-            elem
-          end
-        end
-        v.reject! { |elem| blank ? elem.blank? : elem.nil? }
-        delete(k) if blank ? v.blank? : v.nil?
-      else
-        delete(k) if blank ? v.blank? : v.nil?
+      case obj
+      when String       then obj.strip.empty?
+      when Array, Hash  then obj.empty?
+      else                   false
       end
     end
+
+    each_key.to_a.each do |key|
+      value = self[key]
+
+      case value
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/core_ext/hash/deep_compact.rb
```diff
@@ -0,0 +1,49 @@
...
+          else
+            elem
+          end
+        end
+        v.reject! { |elem| blank ? elem.blank? : elem.nil? }
+        delete(k) if blank ? v.blank? : v.nil?
+      else
+        delete(k) if blank ? v.blank? : v.nil?
+      end
+    end
```
sobrinho: You can probably remove the duplication here since it applies to all cases:

```suggestion
      case v
      when Hash
        v.deep_compact!(blank: blank)
        
      when Array
        v.map! do |elem|
          if elem.is_a?(Hash)
            elem.deep_compact!(blank: blank)
            elem
          elsif elem.is_a?(Array)
            deep_compact_array(elem, blank)
          else
            elem
          end
        end
        v.reject! { |elem| blank ? elem.blank? : elem.nil? }
      end
    end
    delete(k) if blank ? v.blank? : v.nil?
```

amitleshed: Thanks, I've done some refactoring. Feel free to check it out again.

activesupport/lib/active_support/core_ext/hash/deep_compact.rb
```diff
@@ -0,0 +1,46 @@
+class Hash
+  def deep_compact(remove_blank: false)
```
MatheusRich: Rails has compact_blank. IMO this should be deep_compact_blank (or to add two methods, instead of a boolean arg)

amitleshed: Good catch

amitleshed:
@sobrinho @MatheusRich 
Appreciate the engagement guys.

amitleshed:
@rafaelfranca 
Hey guys!
Appreciate all the efforts and feedback.
At this point, I would like to move this forward. Please let me know if there's anything else you'll need on my side!

skipkayhil:
Thank you for the pull request, but new core extensions require a very strong justification (and general applicability) that I don't think this meets.

amitleshed:
@skipkayhil 
I get it.
Do you have any docs of what criterion new core extensions should meet?
I'd love to try again soon, thanks.

skipkayhil:
> Do you have any docs of what criterion new core extensions should meet? I'd love to try again soon, thanks.

If you want to add methods to Ruby's classes, you should propose those changes to Ruby itself. I would not spend too much time on trying to think of core extensions to contribute.

PR #55392, opened 2025-07-23
Fix kill_connection_from_server for PG test

skipkayhil:
and also fix incorrect comment for different test method

This method is intended to make a subsequent query hit the rescue inside
`with_raw_connection`, but this was only actually happening with
`mysql2` and `trilogy`. The `pg` adapter was not being correctly put
into a state where the `SELECT` in the test would need to retry.

This commit fixes the issue by using pg_terminate_backend instead of
pg_cancel_backend. Now when the test is run with `pg` it correctly hits
the rescue in `with_raw_connection`.

### DIFF
--- a/activerecord/test/cases/adapter_test.rb
+++ b/activerecord/test/cases/adapter_test.rb
@@ -882,18 +882,7 @@ module ActiveRecord
       end
 
       test "#execute is retryable" do
-        conn_id = case @connection.adapter_name
-                  when "Mysql2"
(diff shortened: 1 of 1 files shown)


matthewd:
> Now when the test is run with `pg` it correctly hits the rescue in `with_raw_connection`.

Can we somehow improve the test so it fails if `kill_connection_from_server` doesn't do its job? (Ironically, perhaps by re-extracting the "get server-internal connection ID" behaviour, and verifying it's changed after the expected retry? 😅)

PR #55398, opened 2025-07-24
feat: ability to provide a custom error handler to MemCacheStore

nickamorim:
### Motivation / Background

### Detail

### Checklist


zzak:
It's an interesting idea, one thing to note is that typically we wouldn't implement this type of feature for only a single adapter, so it has to make sense for all of the Active Support Cache stores.

The other idea I had was to rather pass in the error handler, to wrap the caller in something like [rescue_from](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html#method-i-rescue_from) that way your business logic to deal with these errors can belong to your application.

nickamorim:
This is effectively a lift + shift from RedisCacheStore

PR #55400, opened 2025-07-25
Add support for plain button rendering using button_to with form: false

uurcank:
## Summary

This PR adds support for passing `form: false` attribute in button_to to render a plain `<button>` without generating a form: 

`<%= button_to "Delete", "#", form: false %>`

## Motivation / Background

I ran into while working to convert my app to use [Shopify Polaris Web Components](https://shopify.dev/docs/api/app-home/using-polaris-components) inside a Rails application. Shopify’s UI system expects buttons to be rendered as plain `<button>` tags in certain places and their JavaScript APIs are designed to interact with button elements only in those areas.

While [button_tag](https://api.rubyonrails.org/v8.0/classes/ActionView/Helpers/FormTagHelper.html#method-i-button_tag) exists, replacing button_to with it often requires additional refactoring.

For example if i want to convert this code into a plain button:

`<%= button_to "Delete", post_path(post), method: :delete, data: { turbo_confirm: "Are you sure?" } %>`

with form false i could simply do 

`<%= button_to "Delete", form: false, post_path(post), method: :delete, data: { turbo_confirm: "Are you sure?" } %>`

But now using button tag i also need to add type: button:

`
<%= button_tag "Delete", type: "button", data: { action: "click->posts#destroy" } %>`

This makes the conversion more verbose and mechanical than it needs to be. Supporting `form: false` provides a more intuitive and streamlined way to disable form behavior while keeping the same button_to interface.

If I ever need to revert away from Polaris Web Components, I would otherwise have to refactor all button tags back to use button_to. With this approach, I can simply remove `form: false` and regain the original behavior without changing helpers or rewriting templates.

If this is perceived as a reasonable addition, it also makes sense for Rails to enforce some security-related constraints in this context. This PR proposes that when `form: false` is used, options like `method:,` `type: "submit"`, and any behavior that would generate hidden `<input>` tags (such as params: or authenticity tokens) are disallowed or ignored by default to prevent misuse and preserve the intent of the button_to helper.

## Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/url_helper.rb
+++ b/actionview/lib/action_view/helpers/url_helper.rb
@@ -207,9 +207,21 @@ module ActionView
         content_tag("a", name || url, html_options, &block)
       end
 
-      # Generates a form containing a single button that submits to the URL created
-      # by the set of +options+. This is the safest method to ensure links that
-      # cause changes to your data are not triggered by search bots or accelerators.
+      # Generates a plain button or a form containing a single button that submits to the URL created
+      # by the set of +options+.
+      #
+      # Passing <tt>form: false</tt> in +html_options+ will render only
+      # a <button> element without generating a <form>. This is useful when you want
+      # a button for JavaScript-driven interactions or UI-only behavior.
+      #
+      # Using a form is the safest method to ensure links that
+      # cause changes to your data are not triggered by search bots, browser prefetching,
+      # or other automated agents.
+      #
+      # Note that when <tt>form: false</tt> is used, options like <tt>:method</tt>, <tt>:params</tt>,
+      # <tt>:authenticity_token</tt>, and form-specific attributes are ignored. Additionally,
+      # the button's <tt>type</tt> is forced to <tt>"button"</tt> to avoid accidental submission
+      # in layouts that may contain nested forms.
       #
       # You can control the form and button behavior with +html_options+. Most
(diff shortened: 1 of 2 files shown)


MatheusRich:
> with form false i could simply do
> 
> <%= button_to "Delete", form: false, post_path(post), method: :delete, data: { turbo_confirm: "Are you sure?" } %>

What would be the point of a button to without a form? I'm sorry, I'm not following why use `button_to` when you don't want many of the options to work.

uurcank:
This change only makes button_to more flexible in modern apps that use JavaScript, Turbo, or Shadow DOM (like Shopify Polaris). Instead of rewriting helpers or switching to button_tag, you can simply toggle form: false to adapt behavior as needed.

It provides a low-friction, reversible path, especially useful in Shopify development, where some actions need to interact with App Bridge (e.g., through JavaScript handlers), while others still rely on traditional Rails form submissions. This lets you incrementally adapt without rewriting.

MatheusRich:
Gotcha. TY for the clarification!

skipkayhil:
Thank you for the pull request, but this change doesn't make sense to me. If you want the behavior of `button_tag` you should use `button_tag`.

uurcank:
This is definitely already covered by button_tag. It could be a good api to have 'form: false' because button tag is type submit by default, just helps with progressive enhancement.

PR #55402, opened 2025-07-25
[Fix #55379] Default branch in github ci templates

### LINKED ISSUE #55379: Git default branch is not specified in the github actions
### Steps to reproduce

```ruby
cat ~/.gitconfig
[init]
    defaultBranch = master
...

rails new demo-app
...

cat demo-app/.github/workflows/ci.yml
name: CI

on:
  pull_request:
  push:
    branches: [ main ]
...
```
### Expected behavior

Having the default branch specified on the gitconfig (git config init.defaultBranch)

### Actual behavior

A static value defined [here](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt#L6)

### System configuration

**Rails version**: Rails 8.0.2

**Ruby version**: ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-linux]

endenis:
### Motivation / Background

Fixes #55379

When generating a new rails app or a plugin, the config templates for GitHub workflow CI contain hardcoded "main" branch name.
Instead, they should take the default branch name from the `git config init.defaultbranch` if it is specified.

### Detail

1. I used the existing method `user_default_branch` inside template files to fetch the name of the default git branch.
2. The `user_default_branch` method was modified so that it uses `capture_command` inside. This way it does not crash on an error (if, for example, git is not installed) and there is no need to call `.strip` on each usage.
3. I changed the existing tests `test_version_control_initializes_git_repo_with_user_default_branch` to also verify the branch name in the generated CI config files.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -529,8 +529,13 @@ module Rails
         using_js_runtime? && %w[bun].include?(options[:javascript])
       end
 
-      def capture_command(command, pattern)
-        `#{command}`[pattern]
+      def capture_command(command, pattern = nil)
+        output = `#{command}`
+        if pattern
+          output[pattern]
+        else
+          output
+        end
       rescue SystemCallError
         nil
       end
@@ -760,11 +765,13 @@ module Rails
       end
 
       def user_default_branch
-        @user_default_branch ||= `git config init.defaultbranch`
+        @user_default_branch ||= capture_command("git config init.defaultbranch").strip.presence || "main"
       end
 
       def git_init_command
-        return "git init" if user_default_branch.strip.present?
+        if capture_command("git config init.defaultbranch").present?
+          return "git init"
+        end
 
         git_version = `git --version`[/\d+\.\d+\.\d+/]
 
--- a/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [ <%= user_default_branch %> ]
 
 jobs:
 <%- unless skip_brakeman? -%>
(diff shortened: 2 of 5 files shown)


skipkayhil (approved):

railties/lib/rails/generators/app_base.rb
```diff
@@ -760,11 +760,11 @@ def keep_file(destination)
       end
 
       def user_default_branch
-        @user_default_branch ||= `git config init.defaultbranch`
+        @user_default_branch ||= `git config init.defaultbranch`.strip
```
Edouard-chin: I'm not sure how frequent this is, but if a user doesn't have git installed and runs `rails new my_app --skip-git` this will now crash the process.

skipkayhil: That's interesting, maybe `--skip-git` should imply `--skip-ci`

endenis: Interesting! Thank you for your comments.
In a new commit I added a begin-rescue block to cover this case.

skipkayhil: ~Maybe an even easier solution: `$default-branch`~

~https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/~

Nevermind, this appears to be a "workflow template" level feature

railties/lib/rails/generators/app_base.rb
```diff
@@ -760,11 +760,16 @@ def keep_file(destination)
       end
 
       def user_default_branch
-        @user_default_branch ||= `git config init.defaultbranch`
+        @user_default_branch ||=
```
Edouard-chin:
```suggestion
        @user_default_branch ||= capture_command('git config init.defaultbranch', /.*/)
```

Edouard-chin (commented):
> That's interesting, maybe --skip-git should imply --skip-ci

That's a good point. I think the fix in this case is simple enough to continue generating a github CI config.
Either way, it seems a bit more robust to rescue from system calls and not fail the whole `rails new` just to get the user's default branch.

railties/lib/rails/generators/app_base.rb
endenis: Thank you, I've applied this change

Edouard-chin (approved):

PR #55403, opened 2025-07-25
Robustify routing parameters against objects that don't respond to empty?

PaceHeartLater:
The optimized URL helpers call `empty?` without checking `respond_to?(:empty?)` first, causing NoMethodError for objects that don't implement the method.

Includes a test that fails before this fix and succeeds after.

# Fix routing parameter `empty?` method bug

## Problem

The Rails routing system's optimized URL helpers have a bug where they call `empty?` on parameter objects without first checking if the object responds to that method. This causes `NoMethodError` exceptions when using objects that don't implement `empty?` as URL parameters.

## What This PR Does

This PR adds a test that demonstrates the bug by:
- Creating a test object that doesn't implement `empty?`
- Attempting to use it as a URL parameter
- Showing that the current code fails with `NoMethodError`

The test fails before this fix and succeeds after.

## Impact

This bug affects any application using custom objects as URL parameters where those objects don't implement `empty?`. The fix is backward compatible and doesn't change behavior for objects that do implement `empty?`.

### DIFF
--- a/actionpack/lib/action_dispatch/routing/route_set.rb
+++ b/actionpack/lib/action_dispatch/routing/route_set.rb
@@ -249,7 +249,7 @@ module ActionDispatch
                 @arg_size.times { |i|
                   key = @required_parts[i]
                   value = args[i].to_param
-                  yield key if value.nil? || value.empty?
+                  yield key if value.nil? || (value.respond_to?(:empty?) && value.empty?)
                   params[key] = value
                 }
                 params
--- a/actionpack/lib/action_dispatch/routing/route_set.rb
+++ b/actionpack/lib/action_dispatch/routing/route_set.rb
@@ -249,7 +249,7 @@ module ActionDispatch
                 @arg_size.times { |i|
                   key = @required_parts[i]
                   value = args[i].to_param
-                  yield key if value.nil? || (value.respond_to?(:empty?) && value.empty?)
+                  yield key if value.nil? || value.blank?
                   params[key] = value
                 }
                 params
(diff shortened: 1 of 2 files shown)


actionpack/lib/action_dispatch/routing/route_set.rb
```diff
@@ -249,7 +249,7 @@ def parameterize_args(args)
                 @arg_size.times { |i|
                   key = @required_parts[i]
                   value = args[i].to_param
-                  yield key if value.nil? || value.empty?
+                  yield key if value.nil? || (value.respond_to?(:empty?) && value.empty?)
```
skipkayhil:
```suggestion
                  yield key if value.blank?
```

Could this be replaced with `blank?` instead (which should be defined on everything)?

PaceHeartLater: Absolutely! Much more elegant. <3

actionpack/test/dispatch/routing_test.rb
```diff
@@ -3912,6 +3912,31 @@ def test_routes_with_double_colon
...
+    end
+
+    non_emptyable_object = non_emptyable_class.new
+
+    begin
+      @app.routes.url_helpers.user_path(non_emptyable_object)
+      assert true, "URL helper correctly handled object without empty? method"
+    rescue NoMethodError => e
+      flunk e.message
+    end
```
skipkayhil: Can you make this a positive assertion instead of a negative one? I'm having trouble understanding what you expect the generated link to be if `to_param` doesn't return a String.

PaceHeartLater: By default, I think it should be the `to_param`'s string representation. Test updated!

PR #52547, opened 2024-08-08
Add ago/in words to time_ago_in_words

n-studio:
### Motivation / Background

I recently encountered the issue https://github.com/rails/mission_control-jobs/issues/136 on mission control and I realized that Rails didn't have a way to localize the word `ago` out of the box.

~This PR will break a lot of projects but I think it's a valuable add, and the update is pretty straightforward.~

EDIT: added `config.action_view.include_prefix_or_suffix_to_time_ago_in_words` to keep retro-compatibility.

### Details

Before:

```ruby
#   time_ago_in_words(3.minutes.from_now)                 # => 3 minutes
#   time_ago_in_words(3.minutes.ago)                      # => 3 minutes
```

After:

```ruby
#   time_ago_in_words(3.minutes.from_now)                 # => in 3 minutes
#   time_ago_in_words(3.minutes.ago)                      # => 3 minutes ago
```

Example of localization:

```yml
ja:
  datetime:
    distance_in_words:
      ago: "%{time}前"
      from_now: "%{time}後"
es:
  datetime:
    distance_in_words:
      ago: "hace %{time}"
      from_now: "en %{time}"
fr:
  datetime:
    distance_in_words:
      ago: "il y a %{time}"
      from_now: "dans %{time}"
```

PR #55405, opened 2025-07-25
Add `relative_time_in_words` helper to ActionView

MatheusRich:
### Motivation / Background

A similar proposal was made in https://github.com/rails/rails/pull/52547. The need for the feature is here, but I proposed a new method not to complicate `time_ago_in_words` and also because returning `"in <future time>"` from a helper called `time_ago` doesn't make sense.

### Detail

This Pull Request adds a new helper `relative_time_in_words` to ActionView.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/date_helper.rb
+++ b/actionview/lib/action_view/helpers/date_helper.rb
@@ -186,6 +186,23 @@ module ActionView
 
       alias_method :distance_of_time_in_words_to_now, :time_ago_in_words
 
+      # Like <tt>time_ago_in_words</tt>, but adds a prefix/suffix depending on whether the time is in the past or future.
+      # You can use the <tt>scope</tt> option to customize the translation scope. All other options
+      # are forwarded to <tt>time_ago_in_words</tt>.
+      #
+      #   relative_time_in_words(3.minutes.from_now) # => "in 3 minutes"
+      #   relative_time_in_words(3.minutes.ago) # => "3 minutes ago"
+      #   relative_time_in_words(10.seconds.ago, include_seconds: true) # => "less than 10 seconds ago"
+      #
+      # See also #time_ago_in_words
+      def relative_time_in_words(from_time, options = {})
+        now = Time.now
+        time = distance_of_time_in_words(from_time, now, options.except(:scope))
+        key = from_time > now ? :future : :past
+
+        I18n.t(key, time: time, scope: options.fetch(:scope, :'datetime.relative'), locale: options[:locale])
+      end
+
       # Returns a set of select tags (one for year, month, and day) pre-selected for accessing a specified date-based
       # attribute (identified by +method+) on an object assigned to the template (identified by +object+).
       #
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Introduce `relative_time_in_words` helper
+
+    ```ruby
+    relative_time_in_words(3.minutes.from_now) # => "in 3 minutes"
+    relative_time_in_words(3.minutes.ago) # => "3 minutes ago"
+    relative_time_in_words(10.seconds.ago, include_seconds: true) # => "less than 10 seconds ago"
+    ```
+
+    *Matheus Richard*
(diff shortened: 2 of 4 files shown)


actionview/lib/action_view/helpers/date_helper.rb
```diff
@@ -186,6 +186,23 @@ def time_ago_in_words(from_time, options = {})
...
+      #   relative_time_in_words(3.minutes.ago) # => "3 minutes ago"
+      #   relative_time_in_words(10.seconds.ago, include_seconds: true) # => "less than 10 seconds ago"
+      #
+      # See also #time_ago_in_words
+      def relative_time_in_words(from_time, options = {})
+        options = { scope: :'datetime.relative' }.merge(options)
+        time = time_ago_in_words(from_time, options.except(:scope))
+        key = from_time > Time.now ? :future : :past
+
+        I18n.t(key, time: time, scope: options[:scope], locale: options[:locale])
```
byroot:
```suggestion
        time = time_ago_in_words(from_time, options)
        key = from_time > Time.now ? :future : :past

        I18n.t(
          key,
          time: time,
          scope: options.fetch(:scope, :'datetime.relative'),
          locale: options[:locale],
        )
```

Is it really important not to forward the `scope` options? Because I think this could be streamlined.

byroot: Ah, I see you essentially followed the same pattern.

byroot: I simplified the implementation a little bit anyway.

MatheusRich: Both this method and the time_ago one have `scope`s, so they'll use the same key if I forward. I'd love being able to pass a scope for each, but idk if there's a pattern for that already.

byroot: Yes, that's what I figured, so I left the `.except`.

MatheusRich: We could use add `distance_of_time_options` option or `relative_time_scope` to disambiguate them. LMK if any of those solutions appeal to you and I can submit another PR.

byroot: I think it's really niche, so probably not worth it.

adrianthedev:
Love it! Thanks!

actionview/lib/action_view/helpers/date_helper.rb
A1bi: It's actually not that niche. In the German language the output of `relative_time_in_words` is grammatically incorrect, because it uses the default translation for `distance_of_time_in_words` (e.g. _"vor ein Jahr"_ and _"vor drei Jahre"_ should be _"vor einem Jahr"_ and _"vor drei Jahren"_). You need to have separate translations for `distance_of_time_in_words` if they are used in conjunction with a preposition from `relative_time_in_words`.

You could choose a different phrasing in `relative_time_in_words`'s translation to make it grammatically correct, but nothing really works without sounding weird in German. I can imagine you have the same problem with other languages.

MatheusRich: @A1bi if you're facing this problem, I encourage you to open a PR with my suggestion.

MatheusRich: @A1bi if you're facing this problem, I encourage you to open a PR with my suggestion.

A1bi: After thinking a little more about this, I came to the conclusion it's probably best if we had completely independent translations for `relative_time_in_words`, because it might not work in all languages to simply have a string interpolation containing the translation of `distance_of_time_in_words`.

I imagine something like this in `en.yml`:

```yml
datetime:
  relative_time_in_words:
    future:
      ...
      x_minutes:
        one:   "in 1 minute"
        other: "in %{count} minutes"
      ...
    past:
      ...
      x_minutes:
        one:   "1 minute ago"
        other: "%{count} minutes ago"
      ...
```

This would obviously introduce a breaking change, so I'm not sure how to go about that. What do you think?

PR #55406, opened 2025-07-25
Allow `postgres` as an alias for `postgresql` in rails generators

aliismayilov:
### Motivation / Background

Recently, I was starting a new project and got the familiar error:
```sh
$ rails new blog --database=postgres
Expected '--database' to be one of mysql, trilogy, postgresql, sqlite3, mariadb-mysql, mariadb-trilogy; got postgres
```

This time instead of cursing at the screen, I took a note to come back to this and submit a PR. It's a small PR, but in the spirit of Ruby it should contribute to the developer happiness.

### Detail

This Pull Request allows `postgres` as an alias for `postgresql` in rails generators.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -286,7 +286,7 @@ module Rails
 
         gem_name, gem_version = database.gem
         GemfileEntry.version gem_name, gem_version,
-          "Use #{options[:database]} as the database for Active Record"
+          "Use #{database.adapter} as the database for Active Record"
       end
 
       def web_server_gemfile_entry # :doc:
@@ -353,7 +353,7 @@ module Rails
       end
 
       def sqlite3? # :doc:
-        !skip_active_record? && options[:database] == "sqlite3"
+        !skip_active_record? && database.adapter == "sqlite3"
       end
 
       def skip_active_record? # :doc:
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Rails 8.1.0.beta1 (September 04, 2025) ##
 
+*   Allow `postgres` as an alias for `postgresql` in rails generators
+
+    *Ali Ismayilov*
+
 *   Add command `rails credentials:fetch PATH` to get the value of a credential from the credentials file.
 
     ```bash
--- a/railties/lib/rails/generators/database.rb
+++ b/railties/lib/rails/generators/database.rb
@@ -3,10 +3,14 @@
 module Rails
   module Generators
     class Database
-      DATABASES = %w( mysql trilogy postgresql sqlite3 mariadb-mysql mariadb-trilogy )
+      DATABASES = %w( mysql trilogy postgresql postgres sqlite3 mariadb-mysql mariadb-trilogy )
 
       module MySQL
         def name
+          "mysql"
+        end
+
+        def adapter
           "mysql"
         end
 
@@ -51,6 +55,10 @@ module Rails
           "mariadb"
         end
 
+        def adapter
+          "mariadb-mysql"
+        end
+
         def port
           3306
         end
@@ -72,7 +80,7 @@ module Rails
         def build(database_name)
           case database_name
           when "mysql" then MySQL2.new
-          when "postgresql" then PostgreSQL.new
+          when "postgresql", "postgres" then PostgreSQL.new
           when "trilogy" then Trilogy.new
           when "sqlite3" then SQLite3.new
           when "mariadb-mysql" then MariaDBMySQL2.new
@@ -93,6 +101,10 @@ module Rails
       end
 
       def name
+        raise NotImplementedError
+      end
+
+      def adapter
         raise NotImplementedError
       end
 
@@ -168,6 +180,10 @@ module Rails
           "postgres"
         end
 
+        def adapter
+          "postgresql"
+        end
+
         def template
           "config/databases/postgresql.yml"
         end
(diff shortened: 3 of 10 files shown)


railties/lib/rails/generators/app_base.rb
```diff
@@ -286,7 +286,7 @@ def database_gemfile_entry # :doc:
 
         gem_name, gem_version = database.gem
         GemfileEntry.version gem_name, gem_version,
-          "Use #{options[:database]} as the database for Active Record"
+          "Use #{database.adapter} as the database for Active Record"
```
MatheusRich, line 289: Why was this needed?

aliismayilov: > Why was this needed?

I believe this line adds a comment in the generated Gemfile, so, I kept it consistent: regardless if you use `postgres` or `postgresql`, it always outputs: Use postgresql as the database for Active Record.

yahonda:
I believe this error message is already clear enough to indicate what needs to be done. Since it only appears when running rails new for the first time, entering the correct database name at that point should be sufficient. I understand that some people refer to PostgreSQL as “Postgres,” but I would prefer not to add this pull request solely to support that alternative naming.

aliismayilov:
> I believe this error message is already clear enough to indicate what needs to be done. Since it only appears when running rails new for the first time, entering the correct database name at that point should be sufficient. I understand that some people refer to PostgreSQL as “Postgres,” but I would prefer not to add this pull request solely to support that alternative naming.

I would like to remind the first pillar: https://rubyonrails.org/doctrine#optimize-for-programmer-happiness.

It's a small feature, but possibly removes the first surprise on your very first project command. Postgres is still the name of the protocol (database url starts with `postgres://`) and it's still pretty common to refer to the project as [postgres](https://en.wikipedia.org/wiki/postgres).

yahonda:
Thank you for the suggestion. I do appreciate the effort behind it, but I think this change would add an unnecessary choice for people who want to use PostgreSQL as their database.

Even if some people call it “Postgres,” the error message you get when using an invalid name with the -d option already makes it clear that postgresql is the correct value to use when creating a Rails app.

If we allow both postgresql and postgres, users may wonder why there are two options and start looking for differences that don’t actually exist. In that sense, it doesn’t really help them and only adds another option to consider.

While I appreciate the intent, I believe it’s better not to introduce extra choices that don’t provide real value.

Cosmo:
What if we accept `postgres`, but show a notice that the generator detected the intent?

Something like:
`Detected '--database=postgres'. Interpreting as '--database=postgresql'.`

"postgres" is the commonly used short form of "PostgreSQL" and many developers naturally default to it.

The form `postgres` also appears frequently in Rails projects. 
For example in `config/database.yml` as the default username and password and in `.devcontainer/compose.yaml` as the service name for PostgreSQL.

railties/lib/rails/generators/app_base.rb
shivabhusal: @aliismayilov I see this was added with a clear purpose, though it seems to have triggered a larger set of changes than expected.

shivabhusal (commented):
I’ve often misspelled postgres/pg/postgre myself and have also wished for an alias like this. 👍🏼 

However, I noticed this change touches quite a few files — maybe we could take a simpler approach.
Please check the diff below; I believe it works as intended.

```diff
diff --git a/railties/lib/rails/generators/app_base.rb 
@@ -39,7 +39,7 @@ def self.add_shared_options_for(name)
                                            desc: "Path to some #{name} template (can be a filesystem path or URL)"
 
         class_option :database,            type: :string, aliases: "-d", default: "sqlite3",
-                                           enum: Database::DATABASES,
-                                           desc: "Preconfigure for selected database"
+                                           enum: Database::DATABASES + Database::ALIAS_OF_DATABASES.keys ,
+                                           desc: "Preconfigure for selected database. Value Aliases: #{Database::ALIAS_OF_DATABASES.keys.join(', ')}"

 
         class_option :skip_git,            type: :boolean, aliases: "-G", default: nil,
@@ -808,7 +808,10 @@ def dockerfile_chown_directories
       end
 
       def database
-        @database ||= Database.build(options[:database])
+        # check if the database option is an alias
+        db_name = Database::ALIAS_OF_DATABASES[options[:database]] || options[:database]
+
+        @database ||= Database.build(db_name)
       end
     end
   end

# diff --git a/railties/lib/rails/generators/database.rb 
@@ -4,6 +4,7 @@ module Rails
   module Generators
     class Database
       DATABASES = %w( mysql trilogy postgresql sqlite3 mariadb-mysql mariadb-trilogy )
+      ALIAS_OF_DATABASES = { 'pg' => 'postgresql', 'postgres' => 'postgresql' }
 
       module MySQL
         def name
~
```

This also automatically adds aliases in help like 
```
  -d,            [--database=DATABASE]                                         # Preconfigure for selected database. Value Aliases: pg, postgres
                                                                               # Default: sqlite3
                                                                               # Possible values: mysql, trilogy, postgresql, sqlite3, mariadb-mysql, mariadb-trilogy, pg, postgres

```

rafaelfranca:
I'm ok with adding an alias, but it should be implemented as an alias, not as a different database or even adapter. @shivabhusal's suggestion seems better

simi:
I'm also not sure this is good idea. Database is called PostgreSQL, Postgres is actually its preceedor. The error mentioned is super nice explaining there is `postgresql` available. Is it really worth to save typing 2 characters? Once the postgres alias starts to spread, there will be no easy way back.

rafaelfranca:
Good point @simi I forgot that there was a previous database called postgres. Thinking about that, I don't think this alias makes sense.

PR #55409, opened 2025-07-26
[ci skip][docs] Remove broken link from CSRF router attack example

yokomaru:
This PR removes a broken link from the CSRF section of the security guide.

### Motivation / Background

The original link (`www.h-online.com`) now redirects to `www.heise.de/en`, but the referenced article is no longer available:

http://www.h-online.com/security/news/item/Symantec-reports-first-active-attack-on-a-DSL-router-735883.html

I initially considered replacing it with a similar article, such as one from Network World. However, based on feedback that the source is not sufficiently authoritative — and that the incident itself dates back nearly 20 years — I decided to remove the link entirely.

Since the key details of the incident are already described in the text, the link was not essential.

### Checklist


zzak:
Do we still need a link to an incident that occurred nearly 20 years ago (2007?), and unless it's from an official or authoritative source, I'd avoid linking to a random blog we might not want to endorse long-term.

I'm not familiar with Network World, but it looks like a for-profit media site rather than an authoritative security reference, which I'd prefer for something in our security guide.

yokomaru:
@zakk

Thank you for the feedback.  
You're absolutely right — there's no strong reason to keep a broken link to a nearly 20-year-old incident, especially when the details are already described in the text itself.

To resolve the broken link issue, I’ve simply removed the outdated link.  
Please have another look when you have a moment.

Edouard-chin:
This security article was written 17 years ago in https://github.com/rails/rails/commit/4705046045dcddad168a43120194ffd8672698f3 , I agree that removing the broken link and keeping the example makes the most sense.

PR #50312, opened 2023-12-09
Implement `button_to` in terms of `form_tag`

seanpdoyle:
### Motivation / Background

While the ergonomics of calling `button_to` match those of `form_for`, `form_with`, or even `form_tag`, the implementation doesn't use much of those helper's existing infrastructure. There are chunks of logic repeated and re-implemented. For example:

* determining the `[action]` and `[method]` attributes for the `<form>`
* rendering an `<input type="hidden" name="_method">` for methods other than `POST` or `GET`
* generating and rendering an authenticity token input
* content exfiltration
* encoding `[data-remote]` for UJS integration

### Detail

This commit implements more of `button_to` in terms of the [form_tag][] helper method defined in `FormTagHelper`. That method handles responsibilities like determining `[method]`, rendering `<input type="hidden">`, etc.

The only outward-facing changes (reflected in test suite cases) involve the order of the `<form>` element's content, and the introduction of an `[accept-charset="UTF-8"]` attribute.

The `<button>` element will be rendered **after** hidden fields instead of **before** (the previous behavior).

Similarly, the `[accept-charset]` attribute wasn't being rendered before, but is rendered by `form_tag`. The `form_tag` helper doesn't currently expose a mechanism to skip that attribute (in the same way that `:enforce_utf8` is being used here to skip the `<input type="hidden">`.

### Additional information

As part of these changes, this commit also excises the `method_tag` and `token_tag` private helpers out of the `UrlHelper` module and into the `FormTagHelper` module, since that's the only remaining call site. The previous implementation relied on the fact that both `UrlHelper` and `FormTagHelper` end up being mixed into an `ActionView::Base` instance, so the cross-module `private` boundary wasn't effective.

In support of that change, this commit includes a `require
"action_view/helpers/tag_helper"` line atop
`actionview/lib/action_view/helpers/url_helper.rb` to ensure the
availability of the `ActionView::Helpers::TagHelper` module methods
(since helpers like `mail_to` invoke helpers such as `content_tag(:a, …)`).

[form_tag]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-form_tag

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
[...]

PR #55410, opened 2025-07-26
Introduce `ActionView::Helper::NavigationHelper`

seanpdoyle:
### Motivation / Background

Related to [#50312][]

Create the `ActionView::Helper::NavigationHelper` to serve as a module for `button_to` and `link_to` (as well as related
`link_to_unless_current`, `link_to_unless`, and `link_to_if`).

### Additional information

As a side-effect of moving those methods, this commit also moves several `private`-scoped methods related to those implementations. Since all `ActionView::Helpers` modules are ultimately mixed-into the `ActionView::Base` class, the `private` method boundaries are largely ignored. Depsite that fact, this commit moves the methods so that they are defined in the same module that they are called.

[#50312]: https://github.com/rails/rails/pull/50312#discussion_r2220564548

### Checklist


actionview/test/template/navigation_helper_test.rb
seanpdoyle, line 1: It felt appropriate to extract test coverage of the `NavigationHelper` module to its own test file.

The side-effect of that extraction is a large diff without any significant changes. If we decide its too disruptive, I'm happy to revert the test changes and continue to cover both `UrlHelper` and `NavigationHelper` in the `actionview/test/template/url_helper_test.rb` file.

skipkayhil:
Would you mind rebasing? I'll merge on green (then we can hopefully get #50312 as well)

p8:
Great to see this getting merged!

PR #55411, opened 2025-07-26
Update action_view_overview.md

lazylester:
Clarifies that a single counter variable is available when instance of different ActiveRecord subclasses are included in the collection. This is the behaviour specified in the test in rails/actionpack/test/controller/render.rb called test_partial_collection_shorthand_with_different_types_of_records

### Motivation / Background

This Pull Request has been created because the current docs suggested a behaviour different from what the test is testing.

### Detail

This Pull Request changes the layout and rendering docs

### Additional information

none

### Checklist

* [Y ] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [ Y] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ N/A] Tests are added or updated if you fix a bug or add a feature.
* [ N/A] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/guides/source/action_view_overview.md
+++ b/guides/source/action_view_overview.md
@@ -524,7 +524,7 @@ view, starting with a value of `0` on the first render.
 This also works when the local variable name is changed using the `as:` option.
 So if you did `as: :item`, the counter variable would be `item_counter`.
 
-As noted above, for collections with instances of different models, Rails chooses the partial appropriate to the model. The counter variable increments for each partial, regardless of the class of the model being rendered.
+NOTE: When rendering collections with instances of different models, the counter variable increments for each partial, regardless of the class of the model being rendered.
 
 Note: The following two sections, [Strict Locals](#strict-locals) and [Local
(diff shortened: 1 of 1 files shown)


guides/source/action_view_overview.md
```diff
@@ -524,6 +524,8 @@ view, starting with a value of `0` on the first render.
 This also works when the local variable name is changed using the `as:` option.
 So if you did `as: :item`, the counter variable would be `item_counter`.
 
+As noted above, for collections with instances of different models, Rails chooses the partial appropriate to the model. The counter variable increments for each partial, regardless of the class of the model being rendered.
```
MatheusRich: I feel like this would be better as a note?
```suggestion
NOTE: When rendering collections with instances of different models, the counter variable increments for each partial, regardless of the class of the model being rendered.
```

MatheusRich (approved):

PR #55412, opened 2025-07-26
`#to_sql` support for calculation and finder methods

joshuay03:
### Motivation / Background

Implements this proposal: https://discuss.rubyonrails.org/t/feature-proposal-to-sql-support-for-calculation-and-finder-methods/89381

cc @byroot

### Detail

This PR enhances `ActiveRecord::Relation#to_sql` to support calculation and finder methods, allowing developers to inspect the SQL that would be generated for complex queries without executing them.

This PR enhances `ActiveRecord::Relation#to_sql` to support calculation and finder methods, allowing developers to inspect the SQL that would be generated for complex queries without executing them.

**Example new capabilities:**
```ruby
> User.where(active: true).to_sql.count
#=> "SELECT COUNT(*) FROM \"users\" WHERE \"users\".\"active\" = 1"

> User.joins(:posts).to_sql.pluck(:name, :email)
#=> "SELECT \"users\".\"name\", \"users\".\"email\" FROM \"users\" INNER JOIN \"posts\" ON \"posts\".\"user_id\" = \"users\".\"id\""

> User.where(role: 'admin').to_sql.exists?
#=> "SELECT 1 AS one FROM \"users\" WHERE \"users\".\"role\" = 'admin' LIMIT 1"
```

**Changes:**
1. Updates `ActiveRecord::Relation#to_sql` to return a proxy object similar to `#explain` so that `ActiveRecord::Calculations` and `ActiveRecord::FinderMethods` can be called on it to return the SQL string for the final query. Missing methods are delegated to the `#to_s` method on the proxy to avoid introducing breaking changes with current code which expects the method to return a `String`.
2. Extracts `ActiveRecord::Relation::ExplainProxy` into its own file, `ActiveRecord::ExplainProxy`.
3. Extends `ExplainProxy` to bring it to parity with `ToSQLProxy` so that both now support all calculation and almost all finder methods (excluding the additional order based methods, e.g. `#second`).

### Additional information

None.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/explain_proxy.rb
+++ b/activerecord/lib/active_record/explain_proxy.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+# :markup: markdown
+
+module ActiveRecord
+  class ExplainProxy # :nodoc:
+    def initialize(relation, options)
+      @relation = relation
+      @options  = options
+    end
+
+    def inspect
+      exec_explain { @relation.send(:exec_queries) }
+    end
+
+    def average(...)
+      exec_explain { @relation.average(...) }
+    end
+
+    def calculate(...)
+      exec_explain { @relation.calculate(...) }
+    end
+
+    def count(...)
+      exec_explain { @relation.count(...) }
+    end
+
+    def exists?(...)
+      exec_explain { @relation.exists?(...) }
+    end
+
+    def find(...)
+      exec_explain { @relation.find(...) }
+    end
+
+    def find_by(...)
+      exec_explain { @relation.find_by(...) }
+    end
+
+    def first(...)
+      exec_explain { @relation.first(...) }
+    end
+
+    def ids(...)
+      exec_explain { @relation.ids(...) }
+    end
+
+    def last(...)
+      exec_explain { @relation.last(...) }
+    end
+
+    def maximum(...)
+      exec_explain { @relation.maximum(...) }
+    end
+
+    def minimum(...)
+      exec_explain { @relation.minimum(...) }
+    end
+
+    def pick(...)
+      exec_explain { @relation.pick(...) }
+    end
+
+    def pluck(...)
+      exec_explain { @relation.pluck(...) }
+    end
+
+    def sum(...)
+      exec_explain { @relation.sum(...) }
+    end
+
+    def take(...)
+      exec_explain { @relation.take(...) }
+    end
+
+    private
+      def exec_explain(&block)
+        @relation.exec_explain(@relation.collecting_queries_for_explain { block.call }, @options)
+      end
+  end
+end
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -3,54 +3,6 @@
 module ActiveRecord
   # = Active Record \Relation
   class Relation
-    class ExplainProxy  # :nodoc:
-      def initialize(relation, options)
-        @relation = relation
-        @options  = options
-      end
-
-      def inspect
-        exec_explain { @relation.send(:exec_queries) }
-      end
-
-      def average(column_name)
-        exec_explain { @relation.average(column_name) }
-      end
-
-      def count(column_name = nil)
-        exec_explain { @relation.count(column_name) }
-      end
-
-      def first(limit = nil)
-        exec_explain { @relation.first(limit) }
-      end
-
-      def last(limit = nil)
-        exec_explain { @relation.last(limit) }
-      end
-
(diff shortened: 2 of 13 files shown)


activerecord/lib/active_record/relation.rb
```diff
@@ -1213,7 +1167,7 @@ def reset
       @future_result&.cancel
       @future_result = nil
       @delegate_to_model = false
-      @to_sql = @arel = @loaded = @should_eager_load = nil
```
joshuay03, line 1223: This seems to be a leftover, I can't find any existing references to it.

activerecord/lib/active_record/explain_proxy.rb
```diff
@@ -0,0 +1,79 @@
...
+      exec_explain { @relation.send(:exec_queries) }
+    end
+
+    def average(column_name)
+      exec_explain { @relation.average(column_name) }
+    end
+
+    def calculate(operation, column_name)
+      exec_explain { @relation.calculate(operation, column_name) }
+    end
```
byroot: Most of these are very similar and may get out of sync if we add new methods on `Relation`, or change some methods signatures.

I'm wondering if it wouldn't be better to just blindly delegate using `method_missing`?

activerecord/lib/active_record/relation.rb
```diff
@@ -65,10 +17,11 @@ def exec_explain(&block)
     VALUE_METHODS = MULTI_VALUE_METHODS + SINGLE_VALUE_METHODS + CLAUSE_METHODS
 
     include Enumerable
-    include FinderMethods, Calculations, SpawnMethods, QueryMethods, Batches, Explain, Delegation
+    include FinderMethods, Calculations, SpawnMethods, QueryMethods, Batches, Explain, ToSQL, Delegation
     include SignedId::RelationMethods, TokenFor::RelationMethods
 
     attr_reader :table, :model, :loaded, :predicate_builder
+    attr_writer :to_sql # :nodoc:
```
byroot, line 25: Can't we make this `protected` instead of `:nodoc:` ?

byroot (commented):
Hum, I'm a bit on the fence about this implementation, that's a whole lot of extra conditionals all over `Relation` methods.

activerecord/lib/active_record/explain_proxy.rb
sobrinho: Looks like `...` could be used everywhere:

```ruby
def calculate(...)
  exec_explain { @relation.calculate(...) }
end
```

Therefore it means `method_missing` would indeed work:

```ruby
def method_missing(method_name, ...)
  if @relation.respond_to?(method_name)
    exec_explain { @relation.public_send(method_name, ...) }
  else
    super
  end
end

def respond_to_missing?(method_name, include_private = false)
  @relation.respond_to?(method_name, include_private) || super
end
```

joshuay03: > I'm wondering if it wouldn't be better to just blindly delegate using method_missing?

I think it would be good to have explicit definitions so things like method introspection with `#methods` works for example, making it easier to quickly check if the proxy supports a method or not.

> Looks like ... could be used everywhere:

I was contemplating using all-argument forwarding, but chose not to because `ExplainProxy` was already defining full signatures. However, I've already split that out into its own file, so might as well use this opportunity to refactor. I'll update both to use `...`.

joshuay03:
> Hum, I'm a bit on the fence about this implementation, that's a whole lot of extra conditionals all over `Relation` methods.

Ya, I didn't realise calculation and finder methods had multiple execution entry points before working on this 😅 

The issue is we need to capture the relation right after it's completely built but before these entry points, since we lose the relation object in some cases downstream - methods like `select_all` aren't guaranteed to receive a relation object, and it didn't seem right to force these methods to always pass one in just to support this feature.

I still think the value outweighs the conditional noise, but I'm open to suggestions if there's a cleaner approach.

activerecord/lib/active_record/explain_proxy.rb
joshuay03: > I'll update both to use `...`.

Done 👍🏽

activerecord/lib/active_record/relation.rb
joshuay03: Not with the current design because `ToSQLProxy` needs to call it on the relation instance in its `exec_to_sql` method: https://github.com/rails/rails/pull/55412/files#diff-30d65db7ab3dea21c1329eaa669adc251e860aa99f3b95607c1d4a7c0a100302R93

PR #55413, opened 2025-07-27
Fix the PostgreSQL version check for pg<1.6

larskanis:
The check was introduced as part of https://github.com/rails/rails/pull/55282 . However the incompatibility is actually due to the changed ABI of libpq of PostgreSQL-18 regarding to PG::Connection#cancel and it's not related to the server version.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -666,8 +666,8 @@ module ActiveRecord
         if database_version < 9_03_00 # < 9.3
           raise "Your version of PostgreSQL (#{database_version}) is too old. Active Record supports PostgreSQL >= 9.3."
         end
-        if database_version >= 18_00_00 && Gem::Version.new(PG::VERSION) < Gem::Version.new("1.6.0")
-          warn "pg gem version #{PG::VERSION} is known to be incompatible with PostgreSQL 18+. Please upgrade to pg 1.6.0 or later."
+        if PG.library_version >= 18_00_00 && Gem::Version.new(PG::VERSION) < Gem::Version.new("1.6.0")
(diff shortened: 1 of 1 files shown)


matthewd:
Oh, thanks for pointing this out!

That means that a previously working application / locked Gemfile, deployed on a newer base OS, could run into this -- even while talking to the same database instance. Hmm, I think I'm feeling less good about our ability to backport this change as a blanket failure case :confused:

@yahonda perhaps we need to re-consider the option of warning on app boot (we can detect this as soon as the adapter is loaded, don't need to wait until we connect), and only raising if we attempt to `cancel`?

yahonda:
I pushed a draft branch to confirm I understand the direction being discussed: https://github.com/rails/rails/tree/add-libpq-cancel-compatibility-check
- Since we can determine the libpq version (PG.library_version) and pg gem version (PG::VERSION) immediately after require "pg" without connecting to the PostgreSQL server, I implemented a warning for incompatible combinations right after requiring "pg".
- I modified the Rails method that calls PG::Connection#cancel (cancel_any_running_query) to raise an exception when encountering incompatible combinations. I'm using NotImplementedError tentatively, though this may not cover all scenarios.
- For tests that call cancel_any_running_query, I made them skip when there's an incompatible combination.

What do you think?

larskanis:
https://github.com/rails/rails/tree/add-libpq-cancel-compatibility-check looks good to me. IMHO that PR could be merged.

yahonda:
@larskanis Thanks for the comment.

This branch is just a proof of concept. The commit messages aren’t properly written yet and it hasn’t been opened as a pull request. I’ll add you as a co-author and clean things up before turning it into a proper pull request

yahonda:
> This branch is just a proof of concept. The commit messages aren’t properly written yet and it hasn’t been opened as a pull request. I’ll add you as a co-author and clean things up before turning it into a proper pull request

Opened #55540

skipkayhil:
Addressed in #55540?

larskanis:
Yes.

PR #53281, opened 2024-10-12
Sort table columns by name when dumping schema

jduff:
### Motivation / Background

In projects when multiple people are creating migrations they can often be run out of order from one another, this results in each developer machine having columns added in a different order. Each time we run migrations the schema is dumped based on our local database base so the schema output has columns out of order from machine to machine resulting in noisy diffs and potentially conflicts in the schema file that make no difference overall. ex:

```
     t.string "county"
-    t.string "country_code"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "country_code"
     t.string "urn"
```

### Detail

This PR adds an option to sort the columns before outputting to the table dump. I found [this previous PR](https://github.com/rails/rails/pull/32111) which does this just for MySQL, but I decided to do this at the base schema dumper level as an option because there could be cases where teams don't want to do this (MySQL being one of them since you can add columns relative to existing columns).

### Additional information

If this approach is accepted I think making a config option to set this would be a good idea and it could be useful to make enabling this the default depending on the database you ere using (ex enabled by default for postgresql and sqlite3, not for mysql).

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #52830, opened 2024-09-07
Load schema when running db:migrate on fresh db

andrewn617:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/52829

When we have a schema file, if we run `db:migrate` before setting up the db, we do not load the schema and then we blow away the contents of the file when we dump the schema after migrations.

You can repro by creating a fresh app and running the following commands:

```
bin/rails g model Post
bin/rails db:migrate
rm -rf db/migrate
bin/rails db:drop
bin/rails db:migrate
```

After following these steps, the schema file will be empty and the posts table will not exist.

### Detail

We should make sure the db is setup before we run migrations. This is what we do in the `db:prepare` task. So, I extracted the code from that task and reused it in the migration task.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #32111, opened 2018-02-26
[ActiveRecord] Schema Dumper Columns Order

Tensho:
### Summary

From project to project written with Rails I spectate the similar snag, when **column definition rows in the schema are shuffled** from time to time (only for `schema.rb` in `:ruby` mode). It's annoying to see the changes in the git index, despite I haven't touched any of them. It's really hard for me to define the steps to reproduce for this bug, but let me put here my thoughts and proposed changes. First consider the next example for better clarity:
```
     t.string "county"
-    t.string "country_code"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "country_code"
     t.string "urn"
```
Here "country_code" column definition changed position even though there was no any related migration code.

First of all I thought this behaviour is connected somehow with database engine version or operating system version divergence between developers in our team. So we setup the same version of Rails 4.2 (later checked on 5.2) and MySQL 5.7.20 on macOS Sierra 10.12.6 at MacBook Pro. But there was no luck – I still spectate schema shuffling.

Then I decided to dig into the schema dumper code to understand how ActiveRecord retrieves columns information and dumps it to the `schema.rb` file. Inside I noticed expected [alphabetical table names sorting](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/schema_dumper.rb#L90):
```
def tables(stream)
  sorted_tables = @connection.tables.sort

   sorted_tables.each do |table_name|
    table(table_name, stream) unless ignored?(table_name)
  end
  ...
```
Then I found [table columns info retrievement](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/schema_dumper.rb#L105):
```
def table(table, stream)
  columns = @connection.columns(table)
...
```
So there is no specific ordering. "Hmmn" 🤔 , – was sparkled in my – "Maybe it's desirable to show columns definition in the same manera as database holds them and don't sort in alphabetical order. But how different databases return schema info and do they comply the result order?". I moved to connection adapters:
• **MySQL** ([`ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter`](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L801))
```
def column_definitions(table_name) # :nodoc:
  execute_and_free("SHOW FULL FIELDS FROM #{quote_table_name(table_name)}", "SCHEMA") do |result|
    each_hash(result)
  end
end
```
[...]

PR #55414, opened 2025-07-28
Add optional setting to `SchemaDumper` to disable sorting of table columns

mackuba:
### Motivation / Background

Pull request https://github.com/rails/rails/pull/53281, merged in January, has changed the behavior of `ActiveRecord::SchemaDumper` to print table columns inside a `create_table` block in `schema.db` sorted alphabetically by name, instead of the actual order in the database.

The rationale was that when multiple people are adding migrations to the same table in parallel, they can end up with different order of the columns in their local databases, which then generate different versions of the schema, causing confusing diffs or merge conflicts. Putting columns in alphabetical order makes the order deterministic.

That is true, but it's also true that the order is now deterministically *different* than the actual order in the database… and sometimes, the order of the columns matters to some degree (otherwise, why have an `:after` / `:before` options in migrations at all?).

This PR intends to add a way to restore the behavior present in AR 8.0 and earlier, while keeping the new behavior as a default.

### Detail

The PR adds a class field `sort_table_columns` in `ActiveRecord::SchemaDumper`, defaulting to true. If set to false, the table columns are printed in original order instead of sorted by name.

### Additional information

I've added one additional test and checked the tests with `rake test:sqlite3`.

Alternative name I've considered was `sort_columns_by_name`, so we could go with that if you prefer.

### DIFF
--- a/activerecord/lib/active_record/schema_dumper.rb
+++ b/activerecord/lib/active_record/schema_dumper.rb
@@ -40,6 +40,11 @@ module ActiveRecord
     # should not be dumped to db/schema.rb.
     cattr_accessor :unique_ignore_pattern, default: /^uniq_rails_[0-9a-f]{10}$/
 
+    ##
+    # :singleton-method:
+    # Specifies if columns in create_table block should be ordered alphabetically by name.
+    cattr_accessor :sort_table_columns, default: true
+
     class << self
       def dump(pool = ActiveRecord::Base.connection_pool, stream = $stdout, config = ActiveRecord::Base)
         pool.with_connection do |connection|
@@ -156,7 +161,8 @@ module ActiveRecord
       end
 
       def table(table, stream)
-        columns = @connection.columns(table)
+        columns = sorted_columns(table)
+
         begin
           self.table_name = table
 
@@ -192,7 +198,7 @@ module ActiveRecord
           tbl.puts ", force: :cascade do |t|"
 
           # then dump all non-primary key columns
-          columns.sort_by(&:name).each do |column|
+          columns.each do |column|
             raise StandardError, "Unknown type '#{column.sql_type}' for column '#{column.name}'" unless @connection.valid_type?(column.type)
             next if column.name == pk
 
@@ -363,6 +369,14 @@ module ActiveRecord
 
       def relation_name(name)
         name
+      end
+
+      def sorted_columns(table)
+        if self.class.sort_table_columns
+          @connection.columns(table).sort_by(&:name)
+        else
+          @connection.columns(table)
+        end
       end
 
       def remove_prefix_and_suffix(table)
--- a/activerecord/test/cases/schema_dumper_test.rb
+++ b/activerecord/test/cases/schema_dumper_test.rb
@@ -175,6 +175,19 @@ class SchemaDumperTest < ActiveRecord::TestCase
     assert_equal %w[account_id client_of description firm_id firm_name name rating status type], column_names
   end
 
+  def test_table_columns_unsorted
+    original_value = ActiveRecord::SchemaDumper.sort_table_columns
+    ActiveRecord::SchemaDumper.sort_table_columns = false
+
+    column_names = column_definition_lines(dump_table_schema("companies")).flatten.filter_map do |line|
+      $1 if line !~ /t\.index/ && line.match(/t\..*"(\w+)"/)
+    end
+
+    assert_equal %w[type firm_id firm_name name client_of rating account_id description status], column_names
+  ensure
+    ActiveRecord::SchemaDumper.sort_table_columns = original_value
+  end
+
   def test_schema_dumps_index_columns_in_right_order
     index_definition = dump_table_schema("companies").split(/\n/).grep(/t\.index.*company_index/).first.strip
     if current_adapter?(:Mysql2Adapter, :TrilogyAdapter)
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -1824,6 +1824,10 @@ whether a foreign key's name should be dumped to db/schema.rb or not. By
 default, foreign key names starting with `fk_rails_` are not exported to the
 database schema dump. Defaults to `/^fk_rails_[0-9a-f]{10}$/`.
 
+#### `ActiveRecord::SchemaDumper.sort_table_columns`
+
+When `true`, table columns are sorted alphabetically in the generated schema file. When `false`, the columns are listed in the same order as in the database. Defaults to `true`.
+
 #### `config.active_record.encryption.support_unencrypted_data`
 
 When `true`, unencrypted data can be read normally. When `false`, it will raise errors. Default: `false`.


fatkodima:
> and sometimes, the order of the columns matters to some degree (otherwise, why have an :after / :before options in migrations at all?).

Can you provide real-world examples when it matters?

mackuba:
It's not about a specific use case, but rather each developer's/team's preference - when you're viewing data using queries in a console like `psql`, it's printed in the table with columns in the order as they are in the database, so you may want to see them in that order, and not a completely different one.

ngan:
I don't have a opinion on sorting vs not sorting, but to answer your question, @fatkodima:

> Can you provide real-world examples when it matters?

Does this count? https://github.com/paper-trail-gem/paper_trail/pull/1457

fatkodima:
> Does this count? https://github.com/paper-trail-gem/paper_trail/pull/1457

No, because schema.rb is for development use only, while production schemas can have different order.

ngan:
I see. So if you wanted to have specific ordering for production, you'd write a migration to change the column ordering and know that this would yield no changes to your schema.rb. Make sense! Although I can see how this could confuse people.

mackuba:
Hmm, schema.rb is for development use only? That's the first time I'm hearing this… so what is supposed to be used in production?

fatkodima:
Migrations are used to change db in production. `schema.rb` file is not needed there.

mackuba:
But `db:migrate` was just changed in 8.0 so that when you initialize a new database, it skips running all migrations and loads `schema.rb` instead (https://github.com/rails/rails/pull/52830), I think that change applies to both dev and production?

bensheldon:
I support this:

- in Postgres, [column ordering affects storage size](https://pganalyze.com/blog/5mins-postgres-reducing-table-size). I do consider this when laying out my database tables.
- For single-tenant and “stamped”/duplicatable applications , I do use schema to install the database and I think that’s a valid use for it. Otherwise one would have to maintain migrations longterm and that also isn’t appropriate imo.

pbstriker38:
> But db:migrate was just changed in 8.0 so that when you initialize a new database, it skips running all migrations and loads schema.rb instead (https://github.com/rails/rails/pull/52830), I think that change applies to both dev and production?

This is definitely going to effect new apps that spend some time in development before being pushed to production for the first time.

This is also going to make it so that you need to be explicit with the column names when copying data between databases. Even in development, everyone could have a different layout depending on when was the last time they dropped their local DB and did a schema:load.

bensheldon:
> Even in development, everyone could have a different layout depending on when was the last time they dropped their local DB and did a schema:load.

I think it's a good practice to regularly sync the production schema with the committed codebase schema. That goes along with encouraging developers to limit usage of bespoke local dev-data and invest time making seeds effective. Which makes local dev environments less precious and more consistent.

stevequinlan:
I think this really is a team preference thing. We find benefit in being able to group "like with like" columns together in our larger tables. It's much easier for us to spot inconsistencies in naming for example. We rely on reading schema.rb for reference, and having the schema dumper undo all this curation by automatically sorting columns was unfortunate for us. So we've disabled the sorting for our project (~360 tables, ~5600 columns).

Hopefully this PR gets merged. Since there's a merge conflict, here is a commit which may be of use, based on this PR, which updates configuring.md instead of the changelog.md file. 
https://github.com/ClinchIO/rails/commit/e4d176d004785537e8eb266b311474cbc2b4c450

mackuba:
Hey! I've rebased the PR based on @stevequinlan's version.

afittro:
👀 following this. I made a patch, for now, to give me what I need. hopefully this will get approved.

tachyons:
We use explicit column order at GitLab. you can read about it [here](https://docs.gitlab.com/development/database/ordering_table_columns/). But we aren't affected by this change as we use structure.sql instead of schema.rb 


Column sorting should be opt-in. Not opt out

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -40,6 +40,11 @@ class SchemaDumper # :nodoc:
     # should not be dumped to db/schema.rb.
     cattr_accessor :unique_ignore_pattern, default: /^uniq_rails_[0-9a-f]{10}$/
 
+    ##
+    # :singleton-method:
+    # Specifies if columns in create_table block should be ordered alphabetically by name.
+    cattr_accessor :sort_table_columns, default: true
```
tachyons, line 46:
```suggestion
    cattr_accessor :sort_table_columns, default: false
```

We should be keep this false by default to restore original behaviour

mackuba:
@tachyons yeah I wouldn't mind that at all, but for now I'm having trouble convincing the team to accept even the opt-out version…

Shuiei:
I second this change. I don't mind opting out of this feature, but I should have the option to prevent columns from being sorted in my `schema.rb`.

Few example:
  - Doing a `SELECT *` and using the index of a column or any script that requires a specific column order will not work anymore (good practice, maybe not, but it can still happen), 
  - Having my columns ordered based on their business commonality is impossible.
  - `created_at` and `updated_at` in the middle of my table, instead of at the end, is something I really dislike.
  - Not reflecting the order of my column in production and development breaks the dev-parity logic.
  - Column order can affect DB size/performance for `PostgreSQL`.
  - It makes the usage of `add_column, after:` and `before:` confusing since it will not be properly reflected in the `schema.rb`

Overall, I do believe devs should have the option to opt out of this until it starts causing issues, and, most importantly, having this configurable is harmless, so I don't understand why there's a battle over whether it should be enforced.

activerecord/lib/active_record/schema_dumper.rb
Gerikon: Sorting by default is already a new standard. We should not change it.
But adding the ability to disable it is great idea.

jasonperrone:
I have a real world example of why I need these columns in database order. I have automated tests which dump a mysql database, uses schema.rb to create a new database and then mysql loads in the data I just dumped... which thanks to this change no longer matches the table columns.

2called-chaos:
If we can't get this merged (I do hope so) can we consider a mechanism that at least dumps both, sql and ruby schema? Because the ruby version is a good reference but now unusable to initialize. 

`config.active_record.schema_format` obviously only accepts either or.

I know that I can use SCHEMA_FORMAT env variable but obviously that would require manual care as migrations will only update one of them automatically.

mackuba:
Looks like this is the no.1 most upvoted currently open PR, if I'm reading this correctly… https://github.com/rails/rails/pulls?q=is%3Apr+is%3Aopen+sort%3Areactions-%2B1-desc

bert-mccutchen:
@mackuba one thing I noticed is that you're missing an entry in the CHANGELOG. I'm not sure if this is dissuading the maintainers from merging, or if they are just busy - most likely the latter.
https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#updating-the-changelog

e.g. PR #53281 has a [CHANGELOG entry](https://github.com/rails/rails/pull/53281/changes#diff-353f4f101b1bb04938d761b6e29c88e0f147084e9e276854a77ae5d9a9a3b2a8).

I'd also just like to throw in my 2 cents into this while I'm already commenting. Having sorted a schema makes things _really weird_ for open source self-hosted applications. When someone spins up one of my applications, all of their columns are sorted and illogical.

mackuba:
@eileencodes hey, any chance someone could look at this again? 🙏🏻

eileencodes:
I don't think we want a config option for this, it should either be sorted or not. If there's a problem caused by sorting other than dev preference, then we should revert. Otherwise the behavior should stay as-is.

Also please don't ping individuals who haven't interacted with the PR, getting attention on a PR is what the discord is for. Thanks for understanding.

jasonperrone:
Well it was fine for 8 major versions, I say it needs to be reverted.

And I don't understand this whole "schema.rb is for dev only" stuff. Nonsense. I should be able to create brand new databases from my schema.rb and I expect them to look exactly like my other databases.

bert-mccutchen:
Let's step back and revisit how we got here.

The main point is that for teams developing overlapping features in parallel, the schema becomes difficult to work with as the dump is not deterministic. This is a valid issue; as DHH points out "Schema dumping should be deterministic regardless of the platform." While I do think this is more of an issue in developer workflow, DHH's point about being deterministic still rings true.

The second point is that it's not guaranteed that every database platform will provide the table schema in an ordinal pattern. _At least that's what I inferred from the previous PRs/issues._

What if instead we move forward and include the ordinal information in the schema alongside the columns, much like we do with `null: {bool}`, `unique: {bool}`, etc. e.g. `position: {int}`

_PS. Please respect the maintainers, and be sure to [follow the contribution guidelines when looking for feedback](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#get-some-feedback)._

mackuba:
> What if instead we move forward and include the ordinal information in the schema alongside the columns, much like we do with `null: {bool}`, `unique: {bool}`, etc. e.g. `position: {int}`

But I think then we're back at the same problem, that dev 1 adds a field `:tags, position: 5` and dev 2 adds a field `:draft, position: 5` on a parallel branch, and then when they merge, they have two fields at position 5?

bert-mccutchen:
That's true, it doesn't fix the problem of development workflow and merges. However, it is still deterministic, and I don't believe that it is the responsibility of Rails to make merges easier anyways.

**Update:** I've come full circle in realizing that the original schema actually was deterministic in nature when it considers ordinal positions of columns **for a given platform**. It's just that for #53281, ordinal positions between the two development databases were truly different. So if the claim in #32111 that different platforms do not guarantee returning accurate ordinal positions of columns is true. Then  it **can't** be deterministic **across different platforms**.

Though, if it simply **only** caused by development workflow... then I'm on the fence that we revert.

bert-mccutchen:
Ok so I did a test, and I have found <ins>no evidence</ins> among the supported ActiveRecord adapters that they produce non-deterministic results due to ordinal columns. This means that even with ordinal columns, schema dumps are deterministic across platforms! (excluding extensions)

<details>

<summary>MySQL Test Script</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "~> 7.2"
  gem "mysql2"
end

require "rake"

ENV["DATABASE_URL"] = "mysql://root@localhost/ar_ordinal_schema_test"

  config.eager_load = false

  config.paths["db"] = ["mysql_db"]
end
Rails.application.initialize!
Rails.application.load_tasks

Rake::Task['db:prepare'].invoke

    t.string "token", null: false
    t.string "name", null: false
    t.string "description"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.datetime "deleted_at", null: false
    t.index ["token"], name: "index_posts_on_token", unique: true
  end
end

Rake::Task['db:schema:dump'].invoke
```
</details>

<details>

<summary>PostgreSQL Test Script</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "~> 7.2"
  gem "pg"
end

require "rake"

ENV["DATABASE_URL"] = "postgresql://localhost/ar_ordinal_schema_test"

  config.eager_load = false

  config.paths["db"] = ["pgsql_db"]
end
Rails.application.initialize!
Rails.application.load_tasks

Rake::Task['db:prepare'].invoke

    t.string "token", null: false
    t.string "name", null: false
    t.string "description"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.datetime "deleted_at", null: false
    t.index ["token"], name: "index_posts_on_token", unique: true
  end
end

Rake::Task['db:schema:dump'].invoke
```
</details>

<details>

<summary>SQLite Test Script</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "~> 7.2"
  gem "sqlite3"
end

require "rake"

  config.eager_load = false

  config.paths["db"] = ["sqlite_db"]
end
Rails.application.initialize!
Rails.application.load_tasks

Rake::Task['db:prepare'].invoke

    t.string "token", null: false
    t.string "name", null: false
    t.string "description"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.datetime "deleted_at", null: false
    t.index ["token"], name: "index_posts_on_token", unique: true
  end
end

Rake::Task['db:schema:dump'].invoke
```
</details>

---

Meanwhile, sorting columns alphabetically produces an incorrect column ordering in **production** when spinning up a Rails application a new region for data residency purposes via `db:prepare`. Something that my company needs to do with every new region we begin to support. I believe this satisfies the condition to revert mentioned by Eileen above.

>  If there's a problem caused by sorting other than dev preference, then we should revert.

---

This essentially means that the original change to sort columns in the schema alphabetically was included due to developer preference and workflows, and has a real-world impact on production environments as a result.

mackuba:
I *think* that what the original PR meant by non-deterministic results is that:

1) dev A adds `field_a` to a table
2) dev B adds `field_b` to the same table
3) someone merges their branches
4) schema becomes `..., field_a, field_b`
5) dev B pulls the latest version, runs migration; `field_a` gets added, `field_b` was already there
6) dev B's database now has `..., field_b, field_a`
7) Rails on dev B's computer dumps the schema and shows a change in the schema file, trying to change the order from `field_a, field_b` to `field_b, field_a`

bert-mccutchen:
Yes, and that result makes complete sense since their schema was dumped from a database that had columns in a different order than the previous schema dump. This does not mean that it's the responsibility of the schema dumper to resolve this failing - especially at the cost of incorrect column ordering in production when using `db:prepare` when spinning up new regions.

### So then how do we properly solve this?

Everyone having this problem is looking in the wrong spot. Trying to fix the symptom (schema.rb), when they should be focusing on the cause.

The cause is ActiveRecord's **migration process**. When ActiveRecord processes old migrations, it does not consider what columns were added after the one it is inserting. To fix this issue, it needs to make sure that it inserts them before and not after.

<details>

<summary>Here is a debug script that outlines this issue.</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

    t.string :email, null: false
  end
end

# db/migrate/1771075560_add_name_to_users.rb
class AddNameToUsers < ActiveRecord::Migration::Current
  def change
    add_column :users, :first_name, :string
    add_column :users, :last_name, :string
  end
end

# db/migrate/1771075634_add_display_name_to_users.rb
class AddDisplayNameToUsers < ActiveRecord::Migration::Current
  def change
    add_column :users, :display_name, :string
  end
end

end

  setup do
    # Reset between.
    AddNameToUsers.migrate(:down)
    AddDisplayNameToUsers.migrate(:down)
  end

  ##
  # Assume the following migration structure
  # db/
  # └── migrate/
  #     ├── 1771075543_create_users.rb
  #     ├── 1771075560_add_name_to_users.rb
  #     └── 1771075634_add_display_name_to_users.rb

  # Developer 1 is adding first_name/last_name.
  def test_migration_in_order_for_developer_1
    AddNameToUsers.migrate(:up)
    AddDisplayNameToUsers.migrate(:up)
    User.reset_column_information

    # Everything is fine since the display name migration is after their migration.
    assert_equal %w[id email first_name last_name display_name], User.columns_hash.keys
  end

  # Developer 2 is adding display_name and has just updated their branch with developer 1's changes.
  def test_migration_out_of_order_for_developer_2
    AddDisplayNameToUsers.migrate(:up)
    AddNameToUsers.migrate(:up)
    User.reset_column_information

    # ActiveRecord migrations fail to recognize that the columns added in migration 1771075560
    # should be inserted before the columns added in migration 1771075634.
    assert_equal %w[id email first_name last_name display_name], User.columns_hash.keys
  end
end
```

</details>

---

To fix this, ActiveRecord would need to store (and check) metadata about each column in regards to what migration created each column. Even though this could be done by yet another migration metadata table, **I don't think we should**. The overhead for such a small developer inconvenience is not worth it.

e.g. `schema_column_migrations`

| table | column | version |
|---|---|---|
| users | id | 1771075543 |
| users | email | 1771075543 |
| users | first_name | 1771075560 |
| users | last_name | 1771075560 |
| users | display_name | 1771075634 |

---

So to summarize everything again, for the busy folks.
1. The original issue with schema conflicts (which is a symptom of another issue entirely) is strictly a minor developer inconvenience.
2. The fix (alphabetical schemas) for this issue is causing real-world problems in production when creating a new database in a new region for regionalized Rails applications (data residency) via `db:prepare`.
3. There was chatter in previous issues about schema dumping not returning columns in a deterministic way across platforms. Which I have disproved.
4. The real issue is with how ActiveRecord handles adding columns from previous migrations - rather than the schema dump.

---

With all of this information, we should revert. Alphabetical columns in schemas is not the right answer to the problem.

eileencodes:
> And I don't understand this whole "schema.rb is for dev only" stuff

Both are kind of only for dev stuff. You can load up a schema for a new db in production but you'd only do that once. Regardless that's not what we're really discussing here.

From my perspective, the schema and structure files are for computers to rebuild the database, not for humans to read, so there should be no applied order to the output.

> With all of this information, we should revert

Can someone open a PR reverting so Rails core can discuss a final decision?

mackuba:
> From my perspective, the schema and structure files are for computers to rebuild the database, not for humans to read

Yes, but the whole point is not about how a `schema.rb` file looks, the point is how the resulting database table looks. Right now this files creates a table that looks different than originally intended.

bert-mccutchen:
I created a PR per Eileen's request: https://github.com/rails/rails/pull/56842

flanger001:
Can someone help me understand here? Was the point of [the original PR](https://github.com/rails/rails/pull/53281) to change the order of columns in `schema.rb`, or the actual order of the columns in the database? If I start from scratch and only use `db:migrate` to create my `schema.rb`, I get the columns in my database in the order I expect. If I drop my database and recreate it `bin/rails db:drop db:migrate`, the columns appear in the order they are in the existing `schema.rb`, which is to say, out of order.

This "`schema.rb` is only for development" stuff feels very much like the "Arel is a private interface" stuff and I really hope we're not trying to do that again. The instructions at the top of the `schema.rb` file say to use `bin/rails db:schema:load` when creating a new database in a new environment. If we're supposed to use this file to create new databases, is that not the mechanism we would use in production? If we do that, that creates the columns in the wrong order.

Please allow this to be configured, or please revert the original change.

jagthedrummer:
Not sure how (or if) votes are being tabulated but I'd vote for #56842 over this one ~because column ordering inside the database is extremely important for database indexes which rely on "leftmost column" optimizations. The changes in #53281 seem to make it impossible to control which columns are leftmost in a fresh database.~

Edited to say that I was mistaken. Column order is important when declaring the _index_ but not for the _table_.

(I still prefer to just revert because I think having `schema.rb` match the database makes the most sense.)

activerecord/lib/active_record/schema_dumper.rb
francois-ferrandis: > Sorting by default is already a new standard

I'd argue that a new standard that is 4-months old is far weaker than a standard that has been stable for 15 years.

Sorting columns solves a problem that occurs in the context of some big projects, at the cost of causing serious issues in most other contexts.

jakeonrails:
For those who are following and want a simple fix, I have created a gem [jakeonrails/unsort-db-schema-columns](https://github.com/jakeonrails/unsort-db-schema-columns) which inverts the column sorting behavior upon inclusion in your Gemfile. I just published to RubyGems.

I am the original author of the gem that let users opt-in to sorted behavior before this PR ever got merged into rails - [fix-db-schema-conflicts](https://rubygems.org/gems/fix-db-schema-conflicts/versions/3.0.1). Feels fitting that I pitch in a gem to help with this too, lol. LMK if you have any issues and please feel free to submit PRs.

rafaelfranca:
I'm closing this one because I don't think this should be configurable. There is enough evidence to prove that we need to revert the sorting. I'm going with https://github.com/rails/rails/pull/56842.

Thank you everyone for the input.

mackuba:
@rafaelfranca Thank you! 🙏🏻

PR #55415, opened 2025-07-28
Add Active Job rate limiting

heka1024:
### Motivation / Background

My application relies on a third-party API that has a rate limit. Instead of exceeding the rate limit and retrying after receiving a 429 error, I want to control the request rate manually.

### Detail

The implementation follows the same patterns as ActionController's rate limiting feature for consistency. It uses ActiveSupport::Cache for persistence.

### Checklist


rafaelfranca:
@matthewd @byroot opinions on the feature? It makes sense to me but this isn't something I see much usage.

byroot:
> opinions on the feature? It makes sense to me but this isn't something I see much usage.

Well, ideally the remote API does feed you with actual rate limit information so that you can handle it in a smarter way, but if it doesn't then yes that feature somewhat makes sense.

I don't have a strong opinion, but I don't see anything particularly wrong with it.

PR #55416, opened 2025-07-28
Requiring only ActiveSupport::CoreExt::Module::Delegation causes NameError

Greg-Myers-SB:
IRB:
```ruby
require 'active_support/core_ext/module/delegation'

class Foo
  delegate :puts, to: Kernel
end
```

```
activesupport-8.0.2/lib/active_support/delegation.rb:45:in `generate': uninitialized constant #<Class:ActiveSupport::Delegation>::Inflector (NameError)

          unless Inflector.safe_constantize(to.name).equal?(to)
                 ^^^^^^^^^
	from lib/ruby/gems/3.3.0/gems/activesupport-8.0.2/lib/active_support/core_ext/module/delegation.rb:161:in `delegate'
```

IRB
```ruby
require 'active_support/inflector'
require 'active_support/core_ext/module/delegation'

class Foo
  delegate :puts, to: Kernel
end
```
👍 

### Motivation / Background

When requiring only the Delegation or CoreExt/Delegation on it's own, the code will fail because of missing requirement on Inflector/Inflection.

### Detail

The delegation module doesn't declare it's requirement. This PR makes the requirement explicit.

### Checklist


Greg-Myers-SB:
A question for maintainers, do we expect projects to be able to require any file in ActiveSupport directly, and it be a valid/working require?

skipkayhil:
> A question for maintainers, do we expect projects to be able to require any file in ActiveSupport directly, and it be a valid/working require?

No, you have to `require "active_support"` first

https://guides.rubyonrails.org/active_support_core_extensions.html#stand-alone-active-support

PR #55419, opened 2025-07-29
Fix `HWIA#transform_keys!` removing defaults

skipkayhil:
Ref #55376 

A [previous change][1] fixed `HWIA#transform_keys!` from clobbering keys if they ended up transforming into another pre-transformed key. However, that change also resulted in `HWIA#transform_keys!` clearing the `default`/`default_proc`.

This commit fixes that issue and adds test coverage to both `#transform_keys` and `#transform_keys!` to ensure that their behavior with regard to `default`/`default_proc` is consistent with `Hash`.

```irb
irb(main):001> Hash.new(:a).transform_keys(&:to_s).default
=> nil
irb(main):002> Hash.new(:a).transform_keys!(&:to_s).default
=> :a
```

[1]: https://github.com/rails/rails/commit/f49d7a0d5a67eccb8d7017c38ac157a5244a7526

cc @byroot

### Checklist

### DIFF
--- a/activesupport/lib/active_support/hash_with_indifferent_access.rb
+++ b/activesupport/lib/active_support/hash_with_indifferent_access.rb
@@ -278,9 +278,7 @@ module ActiveSupport
     #   hash[:a][:c] # => "c"
     #   dup[:a][:c]  # => "c"
     def dup
-      self.class.new(self).tap do |new_hash|
-        set_defaults(new_hash)
-      end
+      copy_defaults(self.class.new(self))
     end
 
     # This method has the same semantics of +update+, except it does not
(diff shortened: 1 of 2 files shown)


byroot:
Good catch. I wasn't aware `Hash#replace` also copies the default / default_proc. The doc doesn't mention it, and it's not speced, but the code is pretty clear:

```c
static VALUE
rb_hash_replace(VALUE hash, VALUE hash2)
{
    rb_hash_modify_check(hash);
    if (hash == hash2) return hash;
    if (hash_iterating_p(hash)) {
        rb_raise(rb_eRuntimeError, "can't replace hash during iteration");
    }
    hash2 = to_hash(hash2);

    COPY_DEFAULT(hash, hash2);

    if (RHASH_AR_TABLE_P(hash)) {
        hash_ar_free_and_clear_table(hash);
    }
    else {
        hash_st_free_and_clear_table(hash);
    }

    hash_copy(hash, hash2);

    return hash;
}
```

byroot:
I pushed a commit to avoid using `Object#tap` as much. Also opened a doc PR on Ruby: https://github.com/ruby/ruby/pull/14046

PR #45680, opened 2022-07-28
Add `action_controller.allowed_redirect_hosts` configuration

Kevinrob:
### Summary

We use Rails with multiple domains (and subdomains). All these domains are configured with `config.hosts`.  
Sometimes we have dynamic redirect from one domain to an other.  
And `ActionController::Redirecting::UnsafeRedirectError` is raised.

It would be great to allow redirections to domains that the app accept in `config.hosts`.
This should not be considered as unsafe.

PR #55420, opened 2025-07-30
Allow hosts redirects from `hosts` Rails configuration

byroot:
Fix: https://github.com/rails/rails/pull/45680/

Reopened the PR because I can't push in the other PR source repo to do the last round of cleanup.

FYI: @q8 @zzak @Kevinrob

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -16,6 +16,18 @@ module ActionController
     included do
       mattr_accessor :raise_on_open_redirects, default: false
       mattr_accessor :action_on_path_relative_redirect, default: :log
+      class_attribute :_allowed_redirect_hosts, :allowed_redirect_hosts_permissions, instance_accessor: false, instance_predicate: false
+      singleton_class.alias_method :allowed_redirect_hosts, :_allowed_redirect_hosts
+    end
+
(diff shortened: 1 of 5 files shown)


Kevinrob:
Thank you very much for your help!

actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -16,6 +16,18 @@ class UnsafeRedirectError < StandardError; end
     included do
       mattr_accessor :raise_on_open_redirects, default: false
       mattr_accessor :action_on_path_relative_redirect, default: :log
+      class_attribute :_allowed_redirect_hosts, :allowed_redirect_hosts_permissions, instance_accessor: false, instance_predicate: false
```
rafaelfranca, line 19: Do it need to be inheritable? I can only see this config be per application, not per controller

byroot: I figured it could be useful to allow in restricted context, e.g. allow redirector to `foo.com` only from that specific controller that integrate with the `foo.com` service, without degrading the security of other controllers.

I'm not hell bent on it. If it bother you, we can make it a global.

rafaelfranca: Didn't thought about that. It is fine

PR #55421, opened 2025-07-30
Default to fiber isolation from Rails 8.1+.

samuel-williams-shopify:
### Motivation / Background

This Pull Request has been created because Rails should adopt fiber isolation mode as the default for new applications in Rails 8.1 to better support modern fiber-based servers and concurrency patterns.

Currently, Rails defaults to thread-based isolation (`isolation_level = :thread`), but fiber-based isolation (`isolation_level = :fiber`) provides better support for:
- Fiber-based web servers like Falcon.
- Modern Ruby concurrency patterns using fibers.
- Better isolation of execution state in fiber-concurrent scenarios.

As Ruby's fiber support continues to mature and more applications adopt fiber-based architectures, Rails should provide sensible defaults that support these patterns out of the box.

### Detail

This Pull Request changes the default `config.active_support.isolation_level` from `:thread` to `:fiber` for newly generated Rails 8.1 applications by:

- Adding `active_support.isolation_level = :fiber` to the Rails 8.1 defaults in `load_defaults "8.1"`.
- Updating tests to reflect the new default while maintaining backward compatibility tests.
- Updating the configuration guide documentation to list the new default.

The change maintains full backward compatibility - existing applications that upgrade to Rails 8.1 but keep `config.load_defaults "8.0"` (or earlier) will continue using thread-based isolation. Only newly generated Rails 8.1 applications will get fiber isolation by default.

### Additional information

- Fiber isolation mode has been available since Rails 7.0 and is well-tested.
- The change follows Rails' established pattern for version-specific configuration defaults.
- Applications can still explicitly override the default by setting `config.active_support.isolation_level = :thread` if needed.
- This change aligns Rails with the direction of modern Ruby concurrency patterns.
- In the vast majority of cases, there shouldn't be any compatibility issues.

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -63,6 +63,7 @@ Below are the default values associated with each target version. In cases of co
 - [`config.action_controller.action_on_path_relative_redirect`](#config-action-controller-action-on-path-relative-redirect): `:raise`
 - [`config.action_controller.escape_json_responses`](#config-action-controller-escape-json-responses): `false`
 - [`config.active_record.raise_on_missing_required_finder_order_columns`](#config-active-record-raise-on-missing-required-finder-order-columns): `true`
+- [`config.active_support.isolation_level`](#config-active-support-isolation-level): `:fiber`
 - [`config.yjit`](#config-yjit): `!Rails.env.local?`
 
 #### Default Values for Target Version 8.0
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Set `config.active_support.isolation_level = :fiber` as the default for Rails 8.1 applications.
+
+    This provides better support for fiber-based servers and modern concurrency patterns.
+    Existing applications upgrading to Rails 8.1 will maintain thread-based isolation
+    unless they update their `config.load_defaults` version.
+
+    *Samuel Williams*
+
 *   Generate static BCrypt password digests in fixtures instead of dynamic ERB expressions.
 
     Previously, fixtures with password digest attributes used `<%= BCrypt::Password.create("secret") %>`,
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -364,6 +364,10 @@ module Rails
           if respond_to?(:active_record)
             active_record.raise_on_missing_required_finder_order_columns = true
           end
+
+          if respond_to?(:active_support)
+            active_support.isolation_level = :fiber
+          end
         else
           raise "Unknown version #{target_version.to_s.inspect}"
         end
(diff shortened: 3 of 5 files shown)


Earlopain (commented):
The docs currently mention this:
> If you use a fiber based server or job processor (e.g. falcon), you should set it to :fiber. Otherwise it is best to use :thread locality. Defaults to :thread.

So I'm not sure what to make of this. The default server is puma, which I don't think even has any support for fibers?

I just use puma since its the default and don't worry much about the rest. But changing the isolation level would be something I am very careful about. I have no clue which (if any) issues I could run into with this.

If this gets changed, it would be very nice if the changelog/framework defaults contains an explanation about what specifically to look out for.

----

Anyways, something more related to the actual code. You need to make additional changes when changing a default. You can orient yourself at previous prs like https://github.com/rails/rails/pull/54643. Specifically `new_framework_defaults_8_1` and the like.

samuel-williams-shopify:
There is nothing wrong with running Puma in fiber mode, every thread has a root fiber.

There are some potential minor incompatibilities with sharing a connection into a lazy enumerator, but that's already problematic, and since most database operations use `with_connection` now, this change would overall improve safety and compatibility.

That being said, it would be good to experiment with this change more, but I believe people have to opt into these new defaults, so that's a safety net.

So:

1. Enabling fiber mode as the default should be compatible with almost all code except some questionable edge cases.
2. Not enabling fiber mode is extremely unsafe for Falcon and any code that uses operation-per-fiber.

That's why I'd like to see this change adopted, because (1) is a much better place to be by default.

Thanks for the link to the other PR, I've updated the template file.

Earlopain:
Thanks for your comment, that sounds good to to me. I was a bit uncertain since you wrote

> In the vast majority of cases, there shouldn't be any compatibility issues.

and that didn't give much hints to what these compatibility issues could be.

> people have to opt into these new defaults

You are expected to uncomment things in the new framework file and also to update `load_defaults` as part of the upgrade process. That said, you still have to manually do all that, so it's not changing without consent. The user should understand the imapct of the new defaults before applying them and flip back to the original if there are concerns. Personally I would call that opt-out but I can see the argument going both ways.

------

I can't make any comments on the actual impact here (and don't want to attempt to do so) but the code itself looks good. You do need to update https://guides.rubyonrails.org/configuring.html#config-active-support-isolation-level though, and add a table documenting the defaults for each rails version (like in the one right below at 3.14.17).

byroot:
> * In the vast majority of cases, there shouldn't be any compatibility issues.

I introduced the configurable isolation level specifically because lots of "context state" was stored in `Thread.current` (fiber local contrary to what the name suggest) and a number of people reported issues around enumerators and such.

So I very strongly disagree with this claim.

As @Earlopain said, as long as Rails defaults use threads (Puma, Solid Queue), it doesn't make sense to change the default isolation level.

ioquatix:
@byroot Were there any other issues apart from lazy enumerators? Because that's the only one I've heard of.

PR #55334, opened 2025-07-14
Structured Event Reporting in Rails

adrianna-chang-shopify:
### Motivation / Background

Ref: https://github.com/rails/rails/issues/50452

This PR adds structured event reporting to Rails. The event reporter, accessible via `Rails.event`, allows you to report structured events to subscriber(s), and provides mechanisms for adding tags and context to events. While the existing `Rails.logger` is great for producing human-readable logs, the unstructured log lines it generates are difficult for modern data analytics / observability platforms to parse. We'd like to add first-class support for structured events, complementing the existing Rails logger. 

Note that events encompass "structured logs", but also "business events", as well as telemetry events such as metrics and logs. The Event Reporter is designed to be a single interface for producing any kind of event in a Rails application. 

This is the API we developed for structured logging at Shopify and are using in our monolith.

#### Basic Event Reporting
```ruby
  # Simple event with payload
  Rails.event.notify("user_created", { id: 123, name: "John Doe" })

  # Event object (for schematized events)
  Rails.event.notify(UserCreatedEvent.new(id: 123, name: "John Doe"))
```
 
#### Event Structure

```ruby
# All events include standardized metadata:
  {
    name: "user_created",
    payload: { id: 123, name: "John Doe" },
    tags: { section: "admin" },
    context: { request_id: "abc123", user_agent: "Mozilla..." },
    timestamp: 1738964843208679035, # nanosecond precision
    source_location: {
      filepath: "app/services/user_service.rb",
      lineno: 45,
      label: "UserService#create"
    }
  }
```

#### Adding Tags / Context

```ruby
 # Tags - Domain-specific context that nests:
  Rails.event.tagged("graphql") do
    Rails.event.tagged(section: "admin") do
      Rails.event.notify("user_created", { id: 123 })
      # Event includes tags: { graphql: true, section: "admin" }
    end
  end

  # Context - Request/job-scoped metadata:
  Rails.event.set_context(request_id: "abc123", shop_id: 456)
  Rails.event.notify("user_created", { id: 123 })
  # All subsequent events in this request include the context
```

#### Debug Mode - Conditional event reporting:
  ```ruby
  Rails.event.with_debug do
    Rails.event.debug("sql_query", { sql: "SELECT * FROM users" })
    # Only reported when debug mode is active
  end
```

 #### Subscriber Interface

EDIT: I've amended this PR to provide some default encoders out of the box, and we may additionally want to provide a default subscriber that converts structured events into unstructured log lines.

[...]

PR #55425, opened 2025-07-30
Add support for `rack.response_finished` callbacks in `ActionDispatch::Executor`

adrianna-chang-shopify:
### Motivation / Background

The executor middleware now supports deferring completion callbacks to later in the request lifecycle by utilizing Rack's `rack.response_finished` mechanism, when available. This enables applications to define `rack.response_finished` callbacks that may rely on state that would be cleaned up by the executor's completion callbacks.

This came up in the context of https://github.com/rails/rails/pull/55334#discussion_r2236757794 -- we want to use `to_complete` hooks to handle clearing context for the Event Reporter, but Shopify's application code wants to use said context in a `rack.response_finished` callback in order to produce a request summary log.

By moving the `to_complete` callbacks to `rack.response_finished` when available instead of `Rack::BodyProxy`, applications are able to define their own `rack.response_finished` callbacks that may rely on state that would be cleaned up (e.g. `ExecutionContext`, `CurrentAttributes`, and eventually the `EventReporter` context).

### Detail

- If `rack.response_finished` is set, we assume the webserver supports it. If there isn't support for it, we continue to use `Rack::BodyProxy`.
- We push the `state.complete!` to the array of callables; we do this _before_ we call the app, because we want it to run _after_ any application `rack.response_finished` callbacks run (and callbacks are executed in reverse order).
- We only run `state.complete!` in the `ensure` block if we're using `BodyProxy`; `rack.response_finished` ["should not raise any errors"](https://github.com/rack/rack/blob/main/SPEC.rdoc#rackresponse_finished-) and are run from an `ensure` block, so even if something went awry we'd invoke the complete callbacks. We don't want to invoke them twice.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/executor.rb
+++ b/actionpack/lib/action_dispatch/middleware/executor.rb
@@ -12,6 +12,10 @@ module ActionDispatch
 
     def call(env)
       state = @executor.run!(reset: true)
+      if response_finished = env["rack.response_finished"]
+        response_finished << -> { state.complete! }
+      end
+
       begin
         response = @app.call(env)
 
@@ -20,7 +24,11 @@ module ActionDispatch
           @executor.error_reporter.report(error, handled: false, source: "application.action_dispatch")
         end
 
-        returned = response << ::Rack::BodyProxy.new(response.pop) { state.complete! }
+        unless response_finished
+          response << ::Rack::BodyProxy.new(response.pop) { state.complete! }
+        end
+        returned = true
+        response
       rescue Exception => error
         request = ActionDispatch::Request.new env
         backtrace_cleaner = request.get_header("action_dispatch.backtrace_cleaner")
@@ -28,7 +36,9 @@ module ActionDispatch
         @executor.error_reporter.report(wrapper.unwrapped_exception, handled: false, source: "application.action_dispatch")
         raise
       ensure
-        state.complete! unless returned
+        if !returned && !response_finished
+          state.complete!
+        end
       end
     end
   end
--- a/actionpack/test/dispatch/executor_test.rb
+++ b/actionpack/test/dispatch/executor_test.rb
@@ -170,10 +170,44 @@ class ExecutorTest < ActiveSupport::TestCase
     end
   end
 
+  def test_complete_callbacks_are_called_on_rack_response_finished
+    completed = false
+    executor.to_complete { completed = true }
+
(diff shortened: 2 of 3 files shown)


eileencodes (approved):
Can you fix the conflicts?

adrianna-chang-shopify:
Done, thanks @eileencodes !

actionpack/test/dispatch/executor_test.rb
```diff
@@ -170,10 +170,44 @@ def test_handled_error_is_not_reported
...
+    completed = false
+    executor.to_complete { completed = true }
+
+    env = { "rack.response_finished" => [] }
+    call_and_return_body(env)
+
+    assert_not completed
+
+    assert_equal 1, env["rack.response_finished"].size
+    env["rack.response_finished"].first.call
```
ioquatix, line 183: This test is insufficient/not to spec (the Rack spec).

```suggestion
    env["rack.response_finished"].first.call({}, 200, {}, nil)
```

ioquatix:
Thanks for working on this.

This change breaks Falcon and I don't believe it's compliant with Rack.

Rack requires the callback to be invoked with arguments `env, status, headers, error`.

However, `->{}` enforces argument size requirements, so `-> { state.complete! }` cannot be called with `env, status, headers, error`.

```ruby
callback = ->{}

env, status, headers, error = {}, 200, {}, nil
=> [{}, 200, {}, nil]

callback.call(env, status, headers, error)
# (irb):3:in 'block in <top (required)>': wrong number of arguments (given 4, expected 0) (ArgumentError)
```

You can replace `->{}` with `proc{}` which has the semantics you are after (ignores arguments).

See <https://github.com/rack/rack/blob/main/SPEC.rdoc#label-rack.response_finished> for more details on the exact requirements.

actionpack/lib/action_dispatch/middleware/executor.rb
```diff
@@ -12,6 +12,10 @@ def initialize(app, executor)
 
     def call(env)
       state = @executor.run!(reset: true)
+      if response_finished = env["rack.response_finished"]
+        response_finished << -> { state.complete! }
```
ioquatix, line 16:
```suggestion
        response_finished << proc { state.complete! }
```

adrianna-chang-shopify:
Hey @ioquatix , thank you for the feedback -- will open a PR to fix!

PR #53220, opened 2024-10-07
Set Current.session if signed-in in unauthenticated access

t27duck:
### Motivation / Background

This is related to the authentication generator.

Currently, if you call `allow_unauthenticated_access` in a controller, `Current.session` won't be set. This can result in rendered views making the user appear to be signed out if they are already signed in.

### Detail

This updates `allow_unauthenticated_access` to set a `before_action` of `resume_session` to allow the user to still be signed in for requests that do not require authentication.

### Additional information

~~Also included in this PR are two quick fixup commits:~~

- ~~Skip querying the `Session` model needlessly if the cookie isn't set.~~
- ~~`SessionsController#destroy` should redirect with `see_other` to match existing scaffolding controller destroy actions.~~

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55428, opened 2025-07-30
[RF-DOCS] Wishlists guide

excid3:
### Motivation / Background

This adds a new tutorial adding Wishlists to the e-commerce store as a follow-up to the Sign Up & Settings Guide. It builds upon the Rails knowledge learned so far and helps reinforce the concepts learned so far.

cc @AmandaPerino 

### Checklist


guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+### Filtering Wishlists
+
+To get a better look at data in the admin area, it's helpful to have filters. We
+can filter wishlists by `User` or by `Product`.
+
+Update `app/views/store/wishlists/index.html.erb` by adding the following form:
+
+```erb#1,3-7
+<h1><%= pluralize @wishlists.count", "Wishlist" %></h1>
```
leegeng: Minor typo breaks ERB parsing

`pluralize` expects the form `pluralize(count, word)`, but the current line has a misplaced quote.


```suggenstion
<h1><%= pluralize @wishlists.count, "Wishlist" %></h1>
```

excid3:
```suggestion
<h1><%= pluralize @wishlists.count, "Wishlist" %></h1>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
+**DO NOT READ THIS FILE ON GITHUB, GUIDES ARE PUBLISHED ON <https://guides.rubyonrails.org>.**
+
+Wishlists
+=========
+
+This guide covers adding Wishlists to the store e-commerce application in the
+[Getting Started Guide](getting_started.html)). We will use the code from the
+[Sign up and Settings Guide](sign_up_and_settings.html) as a starting place.
```
MatheusRich: Wishlist: have all the code from the previous guides ready so I don't have to follow them all before starting this one haha.

rafaelfranca: That isn't a bad idea. Maybe we should put the getting started code in a repository that people can clone?

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+class WishlistProduct < ApplicationRecord
+  belongs_to :product, counter_cache: :wishlists_count
+  belongs_to :wishlist, counter_cache: :products_count
+
+  validates :product_id, uniqueness: { scope: :wishlist_id }
+end
+```
+
+We've specified a column name to update on the associated models. For the
+Product model, we want to use the `wishlists_count` column and for `Wishlist` we
```
rafaelfranca: We aren't being consistent in quoting the model names. In some places we do, in others we don't. I think we need to pick one and stick with it

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+The uniqueness validation also tells Rails to check if a product is already on
+the wishlist. This is paired with the unique index on the wishlist_product table
+so that it's also validated at the database level.
+
+Finally, let's update `app/models/wishlist.rb` with it's associations:
+
+```ruby
+class Wishlist < ApplicationRecord
+  belongs_to :user
+  has_many :wishlist_products, dependent: :destroy
```
rafaelfranca, line 188: I think the proper name for this association and the model would be `wishlisted_products`, no?

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+  end
+
+  def new
+    @wishlist = Wishlist.new
+  end
+
+  def create
+    @wishlist = Current.user.wishlists.new(wishlist_params)
+    if @wishlist.save
+      redirect_to @wishlist, status: :see_other, notice: "Your wishlist was created successfully."
```
rafaelfranca:
```suggestion
      redirect_to @wishlist, notice: "Your wishlist was created successfully."
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<div>
+  <%= link_to wishlist.name, wishlist %>
+</div>
+```
+
+Next let's create the `new` view at `app/views/wishlists/new.html.erb`:
+
+```erb
+<h1>New Wishlist</h1>
+<%= render "form", locals: {wishlist: @wishlist} %>
```
rafaelfranca:
```suggestion
<%= render "form", locals: { wishlist: @wishlist } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+```
+
+And the `edit` view at `app/views/wishlists/edit.html.erb`:
+
+```erb
+<h1>Edit Wishlist</h1>
+<%= render "form", locals: {wishlist: @wishlist} %>
```
rafaelfranca:
```suggestion
<%= render "form", locals: { wishlist: @wishlist } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<% end %>
+```
+
+Create `show` next at `app/views/wishlists/show.html.erb`:
+
+```erb
+<h1><%= @wishlist.name %></h1>
+<% if authenticated? && @wishlist.user == Current.user %>
+  <%= link_to "Edit", edit_wishlist_path(@wishlist) %>
+  <%= button_to "Delete", @wishlist, method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
  <%= button_to "Delete", @wishlist, method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+[Stimulus framework](https://stimulus.hotwired.dev/) to add some JavaScript
+functionality to our UI.
+
+First, let's add a button to `app/views/wishlists/show.html.erb`:
+
+```erb#7
+<h1><%= @wishlist.name %></h1>
+<% if authenticated? && @wishlist.user == Current.user %>
+  <%= link_to "Edit", edit_wishlist_path(@wishlist) %>
+  <%= button_to "Delete", @wishlist, method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
  <%= button_to "Delete", @wishlist, method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<% end %
```
rafaelfranca:
```suggestion
<% end %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+```
+
+Then we can update `app/views/wishlists/show.html.erb` to include a "Remove"
+button:
+
+```erb#13-15
+<h1><%= @wishlist.name %></h1>
+<% if authenticated? && @wishlist.user == Current.user %>
+  <%= link_to "Edit", edit_wishlist_path(@wishlist) %>
+  <%= button_to "Delete", @wishlist, method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
  <%= button_to "Delete", @wishlist, method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<% end %>
+
+<h3><%= pluralize @wishlist.products_count, "Product" %></h3>
+<% @wishlist.wishlist_products.includes(:product).each do %>
+  <div>
+    <%= link_to it.product.name, it.product %>
+    <small>Added <%= l it.created_at, format: :long %></small>
+
+    <% if authenticated? && @wishlist.user == Current.user %>
+      <%= button_to "Remove", [@wishlist, it], method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
      <%= button_to "Remove", [ @wishlist, it ], method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+With multiple wishlists, users may want to move a product from one list to
+another. For example, they might want to move items into a "Christmas" wishlist.
+
+In `app/views/wishlists/show.html.erb`, add the following:
+
+```erb#14-19
+<h1><%= @wishlist.name %></h1>
+<% if authenticated? && @wishlist.user == Current.user %>
+  <%= link_to "Edit", edit_wishlist_path(@wishlist) %>
+  <%= button_to "Delete", @wishlist, method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
  <%= button_to "Delete", @wishlist, method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+    <% if authenticated? && @wishlist.user == Current.user %>
+      <% if (other_wishlists = Current.user.wishlists.excluding(@wishlist)) && other_wishlists.any? %>
+        <%= form_with url: [@wishlist, it], method: :patch do |form| %>
+          <%= form.collection_select :new_wishlist_id, other_wishlists, :id, :name %>
+          <%= form.submit "Move" %>
+        <% end %>
+      <% end %>
+
+      <%= button_to "Remove", [@wishlist, it], method: :delete, data: {turbo_confirm: "Are you sure?"} %>
```
rafaelfranca:
```suggestion
      <%= button_to "Remove", [ @wishlist, it ], method: :delete, data: { turbo_confirm: "Are you sure?" } %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+    <div>
+      <%= yield %>
+    </div>
+  </section>
+<% end %>
+
+<%= render template: "layouts/application" %>
+```
+
+Now we can view wishlists in the admin area
```
rafaelfranca:
```suggestion
Now we can view wishlists in the admin area.
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+Run the following to delete all wishlists in the database:
+
+```bash
+$ bin/rails runner "Wishlist.destroy_all"
+```
+
+Try visiting a product and adding it to a Wishlist now.
+
+The first problem is the select box will be empty. The form will not submit a
+`wishlist_id` param to the server and that will cause ActiveRecord to raise an
```
rafaelfranca:
```suggestion
`wishlist_id` param to the server and that will cause Active Record to raise an
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+```
+
+We've updated the header to display a count of wishlists which will be helpful
+when we apply a filter. The form also reads from the URL's query params to
+pre-select an option in the select boxes if a filter is applied. This form
+submits to the `index` action so it can be used for displaying all records or a
+filtered set of records.
+
+To make this work, we need to apply these filters in our SQL query with
+ActiveRecord. Update the controller to include these filters:
```
rafaelfranca:
```suggestion
Active Record. Update the controller to include these filters:
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+```ruby#4-5
+class Store::WishlistsController < Store::BaseController
+  def index
+    @wishlists = Wishlist.includes(:user)
+    @wishlists = @wishlists.where(user_id: params[:user_id]) if params[:user_id].present?
+    @wishlists = @wishlists.includes(:wishlist_products).where(wishlist_products: {product_id: params[:product_id]}) if params[:product_id].present?
```
rafaelfranca:
```suggestion
    @wishlists = @wishlists.includes(:wishlist_products).where(wishlist_products: { product_id: params[:product_id] }) if params[:product_id].present?
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+  def index
+    @subscribers = Subscriber.includes(:product).filter_by(params)
+  end
+
+  def show
+  end
+
+  def destroy
+    @subscriber.destroy
+    redirect_to store_subscribers_path, notice: "Subscriber has been removed."
```
rafaelfranca:
```suggestion
    redirect_to store_subscribers_path, notice: "Subscriber has been removed.", status: :see_other
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+<h3><%= pluralize @wishlist.products_count, "Product" %></h3>
+<% @wishlist.wishlist_products.includes(:product).each do %>
+  <div>
+    <%= link_to it.product.name, it.product %>
+    <small>Added <%= l it.created_at, format: :long %></small>
+
+    <% if authenticated? && @wishlist.user == Current.user %>
+      <% if (other_wishlists = Current.user.wishlists.excluding(@wishlist)) && other_wishlists.any? %>
+        <%= form_with url: [@wishlist, it], method: :patch do |form| %>
```
rafaelfranca:
```suggestion
        <%= form_with url: [ @wishlist, it ], method: :patch do |form| %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<%= render template: "layouts/application" %>
+```
+
+Now you can view, filter, and remove subscribers in the store's admin area. Try
+it out!
+
+## Adding Links To Products
+
+Now that we've added filters, we can add links to the Product show page for
+viewing wishlists and subscribers for a specific product.
```
rafaelfranca, line 1186: We should tell which file is this. It looks like it is the admin page, not the store page

guides/source/wishlists.md
MatheusRich: A problem is that the repo would get out of sync as we update the tutorials.

AmandaPerino: This came up in conversation with @excid3 when we were updating the Getting Started Guide, but the longterm maintenance as @MatheusRich says would be an issue. 
@rafaelfranca Could this be in the Rails GitHub or would do you think an external repo is better?

rafaelfranca: I think an external repo. There are ways to make maintenance easier. We could write a script that generate the repo and keep it inside the repo and make sure its content is the same as the guide.

guides/source/wishlists.md
MatheusRich: Could be, if you think of this as an adjective ("products that were wishlisted"), but wishlist_products also work as "products that are on a wishlist".

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+```bash
+$ bin/rails generate model Wishlist user:belongs_to name products_count:integer
+```
+
+This model has 3 attributes:
+
+- `user:belongs_to` which associates the Wishlist with the `User` record who
+  owns it
+- `name` which we'll also use for friendly URLs
+- `products_count` for the counter cache we'll add to count how many products
```
MatheusRich: I think we never introduce counter caches (maybe in previous tutorials?). Anyway, a link to the docs wouldn't hurt.
```suggestion
- `products_count` for the [counter cache](https://guides.rubyonrails.org/association_basics.html#counter-cache) we'll add to count how many products
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+option to automatically remove join table records when a Wishlist is deleted.
+
+### Friendly URLs
+
+Wishlists are often shared with friends and family. By default, the ID in the
+URL for a `Wishlist` is a simple Integer. This means we can't easily look at the
+URL to determine which `Wishlist` it's for.
+
+Active Record uses `to_param` method that can be used for generating pretty
+URLs. Let's try it out by creating the `to_param` method in our model:
```
MatheusRich:
```suggestion
Active Record has a `to_param` class method that can be used for generating pretty
URLs. Let's try it out in our model:
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+class Wishlist < ApplicationRecord
+  belongs_to :user
+  has_many :wishlist_products, dependent: :destroy
+  has_many :products, through: :wishlist_products
+
+  def to_param
+    "#{id}-#{name.squish.parameterize}"
+  end
+end
+```
```
MatheusRich, line 215: While reading this, I've found https://edgeapi.rubyonrails.org/classes/ActiveRecord/Integration/ClassMethods.html#method-i-to_param

I've always defined `#to_param` manually, but it looks like we could we use that?

```suggestion
```ruby#6
class Wishlist < ApplicationRecord
  belongs_to :user
  has_many :wishlist_products, dependent: :destroy
  has_many :products, through: :wishlist_products

  to_param :name
end
```
```

File: guides/source/wishlists.md [hunk 25]
```diff
@@ -0,0 +1,1545 @@
...
+
+This returns a String for the URL param made up of the `id` and `name` joined by
+hyphens. `name` is made URL safe by using
+[`squish`](https://api.rubyonrails.org/classes/String.html#method-i-squish) to
+clean up whitespace and
+[`parameterize`](https://api.rubyonrails.org/classes/String.html#method-i-parameterize)
+to replace special characters.
```
MatheusRich:
```suggestion
This will create a `to_param`  instance method that returns a String for the URL param made up of the `id` and `name` joined by
hyphens. `name` is made URL safe by using
[`squish`](https://api.rubyonrails.org/classes/String.html#method-i-squish) to
clean up whitespace and
[`parameterize`](https://api.rubyonrails.org/classes/String.html#method-i-parameterize)
to replace special characters.
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+=> #<Wishlist:0x000000012bb71d68
+ id: 1,
+ user_id: 1,
+ name: "Example Wishlist",
+ products_count: nil,
+ created_at: "2025-07-22 15:21:29.036470000 +0000",
+ updated_at: "2025-07-22 15:21:29.036470000 +0000">
+```
+
+It worked! But how? I thought we had to use Integers to find records?
```
MatheusRich: I think the tutorial usually uses "we"
```suggestion
It worked! But how? Didn't we have to use Integers to find records?
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+The way we're using `to_param` takes advantage of how Ruby converts Strings to
+Integers. Let's convert that param to an integer using `to_i` in the console:
```
MatheusRich:
```suggestion
The way we're using `to_param` takes advantage of [how Ruby converts Strings to
Integers](https://docs.ruby-lang.org/en/master/String.html#method-i-to_i). Let's convert that param to an integer using `to_i` in the console:
```

guides/source/wishlists.md
MatheusRich: Ignore, you do just that later!

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<h1>Edit Wishlist</h1>
+<%= render "form", locals: {wishlist: @wishlist} %>
+```
+
+Along with the `_form` partial at `app/views/wishlists/_form.html.erb`:
+
+```erb
+<%= form_with model: @wishlist do |form| %>
+  <% if form.object.errors.any? %>
+    <div><%= form.object.errors.full_messages.first %></div>
```
MatheusRich: I'm curious about first here. Is that how we've been doing on previous tutorials?

I usually would do .to_sentence but I guess you wan to show one error at a time?

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+    <%= link_to "Cancel", form.object.persisted? ? form.object : wishlists_path %>
+  </div>
+<% end %>
+```
+
+Create `show` next at `app/views/wishlists/show.html.erb`:
+
+```erb
+<h1><%= @wishlist.name %></h1>
+<% if authenticated? && @wishlist.user == Current.user %>
```
MatheusRich, line 615: it `authenticated?` necessary? Wishlist users can't be nil, so I guess this would be safe?
```suggestion
<% if @wishlist.user == Current.user %>
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+<button data-controller="clipboard" data-action="clipboard#copy" data-clipboard-text-value="/wishlists/1-example-wishlist">
+  Copy to clipboard
+</button>
+```
+
+What do these data attributes do? Let's break down each one:
+
+- `data-controller` tells Stimulus to connect to `clipboard_controller.js`
+- `data-action` tells Stimulus when the button is clicked to call the
+  `clipboard` controller's `copy()` function
```
MatheusRich: Not to be pedantic, but
```suggestion
  `clipboard` controller's `copy()` method
```

guides/source/wishlists.md
MatheusRich: Yes. I got confused by this. _"Why end users would be able to see subscribers for a product?"_

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+Our Wishlist model's friendly URLs are an important feature, so let's start by
+writing a test for it.
+
+Open `test/models/wishlist_test.rb` and add this test:
+
+```ruby
+require "test_helper"
+
+class WishlistTest < ActiveSupport::TestCase
+  test "has friendly urls" do
```
MatheusRich: IDK how I feel about this test. This seems mostly to be testing Rails code (which we should assume is tested). Maybe we could test that we can find it? Or maybe no test for this at all, since the integration specs will catch it.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1545 @@
...
+
+Next, we should test viewing wishlists, starting with a user viewing their own
+wishlist.
+
+```ruby
+test "view a wishlist" do
+  user = users(:one)
+  sign_in_as user
+  get wishlist_path(user.wishlists.first)
+  assert_response :success
```
MatheusRich, line 1449: This is a super basic test (which might be okay for a beginner tutorial), but I'd love if we at least checked that the wishlist name is on that page.

MatheusRich (approved):
Thank you @excid3! Great tutorial, as always. I learned something new today!

guides/source/wishlists.md
excid3: Authentication does not resume the session automatically when `allow_unauthenticated_access` is used currently.

https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt#L11

We ran into this in the GoRails Discord and created #53220. Also #55066 seems to address the same thing.

guides/source/wishlists.md
excid3: I can't remember, but I agree we should use `to_sentence` here.

```suggestion
    <div><%= form.object.errors.full_messages.to_sentence %></div>
```

guides/source/wishlists.md
excid3:
```suggestion
viewing wishlists and subscribers for a specific product.

Open `app/views/store/products/show.html.erb` and add the links:
```

guides/source/wishlists.md
excid3: Agreed, I'll swap this with a test for `filter_by` since that is something actually worth unit testing.

guides/source/wishlists.md
excid3: Adding an `assert_select "h1", text: wishlist.name` to these tests. 👍

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
+**DO NOT READ THIS FILE ON GITHUB, GUIDES ARE PUBLISHED ON <https://guides.rubyonrails.org>.**
+
+Wishlists
+=========
+
+This guide covers adding Wishlists to the store e-commerce application in the
```
AmandaPerino:
```suggestion
This guide covers adding Wishlists to the e-commerce application you created in the
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+E-commerce stores often have wishlists for sharing products. Customers can use
+wishlists to keep track of products they'd like to buy or share them with
+friends and family for gift ideas.
+
+Let's get started!
+
+Wishlist Models
+---------------
+
+Our e-commerce store has products and users that we built in the previous
```
AmandaPerino:
```suggestion
Our e-commerce store has products and users that we already built in the previous
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+
+Let's start by creating the `Wishlist` model.
+
+```bash
+$ bin/rails generate model Wishlist user:belongs_to name products_count:integer
+```
+
+This model has 3 attributes:
+
+- `user:belongs_to` which associates the `Wishlist` with the `User` record who
```
AmandaPerino:
```suggestion
- `user:belongs_to` which associates the `Wishlist` with the `User` who
```

AmandaPerino: Either: 
"associates the 'Wishlist' with the 'User' who owns it"
or
"associates the 'Wishlist' with the record of the 'User' who owns it"

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+  owns it
+- `name` which we'll also use for friendly URLs
+- `products_count` for the [counter cache](https://guides.rubyonrails.org/association_basics.html#counter-cache) we'll add to count how many products
```
AmandaPerino:
```suggestion
- `products_count` for the [counter cache](https://guides.rubyonrails.org/association_basics.html#counter-cache) to count how many products
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+`Wishlists` the product is on.
+
+```bash
+$ bin/rails generate migration AddWishlistsCountToProducts wishlists_count:integer
+```
+
+### Default Counter Cache Values
+
+Before we run these new migrations, let's set a default value for the counter
+cache columns.
```
AmandaPerino:
```suggestion
cache columns so that all existing records start with a count of zero instead of NULL.
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+  has_many :wishlist_products, dependent: :destroy
+  has_many :wishlists, through: :wishlist_products
+  has_one_attached :featured_image
+  has_rich_text :description
+```
+
+We added two associations to `Product`. First, we associate the `Product` model
+with the `WishlistProduct` join table. Using this join table, our second
+association tells Rails that a `Product` is a part of many `Wishlists` through
+the `WishlistProduct` join table. From a `Product` record, we can directly
```
AmandaPerino:
```suggestion
the same `WishlistProduct` join table. From a `Product` record, we can directly
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+access the `Wishlists` and Rails will know to automatically `JOIN` the tables in
+SQL queries.
+
+We also set `wishlist_products` as `dependent: :destroy`. When a `Product` is
+destroyed, it will be automatically removed from any Wishlists.
+
+In `app/models/wishlist.rb`, let's update both associations to enable counter
```
AmandaPerino: Can we explain counter caching here for context? Earlier we wrote "to count how many products are on the wishlist".

AmandaPerino:
```suggestion
A counter cache stores the number of associated records to avoid running a separate query each time the count is needed. So in `app/models/wishlist.rb`, let's update both associations to enable counter
```

AmandaPerino: Added suggestion.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+  validates :product_id, uniqueness: { scope: :wishlist_id }
+end
+```
+
+We've specified a column name to update on the associated models. For the
+`Product` model, we want to use the `wishlists_count` column and for `Wishlist` we
+want to use `products_count`. These counter caches update anytime a
+`WishlistProduct` is created or destroyed.
+
+The uniqueness validation also tells Rails to check if a product is already on
```
AmandaPerino: Should uniqueness be `uniqueness`?

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+Just like with `Product`, `wishlist_products` uses the `dependent: :destroy`
+option to automatically remove join table records when a Wishlist is deleted.
+
+### Friendly URLs
+
+Wishlists are often shared with friends and family. By default, the ID in the
+URL for a `Wishlist` is a simple Integer. This means we can't easily look at the
+URL to determine which `Wishlist` it's for.
+
+Active Record has a `to_param` class method that can be used for generating pretty
```
AmandaPerino:
```suggestion
Active Record has a `to_param` class method that can be used for generating more descriptive
```

AmandaPerino: Ignore if "pretty URLs" is a common term!

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+```irb
+store(dev)> reload!
+store(dev)> Wishlist.last.update(name: "A really, really long wishlist name!")
+store(dev)> Wishlist.last.to_param
+=> "1-a-really-really-long"
+```
+
+You can see that the name was truncated to the closest word to 20 characters.
+
+Alright, close the Rails console and start implementing wishlists in the UI.
```
AmandaPerino:
```suggestion
Alright, close the Rails console and let's start implementing wishlists in the UI.
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+
+## Adding Products To Wishlists
+
+The first place a user will probably use wishlists is on the `Product` show page.
+They'll likely be browsing products and want to save one for later. Let's begin
+by building that first.
+
+### Add To Wishlist Form
+
+Start by adding the route for this form to submit to:
```
AmandaPerino:
```suggestion
Start in `config/routes.rb` by adding the route for this form to submit to:
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+```
+
+Since we're in a nested resource route, we find the `Product` using the
+`:product_id` param.
+
+The `create` action is also simpler than normal. If a product is already on the
+wishlist, the `wishlist_product` record will fail to create but we don't need to
+notify the user of this error so we can redirect to the wishlist in either case.
+
+Log in as the user we created a wishlist for earlier and add a product to the
```
AmandaPerino:
```suggestion
Now, log in as the user we created a wishlist for earlier and try adding a product to the 
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+<% end %>
+```
+
+## Managing Wishlists
+
+Next, we need to be able to view and manage our wishlists.
+
+### Wishlists Controller
+
+Start by adding a route for wishlists at the top level:
```
AmandaPerino, line 481: Why is for highlighted here? Is that automatic?

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+        <%= link_to "Settings", settings_root_path %>
+        <%= button_to "Log out", session_path, method: :delete %>
+      <% else %>
+        <%= link_to "Sign Up", sign_up_path %>
+        <%= link_to "Login", new_session_path %>
+      <% end %>
+    </nav>
+```
+
+Refresh the page and click the "Wishlists" link in the navbar to manage your
```
AmandaPerino:
```suggestion
Refresh the page and click the "Wishlists" link in the navbar to view and manage your
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+wishlists.
+
+### Copy To Clipboard
+
+We can make sharing wishlists easier for our users by adding a Copy To Clipboard
```
AmandaPerino:
```suggestion
To make sharing wishlists easier, we can add a “Copy to Clipboard” button that uses a small amount of JavaScript. 
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+button with a little JavaScript.
```
AmandaPerino:
```suggestion
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+
+Since Rails ships with Hotwire on the frontend, we can use its
```
AmandaPerino:
```suggestion
Rails includes Hotwire by default, so we can use its [Stimulus framework](https://stimulus.hotwired.dev/)
 to add some lightweight JavaScript to our UI.
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+[Stimulus framework](https://stimulus.hotwired.dev/) to add some JavaScript
```
AmandaPerino:
```suggestion
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+functionality to our UI.
```
AmandaPerino:
```suggestion
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+```html
+<button data-controller="clipboard" data-action="clipboard#copy" data-clipboard-text-value="/wishlists/1-example-wishlist">
+  Copy to clipboard
+</button>
+```
+
+What do these data attributes do? Let's break down each one:
+
+- `data-controller` tells Stimulus to connect to `clipboard_controller.js`
+- `data-action` tells Stimulus when the button is clicked to call the
```
AmandaPerino:
```suggestion
- `data-action` tells Stimulus to call the
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+  `clipboard` controller's `copy()` method
```
AmandaPerino:
```suggestion
  `clipboard` controller's `copy()` method when the button is clicked
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+```
+
+This Stimulus controller is short. It does two things:
+
+- Registers `text` as a value so we can access it. This is the URL we want to
+  copy to the clipboard.
+- The `copy` function writes the `text` from the HTML to the clipboard when
+  called.
+
+You'll notice we didn't have to add any event listeners or setup & teardown this
```
AmandaPerino:
```suggestion
If you're familiar with JavaScript, you'll notice we didn't have to add any event listeners or setup & teardown this
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+
+To learn more about Stimulus, check out the
+[Stimulus](https://stimulus.hotwired.dev/) website.
+
+### Removing Products
+
+A user may purchase or lose interest in a product and want to remove it from
+their wishlist. Let's add that feature next.
+
+Let's update the wishlists route to contain a nested resource.
```
AmandaPerino:
```suggestion
First we'll update the wishlists route to contain a nested resource.
```

AmandaPerino: This was to remove the "Let's" since it was in 2 back to back sentences.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+    <small>Added <%= l it.created_at, format: :long %></small>
+
+    <% if authenticated? && @wishlist.user == Current.user %>
+      <%= button_to "Remove", [ @wishlist, it ], method: :delete, data: { turbo_confirm: "Are you sure?" } %>
+    <% end %>
+  </div>
+<% end %>
+```
+
+Create `app/controllers/wishlists/wishlist_products_controller.rb` with the
```
AmandaPerino:
```suggestion
Create `app/controllers/wishlists/wishlist_products_controller.rb` and add the
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+<%= form_with url: store_wishlists_path, method: :get do |form| %>
+  <%= form.collection_select :user_id, User.all, :id, :full_name, selected: params[:user_id], include_blank: "All Users" %>
+  <%= form.collection_select :product_id, Product.all, :id, :name, selected: params[:product_id], include_blank: "All Products" %>
+  <%= form.submit "Filter" %>
+<% end %>
+
+<%= render @wishlists %>
+```
+
+We've updated the header to display a count of wishlists which will be helpful
```
AmandaPerino: This paragraph is confusing me, specifically "The form also reads from the URL's query params to
pre-select an option in the select boxes if a filter is applied." Does the following still make sense?

We’ve updated the header to show the total number of wishlists, which makes it easier to see how many results match when a filter is applied. When you submit the form, Rails adds your selected filters to the URL as query params. The form then reads those values when loading the page to automatically re-select the same options in the dropdowns, so your choices stay visible after submitting. Since the form submits to the index action, so it can display either _all_ wishlists or just the filtered results.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+    @wishlists = @wishlists.includes(:wishlist_products).where(wishlist_products: { product_id: params[:product_id] }) if params[:product_id].present?
+  end
+
+  def show
+    @wishlist = Wishlist.find(params[:id])
+  end
+end
+```
+
+Active Record queries are lazy evaluated which means SQL queries aren't executed
```
AmandaPerino:
```suggestion
Active Record queries are _lazy evaluated_ which means SQL queries aren't executed
```

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+until you ask for the results. This allows our controller to build up the query
+step-by-step and include filters if needed.
+
+Try filtering wishlists to a specific user or product or both.
```
AmandaPerino: What should they see if they do this? Right now we don't have many users or products or wishlists. Maybe we should reword this part to indicate what will happen (in the future)?

Once you have more wishlists in the system, you can use the filters to view wishlists by a specific user, product, or a combination of both.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+  end
+end
+```
+
+`filter_by` is almost the same as what we had in the controller, but we start by
+calling
+[`all`](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-all)
+which returns an `ActiveRecord::Relation` for all the records including any
+conditions we may have already applied. Then we apply the filters and return the
+results
```
AmandaPerino:
```suggestion
results. 

Refactoring like this means the controller becomes cleaner, while the filtering logic now lives in the model where it belongs, alongside other database-related logic. This follows the **Fat Model, Skinny Controller** principle, a best practice in Rails.
```

AmandaPerino: It felt like this section needed a way to explain why we just did what we did.

guides/source/wishlists.md
```diff
@@ -0,0 +1,1653 @@
...
+
+## Adding Subscribers To Admin
+
+While we're here, we should also add the ability to view subscribers in the
```
AmandaPerino:
```suggestion
While we're here, we should also add the ability to view and filter subscribers in the
```

AmandaPerino: I was a little confused by this section here. 
We added subscribers in the Getting Started tutorial, and then the admin access in the Sign Up & Settings add-on tutorial. 
What functionality did we just learn that allows us to implement this section now? Was it the filtering?

AmandaPerino (changes requested):
Hey @excid3 - done reviewing! Most of my comments are tiny tweaks for readability or adding context that I felt was missing (as a reader/learner), but there were a few comments where I was a little more confused, most notably 'Adding Subscribers to Admin'.

AmandaPerino (changes requested):
Hey @excid3 - done reviewing! Most of my comments are tiny tweaks for readability or adding context that I felt was missing (as a reader/learner), but there were a few comments where I was a little more confused, most notably 'Adding Subscribers to Admin'.

guides/source/wishlists.md
excid3: Looks to be a bug in GitHub.
Screenshot 2025-10-15 at 12 44 51 PM

rafaelfranca (approved):

PR #55432, opened 2025-07-30
Fix distinct count for composite primary key models

robzolkos:
### Motivation / Background

This Pull Request has been created because Rails throws an SQL error when calling `distinct.count` on models with composite primary keys. The issue occurs because the calculation methods don't properly handle arrays of columns when performing distinct counts, resulting in malformed SQL queries.

Fixes #55401

### Detail

This Pull Request changes the calculation logic in `ActiveRecord::Relation::Calculations` to properly handle composite primary keys when performing distinct count operations. The key changes include:

1. Modified `aggregate_column` method to handle arrays of columns (composite keys) by converting them to Arel nodes
2. Updated `perform_calculation` to properly handle composite primary keys for distinct count operations.
3. Enhanced `operation_over_aggregate_column` to wrap multiple columns in a single count operation.
4. Modified `execute_simple_calculation` to use a subquery approach for distinct counts on composite primary keys, which ensures proper SQL generation

The fix ensures that queries like `Model.distinct.count` work correctly on models with composite primary keys by generating valid SQL using subqueries.

 ### Additional information

The solution uses a subquery approach for composite primary key distinct counts to ensure compatibility across different database adapters. This approach generates SQL like:

  ```sql
  SELECT COUNT(*) FROM (SELECT DISTINCT "column1", "column2" FROM "table") t
  ```

A test case has been added to verify the fix works correctly with the existing composite primary key test models in the Rails test suite.

Checklist

- [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: [Fix #55401]


activerecord/lib/active_record/relation/calculations.rb
```diff
@@ -478,7 +487,15 @@ def possible_aggregation?(column_names)
...
       def operation_over_aggregate_column(column, operation, distinct)
-        operation == "count" ? column.count(distinct) : column.public_send(operation)
+        if operation == "count"
+          if column.is_a?(Array)
+            Arel::Nodes::Count.new(column, distinct)
+          else
+            column.count(distinct)
+          end
+        else
+          column.public_send(operation)
```
akostadinov, line 496: I don't know whether it is possible to have an array of columns for non-count operations. I guess in case that is possible, it would be reported and dealt with in another PR.

But it would be nice if you know the answer off the top of your head :)

And thank you for looking into this!

PR #50858, opened 2024-01-23
Support options for ActionMailer delivery callbacks

fatkodima:
Closes https://github.com/rails/rails/issues/50830.

The code works, but docs/tests etc are currently missing. I want to validate the approach before proceeding.

Inside action mailer I am reusing private methods from the implementation of callbacks from the action controller, see https://github.com/rails/rails/blob/main/actionpack/lib/abstract_controller/callbacks.rb.

The questions to consider are:
1. Should the mailer callbacks really depend on the internals of controller callbacks, considering that it already includes it https://github.com/rails/rails/blob/776626ff987a96201b0bdbd86d716ca6698fa8b3/actionmailer/lib/action_mailer/base.rb#L493
2. Controllers callbacks have a `raise_on_missing_callbacks_action`, see https://github.com/rails/rails/blob/776626ff987a96201b0bdbd86d716ca6698fa8b3/actionpack/lib/abstract_controller/callbacks.rb#L36 which can be set via `config.action_controller. raise_on_missing_callbacks_action`. This kinda makes sense for action mailer too, so should we introduce `config.action_mailer.raise_on_missing_callbacks_action` config? 
3. Controller callbacks use the word `"controller"`, see https://github.com/rails/rails/blob/776626ff987a96201b0bdbd86d716ca6698fa8b3/actionpack/lib/abstract_controller/callbacks.rb#L54, while for mailer we should use the `"mailer"` word or should we make the message text more generic? 

cc @bensheldon

PR #55434, opened 2025-07-31
ActionMailer: add support for callback options

### LINKED ISSUE #50830: Add support for `ActionMailer.*_delivery` callback `only` and `except` options
### Steps to reproduce

Currently ActionMailer's `*_delivery` callbacks do not support `only` and `except` filters, which are quite useful because some callback actions are required for a subset of mailer methods. Technically, only/except can be emulated via `if` option, but another problem is that's not clear from the documentation which options can be passed (currently `only` and `except` are simply ignored).

I believe that many developers might except `ActionMailer` to behave just like `ActionController`,  so current API implementation might lead to unexpected bugs.

Note that documentation on `before_action` (and similar callbacks) explicitly refer to possible options:

> before_action(names, block)[Link](https://api.rubyonrails.org/v7.1.3/classes/AbstractController/Callbacks/ClassMethods.html#method-i-before_action)
> Append a callback before actions. See [_insert_callbacks](https://api.rubyonrails.org/v7.1.3/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_insert_callbacks) for parameter details.

### Expected behavior
At least some explicit info on possible `*_delivery` callback options, or (which would be better) support for `only` and `except` options.

### Actual behavior
Mentioned options are simply ignored, no documentation on possible options.

### System configuration
**Rails version**: 7.1.3

**Ruby version**: 3.0.6

viralpraxis:
### Motivation / Background
Resolves https://github.com/rails/rails/issues/50830

### Detail

This PR is based on https://github.com/rails/rails/pull/50858.

Aside of the mentioned PR which re-uses ActionController's callbacks, a new option `action_mailer.raise_on_missing_callback_actions`
is added. I don't know which default value is preferred so it's `false` for now.

### Checklist


viralpraxis:
Could you take a look please? @fatkodima

viralpraxis:
CI seems to be green (some flaky tests). Any chances for a review? I'm still interested in this.

rafaelfranca:
Thank you. Will merge in #57581

PR #55435, opened 2025-08-01
Fix ActiveSupport::Rescuable not to be coupled with ActionController

byroot:
Followup on https://github.com/rails/rails/pull/55424

I mistakenly added an Action Controller event in the middle of Active Support.

Instead we can decorate the necessary method.

In addition, we can trigger the event before invoking the handler.


rafaelfranca:
❤️

PR #55440, opened 2025-08-02
Fix outdated api.rubyonrails.org links in Testing Guide [ci skip]

Bilka2:
Fix that old Rails versions were included in two api.rubyonrails.org links in the Testing Rails Applications guide. The updated links contain the same docs as the original links, just more up-to-date.


jonathanhefner:
Thank you, @Bilka2! ⬆️

PR #55441, opened 2025-08-03
Fix badly anchored regular expression

kreeksec:
https://github.com/rails/rails/blob/add5a73b26e78d6b13945525874749ae40af21c7/activesupport/lib/active_support/core_ext/module/attribute_accessors_per_thread.rb#L43-L43

fix the problem, the regular expression used to validate the attribute name should use the `\A` and `\z` anchors instead of `^` and `$`. This ensures that the entire string is matched, not just a single line. The change should be made on line 43 of `activesupport/lib/active_support/core_ext/module/attribute_accessors_per_thread.rb`, replacing `/^[_A-Za-z]\w*$/` with `/\A[_A-Za-z]\w*\z/`. No additional imports or method definitions are needed.

### References
Ruby documentation: [Anchors](https://ruby-doc.org/3.2.0/Regexp.html#class-Regexp-label-Anchors)
### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/module/attribute_accessors_per_thread.rb
+++ b/activesupport/lib/active_support/core_ext/module/attribute_accessors_per_thread.rb
@@ -40,7 +40,7 @@ class Module
   #   Current.new.user # => NoMethodError
   def thread_mattr_reader(*syms, instance_reader: true, instance_accessor: true, default: nil) # :nodoc:
     syms.each do |sym|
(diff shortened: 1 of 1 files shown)


skipkayhil:
This appears to be part of a series of low value, AI generated PRs opened against many different open source repositories...

For this change to be useful, a developer would have to intentionally pass a multi-line string to this method even though its documented as accepting symbols.

PR #55443, opened 2025-08-03
Fix incorrect statement about object destruction [ci skip]

yazanalkherret:
Replace the incorrect statement "This does not destroy the objects" with "This destroys the objects" to match the API Docs.

### Motivation / Background

This Pull Request has been created because the Rails guides had a factual error in  [Active Record Associations Section #2.6.1.1](https://guides.rubyonrails.org/association_basics.html#methods-added-by-has-and-belongs-to-many-managing-the-collection). 

### Detail
The guides originally state that [`collection.destroy`](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy) does NOT destroy the  objects. According to the [API Docs](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy), it does destroy the objects.

This Pull Request changes the incorrect statement "This does not destroy the objects" to "This destroys the objects" in order to match the API Docs.

### Checklist

### DIFF
--- a/guides/source/association_basics.md
+++ b/guides/source/association_basics.md
@@ -1325,7 +1325,7 @@ objects.
 ```
 
 The [`collection.destroy`][] method removes one or more objects from the
-collection by deleting records in the join table. This does not destroy the
+collection by deleting records in the join table. This destroys the
 objects.
 
 ```ruby


fatkodima:
It is correct. It destroys records from the join table, but not the records from the associated table.

PR #46603, opened 2022-11-28
Add `perform_all_later` to enqueue multiple jobs at once

Mangara:
### Motivation / Background

Sidekiq has a useful optimisation called `push_bulk` that enqueues many jobs at once, eliminating the repeated Redis roundtrips. However, this feature is not exposed through Active Job, so it only works for `Sidekiq::Worker` jobs. This adds a barrier to Active Job adoption for apps that rely on this feature. It also makes it harder for other queue adapters to implement similar functionality, as they then have to take care of serialization, callbacks, etc. themselves.

### Detail

This PR adds `ActiveJob.perform_all_later([<job 1>, <job2>])`, backed by Sidekiq's `push_bulk` and with a fallback to enqueuing serially if the queue adapter does not support bulk enqueue.

#### Arguments

It's hard to support the full diversity of ways a Ruby method can be called with a simple array. In particular, jobs with a mix of positional and keyword arguments are difficult to distinguish from jobs with positional arguments that include a hash. Sidekiq gets around this by not supporting keyword arguments.

While it is possible to support everything by treating the last hash in the array of arguments as keywords and requiring jobs with a hash as last positional argument to include an additional empty hash of keyword arguments, based on feedback on the PR, we chose to pass instantiated jobs, so we can let Ruby handle all these complexities.

#### Delay

Passing in instantiated jobs also makes it easy for the user to specify delays for each job:

```ruby
user_jobs = users.map.with_index do |user, user_index|
  UserJob.new(user).set(wait: user_index.seconds)
end
ActiveJob.perform_all_later(user_jobs)
```

#### Return value

The return value of `perform_all_later` is limited by Sidekiq's current behaviour. `push_bulk` returns the (provider) ids of all jobs that were successfully enqueued. This means that if we try to enqueue 2 jobs, we may get back a single id with no way of knowing which job it belongs to, so we can't even map it back to Active Job's job ids. I chose to return the number of successfully enqueued jobs, but it may be better to always return `true` for now, so that we can more easily change it later? In an ideal world, I think the return value would be an array of either the job or `false`, to mirror the return value of `perform_later`.

#### Callbacks

Based on feedback in this PR, `perform_all_later` does not run any callbacks. This is in line with the Active Record bulk methods, so it shouldn't be too surprising to users and it is clearly stated in the documentation.

There are several issues with running callbacks:

[...]

PR #55444, opened 2025-08-03
Remove `ActiveJob::ConfiguredJob#perform_all_later` method

fatkodima:
Follow up to https://github.com/rails/rails/pull/46603 (cc @Mangara)

The removed method is currently broken and raises an error
```
undefined method 'perform_all_later' for class TestJob (NoMethodError)

      @job_class.perform_all_later(multi_args, options: @options)
                ^^^^^^^^^^^^^^^^^^
Did you mean?  perform_later
```

The intention of that method was probably to use it like this:
```ruby
TestJob.set(wait: 1.hour).perform_all_later([[job1_arg], [job2_arg]])
```

But then the syntax for `TestJob.perform_all_later([[job1_arg], [job2_arg]])` (note the job name usage instead of `ActiveJob`) should be implemented too, for consistency. 

But I think, this is really not needed, because existing syntax (`ActiveJob.perform_all_later(job1, job2)`) is flexible enough to handle both cases.

### DIFF
--- a/activejob/lib/active_job/configured_job.rb
+++ b/activejob/lib/active_job/configured_job.rb
@@ -19,9 +19,5 @@ module ActiveJob
 
       enqueue_result
     end
-
-    def perform_all_later(multi_args)
-      @job_class.perform_all_later(multi_args, options: @options)
-    end
   end
 end


Mangara (commented):
Yep, this can be safely removed. It's leftover from an older iteration of the PR that had a different API. Apologies for the oversight!

PR #55445, opened 2025-08-04
Update gmt? return true when TZ='GMT'

### LINKED ISSUE #55360: `gmt?` dose not return `true` when TZ='GMT'
The [ActiveSupport::TimeWithZone#utc?](https://github.com/rails/rails/blob/781394fc2316526c6767845a388301f7fa7cc710/activesupport/lib/active_support/time_with_zone.rb#L99-L108) has an alias named `gmt?`.
However, when used with an `ActiveSupport::TimeWithZone` object whose timezone is `'GMT'`, it returns `false`.
This is because the method currently returns `true` only when the timezone is `'UTC'` or `'UCT'`.

The method name and implementation are inconsistent, so I propose adding a check for `'GMT'`.

Additionally, the `utc?` method checks the string returned by `TZInfo::OffsetTimezonePeriod#abbreviation`. 
However, as shown below, the `'UCT'` timezone actually returns `'UTC'`.

```rb
Time.current.in_time_zone('UTC').period.abbreviation 
# => "UTC"

Time.current.in_time_zone('UCT').period.abbreviation 
# => "UTC"
```

Since checking for `'UCT'` is redundant, I propose replacing the `'UCT'` check with a `'GMT'` check to make the `gmt?` alias method work correctly.

```rb
def utc?
  zone == "UTC" || zone == "GMT"
end
alias_method :gmt?, :utc?
```

By the way, I think `'gmt?'` alias does not need to delete and implement as new method, because `'GMT'` and `'UTC'` is different objects, though both are almost the same functionality.

```rb
Time.current.in_time_zone('UTC').period
# => #<TZInfo::OffsetTimezonePeriod:0x0000ffff6b65a660
#  @offset=
#   #<TZInfo::TimezoneOffset: @base_utc_offset=0, @std_offset=0, @abbreviation=UTC>>

Time.current.in_time_zone('GMT').period
# => #<TZInfo::OffsetTimezonePeriod:0x0000ffff6a132620
#  @offset=
#   #<TZInfo::TimezoneOffset: @base_utc_offset=0, @std_offset=0, @abbreviation=GMT>>
```

halllzA:
### Motivation / Background

This Pull Request has been created because it resolve https://github.com/rails/rails/issues/55360.

### Detail

This Pull Request changes `ActiveSupport::TimeWithZone#utc?` to check for `GMT` instead of `UCT`.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -274,4 +274,8 @@
 
     *Jonathan del Strother*
 
+*   `ActiveSupport::TimeWithZone# utc?` to remove `TZ='UCT'` and add `TZ='GMT'.
+
+    *Haruki Kobayashi*
+
 Please check [8-0-stable](https://github.com/rails/rails/blob/8-0-stable/activesupport/CHANGELOG.md) for previous changes.
--- a/activesupport/lib/active_support/time_with_zone.rb
+++ b/activesupport/lib/active_support/time_with_zone.rb
@@ -103,7 +103,7 @@ module ActiveSupport
     #   Time.zone = 'Eastern Time (US & Canada)'    # => 'Eastern Time (US & Canada)'
     #   Time.zone.now.utc?                          # => false
     def utc?
-      zone == "UTC" || zone == "UCT"
+      zone == "UTC" || zone == "GMT"
     end
     alias_method :gmt?, :utc?
 
--- a/activesupport/test/core_ext/time_with_zone_test.rb
+++ b/activesupport/test/core_ext/time_with_zone_test.rb
@@ -70,15 +70,17 @@ class TimeWithZoneTest < ActiveSupport::TestCase
     assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["UCT"]).utc?
     assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Etc/UCT"]).utc?
     assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Etc/Universal"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["GMT"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["GMT0"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Etc/GMT-0"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Abidjan"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Banjul"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Freetown"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Greenwich"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Iceland"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Monrovia"]).utc?
 
-    assert_equal false, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Abidjan"]).utc?
-    assert_equal false, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Banjul"]).utc?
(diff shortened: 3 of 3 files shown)


activesupport/lib/active_support/time_with_zone.rb
```diff
@@ -103,7 +103,7 @@ def dst?
     #   Time.zone = 'Eastern Time (US & Canada)'    # => 'Eastern Time (US & Canada)'
     #   Time.zone.now.utc?                          # => false
     def utc?
-      zone == "UTC" || zone == "UCT"
+      zone == "UTC" || zone == "GMT"
```
byroot, line 106:
```suggestion
      zone == "UTC" || zone == "GMT" || zone == "UCT"
```

I don't think it's worth not checking for `UCT`, that introduce backward compatibility concerns, so might as well keep it, it's not an expensive check.

activesupport/CHANGELOG.md
amitleshed, line 1: It might be worth making this a bit more descriptive to clarify the behavior change for readers going through the log.

Feel free to use this version if helpful:

```
*   Fix `ActiveSupport::TimeWithZone#utc?` (and its alias `gmt?`)
    so it returns `true` when the zone is `"GMT"`. Previously a
    `TimeWithZone` in the "GMT" zone incorrectly reported `gmt?`
    as `false`. (*halllzA*)
```

activesupport/test/core_ext/time_with_zone_test.rb
```diff
@@ -70,15 +70,17 @@ def test_utc?
...
-    assert_equal false, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Monrovia"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["GMT"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["GMT0"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Etc/GMT-0"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Abidjan"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Banjul"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Freetown"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Greenwich"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Iceland"]).utc?
+    assert_equal true, ActiveSupport::TimeWithZone.new(Time.utc(2000), ActiveSupport::TimeZone["Africa/Monrovia"]).utc?
```
byroot, line 81: Actually this change is very fishy.

byroot: I guess it comes down to the meaning of `utc?` vs `gmt?`.  Perhaps the solution is to stop `gmt?` being an alias of `utc?` as the meaning isn't quite the same. But I'm a bit rusty on timezones.

halllzA: I also don't have a deep understanding of timezones, but at least the impact of this fix could have a significant effect in some countries. I also think making `gmt?` an independent method would be better.

byroot: Yes, it's really about the meaning of GMT vs UTC. In a way, UTC means "no timezone offset", and all these countries have "timezone offset 0" all year long, so are effectively GMT aliases. But doesn't mean `GMT == UTC`, at least no more than `nil == 0`.

The problem here really is what is the backward compatibility impact of changing this semantic?

Maybe the better solution is just to deprecate that alias, I'm really not sure what the point of it is, but I may be missing something.

byroot: Nevermind, Ruby also has that alias, so our hands are tied: https://docs.ruby-lang.org/en/master/Time.html#method-i-gmt-3F

I'm afraid the best course of action is to do nothing.

byroot: Ref: https://bugs.ruby-lang.org/issues/11595

halllzA: Ruby's `Time#utc?` behavior appears to have [a clear policy of treating GMT as equivalent to UTC for historical reasons.](https://ruby-doc.org/core-2.6.8/Time.html#:~:text=The%20Time%20class%20treats%20GMT%20(Greenwich%20Mean%20Time)%20and%20UTC%20(Coordinated%20Universal%20Time)%20as%20equivalent.%20GMT%20is%20the%20older%20way%20of%20referring%20to%20these%20baseline%20times%20but%20persists%20in%20the%20names%20of%20calls%20on%20POSIX%20systems.) (This is documentation from Ruby 2.6.x, so the information may be outdated.)  
Actually, there are methods like `getutc` and `utc` that convert timezones to UTC, and `getgm` and `gmtime` are provided as their aliases.

I mentioned in the issue that there was inconsistency between the method name and its behavior, but I think creating a gap with Ruby's implementation policy in the TimeWithZone methods, which are extensions of the Time class, would be a bigger problem.

Therefore, I also think doing nothing is the best approach for the current situation.

halllzA: I have come to understand once again that contributing to Rails also requires deep knowledge of Ruby.
This has been a very educational experience! Thank you!

PR #55447, opened 2025-08-04
Enable hooking LocalCache into Executor

skipkayhil:
Previously, Rails would automatically add the `LocalCache::Middleware`
to the stack if `Rails.cache` is configured to use a `Store` that
supports `LocalCache` (currently, `MemCacheStore`, `RedisCacheStore`,
and `NullStore`).

This commit introduces an additional option to the configuration added
in a previous commit (`config.local_cache_store_strategy`), `:executor`. When
using this option, Rails will register `run` and `complete` hooks on the
application's executor instead of using a middleware.

While Rails automatically adds a middleware for requests to take
advantage of the `LocalCache`, there is no such built in mechanism for
Active Job. Similarly, regular `ActiveSupport::TestCase` tests will not
end up using the `LocalCache` even though they may be testing behavior
that _would_ go through the `LocalCache` during a real request. By using
the executor for configuring the `LocalCache`, requests, job, tests, and
any other application code wrapped in the executor will be able to take
advantage of it. This is also already the behavior of Active Record's
in-memory-cache, so this would align `Rails.cache`'s in-memory-cache
with it.

Finally, using the executor means we have one less middleware in the
request stack and one less BodyProxy allocated and wrapped around the
response, which makes me happy.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/cache/strategy/local_cache.rb
+++ b/activesupport/lib/active_support/cache/strategy/local_cache.rb
@@ -18,13 +18,11 @@ module ActiveSupport
           extend self
 
           def cache_for(local_cache_key)
-            registry = ActiveSupport::IsolatedExecutionState[:active_support_local_cache_registry] ||= {}
-            registry[local_cache_key]
+            ExecutionContext.local_cache_registry[local_cache_key]
           end
 
           def set_cache_for(local_cache_key, value)
-            registry = ActiveSupport::IsolatedExecutionState[:active_support_local_cache_registry] ||= {}
-            registry[local_cache_key] = value
+            ExecutionContext.local_cache_registry[local_cache_key] = value
           end
         end
 
@@ -87,6 +85,14 @@ module ActiveSupport
         # duration of request.
         def middleware
           @middleware ||= Middleware.new("ActiveSupport::Cache::Strategy::LocalCache", self)
+        end
+
+        def install_executor_hooks(executor = ActiveSupport::Executor)
+          key = local_cache_key
+
+          executor.to_run do
+            LocalCacheRegistry.set_cache_for(key, LocalStore.new)
+          end
         end
 
         def clear(options = nil) # :nodoc:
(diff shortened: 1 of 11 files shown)


skipkayhil:
~`./tools/railspect config .` will fail without https://github.com/rails/rails/pull/55446~

Edit: `./tools/railspect config .` passes now ✅

byroot:
The idea make sense to me. An executor hook is preferable to a middleware. But I'll need to find some time to review this carefully.

There might be a concern with the test environment, where the executor can be nested.

rafaelfranca (commented):
I think this change makes sense, but did you try to run in our application to see if it would work?

activesupport/lib/active_support/cache/strategy/local_cache.rb
```diff
@@ -63,6 +63,20 @@ def fetch_entry(key) # :nodoc:
...
         end
 
+        class ExecutorHooks
+          def initialize(local_cache_key)
+            @local_cache_key = local_cache_key
+          end
+
+          def run
+            LocalCacheRegistry.set_cache_for(@local_cache_key, LocalStore.new)
+          end
```
byroot: So I'm afraid this won't play well with https://github.com/rails/rails/pull/55247/

But if you refactor ` LocalCacheRegistry.set_cache_for` to use the execution context, that should allow nesting in test environment.

You can look at what is done for `CurrentAttributes`

skipkayhil: What do you think about https://github.com/rails/rails/pull/55482?

That should make the change in this PR something like:

```ruby
module LocalCacheRegistry
  extend self

  def cache_for(local_cache_key)
    registry = Execution[:active_support_local_cache_registry] ||= {}
    registry[local_cache_key]
  end

  def set_cache_for(local_cache_key, value)
    registry = Execution[:active_support_local_cache_registry] ||= {}
    registry[local_cache_key] = value
  end
end
```

byroot: > What do you think about #55482?

IMO `Execution` alone doesn't make sense as a name, but also it seems to me your PR adds one more indirection and this codepath is really superhot, so I'd rather not change it.

skipkayhil:
Updated to use `ExecutionContext` so that it will work with requests/jobs within tests.

> but did you try to run in our application to see if it would work?

I'll test it with our application now 👍

PR #55453, opened 2025-08-05
docs: document when error raising is skipped for strict loading

### LINKED ISSUE #55128: strict_loading ignored in validation method
### Steps to reproduce

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "activerecord", "8.0.2"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')

  create_table "comments", force: :true do |t|
    t.integer "post_id"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.index ["post_id"], name: "index_comments_on_post_id"
  end

  create_table "posts", force: :true do |t|
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
  end

  add_foreign_key "comments", "posts"
end


  self.abstract_class = true
end

  has_many :comments

  validate :has_valid_comments

  def has_valid_comments
    comments.each do |c|
      # loops on comment
      # Should trigger an error if not included
    end
  end
end

class Comment < ApplicationRecord
  has_one :post
end


class StrictLoadingTestTest < Minitest::Test
  def setup
    Post.create!(comments: [Comment.create!, Comment.create!])
  end

  def test_strict_loading_instance
    post = Post.first
    post.strict_loading!
    assert_raises(ActiveRecord::StrictLoadingViolationError) do
      # Raises as expected
      post.comments.to_a
    end
  end

  def test_strict_loading_validation_should_raise
    post = Post.first
    post.strict_loading!
    assert_raises(ActiveRecord::StrictLoadingViolationError) do
      # Does not raise, but I'd expect to
      post.valid?
    end
  end

  def test_strict_loading_calling_the_validation_method_manually
    post = Post.first
    post.strict_loading!
    assert_raises(ActiveRecord::StrictLoadingViolationError) do
      # Raises as expected
      post.has_valid_comments
    end
  end
end
```

### Expected behavior

I expect ActiveRecord::StrictLoadingViolationError to be raised when validating a Post instance.

This is maybe expected, the validation workflow may have a special behaviour on this. If it's the case, I did not find any mention of it in the strict_loading documentations.

### Actual behavior

ActiveRecord::StrictLoadingViolationError is not raised when the relation is loaded from the validation workflow.

### System configuration

**Rails version**: tested on 7.1.3.4 and 8.0.2

**Ruby version**: tested on 3.3.6 and 3.4.2

jordanmoore753:
### Motivation / Background

Strict loading error raising is skipped during validation, but this behavior was undocumented. This commit adds a note clarifying that validation contexts do not trigger strict loading violations, to help prevent confusion.

Fixes: #55128

### Detail

Docs were not updated after this commit: https://github.com/rails/rails/commit/bda9bc013b63

### Checklist

### DIFF
--- a/activerecord/lib/active_record/core.rb
+++ b/activerecord/lib/active_record/core.rb
@@ -690,6 +690,8 @@ module ActiveRecord
     # Sets the record to strict_loading mode. This will raise an error
     # if the record tries to lazily load an association.
     #
+    # NOTE: Strict loading is disabled during validation in order to let the record validate its association.
+    #
     #   user = User.first
     #   user.strict_loading! # => true
     #   user.address.city


activerecord/lib/active_record/core.rb
```diff
@@ -690,6 +690,8 @@ def strict_loading?
     # Sets the record to strict_loading mode. This will raise an error
     # if the record tries to lazily load an association.
     #
+    # Error raising for strict loading is skipped during validation.
```
Edouard-chin:
```suggestion
    # NOTE: Strict loading is disabled during validation in order to let the record validate its association.
```

Edouard-chin (commented):
Thanks for the PR. I made a suggestion as I think it's a bit more comprehensive with a small explanation on why we disable it.

activerecord/lib/active_record/core.rb
jordanmoore753: Makes sense! For future reference, when should one be using the `NOTE` at the start of the documentation? Is it when it's documenting an edge case or additional behavior?

jordanmoore753: Updated according to suggestion, commit: [8a2fd9c](https://github.com/rails/rails/commit/8a2fd9c0abe28ae770fe828cdf0de6d8d190f1e6)

Edouard-chin: > When should one be using the NOTE at the start of the documentation?

For markdown documentation, NOTE is useful to have a nice banner like this one.
image

For API docs, such as in this case, I think it's mostly to highlight something you believe the reader shouldn't miss.

Edouard-chin:
Thank you!

activerecord/lib/active_record/core.rb
jordanmoore753: Ah I see 👍  Thank you for the explanation.

PR #55457, opened 2025-08-06
Add support to `SchemaDumper.ignore_tables` properly ignore tables in non-public schemas for PostgreSQL

sobrinho:
### Motivation / Background

This Pull Request has been created because I want to ignore tables on `db/structure.sql` that are not on the `public` schema in my PostgreSQL database.

### Detail

This Pull Request changes the `ActiveRecord::StructureDump.ignore_tables` to accept a hash to specify the schema and changes the `pg_dump -T` argument as well so qualified tables can be ignored.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add support to `SchemaDumper.ignore_tables` properly ignore tables in non-public schemas for PostgreSQL.
+
+    *Gabriel Sobrinho*
+
 *   Add `connection.current_transaction.isolation` API to check current transaction's isolation level.
 
     Returns the isolation level if it was explicitly set via the `isolation:` parameter
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -658,6 +658,14 @@ module ActiveRecord
           end
         end
 
+        def data_sources(include_schema: false)
+          if include_schema
+            query(data_source_sql(include_schema: true), "SCHEMA")
+          else
+            query_values(data_source_sql, "SCHEMA")
+          end
+        end
+
         def foreign_tables
           query_values(data_source_sql(type: "FOREIGN TABLE"), "SCHEMA")
         end
@@ -1156,11 +1164,13 @@ module ActiveRecord
               raise(ArgumentError, "Table '#{table_name}' has no unique constraint for #{column || options}")
           end
 
-          def data_source_sql(name = nil, type: nil)
+          def data_source_sql(name = nil, type: nil, include_schema: false)
             scope = quoted_scope(name, type: type)
             scope[:type] ||= "'r','v','m','p','f'" # (r)elation/table, (v)iew, (m)aterialized view, (p)artitioned table, (f)oreign table
 
-            sql = +"SELECT c.relname FROM pg_class c LEFT JOIN pg_namespace n ON n.oid = c.relnamespace"
+            sql = +"SELECT"
+            sql << " n.nspname, " if include_schema
+            sql << " c.relname FROM pg_class c LEFT JOIN pg_namespace n ON n.oid = c.relnamespace"
             sql << " WHERE n.nspname = #{scope[:schema]}"
             sql << " AND c.relname = #{scope[:name]}" if scope[:name]
             sql << " AND c.relkind IN (#{scope[:type]})"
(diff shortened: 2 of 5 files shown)


sobrinho:
This implementation is broken, just a PoC.

Please guide me on the best approach and I will update the PR.

# Use-case

We have range partitioned tables managed by `pg_partman` in our database and we don't wan't local partitions to be dumped into `db/structure.sql` since it will always be different from one machine to another.

We also have 99% of the tables on a schema other than `public` and we can't ignore any of them using `SchemaDumper.ignore_tables` because of that.

# The issue

`connection.data_sources` does not specify the schema a table belongs to and therefore it is impossible to filter tables other than the public schema ones:

```ruby
ActiveRecord::Base.connection.data_sources
#=> ["foo", "bar", "prefix_foo", "ignored_foo"]
```

If `foo` is under `myschema` instead of `public`, this doesn't work:

```ruby
ActiveRecord::SchemaDumper.ignore_tables = ["foo"]
```

Since the final `pg_dump` call will be something like:

```bash
pg_dump --schema-only --no-privileges --no-owner --file /tmp/awesome-file.sql -T foo my-app-db
```

The underlying issue lies in the fact that `pg_dump -T` excludes tables only from the `public` schema and not other schemas unless explicitly set:

```bash
pg_dump --schema-only --no-privileges --no-owner --file /tmp/awesome-file.sql -T myschema.foo my-app-db
```

# Option 1: Always return the qualified schema name on `data_sources`

```ruby
# Before
ActiveRecord::Base.connection.data_sources
#=> ["foo", "bar", "prefix_foo", "ignored_foo"]

# After
ActiveRecord::Base.connection.data_sources
#=> ["myschema.foo", "public.bar", "public.prefix_foo", "public.ignored_foo"]
```

The downside here is that the suggested approach for prefixes would not work anymore:

```ruby
ActiveRecord::SchemaDumper.ignore_tables = [/^prefix_/]
```

It would need to change to this instead:

```ruby
ActiveRecord::SchemaDumper.ignore_tables = [/^public\.prefix_/]
```

`connection.data_sources` method is not widely used on Rails source code and should be fine to patch the places using it if something breaks.

The biggest issue is that it would break existing applications using that recommended approach.

# Option 2: Enhance `SchemaDumper` to assume `public` schema when not specified otherwise

PoC:

```ruby
        ignore_tables = ActiveRecord::SchemaDumper.ignore_tables
        if ignore_tables.any?
          ignore_tables = connection.data_sources(include_schema: true).select do |table|
            table_schema, table_name = ActiveRecord::ConnectionAdapters::PostgreSQL::Utils.extract_schema_qualified_name(table)

            ignore_tables.any? do |pattern|              
              case pattern
              when String
                pattern_schema, pattern_table = ActiveRecord::ConnectionAdapters::PostgreSQL::Utils.extract_schema_qualified_name(pattern)
                pattern_schema ||= "public" # assumes public schema unless specified

                pattern_schema == table_schema && pattern_table == table_name
              else
                pattern === table || pattern === table_name # tries the qualified name and falls back to the raw name
              end
            end
          end
          args += ignore_tables.flat_map { |table| ["-T", table] }
        end
```

The problem here is that it might be too loose and falsely ignore tables like this:

```ruby
ActiveRecord::SchemaDumper.ignore_tables = [/^public./] # user actually meant `^public\.`
```

This would filter tables named like `public.foo` (the overall schema) but also tables named `public.public_repos` (due to the fallback).

_We could use `pattern === table` and mention in the docs that you must use the schema for Regexps, but this starts to feel dirty since String would accept the implicit name.`

# Option 3: Implement a feature toggle

We could avoid the machinery and use opt-in to either always use or not like this:

```ruby
ActiveRecord::SchemaDumper.use_schema_qualified_name = true

# And now you must always specify the schema name
ActiveRecord::SchemaDumper.ignore_tables = [
  /^public\.prefix_/,
  "myschema.foo"
]
```

# Option 4: Allows to specify the schema to filter

Didn't think too much here but seems the cleaner public API (still would require documentation guidance to clarify that):

```ruby
ActiveRecord::SchemaDumper.ignore_tables = {
  public: [/^prefix_/],
  myschema: ["foo", /^another_prefix_/]
]
```

Using an Array would keep the current behavior and using the Hash value would filter by schema. If there's a shared pattern, it could be reused:

```ruby
something = [/^prefix_/]
ActiveRecord::SchemaDumper.ignore_tables = {
  public: [something],
  myschema: ["foo", something]
]
```

# Current workaround

We can use `ActiveRecord::Tasks::DatabaseTasks.structure_dump_flags` but then we say that `SchemaDumper.ignore_tables` doesn't support schemas other than `public` and we should mention in the docs:

```ruby
ActiveRecord::Tasks::DatabaseTasks.structure_dump_flags = {
  postgres: ['-T', 'myschema.foo']
}
```

rafaelfranca:
I like option 4 better. Doesn't change the implementation much for people that don't use multiple schemas

sobrinho:
@rafaelfranca thanks! I will implement it then and let you know once it's ready for review.

sobrinho:
@rafaelfranca would you review again?

PR #55466, opened 2025-08-08
Fix response_finished Executor test for Rack HEAD

skipkayhil:
Ref https://github.com/rack/rack/pull/2365

We're interested in adding more validation to response_finished callbacks, in this case making sure the environment is a valid rack environment.

This commit addresses that change by passing the real env given to the middleware to the callback as well (since its already a valid env and the server will end up doing this kind of thing).


ioquatix:
LGTM.

PR #55469, opened 2025-08-08
Refine code generated by `Module#delegate`

byroot:
Extremely minor, but we can access the receiver from the `NoMethodError`, so we don't need to store it in a local variable.


byroot:
Unfortunately this doesn't work for manually raised `NoMethodError`:

```ruby
>> NameError.new.receiver
(irb):1:in 'NameError#receiver': no receiver is available (ArgumentError)
```

Too bad.

PR #53208, opened 2024-10-06
Unify migration file name formats in "Getting Started with Rails" [ci skip]

wakairo:
### Motivation / Background

This Pull Request has been created because there are two different migration file name formats as below.

- db/migrate/\<timestamp\>_create_articles.rb
- db/migrate/20140120201010_create_comments.rb with a comment: "your name will include a different timestamp"

### Detail

This Pull Request unifies the format as below.

- db/migrate/\<timestamp\>_create_articles.rb
- db/migrate/\<timestamp\>_create_comments.rb

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55472, opened 2025-08-09
[RF-DOCS] Rails Engines

Ridhwana:
### Motivation / Background

This Pull Request has been created to improve the Rails Engines Guides

### Detail

- [x] Add screenshots when talking about the views to aid users to see the UI impact
- [x] Show the structure of the engine 
- [x] A bit more info about the engives vs. plugins
- [x] Update a lot of sections to have further sub sections
- [x]  Update the `:path` option for the gem 
- [x] Fix the example for the overridden view in the application to reflect author.name
- [x] Move Testing an Engine into Improving the Engine
- [x] Update some sections to be more clear
- [x] Update the `Blorgh::Article.find(1)` example to first create an article instead? `Blorgh::Article.create(title: "Hello World")`
- [x] remove references to`attr_accessor`, `cattr_accessor` and keep and link `mattr_accessor` because the former two don't seem necessary.
- [x] Update the title to be consistent "Engines Overview"?
- [x] Maybe the introduction of generating the plugin with `--mountable` should be moved to the "Plugins" guide? We've added this there but also mentioned it here
- [ ] Maybe the `app` directory section could use some code examples instead of describing the namespaced classes inline? I felt like we've covered examples throughout the guide in detail adn I'd even added more so I kept this to be about the structure but if reviewers feel much differently, happy to amend.
- [x] The timestamp placeholder `[timestamp_1]` should maybe be made more consistent with other placeholders in the guide? `<timestamp_1>` See: https://github.com/rails/rails/pull/53208
- [x] The `bin/rails generate migration add_author_id_to_blorgh_articles author_id:integer` command uses snake_case for the migration name while we mostly uses CamelCase in other guides. This should probably be documented in the Guidelines as well.

### Checklist


guides/source/engines.md
```diff
@@ -1,10 +1,10 @@
 **DO NOT READ THIS FILE ON GITHUB, GUIDES ARE PUBLISHED ON <https://guides.rubyonrails.org>.**
 
-Getting Started with Engines
-============================
+Rails Engines Overview
```
p8, line 3: This change also requires the name of the guide to be changed in `documents.yml`

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
-[Refinery CMS](https://github.com/refinery/refinerycms), a CMS engine.
-
-Finally, engines would not have been possible without the work of James Adam,
-Piotr Sarnacki, the Rails Core Team, and a number of other people. If you ever
-meet them, don't forget to say thanks!
+Engines can be thought of as miniature applications that encapsulate specific
+functionality and integrate with a larger Rails application. A Rails application
+is essentially a "supercharged" engine, with the `Rails::Application` class
+inheriting a lot of its behavior from `Rails::Engine`, which in turn [inherits
+from `Rails::Railtie`](engines.html#the-inheritance-hierarchy).
```
p8: Mentioning that a `Rails::Application` is an engine is some nice trivia, but not essential to most readers. You can easily create an engine without this knowledge.

Maybe move it out of the opening paragraph and move it under the other engine examples, as a special example of an engine?
And only explain the Railtie inheritance here?

```suggestion
Engines can be thought of as miniature applications that encapsulate specific
functionality and integrate with a larger Rails application. An engine
extends plugin, with the `Rails::Engine` class
[inheriting](engines.html#the-inheritance-hierarchy) behavior from `Rails::Railtie`.
```

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+
+Some examples of engines in action:
+
+* [Devise](https://github.com/plataformatec/devise) which provides authentication for its parent applications
+* [Thredded](https://github.com/thredded/thredded) which provides forum functionality
+* [Spree](https://github.com/spree/spree) which provides an e-commerce platform
+* [Refinery CMS](https://github.com/refinery/refinerycms) which provides a CMS engine
+* [Active Storage](https://github.com/rails/rails/tree/main/activestorage) which provides a file storage as an engine.
```
p8:
```suggestion
* [Active Storage](https://github.com/rails/rails/tree/main/activestorage) which provides file storage as an engine.
```

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+functionality and integrate with a larger Rails application. A Rails application
+is essentially a "supercharged" engine, with the `Rails::Application` class
+inheriting a lot of its behavior from `Rails::Engine`, which in turn [inherits
+from `Rails::Railtie`](engines.html#the-inheritance-hierarchy).
+
+Some examples of engines in action:
+
+* [Devise](https://github.com/plataformatec/devise) which provides authentication for its parent applications
+* [Thredded](https://github.com/thredded/thredded) which provides forum functionality
+* [Spree](https://github.com/spree/spree) which provides an e-commerce platform
```
p8: Nitpick. While Spree can be considered an engine, I'm not sure it's an example of a "miniature application" that  "integrates in a larger application".
Looking at the code it's also difficult to find examples of the code in this guide.
On the other hand, it is a nice example of how far you can take it.

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+
+Engines are also closely related to plugins - all engines are plugins, but not
+all plugins are engines. The two share a common `lib` directory structure, and
+are both generated using the `rails plugin new` generator.
+
+While a plugin is generated using `rails plugin new <plugin_name>`, an engine is
+generated using either:
+
+- `rails plugin new <engine_name> --full`, or
+- `rails plugin new <engine_name> --mountable`.
```
p8: Maybe move this to a code block?
````suggestion
```bash
rails plugin new <engine_name> --full
# or
rails plugin new <engine_name> --mountable
```
````

File: guides/source/engines.md [hunk 6]
```diff
@@ -104,13 +186,16 @@ that provides the following:
     standard Rails application's `config/application.rb` file:
 
     ```ruby
+    # lib/blorgh/engine.rb
+
     module Blorgh
       class Engine < ::Rails::Engine
       end
     end
```
p8, line 235: For a mountable engine this contains the namespace isolation: `isolate_namespace Blorgh`.
We can probably remove the last item in list (that duplicates this file) as well.

guides/source/engines.md
```diff
@@ -104,13 +186,16 @@ that provides the following:
...
     ```
 
-The `--mountable` option will add to the `--full` option:
+`-- mountable` engines, like above, contain some files that are not present in
+the `--full` option, these are:
 
   * Asset manifest files (`blorgh_manifest.js` and `application.css`)
```
p8: I don't think these are generated anymore?

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+├── config/
+│   ├── initializers/
+│   └── routes.rb
+├── lib/
+│   ├── blorgh/
+│   │   └── engine.rb
+│   ├── blorgh.rb
+│   └── tasks/
+│       └── blorgh_tasks.rake
+├── MIT-LICENSE
```
p8: This no longer seems to be present either:
```suggestion
```

I'm also seeing a rubocop.yml and some github actions files.
These are probably added as part of the application generator.
I'm not sure if we should list those here as well, as these might change and are burdensome to maintain here?

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+The structure of the `--mountable` engine will be as follows:
+
+```
+blorgh/
+├── app/
+│   ├── assets/
+│   │   ├── javascripts/
+│   │   │   ├── blorgh/
+│   │   │   │   └── application.js
+│   │   │   └── blorgh_manifest.js
```
p8: I don't think any javascript is generated anymore.
Locally I did get an image:
```suggestion
│   │   ├── images/
│   │   │   ├── blorgh/
│   │   │   │   └── .keep
```

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
-meet them, don't forget to say thanks!
+Engines can be thought of as miniature applications that encapsulate specific
+functionality and integrate with a larger Rails application. A Rails application
+is essentially a "supercharged" engine, with the `Rails::Application` class
+inheriting a lot of its behavior from `Rails::Engine`, which in turn [inherits
+from `Rails::Railtie`](engines.html#the-inheritance-hierarchy).
+
+Some examples of engines in action:
+
+* [Devise](https://github.com/plataformatec/devise) which provides authentication for its parent applications
```
rafaelfranca:
```suggestion
* [Devise](https://github.com/heartcombo/devise) which provides authentication for its parent applications
```

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+initialization process and allows extensions, such as frameworks (e.g., Active
+Record, Action Mailer) or third-party libraries, to tie into the Rails boot
+sequence.
+
+The [`Rails Engine`](https://api.rubyonrails.org/classes/Rails/Engine.html) builds on
+`Railtie` by adding support for things like routes, isolated namespaces, and
+load paths, making it possible to package complete Rails components. Engines and
+applications also share a common directory structure.
+
+While an engine can be part of a larger Rails application, it does not require
```
rafaelfranca: I don't think this is true. An Engine already need to be part of an application to function, otherwise there is no way to boot it. Where can an engine function without an application?

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+an [`Application`](https://api.rubyonrails.org/classes/Rails/Application.html)
+class to function. However, `Application` can extend `Engine` with additional
+responsibilities like middleware setup, configuration loading, and application
```
rafaelfranca: This is also not true. Engines can setup they own middleware and they can load their own configuration.

guides/source/engines.md
```diff
@@ -21,76 +21,158 @@ After reading this guide, you will know:
...
+needs its own models/controllers but shares the host app's namespace, while the
+`--mountable` option tells the generator that you want to create a fully
+isolated, mountable engine with its own namespace.
+
+You can read more about the different generator options in the [Rails Plugins
+Generator Options](plugins.html#generator-options) section.
+
+### Using an Engine
+
+You add an engine to your host applications `Gemfile`, and then mount it in the
```
rafaelfranca: Not every engine need to be mounted. We need to explain this here.

guides/source/engines.md
```diff
@@ -104,13 +186,16 @@ that provides the following:
...
+    # lib/blorgh/engine.rb
+
     module Blorgh
       class Engine < ::Rails::Engine
       end
     end
     ```
 
-The `--mountable` option will add to the `--full` option:
+`-- mountable` engines, like above, contain some files that are not present in
```
rafaelfranca:
```suggestion
`--mountable` engines, like above, contain some files that are not present in
```

guides/source/engines.md
```diff
@@ -183,79 +283,88 @@ end
...
-for the engine. If this engine is to be used as a stand-alone engine, then you
-would add any customization to its layout in this file, rather than the
-application's `app/views/layouts/application.html.erb` file.
-
-If you don't want to force a layout on to users of the engine, then you can
-delete this file and reference a different layout in the controllers of your
-engine.
+* Add the engine’s `app/` folder to the load path.
+* Load the engine’s models, mailers, controllers, and views.
+* Mount the engine at the specified path when used in a host app.
```
rafaelfranca: This isn't true. You need to mount in the routes files.

guides/source/engines.md
```diff
@@ -142,23 +227,35 @@ following to the dummy application's routes file at
...
+```
+
+This tells Bundler to treat the engine as a gem and load it accordingly.
+
+#### The Engine Entry Point: `lib/blorgh.rb`:
+
+Bundler looks for a file matching the gem name, `lib/blorgh.rb`. This file is the main entry point of the engine. It typically requires the engine definition and sets up a base module:
 
 ```ruby
 require "blorgh/engine"
```
p8, line 304:
```suggestion
# lib/blorgh.rb
require "blorgh/engine"
```

guides/source/engines.md
```diff
@@ -183,79 +283,88 @@ end
...
+The `isolate_namespace Blorgh` from the Engine class is crucial. It prevents your engine’s components (like models, routes, helpers, and controllers) from clashing with those in the host application or other engines.
+
+For example:
+
+* A generated model becomes `Blorgh::Article` rather than `Article`.
+* The table name is `blorgh_articles`, not `articles`.
+* A controller becomes `Blorgh::ArticlesController`, with views in
+  `app/views/blorgh/articles`.
+* A helper becomes `Blorgh::ArticlesHelper`, with views in
+  `app/helpers/blorgh/articles_helper.rb`.
```
p8:
```suggestion
* A helper becomes `Blorgh::ArticlesHelper`.
```

guides/source/engines.md
```diff
@@ -183,79 +283,88 @@ end
...
+
+#### `app/assets`
+
+Inside `app/assets`, you’ll find directories for `images`, `javascripts`, and `stylesheets`. Each of these contains a subdirectory named after your engine—`blorgh` in this case.
+
+This namespacing is important: it ensures that your engine’s assets don’t conflict with those from other engines or the host application. For example:
+
+```
+app/assets/stylesheets/blorgh/application.css
+app/assets/javascripts/blorgh/application.js
```
p8: I don't think this is created anymore:
```suggestion
```

guides/source/engines.md
```diff
@@ -183,79 +283,88 @@ end
...
+
+### Understanding the `app` Directory
+
+The `app` directory in a mountable engine mirrors the familiar structure of a standard Rails application. It includes subdirectories like `assets`, `controllers`, `helpers`, `jobs`, `mailers`, `models`, and `views`.
+
+Here’s what you should know about each of these, especially in the context of a namespaced engine like `blorgh`:
+
+#### `app/assets`
+
+Inside `app/assets`, you’ll find directories for `images`, `javascripts`, and `stylesheets`. Each of these contains a subdirectory named after your engine—`blorgh` in this case.
```
p8:
```suggestion
Inside `app/assets`, you’ll find directories for `images` and `stylesheets`. Each of these contains a subdirectory named after your engine—`blorgh` in this case.
```

guides/source/engines.md
```diff
@@ -603,149 +762,208 @@ The `comment_counter` local variable is given to us by the `<%= render
...
-As described earlier, by placing the gem in the `Gemfile` it will be loaded when
-Rails is loaded. It will first require `lib/blorgh.rb` from the engine, then
-`lib/blorgh/engine.rb`, which is the file that defines the major pieces of
-functionality for the engine.
+NOTE: By including the engine in the application's `Gemfile`, it will be loaded
+automatically when Rails boots. Rails will first require the engine’s entry
+point file at `lib/blorgh.rb`. This file typically sets up the namespace and
+requires additional files, including `lib/blorgh/engine.rb`, which defines the
+`Blorgh::Engine` class. This `Engine` class is responsible for hooking into the
+Rails application and mounting the engine's routes, initializers, and other
```
rafaelfranca:
```suggestion
Rails application and mounting the engine's initializers, and other
```

guides/source/engines.md
```diff
@@ -603,149 +762,208 @@ The `comment_counter` local variable is given to us by the `<%= render
...
 ```bash
 $ bin/rails railties:install:migrations DATABASE=animals
 ```
 
-This command, when run for the first time, will copy over all the migrations
-from the engine. When run the next time, it will only copy over migrations that
-haven't been copied over already. The first run for this command will output
-something such as this:
+NOTE: These tasks are provided by
+[Railtie](https://api.rubyonrails.org/classes/Rails/Railtie.html), the Rails
```
rafaelfranca: Those tasks are provided by the `ActiveRecord::Railtie` not the generic `Railtie`

guides/source/engines.md
```diff
@@ -780,203 +1010,249 @@ def article_params
...
+There also needs to be a way of associating the records in the `blorgh_articles`
+table with the records in the `users` table. Because the association in the
+engine is called `author`, there should be an `author_id` column added to the
+`blorgh_articles` table.
+
+To generate this new column, run this command within the engine's root directory:
 
 ```bash
-$ bin/rails generate migration add_author_id_to_blorgh_articles author_id:integer
+$ bin/rails generate migration AddAuthorIdToBlorghArticles author_id:integer
```
rafaelfranca: Why this change? The previous command was fine. The migration generator documentation prefer that one over the one you wrote.

guides/source/engines.md
Ridhwana: While an engine is packaged like a miniature Rails application, it runs inside a host Rails application. The host `Rails::Application` coordinates boot, executes engine initializers, and builds the overall middleware stack. Engines can also define their own configuration and contribute middleware, but these take effect when the host application boots.

@rafaelfranca Would you say that this is more accurate?

guides/source/engines.md
Ridhwana: CleanShot 2025-08-13 at 20 03 20@2x

Yup you're correct about tthe structure not being up to date, I had generated the application and tested all the commands, but not sure how I got this wrong. 

Edit: However, I'm still seeing MIT-LICENSE.

guides/source/engines.md
rafaelfranca: yes. Perfect

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: This was a mistake, when I was testing out the engine locally I copied the migration name from the file ✅

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
```diff
@@ -21,76 +21,178 @@ After reading this guide, you will know:
...
+generated using either:
+
+```bash
+rails plugin new <engine_name> --full
+# or
+rails plugin new <engine_name> --mountable
+```
+
+The `--full` option tells the generator that you want to create an engine that
+needs its own models/controllers but shares the host app's namespace, while the
```
gmcgibbon:
```suggestion
needs its own models/controllers that share the host app's namespace. The
```
This sentence is a bit long. Let's shorten it.

guides/source/engines.md
```diff
@@ -21,76 +21,178 @@ After reading this guide, you will know:
...
+`--mountable` option tells the generator that you want to create a fully
+isolated, mountable engine with its own namespace.
```
gmcgibbon:
```suggestion
isolated, mountable engine with its own namespace module.
```
Ruby doesn't have namespaces per se, so let's be explicit that it is just a toplevel module.

guides/source/engines.md
```diff
@@ -21,76 +21,178 @@ After reading this guide, you will know:
...
+
+You can read more about the different generator options in the [Rails Plugins
+Generator Options](plugins.html#generator-options) section.
+
+### Using an Engine
+
+You add an engine to your host applications `Gemfile`, and then mount it in the
```
gmcgibbon:
```suggestion
You add an engine to your host application's `Gemfile`, and then mount it in the
```

guides/source/engines.md
gmcgibbon: License generation is dependent on where you generate the app: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/plugin/plugin_generator.rb#L58. You won't get one if you generate it inside of an app. We don't explicitly call out in-app engines as being unique in the guides, but they have specific features.

guides/source/engines.md
```diff
@@ -142,25 +239,38 @@ following to the dummy application's routes file at
...
+
+### Core Engine Setup
 
-#### Critical Files
+#### The `.gemspec` File
 
-At the root of this brand new engine's directory lives a `blorgh.gemspec` file.
-When you include the engine into an application later on, you will do so with
-this line in the Rails application's `Gemfile`:
+At the root of your engine, you’ll find a file named `blorgh.gemspec`. This file defines your engine as a gem. It includes metadata like the gem name, version, authors, dependencies, and which files to include when it's packaged.
```
gmcgibbon:
```suggestion
At the root of our engine, you’ll find a file named `blorgh.gemspec`. This file defines the engine as a gem. It includes metadata like the gem name, version, authors, dependencies, and which files to include when it's packaged.
```
We describe the engine as "ours" above. Though there seems to be a lot of "your" engine references below this one too. I think the tone here could be more consistent.

guides/source/engines.md
```diff
@@ -21,76 +21,178 @@ After reading this guide, you will know:
...
+Since a mounted engine has an isolated namespace, the host application and the
+mounted engine can have a routing helper with the same name (such as
+`articles_path`) without clashing. Along with this, controllers, models and
+table names are also namespaced. You'll see how to do this later in the [Routes
+section](#routes).
 
 Generating an Engine
 --------------------
 
+In the following example, we'll be building an engine, called "blorgh" that
```
gmcgibbon:
```suggestion
In the following example, we'll be building an engine, called "blog" that
```
Since we're rewriting this, and the gem is only ever locally path-sourced to an app, can we simply call it "blog"? There's no intention to publish it to rubygems and it probably reads a little easier.

guides/source/engines.md
```diff
@@ -183,79 +296,86 @@ end
...
+* A generated model becomes `Blorgh::Article` rather than `Article`.
+* The table name is `blorgh_articles`, not `articles`.
+* A controller becomes `Blorgh::ArticlesController`, with views in
+  `app/views/blorgh/articles`.
+* A helper becomes `Blorgh::ArticlesHelper`.
+* Mailers and Jobs are namespaced as well.
+* Engine routes are kept isolated and don’t mix with the main app’s routes. This
+  is discussed later in the [Routes](#routes) section of this guide.
+
+Without this isolation, files from the engine might "leak" into the host app’s namespace or identically named classes might override each other.
```
gmcgibbon: This is repeating what the note is saying. Do we need to say this twice?
```suggestion

```

guides/source/engines.md
```diff
@@ -183,79 +296,86 @@ end
...
+
+```ruby
+# app/controllers/blorgh/application_controller.rb
+module Blorgh
+  class ApplicationController < ActionController::Base
+  end
+end
+```
+
+This controller acts as the base for all controllers in the engine, much like `ApplicationController` does in a full app. Placing it (and all other controllers) in the `blorgh/` namespace ensures they won’t clash with similarly named controllers in the host app or other engines.
```
gmcgibbon:
```suggestion
This controller acts as the base for all controllers in the engine, much like `ApplicationController` does in a full app. Placing it (and all other controllers) in the `blorgh/` module ensures they won’t clash with similarly named controllers in the host app or other engines.
```

guides/source/engines.md
```diff
@@ -183,79 +296,86 @@ end
...
+
+Inside `app/views/layouts`, you’ll find a layout file for the engine:
+
+```
+app/views/layouts/blorgh/application.html.erb
+```
+
+This is the default layout used by views inside the engine. It’s useful if your engine is meant to be used as a self-contained application (e.g., admin dashboards, wikis, etc.). If this engine is to be used as a stand-alone engine, then you would add any customization to its layout in this file, rather than the application's `app/views/layouts/application.html.erb` file.
+
+If you don't want to force a layout on to users of the engine, then you can delete this file and reference a different layout in the controllers of your engine.
```
gmcgibbon:
```suggestion
If you don't want to use a specific layout in the views of the engine, then you can delete this file and reference a different layout in the controllers of your engine.
```

guides/source/engines.md
```diff
@@ -269,22 +389,20 @@ $ bin/rails generate model
...
 
-#### `test` Directory
+#### `/test`
 
 The `test` directory is where tests for the engine will go. To test the engine,
 there is a cut-down version of a Rails application embedded within it at
-`test/dummy`. This application will mount the engine in the
-`test/dummy/config/routes.rb` file:
+`test/dummy`. This application will mount the engine at `/blorgh`, which will
+make it accessible through the application only at that path.
```
gmcgibbon:
```suggestion
make it accessible through the application's routes at that path.
```

guides/source/engines.md
```diff
@@ -399,49 +526,58 @@ module Blorgh
...
-of our engine to run the migration generated by the scaffold generator, and then
-running `bin/rails server` in `test/dummy`. When you open
-`http://localhost:3000/blorgh/articles` you will see the default scaffold that has
-been generated. Click around! You've just generated your first engine's first
-functions.
+#### Exploring the Engine in the Browser
 
-If you'd rather play around in the console, `bin/rails console` will also work just
-like a Rails application. Remember: the `Article` model is namespaced, so to
+You can explore the engine by running `bin/rails db:migrate` at the root of the
```
gmcgibbon:
```suggestion
You can explore the engine by first running `bin/rails db:migrate` at the root of the
```
I originally thought we meant to say console here, but I think we're saying we need to migrate and then run the server.

guides/source/engines.md
```diff
@@ -521,21 +695,22 @@ directory at `app/views/blorgh/comments` and in it a new file called
...
-When this form is submitted, it is going to attempt to perform a `POST` request
-to a route of `/articles/:article_id/comments` within the engine. This route doesn't
-exist at the moment, but can be created by changing the `resources :articles` line
-inside `config/routes.rb` into these lines:
+When this form is submitted, it will attempt to perform a `POST` request to the
+`/articles/:article_id/comments` route within the engine. You can read more
+about the [`form_with` helper](form_helpers.html#working-with-basic-forms) in
+the guides.
+
+However, this route doesn't exist as yet. You can create it by nesting the
```
gmcgibbon:
```suggestion
However, this route doesn't exist yet. You can create it by nesting the
```

guides/source/engines.md
```diff
@@ -603,149 +773,208 @@ The `comment_counter` local variable is given to us by the `<%= render
...
-First, the engine needs to be specified inside the application's `Gemfile`. If
-there isn't an application handy to test this out in, generate one using the
-`rails new` command outside of the engine directory like this:
+To use the engine in a host application, add it to the application's `Gemfile`. If
+you don't already have an application to test with, you can generate one outside
+the engine directory using the `rails new` command:
 
 ```bash
-$ rails new unicorn
+$ cd .. # Go back one folder
```
gmcgibbon:
```suggestion
$ cd .. # Exit the engine folder
```

guides/source/engines.md
```diff
@@ -603,149 +773,208 @@ The `comment_counter` local variable is given to us by the `<%= render
...
 gem "devise"
 ```
 
-However, because you are developing the `blorgh` engine on your local machine,
-you will need to specify the `:path` option in your `Gemfile`:
+However, since the engine is not published as a gem yet, and you're developing it locally, you need to link to it using a relative path in the host application's Gemfile:
 
 ```ruby
-gem "blorgh", path: "engines/blorgh"
+gem "blorgh", path: "../blorgh"
```
gmcgibbon, line 870: It might be worth mentioning here that you can generate engines inside applications and deploy them that way instead of publishing a singular engine gem for one application. Though, it might make sense to rewrite this entire section instead with this in mind.

guides/source/engines.md
```diff
@@ -780,203 +1021,249 @@ def article_params
...
-
 ```
-NOTE Migration [timestamp]_create_blorgh_articles.blorgh.rb from blorgh has been skipped. Migration with the same name already exists.
-NOTE Migration [timestamp]_create_blorgh_comments.blorgh.rb from blorgh has been skipped. Migration with the same name already exists.
-Copied migration [timestamp]_add_author_id_to_blorgh_articles.blorgh.rb from blorgh
+Copied migration <timestamp>_add_author_id_to_blorgh_articles.blorgh.rb from blorgh
 ```
 
+Notice that only _the latest_ migration was copied over. This is because the first
+two migrations were copied over the first time this command was run.
```
gmcgibbon:
```suggestion
two migrations were copied over the first time this command was run in the previous section.
```

guides/source/engines.md
```diff
@@ -780,203 +1021,249 @@ def article_params
...
-engine:
+To define this configuration setting, you should use a
+[`mattr_accessor`](https://api.rubyonrails.org/classes/Module.html#method-i-mattr_accessor)
+inside the `Blorgh` module for the engine. Add this line to the `Blorgh` module:
 
 ```ruby
-mattr_accessor :author_class
+# lib/blorgh.rb
+module Blorgh
+  mattr_accessor :author_class
```
gmcgibbon:
```suggestion
  mattr_accessor :author_class_name
```
This should always be a string to avoid loading problems, so let's call it author_class_name. It also makes the method to constantize below easier to understand, and the following part where we add to_s unnecessary.

guides/source/engines.md
```diff
@@ -1090,43 +1322,64 @@ module MyApp
...
-  # ...
+  validates :title, presence: true, length: { minimum: 10 }
 end
 ```
 
-It is very important that the override _reopens_ the class or module. Using the `class` or `module` keywords would define them if they were not already in memory, which would be incorrect because the definition lives in the engine. Using `class_eval` as shown above ensures you are reopening.
+It is very important that the override _reopens_ the class or module. Using the
+`class` or `module` keywords would define them if they were not already in
+memory, which would be incorrect because the definition lives in the engine.
+Using [`class_eval`](https://api.rubyonrails.org/classes/Module.html#method-i-class_eval) as shown above ensures you are reopening.
```
gmcgibbon:
```suggestion
Using [`class_eval`](https://api.rubyonrails.org/classes/Module.html#method-i-class_eval) as shown above ensures you are reopening an existing module.
```

guides/source/engines.md
```diff
@@ -1090,43 +1322,64 @@ module MyApp
...
-```ruby
-# MyApp/app/models/blorgh/article.rb
+- Adding a new instance method: `time_since_created`
+- Overriding the existing `summary` method
+
+Here’s how you can do it cleanly using `ActiveSupport::Concern`:
 
+##### Adding/Overriding methods
+
+Add the `Article#time_since_created` and override the `Article#summary` in the
```
gmcgibbon:
```suggestion
Add the `Blorgh::Article#time_since_created` and override the `Blorgh::Article#summary` in the
```

guides/source/engines.md
```diff
@@ -1211,135 +1471,287 @@ and put this content in it:
...
+default. This isolation is established by the
+[`isolate_namespace`](#the-engine-class-definition-lib-blorgh-engine-rb) call in
+the engine’s `Engine` class. It allows both the engine and the application to
+define routes with the same names—like `articles_path`—without conflict.
 
-Routes inside an engine are drawn on the `Engine` class within
-`config/routes.rb`, like this:
+#### Defining Engine Routes
+
+You define routes inside the engine using the engine's route set.
```
gmcgibbon:
```suggestion
You can define routes inside the engine using the engine's route set:
```

guides/source/engines.md
```diff
@@ -1211,135 +1471,287 @@ and put this content in it:
...
+only used within the engine and don't need to be included in the host
+application's asset pipeline.
 
-You can define assets for precompilation in `engine.rb`:
+In such situations, it doesn’t make sense for the host app to reference these
+assets manually (e.g., via `stylesheet_link_tag "blorgh/admin"`). Instead, you
+should explicitly tell Rails to precompile these assets so they’re available
+when the engine is used.
+
+To do this, you can add a precompilation hook in your engine’s engine.rb file:
```
gmcgibbon:
```suggestion
To do this, you can add a precompilation hook in your engine's engine.rb file:
```

guides/source/engines.md
```diff
@@ -1211,135 +1471,287 @@ and put this content in it:
...
 end
 ```
 
+This ensures that these engine-specific assets will be compiled when running:
+
+```bash
+bin/rails assets:precompile
+```
+
+NOTE: Be sure to include the full namespaced paths (e.g., `blorgh/admin.css`)
```
gmcgibbon:
```suggestion
NOTE: Be sure to include the full namespaced paths (e.g. `blorgh/admin.css`)
```

guides/source/engines.md
```diff
@@ -1211,135 +1471,287 @@ and put this content in it:
...
+so Sprockets can locate the correct files within your engine.
+
 For more information, read the [Asset Pipeline guide](asset_pipeline.html).
 
-### Other Gem Dependencies
+### Writing Tests
+
+When a Rails engine is generated, it includes a minimal host application inside
+the `test/dummy` directory. This "dummy app" is used solely for development and
+testing—it simulates how your engine will behave when mounted in a real
```
gmcgibbon:
```suggestion
testing—it simulates how your engine will behave when used in a real
```
This is generated for both full and mounted engines, so let's use a general term. It may not have routes.

guides/source/engines.md
```diff
@@ -1211,135 +1471,287 @@ and put this content in it:
...
+    test "title is required" do
+      article = Article.new(text: "Some content")
+      assert_not article.valid?
+      assert_includes article.errors[:title], "can't be blank"
+    end
+  end
+end
+```
+
+Organize your test files under folders like `test/models` and `test/controllers` as needed.
```
gmcgibbon:
```suggestion
Tests are organized in files under folders like `test/models` and `test/controllers` similar to an application.
```

guides/source/engines.md
Ridhwana: Updated as per below ✅

_While an engine is packaged like a miniature Rails application, it runs inside a host Rails application. The host
[Rails::Application](https://api.rubyonrails.org/classes/Rails/Application.html) coordinates boot, executes engine initializers, and builds the overall middleware stack. Engines can also define their own configuration and contribute middleware, but these take effect when the host application boots._

guides/source/engines.md
Ridhwana: Updated this, but there isn't a good link to ActiveRecord::Railtie, so I kept the link to (https://api.rubyonrails.org/classes/Rails/Railtie.html).


_NOTE: These tasks are provided by [`ActiveRecord::Railtie`](https://api.rubyonrails.org/classes/Rails/Railtie.html), the Rails component responsible for managing how Active Record integrates into an application. When run, it looks through all loaded engines and copies any exposed migrations into the host application's `db/migrate` folder, saving you from having to run a separate `install:migrations` task for each engine individually._

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: Updated to go into more detail:

_You add an engine to your host application's `Gemfile`, and depending on how the
engine is designed, you may need to mount it in the main app's routes. Mounting
is required when the engine provides its own routes, as this makes any routes,
controllers, views, or assets defined in the engine available at that mount
point in the host application. We'll cover this later in the section [Using an
Engine in the Host Application](#using-the-engine-in-a-host-application)._

_Some engines, however, don't need to be mounted. These are often backend-only
engines, such as those that provide Active Record models, rake tasks, or other
internal functionality without exposing routes. In those cases, adding the gem
to your application's Gemfile is enough to make its features available._

_When an engine is mounted, it has an isolated namespace. This means the host
application and the mounted engine can have a routing helper with the same name
(such as `articles_path`) without clashing. Along with this, controllers, models,
and table names are also namespaced. You'll see how to do this later in the
[Routes section](#routes)._

guides/source/engines.md
Ridhwana: I've removed this, and I've not added more detail here because we go into more detail about explaining it in Mounting the Engine

guides/source/engines.md
Ridhwana: ✅

Moved it out and added it as the final point to the example list:
_* [`Rails::Application`](https://api.rubyonrails.org/classes/Rails/Application.html) inheriting much of its behavior from `Rails::Engine`._

guides/source/engines.md
Ridhwana: ✅ I've shortened the sentence, but kept the "but" instead of "that" because it signals a contrast: the engine still has its own models/controllers yet shares the namespace — which is the main difference from --mountable.

guides/source/engines.md
Ridhwana: I see your point that Ruby doesn’t have namespaces in the strict sense. However, we’ve been using “namespace” throughout the guide, and I think the intended meaning is implied. Being more explicit here (but not elsewhere) might feel inconsistent.

We could add a short note earlier in the guide clarifying that “namespace” refers to a top-level module. I’ll look into whether that can be made to flow better, but I don’t think it’s strictly necessary here. Happy to change if others feel strongly.

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: This is a good callout, thanks. I'll check the guide for "yours" vs "ours" and make it consistent. Edit: ✅

guides/source/engines.md
Ridhwana: I think there’s value in keeping the engine name as “blorgh.” When we show it mounted in the host application, we use /blog as the path, so the UI and routes remain readable. Using “blorgh” as the internal engine name clearly distinguishes engine code (Blorgh::Article, migrations, etc.) from host application code in the examples, and helps avoid potential naming collisions. This separation makes it easier for readers to follow along and know that “blorgh” always refers to the engine, while “blog” refers to the application.

guides/source/engines.md
Ridhwana: Nicely spotted, thanks ✅

guides/source/engines.md
Ridhwana: ✅

guides/source/engines.md
Ridhwana: > I'm not sure if we should list those here as well, as these might change and are burdensome to maintain here?

I've cleaned this up a bit and decided not to include the MIT LICENSE as well. 

The explanation about MIT-LICENSE was really useful, thanks @gmcgibbon.

guides/source/engines.md
Ridhwana: ✅ I think a note is sufficient in this case

_NOTE: You can also generate an engine inside your application and deploy it for
use by that application only, instead of publishing it as a standalone gem. For
example, placing the engine under `engines/blorgh` and referencing it in your
`Gemfile` allows you to keep it within the same codebase._

guides/source/engines.md
Ridhwana: ✅ https://github.com/rails/rails/pull/55472/commits/1e3d771a7eec349092a722c678f348edbac90fb3

guides/source/engines.md
```diff
@@ -603,149 +824,212 @@ The `comment_counter` local variable is given to us by the `<%= render
...
-If you have multiple engines that need migrations copied over, use
-`railties:install:migrations` instead:
+which will output something like this:
+
+```
+Copied migration <timestamp_1>_create_blorgh_articles.blorgh.rb from blorgh
+Copied migration <timestamp_2>_create_blorgh_comments.blorgh.rb from blorgh
+```
+
+INFO: When run for the first time, it copies over all the migrations from the
```
p8: It was a bit confusing on first read what "it" was referring to:
```suggestion
INFO: When run for the first time, this command copies over all the migrations from the
```

Or maybe even let the INFO block be self-explanatory:
```suggestion
INFO: When run for the first time, `bin/rails blorgh:install:migrations` copies over all the migrations from the
```

guides/source/engines.md
```diff
@@ -603,149 +824,212 @@ The `comment_counter` local variable is given to us by the `<%= render
...
-Copied migration [timestamp_1]_create_blorgh_articles.blorgh.rb from blorgh
-Copied migration [timestamp_2]_create_blorgh_comments.blorgh.rb from blorgh
-```
+#### Running Migrations
 
-The first timestamp (`[timestamp_1]`) will be the current time, and the second
-timestamp (`[timestamp_2]`) will be the current time plus a second. The reason
-for this is so that the migrations for the engine are run after any existing
-migrations in the application.
+To run these migrations within the context of the application, run
```
p8:
```suggestion
To run these migrations within the context of the application, run:
```

guides/source/engines.md
```diff
@@ -603,149 +824,212 @@ The `comment_counter` local variable is given to us by the `<%= render
...
 
-You can specify a custom path in the source engine for the migrations by specifying MIGRATIONS_PATH.
+This will save you from having to run a separate `install:migrations` task for
+each engine individually.
+
+#### Referencing a Custom Path for the Migrations
+
+If your engine stores its migrations in the non-default location, you can
+specify a custom path in the source engine for the migrations using
+`MIGRATIONS_PATH`.
```
p8: It's probably more consistent if we use a colon here?
```suggestion
`MIGRATIONS_PATH`:
```

guides/source/engines.md
```diff
@@ -603,149 +824,212 @@ The `comment_counter` local variable is given to us by the `<%= render
...
 ```bash
-$ bin/rails railties:install:migrations MIGRATIONS_PATH=db_blourgh
+$ bin/rails railties:install:migrations MIGRATIONS_PATH=db_blorgh
 ```
 
-If you have multiple databases you can also specify the target database by specifying DATABASE.
+#### Migrations for Multiple Databases
+
+If you have multiple databases within an engine, you can specify the target
+database by specifying the `DATABASE` option.
```
p8:
```suggestion
database by specifying the `DATABASE` option:
```

guides/source/engines.md
```diff
@@ -754,15 +1038,17 @@ $ bin/rails generate model user name:string
...
+article's author.
 
-First, the `author_name` text field needs to be added to the
-`app/views/blorgh/articles/_form.html.erb` partial inside the engine. This can be
-added above the `title` field with this code:
+Add the `author_name` text field to the
+`app/views/blorgh/articles/_form.html.erb` partial inside the engine. This can
+be added above the `title` field with the following code:
 
 ```html+erb
```
p8, line 1053: Would it make sense here to show the whole template instead, and highlight the added lines?
I don't think we need to do this everywhere but it might make things more clear here.
We do something similar in the "Getting Started" guide:
https://github.com/rails/rails/blob/5a4cc7ae68d6dc0790a1a9154cb19a35a1a0910e/guides/source/getting_started.md?plain=1#L1753-L1766

guides/source/engines.md
```diff
@@ -771,212 +1057,260 @@ added above the `title` field with this code:
...
+methods for this attribute.
 
-To do all this, you'll need to add the `attr_accessor` for `author_name`, the
-association for the author and the `before_validation` call into
-`app/models/blorgh/article.rb`. The `author` association will be hard-coded to the
-`User` class for the time being.
+Start by adding the `attr_accessor` for `author_name`, the association for the
+author and the `before_validation` call into `app/models/blorgh/article.rb`.
 
 ```ruby
```
p8: Here we could also highlight the added lines (I'm not 100% sure about the line numbers):
````suggestion
```ruby#6-15
````

File: guides/source/engines.md [hunk 51]
```diff
@@ -771,212 +1057,260 @@ added above the `title` field with this code:
...
 
-NOTE: Due to the migration's name and the column specification after it, Rails
-will automatically know that you want to add a column to a specific table and
-write that into the migration for you. You don't need to tell it any more than
-this.
+#### Copying and Running the Migration in the Host Application
 
-This migration will need to be run on the application. To do that, it must first
-be copied using this command:
+As discussed in the [Copying the migrations](#copying-the-migrations) section,
```
p8:
```suggestion
As discussed in the [Copying the Migrations](#copying-the-migrations) section,
```

guides/source/engines.md
```diff
@@ -771,212 +1057,260 @@ added above the `title` field with this code:
...
-</p>
+If you open up the rails console in the host application, you can view the
+`Blorgh::Article` record that was created, and see that it is associated with
+the `User` record from the host application:
+
+```irb
+irb> article = Blorgh::Article.last
+=> #<Blorgh::Article id: 1, title: "Hello, World!", text: "This is a test article.", created_at: "2025-07-16 19:04:54.552457000 +0000", updated_at: "2025-07-16 19:04:54.552457000 +0000", author_id: 1>
+
+irb> user = User.find(article.author_id)
```
p8: As the `belongs_to` for the author is also defined on Article, would it be more clear if we call that directly?
```suggestion
irb> user = article.author
```

guides/source/engines.md
```diff
@@ -771,212 +1057,260 @@ added above the `title` field with this code:
...
 
-The `set_author` method in the `Blorgh::Article` model should also use this class:
+The line `self.author = User.find_or_create_by(name: author_name)` in the
+`Blorgh::Article` model should also use this class name instead of `User`:
 
 ```ruby
-self.author = Blorgh.author_class.constantize.find_or_create_by(name: author_name)
+# app/models/blorgh/article.rb
+
+self.author = Blorgh.author_class_name.constantize.find_or_create_by(name: author_name)
```
p8: Maybe make it clearer where this is set?
```suggestion
      def set_author
  self.author = Blorgh.author_class_name.constantize.find_or_create_by(name: author_name)
end
```

guides/source/engines.md
```diff
@@ -997,84 +1331,29 @@ application. The same thing goes if you want to use a standard initializer.
 For locales, simply place the locale files in the `config/locales` directory,
```
p8: Maybe link the locales to the I18n guide?
Also the "Configuring guide" mentioned some lines above should probably be "Configuring Rails Applications guide"? Or '"Configuring Rails Applications" guide'.

guides/source/engines.md
```diff
@@ -997,84 +1331,29 @@ application. The same thing goes if you want to use a standard initializer.
...
 
-Overrides may be organized in a dedicated directory `app/overrides`, ignored by the autoloader, and preloaded in a `to_prepare` callback:
+Here’s how you can set this up in the host application:
 
 ```ruby
 # config/application.rb
-module MyApp
+module HostApplication
     # ...
```
p8, line 1372: I couldn't add a suggestion to the correct line here, but I think it's more common to use `Rails.root.join`:
```suggestion
overrides = Rails.root.join("app/overrides")
```

guides/source/engines.md
```diff
@@ -1184,162 +1509,317 @@ end
...
 
-You can override this view in the application by simply creating a new file at
+You can override this view in the host application by creating a new file at
 `app/views/blorgh/articles/index.html.erb`. Then you can completely change what
 this view would normally output.
 
-Try this now by creating a new file at `app/views/blorgh/articles/index.html.erb`
-and put this content in it:
+Try this now by creating a new file at
+`app/views/blorgh/articles/index.html.erb` and put this content in it:
```
p8: It's already mentioned above, but maybe mention the host application here as well?
```suggestion
`app/views/blorgh/articles/index.html.erb` in the host application and put this content in it:
```
Or maybe delete this whole paragraph, as it basically repeats the paragraph above?

guides/source/engines.md
```diff
@@ -1184,162 +1509,317 @@ end
...
 ```css
 /*
  *= require blorgh/style
  */
 ```
 
-INFO. Remember that in order to use languages like Sass or CoffeeScript, you
+This allows engine styles to be bundled into application-wide stylesheets.
+
+INFO: Remember that in order to use languages like Sass or CoffeeScript, you
```
p8: Haml is probably less obscure than Coffeescript these days:
```suggestion
INFO: Remember that in order to use languages like Sass or Haml, you
```

guides/source/engines.md
```diff
@@ -1184,162 +1509,317 @@ end
...
+
+When a Rails engine is generated, it includes a minimal host application inside
+the `test/dummy` directory. This "dummy app" is used solely for development and
+testing—it simulates how your engine will behave when used in a real
+application. You can extend the dummy app by adding controllers, models, or
+views as needed to help you test the engine’s functionality in context.
+
+The `test/` directory works just like it does in a standard Rails application.
+You can write unit tests, functional tests, and integration tests to ensure your
+engine behaves as expected.
```
p8: It wasn't clear to me if this meant `blorgh/test` or `blorgh/test/dummy/test`.
Maybe swap this paragraph with the first paragraph? Or move it to the next section that mentions the dummy app?

guides/source/engines.md
```diff
@@ -1184,162 +1509,317 @@ end
...
+      article = Article.new(text: "Some content")
+      assert_not article.valid?
+      assert_includes article.errors[:title], "can't be blank"
+    end
+  end
+end
+```
+
+Tests are organized in files under folders like `test/models` and
+`test/controllers` similar to an application.
```
p8: This is pretty similar to lines 1717-1715 and seems a bit out of place here in the end.

guides/source/engines.md
```diff
@@ -1090,43 +1369,82 @@ module MyApp
...
+by:
 
-```ruby
-# MyApp/app/models/blorgh/article.rb
+- Adding a new instance method: `time_since_created`
+- Overriding the existing `summary` method
 
+Here’s how you can do it cleanly using `ActiveSupport::Concern`:
+
+##### Adding/Overriding methods
```
p8: The order of the following sections is a bit confusing to me.
Instead of referencing an unknown concern that is created later, I think it would be clearer if we start with the concern.
After creating the concern we can include it in engine's Article.
We still need the engine's article as other users of the engine might not override it.
As a last step we can show the host application's Article.

Ridhwana: This makes sense, done ✅

guides/source/engines.md
Ridhwana: TIL!

PR #49624, opened 2023-10-13
Remove all deprecated code

rafaelfranca:
Remove all code deprecated in 7.1.

PR #48533, opened 2023-06-20
Call proxy methods from `alias_attribute` generated methods

nvasilevski:
This commit changes bodies of methods generated by `alias_attribute` on Active Record objects.

Previously the body of the `alias_attribute :new_title, :title` was `def new_title; title; end`. This commit changes it to `def new_title; attribute("title"); end`.

This allows for `alias_attribute` to be used to alias attributes named with a reserved names like `id`:
```ruby
  class Topic < ActiveRecord::Base
    self.primary_key = :title
    alias_attribute :id_value, :id
  end

  Topic.new.id_value # => 1
```

### Body generation

Most of the methods for the alias attribute will now be generated by `define_proxy_call` unless the model redefines the `target_name` attribute method, for example:
```ruby
class Post < ApplicationRecord
 alias_attribute :subject, :title

 def title_was
  ...
 end
```

will issue a deprecation warning that asks for the `subject_was` to be overridden as well since next version of Rails will not make `subject_was` method to call into the `title_was`. This is a very unlikely situation and I genuinely expect no applications to ever see the deprecation warning but still something that does require a deprecation cycle. 

### Active Model changes

Even though we don't change the behavior of the Active Model alias attribute definition there are some code changes in this PR that were made to share some common code between Active Model and Active Record. If you find this unnecessary complicated feel free to speak up and we can consider an alternative which means no changes to Active Model but more duplication in Active Record. Some of the duplication will go away though when deprecation is removed

PR #55480, opened 2025-08-12
Fix alias_attribute to preserve custom getter methods

shravan097:
### Motivation / Background

I am upgrading an app from `7.0.8` to latest which is `8.0.2`. I have an existing logic that passes custom method in `alias_attribute` expecting custom method to trigger.  

Currently in Rails 8.0.2, `alias_attribute` generates methods that take precedence over custom methods, causing it to return raw database values instead of custom method results. This creates confusing behavior where developers expect their custom methods to be called but they are silently bypassed.

**Example demonstrating the issue:**

```ruby
# Create a simple table
ActiveRecord::Base.connection.create_table :users, force: true do |t|
  t.string :email
end

# Define a model with custom getter and alias
  def email
    "custom_#{super}"
  end
  
  alias_attribute :username, :email
end

# Create a user
user = User.create!(email: "test@example.com")

# In Rails 7.0.8: Works correctly - custom method is called
user.username
# => "custom_test@example.com"

# In Rails 8.0.2: Bypasses custom method - returns raw DB value
user.username
# => "test@example.com"  # Raw database value, not custom method result
```

This change **prevents** the confusing behavior by raising an `ArgumentError` when `alias_attribute` is used with attributes that have custom methods defined, making the restriction explicit.

---

### Detail

* Fixes `alias_attribute` to raise an `ArgumentError` at alias time when attempting to alias a manually defined method making the behavior explicit instead of silently bypassing the method.
* Refactored test `"#alias_attribute with the same alias as parent doesn't issue a deprecation"` to properly test the new behavior where alias_attribute raises an error instead of silently bypassing custom methods.

**Example of the new behavior:**

```ruby
  def email
    "custom_#{super}"
  end

  alias_attribute :username, :email
end
user = User.create!(email: "test@example.com")
# => ArgumentError: User model aliases `email`, but `email` has a custom method defined, which is not allowed for `alias_attribute`. Consider using `alias_method :username, :email` instead.
```

---

### Additional information

**Related PRs and commits that introduced related changes:**
* https://github.com/rails/rails/pull/48533
* https://github.com/rails/rails/pull/49624

---

### Checklist

### DIFF
--- a/activerecord/lib/active_record/attribute_methods.rb
+++ b/activerecord/lib/active_record/attribute_methods.rb
@@ -90,6 +90,8 @@ module ActiveRecord
         if !abstract_class? && !has_attribute?(old_name)
           raise ArgumentError, "#{self.name} model aliases `#{old_name}`, but `#{old_name}` is not an attribute. " \
             "Use `alias_method :#{new_name}, :#{old_name}` or define the method manually."
+        elsif old_name != "id" && pattern.prefix == "" && (pattern.suffix == "" || pattern.suffix == "=")
+          super
         else
           define_attribute_method_pattern(pattern, old_name, owner: code_generator, as: new_name, override: true)
         end
--- a/activerecord/test/cases/attribute_methods_test.rb
+++ b/activerecord/test/cases/attribute_methods_test.rb
@@ -1539,6 +1539,48 @@ class AttributeMethodsTest < ActiveRecord::TestCase
     assert_equal message, error.message
   end
 
+  test "#alias_attribute returns custom getter value rather than raw DB value" do
+    klass = Class.new(ActiveRecord::Base) do
+      self.table_name = "topics"
+
+      define_method(:title) do
+        "custom_value"
+      end
+
+      alias_attribute :subject, :title
+    end
+
+    rec = klass.create!(title: "raw_db")
+    rec = klass.find(rec.id)
+
+    assert_equal "custom_value", rec.subject
+    assert_equal "raw_db", rec.read_attribute(:title)
+  end
+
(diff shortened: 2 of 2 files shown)


skipkayhil:
I believe the PR was https://github.com/rails/rails/pull/48533 and the change to not call other methods is intended

shravan097:
@skipkayhil @rafaelfranca Thank you for the quick review and for confirming the behavior is intentional.
My concern is the UX: if a class defines a custom method for an attribute and uses it in alias_attribute, Rails currently bypasses that method with no signal. This can be surprising and easy to miss during code reviews.

I saw that #48533 added a deprecation and #49624 replaced it with an ArgumentError, but I believe it missed checking for custom methods.
What do you think about raising the error at alias time when the target is a manually defined method, instead of silently ignoring it?


```ruby
      # active_record/attribute_methods.rb
      def alias_attribute(new_name, old_name)
        old_name = old_name.to_s
        new_name = new_name.to_s

        # Check if the target method is manually defined (custom method)
        method_defined = method_defined?(old_name) || private_method_defined?(old_name)
        manually_defined = method_defined && 
          !self.instance_method(old_name).owner.is_a?(GeneratedAttributeMethods)
        reserved_method_name = ::ActiveRecord::AttributeMethods.dangerous_attribute_methods.include?(old_name)

        if manually_defined && !reserved_method_name
          raise ArgumentError, "`alias_attribute` cannot alias `#{old_name}` because it has a custom method defined."
        end
   ```
        
 Would you be open to this approach? If so, I can make the change and add tests.

rafaelfranca:
I think we should raise, yes. Can you work on that?

PR #55483, opened 2025-08-12
Don't trigger action on path relative redirects starting with ?

TobiasBales:
### Motivation / Background

This is an extension to the [ability to react to path relative redirects](https://github.com/rails/rails/pull/55308).
We realized that I missed the case where the redirect url starts with an `?`
If this is run on `https://www.example.com` you'd end up on e.g. `https://www.example.com?foo=bar`.

### Detail

This PR implements this change and the accompanying test cases

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -199,7 +199,7 @@ module ActionController
       when /\A([a-z][a-z\d\-+.]*:|\/\/).*/i
         options.to_str
       when String
-        if !options.start_with?("/") && !options.empty?
+        if !options.start_with?("/") && !options.start_with?("?") && !options.empty?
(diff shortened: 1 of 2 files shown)


tenderlove (approved):

actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -199,7 +199,7 @@ def _compute_redirect_to_location(request, options) # :nodoc:
       when /\A([a-z][a-z\d\-+.]*:|\/\/).*/i
         options.to_str
       when String
-        if !options.start_with?("/") && !options.empty?
+        if !options.start_with?("/") && !options.start_with?("?") && !options.empty?
```
rafaelfranca:
```suggestion
        if !options.start_with?("/", "?") && !options.empty?
```

PR #55484, opened 2025-08-13
docs: document alias_attribute behavior change in Rails 7.2 upgrading guide

shravan097:
### Motivation / Background

Follow up from https://github.com/rails/rails/pull/55480#issuecomment-3180770021

I am upgrading an app from `7.0.8` to latest which is `8.0.2`. I have an existing logic that passes custom method in `alias_attribute` expecting custom method to trigger.  

The Rails 7.2 upgrading guide needs to be updated to properly document the `alias_attribute` behavior change that was introduced in Rails 7.2.

Currently, the documentation for the `alias_attribute` behavior change is missing from the upgrading guide, making it difficult for developers to understand what changed and how to handle the upgrade.

---

### Detail

* Updates the Rails 7.2 upgrading guide to clarify the `alias_attribute` behavior change

### Additional information

**Related PRs and commits that introduced related changes:**
* https://github.com/rails/rails/pull/48533
* https://github.com/rails/rails/pull/49624

---

### Checklist

### DIFF
--- a/guides/source/upgrading_ruby_on_rails.md
+++ b/guides/source/upgrading_ruby_on_rails.md
@@ -104,6 +104,60 @@ set the `queue_adapter` config to something other than `:test`, but written test
 
 If no config is provided, the `TestAdapter` will continue to be used.
 
+### `alias_attribute` now bypasses custom methods on the original attribute
+
+In Rails 7.2, `alias_attribute` now bypasses custom methods defined on the original attribute and directly accesses the underlying database value. This change was announced via deprecation warnings in Rails 7.1.
+
+**Before (Rails 7.1):**
+
+```ruby
+class User < ActiveRecord::Base
+  def email
+    "custom_#{super}"
+  end
+
+  alias_attribute :username, :email
+end
+
+user = User.create!(email: "test@example.com")
+user.username
+# => "custom_test@example.com"
+```
+
+**After (Rails 7.2):**
+
+```ruby
+user = User.create!(email: "test@example.com")
+user.username
+# => "test@example.com"  # Raw database value
+```
+
+If you received the deprecation warning "Since Rails 7.2 `#{method_name}` will not be calling `#{target_name}` anymore", you should manually define the alias method:
+
+```ruby
+class User < ActiveRecord::Base
+  def email
+    "custom_#{super}"
+  end
+
+  def username
+    email  # This will call the custom email method
+  end
+end
+```
+
+Alternatively, you can use `alias_method`:
+
+```ruby
+class User < ActiveRecord::Base
+  def email
+    "custom_#{super}"
+  end
+
+  alias_method :username, :email
+end
+```
+
 Upgrading from Rails 7.0 to Rails 7.1
 -------------------------------------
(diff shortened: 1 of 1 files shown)


activerecord/test/cases/attribute_methods_test.rb
```diff
@@ -1539,6 +1552,8 @@ def self.name
     assert_equal message, error.message
   end
 
+
+
```
byroot:
```suggestion
```

✂️

activerecord/test/cases/attribute_methods_test.rb
```diff
@@ -1604,7 +1619,13 @@ def publish
...
```
byroot:
```suggestion
```

✂️

byroot:
I'm not sure how I feel about this (cc @rafaelfranca).

On one hand I understand this can be a footgun, on the other it can sometimes be a feature. e.g. if you wish to refactor an attribute to be computed, but want to keep access to the old stored attribute as part of your migration or something.

```ruby
class Invoice < AR::Base
  alias_attribute :stored_total, :total

  def total
    taxless_total * local_tax_rate
  end
end
```

(Not necessarily the best example, but just to give an idea).

Perhaps this calls for an extra option?

matthewd:
Yeah I'm not sure that we'd want to disallow this; the current behaviour was the intended result of the previous change.

@shravan097 you've said you're upgrading from 7.0 to 8.0... did you not get a deprecation on 7.1?

yahonda:
FYI This CI failure is likely due to missing apt-key command that was removed in Debian 13 trixie. https://github.com/docker-library/ruby/pull/515 

I'll open a pull request when I have confirmed https://github.com/rails/buildkite-config/compare/main...yahonda:buildkite-config:support_debian_13_trixie works as expected.

byroot:
@yahonda sorry, I didn't see you edit and already merged: https://github.com/rails/buildkite-config/pull/14

rafaelfranca:
Ah. I missed that we show a deprecation in 7.1 https://github.com/rails/rails/blob/3c5bc25777b2f9b085e84a6bc648454958537c39/activerecord/lib/active_record/attribute_methods.rb#L118-L122

I agree we should not raise in 8.0

shravan097:
@matthewd We missed the deprecation warning, so that’s on us for not reacting sooner.
Thank you @byroot for the example — there are workarounds for this, so it’s not a blocker for my upgrade.

I was confused by this behavior and couldn’t find any documentation until I dug into the source code.

I understand why this change might create conflicting feelings. At the very least, maybe we could add a note to the docs?
@rafaelfranca @byroot I’m okay leaving the behavior as-is, but I think documenting it would be helpful for others in the future.

rafaelfranca:
Yes. Totally. Let's change this PR to document this change in the upgrading guide

shravan097:
sorry i incorrectly rebased, fixing it now. 

**update**: @rafaelfranca commit history should be fixed. sorry again. let me know what you think about the doc change.

thank you again for reviewing my issue.

PR #55487, opened 2025-08-14
Allow unsubscribing from events

zzak:
Some minor API changes around subscriptions, partly taken from ErrorReporter.

* We don't want to publicize the readers for `@subscribers` or `@raise_on_error`.
* Publicize `raise_on_error=` with docs

Some of these we didn't do with ErrorReporter (mainly `attr_*`) but I think it's a cleaner API, for example `ErrorReporter#logger` doesn't need a reader but now that it's public technically removing it should be deprecated. So just being stricter about what we publicize makes maintenance / change easier later.

I also debated whether to add `EventReporter#disable(&block)` for disabling subscriptions because I couldn't think of a use-case but it could always be added later.

### DIFF
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -261,8 +261,9 @@ module ActiveSupport
   #   #    payload: { id: 123 },
   #   #  }
   class EventReporter
-    attr_writer :raise_on_error # :nodoc:
-    attr_reader :subscribers
+    # Sets whether to raise an error if a subscriber raises an error during
+    # event emission, or when unexpected arguments are passed to +notify+.
+    attr_writer :raise_on_error
 
     class << self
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/event_reporter.rb
```diff
@@ -231,8 +231,12 @@ def clear
   #   #    payload: { id: 123 },
   #   #  }
   class EventReporter
-    attr_reader :subscribers
-    attr_accessor :raise_on_error
+    attr_reader :subscribers # :nodoc:
```
byroot: Does it even need to be public?

zzak: No, it was because test was failing and I was being lazy :joy:

adrianna-chang-shopify (approved):

PR #55488, opened 2025-08-14
Catch when msgpack gem is missing and warn

zzak:
When `ActiveSupport::EventReporter.encoder(:msgpack)` is called, typically during boot, we can catch when the user forgot to add the gem and give them a warning.

This is similar to how `ActiveSupport::Messages::SerializerWithFallback.[](format)` works

See:
https://github.com/rails/rails/blob/6f912a5986cd9295e3fbb45b1ec22159e51ad65d/activesupport/lib/active_support/messages/serializer_with_fallback.rb#L9-L14

Also `ActiveSupport::MessagePack`:
https://github.com/rails/rails/blob/2ca26346a563a375277e97a09d879df33682df55/activesupport/lib/active_support/message_pack.rb#L3-L10

Context:
https://github.com/rails/rails/pull/55334#discussion_r2226708177

### DIFF
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -235,8 +235,8 @@ module ActiveSupport
     attr_accessor :raise_on_error
 
     ENCODERS = {
-      json: Encoders::JSON,
-      msgpack: Encoders::MessagePack
+      json: "Encoders::JSON",
+      msgpack: "Encoders::MessagePack"
     }.freeze
 
     class << self
@@ -258,9 +258,10 @@ module ActiveSupport
       #
       # * +KeyError+ - If the encoder format is not found
       def encoder(format)
-        ENCODERS.fetch(format.to_sym) do
+        coder = ENCODERS.fetch(format.to_sym) do
           raise KeyError, "Unknown encoder format: #{format.inspect}. Available formats: #{ENCODERS.keys.join(', ')}"
         end
+        self.const_get(coder)
       end
     end
 
--- a/activesupport/lib/active_support/event_reporter/encoders.rb
+++ b/activesupport/lib/active_support/event_reporter/encoders.rb
@@ -30,61 +30,9 @@ module ActiveSupport
         end
       end
 
-      # JSON encoder for serializing events to JSON format.
-      #
-      #   event = { name: "user_created", payload: { id: 123 }, tags: { api: true } }
-      #   ActiveSupport::EventReporter::Encoders::JSON.encode(event)
-      #   # => {
-      #   #      "name": "user_created",
-      #   #      "payload": {
-      #   #        "id": 123
-      #   #      },
-      #   #      "tags": {
-      #   #        "api": true
-      #   #      },
-      #   #      "context": {}
-      #   #    }
-      #
-      # Schematized events and tags MUST respond to #to_h to be serialized.
-      #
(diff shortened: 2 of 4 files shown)


Earlopain (commented):
This doesn't quite work as is. All the encoders are referenced in `ActiveSupport::EventReporter::ENCODERS` which causes the warning to be emitted even when just accessing `ActiveSupport::EventReporter` (happens during `require "activesupport"`.

Maybe just constantize them in the fetch method and reference the class names as strings instead? Then you can also go back to reraising the `LoadError`. I haven't tested much but this seems to work ok:

```patch
diff --git a/activesupport/lib/active_support/event_reporter.rb b/activesupport/lib/active_support/event_reporter.rb
index da6ec6b0f7..5aafb12f37 100644
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -235,8 +235,8 @@ class EventReporter
     attr_accessor :raise_on_error
 
     ENCODERS = {
-      json: Encoders::JSON,
-      msgpack: Encoders::MessagePack
+      json: "ActiveSupport::EventReporter::Encoders::JSON",
+      msgpack: "ActiveSupport::EventReporter::Encoders::MessagePack"
     }.freeze
 
     class << self
@@ -258,13 +258,10 @@ class << self
       #
       # * +KeyError+ - If the encoder format is not found
       def encoder(format)
-        if format.to_s.include?("msgpack") && !defined?(ActiveSupport::EventReporter::Encoders::MessagePack)
-          require "active_support/event_reporter/encoders/message_pack"
-        end
-
-        ENCODERS.fetch(format.to_sym) do
+        coder = ENCODERS.fetch(format.to_sym) do
           raise KeyError, "Unknown encoder format: #{format.inspect}. Available formats: #{ENCODERS.keys.join(', ')}"
         end
+        coder.constantize
       end
     end
 
diff --git a/activesupport/lib/active_support/event_reporter/encoders/message_pack.rb b/activesupport/lib/active_support/event_reporter/encoders/message_pack.rb
index 1aa5512201..7ad7d80ac3 100644
--- a/activesupport/lib/active_support/event_reporter/encoders/message_pack.rb
+++ b/activesupport/lib/active_support/event_reporter/encoders/message_pack.rb
@@ -4,11 +4,12 @@
 begin
   gem "msgpack"
   require "msgpack"
-rescue LoadError
+rescue LoadError => error
   warn <<~MSG
     ActiveSupport::ErrorReporter::Encoders::MessagePack requires the msgpack gem.
     Please add it to your Gemfile: `gem "msgpack"`
   MSG
+  raise error
 end
 
 module ActiveSupport
```

You can also cache the resolved classes once they are accessed, don't know if that would matter here though.

Earlopain:
I think the documentation also needs some tweaking to get the most impact: https://github.com/rails/rails/pull/55334#discussion_r2237058181

I'm not too deep into this, but I guess `emit` only happens at runtime (or rather when an event as been emitted)? Does that happen before or after boot? Sorry if I misunderstand this part.

activesupport/lib/active_support/event_reporter.rb
```diff
@@ -258,9 +258,10 @@ class << self
       #
       # * +KeyError+ - If the encoder format is not found
       def encoder(format)
```
byroot, line 260: I'd just get rid of that `encoder` method, it doesn't being anything to the table.

e.g.

```ruby
ActiveSupport::EventReporter.encoder(:json)
```

Is no better than

```ruby
ActiveSupport::EventReporter::JSONEncoder
```

I'd even argue it's way worse because it makes it harder to find the implementation.

byroot:
https://github.com/rails/rails/pull/55506

PR #55489, opened 2025-08-14
Filter event reporter payloads

zzak:
For event payloads that are a hash, filter the data before emitting the event.

Since we want to filter any sensitive data as early as possible.

However, custom objects are not resolved until later, so we can't filter those.

---

Based on Shopify/rails#37 and Shopify/rails#38.

Originally discussed here:
https://github.com/rails/rails/pull/55334#discussion_r2212329630

### DIFF
--- a/activesupport/lib/active_support.rb
+++ b/activesupport/lib/active_support.rb
@@ -114,6 +114,8 @@ module ActiveSupport
   @event_reporter = ActiveSupport::EventReporter.new
   singleton_class.attr_accessor :event_reporter # :nodoc:
 
+  cattr_accessor :filter_parameters, default: [] # :nodoc:
+
   def self.cache_format_version
     Cache.format_version
   end
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "active_support/parameter_filter"
+
 module ActiveSupport
   class TagStack # :nodoc:
     EMPTY_TAGS = {}.freeze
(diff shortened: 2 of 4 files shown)


rafaelfranca:
Should we filter twice? This seems like unnecessary overhead

zzak:
> Should we filter twice? This seems like unnecessary overhead

I flip-flopped on this a bit but went with @adrianna-chang-shopify's suggestion in https://github.com/Shopify/rails/pull/38#pullrequestreview-3043294670

So we filter the payload early, but custom objects (non-Hash) are filtered later after emission, during encoding.

adrianna-chang-shopify (commented):
Thanks @zzak ! 

For Rafael -- I believe we shouldn't ever actually double filter -- either the initial payload is a hash, and we filter at resolution time, or it's an "event object", and we defer filtering to encode time.

Should we add a few lines to the docs? IMO would be good to mention:
* Hash-based payloads are automatically filtered based on `Rails.application.config.filter_parameters`
*  The default encoders handle filtering at encode time
* Maybe something along the lines of, if you're using event objects and _not_ relying on any of the provided encoders, subscribers will need to filter sensitive data themselves?

zzak:
:writing_hand: Added docs, thanks for the review!

adrianna-chang-shopify (approved):

byroot:
Heads up that this change seem to have caused some pretty massive performance regression on our app. Some controllers are spending 95% of their time in `Regexp#match?` following this change.

It first caused a noticeable degradation as part of the Rails 8.1 upgrade, but then degraded even more massively when we upgraded to edge, I suspect because now all log subscribers are going through this.

I'm still at early stage investigation, but we might have to revert this :/

byroot:
> I'm still at early stage investigation

So, It appears our filtering list also exploded in size, so this might not be entirely to blame.

PR #55490, opened 2025-08-14
Bump RuboCop to fix some transient failures in CI

Earlopain:
```
rubocop --parallel
Error: `Rails` cops have been extracted to the `rubocop-rails` gem.
(obsolete configuration found in .rubocop.yml, please update it)
````

Like here:
* https://buildkite.com/rails/rails/builds/120729/steps/canvas?sid=0198a576-b57c-485b-803a-150b40ee41df
* https://buildkite.com/rails/rails/builds/120724/steps/canvas?sid=0198a53d-2d42-49b2-8b35-97c2395f01b2

This is fixed since RuboCop 1.74 with https://github.com/rubocop/rubocop/pull/13954 but I can't tell why it would start only now.

`rubocop-md` needs a bump as well to silence some deprecation about a renamed cop.

Other changes are what rubocop now wants.

### DIFF
--- a/activerecord/test/cases/locking_test.rb
+++ b/activerecord/test/cases/locking_test.rb
@@ -859,7 +859,7 @@ end
 
 class PessimisticLockingWhilePreventingWritesTest < ActiveRecord::TestCase
   CUSTOM_LOCK = if current_adapter?(:SQLite3Adapter)
-    true # no-op
+    "FOR UPDATE" # no-op
   else
     "FOR SHARE"
   end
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: activerecord/test/cases/locking_test.rb [hunk 1]
```diff
@@ -859,7 +859,7 @@ def duel(&block)
 
 class PessimisticLockingWhilePreventingWritesTest < ActiveRecord::TestCase
   CUSTOM_LOCK = if current_adapter?(:SQLite3Adapter)
```
Earlopain, line 861: This is what rubocop wants

```suggestion
  CUSTOM_LOCK = current_adapter?(:SQLite3Adapter) || "FOR SHARE"
```

Which is not very nice, so I wrote it differently.

PR #55497, opened 2025-08-15
Add #assert_events_reported test helper

george-ma:
Add #assert_events_reported test helper

The addition of a `#assert_events_reported` helper which was developed with the Rails principles

  -  Optimizes for developer happiness - order doesn't matter, just presence
  - [ErrorReporter's `capture_error_reports`](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Testing/ErrorReporterAssertions.html#method-i-capture_error_reports) similarity - focuses on "did these events get reported?"

You can imagine a stream of logs

```
"checkout.start"
"event-x"
"event-y"
"event-z"
"checkout.end"
```

we can either develop the `assert_events_reported` helper to be
1. Very strict in order (checkout.start must come first, then event-x, event-y, etc)
- but this would make it so if a test _only_ cared about `event-x` and `event-z` for example, you would still have to include the other three events in your `assert_events_reported` block
2. Flexible but strict order (events just have to be ordered in the same way they're sent)
- This would allow for `event-x` and `event-z` to be `assert_events_reported` asserted as long as they're passed into the assertion in the right order

```ruby
        assert_events_reported([
          {
            event: "event-x",
          },
          {
            event: "event-y",
          },
        ])
```

3. No ordering (just a count for whether the provided events occur in the block provided)
I went with this option as it was the most developer friendly and matches the [ErrorReporter](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Testing/ErrorReporterAssertions.html#method-i-capture_error_reports)'s flexibility

### DIFF
--- a/activesupport/lib/active_support/testing/event_reporter_assertions.rb
+++ b/activesupport/lib/active_support/testing/event_reporter_assertions.rb
@@ -158,6 +158,61 @@ module ActiveSupport
           flunk(message)
         end
       end
+
+      # Asserts that the provided events were reported, regardless of order.
+      #
+      #   assert_events_reported([
+      #     { name: "user.created", payload: { id: 123 } },
+      #     { name: "email.sent", payload: { to: "user@example.com" } }
+      #   ]) do
+      #     create_user_and_send_welcome_email
+      #   end
+      #
+      # Supports the same payload and tag matching as +assert_event_reported+.
+      #
+      #   assert_events_reported([
+      #     {
+      #       name: "process.started",
+      #       payload: { id: 123 },
+      #       tags: { request_id: /[0-9]+/ }
+      #     },
+      #     { name: "process.completed" }
+      #   ]) do
+      #     Rails.event.tagged(request_id: "456") do
+      #       start_and_complete_process(123)
+      #     end
+      #   end
+      def assert_events_reported(expected_events, &block)
+        events = EventCollector.record(&block)
+
+        if events.empty? && expected_events.size > 0
+          flunk("Expected #{expected_events.size} events to be reported, but there were no events reported.")
+        end
+
+        events_copy = events.dup
+
+        expected_events.each do |expected_event|
+          name = expected_event[:name]
+          payload = expected_event[:payload] || {}
+          tags = expected_event[:tags] || {}
+
+          matching_event_index = events_copy.find_index { |event| event.matches?(name, payload, tags) }
+
+          if matching_event_index
+            events_copy.delete_at(matching_event_index)
+          else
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/testing/event_reporter_assertions.rb
```diff
@@ -158,6 +158,62 @@ def assert_event_reported(name, payload: nil, tags: {}, &block)
...
+      #       payload: { id: 123 },
+      #       tags: { request_id: /[0-9]+/ }
+      #     },
+      #     { name: "process.completed" }
+      #   ]) do
+      #     Rails.event.tagged(request_id: "456") do
+      #       start_and_complete_process(123)
+      #     end
+      #   end
+      def assert_multiple_events_reported(expected_events, &block)
```
eileencodes: I think the API reads better as `assert_events_reported` as that implies it's multiple events.

```suggestion
      def assert_events_reported(expected_events, &block)
```

activesupport/lib/active_support/testing/event_reporter_assertions.rb
```diff
@@ -158,6 +158,62 @@ def assert_event_reported(name, payload: nil, tags: {}, &block)
           flunk(message)
         end
       end
+
+      # Asserts that the block causes the provided events to be reported,
+      # regardless of order.
```
eileencodes: I think this can be said more simply:

```suggestion
      # Asserts that the provided events were reported, regardless of order.
```

activesupport/test/testing/event_reporter_assertions_test.rb
```diff
@@ -120,6 +120,106 @@ class EventReporterAssertionsTest < ActiveSupport::TestCase
         assert_match(/Expected no 'user\.created' event to be reported, but found:/, e.message)
         assert_match(/user\.created/, e.message)
       end
+
+      test "assert_multiple_events_reported passes when expecting empty array and no events reported" do
+        assert_multiple_events_reported([]) do
+          # No events reported
+        end
```
eileencodes: Can you instead assert that there was no notifier? Otherwise if you can't, I think this test should be deleted.

eileencodes (changes requested):
Thanks for working on this, I think the API should be `assert_events_reported` not `assert_multiple_events_reported`. Let me know when you've made the changes.

activesupport/lib/active_support/testing/event_reporter_assertions.rb
george-ma: Agreed; shorter & simpler the better!

george-ma:
Should be ready now!

PR #53146, opened 2024-10-02
RateLimiting: support method names for `:by` and `:with`

seanpdoyle:
### Motivation / Background

Prior to this commit, `:by` and `:with` options only supported callables. This commit aims to bring rate limiting closer in parity to callbacks declarations like `before_action` and `after_action` by supporting instance method names as well.

Without this change, multi-line customized callables can make declarations difficult to read. Similarly, shared customization would require a local variable to be re-used. For example:

```ruby
rate_limiting_bucket = -> { ... }

rate_limit to: 3, within: 1.minutes, by: rate_limiting_bucket
rate_limit to: 10, within: 5.minutes, by: rate_limiting_bucket
rate_limit to: 30, within: 10.minutes,
  by: -> {
    # ...
    # multiple lines
    # ...
  }, 
  with: -> {
    # ...
    # multiple lines
    # ...
  }
```

### Detail

After this change, method names can be used and shared:

```ruby
rate_limit to: 3, within: 1.minutes, by: :rate_limiting_bucket
rate_limit to: 10, within: 5.minutes, by: :rate_limiting_bucket
rate_limit to: 30, within: 10.minutes, by: :rate_limiting_bucket, with: :rate_limit_exceeded

private
  def rate_limiting_bucket
    # ...
    # multiple lines
    # ...
  end

  def rate_limit_exceeded
    # ...
    # multiple lines
    # ...
  end
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55501, opened 2025-08-15
RateLimiting: raise `ActionController::TooManyRequests` error

seanpdoyle:
### Motivation / Background

Prior to this commit, the default behavior for exceeding a rate limit involved responding with a status of [429 Too many requests][]. The only indication that a rate limit was exceeded was the `429` code.

### Detail

This commit changes the default behavior from a `head :too_many_requests` call to instead raise a new `ActionController::TooManyRequests` error. This change adheres more closely to precedent established by other status codes like `ActionController::BadRequest` to a `:bad_request` (`400 Bad Request`) status, or `ActiveRecord::RecordNotFound` to a `:not_found` (`404 Not Found`) status.

Application-side controllers can be configured to `rescue_from` that exception, or they can rely on a new
`ActionController::TooManyRequests`-to-`429 Too many requests` status mapping entry in the `ActionDispatch/Middleware/ExceptionWrapper` error-to-status mapping.

[429 Too many requests]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429

### Additional information

**This PR was extracted from https://github.com/rails/rails/pull/53146, which was approved**.

### Checklist

### DIFF
--- a/actionpack/test/controller/rate_limiting_test.rb
+++ b/actionpack/test/controller/rate_limiting_test.rb
@@ -66,8 +66,9 @@ class RateLimitingTest < ActionController::TestCase
     get :limited
     assert_response :ok
 
-    get :limited
-    assert_response :too_many_requests
+    assert_raises ActionController::TooManyRequests do
+      get :limited
+    end
   end
 
   test "notification on limit action" do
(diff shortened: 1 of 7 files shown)


actionpack/test/controller/rate_limiting_test.rb
```diff
@@ -66,8 +66,9 @@ class RateLimitingTest < ActionController::TestCase
     get :limited
     assert_response :ok
 
-    get :limited
-    assert_response :too_many_requests
+    assert_raises ActionController::TooManyRequests do
+      get :limited
+    end
```
seanpdoyle, line 71: This change won't have end-user facing consequences in production environments, but has the opportunity to impact consumer application test suites.

While this type of test change is necessary for the Action Dispatch test suite, they will not be necessary for applications that set `config.action_dispatch.show_exceptions = :all` (which is the default value).

However, test suites for applications that set `config.action_dispatch.show_exceptions = :none` will start to fail. Does that constitute enough of a breaking change to demand a new Railtie configuration and default?

byroot (approved):

PR #55502, opened 2025-08-15
Rate Limiting: Add context to request and response

seanpdoyle:
### Motivation / Background

Prior to this commit, the default behavior for exceeding a rate limit involved responding with a status of [429 Too many requests][]. The only indication that a rate limit was exceeded was the `429` code. Custom controller code that responds to rate limited requests does not have access to rate limit contextual information.

### Detail

Add readers and writers for `ActionDispatch::Request#rate_limit` and This commit introduces the `ActionController::RateLimiting::RateLimit` class to access contextual information that pertains to the rate limited request, including:

* `limit`
* `count`
* `within`
* `by`
* `name`
* `scope`
* `cache_key`

Similarly, add readers and writers for
`ActionDispatch::Response#retry_after` to write the RFC 2616-formatted `within` value to the response's [Retry-After][] header.

[429 Too many requests]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429
[Retry-After]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/rate_limiting.rb
+++ b/actionpack/lib/action_controller/metal/rate_limiting.rb
@@ -3,6 +3,40 @@
 # :markup: markdown
 
 module ActionController # :nodoc:
+  # Information derived from a request that has exceeded a rate limit
+  class RateLimit
+    # The number of requests within the time frame
+    attr_reader :count
+
+    # The maximum number of requests allowed within the time frame
+    attr_reader :limit
+
+    # The time frame as a ActiveSupport::Duration instance
+    attr_reader :within
+
+    # The value returned by the identity function used to differentiate the source of the request
+    attr_reader :by
+
+    # The name of the rate limit
+    attr_reader :name
+
+    # The cache store used to count requests. Defaults to `config.action_controller.cache_store`.
+    attr_reader :scope
+
+    # The key used to count requests in the cache store
+    attr_reader :cache_key
+
+    def initialize(count:, limit:, within:, by:, name:, scope:, cache_key:) # :nodoc:
+      @count = count
+      @limit = limit
+      @within = within
+      @by = by
+      @name = name
+      @scope = scope
+      @cache_key = cache_key
+    end
+  end
+
   module RateLimiting
     extend ActiveSupport::Concern
 
@@ -75,6 +109,17 @@ module ActionController # :nodoc:
         cache_key = ["rate-limit", scope, name, by].compact.join(":")
         count = store.increment(cache_key, 1, expires_in: within)
         if count && count > to
+          request.rate_limit = RateLimit.new(
+            limit: to,
+            count: count,
+            within: within,
+            by: by,
+            name: name,
+            scope: scope,
+            cache_key: cache_key
+          )
+          response.retry_after = within.from_now.httpdate
+
           ActiveSupport::Notifications.instrument("rate_limit.action_controller",
               request: request,
               count: count,
--- a/actionpack/lib/action_dispatch/http/request.rb
+++ b/actionpack/lib/action_dispatch/http/request.rb
@@ -498,6 +498,16 @@ module ActionDispatch
       controller_instance.commit_csrf_token(self) if controller_instance.respond_to?(:commit_csrf_token)
     end
 
+    # Write the ActionController::RateLimit values collected during the request
+    def rate_limit=(value)
+      set_header "action_controller.rate_limit", value
+    end
+
+    # Read the ActionController::RateLimit values collected during the request
+    def rate_limit
+      get_header "action_controller.rate_limit"
+    end
+
     private
       def check_method(name)
         if name
--- a/actionpack/lib/action_dispatch/http/response.rb
+++ b/actionpack/lib/action_dispatch/http/response.rb
@@ -396,6 +396,30 @@ module ActionDispatch # :nodoc:
       end
     end
 
+    # Write a Time to the Retry-After HTTP header.
+    #
+    # Expects a value formatted according to RFC 2616 or a non-negative
+    # integer indicating the seconds to delay after the response is received.
+    #
+    #     response.retry_after = 5.minutes.from_now.httpdate
+    #     response.retry_after = 300
+    #
+    # Date, Time, DateTime, and ActiveSupport::TimeWithZone instances will be
+    # coerced to a formatted string.
+    #
+    #     response.retry_after = 5.minutes.from_now
+    def retry_after=(value)
+      value = value.httpdate if value.respond_to?(:httpdate)
+      value = value.to_i unless value.is_a?(String)
+
+      set_header "Retry-After", value
+    end
+
+    # Read from the Retry-After HTTP header.
+    def retry_after
+      get_header "Retry-After"
+    end
+
     # Avoid having to pass an open file handle as the response body. Rack::Sendfile
     # will usually intercept the response and uses the path directly, so there is no
     # reason to open the file.
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Access rate limiting context from `ActionDispatch::Request` and `ActionDispatch::Response`
+
+    Add `ActionDispatch::Request#rate_limit` and `ActionDispatch::Response#retry_after`.
+
+    *Sean Doyle*
+
 *   Define `ActionController::Parameters#deconstruct_keys` to support pattern matching
 
     ```ruby
--- a/actionpack/test/controller/rate_limiting_test.rb
+++ b/actionpack/test/controller/rate_limiting_test.rb
@@ -164,6 +164,32 @@ class RateLimitingTest < ActionController::TestCase
     assert_response :forbidden
   end
 
+  test "exceeding limit writes to the request and response" do
+    freeze_time
+
+    5.times do
+      get :limited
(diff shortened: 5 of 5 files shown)


actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -62,12 +62,26 @@ def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { head :too_
       end
     end
 
+    # Information derived from a request that has exceeded a rate limit
+    RateLimit = Struct.new(:count, :limit, :within, :by, :name, :scope, :cache_key)
```
seanpdoyle: I might be mistaken, but I haven't found other occurrences of publicly documented `Struct` classes. 

What is the recommended way to more thoroughly document the attributes like `count` and `limit`.

Is `:method:` the recommended approach? Is that possible to nest alongside the class-level comment?

Is it better to declare `class RateLimit < Struct.new(…)`, despite the fact that it inserts an additional class in the hierarchy?

```ruby
    # Information derived from a request that has exceeded a rate limit
    class RateLimit < Struct.new(:count, :limit, :within, :by, :name, :scope, :cache_key)
      ##
      # :method: count
      #
      # The number of requests within the time frame

      ##
      # :method: limit
      #
      # The maximum number of requests allowed within the time frame
    end
```

MatheusRich: Thoughts on making this a data object instead?
```suggestion
    RateLimit = Data.define(:count, :limit, :within, :by, :name, :scope, :cache_key)
```

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -68,13 +68,27 @@ def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { raise TooM
       end
     end
 
+    # Information derived from a request that has exceeded a rate limit
+    RateLimit = Data.define(:count, :limit, :within, :by, :name, :scope, :cache_key)
```
zzak: I don't normally see Data being used in runtime, usually just tests

actionpack/lib/action_controller/metal/rate_limiting.rb
seanpdoyle: @zzak relating to https://github.com/rails/rails/pull/55502#discussion_r2383792762, is a `Struct` an improvement on `Data`? If not, is it recommended to pair a constructor with instance variables and attr accessors?

zzak: What is wrong with a normal class?

seanpdoyle: I've pushed up a change to use constructor, instance variables, and `attr_reader` declarations.

actionpack/lib/action_dispatch/http/request.rb
```diff
@@ -498,6 +498,18 @@ def commit_csrf_token
...
+
+    # Write the ActionController::RateLimiting::RateLimit values collected during the request
+    def rate_limit=(value)
+      set_header ACTION_DISPATCH_RATE_LIMIT, value
+    end
+
+    # Read the ActionController::RateLimiting::RateLimit values collected during the request
+    def rate_limit
+      get_header ACTION_DISPATCH_RATE_LIMIT
+    end
```
byroot:
```suggestion
    # Write the ActionController::RateLimiting::RateLimit values collected during the request
    def rate_limit=(value)
      set_header "action_controller.rate_limit", value
    end

    # Read the ActionController::RateLimiting::RateLimit values collected during the request
    def rate_limit
      get_header "action_controller.rate_limit"
    end
```

- I don't think we need that constant. It's an old style that predate frozen literal strings.
- I think the namespace should be `action_controller.` not action dispatch.

actionpack/lib/action_dispatch/http/response.rb
```diff
@@ -396,6 +397,20 @@ def body=(body)
       end
     end
 
+    # Write to the Retry-After HTTP header.
+    #
+    # Expects a value formatted according to RFC 2616.
+    #
+    #     response.retry_after = 5.minutes.from_now.httpdate
+    def retry_after=(value)
+      set_header RETRY_AFTER, value
```
byroot: I'm thinking we could make this method a little bit more handy by accepting time and calling `#http_date` on it automatically. Other setters like `#content_type=` do casting as well.

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -68,13 +68,58 @@ def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { raise TooM
       end
     end
 
+    # Information derived from a request that has exceeded a rate limit
+    class RateLimit
```
byroot: I'm wondering if `ActionController::RateLimit` would be better than `ActionController::RateLimiting::RateLimit`.

There's some needless redundancy here IMO.

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -3,6 +3,40 @@
...
+    # The time frame as a ActiveSupport::Duration instance
+    attr_reader :within
+
+    # The value returned by the identity function used to differentiate the source of the request
+    attr_reader :by
+
+    # The name of the rate limit
+    attr_reader :name
+
+    # The cache store used to count requests. Defaults to `config.action_controller.cache_store`.
```
codergeek121, line 23:
```suggestion
    # The scope to apply the rate limit to. Defaults to the `controller_path` but can be changed to share rate limits
```

actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -75,6 +109,17 @@ def rate_limiting(to:, within:, by:, with:, store:, name:, scope:)
...
+          request.rate_limit = RateLimit.new(
+            limit: to,
+            count: count,
+            within: within,
+            by: by,
+            name: name,
+            scope: scope,
+            cache_key: cache_key
+          )
+          response.retry_after = within.from_now.httpdate
```
codergeek121, line 121: I love the that you set the header, but imo this implementation has one downside:
It doesn't take into account of time that already has passed for a rate limited client. Imagine a `within` of 10.minutes, a rate limited client making a request in minute 6 will see a header pointing 10 minutes in the future, although it should probably be 4 minutes instead.

Does this make sense to you, wdyt?

seanpdoyle: It does! It had crossed my mind. I'm curious if there is a way to know when a rate limiting "window" started. As far as I can tell, the underlying cache store is changes the timestamp with each request. Am I missing a mechanism that could help infer how long it's been since rate limiting started?

codergeek121: I thought about it, but the only accurate way that I could come up with is to look at the cache entry expiry directly. Unfortunately I don't think there's a public api for reading the expiry of an `ActiveSupport::Cache::Store` entry. Alternatively, the cache entry's value could contain a count and an expiry. This way, we could just read both the count and expiry. This solution seems wasteful though...

seanpdoyle: Is that a *limitation* of the implementation, or is that the intent? 

1. If it **is** important to track the start of the window, then requests **within** the window would be ignored, and support for requests would be restored X minutes after the **first limited request**
2. If it **is not** important to track the start of the window, then requests **within** the window sustain or restart the limiting, and support for requests would only be restored X minutes after the **last limited request**


I think it's a question of intent. I'm not sure which was the intended behavior, or which is best. Does that make sense?

codergeek121: I'm not sure if I got you right, but let me try to restate what I understood:

> As far as I can tell, the underlying cache store is changes the timestamp with each request

Maybe this depends on the concrete cachestore implementation, but I'd expect at least Redis to not increment the expiry if called multiple times https://github.com/rails/rails/blob/6152c8f0692f011fd00562cfeb3ec8911eb1ae6d/activesupport/lib/active_support/cache/redis_cache_store.rb#L466 is called  with [NX](https://redis.io/docs/latest/commands/expire/#options) or equivalent. With NX, once a key expires the expiry will not be updated.

> If it is important to track the start of the window, then requests within the window would be ignored, and support for requests would be restored X minutes after the first limited request

You're wondering, if it's intentional that `ActionController::RateLimit` implements a "[fixed window with a user-defined start (1)](https://smudge.ai/blog/ratelimit-algorithms#fixed-window-with-user-defined-start)" rate limit? That's a good question 🤔

If Rails were to implement a simple "[fixed window (2)](https://smudge.ai/blog/ratelimit-algorithms#fixed-windows)" approach, I think we could calculate the end of a window without saving additional information or reading the cache.
I think changing the rate limiting strategy would technically be a breaking change though.

janko: I was studying Rack::Attack a bit, and noticed it uses fixed time buckets, which don't start on first request, but instead start and end at specific times. For example, for a time window of 10 minutes, if the first request came in at start of minute 7, the rate limit would reset after 4 minutes. This allows calculating the end of the time bucket without retrieving cache expiration.

PR #55504, opened 2025-08-17
Fix incorrect cache store comment in production.rb template [ci skip]

gongo:
### Motivation / Background

This Pull Request has been created because the comment in the production.rb template incorrectly describes
the default cache store. The comment states the default is "in-process memory cache store" when it's
actually "file cache store" (as defined in [railties/lib/rails/application/configuration.rb](https://github.com/gongo/rails/blob/092c3e92efa8e762e3d2e32506ef894702f0702d/railties/lib/rails/application/configuration.rb#L58)).

### Detail

This Pull Request changes the comment.

### Additional information

> #### `config.cache_store`
> Configures which cache store to use for Rails caching. (snip). Defaults to `:file_store`.
>
> [guides/source/configuring.md](https://github.com/gongo/rails/blob/092c3e92efa8e762e3d2e32506ef894702f0702d/guides/source/configuring.md#configcache_store)

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -62,7 +62,7 @@ Rails.application.configure do
   # Don't log any deprecations.
   config.active_support.report_deprecations = false
 
-  # Replace the default in-process memory cache store with a durable alternative.
+  # Replace the default file cache store with a durable alternative.
   # config.cache_store = :mem_cache_store
 
   <%- unless options[:skip_active_job] -%>


railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -62,7 +62,7 @@ Rails.application.configure do
   # Don't log any deprecations.
   config.active_support.report_deprecations = false
 
-  # Replace the default in-process memory cache store with a durable alternative.
+  # Replace the default file cache store with a durable alternative.
   # config.cache_store = :mem_cache_store
```
byroot, line 66: Here `mem_cache_store` refers to `memcached`.

The comment is indeed inacurate, but because there is no default.

byroot: Actually, `nil` indeed defaults to `MemoryStore`. So the comment is accurate.

gongo: Thank you for the review. Indeed, [lookup_store](https://github.com/rails/rails/blob/2a0e057570cbfd39fc38e35d97d1c43112c44ebf/activesupport/lib/active_support/cache.rb#L94-L95) defaults to MemoryStore. 💪

PR #55506, opened 2025-08-17
Polish ActiveSupport::EventReporter

byroot:
Followup: https://github.com/rails/rails/pull/55334
Fix: https://github.com/rails/rails/pull/55488

This is just a quick pass on the ore noticeable issues.

- Get rid of useless `typed: true` comments.
- Get rid of `EventReporter.encode` method. This doesn't bring much value and make it hard to lookup the documentation / source code.
- Move json and msgpack encoder in their own autoloaded file, allows to check for dependency presence on load, rather than every time `.encode` is called.
- Report handled errors with the error reporting API rather than Ruby warnings.

After going over this I have a bunch of questions though:

  - Why are we calling raw `JSON.dump` / `MessagePack.pack` shouldn't this use the Active Support JSON and MessagePack layers? e.g. shouldn't this support the `#as_json` interface?

cc @rafaelfranca @adrianna-chang-shopify


adrianna-chang-shopify:
Hey @byroot , thanks for working on this!

> Why are we calling raw JSON.dump / MessagePack.pack shouldn't this use the Active Support JSON and MessagePack layers? e.g. shouldn't this support the #as_json interface?

Yes great question -- this was not intentional, we added the encoders bit later on to that PR, and it was a miss on my part to not integrate with AS JSON / MessagePack.

The JSON use case is straightforward, I'd be keen to get your thoughts on integrating with `ActiveSupport::MessagePack`, though. Since the encoded events are likely to be exported over the wire and consumed by a telemetry / analytics platform, we don't really care about round-trip Ruby serialization / deserialization. Instead, it would make more sense to serialize them as plain data types. Should the expectation be that these event classes also implement `as_json`, which we'd call prior to dumping to MessagePack? For example:

```ruby
    class TestEvent
      def initialize(data)
        @data = data
      end

      def as_json
        { "data" => @data }
      end
    end

    class HttpRequestTag
      def initialize(http_method, http_status)
        @http_method = http_method
        @http_status = http_status
      end

      def as_json
        { "http_method" => @http_method, "http_status" => @http_status }
      end
    end
```

```ruby
  class EventReporter
    module MessagePackEncoder
      include ActiveSupport::MessagePack::Serializer
      extend self

      class << self
        def encode(event)
          # Serialize payload / tag objects to JSON via `as_json` first before dumping to MessagePack ??
          dump(event)
        end
      end
    end
```

byroot:
👋 

> I'd be keen to get your thoughts on integrating with ActiveSupport::MessagePack, though. Since the encoded events are likely to be exported over the wire and consumed by a telemetry / analytics platform, we don't really care about round-trip Ruby serialization / deserialization.

Indeed.

> it would make more sense to serialize them as plain data types.

Yes, which begs the question of whether it should be the responsability of the encoder or of a shared layer above.

In the case of JSON, you can rely on the `as_json` interface, but perhaps you would want to serialize an object differently in event logs and in the "frontend".

Overall, it might be best to stick with Rails moto, which is that it is supposed to be mostly an extraction framework. If the msgpack use case isn't quite clear because you have no use for it, perhaps it simply shouldn't be there until people come up with real world use case.

adrianna-chang-shopify:
Hey @byroot !

Yep, I definitely see your point that this feels like a premature abstraction, and that the shape of what we have doesn't quite feel right given the distinction between serialization / encoding.

I can either:
1) Remove all of the "default encoders" logic entirely, and we (someone else) can revisit once there is a clearer pattern to extract, or
2) We can remove the msgpack support and leave the JSON support given the existence of the `as_json` interface. Acknowledging, however, that this _assumes_ we want to serialize an event object in the same way as a frontend JSON API _and_ that all the `EventReporter::JSONEncoder` would be doing is calling `ActiveSupport::JSON.encode(event)` -- so perhaps not worth having.

It sounds like we may just want to remove the default encoders entirely. What do you think?

byroot:
> What do you think?

Not much. As I said early I don't have real world experience with this, so I don't really understand the ins and out of what this is supposed to do.

But as a more generic answer, if you aren't confident the current interface is what it should be, taking a step back and removing it until you get more production experience with whatever you are building at Shopify would make sense to me.

adrianna-chang-shopify:
Okay, thanks Jean. Let's remove it.

PR #55515, opened 2025-08-19
Remove default JSON and MessagePack encoders from EventReporter.

adrianna-chang-shopify:
### Motivation / Background

Based on some discussion in https://github.com/rails/rails/pull/55506 with @byroot .

Currently, we mandate that event objects implement `to_h` in order to be serialized, and then the default encoders encode them to JSON / MessagePack. But, it would be nice if we could benefit from the automatic serialization that `ActiveSupport::JSON` provides -- and additionally, rely on the `as_json` interface for serialization, rather than `to_h`. When discussing `ActiveSupport::MessagePack`, it became clear that whatever Rails implements as a "default encoder" for the Event Reporter may need to split the ideas of "serialization" and "encoding" for event objects -- this way, applications have the flexibility to serialize event objects however they choose, and select an encoding format independently.

But, the pattern to extract is not clear enough yet -- so, I'm proposing we remove the default encoders until we have a stronger use case (either at Shopify, or in another Rails app) to provide guidance here.

cc @zzak in case you have thoughts here too, and because I know some of your existing PRs deal with the encoders :) 

### Detail

Remove the default encoders from the Event Reporter.

### Checklist


activesupport/CHANGELOG.md
```diff
@@ -50,19 +50,7 @@
...
-    class MySubscriber
-      def emit(event)
-        encoded_event = ActiveSupport::EventReporter::JSONEncoder.encode(event)
-        StructuredLogExporter.export(encoded_event)
-      end
-    end
-
-    Rails.event.subscribe(MySubscriber.new)
-    ```
+    control how events are serialized and emitted.
```
zzak, line 53: I understand why we're removing features we don't need, but this part feels incomplete to me.
Am I missing something, or did we not explain how to write a subscriber somewhere?

I take this as we don't provide a public API, but there is a interface for how to write a subscriber (responds to :emit) -- so should we at least document it? Or should we raise an error if the thing you subscribe with doesn't provide that method?

adrianna-chang-shopify: @zzak I didn't realize that this effectively removes all of the examples of actual subscriber implementations 🤦‍♀️  We do document the `emit` interface, and lay out what the Event hash looks like here:
https://github.com/rails/rails/blob/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/activesupport/lib/active_support/event_reporter.rb#L223-L234

And additionally we raise if a subscriber doesn't implement `#emit`:
https://github.com/rails/rails/blob/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/activesupport/lib/active_support/event_reporter.rb#L242-L244

But agree that having an example would still be helpful (that was an oversight on my part). Let me add something back in to the docs.

zzak: That works, thanks for catching!

PR #55516, opened 2025-08-19
[Docs] Fix some of the API docs for the Event Reporter.

adrianna-chang-shopify:
### Detail

Indentation on one of the code examples was incorrect, and we forgot to add `context` to the expected fields in the `Event` hash.

### DIFF
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -98,19 +98,19 @@ module ActiveSupport
   # If an event object is passed to the +notify+ API, it will be passed through to subscribers as-is, and the name of the
   # object's class will be used as the event name.
   #
-  # class UserCreatedEvent
-  #   def initialize(id:, name:)
-  #     @id = id
-  #     @name = name
-  #   end
+  #   class UserCreatedEvent
+  #     def initialize(id:, name:)
+  #       @id = id
+  #       @name = name
+  #     end
   #
-  #   def to_h
-  #     {
-  #       id: @id,
-  #       name: @name
-  #     }
+  #     def to_h
+  #       {
+  #         id: @id,
+  #         name: @name
+  #       }
+  #     end
   #   end
-  # end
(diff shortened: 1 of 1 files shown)


activesupport/lib/active_support/event_reporter.rb
```diff
@@ -257,6 +257,7 @@ def initialize(*subscribers, raise_on_error: false, tags: nil)
     #   name: String (The name of the event)
     #   payload: Hash, Object (The payload of the event, or the event object itself)
     #   tags: Hash (The tags of the event)
+    #   context: Hash (The context of the event)
```
skipkayhil, line 262: Should this be added to `notify` examples too?

activesupport/lib/active_support/event_reporter.rb
```diff
@@ -98,19 +98,19 @@ def clear
...
-  #     {
-  #       id: @id,
-  #       name: @name
-  #     }
+  #    def to_h
+  #      {
+  #         id: @id,
+  #         name: @name
+  #       }
+  #    end
```
skipkayhil: I think ~these~ `to_h` is still 1 space short?

adrianna-chang-shopify: Yeah the spacing definitely still looks funked up here 🤦‍♀️

activesupport/lib/active_support/event_reporter.rb
adrianna-chang-shopify: Yeah, we probably should 😅  I was trying to keep the other examples brief by omitting an empty context hash, but since we're including `tags: {}` there we should probably include the context for consistency's sake.

adrianna-chang-shopify: Okay, went back to the other examples and filled out any missing attributes. Thanks for the review @skipkayhil ! ❤️

skipkayhil (approved):

PR #55519, opened 2025-08-20
Use target model's `destroy_association_async_job` when available

hubb:
### Motivation / Background

When destroying associations with `dependent: :destroy_async`, Rails now respects the target model's `destroy_association_async_job` configuration instead of always using the owner model's job.

This allows models to define their own custom destruction jobs that will be used regardless of which association triggered the destruction:

```
    class Parent < ApplicationRecord
      self.destroy_association_async_job = ParentDestroyJob
      has_many :children, dependent: :destroy_async
    end

    class Child < ApplicationRecord
      self.destroy_association_async_job = ChildDestroyJob
    end
```

Previously, destroying a parent would use `ParentDestroyJob` for the children. Now it correctly uses `ChildDestroyJob`.

The change maintains backward compatibility by falling back to the owner model's job when the target model has no custom job configured.

### Checklist


activerecord/lib/active_record/associations/association.rb
```diff
@@ -396,7 +396,10 @@ def skip_statement_cache?(scope)
         end
 
         def enqueue_destroy_association(options)
-          job_class = owner.class.destroy_association_async_job
+          job_class = if (association_class = options[:association_class])
+            association_class.safe_constantize.destroy_association_async_job
+          end
+          job_class ||= owner.class.destroy_association_async_job
```
rafaelfranca: Does this even work? `options[:association_class]` is always passed, and `destroy_association_async_job` is always set, so it will never fallback to `owner.class.destroy_association_async_job`, unless of course you set to `nil` explicitly like you did in the test.

rafaelfranca (changes requested):
Not sure this new behavior makes sense. I think `destroy_association_async_job` was made to change the job of all associations declared in the model. If we want to change which job will run to destroy a single association we should pass it as option to the association.

```
has_many :children, dependent: :destroy_async, destroy_job: "SomeJob"`

rafaelfranca (commented):
I think the behavior you want makes more sense. It was never the intention to use the job in the owner class. I think we can just swap this behavior using a configuration option to allow people to move to the new behavior.

matthewd:
> It was never the intention to use the job in the owner class

@rafaelfranca I disagree; it's a configuration on the owner class, describing how to handle its association. The arguments to the job are the owner instance and the association name -- that very much feels like something that belongs to the owner, not the association target class (without even contemplating the possibilities of STI).

The behaviour proposed here would be a `destroy_async_job` or something, that described how to destroy the model being configured.

The current behaviour seems like it would be desirable for anyone implementing e.g. a papertrail-type versioning system for a given model and its associations.

PR #55520, opened 2025-08-20
Allow defining custom attributes on jobs

dasch:
This allows attaching arbitrary attributes to jobs, including from included modules.

### Motivation / Background

The target use cases would be orthogonal to the specific jobs, e.g. consistent auditing metadata. My concrete use case is the ability to attach OTel [Span Links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to the traces for job execution, linking back to the site where the job was enqueued. This allows going back and forth between traces in order to debug e.g. why specific arguments were set on a job. I'm using Datadog for this, but it's an OTel concept.

### Detail

This Pull Request changes Active Job and adds support for setting _attributes_.

For example:

```ruby
class ApplicationJob < ActiveJob::Base
  # Simplified example of adding tracing metadata
  attribute :trace_id
  around_perform { |job, block| Tracer.trace(parent_trace_id: job.trace_id, &block) }
end

# Subclasses of ShardedJob will be executed against a specific shard
class ShardedJob < ApplicationJob
  # This would also be useful for other domain-orthogonal concerns, e.g. sharding
  attribute :shard

  before_enqueue { |job| raise "you must set `shard`!" if job.shard.nil? }

  around_perform do |job, block|
    ActiveRecord::Base.connected_to(shard: job.shard.to_sym, &block)
  end
end
```

The attributes are made available as normal accessors, but are also integrated with `ActiveJob::Base#set`:

```ruby
UserDeletionJob.new(user_id).set(shard: account.shard).enqueue
```

### Checklist

### DIFF
--- a/activejob/lib/active_job/arguments.rb
+++ b/activejob/lib/active_job/arguments.rb
@@ -75,7 +75,7 @@ module ActiveJob
       end
     end
 
-    def deserialize_argument(argument)
+    def deserialize_argument(argument) # :nodoc:
       case argument
       when nil, true, false, String, Integer, Float
         argument
--- a/activejob/lib/active_job/arguments.rb
+++ b/activejob/lib/active_job/arguments.rb
@@ -75,6 +75,25 @@ module ActiveJob
       end
     end
 
+    def deserialize_argument(argument)
+      case argument
+      when nil, true, false, String, Integer, Float
+        argument
+      when Array
+        argument.map { |arg| deserialize_argument(arg) }
+      when Hash
+        if serialized_global_id?(argument)
+          deserialize_global_id argument
+        elsif custom_serialized?(argument)
+          Serializers.deserialize(argument)
+        else
+          deserialize_hash(argument)
+        end
+      else
+        raise ArgumentError, "Can only deserialize primitive arguments: #{argument.inspect}"
+      end
+    end
+
     # Deserializes a set of arguments. Intrinsic types that can safely be
     # deserialized without mutation are returned as-is. Arrays/Hashes are
     # deserialized element by element. All other types are deserialized using
@@ -107,25 +126,6 @@ module ActiveJob
       ].to_set
       private_constant :RESERVED_KEYS, :GLOBALID_KEY,
         :SYMBOL_KEYS_KEY, :RUBY2_KEYWORDS_KEY, :WITH_INDIFFERENT_ACCESS_KEY
-
-      def deserialize_argument(argument)
-        case argument
-        when nil, true, false, String, Integer, Float
-          argument
-        when Array
-          argument.map { |arg| deserialize_argument(arg) }
-        when Hash
-          if serialized_global_id?(argument)
-            deserialize_global_id argument
-          elsif custom_serialized?(argument)
-            Serializers.deserialize(argument)
-          else
-            deserialize_hash(argument)
-          end
-        else
-          raise ArgumentError, "Can only deserialize primitive arguments: #{argument.inspect}"
-        end
-      end
 
       def serialized_global_id?(hash)
         hash.size == 1 && hash.include?(GLOBALID_KEY)
--- a/activejob/lib/active_job/attributes.rb
+++ b/activejob/lib/active_job/attributes.rb
@@ -47,7 +47,7 @@ module ActiveJob
           raise ArgumentError, "Attribute #{name} is already defined"
         end
 
-        attributes.add(name.to_s)
+        attributes.include?(name.to_s)
         attr_accessor name
       end
 
--- a/activejob/lib/active_job/attributes.rb
+++ b/activejob/lib/active_job/attributes.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+
+module ActiveJob
+  # = Active Job \Attributes
+  #
+  # Allows defining custom job attributes that will be serialized and deserialized along with
+  # the job arguments. Useful for cross-cutting concerns.
+  #
+  #   module ShardedJob
+  #     extend ActiveSupport::Concern
+  #
+  #     included do
+  #       attribute :shard
+  #       attribute :class_name
+  #
+  #       around_perform do |job, block|
+  #         klass = job.class_name&.constantize || ActiveRecord::Base
+  #         klass.connected_to(shard: job.shard.to_sym, &block)
+  #       end
+  #     end
+  #   end
+  #
+  #   class UserDeletionJob < ActiveJob::Base
+  #     include ShardedJob
+  #
+  #     # ...
+  #   end
+  #
+  # Once an attribute has been configured, it can be set on job instances with normal assignment
+  # _or_ it can be set via the +set+ method:
+  #
+  #   job = UserDeletionJob.new(user_id)
+  #   job.set(shard: shard_id)
+  #   job.enqueue
+  #
+  # This includes the various ways that +set+ can be called, e.g.
+  #
+  #   UserDeletionJob.set(shard: shard_id).new(user_id).enqueue
+  #   UserDeletionJob.new(user_id).enqueue(shard: shard_id)
+  #
+  module Attributes
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def attribute(name)
+        if respond_to?(name)
+          raise ArgumentError, "Attribute #{name} is already defined"
+        end
+
+        attributes.add(name.to_s)
+        attr_accessor name
+      end
+
+      # The attributes defined on the class, as well as their options.
+      #
+      # Will be inherited to subclasses.
+      def attributes
+        @attributes ||=
+          if superclass.respond_to?(:attributes)
+            superclass.attributes.dup
+          else
+            Set.new
+          end
+      end
+    end
+
+    def set(options = {})
+      options.each_key do |name|
+        if self.class.attributes.include?(name.to_s)
+          send("#{name}=", options.delete(name))
+        end
+      end
+
+      super
+    end
+
+    def serialize
+      data = super
+
+      self.class.attributes.each do |name|
+        data[name] = Arguments.serialize_argument(send(name))
+      end
+
+      data
+    end
+
+    def deserialize(job_data)
+      self.class.attributes.each do |attr|
+        send("#{attr}=", Arguments.deserialize_argument(job_data[attr]))
+      end
+
+      super
+    end
+  end
+end
--- a/activejob/test/cases/attributes_test.rb
+++ b/activejob/test/cases/attributes_test.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+require "helper"
+
+class AccountScopedJob < ActiveJob::Base
+  attribute :account_id
+  attribute :metadata
+
+  def perform
+    JobBuffer.add({ account_id:, metadata: })
+  end
+end
+
+class SubClassedJob < AccountScopedJob
+end
+
+class AttributesTest < ActiveSupport::TestCase
+  setup do
+    JobBuffer.clear
+  end
+
+  test "serializes and deserializes the defined attributes" do
+    job_data = AccountScopedJob.new.set(account_id: 42, metadata: { user_ids: [1, 2] }).serialize
+    job = AccountScopedJob.new
+    job.deserialize(job_data)
+
+    assert_equal 42, job.account_id
+    assert_equal({ user_ids: [1, 2] }, job.metadata)
+  end
+
+  test "integrates with #set" do
+    job = AccountScopedJob.new.set(account_id: 42)
+    assert_equal 42, job.account_id
+  end
+
+  test "defines attribute accessors" do
+    job = AccountScopedJob.new
+    job.account_id = 42
+    assert_equal 42, job.account_id
+  end
+
+  test "passes attributes to the job execution" do
+    AccountScopedJob.set(account_id: 42, metadata: { user_ids: [1, 2] }).perform_now
+    assert_equal({ account_id: 42, metadata: { user_ids: [1, 2] } }, JobBuffer.last_value)
+  end
+
+  test "defaults attributes to nil" do
+    AccountScopedJob.perform_now
+    assert_nil JobBuffer.last_value[:account_id]
+    assert_nil JobBuffer.last_value[:metadata]
+  end
+
+  test "raises when defining an existing attribute" do
+    assert_raises(ArgumentError) do
+      Class.new(ActiveJob::Base) do
+        attribute :queue_name
+      end
+    end
+  end
+
+  test "subclasses inherit the parent class attributes" do
+    job = SubClassedJob.new.set(account_id: 42)
+    assert_equal 42, job.account_id
+  end
+end
--- a/activejob/lib/active_job/base.rb
+++ b/activejob/lib/active_job/base.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_job/core"
+require "active_job/attributes"
 require "active_job/queue_adapter"
 require "active_job/queue_name"
 require "active_job/queue_priority"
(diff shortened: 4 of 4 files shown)


activejob/lib/active_job/core.rb
```diff
@@ -108,7 +111,8 @@ def initialize(*arguments)
       @priority   = self.class.priority
       @executions = 0
       @exception_executions = {}
-      @timezone   = Time.zone&.name
+      @timezone = Time.zone&.name
+      @headers = {}
```
dasch: I don't know if I should instead use HashWithIndifferentAccess here? Maybe that would be more in line with Rails idioms.

bensheldon:
There’s already an existing way to extend jobs with metadata: `#set`. I generally recommend that developers create specific accessors, rather than general ones, for their uses; being specific allows precise serialization, as well as precise rules for how values are mutated during retries.

afaict OTel already does this for Active Job: https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb

dasch:
@bensheldon that seems pretty involved for in-application use cases – would it not make sense to have a more accessible way to associate metadata with jobs, especially for large applications with lots of different job classes?

bensheldon:
> would it not make sense to have a more accessible way to associate metadata with jobs

I don’t think a hash is it. 

As I wrote, Active Job already has a method for adding parameters that are shared across jobs: `set`, and a mechanism for configuring it through code. If I concede that “through code” could be easier, I personally would want to see something closer to the Active Record Attributes API that is explicit in declaring parameters and defining their behavior across the job lifecycle.

dasch:
> [...] their behavior across the job lifecycle

What that just be the serialization or do you have something else in mind?

I could imagine an API like this, but I'm not sure it's _that_ much better than just a plain hash – but I'm up for it if that's the preference!

```ruby
module AccountScoping
  extend ActiveSupport::Concern

  included do
    attribute :account_id, :integer # do we really need typing here?
  end
end

class ApplicationJob < ActiveJob::Base
  include AccountScoping
end

# ...

# I figure this is how it would then work?
job = RemoveStalePostsJob.new(since: 2.weeks.ago)
job.enqueue(account_id: @account.id)
```

bensheldon:
For maximum transparency: I don't have any official authority in the matter, I'm just a motivated member of the Rails community.

Thing that I imagine people would want, either immediately or eventually:

- Default value
- Serializer (I imagine the default would be the argument serializer)
- As you mention, maybe people won't want type coercion, but maybe they would (that's nice if it's already built into Active Model Attributes)

I would expect the api to be:

```ruby
MyJob.set(my_attribute: "something").perform_later(job_args)

# or
my_job.set(my_attribute: "something")

# or
my_job.my_attribute = "something"
my_job.my_attribute #=> "something" 
```

I don't imagine it would be a lot more on top of including Active Model Attributes and then handling set and serialize/deserialize.

dasch:
I'm thinking that of these, only a `required` flag is a use case I would have had in the last 15 years of Rails dev; custom serialization and type coercion do not seem like use cases that need to be supported until there's a proven need. I can add support for that and then leave the other things as future improvements if those use cases do materialize.

dasch:
OK, I've replaced the implementation with one based on "attributes".

dasch:
(there seems to be a regression on `main`)

bensheldon:
This is on track with what I was imagining. Conceptually, the main deviation is that it treats attribute serialization as separate behaior of the job, serializing them into their own `attributes` key, rather than as sugar/dsl on top of set/serialize/deserialize.

So I would imagine that:

```ruby
attribute :trace_id
```

...would be serialized to the `"trace_id"` key in the serialized params.

I also think the guard about the method already being defined would prevent someone from overloading the accessor and calling `super`.

In my mind, things like `priority` or `queue` could ultimately be refactored to use the attribute api we're sketching out.

And I wonder how much of this could be done directly by `include ActiveModel::Attributes` and then overloading the `attribute` class method to slightly change/exclude options.

activejob/lib/active_job/attributes.rb
```diff
@@ -0,0 +1,84 @@
...
+        @attributes ||=
+          if superclass.respond_to?(:attributes)
+            superclass.attributes.dup
+          else
+            {}
+          end
+      end
+    end
+
+    def set(options = {})
```
bensheldon, line 67: This will also need to be implemented on the class method `.set` too.

dasch: My reading of `ConfiguredJob` is that it just delegates to `#set`:

```ruby
def perform_now(...)
  @job_class.new(...).set(@options).perform_now
end
```

dasch:
> This is on track with what I was imagining. Conceptually, the main deviation is that it treats attribute serialization as separate behaior of the job, serializing them into their own `attributes` key, rather than as sugar/dsl on top of set/serialize/deserialize.
> 
> So I would imagine that:
> 
> ```ruby
> attribute :trace_id
> ```
> 
> ...would be serialized to the `"trace_id"` key in the serialized params.

I actually did have an iteration where it worked like that, but I was thinking that other extensions might also mess around with the serialized params, so it would be more resilient to scope everything into a well-known key.

> I also think the guard about the method already being defined would prevent someone from overloading the accessor and calling `super`.

If that happens _after_ the call to `attribute` then it shouldn't be a problem, right?

> In my mind, things like `priority` or `queue` could ultimately be refactored to use the attribute api we're sketching out.

Perhaps, if the attributes API ends up supporting the type casting features.

> And I wonder how much of this could be done directly by `include ActiveModel::Attributes` and then overloading the `attribute` class method to slightly change/exclude options.

I looked into that, but AM::Attributes includes a _lot_ of code – I think it makes sense for models, but I was thinking that something more lightweight would be more appropriate for AJ.

dasch:
Another thing: I might actually remove the `required:` option to `attribute` again – the less we add, the more optionality we have in the future about e.g. building on top of AM::Attributes. And in all the cases I could think of, it would be trivial and arguably better to implement the presence check in the code that reads the attribute, e.g.

```ruby
module ShardedJob
  extend ActiveSupport::Concern

  included do
    attribute :shard

    before_enqueue do |job|
      raise ArgumentError, "you must set `shard` on sharded jobs!" if job.shard.nil?
    end

    around_perform do |job, block|
      ActiveRecord::Base.connected_to(shard: job.shard, &block)
    end
  end
end  
```

dasch:
> > This is on track with what I was imagining. Conceptually, the main deviation is that it treats attribute serialization as separate behaior of the job, serializing them into their own `attributes` key, rather than as sugar/dsl on top of set/serialize/deserialize.
> > So I would imagine that:
> > ```ruby
> > attribute :trace_id
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ...would be serialized to the `"trace_id"` key in the serialized params.
> 
> I actually did have an iteration where it worked like that, but I was thinking that other extensions might also mess around with the serialized params, so it would be more resilient to scope everything into a well-known key.

Also: serialization became way easier, because `serialize_arguments` Just Works on a Hash, whereas I would need to expose `serialize_argument` from `AJ::Arguments` otherwise (or use `AJ::Serializers.serialize`, but I'm not sure what relation those modules have with each other – I figures custom attributes should work like custom arguments).

activejob/lib/active_job/attributes.rb
bensheldon: Aha, you're right 👍🏻

bensheldon:
> I was thinking that other extensions might also mess around with the serialized params, so it would be more resilient to scope everything into a well-known key.

my opinion here is that that's _their_ problem. Rails is the framework and should be confident in its own implementation. We don't want to break other stuff and I don't think it makes sense to speculatively water-down the core implementation. 

> Also: serialization became way easier...

It would simply be something like this, right?

```
def serialize
  serialized_values.merge attributes.map { |attr| [attr, Arguments.serialize(Array(send(attr)))] }.to_h
end

def deserialize
  attributes.each do |attr|
    send "#{attr}=", Arguments.deserialize(serialized_values[attr.to_s]).first
  end
end
```

> I would need to expose `serialize_argument`

Do it 👍🏻 I wrapped the above with an array to make it work, but again, this is the framework. If it makes sense to make a private method public, then go for it.

dasch:
OK, there's now another iteration, this time without the `attributes` hash key.

bensheldon:
@dasch thank you! This looks good to me. Hopefully it still achieves your initial objectives 😅 thanks for all the back and forth on it. 

I gently tapped the core maintainers who have the actual authority to get their feedback and hopefully acceptance. I think this is a really good improvement for Active Job 👍

activejob/test/cases/attributes_test.rb
```diff
@@ -0,0 +1,62 @@
...
+
+class SubClassedJob < AccountScopedJob
+end
+
+class AttributesTest < ActiveSupport::TestCase
+  setup do
+    JobBuffer.clear
+  end
+
+  test "serializes and deserializes the defined attributes" do
```
bensheldon, line 22: Could we have a test serializing a complex attribute? (array, hash, Active Record / GlobalID-able)?

byroot:
👋 Ben pointed me to this PR. Here's my thoughts.

In general:

  - I think you're onto something here, I'm positive on the general idea of being able to set attributes in addition to pass arguments.
  - Timeline wise, the 8.1 beta is already out, and this is too big of a new feature to be merged now.
  - It's quite a big feature so I'd probably probe a few other core members to get a more definitive opinion.

API and implementation wise:

  - Should these attribute the required or optional?
  - How does this play with concept like "unique" jobs: https://github.com/rails/rails/issues/55613
    - e.g. if we were to add `ActiveJob::Uniqueness`, should these attribute be part of the unicity key? In the case of OTEL spans I'd think not.
  - I got some nitpicks on the implementation, but until we really figure out the above details there isn't a whole lot of point doing rounds of review.

activejob/lib/active_job/attributes.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      super
+    end
+
+    def serialize
+      data = super
+
+      self.class.attributes.each do |name|
+        data[name] = send(name)
```
Mangara: Is this serialized elsewhere? I expected a call to `serialize_argument` here.

dasch: Good catch, that was lost in a refactoring; will push a fix ASAP.

dasch:
@byroot thanks for taking a look!

> * I think you're onto something here, I'm positive on the general idea of being able to set attributes in addition to pass arguments.
> * Timeline wise, the 8.1 beta is already out, and this is too big of a new feature to be merged now.
> * It's quite a big feature so I'd probably probe a few other core members to get a more definitive opinion.

Makes sense :+1:

> * Should these attribute the required or optional?
I initially had a bigger API surface that allowed marking attributes as required, but after using the API in a couple of example cases I found that it wasn't adding a lot of value, and figured that keeping the API bare bones initially would allow for more optionality in the future. That's also why I didn't try to use or emulate ActiveModel::Attributes, which has a _lot_ of functionality that's not really needed here.

If someone wants an attribute to be required, it's quite easy to add a before filter that checks for its presence.

> * How does this play with concept like "unique" jobs: [Active Job serialization optimizations break Sidekiq uniqueness #55613](https://github.com/rails/rails/issues/55613)
>   * e.g. if we were to add `ActiveJob::Uniqueness`, should these attribute be part of the unicity key? In the case of OTEL spans I'd think not.

I think it would be no different from current solutions that mess around with `args` – it depends on the values. So if you're passing e.g. trace context, you need to [manage the lock context](https://github.com/sidekiq/sidekiq/wiki/Ent-Unique-Jobs#lock-context) explicitly.

If we were to add uniqueness to Active Job then it might make sense to extend the `attribute` API to allow including/excluding each attribute from the uniqueness key.

dasch:
@byroot any chance of getting another look on this?

byroot:
Right, now that 8.1 has been released, it's time to consider it. I'll ping a couple people to get their opinion.

activejob/lib/active_job/arguments.rb
```diff
@@ -75,6 +75,25 @@ def serialize_argument(argument) # :nodoc:
       end
     end
 
+    def deserialize_argument(argument)
```
byroot:
```suggestion
    def deserialize_argument(argument) # :nodoc:
```

activejob/lib/active_job/attributes.rb
```diff
@@ -0,0 +1,95 @@
...
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def attribute(name)
+        if respond_to?(name)
+          raise ArgumentError, "Attribute #{name} is already defined"
+        end
+
+        attributes.add(name.to_s)
+        attr_accessor name
```
byroot, line 51: The logic isn't correct here. 

  - `respond_to?` check for the method on `self` so it checks for class methods.
  - `attr_accessor` defines an instance method.

Overall, it's probably better to check `attributes.include?(name.to_s)`, but also probably worth checking how ActiveModel does it and try to be consistent.

activejob/lib/active_job/attributes.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      # The attributes defined on the class, as well as their options.
+      #
+      # Will be inherited to subclasses.
+      def attributes
+        @attributes ||=
+          if superclass.respond_to?(:attributes)
+            superclass.attributes.dup
```
byroot, line 60: Note: This is subject to ordering issues. e.g.

```ruby
class MyJob < ParentJob
  attribute :child
end

ParentJob.attribute :test
```

in the above example `MyJob` won't have a `:test` attribute.

I don't have a solution of the top of my head, but we can maybe do better.

activejob/lib/active_job/attributes.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      super
+    end
+
+    def serialize
+      data = super
+
+      self.class.attributes.each do |name|
+        data[name] = Arguments.serialize_argument(send(name))
+      end
```
byroot, line 82: Might make more sense to keep attributes in a hash, so avoid future name clashing.

byroot: Would be worth looking at ActiveModel `Read` and `Write` modules.

byroot (commented):
Alright, I talked with some other people and It comforted me into thinking we should do this.

But as I mentioned I think the implementation needs a lot of work (if not an entirely different approach). I left some comments in case you want to explore on your own.

But I'd like to find some time for myself to play with this and potentially refactor it heavily, I'm just unsure when I'll get the time+energy, so I don't want to do any false promises.

activejob/lib/active_job/attributes.rb
dasch: Unless there's a simple solution, perhaps we can punt on that?

activejob/lib/active_job/attributes.rb
dasch: I had an implementation of that, but @bensheldon [convinced me](https://github.com/rails/rails/pull/55520#issuecomment-3224547631) to instead write attributes directly on the object. I can change it back, but ideally won't go back and forth any more times then :)

mperham:
I have a similar need. I recently had a Sidekiq user ask why he couldn't profile his Active Job.

Sidekiq allows people to add arbitrary key/values to the job payload, enabling optional functionality. For instance, this code works if using `Sidekiq::Job` but does not work if using `ActiveJob::Base` because `ActiveJob::Core#set` discards anything it does not recognize:

```ruby
MyJob.set(profile: "mytoken").perform_later(123)
```

https://github.com/sidekiq/sidekiq/wiki/Profiling#active-job

byroot:
Yeah, I have a need for it to, I do mean to circle back to that at some point.

bensheldon:
fyi https://github.com/rails/rails/pull/57208

PR #50297, opened 2023-12-07
Fix stale state for composite foreign keys in belongs_to association

iamradioactive:
### Motivation / Background

Fixes #50256 
Fixes #51231

**Context :** 
- A Model having query constraints and a belongs_to association.
- Associated record is loaded once.

**Action :** 
- Update Foreign key for the belongs_to association.
- Call `model.association` to refer to the associated object.

**Current Behaviour:**
- State is not changed and no reload is triggered.
- Association still points to the old record.

**Updated Behaviour:**
- State is updated and reload is triggered.
- Association points to the updated record.

### Detail

This Pull Request updates `stale_state` method in `belongs_to association` to handle composite foreign keys

### Additional information

For Additional context Please refer to #50256 

### ~Dependency~
- ~This PR has a dependency on #50347 and is to be merged after those changes are approved and merged.~
- ~It is an independent bug, however when the stale state is updated correctly it results in loss of information because of the bug to be fixed in the mentioned PR.~
- ~Please refer to this [comment](https://github.com/rails/rails/pull/50297#issuecomment-1852817664) for additional description.~

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55521, opened 2025-08-20
Fix stale state for polymorphic relationship

beauraF:
### Motivation / Background

While working on updating to the latest version of rails `8-0-stable` we had some tests failing. After investigation we managed to track down the culprit to this [PR](https://github.com/rails/rails/pull/50297).

In some cases
https://github.com/rails/rails/blob/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/activerecord/lib/active_record/associations/belongs_to_association.rb#L164 
is now returning `[nil]`. 

Making this condition https://github.com/rails/rails/blob/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/activerecord/lib/active_record/associations/belongs_to_polymorphic_association.rb#L44
being always truthy.

### Detail

Aligning the code with this one https://github.com/rails/rails/blob/41f4e981ded1c0d8c2b3bc5bff5fb2af38da04ed/activerecord/lib/active_record/associations/through_association.rb#L84

### Additional information

Unfortunately, we weren't able to come up with a minimalist reproductible case. It only happens on very specific cases on our side. Therefore we also had an hard time coming up with a good test to add.

### Checklist

### DIFF
--- a/activerecord/test/cases/associations/belongs_to_associations_test.rb
+++ b/activerecord/test/cases/associations/belongs_to_associations_test.rb
@@ -2,6 +2,7 @@
 
 require "cases/helper"
 require "support/deprecated_associations_test_helpers"
+require "models/attachment"
 require "models/developer"
 require "models/project"
 require "models/company"
@@ -1451,6 +1452,20 @@ class BelongsToAssociationsTest < ActiveRecord::TestCase
     end
 
     assert_equal touch_time, car.reload.wheels_owned_at
+  end
+
+  def test_polymorphic_stale_state_handles_nil_foreign_keys_correctly
+    klass = Class.new(ActiveRecord::Base) do
+      self.table_name = "records"
+
+      has_one :attachment, as: :record
+
+      def self.polymorphic_name
+        "Blob"
+      end
+    end
+
+    assert_nothing_raised { Attachment.create!(record: klass.build, record_type: "Document") }
   end
 
   def test_build_with_conditions
--- a/activerecord/lib/active_record/associations/belongs_to_association.rb
+++ b/activerecord/lib/active_record/associations/belongs_to_association.rb
@@ -162,9 +162,15 @@ module ActiveRecord
         end
 
         def stale_state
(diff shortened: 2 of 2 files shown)


MatheusRich:
Could you add tests for this?

beauraF:
@MatheusRich unfortunately not easily.

I mean, `#stale_state` is private. So I would like to test the caller instead (eg: `#stale_target?`). But, I'm not able to reproduce a case where `#stale_target?` return `true` instead of `false` in the condition of this bug. In fact, I believe even if we go through the condition, and it's obviously wrong, it's probably harmless because you end up with a `stale_state` that is `[nil, 'Car']` and compare that. That's probably why it went through.

We only detect it because we have some custom code in our codebase along the line of https://shopify.engineering/changing-polymorphic-type-rails and in a very specific case we started comparing `[nil, 'Patient']` with `[nil, 'PatientBases::Patient']`, when both should be `nil` instead. 

I'm gonna try a bit more. I can also go the `send` route. If you have any ideas or insight don't hesitate.

byroot:
> I'm gonna try a bit more.

Ok, let me know how this goes, because indeed a regression test would much very valuable here.

activerecord/test/cases/associations/belongs_to_associations_test.rb
```diff
@@ -1453,6 +1454,22 @@ def test_polymorphic_with_custom_name_touch_old_belongs_to_model
     assert_equal touch_time, car.reload.wheels_owned_at
   end
 
+  def test_polymorphic_stale_state_handles_nil_foreign_keys_correctly
```
beauraF, line 1457: @byroot I added a test. It's a bit indirect, but that's exactly the situation in which we encounter the issue in our codebase. Tell me if you have a better idea!

byroot: Test seems fine to me. I pushed a small refactoring though.

byroot:
Backported to 8.0 and 7.2

Issue #49898, opened 2023-11-02
`saved_changes` can be empty after persisting changes on a model


aprescott:
### Steps to reproduce

If an `Account` belongs to a `Supplier`, and a `Supplier` has one `Account`, there are cases where calling `account.save` then inspecting `account.saved_changes` leads to an empty result of `{}`, even if changes have actually been made.

Code that attempts to do something based on `saved_changes`, such as record an audit event, will incorrectly act as if no changes were made.

For example, this happens when building a new `supplier` associated with the `account`:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :suppliers, force: true do |t|
    t.string :name
  end

  create_table :accounts, force: true do |t|
    t.integer :supplier_id
    t.string :name
  end
end

class Supplier < ActiveRecord::Base
  has_one :account
end

class Account < ActiveRecord::Base
  belongs_to :supplier, optional: true
end

class BugTest < Minitest::Test
  def test_saved_changes_using_existing_child_with_build
    Account.destroy_all
    Supplier.destroy_all

    account = Account.create!(name: "initial account name")
    account.build_supplier(name: "initial supplier name")
    account.name = "updated account name"

    account.save!

    # Fails, `saved_changes` is unexpectedly empty.
    assert_equal account.saved_changes, { "name" => ["initial account name", "updated account name"] }
  end

  def test_saved_changes_using_existing_child_with_assignment
    Account.destroy_all
    Supplier.destroy_all

    account = Account.create!(name: "initial account name")
    account.supplier = Supplier.new(name: "initial supplier name")
    account.name = "updated account name"

    account.save!

    # Fails, `saved_changes` is unexpectedly empty.
    assert_equal account.saved_changes, { "name" => ["initial account name", "updated account name"] }
  end

  def test_saved_changes_using_new_records_saving_via_child
    Account.destroy_all
    Supplier.destroy_all

    account = Account.new(name: "initial account name")
[...]

PR #45280, opened 2022-06-06
Run transactional callbacks on instances most likely to match DB state

cbothner:
When multiple instances of the same ActiveRecord model are used to persist changes within a transaction, Rails always runs `after_commit` callbacks on the first to save. This results in very unintuitive behavior:

```ruby
Product.transaction do
  Product.find(id).update!(title: "T-Shirt")
  Product.find(id).update!(description: "A cool T-shirt")
end
# When the transaction commits, `after_update_commit` callbacks are fired
# on the first instance, which does not have the new description.

Product.transaction do
  Product.find(id).update!(title: "T-Shirt")
  Product.find(id).destroy!
end
# When the transaction commits, `after_update_commit` callbacks are fired,
# even though the product is destroyed.
```

This PR implements a change to instead fire transactional callbacks on instances with internal state most likely to match the committed database state. 

- In general, run transactional callbacks on the last instance to save a given record within the transaction.
- There are two exceptions:
    - If the record is created within the transaction, then updated by another instance, `after_create_commit` callbacks will be run on the second instance. This is instead of the `after_update_commit` callbacks that would naively be run based on that instance’s state.
    - If the record is destroyed within the transaction, then `after_destroy_commit` callbacks will be fired on the last destroyed instance, even if a stale instance subsequently performed an update (which will have affected 0 rows).

Please see the added test cases for a more detailed statement of behavior.

PR #55533, opened 2025-08-22
Fix saved_changes to reflect all updates in a transaction

### LINKED ISSUE #51046: ActiveModel::Dirty is broken in after_commit
#45280 broke the [expected and documented](https://github.com/rails/rails/blob/main/guides/source/active_record_callbacks.md?plain=1#L598-L608) behaviour by introducing a breaking change in Rails 7.1.
It leads to hard-to-find silent bugs and should be reverted, there's no workaround except disabling it completely with `config.active_record.run_commit_callbacks_on_first_saved_instances_in_transaction = true`.
The problem it was trying to solve is well-documented and easily dealt with `reload`.

### Steps to reproduce
Oversimplified example:
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.integer :comments_count, null: false, default: 0
  end

    t.boolean :hidden, null: false, default: false
    t.integer :post_id
  end
end

  has_many :comments

  def update_comments_count!
    self.update!(comments_count: comments.size)
  end
end

  belongs_to :post
  after_commit :update_posts

  def update_posts
    if saved_change_to_post_id?
      post.update_comments_count!
    end
  end
end

  def test_association_stuff
    post_1 = Post.create!
    post_1.comments.create!

    assert_equal 1, post_1.comments_count

    post_2 = Post.create!
    comment = post_2.comments.build
    Post.transaction do
      comment.save!
      comment.update!(hidden: true)
    end
    assert_equal 1, post_1.comments_count
    assert_equal 1, post_2.comments_count
  end
end
```

### Expected behavior
`saved_change_to_post_id? == true`

### Actual behavior
`saved_change_to_post_id? == false`

### System configuration
**Rails version**: 7.1+
**Ruby version**: 3.2.3

Kazakbai:
### Motivation / Background

This Pull Request fixes #51046, #49898, #45280 by ensuring `saved_changes` reflects all updates made to a record within a single transaction.

### Detail

  When updating different fields of the same record multiple times inside a single transaction, Rails previously reset the dirty fields on each update. As a result, the `saved_changes` hash available in `after_commit` only reflected the *last* update, even if multiple attributes had been changed during the transaction.

  This caused two main inconsistencies:

  * **Multiple instances of the same record:**
    When two separate Active Record instances updated different fields of the same record within the same transaction, only one field change was visible in `saved_changes` inside `after_commit`.

  * **Multiple updates on the same instance:**
    When calling `update!` multiple times on the same record instance, changing different fields each time, `saved_changes` only contained the final update instead of all changes applied during the transaction.

  Rails now **aggregates all changes across updates within a transaction**.
  When the transaction commits, the `saved_changes` hash correctly reflects the full set of modifications performed during the transaction, regardless of whether they were made from different instances or through multiple updates on the same instance.

 * Before (multiple instances)

```ruby
      after_commit -> { puts "saved_changes: #{saved_changes.inspect}" }
    end

    User.transaction do
      User.find(1).update!(name: "Alice")
      User.find(1).update!(email: "alice@example.com")
    end

    # => saved_changes: { "email" => [nil, "alice@example.com"] }
    # "name" change is lost
```

* After (multiple instances)

    ```ruby
    # => saved_changes: {
    #      "name"  => [nil, "Alice"],
    #      "email" => [nil, "alice@example.com"]
    #    }
    # both changes are preserved
    ```

* Before (same instance, multiple updates)

    ```ruby
    User.transaction do
      user = User.find(1)
    
      user.update!(name: "Alice")
      user.update!(email: "alice@example.com")
    end
    
    # => saved_changes: { "email" => [nil, "alice@example.com"] }
    # "name" change is lost
    ```

* After (same instance, multiple updates)

    ```ruby
    # => saved_changes: {
    #      "name"  => [nil, "Alice"],
    #      "email" => [nil, "alice@example.com"]
    #    }
    # both changes are preserved
    ```

* **Removal of `run_commit_callbacks_on_first_saved_instances_in_transaction`**

    ```ruby
    config.active_record.run_commit_callbacks_on_first_saved_instances_in_transaction
    ```

    The configuration option is no longer needed and has been removed. Rails now consistently runs transactional callbacks on the instances with the freshest state, making this parameter redundant.

### Checklist


gap777:
Beautiful! How do we get this across the line?

PR #55535, opened 2025-08-22
Allow a CSRF strategy to handle a cross origin request:

Edouard-chin:
### Motivation / Background

We are trying to flip the switch and enable all controllers to be CSRF protected by default (this is the default since ec4a836919c021c0a5cf9ebeebb4db5e02104a55).

Before we can do this in a safe way, we have a custom CSRF strategy that only logs when a CSRF violation is triggered. This strategy is only called when a bad/no CSRF token is sent, but it doesn't account for when a Cross-Origin request happens, and Rails will raise an error, which we don't want yet.

### Detail

One of our controller didn't emit any log WRT to CSRF violation, so we flipped the "raise" switch for that controller, but ultimately realized it was used for serving javascript assets.

E.g. some html files we don't control were calling this endpoint `<script src="/our-controller"></script>`

This ended up creating an incident on our end.

#### Solution

I'd like to be able to implement the `handle_cross_origin_request` method in our strategy which will get called whenever a cross origin request happens (similar to when a bad CSRF token is sent).

This change is backward compatible, application using the Rails builtin strategies (null_session, exception, reset_session) will continue to emit a log and return a 422 when a Cross Origin is detected.

Custom strategies that don't implement `handle_cross_origin_request` will also continue behaving the same way but will get a deprecation warning.

cc/ @that-jill

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,19 @@
+*   Rails now call `handle_cross_origin_request` when a cross origin request is detected.
+
+    Previously, when a cross origin request was detected, Rails would log an error and
+    raise a `ActionController::InvalidCrossOriginRequest` (which would get turned
+    into a 422 response).
+    Now, a custom strategy can implement the `handle_cross_origin_request` method to perform
+    the logic they want, similar to the existent `handle_unverified_request`.
+
+    Applications using Rails built-in strategies such as `:exception`, `:null_session` or
+    `:reset_sesssion` will get the same behaviour as before.
+
+    Applications using a custom strategy will get the same behaviour as before but
+    a deprecation warning in the event where they haven't implemented `handle_cross_origin_request`.
+
+    *Edouard Chin*
+
 *   URL helpers for engines mounted at the application root handle `SCRIPT_NAME` correctly.
 
     Fixed an issue where `SCRIPT_NAME` is not applied to paths generated for routes in an engine
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -258,7 +258,23 @@ module ActionController # :nodoc:
     end
 
     module ProtectionMethods
+      module CrossOriginRequestStrategy
+        extend self
+
+        def handle_cross_origin_request(controller = nil)
+          controller ||= @controller
+
+          if controller.logger && controller.log_warning_on_csrf_failure
+            controller.logger.warn CROSS_ORIGIN_JAVASCRIPT_WARNING
+          end
+
+          raise ActionController::InvalidCrossOriginRequest, CROSS_ORIGIN_JAVASCRIPT_WARNING
+        end
+      end
+
       class NullSession
+        include CrossOriginRequestStrategy
+
         def initialize(controller)
           @controller = controller
         end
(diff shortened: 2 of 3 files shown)


that-jill (commented):
this approach makes sense to me & the tests accompanying it cover the feature well and serve nicely as documentation for the intent. I double-checked to verify that the new module was included in all pre-existing strategies. Thanks for doing this work @Edouard-chin

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -434,11 +454,25 @@ def unverified_request_warning_message
...
-          raise ActionController::InvalidCrossOriginRequest, CROSS_ORIGIN_JAVASCRIPT_WARNING
+        return unless marked_for_same_origin_verification? && non_xhr_javascript_response?
+
+        protection_strategy = forgery_protection_strategy.new(self)
+
+        if protection_strategy.respond_to?(:handle_cross_origin_request)
+          protection_strategy.handle_cross_origin_request
+        else
+          ActionController.deprecator.warn(<<~MSG)
+            Custom CSRF strategy now need to implement the `handle_cross_origin_request` method.
```
gmcgibbon, line 465:
```suggestion
            Custom CSRF strategies now need to implement the `handle_cross_origin_request` method.
```

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -434,11 +454,25 @@ def unverified_request_warning_message
...
+            Rails now calls the strategy when a Cross Origin request is detected.
+
+            To fix this warning, implement the `handle_cross_origin_request` method in the
+            #{protection_strategy.class.name} strategy class.
+
+            The current behaviour is unchanged, a 422 response will be returned. In future
```
gmcgibbon, line 471:
```suggestion
            If the current behaviour is unchanged, a 422 response will be returned. In the next
```

actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+*   Rails now call `handle_cross_origin_request` when a cross origin request is detected.
```
gmcgibbon, line 1:
```suggestion
*   Rails now calls `handle_cross_origin_request` when a cross origin request is detected.
```

actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+    Previously, when a cross origin request was detected, Rails would log an error and
+    raise a `ActionController::InvalidCrossOriginRequest` (which would get turned
+    into a 422 response).
+    Now, a custom strategy can implement the `handle_cross_origin_request` method to perform
+    the logic they want, similar to the existent `handle_unverified_request`.
+
+    Applications using Rails built-in strategies such as `:exception`, `:null_session` or
+    `:reset_sesssion` will get the same behaviour as before.
```
gmcgibbon, line 10:
```suggestion
    `:reset_sesssion` have the same behaviour as before.
```

actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+    Applications using a custom strategy will get the same behaviour as before but
```
gmcgibbon, line 12:
```suggestion
    Applications using a custom strategy will get the same behaviour as before, with
```

PR #21110, opened 2015-08-03
Add a native JSON data type support in MySQL

kamipo:
As of MySQL 5.7.8, MySQL supports a native JSON data type.

http://dev.mysql.com/doc/refman/5.7/en/json.html

Example:

``` ruby
create_table :json_data_type do |t|
  t.json :settings
end
```

Issue #18629, opened 2015-01-21
Setting invalid JSON terminally breaks ActiveRecord object


lawso017:
As illustrated in this minimal test case against edge:

https://gist.github.com/lawso017/1095d2e8ba868fdd8a6c

attempting to set an AR object's JSON or JSONB attribute to an invalid (non-JSON) value results in a JSON::ParserError.

However, subsequent attempts to reset the value fail with the same ParserError, as the object keeps attempting to parse the previously-submitted invalid input:

````
BugTest#test_setting_invalid_json_terminally_breaks_activerecord_object:
JSON::ParserError: 795: unexpected token at 'foo'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/json-1.8.2/lib/json/common.rb:155:in `parse'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/json-1.8.2/lib/json/common.rb:155:in `parse'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activesupport/lib/active_support/json/decoding.rb:26:in `decode'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activerecord/lib/active_record/connection_adapters/postgresql/oid/json.rb:14:in `type_cast_from_database'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activerecord/lib/active_record/type/mutable.rb:5:in `type_cast_from_user'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activerecord/lib/active_record/attribute.rb:108:in `type_cast'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activerecord/lib/active_record/attribute.rb:42:in `original_value'
    /Users/williamlawson/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/rails-14599a575850/activerecord/lib/active_record/attribute.rb:37:in `value'
[...]

PR #55536, opened 2025-08-22
Report json parse errors during json deserialization

Earlopain:
Since `json` [v2.13.0](https://github.com/ruby/json/releases/tag/v2.13.0), duplicate keys are warned during parsing with the goal of raising in 3.0.0

For rails, I encountered this with json column types that merge into themselves. On rails below 8.1, this shows the warning message from the json gem. On main however, the output is silent since https://github.com/rails/rails/pull/54748

```rb
require "bundler/inline"
gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "~> 8.0.0"
  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.json :payload
  end
end

end

  def test_association_stuff
    post = Post.create!(payload: { foo: "bar" })
    [1, 2].each do |i|
      assert_silent do
        post.update(payload: post.payload.merge(state: i))
      end
    end

    assert_equal({ "foo" => "bar", "state" => 2 }, post.payload)
  end
end
```

If rails switches back to its behaviour on 8.0 with json 3.0, then for duplicate keys `nil` will be returned and no error from the json gem will be shown. I looked at the history on why it rescues and found this:
* https://github.com/rails/rails/commit/e8460f8bbe73537897f2162d1cccda943d8c0f4a added the rescue so that assigning a string to a json column does not leave the object in a broken state
* https://github.com/rails/rails/commit/efaa6e4f79d457c2cdd08cbc56d63bc972a6993c allowed json primitive types, removing the test from the previous commit

So perhaps the rescue can simply be removed.

As for writing a test, I have no idea how. This seems like it should work:

```rb
  def test_json_allow_duplicate_key_false_doesnt_raise_for_symbol_string_keys
    Topic.serialize :content, coder: ActiveRecord::Coders::JSON.new(allow_duplicate_key: false)

    t = Topic.create!(content: { foo: "bar" })
    [1, 2].each do |i|
      t.update(content: t.content.merge(state: i))
    end

    assert_equal({ "foo" => "bar", "state" => 2 }, t.content)
  end
```

But the warning goes away when options are specified: https://github.com/rails/rails/blob/1f9ca6c901bb5fc9b80857808dc1c52f1c5bf0bc/activesupport/lib/active_support/json/encoding.rb#L80-L82

That's also why the PR I mentioned above changed behaviour, `escape` is also an option so that codepath is now taken.

### DIFF
--- a/activerecord/lib/active_record/type/json.rb
+++ b/activerecord/lib/active_record/type/json.rb
@@ -13,7 +13,16 @@ module ActiveRecord
 
       def deserialize(value)
         return value unless value.is_a?(::String)
-        ActiveSupport::JSON.decode(value) rescue nil
+        begin
+          ActiveSupport::JSON.decode(value)
+        rescue JSON::ParserError => e
+          # NOTE: This may hide json with duplicate keys. We don't really want to just ignore it
+          # but it's the best we can do in order to still allow updating columns that somehow already
+          # contain invalid json from some other source.
+          # See https://github.com/rails/rails/pull/55536
+          ActiveSupport.error_reporter.report(e, source: "application.active_record")
+          nil
+        end
       end
 
       JSON_ENCODER = ActiveSupport::JSON::Encoding.json_encoder.new(escape: false)
--- a/activerecord/test/cases/adapters/sqlite3/json_test.rb
+++ b/activerecord/test/cases/adapters/sqlite3/json_test.rb
@@ -27,6 +27,20 @@ class SQLite3JSONTest < ActiveRecord::SQLite3TestCase
     assert_equal '{"list":[]}', klass.new.with_defaults_before_type_cast
   end
 
+  def test_invalid_json_can_be_updated
+    model = klass.create!
+    @connection.execute("UPDATE #{klass.table_name} SET payload = '---'")
+
+    model.reload
+    assert_equal "---", model.payload_before_type_cast
+    assert_error_reported(JSON::ParserError) do
+      assert_nil model.payload
+    end
+
+    model.update(payload: "no longer invalid")
+    assert_equal("no longer invalid", model.payload)
+  end
+
   private
     def column_type
       :json
--- a/activerecord/test/cases/json_attribute_test.rb
+++ b/activerecord/test/cases/json_attribute_test.rb
@@ -25,6 +25,20 @@ class JsonAttributeTest < ActiveRecord::TestCase
     end
   end
 
+  def test_invalid_json_can_be_updated
+    model = klass.create!
+    @connection.execute("UPDATE #{klass.table_name} SET payload = '---'")
+
+    model.reload
+    assert_equal "---", model.payload_before_type_cast
+    assert_error_reported(JSON::ParserError) do
+      assert_nil model.payload
+    end
+
+    model.update(payload: "no longer invalid")
+    assert_equal("no longer invalid", model.payload)
+  end
+
   private
     def column_type
       :string


Earlopain:
Quite a bit of explanation for such a short change. This took me a while to figure out whats going on and I'm not convinced this is at all the right thing to do.

I wonder if doing merging symbol keys is even supposed to happen like that. Rails guides create json columns with symbol keys, so I guess it's officially supported somewhat? https://guides.rubyonrails.org/active_record_postgresql.html#json-and-jsonb
My expecation would be somewhere along the lines of `HashWithIndifferentAccess `, but since it's just a plain one they are only stringified after it got saved to the database, not directly after assignment to a specific key like `post.payload[:foo] = "bar"`.

byroot:
So, as far as I can tell, this `rescue nil` has always been there, since the introduction of the JSON type: https://github.com/rails/rails/pull/21110

It's unclear to me what purpose it ever served.

There is indeed a risk with duplicate keys in the future, as they're rejected on parse not dump (because it would be too expensive), so you could write an unparseable blob in your database.

But generally speaking, I don't think it's a good idea to rescue parse errors here. However some people may discover issue when upgrading (e.g. default value that isn't valid json).

Earlopain:
If you look at the PR you linked, the json type was pretty much just moved from one file to another with the same code: https://github.com/rails/rails/pull/21110/files#diff-0d861d11fe8f49c6e36faeae304d5e431e9016d5b14537bcebc5ce60f33f6078

It moved around a few times before, this should be the source: https://github.com/rails/rails/commit/e8460f8bbe73537897f2162d1cccda943d8c0f4a, https://github.com/rails/rails/commit/efaa6e4f79d457c2cdd08cbc56d63bc972a6993c, I linked them above so at least the initial reason for adding it is no longer valid I'm pretty sure.

As for invalid json defaults, I tried and yeah, I guess that could happen. I created a table via `ActiveRecord::Base.connection.execute("CREATE TABLE posts (id INTEGER PRIMARY KEY AUTOINCREMENT, payload json DEFAULT '---')")` and then it would indeed raise on model creation. But I think this is impossible (or very hard) to achieve when using the schema dsl.

But I can think other ways of getting broken data into such a column. sqlite for example doesn't seem to attempt to validate that the column contains valid json (which is why it has no trouble with a default of `---` with no quotes). Using rails itself it still serializes correctly when doing something like `post.update_columns(payload: "---")` but an external system may not.

> so you could write an unparseable blob in your database.

Rails main seems to handle this just fine by accident. It has something to do with this `to_json` call which supresses the warning: https://github.com/rails/rails/blob/1f9ca6c901bb5fc9b80857808dc1c52f1c5bf0bc/activesupport/lib/active_support/json/encoding.rb#L80-L82 (database columns use `escape: false`, so there are options) and because of that it prunes the duplicates beforehand.

Rails 8.0 warns though the value with duplicate keys never reaches the database, it gets cleaned up somewhere, not sure why/where.

byroot:
> If you look at the PR you linked, the json type was pretty much just moved from one file to another

Damn it. I followed like 5 of these moves and missed the last one -_-. GitHub not rendering fully deleted files by default make it easy to fail a cmd+F...

Overall I think not hinding these potential errors would be better, but the issue this `rescue` "fixed" is legit: https://github.com/rails/rails/issues/18629

If somehow you get invalid JSON in there, it might become tricky to correct course. At the very least we should make sure the error is only raised if you try to access that field. AR changed a lot since then, so that might be the case now.

byroot:
So if you have time for it, I'd suggest adding tests around invalid JSON in columns to see what the behavior would be and so we can eventually adjust it.

Earlopain:
I openend https://github.com/rails/rails/pull/55537 to add tests. Unfortunatly, you can't just update the column once it contains invalid json. Somewhere in checking if the attribute has changed it tries to deserialize the invalid json first.

Only sqlite doesn't validate that it actually contains valid json at the database level (I'm not surprised), but you can also use it backed by a txt column.

So, this PR would break that.

byroot:
> Unfortunatly, you can't just update the column once it contains invalid json.

I assume `update_column` would bypass that, or not even?

Earlopain:
That does indeed work. I saw something about timestamps in the backtrace, so it probably was checking if the timestamp should be updated, and `update_column` just skips that step

byroot:
I need to find some time to look at what cause the error and see if it's avoidable.

byroot:
So that's indeed dirty tracking:

```
    lib/active_record/type/json.rb:16:in 'ActiveRecord::Type::Json#deserialize'
    activemodel/lib/active_model/attribute.rb:183:in 'ActiveModel::Attribute::FromDatabase#type_cast'
    activemodel/lib/active_model/attribute.rb:51:in 'ActiveModel::Attribute#original_value'
    activemodel/lib/active_model/attribute.rb:49:in 'ActiveModel::Attribute#original_value'
    activemodel/lib/active_model/attribute.rb:170:in 'ActiveModel::Attribute#changed_from_assignment?'
    activemodel/lib/active_model/attribute.rb:67:in 'ActiveModel::Attribute#changed?'
```

AR want to check if that attribute changed and somehow deserialises to do that. In theory we wouldn't need to, but whether or not it's easy to implement is another question.

byroot:
Ok, so it's doable not to deserialize to check for changes, the problem is that we kinda purposedly do it so that we don't mis interpret things like identical hashes but with different key orders as different...

```
Failure:
SQLite3JSONTest#test_changes_in_place_ignores_key_order [test/cases/json_shared_test_cases.rb:185]:
Expected #<JSONSharedTestCases::JsonDataType id: 1, payload: {"three" => "four", "one" => "two"}, with_defaults: {"list" => []}, settings: nil> to not be changed?.


bin/test test/cases/json_shared_test_cases.rb:176
```

https://github.com/byroot/rails/commit/1975d207a3de91bac690064c8168c0a33c3deca1

So I'm afraid we're stuck here...

byroot:
> There is indeed a risk with duplicate keys in the future, as they're rejected on parse not dump (because it would be too expensive),

I may have found a way to reject duplicated keys on dump without significantly impacting perf: https://github.com/ruby/json/pull/841

byroot:
I'm afraid short term the best we can really do is report that error to the AR error reporter? Still a bit shit but better than entirely swallowing it?

Earlopain:
Yeah, sure. I never used the rails error reporter, so feel free to tweak. Added the tests from the other PR here now

Earlopain:
Specifically for how I ran into this: with your PR from `json`, it raises early enough that the error doesn't get swallowed somewhere. That doesn't really fix this but smoothes it over enough that I'm happy with it now. Very nice that you managed to make that work.

byroot:
BTW, was there a reason you came to look at this `rescue`? Did you hit it in one of your apps?

Earlopain:
That was entirely my fault. I use strict warnings inspired by what rails does, and that rescue swallowed it. But, I failed to also actually print the warning, so in the end I had strange not-null violations in my tests with no real starting point

PR #55539, opened 2025-08-23
Correct irb inspection class [ci skip]

ttilberg:
Just a quick fix in the irb output example. The example uses `Data.define`, but the irb output used a `Struct`. Verified the correction in my own irb:

```rb
irb(main):015> 10.requests_per_hour
=> #<data ApiBoost::RateLimit requests=10, per=:hour>
```


p8:
Thanks @ttilberg !

PR #55540, opened 2025-08-24
 Skip calling PG::Connection#cancel when using libpq >= 18 with pg < 1.6.0

yahonda:
### Motivation / Background

PostgreSQL 18 changes the length of the cancel request key. With libpq 18 or newer the call to PG::Connection#cancel fails when the pg gem is older than 1.6.0. Rails calls PG::Connection#cancel from cancel_any_running_query, so this mismatch leads to runtime errors.

### Detail
Before this change pull request #55368 checked the PostgreSQL server version inside check_version. That was not correct. The incompatibility is between the pg gem version and the client library version (libpq), as discussed in #55368.

### Additional information
Refer to:
https://github.com/ged/ruby-pg/pull/614
https://github.com/rails/rails/pull/55368
https://github.com/rails/rails/pull/55413

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -127,7 +127,14 @@ module ActiveRecord
           def cancel_any_running_query
             return if @raw_connection.nil? || IDLE_TRANSACTION_STATUSES.include?(@raw_connection.transaction_status)
 
-            @raw_connection.cancel
+            # Skip @raw_connection.cancel (PG::Connection#cancel) when using libpq >= 18 with pg < 1.6.0,
+            # because the pg gem cannot obtain the backend_key in that case.
+            # This method is only called from exec_rollback_db_transaction and exec_restart_db_transaction.
+            # Even without cancel, rollback will still run. However, since any running
+            # query must finish first, the rollback may take longer.
+            if !(PG.library_version >= 18_00_00 && Gem::Version.new(PG::VERSION) < Gem::Version.new("1.6.0"))
+              @raw_connection.cancel
+            end
             @raw_connection.block
           rescue PG::Error
           end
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
```diff
@@ -127,6 +127,10 @@ def set_constraints(deferred, *constraints)
           def cancel_any_running_query
             return if @raw_connection.nil? || IDLE_TRANSACTION_STATUSES.include?(@raw_connection.transaction_status)
 
+            if PG.library_version >= 18_00_00 && Gem::Version.new(PG::VERSION) < Gem::Version.new("1.6.0")
+              raise NotImplementedError, "PG::Connection#cancel is not supported with the combination of pg gem #{PG::VERSION} and PostgreSQL libpq #{PG.library_version / 10000}. Please upgrade pg gem to 1.6.0 or later."
+            end
```
rafaelfranca: Instead of raising, let's just not call `cancel` in this combination of versions. Make sure the reason for this conditional is documented in a comment.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -2,6 +2,9 @@
 
 gem "pg", "~> 1.1"
 require "pg"
+if PG.library_version >= 18_00_00 && Gem::Version.new(PG::VERSION) < Gem::Version.new("1.6.0")
```
rafaelfranca: We can remove this warning now.

rafaelfranca (commented):
We should backport to 8.0

activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
yahonda: Updated as suggested.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
yahonda: Done.

yahonda:
I have updated this pull request as discussed offline yesterday.

PR #55544, opened 2025-08-24
Slightly speed up ActiveJob enqueueing with hashes/keyword arguments

Earlopain:
### Motivation / Background

The splat in the case/when is not great since it allocates an array each time it is encountered.

This avoids the allocation, uses a set for the reserved keys, and avoids string allocations for symbol keys.

<p>

```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "../rails"
  gem "benchmark-ips"
end

require "active_job/railtie"

  config.eager_load = false

  config.logger = Logger.new(File::NULL)
end
Rails.application.initialize!

class MyJob < ActiveJob::Base
  def perform
    puts "performed"
  end
end

job_data = ("a".."z").to_h { [it, it] }
Benchmark.ips do |x|
  x.report("perform_later") do
    MyJob.perform_later(job_data)
  end
end
```

</p>
</details> 

Before:
```
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-linux]
       perform_later     1.541k i/100ms
       perform_later     16.415k (±18.0%) i/s   (60.92 μs/i) -     78.591k in   5.031642s
```

After:
```
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [x86_64-linux]
       perform_later     1.834k i/100ms
       perform_later     19.456k (±20.3%) i/s   (51.40 μs/i) -     91.700k in   5.028421s
```

About 18% faster when the hash has many string keys. 

### Checklist

### DIFF
--- a/activejob/lib/active_job/arguments.rb
+++ b/activejob/lib/active_job/arguments.rb
@@ -64,7 +64,7 @@ module ActiveJob
         RUBY2_KEYWORDS_KEY, RUBY2_KEYWORDS_KEY.to_sym,
         OBJECT_SERIALIZER_KEY, OBJECT_SERIALIZER_KEY.to_sym,
         WITH_INDIFFERENT_ACCESS_KEY, WITH_INDIFFERENT_ACCESS_KEY.to_sym,
-      ]
+      ].to_set
       private_constant :RESERVED_KEYS, :GLOBALID_KEY,
         :SYMBOL_KEYS_KEY, :RUBY2_KEYWORDS_KEY, :WITH_INDIFFERENT_ACCESS_KEY
 
@@ -159,10 +159,12 @@ module ActiveJob
 
       def serialize_hash_key(key)
         case key
-        when *RESERVED_KEYS
+        when RESERVED_KEYS
           raise SerializationError.new("Can't serialize a Hash with reserved key #{key.inspect}")
-        when String, Symbol
-          key.to_s
+        when String
+          key
+        when Symbol
+          key.name
         else
           raise SerializationError.new("Only string and symbol hash keys may be serialized as job arguments, but #{key.inspect} is a #{key.class}")
         end
--- a/activejob/lib/active_job/arguments.rb
+++ b/activejob/lib/active_job/arguments.rb
@@ -89,7 +89,10 @@ module ActiveJob
         when ActiveSupport::HashWithIndifferentAccess
           serialize_indifferent_hash(argument)
         when Hash
-          symbol_keys = argument.each_key.grep(Symbol).map!(&:to_s)
+          symbol_keys = argument.keys
+          symbol_keys.select! { |k| k.is_a?(Symbol) }
+          symbol_keys.map!(&:name)
+
           aj_hash_key = if Hash.ruby2_keywords_hash?(argument)
             RUBY2_KEYWORDS_KEY
           else
--- a/activejob/lib/active_job/serializers/symbol_serializer.rb
+++ b/activejob/lib/active_job/serializers/symbol_serializer.rb
@@ -4,7 +4,7 @@ module ActiveJob
   module Serializers
     class SymbolSerializer < ObjectSerializer # :nodoc:
       def serialize(argument)
-        super("value" => argument.to_s)
+        super("value" => argument.name)
       end
 
       def deserialize(argument)


activejob/lib/active_job/arguments.rb
```diff
@@ -158,11 +158,15 @@ def deserialize_hash(serialized_hash)
       end
 
       def serialize_hash_key(key)
-        case key
-        when *RESERVED_KEYS
+        if RESERVED_KEYS.include?(key)
```
byroot: `Set#===` is an alias for `#include?` so you can do `case key when RESERVED_KEYS`.

Earlopain: Neat, TIL

byroot:
Also there's a bit of an issue with using the test adapter, as it keep appending jobs in an array and that stresses the GC a lot.

Earlopain:
I originally was just benchmarking serialization but thought that was a bit too low-level. If I do just that though, the numbers are even better (almost 50% for the benchmark above, just calling `serialize` on an already instantiated job). Didn't think about how the test adapter works tbh

byroot:
Updated bechmark:

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "."
  gem "benchmark-ips"
end

require "active_job/railtie"

STAGE = ENV["STAGE"]

module ActiveJob
  module QueueAdapters
    class BlackholeAdapter < AbstractAdapter
      def enqueue(job)
        job.serialize
      end
    end
  end
end

  config.eager_load = false
  config.active_job.queue_adapter = :blackhole

  config.logger = Logger.new(File::NULL)
end
Rails.application.initialize!

class MyJob < ActiveJob::Base
  def perform
    puts "performed"
  end
end

job_data = ("a".."z").to_h { [it.to_sym, it] }

Benchmark.ips do |x|
  x.report(STAGE || "perform_later") do
    MyJob.perform_later(job_data)
  end
  x.compare!(order: :baseline)
  if STAGE
    x.save!("/tmp/bench-aj")
  end
end
```

```
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after     8.226k i/100ms
               after     84.936k (± 1.8%) i/s   (11.77 μs/i) -    427.752k in   5.037809s

Comparison:
              before:    48629.3 i/s
               after:    84936.3 i/s - 1.75x  faster
```

byroot:
Looks like there a bunch other low hanging fruits in there, would be worth doing a pass.

e.g. the code to handle symbol keys is a bit wasteful:

```ruby
          symbol_keys = argument.keys
          symbol_keys.select! { |k| k.is_a?(Symbol) }
          symbol_keys.map!(&:name)

          aj_hash_key = if Hash.ruby2_keywords_hash?(argument)
            RUBY2_KEYWORDS_KEY
          else
            SYMBOL_KEYS_KEY
          end
          result = serialize_hash(argument)
          result[aj_hash_key] = symbol_keys
          result
```

If we assume most hashes either have all their keys as symbol or none of them, rather than to supply a list we could pass a boolean. Would be faster and more compact. However there are concerns over backward/forward compact, so it need to be done carefully with a configuration flag, etc.

activejob/lib/active_job/arguments.rb
```diff
@@ -64,7 +64,7 @@ def deserialize(arguments)
         RUBY2_KEYWORDS_KEY, RUBY2_KEYWORDS_KEY.to_sym,
         OBJECT_SERIALIZER_KEY, OBJECT_SERIALIZER_KEY.to_sym,
         WITH_INDIFFERENT_ACCESS_KEY, WITH_INDIFFERENT_ACCESS_KEY.to_sym,
-      ]
+      ].to_set
```
byroot, line 67: Also wondering if we're not better to just check for `start_with?("_aj_")`. It's both cheap and ensure that namespace is free for future keys.

Earlopain:
Yup, I noticed that as well. I had a pretty stupid idea for this, assuming all keys are either one or the other is much simpler :sweat_smile:. In practise also often the case I'd say with keyword arguments but never so many as in the benchmark. This was more for when you pass some json over where many keys aren't that uncommon.

byroot:
> but never so many as in the benchmark.

Yeah, the current benchmark may not be super representative. Still, saving having a copy of all the keys most of the time is a good idea.

If you wish to try that I can hold off.

Earlopain:
I can take a look, sure. I'll put something up, I should be able to do so.

Earlopain:
I have a branch here: https://github.com/Earlopain/rails/tree/aj-enqueue-all-symbol-keys

But the numbers aren't really what I would except. Barely any difference or worse, not sure if that is load difference from my system though. It was very hard to measure. The serialization is smaller, but I'd except some improvement in runtime as well (didn't test deserializion)

Feel free to check it out and tweak or just start fresh, your call. I changed the format and content of some of the keys but it should all be backwards compatible. I was testing with this, which is a more realistic benchmark:

<p>

```rb
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "../rails"
  gem "benchmark-ips"
end

require "active_job/railtie"

module ActiveJob
  module QueueAdapters
    class BlackholeAdapter < AbstractAdapter
      def enqueue(job)
        job.serialize
      end
    end
  end
end

  config.load_defaults 8.1
  config.eager_load = false
  config.active_job.queue_adapter = :blackhole

  config.logger = Logger.new(File::NULL)
end
Rails.application.initialize!

class MyJob < ActiveJob::Base
  def perform
    puts "performed"
  end
end

keys = ["first_job_argument", "second_job_argument", "and_one_more"]
symbols = keys.to_h { [it.to_sym, it] }
strings = keys.to_h { [it, it] }
mixed = strings.transform_keys.with_index { |key, i| i % 2 == 0 ? key : key.to_sym }


prefix = "new-"

Benchmark.ips do |x|
  x.report("#{prefix}symbols") do
    MyJob.perform_later(symbols)
  end

  x.report("#{prefix}strings") do
    MyJob.perform_later(strings)
  end

  x.report("#{prefix}mixed") do
    MyJob.perform_later(mixed)
  end
  x.save!("/tmp/bench-aj")

  x.compare!
end
```

</p>
</details>

byroot:
What I had in mind is something like this:

```ruby
Benchmark.ips do |x|
  x.report(STAGE || "perform_later") do
    MyJob.perform_later(foo: 1, bar: 2, baz: { egg: 1, spam: 3})
  end
  x.compare!(order: :baseline)
  if STAGE
    x.save!("/tmp/bench-aj")
  end
end
```

That list of argument seem somewhat typical in my opinion.

With a not too complicated diff https://github.com/rails/rails/compare/main...byroot:rails:aj-hash-args (just on the serialization side)

I get:

```
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after    13.025k i/100ms
               after    133.386k (± 1.0%) i/s    (7.50 μs/i) -    677.300k in   5.078282s

Comparison:
              before:   121386.3 i/s
               after:   133386.2 i/s - 1.10x  faster
```

Which isn't a huge win, but might be worth it?

PR #55548, opened 2025-08-24
Remove the ability to set email_address value via query parameters

### LINKED ISSUE #55168: Unclear usage of `params[:email_address]` in generated `sessions/new.html.erb` view
Hi everyone,

after scaffolding authentication with `rails g authentication`, I've noticed that the form field for the email address in the `sessions/new.html.erb` view looks like this:

```rb
<%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address", value: params[:email_address] %>
```

I get that the intended usage of `value: params[:email_address]` is pre-populating the form field, but I couldn't get it working (as it is right after generation) because `SessionsController` redirects on authentication failure, and of course `params` is empty when `new` is rendered.
I think that this should be a use case for `session` or `flash`.

Is this the intended behavior? Is it meant to take other scenarios into consideration?

Thanks!

### System configuration

**Rails version**: 8.0.2

**Ruby version**: 3.4.4

**UPDATE 20/08/2025**

I think that the above is true also for the `passwords/new.html.erb` view:

```rb
<%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address", value: params[:email_address] %>
```

(same initial value for the field, but nothing in the scaffold that links to the page with an `email_address` parameter).

codergeek121:
The authentication generator form templates allow setting the email_address form field value with query parameters. This is not used in any existing functionality or test.

Fixes #55168

### DIFF
--- a/railties/lib/rails/generators/erb/authentication/templates/app/views/sessions/new.html.erb
+++ b/railties/lib/rails/generators/erb/authentication/templates/app/views/sessions/new.html.erb
@@ -2,7 +2,7 @@
 <%%= tag.div(flash[:notice], style: "color:green") if flash[:notice] %>
 
 <%%= form_with url: session_path do |form| %>
-  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address", value: params[:email_address] %><br>
+  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address" %><br>
   <%%= form.password_field :password, required: true, autocomplete: "current-password", placeholder: "Enter your password", maxlength: 72 %><br>
   <%%= form.submit "Sign in" %>
 <%% end %>
--- a/railties/lib/rails/generators/erb/authentication/templates/app/views/passwords/new.html.erb
+++ b/railties/lib/rails/generators/erb/authentication/templates/app/views/passwords/new.html.erb
@@ -3,6 +3,6 @@
 <%%= tag.div(flash[:alert], style: "color:red") if flash[:alert] %>
 
 <%%= form_with url: passwords_path do |form| %>
-  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address", value: params[:email_address] %><br>
+  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address" %><br>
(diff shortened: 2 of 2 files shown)


railties/lib/rails/generators/erb/authentication/templates/app/views/sessions/new.html.erb
```diff
@@ -2,7 +2,7 @@
 <%%= tag.div(flash[:notice], style: "color:green") if flash[:notice] %>
 
 <%%= form_with url: session_path do |form| %>
-  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address", value: params[:email_address] %><br>
+  <%%= form.email_field :email_address, required: true, autofocus: true, autocomplete: "username", placeholder: "Enter your email address" %><br>
```
MatheusRich, line 5: I think it would be preferable to make this work by forwarding the email param when `authenticate_by` fails

codergeek121: Ah,  I see 🤔
I basically thought the `alert` message `"Try another **email address** or password."`  hinted more to the fact that this is a mistake and not actually intended. But I could adapt it ofc.

I think there are 3 possible solutions:

1. Save the entered email in the flash hash to avoid logging or having the email in the query params
2. Redirect to a path with the email_address query param set
3. Drop the params support completely (basically this PR)

For 1) and 2) I'd also adapt the `alert` message, I  find it confusing to tell the user that the email may be wrong but showing it in the input. What I don't like about 2) would be the visible email address in the browser bar, but it's also not the end of the world.

Imo though, keeping the feature as simple as possible with the least amount of surprises would be solution 3). Since this is only generated code, adapting the default to 1) and 2) would be relatively straightforward.

Thank you for taking look 👍

MatheusRich: I think "Try another email address or password." is okay. We can't tell which one is wrong.

callmesangio: Hi, original reporter of #55168 here.
FWIW, I'm strongly +1 about option 3, since it's the one that makes the least assumptions. I agree that it can be adapted easily to option 1 or 2, if the user wants to do so, but it's a great default on its own.

PR #55549, opened 2025-08-25
Make transaction isolation work smoothly with transactional tests

kirs:
When transactional fixtures are enabled, any code that specifies transaction isolation breaks in tests.

Here's an example:

```ruby
class ExampleTest < ActiveRecord::TestCase
  self.use_transactional_tests = true

    self.table_name = "tags"
  end

  test "omglol" do
    Tag.transaction(isolation: :read_committed) do
      Tag.create!
    end
  end
end

# fails with:
ActiveRecord::TransactionIsolationError: cannot set transaction isolation in a nested transaction
    lib/active_record/connection_adapters/abstract/transaction.rb:418:in 'ActiveRecord::ConnectionAdapters::SavepointTransaction#initialize'
    lib/active_record/connection_adapters/abstract/transaction.rb:538:in 'Class#new'
    lib/active_record/connection_adapters/abstract/transaction.rb:538:in 'block in ActiveRecord::ConnectionAdapters::TransactionManager#begin_transaction'
    /Users/kirs/src/github.com/rails/rails/activesupport/lib/active_support/concurrency/load_interlock_aware_monitor.rb:23:in 'Thread.handle_interrupt'
    /Users/kirs/src/github.com/rails/rails/activesupport/lib/active_support/concurrency/load_interlock_aware_monitor.rb:23:in 'block in ActiveSupport::Concurrency::LoadInterlockAwareMonitorMixin#synchronize'
```

This error makes sense because the transaction is already open by the test framework and you cannot change isolation in the middle of transaction.

The workaround is having to check whether the test transaction has been open in every code path:

```ruby
isolation = if Tag.lease_connection.transaction.open?
  nil
else
  :read_committed
end
Tag.transaction(isolation: isolation) do
  ...
end
```

Similar to how transactional fixtures [mock the replica connection](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/test_fixtures.rb#L239-L262) so that you can smoothly test code that uses DB replicas, I think we can provide better DX and let you pass `isolation` in tests when `use_transactional_tests` is enabled.

This is achieved by accounting for the fact that in tests, the first transaction always comes from transactional fixtures, and mocking that.

I'm open to better ideas how to make this work, but I strongly believe that making it work seamlessly in transactional tests is a good DX.

@byroot @rafaelfranca @matthewd

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -353,6 +353,22 @@ module ActiveRecord
       # isolation level.
       #  :args: (requires_new: nil, isolation: nil, &block)
       def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
+        # If we're running inside the single, non-joinable transaction that
+        # ActiveRecord::TestFixtures starts around each example (depth == 1),
+        # an `isolation:` hint must be validated then ignored so that the
+        # adapter isn't asked to change the isolation level mid-transaction.
+        if isolation && !requires_new && open_transactions == 1 && !current_transaction.joinable?
+          iso = isolation.to_sym
+
+          unless transaction_isolation_levels.include?(iso)
+            raise ActiveRecord::TransactionIsolationError,
+                  "invalid transaction isolation level: #{iso.inspect}"
+          end
+
+          current_transaction.isolation = iso
+          isolation = nil
+        end
+
         if !requires_new && current_transaction.joinable?
           if isolation && current_transaction.isolation != isolation
             raise ActiveRecord::TransactionIsolationError, "cannot set isolation when joining a transaction"
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -436,13 +436,16 @@ module ActiveRecord
         end
       end
 
+      TRANSACTION_ISOLATION_LEVELS = {
+        read_uncommitted: "READ UNCOMMITTED",
+        read_committed:   "READ COMMITTED",
+        repeatable_read:  "REPEATABLE READ",
+        serializable:     "SERIALIZABLE"
+      }.freeze
+      private_constant :TRANSACTION_ISOLATION_LEVELS
+
       def transaction_isolation_levels
-        {
-          read_uncommitted: "READ UNCOMMITTED",
-          read_committed:   "READ COMMITTED",
-          repeatable_read:  "REPEATABLE READ",
-          serializable:     "SERIALIZABLE"
-        }
+        TRANSACTION_ISOLATION_LEVELS
       end
 
       # Begins the transaction with the isolation level set. Raises an error by
--- a/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
@@ -124,6 +124,7 @@ module ActiveRecord
       def after_commit; yield; end
       def after_rollback; end
       def user_transaction; ActiveRecord::Transaction::NULL_TRANSACTION; end
+      def isolation=(_); end
     end
 
     class Transaction # :nodoc:
@@ -154,6 +155,10 @@ module ActiveRecord
       # Returns the isolation level if it was explicitly set, nil otherwise
       def isolation
         @isolation_level
+      end
+
+      def isolation=(isolation) # :nodoc:
+        @isolation_level = isolation
       end
 
       def initialize(connection, isolation: nil, joinable: true, run_commit_callbacks: false)
@@ -424,6 +429,10 @@ module ActiveRecord
       # Delegates to parent transaction's isolation level
       def isolation
         @parent_transaction.isolation
+      end
+
+      def isolation=(isolation) # :nodoc:
+        @parent_transaction.isolation = isolation
       end
 
       def materialize!
--- a/activerecord/test/cases/transaction_isolation_test.rb
+++ b/activerecord/test/cases/transaction_isolation_test.rb
@@ -302,3 +302,55 @@ class TransactionIsolationTest < ActiveRecord::TestCase
       end
   end
 end
+
+class TransactionIsolationWithTransactionalTestsTest < ActiveRecord::TestCase
+  if ActiveRecord::Base.lease_connection.supports_transaction_isolation? && !current_adapter?(:SQLite3Adapter)
+    class Tag < ActiveRecord::Base
+      self.table_name = "tags"
+    end
+
+    test "starting a transaction with isolation does not raise an error" do
+      assert_nothing_raised do
+        Tag.transaction(isolation: :read_committed) do
+          Tag.create!
+        end
+      end
+    end
+
+    test "starting a transaction with isolation sets the isolation level" do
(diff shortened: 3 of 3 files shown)


byroot:
I agree transactional fixtures should do what is necessary to not break code using isolation level, however the implementation isn't satisfactory. The `.extend` in particular isn't suitable.

matthewd:
My main concern is that if code cares enough to be specifying a particular isolation level, then it seems possible that we're going to surprise people by not actually doing what they asked for inside tests -- not to mention the possible dangers from us potentially not validating the requested `isolation:` value.

byroot:
> if code cares enough to be specifying a particular isolation level, then it seems possible that we're going to surprise people by not actually doing what they asked for inside tests

I don't think so. Isolation level only really makes a difference when concurrency is involved, so not really a concern for a test process. 

> not to mention the possible dangers from us potentially not validating the requested isolation: value.

That yes, it should validate the level somehow.

kirs:
Thanks for feedback!

I have reworked it to move the check into `ConnectionAdapters::DatabaseStatements`. It's much cleaner, although you could argue that the test concerns are somewhat leaking into production path.

And I agree with Jean, isolation only makes a difference under concurrency.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -359,6 +359,22 @@ def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
...
+          # adapter isn't asked to change the isolation level mid-transaction.
+          if isolation && open_transactions == 1 && !current_transaction.joinable?
+            iso = isolation.to_sym
+
+            unless transaction_isolation_levels.include?(iso)
+              raise ActiveRecord::TransactionIsolationError,
+                    "invalid transaction isolation level: #{iso.inspect}"
+            end
+
+            current_transaction.instance_variable_set(:@isolation_level, iso)
```
kirs: Happy to introduce `def _set_isolation` on a connection if you think a public method is preferable.

byroot: A `:nodoc:` public method is better than `instance_variable_set` yes.

byroot:
> I have reworked it

I'll try to find some time to have a look.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -359,6 +359,22 @@ def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
           end
           yield current_transaction.user_transaction
         else
+          # If we're running inside the single, non-joinable transaction that
+          # ActiveRecord::TestFixtures starts around each example (depth == 1),
+          # an `isolation:` hint must be validated then ignored so that the
+          # adapter isn't asked to change the isolation level mid-transaction.
+          if isolation && open_transactions == 1 && !current_transaction.joinable?
```
byroot: Maybe it's time we add some more explicit tagging on the test fixtures transaction? I never felt like checking for `joinable? == false` was very clear.

@matthewd what would you think about having a (private API) method such as `current_transaction.fixtures?` (I can't come up with a better name, but there must be one).

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -359,6 +359,22 @@ def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
...
+            iso = isolation.to_sym
+
+            unless transaction_isolation_levels.include?(iso)
+              raise ActiveRecord::TransactionIsolationError,
+                    "invalid transaction isolation level: #{iso.inspect}"
+            end
```
byroot: We probably should just move that in `#isolation=`

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -156,6 +156,10 @@ def isolation
         @isolation_level
       end
 
+      def isolation=(isolation) # :nodoc:
+        @isolation_level = isolation
```
byroot, line 161: We should probably assert that this transaction isn't joinable, because this API used with real transactions would be pretty bad.

From the DB point of view this is a NOOP, setting it does nothing except change the return value of `#isolation`, however, in case of a query retry, we might reconnect and reopen transactions, which would cause the real isolation level to change. That would be super surprising.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
byroot: Or perhaps, we could define a `Transaction` subclass? And check the type?

matthewd: Yeah... `joinable` isn't public API, so technically (despite feeling gross every time I see it) it is valid for us to treat `!joinable?` == `outer_fixture_transaction_we_should_sometimes_pretend_isnt_there?`... but it seems nice to have a clearer direct representation of what we're asking. (`joinable?` was an elegant abstraction originally, when the fixture side was imposing the logic, and just wanted to request one specific behaviour from the transaction... but as soon as started using that in reverse to infer the transaction's purpose, it became overloaded.)

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -359,6 +359,22 @@ def transaction(requires_new: nil, isolation: nil, joinable: true, &block)
...
+
+            current_transaction.isolation = iso
```
matthewd: Is this better placed inside `within_new_transaction`? That already has light `isolation` handling, and it feels slightly less bad layer-breaking to keep it in TransactionManager, maybe?

Should we also restore the previous isolation at the end of the block? Does leaving a false residual isolation level potentially create surprises when interacting with other later transaction blocks?

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -333,6 +346,10 @@ def commit_records
       def full_rollback?; true; end
       def joinable?; @joinable; end
 
+      def fixture_transaction?
```
kirs: @matthewd @byroot I've opted for a method like this rather than a subclass because that seemed less complex than making it fit under `Transaction` / `SavepointTransaction` / `RealTransaction`.

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
kirs: Addressed this above.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
kirs: Great points, addressed both 🙌

kirs:
@matthewd @byroot would appreciate another pass on this one 👀. Thank you!

byroot:
Alright, I did some minor cleanup, this seem good to go now.

I still think we should bite the bullet and have an explicit `fixtures_transaction?` kind of predicate, but it can wait.

kirs:
@byroot something must have screwed up in git because on my branch, https://github.com/rails/rails/commit/4dfe997d62f52d335f87d33f3f470c0f105fb405 is the latest commit which did include `def fixture_transaction?`

byroot:
Yeah, it's my fault 🤦 No big deal.

PR #55553, opened 2025-08-25
Make local cache middleware updatable

gmcgibbon:
### Motivation / Background

This Pull Request has been created because it is currently impossible to update local cache middleware for Rails.cache once the application is booted. If the cache client needs to change for any reason post-boot, mounted middleware will point to the wrong cache key and redundant cache operations will hit the datastore.

### Detail

This Pull Request changes `ActiveSupport::Cache::Strategy::LocalCache::Middleware` to keep reference of the cache instead of the generated cache key (which uses the cache's object ID), and then makes the cache reader an accessor. While I was in the area, I added public local cache resetting methods so that tests and middleware don't need to reach into internals.

### Additional information

Without this patch, there's no way to cleanly update local cache middleware if `Rails.cache` changes. This is a niche use-case, but for cases like forked parallel test workers, it is useful.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Make the cache of `ActiveSupport::Cache::Strategy::LocalCache::Middleware` updatable.
+
+    If the cache client at `Rails.cache` of a booted application changes, the corresponding
+    mounted middleware needs to update in order for request-local caches to be setup properly.
+    Otherwise, redundant cache operations will erroneously hit the datastore.
+
+    *Gannon McGibbon*
+
 *   Add `assert_events_reported` test helper for `ActiveSupport::EventReporter`.
 
     This new assertion allows testing multiple events in a single block, regardless of order:
--- a/activesupport/lib/active_support/cache/strategy/local_cache.rb
+++ b/activesupport/lib/active_support/cache/strategy/local_cache.rb
@@ -68,12 +68,25 @@ module ActiveSupport
           use_temporary_local_cache(LocalStore.new, &block)
         end
 
+        # Set a new local cache.
+        def new_local_cache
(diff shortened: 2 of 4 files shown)


activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Make `ActiveSupport::Cache::Strategy::LocalCache::Middleware` updatable.
```
eileencodes: I think the changelog entry could be a bit more descriptive - because we're really making it possible to update the LocalCache rather than making the Middleware updatable. Meaning that I wouldn't know what is updatable by reading this changelog entry.

eileencodes (commented):
Looks good to me, just a comment on the changelog entry. Also there's a conflict with it.

activesupport/lib/active_support/cache/strategy/local_cache.rb
```diff
@@ -68,12 +68,25 @@ def with_local_cache(&block)
           use_temporary_local_cache(LocalStore.new, &block)
         end
 
+        # Set a new local cache.
+        def new_local_cache
+          LocalCacheRegistry.set_cache_for(local_cache_key, LocalStore.new)
+        end
+
+        # Unset the current local cache.
+        def clear_local_cache
```
byroot: `clear` is perhaps a bit ambiguous here. It can easily be understood as "empty the cache but keep it active" given that's what `Rails.cache.clear` does.

Maybe `unset` or `disable` ?

gmcgibbon: I originally had unset. I'll use that

activesupport/CHANGELOG.md
gmcgibbon: Alright, I'll clarify.

PR #55554, opened 2025-08-25
Clarifies action of `db:migate:reset` in Rails 8.0 release notes [ci skip]

### LINKED ISSUE #55499: Rails guide for v8 release notice  provides deceptively dangerous recommendation
### Steps to reproduce

https://guides.rubyonrails.org/8_0_release_notes.html#active-record-notable-changes

### Actual behavior

Section 8.3 of the v8 release notes suggests using `rails db:migrate:reset` for the "previous behavior"

This does not provide the previous behavior (which was to always run migrations) but instead drops all of the databases and running *all* migrations from scratch. This is dangerous since it could easily result in lost data.

**Rails version**: 

8.0.2.1

marcrohloff:
### Motivation / Background

The `db:migrate:reset` rake task is described in the Rails 8.0 release notes as `the previous behavior`. This is not entirely true as it destructively recreates the databases first and then runs the previous behavior. This can lead to unexpected data loss. There are several recent issues that show this is causing confusion

### Change

This PR expands on the release notes to clarify that 

closes rails/rails#55499

### DIFF
--- a/guides/source/8_0_release_notes.md
+++ b/guides/source/8_0_release_notes.md
@@ -186,7 +186,10 @@ Please refer to the [Changelog][active-record] for detailed changes.
 ### Notable changes
 
 *   Running `db:migrate` on a fresh database now loads the schema before running
-    migrations. (The previous behavior is available as `db:migrate:reset`)
+    migrations. Subsequent calls will run pending migrations.
+    (If you need the previous behavior of running migrations from scratch instead of loading the 
+    schema file, this can be done by running `db:migrate:reset` which 
+    _will drop and recreate the database before running migrations_)
 
 Active Storage
 --------------


guides/source/8_0_release_notes.md
```diff
@@ -186,7 +186,10 @@ Please refer to the [Changelog][active-record] for detailed changes.
 ### Notable changes
 
 *   Running `db:migrate` on a fresh database now loads the schema before running
-    migrations. (The previous behavior is available as `db:migrate:reset`)
+    migrations. Subsequent calls will run pending migrations.
+    (If you need the previous behavior of running migrations from scratch instead of loading the 
+    schema file, this can be done by running `db:migrate:reset` which 
+    will drop and recreate the database before running migrations)
```
MatheusRich: Probably worth some attention
```suggestion
    schema file, this can be done by running `db:migrate:reset` which 
    _will drop and recreate the database before running migrations_)
```

PR #55555, opened 2025-08-25
Rate limiting calls `cache_key` on `by:` if the object responds to it

daniel-sabourin:
### Motivation / Background

This Pull Request has been created because the default configuration of rate limiting by the `remote_ip` doesn't quite work for our use case. There are times where we have an authenticated user that we want to apply the rate limit to, but we don't want to restrict the entire IP incase there's multiple users on the same home/work network. By applying the rate limit to the user itself, each user has their own rate limit.

### Detail

This Pull Request changes the `rate_limit` `to:` argument to allow any kind of object to be returned. If the object responds to `cache_key` (like ActiveRecord objects do), the rate limit will call that method when constructing the cache key. 

```ruby
class CommentsController < ActionController::API
  # Cache key in the store would be `rate-limit:comments:user/1`
  rate_limit to: 2, within: 2.seconds, by: { current_user }

  private

  # Substitute however you load your user object (potentially using Devise)
  def current_user
    User.find(1)
  end
end
```

### Additional information

One might argue that it's not a huge deal to just call the `cache_key` method directly in `to:`, but to me this is a sensible enough default to warrant a change instead of letting the setting propagate the codebase if you have multiple user-centric rate limits.

```ruby
# how it's done now
class CommentsController < ActionController::API
  # Cache key in the store would be `rate-limit:comments:user/1`
  rate_limit to: 2, within: 2.seconds, by: { current_user.cache_key }
end
```

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Rate limiting calls `cache_key` on `by:` if the object responds to it.
+
+    ```ruby
+    class CommentsController < ApplicationController
+      # Cache key in the store would be `rate-limit:comments:user/1`
+      rate_limit to: 2, within: 2.seconds, by: -> { current_user }
+    end
+    ```
+
+    *Daniel Sabourin*
+
 *   Add a configuration for `ActionDispatch::ExceptionWrapper.silent_exceptions` at `config.action_dispatch.silent_exceptions`.
 
     Exceptions on this list do not fall back to framework-level backtraces when there is no application backtrace.
--- a/actionpack/lib/action_controller/metal/rate_limiting.rb
+++ b/actionpack/lib/action_controller/metal/rate_limiting.rb
@@ -20,7 +20,9 @@ module ActionController # :nodoc:
       # Rate limits are by default unique to the ip address making the request, but
       # you can provide your own identity function by passing a callable in the `by:`
       # parameter. It's evaluated within the context of the controller processing the
-      # request.
+      # request. If the callable returns an object that responds to `cache_key`, its
+      # value will be used as the cache key. This is useful if you want the rate limit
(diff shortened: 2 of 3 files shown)


actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+*   Allow `by:` to return an object. If that object responds to `cache_key`,
+    that `cache_key` is used instead of the object in memory, which you'd rarely want
+
+    ```ruby
+    class CommentsController < ActionController::API
+      # Cache key in the store would be `rate-limit:comments:user/1`
+      rate_limit to: 2, within: 2.seconds, by: { current_user }
```
MatheusRich:
```suggestion
      rate_limit to: 2, within: 2.seconds, by: -> { current_user }
```

MatheusRich:
@daniel-sabourin wouldn't `-> { Current.user.cache_key }` be sufficient here? Or are you proposing a general strategy to use `cache_key` whenever it's available?

daniel-sabourin:
@MatheusRich - Exactly. Using `-> { Current.user.cache_key }` would be sufficient/equivalent. I'm proposing that it makes sense to have one of 3 ways of using `by:`
- If you want to rate limit by incoming IP address, that's the default and no configuration is required.
- If you want to rate limit by an ActiveRecord object (or anything that implemented `cache_key`), you can pass it in like `by: -> { resource }`
- If you want to do anything else custom, you return a string `by: -> { "foobar" }`

As it stands right now, returning an object has unexpected results. The cache key ends up being `rate-limit:comments:#<User:0x00000001337545e8>`

MatheusRich:
I'd rename the PR and commit to make that more clear: rate limiting will use `cache_key` if the object responds to that.

PR #55557, opened 2025-08-26
Documentation small change: Add clarification to Concern's save location

manuel1618:
As I was going through the guide I wondered how concerns would be used on multiple models without hurting DRY. 

The guide tells you to create the concern in a model specific folder called products and the sub chapter ends with saying concerns can be used by multiple models.  

I added a short explanation where to put the concerns in this case.


p8:
Related: https://github.com/rails/rails/pull/55304

PR #55562, opened 2025-08-27
Fix words ending in "aves" singular

mochetts:
### Motivation / Background

This Pull Request has been created because "caves" gets singularized to "cafe", which is wrong.

Same happens with "waves", it translates to "wafe".

### Detail

This Pull Request changes the inflextions singular regex pattern for words ending in "ves", so that we exclude "caves" and "waves" from it.  Effectively translating "caves" to "cave" and "waves" to "wave".

### Additional information

Every word that has an "a" before the "ves" shouldn't be singularized using the rule. Word examples:
Caves → Cave
Waves → Wave
Paves → Pave (as in "to pave a road")
Saves → Save (as in "to save a game" or "a goalie's save")
Laves → Lave (an archaic word for "to wash")
Naves → Nave
Braves → Brave (can be an adjective or a noun for a courageous person)
Slaves → Slave
Staves → Stave (depending on the meaning)
Graves → Grave

I had to change the test assertion that checks that the plural of "safe" is "saves". Doesn't seem right. If we're talking about a "safe" being the metal box or cabinet, the plural is "safes".

### Checklist


skipkayhil:
Thank you for the pull request, but as indicated by the comment at the top of the file these inflections are frozen.

https://github.com/rails/rails/blob/f911e3dba94e52dd598f06285b387c8a4df3df32/activesupport/lib/active_support/inflections.rb#L5-L10

Feel free to add additional inflections to your own application

PR #55565, opened 2025-08-28
Creates `parallel_worker_id` helper for running parallel tests

Schwad:
Right now the only opportunity to access the parallel worker id is to grab it in

```
parallelize_setup do |worker_id|
```

Where it must be assigned to a config by hand. This PR makes a helper `parallel_worker_id` available throughout. It is `nil` if not in parallel.

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Create `parallel_worker_id` helper for running parallel tests. This allows users to
+    know which worker they are currently running in.
+
+    *Nick Schwaderer*
+    
 *   Make the cache of `ActiveSupport::Cache::Strategy::LocalCache::Middleware` updatable.
 
     If the cache client at `Rails.cache` of a booted application changes, the corresponding
--- a/activesupport/lib/active_support/test_case.rb
+++ b/activesupport/lib/active_support/test_case.rb
@@ -23,7 +23,22 @@ module ActiveSupport
   class TestCase < ::Minitest::Test
     Assertion = Minitest::Assertion
 
+    # Class variable to store the parallel worker ID
+    @@parallel_worker_id = nil
+
     class << self
+      # Returns the current parallel worker ID if tests are running in parallel,
+      # nil otherwise.
+      #
+      #   ActiveSupport::TestCase.parallel_worker_id # => 2
+      def parallel_worker_id
+        @@parallel_worker_id
+      end
+
+      def parallel_worker_id=(value) # :nodoc:
+        @@parallel_worker_id = value
+      end
+
       # Sets the order in which test cases are run.
       #
       #   ActiveSupport::TestCase.test_order = :random # => :random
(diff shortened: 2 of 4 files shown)


activesupport/lib/active_support/testing/parallelization/worker.rb
```diff
@@ -78,6 +78,7 @@ def safe_record(reporter, result)
         end
 
         def after_fork
+          ActiveSupport::TestCase.parallel_worker_id = @number if defined?(ActiveSupport::TestCase)
```
eileencodes: Hrm, when is `ActiveSupport::TestCase` not defined?

Also style nit, add a new line after this.

```suggestion
          ActiveSupport::TestCase.parallel_worker_id = @number if defined?(ActiveSupport::TestCase)
          
```

activesupport/test/test_case_test.rb
```diff
@@ -666,6 +666,47 @@ class ConstStubbable
 class SubclassOfConstStubbable < ConstStubbable
 end
 
+class ParallelizationTest < ActiveSupport::TestCase
```
eileencodes: Can you put this in it's own file?

activesupport/lib/active_support/test_case.rb
```diff
@@ -22,8 +22,24 @@
...
+      # Returns the current parallel worker ID if tests are running in parallel,
+      # nil otherwise.
+      #
+      #   ActiveSupport::TestCase.parallel_worker_id # => 2
+      def parallel_worker_id
+        @@parallel_worker_id
+      end
+      
+      # Sets the parallel worker ID
+      def parallel_worker_id=(value)
```
eileencodes: I'm not sure we want this API to be public (just the setter) because if someone sets this in their application weird things will happen and they'll think it's a bug in Rails. Only the framework should set this value. So this should have a `# :nodoc` instead of docs.

```suggestion
      def parallel_worker_id=(value) # :nodoc:
```

eileencodes (changes requested):
Thanks @Schwad! Can you add a changelog in addition to addressing the other feedback.

Schwad:
@eileencodes thanks for the feedback! Very good points on all fronts. I believe I have addressed it, and squashed

eileencodes (approved):

activesupport/test/parallelization_test.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class ParallelizationTest < ActiveSupport::TestCase
```
eileencodes, line 5: Oh the builds is failing because AS isn't defined. You're missing `require_relative "abstract_unit"` above this.

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Create `parallel_worker_id` helper for running parallel tests. This allows users to
+    know which worker they are currently running in.
+
+    *Nick Schwaderer*
```
eileencodes, line 4: I was going to just fix this after merge but then saw the failing build so can you fix this and the missing require? Let me know if you can't get to this and I'll do a manual merge and fix the two remaining issues.

```suggestion
    *Nick Schwaderer*
    
```

eileencodes (changes requested):

activesupport/test/parallelization_test.rb
Schwad: Oh dang it I ran the test locally and thought that was good enough I'll sort that out now

Schwad: Okay; I *think* I resolved it, and resquashified

eileencodes (approved):

PR #55566, opened 2025-08-28
Enable debug mode for events in local environments

adrianna-chang-shopify:
### Motivation / Background

Emit debug events by default in local environments. It would be good for devs to be able to see debug events when working locally, and it's also a bit of a pain to have to wrap tests in `Rails.event.with_debug` any time we want to assert that a debug event was emitted.

### Detail

Sets a new option on the Event Reporter, `debug_mode`, which will force the event reporter into debug mode. When bootstrapping the app, we configure that value based on `config.consider_all_requests_local`.

Also two small cleanups -- removed an unused `tags` kwarg from the constructor, and ensured `#notify` returns `nil`.

### Additional information

I considered combining `debug_mode` with `raise_on_error`, but decided it was clearer to keep them separate. We set both of them for local requests, but they don't mean the same thing. The former is enabling the emission of all debug events, and the latter defines behaviour if we encounter an error during event emission / in the subscriber code. 

I've also opted to no'doc `debug_mode`, I don't think apps should set this directly.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -265,15 +265,18 @@ module ActiveSupport
     # event emission, or when unexpected arguments are passed to +notify+.
     attr_writer :raise_on_error
 
+    attr_writer :debug_mode # :nodoc:
+
     class << self
       attr_accessor :context_store # :nodoc:
     end
 
     self.context_store = EventContext
 
-    def initialize(*subscribers, raise_on_error: false, tags: nil)
+    def initialize(*subscribers, raise_on_error: false)
       @subscribers = []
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/event_reporter.rb
```diff
@@ -265,15 +265,18 @@ class EventReporter
...
 
+    attr_writer :debug_mode # :nodoc:
+
     class << self
       attr_accessor :context_store # :nodoc:
     end
 
     self.context_store = EventContext
 
-    def initialize(*subscribers, raise_on_error: false, tags: nil)
```
adrianna-chang-shopify, line 274: `tags` was unused

PR #55569, opened 2025-08-29
Restore Active Storage config to disable variants and analyzers

zzak:
Note: This is a refresh of #55303 because the pr/branch image got stuck with a broken libxml2.

I'm re-opening to see if this fixes it.

---

Fixes a couple of mistakes made in e978ef0.

* Allow setting `config.active_storage.analyzers = []`
* Allow disabling `config.active_storage.variant_processor` to remove warning
* Ensure original default `ActiveStorage.analyzers` array is not broken (all available analyzers)
* Ensure original default `ActiveStorage.variant_processor` is not broken (`:mini_magick` until `load_defaults(7.0)`)

This change restores the configuration of analyzers back to 8.0, instead of trying to gracefully handle loading those constants if the associated gem is missing.

Due to that config being set at require time, this happens prior to initialization and we don't have access to a logger or deprecator to warn the user.

Since `image_processing` gem transitively depends on `ruby-vips` and `mini_magick`, you only need to install that to remove the warning.

However, if you're not using variant transformers, but still using the default analyzers you can remove the warning by including those gems or setting the analyzers config to an empty array.

We determine the variant transformer based on
`config.active_storage.variant_processor` value, but previously we would do nothing if you passed an invalid value (like `mmmmmini_magick`). That behavior is deprecated, and will raise in a future release.

That deprecation is to avoid a scenario where someone was unknowingly using an invalid value, but otherwise working application would now raise an exception on boot when upgrading Rails.

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,27 @@
+*   Allow analyzers and variant transformer to be fully configurable
+
+    ```ruby
+    # ActiveStorage.analyzers can be set to an empty array:
+    config.active_storage.analyzers = []
+    # => ActiveStorage.analyzers = []
+
+    # or use custom analyzer:
+    config.active_storage.analyzers = [ CustomAnalyzer ]
+    # => ActiveStorage.analyzers = [ CustomAnalyzer ]
+    ```
+
+    If no configuration is provided, it will use the default analyzers.
+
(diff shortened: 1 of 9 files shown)


zzak:
Ok.. maybe it is not libxml, I will investigate. :bow:

PR #55572, opened 2025-08-29
Fix for unsubscribing from event reporter using class

aidanharan:
### Motivation / Background

This PR fixes a small issue with unsubscribing using a class in the Event Reporter, which was introduced by https://github.com/rails/rails/pull/55487

The arguments of `===` were in the wrong order which prevented instances of a class being unsubscribed. 

```ruby
Rails.event.unsubscribe(MyEventSubscriber)
```

CC: @zzak 

### Checklist


zzak:
@aidanharan Sorry I just saw this, thank you for investigating and the fix! :bow:

PR #55576, opened 2025-08-29
Fix exec_{insert,update,delete} + query cache

skipkayhil:
[Fix exec_{insert,update,delete} + query cache](https://github.com/rails/rails/commit/80924268dc39bf89b0241841ac28d1d308c1a608)

Previously, these three methods did not clear the query cache even
though execute, exec_query, insert, update, and delete did.

This commit makes the exec_ prefixed methods consistent with their
non-prefixed counterparts: the query_cache clearing is moved to the
exec_ version so that both versions clear the query cache.

(create is also removed as a delegation target because its an alias of
insert, which now clears the query cache with exec_insert)

---

[Fix truncate clearing query cache twice](https://github.com/rails/rails/commit/ee2f027ef697fc2396e4b9eaed32dedcad9d8d75)

Since truncate calls execute and both of those methods are intercepted
to clear the query cache, it unecessarily clears the cache twice.

This commit removes the truncate delegation so that the cache only gets
cleared once (with execute). A test is added to ensure that this doesn't
regress if the implementation of truncate were to be changed in the
future.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
@@ -10,9 +10,9 @@ module ActiveRecord
 
       class << self
         def included(base) # :nodoc:
-          dirties_query_cache base, :exec_query, :execute, :create, :insert, :update, :delete, :truncate,
+          dirties_query_cache base, :exec_query, :execute, :truncate,
             :truncate_tables, :rollback_to_savepoint, :rollback_db_transaction, :restart_db_transaction,
-            :exec_insert_all
+            :exec_insert, :exec_insert_all, :exec_delete, :exec_update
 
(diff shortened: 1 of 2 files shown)


skipkayhil:
~Ah, truncate breaks out of transactional tests... let me fix that~ ✅

matthewd:
`exec_insert` and friends are [intended to be] overridden by adapters, so I think this is pushing the cache-clearing responsibility too low

skipkayhil:
> `exec_insert` and friends are [intended to be] overridden by adapters, so I think this is pushing the cache-clearing responsibility too low

Ah, I see that now. Do you think `dirties_query_cache` could be changed to prepend methods instead of redefining them so that adapters can define their own methods or do you think this just isn't worth changing?

PR #55579, opened 2025-08-30
Improve performance of ActiveJob serializers

camertron:
### Motivation / Background

Hi there! 👋  Thanks for all you do to maintain Rails 😄 

I was looking at the interface for custom `ActiveJob::Serializers` and realized it requires a `serialize?` method that is called for potentially every serializer for every argument in the worst case. Because custom serializers are only considered after built-in serializers for complex types like `DateTime` and `Range` are considered, this behavior can lead to a fairly severe performance penalty for custom serializers. In fact, the more serializers you add, the worse performance becomes.

Currently, serializers are checked in this order:

1. SymbolSerializer
1. DurationSerializer
1. DateTimeSerializer
1. DateSerializer
1. TimeWithZoneSerializer
1. TimeSerializer
1. ModuleSerializer
1. RangeSerializer
1. BigDecimalSerializer
1. Any custom serializers

Benchmarking using [this script](https://gist.github.com/camertron/99a669ca4b20ad2ab7db4e1ab1e56630) produces the following chart showing the degradation in performance as additional custom serializers are added:

custom

This is in addition to the performance penalty already incurred from needing to check all the built-in serializers from the list above.

This pull request makes several changes to ActiveJob serializers that lead to a roughly 2x performance improvement in the linked benchmark, although the actual improvement depends on how may custom serializers are configured and which are used in any particular serialization operation.

### Detail

This pull request makes the following changes:

1. Avoids the overhead of `delegate`d method calls on built-in serializers. This leads to a modest ~3% speed improvement.
1. Introduces the `ClassBasedSerializer` that uses a hash to map classes to their associated built-in serializer. This is where the majority of the performance improvement comes from, since the lookup changes from O(n) complexity to O(1)
1. Allows registration of custom "class-based" serializers via `ActiveJob::Serializers.add_class_based_serializer` that offers the same O(1) lookup performance characteristics.

#### Avoiding delegated method calls

Built-in serializers inherit from `ActiveJob::Serializers::ObjectSerializer`, which is a Ruby singleton. Each is registered by its class with `ActiveJob::Serializers`. This means methods like `serialize?`, `serialize`, etc are called on the class. The first change proposed in this PR registers the singleton instance instead, avoiding the delegated method call overhead. In other words, instead of this:

```ruby
ActiveJob::Serializers.add_serializers(RangeSerializer)
```

the proposed code does this:

```ruby
ActiveJob::Serializers.add_serializers(RangeSerializer.instance)
```

#### The class-based serializer

All the existing built-in serializers implement the `serialize?` method in terms of `Object#is_a?`. For example, `RangeSerializer#serialize?` returns `true` when passed a `Range` object because `argument.is_a?(Range)` returns `true`. This also means `RangeSerializer` will serialize objects that _inherit_ from `Range`, although that is very probably not the common case. The `ClassBasedSerializer` improves the performance of the common case by mapping exact classes to their associated serializer, meaning it will serialize `Range` objects but not objects that inherit from `Range`. This mapping allows   much faster serializer lookups than the existing series of `is_a?` checks.

Since we haven't actually changed `RangeSerializer`, it will continue to work as a fallback for the uncommon case as well.

#### Custom "class-based" serializers

Developers can add their own custom class-based serializers via `ActiveJob::Serializers.add_class_based_serializer(klass, serializer_klass)`.

### Additional information

With the two changes described above, we can see how the number of iterations per second for built-in and custom types improves across the board. `SymbolSerializer` is the first built-in serializer checked, so there is no improvement i.e. both the iterative and hash-based approaches contribute 50% of the total iterations per second.

However as the iterative approach has to check more and more serializers, we can see the hash-based approach performing better and better compared to the iterative approach.

chart

### Checklist

### DIFF
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -8,6 +8,7 @@ module ActiveJob
   module Serializers # :nodoc:
     extend ActiveSupport::Autoload
 
+    autoload :ClassBasedSerializer
     autoload :ObjectSerializer
     autoload :TimeObjectSerializer
     autoload :SymbolSerializer
@@ -55,16 +56,26 @@ module ActiveJob
       def add_serializers(*new_serializers)
         self._additional_serializers += new_serializers.flatten
       end
+
+      # Adds a new serializer that will be used to serialize a specific class. These serializers do not consider
+      # ancestors, and will only be used if the class matches exactly. Class-based serializer lookups are much faster
+      # than regular serializer lookups, which have to iterate through all serializers to find one that can handle the
+      # given object. Most regular serializers call the object's `#is_a?` method to determine if they can handle the
+      # object, which can incur a performance penalty.
+      def add_class_based_serializer(klass, serializer)
+        ClassBasedSerializer.add_serializer(klass, serializer)
+      end
     end
 
-    add_serializers SymbolSerializer,
-      DurationSerializer,
-      DateTimeSerializer,
-      DateSerializer,
-      TimeWithZoneSerializer,
-      TimeSerializer,
-      ModuleSerializer,
-      RangeSerializer,
-      BigDecimalSerializer
+    add_serializers ClassBasedSerializer.instance,
+      SymbolSerializer.instance,
+      DurationSerializer.instance,
+      DateTimeSerializer.instance,
+      DateSerializer.instance,
+      TimeWithZoneSerializer.instance,
+      TimeSerializer.instance,
+      ModuleSerializer.instance,
+      RangeSerializer.instance,
+      BigDecimalSerializer.instance
   end
 end
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Improve performance of ActiveJob serializers
+
+    Introduces a new `add_class_based_serializer` method on `ActiveJob::Serializers`
+    that allows specifying a custom class and serializer pair. These serializers are
+    used to serialize a specific class and, unlike existing serializers, do not consider
+    ancestors of the given class when determining if they can handle an object.
+    Class-based serializer lookups are much faster than regular serializer lookups, which
+    work by iterating over all serializers to find one that can handle the given object.
+
+    *Cameron Dutro*
+
 *   Allow jobs to the interrupted and resumed with Continuations
 
     A job can use Continuations by including the `ActiveJob::Continuable`
--- a/activejob/lib/active_job/serializers/big_decimal_serializer.rb
+++ b/activejob/lib/active_job/serializers/big_decimal_serializer.rb
@@ -13,10 +13,9 @@ module ActiveJob
         BigDecimal(hash["value"])
       end
 
-      private
-        def klass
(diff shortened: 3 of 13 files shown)


activejob/lib/active_job/serializers.rb
```diff
@@ -55,16 +56,26 @@ def serializers
       def add_serializers(*new_serializers)
         self._additional_serializers += new_serializers.flatten
       end
+
+      # Adds a new serializer that will be used to serialize a specific class. These serializers do not consider
+      # ancestors, and will only be used if the class matches exactly. Class-based serializer lookups are much faster
+      # than regular serializer lookups, which have to iterate through all serializers to find one that can handle the
+      # given object. Most regular serializers call the object's `#is_a?` method to determine if they can handle the
```
byroot, line 63:
```suggestion
      # given object. Most regular serializers call the object's `#is_a?` method to determine if they can handle the
```

The performance of `is_a?` is rather negligible if the argument is a class. It used to be quite costly , but was optimized a few years ago in https://github.com/ruby/ruby/pull/5568

byroot:
Thanks for the PR. I'll find some time to review this, but I think there might be simpler solutions.

activejob/lib/active_job/serializers.rb
camertron: Ah yes, I remember reading about that a few years ago. I haven't actually profiled the serializers, but I wonder then if the majority of the perf improvement is just removing a layer of method dispatch?

camertron:
@byroot awesome, a simpler solution would be great! Thanks for taking a look ❤️

activejob/lib/active_job/serializers.rb
byroot: > I wonder then if the majority of the perf improvement is just removing a layer of method dispatch?

Not quite: https://share.firefox.dev/3USDwD6

The current implementation is a bit silly indeed because it's basically:

```ruby
serializers.each do |s|
  return s.serialize(obj) if s.serialize?(obj)
end
```

In your specific benchmark, 66% of the serialization time is just calling `serialize?`, which is agravated by the fast that the default `serialize?` implementation call another method.

e.g.
```ruby
def klass
  Time
end

def serialize?(argument)
  argument.is_a?(klass)
end

serialize(obj)
```

Which is way slower than directly calling `obj.is_a?(Time)`, YJIT can optimize the shit out of the later.

But you patch also help by turning a largely `O(n)` algo into a mostly `O(1)` one.

However I don't think we need a new API to implement that, we can call `#klass` when a new serializer is registered and then have the lookup be:

```ruby
serializer = serializers_by_klass[obj.class] || serializers.detect { ... }
```

But also some types like Symbol should just be "builtin", no need to go through a dispatch for them, they should be handled in the main loop.

Then, there are some serializers, like TimeWithZone that are inherently slow and can be sped up 10x: https://github.com/rails/rails/compare/rails:1dda631...byroot:1e9cf20

And finally, the serialization format isn't ideal, so I'm quite tempted to introduce format versioning and introduce a new more efficient format. Since we're meeting with the core team next week, I'll likely mention it there.

PR #55583, opened 2025-08-30
Optimize Active Job argument serialization by ~5x

byroot:
Fix: https://github.com/rails/rails/pull/55579 (FYI @camertron)

See individual commits for details.

Overall gains (benchmark source is in the first commit)

```
ruby 3.5.0dev (2025-08-27T10:41:07Z master 5257e1298c) +YJIT +PRISM [arm64-darwin24]
opt-aj-serialize-index
                        26.282k i/100ms
opt-aj-serialize-index
                        268.003k (± 1.0%) i/s    (3.73 μs/i) -      1.340M in   5.001883s

Comparison:
                  main:    52026.7 i/s
opt-aj-serialize-index:   268003.4 i/s - 5.15x  faster
```

The benchmark is taken from https://github.com/rails/rails/pull/55579. I'm not convinced it's particularly representative of job arguments payloads out there, but it does test a wide range of serializers.

Some stones left unturned:

  - `DateTime` objects are serialized with `TimeSerializer` calling `DateTime#xmlschema` and somehow this is way slower than `Time#xmlschema`. `DateTime` objects are somewhat deprecated, but could be worth checking what's up.
  - The format itself is very inefficient, as discussed with @Earlopain in https://github.com/rails/rails/pull/55544, things like `_aj_symbol_keys` being an array of string is a waste if we assume most hash either have all or none of their keys are symbol. However this requires a schema change, so is a bit out of scope of this PR.
  - The `_aj_serialized` values being fully qualified class name is very wasteful. At least for the builtin ones we should have smaller names. The benchmark payload is `1.35kB` once serialized in JSON, that's a bit ridiculous. It could be shinked by 25% just by removing the namespace part from the serializers, or `~30%` by also using shorter keys.

### DIFF
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -19,6 +19,7 @@ module ActiveJob
     autoload :ModuleSerializer
     autoload :RangeSerializer
     autoload :BigDecimalSerializer
+    autoload :ActionControllerParametersSerializer
 
     @serializers = Set.new
     @serializers_index = {}
@@ -57,6 +58,14 @@ module ActiveJob
       # Adds new serializers to a list of known serializers.
       def add_serializers(*new_serializers)
         new_serializers = new_serializers.flatten
+        new_serializers.map! do |s|
+          if s.is_a?(Class) && s < ObjectSerializer
+            s.instance
+          else
+            s
+          end
+        end
+
         @serializers += new_serializers
         index_serializers
         @serializers
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -20,15 +20,15 @@ module ActiveJob
     autoload :RangeSerializer
     autoload :BigDecimalSerializer
 
-    mattr_accessor :_additional_serializers
-    self._additional_serializers = Set.new
+    @serializers = Set.new
+    @serializers_index = {}
 
     class << self
       # Returns serialized representative of the passed object.
       # Will look up through all known serializers.
       # Raises ActiveJob::SerializationError if it can't find a proper serializer.
       def serialize(argument)
-        serializer = serializers.detect { |s| s.serialize?(argument) }
+        serializer = @serializers_index[argument.class] || serializers.find { |s| s.serialize?(argument) }
         raise SerializationError.new("Unsupported argument type: #{argument.class.name}") unless serializer
         serializer.serialize(argument)
       end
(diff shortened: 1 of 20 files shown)


activejob/lib/active_job/serializers.rb
```diff
@@ -19,16 +19,17 @@ module Serializers # :nodoc:
...
+    @serializers = Set.new
+    @serializers_index = {}
 
     class << self
       # Returns serialized representative of the passed object.
       # Will look up through all known serializers.
       # Raises ActiveJob::SerializationError if it can't find a proper serializer.
       def serialize(argument)
-        serializer = serializers.detect { |s| s.serialize?(argument) }
+        serializer = @serializers_index[argument.class] || serializers.find { |s| s.serialize?(argument) }
```
camertron, line 32: I thought about adding an optimization like this but I was concerned it could interfere with a serializer who's `serialize?` method returns `true` or `false` based on the properties of the given object, for example:

```ruby
class CustomClass
  attr_reader :foo

  def initialize(foo)
    @foo = foo
  end
end

class CustomClassSerializer < ActiveJob::Serializers::ObjectSerializer
  def serialize?(argument)
    !!argument.foo
  end
end
```

Memoizing the mapping from class -> serializer would use `CustomClassSerializer` for instances of `CustomClass` that have a nil `@foo` and ones that don't.

I doubt this is a common use-case, but it is a change in behavior that could be subtle and difficult to debug for Rails devs.

camertron (commented):
Thanks!

activejob/lib/active_job/serializers.rb
byroot: I know, but I'm not too concerned about that.

Users can define `def klass; nil; end` to skip the indexing.

composerinteralia:
I think this changed the behavior of `ActiveJob::Arguments.serialize` with a hash argument:

Before:

```rb
ActiveJob::Arguments.serialize({a: :b})
#=> 
[[{"_aj_serialized" => "ActiveJob::Serializers::SymbolSerializer", "value" => "a"}, {"_aj_serialized" => "ActiveJob::Serializers::SymbolSerializer", "value" => "b"}]]
```

After:

```rb
ActiveJob::Arguments.serialize({a: :b})
=> {"a" => {"_aj_serialized" => "ActiveJob::Serializers::SymbolSerializer", "value" => "b"}, "_aj_symbol_keys" => ["a"]}
```

The docs say Hashes are serialized element by element: https://api.rubyonrails.org/classes/ActiveJob/Arguments.html#method-i-serialize. Is that meant to be public API?

We started getting `"Can't serialize a Hash with reserved key \"_aj_symbol_keys\"` errors after upgrading GitHub to this commit, and I think this Otel patch what's using `ActiveJob::Arguments.serialize` (unfortunately via a private method): https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/743f7d4c4424caac8691a6115c4491db230158c5/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb#L28

byroot:
I see. Yes I changed the method to accept individual arguments rather than only accept arrays, but indeed that changes how hases are handled. I will restore that behavior.

PR #44699, opened 2022-03-16
Remove sanitize_limit method since ActiveModel::Attribute alerady use…

harrycis:
…s BindParam

### Summary

Remove sanitize_limit method since ActiveModel::Attribute already uses BindParam.

Issue [44489](https://github.com/rails/rails/issues/44489) has covered the topic.

### Other Information

NA

PR #55585, opened 2025-08-30
Simplify limit validation, move to call time

### LINKED ISSUE #44489: connection.sanitize_limit is no longer needed when building limit clause
connection.sanitize_limit [is currently called](https://github.com/rails/rails/blob/de7c495209811f8f918fa9f915e64df61a6080c1/activerecord/lib/active_record/relation/query_methods.rb#L1347) when building limit clause in Arel. 

It doesn't seem necessary any more for the following reasons:
1. build_cast_value results in  ActiveModel::Attribute type, which binds the value, and thus prevents SQLi
2. there is little reason to pass in an argument of Arel::Nodes::SqlLiteral type to +limit+ API

Current:
 arel.take(build_cast_value("LIMIT", connection.sanitize_limit(limit_value))) if limit_value
Expected:
 arel.take(build_cast_value("LIMIT", limit_value.to_i)) if limit_value

skipkayhil:
Fixes #44489
Closes #44699 

Previously, limit values were validated while building the Arel tree,
which can make it difficult to find the source of an invalid limit.

This commit moves the validation to `limit!` (where its actually set) so
that developers can more easily find the source of an invalid limit.

This also has the additional benefit of removing the need for a
connection in `build_arel` (which will be removed in a followup commit).

`sanitize_limit` currently ensures that limit values are either a
SqlLiteral or coerceable to an Integer. However, limits [use bind
params][1] and so end up quoted when compiled into queries:

```ruby
Topic.limit(Arel.sql("1, 1")).to_sql
# => SELECT "topics".* FROM "topics" LIMIT '1, 1'
```

Since SqlLiterals are quoted, you can actually only use SqlLiterals that
end up being a single value anyways (eg. `Arel.sql("1")`). Therefore,
instead of just inlining sanitize_limit into `limit!` it can be replaced
with a simple `Integer()` wrapper.

[1]: https://github.com/rails/rails/commit/574f255629a45cd67babcfb9bb8e163e091a53b8

---

[Remove unused connection arg from {build_,}arel](https://github.com/rails/rails/commit/ed6a2d76580bb5151e6591dd4a0de8ee4ec15419)

Partially revert https://github.com/rails/rails/commit/6ff9e82e4774d494298d6bfca2ceebe92aec64a1 because the
argument is no longer used. Since `aliases` is the only argument again,
I changed it back to be positional (which is what it was before the
connection argument was added in mentioned commit).

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1592,12 +1592,8 @@ module ActiveRecord
     end
 
     # Returns the Arel object associated with the relation.
-    def arel(conn = nil, aliases: nil) # :nodoc:
-      @arel ||= if conn
-        build_arel(conn, aliases)
-      else
-        with_connection { |c| build_arel(c, aliases) }
-      end
+    def arel(aliases = nil) # :nodoc:
+      @arel ||= build_arel(aliases)
     end
 
     def construct_join_dependency(associations, join_type) # :nodoc:
@@ -1751,7 +1747,7 @@ module ActiveRecord
         raise UnmodifiableRelation if @loaded || @arel
       end
 
-      def build_arel(connection, aliases = nil)
+      def build_arel(aliases)
         arel = Arel::SelectManager.new(table)
 
         build_joins(arel.join_sources, aliases)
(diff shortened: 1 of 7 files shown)


activerecord/lib/active_record/relation/query_methods.rb
```diff
@@ -1213,6 +1213,7 @@ def limit(value)
     end
 
     def limit!(value) # :nodoc:
+      value = Integer(value) unless value.nil?
```
byroot, line 1216: Should we check it's positive too?

byroot: Probably no big deal.

PR #55586, opened 2025-08-31
Add retryable option to Arel.sql with bind parameter support

euglena1215:
### Motivation / Background

In our Rails 7 application, we occasionally encountered transient `ConnectionFailed` errors during database migrations and queries. When Rails 8 introduced the `allow_retry` functionality, we upgraded expecting it to resolve these issues.

After upgrading to Rails 8, we did see a reduction in connection errors. However, we still encountered `ConnectionFailed` errors in queries that use bind parameters, such as `where("name LIKE ?", "%foo%")`.
These parameterized queries were not benefiting from the retry mechanism because they couldn't be marked as retryable.

This PR extends the retryable functionality introduced in #54951 by adding support for bind parameters in `Arel.sql`. Now developers can create retryable bound SQL literals for parameterized queries:

```ruby
# Before: Not retryable
User.where("name LIKE ?", search_term)

# After: Retryable with bind parameters
User.where(Arel.sql('name LIKE ?', search_term, retryable: true))
```

This enhancement allows our application (and others facing similar connection stability issues) to leverage the retry mechanism for a broader range of queries, leading to improved database connection resilience.

### Detail

This builds upon the retryable SqlLiteral support introduced in #54951 by extending Arel.sql to accept a retryable option with bind parameters. Previously, only plain SqlLiterals could be marked as retryable, but now developers can create retryable bound SQL literals for parameterized queries.

The retryable option works with both positional (?) and named (:name) bind parameters and integrates with ActiveRecord's existing retry infrastructure.

Example usage:

```rb
User.where(Arel.sql('name LIKE ?', search_term, retryable: true))
Post.where(Arel.sql('id = :id', retryable: true), { id: 1 })
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Add retryable option to `Arel.sql` with bind parameter support.
+
+    ```ruby
+    User.where(Arel.sql('name LIKE ?', search_term, retryable: true))
+    Post.where(Arel.sql('id = :id', retryable: true), { id: 1 })
+    ```
+
+    *euglena1215*
+
 *   Add replicas to test database parallelization setup.
 
     Setup and configuration of databases for parallel testing now includes replicas.
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1620,14 +1620,16 @@ module ActiveRecord
 
         case opts
         when String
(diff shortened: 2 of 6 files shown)


euglena1215:
All Active Record tests passed in my local environment. I've looked into the failing CI tests, and they appear to be unrelated to the changes in this PR. Would it be possible for someone to restart the build?
https://buildkite.com/rails/rails/builds/122117

euglena1215:
A certain amount of time has passed so I think I'll close it. Thank you!

PR #55589, opened 2025-09-01
Remove TableMetadata -> Reflection delegation

skipkayhil:
The delegation is unnecessary since the reflection instance is already available where needed in the predicate builder.

Removing it simplifies TableMetadata (as well as the classes it was previously passed to only to delegate to a reflection), and prevents a TableMetadata allocation except when actually used (querying a through association).

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/predicate_builder.rb
+++ b/activerecord/lib/active_record/relation/predicate_builder.rb
@@ -99,24 +99,26 @@ module ActiveRecord
           elsif value.is_a?(Hash) && !table.has_column?(key)
             table.associated_table(key, &block)
               .predicate_builder.expand_from_hash(value.stringify_keys)
-          elsif table.associated_with?(key)
+          elsif (associated_reflection = table.associated_with?(key))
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/relation/predicate_builder.rb
```diff
@@ -99,24 +99,26 @@ def expand_from_hash(attributes, &block)
           elsif value.is_a?(Hash) && !table.has_column?(key)
             table.associated_table(key, &block)
               .predicate_builder.expand_from_hash(value.stringify_keys)
-          elsif table.associated_with?(key)
+          elsif (associated_reflection = table.associated_with?(key))
```
rafaelfranca, line 102: Using a boolean method to get the value is a breaking the expectation of what this method means. Let's use the non-`?` version, or rename it since it is nodoc.

skipkayhil: https://github.com/rails/rails/pull/55700

PR #55592, opened 2025-09-01
Automatically isolate cache namespaces in tests

### LINKED ISSUE #48341: Rails cache prefix is not properly configured for isolated caching during parallel tests.
I've been investigating why I am seeing a lot of transient test failures due to cache collisions with parallelized tests. I thought it was because Rails doesn't update the cache prefix to protect from collisions but after looking at it; it's looking like Rails _tries_ to do that (by appending `Process.pid` and `Thread.current.object_id` to the namespace), but that it isn't working properly.

For the moment I've worked around it by adding namespace delineation myself in the `parallelize_setup` block of the test helper, but it'd be better if Rails handled this automatically so that no-one has to worry about it.

### Steps to reproduce

Replace `test/test_helper.rb` contents with the following:

```ruby
module ActiveSupport
  class TestCase
    # Run tests in parallel with specified workers
    parallelize(workers: :number_of_processors)

    # Setup all fixtures in test/fixtures/*.yml for all tests in alphabetical order.
    fixtures :all

    parallelize_setup do |worker|
      puts "#{worker.inspect}: #{Rails.cache.options[:namespace]}"
    end

    # Add more helper methods to be used by all tests here...
  end
end
```

### Expected Behavior

The namespaces listed should be unique per worker to isolate the test runners so that they do not collide during parallelization. It should look something like this (assuming that the cache prefix configured is `test:20230518:`, which is what mine is configured to be).

```
6: test:20230518:63512:3580:
5: test:20230518:63513:3580:
0: test:20230518:63514:3580:
3: test:20230518:63515:3580:
7: test:20230518:63516:3580:
4: test:20230518:63517:3580:
1: test:20230518:63518:3580:
2: test:20230518:63519:3580:
```

### Actual Behavior

Instead, the namespaces listed all share identical process identifiers and thread identifiers.

```
6: test:20230518:63512:3580:
5: test:20230518:63512:3580:
0: test:20230518:63512:3580:
3: test:20230518:63512:3580:
7: test:20230518:63512:3580:
4: test:20230518:63512:3580:
1: test:20230518:63512:3580:
2: test:20230518:63512:3580:
```

### System configuration
**Rails version**:

```
% fgrep rails Gemfile.lock
      rails-dom-testing (~> 2.0)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.2.0)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.1, >= 1.2.0)
    importmap-rails (1.1.6)
    rails (7.0.5)
    rails-dom-testing (2.0.3)
    rails-html-sanitizer (1.6.0)
    sentry-rails (5.9.0)
    sprockets-rails (3.4.2)
    tailwindcss-rails (2.0.29-arm64-darwin)
    tailwindcss-rails (2.0.29-x86_64-darwin)
    tailwindcss-rails (2.0.29-x86_64-linux)
    turbo-rails (1.4.0)
  importmap-rails (~> 1.1)
  rails (~> 7.0.5)
  sentry-rails
  sprockets-rails
  tailwindcss-rails (~> 2.0)
  turbo-rails
```

**Ruby version**:

```
% ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21]
```

Schwad:
Fixes: #48341
Ref: #55580

This change helps parallel tests avoid collisions with a shared cache store. It also helps existing test suites by removing the need to run `Rails.cache.clear` in setup or teardown.

With an existing namespace, `puts`'ing out the value during tests with `Rails.cache.namespace` should look like:

```
# Running: 

3eaf68e053car:my-existing-namespace
.704a79c87589r:my-existing-namespace
.8a10fc7d6c2cr:my-existing-namespace
.299e8dbfec5dr:my-existing-namespace
```

Without an existing namespace, it should look like:

```
# Running: 

ab9e605d3e56r:
.860de9cd1f47r:
.c0074cd994aer:
.a440598e8420r:
```

### DIFF
--- a/activesupport/lib/active_support/test_case.rb
+++ b/activesupport/lib/active_support/test_case.rb
@@ -17,6 +17,7 @@ require "active_support/testing/file_fixtures"
 require "active_support/testing/parallelization"
 require "active_support/testing/parallelize_executor"
 require "active_support/testing/notification_assertions"
+require "active_support/testing/cache_isolation"
 require "concurrent/utility/processor_counter"
 
 module ActiveSupport
@@ -197,6 +198,7 @@ module ActiveSupport
     include ActiveSupport::Testing::TaggedLogging
     prepend ActiveSupport::Testing::SetupAndTeardown
     prepend ActiveSupport::Testing::TestsWithoutAssertions
+    prepend ActiveSupport::Testing::CacheIsolation
     include ActiveSupport::Testing::Assertions
     include ActiveSupport::Testing::ErrorReporterAssertions
     include ActiveSupport::Testing::EventReporterAssertions
--- a/activesupport/lib/active_support/testing/cache_isolation.rb
+++ b/activesupport/lib/active_support/testing/cache_isolation.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module ActiveSupport
+  module Testing
+    # Cache Isolation
+    #
+    # Automatically randomizes cache namespaces for each test to ensure proper
+    # isolation. This prevents cache key collisions and removes the need for
+    # explicit cache clearing in teardown.
+    #
+    # This module is automatically included in ActiveSupport::TestCase.
+    module CacheIsolation
+      def self.prepended(klass)
+        klass.set_callback(:setup, :before, :isolate_cache_namespace)
+      end
+
+      private
+        def isolate_cache_namespace
+          return unless defined?(Rails) && Rails.respond_to?(:cache)
+
+          store = Rails.cache
+          return unless store.is_a?(ActiveSupport::Cache::Store)
+          return unless store.respond_to?(:namespace=)
+
+          original_namespace = store.namespace
+
+          isolated_namespace = if original_namespace
+            if original_namespace.include?("r:")
+              "#{SecureRandom.hex(6)}r:" + original_namespace.split("r:", 2)[-1]
+            else
+              "#{SecureRandom.hex(6)}r:#{original_namespace}"
+            end
+          else
+            "#{SecureRandom.hex(6)}r:"
+          end
+
+          store.namespace = isolated_namespace
+        end
+    end
+  end
+end
--- a/activesupport/test/cache/cache_isolation_test.rb
+++ b/activesupport/test/cache/cache_isolation_test.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require_relative "../abstract_unit"
+require "active_support/cache"
+require "active_support/testing/cache_isolation"
+
+# Mock Rails.cache for testing
+module Rails
+  class << self
+    attr_accessor :cache
+  end
+end
+
+class CacheIsolationTest < ActiveSupport::TestCase
+  def setup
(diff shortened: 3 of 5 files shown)


activesupport/lib/active_support/test_case.rb
```diff
@@ -197,6 +198,7 @@ def parallel_worker_id
     include ActiveSupport::Testing::TaggedLogging
     prepend ActiveSupport::Testing::SetupAndTeardown
     prepend ActiveSupport::Testing::TestsWithoutAssertions
+    prepend ActiveSupport::Testing::CacheIsolation
```
gmcgibbon, line 201: I don't think every test should use this. Let's make the module opt-in instead. Also why do we need this to be prepended?

activesupport/test/cache/cache_isolation_test.rb
```diff
@@ -0,0 +1,69 @@
...
+require_relative "../abstract_unit"
+require "active_support/cache"
+require "active_support/testing/cache_isolation"
+
+# Mock Rails.cache for testing
+module Rails
+  class << self
+    attr_accessor :cache
+  end
+end
```
gmcgibbon, line 12: This shouldn't be necessary. Please remove it.

activesupport/test/cache/cache_isolation_test.rb
```diff
@@ -0,0 +1,69 @@
...
+
+class CacheIsolationTest < ActiveSupport::TestCase
+  def setup
+    Rails.cache = ActiveSupport::Cache::MemoryStore.new(namespace: "test_app")
+    @original_namespace = Rails.cache.namespace
+  end
+
+  test "isolate_cache_namespace randomizes Rails.cache namespace" do
+    # Manually call the private method
+    send(:isolate_cache_namespace)
```
gmcgibbon, line 22: Do we actually need send here? It should be within the object.

activesupport/lib/active_support/testing/cache_isolation.rb
```diff
@@ -0,0 +1,43 @@
...
+    #
+    # This module is automatically included in ActiveSupport::TestCase.
+    module CacheIsolation
+      def self.prepended(klass)
+        klass.set_callback(:setup, :before, :isolate_cache_namespace)
+      end
+
+      private
+        def isolate_cache_namespace
+          return unless defined?(Rails) && Rails.respond_to?(:cache)
```
gmcgibbon, line 21: Referencing Railties in Active Support isn't right. Let's instead inject the cache to reset in some glue code that defaults to the rails cache and just make the cache an accessor on the module. This would also allow an escape hatch for anyone who needs to use a cache client that isn't Rails.cache.

gmcgibbon (changes requested):

activesupport/lib/active_support/test_case.rb
Schwad: This should be safe for every test that doesn't manually do weird things with that namespace. But that is fine. I think there is an argument to make it enabled for all _parallel_ tests by default so we are fixing cache collisions like we automatically make dbs parallel friendly e.g. https://github.com/rails/rails/issues/48341

PR #55596, opened 2025-09-01
wip: investigate CI failures

flavorjones:
This is a WIP PR to reproduce, investigate, and try to fix the failing CI jobs in https://github.com/rails/rails/pull/55303/files (e.g. [this build](https://buildkite.com/rails/rails/builds/121227/steps/canvas?jid=019903dc-f72e-49ce-88a1-7c443a0c21ca)).

cc @zzak

### DIFF
--- a/activestorage/lib/active_storage/engine.rb
+++ b/activestorage/lib/active_storage/engine.rb
@@ -107,6 +107,7 @@ module ActiveStorage
             end
 
           ActiveStorage.analyzers = [analyzer].compact.concat(app.config.active_storage.analyzers || [])
+          ActiveStorage.analyzers << ActiveStorage::Analyzer::ImageAnalyzer::Vips
           ActiveStorage.variant_transformer = transformer
         rescue LoadError => error
           case error.message


flavorjones:
@zzak I can't reproduce that failure. Not sure what's going on?

flavorjones:
zzak says I can reproduce locally with

```
gh pr checkout 55303
git clone --depth=1 https://github.com/rails/buildkite-config .buildkite
RUBY_IMAGE="ruby:3.4" docker compose -f .buildkite/docker-compose.yml build base
 IMAGE_NAME=buildkite-base RUBY_IMAGE=ruby:3.4 docker compose -f .buildkite/docker-compose.yml run default runner actiontext 'bin/test test/unit/content_test.rb:83'
```

so I'll try that.

flavorjones:
Yes, the issue here is that libvips is dynamically loading the system libxml2 before Nokogiri is loaded along with its vendored libxml2.

Inside the container I can see Nokogiri is precompiled, which means the issue is that we're exporting libxml2 symbols in the symbol table, which we've long planned to stop doing. See https://github.com/sparklemotion/nokogiri/discussions/2746 for the history and the work we'll be doing in the future.

PR #55599, opened 2025-09-01
Fix instability of mounted_helpers in `ActionDispatch::IntegrationTest`

alpaca-tc:
### Motivation / Background

When using `ActionDispatch::IntegrationTest` with a mounted Engine, the `main_app` helpers could become unstable and return different results depending on the last request.

For example, after accessing an Engine route, `main_app.root_path` unexpectedly changes:

```ruby
main_app.root_path #=> "/"
get "/mounted_engine/posts/1"
main_app.root_path #=> "/mounted_engine/"
```

### Detail

This issue was caused by `url_options[:script_name]`, which varied depending on the last request.

-  `ActionDispatch::IntegrationTest::Session#url_options` uses the `url_options` of the last accessed controller. [code](https://github.com/rails/rails/blob/d18731bed119911867d55983e07b2c853805d7d2/actionpack/lib/action_dispatch/testing/integration.rb#L142)
-  mounted_helpers works through a `RoutesProxy` that replaces `_routes` when generating URLs. [code](https://github.com/rails/rails/blob/d18731bed119911867d55983e07b2c853805d7d2/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L22)
-  This replacement should ensure that `script_name` generation switches correctly, producing the expected path. [code](https://github.com/rails/rails/blob/d18731bed119911867d55983e07b2c853805d7d2/actionpack/lib/action_controller/metal/url_for.rb#L45)
-  However, in `IntegrationTest`, the `_routes` replacement was missing, so the generated paths were incorrectly affected by the request history

This fix ensures that `_routes` replacement is also applied in `IntegrationTest`, making mounted_helpers behave consistently.

A new test was added to confirm that even after accessing an Engine route, `main_app.root_path` always returns `/`.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/testing/integration.rb
+++ b/actionpack/lib/action_dispatch/testing/integration.rb
@@ -329,6 +329,16 @@ module ActionDispatch
           path = location.path
           location.query ? "#{path}?#{location.query}" : path
         end
+
+        def _with_routes(routes, &block)
+          super(routes) do
+            if controller
+              controller.send(:_with_routes, routes, &block)
+            else
+              yield
+            end
+          end
+        end
     end
 
     module Runner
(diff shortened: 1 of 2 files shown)


alpaca-tc:
I noticed that the tests are failing. It looks like my fix was incorrect, so I’ll close this for now :pray:

PR #55600, opened 2025-09-01
Add custom active job serializers in a load hook

andrewn617:
### Motivation / Background

rails/rails@b6c472a optimized looking up active job serializes by indexing them by their klass when they are added.

Calling klass could load autoloaded code. Currently custom serializers are added in an after_initialize hook, meaning we will load autoloaded code at boot time in development and test. Instead, let's add them within an active job load hook to defer adding them until we load a job.

### Checklist


log/development.log
Edouard-chin, line 1: ✂️

PR #55601, opened 2025-09-01
Improve active job serializer #klass deprecation message

andrewn617:
### Motivation / Background

A deprecation was added in rails/rails@b6c472a to require a serializers #klass method to be public.

I was upgrading my app and noticed the deprecation method is slightly confusing. For

```
class MyModelSerializer
  private

  def klass
    MyModel
  end
end
```

We get `MyModel#klass method should be public.`. But `klass` is defined on the serializer not the model so it should be `MyModelSerializer#klass method should be public.`.

### Detail

Update the message to include the name of the serializer.

### Checklist


nvasilevski (approved):
Good catch!

PR #55616, opened 2025-09-05
Use singleton for callbacks default terminator

jhawthorn:
Previously we'd allocate a new proc for each `CallbackChain`. Instead we can use a singleton object that responds to call (we could use a singleton instance of a proc, but this should be slightly simpler for Ruby to execute).

(This could potentially also help make callbacks work inside Ractors, but there would be a lot more to do)

### DIFF
--- a/activesupport/lib/active_support/callbacks.rb
+++ b/activesupport/lib/active_support/callbacks.rb
@@ -573,7 +573,7 @@ module ActiveSupport
           @name = name
           @config = {
             scope: [:kind],
-            terminator: default_terminator
+            terminator: DEFAULT_TERMINATOR
           }.merge!(config)
           @chain = []
           @all_callbacks = nil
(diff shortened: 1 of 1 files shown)


rafaelfranca (approved):

activesupport/lib/active_support/callbacks.rb
```diff
@@ -671,6 +671,7 @@ def default_terminator
               terminate
             end
           end
+          DEFAULT_TERMINATOR = DefaultTerminator.new.freeze
```
bogdan, line 674: Not sure why class is needed. Can just be `DEFAULT_TERMINATOR = Proc.new {...}`

skipkayhil: > we could use a singleton instance of a proc, but this should be slightly simpler for Ruby to execute

PR #55618, opened 2025-09-05
Use frozen_string_literal by default in new apps

jhawthorn:
After this morning's "Chat With Maintainers" at Rails World there was a discussion that we could maybe enable frozen string literal by default.

This does this by modifying `RubyVM::InstructionSequence.compile_option` in the generated boot.rb. This way new Rails applications will have frozen string literal enabled by default (no change to existing applications).

(I made an app and it seems to work 🤞 CI agrees)

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/boot.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/boot.rb.tt
@@ -1,5 +1,10 @@
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
 
+# Enable frozen string literals
+if defined?(RubyVM::InstructionSequence.compile_option=nil)
+  RubyVM::InstructionSequence.compile_option = { frozen_string_literal: true }
+end
+
 require "bundler/setup" # Set up gems listed in the Gemfile.
 <% if depend_on_bootsnap? -%>
 require "bootsnap/setup" # Speed up boot time by caching expensive operations.


jhawthorn:
@byroot says we need to do something to handle bootsnap precompile

viktorianer:
> @byroot says we need to do something to handle bootsnap precompile

Something like https://github.com/rails/rails/issues/54039#issuecomment-3258765839?

byroot:
> Something like [#54039 (comment)](https://github.com/rails/rails/issues/54039#issuecomment-3258765839)?

No it's an entirely different problem.

`bootsnap` does include `RubyVM::InstructionSequence.compile_option` in its cache keys: https://github.com/rails/bootsnap/blob/55d663fbc419d3a4ad32d200f7ba45c64e04d772/lib/bootsnap/compile_cache/iseq.rb#L103-L107

So we need to make sure the compile options are the same when running `rails` and when running `bootsnap precompile`, otherwise the cache generated by `bootsnap precompile` will just be dead weight.

To precompile with frozen string literals we can simply set `RUBYOPT="--enable-frozen-string-literal" bootsnap precompile ...`, the annoying part is how to keep the two in sync. I'm afraid the best we can do is a remainder in comment.

railties/lib/rails/generators/rails/app/templates/config/boot.rb.tt
```diff
@@ -1,5 +1,10 @@
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
 
+# Enable frozen string literals
```
byroot, line 3: We probably need some long comment explaining to remove this if you run into issues in dependencies, etc.

Earlopain:
Ruby seems to want to drag this whole thing out longer once again and basically I agree with this. But I do wonder if it is a bit early: a deprecation warning was only added in 3.4, and a certain test framework doesn't even print deprecations by default (https://github.com/rspec/rspec/pull/161).

I don't remember much trouble with enabling frozen string literals, I did it myself after other big players already did most of the work in popular gems. So for the majority, I feel like gems are already ready for this. But I wonder if people (especially new ruby/rails users) would put one and one together and even know to look into this file to resolve their error.

p8:
With https://github.com/rails/rails/pull/57252 merged I think this one can be closed.

Issue #55617, opened 2025-09-05
public_file_server.headers can't contain procs


jdelStrother:
The new template for config/environments/production.rb sets `public_file_server.headers["cache-control"]` to a proc, but I don't see anything that actually calls that proc.
Instead Rack just dumps it to a string in the output headers, like `#<Proc:0x0000000133195300 /Users/jon/Developer/rails-static-headers/config/environments/development.rb:19 (lambda)>`

### Steps to reproduce

Create a new Rails app, run it with RAILS_ENV=production, or copy the `config.public_file_server.headers` setting from production.rb to development.rb

Fetch a static file & inspect the headers:

```shellsession
$ curl -I localhost:3000/robots.txt
HTTP/1.1 200 OK
last-modified: Fri, 05 Sep 2025 11:24:59 GMT
content-type: text/plain
cache-control: #<Proc:0x0000000133195300 /Users/jon/Developer/rails-static-headers/config/environments/development.rb:19 (lambda)>
content-length: 99
```

### Expected behavior

The response headers should include something like `cache-control: public, max-age=12345`

### Actual behavior

The response headers output a proc that's just been dumped to a string.

### System configuration

**Rails version**: 8.1.0.beta1 (5a0ba202b9c855534a88057c04d741b62fc391bb)

**Ruby version**: 
3.4.5

I've created a reproduction repo at https://github.com/jdelStrother/rails-static-headers with `rails new --main --api`.

---

cc @thisismydesign

PR #55624, opened 2025-09-05
Add callable support for public_file_server headers

codergeek121:
### Motivation / Background

As noted in #55617, there is no support for procs in the public file server. There was a test missing, so this probably went unnoticed. I'd took a stab at adding the functionality and wrote a basic test.

### Detail

This PR allows setting public file server headers with everything that responds to `:call`. In contrast to the original example configuration, I pass a single `Rack::Request` parameter directly instead of `path` and `_`.

### Additional information

If someone is using `ActionDispatch::FileHandler#call` directly and attempt is falsy, we're missing the headers. I think the cleanest way would be to add this to `Rack::Files` directly as well. Maybe I'm missing something, but I'm not sure if there are actual usages of `FileHandler#call` and if it's intended to be public api.

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/static.rb
+++ b/actionpack/lib/action_dispatch/middleware/static.rb
@@ -28,9 +28,22 @@ module ActionDispatch
     end
   end
 
+  # # ActionDispath Files
+  #
+  # `ActionDispatch::Files` serves files with support for setting headers dynamically with callables.
+  class Files < Rack::Files
+    private
+      def assign_headers(headers, request)
+        computed = @headers.transform_values do |value|
+          value.respond_to?(:call) ? value.call(request) : value
+        end
+        headers.merge!(computed)
+      end
+  end
+
   # # Action Dispatch FileHandler
   #
-  # This endpoint serves static files from disk using `Rack::Files`.
+  # This endpoint serves static files from disk using `ActionDispatch::Files`.
   #
   # URL paths are matched with static files according to expected conventions:
   # `path`, `path`.html, `path`/index.html.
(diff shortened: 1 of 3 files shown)


jdelStrother:
It looks like this falls through to Rack::Files when AD::FileHandler doesn't serve a file. I'm not entirely clear when this would be useful, but it doesn't seem like Rack::Files will handle a proc-header properly. Is that a problem?

byroot:
Yeah, we can't really do that without `Rack::Files` support. Given we're in beta, I'll start by reverting all this.

codergeek121:
Makes sense. I'll try to add this to `rack`'s  `Rack::Files` directly, even though I'm not sure if rack is open for this kind of change.

jdelStrother:
Do we need to fall back to Rack::Files here?

```ruby

def call(env)
  attempt(env) || @file_server.call(env)
end
```

I don't think I get the circumstances where AD::Static will fail to serve a file that will be usefully caught by regular Rack::Files.

codergeek121:
iiuc `AD::FileHandler#call` (not `AD::Static`) is not used by Rails itself directly. `AD::Static` is calling `#attempt` but not `#call` on `FileHandler`. But it's public, so removing it would technically be a breaking change if someone were to use `AD::FileHandler` directly, e.g. via `middleware.use ActionDispatch::FileHandler`.

I'm also not sure if there's a useful case where the fall through behavior is useful, but I still agree that the dynamic proc header feature would be best solved in Rack::Files directly.

codergeek121:
I opened up a [PR](https://github.com/rack/rack/pull/2377) in Rack and will update this PR once there's a decision on whether this feature can be added to Rack::Files directly.

bensheldon:
I saw that Rack merged your PR 🎉  Seems like Rails would have to subclass Rack::Files to implement this. I'm excited for the possibility.

Edit: oh, it's in Rack main and not in the Rack 3.2.x release branch. That means it's still unreleased.

codergeek121:
I updated thiss PR to use the new `assign_headers` method in `Rack::Files`, which right now makes it work for `RACK=head`. I'm not sure though how I should handle the compatibility with Rack & Rails once Rack releases the feature. Using older versions of Rack would not break, but also not support the dynamic headers.

The dynamic header configuration in `production.rb.tt` could be a simple comment, stating that if you're using a supported version you can make use of this feature. I'm not sure though if this would be good enough. The failing test could also do a Rack version check to avoid failure in older versions.

Lmk if you have any other suggestions and I'll adapt the PR 👍

PR #55625, opened 2025-09-06
Allow disabling action_controller.logger by setting it to nil or false

robertomiranda:
### Motivation / Background

This PR addresses https://github.com/rails/rails/issues/55518 .Rails guides state that action_controller.logger can be set to nil to disable logging. 

> Accepts a logger conforming to the interface of Log4r or the default Ruby Logger class, which is then used to log information from Action Controller. Set to nil to disable logging.

Previously, it always defaulted to `Rails.logger`, ignoring `nil` or `false` values. This change uses `options.fetch(:logger, Rails.logger)`, enabling apps to opt out of logging when desired.

### DIFF
--- a/railties/test/application/configuration_test.rb
+++ b/railties/test/application/configuration_test.rb
@@ -5059,6 +5059,50 @@ module ApplicationTests
       assert_equal 5, Regexp.timeout
     end
 
+    test "action_controller.logger defaults to Rails.logger" do
+      restore_default_config
+      add_to_config "config.logger = Logger.new(STDOUT, level: Logger::INFO)"
+      app "development"
+
+      output = capture(:stdout) do
+        get "/"
+      end
+
+      assert_equal Rails.application.config.action_controller.logger, Rails.logger
+      assert output.include?("Processing by Rails::WelcomeController#index as HTML")
+    end
+
+    test "action_controller.logger can be disabled by assigning nil" do
+      add_to_config <<-RUBY
+        config.logger = Logger.new(STDOUT, level: Logger::INFO)
+        config.action_controller.logger = nil
+      RUBY
+      app "development"
+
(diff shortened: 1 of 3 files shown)


p8 (approved):

p8:
This also applies to Active Record:
https://github.com/rails/rails/blob/83aec03da6e5dcb0d64ac40c0a51ec3b9f06de29/activerecord/lib/active_record/railtie.rb#L101

Action Cable:
https://github.com/rails/rails/blob/83aec03da6e5dcb0d64ac40c0a51ec3b9f06de29/actioncable/lib/action_cable/engine.rb#L26

Active Job:
https://github.com/rails/rails/blob/83aec03da6e5dcb0d64ac40c0a51ec3b9f06de29/activejob/lib/active_job/railtie.rb#L18

guilleiguaran:
@p8, that's right, we paired on this and found that the issue wasn't only on AC but in most of the other frameworks. I'll update the corresponding code.

railties/test/application/configuration_test.rb
```diff
@@ -5059,6 +5059,50 @@ class Foo < ApplicationRecord
...
+    test "action_controller.logger defaults to Rails.logger" do
+      restore_default_config
+      add_to_config "config.logger = Logger.new(STDOUT, level: Logger::INFO)"
+      app "development"
+
+      output = capture(:stdout) do
+        get "/"
+      end
+
+      assert_equal Rails.application.config.action_controller.logger, Rails.logger
```
chaadow, line 5071: hi @p8 this should be inverted 
```suggestion
      assert_equal Rails.logger, Rails.application.config.action_controller.logger
```

guilleiguaran: 032a286244

PR #55628, opened 2025-09-06
Setting max_connections to nil means no limit

matthewd:
We intend to make this the default: instead of users needing to do maths to choose a pool size based on the number of worker threads they have (in Puma etc.), we can just stop enforcing any limit. Under normal usage, actual consumption will be naturally limited by the number of threads that request a connection. (This natural limiting doesn't apply to code that uses explicit `checkout` calls instead of the standard connection-leasing model: such applications may still benefit from choosing a hard limit.)

In this PR, however, I am _not_ changing the default `max_connections` value from its current **5** -- just making sure there is a way to explicitly specify "unlimited" instead of needing to use an arbitrary large number.

### DIFF
--- a/activerecord/lib/active_record/database_configurations/hash_config.rb
+++ b/activerecord/lib/active_record/database_configurations/hash_config.rb
@@ -71,7 +71,10 @@ module ActiveRecord
       end
 
       def max_connections
-        (configuration_hash[:max_connections] || configuration_hash[:pool] || 5).to_i
+        max_connections = configuration_hash.fetch(:max_connections) {
+          configuration_hash.fetch(:pool, 5)
+        }&.to_i
+        max_connections if max_connections && max_connections >= 0
       end
 
       def min_connections
@@ -86,7 +89,7 @@ module ActiveRecord
       end
 
       def max_threads
-        (configuration_hash[:max_threads] || max_connections).to_i
+        (configuration_hash[:max_threads] || (max_connections || 5).clamp(0, 5)).to_i
       end
 
       def max_age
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/database_configurations/hash_config.rb
```diff
@@ -86,7 +89,7 @@ def min_threads
       end
 
       def max_threads
-        (configuration_hash[:max_threads] || max_connections).to_i
+        (configuration_hash[:max_threads] || max_connections || 5).to_i
```
matthewd: @byroot any thoughts on this?

byroot: I'm tempted to suggest:

```suggestion
        (configuration_hash[:max_threads] || 5).to_i
```

As you probably want to ensure the async query threads can't starve the web server threads out of connections. So seeing it now I feel like this default wasn't great.

byroot (approved):
👍

activerecord/lib/active_record/database_configurations/hash_config.rb
matthewd: I found a way to overcomplicate it instead 😅 

https://github.com/rails/rails/blob/fed08be767b2494950c1c2b5489f77136504221a/activerecord/lib/active_record/database_configurations/hash_config.rb#L92

wdyt?

byroot: Sure.

nateberkopec:
Too-small pools have plagued a lot of apps for years. Happy to see this will be no more (if you choose `nil`)!

nateberkopec:
FWIW I'd be "for" making this the default 🤷‍♂️

PR #49100, opened 2023-09-01
Make enums validatable without raising error

mechnicov:
### Motivation / Background

There is a pretty old issue #13971
@dhh offered to make such contributions in [2017](https://github.com/rails/rails/issues/13971#issuecomment-282989242)
But opened MR #41730 was not maintained since 2021
I tried to take into account the comments in that code review

### Detail

This Pull Request adds `:validate` option for enums

If you want the enum value to be validated before saving, use the option `:validate`:

```ruby
class Conversation < ApplicationRecord
  enum :status, %i[active archived], validate: true
end

conversation = Conversation.new

conversation.status = :unknown
conversation.valid? # => false

conversation.status = nil
conversation.valid? # => false

conversation.status = :active
conversation.valid? # => true
```

It is also possible to pass additional validation options:

```ruby
class Conversation < ApplicationRecord
  enum :status, %i[active archived], validate: { allow_nil: true }
end

conversation = Conversation.new

conversation.status = :unknown
conversation.valid? # => false

conversation.status = nil
conversation.valid? # => true

conversation.status = :active
conversation.valid? # => true
```

Otherwise `ArgumentError` will raise (standard current behavior):

```ruby
class Conversation < ApplicationRecord
  enum :status, %i[active archived]
end

conversation = Conversation.new

conversation.status = :unknown # 'unknown' is not a valid status (ArgumentError)
```

### Additional information

This was intended as non-breaking change

### Checklist

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55632, opened 2025-09-07
Introduce ActiveRecord.validate_enums_by_default config option

julienbourdeau:
### Motivation / Background

In ActiveRecord model, passing an invalid value to an`enum` attribute will raise `ArgumentError`.
@mechnicov introduce a way to make enum attribute validate like any other attribute, raising `ActiveRecord::InvalidRecord` in https://github.com/rails/rails/pull/49100

We end up adding `validate: true` to all enum. This PR adds a new ActiveRecord configuration key to control the default behavior.

~~Ideally, I'd love this to part of 8.1 with `false` by default. I know we're already in beta, so it might be too late.
Then, even default to `true` in the next major.~~

### Detail

- Introduce `ActiveRecord.validate_enums_by_default` config
- `false` by default for backward compatibility, even if I believe `true` makes more sense.
- Update `_enum` method to read config instead of using hardcoded `false` value

I'm open to any better param name of course.

### Additional information

Reminder about enum

```rb
  enum :status, %i[proposed accepted], validate: false # or `enum :status, %i[proposed accepted]` without validate option
end

Book.create(status: :unknown) # will raise ArgumentError

book = Book.new(status: :proposed)
book.status = :unknown # will raise ArgumentError
```

```rb
  enum :status, %i[proposed accepted], validate: true
end

Book.create!(status: :unknown) # will raise ActiveRecord::InvalidRecord

book = Book.new(status: :unknown)
book.valid? # false ; book.errors[:status] is set to ["value_is_invalid"]

book = Book.new(status: :invalid)
book.status = :unknown # nothing happens until validation
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Introduce `ActiveRecord.validate_enums_by_default` to control `enum` validation.
+    Set any value acceptable by the enum validate option. `false` by default.
+
+    ```ruby
+    config.active_record.validate_enums_by_default = true
+    config.active_record.validate_enums_by_default = false
+    config.active_record.validate_enums_by_default = {allow_nil: true}
+    ```
+
+    *Julien Bourdeau*
+
 *   Fix upsert_all when using repeated timestamp attributes.
 
     *Gannon McGibbon*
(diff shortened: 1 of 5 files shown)


MatheusRich:
I think we have to add an entry to https://guides.rubyonrails.org/configuring.html#configuring-active-record too.

mechnicov:
May be need to add some deprecation warning that default value will be change to true?

julienbourdeau:
@mechnicov Yeah, but I'd love some feedback from a maintainer first.

PR #55636, opened 2025-09-08
Disable bootsnap parallel precompilation to workaround QEMU bug

### LINKED ISSUE #54039: Default Dockerfile is broken for arm64 builds on Ruby 3.3.6 due to a Bootsnap bug
I'm on an Intel mac and the Docker build works if I leave the architecture as default. When targeting arm64 with `--platform linux/arm64` however, it hangs with no output from bootsnap.

If I use `docker buildx build .` then it works fine. Downgrading to Ruby 3.2.6 also works fine.

Same issue on Bootsnap GitHub: https://github.com/Shopify/bootsnap/issues/495

The problematic line in the dockerfile is:

```
RUN bundle install && \
    rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git && \
    bundle exec bootsnap precompile --gemfile
```

Breaking it apart into this shows that it's bootsnap, but gives no ouput:

```
RUN bundle install
RUN rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git
RUN bundle exec bootsnap precompile --gemfile --verbose
```

### Steps to reproduce

1. Install Ruby 3.3.6
2. install rails 8.0.1 gem with `gem install rails`
3. `rails new docker_test`
4. `cd docker_test`
5. `docker buildx build --platform linux/arm64 .`

### Expected behaviour:
The docker build completes with no errors.

### Actual behaviour:
Docker build hangs with the output from bundle install as the last lines on the console.

### System configuration
**Rails version**: 8.0.1

**Ruby version**: 3.3.6

**System**: Intel MacOS

viktorianer:
### Motivation / Background

Ref: https://github.com/rails/bootsnap/issues/495

Fixes #54039. Cross-arch Docker Buildx (QEMU) builds occasionally hang during parallel Bootsnap precompile. Single-threading avoids the deadlock.

### Detail

https://github.com/rails/bootsnap/issues/495#issuecomment-2657541922
> For those who are investigating, here are my attempts:
> ...
> `-j 0` takes around 350 seconds for a small Rails project on my machine, which is not bad compared to the time taken before the hang. (supposedly the time taken if `Process.fork` works correctly.)
>  ...
> The `-j 0` option is still the best contender for small projects, IMO.

https://github.com/rails/bootsnap/pull/501
> Try to detect QEMU fork bug

### Additional information

Lightweight safeguard; no Docker integration test (would be slow/flaky). Can be reverted if upstream issue is fully resolved.

### Checklist

* [x] Single focused change.
* [x] Commit message explains rationale and references issue.
* [x] Tests added.
* [ ] CHANGELOG (railties) entry to add if deemed user-visible.

### DIFF
--- a/activesupport/lib/active_support/test_case.rb
+++ b/activesupport/lib/active_support/test_case.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+p [:load_test_case]
+puts caller
+
 require "minitest"
 require "active_support/testing/tagged_logging"
 require "active_support/testing/setup_and_teardown"
--- a/railties/test/fixtures/Dockerfile.test
+++ b/railties/test/fixtures/Dockerfile.test
@@ -27,13 +27,14 @@ RUN apt-get install --no-install-recommends -y build-essential git pkg-config
 COPY Gemfile Gemfile.lock ./
 RUN bundle install && \
     rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git && \
-    bundle exec bootsnap precompile --gemfile
+    # -j 1 disable parallel compilation to avoid a QEMU bug: https://github.com/rails/bootsnap/issues/495
+    bundle exec bootsnap precompile -j 1 --gemfile
 
 # Copy application code
 COPY . .
 
-# Precompile bootsnap code for faster boot times
-RUN bundle exec bootsnap precompile app/ lib/
+# Precompile bootsnap code for faster boot times (single-threaded to mitigate Buildx/QEMU hang)
+RUN bundle exec bootsnap precompile -j 1 app/ lib/
 
 # Precompiling assets for production without requiring secret RAILS_MASTER_KEY
 RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile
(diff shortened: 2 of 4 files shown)


railties/test/fixtures/Dockerfile.test
```diff
@@ -27,13 +27,14 @@ RUN apt-get install --no-install-recommends -y build-essential git pkg-config
 COPY Gemfile Gemfile.lock ./
 RUN bundle install && \
     rm -rf ~/.bundle/ "${BUNDLE_PATH}"/ruby/*/cache "${BUNDLE_PATH}"/ruby/*/bundler/gems/*/.git && \
-    bundle exec bootsnap precompile --gemfile
+    # Force single-threaded bootsnap precompile to avoid Buildx/QEMU hangs
```
byroot: You can't do comments in chained queries like that. (The dockerfile format is 💩 )

byroot: Actually nevermind. I'm wrong on that one.

activesupport/lib/active_support/test_case.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+p [:load_test_case]
+puts caller
```
Earlopain, line 4: ☝️

byroot: Damn it.

byroot: 604e8e057020d3ae522f674e3fdcaf4dea160d31

viktorianer:
Do we need a merge in other Rails versions too? Can I help here?

PR #55638, opened 2025-09-09
[ci skip] Wording correction for url_from documentation

dajabe:
### Motivation / Background

This Pull Request has been created because the current wording for the subdomain handling of url_from implies that subdomains are considered safe but they are not. This is confusing and caused me to waste time debugging why my subdomains were raising a [UnsafeRedirectError](https://api.rubyonrails.org/classes/ActionController/Redirecting/UnsafeRedirectError.html)

### Detail

This Pull Request adds two words to the documentation to be more explicit about the behaviour shown in the example.

### Checklist

* [ x ] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [ x ] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ x ] Tests are added or updated if you fix a bug or add a feature.
* [ x ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -241,7 +241,8 @@ module ActionController
     #     url_from("http://example.com/profile")  # => "http://example.com/profile"
     #     url_from("http://evil.com/profile")     # => nil
     #
-    # Subdomains are considered part of the host:
+    # Subdomains are included when comparing hosts. The hostname must match exactly
+    # for a redirect to be considered internal.
     #
     #     # If request.host is on https://example.com or https://app.example.com, you'd get:
     #     url_from("https://dev.example.com/profile") # => nil


actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -227,7 +227,7 @@ def _compute_redirect_to_location(request, options) # :nodoc:
     #     url_from("http://example.com/profile")  # => "http://example.com/profile"
     #     url_from("http://evil.com/profile")     # => nil
     #
-    # Subdomains are considered part of the host:
+    # Subdomains are not considered part of the same host:
```
shivabhusal: The note _**`Subdomains are considered part of the host`**_ just means the app’s subdomain is included when checking the host. So `url_from` will verify that the URL matches the same host (including subdomains). Even other subdomains of the same main domain are considered different host and it makes sense. 

I think adding **`not`** could be misleading for readers. Instead, we can make the wording more explanatory. But, keeping as-is is also not a bad idea.

dajabe: I spent way to long on a problem with the assumption that this sentence meant that a subdomain should be considered internal. It was my error for not carefully reading the example after the sentence. When I finally did it wasn't till your explanation here that I understood that host is referring to the URL instead of the server.

I still think that the wording is confusing the way it is. I can make a better suggestion now with your explanation. I'll update with something that conveys both the original meaning and providing the clarification that I was looking for when reading these docs.

actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -227,7 +227,7 @@ def _compute_redirect_to_location(request, options) # :nodoc:
     #     url_from("http://example.com/profile")  # => "http://example.com/profile"
     #     url_from("http://evil.com/profile")     # => nil
     #
-    # Subdomains are considered part of the host:
+    # Subdomains are included in the evaluation of the host and are not considered internal.
```
shivabhusal: The new wording `“Subdomains are included in the evaluation of the host and are not considered internal”` is confusing. It could be read as if Rails treats all subdomains as part of the trusted host, which isn’t correct.

For example, if the configured host is `example.com`, a request to `foo.example.com` is not automatically trusted and will return `nil`, which is part of the host authorization security mechanism.

Given that, I find the existing docs clearer on this point, so I’d suggest leaving them as is and closing this PR.

dajabe: I opened the PR because I found the existing wording confusing. On first read it implied that the subdomain is part of the same host therefore internal and should be trusted. The existing wording does nothing to convey that a subdomain is not part of the same trusted host. 

You don't actually explain how the suggested wording is supposed to be confusing. The existing wording says it's part of the same host which has two meanings. It can mean the server itself which would imply it is internal or as it is intended in this case the specific portion of the URL in question. 

At the end of the day I have very little investment in this and was just hoping to help the next poor soul reading this documentation. If you don't see a problem I'm not going to waste any more energy on this. ✌️

ndbroadbent:
I totally agree with @dajabe here, I think the current wording is very ambiguous and could use an improvement.

```
# Subdomains are considered part of the host:
```

This can be easily interpreted in two different ways:

* If you set the host to `example.com`, then any subdomains (such as `foo.example.com`) are considered part of that host. As in, they match.

Or, the way it actually works today:

* If you set the host to `foo.example.com`, then `foo.` is part of the host - as in, we only match if the full hostname is `foo.example.com`

Here are some alternatives:

* The full host name, including any subdomain, must match exactly.
* Host matching includes the subdomain. Only the exact hostname is trusted.
* Subdomains are part of the hostname string and must match exactly to be trusted.


@shivabhusal would you be open to any of these options?

shivabhusal:
@ndbroadbent here’s what came to my mind:

---

**Subdomains are part of the host when evaluating both `request.host` and `location`. For a `location` to be considered trustworthy, its subdomain and domain must match those of `request.host`. Only then is it treated as an internal redirect.**

---

Happy to adjust anything — open to feedback and suggestions! 😄

ndbroadbent:
@shivabhusal thanks for looking into this again! Full disclosure - I got AI to help me review your proposal, and it came up with a few shorter versions that eliminate ambiguity with the shortest possible phrasing:

---

### Option A
Subdomains are included when comparing hosts. The hostname must match exactly for a redirect to be considered internal.

### Option B
Host comparison is exact, including subdomains. A redirect is only trusted if `request.host` and the redirect host are identical.

### Option C
Subdomains are included when comparing `request.host` and the redirect host. A redirect is trusted only if the entire hostname, including subdomain, matches exactly.

---

I think one of these slightly shorter options could be good. I think Option A is pretty clear and gets the point across without being too long

shivabhusal:
@ndbroadbent yes shorter would be good. Lets go with option A

rafaelfranca (approved):

PR #55640, opened 2025-09-09
[ci skip] Fix typo in Update upgrading_ruby_on_rails.md

amrrbakry:
Hello, I think there's a typo/mistake in the guide to upgrade from Rails 7.0 to Rails 7.1 in the Action Dispatch section. The new values are `:all` and `:none`, and they replace `true` and `false` and not the other way round, at least this how I understand it. Please let me know if this is not correct and I'm just being confused.


p8 (approved):

p8:
Thanks @amrrbakry !

PR #55641, opened 2025-09-09
Optimise `ActionDispatch::Http::URL.build_host_url` when protocol in host

joshuay03:
### Motivation / Background

When using URL helpers with a host that includes the protocol (e.g., `{ host: "https://api.example.com" }`), we (@buildkite) found that the already-extracted protocol was being unnecessarily processed through `#normalize_protocol`, creating extra allocations and showing up as a hotspot in our production profiles.

Example local profile: [output.json.gz](https://github.com/user-attachments/files/22236956/output.json.gz) (viewable at https://vernier.prof, select allocations for data source)

### Detail

**Before:**
1. `HOST_REGEXP` extracts `"https://"` from host
2. `normalize_protocol` processes `"https://"` through `PROTOCOL_REGEXP` and creates new string
3. `protocol.dup` creates another allocation

**After:**
1. `HOST_REGEXP` extracts `"https://"` from host  
2. Skip `normalize_protocol` since it's already in correct format
3. Skip `dup` since regex capture groups are mutable

This optimization only applies when no explicit protocol option is provided and `HOST_REGEXP` successfully extracts a protocol. The `HOST_REGEXP` pattern ensures extracted protocols are always in the final `"protocol://"` format that `normalize_protocol` would return, making the normalization step redundant.

The change eliminates 2 string allocations per URL generation when protocol is included in the host, resulting in a ~10-12% performance improvement and ~16% reduction in total allocations for that case. Other cases remain unaffected with no performance regression.

<details>
<summary>Benchmarking</summary>

<br>

**Script:**
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "./"

  gem "benchmark-ips"
  gem "benchmark-memory"
end

require "minitest/autorun"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

end
Rails.application.initialize!

  get "/", to: proc { [200, {}, ["Home"]] }, as: :home
end

class TestHelperWithoutProtocol

  def default_url_options
    { host: "example.org" }
  end
end

class TestHelperWithExplicitProtocol

  def default_url_options
    { host: "example.org", protocol: "http://" }
  end
end

class TestHelperWithProtocolInHost

  def default_url_options
    { host: "http://example.org" }
  end
end

  def test_benchmark
    th1 = TestHelperWithoutProtocol.new
    th2 = TestHelperWithExplicitProtocol.new
    th3 = TestHelperWithProtocolInHost.new

    Benchmark.ips do |x|
      x.report("home_url ips without protocol") { th1.home_url }
      x.report("home_url ips with explicit protocol") { th2.home_url }
      x.report("home_url ips with protocol in host") { th3.home_url }
    end

    puts "\n"

    Benchmark.memory do |x|
      x.report("home_url allocations without protocol") { th1.home_url }
      x.report("home_url allocations with explicit protocol") { th2.home_url }
      x.report("home_url allocations with protocol in host") { th3.home_url }
    end

    assert true
  end
end
```

**Results:**
```
main (604e8e057020d3ae522f674e3fdcaf4dea160d31):

ruby 3.4.5 (2025-07-16 revision 20cda200d3) +YJIT +PRISM [arm64-darwin25]
home_url without protocol
                        24.470k i/100ms
home_url with explicit protocol
                        21.399k i/100ms
home_url with protocol in host
                        22.408k i/100ms
home_url without protocol
                        245.376k (± 2.8%) i/s    (4.08 μs/i) -      1.248M in   5.090068s
home_url with explicit protocol
                        220.910k (± 2.5%) i/s    (4.53 μs/i) -      1.113M in   5.040106s
home_url with protocol in host
                        227.877k (± 2.8%) i/s    (4.39 μs/i) -      1.143M in   5.019124s

home_url allocations without protocol
                         2.680k memsize (   360.000  retained)
                        26.000  objects (     3.000  retained)
                         4.000  strings (     1.000  retained)
home_url allocations with explicit protocol
                         2.928k memsize (   360.000  retained)
                        29.000  objects (     3.000  retained)
                         6.000  strings (     1.000  retained)
home_url allocations with protocol in host
                         3.008k memsize (   360.000  retained)
                        31.000  objects (     3.000  retained)
                         6.000  strings (     1.000  retained)


optimise-build-host-url-with-protocol-in-host rebased on main (604e8e057020d3ae522f674e3fdcaf4dea160d31):

ruby 3.4.5 (2025-07-16 revision 20cda200d3) +YJIT +PRISM [arm64-darwin25]
home_url without protocol
                        24.536k i/100ms
home_url with explicit protocol
                        21.617k i/100ms
home_url with protocol in host
                        25.034k i/100ms
home_url without protocol
                        254.096k (± 2.5%) i/s    (3.94 μs/i) -      1.276M in   5.024416s
home_url with explicit protocol
                        230.278k (± 2.5%) i/s    (4.34 μs/i) -      1.167M in   5.072466s
home_url with protocol in host
                        254.472k (± 2.6%) i/s    (3.93 μs/i) -      1.277M in   5.020847s

home_url allocations without protocol
                         2.680k memsize (   360.000  retained)
                        26.000  objects (     3.000  retained)
                         4.000  strings (     1.000  retained)
home_url allocations with explicit protocol
                         2.928k memsize (   360.000  retained)
                        29.000  objects (     3.000  retained)
                         6.000  strings (     1.000  retained)
home_url allocations with protocol in host
                         2.680k memsize (   360.000  retained)
                        26.000  objects (     3.000  retained)
                         4.000  strings (     1.000  retained)
```

</details>

### Additional information

N/A

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/http/url.rb
+++ b/actionpack/lib/action_dispatch/http/url.rb
@@ -202,24 +202,24 @@ module ActionDispatch
 
           def build_host_url(host, port, protocol, options, path)
             if match = host.match(HOST_REGEXP)
-              protocol ||= match[1] unless protocol == false
-              host       = match[2]
-              port       = match[3] unless options.key? :port
+              protocol_from_host = match[1] if protocol.nil?
+              host               = match[2]
+              port               = match[3] unless options.key? :port
             end
 
-            protocol = normalize_protocol protocol
+            protocol = protocol_from_host || normalize_protocol(protocol).dup
             host     = normalize_host(host, options)
+            port     = normalize_port(port, protocol)
 
-            result = protocol.dup
+            result = protocol
 
             if options[:user] && options[:password]
               result << "#{Rack::Utils.escape(options[:user])}:#{Rack::Utils.escape(options[:password])}@"
             end
 
             result << host
-            normalize_port(port, protocol) { |normalized_port|
-              result << ":#{normalized_port}"
-            }
+
+            result << ":" << port.to_s if port
 
             result.concat path
           end
@@ -265,11 +265,11 @@ module ActionDispatch
             return unless port
 
             case protocol
-            when "//" then yield port
+            when "//" then port
             when "https://"
-              yield port unless port.to_i == 443
+              port unless port.to_i == 443
             else
-              yield port unless port.to_i == 80
+              port unless port.to_i == 80
             end
           end
       end
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Optimize `ActionDispatch::Http::URL.build_host_url` when protocol is included in host.
+
+    When using URL helpers with a host that includes the protocol (e.g., `{ host: "https://example.com" }`),
+    skip unnecessary protocol normalization and string duplication since the extracted protocol is already
+    in the correct format. This eliminates 2 string allocations per URL generation and provides a ~10%
+    performance improvement for this case.
+
+    *Joshua Young*, *Hartley McGuire*
+
 *   Allow `action_controller.logger` to be disabled by setting it to `nil` or `false` instead of always defaulting to `Rails.logger`.
 
     *Roberto Miranda*


skipkayhil:
Could the three branches be combined with something like this?

```diff
diff --git a/actionpack/lib/action_dispatch/http/url.rb b/actionpack/lib/action_dispatch/http/url.rb
index 7570a6a9af..12632736e8 100644
--- a/actionpack/lib/action_dispatch/http/url.rb
+++ b/actionpack/lib/action_dispatch/http/url.rb
@@ -202,15 +202,15 @@ def extract_subdomains_from(host, tld_length)

           def build_host_url(host, port, protocol, options, path)
             if match = host.match(HOST_REGEXP)
-              protocol ||= match[1] unless protocol == false
+              protocol_from_host = match[1] unless protocol == false
               host       = match[2]
               port       = match[3] unless options.key? :port
             end

-            protocol = normalize_protocol protocol
+            protocol = protocol_from_host || normalize_protocol(protocol).dup
             host     = normalize_host(host, options)

-            result = protocol.dup
+            result = protocol

             if options[:user] && options[:password]
               result << "#{Rack::Utils.escape(options[:user])}:#{Rack::Utils.escape(options[:password])}@"
```

joshuay03:
> Could the three branches be combined with something like this?

This line `protocol = protocol_from_host || normalize_protocol(protocol).dup` prioritises host-extracted protocols over explicit user options, breaking the expected behavior where user-provided `protocol` options should take precedence (`protocol ||= match[1] unless protocol == false`). We could replace it with this:
```ruby
  if protocol
    protocol = normalize_protocol(protocol).dup
  else
    protocol = protocol_from_host || normalize_protocol(nil)
  end
```

but I think the current approach is more idiomatic.

joshuay03:
> Could the three branches be combined with something like this?

This line `protocol = protocol_from_host || normalize_protocol(protocol).dup` prioritises host-extracted protocols, breaking the expected behavior where a truthy user-provided `protocol` should take precedence (`protocol ||= match[1] unless protocol == false`). We could replace it with this:
```ruby
if protocol.nil?
  protocol = protocol_from_host || normalize_protocol(nil)  
else
  protocol = normalize_protocol(protocol).dup
end
```

but I think the current approach is more idiomatic.

skipkayhil:
> where a truthy user-provided `protocol` should take precedence

Ah, right. I think the `protocol_from_host` condition could be inverted then?

```diff
diff --git a/actionpack/lib/action_dispatch/http/url.rb b/actionpack/lib/action_dispatch/http/url.rb
index 7570a6a9af..12632736e8 100644
--- a/actionpack/lib/action_dispatch/http/url.rb
+++ b/actionpack/lib/action_dispatch/http/url.rb
@@ -202,15 +202,15 @@ def extract_subdomains_from(host, tld_length)

           def build_host_url(host, port, protocol, options, path)
             if match = host.match(HOST_REGEXP)
-              protocol ||= match[1] unless protocol == false
+              protocol_from_host = match[1] if protocol.nil?
               host       = match[2]
               port       = match[3] unless options.key? :port
             end

-            protocol = normalize_protocol protocol
+            protocol = protocol_from_host || normalize_protocol(protocol).dup
             host     = normalize_host(host, options)

-            result = protocol.dup
+            result = protocol

             if options[:user] && options[:password]
               result << "#{Rack::Utils.escape(options[:user])}:#{Rack::Utils.escape(options[:password])}@"
```

joshuay03:
> Ah, right. I think the `protocol_from_host` condition could be inverted then?

I like it! I've gone ahead and actioned this, however, I still need the result dup branch because the `protocol` variable is used later by `normalize_port()`, but `result` gets mutated with `result << host`. Without the dup, both variables point to the same object, so the mutations corrupt the protocol value that `normalize_port()` expects.

I've also gone ahead and updated the benchmark results with the latest change just in case. TL;DR, the improvement is still around the same.

actionpack/lib/action_dispatch/http/url.rb
```diff
@@ -202,15 +202,15 @@ def extract_subdomains_from(host, tld_length)
...
-              protocol ||= match[1] unless protocol == false
-              host       = match[2]
-              port       = match[3] unless options.key? :port
+              protocol_from_host = match[1] if protocol.nil?
+              host               = match[2]
+              port               = match[3] unless options.key? :port
             end
 
-            protocol = normalize_protocol protocol
+            protocol = protocol_from_host || normalize_protocol(protocol)
```
p8: Previously this would always normalize `protocol`.
Does this skip normalizing `protocol_from_host` now when the regexp matches?

**Edit** Nevermind, that's the purpose of this whole PR 🙃

skipkayhil:
I added a test on main that fails with the current code (addc45c5b4242a899daf4bea02038a8a29c18556), and added a commit to fix it. Let me know if it looks good and I'll squash + merge 👍

joshuay03:
> I added a test on main that fails with the current code ([addc45c](https://github.com/rails/rails/commit/addc45c5b4242a899daf4bea02038a8a29c18556)), and added a commit to fix it. Let me know if it looks good and I'll squash + merge 👍

LGTM, thank you! Please feel free to add yourself to the changelog entry when squashing! :shipit:

PR #46981, opened 2023-01-12
Add test to BUNDLE_WITHOUT

rafaelfranca:
From #46978.

### Motivation / Background

This Pull Request has been created because test gems are not needed at deployment

### Detail

This Pull Request adds :test to BUNDLE_WITHOUT

### Additional information

Our Rails template at fly.io has been this way pretty much since we started creating Dockerfiles for Rails applications.  I haven't seen any issues or complaints.  If there is a reason to include test modules I would appreciate hearing the rationale.

See:

https://github.com/superfly/flyctl/blame/master/scanner/templates/rails/standard/Dockerfile#L38

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`

No tests or CHANGELOG are required for this change.

PR #55656, opened 2025-09-11
Exclude test dependencies from Docker image

jenseickmeyer:
### Motivation / Background

This Pull Request has been created because the "test" dependencies are included in the Docker image which is build based on the default `Dockerfile` generated when the app is created. However, like "development" dependencies the ones for "test" are not needed at runtime. Excluding these dependencies also makes the Docker image a bit smaller.

### Checklist


skipkayhil:
Thank you for the pull request, but this has been previously rejected: https://github.com/rails/rails/pull/46981

PR #55657, opened 2025-09-11
Emit structured debug events in development only

adrianna-chang-shopify:
### Motivation / Background

Ref: https://github.com/rails/rails/pull/55566

Shopify's local Event Reporter emitted debug events in local environments (test and development), so I ported the change over to Rails. However, after further discussion with the team, we decided it's strange to emit debug events in test; this causes extra noise, and tests typically want to verify production behaviour, so emitting debug events by default causes a divergence.

This PR enables debug events in development only. Tests can verify debug events by enabling debug mode explicitly, e.g.
```ruby
def test_debug
  assert_event_reported("test") do
    Rails.event.with_debug do
      something_that_emits_a_debug_event()
    end
  end
end
```

cc @skipkayhil @Edouard-chin 

I haven't added a CHANGELOG entry because the prior behaviour technically wasn't documented. 

We could also opt to make this configurable (and document the option), but not sure whether we want to encourage globally enabling / disabling it vs encouraging use of `with_debug` .

### DIFF
--- a/railties/lib/rails/application/bootstrap.rb
+++ b/railties/lib/rails/application/bootstrap.rb
@@ -73,7 +73,7 @@ module Rails
 
       initializer :initialize_event_reporter, group: :all do
         Rails.event.raise_on_error = config.consider_all_requests_local
-        Rails.event.debug_mode = config.consider_all_requests_local
+        Rails.event.debug_mode = Rails.env.development?
       end
 
       # Initialize cache early in the stack so railties can make use of it.


railties/lib/rails/application/bootstrap.rb
```diff
@@ -73,7 +73,9 @@ module Bootstrap
 
       initializer :initialize_event_reporter, group: :all do
         Rails.event.raise_on_error = config.consider_all_requests_local
-        Rails.event.debug_mode = config.consider_all_requests_local
+        if Rails.env.development?
+          Rails.event.debug_mode = true
+        end
```
p8: Could this just be the following?
```suggestion
        Rails.event.debug_mode = Rails.env.development?
```

adrianna-chang-shopify: Yes, good point @p8 ! 😂 I'll update.

skipkayhil (approved):

PR #55659, opened 2025-09-11
Do not inline process mailers when calling deliver_all_later

chaadow:
### Motivation / Background

This Pull Request has been created to fix https://github.com/rails/rails/issues/55658

### Detail

In our application we store each email attempt/delivery inside a table for tracking purposes, and I was surprised that `deliver_all_later` was taking a lot of time to send around 2000 emails. after looking at the logs i figured that all mailers were processed as if i was calling `deliver_now` 

### Additional information

@fatkodima I still don't know why by looking at this caller stack, but it seems `flatten!` eager load the mails by trying to delegate a method which triggers the inline processing here:
https://github.com/rails/rails/blob/aedfc4af7625312bfdec087a0d03e5ad5da80027/actionmailer/lib/action_mailer/message_delivery.rb#L65-L67

<details>
<summary> Here is the `callers` output </summary>
```
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/gems/sentry-rails-5.26.0/lib/sentry/rails/tracing.rb:56:in 'Sentry::Rails::Tracing::SentryNotificationExtension#instrument'",
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-d18731bed119/activesupport/lib/active_support/notifications.rb:210:in 'ActiveSupport::Notifications.instrument'",
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-d18731bed119/actionmailer/lib/action_mailer/base.rb:651:in 'ActionMailer::Base#process'",
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-d18731bed119/actionmailer/lib/action_mailer/message_delivery.rb:170:in 'block in ActionMailer::MessageDelivery#processed_mailer'",
"<internal:kernel>:91:in 'Kernel#tap'",
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-d18731bed119/actionmailer/lib/action_mailer/message_delivery.rb:169:in 'ActionMailer::MessageDelivery#processed_mailer'",
"/Users/user/.rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-d18731bed119/actionmailer/lib/action_mailer/message_delivery.rb:66:in 'ActionMailer::MessageDelivery#__getobj__'",
[...]

PR #55660, opened 2025-09-11
Add missing test when an array is passed to `ActionMailer.deliver_all_later`

chaadow:
related to https://github.com/rails/rails/pull/55659
cc @fatkodima 

Add test for the `if deliveries.first.is_a?(Array)` case

### DIFF
--- a/actionmailer/test/message_delivery_test.rb
+++ b/actionmailer/test/message_delivery_test.rb
@@ -168,6 +168,14 @@ class MessageDeliveryTest < ActiveSupport::TestCase
 
     assert_not mail1.processed?
     assert_not mail2.processed?
+
+    mail1 = DelayedMailer.test_message(1)
+    mail2 = DelayedMailer.test_message(2)
+
+    ActionMailer.deliver_all_later([mail1, mail2])
+
+    assert_not mail1.processed?
+    assert_not mail2.processed?
   end
 
   test "deliver_all_later enqueues multiple deliveries with correct jobs" do


fatkodima:
Thanks, but I I think it is not currently needed.

chaadow:
@fatkodima I'm sorry but I have to insist. if you remove this line: 

```ruby
deliveries = deliveries.first if deliveries.first.is_a?(Array)
```

All tests still pass. which means my line as it stand does not hold any value theoretically.

However when I add these tests: 
I have the following error

image


Could you please reconsider re-opening and merging please

actionmailer/test/message_delivery_test.rb
```diff
@@ -168,6 +168,14 @@ class DummyJob < ActionMailer::MailDeliveryJob; end
 
     assert_not mail1.processed?
     assert_not mail2.processed?
+
+    mail1 = DelayedMailer.test_message(1)
+    mail2 = DelayedMailer.test_message(2)
+
+    ActionMailer.deliver_all_later([mail1, mail2])
```
chaadow, line 175: @fatkodima could you please tell me why this is not needed? I'm passing an array, and the splat operator transforms it into : `[[ mail1, mail2 ]]` so I still need to test this case to check it does not break

maybe it should be extracted into another, more generic test, `test "deliver_all_later accepts an array of mailers"` 

because this is something that was not tested before with `flatten!` ( if you remove `flatten!` tests were still passing because you didn not add any test with an array of mailers `ActionMailer.deliver_all_later([mail1, mail2])`

In fact if you remove `deliveries.flatten!` and run the tests they still pass, which is an issue for me

PR #55661, opened 2025-09-11
If a serialized job is decoded when ActiveSupport.parse_json_times is true, use the parsed Time object

npezza93:
### Motivation / Background

In SolidQueue the main Job model looks like this:
```ruby
module SolidQueue
  class Job < Record
    class EnqueueError < StandardError; end

    include Executable, Clearable, Recurrable

    serialize :arguments, coder: JSON
```

`arguments` is what stores the serialized data of the job. `coder: JSON` ends up using `ActiveSupport::JSON` for encoding and decoding the data to/from the database.

If `ActiveSupport.parse_json_times` is set to true, the enqueued_at and scheduled_at keys get parsed as Time objects when instantiating the model. But when the job is deserialized those are expected to be strings. This results in an exception being raised because Time.iso8601 expects a string to be passed.

This change checks the type  to see if it's already a Time object and to skip parsing trying to parse it. ME]

### Checklist

### DIFF
--- a/activejob/lib/active_job/core.rb
+++ b/activejob/lib/active_job/core.rb
@@ -167,8 +167,8 @@ module ActiveJob
       self.exception_executions = job_data["exception_executions"]
       self.locale               = job_data["locale"] || I18n.locale.to_s
       self.timezone             = job_data["timezone"] || Time.zone&.name
-      self.enqueued_at          = Time.iso8601(job_data["enqueued_at"]) if job_data["enqueued_at"]
-      self.scheduled_at         = Time.iso8601(job_data["scheduled_at"]) if job_data["scheduled_at"]
+      self.enqueued_at          = deserialize_time(job_data["enqueued_at"])
+      self.scheduled_at         = deserialize_time(job_data["scheduled_at"])
     end
 
     # Configures the job with the given options.
@@ -207,6 +207,14 @@ module ActiveJob
 
       def arguments_serialized?
         @serialized_arguments
+      end
+
+      def deserialize_time(time)
+        if time.is_a?(Time)
+          time
+        elsif time
+          Time.iso8601(time)
+        end
       end
   end
 end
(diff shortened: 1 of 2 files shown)


composerinteralia:
This PR slightly changed some behavior we were relying on at GitHub. We can work around it easily enough, but figured I'd mention it anyway because I'm not sure what is meant to happen here, or if GitHub is even doing something that's supported. The `scheduled_at` from the `#set` call used to survive past the `#deserialize` call, but now it gets wiped out. If we want it to work as before we'd probably need to add back the `if job_data["scheduled_at"]`.

```
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails"
  gem "debug"
end

require "active_job/railtie"
require "minitest/autorun"
require "debug"

  config.eager_load = false

end
Rails.application.initialize!

class MyJob < ActiveJob::Base
  def perform
  end
end

class BuggyJobTest < ActiveJob::TestCase
  def test_stuff
    job = MyJob.new
    serialized = job.serialize
    job.set(wait_until: Time.current)
    job.deserialize(serialized)
    assert job.scheduled_at
  end
end
```

npezza93:
@composerinteralia Hmm we don't alter the data. Just avoid trying to parse an already created Time object. 

> before we'd probably need to add back the if job_data["scheduled_at"].

~~We do this currently in the `elsif` here. I'll dig in further a little later and see what's up.~~ 
Ahh I see what's up, I'll send up a fix. Thanks for reporting!

image

composerinteralia:
If the `time` argument is `nil`, then `deserialize_time` will return `nil` and we'll set `self.scheduled_at = nil` where we didn't set it previously.

```
self.scheduled_at         = deserialize_time(job_data["scheduled_at"])
```

That might be the correct behavior though. It's just different than before.

npezza93:
Yea I opened https://github.com/rails/rails/pull/55682 to restore the old behavior. @rafaelfranca Feel free to close it out the current behavior is correct though.

composerinteralia:
Thanks!

PR #29898, opened 2017-07-23
Fix regression introduced by fixing mounting the same engine in multiple locations

deivid-rodriguez:
### Summary

I introduced this regression in #29662. The previous version would prefer the current `script_name` if one is available in the request for resolving the mount context in engine route helpers. But that would make them incorrect in some cases.

For example (just describing the test I added), if an engine is mounted at

```ruby
resources :fruits do
  mount Bukkits::Engine => "/bukkits"
end
```

and the current request path is `/fruits/1/bukkits/posts`, calling `fruit_bukkits.posts_path(fruit_id: 2)` would incorrectly return `/fruits/1/bukkits/posts`. This is because the `script_name` present in the request (in this case `/fruits/1/bukkits`) was preferred (by using `reverse_merge!` instead of `merge!` [here](https://github.com/rails/rails/blob/127b475dc251a06942fe0cd2de2e0545cf5ed69f/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L34)).

This patch takes care of fixing that. My initial attempt was changing `reverse_merge!` with `merge!` but although that fixed the specific issue I was having, it broke some other tests. The problem is that the previous script name might contain relevant information that is not known to the `lambda` that resolves the script name from the parameters and the mount point location. For example, any context provided to the specific request via the environment's `SCRIPT_NAME` variable is not known to this method.

So instead of preferring the `script_namer` resolution, I merged that with the current script name. So if the previous script name was `/foo/fruits/1/bukkits`, and the resulting one from calling the lambda is `/fruits/2/bukkits`, the resulting script name used for the route helper will be `/foo/fruits/2/bukkits`.

### Other Information

Sorry for the probably not very well explained PR description. I'm not very familliar with rails internals, so it's hard for me to reason about them.

Also, excuse the irrelevant commits. I added them while investigating the problem, but then I found out about the imminent release, so I wanted to open this ASAP to avoid introducing the regression.

**EDIT**: I tried to improve a bit the description of the problem & the solution.

PR #55668, opened 2025-09-13
Fix `SCRIPT_NAME` handling in URL helpers for root-mounted engines

### LINKED ISSUE #34023: URL helpers for isolated engines disregard SCRIPT_NAME when it's set by app middleware
In Basecamp, we use Rack middleware to extract an account ID from the leading segment of each request's path and move it to `env['SCRIPT_NAME']`. This causes Rails to prefix every generated path with the account ID, freeing us from having to pass it as an argument at every call site for every URL helper.

Here’s a simplified example of such a middleware class:

```ruby
class SlugExtractor
  def initialize(app)
    @app = app
  end

  def call(env)
    request = ActionDispatch::Request.new(env)

    # If the first path segment is an integer, pull it off of path_info and put it in script_name.
    if request.path_info =~ /\A(\/(\d+))/
      request.script_name = $1
      request.path_info   = $'.empty? ? '/' : $'
    end

    @app.call(env)
  end
end
```

URL helpers for isolated Rails engines disregard `env['SCRIPT_NAME']` when it’s populated by middleware like `SlugExtractor`. If `SlugExtractor` sets `env['SCRIPT_NAME']` to `"/8675309"`, the following erroneously generates `"/google_sign_in/authorization"` instead of `"/8675309/google_sign_in/authorization"` (assuming [`GoogleSignIn::Engine`](https://github.com/basecamp/google_sign_in) is mounted at `/google_sign_in`):

```ruby
google_sign_in.authorization_path
```

When the code below is reached, it appears that `options[:script_name]` is incorrectly `nil`. As a result, it’s set to `"/google_sign_in"` instead of `"/8675309/google_sign_in"`:

https://github.com/rails/rails/blob/9fa4342970e2c8ac24bcc595f120ab65e88095a4/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L38-L41

For a failing test case, see [this gist](https://gist.github.com/georgeclaghorn/d362d442ab2149f84939e2f65945afd0).

flavorjones:
### Motivation / Background

When an engine is mounted at "/" and `SCRIPT_NAME` is being used, the URL helpers incorrectly drop the script name from the generated URLs.

### Detail

The issue is in the private method `ActionDispatch::Routing::RoutesProxy#merge_script_names` which doesn't handle the edge case of `new_script_name` being a sole slash "/" with no segment following it.

This bug does not manifest for non-root paths with trailing slashes (like "/bar/") because the paths are cleaned up by `Mapper::Mapping#normalize_path` in `Mapper::Resources.add_route`.

The fix I've adopted here is simply to remove any trailing "/" from the `new_script_name` variable. We could also just return `previous_script_name` if `new_script_name == "/"`, however I thought running the names through the existing logic would be preferable to a special-case guard clause.

### Additional information

I think this issue has been present since the changes introduced in #29898 in Rails v5.2.0. cc @deivid-rodriguez in case he has feedback! 🙏

I think this might fix https://github.com/rails/rails/issues/34023. cc @jeremy in case he remembers what was going on there.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   URL helpers for engines mounted at the application root handle `SCRIPT_NAME` correctly.
+
+    Fixed an issue where `SCRIPT_NAME` is not applied to paths generated for routes in an engine
+    mounted at "/".
+
+    *Mike Dalessio*
+
 *   Update `ActionController::Metal::RateLimiting` to support passing method names to `:by` and `:with`
 
     ```ruby
--- a/actionpack/lib/action_dispatch/routing/routes_proxy.rb
+++ b/actionpack/lib/action_dispatch/routing/routes_proxy.rb
@@ -54,6 +54,7 @@ module ActionDispatch
       # dependent part.
       def merge_script_names(previous_script_name, new_script_name)
         return new_script_name unless previous_script_name
+        new_script_name = new_script_name.chomp("/")
 
         resolved_parts = new_script_name.count("/")
         previous_parts = previous_script_name.count("/")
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
Thanks for catching my bug @flavorjones! I completely forgot the context at this point but one thing I can tell you: trailing slashes always get on my way!

jeremy (approved):

PR #55671, opened 2025-09-14
Adding dark mode to http error pages

pardeyke:
Screenshot 2025-09-14 at 11 12 31

I have added a dark appearance to the common http error pages. I have used the same colors as [Frexuz](https://github.com/Frexuz) in his patch for the welcome page #55538.

Also added `height: 100dvh` to the body for better mobile support.


shivabhusal (commented):
Good job! 👍🏼

The only concern I noticed is that in dark mode, the “Bad Request” text isn’t very readable due to low contrast. We should fix the contrast issue so it’s accessible. One idea would be to tweak the color of the “400” (or the heading text) specifically for dark mode.

pardeyke:
Screenshot 2025-09-16 at 11 53 04

You're right, it is kind of hard to read.
I'd rather adjust the red. Because the overlay is overlapping the h1 error code for 422 and also 500.

The preview is #ff4747.

jevin (commented):
This looks good!

MatheusRich:
@pardeyke I think it woudl be helpful to make the color on the error (400, 500, etc) closer to the background color, so it doesn't draw as much attention and the whole thing gets more readable. This is what light mode does too:
image

pardeyke:
done.

Screenshot 2025-09-21 at 11 45 57

shivabhusal:
it looks much better now 👍🏼

MatheusRich (approved):

MatheusRich:
cc @jathayde

jathayde:
Those look nice - my only comment would be that the code itself gets to be out of AA contrast when we go that faint. I do like it better visually, but is there a way we can balance that to the AA minimum?

pardeyke:
That's quite tricky. Had to lower the background brightness as well as the "500". So now we do have AA contrast between the red text and the "500".

Screenshot 2025-09-23 at 09 13 49

PR #55676, opened 2025-09-14
Document THRESHOLD_TO_JUSTIFY_COMPRESSION cannot be changed.

fxn:
This patch documents that the internal constant that decides if compression is applied on attribute encryption (`THRESHOLD_TO_JUSTIFY_COMPRESSION`) cannot be changed.

See why in the code comment.


fxn:
@jorgemanrubia good?

fxn:
I am going to merge, since I believe this code comment is correct and can be useful for maintainers. However, if there's anything else please followup!

PR #55683, opened 2025-09-17
Use PG::Connection#close_prepared when available

### LINKED ISSUE #55653: Postgres: Use Protocol Prepared Statement Deallocation Instead of Statement
Currently it appears that rail deallocates prepared statements via statement instead[0]. This is not efficient and does not work well with pgbouncer. See the Deallocation functionality in psycopg [2] as an example of protocol prepared statement Deallocation.

If I understand correctly it appears that rails uses the ruby-pg library [3] to execute protocol level prepare [4, 5]. Similarly I believe it should be a matter of invoking the `close_prepared` function in ruby-pg.

I have not tested the below code but I believe the solution could be something like this.

```diff
diff --git a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
index 18de389..9001f5a 100644
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -310,7 +310,9 @@ def dealloc(key)
             # don't need the complication of with_raw_connection because
             # a reconnect would invalidate the entire statement pool.)
             if conn = @connection.instance_variable_get(:@raw_connection)
-              conn.query "DEALLOCATE #{key}" if conn.status == PG::CONNECTION_OK
+              if conn.status == PG::CONNECTION_OK
+                conn.close_prepared "#{key}"
+              end
             end
           rescue PG::Error
           end

```

[0] https://github.com/rails/rails/blob/681c86674fa16996e4563b93aab4e9c494f6a651/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L313
[1] https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-CLOSE
[2] https://github.com/psycopg/psycopg/blob/94583702ada92187c8b1bb4da8de0d30a879c8bd/psycopg/psycopg/_connection_base.py#L489
[3] https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter.html
[4] https://deveiate.org/code/pg/PG/Connection.html#method-i-prepare
[5] https://github.com/rails/rails/blob/681c86674fa16996e4563b93aab4e9c494f6a651/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L927
[6] https://deveiate.org/code/pg/PG/Connection.html#method-i-close_prepared

skipkayhil:
Fix #55653

pgbouncer does not support "DEALLOCATE #{key}", but it does support "protocol level Close" (#close_prepared) starting in version 1.21.

This commit changes the PostgreSQL adapter to use #close_prepared (when available) to better support pgbouncer.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -288,6 +288,16 @@ module ActiveRecord
         database_version >= 10_00_00 # >= 10.0
       end
 
+      if PG::Connection.method_defined?(:close_prepared) # pg 1.6.0 & libpq 17
+        def supports_close_prepared? # :nodoc:
+          database_version >= 17_00_00
+        end
+      else
+        def supports_close_prepared? # :nodoc:
+          false
+        end
+      end
+
       def index_algorithms
         { concurrently: "CONCURRENTLY" }
       end
@@ -309,8 +319,12 @@ module ActiveRecord
             # accessed while holding the connection's lock. (And we
             # don't need the complication of with_raw_connection because
             # a reconnect would invalidate the entire statement pool.)
-            if conn = @connection.instance_variable_get(:@raw_connection)
-              conn.query "DEALLOCATE #{key}" if conn.status == PG::CONNECTION_OK
+            if (conn = @connection.instance_variable_get(:@raw_connection)) && conn.status == PG::CONNECTION_OK
+              if @connection.supports_close_prepared?
+                conn.close_prepared key
+              else
+                conn.query "DEALLOCATE #{key}"
+              end
             end
           rescue PG::Error
           end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -425,7 +425,13 @@ GEM
       ast (~> 2.4.1)
       racc
     path_expander (1.1.3)
-    pg (1.5.9)
+    pg (1.6.2)
+    pg (1.6.2-aarch64-linux)
+    pg (1.6.2-aarch64-linux-musl)
+    pg (1.6.2-arm64-darwin)
+    pg (1.6.2-x86_64-darwin)
+    pg (1.6.2-x86_64-linux)
+    pg (1.6.2-x86_64-linux-musl)
     pp (0.6.2)
       prettyprint
     prettyprint (0.2.0)
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Use `PG::Connection#close_prepared` (protocol level Close) to deallocate
+    prepared statements when available.
+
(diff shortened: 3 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -304,15 +308,32 @@ def next_key
...
+              # This is ugly, but safe: the statement pool is only
+              # accessed while holding the connection's lock. (And we
+              # don't need the complication of with_raw_connection because
+              # a reconnect would invalidate the entire statement pool.)
+              if conn = @connection.instance_variable_get(:@raw_connection)
+                if @connection.supports_close_prepared?
+                  conn.close_prepared key if conn.status == PG::CONNECTION_OK
+                else
+                  conn.query "DEALLOCATE #{key}" if conn.status == PG::CONNECTION_OK
+                end
```
skipkayhil: Also wondering if `conn.status == PG::CONNECTION_OK` is still needed after https://github.com/rails/rails/pull/44527

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -304,15 +308,32 @@ def next_key
...
 
         private
-          def dealloc(key)
-            # This is ugly, but safe: the statement pool is only
-            # accessed while holding the connection's lock. (And we
-            # don't need the complication of with_raw_connection because
-            # a reconnect would invalidate the entire statement pool.)
-            if conn = @connection.instance_variable_get(:@raw_connection)
-              conn.query "DEALLOCATE #{key}" if conn.status == PG::CONNECTION_OK
+          if PG::Connection.method_defined?(:close_prepared) # pg 1.6.0 & libpq 17
```
matthewd: Put this on `supports_close_prepared?` instead?

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
matthewd: Without checking, I'd guess it's probably still wanted, to be safe around the dealloc-during-raise process? But can probably move it up as a `return unless` rather than duplicate the conditional.

byroot:
```suggestion
                if conn.status == PG::CONNECTION_OK
                  if @connection.supports_close_prepared?
                    conn.close_prepared key
                  else
                    conn.query "DEALLOCATE #{key}"
                  end
                end
```

matthewd (approved):

byroot (approved):

PR #55684, opened 2025-09-17
[ci skip] docs: Fix out of date docs on cache.namespace

theodorton:
### Motivation / Background

This Pull Request has been created because I was trying to use cache namespaces and they didn't work as intended. Following the documentation and creating an initializer similar to the documented:

```ruby
Rails.cache.namespace = -> { @last_mod_time }
```

caused the application to crash.

### Detail

This Pull Request updates the documented example on how to set the namespace for an already initialized cache store instance.

### DIFF
--- a/activesupport/lib/active_support/cache.rb
+++ b/activesupport/lib/active_support/cache.rb
@@ -184,7 +184,7 @@ module ActiveSupport
     # is a Proc, it will be invoked when each key is evaluated so that you can
     # use application logic to invalidate keys.
     #
-    #   cache.namespace = -> { @last_mod_time }  # Set the namespace to a variable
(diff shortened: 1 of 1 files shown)


fatkodima:
Works for me
```ruby
> Rails.cache.namespace = "foo"
=> "foo"
```

theodorton:
The documentation was really old, so I just assumed it was out of date.

I'm getting `NoMethodErrors` for `namespace=` for `NullStore` and `MemoryStore`. I'll see if I can figure out what's wrong.

theodorton:
Sorry - seems like it was fixed recently in b6023059689263a95cbf22b2c2c7ca3d3d29c3b0 and we're still on Rails 8.0.2

PR #55685, opened 2025-09-17
fix delegation visibility when using `...` in Ruby 3.4.0–3.4.2

lHydra:
### Motivation / Background

Method delegation with the `...` argument allows calling private methods on another object, bypassing their visibility (relevant for Ruby versions 3.4.0–3.4.2, see: https://bugs.ruby-lang.org/issues/21196)

### Detail

Fix delegation visibility issue in Ruby 3.4.0–3.4.2.

In these Ruby versions, using `...` to forward arguments ignores method visibility, allowing private methods to be called. To work around this, delegation now explicitly expands arguments with `*args, **kwargs, &block` instead of `...`:

```ruby
RUBY_VERSION >= "3.4.0" && RUBY_VERSION <= "3.4.2" ? "*args, **kwargs, &block" : "
```
This ensures consistent and expected behavior across all Ruby versions.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Rails 8.1.0.beta1 (September 04, 2025) ##
 
+*   Fix issue with method visibility being ignored when delegating methods with arguments using `...` in Ruby versions 3.4.0 to 3.4.2.
+
+    Add explicit behavior for these Ruby versions:
+
+    ```ruby
+      RUBY_VERSION >= "3.4.0" && RUBY_VERSION <= "3.4.2" ? "*args, **kwargs, &block" : "..."
+    ```
+
+    In the old implementation, it was possible to call private methods of another class through delegation.
+
+    *Georgiy Melnikov*
+
 *   Add `ActiveSupport::Cache::Store#namespace=` and `#namespace`.
 
     Can be used as an alternative to `Store#clear` in some situations such as parallel
--- a/activesupport/lib/active_support/delegation.rb
+++ b/activesupport/lib/active_support/delegation.rb
@@ -93,14 +93,14 @@ module ActiveSupport
                 parameters = method_object.parameters
 
                 if parameters.map(&:first).intersect?([:opt, :rest, :keyreq, :key, :keyrest])
-                  "..."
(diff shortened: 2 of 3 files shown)


byroot:
I considered doing this when the bug was discovered, but decided not to as it adds complexity for a bug that is fixed upstream and you can update your Ruby patch release.

lHydra:
> I considered doing this when the bug was discovered, but decided not to as it adds complexity for a bug that is fixed upstream and you can update your Ruby patch release.

Maybe it makes sense to just use the direct arguments option `*args, **kwargs, &block` (without the `...` operator)? Then there won't be any extra complexity and the behavior will be the same for all versions. I haven't tested the performance yet, though, to see what differences there might be.

byroot:
`...` is way more performant on 3.4+ as it avoid multiple allocations.

Is there a reason you're running 3.4.2?

lHydra:
> `...` is way more performant on 3.4+ as it avoid multiple allocations.
> 
> Is there a reason you're running 3.4.2?

I need to certify the project when changing library versions. And there's currently no way to change the version.

I'm not sure, maybe we can leave a comment explaining why this change was made, and then just remove it in the future when the Rails stops supporting this Ruby version (3.4.Z). And will that be clear enough? It just doesn't seem quite right that a method can trigger incorrect behavior, especially in such a sensitive topic as method visibility. There could potentially be various reasons why people aren't using the latest Ruby version.

byroot:
> I need to certify the project when changing library versions.

Well, so adding a workaround in Rails doesn't help you, as you'll need to do the same certification to upgrade it.

> It just doesn't seem quite right that a method can trigger incorrect behavior

Of course, but it also doesn't quite seem right to fix Ruby inside Rails when the Ruby fix is already available.

lHydra:
> Well, so adding a workaround in Rails doesn't help you, as you'll need to do the same certification to upgrade it.

Yes, just knowing this bug helps me now, so I can simply avoid writing code that causes this behavior. But it was quite unexpected to see this behavior right away. Maybe other people have had a similar problem and aren't aware of it.

If this is just unnecessary complication, that's okay. I can close the PR.

byroot:
Yeah sorry, I'd rather not add that complication.

PR #55687, opened 2025-09-17
Consistently use 3 threads as `RAILS_MAX_THREADS` fallback

p8:
The new fallback is 3 but this wasn't used consistently, including in the database config templates.

### Checklist

### DIFF
--- a/actionmailbox/test/dummy/config/database.yml
+++ b/actionmailbox/test/dummy/config/database.yml
@@ -6,7 +6,7 @@
 #
 default: &default
   adapter: sqlite3
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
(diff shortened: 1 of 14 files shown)


byroot:
I think it's actually good that the connection pool is a bit larger, to accomodate for things like AC::Live etc.

Also if @matthewd merges https://github.com/rails/rails/pull/55628, I think the default should actually be no limit.

javierm:
> I think it's actually good that the connection pool is a bit larger

@byroot Not going to argue about that, but I'm finding the current behavior a bit confusing. If I understand it correctly, by default, the connection pool is a bit larger, but if you define `RAILS_MAX_THREADS` then the connection pool automatically gets the same size as the number of threads. Is that correct? :thinking:

PR #55688, opened 2025-09-17
Do not issue `Benchmark.ms` deprecation when method exists in the gem

nvasilevski:
`benchmark` gem is going to have `ms` method natively in the next version - https://github.com/ruby/benchmark/pull/38
So Rails doesn't need to issue a deprecation since there will be no breaking change for applications using `Benchmark.ms`

### DIFF
--- a/activesupport/lib/active_support/core_ext/benchmark.rb
+++ b/activesupport/lib/active_support/core_ext/benchmark.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "benchmark"
+return if Benchmark.respond_to?(:ms)
 
 class << Benchmark
   def ms(&block) # :nodoc


skipkayhil:
The release manager will handle removing deprecations, there are more to remove before the final 8.1.0 release.

I do think your point about skipping the patch if `Benchmark.ms` is already defined is a good one though, do you want to update the PR to do that instead?

nvasilevski:
>I do think your point about skipping the patch if Benchmark.ms is already defined is a good one though, do you want to update the PR to do that instead?

Done!

nvasilevski:
>The release manager will handle removing deprecations, there are more to remove before the final 8.1.0 release.

I didn't realize we still have a chance of this patch being dropped in `8.1` so perhaps it's better to wait for @rafaelfranca to make a decision as this change is only needed if we are going to keep the deprecation until `8.2`

byroot:
Regardless of deprecations, we shouldn't define that method if it already is defined by the gem. So I'll merge.

Might even backport it.

rafaelfranca:
Good call on backporting!

PR #55696, opened 2025-09-18
Fix query cache for pinned connections

### LINKED ISSUE #55689: Query cache does not work in system tests in Rails 7.2+
### Steps to reproduce

At GitLab, we are upgrading to Rails 7.2 and noticed that all our requests made from system specs were no longer using the query cache.

I dug into the code and found the query cache refactor in https://github.com/rails/rails/commit/85742ce529b028812066cd6070e7de2ff3d44bc8 and other subsequent changes related to that.

Based on that, the pool now owns the caches and it keeps a set of cache stores per thread. Now the problem happens when there is a pinned connection.

Here's the sequence of events:

1. RSpec thread checks out a connection and sets `connection.query_cache` to a new cache store based on the current thread. This connection is then pinned. The cache store for this connection and pool is disabled.
2. RSpec thread makes a web request to Puma
3. Puma processes the request and the Rails executor hook enables the query cache on a new cache store based on the Puma thread.
4. Puma thread checks out a connection and gets the same pinned connection. `checkout_and_verify` is not called in this code path so `connection.query_cache` still points to the cache store owned by the RSpec thread.
 
   This cache store isn't the one enabled by the executor hook so this request does not use the query cache. 

### System configuration

**Rails version**: 7.2

**Ruby version**: 3.3.9

engwan:
When a pinned connection is used across separate threads, they should use the correct cache store for the thread

### Motivation / Background

Fixes https://github.com/rails/rails/issues/55689

### Detail

Right now, a connection's query cache store is set on checkout. For pinned connections, this only happens once and this query cache store is shared by all threads even if we have separate query cache stores per thread.

This PR fixes this so that each thread will use the correct query cache store.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
@@ -147,12 +147,6 @@ module ActiveRecord
             end
         end
 
-        def checkout_and_verify(connection)
-          super
-          connection.query_cache ||= query_cache
-          connection
-        end
-
         # Disable the query cache within the block.
         def disable_query_cache(dirties: true)
           cache = query_cache
@@ -209,15 +203,19 @@ module ActiveRecord
         end
       end
 
-      attr_accessor :query_cache
+      attr_writer :query_cache
 
       def initialize(*)
         super
         @query_cache = nil
       end
 
+      def query_cache
+        @query_cache || pool.query_cache
+      end
+
       def query_cache_enabled
-        @query_cache&.enabled?
+        query_cache&.enabled?
       end
 
       # Enable the query cache within the block.
@@ -256,7 +254,7 @@ module ActiveRecord
 
         # If arel is locked this is a SELECT ... FOR UPDATE or somesuch.
         # Such queries should not be cached.
-        if @query_cache&.enabled? && !(arel.respond_to?(:locked) && arel.locked)
+        if query_cache_enabled && !(arel.respond_to?(:locked) && arel.locked)
           sql, binds, preparable, allow_retry = to_sql_and_binds(arel, binds, preparable, allow_retry)
 
           if async
@@ -280,7 +278,7 @@ module ActiveRecord
 
           result = nil
           @lock.synchronize do
-            result = @query_cache[key]
+            result = query_cache[key]
           end
 
           if result
@@ -299,7 +297,7 @@ module ActiveRecord
           hit = true
 
           @lock.synchronize do
-            result = @query_cache.compute_if_absent(key) do
+            result = query_cache.compute_if_absent(key) do
               hit = false
               yield
             end
--- a/activerecord/test/cases/query_cache_test.rb
+++ b/activerecord/test/cases/query_cache_test.rb
@@ -300,7 +300,6 @@ class QueryCacheTest < ActiveRecord::TestCase
 
         thread_1_connection = ActiveRecord::Base.lease_connection
         ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
-        assert_cache :off, thread_1_connection
 
         started = Concurrent::Event.new
         checked = Concurrent::Event.new
@@ -323,17 +322,16 @@ class QueryCacheTest < ActiveRecord::TestCase
 
             ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
           }.call({})
+
+          assert_cache :off, thread_2_connection
         }
 
         started.wait
 
         thread_1_connection = ActiveRecord::Base.lease_connection
         assert_not_equal thread_1_connection, thread_2_connection
-        assert_cache :dirty, thread_2_connection
         checked.set
         thread.join
-
-        assert_cache :off, thread_2_connection
       }.call({})
 
       ActiveRecord::Base.connection_pool.connections.each do |conn|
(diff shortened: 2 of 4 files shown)


activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
```diff
@@ -147,12 +147,6 @@ def initialize(...)
             end
         end
 
-        def checkout_and_verify(connection)
-          super
-          connection.query_cache ||= query_cache
```
engwan, line 152: Storing the `query_cache` on checkout is problematic for pinned connections because:

1. `checkout_and_verify` is only called once for pinned connections. It's only on the first thread that checks out a connection
2. Even if we would set the value every time a checkout happens, it would still be incorrect because the shared connection object would only be pointing to one cache store. We want connections on separate threads to use different cache stores even though they use the same connection object.

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
```diff
@@ -209,15 +203,19 @@ def query_cache
         end
       end
 
-      attr_accessor :query_cache
```
engwan, line 212: I left the ability to manually set `@query_cache` as-is for compatibility.

Some tests also set the query cache of the connection manually.

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
```diff
@@ -209,15 +203,19 @@ def query_cache
...
+      attr_writer :query_cache
 
       def initialize(*)
         super
         @query_cache = nil
       end
 
+      def query_cache
+        @query_cache || pool.try(:query_cache)
```
engwan: I used `try` because `NullPool` does not have a query cache.

Using `pool.query_cache` here means that we're always consulting the query cache registry. So even if we're using the same connection object, we would use different cache stores if we're on different threads.

activerecord/test/cases/query_cache_test.rb
```diff
@@ -300,7 +300,6 @@ def test_query_cache_across_threads
 
         thread_1_connection = ActiveRecord::Base.lease_connection
         ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
-        assert_cache :off, thread_1_connection
```
engwan, line 303: `clear_active_connections!` disables the query cache but the cache store is still dirty. We can't `assert_cache :off` anymore because this would return the pool that is dirty.

I'm not even sure if disabling the query cache when connections are released is the right behavior. Maybe the intent is to just clear the query cache? Because when you checkout a connection again the query cache would still be disabled for the thread.

In any case, this is a separate issue.

activerecord/test/cases/query_cache_test.rb
```diff
@@ -323,17 +322,16 @@ def test_query_cache_across_threads
...
           }.call({})
+
+          assert_cache :off, thread_2_connection
         }
 
         started.wait
 
         thread_1_connection = ActiveRecord::Base.lease_connection
         assert_not_equal thread_1_connection, thread_2_connection
-        assert_cache :dirty, thread_2_connection
```
engwan, line 332: We can no longer assert on the cache store of `thread_2_connection` outside of thread 2 because this would point to a separate cache store.

We have `assert_cache :dirty` within the thread anyway so I think that should be enough.

engwan:
@byroot can you review this please?

The only failing tests are the activesupport JSON related specs that are also failing on main.

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
byroot: Just add it to `NullPool` if necessary. `try` is **never** what you want.

byroot:
This is quite touchy code, so I'll try to review it tomorrow morning when I'm more clear headed.

NB: I fixed the json issue on main and rebased your PR.

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
engwan: Sure, we can do that.

.tool-versions
```diff
@@ -0,0 +1,37 @@
+# NOTE: This file declares dependencies for GDK itself.
```
byroot: This shouldn't be committed. I'll remove it

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
```diff
@@ -209,15 +203,19 @@ def query_cache
...
-      attr_accessor :query_cache
+      attr_writer :query_cache
 
       def initialize(*)
         super
         @query_cache = nil
       end
 
+      def query_cache
+        @query_cache || pool.query_cache
```
byroot, line 214: I'm a bit concerned about this. `pool.query_cache` is:

```ruby
       def query_cache
          @thread_query_caches.compute_if_absent(ActiveSupport::IsolatedExecutionState.context) do
            Store.new(@query_cache_version, @query_cache_max_size)
          end
      end
```

Which then call into `WeakThreadKeyMap` which before Ruby 3.3.5 can cause perf issues.

I'd like to see if we could do that only for pinned connections. 👀

.tool-versions
engwan: Sorry, my bad.

activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb
engwan: > I'd like to see if we could do that only for pinned connections. 👀

I think it should be easy to do that. We can continue setting `@query_cache` when it's not a pinned connection so we don't fall into this code path. Let me check.

byroot:
Improved version: https://github.com/rails/rails/pull/55703

PR #55697, opened 2025-09-18
Add `image_processing` to the gemfile by default for active_storage

Earlopain:
### Motivation / Background

Right now, a new rails app emits a warning because this gem is eagerly loaded. This prevents that,
and also shows how to opt out of using it alltogether.

See https://github.com/rails/rails/pull/55303 and its trail for more context.

What do you think @zzak?

### Additional information

ActionText wants this gem present. Since it now starts out uncommented, that logic is no longer needed.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/engine.rb
+++ b/activestorage/lib/active_storage/engine.rb
@@ -112,7 +112,8 @@ module ActiveStorage
           when /image_processing/
             ActiveStorage.logger.warn <<~WARNING.squish
               Generating image variants require the image_processing gem.
-              Please add `gem 'image_processing', '~> 1.2'` to your Gemfile.
+              Please add `gem "image_processing", "~> 1.2"` to your Gemfile
+              or set `config.active_storage.variant_processor = :disabled`.
             WARNING
           else
             raise
--- a/actiontext/lib/generators/action_text/install/install_generator.rb
+++ b/actiontext/lib/generators/action_text/install/install_generator.rb
@@ -47,18 +47,6 @@ module ActionText
           "app/views/layouts/action_text/contents/_content.html.erb"
       end
 
-      def enable_image_processing_gem
-        if (gemfile_path = Pathname(destination_root).join("Gemfile")).exist?
-          say "Ensure image_processing gem has been enabled so image uploads will work (remember to bundle!)"
-          image_processing_regex = /gem ["']image_processing["']/
-          if File.readlines(gemfile_path).grep(image_processing_regex).any?
-            uncomment_lines gemfile_path, image_processing_regex
-          else
-            run "bundle add --skip-install image_processing"
-          end
-        end
-      end
-
       def create_migrations
         rails_command "railties:install:migrations FROM=active_storage,action_text", inline: true
       end
--- a/guides/source/active_storage_overview.md
+++ b/guides/source/active_storage_overview.md
@@ -41,12 +41,6 @@ will not install, and must be installed separately:
 * [ffmpeg](http://ffmpeg.org/) v3.4+ for video previews and ffprobe for video/audio analysis
 * [poppler](https://poppler.freedesktop.org/) or [muPDF](https://mupdf.com/) for PDF previews
 
-Image analysis and transformations also require the `image_processing` gem. Uncomment it in your `Gemfile`, or add it if necessary:
-
-```ruby
-gem "image_processing", ">= 1.2"
-```
-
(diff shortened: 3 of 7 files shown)


activestorage/lib/active_storage/engine.rb
```diff
@@ -112,7 +112,8 @@ class Engine < Rails::Engine # :nodoc:
           when /image_processing/
             ActiveStorage.logger.warn <<~WARNING.squish
               Generating image variants require the image_processing gem.
-              Please add `gem 'image_processing', '~> 1.2'` to your Gemfile.
+              Please add `gem "image_processing", "~> 1.2"` to your Gemfile
```
Earlopain, line 115: Small cosmetic change with the quotes here but it matches what is generated in the gemfile and the error message in `image_processing_transformer.rb`

rafaelfranca:
> Right now, a new rails app emits a warning because this gem is eagerly loaded.

Is eagerly loaded by who? That is what is broken.

rafaelfranca:
Now I remember the reason. This makes sense.

zzak:
I had added it in #54166 but I think I forgot it and you could disable it to turn off the warnings if you wanted. Both are fine IMO.

Earlopain:
https://github.com/rails/rails/pull/55303#issuecomment-3311468431 

`image_processing` depends on `ruby-vips` which forces users to have libvips installed. I think that exception should be caught here as well: https://github.com/rails/rails/blob/1f3e490fa2b9b0652e1c15262465b0533bd8d3f1/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb#L15

So this was not 100% complete yet. I will try to open a PR in a bit to fix this.

Earlopain:
Actually, I am unsure

Both vips and image_magick requires extra setup from the user. You can rescue the errors during require but you end up with transformer classes defined that are in a broken state when the require fails for whatever reason.

They take the image blob, try to analyze it but `Vips::Image` is not defined. Not great. I think this still needs more work but no idea what that would actually be.

Seems like the default should actually be `disabled`? So that the user has to actually set it to vips themselves and handle the setup from that. For now, I have no better idea.

I think doing so would line up with previous behavior where `image_processing` was not part of the gemfile by default and analyzer were not eager loaded.

Perhaps this whole chain of PRs should not be part of 8.1 to give it more time still.

claudiob:
Hello there. 👋  Should this (or one of its related PRs like #55716 #55303) be considered a breaking change? Let me explain.

---

My Rails app runs on edge, uses Active Storage, does not use variants, and doesn't have `config.active_storage.variant_processor` in its configuration.

Before, users would upload a picture and no AnalyzeJob was triggered. The log would simply say:

```
ActiveStorage::Blob Update (1.4ms)  UPDATE "active_storage_blobs" SET "metadata" = '{"identified":true,"analyzed":true}' WHERE "active_storage_blobs"."id" = 66
```

After running `bundle update` to get to the latest commit of Rails, the behavior implicitly changed.
AnalyzeJob is triggered and (since I don't have `image_processing` or `vips` in my dependencies) causes a production error:

```
[ActiveJob] Enqueued ActiveStorage::AnalyzeJob ...
[ActiveJob] [ActiveStorage::AnalyzeJob] Performing ActiveStorage::AnalyzeJob ...
[ActiveJob] [ActiveStorage::AnalyzeJob] [0efa0f81-23e0-4a09-9fb2-55ecb484c10a] Error performing ActiveStorage::AnalyzeJob: NameError (uninitialized constant Vips):
```

---

This PR is good because it tells new Rails projects to either include `image_processing` or explicitly set `config.active_storage.variant_processor = :disabled`. In fact, after setting that, the production error was gone.

Could other developers in a similar situation (using Active Storage without variants) also be affected by this change with possible runtime errors? Just thinking out loud whether this is indeed a breaking change or not… 

Thank you 🙇

zzak:
I believe that is the behavior you would get if you used Rails 8 with Active Storage, which is why some apps were using `config.active_storage.analyzers = []` to disable them.

If you have time can you try to reproduce it in a [bug_report](https://github.com/rails/rails/blob/main/guides/bug_report_templates/active_storage.rb) and open a new ticket if so?

I might have time to debug this weekend :bow:

Earlopain:
I think you'd find a warning somewhere during boot that image_processing was not found. The analyzers are still present I suspect and maybe rails should clear them out if it failed to load the necessary gems. Them being present is what causes this since they now no longer check themselves if the gems like libvips are loadable.

The default of `variant_processor` being `vips` is highly unfortunate now that things are eager loaded. I strongly believe that it has to be `disabled`, it would make everything so much easier. I think this is underlined above where even `vips` used to do absolutely nothing before.

PR #55703, opened 2025-09-19
Fix query cache for pinned connections in multi threaded transactional tests

### LINKED ISSUE #55689: Query cache does not work in system tests in Rails 7.2+

byroot:
Fix: https://github.com/rails/rails/issues/55689
Fix: https://github.com/rails/rails/pull/55696

When a pinned connection is used across separate threads, each thread should have its own query cache store.

@engwan

### DIFF
--- a/activerecord/test/cases/query_cache_test.rb
+++ b/activerecord/test/cases/query_cache_test.rb
@@ -729,16 +729,28 @@ class QueryCacheTest < ActiveRecord::TestCase
       ActiveRecord::Base.lease_connection.enable_query_cache!
       assert_cache :clean
 
+      main_thread_cache = ActiveRecord::Base.lease_connection.query_cache
+      assert_same main_thread_cache, ActiveRecord::Base.lease_connection.query_cache
+
       thread_a = Thread.new do
         middleware { |env|
           assert_cache :clean
+
+          # In a background thread, the cache instance must stay consistent but be different from the main
+          # thread.
+          background_thread_cache = ActiveRecord::Base.lease_connection.query_cache
+          assert_same background_thread_cache, ActiveRecord::Base.lease_connection.query_cache
+          assert_not_same main_thread_cache, ActiveRecord::Base.lease_connection.query_cache
           [200, {}, nil]
         }.call({})
       end
 
       thread_a.join
+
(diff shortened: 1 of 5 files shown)


activerecord/test/cases/query_cache_test.rb
```diff
@@ -323,17 +322,16 @@ def test_query_cache_across_threads
...
         started.wait
 
         thread_1_connection = ActiveRecord::Base.lease_connection
         assert_not_equal thread_1_connection, thread_2_connection
-        assert_cache :dirty, thread_2_connection
         checked.set
         thread.join
-
-        assert_cache :off, thread_2_connection
       }.call({})
```
engwan: Since the new behavior only applies to pinned connections, we can actually restore this test to what it was and it should still pass.

Though I guess this is also fine.

engwan (commented):
Thanks @byroot. I think this looks good.

Just a small comment regarding the test.

activerecord/test/cases/query_cache_test.rb
byroot: Ah nice. Good catch.

byroot:
Probably should backport this to 8-0 and 7-2.

PR #55332, opened 2025-07-14
Clearing optional CPK in belongs_to should preserve PK

zzak:
Fixes: #54882

/cc @jarl-dk

PR #55707, opened 2025-09-19
Refactor BelongsToAssociation treatment of composite keys

zzak:
Based on feedback: https://github.com/rails/rails/pull/55332#discussion_r2215995573

```
First I think we should be able to get rid of if
reflection_fk.is_a?(Array) in favor of treating everything as an
Array(), even single-column PK/FKs.

And then perhaps we could semantically split it into two parts: one part
for the "replace" flow and the other one is "nullify" flow where record
is nil.
```

Opening this PR to see if it makes sense and any tests that might fail. :pray:

/cc @nvasilevski

### DIFF
--- a/activerecord/lib/active_record/associations/belongs_to_association.rb
+++ b/activerecord/lib/active_record/associations/belongs_to_association.rb
@@ -11,13 +11,8 @@ module ActiveRecord
         when :destroy
           raise ActiveRecord::Rollback unless target.destroy
         when :destroy_async
-          if reflection.foreign_key.is_a?(Array)
-            primary_key_column = reflection.active_record_primary_key
-            id = reflection.foreign_key.map { |col| owner.public_send(col) }
-          else
-            primary_key_column = reflection.active_record_primary_key
-            id = owner.public_send(reflection.foreign_key)
-          end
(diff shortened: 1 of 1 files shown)


zzak:
:green_circle: ~Some related failures will look into later~

activerecord/lib/active_record/associations/belongs_to_association.rb
```diff
@@ -72,23 +67,24 @@ def decrement_counters_before_last_save
           model_was = klass
         end
 
-        foreign_key_was = owner.attribute_before_last_save(reflection.foreign_key)
+        foreign_key_was = foreign_keys.map { |fk| owner.attribute_before_last_save(fk) }
 
-        if foreign_key_was && model_was < ActiveRecord::Base
+        if foreign_key_was && model_was && model_was < ActiveRecord::Base
```
rafaelfranca: `foreign_key_was` is going to be always true here, no?

zzak:
```suggestion
        if foreign_key_was.any? && model_was && model_was < ActiveRecord::Base
```

True, I think because `attribute_before_last_save` can return the value or nil(?) then this should work.

PR #37898, opened 2019-12-06
Revert "Use `app.config.file_watcher` for watcher in `RoutesReloader`"

tgxworld:
This reverts commit 28e44f472d1cd6853726f85eeb7623e5901c4d37.

A limitation of Listen is that it currently only supports watching directories.
Therefore, watching `config/routes.rb` will end up watching the entire `config` directory
if we use the evented file watcher. This causes problems especially if symlinks are present
in the `config` directory.

PR #55709, opened 2025-09-19
[Fix #55708] Respect the file_watcher config in the routes reloader

grodowski:
Follow-up to #55708, where using `EventedFileUpdateChecker` also proved to be a possible workaround, but we found out that it's hardcoded inside `RoutesReloader`.

### Checklist

### DIFF
--- a/railties/lib/rails/application/routes_reloader.rb
+++ b/railties/lib/rails/application/routes_reloader.rb
@@ -11,12 +11,13 @@ module Rails
       attr_writer :run_after_load_paths, :loaded # :nodoc:
       delegate :execute_if_updated, :updated?, to: :updater
 
-      def initialize
+      def initialize(file_watcher: ActiveSupport::FileUpdateChecker)
         @paths      = []
         @route_sets = []
         @external_routes = []
         @eager_load = false
         @loaded = false
+        @file_watcher = file_watcher
       end
 
       def reload!
@@ -48,7 +49,7 @@ module Rails
             hash[dir.to_s] = %w(rb)
           end
 
-          ActiveSupport::FileUpdateChecker.new(paths, dirs) { reload! }
+          @file_watcher.new(paths, dirs) { reload! }
         end
       end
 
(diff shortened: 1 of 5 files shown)


railties/lib/rails/application/routes_reloader.rb
```diff
@@ -48,7 +48,7 @@ def updater
             hash[dir.to_s] = %w(rb)
           end
 
-          ActiveSupport::FileUpdateChecker.new(paths, dirs) { reload! }
+          Rails.application.config.file_watcher.new(paths, dirs) { reload! }
```
rafaelfranca: We should pass the file watcher down to this class instead of using the Rails.application here.

rafaelfranca:
See https://github.com/rails/rails/pull/37898.

Was that problem fixed?

grodowski:
> See https://github.com/rails/rails/pull/37898.
> Was that problem fixed?

Thanks @rafaelfranca! I wasn't aware that this approach has been tried before without success. Based on this revert PR message:

> A limitation of Listen is that it currently only supports watching directories.
Therefore, watching config/routes.rb will end up watching the entire config directory
if we use the evented file watcher. This causes problems especially if symlinks are present
in the config directory.

it doesn't seem that anything has changed in `listen` since then? 

- `listen` does support `ignore` and `only` [filters](https://github.com/guard/listen/blob/9c21ea7ae71c3c56495763f3ecef1bf2150c5b7a/lib/listen/change.rb#L42), but IIUC they do not change the fact that the underlying watches are per directory
- `listen`'s filters are not used by the `EventedFileUpdateChecker`, which has it's own filtering in [`watching?`](https://github.com/rails/rails/blob/cea7cae6ef90bf33abfe450ba65c3022a1f07fc2/activesupport/lib/active_support/evented_file_update_checker.rb#L152-L172)
- I assume problems with symlinks can get more pronounced with`config/` as with other watched directories and it's irrelevant to how `listen` or the watcher in Rails filter the changes

I was wondering if switching `EventedFileUpdateChecker` to use `listen`'s `only` instead of filtering later with `watching?` would make any difference, but my guess is that it's unlikely. Also depends on what specifically caused the revert.

Do you have any additional context and thoughts on what we could do with this PR. Should I close it?

rafaelfranca:
Let's try this again to see which problems we will find

PR #55713, opened 2025-09-19
Dedupe schema dumps and refactor test helpers

mikeygough:
### Motivation / Background

In working on a large application, @skipkayhil and I noticed that occasionally `db:schema:dump` will dump the same schema more than once. We can speed up `db:migrate` by first checking if a schema has already been dumped before dumping.

Testing these changes on a large application I noticed a roughly 40% improvement in running `bin/rails db:schema:dump`.

This Pull Request has been created because `db:schema:dump` is occasionally dumping the same schema more than once. We can increase performance by checking for duplicate schema dumps before dumping.

While adding test coverage I also noticed several copy-pasted function definitions to stub configurations. I refactored these into a single helper method inside the `DatabaseTasksHelper` module which can be included in the test classes that need it. That full refactoring is available [here](https://github.com/rails/rails/pull/55732) and should be shipped first. Then we can rebase this and ship the schema dump improvements.

### Detail

This Pull Request changes the `dump_all` method to first check if a schema has already been dumped before calling `dump_schema`.

### Additional information

In testing these changes against a large application, I noticed the following time differences in calling `db:schema:dump`. Please note that these performance gains will vary depending on the size of an application's schema.

#### No-deduplication
no-deduplication

#### Deduplication
deduplication

### Checklist

### DIFF
--- a/activerecord/lib/active_record/tasks/database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/database_tasks.rb
@@ -428,9 +428,15 @@ module ActiveRecord
       end
 
       def dump_all
-        with_temporary_pool_for_each do |pool|
-          db_config = pool.db_config
+        seen_schemas = []
+
+        ActiveRecord::Base.configurations.configs_for(env_name: ActiveRecord::Tasks::DatabaseTasks.env).each do |db_config|
+          schema_path = schema_dump_path(db_config, ENV["SCHEMA_FORMAT"] || db_config.schema_format)
+
+          next if seen_schemas.include?(schema_path)
+
           ActiveRecord::Tasks::DatabaseTasks.dump_schema(db_config, ENV["SCHEMA_FORMAT"] || db_config.schema_format)
+          seen_schemas << schema_path
         end
       end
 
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/tasks/database_tasks.rb
```diff
@@ -428,9 +428,14 @@ def reconstruct_from_schema(db_config, file = nil) # :nodoc:
       end
 
       def dump_all
+        seen_schemas = []
+        
         with_temporary_pool_for_each do |pool|
           db_config = pool.db_config
+          next if seen_schemas.include?(db_config.schema_dump)
```
skipkayhil: Reading through `dump_schema` again, I'm wondering if we should handle the `filename = schema_dump_path` case 🤔

I think we would not be de-duping right now if one database was configured with `db/structure.sql` and another was `structure.sql`, but `schema_dump_path()` would end up coercing them both into `db/structure.sql`.

mikeygough: Great catch! 

Think I've got this handled now where we compare paths:

```Ruby
schema_path = schema_dump_path(db_config, ENV["SCHEMA_FORMAT"] || db_config.schema_format)

next if seen_schemas.include?(schema_path)
```

Added another test case for this as well.

eileencodes (approved):

PR #55716, opened 2025-09-20
Swallow error if libvips or ruby-vips gem are missing

zzak:
Follow up to #55303

The previous behavior would be to raise an error only if the gem was missing, this would happen at Runtime before refactoring to startup.

But if trying to boot Rails without libvips installed, this will raise the following error:

```
/home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ffi-1.17.2/lib/ffi/dynamic_library.rb:94:in 'FFI::DynamicLibrary.load_library': Could not open library 'vips.so.42': vips.so.42: cannot open shared object file: No such file or directory. (LoadError)
Could not open library 'libvips.so.42': libvips.so.42: cannot open shared object file: No such file or directory.
Searched in <system library path>
        from /home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ffi-1.17.2/lib/ffi/library.rb:95:in 'block in FFI::Library#ffi_lib'
        from /home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ffi-1.17.2/lib/ffi/library.rb:94:in 'Array#map'
        from /home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ffi-1.17.2/lib/ffi/library.rb:94:in 'FFI::Library#ffi_lib'
        from /home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ruby-vips-2.2.5/lib/vips.rb:46:in '<module:Vips>'
        from /home/zzak/.rbenv/versions/3.5.0/lib/ruby/gems/3.5.0+4/gems/ruby-vips-2.2.5/lib/vips.rb:43:in '<top (required)>'
```

However, because this happens when loading the engine, I think the error was swallowed by Rails startup, and all of the other config setup at the start of loading the engine was not evaluated, so parts like:

```
config.active_storage.paths = ActiveSupport::OrderedOptions.new
config.active_storage.queues = ActiveSupport::InheritableOptions.new
```

Which meant when booting a new Rails app without libvips installed on the system, you actually got this error instead:

```
NoMethodError: undefined method 'analysis=' for nil (NoMethodError)

            active_storage.queues.analysis = :active_storage_analysis
```

If you did manage to boot the app, you could see the other defaults were not setup yet.

```
$ bin/rails c
Loading development environment (Rails 8.1.0.beta1)
>> ActiveStorage.variable_content_types
=> []
```

/cc @jeromedalbert @Earlopain

### DIFF
--- a/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb
+++ b/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb
@@ -12,7 +12,7 @@ begin
   gem "ruby-vips"
   require "ruby-vips"
 rescue LoadError => error
-  raise error unless error.message.include?("ruby-vips")
+  raise error unless error.message.match?(/libvips|ruby-vips/)
 end
 
 module ActiveStorage


Earlopain:
But then you have warning by default again if libvips is missing. Why not swallow the error, then you're just back where when `image_processing` was not available?

I would be ok with a warning/error if the user has to opt in but that isn't the case, so potentially very many will see this. It's also not an active storage logger, so you don't even have to set a log level to see get this.

zzak:
@Earlopain Thanks for the quick feedback! I agree, and swallowing the error like before is the least disruptive behavior and don't expect to get errors until actually trying to use ASto to analyze images or create variants.

byroot:
I'm sorry but I think I lost the plot a bit here.

Why shouldn't this error bubble up? This code is only loaded if you have configured vips to be the image backend, if so we should fail on boot not at runtime, just like how we should fail to boot if you have configured postgresql as your database and we can't load the `pg` gem.

Or am I missing something?

Earlopain:
The problem is that vips is the default. I have a comment in https://github.com/rails/rails/pull/55697#issuecomment-3311596325 that explains some of the current problems I see

zzak:
> we should fail on boot not at runtime

I was thinking about it like this: if the user has the right libraries installed they get the benefit of loading the gem at boot, not during a request.

So the DX of not having the proper dependencies installed is the same as it was before, lazily.

And thus a Day 1 rails dev running `rails new` doesn't get an error about libvips for ActiveStorage which they may not even know they will use yet.

I can see the arugment for raising on boot if dependencies are not met, but I actually think we should delay referencing / loading those gems until an initializer: https://github.com/rails/rails/blob/1f3494d6327eb7f9025862bc3362cbed97b668c1/activestorage/lib/active_storage/engine.rb#L27-L28

But that would be a breaking change, sort of.

byroot:
> The problem is that vips is the default.

I see, but it's also in the default Gemfile, no?

Earlopain:
> I see, but it's also in the default Gemfile, no?

Yes, through `image_processing`. But libvips is not.

To be clear, https://github.com/rails/rails/pull/55697 is partially at fault for exposing this, since it actually started to pull `ruby-vips` in. But practically this is the same experience that actiontext users would have gone through (it uncomments the line in the gemfile as part of the generator), or just for those users uncommenting it themselves.

I imagine this is also the same experience for people before this all changed. But then it was not eagerloaded and only happened when they actually used the functionality.

zzak:
Probably ruby-vips should not install if libvips is not on the system, but that is another issue. I guess they are using FFI, so that explains it.

byroot:
Ok, I'm gonna merge as is because it's better than the current state, and I can't immediately think of a good solution.

But IMO there is a sticky situation to untangle here:

  - When we know during boot something won't work at runtime, we shouldn't boot. Failing to boot is annoying (fail deploy), but failing at runtime is 1000% worse (can cause outage).
  - Problem is Active Storage is enabled by default, and its default backend is vips, but plenty of people might not actually use Active Storage, so I get that it's annoying to force them to install libvips to even boot.

I really don't like this situation.

PR #55717, opened 2025-09-20
Set pipefail for rails-new-docker step

zzak:
This step was previously failing but still marked as green: https://github.com/rails/rails/actions/runs/17870596398/job/50823217427#step:5:200

### DIFF
--- a/.github/workflows/rails-new-docker.yml
+++ b/.github/workflows/rails-new-docker.yml
@@ -24,6 +24,7 @@ jobs:
         bundler-cache: true
     - name: Generate --dev app
       run: |
+        set -euo pipefail
         bundle exec railties/exe/rails new $APP_PATH --dev
     - name: Build image
       run: |


zzak:
Ok, nevermind. I understand now.

AppBase generator allows commands to fail (`abort_on_failure`) by default, I guess so `importmap:install` fails you don't end up with a half-baked app.

Changing that seems more difficult, questionable, so while it'd be nice to find out if `rails new` broke this way sooner, I don't have a good answer now.

skipkayhil:
Maybe we could repurpose the list of expected generated files from the app_generator_test.rb?

zzak:
@skipkayhil So we could call this from within the GH workflow you mean?

```
    assert_empty generated_files_and_folders.difference(default_files)
    assert_empty default_files.difference(generated_files_and_folders)
    assert_equal generated_files_and_folders.sort, default_files, "The expected list of generated files is not alphabetical"

    default_files.each { |path| assert_file path }
```

PR #55718, opened 2025-09-20
Restore support to analyze image blobs with `image_magick`

Earlopain:
### Motivation / Background

Since https://github.com/rails/rails/commit/93c00a8c74c24a3da40d30b6d0c3c667b8ebe9ba, `config.active_storage.analyzers` contains both vips and image magick analyzers. However, they both inherit from `Analyzer::ImageAnalyzer`, which means they just check for `blob.image?` to check if they should accept the image.

That means that even if the user explicitly does `variant_processor = :mini_magick`, vips gets called first (because it comes first in the array), and image magick never gets a chance to do anything.

Before that commit it worked because the analyzers were conditionally added to the array, so there was only ever one or the other present.

This fixes this by once again implementing custom `accept?`, which was previously removed in https://github.com/rails/rails/commit/e978ef011f11102771c5b302f25d9ab7895316cb

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/analyzer/image_analyzer/image_magick.rb
+++ b/activestorage/lib/active_storage/analyzer/image_analyzer/image_magick.rb
@@ -11,6 +11,10 @@ module ActiveStorage
   # This analyzer relies on the third-party {MiniMagick}[https://github.com/minimagick/minimagick] gem. MiniMagick requires
   # the {ImageMagick}[http://www.imagemagick.org] system library.
   class Analyzer::ImageAnalyzer::ImageMagick < Analyzer::ImageAnalyzer
+    def self.accept?(blob)
+      super && ActiveStorage.variant_processor == :mini_magick
+    end
+
     private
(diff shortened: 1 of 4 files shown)


zzak (approved):
Good catch, and sorry for missing this.

Just to confirm, the issue was simply:

* Image analyzers were made optional, based on the variant processor in:
  * [e978ef0](https://github.com/rails/rails/commit/e978ef011f11102771c5b302f25d9ab7895316cb#diff-bbfba04c81dcdc406986d3697ddc362cabcbd130b8888ed7752d074dac1ba8c6L7-L17)
* When we added them all back, we forgot the add `accept?` in:
  * [93c00a8](https://github.com/rails/rails/commit/93c00a8c74c24a3da40d30b6d0c3c667b8ebe9ba#diff-c83e57b202f71e5fdfb58f3df8bb3c8ad7886b7766fc5e5fefdd870fb04b249cR28-R30)

PR #55720, opened 2025-09-20
hidde backtrace when parallel tests get interrupted

aronwolf90:
### Motivation / Background

Right now, if one presses CONTROL+C during the parallel test run, one gets a very long backtrace that is very annoying.

### Additional information

The same problem was solved 11 years ago for the non parallel tests here:

* https://github.com/minitest/minitest/commit/b6ec36d086ac28e47e4e2dc7ffd995f462543f94
* https://github.com/minitest/minitest/issues/503

The reason why the 11 years old fix does not work for the parallel test is, that the rescue happens before the shutdown method gets executed. The shutdown method is where things get executed. In my opinion, this is wrong, but my guess is, that it was done so, because doing it otherwise would probably require changes in the Minitest gem itself.

NOTE: this is my first PR on the rails gem, so I apologize if I made something wrong. Also, there is a big probability that there is a better solution or that I miss some edge case that I am missing because of lack of knowledge.

### Checklist

  - Not sure if this should go into the changelog.

### DIFF
--- a/activesupport/lib/active_support/testing/parallelization/server.rb
+++ b/activesupport/lib/active_support/testing/parallelization/server.rb
@@ -77,10 +77,7 @@ module ActiveSupport
 
           @queue.close
 
-          # Wait until all workers have finished
-          while active_workers?
-            sleep 0.1
-          end
+          wait_for_active_workers
 
           @in_flight.values.each do |(klass, name, reporter)|
             result = Minitest::Result.from(klass.new(name))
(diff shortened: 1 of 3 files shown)


aronwolf90:
@joshuay03 I sow that you worked on something similary. Maybe you can review the PR? 

NOTE:I totaly understand if you have no time for it. At the end removing the backtrace is just a cosmetic thing :) and there are probably a lot more important things to fix than this.

joshuay03:
> @joshuay03 I sow that you worked on something similary. Maybe you can review the PR?
> 
> NOTE:I totaly understand if you have no time for it. At the end removing the backtrace is just a cosmetic thing :) and there are probably a lot more important things to fix than this.

Yes, I can have a look soonish. Please note though that maintainers prefer not to be tagged directly, and you should follow the process documented in the contributing guides:
- https://github.com/rails/rails/blob/main/CONTRIBUTING.md
- https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#get-some-feedback

joshuay03:
Thank you!! @rafaelfranca

aronwolf90:
Thanks! :partying_face:

PR #55722, opened 2025-09-21
Fix lease_connection to preserve pool state when checkout callbacks fail

kozy4324:
### Motivation / Background

This Pull Request addresses a connection pool state issue that can cause `ActiveRecord::Base.connection` to return `nil`.

For example, in Rails console:

```
my-app(dev)> p = ->{raise "some error"}
=> #<Proc:0x000000011f6bab78 (blog-app):1 (lambda)>

my-app(dev)> ActiveRecord::ConnectionAdapters::SQLite3Adapter.set_callback(:checkout, :after, p)
=> [ActiveRecord::ConnectionAdapters::SQLite3Adapter]

my-app(dev)> ActiveRecord::Base.connection
(blog-app):1:in 'block in <main>': some error (RuntimeError)
	from (blog-app):3:in '<main>'

my-app(dev)> ActiveRecord::Base.connection
=> nil
```

This happens because `ConnectionHandling#connection` expects `active_connection` to return a usable connection when `pool.permanent_lease?` is false.

Ref: https://github.com/rails/rails/blob/4fcc1dd2e23e76a31a6f6b6f7bd5b0ebfa87e1fe/activerecord/lib/active_record/connection_handling.rb#L276-L294

The root cause is that `ConnectionPool#connection_lease` previously marked the lease as sticky before assigning `lease.connection`.
When a checkout callback raises an exception, `lease.connection` remains `nil` while `lease.sticky` is `true`.

This is considered a bug because `connection_lease` objects are not expected to have `sticky = true` while `connection = nil`.
This inconsistent state propagates to `active_connection`, causing `ActiveRecord::Base.connection` to unexpectedly return `nil`.

### Detail

This Pull Request changes `ConnectionPool#lease_connection` to assign the `sticky` flag after successfully checking out a connection.

Additionally, a test has been added to ensure that exceptions in checkout callbacks do not affect `permanent_lease?` or `active_connection?` state.

### Checklist

### DIFF
--- a/activerecord/test/cases/connection_pool_test.rb
+++ b/activerecord/test/cases/connection_pool_test.rb
@@ -1477,6 +1477,25 @@ module ActiveRecord
         pool&.disconnect!
       end
 
+      def test_checkout_callback_error_does_not_affect_permanent_lease_and_active_connection_state
+        checkout_error = StandardError.new("error during checkout")
+        proc_to_raise = -> { raise checkout_error }
+        ActiveRecord::ConnectionAdapters::AbstractAdapter.set_callback(:checkout, :after, proc_to_raise)
+
+        assert_predicate @pool, :permanent_lease?
+        assert_not_predicate @pool, :active_connection?
+
+        error = assert_raises StandardError do
+          @pool.lease_connection
+        end
+        assert_same checkout_error, error
+
+        assert_predicate @pool, :permanent_lease?
+        assert_not_predicate @pool, :active_connection?
+      ensure
+        ActiveRecord::ConnectionAdapters::AbstractAdapter.skip_callback(:checkout, :after, proc_to_raise)
+      end
+
       private
         def active_connections(pool)
           pool.connections.find_all(&:in_use?)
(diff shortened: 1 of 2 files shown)


activerecord/test/cases/connection_pool_test.rb
```diff
@@ -1477,6 +1477,24 @@ def test_inspect_does_not_show_secrets
...
+        proc_to_raise = -> { raise msg }
+        ActiveRecord::ConnectionAdapters::AbstractAdapter.set_callback(:checkout, :after, proc_to_raise)
+
+        assert_predicate @pool, :permanent_lease?
+        assert_not_predicate @pool, :active_connection?
+        begin
+          @pool.lease_connection
+        rescue Exception => e
+          assert_equal msg, e.message
+        end
```
byroot:
```suggestion
        begin
          @pool.lease_connection
        rescue Exception => e
          assert_equal msg, e.message
        end
```

This should use `assert_raise`.

PR #55726, opened 2025-09-22
Add fast-path for Object#to_json(escape: false)

etiennebarrie:
With the new defaults from 8.1, we call Object#to_json with `escape: false`:
https://github.com/rails/rails/blob/80827ca7ff33ca3026f2c74fe8249ccc6f513318/actionpack/lib/action_controller/metal/renderers.rb#L171

This results in calling `as_json`:
https://github.com/rails/rails/blob/80827ca7ff33ca3026f2c74fe8249ccc6f513318/activesupport/lib/active_support/json/encoding.rb#L169

which negates the performance gains of not escaping by causing the tree to be walked twice.

### DIFF
--- a/activesupport/lib/active_support/json/encoding.rb
+++ b/activesupport/lib/active_support/json/encoding.rb
@@ -46,6 +46,8 @@ module ActiveSupport
       def encode(value, options = nil)
         if options.nil? || options.empty?
           Encoding.encode_without_options(value)
+        elsif options == { escape: false }.freeze
+          Encoding.encode_without_escape(value)
         else
           Encoding.json_encoder.new(options).encode(value)
         end
@@ -164,7 +166,14 @@ module ActiveSupport
 
 
           def initialize(options = nil)
-            @options = options ? options.dup.freeze : {}.freeze
+            if options
+              options = options.dup
+              @escape = options.delete(:escape) { true }
+              @options = options.freeze
+            else
+              @escape = true
+              @options = {}.freeze
+            end
           end
 
           # Encode the given object into a JSON string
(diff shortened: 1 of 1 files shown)


ghiculescu:
@etiennebarrie this has introduced a regression with some versions of the `json` gem:

I'm scratching my head as to why, but it might be obvious to you.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "json", "2.12" # fail
  gem "json", "2.13" # fail
  # gem "json", "2.14" # pass
  # gem "json", "2.15" # pass

  # gem "rails", "~> 8" # pass
  gem "rails", github: "rails/rails", branch: "main", ref: "7898364a3879862a0802a1e3d6060eb220e49f76" # fail
  # gem "rails", github: "rails/rails", branch: "main", ref: "4e72506c57a9ad570136b0a169d1b1612d9f24b8" # pass (commit before this one)

  gem "sqlite3"
  gem "debug"
end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

  config.load_defaults("7.2")
end
Rails.application.initialize!

  get "/", to: "test#index"
end

    t.timestamps
  end
end

end

  def index
    posts = Post.all
    obj = { posts: posts }
    render json: obj
  end
end

  include Rack::Test::Methods

  def test_returns_success
    3.times { Post.create! }
    get "/"
    puts last_response.body
    assert_equal 3, JSON.parse(last_response.body)["posts"].length
  end

  private
    def app
      Rails.application
    end
end

```

When it passes, the output looks like

```
{"posts":[{"id":1},{"id":2},{"id":3}]}
```

And when it fails it looks like

```
{"posts":"{\"id\" => 1}"}
```

ghiculescu:
Maybe it's not properly compatible with `json` pre https://github.com/ruby/json/pull/864? 🤔  cc @byroot

byroot:
Hum, I thought we addressed this, but apparently not. Thanks for the report, I'll squeeze a quick version check.

PR #55734, opened 2025-09-23
Make CodeStatistics pattern overridable

n-rodriguez:
Make `CodeStatistics` pattern overridable so user can add `haml` file extension.

Thank you!


n-rodriguez:
@byroot Thank you!

Issue #55728, opened 2025-09-22
7.1 -> 7.2 perf regression due to non permanent ActiveRecord connections


cmaion:
I observe a ~5% increase in response time since migrating to Rails 7.2.

This regression stems from the changes introduced with `ActiveRecord::Base.with_connection` and `ActiveRecord::Base.lease_connection`.

Indeed, as there's no permanent AR connection anymore, my app now leases then releases it's connection, at least once for **every ActiveRecord query**.

Traced  those `with_connection` calls from `ActiveRecord::Core::ClassMethods#cached_find_by`, `ActiveRecord::Associations::AliasTracker.create`, `AssociationReflection#association_scope_cache`, `ActiveRecord::Associations::Association#find_target`...

This also triggers many, many calls to `checkin` callbacks per request, which is quite wasteful.

I get that the change can be beneficial on busy heavily multi-threaded workers, but I don't see the benefit on single-threaded web workers (`unicorn`) or when the connection pool is large enough for all threads (and when the database able to deal with all those concurrent connections).

Adding a call to `ActiveRecord::Base.lease_connection` as the first `before_action` in the controllers that requires a connection to the DB restores the performance back to Rails 7.1 level.

Is this new default behavior expected?
Should my explicit `lease_connection` hack in my controllers be needed (and does it have a drawback I didn't get?) or did I miss something that would pin the connection at first use?

Thanks!

cc @byroot 

### System configuration

**Rails version**: 7.2+

**Ruby version**: 3.4

cmaion:
Related to https://github.com/rails/rails/pull/53662 I guess

byroot:
Supposedly the fix is https://github.com/rails/rails/pull/53672. It has been backported to `7-2-stable` but not yet released.

Please try pointing your app at that branch:

```ruby
gem "rails", github: "rails/rails", branch: "7-2-stable"
```

cmaion:
Hi,

Just switched to `7-2-stable`, issue persists.

I guess fix https://github.com/rails/rails/pull/53672 avoids redundant `pings`, but the connection still constantly goes back and forth between the pool and the worker (and triggers `checkin` callbacks each time).

I can verify this by adding a `Rails.logger.debug "Checkin connection"` into `vendor/bundle/ruby/3.4.0/bundler/gems/rails-f96593cdd05c/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#checkin` (line 590, right after `return if @pinned_connection.equal?(conn)`)

[...]

PR #55736, opened 2025-09-23
Micro-optimize Active Record connection checkin

byroot:
Ref: https://github.com/rails/rails/issues/55728

Starting in 7.2, since we no longer pin Active Record connection for the entirely of the request or job cycle, `Pool#checkout` and `Pool#checkin` overhead has to be paid for every query. It's not a whole lot once you consider a query will need to do a network hop, but we should still try to keep that overhead low.

This PR contains many different small optimizations, many of them too small to show a clear gain on a benchmark, but together they add up, see individual commits. (FYI: @cmaion).

Which brings up back to about half-way between 7.1 and `main`:

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
          7-1-stable     3.219k i/100ms
          7-1-stable     88.512k (± 4.1%) i/s   (11.30 μs/i) -    444.222k in   5.027300s

Comparison:
     ar-fast-checkin:    79704.1 i/s
          7-1-stable:    88512.0 i/s - 1.11x  faster
                main:    70279.4 i/s - 1.13x  slower
```

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "."
  gem "sqlite3"
  gem "benchmark-ips"
end

require "benchmark/ips"

  config.eager_load = false
  config.logger = Logger.new(nil)
  config.log_level = :info

end
Rails.application.initialize!


  end

    t.integer :post_id
  end
end

ActiveRecord::Base.release_connection unless Rails::VERSION::STRING < "7.2"

end

BRANCH = `git rev-parse --abbrev-ref HEAD`.strip

Benchmark.ips do |x|
  x.report(BRANCH) { Post.count }
  x.save!("/tmp/bench-ar-data")
  x.compare!(order: :baseline)
end
```

There might be some more optimizations to perform, but I think this is now large enough to be worth merging.


activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
```diff
@@ -179,21 +179,29 @@ def clear(connection)
         end
       end
 
-      class LeaseRegistry # :nodoc:
-        def initialize
-          @mutex = Mutex.new
-          @map = WeakThreadKeyMap.new
+      if RUBY_ENGINE == "ruby"
+        class LeaseRegistry < WeakThreadKeyMap
```
rafaelfranca: Can we add a comment about the GVL?

activesupport/lib/active_support/callbacks.rb
```diff
@@ -905,12 +905,13 @@ def define_callbacks(*names)
...
-            ([self] + self.descendants).each do |target|
-              target.set_callbacks name, CallbackChain.new(name, options)
-            end
-
             module_eval <<-RUBY, __FILE__, __LINE__ + 1
-              def _run_#{name}_callbacks(&block)
+              def _run_#{name}_callbacks
+                yield if block_given?
+              end
+              alias_method :_run_#{name}_callbacks, :_run_#{name}_callbacks
```
rafaelfranca: I didn't get the alias to the same name. Why is that?

activerecord/lib/active_record/runtime_registry.rb
```diff
@@ -7,76 +7,80 @@ module ActiveRecord
...
-  end
-
-  runtime = (finish - start) * 1_000.0
-
-  if payload[:async]
-    ActiveRecord::RuntimeRegistry.async_sql_runtime += (runtime - payload[:lock_wait])
-  end
-  ActiveRecord::RuntimeRegistry.sql_runtime += runtime
-end
+# ActiveSupport::Notifications.monotonic_subscribe("sql.active_record") do |name, start, finish, id, payload|
```
rafaelfranca: Should this be commented out?

activesupport/lib/active_support/callbacks.rb
byroot: It's a classic trick to silence redefinition warnings: https://github.com/rails/rails/blob/529f933fc8b13114d308dd0752f76a9e293c8537/activesupport/lib/active_support/core_ext/module/redefine_method.rb#L7-L13

I could perhaps use that for clarity.

activerecord/lib/active_record/runtime_registry.rb
byroot: No, it's just temporary to measure how much we can save in a case where this is no active `sql.active_record` subscribers. The answer is quite a lot, but then how realistic such a scenario is, I'm not sure.

I suspect it's very common for apps to have some APM or similar so probably at least one subscriber here.

The root of the problem is that `ActiveSupport::Notifications` really isn't as cheap as I'd like, but it already has been optimized quite far, I doubt we can optimize it much more while keep perfect compatibility.

activerecord/lib/active_record/explain_subscriber.rb
```diff
@@ -29,6 +29,10 @@ def ignore_payload?(payload)
         !payload[:sql].match?(EXPLAINED_SQLS)
     end
 
+    def silenced?(event)
+      !ExplainRegistry.collect?
+    end
```
byroot: Better than silencing it, we should just only subscribe on the event when we're collecting.

This means we'd flush the AS::N cache, but I think it's OK as you'd only realistically do this in development.

byroot:
Ok, so this branch currently reclaims a large part of the lost performance:

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
     ar-fast-checkin     7.781k i/100ms
     ar-fast-checkin     77.041k (± 2.8%) i/s   (12.98 μs/i) -    389.050k in   5.053953s

Comparison:
          7-1-stable:    87392.0 i/s
     ar-fast-checkin:    77040.6 i/s - 1.13x  slower
                main:    67712.3 i/s - 1.29x  slower
```

There might be some more to squeeze but it's getting pretty big so might be time to cleanup and merge what can be. My problem is that this is a lot of small optimizations that are within the error  margin when benchmarked individually, so hard to justify them... Hence not to sure how to ship that.

cmaion:
I guess that adding a `ActiveRecord::Base.lease_connection` before running the test still reduces the gap a bit?

byroot:
Yes, if I keep a permanent lease:

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
     ar-fast-checkin     8.581k i/100ms
     ar-fast-checkin     86.840k (± 1.2%) i/s   (11.52 μs/i) -    437.631k in   5.040239s

Comparison:
          7-1-stable:    87392.0 i/s
     ar-fast-checkin:    86839.6 i/s - same-ish: difference falls within error
                main:    67712.3 i/s - 1.29x  slower
```

cmaion:
Ran the bench on my side too:

```
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [x86_64-linux]
         7-2-stable*     3.263k i/100ms
         7-2-stable*     37.049k (± 2.4%) i/s   (26.99 μs/i) -    185.991k in   5.022907s

Comparison:
          7-1-stable:    42964.7 i/s
    ar-fast-checkin*:    38859.4 i/s - 1.11x  slower
         7-2-stable*:    37049.1 i/s - 1.16x  slower
     ar-fast-checkin:    36060.7 i/s - 1.19x  slower
                main:    31789.4 i/s - 1.35x  slower
          7-2-stable:    31203.9 i/s - 1.38x  slower
```

- `7-2-stable*` is 7.2 + permanent lease
- `ar-fast-checkin*` is your branch + permanent lease

Much closer indeed, and the micro optimizations does indeed add up, even if by isolation they are certainly hard to quantify.


EDIT: Ruby 3.4.6:
```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [x86_64-linux]
    ar-fast-checkin*     3.529k i/100ms
    ar-fast-checkin*     37.999k (± 4.0%) i/s   (26.32 μs/i) -    190.566k in   5.022988s

Comparison:
          7-1-stable:    44028.9 i/s
    ar-fast-checkin*:    37999.3 i/s - 1.16x  slower
               main*:    36109.0 i/s - 1.22x  slower
         7-2-stable*:    35829.3 i/s - 1.23x  slower
     ar-fast-checkin:    35292.5 i/s - 1.25x  slower
                main:    32538.4 i/s - 1.35x  slower
          7-2-stable:    30355.4 i/s - 1.45x  slower
```

cmaion:
Benchmark result is not super stable on my laptop...
Re-ran the tests without YJIT to eliminate one potential source of noise.

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +PRISM [x86_64-linux]
    ar-fast-checkin*     1.563k i/100ms
    ar-fast-checkin*     15.658k (± 2.6%) i/s   (63.86 μs/i) -     79.713k in   5.094245s

Comparison:
          7-1-stable:    15920.1 i/s
    ar-fast-checkin*:    15658.2 i/s - same-ish: difference falls within error
         7-2-stable*:    14475.3 i/s - 1.10x  slower
               main*:    14319.4 i/s - 1.11x  slower
     ar-fast-checkin:    13249.6 i/s - 1.20x  slower
                main:    12245.5 i/s - 1.30x  slower
          7-2-stable:    11532.5 i/s - 1.38x  slower
```

Looks like the conclusion still holds.
I also note that holding a permanent connection is still worthwhile when it's reasonable to do so (and obviously when the latency with the database is minimal too).

rafaelfranca:
> My problem is that this is a lot of small optimizations that are within the error margin when benchmarked individually, so hard to justify them... Hence not to sure how to ship that.

I'd not worry with that. If the aggregation of patches make the code faster, I'd just apply them all together (in individual commits)

byroot:
Alright, I've cleaned up the git history, the benchmark is now:

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
          7-1-stable     3.219k i/100ms
          7-1-stable     88.512k (± 4.1%) i/s   (11.30 μs/i) -    444.222k in   5.027300s

Comparison:
     ar-fast-checkin:    79704.1 i/s
          7-1-stable:    88512.0 i/s - 1.11x  faster
                main:    70279.4 i/s - 1.13x  slower
```

I think there is a bit more perf to squeeze in AS:N, but I'd rather not sleep on these gains and merge soon.

I also removed the interlock patch given Matthew merged https://github.com/rails/rails/pull/55753

activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
```diff
@@ -179,21 +179,30 @@ def clear(connection)
...
 
-      class LeaseRegistry # :nodoc:
-        def initialize
-          @mutex = Mutex.new
-          @map = WeakThreadKeyMap.new
+      if RUBY_ENGINE == "ruby"
+        # Thanks to the GVL, the LeaseRegistry doesn't need to be synchronized on MRI
+        class LeaseRegistry < WeakThreadKeyMap # :nodoc:
+          def [](context)
+            super || (self[context] = Lease.new)
```
matthewd, line 186: Is this safe just because we know that we _are_ `context`, and therefore even if we stall during `Lease.new`, we can't possibly be racing someone else?

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -592,8 +592,10 @@ def affected_rows(raw_result)
         end
 
         def preprocess_query(sql)
-          check_if_write_query(sql)
-          mark_transaction_written_if_write(sql)
+          if write_query?(sql)
+            ensure_writes_are_allowed(sql)
+            mark_transaction_written
```
matthewd, line 597: Should we just inline both of these? They're both basically `x if y`, which was probably more method-worthy when they were called separately from each adapter etc? 🤷🏻‍♂️

activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
```diff
@@ -330,8 +328,12 @@ def expire(update_idle = true) # :nodoc:
               "Current thread: #{ActiveSupport::IsolatedExecutionState.context}."
           end
 
-          @idle_since = Process.clock_gettime(Process::CLOCK_MONOTONIC) if update_idle
-          @owner = nil
+          _run_checkin_callbacks do
+            @idle_since = Process.clock_gettime(Process::CLOCK_MONOTONIC) if update_idle
+            @owner = nil
+            enable_lazy_transactions!
```
matthewd, line 334: Not something to change here, but consolidating like this draws my eye to the fairly arbitrary distribution of which "return to neutral state" things happen during checkin, and which during checkout.

byroot:
I feel like I'm taking crazy pills here. 

I added a monkey patch in the benchmark to entirely bypass the database, because 7.1 and this branch use very different version of SQLite 3, and I wanted to reduce variance.

There's still a 9 to 15% difference (varies quite a bit)

```
Comparison:
         7-1-stable*:   114190.0 i/s
    ar-fast-checkin*:   101621.7 i/s - 1.12x  slower
```

But I'm struggling to spot a significant difference on the flame graphs:

7.1: https://share.firefox.dev/3W92sGQ
this branch:  https://share.firefox.dev/4pGMEZo

Benchmark with monkey patches:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", path: "."

  gem "debug"
  gem "sqlite3"
  gem "benchmark-ips"
  gem "vernier"
end

require "benchmark/ips"

  config.eager_load = false
  config.logger = Logger.new(nil)
  config.log_level = :info
  config.cache_classes = true
  config.eager_load = true

end
Rails.application.initialize!


  end

    t.integer :post_id
  end
end

ActiveRecord::Base.release_connection unless Rails::VERSION::STRING < "7.2"

end

if ENV["SKIP_DB"]
  if Rails::VERSION::STRING < "7.2"
    module Sqlite3SkipDb
      def internal_exec_query(sql, name = nil, binds = [], prepare: false, async: false) # :nodoc:
        if sql == "SELECT COUNT(*) FROM \"posts\""
          sql = transform_query(sql)
          check_if_write_query(sql)

          mark_transaction_written_if_write(sql)

          type_casted_binds = type_casted_binds(binds)

          log(sql, name, binds, type_casted_binds, async: async) do
            with_raw_connection do |conn|
              # Don't cache statements if they are not prepared
              verified!

              build_result(columns: ["count(*)"], rows: [[1]])
            end
          end
        else
          super
        end
      end
    end
    ActiveRecord::ConnectionAdapters::SQLite3Adapter.prepend(Sqlite3SkipDb)
  else
    module Sqlite3SkipDb
      def perform_query(raw_connection, sql, binds, type_casted_binds, prepare:, notification_payload:, batch: false)
        if sql == "SELECT COUNT(*) FROM \"posts\""
          verified!

          notification_payload[:affected_rows] = 0
          notification_payload[:row_count] = 1
          ActiveRecord::Result.new(["count(*)"], [[1]], affected_rows: 0)
        else
          super
        end
      end
    end
    ActiveRecord::ConnectionAdapters::SQLite3Adapter.prepend(Sqlite3SkipDb)
  end
end


BRANCH = `git rev-parse --abbrev-ref HEAD`.strip + (ENV["SKIP_DB"] ? "*" : "")

Vernier.profile(out: "time_profile-#{BRANCH}.json") do
  (55060 * 10).times do
    Post.count
  end
end

Benchmark.ips do |x|
  x.report(BRANCH) { Post.count }
  x.save!("/tmp/bench-ar-data")
  x.compare!(order: :baseline)
end
```

activerecord/lib/active_record/explain_registry.rb
```diff
@@ -24,6 +69,11 @@ def initialize
       reset
     end
 
+    def start
+      Subscriber.ensure_subscribed
```
matthewd, line 73: Nice!

activesupport/lib/active_support/notifications/fanout.rb
```diff
@@ -184,25 +187,25 @@ def build_event(name, id, payload)
           end
       end
 
-      def groups_for(name) # :nodoc:
-        groups = @groups_for.compute_if_absent(name) do
-          all_listeners_for(name).reject(&:silenceable).group_by(&:group_class).transform_values do |s|
-            s.map(&:delegate)
-          end
-        end
+      def group_listeners(listeners)
```
matthewd, line 190:
```suggestion
      def group_listeners(listeners) # :nodoc:
```

activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
byroot: Indeed. If we were to check the lease of another thread, we could race on the `self[context] = Lease.new`, but even that wouldn't necessarily be a problem.

Perhaps this may bite us in the future if we're not careful, but I like having minimal locking.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
byroot: I suppose we could yes. More importantly we should pass down whether we expect a read or write query, e.g. if we're called from `select_all`, we should assume it's a read query.

That regexp is about 1.5% on the profile AFAICT.

activesupport/lib/active_support/notifications/fanout.rb
byroot: Thanks, should have waited for your review. I'll fix that on `main`.

PR #40768, opened 2020-12-09
Add `ActiveRecord::FinderMethods#sole` and `#find_sole_by`

kivikakk:
Per a short discussion at https://discuss.rubyonrails.org/t/proposal-patch-findermethods-only-for-asserting-theres-only-one-result-row/76539, I thought I'd submit this as a PR proper.

### Summary

Django ORM’s `QuerySet` has a [`get()`](https://docs.djangoproject.com/en/3.1/topics/db/queries/#retrieving-a-single-object-with-get) method, which asserts that the query only matches a single row. For some reason I thought `find_by` did that in Rails, too, but recent poking around led me to realise that’s not the case and I might’ve relied upon this somewhere :sweat_smile:

The basic usage looks like this (adapted from an actual use case):

```ruby
Product.where(price: price, annual: true).only!
```

It fills the niche where, for whatever reason, adding a validation or constraint that covers the exact circumstances the query is expressing is clunky or untenable, and helps express intention at the point of use that there should be only one matching row.

PR #55741, opened 2025-09-23
feat: Add `:allow_nil` keyword to `#sole`

ixti:
### Motivation / Background

Currently, the `#sole` method enforces uniqueness strictly by raising an error if no or multiple items are found. However, there are scenarios where it is desirable for `#sole` to return `nil` instead of raising an error when there are no matching records, for example:

* When application-level uniqueness is guaranteed, and zero results should be a silent, non-exceptional case.
* When validations or database uniqueness constraints are either insufficient or too costly (for performance reasons).
* To simplify flows where distinguishing between "not found" and "multiple found" is important, without always needing custom error handling.

To address these cases, this change introduces an optional `:allow_nil` keyword argument to #sole, defaulting to `false`. When set to `true`, the method returns `nil` instead of raising an error if no items are found.

### Detail

* Default behavior of `#sole` is unchanged -- it raises an error if zero or more than one item is found.
* With `allow_nil: true`, the method returns `nil` if no records/items exist. If more than one is found, it still raises the relevant error.

#### Example: Enumerable

```ruby
[].sole                       # => raise SoleItemExpectedError
[].sole(allow_nil: false)     # => raise SoleItemExpectedError
[].sole(allow_nil: true)      # => nil

[1].sole                      # => 1
[1].sole(allow_nil: false)    # => 1
[1].sole(allow_nil: true)     # => 1

[1, 2].sole                   # => raise SoleItemExpectedError
[1, 2].sole(allow_nil: false) # => raise SoleItemExpectedError
[1, 2].sole(allow_nil: true)  # => raise SoleItemExpectedError
```

#### Example: ActiveRecord

```ruby
topics = Topic.where(title: "Title Does Not Exist")

topics.count                    # => 0
topics.sole                     # => raise RecordNotFound
topics.sole(allow_nil: false)   # => raise RecordNotFound
topics.sole(allow_nil: true)    # => nil

topics = Topic.where(title: "The First Topic")

topics.count                    # => 1
topics.sole                     # => #<Topic ...>
topics.sole(allow_nil: false)   # => #<Topic ...>
topics.sole(allow_nil: true)    # => #<Topic ...>

topics = Topic.all

topics.count                    # => 2+
topics.sole                     # => raise SoleRecordExceeded
topics.sole(allow_nil: false)   # => raise SoleRecordExceeded
topics.sole(allow_nil: true)    # => raise SoleRecordExceeded
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,30 @@
+*   Add optional `:allow_nil` flag to `#sole` method with defaults to `false`.
+    When `allow_nil: true` no exception will raise if collection has no records.
+    ```ruby
+    topics = Topic.where(title: "Title Does Not Exist")
+
+    topics.count                  # => 0
+    topics.sole                   # => raise RecordNotFound
+    topics.sole(allow_nil: false) # => raise RecordNotFound
+    topics.sole(allow_nil: true)  # => nil
+
+    topics = Topic.where(title: "The First Topic")
+
+    topics.count                  # => 1
+    topics.sole                   # => #<Topic ...>
+    topics.sole(allow_nil: false) # => #<Topic ...>
+    topics.sole(allow_nil: true)  # => #<Topic ...>
+
+    topics = Topic.all
+
+    topics.count                  # => 2+
+    topics.sole                   # => raise SoleRecordExceeded
+    topics.sole(allow_nil: false) # => raise SoleRecordExceeded
+    topics.sole(allow_nil: true)  # => raise SoleRecordExceeded
+    ```
+
+    *Alexey Zapparov*
+
 *   Add support for integer shard keys.
     ```ruby
     # Now accepts symbols as shard keys.
--- a/activerecord/lib/active_record/relation/finder_methods.rb
+++ b/activerecord/lib/active_record/relation/finder_methods.rb
@@ -139,12 +139,28 @@ module ActiveRecord
     # record is found. Raises ActiveRecord::SoleRecordExceeded if more than one
     # record is found.
     #
-    #   Product.where(["price = %?", price]).sole
-    def sole
+    #   Product.where(["price = %?", price]).count                  # => 0
+    #   Product.where(["price = %?", price]).sole                   # => raise RecordNotFound
+    #   Product.where(["price = %?", price]).sole(allow_nil: true)  # => nil
+    #
+    #   Product.where(["price = %?", price]).count                  # => 1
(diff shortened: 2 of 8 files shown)


ixti:
Ideally, it would be nice to introduce `#sole!` and `#find_sole_by!` that would have current behaviour, and change bangless methods to what this PR achieves with `allow_nil: true`. But that will require breaking changes and deprecation period.

If that's something core team might consider - I can prepare a next PR(s). Something like:

``` ruby
def sole!
  sole(allow_nil: false)
end

def sole(allow_nil: LEGACY_SOLE)
  if allow_nil == LEGACY_SOLE
    ActiveSupport::Deprecation.warn("Default behavior of sole is about to change... Use sole!")
    allow_nil = false
  end

  ...
end
```

UPDATE: Realized that having `sole` and `sole!` will make API more confusing.

rafaelfranca:
I don't like adding keyword arguments for this method to change behavior. I'd be ok with a new method for the new behavior, but it will be hard to find a good name for it. See #40768 for context.

cc @matthewd

ixti:
@rafaelfranca oh, I was trying to search if there was a discussion and failed. Frankly in my application I have a small monkey-patch:

``` ruby
# frozen_string_literal: true

module Extensions
  module ActiveRecord
    module FinderMethods
      module SoleOrNothing
        def sole_or_nothing
          suppress(::ActiveRecord::RecordNotFound) { sole }
        end
      end
    end
  end
end

ActiveRecord::FinderMethods.include Extensions::ActiveRecord::FinderMethods::SoleOrNothing
```

Another name I had in mind was `sole_or_none`.

ixti:
Also, I came to realize that probably where I need this functionality it will be more obvious to have:

``` ruby
suppress(ActiveRecord::RecordNotFound) { collection.sole }
```

Or even be more verbose (as it's isolated in a finder method anyway):

``` ruby
def find_active(at: Date.current)
  at = at.to_date
  found, unexpected = model.where(since: at.., till: ...at).first(2)
  return found if unexpected.nil?

  raise UnexpectedRecordsCount, model
end
```

Rather than `sole(allow_nil: true)` or `sole_or_none` (or any other name). Thus, if you don't feel like this efforts are worth any pursuit - feel free to close it.

ixti:
After poking a bit more around, and reading the discussion @rafaelfranca you've linked, I do believe introducing change in this PR, or as a separate method is a bad idea. I mean, sometimes explicit is better than implicit, when implicit adds too much cognitive overhead. Thus, I'm going to close this PR. Sorry for the noise.

PR #55742, opened 2025-09-23
[Getting Started Tutorial] Add note on default string type for `email` column for Subscriber and explain the resulting migration file [ci skip]

harsh183:
### Motivation / Background

When I was doing the tutorial, I noticed that the `email` column was not specified with a type yet it defaulted to a string. This confused me, so I wanted to add more detail about it. 

### Detail

I added a note for the default with showing the generated migration file.

### Additional information

I tried looking for documentation on this, but didn't find anything. When I tried it in a few different projects, this did work. I did find the relevant code in `railties/lib/rails/generators/generated_attribute.rb` where
```ruby
def initialize(name, type = nil, index_type = false, attr_options = {})
  @name           = name
  @type           = type || :string
```

In the tutorial I don't want to link to the actual codebase, but I think there's a few more pages that I can add a note too.
* https://guides.rubyonrails.org/active_record_migrations.html
* https://guides.rubyonrails.org/command_line.html
* https://guides.rubyonrails.org/active_record_basics.html

Not sure if there's any other API specifc pages I can add a note to. If I do add these, should this be in this PR or a separate one?

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2031,19 +2031,36 @@ of these subscribers.
 Let's generate a model called Subscriber to store these email addresses and
 associate them with the respective product.
 
+NOTE: Here we are not specifying a type for `email` as rails automatically defaults to a `string` when a type is not given for migrations.
+
 ```bash
 $ bin/rails generate model Subscriber product:belongs_to email
 ```
+
+By including `product:belongs_to` above, we told Rails that subscribers and products have a one-to-many relationship, meaning a Subscriber "belongs to" a single Product instance.
+
+Next, open the generated migration (`db/migrate/<timestamp>_create_subscribers.rb`) like we did for Product.
+
+```ruby#4-5
+class CreateSubscribers < ActiveRecord::Migration[8.1]
+  def change
+    create_table :subscribers do |t|
+      t.belongs_to :product, null: false, foreign_key: true
+      t.string :email
+
+      t.timestamps
+    end
+  end
+end
+```
+
+This looks quite similar to the migration for `Product`, the main new thing is `belongs_to` which adds a `product_id` foreign key column.
 
 Then run the new migration:
 
 ```bash
 $ bin/rails db:migrate
 ```
-
-By including `product:belongs_to` above, we told Rails that subscribers and
-products have a one-to-many relationship, meaning a Subscriber "belongs to" a
-single Product instance.
 
 A Product, however, can have many subscribers, so we then add
 `has_many :subscribers, dependent: :destroy` to our Product model to add the


MatheusRich (approved):

guides/source/getting_started.md
```diff
@@ -2031,20 +2031,35 @@ of these subscribers.
...
+      t.belongs_to :product, null: false, foreign_key: true
+      t.string :email
+
+      t.timestamps
+    end
+  end
+end
+``` 
+
+This looks quite similar to the migration for `Product`, the main new thing is `belongs_to` which tells Rails that subscribers and products have a one-to-many relationship, meaning a Subscriber "belongs to" a single Product instance.
```
Edouard-chin: > which tells Rails that subscribers and products have a one-to-many relationship

I think this part is a bit misleading. What the belongs_to in the migration does is just adds a reference column. The "one-to-X" is defined on the model (could even be a one-to-one).

The next paragraph [explains it well](https://github.com/rails/rails/pull/55742/files#diff-b0908447c00d5a7a884e8b4347c1bc7fff22e8987d64891a428c55a12e1d9234R2063-R2065).

harsh183: I'll update this. I just initially moved up the next paragraph since I was explaining the string bit. I'll explain that this belongs to adds a reference column in the table and rails knows how to join accordingly.

harsh183: I added a note with more explicit detail about how the linking works and tweaked some other wording. Let me know if this looks good.

harsh183:
I made all updates, let me know if I need to do something else. I'll try to make doc changes for the other places now

guides/source/getting_started.md
```diff
@@ -2031,20 +2031,37 @@ of these subscribers.
...
+      t.belongs_to :product, null: false, foreign_key: true
+      t.string :email
+
+      t.timestamps
+    end
+  end
+end
+``` 
+
+This looks quite similar to the migration for `Product`, the main new thing is `belongs_to` which adds a `product_id` foreign key column. This sets up the database side of a one-to-many relationship, where each Subscriber "belongs to" a single Product instance.
```
rafaelfranca: This paragraph is repeating a lot of what was said above. Let's cut it.
```suggestion
This looks quite similar to the migration for `Product`, the main new thing is `belongs_to` which adds a `product_id` foreign key column.
```

rafaelfranca (changes requested):

guides/source/getting_started.md
harsh183: Done!

rafaelfranca (approved):

PR #55743, opened 2025-09-23
Don't generate system tests by default

eileencodes:
System tests are a nice to have but don't make sense to be generated by default since you don't want to test every endpoint this way. They should be used for testing critical paths, especially where JS is involved.


ghiculescu:
@eileencodes should we remove the system test stuff from https://github.com/rails/rails/blob/cf6b3107753a762b27bfcf9eefcfd9feb43f0242/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt#L222 ? Happy to make a PR if you're 👍  on it.

Arawn-Davies:
Hello :) I think somewhere in these commits a change has broken the GitHub workflow system tests. I think a check needs to see if the system tests are not generated by default, then the system tests GitHub Actions thingmabob shouldn't be there either.

If possible, there needs to be an inverse option where if system tests are generated, then they can be created as a GitHub whatchamacallit too - I still want to generate the system tests personally 

I'm very new to Ruby & Rails and it's super late (4AM currently) so apologies as I've forgotten the CI stuffs name 😆

PR #55749, opened 2025-09-23
ActiveRecord Guide: Fix 404 errors in links to api.rubyonrails.org [ci skip]

Spone:
### Motivation / Background

This Pull Request has been created because I noticed broken external links on this page https://guides.rubyonrails.org/active_record_callbacks.html

### Detail

Removing the `/classes` at the beginning of URL paths fixes the links.

This Pull Request changes all links from:

```diff
- https://api.rubyonrails.org/classes/ActiveRecord/...
+ https://api.rubyonrails.org/ActiveRecord/...
```

### Checklist

* [ ] ~~Tests are added or updated if you fix a bug or add a feature.~~

### DIFF
--- a/guides/source/active_record_callbacks.md
+++ b/guides/source/active_record_callbacks.md
@@ -183,25 +183,25 @@ section](active_record_callbacks.html#after-commit-and-after-rollback) for
 examples using these two callbacks.
 
 [`after_create`]:
-    https://api.rubyonrails.org/classes/ActiveRecord/Callbacks/ClassMethods.html#method-i-after_create
+    https://api.rubyonrails.org/ActiveRecord/Callbacks/ClassMethods.html#method-i-after_create
 [`after_commit`]:
-    https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit
+    https://api.rubyonrails.org/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit
 [`after_rollback`]:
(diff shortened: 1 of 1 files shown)


Edouard-chin:
The links are indeed broken but I don't think it's because the links are wrong. It seems to be related to this change in the docs-server https://github.com/rails/rails-docs-server/commit/8bcc36e9d40559d07c9f3a6a8540acc76f766c6f

https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/Callbacks/ClassMethods.html#method-i-after_create works but https://api.rubyonrails.org/v8.0.3/classes/ActiveRecord/Callbacks/ClassMethods.html#method-i-after_create doesn't

FYI @rafaelfranca

skipkayhil:
https://github.com/ruby/rdoc/pull/1304

Yeah it appears this change in `rdoc` broke our paths, and the correct fix will be to update `sdoc`

skipkayhil (changes requested):
see my comment above

rafaelfranca:
I need to apply https://github.com/rails/sdoc/commit/141f755e81b75a7ce88c2f8408d2d7406e71765c. Doing it now

Edouard-chin:
Ah yeah that's what I was looking at because rdoc reintroduced an API for this https://github.com/ruby/rdoc/pull/1330

rafaelfranca:
Fixed in https://github.com/rails/rails-docs-server/commit/ceb79fc7f1152d298395ea5a5497136409abdb91

PR #55753, opened 2025-09-24
Remove loading behaviour from interlock

matthewd:
The running/unloading interaction is still needed, but "loading" protection was only required for classic autoloading.

This eliminates the need for `permit_concurrent_loads` -- which is the main reason I didn't want to rush into it: once we've stopped recommending that wrapping, it would be very hard to put back again. But at this point I think it's pretty clear it can safely go away.

I've removed all references to `permit_concurrent_loads` from the guide, but kept the method as a silent no-op (no deprecation). We can eventually age it out via a future deprecation, but for now there's no need to bother every existing caller.

In contrast, there should be approximately zero external callers directly invoking "loading" mode on the interlock; those methods are deprecated here just in case, but _are immediately no-op with no locking applied_. Since the switch to Zeitwerk, they've already not been interacting with framework-level loading, so this should only create an observable change if someone was using our lock as if it was their own mutex.

I've also changed references to the "Load Interlock" in the guide to "Reloading Interlock"; it's a longer name, a different name, and breaks the in-page anchor... but it just seems confusing to keep the old name when it's (already!) not involved in Load, only Reload.

### DIFF
--- a/actionpack/lib/action_controller/metal/live.rb
+++ b/actionpack/lib/action_controller/metal/live.rb
@@ -236,12 +236,7 @@ module ActionController
 
       private
         def each_chunk(&block)
-          loop do
-            str = nil
-            ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
-              str = @buf.pop
-            end
-            break unless str
+          while str = @buf.pop
             yield str
           end
         end
@@ -306,9 +301,7 @@ module ActionController
         end
       end
 
-      ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
-        @_response.await_commit
-      end
+      @_response.await_commit
 
       raise error if error
     end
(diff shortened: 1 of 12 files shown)


byroot:
You need to rebase 😄

byroot (approved):
❤️

matthewd:
Open questions / points of hesitation:

* Is it reasonable to rush the deprecation like this, rather than going through a release that warns while maintaining identical behaviour? (Per above I'm only really considering this because it turns out the "loading" state is _already_ not used by any built-in behaviour.)
* Especially given that turbo deprecation, should this wait, since we've already cut beta 1, for the next release?

byroot:
> * Is it reasonable to rush the deprecation like this

I honestly don't care so much whether it's deprecated or nor as long as it's out of the hot paths.

That being said do we really expect much people calling this? I've never seen it in user code, nor gems.

matthewd:
Yeah the only mention I can see in GH code search is old forks that predate https://github.com/ManageIQ/manageiq-api/commit/8885e1ea6d1ba5656ec7a29d9d7af8aec5b5fd17 -- and that's one more [past] caller than I'd have expected.

But that's the question: it's only safe to empty out `permit_concurrent_loads` as I'm doing here if we're confident no-one is relying on being able to actually acquire a load lock [via this deprecated API]

I think it's probably fine, though 🤷🏻‍♂️

byroot:
Yes, shipit

rafaelfranca (approved):

PR #55757, opened 2025-09-24
Special case root SCRIPT_NAME [Fix #55756]

ianterrell:
### Motivation / Background

This Pull Request has been created because route generation fails unexpectedly in an app with several interacting engines. I have detailed the bug here:

- https://github.com/rails/rails/issues/55756

In brief, this fails:

```ruby
assert_equal "/dashboard", user.dashboard_path
get admin.dashboard_path
assert_equal "/dashboard", user.dashboard_path
```

### Detail

This Pull Request changes recent changes in 8.0.3 from this PR:

- https://github.com/rails/rails/pull/55668

### Checklist

I have not yet been able to get a good test environment set up for this, and haven't been able to dive into how to test this within the Rails test suite. I do see that this change breaks the test introduced in the prior PR. I am hoping someone with greater familiarity can help reconcile what needs to be done here, as I think the test case outlined above should trivially pass.

### DIFF
--- a/actionpack/lib/action_dispatch/routing/routes_proxy.rb
+++ b/actionpack/lib/action_dispatch/routing/routes_proxy.rb
@@ -54,7 +54,7 @@ module ActionDispatch
       # dependent part.
       def merge_script_names(previous_script_name, new_script_name)
         return new_script_name unless previous_script_name
-        new_script_name = new_script_name.chomp("/")
+        new_script_name = new_script_name.chomp("/") unless new_script_name == "/"
 
         resolved_parts = new_script_name.count("/")
         previous_parts = previous_script_name.count("/")


rafaelfranca:
From the original PR:

> The issue is in the private method ActionDispatch::Routing::RoutesProxy#merge_script_names which doesn't handle the edge case of new_script_name being a sole slash "/" with no segment following it.

You are putting back the code that was broken before, that is why the tests added in that PR are broken.

I don't think this is the right fix. If the behaviors are exact opposite, we will probably have to decide which one we will not support.

flavorjones:
👋 Hi, I'm the author of the mentioned PR that changed this behavior. Let me dig in to what's happening.

flavorjones:
I appreciate opening a pull request to try to fix this; but this essentially undoes the PR, and if that's the solution we should instead do a clean revert.

That said, I don't think this is the right fix, and I'd like to take the conversation back to the bug report https://github.com/rails/rails/issues/55756 to dig into the root causes. Thank you!

PR #55760, opened 2025-09-24
Deprecate usage of custom ActiveJob serializers without public #klass methods

that-jill:
### Motivation / Background

The performance gains from https://github.com/rails/rails/pull/55583 meant that, when testing upgrades from 8.0 to 8.1, suddenly the app crashed and stopped evaluating due to the raised `NotImplementedError` in the parent class:

Screenshot 2025-09-24 at 5 04 16 PM

[`index_serializers` checks if the class responds to `:klass`](https://github.com/rails/rails/blob/main/activejob/lib/active_job/serializers.rb#L78)

but in the case of custom serializers, it will also do, even if it doesn't have an implemented method because [it inherits from the ObjectSerializer, which raises the `NotImplementedError`.](https://github.com/rails/rails/blob/main/activejob/lib/active_job/serializers/object_serializer.rb#L52)

Which is what was causing things to raise, tests to fail, etc.

 This indexer isn't strictly required (there's a fallback), so it feels appropriate to do a proper deprecation cycle and then remove this behavior in a later version.

### Detail

This Pull Request catches the NotImplementedError in this method and instead reports the deprecation. When that happens, the indexer will fall back to the lookup, so they don't get the performance gains, but their app will continue to run until the deprecation is addressed.

I also added some clarity to the deprecations so people will know what to expect.

### Checklist

* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. _Note: no changelog is included because the notes from https://github.com/rails/rails/pull/55583 feel sufficient to me_

### DIFF
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -76,11 +76,20 @@ module ActiveJob
           @serializers_index.clear
           serializers.each do |s|
             if s.respond_to?(:klass)
-              @serializers_index[s.klass] = s
+              begin
+                @serializers_index[s.klass] = s
+              rescue NotImplementedError => e
+                ActiveJob.deprecator.warn(
+                  <<~MSG.squish
+                    #{e.message}. This will raise an error in Rails 8.2.
+                  MSG
+                )
+              end
             elsif s.respond_to?(:klass, true)
               klass = s.send(:klass)
               ActiveJob.deprecator.warn(<<~MSG.squish)
                 #{s.class.name}#klass method should be public.
+                This will raise an error in Rails 8.2.
               MSG
               @serializers_index[klass] = s
             end
--- a/activejob/test/cases/serializers_test.rb
+++ b/activejob/test/cases/serializers_test.rb
@@ -30,6 +30,8 @@ class SerializersTest < ActiveSupport::TestCase
     end
   end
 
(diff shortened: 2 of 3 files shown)


Edouard-chin:
Do you know how the custom serializers that have issues currently work if they didn't implement the `klass` method ? Active Job expects this method to be implemented per the interface of `ObjectSerializer`.

Edouard-chin:
Ah ok I see, they override the `serialize?` method which is the one that make uses of `#klass`. That seems like a breach of contract :( .

```ruby
class MySerializer < ObjectSerializer
  def serialize?(arg)
     arg.is_a?(Something)
  end
end
```

gmcgibbon (approved):
I think this makes sense. We don't want apps to crash over an opt-in performance optimization. Warning first, and then raising later when/if this feature is mandatory seems best, though I'm not sure we can avoid the array search for subclasses.

matthewd:
If we want a deprecation here, can we put it inside the method, rather than doing the raise -> rescue -> warn?

that-jill:
> If we want a deprecation here, can we put it inside the method, rather than doing the raise -> rescue -> warn?

happy to do so -- updated in https://github.com/rails/rails/pull/55768

Issue #55762, opened 2025-09-25
`ActionController::Live` causes thread pool exhaustion for non-streaming actions


moznion:
### Steps to reproduce

1. Create a controller with ActionController::Live included
2. Add both SSE endpoints and regular JSON API endpoints
3. Send high volume of requests to the non-streaming endpoints
4. Monitor thread pool usage - threads will be consumed and not released properly
5. Eventually, the application will stop responding to new requests

### Expected behavior

It would be preferable if a controller including `ActionController::Live` could handle both streaming and non-streaming actions correctly without leading to thread exhaustion.

### Actual behavior

```ruby
class MixedController < ApplicationController
  include ActionController::Live

  # This action needs Live for SSE
  def events
    response.headers['Content-Type'] = 'text/event-stream'
    sse = SSE.new(response.stream, retry: 300, event: "event-name")
    # ... SSE implementation
  ensure
    sse.close
  end

  # These actions DON'T need Live but still run in separate threads!
  def index
    render json: { data: "..." }  # Runs in a thread unnecessarily
  end

  def show
    @user = User.find(params[:id])
    render json: @user  # Also runs in a thread unnecessarily
  end
end
```

In the above example, even index and show actions, which return regular JSON responses, are executed in separate threads from the thread pool. Under high traffic,
these non-streaming actions can exhaust the thread pool, causing the application to become unresponsive.

### System configuration

**Rails version**: Rails 8.0.2.1

**Ruby version**: 3.4.4

PR #55763, opened 2025-09-25
Add documentation on how to prevent thread exhaustion in `ActionController::Live`

moznion:
### Motivation / Background

Ref: https://github.com/rails/rails/issues/55762

While it would be preferable to resolve the above problem fundamentally, this pull request provides documentation to alert users on how to prevent thread exhaustion in `ActionController::Live` as first-aid workaround.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/live.rb
+++ b/actionpack/lib/action_controller/metal/live.rb
@@ -26,18 +26,42 @@ module ActionController
   #       end
   #     end
   #
-  # There are a few caveats with this module. You **cannot** write headers after
-  # the response has been committed (Response#committed? will return truthy).
-  # Calling `write` or `close` on the response stream will cause the response
-  # object to be committed. Make sure all headers are set before calling write or
-  # close on your stream.
+  # ## Considerations
+  #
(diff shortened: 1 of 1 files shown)


bensheldon:
I think this could be a written a little more precisely and tightly. I don't think it's Thread Exhaustion (CachedThreadPool can create up to [2 billion](https://github.com/ruby-concurrency/concurrent-ruby/blob/05523c01f6f2eebcb7579c2b8efc29e595010bb0/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb#L15) [threads](https://github.com/ruby-concurrency/concurrent-ruby/blob/05523c01f6f2eebcb7579c2b8efc29e595010bb0/lib/concurrent-ruby/concurrent/executor/cached_thread_pool.rb#L42)) but just "resource exhaustion".

The main takeaway here could be a bit sharper of like: 

> When you include `ActionController::Live`, all actions in that controller will be execute in their own thread, which can lead to resource exhaustion. To avoid this, limit the number of actions in the controller to only those requiring streaming.

Generally I think this is good to document 👍🏻

PR #55765, opened 2025-09-25
[ci skip][docs] Adds guide for Active Record Aggregations With Examples And References

shivabhusal:
### Motivation / Background
While browsing the Active Record codebase, I came across the `Aggregations` module. I realized that, despite its usefulness, it isn’t currently documented in the Rails Guides. After confirming that no guide exists for this feature, I decided to add one so that developers can better understand and utilize Active Record Aggregations.

### Details
This pull request introduces a new guide for **Active Record Aggregations**, explaining how to use `composed_of` with value objects through examples and caveats.  
Additionally, it updates the Rails Guides index and navigation menu to include this new document for easier discovery.

Suggestions, improvements, and feedback are very welcome.

See [**Added Doc Latest Preview**:](https://shivabhusal-active-record-ag.rails-docs-preview.pages.dev/guides/active_record_aggregations)

### Checklist

### DIFF
--- a/guides/source/active_record_aggregations.md
+++ b/guides/source/active_record_aggregations.md
@@ -1,0 +1,417 @@
+**DO NOT READ THIS FILE ON GITHUB, GUIDES ARE PUBLISHED ON <https://guides.rubyonrails.org>.**
+
+Active Record Aggregations
+==========================
+
+This guide covers how Rails supports aggregations in Active Record models.
+
+After reading this guide, you will know:
+
+* What are value objects and how to define them.
+* How to use Value Objects with Active Record aggregations.
+* How to query and manipulate aggregated attributes.
+
+--------------------------------------------------------------------------------
+
+What is Active Record Aggregations?
+-----------------------------------
+
+In many domains, simple columns like strings and integers are not expressive enough on their own.
+You may want to introduce a [*value object*](#what-are-value-objects-questionmark) to represent a richer concept in your domain.
+
+For example:
+
+- a `Money` class to represent monetary values,
+- an `Address` class for postal addresses,
+- a `Dimension` class for measurements,
+- or even complex numbers.
+
+**Active Record Aggregations** provides a way to map these `value objects` attributes to database columns using
+the [`composed_of`](https://api.rubyonrails.org/classes/ActiveRecord/Aggregations/ClassMethods.html) method.
+It is used to express `is-composed-of` relationships.
+
+For example:
+
+- An `Account` *is composed of* `Money`.
+- A `User` *is composed of* `Address`.
+- A `Product` *is composed of* `Dimension`.
+
+What are Value Objects?
+-----------------------
+
+Value objects are immutable and interchangeable objects that represent a single value or concept, so no setter methods are allowed.
+This improves clarity, keeps business rules close to the data they describe, and avoids the pitfalls of using
+raw primitive data-types throughout your models and business logic.
+They are often used to encapsulate related attributes and behavior, such as Money object can represent `$100`, and
+Address object can represent `123 Main St, Springfield, IL`.
+
+### Value Object Comparison
+
+Any two value objects with the same attributes or calculative equal values are considered equal.
+
+For example:
+
+* two `Money` objects with the same amount and currency are considered equal, even if they are different instances.
+* Also, two money objects with different amounts but equal value when exchanged to the same currency are considered equal.
+
+While defining a value object, you typically override the `==` and `<=>` methods to compare the attributes of the objects.
+Also, you define a way to convert between different units if applicable.
+
+NOTE: Entity objects like descendants of `ApplicationRecord` are not value objects, as they have a distinct identity and
+lifecycle. They are uniquely identified by their primary keys or object ids.
+
+### Benefits of Using Value Objects
+
+Using value objects in your application can provide several benefits:
+
+* **Maintainability**: Value objects can help to keep your code organized and maintainable by separating concerns and
+    reducing duplication. Instead of juggling raw strings and integers everywhere, you use meaningful types.
+
+* **Immutability**: Value objects are typically immutable, which can help to prevent unintended side effects and make your code more predictable.
+
+* **Reduced Complexity**: Value objects can make your code more expressive and easier to understand by encapsulating related attributes and behavior.
+
+### Defining a Value Object
+
+A value object is typically defined as a plain Ruby class that includes the `Comparable` module. They are importantly
+immutable and do not let you change the value after creation.
+
+Instead, you need to create a new instance with the new value.
+The immutable requirement is enforced by Active Record by freezing any object assigned as a value object. Attempting to
+change it afterward will result in a `RuntimeError`.
+
+```ruby
+
+class Money
+  include Comparable
+  attr_reader :amount, :currency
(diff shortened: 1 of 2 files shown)


guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+to use a `Money` value object to represent the price of the product.
+
+You can define the `Product` model like this:
+
+```ruby
+class Product < ApplicationRecord
+  composed_of :price,
+              class_name: "Money",
+              mapping: [%w(price_cents amount), %w(price_currency currency)],
+              # mapping: {price_cents: :amount, price_currency: :currency}, # alternative syntax
```
MatheusRich: I know the docs often use the array form, but I think the hash syntax is much nicer and clearer. Could we use that as the default?

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+
+Querying Aggregated Attributes
+-----------------------------------------------
+
+You can make simple queries using the aggregated attributes.
+When you query using an aggregated attribute, Active Record will translate the query to use the underlying
+table columns.
+
+```ruby
+expensive_products  = Product.where("price_cents > ?", 1000)
```
MatheusRich: Shouldn't we be using the Money object here?

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+  composed_of :price,
+              class_name: "Money",
+              mapping: [%w(price_cents amount), %w(price_currency currency)]
+  validates   :price, presence: true
+  scope       :expensive, -> { where("price_cents > ?", 1000) }
+end
+```
+
+Since you cannot use aggregated attribute(like `price`) directly in database queries,
+you need to use the underlying table columns(like `price_cents`) in scopes.
```
MatheusRich: In the first paragraph you say

> You can also use the aggregated attributes in scopes and validations

And here you say that we have to use the underlying table columns. I'm confused what is the aggregation doing here. It seems to only work on validations.

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+
+Other Option Examples
+---------------------
+
+```ruby
+composed_of :temperature, mapping: { reading: :celsius }
+composed_of :balance, class_name: "Money", mapping: { balance: :amount }
+composed_of :address, mapping: { address_street: :street, address_city: :city }
+composed_of :address, mapping: [%w(address_street street), %w(address_city city)]
+composed_of :gps_location
```
MatheusRich, line 302: Don't we need mappings other options for this?

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+  scope       :expensive, -> { where("price_cents > ?", 1000) }
+end
+```
+
+Since you cannot use aggregated attribute(like `price`) directly in database queries,
+you need to use the underlying table columns(like `price_cents`) in scopes.
+
+Other Option Examples
+---------------------
+
```
MatheusRich, line 294: I think need a sentence before the code example. 
```suggestion

Here are some examples of value objects you might build using Active Record Aggregations in your app:

```

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+            mapping: { ip: :to_i },
+            constructor: Proc.new { |ip| IPAddr.new(ip, Socket::AF_INET) },
+            converter: Proc.new { |ip| ip.is_a?(Integer) ? IPAddr.new(ip, Socket::AF_INET) : IPAddr.new(ip.to_s) }
+```
+
+
+Caveats
+-------
+
+### Query with values in different units
```
MatheusRich: This is not exclusive to units. Value objects are identified by their _values_, so if you're using a different value, it's a different object.

The same would happen with a Distance object with kilometers vs miles.

A possible solution would be keep the db unit constant (dollars, meters, etc) and perform conversions/compasissons at the class level

guides/source/active_record_aggregations.md
```diff
@@ -0,0 +1,350 @@
...
+
+For example:
+
+```ruby
+product       = Product.find(1)
+product.price = Money.new(2000, "EUR")
+product.changed?                          # => true
+product.changes_to_save                   # => {"price_cents"=>[1000, 2000], "price_currency"=>["USD", "EUR"]}
+
+product.price_changed?                    # => NoMethodError (no direct price_changed?)
```
MatheusRich, line 402: Unrelated to this PR, but at first sight this seems doable. We could delegate methods back to the original model (if we have access to that)

MatheusRich (commented):
Love this addition!

guides/source/active_record_aggregations.md
shivabhusal: https://github.com/rails/rails/blob/529f933fc8b13114d308dd0752f76a9e293c8537/activerecord/lib/active_record/aggregations.rb#L225-L236

Yes, it is also possible. Mapping is derived from `part_id`, i.e. `:gps_location` and other options are optional.

guides/source/active_record_aggregations.md
shivabhusal: done!

guides/source/active_record_aggregations.md
shivabhusal: yes we should. done!

guides/source/active_record_aggregations.md
shivabhusal: made corrections

guides/source/active_record_aggregations.md
shivabhusal: accepted your suggestion

guides/source/active_record_aggregations.md
shivabhusal: I have added new descriptive example

guides/source/active_record_aggregations.md
shivabhusal: I will have a look later if we can  make such changes

MatheusRich (approved):

shivabhusal:
@byroot When you have a moment, I’d really appreciate your thoughts on this PR.

munabhandari:
Nice Addition, thanks! 👍

rafaelfranca:
@p8 any opinions here? Not very Rails feature needs a guide and it seems this isn't a common enough feature to need a guide. I don't see anything in this guide that isn't already covered in https://api.rubyonrails.org/classes/ActiveRecord/Aggregations/ClassMethods.html

p8:
> @p8 any opinions here? Not very Rails feature needs a guide and it seems this isn't a common enough feature to need a guide. I don't see anything in this guide that isn't already covered in https://api.rubyonrails.org/classes/ActiveRecord/Aggregations/ClassMethods.html

Yes, although nicely written, I'm not sure there is enough content here to justify its own guide. It's mostly about a single method with a limited number of options. Without all the examples and caveats the guide would be a bit smaller.

Maybe `composed_of` is a not commonly used method in part because it isn't described anywhere in the guides?
Maybe we could add a "Aggregations" section at the bottom of the "Active Record Basics" guide?
It could show a simple example and link to the API docs for more extensive examples and caveats.

MatheusRich:
Is there an opportunity to improve the API docs with something that was written here?

rafaelfranca:
> Is there an opportunity to improve the API docs with something that was written here?

Yes. I think so. @shivabhusal do you want to improve the API documentation?

shivabhusal:
@rafaelfranca sounds good! i will push a commit and let you know.

PR #55767, opened 2025-09-25
ActionText: Improved heading and table plain text

intrip:
### Motivation / Background

Table formatting in the ActionText plaintext conversions.

### Detail

- Treat all heading tags (`<h1>-<h6>`) as blocks, not just `<h1>`.
- Format tables - line per row, and a space between cells.

This is a minor formatting improvement and likely does not require a CHANGELOG entry, but I can add one if needed.


intrip:
The test failure looks unrelated.

PR #55769, opened 2025-09-25
Add replicas to test database parallelization setup

### LINKED ISSUE #55758: Parallel ActionDispatch::IntegrationTest tests, when using multiple databases and use_transactional_tests=false do not use the correct database
This is a really niche issue, but I was able to reproduce it in a sample app.

### Steps to reproduce

I've generated a [minimal sample app](https://github.com/a-maas/testapp) that reproduces this issue. The app has:

1. Multiple [postgresql databases](https://github.com/a-maas/testapp/blob/main/docker-compose.yml), [one primary, and one replica](https://github.com/a-maas/testapp/blob/main/config/database.yml#L16-L39).
2. Standard minitest, with the tests running in parallel so that there are `n` number of test databases for each parallel worker, i.e. `test_db-0` & `test_db-1`
3. The IntegrationTest class should use `use_transactional_tests=false`. This issue does not happen when `use_transactional_tests=true`, but assume that it's not possible to use `use_transactional_tests=true` in this case.

To reproduce with the sample app:

1. Pull down the sample app (requires docker for postgresql)
2. docker compose to get the databases working `docker compose up -d`
3. `./bin/setup`
4. Run the tests in parallel `PARALLEL_WORKERS=5 ./bin/rails test`
5. Observe that all the tests fail with 404

### Expected behavior

When running the tests in parallel, the IntegrationTest should use the same database for the controller request that it does for test setup.

i.e.
[Example test file](https://github.com/a-maas/testapp/blob/main/test/controllers/tests_controller_test.rb)
```
class TestsControllerTest < ActionDispatch::IntegrationTest
  self.use_transactional_tests = false
  test "should get show" do
    test = Test.create(name: "test") # Should use `test_db-{n}`
    get test_url(test)  # Should use `test_db-{n}`
    assert_response :success
  end
end
```

### Actual behavior

When running the tests in parallel, the IntegrationTest uses the numbered parallel databases, `test_db-{n}`, for test setup, but uses non-number test database, i.e. `test_db` for the controller request, leading to not found errors.

```
class TestsControllerTest < ActionDispatch::IntegrationTest
  self.use_transactional_tests = false
  test "should get show" do
    test = Test.create(name: "test") # Uses `test_db-{n}`
    get test_url(test)  # Uses `test_db`
    assert_response :success # 404
  end
end
```

### System configuration

**Rails version**: 8.0.3

**Ruby version**: 3.4.3

a-maas:
### Motivation / Background

Fixes #55758

### Detail

This Pull Request makes a change so that setup and configuration of databases for parallel testing now includes replicas.
    
This fixes an issue when using a replica database, database selector middleware, and non-transactional tests, where integration tests running in parallel would select the base test database, i.e. `db_test`, instead of the numbered parallel worker database, i.e. `db_test_{n}`.

### Additional information

The main issue revolves around how transactional tests work, and how they affect database selection when using a replica database. When using transaction tests, which is default, the test is wrapped in a transaction, which will default the entire transaction to the primary (writer) database. Because of this transaction, the test will use the primary database for the entire test, and the issue does not happen.

When you disable transactional tests, `self.use_transactional_tests = false`, this transaction is not present, so there is no initial/inherent database selection. In an integration test, the test setup code will use the primary database, but the controller request (GET, specifically) will run through the [database selector middleware](https://api.rubyonrails.org/classes/ActiveRecord/Middleware/DatabaseSelector.html), since we have this enabled when using a replica. The middleware will select the replica, because the test setup does not go through the middleware when doing the initial writes.

This creates an issue when running your tests in parallel, because replica databases do not get configured with a parallel worker number, i.e. `test_db-{n}`. That is because [configs_for](https://github.com/rails/rails/blob/v8.0.3/activerecord/lib/active_record/test_databases.rb#L14-L15) hides replica database configs unless you pas the `include_hidden: true` flag.

With the integration test running in parallel, when the database selector middleware selects the replica during the GET request, it'll use the base test database, i.e. `test_db`, not the numbered version, `test_db-{n}`, since these were not configured to use a parallel worker database. This ends up raising an error because the test setup was written to the parallel numbered database `test_db-{n}`, while the request reads from the base replica database `test_db`.

i.e. 
```ruby
class TestsControllerTest < ActionDispatch::IntegrationTest
  self.use_transactional_tests = false
  test "should get show" do
    test = Test.create(name: "test") # Uses `test_db-{n}`
    get test_url(test)               # Uses `test_db`
    assert_response :success         # 404
  end
end
```

To fix this issue, I've included replica databases in the parallel testing setup by passing the `include_hidden: true` flag to `configs_for`. Additionally, I skip `reconstruct_from_schema` for replicas, since these will replicate to replicas if using replication.

More information about this can be found in the original issue:  #55758 

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Add replicas to test database parallelization setup.
+
+    Setup and configuration of databases for parallel testing now includes replicas.
+
+    This fixes an issue when using a replica database, database selector middleware,
+    and non-transactional tests, where integration tests running in parallel would select
+    the base test database, i.e. `db_test`, instead of the numbered parallel worker database,
+    i.e. `db_test_{n}`.
+
+    *Adam Maas*
+
 *   Optimize schema dumping to prevent duplicate file generation.
 
     `ActiveRecord::Tasks::DatabaseTasks.dump_all` now tracks which schema files
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -19,10 +19,12 @@ module ActiveRecord
     def self.create_and_load_schema(i, env_name:)
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
-      ActiveRecord::Base.configurations.configs_for(env_name: env_name).each do |db_config|
+      ActiveRecord::Base.configurations.configs_for(env_name: env_name, include_hidden: true).each do |db_config|
         db_config._database = "#{db_config.database}_#{i}"
 
-        ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
+        if db_config.database_tasks?
+          ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
+        end
       end
     ensure
       ActiveRecord::Base.establish_connection
--- a/activerecord/test/cases/test_databases_test.rb
+++ b/activerecord/test/cases/test_databases_test.rb
@@ -101,5 +101,35 @@ class TestDatabasesTest < ActiveRecord::TestCase
       ActiveRecord::Base.establish_connection(:arunit)
       ENV["RAILS_ENV"] = previous_env
     end
+
+    def test_create_databases_after_fork_with_replica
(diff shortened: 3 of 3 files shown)


eileencodes:
Thanks!!

PR #37625, opened 2019-11-02
Add predicate methods for previous and next day checks

jatindhankhar:
### Summary

This add supports for checking time against previous day 
`prev_day? `/ `yesterday?`

and next day 

`next_day?` / `tomorrow?`  
#### Intention

While at work, we wanted to check if a date belonged to yesterday, to implement the business logic and were surprised this wasn't part of `activesuppport` 
[Relevant Stackoverflow question](https://stackoverflow.com/questions/26578553/check-if-datetime-value-is-today-tomorrow-or-later)

  
This completes the `today?` predicate with its complimentary helper methods `tomorrow?`, `next_day?`, `yesterday?` and `prev_day?` 

So people can simply use the helper predicates to check it the date in question belongs to the previous or next day.

PR #55770, opened 2025-09-25
Introduce `this_week?`, `this_month?`, and `this_year?` to Date/Time

MatheusRich:
Similar to `today?`, `tomorrow?`, and `yesterday?`, these methods are useful to
query time instances against the current period.

```ruby
unless some_date.in_this_week?
  link_to "See week recap", week_recap_path(some_date)
end
```

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Introduce `this_week?`, `this_month?`, and `this_year?` methods to Date/Time
+
+    Similar to `today?`, `tomorrow?`, and `yesterday?`, these methods are useful to
+    query time instances against the current period.
+
+    ```ruby
+    unless date.this_week?
+      link_to "See week recap", week_recap_path(date)
+    end
+    ```
+
+    *Matheus Richard*
+
 *   Removed the deprecated `ActiveSupport::Multibyte::Chars` class.
 
     As well as `String#mb_chars`
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -4,7 +4,7 @@
     query time instances against the current period.
 
     ```ruby
-    unless date.this_week?
+    unless post.created_at.this_week?
       link_to "See week recap", week_recap_path(date)
     end
     ```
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/core_ext/date_and_time/calculations.rb
```diff
@@ -63,6 +63,24 @@ def on_weekday?
       !WEEKEND_DAYS.include?(wday)
     end
 
+    # Returns true if the date/time falls within the current week.
+    def this_week?
+      ::Date.current.all_week.include?(to_date)
```
MatheusRich, line 68: I thought comparing as date made sense for these units. Unless we plan to add `current_hour?` too, which then it would be best to convert `to_time` instead.

p8:
A comment about a similar PR to add methods like `yesterday?` probably applies here as well:
> New Active Support core extensions that aren't used internally by Rails are often rejected in favour of being proposed to Ruby directly, but since these complement existing Active Support methods it seems reasonable to consider adding them here. That decision will have to be made by a committer though.

https://github.com/rails/rails/pull/37625#issuecomment-597903034

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+*   Introduce `in_this_week?`, `in_this_month?`, and `in_this_year?` methods to Date/Time
+
+    Similar to `today?`, `tomorrow?`, and `yesterday?`, these methods are useful to
+    query time instances against the current period.
+
+    ```ruby
+    unless some_date.in_this_week?
+    link_to "See week recap", week_recap_path(some_date)
+    end
```
felixefelip: It looks like there's a missing spacing
```suggestion
    unless some_date.in_this_week?
      link_to "See week recap", week_recap_path(some_date)
    end
```

MatheusRich:
@jeremy @dhh sorry to ping y'all directly, I'd love a thumbs up/down on this. We have simimilar query methods, and I'm assuming you have experience with similar methods because of HEY.

dhh:
I prefer skipping the aliases. We already use #in_ for conversions, like #in_minutes. But #this_week? matches #today? So I'm good with this without the #in_ aliases 👍

MatheusRich:
@dhh Done. Thank you!

PR #55774, opened 2025-09-26
Remove Code of Conduct

jclusso:
This only serves to give legitmacy to the woke regime so I think we should remove it. (related: https://github.com/rails/website/pull/577)


dhh:
The former code of conduct was indeed removed. We used to use the Contributor Covenant. Now we have have a set of guidelines that Matz picked for Ruby, which I think are great. You can see the diff here: https://github.com/rails/website/commit/0351bc6710a59bab94d07bf65a31769b3c9cd818

PR #55777, opened 2025-09-26
Add `Date.upcoming` method to find next occurrence of month/day

elalemanyo:
### Motivation / Background

This Pull Request has been created because developers frequently need to calculate the next occurrence of recurring dates (holidays, birthdays, anniversaries) and currently have to manually handle year transitions and leap year edge cases. There's no convenient method in Rails to find "the next Christmas" or "the next February 29th" that properly handles all the calendar complexities.

### Detail

This Pull Request changes ActiveSupport's Date class by adding a new `Date.upcoming(month:, day:)` class method that:

- Returns the next occurrence of a specific month and day
- Automatically handles year transitions (if the date has passed this year, returns next year)
- Properly handles leap year edge cases (e.g., Feb 29 in non-leap years finds the next leap year)
- Validates input parameters to prevent infinite loops and provide clear error message
- Respects time zone configuration when `Time.zone` is set
- Uses keyword arguments for international clarity and consistency with Rails conventions (matches `advance` method naming)

**Usage examples:**
```ruby
Date.upcoming(month: 12, day: 25)  # => Next Christmas (Dec 25)
Date.upcoming(month: 1, day: 1)    # => Next New Year's Day (Jan 1)
Date.upcoming(month: 2, day: 29)   # => Next Feb 29 (finds next leap year if needed)

# Invalid parameters raise clear errors
Date.upcoming(month: 13, day: 1)   # => ArgumentError: Invalid date: month 13, day 1
Date.upcoming(month: 2, day: 30)   # => ArgumentError: Invalid date: month 2, day 30
```

### Additional information

**Edge cases handled:**

- Leap years: Correctly finds next Feb 29 even from non-leap years, including century year rules (2100 → 2104)
- Multiple year jumps: Efficiently handles cases requiring several years to find valid date
- Time zones: Uses Time.zone when configured, consistent with other Date class methods
- Same day matching: Returns current date if it matches the requested month/day

**Test coverage includes:**

- Basic functionality (current year vs next year)
- Leap year scenarios (simple and complex cases)
- Time zone handling (with and without Time.zone set)
- Edge cases (end of year, beginning of year, century years)
- Multiple year jumps (e.g., 2021 → 2024 for Feb 29)
- Parameter validation (invalid months, days, and combinations)

**Implementation uses:**

- Early parameter validation using `Date.new(2000, month, day) to test validity
- Proper error handling for invalid dates with `Date::Error rescue
- Clear `ArgumentError` messages for invalid input parameters
- Safe navigation with `&.` for time zone handling

### Checklist


sevos:
I would just name it `Date.upcoming`
Also, I am not sure if the month as the first argument is aligned with Rails' conventions (honestly don't remember) - something to look at.

wintermeyer:
For me it somehow feels half without the opposite in the past.

activesupport/lib/active_support/core_ext/date/calculations.rb
```diff
@@ -48,6 +48,35 @@ def tomorrow
...
       ::Time.zone ? ::Time.zone.today : ::Date.today
     end
+
+    # Returns the upcoming date for the specified month and day.
+    # If the date has already passed this year, returns the date for next year.
+    # Uses the configured time zone when <tt>Time.zone</tt> is set.
+    #
+    #   Date.upcoming_date(12, 25)  # => Date for next Christmas (Dec 25)
+    #   Date.upcoming_date(1, 1)    # => Date for next New Year's Day (Jan 1)
+    def upcoming_date(month, day)
```
MatheusRich: The argument ordering doesn't make sense right away IMO. I'm not from the US, so I don't expect month first. Can we make these kwargs instead? If we go that route, maybe match the same argument names we use in `advance`? I think we use `mon/day`

activesupport/lib/active_support/core_ext/date/calculations.rb
```diff
@@ -48,6 +48,35 @@ def tomorrow
...
+      today = ::Time.zone&.now || ::Time.now
+
+      begin
+        current_year_date = ::Date.new(today.year, month, day)
+        return current_year_date if current_year_date >= today.to_date
+      rescue Date::Error
+        # Invalid date for current year (e.g., Feb 29 in non-leap year)
+      end
+
+      (today.year + 1..Float::INFINITY).lazy.each do |year|
```
MatheusRich: How does this work? You're adding an integer and a range?

activesupport/lib/active_support/core_ext/date/calculations.rb
elalemanyo: Great feedback! You're absolutely right about the international perspective and argument clarity. I've updated the method to use keyword arguments with the same naming convention as the `advance` method.

```ruby
Date.upcoming_date(month: 12, day: 25)   # Christmas
Date.upcoming_date(month: 1, day: 1)     # New Year's Day
Date.upcoming_date(month: 2, day: 29)    # Next leap year Feb 29
```

This eliminates the US date format assumption and makes the API much clearer for international developers. The parameter order is now irrelevant, and it's immediately obvious what each value represents. Thanks for the suggestion!

MatheusRich:
@wintermeyer maybe `previous`?

activesupport/lib/active_support/core_ext/date/calculations.rb
elalemanyo: The line is not adding an integer and a range. Here's what's actually happening:

1. `today.year + 1` - This evaluates to an integer (e.g., if today is 2025, this becomes 2026)

2. `(2026..Float::INFINITY)` - This creates a range from 2026 to infinity

3. `.lazy.each` - This lazily iterates through each year in that range, evaluates only when needed.

The parentheses are crucial here - they ensure the addition happens first, then the range is created from that result to infinity. It's equivalent to:

```ruby
start_year = today.year + 1
(start_year..Float::INFINITY).lazy.each do |year|
```

So we're creating a range from "next year" to infinity, not adding an integer to a range. The method uses this to search future years until it finds a valid date for the given month/day combination (handling edge cases like Feb 29 in non-leap years).

MatheusRich: Oh I see. Can we do `(today.year + 1)..` or `today.next_year.year..`? I think the INFINITY can be omited with an end-less range.

elalemanyo:
Regarding the `previous_date` method suggestion:

I don't see strong justification for adding a `previous_date` counterpart at this time 😬. Here's why:

- **Limited use cases:** While `upcoming_date` serves clear planning needs (scheduling, reminders, deadlines), `previous_date` scenarios are much narrower and less common in typical applications.

- `Simpler alternatives exist:` Past dates can usually be calculated more directly with existing methods (e.g., `Date.new(year - 1, month, day)` for most cases).
Maybe we could implement `previous_date` using `upcoming_date` as a foundation 🤔. This would be much simpler and reuse the existing logic:

```ruby
def previous_date(month:, day:)
  upcoming = upcoming_date(month: month, day: day)

  (upcoming.year - 1).downto(upcoming.year - 4).each do |year|
    return ::Date.new(year, month, day)
  rescue Date::Error
    next
  end
end
```

- **Complexity vs. benefit:** A `previous_date` method would require similar complexity (handling leap years, backward iteration) but for significantly fewer real-world use cases.

- **API design principle:** Symmetry isn't always valuable - the usage patterns for future vs. past date lookups are fundamentally different. Future planning is proactive and common; past lookups are typically reactive and infrequent.

I'd suggest waiting for concrete user demand with specific use cases before adding this. If users can't immediately articulate why they need it beyond "it would be symmetrical," that usually indicates the method isn't solving a real problem.

I'm adding `upcoming_date` because it's a method I've genuinely needed in past projects. However, I'm absolutely open to adding `previous_date` if there are real, documented needs for it - I just haven't encountered those use cases myself yet 😅

activesupport/lib/active_support/core_ext/date/calculations.rb
elalemanyo: @MatheusRich Maybe you're right, and it's better to just do this: `(today.year + 1..today.year + 3)`

elalemanyo:
Following the suggestions, I have changed the name to `upcoming` 🙂

p8:
Can't this already be solved with `advance`?
```ruby
Date.new(2004, 9, 30).advance(years: 1)
# => Sun, 9 Sept 2005
```

elalemanyo:
@p8 `advance` works for relative changes like `this date + N years/months/days`, but not for finding the next occurrence of a fixed month/day.

For example, if today is October 5 and you want the next September 30, advance can’t express that — you have to manually check if it’s already passed this year and handle leap years like Feb 29 yourself.

`Date.upcoming` wraps that common logic (year rollover, leap year handling, time zone) into a single, intention-revealing method.

rafaelfranca:
@jorgemanrubia you built a calendar application, so you might help me to decide if this is useful or not. Did you see the need for methods like this?

I'm asking because I'm not sure this belongs to Active Support given I don't see a lot of use cases.

jorgemanrubia:
@rafaelfranca In HEY Calendar we don't have this need, but probably because the app supports iCal subscriptions, which is what you usually use to track holidays and such. I like the method, though. Active Support date extensions have many methods that you rarely use, but when you find them they are super nice to have (like `#tomorrow?` or `#future?`). I think this is a good fit for the API.

MatheusRich:
@jorgemanrubia @rafaelfranca I'd love feedback on a similar PR: https://github.com/rails/rails/pull/55770

p8:
I would expect this to work as an instance method on any date (not just today), similar to [next_occuring](https://edgeapi.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-next_occurring).
```ruby
today = Date.today               # => Thu, 14 Dec 2017
today.next_occurring(:monday)    # => Mon, 18 Dec 2017
today.next_occurring(:thursday)  # => Thu, 21 Dec 2017
```
Maybe that method should allow keyword arguments?
```ruby
today = Date.parse('2017-10-14')
today.next_occurring(month: 12, day: 25)    # => 25 Dec 2017
today.next_occurring(month: 1, day: 1)    # => 1 Jan 2016
# This would also allow chaining:
today.
  next_occurring(month: 12, day: 25).
  next_occurring(month: 12, day: 25)   # => 25 Dec 2018
```

rafaelfranca:
I like @p8's suggestion. I prefer that than a class method

elalemanyo:
@p8 @rafaelfranca Should I use the `next_occurring` method and enable keyword arguments?

PR #54273, opened 2025-01-17
Prevent persisting invalid record:

Edouard-chin:
Prevent persisting invalid record:

- Fix #54267

### Motivation / Background

- In #53951, the goal was to allow saving a record and its association even if the association had existing invalid records in the DB. We would not validate the association in that case.

  The problem of not validating, is that it stops the validation chain and it's possible that other records in subsequent relations are being modified and are invalid. e.g. `Company -has_many-> Developers -has_many-> Computers`, if a computer is changed with invalid value and the company get saved, this would bypass the computer validation and persist it.

  This commit fixes that by ensuring there are no validation errors from *changed* records in the whole association chain.

This Pull Request has been created because [REPLACE ME]

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #53951, opened 2024-12-14
Do not validate associated records that haven't changed

byroot:
Fix: https://github.com/rails/rails/issues/17621
Redo of: https://github.com/rails/rails/pull/18612

You should be able to create a reference to an invalid record as long as you don't modify it.

This was initially fixed in https://github.com/rails/rails/pull/18612 but was reverted in https://github.com/rails/rails/commit/2c02bc0a47777ad8cf98e1465c08b1a68151803e because it was causing other bugs.

Some tests were added as part of the revert to avoid future regressions.

PR #55778, opened 2025-09-26
Prefer changed_for_autosave? [Fix #55771]

### LINKED ISSUE #55771: Rails 8.0.2+ saves invalid records when required distantly associated records are marked for deletion (is correct on 8.0.1)
### Steps to reproduce

Have a model relationship with a `foo --* bar --* baz` like structure where nested attributes are accepted for the deeper models. Delete a distant model `baz` via nested attributes when `bar`'s validations require it not be deleted.

In this case, the instance of `bar` will be invalid, and therefore `foo` will be invalid, but the deletion will occur anyway in Rails 8.0.2 and up. This leads to an invalid model structure when it was valid before.

The deletion does not occur in Rails 8.0.1 and validity is preserved.

A reproduction script follows, where the models are named `Parent`, `Child`, and `Grandchild` for simplicity, but I do not believe the tree-like structure is necessary.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # The tests pass in 8.0.1
  # gem "rails", "8.0.1"

  # The tests fail in 8.0.2
  gem "rails", "8.0.2"

  # and in 8.0.3
  # gem "rails", "8.0.3"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end

    t.integer :parent_id
  end

  create_table :grandchildren, force: true do |t|
    t.integer :child_id
  end
end

  has_many :children
  accepts_nested_attributes_for :children, allow_destroy: true
end

class Child < ActiveRecord::Base
  belongs_to :parent
  has_many :grandchildren
  accepts_nested_attributes_for :grandchildren, allow_destroy: true

  validate :has_at_least_two_grandchildren

  # This is a workaround for 8.0.2:
  # def changed?
  #   grandchildren.any? { |gc| gc.marked_for_destruction? } || super
  # end

  private

  def has_at_least_two_grandchildren
    current_grandchildren = grandchildren.select { |gc| !gc.marked_for_destruction? }
    errors.add(:grandchildren, "must have at least two grandchildren") if current_grandchildren.size < 2
  end
end

class Grandchild < ActiveRecord::Base
  belongs_to :child
end

  setup do
    @parent = Parent.create!
    @child = @parent.children.new
    @grandchild1 = @child.grandchildren.new
    @grandchild2 = @child.grandchildren.new
    @child.save!

    @deletion_params = {
      "children_attributes" => {
        "0" => {
          "id" => @child.id,
          "grandchildren_attributes" => {
            "0" => {
              "id" => @grandchild1.id,
              "_destroy" => '1',
            },
          }
        }
      }
    }
  end

  test "starts out in a valid state" do
    assert_predicate @parent, :valid?
    assert_predicate @child, :valid?
    assert_predicate @grandchild1, :valid?
    assert_predicate @grandchild2, :valid?
  end

  test "cannot update a parent with invalid child via deleting a grandchild" do
    refute @parent.update(@deletion_params), "Parent should not update successfully"
  end

  test "failed update should not delete grandchild" do
    @parent.update(@deletion_params)
    assert_nothing_raised do
      @grandchild1.reload
    end
  end

  test "child should be invalid after failed update" do
    @parent.update(@deletion_params)
    refute_predicate @child, :valid?
    assert_includes @child.errors[:grandchildren], "must have at least two grandchildren"
  end

  test "parent should be invalid when child is invalid after failed update" do
    @parent.update(@deletion_params)
    refute_predicate @parent, :valid?
    assert_includes @parent.errors[:"children.grandchildren"], "must have at least two grandchildren"
  end
end
```

Note that you can change the version to 8.0.1 and see that the tests pass. There is also a workaround commented out that can be uncommented to have the tests pass in 8.0.2 and above. 

It also fails in 8.0.3, but 8.0.2 is used a reference since that is the first failing version.

### Expected behavior

All tests should pass as they do in Rails 8.0.1.

In this case, the `Parent` instance should not be saved, because its existing `Child` instance is now invalid. The child is properly validated and expresses that it is invalid through adding an error.

### Actual behavior

The `Parent` is saved, its `Child` instance is saved, and the `Grandchild` instance is deleted. Now the `Child` is invalid and the `Parent` is invalid, despite all records starting as valid.

### Notes

This behavior was changed in

- https://github.com/rails/rails/pull/53951 (cc @byroot)
- https://github.com/rails/rails/pull/54273 (cc @Edouard-chin)

Notably it is caused by this in `activerecord/lib/active_record/autosave_association.rb`:

```ruby
if record.changed? || record.new_record? || context
  associated_errors = record.errors.objects
else
  # If there are existing invalid records in the DB, we should still be able to reference them.
  # Unless a record (no matter where in the association chain) is invalid and is being changed.
  associated_errors = record.errors.objects.select { |error| error.is_a?(Associations::NestedError) }
end
```

The issue is there are _not_ existing invalid records in the database. Unfortunately, the change is occurring via an association with nested attributes and this is not reported via dirty tracking.

As a workaround, we can modify our own code like so:

```ruby
def changed?
  grandchildren.any? { |gc| gc.marked_for_destruction? } || super
end
```

But this is surprising behavior for a patch release since it worked through 8.0.1.

### System configuration

**Rails version**: 8.0.2+

**Ruby version**: 3.4.1

ianterrell:
Fix: https://github.com/rails/rails/issues/55771

- The original code intends to avoid already persisted invalid records
  from preventing new associated records from being saved.
- Issue #55771 documents that this could lead to previously valid data
  being saved and becoming invalid.
- The code originated in https://github.com/rails/rails/pull/53951
- And was updated in https://github.com/rails/rails/pull/54273
- The solution suggested by @axlekb in https://github.com/rails/rails/pull/53951#pullrequestreview-3272124308

### Motivation / Background

This Pull Request has been created because 8.0.2 introduced a regression where previously valid models could become invalid through autosave when a distantly related record was deleted via nested attributes.

Bug reported in

- #55771 

### Detail

This Pull Request changes `changed?` to `changed_for_autosave?` at suggestion of @axlekb in discussion with @byroot.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/autosave_association.rb
+++ b/activerecord/lib/active_record/autosave_association.rb
@@ -374,7 +374,7 @@ module ActiveRecord
         context = validation_context if custom_validation_context?
         return true if record.valid?(context)
 
-        if record.changed? || record.new_record? || context
+        if context || record.changed_for_autosave?
           associated_errors = record.errors.objects
         else
           # If there are existing invalid records in the DB, we should still be able to reference them.
--- a/activerecord/test/cases/autosave_association_test.rb
+++ b/activerecord/test/cases/autosave_association_test.rb
@@ -2443,3 +2443,58 @@ class TestAutosaveAssociationOnABelongsToAssociationDefinedAsRecord < ActiveReco
     end
   end
 end
+
+class TestAutosaveAssociationWithNestedAttributes < ActiveRecord::TestCase
+  class Part < ActiveRecord::Base
+    self.table_name = "ship_parts"
+  end
+
+  class Ship < ActiveRecord::Base
+    self.table_name = "ships"
+    has_many :parts, class_name: Part.name
+    accepts_nested_attributes_for :parts, allow_destroy: true
+
+    validate :has_at_least_two_parts
+    def has_at_least_two_parts
+      current_parts = parts.select { |p| !p.marked_for_destruction? }
+      errors.add(:parts, "must have at least two parts") if current_parts.size < 2
+    end
+  end
+
+  class Pirate < ActiveRecord::Base
+    self.table_name = "pirates"
+    has_many :ships, class_name: Ship.name
+    accepts_nested_attributes_for :ships, allow_destroy: true
+  end
+
+  def test_should_be_invalid_when_nested_attributes_deletion_breaks_validation
+    pirate = Pirate.create!
+    ship = pirate.ships.new
+    2.times do |i|
+      ship.parts.build
+    end
+    part = ship.parts.first
+    ship.save!
+
+    deletion_params = {
+      "ships_attributes" => {
+        "0" => {
+          "id" => ship.id,
+          "parts_attributes" => {
+            "0" => {
+              "id" => part.id,
+              "_destroy" => "1",
+            },
+          }
+        }
+      }
+    }
+
+    assert_not pirate.update(deletion_params)
+    assert_nothing_raised do
+      part.reload
+    end
+    assert_includes pirate.errors[:"ships.parts"], "must have at least two parts"
+    assert_includes ship.errors[:parts], "must have at least two parts"
+  end
+end


byroot:
Not too sure what's going on with the connection pool tests, But liekly unrelated to your PR. The rubocop failure is legit though.

activerecord/lib/active_record/autosave_association.rb
```diff
@@ -374,7 +374,7 @@ def association_valid?(association, record)
         context = validation_context if custom_validation_context?
         return true if record.valid?(context)
 
-        if record.changed? || record.new_record? || context
+        if record.changed_for_autosave? || record.new_record? || context
```
byroot:
```suggestion
        if context || record.changed_for_autosave?
```

Two nitpicks:

  - `changed_for_autosave?` is `new_record? || has_changes_to_save? || ..`, so no need to check for new_record again.
  - Best to check the cheaper condition first.

ianterrell: Suggestion integrated, and I think I have fixed the linting errors as well.

byroot:
Thank you very much, I really didn't have the energy to look into that one today.

byroot:
Backported to 8-0-stable and 7-2-stable.

PR #54661, opened 2025-03-01
Fix Horizontal Scroll on Mobile [ci skip]

etmartinkazoo:
Closes #54590

### Motivation / Background

On mobile devices, the Rails Guides have horizontal scroll which is not ideal for responsive design. C.f. [See here](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#follow-the-coding-conventions) on a mobile device as an example.

### Solution

This can be fixed by adjusting the width, white-space and overflow attributes on .code and .code > pre respectively.

PR #55782, opened 2025-09-27
Remove the scroll overflow on code blocks:

### LINKED ISSUE #55781: Padding on the code block looks off
### Steps to reproduce

Navigate to any page on the rails documentation site.

For example: https://edgeguides.rubyonrails.org/getting_started.html

Image

### Expected behavior

Code inside code block doesn't render the white background on the bottom-right.

### Actual behavior

Code block is adding an odd-padding at the bottom-right.

### System configuration

**Rails version**: n/a

**Ruby version**: n/a

Edouard-chin:
### Motivation / Background

Fixes #55781

We introduced a change in #54661 to prevent overflowing codeblocks into other elements.

This change creates white scroll bars on MacOS (maybe other platfroms too) only when you connect a mouse, it's a behaviour of MacOS that it displays scroll bar on elements.

This commit removes the overflow property without affecting overflow problem, that is because the PR linked above restrained the element to a 100% width.

### Before

image

### After

image

### After on small screens

image

### DIFF
--- a/guides/assets/stylesrc/components/_code-container.scss
+++ b/guides/assets/stylesrc/components/_code-container.scss
@@ -2,7 +2,7 @@
 // Containers
 //
 // These are interstitial elements used throughout the guides, providing help,
-// context, more info, or warnings to readers. 
+// context, more info, or warnings to readers.
 // ----------------------------------------------------------------------------
 
 /* Same bottom margin for special boxes than for regular paragraphs, this way
@@ -49,7 +49,6 @@ dl dd .interstitial {
 
     pre {
       margin: 0;
-      overflow: scroll;
       white-space: pre;
     }
 


frou:
The CSS before the original PR was the following including a comment:

`overflow: visible; // allows for the blue highlight to be seen`

[Here's an example](https://edgeguides.rubyonrails.org/getting_started.html#showing-individual-products) of where the blue marks on the left are used to call attention to certain lines. Are they still working with this PR or have they been lost in the shuffle?

Screenshot 2025-09-27 at 09 47 31

mikeygough:
@Edouard-chin I think you can bring back the left blue border by dropping the margin-left and padding-left here:

https://github.com/rails/rails/blob/cf6b3107753a762b27bfcf9eefcfd9feb43f0242/guides/assets/stylesrc/highlight.scss#L66

For example: 

margin-padding-removal-blue-left-border-fix

Edouard-chin:
Thanks @mikeygough ! I'll have a look

PR #55786, opened 2025-09-27
[Fix #55776] `class_attribute` on instance singleton class raises `NameError`

joshuay03:
### Motivation / Background

Closes #55776.

### Detail

When `class_attribute` is defined on an instance's singleton class (e.g., `object.singleton_class.class_attribute :foo`), accessing the attribute through the instance (`object.foo`) raises a `NameError` because the generated instance method tries to call `__class_attr_foo` as an instance method, but this method only exists as a private class method on the singleton class.

This fix changes the singleton class instance reader generation from `"def #{name}; #{namespaced_name}; end"` to `"def #{name}; self.singleton_class.#{name}; end"`, which properly delegates to the singleton class's public interface. The change preserves all existing behavior (including when `class_attribute` is defined on a class's singleton class) while fixing the `NameError` for instance singleton class attributes.

### Additional information

Also see https://github.com/rails/rails/issues/55776#issuecomment-3341321483.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/class/attribute.rb
+++ b/activesupport/lib/active_support/core_ext/class/attribute.rb
@@ -96,14 +96,16 @@ class Class
       namespaced_name = :"__class_attr_#{name}"
       ::ActiveSupport::ClassAttribute.redefine(self, name, namespaced_name, default)
 
-      delegators = [
-        "def #{name}; #{namespaced_name}; end",
-        "def #{name}=(value); self.#{namespaced_name} = value; end",
-      ]
+      class_methods << "def #{name}; #{namespaced_name}; end"
+      class_methods << "def #{name}=(value); self.#{namespaced_name} = value; end"
 
-      class_methods.concat(delegators)
       if singleton_class?
-        methods.concat(delegators)
+        methods << <<~RUBY if instance_reader
+          silence_redefinition_of_method(:#{name})
+          def #{name}
+            self.singleton_class.#{name}
+          end
+        RUBY
       else
         methods << <<~RUBY if instance_reader
           silence_redefinition_of_method def #{name}
--- a/activesupport/test/core_ext/class/attribute_test.rb
+++ b/activesupport/test/core_ext/class/attribute_test.rb
@@ -116,7 +116,7 @@ class ClassAttributeTest < ActiveSupport::TestCase
     assert_equal "plop", @klass.setting
   end
 
-  test "when defined in a class's singleton" do
+  test "when defined in a class's singleton class" do
     @klass = Class.new do
       class << self
         class_attribute :__callbacks, default: 1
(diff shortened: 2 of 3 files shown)


activesupport/test/core_ext/class/attribute_test.rb
```diff
@@ -130,6 +130,26 @@ class << self
...
+    @klass.singleton_class.__callbacks = 4
+    assert_equal 4, @klass.__callbacks
+    assert_equal 4, @klass.singleton_class.__callbacks
+  end
+
+  test "when defined on an instance's singleton class" do
+    object = @klass.new
+
+    object.singleton_class.class_attribute :external_attr, default: "default_value"
+    assert_equal "default_value", object.external_attr
```
joshuay03, line 143: On main:
```
Error:
ClassAttributeTest#test_when_defined_on_an_instance's_singleton_class:
NameError: undefined local variable or method '__class_attr_external_attr' for #<#<Class:0x00000001289fb5e0>:0x0000000128b7bac8>
    test/core_ext/class/attribute_test.rb:142:in 'external_attr'
    test/core_ext/class/attribute_test.rb:143:in 'block in <class:ClassAttributeTest>'
```

byroot:
The failure is legit.

```
/rails/activesupport/test/core_ext/class/attribute_test.rb:142: warning: method redefined; discarding old external_attr
/rails/activesupport/lib/active_support/class_attribute.rb:11: warning: previous definition of external_attr was here
E
Error:
ClassAttributeTest#test_when_defined_on_an_instance's_singleton_class:
RailsStrictWarnings::WarningError: /rails/activesupport/test/core_ext/class/attribute_test.rb:142: warning: method redefined; discarding old external_attr
/rails/activesupport/lib/active_support/class_attribute.rb:11: warning: previous definition of external_attr was here
    /rails/tools/strict_warnings.rb:35:in 'RailsStrictWarnings#warn'
    test/core_ext/class/attribute_test.rb:142:in 'Class#class_attribute'
    lib/active_support/core_ext/class/attribute.rb:131:in 'Module#class_eval'
    lib/active_support/core_ext/class/attribute.rb:131:in 'Class#class_attribute'
    test/core_ext/class/attribute_test.rb:142:in 'block in <class:ClassAttributeTest>'
bin/rails test /rails/activesupport/test/core_ext/class/attribute_test.rb:139
```

activesupport/lib/active_support/core_ext/class/attribute.rb
```diff
@@ -96,14 +96,16 @@ def class_attribute(*attrs, instance_accessor: true,
...
 
-      class_methods.concat(delegators)
       if singleton_class?
-        methods.concat(delegators)
+        methods << <<~RUBY if instance_reader
+          silence_redefinition_of_method(:#{name})
+          def #{name}
+            self.singleton_class.#{name}
+          end
+        RUBY
```
joshuay03, line 108: The writer that was previously added here shouldn't be needed. It is correctly set below when `instance_writer` is set. Also, previously, the reader was added unconditionally when it should be checking `instance_reader`.

Note however that this reader definition preserves some confusing behaviour, see: https://github.com/rails/rails/issues/55776#issuecomment-3341321483.

joshuay03:
> The failure is legit.

I've pushed a fix 👍🏽

activesupport/lib/active_support/core_ext/class/attribute.rb
byroot: Yeah, `class_attribute` behavior on singleton classes always has been very weird. Honestly I think we should probably deprecate that usage, but no isn't the right time.

byroot:
Backported to 8-0-stable.

PR #53375, opened 2024-10-19
Deprecate setting `config.active_job.enqueue_after_transaction_commit`

rafaelfranca:
Also change the configuration to be a boolean instead of a symbol.

The value of this setting isn't delegated to the queue adapter anymore, since the behavior that this setting controls is too dangerous to leave this change of behavior as something that users don't control.

Fixes #52675.

PR #55788, opened 2025-09-27
[Activejob] Enable enqueue_after_transaction_commit by default with ActiveRecord and validate invalid configurations

mugitti9:
### Motivation / Background
Until now, the default value of ```enqueue_after_transaction_commit``` has been false.

This meant that jobs enqueued inside a transaction were dispatched to the queue immediately and could still run even if the transaction was rolled back. In practice, this is not a rare edge case—jobs that depend on uncommitted data often end up running prematurely, breaking application consistency.

With this change, the default is switched to true only when ActiveRecord is defined. This ensures that jobs are enqueued only after the surrounding transaction has successfully committed, preventing jobs tied to rolled-back work from being executed unnecessarily.

As DHH also stated in [#53375](https://github.com/rails/rails/pull/53375#issuecomment-2555694252)
```Yes, by default, all jobs should enqueue after the commit.``` 
I strongly agree with this position. At the same time, this change does not remove flexibility: if immediate enqueueing is still desired, it can be explicitly enabled by setting ```self.enqueue_after_transaction_commit = false``` on a per-job basis.

### Detail

- The default value of ```enqueue_after_transaction_commit``` is now set to true only when ActiveRecord is available.
- A validation has been added: attempting to set ```enqueue_after_transaction_commit = true``` when ActiveRecord is not defined will now raise an error, since such a configuration would not have meaningful semantics without transaction hooks.

### Why this change is needed (what problem it solves)
With the current behavior on main, jobs can be enqueued before the surrounding database transaction has successfully committed. For example:

```rb
  user = User.create!(name: "James")
  UserCreationNotifyJob.perform_later(user:)
end
```

In this case, UserCreationNotifyJob may be enqueued while the user record is still uncommitted. If the transaction is rolled back, the job can still run and attempt to operate on data that does not actually exist. This leads to subtle and easily overlooked inconsistencies.

Until Rails 7.2, most adapters—Sidekiq, Resque, and Solid Queue—effectively behaved as if enqueue_after_transaction_commit = true by default. As a result, many applications implicitly relied on jobs being enqueued only after the transaction had fully committed. With the current default on main, however, jobs may run earlier than expected, and this can introduce hard-to-debug issues.

In our application, we currently mitigate this by explicitly setting enqueue_after_transaction_commit = true on almost all jobs. But this shifts responsibility for a fundamental safety guarantee from the framework to individual applications, increases boilerplate, and risks inconsistencies when someone forgets to set it.

This issue is not theoretical or rare. For example, at a recent Ruby/Rails conference in Japan ("Kaigi on Rails"), there was a well-received talk titled “Never enqueue async jobs inside a transaction—seriously, never!”, which highlighted exactly this class of problems. The talk resonated widely, which shows how many developers have been impacted by premature job execution.

### Additional information

When ActiveRecord is not defined, transaction callbacks are not available, and therefore enabling ```enqueue_after_transaction_commit``` would result in misleading or inconsistent behavior. Explicit validation has been added to surface this misuse early and clearly.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
@@ -18,9 +18,7 @@
 # commits before being dispatched to the queue. This prevents jobs from
 # operating on rolled-back data.
 #++
-# ActiveSupport.on_load(:active_job) do
-#   ActiveJob::Base.enqueue_after_transaction_commit = true
-# end
+# Rails.application.config.active_job.enqueue_after_transaction_commit = true
 
 ###
 # Use Sec-Fetch-Site header for CSRF protection instead of tokens.
--- a/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
@@ -9,6 +9,19 @@
 # Read the Guide for Upgrading Ruby on Rails for more info on each option.
 # https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
 
+# Active Job
+
+###
+# Defer job enqueueing until after the database transaction commits.
+#
+# When enabled, jobs enqueued inside a transaction wait until the transaction
+# commits before being dispatched to the queue. This prevents jobs from
+# operating on rolled-back data.
+#++
+# ActiveSupport.on_load(:active_job) do
+#   ActiveJob::Base.enqueue_after_transaction_commit = true
+# end
+
 ###
 # Use Sec-Fetch-Site header for CSRF protection instead of tokens.
 #
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Jobs are now enqueued after transaction commit.
+
+    This fixes that jobs would surprisingly run against uncommitted and
+    rolled-back records.
+
+    New Rails 8.2 apps (and apps upgrading to `config.load_defaults "8.2"`)
+    default `ActiveJob::Base.enqueue_after_transaction_commit` to true.
+    Uncomment the setting in `config/initializers/new_framework_defaults_8_2.rb`
+    to opt in.
+
+    *mugitti9*
+
 *   Deprecate built-in `sneakers` adapter.
 
     *Dino Maric*
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -4,11 +4,18 @@
     rolled-back records.
 
     New Rails 8.2 apps (and apps upgrading to `config.load_defaults "8.2"`)
-    default `ActiveJob::Base.enqueue_after_transaction_commit` to true.
+    have `config.active_job.enqueue_after_transaction_commit = true` by default.
     Uncomment the setting in `config/initializers/new_framework_defaults_8_2.rb`
     to opt in.
 
     *mugitti9*
+
+*   Un-deprecate the global `config.active_job.enqueue_after_transaction_commit`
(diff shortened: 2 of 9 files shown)


jeremy (commented):
-1 on the AR dep. Would either enqueue immediately if AR is unavailable, gracefully behaving as anyone would expect, or raise an exception indicating that enqueue_after_transaction_commit should be set to false.

mugitti9:
@jeremy 
Thanks for your review and for pointing that out!

You're absolutely right — adding ActiveRecord as a hard dependency was too much.
I've updated the implementation based on your suggestion:

- The default value of enqueue_after_transaction_commit is now set to true only when ActiveRecord is defined.
- When ActiveRecord is not available, setting enqueue_after_transaction_commit = true doesn’t make sense, so an exception is raised in that case to surface the misconfiguration clearly.

I'd appreciate it if you could take another look.

zzak:
I'm confused that we would re-introduce this config (see: 582f91e), but somehow makes sense in the context of changing the default? :thinking:

jeremy:
> I'm confused that we would re-introduce this config (see: [582f91e](https://github.com/rails/rails/commit/582f91e9330071200b1b72bcfef46a6960ea37e3)), but somehow makes sense in the context of changing the default? 🤔

582f91e deprecated setting it to symbol values in job adapters; it continues as an explicit boolean app-controlled setting. (The commit message does imply that the entire setting is deprecated when it was not.)

zzak:
Unless I'm missing something, the flag itself was deprecated and removed in 8.1, and we'd be re-introducing it in 8.2 / 9.0:
https://github.com/rails/rails/blob/582f91e9330071200b1b72bcfef46a6960ea37e3/activejob/lib/active_job/railtie.rb#L33-L38

jeremy:
> Unless I'm missing something, the flag itself was deprecated and removed in 8.1, and we'd be re-introducing it in 8.2 / 9.0:
> 
> https://github.com/rails/rails/blob/582f91e9330071200b1b72bcfef46a6960ea37e3/activejob/lib/active_job/railtie.rb#L33-L38

Ah I see what you mean: global config vs per-job config. This is a bit of a mess; I'll consider.

jeremy:
@zzak, revised to avoid re‑introducing global config—for now. I think it's an awkward mistake, though. You can see how much worse the code is. Not going to foist that upon anyone. Apps need an app-wide toggle for this. I aim to un-deprecate, particularly considering it wasn't removed on schedule.

zzak:
Yeah, if you want to change the default, old apps need a way to maintain the existing behavior without any changes so a global config is kind of the only option. :thinking: I was just pointing out that bit of context.

railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
```diff
@@ -9,6 +9,19 @@
...
+###
+# Defer job enqueueing until after the database transaction commits.
+#
+# When enabled, jobs enqueued inside a transaction wait until the transaction
+# commits before being dispatched to the queue. This prevents jobs from
+# operating on rolled-back data.
+#++
+# ActiveSupport.on_load(:active_job) do
+#   ActiveJob::Base.enqueue_after_transaction_commit = true
+# end
```
zzak: I think it's a bit of an anti-pattern to put initializer code here, typically we only put `Rails.application.config` afaik

Maybe it's ok to keep the config for this since the previous deprecation cycle is concluded, and we give people at least a MINOR version to adopt the change.

jeremy: Right, this is the ugly code foisted upon apps referenced above. I'm changing it back to global config and un-deprecating.

jeremy (approved):
Un-deprecated the global config toggle, added docs, switched to versioned config (`load_defaults "8.2"` with the new framework defaults initializer), and gated config on AR load (rather than presence of the constant).

Issue #48429, opened 2023-06-09
Pattern matching for ActionController::Parameters


Timmitry:
### Steps to reproduce

```ruby
ActionController::Parameters.new(a: 1, b: 2) => {a:}
```

### Suggested behavior

The statement will match and assign `a = 1`.

### Actual behavior

```
NoMatchingPatternError: #<ActionController::Parameters {"a"=>1, "b"=>2} permitted: false>: #<ActionController::Parameters {"a"=>1, "b"=>2} permitted: false> does not respond to #deconstruct_keys
```

### Workaround

```ruby
ActionController::Parameters.new(a: 1, b: 2).to_unsafe_hash => {a:}
```

### System configuration
**Rails version**:
7.0.4.3

**Ruby version**:
3.1.3

### Comments

I wonder if this is something that might be worth implementing. It would be as easy as to add the following code:

```ruby
# lib/action_controller/metal/strong_parameters.rb
def deconstruct_keys(_keys)
  @parameters
end
```

Since we are explicitly matching variables here, I do not see the need for something like `permit` - it's the same as using `a = params.fetch(:a)`, which is perfectly valid for me.

Some thought should be taken to the usage of

```ruby
ActionController::Parameters.new(a: 1, b: 2) => {a:, **kwargs}
```

which will lead to a variable `kwargs = { b: 2 }`, which again would be problematic if used for mass updating.

skipkayhil:
Hey @Timmitry, thanks for the suggestion! The best place to discuss this would be on the [forum](https://discuss.rubyonrails.org/) or [discord](https://discord.gg/d8N68BCw49). Since we reserve the issue tracker for bugs only, I'm going to close this

(I will link you to a related discussion about pattern matching for Active Model: https://github.com/rails/rails/pull/45553#issuecomment-1179568855)

seanpdoyle:
This has been implemented by https://github.com/rails/rails/pull/55789.

PR #55789, opened 2025-09-27
`ActionController::Parameters#deconstruct_keys` for pattern matching

seanpdoyle:
Related to [#48429][]

### Motivation / Background

Define `ActionController::Parameters#deconstruct_keys` in a `Hash`-like way, so that `params` can integrate with [pattern matching][] (for statements like `case … in`, `if … in`, and `params => { … }`).

### Detail

To do so, declare the [#deconstruct_keys][] method in way that loops over the specified keys and returns a
`ActiveSupport::HashWithIndifferentAccess` instance with only the root-level keys. Any primitives will remain primitives, and any `ActionController::Parameters` values below the first level of nesting will remain `ActionController::Parameters` instances.

### Additional information

This integration can be useful for controller parameters (either at the root level or nested within) that are to be passed to objects that don't expect attributes or keyword arguments:

```ruby
if params in { search:, page: }
  Model.search(search).limit(page)
else
  …
end

case (value = params[:primitive_or_nested])
in String
  # do something with a String `value`…
in { nested_key: }
  # do something with `nested_key` or `value`
else
  # …
end
```

[#48429]: https://github.com/rails/rails/issues/48429
[pattern matching]: https://docs.ruby-lang.org/en/master/syntax/pattern_matching_rdoc.html
[#deconstruct_keys]: https://docs.ruby-lang.org/en/master/syntax/pattern_matching_rdoc.html#label-Matching+non-primitive+objects-3A+deconstruct+and+deconstruct_keys

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,24 @@
+*   Define `ActionController::Parameters#deconstruct_keys` to support pattern matching
+
+    ```ruby
+    if params in { search:, page: }
+      Article.search(search).limit(page)
+    else
+      …
+    end
+
+    case (value = params[:string_or_hash_with_nested_key])
+    in String
+      # do something with a String `value`…
+    in { nested_key: }
+      # do something with `nested_key` or `value`
+    else
+      # …
+    end
+    ```
+
+    *Sean Doyle*
+
 *   Submit test requests using `as: :html` with `Content-Type: x-www-form-urlencoded`
 
     *Sean Doyle*
--- a/actionpack/lib/action_controller/metal/strong_parameters.rb
+++ b/actionpack/lib/action_controller/metal/strong_parameters.rb
@@ -316,6 +316,10 @@ module ActionController
       [self.class, @parameters, @permitted].hash
     end
 
+    def deconstruct_keys(keys)
+      slice(*keys).each.with_object({}) { |(key, value), hash| hash.merge!(key.to_sym => value) }
+    end
+
     # Returns a safe ActiveSupport::HashWithIndifferentAccess representation of the
     # parameters with all unpermitted keys removed.
     #
--- a/actionpack/test/controller/parameters/equality_test.rb
+++ b/actionpack/test/controller/parameters/equality_test.rb
@@ -57,4 +57,17 @@ class ParametersAccessorsTest < ActiveSupport::TestCase
     params = ActionController::Parameters.new(foo: { bar: "baz" })
     assert params.has_value?(ActionController::Parameters.new("bar" => "baz"))
   end
+
+  test "deconstruct_keys works with parameters" do
+    assert_pattern { @params => { person: { age: "32" } } }
+    refute_pattern { @params => { person: { addresses: ["does not match"] } } }
+  end
+
+  test "deconstruct_keys returns instances of ActionController::Parameters for nested values" do
+    @params => { person: }
+    person => { addresses: }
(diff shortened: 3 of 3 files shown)


actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -316,6 +316,12 @@ def hash
       [self.class, @parameters, @permitted].hash
     end
 
+    def deconstruct_keys(keys)
+      slice(*keys).each.with_object(ActiveSupport::HashWithIndifferentAccess.new) do |(key, value), hash|
+        hash.merge!(key => value)
+      end
```
MatheusRich: Would this work?
```suggestion
      ActiveSupport::HashWithIndifferentAccess.new(slice(*keys))
```

actionpack/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Define `ActionController::Parameters#deconstruct_keys` to support pattern matching
```
MatheusRich, line 1: I think it would be helpful to have one of the PR examples here too

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -316,6 +316,10 @@ def hash
       [self.class, @parameters, @permitted].hash
     end
 
+    def deconstruct_keys(keys)
+      slice(*keys).each.with_object({}) { |(key, value), hash| hash.merge!(key.to_sym => value) }
```
MatheusRich, line 320:
```suggestion
      slice(*keys).symbolize_keys
```

seanpdoyle: `ActionController::Parameters` does not define `#symbolize_keys` (it does not inherit from `Hash`).

Similarly, passing `slice(*keys)` to `ActiveSupport::HashWithIndifferentAccess` is a "deep" operation, which means that the values that are returned deeper than the first level of nesting are transformed from `ActionController::Parameter` instances to `ActiveSupport::HashWithIndifferentAccess` instances.

The current implementation proposed in this PR achieves both goals:

* it symbolizes the top-level keys so that the returned `Hash` instance can be destructured
* it does not coerce any values, and retains nested "hash"-like `ActionController::Parameter` instances

`ActionController::Parameter` instances are more `ActiveSupport::HashWithIndifferentAccess` than they are `Hash` (since you can access values with either symbol or string keys), so I'm hesitant to propose adding a `symbolize_keys` or `symbolize_keys!` method, since it wouldn't make much sense outside of this method.

guilleiguaran:
Please resolve conflicts in your branch

actionpack/lib/action_controller/metal/strong_parameters.rb
rainerborene, line 320: @seanpdoyle It's possible to replace `each.with_object` with `each_with_object` instead, right?

seanpdoyle: In my experience, no:

```
NoMethodError: undefined method `each_with_object' for an instance of ActionController::Parameters
```

Since ActionController::Parameters inherits from `Object` (rather than an `Enumerable` like `Hash`), it doesn't implement all the methods derived from `#each` (including `#each_with_object`). The enumerator returned from a call to `#each` without a block *does* respond to `#with_object`, so we used that here instead.

PR #55790, opened 2025-09-27
[ci skip] Docs: Improve clarity of Rails console examples in Getting Started guide

briankb:
### Motivation / Background

When reading through the *Getting Started* guide, console examples sometimes
show the input command and the returned output directly together in the same
block. For beginners, this can be confusing because it isn’t clear where the
command ends and the output begins. Separating them makes it much easier to
understand what should be typed vs. what Rails prints back.

This PR improves the clarity of Rails console examples for new learners.

### Detail

- Updated `guides/source/getting_started.md`
- Console input (`store(dev)> ...`) is now shown in its own fenced `irb` block
- The corresponding output (`=> ...`) is separated into a new fenced block
- Added a blank line for visual separation between input and output

No other guide content has been changed.

### Checklist

* [x] This Pull Request is related to one change (console formatting in docs).

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -414,9 +414,6 @@ We can instantiate a new Product record with the following code:
 
 ```irb
 store(dev)> product = Product.new(name: "T-Shirt")
-```
-And you should see:
-```irb
 => #<Product:0x000000012e616c30 id: nil, name: "T-Shirt", created_at: nil, updated_at: nil>
 ```
 
@@ -427,9 +424,6 @@ We can call `save` to write the record to the database.
 
 ```irb
 store(dev)> product.save
-```
-And you should see:
-```irb
   TRANSACTION (0.1ms)  BEGIN immediate TRANSACTION /*application='Store'*/
   Product Create (0.9ms)  INSERT INTO "products" ("name", "created_at", "updated_at") VALUES ('T-Shirt', '2024-11-09 16:35:01.117836', '2024-11-09 16:35:01.117836') RETURNING "id" /*application='Store'*/
   TRANSACTION (0.9ms)  COMMIT TRANSACTION /*application='Store'*/
@@ -445,9 +439,6 @@ the `product` variable.
 
 ```irb
 store(dev)> product
-```
-And you should see:
-```irb
 => #<Product:0x00000001221f6260 id: 1, name: "T-Shirt", created_at: "2024-11-09 16:35:01.117836000 +0000", updated_at: "2024-11-09 16:35:01.117836000 +0000">
 ```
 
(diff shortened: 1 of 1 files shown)


matthewd:
This seems more confusing to me. The suggested structure implies it's explicitly separating input and output, but it's actually not, because the "input" contains the prompt. (Plus the visual separation really detaches the input/output pairs :confused:)

briankb:
Thanks for the feedback! I understand how this might look more natural if you’re
already familiar with how Rails console works. But coming into Rails fresh, that
wasn’t my perspective.

In my case, the combined examples made it unclear that `=>` lines were the
console’s output — I actually tried to type them in as input at first. For
example, in section 7.1 “Creating Records,” I didn’t know that `=>` was the
response to the `product = Product.new(name: "T-Shirt")` command.

CleanShot Getting Started with Rails — Ruby on Rails Guides2025-09-27 at 14 44 40

Part of the confusion was that in other languages I’ve used (`=>` is valid
syntax — e.g. JavaScript arrow functions), so it felt natural to assume it was part of what I needed to type.

The change here is aimed at helping first‑time learners see more clearly that
there are two separate things: what you type vs. what Rails returns. That
distinction was confusing to me as a new Rails user, and separating them made
the intent much clearer.

briankb:
I’m happy to keep iterating on this to refine the presentation if that would help
make it more clear and stay consistent with the Guides’ style. My main goal is
just to reduce the initial confusion I ran into as a new Rails user so the
examples read more naturally to beginners.

p8:
Separating the input and output is done in other places in the guide.
For example when starting the server:
https://github.com/rails/rails/blob/main/guides/source/getting_started.md?plain=1#L181-L204
Or when generating a model
https://github.com/rails/rails/blob/main/guides/source/getting_started.md?plain=1#L248-L265

Maybe it makes sense to do it only the first time for the console as well?
cc @excid3

excid3:
IIRC the previous version of the guide had split input & output at the beginning to explain the difference.

If the goal is making it clear what is prompt, input and output, then I think it needs more separation (like syntax highlighting changes) as opposed to just splitting the code blocks. However, the goal is for it to visually match what you'd actually see in the terminal.

A simpler improvement might just be adding a small section to explain the output in the Rails Console section? It explains the prompt but it doesn't explain the output which would address this without needing to change every single code block.

briankb:
I thought about just adding a declaration as was suggested. once you use rails console a few times the => isn't nearly as confusing but the Getting Started is aimed for new learners, like me for rails. But I'm familiar with other languages and frameworks so I assumed the => was part of the command. I will work on an explainer block and post it back here in a few days to review. thanks for the feedback!

excid3:
Thanks @briankb! 

I can see the prompt being a bit confusing for the same reasons for beginners. Maybe an image with some labels pointing to each part could help visually call out each piece?

briankb:
I reverted the file back to its original version and then made this new change based on the feedback from the earlier comments. The section now has a short explanation in the Rails Console part to show the difference between what you type, the logs Rails prints, and the result that comes back. Hopefully this makes it clearer for beginners before they get to the “It works!” example.

skipkayhil:
I think the actual issue here is Rouge isn't highlighting IRB output lines correctly (as comments) because [it only looks for consoles named "irb" or "pry"](https://github.com/rouge-ruby/rouge/blob/b30020bb8ac965ef2a29560e115ebf0fc3da32d1/lib/rouge/lexers/irb.rb#L29). If someone wants to try sending a PR to Rouge to allow other console names (since IRB allows configuring them), that should fix the output lines not being highlighted like comments.

Edit: maybe I'm wrong actually, the IRB output highlighting seems generally inconsistent :/

PR #55792, opened 2025-09-28
Make transaction isolation work with transactional tests: follow-up

kirs:
@byroot this includes the part that was lost at the merge of https://github.com/rails/rails/pull/55549.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
@@ -171,7 +171,7 @@ module ActiveRecord
         @isolation_level = isolation
       end
 
-      def initialize(connection, isolation: nil, joinable: true, run_commit_callbacks: false)
+      def initialize(connection, isolation: nil, fixtures: false, run_commit_callbacks: false)
         super()
         @connection = connection
         @state = TransactionState.new
@@ -179,11 +179,11 @@ module ActiveRecord
         @records = nil
         @isolation_level = isolation
         @materialized = false
-        @joinable = joinable
+        @fixture_transaction = fixtures
         @run_commit_callbacks = run_commit_callbacks
         @lazy_enrollment_records = nil
         @dirty = false
-        @user_transaction = joinable ? ActiveRecord::Transaction.new(self) : ActiveRecord::Transaction::NULL_TRANSACTION
+        @user_transaction = fixtures ? ActiveRecord::Transaction::NULL_TRANSACTION : ActiveRecord::Transaction.new(self)
         @instrumenter = TransactionInstrumenter.new(connection: connection, transaction: @user_transaction)
       end
 
@@ -346,10 +346,11 @@ module ActiveRecord
       end
 
       def full_rollback?; true; end
-      def joinable?; @joinable; end
 
-      def fixture_transaction?
-        !joinable?
+      def fixture_transaction?; @fixture_transaction; end
+
+      def joinable?
+        !@fixture_transaction
       end
 
       protected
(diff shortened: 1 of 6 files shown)


activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -674,6 +687,10 @@ def within_new_transaction(isolation: nil, joinable: true)
             transaction&.incomplete!
           end
         end
+      ensure
+        if open_transactions == 1 && current_transaction.fixture_transaction?
+          current_transaction.isolation = old_isolation
+        end
```
kirs: It seems important that we reset this back.

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -641,6 +648,12 @@ def rollback_transaction(transaction = nil)
       end
 
       def within_new_transaction(isolation: nil, joinable: true)
+        old_isolation = current_transaction.isolation
+        if isolation && open_transactions == 1 && current_transaction.fixture_transaction?
```
byroot:
```suggestion
        if isolation && current_transaction.fixture_transaction?
```

Supposedly we no longer need to check the depth.

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -338,6 +348,10 @@ def commit_records
       def full_rollback?; true; end
       def joinable?; @joinable; end
 
+      def fixture_transaction?
+        !joinable?
+      end
```
byroot, line 354: I'm feeling like it should be the inverse.

```ruby
def joinable?
  !@fixture_transaction
end
```

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -641,6 +648,12 @@ def rollback_transaction(transaction = nil)
       end
 
       def within_new_transaction(isolation: nil, joinable: true)
```
byroot:
```suggestion
      def within_new_transaction(isolation: nil, fixtures: false)
```

I think we should rename the keyword arg too.

byroot (commented):
cc @matthewd are my comments in line with what you had in mind too?

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
matthewd: Despite being private API it's been around for a long time, so maybe a deprecation would be kind, but ultimately I think "joinable" can just go away.

(Read: "yes, and.." :smile:)

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
matthewd: I don't like repeating original conditionals in `ensure`... how about at assignment time we store `current_transaction` in a local, and then use that to indicate we should restore?

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -125,6 +125,7 @@ def after_commit; yield; end
       def after_rollback; end
       def user_transaction; ActiveRecord::Transaction::NULL_TRANSACTION; end
       def isolation=(_); end
+      def fixture_transaction?; false; end
```
kirs, line 128: I'm unsure about this bit. If we say that `fixture_transaction? == !joinable` then this should be `true`.

kirs:
@byroot @matthewd thanks for feedback! I've replaced `joinable?` with `fixture_transaction?` in private APIs and kept `joinable` as an argument in public APIs.

activerecord/lib/active_record/railties/console_sandbox.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 ActiveRecord::ConnectionAdapters::AbstractAdapter.set_callback(:checkout, :after) do
-  begin_transaction(joinable: false)
+  begin_transaction(fixtures: true)
```
kirs, line 4: @matthewd I had to rename it here, which perhaps is at least one good argument to keep it called as `joinable` instead of `fixtures`.

We could do that and keep `joinable` and then just provide `def fixture_transaction?` as an accessor.

PR #55794, opened 2025-09-28
[Fix #55513] parallel tests hanging when worker processes die abruptly

joshuay03:
### Motivation / Background

Closes https://github.com/rails/rails/issues/55513.

### Detail

When parallel tests are running and a worker process dies abruptly (e.g., OOM killed, `kill -9`), the test suite would hang forever waiting for the dead worker to call `stop_worker`, which it never could.

The fix detects dead workers during shutdown by using `Process.waitpid` with `WNOHANG` to check which processes have exited, then removes them from the server's active worker list before waiting for remaining workers to finish. The server now tracks PIDs alongside worker IDs, allowing it to map dead processes back to their worker entries for cleanup.

### Checklist

### DIFF
--- a/activesupport/test/parallelization_test.rb
+++ b/activesupport/test/parallelization_test.rb
@@ -33,4 +33,21 @@ class ParallelizationTest < ActiveSupport::TestCase
     instance = subclass.new("test")
     assert_equal 5, instance.parallel_worker_id
   end
+
+  test "shutdown handles dead workers gracefully" do
+    parallelization = ActiveSupport::Testing::Parallelization.new(1)
+    parallelization.start
+
+    sleep 0.25
+
+    server = parallelization.instance_variable_get(:@queue_server)
+    assert server.active_workers?
+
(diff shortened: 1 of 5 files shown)


activesupport/test/parallelization_test.rb
```diff
@@ -33,4 +33,21 @@ def teardown
...
+    sleep 0.25
+
+    server = parallelization.instance_variable_get(:@queue_server)
+    assert server.active_workers?
+
+    worker_pids = parallelization.instance_variable_get(:@worker_pool)
+    Process.kill("KILL", worker_pids.first)
+    sleep 0.25
+
+    Timeout.timeout(2.5, Minitest::Assertion, "Expected shutdown to not hang") { parallelization.shutdown }
```
joshuay03, line 50: On main:

Image

byroot (approved):
LGTM, I'll merge this in a bit, but need to backport etc.

Issue #54808, opened 2025-03-24
Unsetting cache headers after calling `stale?` or `fresh_when` is a rough experience


Mr0grog:
### Steps to reproduce

I’ve been using `stale?` to manage some extremely expensive computation, [more-or-less as the docs describe](https://api.rubyonrails.org/classes/ActionController/ConditionalGet.html#method-i-stale-3F):

```rb
if stale?(freshness_options)
  really_expensive_call
end
```

…but recently discovered that, if `really_expensive_call` raises an exception, I wind up returning a 500 response with some long-lived cache headers, which creates huge problems.

I’d like to reset those headers when there’s an exception, but this can be pretty hard since you have to know which headers `fresh_when` (which `stale?` calls) sets and what those headers’ previous states were to reset them.

### Expected behavior

I think having a method that takes a block and resets the headers if the block raises would be ideal:

```rb
# Resets the relevant headers if `really_expensive_call` raises
when_stale(freshness_options) do
  really_expensive_call
end
```

…but making `expires_now` or `no_store` reset other headers that `fresh_when` sets might be another way to solve the problem.

In my own code, I’ve currently implemented a not-very-good version of `when_stale`, but it would be nice if it were built-in so it was more future-proof and integrated with the rest of the cache tooling:

```rb
class MyController
  def when_stale(object = nil, **options)
    _cache_control = response.cache_control.dup
    _last_modified = response.last_modified
    yield if stale?(object, **options)
  rescue StandardError => e
    # Unfortunately these are some complicated getters and setters, so you
    # can't just set all of them back to their previous values.
    response.cache_control.clear.merge!(_cache_control)
    response.delete_header('ETag')
    if _last_modified
      response.last_modified = _last_modified
    else
      response.delete_header(ActionDispatch::Http::Cache::Response::LAST_MODIFIED)
    end
    raise e
  end
end
```

### System configuration

**Rails version**: 8.0.1

**Ruby version**: 3.4.2

skipkayhil:
Normally, I would expect uncaught exceptions to go through `ShowExceptions`, which doesn't look like it has access to a controller response's header during exception handling. How do you end up in a situation where a 500 is returning response headers from the controller?

Mr0grog:
> I would expect uncaught exceptions to go through `ShowExceptions`, which doesn't look like it has access to a controller response's header during exception handling. How do you end up in a situation where a 500 is returning response headers from the controller?

[...]

PR #55797, opened 2025-09-28
[ci skip] Document safe usage of `rescue_from`

Mr0grog:
### Motivation / Background

When responding with a non-2xx HTTP response, `ActionController::Base#rescue_from` handlers should usually create a fresh response object in order to avoid inheriting cookies, caching info, or other headers that were set during normal request handling. In most cases, these headers won’t be relevant to an error and can cause problems if they are added to an error response (for example, you probably don’t want an error to get cached by a CDN or other proxy). The guides and API docs never mention that, though.

I ran into this issue when using `if stale?` ([the recommended way to do conditional GET](https://guides.rubyonrails.org/caching_with_rails.html#conditional-get-support)) on a controller. Not realizing I needed to use a fresh response or clear the headers led to intermittent caching problems I didn’t know about and left unaddressed for a long time!

### Detail

This adds a warning and updates examples in the Rails Guides and API docs about this issue. It suggests always creating a fresh response object by calling `set_response! self.class.make_response! request`, which AFAICT is the most correct approach. I’m happy to change this if there is a better way!

The behavior change I suggested for `stale?` in #54808 doesn’t seem to have much interest, so I’m attempting to document things so fewer people hit this problem. I suspect there are probably other controller methods with similar issues, so this documentation update is probably a good idea anyway.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/rescuable.rb
+++ b/activesupport/lib/active_support/rescuable.rb
@@ -31,20 +31,28 @@ module ActiveSupport
       # which <tt>exception.is_a?(klass)</tt> holds true is the one invoked, if
       # any.
       #
+      # When used on an <tt>ActionController::Base</tt> subclass, handlers
+      # should usually create a fresh response object by calling
+      # <tt>set_response! self.class.make_response! request</tt>. This avoids
+      # inheriting headers that were set during request handling.
+      #
       #   class ApplicationController < ActionController::Base
       #     rescue_from User::NotAuthorized, with: :deny_access
       #     rescue_from ActiveRecord::RecordInvalid, with: :show_record_errors
       #
       #     rescue_from "MyApp::BaseError" do |exception|
+      #       set_response! self.class.make_response! request
       #       redirect_to root_url, alert: exception.message
       #     end
       #
       #     private
       #       def deny_access
+      #         set_response! self.class.make_response! request
       #         head :forbidden
       #       end
       #
       #       def show_record_errors(exception)
+      #         set_response! self.class.make_response! request
       #         redirect_back_or_to root_url, alert: exception.record.errors.full_messages.to_sentence
       #       end
       #   end
--- a/guides/CHANGELOG.md
+++ b/guides/CHANGELOG.md
@@ -1,2 +1,6 @@
+*   Document safe usage of `rescue_from` by creating a fresh response object
+    before actually responding with an HTTP error.
+
+    *Rob Brackett*
 
 Please check [8-1-stable](https://github.com/rails/rails/blob/8-1-stable/guides/CHANGELOG.md) for previous changes.
--- a/guides/source/action_controller_advanced_topics.md
+++ b/guides/source/action_controller_advanced_topics.md
@@ -608,6 +608,7 @@ class ApplicationController < ActionController::Base
 
   private
     def record_not_found
+      set_response! self.class.make_response! request
       render plain: "Record Not Found", status: 404
     end
 end
@@ -652,6 +653,12 @@ end
 WARNING: Using `rescue_from` with `Exception` or `StandardError` would cause
 serious side-effects as it prevents Rails from handling exceptions properly. As
 such, it is not recommended to do so unless there is a strong reason.
+
+WARNING: If you use `rescue_from` to respond with a non-2xx response, you
+should *usually* create a fresh response object by calling
+`set_response! self.class.make_response! request`. This ensures the error
+response does not inherit cookies, caching information, or other headers that
+were set as part of your application’s normal request handling.
 
 NOTE: Certain exceptions are only rescuable from the `ApplicationController`
 class, as they are raised before the controller gets initialized, and the action


Mr0grog:
This is still a concern on Rails 8.1.1, so I’ve rebased it and cleaned up the merge conflict on the changelog.

Again, happy to make any changes to the recommended advice here; this is the most reasonable solution I could figure based on where discussion died off on #54808.

matthewd:
Discussion in the linked issue seems to have progressed through you describing a behaviour that was not present, being corrected, and then making a modified/reduced claim, more than once.

While (my also-untested understanding of) the current behaviour does have the potential to surprise, I do not think this is a reasonable change.

If this were the right thing to do, then `rescue_from` should do it by default (_and_ we should have a real API to do it). But I also don't think that's the case: outside of potentially-temporary 5xx responses, the `if stale?` block is going to continue raising the same exception, and therefore the response _should_ be cacheable. (Otherwise the issue is in the stale condition, and would equally apply to any other situation where the user's code has arranged for caching to persist longer than their response is actually stable.)

Mr0grog:
> Discussion in the linked issue seems to have progressed through you describing a behaviour that was not present, being corrected, and then making a modified/reduced claim, more than once.

Oh! I’m sorry, I didn’t realize this came across as me changing the issue.

From my perspective, I discovered that the headers from `stale?` where winding up in my error responses, and TBH, did not even consider initially that whatever bottom-level error handling Rails does in production would be meaningfully different from what `rescue_from` does. The discussion in the issue revealed that to me (lesson learned!), but it also didn’t seem like it materially affected the issue: I was using the recommended error handling methods (at least so far as I understand it) and winding up with lingering headers from `stale?` that were causing real problems in production and that the `when_stale` solution I described there *did* solve (IIRC the caching of the error was happening in AWS CloudFront, but it’s so long ago now I don’t remember for sure) or that the approach described in this PR could also solve.

I guess, if it seemed like I was redefining or changing things in a problematic way, it would have been helpful to hear that as feedback, or have the issue closed as “won’t fix,” or something, instead of just being left alone with no response for a few months.

As it was, I figured no response there meant that there was no interest in changing the Rails code to solve the issue (whether it was my proposed `when_stale` or something else), so I figured it would at least be useful and less controversial to document what felt to me like an easy-to-make error. I’m sorry if that was a bad approach; I wasn’t really sure what else to do with no feedback from Rails core folks.

> If this were the right thing to do, then` rescue_from` should do it by default (and we should have a real API to do it).

Great! I’d love to have that change, or help with that change, or whatever.

> outside of potentially-temporary 5xx responses, the `if stale?` block is going to continue raising the same exception, and therefore the response should be cacheable.

In my case, at least, the `if stale?` block only intermittently raises an exception and in a way that has nothing to do with the inputs, so it does *not* actually continue raising the same exception. It is guarding a call to a separate service that does a very expensive calculation, plus other network calls to load the data for the calculation. As with any network boundary, every once in a while, those calls error out for wide variety of reasons. There might be other ways someone might find themselves in this sort of situation; I don’t know.

With respect to `rescue_from`, I guess I was also imagining that this is more general than `stale?`. I don’t know if there are other methods on `ActionController::Base` that might cause similar situations, but someone could certainly just start setting some headers in an action method and then later hit an exception before actually sending the response, which would leave them in a similar, broken state.

I realize that’s expanding the scope well beyond my original issue about `stale?`, but it wasn’t a type of issue I was even aware of before @skipkayhil’s comments on #54808, and those comments implied to me that the problem was really with my usage of `rescue_from` (since they noted this isn’t an issue for the error middleware because it starts a new response from scratch and stopped responding once I noted that it must be unique to `rescue_from` because it does not start a new response). So, this seemed like the most reasonable, in-keeping-with-the-framework approach I could come up with in the absence of any further feedback.

Mr0grog:
> While (my also-untested understanding of) the current behaviour does have the potential to surprise

@matthewd FWIW, I didn’t include a repro originally because I didn’t think there was a bug or anything wrong (the headers sticking around seemed like they must have been “as designed” at this point in Rails’s maturity!) and I was just suggesting a new feature. But I probably should have after @skipkayhil’s original response. That would have saved one extra round of back-and-forth.

Anyway, here’s an example if you need something to test!

<details><summary>Test Case Script (collapsed to keep this thread readable)</summary>

```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

end
Rails.application.initialize!

  get "/rescue_basic", to: "persistent_headers_test#index"
  get "/rescue_with_fresh_response", to: "fresh_response_test#index"
end

class MyAppError < StandardError
end

class PersistentHeadersTestController < ActionController::Base

  rescue_from MyAppError, with: :render_error

  def index
    # For test purposes, generate a random etag and current modified time so the `if` body always runs.
    if stale?(etag: SecureRandom.hex, last_modified: Time.now, public: true)
      raise MyAppError, "Problem inside the stale? block."

      # Never runs, but in a realistic scenario, the above `raise` would be intermittent.
      render plain: "New computed data from inside the stale? block."
    end
  end

  protected

  def render_error(error)
    render plain: "Something broke! #{error.message}", status: 500
  end
end

class FreshResponseTestController < PersistentHeadersTestController
  protected

  def render_error(error)
    set_response! self.class.make_response! request
    super
  end
end

class RescueHeadersTest < ActiveSupport::TestCase
  include Rack::Test::Methods

  def test_headers_persist_on_rescue_from
    get "/rescue_basic"
    assert_equal last_response.status, 500
    assert last_response.headers["etag"]
    assert last_response.headers["cache-control"] == "public"
  end

  def test_fresh_response_clears_cache_headers
    get "/rescue_with_fresh_response"
    assert_equal last_response.status, 500
    assert_nil last_response.headers["etag"]
    assert_equal last_response.headers["cache-control"], "no-cache"
  end

  private

  def app
    Rails.application
  end
end
```

</details>

PR #55802, opened 2025-09-29
Fix Module#anonymous? to handle overridden name method

americodls:
### Motivation / Background

`Module#anonymous?` breaks when classes override the `name` method, causing exceptions during introspection. This affects frameworks and CI optimizations that rely on determining if a module is attached to a constant.

The current implementation assumes `name` is never overridden, but this happens in practice:

```ruby
class Parent
  def self.name
    raise NotImplementedError, "Children classes must implement #{__method__}"
  end
end

Parent.anonymous? # raises NotImplementedError instead of returning false
```

### Details

This Pull Request changes `Module#anonymous?` to use `bind_call` to access the original `Module#name` directly, bypassing user-defined overrides on class or module definitions. This ensures the method accurately reflects the true anonymous state based on constant assignment.

### Additional information

This issue could affect any code that uses `Module#anonymous?` for introspection, particularly in frameworks or libraries that need to determine if a class/module has been properly defined and assigned to a constant.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Fix `Module#anonymous?` to handle overridden `name` methods.
+
+    The `anonymous?` method now uses `bind_call` to bypass any user-defined name method overrides and access the actual module name from Ruby's internal state. This ensures correct behavior when classes or modules override their `name`.
+
+    *Americo Duarte*
+
 *   Fix parallel tests hanging when worker processes die abruptly.
 
     Previously, if a worker process was killed (e.g., OOM killed, `kill -9`) during parallel
--- a/activesupport/lib/active_support/core_ext/module/anonymous.rb
+++ b/activesupport/lib/active_support/core_ext/module/anonymous.rb
@@ -25,6 +25,7 @@ class Module
   #   m.name         # => "M"
   #   m.anonymous?   # => false
   def anonymous?
-    name.nil?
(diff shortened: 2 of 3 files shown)


matthewd:
We don't generally try to protect ourselves against people overriding Ruby built-ins in ways that break their existing contract. I don't think I see a reason to do things differently here.

americodls:
Hey @matthewd 👋 

Thank you for looking at my proposal.

I completely see where you're coming from — I actually had the same initial thought as well.

The reason I leaned this way is that since the core extension `anonymous?` was added to `Module`, it feels more natural for its implementation to rely directly on `Module` itself rather than on the user's definitions themselves.

I also considered that the current approach is that the extra indirection could be unintentionally overridden or interfered with by users, which might introduce confusion or subtle bugs (as has happened to me).

On top of that, I also considered that if Ruby ever decided to adopt `Module#anonymous?`, it would likely be closer to this approach than to the current implementation.

I hope that perspective makes sense!

byroot:
I agree with Matthew here. In addition, "anonymous" isn't so clear cut these days with the introduction of `set_temporary_name`: https://docs.ruby-lang.org/en/master/Module.html#method-i-set_temporary_name

If you got some code that does low level introspection, it's best to do it like Zeitwerk: https://github.com/fxn/zeitwerk/blob/a09f0cb48fe7d56578a2880037b5a5a4474c30aa/lib/zeitwerk/real_mod_name.rb#L17

PR #55804, opened 2025-09-29
Store destroy dependent record message on reflection name

### LINKED ISSUE #53448: ActiveRecord::Associations.handle_dependency downcases attribute name in error messages
### Steps to reproduce
```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'rails'
  gem 'sqlite3'
  gem('rails-i18n', '~> 7.0')
end

require 'active_record'
require 'minitest/autorun'
require 'logger'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')

  end

    t.integer :post_id
  end
end

  has_many :comments, dependent: :restrict_with_error
end

  belongs_to :post
end

  setup do
    I18n.backend.store_translations(
      :en,
      {
        activerecord: {
          attributes: {
            post: {
              comments: 'Some Comments'
            }
          }
        }
      }
    )
  end
  def test_association_stuff
    post = Post.create!
    post.comments << Comment.create!

    # Attempt to destroy post, which should fail due to restrict_with_error
    assert_not post.destroy
    # Verify that the appropriate error message is added to the post’s errors.
    # The error message should include the human-readable name 'Some Comments' as defined
    # in the custom translation.
    assert_equal ['Cannot delete record because dependent Some Comments exist'], post.errors[:base]
  end
end

```

### Expected behavior
Attribute name should remain the same as in translation

### Actual behavior
Attribute name has been downcased from `Some Comments` to `some comments`

### System configuration
**Rails version**: 6.1.7.9

**Ruby version**: 3.2.5

tvdeyen:
Currently we store the error on the `base` attribute which does not allow to translate individual error messages per assocation.

Furthermore the record name was downcased not allowing to have a capitalized record, which makes the error message hard to use in places where the associated record is not an english word.

Fixes #53448

### Checklist

### DIFF
--- a/activerecord/test/cases/associations/has_many_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_associations_test.rb
@@ -2001,14 +2001,29 @@ class HasManyAssociationsTest < ActiveRecord::TestCase
 
     assert_not_empty firm.errors
 
-    assert_equal "Cannot delete record because dependent companies exist", firm.errors[:base].first
+    assert_equal "Cannot delete record because dependent companies exist", firm.errors[:companies].first
     assert RestrictedWithErrorFirm.exists?(name: "restrict")
     assert firm.companies.exists?(name: "child")
   end
 
   def test_restrict_with_error_with_locale
     I18n.backend = I18n::Backend::Simple.new
-    I18n.backend.store_translations "en", activerecord: { attributes: { restricted_with_error_firm: { companies: "client companies" } } }
+    I18n.backend.store_translations "en",
+      activerecord: {
+        errors: {
+          models: {
+            restricted_with_error_firm: {
+              attributes: {
+                companies: {
+                  restrict_dependent_destroy: {
+                    has_many: "Cannot delete record because Companies still exist"
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
     firm = RestrictedWithErrorFirm.create!(name: "restrict")
     firm.companies.create(name: "child")
 
@@ -2018,7 +2033,7 @@ class HasManyAssociationsTest < ActiveRecord::TestCase
 
     assert_not_empty firm.errors
 
-    assert_equal "Cannot delete record because dependent client companies exist", firm.errors[:base].first
+    assert_equal "Cannot delete record because Companies still exist", firm.errors[:companies].first
     assert RestrictedWithErrorFirm.exists?(name: "restrict")
     assert firm.companies.exists?(name: "child")
   ensure
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Store destroy dependent record message on reflection name
+
+    Store the error from a `has_many :companies, dependent: :restrict_with_error` association
+    on the `companies` attribute and not `base` as before.
+
+    Please change your translations accordingly if you have customized the message.
+
+    *Thomas von Deyen*
+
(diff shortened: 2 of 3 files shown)


activerecord/test/cases/associations/has_many_associations_test.rb
```diff
@@ -2001,14 +2001,29 @@ def test_restrict_with_error
...
   def test_restrict_with_error_with_locale
     I18n.backend = I18n::Backend::Simple.new
-    I18n.backend.store_translations "en", activerecord: { attributes: { restricted_with_error_firm: { companies: "client companies" } } }
+    I18n.backend.store_translations "en",
+      activerecord: {
+        errors: {
+          models: {
+            restricted_with_error_firm: {
+              attributes: {
+                companies: {
```
Edouard-chin, line 2017: This doesn't solve the issue referenced in the PR.

The issue is about modifying the translation for a model name. To reproduce the issue, the translation should be

```
{
  activerecord: {
    attributes: {
      post: {
        comments: 'Some Comments'
      }
    }
  }
}
```

We want to ensure that the "Some Comments" part remains as is and doesn't become "some comments".

Edouard-chin (changes requested):
Thank you for the PR but I have one opened and will change it according to the review from the team

activerecord/test/cases/associations/has_many_associations_test.rb
tvdeyen: It does solve the issue, but in a different way. The underlying issue is that the error is added to `base` instead of the relation name (what would be much more intuitive) and allows to translate the error message based on the relation name. Much more useful imo, because it allows to use completely different error messages per association instead of an combined one that soley allows one sentence.

I debugged this with the [great `i18n-debug` gem.](https://github.com/fphilipe/i18n-debug)

I have two models `Alchemy::Page` and `Alchemy::Language`, where language `has_many :pages,  `dependent: :restrict_with_error`. If you want to delete an language with pages attached this is the lookup.

```
[i18n-debug] en.activerecord.errors.models.alchemy/language.attributes.base.restrict_dependent_destroy.has_many => "There are still %{record} attached to this language. Please remove them first."
```

With my change we get this

```
[i18n-debug] en.activerecord.errors.models.alchemy/language.attributes.pages.restrict_dependent_destroy.has_many => "There are still pages attached to this language. Please remove the pages first."
```

I think your approach is also very useful, but it solves a different problem. 

I tried to keep the current downcased default translation in order to not "break" existing apps, but I am happy to remove this in order to make it more aligned with your PR. I thought it had a better chance to be merge in a minor or patch level release that way ;)

XSpielinbox:
What is preventing this PR from moving forward?

I would appreciate it, if there would be an easy way to have a generic error message translation that has some place holder that gets filled with the name of the dependent object in correct capitalization. This could be either by just removing the downcase in the current code or by introducing a new key that is looked up for translations to be compatible with old apps (though I doubt that someone relies on the old behavior).

Independent of that I see the value of allowing custom error messages per association, but see this as a distinct improvement rather than a highly awaited bugfix, because if I understand the proposal correctly, I would not be able to have a generic template but would have to write a complete error message for each association?

Edouard-chin:
Going to remove my "requested changes" and will let the team review/merge it since I ended up closing the PR I had opened a few weeks ago, since I lost context and couldn't remember what was the issue.

PR #55805, opened 2025-09-30
Only use JSON::Coder with `json >= 2.15`.

byroot:
Ref: https://github.com/rails/rails/pull/55726#issuecomment-3349478548 (cc @ghiculescu)

Older versions are harder to support, so it's not worth it.


ghiculescu:
Thank you!

PR #55735, opened 2025-09-23
Remove incorrect ActiveSupport changelog entry announced in the Rails 7.1.3

pjurewicz:
### Motivation / Background
This Pull Request has been created because the ActiveSupport changelog contains an entry that refers to a change which was later reverted and is **not part of the `7-1-stable` branch**. The entry was mistakenly announced in the Rails **7.1.3** release notes, even though the change had already been reverted. Keeping it is misleading for developers upgrading to Rails 7.1, since they might expect the reverted feature/fix to be available.

### Detail
This Pull Request removes the incorrect changelog entry from `activesupport/CHANGELOG.md`.
No code is affected — this is purely a documentation cleanup.

### Additional information
The change was reverted in commit: https://github.com/rails/rails/commit/2ac84355ca8498c2b679cc85c1cfc1064e5a6537  
No tests are needed, as this PR only updates documentation.

### Checklist
Before submitting the PR make sure the following are checked:
* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why.
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature.

PR #55806, opened 2025-09-30
Correct ActiveSupport changelog for JSON.encode fix and revert

pjurewicz:
### Motivation / Background

In my [previous PR](https://github.com/rails/rails/pull/55735), I stated that this `JSON.encode` fix had been reverted before the 7.1.3 release, and I removed the corresponding ActiveSupport 7.1.3 changelog entry.
In fact, the fix was actually included in 7.1.3 and then reverted in 7.1.4, which was never documented in the changelog.

### Detail

This PR restores the removed ActiveSupport 7.1.3 changelog entry and adds a note about its reversion in ActiveSupport 7.1.4. It also adds code samples for clarity.

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -111,6 +111,13 @@
 
     *Eugene Kenny*
 
+*   Fix `ActiveSupport::JSON.encode` to prevent duplicate keys.
+
+    If the same key exist in both String and Symbol form it could
+    lead to the same key being emitted twice.
+
+    *Manish Sharma*
+
 *   Fix `ActiveSupport::Cache::Store#read_multi` when using a cache namespace
     and local cache strategy.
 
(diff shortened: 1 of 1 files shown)


chaadow:
this change is quite annoying, we've been used all our entire "rails" lives to not care whether we pass symbols or strings ( especially with HashWithIndifferentAccess )

and now we will have duplicate keys if let's say we start from the params object ( having string keys) then we merge a symbol key with the intention of **overriding/taking precedence**

rafaelfranca:
You should always care if it is a symbol or a string, specially when you have both inside the hash and want to serialize. Who takes precedence? `HashWithIndifferentAccess` don't allow you to store a symbol and a string key inside the hash.

params doesn't store things inside it as symbols keys, it always use strings internally, so you should not have the precedence problem.

rafaelfranca:
BTW, this change was just bringing back the behavior Rails had for years. Rails always encoded duplicated keys in json, except for a release were we introduced a bug where we stoped doing it.

PR #54341, opened 2025-01-24
Fix `Enumerable#sole` for infinite collections

fatkodima:
Fixes #54335.

PR #55808, opened 2025-09-30
Fix `Enumerable#sole` when element is a tuple

### LINKED ISSUE #55807: `Enumerable#sole` returns only the first element of tuples
### Steps to reproduce

As of Rails 8.0.3:
```ruby
Rails::VERSION::STRING
# => "8.0.3"

enum = Enumerator.new(1) do |yielder|
  yielder.yield(1, "one")
end

enum.first
# => [1, "one"]

enum.rewind

enum.sole
# => 1
```

### Expected behavior

`Enumerable#sole` should return the full tuple, and the same value as `Enumerable#first`. This was the case prior to Rails 8.0.3:
```ruby
Rails::VERSION::STRING
# => "8.0.2.1"

enum = Enumerator.new(1) do |yielder|
  yielder.yield(1, "one")
end

enum.first
# => [1, "one"]

enum.rewind

enum.sole
# => [1, "one"]
```

### Actual behavior

#54341 changed the behavior of `Enumerable#sole`, causing it to only return the first element of tuples instead of the full tuple.

### System configuration

**Rails version**: 8.0.3

**Ruby version**: 3.4.4

olivier-thatch:
### Motivation / Background

This Pull Request has been created because #54341 introduced a behavior change for `Enumerable#sole` when the sole element is a tuple.

Previously, `Enumerable#sole` would return the full tuple (same as `Enumerable#first`). #54341 changed the behavior so that `Enumerable#sole` only returns the first element of the tuple.

Fixes #55807.

### Detail

This Pull Request changes `Enumerable#sole` to return the full tuple, i.e. the same value as `Enumerable#first`.

### Additional information

-

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/enumerable.rb
+++ b/activesupport/lib/active_support/core_ext/enumerable.rb
@@ -212,12 +212,12 @@ module Enumerable
     result = nil
     found = false
 
-    each do |element|
+    each do |*element|
       if found
         raise SoleItemExpectedError, "multiple items found"
       end
 
-      result = element
+      result = element.size == 1 ? element[0] : element
       found = true
     end
 
--- a/activesupport/test/core_ext/enumerable_test.rb
+++ b/activesupport/test/core_ext/enumerable_test.rb
@@ -401,6 +401,26 @@ class EnumerableTests < ActiveSupport::TestCase
     assert_raise(expected_raise) { GenericEnumerable.new(1..).sole }
   end
 
+  def test_sole_returns_same_value_as_first_for_tuples
+    enum = Enumerator.new(1) { |yielder| yielder.yield(1, "one") }
+    assert_equal [1, "one"], enum.sole
+    assert_equal enum.first, enum.sole
+  end
+
+  class KeywordYielder
+    include Enumerable
+
+    def each
+      yield 1, two: 3
+    end
+  end
+
+  def test_sole_keyword_arguments
+    yielder = KeywordYielder.new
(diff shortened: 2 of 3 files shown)


activesupport/lib/active_support/core_ext/enumerable.rb
```diff
@@ -212,12 +212,12 @@ def sole
     result = nil
     found = false
 
-    each do |element|
+    each do |*element|
```
byroot, line 215: I fear this may have a similar issue with keyword arguments, but I need to try a few things.

matthewd:
I imagine this would now break `[[1]].sole` too 😕

byroot:
> I imagine this would now break `[[1]].sole` too 😕

Nope.

activesupport/test/core_ext/enumerable_test.rb
```diff
@@ -401,6 +401,26 @@ def test_sole
...
+    def each
+      yield 1, two: 3
+    end
+  end
+
+  def test_sole_keyword_arguments
+    yielder = KeywordYielder.new
+    assert_equal [1, {two: 3}], yielder.sole
+    assert_equal yielder.first, yielder.sole
+  end
```
byroot, line 422: I pushed an extra test to make sure `*arguments` do also collect keyword args.

byroot:
Thanks for the fix, backported to 8-0-stable, and 7-2-stable.

olivier-thatch:
Thanks for the quick merge!

PR #55811, opened 2025-10-01
Array#duplicates?

dpep:
### Motivation / Background

This Pull Request has been created because I recently had to determine whether an array contained duplicates and was surprised Rails didn't have a helper to make this trivial.  Given all the threads on stack overflow regarding this subject, it seems like a common need.

### Detail

This Pull Request changes Array to add a `.duplicates?` method, which determines if the array contains duplicate elements.

### Additional information

common, although less efficient alternatives that iterate through the entire array instead of short circuiting:
```ruby
arr = [1, 2, 3, 2, 4]

arr.uniq.length != arr.length

arr.group_by(&:itself).select { |_k, v| v.size > 1 }.any?
```

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add `Array#duplicates?` method to easily determine whether arrays contain duplicate elements.
+
+    ```ruby
+    [ 1, 2, 2 ].duplicates? # => true
+    ```
+
+    *Daniel Pepper*
+
 *   Fix `Enumerable#sole` to return the full tuple instead of just the first element of the tuple.
 
     *Olivier Bellone*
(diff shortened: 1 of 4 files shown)


skipkayhil:
Its generally preferable to add new methods to Ruby instead of core extensions. I found some previous proposals for this type of method here: 

- https://bugs.ruby-lang.org/issues/16428
- https://bugs.ruby-lang.org/issues/15590

which looks like they are missing strong examples of how the method would be used. How would you plan to use this method?

dpep:
thanks for the pointers @skipkayhil !  Reading through them, it seems like there's two use cases: one to determine whether an array contains duplicates and the other to enumerate them.  Plus a naming question of `.uniq?` v `.duplicates?`.  My use case arose from needing to validate user input and ensure there were no duplicates entries.

skipkayhil:
> My use case arose from needing to validate user input and ensure there were no duplicates entries.

Right, this is the important bit. Looks like `uniq!` also may serve this purpose?

dpep:
yeah, something like:

```ruby
raise 'duplicates found' if input.uniq!
```

it feels a little awkward, but does reduce the argument for adding a new method.  thanks for your consideration!

PR #55813, opened 2025-10-01
Add environment config file existence check

### LINKED ISSUE #55779: Running `rails server` with missing or mis-spelled environments file silently hides the mistake
### Steps to reproduce

```
rails new lol --minimal --skip-active-record
cd lol
cat << EOF > config/environments/staging.rb
require_relative "production"

end
EOF
RAILS_ENV=stage rails server
```

### Expected behavior

If I have a file `config/environments/staging.rb` and I accidentally try to run `RAILS_ENV=stage rails server` (note it is `stage` but the file is `staging`). I expect I will get an error.

### Actual behavior

It boots with no problem and no clear warning:

```
$ RAILS_ENV=stage rails server
=> Booting Puma
=> Rails 8.0.3 application starting in stage
=> Run `bin/rails server --help` for more startup options
config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true

Puma starting in single mode...
* Puma version: 7.0.4 ("Romantic Warrior")
* Ruby version: ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
*  Min threads: 3
*  Max threads: 3
*  Environment: stage
*          PID: 86318
* Listening on http://0.0.0.0:3000
Use Ctrl-C to stop
```

## Suggestion

> WARNING: You are using `RAILS_ENV=stage`, but no `config/environments/stage.rb` was found. Known environments:
>
> - test.rb
> - production.rb
> - development.rb
>
> If this is intentional, you can ignore this warning by setting `RAILS_IGNORE_MISSING_ENV=stage`

Related: We could also have `Rails.env.<environment>?`, also check for the existence of a specific file for proof that it's a valid environment. Otherwise, `Rails.env.staaging?` has the same problem (silent typo failure, hard to find).

### System configuration

**Rails version**: 8.0.3

**Ruby version**: 3.4.6

daniel-niknam:
### Motivation / Background

Fixes #55779

### Detail

This Pull Request adds a behaviour to `Rails::Application` which will raise an error if unable to load any environment file.

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Add environment config file existence check
+
+    `Rails::Application` will raise an error if unable to load any environment file.
+
+    *Daniel Niknam*
+
 *   `Rails::Application::RoutesReloader` uses the configured `Rails.application.config.file_watcher`
 
     *Jan Grodowski*
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -636,6 +636,18 @@ module Rails
     end
 
     private
+      def missing_environment_file
+        raise "Rails environment has been set to #{Rails.env} but config/environments/#{Rails.env}.rb does not exist."
+      end
+
+      def any_environment_files?
+        paths["config/environments"]
+          .paths
+          .flat_map { |path| [path, *path.glob("*.rb")] }
+          .select(&:file?)
+          .any?
+      end
+
       def build_request(env)
         req = super
         env["ORIGINAL_FULLPATH"] = req.fullpath
--- a/railties/lib/rails/engine.rb
+++ b/railties/lib/rails/engine.rb
@@ -562,8 +562,14 @@ module Rails
     end
 
     initializer :load_environment_config, before: :load_environment_hook, group: :all do
-      paths["config/environments"].existent.each do |environment|
-        require environment
+      env_files = paths["config/environments"].existent
+
+      if env_files.empty? && any_environment_files?
+        missing_environment_file
+      else
+        env_files.each do |environment|
+          require environment
+        end
       end
     end
 
@@ -687,6 +693,12 @@ module Rails
       end
 
     private
+      def missing_environment_file; end
+
+      def any_environment_files?
+        false
+      end
+
       def load_config_initializer(initializer) # :doc:
         ActiveSupport::Notifications.instrument("load_config_initializer.railties", initializer: initializer) do
           load(initializer)
--- a/railties/test/application/configuration/custom_test.rb
+++ b/railties/test/application/configuration/custom_test.rb
@@ -9,7 +9,7 @@ module ApplicationTests
 
       def setup
         build_app
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
       end
 
       def teardown
--- a/railties/test/application/loading_test.rb
+++ b/railties/test/application/loading_test.rb
@@ -111,6 +111,15 @@ class LoadingTest < ActiveSupport::TestCase
     assert ::Rails.application.config.loaded
   end
 
+  test "raises RuntimeError when config/environments/ENV.rb file does not exist" do
+    msg = <<~MSG.squish
+      Rails environment has been set to non_existent_environment but
+      config/environments/non_existent_environment.rb does not exist.
+    MSG
+
+    assert_raise(RuntimeError, match: msg) { boot_app "non_existent_environment" }
+  end
+
   test "descendants loaded after framework initialization are cleaned on each request if reloading is enabled" do
     add_to_config <<-RUBY
       config.enable_reloading = true
--- a/railties/test/commands/credentials_test.rb
+++ b/railties/test/commands/credentials_test.rb
@@ -129,6 +129,7 @@ class Rails::Command::CredentialsTest < ActiveSupport::TestCase
   end
 
   test "edit command does not raise when an initializer tries to access non-existent credentials" do
+    File.write(app_path("config", "environments", "qa.rb"), "")
     app_file "config/initializers/raise_when_loaded.rb", <<-RUBY
       Rails.application.credentials.missing_key!
     RUBY
@@ -286,12 +287,14 @@ class Rails::Command::CredentialsTest < ActiveSupport::TestCase
   end
 
   test "diff for custom environment" do
+    File.write(app_path("config", "environments", "custom.rb"), "")
     run_edit_command(environment: "custom")
 
     assert_match(/access_key_id: 123/, run_diff_command("config/credentials/custom.yml.enc"))
   end
 
   test "diff for custom environment when key is not available" do
+    File.write(app_path("config", "environments", "custom.rb"), "")
     run_edit_command(environment: "custom")
     remove_file "config/credentials/custom.key"
 
--- a/railties/test/application/middleware/cookies_test.rb
+++ b/railties/test/application/middleware/cookies_test.rb
@@ -14,7 +14,7 @@ module ApplicationTests
 
     def setup
       build_app
-      FileUtils.rm_rf("#{app_path}/config/environments")
+      reset_environment_configs
     end
 
     def app
--- a/railties/test/application/paths_test.rb
+++ b/railties/test/application/paths_test.rb
@@ -8,7 +8,7 @@ module ApplicationTests
 
     def setup
       build_app
-      FileUtils.rm_rf("#{app_path}/config/environments")
+      reset_environment_configs
       app_file "config/environments/development.rb", ""
       add_to_config <<-RUBY
         config.root = "#{app_path}"
--- a/railties/test/application/rake/dbs_test.rb
+++ b/railties/test/application/rake/dbs_test.rb
@@ -10,7 +10,7 @@ module ApplicationTests
 
       def setup
         build_app
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
       end
 
       def teardown
--- a/railties/test/application/rake/framework_test.rb
+++ b/railties/test/application/rake/framework_test.rb
@@ -9,7 +9,7 @@ module ApplicationTests
 
       def setup
         build_app
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
       end
 
       def teardown
--- a/railties/test/application/rake/migrations_test.rb
+++ b/railties/test/application/rake/migrations_test.rb
@@ -7,7 +7,7 @@ module ApplicationTests
     class RakeMigrationsTest < ActiveSupport::TestCase
       def setup
         build_app
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
         add_to_config("config.active_record.timestamped_migrations = false")
       end
 
--- a/railties/test/application/rake/multi_dbs_test.rb
+++ b/railties/test/application/rake/multi_dbs_test.rb
@@ -9,7 +9,7 @@ module ApplicationTests
 
       def setup
         build_app(multi_db: true)
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
         add_to_config("config.active_record.timestamped_migrations = false")
       end
 
--- a/railties/test/isolation/abstract_unit.rb
+++ b/railties/test/isolation/abstract_unit.rb
@@ -144,6 +144,12 @@ module TestHelpers
       add_to_env_config :production, "config.log_level = :error"
     end
 
+    def reset_environment_configs
+      Dir["#{app_path}/config/environments/*.rb"].each do |path|
+        File.write(path, "")
+      end
+    end
+
     def teardown_app
       ENV["RAILS_ENV"] = @prev_rails_env if @prev_rails_env
       Rails.app_class = @prev_rails_app_class if @prev_rails_app_class
--- a/railties/test/railties/railtie_test.rb
+++ b/railties/test/railties/railtie_test.rb
@@ -8,7 +8,7 @@ module RailtiesTest
 
     def setup
       build_app
-      FileUtils.rm_rf("#{app_path}/config/environments")
+      reset_environment_configs
       require "rails/all"
     end
 


daniel-niknam:
@byroot, I followed your advice on creating a new initializer, but I went back and forth and couldn't decide whether or not `Rails::Engine` is the right place because I think this would warn the user every time a Rails::Engine is loaded. I wasn't able to figure out the exact behaviour of the `group:` attribute, but I'm guessing we should be able to control that to be executed only on the higher level, like `Rails::Application`. Also, I realised the tests are going to output a lot of warnings. Before going any further, I wanted to share this PR to get some feedback to make sure I'm on the right track. Please let me know what you think.

skipkayhil:
> because I think this would warn the user every time a Rails::Engine is loaded

Right, I don't think this is what we want. You should probably put the initializer in `Rails::Application` itself (?), which subclasses `Rails::Engine`, so that it only warns when the top level files are missing.

byroot:
Looks like a lot of `railties` tests have:

```ruby
FileUtils.rm_rf("#{app_path}/config/environments")
```

byroot:
I'm actually surprised `:load_environment_config` is executed for every engine. Granted I'm not the most familoar with initializers.

Is it expected that engines have their own `config/environments/*.rb` files?

byroot:
Ok, I pushed a commit to show how I think it would work. A bunch of tests would need cleanup though.

byroot:
So aside from the vast majority of Railties tests needing to be updated, another potential blocker is:

```ruby
    test "the application can be eager loaded even when there are no frameworks" do
      FileUtils.rm_rf("#{app_path}/app/jobs/application_job.rb")
      FileUtils.rm_rf("#{app_path}/app/models/application_record.rb")
      FileUtils.rm_rf("#{app_path}/app/mailers/application_mailer.rb")
      FileUtils.rm_rf("#{app_path}/config/environments")
      add_to_config <<-RUBY
        config.enable_reloading = false
        config.eager_load = true
      RUBY

      use_frameworks []

      assert_nothing_raised do
        app "development"
      end
    end
```

Rails is meant to be bootable from single file. So it's legit not to have the corresponding env file.

A solution might be to make the logic smarter, e.g.:

```ruby
if env_file_exists?
  require it
elsif any_env_file_exists?
  raise error
else
  # don't care.
end
```

daniel-niknam:
> I'm actually surprised `:load_environment_config` is executed for every engine. Granted I'm not the most familoar with initializers.
> 
> Is it expected that engines have their own `config/environments/*.rb` files?

@byroot Yes, we are on the same boat 😄 I suspect that would be the case because both `Rails::Application` and any engines will be using the same underlying class, which is the `Rails::Engine`.

I confirmed this behaviour by generating a Rails app first, generating a plugin inside the Rails app, creating an environment file in `config/environments/development.rb` inside the plugin with a content of `raise "inside engine"`, and after running Rails, I got the error `/private/tmp/blah/blorgh/config/environments/development.rb:1:in '<main>': inside engine (RuntimeError)`. 

So yes, the new engine is not generated with a set of environment files, but if you create it, it's going to load them. I concluded that means the block of initialiser is executed in any engine instance.

daniel-niknam:
> Rails is meant to be bootable from single file. So it's legit not to have the corresponding env file.
> 
> A solution might be to make the logic smarter, e.g.:
> 
> ```ruby
> if env_file_exists?
>   require it
> elsif any_env_file_exists?
>   raise error
> else
>   # don't care.
> end
> ```

That's very important to keep in mind, thanks. 

Let me see how I can get that to work!

daniel-niknam:
> > because I think this would warn the user every time a Rails::Engine is loaded
> 
> Right, I don't think this is what we want. You should probably put the initializer in `Rails::Application` itself (?), which subclasses `Rails::Engine`, so that it only warns when the top-level files are missing.

Thanks for the feedback @skipkayhil. I didn't implement it inside the `Rails::Application` class because I couldn't see any existing pattern that looks the same. But I think I need to find a way to make sure that logic is only executed on the main application, so I might end up introducing it. I want to play with that `group:` attribute to see if that does something useful for this use case 😄

railties/lib/rails/application.rb
```diff
@@ -636,6 +636,13 @@ def ensure_generator_templates_added
     end
 
     private
+      def missing_environment_file
+        raise <<~TEXT
+          Rails environment has been set to #{Rails.env} but
+          config\/environments\/#{Rails.env}.rb does not exist"
+        TEXT
+      end
```
daniel-niknam, line 641: Maybe passing available environments and making use of `DidYouMean::SpellChecker` to tell the user about a possible typo?

railties/lib/rails/engine.rb
```diff
@@ -562,8 +562,14 @@ def load_seed
     end
 
     initializer :load_environment_config, before: :load_environment_hook, group: :all do
-      paths["config/environments"].existent.each do |environment|
-        require environment
+      env_files = paths["config/environments"].existent
+
+      if env_files.empty? && paths["config/environments"].paths.first.glob("*.rb").any?
```
daniel-niknam: I saw some examples in the tests that we push a specific file to `paths["config/environments"]`. I'm wondering if I should loop over all `paths` instead of using the first one? Things could go wrong if the user `unshift`s something. It might get complicated to cover that scenario, but I haven't looked into it.

railties/test/application/configuration/custom_test.rb
```diff
@@ -9,7 +9,7 @@ class CustomTest < ActiveSupport::TestCase
 
       def setup
         build_app
-        FileUtils.rm_rf("#{app_path}/config/environments")
+        reset_environment_configs
```
daniel-niknam, line 12: @byroot Thanks for adding these method calls. It made it a lot easier for me to work my way through failed tests.

railties/test/commands/credentials_test.rb
```diff
@@ -129,6 +129,7 @@ class Rails::Command::CredentialsTest < ActiveSupport::TestCase
   end
 
   test "edit command does not raise when an initializer tries to access non-existent credentials" do
+    File.write(app_path("config", "environments", "qa.rb"), "")
```
daniel-niknam, line 132: In this test, we already have `development.rb`, `test.rb` and `production.rb`. I've added the extra env file only in the test. The other way would be to extend the `setup :build_app` somehow to accept extra env. Maybe out of scope of this PR.

railties/lib/rails/application.rb
```diff
@@ -636,6 +636,13 @@ def ensure_generator_templates_added
     end
 
     private
+      def missing_environment_file
+        raise <<~TEXT
```
daniel-niknam: Should users have a config to turn off this? I don't know, but I'm guessing maybe someone could have a case in which they are running Rails on some weird environment name and everything is ok for them. We could check for config and switch from `raise` to `warn`.

daniel-niknam (commented):
I managed to have another look at this PR finally 😅 We’ve had a bit of a cold going around the house, everyone caught it one by one, and I was the last one standing. Finally caught it myself, but all good now! 😄 

@byroot When you can, can you please have a look at this PR one more time? Before I do a cleanup, it would be good to know what you think. I added some comments about some possible changes, but I'm not sure because I'm new.

railties/test/application/loading_test.rb
```diff
@@ -111,6 +111,10 @@ class User < ActiveRecord::Base
     assert ::Rails.application.config.loaded
   end
 
+  test "Raises RuntimeError when config/environments/ENV.rb file does not exist" do
```
rafaelfranca:
```suggestion
  test "raises RuntimeError when config/environments/ENV.rb file does not exist" do
```

railties/test/application/loading_test.rb
```diff
@@ -111,6 +111,10 @@ class User < ActiveRecord::Base
...
+    assert_raise(RuntimeError, /Rails environment has been set to non_existent_environment but config\/environments\/non_existent_environment.rb does not exist./) { boot_app "non_existent_environment" }
```
rafaelfranca:
```suggestion
    assert_raises(RuntimeError, match: /Rails environment has been set to non_existent_environment but config\/environments\/non_existent_environment.rb does not exist./) { boot_app "non_existent_environment" }
```

railties/lib/rails/engine.rb
rafaelfranca: We should check all paths. And we should only check if it is application, so maybe extract this to a method that just return `false` in engines?

byroot:
> When you can, can you please have a look at this PR one more time?

I'll make some time soon. 

One thing to check (I will but you can do it first if you want) is to ensure your branch don't break "single file apps", like our repro scripts: https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case

railties/lib/rails/application.rb
byroot: Nah, it's extra complexity for nothing. They can create an empty file if needed.

byroot:
Generally looks good to me, once Rafael feedback is addressed and the git history cleaned up I think it's ready for merge. Will probably wait for 8.1 final to be released though.

railties/lib/rails/application.rb
daniel-niknam: @byroot what do you think of this comment? I can experiment with that and see how complicated it gets.

railties/lib/rails/application.rb
daniel-niknam: Make sense.

railties/test/application/loading_test.rb
daniel-niknam: Thank you. I'm new to minitest and thought it was working like RSpec 🤦
TIL: without `match:`, the test ignored the regex.

railties/lib/rails/application.rb
daniel-niknam: I leave it out for now, but I managed to get it working with something like:

```ruby
      def missing_environment_file
        corrections = DidYouMean::SpellChecker.new(dictionary: all_environment_files.map {|path| path.basename(".rb") }).correct(Rails.env)

        message = "Rails environment has been set to #{Rails.env} but config/environments/#{Rails.env}.rb does not exist."
        message += DidYouMean::Formatter.message_for(corrections) if corrections.any?

        raise message
      end
```

daniel-niknam: In `railtie`, `DidYouMean` has been used inside commands and test runner only.

daniel-niknam:
Thanks @rafaelfranca for the feedback. I've addressed your comments. I have an error on the build, but it seems to be a flaky one. I'm going to do some refactoring and squash everything to clean it up.

railties/lib/rails/application.rb
```diff
@@ -636,6 +636,18 @@ def ensure_generator_templates_added
...
 
     private
+      def missing_environment_file
+        raise "Rails environment has been set to #{Rails.env} but config/environments/#{Rails.env}.rb does not exist."
+      end
+
+      def any_environment_files?
+        paths["config/environments"]
+          .paths
+          .flat_map { |path| [path, *path.glob("*.rb")] }
```
daniel-niknam, line 646: paths["config/environments"] could contain both files and directories. Using `[path, *path.glob("*.rb")]` with a combination of `flat_map` we can make sure both cases are considered. 

I could rewrite this like:

```suggestion
          .flat_map { |path| path.directory? ? path.glob("*.rb") : path }
```

But I didn't like the ternary operator. Happy to use that or `if` statement if this doesn't read well.

railties/test/application/loading_test.rb
```diff
@@ -111,6 +111,10 @@ class User < ActiveRecord::Base
     assert ::Rails.application.config.loaded
   end
 
+  test "raises RuntimeError when config/environments/ENV.rb file does not exist" do
+    assert_raise(RuntimeError, match: /Rails environment has been set to non_existent_environment but config\/environments\/non_existent_environment.rb does not exist./) { boot_app "non_existent_environment" }
```
zzak:
```suggestion
    msg = <<~MSG.squish
      Rails environment has been set to non_existent_environment but
      config/environments/non_existent_environment.rb does not exist.
    MSG
    assert_raise(RuntimeError, match: msg) { boot_app "non_existent_environment" }
```

railties/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+*   Add environment validation check
```
zzak:
```suggestion
*   Add environment config file exists check
```

Maybe a little more specific works?

daniel-niknam: Thanks for the feedback @zzak. How about `Add environment config file existence check`? It reads a bit better to me, but I'm happy to go with your suggestion.

zzak: Yeah I think it's better, let's go!

railties/test/application/loading_test.rb
daniel-niknam: Thanks, this reads much better ❤️

zzak (approved):

daniel-niknam:
I just realised I messed up the git history again 🤦 😅 It's ready now. Thank you 🙇

daniel-niknam:
> > When you can, can you please have a look at this PR one more time?
> 
> I'll make some time soon.
> 
> One thing to check (I will but you can do it first if you want) is to ensure your branch don't break "single file apps", like our repro scripts: https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case

@byroot Sorry, I forgot to reply to your comment about those scripts. I've updated the gem path to use my own working directory in this PR, and I managed to execute a few of them without any issues. I've checked them all, none of them is setting any environment.rb file and I could run them without any issues. I don't see any particular reason for people to try to do that in those scripts. I think we are safe ✅

railties/lib/rails/engine.rb
```diff
@@ -687,6 +693,12 @@ def run_tasks_blocks(*) # :nodoc:
       end
 
     private
+      def missing_environment_file; end
+
+      def any_environment_files?
+        []
+      end
```
byroot, line 700:
```suggestion
      def any_environment_files?
        false
      end
```

daniel-niknam: Dang it. I must have missed it during rebase 😂

schneems:
Thanks all!

PR #55814, opened 2025-10-01
Exclude notes from excluded directories

siaw23:
### Motivation / Background

The /builds folder is where compiled frontend assets are written to at build time, When we run 'rails notes' notes from this folder clutter the terminal, but these are notes from 3rd parties that we didn't write

### Checklist

### DIFF
--- a/railties/lib/rails/source_annotation_extractor.rb
+++ b/railties/lib/rails/source_annotation_extractor.rb
@@ -79,6 +79,16 @@ module Rails
         directories.push(*dirs)
       end
 
+      def self.excluded_directories
+        @@excluded_directories ||= []
+      end
+
+      # Registers directories to be excluded from scanning
+      #   Rails::SourceAnnotationExtractor::Annotation.register_excluded_directories("app/assets/builds")
+      def self.register_excluded_directories(*dirs)
+        excluded_directories.push(*dirs)
+      end
+
       def self.tags
         @@tags ||= %w(OPTIMIZE FIXME TODO)
       end
@@ -167,6 +177,8 @@ module Rails
     # taken into account. Only files with annotations are included.
     def find_in(dir)
       results = {}
+
+      return results if Annotation.excluded_directories.include?(dir)
 
       Dir.glob("#{dir}/*") do |item|
         next if File.basename(item).start_with?(".")
--- a/railties/test/commands/notes_test.rb
+++ b/railties/test/commands/notes_test.rb
@@ -167,6 +167,13 @@ class Rails::Command::NotesTest < ActiveSupport::TestCase
     assert_empty run_notes_command
   end
 
+  test "does not display results from excluded directories" do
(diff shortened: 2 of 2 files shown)


railties/lib/rails/source_annotation_extractor.rb
```diff
@@ -170,6 +170,7 @@ def find_in(dir)
 
       Dir.glob("#{dir}/*") do |item|
         next if File.basename(item).start_with?(".")
+        next if item.match?(%r{(^|/)app/assets/builds(/|$)})
```
skipkayhil: Since `find_in` appears to be recursive, should we move this up out of the `glob` iteration so that it only has to check the current directory once?

siaw23: Right, moved this up to avoid unnecessary work.

railties/lib/rails/source_annotation_extractor.rb
```diff
@@ -168,6 +168,8 @@ def find(dirs)
     def find_in(dir)
       results = {}
 
+      return results if dir.match?(%r{(^|/)app/assets/builds(/|$)})
```
p8: I'm wondering if this should be made more generic, there might be other directories to exclude as well.

```suggestion
      return results if dir.in?(self.class.excluded_directories)
```

This would work similar to how directories are registered:
```ruby
    class Annotation < Struct.new(:line, :tag, :text)
      def self.excluded_directories
        @@excluded_directories ||= []
      end

      def self.register_excluded_directories(*dirs)
        excluded_directories.push(*dirs)
      end
    ...
```

p8: Also, this `app/assets/builds` is configured in js-bundling-rails:
https://github.com/rails/jsbundling-rails/blob/main/lib/install/install.rb#L2-L3
Rails itself doesn't know about this directory, especially with import maps being the default.
So with the above change js-bundling-rails could call:
```ruby
Annotation.register_excluded_directories("app/assets/builds")
```

siaw23: @p8 this looks more elegant. I've implemented the changes and will need to check how the wiring is set up from js-bundling-rails and create a PR there too. thanks Petrik!

railties/test/commands/notes_test.rb
```diff
@@ -167,6 +167,13 @@ class Rails::Command::NotesTest < ActiveSupport::TestCase
     assert_empty run_notes_command
   end
 
+  test "does not display results from generated assets under app/assets/builds" do
```
p8:
```suggestion
  test "does not display results from excluded directories" do
```

p8 (changes requested):

railties/test/commands/notes_test.rb
siaw23: fixed @p8

PR #49894, opened 2023-11-02
Don’t ignore X-Forwarded-For IPs with ports attached (again)

ybiquitous:
Fix: https://github.com/rails/rails/pull/55816

### Motivation / Background

This Pull Request has been created because I'd like to merge #39134 again, which was reverted due to Rack versions compatibility.

### Detail

This Pull Request doesn't ignore X-Forwarded-For IPs with ports attached, e.g. `3.4.5.6:1234` or `[fe80:0000:0000:0000:0202:b3ff:fe1e:8329]:3000`, in the `ActionDispatch::RemoteIp` middleware. For details, see the description in #39134 (thanks to @duncanjbrown).

I've cherry-picked the commit <https://github.com/rails/rails/pull/39134/commits/c2077c850715b359a24a6b083d2e042565587c52> and modified partially for the latest Rails code (added a test case for IPv6, too).

### Additional information

I believe there is no longer a compatibility issue because:

- the latest `actionpack` gem requires Rack 2.2.4 or later,
  https://github.com/rails/rails/blob/bbc6230ae7a708750a9bc14102bd4627ba85cd1a/actionpack/actionpack.gemspec#L40
- Rack 2.2.x series have supported the `Rack::Request#forwarded_for` API:
  https://github.com/rack/rack/blob/v2.2.1/lib/rack/request.rb#L330

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55816, opened 2025-10-01
Strip port numbers from `X-Forwarded-For` IP addresses

adam12:
### Motivation / Background

RFC 7239 section 5.2 specifies that forwarded IP addresses in proxy headers may include port information (e.g., `192.168.1.1:8080`). Some proxies and clients include this port information when forwarding requests, but the `RemoteIp` middleware currently rejects these IP addresses as invalid during IPAddr parsing.

This causes legitimate forwarded IP addresses to be ignored.

### Details

I've seen this in the wild for an application deployed to Azure Web Apps.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/remote_ip.rb
+++ b/actionpack/lib/action_dispatch/middleware/remote_ip.rb
@@ -177,6 +177,19 @@ module ActionDispatch
         return [] unless header
         # Split the comma-separated list into an array of strings.
         ips = header.strip.split(/[,\s]+/)
+        ips.map! do |ip|
+          # Remove port number if present, per RFC 7239 section 5.2.
+          case ip
+          when /\A\[(.+)\]:(\d+)\z/ # [IPv6]:port
+            $1
+          when /\A\[(.+)\]\z/       # [IPv6] without port
+            $1
+          when /\A([^:]+):(\d+)\z/  # IPv4:port (no other colons)
+            $1
+          else                      # Plain IP without port
+            ip
+          end
+        end
         ips.select! do |ip|
           # Only return IPs that are valid according to the IPAddr#new method.
           range = IPAddr.new(ip).to_range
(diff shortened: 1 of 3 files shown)


byroot:
This seem sensible, and I'd even consider this a bugfix worth of a backport, but I think we have to be careful here because there might be security considerations.

actionpack/lib/action_dispatch/middleware/remote_ip.rb
```diff
@@ -177,6 +177,19 @@ def ips_from(header) # :doc:
...
+          case ip
+          when /\A\[(.+)\]:(\d+)\z/ # [IPv6]:port
+            $1
+          when /\A\[(.+)\]\z/       # [IPv6] without port
+            $1
+          when /\A([^:]+):(\d+)\z/  # IPv4:port (no other colons)
+            $1
+          else                      # Plain IP without port
+            ip
+          end
```
byroot, line 191: I feel like it would be simpler to just look for `:(\d+)\z` and strip that?

adam12: That was my original implementation but it fails for IPv6.

adam12:
> there might be security considerations

I agree.

taketo1113:
It seems that #49894 addresses the same issue.

In that Pull Request, the retrieval of `X-Forwarded-For` IP addresses is changed to use the `forwarded_for` method from the rack gem instead of the `x_forwarded_for` method.
The `forwarded_for` method in rack returns only the IP address portion of the `X-Forwarded-For` header, excluding any port numbers.

I believe using the `forwarded_for` method to exclude port numbers results in a more secure implementation.

byroot:
Agreed, I rebased https://github.com/rails/rails/pull/49894 and will merge it on green CI. Thanks a lot for pointing it out.

PR #55822, opened 2025-10-02
Add prefix option to has_secure_token for improved token identification

benngarcia:
Adds an optional :prefix paramer to has_secure_token that prepends a string to generated tokens, making token types immediately identifiable in logs, debugging sessions, and error messages.

Before:
  user.auth_token # => "pX27zsMN2ViQKta1bGfLmVJE"
  user.reset_token # => "tU9bLuZseefXQ4yQxQo8wjtB"

After:
  has_secure_token :auth_token, prefix: "auth_"
  has_secure_token :reset_token, prefix: true

  user.auth_token # => "auth_pX27zsMN2ViQKta1bGfLmVJE"
  user.reset_token # => "reset_token_tU9bLuZseefXQ4yQxQo8wjtB"

This enables better developer clarity without additional database columns or complex token parsing logic. Particularly valuable when analyzing application logs, supporting operations teams, and debugging authentication flows with multiple token types.

Anecdotally, I have dozens of classes that I override the #generate_unique_secure_token method, passing in a hardcoded prefix. It _feels_ far nicer to copy/paste a token with context to others than an string without any purpose encoded.

Notably, ~~the current implementation makes this a breaking change for anyone who has overriden the #generates_unique_secure_token like myself~~. This is no longer true with https://github.com/rails/rails/pull/55822/commits/38cad158e34a914fbf4e45ae2c0f89d8f4e40116

### Checklist

### DIFF
--- a/activerecord/lib/active_record/secure_token.rb
+++ b/activerecord/lib/active_record/secure_token.rb
@@ -14,14 +14,17 @@ module ActiveRecord
       #   # Schema: User(token:string, auth_token:string)
       #   class User < ActiveRecord::Base
       #     has_secure_token
-      #     has_secure_token :auth_token, length: 36
+      #     has_secure_token :invite_token, prefix: true
+      #     has_secure_token :auth_token, length: 36, prefix: "auth_"
       #   end
       #
       #   user = User.new
       #   user.save
       #   user.token # => "pX27zsMN2ViQKta1bGfLmVJE"
-      #   user.auth_token # => "tU9bLuZseefXQ4yQxQo8wjtBvsAfPc78os6R"
+      #   user.invite_token # => "invite_token_srUP2WWCb6yCtfy6CAKVdzxF"
+      #   user.auth_token # => "auth_tU9bLuZseefXQ4yQxQo8wjtBvsAfPc78os6R"
       #   user.regenerate_token # => true
+      #   user.regenerate_invite_token # => true
       #   user.regenerate_auth_token # => true
       #
       # +SecureRandom::base58+ is used to generate at minimum a 24-character unique token, so collisions are highly unlikely.
@@ -33,8 +36,7 @@ module ActiveRecord
       # ==== Options
       #
       # [+:length+]
-      #   Length of the Secure Random, with a minimum of 24 characters. It will
-      #   default to 24.
+      #   Length of the randomly generated token, with a minimum of 24 characters. Defaults to 24.
       #
       # [+:on+]
       #   The callback when the value is generated. When called with <tt>on:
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/secure_token.rb
```diff
@@ -14,13 +14,13 @@ module ClassMethods
       #   # Schema: User(token:string, auth_token:string)
       #   class User < ActiveRecord::Base
       #     has_secure_token
-      #     has_secure_token :auth_token, length: 36
+      #     has_secure_token :auth_token, length: 36, prefix: "auth_"
```
MatheusRich, line 18: Should `length` take the prefix into consideration? If not (i believe is the case), should we clarify that in the docs?

activerecord/lib/active_record/secure_token.rb
```diff
@@ -43,23 +43,26 @@ module ClassMethods
       #   in a <tt>before_</tt> callback. When not specified, +:on+ will use the value of
       #   <tt>config.active_record.generate_secure_token_on</tt>, which defaults to +:initialize+
       #   starting in \Rails 7.1.
-      def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on)
+      #
+      # [+:prefix+]
+      #   An optional prefix to provide context about the token in different settings such as in logs. Defaults to nil.
```
MatheusRich: I'd expect tokens to be redacted in the logs. Maybe a different example here?

activerecord/lib/active_record/secure_token.rb
benngarcia: I don't believe it should. I would expect the length of the token to refer to the actual token.

Clarifying in the docs is a great idea though.

zzak:
> making token types immediately identifiable in logs, debugging sessions, and error messages.

Shouldn't these be filtered from logs and error messages?

I can see wanting a mask for `puts`ing them too.

benngarcia:
> > making token types immediately identifiable in logs, debugging sessions, and error messages.
> 
> Shouldn't these be filtered from logs and error messages?
> 
> I can see wanting a mask for `puts`ing them too.

It depends on the token's sensitivity. I mask sensitive tokens (like private API keys) using [`filter_attributes`](https://api.rubyonrails.org/classes/ActiveRecord/Core/ClassMethods.html). Public-ish tokens I don't.

I was referring to contexts where you *want* the token visible: debugging sessions, explicit display (`<%= @user.token %>`), client identification. Not automated request logs or AR object dumps.

That said, you make a good point: it'd be reasonable for `has_secure_token` to automatically append to `filter_attributes` by default, given the method name implies security - can spin up a PR but interested in more thoughts/feedback first.

zzak:
It'd be interesting if you could mask it like `token=reset_token_REDACTED` to include the prefix part outside of what gets filtered :thinking: 

I think opening a PR is the best way to get feedback, but not trying to prescribe you work.

benngarcia:
Thought that was a really great idea - appreciate the feedback. I do generally keep the default attributes_for_inspect, but for cases where I don't then this would be preferable.

I took a look this morning and found out that `filter_attributes` accepts a block, making the change fairly trivial. PR here:
https://github.com/rails/rails/pull/55972

PR #55823, opened 2025-10-03
Backwards compatibility in migrations with remove_foreign_key syntax

that-jill:
### Motivation / Background

When upgrading to Rails 8.1.0.beta1, older migrations were failing.

### Detail

There are several different ways to remove foreign keys in migrations:

```ruby
def up
  change_table(:sub_testings) do |t|
    t.remove_foreign_key :testings, column: "testing_id"
  end
end
```

also

```ruby
def up
  remove_foreign_key :sub_testings, column: "testing_id"
end
```

in the last syntax the to_table variable is not defined. The error we were seeing is that this was getting passed into rails as a nil and then transformed into an empty string, and then failing when the migration ran:

```
Caused by:
ArgumentError: Table '<table_name>' has no foreign key for  (ArgumentError)

            raise(ArgumentError, "Table '#{from_table}' has no foreign key for #{to_table || options}")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Additional information

#proper_table_name has been around for a long time. I can't think of any reason why we'd ever want to return an empty string instead of a nil if we don't pass in a name, but there may be a use case. None of the tests failed when I made this change. 

I added a test with a migration for `remove_foreign_key` versioned to Rails 7.0 because that's where I saw this migration originally fail. However, if I version that migration to 8.0, it fails as well.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/migration.rb
+++ b/activerecord/lib/active_record/migration.rb
@@ -1122,6 +1122,8 @@ module ActiveRecord
     # Uses the Active Record object's own table_name, or pre/suffix from the
     # options passed in.
     def proper_table_name(name, options = {})
+      return if name.nil?
+
       if name.respond_to? :table_name
         name.table_name
       else
--- a/activerecord/test/cases/migration/compatibility_test.rb
+++ b/activerecord/test/cases/migration/compatibility_test.rb
@@ -718,6 +718,26 @@ module ActiveRecord
         ActiveRecord::Base.clear_cache!
       end
 
+      def test_remove_foreign_key_on_7_0
+        connection.create_table(:sub_testings) do |t|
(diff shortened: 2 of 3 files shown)


that-jill:
I still have work that I want to do with this PR, but it feels important to raise this early, as it's a bug with 8.1. I need to figure out which changed caused the breakage, do more digging into the use cases with `#proper_table_name`, update the changelog, etc. This may or may not be the correct fix for the problem.

zzak:
Haven't looked into this deeply but reminds me of #55086, and I wonder if that is related. :pray:

Issue #25333, opened 2016-06-09
ActionCable: consider deferring `transmit`s until after `confirm_subscription`


machty:
### Steps to reproduce

Create an ActionCable channel that transmits data within `subscribed`, which seems like a reasonable place to send an initial payload upon subscription (e.g. if you're subscribing to a `TweetsChannel`, this could send the most recent 15 tweets).

```
class MyChannel < ApplicationCable::Channel
  def subscribed
    transmit({ foo: 123 })
  end
end
```

### Expected behavior

The `confirm_subscription` websocket event should arrive before the transmitted payload.

### Actual behavior

The payload is sent _before_ the `confirm_subscription` message is sent. Given that the above pattern feels very natural and correct, it seems like Rails should internally defer transmits until the `confirm_subscription` is sent since the present behavior is confusing and it complicates the logic on the client side of things (FYI I'm integrating with a JS app).

### System configuration
**Rails version**: 5.0.0.rc1

**Ruby version**: 2.3.0

maclover7:
This is expected behavior per [here](https://github.com/rails/rails/blob/master/actioncable/lib/action_cable/channel/base.rb#L234-244), but not sure if we should change this. cc @jeremy

machty:
Oh wow, I'm sorry for not giving this due diligence. The `defer_subscription_confirmation` API seems completely reasonable.

andreychernih:
I am having the same problem. Trying to broadcast initial payload in `Channel#subscribed` method, but it is being transmitted before subscription:

```ruby
class ScoreChannel < ApplicationCable::Channel
  def subscribed
    stream_from(broadcast_name)
    ActionCable.server.broadcast(broadcast_name, Cable::V1::ScoreSerializer.new(current_user).as_json)
  end
end
```

Produces the following in Redis:

```
1472430005.884742 [1 172.18.0.6:50242] "subscribe" "action_cable/Z2lkOi8vYXBwLWVuZ2luZS9Vc2VyLzE"
1472430010.988077 [1 172.18.0.6:50244] "publish" "score_cc112e3fdfb5411a965a31f9468abf98" "{\"score\":17,\"rank\":1}"
1472430010.988773 [1 172.18.0.6:50242] "subscribe" "score_cc112e3fdfb5411a965a31f9468abf98"
```

If this is by design then what is the conventional way for sending initial payload when client connects?

wpp:
Hi guys,

[...]

PR #55824, opened 2025-10-03
Add `success_callback` to Action Cable's `stream_from` and `stream_to`

bensheldon:
### Motivation / Background

Action Cable channels have a `success_callback` that is used in the implementation of `stream_from` and `stream_to`, which is used to ensure that streams subscriptions are established before the channel subscription confirmation is sent to the client.

👉🏻 This changes `stream_from` and `stream_to` to expose their own `success_callback` too. This callback can be used in Channels to defer behavior until the stream subscription is established. 

For example, a channel might want to ensure there aren't any missed updates between when a record is fetched and rendered for the frontend, and when an Action Cable stream subscription is ultimately established:

```ruby
class ReliableChannel < ActionCable::Base
  include Turbo::Streams::ActionHelper
  include ActionView::RecordIdentifier

  def subscribed
    resource = GlobalId::Locator.locate_signed params[:resource]
    rendered_at = Time.parse params[:rendered_at]
  
    #                    👇 👇 👇 👇 👇 👇
    stream_for(resource, success_callback: (lambda do
      resource.reload
      if resource.updated_at > rendered_at
        transmit turbo_stream_action_tag(
          :update, 
          target: dom_id(resource), 
          template: ApplicationController.render(resource, format: :html)
        )
      end
    end)
  end
end
```

### Additional information

I think this is conceptually a good change because it better aligns Stream subscription lifecycle options with Channel subscription lifecycle options.

### Checklist

### DIFF
--- a/actioncable/test/channel/stream_test.rb
+++ b/actioncable/test/channel/stream_test.rb
@@ -94,6 +94,33 @@ module ActionCable::StreamTests
       end
     end
 
+    test "success_callback is invoked" do
+      run_in_eventmachine do
+        connection = TestConnection.new
+
+        channel = OrderedCallbackChannel.new connection, "{id: 9}", id: 9
+
+        fake_pubsub = Class.new do
+          def subscribe(channel_name, handler, success)
+            success.call
+          end
+        end.new
+
+        connection.stub(:pubsub, fake_pubsub) do
+          channel.subscribe_to_channel
+          wait_for_async
+        end
+
+        assert_equal [
+          :subscribed_method_end,
+          :after_subscribe_callback,
+          :room_1_success,
+          :room_2_success,
+          :transmit_subscription_confirmation
+        ], channel.order
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


actioncable/test/channel/stream_test.rb
```diff
@@ -94,6 +94,33 @@ class StreamTest < ActionCable::TestCase
...
+          channel.subscribe_to_channel
+          wait_for_async
+        end
+
+        assert_equal [
+          :subscribed_method_end,
+          :after_subscribe_callback,
+          :room_1_success,
+          :room_2_success,
+          :transmit_subscription_confirmation
```
bensheldon, line 119: This test surfaces the `after_subscribe` callback order happening _before_ subscription confirmation. It doesn't imply that's intended though (#25333). I'm planning to look at that next.

bensheldon: That's fixed in #55825

PR #55828, opened 2025-10-03
Ensure one stream write per SSE event

almathie:
### Motivation / Background

This Pull Request has been created to optimize Server-Sent Events (SSE) streaming in `ActionController::Live::SSE` by ensuring each logical SSE is delivered via exactly one underlying `stream.write` call.

While building an AI chatbot client that consumes streaming responses (using an API that yields chunks as they arrive), I noticed a discrepancy between OpenAI's streaming responses and those produced by a Rails application using `ActionController::Live::SSE`.

OpenAI sends each SSE event as a single TCP/application chunk (i.e. one cohesive unit containing the optional `event`, `id`, `retry`, and all `data:` lines). Rails, however, was emitting *multiple* writes per logical event: one for each option/value lines and an additional one for the `data:` payload. For consumers that process chunks incrementally, this results in:

* Extra syscalls / write overhead.
* Partial events being observed mid-assembly (forcing additional buffering logic client-side).
* Reduced efficiency when the event is the natural unit of work (especially for LLM / AI streaming token consumption patterns).

This change reduces fragmentation so downstream clients receive a fully formed SSE in a single chunk whenever possible.

### Detail

This Pull Request changes the implementation of `ActionController::Live::SSE#perform_write` to build the entire event payload (all permitted option lines plus transformed multiline data) into a single mutable string and invoke `@stream.write` once per `SSE#write` call.

No public API changes are introduced; behavior is the same from a semantic perspective except for improved chunk consolidation.

### Checklist

* [x] Commit message will include a detailed description of what changed and why.
* [x] Tests are added or updated (new unit test file asserting single write per event).
* [ ] CHANGELOG files are updated if the core team deems this worthy of mention (currently omitted as an internal optimization without public behavior change).

### DIFF
--- a/actionpack/lib/action_controller/metal/live.rb
+++ b/actionpack/lib/action_controller/metal/live.rb
@@ -133,15 +133,16 @@ module ActionController
       private
         def perform_write(json, options)
           current_options = @options.merge(options).stringify_keys
-
+          event = +""
           PERMITTED_OPTIONS.each do |option_name|
             if (option_value = current_options[option_name])
-              @stream.write "#{option_name}: #{option_value}\n"
+              event << "#{option_name}: #{option_value}\n"
             end
           end
 
           message = json.gsub("\n", "\ndata: ")
-          @stream.write "data: #{message}\n\n"
+          event << "data: #{message}\n\n"
+          @stream.write event
         end
     end
 
--- a/actionpack/test/controller/sse_perform_write_test.rb
+++ b/actionpack/test/controller/sse_perform_write_test.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+require "abstract_unit"
+
+module ActionController
+  # Unit tests for ActionController::Live::SSE focusing on underlying write semantics.
+  # These tests ensure that each call to SSE#write results in exactly one call to the
+  # underlying stream#write, even when multiple options (event, retry, id) are supplied
+  # and when multiline data requires transformation.
+  class SSEPerformWriteTest < ActiveSupport::TestCase
+    class FakeStream
+      attr_reader :writes
+
+      def initialize
+        @writes = []
+      end
+
+      def write(chunk)
+        @writes << chunk
+      end
+
+      def close; end
+    end
+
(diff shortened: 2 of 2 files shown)


almathie:
I am not sure I understand the buildkite errors here.

- One is about a ruby image not found
- The other is one the active record test suite, which my PR doesn't touch at all

If anyone give me any pointer, I am willing to dig further but I am a bit puzzled

zzak:
You can ignore the `buildkite/doc-preview` failure, it happens when there hasn't been a fresh build of `main` recently.

The other failure is just a flake, I've issued a rebuild and it passed.

actionpack/lib/action_controller/metal/live.rb
```diff
@@ -133,15 +133,19 @@ def write(object, options = {})
       private
         def perform_write(json, options)
           current_options = @options.merge(options).stringify_keys
-
+          # Build the full SSE payload in a single mutable String so we only
+          # perform one underlying @stream.write per SSE event. Use a mutable
+          # String (String.new) because frozen_string_literal is enabled.
+          event = String.new
```
byroot:
```suggestion
          event = +""
```
Otherwise that string will be binary encoded which may cause issues with some servers. AFAIK SSE are UTF-8.

actionpack/lib/action_controller/metal/live.rb
```diff
@@ -133,15 +133,19 @@ def write(object, options = {})
       private
         def perform_write(json, options)
           current_options = @options.merge(options).stringify_keys
-
+          # Build the full SSE payload in a single mutable String so we only
+          # perform one underlying @stream.write per SSE event. Use a mutable
+          # String (String.new) because frozen_string_literal is enabled.
```
byroot:
```suggestion
```

That comment isn't necessary.

byroot:
Unless special circumstances, PRs should have a single commit.

actionpack/lib/action_controller/metal/live.rb
almathie: Nice catch, thanks !

PR #55833, opened 2025-10-04
[Fix #55582] Skip eager loading strategy for `#pluck`/`#calculate`/`#ids`

joshuay03:
### Motivation / Background

Closes https://github.com/rails/rails/issues/55582.

### Detail

When `pluck`, `calculate`, or `ids` are called on a relation with `includes` and `limit`/`offset`, they incorrectly apply the two-query eager loading strategy via `distinct_relation_for_primary_key`. This causes aggregation queries on empty tables to return `[]` instead of a row with null/zero values.

The two-query strategy exists to ensure the correct number of parent records are loaded when eager loading associations. For example, `Post.eager_load(:comments).limit(5)` should load exactly 5 posts, not 5 arbitrary join rows. However, `pluck`, `calculate`, and `ids` don't load records - they extract values. The optimization is unnecessary and breaks aggregation semantics.

This changes `apply_join_dependency` in these methods to pass `eager_loading: false`, skipping the two-query path. Aggregations now correctly return a result row even on empty relations, and `limit` applies to the result set rather than being used to select parent record IDs first. The `pick` method is also fixed since it calls `pluck` internally with a limit.

### Checklist

### DIFF
--- a/activerecord/test/cases/calculations_test.rb
+++ b/activerecord/test/cases/calculations_test.rb
@@ -1227,6 +1227,13 @@ class CalculationsTest < ActiveRecord::TestCase
     assert_equal [], Topic.includes(:replies).limit(1).where("0 = 1").ids
   end
 
+  def test_ids_with_includes_joins_limit_and_empty_table
+    Post.delete_all
+    Comment.delete_all
+
+    assert_equal [], Post.includes(:comments).joins(:comments).limit(1).ids
+  end
+
   def test_ids_with_includes_offset
     assert_equal [5], Topic.includes(:replies).order(:id).offset(4).ids
     assert_equal [], Topic.includes(:replies).order(:id).offset(5).ids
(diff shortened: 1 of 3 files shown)


activerecord/test/cases/calculations_test.rb
```diff
@@ -1182,6 +1189,20 @@ def test_pluck_with_includes_limit_and_empty_result
...
+    Comment.delete_all
+
+    assert_equal [], Post.includes(:comments).joins(:comments).limit(1).pluck(:id)
+  end
+
+  def test_pluck_aggregation_with_includes_joins_limit_and_empty_table
+    Post.delete_all
+    Comment.delete_all
+
+    assert_equal [[nil, 0]], Post.includes(:comments).joins(:comments).limit(1).pluck(Arel.sql("MAX(posts.id), COUNT(*)"))
```
joshuay03, line 1258: On main:
Image

activerecord/test/cases/calculations_test.rb
```diff
@@ -1442,6 +1471,13 @@ def test_pick_loaded_relation_aliased_attribute
     end
   end
 
+  def test_pick_aggregation_with_includes_joins_limit_and_empty_table
+    Post.delete_all
+    Comment.delete_all
+
+    assert_equal [nil, 0], Post.includes(:comments).joins(:comments).pick(Arel.sql("MAX(posts.id), COUNT(*)"))
```
joshuay03, line 1537: On main:
Image

PR #55836, opened 2025-10-04
[ci skip][docs] fix missing example in documentation for Delegated Types

### LINKED ISSUE #55815: documentation: missing example/explanation in documentation for Delegated Types
### Steps to reproduce

Link to documentation in question: https://guides.rubyonrails.org/association_basics.html#setting-up-delegated-types

On the page for Association Basics, under the section for Delegated Types, the following text appears:

> In order to apply this to our example above, we need to regenerate our models.

There's no example above which seems relevant. The immediate example above is about vehicles, whereas the examples for Delegated Types are about comments. There do not appear to be any models to "regenerate", since the examples in this section are the first mention of "message" or "comment" in the context of model names.

### Expected behavior

So that the context and intention of the documentation is clear, I expect that the sample data models would be explained somewhere in this section on Delegated Types, or that the examples in the Delegated Types would be written using existing data models used elsewhere on the page.

I'll make a PR if someone can tell me what should be present or changed.

### Actual behavior

There is a reference to "above" and "existing models" which do not seem to have a referent. There seems to be an expectation that the user should be familiar with the examples being used and what they are trying to accomplish. 
 
### System configuration

**Rails version**: 

**Ruby version**:

drymar:
### Motivation / Background

The "Delegated Types" section in the Association Basics guide contained the phrase:
> "In order to apply this to our example above, we need to regenerate our models."

This was confusing because the preceding example is about STI and vehicles, which is unrelated to delegated types. As a result, readers could not easily understand what "example above" referred to.

This Pull Request clarifies the documentation so that the Delegated Types section is self-contained and easier to follow.

Fixes [#55815](https://github.com/rails/rails/issues/55815)

### Detail

This Pull Request changes the `documentation in guides/source/association_basics.md` by:
- Adding a **NOTE** clarifying that Delegated Types examples are independent from the STI “Vehicle” example.
- Rewording the introduction in _6.2 Generating Models_ to explicitly reference the `Entry`, `Message`, and `Comment` models instead of “example above.”

No code changes were made. Only documentation was updated.

### Additional information

- Improves clarity for newcomers learning Delegated Types.
- Ensures the section stands alone without relying on unrelated STI examples.

### Checklist

* [x] Commit message has a detailed description of what changed and why. Includes reference to issue #55815.
* [x] No tests or CHANGELOG updates needed (documentation-only change).

### DIFF
--- a/guides/source/association_basics.md
+++ b/guides/source/association_basics.md
@@ -1961,67 +1961,75 @@ To use delegated types, we need to model our data as follows:
 This eliminates the need to define attributes in a single table that are
 unintentionally shared among all subclasses.
 
+NOTE: The delegated types examples below are independent of the STI “Vehicle” example in the previous section.
+These examples start from scratch with an `Post` model(the delegator) and three concrete types: `TextPost`, `ImagePost` and `VideoPost`.
+
 ### Generating Models
 
-In order to apply this to our example above, we need to regenerate our models.
+Generate these models as follows:
 
-First, let's generate the base `Entry` model which will act as our superclass:
-
-```bash
-$ bin/rails generate model entry entryable_type:string entryable_id:integer
-```
-
-Then, we will generate new `Message` and `Comment` models for delegation:
+First, let's generate the base `Post` model which will act as our superclass:
 
 ```bash
-$ bin/rails generate model message subject:string body:string
-$ bin/rails generate model comment content:string
+$ bin/rails generate model post postable_type:string postable_id:integer
 ```
 
-NOTE: If you don't specify a type for a field (e.g., `subject` instead of `subject:string`), Rails will default to type `string`.
+Then, we will generate new `TextPost`, `ImagePost` and `VideoPost` models for delegation:
+
+```bash
+$ bin/rails generate model text_post title:string body:string
+$ bin/rails generate model image_post title:string url:string
+$ bin/rails generate model video_post title:string url:string
+```
+
+NOTE: If you don't specify a type for a field (e.g., `title` instead of `title:string`), Rails will default to type `string`.
 
 After running the generators, our models should look like this:
 
 ```ruby
-# Schema: entries[ id, entryable_type, entryable_id, created_at, updated_at ]
-class Entry < ApplicationRecord
+# Schema: posts[ id, postable_type, postable_id, created_at, updated_at ]
+class Post < ApplicationRecord
 end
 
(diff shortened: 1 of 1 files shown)


guides/source/association_basics.md
```diff
@@ -1961,9 +1961,13 @@ To use delegated types, we need to model our data as follows:
 This eliminates the need to define attributes in a single table that are
 unintentionally shared among all subclasses.
 
+NOTE: The delegated types examples below are independent of the STI “Vehicle” example in the previous section.
```
shivabhusal, line 1964: I’m not sure this change adds much value.

That said, I think there’s a good opportunity to improve the example below. The current `Entry` example feels a bit abstract and might be harder for new readers to relate to. Perhaps we could use something like a `Post` model with `TextPost`, `ImagePost`, and `VideoPost` types — that might make the concept clearer and more intuitive.

let me know what you think. Thank you for the contribution 😄

guides/source/association_basics.md
```diff
@@ -1961,9 +1961,13 @@ To use delegated types, we need to model our data as follows:
...
+These examples start from scratch with an `Entry` model(the delegator) and two concrete types: `Message` and `Comment`.
+
 ### Generating Models
 
-In order to apply this to our example above, we need to regenerate our models.
```
shivabhusal, line 1966: Simply removing this line should work, I think.

drymar: @shivabhusal – thank you, I agree. It may be more clean and explicit. It is updated and looks like below:
image

guides/source/association_basics.md
drymar: @shivabhusal – that makes sense to me. It may be more logical to use more real world example. I can do it in a separate PR. Or would you prefer updating it here?

Thank you

shivabhusal: I think updating in this PR should work

drymar: @shivabhusal – ok thank you. I just pushed an updated doc 👍

shivabhusal:
<table>
<td>
<h2>Your Version</h2>

https://drymar-fix-delegated-types-d.rails-docs-preview.pages.dev/guides/association_basics#setting-up-delegated-types
image

</td>

<td>
<h2>Original Version</h2>

https://guides.rubyonrails.org/association_basics.html#delegated-types
image

</td>
</table>

PR #55838, opened 2025-10-04
Add structured event for Rails deprecations, when `config.active_support.deprecation` is set to `:notify`.

zzak:
Opted to make it dependent on the `:notify` behavior than adding yet another behavior just for structured events, but happy to go that way if there is a strong opion.

/cc @gmcgibbon @adrianna-chang-shopify

### DIFF
--- a/railties/test/structured_event_subscriber_test.rb
+++ b/railties/test/structured_event_subscriber_test.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require "abstract_unit"
+require "active_support/testing/event_reporter_assertions"
+require "rails/structured_event_subscriber"
+
+module Rails
+  class StructuredEventSubscriberTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::EventReporterAssertions
+
+    def run(*)
+      ActiveSupport.event_reporter.with_debug do
+        super
+      end
+    end
+
+    def test_deprecation_is_notified_when_behavior_is_notify
+      Rails.deprecator.with(behavior: :notify) do
+        event = assert_event_reported("rails.deprecation", payload: { gem_name: "Rails" }) do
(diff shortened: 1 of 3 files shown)


railties/test/structured_event_subscriber_test.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require "abstract_unit"
+require "active_support/testing/event_reporter_assertions"
+require "rails/structured_event_subscriber"
+
+module Rails
+  class StructuredEventSubscriberTest < ActionMailer::TestCase
```
adrianna-chang-shopify:
```suggestion
  class StructuredEventSubscriberTest < ActiveSupport::TestCase
```

adrianna-chang-shopify (approved):

railties/test/structured_event_subscriber_test.rb
zzak: Whoops haha :joy:

railties/test/structured_event_subscriber_test.rb
```diff
@@ -0,0 +1,42 @@
...
+require "abstract_unit"
+require "active_support/testing/event_reporter_assertions"
+require "rails/structured_event_subscriber"
+
+module Rails
+  class StructuredEventSubscriberTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::EventReporterAssertions
+
+    def run(*)
+      ActiveSupport.event_reporter.with_debug do
```
gmcgibbon, line 12: This is only needed for debug events. Did we mean to use `emit_debug_event`? Debug events only happen in development by default, so I assume not. Perhaps this can be removed.

zzak: Whoops, nice catch, didn't mean to copy this into the test :sweat_smile:

PR #55839, opened 2025-10-05
Skip analyze job if ruby-vips or mini_magick gem are missing

zzak:
If you're using Active Storage, but don't have `image_processing` gem installed, because the [default analyzers][^1] includes Vips and ImageMagick analyzers, you would see the following error:

```
[ActiveJob] Enqueued ActiveStorage::AnalyzeJob ...
[ActiveJob] [ActiveStorage::AnalyzeJob] Performing ActiveStorage::AnalyzeJob ...
[ActiveJob] [ActiveStorage::AnalyzeJob] [0efa0f81-23e0-4a09-9fb2-55ecb484c10a] Error performing ActiveStorage::AnalyzeJob: NameError (uninitialized constant Vips):
```

Previously, this would rescue LoadError because the gems were required lazily.[^2]

```
def read_image
  begin
    require "ruby-vips"
  rescue LoadError
    logger.info "Skipping image analysis because the ruby-vips gem isn't installed"
    return {}
  end

  # ...
```

---

Thanks to @claudiob for [reporting][^3] and @Earlopain for all your help with this issue!

[^1]: https://guides.rubyonrails.org/configuring.html#config-active-storage-analyzers
[^2]: https://github.com/rails/rails/blob/ab8e833991e493f57f03c10ee2a8a7fda218faae/activestorage/lib/active_storage/analyzer/image_analyzer/vips.rb#L13-L18
[^3]: https://github.com/rails/rails/pull/55697#issuecomment-3364212327

### DIFF
--- a/activestorage/test/analyzer/image_analyzer/image_magick_test.rb
+++ b/activestorage/test/analyzer/image_analyzer/image_magick_test.rb
@@ -70,6 +70,23 @@ class ActiveStorage::Analyzer::ImageAnalyzer::ImageMagickTest < ActiveSupport::T
     end
   end
 
+  test "when image_magick is not installed" do
+    stub_const(ActiveStorage, :MINIMAGICK_AVAILABLE, false) do
+      blob = create_file_blob(filename: "racecar.jpg", content_type: "image/jpeg")
+
+      output = StringIO.new
+      logger = ActiveSupport::Logger.new(output)
+
+      ActiveStorage.with(logger: logger) do
+        analyze_with_image_magick do
+          blob.analyze
+        end
+      end
+
+      assert_includes output.string, "Skipping image analysis because the mini_magick gem isn't installed"
+    end
+  end
+
   private
     def analyze_with_image_magick
       previous_processor, ActiveStorage.variant_processor = ActiveStorage.variant_processor, :mini_magick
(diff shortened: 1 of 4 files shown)


Earlopain (commented):
I think it would be better to instead save if the require succeeded instead of doing this error check. Something like

```rb
# need to move this further down into the class I think
begin
  require "ruby-vips"
  VIPS_AVAILABLE = true
rescue
  VIPS_AVAILABLE = false
end
```

You can then return early.

I was about to suggest to then simply put this into `accept?` but I think you want to log in such cases (like it was previously/you already do in this PR).

zzak:
:+1: I will do that, because it will make it easier to test too. Thanks!

activestorage/test/analyzer/image_analyzer/image_magick_test.rb
```diff
@@ -70,6 +70,28 @@ class ActiveStorage::Analyzer::ImageAnalyzer::ImageMagickTest < ActiveSupport::T
     end
   end
 
+  test "when image_magick is not installed" do
+    old_value = ActiveStorage.const_get(:MINIMAGICK_AVAILABLE)
```
Earlopain: You can use [stub_const](https://api.rubyonrails.org/classes/ActiveSupport/Testing/ConstantStubbing.html#method-i-stub_const) for this

zzak: Thank you, I was looking for that but had to step away. Fixed!

PR #55841, opened 2025-10-05
Fix remove_foreign_key compatibility signature

zzak:
Closes #55823

/cc @that-jill

### DIFF
--- a/activerecord/lib/active_record/migration/compatibility.rb
+++ b/activerecord/lib/active_record/migration/compatibility.rb
@@ -33,7 +33,7 @@ module ActiveRecord
 
       class V8_0 < V8_1
         module RemoveForeignKeyColumnMatch
-          def remove_foreign_key(from_table, to_table = nil, **options)
+          def remove_foreign_key(*args, **options)
             options[:_skip_column_match] = true
             super
           end
--- a/activerecord/test/cases/migration/compatibility_test.rb
+++ b/activerecord/test/cases/migration/compatibility_test.rb
@@ -718,6 +718,26 @@ module ActiveRecord
         ActiveRecord::Base.clear_cache!
       end
 
+      def test_remove_foreign_key_on_7_0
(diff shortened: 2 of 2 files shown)


that-jill (commented):
For thoroughness, I tested this on the application I saw fail and it succeeded. I am a bit worried that we may eventually get bitten by the bug that is #proper_table_name being passed in a nil and returning an empty string. But overall I think this solution is elegant. Do you know which PR this bug came from?

zzak:
@that-jill Thanks for trying it!

This bug was introduced (by me) in #55086.

Basically the method signature should be:

```ruby
def remove_foreign_key(*args, **options)
  raise_on_if_exist_options(options)
  @base.remove_foreign_key(name, *args, **options)
end
```

It comes from the schema table definition:
https://github.com/rails/rails/blob/4dee2b25c0db74e6672c35092ff71f3de28a81ad/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L902-L905

But the method definition I was using comes from the schema statements:
https://github.com/rails/rails/blob/4dee2b25c0db74e6672c35092ff71f3de28a81ad/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1247-L1257

```ruby
def remove_foreign_key(from_table, to_table = nil, **options)
```

Because of this you end up with an argument that is empty, when I put a print here and run your test:
https://github.com/rails/rails/blob/4dee2b25c0db74e6672c35092ff71f3de28a81ad/activerecord/lib/active_record/connection_adapters/sqlite3/schema_statements.rb#L65

```
remove_foreign_key(from_table, to_table, options)
remove_foreign_key(sub_testings, "", {column: "testing_id", _skip_column_match: true})
```

Sorry that's not a great explanation, I don't think we should ever get to `proper_table_name` with an empty string though otherwise. There is such a thing as overly defensive results in restrictive APIs, which I think Rails tries to avoid.

Issue #55843, opened 2025-10-05
Security Vulnerability in Cookie Storage Technique for Rails authentication generator


n-at-han-k:
Referring to the issue @ joemasilotti/daily-log#21

> [By default](https://api.rubyonrails.org/classes/ActionDispatch/Cookies.html), Rails doesn't store a cookie as "secure". This means that the cookie is used for both HTTP and HTTPS requests. This means that during an unencrypted HTTP request, the cookie is completely visible.
> 
> At first thought, this doesn't sound like an issue. If your website/app never makes an unencrypted HTTP request then you're safe aren't you? Yes you are. Until a bad actor who is also snooping on the network gets you to make an HTTP request instead. That request may fail (or not, depending on your server config) but it doesn't matter, the bad actor only wanted to read the request, and now they have the cookie. The unsuspecting user would have absolutely no idea that their credentials had been stolen.
> 
> This [stack overflow post](https://stackoverflow.com/questions/4015746/is-a-cookie-secure-in-a-https-connection#4015769) has more details. A comment from a user highlighted the error in the answer.

This section of code should be replaced:

https://github.com/rails/rails/blob/dac26fdf22498d79fe78501c55fdbf386f20590a/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt#L41-L46

Something like this would suffice

```ruby
def start_new_session_for(user)
  user.sessions.create!(user_agent: request.user_agent, ip_address: request.remote_ip).tap do |session|
    Current.session = session
    cookies.signed.permanent[:session_id] = {
      value: session.id,
      httponly: true,
      same_site: :lax,
      secure: Rails.env.production? # <- new value here
    }
  end
end
```

Ref:
Pull request @ joemasilotti/daily-log#22

PR #55844, opened 2025-10-05
Enhance Cookie Security in Authentication Generator to prevent man-in-the-middle attack via #55843

n-at-han-k:
### Motivation / Background

This Pull Request has been created because of security vulnerability detailed in #55843. The default rails authentication generator produces code that could allow a targeted man-in-the-middle attack to take place.

### Detail

This Pull Request changes the options passed when creating the session cookie.
Simply adding the `secure: true` option stops the cookie being sent over insecure connections.

### Additional information

- [Rails Issue](https://github.com/rails/rails/issues/55843)
- [Stack overflow post explaining vulnerability](https://stackoverflow.com/questions/4015746/is-a-cookie-secure-in-a-https-connection#4015769)
- [Rails Docs which show that the secure option defaults to false](https://api.rubyonrails.org/classes/ActionDispatch/Cookies.html)

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
+++ b/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
@@ -41,7 +41,12 @@ module Authentication
     def start_new_session_for(user)
       user.sessions.create!(user_agent: request.user_agent, ip_address: request.remote_ip).tap do |session|
         Current.session = session
-        cookies.signed.permanent[:session_id] = { value: session.id, httponly: true, same_site: :lax }
+        cookies.signed.permanent[:session_id] = {
+          value: session.id,
+          httponly: true,
+          same_site: :lax,
+          secure: Rails.env.production? # prevents man-in-the-middle attack by telling the browser to only send over HTTPS.
+        }
       end
     end
 
--- a/.github/workflows/tag-gem-version-bump.yml
+++ b/.github/workflows/tag-gem-version-bump.yml
@@ -1,0 +1,47 @@
+name: Auto-tag on version bump
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  tag:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.4'
+
+      - name: Read version from gemspec
+        id: read
+        run: |
+          version=$(ruby -e '
+            spec = Gem::Specification.load(Dir["*.gemspec"].first)
+            puts spec.version
+          ')
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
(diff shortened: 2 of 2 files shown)


railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
```diff
@@ -41,7 +41,12 @@ module Authentication
     def start_new_session_for(user)
       user.sessions.create!(user_agent: request.user_agent, ip_address: request.remote_ip).tap do |session|
         Current.session = session
-        cookies.signed.permanent[:session_id] = { value: session.id, httponly: true, same_site: :lax }
+        cookies.signed.permanent[:session_id] = {
+          value: session.id,
+          httponly: true,
+          same_site: :lax,
+          secure: Rails.env.production? # prevents man-in-the-middle attack by telling the browser to only send over HTTPS.
```
bdewater-thatch, line 48: You'll also want this for other deployments (eg staging), so to enable it for anything except development and test:
```suggestion
          secure: !Rails.env.local? # prevents man-in-the-middle attack by telling the browser to only send over HTTPS.
```

rafaelfranca: The more I look at this I think we should just use the `session` object. It knows to set the `secure` all the other options by default.

I don't remember why we didn't use the `session` object here. @dhh?

dhh: I thought default cookie session behavior was that it only lasted until the browser is closed? We want this session to be permanent.

rafaelfranca: Yes, that is true. There is no `Expire` in the session cookies, so it would only last until the browser is closed.

I'll see what we can do here. It is easier to get security right with the session store than in regular cookies, and just allow the session store to write explicit expiration dates.

dhh: It feels like mixing things together a little. I think folks would be surprised if the session suddenly lasted beyond the scope of the single browser session. I'd rather find a way to have "secure cookies" be something that's easy to do. It shouldn't be hard to get this right. Maybe we can do something like `cookies.secure[:session_id] =` ala how we have `.permanent`. Or we simply make the cookie setting secure by default! (even better, but there could be backwards compatibility issues there).

htetwunsan: Btw, setting force_ssl to `true` will make cookies secure by default. However, it is not mentioned in the documentation. And it is already set to true in `production.rb` config template.

References:
- Configuration guide: https://edgeguides.rubyonrails.org/configuring.html#config-force-ssl
- Test coverage: https://github.com/rails/rails/blob/main/railties/test/application/middleware/session_test.rb#L24
- Production config template: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L34

PR #42283, opened 2021-05-24
cause rails to correctly place optional path parameter booleans

HParker:
Previously, if you specify a url parameter that is part of the path as false it would include that part of the path as parameter for example:
`get "(/optional/:optional_id)/things" => "foo#foo", as: :things`
`things_path(optional_id: false)  # => /things?optional_id=false`

this is not the case for empty string,
`things_path(optional_id: '')  # => "/things"

this is due to a quark in how path parameters get removed from the parameters.

we where doing `(paramter || recall).nil?` which returns nil if both values are nil however it also return nil if one value is false and the other value is nil. i.e. `(false || nil).nil # => nil` which is confusing.

After this change, `true` and `false` will be treated the same when used as optional path parameters. meaning now

```
get '(this/:my_bool)/that' as: :that

that_path(my_bool: true) # => `/this/true/that`
that_path(my_bool: false) # => `/this/false/that`
```

PR #55846, opened 2025-10-06
Treat empty strings as nil for optional route segments

### LINKED ISSUE #55845: Optional route segments treat empty strings inconsistently with nil values
When using routes with optional segments, nil values are correctly omitted from the generated path, but empty strings ('') are not, unless they are at the end of the route pattern. This creates malformed URLs with double slashes and incorrect query parameters.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'rails', github: 'rails/rails', branch: 'main'
end

require 'action_controller/railtie'
require 'minitest/autorun'

  config.root = __dir__
  config.eager_load = false
end

Rails.application.initialize!

  get '(/a/:a_id)(/b/:b_id)(/c/:c_id)', to: 'test#index', as: :optional_segments
end

  def index
    render plain: 'Hello'
  end
end

  def test_leading_optional_segment_with_empty_string
    assert_equal '/b/2/c/3', optional_segments_path(a_id: '', b_id: 2, c_id: 3)
  end

  def test_leading_optional_segment_with_nil
    assert_equal '/b/2/c/3', optional_segments_path(a_id: nil, b_id: 2, c_id: 3)
  end

  def test_trailing_optional_segment_with_empty_string
    assert_equal '/a/1/b/2', optional_segments_path(a_id: 1, b_id: 2, c_id: '')
  end

  def test_trailing_optional_segment_with_nil
    assert_equal '/a/1/b/2', optional_segments_path(a_id: 1, b_id: 2, c_id: nil)
  end

  def test_middle_optional_segment_with_empty_string
    assert_equal '/a/1/c/3', optional_segments_path(a_id: 1, b_id: '', c_id: 3)
  end

  def test_middle_optional_segment_with_nil
    assert_equal '/a/1/c/3', optional_segments_path(a_id: 1, b_id: nil, c_id: 3)
  end
end
```

### Expected behavior

Empty strings should be treated the same as nil for optional route segments, wherever these blank values occur, meaning that both my_path(a: "", b: 2, c: 3) and my_path(a: nil, b: 2, c: 3) give "/b/2/c/3"

### Actual behavior

Empty strings in anything but the final optional segment position fail to omit the key and add a trailing slash, the example above - my_path(a: "", b: 2, c: 3) with a route pattern of "(/a/:a)(/b/:b)(/c/:c)" will give "/a//b/2/c/3" instead of "/b/2/c/3"

### System configuration

**Rails version**: 8.0.3

**Ruby version**: 3.4.6

carldaws:
### Motivation / Background

Fixes #55845.

Rails omits an empty string from a single optional route segment, but an empty string in a non-trailing optional group remains in the generated path:

```ruby
get "(/a/:a_id)(/b/:b_id)(/c/:c_id)", to: "test#index", as: :optional_segments

optional_segments_path(a_id: nil, b_id: 2, c_id: 3) # => "/b/2/c/3"
optional_segments_path(a_id: "", b_id: 2, c_id: 3)  # => "/a//b/2/c/3"
```

This is inconsistent with the established empty-string behavior covered by #42283 and commonly surfaces when route values come from form parameters.

### Detail

This change normalizes empty strings to `nil` for optional route parts before parameter extraction, allowing the existing nil handling to omit the corresponding optional group.

The normalization is intentionally limited to optional parts. Regression coverage verifies that required empty segments, `false` route values, and unrelated empty query parameters retain their existing behavior.

### Testing

- `bin/test actionpack/test/dispatch/routing_test.rb actionpack/test/dispatch/url_generation_test.rb`
- `PARALLEL_WORKERS=1 actionpack/bin/test` (4,249 tests, 19,971 assertions)
- `bundle exec rubocop actionpack/lib/action_dispatch/journey/formatter.rb actionpack/test/dispatch/routing_test.rb`

### Checklist

- [x] The commit message describes what changed and why.
- [x] Tests are added for the bug and compatibility boundaries.
- [x] No CHANGELOG entry is included because this is a minor bug fix.

### DIFF
--- a/actionpack/lib/action_dispatch/journey/formatter.rb
+++ b/actionpack/lib/action_dispatch/journey/formatter.rb
@@ -71,6 +71,10 @@ module ActionDispatch
         missing_keys = nil
 
         match_route(name, constraints) do |route|
+          (route.parts - route.required_parts).each do |part|
+            options[part] = nil if options[part] == ""
+          end
+
           parameterized_parts = extract_parameterized_parts(route, options, path_parameters)
 
           # Skip this route unless a name has been provided or it is a standard Rails
--- a/actionpack/test/dispatch/routing_test.rb
+++ b/actionpack/test/dispatch/routing_test.rb
@@ -2713,6 +2713,42 @@ class TestRoutingMapper < ActionDispatch::IntegrationTest
       url_for(controller: "photos", action: "index", username: nil)
   end
 
+  def test_url_generator_treats_empty_strings_like_nil_for_optional_segments
+    draw do
+      get "(/locale/:locale)(/currency/:currency)/products(/:id)" => "products#show"
+    end
+
+    assert_equal "http://www.example.com/currency/USD/products/123",
+      url_for(controller: "products", action: "show", locale: "", currency: "USD", id: 123)
+    assert_equal "http://www.example.com/currency/USD/products/123",
+      url_for(controller: "products", action: "show", locale: nil, currency: "USD", id: 123)
+
+    assert_equal "http://www.example.com/locale/en/products/123",
+      url_for(controller: "products", action: "show", locale: "en", currency: "", id: 123)
+    assert_equal "http://www.example.com/locale/en/products/123",
+      url_for(controller: "products", action: "show", locale: "en", currency: nil, id: 123)
+
+    assert_equal "http://www.example.com/locale/en/currency/USD/products",
+      url_for(controller: "products", action: "show", locale: "en", currency: "USD", id: "")
+    assert_equal "http://www.example.com/locale/en/currency/USD/products",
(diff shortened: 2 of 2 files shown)


carldaws:
Rebased onto current main and refreshed the patch. The normalization is now limited to optional route parts, with regression coverage preserving required empty segments, false route values, and unrelated empty query parameters. The focused routing tests, full Action Pack suite (4,249 tests / 19,971 assertions, run serially), and RuboCop all pass locally. This is ready for another look. Thanks!

carldaws:
Closing in favor of #58412, which fixes the same issue without changing shared state while candidate routes are being tried.

PR #55856, opened 2025-10-06
Refactor gcs_service public interface to bring more in line with s3_service

benngarcia:
The s3_service exposes it's respective client and bucket attributes publicly. I think this makes sense in the context of active storage, as it allows users to rely on the same adapter & configs that the framework is using for Activestorage without having to duplicate your own client and ensure it matches the original, which can be modified at runtime, in order to perform operations outside of the scope of ActiveStorage.

This commit serves to make the same true of the gcs_service, which currently defined it's respective client & bucket as private method definitions.

I did a bit of git snooping and didn't see any mention of this being an intentional decision when first released.

### DIFF
--- a/activestorage/lib/active_storage/service/gcs_service.rb
+++ b/activestorage/lib/active_storage/service/gcs_service.rb
@@ -10,12 +10,17 @@ module ActiveStorage
   # Wraps the Google Cloud Storage as an Active Storage service. See ActiveStorage::Service for the generic API
   # documentation that applies to all services.
   class Service::GCSService < Service
+    attr_reader :client, :bucket
     class MetadataServerError < ActiveStorage::Error; end
     class MetadataServerNotFoundError < ActiveStorage::Error; end
 
     def initialize(public: false, **config)
-      @config = config
+      @client = Google::Cloud::Storage.new(**config.except(:bucket, :cache_control, :iam, :gsa_email))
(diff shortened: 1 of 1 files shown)


activestorage/lib/active_storage/service/gcs_service.rb
```diff
@@ -188,14 +193,6 @@ def stream(key)
...
       end
 
-      def bucket
-        @bucket ||= client.bucket(config.fetch(:bucket), skip_lookup: true)
-      end
-
-      def client
-        @client ||= Google::Cloud::Storage.new(**config.except(:bucket, :cache_control, :iam, :gsa_email))
-      end
-
```
chaadow, line 198: I think these methods should have been just moved to be public

because when defined in the `def initialize` method, they can potentially perform a network call at initialization ( to retrieve bucket metadata or idk), instead of when they're actually used

rafaelfranca: Good point. I thought the same thing, but since the others also do in the initialize I thought it would be fine. I'll fix it.

chaadow: @rafaelfranca Thanks!  i've just tweaked your fix a bit more here: https://github.com/rails/rails/pull/55953

PR #55862, opened 2025-10-07
[RF-Docs] [ci-skip] Rewrite of the Rails Initialization Process Guide

bhumi1102:
## Overview

The goal of this PR is to add documentation for the Rails Initialization Process Guide.

Note: This guide has be rewritten entirely. I kept the the old version at the bottom so that the "diff" is clean and easy to read (without meaningless +-). The old guide will of course be deleted after the review and before merging.


bhumi1102:
CC: @fxn This one is ready for your review.

chabgood:
👍

fxn:
Hi! I have done a pass until the `Rails.application.initialize!` section. Some notes!

---

> At a high level, the Rails initialization process has two parts

I think that would rather be:

At a high level, when you launch the web server two things happen:

1. The application is loaded and initialized
2. The web server is started

---

> This is where your application is defined and configured.

You could say the application is defined and leave it there. Some configuration goes there, but configuration is spread over multiple files. So that file is not where the application is configured, really.

That would be also consistent with the wording later in "Requiring config/application".

---

> (The config.ru file has this line, as we'll see in a later section

Missing closing parenthesis.

---

> The config/boot.rb file is responsible for loading Bundler and setting it up. The boot.rb file contains:

The `config/boot.rb` file is responsible for setting up Bundler **and Bootsnap**.

---

> Requiring application.rb in ServerCommand#perform

For consistency, we always write file names relative to the application root, so `config/application.rb`. This might need a pass over the document.

---

A generic comment for your consideration.

While the `server` command is a common one, as `console` is, for example, I wonder if its inherent complexity, lines of code in its listing, relationship with `config.ru`, etc. is unnecessary overhead when explaining how things work.

The intro says that in order to boot from a Ruby script you load `config/environment.rb`. That is waaayy simple, that goes to the core of it.

Another option would be to base this on the `boot` command, which involves "commands" if you will, but it is minimal. Just what you want the reader to understand.

Then, with things explained clearly, you can link to other commands saying that other commands like such and such load and initialize the application in a similar way, and add their own logic on top of that.

---

> NOTE: Zooming out, so far we have been following what happens in the bin/rails ruby script

Ruby, capital "R".

---

> the main line is when config/boot.rb is required, which sets up the Gems.

Sets up Bundler and Bootsnap, or sets up "the gem dependencies" (lowercase "g").

---

> call is loading all the Railties and Engines from various Rails framework components

We normally use lowercase "railties" and "engines". The same way we do not capitalize "application". Look for instance at the [guide for engines](https://guides.rubyonrails.org/engines.html).

May need a pass too.

---

> ServerCommand#perform

In general, it is better to use the full constant path, because, while you can understand from the context, `ServerCommand` is technically a top-level constant that does not exist.

bhumi1102:
I've applied all of the suggested changes above and made a pass over the entire guide for the following: lowercase railtie and engine, ServerCommand to Rails::Command::ServerCommand, file names relative to Rails root 'config/application.rb', gem lowercase.

@fxn - want to clarify the remaining comment: 

> While the server command is a common one, as console is, for example, I wonder if its inherent complexity, lines of code in its listing, relationship with config.ru, etc. is unnecessary overhead when explaining how things work. ...

I'm trying to recall my thinking from one year ago when I wrote this guide. I think I kept the perspective of "here's what happens run when you run `bin/rails server`" from the previous version of this guide. I agree that the the `server` command is the most complex and brings in config.ru, Rack, and puma. Since this guide is called the Rails initialization Guide, we don't necessarily have to focus on explaining what happens when you start the server though. I am up for simplifying this guide further and focusing on the `require "config/environment"` and removing the focus on `server`. Is that what you mean by the above comment?

PR #55863, opened 2025-10-07
[Getting Started] Use .positive? for more explicit positive checks rather than implicitly field itself [ci skip]

harsh183:
### Motivation / Background

In the tutorial we use `product.inventory_count?` to check if the product is in stock. This is relying on implicit rails behavior that can be confusing and hard to reason about for newcomers.

### Detail

This change uses `product.inventory_count.positive?` instead, which is more explicit and easier to understand. I also standardized on using it throughout the tutorial for consistency.

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -1975,6 +1975,17 @@ stock. We can generate this migration using the following command:
 $ bin/rails generate migration AddInventoryCountToProducts inventory_count:integer
 ```
 
+This will generate a migration file. Open it and add a default value of `0` to
+ensure `inventory_count` is never `nil`:
+
+```ruby
+class AddInventoryCountToProducts < ActiveRecord::Migration[8.1]
+  def change
+    add_column :products, :inventory_count, :integer, default: 0
+  end
+end
+```
+
 Then let's run the migration.
 
 ```bash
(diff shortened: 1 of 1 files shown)


guides/source/getting_started.md
```diff
@@ -2119,7 +2119,7 @@ Create a new partial at `app/views/products/_inventory.html.erb` and add the
 following:
 
 ```erb
-<% if product.inventory_count? %>
+<% if product.inventory_count.positive? %>
```
MatheusRich, line 2133: Just double checking that this is never `nil`

MatheusRich (commented):
I'm positive (pun intended) about this change.

MatheusRich (approved):

guides/source/getting_started.md
harsh183: I'll double check in my tutorial code. If it is nil, maybe it's a good time to introduce default column values.

harsh183: Or I can do &.positive?, that's easier

p8:
This has been discussed before: https://github.com/rails/rails/pull/55344

harsh183:
Nice! Glad to know I wasn't the only person who thought that was a little confusing for beginners.

guides/source/getting_started.md
harsh183: This was a real error to worry about. I added a `default: 0` when introducing the migration to guard against this. As a bonus, it's another nice rails feature to throw in the getting started tutorial.

PR #55864, opened 2025-10-07
Fix typo in idle_time method docs

RicardoTrindade:
[idle_time](https://api.rubyonrails.org/classes/ActiveSupport/Notifications/Event.html#method-i-idle_time) method contains I think a duplicate word time that can be removed.


Edouard-chin:
Thank you

PR #55868, opened 2025-10-08
avoid direct interpolation for args in structure_load

kwkr:
### Motivation / Background

This Pull Request has been created because the current implementation of `structure_load` in `ActiveRecord::Tasks::SQLiteDatabaseTasks` uses a backtick-executed shell command to invoke `sqlite3`.  
This approach directly interpolates supplied values (such as the database name) into the command string, which could allow **command injection** if a maliciously crafted database path is used.  

By switching to `Kernel.system` with properly separated arguments, we prevent unescaped shell interpretation and ensure the command is executed safely.  

No functional behavior changes are expected under normal conditions.

From H1 report @jhawthorn @matthewd 

### Checklist

### DIFF
--- a/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
@@ -51,8 +51,10 @@ module ActiveRecord
       end
 
       def structure_load(filename, extra_flags)
-        flags = extra_flags.join(" ") if extra_flags
-        `sqlite3 #{flags} #{db_config.database} < "#{filename}"`
+        args = []
+        args.concat(extra_flags) if extra_flags
+        args << db_config.database
+        run_cmd("sqlite3", *args, in: filename)
       end
 
       def check_current_protected_environment!(db_config, migration_class)
--- a/railties/test/application/rake/multi_dbs_test.rb
+++ b/railties/test/application/rake/multi_dbs_test.rb
@@ -106,7 +106,7 @@ module ApplicationTests
             assert_match(/CREATE TABLE (?:IF NOT EXISTS )?"dogs"/, schema_dump_animals)
           end
 
-          rails "db:schema:load"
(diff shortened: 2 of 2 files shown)


gregmolnar:
How is the database name a user supplied value? That is set by the developer.

kwkr:
sorry, my bad, this is not user-supplied. I edited the description.

gregmolnar:
But how would be a database name be maliciously crafted? I believe this is a no issue at all.

kwkr:
The suggestion to create this PR came from feedback I received on a HackerOne report I submitted. The original concern was that a malicious `config/database.yml` could be used to inject commands, but this was dismissed because ERB evaluation would already allow code execution in that scenario. An additional benefit of this change is that it would properly handle filenames and database paths containing spaces.

gregmolnar:
> An additional benefit of this change is that it would properly handle filenames and database paths containing spaces.

This is a valid point, but there is no command injection issue with the original implementation.

kwkr:
updated the title so it doesn't sound like a security issue.

zzak:
Personally I find this test harder to read, and since there is no attack surface I don't see the point of obfuscating the test in the name of security.

byroot:
I'm in favor of this change, I indeed don't think it is a security issue, but it's always preferable to supply subprocess arguments as an array of strings to avoid escaping issues and such.

However the CI failures do look legit. Maybe because previously the command output was discard and that's no longer the case?

byroot:
> Maybe because previously the command output was discard and that's no longer the case?

Actually no. Previously there was no error checking at all, so the failure was swept under the rug.

byroot:
Given the tests that had to be updated, I'm now 100% convinced this is a positive change. All other adapters worked this way, and here we were sweeping errors under the rug.

rafaelfranca:
Agree. This change is great. Thank you for the pull request

PR #55875, opened 2025-10-08
feat(sqlite3): Restarting the serve automatically after dropping DB

yokonao:
### Motivation / Background

This Pull Request has been created because #55874 

### Detail

Restarting the serve automatically after dropping DB
close #55874 

### Checklist

### DIFF
--- a/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
@@ -3,9 +3,10 @@
 module ActiveRecord
   module Tasks # :nodoc:
     class SQLiteDatabaseTasks < AbstractTasks # :nodoc:
-      def initialize(db_config, root = ActiveRecord::Tasks::DatabaseTasks.root)
+      def initialize(db_config, root = ActiveRecord::Tasks::DatabaseTasks.root, env = ActiveRecord::Tasks::DatabaseTasks.env)
         @db_config = db_config
         @root = root
+        @env = env
       end
 
       def create
@@ -20,7 +21,7 @@ module ActiveRecord
         file = File.absolute_path?(db_path) ? db_path : File.join(root, db_path)
         FileUtils.rm(file)
         FileUtils.rm_f(["#{file}-shm", "#{file}-wal"])
-        FileUtils.touch("tmp/restart.txt") if Rails.env.development?
+        FileUtils.touch("tmp/restart.txt") if @env&.development?
       rescue Errno::ENOENT => error
         raise NoDatabaseError.new(error.message)
       end
--- a/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
@@ -20,6 +20,7 @@ module ActiveRecord
         file = File.absolute_path?(db_path) ? db_path : File.join(root, db_path)
         FileUtils.rm(file)
         FileUtils.rm_f(["#{file}-shm", "#{file}-wal"])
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
```diff
@@ -20,6 +20,7 @@ def drop
         file = File.absolute_path?(db_path) ? db_path : File.join(root, db_path)
         FileUtils.rm(file)
         FileUtils.rm_f(["#{file}-shm", "#{file}-wal"])
+        FileUtils.touch("tmp/restart.txt") if Rails.env.development?
```
eileencodes: We can't call Rails from within Active Record as it's not a dependency of this component (and since the "contract" is AR can be used independently from the rest of the framework it must maintain that contract). The exceptions are codepaths in AR that won't work without Rails.

I think this is ultimately a limitation of SQLite3 since it's a file, the restart is so that file is reloaded. Realistically, how often are you deleting the db? I don't keep dev dbs around for long either but it's not a weekly occurrence that I start from scratch. More like monthly. I'm not sure this is a common enough use case to solve on the framework level.

eileencodes (changes requested):

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
yokonao: Thanks! I has removed direct dependency on Rails objects. (I will fix the test later, but I will answer your question first.)

Regarding the latter question, it's a case by case basis.
I also think that frequent database deletion is rare in the development of mature applications.

However, during the initial development of an application, especially when you are exploring the ideal database schema, you will want to reset the database frequently. I have also experienced use cases where you want to run db:reset and db:seed frequently in order to modify the development seed data.

Also, executing db:drop or db:reset while the development server is running is a natural action for developers. Shouldn't the framework provide some care to prevent inconsistencies in such cases?

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
```diff
@@ -20,6 +21,7 @@ def drop
         file = File.absolute_path?(db_path) ? db_path : File.join(root, db_path)
         FileUtils.rm(file)
         FileUtils.rm_f(["#{file}-shm", "#{file}-wal"])
+        FileUtils.touch("tmp/restart.txt") if @env == "development"
```
seuros, line 24: the restart.txt is not the correct way to restart the server.

It a currently a Puma only technic... And since version 5+, it an opt-in plugin.

seuros (changes requested):

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
eileencodes: > Also, executing db:drop or db:reset while the development server is running is a natural action for developers.

Is it though? Anytime I've done work that blows away the database I've gotten tons of complaints about devs wanting to keep their local data around. I'm not saying you don't do that, but I don't think it's as common as you feel it is. Usually when I'm blowing away the it's right after a pull of main so the server isn't running yet.

I don't see a clean way to do this since as @seuros noted this isn't the right way to restart either.

yokonao: Admittedly, my use case may not be typical.

And it's important to note that there's no clean way to restart the server. I'll probably have to give up on this approach.

activerecord/lib/active_record/tasks/sqlite_database_tasks.rb
yokonao: Thanks for teaching me！

Issue #39566, opened 2020-06-07
Add top-level Rails.application.url for canonical reference


dhh:
We often need to know the application's URL outside of the request/response cycle. This can be in jobs or in mailers or for configuration purposes. Let's extract this concept as `Rails.application.url`, which is configured by default in the environment files.

What we'd like is for config/development.rb to have something like:

```ruby
  # Application URL
  config.url = "https://hey.test"
```

Which is then turned into a `ActionDispatch::Http::URI` behind the scenes, and made available as `Rails.application.url`. This URL should automatically be configured as the default URL for mailers and renderers.

Here's a quick stab at an extraction that take the shared logic from ActionDispatch::Http::URL and turns into a decoration on top of stdlib's `URI`. The ActionDispatch::Http::URL is a better name, but it's unfortunately setup as a module that's included in ActionDispatch::Http::Request. It's possible that we could just rename that, though, and then reuse this name instead of URI:

```ruby
class ActionDispatch::Http::URI
  mattr_accessor :tld_length, default: 1

  delegate :scheme, :host, :port, :path, :query, :to_s, to: :uri
  delegate :scheme=, :host=, :port=, :path=, :query=, to: :uri

  class << self
    # Returns the domain part of a host given the domain level.
    #
    #    # Top-level domain example
    #    extract_domain('www.example.com', 1) # => "example.com"
    #    # Second-level domain example
    #    extract_domain('dev.www.example.co.uk', 2) # => "example.co.uk"
    def extract_domain(host, tld_length)
      extract_domain_from(host, tld_length) if named_host?(host)
    end

    # Returns the subdomains of a host as an Array given the domain level.
    #
    #    # Top-level domain example
    #    extract_subdomains('www.example.com', 1) # => ["www"]
    #    # Second-level domain example
    #    extract_subdomains('dev.www.example.co.uk', 2) # => ["dev", "www"]
    def extract_subdomains(host, tld_length)
      if named_host?(host)
        extract_subdomains_from(host, tld_length)
      else
        []
      end
    end

    # Returns the subdomains of a host as a String given the domain level.
    #
    #    # Top-level domain example
    #    extract_subdomain('www.example.com', 1) # => "www"
    #    # Second-level domain example
    #    extract_subdomain('dev.www.example.co.uk', 2) # => "dev.www"
    def extract_subdomain(host, tld_length)
      extract_subdomains(host, tld_length).join(".")
    end

    private
      def extract_domain_from(host, tld_length)
[...]

PR #55876, opened 2025-10-08
add config.activestorage.default_url_options

agomezcampero:
### Motivation / Background

This Pull Request has been created because when using Disk service in development and something else in production you end up having to set ActiveStorage::Current.url_options conditionally. This usually leads to code in a controller that looks like this:

```ruby
ActiveStorage::Current.url_options = { host: , protocol: } if Rails.env.development?
```

I believe it would be cleaner to set this up in the configs and not have to add any code to the controller just to get ActiveStorage working in development.

### Detail

This Pull Request adds the ability to add config.active_storage.default_url_options in your config file to avoid having to set ActiveStorage::Current.url_options.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Add `config.active_storage.default_url_options` configuration option.
+
+    Allows setting default URL options for Active Storage disk service at the application level.
+    This provides a fallback when `ActiveStorage::Current.url_options` is not set.
+
+    ```ruby
+    # config/application.rb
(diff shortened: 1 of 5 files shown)


rafaelfranca:
Related https://github.com/rails/rails/issues/39566

agomezcampero:
> Related #39566

Should I just close this PR then? I see some people are already tackling that issue and aims to be shipped in 8.2

This was a quick easy fix, but I don't think I have the time or the experience in the Rails codebase to tackle it myself.

PR #55879, opened 2025-10-09
Add :restart option to retry_job and retry_on

lHydra:
### Motivation / Background

Adds a new `:restart` option to `retry_job` and `retry_on` for jobs
including `ActiveJob::Continuable`. By default, retried jobs resume from the last
saved checkpoint. In some cases, certain exceptions require restarting the job
from the beginning instead of resuming partially completed work.

This feature allows developers to selectively reset the job progress when needed,
while preserving the default behavior for normal continuable jobs.

It also ensures that internal continuation interrupts (`Continuation::Interrupt`)
still always resume from the last checkpoint, keeping the built-in semantics of
`ActiveJob::Continuable` unchanged.

### Detail

* Add `restart` option to `retry_job`:
  `restart: false` → resume from last saved checkpoint (default)  
  `restart: true` → discard saved progress and start from the beginning
* Updated `retry_on` to accept and pass the `:restart` option to `retry_job`.
* Internal `Continuation::Interrupt` errors always force `restart: false` to maintain expected continuation flow and avoid resetting progress caused by normal pause or reschedule events.

Example usage:

```ruby
class MyJob < ApplicationJob

  self.resume_errors_after_advancing = false
  retry_on RestartError, restart: true
  retry_on StandardError

  def perform(*args)
    step :step1 do
      # some step logic
    end

    step :step2 do
      # some step 2 logic that may cause RestartError
    end
  end
end
```

The `:restart` option can also be defined at the job level through `resume_options`,
allowing the job to fully restart on any `StandardError`.

Example usage:

```ruby
class MyJob < ApplicationJob

  self.resume_options = { restart: true }

  def perform(*args)
    step :step1 do
      # some step logic
    end

    step :step2 do
      # some step 2 logic that may cause StandardError
    end
  end
end
```

### Additional information

Although I was thinking — it might make sense to name the option for
`resume_options` (and the corresponding `retry_job` method) something like
`:restart_on_exception`, to make the behavior more explicit,
while keeping `:restart` for the `retry_on` method. Do you think that would be a better choice?

### Checklist

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Add `:restart` option to `retry_job` and `retry_on` for jobs including `ActiveJob::Continuable`.
+
+    This option controls whether a retried job should resume from its last
+    saved checkpoint (`restart: false`, default) or restart from
+    the beginning (`restart: true`). It allows finer control over retry
+    behavior when certain exceptions require a full restart instead of resuming
+    partial progress.
+
+    *Georgiy Melnikov*
+
 *   Add structured events for Active Job:
     - `active_job.enqueued`
     - `active_job.bulk_enqueued`
--- a/activejob/lib/active_job/continuable.rb
+++ b/activejob/lib/active_job/continuable.rb
@@ -91,7 +91,9 @@ module ActiveJob
       def resume_job(exception) # :nodoc:
         executions_for(exception)
         if max_resumptions.nil? || resumptions < max_resumptions
-          retry_job(**self.resume_options)
+          opts = self.resume_options
+          opts[:restart] = false if exception.is_a?(Continuation::Interrupt)
+          retry_job(**opts)
         else
           raise Continuation::ResumeLimitError, "Job was resumed a maximum of #{max_resumptions} times"
         end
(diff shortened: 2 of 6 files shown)


lHydra:
@djmb Hi! Could you please review this PR when you have a chance? I’d like to know your opinion on whether this functionality makes sense. I’m addressing you as the author of this related feature (https://github.com/rails/rails/pull/55127)

djmb:
@lHydra - what's the situation where you would want to do this? If an error in step 2 means that you want to retry step 1, then I wonder if they should be a single step?

lHydra:
@djmb I have a current example where I upload SSH keys to a server and gather system facts. Sometimes the server becomes unreachable due to a timeout, and in those cases, I only want to retry the current step. However, there are scenarios where a server admin is making changes in parallel and overwrites the authorized_keys file. In that situation, the keys were initially uploaded, but by the time we reach the `:gather_facts` step, access is denied. In such a case, it would be very useful to restart the job from the beginning.

```ruby
class Servers::PrepareJob < ApplicationJob

  retry_on PermissionDeniedError, restart: true
  retry_on ConnectionTimeoutError

  self.resume_errors_after_advancing = false

  def perform(server_id)
    @server = Server.find(server_id)

    step :add_keys do
      Servers::AddKeys.call(@server)
    end

    step :install_dependencies do
      Servers::InstallDependencies.call(@server) # may produce ConnectionTimeoutError or PermissionDeniedError
    end

    step :gather_facts do
      Servers::GatherFacts.call(@server) # may produce ConnectionTimeoutError or PermissionDeniedError
    end
  end
end
```

I understand that this is a fairly specific case in my situation, but I think the new step structure makes it convenient to separate logical blocks. If we always assume that a step requiring a full restart must contain all related logic in a single step, it can sometimes break the logical clarity

PR #52297, opened 2024-07-09
Add setting for logging redirect source locations

dennispaagman:
### Motivation / Background

This introduces a new config setting: `action_dispatch.verbose_redirect_logs`. It logs the source location of redirects. It works similarly to `active_record.verbose_query_logs` and `active_job.verbose_enqueue_logs`.

Redirects are of course used a lot. They are often defined in different locations. For example in your authentication code, authorization code, fallbacks for resources that can't be found, and to redirect users to a different page after performing an action, possibly with multiple redirect options in one controller action.

I think it would be beneficial to know exactly which line of code triggered a redirect. 

### Detail

This adds a config setting that logs the source location of the redirect.

There are three main use cases, they would look like this:

1. Redirects in filters (these already have a "filter chain halted" log as well, which points you in the right direction a bit, but the exact code location can still be useful):

```
Redirected to http://localhost:3000/
↳ app/controllers/posts_controller.rb:76:in `check_something'
Filter chain halted as :redirect rendered or redirected
Completed 302 Found in 18ms (ActiveRecord: 0.1ms (1 query, 0 cached) | GC: 0.0ms)
```

2. Redirects in controller actions:

```
Redirected to http://localhost:3000/posts/7
↳ app/controllers/posts_controller.rb:32:in `block (2 levels) in create'
Completed 302 Found in 36ms (ActiveRecord: 1.3ms (1 query, 0 cached) | GC: 0.0ms)
```

3. Redirects in routes:

```
Started GET "/redirect" for ::1 at 2024-07-09 12:40:23 +0200
Redirected to http://localhost:3000/
↳ /Users/dennis/Code/rails-test-app/config/routes.rb:14
Completed 301 Moved Permanently in 9ms
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

---

There are a few points I'm not sure how to approach and like some guidance in:

[...]

PR #55889, opened 2025-10-12
Fix verbose redirect output and route source mapping

gmcgibbon:
### Motivation / Background

This Pull Request has been created because https://github.com/rails/rails/pull/52297 isn't working exactly the way I expected.

Previously, we would sometimes log nil when verbose logs are disabled, and dispatch route level source locations could not be found.

### Detail

This Pull Request changes mostly dispatch code to record some source location in redirect endpoints when needed. This allows this feature to work in environments other than development, if needed, instead of depending on the other route source locations feature.

This adds an additional parameter to `redirect.action_dispatch` which may not be desirable. We can instead hack the caller trace to include the source location or use route source locations. Open to feedback on this, but I prefer to keep things simple here with an event parameter. 

### Checklist

### DIFF
--- a/guides/source/8_1_release_notes.md
+++ b/guides/source/8_1_release_notes.md
@@ -96,6 +96,8 @@ Please refer to the [Changelog][action-pack] for detailed changes.
 
 ### Notable changes
 
+*   Redirects are now verbose in development for new Rails apps. To enable it in an existing app, add `config.action_dispatch.verbose_redirect_logs = true` to your `config/development.rb` file.
+
 Action View
 -----------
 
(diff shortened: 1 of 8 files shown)


guides/source/8_1_release_notes.md
```diff
@@ -96,6 +96,8 @@ Please refer to the [Changelog][action-pack] for detailed changes.
 
 ### Notable changes
 
+*   Redirects are now verbose in development for new Rails apps. To enable it in an existing app, add `config.action_dispatch.verbose_redirect_logs = true` to your `config/development.rb` file.
```
chaadow, line 99: Thanks!

PR #54468, opened 2025-02-07
Make ActiveStorage::Service responsible for checksums

mrpasquini:
### Motivation / Background

This PR moves responsibility for providing the class used to calculate checksums to the ActiveStorage::Service class. This makes the storage service class fully responsible for calculating digests on the back end and removes the need to expose the digest class or ActiveStorage.checksum_implementation in other classes.

### Detail
S3, GCS, and Azure all support file integrity checksums in addition to MD5. MD5 is not the recommended algorithm by most of these cloud services and cannot be used in a FIPS compliant environment. This is a first step toward allowing these service classes to support additional/alternative algorithms for file integrity.

This PR:
- removes the private implementation of `calculate_checksum_in_chunks` from Blob in favor a new public method `calculate_checksum` on the storage service class to manage all digest calculation for that service. 
- ~removes direct use of `ActiveStorage.checksum_implementation` by other class/services when verifying files in favor of `calculate_checksum`.~ Removes direct use of OpenSSL::Digest::MD5 by services in favor of a new public `checksum_implementation` method on the base service class. This method returning `OpenSSL::Digest::MD5` by default to maintain backwards compatibility.
- removes the separate ActiveStorage::Downloader class in favor of a combined private `download_and_verify_tempfile` implementation in the base service.
- refactors direct uploads to separate algorithms from framework functionality.

This PR also implements SHA256 support for the S3 service including support for direct uploads.

See #54123, #34443 or #54040 for additional discussion. 

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #54040, opened 2024-12-23
Add config option for MD5 class to support FIPS compliance

mrpasquini:
### Motivation / Background

This Pull Request has been created because `OpenSSL::Digest::MD5` is not available when OpenSSL is running in FIPS mode #34443

Fix: #34443

### Detail

This Pull Request abstracts usages of `OpenSSL::Digest::MD5` to `ActiveStorage.digest_class` which may be set by a new configuration option `config.active_storage.digest_class`. The default continues to be `OpenSSL::Digest::MD5`.

This allows an alternative implementation of MD5 to be used when it is not available in OpenSSL due to FIPS compliance mode.

### Additional information

A longer term term solution may be a refactoring of `ActiveStorage::Service::*Service` classes to support each digest type supported by each cloud service and a way to configure the digest to use per service or instance of service.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55890, opened 2025-10-12
Rename AS checksum_implementation to md5_checksum_implementation

mrpasquini:
### Motivation / Background

This Pull Request has been created because discussion here: https://github.com/rails/rails/pull/54468#discussion_r2424494487

checksum_implementation only supports alternative MD5 implementations for the default service and should reflect that.  

### Detail

This Pull Request renames the method used to access the MD5 implementation class from `checksum_implementation` to `md5_checksum_implementation`. Adding support for additional checksum implementations would at the service class level.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -75,7 +75,7 @@
 
     *Sean Doyle*
 
-*   Add support for alternative MD5 implementation through `config.active_storage.checksum_implementation`.
+*   Add support for alternative MD5 implementation through `config.active_storage.md5_checksum_implementation`.
 
     Also automatically degrade to using the slower `Digest::MD5` implementation if `OpenSSL::Digest::MD5`
     is found to be disabled because of OpenSSL FIPS mode.
--- a/activestorage/app/models/active_storage/blob.rb
+++ b/activestorage/app/models/active_storage/blob.rb
@@ -332,7 +332,7 @@ class ActiveStorage::Blob < ActiveStorage::Record
     def compute_checksum_in_chunks(io)
       raise ArgumentError, "io must be rewindable" unless io.respond_to?(:rewind)
 
-      ActiveStorage.checksum_implementation.new.tap do |checksum|
+      ActiveStorage.md5_checksum_implementation.new.tap do |checksum|
         read_buffer = "".b
         while io.read(5.megabytes, read_buffer)
           checksum << read_buffer
--- a/activestorage/lib/active_storage.rb
+++ b/activestorage/lib/active_storage.rb
@@ -362,13 +362,13 @@ module ActiveStorage
 
   mattr_accessor :track_variants, default: false
 
-  singleton_class.attr_accessor :checksum_implementation
-  @checksum_implementation = OpenSSL::Digest::MD5
+  singleton_class.attr_accessor :md5_checksum_implementation
+  @md5_checksum_implementation = OpenSSL::Digest::MD5
   begin
-    @checksum_implementation.hexdigest("test")
+    @md5_checksum_implementation.hexdigest("test")
   rescue # OpenSSL may have MD5 disabled
     require "digest/md5"
-    @checksum_implementation = Digest::MD5
+    @md5_checksum_implementation = Digest::MD5
   end
 
   mattr_accessor :video_preview_arguments, default: "-y -vframes 1 -f image2"
--- a/activestorage/lib/active_storage/downloader.rb
+++ b/activestorage/lib/active_storage/downloader.rb
@@ -35,7 +35,7 @@ module ActiveStorage
       end
 
       def verify_integrity_of(file, checksum:)
-        unless ActiveStorage.checksum_implementation.file(file).base64digest == checksum
+        unless ActiveStorage.md5_checksum_implementation.file(file).base64digest == checksum
           raise ActiveStorage::IntegrityError
         end
       end
--- a/activestorage/lib/active_storage/engine.rb
+++ b/activestorage/lib/active_storage/engine.rb
@@ -149,8 +149,8 @@ module ActiveStorage
         ActiveStorage.binary_content_type = app.config.active_storage.binary_content_type || "application/octet-stream"
         ActiveStorage.video_preview_arguments = app.config.active_storage.video_preview_arguments || "-y -vframes 1 -f image2"
         ActiveStorage.track_variants = app.config.active_storage.track_variants || false
-        if app.config.active_storage.checksum_implementation
-          ActiveStorage.checksum_implementation = app.config.active_storage.checksum_implementation
+        if app.config.active_storage.md5_checksum_implementation
+          ActiveStorage.md5_checksum_implementation = app.config.active_storage.md5_checksum_implementation
         end
       end
     end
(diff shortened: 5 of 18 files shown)


benngarcia:
The current `ActiveStorage.checksum_implementation` interface doesn’t appear to support arbitrary algorithms like `Digest::SHA2` at runtime. The naming suggests it is specific to MD5, especially since there’s also an `md5_checksum_implementation`. Based on Rails configuration conventions, that implies the ability to swap *implementations* of MD5 (for example, OpenSSL vs. Digest) rather than to choose a different hash algorithm entirely. From a developer’s perspective, this distinction isn’t intuitive or particularly useful.

After reviewing the linked PR, the design intent makes more sense if the goal is to support per-global, per-service-class, and per-service-instance `checksum_implementation`s. The confusing part is here:
[https://github.com/rails/rails/pull/54468/files/43005f16ebd8393c76197a9245cb7a4712dfca9f#diff-f64fd0fdc47fb391533ab11d1ea529e7ba6c28a5fbc905884869fc50cdcc626bR17](https://github.com/rails/rails/pull/54468/files/43005f16ebd8393c76197a9245cb7a4712dfca9f#diff-f64fd0fdc47fb391533ab11d1ea529e7ba6c28a5fbc905884869fc50cdcc626bR17)

That code assumes `:md5` will always be used for `ActiveStorage.checksum_implementation` and then adds special-case handling. Instead, the global, class-level, and instance-level initialization should share a single logic path that determines the actual digest implementation. The default value should come from `ActiveStorage.checksum_implementation`, similar to other ActiveStorage configurations.

Conceptually:

```ruby
def initialize(bucket:, upload: {}, public: false, digest_type: ActiveStorage.checksum_implementation, **options)
  ...
end

def checksum_implementation(check_digest_type: digest_type)
  return ActiveStorage.resolve_checksum_implementation(digest_type) if digest_type.is_a?(Symbol)
  digest_type
end

class ActiveStorage
  def self.resolve_checksum_implementation(type)
    case type
    when :md5
      OpenSSL::Digest::MD5 rescue Digest::MD5
    when :sha256
      OpenSSL::Digest::SHA256 rescue Digest::SHA256
    else
      raise ArgumentError, "Unsupported checksum type: #{type.inspect}"
    end
  end
end
```

This makes `ActiveStorage.checksum_implementation` globally consistent, intuitive, and extensible. You can use `:md5`, `:sha256`, or provide a direct digest class. It aligns with broader Rails configuration semantics and removes the need for special-casing MD5.

mrpasquini:
Not to drag all of the other discussion here, but why does ActiveStorage need to be aware of the different algorithms at all?
```
AWS: :MD5, :CRC32, :CRC32c, SHA1, :SHA256, :CRC64NVMe
Azure: :MD5, :CRC64,
GCS:  :MD5, :CRC32c,
```
MD5 is the only one supported across AWS/Azure/GCS. CRC32c overlaps AWS/GCS. Conceptually these should be features of the services and should be supported by the service classes, not at the global AS level.

Even with the above code, only the service classes should be calling `ActiveStorage.resolve_checksum_implementation`. Everything else would need to call the service's `checksum_implementation` method since that is where the service would implement support for its own custom algorithms.

Does it make sense to get rid of the configuration option but keep the global method (and i guess fallback?) for now? The ability to select (or provide a class for) an algorithm seems like it should be features added at the service class level. I would argue to keep `ActiveStorage.checksum_implementation` or `ActiveStorage.md5_checksum_implementation` because it improves the existing monkey patch situation by consolidating all calls to the MD5 class to a single method to be patched. Maybe additional PRs can move MD5 from the global AS space to the base service class?

benngarcia:
Yeah that's a great observation. A global variable for a service-level constant doesn't make sense. 

>  I would argue to keep ActiveStorage.checksum_implementation or ActiveStorage.md5_checksum_implementation because it improves the existing monkey patch situation by consolidating all calls to the MD5 class to a single method to be patched.

Agreed

I am of the opinion that keeping `ActiveStorage.checksum_implementation` in the interim makes more sense, as I still weigh the confusion of an `md5_checksum_implementation` (implying there should be a `sha256_checksum_implementation`, etc. as the worse of two not-ideal choices.

I think ideal would be defining the checksum_implementation as an ivar at initialization of the service, and then just a VALID_CHECKSUM_IMPLEMENTATIONS-like constant (example name) per service class, and then we can define the single  
`resolve_checksum_implementation` method in the base service to check the classes constant for presence & do resolution as defined above.

This would raise errors on initialization when an invalid checksum algorithm is passed, and you can always override the method in your custom service to support a different service.

mrpasquini:
> I am of the opinion that keeping `ActiveStorage.checksum_implementation` in the interim makes more sense, as I still weigh the confusion of an `md5_checksum_implementation` (implying there should be a `sha256_checksum_implementation`, etc. as the worse of two not-ideal choices.

I'm not sure I agree, but I also dont feel super strongly on this. I just did not want to include requested changes around  `ActiveStorage.checksum_implementation` in the same PR as moving responsibility of calculating and using checksums to the service classes in https://github.com/rails/rails/pull/54468 

I am trying to keep this into separate, digestible PRs so if we want to rename this method (as requested in #54468), remove the configuration option, or maybe even wrap all calls to `ActiveStorage.checksum_implementation` calls through a method in the base service class (about a +/-40 line change, mostly tests) I would like to do that here.

> I think ideal would be defining the checksum_implementation as an ivar at initialization of the service, and then just a VALID_CHECKSUM_IMPLEMENTATIONS-like constant (example name) per service class, and then we can define the single `resolve_checksum_implementation` method in the base service to check the classes constant for presence & do resolution as defined above.
> 
> This would raise errors on initialization when an invalid checksum algorithm is passed, and you can always override the method in your custom service to support a different service.

I broadly agree with having a config at the service level for the default algorithm to use and which algorithms are allowed. What that exactly looks like I would probably save for a later PR, but I'm not sure we'd want to force service classes into a specific implementation of multiple types or even configuration in a specific way. Ideally the checksum attribute is left wholly to the service class to manage and the configuration of the service is also left to the service.
A single `resolve_checksum_implementation` may not even work in cases like S3 Multipart upload support being added where the supported list of algorithms on the AWS side differs and could possibly even need another set of configuration options in the S3 service: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#MultipartUploads-Checksums

I see this currently shaping up as
- This PR to address the MD5 functionality that was already merged (lets discuss what that means)
- #54468 to move responsibility of creating and validating checksums to the service + meet rails maintainer's request to show an implementation in a service class
- A PR to fully add support to use SHA256 for S3 and generically to direct upload (+ S3 service configuration options, etc)

rafaelfranca (changes requested):
We can't rename things without keeping backward compatibility and deprecating the old name.

mrpasquini:
> We can't rename things without keeping backward compatibility and deprecating the old name.

Its in https://github.com/rails/rails/releases/tag/v8.1.0.beta1 release, would it still require deprecation if renamed/removed before 8.1.0 is cut?

I dont feel strongly on the renaming, but I do think we should not introduce the AS level configuration option. Would that be able to be remove before 8.1.0 is cut?

I guess the other option could be to revert completely: https://github.com/rails/rails/pull/54040/commits/bbc0cf1712da74d52874a4e012321b5e27e85e39

rafaelfranca:
Good point. I don't think accepting different MD5 checksum implementation is worth it. I'll just revert that one. We should support changing the checkum algorithm entirely, not only accepting MD5

PR #55891, opened 2025-10-12
Update comment for ActiveStorage::VariantWithRecord

connorshea:
### Motivation / Background

This Pull Request has been created because I found the phrasing of the documentation a bit misleading, as it implies the default behavior is not the default behavior ("this is _only_ used if <default setting since Rails 6.1> is enabled.").

### Detail

This Pull Request changes the docs to clarify the usage of `ActiveStorage::VariantRecord`, since `track_variants` has been the default for a few years now.

### Checklist

### DIFF
--- a/activestorage/app/models/active_storage/variant_with_record.rb
+++ b/activestorage/app/models/active_storage/variant_with_record.rb
@@ -3,8 +3,7 @@
 # = Active Storage \Variant With Record
 #
 # Like an ActiveStorage::Variant, but keeps detail about the variant in the database as an
-# ActiveStorage::VariantRecord. This is used if +ActiveStorage.track_variants+ is enabled,
-# which has been the default since Rails 6.1.
+# ActiveStorage::VariantRecord. This is used if +ActiveStorage.track_variants+ is enabled.
 class ActiveStorage::VariantWithRecord
   include ActiveStorage::Blob::Servable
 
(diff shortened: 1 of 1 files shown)


activestorage/app/models/active_storage/variant_with_record.rb
```diff
@@ -3,7 +3,8 @@
 # = Active Storage \Variant With Record
 #
 # Like an ActiveStorage::Variant, but keeps detail about the variant in the database as an
-# ActiveStorage::VariantRecord. This is only used if +ActiveStorage.track_variants+ is enabled.
+# ActiveStorage::VariantRecord. This is used if +ActiveStorage.track_variants+ is enabled,
+# which has been the default since Rails 6.1.
```
zzak: We generally don't document which version things are introduced or changed in the API docs, and leave that to the upgrade guide.

connorshea: Should I instead phrase it as "if track_variants is enabled, which is the default."?

rafaelfranca: `This is used if +ActiveStorage.track_variants+ is enabled.` is enough. When it is updated is up to the configuration guide.

connorshea: Updated the phrasing to Rafael's suggestion 👍

PR #55896, opened 2025-10-13
Improve support for SQLite database URIs

flavorjones:
### Motivation / Background

A SQLite database may be configured with a relative path, an absolute path, or a nonstandard `file:` URIs (possibly with params). When it is a URI, it's non-trivial to determine the filesystem path to the database, and Rails has historically not handled them very well.

Specifically, when using a SQLite database URI:

- `db:create` does not correctly check for file existence
- `db:drop` does not work at all
- opening a database connection to a nonexistent file does not implicitly create the parent directory

All of these things work fine if a filesystem path is used, and so I want this behavior when I use URIs.

### Detail

This commit introduces a new singleton method on the connection adapter, `SQLite3Adapter.resolve_path`, to reliably resolve a filesystem path from the database configuration.

It uses this method in the connection adapter's initializer to ensure the database file's directory exists. Previously the adapter omitted this check for database URIs.

It also introduces usage of this method in the `create` and `drop` database tasks which also previously did not fully support database URIs.

### More information

You can read up on SQLite URIs here: https://www.sqlite.org/uri.html

They are, unfortunately, nonstandard around the edges and so we're unable to "just use" Ruby's `URI` class.

This code has been extracted from the [`activerecord-tenanted` gem](https://github.com/basecamp/activerecord-tenanted) where it's been working well for us for many months.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,2 +1,8 @@
+*   Improve support for SQLite database URIs.
+
+    The `db:create` and `db:drop` tasks now correctly handle SQLite database URIs, and the
+    SQLite3Adapter will create the parent directory if it does not exist.
+
+    *Mike Dalessio*
 
 Please check [8-1-stable](https://github.com/rails/rails/blob/8-1-stable/activerecord/CHANGELOG.md) for previous changes.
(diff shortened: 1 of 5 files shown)


flavorjones:
Looking into the railtie failures.

flavorjones:
Ready for review now. I think the test failure is an unrelated flake.

guilleiguaran (approved):

guilleiguaran:
Great work @flavorjones !!

PR #55897, opened 2025-10-13
Introduce Parameter Object: QueryIntent

matthewd:
This has been on my todo list for a while, as we've continued to gradually accumulate extra context that we end up needing to pass along with queries.

~~**This is a draft** -- for exploration, I've changed a number of low-level-but-public APIs. Even if we agree that's a reasonable end-goal, they would require deprecations along the way.~~

Affected Public APIs:

`exec_update` and `exec_delete` are deprecated, pointing callers to `update` and `delete` instead. Those methods (the only internal callers) now just `affected_rows(raw_execute(intent))` directly.

`exec_insert` is likewise deprecated in favour of `insert`. We [currently?] still need to keep it for overriding, so the internal intent-using one is now `_exec_insert`. That's ugly, but I'm hopeful we can later evolve it away, and failing that we'll be able to reclaim `exec_insert` post-deprecation.

`log` is now called exclusively with an `intent`, but still accepts `sql, name, binds` calls too (with deprecation). In practice I don't anticipate many people calling it with existing parameters, but it is technically documented (as a `:doc:` private method). I anticipate most external users of `log` are overriding/wrapping it, so will be broken by the signature change on our existing calls -- but I think that's unavoidable.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters.rb
+++ b/activerecord/lib/active_record/connection_adapters.rb
@@ -76,6 +76,7 @@ module ActiveRecord
     autoload :Column
     autoload :PoolConfig
     autoload :PoolManager
+    autoload :QueryIntent
     autoload :SchemaCache
     autoload :BoundSchemaReflection, "active_record/connection_adapters/schema_cache"
     autoload :SchemaReflection, "active_record/connection_adapters/schema_cache"
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -565,21 +565,12 @@ module ActiveRecord
 
         # Lowest level way to execute a query. Doesn't check for illegal writes, doesn't annotate queries, yields a native result object.
         def raw_execute(intent)
-          sql = intent.sql
-          name = intent.name
-          binds = intent.binds
-          prepare = intent.prepare
-          async = intent.async
-          allow_retry = intent.allow_retry
-          materialize_transactions = intent.materialize_transactions
-          batch = intent.batch
-
-          intent.type_casted_binds = type_casted_binds = type_casted_binds(binds)
-          log(sql, name, binds, type_casted_binds, async: async, allow_retry: allow_retry) do |notification_payload|
+          intent.type_casted_binds = type_casted_binds(intent.binds)
+          log(intent) do |notification_payload|
             intent.notification_payload = notification_payload
-            with_raw_connection(allow_retry: allow_retry, materialize_transactions: materialize_transactions) do |conn|
-              result = perform_query(conn, sql, binds, type_casted_binds, prepare: prepare, notification_payload: notification_payload, batch: batch)
-              handle_warnings(result, sql)
+            with_raw_connection(allow_retry: intent.allow_retry, materialize_transactions: intent.materialize_transactions) do |conn|
+              result = perform_query(conn, intent.sql, intent.binds, intent.type_casted_binds, prepare: intent.prepare, notification_payload: intent.notification_payload, batch: intent.batch)
+              handle_warnings(result, intent.sql)
               result
             end
           end
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -546,8 +546,8 @@ module ActiveRecord
       end
 
       # Same as raw_execute but returns an ActiveRecord::Result object.
-      def raw_exec_query(...) # :nodoc:
-        cast_result(raw_execute(...))
+      def raw_exec_query(intent) # :nodoc:
+        cast_result(raw_execute(intent))
       end
 
       # Execute a query and returns an ActiveRecord::Result
@@ -564,7 +564,16 @@ module ActiveRecord
         private_constant :DEFAULT_INSERT_VALUE
 
         # Lowest level way to execute a query. Doesn't check for illegal writes, doesn't annotate queries, yields a native result object.
-        def raw_execute(sql, name = nil, binds = [], prepare: false, async: false, allow_retry: false, materialize_transactions: true, batch: false)
+        def raw_execute(intent)
+          sql = intent.sql
+          name = intent.name
+          binds = intent.binds
+          prepare = intent.prepare
+          async = intent.async
+          allow_retry = intent.allow_retry
+          materialize_transactions = intent.materialize_transactions
+          batch = intent.batch
+
           type_casted_binds = type_casted_binds(binds)
           log(sql, name, binds, type_casted_binds, async: async, allow_retry: allow_retry) do |notification_payload|
             with_raw_connection(allow_retry: allow_retry, materialize_transactions: materialize_transactions) do |conn|
@@ -610,12 +619,31 @@ module ActiveRecord
         # Same as #internal_exec_query, but yields a native adapter result
         def internal_execute(sql, name = "SQL", binds = [], prepare: false, async: false, allow_retry: false, materialize_transactions: true, &block)
           sql = preprocess_query(sql)
-          raw_execute(sql, name, binds, prepare: prepare, async: async, allow_retry: allow_retry, materialize_transactions: materialize_transactions, &block)
+          intent = QueryIntent.new(
+            sql: sql,
+            name: name,
+            binds: binds,
+            prepare: prepare,
+            async: async,
+            allow_retry: allow_retry,
+            materialize_transactions: materialize_transactions
+          )
+          raw_execute(intent, &block)
         end
 
         def execute_batch(statements, name = nil, **kwargs)
           statements.each do |statement|
-            raw_execute(statement, name, **kwargs)
+            intent = QueryIntent.new(
+              sql: statement,
+              name: name,
+              binds: kwargs[:binds] || [],
+              prepare: kwargs[:prepare] || false,
+              async: kwargs[:async] || false,
+              allow_retry: kwargs[:allow_retry] || false,
+              materialize_transactions: kwargs[:materialize_transactions] != false,
+              batch: kwargs[:batch] || false
+            )
+            raw_execute(intent)
           end
         end
 
@@ -691,13 +719,15 @@ module ActiveRecord
 
             # We make sure to run query transformers on the original thread
             sql = preprocess_query(sql)
-            future_result = async.new(
-              pool,
-              sql,
-              name,
-              binds,
+            intent = QueryIntent.new(
+              sql: sql,
+              name: name,
+              binds: binds,
               prepare: prepare,
+              async: true,
+              allow_retry: allow_retry
             )
+            future_result = async.new(pool, intent)
             if supports_concurrent_connections? && !current_transaction.joinable?
               future_result.schedule!(ActiveRecord::Base.asynchronous_queries_session)
             else
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -569,14 +569,14 @@ module ActiveRecord
           log(intent) do |notification_payload|
             intent.notification_payload = notification_payload
             with_raw_connection(allow_retry: intent.allow_retry, materialize_transactions: intent.materialize_transactions) do |conn|
-              result = perform_query(conn, intent.sql, intent.binds, intent.type_casted_binds, prepare: intent.prepare, notification_payload: intent.notification_payload, batch: intent.batch)
+              result = perform_query(conn, intent)
               handle_warnings(result, intent.sql)
               result
             end
           end
         end
 
-        def perform_query(raw_connection, sql, binds, type_casted_binds, prepare:, notification_payload:, batch:)
+        def perform_query(raw_connection, intent)
           raise NotImplementedError
         end
 
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -51,6 +51,15 @@ module ActiveRecord
       end
       private :to_sql_and_binds
 
+      # Compiles Arel in the given QueryIntent if needed
+      def compile_arel_in_intent(intent) # :nodoc:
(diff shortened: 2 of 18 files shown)


activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -0,0 +1,68 @@
...
+          batch: batch,
+          type_casted_binds: type_casted_binds,
+          notification_payload: notification_payload
+        }
+      end
+
+      # Returns a string representation showing key attributes
+      def inspect
+        "#<#{self.class.name} name=#{name.inspect} allow_retry=#{allow_retry} materialize_transactions=#{materialize_transactions}>"
+      end
```
byroot, line 65: We have `#instance_variables_to_inspect` now: https://github.com/ruby/ruby/pull/13555

activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    class QueryIntent # :nodoc:
```
byroot, line 5: Why not just `Query`?

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -564,18 +591,24 @@ def default_insert_value(column) # :nodoc:
...
-            with_raw_connection(allow_retry: allow_retry, materialize_transactions: materialize_transactions) do |conn|
-              result = perform_query(conn, sql, binds, type_casted_binds, prepare: prepare, notification_payload: notification_payload, batch: batch)
-              handle_warnings(result, sql)
+        def raw_execute(intent)
+          # Handle Arel compilation if needed
+          compile_arel_in_intent(intent)
+
+          # Handle SQL preprocessing if needed
+          intent.processed_sql ||= preprocess_query(intent.raw_sql) if intent.raw_sql
+          intent.type_casted_binds = type_casted_binds(intent.binds)
```
byroot, line 607: This feels like it should be part of the constructor, or at least of `compile_arel_in_intent`?

I see there are other codepath only calling one of the two, or none of them but I don't understand why.

activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -0,0 +1,68 @@
...
+      attr_reader :arel, :name, :prepare, :allow_retry,
+                  :materialize_transactions, :batch
+      attr_accessor :raw_sql, :binds, :async, :processed_sql, :type_casted_binds, :notification_payload
+
+      def initialize(arel: nil, raw_sql: nil, processed_sql: nil, name: "SQL", binds: [], prepare: false, async: false,
+                     allow_retry: false, materialize_transactions: true, batch: false)
```
byroot, line 11: Another thing I'd like to store in there is whether we know if it's a read or write query (or unknown).

e.g. you used `#select_all`, we know it's a read query and skip trying to match it against the shitty regexps.

byroot (commented):
I had something like that in mind for a while, but I really think it should just be a "query object".

activerecord/lib/active_record/connection_adapters/query_intent.rb
matthewd: I wanted a name that 1) sounded like an internal implementation detail, 2) conveyed a sense that it's a particular execution of a query and not a more long-term storable thing, 3) can be used as a parameter/variable name without introducing ambiguity about whether it contains an instance of this class vs a string vs an Arel object.

(I also anticipate this overall concept eventually needing more than one class -- e.g. a deeply-frozen encapsulation of just the things that actually need to be passed into the DB driver -- so it feels cleaner to decorate everything with slight nonsense-words to keep them mentally separated.)

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
matthewd: Yeah, I wanted to retain the current behaviour everywhere for now (at the cost of more complicated handling of arguably the most important contents in the object), to put off which call sites' local semantics it's safe to change.

I think `insert` -> `exec_insert` -> `sql_for_insert` is the call path that most clearly [currently] leans on all these steps being separately addressable.

activerecord/lib/active_record/connection_adapters/query_intent.rb
byroot: > sounded like an internal implementation detail

I suspected as much, but I don't think we should make our code less expressive to scare users away from private APIs.

> conveyed a sense that it's a particular execution of a query and not a more long-term storable thing

I don't think `Query` implies that but alright.

> can be used as a parameter/variable name without introducing ambiguity about whether it contains an instance of this class vs a string vs an Arel object.

If anything, I'm more weirded out by all these `intent` variable, more than by the class name.

If you really don't want `Query`, what about `QueryInfo`?

n-rodriguez: Or QueryParams?

benngarcia: fwiw I think of a `QueryIntent` (or any Intent suffixed object) as implying a writeable-until-maybe-execution and read-only-post-execution lifecycle. Not sure if others interpret this semantic the same way, and I haven't read the changes to evaluate if I agree with the name, just offering a data point.

byroot: > sounded like an internal implementation detail

I'm actually wondering if this should be public API.

I was talking with @beauraF earlier today, and they're doing some pretty heavy SQL parsing to figure out what they can redirect to read replicas or not. I think having access to the Arel AST could potentially save them a lot of effort.

Similarly, having the info of whether the query is known to be a read or a write, can be very useful, etc.

The more I think about it, the more it make sense to expose a proper query object rather than just a SQL string and a few things on the side.

NB: I understand designing a public API may require more thoughts and effort than a purely internal refactor. I'm mostly just thinking out loud.

beauraF: Indeed, we rely on this concept of `Database::CurrentQuery` in our codebase, which is set using [pg_query](https://github.com/pganalyze/pg_query):

```ruby
class Database::CurrentQuery < ActiveSupport::CurrentAttributes
  attribute :normalized_sql
  attribute :raw_sql
  attribute :query_id
  attribute :parsed_query
  attribute :type
  attribute :locking
  attribute :before_table_write
  attribute :past_parsed_queries
  attribute :past_written_tables
  attribute :auto_offloadable
  attribute :sql_length
  attribute :where_in_metadata

  # ...
end
```

We primarily use it to automatically route read queries to replicas, through logic like this:

```ruby
    def auto_offloadable?(sql, name)
      Database::CurrentQuery.set(sql)

      # ...

      # Don't offload if we are not reading the database
      return false if Database::CurrentQuery.type != :read

      # Don't offload if we have previously written to one of these tables
      return false unless Database::CurrentQuery.before_table_write

       # ...

      true
    end
```

We patch both the `PostgreSQLAdapter` and the `ConnectionPool` to make this behavior transparent.
We also use it for observability and resiliency use cases (eg: discussions around circuit breaker based on `query_id`). 

As @byroot mentioned, exposing a public API for query introspection would be a strong enabler for us. It would likely allow us to simplify our internal implementation and potentially open the door to upstreaming some of this logic to improve the current automatic role switching mechanism 😄

matthewd:
@byroot I'm thinking of moving forward with the current name for now, with an intention (heh) of revisiting the question later, when we can see what [if anything] further I've ended up moving onto it, etc. To that end, I think we're best served by a name that stands out for easy search & [interactive] replace, vs something like "query" which would be hard to track down.

byroot:
@matthewd yes no worries, my concern isn't a blocking one, your refactor seem positive to me regardless.

rafaelfranca (approved):

matthewd:
Mostly just for my own future reference / comparison, adapter methods that accept and run a query:

Before:

mermaid-diagram-2025-10-29-170613

Now:

mermaid-diagram-2025-10-29-175554

(Blue = public API, purple = nodoc, orange = private, blue-grey = deprecated, grey = adapter-external [calling non-API method], red = internal core; black arrow = passing Arel/SQL query, pink arrow = passing intent)

activerecord/lib/active_record/connection_adapters.rb
```diff
@@ -76,6 +76,7 @@ def resolve(adapter_name) # :nodoc:
     autoload :Column
     autoload :PoolConfig
     autoload :PoolManager
+    autoload :QueryIntent
```
byroot, line 79: I think this should be in `eager_autoload`.

activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
```diff
@@ -1197,23 +1197,53 @@ def translate_exception_class(native_error, sql, binds)
...
+              &block
+            )
+          end
         rescue ActiveRecord::StatementInvalid => ex
-          raise ex.set_query(sql, binds)
+          if intent
+            raise ex.set_query(intent.processed_sql, intent.binds)
+          else
+            raise ex.set_query(sql, binds)
+          end
```
byroot, line 1246: Nitpick: I'd move that rescue in each branch of the `if`, so we have a single conditional instead of two.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -51,6 +51,15 @@ def to_sql_and_binds(arel_or_sql_string, binds = [], preparable = nil, allow_ret
...
       private :to_sql_and_binds
 
+      # Compiles Arel in the given QueryIntent if needed
+      def compile_arel_in_intent(intent) # :nodoc:
+        return unless intent.needs_arel_compilation?
+
+        sql, binds, preparable, allow_retry = to_sql_and_binds(intent.arel, intent.binds, intent.prepare, intent.allow_retry)
+        intent.set_compiled_result(raw_sql: sql, binds: binds, prepare: preparable, allow_retry: allow_retry)
+      end
+      private :compile_arel_in_intent
```
byroot, line 61: I believe this logic should move inside `QueryAttempt`, e.g. `intent.compile(self)`. It would be better encapsulation.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -158,21 +169,40 @@ def exec_query(sql, name = "SQL", binds = [], prepare: false)
...
+
+        raw_exec_query(intent)
       end
 
       # Executes delete +sql+ statement in the context of this connection using
       # +binds+ as the bind substitutes. +name+ is logged along with
       # the executed +sql+ statement.
       def exec_delete(sql, name = nil, binds = [])
+        ActiveRecord.deprecator.warn("exec_delete is deprecated and will be removed in Rails 8.2. " \
+                                     "Please use #delete instead.")
```
byroot: Nitpick: You could use `deprecate_methods` helper for all of these.

byroot (approved):
It's big so I haven't reviewed everything in close detail, but seems like a positive refactor to me, we can always refine later.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
matthewd: For now I prefer to keep the adapter being obsessed with intent's internals rather than teaching the intent about `to_sql_and_binds`. I think it would be interesting to consider moving all of `to_sql_and_binds` into QueryIntent, but am also conscious that would mean constructing an intent just to do `to_sql`, so I'm holding off on that at this stage.

activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
matthewd: I would if the conditional wasn't on a local, but between the extra nesting of the `begin` and the possibility of two `rescue` blocks drifting.. 🤷🏻‍♂️

activerecord/lib/active_record/connection_adapters.rb
matthewd: Probably true, but IMO better considered together with a bunch of these peers -- I think AbstractAdapter has specific reasons it's currently eager loaded even though nothing else is, and I wouldn't want to imply this is a similar outlier.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -564,18 +598,24 @@ def default_insert_value(column) # :nodoc:
...
         # Lowest level way to execute a query. Doesn't check for illegal writes, doesn't annotate queries, yields a native result object.
-        def raw_execute(sql, name = nil, binds = [], prepare: false, async: false, allow_retry: false, materialize_transactions: true, batch: false)
-          type_casted_binds = type_casted_binds(binds)
-          log(sql, name, binds, type_casted_binds, async: async, allow_retry: allow_retry) do |notification_payload|
-            with_raw_connection(allow_retry: allow_retry, materialize_transactions: materialize_transactions) do |conn|
-              result = perform_query(conn, sql, binds, type_casted_binds, prepare: prepare, notification_payload: notification_payload, batch: batch)
-              handle_warnings(result, sql)
+        def raw_execute(intent)
+          # Handle Arel compilation if needed
+          compile_arel_in_intent(intent)
```
bquorning, line 603: I may be wrong, but is this an unintended API change? I think that `#raw_execute` did not previously accept Arel, but only raw sql? And it looks like you changed all the internal code to compile Arel into sql before calling `#raw_execute`.

matthewd: Intentional simplification.

Most public API already accepts at least Arel, and my plan is to drive towards full consistency. I haven't widened any documentation yet, because things are still in flux (and "surprise: it works" is a valid version of the undefined behaviour that's possible when supplying a parameter that's not documented as supported).

Also note at this stage this only applies where the QueryIntent is constructed with `arel:`, so does not affect e.g. `exec_update` which [via `internal_execute`] skips straight to `raw_sql:`. I'm semi-anticipating this changing (hence the above thinking/reasoning), but I'm actually not sure this PR alone does cause additional public APIs to accept Arel.

PR #55902, opened 2025-10-14
Fix ActiveRecord::Migration.check_all_pending! to use a separate connection class

engwan:
### Motivation / Background

GitLab started using `ActiveRecord::Migration.check_all_pending!` to check for pending migrations and we saw connection pool errors because the `ActiveRecord::Base` connection pool was being mutated when this check ran.

More details in https://gitlab.com/gitlab-org/gitlab/-/issues/573581

### Detail

This prevents `ActiveRecord::Migration.check_all_pending!` from mutating `ActiveRecord::Base` connection pools which can cause errors when other threads are also using the same connection.

I just made this method an alias of `check_pending_migrations` since they achieve the same thing but with `check_pending_migrations` using a separate `PendingMigrationConnection`. 

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Prevent connection pool errors when calling `ActiveRecord::Migration.check_all_pending!`
+
+    Use `ActiveRecord::PendingMigrationConnection` so that `ActiveRecord::Base` connection pools are
+    not affected.
+
+    *Heinrich Lee Yu*
+
 *   Fix SQLite3 data loss during table alterations with CASCADE foreign keys.
 
     When altering a table in SQLite3 that is referenced by child tables with
--- a/activerecord/lib/active_record/migration.rb
+++ b/activerecord/lib/active_record/migration.rb
@@ -688,24 +688,6 @@ module ActiveRecord
         delegate || superclass.nearest_delegate
       end
 
-      # Raises ActiveRecord::PendingMigrationError error if any migrations are pending
-      # for all database configurations in an environment.
-      def check_all_pending!
-        pending_migrations = []
-
-        ActiveRecord::Tasks::DatabaseTasks.with_temporary_pool_for_each(env: env) do |pool|
-          if pending = pool.migration_context.open.pending_migrations
-            pending_migrations << pending
-          end
-        end
-
-        migrations = pending_migrations.flatten
-
-        if migrations.any?
-          raise ActiveRecord::PendingMigrationError.new(pending_migrations: migrations)
-        end
-      end
-
       def load_schema_if_pending!
         if any_schema_needs_update?
           load_schema!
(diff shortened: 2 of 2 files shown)


engwan:
@eileencodes can you review this PR since you worked on https://github.com/rails/rails/commit/64f57c2c9d7b356ebd120751171a0fadf8946e90?

Thanks!

activerecord/lib/active_record/migration.rb
```diff
@@ -688,24 +688,6 @@ def nearest_delegate # :nodoc:
...
-            pending_migrations << pending
-          end
-        end
-
-        migrations = pending_migrations.flatten
-
-        if migrations.any?
-          raise ActiveRecord::PendingMigrationError.new(pending_migrations: migrations)
-        end
-      end
```
eileencodes, line 707: I'm not sure why I didn't fix this when I opened that PR if this fix is correct. What is calling `check_all_pending!` in your app? It's not called from Rails.

activerecord/lib/active_record/migration.rb
```diff
@@ -736,13 +718,16 @@ def disable_ddl_transaction!
         @disable_ddl_transaction = true
       end
 
+      # Raises ActiveRecord::PendingMigrationError error if any migrations are pending
+      # for all database configurations in an environment.
       def check_pending_migrations # :nodoc:
```
eileencodes: You need to remove the nodoc, otherwise the above docs won't show in the API docs.

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+*   Prevent connection pool errors when calling `ActiveRecord::Migration.check_all_pending!`
```
eileencodes, line 1: I think this is specific to your app and/or goodjob rather than all apps, so I wouldn't add a changelog. If we did add one it'd be more like "alias `check_all_pending!` to `check_pending_migrations`" since it's not fixing a bug everyone has, it's if you're calling `ActiveRecord::Base.connection` on boot while running this.

eileencodes (changes requested):

activerecord/lib/active_record/migration.rb
engwan: Removed.

activerecord/lib/active_record/migration.rb
engwan: Yeah, it's not called from Rails.

We use the `health_check` gem which [calls `check_pending!`](https://github.com/Purple-Devs/health_check/blob/master/lib/health_check/utils.rb#L37).

Since this is removed in Rails 7.2, I changed our code to call `check_all_pending!` since that's the suggested replacement.

As a workaround to the current bug, I just called `ActiveRecord::Migration.check_pending_migrations` which does the same thing without the connection problem and which is the method we're aliasing here.

activerecord/CHANGELOG.md
engwan: > it's if you're calling ActiveRecord::Base.connection on boot while running this

I don't think it's limited to this scenario. As long as `ActiveRecord::Migration.check_all_pending!` is called while another thread tries to use a connection, the problem happens.

In our case, we call `check_all_pending!` within a controller endpoint (`/health_check`). And we use Puma which is multi-threaded.

When a request comes in for `/health_check`, it alters the pool and when another request comes in at the same time for any other route, DB connection fails.

I know it's not a very common scenario to call `ActiveRecord::Migration.check_all_pending!` within a request so I'm also fine with just removing this changelog entry.

PR #55907, opened 2025-10-15
[#55866] Fix SQLite3 data loss during table alterations with CASCADE foreign keys.

### LINKED ISSUE #55866: Migrations that alter SQLite table structure can cause data loss
[A mere add_foreign_key can wipe out your whole Rails+SQLite production table](https://kyrylo.org/software/2025/09/27/a-mere-add-foreign-key-can-wipe-out-your-whole-rails-sqlite-production-table.html)

We likely should be disabling and re-enabling the `foreign_keys` PRAGMA before/after the transaction to change the table per the [sqlite docs](https://www.sqlite.org/lang_altertable.html#otheralter).

Edit: actually it looks like we already do that since https://github.com/rails/rails/commit/45881b0a64b4039e3b0d27e2c4d959dc2d744b9d

Edit2: Ah, this is complex. We currently do

```
transaction do
  disable_referential_integrity do
```

but it should be

```
disable_pragma_foreign_keys do
  transaction do
    enable_defer_foreign_keys do
```

because `PRAGMA foreign_keys`

> is a no-op within a transaction; foreign key constraint enforcement may only be enabled or disabled when there is no pending [BEGIN](https://www.sqlite.org/lang_transaction.html) or [SAVEPOINT](https://www.sqlite.org/lang_savepoint.html).

per [doc](https://www.sqlite.org/pragma.html#pragma_foreign_keys), and "`PRAGMA foreign_keys` before `TRANSACTION`" is the order recommended in the first sqlite doc link above. However, `PRAGMA defer_foreign_keys` is [set _per transaction_](https://www.sqlite.org/pragma.html#pragma_defer_foreign_keys) so it has to be inside the transaction block.

ruyrocha:
### Motivation / Background

This Pull Request has been created because #55866.

### Detail

Fix SQLite3 data loss during table alterations with CASCADE foreign keys

When altering a table in SQLite3 that is referenced by child tables with `ON DELETE CASCADE` foreign keys, ActiveRecord would silently delete all data from the child tables. This occurred because SQLite requires table recreation for schema changes, and during this process:

1. The original table is temporarily dropped
2. CASCADE deletes fire immediately on child tables
3. The table is recreated, but child data is already gone

This issue could result in catastrophic production data loss when adding columns, renaming tables, or adding foreign keys to tables that are referenced by other tables with CASCADE constraints.

The fix simply reverses the order to follow the [official SQLite 12-step ALTER TABLE procedure](https://www.sqlite.org/lang_altertable.html#otheralter).

This Pull Request fixes #55866.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
@@ -639,8 +639,8 @@ module ActiveRecord
             yield definition if block_given?
           end
 
-          transaction do
-            disable_referential_integrity do
+          disable_referential_integrity do
+            transaction do
               move_table(table_name, altered_table_name, options.merge(temporary: true))
               move_table(altered_table_name, table_name, &caller)
             end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Fix SQLite3 data loss during table alterations with CASCADE foreign keys.
+
+    When altering a table in SQLite3 that is referenced by child tables with
+    `ON DELETE CASCADE` foreign keys, ActiveRecord would silently delete all
+    data from the child tables. This occurred because SQLite requires table
+    recreation for schema changes, and during this process the original table
+    is temporarily dropped, triggering CASCADE deletes on child tables.
+
+    The root cause was incorrect ordering of operations. The original code
+    wrapped `disable_referential_integrity` inside a transaction, but
+    `PRAGMA foreign_keys` cannot be modified inside a transaction in SQLite -
+    attempting to do so simply has no effect. This meant foreign keys remained
+    enabled during table recreation, causing CASCADE deletes to fire.
+
+    The fix reverses the order to follow the official SQLite 12-step ALTER TABLE
+    procedure: `disable_referential_integrity` now wraps the transaction instead
+    of being wrapped by it. This ensures foreign keys are properly disabled
+    before the transaction starts and re-enabled after it commits, preventing
+    CASCADE deletes while maintaining data integrity through atomic transactions.
+
+    *Ruy Rocha*
+
 *   Fix negative scopes for enums to include records with `nil` values.
 
     *fatkodima*
--- a/activerecord/test/cases/adapters/sqlite3/sqlite3_adapter_test.rb
+++ b/activerecord/test/cases/adapters/sqlite3/sqlite3_adapter_test.rb
@@ -1185,6 +1185,132 @@ module ActiveRecord
         end
       end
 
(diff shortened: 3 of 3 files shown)


activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
```diff
@@ -639,11 +639,28 @@ def alter_table(
...
+
+            if fk_enabled
+              violations = execute("PRAGMA foreign_key_check").to_a
+              if violations.any?
+                raise ActiveRecord::StatementInvalid,
+                      "Foreign key violations detected after schema change: #{violations.inspect}"
+              end
+            end
+          rescue => _
+            execute("PRAGMA foreign_keys = ON") if fk_enabled
```
rafaelfranca: Would not this execute `execute("PRAGMA foreign_keys = ON")` again if there is any violation?

activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
```diff
@@ -639,11 +639,28 @@ def alter_table(
...
+
+            execute("PRAGMA foreign_keys = ON") if fk_enabled
+
+            if fk_enabled
+              violations = execute("PRAGMA foreign_key_check").to_a
+              if violations.any?
+                raise ActiveRecord::StatementInvalid,
+                      "Foreign key violations detected after schema change: #{violations.inspect}"
+              end
+            end
```
skipkayhil: I believe this is supposed to run inside the transaction according the the [sqlite docs](https://sqlite.org/lang_altertable.html) (steps 10-12)

skipkayhil: Thinking out loud: 

- I wonder if the real problem here is just the ordering of `disable_referential_integrity` and the `transaction`
- `defer_foreign_keys` inside `disable_referential_integrity` seems to be unnecessary as long as it wraps `transaction` and not the opposite
- fixture insertion also has `transaction` wrapping `disable_referential_integrity`, maybe could also be inverted?

activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
ruyrocha: It won't be required anymore, @rafaelfranca :)

activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb
ruyrocha: You're absolutely right, @skipkayhil, regarding the order - it's fixed now.

Regarding fixtures insertion, should we address this in the same PR or create a separate issue/PR for the fixture loading bug?

byroot:
Thank you. Backported to 8.1, 8.0 and 7.2

PR #55911, opened 2025-10-15
Feature Proposal: Enable Custom Validation Contexts

shetty-tejas:
### Motivation / Background

Currently, `ActiveModel::Validations` allows passing a validation context to `valid?` and `validate` methods, enabling dynamic validation behavior based on context. However, persistence methods like `create`, `update`, `find_or_create_by`, etc., do not support passing validation contexts, which limits their usefulness in certain workflows.

This is particularly problematic when using methods like `find_or_create_by`, which rely on database constraint violations rather than validation failures. In such cases, we may want to skip specific validations (like uniqueness checks) while still maintaining them for normal create/update operations.

### Detail

This Pull Request adds support for setting `validation_context` within blocks passed to ActiveRecord persistence methods such as `create`, `update`, `find_or_create_by`, `create_or_find_by`, and other similar methods.

**Example usage:**
```ruby
  validates :email, uniqueness: true, unless: -> { validation_context == :skip_validations }
end

# Set validation context during creation
User.create(name: 'Test User', email: 'test@example.com') do |user|
  user.validation_context = :skip_validations
end

# Set validation context with find_or_create_by
User.find_or_create_by(email: 'test@example.com') do |user|
  user.name = 'Test User'
  user.validation_context = :skip_validations
end
```

This allows validations to run normally in most cases while selectively skipping them in specific workflows where database constraints are the primary safeguard.

### Additional information

This change maintains backward compatibility as it only adds functionality when `validation_context` is explicitly set within the block. Existing code should continue to work without modification.

This PR also uses AI for writing tests and updating documentations. All AI generated code has been personally reviewed by me before committing.

### Checklist


rafaelfranca:
Makes sense to me, but of course, this PR needs more work to add tests and document the feature

benvilliere:
The provided examples are giving me a feeling that this may be achievable within the current Rails ergonomics. Exposing `validation_context` (part of the validation "guts") _feels_ off to me. 

I would suggest doing something like this instead:

```rb
  attr_accessor :skip_validations

  validates :email, uniqueness: true, unless: :skip_validations
end

User.create(name: 'Test User', email: 'test@example.com') do |user|
  user.skip_validations = true
end

User.find_or_create_by(email: 'test@example.com') do |user|
  user.name = 'Test User'
  user.skip_validations = true
end
```

If that's a pattern that you wind up using a lot in your codebase, I would consider isolating this into a concern.

shetty-tejas:
@rafaelfranca, yup, on it!

@benvilliere, while I see your point about using the `attr_accessor` pattern, I have some concerns with that approach:

1. If everything (including contexts) can be handled by defining attr_accessors, what's the purpose of having validation contexts in the first place? Since, they already exist as a feature, it makes sense to me to support them consistently.
2. If there's complex logic requiring multiple validation contexts, you'd need to create multiple `attr_accessors`, whereas using validation contexts lets you simply pass in the appropriate context. IMO, it keeps your logic cleaner and much more maintainable.
3. Exposing `validation_context` makes sense to me considering you can already read the current context from `#validation_context` (which is public at the moment). Being able to set it while validating (via `valid?`) but not while saving records seems incorrect. This creates a weird situation where you can validate a record with a specific context, but then can't actually save that record using the same context, as you'd have to resort to workarounds like `save(validate: false)`, which bypasses all validations rather than using the intended context-aware validation.

Basically, if validation contexts exist and work with `valid?`, they should work with persistence methods too, and this seems to be the easiest way to achieve without changing many things.

benvilliere:
Thank you for your reply @shetty-tejas. 

> [...] what's the purpose of having validation contexts in the first place? Since, they already exist as a feature, it makes sense to me to support them consistently.

Solid point! Given this context (pun intended), it makes more sense to surface `validation_context`. 

I've never met the kind of friction you're experiencing with your use case; I can understand how the `attr_accessor` wouldn't be a scalable solution for you. And it's fair to expect a framework feature to be thoroughly exposed. We want sharp knives!

viniciusoyama:
I get the motivation and I would like to discuss a few things and an alternative solution. 

A few thoughts:

**1) Making `validation_context` public allows us to do something like this:** 

```ruby
  user.validation_context = :step1
  user.save # using the context 
```

Which I dont think is a good thing. Contexts should work on a "per validation" basis. 

So - If `save` uses validations - we should pass the context to `save`. 

Which is exactly how it's implemented right now: https://guides.rubyonrails.org/active_record_validations.html#on)

So we CAN specify a context when saving, but not when using `create`.

**2) Using a block to change flow**

The intended use of a block in these methods (`create`, `update`, etc.) is to change a few attributes for persistance. 

Using it to affect validations or other flow related stuff can be a little bit confusing.

**3) When to use contexts?**

Related to your answer/question "If everything (including contexts) can be handled by defining attr_accessors, what's the purpose of having validation contexts in the first place?..."

Thinking about your specific example, I don't see it as a a context.

For me, contexts are related to a specific business use cases, workflows, scenarios, etc. "skip a validation" would be a intended effect of using a context but not a context itself. 

Your case is more like a flag to skip validations and using an accessor would be more appropriate in my opinion. 

We could very well have multiple validations on different contexts that still check for the skip_validation flag (this would be very confusing though 😅).

# Solution 

I agree that we should be able to specify a context for methods like `create`. But we can follow the `save` approach: if `create` uses validations, we should pass the context to `create`. 

So, an alternative solution would be adding a second Hash argument, like in the old times when we didn't have strong parameters: 

```ruby
User.create({ :first_name => 'Jamie', :is_admin => true }, :as => :admin)
```
https://apidock.com/rails/v3.1.0/ActiveRecord/Base/create/class

A potential implementation could be something like this:

```ruby
User.create({ first_name: 'Vini' }, context: :step_1)
```

What do you think?

shetty-tejas:
@viniciusoyama Interesting points! Let me respond to them one by one!

> Making validation_context public allows us to do something like this:
>   ```
>   user.validation_context = :step1
>   user.save # using the context 
>   ```
> Which I dont think is a good thing. Contexts should work on a "per validation" basis.

You're not wrong, this was a valid concern in my mind too. I was considering something like:
```rb
user.with_validation_context(:step1) do
  user.save
end
# or in short => user.with_validation_context(:step1, &:save)
```

which could scope the context to a particular block, but honestly I'm not sure if this adds much value over the existing `save(context: :step1)` for simple cases.

However, the real issue I'm trying to solve is with methods like `find_or_create_by` and `create_or_find_by`. How would the `context:` parameter work here?
```ruby
# Does this make sense?
User.find_or_create_by({ email: 'test@example.com' }, context: :skip_validations) do |user|
  user.name = 'Test User'
end
```

This creates ambiguity about when the context applies and how it interacts with the block. With the block-based approach, it's much clearer:
```ruby
User.find_or_create_by(email: 'test@example.com') do |user|
  user.name = 'Test User'
  user.validation_context = :skip_validations
end
```

> The intended use of a block in these methods (create, update, etc.) is to change a few attributes for persistence.
> Using it to affect validations or other flow related stuff can be a little bit confusing.

I disagree with your point that the intended use of a block is just to change attributes for persistence. The block is for changing the state of the object before it's written to the database, and `validation_context` is part of that state as it determines how the object gets validated before persistence. Setting it in the block is as explicit as setting any other attribute.

I would propose a better name rather than `validation_context`, but this was already existing as a private method, so I assume it follows Rails conventions.

Finally,

> Thinking about your specific example, I don't see it as a context.
> For me, contexts are related to a specific business use cases, workflows, scenarios, etc. "skip a validation" would be a intended effect of using a context but not a context itself.

I think you're generalizing the functionality of `validation_context` here. It's meant to be used specifically for validation (hence the name), and 'skipping a validation' is not an effect, but the actual use case of having a validation context. The only reason validations have `on:` or `except_on:` options is to determine whether we should run the validation or not, and these contexts aren't used in any other parts of the code (at least from what I've seen in the codebase).

So I wouldn't look at it as a generic context to be used for determining workflow, but as a very specific feature for validation itself.

That being said, I do see the value in your potential implementation with the `context:` parameter, but it might break what's already working. Maybe we could additionally support the `#with_validation_context` method I mentioned above?
```rb
user.with_validation_context(:some_context) do |u|
  u.save
end
# or in short => user.with_validation_context(:some_context, &:save)

# We can do anything with the AR object within the block
user.with_validation_context(:some_context) do |u|
  u.update!(some_attributes)
end

# While for creation we use the already existing block flavour
User.create(**attributes) do |u|
  u.validation_context = :some_context
end
```

What do you think?

viniciusoyama:
Hey, thanks for answering with so much detail! 

# On when to use the block + create

```
 I disagree with your point that the intended use of a block is just to change attributes for persistence....
```

I agree with your disagreement 😅. What I meant is that even though the block allows us to change any state, changing something because we intentionallly want to skip validations is possible but would be confusing (for me). But It's definitely a personal opinion. 

In the end it's more about keeping the usage consistent: if context are already per-validation, per-save, is there a way to keep this same pattern for other cases? 

There's also some degree "defensive programming" to avoid surprises. 

Currently, as it scoped per validation, when we do `save(context: :skip_validations)`, it will skip validations just for this save. 
If we call `save` or `valid?` it will run the validations again without the context. (user.validation_context is nil).

I can imagine a cases were we set `user.validation_context = :skip_validations`, do our things and pass the object along to other classes or methods that may call `valid?` or `save` or `create` again and the object would have the `validation_context` set. 

If the idea is to keep the behaviour of 'cleaning after using', allowing us to set `user.validation_context = :skip_validations` using a public method, would be strange too because, if it's a "state" that I intentionally want to set, I would not expect it to be cleaned after using. Example:

```ruby
user.validation_context = :skip_validations
user.save
user.validation_context # => nil
user.save # No validation_context??? This is odd (even if it's documented).
```

# Solution 

Your suggestion about `with_validation_context` is interesting because it was also my first idea haha. Something like:

`User.with_validation_context(:skip_validations).create({})`

But it would be more troublesome to implementand also it I think it's better to keep the consistency with `save`.

We don't need to break things: we can implement it as Rails does in a lot of other cases: if there's an aditional hash argument, we consider it as options. 

Does this make sense?

shetty-tejas:
@viniciusoyama Welcome back! 😄 Gonna go through it one by one again!

> I can imagine cases where we set user.validation_context = :skip_validations, do our things and pass the object along to other classes or methods that may call valid? or save or create again and the object would have the validation_context set.

You have a very fair point about defensive programming. The auto-cleanup behaviour (where `validation_context` may or may not be reset after save) does make it weird if we're exposing it as a settable attribute. That's why I recommended the `#with_validation_context` flavour in the previous response that makes it scoped with minimal effort.

> We don't need to break things: we can implement it as Rails does in a lot of other cases: if there's an additional hash argument, we consider it as options.

I won't lie, I do like this approach a lot (for consistency with `save(context:)`). However, there are **a lot** of methods in ActiveRecord that trigger validations - `create`, `create!`, `update`, `update!`, `find_or_create_by`, `find_or_create_by!`, `create_or_find_by`, `create_or_find_by!`, and potentially more. 

To keep the API consistent, we'd need to:
1. Add the `context:` parameter to all of these methods.
2. Ensure it's consistently implemented across all of them.
3. Maintain this pattern going forward - if someone adds a new creation/update method in the Rails codebase, they'd need to remember to support `context:` as well.

That's a significant amount of implementation work and ongoing maintenance burden, which was the main reason why I didn't want to dive deep into that syntax right now. With the block-based approach, it works across all these methods automatically without needing to modify each one individually, while being explicit enough (just a bit verbose 😄).

Also, as much as I hate the double-hash syntax (personal opinion lol), I definitely agree it's more consistent with the Rails conventions.

What do you think? Do you reckon the maintenance overhead of maintaining the `context` argument as a pattern makes sense for this feature?

viniciusoyama:
Hi! This is a tough one. I'm here because I faced the same issue and I asked myself: is `context:` only available for the `save` methods? 

Maybe this was forgotten or intentionally ignored as the effort to implement may not be worth it, given that `save` / `save!` covers all the cases (even the `create_or_find_by`, we just have to do the rescue/retry on duplicate record error by hand). 

Now we have a few options: 

1) Make `validation_context` public and use the block as you first suggested 
2) Make all other methods compatible. 
3) Simply do nothing and keep things as it is 

As you've mentioned: "... there are a lot of methods in ActiveRecord that trigger validations". This is basically a dependency problem, right? All these methods depends on validations. If that's the case, when validation interface changes, we "should" (not necessarily "have to") change these methods as well or create a workaround. 🥲

As a developer (using Rails), I would find solution 2 more appealing. As a developer (contributing to Rails), I would find solution 1 a good compromise. We just have to document these edge cases well (and I'm favoring for the value to be reset after validation, as I think it's safer). 

# We can go with 1 or ask Rafael for his opinion
I understand your points, just wanted to share my thoughts here. 

If you're also satisfied with solution 1, let just go for it. If you're interested in exploring solution 2, I think that should be good to hear a third opinion, because this is also a decision about how consistent we want to be with the Rails API. (So asking about Rafael's opinion may be good to get a broader view.)

shetty-tejas:
@viniciusoyama got back to this one and had a chance to think more. ActiveSupport has an Object#with method which can be used without writing a custom `with_validation_context` method. This way we are using pre-existing APIs while keeping backward compatibility by not changing any existing APIs

CC: @rafaelfranca It would be good for review if the CI turns out green.

PR #55912, opened 2025-10-15
Fix negative scopes for enums to include records with `nil` values

fatkodima:
Since NULL != NULL in SQL, the condition where `column != 'value'` always ignores rows with NULLs, but should include them too.

Fixes #55908.


rafaelfranca:
This is very likely to break a lot of people applications. But it is correct. I don't see a way to deprecate it.

But can we add a CHANGELOG entry and and entry in the upgrading guide?

fatkodima:
Thanks 👍. Added some docs.

rafaelfranca (approved):

PR #55914, opened 2025-10-15
[Getting Started] Introduce flash earlier in the tutorial and give it more detail [ci skip]

harsh183:
### Motivation / Background

We introduce flash quite late in the tutorial, section 16, but I think it can be useful to introduce it earlier since we can utilize it with the earlier CRUD actions. This way the reader of the tutorial gets some more debug information while running their scripts and have more than just a 422 http error to know a validation failed.

### Detail

* moved the first reference to flash earlier in the tutorial
* added flash being set in more parts of the tutorial since we now have it available and it's a good practice
* migrated the mentions of `notice` to the actual `flash` variable
* added some explanation about it because exposing the reader to the official flash API to handle notices felt quite worth it so I added some more explanation of how it works as well.

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -1115,6 +1115,48 @@ Let's refactor this to use these helpers:
 </div>
 ```
 
+### Displaying Flash Messages
+
+The
+[Flash](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash.html)
+provides a way to pass temporary messages between actions. You'll commonly use
+it to display success messages or errors to users after they perform an action
+like creating or updating a record.
+
+Flash messages are stored temporarily and cleared automatically after being
+displayed. The flash is a hash, so you can use any key you want. Common keys
+are `:notice` for success messages and `:alert` for errors.
+
+Let's add flash messages to the application layout. The layout file wraps around
+all your views, making it the perfect place for elements that appear on every
+page.
+
+Update `app/views/layouts/application.html.erb`:
+
+```erb#11-12
+<!DOCTYPE html>
+<html>
+  <head>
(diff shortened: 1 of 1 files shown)


rafaelfranca:
@excid3 can you review this change please?

guides/source/getting_started.md
```diff
@@ -1115,6 +1115,48 @@ Let's refactor this to use these helpers:
 </div>
 ```
 
+### Displaying Flash Messages
+
+The
+[Flash](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash.html)
```
MatheusRich, line 1121: I know this is how it's written in the docs, but "the Flash" sounds weird to me. Maybe 
```suggestion
The
[`flash` object](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash.html)
```

zzak: This is what it's been called for years :joy:
https://github.com/rails/rails/blob/e17e25cd23e8abd45b1706463dd57c90fa6dcb7c/railties/guides/source/getting_started.textile?plain=1#L994

MatheusRich: I had no idea!

harsh183: I guess we keep it then?

MatheusRich: Yes

excid3:
I like expanding the Flash explanation here @harsh183!

I think we should use form errors for this instead of `flash.now` so I've made an alternative PR for this that incorporates your direction. #55951

harsh183:
closing since I like the alternative PR. I'll reopen if needed

PR #55916, opened 2025-10-16
Fix passing both module: and shallow: to resources

### LINKED ISSUE #55915: `shallow` resources option is not respected when module is given
`shallow` resources option is not respected when `module` option is given on `main` and `8-1-stable`. `git bisect` is showing 3b4255e1800d0b53c96db5aeb88aaa66163f74d5 (#53689) as a source of this issue.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails"
  gem "rails", github: "rails/rails", branch: "8-1-stable"
end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

end
Rails.application.initialize!

  resources :evaluators do
    resources :metrics, module: :evaluators, shallow: true
  end
end

class BugTest <  ActionDispatch::IntegrationTest
  include Rack::Test::Methods

  def test_shallow_route
    assert_equal url_for(controller: "evaluators/metrics", action: :update, id: 1, only_path: true), "/metrics/1"
  end

  private
    def app
      Rails.application
    end
end
```

### Expected behavior

Following routes should be generated:

```
   evaluator_metrics GET    /evaluators/:evaluator_id/metrics(.:format)     evaluators/metrics#index
                     POST   /evaluators/:evaluator_id/metrics(.:format)     evaluators/metrics#create
new_evaluator_metric GET    /evaluators/:evaluator_id/metrics/new(.:format) evaluators/metrics#new
         edit_metric GET    /metrics/:id/edit(.:format)                     evaluators/metrics#edit
              metric GET    /metrics/:id(.:format)                          evaluators/metrics#show
                     PATCH  /metrics/:id(.:format)                          evaluators/metrics#update
                     PUT    /metrics/:id(.:format)                          evaluators/metrics#update
                     DELETE /metrics/:id(.:format)                          evaluators/metrics#destroy
```

### Actual behavior

```
    evaluator_metrics GET    /evaluators/:evaluator_id/metrics(.:format)          evaluators/metrics#index
                      POST   /evaluators/:evaluator_id/metrics(.:format)          evaluators/metrics#create
 new_evaluator_metric GET    /evaluators/:evaluator_id/metrics/new(.:format)      evaluators/metrics#new
edit_evaluator_metric GET    /evaluators/:evaluator_id/metrics/:id/edit(.:format) evaluators/metrics#edit
     evaluator_metric GET    /evaluators/:evaluator_id/metrics/:id(.:format)      evaluators/metrics#show
                      PATCH  /evaluators/:evaluator_id/metrics/:id(.:format)      evaluators/metrics#update
                      PUT    /evaluators/:evaluator_id/metrics/:id(.:format)      evaluators/metrics#update
                      DELETE /evaluators/:evaluator_id/metrics/:id(.:format)      evaluators/metrics#destroy

```

```
url_for(controller: "evaluators/metrics", action: :update, id: 1, only_path: true) =>
ActionController::UrlGenerationError: No route matches {action: "update", controller: "evaluators/metrics", id: 1}
```

### System configuration

**Rails version**:  `main` and `8-1-stable`

**Ruby version**:  `3.4.6`

skipkayhil:
Fix #55915

This broke because the keyword arg change started always passing
`shallow:` through to this `scope` block. However, this caused an issue
because passing the `shallow` option caused it to _always_ override the
scope's current `shallow` value, instead of only doing that when
specified.

This commit fixes the behavior by only passing `shallow` to the `scope`
when it is specified, instead of unconditionally.

(no changelog because this fixes an issue in 8.1.0 which hasn't been released?)

### DIFF
--- a/actionpack/lib/action_dispatch/routing/mapper.rb
+++ b/actionpack/lib/action_dispatch/routing/mapper.rb
@@ -1948,8 +1948,10 @@ module ActionDispatch
             end
 
             scope_options = options.slice!(*RESOURCE_OPTIONS)
-            if !scope_options.empty? || !shallow.nil?
-              scope(**scope_options, shallow:) do
+            scope_options[:shallow] = shallow unless shallow.nil?
+
+            unless scope_options.empty?
+              scope(**scope_options) do
                 public_send(method, resources.pop, **options, &block)
               end
               return true
--- a/actionpack/test/controller/resources_test.rb
+++ b/actionpack/test/controller/resources_test.rb
@@ -18,6 +18,10 @@ module Backoffice
   end
 end
 
+module Products
+  class ImagesController < ResourcesController; end
+end
+
 class ResourcesTest < ActionController::TestCase
   def test_default_restful_routes
     with_restful_routing :messages do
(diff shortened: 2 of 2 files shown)


skipkayhil:
~One real test failure...~ ✔️

PR #55900, opened 2025-10-14
Make framework log subscribers to consume structured events

gmcgibbon:
### Motivation / Background

This Pull Request has been created because framework log subscribers currently consume notification events. Since the introduction of structured event subscribers in https://github.com/rails/rails/pull/55690, this doesn't necessarily need to happen anymore. We can instead use the structured event subscribers to produce logs instead for environments like development.

### Detail

This Pull Request adds a base class for framework structured event log subscribers, and changes current log subscribers (private API since https://github.com/rails/rails/pull/55755) to consume structured events instead of notification events.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #55922, opened 2025-10-16
Add missing structured event data for logs

gmcgibbon:
Followup on https://github.com/rails/rails/pull/55904.

Related to #55900.

This will be backported to 8.1.


rafaelfranca:
👍🏽

PR #55924, opened 2025-10-17
Use WeakRef to avoid leaking constants in ActiveSupport::LazyLoadHooks

### LINKED ISSUE #55923: ActiveSupport.run_load_hooks memory leaks on reload
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  gem "debug"
end

require "active_support"
require "active_support/core_ext/object/blank"
require "active_support/testing/autorun"
require "zeitwerk"
require "tmpdir"

  def setup
    @dir = Dir.mktmpdir
    @autoload_path = File.join(@dir, "app", "models")
    FileUtils.mkdir_p(@autoload_path)

    # Minimal Rails-like autoloading setup
    @loader = Zeitwerk::Loader.new
    @loader.push_dir(@autoload_path)
    @loader.enable_reloading
    @loader.setup

    define_foo_class
  end

  def teardown
    FileUtils.remove_entry(@dir)
  end

  def define_foo_class
    File.write(File.join(@autoload_path, "foo.rb"), <<~RUBY)
      class Foo
        ActiveSupport.run_load_hooks(:foo, self)
      end
    RUBY
  end

  def test_reloading_retains_old_class
    # First load
    debugger
    @loader.reload
    foo_1 = Foo # Trigger loading

    # Reload and get the new version
    @loader.reload
    foo_2 = Foo

    # Confirm redefinition
    refute_equal foo_1, foo_2, "Expected Foo to be redefined on reload"

    # Check the ActiveSupport hook registry
    loaded = ActiveSupport.instance_variable_get(:@loaded)[:foo]
    p loaded # => [Foo (old), Foo (new)]

    assert_equal 1, loaded.uniq.size, "Expected only one Foo class to be kept after reload"
  end
end
```

### Expected behavior

After reloading the old classes should be deleted from the array, so the Garbage Collector can reclaim that memory back.

### Actual behavior

The `@loaded` variable is retaining references to the classes, so the GC cannot reclaim that memory in Zeitwerk reloads.

This is not a hypothetical issue, is a memory leaks we are having with the following flow in our application.

1. We use Devise gem, that use the hook in `DeviseController` class [here](https://github.com/heartcombo/devise/blob/main/app/controllers/devise_controller.rb#L225).
2. The `DeviseController` is autoloaded, therefore it is retained in the `@loaded` array.
3. The `DeviseController` class retained inherits from our `ApplicationController`. And out `ApplicationController` have some KB because it contains a lot of logic, therefore producing a constant leak on every reload.

With this, summed to the fact that DelayedJob reloads every 5 seconds by default, the `bin/rails jobs:work` end up eating a lot of memory after hours of development. The same could happen without DelayedJob on the development reloads of `bin/rails s`

### System configuration

**Rails version**: 8.0.3, but also seen in 7.1.5.2

**Ruby version**: 3.3.7

domingo2000:
### Motivation / Background

Fixes #55923

### Detail

This Pull Request changes `ActiveSupport::LazyLoadHooks` to use `WeakRef` in its `@loaded`, variable so it does not memory leaks on rails reloads as described in the Issue.

### Additional information

Rescuing in lost WeakRefs is not ideal, and those object should be removed from the Array. We could implement something like a WeakArray or use another data structure in the `@loaded` variable, but for now it solves the problem. Also the reloads are not high in production, so the array will grow at the same rate as the `reload!` autoloads constants using the hooks, just like is doing right now.

### Checklist

* [ ] Tests are added or updated if you fix a bug or add a feature. (Not sure if this apply to this king of changes, if it do, i can add a test similar to the one of the issue.)

### DIFF
--- a/activesupport/lib/active_support/lazy_load_hooks.rb
+++ b/activesupport/lib/active_support/lazy_load_hooks.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "weakref"
+
 module ActiveSupport
   # = Lazy Load Hooks
   #
@@ -59,7 +61,8 @@ module ActiveSupport
     # * <tt>:run_once</tt> - Given +block+ will run only once.
     def on_load(name, options = {}, &block)
       @loaded[name].each do |base|
-        execute_hook(name, base, options, block)
+        execute_hook(name, base.__getobj__, options, block)
+      rescue WeakRef::RefError
       end
 
       @load_hooks[name] << [block, options]
@@ -73,9 +76,10 @@ module ActiveSupport
     # In the case of the above example, it will execute all hooks registered
     # for +:active_record+ within the class +ActiveRecord::Base+.
     def run_load_hooks(name, base = Object)
-      @loaded[name] << base
+      @loaded[name] << WeakRef.new(base)
       @load_hooks[name].each do |hook, options|
         execute_hook(name, base, options, hook)
+      rescue WeakRef::RefError
       end
     end
 


rafaelfranca:
By default reloads aren't enabled at all in production. And in development, it is expected this list to grow. I'm not sure if it is worth this complexity. How is this leak impacting your application and why?

activesupport/lib/active_support/lazy_load_hooks.rb
```diff
@@ -59,7 +61,8 @@ def self.extended(base) # :nodoc:
     # * <tt>:run_once</tt> - Given +block+ will run only once.
     def on_load(name, options = {}, &block)
       @loaded[name].each do |base|
-        execute_hook(name, base, options, block)
+        execute_hook(name, base.__getobj__, options, block)
+      rescue WeakRef::RefError
```
rafaelfranca, line 65: Should we use a `ObjectSpace::WeakMap` instead?

domingo2000:
> By default reloads aren't enabled at all in production. And in development, it is expected this list to grow. I'm not sure if it is worth this complexity. How is this leak impacting your application and why?

The reloads caused by DelayedJob on development are consuming memory on each reload ([this issue](https://github.com/collectiveidea/delayed_job/issues/776)). So this small leak ends up eating a lot of memory over time.

activesupport/lib/active_support/lazy_load_hooks.rb
domingo2000: I am not familiar with the current implementation. but `@loaded` is an Array, not a hash, that's why i didn't use a WeakMap. 

If you guide how can we change from Array to Hash i could use the `ObjectSpace::WeakMap`

domingo2000:
Just watched that ActiveJob::Execution reloads the code without the need of manual reloading that DelayedJob does. 

This becomes less important, cause rails will only reload if changes are detected, making the problem smaller.

So there is a leak, but the impact is small, if you think that this add more complexity I am ok with closing the issue + PR.

activesupport/lib/active_support/lazy_load_hooks.rb
zzak: I could be missing something but isn't it just:

```diff
diff --git a/activesupport/lib/active_support/lazy_load_hooks.rb b/activesupport/lib/active_support/lazy_load_hooks.rb
index 456e7be84f..a611656793 100644
--- a/activesupport/lib/active_support/lazy_load_hooks.rb
+++ b/activesupport/lib/active_support/lazy_load_hooks.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "objspace"
+
 module ActiveSupport
   # = Lazy Load Hooks
   #
@@ -44,7 +46,7 @@ module LazyLoadHooks
     def self.extended(base) # :nodoc:
       base.class_eval do
         @load_hooks = Hash.new { |h, k| h[k] = [] }
-        @loaded     = Hash.new { |h, k| h[k] = [] }
+        @loaded     = Hash.new { |h, k| h[k] = ObjectSpace::WeakMap.new }
         @run_once   = Hash.new { |h, k| h[k] = [] }
       end
     end
@@ -58,7 +60,7 @@ def self.extended(base) # :nodoc:
     # * <tt>:yield</tt> - Yields the object that run_load_hooks to +block+.
     # * <tt>:run_once</tt> - Given +block+ will run only once.
     def on_load(name, options = {}, &block)
-      @loaded[name].each do |base|
+      @loaded[name].each_key do |base|
         execute_hook(name, base, options, block)
       end
 
@@ -73,7 +75,7 @@ def on_load(name, options = {}, &block)
     # In the case of the above example, it will execute all hooks registered
     # for +:active_record+ within the class +ActiveRecord::Base+.
     def run_load_hooks(name, base = Object)
-      @loaded[name] << base
+      @loaded[name][base] = true
       @load_hooks[name].each do |hook, options|
         execute_hook(name, base, options, hook)
       end
```

byroot: Pretty much yes. The need here is indeed for an array, but WeakMap can be used as a Set which is close enough, and it's generally easier to use than WeakRef.

byroot:
So I get the issue here, and I guess I'm mostly fine with the change, however I wonder if this could introduce the opposite effect, e.g. some code with:

```ruby
ActiveSupport.run_load_hooks(:foo, SomeClass.new)
```

Where the load hooks are currently the only live reference to the object.

Another thing I wonder, is whether this should always be a weakref, or only when reloading is enabled?

fxn:
Let me go a step back: Why is `@loaded` values a collection? Which are the use cases for wanting to run load hooks on multiple bases? Are users aware that their `on_load` blocks may be executed an arbitrary number of times? (Docs do not say so.)

fxn:
I see Rails itself may run

```
ActiveSupport.run_load_hooks(:action_controller, self)
```

in two places, for API and regular controllers. At least, we miss docs on this, people should be told `on_load` hooks may run multiple times.

But we need to think from a logical point of view, this feature is not designed to be called with reloadable objects. Because while `on_load` hooks may be OK when be called with multiple _current_ bases, what does not seem to be well designed is that they may be called with _stale_ objects.

This feature, as it is written today, is not meant to be used with reloadable objects (objects stored in a reloadable constant).

fxn:
But, at the same time, the contract for `on_load` is:

> Declares a block that will be executed when a Rails component is fully loaded.

So, if an app has API and regular controllers, is the "action_controller" component fully loaded **twice**? How does that make sense?

Multiple bases for the same "component" seem odd to me.

rafaelfranca:
In this case it is just the documentation that is using the wrong word. What we want is a way to hook lazy behavior when something is loaded. It could be single or multiple "things" for a single hook.

rafaelfranca:
> This feature, as it is written today, is not meant to be used with reloadable objects (objects stored in a reloadable constant).

I'm not sure that is true. We use in Rails in reloadable objects.

See https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/actionmailbox/app/models/action_mailbox/inbound_email.rb#L55
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/actionmailbox/app/models/action_mailbox/record.rb#L9
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/actiontext/app/models/action_text/encrypted_rich_text.rb#L11
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/actiontext/app/models/action_text/record.rb#L11
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/actiontext/app/models/action_text/rich_text.rb#L96
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/activestorage/app/models/active_storage/attachment.rb#L176
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/activestorage/app/models/active_storage/blob.rb#L379
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/activestorage/app/models/active_storage/record.rb#L7
https://github.com/rails/rails/blob/472ef0697a2ab066ae4544ebddeea8b793f4c4f2/activestorage/app/models/active_storage/variant_record.rb#L8

I don't see why it should not be. The use case is pretty much like `Zeitwerk#on_load` hook, that also works in reloadable objects.

fxn:
> I'm not sure that is true. We use in Rails in reloadable objects.

This is what I mean:

```
% cat app/models/foo.rb 
class Foo
  p object_id
  ActiveSupport.run_load_hooks(:foo, self)
end
% bin/rails c
Loading development environment (Rails 8.0.3)
> Foo
7648
=> Foo
> reload!
Reloading...
=> nil
> Foo
9808
=> Foo
> ActiveSupport.on_load(:foo) { p self.object_id }
7648
9808
=> [[#<Proc:0x0000000129a92610 (del-hook):4>, {}]]
```

That is wrong, the user does not want to be called with stale objects (7648).

fxn:
The general rule is that storing objects in a place that survives reloads is not designed to be used with reloadable code. In this case, that place is `@loaded`.

rafaelfranca:
Ah, I see what you mean. I agree. Whould not the WeakRef that care of that?

domingo2000:
> Ah, I see what you mean. I agree. Whould not the WeakRef that care of that?

The WeakRef is going to work to clear that, already tested in our app. The problem is the one commented here

> So I get the issue here, and I guess I'm mostly fine with the change, however I wonder if this could introduce the opposite effect, e.g. some code with:
> 
> ```ruby
> ActiveSupport.run_load_hooks(:foo, SomeClass.new)
> ```
> 
> Where the load hooks are currently the only live reference to the object.
> 
> Another thing I wonder, is whether this should always be a weakref, or only when reloading is enabled?

For example, [in tests](https://github.com/rails/rails/blob/main/activesupport/test/lazy_load_hooks_test.rb#L18) of the method, that case happen. 

Because of that, if GC enter running those tests, you could free that object early, and the tests could become flaky. 

But looking into [github grep](https://grep.app/search?q=ActiveSupport.run_load_hooks) that seems like an edge case, and only is beign used at rails internal tests.

fxn:
@rafaelfranca it would indeed.

But then, being a stale reloadable reference and being a candidate for garbage collection are not equivalent. That was the point in @byroot's comment [here](https://github.com/rails/rails/pull/55924#issuecomment-3417939182).

This one may be difficult to adddress, the "leak" is just a symptom of something more fundamental, it is a problem with the API not being strict enough.

domingo2000:
My other thoughts were to just add a callback on Zeitwerk [on_unload](https://github.com/fxn/zeitwerk/blob/a09f0cb48fe7d56578a2880037b5a5a4474c30aa/lib/zeitwerk/loader/config.rb#L255-L276) as [suggested here](https://github.com/sorbet/sorbet/issues/9455#issuecomment-3403638827) in a similar leak on sorbet-runtime. A block in that callback could remove the entries in the `@loaded` variable

But I don't know if rails reloading mechanism supports a similar hooks when not using Zeitwerk as autoloader.

But as @fxn maybe it is a more fundamental cause of how the `@loaded` variable is beign handle as a class instance variable that survives reloads.

fxn:
Off the top of my head, we could perhaps have an `on_unload` callback set that unregisters loaders. Mod details and performance, that would do it.

fxn:
@domingo2000 great minds think alike! `@loaded` per se is fine, what we need to clear is its contents.

fxn:
> But I don't know if rails reloading mechanism supports a similar hooks when not using Zeitwerk as autoloader.

This is not a supported scenario, we have only one autoloading mechanism since 7.0.

The railtie of Active Record already uses `on_unload` ([here](https://github.com/rails/rails/blob/bfd0ebc0e765d14de267fcbbc6a293db1cacc321/activerecord/lib/active_record/railtie.rb#L423-L435)).

domingo2000:
> > But I don't know if rails reloading mechanism supports a similar hooks when not using Zeitwerk as autoloader.
> 
> This is not a supported scenario, we have only one autoloading mechanism since 7.0.
> 
> The railtie of Active Record already uses `on_unload` ([here](https://github.com/rails/rails/blob/bfd0ebc0e765d14de267fcbbc6a293db1cacc321/activerecord/lib/active_record/railtie.rb#L423-L435)).

Oh, that's good to know, if only Zeitwerk as autoloding mechanism is supported then I could implement that as an `initializer`. If @rafaelfranca or @byroot agrees I can change the current implementation to that.

rafaelfranca:
I think it is ok to couple this with zeitwerk

fxn:
There is still one limitation in that approach.

As you know, Ruby does no have reloadable references. What we have is strictly constant reloading.

Normally, reloaded constants stored objects that were not stored anywhere else and are subject to garbage collection. That is the mental model with the `User` model.

But think about `X = 1` in `app/models/x.rb`. You can autoload `X`, and on reload you'll get `X` wiped and loaded again (on demand). But the object reference survives the reload.

The consequence is that even if `on_unload` tells you a certain constant path storing a certain value is being unloaded, you cannot assume any of the values with the same object ID stored in `@loaded` came from that constant.

Certainly an edge case, but one that is telling us the solution is not quite exactly right. Again, I believe this is hard to reconcile.

fxn:
I studied whether we could replace the array of bases with a single value, on the assumption that `run_load_hooks` would perhaps be thought to be called "once" for a given event (mod reloads). But the support for multiple calls with different bases seems to be intentional.

On the other hand, Ruby does not have a notion of "reloadable objects", we only have "reloadable constants", and that makes stale objects not detectable.

The only way I can think of to address the leak and the multiple unwanted `on_load` calls with stale objects is to provide API for the user to decide:

```ruby
ActiveSupport.run_load_hooks(:foo, base, forget_on_reload: true)
```

Not elegant, but I don't see another option that is correct.

fxn:
I suspect the design of these lazy hooks has evolved without clear boundaries. The docs of `on_load` say:

> Declares a block that will be executed when a Rails component is fully loaded. If the component has already loaded, the block is executed immediately.

That implies:

1. `run_load_hooks` is executed at most _once_, because conceptually you fully load a Rails component once.
2. Bases are not reloadable, since Rails components aren't.

But that is not what's implemented, it stores objects, one or more, and it is not designed to be used with reloadable bases. That use case sneaked in somehow.

PR #55926, opened 2025-10-17
Optional event reporter-specific filter parameters

gmcgibbon:
### Motivation / Background

Related to https://github.com/rails/rails/pull/55489

This Pull Request has been created because `ActiveSupport::EventReporter` is currently forced to use the global filter parameters set by Rails. This isn't always desired behaviour, especially if you want to filter different keys for events vs parameters. So, we should provide an optional override.

### Detail

This Pull Request changes `ActiveSupport::EventReporter` to allow use of separate filter parameters for event payload filtering. Filtering will by default inherit the global filter parameters, and use specific ones if they are set.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/event_reporter.rb
+++ b/activesupport/lib/active_support/event_reporter.rb
@@ -278,6 +278,9 @@ module ActiveSupport
     attr_reader :subscribers # :nodoc
 
     class << self
+      # Filter parameters used to filter event payloads. If nil,
+      # Active Support's filter parameters will be used instead.
+      attr_accessor :filter_parameters
       attr_accessor :context_store # :nodoc:
     end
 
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/event_reporter.rb
```diff
@@ -278,6 +278,7 @@ class EventReporter
     attr_reader :subscribers # :nodoc
 
     class << self
+      attr_accessor :filter_parameters
```
adrianna-chang-shopify, line 283: Should we doc this then?

activesupport/test/event_reporter_test.rb
```diff
@@ -259,6 +259,32 @@ def emit(event)
       end
     end
 
+    test "default filter_patameters is used by default" do
```
adrianna-chang-shopify:
```suggestion
    test "default filter_parameters is used by default" do
```

activesupport/lib/active_support/event_reporter.rb
gmcgibbon: Yeah, I think it will appear in docs by default if we don't nodoc. I'll add a description 👍

PR #43485, opened 2021-10-18
Add `ActiveRecord::Base.prohibit_shard_swapping`

seejohnrun:
### Summary

When using sharded databases for the lifecycle of an entire request,
it's often desirable to ensure that the databases shard is not
unintentionally changed.

This commit introduces `ActiveRecord::Base.prohibit_shard_swapping`
which takes a block and prevents swapping of shards for its duration.

cc / @eileencodes

PR #55927, opened 2025-10-17
Update shard swap prohibition to be more granular

flavorjones:
### Motivation / Background

In #43485, shard swapping prohibition was introduced. It was implemented as a boolean that applied to all databases (within the isolated execution state).

I believe the spirit of the original feature was to prevent swapping shards on a specific database, and so I think the implementation should support the use case of a multi-database application.

cc @eileencodes @seejohnrun @matthewd 

### Detail

This change scopes the shard swap prohibition to the "connection specification name", which allows an application to prohibit shard swapping on a specific database while allowing it on all others.

Test coverage has been added to cover the `prohibit_shard_swapping(false)` use case.

### Additional information

Depends on https://github.com/rails/rails/pull/55933, please review that PR first. 🙏

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Database-specific shard swap prohibition
+
+    In #43485 (v7.0.0), shard swapping prohibition was introduced as a global
+    switch that applied to all databases.
+
+    For the use case of a multi-database application, the global prohibition is
+    overly broad, and so with this change the method `prohibit_shard_swapping`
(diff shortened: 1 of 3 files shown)


flavorjones:
Still working on making this work for `connected_to_many`.

flavorjones:
@eileencodes If you get a minute, I would appreciate your feedback on this -- I'm not sure whether this is aligned with the original feature's intent.

If the intention really was to block shard changes on all databases, then I can modify this PR to leave that as the default behavior add an explicit parameter to opt into the granular behavior. LMK?

PR #55928, opened 2025-10-17
Batch reset sequence

gmcgibbon:
### Motivation / Background

This Pull Request has been created because the sequence resetting we use in fixture code is a little slow when you have a lot of tables. On Shopify, it currently takes multiple seconds. Some of this information can be derived from the schema cache, or assumed if we know the application is using default sequence names.

### Detail

This Pull Request changes the postgres adapter by introducing `reset_column_sequences!` to batch reset column sequences. Due to the complex nature of the queries used to get this info, and the fact that most of this data can be found in the schema cache, I've opted to just expose a batching method with all values provided. Later on, we might want to change the fixture implementation to do this faster using this method. This is all nodoc anyway, so the risk here is relatively low.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -615,6 +615,25 @@ module ActiveRecord
         end
       end
 
+      # Executes SQL statements in the context of this connection without
+      # returning a result.
+      def execute_batch(statements, name = nil, **kwargs) # :nodoc:
+        statements.each do |statement|
+          intent = QueryIntent.new(
+            adapter: self,
+            processed_sql: statement,
+            name: name,
+            binds: kwargs[:binds] || [],
+            prepare: kwargs[:prepare] || false,
+            allow_retry: kwargs[:allow_retry] || false,
+            materialize_transactions: kwargs[:materialize_transactions] != false,
+            batch: kwargs[:batch] || false
+          )
+          intent.execute!
+          intent.finish
+        end
+      end
+
       private
         DEFAULT_INSERT_VALUE = Arel.sql("DEFAULT").freeze
         private_constant :DEFAULT_INSERT_VALUE
@@ -645,23 +664,6 @@ module ActiveRecord
             allow_retry: allow_retry,
             materialize_transactions: materialize_transactions
           )
-        end
-
-        def execute_batch(statements, name = nil, **kwargs)
-          statements.each do |statement|
-            intent = QueryIntent.new(
-              adapter: self,
-              processed_sql: statement,
-              name: name,
-              binds: kwargs[:binds] || [],
-              prepare: kwargs[:prepare] || false,
-              allow_retry: kwargs[:allow_retry] || false,
-              materialize_transactions: kwargs[:materialize_transactions] != false,
-              batch: kwargs[:batch] || false
-            )
-            intent.execute!
-            intent.finish
-          end
         end
 
         def build_fixture_sql(fixtures, table_name)
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -357,32 +357,172 @@ module ActiveRecord
           end
         end
 
-        # Resets the sequence of a table's primary key to the maximum value.
-        def reset_pk_sequence!(table, pk = nil, sequence = nil) # :nodoc:
-          unless pk && sequence
-            default_pk, default_sequence = pk_and_sequence_for(table)
+        class SequenceReset # :nodoc:
+          Data = Struct.new(:column, :sequence, :max_value, :min_value)
 
-            pk ||= default_pk
-            sequence ||= default_sequence
+          def initialize(adapter, tables)
+            @adapter = adapter
+            @tables = tables.to_h { |table, data| [table.to_s, Data.new(*data)] }
           end
 
-          if @logger && pk && !sequence
-            @logger.warn "#{table} has primary key #{pk} with no default sequence."
+          def reset
+            backfill
+
+            reset_sqls = @tables.values.filter_map do |data|
+              next unless data.sequence && (data.max_value || data.min_value)
+              reset_sequence_sql(data.sequence, data.max_value, data.min_value)
+            end
+
+            @adapter.send(:execute_batch, reset_sqls, "SCHEMA")
           end
 
-          if pk && sequence
-            quoted_sequence = quote_table_name(sequence)
-            max_pk = query_value("SELECT MAX(#{quote_column_name pk}) FROM #{quote_table_name(table)}")
-            if max_pk.nil?
-              if database_version >= 10_00_00
-                minvalue = query_value("SELECT seqmin FROM pg_sequence WHERE seqrelid = #{quote(quoted_sequence)}::regclass")
-              else
-                minvalue = query_value("SELECT min_value FROM #{quoted_sequence}")
+          private
+            def backfill
+              backfill_sequences_with_primary_keys
+              backfill_sequences_with_uuids
+              backfill_max_values
+              backfill_min_values
+            end
+
+            def backfill_sequences_with_primary_keys
+              tables_without_sequences = @tables.select do |table, data|
+                !data.column || !data.sequence
+              end.keys
+
+              return if tables_without_sequences.none?
+
+              sql = primary_key_column_and_sequence_sql(tables_without_sequences)
+              primary_key_column_and_sequences = @adapter.exec_query(sql, "SCHEMA")
+
+              primary_key_column_and_sequences.rows.each do |table, column, namespace, sequence|
+                table = table.delete_prefix('"').delete_suffix('"')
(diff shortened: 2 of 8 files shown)


zzak:
Is it just postgres which is affected by this?

gmcgibbon:
Yes, mysql and sqlite don't have sequences, only postgres. We reset sequences when inserting fixtures to reset auto incrementing values so that we don't accidentally increment to an already exiting value. We reset sequences for postgres only by using respond_to?.

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -335,47 +335,58 @@ def serial_sequence(table, column)
...
             end
           end
         end
 
-        # Resets the sequence of a table's primary key to the maximum value.
-        def reset_pk_sequence!(table, pk = nil, sequence = nil) # :nodoc:
-          unless pk && sequence
-            default_pk, default_sequence = pk_and_sequence_for(table)
+        # Resets the sequence of a table's sequence column (by default, the primary key) to the maximum value.
+        def reset_column_sequence!(table, column = nil, sequence = nil) # :nodoc:
```
skipkayhil: While the method is `nodoc`d, I worry that its stability for so long has lead to a lot of use: https://github.com/search?type=code&q=reset_pk_sequence

Since we're just renaming it but not really changing any of its functionality, is this worth the cost of breaking backwards compatibility?

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -335,47 +335,58 @@ def serial_sequence(table, column)
...
             end
 
-            internal_execute("SELECT setval(#{quote(quoted_sequence)}, #{max_pk || minvalue}, #{max_pk ? true : false})", "SCHEMA")
+            internal_execute("SELECT setval(#{quote(quoted_sequence)}, #{max_column_value || min_column_value}, #{max_column_value ? true : false})", "SCHEMA")
+          end
+        end
+
+        # Batch resets column sequences. Passed in the shape of [[table, column, sequence, max_column_value, min_column_value]].
+        # Reduces query round trips for sequence resets when there are a large amount of tables.
+        def reset_column_sequences!(args) # :nodoc:
```
skipkayhil: While _we_ want to use this method in our own app, does it make sense to define it (`nodoc`) without any usage inside Active Record itself?

gmcgibbon: Right. I think the problem is mostly rooted in the batch query method being private, which makes us want to upstream early. I can likely hook this method into the current fixture code. It won't be the fastest possible, but it will be faster that the current, and integrated to an acceptable point after that.

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
gmcgibbon: A lot of that is old code that will likely never be updated, so I'm not concerned. If developers use private API, they should be prepared to fix breakages. I think the naming here should be changed, but we can discuss that separately. The batching is more important, so I'll revert the first commit for now. We can at least keep the real behaviour noted in the comments.

matthewd:
It definitely sounds like there's a potential improvement here, but this doesn't look like a complete change in its current state.

gmcgibbon:
We need to batch pk_and_sequence_for and the min / max value queries. I'll work on that next, but it will take a little more effort to get right.

gmcgibbon:
I've added a proper batching solution, hooked it up to fixtures, and made sure it backfills data lazily. This should be much faster for apps with lots of tables. I will draw a quick benchmark.

gmcgibbon:
On an app with ~50 active records, you can see a slight difference:
Before:
Screenshot 2025-10-25 at 2 06 47 AM
After:
Screenshot 2025-10-25 at 2 06 38 AM
This really only becomes an issue when you have more than 500 models. Then, the current approach can take a few seconds.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -555,6 +555,10 @@ def internal_exec_query(...) # :nodoc:
         cast_result(internal_execute(...))
       end
 
+      def internal_batch_exec_query(...) # :nodoc:
+        cast_result(execute_batch(...))
+      end
```
skipkayhil: Should the `SequenceReset` class just return the SQL to execute so the adapter could continue calling private `execute_batch`?

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -357,32 +357,172 @@ def set_pk_sequence!(table, value) # :nodoc:
...
+                sequence ? select_min_column_value_sql(sequence) : nil
+              end
+              min_values = tables_min_value_sqls.map { |sql| @adapter.query_value(sql, "SCHEMA") if sql }
+
+              tables_without_max_or_min_values.zip(min_values).each do |table, (min_value)|
+                @tables[table].min_value = min_value
+              end
+            end
+
+            def primary_key_column_and_sequence_sql(tables)
```
skipkayhil, line 454: It feels odd to me that we pull sql/querying into this custom class when the same sql/querying (but just for a single table) is on the adapter itself. Should we just be making the existing methods work with multiple tables?

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -357,32 +357,172 @@ def set_pk_sequence!(table, value) # :nodoc:
...
+                sequence ? select_max_column_value_sql(table, column) : nil
+              end
+              max_values = tables_max_value_sqls.map { |sql| @adapter.query_value(sql, "SCHEMA") if sql }
+
+              tables_without_max_or_min_values.zip(max_values).each do |table, max_value|
+                @tables[table].max_value = max_value
+              end
+            end
+
+            def backfill_min_values
```
skipkayhil, line 438: Since the implementation is new, I'd love to discuss whether we actually need to reset sequences back to their min value.

My understanding of the need to reset sequences is that we're trying to ensure a sequence doesn't generate a value that's been hard-coded in a fixture. If we query for max and get no value, I'd expect that there are no fixtures for that table, and so we shouldn't need to reset the column in that case?

gmcgibbon: We have tests that cover this use-case, but I haven't looked into why precisely.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
gmcgibbon: Sure, I can make that change.

gmcgibbon: Actually, since the adapter is needed to backfill data, having this in one class is a bit simpler. We'd just be making this change solely to appease the current API, so I'm not 100% sold on it. Other query methods have nodoc public endpoints, so batching should too IMO.

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
gmcgibbon: I originally wanted to deprecate the original methods, but they're used elsewhere in non-batching manners. The backfill logic doesn't belong on the adapter IMO, so we can extract this into a separate method, but I would rather do that when it is necessary as opposed to prematurely.

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
gmcgibbon: Looks like we have test coverage for minimum values here:
```
ARCONN=postgresql bin/test test/cases/fixtures_test.rb:718 test/cases/fixtures_test.rb:729
```
It isn't specific as to why. This decision looks rather ancient: https://github.com/rails/rails/commit/7c8f3edc99560d15ae715bdbe99a32bc538e9396

It _looks_ like we could remove this assumption in code without breaking anything, but I'd rather do that in a followup PR because the potential to revert that is a lot higher.

activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
```diff
@@ -357,32 +357,172 @@ def set_pk_sequence!(table, value) # :nodoc:
...
+              backfill_max_values
+              backfill_min_values
+            end
+
+            def backfill_sequences_with_primary_keys
+              tables_without_sequences = @tables.select do |table, data|
+                !data.column || !data.sequence
+              end.keys
+
+              return unless tables_without_sequences.any?
```
skipkayhil:
```suggestion
              return if tables_without_sequences.empty?
```

I like `empty?` over `any?` because `empty?` is a length check where `any?` iterates, but not blocking

gmcgibbon:
@matthewd when you have a minute, can let me know if you would support making `execute_batch` nodoc public? It seems similar to the `execute` method, but it is currently private.

PR #55929, opened 2025-10-17
Allow autosave association methods to be overwritten

Noah-Silvera:
### Motivation / Background

There are many rails gems out there that define active record classes to be consumed.

Sometimes, consumers of those gems want to override the properties, such as associations of the active record classes in the gem. When you override an association by redefining it using a concern, prepended module, etc, rails generally lets you do this and respect it. For example, passing association arguments such as `disable_joins`, `foreign_key`, `inverse_of` to `has_many` in an overriding class will change the original ActiveRecord Class' behavior (e.g. association.rb:45:219:360)

However, if you try to override the `autosave` option, it does not change the behavior of the underlying class.

```
class OriginalClass < ActiveRecord::Base
  has_many :items
end

class MyExtension
  extend ActiveSupport::Concern

  # Adding autosave: true here doesn't do anything
  has_many :items, autosave: true

  OriginalClass.prepend(self)
end
```

This is because even though you can override the reflection attached to the class, the original reflection is still referenced within blocks that define the methods related to autosaving associations. This means that the original reflection behavior (in this example, autosave: false) is still used.

### Detail

This Pull Request changes the AutosaveAssociation behavior so that if a prepended module/subclass etc calls association methods (`has_many`, `has_one`, etc) that match an existing association on the class being modified, the old `autosave` association option value will be overwritten and the new value respected, like the behavior of other association options. 

### Additional information

The `add_autosave_association_callbacks` method are only defined when an association is first built; or in the example I gave above, when an association is rebuilt. I was not able to get this method to fire more than the number of times association helpers like `has_many` were called in the app with a specific association name. I don't think that checking if the method was already defined was buying any performance improvements, but it was preventing the existing behavior on a active record class from being overwritten.

I did consider an alternative that would keep the existing method defined checks but still allow for overriding the existing callbacks by passing a digest of the reflection. However, IMO this had more performance and complexity implications than what would be gained.

[Alternative Approach](https://gist.github.com/Noah-Silvera/5b156d60f8532bcd6b4f01b4611e8246)

### Checklist

### DIFF
--- a/activerecord/lib/active_record/autosave_association.rb
+++ b/activerecord/lib/active_record/autosave_association.rb
@@ -216,18 +216,23 @@ module ActiveRecord
 
         def define_autosave_validation_callbacks(reflection)
           validation_method = :"validate_associated_records_for_#{reflection.name}"
-          if reflection.validate? && !method_defined?(validation_method)
+          if reflection.validate?
             if reflection.collection?
+              # TODO - I'll have to redefine this method, it depends on autosave
               method = :validate_collection_association
             elsif reflection.has_one?
+              # This one depends on inverse_of and stuff
               method = :validate_has_one_association
             else
               method = :validate_belongs_to_association
             end
+            # TODO - is calling `validate` only once here enough?
+            if !method_defined?(validation_method)
+              validate validation_method
+              after_validation :_ensure_no_duplicate_errors
+            end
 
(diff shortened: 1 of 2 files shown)


fatkodima:
We do not accept WIP PRs. Please finish the implementation and open a non-WIP PR.

PR #51432, opened 2024-03-27
[RF-DOCS] Action View Helpers Documentation [ci-skip]

Ridhwana:
### Motivation / Background

This Pull Request has been created to provide clarity to the Action View Helpers guide.

### Detail

Here are some notes on what has changed:

Some notes:

- AssetTagHelper talks about a bunch of *_tag helpers, but also some of them mention *_url and *_path helpers. (which live in AssetUrlHelper). Those are helpful to know, but less frequently used I think. It's also incomplete (some mention it, others don't, *_tag helpers have their *_url/*_path counterparts, or use the basic `asset_path/url`.) I don't think it's worth expanding with all these variants, I'd rather remove the individual ones, and if we want to mention, we can maybe bundle them together within the same subsection instead, or just as examples of the related *_tag sections.
- The cache helper could use some more info. I don't think it's worth linking to the Fragments API, but rather to the [fragments caching piece of the caching guide](https://edgeguides.rubyonrails.org/caching_with_rails.html#fragment-caching), where it goes into more details. The example could also be expanded to show something that at least gets a record as an example too.
- `content_for` could perhaps benefit from showing `content_for?` in the example, somehow.
- We do not want to show any form helpers here, just link to the specific guide, but we can maybe make that paragraph clearer.
- `escape_javascript` could use an example (not the one that's on the API as it seems very jquery-related... maybe some other example of embedding some Ruby in JS or something)
- mention of `rails-html-sanitizer` could be linked.
- "NB: The output may still contain unescaped '<', '>', '&' characters and confuse browsers." should be a note?
- UrlHelper could have a few extra items like `current_page?` and `mail_to`, which can be commonly used. (no need to show each and every method, just the most common ones)
- CsrfHelper seems unnecessary to explain here. (we don't add CspHelper for instance... not sure we need to get into those, can probably skip them.)
- We could have a section on `TextHelper` mentioning things like `simple_format`, `truncate`, `excerpt`, maybe some other(s). (again, not everything, just the most commonly used ones)
- We could have a section on `TagHelper` mentioning `content_tag` and `tag`.

- Updated the 'what you will know section'
- Formatted all sections to be similar where there is some description under the helper and then we get into the methods
- Column count adherence.
- Removed Benchmark that was deprecated
- Made tweaks to language and clarity. 
- Added more clarity to `capture` and `content_for`
- Added more examples where necessary.
- Remove WIP from index page
- Rearranged the sections 

### Checklist

Before submitting the PR make sure the following are checked:

[...]

PR #55932, opened 2025-10-18
[Getting Started] Add file paths to code blocks [ci skip]

harsh183:
### Motivation / Background

I talked to some people who did the tutorial and found that they were finding the blocks confusing since the filename was not always mentioned, especially later in the tutorial. For someone familiar with rails and path conventions its easy to find the way back so I didn't notice this at first but I see how this can helped. After talking to @rafaelfranca  on the rails discord I decided to explore it.

### Detail

* Added filepath with comments at the top of each block as per
  convention I've seen online
* Incremented line numbers accordingly so that the offset is right.

In some cases it does feel a bit redundant early on since we are quite explicit about which file, but it does get more useful as the tutorial goes on and following the convention consistently is useful.

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -284,11 +284,14 @@ helps keep track of changes we make in development (only on our computer) so
 they can be deployed to production (live, online!) safely.
 
 In your code editor, open the migration Rails created for us so we can see what
-the migration does. This is located in
-`db/migrate/<timestamp>_create_products.rb`:
+the migration does.
+
+NOTE: As convention we'll state the filepath as a comment on top of each file
+
 
 ```ruby
-class CreateProducts < ActiveRecord::Migration[8.2]
+# db/migrate/<timestamp>_create_products.rb
+class CreateProducts < ActiveRecord::Migration[8.1]
   def change
     create_table :products do |t|
       t.string :name
@@ -376,6 +379,7 @@ a file at `app/models/product.rb`. This file creates a class that uses Active
 Record for interacting with our `products` database table.
 
 ```ruby
+# app/models/product.rb
 class Product < ApplicationRecord
 end
 ```
@@ -613,6 +617,7 @@ Let's add a `presence` validation to the Product model to ensure that all
 products must have a `name`.
 
 ```ruby
+# app/models/product.rb
 class Product < ApplicationRecord
   validates :name, presence: true
 end
@@ -723,6 +728,7 @@ To define a route in Rails, let's go back to your code editor and add the
 following route to `config/routes.rb`
 
 ```ruby
+# config/routes.rb
 Rails.application.routes.draw do
   get "/products", to: "products#index"
 end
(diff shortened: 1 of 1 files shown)


guides/source/getting_started.md
```diff
@@ -907,6 +923,7 @@ The `index` action will render `app/views/products/index.html.erb`. If we open
 up that file in our code editor, we'll see the HTML it renders.
 
 ```erb
+<!-- app/views/products/index.html.erb -->
```
p8: We currently use ERB comments in similar cases, as they won't show up in the generated HTML:
```suggestion
<%# app/views/products/index.html.erb %>
```

harsh183: I never thought about using ERB this way, this is great. I'll change the PR accordingly

guides/source/getting_started.md
```diff
@@ -1390,7 +1423,8 @@ To learn more about view partials, check out the
 
 Now we can add an Edit link to `app/views/products/show.html.erb`:
 
-```erb#4
+```erb#5
+<!-- app/views/products/show.html.erb -->
```
rafaelfranca: ERB comment here

AmandaPerino:
Edit: I saw @p8's comment above about using it elsewhere in the Guides, so happy to disregard this idea for consistency across tutorials & guides!

> @harsh183 Specifically calling out the file name is such a good idea for beginners, thank you for the suggestion! However, would it not be better to do that outside of the code block, just above it?
> 
> We can use the `<code></code>` formatting in **bold** with a colon at the end, like so:
> 
> Screenshot 2025-10-21 at 13 43 40
> 
> There is already so much code to read and dissect inside the code blocks, and if you use the copy function, it will also copy the notes - code that we don't necessarily want to transfer.
> 
> Thoughts @rafaelfranca @p8 @excid3 ?

p8:
@AmandaPerino The copy function ignores comments. Also an older discussion on this topic: https://github.com/rails/rails/pull/51432#discussion_r1556389845

AmandaPerino:
> @AmandaPerino The copy function ignores comments. Also an older discussion on this topic: [#51432 (comment)](https://github.com/rails/rails/pull/51432#discussion_r1556389845)

There's nothing like rehashing old conversations!  😅 You have a great memory. I was getting mixed reviews testing the copy/paste, but I suspect it has to do with where the comments fall in the block.

harsh183:
Oh wow I didn't know the copy function was so smart. I'm not opposed to footer or header either but since we have comments in other cases it's best to be consistent.

rafaelfranca (approved):
I prefer the comment version since it is used everywhere in the guides and make is possible for us to use the same code block for multiple files if that usage make sense somewhere in the guide, but I'll let @p8 and @AmandaPerino decide.

AmandaPerino:
Agree with all of you! :) Thanks again for working on this @harsh183!

harsh183:
I tried rebase and that didn't give the results I expected. Let me try to fix this.

harsh183:
the merge conflicts are fixed but I have no way to remove the labels that were automatically added. Lmk if I can do something else and apologies for the `rebase` mess.

PR #55934, opened 2025-10-18
Kamal localhost registry

dhh:
Kamal 2.8.0 is shipping with the long-awaited localhost registry option. We should use that out of the box for new installs, so there's one less thing to setup before you can get that "hurrah, I'm live in production" feeling.

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/deploy.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/deploy.yml.tt
@@ -21,15 +21,15 @@ proxy:
   ssl: true
   host: app.example.com
 
-# Credentials for your image host.
+# Registry is where your images are kept.
 registry:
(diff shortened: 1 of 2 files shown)


railties/lib/rails/generators/rails/app/templates/config/deploy.yml.tt
npezza93: With the local registry we no longer need to namespace the image name since it's only you using the registry.

jeromedalbert:
Edit: fixed in Kamal 2.8.1.

I was not able to make this config work with Kamal 2.8.0 on macOS with Docker Desktop. I filed an issue with details in the Kamal repo: https://github.com/basecamp/kamal/issues/1674. It might just be me, but I just wanted to share this here in case a different config or different instructions are needed for macOS.

PR #55936, opened 2025-10-18
Fix ArgumentError when resolving nested class names in reflections

### LINKED ISSUE #55935: ArgumentError in reflection class resolution when nested class name conflicts with top-level module
When resolving an association's class name, if a nested class (e.g., `Company::Department`) shares its demodulized name with a top-level module (e.g., `module Department` ), ActiveRecord raises an `ArgumentError` instead of correctly resolving the nested class.

This appears to happen in `ActiveRecord::Reflection::MacroReflection#_klass`. The code attempts to resolve the class name at the top level (e.g., `::Department`). This finds the top-level `module Department`, which is not an `ActiveRecord::Base subclass`, triggering an `ArgumentError`.

I will open a PR shortly.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  create_table :departments, force: true do |t|
  end

  create_table :department_hierarchies, force: true do |t|
    t.references :department
  end
end

module Company
end

module Department
end

class Company::Department < ActiveRecord::Base
  has_many :department_hierarchies, class_name: "Company::DepartmentHierarchy"
  has_many :child_departments, through: :department_hierarchies, source: :department
end

class Company::DepartmentHierarchy < ActiveRecord::Base
  belongs_to :department
end

  def test_association_stuff
    Company::DepartmentHierarchy.create!(department: Company::Department.new)

    assert Company::Department.first.child_departments
  end
end

```

### Expected behavior

The `child_departments` association should be loaded correctly, resolving to `Company::Department`.

### Actual behavior

The test fails with an `ArgumentError`.

```
  1) Error:
BugTest#test_association_stuff:
ArgumentError: The ::Department model class for the Company::Department#child_departments association is not an ActiveRecord::Base subclass.
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/reflection.rb:512:in 'ActiveRecord::Reflection::AssociationReflection#compute_class'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/reflection.rb:429:in 'ActiveRecord::Reflection::MacroReflection#_klass'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/reflection.rb:1008:in 'ActiveRecord::Reflection::ThroughReflection#klass'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/associations/association.rb:166:in 'ActiveRecord::Associations::Association#klass'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/associations/association.rb:245:in 'ActiveRecord::Associations::Association#ensure_klass_exists!'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/associations/collection_association.rb:35:in 'ActiveRecord::Associations::CollectionAssociation#reader'
    /Users/ima1zumi/.asdf/installs/ruby/3.4.7/lib/ruby/gems/3.4.0/bundler/gems/rails-da3f2c5cfc70/activerecord/lib/active_record/associations/builder/association.rb:107:in 'Company::Department::GeneratedAssociationMethods#child_departments'
    test.rb:58:in 'BugTest#test_association_stuff'

1 runs, 0 assertions, 0 failures, 1 errors, 0 skips
```

### System configuration

**Rails version**: 

main@da3f2c5cfc7041f0c1000136dbd1759c2598b3d7

**Ruby version**: 

3.4.7

ima1zumi:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/55935
Follow up to https://github.com/rails/rails/pull/53969

### Detail

Fix reflection fallback logic for nested class/module name conflicts
When resolving an association for a nested class (e.g., `Nested::Child`),
if a top-level constant (e.g., `module Child`) exists with the same
demodulized name, the reflection mechanism would fail.

The `_klass` method attempts to resolve the top-level constant first
(e.g., `::Child`). The `compute_class` method would find the module but
then raise an error because it is not an `ActiveRecord::Base` subclass.

Previously, this check raised a generic `ArgumentError`. The `_klass`
method only rescued `NameError`, so this `ArgumentError` was not
caught, preventing the intended fallback to the nested class name
(e.g., `Nested::Child`).

This commit introduces two changes:

1.  Changes `compute_class` to raise a specific
    `ActiveRecord::AssociationClassNotActiveRecordBaseError`
    when the resolved class is not an `ActiveRecord::Base` subclass.
    This avoids catching unrelated `ArgumentError` exceptions
    (like "wrong number of arguments") that might occur during
    constantization.

2.  Updates `_klass` to rescue this new
    `AssociationClassNotActiveRecordBaseError` (in addition to
    `NameError`).

This ensures that when the top-level constant lookup fails the
`ActiveRecord::Base` check, the fallback logic correctly proceeds
to resolve the qualified nested class name.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/reflection.rb
+++ b/activerecord/lib/active_record/reflection.rb
@@ -427,7 +427,7 @@ module ActiveRecord
         if active_record.name.demodulize == class_name
           begin
             compute_class("::#{class_name}")
-          rescue NameError
+          rescue NameError, ArgumentError
             compute_class(class_name)
           end
         else
--- a/activerecord/test/cases/reflection_test.rb
+++ b/activerecord/test/cases/reflection_test.rb
@@ -221,6 +221,18 @@ class ReflectionTest < ActiveRecord::TestCase
     assert_equal Nested::NestedUser, reflection.klass
   end
 
+  def test_reflection_klass_for_nested_association_with_top_level_module
+    reflection = ActiveRecord::Reflection.create(
+      :has_many,
+      :children,
+      nil,
+      {},
+      Nested::Child
+    )
+
+    assert_equal Nested::Child, reflection.klass
+  end
+
   def test_aggregation_reflection
     reflection_for_address = AggregateReflection.new(
       :address, nil, { mapping: [ %w(address_street street), %w(address_city city), %w(address_country country) ] }, Customer
--- a/activerecord/test/models/user.rb
+++ b/activerecord/test/models/user.rb
@@ -27,6 +27,9 @@ class UserWithNotification < User
   after_create -> { Notification.create! message: "A new user has been created." }
 end
 
+module Child
+end
+
 module Nested
   class User < ActiveRecord::Base
     self.table_name = "users"
@@ -34,5 +37,8 @@ module Nested
 
   class NestedUser < ActiveRecord::Base
     has_many :nested_users
+  end
+
+  class Child < ActiveRecord::Base
   end
 end


activerecord/lib/active_record/reflection.rb
```diff
@@ -427,7 +427,7 @@ def _klass(class_name) # :nodoc:
         if active_record.name.demodulize == class_name
           begin
             compute_class("::#{class_name}")
-          rescue NameError
+          rescue NameError, ArgumentError
```
daniel-niknam, line 430: I've executed the test, and it passed without this line of code. I think the problem you have raised happens only through providing the source due to the name difference in the associations. Maybe try to replicate exactly what you have in your issue.

On a side note, I'm hesitant towards rescuing ArgumentError. Not sure if there is enough test to cover it, but have you checked what happens if you actually provide some sort of arguments that should result in an ArgumentError? Maybe it's not possible 🤷 .

ima1zumi: On my local, the test failed when I commented out only `ArgumentError` as shown below:

`rescue NameError # , ArgumentError`

How did you run it?

ima1zumi: > On a side note, I'm hesitant towards rescuing ArgumentError. 

You're right, rescuing `ArgumentError` is indeed a bit too broad. Thanks!
My goal is to rescue the error raised from this, so it seems better to perform this check on the caller side.

https://github.com/rails/rails/blob/513d1db9b6503b50caf12da6e52b59feb3e2d50c/activerecord/lib/active_record/reflection.rb#L511-L513

ima1zumi: I've updated it to rescue a custom exception class instead of `ArgumentError`.

daniel-niknam: > On my local, the test failed when I commented out only `ArgumentError` as shown below:
> 
> `rescue NameError # , ArgumentError`
> 
> How did you run it?

Sorry, I think I was wrong. I might have executed the test while I played with the code and got it to pass the test, but I forgot to revert my changes before testing your code. Please ignore that part.

daniel-niknam: > > On a side note, I'm hesitant towards rescuing ArgumentError.
> 
> You're right, rescuing `ArgumentError` is indeed a bit too broad. Thanks! My goal is to rescue the error raised from this, so it seems better to perform this check on the caller side.
> 
> [rails/activerecord/lib/active_record/reflection.rb](https://github.com/rails/rails/blob/513d1db9b6503b50caf12da6e52b59feb3e2d50c/activerecord/lib/active_record/reflection.rb#L511-L513)
> 
> Lines 511 to 513 in [513d1db](/rails/rails/commit/513d1db9b6503b50caf12da6e52b59feb3e2d50c)
> 
>  unless klass < ActiveRecord::Base 
>    raise ArgumentError, "The #{name} model class for the #{active_record}##{self.name} association is not an ActiveRecord::Base subclass." 
>  end

Yes, I understand what you are trying to do. If we could be specific about how we are rescuing, it would be better. ~~I wonder if it's possible not to raise an exception in the first place.~~ The way I'm understanding it is that this is by design. The error must be raised to be captured elsewhere.

daniel-niknam: > I've updated it to rescue a custom exception class instead of `ArgumentError`.

I'm still seeing the same code as before.

daniel-niknam (commented):
As you mentioned in your issue, the core of the problem is that the author in [Commit 2fe2cd4](https://github.com/rails/rails/commit/2fe2cd4bcd52ea3bd3a534b766b07f84c52f6467) interpreted the code:

```ruby
def _klass(class_name) # :nodoc:
  if active_record.name.demodulize == class_name
    return compute_class("::#{class_name}") rescue NameError
  end

  compute_class(class_name)
end
```

to be the same as 

```ruby
def _klass(class_name) # :nodoc:
  if active_record.name.demodulize == class_name
    begin
      compute_class("::#{class_name}")
    rescue NameError
      compute_class(class_name)
    end
  else
    compute_class(class_name)
  end
end
```

while what actually does is:

```ruby
def _klass(class_name) # :nodoc:
  if active_record.name.demodulize == class_name
    begin
      return compute_class("::#{class_name}")
    rescue
      NameError # This is not the same as `raise NameError` or `return NameError`. It's just `NameError`.
    end
  end

  compute_class(class_name)
end
```

I 100% agree with the author that `code rescue Error` is a confusing syntax. Unfortunately, we missed catching the wrong implementation when we rewrote it.

So technically, adding an argument error to the capture is not a problem since we used to rescue everything. I wonder how many other regressions we have introduced by this. Maybe it would be better to revert the code to the original state (without confusion of `code rescue Error`)?

activerecord/lib/active_record/reflection.rb
daniel-niknam: Resolving the comment in favour of [this comment](https://github.com/rails/rails/pull/55936#pullrequestreview-3356021304).

daniel-niknam: I can't resolve it for some reason! 🤷

PR #55938, opened 2025-10-18
Use truthy condition in instead of true for MSSQL support

aidanharan:
### Motivation / Background

This contains just a minor change to the `WithTest#test_with_when_hash_with_multiple_elements_of_different_type_is_passed_as_an_argument` test so that it doesn't need to be reimplemented in the [ActiveRecord SQL Server Adapter](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter). 

MySQL/Postgres/SQLite support queries that contain `WHERE true` but SQL Server doesn't. Instead SQL Server (and MySQL/Postgres/SQLite) support `WHERE 1=1` which serves the same purpose. The change doesn't affect the original purpose of the test. 

### Checklist


aidanharan:
The failing test is unrelated to the PR.

skipkayhil (approved):

skipkayhil:
Thanks! Backported to `8-1-stable` and `8-0-stable` following the backports in #55918

PR #55939, opened 2025-10-18
Remove filter parameter callback mechanisms

gmcgibbon:
### Motivation / Background

This Pull Request has been created because we have too many callback mechanisms to keep filter parameters, filter attributes, and encrypted attributes in sync. We can push the responsibility down to Active Support, and use it to filter event reporter payloads, request parameters, and record attributes.

### Detail

This Pull Request removes callback classes because we don't really need them. We can instead use the recently introduced `ActiveSupport.filter_parameters` as a base source of truth for other frameworks to use as they need. While I was making this change, I also opted to pass a filter parameter to SQL logs originating from an Active Record model, so that subclasses with more specific filters can be used to filter. 

### Additional information

I failed to fix one bug here. Active Record filter attributes being stored on Active Record classes makes it so that filter parameters only apply immediately in eager-loaded environments. In local autoloaded environments where `ActiveRecord::Base` loading is deferred, parameter filtering for specific Active Record attributes won't work until the right model is loaded, which should be addressed at some point. This problem exists in the current implementation though, and will likely require deprecation of `filter_attributes` in Active Records, so we can do this later.

### Checklist


activerecord/test/cases/encryption/encryption_schemes_test.rb
```diff
@@ -4,6 +4,13 @@
 require "models/author_encrypted"
 
 class ActiveRecord::Encryption::EncryptionSchemesTest < ActiveRecord::EncryptionTestCase
+  def run(*)
+    ActiveRecord::Encryption.config.add_to_filter_parameters = false
+    super
+  ensure
+    ActiveRecord::Encryption.config.add_to_filter_parameters = true
+  end
```
gmcgibbon: We probably want this to be false by default, and turned on when the application boots.

activerecord/lib/active_record/encryption.rb
```diff
@@ -8,7 +8,6 @@ module Encryption
     extend ActiveSupport::Autoload
 
     eager_autoload do
-      autoload :AutoFilteredParameters
```
gmcgibbon, line 11: I think this is a breaking change, we may need a deprecation cycle here. The other callback class is rightfully nodoc.

activerecord/lib/active_record/railtie.rb
```diff
@@ -325,16 +325,12 @@ class Railtie < Rails::Railtie # :nodoc:
       end
     end
 
-    initializer "active_record.set_filter_attributes" do
+    initializer "active_record.set_filter_attributes", after: "active_support.set_filter_parameters" do
       ActiveSupport.on_load(:active_record) do
-        self.filter_attributes += Rails.application.config.filter_parameters
+        self.filter_attributes += ActiveSupport.filter_parameters
```
zzak, line 342: 👍

activerecord/test/cases/filter_attributes_test.rb
```diff
@@ -5,7 +5,6 @@
 require "models/admin/user"
 require "models/admin/account"
 require "models/user"
-require "pp"
```
zzak, line 8: 🙈

activerecord/test/models/author_encrypted.rb
```diff
@@ -2,6 +2,8 @@
 
 require "models/author"
 
+ActiveRecord::Encryption.config.add_to_filter_parameters = false
```
zzak: It's interesting to turn this off when loading the model 🤔

activerecord/lib/active_record/encryption.rb
zzak: It seems it should have been private anyway, the only usage is in an initializer:
https://github.com/rails/rails/blob/16e60c9ed00134c585643ac49385a0da8f8a50c4/activerecord/lib/active_record/railtie.rb#L370-L371

No public API is returning the object or modifying it's value afaict

activerecord/lib/active_record/encryption/encryptable_record.rb
```diff
@@ -84,6 +84,9 @@ def global_previous_schemes_for(scheme)
           def encrypt_attribute(name, key_provider: nil, key: nil, deterministic: false, support_unencrypted_data: nil, downcase: false, ignore_case: false, previous: [], compress: true, compressor: nil, **context_properties)
             encrypted_attributes << name.to_sym
 
+            param = self.name ? "#{model_name.element}.#{name}" : name
+            ActiveSupport.filter_parameters << param if ActiveRecord::Encryption.config.add_to_filter_parameters && !name.in?(ActiveRecord::Encryption.config.excluded_from_filter_parameters) && !name.in?(ActiveSupport.filter_parameters)
```
zzak: Maybe refactor to methods?

activerecord/test/models/author_encrypted.rb
gmcgibbon: https://github.com/rails/rails/pull/55939#discussion_r2442597233

Yeah I don't like it either. Let's change the default instead of this hack.

activerecord/lib/active_record/encryption.rb
gmcgibbon: I've opened #56798 to nodoc it before this gets merged.

PR #55941, opened 2025-10-19
PostgreSQL drop database FORCE option.

swebb:
### Motivation / Background

One of the benefits of developing with MySQL is that it allows dropping the current database without first disconnecting clients. As a result developers can use `bin/rails db:reset` and similar, without first shutting down instances of the app, Rails consoles, background workers, etc. By default PostgreSQL fails to drop a database when clients are connected and displays the following error:

  > PG::ObjectInUse: ERROR:  database "xyz" is being accessed by other users (PG::ObjectInUse)

This is frustrating when working in development where the database may be dropped frequently.

### Detail

PostgreSQL 13 added the `FORCE` option to the `DROP DATABASE` statement ([PostgreSQL docs](https://www.postgresql.org/docs/current/sql-dropdatabase.html)) which automatically disconnects clients before dropping the database. This PR adds the `RAILS_PG_DROP_FORCE` environment variable which sets this new option. This allows developers to easily drop/reset the database without shutting down their app, etc. E.g.,

  $ RAILS_PG_DROP_FORCE=true bin/rails db:reset

### Additional information

My first attempt at this feature used the PostgreSQL function [pg_terminate_backend](https://www.postgresql.org/docs/current/functions-admin.html), similar to the [test helper](https://github.com/rails/rails/blob/44b6847ead440f8ab5b17e3a0036fb4e75c6e2d4/activerecord/test/support/adapter_helper.rb#L173). By using that function it would work for Postgres versions < 13; however, it felt significantly more complicated. As those versions have already reached EOL I didn't think the extra complexity was justified.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,24 @@
+*   PostgreSQL enable drop database FORCE option.
+
+    One of the benefits of developing with MySQL is that it allows dropping the
+    current database without first disconnecting clients. As a result developers
+    can use `bin/rails db:reset` and similar, without first shutting down
+    instances of the app, Rails consoles, background workers, etc. By default
+    PostgreSQL fails to drop a database when clients are connected and displays
+    the following error:
+
+      > PG::ObjectInUse: ERROR:  database "xyz" is being accessed by other users (PG::ObjectInUse)
+
+    This is frustrating when working in development where the database may be
+    dropped frequently.
+
+    PostgreSQL 13 added the `FORCE` option to the `DROP DATABASE` statement
+    ([PostgreSQL docs](https://www.postgresql.org/docs/current/sql-dropdatabase.html))
+    which automatically disconnects clients before dropping the database.
+    This option is automatically enabled for supported PostgreSQL versions.
+
+    *Steven Webb*
+
 *   Raise specific exception when a prohibited shard change is attempted.
 
     The new `ShardSwapProhibitedError` exception allows applications and
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -55,8 +55,15 @@ module ActiveRecord
         #
         # Example:
         #   drop_database 'matt_development'
-        def drop_database(name) # :nodoc:
-          execute "DROP DATABASE IF EXISTS #{quote_table_name(name)}"
+        #
+        # Note, for PostgreSQL versions >= 13 the SQL statement will include `WITH (FORCE)` to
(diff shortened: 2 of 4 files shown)


viralpraxis:
> My first attempt at this feature used the PostgreSQL function [pg_terminate_backend](https://www.postgresql.org/docs/current/functions-admin.html), similar to the [test helper](https://github.com/rails/rails/blob/44b6847ead440f8ab5b17e3a0036fb4e75c6e2d4/activerecord/test/support/adapter_helper.rb#L173). By using that function it would work for Postgres versions < 13; however, it felt significantly more complicated. As those versions have already reached EOL I didn't think the extra complexity was justified.

Just curious -- could you please clarify whether the `pg_terminate_backend` approach no longer works?

I've been using this for a couple years:

```shell
bundle exec rails dbconsole <<EOF
  SELECT COUNT(pg_terminate_backend(pg_stat_activity.pid)) "terminated connections"
  FROM pg_stat_activity
  WHERE datname = current_database()
    AND pid <> pg_backend_pid();
EOF
```

And it works fine (at least with PG17).

I guess it could be turned into a Rake task, like `bin/rails pg:terminate_connections db:reset`

swebb:
@viralpraxis using `pg_terminate_backed` did also work. It just felt more complex than using `WITH (FORCE)`.

A rake task would also work. Personally I like using an environment variable because it pairs well with [dotenv](https://github.com/bkeepers/dotenv). I can switch between apps using MySQL and PostgreSQL and `bin/rails db:reset` will work the same way.

skipkayhil:
I don't think an environment variable is the way to go here. To me it sounds like we should always be using `WITH (FORCE)`, except when the version of PostgreSQL doesn't support it (which we should version detect). Is there any other reason to not include it?

swebb:
> I don't think an environment variable is the way to go here. To me it sounds like we should always be using `WITH (FORCE)`, except when the version of PostgreSQL doesn't support it (which we should version detect). Is there any other reason to not include it?

Always using it when available sounds good to me. I'll make those changes.

I originally gated it behind an environment variable to be conservative and for a small amount of safety. If someone were to accidentally run `bin/rails db:drop` in production without this change the rake task would fail and they wouldn't lose any data. Practically speaking this feels like a pretty poor safety mechanism to me so I'm happy to go with always using it. Alternatively maybe I could ensure it only happens in development and test?

skipkayhil:
> If someone were to accidentally run `bin/rails db:drop` in production

This should be prevented already with the `check_protected_environments` task dependency https://github.com/rails/rails/blob/2c76861864814ec0a95ea13a6e786cab9b2f66cd/activerecord/lib/active_record/railties/databases.rake#L18-L20

swebb:
@skipkayhil I've updated my PR with your suggestions.

guilleiguaran (approved):

PR #55951, opened 2025-10-20
[RF-Docs] [ci-skip] Explain flash better in Getting Started guide

excid3:
### Motivation / Background

This improves the Getting Started guide by adding more explanation for the Flash and links to the Action Controller Overview for further reading.

### Detail

Instead of using the `notice` helper, using `flash` directly clarifies a bit more of how flash works.

While we're there, adding the `alert` flash in the Getting Started guide helps showcase it can be used with multiple names and simplifies future tutorials (and anyone who goes on side quests adding their own features).

This also improves the form partial to render errors inline instead than using a flash.now in #55914. Errors and alerts will be displayed in red so they stand out like notices as well.

### Additional information

This PR is an alternative to #55914

### Checklist

cc @AmandaPerino @p8 @zzak

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -1348,8 +1348,14 @@ We also want to replace any instance variables with a local variable, which we
 can define when we render the partial. We'll do this by replacing `@product`
 with `product`.
 
-```erb#1
+Let's also display any errors from the form submission inside the form.
+
+```erb#1-4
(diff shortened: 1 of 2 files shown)


harsh183:
Oh I like this better. Showing the errors in form is another cool rails thing to do and the flash explanation is good. Not a maintainer so I'll just give a lgtm in the comments and close out my PR for now

PR #55956, opened 2025-10-21
Getting Started guide (section 16.4 — Extracting a Concern) Doc oversight

Gabriel-Penteadoo:
The Getting Started guide (section 16.4 — Extracting a Concern) appears to be mixing two naming patterns:

It defines a namespaced module (`Product::Notifications`)

but then includes it as if it were a top-level module (`include Notifications`)

Should be a doc oversight, the correct line would be:

```ruby
include Product::Notifications
```

And that would match the file path and the constant definition.

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2381,7 +2381,7 @@ Notifications module.
 
 ```ruby#2
 class Product < ApplicationRecord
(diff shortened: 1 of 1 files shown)


p8:
Duplicate of: https://github.com/rails/rails/pull/55304

PR #55961, opened 2025-10-22
[ci skip] Small fix on the formatting of a code block in the 8.1 release notes.

sauloefo:
The missing blank line before the code block cause d rendering issues in the html page. I noticed the
 problem is not exhibited in the markdown preview.

**I'm aware of the policy about not merging small or cometic changes but the result of this bug was so bad (and easy to fix) that I thought it would worth to give it a try:**

## Result After Build

image

### DIFF
--- a/guides/source/8_1_release_notes.md
+++ b/guides/source/8_1_release_notes.md
@@ -94,6 +94,7 @@ end
 ```
 
 As well as context:
+
 ```ruby
 # All events will contain context: {request_id: "abc123", shop_id: 456}
 Rails.event.set_context(request_id: "abc123", shop_id: 456)


yahonda:
https://fc7128d2.rails-docs-preview.pages.dev/guides/8_1_release_notes#structured-event-reporting looks good. Thank you.

yahonda:
Backport to 8-1-stable branch via https://github.com/rails/rails/commit/6540e3cac4566c143618ecb9b543bd2d50653992

skipkayhil:
> I'm aware of the policy about not merging small or cometic changes

I think that mostly applies to code changes, this is definitely something worth fixing, thank you!

PR #55965, opened 2025-10-22
Document risks of using `.descendants`

alexanderadam:
### Motivation / Background

This Pull Request has been created because developers should be aware of the risks of using `Class.descendants`, particularly when using autoloading or in test environments with dynamically defined classes.

As discussed in [rubocop/rubocop-rails#1544](https://github.com/rubocop/rubocop-rails/issues/1544), these methods have two limitations:

1. They only return classes that have already been loaded, missing classes that haven't been autoloaded yet
2. They are non-deterministic with regards to Garbage Collection, especially problematic in test environments where classes are dynamically defined

Without proper documentation, developers may unknowingly use these methods in production code expecting complete and reliable results, leading to subtle bugs that are difficult to diagnose.

### Detail

This Pull Request changes the documentation for `Class#descendants`

### Additional information

Related discussion at https://github.com/github/rubocop-github/issues/110

### Checklist

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev
**PPS:** would you be so kind and add the `hacktoberfest-accepted` label to this issue in case you find that PR helpful? :pleading_face:

### DIFF
--- a/activesupport/lib/active_support/core_ext/class/subclasses.rb
+++ b/activesupport/lib/active_support/core_ext/class/subclasses.rb
@@ -16,6 +16,19 @@ class Class
   #
   #   class D < C; end
   #   C.descendants # => [B, A, D]
+  #
+  # WARNING: This method is unreliable in certain situations:
+  #
+  # * It only returns classes that have already been loaded. When using
+  #   autoloading, classes that haven't been referenced yet won't appear
+  #   in the results, even if their files exist in the codebase.
+  #
+  # * The results are non-deterministic with regards to Garbage Collection.
+  #   If you use this method in tests where classes are dynamically defined,
+  #   GC is unpredictable about when those classes are cleaned up and removed.
+  #
+  # Consider these limitations carefully when using this method, especially in
+  # production code.
   def descendants
     subclasses.concat(subclasses.flat_map(&:descendants))
   end
--- a/guides/source/active_support_core_extensions.md
+++ b/guides/source/active_support_core_extensions.md
@@ -1064,6 +1064,12 @@ C.descendants # => [B, A, D]
 
 The order in which these classes are returned is unspecified.
 
+WARNING: This method is unreliable in certain situations and should be used with caution:
+
+* **Autoloading**: It only returns classes that have already been loaded. When using autoloading (the default in Rails), classes that haven't been referenced yet won't appear in the results, even if their files exist in the codebase. This can lead to incomplete results in production code.
+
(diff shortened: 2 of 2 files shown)


guides/source/active_support_core_extensions.md
```diff
@@ -1040,6 +1040,8 @@ C.subclasses # => [B, D]
 
 The order in which these classes are returned is unspecified.
 
+WARNING: Like `descendants`, this method is unreliable in certain situations. It only returns classes that have already been loaded (relevant when using autoloading), and the results are non-deterministic with regards to Garbage Collection. See the `descendants` section below for more details.
```
skipkayhil: `Class#subclasses` actually isn't defined in Active Support anymore, its part of Ruby [3.1+](https://docs.ruby-lang.org/en/3.1/Class.html#method-i-subclasses).

Would you mind opening another PR that just removes this whole section since this file is only for documenting core extensions and `subclasses` isn't one? (I'm happy to if you don't want to)

alexanderadam: That's reasonable. I was wondering about that. :laughing: 
I removed that change from this PR here and created [the other one](https://github.com/rails/rails/pull/55966/files) as you requested.

activesupport/lib/active_support/core_ext/class/subclasses.rb
```diff
@@ -16,6 +16,19 @@ class Class
   #
   #   class D < C; end
   #   C.descendants # => [B, A, D]
+  #
+  # WARNING: This method is unreliable in certain situations:
+  #
+  # * It only returns classes that have already been loaded. When using
```
rafaelfranca, line 22: Those are also true for `subclasses` no? `descendants` is implemented using `subclasses` if `descendants` isn't unreliable, neither is `subclasses`. And in production none of those conditions are true, so saying "carefully when using this method, especially in production code." is wrong.

rafaelfranca:
Think about this, this kind of fear inducing documentation don't belong to Rails.

I'm using `subclasses` and `descendants` in production codebases for years and never had any of the problems described here.

Of course, trying to get a non-loaded object to appear in any collection is a wrong expectation. If the class isn't loaded, Ruby doesn't know about it, thus, it doesn't exist.

And yes, dynamic defined objects are removed by the garbage collector unless you hold a reference to it. That is just how GC works. This object being a class or a module doesn't matter.

Thank you for the contribution.

PR #55972, opened 2025-10-22
Mask secure tokens by default while displaying prefix

benngarcia:
Following up on #55822 - It makes sense to mask token values by default for a module named SecureToken.

Notably, this PR still respects prefix definitions.

* Before 
`user.inspect => #<User ..., token: token_tU9bLuZseefXQ4yQxQo8wjtB>`

* After
`user.inspect => #<User ..., token: token_[FILTERED]>`

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add `prefix:` and `skip_inspection_filter:` options to `has_secure_token` and filter token values by default.
+
+    *Benjamin Garcia*
+
 *   PostgreSQL enable drop database FORCE option.
 
     One of the benefits of developing with MySQL is that it allows dropping the
--- a/activerecord/lib/active_record/secure_token.rb
+++ b/activerecord/lib/active_record/secure_token.rb
@@ -50,7 +50,10 @@ module ActiveRecord
       #   An optional string prepended to the generated token. Intended primarily for keys that
       #   will be publicly visible (e.g. API keys, URLs, etc.) so that their purpose can be identified at a glance.
       #   The prefix does not count toward +:length+.
-      def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on, prefix: nil)
+      #
+      # [+:skip_inspection_filter]
+      #   Skip adding the column to the models <tt>filter_attributes</tt>. Defaults to false.
+      def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on, prefix: nil, skip_inspection_filter: false)
         if length < MINIMUM_TOKEN_LENGTH
           raise MinimumLengthError, "Token requires a minimum length of #{MINIMUM_TOKEN_LENGTH} characters."
         end
@@ -67,6 +70,10 @@ module ActiveRecord
           generate_token = -> { self.generate_unique_secure_token(length: length) }
         end
 
+        unless skip_inspection_filter
+          self.filter_attributes += [secure_token_filter_attribute(attribute, prefix)]
+        end
+
         # Load securerandom only when has_secure_token is used.
         require "active_support/core_ext/securerandom"
         define_method("regenerate_#{attribute}") { update! attribute => generate_token.call }
@@ -80,6 +87,19 @@ module ActiveRecord
       def generate_unique_secure_token(length: MINIMUM_TOKEN_LENGTH)
         SecureRandom.base58(length)
       end
+
+      private
+        def secure_token_filter_attribute(attribute, prefix)
+          return attribute unless prefix
+
+          ->(key, value) do
+            return unless key == attribute.to_s
+
+            value.replace(
+              "#{prefix}#{ActiveSupport::ParameterFilter::FILTERED}"
+            )
+          end
+        end
     end
   end
 end
--- a/activerecord/test/cases/secure_token_test.rb
+++ b/activerecord/test/cases/secure_token_test.rb
@@ -130,4 +130,19 @@ class SecureTokenTest < ActiveRecord::TestCase
     assert_match(/^token_/, user.token)
     assert_match(/^auth/, user.auth_token)
   end
+
+  def test_token_masked_on_inspect
+    model = Class.new(ActiveRecord::Base) do
+      self.table_name = "users"
+      attribute :auth_token
+      has_secure_token on: :initialize, prefix: true, skip_inspection_filter: true
+      has_secure_token :auth_token, on: :initialize, prefix: "auth_"
+    end
+
+    user = model.new
+
+    assert_not_includes user.inspect, "token: token_[FILTERED]"
+    assert_includes user.inspect, "auth_token: auth_[FILTERED]"
+    assert_equal 1, user.inspect.scan("[FILTERED]").length
+  end
 end


activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+*   Mask secure tokens by default while displaying prefix
+
+    Following up on behavior added in [e434aa4](https://github.com/rails/rails/commit/e434aa439a8ab8171cc63dbae821ca4f4da64623),
```
zzak: We generally don't reference other commits in the changelog, but also this can be considered one change which includes the other PR.

So can you make the changelog entry for both?

zzak:
Could you [squash your commits](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#squashing-commits) please?

benngarcia:
Done 👍

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Add `prefix:` option to `has_secure_token` and filter token values by default.
```
zzak: Should mention `skip_inspection_filter` option here too. :pray:

benngarcia: Done! Appreciate your reviews here :)

benngarcia:
Friendly ping — following up to see if anything else is needed from me.

activerecord/lib/active_record/secure_token.rb
```diff
@@ -50,7 +50,10 @@ module ClassMethods
       #   An optional string prepended to the generated token. Intended primarily for keys that
       #   will be publicly visible (e.g. API keys, URLs, etc.) so that their purpose can be identified at a glance.
       #   The prefix does not count toward +:length+.
-      def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on, prefix: nil)
+      #
+      # [+:skip_inspection_filter]
```
zzak, line 54: Maybe the name should be `skip_inspect_filter` instead 🤔

activerecord/lib/active_record/secure_token.rb
```diff
@@ -67,6 +70,11 @@ def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: Activ
           generate_token = -> { self.generate_unique_secure_token(length: length) }
         end
 
+        unless skip_inspection_filter
+          filter_attribute = prefix ? ->(key, value) { value.replace("#{prefix}#{ActiveSupport::ParameterFilter::FILTERED}") if key == attribute.to_s } : attribute
```
zzak: I think this line is a bit long with multiple conditions which makes it harder to read, if we are knit-picking

activerecord/lib/active_record/secure_token.rb
benngarcia: I actually gave this a pretty deep thought when first implementing and ultimately kept skip_inspection_filter because within `ActiveRecord::Core`, the relevant methods all reference an `inspection_filter`. I think the semantic consistency is more important.

~~Alternatively, we could also change to `skip_filter_attributes`~~. Not a huge fan of that actually.

activerecord/lib/active_record/secure_token.rb
benngarcia: I agree, it was just a surprisingly difficult tradeoff between redunandancy & readability given the multiple prefix conditionals in separate contexts, most of the other variations I could think of ended up having the same conditional being evaluated in multiple places nested or unnested within other conditionals, which I found harder to comprehend. I gave this good think and came up with this. I like this better so I'm updating the PR with it.

```ruby
def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on, prefix: nil, skip_inspection_filter: false) 
  if length < MINIMUM_TOKEN_LENGTH                                                                                                                            
    raise MinimumLengthError, "Token requires a minimum length of #{MINIMUM_TOKEN_LENGTH} characters."                                                        
  end                                                                                                                                                         
                                                                                                                                                              
  prefix = "#{attribute}_" if prefix == true                                                                                                                  
                                                                                                                                                            
  if prefix                                                                                                                                                   
    generate_token = -> do                                                                                                                                    
      token = self.generate_unique_secure_token(length: length)                                                                                               
                                                                                                                                                              
      "#{prefix}#{token}"                                                                                                                                     
    end                                                                                                                                                       
                                                                                                                                                              
    filter_attribute = ->(key, value) do                                                                                                                      
      if key == attribute.to_s                                                                                                                                
        value.replace("#{prefix}#{ActiveSupport::ParameterFilter::FILTERED}")                                                                                 
      end                                                                                                                                                     
    end                                                                                                                                                       
  else                                                                                                                                                        
    generate_token = -> { self.generate_unique_secure_token(length: length) }                                                                                 
    filter_attribute = attribute                                                                                                                              
  end                                                                                                                                                         
                                                                                                                                                              
  unless skip_inspection_filter                                                                                                                               
    self.filter_attributes += [filter_attribute]                                                                                                              
  end                                                                                                                                                         
                                                                                                                                                              
  # Load securerandom only when has_secure_token is used.                                                                                                     
  require "active_support/core_ext/securerandom"                                                                                                              
  define_method("regenerate_#{attribute}") { update! attribute => generate_token.call }                                                                       
  set_callback on, on == :initialize ? :after : :before do                                                                                                    
    if new_record? && !query_attribute(attribute)                                                                                                             
      send("#{attribute}=", generate_token.call)                                                                                                              
    end                                                                                                                                                       
  end                                                                                                                                                         
end                                                                                                                                                           

```

activerecord/lib/active_record/secure_token.rb
zzak: Yeah, if AR is using `inspection_filter` then lets keep it, thanks for thinking about it!

activerecord/lib/active_record/secure_token.rb
zzak: I see what you mean, I think there might be a path where we could define those procs as regular methods, but I don't know if that's any better.

What you have is an improvement over the initial PR, so I will mark this as resolved. Thanks!

benngarcia:
```ruby
      def has_secure_token(attribute = :token, length: MINIMUM_TOKEN_LENGTH, on: ActiveRecord.generate_secure_token_on, prefix: nil, skip_inspection_filter: false)
        if length < MINIMUM_TOKEN_LENGTH
          raise MinimumLengthError, "Token requires a minimum length of #{MINIMUM_TOKEN_LENGTH} characters."
        end

        prefix = "#{attribute}_" if prefix == true
        generate_token = -> do
          token = generate_unique_secure_token(length:)
          prefix ? "#{prefix}#{token}" : token
        end

        unless skip_inspection_filter
          self.filter_attributes += [secure_token_filter_attribute(attribute, prefix)]
        end

        # Load securerandom only when has_secure_token is used.
        require "active_support/core_ext/securerandom"
        define_method("regenerate_#{attribute}") { update! attribute => generate_token.call }
        set_callback on, on == :initialize ? :after : :before do
          if new_record? && !query_attribute(attribute)
            send("#{attribute}=", generate_token.call)
          end
        end
      end

      def generate_unique_secure_token(length: MINIMUM_TOKEN_LENGTH)
        SecureRandom.base58(length)
      end

      def secure_token_filter_attribute(attribute, prefix)
        return attribute unless prefix

        ->(key, value) do
          return unless key == attribute.to_s

          value.replace(
            "#{prefix}#{ActiveSupport::ParameterFilter::FILTERED}"
          )
        end
      end
```

I think something like this could be an improvement.

benngarcia:
Another friendly ping :)

PR #55978, opened 2025-10-22
Enable `config.asset_host` to read from environment by default

stevepolitodesign:
### Motivation / Background

This makes it so no code changes are needed in order to setup a CDN to
serve static assets.

### Detail

If `ENV["CDN_HOST"]` is not set, then `config.asset_host` will be set to
`nil`, meaning this commit does not change behavior.

### Additional information

Chose `ENV["CDN_HOST"]` to be consistent with the [example][] used in the
Guides.

[example]: https://guides.rubyonrails.org/asset_pipeline.html#set-up-a-cdn-to-serve-static-assets

### Checklist

-   [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
-   [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
-   [x] Tests are added or updated if you fix a bug or add a feature.
-   [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -21,7 +21,7 @@ Rails.application.configure do
   config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.asset_host = "http://assets.example.com"
+  config.asset_host = ENV["CDN_HOST"]
 
   <%- unless skip_active_storage? -%>
   # Store uploaded files on the local file system (see config/storage.yml for options).
--- a/actionmailbox/test/dummy/config/environments/production.rb
+++ b/actionmailbox/test/dummy/config/environments/production.rb
@@ -30,7 +30,7 @@ Rails.application.configure do
   config.assets.compile = false
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.asset_host = "http://assets.example.com"
+  config.asset_host = ENV["CDN_HOST"]
 
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
--- a/actiontext/test/dummy/config/environments/production.rb
+++ b/actiontext/test/dummy/config/environments/production.rb
@@ -30,7 +30,7 @@ Rails.application.configure do
   config.assets.compile = false
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.asset_host = "http://assets.example.com"
+  config.asset_host = ENV["CDN_HOST"]
 
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
--- a/activestorage/test/dummy/config/environments/production.rb
+++ b/activestorage/test/dummy/config/environments/production.rb
@@ -30,7 +30,7 @@ Rails.application.configure do
   config.assets.compile = false
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.asset_host = "http://assets.example.com"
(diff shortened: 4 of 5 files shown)


stevepolitodesign:
CC @seanpdoyle

railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -21,7 +21,7 @@ Rails.application.configure do
   config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.asset_host = "http://assets.example.com"
+  config.asset_host = ENV["CDN_HOST"]
```
stevepolitodesign, line 24: A less aggressive commit could be this:


```suggestion
  # config.asset_host = ENV["CDN_HOST"]
```

seanpdoyle: By default, `config.asset_host == nil`. To me, newly generated applications reading `nil` from an undefined environment variable feels like an improvement, since that provides applications with a hook to enable or disable (in the case of an outage, for example) a CDN without the need for a deployment or code change.

seanpdoyle:
This line appears throughout a variety of test suite files (based on https://github.com/rails/rails/commit/76e432341c5457372b5c38b62951ee89d43b470f):

```
# config.asset_host = "http://assets.example.com
```

Would it make sense to change those occurences to read from `ENV["CDN_HOST"]` as well?

railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
seanpdoyle: There is also mention of setting `config.asset_host` in the Action Mailer Basics guides:

https://github.com/rails/rails/blob/db82a005e422ea7461117eb49e7b242c7d327e27/guides/source/action_mailer_basics.md#L598

Should that include this same suggestion?

seanpdoyle (approved):

brunoprietog:
Is it even worth suggesting that in the configuration file? I'd suggest removing that suggestion entirely, considering that Rails comes with Thruster by default, and also, if you use a CDN like Cloudflare, the assets would already be served from the CDN without the need for another host. I know there are ideas for reducing the suggested configurations, and this could easily be one of them.

seanpdoyle:
> if you use a CDN like Cloudflare, the assets would already be served from the CDN without the need for another host.

In our experience, Cloudflare's universal caching for assets served from origin like `www.domainname.com` (as opposed to a CDN_HOST like `assets.domainname.com`). This can be troublesome for files served from `public/` (like `404.html`, `icon.png`, etc.). Files from `public/` are not fingerprinted, so aggressively caching them permanently can lead to issues when those assets change.

We've skipped Cloudflare's automatic caching for all assets served from origin (`www`), but instead permanently cache `assets` served assets. By encouraging the setting of `config.asset_hosts` to `CDN_HOST`, applications can make a similar decision: permanently cache assets and don't cache `public/` files.

guides/source/action_mailer_basics.md
```diff
@@ -595,7 +595,7 @@ Usually the `:asset_host` is consistent across the application, so you can
 configure it globally in `config/application.rb`:
 
 ```ruby
-config.action_mailer.asset_host = "http://example.com"
+config.action_mailer.asset_host = ENV["CDN_HOST"]
```
rafaelfranca: I don't think this one should change.

guides/source/action_view_helpers.md
```diff
@@ -471,7 +471,7 @@ typically in `config/environments/production.rb`.
 For example, let's say your asset host is `assets.example.com`:
 
 ```ruby
-config.asset_host = "assets.example.com"
+config.asset_host = ENV["CDN_HOST"]
```
rafaelfranca: I don't think this one should change

PR #55981, opened 2025-10-23
Correct WHERE condition in CTE test

vlado:
`posts_with_tags_and_comments` should filter on comments count instead of tags count. While the test still passed, this change aligns the query with the CTE’s intent and naming.

Introduced in https://github.com/rails/rails/pull/55918 when adding support for bound SQL literals in CTEs.


vlado:
@baarde FYI

PR #55987, opened 2025-10-23
Add tests for guides code

p8:
The guides have a lot of untested code, making it easy to break things.

This sets up the guides directory to also support tests. `guides/rails_guides.rb`
will no longer generate the guides by default so we can include it in tests
without generating the guides.

 The existing `test` task for bug_report_templates has been moved to a
namespace: `test:bug_report_templates`.

The w3c_validator script requires being run from a Rake task now as well
for consistency.

For the tests to run we'd also need to update the buildkite config.

### Checklist

### DIFF
--- a/guides/Rakefile
+++ b/guides/Rakefile
@@ -1,5 +1,14 @@
 # frozen_string_literal: true
 
+require "rake/testtask"
+
+Rake::TestTask.new do |t|
+  t.libs << "test"
+  t.test_files = FileList["test/**/*_test.rb"]
+  t.verbose = true
+  t.options = "--profile" if ENV["CI"]
+end
+
 namespace :guides do
   desc 'Generate guides (for authors), use ONLY=foo to process just "foo.md"'
   task generate: "generate:html"
@@ -7,7 +16,7 @@ namespace :guides do
   namespace :generate do
     desc "Generate HTML guides"
     task :html do
-      ruby "-Eutf-8:utf-8", "rails_guides.rb"
+      generate_guides
     end
 
     desc "Generate .mobi file"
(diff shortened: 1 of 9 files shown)


guides/Rakefile
```diff
@@ -7,7 +7,7 @@ namespace :guides do
   namespace :generate do
     desc "Generate HTML guides"
     task :html do
-      ruby "-Eutf-8:utf-8", "rails_guides.rb"
```
p8, line 10: I'm not sure if removing the encoding breaks things now.

guides/rails_guides/markdown.rb
```diff
@@ -4,7 +4,6 @@
 require "nokogiri"
 require "rails_guides/markdown/renderer"
 require "rails_guides/markdown/epub_renderer"
-require "rails-html-sanitizer"
```
p8, line 7: This require was resulting in a circular dependencies error.

guides/Rakefile
```diff
@@ -105,15 +115,37 @@ HELP
...
 end
 
-task :test do
-  templates = Dir.glob("bug_report_templates/*.rb")
-  counter = templates.count do |file|
-    puts "--- Running #{file}"
-    Bundler.unbundled_system(Gem.ruby, "-w", file) ||
-      puts("+++ 💥 FAILED (exit #{$?.exitstatus})")
+namespace :test do
+  task :bug_report_templates do
```
rafaelfranca: This isn't going to run on CI anymore. We need to make it run on CI by making `rake test` run the tests you added and those one. Maybe you can add as a dependency to the new `:test` tasks that was added.

p8: Sure!

rafaelfranca:
You also need to change the `preview_docs` task

guides/Rakefile
zzak: This is probably from before Ruby defaulted to utf8

zzak:
@p8 I actually don't see bug report tests running on CI anymore :eyes: 

Example: https://buildkite.com/rails/rails/builds/124548#019b0c8f-74ea-4ce6-a98c-098e82f8d500/1557

It takes 49s to start the container and finishes in 2s. On my machine `bin/test guides` over 80s to run.

PR #55990, opened 2025-10-23
Support 'remove_hidden_field_autocomplete' in the hidden_field helper

### LINKED ISSUE #55984: `autocomplete="off"` is not removed from hidden fields everywhere
The new option `action_view.remove_hidden_field_autocomplete` does mostly what it says on the tin, but calling `hidden_field` on a form object appears to be overlooked and still unconditionally produces an autocomplete attribute.

[offending code](https://github.com/rails/rails/blob/7061d801307ed15fab582fc8f139c7a145d5ee0d/actionview/lib/action_view/helpers/tags/hidden_field.rb#L8)
Previous issue & fix: https://github.com/rails/rails/issues/55209 https://github.com/rails/rails/pull/55336

### Steps to reproduce

```ruby
  config.action_view.remove_hidden_field_autocomplete = true
end

# in an erb template
<%= form_for :user do |form| %>
  <%= form.hidden_field :name %>
<% end %>
```

### Expected behavior

`<form><input type="hidden" name="user[name]" id="user_name" /></form>`

### Actual behavior

`<form><input autocomplete="off" type="hidden" name="user[name]" id="user_name" /></form>`

### System configuration

**Rails version**: 8.1.0

**Ruby version**: 3.4.4

codergeek121:
Fixes #55984 and adds unit tests for both configuration variants.

### DIFF
--- a/actionview/lib/action_view/helpers/tags/hidden_field.rb
+++ b/actionview/lib/action_view/helpers/tags/hidden_field.rb
@@ -5,7 +5,7 @@ module ActionView
     module Tags # :nodoc:
       class HiddenField < TextField # :nodoc:
         def render
-          @options.reverse_merge!(autocomplete: "off")
+          @options.reverse_merge!(autocomplete: "off") unless ActionView::Base.remove_hidden_field_autocomplete
           super
         end
       end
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thanks! See #55989. But I'll add you as co-author of that one.

PR #55991, opened 2025-10-23
Add AGENTS.md with Rails codebase guide for AI coding agents

rafaelfranca:
I just opened https://github.com/rails/rails/pull/55989 using GitHub Copilot exclusively, and it worked quite well. To help future AI coding agents contribute effectively to the Rails codebase, I've created a comprehensive guide in AGENTS.md based on my session.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,167 @@
+# Rails Codebase Guide for AI Coding Agents
+
+## Architecture Overview
+
+Rails is a **monorepo containing 10+ independent frameworks** that can work standalone or together. Each framework lives in its own directory at the root level:
+
+- **Active Record** (`activerecord/`) - ORM and database abstraction
+- **Action Pack** (`actionpack/`) - Controllers and routing (contains Action Controller and Action Dispatch)
+- **Action View** (`actionview/`) - View templates and helpers (extracted from Action Pack in Rails 3)
+- **Active Model** (`activemodel/`) - Model interfaces without database dependency
+- **Active Support** (`activesupport/`) - Core extensions and utilities used across all frameworks
+- **Action Mailer** (`actionmailer/`), **Action Mailbox** (`actionmailbox/`) - Email sending/receiving
+- **Active Job** (`activejob/`) - Background job abstraction
+- **Action Cable** (`actioncable/`) - WebSocket integration
+- **Active Storage** (`activestorage/`) - File uploads and cloud storage
+- **Action Text** (`actiontext/`) - Rich text content
+- **Railties** (`railties/`) - Rails CLI, generators, and framework glue
+
+**Key architectural principle**: Frameworks are loosely coupled. Changes to one framework should not break others unless there's an explicit dependency.
+
+## Testing Commands
+
+### Running Tests in a Framework
+```bash
+# From framework directory (preferred method)
+cd actionview && bin/test                    # Run all tests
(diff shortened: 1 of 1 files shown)


harsh183:
Should there be any link to instructions related to set up or does this work fine without?

rafaelfranca:
> Should there be any link to instructions related to set up or does this work fine without?

The devcontainer does the entire setup

fxn:
@rafaelfranca I think we have to have `AGENTS.md`.

Since you are busy these days, I took the liberty to do a few minor edits and will merge.

Of course, feel free to followup if you'd like to change anything.

fxn (approved):

seuros:
This file can be replaced with 1 line .

```
The workspace is the rails/rails source!
```

9 tokens instead of 1618 tokens.

Any model already knows the structure and the flows. It's all baked into training.

You get this for free when your codebase has consistency and real conventions over vibes.

fxn:
@seuros this was a first version, but indeed I believe it can be simplified. I believe we also need one per component. Plan to iterate in the next days.

seuros:
@fxn I get it. I'm not saying it is wrong. AGENTS.md is needed because the harness asks the model to look for it if it is not in the right place.

My point is simple: unless you introduce a new pattern or a new component, everything in this repo is already known by LLMs from training, even the flows and the commit styles.

I see this working with every pre-ai gems I maintain or use.

You can change the instructions however you want, but this is the most token-efficient and fastest AGENTS.md you can have.

Bonus: if the code hasn't moved or changed for years, the model can make precise changes using grep/sed without even opening the file.

I tried it 1-2 years ago and it still valid today. At first i was suspecting that the harness send everything in background.  It wasn't.

Open 2 worktree and try by yourself.

fxn:
@seuros since this is hard to test, my authoritative source is going to be
AGENTS.md docs themselves.

claasz:
https://www.heise.de/en/background/AGENTS-md-Helpful-agent-briefing-or-token-hog-11245317.html which is basically a summary of a study from ETH, has some support for @seuros position.

fxn:
@claasz Yep, @seuros and I exchanged some emails. I'll come back to this, thanks for that pointer!

simi:
Wouldn't be easier to just point to various files like https://github.com/rails/rails/blob/main/CONTRIBUTING.md and individual project READMEs to keep one source of docs working for both :people_holding_hands: and :robot:?

fxn:
@simi in theory, you have instructions for agents, and instructions for humans.

If you tell the agent to read that document, it will do, but might actually be inefficient, because the model already has (some version of) it.

The theory is that since all of that is available online, the model may not need that we reiterate any of that.

We need to find the right compromise.

I have to say that in my personal sessions I say "This is the repository of the Ruby on Rails web framework", and that is all.

simi:
> @simi in theory, you have instructions for agents, and instructions for humans.

@fxn I understand, but what's specific in current instructions to agents not useful for humans? :thinking: My fear is those "agentic specific instructions" are not going to be easily kept in sync with latest updated on "human docs". - in short, why to break DRY on docs level? :thinking:

seuros:
@simi 

Agents and humans are different identities . 

You need prose, they need **instructions** that don't fight against/invalidate their training data.

An agent already know rails/rails use mini test, it directory structure, it also know all it usage.

Humans needs more hand-holding.

PR #55992, opened 2025-10-23
ActiveJob/serializers: Only index new serializers

jsharpify:
Instead of reprocessing every serializer everytime a new serializer is added, just process the new one. This has the intended side-effect of only issuing the deprecation warning for non-compliant serializers once (when added) instead of on every subsequent call to add serializers. This also ensures that the location deperecation warning does not point to the innocent compliant serializers. Currently when it reports the deprecation it will include the location of the most recent call to add_serializers which falsely implicates those serializers.

### DIFF
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -51,30 +51,34 @@ module ActiveJob
       attr_reader :serializers
 
       def serializers=(serializers)
-        @serializers = serializers
-        index_serializers
+        @serializers_index.clear
+        @serializers = Set.new
+        add_new_serializers(serializers)
       end
 
       # Adds new serializers to a list of known serializers.
       def add_serializers(*new_serializers)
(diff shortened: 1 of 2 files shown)


skipkayhil (dismissed):

rafaelfranca:
CI is broken

activejob/lib/active_job/serializers.rb
```diff
@@ -52,29 +52,33 @@ def deserialize(argument)
 
       def serializers=(serializers)
         @serializers = serializers
-        index_serializers
+        @serializers_index.clear
+        add_new_serializers(serializers)
```
skipkayhil, line 56: `@serializers` is a Set, but aren't we adding them twice now?

jsharpify: Oops, yup. Emptying it out first and then assigning

PR #55993, opened 2025-10-23
Fallback DiskService URL options

JonathanFrias:
We should allow the DiskService to be configured

### Motivation / Background

There are many times when I have a small or simple setup one a single server, and I want to be able to test my active storage download url for example. This change allows for syntax like the following:

```ruby
  # config/environments/development.rb
  config.active_storage.url_options = { host: "example.com", protocol: "https", port: 443 }
  
  # Extends the same convention as the existing actionmailer gem
  config.action_mailer.default_url_options = { host: "example.com", port: 8443 }
```

This Pull Request has been created because it is often the case that I do not have nor want to setup a full s3 integration, and I think it makes sense to give users the ability to opt-into more complete functionality if they choose to. 

### Detail

This Pull Request changes the disk service url options to have a fallback to an environment specific configuration if defined.

### Additional information

This may not be the best implementation, it might make more sense to call `ActiveStorage::Current.url_options=(value)` during initialization.

This is what I have in my application as a workaround.

```ruby
  config.after_initialize do
    ActiveStorage::Current.attribute :url_options, default: { host: "example.com", protocol: "https", port: 443 }
  end
```

IMO I don't think I should put this snippet in an initializer because that spreads the configuration over multiple files without adding much. However this makes my environment inconsistent and more complicated that it should be.

We should consider that perhaps this DiskService should not depend on Rails directly, but we already have `url_helpers` method that depends on `Rails`, so I think this is not an issue. Users of activestorage without rails will have to always set `url_options` manually anyway.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/service/disk_service.rb
+++ b/activestorage/lib/active_storage/service/disk_service.rb
@@ -172,7 +172,7 @@ module ActiveStorage
       end
 
       def url_options
-        ActiveStorage::Current.url_options
+        ActiveStorage::Current.url_options || Rails.application.config.active_storage.url_options
       end
   end
 end


zzak:
At the moment I can't think of an alternative, but this change is not acceptable as is for the reasons you mentioned, thank you for your patch.

* DiskService should not depend on Rails directly

I doubt there is a strong argument for using a Rails engine outside of Rails, but there is a (stronger) argument for not reversing the dependency on config, it should be set by an initializer and dependents should call that initialized attribute.

This issue seems to relate to #39566, and is something that will be fixed when that is, or so I think.

PR #56000, opened 2025-10-24
Introduce a specific exception for unresolved models in AssociationReflection#compute_class

daniel-niknam:
### Motivation / Background

Follow up #55968

### Detail

`AssociationReflection#compute_class` previously raised a generic `ArgumentError` when it couldn’t
derive the model name. Callers that needed to recover (e.g., retrying lookup within a nested module
when a top-level constant shadows the model name) had to rescue `ArgumentError`, which risks masking
unrelated bugs that also manifest as `ArgumentError`.

This change defines and raises a dedicated exception for this case, allowing callers to rescue
precisely the failure they intend to handle and eliminating the need for a blank/broad `rescue`.

### Additional information

**Why**

* Avoids swallowing unrelated `ArgumentError`s.
* Makes the recovery path (nested module lookup) explicit and safe.
* Improves readability and intent, replacing a blanket rescue with targeted handling.

**What**

* Add a new error class for “unable to compute model” failures.
* Update `AssociationReflection#compute_class` to raise this specific error.
* Adjust call sites to rescue the new error only where fallback lookup is required.

**Behavior**

* No change to successful code paths.
* Failure is now surfaced via a specific, self-describing exception, reducing the chance of hiding genuine bugs.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/errors.rb
+++ b/activerecord/lib/active_record/associations/errors.rb
@@ -174,6 +174,9 @@ module ActiveRecord
     end
   end
 
+  class AssociationModelTypeError < ActiveRecordError # :nodoc:
+  end
+
   class ThroughCantAssociateThroughHasOneOrManyReflection < ActiveRecordError # :nodoc:
     def initialize(owner = nil, reflection = nil)
       if owner && reflection
--- a/activerecord/lib/active_record/reflection.rb
+++ b/activerecord/lib/active_record/reflection.rb
@@ -427,8 +427,10 @@ module ActiveRecord
         if active_record.name.demodulize == class_name
           begin
             return compute_class("::#{class_name}")
-          rescue
-            # Ignored
+          rescue NameError, AssociationModelTypeError
+            # If NameError or AssociationModelTypeError raised during above method call,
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/associations/errors.rb
```diff
@@ -174,6 +174,9 @@ def initialize(owner_class_name = nil, reflection = nil, through_reflection = ni
     end
   end
 
+  class AssociationModelTypeError < ActiveRecordError
```
zzak:
```suggestion
  class AssociationModelTypeError < ActiveRecordError # :nodoc:
```

daniel-niknam: Thanks @zzak

zzak: @daniel-niknam You can squash it, don't need to add me, thanks!

daniel-niknam:
I think I have a valid broken test 🤦

daniel-niknam:
@zzak I think it's ready for review again 🙏

zzak (approved):

PR #56001, opened 2025-10-24
[ci-skip][doc] Fix typos and errors in wishlists.md

hachi8833:
### Motivation / Background

This Pull Request has been created because wishlists.md contains some errors and typos that disturbs running tests.

### Detail

This Pull Request contains the following:

- Removing `has_many :subscribers, dependent: :destroy` from `Product` model because  it includes `Notifications` concern that already declares the `has_many: subscribers` association. Current declaration is redundant and inconsistent with the final `Product` model in [getting_started.md](https://guides.rubyonrails.org/getting_started.html#extracting-a-concern).
- Fixing incorrect filename `app/models/wishlist.rb` to `app/models/wishlist_product.rb`.
- Fixing incorrect path `store_products_path` to `store_wishlists_path`.
- Fixing incorrect call `pluralize "Subscriber", @subscribers.count` to `pluralize @subscribers.count, "Subscriber"`.
- Fixing incorrect assertion "T-Shirt is already on Second Wishlist." to "T-Shirt is already on Second.".

### Checklist

### DIFF
--- a/guides/source/wishlists.md
+++ b/guides/source/wishlists.md
@@ -142,7 +142,6 @@ Then in `app/models/product.rb`, add:
 class Product < ApplicationRecord
   include Notifications
 
-  has_many :subscribers, dependent: :destroy
   has_many :wishlist_products, dependent: :destroy
   has_many :wishlists, through: :wishlist_products
   has_one_attached :featured_image
@@ -159,7 +158,7 @@ SQL queries.
 We also set `wishlist_products` as `dependent: :destroy`. When a `Product` is
 destroyed, it will be automatically removed from any Wishlists.
 
-A counter cache stores the number of associated records to avoid running a separate query each time the count is needed. So in `app/models/wishlist.rb`, let's update both associations to enable counter
+A counter cache stores the number of associated records to avoid running a separate query each time the count is needed. So in `app/models/wishlist_product.rb`, let's update both associations to enable counter
 caching:
 
 ```ruby#2-5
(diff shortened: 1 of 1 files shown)


guides/source/wishlists.md
```diff
@@ -142,7 +142,6 @@ Then in `app/models/product.rb`, add:
 class Product < ApplicationRecord
   include Notifications
 
-  has_many :subscribers, dependent: :destroy
```
skipkayhil, line 145: Do the highlighted lines need to be updated?

hachi8833: I think so. In [getting_started.md](https://guides.rubyonrails.org/getting_started.html#extracting-a-concern), the concern `pp/models/product/notifications.rb` already has the one:

```ruby
module Product::Notifications
  extend ActiveSupport::Concern

  included do
    has_many :subscribers, dependent: :destroy # this should work
    after_update_commit :notify_subscribers, if: :back_in_stock?
  end
```

As far as I followed the steps in getting_started.md, the highlighted line `has_many :subscribers, dependent: :destroy` is unnecessary.

skipkayhil: I'm sorry, I should have been more specific: at the top of this code block is `rb#5-6` which causes lines 5 through 6 to be highlighted in the generated guide. Since we're removing a line shouldn't we update which line numbers should be highlighted as well?

hachi8833: Aha, now I get this😅 Yes, I should have also updated the highlight from 5-6 to 4-5. Just a moment.

PR #56005, opened 2025-10-24
Fix subsequent call to AS::JSON.encode after a nesting error

chaadow:
When a circular reference is triggered, all subsequent attempts to encode any ruby hash fail with the same `JSON::NestingError`

The fix for now is to initialize a new coder each time there is an encoding attempt.

A better fix probably would be on the `json` gem so that the same JSON::Coder instance can be called multiple times even when one call raises a `JSON::NestingError`

cc @byroot 

### Detail

More details can be found here https://github.com/rails/rails/commit/90616277e3d8fc46c9cf35d6a7470ff1ea0092f7#r168752947

### Additional information

In my production app, this made all of my Sidekiq jobs on the same ruby process fail, even internal ActiveStorage job such as preview jobs started failing.
I fixed the issue by restarting the dyno.

- But this issue is really wild. This means if for some reason, my code or any gems in my application triggers a circular reference ==> the entire JSON generation of valid ruby hashes would fail for the entire duration of the ruby process until shutdown or restart!

### Checklist

### DIFF
--- a/activesupport/lib/active_support/json/encoding.rb
+++ b/activesupport/lib/active_support/json/encoding.rb
@@ -149,25 +149,6 @@ module ActiveSupport
       if defined?(::JSON::Coder) && Gem::Version.new(::JSON::VERSION) >= Gem::Version.new("2.15")
         class JSONGemCoderEncoder # :nodoc:
           JSON_NATIVE_TYPES = [Hash, Array, Float, String, Symbol, Integer, NilClass, TrueClass, FalseClass, ::JSON::Fragment].freeze
-          CODER = ::JSON::Coder.new do |value, is_key|
-            json_value = value.as_json
-            # Keep compatibility by calling to_s on non-String keys
-            next json_value.to_s if is_key
-            # Handle objects returning self from as_json
-            if json_value.equal?(value)
-              next ::JSON::Fragment.new(::JSON.generate(json_value))
-            end
-            # Handle objects not returning JSON-native types from as_json
-            count = 5
-            until JSON_NATIVE_TYPES.include?(json_value.class)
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/json/encoding.rb
```diff
@@ -149,25 +149,6 @@ def stringify(jsonified)
       if defined?(::JSON::Coder) && Gem::Version.new(::JSON::VERSION) >= Gem::Version.new("2.15")
         class JSONGemCoderEncoder # :nodoc:
           JSON_NATIVE_TYPES = [Hash, Array, Float, String, Symbol, Integer, NilClass, TrueClass, FalseClass, ::JSON::Fragment].freeze
-          CODER = ::JSON::Coder.new do |value, is_key|
```
chaadow, line 152: Ideally the fix should be done on `JSON::Coder` to allow one global instance to reset its state when encountering an error

but in the meantime for rails users on version `2.15` of the json gem i see two solutions :

1- Either delete/comment this entire if branch until it's fixed on the `json`
2- Or initialiaze a new `JSON::Coder` instance everytime we try to encode/dump

byroot:
Thank you, but the proper fix has to be in the json gem: https://github.com/ruby/json/pull/874

byroot:
`json` `2.15.2` is released with the fix.

chaadow:
thanks @byroot I'm going to open another PR to change this: 

- change this line: `if defined?(::JSON::Coder) && Gem::Version.new(::JSON::VERSION) >= Gem::Version.new("2.15.2")` to be strict about using `2.15.2`
- update Gemfile.lock

PR #56006, opened 2025-10-24
Don't destroy ActiveStorage::Blob when its file deletion on the service fails

sherpalabsio:
### Motivation / Background

To fix the following scenario:
- Call `ActiveStorage::Blob#purge_later` which calls `ActiveStorage::Blob#purge`.
- System removes the blob record from the DB.
- System tries to remove the file from the storage (`ActiveStorage::Blob#delete`).
  - It crashes, e.g., with a `Google::Cloud::UnavailableError`.
- File stays in the storage, the DB record gallops to the sunset.
- The file in the storage becomes orphaned and difficult to remove when we have hundreds of terabytes of data in the storage.
- Retrying the job won't help because the blob record got deleted in the first run that failed halfway

### Details

This Pull Request allows removing `ActiveStorage::Blob` records only after their file in the service got removed successfully.

The provided solution
- Checks if the blob can be destroyed.
- Deletes its file from the service.
- If the file deletion is successful, removes the blob record from the DB.

The `with_no_destroy_preconditions_check` method is added to avoid sending duplicate DB queries caused by `attachments.exists?`.

Alternative solution would be to wrap `destroy` and `delete` calls in a transaction, like below:

```rb
def purge
    destroy
    delete if previously_persisted?
  end
rescue ActiveRecord::InvalidForeignKey
end
```

I'd go with the transactional solution if there is no argument against it.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -3,4 +3,12 @@
 
     *Bart de Water*
 
+*   `ActiveStorage::Blob#purge` won't delete blob records when their file deletion on the service fails.
+    This prevents dangling files on the storage.
+
+    For example, GCS sometimes fails with `Google::Cloud::UnavailableError`. In such cases, we want to keep the blob
+    record intact so that we can retry deletion later.
+
+    *Peter Toth*
+
 Please check [8-1-stable](https://github.com/rails/rails/blob/8-1-stable/activestorage/CHANGELOG.md) for previous changes.
--- a/activestorage/app/models/active_storage/blob.rb
+++ b/activestorage/app/models/active_storage/blob.rb
@@ -45,9 +45,7 @@ class ActiveStorage::Blob < ActiveStorage::Record
 
   after_update_commit :update_service_metadata, if: -> { content_type_previously_changed? || metadata_previously_changed? }
 
-  before_destroy(prepend: true) do
-    raise ActiveRecord::InvalidForeignKey if attachments.exists?
-  end
+  before_destroy(prepend: true) { check_destroy_preconditions! }
 
   validates :service_name, presence: true
   validates :checksum, presence: true, unless: :composed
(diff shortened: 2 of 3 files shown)


guilleiguaran:
We need a CHANGELOG entry for the change.

sherpalabsio:
CHANGELOG  entry is added.
How about using a transaction instead of the current solution?
Should I backport the final solution to older versions?

zzak:
What about retrying on `UnavailableError`?

sherpalabsio:
The retry won't find the blob record because it got deleted by the first run that failed halfway.
No blob record -> no key -> no way to know what file to remove from the service.
Added this to the PR desc.

zzak:
Maybe a transaction or after_commit, probably holding open a transaction waiting on gcs to delete the file is not preferrable though.

sherpalabsio:
Well, I see only two options: either we run it in a transaction or we delete the file from the service before deleting it from the DB. The current diff is for the 2nd.

I think it boils down to whether we want a simpler, possibly less performant or a bit more complex, possibly a bit more performant solution.

The current solution has some indirections which makes it more complicated but we need that to support the built in safety mechanism to check if the blob is still attached.

On the other hand the transaction version would be very simple. But if the service is slow it can keep the transaction open. Ideally the job should be concurrency limited and run in off hours so not sure how big problem this would cause.

Side story: I already monkey patched ActiveStorage::Blob#purge in my work to make it run in a transaction. We have just finished cleaning up ~30 million unattached blobs in a few days with this setup. The ActiveStorage::PurgeJob jobs ran for ~200ms on average, the HTTP delete requests to GCS took ~100ms. This means that we kept the transaction open for a bit longer than 100ms. The amount of jobs put into the system was throttled. It was working out fine.

zzak:
I'm thinking about this and wrapping it in a transaction is probably not what we want, this can lead to many other complications. This approach seems reasonable, in that we shouldn't delete the record if file deletion fails, but it's also a change that might warrant a config to decide whether or not we want to roll this out immediately -- and also should the behavior change if being run inside of `purge_later` or not.

@sherpalabsio Since this issue is currently affecting you, I would appreciate if you could investigate this a bit more :bow:

PR #56008, opened 2025-10-25
Use strict version of `json` gem greater than or equal to 2.15.2

chaadow:
related to #56005 

I feel we need to be stricter on the JSON version here

cc @byroot

### DIFF
--- a/activesupport/test/json/encoding_test.rb
+++ b/activesupport/test/json/encoding_test.rb
@@ -520,6 +520,23 @@ EXPECTED
     end
   end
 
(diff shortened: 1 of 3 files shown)


activesupport/test/json/encoding_test.rb
```diff
@@ -520,6 +520,25 @@ def test_as_json_too_recursive
     end
   end
 
+  if defined?(::JSON::Coder) && Gem::Version.new(::JSON::VERSION) >= Gem::Version.new("2.15.2")
```
chaadow: Since I've updated the lock, I don't know if we need to keep testing this?

PR #56012, opened 2025-10-26
Memoize IAM client and set authorization to ADC for GCS Active Storage service

jmalcic:
### Motivation / Background

This Pull Request has been created because #55353 breaks signed URLs for applications using the default ADC auth behaviour.

### Detail

This Pull Request restores the previous behaviour where ADC is used when `iam: true` is set, but avoids excessive ADC calls by memoising the client. This means that calls are only made at instantiation time and when the credentials eventually expire (which the client takes care of). It also still allows overriding the auth used by either the mechanism suggested in #55353 (using `Google::Apis::RequestOptions.default.authorization`, which affects all clients) or by setting the `authorization` value on the IAM client directly.

This is necessary because the assumption in #55353 that the libraries will request auth themselves with the right scope isn't true in practice. So e.g [this test](https://github.com/rails/rails/blob/4a753c2f79a9c41ec3ab5551724599193da180e5/activestorage/test/service/gcs_service_test.rb#L193) when run locally and where `GOOGLE_APPLICATION_CREDENTIALS` is set to the path to a service account key will fail with `Google::Apis::AuthorizationError: Unauthorized`. This is also what we have seen in production.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/service/gcs_service.rb
+++ b/activestorage/lib/active_storage/service/gcs_service.rb
@@ -152,6 +152,21 @@ module ActiveStorage
       @client ||= Google::Cloud::Storage.new(**@config.except(:bucket, :cache_control, :iam, :gsa_email))
     end
 
+    # Returns the IAM client used for direct uploads and signed URLs.
+    # By default, the authorization for the IAM client is set to
+    # {Application Default Credentials}[https://docs.cloud.google.com/docs/authentication/application-default-credentials],
+    # fetched once at instantiation time, then refreshed automatically when expired.
+    # This can be set to a different value to use other authorization methods.
+    #
+    #   ActiveStorage::Blob.service.iam_client.authorization = Google::Auth::ImpersonatedServiceAccountCredentials.new(options)
+    def iam_client
+      @iam_client ||= Google::Apis::IamcredentialsV1::IAMCredentialsService.new.tap do |client|
+        client.authorization ||= Google::Auth.get_application_default(["https://www.googleapis.com/auth/iam"])
+      rescue
+        nil
+      end
+    end
+
     private
       def private_url(key, expires_in:, filename:, content_type:, disposition:, **)
         args = {
@@ -208,19 +223,6 @@ module ActiveStorage
       def signer
         # https://googleapis.dev/ruby/google-cloud-storage/latest/Google/Cloud/Storage/Project.html#signed_url-instance_method
         lambda do |string_to_sign|
-          iam_client = Google::Apis::IamcredentialsV1::IAMCredentialsService.new
-
-          # We explicitly do not set iam_client.authorization so that it uses the
-          # credentials set by the application at Google::Apis::RequestOptions.default.authorization.
-          # If the application does not set it, the GCP libraries will automatically
-          # determine it on each call. This code previously explicitly set the
-          # authorization to Google::Auth.get_application_default which triggers
-          # an explicit call to the metadata server - given this lambda is called
-          # for a significant number of file operations, it can lead to considerable
-          # tail latencies and even metadata server overloads. Additionally, that
-          # prevented applications from being able to configure the credentials
-          # used to perform the signature operation.
-
           request = Google::Apis::IamcredentialsV1::SignBlobRequest.new(
             payload: string_to_sign
           )
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,20 @@
+*   Restore ADC when signing URLs with IAM for GCS
+
+    ADC was previously used for automatic authorization when signing URLs with IAM.
+    Now it is again, but the auth client is memoized so that new credentials are only
+    requested when the current ones expire. Other auth methods can now be used
+    instead by setting the authorization on `ActiveStorage::Service::GCSService#iam_client`.
+
+    ```ruby
+    ActiveStorage::Blob.service.iam_client.authorization = Google::Auth::ImpersonatedServiceAccountCredentials.new(options)
+    ```
+
+    This is safer than setting `Google::Apis::RequestOptions.default.authorization`
+    because it only applies to Active Storage and does not affect other Google API
+    clients.
+
+    *Justin Malčić*
+
 *   Move responsibility for checksums storage service
 
     The storage service should implement calculating and
--- a/activestorage/test/service/gcs_service_test.rb
+++ b/activestorage/test/service/gcs_service_test.rb
@@ -198,6 +198,37 @@ if SERVICE_CONFIGURATIONS[:gcs]
         assert_match(/storage\.googleapis\.com\/.*response-content-disposition=inline.*test\.txt.*response-content-type=text%2Fplain/,
           service.url(key, expires_in: 2.minutes, disposition: :inline, filename: ActiveStorage::Filename.new("test.txt"), content_type: "text/plain"))
       end
+
+      test "default IAM client ADC" do
+        WebMock.enable!
+        config_with_adc = { gcs: SERVICE_CONFIGURATIONS[:gcs].merge({ iam: true }) }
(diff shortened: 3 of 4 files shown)


zzak:
It took me a bit, but I was able to produce this error in the test with a service account using IAM:

```
Error:
ActiveStorage::Service::GCSServiceTest#test_direct_upload_with_IAM_signing:
Google::Apis::AuthorizationError: Unauthorized
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:235:in 'Google::Apis::Core::HttpCommand#check_status'
    google-apis-core-0.15.1/lib/google/apis/core/api_command.rb:137:in 'Google::Apis::Core::ApiCommand#check_status'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:207:in 'Google::Apis::Core::HttpCommand#process_response'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:326:in 'Google::Apis::Core::HttpCommand#execute_once'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:131:in 'block (2 levels) in Google::Apis::Core::HttpCommand#do_retry'
    retriable-3.1.2/lib/retriable.rb:61:in 'block in Retriable.retriable'
    retriable-3.1.2/lib/retriable.rb:56:in 'Integer#times'
    retriable-3.1.2/lib/retriable.rb:56:in 'Retriable.retriable'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:128:in 'block in Google::Apis::Core::HttpCommand#do_retry'
    retriable-3.1.2/lib/retriable.rb:61:in 'block in Retriable.retriable'
    retriable-3.1.2/lib/retriable.rb:56:in 'Integer#times'
    retriable-3.1.2/lib/retriable.rb:56:in 'Retriable.retriable'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:118:in 'Google::Apis::Core::HttpCommand#do_retry'
    google-apis-core-0.15.1/lib/google/apis/core/http_command.rb:109:in 'Google::Apis::Core::HttpCommand#execute'
    google-apis-core-0.15.1/lib/google/apis/core/base_service.rb:477:in 'Google::Apis::Core::BaseService#execute_or_queue_command'
    google-apis-iamcredentials_v1-0.22.0/lib/google/apis/iamcredentials_v1/service.rb:191:in 'Google::Apis::IamcredentialsV1::IAMCredentialsService#sign_service_account_blob'
    lib/active_storage/service/gcs_service.rb:239:in 'block in ActiveStorage::Service::GCSService#signer'
    google-cloud-storage-1.54.0/lib/google/cloud/storage/file/signer_v2.rb:136:in 'Google::Cloud::Storage::File::SignerV2#generate_signature'
    google-cloud-storage-1.54.0/lib/google/cloud/storage/file/signer_v2.rb:129:in 'Google::Cloud::Storage::File::SignerV2#signed_url'
    google-cloud-storage-1.54.0/lib/google/cloud/storage/bucket.rb:2253:in 'Google::Cloud::Storage::Bucket#signed_url'
    lib/active_storage/service/gcs_service.rb:120:in 'block in ActiveStorage::Service::GCSService#url_for_direct_upload'
    activesupport/lib/active_support/notifications.rb:212:in 'ActiveSupport::Notifications.instrument'
    lib/active_storage/service.rb:169:in 'ActiveStorage::Service#instrument'
    lib/active_storage/service/gcs_service.rb:93:in 'ActiveStorage::Service::GCSService#url_for_direct_upload'
    test/service/gcs_service_test.rb:177:in 'block in <class:GCSServiceTest>'


bin/test test/service/gcs_service_test.rb:170
```

@jmalcic Is that what you mean?

I have this in my services config:

```yaml
gcs:
  service: GCS
  iam: true
  gsa_email: "[X]@[Y].iam.gserviceaccount.com"
  project: [REDACTED]
  bucket: [REDACTED]
```

```
# .envrc
export $GOOGLE_APPLICATION_CREDENTIALS=path/to/gcs/sa.json
```

```json
{
  "type": "service_account",
  "project_id": "REDACTED",
  "private_key_id": "REDACTED",
  "private_key": "REDACTED",
  "client_email": "X@Y.iam.gserviceaccount.com",
  "client_id": "REDACTED",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/X%40Y.iam.gserviceaccount.com",
  "universe_domain": "googleapis.com"
}
```

jmalcic:
> @jmalcic Is that what you mean?

Yes exactly! I was interested in signed URLs, but direct uploads are also affected.

https://github.com/rails/rails/blob/4a753c2f79a9c41ec3ab5551724599193da180e5/activestorage/test/service/gcs_service_test.rb#L170-L200

Is the test still failing for you with the change?

zzak:
@jmalcic Thanks, no I could produce the error in both those tests and fixed with this patch.

activestorage/lib/active_storage/service/gcs_service.rb
```diff
@@ -152,6 +152,15 @@ def client
       @client ||= Google::Cloud::Storage.new(**@config.except(:bucket, :cache_control, :iam, :gsa_email))
     end
 
+    def iam_client
```
zzak, line 162: @jmalcic I think this should be private though, not that user needs to change it, wdyt?

jmalcic: The user does need to be able to change the `#authorization` inside if they want to use other credentials, and modifying `Google::Apis::RequestOptions.default.authorization` to achieve this is not great because it's global. If the client needs to be private I'd want to add a public writer on the service e.g. `#iam_client_authorization=`

zzak: I guess we make the client public so giving the iam_client a reader so you can modify it's internals is maybe acceptable?

jmalcic: That had been my thinking yeah. I will also add a rescue in case ADC isn't set up at all (then you need to deal with authorization yourself because it can't be inferred)

jmalcic: @zzak any more thoughts on this?

zzak: Yeah, I think we should add a test for this or at least could you explain how you will use this when it happens. Also, it would be good to get feedback from folks like @drcapulet who added this in #55353.

kimihiko-nakamura:
@jmalcic thank you for the work!

I came through #56391 and tested your branch, in GKE iam: true environment.

I could confirm creating url which failed on 8.1 works with your branch 👍 

Hope this info helps

activestorage/lib/active_storage/service/gcs_service.rb
jmalcic: I've added tests for ADC and manually configured auth

palexvs:
Had the same issue on prod, any chance this PR could be merged to unblock Rails upgrade?

eileencodes (changes requested):
Thanks for the PR, it needs some work before it can be merged.

Can you add a changelog? Since this method is public, it also needs documentation (comments inside a method won't be picked up by rdoc).

CI is broken -
`/usr/local/lib/ruby/4.0.0/bundled_gems.rb:60:in 'Kernel.require': cannot load such file -- minitest/mock (LoadError)` on most builds, and the rubocop build is also failing.

eileencodes:
Thanks for quickly fixing the changes.

eileencodes:
Backported to 8-1-stable in 528351457f6aa2a472c22b6cf481bea7d2621e0b

PR #56015, opened 2025-10-27
Fix strict_loading!(false) to override the strict_loading option in macro

### LINKED ISSUE #56014: Can't disable strict loading if strict_loading option is specified in the association macro
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end

    t.integer :post_id
  end
end

  self.strict_loading_by_default = true

  has_many :comments
  has_many :strict_loading_comments, strict_loading: true, class_name: "Comment"
end

  belongs_to :post
end

class Rating < ActiveRecord::Base
  belongs_to :post
end

  def test_association_stuff
    post = Post.create!

    assert_raises(ActiveRecord::StrictLoadingViolationError) { post.comments.to_a }
    assert_raises(ActiveRecord::StrictLoadingViolationError) { post.strict_loading_comments.to_a }

    post.strict_loading!(false)
    assert_nothing_raised { post.comments.to_a }
    assert_nothing_raised { post.strict_loading_comments.to_a }
  end
end
```

### Expected behavior

After calling `strict_loading!(false)`, `ActiveRecord::StrictLoadingViolationError` should not be raised for associations with `strict_loading: true`.

### Actual behavior

Even after calling `strict_loading!(false)`, accessing associations with `strict_loading: true` still raises `ActiveRecord::StrictLoadingViolationError`.

### System configuration

**Rails version**: Rails 8.1.0

shouichi:
Problem:

When `strict_loading: true` is specified in has_many/has_one macro, calling `strict_loading!(false)` does not override the option.

Closes https://github.com/rails/rails/issues/56014.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -286,6 +286,9 @@ module ActiveRecord
 
           return unless owner.validation_context.nil?
 
+          strict_loading = owner.instance_eval { @strict_loading }
+          return strict_loading && !owner.strict_loading_n_plus_one_only? unless strict_loading.nil?
+
           return reflection.strict_loading? if reflection.options.key?(:strict_loading)
 
           owner.strict_loading? && !owner.strict_loading_n_plus_one_only?
--- a/activerecord/lib/active_record/core.rb
+++ b/activerecord/lib/active_record/core.rb
@@ -696,7 +696,9 @@ module ActiveRecord
 
     # Returns +true+ if the record is in strict_loading mode.
     def strict_loading?
-      @strict_loading
+      return @strict_loading if defined?(@strict_loading)
+
+      self.class.strict_loading_by_default
     end
 
     # Sets the record to strict_loading mode. This will raise an error
@@ -855,7 +857,6 @@ module ActiveRecord
         klass = self.class
 
         @primary_key         = klass.primary_key
-        @strict_loading      = klass.strict_loading_by_default
         @strict_loading_mode = klass.strict_loading_mode
 
         klass.define_attribute_methods
--- a/activerecord/test/cases/strict_loading_test.rb
+++ b/activerecord/test/cases/strict_loading_test.rb
@@ -499,6 +499,16 @@ class StrictLoadingTest < ActiveRecord::TestCase
     assert_nothing_raised { developer.strict_loading_mentor }
   end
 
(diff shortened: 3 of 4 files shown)


tahsin352 (commented):
activestorage attachment tests for strict loading are failing??

bhushangahire:
Took a look at this — the root-cause analysis matches the fix, but I have a few concerns and a small gap to flag:

**1. `owner.instance_eval { @strict_loading }` in `association.rb`**

Reaching into the owner with `instance_eval` to read an ivar feels like a code smell — especially now that `strict_loading?` itself ([`core.rb:698`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/core.rb#L698)) is the canonical accessor. The reason `instance_eval` is needed seems to be that callers need to distinguish "explicitly set" from "never set" (so they can fall through to the macro-level `reflection.strict_loading?`). Could that be exposed as a small private predicate on the owner (e.g. `strict_loading_set?`) instead, so `Association` doesn't poke at internals? Same effect, cleaner boundary.

**2. Missing `has_many` test**

The original issue (#56014) reproduces with `has_many` — the new tests only cover `belongs_to` and `has_one`. Could you add the `has_many` case to make sure all three macros are exercised?

**3. Unrelated activestorage test change**

```diff
-    @user.strict_loading!(true)
-    @user.avatar.attach(blob)
+    with_strict_loading_by_default do
+      @user.avatar.attach(blob)
+    end
```

This swaps instance-level strict loading for class-level default, which tests a different code path than the original assertion. If the test still passes with the simpler instance-level form under your fix, it'd be safer to leave it unchanged. If it doesn't, that's worth understanding before merging — could indicate the fix has a subtler behavior change. Why the swap?

**4. Asymmetric init**

[`init_internals`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/core.rb#L853) now lazy-initializes `@strict_loading` but still eagerly initializes `@strict_loading_mode` on the next line. Intentional? If both should follow the same lazy pattern for consistency it'd be a quick follow-up.

I haven't run the AR test suite locally yet, but the logic in `core.rb` and the `n_plus_one_only` interaction in `association.rb` both read correctly to me.

PR #56017, opened 2025-10-27
Add `Rails::CodeStatistics#register_extension` to register file extensions for `rails stats`

taketo1113:
### Motivation / Background

In the following Pull Request, support was added for customizing the file extensions included in the `rails stats` command by overriding `Rails::CodeStatistics.pattern`:
- https://github.com/rails/rails/pull/55734

However, since this approach overrides `Rails::CodeStatistics.pattern`, any future changes to the default `PATTERN` cannot be detected or followed automatically.
I'd like to make it possible to add extensions while keeping compatibility with future updates to the default pattern.

Below is an example from Rails v8.1.0 showing how to add the `.slim` extension:
```ruby
# default value
# PATTERN = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb)$/

#  override value to add the `slim` extension
Rails::CodeStatistics.pattern = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb|slim)$/
```

If a new default extension (e.g., `.txt`) is added in the future, we would ideally like the updated default pattern plus our custom extensions.
However, because the entire pattern is overridden, this is not currently possible:
```ruby
# default value after adding the `txt` extension
# PATTERN = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb|txt)$/

# actual behavior: .txt is not included
Rails::CodeStatistics.pattern = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb|slim)$/

# expected behavior: .txt is included
Rails::CodeStatistics.pattern = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb|txt|slim)$/
```

In addition, due to this behavior, when a third-party gem such as the `slim-rails` gem tries to add an extension, it ends up overwriting the existing `pattern` and therefore cannot actually add it.

### Detail

This Pull Request proposes adding a `register_extension` method to `CodeStatistics` for adding file extensions.
By using `register_extension`, users can add extensions without needing to override `Rails::CodeStatistics.pattern`, ensuring compatibility with future updates to the default pattern.

```ruby
Rails::CodeStatistics.register_extension("slim")
```

With `register_extension`, file extensions can be added to the default pattern, enabling third-party gems such as the `slim-rails` gem to register their own extensions.

Note that the existing `Rails::CodeStatistics.pattern` will continue to work.

### Additional information

Although the new `register_extension` method could eventually make `Rails::CodeStatistics.pattern` unnecessary, it is already available in Rails 8.1.0 and may be in use by existing users.
The `Rails::CodeStatistics.pattern` feature was originally introduced in the following Pull Request to allow overriding extensions such as `haml`
- https://github.com/rails/rails/pull/55734

### Checklist

### DIFF
--- a/railties/lib/rails/code_statistics.rb
+++ b/railties/lib/rails/code_statistics.rb
@@ -42,10 +42,13 @@ module Rails
 
     HEADERS = { lines: " Lines", code_lines: "   LOC", classes: "Classes", methods: "Methods" }
 
-    PATTERN = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb)$/
+    EXTENSIONS = %w[rb js ts css scss coffee rake erb]
+
+    PATTERN = /^(?!\.).*?\.(#{EXTENSIONS.join("|")})$/
 
     class_attribute :directories, default: DIRECTORIES
     class_attribute :test_types, default: TEST_TYPES
+    class_attribute :extensions, default: EXTENSIONS
     class_attribute :pattern, default: PATTERN
 
     # Add directories to the output of the <tt>bin/rails stats</tt> command.
@@ -58,6 +61,14 @@ module Rails
     def self.register_directory(label, path, test_directory: false)
       self.directories << [label, path]
       self.test_types << label if test_directory
+    end
+
+    # Add extensions to the output of the <tt>bin/rails stats</tt> command.
+    #
+    #   Rails::CodeStatistics.register_extension("txt")
+    def self.register_extension(extension)
+      self.extensions += [extension]
+      self.pattern = /^(?!\.).*?\.(#{Regexp.union(extensions.map { |e| Regexp.escape(e) })})$/
     end
 
     def initialize(*pairs)
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add `Rails::CodeStatistics#register_extension` to register file extensions for `rails stats`
+
+    ```ruby
+    Rails::CodeStatistics.register_extension("txt")
+    ```
+
+    *Taketo Takashima*
+
 *   Wrap console command with an executor by default
 
     This can be disabled with `-w` or `--skip_executor`, same as runner.
--- a/railties/test/code_statistics_test.rb
+++ b/railties/test/code_statistics_test.rb
@@ -29,6 +29,14 @@ class CodeStatisticsTest < ActiveSupport::TestCase
     Rails::CodeStatistics.test_types.delete("Model specs")
   end
 
+  test "register test extensions" do
+    Rails::CodeStatistics.register_extension("slim")
+    assert Rails::CodeStatistics.pattern.to_s.include?("slim")
+  ensure
+    Rails::CodeStatistics.extensions = Rails::CodeStatistics::EXTENSIONS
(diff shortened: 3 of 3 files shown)


railties/lib/rails/code_statistics.rb
```diff
@@ -42,10 +42,20 @@ class CodeStatistics
...
 
-    PATTERN = /^(?!\.).*?\.(rb|js|ts|css|scss|coffee|rake|erb)$/
+    EXTENSIONS = ["rb",
+                  "js",
+                  "ts",
+                  "css",
+                  "scss",
+                  "coffee",
+                  "rake",
+                  "erb"]
```
zzak:
```suggestion
    EXTENSIONS = %w[rb js ts css scss coffee rake erb]
```

taketo1113: @zzak Thank you for reviewing.
I have updated it!

taketo1113:
I have fixed the conflict in CHANGELOG.

zzak:
Would be curious to hear what @p8 thinks of this

taketo1113:
@byroot Since you adopted #55734, what do you think about this Pull Request?

I would like to make it possible to support cases where both `slim-rails` and `haml-rails` add their own  custom extensions for `rails stats`.
Currently, when a 3rd-party gem adds custom extensions, it ends up overriding `Rails::CodeStatistics.pattern`, so only one of the added extensions is actually used.

railties/lib/rails/code_statistics.rb
```diff
@@ -60,6 +63,14 @@ def self.register_directory(label, path, test_directory: false)
       self.test_types << label if test_directory
     end
 
+    # Add extensions to the output of the <tt>bin/rails stats</tt> command.
+    #
+    #   Rails::CodeStatistics.register_extension("txt")
+    def self.register_extension(extension)
+      self.extensions << extension
```
byroot: I think we end up mutating the default here.

railties/lib/rails/code_statistics.rb
```diff
@@ -60,6 +63,14 @@ def self.register_directory(label, path, test_directory: false)
...
+      self.pattern = /^(?!\.).*?\.(#{self.extensions.join("|")})$/
```
byroot: nitpick: It's better to use `Regexp.union`.

byroot (commented):
Seems fine to me overall, but a couple questions.

railties/lib/rails/code_statistics.rb
taketo1113: @byroot Thank you for the comments.

Is this the kind of usage you had in mind for `Regexp.union`?
```suggestion
      self.pattern = Regexp.union(self.extensions.map { |ext| /^(?!\.).*?\.(#{Regexp.escape(ext)})$/ })
```
(It also added `Regexp.escape`.)

byroot: More like: 

```suggestion
      self.pattern = /^(?!\.).*?\.(#{Regexp.union(extensions.map { |e| Regexp.escape(e) })})$/
```

NB: The `Regexp.escape` is a bit superfluous because I can hardly imagine an extension containing a special character, but might as well.

taketo1113: Thanks. I have updated it.

PR #56018, opened 2025-10-27
Railties: raise when both config.force_ssl and config.assume_ssl are enabled

### LINKED ISSUE #55695: Production SSL not enforced by default
### Steps to reproduce
SSL is not enforced by default as `config.assume_ssl` and `config.force_ssl` don't work when both set to `true`.

```sh
# Create a new rails app
$ rails new ssl-check
$ rails g controller Static index

# Check default options
$ cat config/environments/production.rb | grep ssl
config.assume_ssl = true
config.force_ssl = true

# Build and run container
$ docker build -t ssl-check .
$ docker run -d -p 80:80 -e RAILS_MASTER_KEY=<master-key> --name ssl-check ssl-check
Rails 8.0.2.1 application starting in production

# HTTP Request
$ curl -I http://localhost/static/index
HTTP/1.1 200 OK
```

### Expected behavior

- HTTP requests should force redirect to HTTPS unless: `config.force_ssl = false`

### Actual behavior

- HTTP requests are allowed when both `config.assume_ssl` and `config.force_ssl` are set to `true`
- HTTPS forced when `config.assume_ssl = false` and `config.force_ssl = true`

### System configuration

**Rails version**: 8.0.2.1

**Ruby version**:  3.4.5

daniel-niknam:
### Motivation / Background

Fixes #55695

`config.force_ssl` (adds `ActionDispatch::SSL`) and `config.assume_ssl` (adds `ActionDispatch::AssumeSSL`) are intended for different deployment setups and should not be enabled together. Enabling both can lead to confusion and makes the effective SSL policy ambiguous.

This change raises early if an application attempts to enable both settings at the same time, guiding the user to choose a single, correct configuration for their environment.

### Detail

**What changed**

* Convert `assume_ssl` and `force_ssl` from simple accessors to reader methods with custom writer methods that validate exclusivity.

* Add a private guard `check_for_only_assume_or_force_ssl` that raises:

  > Both config.force_ssl and config.assume_ssl cannot be enabled
  > simultaneously. The config.force_ssl and config.assume_ssl are
  > mutually exclusive, consider enabling only one of them.

* Add a test to assert a `RuntimeError` is raised when both options are set to `true`.

**Why**

* Prevents misconfiguration at boot time.
* Matches Rails’ pattern of validating incompatible configuration options early and providing a clear, actionable error message.

**Tests**

* `railties/test/application/configuration_test.rb`:

  * Adds `config.force_ssl = true and config.assume_ssl = true` test that asserts a `RuntimeError` during boot.

### Additional information

**Backwards compatibility**

* Applications that previously set both options will now fail fast with a clear error, prompting the operator to select exactly one of the two settings based on their deployment (reverse proxy/LB vs. app-level enforcement).

### Checklist

### DIFF
--- a/railties/test/application/configuration_test.rb
+++ b/railties/test/application/configuration_test.rb
@@ -1559,6 +1559,22 @@ module ApplicationTests
       assert_equal false, ActionView::Resolver.caching?
     end
 
+    test "config.force_ssl = true and config.assume_ssl = true" do
+      add_to_config <<~RUBY
+        config.assume_ssl = true
+        config.force_ssl = true
+      RUBY
+
+      e = assert_raise RuntimeError do
+        app "development"
+      end
+
+      assert_equal <<~MSG.squish, e.message
+        Both config.force_ssl and config.assume_ssl cannot be enabled simultaneously. The
+        config.force_ssl and config.assume_ssl are mutually exclusive, consider enabling only one of them.
+      MSG
+    end
+
     test "ActionController::Base::renderer uses Rails.application.default_url_options and config.force_ssl" do
       add_to_config <<~RUBY
         config.force_ssl = true
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Raise RuntimeError when both config.force_ssl and config.assume_ssl are enabled.
+
+    **This is a breaking change:** These settings target different deployment setups (`ActionDispatch::SSL` vs. `ActionDispatch::AssumeSSL`) and
+    are mutually exclusive. Attempting to enable both now raises early to prevent ambiguous SSL configuration.
+
+    *Daniel Niknam*
+
 *   Do not assume and force SSL in production by default when using Kamal, to allow for out of the box Kamal deployments.
 
     It is still recommended to assume and force SSL in production as soon as you can.
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -11,10 +11,10 @@ require "rails/source_annotation_extractor"
 module Rails
   class Application
     class Configuration < ::Rails::Engine::Configuration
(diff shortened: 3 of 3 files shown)


railties/test/application/configuration_test.rb
```diff
@@ -1559,6 +1559,17 @@ def index
...
     end
 
+    test "config.force_ssl = true and config.assume_ssl = true" do
+      add_to_config <<~RUBY
+        config.assume_ssl = true
+        config.force_ssl = true
+      RUBY
+
+      assert_raise RuntimeError do
+        app "development"
```
b1tfury, line 1569: good to have assertion around error message "Both config.force_ssl and config.assume_ssl cannot be enabled simultaneously. The
            config.force_ssl and config.assume_ssl are mutually exclusive, consider enabling only one of them."

daniel-niknam: @b1tfury I forgot to include that in the commit 🤦 Thanks for noticing. I'll update it soon.

maxfelsher:
I'm not sure it's true that `config.force_ssl` and `config.assume_ssl` are mutually exclusive. Setting `assume_ssl` will make it so that the HTTP->HTTPS redirect from `ActionDispatch::SSL` won't occur, but the functionality to set HSTS headers and secure cookies should still occur even when both `force_ssl` and `assume_ssl` are turned on. (And I don't think they'll happen if only `assume_ssl` is turned on.)

Up until https://github.com/rails/rails/commit/87feb28905bdb64eb6b2a0995782f1b105990d6c there was even a note in the default production environment file talking about how you can combine the two settings, and the commit message makes it sound like that note was only removed because it seemed redundant.

I think the real issue might be that some settings in Rails really rely on knowing what deployment setups they should be combined with, and that maybe should be explained better.

p8:
@maxfelsher Is correct and this shouldn’t raise.

daniel-niknam:
Thanks for the feedback. I’ll close this PR then.

PR #56019, opened 2025-10-27
Add structured event for open redirects

adrianna-chang-shopify:
### Motivation / Background

Follow up to: https://github.com/rails/rails/pull/55496

This Pull Request has been created so that applications with `config.action_controller.action_on_open_redirect = :notify` automatically get a structured log event, rather than having to configure their own subscriber to the Active Support Notification.

### Detail

Adds a method for `open_redirect` to `ActionController::StructuredEventSubscriber`. Uses the last line of the stacktrace for `stack_trace` to match `#rescue_from_callback`, but we could offer a bit more here if desired (return the last X lines, filter the trace, etc.)

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -117,7 +117,7 @@ module ActionController
     # The `action_on_open_redirect` configuration option controls the behavior when an unsafe
     # redirect is detected:
     # * `:log` - Logs a warning but allows the redirect
-    # * `:notify` - Sends an Active Support notification for monitoring
+    # * `:notify` - Sends an Active Support notification and structured event for monitoring
     # * `:raise` - Raises an UnsafeRedirectError
     #
     # To allow any external redirects pass `allow_other_host: true`, though using a
--- a/actionpack/lib/action_controller/structured_event_subscriber.rb
+++ b/actionpack/lib/action_controller/structured_event_subscriber.rb
@@ -69,6 +69,17 @@ module ActionController
       emit_event("action_controller.data_sent", filename: event.payload[:filename], duration_ms: event.duration.round(1))
     end
 
+    def open_redirect(event)
+      payload = event.payload
+
+      emit_event("action_controller.open_redirect",
+        location: payload[:location],
+        request_method: payload[:request]&.method,
+        request_path: payload[:request]&.path,
+        stacktrace: payload[:stack_trace],
+      )
+    end
+
     def unpermitted_parameters(event)
       unpermitted_keys = event.payload[:keys]
       context = event.payload[:context]
--- a/actionpack/test/controller/structured_event_subscriber_test.rb
+++ b/actionpack/test/controller/structured_event_subscriber_test.rb
@@ -4,6 +4,8 @@ require "abstract_unit"
 require "active_support/testing/event_reporter_assertions"
 require "action_controller/structured_event_subscriber"
 
+require "active_support/core_ext/object/with"
+
 module ActionController
   class StructuredEventSubscriberTest < ActionController::TestCase
     module Another
@@ -43,6 +45,10 @@ module ActionController
 
         def raise_error
           raise StandardError, "Something went wrong"
+        end
+
+        def open_redirector
+          redirect_to "example.com"
         end
       end
     end
@@ -177,6 +183,20 @@ module ActionController
     ensure
       ActionController::Base.enable_fragment_cache_logging = original_enable_fragment_cache_logging
       FileUtils.rm_rf(cache_path)
+    end
+
+    def test_open_redirect
+      ActionController::Base.with(action_on_open_redirect: :notify) do
+        event = assert_event_reported("action_controller.open_redirect", payload: {
+          location: "http://test.hostexample.com",
+          request_method: "GET",
+          request_path: "/action_controller/structured_event_subscriber_test/another/structured_event_subscribers/open_redirector"
+        }) do
+          get :open_redirector
+        end
+
+        assert(event[:payload][:stacktrace].find { |line| line.include?("structured_event_subscriber_test.rb:51") })
+      end
     end
   end
 end
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Emit a structured event when `action_on_open_redirect` is set to `:notify`
+    in addition to the existing Active Support Notification.
+
+    *Adrianna Chang*, *Hartley McGuire*
+
 *   Support `text/markdown` format in `DebugExceptions` middleware.
 
     When `text/markdown` is requested via the Accept header, error responses


actionpack/test/controller/structured_event_subscriber_test.rb
```diff
@@ -178,5 +182,26 @@ def test_fragment_cache
       ActionController::Base.enable_fragment_cache_logging = original_enable_fragment_cache_logging
       FileUtils.rm_rf(cache_path)
     end
+
+    def test_open_redirect
+      with_action_on_open_redirect(:notify) do
```
skipkayhil:
```suggestion
      ActionController::Base.with(action_on_open_redirect: :notify) do
```

I'd require the `with` extension instead of creating a method

adrianna-chang-shopify: Done!

eileencodes:
For some reason this test is failing only for Ruby 3.4, can you take a look? https://buildkite.com/rails/rails/builds/123220#019a262f-c5ec-408a-af81-f85550424863

actionpack/lib/action_controller/structured_event_subscriber.rb
```diff
@@ -65,6 +65,16 @@ def send_data(event)
       emit_event("action_controller.data_sent", filename: event.payload[:filename], duration_ms: event.duration.round(1))
     end
 
+    def open_redirect(event)
+      payload = event.payload
+      emit_event("action_controller.open_redirect_detected",
+        location: payload[:location],
+        request_method: payload[:request]&.method,
+        request_path: payload[:request]&.path,
+        stack_trace: payload[:stack_trace]&.first&.delete_prefix("#{Rails.root}/")
```
gmcgibbon:
```suggestion
        backtrace: payload[:stack_trace]&.first&.delete_prefix("#{Rails.root}/")
```
To match the naming of the current `exception_backtrace` naming.

gmcgibbon: This isn't an error though so maybe `stacktrace` is better? The name sounds a bit funny in contrast to the current standard.

adrianna-chang-shopify: Yeah, IMO `stacktrace` is more "correct" since it's not an exception. I can remove the `_` if you'd prefer -- and update the ASN to match?

adrianna-chang-shopify:
Still seeing flakiness with the new test on ~Ruby 3.4~ (apparently not just Ruby 3.4), trying to repro locally...

adrianna-chang-shopify:
Aha, okay figured it out -- tests in redirect_test.rb were calling `ActiveSupport::Notifications.unsubscribe("open_redirect.action_controller")`, which was unsubscribing the structured event subscriber and resulting in failures depending on which order the tests ran in.

I've fixed all of the tests in `actionpack/test/controller/redirect_test.rb` to unsubscribe from _only_ the subscriber added in the test, not all subscribers. Flakiness should be resolved now.

actionpack/lib/action_controller/structured_event_subscriber.rb
```diff
@@ -65,6 +65,16 @@ def send_data(event)
       emit_event("action_controller.data_sent", filename: event.payload[:filename], duration_ms: event.duration.round(1))
     end
 
+    def open_redirect(event)
+      payload = event.payload
+      emit_event("action_controller.open_redirect_detected",
+        location: payload[:location],
+        request_method: payload[:request]&.method,
+        request_path: payload[:request]&.path,
+        stacktrace: payload[:stacktrace]&.first&.delete_prefix("#{Rails.root}/")
```
Edouard-chin: We probably need a `defined?(Rails)`, otherwise this won't work in projects using actionpack as a standalone gem

adrianna-chang-shopify: Good point, addressed 👍 I fixed a couple of the other callsites in the subscribers as well.

actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -292,7 +292,7 @@ def _enforce_open_redirect_protection(location, allow_other_host:)
           ActiveSupport::Notifications.instrument("open_redirect.action_controller",
             location: location,
             request: request,
-            stack_trace: caller,
+            stacktrace: caller,
```
skipkayhil: This was released in 8.1, so I'm not sure its worth breaking backwards compatibility to remove the underscore

gmcgibbon: We can remove the underscore in the new structured event subscriber method, but not here. The notification must remain the same.

adrianna-chang-shopify: Oops right ofc, forgot we just released 🤦‍♀️ Fixed!

gmcgibbon:
We should also add a changelog since this adds a new event. I don't think we've documented events yet, so we can do that later.

adrianna-chang-shopify:
> We should also add a changelog since this adds a new event. I don't think we've documented events yet, so we can do that later.

Yeah, none of the structured events are documented yet right? I can open a PR to do that separately.

hmcguire-shopify:
~I pushed up new commits to split out the backportable changes from the new feature. (I also dropped the change in log_subscriber because https://github.com/rails/rails/pull/56059 makes it unnecessary)~ (Edit: merged the backportable commits separately)

However, now that I've spent more time with the code I don't think we should be filtering the `stacktrace` in the event payload (related: https://github.com/rails/rails/issues/56060). We are more frequently interested in frames further up the stack than just the first one, and we lose that information by filtering here.

Edouard-chin (approved):

adrianna-chang-shopify:
> However, now that I've spent more time with the code I don't think we should be filtering the stacktrace in the event payload (related: https://github.com/rails/rails/issues/56060). We are more frequently interested in frames further up the stack than just the first one, and we lose that information by filtering here.

Yeah, agree 👍

actionpack/lib/action_controller/structured_event_subscriber.rb
```diff
@@ -69,6 +69,18 @@ def send_data(event)
       emit_event("action_controller.data_sent", filename: event.payload[:filename], duration_ms: event.duration.round(1))
     end
 
+    def open_redirect(event)
+      payload = event.payload
+      stacktrace = payload[:stack_trace]&.first
+      stacktrace = stacktrace&.delete_prefix("#{Rails.root}/") if defined?(Rails.root) && Rails.root
+      emit_event("action_controller.open_redirect_detected",
```
gmcgibbon:
```suggestion
      emit_event("action_controller.open_redirect",
```
I'd rather the event keep the same name as the notification. We do this already for `unpermitted_parameters`. Adding "detected" makes it vague as to when it actually happened IMO.

gmcgibbon (approved):
LGTM, but I think we should still add a changelog for this. The structured event subscriber is released in 8.1.

PR #56025, opened 2025-10-28
Add delegated_type support to model generator

eelcoj:
### Motivation / Background

Add delegated_type support to model generator

Allows generating models with delegated_type associations using:
```bash
rails generate model Entry entryable:delegated_type

# => delegated_type :entryble, types: []
```

### Checklist


zzak:
I'm not sure this will be accepted, the core team is kind of picky about what goes into generators and this feels like an uncommon need. Still you probably want to add a changelog in case they are willing to support it.

PR #56026, opened 2025-10-28
Add ActiveRecord.parallel_test_table_reset_method

djmb:
### Motivation / Background

Truncating tables is very slow on MySQL compared to deleting from them, so allow the method used by parallel tests to reset tables to be configurable.

### Detail

When doing parallel testing with many databases, truncating all the tables is very slow, at least on MySQL. Deleting from the tables is much faster.

Add a configuration option `parallel_test_table_reset_method` that can be set to `:truncate`, `:delete`, or `:skip` to control how test database tables are emptied.

The `:skip` option makes the `SKIP_TEST_DATABASE_TRUNCATE` env variable obsolete, but that has been kept for backward compatibility.

There's no difference between the two methods for SQLite, as the adapter uses DELETE for truncation anyway.

On MySQL and PostgreSQL, TRUNCATE resets the auto-increment counters, while DELETE preserves them.
### Additional information

Running on a schema with about 200 tables and 32 parallel databases, I can delete from all tables in about 3 seconds, while truncating takes about 20 seconds.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/railtie.rb
+++ b/activerecord/lib/active_record/railtie.rb
@@ -30,6 +30,7 @@ module ActiveRecord
     config.active_record.use_schema_cache_dump = true
     config.active_record.check_schema_cache_dump_version = true
     config.active_record.maintain_test_schema = true
+    config.active_record.parallel_test_table_reset_method = :truncate
     config.active_record.has_many_inversing = false
     config.active_record.query_log_tags_enabled = false
     config.active_record.query_log_tags = [ :application ]
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,21 @@
+*   Add `ActiveRecord.parallel_test_table_reset_method` configuration to control
+    how test database tables are reset during parallel test database initialization.
+
+    Can be set to `:truncate` (default), `:delete`, or `:skip`.
+
+    When running parallel tests, each worker gets a separate database copy. If the
+    schema is up to date, by default the tables are truncated to reset them.
+
+    On MySQL deletion can be significantly faster, so this allow us to use that as
+    an alternative method to clear out the tables.
+
+    Example:
+
+        # config/environments/test.rb
+        config.active_record.parallel_test_table_reset_method = :delete
+
+    *Donal McBreen*
+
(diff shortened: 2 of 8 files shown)


activerecord/lib/active_record/railtie.rb
```diff
@@ -30,6 +30,7 @@ class Railtie < Rails::Railtie # :nodoc:
     config.active_record.use_schema_cache_dump = true
     config.active_record.check_schema_cache_dump_version = true
     config.active_record.maintain_test_schema = true
+    config.active_record.parallel_test_table_reset_method = :truncate
```
flavorjones, line 33: @djmb Can you please also add documentation for the new option to `guides/source/configuring.md` in the [Configuring Active Record](https://guides.rubyonrails.org/configuring.html#configuring-active-record) section?

djmb: Thanks @flavorjones - I've added that documentation now.

dhh:
I'd prefer that we just do the fast thing without any configuration. We should just lean on configuration here. If truncating is slow, but deleting gives us the same outcome, we should just be deleting. Don't even offer truncating as an option.

djmb:
Truncating resets the auto increment IDs so we should probably make this configurable to start with and then deprecate the current behaviour. Relying on the IDs seems like an antipattern anyway.

Delete is much faster on MySQL and on SQLite truncating does a DELETE under the hood so there's no difference. For Postgres I'll need to test it out - if truncating is significantly faster there then we can make the default behaviour adapter dependant, otherwise we can switch to deleting everywhere by default.

djmb:
Truncation and deletion are both very fast on Postgres, though deletion is faster

On a dummy app with 200 tables, running tests across 32 processes I get:

```
DB          Truncation          Deletion
----------------------------------------
MySQL              18s             290ms
Postgres         300ms              50ms
SQLite             N/A              20ms 
```

So I think the way forward is to switch to deletion, but provide a config option for the existing behaviour that we can remove later.

djmb:
I've created https://github.com/djmb/test_db_cleanup if anyone else wants to try to reproduce my timings.

djmb:
Closing in favour of https://github.com/rails/rails/pull/56097

PR #56030, opened 2025-10-29
[ci skip] fix(guides): Update AJ serializer example to have public klass

ixti:
### Motivation / Background

As of Rails 8.1 ActiveJob custom serializer example yields deprecation warnings:

```
DEPRECATION WARNING: IpAddrSerializer#klass method should be public. 
This will raise an error in Rails 8.2.
```

### Detail

This Pull Request changes ActiveJob custom serializer example in guides.

### Checklist

### DIFF
--- a/guides/source/active_job_basics.md
+++ b/guides/source/active_job_basics.md
@@ -988,11 +988,10 @@ class MoneySerializer < ActiveJob::Serializers::ObjectSerializer
     Money.new(hash["amount"], hash["currency"])
   end
 
-  private
-    # Checks if an argument should be serialized by this serializer.
-    def klass
-      Money
-    end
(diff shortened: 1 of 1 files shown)


ixti:
If IPAddr serializer can be potentially added to the core - I'll be happy to open PR with something like:

``` ruby
module ActiveJob
  module Serializers
    class IPAddrSerializer < ObjectSerializer
      def serialize(ip_addr)
        super("value" => ip_addr.to_s)
      end

      def deserialize(hash)
        IPAddr.new(hash["value"])
      end

      def klass
        IPAddr
      end
    end
  end
end
```

skipkayhil (approved):

PR #56036, opened 2025-10-29
Add ability to use a block when rendering a collection

genezys:
### Motivation / Background

When using a partial that yields, we can already use this partial with `render partial: ` or directly `render @model`.
But this was not working with `render collection` or `render @models`. 

The alternative was to use `@models.each { |model| render model }` but we lose some advantages of `render collection:` like collection caching.

This allows to use a code like this:

```
# _post.html.erb
<article class="post">
  <h1><%= post.title %></h1>
  <div class="post__body"><%= post.body %></div>

  <%# Render additional content %>
  <%= yield post if block_given? %>
</article>

# index.html.erb
<%= render @posts do |post| %>
  <%= link_to "Edit", [:edit, post] %>
<% end %>
```

### Detail

This Pull Request changes the `CollectionRenderer` to make use of the already existing `block` argument. The block is sent to every rendered element.

### Additional information

For my test, I was unable to use `render partial:`and had to use `render layout:`. Otherwise, I end up with a nil partial. I was unable to investigate this issue, but the code works fine in an application using `render partial:` or `render layout:`.

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Add ability to pass a block when rendering collection. The block will be executed for each rendered element in the collection.
+
+    *Vincent Robert*
+
+
 ## Rails 8.1.1 (October 28, 2025) ##
 
 *   Respect `remove_hidden_field_autocomplete` config in form builder `hidden_field`.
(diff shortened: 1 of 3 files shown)


genezys:
Closing to fix the tags. I opened this PR with the wrong source base branch.

PR #56044, opened 2025-10-30
Add ability to use a block when rendering a collection

genezys:
### Motivation / Background

When using a partial that yields, we can already use this partial with `render partial: ` or directly `render @model`.
But this was not working with `render collection` or `render @models`. 

The alternative was to use `@models.each { |model| render model }` but we lose some advantages of `render collection:` like collection caching.

This allows to use a code like this:

```
# _post.html.erb
<article class="post">
  <h1><%= post.title %></h1>
  <div class="post__body"><%= post.body %></div>

  <%# Render additional content %>
  <%= yield post if block_given? %>
</article>

# index.html.erb
<%= render @posts do |post| %>
  <%= link_to "Edit", [:edit, post] %>
<% end %>
```

### Detail

This Pull Request changes the `CollectionRenderer` to make use of the already existing `block` argument. The block is sent to every rendered element.

### Additional information

For my test, I was unable to use `render partial:`and had to use `render layout:`. Otherwise, I end up with a nil partial. I was unable to investigate this issue, but the code works fine in an application using `render partial:` or `render layout:`.

### Checklist

### DIFF
--- a/actionview/lib/action_view/renderer/collection_renderer.rb
+++ b/actionview/lib/action_view/renderer/collection_renderer.rb
@@ -167,10 +167,10 @@ module ActionView
 
           collection_body = if template
             cache_collection_render(payload, view, template, collection) do |filtered_collection|
-              collection_with_template(view, template, layout, filtered_collection)
+              collection_with_template(view, template, layout, filtered_collection, block)
             end
           else
-            collection_with_template(view, nil, layout, collection)
+            collection_with_template(view, nil, layout, collection, block)
           end
 
           return RenderedCollection.empty(@lookup_context.formats.first) if collection_body.empty?
@@ -179,7 +179,7 @@ module ActionView
         end
       end
 
-      def collection_with_template(view, template, layout, collection)
+      def collection_with_template(view, template, layout, collection, block)
         locals = @locals
         cache = {}
 
@@ -194,7 +194,7 @@ module ActionView
 
           _template = (cache[path] ||= (template || find_template(path, @locals.keys + [as, counter, iteration])))
 
-          content = _template.render(view, locals, implicit_locals: [counter, iteration])
+          content = _template.render(view, locals, implicit_locals: [counter, iteration], &block)
           content = layout.render(view, locals) { content } if layout
           partial_iteration.iterate!
           build_rendered_template(content, _template)
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add ability to pass a block when rendering collection. The block will be executed for each rendered element in the collection.
+
+    *Vincent Robert*
+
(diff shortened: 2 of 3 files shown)


MatheusRich:
I wonder if we should add a note that the block's content doesn't affect caching

genezys:
I was wondering about your comment and tested. This PR currently raises an error when rendering a cached collection. I will further investigate.

genezys:
My error was an early return in my ERB partial. It was entirely my fault. No problem with the code in this PR.

genezys:
@MatheusRich Yes the block's content does not affect caching, similar to how blocks work with other render methods.

This PR was just about aligning the behaviour of render collection with other renders, but I can also include some documentation if that seems appropriate. Should that be in the [render guide] or the [caching guide]?

[render guide]: https://github.com/rails/rails/blob/main/guides/source/layouts_and_rendering.md#using-partials-to-simplify-views
[caching guide]: https://github.com/rails/rails/blob/main/guides/source/caching_with_rails.md#russian-doll-caching

MatheusRich:
@genezys I think it would be helpful to have that in one of the guides. It could be a separate PR after this one is merged.

actionview/lib/action_view/renderer/collection_renderer.rb
```diff
@@ -179,7 +179,7 @@ def render_collection(collection, view, path, template, layout, block)
         end
       end
 
-      def collection_with_template(view, template, layout, collection)
+      def collection_with_template(view, template, layout, collection, block)
```
chaadow, line 182: hi @guilleiguaran @rafaelfranca 

This broke `jbuilder` here: https://github.com/rails/jbuilder/blob/38339adaa9d44ba89c0dde2a795338a886941e6f/lib/jbuilder/collection_renderer.rb#L47

and consequently my production application 😭

genezys: This should easily be fixed by adding the new argument in jbuilder. Meanwhile, you could always monkey-patch ScopedIterator in your application if you need a immediate fix.

chaadow: I'm used to run on edge branches of gems, so naturally I forked the jbuilder gem and applied the fix for myself. but since i'm feeling generous, i'm sharing with you to avoid breaking production apps in the future! :D

chaadow: ( I'm also used to open PRs to fix, but since I'm not sure about the best strategy here for foreward and backward compatibility, i'll let maintainers of the gem decide 👌🏼 )

zzak: Wouldn't `&block` work here?

chaadow: @zzak i thought of that, but didnt have the patience to test it out xD. just saw your related PR, glad it worked!

PR #56048, opened 2025-10-30
Add a summary report at the end of `ContinousIntegration.run`

flavorjones:
### Motivation / Background

In applications where there are multiple verbose CI steps, I frequently find myself scrolling backwards in my terminal to see what steps failed. A summary report at the end of a run helps me quickly understand what has failed, and whether there are multiple failures.

### Detail

The `@results` ivar is changed to hold the step title in addition to the success boolean, and any multi-step `run` or `step` block will print the failed steps.

The output looks like:

```
❌ Continuous Integration failed in 0.02s
   ↳ Tests: Rails failed
   ↳ Tests: Engine failed
```

In color:

image

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56049, opened 2025-10-30
Add a summary report at the end of `ContinousIntegration.run`

flavorjones:
### Motivation / Background

In applications where there are multiple verbose CI steps, I frequently find myself scrolling backwards in my terminal to see what steps failed. A summary report at the end of a run helps me quickly understand what has failed, and whether there are multiple failures.

### Detail

The `@results` ivar is changed to hold the step title in addition to the success boolean, and any multi-step `run` or `step` block will print the failed steps.

The output looks like:

```
❌ Continuous Integration failed in 0.02s
   ↳ Tests: Rails failed
   ↳ Tests: Engine failed
```

In color:

image

### Other details

See previous conversation at https://github.com/rails/rails/pull/56048

### Checklist

### DIFF
--- a/activesupport/lib/active_support/continuous_integration.rb
+++ b/activesupport/lib/active_support/continuous_integration.rb
@@ -74,12 +74,12 @@ module ActiveSupport
     #   step "Single test", "bin/rails", "test", "--name", "test_that_is_one"
     def step(title, *command)
       heading title, command.join(" "), type: :title
-      report(title) { results << system(*command) }
+      report(title) { results << [ system(*command), title ] }
     end
 
     # Returns true if all steps were successful.
     def success?
-      results.all?
+      results.map(&:first).all?
     end
 
     # Display an error heading with the title and optional subtitle to reflect that the run failed.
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/continuous_integration.rb
```diff
@@ -123,13 +123,25 @@ def report(title, &block)
...
       end
 
       results.concat ci.results
     ensure
       Signal.trap("INT", "-")
     end
 
+    # :nodoc:
+    def failure_summary(title)
+      return if success? || results.size < 2
```
dhh: We're already on the failure branch of the former conditional, so think we can drop `success?` and I also think we should probably extract `results.size < 2` as `multiple_steps?` and put that into the caller, so it's `ci.failure_summary(title) if multiple_steps?`, so it's clear where the presentation is being made whether it needs a summary.

activesupport/lib/active_support/continuous_integration.rb
```diff
@@ -123,13 +123,25 @@ def report(title, &block)
         echo "\n✅ #{title} passed in #{elapsed}", type: :success
       else
         echo "\n❌ #{title} failed in #{elapsed}", type: :error
+        ci.failure_summary(title)
```
dhh: Actually, I would inline the production of the output here like this:

```ruby
def report(title, &block)
  Signal.trap("INT") { abort colorize(:error, "\n❌ #{title} interrupted") }

  ci = self.class.new
  elapsed = timing { ci.instance_eval(&block) }

  if ci.success?
    echo "\n✅ #{title} passed in #{elapsed}", type: :success
  else
    echo "\n❌ #{title} failed in #{elapsed}", type: :error

    failures.each do |title|
      echo "   ↳ #{title} failed", type: :error
    end
  end

  results.concat ci.results
ensure
  Signal.trap("INT", "-")
end

def failures
  results.select(&:first)
end
```

activesupport/lib/active_support/continuous_integration.rb
flavorjones: Done.

activesupport/lib/active_support/continuous_integration.rb
flavorjones: Done. We need to still check that there are multiple steps/results because every step creates its own CI instance with a single result, and we don't want to print reports in that case.

PR #56054, opened 2025-10-31
Add code examples section to guides guidelines.

p8:
image

### Checklist

### DIFF
--- a/guides/source/ruby_on_rails_guides_guidelines.md
+++ b/guides/source/ruby_on_rails_guides_guidelines.md
@@ -51,6 +51,109 @@ Use the same inline formatting as regular text:
 ##### The `:content_type` Option
 ```
 
+Code Examples
+-------------
+
+Wrap code examples with code fence syntax using backticks:
+
+````markdown
+```ruby
+puts "Hello World!"
+```
+````
+
+Always specify the language used in the example. Supported languages labels
+include: `ruby`, `html`, `irb`, `html+erb`, `bash` and `js`.
+
+````markdown
+```js
+alert("Hello World!")
+```
+````
+
+### Filenames for Examples
+
+If an example references a specific file add the filename in the comments:
+
+````markdown
+```ruby
+# app/model/product.rb
+
+class Product < ApplicationRecord
+end
+```
+````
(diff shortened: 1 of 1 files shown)


guides/source/ruby_on_rails_guides_guidelines.md
```diff
@@ -51,6 +51,109 @@ Use the same inline formatting as regular text:
...
+```ruby
+puts "Hello World!"
+```
+````
+
+Always specify the language used in the example. Supported languages labels
+include: `ruby`, `html`, `irb`, `html+erb`, `bash` and `js`.
+
+````markdown
+```html+erb
```
harsh183: very very minor nit: should we also add `<% app/views/products/index.html.erb %>` in the file here.

Alternatively maybe this snippet can be something that doesn't look like it's referencing a specific file
````markdown
```html+erb
<p>Hello, <%= "World" %></p>
```
````

harsh183:
I really like this!

guides/source/ruby_on_rails_guides_guidelines.md
shivabhusal: > should we also add <% app/views/products/index.html.erb %> in the file here.

It looks like this is just an example to demonstrate syntax-highlighted code in markdown. Adding it wouldn’t cause any issue, and while it may not be strictly necessary, including it could help keep things consistent with the other examples.

p8: The other `html+erb` example includes the filename as a comment (in the section about filenames), but maybe we should always include filenames for `html+erb`.
I'll use a different example instead, so the filename comments won't distract.

p8: I've replaced the example with a JavaScript example (to also make it clear `js` is used for JavaScript).

shivabhusal:
Awesome work @p8 👍🏼 

**For Reviewers**: Here is the link to the live preview to this change : https://p8-guides-file-names-comment.rails-docs-preview.pages.dev/guides/ruby_on_rails_guides_guidelines#code-examples

PR #56056, opened 2025-10-31
Add block support to ActiveModel::Access#slice

arzezak:
Allow `slice` to accept a block that filters public instance methods dynamically. When a block is provided, it selects methods from the object's public instance methods based on the block's criteria.

Examples:

```
person.slice { |method| [:id, :name].include?(method) }
person.slice { |method| method.start_with?("display_") }
```

Also adds documentation for the `Access` module and its methods, including usage examples for both `slice` and `values_at`.

### DIFF
--- a/activemodel/lib/active_model/access.rb
+++ b/activemodel/lib/active_model/access.rb
@@ -4,11 +4,43 @@ require "active_support/core_ext/enumerable"
 require "active_support/core_ext/hash/indifferent_access"
 
 module ActiveModel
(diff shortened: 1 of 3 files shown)


byroot:
This method is mirrored from `Hash#slice`, as such it should keep a similar signature and behavior.

The behavior you want is more similar to `Hash#select`: https://docs.ruby-lang.org/en/master/Hash.html#method-i-select

So if such feature had to be accepted, it would be through a `select` method.

PR #56059, opened 2025-11-01
Fix double filtering rescue_from_handled backtrace

hmcguire-shopify:
When ActionController::StructuredEventSubscriber was created to consume events instead of notifications, it kept its previous logic to take the first frame of the backtrace and remove the Rails root. However, the event being consumed _already did this_, so the backtrace output was a single character.

This commit fixes the issue by removing the duplicate backtrace filtering. Now its only filtered once in the event and the log message properly displayed a whole path.


skipkayhil:
Backported the test to `8-1-stable` in 27e709a

PR #56062, opened 2025-11-01
Use -infinity for lower value of unbounded PG time ranges

Martin-Alexander:
Fixes: #56061

Use `-infinity` rather than `NULL` for the lower value of PostgreSQL time ranges when saving records with a Ruby range that begins with `nil`.

### Motivation / Background

In PostgreSQL if you have a time range column (`daterange`, `tstzrange`, etc) and save two records like so...

```ruby
  create_table :products do |t|
    t.tsrange :period
  end
end

class Product < ActiveRecord::Base; end

t = Time.utc(2000)

Product.create(period: t...nil)
Product.create(period: nil...t)
```

The resultant rows will have these respective values in the `period` column

```
["2000-01-01 00:00:00",infinity)
(NULL,"2000-01-01 00:00:00")
```

This can results in unexpected behaviour when querying with PostgreSQL's range operators. For instance, `lower_inf` treats `NULL` and `-infinity` differently.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
@@ -205,14 +205,13 @@ module ActiveRecord
           end
 
           def encode_range(range)
-            lower_bound = type_cast_range_value(range.begin)
-            upper_bound = if date_or_time_range?(range)
-              # Postgres will convert `[today,]` to `[today,)`, making it exclusive.
-              # We can use the special timestamp value `infinity` to force inclusion.
-              # https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-INFINITE
-              range.end.nil? ? "infinity" : type_cast(range.end)
+            lower_bound, upper_bound = if date_or_time_range?(range)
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
```diff
@@ -205,14 +205,13 @@ def encode_array(array_data)
           end
 
           def encode_range(range)
-            lower_bound = type_cast_range_value(range.begin)
-            upper_bound = if date_or_time_range?(range)
-              # Postgres will convert `[today,]` to `[today,)`, making it exclusive.
-              # We can use the special timestamp value `infinity` to force inclusion.
-              # https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-INFINITE
```
Martin-Alexander, line 212: Unsure if this comment is still necessary since there's less of a "special case" thing going on -- we're treating both ends of the range the same now.

guilleiguaran:
We need a CHANGELOG entry for this.

guilleiguaran (approved):

PR #56063, opened 2025-11-01
[ActiveSupport ContinuousIntegration] Add attempts option to step to allow CI steps to retry

harsh183:
### Motivation / Background

Some CI steps might flake and currently that fails the whole local CI run. I'd like the option to retry certain steps.

### Detail

To enable that, I'm introducing a new `attempts: 1` optional argument that can be set on a `step` to retry until success. 

By default it's 1, meaning no retries and existing step calls should work exactly the same. If set to a higher number, the step will be retried up to that many times before failing. If it's not positive, there will be an ArgumentError. New test cases have been added for this feature.

### Additional information

While I've made many contributions to the guides before, this is my first attempt at trying something with the codebase so let me know if I should change anything to match conventions. 

I also could not figure out the best way to reliably test a flaky scenario without modifying external state, I'm open to ideas on that.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -16,5 +16,14 @@
 
     *Gannon McGibbon*
 
+*   Add `attempts` parameter to `ActiveSupport::ContinuousIntegration#step` to
+    allow retrying flaky CI steps a specified number of times before marking
+    them as failed.
+    ```ruby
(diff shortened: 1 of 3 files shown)


harsh183:
bumping this

harsh183:
https://github.com/rails/rails/pull/56774 - with parallelism now in play I might want to rethink retries, I'll think of another approach

PR #56065, opened 2025-11-02
Add zeitwerk:check to default CI workflow and bin/ci

trevorturk:
## Motivation

The `zeitwerk:check` rake task has existed since Rails 6 (April 2019) but is not included in the default CI configuration added in Rails 7.2.

While `config.eager_load = ENV["CI"].present?` causes tests to eager load in CI, a dedicated `zeitwerk:check` provides:

- **Fail-fast feedback**: ~2s check before running the full test suite
- **Explicit validation**: Clear CI step showing autoloading is verified
- **Exhaustive checking**: Tests might not exercise all autoload paths

## Real-World Impact

This check would have caught production crashes we experienced with:
- Missing Zeitwerk inflector rules for acronyms (MCP → Mcp)
- Files not defining expected constants

See upstream PRs for examples:
- https://github.com/parruda/swarm/pull/163 (MCP inflector fix)
- https://github.com/crmne/ruby_llm/pull/486 (Railtie eager loading fix)

## Implementation

Adds `zeitwerk:check` to:
1. **GitHub Actions workflow** (`lint` job, after rubocop)
2. **bin/ci script** (after Style: Ruby step)

Both are structural validation checks that fail fast before running expensive tests.

## Testing

- Added test assertion in `bin_ci_test.rb`
- Updated `CHANGELOG.md`

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Add `zeitwerk:check` to default CI workflow and `bin/ci`.
+
+    The `zeitwerk:check` rake task verifies that the application can eager load
+    all constants without errors. This check is now included in both the GitHub
+    Actions workflow template and the `bin/ci` script to catch autoloading issues
+    before deployment.
+
+    *Trevor Turk*
+
 *   Do not assume and force SSL in production by default when using Kamal, to allow for out of the box Kamal deployments.
 
     It is still recommended to assume and force SSL in production as soon as you can.
--- a/railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
@@ -5,6 +5,7 @@ CI.run do
 <% unless options.skip_rubocop? %>
   step "Style: Ruby", "bin/rubocop"
 <% end -%>
+  step "Zeitwerk: Eager loading check", "bin/rails zeitwerk:check"
 
 <% unless options.skip_bundler_audit? -%>
   step "Security: Gem audit", "bin/bundler-audit"
--- a/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
@@ -74,6 +74,9 @@ jobs:
       - name: Lint code for consistent style
         run: bin/rubocop -f github
 
+      - name: Check Zeitwerk eager loading
+        run: bin/rails zeitwerk:check
+
 <% end -%>
 <% unless options[:skip_test] -%>
   test:
--- a/railties/test/application/bin_ci_test.rb
+++ b/railties/test/application/bin_ci_test.rb
@@ -18,6 +18,7 @@ module ApplicationTests
 
         # Default steps
         assert_match(/bin\/rubocop/, content)
+        assert_match(/bin\/rails zeitwerk:check/, content)
         assert_match(/bin\/brakeman/, content)
         assert_match(/bin\/bundler-audit/, content)
         assert_match(/"bin\/rails test"$/, content)


fxn:
@trevorturk Let me think about it.

The task is not there because all `zeitwerk:check` does is eager load the application. As a side-effect, when files are loaded the error checks of Zeitwerk run, as they do in lazy mode too.

To do this, `zeitwerk:check` calls Rails API that eager loads the application.

But `config.eager_load = true` is more than that, because Rails is the one eager loading and **with the flag set to true**. Which is closer to production. And that covers what `zeitwerk:check` does and more. That is closer to production, because everything else affected by `config.eager_load = true` runs with the flag set.

So, was CI in that project eager loading?

trevorturk:
Ok! Interesting to know that `config.eager_load = true` does even more here...

I thought about it and created an alternative that might be more appealing here: https://github.com/rails/rails/pull/56071

 We could consider one or both of these ideas, and I'm happy to be convinced it's not necessary as well. The default `config/environments/test.rb` is pretty reasonable with the comment I see now:

```ruby
  # Eager loading loads your entire application. When running a single test locally,
  # this is usually not necessary, and can slow down your test suite. However, it's
  # recommended that you enable it in continuous integration systems to ensure eager
  # loading is working properly before deploying your code.
  config.eager_load = ENV["CI"].present?
```

...but perhaps even better for the default `rails new` to have `ENV["CI"]` set already, so you don't need to read and understand the comment.

fxn:
But, let me understand.

Your CI was not eager loading? The point of the env variable is that it is set by CIs, you do not need to do that. If there was an exception, it would be helpful to know..

trevorturk:
Yes, I'm confused now. Perhaps I did something wrong here, or it's a different bug / misunderstanding. Let me try to make a simple reproduction and I'll report back!

trevorturk:
Ok, I think I have a pretty straightforward reproduction in a sample Rails app. You can browse the commits and GitHub CI workflows that have run to see what I mean, or view this issue which links to a couple of Gists which demonstrate the issue: https://github.com/trevorturk/eager-loading-example/issues/3

The important part here is that in the "before" case, I have `rails new` app which adds a single gem (within known loading issues that have since been fixed, but we're using the known-bad version) and both local and GitHub CI pass, when deploying to Heroku crashes. Then, in the "after" case, I'm not fixing the underlying issue, but just demonstrating that `zeitwerk:check` would have flagged the issue in local and GitHub CI (which is what my PRs are attempting to solve). 

I may be misunderstanding a bunch of things here (for example, I didn't know that GitHub set `CI=true`) and there may be Zeitwerk bugs, Rails bugs, etc that would be the "real" fix, I'm really not sure yet. But I do want to emphasize the issue I'm trying to identify is that I have an brand new Rails app that passes local and GitHub CI, but crashes in production, and I'd like to get to the bottom of it!

Thanks for having a look, @fxn and @skipkayhil, I appreciate the 👀s!

fxn:
@trevorturk awesome, thanks very much.

> But I do want to emphasize the issue I'm trying to identify is that I have an brand new Rails app that passes local and GitHub CI, but crashes in production, and I'd like to get to the bottom of it!

Absolutely!

In my machine, the test suite fails.

Reproduction steps:

```
rails new issue_56065 --minimal
cd issue_56065
bundle add swarm_sdk --version=2.1.1
bin/rails g model User
bin/rails db:migrate
CI=1 bin/rails test
```

That raises

```
undefined method 'clean_parameters' for module 'RubyLLM::Providers::Anthropic::Tools' (NameError)
```

Which CI service are you using? Can you verify without doubt that it is not setting the `CI` environment variable?

trevorturk:
Interesting, it seems the addition of the User model as you show does trigger the crash. Perhaps my repro/demo app (https://github.com/trevorturk/eager-loading-example) didn't have the issue because I had no model tests? 

Here I'm moving the `zeitwerk:check` so it happens after `rails test` in a GitHub CI workflow, which is documented to have the env var set: https://github.com/trevorturk/eager-loading-example/actions/runs/19110486642/job/54606013435 

You can see `rails test` doesn't fail, but perhaps because there are no tests to run and therefore it exits early? 

I also did some debugging and it appears the env var is correctly set for `bin/ci` as mentioned in #56071. 

I'm a bit confused (and might have a hard time attempting to reproduce) because my private repo definitely had tests for `rails test` to run, but they didn't trigger the issue, only `zeitwerk:check` did. There's one thing I could check where I had `require: false` in the `Gemfile` for `swarm_sdk` which could be playing a role here.

I'll attempt to reproduce with that in mind and report back!

fxn:
@trevorturk yeah, when we do `bin/rails test`, minitest autorun is invoked. If there are no tests, nothing runs. I created a model test because it was an easy automation, but any test will do.

The file that boots the app is `test/test_helper.rb`:

```ruby
ENV["RAILS_ENV"] ||= "test"
require_relative "../config/environment" # <-- this is what boots
...
```

so if no tests are loaded, that file is not loaded, and the app does not even boot. That is the trick.

We are hesitant to add an explicit `zeitwerk:check` step for CI because:

1. `config.eager_load = true` is what you have in production normally, that is what you want CI to do.
2. (1) eager loads, and anything else that depends conditionally on that flag runs too.
3. As a side-effect of eager loading, Zeitwerk checks its error conditions.
4. `zeitwerk:check` is not synthetic, all it does is eager load (to trigger (3)).
5. If we add `zeitwerk:check` to the generated steps, not only it is redundant, it will make CI eager load the application twice, which may add unnecessary build time.

So, I suspect some lesson is going come out of this ticket. Even if it is documentation or a better code comment, but that is the context :).

trevorturk:
Thanks for the thoughtful comment. It's on my list to try to get to the bottom of what exactly went wrong in my app, as I suspect it may be a Zeitwerk usage issue with the gem (I've been trying to fix them as I encounter them) or maybe the way I was including but not requiring the gem in my Gemfile. I'll report back ASAP, but I'm behind on my client work at the moment, so it may be next week. 

I totally understand we may not want to add this additional check, which would result in many, many hours of additional server usage across all Rails apps on GitHub CI etc etc, so I don't take it lightly! I do want to try to get to the bottom of the issue I had, though, because it's no fun to have a green CI but then the app crash when you deploy, and I'd like to make sure I understand why!

So, tl;dr, I totally understand and we can close this PR, or leave it open for now until I report back with additional findings. Thanks again!

trevorturk:
Ok! I'm sorry for the delay! I think I might have found a way to reproduce. 

I worked with Claude Code to do a kind of binary search strategy which eventually found that adding the `tailwindcss-rails` gem seems to trigger the issue. 

To reproduce, you should be able to run:

```
rails new issue_56065 --css tailwind
cd issue_56065
bundle add swarm_sdk --version=2.1.3
bundle install
bin/rails g model User
bin/rails db:migrate
# uncomment the generated test in: test/models/user_test.rb
CI=1 bin/rails test # passes
CI=1 bin/rails zeitwerk:check # fails
```

I also pushed the repro app to GitHub. You can see the commit adding the `zeitwerk:check` here, and see that `rails test` passes while `zeitwerk:check` fails: https://github.com/trevorturk/rails_issue_56065/commit/205a209c -- click the small red X to see the CI results.

I totally understand if this doesn't change the calculation in terms of adding `zeitwerk:check` to the generated CI template etc, so don't worry about that aspect of things. Perhaps we'll end up wanting a fix in `tailwindcss-rails` or something, I'm not sure yet. I'm just happy to see that we might be onto something here!

trevorturk:
Ok! I worked with Claude on this a bit more (please forgive me) and I think we have another potential option worth investigating a bit more... and then possibly considering for inclusion in the default generators. 

I worked up a new PR here: https://github.com/rails/rails/pull/56212

tl;dr it's suggesting that we add `config.rake_eager_load = ENV["CI"].present?` to `config/environments/test.rb`. 

I'm a bit unclear on the finer points here w/r/t `eager_load` and `rake_eager_load` (I haven't encountered `rake_eager_load ` directly before...) but I think this gives us something to discuss further, in any case. 

I tested this fix with my repro app here: https://github.com/trevorturk/rails_issue_56065/commit/2e76316a -- click the small red x to see the CI results. In this case, we have the desired outcome, which is that `CI=1 bin/rails test` correctly fails with a `zeitwerk` error. So, we presumably would have caught this issue in CI without adding the additional standalone `zeitwerk:check` bit. 

Apologies again for the delay in getting back to this, and for the long-winded and somewhat-incomplete investigation, but I hope this is helpful and/or interesting anyway. Thank you!

fxn:
Hi, no need to apologyze! :)

Thanks, I could reproduce, I'll investigate the root cause of this.

I'd like to underline one thing: The problem in this ticket is that the application **is not being eager loaded**. This is conceptually orthogonal to `zeitwerk:check`. If the checker was implemented as a parser, the application would not be even loaded and the error would have not surfaced.

That gem in that version cannot be eager loaded, it is doing an invalid method call. `zeitwerk:check` would have shown that, but that is a red herring, note that no Zeitwerk-violation is being found.

What we want to ensure is that the application is eager loaded when the test suite runs in CI. That's the key point in this ticket.

I'll write back!

trevorturk:
Thanks @fxn! Agreed with your underscore and key point above. I think the way I've conceptualized this bug hunt is that my (relatively new and vanilla Rails app) passed CI but crashed in production and I was like... what the heck!? 🤣

Initially I thought we weren't even setting `CI=true` and thus opened the misguided #56071. Then, for this PR, I was thinking we might consider leveraging the `zeitwerk:check` to "fail fast" with a simple check with clear intentions, but I agree this has other concerns such as wasted CI time, and in fact the core issue is that we want, as you said, to make sure we're eager loading the application when running in CI. 

If you didn't see it already, check out the quick PR I put together in #56212, which @guilleiguaran kindly merged. I think this does show and explain the root cause, and there's a +1 from @chaadow who encountered this before. However I'm still a bit new to this area and of course would love to have your eyes on it, since it's eager_loading adjacent in any case. I want to make sure we're not unintentionally causing some other knock-on issue in the process, if that makes sense.

fxn:
I guessed something like that was the explanation.

But I am not sure about that patch. We do not want ALL Rake task in your CI to eager load the app. Is there a better solution?

trevorturk:
Yes, I agree, it seems sensible at first, but perhaps it's a bit "too big" in that it seems like a small fix but actually changes quite a lot. 

I'm not sure what's better, though. Perhaps we could be specific about only changing the `bin/rails test` behavior somehow? I could imagine having a check to assert that eager loading is enabled or something? In some way we'd want acknowledge that running the test suite is a special sort of rake task, if you see what I mean. 

I'm just brainstorming, and this would be a weird, new sort of idea, but I could imagine even having a generated unit test that would fail if `CI=true` but eager_load is not enabled?

Anyway, no worries if we want to revert that earlier merge for now. I should have made it a draft PR since this is still under consideration and a bit tricker than it might initially appear!

fxn:
OK.

Since what we learnt is that eager loading was the thing missing in your CI and it is conceptually orthogonal to `zeitwerk:check`, I think we can close this one.

We may still think if the solution in https://github.com/rails/rails/pull/56212 can be improved.

trevorturk:
Sounds good. Thank you!

PR #56071, opened 2025-11-02
Set ENV["CI"] in default CI configuration

trevorturk:
## Summary

This PR activates eager loading in CI by setting `ENV["CI"]` in the default CI configuration templates (bin/ci and GitHub Actions workflow).

Since Rails 6.1 ([#43508](https://github.com/rails/rails/pull/43508)), `config.eager_load = ENV["CI"].present?` has been the default in `config/environments/test.rb`, but the `CI` environment variable was not set in the CI configuration templates, so eager loading was not enabled by default.

## Changes

- Sets `CI=true` in `bin/ci` test commands (using `env CI=true` prefix)
- Sets `CI: true` in GitHub Actions workflow test job env blocks

## Alternative Approach

See [#56065](https://github.com/rails/rails/pull/56065) for an alternative approach that adds an explicit `zeitwerk:check` step to CI.

### ENV["CI"] approach (this PR)
**Pros:**
- Leverages existing Rails configuration pattern from #43508
- Tests eager loading in actual test context (more realistic)
- Catches runtime issues that may only appear during actual usage
- No additional CI step needed

**Cons:**
- Less explicit about what's being validated
- Slower (eager loading overhead on all test runs)
- Error messages can be buried in test output
- May not exercise all autoload paths if tests fail early

### Explicit zeitwerk:check approach (#56065)
**Pros:**
- Clear, visible validation step in CI logs
- Fails fast (~2s check before running full test suite)
- Clean, focused error messages
- Properly placed in lint job alongside other static checks
- Exhaustive checking of all autoload paths

**Cons:**
- Adds another CI step
- Doesn't test eager loading in actual runtime context
- Slight duplication with the eager_load configuration

Both approaches are valid. This PR is simpler and uses existing Rails patterns, while the explicit approach provides faster feedback and clearer validation.

Related: #43508

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
@@ -20,10 +20,10 @@ CI.run do
 <% end -%>
 <% unless options[:skip_test] -%>
 <% if options[:api] || options[:skip_system_test] -%>
-  step "Tests: Rails", "bin/rails test"
+  step "Tests: Rails", "env CI=true bin/rails test"
 <% else %>
-  step "Tests: Rails", "bin/rails test"
-  step "Tests: System", "bin/rails test:system"
+  step "Tests: Rails", "env CI=true bin/rails test"
+  step "Tests: System", "env CI=true bin/rails test:system"
 <% end -%>
 <% unless options.skip_active_record? -%>
   step "Tests: Seeds", "env RAILS_ENV=test bin/rails db:seed:replant"
--- a/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
+++ b/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
@@ -136,6 +136,7 @@ jobs:
 
       - name: Run tests
         env:
+          CI: true
           RAILS_ENV: test
           <%- if options[:database] == "mysql" -%>
           DATABASE_URL: mysql2://127.0.0.1:3306
@@ -209,6 +210,7 @@ jobs:
 
       - name: Run System Tests
         env:
+          CI: true
           RAILS_ENV: test
           <%- if options[:database] == "mysql" -%>
           DATABASE_URL: mysql2://127.0.0.1:3306
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Enable eager loading in CI by setting `ENV["CI"]` in default CI configuration.
+
+    The `CI` environment variable is now set to `"true"` when running tests via `bin/ci`
+    and GitHub Actions, which activates `config.eager_load = ENV["CI"].present?` in
+    `config/environments/test.rb`. This ensures autoloading issues are caught during
+    test runs in continuous integration environments.
+
+    *Trevor Turk*
+
 *   Do not assume and force SSL in production by default when using Kamal, to allow for out of the box Kamal deployments.
 
     It is still recommended to assume and force SSL in production as soon as you can.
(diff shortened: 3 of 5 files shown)


railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
```diff
@@ -136,6 +136,7 @@ jobs:
 
       - name: Run tests
         env:
+          CI: true
```
skipkayhil, line 139: https://docs.github.com/en/actions/reference/workflows-and-actions/variables
I believe GitHub Actions does this by default

railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
```diff
@@ -20,10 +20,10 @@ CI.run do
 <% end -%>
 <% unless options[:skip_test] -%>
 <% if options[:api] || options[:skip_system_test] -%>
-  step "Tests: Rails", "bin/rails test"
+  step "Tests: Rails", "env CI=true bin/rails test"
```
skipkayhil, line 23: https://github.com/rails/rails/blob/d3d17c297a70d3e255a34693a1fb0074712e9930/activesupport/lib/active_support/continuous_integration.rb#L57
Is this not adding `CI=true` already?

skipkayhil (commented):
Thank you for the pull request, but I'm not sure any of this is needed

trevorturk:
@skipkayhil thanks for the review! See also #56065 for a bit more detail, because I'm confused now...

I see that https://docs.github.com/en/actions/reference/workflows-and-actions/variables appears to set `CI=true` automatically, so I would expect `config.eager_load = ENV["CI"].present?` to have trigger eager loading issues I saw in production, but my GitHub and `bin/ci` worked fine, just that the specific `zeitwerk:check` did fail. Otherwise, it was only when running in production that I noticed anything wrong. 

Perhaps it's worth me creating a `rails new` app and documenting the steps I took to make sure I can reproduce? Otherwise I think I have to assume that I did something wrong here. I can also try to reproduce again in my current app where I noticed the issues. 

In any case, I'd be curious if you have thoughts on #56065, or if it's best for me to try to reproduce carefully and report back?

I appreciate the review and advice! Thanks!

skipkayhil:
> I can also try to reproduce again in my current app where I noticed the issues.

Yeah I think this is the best path forward.

trevorturk:
Ok I will do and report back, should be in the next few days. Thanks again!

trevorturk:
@skipkayhil I think I have a pretty straightforward reproduction, but I'm not sure what the best course of action is yet. Would you mind taking the conversation here, please? https://github.com/rails/rails/pull/56065 ...otherwise I'll just be cross-posting to both. Thanks again!

railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt
byroot: Yes. This isn't needed.

railties/lib/rails/generators/rails/app/templates/config/ci.rb.tt
```diff
@@ -20,10 +20,10 @@ CI.run do
...
 <% else %>
-  step "Tests: Rails", "bin/rails test"
-  step "Tests: System", "bin/rails test:system"
+  step "Tests: Rails", "env CI=true bin/rails test"
+  step "Tests: System", "env CI=true bin/rails test:system"
```
byroot, line 26: This is already set by the helper https://github.com/rails/rails/blob/65dc2163e314faa42d0e8393929ba2699e35090f/activesupport/lib/active_support/continuous_integration.rb#L57

byroot:
Thank you for the PR, but unless I'm mistaken that env var is already set globally.

PR #56072, opened 2025-11-02
Fix foreign key order and table name prefixing when dumping multiple schemas in postgres

robinator:
Closes #56070.

### Motivation / Background

We have a postgres database with multiple schemas and foreign keys that reference tables across the schemas.  I noticed some issues when running `rails db:schema:dump` on this database:

1. The dumper does not handle table names properly if they're already in a schema, e.g it will dump the table reference `vault.credit_cards` as `public.vault.credit_cards`.

2. The dumper appends foreign keys after all tables have been dumped but within each schema.  This causes load order issues when keys in one schema reference tables in another which might not have been created yet.

### Detail

I had to do a larger refactor of the foreign key dumping strategy to fix the ordering but I think the result is clear.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
@@ -42,7 +42,6 @@ module ActiveRecord
                 types.sort.each do |name, values|
                   stream.puts "  create_enum #{relation_name(name).inspect}, #{values.inspect}"
                 end
-                stream.puts
               end
             end
           end
@@ -64,6 +63,13 @@ module ActiveRecord
               stream.puts if previous_schema_had_tables
               super
               previous_schema_had_tables = @connection.tables.any?
+            end
+          end
+
+          def foreign_keys(stream)
+            previous_schema_had_tables = false
+            within_each_schema do
+              super
             end
           end
 
@@ -152,7 +158,7 @@ module ActiveRecord
           end
 
           def relation_name(name)
-            if @dump_schemas.size == 1
+            if @dump_schemas.size == 1 || name.include?(".")
               name
             elsif name.include?(".")
               name  # Already schema-qualified, don't add another prefix
--- a/activerecord/lib/active_record/schema_dumper.rb
+++ b/activerecord/lib/active_record/schema_dumper.rb
@@ -63,6 +63,7 @@ module ActiveRecord
       extensions(stream)
       types(stream)
       tables(stream)
+      foreign_keys(stream)
       virtual_tables(stream)
       trailer(stream)
       stream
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
```diff
@@ -67,6 +66,15 @@ def tables(stream)
...
           end
 
+          def foreign_keys(stream)
+            previous_schema_had_tables = false
+            within_each_schema do
+              stream.puts if previous_schema_had_tables
+              super
+              previous_schema_had_tables = @connection.tables.any?
+            end
+          end
```
agrobbin, line 74: This ended up causing a handful of extra newlines around the foreign key declarations, I think due to some of our schemas not including foreign keys. To temporarily apply this patch in our app, I ended up adjusting this method definition to:

```ruby
def foreign_keys(stream)
  within_each_schema do
    super
  end
end
```

Hopefully that's helpful!

chiperific:
I fixed the `"public.vault.credit_cards"` issue here: https://github.com/rails/rails/pull/56556

And I have a PR up to fix the order of emitted tables specifically for Postgres here: https://github.com/rails/rails/pull/56560

Maybe we need to combine our work?

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -63,6 +63,7 @@ def dump(stream)
       extensions(stream)
       types(stream)
       tables(stream)
+      foreign_keys(stream)
```
chiperific, line 66: @rafaelfranca [didn't like this approach on my PR](https://github.com/rails/rails/pull/56560#issuecomment-3725817292), he pointed out that not all adapters care about referential integrity in the schema.rb file.

activerecord/test/cases/adapters/postgresql/schema_test.rb
```diff
@@ -1062,8 +1066,18 @@ def test_schema_dump_with_dump_schemas_schema_search_path
...
       end
     end
   end
+
+  def test_foreign_keys_for_all_schemas_are_dumped_at_the_bottom_to_circumvent_dependency_issues
+    with_dump_schemas(:all) do
+      output = dump_all_table_schema
+
+      # a create_table should never occur after an add_foreign_key
+      assert_no_match(/add_foreign_key[\s\S]*create_table/m, output)
```
chiperific, line 1111: Oh! This is a slick test. Nice!

activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
robinator: Good catch!  I updated the method based on your feedback.  thanks!

robinator:
@chiperific It seems like your fixes have a lot more momentum than mine so feel free to take whatever you think is useful from it and I can close this PR and issue in favor of yours.

activerecord/lib/active_record/schema_dumper.rb
robinator: I think it makes more sense where I had put it because that's where the dumper is doing all the pooping of the main objects.  Leaving it only in postgres would make _that_ dumper do extra backflips to get it to work but I just want it to work at the end of the day 😇

chiperific:
> @chiperific It seems like your fixes have a lot more momentum than mine so feel free to take whatever you think is useful from it and I can close this PR and issue in favor of yours.

Ha! I wish this was the case! Mine's stagnating too.

PR #56075, opened 2025-11-02
Various event related fixes

skipkayhil:
(extracted from #56019)

---

[Fix ActiveSupport -> Active Support](https://github.com/rails/rails/commit/b10a43e67dbb1df82c1c4fd5b819718742bb564b)

We're describing the framework, not the module.

---

[Fix dependency on Rails constant](https://github.com/rails/rails/commit/3d4e3bd35f87ffec543034c4e987e51dfbc3f664)

Action Pack can be used without Rails

---

[Fix redirect_test leaking subscription state](https://github.com/rails/rails/commit/fd27a9c5c0dd58a70c72f3d093ba32520c04b239)

Previously these tests would unsubscribe _all_ subscribers for
{unsafe,open}_redirect notifications. This becomes a problem when Action
Pack provides a default subscription for these events because those
subscribers will get unsubscribed as well.

This commit fixes the issue by only unsubscribing the subscriber used
in the tests.

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -117,7 +117,7 @@ module ActionController
     # The `action_on_open_redirect` configuration option controls the behavior when an unsafe
     # redirect is detected:
     # * `:log` - Logs a warning but allows the redirect
-    # * `:notify` - Sends an ActiveSupport notification for monitoring
+    # * `:notify` - Sends an Active Support notification for monitoring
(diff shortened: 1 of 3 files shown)


skipkayhil:
Backported to `8-1-stable` in:

873cda559de396b9b9e63b9b2c62b96a1267bf15
13589db95b8380411baeea21dfdd2cc8e7843231
4388688077b6406e9d71fd622f98677a83a2c2d0

zzak:
I wonder if we could detect when subscriptions are leaked between tests

PR #56080, opened 2025-11-03
Add Rails.app.revision for deployment tracking

seuros:
## Summary

Add `Rails.app.revision` for deployment tracking and error reporting.

## Motivation

Common pattern seen everywhere:
```ruby
# Reading REVISION file (Capistrano)
APP_REVISION = File.read(Rails.root.join("REVISION")).strip rescue nil
```

Useful for error tracking and deployment verification.

## Usage

```ruby
# config/application.rb
config.revision = ENV["GIT_SHA"]
# or
config.revision = -> { File.read("BUILD_ID").strip }
```

```ruby
Rails.app.revision  # => "abc123def456"
```

## Revision Sources (priority order)
1. `config.revision` (string or proc)
2. `REVISION` file in Rails root
3. `nil`

## Integration

**Rails::Info** shows revision in the properties list.

Future: Could be added to health check endpoint.

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -538,6 +538,18 @@ Enables or disables reloading of classes only when tracked files change. By defa
 
 Causes the app to not boot if a master key hasn't been made available through `ENV["RAILS_MASTER_KEY"]` or the `config/master.key` file.
 
+#### `config.revision`
+
+Sets the application revision for deployment tracking and error reporting. Can be a string or a proc. When not set, reads from `REVISION` file in the application root (default: `nil`).
+
+```ruby
+config.revision = ENV["GIT_SHA"]
+# or
+config.revision = -> { File.read("BUILD_ID").strip }
+```
+
+Revision can be accessed via `Rails.app.revision`.
+
 #### `config.sandbox_by_default`
 
 When `true`, rails console starts in sandbox mode. To start rails console in non-sandbox mode, `--no-sandbox` must be specified. This is helpful to avoid accidental writing to the production database. Defaults to `false`.
--- a/railties/lib/rails/app_version.rb
+++ b/railties/lib/rails/app_version.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Rails
+  # Manages the application revision for deployment tracking and error reporting.
+  #
+  # The revision is determined in order of precedence:
+  # 1. Custom proc/string via config.revision
+  # 2. REVISION file in application root (created by deployment tools like Capistrano)
+  # 3. nil if none available
+  module AppVersion
+    class << self
+      def revision(config_revision)
+        @revision ||= load_revision(config_revision)
+      end
+
+      def reset!
+        @revision = nil
+      end
+
+      private
+        def load_revision(config_revision)
+          custom_revision(config_revision) || read_revision_file
+        end
+
+        def custom_revision(config_revision)
+          return unless config_revision
+
+          case config_revision
+          when Proc
+            config_revision.call&.to_s&.strip.presence
+          when String
+            config_revision.strip.presence
+          end
+        end
+
+        def read_revision_file
+          revision_file = Rails.root.join("REVISION")
+          revision_file.read.strip.presence if revision_file.exist?
+        rescue
+          nil
+        end
+    end
+  end
+end
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -13,6 +13,7 @@ require "active_support/configuration_file"
 require "active_support/parameter_filter"
 require "rails/engine"
 require "rails/autoloaders"
+require "rails/app_version"
 
 module Rails
   # An Engine with the responsibility of coordinating the whole boot process.
@@ -101,6 +102,14 @@ module Rails
     attr_accessor :assets, :sandbox
     alias_method :sandbox?, :sandbox
     attr_reader :reloaders, :reloader, :executor, :autoloaders
+
+    # Returns the application's revision (deployment identifier).
+    # Useful for error reporting and deployment verification.
+    #
+    # Set via config.revision (string or proc) or REVISION file.
+    def revision
+      @revision ||= Rails::AppVersion.revision(config.revision)
+    end
 
     delegate :default_url_options, :default_url_options=, to: :routes
 
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -24,7 +24,7 @@ module Rails
                     :content_security_policy_nonce_auto,
                     :require_master_key, :credentials, :disable_sandbox, :sandbox_by_default,
                     :add_autoload_paths_to_load_path, :rake_eager_load, :server_timing, :log_file_size,
-                    :dom_testing_default_html_version, :yjit
+                    :dom_testing_default_html_version, :yjit, :revision
 
       attr_reader :encoding, :api_only, :loaded_config_version, :log_level
 
@@ -85,6 +85,7 @@ module Rails
         @server_timing                           = false
         @dom_testing_default_html_version        = :html4
         @yjit                                    = false
+        @revision                                = nil
       end
 
       # Loads default configuration values for a target version. This includes
--- a/railties/lib/rails/info.rb
+++ b/railties/lib/rails/info.rb
@@ -93,6 +93,11 @@ module Rails
       Rails.env
     end
 
+    # The application's revision (deployment identifier)
+    property "Application revision" do
+      Rails.app.revision
+    end
+
     # The name of the database adapter for the current environment.
     property "Database adapter" do
       ActiveRecord::Base.connection_pool.db_config.adapter
--- a/railties/test/application/app_version_test.rb
+++ b/railties/test/application/app_version_test.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+require "isolation/abstract_unit"
+
+module ApplicationTests
+  class AppVersionTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    def setup
+      build_app
+    end
+
+    def teardown
+      teardown_app
+    end
+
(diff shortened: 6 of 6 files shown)


bdewater-thatch:
I like this idea! Couple of suggestions and ideas:
- Include this info in the default context for the Rails error reporter, like `:controller` and `:job` are set automatically
- In my experience nearly all Rails apps are worked on with a trunk-based development approach so things like GitLab or Mastodon are the exception, not the norm. Instead of a default "0.0.0" version, it should be nilable (like revision). 
- Perhaps even better is that the "version" is just a string that the framework doesn't attempt to give any special kind of meaning to (so no parsing), then you can use a source code revision, timestamp, code name, day of the week, or anything else as the desired "version" that makes sense for you. That would make the revision as a separate thing probably redundant.

seuros:
1. For the errors reporter, i already build and opensource [lapsoss](https://github.com/seuros/lapsoss) (the gem is few years old, but i just released) 

2. I don't understand this question. but What i do is to update ENV vars in docker or the host, if i cannot use file approch.

3. The version has to be valid Version, That not my decision , there are plenty of APM, error collector, that requires a sortable version.

additionally, i use the version in the cache... a rebel string means: people can put spaces, symbols, emojis, a tweet.

ruby Version is already constrained.

In short, i did not overengineer this implementation, as you can see i have an operational gem you can add to your app.

Those were restrictions i added after i saw many devs doing the same error with the vendored gem,

actionpack/lib/action_dispatch/middleware/app_info.rb
```diff
@@ -0,0 +1,52 @@
...
+      @environment_header = options[:environment_header] || "X-App-Environment"
+    end
+
+    def call(env)
+      status, headers, response = @app.call(env)
+
+      if Rails.application.config.app_version.add_headers
+        headers[@version_header] = version_string
+        headers[@environment_header] = Rails.application.app_environment.to_s
+      end
```
byroot: Rather than a middleware that is a noop based on config, we should only insert the middleware if the config is enabled.

actionpack/lib/action_dispatch/middleware/app_info.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      [status, headers, response]
+    end
+
+    private
+      def version_string
+        version = Rails.application.version
+        include_revision = Rails.application.config.app_version.include_revision
```
byroot: Similarly, this should be passed to the middleware `initialize`

railties/lib/rails/app_version.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+
+module Rails
+  # This module manages the application version and revision information.
+  # It provides a unified way to track and expose version details throughout
+  # the Rails application, similar to how Rails::Info exposes system information.
+  module AppVersion
+    # Semantic version representation with revision support
+    class Version < Gem::Version
```
byroot: We should be careful about inheriting a class from rubygems. Ideally we wouldn't because we don't know when it might change.

railties/lib/rails/app_version.rb
```diff
@@ -0,0 +1,123 @@
...
+          @patch = parts[2]&.to_i
+        end
+    end
+
+    class << self
+      attr_accessor :version_instance, :app_environment
+
+      def version
+        @version_instance ||= load_version
+      end
```
byroot: Maybe it's personal bias, but I've never seen any Rails app bother with actual versions. The git revision is generally plenty.

railties/lib/rails/app_version.rb
```diff
@@ -0,0 +1,123 @@
...
+          revision_file.read.strip if revision_file.exist?
+        rescue
+          nil
+        end
+
+        def read_git_revision
+          `git rev-parse HEAD 2>/dev/null`.strip.presence
+        rescue
+          nil
+        end
```
byroot, line 41: I think we should offer a proc based escape hatch so that users with different ways to expose the revision can still plug it.

actionpack/lib/action_dispatch/middleware/app_info.rb
```diff
@@ -0,0 +1,52 @@
...
+      @version_header = options[:version_header] || "X-App-Version"
+      @environment_header = options[:environment_header] || "X-App-Environment"
+    end
+
+    def call(env)
+      status, headers, response = @app.call(env)
+
+      if Rails.application.config.app_version.add_headers
+        headers[@version_header] = version_string
+        headers[@environment_header] = Rails.application.app_environment.to_s
```
byroot: I'm not convinced the environment is useful.

byroot (commented):
I'm generally in favor of having `Rails.application.revision`, as that's indeed something I've seen a lot and that is extremely useful to expose in `Rails.error` reports and such.

I'm not really convinced about the need for a `version` though, and overall I think this PR could be way simpler.

railties/lib/rails/app_version.rb
seuros: git revision are not ordered.

I use the version in telemetry and cache invalidation.


##### Caching

using Rails.application.version.major  will allow me to invalidate everything when i do a breaking change.

#### Telemetry

I can see regression from version 3.2.15 to 3.3.28 in speed.  
With revisions, f95cc38 and 554fd77 , could be anytime life cycle of the app. 554fd77 could be rails 8 and the other 8.2.

### Error handling 

Having the error aggregator shows me tha ti have a recurrent error from verion 2.1.1 of my app, is quicker to identify than a sha.

byroot: I'm not denying you have use for it. But I don't see most Rails users doing the same, and we aim to only merge features that are useful to most users.

Hence if you wish to add `revision` I'm 👍 , but I'm 👎 on `version`.

railties/lib/rails/app_version.rb
seuros: Valid concern, and i had Data before. 

But the interface is pretty stable for years.  In addition, there is 19 other place using it in rails repo. activerecord/lib/active_record/gem_version.rb for example.

The API didn't change for decades. Changing it will break the whole ruby ecosystem.

railties/lib/rails/app_version.rb
seuros: Also valid! Already handled in line 41

guides/source/configuring.md
```diff
@@ -217,6 +217,22 @@ end
 Controls whether requests should be handled concurrently. This should only
 be set to `false` if application code is not thread safe. Defaults to `true`.
 
+#### `config.app_version`
+
+Configures application version management. This allows Rails to track and expose version and environment information through various endpoints. Accepts an `ActiveSupport::OrderedOptions` object with the following settings:
+
+* `enabled`: Enable or disable version management (default: `true`)
+* `add_headers`: Add version and environment headers to HTTP responses (default: `true`)
```
zzak: This is maybe a old opinion but exposing version information to the public is probably not preferrable in some cases to make it the default

railties/lib/rails/app_version.rb
seuros: I get the confusion here. 
This pattern i saw it in many applications . 
people hardcode a single string in their initialisers or use lot of env vars.

revision is from Capistrano , it get the SHA of the commit.  REVISION file gets created by capistrano at each deploy. Kamal use the same concept but require the change to get committed locally.

version is another concept used by error reporting/APM. Datadog use DD_VERSION, Sentry use SENTRY_RELEASE.

----

The end user will be using only `Rails.applicaiton.version.to_s` (i'm open to rename)

The version will output something like `2.3.3 (554fd77)` if revision set. 
`2.3.3` if not set.

but as i said in the other comment, version for me is sortable, i can have conditions. Revisions are by default just string. 

So i'm open to rename/refine the concept if we are going to redefine things.

guides/source/configuring.md
seuros: I will make it opt-in instead.

actionpack/lib/action_dispatch/middleware/app_info.rb
seuros: app_env is context of the app not Rails.

You can have qa, dev, pre-production... 
Those application need to run exactly like production with Rails.env.production? #=> true . 

They just  have different databases and caches.

A use case could be having QA running against live database, but `Rails.application.app_environment.qa?` allow your QAs to use 4242... to process payment/refund.

railties/lib/rails/app_version.rb
byroot: There is no confusion.

railties/lib/rails/app_version.rb
```diff
@@ -0,0 +1,44 @@
...
+  # Manages the application revision for deployment tracking and error reporting.
+  #
+  # The revision is determined in order of precedence:
+  # 1. Custom proc/string via config.revision
+  # 2. REVISION file in application root (created by deployment tools like Capistrano)
+  # 3. nil if none available
+  module AppVersion
+    class << self
+      def revision(config_revision)
+        @revision ||= load_revision(config_revision)
```
byroot, line 13: This means that if it's not configured, we'll keep trying to check for the `REVISION` file. 

Given it's likely we'll expose the revision in various places (typically `Rails.error.report`) we should be able to memoize `nil`.

guides/source/configuring.md
```diff
@@ -538,6 +538,18 @@ Enables or disables reloading of classes only when tracked files change. By defa
...
 Causes the app to not boot if a master key hasn't been made available through `ENV["RAILS_MASTER_KEY"]` or the `config/master.key` file.
 
+#### `config.revision`
+
+Sets the application revision for deployment tracking and error reporting. Can be a string or a proc. When not set, reads from `REVISION` file in the application root (default: `nil`).
+
+```ruby
+config.revision = ENV["GIT_SHA"]
+# or
+config.revision = -> { File.read("BUILD_ID").strip }
```
byroot, line 548: Another issue I'm concerned about is code accessing `Rails.app.revision` early during boot. Problem is we don't have a good solution for that...

seuros:
replaced by #56490

PR #56084, opened 2025-11-03
[ci skip] [docs] Make code blocks horizontally scrollable when lines exceed the visible area instead of wrapping

shivabhusal:
### Motivation / Background

Previously, long lines in code examples wrapped to the next line, making the code harder to read. 

This change enables horizontal scrolling to maintain proper code formatting and improve readability.

---

rails-guide-code-line-wrap

---

[Click To See The Live Preview](https://shivabhusal-guide-code-block.rails-docs-preview.pages.dev/guides/association_basics#has-and-belongs-to-many)

### DIFF
--- a/guides/assets/stylesrc/_main.scss
+++ b/guides/assets/stylesrc/_main.scss
@@ -372,6 +372,13 @@ body.guide {
     word-wrap: break-word; /* Internet Explorer 5.5+ */
   }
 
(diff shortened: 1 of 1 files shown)


p8:
I think a change like this is nice, but I also think the examples code should wrap to 80 columns so you won't have to scroll.

PR #56086, opened 2025-11-03
Fix rails new to use stylesheet_link_tag "application" when --css is present

kryzhovnik:
When generating a Rails app with CSS bundling, e.g., `rails new myapp --css tailwind --js esbuild`, the browser console shows:

   GET /assets/tailwindcss net::ERR_ABORTED 404 (Not Found)

The problem:

1. CSS build tools output to `app/assets/builds/application.css`
2. The generated layout has `stylesheet_link_tag :app`, which expands to link tags for ALL .css files in app/assets/:

   <link rel="stylesheet" href="/assets/application-{hash}.css">
   <link rel="stylesheet" href="/assets/application.tailwind-{hash}.css">

3. The source file application.tailwind.css shouldn't be served directly: only the built output (application.css) should be linked.

This fix:
- Adds `using_css_bundling?` helper to detect when CSS bundlers are used
- Updates layout template to use "application" when --css is present
- Applies to all CSS bundlers (tailwind, sass, postcss, bootstrap, bulma)

Related: rails/cssbundling-rails#175, rails/cssbundling-rails@1332e42

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -536,6 +536,10 @@ module Rails
         using_js_runtime? && %w[bun].include?(options[:javascript])
       end
 
+      def using_css_bundling?
+        options[:skip_asset_pipeline] || options[:css]
+      end
+
       def capture_command(command, pattern = nil)
         output = `#{command}`
         if pattern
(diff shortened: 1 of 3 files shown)


railties/lib/rails/generators/app_base.rb
```diff
@@ -536,6 +536,10 @@ def using_bun?
         using_js_runtime? && %w[bun].include?(options[:javascript])
       end
 
+      def using_css_bundling?
+        options[:skip_asset_pipeline] || options[:css]
```
brunoprietog, line 540: This is not correct. You can use the `tailwindcss-rails` gem without using CSS Bundling, and this change will not include the `tailwind.css` file that is generated after Tailwind does the build.

brunoprietog: I've opened #56111 to fix this, cc @guilleiguaran

PR #56092, opened 2025-11-05
Document change in config for open redirects

that-jill:
### Motivation / Background

The behavior introduced in https://github.com/rails/rails/pull/55496 is fantastic and totally makes sense. It felt worth calling out in the upgrade guides, just for people who may be using `load_defaults` in their applications so would now be getting raises, regardless of whether they'd previously configured something else.

### Checklist

### DIFF
--- a/guides/source/upgrading_ruby_on_rails.md
+++ b/guides/source/upgrading_ruby_on_rails.md
@@ -113,6 +113,10 @@ For more information on changes made to Rails 8.1 please see the [release notes]
 
 Active Record now alphabetically sorts table columns in `schema.rb` by default, so dumps are consistent across machines and don’t flip-flop with migration order -- meaning fewer noisy diffs. `structure.sql` can still be leveraged to preserve exact column order. [See #53281 for more details on alphabetizing schema changes.](https://github.com/rails/rails/pull/53281)
 
+### More configuration for open redirects
+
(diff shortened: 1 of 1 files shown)


guides/source/upgrading_ruby_on_rails.md
```diff
@@ -113,6 +113,10 @@ For more information on changes made to Rails 8.1 please see the [release notes]
 
 Active Record now alphabetically sorts table columns in `schema.rb` by default, so dumps are consistent across machines and don’t flip-flop with migration order -- meaning fewer noisy diffs. `structure.sql` can still be leveraged to preserve exact column order. [See #53281 for more details on alphabetizing schema changes.](https://github.com/rails/rails/pull/53281)
 
+### More configuration for open redirects
+
+Rails 8.1 replaced `raise_on_open_redirects` with `action_on_open_redirects`, adding `:log` and `:notify` actions. This also changed default behavior for apps using `load_defaults` -- you app will now raise if it detects open redirects. Using the new configuration will update the behavior. [See #55496 for more details.](https://github.com/rails/rails/pull/55496)
```
skipkayhil, line 118: > This also changed default behavior for apps using `load_defaults`

Have we seen this happen for applications upgrading from 8.0 to 8.1 or only on applications running edge? (if the former, there's a bug we need to fix, if the latter, it may have been fixed already)

that-jill:
this was fixed in https://github.com/rails/rails/commit/808ca69b7e9d973c46541a9176fded00560ca729 -- thanks!

PR #56093, opened 2025-11-05
Fix Arguments#serialize missing custom serializers

### LINKED ISSUE #56090: Custom serializer implementation from guides doesn't work in Rails 8.1
Custom serializers appear to be broken in Rails 8.1. The approach detailed in the Rails guides adds the custom serializers to the config, but does so **after** the Rails initialization process calls `ActiveJob::Serializers.add_serializers Rails.application.config.active_job.custom_serializers`. The result is that `ActiveJob::Serializers` array is out of sync with the config.

### Steps to reproduce

Repro here: https://github.com/brian-kephart/serializers
This repro is a basic implementation of custom serializers per the Rails guides. There are a couple of alternative approaches in comments that can be evaluated as well.

### Expected behavior

Serialization should succeed

### Actual behavior

`ActiveJob::SerializationError` when trying to use custom serializers

### System configuration

**Rails version**: 8.1.0, 8.1.1

skipkayhil:
Fix https://github.com/rails/rails/issues/56090

Previously, custom serializer configuration was [moved][1] from `after_initialize` to `on_load(:active_job)` because the usage of `Serializer#klass` during boot meant loading custom serializers was now more likely to load autoloaded classes. However, this caused a problem for users of `ActiveJob::Arguments.serialize` since it is a documented public method and can be called without loading `ActiveJob::Base`.

This commit fixes the issue by giving `ActiveJob::Arguments` its own load hook. Now, `ActiveJob::Arguments.serialize` will correctly have custom serializers configured.

Adding this load hook also required resolving a circular dependency between ActiveJob::Serializers and ActiveJob::Arguments. Previously, the OBJECT_SERIALIZER_KEY was found under Arguments, so the ObjectSerializer singletons would cause Arguments to be loaded as soon as ActiveJob::Serializers is referenced (such as when subclassing ObjectSerializer). This commit moves OBJECT_SERIALIZATION_KEY under Serializers so that Arguments depends on Serializers but never the opposite.

[1]: https://github.com/rails/rails/commit/b7cd3018ef4c39fdc7fc0c22fb06b8a0cd035c12

### DIFF
--- a/activejob/lib/active_job/serializers.rb
+++ b/activejob/lib/active_job/serializers.rb
@@ -38,7 +38,7 @@ module ActiveJob
       # Will look up through all known serializers.
       # If no serializer found will raise <tt>ArgumentError</tt>.
       def deserialize(argument)
-        serializer_name = argument[Arguments::OBJECT_SERIALIZER_KEY]
+        serializer_name = argument[OBJECT_SERIALIZER_KEY]
         raise ArgumentError, "Serializer name is not present in the argument: #{argument.inspect}" unless serializer_name
 
         serializer = serializer_name.safe_constantize
@@ -99,6 +99,9 @@ module ActiveJob
           end
         end
     end
+
+    # :nodoc:
+    OBJECT_SERIALIZER_KEY = "_aj_serialized"
 
     add_serializers SymbolSerializer.instance,
       DurationSerializer.instance,
--- a/railties/test/application/active_job_railtie_test.rb
+++ b/railties/test/application/active_job_railtie_test.rb
@@ -20,5 +20,23 @@ module ApplicationTests
 
       assert_equal "false", rails("runner", "p FooJob.enqueue_after_transaction_commit").strip
     end
+
+    test "custom serializers are loaded for Arguments#serialize" do
+      app_file "config/initializers/custom_serializers.rb", <<~RUBY
+        class Money; end
+
+        class MoneySerializer < ActiveJob::Serializers::ObjectSerializer
+          def klass = Money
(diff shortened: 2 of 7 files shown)


activejob/lib/active_job/serializers.rb
```diff
@@ -100,14 +100,18 @@ def index_serializers(new_serializers)
...
-      DateTimeSerializer.instance,
-      DateSerializer.instance,
-      TimeWithZoneSerializer.instance,
-      TimeSerializer.instance,
-      ModuleSerializer.instance,
-      RangeSerializer.instance,
-      BigDecimalSerializer.instance
+    private
+      # :nodoc:
+      OBJECT_SERIALIZER_KEY = "_aj_serialized"
```
rafaelfranca: This isn't private, so why move here? `private` doesn't change the constant visibility. Also, neither do anything with `add_serializers`

rafaelfranca (approved):

railties/test/application/active_job_railtie_test.rb
```diff
@@ -20,5 +20,23 @@ class FooJob < ActiveJob::Base
 
       assert_equal "false", rails("runner", "p FooJob.enqueue_after_transaction_commit").strip
     end
+
+    test "custom serializers are loaded for Arguments#serialize" do
+      app_file "config/initializers/custom_serializers.rb", <<-RUBY
```
Edouard-chin: Super small nitpick but maybe `<<~RUBY` instead? Would allow for indentation of the code for easier reading

PR #56097, opened 2025-11-06
Reset parallel test tables by deleting

djmb:
### Motivation / Background

Switch from truncating to deleting when resetting parallel test databases. This PR is an alternative to https://github.com/rails/rails/pull/56026 without the configuration option.

When doing parallel testing with many databases, truncating all the tables is very slow on MySQL. Deleting from the tables is much faster even if they contain a reasonable amount of data.

### Detail

Running on a schema with about 200 empty tables and 32 parallel databases, deleting from all tables takes about 0.3 seconds, while truncating takes 18 seconds. Even with 1000 rows to delete in each table deleting is about 2.7x faster than truncating.

On PostgreSQL, truncating the tables is slower than deleting if they have small amounts of data which is typical for test databases. Both methods are very quick anyway.

There's no difference between the two methods for SQLite, as the adapter uses DELETE for truncation anyway.

Functionally DELETE and TRUNCATE are identical on PostgreSQL, but on MySQL TRUNCATE will reset AUTO_INCREMENT counters while DELETE will not.

Tests should not be relying on them being reset and since this is only for parallel tests where test orders are going to be randomised anyway so the IDs are unlikely to be consistent between test runs anyway.

By switching to DELETE all three databases will behave identically and not reset the ID sequence counters.

If there's much more data in the tables then TRUNCATE may become faster, but in that case the tests are likely to be slow as they'll need to generate the data.

This PR also removed the SKIP_TEST_DATABASE_TRUNCATE environment variable - we no longer truncate so the name doesn't make sense and deleting should be fast enough to not need to skip.

### Additional information

Timings for resetting tables with TRUNCATE vs DELETE, based on 32 databases with 200 tables each, with 0, 100 and 1000 rows per table:
```
  ┌──────────────┬─────────┬──────────┬───────┬───────┬───────┬─────────┐
  │   Database   │  Rows/  │  Method  │  Min  │  Avg  │  Max  │ Winner  │
  │              │  Table  │          │ (sec) │ (sec) │ (sec) │         │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │ 0.269 │ 0.535 │ 0.578 │         │
  │ PostgreSQL   │    0    │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 0.038 │ 0.051 │ 0.067 │  10.6x  │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │ 0.174 │ 0.323 │ 0.408 │         │
  │ PostgreSQL   │   100   │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 0.104 │ 0.132 │ 0.150 │  2.45x  │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │ 0.511 │ 0.578 │ 0.604 │  2.48x  │
  │ PostgreSQL   │  1000   │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 1.372 │ 1.434 │ 1.568 │         │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │17.485 │18.413 │18.978 │         │
  │ MySQL        │    0    │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 0.311 │ 0.341 │ 0.356 │  53.6x  │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │16.892 │18.281 │18.845 │         │
  │ MySQL        │   100   │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 0.534 │ 1.406 │ 2.003 │  13.0x  │
  ├──────────────┼─────────┼──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ TRUNCATE │16.723 │17.939 │18.544 │         │
  │ MySQL        │  1000   │──────────┼───────┼───────┼───────┼─────────┤
  │              │         │ DELETE   │ 3.789 │ 6.635 │ 8.400 │  2.70x  │
  └──────────────┴─────────┴──────────┴───────┴───────┴───────┴─────────┘
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,12 +1,6 @@
-*   Change parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
+*   On MySQL parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
 
-    When running parallel tests, each worker gets a separate database copy. If the
-    schema is up to date, tables are now reset using `DELETE FROM` instead of `TRUNCATE`.
-
-    - On MySQL, `DELETE FROM` is significantly faster than `TRUNCATE`.
-    - For Postgresql, `DELETE FROM` is also slightly faster for tables with 100s of records,
-      though `TRUNCATE` is pretty quick as well.
-    - On SQLite there is no difference as the adapter already converted `TRUNCATE` to `DELETE FROM`.
+    Truncating on MySQL is very slow even on empty or nearly empty tables.
 
     As a result of this change auto increment counters are now no longer reset between test
     runs on MySQL and the `SKIP_TEST_DATABASE_TRUNCATE` environment variable no longer has
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,19 @@
+*   Change parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
+
+    When running parallel tests, each worker gets a separate database copy. If the
+    schema is up to date, tables are now reset using `DELETE FROM` instead of `TRUNCATE`.
+
+    - On MySQL, `DELETE FROM` is significantly faster than `TRUNCATE`.
+    - For Postgresql, `DELETE FROM` is also slightly faster for tables with 100s of records,
+      though `TRUNCATE` is pretty quick as well.
+    - On SQLite there is no difference as the adapter already converted `TRUNCATE` to `DELETE FROM`.
+
+    As a result of this change auto increment counters are now no longer reset between test
+    runs on MySQL and the `SKIP_TEST_DATABASE_TRUNCATE` environment variable no longer has
+    any effect.
+
+    *Donal McBreen*
+
 *   Fix inconsistency in PostgreSQL handling of unbounded time range types
 
     Use `-infinity` rather than `NULL` for the lower value of PostgreSQL time
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -541,7 +541,7 @@ module ActiveRecord
 
       def insert_fixtures_set(fixture_set, tables_to_delete = [])
         fixture_inserts = build_fixture_statements(fixture_set)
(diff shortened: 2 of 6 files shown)


byroot:
Which version of MySQL did you test against? Because supposedly, it got some pretty drastic improvements lately: https://dev.mysql.com/blog-archive/a-250x-improvement-to-tablespace-truncation-in-mysql-server-8-0-23/

Overall I'm OK with this, but I suspect this should be an adapter logic, e.g. `AbstractAdapter#empty_all_tables(database)`, so that adapters can use implementation specific strategies.

byroot:
Also, a bit tangential, but when we look at this at Shopify, at some point rather than to empty table on test start, we did attempt to checksum the tables to ensure the inserted fixtures were still valid. Problem is it doesn't play very well with dybamic fixtures such as `created_at: <%= 10.minutes.ago %>`.

djmb:
@byroot - I tested against MySQL 8.0.43, and Postgres 18.0.

Adding `empty_all_tables` to the adapter seems a bit dangerous, but I suppose there's already `drop_database` there if you are looking for footguns.

If we go with that, we could stick with truncate for Postgres I guess? It doesn't matter a huge amount, either way is pretty fast. 

It would also mean we don't have to use send to call `execute_batch`, though maybe it should be a public method anyway - it seems like something that could be generally useful?

byroot:
I don't see what's dangerous about it.



> `execute_batch`, though maybe it should be a public method anyway - it seems like something that could be generally useful?

It was considered, but there are security implications about it, so I'm not big on it, and also it's designed as internal API, so if we wanted to expose batch capabilities to users, we should think about a proper API rather than just flip the visibility of an internal method.

djmb:
@byroot - thanks for the review. I've added `empty_all_tables` to the `AbstractAdapter` now and defaulted to truncating. I've only changed the MySQL adapter now to use delete - there's not much difference between delete and truncate on Postgresql.

This all means that also we won't automatically change how it works for third part adapters which is probably for the best.

PR #56100, opened 2025-11-06
Introduce ActiveRecord::Result#to_csv

schmijos:
I was annoyed whenever I needed to come up with CLI fiddlery like this one. I can never remember the API.

```rb
result = ActiveRecord::Base.connection.exec_query(query)
CSV.generate do |csv|
  csv << result.columns
  result.each do |row|
    csv << row.values
  end
end
```

So I propose a new method `to_csv` to be answered by `ActiveRecord::Result`. So that I can comfortably get my CSV.

```rb
result = ApplicationRecord.connection.exec_query("SELECT id, title, body FROM posts")
puts result.to_csv
# id,title,body
# 1,title_1,body_1
# 2,title_2,body_2
# ...
```

There is one downside: The `csv` gem is not a default gem anymore since Ruby 3.4. So I need to add it to the gemspec.


activerecord/activerecord.gemspec
```diff
@@ -38,4 +38,5 @@ Gem::Specification.new do |s|
   s.add_dependency "activesupport", version
   s.add_dependency "activemodel",   version
   s.add_dependency "timeout", ">= 0.4.0"
+  s.add_dependency "csv", ">= 1.0.0"
```
schmijos, line 41: Stable API for `generate` since then.

fatkodima:
Thanks, but this is better implemented as a helper in the app, because then we need to_json, to_xml etc.

PR #56101, opened 2025-11-06
Make schema cache backward compatible with the changes in column

beauraF:
### Motivation / Background

We recently rolled out Rails 8.1 to production and encountered a backward compatibility issue when older Rails 8.0 processes load a schema cache generated under Rails 8.1.

### Detail

In Rails 8.1, the default value for a column is now assigned like this (see https://github.com/rails/rails/commit/205cdd50ba626325b26ce2919d5e961c6df50df0):

```ruby
@default = default.nil? || cast_type.mutable? ? default : cast_type.deserialize(default)
```

This means that when default is `nil`, `@default` is set to `true`. Later on, during deduplication, Rails handles this case correctly within 8.1 https://github.com/rails/rails/blob/205cdd50ba626325b26ce2919d5e961c6df50df0/activerecord/lib/active_record/connection_adapters/column.rb#L120

However, if a schema cache produced by 8.1 is deserialized by Rails 8.0, the `default = true` value is processed by Rails 8.0 code that expects default to always be either a `string` or `nil`
https://github.com/rails/rails/blob/433c70c470c2bbaddfc29cde40966b9700cccfe7/activerecord/lib/active_record/connection_adapters/column.rb#L107

Since `default` is now `true`, this triggers:

```
NoMethodError: undefined method `-@' for true:TrueClass
```

To fix it, I propose to backport @byroot's changes in `Column#deduplicated`.  It should restore backward compatibility, with no impact on 8.0 application.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/column.rb
+++ b/activerecord/lib/active_record/connection_adapters/column.rb
@@ -104,7 +104,7 @@ module ActiveRecord
         def deduplicated
           @name = -name
           @sql_type_metadata = sql_type_metadata.deduplicate if sql_type_metadata
-          @default = -default if default
+          @default = -default if String === default
           @default_function = -default_function if default_function
           @collation = -collation if collation
           @comment = -comment if comment
--- a/activerecord/test/assets/schema_dump_8_1.yml
+++ b/activerecord/test/assets/schema_dump_8_1.yml
@@ -1,0 +1,40 @@
+--- !ruby/object:ActiveRecord::ConnectionAdapters::SchemaCache
+columns:
+  posts:
+  - &1 !ruby/object:ActiveRecord::ConnectionAdapters::Column
+    name: id
+    table_name: posts
+    sql_type_metadata: !ruby/object:ActiveRecord::ConnectionAdapters::SqlTypeMetadata
+      sql_type: INTEGER
+      type: :integer
+      limit:
+      precision:
+      scale:
+    'null': false
+    default:
+    default_function:
+    collation:
+    comment:
+  - &2 !ruby/object:ActiveRecord::ConnectionAdapters::Column
+    name: published
+    table_name: posts
+    sql_type_metadata: !ruby/object:ActiveRecord::ConnectionAdapters::SqlTypeMetadata
+      sql_type: boolean
+      type: :boolean
+      limit:
+      precision:
+      scale:
+    'null': true
+    default: true
+    default_function:
+    collation:
+    comment:
+columns_hash:
+  posts:
+    id: *1
+    published: *2
+primary_keys:
+  posts: id
+data_sources:
+  posts: true
+version: 0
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/connection_adapters/column.rb
```diff
@@ -104,7 +104,7 @@ def virtual?
         def deduplicated
           @name = -name
           @sql_type_metadata = sql_type_metadata.deduplicate if sql_type_metadata
-          @default = -default if default
+          @default = -default if String === default
```
byroot, line 107:
```suggestion
          @default = String === default ? -default : default
```

This is a bit defensive, but I think we should consider `default` could be anything.

activerecord/test/assets/schema_dump_8_1.yml
```diff
@@ -0,0 +1,40 @@
...
+    name: published
+    table_name: posts
+    sql_type_metadata: !ruby/object:ActiveRecord::ConnectionAdapters::SqlTypeMetadata
+      sql_type: boolean
+      type: :boolean
+      limit:
+      precision:
+      scale:
+    'null': true
+    default: true
```
byroot, line 28: Actually, that's the case here, `default` is `true` so we must assign it.

activerecord/lib/active_record/connection_adapters/column.rb
byroot: Nevermind 🤦 , `deduplicated` is called alter, so `@default` is already assigned.

rafaelfranca:
Can you help me understand how the 8.1 schema was loaded by a process running 8.0? The right way to do an upgrade is to only generate the 8.1 schema after you know there is no process running with 8.0. It doesn't make much sense to change a already released version of Rails to work with a newer version of code.

Also this isn't making the cache backwards compatible, it is making 8.0 forward compatible, and this we never do.

beauraF:
> Can you help me understand how the 8.1 schema was loaded by a process running 8.0? The right way to do an upgrade is to only generate the 8.1 schema after you know there is no process running with 8.0. It doesn't make much sense to change a already released version of Rails to work with a newer version of code.
> 
> Also this isn't making the cache backwards compatible, it is making 8.0 forward compatible, and this we never do.

We’re using k8s [ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) to distribute the schema cache. When we roll out a new version, the `ConfigMap` gets updated and is mounted into all pods, including pods still running the previous Rails version.

This wasn’t really an issue before because schema cache was eagerly loaded at boot. But since moving to a multi-database setup, schema cache loading is now [lazy](https://guides.rubyonrails.org/active_record_multiple_databases.html#schema-caching), so a pod running 8.0 can end up loading a cache generated by 8.1.

We also generate the schema cache as part of the deployment process, whether there is a migration to run or not. So the cache change on every deploy, even if the schema itself didn’t evolve.

Overall, this is coming from our deployment setup rather than Rails itself.

> Also this isn’t making the cache backwards compatible, it is making 8.0 forward compatible, and this we never do.

Indeed. I understand your position. The risk is now identified on our side and we’ll adapt our schema cache distribution strategy. 

Thanks for clarifying @rafaelfranca 🙏

PR #48993, opened 2023-08-21
Make `has_secure_password` pluginable/extendable with different password hashing algorithms

f3ndot:
### Motivation / Background

Based on an idea by @rafaelfranca in https://github.com/rails/rails/issues/41420#issuecomment-1686745412.

While BCrypt may be the desirable default password hashing algorithm in Rails for the foreseeable future, making the mechanism for hashing and verifying passwords in `has_secure_password` extendable/modular is valuable.

Developers may opt for different hashing algorithms and can now extend or plug in their own alongside BCrypt in a clean and maintainable way. From the referenced thread, there is discourse amongst people on what should be the _de facto_ password algorithm and disagreement on whether Argon2 should be it. Rails should make it easier for the community to progress on best password hashing practices.

This also unlocks Rails shipping with more than BCrypt as an available hashing algorithm.

I _could_ see a world where Rails offers PBKDF2 (for those wanting FIPS 140 compliance), BCrypt (the safe default), and Argon2 (since OWASP recommends it).

### Detail

This PR takes BCrypt-specific code in `ActiveModel::SecurePassword` and extracts it out into an object `ActiveModel::SecurePassword::BCrypt`. The object adheres to a simple interface (`ActiveModel::SecurePassword::Base`) so developers can supplement with different algorithms (or different libraries of the same underlying algo). 

A user of the Rails framework uses the newly added `:algorithm` option in `has_secure_password` to declare what password hashing algo should be used on a given attribute.

Developers can write libraries or even initializers to add different password hashing algorithm support following a very clean interface.

Below is an example of someone opting to use Argon2 using `ruby-argon2`:

```ruby
# config/initializers/add_argon2_to_activemodel_securepassword.rb

require "active_model/secure_password/base"

module ActiveModel
  module SecurePassword
    class Argon2id < Base
      def initialize
        require "argon2"
      rescue LoadError
        warn "You don't have argon2 installed in your application. Please add it to your Gemfile and run bundle install."
        raise
      end

      def self.algorithm_name
        :argon2id
      end

      def hash_password(unencrypted_password, options = {})
        if options[:min_cost]
          hasher = Argon2::Password.new(t_cost: 1, m_cost: 3, p_cost: 1)
        else
          hasher = Argon2::Password.new(t_cost: 2, m_cost: 19, p_cost: 1)  # Latest OWASP recommended values
        end
        hasher.create(unencrypted_password)
      end

      def verify_password(password, digest)
        Argon2::Password.verify_password(password, digest)
      end

[...]

PR #37628, opened 2019-11-03
Support cookie migration in distributed settings.

qnighy:
### Summary

In a distributed configuration like rolling update or multiple Rails servers sharing sessions, users may observe both old and new instances during deployment.
Users may be served by a new instance and then by an old instance.

That means, users may lose their sessions if they access the server during deployment in the following cases:

- The server changed `cookies_serializer` from `:marshal` to `:hybrid`.
  - Reason: `:hybrid` generates a JSON-serialized session, but `:marshal` can't read it.
- The server changed `use_authenticated_cookie_encryption` from `false` to `true`.
  - Reason: `use_authenticated_cookie_encryption: true` generates a GCM-encrypted session, but `use_authenticated_cookie_encryption: false` can't read it.

This patch adds fallbacks to properly downgrade the cookie format when necessary during deployment.

Generally speaking, there are three ways to address this:

- (A) When we see a new cookie format in an old server, accept the cookie and always overwrite the cookie in the old format.
- (B) When we see a new cookie format in an old server, accept the cookie and write the cookie in the old format if necessary.
- (C) When we see a new cookie format in an old server, accept the cookie and write the cookie in the new format if necessary.

Any of them is fine and I chose the easiest way. (B) is easiest for the serializer case and (A) is easiest for the encryption case. (C) is ideal for minimizing the number of cookie writes.

PR #56104, opened 2025-11-07
Add a top-level switch for has_secure_password algorithm and default to argon2

zzak:
Follow up to #56057 per:
https://github.com/rails/rails/pull/48993#issuecomment-2266154188

> the bigger value would be if we added support for argon2id as an alternative, and then had a top-level switch to set the default algorithm, so it can be set for new apps to be argon2id

### DIFF
--- a/railties/lib/rails/generators/test_unit/authentication/templates/test/fixtures/users.yml.tt
+++ b/railties/lib/rails/generators/test_unit/authentication/templates/test/fixtures/users.yml.tt
@@ -1,4 +1,6 @@
-<%% password_digest = BCrypt::Password.create("password") %>
+<%% algorithm = Rails.application.config.active_model.secure_password_algorithm || :argon2 %>
+<%% password_algorithm = ActiveModel::SecurePassword.lookup_algorithm(algorithm).new %>
+<%% password_digest = password_algorithm.hash_password("password") %>
 
 one:
   email_address: one@example.com
--- a/railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
+++ b/railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
@@ -4,7 +4,9 @@
 <%= name %>:
 <% attributes.each do |attribute| -%>
   <%- if attribute.password_digest? -%>
-  password_digest: <%= BCrypt::Password.create("secret") %>  # Generated with BCrypt::Password.create("secret")
+  <%- algorithm = Rails.application.config.active_model.secure_password_algorithm || :argon2 -%>
+  <%- password_algorithm = ActiveModel::SecurePassword.lookup_algorithm(algorithm).new -%>
+  password_digest: <%= password_algorithm.hash_password("secret") %>  # Generated with <%= password_algorithm.class.name %>.new.hash_password("secret")
   <%- elsif attribute.reference? -%>
   <%= yaml_key_value(attribute.column_name.delete_suffix("_id"), attribute.default || name) %>
   <%- elsif !attribute.virtual? -%>
--- a/activemodel/lib/active_model/railtie.rb
+++ b/activemodel/lib/active_model/railtie.rb
@@ -13,9 +13,10 @@ module ActiveModel
       app.deprecators[:active_model] = ActiveModel.deprecator
     end
 
-    initializer "active_model.secure_password" do
+    initializer "active_model.secure_password" do |app|
       ActiveSupport.on_load(:active_model_secure_password) do
         ActiveModel::SecurePassword.min_cost = Rails.env.test?
+        ActiveModel::SecurePassword.algorithm = app.config.active_model.secure_password_algorithm
       end
     end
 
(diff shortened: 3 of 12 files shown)


zzak:
Oh, I forgot a changelog but I want to see if CI will pass first

railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
```diff
@@ -4,7 +4,7 @@
 <%= name %>:
 <% attributes.each do |attribute| -%>
   <%- if attribute.password_digest? -%>
-  password_digest: <%= BCrypt::Password.create("secret") %>  # Generated with BCrypt::Password.create("secret")
+  password_digest: <%= Argon2::Password.create("secret") %>  # Generated with Argon2::Password.create("secret")
```
morgoth: Shouldn't this follow the configuration? For folks using Bcrypt, this will always have to be changed manually

railties/lib/rails/generators/test_unit/authentication/templates/test/fixtures/users.yml.tt
```diff
@@ -1,4 +1,4 @@
-<%% password_digest = BCrypt::Password.create("password") %>
+<%% password_digest = Argon2::Password.create("password") %>
```
guilleiguaran:
```suggestion
<%% algorithm = Rails.application.config.active_model.secure_password_algorithm || :argon2 %>
<%% password_algorithm = ActiveModel::SecurePassword.lookup_algorithm(algorithm).new %>
<%% password_digest = password_algorithm.hash_password("password") %>
```

railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
guilleiguaran:
```suggestion
  <%- algorithm = Rails.application.config.active_model.secure_password_algorithm || :argon2 -%>
  <%- password_algorithm = ActiveModel::SecurePassword.lookup_algorithm(algorithm).new -%>
  password_digest: <%= password_algorithm.hash_password("secret") %>  # Generated with <%= password_algorithm.class.name %>.new.hash_password("secret")
```

ezekg:
Won't this now break existing implementations of `has_secure_password` by default, since Rails will now try to verify passwords using Argon2 and not Bcrypt? If the default is going to change, I feel like it should be backward-compatible in the same way a new encryption algorithm would be, e.g. by default try Argon2 and then Bcrypt, unless configured to only use one of them.

I feel like this change has too much potential impact on nearly all existing Rails apps using Bcrypt. In addition, the ability to migrate from Bcrypt to Argon2, gradually, would be super nice, as I've always hated the length limit of Bcrypt.

railties/lib/rails/generators/test_unit/model/templates/fixtures.yml.tt
zzak: Yeah good catch, I committed the fix from Guillermo. Thanks for your review!

zzak:
@ezekg It's a valid concern, but the default is not changing for Active Model the default is still bcrypt. If you use `config.load_defaults(8.2)` you will get the new default `:argon2` instead.

Apps that want to keep using bcrypt must explicitly set `config.active_model.secure_password_algorithm = :bcrypt`.
I've tried to update the changelog to reflect this, suggestions welcome. :pray:

ezekg:
Consider this scenario: somebody upgrades Rails and enables the 8.2 defaults, like I do when upgrading Rails. All tests eventually pass, because new users use argon2, and the app uses argon2. You deploy to production but now old users cannot login because they use bcrypt. You rollback the broken deploy, but then realize users that signed up within that timeframe use argon2 but the app now uses bcrypt, meaning they cannot login after the rollback. 😃

zzak:
The recommended upgrade process regarding load_defaults is clearly documented here:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults

Blindly updating load_defaults without understanding how it will impact your application is not a valid argument IMO

You would also have to add `argon2` to your Gemfile before the app would even boot in production, so I think this even further the signals to a developer they need to pay attention to the details.

I understand your concern though, but there is only so much we can do and this is the safest way for the framework to move forward.

ojab:
>I understand your concern though, but there is only so much we can do and this is the safest way for the framework to move forward.

Have you considered retaining backward compatibility via updating passwords to the new hashing alg without breaking compatibility with the old one (e. g. like https://github.com/rails/rails/pull/37628 was implemented)?

I don't quite see how existing apps could be realistically updated to this new framework default except 
- hard-switch (and all the passwords needs to be reset)
- implementing custom `hybrid` scheme via  (pseudocode):
```ruby
def verify_password
  return true if Argon2::Password.verify_password(password, digest)

  if BCrypt::Password.new(digest).is_password?(password)
    return update!(password:)
  end

  false
end
```
that's quite hard to write correctly (considering possible timing attacks/etc).

zzak:
I don't think so no, because this is intended for new applications or possibly introducing `has_secure_password` to an existing application to adopt the new default, existing applications already using bcrypt can continue to do so. There is no reason to migrate their applications, that's not what we're trying to do here.

PR #56107, opened 2025-11-07
Strict loading violation is raised when mode is n_plus_one_only with strict loading by default, fixes #56094

### LINKED ISSUE #56094: Strict loading with `:n_plus_one_only` is not detecting N+1 queries
### Steps to reproduce

Configure Active Record's Strict Loading to be `:n_plus_one_only` with Strict Loading by default:

```diff
+  config.active_record.strict_loading_by_default = true
+  config.active_record.strict_loading_mode = :n_plus_one_only
```

Create multiple records with associated `has_many …` records (in the script below, `Post` <- `Comment`).

Iterate over the list of `Post` records, querying each for its associated `Comment` records. With two `Post` records and no eager loading, that should result in 3 total queries (1 for all `Post` records, then 1 for the collection of each `Post` records' associated `Comment` records):

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 8.1.1"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

  config.active_record.strict_loading_by_default = true
  config.active_record.strict_loading_mode = :n_plus_one_only
end
Rails.application.initialize!

    t.text :name
  end

    t.text :body
    t.integer :post_id
  end
end

  has_many :comments
end

  belongs_to :post
end

  def test_strict_loading_errors_raised
    one, two = Post.create!([ { name: "one" }, { name: "two" } ])
    Comment.create!([
      { body: "one comment", post: one },
      { body: "two comment", post: two }
    ])

    assert_raises match: /`Post` is marked for strict_loading/ do
      Post.all.flat_map { |post| post.comments.map(&:body) }
    end
  end
end
```

### Expected behavior

With both `config.active_record.strict_loading_by_default = true` and `config.active_record.strict_loading_mode = :n_plus_one_only` set, the expected behavior is that Active Record would detect the N+1 query issues, then raise a Strict Loading Violation error.

### Actual behavior

No error is raised.

When `config.active_record.strict_loading_mode = :n_plus_one_only` is replaced with `:all`, the Strict Loading Violation is raised:

```diff
-  config.active_record.strict_loading_mode = :n_plus_one_only
+  config.active_record.strict_loading_mode = :all
```

### System configuration

**Rails version**: both `8.1.1` and `branch: "rails/rails"` as of [c6de6cf](https://github.com/rails/rails/commit/c6de6cfcedbadbad1a9f1908660da554d60bd012).

**Ruby version**: `3.4.5`

tahsin352:
…

### Motivation / Background

This Pull Request has been created because it fixes #56094 

### Detail
This PR  allows to configure Active Record's Strict Loading to be :n_plus_one_only with Strict Loading by default. It now raises strict loading violations error.

After this fix, now the below script passes:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 8.1.1"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

  config.active_record.strict_loading_by_default = true
  config.active_record.strict_loading_mode = :n_plus_one_only
end
Rails.application.initialize!

    t.text :name
  end

    t.text :body
    t.integer :post_id
  end
end

  has_many :comments
end

  belongs_to :post
end

  def test_strict_loading_errors_raised
    one, two = Post.create!([ { name: "one" }, { name: "two" } ])
    Comment.create!([
      { body: "one comment", post: one },
      { body: "two comment", post: two }
    ])

    assert_raises match: /`Post` is marked for strict_loading/ do
      Post.all.flat_map { |post| post.comments.map(&:body) }
    end
  end
end
```

With both `config.active_record.strict_loading_by_default = true` and `config.active_record.strict_loading_mode = :n_plus_one_only` set, the expected behavior is that Active Record would detect the N+1 query issues, then raise a Strict Loading Violation error. This PR fixes it accordingly.

### Checklist


fatkodima:
Where is the fix in this PR?

tahsin352:
> Where is the fix in this PR?

Strict Loading violation is only raised [here](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/association.rb#L250). But for the above case `violates_strict_loading?` remains false due to this condition [here](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/association.rb#L291). If we remove ` && !owner.strict_loading_n_plus_one_only?` from that checking, then other test cases fails. I have incorporated the test case and it truly fails to raise error.

olimart:
@tahsin352 I think you forgot to commit the actual change.

tahsin352:
@fatkodima @olimart I have added my fixes. But one builds are failing, though I have tested with the above script, that passes with my fixes. I have integrated it as a test case also, it also passes. Thanks.

tahsin352:
All test passed

PR #37295, opened 2019-09-25
Fix autosave validation context regression

tekin:
This fixes a regression introduced by #36671 by ensuring autosaved associations always perform validations when a custom validation context is being used.

Prior to #36671, validations on a dependent singular association would always fire, regardless of whether the associated record was `changed_for_autosave?` or not. That change brought the autosave behaviour for both singular and collection associations inline such that dependent associations would only be validated/saved if they needed to be, according to `changed_for_autosave?`. The problem this highlighted/introduced is that the use of custom validation contexts became a bit broken in that the validations on the associated records now only fire if the associated records had actual changes on them, ignoring the fact that a custom context is a strong indicator that the context might be different to when the record was last save, and therefor their validity should be reconsidered.

This PR reinstates the previous behaviour for singular associations, but also updates the behaviour for collection associations so that they both behave consistently: if a custom validation context is used, the validations will always fire on the association records.

The first commit updates an existing regression test that became a false-positive when #36671 was merged.

Resolves #37192 and #36822.

Issue #37192, opened 2019-09-13
Validations failing to fire on non-dirty belongs_to associations when a custom context is used


tekin:
Since https://github.com/rails/rails/pull/36671 was introduced, validations are no longer fired on non-dirty dependent association records, even when a custom validation context is used.

Skipping the associations validations when it hasn't been marked as dirty is undesirable is that it breaks the ability to vary validations based on contexts. This test fails where previously it would have passed: 

```ruby
  def test_belongs_to_validations_fire_based_on_validation_context
    old_validators = Mouse._validators.deep_dup
    old_callbacks = Mouse._validate_callbacks.deep_dup
    Mouse.clear_validators!
    Mouse.validates :name, presence: true, on: :rename

    mouse = Mouse.create!
    squeak = Squeak.create!(mouse: mouse)
    assert_equal true, squeak.valid?
    assert_equal false, squeak.valid?(:rename) # this now fails
  ensure
    Mouse._validators = old_validators
    Mouse._validate_callbacks = old_callbacks
  end
```

Other related issues: https://github.com/rails/rails/issues/36822 and https://github.com/rails/rails/pull/36828.

Issue #36822, opened 2019-07-31
6.0.0.rc2: The behavior of validation when using `accept_nested_attributes_for` is changed, and ActiveRecord has not working validation for associated models.


sinsoku:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  if ENV["RAILS_BRANCH"]
    gem "rails", github: "rails/rails", branch: ENV["RAILS_BRANCH"]
  else
    gem "rails", "5.2.3"
  end
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table(:articles, force: true)

  create_table :top_pages, force: true do |t|
    t.integer :article_id
  end
end

class Article < ActiveRecord::Base
  has_one :top_page
  accepts_nested_attributes_for :top_page
end

class TopPage < ActiveRecord::Base
  belongs_to :article

  validate -> { errors.add(:base, 'does not publish') }, on: :publish

  def valid?(*args)
    logger.debug "valid?(*args): #{args}"
    super
  end
end

puts
puts "ActiveRecord.version = #{ActiveRecord.version}"
puts

class BugTest < Minitest::Test
  def test_to_valid
    article = Article.create
    top_page = TopPage.create
    article.top_page = top_page

    assert_equal false, article.valid?(:publish)
    assert_includes article.top_page.errors.full_messages, 'does not publish'
  end
end
```

### Expected behavior (5.2.3)

I want the validation of associated models to work as well as v5.2.3.

In v5.2.3, the `valid?` method executed three times, and the last call is passed context.

```
$ ruby issue.rb

ActiveRecord.version = 5.2.3

Run options: --seed 45000

# Running:

D, [2019-07-31T20:09:22.267533 #8937] DEBUG -- :    (0.1ms)  begin transaction
D, [2019-07-31T20:09:22.268150 #8937] DEBUG -- :   Article Create (0.1ms)  INSERT INTO "articles" DEFAULT VALUES
D, [2019-07-31T20:09:22.268418 #8937] DEBUG -- :    (0.1ms)  commit transaction
[...]

PR #36671, opened 2019-07-13
Don't validate non dirty association targets

wjessop:
Fixes #36581.

### Summary

This fixes an issue where validations would return differently after a previously saved invalid association was loaded. Here the relationship between Mouse and Squeak is:

```ruby
class Mouse < ActiveRecord::Base
  has_many :squeaks, autosave: true
  validates :name, presence: true
end

class Squeak < ActiveRecord::Base
  belongs_to :mouse
  accepts_nested_attributes_for :mouse
end
```

Currently, if we have an invalid mouse association that was previously saved at some point in the past (eg. with `mouse.save! validate: false`) then loading the association will cause the change in return value for the parent object's `valid?` method, here `squeak.valid?`):

```ruby
squeak.valid? # => true
squeak.mouse.present? # invalid mouse association is loaded
squeak.valid?  # => false
```

Limiting validations to associations that will be saved (using autosave: true) due to being changed means that loading invalid associated relations will not change the return value of the parent relation's `valid?` method, after this patch:

```ruby
squeak.valid? # => true
squeak.mouse.present? # invalid mouse association is loaded, but not changed
squeak.valid?  # => true
```

This change also updates the way save/save! works when the association is unchanged, I think for the better. Before the change:

```ruby
squeak = Squeak.first
squeak.valid? # => true
squeak.save # => true
squeak.mouse.present? # invalid mouse association is loaded, but unchanged
squeak.valid?  # => false # valid? method behaviour has changed even though we never altered the mouse association
squeak.save # => false
```

After the change:

```ruby
squeak = Squeak.first
squeak.valid? # => true
squeak.save # => true
squeak.mouse.present? # invalid mouse association is loaded
squeak.valid?  # => true # valid? method behaviour is the same because we never altered the association and it would not be saved
squeak.save # => true
```

This seems reasonable, as we are only operating on the squeak parent object. With the new code if we change the mouse association we still get the expected behaviour:

```ruby
squeak = Squeak.first
squeak.valid? # => true
squeak.mouse.present? # invalid mouse association is loaded
squeak.save # => true
squeak.mouse.name = "Jerry" # Make the mouse object valid again
squeak.save # => true
squeak.mouse.name =  nil # Re-invalidate the mouse object
squeak.save # => false
```

PR #56109, opened 2025-11-07
Fix validation context regression for unchanged associations

### LINKED ISSUE #38036: Associated models are validated when there is a custom validation context
### Steps to reproduce
After we upgraded from Rails 5.2.3 to Rails 5.2.4, one model cannot be saved due to validation errors. We found it was caused by the validation on an associated model which was not validated on Rails 5.2.3.

The problem is likely caused by the newly added `custom_validation_context?` boolean checking at https://github.com/rails/rails/pull/37295/files#diff-829fd5510b886395117cc530518ef7f7R273, which enforces validating associated collections when there is a custom validation context.

Sample code:
```
class Subscription
  validates :price, presence: true
end

class User
  has_many :subscriptions, dependent: :delete_all
end

# insert into a valid user record in the database and a subscription without price
user = User.first
user.subscriptions.first.valid? # this returns false
user.valid?(:registration_context) # this used to return true on Rails 5.2.3 and now returns false
```

### Expected behavior
Associated models should NOT be validated when there is no `validates_associated` for a relationship, as described in https://guides.rubyonrails.org/active_record_validations.html#validates-associated

### Actual behavior
Associated models are validated.

### System configuration
**Rails version**:
5.2.4
**Ruby version**:
2.6.3

PhilCoggins:
This PR fixes a regression where calling `valid?(context)` with a custom validation context incorrectly validates unchanged, persisted association records that were previously saved with `validate: false`.

## The Problem
When you save an invalid association record using `validate: false` and then later call `valid?(custom_context)` on the parent, the parent becomes invalid even though: 
1. The association record hasn't changed
2. The parent record itself is valid
3. The default `valid?` call (without context) correctly passes

## Root Cause
The regression was introduced in #37295, which added `|| custom_validation_context?` checks in three places:
* `#validate_has_one_association` (line 332)
* `#validate_belongs_to_association` (line 346)
* `#associated_records_to_validate_or_save` (line 299)

The intent of #37295 was to ensure validations fire when using custom contexts because validation logic might be context-dependent. However, this made the assumption too broad - any custom validation context now causes all loaded association records to be validated, including unchanged, persisted records. This breaks the principle established in #36671 that associations should only be validated if they've actually changed (`changed_for_autosave?`).

## The Solution
Remove the `|| custom_validation_context?` condition from all three validation methods. This restores the behavior where associations are only validated if they have been modified, regardless of whether a custom validation context is being used.
While this means we lose the ability to re-validate unchanged associations when their validation logic might depend on the parent's state (the original motivation for #37295), that edge case should *be handled explicitly by the application*, rather than assuming all custom contexts require full re-validation of unchanged records.
Changes
Remove `|| custom_validation_context?` from `#validate_has_one_association`
Remove `|| custom_validation_context?` from `#validate_belongs_to_association`
Remove `|| custom_validation_context?` from `#associated_records_to_validate_or_save`
Add regression test in test/cases/validations/association_validation_test.rb

## Related Issues
Fixes #38036
Partially reverts #37295 (which was fixing #37192 and #36822)
Restores behavior from #36671

### DIFF
--- a/activerecord/test/cases/autosave_association_test.rb
+++ b/activerecord/test/cases/autosave_association_test.rb
@@ -2272,11 +2272,18 @@ class TestAutosaveAssociationValidationsOnAHasManyAssociation < ActiveRecord::Te
     assert_equal(book_count_before_save, Book.count)
   end
 
-  def test_validations_still_fire_on_unchanged_association_with_custom_validation_context
+  test "validations still fire on unchanged association with validates_associated and custom validation context" do
     pirate = FamousPirate.create!(catchphrase: "Avast Ye!")
-    pirate.famous_ships.create!
+    FamousPirate.validates_associated :famous_ships
+    ship = pirate.famous_ships.new
+    ship.save(validate: false)
 
-    assert_predicate pirate, :valid?
+    pirate.reload
+    assert pirate.valid?
+    assert_not pirate.valid?(:conference)
+
+    pirate.famous_ships.load
+    assert pirate.valid?
     assert_not pirate.valid?(:conference)
   end
 end
@@ -2318,13 +2325,6 @@ class TestAutosaveAssociationValidationsOnABelongsToAssociation < ActiveRecord::
     assert_predicate @pirate, :valid?
     @pirate.non_validated_parrot = Parrot.new(name: "")
     assert_predicate @pirate, :valid?
-  end
-
-  def test_validations_still_fire_on_unchanged_association_with_custom_validation_context
-    firm_with_low_credit = Firm.create!(name: "Something", account: Account.new(credit_limit: 50))
-
-    assert_predicate firm_with_low_credit, :valid?
-    assert_not firm_with_low_credit.valid?(:bank_loan)
   end
 end
 
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,37 @@
+*   Fix validation regression when using custom context with unchanged associations.
+
+    When calling `valid?(custom_context)` on a record with loaded but unchanged
+    associations, ActiveRecord was incorrectly validating those association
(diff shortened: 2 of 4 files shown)


activerecord/test/cases/autosave_association_test.rb
```diff
@@ -2271,14 +2271,6 @@ def setup
     assert_equal(author_count_before_save, Author.count)
     assert_equal(book_count_before_save, Book.count)
   end
-
-  def test_validations_still_fire_on_unchanged_association_with_custom_validation_context
-    pirate = FamousPirate.create!(catchphrase: "Avast Ye!")
-    pirate.famous_ships.create!
-
-    assert_predicate pirate, :valid?
-    assert_not pirate.valid?(:conference)
```
rafaelfranca, line 2280: Can we add a test similar to this but with the `validate_associated` defined in the model? Maybe we already have it, but I didn't look.

PhilCoggins: Done 👍

activerecord/test/cases/autosave_association_test.rb
```diff
@@ -2272,12 +2272,19 @@ def setup
...
 
     assert_predicate pirate, :valid?
-    assert_not pirate.valid?(:conference)
+    assert pirate.valid?(:conference)
+
+    pirate.reload
+    pirate.famous_ships.load
+
+    assert_predicate pirate, :valid?
+    assert pirate.valid?(:conference)
```
rafaelfranca: I think what we want is for this to return false when the famous_ship don't have a name. For that we would need to add it to the database without validating

PhilCoggins: Ah right, good catch.

PhilCoggins:
Bumping @rafaelfranca , anything else we'd like to see here?

PR #56113, opened 2025-11-08
Add configuration option for negative scopes to include records with nil values

QWYNG:
### Motivation / Background

In #55912 and c08989565fc7119ada007f8a20621db8df6f50a7, the behavior of negative scopes for enums was changed to include records with `nil` values using `IS DISTINCT FROM` operator. 
While this is a more correct SQL behavior, it's a breaking change for existing applications that may rely on the previous behavior of excluding `nil` values.

This Pull Request provides a configuration option to allow applications to choose between the new behavior (including `nil` values) and the old behavior (excluding `nil` values), giving developers control over this behavior change.

### Detail

This Pull Request adds a new configuration option `config.active_record.deprecated_negative_enum_scopes_exclude_nil`:

- When set to `false` (default for Rails 8.2+), negative scopes for enums will include records with `nil` values using `IS DISTINCT FROM`
- When set to `true`, negative scopes will exclude records with `nil` values using `WHERE.NOT`

```ruby
  enum :status, [:proposed, :written, :published]
end

book1 = Book.create!(status: :published)
book2 = Book.create!(status: :written)
book3 = Book.create!(status: nil)

# With deprecated_negative_enum_scopes_exclude_nil = false (8.2 default)
Book.not_published # => [book2, book3]

# With deprecated_negative_enum_scopes_exclude_nil = true
Book.not_published # => [book2]
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -6,6 +6,11 @@
     When set to `false` (8.2 default), negative scopes will include records with `nil`
     values. When set to `true`, negative scopes will exclude records with `nil` values.
 
+    This configuration is deprecated and will be removed in a future version of Rails.
+    It is intended as a temporary measure to ease upgrades. Setting this to `true` will
+    trigger a deprecation warning. Applications should instead manually override the
+    negative scope in their models if they need to maintain the old behavior.
+
     *QWYNG*
 
 *   Fix bug when `current_transaction.isolation` would not have been reset in test env.
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Add `deprecated_negative_enum_scopes_exclude_nil` configuration option.
+
+    This allows applications to choose whether negative scopes for enums should
+    include records with `nil` values or exclude them.
+
+    When set to `false` (8.2 default), negative scopes will include records with `nil`
+    values. When set to `true`, negative scopes will exclude records with `nil` values.
+
+    *QWYNG*
+
 *   Fix bug when `current_transaction.isolation` would not have been reset in test env.
 
     Additionally, extending the change in [#55549](https://github.com/rails/rails/pull/55549)
--- a/guides/source/upgrading_ruby_on_rails.md
+++ b/guides/source/upgrading_ruby_on_rails.md
@@ -104,7 +104,17 @@ Book.not_published # => [book2]
 Book.not_published # => [book2, book3]
 ```
 
-If you want to maintain the previous behavior (excluding `nil` values), you can set the following configuration:
+If you want to maintain the previous behavior (excluding `nil` values), you should manually override the negative scope in your model:
+
+```ruby
+class Book < ApplicationRecord
+  enum :status, [:proposed, :written, :published]
+  # Override to exclude nil values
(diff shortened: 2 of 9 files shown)


railties/test/application/configuration_test.rb
```diff
@@ -3138,6 +3138,7 @@ class Post < ActiveRecord::Base
       assert_match(/no such column: "?non_existent"?/, error.message)
     end
 
+
```
zzak:
```suggestion
```

guides/source/upgrading_ruby_on_rails.md
```diff
@@ -104,6 +104,14 @@ Book.not_published # => [book2]
...
 ```
 
+If you want to maintain the previous behavior (excluding `nil` values), you can set the following configuration:
+
+```ruby
+config.active_record.negative_scopes_for_enums_to_include_records_with_nil = false
+```
+
+
+
```
zzak, line 122:
```suggestion
```

activerecord/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
...
+
+    This allows applications to choose whether negative scopes for enums should
+    include records with `nil` values or exclude them.
+
+    When set to `true` (default), negative scopes will include records with `nil`
+    values. When set to `false`, negative scopes will exclude records with `nil` values.
+
+    *QWYNG*
+
+* Fix inconsistency in PostgreSQL handling of unbounded time range types
```
zzak:
```suggestion
*   Fix inconsistency in PostgreSQL handling of unbounded time range types
```

The heading needs to be indented, I think that will fix CI

zzak:
I think this makes sense, should we deprecate the flag too? Then the name should be `deprecated_...` or something :thinking:

QWYNG:
@zzak 
> I think this makes sense, should we deprecate the flag too? Then the name should be deprecated_... or something 🤔

Thank you for the review and good idea.

How about `deprecated_negative_enum_scopes_exclude_nil` with a default of `false`?
This makes it explicit for users who set it to `true` that they're using the deprecated behavior.

QWYNG:
Alternatively, `deprecated_negative_scopes_for_enums_to_exclude_records_with_nil` could work as well.

zzak:
I think making deprecated_whatever default to false in 8.2 and otherwise true for the old behavior makes sense, shorter name looks better IMO

QWYNG:
Switched to `deprecated_negative_enum_scopes_exclude_nil` and set default to `false` (in 8.2, `true` for old version).
Some tests are failing, but it's unclear if these are flaky tests or caused by the changes here.

zzak:
It's just a flake, I've rerun the failed job. Let's wait for more feedback on this, thanks for your work! :bow:

rafaelfranca (changes requested):
I'm ok with a temporary configuration to make easier for people to upgrade, but we need a plan for developers to migrate to the new behavior. We don't want this behavior to be configurable forever.

rafaelfranca:
Maybe a solution for that would to show a deprecation when that configuration is `true` and tell users to override the scope with the previous behavior if they want to keep the old behavior?

QWYNG:
@rafaelfranca 
Thank you for your suggestion. I have implemented a deprecation warning to be displayed when this deprecated setting is set to true.

PR #56115, opened 2025-11-08
How to fix InvalidAuthenticityToken with formmethod [ci skip]

### LINKED ISSUE #52290: ActionController::InvalidAuthenticityToken when submitting a form via button with `formmethod: :delete`
### Steps to reproduce
1. Create a new app with `rails _7.2.0.beta2_  new csrf_repro -d sqlite3`
2. Add scaffold with `rails g scaffold Products`
3. Replace `app/views/products/show.html.erb` with following:
  ```erb
  <%= form_with model: @product do |f| %>
    <%= f.button "Remove product", formmethod: :delete %>
  <% end %>
  ```

4. Disable JavaScript
5. Navigate to http://localhost:3000/products/new
6. Create new product
7. Click "Remove product"

Repo with the project repro: https://github.com/pokonski/csrf_repro

### Expected behavior
products#destroy action should be properly executed and product removed. 

The docs suggest this is possible (https://guides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark) b

### Actual behavior
`ActionController::InvalidAuthenticityToken` is raised instead when Hotwire is disabled (via Javascript disable or just removing it from application.js)

The difference seems to be in how `_method` param is submitted:

- with JS and Hotwire enabled there's only a single param:

image

- without JS/Hotwire the `_method` param appears twice with both the original method (`patch`) and `delete`

image

Issue is present in both Firefox 127 and Chrome 126

### System configuration
**Rails version**: 7.2.0-beta2

**Ruby version**: ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin23]

zzak:
Fixes #52290

Screenshot from 2025-11-08 19-40-25

/cc @p8

### DIFF
--- a/guides/source/form_helpers.md
+++ b/guides/source/form_helpers.md
@@ -462,9 +462,17 @@ Rails works around this issue by emulating other methods over POST through a com
 
 In this case, the "Update" button will be treated as `PATCH` and the "Delete" button will be treated as `DELETE`.
 
+NOTE: If you're not using [JavaScript][js-guide], the per-form CSRF validation will raise a `ActionController::InvalidAuthenticityToken` error when clicking the `formmethod` button.<br>
+To avoid this either: disable per-form validation globally with [`config.action_controller.per_form_csrf_tokens = false`][per-form-csrf-tokens],
+add a `<%= hidden_field_tag :authenticity_token, form_authenticity_token %>` in the form,
+or use `form_with(authenticity_token: form_authenticity_token)`.
+
 [formmethod]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod
 [button-name]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-name
 [button-value]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-value
+[js-guide]: https://guides.rubyonrails.org/working_with_javascript_in_rails.html
+[per-form-csrf-tokens]: https://guides.rubyonrails.org/configuring.html#config-action-controller-per-form-csrf-tokens
+
 
 Making Select Boxes with Ease
 -----------------------------


MatheusRich (approved):

guides/source/form_helpers.md
```diff
@@ -462,9 +462,17 @@ Rails works around this issue by emulating other methods over POST through a com
 
 In this case, the "Update" button will be treated as `PATCH` and the "Delete" button will be treated as `DELETE`.
 
+NOTE: If you're not using [JavaScript][js-guide], the per-form CSRF validation will raise a `ActionController::InvalidAuthenticityToken` error when clicking the `formmethod` button.<br>
+To avoid this either: disable per-form validation globally with [`config.action_controller.per_form_csrf_tokens = false`][per-form-csrf-tokens],
+add a `<%= hidden_field_tag :authenticity_token, form_authenticity_token %>` in the form,
+or use `form_with(authenticity_token: form_authenticity_token)`.
```
p8, line 468: Maybe we can expand a bit on why this fails?
The `authenticity_token` in a form is [bound to the form method](https://github.com/rails/rails/blob/main/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L721) if per-form CSRF validation is enabled. If Javascript is enabled it will use the global `authenticity_token` instead, which doesn't look at the form method.

Maybe disabling `per_form_csrf_tokens` is a bit drastic if this is used in just one form?
Also, adding the `hidden_field_tag` will add another hidden field with the new `form_authenticity_token`. Overriding the `authenticity_token` basically works the same but is a bit cleaner as it replaces the existing hidden field.
Maybe this should be recommended even if Javascript is enabled?

Maybe something like:
```suggestion
NOTE: The `authenticity_token` in a form is bound to the form method if per-form CSRF validation is enabled. Submitting a form with a different method will raise an `ActionController::InvalidAuthenticityToken` error. If [JavaScript][js-guide] is used to submit the form it will typically send the global `authenticity_token` instead, which isn't bound to the form.
<br>
To make sure the form works without JavaScript, pass the global `form_authenticity_token` to `form_with` to allow all methods to be submitted `form_with(authenticity_token: form_authenticity_token)`.
```

PR #56117, opened 2025-11-08
add deep compact methods to hash

italomatos:
## Motivation / Background

While Ruby's `Hash#compact` removes nil values from a hash, it only operates at the top level. When working with nested hashes (common in Rails applications, especially with params, JSON responses, and nested attributes), there's no built-in way to remove nil or blank values recursively.

This PR adds four new methods to Hash that recursively remove nil or blank values from nested structures.

## Detail

This PR introduces four new methods to `Hash`:

- **`deep_compact`**: Returns a new hash with all nil values removed recursively
- **`deep_compact!`**: Destructive version of `deep_compact`
- **`deep_compact_blank`**: Returns a new hash with all blank values removed recursively (using `Object#blank?`)
- **`deep_compact_blank!`**: Destructive version of `deep_compact_blank`

All methods handle nested hashes and arrays containing hashes.

## Examples

# deep_compact - removes nil values recursively
hash = { a: 1, b: nil, c: { d: 2, e: nil, f: { g: nil, h: 3 } } }
hash.deep_compact
# => { a: 1, c: { d: 2, f: { h: 3 } } }

# deep_compact_blank - removes blank values recursively
hash = { a: 1, b: "", c: nil, d: [], e: {}, f: false, g: { h: "", i: 2 } }
hash.deep_compact_blank
# => { a: 1, g: { i: 2 } }

# Works with arrays of hashes
hash = { a: [{ b: 1, c: nil }, { d: nil, e: 2 }] }
hash.deep_compact
# => { a: [{ b: 1 }, { e: 2 }] }## Use Cases

- Cleaning up API responses before sending to clients
- Removing blank nested attributes before validation
- Sanitizing form params with nested structures
- Data cleanup in JSON serialization

## Implementation Details

The implementation follows the same pattern as existing deep methods like `deep_transform_values` and `deep_transform_keys`:
- Uses private helper methods for recursion
- Handles both Hash and Array types
- Non-destructive versions create new objects
- Destructive versions (with `!`) modify in place

## Additional Information

⚠️ **Note**: According to Rails contributing guidelines, changes to Active Support core extensions are generally discouraged to avoid conflicts with future Ruby versions. This PR is being submitted for discussion and consideration by the core team. If accepted here, a similar proposal could be made to Ruby core, or this could be considered for Rails-specific use cases where the functionality is particularly valuable.

## Checklist

- [x] Implementation follows existing Rails patterns
- [x] Tests added for all new methods
- [x] Tests cover edge cases (nested hashes, arrays, empty values)
- [x] Both destructive and non-destructive variants implemented
- [x] Documentation includes clear examples
- [x] No linter errors

### DIFF
--- a/activesupport/lib/active_support/core_ext/hash.rb
+++ b/activesupport/lib/active_support/core_ext/hash.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "hash/conversions"
+require_relative "hash/deep_compact"
 require_relative "hash/deep_merge"
 require_relative "hash/deep_transform_values"
 require_relative "hash/except"
--- a/activesupport/lib/active_support/core_ext/hash/deep_compact.rb
+++ b/activesupport/lib/active_support/core_ext/hash/deep_compact.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+class Hash
+  # Returns a new hash with all nil values removed recursively.
+  # This includes the values from the root hash and from all
+  # nested hashes and arrays.
+  #
+  #   hash = { a: 1, b: nil, c: { d: 2, e: nil } }
+  #
+  #   hash.deep_compact
+  #   # => { a: 1, c: { d: 2 } }
+  def deep_compact
+    _deep_compact_in_object(self)
+  end
+
+  # Destructively removes all nil values recursively.
+  # This includes the values from the root hash and from all
+  # nested hashes and arrays.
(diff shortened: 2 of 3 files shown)


skipkayhil:
Thank you for the pull request, but this has been previously rejected:

https://github.com/rails/rails/pull/17608
https://github.com/rails/rails/pull/28588
https://github.com/rails/rails/pull/55390

PR #56118, opened 2025-11-08
Add support for PostgreSQL index `WITH (...)` options

### LINKED ISSUE #56116: Add support for PostgreSQL index `WITH (...)` options in `add_index` and `ActiveRecord::SchemaDumper`
### Steps to reproduce

```ruby
# frozen_string_literal: true
require "active_record"
require "minitest/autorun"

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "test_db")

  create_table :examples, force: true do |t|
    t.string :title
  end
end

class MissingWithOptionTest < Minitest::Test
  def setup
    ActiveRecord::Base.connection.execute("DROP INDEX IF EXISTS examples_title_idx;")
  end

  def test_add_index_with_with_option
    ActiveRecord::Base.connection.add_index(
      :examples,
      :title,
      name: "examples_title_idx",
      using: :btree,
      with: "fillfactor = 70"
    )
  end
end
```

### Expected behavior

Rails should:

- Accept the `with:` keyword in `add_index` and include it in the generated SQL, e.g.:

  ```sql
  CREATE INDEX examples_title_idx ON examples USING btree (title) WITH (fillfactor = 70);
  ```

- Persist the `with:` option through `SchemaDumper`, so that `schema.rb` round-trips correctly:

  ```ruby
  t.index ["title"], name: "examples_title_idx", using: :btree, with: "fillfactor = 70"
  ```

### Actual behavior

- `add_index` currently ignores the `with:` option and produces SQL without the `WITH (...)` clause.  
- `SchemaDumper` also lacks awareness of the `WITH (...)` parameters, so even if such an index exists, it dumps as:

  ```ruby
  t.index ["title"], name: "examples_title_idx", using: :btree
  ```

  Reloading the schema via `db:schema:load` attempts to recreate the index without the required `WITH (...)`, leading to PostgreSQL errors for index types that depend on it.

### Use case

PostgreSQL supports index storage parameters via `WITH (...)`, for example:

```sql
CREATE INDEX ... USING btree (...) WITH (fillfactor = 80);
CREATE INDEX ... USING gin (...) WITH (fastupdate = off);
CREATE INDEX ... USING bm25 (...) WITH (key_field = 'id', text_fields = '{...}');
```

There is no current Active Record DSL to represent these options. This prevents accurate schema dumping and restoration for any PostgreSQL index that requires `WITH (...)`.

### Proposed change

- Add a `with` attribute to `ActiveRecord::ConnectionAdapters::IndexDefinition`.
- Extend PostgreSQL adapter’s `#add_index` and `SchemaDumper` to support and round-trip the `with:` option.

### System configuration

**Rails version:** 8.1.1
**Ruby version:** 3.3.x
**Database:** PostgreSQL 17

dannote:
### Motivation / Background

PostgreSQL’s `WITH (...)` storage parameters are required by some extensions and index types. ParadeDB’s `bm25` search index, for example, must include a `WITH` clause that defines tokenizers, stemmers, and other schema details. Without DSL support, these indexes can’t be expressed in migrations, inspected via `index_exists?`, or round-tripped through schema dumps.

Closes #56116.

### Detail

- `IndexDefinition` now records a `with` hash for storage parameters.
- `add_index` / `index_exists?` accept `with:` hash and normalize it, ensuring migrations, checks, and schema dumps stay consistent.
- PostgreSQL adapter parses `WITH (...)` into the new `with` attribute and emits the clause when generating SQL.
- `SchemaDumper` includes `with:` in schema.rb, completing the round trip.

### Additional Information

- ParadeDB `bm25` example: https://docs.paradedb.com/v2/indexing/create-index#create-an-index
- PostgreSQL reference: https://www.postgresql.org/docs/current/sql-createindex.html

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Allow PostgreSQL index storage parameters to be managed via the `with:` option
+    on `add_index`, and include them in schema dumps and `index_exists?` lookups.
+
+    *Danila Poyarkov*
+
 *   Fix inconsistency in PostgreSQL handling of unbounded time range types
 
     Use `-infinity` rather than `NULL` for the lower value of PostgreSQL time
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_creation.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_creation.rb
@@ -117,6 +117,9 @@ module ActiveRecord
           sql << "(#{quoted_columns(index)})"
           sql << "INCLUDE (#{quoted_include_columns(index.include)})" if supports_index_include? && index.include
           sql << "NULLS NOT DISTINCT" if supports_nulls_not_distinct? && index.nulls_not_distinct
+          if clause = index_with_clause(index)
+            sql << clause
+          end
           sql << "WHERE #{index.where}" if supports_partial_index? && index.where
 
           sql.join(" ")
@@ -137,6 +140,8 @@ module ActiveRecord
         def supports_index_using?
           true
         end
+
+        def index_with_clause(index); end
 
         def add_table_options!(create_sql, o)
           create_sql << " #{o.options}" if o.options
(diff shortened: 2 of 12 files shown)


fatkodima:
I have the same opinion as before - https://github.com/rails/rails/issues/56116#issuecomment-3506710672 and think that this is not needed. This a postgres only feature, very rarely used in practice and if used, then most of the time it is used for things like fillfactor (and these are usually managed manually on production by DBAs). 

If you use some db engine specific feature, I suggest switching to structure.sql as soon as possible - there is no reason to stick to schema.rb. And these can be configured via custom `exec` queries in migration files.

PR #55690, opened 2025-09-17
Structured event subscribers

gmcgibbon:
### Motivation / Background

This Pull Request has been created to add structured events for framework notifications via the use of `ActiveSupport::StructuredEventSubscriber` subclasses. Events may be emitted with the `#emit_event` or `#emit_debug_event` methods like this, similar to a traditional `LogSubscriber`:

```ruby
class MyStructuredEventSubscriber < ActiveSupport::StructuredEventSubscriber
  def notification(event)
    emit_event("my.notification", data: event.payload)
  end
end

MyStructuredEventSubscriber.attach_to :namespace
```

### Detail

This Pull Request adds structured event subscribers for each framework library that has log subscribers for notifications. The idea is to mirror structured events for each existing framework log subscriber.

### Additional information

Eventually, we would like to unify notification subscribing and emit logs for these structured events (instead of maintaining separate log subscribers), and that will happen in a followup change. For now, this is focused on adding support for events and minimizing the set of changes as to not break anything.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56119, opened 2025-11-09
Emit structured events for Active Record work on v8.1

y-yagi:
### Motivation / Background

This PR has been created because structured events for Active Record doesn't emit on v8.1.

### Detail

This is because `active_record/structured_event_subscriber` isn't loaded(I guess just forgot to require it in #55690).

In `main` branch, this was already fixed by #55900. I think we should fix this on v8.1 too because we mentioned structured events for Active Record on CHANGELOG.
https://github.com/rails/rails/blob/8-1-stable/activerecord/CHANGELOG.md#rails-810-october-22-2025

### Checklist


activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Emit structured events for Active Record.
```
zzak: Thanks for your PR!

I think it doesn't need changelog because it's bug fix and should have been working.

We will need to backport this too. /cc @gmcgibbon

y-yagi: Yes, this is a bug fix. But, I added a CHANGELOG entry because this PR aims 8-1-stable, not main.

zzak: Ahh I missed the base.

There is already a changelog for this in 8-1-stable:
https://github.com/rails/rails/blob/d401e4b5477f157021872bd4fa43dcf722dc8b02/activerecord/CHANGELOG.md?plain=1#L72-L76

We should commit the bugfix to main and backport it to 8-1-stable branch:
https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#backporting

:bow:

y-yagi: I described in the description of PR, this bug was already fixed in `main` by #55900. `8-1-stable` is the only branch that we need to fix a bug.

zzak: Ugh, sorry I should read the description, I'm just used to looking at the diff. I still don't think the changelog is necessary though :pray: Sorry for the confusion!

skipkayhil (approved):

skipkayhil:
Nice catch, thanks!

PR #56123, opened 2025-11-09
counter_cache issue when using string IDs

### LINKED ISSUE #56122: Resetting counter_cache using string ids sets them to 0 incorrectly
### Steps to reproduce

When using integer for PK, following code will set count to 0 incorrectly. Code taken from `counter_cache_test.rb`
```ruby
Topic.reset_counters(@topic.to_param, "replies")
```

### Expected behavior

Set `replies_count` to correct number

### Actual behavior

Sets `replies_count` to 0

### System configuration

**Rails version**: 
Latest `main` branch

**Ruby version**: 
3.2.2 and 3.4.7

iuri-gg:
### Motivation / Background

This Pull Request has been created because of bug in counterculture implementation. It fixes https://github.com/rails/rails/issues/56122

### Detail

This Pull Request changes `counter_culture` implementation to pluck ids so the type of the passed in value matches to what we have in DB. 

### Checklist

### DIFF
--- a/activerecord/lib/active_record/counter_cache.rb
+++ b/activerecord/lib/active_record/counter_cache.rb
@@ -42,7 +42,7 @@ module ActiveRecord
             [id]
           end
         else
-          Array(id)
+          where(primary_key => id).pluck(primary_key)
         end
 
         updates = Hash.new { |h, k| h[k] = {} }
--- a/activerecord/test/cases/counter_cache_test.rb
+++ b/activerecord/test/cases/counter_cache_test.rb
@@ -127,6 +127,14 @@ class CounterCacheTest < ActiveRecord::TestCase
     end
   end
 
+  test "reset counters with to_param argument" do
+    Topic.increment_counter("replies_count", @topic.id)
+
+    assert_difference "@topic.reload.replies_count", -1 do
+      Topic.reset_counters(@topic.to_param, "replies")
+    end
+  end
+
   test "reset counters with modularized and camelized classnames" do
     special = SpecialTopic.create!(title: "Special")
     SpecialTopic.increment_counter(:replies_count, special.id)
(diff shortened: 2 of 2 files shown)


iuri-gg:
CI failure seems to be unrelated to my changes. there is a bug on main branch `--seed 29353` fails in `activerecord`. to reproduce use following command within the `activerecord` directory:
```
ARCONN=postgresql bundle exec ruby -w -Itest -Ilib -I../activesupport/lib -I../activemodel/lib test/cases/transactions_test.rb --seed 29353
```

activerecord/lib/active_record/counter_cache.rb
```diff
@@ -42,7 +42,7 @@ def reset_counters(id, *counters, touch: nil)
             [id]
           end
         else
-          Array(id)
+          where(primary_key => id).pluck(primary_key)
```
skipkayhil, line 45: I don't think it makes sense to make an additional query here. You end up making all callers pay the cost of the query instead of normalizing the value for the specific cases you need it

iuri-gg: Is there a good way to typecast the id especially if it is a composite key to avoid extra query?

skipkayhil: I'm not sure yet if I'm understanding your problem. `to_param` will always return a string, why are you using it here?

iuri-gg: The old codebase used to work and then broke. I think you are right that to_param shouldn't be used here, but either way, unexpectedly, it sets the wrong value. I would expect no change or correct count, not 0

byroot: > instead of normalizing the value for the specific cases you need it

Agreed, it's a general issue with AR class methods that do queries, they should access the primary key attribute from the model, and do the coercion.

PR #56128, opened 2025-11-11
Enhance rate limiting to support dynamic `to:` and `within:` options

muriloduarte:
### Motivation / Background

The `rate_limit` API already supports method names for `:by` and `:with`, bringing parity with callback declarations like `before_action`. Extending this to `:to` and `:within` allows applications to express dynamic limits (per-user, per-plan, per-tenant) without boilerplate conditionals in controllers. This improves flexibility while keeping existing static usages untouched. Related prior work: support for method names on `:by` and `:with` (see commit `63166fe`) [link](https://github.com/muriloduarte/rails/commit/63166fe6233527ddbe16f741a5a49d8b0854cdc3).

### Detail

Changes `ActionController::RateLimiting` so that `to:` and `within:` accept:
- Static values (unchanged behavior), e.g. `to: 10`, `within: 1.minute`.
- Instance method names (symbols), evaluated in the controller context via `send`.
- Callables (lambdas/procs), evaluated in the controller context via `instance_exec`.

Resolution logic:
- If value is a Symbol → `send(value)`
- Else if it responds to `call` → `instance_exec(&value)`
- Else → use as-is

### Additional information

- Aligns with earlier change to allow method names for `:by` and `:with` (commit `63166fe`) [link](https://github.com/muriloduarte/rails/commit/63166fe6233527ddbe16f741a5a49d8b0854cdc3).
- Negligible overhead (a couple of conditional checks) on the rate-limited path.
- Considered edge case: objects that respond to `call` but are intended as literals; current approach mirrors existing conventions in the codebase.

### Checklist

* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message.`

### DIFF
--- a/actionpack/lib/action_controller/metal/rate_limiting.rb
+++ b/actionpack/lib/action_controller/metal/rate_limiting.rb
@@ -10,8 +10,12 @@ module ActionController # :nodoc:
       # Applies a rate limit to all actions or those specified by the normal
       # `before_action` filters with `only:` and `except:`.
       #
-      # The maximum number of requests allowed is specified `to:` and constrained to
+      # The maximum number of requests allowed is specified by `to:` and constrained to
       # the window of time given by `within:`.
+      #
+      # Both `to:` and `within:` can be static values, callables,
+      # or method names (as symbols) that will be evaluated in the context of the
+      # controller processing the request.
       #
       # Rate limits are by default unique to the ip address making the request, but
       # you can provide your own identity function by passing a callable in the `by:`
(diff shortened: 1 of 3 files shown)


actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -10,9 +10,13 @@ module ClassMethods
       # Applies a rate limit to all actions or those specified by the normal
       # `before_action` filters with `only:` and `except:`.
       #
-      # The maximum number of requests allowed is specified `to:` and constrained to
+      # The maximum number of requests allowed is specified by `to:` and constrained to
       # the window of time given by `within:`.
       #
+      # Both `to:` and `within:` can be static values, callables (lambdas or procs),
```
MatheusRich:
```suggestion
      # Both `to:` and `within:` can be static values, callables,
```

MatheusRich (approved):

PR #56129, opened 2025-11-11
QueryIntent#execute!

matthewd:
Follow-up to #55897

QueryIntent now knows which adapter instance it belongs to (with the exception of the async queue, which needs a little extra handling, QueryIntent instances live and die all within a single `select_all`-or-similar method call on an adapter, so there's no lifetime issue).

That means QueryIntent can now ask the adapter to do work for it, and therefore is able to take the driver's seat for process steps where it was previously a passive container, like query preprocessing.

And all of the above sets us up to allow QueryIntent to `execute!`, storing the result for the caller to then request in whichever raw/cast form they prefer. 

Before:

mermaid-diagram-2025-10-29-175554

Now:

mermaid-diagram-2025-11-12-040028

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -119,6 +119,10 @@ module ActiveRecord
         internal_exec_query(sql, name, allow_retry:, materialize_transactions:).rows
       end
 
+      def query_command(sql, name = nil, allow_retry: false, materialize_transactions: true) # :nodoc:
+        affected_rows(internal_execute(sql, name, allow_retry: allow_retry, materialize_transactions: materialize_transactions))
+      end
+
       # Determines whether the SQL statement is a write query.
       def write_query?(sql)
         raise NotImplementedError
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -142,7 +142,7 @@ module ActiveRecord
         query_all(...).rows
       end
 
-      def query_all(sql, name = nil, allow_retry: true, materialize_transactions: true) # :nodoc:
+      def query_all(sql, name = "SCHEMA", allow_retry: true, materialize_transactions: false) # :nodoc:
         intent = internal_build_intent(sql, name, allow_retry:, materialize_transactions:)
         intent.execute!
         intent.cast_result
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -79,9 +79,28 @@ module ActiveRecord
           allow_retry: allow_retry
         )
 
-        select(intent, async: async && FutureResult::SelectAll)
-      rescue ::RangeError
-        ActiveRecord::Result.empty(async: async)
+        if async && async_enabled?
+          if current_transaction.joinable?
+            raise AsynchronousQueryInsideTransactionError, "Asynchronous queries are not allowed inside transactions"
+          end
+
+          future_result = FutureResult::SelectAll.new(pool, intent)
+          future_result.schedule!(ActiveRecord::Base.asynchronous_queries_session)
+          future_result
+        else
+          begin
+            intent.execute!
+            result = intent.cast_result
+          rescue ::RangeError
+            result = ActiveRecord::Result.empty
+          end
+
+          if async
+            FutureResult.wrap(result)
+          else
+            result
+          end
+        end
       end
 
       # Returns a record hash with the column names as keys and column values
@@ -120,7 +139,9 @@ module ActiveRecord
       end
 
       def query_command(sql, name = nil, allow_retry: false, materialize_transactions: true) # :nodoc:
-        affected_rows(internal_execute(sql, name, allow_retry: allow_retry, materialize_transactions: materialize_transactions))
+        intent = internal_build_intent(sql, name, allow_retry: allow_retry, materialize_transactions: materialize_transactions)
+        intent.execute!
+        intent.finish
       end
 
       # Determines whether the SQL statement is a write query.
@@ -143,7 +164,9 @@ module ActiveRecord
       # method may be manually memory managed. Consider using #exec_query
       # wrapper instead.
       def execute(sql, name = nil, allow_retry: false)
-        internal_execute(sql, name, allow_retry: allow_retry)
+        intent = internal_build_intent(sql, name, allow_retry: allow_retry)
+        intent.execute!
+        intent.raw_result
       end
 
       # Executes +sql+ statement in the context of this connection using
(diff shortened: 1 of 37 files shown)


activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -81,16 +71,36 @@ def cacheable_query(klass, arel) # :nodoc:
...
         intent = QueryIntent.new(
+          adapter: self,
           arel: arel,
           name: name,
           binds: binds,
           prepare: preparable,
           allow_retry: allow_retry
         )
 
-        select(intent, async: async && FutureResult::SelectAll)
```
matthewd, line 91: Inlined `select`: this is the only caller, so it seems clearer to keep the below work adjacent to the construction of the intent.

(It turns out that `FutureResult::SelectAll` is the only instance we ever use -- never plain `FutureResult`.)

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -631,49 +663,32 @@ def affected_rows(raw_result)
...
-        def internal_execute(sql, name = "SQL", binds = [], prepare: false, async: false, allow_retry: false, materialize_transactions: true, &block)
-          intent = QueryIntent.new(
+        def internal_build_intent(sql, name = "SQL", binds = [], prepare: false, allow_retry: false, materialize_transactions: true, &block)
+          QueryIntent.new(
+            adapter: self,
             raw_sql: sql,
             name: name,
             binds: binds,
             prepare: prepare,
-            async: async,
```
matthewd, line 661: This method / its predecessor is never `async`, and wouldn't actually run async even if it was.

activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -63,6 +53,73 @@ def to_h
...
+        nil
+      end
+
+      def finish
+        affected_rows # just to consume/close the result
+        nil
+      end
+
+      def cast_result
+        raise "Cannot call cast_result before query has executed" unless @executed
```
matthewd, line 107: There's huge potential for cleanup by making this (or more properly, `raw_result`) auto-`execute!` on demand. For this change I wanted to keep the callers a little more clearly imperative, though, to avoid going too far all at once.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -117,15 +127,31 @@ def select_rows(arel, name = nil, binds = [], async: false)
...
+        query_all(...).rows
+      end
+
+      def query_all(sql, name = "SCHEMA", allow_retry: true, materialize_transactions: false) # :nodoc:
+        intent = internal_build_intent(sql, name, allow_retry:, materialize_transactions:)
+        intent.execute!
+        intent.cast_result
+      end
+
+      def query_command(sql, name = nil, allow_retry: false, materialize_transactions: true) # :nodoc:
```
matthewd, line 151: The rest of the above are direct mirrors of `select_*` methods (but defaulting to assuming they're [retryable] schema queries); this one is closest to being the `query_*` family's equivalent of `exec_query`, based on how things actually get used, even though that method [unlike this] still returns a result. `query_command` is strictly "we care that this query has been executed [and hasn't failed/raised], but that's all".

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -49,16 +49,6 @@ def to_sql_and_binds(arel_or_sql_string, binds = [], preparable = nil, allow_ret
           [arel_or_sql_string, binds, preparable, allow_retry]
         end
       end
-      private :to_sql_and_binds
-
-      # Compiles Arel in the given QueryIntent if needed
-      def compile_arel_in_intent(intent) # :nodoc:
```
matthewd, line 55: I changed my mind about moving this (from https://github.com/rails/rails/pull/55897#discussion_r2493448527). I do now expect to involve an intent instance in `to_sql`, in a separate follow-up.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -579,42 +609,44 @@ def high_precision_current_timestamp
...
-      def internal_exec_query(...) # :nodoc:
-        cast_result(internal_execute(...))
+      def preprocess_query(sql) # :nodoc:
+        if write_query?(sql)
+          ensure_writes_are_allowed(sql)
+          mark_transaction_written
+        end
+
+        # We call tranformers after the write checks so we don't add extra parsing work.
+        # This means we assume no transformer whille change a read for a write
```
byroot:
```suggestion
        # This means we assume no transformer while change a read for a write
```

activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -63,6 +53,73 @@ def to_h
...
+        @raw_sql ||
+          begin
+            compile_arel!
+            @raw_sql
+          end
+      end
+
+      # Returns preprocessed SQL, memoized
+      def processed_sql
+        @processed_sql ||= adapter.preprocess_query(raw_sql)
```
byroot: Shouldn't this pass `self`? Because the QueryIntent will often know if it's a read or write hence save havign to use a much less reliable regexp.

matthewd: Done 👍🏻

https://github.com/rails/rails/pull/56129/commits/d632189604c3203623241b14abf16c5cdc4054ec

activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -63,6 +54,110 @@ def to_h
...
+          if write_query?
+            adapter.mark_transaction_written
+          end
+
+          sql = raw_sql
+
+          # We call tranformers after the write checks so we don't need to parse the
+          # transformed result (which probably just adds comments we'd need to ignore).
+          # This means we assume no transformer will change a read into a write.
+          ActiveRecord.query_transformers&.each do |transformer|
```
matthewd, line 145: I'm curious if there are any `query_transformers` other than `QueryLogs` out in the wild (despite it being a private API)... if we can constrain it to exclusively (prepending || appending) a comment, it might make more sense to keep that comment as a separate QueryIntent attribute instead of carrying the two `.._sql` values 🤔 

_Edit:_ I have now found some Shopify-internal usage, where a transformer is used to inject query hints -- so still comments, but in a specific mid-string position -- as well as some public users that seem to potentially have very specific comment formatting requirements. I'll continue pondering, but probably won't nuke transformers for now (while reserving the right to do so, as it is still private API).

byroot:
So one big reservation I got here, is that as we mentioned in https://github.com/rails/rails/pull/55897, I think one of the major use case for `QueryIntent` is for people like @beauraF to be able to direct arbitrary queries at a read replica etc.

But how would that work if the `adapter` is already selected when the intent is created?

matthewd:
Yeah, I'm not quite sure how to break that dependency: we need an adapter to know whether a query is readonly, let alone everything else happening here. 

... which ties in to the fact that we're already inside a chosen adapter by the time we construct a QueryIntent.

I think it'd be possible to (carefully) use a similar detach/reattach handoff as I have here in `schedule!`/`FutureResult`.

PR #56139, opened 2025-11-12
Fix AppGeneratorTest devcontainer MySQL tests when using socket locally

zzak:
Fixes #56134


yahonda:
Thanks for the pull request. I'm wondering if the `socket: /var/run/mysqld/mysqld.sock` is accessible from the Rails container. This socket is visible inside the MySQL/MariaDB container, but not sure it is exported.

zzak:
We'd have to bind mount the socket to share it, I think, that seems weird to me. I think socket is only intended to be used on the same host. I could be wrong though.

guilleiguaran:
@zzak @yahonda please check #56166

zzak:
@guilleiguaran Thanks for your patch! I think it's better to just not use socket if devcontainer was set. :+1:

PR #56146, opened 2025-11-13
Make Rails.error.unexpected always report

mgloc:
### Motivation / Background

I love `Rails.error.unexpected`. I treat it like an `assert`, sprinkling it throughout the codebase to catch any strange behavior in dev or prod. Being able to attach custom context is useful for debugging production issues.

In development, it is not modular at all: when an unexpected case happens, we get an exception and a stack trace. All the benefits of the context hash and custom dev error subscribers are lost.

A simple fix would be to have it call `report` before raising the error.

### Detail

This Pull Request changes `ActiveSupport::ErrorReporter#unexpected` to always calls `report` before raising the error.

### Additional information

- [Original discussion](https://discuss.rubyonrails.org/t/activesupport-error-reporter-make-unexpected-call-report-before-raising-the-error/89911)

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Make `ActiveSupport::ErrorReporter#unexpected` always report the error.
+
+    *Matteo Glock*
+
 *   Add a detailed failure summary to `ActiveSupport::ContinuousIntegration`.
 
     *Mike Dalessio*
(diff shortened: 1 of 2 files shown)


activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Make `ActiveSupport::ErrorReporter#unexpected` report error in debug mode.
```
zzak: It's not in debug mode, it's always, right?

zzak:
@andrewn617 Do you know if this would cause problems for you if unexpected always calls report? I expect maybe, if those subscribers are expecting to only be reporting in production (debug=false), but not sure.

PR #56148, opened 2025-11-13
Deduplicate dynamic regexp in journey

byroot:
In an app I'm working on, the routes end up generating many duplicated regexps:

```
 181.28 kB    4532  "\\A"
              4529  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:77
                 1  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:120

 181.28 kB    4532  "\\."
              4530  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:98

 181.24 kB    4531  "([^/.?]+)"
              4531  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:72

 181.20 kB    4530  "\\Z"
              4529  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:77

 181.16 kB    4529  "\\.([^/.?]+)"
              4529  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:81

 362.32 kB    4529  "(?:\\.([^/.?]+))?"
              4529  actionpack-8.0.4/lib/action_dispatch/journey/path/pattern.rb:94
```

Deduplicating them saves about 15-16MiB of memory:

```ruby
>> def mem(str, count) = (ObjectSpace.memsize_of(str) + ObjectSpace.memsize_of(Regexp.new(str))) * count
>> mem("(?:\\.([^/.?]+))?", 4529) + mem("\\.([^/.?]+)", 4529) + mem("\\Z", 4530) + mem("([^/.?]+)", 4531) + mem("\\.", 4532) + mem("\\A", 4532)
=> 15960677
```

### DIFF
--- a/actionpack/lib/action_dispatch/journey/path/pattern.rb
+++ b/actionpack/lib/action_dispatch/journey/path/pattern.rb
@@ -6,6 +6,14 @@ module ActionDispatch
   module Journey # :nodoc:
     module Path # :nodoc:
       class Pattern # :nodoc:
+        REGEXP_CACHE = {}
+
+        class << self
+          def dedup_regexp(regexp)
+            REGEXP_CACHE[regexp.source] ||= regexp
+          end
+        end
+
         attr_reader :ast, :names, :requirements, :anchored, :spec
 
         def initialize(ast, requirements, separators, anchored)
@@ -74,7 +82,7 @@ module ActionDispatch
           end
 
           def accept(node)
-            %r{\A#{visit node}\Z}
+            Pattern.dedup_regexp(%r{\A#{visit node}\Z})
           end
 
           def visit_CAT(node)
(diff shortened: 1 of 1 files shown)


rkh:
This is neat, I might steal it.

Is there any scenario in which this could lead to a memory leak, since it would prevent Regexp instances from being garbage collected? Or is the assumption that the number of generated regular expressions is finite?

byroot:
In this case they're both finite (declared in routes.rb), and already immortal.

If you have a case with non immortal regexps, you can dedup them safely with https://docs.ruby-lang.org/en/master/ObjectSpace/WeakKeyMap.html#method-i-getkey

rkh:
Yep, that's what I ended up doing, thanks. Reduced memory footprint by around 25% for the test case I'm using.

rania33840-netizen:
set

PR #56149, opened 2025-11-13
Add support for parameterized tests

thehale:
### Motivation / Background

This Pull Request has been created to add an additional way to make writing tests more expressive, particularly when making identical actions/assertions about varying inputs.

Consider some of the existing approaches to writing parameterized tests in the ActiveSupport test suite...

**Inlining** https://github.com/rails/rails/blob/217d6a4a8614ac5ac22a5f8f12cb66bd0f6d2de2/activesupport/test/core_ext/string_ext_test.rb#L76-L80

**Metaprogramming**  https://github.com/rails/rails/blob/217d6a4a8614ac5ac22a5f8f12cb66bd0f6d2de2/activesupport/test/inflector_test.rb#L127-L132

Personally, I prefer how the inline test keeps the test method declaration at the root level of the containing class, but here are two things I greatly value about the metaprogramming version that I don't think are possible with the inline version.

1. Every test case is always executed, even if one or more fail.
	
	By contrast, with inlining, if an early test case fails, none of the later cases are executed (because iteration aborts on the first assertion failure). If multiple test cases would have failed, only the first failure is shown, potentially obfuscating a problem.

2. Test failure messages are automatically more detailed/helpful

	With metaprogramming, each test case is literally a separate test. When one fails, the failure message includes both the assertion failure and the detailed test case in the test name. Running the tests in verbose mode clearly shows the variety of test cases being evaluated.

	By contrast, with inlining, there's one single test with a generic test name. When it fails, there's only the assertion failure which doesn't indicate the specific failing test case (unless an explicit message is added to the assertion).

What if we could have the best of both worlds?

### Detail

This Pull Request adds an `each: <iterable>` paremeter to the `ActiveSupport::TestCase#test` helper, which will generate a new test method per item in the iterable.

For example, this code
```ruby
test "greater than 0", each: [1, 2, 3] do |value|
  assert_operator value, :>, 0
end
```

Produces the following tests
```ruby
$ bin/rails test --verbose

...
test_greater_than_0 |1|
test_greater_than_0 |2|
test_greater_than_0 |3|
...
```

> [!NOTE]
> Notice how the parameterized test method name places the value within `|pipes|` to mirror how the value is passed to the test block.

Using this syntax, the previous examples from the ActiveSupport test suite now look like this:

```ruby
test "singularize", each: SingularToPlural do |singular, plural|
  assert_equal singular, plural.singularize
end
```

```ruby
test "titleize mixture to title case", each: MixtureToTitleCase do |before, titleized|
  assert_equal titleized, ActiveSupport::Inflector.titleize(before)
end
```

Multiple portions of the test suite have been updated to use these parameterized tests to improve readability. These updates are in distinct commits so they can be accepted/rejected independent of the core addition to `ActiveSupport::TestCase#test`

### Additional information

Testing tools in other languages/frameworks include this kind of feature under the name ["parameterized tests"](https://docs.pytest.org/en/stable/how-to/parametrize.html), ["subtests"](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests), or "table driven tests" ([JavaScript](https://jestjs.io/docs/api#testeachtablename-fn-timeout), [Go](https://go.dev/wiki/TableDrivenTests))

### Checklist

### DIFF
--- a/activesupport/lib/active_support/testing/declarative.rb
+++ b/activesupport/lib/active_support/testing/declarative.rb
@@ -10,18 +10,36 @@ module ActiveSupport
         #   test "verify something" do
         #     ...
         #   end
-        def test(name, &block)
+        #
+        # Also supports parameterized tests.
+        #
+        #   test "verify something", each: [ ... ] do |value|
+        #     ...
+        #   end
+        def test(name, each: [], &block)
           test_name = "test_#{name.gsub(/\s+/, '_')}".to_sym
-          defined = method_defined? test_name
-          raise "#{test_name} is already defined in #{self}" if defined
-          if block_given?
-            define_method(test_name, &block)
-          else
-            define_method(test_name) do
-              flunk "No implementation provided for #{name}"
+          test_with_name(test_name, each: each, &block)
+        end
+
+        private
+          def test_with_name(test_name, each: [], &block)
+            if each.any?
+              each.each do |value|
+                parameterized_test_name = "#{test_name} |#{value.inspect}|".to_sym
+                test_with_name(parameterized_test_name) { instance_exec(value, &block) }
+              end
+            else
+              defined = method_defined? test_name
+              raise "#{test_name} is already defined in #{self}" if defined
+              if block_given?
+                define_method(test_name, &block)
+              else
+                define_method(test_name) do
+                  flunk "No implementation provided for #{test_name}"
+                end
+              end
             end
           end
-        end
       end
     end
   end
--- a/activesupport/test/core_ext/object/blank_test.rb
+++ b/activesupport/test/core_ext/object/blank_test.rb
@@ -19,25 +19,32 @@ class BlankTest < ActiveSupport::TestCase
   BLANK = [ EmptyTrue.new, nil, false, "", "   ", "  \n\t  \r ", "　", "\u00a0", [], {} ]
   NOT   = [ EmptyFalse.new, Object.new, true, 0, 1, "a", [nil], { nil => 0 }, Time.now ]
 
-  def test_blank
-    BLANK.each { |v| assert_equal true, v.blank?,  "#{v.inspect} should be blank" }
-    NOT.each   { |v| assert_equal false, v.blank?, "#{v.inspect} should not be blank" }
+  test "blank", each: BLANK do |value|
+    assert_equal true, value.blank?
   end
 
-  def test_blank_with_bundled_string_encodings
-    Encoding.list.reject(&:dummy?).each do |encoding|
-      assert_predicate " ".encode(encoding), :blank?
-      assert_not_predicate "a".encode(encoding), :blank?
-    end
+  test "not blank", each: NOT do |value|
+    assert_equal false, value.blank?
   end
 
-  def test_present
-    BLANK.each { |v| assert_equal false, v.present?, "#{v.inspect} should not be present" }
-    NOT.each   { |v| assert_equal true, v.present?,  "#{v.inspect} should be present" }
+  test "blank with bundled string encodings", each: Encoding.list.reject(&:dummy?) do |encoding|
+    assert_predicate " ".encode(encoding), :blank?
+    assert_not_predicate "a".encode(encoding), :blank?
   end
 
-  def test_presence
-    BLANK.each { |v| assert_nil v.presence, "#{v.inspect}.presence should return nil" }
-    NOT.each   { |v| assert_equal v,   v.presence, "#{v.inspect}.presence should return self" }
+  test "present", each: NOT do |value|
+    assert_equal true, value.present?
+  end
+
+  test "not present", each: BLANK do |value|
+    assert_equal false, value.present?
+  end
+
+  test "presence is self", each: NOT do |value|
+    assert_equal value, value.presence
+  end
+
+  test "presence is nil", each: BLANK do |value|
+    assert_nil value.presence
   end
 end
--- a/activesupport/test/core_ext/object/with_test.rb
+++ b/activesupport/test/core_ext/object/with_test.rb
@@ -92,12 +92,7 @@ class WithTest < ActiveSupport::TestCase
     assert_equal "1", @object.with(public_attr: "1", &:public_attr)
   end
 
-  test "basic immediates don't respond to #with" do
-    assert_not_respond_to nil, :with
-    assert_not_respond_to true, :with
-    assert_not_respond_to false, :with
-    assert_not_respond_to 1, :with
-    assert_not_respond_to 1.0, :with
-    assert_not_respond_to :sym, :with
+  test "basic immediates don't respond to #with", each: [nil, true, false, 1, 1.0, :sym] do |value|
+    assert_not_respond_to value, :with
   end
 end
--- a/activesupport/test/core_ext/string_ext_test.rb
+++ b/activesupport/test/core_ext/string_ext_test.rb
@@ -56,16 +56,20 @@ class StringInflectionsTest < ActiveSupport::TestCase
     EOS
   end
 
-  def test_pluralize
-    SingularToPlural.each do |singular, plural|
-      assert_equal(plural, singular.pluralize)
-    end
+  test "pluralize", each: SingularToPlural do |singular, plural|
+    assert_equal(plural, singular.pluralize)
+  end
 
+  test "pluralize already pluralized word" do
     assert_equal("plurals", "plurals".pluralize)
+  end
 
-    assert_equal("blargles", "blargle".pluralize(0))
-    assert_equal("blargle", "blargle".pluralize(1))
-    assert_equal("blargles", "blargle".pluralize(2))
+  test "pluralize with count", each: [
+    ["blargle", 0, "blargles"],
+    ["blargle", 1, "blargle"],
+    ["blargle", 2, "blargles"]
+  ] do |singular, count, plural|
+    assert_equal plural, singular.pluralize(count)
   end
 
   test "pluralize with count = 1 still returns new string" do
@@ -73,22 +77,16 @@ class StringInflectionsTest < ActiveSupport::TestCase
     assert_not_same name.pluralize(1), name
   end
 
-  def test_singularize
-    SingularToPlural.each do |singular, plural|
-      assert_equal(singular, plural.singularize)
-    end
+  test "singularize", each: SingularToPlural do |singular, plural|
+    assert_equal(singular, plural.singularize)
   end
 
-  def test_titleize
-    MixtureToTitleCase.each do |before, titleized|
-      assert_equal(titleized, before.titleize)
-    end
+  test "titleize", each: MixtureToTitleCase do |before, titleized|
+    assert_equal(titleized, before.titleize)
   end
 
-  def test_titleize_with_keep_id_suffix
-    MixtureToTitleCaseWithKeepIdSuffix.each do |before, titleized|
-      assert_equal(titleized, before.titleize(keep_id_suffix: true))
-    end
+  test "titleize with keep_id_suffix", each: MixtureToTitleCaseWithKeepIdSuffix do |before, titleized|
+    assert_equal(titleized, before.titleize(keep_id_suffix: true))
   end
 
   def test_downcase_first
@@ -117,10 +115,8 @@ class StringInflectionsTest < ActiveSupport::TestCase
     assert_not_predicate "".upcase_first, :frozen?
   end
 
-  def test_camelize
-    CamelToUnderscore.each do |camel, underscore|
-      assert_equal(camel, underscore.camelize)
-    end
+  test "camelize", each: CamelToUnderscore do |camel, underscore|
+    assert_equal(camel, underscore.camelize)
   end
 
   def test_camelize_lower
@@ -138,25 +134,23 @@ class StringInflectionsTest < ActiveSupport::TestCase
     assert_equal("Invalid option, use either :upper or :lower.", e.message)
   end
 
-  def test_dasherize
-    UnderscoresToDashes.each do |underscored, dasherized|
-      assert_equal(dasherized, underscored.dasherize)
-    end
+  test "dasherize", each: UnderscoresToDashes do |underscored, dasherized|
+    assert_equal(dasherized, underscored.dasherize)
   end
 
-  def test_underscore
-    CamelToUnderscore.each do |camel, underscore|
-      assert_equal(underscore, camel.underscore)
-    end
-
-    assert_equal "html_tidy", "HTMLTidy".underscore
-    assert_equal "html_tidy_generator", "HTMLTidyGenerator".underscore
+  test "underscore", each: CamelToUnderscore do |camel, underscore|
+    assert_equal(underscore, camel.underscore)
   end
 
-  def test_underscore_to_lower_camel
-    UnderscoreToLowerCamel.each do |underscored, lower_camel|
-      assert_equal(lower_camel, underscored.camelize(:lower))
-    end
+  test "underscore acronyms", each: [
+    ["HTMLTidy", "html_tidy"],
+    ["HTMLTidyGenerator", "html_tidy_generator"]
+  ] do |acronym, underscored|
+    assert_equal underscored, acronym.underscore
+  end
+
+  test "underscore to lower camel", each: UnderscoreToLowerCamel do |underscored, lower_camel|
+    assert_equal(lower_camel, underscored.camelize(:lower))
   end
 
   def test_demodulize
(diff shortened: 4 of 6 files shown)


activesupport/test/core_ext/object/blank_test.rb
```diff
@@ -19,25 +19,32 @@ def empty?
   BLANK = [ EmptyTrue.new, nil, false, "", "   ", "  \n\t  \r ", "　", "\u00a0", [], {} ]
   NOT   = [ EmptyFalse.new, Object.new, true, 0, 1, "a", [nil], { nil => 0 }, Time.now ]
 
-  def test_blank
-    BLANK.each { |v| assert_equal true, v.blank?,  "#{v.inspect} should be blank" }
-    NOT.each   { |v| assert_equal false, v.blank?, "#{v.inspect} should not be blank" }
+  test "blank", each: BLANK do |value|
+    assert_equal true, value.blank?
```
thehale, line 23: Parameterized tests can obviate the need for assertion messages in some situations.

When one of these specific test cases fails, the default message makes the precise shortcoming obvious:

```text
Failure:
BlankTest#test_blank |""| [test/core_ext/object/blank_test.rb:23]:
Expected: true
  Actual: false
```

activesupport/test/core_ext/object/with_test.rb
```diff
@@ -92,12 +92,7 @@ def initialize
     assert_equal "1", @object.with(public_attr: "1", &:public_attr)
   end
 
-  test "basic immediates don't respond to #with" do
-    assert_not_respond_to nil, :with
-    assert_not_respond_to true, :with
-    assert_not_respond_to false, :with
-    assert_not_respond_to 1, :with
-    assert_not_respond_to 1.0, :with
-    assert_not_respond_to :sym, :with
```
thehale, line 101: This is another variant of the **inlining** approach to parameterized tests.

activesupport/test/core_ext/string_ext_test.rb
```diff
@@ -56,220 +56,192 @@ def test_strip_heredoc_on_a_regular_indented_heredoc_with_blank_lines
...
-    assert_equal("blargles", "blargle".pluralize(0))
-    assert_equal("blargle", "blargle".pluralize(1))
-    assert_equal("blargles", "blargle".pluralize(2))
+  test "pluralize with count", each: [
+    ["blargle", 0, "blargles"],
+    ["blargle", 1, "blargle"],
+    ["blargle", 2, "blargles"]
+  ] do |singular, count, plural|
+    assert_equal plural, singular.pluralize(count)
   end
```
thehale, line 73: Sometimes a test method contains a mix of discrete test cases, only some of which are conducive to parameterization.

In this kind of situation, it's helpful to manually create separate test methods for the unparameterizable test cases.

activesupport/test/inflector_test.rb
```diff
@@ -222,88 +202,83 @@ def test_acronyms
...
 
-    assert_equal("LegacyApi", ActiveSupport::Inflector.camelize("legacyapi"))
-    assert_equal("LegacyAPI", ActiveSupport::Inflector.camelize("legacy_api"))
-    assert_equal("SomeLegacyApi", ActiveSupport::Inflector.camelize("some_legacyapi"))
-    assert_equal("Nonlegacyapi", ActiveSupport::Inflector.camelize("nonlegacyapi"))
+    assert_equal "LegacyApi", ActiveSupport::Inflector.camelize("legacyapi")
+    assert_equal "LegacyAPI", ActiveSupport::Inflector.camelize("legacy_api")
+    assert_equal "SomeLegacyApi", ActiveSupport::Inflector.camelize("some_legacyapi")
+    assert_equal "Nonlegacyapi", ActiveSupport::Inflector.camelize("nonlegacyapi")
   end
```
thehale: Parameterized tests aren't always the right choice, especially if test cases share significant/expensive setup.

In this example, the inflector setup currently runs once, followed by several parameterizable test cases. If we parameterized this test directly the inflector setup would re-run for each test case.

Extracting a separate test class would allow separating the expensive setup code from the parameterized test cases, but for small tests scenarios like this, that can reasonably be considered excessive.

activesupport/test/inflector_test.rb
```diff
@@ -604,7 +558,7 @@ def test_clear_all_resets_camelize_and_underscore_regexes
...
 
   Irregularities.each do |singular, plural|
-    define_method("test_pluralize_of_irregularity_#{plural}_should_be_the_same") do
+    test "pluralize of irregularity #{plural} should be the same" do
```
thehale, line 572: Parameterized tests aren't always the right choice, particularly when the test author wants precise control over the test name.

If the default parameterized test name is undesirable, then metaprogramming is a great way to get the precise desired formatting.

Other testing frameworks support string substitution in parameterized test names (e.g. [Jest](https://jestjs.io/docs/api#1-testeachtablename-fn-timeout)). That's potentially worth considering in a future PR.

activesupport/test/inflector_test.rb
```diff
@@ -315,184 +290,163 @@ def test_deconstantize
...
+  test "clear uncountables" do
+    ActiveSupport::Inflector.inflections.clear :uncountables
+    assert ActiveSupport::Inflector.inflections.uncountables.empty?
   end
 
-  def test_clear_acronyms_resets_to_reusable_state
+  test "clear humans" do
+    ActiveSupport::Inflector.inflections.clear :humans
+    assert ActiveSupport::Inflector.inflections.humans.empty?
+  end
```
thehale, line 458: Parameterized tests aren't always the right choice, especially when the test implementation doesn't reflect actual usage.

In an earlier draft, I had parameterized this test like so:

```ruby
test "clear", [ :plurals, :singulars, :uncountables, :humans ] do | inflection_type |
  ActiveSupport::Inflector.inflections.clear inflection_type
  assert ActiveSupport::Inflector.inflections.public_send(inflection_type).empty?
end
```

But neither the original implementation nor my `public_send` variant clearly reflect how actual users will write their code. I believe the clarity of the test code is worth the visual repetition.

activesupport/test/core_ext/string_ext_test.rb
```diff
@@ -341,7 +313,7 @@ def test_truncate_bytes
     end
   end
```
thehale: Parameterization isn't always the right choice. Even when it's viable, it can overly obfuscate the intent of a test -- particularly when lots of values are parameterized.

This test method is one such example. If parameterized it could look something like this:

```ruby
test "truncate bytes", each: [
  [ 16, nil, "👍👍👍👍"],
  [ 16, " ", "👍👍👍👍"],
  ... # over a dozen more test cases
] do | count, omission, result|
  assert_equal result, "👍👍👍👍".truncate_bytes(count, omission: omission)
end
```

To me, the original test is abundantly clear about the relationship between the input value, its transformation, and expected output. The parameterized version makes those relationships significantly harder to see.

Furthermore, the test cases without an omission (e.g. `.truncate_bytes(count)`) would have to be extracted to a separate test method, further obfuscating the comparisons between the default behavior and the `omission` behavior.

thehale:
One more example...

Parameterized tests also aren't the right choice when `value.inspect` obfuscates the test case.

Consider this existing ActiveSupport test

https://github.com/rails/rails/blob/e0789cb3b1a4ac96f8ecfcbca50031a7519ce5dd/activesupport/test/core_ext/numeric_ext_test.rb#L27-L36

It could be parameterized like this

```ruby
  test "irregular durations", each: {
    3000.days.since(NOW) => NOW.advance(days: 3000),
    1.month.since(NOW) => NOW.advance(months: 1),
    1.month.until(NOW) => NOW.advance(months: -1),
    20.years.since(NOW) => NOW.advance(years: 20),
    3000.days.since(DTNOW) => DTNOW.advance(days: 3000),
    1.month.since(DTNOW) => DTNOW.advance(months: 1),
    1.month.until(DTNOW) => DTNOW.advance(months: -1),
    20.years.since(DTNOW) => DTNOW.advance(years: 21)
  } do |actual, expected|
    assert_equal expected, actual
  end
```

But the full test names have no clear connection to the test case _and_ they change over time. Additionally, since each test case shares the same line for their `assert_equal` call, if a test case fails, it's very difficult to know which one it was.

```text
NumericExtTimeAndDateTimeTest#test_irregular_durations = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[Mon, 10 Feb 2025 15:30:45 +0000, Mon, 10 Feb 2025 15:30:45 +0000]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[Mon, 29 Apr 2013 15:30:45 +0000, Mon, 29 Apr 2013 15:30:45 +0000]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[2005-01-10 15:30:45 -0700, 2005-01-10 15:30:45 -0700]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[2025-02-10 15:30:45 -0700, 2025-02-10 15:30:45 -0700]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[2013-04-29 15:30:45 -0700, 2013-04-29 15:30:45 -0700]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[2005-03-10 15:30:45 -0700, 2005-03-10 15:30:45 -0700]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[Thu, 10 Mar 2005 15:30:45 +0000, Thu, 10 Mar 2005 15:30:45 +0000]| = 0.00 s = .
NumericExtTimeAndDateTimeTest#test_irregular_durations |[Mon, 10 Jan 2005 15:30:45 +0000, Mon, 10 Jan 2005 15:30:45 +0000]| = 0.00 s = .
```

activesupport/lib/active_support/testing/declarative.rb
```diff
@@ -10,18 +10,38 @@ module Declarative
...
-            define_method(test_name) do
-              flunk "No implementation provided for #{name}"
+        #
+        # Also supports parameterized tests.
+        #
+        #   test "verify something", each: [ ... ] do |value|
+        #     ...
+        #   end
+        def test(name, each: [], &block)
+          test_name = "test_#{name.gsub(/\s+/, '_')}"
```
seanpdoyle: Is it significant that the `to_sym` is no longer invoked?
```suggestion
          test_name = "test_#{name.gsub(/\s+/, '_')}".to_sym
```

activesupport/lib/active_support/testing/declarative.rb
```diff
@@ -10,18 +10,38 @@ module Declarative
...
+
+          if each.any?
+            each.each do |value|
+              parameterized_test_name = "#{test_name} |#{value.inspect}|"
+              define_test_method(parameterized_test_name) { instance_exec(value, &block) }
```
seanpdoyle: Could this be defined recursively in terms of `test` targeting this conditional's `else` branch?

```suggestion
              test(parameterized_test_name) { instance_exec(value, &block) }
```

activesupport/lib/active_support/testing/declarative.rb
```diff
@@ -10,18 +10,38 @@ module Declarative
...
             end
+          else
+            define_test_method(test_name, &block)
           end
         end
+
+        private
+          def define_test_method(test_name, &block)
+            test_name = test_name.to_sym
```
seanpdoyle: If the original `to_sym` is restored and the `define_test_method` call within the `each { … }` block is replaced by a call to `test` itself, could this coercion be omitted?

```suggestion
```

activesupport/test/inflector_test.rb
```diff
@@ -222,88 +202,83 @@ def test_acronyms
...
+      assert_not_predicate ActiveSupport::Inflector.underscore(under), :frozen?
+      assert_not_predicate ActiveSupport::Inflector.underscore(camel), :frozen?
+
+      assert_equal title, ActiveSupport::Inflector.titleize(under)
+      assert_equal title, ActiveSupport::Inflector.titleize(camel)
+      assert_not_predicate ActiveSupport::Inflector.titleize(under), :frozen?
+      assert_not_predicate ActiveSupport::Inflector.titleize(camel), :frozen?
+
+      assert_equal human, ActiveSupport::Inflector.humanize(under)
+      assert_not_predicate ActiveSupport::Inflector.humanize(camel), :frozen?
```
seanpdoyle: Is it significant that these wrapping parentheses are removed? If they were restored, this file's diff could be reduced dramatically.

activesupport/lib/active_support/testing/declarative.rb
```diff
@@ -10,18 +10,38 @@ module Declarative
...
+            defined = method_defined? test_name
+            raise "#{test_name} is already defined in #{self}" if defined
+            if block_given?
+              define_method(test_name, &block)
+            else
+              define_method(test_name) do
+                flunk "No implementation provided for #{name}"
+              end
+            end
+          end
```
seanpdoyle, line 42: Re-considering my prior feedback, I wonder if this `private` method be inlined back into the `else` branch of the conditional to further limit the scope of the implementation changes being proposed?

```suggestion
        def test(name, each: [], &block)
          test_name = "test_#{name.gsub(/\s+/, '_')}".to_sym

          if each.present?
            each.each do |value|
              test("#{test_name} |#{value.inspect}|") { instance_exec(value, &block) }
            end
          else
            defined = method_defined? test_name
            raise "#{test_name} is already defined in #{self}" if defined
            if block_given?
              define_method(test_name, &block)
            else
              define_method(test_name) do
                flunk "No implementation provided for #{name}"
              end
            end
          end
        end
```

activesupport/lib/active_support/testing/declarative.rb
```diff
@@ -10,18 +10,38 @@ module Declarative
...
-          else
-            define_method(test_name) do
-              flunk "No implementation provided for #{name}"
+        #
+        # Also supports parameterized tests.
+        #
+        #   test "verify something", each: [ ... ] do |value|
+        #     ...
+        #   end
+        def test(name, each: [], &block)
```
MatheusRich, line 19: `each` doesn't sound correct to me. Maybe `values` or `params`/`parameters` since this is about test parameters?

thehale: I did consider names like `values`/`params`, but I like `each` better for a few reasons:
  - I like the sound of "testing `each` scenario" from a collection or "asserting `test "something"` holds true for `each` case".
  - `each` mirrors the standard Ruby way to iterate over a collection (e.g. `Hash.each` and `Array.each`).
    ```ruby
    test "something", each: [ ... ] { |value| ... }
    # mirrors
    [ ... ].each { |value| ... }
    ```
  - I think using `each` clearly indicates that an iteration will occur to run multiple separate test cases, whereas `values/parameters` could be ambiguous (are they passed to the block one at a time, or all at once)?

Your thoughts? Feel free to critique these perspectives. Dissenting voices welcome!

activesupport/lib/active_support/testing/declarative.rb
thehale: As you already noted, it was moved to `define_test_method` to ensure we are properly symbolizing both normal and parameterized test method names prior to checking their existence.

activesupport/lib/active_support/testing/declarative.rb
seanpdoyle: I've experimented with this locally, and was able to test out the suggestion I'd made in https://github.com/rails/rails/pull/56149/files#r2525222343:

```sh
❯ git diff main -w activesupport/lib/active_support/testing/declarative.rb
```
```diff
diff --git a/activesupport/lib/active_support/testing/declarative.rb b/activesupport/lib/active_support/testing/declarative.rb
index 7c3403684d..3c3cb1cd2e 100644
--- a/activesupport/lib/active_support/testing/declarative.rb
+++ b/activesupport/lib/active_support/testing/declarative.rb
@@ -10,8 +10,20 @@ module Declarative
         #   test "verify something" do
         #     ...
         #   end
-        def test(name, &block)
+        #
+        # Also supports parameterized tests.
+        #
+        #   test "verify something", each: [ ... ] do |value|
+        #     ...
+        #   end
+        def test(name, each: [], &block)
           test_name = "test_#{name.gsub(/\s+/, '_')}".to_sym
+
+          if each.any?
+            each.each do |value|
+              test("#{test_name} |#{value.inspect}|") { instance_exec(value, &block) }
+            end
+          else
             defined = method_defined? test_name
             raise "#{test_name} is already defined in #{self}" if defined
             if block_given?
@@ -26,3 +38,4 @@ def test(name, &block)
       end
     end
   end
+end
```

That shrinks the implementation-side diff dramatically. Had you considered relying on recursion?

activesupport/lib/active_support/testing/declarative.rb
thehale: I really like this idea -- and started implementing it -- but the [first line normalizing the test name](https://github.com/thehale/rails/blob/93494d4c62d2f8f3f2de5a545fa82ba1e5e0b131/activesupport/lib/active_support/testing/declarative.rb#L20) creates two roadblocks that I think justify keeping the current approach.

1. We would lose the whitespace in any parameters. Aside from altering the actual values in the test case, this can cause test instantiation failures.

	```ruby
	test "something about whitespace", each: [ " ", "   "] { ... }
	
	# would become
	
	test_test_something_about_whitespace_|"_"|
	test_test_something_about_whitespace_|"_"|  # error: method already defined.
	```
	
	This isn't hypothetical. One of the [existing string inflection tests](https://github.com/thehale/rails/blob/93494d4c62d2f8f3f2de5a545fa82ba1e5e0b131/activesupport/test/inflector_test_cases.rb#L261-L265) uses `"employee_id" => "Employee id"` and `"employee id" => "Employee id"` which the current test name regex would normalize to identical test names.
	
	We could workaround the issue by adjusting the regex to omit anything between the `|pipes|`, but to me that seems too "special-casey".

2. We'd also need to slice off the `test_` prefix when forming the `parameterized_test_name`. Otherwise, the final test name would have the prefix `test_test_`.

	```diff
	-parameterized_test_name = "#{test_name} |#{value.inspect}|"
	+parameterized_test_name = "#{test_name.slice('test_'.size..)} |#{value.inspect}|"
	```

   Definitely doable, but I think this slightly muddies the implementation, potentially negating the visual benefit of the recursive approach.

activesupport/lib/active_support/testing/declarative.rb
thehale: Yes, but there are some roadblocks to pursuing the recursive approach. Let's make my [above comment](https://github.com/rails/rails/pull/56149/files#r2528453532) the place for that discussion.

activesupport/lib/active_support/testing/declarative.rb
thehale: > Had you considered relying on recursion?

Not until you suggested it. This was the comment that got me most excited about trying that approach. Unfortunately, the recursive approach introduces some other concerns. Let's make my [above comment](https://github.com/rails/rails/pull/56149/files#r2528453532) the place for that discussion.

activesupport/test/inflector_test.rb
thehale: Not significant. I was following the principle of ["clean the campsite"](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) when visiting a file. You'll notice I also standardized the tests on the declarative format.

Not opposed to removing these types of changes so the diff only modifies the tests that now use the new `each: <iterable>` format.

activesupport/lib/active_support/testing/declarative.rb
thehale: After some more experimentation, we can make a recursive implementation work by removing the `test_name` normalization from the recursive step.

```diff
diff --git a/activesupport/lib/active_support/testing/declarative.rb b/activesupport/lib/active_support/testing/declarative.rb
index 7c3403684d..ad22339543 100644
--- a/activesupport/lib/active_support/testing/declarative.rb
+++ b/activesupport/lib/active_support/testing/declarative.rb
@@ -10,15 +10,33 @@ module Declarative
         #   test "verify something" do
         #     ...
         #   end
-        def test(name, &block)
+        #
+        # Also supports parameterized tests.
+        #
+        #   test "verify something", each: [ ... ] do |value|
+        #     ...
+        #   end
+        def test(name, each: [], &block)
           test_name = "test_#{name.gsub(/\s+/, '_')}".to_sym
+          create_test(test_name, each: each, &block)
+        end
+
+        private
+          def create_test(test_name, each: [], &block)
+            if each.any?
+              each.each do |value|
+                parameterized_test_name = "#{test_name} |#{value.inspect}|".to_sym
+                create_test(parameterized_test_name) { instance_exec(value, &block) }
+              end
+            else
               defined = method_defined? test_name
               raise "#{test_name} is already defined in #{self}" if defined
               if block_given?
                 define_method(test_name, &block)
               else
                 define_method(test_name) do
-              flunk "No implementation provided for #{name}"
+                  flunk "No implementation provided for #{test_name}"
+                end
               end
             end
           end
```

This still changes most every line, though many changes are now whitespace-only.

Thoughts?

thehale: Added this in the latest rebase.

activesupport/test/inflector_test.rb
thehale: Removed changes to tests that are not parameterized in some way in the latest rebase.

activesupport/test/core_ext/string_ext_test.rb
```diff
@@ -56,39 +56,37 @@ def test_strip_heredoc_on_a_regular_indented_heredoc_with_blank_lines
     EOS
   end
 
-  def test_pluralize
-    SingularToPlural.each do |singular, plural|
-      assert_equal(plural, singular.pluralize)
-    end
+  test "pluralize", each: SingularToPlural do |singular, plural|
+    assert_equal plural, singular.pluralize
```
seanpdoyle: I do think that retaining spacing and parenthesis usage will help to shrink this diff to its most important parts, and will facilitate review from Core or a Committer:

```suggestion
    assert_equal(plural, singular.pluralize)
```

thehale: Done

zzak:
Couldn't this be a gem instead?

jeremy (commented):
This is an excellent PR, implementation, and refinement, but I don't think the abstraction is a fit. 

Would contrast the opening code sample with: 
```ruby
   MixtureToTitleCase.each do |string, expected|
     test "titleize #{string}" do
       assert_equal expected, ActiveSupport::Inflector.titleize(string)
     end 
   end
```

"I have these cases. Each becomes a test." The expression, sequencing, and result all match the mental model. The code is digestible at a glance; its meaning is immediately understood.

To trade away fluent Ruby idioms we must be achieving some compelling conceptual compression, some manyfold payback. I'm not feeling that here. We're reorganizing syntax, transposing, not clarifying (and, realistically, obfuscating).

There is a legit issue with assertions failing mid-loop and only seeing the first failure, often with an insufficient message. Would address that underlying issue: improve assertion failure messages; hoist in-test iterations out of them, to test declaration; introduce a test/unit-style `assert_all` that collects failures across multiple examples.

PR #56152, opened 2025-11-13
Implement LocalCache strategy on MemoryStore

mikeygough:
### Motivation / Background

This Pull Request has been created because the `ActiveSupport::Cache::MemoryStore` currently doesn't implement the local cache strategy. It should respond to the same interface as other cache stores (e.g. `ActiveSupport::Cache::NullStore`). While the backing cache in this case is just another memory cache, I think we can guarantee the same behaviour the easiest way using the local cache strategy module.

### Detail

This Pull Request simply prepends `Strategy::LocalCache` to `MemoryStore` which adds common behavior for cache stores. I've added two tests to accompany these changes.

### Additional information

N.A.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Implement LocalCache strategy on `ActiveSupport::Cache::MemoryStore`. The memory store
+    needs to respond to the same interface as other cache stores (e.g. `ActiveSupport::NullStore`).
+
+    *Mikey Gough*
+
 *   Add a detailed failure summary to `ActiveSupport::ContinuousIntegration`.
 
     *Mike Dalessio*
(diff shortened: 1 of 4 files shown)


gmcgibbon:
It looks like `test/commands/middleware_test.rb:317` is failing, can you take a look? It might be expecting the memory store to not implement the local cache strategy.

mikeygough:
> It looks like `test/commands/middleware_test.rb:317` is failing, can you take a look? It might be expecting the memory store to not implement the local cache strategy.

@gmcgibbon I think that's right... I've updated the test to set `config.cache_store = :file_store` which doesn't get added to middleware.

gmcgibbon (approved):
Thanks! The file store might also want to respond to local cache methods at some point, in which case we can probably get rid of that test.

Issue #48398, opened 2023-06-05
Should Active Record disable prepared statements if QueryLogs are enabled?


casperisfine:
Ref: https://github.com/rails/rails/pull/42240

`ActiveRecord::QueryLogs` was a direct port of Marginalia, and it just hit me that Marginalia was very purposely disabling prepared statements: https://github.com/basecamp/marginalia/blob/c02d3d8404b15004c3432783543b8fc04306927b/lib/marginalia.rb#L67

I think `AR::QueryLogs` should do the same to avoid generating tons of prepared statements, as with query logs each query has a high chance to be unique, making prepared statements a useless overhead.

cc @keeran @eileencodes

exterm:
I believe that since https://github.com/rails/rails/issues/41896, prepared statements are used as a way to filter sensitive data from SQL logs (not just for performance). Enabling `ActiveRecord::QueryLogs` now sanitizing of SQL logs (by disabling prepared statements). This was surprising to me.

By default, we seem to be getting the `application` key (should be constant) and either `job` or `index` and `controller`.

I'm thinking the cardinality should be OK?

exterm:
Would it make sense for Rails to use explicit binds (enabling #41896) even when prepared statements are disabled?

byroot:
Can you reproduce this with a script? And if so open an issue. Because I don't think that is the case. Prepared statements or not shouldn't change anything about logging.

exterm:
I may be misunderstanding but I believe that https://github.com/rails/rails/pull/42006/files filters the binds, and without prepared statements there don't seem to be binds to filter. My understanding (and observation) is that without prepared statements, the values will be in the sql string itself.

Reproduce with a script? 🤔 I think one could set up a minimal Rails app to reproduce this, yes. Just enable `ActiveRecord::QueryLogs`, create a table with an `email` column, and look at the logs for queries for it. I don't really have the time right now but I'll try to get to it

byroot:
There are binds regardless of whether prepared statements are enabled or not. At least up until the logging point.

If you are experencing a problem please open an issue with repro steps.

byroot:
See https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case for how to provide a reproduction.

PR #56153, opened 2025-11-13
Do not disable prepared statements when query log tags are enabled

aliismayilov:
### Motivation / Background

The earliest discussion I found on this topic is from here: https://github.com/rails/rails/issues/48398

It is true that SQL comments are stored in the prepared statements and adding a tag with high cardinality would generate an extra statement per tag. But I'd like to make a case for keeping the prepared statements even when query log tags are enabled.

Default query log tags are `[:application, :job, :controller, :action]`. There's a high chance that any non-basic query has only one or few combination of the these tags. Which means that the prepared statements will be mostly the same and useful.

It's much trickier to reenable the prepared statements, given the rails's initialization process. I might be missing something, but I end up every time having to write the following code to reenable the prepared statements which feels like fighting against the framework:
```ruby
config.active_record.query_log_tags_enabled = true
initializer "reenable_prepared_statements", after: "active_record.query_log_tags_config" do
  config.after_initialize do
    ActiveRecord.disable_prepared_statements = false
  end
end
```

On the other hand, disabling prepared statments can easily be done in database.yml file.

Prepared statements is a good default for for saving time parsing queries and filtering sensitive information when SQL logging is enabled.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Do not disable prepared statements by default when query log tags are enabled.
+
+    *Ali Ismayilov*
+
 *   On MySQL parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
 
     Truncating on MySQL is very slow even on empty or nearly empty tables.
--- a/activerecord/lib/active_record/query_logs.rb
+++ b/activerecord/lib/active_record/query_logs.rb
@@ -63,6 +63,9 @@ module ActiveRecord
   #       },
   #     ]
   #
+  # WARNING: Avoid adding high cardinality tags (like request ID), since this will cause each prepared statement to be unique.
+  # Consider disabling prepared_statements in config/database.yml.
+  #
   # By default the name of the application, the name and action of the controller, or the name of the job are logged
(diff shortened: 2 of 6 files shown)


zzak:
You can still enable them yourself of you want, right? Don't see the need for a new default for this. 🤔

aliismayilov:
> You can still enable them yourself of you want, right? Don't see the need for a new default for this. 🤔

Right, but given the default set of tags, I thought, it's a better to not disable prepared statements, since it's a better default.

zzak:
But that ship may have already sailed, Rails doesn't normally revert defaults unless there is a good reason and you're the first person to report this since it was merged.

Not that your concern is invalid, but it may take more evidence to persuade the team to go back on this.

Thank you for your patch!

byroot:
It's just too much of a footgun, and as @zzak said, this is just a safe default, you can perfectly have both enabled if you know what you are doing.

aliismayilov:
I understand, thanks for taking time to look into it @zzak @byroot.

Is it worth spending time to make a PR to the guides on how to reenable the prepared statements? Currently, one needs to dig through the source code to come up with the necessary code.

byroot:
Sure. Please open a documentation PR

PR #56159, opened 2025-11-14
Restore missing Postgres type decoding

matthewd:
We've had support for typecasting `money` and `bytea` on the purely-internal `query` method forever: https://github.com/rails/rails/blob/077c3ad60db4c43cccb8f4637b53b8d8eb7a3c19/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L1167-L1168

... but `select_value` and friends (up to `exec_query` -- all the public query-running API methods) lost this behaviour in a92af3fbf0 (shipped in 4.0).

For model-layer behaviour, this does not create any change: the value will be decoded slightly earlier as it's loaded from the query, but model attribute typecasting is already handling these values, so the user-visible effect is only in situations that attribute typecasting doesn't apply (direct queries; `pluck` with non-attribute SQL expressions).

As a safety measure, this PR adds a monkey-patch to `PG::Connection.unescape_bytea` so that existing code of the form `unescape_bytea(select_value("SELECT bytea_value"))` will not double-unescape even when the config is enabled and `select_value` starts correctly returning the already-decoded binary value.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -1176,7 +1176,7 @@ module ActiveRecord
           }
           coders_by_name["date"] = PG::TextDecoder::Date if decode_dates
           coders_by_name["money"] = MoneyDecoder if decode_money
-          coders_by_name["bytea"] = PG::TextDecoder::Bytea if decode_bytea
+          coders_by_name["bytea"] = decode_bytea ? ByteaDecoder : ByteaMarker
 
           known_coder_types = coders_by_name.keys.map { |n| quote(n) }
           query = <<~SQL % known_coder_types.join(", ")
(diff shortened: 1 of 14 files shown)


activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -1205,7 +1242,30 @@ def decode(value, tuple = nil, field = nil)
...
+            if string.instance_variable_get(:@ar_pg_bytea_decoded) == true
+              ActiveRecord.deprecator.warn(<<~MSG.squish)
+                unescape_bytea called on a query result value that has already been
+                automatically unescaped due to `config.active_record.postgresql_adapter_decode_bytea`.
+                In a future Rails release, this will double-unescape the value as
+                instructed, which could cause data corruption.
+              MSG
+
+              result = string.dup
+              result.remove_instance_variable(:@ar_pg_bytea_decoded)
```
byroot, line 1268: Why remove the ivar?

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -1185,6 +1205,23 @@ def decode(value, tuple = nil, field = nil)
...
+            value.instance_variable_set(:@ar_pg_bytea_decoded, false) if value.encoding == Encoding::BINARY
+            value
+          end
+        end
+
+        class ByteaDecoder < PG::SimpleDecoder # :nodoc:
+          def decode(value, tuple = nil, field = nil)
+            return if value.nil?
+            decoded = PG::Connection.unescape_bytea(value)
+            decoded.instance_variable_set(:@ar_pg_bytea_decoded, true)
```
byroot, line 1231: Have you considered using a string subclass rather than an ivar? Ivars on primitives aren't great.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
matthewd: If someone is currently doing `unescape(unescape(select_value(..))`, I only want to elide the inner one [because it's already happened]

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
matthewd: I wanted to minimise the potential for observable change, especially as it's only a transitional mechanism. Given we're talking about blobs, I don't think object-local non-great-ness is a worry -- does it have a wider-reaching deopt effect on all strings or something?

byroot: No the deopt is local. That particular string end up having an external ivar entry, so it's a bit slower to sweep.

matthewd: Okay, so yeah given they're expected to be few and bulky, I think this is probably fine, for the benefit of it not being user-distinguishable (so we can remove it again post-transition, without needing a whole other config value to switch back from the subclass to plain String)

PR #56164, opened 2025-11-14
Add Hash filtering helpers for key/value predicates

faraquet:
### Summary
Introduce new Hash helpers to filter by key/value predicates:

- `Hash#reject_if_value`, `Hash#select_if_value`
- `Hash#reject_if_key`, `Hash#select_if_key`
- Bang variants: `reject_if_value!`, `select_if_value!`, `reject_if_key!`, `select_if_key!`

Predicates can be provided as:
- a method name (Symbol/String) called on the key/value
- a callable (responds to `#call`)
- a block

### Motivation
Filtering Hashes by key or value is a common pattern in Rails apps. While `select`/`reject` exist, they’re block-only and require repeating predicate logic at callsites. Providing first-class helpers improves readability and mirrors other Active Support ergonomic APIs.

### Public API
- `hash.reject_if_value(condition = nil, *args, &block)`
- `hash.select_if_value(condition = nil, *args, &block)`
- `hash.reject_if_key(condition = nil, *args, &block)`
- `hash.select_if_key(condition = nil, *args, &block)`
- `hash.reject_if_value!(...)`, `hash.select_if_value!(...)`
- `hash.reject_if_key!(...)`, `hash.select_if_key!(...)`

Behavior:
- If given a block, it’s used as the predicate.
- If given a callable, it’s invoked with the key or value.
- If given a method name, it’s called on the key/value when it responds to that method; otherwise the predicate returns false.
- Bang variants mutate and always return `self`.

Note: No exceptions are rescued within the predicate. For methods that can raise with incompatible types (e.g., `:>`), filter to a compatible type first (see examples).

### Examples

```ruby
{ a: "", b: 1, c: nil }.reject_if_value(:blank?)
# => { b: 1 }

{ "user_1" => 1, "admin_1" => 2 }.select_if_key(:match?, /\Auser_/)
# => { "user_1" => 1 }

h = { a: "x", b: 2, c: 3 }
numerics = h.select_if_value(:is_a?, Numeric)
numerics.select_if_value(:>, 2)
# => { c: 3 }

# Bang variants
h = { a: "", b: "x", c: [] }
h.reject_if_value!(:empty?) # => self, h is now { b: "x" }

# Block predicate on values
{ a: 1, b: 2, c: 3 }.reject_if_value { |v| v.odd? }
# => { b: 2 }

# Callable predicate on values
even = ->(v) { v.even? }
{ a: 1, b: 2, c: 3 }.select_if_value(even)
# => { b: 2 }

# Key filtering with regex via :match?
{ "user_1" => 1, "admin_1" => 2, "user_2" => 3 }.reject_if_key(:match?, /\Aadmin_/)
# => { "user_1" => 1, "user_2" => 3 }

# Numeric key predicates
{ 1 => "a", 2 => "b", 3 => "c" }.select_if_key(:odd?)
# => { 1 => "a", 3 => "c" }

# Value method with arguments (multiple)
{ a: "alpha", b: "beta", c: "gamma" }.select_if_value(:start_with?, "a", "b")
# => { a: "alpha", b: "beta" }

# Chaining: narrow to numerics, then apply operator
h = { a: "x", b: 2, c: 3 }
h.select_if_value(:is_a?, Numeric).select_if_value(:>, 2)
# => { c: 3 }

# Key length via block
{ a: 1, ab: 2, abc: 3 }.select_if_key { |k| k.to_s.length >= 2 }
# => { ab: 2, abc: 3 }

# Bang variants mutate and return self
h = { a: "", b: "x", c: [] }
h.reject_if_value!(:empty?) # => self; h is now { b: "x" }

# Defaults/default_proc preserved with bang variants
h = Hash.new { |hh, k| hh[k] = [] }
h.update(a: [], b: [1]).select_if_value!(:empty?)
h[:missing] # => [] (default_proc still active)
```


fatkodima:
Can you explain, why do you think 
`h.select_if_value(:is_a?, Numeric)`
is better than 
`h.select { |k, v| v.is_a?(Numeric) }` ?

faraquet:
Hi @fatkodima 👋
Thanks for taking a look.

It's syntactic sugar. In a one-liner, the difference isn't much significant, but in methods with many iterations, it becomes more noticeable.

In terms of intent clarity, it expresses 'filter by value' directly, without the `|_, v|` noise, which I personally find the most distracting. 

```ruby
h = { "user_1" => "Doe John", "admin_1" => "Doe Steve", "user_2" => "Hope Jenny" }

# Old
x = h.reject { |k, _| k.match?(/admin/) }
# some stuff in between
x.select { |_, v| v.start_with?("Doe") }

# New
x = h.reject_if_key(:match?, /admin/)
# some stuff in between
x.select_if_value(:start_with?, "Doe")

```

I also like the predicate flexibility: passing a method name, a callable, or a block (with arguments) allows for reusing and composing predicates.

```ruby
cfg = { "cfg_paths" => ["/a"], "cfg_env" => [], "path" => "/tmp" }

cfg_key       = ->(k) { k.to_s.start_with?("cfg_") }
non_empty_ary = ->(v) { v.is_a?(Array) && !v.empty? }

# Old
cfg.select { |k, _| cfg_key.call(k) }
   .select { |_, v| non_empty_ary.call(v) }

# New
cfg.select_if_key(cfg_key)
   .select_if_value(non_empty_ary)

```

Here are a few other useful notes: 

Consistent naming is helpful when grepping a codebase, as it makes similar operations easier to locate.

Chaining reads naturally as a pipeline, which makes multi-step filters easier to read. `h.select_if_value(:is_a?, Numeric).select_if_value(:>, 2)`

For in-place operations, I prefer that the bang methods always return self. Unlike `Hash#select!`, which can return nil, `select_if_value!` is predictable, so mutation flows are safer to write and reason about.

fatkodima:
Thanks for the PR! But I think the current approach is nicer and it does not introduce another method to the API - it is just easier to use original method than to remember the new syntax.

faraquet:
Thanks for the review anyway @fatkodima 👍

PR #56167, opened 2025-11-15
Replace TimeZoneConversion's DelegateClass

hmcguire-shopify:
ActiveSupport's `delegate` is much faster than `DelegateClass`, and the slowness of `DelegateClass` shows up when using `InsertAll`.

With the benchmark from a [previous commit]:

Before:

```
=== Running benchmark-ips comparison ===
ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [arm64-darwin23]
         insert_all!     1.000 i/100ms
             raw_sql     1.000 i/100ms
         insert_all!      4.562 (± 0.0%) i/s  (219.23 ms/i) -     23.000 in   5.044978s
             raw_sql     13.340 (± 0.0%) i/s   (74.96 ms/i) -     67.000 in   5.025671s

Comparison:
             raw_sql:       13.3 i/s
         insert_all!:        4.6 i/s - 2.92x  slower
```

After:

```
=== Running benchmark-ips comparison ===
ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [arm64-darwin23]
         insert_all!     1.000 i/100ms
             raw_sql     1.000 i/100ms
         insert_all!      4.875 (± 0.0%) i/s  (205.11 ms/i) -     25.000 in   5.132894s
             raw_sql     12.756 (± 0.0%) i/s   (78.40 ms/i) -     64.000 in   5.021686s

Comparison:
             raw_sql:       12.8 i/s
         insert_all!:        4.9 i/s - 2.62x  slower
```

[previous commit]: https://github.com/rails/rails/commit/026df3ca8b94b68a7d611dd44d2b43c04edc130a

### DIFF
--- a/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb
+++ b/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb
@@ -5,23 +5,27 @@ require "active_support/core_ext/object/try"
 module ActiveRecord
   module AttributeMethods
     module TimeZoneConversion
-      class TimeZoneConverter < DelegateClass(Type::Value) # :nodoc:
+      class TimeZoneConverter # :nodoc:
         def self.new(subtype)
           self === subtype ? subtype : super
         end
 
+        def initialize(subtype)
+          @subtype = subtype
+        end
+
         def deserialize(value)
-          convert_time_to_time_zone(super)
+          convert_time_to_time_zone(@subtype.deserialize(value))
         end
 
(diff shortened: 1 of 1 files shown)


byroot:
Ref: https://github.com/ruby/delegate/pull/46

But also we probably should just implement our own class delegator as we can do even better if we're not bound to the `delegate.rb` spec (e.g. calling `__getobj__` etc.

PR #56170, opened 2025-11-15
Add structured event `active_record.deprecated_association` when mode is set to `:notify`

zzak:
Currently, mode can only be configured by global config:

```
config.active_record.deprecated_associations_options = { mode: :notify }
```

The default is `:warn`:
<https://edgeguides.rubyonrails.org/configuring.html#config-active-record-deprecated-associations-options>

/cc @gmcgibbon @adrianna-chang-shopify @fxn

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add structured event `active_record.deprecated_association` when mode is set to `:notify`.
+
+    ```
+    config.active_record.deprecated_associations_options = { mode: :notify }
+    ```
+
+    *zzak*
+
 *   On MySQL parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
 
     Truncating on MySQL is very slow even on empty or nearly empty tables.
--- a/activerecord/lib/active_record/structured_event_subscriber.rb
+++ b/activerecord/lib/active_record/structured_event_subscriber.rb
@@ -56,6 +56,18 @@ module ActiveRecord
     end
     debug_only :sql
 
+    def deprecated_association(event)
+      payload = {
+        reflection: event.payload[:reflection],
+        message: event.payload[:message],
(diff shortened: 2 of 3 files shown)


fxn:
Good, we do not want to instrument anything, we want to send an event with event metadata and that's it.

This is better interface and maybe we could even deprecate the notification.

I wonder if there is a pattern here `forward_notification` which emits a structured event with the same name and same payload except for the instrumentation keys.

zzak:
> This is better interface and maybe we could even deprecate the notification.

This is built on top of the notification so we shouldn't deprecate it, maybe we want to make it default (always emit) though?

> RE: `forward_notification`

I'm hesitant to add it because there are not so many events I've seen that are this simple, and having explicit payload means we don't accidentally break compatibility by modifying the underlying notification somehow (I guess as a strawman argument).

fxn:
> This is built on top of the notification so we shouldn't deprecate it, maybe we want to make it default (always emit) though?

When I implemented deprecated associations, I emitted an AS notification because that is what we had, but I did not like it.

The fact that this patch uses a subscriber is an implementation detail, no? The patch that I would like is that the core of the deprecated associations feature directly emits a structured event.

fxn:
Yeah, so after sleeping on this, the natural evolution that I see here is to add another mode:

```ruby
config.active_record.deprecated_associations_options = { mode: :report_event }
```

Precisely because structured events are different than instrumentation. I don't think we should mix them under the same mode.

Would "report" be an idiomatic verb? Maybe "publish"?

zzak:
@fxn I've opened another PR based on your feedback, and left my thoughts.

> the natural evolution that I see here is to add another mode:

This is basically what I did but with the name `:event` instead, we can discuss it over there if you prefer! :bow:

zzak:
I think we're going to use an artisnal hand-crafted event instead of this, so closing :joy:

PR #56175, opened 2025-11-16
Make flaky parallel tests easier to diagnose by deterministically assigning tests to workers

jeremy:
Rails assigns tests to workers in round-robin order so the same `--seed` and worker count will result in the same sequence of tests running on each worker (whether processes or threads) increasing the odds of reproducing test failures caused by test interdependence.

This can make test runtime slower and spikier when one worker gets most of the slow tests. Enable `work_stealing: true` to allow idle workers to steal tests from busy workers in deterministic order, smoothing out runtime at the cost of less reproducible flaky-test failures.

(Update: removed concurrent test reportsfor each test failures.)

Performance:
* Seeing statistically insignificant difference in test suite times before/after, and with work stealing enabled/disabled. Too much variance; need more samples.
* See https://github.com/basecamp/fizzy/pull/2037 for validation.

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Make flaky parallel tests easier to diagnose by deterministically assigning
+    tests to workers.
+
+    Rails assigns tests to workers in round-robin order so the same `--seed`
+    and worker count will result in the same sequence of tests running on each
+    worker (whether processes or threads) increasing the odds of reproducing
+    test failures caused by test interdependence.
+
+    This can make test runtime slower and spikier when one worker gets most of
+    the slow tests. Enable `work_stealing: true` to allow idle workers to steal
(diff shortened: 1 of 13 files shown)


dhh:
Why would this not just be the default?

> This can make test runtime slower and spikier when one worker gets most of the slow tests. Enable work_stealing: true to allow idle workers to steal tests from busy workers in deterministic order, smoothing out runtime at the cost of less reproducible flaky-test failures.

jeremy:
> Why would this not just be the default?

Because it won't be noticeable in most test suites (you need a *really* lumpy test suite to benefit) and once you start hitting flaky test issues, you'd have to turn it off for more predictable diagnostics.

Realistically, I added it prematurely anticipating pushback re. imbalanced test distribution in outlier suites. (And, because it's cool.)

PR #56177, opened 2025-11-16
Fix typos in test descriptions & code for clarity

shivabhusal:
### Motivation / Background

```ruby
assert_not_includes ActiveRecord::Base.filter_attributes, "messsage.special_attr"
```

Test uses "messsage.special_attr" (with a typo), which caused the test to assert against an unintended string. 

This made the test misleading because it appeared to validate filtering behavior, but the assertion was actually checking for a misspelled attribute, not the intended `"message.special_attr"` key.


shivabhusal:
The failing test doesn’t seem related to this PR. Please let me know how to proceed.

PR #56156, opened 2025-11-14
AS Structured Events Guide [ci skip]

zzak:
### :camera_flash: Preview: <https://guides-structured-events.rails-docs-preview.pages.dev/guides/active_support_structured_events>

Been meaning to put this down into words, basically summarize the Event Reporting and Structured Events APIs in the guides, similar to AS::Notifications. A lot of this info comes from @adrianna-chang-shopify's excellent [talk](https://www.youtube.com/watch?v=JK90_2L_NUM) from RailsWorld this year.

Basically to try and answer what are they, how are they different from notifications, and what events does Rails come with.

Decided to make this an upstream branch, I'm not sure I will be able to take it to the :checkered_flag: finish line, so I appreciate any review and feedback here.

PR #56179, opened 2025-11-16
Use event reporter for active_record.deprecated_association

zzak:
Replaces the existing notification and deprecates `:notify` option.

Based on feedback: https://github.com/rails/rails/pull/56170#issuecomment-3536213783

> When I implemented deprecated associations, I emitted an AS notification because that is what we had, but I did not like it.
> 
> The fact that this patch uses a subscriber is an implementation detail, no? The patch that I would like is that the core of the deprecated associations feature directly emits a structured event.

This is an alternative to #56170 that uses event reporter directly to emit a structured event for deprecated associations.

The benefits to this approach are simpler, and less overhead than going through notifications and structured event subscribers.

My concern is that this introduces a new pattern for emitting events in the framework, where the other PR follows the established pattern of (emit notification -> event subscribers -> reporter). Is this a one-off, or are there other cases where we would replace existing business specific notifications with events.

The other concern is how neatly this fits in to existing event subscribers, but since everything flows through the reporter (there should only be one) then it's probably fine -- I just wanted to note that as a possible concern even if it's lower risk. In theory, event subscribers will emit it like any other event, or can filter if necessary.

I think this feature also sits separately from the other notifications, for example `active_record.sql` or `active_job.completed` are more _framework-level_ telling us what's happening in the system. But associations are very much apart of the business domain of the application, not an internal event or side-effect of something Rails did, like process a request. So I can see the argument for not using notifications.

That said, I would like to get feedback from @gmcgibbon, @adrianna-chang-shopify, and @rafaelfranca on this.

/cc @fxn

### DIFF
--- a/activerecord/test/cases/associations/deprecation_test.rb
+++ b/activerecord/test/cases/associations/deprecation_test.rb
@@ -3,6 +3,8 @@
 require "cases/helper"
 require "models/dats"
 
+require "active_support/event_reporter/test_helper"
+
 # The logic of the `guard` method is extensively tested indirectly, via the test
 # suites of each type of association.
 #
@@ -54,10 +56,10 @@ module AssociationDeprecationTest
 
     test "valid options, both" do
       ActiveRecord.deprecated_associations_options = {
-        mode: :notify,
+        mode: :event,
         backtrace: true
       }
-      assert_equal :notify, ActiveRecord::Associations::Deprecation.mode
+      assert_equal :event, ActiveRecord::Associations::Deprecation.mode
       assert_equal true, ActiveRecord::Associations::Deprecation.backtrace
     end
 
@@ -79,15 +81,26 @@ module AssociationDeprecationTest
       error = assert_raises(ArgumentError) do
         ActiveRecord.deprecated_associations_options = { mode: :invalid }
       end
-      assert_equal "invalid deprecated associations mode :invalid (valid modes are :warn, :raise, and :notify)", error.message
+      assert_equal "invalid deprecated associations mode :invalid (valid modes are :warn, :raise, and :event)", error.message
     end
   end
 
   class ModeWriterTest < TestCase
     test "valid values" do
-      [:warn, :raise, :notify].each do |mode|
+      [:warn, :raise, :event].each do |mode|
         ActiveRecord::Associations::Deprecation.mode = mode
         assert_equal mode, ActiveRecord::Associations::Deprecation.mode
+      end
+    end
+
+    test ":notify is deprecated" do
+      msg = <<~MSG.squish
+        The :notify mode for deprecated associations is deprecated and will be removed in Rails 9.0.
+        Please use `config.active_record.deprecated_associations_options[:mode] = :event` and instead.
+      MSG
+      assert_deprecated(ActiveRecord.deprecator, msg: msg) do
+        ActiveRecord::Associations::Deprecation.mode = :notify
+        assert_equal :notify, ActiveRecord::Associations::Deprecation.mode
       end
     end
 
@@ -95,7 +108,7 @@ module AssociationDeprecationTest
       error = assert_raises(ArgumentError) do
         ActiveRecord::Associations::Deprecation.mode = :invalid
       end
-      assert_equal "invalid deprecated associations mode :invalid (valid modes are :warn, :raise, and :notify)", error.message
+      assert_equal "invalid deprecated associations mode :invalid (valid modes are :warn, :raise, and :event)", error.message
     end
 
     test "the backtrace flag becomes a true/false singleton" do
@@ -203,7 +216,9 @@ module AssociationDeprecationTest
   class NotifyModeTest < TestCase
     def setup
       super
-      ActiveRecord::Associations::Deprecation.mode = :notify
+      assert_deprecated(ActiveRecord.deprecator) do
+        ActiveRecord::Associations::Deprecation.mode = :notify
+      end
       ActiveRecord::Associations::Deprecation.backtrace = true
     end
 
@@ -269,6 +284,55 @@ module AssociationDeprecationTest
 
     test "HABTM receives the user-facing reflection in the payload" do
       assert_user_facing_reflection(DATS::Category, :deprecated_posts)
+    end
+  end
+
+  class EventModeTest < TestCase
+    include ActiveSupport::EventReporter::TestHelper
+
+    def setup
+      super
+      @subscriber = ActiveSupport::EventReporter::TestHelper::EventSubscriber.new
+      ActiveSupport.event_reporter.subscribe(@subscriber)
+      ActiveRecord::Associations::Deprecation.mode = :event
+      ActiveRecord::Associations::Deprecation.backtrace = true
+    end
+
+    def teardown
+      super
+      ActiveSupport.event_reporter.unsubscribe(@subscriber)
+    end
+
+    def test_deprecated_associations
+      bc = ActiveSupport::BacktraceCleaner.new
+      bc.add_silencer { !_1.include?(__FILE__.sub(%r(.*?/activerecord/(?=test/)), "")) }
+      ActiveRecord::LogSubscriber.backtrace_cleaner = bc
+
+      line = __LINE__ + 1
+      DATS::Car.new.deprecated_tires
+
+      assert_equal 1, @subscriber.events.size
+
+      event = @subscriber.events.first
+
+      if RUBY_VERSION < "3.4"
+        assert_match "report", event[:source_location][:label]
+      else
+        assert_equal "ActiveRecord::Associations::Deprecation.report", event[:source_location][:label]
+      end
+
+      payload = event[:payload]
+
+      assert_equal DATS::Car.reflect_on_association(:deprecated_tires), payload[:reflection]
+
+      re = /The association DATS::Car#deprecated_tires is deprecated, the method deprecated_tires was invoked \(#{__FILE__}:#{line}:in/
+      assert_match re, payload[:message]
+
+      assert_equal __FILE__, payload[:location].path
+      assert_equal line, payload[:location].lineno
+
+      assert_equal __FILE__, payload[:backtrace][0].path
+      assert_equal line, payload[:backtrace][0].lineno
     end
   end
 end
--- a/guides/source/active_support_instrumentation.md
+++ b/guides/source/active_support_instrumentation.md
@@ -659,23 +659,6 @@ In practice, you cannot do much with the transaction object, but it may still be
 helpful for tracing database activity. For example, by tracking
 `transaction.uuid`.
 
-#### `deprecated_association.active_record`
-
-This event is emitted when a deprecated association is accessed, and the
-configured deprecated associations mode is `:notify`.
-
-| Key                  | Value                                                |
-| -------------------- | ---------------------------------------------------- |
-| `:reflection`        | The reflection of the association                    |
-| `:message`           | A descriptive message about the access               |
-| `:location`          | The application-level location of the access         |
-| `:backtrace`         | Only present if the option `:backtrace` is true      |
-
-The `:location` is a `Thread::Backtrace::Location` object, and `:backtrace`, if
-present, is an array of `Thread::Backtrace::Location` objects. These are
-computed using the Active Record backtrace cleaner. In Rails applications, this
-is the same as `Rails.backtrace_cleaner`.
-
 ### Active Storage
 
 #### `preview.active_storage`
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add structured event `active_record.deprecated_association` when mode is set to `:notify`.
+
+    ```
+    config.active_record.deprecated_associations_options = { mode: :notify }
+    ```
+
+    *zzak*
+
 *   On MySQL parallel test database table reset to use `DELETE` instead of `TRUNCATE`.
 
     Truncating on MySQL is very slow even on empty or nearly empty tables.
--- a/activerecord/lib/active_record/associations/deprecation.rb
+++ b/activerecord/lib/active_record/associations/deprecation.rb
@@ -7,15 +7,21 @@ module ActiveRecord::Associations::Deprecation # :nodoc:
   EVENT = "deprecated_association.active_record"
   private_constant :EVENT
 
-  MODES = [:warn, :raise, :notify].freeze
+  MODES = [:warn, :raise, :notify, :event].freeze
   private_constant :MODES
 
   class << self
     attr_reader :mode, :backtrace
 
     def mode=(value) # private setter
+      if value == :notify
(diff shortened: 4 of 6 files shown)


fxn:
> My concern is that this introduces a new pattern for emitting events in the framework, where the other PR follows the established pattern of (emit notification -> event subscribers -> reporter). Is this a one-off, or are there other cases where we would replace existing business specific notifications with events.

Let me put it this way: If I had structured events at the time, I would **not** have emitted an AS notification.

We have to break the inertia, in my view. When you want to "notify" about something, is there instrumentation-related data that you want to send? Like, how long did a query take? If yes, then you want a notification. If no, I only want to signal something happened, that is an event.

fxn:
Of course, open also to what @gmcgibbon, @adrianna-chang-shopify, and @rafaelfranca think, they have the vision for events.

adrianna-chang-shopify:
> If yes, then you want a notification. If no, I only want to signal something happened, that is an event.

Yes, this matches my mental model too; seems like this is a good opportunity to delineate "instrumentation events" from "reporting events". We had the Event Reporter consume all existing AS notifications for backwards compatibility, but for new events that are "signal-only", it makes sense to me that those would be emitted directly via the Event Reporter. Rather than supporting two distinct patterns, users should reach for the Event Reporter API vs writing their own subscriber to consume the Active Support notification.

activerecord/test/cases/associations/deprecation_test.rb
```diff
@@ -271,4 +286,53 @@ def assert_user_facing_reflection(model, association)
...
+
+    def test_deprecated_associations
+      bc = ActiveSupport::BacktraceCleaner.new
+      bc.add_silencer { !_1.include?(__FILE__.sub(%r(.*?/activerecord/(?=test/)), "")) }
+      ActiveRecord::LogSubscriber.backtrace_cleaner = bc
+
+      line = __LINE__ + 1
+      DATS::Car.new.deprecated_tires
+
+      assert_equal 1, @subscriber.events.size
```
adrianna-chang-shopify, line 314: Instead of instantiating a test subscriber and inspecting its events, can we use `assert_event_reported`?

```ruby
reflection = DATS::Car.reflect_on_association(:deprecated_tires)
event = assert_event_reported("active_record.deprecated_association", payload: { reflection: reflection }) do
  DATS::Car.new.deprecated_tires
end

if RUBY_VERSION < "3.4"
  assert_match "report", event[:source_location][:label]
else
  assert_equal "ActiveRecord::Associations::Deprecation.report", event[:source_location][:label]
end
...
```

guides/source/active_support_instrumentation.md
```diff
@@ -659,23 +659,6 @@ In practice, you cannot do much with the transaction object, but it may still be
 helpful for tracing database activity. For example, by tracking
 `transaction.uuid`.
 
-#### `deprecated_association.active_record`
```
adrianna-chang-shopify, line 662: Will be great to have a corresponding guide for events, taking a look at your documentation PR next 🙇‍♀️

activerecord/test/cases/associations/deprecation_test.rb
zzak: I wanted to but because the backtrace and location were giving me trouble, I just used the style from the previous notification test for this, I will give it another go though :muscle: Thanks for your review!

adrianna-chang-shopify:
Hey @zzak and @fxn! After some discussion with folks on my team (particularly @Edouard-chin), I'm actually reconsidering the original proposal to have "signal-only" events bypass Active Support Notifications.

The challenge is this: "instrumentation vs signal-only" is a less useful distinction than "is this an event a user would want to hook into to invoke some behaviour, or is this purely information reporting?" We _don't_ want users to have to sometimes hook into Active Support notifications and other times into the Event Reporter; anything that is a "framework hook" should go through Active Support notifications IMHO.

With the existing framework hooks, we can't easily predict which events users will find valuable for hooking into. Arguably all of them are reasonable hook points (including `active_record.deprecated_association`). 

I'm wondering if the right approach for `deprecated_association` would be to _always_ emit the ASN event (as a standard framework hook point), and then have `:notify` trigger sending the event to the Event Reporter. The mode configuration in its current form is indeed a bit odd - typically Rails just instruments with ASN by default, and users subscribe if they care and ignore if they don't. So to me, it would make sense to always produce the Active Support notification, which users can hook into if they want, and then provide a `:notify` option that handles reporting the event via the Event Reporter.

Thoughts?

fxn:
@adrianna-chang-shopify @Edouard-chin let me give you my point of view.

Active Support Notifications were historically introduced to provide a framework for APMs. So, the method is `instrument`, not `publish`, or `notify`, and what you get is a payload with instrumentation data:

```ruby
ActiveSupport::Notifications.subscribe('render') do |name, start, finish, id, payload|
  # ...
end
```

it has `start` and `finish`, it could as well have come with `allocated_objects` or whatever cheap instrumentation information you could add to that. Because that is what the API is about.

The problem is that since that existed, its usage has been abused during the years for stuff that is not instrumented. You just want to "notify" that something happened. As is the case here, I want you to know that a deprecated association was used. Instrumentation has nothing to do with this.

As emitter, I'll only do the work that I need to do, no time checking, no allocated objects counting (in that hypothetical example I am using to help my point be more apparent). And as subscriber, you don't get weird, irrelevant input. You get the data that makes sense for the event.

Sometimes I played with the idea of adding `publish` to AS notifications, which would have no instrumentation. But I was not convinced with how it looked in practice.

However, now we have structured events. The way I see it, this fits very well what I think the framework was missing. Finally, a way to send pure events unrelated to instrumentation.

That is why I think about this in terms of "inertia". We are so used to use instrumentation as notification that the mental model persists. But I believe it is the time to start departing from that.

(BTW, my pet peeve with structured events is that subscribers are supposed to `emit`. Why emit? The reason why a subscriber "listens" to the events may not necessarily be emitting anything.)

zzak:
Another side of this is other than event reporting, this feature can also be configured to raise an exception -- but I assume that is not intended for production; therefor why not use the error reporter?

To take that further, why shouldn't any part of rails that has the option to raise or not either (always) be emitting a framework event or using the error reporter (sans-compatibility)?

Not to distract the thread, but just a thought I had.

> We don't want users to have to sometimes hook into Active Support notifications and other times into the Event Reporter; anything that is a "framework hook" should go through Active Support notifications IMHO.

I thought that the Structured Event Subscribers job was to capture all of those notifications for the event reporter, which maybe don't have events, and then user only needs to care about whether their event subscriber wants to deal with it. When I was looking at the `diff` of events vs. notification hooks I was thinking should these all be implemented eventually? There are some events I thought about adding after this, like rate limit. I think it's important to have this discussion before migrating more events. :pray:

Also, I think we've never had to deprecate a notification or part of the payload, and only been making additive changes afaik. Which maybe lead to this situation of "intertia" where the notifications API has been used for not instrumentation events.

The question is do we want to try and break that pattern, is this feature a good reason to? I'm seeing both sides of the argument and maybe I prefer to say "let's just keep all framework related hooks as notifications" and then decide which ones are worth making events too. So we have a consistent underlying internal API to build on top of, if we want to do other stuff like sampling or deprecations, idk. Or is Event Reporter actually the right place for that and we should leave ASN for instrumentation + all the other stuff we added until now.

adrianna-chang-shopify:
@fxn thanks for sharing your mental model. I definitely understand the motivation to break the inertia, and wanting to scope ASN towards true instrumentation events.

I think the challenge is that we designed the Event Reporter to solve a very specific problem: reporting telemetry and business events to observability / logging pipelines. It wasn't really meant to be replace the existing "Rails framework hooks" mechanism with AS Notifications (hence us choosing `emit` as the subscriber API; when we built this internally it was "emit to the logging pipeline"). The Event Reporter is designed to have subscribers representing different "sinks" we want to send structured data to, not subscribers listening for individual events and invoking certain behaviour.

> I thought that the Structured Event Subscribers job was to capture all of those notifications for the event reporter, which maybe don't have events, and then user only needs to care about whether their event subscriber wants to deal with it.

Right, in my mind, the AS notification is the "framework event", and then we turn these into "reporting-level events" (structured events), just as we turn them into "unstructured log lines" via the `LogSubscribers`. My mental model is that pretty much all "framework events" get picked up by the Event Reporter as "structured logs", and subscribers can choose to filter out any they don't care about.

> I prefer to say "let's just keep all framework related hooks as notifications" and then decide which ones are worth making events too

This is where I'm at. I'm open to further discussion about using the Event Reporter for signal-only framework events, but it seems strange to have two very separate APIs where one is for events that include a couple of extra bits of instrumentation data, and the other for events without that data. (Especially if _everything_ is going through the Event Reporter anyways as a structured log, it feels confusing to sometimes be subscribing to the event reporter directly to invoke behaviour vs sometimes to an AS notification, simply based on whether the event is expected to contain instrumentation-related metadata)

cc @rafaelfranca I'd be very interested to hear your thoughts here as well.

Edouard-chin:
@fxn and @zzak Thanks for sharing your point of view. 

> I want you to know that a deprecated association was used. Instrumentation has nothing to do with this. 

The thing is that the line is blurry on other signals that Rails sends. For instance when a request is being redirected, Action Controller sends a notification. Looking at the implementation, there is nothing to instrument (it just sets some headers and generate the redirection path), so sending an Event would be more suited. But as a user, how would I know that beforehand? 

I feel like each time I'd have to create a subscriber I'd have to look at the docs to know what signal I have to subscribe to.

Similarly, in the perspective of a user, it seems also less ideal that I can't create a single subscriber listen to all request related events and now have to deal with a notification payload that has a different structure.

As the framework evolves, some existing ASN may become more suited for Event and vice versa, would this mean we'll change the signal we send and expects user to change their subscription?

It seems to me that sending an event or an ASN is an implementation detail of the framework and that it shouldn't affect how the end user consumes those signals.

fxn:
Hey, just a heads up that I have this pending to reply.

gmcgibbon:
I think it is important to reflect on the decisions we've already made.

- We have Active Support Notifications for every kind of event historically.
- We intentionally made a structured event subscriber to convert select notifications to structured events.
- We refactored log subscribers to consume structured events and spit out logs.

To me, it depends on the level at which the event matters, similar to what @fxn was saying about actual instrumentation vs signalling something happened.  Notifications have been historically abused and we've gotten used to using that as generic endpoints for all events, but I agree we should step away from it where we can.

We currently have 4 (maybe more?) signals-as-notifications: `open_redirect.action_controller`, `unsafe_redirect.action_controller`, `strict_loading_violation.active_record`, `deprecated_association.active_record`. Some of them also optionally write directly to the log, which convolutes things a little more. I think these should be using the event reporter from a framework maintainer standpoint, but I understand it is confusing from a user standpoint how we should be listening.

More specifically related to this PR, perhaps we can expose a `notify_behavior` option in the config to say what `notify` options do (emit an `event` or `notification`)? We could emit a deprecation when it is set to `notification` and flip it in the next Rails version to give developers time to switch. That way, these signals would go to the right place eventually, and we can save instrumentation notifications for things that really need them.

Ultimately, it is harder to take notifications away than it is to add them retrospectively if we want to. IMO the developer building a Rails app should be interfacing with the event reporter as much as possible, and notification instrumentation as little as possible. This should help set that standard.

fxn:
> But as a user, how would I know that beforehand?

You know about the notification name and type because the docs tell you about it (that is how you learn the event exists in the first place).

Right now [this subscriber](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/structured_event_subscriber.rb) is forwarding a bunch of Active Record notifications. How is the user supposed to know those are available in both forms? They know because they have read the docs.

So, that, by itself, does not seem like an issue to me.

However, I find that subscriber to be highly dubious. For example, why are we forwarding `sql.active_record` at all? I don't think that is a good idea. That is an instrumentation event, if you want to integrate it in your events pipeline, you do it in your app. Otherwise, why stop in those few? Given those, people are going to ask "and why not this other one? here's a patch!"

You all of a sudden have lowered the barrier, and everything is the same. But it should not be the same. Events are events, and instrumentation is instrumentation. Nowadays, that is blurred, but for historical reasons. Now that we have the tools, we can fix that.

And, personally, I would delete that subscriber, or use it strictly for the non-instrumentation events that today are sent as AS notifications, as a transition measure that also conveys this message.

> IMO the developer building a Rails app should be interfacing with the event reporter as much as possible, and notification instrumentation as little as possible.

Agree, that is one of those framework-vision decisions.

/cc @Ridhwana

fxn:
Another option could be to standardize on structured events.

In this alternative, the framework sends structured events for everything. Some of them may also be instrumented (meaning, there is an AS notification you can subscribe to).

For example, the times I have needed to monitor SQL events it has been for logic unrelated to instrumentation.

@gmcgibbon indeed there are more notifications that are not instrumented, some additional examples:

```
halted_callback.action_controller
rescue_from_callback.action_controller
start_processing.action_controller
unpermitted_parameters.action_controller
start_transaction.active_record
...
```

adrianna-chang-shopify:
>  IMO the developer building a Rails app should be interfacing with the event reporter as much as possible

This requires a fundamental reframing of what the Event Reporter was built for, though. We built it as a structured logger; a way to report structured logs to an Observability service. _Not_ as a general pub/sub mechanism, or a way to hook into framework events. I did not foresee users writing "subscribers" to events to invoke application code; subscribers were meant to be listen to events and **emit** them to an Observability / analytics platform. The current API docs don't make this clear enough, which is probably contributing to this problem, but the [guide](https://github.com/rails/rails/pull/56156) @zzak has been working on makes this clearer.

> Right now [this subscriber](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/structured_event_subscriber.rb) is forwarding a bunch of Active Record notifications. How is the user supposed to know those are available in both forms?

IMO a user shouldn't really need to "know" about the events coming from the Event Reporter; it's equivalent to the unstructured log the `ActiveRecord::LogSubscriber` produces, but in a structured form. Most applications (ours included) implement a  subscriber that listens to _all_ events and passes them onto the logging pipeline. But, if an application wanted to ignore certain structured logs, they could do so using a filter.

I'm still of the opinion that if applications want to hook into framework events to invoke some behaviour, ASN should be used to do this, not Event Reporter subscribers. And I believe that all AS Notifications _should_ go to the Event Reporter; that's about capturing everything that occurred in a Rails request from an Observability standpoint.

Re: signals-as-notifications, I do think it makes sense to have these emit structured logs instead of notifications across the board. `:event` IMO is the structured equivalent to `:log`. And, if there are currently AS Notifications that we _only_ want to "log" to users (without offering up a hook point), those could be converted to calling the Event Reporter directly.

fxn:
@adrianna-chang-shopify I guess the name "structured event" does not help, because it all seems like a pub/sub system until you arrive to the name `emit`. When I looked into it I thought (perhaps mistakenly) that it was a generic pub/sub interface where in the very last step, the intended internal usage leaked in that name.

Maybe "observability events" would have communicated better, because from conversations I believe the expectation needs to be corrected.

Are there design choices that say: "look, this is not a generic pub/sub system, you can see it in how this and that works"?

PR #56182, opened 2025-11-16
fix: acronym inflector should give precedence to longer matches

### LINKED ISSUE #56181: [Active Support] Acronym inflector should give precedence to longer matches
### Steps to reproduce

When two acronym inflections are defined where one is a prefix of the other (e.g., "JWK" and "JWKS"), the inflector incorrectly matches the shorter acronym first when processing the longer string.

This causes incorrect `underscore` output for the longer acronym.

### Expected behavior

Given the following inflections:
```ruby
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym "JWK"  # JSON Web Key
  inflect.acronym "JWKS" # JSON Web Key Set
end
```
We expect `"JWKS".underscore` to return `"jwks"`.

### Actual behavior

`"JWKS".underscore` incorrectly returns `"jwk_s"`.

The inflector appears to match "JWK" and then incorrectly treats the "S" as a separate word to be underscored. The rules should be applied to find the longest possible match first.

### System configuration

**Rails version**: (Please fill in your Rails version, e.g., 7.1.2 or `main` branch)
**Ruby version**: (Please fill in your Ruby version, e.g., 3.3.0)

---

### Executable Test Case

Here is an executable test script based on the "Generic template for other issues". You can run this file with `ruby bug_report.rb` to demonstrate the failure.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "7.1.2" # Specify the Rails version you are using
  # Or use main
  gem "minitest"
end

require "active_support"
require "active_support/inflector"
require "minitest/autorun"

class AcronymPrecedenceTest < Minitest::Test
  def setup
    # Ensure a clean state by clearing inflections
    ActiveSupport::Inflector.inflections(:en).clear

    # Define inflections in the order that demonstrates the bug
    # Shorter acronym is defined first
    ActiveSupport::Inflector.inflections(:en) do |inflect|
      inflect.acronym "JWK"  # JSON Web Key
      inflect.acronym "JWKS" # JSON Web Key Set
    end
  end

  def teardown
    # Clear inflections after the test
    ActiveSupport::Inflector.inflections(:en).clear
  end

  def test_longer_acronym_takes_precedence
    expected = "jwks"
    actual = "JWKS".underscore

    # This assertion will fail, demonstrating the bug.
    assert_equal expected, actual, "Expected 'JWKS' to underscore to 'jwks', but got '#{actual}'"
  end

  def test_shorter_acronym_still_works_correctly
    expected = "jwk"
    actual = "JWK".underscore

    # This assertion should pass
    assert_equal expected, actual, "Expected 'JWK' to underscore to 'jwk'"
  end
end
```

zubin:
### Motivation / Background

When multiple acronyms are defined where one is a prefix of another (e.g., "JWK" and "JWKS"), the inflector would incorrectly match the shorter acronym first, causing "JWKS".underscore to return "jwk_s" instead of "jwks".

Fixes #56181

### Detail

This PR fixes the issue by sorting acronyms by length in descending order before building the regex pattern, ensuring longest matches take precedence over shorter overlapping ones.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/inflector/inflections.rb
+++ b/activesupport/lib/active_support/inflector/inflections.rb
@@ -264,9 +264,13 @@ module ActiveSupport
 
       private
         def define_acronym_regex_patterns
-          @acronym_regex             = @acronyms.empty? ? /(?=a)b/ : /#{@acronyms.values.join("|")}/
+          @acronym_regex             = @acronyms.empty? ? /(?=a)b/ : /#{sorted_acronyms.values.join("|")}/
           @acronyms_camelize_regex   = /^(?:#{@acronym_regex}(?=\b|[A-Z_])|\w)/
           @acronyms_underscore_regex = /(?:(?<=([A-Za-z\d]))|\b)(#{@acronym_regex})(?=\b|[^a-z])/
+        end
+
+        def sorted_acronyms
+          @acronyms.sort_by { |_key, acronym| -acronym.length }.to_h
         end
     end
 
--- a/activesupport/test/inflector_test.rb
+++ b/activesupport/test/inflector_test.rb
@@ -265,6 +265,16 @@ class InflectorTest < ActiveSupport::TestCase
     assert_equal("htmlAPI", ActiveSupport::Inflector.camelize("HTMLAPI", false))
   end
 
+  def test_underscore_acronym_precedence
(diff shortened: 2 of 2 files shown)


zubin:
Implemented in #56679.

PR #33773, opened 2018-09-01
Make null_store the default cache store in test environment config

mcnelson:
### Summary

It's surprising to me that by default, [low-level caching](https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching) is functional in the test environment. This simply adds `config.cache_store = :null_store` in `config/environments/test.rb`.

The existing defaults in `test.rb` as well as the Rails Guides kinda have you believe the `null_store` is the default.

Really was on the fence about opening this as it may not be a substantial enough change to warrant dev resources. But since Rails has such a great reputation of providing sane defaults, I felt like this would be worth it.

### Other Information

I perused the tests and wasn't sure where best to add anything. `railties/test/application/generators_test.rb`?

Appropriate to update CHANGELOG?

PR #56183, opened 2025-11-17
Document how to enable Rate Limiting in tests

Kerrick:
### Motivation / Background

This Pull Request has been created because I had trouble finding this information.

Rails applications are generated to use the `ActiveSupport::Cache::NullStore` in the test environment by default. This means `rate_limit` doesn't work without intervention. This was a surprising foot gun when I tried covering the controllers generated by the Rails 8 authentication generator.

I brought this up at https://github.com/rails/rails/pull/33773#issuecomment-3538502849 and @zzak requested I send a PR.

### Detail

This Pull Request adds a paragraph to the documentation of [`rate_limit` method in `ActionController::RateLimiting::ClassMethods`](https://api.rubyonrails.org/classes/ActionController/RateLimiting/ClassMethods.html#method-i-rate_limit). I placed it immediately after the existing paragraph about backing cache stores. I added it as a separate paragraph because it would've been quite long together, and it is a separate (though related) topic.

### DIFF
--- a/actionpack/lib/action_controller/metal/rate_limiting.rb
+++ b/actionpack/lib/action_controller/metal/rate_limiting.rb
@@ -77,6 +77,10 @@ module ActionController # :nodoc:
       #       rate_limit to: 3, within: 2.seconds, name: "short-term"
       #       rate_limit to: 10, within: 5.minutes, name: "long-term"
       #     end
+      #
+      # For testing with rate limits, you may need to switch the cache store to
+      # ActiveSupport::Cache::MemoryStore (`config.cache_store = :memory_store`)
+      # or any store that will persist data.
       def rate_limit(to:, within:, by: -> { request.remote_ip }, with: -> { raise TooManyRequests }, store: cache_store, name: nil, scope: nil, **options)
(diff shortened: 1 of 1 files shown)


actionpack/lib/action_controller/metal/rate_limiting.rb
```diff
@@ -38,6 +38,12 @@ module ClassMethods
       # datastore as your general caches, you can pass a custom store in the `store`
       # parameter.
       #
+      # Rate limiting does not work in automated tests when your backing cache store is
+      # a `ActiveSupport::Cache::NullStore`. The global `config.cache_store` uses this
+      # for the test environment by default. To enable rate limiting in your tests, edit
+      # `config/environments/test.rb` and change the value of `config.cache_store` from
+      # `:null_store` to `:memory_store`.
+      #
```
zzak: I would make this a footnote at the end of the documentation, after the example, and just say something like:

If your tests depend on rate limit, then you may need to switch the cache store to use ActiveSupport::Cache::MemoryStore or any store that will persist data.

I don't recommend changing it for all tests, as it could unnecessarily slow things down.

zzak:
Thank you!

PR #53687, opened 2024-11-20
Rails Guides: PostgreSQL timestamp with time zone [ci skip]

andyatkinson:
This addition to the Active Record PostgreSQL Rails Guide documents how to use a configuration option to change the default timestamp data type from `timestamp without time zone` to `timestamp with time zone` (timezone-aware timestamps).

Support for this data type was added to Active Record earlier but is not used by default, requiring a configuration change to use. From the PR comments where it was added, there was a suggestion to document how to use this configuration option, prompting this PR.

I created a new Rails 8 application, generating migrations for scaffolds, and realized they weren't using this data type. I made this configuration change, destroyed and regenerated my scaffold migration, and verified the new data type was used.

I welcome your feedback on this guide content, placement within the guide, or anything else.

- Related PR: https://github.com/rails/rails/pull/41084
- Guide page: https://guides.rubyonrails.org/active_record_postgresql.html
- PostgreSQL best practices:
https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29

Thanks!

### Checklist

Before submitting the PR make sure the following are checked:

* [ ] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [ ] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #41084, opened 2021-01-12
Active Record + PostgreSQL: native support for `timestamp with time zone`

ghiculescu:
In https://github.com/rails/rails/issues/21126 it was suggested to make "timestamp with time zone" the default type for datetime columns in PostgreSQL. This is in line with PostgreSQL [best practices](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29). This PR lays some groundwork for that.

This PR adds a configuration option, `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.datetime_type`. The default is `:timestamp` which preserves current Rails behavior of using "timestamp without time zone" when you do `t.datetime` in a migration. If you change it to `:timestamptz`, you'll get "timestamp with time zone" columns instead.

If you change this setting in an existing app, you should immediately call `bin/rails db:migrate` to ensure your `schema.rb` file remains correct. If you do so, then existing columns will not be impacted, so for example if you have an app with a mixture of both types of columns, and you change the config, schema dumps will continue to output the correct types.

This PR also adds two new types that can be used in migrations: `t.timestamp` and `t.timestamptz`.

```ruby
ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.datetime_type = :timestamp # default value is :timestamp

create_table("foo1") do |t|
  t.datetime :default_format # "timestamp without time zone"
  t.timestamp :without_time_zone # "timestamp without time zone"
  t.timestamptz :with_time_zone # "timestamp with time zone"
end

ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.datetime_type = :timestamptz

create_table("foo2") do |t|
  t.datetime :default_format # "timestamp with time zone" <-- note how this has changed!
  t.timestamp :without_time_zone # "timestamp without time zone"
  t.timestamptz :with_time_zone # "timestamp with time zone"
end

ActiveRecord::ConnectionAdapters::PostgreSQLAdapter::NATIVE_DATABASE_TYPES[:my_custom_type] = { name: "custom_datetime_format_i_invented" }
ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.datetime_type = :my_custom_type

create_table("foo3") do |t|
  t.datetime :default_format # "custom_datetime_format_i_invented"
  t.timestamp :without_time_zone # "timestamp without time zone"
  t.timestamptz :with_time_zone # "timestamp with time zone"
end
```

**Notes**

[...]

PR #44601, opened 2022-03-02
Add `timestamptz` as a time zone aware type for PostgreSQL

ghiculescu:
https://github.com/rails/rails/pull/41395 added support for the `timestamptz` type on the Postgres adapter.

As we found [here](https://github.com/rails/rails/pull/41084#issuecomment-1056430921) this causes issues because in some scenarios the new type is not considered a time zone aware attribute, meaning values of this type in the DB are presented as a `Time`, not an `ActiveSupport::TimeWithZone`.

This PR fixes that by ensuring that `timestamptz` is always a time zone aware type, for Postgres users. A workaround is to do this in an initializer:

```ruby
ActiveRecord::Base.time_zone_aware_types << :timestamptz
```

Or if you don't want this you can opt out in an initializer:

```ruby
ActiveRecord::Base.time_zone_aware_types -= [:timestamptz]
```

I think this should be backported to `7-0-stable`.

Fixes https://github.com/rails/rails/issues/45270

PR #56184, opened 2025-11-17
[skip ci] DOCS / PostgreSQL / Add missing configuration for timestamptz 

chaadow:
More info: https://github.com/rails/rails/pull/41084#issuecomment-3541856880

By setting `timestamptz`, `timestamp` is no longer time zone aware which breaks existing application that relied on this behavior. `timestamp` used to be converted to `datetime` which is time zone aware.

Related to : 
- https://github.com/rails/rails/pull/53687
- https://github.com/rails/rails/pull/41084

cc @ghiculescu @andyatkinson

### DIFF
--- a/guides/source/active_record_postgresql.md
+++ b/guides/source/active_record_postgresql.md
@@ -539,6 +539,8 @@ To configure `timestamp with time zone` as your new timestamp default data type,
 # config/application.rb
 ActiveSupport.on_load(:active_record_postgresqladapter) do
   self.datetime_type = :timestamptz
+  # If you are adding this to an existing application with an existing database, add this line:
+  ActiveRecord::Base.time_zone_aware_types << :timestamp
 end
 ```
 


ghiculescu:
Hmm. If this is actually an issue, I think we should fix it in `time_zone_aware_types`. Why not do it [here](https://github.com/rails/rails/blob/6388aec724df438bb7ed6909a710fcbd88bd45a3/activerecord/lib/active_record/railtie.rb#L95)? (see https://github.com/rails/rails/pull/44601)

p8:
https://github.com/rails/rails/pull/55135 will remove the PostgreSQL specific guide.

chaadow:
> Hmm. If this is actually an issue, I think we should fix it in `time_zone_aware_types`. Why not do it [here](https://github.com/rails/rails/blob/6388aec724df438bb7ed6909a710fcbd88bd45a3/activerecord/lib/active_record/railtie.rb#L95)? (see #44601)

@ghiculescu because technically `timestamp` is not timezone aware. it's just that when you manually set ` self.datetime_type = :timestamptz`, everything falls apart, because existing columns that are "TIMESTAMP WITHOUT TIME ZONE` are considered not time zone aware. (which is technically true in the database)

but as i said, everybody relies on the fact that Rails applies the proper timezone ( because of this  : https://github.com/rails/rails/blob/77191f5afb9c39aa5968219047ae69cdb2f48b14/activerecord/lib/active_record/connection_adapters/postgresql/oid/date_time.rb#L29-L31 )

and setting it in the initiliazer as you suggest is technically not true, because once an application is fully migrated to use `timestampz` then it makes sense to distinguish between the two. meaning when an application developer decides for some reason to use `timestamp` they will want it to NOT be timezone aware.

In conclusion, what I"m suggesting in my PR is that developers need to add this line temporarily so that the migration becomes smooth. and once they're done they can remove this line!

PR #56171, opened 2025-11-15
Faster class delegator

byroot:
Fix: https://github.com/rails/rails/pull/56167

I opened https://github.com/ruby/delegate/pull/46 upstream to improve `ClassDelegator` but it has a bunch of backward compatibility concerns that limits how much it can be optimized.

For internal use cases, we can define our own private class delegator and optimize it much further.

```
== no arguments ==
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
       DelegateClass     1.014M i/100ms
                 Opt     3.273M i/100ms
Calculating -------------------------------------
       DelegateClass     12.322M (± 0.8%) i/s   (81.15 ns/i) -     61.838M in   5.018661s
                 Opt     56.866M (± 0.4%) i/s   (17.59 ns/i) -    284.716M in   5.006894s

Comparison:
       DelegateClass: 12322457.6 i/s
                 Opt: 56865721.7 i/s - 4.61x  faster

== many arguments ==
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
       DelegateClass   555.419k i/100ms
                 Opt     3.214M i/100ms
Calculating -------------------------------------
       DelegateClass      6.256M (± 0.6%) i/s  (159.85 ns/i) -     31.659M in   5.060874s
                 Opt     53.902M (± 0.0%) i/s   (18.55 ns/i) -    269.936M in   5.007857s

Comparison:
       DelegateClass:  6255815.2 i/s
                 Opt: 53902461.2 i/s - 8.62x  faster

== optional arguments ==
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
       DelegateClass   826.558k i/100ms
                 Opt     1.645M i/100ms
Calculating -------------------------------------
       DelegateClass      9.821M (± 0.8%) i/s  (101.82 ns/i) -     49.593M in   5.049895s
                 Opt     21.177M (± 0.8%) i/s   (47.22 ns/i) -    106.952M in   5.050788s

Comparison:
       DelegateClass:  9821329.7 i/s
                 Opt: 21176752.0 i/s - 2.16x  faster
```

Benchmark:

```ruby
require "delegate"
require "bundler/inline"

gemfile do
  gem "benchmark-ips"
  gem "rails", path: "."
end

[...]

PR #56185, opened 2025-11-17
Fix DelegateClass defining ... signatures

skipkayhil:
Ref #56171

Previously, the `signature` variable was only getting set in the first two branches, but not the third. This caused methods that _should_ have been defined with triple dot delegation to instead have no arguments defined.

This commit fixes the issue by pulling the `signature =` assignment out of the condition so that it always happens.

### DIFF
--- a/activesupport/lib/active_support/delegation.rb
+++ b/activesupport/lib/active_support/delegation.rb
@@ -226,11 +226,12 @@ module ActiveSupport
         delegated_methods.each do |method_name|
           if method_name.match?(/\A[a-zA-Z]\w+[!?]?\z/)
             parameters = superclass.instance_method(method_name).parameters
-            if parameters.empty?
-              signature = ""
+
+            signature = if parameters.empty?
+              ""
             elsif parameters.all? { |type, _name| type == :req || type == :keyreq || type == :block }
               anonymous = 0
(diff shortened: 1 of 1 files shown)


skipkayhil:
cc @chaadow can you confirm this fixes the issue?

chaadow:
yes it does!!! thank you so much! @skipkayhil

chaadow:
@skipkayhil If you don't mind, from the ambigous stack trace that i sent earlier, namely the line 8 with the Delegate inheritance, I would like to know, how did you figure it out. I'd say probably from the arguments mismatch.

Rather, my question is, was there a way for me, with that stack trace to figure it out myself :D

skipkayhil:
> @skipkayhil If you don't mind, from the ambigous stack trace that i sent earlier, namely the line 8 with the Delegate inheritance, I would like to know, how did you figure it out. I'd say probably from the arguments mismatch.
> 
> Rather, my question is, was there a way for me, with that stack trace to figure it out myself :D

Yeah the arguments mismatch pointed me there (since I knew the PR introduced new meta-programming). I added `binding.irb if method_name == :try` inside the `method_name.match?` condition and then stepped through the `parameters` conditional, eventually seeing that `arguments` was `nil` even though the `else` block was taken.

PR #56195, opened 2025-11-18
Allow create_join_table to accept a primary key

gsamokovarov:
### Motivation / Background

Historically, `create_join_table` did not allow setting primary keys because it hard-coded id: false when calling the underlying `create_table` method. This meant that even if you passed the `:primary_key` option, it would be ignored since `id: false` took precedence and prevented any primary key from being added.

Postgres logical replication requires all tables to have primary keys for it to work correctly, and in this case, it's handy to be able to specify a primary key, such as a composite one.

At work (Dext), we are using Postgres replication, and we cannot use `create_join_table` in its current form. We have to rely on `create_table` for join tables, or define a custom migration method and deal with its reversibility, which is not pretty to say the least.

### Detail

The `id: false` option is no longer hard-coded and passed to the underlying `create_table` call, but it remains false by default to maintain backwards compatibility. When the `:primary_key`
option is provided, `:id` is automatically set to `:primary_key`, allowing primary keys to be configured. 

Here is a usage example:

```ruby
create_join_table :assemblies, :parts, primary_key: [:assembly_id, :part_id]
```

I could have left the interface to be:

```ruby
create_join_table :assemblies, :parts, id: :primary_key, primary_key: [:assembly_id, :part_id]
```

But the former is more concise, which is what we want when setting primary keys on the join tables, in my opinion.
    
### Additional information

While working on this PR, I stumbled on the `build_create_tjoin_table_definition` method, which is a `:nodoc` method invoked only in Rails' test suite. Do you think I can just remove it? It's not public by Rails API standards, and a simple [GH search](https://github.com/search?q=build_create_join_table_definition&type=code&p=1) does not show it being used in anything serious – type declaration projects, Rails forks, and the likes.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Allow `create_join_table` to accept a primary key.
+
+    This is useful for databases like PostgreSQL where logical replication requires primary
+    keys on all tables.
+
+    ```ruby
+    create_join_table :assemblies, :parts, primary_key: [:assembly_id, :part_id]
+    ```
+
+    This generates a join table with a composite primary key on both foreign key columns.
+
+    *Genadi Samokovarov*
+
 *   Fix Active Record Pool Reaper thread leak after `Parallelization#shutdown`.
 
     After parallelized test runs, the parent process leaked the Active Record Pool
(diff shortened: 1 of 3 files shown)


fatkodima:
Do the following syntax work currently?
```ruby
create_join_table ... do |t|
  t.primary_key ...
  ...
end
```

gsamokovarov:
Nope, `t.primary_key` creates a new column, and if you try to make a composite primary key, you get a funny-named column that is indeed the primary key. For example:

```
"[:supplier_id, :receipt_id]" bigint NOT NULL
```

To set a composite primary key, which is probably what you want for the join tables in the replication context, you can get around by calling `set_primary_key`, which works correctly.
```
create_join_table ... do |t|
  t.set_primary_key ...
  ...
end
```

This is what I used for our custom migration method. The problem with custom migration methods is that making them reversible is neither trivial nor pretty.

And using the `create_join_table` helper with a block to specify the primary key kinda invalidates its ergonomics.

PR #56197, opened 2025-11-19
Remove unused test fixtures from ActionView

thiagopradi:
While reviewing the ActionView test suite, I discovered a couple of fixture files that are not referenced anywhere in the tests. 

### Detail

This Pull Request removes two unused file from the ActionView test suite.

### DIFF
--- a/actionview/test/fixtures/test/dont_pick_me
+++ b/actionview/test/fixtures/test/dont_pick_me
@@ -1,1 +1,0 @@
-non-template file
--- a/actionview/test/fixtures/test/hello_world.erb~
+++ b/actionview/test/fixtures/test/hello_world.erb~
@@ -1,1 +1,0 @@
-Don't pick me!


skipkayhil:
`dont_pick_me` was added in https://github.com/rails/rails/commit/2fc6c7dd05b9481ed81fe19c8fc0cc7915868404 🤔 unclear how it was used at the time

[Lighthouse archive](https://web.archive.org/web/20220526081339/https://rails.lighthouseapp.com/projects/8994/tickets/1187)

thiagopradi:
@skipkayhil yeah, I couldn't find any mentions or usage on a test, so I assumed it was added by mistake.

skipkayhil:
The other one was added in https://github.com/rails/rails/commit/e4a7c0bb5b175e6eceb2fe808075d6122521237c [Lighthouse](web.archive.org/web/20220526081339/https://rails.lighthouseapp.com/projects/8994/tickets/1818)

They were also intentionally kept when extracting `actionview` from `actionpack`: https://github.com/rails/rails/pull/11032#issuecomment-19778195

I assume these were added because they would make some test fail, but at this point it may be difficult to reproduce the Rails 2 test suite to validate that assumption.

PR #56198, opened 2025-11-19
Make explain accept hash format syntax

gmcgibbon:
### Motivation / Background

This Pull Request has been created because Active Record currently requires adapter specific syntax for explain queries when using the FORMAT option. MySQL uses `FORMAT=JSON` whereas Postgres uses `FORMAT JSON`. We should be able to accept a hash that maps to the right format syntax as an option.

### Detail

This Pull Request changes explain clause building for mysql and postgres to map hashes to the correct syntax. Example:
```rb
Car.all.explain(format: :json)
# or
Car.connection.explain(query, [], [{format: :json}]
```

### Additional information

I previously tried gsub-ing the SQL string but I think this approach with hashes is cleaner.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Allow either explain format syntax for EXPLAIN queries.
+
+    MySQL uses FORMAT=JSON whereas Postgres uses FORMAT JSON. We should be
+    able to accept both formats as options.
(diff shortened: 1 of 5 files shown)


gmcgibbon:
I think I actually like the hash syntax better. I'll update the PR to use that instead of mangling SQL.

PR #56202, opened 2025-11-20
Prepping rails for minitest 6

zenspider:
This branch contains all the fixes needed for minitest 6 to work smoothly with rails. I've now tested extensively on MT5 as well as MT6. 

I have another branch testing out the actual MT6 bump and it still has some sort of flaky going on that I'm investigating.

### DIFF
--- a/actionpack/test/dispatch/routing/route_set_test.rb
+++ b/actionpack/test/dispatch/routing/route_set_test.rb
@@ -20,13 +20,13 @@ module ActionDispatch
       end
 
       test "not being empty when route is added" do
-        assert_predicate self, :empty?
+        assert_empty @set
 
         draw do
           get "foo", to: SimpleApp.new("foo#index")
         end
 
-        assert_not empty?
+        assert_not_empty @set
       end
 
       test "URL helpers are added when route is added" do
--- a/activesupport/lib/active_support/testing/assertions.rb
+++ b/activesupport/lib/active_support/testing/assertions.rb
@@ -124,7 +124,8 @@ module ActiveSupport
           actual = exp.call
           rich_message = -> do
             code_string = code.respond_to?(:call) ? _callable_to_source_string(code) : code
-            error = "`#{code_string}` didn't change by #{diff}, but by #{actual - before_value}"
+            error = "`#{code_string}` didn't change by #{diff}, but by #{actual - before_value}."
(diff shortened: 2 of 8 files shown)


activesupport/test/backtrace_cleaner_test.rb
```diff
@@ -2,6 +2,8 @@
 
 require_relative "abstract_unit"
 
+module Rails; end # needed to run this file solo?
```
guilleiguaran: Added a missing require in 8aebd0b507

actionpack/test/dispatch/routing/route_set_test.rb
```diff
@@ -20,13 +20,13 @@ def call(env)
       end
 
       test "not being empty when route is added" do
-        assert_predicate self, :empty?
+        assert_empty @set
```
tenderlove, line 23: Why do these need to be changed?

activesupport/lib/active_support/testing/assertions.rb
```diff
@@ -296,8 +297,9 @@ def assert_no_changes(expression, message = nil, from: UNTRACKED, &block)
 
         rich_message = -> do
           code_string = expression.respond_to?(:call) ? _callable_to_source_string(expression) : expression
-          error = "`#{code_string}` changed"
+          error = "`#{code_string}` changed."
```
tenderlove, line 300: Are these periods load bearing?  Why do they need to be added?

tenderlove (commented):
The changes seem fine, but I have a couple questions.  If some of these changes are necessary, I would like to understand why.  If they aren't necessary, then maybe remove them from the PR? (I don't actually mind whether or not those changes are there, I just want to know if they are there _because_ of the upgrade to MT6)

actionpack/test/dispatch/routing/route_set_test.rb
zenspider: because y'all made `empty?` private and `assert_predicate` does a send on self. Breaks. `assert_empty` on `@set` passes.

activesupport/lib/active_support/testing/assertions.rb
zenspider: The `message` helper has gotten a lot simpler/cleaner for MT6 and doesn't end sentences for you anymore. changing the error message to include the period passes on both MT5 and MT6. Otherwise I have to fight versioning on the assert_equal against the message output.

tenderlove (approved):

PR #56204, opened 2025-11-20
Add support for configuring migration strategy on a per-adapter basis.

adrianna-chang-shopify:
### Motivation / Background

This allows individual databases to customize migration execution logic, rather than only allowing overrides at the global level. This is useful for multi-db applications that may be using different DB drivers, with different requirements for migration behaviour.

For example, to use a custom migration strategy for PostgreSQL:

```ruby
class CustomPostgresStrategy < ActiveRecord::Migration::DefaultStrategy
  def drop_table(*)
    # Custom logic specific to PostgreSQL
  end
end

ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.migration_strategy = CustomPostgresStrategy
```

### Detail

Add a `class_attribute` for `migration_strategy` to `AbstractAdapter`. When we set the `execution_strategy` inside the migration, we first check whether the adapter has a strategy configured, otherwise we use the global value.

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -1188,6 +1188,23 @@ end
 config.active_record.migration_strategy = CustomMigrationStrategy
 ```
 
+You can also configure migration strategies on a per-adapter basis by setting the `migration_strategy` class on the adapter itself.
+This is useful when you want to customize migration behavior for a specific database type.
+
+For example, to use a custom migration strategy for PostgreSQL:
+
+```ruby
+class CustomPostgresStrategy < ActiveRecord::Migration::DefaultStrategy
+  def drop_table(*)
+    # Custom logic specific to PostgreSQL
+  end
+end
+
+ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.migration_strategy = CustomPostgresStrategy
+```
+
+By assigning to the adapter class, all migrations run through connections using that adapter will use the specified strategy.
+
 #### `config.active_record.schema_versions_formatter`
 
 Controls the formatter class used by schema dumper to format versions information. Custom class can be provided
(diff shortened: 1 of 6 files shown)


eileencodes (approved):

guides/source/configuring.md
```diff
@@ -1188,6 +1188,23 @@ end
...
+For example, to use a custom migration strategy for PostgreSQL:
+
+```ruby
+class CustomPostgresStrategy < ActiveRecord::Migration::DefaultStrategy
+  def drop_table(*)
+    # Custom logic specific to PostgreSQL
+  end
+end
+
+ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.migration_strategy = CustomPostgresStrategy
```
chaadow, line 1203: Hi! maybe it's worth mentionning that we can use this in an initializer
```ruby
ActiveSupport.on_load(:active_record_postgresqladapter) do
  self.migration_strategy = CustomPostgresStrategy
end
```

jrochkind:
Saw this PR listed in "This Week In Rails" -- thanks!

I was not aware there was a feature called "migration strategy" that was global and is now being made per-adapter!  But I am interested in it. I have not been able to find anything more about trying to search documentation, github issues/PRs, or blogs. (Could be I'm bad at searching). 

Please don't clutter up the comments with discussion, but if anyone were able to simply post links to more info on this feature and how to use it, I and possibly others coming from This Week In Rails would appreciate it, thanks!

adrianna-chang-shopify:
Hey @jrochkind ! We added this a little while back to support our needs at Shopify, we did _not_ do a great job of publicizing it though (clearly) 😂 I'll make some time to write something for Shopify's [Rails at Scale](https://railsatscale.com/) blog in the next few weeks, glad to hear that kind of thing might be helpful!

guides/source/configuring.md
zzak: I think this guide is already really long, and if we try to explain all the ways to set config for each option it will grow even longer.

PR #56208, opened 2025-11-21
Fix bug when txn isolation would not reset or when requires_new is set

kirs:
Opening this since it's a smaller change that's perhaps easier to land than https://github.com/rails/rails/pull/55792.

In https://github.com/rails/rails/pull/55549, I've missed:

1) Resetting `current_transaction.isolation` back to its original value
2) Handling `requires_new: true`

This is a trivial fix and this bug is really impacting our isolation workflows at Shopify.

@matthewd @eileencodes @byroot


gmcgibbon (approved):
LGTM, but the previous PR is released in 8.1, so we need a changelog describing the bug fix. Please add one and this can be merged.

PR #53970, opened 2024-12-15
Deprecate `ActiveSupport::Configurable`

seanpdoyle:
### Motivation / Background

This proposal is based on a [code review comment][]:

> I think we should deprecate and remove this module. It is only used
> once inside the framework, in a place that perhaps don't even need it
> anymore and I don't think it adds much for our users. They would be
> better served with their own config object that is just a hash.

### Detail

To achieve this outcome, replace the only internal occurrence of `ActiveSupport::Configurable` with a [class_attribute][]. Similarly, replace [config_accessor][] calls with class- and instance-level calls to [delegate][] for the readers and writers.

[code review comment]: https://github.com/rails/rails/pull/53796#issuecomment-2512417545
[class_attribute]: https://edgeapi.rubyonrails.org/classes/Class.html#method-i-class_attribute
[config_accessor]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Configurable/ClassMethods.html#method-i-config_accessor
[delegate]: https://edgeapi.rubyonrails.org/classes/Module.html#method-i-delegate

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56209, opened 2025-11-21
Update Configurable deprecaton warning

skipkayhil:
Ref https://github.com/rails/rails/pull/53970

The blanket suggestion to use class_attribute can be misleading, and I've seen gems default to using class_attribute because of this warning instead of plain attr_accessors.

This commit tries to be less prescriptive of how to replace Configurable by suggesting multiple alternatives, since different usages of Configurable will require different replacements.


skipkayhil:
cc @janko Would appreciate your thoughts since you expressed concern in the referenced pull request

PR #53340, opened 2024-10-16
[ci skip] Fix ENV check for solid_queue plugin

Jay0921:
### Motivation / Background

This PR updates the ENV check for the solid_queue plugin to ensure it is only activated when `ENV["SOLID_QUEUE_IN_PUMA"]` is explicitly set to `true`.

The current check activates solid_queue as long as there is any value in `ENV["SOLID_QUEUE_IN_PUMA"]`, even if the value is `false`.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [ ] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56210, opened 2025-11-21
Clarify that SOLID_QUEUE_IN_PUMA needs to be removed to run Solid Queue separately

TheRusskiy:
### Motivation / Background

This Pull Request has been created to clarify the behaviour of `SOLID_QUEUE_IN_PUMA` variable.

### Detail

Setting it to `false` does nothing because of how we check the variable:
https://github.com/Jay0921/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/puma.rb.tt#L39
`"false"`  is still truthy.

My first though was to update the check within Puma,
but I see that a similar PR was closed: https://github.com/rails/rails/pull/53340
so going to just document it, hopefully saves other people time.

### Checklist


rafaelfranca:
Gone with #57424

Issue #56112, opened 2025-11-08
ActiveSupport::TimeWithZone#xmlschema broken when used with utc DateTime


rahearn:
### Steps to reproduce

```
tz = ActiveSupport::TimeZone["America/New_York"]
twz = ActiveSupport::TimeWithZone.new(nil, tz, DateTime.new(2025, 11, 7, 12))
puts twz.xmlschema
```

the script will output `2025-11-07T12:00:00+00:0-05:00`

### Expected behavior

The output will be in proper iso8601 format.

### Actual behavior

The output contains an extra `+00:0` between the time section and the actual timezone offset.

This issue was introduced in https://github.com/rails/rails/pull/55583. That PR assumes that the local `time.xmlschema` will always end in `Z`, which is not true for `DateTime` objects

### System configuration

Rails 8.1.1
Ruby 3.4.7

daniel-niknam:
Here is the test to confirm the behaviour:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"
end

require "active_support"
require "active_support/core_ext/object/blank"
require "minitest/autorun"
Minitest::Test.make_my_diffs_pretty!

class BugTest < ActiveSupport::TestCase
  def test_failed
    tz = ActiveSupport::TimeZone["America/New_York"]
    twz = ActiveSupport::TimeWithZone.new(nil, tz, DateTime.new(2025, 11, 7, 12))

    assert_equal "2025-11-07T12:00:00-05:00", twz.xmlschema
  end
end
```

output:

```bash
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Run options: --seed 6848

# Running:

F

Finished in 0.029187s, 34.2618 runs/s, 34.2618 assertions/s.

  1) Failure:
BugTest#test_failed [56112.rb:23]:
--- expected
+++ actual
@@ -1 +1 @@
-"2025-11-07T12:00:00-05:00"
+"2025-11-07T12:00:00+00:0-05:00"


1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

daniel-niknam:
[...]

PR #56216, opened 2025-11-22
Fix TimeWithZone#xmlschema when wrapping DateTime local time

drymar:
### Motivation / Background

This Pull Request has been created because of issue: https://github.com/rails/rails/issues/56112

It fixes a regression where `ActiveSupport::TimeWithZone#xmlschema` produced invalid ISO-8601 strings when instantiated with a `DateTime` used as a local time.

Example:
```ruby
tz  = ActiveSupport::TimeZone["America/New_York"]
twz = ActiveSupport::TimeWithZone.new(nil, tz, DateTime.new(2025, 11, 7, 12))

twz.xmlschema
# => "2025-11-07T12:00:00+00:0-05:00"
```

Expected:
```ruby
"2025-11-07T12:00:00-05:00"
```

This happens because the implementation assumed that `time.iso8601` always ends with `"Z"`, which is not true for `DateTime`. This PR ensures correct ISO-8601 output regardless of whether the wrapped value is a `Time` or `DateTime`.

### Detail

This Pull Request changes:
- Updates `ActiveSupport::TimeWithZone#xmlschema` to replace the entire timezone suffix (`Z` or `±HH:MM`) instead of the final character
- Ensures correct calculation of the timezone offset for both Time and DateTime instances
- Adds a regression `test test_xmlschema_with_datetime_local_time`, which reproduces the issue and confirms the fix

Implementation change:
```ruby
str.sub(/(Z|[+-]\d{2}:\d{2})\z/, formatted_offset(true, "Z"))
```

### Additional information

It does not introduce behavioral changes outside of the reported bug

### Checklist

* [x] This Pull Request is related to a single change.
* [x] Commit message contains a detailed description and references the issue.
* [x] Tests were added and pass successfully.
* [x] No CHANGELOG is needed, this is a minor bug fix.

### DIFF
--- a/activesupport/lib/active_support/time_with_zone.rb
+++ b/activesupport/lib/active_support/time_with_zone.rb
@@ -152,11 +152,15 @@ module ActiveSupport
     #
     #   Time.zone.now.xmlschema  # => "2014-12-04T11:02:37-05:00"
     def xmlschema(fraction_digits = 0)
+      precision = fraction_digits || 0
+
       if @is_utc
-        utc.iso8601(fraction_digits || 0)
+        utc.iso8601(precision)
       else
-        str = time.iso8601(fraction_digits || 0)
-        str[-1] = formatted_offset(true, "Z")
+        str = time.iso8601(precision)
+        offset = formatted_offset(true, "Z")
+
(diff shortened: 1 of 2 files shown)


byroot:
Thanks, backported to `8-1-stable`.

PR #56221, opened 2025-11-24
fix(activerecord): support backticked table names in MariaDB with cus…

### LINKED ISSUE #56200: Inserting with backticks into MariaDB table with custom PK failing
### Steps to reproduce

Inserting into some MariaDB table written in backticks and with custom primary key fails when the inserted values match the regular expression in ActiveRecord::ConnectionAdapters::DatabaseStatements#extract_table_ref_from_insert_sql, producing something that can't be resolved by "primary_key(table_ref)" in some cases.

```ruby
require 'test_helper'

class DatabaseStatementsTest < ActiveSupport::TestCase
  def test_query_with_backticks_without_pk
    # create_table "my_texts", primary_key: "pk_column", id: { type: :string, limit: 13 }, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
    #   t.text "text", size: :medium
    #   t.datetime "created_at", null: false
    #   t.datetime "updated_at", null: false
    # end

    sql_prefix = "INSERT INTO `my_texts` (`pk_column`,`text`,`created_at`,`updated_at`) VALUES "
    sql_suffix =
      ' ON DUPLICATE KEY UPDATE `my_texts`.`text`=VALUES(`text`),`my_texts`.`created_at`=VALUES(`created_at`),' \
        '`my_texts`.`updated_at`=VALUES(`updated_at`)'
    batch_values_trials =
      [
        ["('4005556237319','Inserting \"into \" should not fail','2025-11-19 22:23:40.918912','2025-11-19 22:23:40.918912')"],
        ["('4005556237319','Inserting \"into .\" should not fail','2025-11-19 22:23:40.918912','2025-11-19 22:23:40.918912')"]
      ]

    batch_values_trials.each do |batch_values|
      assert_nothing_raised do
        ActiveRecord::Base.connection.insert_many(
          [sql_prefix, sql_suffix],
          batch_values,
          {
            model: MyText, validate: false, timestamps: true, track_validation_failures: false,
            on_duplicate_key_update: [:text, :created_at, :updated_at], primary_key: 'pk_column'
          },
          "MyText Create Many"
        )
      end
    end
  end
end
```

### Expected behavior

Should successfully insert the defined values into the database table.

### Actual behavior

Depending on the inserted values, raises ArgumentException or "Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'ORDER BY seq_in_index".

The regular expression in ActiveRecord::ConnectionAdapters::DatabaseStatements#extract_table_ref_from_insert_sql doesn't evaluate backticks for table names. When it matches the inserted values instead, the table name will be extracted from these, producing table name like empty string or ".". With that, "primary_key(table_ref)" will raise several kinds of exception depending on the irregularly found table name.

### System configuration

**Rails version**: 
8.0.3

**Ruby version**: 
3.4.7

**MariaDB version**:
 10.11.13

harsha08-2k6:
…tom primary keys - Fixes #56200

Updated the extract_table_ref_from_insert_sql method to properly handle backticked table identifiers used in MariaDB.

The regex pattern now matches three identifier formats:
1. Backticked identifiers: `table_name`
2. Double-quoted identifiers: "table_name"
3. Unquoted identifiers: table_name

Also updated the string processing to delete backticks in addition to double quotes when extracting table references.

This resolves the issue where INSERT statements with backticked MariaDB table names and custom primary keys were failing due to improper table reference extraction.

### Motivation / Background

### Detail

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -817,10 +817,8 @@ module ActiveRecord
         end
 
         def extract_table_ref_from_insert_sql(sql)
-          if sql =~ /into\s("[A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
-            $1.delete('"').strip
-          end
-        end
+    if sql =~ /into\s(`[A-Za-z0-9_.`]+`|"[A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_.\[\]]+)\s*/im            $1.delete('"').strip
+      $1.delete('"` ').strip        end
     end
   end
 end
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -817,8 +817,8 @@ module ActiveRecord
         end
 
         def extract_table_ref_from_insert_sql(sql)
-    if sql =~ /into\s(`[A-Za-z0-9_.`]+`|"[A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_.\[\]]+)\s*/im            $1.delete('"').strip
-      $1.delete('"` ').strip        end
+            if sql =~ /into\s((?:`[^`]+`|"[^"]*"|\S+))\s*/im
+            $1.delete("`\"").strip
     end
   end
 end


activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -817,10 +817,8 @@ def arel_from_relation(relation)
         end
 
         def extract_table_ref_from_insert_sql(sql)
-          if sql =~ /into\s("[A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
-            $1.delete('"').strip
-          end
-        end
+            if sql =~ /into\s((?:`[^`]+`|"[^"]*"|\S+))\s*/im
+            $1.delete("`\"").strip
```
shivabhusal, line 821: Thanks for the PR.

Please update the PR title/description.
Also, please add tests for the new behavior, and include a **`CHANGELOG`** entry if this affects user-facing behavior. A short explanation of any non-obvious changes would also be helpful.

PR #56222, opened 2025-11-24
Fix remove_index to handle custom name with JSONB expression

alexesba:
### Motivation / Background

When reversing a migration that adds an index with a JSONB expression and a custom name (using `algorithm: :concurrently`), Rails fails to remove the index. The column-based matching logic generates a different index name than the custom one provided, resulting in `ArgumentError: No indexes found`. This PR addresses that bug.

### Detail

This Pull Request updates `index_name_for_remove` to skip column-based matching when both a custom name and a complex expression are provided, relying solely on name-based matching. This ensures that indexes created with custom names and expressions can be reliably removed.

### Additional information

- Backward compatible; only affects this edge case.
- No breaking changes.
- Affects Rails 6.1+.
- Includes a test that creates and removes such an index, verifying the fix.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## Rails 6.1.7.10 (October 23, 2024) ##
 
-*   No changes.
+*   Fix `remove_index` to handle custom name with JSONB expression.
+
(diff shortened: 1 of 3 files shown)


yahonda:
Thanks for the pull request.
We’re not accepting pull requests for unmaintained versions. https://rubyonrails.org/maintenance 
Would you try this on the Rails main branch and, if the issue persists, consider submitting a pull request against main?

PR #56227, opened 2025-11-25
PostgreSQLAdapter: fully reset the connection when reconnecting

byroot:
Ref: https://github.com/rails/rails/issues/51780#issuecomment-3571653983

Postgres connections require a lot of associated state in `@type_map` and other values. The configuration process is very long and contain many pieces.

If it's interupted by something like `Timeout.timeout`, all bets are off as to what state we're in, and we may end up in a finally broken state where the connection report being active but is unusable.

To ensure that even such broken connection won't remain in such state forever, this commit de-optimize `PostgreSQLAdapter#reconnect`.

We no longer try to reset and re-use the connection, we do a full reconnect and reconfigure.

Big thanks to Shayon Mukherjee for the repro.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -362,6 +362,14 @@ module ActiveRecord
       def active?
         @lock.synchronize do
           return false unless @raw_connection
+
+          # Defensive programing: some users may interupt the configuration
+          # process with Timeout.timeout and cause connections to be broken
+          # beyond repair.
+          # This cheap check tries to catch this case.
+          # Ref: https://github.com/rails/rails/issues/51780
+          return false unless @type_map && @type_map_for_results
+
           @raw_connection.query ";"
           verified!
         end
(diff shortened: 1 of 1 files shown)


byroot:
Damn it, that breaks a bunch of tests. I'll try to come back to this ASAP.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -962,13 +981,12 @@ def connect
...
-            @raw_connection = nil
-          end
-
-          connect unless @raw_connection
+          # We fully disconnect because PostgreSQLAdapter has lots of state such as `@type_map`
+          # that can easily go out of sync if the configuration process is interupted.
+          # So to ensure we'll be back to a clean state, we restart from scratch.
+          @raw_connection&.close rescue nil
+          unconfigure_connection
+          connect
```
shayonj, line 989: nice!

do you reckon https://github.com/byroot/rails/blob/c9758b26f135059c6ffdac666f17a20b348108f1/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1224 is still needed? (vs just `StandardError`)

From: https://github.com/rails/rails/pull/54738/

byroot:
Superseeded by: https://github.com/rails/rails/pull/56231

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
byroot: Yes, to handle timeouts: https://github.com/ruby/timeout/blob/983cbf636a173bbea516c4cc2354b970108bd137/lib/timeout.rb#L26

Issue #51217, opened 2024-02-28
`ActiveRecord::Persistence#becomes` does not clone eager-loading state and can raise Strict Loading violations


seanpdoyle:
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"

  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new($stdout)

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
    t.text :type
  end

  create_table :comments, force: true do |t|
    t.integer :post_id
  end
end

class Post < ActiveRecord::Base
  has_many :comments, strict_loading: true
end

class Post::Special < Post
end

class Comment < ActiveRecord::Base
  belongs_to :post
end

class BugTest < Minitest::Test
  def test_becomes_eager_loading
    comment = Comment.new

    Post.create! comments: [comment]
    post = Post.includes(:comments).last!

    assert_equal [comment], post.comments
    assert_equal [comment], post.becomes(Post::Special).comments
    # ActiveRecord::StrictLoadingViolationError: `Post::Special` is marked for strict_loading. The Comment association named `:comments` cannot be lazily loaded.
    #     activerecord/lib/active_record/core.rb:256:in 'ActiveRecord::Base.strict_loading_violation!'
    #     activerecord/lib/active_record/associations/association.rb:250:in 'ActiveRecord::Associations::Association#find_target'
    #     activerecord/lib/active_record/associations/collection_association.rb:274:in 'ActiveRecord::Associations::CollectionAssociation#load_target'
    #     activerecord/lib/active_record/associations/collection_proxy.rb:45:in 'ActiveRecord::Associations::CollectionProxy#load_target'
    #     activerecord/lib/active_record/associations/collection_proxy.rb:981:in 'ActiveRecord::Associations::CollectionProxy#=='
  end
end
```

### Expected behavior

[...]

PR #56228, opened 2025-11-25
Optimize ActiveRecord::Base#becomes when no class conversion needed

bogdan:
### Motivation / Background

This Pull Request has been created because `becomes` can be optimized a little.

### Detail

This Pull Request changes `ActiveRecord::Base#becomes` to return self when no class conversion needed.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/persistence.rb
+++ b/activerecord/lib/active_record/persistence.rb
@@ -485,6 +485,8 @@ module ActiveRecord
     #
     # If you want to change the STI column as well, use #becomes! instead.
     def becomes(klass)
+      return clone if self.class == klass
+
       became = klass.allocate
 
       became.send(:initialize) do |becoming|


activerecord/lib/active_record/persistence.rb
```diff
@@ -485,6 +485,8 @@ def destroy!
     #
     # If you want to change the STI column as well, use #becomes! instead.
     def becomes(klass)
+      return self if self.class == klass
+
```
shivabhusal, line 489: Nice addition!

I suggest adding a test to cover the “same class” case. Something like:

```ruby
 # activerecord/test/cases/persistence_test.rb:528

  def test_becomes_when_same_class_instance_is_passed
    client = LargeClient.new(name: "ShrinkingCompany")

    new_client = client.becomes(LargeClient)

    assert_equal client.object_id, new_client.object_id
  end
```

activerecord/lib/active_record/persistence.rb
```diff
@@ -485,6 +485,8 @@ def destroy!
     #
     # If you want to change the STI column as well, use #becomes! instead.
     def becomes(klass)
+      return self if self.class == klass
```
byroot:
```suggestion
      return clone if self.class == klass
```
I think we should still return a copy, because until now this method always did. Otherwise this optimization my introduce bugs in apps that depends on mutating the return of `becomes` without mutating the original value.

seanpdoyle:
This PR (inadvertently?) resolved https://github.com/rails/rails/issues/51217 in its original form.

I've adjusted the test script to be more realistic. After making those changes, it is once again failing.

@bogdan if you're interested, I would appreciate some help troubleshooting that behavior!

dennispaagman:
Just came across this when I had a similar idea.

Are you sure this is implemented as you wanted @bogdan? As your description mentions it returns `self`, but the actual code returns `clone`. The test you implemented in #56248 also tests against not having the same object_id, so it does seem deliberate.

If there's a good reason to clone instead of just returning self that's ok of course, I'm just wondering if there is and why we cannot just return self 🤔

bogdan:
@dennispaagman https://github.com/rails/rails/pull/56228#discussion_r2563856098

dennispaagman:
Oh yes thanks I guess I missed that, because it was collapsed! Thanks for the quick response 🙇‍♂️

PR #51951, opened 2024-05-30
`ActiveStorage` immediate variants

tomrossi7:
### Motivation / Background

ActiveStorage currently allows variants to be created on demand (by default) or in the background (using preprocessed). This can be problematic for applications that have a lot of traffic. Immediately after an attachment is created, thousands of requests could be created for a single variant. Each controller receiving the request as well as a background job could all be racing to create the variant.

In order to give more control variant creation, we are proposing an immediate option to allow variants to be created simultaneously with the attachment.

This Pull Request has been created because products like ours need more control over variant creation.

### Detail

This Pull Request changes introduces the `immediate` option and follows the existing patterns around `preprocessed`.

```
  has_one_attached :avatar_with_immediate do |attachable|
    attachable.variant :thumb, resize_to_limit: [4, 4], immediate: true
  end
```

### Checklist

Before submitting the PR make sure the following are checked:

* [X] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [X] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [X] Tests are added or updated if you fix a bug or add a feature.
* [X] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56233, opened 2025-11-25
Schedule a single job per Blob that schedules preprocessing variants jobs

p8:
Similar to `PreviewImageJob`, instead of scheduling jobs for processing each variant, schedule a single job that schedules jobs for processing each variant.

This makes things more performant as less jobs get scheduled in a single request. It also makes things more consistent, as for previewable blobs we do the same thing.

# Details 

### For attachments we schedule a job for each variant

If an ActiveStorage attachment has multiple preprocessed variants, currently a job gets created for each variant, resulting in a lot of jobs when multiple blobs are attached with multiple variants in a single request.
A request with 10 uploads with 5 variants each, schedules 50 jobs.

```mermaid
flowchart TD
    A[Attachment] -->|transform_variants_later| B@{shape: processes, label: TransformJob}
```

This is done in `transform_variants_later` where `blob.preprocessed` schedules a `TransformJob` for each variation:
https://github.com/rails/rails/blob/4c53863e354ad94dbfaf0643d9cdc99393a1a583/activestorage/app/models/active_storage/attachment.rb#L144-L146

### For previewable attachments we schedule a single job

For a previewable blob, for example a PDF has a image preview, all variants are based on the preview.
As the variants require the preview, we create a single `PreviewImageJob` that creates the preview and then schedules all the variants.

```mermaid
flowchart TD
    C[Attachment] -->|transform_variants_later| D[ImagePreviewJob]
    D -->|perform| E@{shape: processes, label: TransformJob}
```

`create_preview_image_later` schedules a `PreviewimageJob` for the blob:

https://github.com/rails/rails/blob/4c53863e354ad94dbfaf0643d9cdc99393a1a583/activestorage/app/models/active_storage/attachment.rb#L142

`PreviewImageJob` processes the required image preview and schedules jobs for all preprocessed variants:

https://github.com/rails/rails/blob/4c53863e354ad94dbfaf0643d9cdc99393a1a583/activestorage/app/jobs/active_storage/preview_image_job.rb#L10-L14

### Schedule a single job for all variant processing

For non-previewable blobs we can do the same: schedule a single job that schedules jobs for all preprocessed variants.

```mermaid
flowchart TD
    C[Attachment] -->|transform_variants_later| D[VariantsJob]
    D -->|perform| E@{shape: processes, label: TransformJob}
```

This moves some logic (with feature envy on the Blob) from Attachment to a single method on Blob.
This is similar to other methods defined on Attachment that delegate to the Blob instead, like `analyze_blob_later`, `mirror_blob_later` and `purge_later`.

This reuses some code from: https://github.com/rails/rails/pull/51951

Co-authored-by: Tom Rossi <tom@higherpixels.com>

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Schedule a single job per blob for creating variants
+
+    *Petrik de Heus, Tom Rossi <tom@higherpixels.com>*
+
 *   `ActiveStorage::Blob#open` can now be used without passing a block, like `Tempfile.open`. When using this form the
     returned temporary file must be unlinked manually.
 
--- a/activestorage/app/jobs/active_storage/variants_job.rb
+++ b/activestorage/app/jobs/active_storage/variants_job.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ActiveStorage::VariantsJob < ActiveStorage::BaseJob
+  queue_as { ActiveStorage.queues[:transform] }
+
+  discard_on ActiveRecord::RecordNotFound, ActiveStorage::UnrepresentableError
+  retry_on ActiveStorage::IntegrityError, attempts: 10, wait: :polynomially_longer
+
+  def perform(blob, variations)
+    variations.each do |transformations|
+      blob.preprocessed(transformations)
+    end
+  end
+end
--- a/activestorage/app/models/active_storage/attachment.rb
+++ b/activestorage/app/models/active_storage/attachment.rb
@@ -137,14 +137,7 @@ class ActiveStorage::Attachment < ActiveStorage::Record
           named_variant.transformations
         end
       }
-
-      if blob.preview_image_needed_before_processing_variants? && preprocessed_variations.any?
-        blob.create_preview_image_later(preprocessed_variations)
-      else
-        preprocessed_variations.each do |transformations|
-          blob.preprocessed(transformations)
-        end
-      end
+      blob.transform_variants_later(preprocessed_variations) if preprocessed_variations.any?
     end
 
     def purge_dependent_blob_later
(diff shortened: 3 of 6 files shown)


byroot:
> a job gets created for each variant, resulting in a lot of jobs

I've never been a big Active Storage user, but is this bad? It's generally good practice to enqueue multiple small jobs (within reason) over one big one, as if one of the job fail only that part will fail and/or be retried, instead of all of it.

So what's the issue exactly?

p8:
This single job just schedules multi small jobs, 1 for each variant.

If someone uploads 10 images in a single request and 5 variants need be created, currently this creates 50 jobs in the request.
With this change it would create 10 jobs that eventually schedule 50 jobs.

We already do this for the PreviewImageJob because we need to create the first preview before the variants can be created:
```
# Process the first preview directly because it's required for all variations
blob.preview({}).processed 

 variations.each do |transformations| 
   # preprocessed schedules a TransformJob for each variations
   blob.preprocessed(transformations) 
 end 
```

This is what preprocessed looks like:
https://github.com/rails/rails/blob/4c53863e354ad94dbfaf0643d9cdc99393a1a583/activestorage/app/models/active_storage/blob/representable.rb#L175-L177

So this change would also make things more consistent: we always schedule a single job to schedule processing of variants.

I also think #51951 will be a smaller change with this PR.

activestorage/app/jobs/active_storage/create_variants_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ActiveStorage::CreateVariantsJob < ActiveStorage::BaseJob
```
p8: Maybe this should just be named VariantsJob, more inline with AnalyzeJob, TransformJob, PreviewImageJob, ...
```suggestion
class ActiveStorage::VariantsJob < ActiveStorage::BaseJob
```

p8:
Closing now that #51951 has been merged.

PR #56242, opened 2025-11-27
Ensure `TimeWithZone#as_json` always return an UTF-8 string

### LINKED ISSUE #56241: Regression in `ActiveSupport::TimeWithZone#as_json` encoding in 8.1.0
`TimeWithZone#as_json` returns US-ASCII encoded strings instead of UTF-8 in Rails 8.1.0, breaking JSON serialization for Active Record timestamp columns.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails", "8.0.4"
  gem "rails", "8.1.0"
end

require "active_support"
require "active_support/core_ext/object/blank"
require "minitest/autorun"

  def test_active_support_time_with_zone_as_json_encoding
    assert_equal Encoding::UTF_8, Time.now.in_time_zone("Australia/Brisbane").as_json.encoding
  end
end
```

### Expected behavior

Returns a UTF-8 String.

This happens on 8.0.4:
```
# Running:

.

Finished in 0.016225s, 61.6333 runs/s, 61.6333 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

### Actual behavior

Returns a US-ASCII String.

This happens on 8.1.0:
```
# Running:

F

Finished in 0.019657s, 50.8725 runs/s, 50.8725 assertions/s.

  1) Failure:
BugTest#test_active_support_time_with_zone_as_json_encoding [test.rb:18]:
Expected: #<Encoding:UTF-8>
  Actual: #<Encoding:US-ASCII>

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

### System configuration

**Rails version**: 8.1.0

**Ruby version**: 3.4.7

cc @byroot I haven't done a proper bisect, but by mucking around in the app at work I'm pretty sure the regression was introduced in https://github.com/rails/rails/pull/55583/commits/4b27a2a498655e96f6f4b5e66b5cbec83191cd98

byroot:
Fix: https://github.com/rails/rails/issues/56241

`Time#xmlschema` returns an US-ASCII string.

cc @joshuay03

### DIFF
--- a/activesupport/lib/active_support/time_with_zone.rb
+++ b/activesupport/lib/active_support/time_with_zone.rb
@@ -181,7 +181,7 @@ module ActiveSupport
     #   # => "2005/02/01 05:15:10 -1000"
     def as_json(options = nil)
       if ActiveSupport::JSON::Encoding.use_standard_json_time_format
-        xmlschema(ActiveSupport::JSON::Encoding.time_precision)
+        xmlschema(ActiveSupport::JSON::Encoding.time_precision).force_encoding(Encoding::UTF_8)
       else
         %(#{time.strftime("%Y/%m/%d %H:%M:%S")} #{formatted_offset(false)})
       end


joshuay03 (approved):
Thank you!

joshuay03:
I've confirmed this fixes the issue I described [here](https://github.com/rails/rails/issues/56241) btw.

byroot:
Backported.

PR #56245, opened 2025-11-27
Introduce DevToolsController for chrome workspaces

coorasse:
### Motivation / Background

[Chromium is able to automatically detect workspace folders](https://chromium.googlesource.com/devtools/devtools-frontend/+/main/docs/ecosystem/automatic_workspace_folders.md). In order to do that, performs a call to `.well-known/appspecific/com.chrome.devtools.json`.

This PR introduces a default controller to respond to this call.

### Detail

A Rails::DevtoolsController is provided by the Rails engine and a path is added (commented out) to the routes.rb file:

```ruby
get ".well-known/appspecific/com.chrome.devtools.json", to: "rails/devtools#show"
```

The controller returns the `Rails.root` path and a uuid that changes only if the Rails.root path changes.

### Additional information

There's plenty of other ways we can implement this:
* a middleware
* a non-modifiable route
* clone the controller in the project

I think this implementation is a good trade-off: it provides a good default, which can be easily enabled/disabled also in new apps. It can be easily overridden if the functionality changes.

### Checklist

### DIFF
--- a/railties/lib/rails.rb
+++ b/railties/lib/rails.rb
@@ -30,6 +30,7 @@ module Rails
   autoload :InfoController
   autoload :MailersController
   autoload :WelcomeController
+  autoload :DevtoolsController
 
   eager_autoload do
     autoload :HealthController
--- a/railties/lib/rails/devtools_controller.rb
+++ b/railties/lib/rails/devtools_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class Rails::DevtoolsController < ActionController::Base # :nodoc:
+  def show
+    root_path = Rails.root.to_s
+    hash = Digest::SHA1.hexdigest(root_path)
+    uuid = "#{hash[0..7]}-#{hash[8..11]}-#{hash[12..15]}-#{hash[16..19]}-#{hash[20..31]}"
+
+    render json: {
+      "workspace": {
+        "root": root_path,
+        "uuid": uuid
+      }
+    }
+  end
+end
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Add a new internal route in development to respond to chromium devtools GET request.
+
+    This allows the app folder to be easily connected as a workspace in chromium-based browsers.
+
+    *coorasse*
+
 *   Set `config.rake_eager_load` in generated test environment to match `config.eager_load` behavior in CI.
 
     This ensures eager loading works consistently in CI when rake tasks invoke the `:environment` task before tests run.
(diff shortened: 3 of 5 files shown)


coorasse:
CI failure seems unrelated. I'll add the CHANGELOG if the proposal looks good.

railties/lib/rails/devtools_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class Rails::DevtoolsController < ActionController::Base # :nodoc:
+  def show
+    require "digest"
```
byroot: `require` shouldn't be called from controller actions.

railties/lib/rails/generators/rails/app/templates/config/routes.rb.tt
```diff
@@ -9,6 +9,7 @@ Rails.application.routes.draw do
   # Render dynamic PWA files from app/views/pwa/* (remember to link manifest in application.html.erb)
   # get "manifest" => "rails/pwa#manifest", as: :pwa_manifest
   # get "service-worker" => "rails/pwa#service_worker", as: :pwa_service_worker
+  # get ".well-known/appspecific/com.chrome.devtools.json", to: "rails/devtools#show"
```
byroot: We shouldn't add framework routes to people's `config/routes`, especially since in this case it's meant to be a development only route.

Look for `initializer :add_internal_routes` in `finisher.rb` to see how to do that cleanly.

byroot (changes requested):
I'm positive on the feature, but the implementation needs some work.

railties/lib/rails/devtools_controller.rb
coorasse: I think I can completely remove it.

railties/lib/rails/generators/rails/app/templates/config/routes.rb.tt
coorasse: right! Very good point.

railties/lib/rails.rb
```diff
@@ -34,6 +34,7 @@ module Rails
   eager_autoload do
     autoload :HealthController
     autoload :PwaController
+    autoload :DevtoolsController
```
byroot: This doesn't need to be in `eager_autoload` as it won't be used in production.

byroot:
Please squash your commits, and then it's good for me.

coorasse:
commits squashed. I wonder cif it's a good idea to test the format of the uuid and that is a valid v4 UUID

byroot (approved):

railties/lib/rails/devtools_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Rails::DevtoolsController < ActionController::Base # :nodoc:
+  def show
+    root_path = Rails.root.to_s
+    hash = Digest::SHA1.hexdigest(root_path)
+    uuid = "#{hash[0..7]}-#{hash[8..11]}-#{hash[12..15]}-#{hash[16..19]}-#{hash[20..31]}"
```
schmijos, line 7: @coorasse why not use [`uuid_from_hash`](https://api.rubyonrails.org/classes/Digest/UUID.html#method-c-uuid_from_hash)?

PR #56249, opened 2025-11-28
Simplify setting of threads in Puma config template

joshuay03:
### Motivation / Background

https://github.com/puma/puma/pull/3309 allowed for passing in a single argument to `Puma::DSL#threads` which will be considered to be both the min and the max. This was released in 6.5.0, the latest release is 7.1.0. Since Rails uses the same value for both min and max, we can simplify this by passing just one value in.

Puma Gemfile entry for reference: https://github.com/rails/rails/blob/1e1d8497a0fdad1f4bd3a6efbe7453217b4e56c5/railties/lib/rails/generators/app_base.rb#L295-L297

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/puma.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/puma.rb.tt
@@ -25,8 +25,7 @@
 # Any libraries that use a connection pool or another resource pool should
 # be configured to provide at least as many connections as the number of
 # threads. This includes Active Record's `pool` parameter in `database.yml`.
-threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
-threads threads_count, threads_count
+threads ENV.fetch("RAILS_MAX_THREADS", 3)
 
 # Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 port ENV.fetch("PORT", 3000)


railties/lib/rails/generators/rails/app/templates/config/puma.rb.tt
```diff
@@ -26,7 +26,7 @@
 # be configured to provide at least as many connections as the number of
 # threads. This includes Active Record's `pool` parameter in `database.yml`.
 threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
-threads threads_count, threads_count
+threads threads_count
```
morgoth: Maybe it would even make sense to inline all, ie
`threads ENV.fetch("RAILS_MAX_THREADS", 3)`

just like port is set

joshuay03: Thanks, updated.

maxfelsher:
> ...This was released in 6.5.0, the latest release is 7.1.0. Since Rails uses the same value for both min and max, we can simplify this by passing just one value in.
> 
> Puma Gemfile entry for reference:
> 
> https://github.com/rails/rails/blob/1e1d8497a0fdad1f4bd3a6efbe7453217b4e56c5/railties/lib/rails/generators/app_base.rb#L295-L297
> 

If the generated Puma config requires Puma >= 6.5.0 in order to not break, I wonder if the requirement in the generated Gemfile should be updated to match? I know that by default Bundler will pick the latest version available here, but it doesn't seem out of the question that an application developer could add a different gem to their Gemfile that forces the version of Puma to drop below 6.5.0. Maybe that's on the application developer to figure out since they added the gem? I'm not sure.

joshuay03:
> Maybe that's on the application developer to figure out since they added the gem? I'm not sure.

I think that's fair. The goal is for a new Rails app to boot successfully; any changes from there on out are on the developer.

**Edit:** Actually, I changed my mind on this. Will send a follow-up.

joshuay03:
> **Edit:** Actually, I changed my mind on this. Will send a follow-up.

https://github.com/rails/rails/pull/56254

Issue #45636, opened 2022-07-22
Invalid template types should raise an exception


axos88:
### Steps to reproduce

  - Create a template in views/foo/bar.txt.erb # Note .txt, and not .text

```ruby
# Your reproduction script goes here
```
details = { locale: [:en], formats: [:txt], handlers: [:erb], variants: []}
ActionController::Base.view_paths.find("bar", 'foo', false, details, nil, [])

### Expected behavior

Should throw an exception that :txt is not a registered type in Template::Types

### Actual behavior

Silently fails to find the otherwise existing template, specifying the correct directory it the file resides in. 
Confustion intensifies.

### System configuration
**Rails version**: 7.0.3
**Ruby version**: 2.7.6

axos88:
Still a problem

axos88:
Still outstanding

axos88:
Still outstanding

jhawthorn:
> ActionController::Base.view_paths.find("bar", 'foo', false, details, nil, [])

ViewPaths is an instance of PathSet which is a private (`:nodoc:`) API.  Do you have an example using a public API?

Issue #42201, opened 2021-05-11
Final newline rendered in partials


boardfish:
### Steps to reproduce

See [test repo](https://github.com/boardfish/view_component-bug). `bundle && rails s`, then navigate to localhost:3000.

The test repo has a partial that renders an anchor tag. The file ends with a newline, which is commonly established as a standard.

Identified in github/view_component#913. It's worth checking out that issue and the linked PR for some further insight.

### Expected behavior

The following should be rendered:

```html
<p>
  <a href="https://simon.fish">Add Test template here</a>, <a href="https://simon.fish">Add Test template here</a>, and also <a href="https://simon.fish">Add Test template here</a>
</p>
```

### Actual behavior

The following is rendered:

```html
<p>
  <a href="https://simon.fish">Add Test template here</a>
, <a href="https://simon.fish">Add Test template here</a>
, and also
  <a href="https://simon.fish">Add Test template here</a>
</p>
```

The newlines introduced by `render` appear as a singular space before the commas in the first two instances. It renders like this:

![Screen Shot 2021-05-12 at 13 52 09](https://user-images.githubusercontent.com/20680337/117818049-20184c80-b260-11eb-8dda-f60ee53f0a1f.png)

I'm aware that `<%= -%>` is expected to strip trailing whitespace, but it doesn't appear to do so when rendering partials or ViewComponents. I figure this is most likely an ActionView-specific problem.

### System configuration
**Rails version**: 6.1.3.2

**Ruby version**: 2.7.2p137

lfalcao:
@boardfish both partials have a newline at the end. Its not the `render` method who is adding them...

boardfish:
What I'm saying is that the newline should be ignored. I'll copy my comment from github/view_component#920 here:

> It seems as though the final newline of the template file is rendered, even though this usually indicates the end of a file.
>
> As @\Spone mentioned, trailing newlines are something of a standard ([here's one stance on why](https://unix.stackexchange.com/a/18789), if you're
> curious). It's definitely something we lint for over @raisedevs and something [RuboCop](https://www.rubydoc.info/gems/rubocop/0.20.1/Rubocop/Cop/Style/FinalNewline) checks out.
>
> A per-component or global option would be a great idea. :sparkles:

boardfish:
When you see the :no_entry: symbol at the end of a GitHub diff, that means there's no trailing newline to mark the end of the file.

zzak:
[...]

PR #56251, opened 2025-11-28
Adds template-file rendering examples in the bug report test-template for ActionView

shivabhusal:
## Motivation

When developers report view-template rendering issues, they often struggle to provide reproducible code. As a result, many issue descriptions are unclear or incomplete, making them difficult to understand or reproduce. This is largely because we do not currently provide a predefined template for such reports.

## Details

This PR enhances the existing bug report test template for Action View by adding clearer guidance and structure for issues related to template file rendering. The updated template helps contributors create executable, reproducible test cases, improving the quality and efficiency of issue triage and resolution.

### DIFF
--- a/guides/bug_report_templates/action_view.rb
+++ b/guides/bug_report_templates/action_view.rb
@@ -23,13 +23,40 @@ end
 Rails.application.initialize!
 
 class BugTest < ActionView::TestCase
+  setup do
+    @view_path = Pathname.new(Dir.mktmpdir)
+    ActionController::Base.prepend_view_path @view_path
+  end
+
+  teardown do
+    @view_path.rmtree
+  end
+
   helper do
     def upcase(value)
       value.upcase
     end
   end
 
-  def test_stuff
+  def test_action_view_render_template
+    view_file "posts/index.html.erb", <<~ERB
+      <h1>Posts</h1>
+
+      <% posts.each do |post| %>
+        <%= render partial: "posts/post", locals: { post: post } %>
+      <% end %>
+    ERB
+
+    view_file "posts/_post.html.erb", <<~ERB
+      <p><%= upcase(post) %></p>
+    ERB
+
+    render template: "posts/index", locals: { posts: ["hello world"] }
+
+    assert_equal "HELLO WORLD", rendered.html.at("p").text
+  end
+
+  def test_action_view_render_inline
     render inline: <<~ERB, locals: { key: "value" }
       <p><%= upcase(key) %></p>
     ERB
@@ -38,4 +65,11 @@ class BugTest < ActionView::TestCase
 
     assert_equal element.text, "VALUE"
   end
+
+  private
+    def view_file(filename, contents)
+      pathname = @view_path.join(filename)
+      pathname.dirname.mkpath
+      pathname.write(contents.chomp)
+    end
 end


zzak:
Can you provide a few examples where this was the case?

> When developers report view-template rendering issues, they often struggle to provide reproducible code. As a result, many issue descriptions are unclear or incomplete, making them difficult to understand or reproduce. This is largely because we do not currently provide a predefined template for such reports.

shivabhusal:
@zzak Here are a few example issues that I believe would be clearer with the new reproducible template:

- https://github.com/rails/rails/issues/56239
- https://github.com/rails/rails/issues/45636
- https://github.com/rails/rails/issues/14991
- https://github.com/rails/rails/issues/55161
- https://github.com/rails/rails/issues/42201

Happy to iterate further. Suggestions to make the template even more useful are very welcome.

zzak:
At that point, why not just make this the default for AV reports?

guides/bug_report_templates/action_view_with_template_files.rb
```diff
@@ -0,0 +1,81 @@
...
+    # config ActionController to search for templates at our tmp dir
+    ActionController::Base.prepend_view_path @tmp_path
+  end
+
+  def teardown
+    # delete the tmp dir with files
+    FileUtils.rm_rf @tmp_path
+
+    # restores the old view path
+    ActionController::Base.view_paths = @old_view_path
```
zzak: For most bug reports we probably only need to setup the paths once, not on each test. That might simplify this code too.

shivabhusal:
> At that point, why not just make this the default for AV reports?

@zzak I don’t think everyone needs a multi-file/directory-based test template for every rendering issue. Inline rendering is often sufficient and solves most cases. For situations where a fuller setup is required, this template is available. 

So I’d prefer keeping both options—similar to how we have two templates for `ActiveRecord`.

seanpdoyle:
> I don’t think everyone needs a multi-file/directory-based test template for every rendering issue. Inline rendering is often sufficient and solves most cases. For situations where a fuller setup is required, this template is available.

I believe that multiple templates increases the odds that the bug reporter makes a wrong choice when creating a reproduction script. The existing Action View template would benefit from a more maximal setup that could be self-edited by the reporter to reduce its scope to better suit their report.

guides/bug_report_templates/action_view_with_template_files.rb
```diff
@@ -0,0 +1,77 @@
...
+
+Rails.application.initialize!
+
+class BugTest < ActionView::TestCase
+  def setup
+    @tmp_path = Pathname(Dir.mktmpdir)
+
+    # config ActionController to search for templates at our tmp dir
+    ActionController::Base.prepend_view_path @tmp_path
+  end
```
seanpdoyle: This is an opportunity to call the [Pathname.mktmpdir](https://docs.ruby-lang.org/en/master/Pathname.html#method-c-mktmpdir) class method to combine these steps into a single call.

What do you think about renaming the variable to self-document its intent?

```suggestion
  setup do
    @view_path = Pathname.mktmpdir
    ActionController::Base.prepend_view_path @view_path
  end
```

guides/bug_report_templates/action_view_with_template_files.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def teardown
+    # delete the tmp dir with files
+    FileUtils.rm_rf @tmp_path
+  end
```
seanpdoyle: Could this utilize [Pathname#rmtree](https://docs.ruby-lang.org/en/master/Pathname.html#method-i-rmtree)?

```suggestion
  teardown do
    @view_path.rmtree
  end
```

guides/bug_report_templates/action_view_with_template_files.rb
```diff
@@ -0,0 +1,77 @@
...
+      ERB
+    end
+
+    render template: template_name
+
+    element_contents = rendered.html.css("p").map(&:text)
+
+    assert_equal %w[value1 value2 value3], element_contents
+  end
+end
```
seanpdoyle: Methods defined within the `helper` block will be available from the view template, whereas the `partial_content` and `template_content` methods are more focused on creating files in the `@view_path` directory.

Imagining the experience of a bug reporter, there seems to be a lot of indirection, with strings, values, and concepts spread across a variety of places in the file. 

What do you think about simplifying this initial code to separate out Action View-centric `helper { … }` block from the test case-centric file writing methods? Would it help to reduce the number of utilities available? Would utilities benefit from accepting HEREDOC strings directly, rather than blocks that yield HEREDOC strings?

```suggestion
  helper do
    def titleize(string)
      string.titleize
    end
  end

  test "Action View rendering" do
    view_file "posts/index.html.erb", <<~ERB
      <h1>Posts</h1>

      <% posts.each do |post| %>
        <%= render partial: "posts/post", locals: { post: post } %>
      <% end %>
    ERB
    view_file "posts/_post.html.erb", <<~ERB
      <p><%= titleize(post) %></p>
    ERB

    render template: "posts/index", locals: { posts: ["hello world"] }

    assert_equal "Hello World", rendered.html.at("p").text
  end

  private

  def view_file(filename, contents)
    pathname = @view_path.join(filename)
    pathname.dirname.mkpath
    pathname.write(contents.chomp)
  end
end
```

seanpdoyle:
Are bug report templates executed as part of CI? IF so, should the PR description remove the `[ci skip]` suffix so that these proposed changes are evaluated and pass before merge?

shivabhusal:
> I believe that multiple templates increases the odds that the bug reporter makes a wrong choice when creating a reproduction script. The existing Action View template would benefit from a more maximal setup that could be self-edited by the reporter to reduce its scope to better suit their report.

@seanpdoyle  Let me make this the default flle.

I see some errors in the template after applying patches, working to fix it. 

> Are bug report templates executed as part of CI? IF so, should the PR description remove the [ci skip] suffix so that these proposed changes are evaluated and pass before merge?

Looks like yes, they are tested in CI. I will remove the `Ci-Skip`

shivabhusal:
The failing tests seem unrelated to this PR

guides/bug_report_templates/action_view.rb
```diff
@@ -16,26 +16,62 @@
 
 class TestApp < Rails::Application
   config.load_defaults Rails::VERSION::STRING.to_f
-  config.eager_load = false
-  config.logger = Logger.new($stdout)
-  config.secret_key_base = "secret_key_base"
+  config.eager_load       = false
+  config.logger           = Logger.new($stdout)
+  config.secret_key_base  = "secret_key_base"
```
zzak: Please remove these cosmetic changes :pray:

guides/bug_report_templates/action_view.rb
```diff
@@ -16,26 +16,62 @@
...
+    ActionController::Base.prepend_view_path @view_path
+  end
+
+  teardown do
+    @view_path.rmtree
+  end
+
   helper do
-    def upcase(value)
-      value.upcase
```
zzak, line 28: I think the existing helper is fine as-is

guides/bug_report_templates/action_view.rb
```diff
@@ -16,26 +16,62 @@
...
+    def titleize(string)
+      string.titleize
     end
   end
 
-  def test_stuff
+  test "Action View template rendering" do
```
zzak:
```suggestion
  def test_action_view_render_template
```

guides/bug_report_templates/action_view.rb
```diff
@@ -16,26 +16,62 @@
...
+      <p><%= titleize(post) %></p>
+    ERB
+
+    render template: "posts/index", locals: { posts: ["hello world"] }
+
+    assert_equal "Hello World", rendered.html.at("p").text
+  end
+
+
+  test "Action View inline rendering" do
```
zzak:
```suggestion
  def test_action_view_render_inline
```

guides/bug_report_templates/action_view.rb
```diff
@@ -20,16 +20,44 @@ class TestApp < Rails::Application
   config.logger = Logger.new($stdout)
   config.secret_key_base = "secret_key_base"
 end
+
```
zzak:
```suggestion
```

zzak:
Can you please [squash your commits](https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#squashing-commits) and I think the title/description should call out we're _expanding_ the Action View bug report to include template file rendering.

shivabhusal:
@zzak I squashed my commits.

seanpdoyle (approved):

PR #56252, opened 2025-11-28
[ci skip] Improve docs about system test generation

callmesangio:
### Motivation / Background

This Pull Request has been created because the docs about system test generation could be improved.

### Detail

This is a follow-up to #55743 and #56272.

Testing guide:

- Update the code block in the "Test Setup" section to show the actual output of the `ls -F test` command.
- Reword the "Test Directories" section to emphasize that contents related to system tests must be explicitly generated.
- Fix a rendering issue in the "Generating System Tests" section.
- Update code blocks showing the output of system test generation commands.

Testing guide, API docs:

- The `application_system_test_case.rb` file is no longer generated by default with a new application or scaffold, so fix the docs wherever they suggest otherwise.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/system_test_case.rb
+++ b/actionpack/lib/action_dispatch/system_test_case.rb
@@ -22,8 +22,8 @@ module ActionDispatch
   #
   # To create a system test in your application, extend your test class from
   # `ApplicationSystemTestCase`. System tests use Capybara as a base and allow you
-  # to configure the settings through your `application_system_test_case.rb` file
-  # that is generated with a new application or scaffold.
+  # to configure the settings through the `application_system_test_case.rb` file,
+  # which is created when you generate your first system test.
   #
   # Here is an example system test:
   #
@@ -41,7 +41,7 @@ module ActionDispatch
   #       end
   #     end
   #
-  # When generating an application or scaffold, an
+  # When generating system tests, an
   # `application_system_test_case.rb` file will also be generated containing the
   # base class for system testing. This is where you can change the driver, add
   # Capybara settings, and other configuration for your system tests.
--- a/guides/source/testing.md
+++ b/guides/source/testing.md
@@ -40,7 +40,7 @@ then you will see:
 
 ```bash
 $ ls -F test
-controllers/                     helpers/                         mailers/                         fixtures/                        integration/                     models/                          test_helper.rb
+controllers/  fixtures/  helpers/  integration/  mailers/  models/  test_helper.rb
 ```
 
 ### Test Directories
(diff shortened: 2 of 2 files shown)


callmesangio:
The rendering of the `ls -F test` example in the testing guide is different from the one currently shown at `guides.rubyonrails.org`, but matches the one at `edgeguides.rubyonrails.org`:

This PR: https://callmesangio-fix-testing-doc.rails-docs-preview.pages.dev/guides/testing#test-setup
guides: https://guides.rubyonrails.org/testing.html#test-setup
edgeguides: https://edgeguides.rubyonrails.org/testing.html#test-setup

If edgeguides rendering is the new reference, I think it's probably worth reworking the spacing between files/directories in the example in this PR to make it more compact.

callmesangio:
I've tracked down the new rendering behavior to https://github.com/rails/rails/pull/56084, so I assume it is intended.
I've updated the PR with the output of `ls -F test` (`ls` from GNU coreutils).

actionpack/lib/action_dispatch/system_test_case.rb
```diff
@@ -23,7 +23,7 @@ module ActionDispatch
   # To create a system test in your application, extend your test class from
   # `ApplicationSystemTestCase`. System tests use Capybara as a base and allow you
   # to configure the settings through your `application_system_test_case.rb` file
-  # that is generated with a new application or scaffold.
+  # that is generated with a new application.
```
eileencodes: It's no longer generated with an application or a scaffold so this change is only half correct. Can you fix the doc updates? This is the PR that updated the behavior https://github.com/rails/rails/pull/55743

eileencodes (changes requested):

actionpack/lib/action_dispatch/system_test_case.rb
callmesangio: Hi @eileencodes, thanks for reviewing this.

I'm not sure I'm following: by "new application" I assumed "by running `rails new`", which still generates the `application_system_test_case.rb` file as of Rails 8.1.1.
This is the reason why this PR also fixes the `ls -F test` example in the guide.
Under this assumption, removing just the scaffold part in the docs seems correct to me. Am I missing something?

eileencodes: Oh guess I missed that. I just opened this [PR](https://github.com/rails/rails/pull/56272), can you update the docs to match the new behavior?

actionpack/lib/action_dispatch/system_test_case.rb
```diff
@@ -22,8 +22,8 @@ module ActionDispatch
   #
   # To create a system test in your application, extend your test class from
   # `ApplicationSystemTestCase`. System tests use Capybara as a base and allow you
-  # to configure the settings through your `application_system_test_case.rb` file
-  # that is generated with a new application or scaffold.
+  # to configure the settings through the `application_system_test_case.rb` file
+  # that is created when you first generate them.
```
brkn: Would this read better here?

```suggestion
  # to configure the settings through the `application_system_test_case.rb` file,
  # which is created when you generate your first system test.
```

callmesangio: Thanks for the suggestion.
I'm not a native English speaker, but I honestly think that sentence reads ok as it is:

> System tests [...] allow you to configure the settings through the [...] file that is created when you first generate them.

actionpack/lib/action_dispatch/system_test_case.rb
callmesangio: @eileencodes Hi, docs updated. This PR is reviewable again. Thanks!

actionpack/lib/action_dispatch/system_test_case.rb
eileencodes: I prefer the suggestion, while the original is ok, the `which is created when you generate your first system test` is clearer about how this file gets generated.

eileencodes (commented):
Looks good, I think this should accept the suggestion from @brkn because it's a little bit clearer about when the application_system_test_case.rb gets generated. Thanks for iterating on this.

actionpack/lib/action_dispatch/system_test_case.rb
callmesangio: Sure! Change integrated.

callmesangio:
> Looks good, I think this should accept the suggestion from @brkn because it's a little bit clearer about when the application_system_test_case.rb gets generated. Thanks for iterating on this.

All should be good now.
Thank you again for reviewing!

eileencodes:
I backported this to 8-1-stable. Thanks @callmesangio!

PR #56253, opened 2025-11-28
[ActiveSupport] Fix `in_order_of` with `filter: false` to work with procs and lambdas

igordepolli:
### Motivation / Background

This Pull Request has been created to fix `in_order_of` with `filter: false` to work with procs and lambdas.

### Detail

This Pull Request changes the `in_order_of` method of the `Enumerable` class. Now the method can handle procs and lambdas.

### Additional information

```rb
arr_a = [{ id: 1 }, { id: 5}, { id: 3}]
arr_b = [5, 3]
```

When `in_order_of` is called with `filter: true` (default), this works and returns:

```rb
arr_a.in_order_of(-> { it[:id] }, arr_b)
=> [{id: 5}, {id: 3}]
```

Currently, when we set `filter: false`, an exception is raised:

```rb
 arr_a.in_order_of(-> { it[:id] }, arr_b, filter: false)
#<Proc:0x00000001335597c0 (simbios):10 (lambda)> is not a symbol nor a string (TypeError)
# sort_by { |v| series.index(v.public_send(key)) || series.size }.compact
```

Expected behavior is:

```rb
arr_a.in_order_of(-> { it[:id] }, arr_b, filter: false)
=> [{id: 5}, {id: 3}, {id: 1}]
```

This PR fixes this issue.

### Checklist


activesupport/test/core_ext/enumerable_test.rb
```diff
@@ -391,6 +391,22 @@ def test_in_order_of_with_filter_false
...
+  end
+
+  def test_in_order_of_works_with_lambdas
+    values = [ { "price" => 3 }, { "price" => 5 }, { "price" => 1 } ]
+    assert_equal [ { "price" => 1 }, { "price" => 5 } ], values.in_order_of(lambda { |e| e["price"] }, [ 1, 5 ])
+
+    values = [ { "price" => 3 }, { "price" => 5 }, { "price" => 1 } ]
+    assert_equal [ { "price" => 1 }, { "price" => 5 }, { "price" => 3 } ], values.in_order_of(lambda { |e| e["price"] }, [ 1, 5 ], filter: false)
+  end
+
```
willianveiga, line 409: I’d also add similar specs for the case where the filter is true.

willianveiga (commented):
Good job so far. Could you also fix the merge conflicts? Thanks!

activesupport/test/core_ext/enumerable_test.rb
igordepolli: @willianveiga Ty for your review, bro! 

The first assert in both tests already tests that, since the default value for the filter argument is true.

fatkodima:
Thanks for the PR, but this is a very specific use case and not clear its real world usage and how it will be implemented for Active Record version. Closing for now.

igordepolli:
> Thanks for the PR, but this is a very specific use case and not clear its real world usage and how it will be implemented for Active Record version. Closing for now.

Ty for your review!

About the PR, i think i don't follow you... The purpose here was only fix the current "strange" behavior when we try use this method with `filter: false`. Without it, works normally, but with this option, breaks without a clear message. The idea isn’t to create a new feature, either for AS or AR, but rather to fix some weird behavior that the extra option can cause.

PR #52541, opened 2024-08-08
Remove redundant Puma configuration settings

dhh:
1. Puma automatically sets its environment from RAILS_ENV
2. preload_app! is automatically set when in cluster mode
3. WEB_CONCURRENCY is automatically used and is now parsed directly so can't be extended with "auto" here.
4. worker_timeout is a cluster mode feature, and we don't run that in dev.

PR #56254, opened 2025-11-29
Update puma gem constraint from >= 5.0 to >= 7.1

joshuay03:
### Motivation / Background

Inspired by a few things:
- https://github.com/rails/rails/pull/56249#issuecomment-3590664309 
- https://github.com/rails/rails/pull/52541#issuecomment-3325028732
- https://github.com/hanami/hanami-cli/pull/372

TL;DR: The change in https://github.com/rails/rails/pull/56249 relies on puma 6.5, but instead of constraining to that I went with 7.0 because it includes a fix for a very confusing inconsistency bug if users decide to configure workers with `Puma::DSL#workers` instead of `WEB_CONCURRENCY`.

**Edit:** Going to 7.1, see https://github.com/rails/rails/pull/56254#discussion_r2572767896

cc @guilleiguaran

### Checklist

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -293,7 +293,7 @@ module Rails
       end
 
       def web_server_gemfile_entry # :doc:
-        GemfileEntry.new "puma", ">= 5.0", "Use the Puma web server [https://github.com/puma/puma]"
+        GemfileEntry.new "puma", ">= 7.1", "Use the Puma web server [https://github.com/puma/puma]"
       end
 
       def asset_pipeline_gemfile_entry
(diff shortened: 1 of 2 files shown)


railties/lib/rails/generators/app_base.rb
```diff
@@ -293,7 +293,7 @@ def database_gemfile_entry # :doc:
       end
 
       def web_server_gemfile_entry # :doc:
-        GemfileEntry.new "puma", ">= 5.0", "Use the Puma web server [https://github.com/puma/puma]"
+        GemfileEntry.new "puma", ">= 7.0", "Use the Puma web server [https://github.com/puma/puma]"
```
guilleiguaran: Let's go for 7.1 👍

joshuay03: Done!

PR #56258, opened 2025-11-29
Add schematized json for has_json

dhh:
Provides a schema-enforced access object for a JSON attribute. This allows you to assign values
directly from the UI as strings, and still have them set with the correct JSON type in the database.

Only the three basic JSON types are supported: boolean, integer, and string. No nesting either.
These types can either be set by referring to them by their symbol or by setting a default value.
Default values are set when a new model is instantiated and on +before_save+ (if defined).

Examples:

```ruby
  has_json :settings, restrict_creation_to_admins: true, max_invites: 10, greeting: "Hello!"
  has_delegated_json :flags, beta: false, staff: :boolean
end

a = Account.new
a.settings.restrict_creation_to_admins? # => true
a.settings.max_invites = "100" # => Set to integer 100
a.settings = { "restrict_creation_to_admins" => "false", "max_invites" => "500", "greeting" => "goodbye" }
a.settings.greeting # => "goodbye"
a.staff # => nil
a.staff = true
a.staff? # => true
```

### DIFF
--- a/activemodel/lib/active_model/schematized_json.rb
+++ b/activemodel/lib/active_model/schematized_json.rb
@@ -12,13 +12,16 @@ module ActiveModel
       # directly from the UI as strings, and still have them set with the correct JSON type in the database.
       def has_json(attr, **schema)
         define_method(attr) do
+          # Ensure the attribute is set if nil, so we can pass the reference to the accessor for defaults.
           _write_attribute(attr.to_s, {}) if attribute(attr.to_s).nil?
+
+          # No memoizationen used in order to stay compatible with #reload (and because it's such a thin accessor)
           ActiveModel::SchematizedJson::DataAccessor.new(schema, data: attribute(attr.to_s))
         end
 
         define_method("#{attr}=") { |data| public_send(attr).assign_data_with_type_casting(data) }
 
-        # Ensure default values are set before saving
+        # Ensure default values are set before saving by relying on DataAccessor instantiation to do it.
         before_save -> { send(attr) } if respond_to?(:before_save)
       end
     end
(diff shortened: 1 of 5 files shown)


n-studio:
@dhh Shouldn't the example be `a.settings.max_invites = "100" # => Set to integer 100` since it's not delegated?

amerritt14:
Any chance of this idea getting implemented in `ActiveRecord::Store` as well?
I've always thought that `store_accessor` should have a default value option.

`store_accessor :settings, :restrict_creation_to_admins, default: false, type: :boolean`

activemodel/lib/active_model/schematized_json.rb
```diff
@@ -0,0 +1,124 @@
...
+      #   a.settings = { "restrict_creation_to_admins" => "false", "max_invites" => "500", "greeting" => "goodbye" }
+      #   a.settings.greeting # => "goodbye"
+      #   a.flags.staff # => nil
+      #   a.flags.staff? # => false
+      def has_json(attr, **schema)
+        define_method(attr) do
+          # Ensure the attribute is set if nil, so we can pass the reference to the accessor for defaults.
+          _write_attribute(attr.to_s, {}) if attribute(attr.to_s).nil?
+
+          # No memoizationen used in order to stay compatible with #reload (and because it's such a thin accessor)
```
henrik, line 37: Typo ”memoizationen”

jrochkind:
I'm having trouble telling the difference between has_json and has_delegated_json from the comment docs.

nikolaokonesh:
How to use has_json in SQLite3 database?

dhh:
SQLite3 supports json columns, so it works as advertised.

PR #56261, opened 2025-11-30
[Fix #56260] Include `env_name` and `name` in `configuration_hash` and respect them in `establish_connection`

### LINKED ISSUE #56260: `ActiveRecord::Base.establish_connection` with `configuration_hash` always connects to primary database
### Steps to reproduce

Passing a hash to `ActiveRecord::Base.establish_connection` with a non-primary configuration always connects to the primary database. The issue here is twofold:
1. `ActiveRecord::DatabaseConfigurations::DatabaseConfig#configuration_hash` does not include the `env_name` and `name`, so unlike with a `DatabaseConfig`/`HashConfig` object, the config state isn't preserved to re-establish a fresh connection using that hash.
2. Even if `env_name` and `name` are provided in a Hash arguement for `ActiveRecord::Base.establish_connection`, the environment always defaults to the Rails environment and the name always defaults to primary

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "8.1.1"

  gem "sqlite3"
end

require "minitest/autorun"

FileUtils.mkdir_p("config")

File.open("config/database.yml", "w") do |file|
  file.write(<<~YAML)
    development:
      primary:
        adapter: sqlite3
        database: ":memory:"
      shard_001:
        adapter: sqlite3
        database: ":memory:"
      shard_002:
        adapter: sqlite3
        database: ":memory:"
  YAML
end

  config.eager_load = false

end
Rails.application.initialize!

  Minitest.after_run do
    FileUtils.rm_f("config")
  end

  def setup
    ActiveRecord::Base.establish_connection(:primary)
    @primary_db_config = ActiveRecord::Base.connection.pool.db_config
    ActiveRecord::Base.establish_connection(:shard_001)
    @shard_001_db_config = ActiveRecord::Base.connection.pool.db_config

    ActiveRecord::Base.establish_connection(:primary)
  end

  def test_establish_connection_symbol
    ActiveRecord::Base.establish_connection(:primary)
    assert_equal "primary", ActiveRecord::Base.connection.pool.db_config.name
    ActiveRecord::Base.establish_connection(:shard_001)
    assert_equal "shard_001", ActiveRecord::Base.connection.pool.db_config.name
  end

  def test_establish_connection_hash_config
    ActiveRecord::Base.establish_connection(@primary_db_config)
    assert_equal "primary", ActiveRecord::Base.connection.pool.db_config.name
    ActiveRecord::Base.establish_connection(@shard_001_db_config)
    assert_equal "shard_001", ActiveRecord::Base.connection.pool.db_config.name
  end

  def test_establish_connection_hash
    ActiveRecord::Base.establish_connection(@primary_db_config.configuration_hash)
    assert_equal "primary", ActiveRecord::Base.connection.pool.db_config.name
    ActiveRecord::Base.establish_connection(@shard_001_db_config.configuration_hash)
    assert_equal "shard_001", ActiveRecord::Base.connection.pool.db_config.name # ❌ Expected: "shard_001", Actual: "primary"
  end
end
```

### Expected behavior

When establishing a connection using the `configuration_hash` of an existing non-primary config, the connection should be made to that non-primary database.

### Actual behavior

When establishing a connection using the `configuration_hash` of an existing non-primary config, the connection is made to the primary database:

```
rails [main] ruby test.rb
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Run options: --seed 44621

# Running:

I, [2025-11-30T20:09:33.479838 #41205]  INFO -- : ----------------------------------------------
I, [2025-11-30T20:09:33.479897 #41205]  INFO -- : BugTest: test_establish_connection_hash_config
I, [2025-11-30T20:09:33.479911 #41205]  INFO -- : ----------------------------------------------
.I, [2025-11-30T20:09:33.484290 #41205]  INFO -- : ---------------------------------------
I, [2025-11-30T20:09:33.484326 #41205]  INFO -- : BugTest: test_establish_connection_hash
I, [2025-11-30T20:09:33.484332 #41205]  INFO -- : ---------------------------------------
FI, [2025-11-30T20:09:33.485269 #41205]  INFO -- : -----------------------------------------
I, [2025-11-30T20:09:33.485277 #41205]  INFO -- : BugTest: test_establish_connection_symbol
I, [2025-11-30T20:09:33.485282 #41205]  INFO -- : -----------------------------------------
.

Finished in 0.115594s, 25.9529 runs/s, 51.9058 assertions/s.

  1) Failure:
BugTest#test_establish_connection_hash [test.rb:77]:
Expected: "shard_001"
  Actual: "primary"

3 runs, 6 assertions, 1 failures, 0 errors, 0 skips
```

### System configuration

**Rails version**: 8.1.1

**Ruby version**: 3.4.7

joshuay03:
### Motivation / Background

Closes https://github.com/rails/rails/issues/56260.

When passing a hash to `ActiveRecord::Base.establish_connection` (e.g., from `configuration_hash`), the connection always defaults to the primary database instead of preserving the original database name and environment. This makes it impossible to programmatically switch between shards or non-primary databases using configuration hashes like you can with `DatabaseConfig`/`HashConfig` objects.

### Detail

This Pull Request changes two things:

1. `DatabaseConfig#configuration_hash` now includes `env_name` and `name` so the configuration state is preserved when re-establishing connections
2. `DatabaseConfigurations#build_configs` (used by `#initialize`) and `#resolve`  now check for `name` in the hash first before defaulting to "primary", and the latter also checks for `env_name` in the  hash first before defaulting to the Rails environment.

This ensures that `establish_connection(db_config.configuration_hash)` behaves identically to `establish_connection(db_config)` for non-primary databases.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix `establish_connection` to preserve environment and database name when using configuration hash.
+
+    `ActiveRecord::DatabaseConfigurations::DatabaseConfig#configuration_hash` now includes
+    `env_name` and `name`, and `ActiveRecord::Base.establish_connection` respects these
+    values from the hash before defaulting to the Rails environment and "primary". This
+    allows programmatic database switching using configuration hashes.
+
+    *Joshua Young*
+
 *   Add support for configuring migration strategy on a per-adapter basis.
 
     `migration_strategy` can now be set on individual adapter classes, overriding
--- a/activerecord/lib/active_record/database_configurations.rb
+++ b/activerecord/lib/active_record/database_configurations.rb
@@ -179,8 +179,10 @@ module ActiveRecord
       case config
       when Symbol
         resolve_symbol_connection(config)
-      when Hash, String
+      when String
         build_db_config_from_raw_config(default_env, "primary", config)
+      when Hash
+        build_db_config_from_raw_config(config[:env_name]&.to_s || default_env, config[:name]&.to_s || "primary", config)
       else
         raise TypeError, "Invalid type for configuration. Expected Symbol, String, or Hash. Got #{config.inspect}"
       end
@@ -204,8 +206,12 @@ module ActiveRecord
         return configs if configs.is_a?(Array)
 
         db_configs = configs.flat_map do |env_name, config|
-          if config.is_a?(Hash) && config.values.all?(Hash)
-            walk_configs(env_name.to_s, config)
+          if config.is_a?(Hash)
+            if config.values.all?(Hash)
+              walk_configs(env_name.to_s, config)
+            else
+              build_db_config_from_raw_config(env_name.to_s, config[:name]&.to_s || "primary", config)
+            end
           else
             build_db_config_from_raw_config(env_name.to_s, "primary", config)
           end
--- a/activerecord/lib/active_record/database_configurations/hash_config.rb
+++ b/activerecord/lib/active_record/database_configurations/hash_config.rb
@@ -37,7 +37,7 @@ module ActiveRecord
       #
       def initialize(env_name, name, configuration_hash)
         super(env_name, name)
-        @configuration_hash = configuration_hash.symbolize_keys.freeze
+        @configuration_hash = configuration_hash.symbolize_keys.merge(env_name: env_name, name: name).freeze
         validate_configuration!
       end
 
--- a/activerecord/test/cases/adapters/mysql2/mysql2_rake_test.rb
+++ b/activerecord/test/cases/adapters/mysql2/mysql2_rake_test.rb
@@ -26,7 +26,7 @@ module ActiveRecord
       db_config = ActiveRecord::DatabaseConfigurations::HashConfig.new("default_env", "primary", @configuration)
 
       mock = Minitest::Mock.new
-      mock.expect(:call, nil, [adapter: "mysql2", database: nil])
+      mock.expect(:call, nil, [env_name: "default_env", name: "primary", adapter: "mysql2", database: nil])
       mock.expect(:call, nil, [db_config])
 
       ActiveRecord::Base.stub(:lease_connection, @connection) do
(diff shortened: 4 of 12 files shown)


byroot:
I think it's a good idea. @eileencodes any objections?

eileencodes (changes requested):
Sorry I hope this doesn't come across as harsh, but I'm very against this change. It was by design to not include these in the configuration hash since what's in the hash is "what is needed by the client to establish the connection". Client libraries don't care about `env_name` and `name`, those are specific to Rails.

If applications are using multiple databases, they should be using `connects_to` to establish connections or `connected_to` to switch the context in which connections are chosen. They shouldn't use `Base#establish_connection`, especially in runtime code.

I don't follow this point:

> DatabaseConfig#configuration_hash now includes env_name and name so the configuration state is preserved when re-establishing connections

When is a connection losing configuration state other than in Rake tasks and tests? This shouldn't happen at runtime in development or production code. And that can only happen if the hash changed, otherwise the existing connection is picked up (unless you're setting `clobber: true`).

> DatabaseConfigurations#build_configs (used by #initialize) and #resolve now check for name in the hash first before defaulting to "primary", and the latter also checks for env_name in the hash first before defaulting to the Rails environment.

We default to primary because most applications are single database and so we can assume they are primary.

> This makes it impossible to programmatically switch between shards or non-primary databases using configuration hashes like you can with DatabaseConfig/HashConfig objects.

Why are you using configuration hashes to switch connections? This entirely goes against the design of multiple databases to use the `role` and `shard` to switch connection context. I'm a bit concerned your application is using multiple databases in a way it was not designed. I don't want to change the design of Rails to fit an anti-pattern that I deliberately chose to move away from.

joshuay03:
> When is a connection losing configuration state other than in Rake tasks and tests? This shouldn't happen at runtime in development or production code.

It was a Rake task in this case.

> Why are you using configuration hashes to switch connections? This entirely goes against the design of multiple databases to use the `role` and `shard` to switch connection context. I'm a bit concerned your application is using multiple databases in a way it was not designed. I don't want to change the design of Rails to fit an anti-pattern that I deliberately chose to move away from.

Tried to explain here: https://github.com/rails/rails/issues/56260#issuecomment-3636455592

joshuay03:
Closing based on https://github.com/rails/rails/issues/56260#issuecomment-3638633746.

PR #56267, opened 2025-12-01
Move async execution into QueryIntent

matthewd:
FutureResult remains as the user-facing object, but is no longer directly involved in the actual execution process: the pool's async queue now contains the QueryIntent instance.

Before:

mermaid-diagram-2025-11-12-040028

Now:

mermaid-diagram-2025-12-01-201116

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/query_intent.rb
+++ b/activerecord/lib/active_record/connection_adapters/query_intent.rb
@@ -90,25 +90,6 @@ module ActiveRecord
         "#<#{self.class.name} name=#{name.inspect} allow_retry=#{allow_retry} materialize_transactions=#{materialize_transactions}>"
       end
 
-      # Schedule this intent for async execution via thread pool
-      def async_schedule!(session)
-        # Upgrade to real mutex now that we'll have concurrent access
-        @mutex = Mutex.new
-        @session = session
-
-        # Force preprocessing on original thread before queuing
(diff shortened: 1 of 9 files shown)


activerecord/lib/active_record/connection_adapters/query_intent.rb
```diff
@@ -93,29 +161,95 @@ def has_binds?
...
+        raise @error if @error
+      end
+
       def cast_result
         raise "Cannot call cast_result before query has executed" unless @executed
         raise "Cannot call cast_result after affected_rows has been called" if defined?(@affected_rows)
-        @cast_result ||= adapter.send(:cast_result, raw_result)
+
+        ensure_result
+        @cast_result ||= adapter.send(:cast_result, @raw_result)
```
byroot, line 222: Not new and somewhat tangential, but I think we should make `cast_result` public but `:nodoc:` to avoid that `send`.

byroot (approved):

activerecord/lib/active_record/connection_adapters/query_intent.rb
matthewd: Yeah, I'm currently putting it off just to avoid the extra diff churn on all the adapters before everything feels settled (even though it seems pretty unlikely this call will change), but won't leave it like this 👍🏻

PR #56270, opened 2025-12-02
Fix ERB strict locals parsing when comment spans multiple lines

Saidbek:
### Motivation / Background

Fixes #56239

This PR has been created because ERB partial locals parsing fails when the locals: magic comment spans multiple lines. 
For example, this works:

```ruby
<%# locals: (arg_1:, arg_2: nil, arg_3: []) %>
```
But this raises `ActionView::Template::Error: undefined local variable or method 'arg_2'`:
```ruby
<%# locals: (arg_1:,
             arg_2: nil,
             arg_3: []) %>
```

### Additional information

Now, the below script passes:
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

end

require "action_view/railtie"
require "minitest/autorun"

  config.eager_load = false
end

Rails.application.initialize!

class MultilineLocalsTest < ActionView::TestCase
  def test_partial_with_single_line_strict_locals
    tmp_dir = Dir.mktmpdir
    tmp_path = Pathname(tmp_dir)

    # Single line magic comment (this is working)
    File.write(tmp_path.join("_partial.html.erb"), <<~HTML)
      <%# locals: (arg_1:, arg_2: nil, arg_3: []) %>
      <p class="arg1"><%= arg_1 %></p>
      <p class="arg2"><%= arg_2 %></p>
      <p class="arg3"><%= arg_3 %></p>
    HTML

    File.write(tmp_path.join("main.html.erb"), <<~ERB)
      <div>
        <%= render partial: "partial", locals: { arg_1: "First" } %>
      </div>
    ERB

    ActionController::Base.prepend_view_path tmp_path

    render template: "main"

    assert_equal "First", rendered.html.at_css(".arg1").text
    assert_equal "", rendered.html.at_css(".arg2").text
    assert_equal "[]", rendered.html.at_css(".arg3").text
  ensure
    FileUtils.rm_rf tmp_dir
  end

  def test_partial_with_multiline_strict_locals
    tmp_dir = Dir.mktmpdir
    tmp_path = Pathname(tmp_dir)

    # Multiline magic comment (this is not working)
    File.write(tmp_path.join("_partial.html.erb"), <<~HTML)
      <%# locals: (arg_1:,
                   arg_2: nil,
                   arg_3: []) %>
      <p class="arg1"><%= arg_1 %></p>
      <p class="arg2"><%= arg_2 %></p>
      <p class="arg3"><%= arg_3 %></p>
    HTML

    File.write(tmp_path.join("main.html.erb"), <<~ERB)
      <div>
        <%= render partial: "partial", locals: { arg_1: "First" } %>
      </div>
    ERB

    ActionController::Base.prepend_view_path tmp_path

    render template: "main"

    assert_equal "First", rendered.html.at_css(".arg1").text
    assert_equal "", rendered.html.at_css(".arg2").text
    assert_equal "[]", rendered.html.at_css(".arg3").text
  ensure
    FileUtils.rm_rf tmp_dir
  end
end
```

### Checklist

### DIFF
--- a/actionview/lib/action_view/template.rb
+++ b/actionview/lib/action_view/template.rb
@@ -7,7 +7,7 @@ module ActionView
   class Template
     extend ActiveSupport::Autoload
 
-    STRICT_LOCALS_REGEX = /\#\s+locals:\s+\((.*)\)/
+    STRICT_LOCALS_REGEX = /\#\s+locals:\s+\((.*?)\)(?=\s*-?%>|\s*$)/m
 
     # === Encodings in ActionView::Template
     #
--- a/actionview/test/template/template_test.rb
+++ b/actionview/test/template/template_test.rb
@@ -220,6 +220,28 @@ class TestERBTemplate < ActiveSupport::TestCase
     assert_equal "Hello", render
   end
 
+  def test_locals_with_parentheses
+    @template = new_template("<%# locals: (message: '(Hello)') -%>\n<%= message %>")
+    assert_equal "(Hello)", render
+
+    @template = new_template("<%# locals: (message: '(Hello') -%>\n<%= message %>")
+    assert_equal "(Hello", render
+
+    @template = new_template("<%# locals: (message: 'Hello)') -%>\n<%= message %>")
+    assert_equal "Hello)", render
+  end
+
+  def test_locals_can_spread_to_multiple_lines
+    template = <<~ERB.chomp
+        <%# locals: (arg_1:,
+                     arg_2: nil,
+                     arg_3: []) -%>
+        <%= arg_1 %><%= arg_2 %><%= arg_3 %>
+    ERB
+    @template = new_template(template)
+    assert_equal "First[]", render(arg_1: "First")
+  end
+
   def test_required_locals_must_be_specified
     error = assert_raises(ActionView::Template::Error) do
       @template = new_template("<%# locals: (message:) -%>")


actionview/test/template/template_test.rb
```diff
@@ -220,6 +220,11 @@ def test_default_locals_can_be_specified
     assert_equal "Hello", render
   end
 
+  def test_multiline_locals_can_be_specified
```
shivabhusal:
```suggestion
  def test_locals_can_spread_to_multiple_lines
    template = <<~ERB.chomp
        <%# locals: (arg_1:,
                     arg_2: nil,
                     arg_3: []) -%>
        <%= arg_1 %><%= arg_2 %><%= arg_3 %>
    ERB
    @template = new_template(template)
```

Looks great! 👍🏼
This is just a non-blocking suggestion to slightly improve readability and formatting.

Saidbek:
The failing test seems unrelated (railties errors).

actionview/test/template/template_test.rb
```diff
@@ -220,6 +220,22 @@ def test_default_locals_can_be_specified
     assert_equal "Hello", render
   end
 
+  def test_locals_with_parantheses
+    @template = new_template("<%# locals: (message: '(Hello)') -%>\n<%= message %>")
+    assert_equal "(Hello)", render
+  end
```
byroot: I pushed an extra commit with this test to demonstrate that this PR causes a regression if one of the default values contains a paranthese.

Saidbek: Good catch! I've updated the regex to handle parens in default values. The fix now uses a non-greedy match with a lookahead `(?=\s*-?%>)` to ensure we stop at the `)` that closes the locals declaration, not one inside a default value like `'(Hello)'`.

shivabhusal:
I think at this time, we should also support a case when the closing parenthesis is on next/different line than the last key-value pair. This situation occurs when we format code for alignment.

**Change Suggestion**: https://github.com/rails/rails/commit/55bd348dfb47fcf6c2740d7bc4312da653ac404d

actionview/lib/action_view/template.rb
```diff
@@ -7,7 +7,7 @@ module ActionView
   class Template
     extend ActiveSupport::Autoload
 
-    STRICT_LOCALS_REGEX = /\#\s+locals:\s+\((.*)\)/
+    STRICT_LOCALS_REGEX = /\#\s+locals:\s+\((.*?)\)(?=\s*-?%>)/m
```
byroot: So this works for ERB, but I fear that some people were using it with HAML or Slim :/

Saidbek: Good point. The [fix](https://github.com/rails/rails/pull/56270/commits/90f0b28311554a0a59cf773628a0f76f0039a14b) handles it, so it works for:

ERB: `<%# locals: (arg:) -%>`
HAML: `-# locals: (arg:)\n`
Slim: `/ # locals: (arg:)\n`

Saidbek: @byroot the expectation on line 231 looks incorrect. If the default value is `'Hello)'`, i believe it should render as `"Hello)"`, not `"(Hello)"`, correct me if i'm wrong

PR #56278, opened 2025-12-03
avoid schema conflicts by dumping enum types with fully qualified name in PostgreSQL

agrobbin:
If there are multiple enum types with the same name, but in different schemas, the `:enum_type` option dumped as part of `db:schema:dump` will be ambiguous, potentially resulting in the wrong type being used when loading the schema.

Now, if the enum type is not already fully qualified, and multiple schemas are being dumped, the type includes the schema name.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
@@ -108,7 +108,17 @@ module ActiveRecord
               spec = { type: schema_type(column).inspect }.merge!(spec)
             end
 
-            spec[:enum_type] = column.sql_type.inspect if column.enum?
+            if column.enum?
+              # If only dumping one schema, or the enum type is already schema-qualified,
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
```diff
@@ -108,7 +108,17 @@ def prepare_column_options(column)
...
+              # If only dumping one schema, or the enum type is already schema-qualified,
+              # use the enum type as is. Otherwise, qualify it with the current schema name.
+              enum_type = if @dump_schemas.size == 1 || column.sql_type.include?(".")
+                column.sql_type
+              else
+                "#{schema_name}.#{column.sql_type}"
+              end
+
+              spec[:enum_type] = enum_type.inspect
+            end
```
fatkodima, line 121:
```suggestion
            spec[:enum_type] = relation_name(column.sql_type).inspect if column.enum?
```

fatkodima (changes requested):

PR #56279, opened 2025-12-03
Use pattern matching on sql instead of string match event payload name

zzak:
Follow up to #56225

I think string matching the event name could be brittle if the notification changes the format of that message this test will fail for no reason, so instead we can match on the SQL which is less-likely to change.

/cc @jeremy @rosa

### DIFF
--- a/activestorage/test/controllers/representations/redirect_controller_test.rb
+++ b/activestorage/test/controllers/representations/redirect_controller_test.rb
@@ -93,9 +93,11 @@ class ActiveStorage::Representations::RedirectControllerWithPreviewsTest < Actio
     variant_record_loaded_count = 0
 
     query_subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |event|
-      if event.payload[:name] == "ActiveStorage::VariantRecord Create"
+      case event.payload[:sql]
+      when /INSERT INTO "active_storage_variant_records"/
         variant_record_created = true
-      elsif event.payload[:name] == "ActiveStorage::VariantRecord Load" && variant_record_created
+      when /SELECT "active_storage_variant_records".* FROM "active_storage_variant_records"/
+        next unless variant_record_created
         variant_record_loaded_count += 1
       end
     end


rosa:
@zzak sounds good! In fact, I had this matching on the SQL initially 😅 I changed it because I thought the other approach looked nicer, but I agree, it's more brittle. 

Thank you!

activestorage/test/controllers/representations/redirect_controller_test.rb
```diff
@@ -93,10 +93,14 @@ class ActiveStorage::Representations::RedirectControllerWithPreviewsTest < Actio
     variant_record_loaded_count = 0
 
     query_subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |event|
-      if event.payload[:name] == "ActiveStorage::VariantRecord Create"
+      case event.payload[:sql]
+      in /INSERT INTO "active_storage_variant_records"/
```
fatkodima: Maybe simply `case`/`when`?

zzak: Ahh right :facepalm: fixed in 414a0b74c4, thank you!

activestorage/test/controllers/representations/redirect_controller_test.rb
```diff
@@ -93,9 +93,11 @@ class ActiveStorage::Representations::RedirectControllerWithPreviewsTest < Actio
     variant_record_loaded_count = 0
 
     query_subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |event|
-      if event.payload[:name] == "ActiveStorage::VariantRecord Create"
+      case event.payload[:sql]
+      when /INSERT INTO "active_storage_variant_records"/
         variant_record_created = true
-      elsif event.payload[:name] == "ActiveStorage::VariantRecord Load" && variant_record_created
+      when /SELECT "active_storage_variant_records".* FROM "active_storage_variant_records"/
```
skipkayhil, line 99: Does this work with databases that don't use double quotes for identifiers (MySQL)?

zzak: @skipkayhil Active Storage only ever uses `sqlite3::memory:` for testing: https://github.com/rails/rails/blob/main/activestorage/test/dummy/config/database.yml

So not sure portability is a concern here, wdyt?

PR #56282, opened 2025-12-03
Fix #56232: Clean output directory when generating guides to remove stale output files and add guide generation tests

harsh183:
### Motivation / Background

See issue https://github.com/rails/rails/issues/56232 for more details but basically we had some leftover old output files that would still be accessible and indexed by search engines.

### Detail

This PR clears the output dir before a new batch of output files are added. This should take care of stale output.

### Checklist

### DIFF
--- a/guides/rails_guides/generator.rb
+++ b/guides/rails_guides/generator.rb
@@ -18,7 +18,7 @@ module RailsGuides
   class Generator
     GUIDES_RE = /\.(?:erb|md)\z/
 
-    def initialize(edge:, version:, all:, only:, epub:, language:, direction: nil, lint:)
+    def initialize(edge:, version:, all:, only:, epub:, language:, direction: nil, lint:, guides_dir: nil)
       @edge         = edge
       @version      = version
       @all          = all
@@ -34,7 +34,7 @@ module RailsGuides
         register_special_mime_types
       end
 
-      initialize_dirs
+      initialize_dirs(guides_dir:)
       create_output_dir_if_needed if !dry_run?
       initialize_markdown_renderer
     end
@@ -82,8 +82,8 @@ module RailsGuides
         epub_filename << ".epub"
       end
 
-      def initialize_dirs
-        @guides_dir = File.expand_path("..", __dir__)
+      def initialize_dirs(guides_dir:)
+        @guides_dir = guides_dir || File.expand_path("..", __dir__)
 
         @source_dir  = "#{@guides_dir}/source"
         @source_dir += "/#{@language}" if @language
@@ -129,6 +129,7 @@ module RailsGuides
       end
 
       def cleanup_assets
+        FileUtils.rm_f(Dir.glob("#{@output_dir}/*.html"))
         FileUtils.rm_rf(Dir.glob("#{@output_dir}/{stylesheets,javascripts}"))
       end
 
--- a/guides/test/generator_test.rb
+++ b/guides/test/generator_test.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "fileutils"
+require "rails_guides/generator"
+require "tmpdir"
+
+class GeneratorTest < ActiveSupport::TestCase
+  def teardown
+    # Reset class-level attributes that Generator sets globally
+    RailsGuides::Markdown::Renderer.edge = nil
+    RailsGuides::Markdown::Renderer.version = nil
+  end
+
+  test "generate creates new files and removes stale html/css/js files" do
+    Dir.mktmpdir do |guides_dir|
+      output_dir = File.join(guides_dir, "output")
+
+      real_guides_dir = File.expand_path("..", __dir__)
+      FileUtils.mkdir_p(File.join(guides_dir, "source"))
+      %w[getting_started.md layout.html.erb _license.html.erb].each do |file|
+        FileUtils.cp(
+          File.join(real_guides_dir, "source", file),
+          File.join(guides_dir, "source", file)
+        )
+      end
+
+      assets_dir = File.join(guides_dir, "assets")
+      FileUtils.mkdir_p(File.join(assets_dir, "javascripts"))
+
+      style_dir = FileUtils.mkdir_p(File.join(assets_dir, "stylesrc"))
+      style_files = %w[style.scss highlight.scss print.scss]
+      style_files.each do |file|
+        File.write(File.join(style_dir, file), "")
+      end
+
+      FileUtils.mkdir_p(output_dir)
+      stale_html = File.join(output_dir, "stale_test_guide.html")
+      FileUtils.touch(stale_html)
+
+      css_dir = File.join(output_dir, "stylesheets")
+      FileUtils.mkdir_p(css_dir)
+      stale_css = File.join(css_dir, "old_stale_test.css")
+      FileUtils.touch(stale_css)
+
+      js_dir = File.join(output_dir, "javascripts")
+      FileUtils.mkdir_p(js_dir)
+      stale_js = File.join(js_dir, "old_stale_test.js")
+      FileUtils.touch(stale_js)
+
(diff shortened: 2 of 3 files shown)


guides/rails_guides/generator.rb
```diff
@@ -66,6 +67,10 @@ def dry_run?
         [@lint].any?
       end
 
+      def cleanup_output_dir
+        FileUtils.rm_rf(Dir.glob(File.join(@output_dir, "*"))) if Dir.exist?(@output_dir)
```
zzak: How about changing `create_output_dir_if_needed` to always clear the contents?
I don't think we want to completely remove the directory because it's being used as a symlink, but I'm also concerned that what if `@output_dir` here is nil or something crazy like `/`. Can we do this safely, maybe `cd @output_dir` first?

zzak: There is also a `cleanup_assets` method, but we really only care about `.html` files, right? so maybe making that more specific also reduces risk.

harsh183: yeah a more specific one is a better move, I'll try that in a bit.

guides/rails_guides/generator.rb
```diff
@@ -129,6 +129,7 @@ def select_only(guides)
       end
 
       def cleanup_assets
+        FileUtils.rm_rf(Dir.glob("#{@output_dir}/*.html"))
```
harsh183: @zzak I updated it to just delete the *.html pages based on
```rb
      def output_path_for(output_file)
        File.join(@output_dir, File.basename(output_file))
      end
```

zzak: It doesn't need to be recursive if we're globbing files

harsh183: Fixed that, good catch

tahsin352 (commented):
Thanks for the PR. Could you please update the Changelog and add a test case for the corresponding change. Also the build seems failing!

harsh183:
The build is failing on
```
Error:
ActiveSupport::Cache::RedisCacheStoreTests::ConnectionPoolBehaviorTest#test_pool_options_work:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /usr/local/bundle/gems/connection_pool-3.0.2/lib/connection_pool.rb:48:in 'initialize'
    lib/active_support/cache/redis_cache_store.rb:163:in 'Class#new'
    lib/active_support/cache/redis_cache_store.rb:163:in 'ActiveSupport::Cache::RedisCacheStore#initialize'
    lib/active_support/cache.rb:91:in 'Class#new'
    lib/active_support/cache.rb:91:in 'ActiveSupport::Cache.lookup_store'
    test/cache/stores/redis_cache_store_test.rb:316:in 'ActiveSupport::Cache::RedisCacheStoreTests::ConnectionPoolBehaviorTest#test_pool_options_work'
```

and I'm not sure if it's related to my change. Maybe I'll sync up to `main` and this can fix buildkite

harsh183:
I added to the changelog, and created the first ever tests for the generator scripts. Locally, my test does pass:

```
Finished in 0.274830s, 3.6386 runs/s, 25.4703 assertions/s.

1 runs, 7 assertions, 0 failures, 0 errors, 0 skips
```

---

 The build passed the first time I synced with `main` but now has multiple breaks that seem unrelated to change and might be flaky. Here are some of the breaks I'm seeing in the most recent build:
* failed to solve: ruby:4.0: docker.io/library/ruby:4.0: not found
*     /usr/local/bundle/gems/connection_pool-3.0.2/lib/connection_pool.rb:48:in 'initialize'
  - various ActiveSupport::Cache::RedisCacheStoreTests::FailureSafetyFromUnavailableClientTest:
ArgumentError: wrong number of arguments (given 1, expected 0)
  - Expected /ActionController::InvalidAuthenticityToken/ to match "/usr/local/bundle/gems/connection_pool-3.0.2/lib/connection_pool.rb:48 
* ApplicationTests::RakeTests::RakeDbsTest#test_destructive_tasks_are_protected:
ArgumentError: wrong number of arguments (given 1, expected 0)
    test/application/rake/dbs_test.rb:878:in 'block in <class:RakeDbsTest>'
* ApplicationTests::QueryLogsTest#test_controller_tags_are_not_doubled_up_if_already_configured:
ArgumentError: wrong number of arguments (given 1, expected 0)
    test/application/query_logs_test.rb:304:in 'ApplicationTests::QueryLogsTest#boot_app'

I'm not sure what I should do here, I can try to sync with master again but I might get more build flakes.

harsh183:
I got the build to run again and it's a different set of failing tests 😅

harsh183:
@zzak following up, is there anything else I can do here? I don't know what to do about the unrelated and potentially flaky rails build.

guides/test/generator_test.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "fileutils"
+require "rails_guides/generator"
+
+class GeneratorTest < ActiveSupport::TestCase
+  GUIDES_DIR = File.expand_path("..", __dir__)
+  OUTPUT_DIR = File.join(GUIDES_DIR, "output")
```
zzak: Why not use a tmpdir for testing? Just make sure to clean it up after.

harsh183: @zzak I just added it. I initially didn't do this since I saw `generator.rb` had hard coded a output directory so I added an override in the initializer to allow this. Let me know if this looks good to you.

guides/test/generator_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      assert Dir.exist?(css_dir), "Stylesheets directory should be recreated"
+      assert_not File.exist?(stale_css), "Stale CSS should be removed"
+
+      assert Dir.exist?(js_dir), "Javascripts directory should be recreated"
+      assert_not File.exist?(stale_js), "Stale JS should be removed"
+
+      assert File.exist?(keep_file), "Non-HTML files should be preserved"
+
+      getting_started_html = File.join(output_dir, "getting_started.html")
```
harsh183, line 76: since we're creating a bare scaffold of the rails guides outputs to test the stale removals, I figured we might as well lightly test if the other basic files are created

guides/rails_guides/generator.rb
```diff
@@ -18,7 +18,7 @@ module RailsGuides
   class Generator
     GUIDES_RE = /\.(?:erb|md)\z/
 
-    def initialize(edge:, version:, all:, only:, epub:, language:, direction: nil, lint:)
+    def initialize(edge:, version:, all:, only:, epub:, language:, direction: nil, lint:, guides_dir: nil)
```
harsh183, line 21: introduced this optional param solely for overriding it for test cases

harsh183:
I still have no idea on the broader build failures, locally my tests do pass for my change:
```
Finished in 0.605450s, 1.6517 runs/s, 16.5166 assertions/s.
1 runs, 10 assertions, 0 failures, 0 errors, 0 skips
```

harsh183:
After rebasing the code from `main` and fixing a variable pollution issue from the generate function, the build is finally passing. Lmk if I need to do anything else

harsh183:
@zzak just nudging on this PR. The build is green, I've addressed your comments and I think it should be good to go

eileencodes:
@harsh183 thank you for the PR, I'm merging this but will remove the changelog as it wasn't necessary.

@tahsin352 in the future - the changelog is for changes users need to know about, the guides are an internal tool used by the framework, therefore a changelog is not necessary here. In addition the failing tests were unrelated to this PR so it would have been fine for @harsh183 to ignore them. Please only ask contributors to fix tests they wrote or broke.

harsh183:
@eileencodes thank you for the review and the merge! I'll definitely keep the notes about the change log in mind for the future too

On a side note, I've really loved seeing your rails conference and alt Brighton ruby talks, and that inspired me to try to contribute to rails and join the discord chatroom for it

PR #56283, opened 2025-12-03
ActionText: Validate `RemoteImage` URLs

flavorjones:
### Motivation / Background

If an image is attached to rich text with a URL like `image.png`, the rendering pipeline will attempt to ask Propshaft to resolve it as a local asset, which will either then:

- raise `ActionView::Template::Error: The asset 'image.png' was not found in the load path.`
- or potentially resolve to something with the same name that is in the asset load path.

The first leads to a 500 error unless the application developer adds code to rescue the exception. The second is bad behavior at best, and a potential exploit vector at worst.

### Detail

RemoteImage.from_node now validates the URL before creating a RemoteImage object, using the same regex that AssetUrlHelper uses during rendering. URLs like "image.png" that would previously have been passed to the asset pipeline are rejected and fall through to MissingAttachable.

### Checklist

### DIFF
--- a/actiontext/CHANGELOG.md
+++ b/actiontext/CHANGELOG.md
@@ -1,2 +1,10 @@
+*   Validate `RemoteImage` URLs at creation time.
+
+    `RemoteImage.from_node` now validates the URL before creating a `RemoteImage` object, using the
+    same regex that `AssetUrlHelper` uses during rendering. URLs like "image.png" that would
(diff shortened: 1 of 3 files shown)


flavorjones:
The failing test seems unrelated (actioncable sauce labs errors).

seanpdoyle (approved):

guilleiguaran (approved):

PR #56288, opened 2025-12-04
Add SecureRandom.base32

monorkin:
### Motivation / Background

For one-time codes that are meant to be typed in by hand, we want to generate codes that are easy to read, easy to say out loud, and hard to confuse.

Crockford's Base32 gives us a case insensitive alphabet that avoids characters ambiguous to humans. This makes it a good fit for short codes that people can copy from an email into a form or read over the phone, such as two factor authentication codes.

Implementing a similar code generator is trivial in Ruby. Implementing it in a *cryptographically safe* way is less intuitive, as [I learned while working on Fizzy](https://github.com/basecamp/fizzy/pull/1868). By upstreaming this, I want to provide a helper that produces human-friendly codes that can safely be used for magic links or 2FA, so individual apps don’t need to re-solve the same problem.

### Detail

- Adds a new alphabet to `SecureRandom::BASE32_ALPHABET`
- Reuses the existing pattern for generating keys in different bases

Note: Base32 can mean different things, look at the Wikipedia article below, I went with `base32` because I saw that `base58` uses the Bitcoin version without specifically calling it out.

### Additional information

- [Crockford's Base32 definition](https://www.crockford.com/base32.html)
- [Base32 wikipedia article](https://en.wikipedia.org/wiki/Base32#Crockford's_Base32)

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Add `SecureRandom.base32` for generating case-insensitive keys that are unambiguous to humans.
+
+    *Stanko Krtalic Rusendic & Miha Rekar*
+
 *   Add a fast failure mode to `ActiveSupport::ContinuousIntegration` that stops the rest of
     the run after a step fails. Invoke by running `bin/ci --fail-fast` or `bin/ci -f`.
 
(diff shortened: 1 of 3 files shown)


monorkin:
Fixed the changelog check, the ActionCable tests are unrelated to this change from what I can tell.

jeremy (approved):
Note to ourselves that we'll want to move our SecureRandom extensions over to [Random::Formatter](https://docs.ruby-lang.org/en/master/Random/Formatter.html).

PR #56293, opened 2025-12-04
Combine rather than overwrite tag content supplied via both parameter and block

dhh:
Before this, we'd silently just overwrite the parameter content with the block content, so `tag.div("Hello ") { "World" }` would just return `<div>World</div>`. With this fix, it's going to return `<div>Hello World</div>`.


zzak:
@dhh Should this be backported?

dhh:
Hesitant to backport as there's a theoretical chance that someone might have a buggy call that changes behavior in their app. That's of course always a possibility when you fix bugs, but I think in this case it's better to wait for next 8.2/9.0.

PR #56292, opened 2025-12-04
Fix `RedisCacheStore` for `connection_pool` >= 3

fatkodima:
Fixes #56291.

PR #56294, opened 2025-12-05
Fix `MemCacheStore` for `connection_pool` >= 3

fatkodima:
Follow up to #56292 (forgot about mem_cache_store 😞).

### DIFF
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -59,7 +59,7 @@ module ActiveSupport
         pool_options = retrieve_pool_options(options)
 
         if pool_options
-          ConnectionPool.new(pool_options) { Dalli::Client.new(addresses, options.merge(threadsafe: false)) }
+          ConnectionPool.new(**pool_options) { Dalli::Client.new(addresses, options.merge(threadsafe: false)) }
         else
           Dalli::Client.new(addresses, options)
         end


Aqualon:
@fatkodima will this be released in all supported versions of rails or should the older versions get a release with connection_pool pinned down to '< 3'?

fatkodima:
This will be released in all supported rails versions soon.

beetlegius-somos:
> This will be released in all supported rails versions soon.

how soon? :P @fatkodima

davidesantangelo:
When will the new version be released?

fatkodima:
No one knows (or at least I don't know people who know). Asking it multiple times will not make it quicker. 
You can point to the stable branch via `gem "rails", github: "rails/rails", branch: "8-0-stable"`.

afdev82:
You are right @fatkodima , but I have done as suggested and it doesn't work.

I think because bundler still picks up the `activesupport` gem using the version (8.0.4) and not the HEAD of the branch.
I have verified it using the `bundle open activesupport` command and in fact the fix is not there, even if the commit for the `rails` gem is the latest of the current `8-0-stable` branch.
How could I solve this?
Thank you for your support.

> No one knows (or at least I don't know people who know). Asking it multiple times will not make it quicker. You can point to the stable branch via `gem "rails", github: "rails/rails", branch: "8-0-stable"`.

brendon:
The latest Rails `8.1.2` has the fix.

afdev82:
Yes, I know, thank you.  
Unfortunately, I can't easily update to Rails 8.1.x because of one dependency.  
I have replied only because I am waiting for the fix for the 8.0.x branch and the trick to pin the github branch instead of the version didn't work.

> The latest Rails `8.1.2` has the fix.

mlarraz:
FYI this could use a backport to Rails 7.2 as well since iirc it's still supported

skipkayhil:
> since iirc it's still supported

7.2 is security fixes only, per the [maintenance policy](https://rubyonrails.org/maintenance)

brendon:
Is there a case for then just restricting the dependency in a 7.2 release so that it doesn't just break? I'd assume the Rails test suite would break if they did another patch release of 7.2 from now on without fixing this dependency?

Sorry @afdev82, I misunderstood you :)

afdev82:
@brendon no problem.
In the meanwhile, waiting for the patch release of the 8.0, I am simply forcing connection_pool on v2 in my bundle.
But in my opinion, it's a quite urgent fix, since the v8.0.4 and others are now broken with the release of connection_pool v3, anyway forcing v2 is quite easy (if other dependencies are still compatible too).

PR #56295, opened 2025-12-05
Make becomes preserve association cache

### LINKED ISSUE #51217: `ActiveRecord::Persistence#becomes` does not clone eager-loading state and can raise Strict Loading violations

bogdan:
### Motivation / Background

This Pull Request has been created because it fixes #51217 

### Detail

This Pull Request changes `becomes` to preserve association cache from original model.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/persistence.rb
+++ b/activerecord/lib/active_record/persistence.rb
@@ -500,6 +500,12 @@ module ActiveRecord
         becoming.errors.copy!(errors)
       end
 
+      @association_cache.each do |name, assoc|
+        next unless assoc.loaded? && assoc.reflection.strict_loading?
+        next unless assoc.reflection.equal?(klass.reflect_on_association(name))
+        became.association(name).target = assoc.target
+      end
+
       became
     end
 
--- a/activerecord/test/cases/persistence_test.rb
+++ b/activerecord/test/cases/persistence_test.rb
@@ -542,6 +542,20 @@ class PersistenceTest < ActiveRecord::TestCase
     assert_not_equal original.object_id, clone.object_id
   end
 
+  def test_becomes_keeps_strict_loading_association_cache
+    ship = Ship.create!(name: "Endurance", developer: Developer.first)
+    developer = Developer.preload(:strict_loading_ship).find(ship.developer_id)
+    sub = developer.becomes(SubDeveloper)
+    assert_predicate sub.association(:strict_loading_ship), :loaded?
+    assert_equal ship, sub.strict_loading_ship
+  end
+
+  def test_becomes_does_not_copy_non_strict_loading_association_cache
+    firm = Firm.preload(:account).first!
+    agency = firm.becomes(Agency)
+    assert_not_predicate agency.association(:account), :loaded?
+  end
+
   def test_delete_many
     original_count = Topic.count
     Topic.delete(deleting = [1, 2])


seanpdoyle:
I can confirm that this resolves https://github.com/rails/rails/issues/51217. Thank you!

Does copying the association cache affect other behaviors? Would it be worthwhile to document this change with an Active Record CHANGELOG entry?

bogdan:
I can confirm that this resolves https://github.com/rails/rails/issues/51217. Thank you!

> Does copying the association cache affect other behaviors? 

Hard to say. On one end: it should be behave like clone. On other end, imagine there are differences in associations of target class and source class. I think if accidentally both classes will have an association with same name, that can cause the result instance association to become broken, preloaded incorrectly etc.  

Alternative way of fixing it is to reuse this code: https://github.com/rails/rails/blob/75f0efa212444e7da6dce071012bc6f9fb771dfe/activerecord/lib/active_record/persistence.rb#L862
It specifically targets fixing `strict_loading` option when new instance of a class is required.


> Would it be worthwhile to document this change with an Active Record CHANGELOG entry?

I don't feel like I have the understanding to decide that.

genezys:
Maybe @byroot for a review?

byroot:
A bit on the fence about this one. Do we even know if the new class even have these associations? Perhaps it's not a big deal.

However at the very least I think we should shallow copy the cache. (even though interestingly `becomes` seems to share `@attributes`?, I don't think I ever used it, so I'm uncertain about the desired semantic, but this seems weird/dangerous to me).

bogdan:
Agree with @byroot.

I made a more conservative approach by taking association `target` from `association_cache` only for loaded associations with `strict_loading` flag and only when association reflection is the same (by `object_id`). This makes sure that `Association` instance remains distinct between source and destination instance in `become`.

Side note:

I tried to solve problem by setting `strict_loading!(false)` on the created instance, but was surprised that `strict_loading: true` association option takes precedence on instance level config. I never used strict loading feature myself, but that seemed confusing because usually instance level configurations take precedence on class level configurations.

PR #38116, opened 2019-12-28
Fix regression where AJ#perform_now doesn't return the job's value:

Edouard-chin:
This PR contains 2 commits that aren't related to each other but they were both making the wrong assumption (I initially thought that a job was either returning `true` / `false`.

cc/ @rafaelfranca

PR #56296, opened 2025-12-05
Remove unnecessary defined? check in Active Job instrumentation

shivamsinghchahar:
### Motivation / Background

Now that we dropped support for Ruby 2.7, we no longer need to check if variables are defined before accessing them to avoid the undefined variable warning.

#### Before
```rb
ActiveSupport::Notifications.instrument("#{operation}.active_job", payload) do |payload|
  value = block.call(payload) if block
  payload[:aborted] = @_halted_callback_hook_called if defined?(@_halted_callback_hook_called)
  @_halted_callback_hook_called = nil
  value
end
```
#### After
```rb
ActiveSupport::Notifications.instrument("#{operation}.active_job", payload) do |payload|
  value = block.call(payload) if block
  payload[:aborted] = true if @_halted_callback_hook_called
  @_halted_callback_hook_called = nil
  value
end
```

### Additional information

Similar to changes in https://github.com/rails/rails/pull/50597

### Checklist

### DIFF
--- a/activejob/lib/active_job/instrumentation.rb
+++ b/activejob/lib/active_job/instrumentation.rb
@@ -32,7 +32,7 @@ module ActiveJob
 
       ActiveSupport::Notifications.instrument("#{operation}.active_job", payload) do |payload|
         value = block.call(payload) if block
-        payload[:aborted] = @_halted_callback_hook_called if defined?(@_halted_callback_hook_called)
+        payload[:aborted] = true if @_halted_callback_hook_called
         @_halted_callback_hook_called = nil
         value
       end


activejob/lib/active_job/instrumentation.rb
```diff
@@ -32,7 +32,7 @@ def instrument(operation, payload = {}, &block) # :nodoc:
 
       ActiveSupport::Notifications.instrument("#{operation}.active_job", payload) do |payload|
         value = block.call(payload) if block
-        payload[:aborted] = @_halted_callback_hook_called if defined?(@_halted_callback_hook_called)
+        payload[:aborted] = @_halted_callback_hook_called
```
byroot: The original code checked if the ivar was true: https://github.com/rails/rails/pull/38116/files#diff-41ecb8d7126485e207a879c9e3e28f73837309b8370d0db939a23874b97f66b1R22

It somehow was lost without justification in https://github.com/rails/rails/commit/9574a39a7acf38453d08a31f56a9797c7e417e94

shivamsinghchahar: @byroot Thanks for catching this! I've update the code to
```rb
payload[:aborted] = true if @_halted_callback_hook_called
```
This removes the unnecessary `defined?` check while preserving the intent that `:aborted` should only be set to true when the callback chain is actually halted.

Please let me know if this is alright

PR #56297, opened 2025-12-05
Wrap console command with an executor by default

### LINKED ISSUE #56277: ActiveRecord: async query methods not working in console
### Steps to reproduce
1. Create new rails app `rails new testing` or use application with rails version `>= 8.0` .
2. Create new model `rails g model User first_name:string` and migrate `rails db:migrate`
3. Go to `config/application.rb` or `config/development.rb` and set async_query_executor `config.active_record.async_query_executor = :global_thread_pool`
4. Run console and try run `User.limit(1).load_async`

### Expected behavior

Perform query and load user from a database.

### Actual behavior

Raises and exception `Can't perform asynchronous queries without a query session (ActiveRecord::ActiveRecordError)`

### System configuration

Rails version 8.1
Ruby version 3.4

### More info

I do not see this bug in rails 7.1 but It's present in rails 8.0 and 8.1.
Other queries like `async_pluck` etc. not working either.

### Workaround

In console run `ActiveRecord::Base.asynchronous_queries_tracker.start_session` - and async queries will start working.

zzak:
This can be disabled with `-w` or `--skip_executor`, same as runner.

Fixes #56277

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Wrap console command with an executor by default
+
+    This can be disabled with `-w` or `--skip_executor`, same as runner.
+
+    *zzak*
+
 *   Add a new internal route in development to respond to chromium devtools GET request.
 
     This allows the app folder to be easily connected as a workspace in chromium-based browsers.
--- a/railties/lib/rails/commands/console/console_command.rb
+++ b/railties/lib/rails/commands/console/console_command.rb
@@ -68,6 +68,8 @@ module Rails
       class_option :sandbox, aliases: "-s", type: :boolean, default: nil,
         desc: "Rollback database modifications on exit."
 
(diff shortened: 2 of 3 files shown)


ghiculescu:
This is a pretty surprising breaking change. I don't think most console users would consider a console session to be a equivalent to a request. Would you reconsider this?

byroot:
> Would you reconsider this?

Not until someone actually explain what the problem is. So far I've seen a few people say "breaking change", but no one bothered to explain what they meant, so I'm reading it as "I don't like change".

Now if someone actually bother making a case, explaining what's problematic, then yes, I'm absolutely open to changes.
Likely not to remove the executor, but disabling the query cache by default surely.

Personally I think it's a good change, because it actually allow you to test how a snippet of code would behave in a controller.

bensheldon:
(I replied to byroot's question being crossposted over here: https://github.com/rails/rails/issues/56473#issuecomment-3790166517)

PR #56300, opened 2025-12-05
Fix typos in JavaScript guide

Saidbek:
### Motivation / Background

- Improve formatting for raw file readability
- Fixed async js code example

### Checklist

### DIFF
--- a/guides/source/working_with_javascript_in_rails.md
+++ b/guides/source/working_with_javascript_in_rails.md
@@ -339,13 +339,13 @@ import { FetchRequest } from '@rails/request.js'
 
 ....
 
-async myMethod () {
-  const request = new FetchRequest('post', 'localhost:3000/posts', {
+async function myMethod () {
+  const request = new FetchRequest('post', 'http://localhost:3000/posts', {
     body: JSON.stringify({ name: 'Request.JS' })
   })
   const response = await request.perform()
   if (response.ok) {
-    const body = await response.text()
+    const body = await response.text
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


guides/source/working_with_javascript_in_rails.md
```diff
@@ -63,7 +63,8 @@ import ReactDOM from "react-dom"
 ```
 
 Adding npm Packages with JavaScript Bundlers
---------
+--------------------------------------------
```
gmcgibbon, line 66: This doesn't affect the output of the markdown, but it does look better when viewing the raw file, so this is fine. I wouldn't mention it as a fix though.

guides/source/working_with_javascript_in_rails.md
```diff
@@ -344,7 +345,7 @@ async myMethod () {
   })
   const response = await request.perform()
   if (response.ok) {
-    const body = await response.text
+    const body = await response.text()
```
gmcgibbon: This is actually incorrect, it is a property. But the function is also still wrong. Use this instead:
```js
async function myMethod () {
  const request = new FetchRequest('post', 'http://localhost:3000/posts', {
    body: JSON.stringify({ name: 'Request.JS' })
  })
  const response = await request.perform()
  if (response.ok) {
    const body = await response.text
  }
}
```

gmcgibbon (changes requested):

guides/source/working_with_javascript_in_rails.md
Saidbek: done

gmcgibbon (approved):
Thanks!

PR #56302, opened 2025-12-06
Dispatch `direct-upload:success` event with `detail.hiddenInput`

seanpdoyle:
### Motivation / Background

The issue
---

The [direct-upload:end][] event is dispatched when a `DirectUploadController` finishes a file upload. At that point, the upload has either succeeded or failed. On success, the `<form>` element has a new `<input type="hidden">` element with the signed Blob response from the Rails server. However, listeners cannot access the signed value reliably, nor can they access other metadata from the response.

### Detail

The proposal
---

Introduce the `direct-upload:success` event dispatched after the final `direct-upload:progress` and before `direct-upload:end`. The `event.detail` object has two properties:

* `hiddenInput` the `<input type="hidden">` element that stores the Blob's signed id in its `[value]` and `.value` property
* `blob` the entire Blob metadata response object

[direct-upload:end]: https://guides.rubyonrails.org/active_storage_overview.html#direct-uploads

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Dispatch `direct-upload:success` with `blob` and `hiddenInput`
+
+    The `direct-upload:success` event is dispatched after the final
+    `direct-upload:progress` and before `direct-upload:end`. The `blob` is the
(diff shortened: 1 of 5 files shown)


joelhawksley (commented):
This looks good to me 👍🏻

PR #56308, opened 2025-12-06
Fix update, insert, and delete to raise IrreversibleMigration on roll…

### LINKED ISSUE #51570: [ActiveRecord] migrations: `update` statements are silently being executed in both directions
I recently stumbled across a weird behavior in ActiveRecord migrations that appears to be a bug: `update` statements within the `change`   method are being **silently executed** during the rollback phase when they  are neither wrapped in an `up_only` nor a `reversible` block. For example:

```rb
class ResetUsersUpdatedAt < ActiveRecord::Migration[7.1]
  
  def change
    update(<<~SQL.squish)
      UPDATE users
      SET updated_at = NOW()
    SQL
  end
  
end
```

When I run this migration upwards, it behaves as expected. `db:migrate` yields the following output:
```
== 20240411074748 ResetUsersUpdatedAt: migrating ==========================
-- execute("UPDATE users SET updated_at = NOW()")
   -> 0.0007s
== 20240411074748 ResetUsersUpdatedAt: migrated (0.0188s) =================
```

However, the output of `db:rollback` is empty:
```
== 20240411074748 ResetUsersUpdatedAt: reverting ==========================
== 20240411074748 ResetUsersUpdatedAt: reverted (0.0680s) =================
```

A quick glance at the `updated_at` column reveals that the `update` statement was being silently executed in both direction.
My expectation would have been that it would raise an `ActiveRecord::IrreversibleMigration` just like `execute` – or otherwise print the executed statement to STDOUT.

### Steps to reproduce
Run the code below:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

  create_table :payments, force: true do |t|
    t.text :migration_type
  end
end

class Payment < ActiveRecord::Base
end

class SetMigrationTypeWithUpdate < ActiveRecord::Migration::Current
  def change
    migration_type = reverting? ? 'down' : 'up'
    update(<<~SQL.squish)
      UPDATE payments
      SET migration_type = "#{migration_type}"
    SQL
  end
end

class SetMigrationTypeWithExecute < ActiveRecord::Migration::Current
  def change
    migration_type = reverting? ? 'down' : 'up'
    execute(<<~SQL.squish)
      UPDATE payments
      SET migration_type = "#{migration_type}"
    SQL
  end
end

  def test_update_migration_up
    Payment.create!(migration_type: 'none')
    SetMigrationTypeWithUpdate.migrate(:up)

    assert_equal "up", Payment.last!.migration_type
  end

  # This test case is red and yields two unexpected assertions:
  # * no exception is being raised
  # * the update statement was executed (the value was changed to "down")
  #
  # And on top of that, no output to STDOUT was being made so it happened silently..
  def test_update_migration_down
    assert_raises ActiveRecord::IrreversibleMigration do
      SetMigrationTypeWithUpdate.migrate(:down)
      assert_equal "up", Payment.last!.migration_type
    end
  end

  def test_execute_migration_up
    Payment.create!(migration_type: 'none')
    SetMigrationTypeWithExecute.migrate(:up)

    assert_equal "up", Payment.last!.migration_type
  end

  def test_execute_migration_down
    assert_raises ActiveRecord::IrreversibleMigration do
      SetMigrationTypeWithExecute.migrate(:down)
      assert_equal "up", Payment.last!.migration_type
    end
  end
end
```

### Expected behavior
* The test case `test_update_migration_down` should be green.
* If I use `update` within `def change` and try to roll back that migration, it should raise an `ActiveRecord::IrreversibleMigration` exception
* Any SQL statement that is being executed during the rollback phase is being printed to the STDOUT

### Actual behavior
* The test case `test_update_migration_down` is red
* No exception is being made
* The rollback phase does not print the "silent" update statement to the STDOUT

### System configuration
**Rails version**:  7.1.3.2
**Ruby version**: 3.3.0

Saidbek:
Fixes #51570

### Motivation / Background

`update`, `insert`, and `delete` statements in migrations were silently being executed during rollback when used in a change method. This is inconsistent with `execute`, which correctly `raises ActiveRecord::IrreversibleMigration`.

For example:
```ruby
class ResetUsersUpdatedAt < ActiveRecord::Migration[8.0]
  def change
    update("UPDATE users SET updated_at = NOW()")
  end
end
```
When rolling back this migration, the `UPDATE` statement would execute silently (potentially corrupting data) instead of raising an error.

### Detail

This PR adds `:update`, `:insert`, and `:delete` to `ReversibleAndIrreversibleMethods` in `CommandRecorder`.
This ensures these methods are recorded by the `CommandRecorder` and checked for reversibility when rolling back. Since they have no inverse operation defined, they now correctly raise `ActiveRecord::IrreversibleMigration` on rollback.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -2,6 +2,13 @@
 
     *Ngan Pham*
 
+*   Fix `update`, `insert`, and `delete` in migrations to raise `ActiveRecord::IrreversibleMigration` on rollback.
+
+    Previously, these methods would silently execute during rollback instead of
+    raising an error like `execute` does.
+
+    *Said Kaldybaev*
+
 *   Fix bug when `current_transaction.isolation` would not have been reset in test env.
 
     Additionally, extending the change in [#55549](https://github.com/rails/rails/pull/55549)
--- a/activerecord/lib/active_record/migration/command_recorder.rb
+++ b/activerecord/lib/active_record/migration/command_recorder.rb
@@ -61,7 +61,8 @@ module ActiveRecord
         :create_enum, :drop_enum, :rename_enum, :add_enum_value, :rename_enum_value,
         :create_schema, :drop_schema,
         :create_virtual_table, :drop_virtual_table,
-        :enable_index, :disable_index
+        :enable_index, :disable_index,
+        :update, :insert, :delete
       ]
       include JoinTable
 
--- a/activerecord/test/cases/migration/command_recorder_test.rb
+++ b/activerecord/test/cases/migration/command_recorder_test.rb
@@ -54,6 +54,42 @@ module ActiveRecord
         end
       end
 
+      def test_inverse_of_raise_exception_on_update
+        assert_raises(ActiveRecord::IrreversibleMigration) do
+          @recorder.inverse_of :update, ["UPDATE users SET name = 'test'"]
+        end
+      end
+
+      def test_inverse_of_raise_exception_on_insert
+        assert_raises(ActiveRecord::IrreversibleMigration) do
+          @recorder.inverse_of :insert, ["INSERT INTO users (name) VALUES ('test')"]
+        end
+      end
+
+      def test_inverse_of_raise_exception_on_delete
+        assert_raises(ActiveRecord::IrreversibleMigration) do
+          @recorder.inverse_of :delete, ["DELETE FROM users WHERE id = 1"]
+        end
+      end
+
+      def test_irreversible_update_raises_exception
+        assert_raises(ActiveRecord::IrreversibleMigration) do
+          @recorder.revert { @recorder.update "UPDATE users SET name = 'test'" }
+        end
+      end
+
+      def test_irreversible_insert_raises_exception
+        assert_raises(ActiveRecord::IrreversibleMigration) do
+          @recorder.revert { @recorder.insert "INSERT INTO users (name) VALUES ('test')" }
+        end
+      end
+
+      def test_irreversible_delete_raises_exception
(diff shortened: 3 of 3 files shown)


fatkodima:
What about 
```ruby
class MyMigration < ActiveRecord::Migration[8.0]
  def change
    select_value("drop database my_database")
  end
end
```
?

makmic:
Hi,
thanks for looking into this issue.

The patch looks fine to me on the first glance, but I don't know much about the underlying code. One thing that I asked myself is if it breaks update statements that are within a [reversible](https://guides.rubyonrails.org/active_record_migrations.html#using-reversible) or [up_only](https://api.rubyonrails.org/classes/ActiveRecord/Migration.html#method-i-up_only) block. We should make sure this case is tested, it should not raise in this case.

Saidbek:
@makmic I just had a quick look into the tests and this is already covered. The fix doesn't break `update` inside `reversible` or `up_only` blocks because both are skipped during the rollback. [up_only](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/migration.rb#L934) and [reversible](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/migration.rb#L915)

Also, I've found an existing test in [invertible_migration_test](https://github.com/rails/rails/blob/main/activerecord/test/cases/invertible_migration_test.rb#L565) which already verifies it and confirms rollbacks works without raising an error.

fatkodima:
Unfortunately, this does not cover all the possible cases (see my comment above) and it is not even possible to cover them.

PR #56311, opened 2025-12-06
Deprecate PreviewImageJob

p8:
PreviewImageJob is no longer called by Rails but it is a documented
class making it public API that requires deprecation.

Another option is to just delete it but maybe there are code bases that subclass it or call it in tests.

The preprocessed method no longer exist so inline the original implementation instead.

This was missed in #51951.

### Checklist

### DIFF
--- a/activestorage/app/jobs/active_storage/preview_image_job.rb
+++ b/activestorage/app/jobs/active_storage/preview_image_job.rb
@@ -6,11 +6,20 @@ class ActiveStorage::PreviewImageJob < ActiveStorage::BaseJob
   discard_on ActiveRecord::RecordNotFound, ActiveStorage::UnrepresentableError
   retry_on ActiveStorage::IntegrityError, attempts: 10, wait: :polynomially_longer
 
+  def initialize(*arguments)
+    ActiveStorage.deprecator.warn(<<~MSG.squish)
+      ActiveStorage::PreviewImageJob is no longer used by Rails.
+      It is deprecated and will be removed in Rails 9.0.
+      Use the ActiveStorage::CreateVariantsJob instead.
+    MSG
+    super
+  end
+
   def perform(blob, variations)
     blob.preview({}).processed
 
     variations.each do |transformations|
-      blob.preprocessed(transformations)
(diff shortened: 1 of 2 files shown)


activestorage/app/jobs/active_storage/preview_image_job.rb
```diff
@@ -6,11 +6,20 @@ class ActiveStorage::PreviewImageJob < ActiveStorage::BaseJob
...
   retry_on ActiveStorage::IntegrityError, attempts: 10, wait: :polynomially_longer
 
+  def initialize(*arguments)
+    ActiveStorage.deprecator.warn(<<~MSG.squish)
+      ActiveStorage::PreviewImageJob is no longer used by Rails.
+      It is deprecated and will be removed in Rails 9.0.
+      Use the ActiveStorage::CreateVariantsJob instead.
+    MSG
+    super
+  end
```
byroot, line 16: Wouldn't it be simpler to use `deprecate_constant` ?

activestorage/test/jobs/preview_image_job_test.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "database/setup"
+
+class ActiveStorage::PreviewImageJobTest < ActiveJob::TestCase
```
byroot, line 6: I don't think it's worth adding tests for something we're about to remove.

p8: This restores the test removed in https://github.com/rails/rails/pull/51951.
Running the tests showed that PR also breaks the PreviewImageJob as `Blob#preprocessed` was removed, hence the change in PreviewImageJob.

activestorage/app/jobs/active_storage/preview_image_job.rb
p8: I tried using that to use with the new constant being `ActiveStorage::CreateVariantsJob`.
However, the signature of `ActiveStorage::CreateVariantsJob#perform` is slightly different (it also takes a `variants:` argument), which would still be a breaking change I think.

PR #56312, opened 2025-12-06
Add setting for requiring explicit SELECT column names

Saidbek:
### Motivation / Background

Add a new configuration setting `ActiveRecord::Base.require_explicit_select`. When truthy, ActiveRecord will raise an `ActiveRecord::ExplicitSelectRequired` error when a query is executed without an explicit select clause, forcing developers to be explicit about which columns to fetch.

### Detail

Using SELECT * can cause several issues in production applications:
- Increased memory usage by loading unnecessary data into Ruby objects
- Slower queries by transferring mode data than needed from the database

Currently, there's no way easy way to enforece explicit `select` usage accross the app. Teams that care about this end up:
- Adding custom rubocop rules to lint for missing `select` calls
- Doing manual code reviews to catch `select *` queries
- Building custom query wrappers that require column specs

Adding a config option to enforce this behaviour all the time seems cleaner, and makes it easier to opt-in on per-model or per-environment basis.
```ruby
# config/application.rb
module MyApp
    config.active_record.require_explicit_select = true
  end
end
```
Or on a per-model basis:
```ruby
# app/models/user.rb
  self.require_explicit_select = true
end
```

Note: This does not affect aggregate queries like `count`, `sum`, `pluck`, `exists`?, `ids`, etc. as they already specify their own columns internally.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,32 @@
+*   Add `ActiveRecord::Base.require_explicit_select` configuration option.
+
+    When set to `true`, raises `ActiveRecord::ExplicitSelectRequired` when a query
+    is executed without an explicit `select` clause. This encourages developers to
+    be explicit about which columns to fetch, improving performance by reducing
+    memory usage and data transfer.
+
+    ```ruby
+    # config/application.rb
+    config.active_record.require_explicit_select = true
+
+    # Or per-model
+    class User < ApplicationRecord
+      self.require_explicit_select = true
+    end
+
+    User.all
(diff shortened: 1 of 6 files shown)


fatkodima:
My 2 cents: using explicit SELECT is not needed in 99.99% of cases and will not make a performance difference and enforcing it for all the queries will be just annoying to the developers and make the code much verbose and uglier. It is better to use it only where it is needed (identified and confirmed via profiling that it will make a difference).

I am not sure if this will be merged, but I suggest implementing it as a custom rails extension in the app or a new gem.

PR #56314, opened 2025-12-07
Fix `rails new` swallowing `bundle` errors

strugee:
### Motivation / Background

Fixes #56313

### Detail

I think the diff is pretty self-explanatory when paired with the Ruby docs: https://ruby-doc.org/3.4.1/Kernel.html#method-i-system

### Additional information

Not sure if this needs a test, given that it's... somewhat hard to programmatically reproduce? I can dive into the test suite if someone decides this is required.

Also, please review harshly... I hit this bug _literally_ while running the Rails tutorial. And I don't really know Ruby.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/bundle_helper.rb
+++ b/railties/lib/rails/generators/bundle_helper.rb
@@ -24,9 +24,9 @@ module Rails
         def exec_bundle_command(bundle_command, command, env, params)
           full_command = %Q["#{Gem.ruby}" "#{bundle_command}" #{command}]
           if options[:quiet] || params[:quiet]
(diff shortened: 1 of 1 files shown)


strugee:
I realized that this is _in theory_ decently easy to test by just invoking `bundle_command "exec false"` - no need to test that `bundle install` errors specifically fail. However I just spent the last hour and a half trying to make this work, and the `BundleHelper` implementation is all tied up with Thor in ways that I don't understand.

Again, let me know if a test is actually required here. (Presumably yes, given that the `exec false` path is viable? But I might need some help.)

PR #56315, opened 2025-12-07
gitignore `/vendor`

strugee:
### Motivation / Background

Bundler is populating `vendor/cache` on my system, and obviously it's wrong for git to be looking at these cache files.

That being said: this seems pretty obvious, which makes me think it might be just a misconfiguration on my end? `type bundle` points to some `rbenv` shim binary, and `rbenv version` returns `system` so this might be related. (I set this up a while ago and am not 100% confident I remember how it's configured.)

If y'all think this is just a me problem, possibly for using system Ruby, feel free to close.

### Detail

N/A

### Additional information

N/A

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /*/doc/
 /*/test/tmp/
 /.bundle
+/vendor
 /dist/
 /doc/
 /guides/output/


byroot:
Yes, bundler doesn't install gems there by default, if it does it's because you configured it to do so.

Also the whole concept of "vendoring" is to add your dependencies in your own repo, so adding `vendor` to `gitignore` makes zero sense.

strugee:
I promise you that I did not _actively_ configure Bundler to install to `/vendor`. It also is only Gem cache files (`cache` is the only file or directory in `vendor`), the actual bundle seems to be installed somewhere else. Could just add `/vendor/cache` to `gitignore`.

I am aware of what the concept of vendoring is, but clearly `vendor` isn't being used that way in rails.git.

PR #56319, opened 2025-12-07
Fix preload for scoped has_one associations with LIMIT

drymar:
### Motivation / Background

This Pull Request addresses issue: https://github.com/rails/rails/issues/56276

Rails currently applies `LIMIT` incorrectly when preloading a `has_one` association where the scope contains `.limit(1)`. Because `preload` performs a single query for all parent records, the LIMIT is applied globally rather than per parent record.

Example:
```ruby
  has_one :last_payment, -> { order(created_on: :desc).limit(1) }
end

Account.preload(:last_payment)
```

Generates SQL similar to:
```sql
SELECT * FROM payments
WHERE account_id IN (1, 2)
ORDER BY created_on DESC
LIMIT 1
```

Only one associated record is loaded for all accounts, causing incorrect results(one account receives `nil`).

### Detail

This PR fixes the issue by stripping the LIMIT clause from the preload scope for has_one (non-collection) associations.

  The key insight is that for `has_one` associations, the "one record per parent" filtering is already handled in Ruby code within `Preloader::Association#load_records`, which only assigns the first matching record to each owner. The `LIMIT` in the scope is therefore redundant for preloading and harmful when applied globally.

The fix(in `preloader/association.rb`):
```ruby
scope.limit!(nil) if !reflection.collection? && scope.limit_value
```

This approach:
  - Preserves `preload`'s efficient batch loading (single query for all parents)
  - Removes only the problematic global `LIMIT`
  - Lets existing Ruby code handle per-parent record assignment
  - Does not affect `has_many` or other collection associations

Additional Information
  - This change does not modify behavior of `has_many`, `belongs_to`, or `has_one` associations without `LIMIT` in scope
  - The fix applies to both preload and includes (when using preload strategy)
  - eager_load was already working correctly (uses JOINs, not affected by this issue)
  - The fix is backward-compatible and minimal

### Checklist

* [x] No CHANGELOG is needed, this is a small bug fix.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix `preload` and `includes` applying `LIMIT` globally for `has_one` associations with `limit` in scope.
+
+    When preloading a `has_one` association defined with a `limit` scope like
+    `has_one :last_comment, -> { order(id: :desc).limit(1) }`, the `LIMIT` was
+    incorrectly applied to the entire batch query instead of per-parent-record.
+    This caused some parent records to incorrectly return `nil` for the association.
+
+    *Dmytro Rymar*
+
 *   Deprecate the `schema_order` option in PostgreSQL database configurations.
 
     Use `schema_search_path` instead. The `schema_order` alias will be
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -304,6 +304,8 @@ module ActiveRecord
               scope.merge!(preload_scope)
             end
 
+            scope.limit!(nil) if !reflection.collection? && scope.limit_value
+
             cascade_strict_loading(scope)
           end
 
--- a/activerecord/test/cases/associations/preload_has_one_limit_test.rb
+++ b/activerecord/test/cases/associations/preload_has_one_limit_test.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require "cases/helper"
+require "models/post"
+require "models/comment"
+
+class PreloadHasOneLimitTest < ActiveRecord::TestCase
+  fixtures :posts, :comments
+
+  def test_preload_has_one_with_limit_returns_correct_record_per_parent
+    posts = Post.where(id: [posts(:welcome).id, posts(:thinking).id])
+                .order(:id)
+                .preload(:limited_last_comment)
+                .to_a
+
+    assert_equal comments(:more_greetings), posts[0].limited_last_comment
+    assert_equal comments(:does_it_hurt), posts[1].limited_last_comment
+  end
+
+  def test_preload_has_one_with_limit_does_not_cause_n_plus_one
+    posts = Post.where(id: [posts(:welcome).id, posts(:thinking).id])
+                .preload(:limited_last_comment)
+                .to_a
+
+    assert_no_queries { posts.each(&:limited_last_comment) }
+  end
+
+  def test_preload_has_one_with_limit_alongside_other_associations
+    expected_comment_1 = comments(:more_greetings)
+    expected_comment_2 = comments(:does_it_hurt)
+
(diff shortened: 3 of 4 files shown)


drymar:
I realized it is more complex, closing for now. I will try to figure it

tahsin352:
When the association is has_one and its scope contains a LIMIT clause, then it should be `eager_loaded`. For this case 
```ruby
def eager_loading?
      @should_eager_load ||=
        eager_load_values.any? ||
        includes_values.any? && (joined_includes_values.any? || references_eager_loaded_tables?)
    end
```
Should also contain an additional checking for
```ruby
def eager_loading?
      @should_eager_load ||=
        eager_load_values.any? ||
        includes_values.any? && (joined_includes_values.any? || references_eager_loaded_tables?) ||
        includes_values.any? && has_limit_scope?
    end
```
And the has_limit_scope? method contain the checking for has_one association contains a limit in the scope like
```ruby
def has_limit_scope?
      preload = preload_values
      preload += includes_values unless eager_loading?
      has_one_limit_assocs = []

      preload.each do |assoc_group|
        Array(assoc_group).each do |assoc|
          reflection = model._reflect_on_association(assoc)
          next unless reflection&.macro == :has_one

          assoc_scope = reflection.scope ? model.unscoped.instance_exec(&reflection.scope) : nil
          has_one_limit_assocs << assoc if assoc_scope&.limit_value
        end
      end
      has_one_limit_assocs.any?
    end
```
But doing this, `has_limit_scope?` method returns false, where it should return true.

drymar:
@tahsin352 Thank you for the suggestion! I considered the `eager_load` approach but chose to strip the `LIMIT` in the `preloader` for these reasons:

  1. The LIMIT is redundant for `has_one` - Rails already adds `LIMIT 1` for lazy loading(`AssociationScope#scope`) and the preloader's code already filters to one record per parent in `load_records`. The explicit `limit(1)` is effectively documentation.
  2. Performance - preload's batch query(`WHERE id IN (...)`) is more efficient than `eager_load`'s JOINs, especially with many parent records.
  3. Targeted fix - The scope is already properly evaluated in `Preloader::Association#build_scope`, so we can detect and strip the limit at exactly the right place with a one-line change.

  The approach of switching to `eager_load` would work but would change the loading strategy with potential performance implications rather than fixing the specific issue with preload.

tylerwood-io:
This and #56543 are competing fixes for #56276 — same root cause, slightly different approaches. Comparing both:

**This PR:** strips `limit` from the preload scope unconditionally for singular associations via `scope.limit!(nil) if !reflection.collection? && scope.limit_value`.

**#56543:** strips both `limit` and `offset` via `scope.merge!(reflection_scope.except(:limit, :offset))` for non-collection associations.

#56543's approach is slightly more defensive — it also strips `offset`, which would have the same global-vs-per-parent problem. However, it restructures the merge flow, while this PR is a more minimal one-line addition.

A few observations:

**The fix location is correct.** `build_scope` in `Preloader::Association` is the right place — this is where the scope gets materialized before the batch query. Stripping `limit` here preserves the ordering (so the right "last" record is first in results per owner) while removing the global constraint.

**Test coverage is thorough.** Six test cases covering preload, includes, combined associations, and nil-record scenarios. The fixture-based approach with `limited_last_comment` on Post is clean.

**One consideration:** is there a case where a `has_one` scope with `limit` is intentionally used for something other than "give me the first/last N=1 record"? For `has_one`, `limit(1)` is always semantically redundant since the association already returns one record — but what about `limit(0)` or `limit(N>1)` on a `has_one` scope? Those would be unusual but stripping them unconditionally could change behavior. Probably not a real concern in practice, but an AR maintainer might probe it.

**Missing CHANGELOG:** The checklist says "No CHANGELOG is needed" but Rails convention requires CHANGELOG entries for behavior changes, even small bug fixes. #56543 includes one. This could be a review nit.

tylerwood-io:
Quick correction to my comment above: I stated that #56543 includes a CHANGELOG entry — it does not. Neither PR currently has one. Apologies for the confusion.

PR #56322, opened 2025-12-07
ActiveRecord log to STDOUT on console

coorasse:
### Motivation / Background

Every time I set up a new Rails app, I find myself configuring the following:

```
# config/application.rb
console do
end
```

because I want to see the ActiveRecord logs printed on the console. Working with an ActiveRecord object is the Number 1 reason why I use the rails console.

I wonder if this is the right approach or if we should set this by default in application.rb in new rails apps, but I decided to bring the point nevertheless.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/railtie.rb
+++ b/activerecord/lib/active_record/railtie.rb
@@ -73,6 +73,7 @@ module ActiveRecord
       end
       ActiveRecord.verbose_query_logs = false
       ActiveRecord::Base.attributes_for_inspect = :all
+      ActiveRecord::Base.logger = Logger.new(STDOUT)
     end
 
     runner do


Hanaffi:
Nice work! Should this be added **_only_** to the development environment? Because in the test environment, it might slow down the CI, and I have seen multiple blogs recommending its removal.

coorasse:
This is affecting only the rails console. I think your point is about logging during the tests execution, but in that case, this block would not be executed.

schmijos:
To add an expectation:

Even if _ActiveRecord_ didn't log to _STDOUT_ per default in the Rails console, I'd expect that it did after executing `ActiveRecord.verbose_query_logs = true`. Enabling "verbose" means I want to see things.

And yes: even in `production` environments.

So the config I'd love would look something like this: `ActiveRecord.verbose_query_logs_here_and_right_now!`

p8:
@schmijos You could try adding a new method to:
https://github.com/rails/rails/blob/69763325dd86f839a58e2096785b31a66b763378/railties/lib/rails/commands/console/irb_console.rb

activerecord/lib/active_record/railtie.rb
```diff
@@ -73,6 +73,7 @@ class Railtie < Rails::Railtie # :nodoc:
       end
       ActiveRecord.verbose_query_logs = false
       ActiveRecord::Base.attributes_for_inspect = :all
+      ActiveRecord::Base.logger = Logger.new(STDOUT)
```
skipkayhil, line 76: Would this not cause double logging since `STDERR` is added to the logger's broadcast just above?

skipkayhil: Before (main):
```
$ bin/rails c
Loading development environment (Rails 8.2.0.alpha)
Type 'help' for help.
dev(dev):001> Post.all
  Post Load (9.2ms)  SELECT "posts".* FROM "posts" /* loading for pp */ LIMIT 11 /*application='Dev'*/
=> []
```

After (this PR):
```
$ bin/rails c
Loading development environment (Rails 8.2.0.alpha)
Type 'help' for help.
dev(dev):001> Post.all
D, [2026-06-23T00:27:28.169048 #143127] DEBUG -- :   Post Load (4.4ms)  SELECT "posts".* FROM "posts" /* loading for pp */ LIMIT 11 /*application='Dev'*/
=> []
```

So I don't think this is needed, I'm going to close

PR #56326, opened 2025-12-08
Add slice/except helpers to ActiveModel::Errors

Saidbek:
### Motivation / Background

We often need to scope errors for specific contexts (sectioned forms, partial PATCH responses, logging). Today, doing that requires manual filtering over errors. E.g:
```ruby
err = ActiveModel::Errors.new(user)
user.errors.each do |error|
  err.import(error) if [:billing_address, :billing_city, :billing_zip].include?(error.attribute)
end
logger.warn(err.details)
render json: err.full_messages
```
With a helper, this would become simpler and clearer:
```ruby
err= user.errors.slice(:billing_address, :billing_city, :billing_zip)
logger.warn(err.details)
render json: err.full_messages
```

If these helpers make sense to merge, I can also update the changelog.

### Detail

This PR adds two helpers to `ActiveModel::Errors`:
- `slice(*attributes)`: returns a new `Errors` with only the given attributes’ errors.
- `except(*attributes)`: returns a new `Errors` without the given attributes’ errors.

Keeping this on `Errors` preserves full semantics (types, metadata/options, nested/NestedError wrapping, and downstream helpers like messages, full_messages, details, serialization).

### Checklist

### DIFF
--- a/activemodel/lib/active_model/errors.rb
+++ b/activemodel/lib/active_model/errors.rb
@@ -190,6 +190,20 @@ module ActiveModel
       }
     end
 
+    # Returns a new +Errors+ containing only the errors for the given attributes.
+    #
+    #   person.errors.slice(:name, :email)
+    def slice(*attributes)
+      filter_errors(attributes.flatten.map!(&:to_sym), keep: true)
+    end
+
+    # Returns a new +Errors+ excluding errors for the given attributes.
+    #
+    #   person.errors.except(:email)
+    def except(*attributes)
+      filter_errors(attributes.flatten.map!(&:to_sym), keep: false)
+    end
+
     # Returns +true+ if the error messages include an error for the given key
     # +attribute+, +false+ otherwise.
     #
@@ -499,6 +513,19 @@ module ActiveModel
         end
 
         [attribute.to_sym, type, options]
+      end
+
+      def filter_errors(attributes, keep:)
+        filtered = self.class.new(@base)
+
+        @errors.each do |error|
+          included = attributes.include?(error.attribute)
+          next unless keep ? included : !included
+
+          filtered.import(error)
+        end
+
+        filtered
       end
   end
 
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Add `ActiveModel::Errors#slice` and `#except` to filter errors by attribute.
+
+    ```ruby
+    user.errors.slice(:email, :name)
+    user.errors.except(:base, :admin)
+    ```
+
+    *Said Kaldybaev*
+
 *   Add `has_json` and `has_delegated_json` to provide schema-enforced access to JSON attributes.
 
     ```ruby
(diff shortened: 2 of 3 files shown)


seanpdoyle:
Could you add an entry to the [activemodel/CHANGELOG.md](https://github.com/rails/rails/blob/main/activemodel/CHANGELOG.md) file?

shivabhusal:
Thanks for the PR!

I was trying to understand the exact use case for `Errors#slice`, and I’d like to check if I’m following correctly.

In the current `main` branch, if the goal is only to filter errors for something like JSON serialization, the same result can already be achieved with:

**Your sample**

```ruby
err = ActiveModel::Errors.new(user)
user.errors.each do |error|
  err.import(error) if [:billing_address, :billing_city, :billing_zip].include?(error.attribute)
end
render json: err.to_hash
```

**Possible simplifications**

```ruby
render json: user.errors.to_hash.slice(:billing_address, :billing_city, :billing_zip)
```

or

```ruby
render json: user.errors.messages.slice(:billing_address, :billing_city, :billing_zip)
```

So I’m guessing the motivation for adding slice directly to `ActiveModel::Errors` is that it returns a new `Errors` object (preserving types, metadata, nested errors, etc.) rather than just a sliced `hash`. If that’s the intended use case, that makes sense.

If I’m still misunderstanding the purpose, please correct me. I appreciate the effort you’ve put into this improvement!

activemodel/lib/active_model/errors.rb
```diff
@@ -190,6 +190,20 @@ def where(attribute, type = nil, **options)
...
+    def slice(*attributes)
+      filter_errors(attributes.flatten.map!(&:to_sym), keep: true)
+    end
+
+    # Returns a new +Errors+ excluding errors for the given attributes.
+    #
+    #   person.errors.except(:email)
+    def except(*attributes)
+      filter_errors(attributes.flatten.map!(&:to_sym), keep: false)
+    end
```
seanpdoyle, line 205: If the `flatten` call is important for the sake of making the interface accept both `Array` and variable arguments, could you include test coverage to exercise that coercion?

Similarly, if the interface intends to accept both String and Symbol arguments, could you include coverage for that casting as well?

Saidbek: done

fatkodima:
> `render json: user.errors.slice(:billing_address, :billing_city, :billing_zip).to_hash`

Why not `render json: user.errors.to_hash.slice(:billing_address, :billing_city, :billing_zip)`?

Saidbek:
> > `render json: user.errors.slice(:billing_address, :billing_city, :billing_zip).to_hash`
> 
> Why not `render json: user.errors.to_hash.slice(:billing_address, :billing_city, :billing_zip)`?

Sorry, maybe not the best example, here's a better one:

```ruby
billing_errors = user.errors.slice(:billing_address, :billing_city, :billing_zip)
logger.warn(billing_errors.details)
@billing_messages = billing_errors.full_messages
```

Basically, it enables a structured errors collection to reuse, whereas slicing a plain hash loses that and is a one-way snapshot.

PR #56330, opened 2025-12-09
Enable debug events by default

gmcgibbon:
### Motivation / Background

This Pull Request has been created because debug events don't get emitted by default anymore since we introduced the new log subscribers that use the new event reporter.

### Detail

This Pull Request changes the event reporter to default to debug mode.

Closes https://github.com/rails/rails/issues/56230

Enables debug events when the application hasn't been initialized. This use-case is primarily for test purposes, or when using framework libraries individually outside of Rails. In these cases, we shouldn't assume debug events need to be obfuscated.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Make `ActiveSupport::EventReporter#debug_mode?` true by default to emit debug events
+    outside of Rails application contexts.
+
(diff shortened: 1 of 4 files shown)


adrianna-chang-shopify (approved):
Seems reasonable to me 👍

jeremy (approved):

PR #56333, opened 2025-12-09
Do not clean directories directly under the application root with `Rails::BacktraceCleaner`

alpaca-tc:
### Motivation / Background

- We run ad-hoc scripts under `./script` via `bin/rails runner ./script/<file>`.
  - I think it's common to run files as temporary scripts when you want to avoid putting complex data patches into DB migrations.
- When debugging, we need accurate backtraces and query call locations, but `Rails.backtrace_cleaner` hides frames outside the fixed "application code" paths.
- As a result, useful frames from `./script` (and similar custom dirs) are removed.

*Problem Example*

```
# ./script/verbose_logs.rb
User.where(id: 1).load
```

```
# On main, frames under `./script` are cleaned, so their backtraces don't appear.
$ bundle exec rails runner ./script/verbose_logs.rb
  User Load (1.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = 1
```

```
# Backtrace includes gem internals but omits our script frames we actually want.
$ BACKTRACE=1 bundle exec rails runner ./script/verbose_logs.rb
  User Load (1.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = 1
  ↳ activerecord (7.2.3) lib/active_record/log_subscriber.rb:122:in 'ActiveRecord::LogSubscriber#log_query_source'
```

### Detail

I updated `Rails::BacktraceCleaner` it so that directories directly under the application root - except for vendor - are not cleaned, based on the advice in https://github.com/rails/rails/pull/56333#issuecomment-3635662172.

```
# This PR
$ bundle exec rails runner ./script/verbose_logs.rb
  User Load (1.1ms)  SELECT "users".* FROM "users" WHERE "users"."id" = 1
  ↳ script/verbose_logs.rb:1:in '<top (required)>'
```

### Additional information

Who benefits

- Teams that keep runnable utilities in user-defined dirs, or modular-monolith setups that store app code in `./packs`, can keep those paths visible in backtraces for debugging.
- `ActiveSupport::BacktraceCleaner#remove_filters!` isn't a viable alternative because it strips out the useful default filters.

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Do not clean directories directly under the application root with `Rails::BacktraceCleaner`
+
+    Improved `Rails.backtrace_cleaner` so that most paths located directly under the application's root directory are no longer silenced.
+
+    *alpaca-tc*
+
 *   Add `Rails::CodeStatistics#register_extension` to register file extensions for `rails stats`
 
     ```ruby
--- a/railties/lib/rails/backtrace_cleaner.rb
+++ b/railties/lib/rails/backtrace_cleaner.rb
@@ -5,7 +5,6 @@ require "active_support/core_ext/string/access"
 
 module Rails
   class BacktraceCleaner < ActiveSupport::BacktraceCleaner # :nodoc:
-    APP_DIRS_PATTERN = /\A(?:\.\/)?(?:app|config|lib|test|\(\w+(?:-\w+)*\))/
     RENDER_TEMPLATE_PATTERN = /:in [`'].*_\w+_{2,3}\d+_\d+'/
 
     def initialize
@@ -23,7 +22,10 @@ module Rails
           line
         end
       end
-      add_silencer { |line| !APP_DIRS_PATTERN.match?(line) }
+
+      add_silencer do |line|
+        line.start_with?(File::SEPARATOR, "vendor/", "bin/")
+      end
     end
 
     def clean(backtrace, kind = :silent)


byroot:
So why not, but I think we could start by making the default better.

This list was established 16 years ago in https://github.com/rails/rails/commit/eb39d0f7b999f09c4e13f035634887a8f5592443

I'm thinking we could exclude directories rather than to use an allow list.

Aside from `vendor/` I think anything in the app root is fine?

alpaca-tc:
@byroot Thank you for your feedback 💛 

`/\A(?:\.\/)?(?!vendor)/` does achieve the behavior I'm expecting 👍 
However, since some users install their bundle in paths other than `vendor/bundle` - not ideal, but some people do use `.bundle` - there's a chance this change might receive negative feedback from a subset of users 😕 

If that happens, one option could be to add a silencer like the following:

```
if defined?(Bundler)
  @bundle_path ||= Bundler.bundle_path.to_s
  add_silencer { _1.start_with?(@bundle_path) }
end
```

What do you think?

byroot:
> one option could be to add a silencer like the following:

Bundler's path is already silenced in the parent class.

skipkayhil:
To me it seems like a better configuration point would be to allow customizing `Rails.backtrace_cleaner` instead of allowing the `app_dirs` to be configured. `Rails::BacktraceCleaner` is just a (private) `ActiveSupport::BacktraceCleaner` that adds a few things:

- `BACKTRACE` to disable
- Convert absolute paths -> relative to `Rails.root`
- template render filtering
- app directory filtering

byroot:
> it seems like a better configuration point would be to allow customizing `Rails.backtrace_cleaner` i

Yes, I didn't realize `Rails.backtrace_cleaner =` isn't defined. It definitely should.

But yes, I don't want any more config on the default cleaner, that's why I suggested a behavior that wouldn't require configuring.

alpaca-tc:
@byroot Thanks for the advice!
I've updated the changes based on your suggestions. 😄 

In the original tests, the `bin/` directory was being cleaned, but with this PR it is no longer cleaned. Personally, I believe this behavior matches the expected intention, but it could also be viewed as a compatibility-affecting change.
Could you review this changes 🙏

PR #56335, opened 2025-12-09
Analyze attachments before validation

jeremy:
Attachment metadata (width, height, duration, etc.) is now available for model validations:

```ruby
      has_one_attached :avatar

      validate :validate_avatar_dimensions, if: -> { avatar.attached? }

      def validate_avatar_dimensions
        if avatar.metadata[:width] < 200 || avatar.metadata[:height] < 200
          errors.add(:avatar, "must be at least 200x200")
        end
      end
    end
```

Configure when analysis is performed:

* `analyze: :immediately` (default in 8.2) - Analyze before validation
* `analyze: :later` - Analyze after upload from local IO or via background job
* `analyze: :lazily` - Skip automatic analysis; analyze on-demand

```ruby
    has_one_attached :document, analyze: :later
    has_many_attached :files, analyze: :lazily

    # Or set the global default:
    config.active_storage.analyze = :later
```

Direct uploads bypass the server so the file isn't locally available for analysis. In this case, :immediately falls back to :later, analyzing via background job after upload completes. Metadata isn't available for validation; validate on the client side instead.

/cc @tomrossi7

### DIFF
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -375,6 +375,10 @@ module Rails
             active_record.postgresql_adapter_decode_bytea = true
             active_record.postgresql_adapter_decode_money = true
           end
+
+          if respond_to?(:active_storage)
+            active_storage.analyze = :immediately
+          end
         else
           raise "Unknown version #{target_version.to_s.inspect}"
         end
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,8 +1,40 @@
-*   Eagerly analyze attachments with immediate variants
+*   Analyze attachments before validation
 
-    Attachments with `process: :immediately` variants now eagerly analyze
-    during validation, making metadata like image dimensions available for
-    custom validations.
+    Attachment metadata (width, height, duration, etc.) is now available for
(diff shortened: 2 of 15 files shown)


guilleiguaran (approved):

railties/lib/rails/application/configuration.rb
```diff
@@ -375,6 +375,10 @@ def load_defaults(target_version)
             active_record.postgresql_adapter_decode_bytea = true
             active_record.postgresql_adapter_decode_money = true
           end
+
+          if respond_to?(:active_storage)
+            active_storage.analyze = :immediately
```
chaadow, line 380: @jeremy Sorry for my naive question. But how can someone mitigate uploads of big files, the analyze time directly in the request thread would be huge leading probably to timeouts. Thanks in advance

chaadow: @jeremy since this is done using the local IO, is there a possibility to add another configuration such as if the file is under a certain size, we analyze immediately, otherwise we fallback to "later" ?

jeremy: Selective analysis is a good thread to pull, @chaadow; may lead somewhere good.

PR #56336, opened 2025-12-09
Split dbs_test.rb to whittle away at CI bottlenecks

jeremy:
`railties/test/application/rake/dbs_test.rb` regularly times out:
- Estimated at 82s but actually takes 20+ minutes
- 49 isolated tests = 49 process forks with full app builds
- Single file can't parallelize across CI workers

This PR:
* Bumps up inaccurate duration estimates
* Combines 7 identical ERB YAML tests into one, saving 6 forks where full isolation isn't needed
* Splits into 4 focused tests that CI can parallelize


jeremy:
These are really flaky: sometimes they go through quickly, sometimes block/lag for whatever reason for ages. Lower timeouts may be a better approach.

PR #56338, opened 2025-12-10
nodoc ActiveRecord::QueryMethods#uniq!

p8:
All other bang methods in QueryMethods are nodoc as well.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1539,7 +1539,7 @@ module ActiveRecord
     end
 
     # Deduplicate multiple values.
-    def uniq!(name)
+    def uniq!(name) # :nodoc:
       if values = @values[name]
         values.uniq! if values.is_a?(Array) && !values.empty?
       end


yahonda:
Hi, I'm not sure if we can mark this method as nodoc just because other bang methods in QueryMethods are nodoc.

This method was added in https://github.com/rails/rails/pull/39358 and has been public since Rails 6.1.0. The uniq! method is also documented in the changelog entry: https://github.com/rails/rails/blob/6-1-stable/activerecord/CHANGELOG.md?plain=1#L1101-L1123

p8:
Argh, I should have looked at the commit history.
I thought it was related to the `uniq` method related to `distinct` https://api.rubyonrails.org/v5.0/classes/ActiveRecord/QueryMethods.html#method-i-uniq

p8:
If I'm reading the original PR correctly it seems `uniq!` was introduced in Rails 6.1 to get Rails 7.0 behaviour: deduplicating group fields. As Rails 7.0 deduplicates by default I'm not sure this method should be kept.
Also, that `uniq!` deduplicates group fields isn't obvious from the documentation.
@kamipo Is this method still relevant?

kamipo:
Yeah, this method was introduced for the deprecation cycle, and since all multi values are now deduped, it is no longer relevant.

I originally intended to deprecate this method once the migration was done, so it would be a good idea to deprecate it.

PR #55865, opened 2025-10-08
Make the warning about active storage redirect and proxy mode stronger

pwim:
The existing documentation implied that the "hard to guess" URLs active storage generates provides some sort of access control. This is not the case. Rather, as they rely on ActiveRecord::SignedId, they're tamper proof and don't expose the underlying id.

The security risk of using these URLs isn't that someone will guess them. Rather, it is if the URLs are ever leaked, the files will be exposed.

[This happened to us]( https://www.tokyodev.com/articles/identifying-and-fixing-a-resume-visibility-mistake) via Cloudflare's crawler hints feature.

But there is a myriad of other ways it could happen. E.g. anyone with access to the request log has access to them, they might be exposed via certain analytics tools, even if a user has access to the app revoked they'll still have access to any files, etc.

Because of this, whenever any access control is needed, proxy and redirect mode should not be used.

PR #56340, opened 2025-12-10
[RF-Docs] [ci-skip]  Active Storage Guide

bhumi1102:
This PR is for updating the Active Storage guide. Main changes:

- Restructure the entire guide to be more cohesive
- Add explanations for Active Storage specific concepts 
- Correct inaccurate/confusing text about how `.attach` method works with `has_many_attached` association (the existing text implied that `.attach` replaces existing attachments when adding a new one - not true.)
- Clarify proxy mode vs. redirect mode
- Document `processed` option

Other updates from internal audit and review:
- [X] The "Requirements" section should probably be moved from "What is Active Storage?" to the "Setup" section instead?
- [X] Section 2 "Setup" has a lot of configuration. Maybe we can start with just the Disk service and move S3 and other services to a separate section?
Also, this is worded in a commanding style (Do A, Do B, etc.) without mentioning why A or B should be done.
- [X] For S3 we include the CacheControl config in the main S3 configuration, for GCS it's mentioned as an additional step. This also doesn't mention what this does exactly (making sure uploaded files have CacheControl header set, so an image downloaded from the server won't get loaded again by the browser if it's present the browser's cache and not expired).
- [X] `has_many_attached` doesn't mention the preprocessed option (it's also not mentioned in the API docs, but that is another issue).
- [X] Section 8 "Displaying Images, Videos, and PDFs" mentions 3 different types of "previews": `representable`, `variant` and `preview`. `representable` doesn't have it's own section header. Also, I think `variant` should be mentioned first, as it's probably the more common case.
- [X] The "Lazy vs Immediate Loading" section should probably be mentioned last as it applies to all 3 types.

- [X] in the table in Section 2 - setup, we mention polymorphic associations. This could do with a link to where this appears in a guide, as a beginner may not understand this terminology.
- [X]  When we mention 'bucket names' just before section 2.1, clarify that we mean storage containers by buckets.
- [X] "Continue reading for more information on the built-in service adapters (e.g. Disk and S3) and the configuration they require." -> context needed here about what a service adapter is and why we might need to read about it.
- [X] This guide starts with a lot of configuration that might mean we lose some beginners... maybe they won't need the granular level here, or maybe they will be working within an app that's already configured. With this in mind, maybe we should either move Section 3 earlier in the guide (perhaps with a quick 'get started' section beforehand) or provide a link in the earlier section to this part, with instructions to skip ahead if this is the section that's relevant to you.
- [X] 5.2. Proxy Mode. This opening paragraph is obscure for people that have never met this before. Can we build this out slightly with simpler terms/links and more detailed examples and explanation, so we don't lose people?
- [X] Section 8.1: N+1 queries are mentioned without context - could be confusing for those who have not seen these before.
- [X] 9.1 Usage - 'configure CORS on third-party storage services to allow direct upload requests' (need to add to this point that the instructions for how to do this are in the section below. If you're reading this out of context, it seems like there is expansion/examples missing from this part).
- [X] 12 Purging Unattached Uploads. This is a short section, so does this need to be its own section like this, or could we include all information where the help box first mentions it in section 9.7?


guides/source/active_storage_overview.md
```diff
@@ -8,578 +8,297 @@ This guide covers how to attach files to your Active Record models.
...
+<%= form.file_field :images, multiple: true %>
 ```
 
-#### Variant Generation: Lazily, Later, Immediately
+The above code resubmits the already-attached images back to Rails using hidden fields, so Active Storage keeps the existing attached images when adding a new one.
+
+[ActiveStorage::Blob#signed_id]: https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-signed_id
 
-When you know in advance which variants you'll generate, use the `process` option to control when they're generated:
+### Attaching Files
```
bhumi1102:
```suggestion
### Attaching Files From Disk
```

guides/source/active_storage_overview.md
```diff
@@ -8,578 +8,297 @@ This guide covers how to attach files to your Active Record models.
...
-
-google:
-  service: GCS
-  # ...
-  bucket: your_own_bucket-<%= Rails.env %>
-```
-
-Continue reading for more information on the built-in service adapters (e.g.
-`Disk` and `S3`) and the configuration they require.
+You can find detailed information about [configuring cloud services](#configure-cloud-services) in a later section.
```
bhumi1102:
```suggestion
You can find detailed information about [configuring cloud services](#configuring-cloud-services) in a later section.
```

guides/source/active_storage_overview.md
```diff
@@ -845,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+```
+
+The variant tracker runs automatically. It is enabled by default but can be
+disabled using [`config.active_storage.track_variants`][].
+
+[`config.active_storage.track_variants`]: configuring.html#config-active-storage-track-variants
+[`ActiveStorage::Representations::RedirectController`]: https://api.rubyonrails.org/classes/ActiveStorage/Representations/RedirectController.html
+[`with_all_variant_records`]: https://api.rubyonrails.org/classes/ActiveStorage/Attachment.html#method-c-with_all_variant_records
+
+Configure Cloud Services
```
bhumi1102:
```suggestion
Configuring Cloud Services
```

guides/source/active_storage_overview.md
```diff
@@ -8,578 +8,297 @@ This guide covers how to attach files to your Active Record models.
...
 
 ```ruby
-# Store files locally.
+# config/environments/development.rb
 ```
 
-To use the S3 service in production, you would add the following to
-`config/environments/production.rb`:
+In a production environment you wouldn't use the local disk-based service. So, the `config/storage.yml` file is also where cloud services can be configured. For example, assuming there is a serviced called `amazon` in the `config/storage.yml` file, in order to use that service in production:
```
ayushn21: > In a production environment you wouldn't use the local disk-based service.

Given the push towards self-hosting with Kamal, do we still want to say this?

bhumi1102: Updated

guides/source/active_storage_overview.md
```diff
@@ -8,373 +8,126 @@ This guide covers how to attach files to your Active Record models.
...
+* `active_storage_blobs` - stores data about uploaded files, such as filename and content type.
+* `active_storage_attachments` - a polymorphic join table that [connects your models to blobs](#attaching-files-to-records). This is a [polymorphic association](association_basics.html#polymorphic-associations) so if your model's class name changes, you will need to run a migration to update the underlying `record_type` column in this table to the new name.
+* `active_storage_variant_records` - if [variant tracking](#attaching-files-to-records) is enabled, this table stores records for each variant that has been generated.
 
-WARNING: If you are using UUIDs instead of integers as the primary key on your models, you should set `Rails.application.config.generators { |g| g.orm :active_record, primary_key_type: :uuid }` in a config file.
+WARNING: If you are using UUIDs instead of integers as the primary key on your models, you will need to set `Rails.application.config.generators { |g| g.orm :active_record, primary_key_type: :uuid }` in a config file.
```
p8: Should this be moved just under the install commands, to make it clear this needs to be set before running the commands? Changing it afterwards won't work.

Maybe also clarify this in the text?
```suggestion
WARNING: If you are using UUIDs instead of integers as the primary key on your models, you will need to set `Rails.application.config.generators { |g| g.orm :active_record, primary_key_type: :uuid }` in a config file before installing.
```

guides/source/active_storage_overview.md
```diff
@@ -8,373 +8,126 @@ This guide covers how to attach files to your Active Record models.
...
 ```
 
+Then in the User controller, add `:profile_photo` to the allowed params:
+
 ```ruby
-class SignupController < ApplicationController
+class UserController < ApplicationController
   def create
     user = User.create!(user_params)
     session[:user_id] = user.id
```
p8: If this is now just a user controller, there is no need to set the user_id in the session:
```suggestion
```

guides/source/active_storage_overview.md
```diff
@@ -383,148 +136,89 @@ class SignupController < ApplicationController
...
 ```
 
-Overriding the default service is done the same way as `has_one_attached`, by using the `service` option:
-
-```ruby
-class Message < ApplicationRecord
-  has_many_attached :images, service: :s3
-end
-```
+NOTE: When using `has_many_attached`, calling `images.attach(...)` adds new attachments to the list of existing attachments. It does not replace or overwrite existing attachments. If you want to replace existing images, you must explicitly [purge](#removing-files) the old attachments before attaching new ones.
```
p8: I think this no longer applies, the attachments are always replaced now when calling `attach`:
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#active-storage-assignment-behavior-change

```suggestion
```

guides/source/active_storage_overview.md
```diff
@@ -538,49 +232,73 @@ end
...
 
-### Attaching File/IO Objects
+When working with the `has_many_attached` association, it’s important to distinguish between calling `.attach` directly in Ruby and assigning attachments through form parameters.
 
-Sometimes you need to attach a file that doesn’t arrive via an HTTP request.
-For example, you may want to attach a file you generated on disk or downloaded
-from a user-submitted URL. You may also want to attach a fixture file in a
-model test. To do that, provide a Hash containing at least an open IO object
-and a filename:
+Calling `.attach` always appends new files. It never replaces existing attachments:
```
p8: This also no longer applies:
```suggestion
```

guides/source/active_storage_overview.md
```diff
@@ -538,49 +232,73 @@ end
...
 
 ```ruby
-@message.images.attach(io: File.open("/path/to/file"), filename: "file.pdf")
+# Appends new images, previously attached images remain.
+@product.images.attach(params[:new_images])
+```
+
```
p8, line 324: This example can be removed:
```suggestion
```

guides/source/active_storage_overview.md
```diff
@@ -538,49 +232,73 @@ end
...
+However, form assignment behaves differently. When a form submits `images: params`, Rails treats the submitted list as the entire intended set of attachments for that field. If the form only includes the newly uploaded files, Rails will interpret that as replacing the collection.
```
p8:
```suggestion
When a form submits `images: params`, Rails treats the submitted list as the entire intended set of attachments for that field. If the form only includes the newly uploaded files, Rails will interpret that as replacing the collection.
```

guides/source/active_storage_overview.md
```diff
@@ -601,145 +321,125 @@ end
...
 ### Redirect Mode
 
-To generate a permanent URL for a blob, you can pass the attachment or the blob to
-the [`url_for`][ActionView::RoutingUrlFor#url_for] view helper. This generates a
-URL with the blob's [`signed_id`][ActiveStorage::Blob#signed_id]
-that is routed to the blob's [`RedirectController`][`ActiveStorage::Blobs::RedirectController`]
+To generate a permanent URL for a blob, you can pass the attachment or the blob
+to the [`url_for`][ActionView::RoutingUrlFor#url_for] view helper. This
+generates a URL with the blob's [`signed_id`][ActiveStorage::Blob#signed_id]
+which points to the the blob's
```
p8:
```suggestion
which points to the blob's
```

guides/source/active_storage_overview.md
```diff
@@ -601,145 +321,125 @@ end
...
+(e.g. an expiring S3 URL). This indirection decouples your application’s public
+URLs from the underlying storage service and enables features such as mirroring
+attachments across multiple services for high-availability. The redirect
+response is cached by the browser for 5 minutes.
 
-To create a download link, use the `rails_blob_{path|url}` helper. Using this
-helper allows you to set the disposition.
+To create a download link, use the `rails_blob_{path|url}` helpers. These
+helpers generate the same permanent Rails URL but allow you to specify the file
+disposition.
```
p8: `disposition` isn't explained anywhere. Maybe link to Mozilla's Content-Disposition Header page here?
```suggestion
[Content-Disposition Header].

[Content-Disposition Header]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition
```

guides/source/active_storage_overview.md
```diff
@@ -789,23 +489,21 @@ end
...
-require a login. If your files require a higher level of protection, you can
-implement your own authenticated controllers, based on the
+By default, all Active Storage controllers are publicly accessible. The URLs they generate contain a blob’s [signed_id][ActiveStorage::Blob#signed_id], which is hard to guess but permanent. Anyone who knows the URL can access the file, even if the rest of your application requires authentication. before_actions in your own controllers (such as requiring a logged-in user) do not apply to Active Storage’s built-in controllers.
+
+If your files require stricter access control, such as “a user may only view their own files”, you can replace the built-in controllers with your own authenticated controllers. These controllers should wrap the behavior of the following built-in controllers but apply your own authorization logic before serving the file :
+
 [`ActiveStorage::Blobs::RedirectController`][],
 [`ActiveStorage::Blobs::ProxyController`][],
 [`ActiveStorage::Representations::RedirectController`][] and
 [`ActiveStorage::Representations::ProxyController`][]
```
p8: Would this be more readable in a list?
```suggestion
* [`ActiveStorage::Blobs::RedirectController`][]
* [`ActiveStorage::Blobs::ProxyController`][]
* [`ActiveStorage::Representations::RedirectController`][]
* [`ActiveStorage::Representations::ProxyController`][]
```

guides/source/active_storage_overview.md
```diff
@@ -789,23 +489,21 @@ end
...
 
 ### Authenticated Controllers
 
-All Active Storage controllers are publicly accessible by default. The generated
-URLs use a plain [`signed_id`][ActiveStorage::Blob#signed_id], making them hard to
-guess but permanent. Anyone that knows the blob URL will be able to access it,
-even if a `before_action` in your `ApplicationController` would otherwise
-require a login. If your files require a higher level of protection, you can
-implement your own authenticated controllers, based on the
+By default, all Active Storage controllers are publicly accessible. The URLs they generate contain a blob’s [signed_id][ActiveStorage::Blob#signed_id], which is hard to guess but permanent. Anyone who knows the URL can access the file, even if the rest of your application requires authentication. before_actions in your own controllers (such as requiring a logged-in user) do not apply to Active Storage’s built-in controllers.
```
p8: Nitpick, but the plain before_actions looks a bit akward:
```suggestion
By default, all Active Storage controllers are publicly accessible. The URLs they generate contain a blob’s [signed_id][ActiveStorage::Blob#signed_id], which is hard to guess but permanent. Anyone who knows the URL can access the file, even if the rest of your application requires authentication. `before_action`s in your own controllers (such as requiring a logged-in user) do not apply to Active Storage’s built-in controllers.
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+  has_one_attached :document, analyze: :later
+
+  # Skip automatic analysis - analyze on-demand when metadata is accessed
+  has_many_attached :files, analyze: :lazily
+end
+```
+
+NOTE: Attachments with `process: :immediately` variants automatically analyze immediately to ensure metadata is available before processing.
+
+You can set the an application level default for the `analyze` option your in your Rails application configuration as well:
```
p8:
```suggestion
You can set the application level default for the `analyze` option in your Rails application configuration as well:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+```irb
+> user.profile_photo.analyzed?
+=> true
+```
+
+Once a file has been analyzed, the metadata is stored in the `active_storage_blobs` table and can be viewed with `user.profile_photo.blob.metadata`.
+
+Image analysis provides `width` and `height` attributes. Video analysis provides these, as well as `duration`, `angle`, `display_aspect_ratio`, and `video` and `audio` booleans to indicate the presence of those channels. Audio analysis provides `duration` and `bit_rate` attributes.
+
+### Validating Attachment Metadata
```
p8: This section mentions some of the ways attachments can be analyzed: `immediately` and `later`, but the next section explains these.
Maybe move `Controlling When Analysis is Performed` before this section?

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+  root: <%= Rails.root.join("storage") %>
 ```
 
-NOTE: [Direct uploads](#direct-uploads) bypass the server so the file isn't locally available for analysis. In this case, `:immediately` falls back to `:later`, analyzing via background job after upload completes. Metadata isn't available for validation; validate on the client side using JavaScript instead.
-
-### Controlling When Analysis is Performed
+### S3 Service (Amazon S3 and S3-compatible APIs)
 
-Pass the `analyze:` option to control when analysis is performed:
+Active Storage’s built-in S3 service adapter relies on the official AWS SDK to communicate with Amazon S3 (or any S3-compatible service). Rails does not bundle the AWS SDK by default, so you must add the `aws-sdk-s3 gem` to your application’s Gemfile:
```
p8:
```suggestion
Active Storage’s built-in S3 service adapter relies on the official AWS SDK to communicate with Amazon S3 (or any S3-compatible service). Rails does not bundle the AWS SDK by default, so you must add the `aws-sdk-s3` gem to your application’s Gemfile:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+  # ...and other options
 ```
 
-Internally, `representation` calls `variant` for images, and `preview` for
-previewable files. You can also call these methods directly.
+NOTE: The core features of Active Storage require the following permissions: `s3:ListBucket`, `s3:PutObject`, `s3:GetObject`, and `s3:DeleteObject`. [Public access](#public-access) additionally requires `s3:PutObjectAcl`. If you have additional upload options configured such as setting ACLs then additional permissions may be required.
 
-[`representable?`]: https://api.rubyonrails.org/classes/ActiveStorage/Blob/Representable.html#method-i-representable-3F
-[`representation`]: https://api.rubyonrails.org/classes/ActiveStorage/Blob/Representable.html#method-i-representation
+There are many other options available. You can see them in [AWS S3 Client](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method) documentation.
```
p8:
```suggestion
There are many other options available. You can see them in the [AWS S3 Client](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method) documentation.
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
 
 ```ruby
-image_tag file.representation(resize_to_limit: [100, 100])
+gem "google-cloud-storage", "~> 1.11", require: false
 ```
 
-Will generate an `<img>` tag with the `src` pointing to the
-[`ActiveStorage::Representations::RedirectController`][]. The browser will
-make a request to that controller, which will perform the following:
+Then you can declare a Google Cloud Storage service in `config/storage.yml`:
```
p8, line 1271: It may be worth mention why the `require: false` option is used here as well, similar to the S3 section above:
```suggestion
The `require: false` option avoids loading the gem automatically. Active Storage will load it only when the GCS service is used, keeping application boot time and memory usage lower.

Then you can declare a Google Cloud Storage service in `config/storage.yml`:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+    token_uri: "https://accounts.google.com/o/oauth2/token"
+    auth_provider_x509_cert_url: "https://www.googleapis.com/oauth2/v1/certs"
+    client_x509_cert_url: ""
+  project: ""
+  bucket: your_own_bucket-<%= Rails.env %>
+  cache_control: "public, max-age=3600"
+```
 
-If you want to generate URLs for images immediately, you can call `.processed.url`:
+You can optionally use [IAM](https://cloud.google.com/storage/docs/access-control/signed-urls#signing-iam) instead of the `credentials` when signing URLs. This is useful if you are authenticating your GKE applications with Workload Identity, see [this Google Cloud blog post](https://cloud.google.com/blog/products/containers-kubernetes/introducing-workload-identity-better-authentication-for-your-gke-applications) for more information.
```
p8: GKE isn't explained anywhere:
```suggestion
You can optionally use [IAM](https://cloud.google.com/storage/docs/access-control/signed-urls#signing-iam) instead of the `credentials` when signing URLs. This is useful if you are authenticating your GKE (Google Kubernetes Engine) applications with Workload Identity, see [this Google Cloud blog post](https://cloud.google.com/blog/products/containers-kubernetes/introducing-workload-identity-better-authentication-for-your-gke-applications) for more information.
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
-WARNING: It should be considered unsafe to provide arbitrary user supplied
-transformations or parameters to variant processors. This can potentially
-enable command injection vulnerabilities in your app. It is also recommended
-to implement a strict [ImageMagick security policy](https://imagemagick.org/script/security-policy.php)
-when MiniMagick is the variant processor of choice.
+NOTE: Mirroring is not atomic. It’s possible for an upload to succeed on the primary service but fail on one or more mirrors. Before switching fully to the new service, ensure that all files have been successfully copied.
 
-If a variant is requested, Active Storage will automatically apply
-transformations depending on the image's format:
+In order to define a `Mirror` service, first, define each service you want to mirror as usual. Then, reference them by name in the `Mirror` service configuration:
```
p8: This comma feels a bit redundant:
```suggestion
In order to define a `Mirror` service, first define each service you want to mirror as usual. Then, reference them by name in the `Mirror` service configuration:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+Mirror services are compatible with [direct uploads](#direct-uploads). New files
+are directly uploaded to the primary service. When a directly-uploaded file is
+attached to a record, a background job is enqueued to copy it to the secondary
+services.
 
-Some non-image files can be previewed: that is, they can be presented as images.
-For example, a video file can be previewed by extracting its first frame. Out of
-the box, Active Storage supports previewing videos and PDF documents. To create
-a link to a lazily-generated preview, use the attachment's [`preview`][] method:
+### Public access
```
p8:
```suggestion
### Public Access
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+To use direct uploads, you'll need to include the library in your application’s JavaScript bundle and enable the `direct_upload: true` option on your file input fields. This allows Rails and the storage service to coordinate securely using signed IDs, without requiring extra backend configuration.
 
-    Requiring it directly in the application HTML with autostart, instead of
-    bundling it through the asset pipeline:
+There are several ways to include the Active Storage JavaScript library in your application:
 
-    ```erb
-    <%= javascript_include_tag "activestorage" %>
-    ```
+1. Directly via `javascript_include_tag` - include the library in your HTML without bundling through the asset pipeline. Autostart is enabled automatically:
```
p8: The numbering here is a bit off (all items start with 1).

Maybe just use a sub-heading?
```suggestion
#### Directly via `javascript_include_tag`

Include the library in your HTML without bundling through the asset pipeline. Autostart is enabled automatically:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
 
-    Requiring via importmap-rails as an ESM in the application HTML, instead of
-    bundling it through the asset pipeline and using autostart:
+```erb
+<%= javascript_include_tag "activestorage" %>
+```
+
+1. Using Importmap (ESM) - pin the library in `config/importmap.rb`:
```
p8:
```suggestion
#### Using Importmap (ESM)

Pin the library in `config/importmap.rb`:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+</script>
+```
 
-    ```html
-    <script type="module-shim">
-      import * as ActiveStorage from "@rails/activestorage"
-      ActiveStorage.start()
-    </script>
-    ```
+1. Using the Asset Pipeline - require the library in a JavaScript manifest:
```
p8:
```suggestion
#### Using the Asset Pipeline

Require the library in a JavaScript manifest:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
 
-    Using the asset pipeline:
+```javascript
+//= require activestorage
+```
 
-    ```js
-    //= require activestorage
-    ```
+1. Using the npm package - install via npm/yarn and import it in your JavaScript bundle:
```
p8:
```suggestion
#### Using the npm package

Install via npm/yarn and import it in your JavaScript bundle:
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
+```
 
-    ```js
-    import * as ActiveStorage from "@rails/activestorage"
-    ActiveStorage.start()
-    ```
+All of these approaches provide the same functionality; choose the one that matches your application’s JavaScript setup.
 
-2. Add `direct_upload: true` option to your [`file_field`
+Next step is to set the `direct_upload: true` option in your [`file_field`
```
p8, line 1503: Wondering if it would be clearer to add a heading here as well?

```suggestion
#### Enabling Direct Uploads on the Input

The next step is to set the `direct_upload: true` option in your [`file_field`
```

guides/source/active_storage_overview.md
```diff
@@ -846,285 +543,644 @@ to prevent files being accessed with the publicly accessible URLs.
...
 --------------
 
+By default, files uploaded through Active Storage are sent to your Rails server first, then forwarded to the configured storage service. Direct uploads bypass the Rails server entirely, sending files straight from the browser to the storage service. Direct uploads provide improved performance as large files do not have to pass through your Rails server.
+
+Direct uploads integrate seamlessly with Active Storage’s attachments and variants, allowing you to use the same models, validations, and background processing workflows as standard uploads.
+
 Active Storage, with its included JavaScript library, supports uploading
 directly from the client to the cloud.
 
 ### Usage
```
p8: Maybe use `Setup` here, just like in the first section?

```suggestion
### Setup
```

guides/source/active_storage_overview.md
```diff
@@ -1289,11 +1343,12 @@ input[type=file][data-direct-upload-url][disabled] {
 
 ### Custom drag and drop solutions
```
p8:
```suggestion
### Custom Drag and Drop Solutions
```

guides/source/active_storage_overview.md
```diff
@@ -1338,10 +1393,10 @@ const uploadFile = (file) => {
 
 ### Track the progress of the file upload
```
p8:
```suggestion
### Track the Progress of the File Upload
```

guides/source/active_storage_overview.md
p8, line 1576: Maybe add a description what these events can be used for?

```suggestion
### Direct Upload JavaScript Events

The JavaScript library supports events that can be used for the upload form:
```

guides/source/active_storage_overview.md
```diff
@@ -1202,8 +1258,7 @@ You can use these events to show the progress of an upload.
 To show the uploaded files in a form:
```
p8:
```suggestion
To show the progress of the uploaded files in a form add the following javascript:
```

guides/source/active_storage_overview.md
```diff
@@ -1246,8 +1301,7 @@ addEventListener("direct-upload:end", event => {
 Add styles:
```
p8: Maybe this could be expanded slightly?

```suggestion
Add CSS to style the progress of the uploaded files:
```

guides/source/active_storage_overview.md
```diff
@@ -1375,10 +1430,11 @@ class Uploader {
 
 ### Integrating with Libraries or Frameworks
 
-Once you receive a file from the library you have selected, you need to create
-a `DirectUpload` instance and use its "create" method to initiate the upload process,
-adding any required additional headers as necessary. The "create" method also requires
-a callback function to be provided that will be triggered once the upload has finished.
+Once you receive a file from the library you have selected, you need to create a
+`DirectUpload` instance and use its "create" method to initiate the upload
```
p8:
```suggestion
`DirectUpload` instance and use its `create` method to initiate the upload
```

guides/source/active_storage_overview.md
```diff
@@ -1431,21 +1487,20 @@ end
 NOTE: Using [Direct Uploads](#direct-uploads) can sometimes result in a file that uploads, but never attaches to a record. Consider [purging unattached uploads](#purging-unattached-uploads).
 
 Testing
```
p8, line 1832: For some other guides we link to the testing guide instead.
For example:
https://guides.rubyonrails.org/active_job_basics.html#job-testing
https://guides.rubyonrails.org/action_cable_overview.html#testing

Maybe move this as well?

guides/source/active_storage_overview.md
```diff
@@ -1608,20 +1649,20 @@ end
...
 end
 ```
 
-Without `config/storage/test.yml`, the `s3` service configured in `config/storage.yml` is used - even when running tests.
+Without configuring a test `s3` service in `config/storage/test.yml`, the `s3` service configured in `config/storage.yml` is used - even when running tests.
 
 The default configuration would be used and files would be uploaded to the service provider configured in `config/storage.yml`.
 
-In this case, you can add `config/storage/test.yml` and use Disk service for `s3` service to prevent sending requests.
+In this case, you can add `config/storage/test.yml` and use Disk service for a service named `s3` to prevent sending requests.
```
p8:
```suggestion
In this case, you can add `config/storage/test.yml` and use Disk service for a service named `s3` to prevent sending requests to S3.
```

guides/source/active_storage_overview.md
bhumi1102: I like the idea of making it clear in the warning text. I added "You need to set this configuring before running the `active_storage:install` command"

I tried moving the warning up but didn't like how it interrupts the reading flow. So going with making it super clear in the text.

guides/source/active_storage_overview.md
bhumi1102: Yeah good call, done!

guides/source/active_storage_overview.md
bhumi1102: I like this subheading to break up the section, added!

guides/source/active_storage_overview.md
bhumi1102: Thanks for spotting that! I like the numbering 1, 2, 3, 4 so let me know try fixing that. If the formatting gets messed up again, I can go with the short subsections.

guides/source/active_storage_overview.md
bhumi1102: I moved the testing section to the Testing guide and added a link in the Active Storage guide.

guides/source/active_storage_overview.md
bhumi1102: Hi @p8 - I think this has been confusing in the past. But the behavior of `attach` is and always has been "append" not "replace". Reading that upgrade guide, it does say "#attach can be used to add new attachments without removing the existing ones:" I think the distinction is `attach` vs. something like`product.update!`, the latter _will_ "replace" and not "append" (though it used to "append" same as `attach` and that behavior got phased out and then deprecated in favor of `.update` doing "replace" behavior.) 

So the above note is accurate still I think. That's my reading and understanding, let me know if that distinction makes sense to you as well or if I've missed something! :)

Update: tried this out in an fresh Rails app and yes `.attach` does still add new attachments to the list and does not replace them.

guides/source/active_storage_overview.md
```diff
@@ -8,523 +8,216 @@ This guide covers how to attach files to your Active Record models.
...
+image representations of non-image uploads like PDFs and videos, and extract
+metadata.
 
-### Requirements
+For cloud storage services, Active Storage supports mirroring files to secondary
+services to serve as a backup or to allow migration between services. Active Storage also supports Direct Uploads, allowing files to be uploaded straight from the client's browser to the configured cloud storage service. This avoids routing large files through your Rails servers.
 
-Various features of Active Storage depend on third-party software which Rails
-will not install, and must be installed separately:
+For local development and testing, Active Storage has a `Disk` service which
```
ayushn21: Should we still say that the `Disk` service is for local development and testing?

guides/source/active_storage_overview.md
```diff
@@ -8,523 +8,216 @@ This guide covers how to attach files to your Active Record models.
...
 What is Active Storage?
 -----------------------
 
-Active Storage facilitates uploading files to a cloud storage service like
-Amazon S3, or Google Cloud Storage and attaching those
-files to Active Record objects. It comes with a local disk-based service for
-development and testing and supports mirroring files to subordinate services for
-backups and migrations.
+Active Storage facilitates attaching files to Active Record objects and
+uploading those files to cloud storage services.
```
ayushn21: Is it worth rewording this?

```suggestion
uploading those files to your server or to a cloud storage service.
```

guides/source/active_storage_overview.md
bhumi1102: Same reason as above

guides/source/active_storage_overview.md
bhumi1102: Good to have still I think, for the same reason as above.

guides/source/active_storage_overview.md
```diff
@@ -953,178 +770,422 @@ the file instead.
...
+  import * as ActiveStorage from "@rails/activestorage"
+  ActiveStorage.start()
+</script>
+```
 
-    ```ruby
-    # config/importmap.rb
-    pin "@rails/activestorage", to: "activestorage.esm.js"
-    ```
+3. Using the Asset Pipeline - require the library in a JavaScript manifest:
```
ayushn21: I don't think this technique applies to Propshaft. Since that's the default, perhaps we should remove it and the import map example above should be sufficient?

ayushn21 (commented):
This is one of the most insightful posts on Active Storage I've read: https://discuss.rubyonrails.org/t/active-storage-in-production-lessons-learned-and-in-depth-look-at-how-it-works/83289.

I wonder if it's worth linking to somewhere in the guide?

guides/source/active_storage_overview.md
p8: Yes!

guides/source/active_storage_overview.md
p8: Ah sorry @bhumi1102, you are correct. I thought I tested locally on a mature app but can't reproduce it anymore. I also tested this behaviour in a gist: https://gist.github.com/p8/eb9010de1708e80718fc97c5fb031f84
Thanks for checking 😄

guides/source/active_storage_overview.md
p8: I think that is still the most common use case for `Disk` service.

p8:
> This is one of the most insightful posts on Active Storage I've read: https://discuss.rubyonrails.org/t/active-storage-in-production-lessons-learned-and-in-depth-look-at-how-it-works/83289.
> 
> I wonder if it's worth linking to somewhere in the guide?

With the rewrites we've been removing links to blogs and articles mostly. Some of these links no longer work or the information hasn't been updated for 10 years and is outdated. I think the only links we still support are Wikipedia, Mozilla and links to libraries.

If this guide is missing information we should add it to the guide.
I'm also not sure if all information in that post is supported by Rails Core.

guides/source/active_storage_overview.md
bhumi1102: No worries at all @p8! This has been a bit confusing so glad we are clarifying it in the guide now! 

Oh and I like that gist, nice minimal way to test out a feature. I keep spinning up demo Rails apps, haha.

guides/source/active_storage_overview.md
ayushn21: As of today that's probably true, but going forward, is this still the Core Team's position? I feel that the way we use language is important and if the official position is that the Disk service is fine for production, we shouldn't use language suggesting it's best used for development and testing.

Even before Rails 8, SQLite could be used in production with Rails, even though there was a warning against it. With the big SQLite push with Rails 8, that warning was removed, so I'm thinking along those lines here. 

I only raise this because looking at the source of apps like Campfire, Fizzy etc, they are using the Disk service for Active Storage, and I don't want to create an impression that it isn't for production use.

ayushn21:
> If this guide is missing information we should add it to the guide. I'm also not sure if all information in that post is supported by Rails Core.

Yeah that's fair enough, thanks :)

guides/source/active_storage_overview.md
bhumi1102: Updated wording to remove environment mention

guides/source/active_storage_overview.md
```diff
@@ -601,162 +385,181 @@ end
...
 
 WARNING: All Active Storage controllers are publicly accessible by default. The
-generated URLs are hard to guess, but permanent by design. If your files
-require a higher level of protection consider implementing
-[Authenticated Controllers](#authenticated-controllers).
+Rails generated URLs are hard to guess, but permanent by design. If your files
+require a higher level of protection consider implementing [Authenticated
+Controllers](#authenticated-controllers).
 
 ### Redirect Mode
```
p8, line 482: @bhumi1102 I recently had a look at these controllers for #55865 and found out you can also create expiring URLs:
https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blobs/RedirectController.html
https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blobs/ProxyController.html

I think we should mention the `:expires_at` option and `config.active_storage.urls_expire_in` in this guide as well.

bhumi1102: Hi @p8 - agree, thanks for the heads up on this PR! I read the details and the blog post linked. I've added a little subsection to explain the `expires_in` option. Hopefully this conveys the difference between access control with authentication controller and other options. Let me know if you have notes on that new section and I'll address it to get this one guide ready to go!

guides/source/active_storage_overview.md
```diff
@@ -832,305 +642,835 @@ end
...
+--------------
 
-Active Storage analyzes files to extract metadata like image dimensions, video duration, and audio bit rate. Analyzed files store this information in their metadata hash, including `analyzed: true`. You can check whether a blob has been analyzed by calling [`analyzed?`][] on it.
+### Purging Attachments From a Model
 
-Image analysis provides `width` and `height` attributes. Video analysis provides these, as well as `duration`, `angle`, `display_aspect_ratio`, and `video` and `audio` booleans to indicate the presence of those channels. Audio analysis provides `duration` and `bit_rate` attributes.
+To remove an attachment from a model, call [`purge`][Attached::One#purge] on the
+attachment. If your application is set up to use Active Job, removal can be done
+in the background instead by calling [`purge_later`][Attached::One#purge_later].
+Purging deletes the blob and the file from the storage service.
```
p8: Nitpick. Some readers might think the blob is deleted from the storage service as well.

Also attachments are deleted/destroyed as well and blobs are only destroyed when there are no more attachments attached (I don't think we mention anywhere in the guide that only one Blob is created for a file with the same checksum).

Maybe something like:
```suggestion
Purging destroys the attachment record. If the blob has no more attachments it gets destroyed as well and the file is deleted from the storage service.
```

guides/source/active_storage_overview.md
```diff
@@ -8,523 +9,265 @@ This guide covers how to attach files to your Active Record models.
...
+routing large files through your Rails servers.
 
-Various features of Active Storage depend on third-party software which Rails
-will not install, and must be installed separately:
+Active Storage also supports a `Disk` service which uses the local filesystem by default.
 
-* [libvips](https://github.com/libvips/libvips) v8.6+ or [ImageMagick](https://imagemagick.org/index.php) for image analysis and transformations
-* [ffmpeg](http://ffmpeg.org/) v3.4+ for video previews and ffprobe for video/audio analysis
-* [poppler](https://poppler.freedesktop.org/) or [muPDF](https://mupdf.com/) for PDF previews
+### Setup
```
p8: In other guides "Setup" or "Installation" is a top section.
- [Action Mailbox > Setup](https://edgeguides.rubyonrails.org/action_mailbox_basics.html#setup)
- [Action Text > Installation](https://edgeguides.rubyonrails.org/action_text_overview.html#installation)
- [Active Storage > Setup](https://edgeguides.rubyonrails.org/active_storage_overview.html#setup)
- [I18n > Setup the Rails Application for Internationalization](https://edgeguides.rubyonrails.org/i18n.html#setup-the-rails-application-for-internationalization)

For consistency we should probably do it here as well?
```suggestion
Setup
-----
```

guides/source/active_storage_overview.md
```diff
@@ -603,96 +387,94 @@ end
...
 
-[Attached::One#purge]: https://api.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-purge
-[Attached::One#purge_later]: https://api.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-purge_later
+This query executes against the `active_storage_blobs` table rather than the
+attachment records themselves, since the join created by `joins(:images_blobs)`
+operates on the blob side of the association. You can combine such blob-based
+filters with additional scope conditions in the same way you would with any
+standard Active Record query.
 
 Serving Files
```
p8, line 495: This was a pleasure to read.

guides/source/active_storage_overview.md
```diff
@@ -832,305 +642,835 @@ end
...
-[Blob#open]: https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-open
+[Blob#download]:
+https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-download
+[Blob#open]:
+https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-open
 
-Analyzing Files
----------------
+Removing Files
+--------------
```
p8, line 757: It would be nice if we added some introduction paragraph here explaining when you want to remove files.

There also is the `detach` method, that we don't describe.
It removes the attachment and leaves the Blob in place.
https://edgeapi.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-detach

I'm not sure what the use-case is (making sure a Blob isn't accidentally deleted?) but we probably need to add it?

guides/source/active_storage_overview.md
```diff
@@ -832,305 +642,835 @@ end
...
 
-Active Storage analyzes files to extract metadata like image dimensions, video duration, and audio bit rate. Analyzed files store this information in their metadata hash, including `analyzed: true`. You can check whether a blob has been analyzed by calling [`analyzed?`][] on it.
+### Purging Attachments From a Model
```
p8: I noticed there is a `purge` method defined for `has_many_attached` as well. It purges all attached files.
Should we describe that method as well?

Or maybe more importantly explain how to remove a single file from a `has_many_attached`?

Also if we'll add the `detach` method, maybe rename this section to "Removing Attachments from a Model"?

guides/source/active_storage_overview.md
```diff
@@ -832,305 +642,835 @@ end
...
+WARNING: The query generated by `ActiveStorage::Blob.unattached` can be slow and
+potentially disruptive on applications with larger databases.
+
+Analyzing Files For Metadata
+----------------------------
+
+Active Storage analyzes files to extract metadata like image dimensions, video
+duration, and audio bit rate. Analyzed files will store additional information
+in the metadata hash, including `analyzed: true`. You can check whether a blob
+has been analyzed by calling the [`analyzed?`][] method on it.
```
p8: Before describing `analyzed?` I think we should describe the `metadata` method (I've just added this method to the API docs).
Something like:
````suggestion
Active Storage analyzes files to extract metadata like image dimensions, video
duration, and audio bit rate.  Once a file has been analyzed, the metadata is stored in the
`active_storage_blobs` table and can be viewed with
the [`metadata`][] method :

```irb
> user.profile_photo.metadata
=> {"identified" => true, "width" => 112, "height" => 243, "created_at" => "2026-04-05T00:11:48+02:00", "analyzed" => true}
```

Analyzed files will store additional information
in the metadata hash, including `analyzed: true`. You can check whether a blob
has been analyzed by calling the [`analyzed?`][] method on it.
````

File: guides/source/active_storage_overview.md [hunk 46]
```diff
@@ -832,305 +642,835 @@ end
...
-bundle or reference it directly.
+To use direct uploads, you'll need to include the library in your application’s
+JavaScript bundle and enable the `direct_upload: true` option on your file input
+fields. This allows Rails and the storage service to coordinate securely using
+signed IDs, without requiring extra backend configuration.
 
-    Requiring it directly in the application HTML with autostart, instead of
-    bundling it through the asset pipeline:
+There are several ways to include the Active Storage JavaScript library in your
+application:
```
p8, line 1459: This list isn't properly displaying:

Image

p8 (commented):
This is looking great @bhumi1102 !
I've added some final comments.

guides/source/active_storage_overview.md
bhumi1102: sure!

guides/source/active_storage_overview.md
bhumi1102: Oh good, thank you :)

guides/source/active_storage_overview.md
bhumi1102: good point!

guides/source/active_storage_overview.md
bhumi1102: Thank you for that addition and the diff for it!

guides/source/active_storage_overview.md
bhumi1102: man, I thought I fixed this. alright, I might need to just go with subheading like you suggested a while back haha

guides/source/active_storage_overview.md
bhumi1102: Good points. Added an intro sentence. Also added `detach` method (in the 2nd section along with `unattached`). It seems the use case may be when you want to leave the file on the storage service but not have it associated with a record? Either way, I added a description with an example.

guides/source/active_storage_overview.md
bhumi1102: Agree. Renamed. Added an example to show how to remove a single file from `has_many_attached`

guides/source/active_storage_overview.md
```diff
@@ -832,305 +643,856 @@ end
...
+
+
+```irb
+> user.profile_photo.analyzed?
+=> true
+```
+
+Once a file has been analyzed, the metadata is stored in the
+`active_storage_blobs` table and can be viewed with
+`user.profile_photo.blob.metadata`.
```
p8: I think this suggestion didn't apply correctly.

````suggestion
```irb
> user.profile_photo.metadata
=> {"identified" => true, "width" => 112, "height" => 243, "created_at" => "2026-04-05T00:11:48+02:00", "analyzed" => true}
```
````

I don't have a strong opinion about keeping the `analyzed?` method example.
So it could also be:

````suggestion
```irb
> user.profile_photo.metadata
=> {"identified" => true, "width" => 112, "height" => 243, "created_at" => "2026-04-05T00:11:48+02:00", "analyzed" => true}
```

You can check whether a blob has been analyzed by calling the [`analyzed?`][]
method on it:

```irb
> user.profile_photo.analyzed?
=> true
```
````

File: guides/source/active_storage_overview.md [hunk 48]
```diff
@@ -832,305 +643,856 @@ end
...
-NOTE: [Direct uploads](#direct-uploads) bypass the server so the file isn't locally available for analysis. In this case, `:immediately` falls back to `:later`, analyzing via background job after upload completes. Metadata isn't available for validation; validate on the client side using JavaScript instead.
+NOTE: Since [Direct uploads](#direct-uploads) bypass the server, files aren't
+locally available for analysis. In this case, `:immediately` falls back to
+`:later`, analyzing via background job after upload completes. So model
+validations using metadata isn't possible. You can validate on the client side
+using JavaScript instead.
 
-### Controlling When Analysis is Performed
+[`analyzed?`]:
+https://api.rubyonrails.org/classes/ActiveStorage/Blob/Analyzable.html#method-i-analyzed-3F
```
p8, line 902:
```suggestion
[`metadata`][]:
https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-metadata
[`analyzed?`]:
https://api.rubyonrails.org/classes/ActiveStorage/Blob/Analyzable.html#method-i-analyzed-3F
```

p8:
@bhumi1102 Some final suggestions. Could you also look at the merge conflict?

bhumi1102:
> @bhumi1102 Some final suggestions. Could you also look at the merge conflict?

Hi @p8, ready for you again, thanks! :)

1. resolved the merge conflict, there was a PR that added the `config.active_storage.routes_prefix`. It didn't flow at the original location the PR added, so I ended up creating a little subsection for it under "Setup and Configuration". I'm not sure if this is needed in this guide or it being in the configuration guide is enough but I didn't want to just ignore the incoming PR so added. Let me know if you have an opinion either way!

2. manually applied that last change, github UI was not playing nice with that one for some reason.

guides/source/active_storage_overview.md
```diff
@@ -8,523 +9,289 @@ This guide covers how to attach files to your Active Record models.
...
+`https://example.com/rails/active_storage/blobs/redirect/eyJf.../photo.jpg`. You can see all the routes by running:
 
-```yaml
-google:
-  service: GCS
-  ...
-  iam: true
-  gsa_email: "foobar@baz.iam.gserviceaccount.com"
+```bash
+bin/rails routes | grep active_storage
```
p8: This is missing the prompt. Also we should use the build-in grep option as that keeps the table headers in the output.
```suggestion
$ bin/rails routes --grep active_storage
```

p8:
>     1. resolved the merge conflict, there was a PR that added the `config.active_storage.routes_prefix`. It didn't flow at the original location the PR added, so I ended up creating a little subsection for it under "Setup and Configuration". I'm not sure if this is needed in this guide or it being in the configuration guide is enough but I didn't want to just ignore the incoming PR so added. Let me know if you have an opinion either way!

I think that's a great location for the routes section!

bhumi1102:
@p8 thanks! applied that last change and this one is good to go again!

PR #51940, opened 2024-05-29
Fix query result of eager_load example in the guides. [ci-skip]

p8:
Calling `eager_load` will execute a single query with a join, not two. The original queries were probably the result of executing the code in the console, where `books = Book.eager_load(:author).limit(10) ` would execute the query.

PR #56341, opened 2025-12-10
[RF Docs] Active Record Query Interface

Ridhwana:
### Motivation / Background

This Pull Request has been created to imrove the Active Record Query Interface Guide

### Detail

- [ ] Does the example need join_table in the has_and_belongs_to_many definition? -> Probbaly not. 
_I kept it as is because I like that we show a real world holistic example_
- [x] Scope out_of_print_and_expensive could be built in terms of out_of_print.costs_more_than(500) (and moved to last?) ->  I think it's used as an example of chaining scopes. We don't need it in the initial example I think.
- [x] The AR Basics guide has a note about SQL knowledge and a couple of links to learn more, maybe a similar note would be useful here as this guide talks a lot about SQL. -> +1
- [x] There's opportunity to link to composite primary keys guide, or whenever we mention a method that's not been explained (like order) we can link to that within the guide too (we do it for default_scope for instance) -> Maybe we can link to other guides in the beginning? So searching for "composite" or "transaction" would at least find a result with a link to those other guides? Some of the interleaved examples of composite keys might be better kept in the "Composite Primary Guide" I think. The "Query Interface" guide is already pretty overwhelming.
- [x] There's a section on conditions with strings that warns about SQL Injection, could link to the security guide on the topic. (which we do link a in the subsequent section on array conditions, maybe all of that can be slightly simplified) -> +1
- [ ] Hash range conditions could use methods like all_day and beginning_of_day to simplify the samples. -> +1 
  _NOTE: I kept it as is because it avoids timezone confusion. Date.yesterday.all_day is evaluated in UTC by default, which can cause subtle differences between local and UTC times, especially if a developer’s config.time_zone is set to something else (e.g., "America/New_York" or "Pretoria"). Time.now.midnight (or, even better, Time.zone.now.midnight) makes the range clearly time-zone aware_
- [x] Samples that show boolean queries with 0 or 1 should likely be changed to false / true instead. (it's simpler to understand, independently of the underlying db) -> +1
- [x] The examples where it uses created_at as a "date" are not entirely right, especially with grouping, because it will group by timestamp. (not sure how we can show it in an agnostic way, since converting timestamp -> date is usually db specific) -> Maybe group on `status` instead, like the other examples? _grouped by status and customer_id_
- [x] The guide explains only, but there's also [except](https://edgeapi.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-except) now. (the reverse of only) -> +1
- [x] reorder explicitly mentions overriding default scope, but it's actually any previously defined order, that can be made clearer -> +1
- [x] lock and with_lock could be linked to the API. -> +1
- [x] Mentions references in the includes section, could link to the API. -> +1
- [x] The query sample for eager_load seems odd, I don't think it'd run two queries like that. -> I have an open PR for that: https://github.com/rails/rails/pull/51940
- [x] Under scopes arguments & conditionals, we show examples of scopes & class methods being similar, and note at the end one important caveat. It might be a good place to show an example that returning self in a class method would essentially work like a scope. -> +1
- [x] Duplicated default_scope example after the note. -> I think that example is part of the note, as it ends with a colon: "...applied while updating a record. E.g.:"
- [x] "The Active Record pattern implements Method Chaining" -> I think "pattern" is wrong here. -> Maybe rename it to "Fluent Interface": https://martinfowler.com/bliki/FluentInterface.html. edit: Ah, I misread this. 'Active Record pattern' is unrelated to Method Chaining indeed.
- [x] Under Find or Build a New Object, there's good examples of find_or_create_by(!). There's a more recent [create_or_find_by(!)](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by) that we could expand on the guide. There's also one caveat with the find_or_create method that it's not atomic, so race conditions could potentially create two record. -> +1
- [ ] Food for thought: locking on its own could be an interesting topic for a new guide. It's less about querying, it's kind of a somewhat related but separate concept. The current content is simple / very bare, could likely be expanded with more intricate examples." Maybe even something about transactions & locking together. I added [Active Model Transactions & Locking - Documentation Updates](https://3.basecamp.com/3076981/buckets/35498807/todos/7082186922) to discuss it further, should we decide to go ahead with a new guide.
_NOTE: deferred_ 
- [x] The first tip might be better placed under the examples? Now it obfuscates the line above it: "Code examples throughout this guide will refer to one or more of the following models:"
- [x] Also the first bookstore example could probably use a heading or maybe be moved under "Retrieving Object from the Database", as the example doesn't seem to belong to "What is the Active Record Query Interface?"?
- [x] I don't think `create_with` is a finder method, so we can probably remove it from the list.
- [x] "Retrieving Objects from the Database" seems to miss "Retrieving Multiple Objects". Should we add it with an example of `Customer.all` before the "Retrieving Multiple Objects in batches" section? Or maybe mention `where` and all the other sections later on ("Conditions", "Ordering", "Limit", "Grouping", etc)?
- [x] Maybe rename "Conditions" to the less ambiguous "Where Conditions" or "Filtering Results" or "Filtering Results with Where"? 
- [x] Maybe rename "Ordering" to the less ambiguous "Ordering Results"? 
- [x] The `distinct` section seems somewhat out-of-place in the "Selecting Specific Fields" section. Maybe move it to "Limit and Offset" and rename that section to "Limiting Results"?
- [x] In the "Joining Tables" section, the "in English" descriptions are inconsistent with emphasizing and. 
- [x] "Overriding Conditions" describes more that overriding where conditions. Not sure what a better name would be though. "Overriding Clauses", "Overriding SQL Clauses"?
- [ ] Moving locking and transactions to a separate guide seem like a good idea to me. Maybe that could also include the "Read Only" section?_NOTE: tackle as a separate follow up task_ 
- [ ] Maybe we could move the "Eager Loading" and "Running Explain" section to an "Active Record Performance" guide?
- [x] Maybe move the "Dynamic Finders" section to "Retrieving a Single Object" after the examples of `find_by`?
- [x] In the "Understanding Method Chaining" section should we add a NOTE about "how in the console queries are sometimes unexpectedly executed because it will call inspect on the result"?
- [x] The "Find or Build a New Object" section should mention [create_or_find_by](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by) as well.
- [x] The "Finding by SQL" section might have a better name as it also describes `pluck` and `ids`. Or maybe move the `pluck`, `pick` and `ids` methods to the "Retrieving Objects from the Database" section under a new "Retrieving Columns" section?
- [x] Also the `find_by_sql` should have it's own heading, like the other methods have.
- [ ] We probably need to mention Async methods like:
https://edgeapi.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-async_pluck
_NOTE: tackle as a separate follow up task,  maybe those could be added to an Advanced guide?_

### Additional information

-  Running EXPLAIN and eager loading can probably be moved to a Performance Guide as suggested. I did not attempt to do this as part of this task, but I will create a separate Basecamp task so that we can determine what else we'd move into this guide and how to give a full view of performance. Thereafter, we can extract it.

- As per Carlos's and Petrik's suggestions, I agree that we can have a separate guide [Active Model Transactions & Locking](https://3.basecamp.com/3076981/buckets/35498807/todos/7082186922). However, I'd prefer to tackle this separately. Once we tackle this task, I'll move the content from here and other places to this new guide.

### Checklist


guides/source/active_record_querying.md
```diff
@@ -9,25 +9,40 @@ After reading this guide, you will know:
...
+* How to join tables and work with data from multiple tables.
 * How to use eager loading to reduce the number of database queries needed for data retrieval.
-* How to use dynamic finder methods.
+* How to use scopes to create reusable query logic.
 * How to use method chaining to use multiple Active Record methods together.
 * How to check for the existence of particular records.
+* How to use pluck, pick, and ids to efficiently retrieve specific values.
 * How to perform various calculations on Active Record models.
-* How to run EXPLAIN on relations.
+* How to use locking mechanisms for concurrent access control.
```
p8, line 19: This is a pretty long list right now.
Can we remove some of these? For example:
```
* How to use method chaining to use multiple Active Record methods together.
* How to check for the existence of particular records.
```

Maybe combine the following into a single item?
```
* How to retrieve records in batches to efficiently process large datasets.
* How to use eager loading to reduce the number of database queries needed for data retrieval.
* How to use pluck, pick, and ids to efficiently retrieve specific values.
```
as:
```
* How to retrieve data efficiently.
```

guides/source/active_record_querying.md
```diff
@@ -9,25 +9,40 @@ After reading this guide, you will know:
...
+* How to run .explain on relations.
```
p8: We don't prepend the pluck and pick methods with a period, so this should probably not use as period either?

```suggestion
* How to run explain on relations.
```

Maybe also describe what explain is used for?
```suggestion
* How to run explain to analyze queries.
```

guides/source/active_record_querying.md
```diff
@@ -84,6 +99,8 @@ class Supplier < ApplicationRecord
 end
 ```
 
+NOTE: All of the following models use `id` as the primary key, unless specified otherwise.
```
p8: This note is now defined under the models (or between when taking into account the UML diagram):
```suggestion
NOTE: All of the models use `id` as the primary key, unless specified otherwise.
```
Or maybe:
```suggestion
NOTE: These models use `id` as the primary key, unless specified otherwise.
```

guides/source/active_record_querying.md
p8: Shouldn't we only list the methods from [FinderMethods](https://edgeapi.rubyonrails.org/classes/ActiveRecord/FinderMethods.html)
here? All the conditioning, ordering, limiting, joining methods will be handled later.

The following is not complete:
```suggestion
* [`find_by`][]
* [`find_sole_by`][]
* [`first`][]
* [`last`][]
* [`second`][]
* [`second_to_last`][]
* [`sole`][]
* [`take`][]
```

guides/source/active_record_querying.md
```diff
@@ -183,12 +202,13 @@ SELECT * FROM customers WHERE (customers.id = 10) LIMIT 1
 
 The `find` method will raise an `ActiveRecord::RecordNotFound` exception if no matching record is found.
 
-You can also use this method to query for multiple objects. Call the `find` method and pass in an array of primary keys. The return will be an array containing all of the matching records for the supplied _primary keys_. For example:
+You can also use this method to query for multiple objects. Call the `find` method and pass in an array of primary keys. The return value will be an array containing all of the matching records for the supplied _primary keys_. For example:
 
 ```irb
 # Find the customers with primary keys 1 and 10.
 irb> customers = Customer.find([1, 10]) # OR Customer.find(1, 10)
```
p8: Can you replace all irb prompts with `store(dev)`?
The app name and env is now the default, and store seems to make sense for a bookstore.
```suggestion
store(dev)> customers = Customer.find([1, 10]) # OR Customer.find(1, 10)
```

guides/source/active_record_querying.md
p8: Could we remove the examples of composite primary keys here and link to the composite primary guide instead?

```suggestion
If your table uses a [composite primary key](active_record_composite_primary_keys.html), you'll need to [pass in an array to find](active_record_composite_primary_keys.html#using-find) a single item. 
```

guides/source/active_record_querying.md
```diff
@@ -296,7 +320,7 @@ The SQL equivalent of the above is:
 SELECT * FROM customers ORDER BY customers.id ASC LIMIT 3
 ```
 
-Models with composite primary keys will use the full composite primary key for ordering.
+Models with [composite primary keys](active_record_composite_primary_keys.html) will use the full composite primary key for ordering.
```
p8: Could we remove the examples of composite primary keys here and link to the composite primary guide instead?

The examples can be moved to the composite primary keys guide.
First isn't describe there yet.

guides/source/active_record_querying.md
```diff
@@ -464,11 +489,72 @@ irb> Customer.find_by(id: customer.id_value) # Customer.find_by(id: 10)
...
 
+#### Dynamic Finders
+
+For every field (also known as an attribute) you define in your table,
+Active Record provides a finder method. If you have a field called `first_name` on your `Customer` model for example,
+you get the instance method `find_by_first_name` for free from Active Record.
+If you also have a `locked` field on the `Customer` model, you also get `find_by_locked` method.
+
+You can specify an exclamation point (`!`) on the end of the dynamic finders
+to get them to raise an `ActiveRecord::RecordNotFound` error if they do not return any records, like `Customer.find_by_first_name!("Ryan")`
```
p8:
````suggestion
You can specify an exclamation point (`!`) on the end of the dynamic finders
to get them to raise an `ActiveRecord::RecordNotFound` error if they do not return any records:

```irb
store(dev> Customer.find_by_first_name!("Ryan") # raises an ActiveRecord::RecordNotFound
```
````

File: guides/source/active_record_querying.md [hunk 7]
```diff
@@ -464,11 +489,72 @@ irb> Customer.find_by(id: customer.id_value) # Customer.find_by(id: 10)
...
+
+If you want to find both by `first_name` and `orders_count`, you can chain these finders together by simply typing "`and`" between the fields.
+For example, `Customer.find_by_first_name_and_orders_count("Ryan", 5)`.
```
p8: In the previous example Ryan wasn't found. Another example found Bhumi.
````suggestion

```irb
store(dev)> Customer.find_by_first_name_and_orders_count("Bhumi", 5)
=> #<Customer id: 25, first_name: "Bhumi">
```
````

File: guides/source/active_record_querying.md [hunk 8]
```diff
@@ -464,11 +489,72 @@ irb> Customer.find_by(id: customer.id_value) # Customer.find_by(id: 10)
 
 [`id_value`]: https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema.html#method-i-id_value
 
+#### Dynamic Finders
+
+For every field (also known as an attribute) you define in your table,
+Active Record provides a finder method. If you have a field called `first_name` on your `Customer` model for example,
+you get the instance method `find_by_first_name` for free from Active Record.
```
p8: To me this section could be clearer with irb code examples:
````suggestion
you get the instance method `find_by_first_name` for free from Active Record.

```irb
store(dev)> Customer.find_by_first_name("Bhumi")
=> #<Customer id: 25, first_name: "Bhumi">
```

````

File: guides/source/active_record_querying.md
Comment (p8) [reply guides/source/active_record_querying.md]: Maybe we should mention the list isn't complete and only list the methods we describe in this guide? The links could also link to the guide sections?

File: guides/source/active_record_querying.md [hunk 9]
```diff
@@ -359,13 +383,15 @@ The SQL equivalent of the above is:
 SELECT * FROM customers ORDER BY customers.store_id DESC, customers.id DESC LIMIT 1
```
p8: Could we move these examples of composite primary keys to the "Composite Primary Keys"  guide and link to that instead?

```suggestion
Models with [composite primary keys](active_record_composite_primary_keys.html) will use the [full composite primary key for ordering](link).
```

guides/source/active_record_querying.md
```diff
@@ -464,11 +489,72 @@ irb> Customer.find_by(id: customer.id_value) # Customer.find_by(id: 10)
 
 [`id_value`]: https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema.html#method-i-id_value
```
p8: The "Composite Primary Key" guide [links to this section](
https://edgeguides.rubyonrails.org/active_record_composite_primary_keys.html#conditions-with-id).

As this is about Composite Primary keys, I think it makes more sense to move this section to the "Composite Primary Key" guide instead and link to that here.

guides/source/active_record_querying.md
```diff
@@ -439,7 +465,7 @@ Customer.where(first_name: "does not exist").take!
 When specifying conditions on methods like [`find_by`][] and [`where`][], the use of `id` will match against
 an `:id` attribute on the model. This is different from [`find`][], where the ID passed in should be a primary key value.
 
-Take caution when using `find_by(id:)` on models where `:id` is not the primary key, such as composite primary key models.
+Take caution when using `find_by(id:)` on models where `:id` is not the primary key, such as [composite primary key](active_record_composite_primary_keys.html) models.
```
p8: Maybe make this a warning?

```suggestion
WARNING: Take caution when using `find_by(id:)` on models where `:id` is not the primary key, such as [composite primary key](active_record_composite_primary_keys.html) models.
```

guides/source/active_record_querying.md
```diff
@@ -464,11 +489,72 @@ irb> Customer.find_by(id: customer.id_value) # Customer.find_by(id: 10)
...
+```
+
+Since `all` returns an `ActiveRecord::Relation` and relations are lazy-loaded, calling `all` first is optional and doesn't change the query behavior.
+
+The SQL equivalent is:
+
+```sql
+SELECT * FROM customers WHERE (customers.active = true)
+```
+
```
p8: The SQL equivalent probably needs to be mentioned first?
I also wonder if we should mention that the console always calls inspect on the return value. This can be confusing when calling `all`.

````suggestion
The SQL equivalent is:

```sql
SELECT * FROM customers WHERE (customers.active = true)
```

Since `all` returns an `ActiveRecord::Relation` and relations are lazy-loaded, calling `all` first is optional and doesn't change the query behavior.

NOTE: You might have noticed in the console `Customer.all` does execute the query. In the console the return value of the last statement is shown by calling `inspect` on the statement. Running `Customer.all` will show `Customer.all.inspect` which executes the query and loads the relations.
````

File: guides/source/active_record_querying.md
Comment (p8) [inline guides/source/active_record_querying.md]: The `finish` id is also included in the result.
```suggestion
The `finish` option allows specifying the ending ID of the records to be retrieved. The code below shows the case of retrieving customers in batches, up to including the customer with ID: 7000:
```

p8:
```suggestion
Customer.find_each(start: 2000, finish: 9999) do |customer|
```

p8: The finish id is included in the result, so I think it's clearer if we change it to `9999`. The next call could be `start: 10000, finish: 19999`.
```suggestion
For example, to send newsletters only to customers with the primary key starting from 2000 up to including 9999:
```

*999 is also used in the API docs.

guides/source/active_record_querying.md
```diff
@@ -634,50 +721,56 @@ end
 
 The `error_on_ignore` option overrides the application config to specify if an error should be raised when a specific order is present in the relation.
 
-Conditions
-----------
+Filtering Records
+-----------------
 
 The [`where`][] method allows you to specify conditions to limit the records returned, representing the `WHERE`-part of the SQL statement. Conditions can either be specified as a string, array, or hash.
```
p8: As this section is about filtering we should probably use it here as well:
```suggestion
The [`where`][] method allows you to specify conditions to filter the records returned, representing the `WHERE`-part of the SQL statement. Conditions can either be specified as a string, array, or hash.
```

guides/source/active_record_querying.md
```diff
@@ -634,50 +721,56 @@ end
...
+```
+
+This will find all books where the `title` field value is 'Introduction to Algorithms'.
+
+WARNING: Building your own conditions as pure strings can leave you vulnerable to SQL injection exploits. For example, `Book.where("title LIKE '%#{params[:title]}%'")` is not safe. See the next section for the preferred way to handle conditions using an array. For more background, see the [Ruby on Rails Security Guide on SQL injection](security.html#sql-injection).
 
 ### Array Conditions
 
-Now what if that title could vary, say as an argument from somewhere? The find would then take the form:
+If an object attribute is variable or dependent on an argument, the find could then take the form:
```
p8: This sentence is a bit confusing to me. Isn't passing an argument the only use case?
Maybe:
```suggestion
If a condition is dependent on an argument, the condition could then take the following form:
```

guides/source/active_record_querying.md
```diff
@@ -634,50 +721,56 @@ end
...
 
 ```ruby
 Book.where("title = ?", params[:title])
 ```
 
 Active Record will take the first argument as the conditions string and any additional arguments will replace the question marks `(?)` in it.
```
p8: Maybe we can explain here that this form protects against injection attacks? (Some sentences copied from the API docs).
And show some examples?

````suggestion
Active Record will take the first argument as the conditions string and any additional arguments will replace the question marks `(?)` in it.

To avoid SQL injection attacks Active Record will convert from the ruby type to the database type where needed.

Using a unsafe string condition would allow getting all books (`OR '1' = '1'` will match any book) :

```ruby
unsafe_title = "a' or '1' = '1"
Book.where("title = '#{unsafe_title}'")
# SELECT `books`.* FROM `books` WHERE (title = 'a' or '1'='1')
```

Whereas the array condition will escape the SQL inject:

```ruby
Book.where("title = ?", unsafe_title)
# SELECT `books`.* FROM `books` WHERE `books`.`title` = 'a\\' or \\'1\\'=\\'1'
```

````

File: guides/source/active_record_querying.md [hunk 16]
```diff
@@ -634,50 +721,56 @@ end
...
+WARNING: Building your own conditions as pure strings can leave you vulnerable to SQL injection exploits. For example, `Book.where("title LIKE '%#{params[:title]}%'")` is not safe. See the next section for the preferred way to handle conditions using an array. For more background, see the [Ruby on Rails Security Guide on SQL injection](security.html#sql-injection).
 
 ### Array Conditions
 
-Now what if that title could vary, say as an argument from somewhere? The find would then take the form:
+If an object attribute is variable or dependent on an argument, the find could then take the form:
 
 ```ruby
 Book.where("title = ?", params[:title])
 ```
```
p8, line 834: I'm also wondering if it's clear to beginners that by array we also mean a splatted array? Should we pass an actual array to the first example?

````suggestion
```ruby
Book.where(["title = ?", params[:title]])
```

You don't have to pass an actual array. A list of arguments is supported as well:

```ruby
Book.where("title = ?", params[:title])
```
````

File: guides/source/active_record_querying.md [hunk 17]
```diff
@@ -634,50 +721,56 @@ end
...
 
-If you want to specify multiple conditions:
+You can also specify multiple conditions:
 
 ```ruby
 Book.where("title = ? AND out_of_print = ?", params[:title], false)
 ```
 
-In this example, the first question mark will be replaced with the value in `params[:title]` and the second will be replaced with the SQL representation of `false`, which depends on the adapter.
+In the above example, the first question mark will be replaced with the value in `params[:title]` and the second will be replaced with the SQL representation of `false`, which depends on the adapter.
```
p8: Maybe highlight that the value will be escaped?
```suggestion
In the above example, the first question mark will be replaced with the escaped value in `params[:title]` and the second will be replaced with the SQL representation of `false`, which depends on the adapter.
```

guides/source/active_record_querying.md
```diff
@@ -634,50 +721,56 @@ end
...
+instead of this code:
 
 ```ruby
 Book.where("title = #{params[:title]}")
 ```
 
-because of argument safety. Putting the variable directly into the conditions string will pass the variable to the database **as-is**. This means that it will be an unescaped variable directly from a user who may have malicious intent. If you do this, you put your entire database at risk because once a user finds out they can exploit your database they can do just about anything to it. Never ever put your arguments directly inside the conditions string.
-
-TIP: For more information on the dangers of SQL injection, see the [Ruby on Rails Security Guide](security.html#sql-injection).
+WARNING:Adding a variable directly into the conditions string sends it to the database **as-is**, without any escaping or protection. If that value comes from user input, it can introduce malicious SQL and put your entire database at risk. For that reason, you should avoid placing arguments directly inside the conditions string.
```
p8: This could probably be removed because it's already mentioned earlier in the String condition section?
```suggestion
```

guides/source/active_record_querying.md
```diff
@@ -688,7 +781,7 @@ Book.where("created_at >= :start_date AND created_at <= :end_date",
   { start_date: params[:start_date], end_date: params[:end_date] })
```
p8: This example can now be written as `Book.where(created_at: params[:start_date]..params[:end_date])`. 

So maybe we can use title and out_of_print instead (like an earlier example) and also remove the curly braces:
```suggestion
Book.where("title = :title AND out_of_print = :out_of_print",
title: params[:title], out_of_print: false)
```

guides/source/active_record_querying.md
```diff
@@ -943,96 +1035,95 @@ The SQL query used by this find call will be somewhat like:
...
+And this will give you a single `Order` object for each unique status value in the database.
 
 The SQL that would be executed would be something like this:
 
 ```sql
-SELECT created_at
+SELECT status
 FROM orders
-GROUP BY created_at
+GROUP BY status
```
p8: I seem to recall we try to use indentation for multi line SQL:
```suggestion
SELECT status
  FROM orders
  GROUP BY status
```
There are more SQL examples in this guide where this applies.

guides/source/active_record_querying.md
```diff
@@ -943,96 +1035,95 @@ The SQL query used by this find call will be somewhat like:
...
-If you would like to only grab a single record per unique value in a certain field, you can use [`distinct`][]:
+The limit method specifies how many records should be returned, while offset determines how many records to skip before retrieving results. For example:
 
 ```ruby
-Customer.select(:last_name).distinct
+Customer.limit(5)
 ```
 
-This would generate SQL like:
+This example will return a maximum of 5 customers, and as it specifies no offset first 5 in the table will be returned. The SQL it executes looks like this:
```
p8:
```suggestion
This example will return a maximum of 5 customers, and as it specifies no offset, the first 5 in the table will be returned. The SQL it executes looks like this:
```

guides/source/active_record_querying.md
```diff
@@ -943,96 +1035,95 @@ The SQL query used by this find call will be somewhat like:
...
 
-To apply a `GROUP BY` clause to the SQL fired by the finder, you can use the [`group`][] method.
+If you want to group records, you can use the [`group`][] method to apply a `GROUP BY` clause to the SQL generated by the relation.
 
-For example, if you want to find a collection of the dates on which orders were created:
+For example, if you want to find a collection of orders grouped by status:
 
 ```ruby
-Order.select("created_at").group("created_at")
+Order.select("status").group("status")
```
p8: It might be clearer if we ignore the `select` in this example (and the following SQL), as that is not required.
```suggestion
Order.group("status")
```

guides/source/active_record_querying.md
```diff
@@ -1095,25 +1190,73 @@ SELECT * FROM books WHERE id > 100 LIMIT 20
...
 
 ```ruby
 Book.order("id desc").merge(Book.unscope(:order))
-# SELECT books.* FROM books
+```
+
+This will generate the following SQL:
+
+```sql
+SELECT books.* FROM books ORDER BY id desc
```
p8: I think the order should be removed here, as it's unscoped:
```suggestion
SELECT books.* FROM books
```

guides/source/active_record_querying.md
```diff
@@ -1095,25 +1190,73 @@ SELECT * FROM books WHERE id > 100 LIMIT 20
...
+Book.where(out_of_print: true).unscoped.all
+```
+
+Both of the above will generate the following SQL:
+
+```sql
+SELECT books.* FROM books
+```
+
+`unscoped` can also accept a block:
```
p8: Maybe explain what happens here (I've copied the sentence from the API docs):
```suggestion
`unscoped` can also accept a block. All queries inside the block will not use the previously set scopes.
```

guides/source/active_record_querying.md
```diff
@@ -1095,25 +1190,73 @@ SELECT * FROM books WHERE id > 100 LIMIT 20
...
+
+```ruby
+Book.unscoped { Book.out_of_print }
```
p8: Maybe add a scope that gets unscoped here?
```suggestion
Book.in_print.unscoped { Book.out_of_print }
```

guides/source/active_record_querying.md
p8: Can we add a `default_scope` to Book instead?
This makes the examples a bit clearer because we don't need the author:
```suggestion
class Book < ApplicationRecord
  default_scope { order(year_published: :desc) }
end
```

guides/source/active_record_querying.md
```diff
@@ -1172,14 +1346,14 @@ And you execute this:
 Author.find(10).books
```
p8:
```suggestion
Book.all
```

guides/source/active_record_querying.md
```diff
@@ -1172,14 +1346,14 @@ And you execute this:
...
 ```
 
-The SQL that would be executed:
+This will generate the following SQL:
 
 ```sql
 SELECT * FROM authors WHERE id = 10 LIMIT 1
 SELECT * FROM books WHERE author_id = 10 ORDER BY year_published DESC
```
p8:
```suggestion
SELECT * FROM books ORDER BY year_published DESC
```

guides/source/active_record_querying.md
```diff
@@ -1172,14 +1346,14 @@ And you execute this:
...
 ```
 
-You can using the `reorder` clause to specify a different way to order the books:
+You can use the `reorder` clause to specify a different way to order the books:
 
 ```ruby
 Author.find(10).books.reorder("year_published ASC")
```
p8:
```suggestion
Book.reorder("year_published ASC")
```

guides/source/active_record_querying.md
```diff
@@ -1192,6 +1366,14 @@ SELECT * FROM authors WHERE id = 10 LIMIT 1
 SELECT * FROM books WHERE author_id = 10 ORDER BY year_published ASC
```
p8:
```suggestion
SELECT * FROM books ORDER BY year_published ASC
```

guides/source/active_record_querying.md
```diff
@@ -1257,13 +1439,13 @@ The [`regroup`][] method overrides an existing, named `group` condition. For exa
 Book.group(:author).regroup(:id)
```
p8: I think this should reference the foreign_key column:
```suggestion
Book.group(:author_id).regroup(:id)
```

guides/source/active_record_querying.md
```diff
@@ -1257,13 +1439,13 @@ The [`regroup`][] method overrides an existing, named `group` condition. For exa
...
 
 ```sql
 SELECT * FROM books GROUP BY id
 ```
 
-If the `regroup` clause is not used, the group clauses are combined together:
+If a regular `group` is used instead of the `regroup` clause, the group clauses are combined together:
 
 ```ruby
 Book.group(:author).group(:id)
```
p8:
```suggestion
Book.group(:author_id).group(:id)
```

guides/source/active_record_querying.md
p8:
```suggestion
SELECT * FROM books GROUP BY author_id, id
```

guides/source/active_record_querying.md
```diff
@@ -1345,8 +1529,11 @@ c2 = Customer.find(1)
...
 
+c1.lock_version # => 1
+c2.lock_version # => 0
+
 c2.first_name = "Michael"
-c2.save # Raises an ActiveRecord::StaleObjectError
+c2.save  # Raises an ActiveRecord::StaleObjectError
 ```
 
 You're then responsible for dealing with the conflict by rescuing the exception and either rolling back, merging, or otherwise apply the business logic needed to resolve the conflict.
```
p8: I think the verb form here should be the same as "rolling back" and "merging":
```suggestion
You're then responsible for dealing with the conflict by rescuing the exception and either rolling back, merging, or otherwise applying the business logic needed to resolve the conflict.
```

guides/source/active_record_querying.md
p8: Not sure if a comment is the best here, but I think we need to mention that the book is reloaded here:
```suggestion
# Reload book with a lock before yielding.
book.with_lock do
```

guides/source/active_record_querying.md
```diff
@@ -1409,10 +1596,13 @@ end
 Joining Tables
 --------------
 
+Joining tables allows you to retrieve records from multiple tables in a single query, for example, fetching books together with their authors.
+
 Active Record provides two finder methods for specifying `JOIN` clauses on the
-resulting SQL: `joins` and `left_outer_joins`.
-While `joins` should be used for `INNER JOIN` or custom queries,
-`left_outer_joins` is used for queries using `LEFT OUTER JOIN`.
+resulting SQL; `joins` and `left_outer_joins`:
```
p8: In other sections we link the methods the first time they are mentioned.
Maybe:
```suggestion
resulting SQL; [`joins`][] and [`left_outer_joins`][]:
```

guides/source/active_record_querying.md
p8: If we link it earlier we could remove the link here?
```suggestion
There are multiple ways to use the `joins` method.
```

p8, line 1980: I think it would be nice to have a single introduction sentence for the join examples:

```suggestion

Pass the name of the association to join a single table:

```

guides/source/active_record_querying.md
```diff
@@ -1444,54 +1635,59 @@ All of the following will produce the expected join queries using `INNER JOIN`:
...
 SELECT books.* FROM books
   INNER JOIN reviews ON reviews.book_id = books.id
 ```
 
-Or, in English: "return a Book object for all books with reviews". Note that you will see duplicate books if a book has more than one review.  If you want unique books, you can use `Book.joins(:reviews).distinct`.
+The SQL query will return a Book object for all books with reviews.
+
+NOTE: You will see duplicate books if a book has more than one review.  If you want unique books, you can use `Book.joins(:reviews).distinct`.
 
 #### Joining Multiple Associations
```
p8, line 1997:
```suggestion

Pass multiple association names to join multiple tables:

```

guides/source/active_record_querying.md
```diff
@@ -1444,54 +1635,59 @@ All of the following will produce the expected join queries using `INNER JOIN`:
...
   INNER JOIN reviews ON reviews.book_id = books.id
 ```
 
-Or, in English: "return all books that have an author and at least one review". Note again that books with multiple reviews will show up multiple times.
+The SQL query will return all books that have an author _and_ at least one review.
+
+NOTE: You will see duplicate books if a book has more than one review.  If you want unique books, you can use `Book.joins(:reviews).distinct`.
+
 
 ##### Joining Nested Associations (Single Level)
```
p8, line 2024:
```suggestion

Pass a hash of association names to join tables on other joined tables:

```

guides/source/active_record_querying.md
```diff
@@ -1444,54 +1635,59 @@ All of the following will produce the expected join queries using `INNER JOIN`:
...
 ```sql
 SELECT books.* FROM books
   INNER JOIN reviews ON reviews.book_id = books.id
   INNER JOIN customers ON customers.id = reviews.customer_id
 ```
 
-Or, in English: "return all books that have a review by a customer."
+The SQL query will return all books that have a review by a customer.
 
 ##### Joining Nested Associations (Multiple Level)
```
p8, line 2046:
```suggestion

For more complex joining, use a combination of hashes and arrays:

```

guides/source/active_record_querying.md
```diff
@@ -1444,54 +1635,59 @@ All of the following will produce the expected join queries using `INNER JOIN`:
...
 SELECT books.* FROM books
   INNER JOIN reviews ON reviews.book_id = books.id
 ```
 
-Or, in English: "return a Book object for all books with reviews". Note that you will see duplicate books if a book has more than one review.  If you want unique books, you can use `Book.joins(:reviews).distinct`.
+The SQL query will return a Book object for all books with reviews.
+
+NOTE: You will see duplicate books if a book has more than one review.  If you want unique books, you can use `Book.joins(:reviews).distinct`.
 
 #### Joining Multiple Associations
```
p8: I think this heading level should be the same as "Joining a Single Association"

```suggestion
##### Joining Multiple Associations
```

guides/source/active_record_querying.md
p8: Maybe first explain the "Inner join" again for clarity?
```suggestion
Inner joins will only return records that have the associated records.

If you want to select a set of records, whether or not they have associated
records, you can use the [`left_outer_joins`][] method.
```

guides/source/active_record_querying.md
```diff
@@ -1502,7 +1698,7 @@ SELECT authors.* FROM authors
 INNER JOIN suppliers ON suppliers.id = books.supplier_id
```
p8:
```suggestion
  INNER JOIN suppliers ON suppliers.id = books.supplier_id
```

guides/source/active_record_querying.md
p8: Maybe expand slightly on this explanation and highlight that an empty where is required?
```suggestion
To use `where.associated`, begin with an empty `where` followed by `associated` with the association name:
```

guides/source/active_record_querying.md
```diff
@@ -1573,37 +1770,42 @@ To use `where.associated`:
...
-WHERE reviews.id IS NOT NULL
+  INNER JOIN reviews ON reviews.customer_id = customers.id
+  WHERE reviews.id IS NOT NULL
 ```
 
-Which means "return all customers that have made at least one review".
+The SQL query will return all customers that have made at least one review.
 
-To use `where.missing`:
+You can use `where.missing` to select records that do not have an association:
```
p8: Maybe highlight that `where.missing` is the opposite?
```suggestion
`where.missing` is the opposite of `where.associated`.
You can use `where.missing` to select records that do not have an association:
```

guides/source/active_record_querying.md
```diff
@@ -1573,37 +1770,42 @@ To use `where.associated`:
...
-WHERE reviews.id IS NULL
+  LEFT OUTER JOIN reviews ON reviews.customer_id = customers.id
+  WHERE reviews.id IS NULL
 ```
 
-Which means "return all customers that have not made any reviews".
+The SQL query will return all customers that have not made any reviews.
 
+If a join is already defined `associated` will use that join instead:
+```ruby
```
p8, line 2139: This is missing a linebreak:

Image

```suggestion

```ruby
```

guides/source/active_record_querying.md
p8: I think we should probably explain the `N` here somewhere.
Maybe someting like:
```suggestion
Retrieving a list of records N (where N is a number greater that 1) in a single query can sometimes trigger N extra queries: one for each record.
Consider the following code, which finds 10 (N) books and prints their authors' last_name:
```

guides/source/active_record_querying.md
```diff
@@ -1573,37 +1770,42 @@ To use `where.associated`:
...
+# associated will use LEFT JOIN for this query instead of using JOIN
+Post.left_joins(:author).where.associated(:author)
+```
 
 Eager Loading Associations
 --------------------------
 
-Eager loading is the mechanism for loading the associated records of the objects returned by `Model.find` using as few queries as possible.
+Eager loading is the mechanism for loading the associated records of the objects returned by `ActiveRecord::Relation` using as few queries as possible.
```
p8: Sometimes it's better to have separate queries instead of as single join, so eager loading is about 'most performant' instead of as 'few queries'.
```suggestion
Eager loading is the mechanism for loading the associated records of the objects returned by `ActiveRecord::Relation` using the most performant queries possible.
```

guides/source/active_record_querying.md
```diff
@@ -1653,7 +1855,7 @@ SELECT authors.* FROM authors
 
 #### Eager Loading Multiple Associations
 
-Active Record lets you eager load any number of associations with a single `Model.find` call by using an array, hash, or a nested hash of array/hash with the `includes` method.
+Active Record lets you eager load any number of associations with a single `ActiveRecord::Relation` call by using an array, hash, or a nested hash of array/hash with the `includes` method.
 
 ##### Array of Multiple Associations
```
p8: This section is pretty heading heavy, maybe change them to regular sentences instead?
Unlike short statements with headings, this to me also reads more like guiding text:
```suggestion
To eager load multiple associations pass an array of association names:
```

guides/source/active_record_querying.md
```diff
@@ -1631,7 +1833,7 @@ The methods are:
```
p8, line 2161: Not sure what would be a good way to describe it here, but `includes` is the preferred method as it will do either a `preload` or an `eager_load` depending on the conditions.
Maybe an info block?
```suggestion

INFO: Prefer using `includes`, as it is a higher level method that will do either an `preload` or an `eager_load` depending on the situation.

```

guides/source/active_record_querying.md
p8: Maybe mention that this is the same as `includes` for the simple cases. Otherwise readers will have a hard time finding the difference.
```suggestion
With `preload`, Active Record loads each specified association using one query per association. This is exactly the same as what `includes` will do when there are no conditions.
```

p8: I think we show most configuration overrides as code example blocks:
````suggestion
To enable strict loading for all relations, change the configuration to `true`:
```ruby
config.active_record.strict_loading_by_default = true
```
````

Comment (p8) [inline guides/source/active_record_querying.md]:
````suggestion
To send violations to the logger instead, change it to `:log`:
```ruby
config.active_record.action_on_strict_loading_violation = :log
```
````

File: guides/source/active_record_querying.md [hunk 44]
```diff
@@ -1953,9 +2164,7 @@ irb> Book.unscoped.new
 => #<Book id: nil, out_of_print: nil>
 ```
 
-Be aware that, when given in the `Array` format, `default_scope` query arguments
-cannot be converted to a `Hash` for default attribute assignment. E.g.:
-
+Be aware that when scope arguments are given as an `Array`, `default_scope` cannot convert the arguments to a `Hash` for default attribute assignment. For example:
 ```ruby
```
p8, line 2482: This needs an extra linebreak:

Image

````suggestion

```ruby
````

File: guides/source/active_record_querying.md [hunk 45]
```diff
@@ -1936,9 +2147,9 @@ class Book < ApplicationRecord
 end
 ```
 
-NOTE: The `default_scope` is also applied while creating/building a record
+The `default_scope` is also applied while creating/building a record
 when the scope arguments are given as a `Hash`. It is not applied while
-updating a record. E.g.:
+updating a record. For example, if you have a `default_scope` that sets `out_of_print` to `false`, and you create a new book with the `out_of_print` attribute set to `true`, the `default_scope` will be applied:
```
p8: The sentence about "not applying the default_scope while updating a record", is easily overlooked.
Maybe add a linebreak?
```suggestion
updating a record.

For example, if you have a `default_scope` that sets `out_of_print` to `false`, and you create a new book with the `out_of_print` attribute set to `true`, the `default_scope` will be applied:
```

guides/source/active_record_querying.md
```diff
@@ -1971,7 +2180,17 @@ irb> Book.new
 
 ### Merging of Scopes
 
-Just like `where` clauses, scopes are merged using `AND` conditions.
+You can chain small, reusable scopes to build more specific queries. Scopes can call other scopes:
+
+```ruby
```
p8, line 2500: Maybe highlight the line that combines other scopes:
```suggestion
```ruby#4
```

guides/source/active_record_querying.md
p8: `recent` doesn't seem to get used, so we can probably delete it?
```suggestion
```

guides/source/active_record_querying.md
```diff
@@ -1971,7 +2180,17 @@ irb> Book.new
 
 ### Merging of Scopes
 
-Just like `where` clauses, scopes are merged using `AND` conditions.
+You can chain small, reusable scopes to build more specific queries. Scopes can call other scopes:
```
p8: In this section we first combine two scopes into a new scope. Later we call multiple scopes on a single model.

I think it makes more sense to swap these: first show an example of combining scopes, then show an example of creating a new scope for these combined scopes (using the same example fir clarity?).

guides/source/active_record_querying.md
```diff
@@ -2023,12 +2241,11 @@ SELECT books.* FROM books WHERE (year_published >= 1969)
 irb> Book.in_print
 SELECT books.* FROM books WHERE (year_published >= 1969) AND books.out_of_print = false
 
-irb> Book.where('price > 50')
+irb> Book.where("price > 50")
 SELECT books.* FROM books WHERE (year_published >= 1969) AND (price > 50)
```
p8: Not sure this third example adds much
```suggestion
```

If we really want another example I would add an example that uses the same columns as the default_scope:
```suggestion
irb> Book.where(year_published: 2020)
SELECT books.* FROM books WHERE (year_published >= 1969) AND (year_published = 2020)
```

guides/source/active_record_querying.md
p8, line 2630: This example doesn't add much I think.
It uses the uncommon `load` method and doesn't show the resulting query.
The next `irb` block is much clearer.
```suggestion
```

Another option would be showing a Book with a default scope (although this example is shown a lot). This would also clarify the `unscoped` examples, particularly the block example.

```suggestion
```ruby
  default_scope { where(out_of_print: false) }
end
```

```

Comment (p8) [inline guides/source/active_record_querying.md]: Maybe clarify the methods are dynamic?
Also, this whole section is duplicated in 3.1.6.
```suggestion
Dynamic Finder Methods
----------------------
```

p8:
````suggestion
For example:

```ruby
Customer.find_by_first_name_and_orders_count("Ryan", 5)
```
````

Comment (p8) [inline guides/source/active_record_querying.md]:
```suggestion
If you want to find both by `first_name` and `orders_count`, you can chain these finders together by simply typing `_and_` between the fields.
```

p8: To me these examples are a bit easier to read with code example blocks (Ah, I already made the same suggestions in 3.1.6 😄)

````suggestion
You can specify an exclamation point (`!`) on the end of the dynamic finders
to get them to raise an `ActiveRecord::RecordNotFound` error if they do not return any records, like:

```ruby
Customer.find_by_first_name!("Ryan")
```
````

File: guides/source/active_record_querying.md [hunk 49]
```diff
@@ -2129,23 +2346,17 @@ Enums
 
 An enum lets you define an Array of values for an attribute and refer to them by name.  The actual value stored in the database is an integer that has been mapped to one of the values.
```
p8: Maybe start with explaining the use-case for `Enums`?
```suggestion
Sometimes you might want to restrict the value of an attribute to a predefined list of values.

An enum lets you define an Array of values for an attribute and refer to them by name.  The actual value stored in the database is an integer that has been mapped to one of the values.
```

guides/source/active_record_querying.md
```diff
@@ -2154,7 +2365,7 @@ irb> Order.not_shipped
 => #<ActiveRecord::Relation> # all orders with status != :shipped
 ```
 
-These instance methods are created automatically and query whether the model has that value for the `status` enum:
+Predicate methods are created automatically for each enum value that query whether the model has that value for the `status` enum:
```
p8: Maybe explain predicate methods?
```suggestion
Predicate methods (methods that return true of false) are created automatically for each enum value that query whether the model has that value for the `status` enum:
```

Or maybe change the wording the make it clearer they return true of false?

```suggestion
Predicate methods are created automatically for each enum value and return whether the model has that value for the `status` enum:
```

guides/source/active_record_querying.md
```diff
@@ -2174,24 +2385,23 @@ UPDATE "orders" SET "status" = ?, "updated_at" = ? WHERE "orders"."id" = ?  [["s
 => true
 ```
 
-Full documentation about enums can be found [here](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html).
+You can read more about enums in the [ActiveRecord::Enum documentation](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html).
 
 [`enum`]: https://api.rubyonrails.org/classes/ActiveRecord/Enum.html#method-i-enum
 
 Understanding Method Chaining
```
p8: This section feels a bit late in the guide. There are multiple sections that use method chaining, then we explain other concepts like eager loading, scopes and enums, before we end up here.
To me it feels like it should be a sub section of "Retrieving Objects from the Database"? What do you think?

guides/source/active_record_querying.md
```diff
@@ -2362,8 +2632,10 @@ This method will return an instance of `ActiveRecord::Result` class and calling
 object would return you an array of hashes where each hash indicates a record.
 
 ```irb
-irb> Customer.lease_connection.select_all("SELECT first_name, created_at FROM customers WHERE id = '1'").to_a
-=> [{"first_name"=>"Rafael", "created_at"=>"2012-11-10 23:23:45.281189"}, {"first_name"=>"Eileen", "created_at"=>"2013-12-09 11:22:35.221282"}]
+irb> Customer.lease_connection.select_all("SELECT first_name, created_at FROM customers WHERE id = \"1\"").to_a
+```
```
p8: The example hasn't ended here yet, so these mess up the markup:
```suggestion
```

guides/source/active_record_querying.md
p8: This example is a bit confusing to me. Of course pluck doesn't work on `name`.
I think we should highlight that overriding attribute methods work with `map` but not with `pluck`:
````suggestion
```ruby
class Customer < ApplicationRecord
  def first_name
    "I am #{super}"
  end
end
```

```irb
irb> Customer.select(:first_name).map &:first_name
=> ["I am David", "I am Jeremy", "I am Jose"]

irb> Customer.pluck(:first_name)
=> ["David", "Jeremy", "Jose"]
```
````

File: guides/source/active_record_querying.md [hunk 53]
```diff
@@ -2340,14 +2551,73 @@ irb> nina.save
...
+```irb
+irb> Customer.create_or_find_by!(first_name: "Andy", orders_count: nil)
+ActiveRecord::RecordInvalid: Validation failed: Orders count can't be blank
+```
+
+However, if the failure is due to a uniqueness constraint violation, it will find and return the existing record (just like `create_or_find_by`), rather than raising an exception.
+
+[`create_or_find_by!`]: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by-21
+
+Finding Records and Values
```
p8: To me it seems this section is should be shown a lot earlier.
I think it would fit as a subsection of "Retrieving Objects from the Database", as these methods also describe getting objects from the database.
What do you think?

guides/source/active_record_querying.md
```diff
@@ -2232,45 +2442,43 @@ WHERE books.title = $1 [["title", "Abstraction and Specification in Program Deve
...
+The [`find_or_create_by`][] method checks whether a record with the specified attributes exists. If it doesn't, then `create` is called.
 
-Suppose you want to find a customer named "Andy", and if there's none, create one. You can do so by running:
+Suppose you want to find a customer named "Andy", and if there's no customer with that name, then you want to create one. You can do this by running:
 
 ```irb
-irb> Customer.find_or_create_by(first_name: 'Andy')
+irb> Customer.find_or_create_by(first_name: "Andy")
 => #<Customer id: 5, first_name: "Andy", last_name: nil, title: nil, visits: 0, orders_count: nil, lock_version: 0, created_at: "2019-01-17 07:06:45", updated_at: "2019-01-17 07:06:45">
 ```
```
p8: Nitpick. It's common to have multiple customers named Andy.
I know this is just an example, but I think examples make more impact with typical use cases.
For example (the rest of the text would probably need to be changed as well)

```suggestion
Suppose you want to find a customer with the email "andy@example.com", and if there's no customer with that email, then you want to create one. You can do this by running:

```irb
irb> Customer.find_or_create_by(email: "andy@example.com")
=> #<Customer id: 5, email: "andy@example.com", last_name: nil, title: nil, visits: 0, orders_count: nil, lock_version: 0, created_at: "2019-01-17 07:06:45", updated_at: "2019-01-17 07:06:45">
```
```

File: guides/source/active_record_querying.md [hunk 55]
```diff
@@ -2232,45 +2442,43 @@ WHERE books.title = $1 [["title", "Abstraction and Specification in Program Deve
...
-The SQL generated by this method looks like this:
+The SQL generated by this method will look like this:
 
 ```sql
-SELECT * FROM customers WHERE (customers.first_name = 'Andy') LIMIT 1
+SELECT * FROM customers WHERE (customers.first_name = "Andy") LIMIT 1
 BEGIN
-INSERT INTO customers (created_at, first_name, locked, orders_count, updated_at) VALUES ('2011-08-30 05:22:57', 'Andy', 1, NULL, '2011-08-30 05:22:57')
+INSERT INTO customers (created_at, first_name, locked, orders_count, updated_at) VALUES ("2011-08-30 05:22:57", "Andy", 1, NULL, "2011-08-30 05:22:57")
 COMMIT
```
p8:
```suggestion
SELECT * FROM customers WHERE (customers.email = "andy@example.com") LIMIT 1
BEGIN
INSERT INTO customers (created_at, email, locked, orders_count, updated_at) VALUES ("2011-08-30 05:22:57", "andy@example.com", 1, NULL, "2011-08-30 05:22:57")
COMMIT
```

guides/source/active_record_querying.md
```diff
@@ -2340,14 +2551,73 @@ irb> nina.save
...
+
+```sql
+BEGIN
+INSERT INTO customers (created_at, first_name, locked, orders_count, updated_at) VALUES ("2011-08-30 05:22:57", "Andy", 1, NULL, "2011-08-30 05:22:57")
+ROLLBACK
+SELECT * FROM customers WHERE (customers.first_name = "Andy") LIMIT 1
+```
+
+The key difference between `create_or_find_by` and `find_or_create_by` is the order of operations and atomicity:
+- `find_or_create_by`: First tries to find, then creates if not found. This is **not atomic** and can have race conditions where duplicate records may be created.
```
p8: This needs a linebreak for markup:

Image

```suggestion

- `find_or_create_by`: First tries to find, then creates if not found. This is **not atomic** and can have race conditions where duplicate records may be created.
```

guides/source/active_record_querying.md
```diff
@@ -2232,45 +2442,43 @@ WHERE books.title = $1 [["title", "Abstraction and Specification in Program Deve
...
 
-NOTE: Note that if a query matches multiple records, `find_by` will
-fetch only the first one and ignore the others (see the `LIMIT 1`
-statement above).
+NOTE: If a query matches multiple records, `find_by` will fetch only the first
+one and ignore the others, as specified by the `LIMIT 1` statement above.
 
-Find or Build a New Object
---------------------------
+Finding or Building a New Object
```
p8: To me it seems this whole section belongs as a top section after "Retrieving Objects from the Database". There is a lot of overlap in context.

guides/source/active_record_querying.md
p8: I would expect to see the generated SQL here as well:
```suggestion
Customer.exists?(1)
# SELECT 1 AS one FROM `customers` WHERE `users`.`id` = 1 LIMIT 1
```

guides/source/active_record_querying.md
```diff
@@ -2535,6 +2814,8 @@ Customer.exists?
 
 The above returns `false` if the `customers` table is empty and `true` otherwise.
 
+### `any?` and `many?`
```
p8: Maybe split this section into two separate sections (one for `any?` and one for `many?`)?
It's easier if the examples only use one method instead of two.

Also, I think we should mention `any` and `many`  don't talk to the database if the records are already loaded:
```ruby
users = User.limit(10).load
# SELECT `users`.* FROM `users` LIMIT 10
users.any?
=> true
```

guides/source/active_record_querying.md
```diff
@@ -2564,57 +2845,58 @@ Customer.first.orders.many?
...
 
 assuming that Order has `enum status: [ :shipped, :being_packed, :cancelled ]`.
 
 ### `count`
 
 If you want to see how many records are in your model's table you could call `Customer.count` and that will return the number.
-If you want to be more specific and find all the customers with a title present in the database you can use `Customer.count(:title)`.
 
-For options, please see the parent section, [Calculations](#calculations).
+If you want to be more specific and find all the customers with a title present in the database you can use `Customer.count(:title)`.
```
p8: Some of the similar examples in this section use a code example block (although with a string instead of a symbol):
````suggestion
If you want to be more specific and find all the customers with a title present in the database you can use pass the `:title`:

```ruby
Customer.count(:title)`.
```
````

File: guides/source/active_record_querying.md [hunk 60]
```diff
@@ -2647,24 +2927,30 @@ If you want to find the sum of a field for all records in your table you can cal
...
 
 You can run [`explain`][] on a relation. EXPLAIN output varies for each database.
 
-For example, running:
+You can also chain `explain` with calculation methods like [`count`][], [`first`][], [`last`][], [`average`][], [`maximum`][], [`minimum`][], [`sum`][], and [`pluck`][] to see the query plan for those operations:
+
+```ruby
+Customer.where(active: true).explain.count
+Customer.order(:created_at).explain.first
+```
```
p8: This describes the more complex examples of using explain on the calculation methods.
I would expect to see this after the more simple example of explain with the output of the command that is coming next.

p8 (commented):
Looks great @Ridhwana !

guides/source/active_record_querying.md
Ridhwana: Agreed!

guides/source/active_record_querying.md
Ridhwana: ✅

guides/source/active_record_querying.md
```diff
@@ -84,94 +113,112 @@ class Supplier < ApplicationRecord
 end
 ```
 
-![Diagram of all of the bookstore models](images/active_record_querying/bookstore_models.png)
+NOTE: These models use `id` as the primary key, unless specified otherwise.
+
+![Diagram of all of the bookstore
+models](images/active_record_querying/bookstore_models.png)
 
 Retrieving Objects from the Database
```
p8: @Ridhwana A final thing I just noticed is that we use both Object and Record for the same thing in this guide. I think we should stick to Record for consistency.
There are a lot more occurences of Object in this guide that  should probably be replaced as well.
```suggestion
Retrieving Records from the Database
```

Ridhwana: Hey @p8, thanks for catching this!  I’ve changed most of the model-instance wording to Record, see https://github.com/rails/rails/pull/56341/commits/95a8a9be6ece357a4e4634ccbeb3b7b331152f2d, but I left a few object usages where that term is still accurate. For example: ActiveRecord::Relation and ActiveRecord::Result should still be described as objects, or generic lifecycle/association-proxy wording also still uses object where it doesn’t mean a model record.

p8: Thanks @Ridhwana ! I think that makes sense.

guides/source/active_record_querying.md
```diff
@@ -431,2241 +481,2958 @@ This is equivalent to writing:
...
+
+store(dev)> Customer.first.orders.many?
+=> true
 ```
 
-You can also order from a joined table
+[`exists?`]:
+    https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F
+
+### Retrieving Multiple Records in Batches
```
p8, line 1161: @Ridhwana I just noticed this. Shouldn't this section be moved under: "Retrieving Records from the Database"?
So we'd have:
- Retrieving a Single Record
- Retrieving Multiple Records
- Understanding Method Chaining
- Finding Records and Values
- Retrieving Multiple Records in Batches

Or:
- Retrieving a Single Record
- Retrieving Multiple Records
- Retrieving Multiple Records in Batches
- Understanding Method Chaining
- Finding Records and Values

PR #56343, opened 2025-12-10
Add delete_in_batches for batch deletions

shayonj:
### Motivation / Background

Currently, the default option for `order` is `asc` and can also be set to `desc` or passed an array with those values. This works great in most cases, however sometimes forcing the `ORDER` clause for the primary key can result into a sort + heap scan in Postgresql, by passing an index lookup and may cause a performance tax and in those situations where we don't care about order, ~it'd be lovely to continue using `in_batches` with no order instead of having to write custom SQL.~

**Update:** Based on feedback from @jeremy, @p8, and @skipkayhil, this PR now introduces `delete_in_batches` instead of adding `order: nil` support.

### Solution

Add `delete_in_batches` method:

```ruby
Event.where(story_run_id: ids).delete_in_batches
Event.where(processed: true).delete_in_batches(of: 5000)
```

For bulk deletes where order doesn't matter, something like can actually be faster because the lookups just hit the primary key index.

Added context (from the conversations below): This is cleaner than `order: nil` because the intent is explicit (as this is for deletions) and we don't need to modify the order function either. 

### Checklist

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Add `delete_in_batches` for deleting records in batches.
+
+    Unlike `in_batches.delete_all`, this does not use cursor tracking or
+    `ORDER BY` since deleted records are excluded from subsequent batches.
+
+    ```ruby
+    Event.where("created_at < ?", 1.year.ago).delete_in_batches
+    Event.where(processed: true).delete_in_batches(of: 5000)
+    ```
+
+    *Shayon Mukherjee*
+
 *   Yield the transaction object to the block when using `with_lock`.
 
     *Ngan Pham*
--- a/activerecord/lib/active_record/querying.rb
+++ b/activerecord/lib/active_record/querying.rb
@@ -11,7 +11,7 @@ module ActiveRecord
       :find_or_create_by, :find_or_create_by!, :find_or_initialize_by,
       :create_or_find_by, :create_or_find_by!,
       :destroy, :destroy_all, :delete, :delete_all, :update_all, :touch_all, :destroy_by, :delete_by,
-      :find_each, :find_in_batches, :in_batches,
+      :find_each, :find_in_batches, :in_batches, :delete_in_batches,
       :select, :reselect, :order, :regroup, :in_order_of, :reorder, :group, :limit, :offset, :joins, :left_joins, :left_outer_joins,
       :where, :rewhere, :invert_where, :preload, :extract_associated, :eager_load, :includes, :from, :lock, :readonly,
       :and, :or, :annotate, :optimizer_hints, :extending,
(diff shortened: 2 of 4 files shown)


p8:
Without an order isn't the resulting order non-deterministic?
If it's non-deterministic, how do you ensure the next batch isn't missing or duplicating rows?

shayonj:
My intention with this was to use it for deletes basically, something like

```ruby
Event
  .where(story_run_id: ids)
  .lock("FOR UPDATE SKIP LOCKED")
  .in_batches(of: 1000, order: nil, &:delete_all)
```

Given, its deletes, the ordering won't matter is what I was thinking. I can see this only being in useful in the context of deletes perhaps?

jeremy (commented):
Really clever usage with batched deletions.

I'd document it separately as a special-case advanced usage.

Can you `.reorder(nil)` to hack out the ordering? If there's an escape hatch that works, it's probably worth not muddying the method.

p8:
We also have a `find_in_batches`. If this is only useful for deletions, maybe we should make it more explicit and add a `delete_in_batches`?

shayonj:
@jeremy - I confirmed `.reorder(nil)` won't work as an escape hatch. The batching code ignores any existing order and then forces its own via `.reorder(batch_orders.to_h)`.

@p8 - I like the `delete_in_batches` idea! It makes the intent explicit and the implementation is simpler since we don't need ORDER BY or cursor tracking. Could do something like:

```ruby
def delete_in_batches(of: 1000)
  total = 0
  loop do
    deleted = limit(of).delete_all
    break if deleted == 0
    total += deleted
  end
  total
end
```

Usage could be:
```ruby
Event.where(story_run_id: ids).delete_in_batches(of: 1000)
```

Happy to update accordingly

skipkayhil:
`+1` for `delete_in_batches`, I think that makes much more sense than adding an option to the existing API which is only useful for deletes (especially if it keeps the `in_batches` code cleaner)

shayonj:
Thank you for the feedback and upvotes. I agree, `delete_in_batches` is cleaner and sort of mimics some of the internal function APIs we have too around this. Updated accordingly and open for feedback 🙏🏾

shayonj:
👋🏾 Hello team - wanted to see if this proposal was still something you were interested in? Happy to rebase accordingly.

fxn:
When working with batches, I am always concerned about the unlikely but possible race conditions from rows being updated or inserted in busy tables.

Would you consider breaking the loop when the number of deleted rows is strictly less than the batch size? Race conditions still possible, but the cut is clear and there is one less statement.

**EDIT:** Nah, forget it, the query is arbitrary and could return duplicates.

PR #56350, opened 2025-12-12
Use a modern approach for cross-site request forgery protection

rosa:
### Motivation / Background

CSRF protection has always been a source of issues related to cached pages, stale tokens, the session being somehow cleared, all resulting in false positives. We still get these in our apps. 

This PR proposes to modernise this, with the goal of eventually deprecating and getting rid of CSRF tokens completely. It takes advantage of the `Sec-Fetch-Site` header, which is a fetch metadata request header that modern browsers send, and that indicates whether a request for a resource is coming from the same origin, the same site or a different site. The server can use this to decide whether the request should be allowed or not.

In this PR, this approach is used by default, in a very conservative and strict way: only requests with a `Sec-Fetch-Site` value of `same-site` or `same-origin` are allowed. Requests with a value of `cross-site` are rejected unless they come from a configured trusted origin (see below, in the Detail section). Requests that don't include a `Sec-Fetch-Site` header or that have it set to `none` are rejected as well. Assuming a modern browser, user-initiated requests (like direct navigation or clicking a bookmark) shouldn't cause a non-GET/HEAD request, so these wouldn't be subject to the CSRF check in any case. Requests that don't include the header could be:
- Not from a browser, in which case they wouldn't include the authenticity token either. These would be excluded from the check in another way.
- From a browser, but having the header stripped by some middle proxy or firewall (we've seen this in the wild). In this case, we can't allow the request without making the user vulnerable to CSRF.

This is what we're currently using in HEY, with a custom strategy class for unverified request handling. 

For apps that can't use this approach yet because they need to support older browsers that don't send the `Sec-Fetch-Site` header, the usual authenticity token approach is kept as a fallback. When a decision can't be clearly made with the header value because it's missing or 'none', the token is checked as before.

The `Origin` check is not changed; it continues to work in the same way, combined with the `Sec-Fetch-Site `or authenticity token checks.

### Detail
This PR introduces a new option, `using`, that can be passed to `protect_from_forgery`, to indicate the approach used for verifying requests. The possible values are `header_only`, which relies exclusively on the `Sec-Fetch-Site` header, or `header_or_legacy_token`, which uses the authenticity token when a decision can't be made with the header alone. The default would be `header_only` for Rails 8.2. Example: 

```ruby
class ApplicationController < ActionController::Base
  protect_from_forgery using: :header_or_legacy_token, with: :exception
```

When using the `header_or_legacy_token` approach, a warning is logged when the fallback is necessary. In this way, developers can check which requests don't include valid `Sec-Fetch-Site` headers and progressively fix them so that they can eventually stop using the CSRF token. 

A new option, `trusted_origins`, is also added to set trusted origins for legitimate cross-site requests (OAuth callbacks, third-party embeds):
 
```ruby
protect_from_forgery trusted_origins: %w[ https://accounts.google.com ]
```

I'll include some comments in the PR to point out other important details. 

### Additional information
Some references about this: 
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site
- https://web.dev/articles/fetch-metadata
- https://github.com/golang/go/issues/73626
- https://www.alexedwards.net/blog/preventing-csrf-in-go

### Open questions
Besides the obvious question of whether this is an acceptable approach for Rails and whether this makes sense with the current implementation, a possible open question is: should the values allowed for `Sec-Fetch-Site` be configurable? This would allow for an even stricter policy where `same-site` requests aren't allowed, only `same-origin`. This would prevent requests coming from a different subdomain, for example. It would also allow for a more relaxed policy where `none` or missing values are allowed because the risk of people running older browsers or some kind of proxy is deemed acceptable. 

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -119,7 +119,7 @@ module ActionController # :nodoc:
       # * :token_fallback - Combined approach: Sec-Fetch-Site with fallback to token
       singleton_class.delegate :forgery_protection_verification_strategy, :forgery_protection_verification_strategy=, to: :config
       delegate :forgery_protection_verification_strategy, :forgery_protection_verification_strategy=, to: :config
-      self.forgery_protection_verification_strategy = :fetch_metadata
+      self.forgery_protection_verification_strategy = :token_fallback
 
       # Origins allowed for cross-site requests, such as OAuth/SSO callbacks,
       # third-party embeds, and legitimate remote form submission.
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -566,8 +566,6 @@ module ActionController # :nodoc:
       # For all strategies, GET and HEAD requests are allowed without verification.
       #
       def verified_request? # :doc:
-        return true if request.get? || request.head? || !protect_against_forgery?
-
         request.get? || request.head? || !protect_against_forgery? ||
           (valid_request_origin? && verified_request_for_forgery_protection?)
       end
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -7,11 +7,13 @@ require "action_controller/metal/exceptions"
 require "active_support/security_utils"
 
 module ActionController # :nodoc:
-  class InvalidAuthenticityToken < ActionControllerError # :nodoc:
-  end
-
   class InvalidCrossOriginRequest < ActionControllerError # :nodoc:
   end
+
+  include ActiveSupport::Deprecation::DeprecatedConstantAccessor
+  deprecate_constant "InvalidAuthenticityToken", "ActionController::InvalidCrossOriginRequest",
+    deprecator: ActionController.deprecator,
+    message: "ActionController::InvalidAuthenticityToken has been deprecated and will be removed in Rails 9.0. Use ActionController::InvalidCrossOriginRequest instead."
 
   # # Action Controller Request Forgery Protection
   #
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -115,11 +115,11 @@ module ActionController # :nodoc:
       self.csrf_token_storage_strategy = SessionStore.new
 
       # The strategy to use for verifying requests. Options are:
-      # * :fetch_metadata - Use Sec-Fetch-Site header only (default, modern browsers)
-      # * :token_fallback - Combined approach: Sec-Fetch-Site with fallback to token
+      # * :header_only - Use Sec-Fetch-Site header only (default, modern browsers)
+      # * :header_or_legacy_token - Combined approach: Sec-Fetch-Site with fallback to token
       singleton_class.delegate :forgery_protection_verification_strategy, :forgery_protection_verification_strategy=, to: :config
       delegate :forgery_protection_verification_strategy, :forgery_protection_verification_strategy=, to: :config
-      self.forgery_protection_verification_strategy = :token_fallback
+      self.forgery_protection_verification_strategy = :header_or_legacy_token
 
       # Origins allowed for cross-site requests, such as OAuth/SSO callbacks,
       # third-party embeds, and legitimate remote form submission.
@@ -231,7 +231,7 @@ module ActionController # :nodoc:
       #
       # Built-in verification strategies are:
       #
-      # *   `:fetch_metadata` - Uses the `Sec-Fetch-Site` header sent by modern
+      # *   `:header_only` - Uses the `Sec-Fetch-Site` header sent by modern
       #     browsers to verify that requests originate from the same site. This
       #     approach does not require authenticity tokens but only works with
       #     browsers that support the Fetch Metadata Request Headers. Requests
@@ -239,12 +239,12 @@ module ActionController # :nodoc:
       #     default, as Rails only supports modern browsers by default (see
       #     `allow_browser`).
       #
-      # *   `:token_fallback` - A hybrid approach that first checks the `Sec-Fetch-Site`
+      # *   `:header_or_legacy_token` - A hybrid approach that first checks the `Sec-Fetch-Site`
       #     header. If the header indicates same-site or same-origin, the request is
       #     allowed. Requests with a cross-site value are rejected. When the header is
       #     missing or "none", it falls back to checking the authenticity token.
       #     This mode logs when falling back to authenticity token to help identify
-      #     requests that should be fixed to work with `:fetch_metadata`. Use this
+      #     requests that should be fixed to work with `:header_only`. Use this
       #     if you need to support older browsers that don't send the `Sec-Fetch-Site` header.
       #
       # *   `:trusted_origins` - Array of origins to allow for cross-site requests,
@@ -255,10 +255,10 @@ module ActionController # :nodoc:
       #
       #     class ApplicationController < ActionController::Base
       #       # Modern browsers only (default)
-      #       protect_from_forgery using: :fetch_metadata, with: :exception
+      #       protect_from_forgery using: :header_only, with: :exception
       #
       #       # Hybrid approach with fallback for older browsers
-      #       protect_from_forgery using: :token_fallback, with: :exception
+      #       protect_from_forgery using: :header_or_legacy_token, with: :exception
       #
       #       # Allow cross-site requests from trusted origins
       #       protect_from_forgery trusted_origins: %w[ https://accounts.google.com ]
@@ -270,7 +270,7 @@ module ActionController # :nodoc:
         self.request_forgery_protection_token ||= :authenticity_token
 
         self.csrf_token_storage_strategy = storage_strategy(options[:store] || SessionStore.new)
-        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || :fetch_metadata)
+        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || :header_only)
         self.forgery_protection_trusted_origins = Array(options[:trusted_origins]) if options.key?(:trusted_origins)
 
         before_action :verify_request_for_forgery_protection, options
@@ -321,10 +321,10 @@ module ActionController # :nodoc:
 
         def verification_strategy(name)
           case name
-          when :fetch_metadata, :token_fallback
+          when :header_only, :header_or_legacy_token
             name
           else
-            raise ArgumentError, "Invalid request forgery verification strategy, use :fetch_metadata or :token_fallback."
+            raise ArgumentError, "Invalid request forgery verification strategy, use :header_only or :header_or_legacy_token."
           end
         end
     end
@@ -490,7 +490,7 @@ module ActionController # :nodoc:
       def cross_origin_request?
         !valid_request_origin? ||
           sec_fetch_site_value == "cross-site" ||
-          using_only_fetch_metadata_for_forgery_protection?
+          using_header_only_for_forgery_protection?
       end
 
       def unverified_request_warning_message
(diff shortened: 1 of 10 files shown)


actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -40,7 +46,7 @@ class InvalidCrossOriginRequest < ActionControllerError # :nodoc:
   #
   # Subclasses of ActionController::Base are protected by default with the
   # `:exception` strategy, which raises an
-  # ActionController::InvalidAuthenticityToken error on unverified requests.
+  # ActionController::InvalidCrossOriginRequest error on unverified requests.
```
rosa, line 51: This would be a breaking change, as I understand this exception is publicly documented and people are even encouraged to rescue it to do specific tasks upon this error. 

The problem is that the exception is not really descriptive if not using a token to defend against CSRF. It was even used in place of `InvalidCrossOriginRequest` when `Origin` is set to `null`. I think it'd be great to change this and start using an exception with a more correct name. I've just changed it here to get your take about how this could be done: 
- Would this breaking change be ok for a version change (i.e. Rails 8.2), if documented properly? 
- Would it be better to have another exception that extends `InvalidAuthenticityToken` for backwards compatibility, that's later deprecated and replaced with a new one? For example, `UnverifiedRequestError`, or `UnverifiedCrossSiteRequestError`. 
- Should we keep the incorrectly named exception for everything?

jeremy: Can deprecate the constant! https://github.com/rosa/rails/pull/2

rosa: Ohhh TIL! Brilliant, thanks a lot @jeremy!

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -460,16 +550,63 @@ def non_xhr_javascript_response? # :doc:
...
+      # *   `:token_fallback` - Uses Sec-Fetch-Site header with fallback to token
+      #
+      # For all strategies, GET and HEAD requests are allowed without verification.
       #
       def verified_request? # :doc:
+        return true if request.get? || request.head? || !protect_against_forgery?
+
         request.get? || request.head? || !protect_against_forgery? ||
-          (valid_request_origin? && any_authenticity_token_valid?)
+          (valid_request_origin? && verified_request_for_forgery_protection?)
```
jeremy, line 569:
```suggestion
        valid_request_origin? && verified_request_for_forgery_protection?
```

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -106,6 +112,20 @@ module RequestForgeryProtection
       delegate :csrf_token_storage_strategy, :csrf_token_storage_strategy=, to: :config
       self.csrf_token_storage_strategy = SessionStore.new
 
+      # The strategy to use for verifying requests. Options are:
+      # * :fetch_metadata - Use Sec-Fetch-Site header only (default, modern browsers)
+      # * :token_fallback - Combined approach: Sec-Fetch-Site with fallback to token
```
jeremy: Suggest clarifying naming: https://github.com/rosa/rails/pull/4

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -460,16 +552,63 @@ def non_xhr_javascript_response? # :doc:
...
+      end
+
+      def verified_with_token_fallback?
+        case sec_fetch_site_value
+        when "same-origin", "same-site"
+          true
+        when "cross-site"
+          origin_trusted?
+        else # "none" or missing
+          logger.warn "Falling back to CSRF token check for forgery protection" if logger && log_warning_on_csrf_failure
```
bdewater-thatch, line 596: IDK if this should be Rubocop rule or documented somewhere, but I think Rails framework components shouldn't log directly and use Active Support Notifications (or the new Event Reporter) instead so that applications can have control over the log output and hook up their own additional behaviour in a clean way.

rosa: Good point! I had used Active Support notifications on my first approach, but changed to this to match the other logging done in this module. https://github.com/rails/rails/blob/9236403fbac77d18490dd7e8bb25eb60e92191f9/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L402


Maybe everything should be switched instead of mixing the two ways?

actionpack/lib/action_controller/metal/request_forgery_protection.rb
rosa: D'oh, I left over the guard from a previous implementation, will remove the guard instead! Thanks for spotting this 🙏🏻

actionpack/lib/action_controller/metal/request_forgery_protection.rb
jeremy: Agree re. sticking with existing usage. We can introduce structured events in a separate PR.

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -203,25 +225,66 @@ module ClassMethods
...
+      #     default, as Rails only supports modern browsers by default (see
+      #     `allow_browser`).
+      #
+      # *   `:header_or_legacy_token` - A hybrid approach that first checks the `Sec-Fetch-Site`
+      #     header. If the header indicates same-site or same-origin, the request is
+      #     allowed. Requests with a cross-site value are rejected. When the header is
+      #     missing or "none", it falls back to checking the authenticity token.
+      #     This mode logs when falling back to authenticity token to help identify
+      #     requests that should be fixed to work with `:header_only`. Use this
+      #     if you need to support older browsers that don't send the `Sec-Fetch-Site` header.
```
jeremy: Indentation fixed in https://github.com/rosa/rails/pull/6

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -203,25 +225,65 @@ module ClassMethods
...
+      #       protect_from_forgery trusted_origins: %w[ https://accounts.google.com ]
+      #     end
       def protect_from_forgery(options = {})
         options = options.reverse_merge(prepend: false)
 
         self.forgery_protection_strategy = protection_method_class(options[:with] || :null_session)
         self.request_forgery_protection_token ||= :authenticity_token
 
         self.csrf_token_storage_strategy = storage_strategy(options[:store] || SessionStore.new)
+        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || :header_only)
```
chaadow, line 272: I believe this should be 
```suggestion
        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || forgery_protection_verification_strategy)
```

For those who still want to keep the old behaviour

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -203,25 +225,65 @@ module ClassMethods
...
       # Turn off request forgery protection. This is a wrapper for:
       #
-      #     skip_before_action :verify_authenticity_token
+      #     skip_before_action :verify_request_for_forgery_protection
       #
       # See `skip_before_action` for allowed options.
       def skip_forgery_protection(options = {})
-        skip_before_action :verify_authenticity_token, options.reverse_merge(raise: false)
+        skip_before_action :verify_request_for_forgery_protection, options.reverse_merge(raise: false)
       end
```
chaadow, line 287: There was  kind of a breaking change ( renaming `verify_authenticity_token` to `verify_request_for_forgery_protection`)  , I've just realized it in my application. I know i can just use `skip_forgery_protection` to fix it. But `verify_authenticity_token` is used in the documentation, and I believe in many applications ( as well as many Rails articles ) use it especially in the `Rails API only` sphere.

I had this error by running: `bin/rails zeitwerk:check` ( aka eager loading the application )

Image

@jeremy @rosa I couldn't figure out an easy way to handle both foreward and backward compatibility, so I'm opening a documention PR

chaadow: https://github.com/rails/rails/pull/56358

rosa: Ohh, interesting 🤔 I thought this wouldn't be used directly because it's discouraged, these methods are used, and Rails API only disables forgery protection, so in theory you shouldn't need to handle these methods. 

Maybe we should deprecate it instead.

p8:
Linking this with a recent discussion in the Rack repo: https://github.com/rack/rack/issues/2367

actionpack/lib/action_controller/metal/request_forgery_protection.rb
chaadow: in my case, i have a hybrid application, where i have only a few "API" routes that have their own "Base api controller" and the rest is "standard rails with CSRF enable"

I've just done a quick search on Github and `skip_before_action :verify_authenticity_token` [is used quite a lot ](https://github.com/search?q=skip_before_action+%3Averify_authenticity_token&type=code&p=1)

chaadow: > Maybe we should deprecate it instead.

@rosa I've considered this option, but I couldn't find a simple way to handle both backward and foreward compatibilty.
because people use it in this way : `skip_before_action :verify_authenticity_token`

That presumes that a `before_action :verify_authenticity_token` was set somewhere  ( by this module)
I'm lost :D 
That's why i opted into updating the upgrade guide after thoughtful consideration.
However if you can find a way of deprecating it then please do!

chaadow: I think a `rubocop-rails` cop detecting `skip_before_action :verify_authenticity_token` and changing it to `skip_forgery_protection` can be cool as well ( Just throwing some ideas that I've just thought of)

chaadow: here is an example of the `Gitlab` codebase : https://github.com/gitlabhq/gitlabhq/blob/5395a822abdb4af1415a875a4d70c19c3a3d4118/app/controllers/jwt_controller.rb#L6

rosa: I think we should revert the rename, then, and add a deprecation notice 🤔 Then rename in the next version. I'll open a PR with this proposal.

rosa:
Oh! Thanks a lot @p8! I hadn't seen that discussion!

actionpack/lib/action_controller/metal/request_forgery_protection.rb
chaadow: I'll be interested to see how you can add a deprecation notice in this case.
 I might be missing something, but since `protect_from_forgery` is the one calling `before_action :verify_authenticity_token`, and 99,99% of applications call `protect_from_forgery` in their controllers in some form, that means all applications will have this deprecation warning :/

rosa: Yes, it's indeed tricky... I came up with https://github.com/rails/rails/pull/56369, that I think could work, but only for `skip_before_action :verify_authenticity_token`, not for other usages of this method. I think that should be good enough for this, though, looking into how people are using this.

bensheldon:
I just ran across this in my GoodJob test suite, specifically for system tests that are driven by `rack_test` and thus don't have the appropriate simulated request. (or that's what I think that's going on). I'm not sure the right way to handle this for my test suite: either downgrading the security by introspcting test configuration (or vice versa) or to suggest a change to rack_test. 

**Update:** I added this to my test setup (rspec) to add the header to system test requests driven by rack_test:

```ruby
RSpec.configure do |config|
  config.prepend_before(:each, type: :system) do |example|
    #...
    driven_by :rack_test
    page.driver.header "sec-fetch-site", "same-site" # <-----
  end
end

Atulin:
Correct me if I'm wrong, but can't the `Sec-Fetch-*` headers be trivially spoofed by a non-browser source?

bensheldon:
@Atulin yes, for non-browser sources. And not more trivially imo than the cookie+hidden-field strategy. CSRF is a browser-context vulnerability. 

I personally like Sec-Fetch-Site because it is purely a CSRF protection that is simple.

The Cookie+hidden-field strategy does have some form-stuffing protection from non-browser sources, cause it requires a GET before a POST. But form-stuffing protection can be made better/simpler by not entangling it with CSRF protection.

PR #41684, opened 2021-03-16
Disable parallel testing when running individual files

ricardotk002:
### Summary

Fixes #41473

Setting up the parallel workers could be an overhead when running
individual files.

This patch disables that process in case the number of files to run
is less than one.

Results running a sample file:

Before:

```
actionpack $ bin/test test/controller/parameters/accessors_test.rb
Run options: --seed 48261

........................................................................

Finished in 0.211923s, 339.7460 runs/s, 552.0873 assertions/s.
72 runs, 117 assertions, 0 failures, 0 errors, 0 skips
```

After

```
actionpack $ bin/test test/controller/parameters/accessors_test.rb

Run options: --seed 5461

........................................................................

Finished in 0.008411s, 8560.2189 runs/s, 13910.3557 assertions/s.
72 runs, 117 assertions, 0 failures, 0 errors, 0 skips
```

Of course this improvement doesn't apply to all cases, so the environment
variable `PARALLEL_WORKERS` can still be used to enforce parallelization.

PR #56352, opened 2025-12-12
Fix PARALLEL_WORKERS ignoring test parallelization threshold

justinko:
Previously, setting PARALLEL_WORKERS would enable test parallelization regardless of whether the test count exceeded the threshold. Now parallelization is only enabled when tests_count > threshold.

I was fiddling with the threshold argument and noticed parallelization was always activating, causing a 4 second slowdown for running a single test. I have PARALLEL_WORKERS set in my `.env.test.local` and that was the cause. 

I can't think of a reason why we would ever want PARALLEL_WORKERS to override the behavior of threshold.

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Fix `PARALLEL_WORKERS` environment variable bypassing the parallelization threshold.
+
+    Previously, setting `PARALLEL_WORKERS` would enable parallelization regardless of
+    the test count threshold. Now parallelization only occurs when the number of tests
+    exceeds the configured threshold.
+
+    *Justin Ko*
+
 *   Make flaky parallel tests easier to diagnose by deterministically assigning
     tests to workers.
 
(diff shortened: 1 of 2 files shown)


eileencodes:
Hi @justinko - the reason I implemented it this way is `PARALLEL_WORKERS` env var is supposed to win so that it overrides threshold. Sometimes you want to test parallelism, regardless of the number of tests (like when I'm testing changes to parallel testing) without having to also update the threshold. This change would make it impossible to always run in parallel, regardless of what else is going on.

I appreciate your opening a PR but I don't want to change this behavior.

justinko:
1.) This behavior isn't documented anywhere. Imagine the amount of collective time that has been wasted the past couple of years. Setting PARALLEL_WORKERS is very common on non-virtualized environments, I'm assuming, because utilizing all cores usually results in a slower runtime. 
2.) I understand the difficultly with testing. Perhaps we should consider adding an env var for threshold. 

Which one is more appealing to you?

eileencodes:
I think we should document the behavior, but I don't think it should change since it was explicitly designed this way. 

It's not documented but it was mentioned in the changelog https://github.com/rails/rails/pull/41684/files#diff-2bf9f348945534785fefa2d491085e108aa71faa52f2e3dd140eade408b06bf3R3 and there's a test for the behavior (see the failing build).

PR #56353, opened 2025-12-12
Make `delegate` and `delegate_missing_to` work in BasicObject subclasses

rafaelfranca:
When those features are used in `BasicObject` subclasses, the generated code must use `::Kernel.raise` to raise exceptions, because `BasicObject` does not have a built-in raise method. Similarly, references to exception classes like `NoMethodError` must be fully qualified with the `::` prefix to ensure they are resolved correctly.

Without this, `delegate_to_missing` would raise a `SystemStackError` when attempting to raise a `DelegationError` from a `BasicObject` subclass, due to infinite recursion in `method_missing`.


byroot (approved):

rafaelfranca:
Thanks for the review @byroot!

PR #56356, opened 2025-12-12
Fix protect_from_forgery to use the default value from config

chaadow:
Related to https://github.com/rails/rails/pull/56350/
and this [discussion](https://github.com/rails/rails/pull/56350/changes#r2615865992)

When using
```ruby
config.action_controller.forgery_protection_verification_strategy = :header_or_legacy_token
```

We expect that this will be used in controllers by default. But in actuality, `header_only` is used as a fallback

cc @rosa @jeremy

### DIFF
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -269,7 +269,7 @@ module ActionController # :nodoc:
         self.request_forgery_protection_token ||= :authenticity_token
 
         self.csrf_token_storage_strategy = storage_strategy(options[:store] || SessionStore.new)
-        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || :header_only)
+        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || forgery_protection_verification_strategy)
         self.forgery_protection_trusted_origins = Array(options[:trusted_origins]) if options.key?(:trusted_origins)
 
         before_action :verify_request_for_forgery_protection, options
--- a/actionpack/test/controller/request_forgery_protection_test.rb
+++ b/actionpack/test/controller/request_forgery_protection_test.rb
@@ -76,16 +76,16 @@ end
 # sample controllers
 class RequestForgeryProtectionControllerUsingResetSession < ActionController::Base
   include RequestForgeryProtectionActions
-  protect_from_forgery only: %w(index meta same_origin_js negotiate_same_origin), with: :reset_session, using: :header_or_legacy_token
+  protect_from_forgery only: %w(index meta same_origin_js negotiate_same_origin), with: :reset_session
 end
 
 class RequestForgeryProtectionControllerUsingException < ActionController::Base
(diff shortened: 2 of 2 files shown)


actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -269,7 +269,7 @@ def protect_from_forgery(options = {})
         self.request_forgery_protection_token ||= :authenticity_token
 
         self.csrf_token_storage_strategy = storage_strategy(options[:store] || SessionStore.new)
-        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || :header_only)
+        self.forgery_protection_verification_strategy = verification_strategy(options[:using] || forgery_protection_verification_strategy)
```
chaadow, line 272: `forgery_protection_verification_strategy` will get delegated to `config` which will use the value set in config as default, instead of hard coding `:header_only`

jeremy (changes requested):
Nice catch! Thanks. Would you add test coverage? Something like:
```ruby
  def test_protect_from_forgery_respects_configured_verification_strategy
    old_strategy = ActionController::Base.forgery_protection_verification_strategy
    ActionController::Base.forgery_protection_verification_strategy = :header_or_legacy_token

    # Call protect_from_forgery without :using option
    # Assert it uses :header_or_legacy_token, not :header_only
  ensure
    ActionController::Base.forgery_protection_verification_strategy = old_strategy
  end
```

actionpack/test/controller/request_forgery_protection_test.rb
```diff
@@ -118,7 +118,7 @@ def handle_unverified_request
     end
   end
 
-  protect_from_forgery only: %w(index meta same_origin_js negotiate_same_origin), with: CustomStrategy, using: :header_or_legacy_token
```
chaadow, line 121: I removed all these `using :header_or_legacy_token` to also test that when not provided, it is used by default as per this line: https://github.com/rails/rails/blob/42f327ecd10623ca492db415c3a304786aa3ea19/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L122

actionpack/test/controller/request_forgery_protection_test.rb
```diff
@@ -1694,6 +1693,24 @@ def forgery_protection_origin_check
     end
 end
 
+class ConfiguredVerificationStrategyTest < ActiveSupport::TestCase
+  test "protect_from_forgery respects configured verification strategy when :using is not provided" do
+    old_strategy = ActionController::Base.forgery_protection_verification_strategy
+    assert_equal :header_or_legacy_token, old_strategy
```
chaadow, line 1699: I've added this assert because of https://github.com/rails/rails/blob/42f327ecd10623ca492db415c3a304786aa3ea19/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L122

to make it clear that the default value is `:header_or_legacy_token`

jeremy (approved):

PR #56213, opened 2025-11-22
Respect config.log_level to emit debug events

claudiob:
This is a follow-up to #55657, a PR that enables debug events in development only.

The problem is that some users might want to see debug events also in test or in production. 
A real-world example is detailed in https://github.com/rails/rails/pull/55900#issuecomment-3445390380

---

This PR slightly changes the behavior replacing `Rails.env.development?` with `config.log_level == "debug"`. 
As explained in [Rails documentation](https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels):

> The default Rails log level is :debug. However, it is set to :info for the production environment
> in the default generated config/environments/production.rb.

As a result, this PR maintains the existing behavior when defaults are applied, but opens up the possibility to see debug events in production for those who want to do so: they simply have to set `RAILS_LOG_LEVEL=debug`.

PR #56357, opened 2025-12-13
Only emit logs if logger is defined

claudiob:
See https://github.com/rails/rails/pull/56213#issuecomment-3642535238

Fixes the following error on main:

1. First generate a new Rails app:

```sh
rails/railties/exe/rails new issue56230 --dev
cd issue56230/
RAILS_ENV=production RAILS_LOG_LEVEL=debug bin/rails server
```

Then edit config/environments/production.rb replacing these lines:

```diff
-  config.consider_all_requests_local = false
+  config.consider_all_requests_local = true

+  config.action_view.logger = nil
```

Then open http://localhost:3000/up in the browser: the error is there

```
NoMethodError (undefined method 'debug?' for nil)
Caused by: NoMethodError (undefined method 'debug?' for nil)
Information for: NoMethodError (undefined method 'debug?' for nil):
/rails/activesupport/lib/active_support/event_reporter/log_subscriber.rb:9:in 'block in '
/rails/activesupport/lib/active_support/event_reporter/log_subscriber.rb:47:in 'ActiveSupport::EventReporter::LogSubscriber#emit'
```

### DIFF
--- a/activesupport/lib/active_support/event_reporter/log_subscriber.rb
+++ b/activesupport/lib/active_support/event_reporter/log_subscriber.rb
@@ -42,6 +42,7 @@ module ActiveSupport
       class_attribute :log_levels, default: {} # :nodoc:
 
       def emit(event)
+        return unless logger
         name = event[:name]
         event_method = name[name.index(".") + 1, name.length]
         public_send(event_method, event) if LEVEL_CHECKS[log_levels[event_method]]&.call(logger)
--- a/activesupport/test/event_reporter/log_subscriber_test.rb
+++ b/activesupport/test/event_reporter/log_subscriber_test.rb
@@ -73,6 +73,18 @@ class ActiveSupport::EventReporter::LogSubscriberTest < ActiveSupport::TestCase
     assert_instance_of(ActiveSupport::LogSubscriber::TestHelper::MockLogger, subclass.logger)
   end
 
+  test "nil logger" do
+    MyLogSubscriber.logger = nil
+
+    subclass = Class.new(MyLogSubscriber) do
+      def self.default_logger
+        nil
+      end
+    end
+
+    assert_nothing_raised { subclass.new.emit(name: "test.debug_only") }
+  end
+
   test ".subscription_filter" do
     event_reporter_raise_on_error do
       ActiveSupport.event_reporter.notify("other_namespace_that_shouldnt_work.info_only")


adrianna-chang-shopify (approved):

gmcgibbon:
LGTM but let's write a test and add a changelog please.

adrianna-chang-shopify:
Does this need a changelog entry? I figured this would be considered a bug fix.

gmcgibbon:
> Does this need a changelog entry?

~Yes, the log subscriber is released.~

I don't see it in any release tag though I swear it was. I guess I'm wrong. Just the test is needed then.

claudiob:
Thanks for the review @gmcgibbon -- I'll get to it.

Is your suggestion to add a test to https://github.com/rails/rails/blob/main/activesupport/test/event_reporter/log_subscriber_test.rb for the scenario where `logger` is not defined?

gmcgibbon:
Yeah, override the default logger in a subclass and have the method return nil to prevent regression.

claudiob:
@gmcgibbon I wrote a test that fails on `main` and passes with this patch. Let me know if that works.

### on main

```
~/code/rails/activesupport$ bin/test test/event_reporter/log_subscriber_test.rb:85

E

Error:
ActiveSupport::EventReporter::LogSubscriberTest#test_nil_logger:
NoMethodError: undefined method 'debug?' for nil
    lib/active_support/event_reporter/log_subscriber.rb:9:in 'block in <class:LogSubscriber>'
    lib/active_support/event_reporter/log_subscriber.rb:47:in 'ActiveSupport::EventReporter::LogSubscriber#emit'
    test/event_reporter/log_subscriber_test.rb:85:in 'block (2 levels) in <class:LogSubscriberTest>'
    lib/active_support/testing/assertions.rb:49:in 'ActiveSupport::Testing::Assertions#assert_nothing_raised'
    test/event_reporter/log_subscriber_test.rb:85:in 'block in <class:LogSubscriberTest>'


bin/test test/event_reporter/log_subscriber_test.rb:76
```

### after this patch

```
~/code/rails/activesupport$ bin/test test/event_reporter/log_subscriber_test.rb:85

.
```

gmcgibbon (approved):

gmcgibbon:
Thanks!

PR #56358, opened 2025-12-13
[Guides] Add an upgrade note about the request forgery callback rename

chaadow:
Related to https://github.com/rails/rails/pull/56350/ ( [Comment](https://github.com/rails/rails/pull/56350/changes#r2616298971) )

There was  kind of a breaking change ( renaming `verify_authenticity_token` to `verify_request_for_forgery_protection`)  , I've just realized it in my application. I know developers can just use `skip_forgery_protection` to fix it. But `verify_authenticity_token` is referenced quited a bit in the documentation, and I believe in many applications ( as well as many Rails articles ) use it frequently in the `Rails API only` sphere.

I had this error by running: `bin/rails zeitwerk:check` ( aka eager loading the application )

Image

I couldn't figure out an easy way to handle both foreward and backward compatibility, so I'm opening a documentation PR with an upgrade note to help developers upgrade  their applications from Rails 8.1 to 8.2

references across Github can be found here: https://github.com/search?q=skip_before_action+%3Averify_authenticity_token&type=code&p=1

cc @jeremy @rosa

### DIFF
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -286,6 +286,37 @@ module ActionController # :nodoc:
         skip_before_action :verify_request_for_forgery_protection, options.reverse_merge(raise: false)
       end
 
+      # `verify_request_for_forgery_protection` is the callback added by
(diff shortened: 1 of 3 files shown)


chaadow:
@rosa closing in favor of #56369

PR #56367, opened 2025-12-15
Use Tempfile not to overwrite existing file

nobu:
### Motivation / Background

This Pull Request has been created because the current implementation of `File.probe_stat_in` **can** clobber an existing file.

### Detail

This Pull Request changes to use `Tempfile.create` not to overwrite an existing file.
First drop the file created by `Tempfile`, which is always permission 0600, then `stat` the re-opened file with the default permission.

### Additional information

`File::SHARE_DELETE | File::BINARY` is needed to unlink on Windows, c.f. `Tempfile.create_anonymous`.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/file/atomic.rb
+++ b/activesupport/lib/active_support/core_ext/file/atomic.rb
@@ -54,19 +54,14 @@ class File
 
   # Private utility method.
   def self.probe_stat_in(dir) # :nodoc:
-    basename = [
-      ".permissions_check",
-      Thread.current.object_id,
-      Process.pid,
-      rand(1000000)
-    ].join(".")
+    mode = File::SHARE_DELETE | File::BINARY # to unlink on Windows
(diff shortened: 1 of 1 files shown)


byroot:
> This Pull Request has been created because the current implementation of `File.probe_stat_in` **can** clobber an existing file.

It seems extremely unlikely given the name generation, but OK.

I just think Tempfile is making things more complicated than adding and handling `O_EXCL` to the existing code.

byroot:
I think https://github.com/rails/rails/pull/56368 is simpler.

jeremy:
#56368 is a smaller change; however, the Tempfile code follows the principle of relying on Ruby where we can, and it's just less code.

byroot:
@jeremy well, it's relying on ruby but also fighting it, e.g. the default permissions you can't set, so you have to delete the file and recreate it.

I strongly believe https://github.com/rails/rails/pull/56368 is much easier to understand, and cleaner.

PR #56369, opened 2025-12-15
Support but deprecate `skip_before_action :verify_authenticity_token`

rosa:
In #56350, `verify_authenticity_token` was renamed to `verify_request_for_forgery_protection`, to more accurately reflect its purpose, now that  an authenticity token is not necessarily verified.

However, because many people rely on `skip_before_action :verify_authenticity_token`, to opt out of forgery protection, we need to keep this working and deprecate it.  We simply mark it as having ran, as part of `protect_from_forgery`, and when verifying the request, we check if the method ran. If it didn't, it's because it was skipped on its own and not via `skip_forgery_protection`, so we can emit the deprecation warning.

cc @chaadow

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -272,7 +272,7 @@ module ActionController # :nodoc:
         self.forgery_protection_verification_strategy = verification_strategy(options[:using] || forgery_protection_verification_strategy)
         self.forgery_protection_trusted_origins = Array(options[:trusted_origins]) if options.key?(:trusted_origins)
 
-        before_action :verify_request_for_forgery_protection, options
+        before_action :verify_authenticity_token, :verify_request_for_forgery_protection, options
         append_after_action :verify_same_origin_request
         append_after_action :append_sec_fetch_site_to_vary_header, options
       end
@@ -444,6 +444,7 @@ module ActionController # :nodoc:
     def initialize(...)
       super
       @_marked_for_same_origin_verification = nil
+      @_verify_authenticity_token_ran = false
     end
 
     def reset_csrf_token(request) # :doc:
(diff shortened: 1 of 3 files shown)


chaadow:
Oh I've just thought of this: https://github.com/rails/rails/pull/56358/changes/871259f98fefb66cbe76f6d0217e61d0a52e52e9

rosa:
That was another alternative I thought! In the end, I went with this one, but I don't feel strongly, TBH.

actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -272,7 +272,7 @@ def protect_from_forgery(options = {})
         self.forgery_protection_verification_strategy = verification_strategy(options[:using] || forgery_protection_verification_strategy)
         self.forgery_protection_trusted_origins = Array(options[:trusted_origins]) if options.key?(:trusted_origins)
 
-        before_action :verify_request_for_forgery_protection, options
+        before_action :verify_authenticity_token, :verify_request_for_forgery_protection, options
```
chaadow, line 275: nice that's clever!

chaadow (commented):
cool!

actionpack/lib/action_controller/metal/request_forgery_protection.rb
george-ma: Heya @rosa! I'm bumping the company's monolith this week to use this commit and I found a potential edge case - with `before_action :verify_authenticity_token, :verify_request_for_forgery_protection`, when we call

[protect_from_forgery ... prepend: true](https://github.com/iconben/gitlabhq/blob/1f9093a2dd1cf8a096cdcdd786ec9f95f3cce2a5/app/controllers/metrics_controller.rb#L6), this results in

[]
[:verify_authenticity_token]  <- first prepend
[:verify_request_for_forgery_protection, :verify_authenticity_token] <- second prepend

causing the deprecation warning to fire when `:verify_request_for_forgery_protection` runs as `:verify_authenticity_token` has not ran yet

still investigating some things that prevent our monolith from working around this so just wanted to leave a note before I sign off for the day as I _think_ this is unintended, thank you!

first-thought to a fix:
```
if options[:prepend]
  prepend_before_action :verify_authenticity_token, options
  before_action :verify_request_for_forgery_protection, options.except(:prepend)
else
   before_action :verify_authenticity_token, :verify_request_for_forgery_protection, options
end
```

rosa: Oh! Indeed, I didn't consider this case! I think your fix could work.

rosa: @george-ma, would you like to submit a PR with the fix? Or should I do it? I don't mind either way.

george-ma: Writing it up rn! will cc you shortly~

george-ma: https://github.com/rails/rails/pull/56392

rosa: Thank you! 🙏🏻

PR #56370, opened 2025-12-15
Enable Minitest/AssertEqual cop and standardize test assertions

shivamsinghchahar:
### Motivation / Background

This PR enables the `Minitest/AssertEqual` RuboCop cop to enforce consistent assertion style across the test suite. Using `assert_equal` instead of `assert x == y` provides better failure messages showing both expected and actual values, while `assert x == y` only shows `false`.

### Detail

**Changes made:**
1. **Enabled `Minitest/AssertEqual` cop** in `.rubocop.yml`
2. **Updated test assertions** across multiple modules to use `assert_equal(expected, actual)` instead of `assert expected == actual`
3. **Excluded 3 test files** that use custom equality operators:
   - `actionpack/test/dispatch/mime_type_test.rb` - `Mime::Type` has custom `==` for string/symbol comparison
   - `activemodel/test/cases/secure_password_test.rb` - `BCrypt::Password` uses custom `==` for digest comparison
   - `railties/test/generators/scaffold_generator_test.rb` - `BCrypt::Password` uses custom `==` for password verification

**Modules affected:**
- ActiveModel (2 files)
- ActiveRecord (3 files)
- ActiveStorage (2 files)
- ActiveSupport (3 files)
- ActionPack (2 files)
- ActionMailer (1 file)
- Railties (1 file)

**Benefits:**
- More descriptive test failure messages
- Consistent assertion style across the codebase
- Prevents future `assert ==` patterns from being introduced

### Additional information

The excluded files require `assert_operator` or `assert ==` because their custom equality operators only work when the custom object is on the left side of the comparison. Using `assert_equal` would reverse the comparison and break these tests.

### Checklist


yahonda:
Given that this pull request addresses this manually, it might be worth enabling the `Minitest/AssertEqual` cop to prevent similar cases in the future.
For example, this cop can detect and auto-correct the changes in this pull request:

```ruby
$ bundle exec rubocop --only Minitest/AssertEqual activestorage/test/service/s3_service_test.rb -a
Inspecting 1 file
C

Offenses:

activestorage/test/service/s3_service_test.rb:173:9: C: [Corrected] Minitest/AssertEqual: Prefer using assert_equal(data, service.download(key)).
        assert data == service.download(key)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
activestorage/test/service/s3_service_test.rb:187:9: C: [Corrected] Minitest/AssertEqual: Prefer using assert_equal(data, service.download(key)).
        assert data == service.download(key)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 2 offenses detected, 2 offenses corrected
yahonda@myubuntu:~/src/github.com/rails/rails$ git diff
diff --git a/activestorage/test/service/s3_service_test.rb b/activestorage/test/service/s3_service_test.rb
index 11ca847b52..fd28a7bc55 100644
--- a/activestorage/test/service/s3_service_test.rb
+++ b/activestorage/test/service/s3_service_test.rb
@@ -170,7 +170,7 @@ class ActiveStorage::Service::S3ServiceTest < ActiveSupport::TestCase
         data = SecureRandom.bytes(8.megabytes)
 
         service.upload key, StringIO.new(data), checksum: OpenSSL::Digest::MD5.base64digest(data)
-        assert data == service.download(key)
+        assert_equal data, service.download(key)
       ensure
         service.delete key
       end
@@ -184,7 +184,7 @@ class ActiveStorage::Service::S3ServiceTest < ActiveSupport::TestCase
         data = SecureRandom.bytes(3.megabytes)
 
         service.upload key, StringIO.new(data), checksum: OpenSSL::Digest::MD5.base64digest(data)
-        assert data == service.download(key)
+        assert_equal data, service.download(key)
       ensure
         service.delete key
       end
$
```

shivamsinghchahar:
@yahonda Thanks a lot for reviewing. I've enabled the Minitest/AssertEqual cop as suggested.

.rubocop.yml
```diff
@@ -403,6 +403,9 @@ Minitest/SkipEnsure:
 Minitest/UnreachableAssertion:
   Enabled: true
 
+Minitest/AssertEqual:
```
yahonda: Would you move the Minitest/AssertEqual cop above Minitest/AssertNil to keep cops in alphabetical order?

yahonda:
Would you run `bundle exec rubocop -a` to address these offenses? https://buildkite.com/rails/rails/builds/124727#019b2a98-b8a7-4a67-8ac9-3f4510a2e954

yahonda (changes requested):

shivamsinghchahar:
@yahonda Happy new year! Apologies for not getting back on this earlier.

After enforcing the Minitest/AssertEqual cop across all test files, I've identified two categories of failures where the cop's auto-correction produces incorrect results:

## 1. Custom Equality Operators (Mime::Type)
**Issue**: `Mime::Type` implements a custom `==` operator that allows comparison with both strings and symbols:
```rb
# actionpack/lib/action_dispatch/http/mime_type.rb:297-302
def ==(mime_type)
 return false unless mime_type

 (@synonyms + [ self ]).any? do |synonym|
   synonym.to_s == mime_type.to_s || synonym.to_sym == mime_type.to_sym
 end
end
```
Test failure in `actionpack/test/dispatch/mime_type_test.rb`:
```rb
# ❌ Fails - assert_equal compares in wrong direction
assert_equal Mime[:html], "application/xhtml+xml"  # Expected: Mime::Type, Actual: String
assert_equal Mime[:html], :html                     # Expected: Mime::Type, Actual: Symbol

# ✅ Works - uses custom == operator correctly
assert_operator Mime[:html], :==, "application/xhtml+xml"
assert_operator Mime[:html], :==, :html
```
The issue is that `assert_equal(expected, actual)` internally does expected `==` actual, but `String#==` and `Symbol#==` don't know how to compare with `Mime::Type` objects. The custom `Mime::Type#==` only works when the Mime object is on the left side.

## 2. BCrypt Password Verification
**Issue**: `BCrypt::Password` has a custom `==` operator that cryptographically verifies passwords:
```rb
# From bcrypt gem
class BCrypt::Password
 def ==(secret)
  # Performs secure password comparison
 end
end
```
Test failure in `railties/test/generators/scaffold_generator_test.rb`:
```rb
digest = content.match(/password_digest: ([^#\s]+)/)[1].strip
# ❌ Fails - compares BCrypt hash string with plain string
assert_equal BCrypt::Password.new(digest), "secret"
# Expected: BCrypt::Password object
# Actual: "secret" (plain string)

# ✅ Works - uses BCrypt's custom == to verify password
assert BCrypt::Password.new(digest) == "secret"
```
### Proposed Solution
I see three options:
1. Add rubocop:disable comments for these specific cases (preserves the cop's value elsewhere)
2. Exclude these test files from the Minitest/AssertEqual cop
3. Don't enable the cop due to these edge cases

What do you think? I'm happy to update the PR with whichever approach you prefer.

yahonda:
Thanks for the detailed investigation. Let's exclude these files from the `Minitest/AssertEqual` cop.

> Exclude these test files from the Minitest/AssertEqual cop

.rubocop.yml
```diff
@@ -382,6 +382,13 @@ Performance/RedundantStringChars:
 Performance/StringInclude:
   Enabled: true
 
+Minitest/AssertEqual:
+  Enabled: true
+  Exclude:
```
yahonda, line 387: Would you add some comments why these files are excluded?

activerecord/test/cases/migration_test.rb
```diff
@@ -1050,7 +1050,7 @@ def test_generate_migrator_advisory_lock_id
       salt = ActiveRecord::Migrator::MIGRATOR_SALT
       expected_id = Zlib.crc32(current_database) * salt
 
-      assert lock_id == expected_id, "expected lock id generated by the migrator to be #{expected_id}, but it was #{lock_id} instead"
+      assert_equal lock_id, expected_id, "expected lock id generated by the migrator to be #{expected_id}, but it was #{lock_id} instead"
```
yahonda:
```suggestion
      assert_equal expected_id, lock_id, "expected lock id generated by the migrator to be #{expected_id}, but it was #{lock_id} instead"
```

`assert_equal(expected, actual, message = nil)` then the expected_id should comes first.

yahonda (changes requested):

shivamsinghchahar:
@yahonda I've made the suggested changes. Please check

rafaelfranca (commented):
I'm almost sure this is the kind of cop @dhh would complain that is enabled.

We usually prefer to allow us to write the assertion in the way we believe would fit best in the case.

@dhh what do you think?

dhh:
You're right to flag this @rafaelfranca! I'm usually very sensitive to this kind of stuff. But in this case, I'm happy to add it. I didn't even realize that people would manually be doing == instead of assert_equal! I don't love that we need the three exceptions, and I probably would have been cursing if I was the one writing something like that, which needed to encode the exception, but given the volume of unnecessary `assert ==`'s we have in the code base, I'd still say it's worth it.

yahonda (approved):
Rafael is the code owner for .rubocop.yml, so I'll leave the decision to him.
https://github.com/rails/rails/blame/3a4961048ad251b50991ae83135d760a8a9e8ae3/.github/CODEOWNERS

shivamsinghchahar:
@rafaelfranca Requesting your review on this. Please let me know if anything else is needed here.

PR #30702, opened 2017-09-25
ActionCable: use find method when unsubscribing

rmachielse:
### Summary

If a frontend for some reason tries to unsubscribe from a non existing action cable subscription, the following error is logged:

```
Could not execute command from ({"command"=>"unsubscribe", "identifier"=>"{\"channel\":\"SomeChannel\"}"}) [NoMethodError - undefined method `unsubscribe_from_channel' for nil:NilClass]:
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:48:in `remove_subscription'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:44:in `remove'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:16:in `execute_command'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/base.rb:85:in `dispatch_websocket_message' /app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/server/worker.rb:58:in `block in invoke'
```

Instead, it will now properly log:

```
Could not execute command from ({"command"=>"unsubscribe", "identifier"=>"{\"channel\":\"SomeChannel\"}"}) [RuntimeError - Unable to find subscription with identifier: {"channel":"SomeChannel"}]:
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:76:in `find'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:44:in `remove'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/subscriptions.rb:16:in `execute_command'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/connection/base.rb:85:in `dispatch_websocket_message'
/app/vendor/bundle/ruby/2.4.0/gems/actioncable-5.1.4/lib/action_cable/server/worker.rb:58:in `block in invoke
```

It's a subtle difference but it helps a lot when debugging.

Issue #25381, opened 2016-06-13
ActionCable: Unable to find subscription with identifier...


tomekw:
Rails 5.0.0.rc1
Ruby 2.3.1 MRI + Puma

I've only one channel `ApiChannel`, the client (iOS app) connects to the back-end, receives the subscription confirmation and immediately sends a command. Sometimes the client is greeted with `RuntimeError` exception:
https://github.com/rails/rails/blob/a8df1bc345bf062e98bda64ef9638f65292c4b56/actioncable/lib/action_cable/connection/subscriptions.rb#L70

I've added some debugging statements to the code and from time to time the `subscriptions` Hash is indeed empty (the client issues `subscribe` command before), sometimes `identifiers` method returns empty Array for non-empty `subscriptions` (but it's just `.keys` on it).

I spent a lot of time debugging this and I'm out of ideas. Am I doing anything wrong here? Any help would be appreciated!

maclover7:
Hmm, are you verifying that the subscription confirmation is received before you are sending any commands? Also, is there any reliable way to reproduce this? 😬

tomekw:
@maclover7 the client sends command **immediately** after receiving `{"identifier"=>"{\"channel\":\"ApiChannell\"}", "type"=>"confirm_subscription"}` payload.

Also, the back-end propagates client status updated in `subscribed` callback to other connected clients (sort of home-made presence feature).

Reliable way? Hard to tell right now, looks like some kind of race condition... Maybe I would be able to prepare some minimal app to reproduce this, let me try!

maclover7:
Yep, seems like a race condition to me as well 😞😬

jbence:
Is there perhaps a spelling issue here? I notice that it is called ```ApiChannel``` in the initial description, but the reported payload comment has an extra lower-case L on the end.

Looking at the code, there are conditions in the ```add``` method that would result in an empty ```subscriptions``` hash, with an error message going to the log. Checking for that error message might shed some light.

Spelling would produce a consistent error, so this would have to be a situation where sometimes it is spelled correctly... that seems unlikely. But there is that path in the code for ```add``` that results in no update of the ```subscriptions``` hash. Perhaps that is the path taken before the failure occurs.

tomekw:
@jbence I've made a typo in an issue description. Thanks! And no, there is no typo in the code.

Right now we've added some small sleep after subscription is confirmed and, as for now, the error didn't occur. Will post more findings tomorrow.

[...]

PR #56375, opened 2025-12-15
Make ActionCable unsubscribe idempotent

matthewbjones:
### Motivation / Background

`ActionCable` currently raises a `RuntimeError` when a client attempts to unsubscribe from a subscription that no longer exists on the server. This commonly occurs due to race conditions: reconnects, server restarts, or rapid subscribe/unsubscribe cycles (especially with Turbo Streams).

The error is caught by `execute_command`'s rescue block, which calls `rescue_with_handler`. This triggers any `rescue_from` handlers defined in the application's connection class, commonly used to report exceptions to error monitoring services like Sentry, creating noise in production error tracking. It also represents incorrect semantics, as unsubscribe should be idempotent.

Related to #25381 and #30702.

### Detail

This Pull Request changes `find` to return `nil` instead of raising and moves the error handling to the callers. Currently, `find` is only called in two places:

- `remove` now checks for nil and returns early (idempotent no-op)
- `perform_action` still raises if subscription not found (this indicates a real client bug)

This distinction is important: unsubscribing from something that doesn't exist is a successful no-op, while trying to perform an action on a non-existent subscription indicates a real client bug that should still raise.

### Additional information

One issue #25381 was opened in 2016. PR #30702 improved the error message but preserved the raising behavior. This PR completes the fix by making unsubscribe truly idempotent while preserving the error for `perform_action`.

### Checklist

### DIFF
--- a/actioncable/lib/action_cable/connection/subscriptions.rb
+++ b/actioncable/lib/action_cable/connection/subscriptions.rb
@@ -44,6 +44,12 @@ module ActionCable
         end
       end
 
+      class MissingIdentifier < Error
+        def initialize
+          super "Identifier is required"
+        end
+      end
+
       def initialize(connection)
         @connection = connection
         @subscriptions = {}
@@ -78,8 +84,11 @@ module ActionCable
       end
 
       def remove(data)
+        raise MissingIdentifier unless data["identifier"].present?
+
         logger.info "Unsubscribing from channel: #{data['identifier']}"
-        remove_subscription find(data)
+        subscription = find(data)
+        remove_subscription(subscription) if subscription
       end
 
       def remove_subscription(subscription)
@@ -88,7 +97,9 @@ module ActionCable
       end
 
       def perform_action(data)
-        find(data).perform_action ActiveSupport::JSON.decode(data["data"])
+        subscription = find(data)
+        raise UnknownSubscription.new(data["identifier"]) unless subscription
+        subscription.perform_action ActiveSupport::JSON.decode(data["data"])
       end
 
       def identifiers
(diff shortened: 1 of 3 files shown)


actioncable/lib/action_cable/connection/subscriptions.rb
```diff
@@ -74,11 +77,7 @@ def unsubscribe_from_all
         delegate :logger, to: :connection
 
         def find(data)
-          if subscription = subscriptions[data["identifier"]]
-            subscription
-          else
-            raise "Unable to find subscription with identifier: #{data['identifier']}"
-          end
+          subscriptions[data["identifier"]]
```
thadeu, line 118: Instead of `def find`, why did you dont use `key?` like

`subscriptions.key?(data["identifier"])` it's more readable to us, like: `Did you have a key X?`

matthewbjones: I'm sorry, but I don't quite understand the code review feedback. The `find()` method is existing and goes from existing functionality of returning non-nil when there is a subscription or raising an exception if missing, to returning non-nil or nil. There are 2 call sites that need to look up a subscription, so I thought keeping `find()` still made sense since it centralizes the logic of how to find a subscription for a given `data`. However, `remove()` should not raise, but `perform_action()` still should raise.

I'm not understanding the feedback of using `key?`, especially given the context that the existing code for `find()` did not make use of `key?` either.

thadeu: The `subscriptions[data["identifier"]]` will be load the entire data about subscription and `subscriptions.key?(data["identifier"])` just tell us if the subscription exist or not in the subscriptions hashmap.

I think that is will allocate less memory

PR #56376, opened 2025-12-16
Allow disabling transactional callbacks on a model

kirs:
All validations and model callbacks are wrapped into `BEGIN` / `COMMIT` for transactional consistency.

However, in a large-scale apps with high concurrency and high DB connections usage, that brings the downside of that `BEGIN` holding the connection from being reused by another workers.

On top of that, in a large scale app, you'd run in isolation level like `READ COMMITTED` where keeping callbacks part of the same txn doesn't give you much.

I've tried to write a version of this PR that would automatically kick in if the model has no callbacks, but then I've realized that validations are also part of this. And there's no good way to check whether validations are pure Ruby (`validates :author_id, presence: true`) or actual database (`validates :author, presence: true`).

So I've opted to make this a manual thing per model.

This would already help us at Shopify to save some connections usage across some very hot paths.

### DIFF
--- a/activerecord/lib/active_record/transactions.rb
+++ b/activerecord/lib/active_record/transactions.rb
@@ -10,6 +10,8 @@ module ActiveRecord
       define_callbacks :commit, :rollback,
                        :before_commit,
                        scope: [:kind, :name]
+
+      class_attribute :transactional_callbacks, instance_writer: false, default: true
     end
 
     attr_accessor :_new_record_before_last_commit, :_last_transaction_return_status # :nodoc:
@@ -425,6 +427,10 @@ module ActiveRecord
     # This method is available within the context of an ActiveRecord::Base
     # instance.
     def with_transaction_returning_status # :nodoc:
+      unless self.class.transactional_callbacks
+        return yield
+      end
+
       self.class.with_connection do |connection|
         connection.pool.with_pool_transaction_isolation_level(ActiveRecord.default_transaction_isolation_level, connection.transaction_open?) do
           status = nil
--- a/activerecord/test/cases/transactions_test.rb
+++ b/activerecord/test/cases/transactions_test.rb
@@ -1525,6 +1525,56 @@ class TransactionTest < ActiveRecord::TestCase
     assert_array_match expected_queries, actual_queries
   end
 
+  class TopicWithoutCallbacks < ActiveRecord::Base
+    self.table_name = "topics"
+    self.transactional_callbacks = false
+  end
+
+  def test_save_without_callbacks_does_not_materialize_transaction
+    topic = TopicWithoutCallbacks.find(topics(:fifth).id)
+    actual_queries = capture_sql do
+      topic.title = "foo"
+      topic.save!
+    end
+
+    assert_array_match [/UPDATE/i], actual_queries
+  end
+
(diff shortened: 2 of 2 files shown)


activerecord/test/cases/transactions_test.rb
```diff
@@ -1525,6 +1525,56 @@ def test_nested_transactions_skip_excess_savepoints
     assert_array_match expected_queries, actual_queries
   end
 
+  class TopicWithoutCallbacks < ActiveRecord::Base
+    self.table_name = "topics"
+    self.transactional_callbacks = false
+  end
+
+  def test_save_without_callbacks_does_not_materialize_transaction
```
juniper-shopify, line 1533: I think it's worth adding at least one test case that asserts the transaction *is* opened on the connection when `transactional_callbacks` is `true`.

There's no test case in this file that explicitly checks for that. The closest, I think, is `"test_cancellation_from_before_filters_rollbacks_in_#{filter}"` (line 715) that asserts the transaction gets rolled back when the validation fails, which technically does cover this behaviour but in a very oblique way.

activerecord/lib/active_record/transactions.rb
```diff
@@ -10,6 +10,8 @@ module Transactions
       define_callbacks :commit, :rollback,
                        :before_commit,
                        scope: [:kind, :name]
+
+      class_attribute :transactional_callbacks, instance_writer: false, default: true
```
eileencodes, line 14: Does this make sense as an application wide setting as well? Something like `config.active_record.disable_transactional_callbacks = true`?

I also think the controller level API should be `disable_transactional_callbacks` and default to `false`. So it's clearer what's happening in a controller

eileencodes (changes requested):
Can you add a changelog? I also left a comment about the API.

activerecord/lib/active_record/transactions.rb
kirs: I'm not sure it makes sense per application -- especially given that you could set it on `ApplicationRecord` anyways.

I specifically wanted to avoid something like `disable_transactional_callbacks` because in my opinion that double negation takes more mental cycles to understand.

It's also more in line with other existing accessors like `record_timestamps`, `pluralize_table_names`, `verify_foreign_keys_for_fixtures`, `automatic_scope_inversing`.

what do you think?

`use_transactional_callbacks` is another slight modification of the accessor to consider.

eileencodes: > I'm not sure it makes sense per application -- especially given that you could set it on ApplicationRecord anyways.

Ok works for me, we can always add it later.

> what do you think?

IMO `record_`, `pluralize_` and `verify_` are all "actions" the user is taking to change default behavior. `automatic_scope_inversing` is the odd one that doesn't follow that pattern. I think `use_transactional_callbacks` is fine, but I would prefer the config option start with an action (`enable`, `disable`, `use`). Curious if anyone else from core has an opinion. What do you think @matthewd?

kirs:
Discussed this with @matthewd and his feedback was:

> it just seems unreasonably dangerous for people to configure a model once to turn off that guarantee, and have that apply to any future-defined callback, from some AR-using gem, say

Which I agree with.

https://github.com/rails/rails/pull/56732 might be able to solve this in a cleaner way.

PR #56379, opened 2025-12-16
fix: Reset the schema_search_path during pg reconnect and reset

eglitobias:
### Motivation / Background

This Pull Request has been created to fix a PostgreSQL adapter bug where the configured `schema_search_path` is not reapplied after a connection is reset or reconnected.

In Rails 8.1, calling `ActiveRecord::Base.connection.reconnect!` (or triggering a connection reset via the pool) creates a new PostgreSQL session, which resets all session-level state such as `search_path`. However, `PostgreSQLAdapter` caches the previously applied `@schema_search_path`, causing `schema_search_path=` to incorrectly short-circuit and skip reapplying the configured search path.

This results in PostgreSQL falling back to its default `"$user", public` search path, breaking applications that rely on multiple schemas.

**Fixes #56378; closes #56382**

---

### Detail

This Pull Request clears the cached `@schema_search_path` whenever PostgreSQL session state is invalidated:

* When reconnecting via `PG::Connection#reset`
* When resetting the session via `DISCARD ALL`

By resetting the cached value in these code paths, `configure_connection` correctly reapplies the configured `schema_search_path` on the new session.

This preserves existing guard logic and avoids unnecessary `SET search_path` calls during normal operation, while ensuring correctness after reconnects and resets.

---

### Additional information

* This issue affects both the `postgresql` and `postgis` adapters
* The behavior is reproducible with a minimal standalone Active Record script (see linked issue)
* Setting the search path at the database or role level works as a workaround, but `database.yml` configuration should be sufficient
* The fix is PostgreSQL-specific and does not affect other adapters

---

### Checklist

- [x] This Pull Request is related to one change. Unrelated changes are not included.
- [x] Commit message includes a detailed description and references the issue.
- [x] Tests are added / updated to cover reconnect and reset behavior.
- [x] CHANGELOG update not required for a bug fix.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -314,7 +314,9 @@ module ActiveRecord
 
         # Returns the active schema search path.
         def schema_search_path
-          @schema_search_path ||= query_value("SHOW search_path")
+          @schema_search_path ||=
+            with_raw_connection { |conn| conn.parameter_status("search_path") } ||
+            query_value("SHOW search_path")
         end
 
         # Returns the current client message level.
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -413,6 +413,11 @@ module ActiveRecord
         end
       end
 
+      def clear_cache!(new_connection: false)
+        super
+        @schema_search_path = nil if new_connection
+      end
+
       # Disconnects from the database if already connected. Otherwise, this
       # method does nothing.
       def disconnect!
@@ -1038,6 +1043,8 @@ module ActiveRecord
 
           add_pg_encoders
           add_pg_decoders
+
+          schema_search_path # populate cache
 
           reload_type_map
         end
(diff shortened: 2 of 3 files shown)


eglitobias:
This issue was caused by https://github.com/rails/rails/pull/54698
Currently a spec is failing because of my change. If I just revert 54698 everything (including my new tests) is green.

eglitobias:
test_querying_a_'clean'_recently-used_but_now-failed_connection_skips_verification fails when run in isolation on main. This change does not introduce the failure; it makes an existing order dependency more visible.

I verified this locally (on the current main):

Running the test in isolation:
```
ARCONN=postgresql \
PGHOST=localhost \
PGPORT=5432 \
PGUSER=postgres \
bin/test activerecord/test/cases/adapter_test.rb:701
```
→ fails 

Running the full file:
```
ARCONN=postgresql \
PGHOST=localhost \
PGPORT=5432 \
PGUSER=postgres \
bin/test activerecord/test/cases/adapter_test.rb
```
→ passes

What would you recommend as the best way forward here - should I adjust the fix (just rollback https://github.com/rails/rails/pull/54698), or can someone show me how tho fix this spec?

I have now pushed a small fix for the spec. What do you think?

matthewd:
Thanks @eglitobias!

I changed the attribution line in your commit messages, just because GitHub can make being at-mentioned in public commits a pain; and removed the changelog entry because I don't see it as a user-facing change... and then ended up having to rebase to catch up with CI changes.

eglitobias:
@matthewd Thanks for getting this merged!
What’s the process for getting a fix like this into Rails 8.1? Should I open a PR against 8-1-stable, and who decides whether both the improvement and the fix are backported?

PR #56382, opened 2025-12-16
Fix schema_search_path cache invalidation after reconnect/reset

matthewd:
The cached value wasn't cleared when the connection was reset, so the setter's guard would skip the `SET` command when `configure_connection` tried to restore the configured value.

In passing, eagerly populate the cache during connection setup, and use `parameter_status` on PG18+ to avoid the query entirely.

Fixes #56378; alternative to / closes #56379


andreimaxim:
Hi @matthewd is there something I could do to help merge this ticket? We've had intermittent issues due to the schema search path since upgrading to Rails 8.1 and, after days of debugging, finally found this PR.

We've worked around this by creating a role with a preset search path, but ideally this could be fixed in Rails.

PR #56384, opened 2025-12-16
Add ActionController::Parameters#fetch_values for multi-key fetching

Saidbek:
### Motivation / Background

This PR has been created because I'd like a nicer way of fetching multiple required params at once when using `ActionController::Parameters`. Currently, I need to call `fetch` multiple times, which is verbose and repetitive. This is similar to Ruby's `Hash#fetch_values` method, which makes it easy to fetch multiple values in a single call.

### Details
```ruby
params = ActionController::Parameters.new(name: "Francesco", age: 22)
params.fetch_values(:name, :age)                # => ["Francesco", 22]
params.fetch_values(:name, :none)               # => ActionController::ParameterMissing: param is missing or the value is empty or invalid: none
params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
```

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/strong_parameters.rb
+++ b/actionpack/lib/action_controller/metal/strong_parameters.rb
@@ -833,6 +833,28 @@ module ActionController
       )
     end
 
+    # Returns parameters for the given keys. If a key can't be found, there are
+    # several options: With no other arguments, it will raise an
+    # ActionController::ParameterMissing error; if a block is given, then that will
+    # be run and its result returned for the missing key.
+    #
+    #     params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    #     params.fetch_values(:name, :age)                # => ["Francesco", 22]
+    #     params.fetch_values(:name, :none)               # => ActionController::ParameterMissing: param is missing or the value is empty or invalid: none
+    #     params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
+    def fetch_values(*keys)
+      original_key_lookup = keys.index_by { |key| key.to_s }
+      values = @parameters.fetch_values(*keys) do |missing_key|
+        original_key = original_key_lookup[missing_key]
+        if block_given?
+          yield original_key
+        else
+          raise ActionController::ParameterMissing.new(original_key, @parameters.keys)
+        end
+      end
+      values.map! { |value| convert_value_to_parameters(value) }
+    end
+
     # Extracts the nested parameter from the given `keys` by calling `dig` at each
     # step. Returns `nil` if any intermediate step is `nil`.
     #
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,17 @@
+*   Add `ActionController::Parameters#fetch_values` for fetching multiple parameter values
+
+    ```ruby
+    params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    name, age = params.fetch_values(:name, :age)
+    # => ["Francesco", 22]
+
+    # With default values via block
+    name, email = params.fetch_values(:name, :email) { |key| "default_#{key}" }
+    # => ["Francesco", "default_email"]
+    ```
+
+    *Said Kaldybaev*
+
 *   Serve static CSS and HTML files with `charset=utf-8` in the Content-Type header.
 
     Static CSS and HTML files served by `ActionDispatch::Static` now include
--- a/actionpack/test/controller/parameters/accessors_test.rb
+++ b/actionpack/test/controller/parameters/accessors_test.rb
@@ -191,6 +191,47 @@ class ParametersAccessorsTest < ActiveSupport::TestCase
     assert_equal :missing, key
   end
 
+  test "fetch_values returns values for given keys" do
+    params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    name, age = params.fetch_values(:name, :age)
+
+    assert_equal "Francesco", name
+    assert_equal 22, age
+  end
+
+  test "fetch_values raises ParameterMissing when string key not found" do
+    params = ActionController::Parameters.new(name: "Francesco")
+    error = assert_raises(ActionController::ParameterMissing) do
+      params.fetch_values(:name, "missing_key")
+    end
+    assert_equal "missing_key", error.param
+  end
+
+  test "fetch_values raises ParameterMissing when symbol key not found" do
+    params = ActionController::Parameters.new(name: "Francesco")
+    error = assert_raises(ActionController::ParameterMissing) do
+      params.fetch_values(:name, :missing_key)
+    end
+    assert_equal :missing_key, error.param
+  end
+
+  test "fetch_values with block returns block result for missing keys" do
+    params = ActionController::Parameters.new(name: "Francesco")
+    assert_equal ["Francesco", :age], params.fetch_values(:name, :age, &:itself)
+    assert_equal ["Francesco", "age"], params.fetch_values(:name, "age", &:itself)
+  end
+
+  test "fetch_values retains permitted status" do
+    @params.permit!
+    values = @params.fetch_values(:person)
+    assert_predicate values[0], :permitted?
+  end
+
+  test "fetch_values retains unpermitted status" do
+    values = @params.fetch_values(:person)
+    assert_not_predicate values[0], :permitted?
+  end
+
   test "has_key? returns true if the given key is present in the params" do
     assert @params.has_key?(:person)
   end


byroot:
The feature seem valuable, but the implementation isn't correct.

Look at the `#fetch` method, you have to convert sub hashes into `Parameter` instances.

Also `Hash#fetch_values` is a thing so you can just delegate to that rather than implement it again.

Also please have a clean git history with just one commit.

Saidbek:
@byroot thanks! Updated the code - mind taking another look?

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -833,6 +833,27 @@ def fetch(key, *args)
...
+    # ActionController::ParameterMissing error; if a block is given, then that will
+    # be run and its result returned for the missing key.
+    #
+    #     params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    #     params.fetch_values(:name, :age)                # => ["Francesco", 22]
+    #     params.fetch_values(:name, :none)               # => ActionController::ParameterMissing: param is missing or the value is empty or invalid: none
+    #     params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
+    def fetch_values(*keys)
+      values = @parameters.fetch_values(*keys) do |missing_key|
+        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym)
```
seanpdoyle: What is the significance of casting the key to be a symbol? How is yielding the key value returned by the underlying instance (a string, likely) insufficient?

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -833,6 +833,27 @@ def fetch(key, *args)
...
+        if block_given?
+          yield missing_key
+        else
+          raise ActionController::ParameterMissing.new(missing_key, @parameters.keys)
+        end
+      end
+      keys.zip(values).map { |key, value| convert_hashes_to_parameters(key, value) }
```
seanpdoyle: Could this be simplified to use `convert_value_to_parameters`?

```suggestion
      values.map { |value| convert_value_to_parameters(value) }
```

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -833,6 +833,27 @@ def fetch(key, *args)
...
+    #
+    #     params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    #     params.fetch_values(:name, :age)                # => ["Francesco", 22]
+    #     params.fetch_values(:name, :none)               # => ActionController::ParameterMissing: param is missing or the value is empty or invalid: none
+    #     params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
+    def fetch_values(*keys)
+      values = @parameters.fetch_values(*keys) do |missing_key|
+        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym)
+        if block_given?
+          yield missing_key
```
seanpdoyle: The [Hash#fetch_values](https://docs.ruby-lang.org/en/master/Hash.html#method-i-fetch_values) method yields the missing key, so it's good that this implementation does as well.

While reviewing this change, I realized that the `ActionController::Parameters#fetch` method *does not* yield the missing key to the block, despite the fact that [Hash#fetch](https://docs.ruby-lang.org/en/master/Hash.html#method-i-fetch) does.

I've opened https://github.com/rails/rails/pull/56400 to match the built-in behavior.

actionpack/lib/action_controller/metal/strong_parameters.rb
Saidbek: Otherwise, `HashWithIndifferentAccess#fetch_values` hands us a string key, so the error would carry a string instead of the symbol that `fetch` raises with - casting keeps `ParameterMissing#param` consistent with `fetch`.
```ruby
irb(main):009> h = {a: 1}.with_indifferent_access; h.fetch_values(:b) {|k| k}
=> ["b"]
```

seanpdoyle: Hmm, that surprised me. I was able to add the following assertions for both String and Symbol keys when evaluating the `fetch(…) { … }` block:

https://github.com/rails/rails/blob/45ab692176ea18728fcf0b01f0a22e30d672a5d6/actionpack/test/controller/parameters/accessors_test.rb#L182-L192

Is there a difference between `ActiveSupport::HashWithIndifferentAccess#fetch` and `ActiveSupport::HashWithIndifferentAccess#fetch_values`:

https://github.com/rails/rails/blob/45ab692176ea18728fcf0b01f0a22e30d672a5d6/activesupport/lib/active_support/hash_with_indifferent_access.rb#L211-L213

https://github.com/rails/rails/blob/45ab692176ea18728fcf0b01f0a22e30d672a5d6/activesupport/lib/active_support/hash_with_indifferent_access.rb#L267-L270

seanpdoyle: Ah, I see now. The `yield key` expression used `ActionController::Parameters#fetch` references the `key` method argument, and not the key being yielded by the `ActiveSupport::HashWithIndifferentAccess#fetch` call.

I wonder if it's worthwhile to change the implementation to account for maintaining the type of the key, rather than coercing all missing keys to symbols.

```diff
diff --git a/actionpack/lib/action_controller/metal/strong_parameters.rb b/actionpack/lib/action_controller/metal/strong_parameters.rb
index 30e20c56e1..9a93bb906c 100644
--- a/actionpack/lib/action_controller/metal/strong_parameters.rb
+++ b/actionpack/lib/action_controller/metal/strong_parameters.rb
@@ -844,7 +844,8 @@ def fetch(key, *args)
     #     params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
     def fetch_values(*keys)
       values = @parameters.fetch_values(*keys) do |missing_key|
-        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym)
+        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym) && !keys.include?(missing_key)
+
         if block_given?
           yield missing_key
         else
diff --git a/actionpack/test/controller/parameters/accessors_test.rb b/actionpack/test/controller/parameters/accessors_test.rb
index 34e2fdcf54..a7f46316c9 100644
--- a/actionpack/test/controller/parameters/accessors_test.rb
+++ b/actionpack/test/controller/parameters/accessors_test.rb
@@ -199,7 +199,15 @@ class ParametersAccessorsTest < ActiveSupport::TestCase
     assert_equal 22, age
   end
 
-  test "fetch_values raises ParameterMissing when key not found" do
+  test "fetch_values raises ParameterMissing when string key not found" do
+    params = ActionController::Parameters.new(name: "Francesco")
+    error = assert_raises(ActionController::ParameterMissing) do
+      params.fetch_values(:name, "missing_key")
+    end
+    assert_equal "missing_key", error.param
+  end
+
+  test "fetch_values raises ParameterMissing when symbol key not found" do
     params = ActionController::Parameters.new(name: "Francesco")
     error = assert_raises(ActionController::ParameterMissing) do
       params.fetch_values(:name, :missing_key)
@@ -209,8 +217,8 @@ class ParametersAccessorsTest < ActiveSupport::TestCase
 
   test "fetch_values with block returns block result for missing keys" do
     params = ActionController::Parameters.new(name: "Francesco")
-    values = params.fetch_values(:name, :age) { |key| "default_#{key}" }
-    assert_equal ["Francesco", "default_age"], values
+    assert_equal ["Francesco", :age], params.fetch_values(:name, :age, &:itself)
+    assert_equal ["Francesco", "age"], params.fetch_values(:name, "age", &:itself)
   end
 
   test "fetch_values retains permitted status" do
```

seanpdoyle: In addition to the tests mentioned above, this line would change:

```suggestion
        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym) && !keys.include?(missing_key)
```

seanpdoyle (approved):

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -833,6 +833,27 @@ def fetch(key, *args)
...
+    def fetch_values(*keys)
+      values = @parameters.fetch_values(*keys) do |missing_key|
+        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym) && !keys.include?(missing_key)
+        if block_given?
+          yield missing_key
+        else
+          raise ActionController::ParameterMissing.new(missing_key, @parameters.keys)
+        end
+      end
+      values.map { |value| convert_value_to_parameters(value) }
```
seanpdoyle: Would it be more efficient to invoke [Array#map!](https://docs.ruby-lang.org/en/master/Array.html#method-i-map-21), since the `values` array was allocated and populated on the prior line:

```suggestion
      values.map! { |value| convert_value_to_parameters(value) }
```

Saidbek: Sure, manipulating in place would avoid allocating another array

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -833,6 +833,27 @@ def fetch(key, *args)
...
+    # ActionController::ParameterMissing error; if a block is given, then that will
+    # be run and its result returned for the missing key.
+    #
+    #     params = ActionController::Parameters.new(name: "Francesco", age: 22)
+    #     params.fetch_values(:name, :age)                # => ["Francesco", 22]
+    #     params.fetch_values(:name, :none)               # => ActionController::ParameterMissing: param is missing or the value is empty or invalid: none
+    #     params.fetch_values(:name, :none) { |key| key } # => ["Francesco", :none]
+    def fetch_values(*keys)
+      values = @parameters.fetch_values(*keys) do |missing_key|
+        missing_key = missing_key.to_sym if missing_key.respond_to?(:to_sym) && !keys.include?(missing_key)
```
byroot: that `!keys.include?` is terrible for performance, means the method is now `O(N*M)`.

If you need to check the key you can directly do `key?(missing_key)` for `O(1)` access, but also I'm not sure it's really worth checking for that.

Saidbek: That line preserves the original key type so `fetch_values` matches `fetch` behavior. As an additional improvement, I replaced the O(NM) `keys.include?` check with an O(1) hash lookup using `index_by`, reducing overall complexity to O(N). https://github.com/rails/rails/pull/56384/commits/0c1e27c70b0c36e7bfb0648ad670c03a73086b41

PR #56385, opened 2025-12-17
In MT6, test_order -> run_order. Add AS::TC#run_order as needed

zenspider:
Rather than rename everything on the rails side, only provide a bridge
from AS::TC#test_order to MT6's MT::Test#run_order if AS::TC responds
to it.

### DIFF
--- a/activesupport/lib/active_support/test_case.rb
+++ b/activesupport/lib/active_support/test_case.rb
@@ -62,6 +62,12 @@ module ActiveSupport
         ActiveSupport.test_order ||= :random
       end
 
+      if Minitest.respond_to? :run_order # MT6 API change
+        def run_order # :nodoc:
+          test_order
+        end
+      end
+
       # Parallelizes the test suite.
       #
       # Takes a +workers+ argument that controls how many times the process


activesupport/lib/active_support/test_case.rb
```diff
@@ -62,6 +62,12 @@ def test_order
         ActiveSupport.test_order ||= :random
       end
 
+      if Minitest.respond_to? :run_order # MT6 API change
+        def run_order # :nodoc:
+          test_order
+        end
+      end
```
vlad-pisanov, line 69: Shouldn't this be 

```ruby
if Minitest::Test.respond_to? :run_order
```

zenspider: Good catch! Actually should be `Minitest::Runnable` to be thoroughly pedantic.

You wanna do a PR or should I? (fine either way)

zenspider: ping @vlad-pisanov

vlad-pisanov: go ahead

PR #56389, opened 2025-12-17
[Bug] Fix input[type=file] accept attribute array value separator as `,`

bogdan:
### Motivation / Background

It is great when all bugs are fixed.

### Detail

This Pull Request changes `input[type=file]` `accept` attribute to join multiple values as `,` instead of ` ` as per specification:

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/accept

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/tags/file_field.rb
+++ b/actionview/lib/action_view/helpers/tags/file_field.rb
@@ -6,6 +6,9 @@ module ActionView
       class FileField < TextField # :nodoc:
         def render
           include_hidden = @options.delete(:include_hidden)
+          if @options[:accept].is_a?(Array)
+            @options[:accept] = @options[:accept].join(",")
+          end
           options = @options.stringify_keys
           add_default_name_and_field(options)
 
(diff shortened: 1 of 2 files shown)


actionview/lib/action_view/helpers/tags/file_field.rb
```diff
@@ -6,6 +6,9 @@ module Tags # :nodoc:
       class FileField < TextField # :nodoc:
         def render
           include_hidden = @options.delete(:include_hidden)
+          if @options[:accept].is_a?(Array)
+            @options[:accept] = @options[:accept].join(",")
+          end
```
seanpdoyle, line 11: Is this a [breaking change](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#breaking-changes)? It might require introducing a configuration to control the [change in behavior](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#changing-behavior).

seanpdoyle (commented):
Could you include a [CHANGELOG](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#updating-the-changelog) entry?

actionview/lib/action_view/helpers/tags/file_field.rb
byroot: It's a bug fix.

byroot:
So technically, as documented `accept:` is supposed to be provided as a String, but I do agree it's natural to pass an Array there hence I do interpret it as a bug, and I backported it to 8.0 and 8.1.

PR #56393, opened 2025-12-17
Implement exclude keys for ActionController::Live execution sharing

### LINKED ISSUE #52906: `ActionController::Live` and `ActiveRecord::Base.connected_to` stale db state
### Steps to reproduce

Repro repo: https://github.com/jscheid/prevent-writes-test

Possibly related to #36830.

I've been hunting down an issue in our multi-db, passenger-based, postgres-based app where we see spurious errors "`Write query attempted while in readonly mode`" raised in code paths where we'd definitely expect the app to be connected to the primary (writable) db role.

I've managed to reliably reproduce in our app by limiting to one passenger worker (so that one process is being reused for all requests) and:
1. Hitting an `ActionController::Live` endpoint which uses `ActiveRecord::Base.connected_to(..., prevent_writes: true)`.
2. Provoking an `ActionController::Live::ClientDisconnected` to be raised during processing of said endpoint (just by disconnecting the client early.)
3. In the next request (served by the same worker), `preventing_writes?` will return true unexpectedly.

I can't share our whole app here and so instead I've built the repro test case linked above. Unfortunately I didn't yet manage to reproduce the exact same behavior, however it reliably demonstrates very similar behavior, except that the error thrown in step (3) is instead `No connection pool for 'ActiveRecord::Base' found for the role.` It's not the same error but it seems to indicate that some db state isn't being cleaned up properly in this scenario?

Things I've tried:
- Using a single-threaded, single-worker Puma server instead of a single-worker Passenger server - and I can see the same behavior, so it doesn't seem to depend on the web server being used.
- Using both latest Rails (7.2.1) and the version we're currently running in production (7.1.3.1), the behavior is the same in both cases.
- Using `Rails.application.executor.wrap` and/or `ActiveRecord::Base.connection_pool.with_connection` inside the `ActionController::Live` thread in hopes that it will isolate the environment stood up by `ActiveRecord::Base.connected_to`, but to no avail.
- Raising a `RuntimeError` from inside the Live thread (instead of `ActionController::Live::ClientDisconnected`) to see what happens, but interestingly this **doesn't** seem to trigger the same behavior.

Things I haven't tried yet:
- Using a database (connection handler) other than Postgres; but from reading the Rails code I think all the role/`prevent_writes` stuff is happening at the level of the abstract connection so I don't think it would help shed much light.

I can't rule out some kind of misconfiguration or misuse of `ActionController::Live` or `ActiveRecord::Base.connected_to` on our part - if so my apologies and I hope someone can point me into the right direction.

### Expected behavior

When using `ActiveRecord::Base.connected_to` inside `ActionController::Live.stream`, and when the client disconnects early, it should clean up after itself so that subsequent requests served by the same worker process have a clean database environment.

### Actual behavior

When using `ActiveRecord::Base.connected_to` inside `ActionController::Live.stream`, and when the client disconnects early, the database environment used for the next requests seems to be corrupted in some way that I don't fully understand yet.

### System configuration
**Rails version**: `7.2.1`, but also tested with `7.1.3.1` with identical results.
**Ruby version**: 3.3.5

eileencodes:
When using `ActionController::Live` the execution context is shared across threads. In some cases, like database context switching from a writer to a replica connection, you may not want that context to be shared.

This PR implements a config option for excluding specific keys from the shared execution. By default no keys are shared to support backwards compatibility in order to maintain current behavior while implementing a way for applications experiencing an issue with database connections to opt-out. See #52906.

To exclude a specific key application-wide, set the following config option:

```ruby
config.action_controller.live_streaming_excluded_keys = [:active_record_connected_to_stack]
```

If you want finer-grained control over which controllers exclude these keys you can set `live_streaming_excluded_keys` in the controller itself.

```ruby
class MyController < ApplicationController
  include ActionController::Live
  self.live_streaming_excluded_keys = [:active_record_connected_to_stack]

  def index
    ...
  end
end
```

In the future we may want to default to excluding this key since I'm not sure anyone wants to share database context. However I chose not to do that in this PR because we need to backport this to use as a bug fix, and if it changes behavior that could break applications.

Fixes #52906

### DIFF
--- a/actionpack/lib/action_controller/metal/live.rb
+++ b/actionpack/lib/action_controller/metal/live.rb
@@ -53,8 +53,49 @@ module ActionController
   #       response.headers["Last-Modified"] = Time.now.httpdate # Add this line if your Rack version is 2.2.x
   #       ...
   #     end
+  #
+  # ## Streaming and Execution State
+  #
+  # When streaming, the action is executed in a separate thread. By default, this thread
+  # shares execution state from the parent thread.
+  #
+  # You can configure which execution state keys should be excluded from being shared
+  # using the `config.action_controller.live.streaming_excluded_keys` configuration:
+  #
+  #   # config/application.rb
+  #   config.action_controller.live.streaming_excluded_keys = [:active_record_connected_to_stack]
+  #
+  # This is useful when using ActionController::Live inside a `connected_to` block. For example,
+  # if the parent request is reading from a replica using `connected_to(role: :reading)`, you may
+  # want the streaming thread to use its own connection context instead of inheriting the read-only
+  # context:
+  #
+  #   # Without configuration, streaming thread inherits read-only connection
+  #   ActiveRecord::Base.connected_to(role: :reading) do
+  #     @posts = Post.all
+  #     render stream: true # Streaming thread cannot write to database
+  #   end
+  #
+  #   # With configuration, streaming thread gets fresh connection context
+  #   # config.action_controller.live.streaming_excluded_keys = [:active_record_connected_to_stack]
+  #   ActiveRecord::Base.connected_to(role: :reading) do
+  #     @posts = Post.all
+  #     render stream: true # Streaming thread can write to database if needed
+  #   end
+  #
+  # Common keys you might want to exclude:
+  # - `:active_record_connected_to_stack` - Database connection routing and roles
+  # - `:active_record_prohibit_shard_swapping` - Shard swapping restrictions
+  #
+  # By default, no keys are excluded to maintain backward compatibility.
   module Live
     extend ActiveSupport::Concern
+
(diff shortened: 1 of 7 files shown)


actionpack/lib/action_controller/metal/live.rb
```diff
@@ -53,9 +53,50 @@ module ActionController
...
   #       ...
   #     end
+  #
+  # ## Streaming and Execution State
+  #
+  # When streaming, the action is executed in a separate thread. By default, this thread
+  # shares execution state from the parent thread.
+  #
+  # You can configure which execution state keys should be excluded from being shared
+  # using the `config.action_controller.live.streaming_excluded_keys` configuration:
```
tavianator, line 63: Should this be
```suggestion
  # using the `config.action_controller.live_streaming_excluded_keys` configuration:
```
?

PR #56395, opened 2025-12-18
ActionView#content_for uses keyword argument to replace options hash

drjayvee:
# Motivation / Background

This PR improves the signature for `content_for` by replacing the `options` hash with a modern keyword argument.

The benefits are:
* Invalid arguments will eventually (see below) `raise ArgumentError` instead of being silently ignored.
  (e.g., `content_for :foo, fluhs: true { "whoops" }`)
* Keyword arguments are documented in the method's signature instead of in its body.
  Better for humans, IDEs and LLMs.
* Remove confusing argument shifting (`options = content if content` 🤨).

## Update
Based on seanpdoyle's review, the scope of this change was reduced to maintain backwards compatibility.

The only change now is that options other than `flush` will trigger a deprecation warning.

# Backwards-compatibility
The current behavior is kept but deprecated. The added warning will not only warn users of a future change, it will also alert them to bugs in their code. (Any invalid option is almost guaranteed to be a typo!)

~~Note that I've added the BC-keeping code in the second commit. If this is deemed overkill ("progress over stability" 😉), I can easily drop it. Otherwise, I'd be more than happy to squash the two commits.~~

# Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/capture_helper.rb
+++ b/actionview/lib/action_view/helpers/capture_helper.rb
@@ -148,7 +148,7 @@ module ActionView
       #
       #   <ul><%= content_for :navigation %></ul>
       #
-      # If the flush parameter is +true+ <tt>content_for</tt> replaces the blocks it is given. For example:
+      # If the flush parameter is +true+, <tt>content_for</tt> replaces the blocks it is given. For example:
       #
       #   <% content_for :navigation do %>
       #     <li><%= link_to 'Home', action: 'index' %></li>
@@ -175,8 +175,15 @@ module ActionView
             options = content if content
             content = capture(&block)
           end
+          flush = options.delete(:flush) { false }
+          if options.any?
+            ActionView.deprecator.warn(<<-MSG.squish)
+              Passing invalid options to content_for is deprecated.
+              Options #{options.keys.join(', ')} will be ignored.
+            MSG
+          end
           if content
-            options[:flush] ? @view_flow.set(name, content) : @view_flow.append(name, content)
+            flush ? @view_flow.set(name, content) : @view_flow.append(name, content)
           end
           nil
         else
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Deprecate invalid options passed to `content_for`. Invalid options were previously silently
+    ignored. For example, `content_for :foo, fluhs: true { "bar" }` (note the typo) would not
+    flush content, with no indication of the mistake.
+
+    *Jeroen Versteeg*
+
 *   Fix tag parameter content being overwritten instead of combined with tag block content.
     Before `tag.div("Hello ") { "World" }` would just return `<div>World</div>`, now it returns `<div>Hello World</div>`.
 
--- a/actionview/test/template/capture_helper_test.rb
+++ b/actionview/test/template/capture_helper_test.rb
@@ -178,6 +178,14 @@ class CaptureHelperTest < ActionView::TestCase
     assert_predicate content_for(:title), :html_safe?
   end
 
+  def test_content_for_deprecates_invalid_options
+    assert_deprecated(/Passing invalid options to content_for is deprecated/, ActionView.deprecator) do
+      content_for :name, invalid: true do
+        "foo"
+      end
+    end
+  end
+
   def test_provide
     assert_not content_for?(:title)
     provide :title, "hi"
--- a/guides/source/8_2_release_notes.md
+++ b/guides/source/8_2_release_notes.md
@@ -94,6 +94,15 @@ Please refer to the [Changelog][action-view] for detailed changes.
 
 ### Deprecations
 
+*   Deprecate invalid options (i.e., anything other than `flush:`) passed to `content_for`.
+
+    ```ruby
+    content_for(:foo, invalid: true) { "foo" }
+    ```
+
+    Invalid options were previously silently ignored. This will now emit deprecation warnings.
+    In the next major release, this will instead `raise ArgumentError`.
+
 ### Notable changes
 
 *   Add ability to pass a block when rendering a collection. The block is executed


actionview/lib/action_view/helpers/capture_helper.rb
```diff
@@ -169,14 +169,20 @@ def capture(*, **, &block)
       #   <% content_for :script, javascript_include_tag(:defaults) %>
       #
       # WARNING: <tt>content_for</tt> is ignored in caches. So you shouldn't use it for elements that will be fragment cached.
-      def content_for(name, content = nil, options = {}, &block)
+      def content_for(name, content = nil, flush: false, **options, &block)
+        if options.any?
+          ActionView.deprecator.warn(<<-MSG.squish)
+            Passing invalid options to content_for is deprecated.
+            Options #{options.keys.join(', ')} will be ignored.
+          MSG
```
seanpdoyle: Is this worth mentioning in the Action View Deprecations section of the [8.2 Release Notes](https://github.com/rails/rails/blob/61b5c7e2953fd0ccfab151b233ce69283c567a34/guides/source/8_2_release_notes.md#L57-L64)?

seanpdoyle (approved):

actionview/lib/action_view/helpers/capture_helper.rb
drjayvee: I guess? I'm new to Rails, so I can't judge whether this is worth mentioning.

Also, will this be *deprecated* in 8.2 and removed in 8.3 (or 8.4)? Or deprecated in 8.1 and removed in 8.2? That seems a bit rushed to me, since invalid options/keyword args will cause problems.

Please advise and I'll make any changes you deem sensible.

seanpdoyle: If this change is merged, the next release that it will be part of is 8.2.

In its current form, it will start emitting deprecation warnings. The Release Notes serve as an outward-facing list of behaviors that are deprecated, so adding an entry makes the motivation behind this underlying change more discoverable.

> Also, will this be deprecated in 8.2 and removed in 8.3

If the commit is merged ahead of the 8.2 release, it will be part of the 8.2 release. The timing of its removal (whether 8.3 or later) is up to Core.

drjayvee: All right, I'll add it to the `8_2_release_notes`.

drjayvee: I've added the following.

> Deprecate invalid options (i.e., anything other than flush:) passed to render_to.
>
> ```ruby
> render_to(:foo, invalid: true) { "foo" }
> ```
> 
> Invalid options were previously silently ignored. This will now emit deprecation warnings.
>
> In the next major release, this will instead `raise ArgumentError`.

Thoughts?

drjayvee:
@seanpdoyle, could you please help me get this PR merged?

I've rebased on top of `main` to update it.

actionview/lib/action_view/helpers/capture_helper.rb
```diff
@@ -169,14 +169,20 @@ def capture(*, **, &block)
       #   <% content_for :script, javascript_include_tag(:defaults) %>
       #
       # WARNING: <tt>content_for</tt> is ignored in caches. So you shouldn't use it for elements that will be fragment cached.
-      def content_for(name, content = nil, options = {}, &block)
+      def content_for(name, content = nil, flush: false, **options, &block)
+        if options.any?
```
seanpdoyle: For what it's worth, replacing `options = {}` with `**options` constitutes a breaking change to the method's interface. I'm not sure of Core's process for "upgrading" methods to use keyword arguments.

Could the deprecation be achieved without this change?


```suggestion
      def content_for(name, content = nil, options = {}, &block)
        flush = options.delete(:flush) { false }
        
        if options.any?
```

That way, the interface is preserved, the outcome is achieved, and a future PR could propose migrating to keyword arguments.

drjayvee: Thanks for your feedback.

> replacing options = {} with **options constitutes a breaking change to the method's interface.

Really? I don't understand how.

The added `test_content_for_deprecates_invalid_options` demonstrates that `content_for :name, invalid: true` triggers a deprecation notice but does not raise an error. Is that considered breaking?

> Could the deprecation be achieved without this change? That way, the interface is preserved, the outcome is achieved, and a future PR could propose migrating to keyword arguments.

As a newcomer to Ruby & Rails, this method's signature is confusing to me. _My goal_ for this PR is to replace the legacy options hash with modern keyword arguments to clarify the API. (Specifically: `flush` is the only valid keyword argument.)

seanpdoyle: The interface change from `options = {}` to `**options` is a breaking change for callers might have been passing in a `Hash` argument from outside the method invocation:

```
irb(main):001* def method_with_opts(opts = {})
irb(main):002*   puts opts
irb(main):003> end
=> :method_with_opts
irb(main):004* def method_with_kwargs(**opts)
irb(main):005*   puts opts
irb(main):006> end
=> :method_with_kwargs
irb(main):007> hash = { key: "value" }
=> {key: "value"}
irb(main):008> method_with_opts(hash)
{key: "value"}
=> nil
irb(main):009> method_with_kwargs(hash)
(irb):4:in 'method_with_kwargs': wrong number of arguments (given 1, expected 0) (ArgumentError)
        from (irb):9:in '<main>'
        from <internal:kernel>:168:in 'Kernel#loop'
        from …/.asdf/installs/ruby/3.4.8/lib/ruby/gems/3.4.0/gems/irb-1.17.0/exe/irb:9:in '<top (required)>'
        from …/.asdf/installs/ruby/3.4.8/bin/irb:25:in 'Kernel#load'
        from …/.asdf/installs/ruby/3.4.8/bin/irb:25:in '<main>'
irb(main):010> method_with_kwargs(**hash)
{key: "value"}
=> nil
```

drjayvee: Oh, I see! Thanks for pointing that out. Very well, I agree we have to avoid that breaking change.

One big downside is that I need to bring back the argument shifting logic which can assign `options = content`. Yuck!

Anyway, I restored the signature and squashed the commits.

actionview/lib/action_view/helpers/capture_helper.rb
```diff
@@ -175,8 +175,15 @@ def content_for(name, content = nil, options = {}, &block)
             options = content if content
             content = capture(&block)
           end
+          flush = options.delete(:flush) { false }
+          if options.any?
+            ActionView.deprecator.warn(<<-MSG.squish)
+              Passing invalid options to content_for is deprecated.
+              Options #{options.keys.join(', ')} will be ignored.
+            MSG
```
seanpdoyle, line 183: @skipkayhil I suspect that the follow-up to the deprecation would require either:

1. switching from the relax `options = {}` to a more restrictive `flush: false` keyword argument (a breaking change as discussed in https://github.com/rails/rails/pull/56395#discussion_r2878884533)
2. retaining the `options = {}`, and replacing the deprecation warning with something like a call to [Hash#assert_valid_keys](https://edgeapi.rubyonrails.org/classes/Hash.html#method-i-assert_valid_keys): `options.assert_valid_keys(:flush)` to raise instead of warn

Are there other possible outcomes worth considering? If not, I think that this PR is ready for final review.

PR #56396, opened 2025-12-18
Ensure Rails::TestUnitRailtie is defined and `config.generators.system_tests` is set when generating a system test

### LINKED ISSUE #56022: Standalone system_test generator not producing code
### Steps to reproduce

```
rails new hello_world
cd hello_world
bin/rails g model Account
bin/rails db:migrate
bin/rails g system_test Account
```

### Expected behavior

System test directory and files should be created.

E.g. what is produced using Rails 6.1.7.8
```
$ bin/rails g system_test Account
      invoke  test_unit
      create    test/application_system_test_case.rb
      create    test/system/accounts_test.rb
```
Or per the guide with model User (https://guides.rubyonrails.org/testing.html#system-testing
)
```
$ bin/rails generate system_test users
      invoke test_unit
      create test/system/users_test.rb
```

### Actual behavior

System test directory and files are not created. 

I debugged in railties code and viewed the guide but was unable to make sense of the behavior.
https://guides.rubyonrails.org/testing.html#system-testing

### System configuration

**Rails version**:  7.2.2.2, 8.0.3, 8.1.0

**Ruby version**:  3.2.5

**OS**: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000 arm64

zzak:
Fixes #56022

When `rails new --skip-test` is called, we put `config.generators.system_tests = nil` in `config/application.rb`:
https://github.com/rails/rails/blob/61b5c7e2953fd0ccfab151b233ce69283c567a34/railties/lib/rails/generators/rails/app/templates/config/application.rb.tt#L36-L39

If the user then tries to run `bin/rails g system_test Whatever` or `bin/rails g integration_test Whatever` it will do nothing.

This change will now raise an exception in that case to give the user some feedback.

---

Originally I was going to leave out the const check because it creates a dependency on `Rails::TestUnit`, but I checked and `rspec-rails` is not using this generator -- they provide their own `bin/rails g rspec:integration` and `bin/rails g rspec:system`.

So the behavior is the same if you are using `rspec-rails` or not, if `rails/test_unit/railtie` is not loaded.

I couldn't find any other uses of the generator in the wilds, and I'm unaware of other test frameworks that could be affected by this change.

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Ensure `Rails::TestUnitRailtie` is defined and
+    `config.generators.system_tests` is set when generating a system test.
+
+    For integration tests we added check for `Rails::TestUnitRailtie`.
+
+    *zzak*
+
 *   Do not clean directories directly under the application root with `Rails::BacktraceCleaner`
 
     Improved `Rails.backtrace_cleaner` so that most paths located directly under the application's root directory are no longer silenced.
--- a/railties/lib/rails/generators/rails/integration_test/integration_test_generator.rb
+++ b/railties/lib/rails/generators/rails/integration_test/integration_test_generator.rb
@@ -3,6 +3,14 @@
 module Rails
   module Generators
     class IntegrationTestGenerator < NamedBase # :nodoc:
+      def check_if_test_unit_loaded!
+        unless defined?(Rails::TestUnitRailtie)
+          raise <<~ERR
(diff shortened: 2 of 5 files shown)


zzak:
I think the docker cache on this PR got corrupted because it was starting from an older sha during the minitest changes. :cry:

PR #56399, opened 2025-12-18
Add minitest-mock as run-time dependency of AS

zzak:
It's used in ActiveSupport::Testing::MethodCallAssertions:
https://github.com/rails/rails/blob/61b5c7e2953fd0ccfab151b233ce69283c567a34/activesupport/lib/active_support/testing/method_call_assertions.rb#L3

This might explain why my builds are failing, but why not all builds are affected :shrug: 
https://buildkite.com/rails/rails/builds/124759#019b3073-e2ce-4883-b4eb-212f6e3dea86/L1482

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -70,6 +70,7 @@ PATH
       json
       logger (>= 1.4.2)
       minitest (>= 5.1)
+      minitest-mock
       securerandom (>= 0.3)
       tzinfo (~> 2.0, >= 2.0.5)
       uri (>= 0.13.1)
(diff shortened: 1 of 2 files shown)


byroot:
Failures look legit.

byroot:
NB: I locked minitest to `< 6` in the Gemfile so that until we sort this stuff out we have a working CI.

If you wish to finish this PR you probably want to rebase and remove the lock.

Earlopain:
The module is nodoc, not intended to be used outside of rails. If mock is only used there for use of testing rails itself, I don't think it's the right move to make this a runtime dependency.

zenspider:
I already have a branch + PR up to add minitest-mock and it was green last time I looked at it.

ETA: https://github.com/rails/rails/pull/56207

zzak:
Yeah, being needed only for one assertion method (`assert_called_with`) makes it questionable as a runtime dep, so I will change it but users will have to update their Gemfile

zzak:
I guess we can combine it into a single PR then, since builds are no longer blocked

PR #56403, opened 2025-12-18
Add Rails.app alias for Rails.application

dhh:
This is especially so when it's followed by a long name, like `credentials`. Now that's `Rails.app.credentials`. Reads a lot more pleasantly too.


justinko:
How do you feel about `Rails.app.creds`?

PR #56404, opened 2025-12-19
Add Rails.app.creds to provide combined credentials lookup in ENV and encrypted file

dhh:
Many applications use a mix of ENV and encrypted keys to store cloud service keys and the like. Depending on the service, it might make sense only to inject the ENV during Docker boot or to keep it as part of the Rails encrypted credentials. And applications might migrate from one form to another.

But whether ENV or encrypted keys are both used or you wish to migrate in the future, you shouldn't have to change your application code.

So these combined credentials are now available via `Rails.app.creds`.

```ruby
Rails.app.creds.require(:db_host) # ENV.fetch("DB_HOST") || Rails.app.credentials.require(:db_host)
Rails.app.creds.require(:aws, :access_key_id) # ENV.fetch("AWS__ACCESS_KEY_ID") || Rails.app.credentials.require(:aws, :access_key_id)
Rails.app.creds.option(:cache_host) # ENV["CACHE_HOST"] || Rails.app.credentials.option(:cache_host)
Rails.app.creds.option(:cache_host, default: "cache-host-1") # ENV["CACHE_HOST"] || Rails.app.credentials.option(:cache_host) || "cache-host-1"
Rails.app.creds.option(:cache_host, default: -> { "cache-host-1" }) # ENV["CACHE_HOST"] || Rails.app.credentials.option(:cache_host) || "cache-host-1"
```

Provided by the new `ActiveSupport::CombinedConfiguration` and `ActiveSupport::EnvConfiguration` classes.

You can also access just envs using the new API via `Rails.app.envs`, which uses the same require/option API. And finally, you can set a custom set of backends for Rails.app.creds by assigning your own combined configuration:

```ruby
Rails.app.creds = ActiveSupport::CombinedConfiguration.new(Rails.app.envs, OnePasswordConfiguration.new)
```


activesupport/lib/active_support/env_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    # Find a upcased and double-underscored-joined string-version of the +keys+ in ENV.
+    #
+    # Examples:
+    #
+    #   config.dig(:db_host) # => ENV["DB_HOST"]
+    #   config.dig(:database, :host) # => ENV["DATABASE__HOST"]
+    def dig(*keys)
+      lookup(Array(keys).collect { |key| envify(key) }.join("__"))
```
byroot:
```suggestion
      lookup(keys.collect { |key| envify(key) }.join("__"))
```

keys is guaranteed to be an array.

activesupport/lib/active_support/env_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    private
+      def lookup(env_key)
+        ENV[env_key]
```
byroot: I've been meaning to implement something like that for a while, but a big reason for it is that accessing env in tight loops is quite costly. 

I think we should consider storing a copy of the ENV (`env.to_h`) and using that.

The downside of course is that env assignment won't be reflected, so maybe that's too annoying for users.

activesupport/lib/active_support/env_configuration.rb
```diff
@@ -0,0 +1,58 @@
...
+    #
+    #   creds.grab(:db_host) # => creds[:db_host]
+    #   creds.grab(:database, :host) # => creds.dig(:database, :host)
+    def grab(*keys)
+      if keys.many?
+        dig(*keys)
+      else
+        self[keys.first]
+      end
+    end
```
byroot:
```suggestion
    def grab(*keys)
      dig(*keys)
    end
```

activesupport/lib/active_support/combined_configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+    #
+    #   creds.grab(:db_host) # => creds[:db_host]
+    #   creds.grab(:database, :host) # => creds.dig(:database, :host)
+    def grab(*keys)
+      if keys.many?
+        dig(*keys)
+      else
+        self[keys.first]
+      end
+    end
```
byroot:
```suggestion
    def grab(*keys)
      dig(*keys)
    end
```

activesupport/lib/active_support/combined_configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+  class CombinedConfiguration
+    def initialize(*configurations)
+      @configurations = configurations
+    end
+
+    # Combined interface for #[] and #dig. Examples:
+    #
+    #   creds.grab(:db_host) # => creds[:db_host]
+    #   creds.grab(:database, :host) # => creds.dig(:database, :host)
+    def grab(*keys)
```
byroot: Shouldn't just this be `dig`? Since the behavior is identical, might as well use a well known name.

activesupport/lib/active_support/env_configuration.rb
byroot: > The downside of course is that env assignment won't be reflected, so maybe that's too annoying for users.

Actually the more I think about it, if you write into your confile file, that doesn't reload your config automatically, so kinda make sense in a way.

We can provide a `reload` method that works for both.

dhh: I like that idea. Think it's totally fine to cache upfront too with a #reload. This is almost always about having these values set at boot time anyway.

activesupport/lib/active_support/env_configuration.rb
dhh: Doh. Yeah, we don't need grab at all.

activesupport/lib/active_support/combined_configuration.rb
dhh: Yes, we should. I was thinking it was either dig for multiple or [] for single, but of course you can just use dig for both.

activesupport/lib/active_support/env_configuration.rb
dhh: Implemented that in [116e36b](https://github.com/rails/rails/pull/56404/commits/116e36b159e7a95fa27411dba00b58f93005459a).

byroot:
Looks good!👍 Le 19 déc. 2025 à 17:38, David Heinemeier Hansson ***@***.***> a écrit :﻿
@dhh commented on this pull request.



In activesupport/lib/active_support/env_configuration.rb:
> +      lookup envify(key)
+    end
+
+    # Find a upcased and double-underscored-joined string-version of the +keys+ in ENV.
+    #
+    # Examples:
+    #
+    #   config.dig(:db_host) # => ENV["DB_HOST"]
+    #   config.dig(:database, :host) # => ENV["DATABASE__HOST"]
+    def dig(*keys)
+      lookup(Array(keys).collect { |key| envify(key) }.join("__"))
+    end
+
+    private
+      def lookup(env_key)
+        ENV[env_key]

Implemented that in 116e36b.

activesupport/lib/active_support/combined_configuration.rb
```diff
@@ -0,0 +1,56 @@
...
+
+module ActiveSupport
+  # Allows for configuration keys to be pulled from multiple backends. Keys are pulled in first-found order from
+  # the configuration backends that the combined configuration has been initialized with.
+  #
+  # This is used by Rails to offer a unified API for fetching credentials from both ENV and the encrypted file.
+  # You can access this through +Rails.app.creds+ within a Rails application.
+  class CombinedConfiguration
+    def initialize(*configurations)
+      @configurations = configurations
```
byroot, line 13: Another idea: it would be nice if there was a way to append/prepend another configuration/secret store.

This would allow to integrate things like https://rubygems.org/gems/ejson-rails in a much cleaner way.

activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -58,6 +58,42 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
+    # Find a upcased and double-underscored-joined string-version of the +keys+ in ENV.
+    # Returns nil if the key isn't found or the value of default when passed If default is
+    # a block, it's called first.
+    #
+    # Examples:
+    #
+    #   config.option(:db_host)                                    # => ENV["DB_HOST"]
+    #   config.option(:database, :host)                            # => ENV["DATABASE__HOST"]
+    #   config.option(:database, :host, default: "missing")        # => ENV.fetch("DATABASE__HOST", "missing")
+    #   config.option(:database, :host, default: -> { "missing" }) # => ENV.fetch("DATABASE__HOST", default.call)
```
byroot, line 87: Why not:

```suggestion
    #   config.option(:database, :host) { "missing" } # => ENV.fetch("DATABASE__HOST", default.call)
```

activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -58,6 +58,42 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
+    def option(*keys, default: nil)
+      if default.respond_to?(:call)
+        dig(*keys) || default.call
+      elsif default
+        dig(*keys) || default
```
byroot: This may be an issue for boolean configs.

activesupport/lib/active_support/encrypted_configuration.rb
dhh: I considered this, as it would match fetch, but never liked the unexplained block. And since we're doing a totally new API here with require/option, I prefer to be explicit about the default. Like we do in other places, like attributes.

activesupport/lib/active_support/encrypted_configuration.rb
dhh: Good point.

dhh: Fixed in 82b6e09c556b83f835e9fc69cbcc4155eb1e66d8.

activesupport/lib/active_support/combined_configuration.rb
dhh: That's a good idea, but I think in most cases, you'd want this inserted after the ENVs lookups. So you can still overwrite through that. So I've added `Rails.app.creds=`, which can take a custom `ActiveSupport::CombinedConfiguration` instance. I've also split out `Rails.app.envs`, so you could do:

```ruby
Rails.app.creds = ActiveSupport::CombinedConfiguration.new \
  Rails.app.envs, EjsonConfiguration.new
```

activesupport/lib/active_support/combined_configuration.rb
```diff
@@ -0,0 +1,56 @@
...
+    #
+    #   option(:db_host)                             # => ENV["DB_HOST"] || Rails.app.credentials.option(:db_host)
+    #   option(:database, :host)                     # => ENV["DATABASE__HOST"] || Rails.app.credentials.option(:database, :host)
+    #   option(:database, :host, default: "missing") # => ENV["DATABASE__HOST"] || Rails.app.credentials.option(:database, :host) || "missing"
+    #   option(:database, :host, default: -> { "missing" }) # => ENV["DATABASE__HOST"] || Rails.app.credentials.option(:database, :host) || "missing"
+    def option(*key, default: nil)
+      @configurations.each do |config|
+        if value = config.option(*key)
+          return value
+        end
```
byroot:
```suggestion
        value = config.option(*key)
        return value unless value.nil?
```

Need to handle `false` config here here.

activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -58,6 +58,42 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
 
+    # Find the referenced key
+    # Raises +KeyError+ if not found.
+    #
+    # Examples:
+    #
+    #   require(:db_host)         # => ENV.fetch("DB_HOST")
+    #   require(:database, :host) # => ENV.fetch("DATABASE__HOST")
+    def require(*key)
+      dig(*key) || raise(KeyError)
```
byroot:
```suggestion
      value = dig(*key)
      raise(KeyError) if value.nil?
      value
```
Here too.

dhh: Ah yes. What an awkward implementation that gives, though. Wish Ruby had an || like operator for this purpose. |? or whatever.

activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -58,6 +59,48 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
+    #   require(:db_host)         # => ENV.fetch("DB_HOST")
+    #   require(:database, :host) # => ENV.fetch("DATABASE__HOST")
+    def require(*key)
+      value = dig(*key)
+
+      if value.defined?
+        value
+      else
+        raise(KeyError)
+      end
```
byroot:
```suggestion
      raise(KeyError) if value.nil?
      value
```

I really don't think a new core ext is warranted here, especially one with a name that matches an existing keyword. A simple guard clause is just as readable IMO.

dhh: I've wanted this for a long time. Just like we have present/blank. Every existential predicate should come in a positive and a negative. This goes right to Ruby's if/unless insight. So this is just riding along with this PR, because it highlighted it.

Also, I really dislike guard statements inside methods that produce multi exit wounds.

dhh: But would entertain a different method name than `defined?` if you have a great one.

dhh: Actually, fair point that we already have Kernel#defined? with different semantics. I'll yank and find a better name before reintroducing in a separate PR (maybe #set?).

maxim:
@dhh If interested, I've been thinking through some of these ideas when designing [this gem](https://github.com/maxim/loadout) including thoughts like "how to deal with inherited configs". For example, it's hard to inherit a required config and then override it as optional, because the required one will error-out immediately. I wrote up some tricks in the gem's README about it.

dhh:
Not sure I follow? The combined configuration that wraps Rails.app.creds will check all backends on #require before throwing an exception. So you should be good with this setup 👍

maxim:
Ah nice. My solution is simple/naive, because it doesn't touch anything in Rails, only adds helpers for the right hand side (an intentional constraint). Either way, wanted to share some ideas around this since this feels somewhat similar.

OrestF:
Looks like one of my gems is going to be deprecated soon  — [RCreds](https://github.com/coaxsoft/r_creds)￼ 🥲. I developed it specifically for the scenario you described, but I appreciate the Rails ecosystem and am going to promote the native approach in my README soon.

p8:
With Rails.application.secrets deprecated and removed in 7.2, could we reuse “secrets” for this instead? It might be confusing with older documentation describing the old secret implementation, but having creds and credentials having different behaviour might cause some confusion as well.

Getting the application secret (which isn’t really a credential?) from “secrets” instead of “creds” seem clearer to me as well.

PR #56405, opened 2025-12-19
zenspider mt6 pt3

zenspider:
- **Add Minitest.load :rails where needed.**
- **Fix ThreadPoolExecutor to work with minitest 5 & 6.**


zzak:
We just wanted to check if the tests still passed before bumping the dep, I think we can keep #56207 as is and just merge it as a single PR.

PR #56407, opened 2025-12-19
Add password manager support for Rails Credentials keys

jordan-brough:
### Motivation / Background

When using per-environment credentials files, I'd like to be able to use a lazily-invoked password manager when:

1. **Editing or viewing credentials** for my production environment, in particular.
2. **Viewing git diffs** with changes to `config/credentials/production.yml.enc`.
3. **Deploying** e.g. via Kamal. (This isn't addressed in the current PR, though)

### Detail

This PR adds `config.credentials.key_command`, similar to the existing `config.credentials.key_path`.

If present, `config.credentials.key_command` is executed as a shell command, e.g.:
```ruby
config.credentials.key_command = "op read op://vault/my_app/master_key"
```

### Additional information

This approach seems convenient and flexible to me, but some alternative options could be:

- Re-use `config.credentials.key_path` by detecting when it's a shell command instead of a path and executing it.
- Have `key_command` be a callable object instead of a shell command.
- Allow `RAILS_<ENVIRONMENT>_MASTER_KEY` env vars, and leave it to the user to set the appropriate environment values.  I think this might end up being eagerly loaded locally though, which I don't love for speed or for production keys.
- Provide explicit integrations with password managers, similar to what Kamal does. E.g.:
    ```ruby
    config.credentials.key_provider = {
      adapter: "1password",
      account: "myaccount",
      vault: "myvault",
      item: "myitem",
      key: "rails_master_key",
    }
    ```

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Add `key_command` option to `EncryptedFile` and `EncryptedConfiguration` to
+    fetch encryption keys from shell commands (e.g., password managers).
+
+    *Jordan Brough*
+
 *   Make flaky parallel tests easier to diagnose by deterministically assigning
     tests to workers.
 
--- a/activesupport/lib/active_support/encrypted_configuration.rb
+++ b/activesupport/lib/active_support/encrypted_configuration.rb
@@ -51,9 +51,9 @@ module ActiveSupport
 
     delegate_missing_to :options
 
-    def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
-      super content_path: config_path, key_path: key_path,
-        env_key: env_key, raise_if_missing_key: raise_if_missing_key
+    def initialize(config_path:, env_key:, key_command: nil, key_path:, raise_if_missing_key:)
+      super content_path: config_path, env_key: env_key, key_command: key_command,
+        key_path: key_path, raise_if_missing_key: raise_if_missing_key
       @config = nil
       @options = nil
     end
--- a/activesupport/lib/active_support/encrypted_file.rb
+++ b/activesupport/lib/active_support/encrypted_file.rb
@@ -26,6 +26,12 @@ module ActiveSupport
       end
     end
 
+    class KeyCommandError < RuntimeError
+      def initialize(key_command:, message:)
+        super "key_command `#{key_command}` #{message}"
+      end
+    end
+
     CIPHER = "aes-128-gcm"
 
     def self.generate_key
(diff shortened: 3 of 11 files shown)


jordan-brough:
I should have made it clearer that the goal of this PR is around editing & diffing credentials securely and easily. Not providing the key in a deployed app.
This PR probably needs some updates to scope the changes more tightly to those things. (But I'd love to hear feedback on the general idea first)

It feels like `rails credentials:edit --environment=production` and `git diff config/credentials/production.yml.enc` should "just work" locally, without storing the master key on disk in plaintext and without having the production key always in memory (e.g. env var).

jordan-brough:
Another option: A config file, that perhaps Kamal could also use:
```yaml
# config/credential_keys.yml
test:
  plaintext: "1234567890asdf1234567890asdf"
development:
  file: config/credentials/development.key
production:
  command: "op read op://Engineering/MyApp/production_master_key"
```

guilleiguaran:
I like the general proposal, but I'll give more of thinking on the proposal vs the other proposed alternatives.

jordan-brough:
Thanks @guilleiguaran.

FYI, I'm assuming the deployment process would also extract the key from my password manager, and set it as `ENV["RAILS_MASTER_KEY"]` in the deployed app, which overrides this new setting so that the deployed app doesn't try to load the value from a password manager.

For that reason it feels like it'd be nice if Kamal could share the config, so that I don't have to configure the same password manager  in multiple places (in `.kamal/secrets` and somewhere for this). So maybe a `config/credential_keys.yml` could be a better option that both could share?

Also I'm kind of assuming people bothering to use a password manager for this would have separate credentials files for each environment. This would add delay and maybe extra steps (e.g. biometric auth when grabbing the key) for development & test environments.

morgoth:
Looks like this could be achieved via newly added https://github.com/rails/rails/pull/56404 by creating your own class and setting it to Rails.app.creds

jordan-brough:
@morgoth that PR from @dhh is great! I'll look through it more later today, but at first glance I think it might be targeting a different issue?

I think that's addressing storing _secrets_ in 1Password (and ENV and any other custom locations), whereas I'm wanting to continue to use `Rails.app.credentials` (and being able to commit and ship the encrypted file), and just having the _master key_ in a password manager. (And having it work for `rails credentials:xxx` and `git diff` and, ideally, Kamal as well).

I'll take a closer look later though.

EDIT: OK I think I misunderstood the PR a bit, and @morgoth is probably right that it could help solve this! Will look more.

PR #56417, opened 2025-12-21
Active storage requirements [skip ci]

tahsin352:
### Motivation / Background

I was playing with Active storage in my freshly setup Macbook Pro. When I made some modifications into the codebase and run the tests `bin/test`, there were lots of errors regarding missing required components, dependencies. I searched into the active storage Readme file and there were no suggestions about active storage requirements. I googled and found in the guide sections regarding the requirements. Though I found I need both libvips and imagemagick into my setup. For contributors, it should be also mentioned in the active storage readme file. And I also updated the installation with one more additional step that I faced required. Hence, I created this Pull Request.

### Detail

This Pull Request updates readme file with active storage requirements section and it also updates installation section with one more additional step that I faced required.

### Checklist

### DIFF
--- a/activestorage/README.md
+++ b/activestorage/README.md
@@ -14,11 +14,22 @@ A key difference to how Active Storage works compared to other attachment soluti
 
 `Blob` models store attachment metadata (filename, content-type, etc.), and their identifier key in the storage service. Blob models do not store the actual binary data. They are intended to be immutable in spirit. One file, one blob. You can associate the same blob with multiple application models as well. And if you want to do transformations of a given `Blob`, the idea is that you'll simply create a new one, rather than attempt to mutate the existing one (though of course you can delete the previous version later if you don't need it).
 
+## Requirements
+
+Various features of Active Storage depend on third-party software which Rails
+will not install, and must be installed separately:
+
+* [libvips](https://github.com/libvips/libvips) v8.6+ and [ImageMagick](https://imagemagick.org/index.php) for image analysis and transformations
+* [ffmpeg](http://ffmpeg.org/) v3.4+ for video previews and ffprobe for video/audio analysis
+* [poppler](https://poppler.freedesktop.org/) or [muPDF](https://mupdf.com/) for PDF previews
+
 ## Installation
 
-Run `bin/rails active_storage:install` to copy over active_storage migrations.
+* Run `bin/rails active_storage:install` to copy over active_storage migrations.
 
 NOTE: If the task cannot be found, verify that `require "active_storage/engine"` is present in `config/application.rb`.
+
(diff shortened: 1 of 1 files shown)


activestorage/README.md
```diff
@@ -14,12 +14,23 @@ A key difference to how Active Storage works compared to other attachment soluti
 
 `Blob` models store attachment metadata (filename, content-type, etc.), and their identifier key in the storage service. Blob models do not store the actual binary data. They are intended to be immutable in spirit. One file, one blob. You can associate the same blob with multiple application models as well. And if you want to do transformations of a given `Blob`, the idea is that you'll simply create a new one, rather than attempt to mutate the existing one (though of course you can delete the previous version later if you don't need it).
 
+## Requirements
+
+Various features of Active Storage depend on third-party software which Rails
+will not install, and must be installed separately:
+
+* [libvips](https://github.com/libvips/libvips) v8.6+ and [ImageMagick](https://imagemagick.org/index.php) for image analysis and transformations
```
zzak: It's already mentioned earlier that minimagick or vips are used for transformations.

I'm not sure ffmpeg or poppler are common enough to be declared at the top of the guide, but maybe you can fit it in (briefly) below that sentance above?

tahsin352: Thanks for your feedback. Yes, I know that ffmpeg, ImageMagick, poppler are common. As I mentioned into the description, I faced the errors in a freshly setup Mac. But I it would be better to mention the requirements for Active Storage in both readme and guides.

zzak: Right, I'm saying you can add something here but maybe try to keep it as concise as possibe:
https://github.com/rails/rails/blob/92481359a7f74261059a292a7cea61e68144b57e/activestorage/README.md?plain=1#L7

This file is rendered here under the top-level module:
https://edgeapi.rubyonrails.org/classes/ActiveStorage.html#module-ActiveStorage-label-Active+Storage

tahsin352: Thanks for your valuable suggestions. I have made the requirement more concise.

p8:
Instead of duplicating information from the guides, I think we should keep these README's short and link to the guides.
Similar to:
https://github.com/rails/rails/blob/main/actioncable/README.md
https://github.com/rails/rails/blob/main/actiontext/README.md
https://github.com/rails/rails/blob/main/actionview/README.rdoc

skipkayhil:
It sounds like the additions are for people developing Active Storage, which I don't think we want to add to the Active Storage README (which is more for application developers _using_ Active Storage).

Everything added is covered in https://guides.rubyonrails.org/development_dependencies_install.html. Is there a way we could make that more prominent for people trying to work on Rails?

Issue #56235, opened 2025-11-26
ActionView::TestCase#rendered behavior conflicts with documentation


drjayvee:
# Issue

The documentation and test for `ActionView::TestCase#rendered` are in conflict with the behavior.

The [documentation for `rendered`](https://github.com/rails/rails/blob/a9a12a148b7782ef8d22b73c0d948a06e4a97166/actionview/lib/action_view/test_case.rb#L65) reads (emphasis mine)

> Returns the content rendered by the *last* +render+ call.

The [test name](https://github.com/rails/rails/blob/5c0b7496ab32c25c80f6d1bdc8b32ec6f75ce1e4/actionview/test/template/test_case_test.rb#L398) is

> #rendered resets after each render

However, it actually asserts that `#rendered` accumulates output of consecutive `render` calls. Here's a slightly simplified version of the test:

```ruby
render "developers/developer", developer: DeveloperStruct.new("first")

assert_includes rendered, "first"
assert_not_includes rendered, "second"

render "developers/developer", developer: DeveloperStruct.new("second")

assert_includes rendered, "first"
assert_includes rendered, "second"
```

### Expected behavior

Based on the documentation and test name, I would expect the actual behavior to be

```ruby
render "developers/developer", developer: DeveloperStruct.new("first")

assert_includes rendered, "first"
assert_not_includes rendered, "second"

render "developers/developer", developer: DeveloperStruct.new("second")

assert_not_includes rendered, "first"
assert_includes rendered, "second"
```

# History

The behavior of `ActionView::TestCase#rendered` was changed in #51093. (ping @seanpdoyle)

`#render` [appends to](https://github.com/rails/rails/blob/a9a12a148b7782ef8d22b73c0d948a06e4a97166/actionview/lib/action_view/test_case.rb#L287) to `@rendered`:

```ruby
@rendered << output = view.render(options, local_assigns, &block)
```

# Solutions
We could revert the behavior to match the documentation. However, that will likely break some tests ([again](https://github.com/rails/rails/pull/49856#issuecomment-1945039015)).

Therefore, it may be better to adjust the documentation (and test name) to match the actual behavior.

A third alternative would be to add a parameter `reset:` to `render` or `rendered`, or to add a `reset_rendered` method. (None of those look appealing to me.) Update: users can use `rendered.clear` to get the behavior.

I'd be more than happy to create a PR, but I thought it would be best to describe the problem here and ask what the best solution should be. I'd personally prefer restoring the original behavior.

[...]

PR #51093, opened 2024-02-15
Action View Test Case `rendered` memoization

seanpdoyle:
### Motivation / Background

The introduction of memoization as an optimization posed a backwards incompatible change to View tests that call `render` multiple times.

### Detail

Follow-up to [49856][]
Follow-up to [49194][]

This commit changes the `@rendered` instance variable from a `String` to an instance of the `RenderedViewContent` specialized `String` subclass.
    
The end result is that there is no memoization to reset, and the memoization optimization side-effect is preserved after rendering for test cases where `rendered` (or parser methods like `rendered.html`) might be invoked more than once.

[49856]: https://github.com/rails/rails/pull/49856#issuecomment-1945039015
[49194]: https://github.com/rails/rails/pull/49194/files#diff-ce84a807f3491121a5230d37bd40454bb1407fcca71179e1a2fa76d4c0ddfa2aR293

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56422, opened 2025-12-22
ActionView::TestCase#render resets rendered

drjayvee:
### Motivation / Background

See #56235 for details.

### Detail

`ActionView::TestCase#render` resets `rendered` by calling `@rendered.clear`.

This behavior was broken when memoization was added in #51093.

### Additional information

One alternative is to create a new `@rendered` object.

After all, users may provide their own class by setting the `content_class` class attribute. The only contract / dependency is that the object `responds_to :<<`, so `:clear` might not be implemented.

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Fix `ActionView::TestCase#render` to reset `rendered`.
+    The behavior was changed when memoization was added in #51093. Now it once again conforms to the documentation.
+
+    *Jeroen Versteeg*
+
 *   Fix `FormBuilder#to_partial_path` returning `nil` for subclasses whose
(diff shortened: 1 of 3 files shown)


drjayvee:
@seanpdoyle, sorry again for the nagging, but could you please take a look at this PR and the related bug?

rafaelfranca (approved):

drjayvee:
Thank you Rafael for picking this up and fixing the issue with sub-rendering.

I'll have a look at the failing test. My hunch (without looking) is that it depends on the previous behavior (i.e. `render` does *not* flush).

drjayvee:
I can't reproduce the error locally. Both `bundle exec ruby -Itest test/template/test_case_test.rb` and `rake test` (to run all tests) pass.

PR #56428, opened 2025-12-22
Available to Set Credentials via CLI Directly Using `rails credentials:set path=value`

berniechiu:
### Motivation / Background

This Pull Request has been created because, for CLI commands, it is common for servers or DevOps to manage keys using one-liner assignments like `$ heroku config:set KEY:VALUE`. This can be done without opening an editor, simply through straightforward CLI command instructions. REF: https://devcenter.heroku.com/articles/config-vars#using-the-heroku-cli

### Detail

This Pull Request changes:
- `railties/lib/rails/commands/credentials/credentials_command.rb`
- `railties/test/commands/credentials_test.rb`

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Add `bin/rails credentials:set` commands to set encrypted credentials from the command line,
+    making it easier to update credentials without opening an editor.
+
+    *berniechiu*
+
 *   Add `Rails.app.creds` to provide combined access to credentials stored in either ENV or the encrypted credentials file.
     Provides a new require/option API for accessing these values. Examples:
 
--- a/railties/lib/rails/commands/credentials/credentials_command.rb
+++ b/railties/lib/rails/commands/credentials/credentials_command.rb
@@ -77,6 +77,38 @@ module Rails
         end
       end
 
+      desc "set PATH=VALUE", "Set a value in the decrypted credentials"
+      def set(*pairs)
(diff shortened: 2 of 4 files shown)


brunoprietog:
I'd make the API more similar to `bin/rails credentials:set database.url=value`, using `=`, Or even as a separate argument, but that's just my opinion of course. I'm thinking about it in case you have `:` as part of the value, and because handling those things comes free with argument handling, quotes, escape characters, right? What do you think?

berniechiu:
> I'd make the API more similar to `bin/rails credentials:set database.url=value`, using `=`, Or even as a separate argument, but that's just my opinion of course. I'm thinking about it in case you have `:` as part of the value, and because handling those things comes free with argument handling, quotes, escape characters, right? What do you think?

Ah, yes! Actually, the code logic is set by `=`. Just realized the typos in some sentences and the PR title. Thanks.

byroot:
I never really used credentials, but aren't you able to store comments in them? This feature would strip all your comments, wouldn't it?

berniechiu:
> I never really used credentials, but aren't you able to store comments in them? This feature would strip all your comments, wouldn't it?

That could be a concern, will revisit this soon. Thanks for pointing out!

brunoprietog:
Now this is easy!

https://kddnewton.com/2025/12/25/psych-pure.html

byroot:
Yes, but I don't think such feature is worth adding a new dependency.

PR #56429, opened 2025-12-22
Fixes for default handling in CombinedConfiguration

jordan-brough:
### Motivation / Background

A few fixes for the recently-added `CombinedConfiguration` from https://github.com/rails/rails/pull/56404

### Detail

1. Fix `default` in `env_configuration.rb` to be invoked lazily, like `encrypted_configuration.rb` does.
2. Fix `default` in `env_configuration.rb` to return `false` for `default: false`. Previously it would return `nil`.
3. Fix `default` in `encrypted_configuration.rb` when the default callback returns `false`. Previously it would return the callback itself in that case.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/env_configuration.rb
+++ b/activesupport/lib/active_support/env_configuration.rb
@@ -39,14 +39,12 @@ module ActiveSupport
     #   option(:db_host)                                    # => ENV["DB_HOST"]
     #   option(:database, :host)                            # => ENV["DATABASE__HOST"]
     #   option(:database, :host, default: "missing")        # => ENV.fetch("DATABASE__HOST", "missing")
-    #   option(:database, :host, default: -> { "missing" }) # => ENV.fetch("DATABASE__HOST", default.call)
+    #   option(:database, :host, default: -> { "missing" }) # => ENV.fetch("DATABASE__HOST") { default.call }
     def option(*key, default: nil)
       if default.respond_to?(:call)
-        @envs.fetch envify(*key), default.call
-      elsif default
-        @envs.fetch envify(*key), default
+        @envs.fetch(envify(*key)) { default.call }
       else
-        @envs[envify(*key)]
+        @envs.fetch envify(*key), default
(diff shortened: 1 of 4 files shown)


activesupport/lib/active_support/env_configuration.rb
```diff
@@ -39,14 +39,12 @@ def require(*key)
...
     #   option(:database, :host)                            # => ENV["DATABASE__HOST"]
     #   option(:database, :host, default: "missing")        # => ENV.fetch("DATABASE__HOST", "missing")
-    #   option(:database, :host, default: -> { "missing" }) # => ENV.fetch("DATABASE__HOST", default.call)
+    #   option(:database, :host, default: -> { "missing" }) # => ENV.fetch("DATABASE__HOST") { default.call }
     def option(*key, default: nil)
       if default.respond_to?(:call)
-        @envs.fetch envify(*key), default.call
-      elsif default
-        @envs.fetch envify(*key), default
+        @envs.fetch(envify(*key)) { default.call }
```
byroot, line 45: this doesn't seem correct to me. We're in the `else` branch of `if default.respond_to?(:call)`

jordan-brough: The line you commented on is not the `else` branch.
Did you possibly misread the diff? We went from 3 branches to 2 branches, so the diff is a bit confusing.

jordan-brough: We could also rewrite like this:
```ruby
@envs.fetch(envify(*key)) do
  default.respond_to?(:call) ? default.call : default
end
```
That might be clearer?
Also means we don't even have to check `respond_to?` if the key is present.

byroot: What the hell, is there a bug in GitHub diff rendering?

byroot: 🤦 nevermind.

PR #56430, opened 2025-12-22
Fix & standardize comments for CombinedConfiguration

jordan-brough:
### Motivation / Background

Some minor comment updates related to https://github.com/rails/rails/pull/56404

1. Fix comments in `encrypted_configuration.rb` that were referencing `ENV`.
2. Standardize comments across the three different classes. I tried to take what seemed like the best pieces from each and standardize them.
3. Note in comments that `nil` values are considered "missing". ~Though I'm not 100% sure if that's intentional.~
4. Use scenarios for comment examples

### Checklist


activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -58,13 +58,13 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
-    # Raises +KeyError+ if not found.
+    # Find singular or nested keys.
+    # Raises +KeyError+ if not found or nil.
     #
     # Examples:
     #
-    #   require(:db_host)         # => ENV.fetch("DB_HOST")
-    #   require(:database, :host) # => ENV.fetch("DATABASE__HOST")
+    #   require(:db_host)         # => dig(:db_host) || raise(KeyError)
+    #   require(:database, :host) # => dig(:database, :host) || raise(KeyError)
```
jordan-brough: I think the comments in here were a copy-paste leftover from `env_configuration.rb`?

I dunno if what I put here is that helpful, and it isn't 100% correct. E.g. the comment doesn't accurately show what happens when the value is `false`.
To be fully accurate we'd have to duplicate the method code in the comment.

I wonder if it might be better for the comments here and in the other two files to instead lay out a mini scenario and then show the results?

E.g. maybe something like:
```ruby
# Examples:
#
# db_host: "db.example.com"
# database:
#   host: "db.example.com"
#
# require(:db_host)         # => "db.example.com"
# require(:database, :host) # => "db.example.com"
# require(:missing)         # => KeyError
```
?

activesupport/lib/active_support/combined_configuration.rb
```diff
@@ -13,7 +13,8 @@ def initialize(*configurations)
       @configurations = configurations
     end
 
-    # Find singular or nested keys across all backends. If no backend holds the key, it raises +KeyError+.
+    # Find singular or nested keys across all backends.
+    # Raises +KeyError+ if no backend holds the key or if the value is nil.
```
jordan-brough, line 17: Behavior question RE `if the value is nil`:

Currently if any Configuration has an _explicit_ nil value, it'll be skipped and we'll move on to the next Configuration looking for the key.  For `EnvConfiguration` you can't have explicit nils since you can't have explicit nulls in the environment, only empty strings.
But for EncryptedConfiguration or other backends you can have explicit nulls.

E.g. for EncryptedConfiguration:
```ruby
# some_config:
#   some_key: null
#
Rails.app.creds(:some_config, :some_key)
```
will skip the explicit `null` and move on to looking for the value in the next backend, and falling back to the `default` if needed.
Is that what we want, or should explicit `null`s be respected if found?

dhh: I don't think nil should be respected, no. We should search through all the configurations for a not-nil value. If we can't find one, we raise MissingKey.

activesupport/lib/active_support/encrypted_configuration.rb
dhh: Yes, that's better.

jordan-brough: @dhh I've updated to use scenarios for the comments. How does that look?

activesupport/lib/active_support/combined_configuration.rb
jordan-brough: @dhh that seems reasonable 👍 

Just to check: What do you think of `blank?` values? (i.e. empty strings or whitespace strings)

`EncryptedConfiguration` has had:
* `my_config.foo!`,  which raises for **both** blank and nil values.
* `my_config.fetch(:foo)` which doesn't raise for blank or for `nil` values.

Doesn't seem like this new stuff needs to behave exactly the same as either of those, but I thought I'd check.

dhh: When you do #require, we should blow up if nil or blank 👍

jordan-brough: @dhh I added a commit to check `blank?` instead of `nil?`.
Lmk if I should move that into a separate PR or anything.

jordan-brough: @dhh I removed the `blank?` change so that this is again a comments-only change, reflecting the current state of the code.

Hopefully that makes this easier to merge?

I also moved the `blank?` change into a [separate commit](https://github.com/rails/rails/commit/d62e49a28651b07c894e70d1553262055bf314e3) that I can turn in to a separate PR after we merge this, if that sounds OK.

jordan-brough:
cc @byroot 🙏 since you approved [another CombinedConfiguration-related PR](https://github.com/rails/rails/pull/56429) of mine 🙏

activesupport/lib/active_support/combined_configuration.rb
jordan-brough: For the record, this is the follow-up PR for blank-handling: https://github.com/rails/rails/pull/56818

PR #56431, opened 2025-12-22
Add "Missing key: ..." to `KeyError` message in EncryptedConfiguration

jordan-brough:
Minor update for recent `CombinedConfiguration ` PR https://github.com/rails/rails/pull/56404.

This is helpful info for the error message.
And this is what CombinedConfiguration does for `KeyError` already.

### DIFF
--- a/activesupport/lib/active_support/encrypted_configuration.rb
+++ b/activesupport/lib/active_support/encrypted_configuration.rb
@@ -71,7 +71,7 @@ module ActiveSupport
       if !value.nil?
         value
       else
-        raise(KeyError)
+        raise KeyError, "Missing key: #{key.inspect}"
       end
     end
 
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -71,7 +71,7 @@ def require(*key)
       if !value.nil?
         value
       else
-        raise(KeyError)
+        raise KeyError, "Missing key: #{key.inspect}"
```
jordan-brough, line 74: Same as what `CombinedConfiguration` does:
https://github.com/rails/rails/blob/a9ee16f0b94fe53bd3b715aa4e6e3f63e69a215a/activesupport/lib/active_support/combined_configuration.rb#L28

PR #56433, opened 2025-12-22
don't create unused routes in authentication generator

gregmolnar:
### Motivation / Background
Not all resource routes are used in the authentication generator's controllers, so it would be better to not even create those routes.

### Checklist


skipkayhil:
FYI this was previously rejected: https://github.com/rails/rails/pull/52369#discussion_r1699079086

byroot:
Yes, but this keeps coming up, and it's not a big deal, I'd rather get it solved.

PR #56434, opened 2025-12-23
[8-1-stable] Minitest 6 support

zzak:
Backports #56207, #56385, and #56202 (sans c5c678a67a09b1b83f0f7616290512b48bf2d071) to `8-1-stable` branch.

### DIFF
--- a/activesupport/lib/active_support/testing/parallelization/worker.rb
+++ b/activesupport/lib/active_support/testing/parallelization/worker.rb
@@ -46,8 +46,10 @@ module ActiveSupport
 
           set_process_title("#{klass}##{method}")
 
-          result = klass.with_info_handler reporter do
-            Minitest.run_one_method(klass, method)
+          result = if Minitest.respond_to? :run_one_method then
+            Minitest.run_one_method klass, method
+          else
+            klass.new(method).run
           end
 
           safe_record(reporter, result)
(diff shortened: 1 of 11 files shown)


activesupport/lib/active_support/testing/parallelization/worker.rb
```diff
@@ -46,8 +46,10 @@ def perform_job(job)
 
           set_process_title("#{klass}##{method}")
 
-          result = klass.with_info_handler reporter do
```
gstokkink-bex, line 49: @zzak can it be that you accidentally copied the updated code from the [thread pool executor](https://github.com/rails/rails/pull/56207/changes#diff-71d234acea2ee9c358f05f60594fe6b21892c1d97257f30af2667a556faea2c1R42) here, instead of the updated code from the [worker](https://github.com/rails/rails/pull/56202/changes#diff-1da4187009a2235b3dc012d208c7151864a6aa4dc53986bd152fc831428f20cbR51)? The lack of `with_info_handler` wrapping causes some minitest-based gems to break, like [minitest-hooks](https://github.com/jeremyevans/minitest-hooks/issues/31).

zzak: Gah! Sorry, that is a good catch. We probably should fix that.

gstokkink-bex: Yeah, I think so 😋 I will use a monkeypatch in the meantime.

PR #56436, opened 2025-12-23
[7-2-stable] Minitest 6 support

zzak:
Backports https://github.com/rails/rails/pull/56207, https://github.com/rails/rails/pull/56385, and https://github.com/rails/rails/pull/56202 (sans https://github.com/rails/rails/commit/c5c678a67a09b1b83f0f7616290512b48bf2d071) to 7-2-stable branch.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,14 @@
 source "https://rubygems.org"
 gemspec
 
-gem "minitest"
+# Minitest 6 requires Ruby 3.2 or higher while Rails 7.2 supports Ruby 3.1
+# https://github.com/minitest/minitest/commit/8a50ebfee5d17dc231e5fb87bf936bdf250429a1
+if RUBY_VERSION < "3.2"
+  gem "minitest", "< 6.0"
+else
+  gem "minitest", "~> 6.0"
+  gem "minitest-mock"
+end
 
 # We need a newish Rake since Active Job sets its test tasks' descriptions.
 gem "rake", ">= 13"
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -3,7 +3,8 @@
 source "https://rubygems.org"
 gemspec
 
-gem "minitest"
+gem "minitest", "~> 6.0"
```
yahonda: It should not work with Ruby 3.1.
Minitest 6 requires Ruby 3.2 or higher while Rails 7.2 supports Ruby 3.1.
https://github.com/minitest/minitest/commit/8a50ebfee5d17dc231e5fb87bf936bdf250429a1

yahonda: https://buildkite.com/rails/rails/builds/124977#019b49aa-2ea8-4d97-b3a0-5e0d789a1e89/L2059

zzak: @yahonda The tests should pass with either version, can we put a guard in the Gemfile for 3.1?

```ruby
if RUBY_VERSION >= "3.2"
  gem "minitest", "~> 6.0"
  gem "minitest-mock"
else
  gem "minitest", "< 6.0"
end
```

byroot: We can , but at the end of the day, there's only one Gemfile.lock.

7.2 is in security issue only I think we can stick with minitest 5 there.

yahonda:
> 7.2 is in security issue only I think we can stick with minitest 5 there.

Opened https://github.com/rails/rails/pull/56442 to pin the Minutest version on 7-2-stable branch.

PR #56442, opened 2025-12-23
Pin Minitest to version 5 on 7-2-stable

yahonda:
### Motivation / Background

Since Minitest 6 requires Ruby 3.2 while Rails 7.2 supports Ruby 3.1, pin Minitest to version 5 on the 7-2-stable branch.

Rails 7.2.x is under security maintenance only, so this approach is reasonable.

### Detail
None

### Additional information

https://github.com/minitest/minitest/commit/8a50ebfee5d17dc231e5fb87bf936bdf250429a1 https://github.com/rails/rails/blob/a24be8d963d2d43c295c89c997c98508ab6237bf/rails.gemspec#L12 https://rubyonrails.org/maintenance

### Checklist

### DIFF
--- a/activesupport/activesupport.gemspec
+++ b/activesupport/activesupport.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_dependency "tzinfo",          "~> 2.0", ">= 2.0.5"
   s.add_dependency "concurrent-ruby", "~> 1.0", ">= 1.3.1"
   s.add_dependency "connection_pool", ">= 2.2.5"
-  s.add_dependency "minitest",        ">= 5.1"
+  s.add_dependency "minitest",        ">= 5.1", "< 6"
   s.add_dependency "base64"
   s.add_dependency "drb"
   s.add_dependency "bigdecimal"


byroot:
It's already locked in the Gemfile.

yahonda:
Maybe I've overlooked something, As far as I understand, Gemfile at the 7-2-stable branch does not lock the Minitest version (yet).
https://github.com/rails/rails/blob/a24be8d963d2d43c295c89c997c98508ab6237bf/Gemfile#L6

Even if we lock the Minitest version in Gemfile, this would satisfy Rails framework testing, it does not lock the Minitest version for Rails 7.2.x application since https://github.com/rails/rails/blob/a24be8d963d2d43c295c89c997c98508ab6237bf/activesupport/activesupport.gemspec#L41 allows Minitest 6.

byroot:
> Maybe I've overlooked something, As far as I understand, Gemfile at the 7-2-stable branch does not lock the Minitest version (yet).

My bad, I thought I backported that to 7.2, seems like I didn't.




> it does not lock the Minitest version for Rails 7.2.x application since

Yeah, but that's the thing, I'm not too sure it's useful. We're not supposed to cut any new feature release of 7.2, only security releases that won't contain that change anyway. So the only thing we need is whatever we need for a green CI.

But I guess your PR acheive that too so feel free to merge.

yahonda:
> only security releases that won't contain that change anyway.

Ah, Security releases will be based on the v7.2.3, not based on 7-2-stable.

yahonda:
Let's keep the 7-2-stable CI green. I'm merging it.

Abdullah-l:
So rails 8 is required to use minitest 6?

byroot:
No.

tashoaib:
When do you expect it be on a new patch version as it is a breaking change?

byroot:
There is no breaking change here.

byroot:
If you are talking about 7.2, it won't be released because it's in a security release only mode. If you are talking about 8.x, we don't know, releases aren't scheduled in advance. If you need a fix urgently, just point your Gemfile to the corresponding stable branch.

PR #56443, opened 2025-12-23
Add MaximumSideEffectsReached to TestAdapter

mikeygough:
### Motivation / Background

Working inside a large codebase I've noticed many tests which perform _all_ enqueued jobs:

```ruby
perform_enqueued_jobs do
  ...
end
```

For many tests performing all the enqueued jobs is overkill. For instance, I've come across tests which unnecessarily run an ElasticSearchIndexingJob along with the actual job being tested. This ends up slowing down the test suite with unrelated job runs.

This PR adds functionality allowing users to set a maximum number of 'side effects'. The maximum can be set locally as an argument inside `perform_enqueued_jobs`, for example:

```ruby
# will raise ActiveJob::QueueAdapters::TestAdapter::MaximumSideEffectsReached
perform_enqueued_jobs(maximum_side_effects: 1) do
  assert_raises ActiveJob::QueueAdapters::TestAdapter::MaximumSideEffectsReached do
    HelloJob.perform_later
    GoodbyeJob.perform_later
  end
end
```

Or, the maximum number can be set globally for all tests, for example:

```ruby
# will raise ActiveJob::QueueAdapters::TestAdapter::MaximumSideEffectsReached
ActiveJob::QueueAdapters::TestAdapter.maximum_side_effects = 1
  perform_enqueued_jobs do
    assert_raises ActiveJob::QueueAdapters::TestAdapter::MaximumSideEffectsReached do
      HelloJob.perform_later
      LoggingJob.perform_later
    end
end
```

When this threshold is breached, a `MaximumSideEffectsReached` error will be raised, encouraging users to specify which jobs to perform with the `only:` argument available to `perform_enqueued_jobs`:

"Maximum side effects threshold of #{threshold} reached. #{performed_jobs_count} jobs have already been performed. Consider using perform_enqueued_jobs with `only:` or `except:` or increase the threshold."

### Detail

This Pull Request adds an optional class attribute to `TestAdapter` for global configuration and an optional argument to `perform_enqueued_jobs` for local configuration.

It then implements a counter to check the number of jobs performed against this threshold both on a block and block-less call to `perform_enqueued_jobs`.

I've added test coverage for each code path but welcome suggestions.

### Additional information

In reworking tests on a large codebase, specifying which job to perform in calls to `performed_enqueued_jobs` shaved several seconds off individual tests. My hope is that by setting a threshold value for side effects other developers can identify unrelated jobs being performed in their tests and speed up their tests : )

### Checklist

### DIFF
--- a/activejob/lib/active_job/queue_adapters/test_adapter.rb
+++ b/activejob/lib/active_job/queue_adapters/test_adapter.rb
@@ -12,7 +12,10 @@ module ActiveJob
     #
     #   Rails.application.config.active_job.queue_adapter = :test
     class TestAdapter < AbstractAdapter
-      attr_accessor(:perform_enqueued_jobs, :perform_enqueued_at_jobs, :filter, :reject, :queue, :at, :stopping)
+      class MaximumSideEffectsReached < StandardError; end
+
+      class_attribute :maximum_side_effects, default: nil
+      attr_accessor(:perform_enqueued_jobs, :perform_enqueued_at_jobs, :filter, :reject, :queue, :at, :stopping, :maximum_side_effects_override)
       attr_writer(:enqueued_jobs, :performed_jobs)
 
       # Provides a store of all the enqueued jobs with the TestAdapter so you can check them.
@@ -23,6 +26,25 @@ module ActiveJob
       # Provides a store of all the performed jobs with the TestAdapter so you can check them.
       def performed_jobs
         @performed_jobs ||= []
+      end
+
+      # Provides a count of all the performed jobs to check if the maximum side effects threshold has been reached.
+      def performed_jobs_count
+        @performed_jobs_count ||= 0
+      end
+
+      # Tracks the execution of a job and raises an error if the maximum side effects threshold has been reached.
+      def track_job_execution!
+        check_side_effects_threshold!
+        increment_performed_jobs_count!
+      end
+
+      def increment_performed_jobs_count!
+        @performed_jobs_count = performed_jobs_count + 1
+      end
+
+      def reset_performed_jobs_count!
+        @performed_jobs_count = 0
       end
 
       def enqueue(job) # :nodoc:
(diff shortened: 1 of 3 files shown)


bensheldon:
I'm not sure I understand this interface. A "side effect" is a very generic phrase. I might be more amenable if it was more concretely described for what it is intended to track/assert.

mikeygough:
> I'm not sure I understand this interface. A "side effect" is a very generic phrase. I might be more amenable if it was more concretely described for what it is intended to track/assert.

In this case, side effects are performed jobs. Coming from the situation where a specific job's test is calling `perform_enqueued_jobs` which performs all enqueued jobs, most of which are not needed for the test.

I could make this more explicit by calling it something like job count threshold...

`MaximumSideEffectsReached` => `MaximumJobCountThresholdReached`

`maximum_side_effects_override` => `maximum_job_count_threshold_override`

`check_side_effects_threshold` => `check_maximum_job_count_threshold`

... How is this @bensheldon ? Naming is hard and I'm open to suggestions 😅

bensheldon:
I think it's possible to do this:

```ruby
test "enqueue, assert, and run jobs" do
  perform_enqueued_jobs do
    assert_enqueued_jobs 2 do
      2.times { MyJob.perform_later }
    end
  end
end
```

...so it seems like the proposal would be to allow assert_enqueued_jobs to handle more numerical comparisons (less than, etc.) Are there other existing Rails test helpers that allow numerical comparisons like that?

Edouard-chin:
@bensheldon Thanks for the suggestion on wrapping `perform_enqueued_jobs` / `assert_enqueued_jobs`. I guess this would work if we created a cop to make sure all `perform_enqueued_jobs` assertion are paired with an `assert_enqueued_jobs` and that their arguments matches (e.g. `assert_enqueued_jobs(2, only: MyJob)` has to match with `perform_enqueued_jobs(only: MyJob)` otherwise nothing blocks performing *all* other jobs).

The reason we need this in our monolith is because we have many code that looks like

```ruby
  perform_enqueued_jobs do
    Payment.process
  end
```

A oneliner may ends up enqueueing dozens of jobs that are not required to be performed for the intent of the test, resulting in a test that takes seconds to run.

The idea was to make sure developers are aware of those "side effects" (which I agree isn't the best way to describe this, sorry for the suggestion @mikeygough !) and have their test fail loudly when too many jobs get performed, so that they can modify the call to `perform_enqueued_jobs` to **only** perform the jobs needed by their test.

-----

I may be wrong but I think this could be helpful to other applications and the reason I thought about adding this in the Active Job test helper rather than fixing it only for our app with a monkeypatch or a Rubocop cop.

PR #56445, opened 2025-12-24
Allow numericality :in to accept proc or symbol returning a Range

opsys-saito:
## Motivation / Background

The `:in` option of `numericality` currently only accepts a static `Range`,
while other numericality options such as `:greater_than` and
`:less_than_or_equal_to` already support procs and symbols.

In practice, it is common for the valid range of a numeric attribute to depend
on other attributes of the same record. For example, a `Product` model may
define a minimum and maximum allowed price:

```ruby
  # columns: price, min_price, max_price
end
```

Validating that `price` falls between `min_price` and `max_price` currently
requires a custom `validate` method, even though the intent fits naturally
within `numericality`.

Allowing `:in` to accept a proc or symbol that resolves to a `Range` enables
this use case to be expressed declaratively:

```ruby
validates :price, numericality: { in: ->(p) { p.min_price..p.max_price } }
```

This change brings the `:in` option in line with other numericality options and
with validators such as `inclusion` and `exclusion`, while remaining fully
backward compatible.

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -137,4 +137,9 @@
 
     *Sean Doyle*
 
+
+*   Allow `numericality :in` to accept a proc or symbol.
+
+    *opsys-saito*
+
 Please check [8-1-stable](https://github.com/rails/rails/blob/8-1-stable/activemodel/CHANGELOG.md) for previous changes.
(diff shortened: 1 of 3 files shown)


fatkodima:
Thanks for the PR! There was a similar PR opened before https://github.com/rails/rails/pull/46262 and we decided to proceed with it.

PR #56455, opened 2025-12-24
Allow Rails.app.creds to access .env values in dev

dhh:
In addition to ENV and the encrypted file, let's allow `Rails.app.creds` to also access values from .env files in development mode.

This will now be done in the order of 1) ENV, 2) .env, 3) encrypted credentials.

Values can be both interpolated, like:

```
DB_HOST=localhost
DB_PORT=5432
DATABASE_URL=postgres://${DB_HOST}:${DB_PORT}/app 
```

...and use command execution, so you can use third-party credential providers like 1password:

```
DB_HOST=$(op read op://Vault/item/value --account=MyAccount)
```

This adds `ActiveSupport::DotEnvConfiguration` with the same API as `ActiveSupport::EnvConfiguration`.

### DIFF
--- a/activesupport/lib/active_support/dot_env_configuration.rb
+++ b/activesupport/lib/active_support/dot_env_configuration.rb
@@ -15,6 +15,11 @@ module ActiveSupport
   # - Empty lines (ignored)
   # - Quoted values (single or double quotes)
   # - Variable interpolation with ${VAR} syntax
+  # - Command execution with $(command) syntax
+  #
+  # The command execution allows for easy integration with third-party credential providers, like 1password:
+  #
+  #   DB_HOST=$(op read op://Vault/item/value --account=MyAccount)
   #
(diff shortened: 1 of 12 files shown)


activesupport/lib/active_support/dot_env_configuration.rb
```diff
@@ -0,0 +1,86 @@
...
+        if File.exist?(@path.to_s)
+          envs = {}
+
+          File.foreach(@path) do |line|
+            line = line.strip
+
+            next if line.empty? || line.start_with?("#")
+
+            # Match KEY=value pattern
+            if line =~ /\A([A-Za-z_][A-Za-z0-9_]*)=(.*)\z/
```
Roriz: Should we support spaces around the = sign? It is common for other libraries to support this syntax:
```suggestion
            if line =~ /\A([A-Za-z_][A-Za-z0-9_]*)\s*=\s*(.*)\z/
```

PR #56459, opened 2025-12-25
Fix connection error handling in create_or_find_by methods

afurm:
## Problem

When `create_or_find_by` or `create_or_find_by!` catches `ActiveRecord::RecordNotUnique` and attempts to check if the transaction is open using `connection.transaction_open?`, if this method raises an exception (e.g., due to a closed connection or connection in a bad state), the exception was not being handled and would propagate. This caused the method to fail even though the record exists in the database.

### Example Scenario

```ruby
# Existing record
subscriber = Subscriber.create!(nick: "alice")

# Connection is in a bad state
connection.transaction_open? # raises "Connection error"

# This would fail with an unhandled exception
Subscriber.create_or_find_by(nick: "alice")
```

## Solution

Added error handling around the `connection.transaction_open?` check in both `create_or_find_by` and `create_or_find_by!` methods. If the check raises an exception, we now default to `false` (assuming the transaction is closed) and proceed with `find_by!` without a lock. This ensures the method can still successfully find the existing record even when the connection state check fails.

### Code Changes

**Before:**
```ruby
rescue ActiveRecord::RecordNotUnique
    where(attributes).lock.find_by!(attributes)
  else
    find_by!(attributes)
  end
end
```

**After:**
```ruby
rescue ActiveRecord::RecordNotUnique
  # Safely check if transaction is open, defaulting to false if check fails
  transaction_open = begin
    connection.transaction_open?
  rescue
    false
  end

  if transaction_open
    where(attributes).lock.find_by!(attributes)
  else
    find_by!(attributes)
  end
end
```

## Changes Made

1. **`activerecord/lib/active_record/relation.rb`**
   - Added error handling in `create_or_find_by` method (line 283-287)
   - Added error handling in `create_or_find_by!` method (line 310-314)

2. **`activerecord/test/cases/relations_test.rb`**
   - Added `test_create_or_find_by_handles_connection_error_when_checking_transaction_status` test
   - Added `test_create_or_find_by_bang_handles_connection_error_when_checking_transaction_status` test

## Testing

Added comprehensive test cases that:
- Create a duplicate record scenario (triggers `RecordNotUnique`)
- Simulate a connection error when checking transaction status
- Verify the method still finds the existing record despite the error

Both tests pass successfully:
- ✅ `test_create_or_find_by_handles_connection_error_when_checking_transaction_status`
- ✅ `test_create_or_find_by_bang_handles_connection_error_when_checking_transaction_status`

## Impact

- **Backward Compatible**: The fix maintains existing behavior for normal cases
- **Error Resilient**: Handles edge cases where connection state checks fail
- **No Breaking Changes**: All existing tests continue to pass

## Related Issues

This fix addresses a potential failure scenario where connection state checks could raise exceptions in edge cases (network issues, connection pool problems, etc.), making the `create_or_find_by` methods more robust and reliable.

### DIFF
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -279,10 +279,10 @@ module ActiveRecord
         end
         record
       rescue ActiveRecord::RecordNotUnique
-        # Safely check if transaction is open, defaulting to false if check fails
+        # Safely check if transaction is open, defaulting to false if connection is unavailable
         transaction_open = begin
           connection.transaction_open?
-        rescue
+        rescue ActiveRecord::ConnectionNotEstablished, ActiveRecord::ConnectionFailed
           false
         end
 
@@ -306,10 +306,10 @@ module ActiveRecord
         end
         record
       rescue ActiveRecord::RecordNotUnique
-        # Safely check if transaction is open, defaulting to false if check fails
+        # Safely check if transaction is open, defaulting to false if connection is unavailable
         transaction_open = begin
           connection.transaction_open?
-        rescue
+        rescue ActiveRecord::ConnectionNotEstablished, ActiveRecord::ConnectionFailed
           false
         end
 
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -279,7 +279,14 @@ module ActiveRecord
         end
         record
       rescue ActiveRecord::RecordNotUnique
-        if connection.transaction_open?
+        # Safely check if transaction is open, defaulting to false if check fails
+        transaction_open = begin
+          connection.transaction_open?
+        rescue
+          false
+        end
+
+        if transaction_open
           where(attributes).lock.find_by!(attributes)
         else
           find_by!(attributes)
@@ -299,7 +306,14 @@ module ActiveRecord
         end
         record
       rescue ActiveRecord::RecordNotUnique
-        if connection.transaction_open?
+        # Safely check if transaction is open, defaulting to false if check fails
+        transaction_open = begin
+          connection.transaction_open?
+        rescue
+          false
+        end
+
+        if transaction_open
           where(attributes).lock.find_by!(attributes)
         else
           find_by!(attributes)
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/relation.rb
```diff
@@ -279,7 +279,14 @@ def create_or_find_by(attributes, &block)
...
+        transaction_open = begin
+          connection.transaction_open?
+        rescue
+          false
+        end
+
+        if transaction_open
           where(attributes).lock.find_by!(attributes)
         else
           find_by!(attributes)
```
tahsin352, line 292: If transaction connection is not open (false), then how we will search the DB for the attributes?

activerecord/lib/active_record/relation.rb
```diff
@@ -279,7 +279,14 @@ def create_or_find_by(attributes, &block)
         end
         record
       rescue ActiveRecord::RecordNotUnique
-        if connection.transaction_open?
+        # Safely check if transaction is open, defaulting to false if check fails
+        transaction_open = begin
+          connection.transaction_open?
+        rescue
```
tahsin352: This is a very bad practice to swallow all errors, please use specific exceptions like `rescue ActiveRecord::ConnectionFailed` instead.

activerecord/lib/active_record/relation.rb
afurm: `transaction_open?` only checks if we’re inside an explicit transaction. Even when it’s false, `find_by!` still queries the DB using the open connection from `with_connection`. We only add `.lock` when a transaction is open because `FOR UPDATE` requires one. So the lookup still happens; it just doesn’t take a row lock outside a transaction.

activerecord/lib/active_record/relation.rb
afurm: Thanks for the catch — agree we shouldn’t swallow everything. The intent was only to guard against connection-loss cases, so I’ll narrow this to connection errors (e.g., `ActiveRecord::ConnectionNotEstablished` / `ActiveRecord::ConnectionFailed`) and let anything else bubble. If you think `transaction_open?` shouldn’t raise at all, I’m also happy to drop the rescue entirely.

PR #56460, opened 2025-12-25
[ci skip] Use `/app/models/concerns/` for the `Notifications` concern in Getting Started with Rails

dimasusername:
### Motivation / Background

**Getting Started with Rails, 16.4 Extracting a Concern** instructs the reader to place the custom `Product::Notifications` concern inside the `/app/models/product/` folder.

I think this is actually a great opportunity to introduce the reader to the dedicated `/app/models/concerns/` directory by creating a `Notifications` concern there instead.

### Detail

This Pull Request changes `guides/source/getting_started.md` section 16.4.

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2455,11 +2455,11 @@ Concern is a Ruby module with some syntactic sugar to make using them easier.
 
 First let’s create the Notifications module.
 
-Create a file at `app/models/product/notifications.rb` with the following:
+Create a file at `app/models/concerns/notifications.rb` with the following:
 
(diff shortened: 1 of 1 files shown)


benngarcia:
The concern here relies on the including model having an inventory_count attribute, which makes it model-specific rather than a generic concern.

This seems like a good opportunity to clarify when concerns should live under app/models/<model> (schema- or domain-specific) versus app/models/concerns (reusable across models).

dimasusername:
@benngarcia yes, I agree and like this better. I'll update the PR.

PR #56462, opened 2025-12-26
Add alias for sunday and monday methods

Yuhi-Sato:
### Motivation / Background

This Pull Request has been created to improve the clarity of the `sunday` and `monday` methods in `DateAndTime::Calculations`, which is included by Date class and Time class.

The current method names `sunday` and `monday` do not explicitly convey their behavior - specifically, that `monday` returns previous monday (or current date if today is Monday), and sunday returns next sunday (or current date if today is Sunday). This implicit behavior can be confusing for developers who are unfamiliar with these methods.

### Detail

  This Pull Request adds more descriptive aliases for the existing `sunday` and `monday` methods:

  - `previous_or_current_monday` as an alias for `monday` - makes it clear that the method returns either the previous Monday or the current date if today is Monday
  - `next_or_current_sunday` as an alias for `sunday` - makes it clear that the method returns either the next Sunday or the current date if today is Sunday

  The original method names remain unchanged to maintain backward compatibility. These aliases provide a more explicit and self-documenting alternative for developers who prefer clearer method names.

```.ruby
  # Both return the same result
  date.monday
  date.previous_or_current_monday

  # Both return the same result  
  date.sunday
  date.next_or_current_sunday
```

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/date_and_time/calculations.rb
+++ b/activesupport/lib/active_support/core_ext/date_and_time/calculations.rb
@@ -275,6 +275,7 @@ module DateAndTime
     def monday
       beginning_of_week(:monday)
     end
+    alias :previous_or_current_monday :monday
 
     # Returns a new date/time representing the end of this week on the given day.
     # Week is assumed to start on +start_day+, default is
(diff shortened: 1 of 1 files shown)


byroot:
Thank you for the PR, but I don't think such alias is necessary.

PR #56470, opened 2025-12-27
Change production host default to IPv6 (`::`)

schneems:
The `bin/rails server` command now binds to `::` (all IPv6 interfaces) in non-development environments, instead of `0.0.0.0`. The development environment continues to use `localhost` for backwards compatibility.

IPv6 sockets for `::` on most operating systems accept both IPv4 and IPv6 connections, so existing workflows will continue to work. This change prepares Rails for the increasing adoption of IPv6 and reduces reliance on increasingly expensive public IPv4 addresses (https://aws.amazon.com/blogs/aws/new-aws-public-ipv4-address-charge-public-ip-insights/). 

Here's how different hosts map to different urls: 

| Host | localhost:3000 | 127.0.0.1:3000 | \[::1\]:3000 | 0.0.0.0:3000 | \[::\]:3000 | Network Open |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `localhost` | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ No |
| `127.0.0.1` | ✅ | ✅ | ❌ | ✅ | ❌ | ❌ No |
| `::1` | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ No |
| `0.0.0.0` | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ Yes |
| `::` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ Yes |

This change:

- `0.0.0.0` => `::` (production)

I have a corresponding issue to update the defaults in Puma to IPv6 https://github.com/puma/puma/issues/3812 and a PR https://github.com/puma/puma/pull/3847.

### Checklist


seuros:
This is tricky. 

I currently have systems with IPv6 disabled at the kernel level where `socket(AF_INET6, ...)` returns `EAFNOSUPPORT (47)` - "Address family not supported".                                                                     
                                                                                                                                                                                                                                   
   Other cases where this could break:                                                                                                                                                                                             
   - Docker containers without `--ipv6` or network driver support                                                                                                                                                                  
   - Some cloud VMs (need upgrade)                                                                                                                                                                                                 
   - Corporate networks with IPv6 stripped
   - Loadbalancer or reverse proxies than have just Ipv4 active
   - Configuration that is hardcoded for 127.1 instead of ::

---

Edit: check below

schneems:
> Docker containers without --ipv6 or network driver support

Can you give me an example of a failure case? I tried reproducing with docker and couldn't get it to fail:

```
$ docker network create --ipv6=false disable_ipv6
$ docker build -t rails-ipv6-test .
$ $ docker run --rm \
  --network disable_ipv6 \
  -e RAILS_ENV=production \
  -e SECRET_KEY_BASE=dummy \
  -p 3000:3000 \
  rails-ipv6-test
[1] Puma starting in cluster mode...
[1] * Puma version: 7.1.0 ("Neon Witch")
[1] * Ruby version: ruby 3.3.10 (2025-10-23 revision 343ea05002) [aarch64-linux]
[1] *  Min threads: 3
[1] *  Max threads: 3
[1] *  Environment: production
[1] *   Master PID: 1
[1] *      Workers: 2
[1] *     Restarts: (✔) hot (✖) phased (✖) refork
[1] * Preloading application
[1] * Listening on http://[::]:3000
[1] Use Ctrl-C to stop
[1] - Worker 0 (PID: 12) booted in 0.0s, phase: 0
[1] - Worker 1 (PID: 15) booted in 0.0s, phase: 0
source=rack-timeout id=f19c45cb-7b84-49a3-b143-b7a71adf6ba1 timeout=15000ms state=ready
[f19c45cb-7b84-49a3-b143-b7a71adf6ba1] Started GET "/" for ::ffff:192.168.65.1 at 2025-12-29 22:54:21 +0000
[f19c45cb-7b84-49a3-b143-b7a71adf6ba1] Processing by WelcomeController#index as HTML
```

Assuming my reproduction is incorrect somehow, and it does fail for that case, if it fails with a nice error message, then I think that's good enough. For example, `Errno::EAFNOSUPPORT` would fail with a nice error message indicating the problem, but I'm unsure if that's what would actually happen in the wild:

```
Address family not supported by protocol - socket(2) for "::" port 3000
```

Then someone would see the problem right away and address it. I pasted that error into a claude with a fresh, empty directory and it nailed it first time https://gist.github.com/schneems/9de0837405834891986cce72b577a817.

> Some cloud VMs (need upgrade)

World IPv6 Launch Day was June 6, 2012 which is 4,954 days ago. I think that systems without ipv6 enabled are the exception rather than the rule (or if not now, leaning that way into the future). And like above, if the app fails with a good error message, I think that's good-enough.

> Loadbalancer or reverse proxies than have just Ipv4 active

It can still receive ipv4 traffic

> Configuration that is hardcoded for 127.1 instead of ::

Binding to `::` is still reachable via 127.1 still works with `::` see the chart above, though `localhost` is better for compatibility.

schneems:
As a data point Puma 8.0 was released with the linked change 7 days ago, and we've had zero reports of problems. This graph by google today is saying that 50% of traffic is now IPv6 https://news.ycombinator.com/item?id=47777894.

seuros:
Thank you for bumping this thread.

I forgot to update, i the concern is not current anymore.   I checked the major operating system Linux and BSDs code source and i can assert that fallback was implement in most of them without causing hard crash at least in in past 4 years and backported to earlier version.

Using :: with a system that was stripped from ipv6 will just fall back to 0.0.0.0.

nateberkopec:
This is now mainstreamed in Puma.

PR #56472, opened 2025-12-27
authenticated users shouldn't access new session path in authenticati…

gregmolnar:
…on generator

### Motivation / Background

Currently, the code generated by the authentication generator permits a user to access the sign in page even with an authenticated session. That shouldn't be permitted and with this change, they will be redirected to the root_url. There is also a new helper method to add the same behavior to a sign up flow.,
### Checklist

* [] Tests are added or updated if you fix a bug or add a feature.
* [] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
+++ b/railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
@@ -10,6 +10,11 @@ module Authentication
     def allow_unauthenticated_access(**options)
       skip_before_action :require_authentication, **options
     end
+
+    def ensure_unauthenticated_access(**options)
+      allow_unauthenticated_access **options
+      before_action :require_unauthenticated, **options
+    end
   end
 
   private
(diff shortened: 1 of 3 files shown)


callmesangio:
Should the same change be applied to `PasswordsController` as well?

gregmolnar:
@callmesangio good point. I will change that too.

gregmolnar:
Is there any change I should do to this to get it merged?

railties/lib/rails/generators/rails/authentication/templates/app/controllers/concerns/authentication.rb.tt
```diff
@@ -21,6 +26,10 @@ module Authentication
       resume_session || request_authentication
     end
 
+    def require_unauthenticated
+      redirect_to root_url if authenticated?
```
callmesangio, line 30: I think redirecting to `root_path` would probably be more appropriate here, since we're not changing host.

gregmolnar: The after login part also uses the _url helper, so I believe we should keep all of them the same. @dhh do you prefer to use the _url or the _path helpers for the internal redirects?

dhh:
I don't think it's worth adding this because I actually think it's fine that you can login as a different user even though you're already logged in.

PR #56474, opened 2025-12-28
Add `ActionDispatch::Request#bearer_token` to extract the bearer token

dhh:
Add `ActionDispatch::Request#bearer_token` to extract the bearer token from the Authorization header. Bearer tokens are commonly used for API and MCP requests.

### DIFF
--- a/actionpack/lib/action_dispatch/http/request.rb
+++ b/actionpack/lib/action_dispatch/http/request.rb
@@ -469,6 +469,11 @@ module ActionDispatch
       get_header("REDIRECT_X_HTTP_AUTHORIZATION")
     end
 
+    # Returns the bearer token embedded in the authorization header or nil if missing.
+    def bearer_token
+      authorization.to_s[/\ABearer (.+)\z/, 1]
+    end
+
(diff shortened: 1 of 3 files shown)


dhh:
(Unrelated flaky test).

actionpack/lib/action_dispatch/http/request.rb
```diff
@@ -469,6 +469,11 @@ def authorization
       get_header("REDIRECT_X_HTTP_AUTHORIZATION")
     end
 
+    # Returns the bearer token embedded in the authorization header or nil if missing.
+    def bearer_token
+      authorization.to_s[/\ABearer (.+)\z/, 1]
```
songjiz, line 474: Would it make sense to support both Token and Bearer prefixes here?

Currently, [ActionController::HttpAuthentication::Token](https://github.com/rails/rails/blob/0cfd49f401f12ef5d32f811f93150c6cbf92c1b3/actionpack/lib/action_controller/metal/http_authentication.rb#L427) uses the following regex: 

```ruby
TOKEN_REGEX = /^(Token|Bearer)\s+/
```

PR #56476, opened 2025-12-29
Add LeakChecker to tools and fail on ENV leaks

zzak:
This adds some support for detecting leaked env vars to our test suite, and fixes any detected cases.

I was debugging this error from a flakey test:
https://buildkite.com/rails/rails/builds/125076#019b624d-1383-4554-97f3-296c60ddeff4/L1729

```
Error:
InfoControllerTest#test_info_controller_returns_fuzzy_matches_for_route_paths:
ArgumentError: `secret_key_base` for test environment must be a type of String`
    lib/rails/application/configuration.rb:559:in `secret_key_base='
    lib/rails/application/configuration.rb:543:in `secret_key_base'
    lib/rails/application.rb:482:in `secret_key_base'
    lib/rails/application.rb:325:in `env_config'
bin/test railties/test/rails_info_controller_test.rb:207
```

First, I thought it might be something leaking `SECRET_KEY_BASE` or `SECRET_KEY_BASE_DUMMY`, but ended up digging this hole instead.

Inspired by Ruby's [leakchecker](https://github.com/ruby/ruby/blob/master/tool/lib/leakchecker.rb) if only in spirit, but at least wanted to put this up to get feedback. :bow:

### DIFF
--- a/tools/support/leak_checker.rb
+++ b/tools/support/leak_checker.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module LeakChecker
+  # These keys are out of our control.
+  # Either added by the system or by other libraries.
+  ALLOWED_KEYS = %w[
+    VIPSHOME
+  ]
+
+  def before_setup
+    @__leak_checker_before_env = ENV.to_h
+    super
+  end
+
+  def after_teardown
+    super
+
+    after = ENV.to_h
+    before = @__leak_checker_before_env
+
+    ALLOWED_KEYS.each do |k|
+      after.delete(k)
+      before.delete(k)
+    end
+
+    if after != before
+      message = +"Environment leak detected:\n"
+      added_keys = after.keys - before.keys
+      unless added_keys.empty?
+        message << "  - Added variables:\n"
+        added_keys.each do |k|
+          message << "    - #{k}"
+        end
+      end
+
+      removed_keys = before.keys - after.keys
+      unless removed_keys.empty?
+        message << "  - Removed variables:\n"
+        removed_keys.each do |k|
+          message << "    - #{k}"
+        end
+      end
+
+      changed_keys = (before.keys & after.keys).select { |k| before[k] != after[k] }
+      unless changed_keys.empty?
+        message << "  - Changed variables:\n"
+        changed_keys.each do |k|
+          message << "    - #{k} from #{before[k].inspect} to #{after[k].inspect}"
+        end
+      end
+
+      flunk message
+    end
+  end
+end
--- a/tools/test_common.rb
+++ b/tools/test_common.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "support/leak_checker"
+
 ActiveSupport::TestCase.alias_method :force_skip, :skip
 
 ENV["RAILS_TEST_EXECUTABLE"] = "bin/test"
@@ -20,3 +22,5 @@ if ENV["BUILDKITE"]
   end
   ActiveSupport::TestCase.include(DisableSkipping)
 end
+
+ActiveSupport::TestCase.prepend(LeakChecker)
--- a/actionview/lib/action_view/test_case.rb
+++ b/actionview/lib/action_view/test_case.rb
@@ -266,6 +266,10 @@ module ActionView
         end
       end
 
+      def after_setup
+        @__setup_ivars = instance_variables << :@__setup_ivars
+      end
+
       def setup_with_controller
         controller_class = Class.new(ActionView::TestCase::TestController)
         @controller = controller_class.new
@@ -399,7 +403,7 @@ module ActionView
       ]
 
       def _user_defined_ivars
-        instance_variables - INTERNAL_IVARS
+        instance_variables - INTERNAL_IVARS - @__setup_ivars
       end
 
       # Returns a Hash of instance variables and their values, as defined by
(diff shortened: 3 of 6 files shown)


railties/test/application/configuration_test.rb
```diff
@@ -274,6 +274,8 @@ def change
 
       ENV["RAILS_GROUPS"] = "javascripts,stylesheets"
       assert_equal [:default, "test", "javascripts", "stylesheets"], Rails.groups
+    ensure
+      ENV.delete "RAILS_GROUPS"
```
byroot: We shouldn't `ENV.delete` but restore to whatever it was. We got a helper for that (`set_env` I think?) that is copy pasted in a bunch of places. Would be good to clean that up.

railties/test/application/initializers/frameworks_test.rb
```diff
@@ -368,7 +368,11 @@ def show
       assert_match(/#{database_url_db_name}/, ActiveRecord::Base.connection_db_config.database)
     ensure
       ActiveRecord::Base.remove_connection
-      ENV["DATABASE_URL"] = orig_database_url if orig_database_url
+      if orig_database_url
+        ENV["DATABASE_URL"] = orig_database_url
+      else
+        ENV.delete("DATABASE_URL")
```
byroot: Same here.

byroot (approved):
Generally in favor of this. Tests leak turns into weird CI flakiness, so catching it up upfront is best practice.

Once the `ENV.delete` are removed I'm 👍

tools/support/leak_checker.rb
```diff
@@ -0,0 +1,36 @@
...
+      keys = LeakChecker.env_keys_delta
+      if keys.any?
+        flunk "Environment leak detected! #{keys.join(", ")}"
+      end
+    end
+  end
+
+  class << self
+    def setup
+      @before_env_keys = ENV.keys
```
byroot: Why check only the keys? You can get a snapshot of the values too with `ENV.to_h`

zzak: I guess my instinct is "don't store env values anywhere if I don't need them" but I've changed it as you suggested. :bow:

railties/test/application/configuration_test.rb
```diff
@@ -51,18 +51,6 @@ def copy_app
       FileUtils.cp_r(app_path, new_app)
     end
 
-    def app(env = "development")
```
zzak: This is actually just the same as here but with a different default:
https://github.com/rails/rails/blob/60d92e4e7dfe923528ccdccc18820ccfe841b7b8/railties/test/isolation/abstract_unit.rb#L71

activesupport/test/test_case_test.rb
```diff
@@ -578,6 +578,18 @@ def teardown
...
   private
+    def callbacks(name)
+      self.class.send(name).reject do |callback|
+        if callback.filter.respond_to?(:source_location)
+          callback.filter.source_location.first.match?(
+            Regexp.union(
+              /leak_checker/,
+            )
+          )
+        end
```
byroot: This is really ugly :/

Surely there is a cleaner and simpler way to achieve this.

byroot:
Can you move the fixes to another PR, so we can merge that right away, and then work on cleaning up the leak checker code?

I don't think using a concern is helping, the previous version you got was simpler.

tools/support/leak_checker.rb
```diff
@@ -0,0 +1,21 @@
...
+  ]
+
+  def before_setup
+    @@before_env = ENV.to_h
+    super
+  end
+
+  def after_teardown
+    super
+    if (keys = ENV.keys - @@before_env.keys - ALLOWED_KEYS) && keys.any?
```
byroot: I still think we should not only check the keys but also the values.

`if ENV.to_h != @@before_env then figure_out_the_diff`

tools/support/leak_checker.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module LeakChecker
+  # These keys are out of our control.
+  # Either added by the system or by other libraries.
+  ALLOWED_KEYS = %w[
+    VIPSHOME
+  ]
+
+  def before_setup
+    @@before_env = ENV.to_h
```
byroot: Not sure we should use class vars here. It's rare but technically you can "parallelize" tests with threads.

tools/test_common.rb
```diff
@@ -20,3 +22,5 @@ def skip(message = nil, *)
   end
   ActiveSupport::TestCase.include(DisableSkipping)
 end
+
+ActiveSupport::TestCase.prepend(LeakChecker)
```
byroot, line 26: I changed it so it also runs locally. I don't think it will cause issues, but if it does we can always reconsider and make it CI only again.

tools/support/leak_checker.rb
zzak: I was a bit worried about including the values, and inspecting them because there might be a safety concern but generally we don't provide that value to the build runner anyways -- because they can just `puts ENV` anyways so no point in worrying :joy:

zzak:
Thanks for getting this over the line! :bow:

george-ma:
Hey @zzak I'm investigating a regression in our test suite after this PR and I wanted to confirm if this behaviour change was intentional

```ruby
def _user_defined_ivars
  instance_variables - INTERNAL_IVARS - @__setup_ivars
end
```

`_user_defined_ivars` is used by [view_assigns](https://github.com/rails/rails/blob/e69d30b0dbbd4da4caebb66c978716841a55bde2/actionview/lib/action_view/test_case.rb#L409C1-L417C10) to determine which ivars are passed to views during `render()`

After this change (`- @__setup_ivars`) ivars set in setup blocks are no longer passed to render views. If intentional should this be documented as a breaking change? (and what is the recommended pattern going forward), and if not should should `@__setup_ivars` filtering be removed from `_user_defined_ivars`?

Thanks in advance!

byroot:
I did this part of the change. I didn't expect people to rely on this. Can you file an issue and tag me ? I'll try to find a solution soon.

PR #56477, opened 2025-12-29
Do not concat twice to the same draw_paths

kuahyeow:
When this initializer is called for the main application `routes` and `app.routes` are the same
object. So we do not need to concat it again.

### Motivation / Background

Fixes https://github.com/rails/rails/issues/56475

### Detail

### Checklist

### DIFF
--- a/railties/lib/rails/engine.rb
+++ b/railties/lib/rails/engine.rb
@@ -601,7 +601,11 @@ module Rails
       routing_paths = paths["config/routes.rb"].existent
       external_paths = self.paths["config/routes"].paths
       routes.draw_paths.concat(external_paths)
-      app.routes.draw_paths.concat(external_paths)
+
+      # Concat engine draw paths to the app also.
+      # If this initializer is being called for the app, it's redundant, so
+      # skip.
+      app.routes.draw_paths.concat(external_paths) unless app == self
 
       if routes? || routing_paths.any?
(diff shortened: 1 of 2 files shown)


railties/lib/rails/engine.rb
```diff
@@ -601,7 +601,7 @@ def load_seed
       routing_paths = paths["config/routes.rb"].existent
       external_paths = self.paths["config/routes"].paths
       routes.draw_paths.concat(external_paths)
-      app.routes.draw_paths.concat(external_paths)
+      app.routes.draw_paths.concat(external_paths) unless app == self
```
kuahyeow, line 608: Honestly, I am not sure this is the best fix.

kuahyeow: Alternative: we could change `draw_paths` to be initialized as `Set.new` rather than `[]`. But Set does not have a `concat` method. Maybe this might break applications and engines ?

shivabhusal: an alternative could be using this union-assignment(`|=`) operator. It does in-place replacement(actually mutated) after performing union. It only keeps uniq values maintaining order.

```ruby
app.routes.draw_paths |= external_paths
```

kuahyeow: OK, I think `app == self` is the best way for now. Added a code comment as well

PR #56490, opened 2025-12-31
Add Rails.app.revision

byroot:
Useful in many context, deployment tracking, error reporting, monitoring, even sometimes in cache keys.

Fix: https://github.com/rails/rails/pull/56080

Co-Authored-By: @seuros

### DIFF
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -119,6 +119,8 @@ module Rails
       @message_verifiers = nil
       @deprecators       = nil
       @ran_load_hooks    = false
+      @revision          = nil
+      @revision_initialized = false
 
       @executor          = Class.new(ActiveSupport::Executor)
       @reloader          = Class.new(ActiveSupport::Reloader)
@@ -394,7 +396,35 @@ module Rails
       self.class.isolate_namespace(mod)
     end
 
+
+    # Returns the application's revision (deployment identifier).
+    # Useful for error reporting and deployment verification.
+    #
+    # Set via config.revision (string or proc) or REVISION file.
+    # Always either a String or +nil+.
+    def revision
+      unless @revision_initialized
+        @revision = begin
+          root.join("REVISION").read.strip.presence
+        rescue SystemCallError
+          if Dir.exist?(".git")
+            rev = `git rev-parse HEAD 2> /dev/null`.strip.presence
+            rev if $?.success?
+          end
+        end
+        @revision_initialized = true
+      end
+      @revision
+    end
+
     ## Rails internal API
+
+    def revision=(rev) # :nodoc:
+      rev = rev.call if rev.respond_to?(:call)
+      rev = rev&.to_s
+      @revision = rev
+      @revision_initialized = true
+    end
 
     # This method is called just after an application inherits from Rails::Application,
     # allowing the developer to load classes in lib and use them during application
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -538,6 +538,20 @@ Enables or disables reloading of classes only when tracked files change. By defa
 
 Causes the app to not boot if a master key hasn't been made available through `ENV["RAILS_MASTER_KEY"]` or the `config/master.key` file.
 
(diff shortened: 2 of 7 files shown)


seuros (approved):

bbenno:
Would it be reasonable to allow users to append an environment-dependent suffix? For example, I prefer a `-dev` suffix for development revisions. Enabling the direct specification of such environment-specific suffixes would obviate the need for ad hoc extensions and would allow me to remove my previous, largely equivalent implementation of this functionality.

I am currently occupied with other commitments; however, I expect to revisit this matter in the coming weeks and, time permitting, to follow up with a pull request should this be considered beneficial.

seuros:
@bbenno , the initial proposal was based on this [gem](https://github.com/seuros/rails_app_version) , 

Maybe we could still get the .environment . 

I use that because everything except dev/test is **production**.

railties/lib/rails/application.rb
```diff
@@ -394,8 +396,36 @@ def isolate_namespace(mod)
...
+        end
+        @revision_initialized = true
+      end
+      @revision
+    end
+
     ## Rails internal API
 
+    def revision=(rev) # :nodoc:
+      rev = rev.call if rev.respond_to?(:call)
```
ghiculescu, line 423: What's the point of supporting a proc here, if it will get called immediately on assignment?

this

```ruby
config.revision = -> { File.read("BUILD_ID").strip }
```

is effectively identical to this:

```ruby
config.revision = File.read("BUILD_ID").strip
```

byroot: Yeah, you're right.

byroot: https://github.com/rails/rails/pull/56508

G-Rath:
@byroot apologies in advance if this isn't the right place to do this, but we're found it useful to have the revision time as well which is something [I implemented support for in Capistrano](https://github.com/capistrano/capistrano/pull/2155).

Would you accept a PR implementing that too, mirroring the work done here?

byroot:
I don't see it as useful enough to be Rails core.

railties/lib/rails/application.rb
fxn: Hey! Should `Dir.exist?(".git")` be also relative to `root`?

fxn: Maybe even pass `-C root` to the command?

railties/lib/rails/application.rb
```diff
@@ -394,8 +396,36 @@ def isolate_namespace(mod)
...
+    # Useful for error reporting and deployment verification.
+    #
+    # Set via config.revision (string or proc) or REVISION file.
+    # Always either a String or +nil+.
+    def revision
+      unless @revision_initialized
+        @revision = begin
+          root.join("REVISION").read.strip.presence
+        rescue SystemCallError
+          if Dir.exist?(".git")
```
Saidbek, line 410: https://github.com/rails/rails/pull/56524 to ensure the revision is read from the app's git repo

cc @byroot

railties/lib/rails/application.rb
byroot: No, I've done this on purpose to handle tests that do `Dir.chdir`.

fxn: What do you mean?

I mean, you cannot generally assume the current working directory is `Rails.root`. So you compute the absolute path to `.git` relative to `Rails.root`, and pass `-C` to `git`.

Same reason the code checks for `root.join("REVISION")` and not `File.read("REVISION")`. No?

You have something else in mind?

byroot: In a previous version of this PR, I was indeed checking for `Rails.root.join(".git").exist?` but then ran into an issue with some tests that use the `build_app` helper.

That being said, as mentioned in the followup PR, I think we shouldn't check for the directory at all, and directly shell out to `git`, because it's legitimate for the app to be in a subdirectory of of the repo. So yes, I think `-C Rails.root` might be the way to go.

PR #56496, opened 2026-01-01
Add PostgreSQLAdapter.register_type_mapping

seuros:
Gems adding types to postgresql for [postgis](https://github.com/seuros/activerecord-postgis) or  for [vector](https://github.com/ankane/neighbor) monkey patch the adapter to add the types. 

This interface will allow us to add types with a cleaner interface.

cc @ankane

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -775,6 +775,25 @@ module ActiveRecord
             OID::Interval.new(precision: precision)
           end
         end
+
+        # Registers a callback to extend the type map during initialization.
+        # Useful for third-party gems that need to register custom SQL types.
+        #
+        #   ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.register_type_mapping do |type_map|
+        #     type_map.register_type("geometry") do |oid, fmod, sql_type|
+        #       MyGeometryType.new(sql_type)
+        #     end
+        #   end
+        #
+        def register_type_mapping(&block)
+          raise ArgumentError, "block required" unless block_given?
+          @@type_mapping_callbacks ||= []
+          @@type_mapping_callbacks << block
+        end
+
+        def clear_type_mapping_callbacks! # :nodoc:
+          @@type_mapping_callbacks = [] if defined?(@@type_mapping_callbacks)
+        end
       end
 
       private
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -770,6 +770,38 @@ def initialize_type_map(m) # :nodoc:
...
+        # Registers a callback to extend the type map during initialization.
+        # Useful for third-party gems that need to register custom SQL types.
+        #
+        #   ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.register_type_mapping do |type_map|
+        #     type_map.register_type("geometry") do |oid, fmod, sql_type|
+        #       MyGeometryType.new(sql_type)
+        #     end
+        #   end
+        #
+        def register_type_mapping(priority: 0, &block)
```
byroot: Is the priority thing really necessary? I don't get the point of it.

seuros: Sorry i missed the notification.

The priority parameter enables layered type matching for complex types like PostGIS geometries:
Without priority:
1. Generic geometry handler catches everything =>  POINT Z, POINT ZM all match first (They are similar but hold extra value)
2. Specific subtypes never get a chance to match

With priority:
```
# High priority = runs LAST (most specific, can build on previous results)
ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.register_type_mapping(priority: 100) do |type_map|
  type_map.register_type("pointz") do |oid, fmod, sql_type|
    PointZ.new(sql_type)
  end
  type_map.register_type("pointzm") do |oid, fmod, sql_type|
    PointZM.new(sql_type)
  end
end
# Low priority = runs FIRST (generic fallback)
ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.register_type_mapping(priority: 0) do |type_map|
  type_map.register_type("geometry") do |oid, fmod, sql_type|
    Geometry.new(sql_type)
  end
end
```
This allows gems like PostGIS to register specific subtypes (POINT Z, POINT ZM, POLYGON Z, etc.) at high priority.

If this pattern apply only to postgis, we can drop the priority param and have correct order in the gem.

However if other extensions have such pattern, it preferable to keep it.

byroot: Yeah, I'd rather avoid YAGNI complexity.

Please remove it. Other than that this PR is fine to me.

PR #56503, opened 2026-01-02
[ci skip] Add parameter keep_id_suffix into API docs for humanize method 

tahsin352:
### Motivation / Background

This Pull Request has been created because if we set `capitalize: true` and `keep_id_suffix: true`, then `humanize` and `titleize` methods will display `ID` instead of `Id`, as it is more readable, user friendly. Currently it displays `Id` which is not user friendly.

### Detail

Currently, the `keep_id_suffix` parameter prevents a trailing "_id" from being stripped, but does not capitalize it.

```ruby
ActiveSupport::Inflector.humanize("author_id", keep_id_suffix: true) # => "Author id"
"author_id".humanize(keep_id_suffix: true) # => "Author id"
```

This Pull Request changes to

```ruby
ActiveSupport::Inflector.humanize("author_id", keep_id_suffix: true) # => "Author ID"
"author_id".humanize(keep_id_suffix: true) # => "Author ID"
```

Which is more user friendly and readable.

### Checklist

### DIFF
--- a/guides/source/active_support_core_extensions.md
+++ b/guides/source/active_support_core_extensions.md
@@ -1781,12 +1781,16 @@ Specifically, it performs these transformations:
 The capitalization of the first word can be turned off by setting the
 `:capitalize` option to false (default is true).
 
+The trailing "_id" can be kept and capitalized by setting the optional
+parameter `keep_id_suffix` to true (default is true).
+
 ```ruby
-"name".humanize                         # => "Name"
-"author_id".humanize                    # => "Author"
-"author_id".humanize(capitalize: false) # => "author"
-"comments_count".humanize               # => "Comments count"
-"_id".humanize                          # => "Id"
+"name".humanize                            # => "Name"
+"author_id".humanize                       # => "Author"
+"author_id".humanize(capitalize: false)    # => "author"
+"comments_count".humanize                  # => "Comments count"
+"_id".humanize                             # => "Id"
+"author_id".humanize(keep_id_suffix: true) # => "Author id"
 ```
 
 If "SSL" was defined to be an acronym:


callmesangio:
The current semantics of the `capitalize` argument of `#humanize` are about the first word of the resulting output. I'm not sure that extending them to also control the _upcasing_ of the "id" suffix (where upcasing != capitalization) is something actually expected or even desirable.

For instance, I don't think this is what we want:

```ruby
"author_id_and_book_id".humanize(keep_id_suffix: true)  # => "Author id and book ID"
```

tahsin352:
'ID' is generally considered the clearer, more established abbreviation in most forms of communication. 'id' is less intuitive for the reader when abbreviating "identification". 

As the `keep_id_suffix` parameter only handles `_id`, so middle '_id_' remains as 'id'.

callmesangio:
> 'ID' is generally considered the clearer, more established abbreviation in most forms of communication. 'id' is less intuitive for the reader when abbreviating "identification".

My point is that historically `capitalize: true` has meant something completely unrelated, so it should not be coupled to the upcasing of the "_id" suffix.

> As the `keep_id_suffix` parameter only handles `_id`, so middle '_id_' remains as 'id'.

Yes, and this change creates an inconsistency we currently don't have.

A couple of other considerations:

- The API docs should be updated to reflect the new behavior (basically everywhere `keep_id_suffix` is accepted as an argument).
- In the regexes you use, you probably want to match the end of the string (`\z`) instead of the end of the line.

guides/source/active_support_core_extensions.md
```diff
@@ -1781,12 +1781,16 @@ Specifically, it performs these transformations:
...
+The trailing "_id" can be kept and capitalized by setting the optional
+parameter `keep_id_suffix` to true (default is true).
+
 ```ruby
 "name".humanize                         # => "Name"
 "author_id".humanize                    # => "Author"
 "author_id".humanize(capitalize: false) # => "author"
 "comments_count".humanize               # => "Comments count"
 "_id".humanize                          # => "Id"
+"author_id".humanize(keep_id_suffix: true) # => "Author id"
```
callmesangio, line 1793: I think it would be nice if the comments with the output were all vertically aligned, as above and in the [API docs](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/string/inflections.rb#L259).

guides/source/active_support_core_extensions.md
```diff
@@ -1781,12 +1781,16 @@ Specifically, it performs these transformations:
 The capitalization of the first word can be turned off by setting the
 `:capitalize` option to false (default is true).
 
+The trailing "_id" can be kept and capitalized by setting the optional
+parameter `keep_id_suffix` to true (default is true).
```
callmesangio, line 1785: - `keep_id_suffix` keeps but _does not_ capitalize the trailing "_id".
- The default of `keep_id_suffix` is `false`.

The API docs of `humanize` have been recently fixed, see https://github.com/rails/rails/pull/56427 for an updated description of the behavior of `keep_id_suffix`.

callmesangio (commented):
Probably commits should be squashed into one.

guides/source/active_support_core_extensions.md
tahsin352: Updated.

tahsin352:
> Probably commits should be squashed into one.

Rebased and squashed into one.

Issue #52929, opened 2024-09-14
Eject most Active Job adapters into gems·


dhh:
We used to ship with a million database adapters in the box, but eventually settled on just offering official support for mysql/pgsql/sqlite. Folks can of course use whatever database they desire, but the work to maintain db2 or sqlserver or oracle falls to those vendors or the broader community.

With Solid Queue becoming the default for Rails 8, it's time to do the same for Active Job. For starters, I think we can boil it down to the single-process adapters (inline/async/test) and then adapters for resque, sidekiq, and then of course Solid Queue (which actually just has the adapter in the gem!). Everything else should be extracted into separate adapter gems, like activejob-suckerpunch or whatever.

byroot:
IMO this is welcome because that will get rid of quite a bit of dependencies in the rails Gemfile.

However before creating `activejob-*` gems, it would probably be worth reaching to the underlying gem maintainers to ask if they'd rather include the Active Job adapter in their gem.

IMO the list would be:

| Adapter                |  Task   | PR |
|------------------------|---------|----|
| backburner_adapter     | Extract |    |
| inline_adapter         | Keep    |    |
| resque_adapter         | Depends   |    |
| sneakers_adapter       | Extract |    |
| test_adapter           | Keep    |    |
| async_adapter          | Keep    |    |
| delayed_job_adapter    | Extract |    |
| queue_classic_adapter  | Extract |    |
| sidekiq_adapter        | Depends   |    |
| sucker_punch_adapter   | Extract |    |

- `Extract`: try to get it upstreamed and if not create a gem
- `Depends`: try to get it upstreamed but if not, keep.
- `Keep`: Should remain in core.

dixpac:
I like it, and I’m open to helping with this feature.

> However before creating activejob-* gems, it would probably be worth reaching to the underlying gem maintainers to ask if they'd rather include the Active Job adapter in their gem.

👍

dhh:
I'd love to get these adapters upstreamed. But maybe we can just do both. Would be nice to dump all these dependencies for 8.0. So maybe if upstreaming immediately isn't an option, we can at least put the long tail of adapters in a gem for staging for that to happen.

byroot:
Is there really a rush? While as I said I'd love to reduce the dependencies needed to contribute to Rails, these doesn't appear actively harmful currently. So I don't really see why it would be seen as a blocker for Rails 8.0.

dhh:
[...]

PR #53058, opened 2024-09-26
Deprecate built-in `sidekiq` adapter

fatkodima:
Reference #52929.

Opening this to not miss out.

The next sidekiq release will add an Active Job adapter by default - https://github.com/sidekiq/sidekiq/pull/6430.

PR #56506, opened 2026-01-02
Deprecate delayed_job Active Job adapter

morgoth:
Follow up to https://github.com/collectiveidea/delayed_job/pull/1236

refs https://github.com/rails/rails/issues/52929

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Deprecate built-in `delayed_job` adapter.
+
+    If you're using this adapter, upgrade to `delayed_job` 4.2.0 or later to use the `delayed_job` gem's adapter.
+
+    *Dino Maric, David Genord II, Wojciech Wnętrzak*
+
 *   Deprecate built-in `backburner` adapter.
 
     *Dino Maric, Nathan Esquenazi, Earlopain*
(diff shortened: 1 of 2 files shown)


morgoth:
I wanted to add a test in a similar manner as https://github.com/rails/rails/pull/53058/changes
but delayed_job testing was removed in https://github.com/rails/rails/commit/d55ec9d5831b05ea5de75c105635c80376c0bf11 so without bringing it back it won't be possible.
Not sure if it is worth it.

Anyway, tested locally and deprecation works

albus522:
The downside is that the deprecation notice won't go away after upgrading. We'll probably need some form of version check as well

activejob/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+*   Deprecate built-in `delayed_job` adapter.
+
+    If you're using this adapter, upgrade to `delayed_job` 4.2.0 or later to use the `delayed_job` gem's adapter.
+
+    *Wojciech Wnętrzak*
```
dixpac:
```suggestion
    *Wojciech Wnętrzak, David Genord II, Dino Maric*
```

dixpac (commented):
🙌
Thanks @albus522  for your help

morgoth:
@albus522 I guess loading the adapter from gem should be handled in similar way as https://github.com/sidekiq/sidekiq/pull/6430/changes so no version checking required on Rails side

PR #56511, opened 2026-01-04
Drop support for Ruby 3.2

byroot:
Our support policy is that we support all Ruby release that aren't EOL at the time we release a new minor of Rails.

Ruby 3.2 will be EOL on April 1st, so by the time we'll release Rails 8.2 it will be EOL.

Might as well drop support now and spare CI resources.

NB: We don't have any Ruby 3.2 specific codepath  that I could find, so this does allow any cleanup. Dropping 3.3. will though.

### DIFF
--- a/actioncable/actioncable.gemspec
+++ b/actioncable/actioncable.gemspec
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
   s.summary     = "WebSocket framework for Rails."
   s.description = "Structure many real-time application concerns into channels over a single WebSocket connection."
 
-  s.required_ruby_version = ">= 3.2.0"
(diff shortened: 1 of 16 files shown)


matthewd:
> Our support policy is that we support all Ruby release that aren't EOL at the time we release a new minor of Rails.

Not really... we've often bumped our minimum ahead of upstream EOL, and will continue to sometimes do so. Our main policy is the inverse: that we do _not_ support Ruby releases that are already EOL at the time of our release.

(Commenting here because otherwise people will point to this and claim it's a commitment.)

byroot:
Right, subtle difference, but difference indeed :).

To be very honest, the only reason I dropped 3.2 support is CI resources. There's no maintenance burden in this specific case.

PR #35698, opened 2019-03-22
Output junit format test report

mtsmfm:
### Summary

Hi, I'm interested in analyzing Rails CI.

I created a tool to find flaky tests.

app: https://has-it-failed.herokuapp.com/
repo: https://github.com/mtsmfm/has-it-failed
slides: https://speakerdeck.com/mtsmfm/analyze-rails-ci

I wrote a parser to extract which test was failed from raw logs.

https://github.com/mtsmfm/has-it-failed/blob/75eb7f367d42b68ba3ff89d6a1618d03ef7692d5/parser/rails_ci.pegjs

But of course, it's not perfect and in some case failed to parse.
As you may know, junit format test report is de facto standard to output machine readable test result.
Unfortunately, TravisCI doesn't provide data store for build artifacts so we need AWS account.
https://docs.travis-ci.com/user/uploading-artifacts/

I heard Rails team is trying to use buildkite so let's upload test report!
https://buildkite.com/docs/pipelines/artifacts

`Minitest::Reporters::JUnitReporter` outputs test report to `<subdir of rails/rails>/test/reports/*.xml` by default.

What do you think?

PR #56512, opened 2026-01-04
Remove minitest-ci

byroot:
It was introduced in https://github.com/rails/rails/pull/35698 to allow automatic analysis of failures and such.

Overall it makes sense but it has been broken for ages, so let's just get rid of it. If someone is interested in fixing it, it can be re-introduced.

Until then, it get rid of a CGI warning on Ruby 4.0


zenspider:
Hopefully this addresses things: https://github.com/CircleCI-Public/minitest-ci/pull/35

PR #56513, opened 2026-01-04
Attempt to fix secret_key_base test failure

byroot:
We've been experiencing a nasty flaky test, yet it's not very clear what is happening:

```
Error:
InfoControllerTest#test_info_controller_allows_requests_when_all_requests_are_considered_local:
ArgumentError: `secret_key_base` for test environment must be a type of String, got: ""`
    lib/rails/application/configuration.rb:563:in 'Rails::Application::Configuration#secret_key_base='
    lib/rails/application/configuration.rb:547:in 'Rails::Application::Configuration#secret_key_base'
    lib/rails/application.rb:511:in 'Rails::Application#secret_key_base'
    lib/rails/application.rb:328:in 'Rails::Application#env_config'
```

### DIFF
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -679,13 +679,18 @@ module Rails
         def generate_local_secret
           key_file = root.join("tmp/local_secret.txt")
 
-          unless File.exist?(key_file)
-            random_key = SecureRandom.hex(64)
-            FileUtils.mkdir_p(key_file.dirname)
-            File.binwrite(key_file, random_key)
+          random_key = begin
+            File.binread(key_file)
+          rescue SystemCallError
+            nil
           end
 
-          File.binread(key_file)
+          return random_key if random_key.present?
+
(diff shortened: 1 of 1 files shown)


chaadow:
I occasionally had this exact same error on my CI ( Github CI ) when I run `bin/rails parallel:drop parallel:setup` ( with the `parallel_tests` gem, I ended up NOT relying on `generate_local_secret` but instead hardcode `SECRET_KEY_BASE` because it pissed me off.
I couldn't figure out a fix AND I didn't want to monkey patch to add what you did here: https://github.com/rails/rails/commit/0d1b4918594caad1df15aefaa0b8393bbe7e232e 

Because I couldn't reliably reproduce the bug.

And since I couldn'r reliably reproduce the bug, I didn't bother open a Rails issue...

chaadow:
@byroot Maybe add a suffix to `"tmp/local_secret#{ENV['PARALLEL_PROCESS_NUMBER'].txt"` per process, so that each process handles its own file in a multi process test environment. idk just throwing out ideas

PR #53335, opened 2024-10-16
Deprecate built-in snekaers adapter

dixpac:
[`sneakers` gem development](https://github.com/jondot/sneakers?tab=readme-ov-file#sneakers-has-a-new-home-and-a-new-name) has been moved to [kicks gem](https://github.com/ruby-amqp/kicks)

Connected to https://github.com/rails/rails/issues/52929

PR #53334, opened 2024-10-16
Deprecate built-in ActiveJob backburner adapter

dixpac:

PR #56517, opened 2026-01-05
Fix deprecation horizon for AJ adapters

morgoth:
They will be deprecated in 8.2

Follow up to https://github.com/rails/rails/pull/53334 and https://github.com/rails/rails/pull/53335

cc @guilleiguaran @yahonda 
https://github.com/rails/rails/pull/53334#issuecomment-3708536258


chaadow:
@morgoth  yeah I've reported it here: https://github.com/rails/rails/pull/53335#issuecomment-3676760483 . thanks

PR #56518, opened 2026-01-05
Remove deprecated sidekiq Active Job adapter

morgoth:
It is available in the sidekiq gem


guilleiguaran:
I think we should keep the mention of third-party gems in the AJ README so users know where to find more info outside of Rails.

morgoth:
@guilleiguaran This is still present here https://github.com/rails/rails/blob/main/guides/source/active_job_basics.md#starting-the-backend

I only removed mentions to the built in adapter

PR #56519, opened 2026-01-05
Use the random suffix in the atomic temp file name

djmb:
In https://github.com/rails/rails/commit/f781cc50e5b27502d01e38039f401bf7cc2fd369, the suffix was not appended to the temp file name, leading to collisions when concurrently writing to the same file.


byroot:
:facepalm:

PR #52904, opened 2024-09-12
Introduce ActiveModel::Inspect

p8:
This moves the Active Record inspect implementation to Active Model, to allow Active Model objects to show a formatted string with relevant attributes when calling `inspect`.

```ruby
class Person
  include ActiveModel::API
  include ActiveModel::Attributes
  include ActiveModel::Inspect
  attribute :name, :string
end
person = Person.new(name: "Amanda")
person.inspect
=> "#<Person name: \"Amanda\">"
```

Sensitive attributes can be filtered, just like in Active Record:

```ruby
  Person.filter_attributes = [:name]
  person.inspect
  => #<Person name: [FILTERED]>
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56521, opened 2026-01-05
Add inspect features to `ActiveModel::Attributes`

viniciusalmeida:
### Motivation / Background

`ActiveRecord::Base#inspect` provides a rich, human-readable representation of model instances with features like attribute filtering, string truncation, and date formatting. However, `ActiveModel::Model` classes fall back to Ruby's default `Object#inspect`, which outputs all instance variables in an unhelpful format:

  ```ruby
  # ActiveRecord - nice output
  User.new(name: "Alice").inspect
  #=> "#<User id: nil, name: \"Alice\">"

  # ActiveModel - falls back to Object#inspect
  class Person
    include ActiveModel::Model
    attribute :name, :string
  end

  Person.new(name: "Alice").inspect
  #=> "#<Person:0x00007f... @attributes=...>"
  ```
  
  ### Detail

This Pull Request adds `ActiveModel::Inspection` (included automatically via `ActiveModel::Attributes`) with full inspection feature parity to `ActiveRecord`:

  - `#inspect` - Formatted attribute output: `#<Person name: "Alice", age: 30>`
  - `#full_inspect` - Shows all attributes regardless of `attributes_for_inspect`
  - `#pretty_print` - Support for pp with proper formatting
  - `.filter_attributes` - Mask sensitive data with `[FILTERED]`
  - `.attributes_for_inspect` - Limit which attributes appear in output
  - String truncation at 50 characters
  - Date/time formatting with `to_fs(:inspect)`

### DIFF
--- a/activemodel/lib/active_model/inspection.rb
+++ b/activemodel/lib/active_model/inspection.rb
@@ -79,7 +79,7 @@ module ActiveModel
 
       # Returns the inspection filter used to mask sensitive attributes.
       def inspection_filter # :nodoc:
-        if defined?(@filter_attributes)
+        if defined?(@filter_attributes) && @filter_attributes
           @inspection_filter ||= begin
             mask = InspectionMask.new(ActiveSupport::ParameterFilter::FILTERED)
             ActiveSupport::ParameterFilter.new(@filter_attributes, mask: mask)
--- a/activemodel/lib/active_model/inspection.rb
+++ b/activemodel/lib/active_model/inspection.rb
@@ -1,0 +1,229 @@
+# frozen_string_literal: true
+
+module ActiveModel
+  # = Active \Model \Inspection
+  #
+  # Provides a human-readable representation of your model instances through
+  # the +inspect+ and +pretty_print+ methods.
+  #
+  # This module requires that the including class has an +@attributes+ instance
+  # variable (as provided by ActiveModel::Attributes) and responds to
+  # +attribute_names+.
+  #
+  # == Configuration
+  #
+  # The inspection output can be customized through two class attributes:
+  #
+  # * +filter_attributes+ - An array of attribute names whose values should be
+  #   masked in the output. Useful for sensitive data like passwords.
+  #
+  # * +attributes_for_inspect+ - An array of attribute names to include in
+  #   the output, or +:all+ to show all attributes (default).
+  #
+  # == Example
+  #
+  #   class Person
+  #     include ActiveModel::Model
+  #     include ActiveModel::Attributes
+  #
+  #     attribute :name, :string
+  #     attribute :age, :integer
+  #     attribute :password, :string
+  #   end
+  #
+  #   person = Person.new(name: "Alice", age: 30, password: "secret")
+  #   person.inspect
+  #   # => "#<Person name: \"Alice\", age: 30, password: \"secret\">"
+  #
+  #   Person.filter_attributes = [:password]
+  #   person.inspect
+  #   # => "#<Person name: \"Alice\", age: 30, password: [FILTERED]>"
+  #
+  #   Person.attributes_for_inspect = [:name]
+  #   person.inspect
+  #   # => "#<Person name: \"Alice\">"
+  #
+  module Inspection
+    extend ActiveSupport::Concern
+
+    included do
+      class_attribute :attributes_for_inspect, instance_accessor: false, default: :all
+    end
+
+    module ClassMethods
+      # Returns an array of attribute names whose values should be masked in
+      # the output of +inspect+.
+      #
+      #   Person.filter_attributes = [:password]
+      #   Person.filter_attributes # => [:password]
+      def filter_attributes
+        if defined?(@filter_attributes)
+          @filter_attributes
+        elsif superclass.respond_to?(:filter_attributes)
+          superclass.filter_attributes
+        else
+          []
+        end
+      end
+
+      # Specifies attributes whose values should be masked in the output of
+      # +inspect+.
+      #
+      #   Person.filter_attributes = [:password]
+      #   Person.new(password: "secret").inspect
+      #   # => "#<Person password: [FILTERED]>"
+      def filter_attributes=(attributes)
+        @inspection_filter = nil
+        @filter_attributes = attributes
+      end
+
+      # Returns the inspection filter used to mask sensitive attributes.
+      def inspection_filter # :nodoc:
+        if defined?(@filter_attributes)
+          @inspection_filter ||= begin
+            mask = InspectionMask.new(ActiveSupport::ParameterFilter::FILTERED)
+            ActiveSupport::ParameterFilter.new(@filter_attributes, mask: mask)
+          end
+        elsif superclass.respond_to?(:inspection_filter)
+          superclass.inspection_filter
+        else
+          @inspection_filter ||= begin
+            mask = InspectionMask.new(ActiveSupport::ParameterFilter::FILTERED)
+            ActiveSupport::ParameterFilter.new([], mask: mask)
+          end
+        end
+      end
+    end
+
+    # Returns the attributes of the model as a nicely formatted string.
+    #
+    #   Person.new(name: "Alice", age: 30).inspect
+    #   # => "#<Person name: \"Alice\", age: 30>"
+    #
+    # The attributes can be limited by setting +.attributes_for_inspect+.
+    #
+    #   Person.attributes_for_inspect = [:name]
+    #   Person.new(name: "Alice", age: 30).inspect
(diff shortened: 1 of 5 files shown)


activemodel/lib/active_model/inspection.rb
```diff
@@ -0,0 +1,140 @@
...
+  #   person.inspect
+  #   # => "#<Person name: \"Alice\", age: 30, password: [FILTERED]>"
+  #
+  #   Person.attributes_for_inspect = [:name]
+  #   person.inspect
+  #   # => "#<Person name: \"Alice\">"
+  #
+  module Inspection
+    extend ActiveSupport::Concern
+    include ActiveSupport::Inspection
```
seanpdoyle: Is it significant that this line is outside of an [included](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Concern.html#method-i-included) block that's typical of `ActiveSupport::Concern` modules?

```suggestion
    included do
      include ActiveSupport::Inspection
    end
```

activesupport/lib/active_support/inspection.rb
```diff
@@ -0,0 +1,148 @@
...
+  # to be included by both ActiveModel and ActiveRecord to share inspection
+  # logic.
+  #
+  # == Usage
+  #
+  # Include this module and implement the required methods:
+  #
+  # * +attribute_names+ - Returns an array of attribute names to inspect
+  # * +_read_attribute_for_inspection(name)+ - Returns the value of the attribute
+  # * +_has_attribute_for_inspection?(name)+ - Returns true if the attribute exists
```
seanpdoyle: The class-level concept of `attribute_names` is directly tied to [ActiveModel::Attributes](https://edgeapi.rubyonrails.org/classes/ActiveModel/Attributes.html#method-i-attribute_names). I'm unsure if this interface is applicable outside of Active Model. 

In particular, the `_`-prefixed attribute methods indicate that they aren't for public consumption.

In an effort to reduce the surface area of this changeset, are you open to inlining the `ActiveSupport::Inspection` implementation back into `ActiveModel::Inspection`?

Based on feedback from Core (and other code review feedback), a second PR that extracts `ActiveSupport::Inspection` on its own merits might be more viable than a single combined PR.

seanpdoyle (changes requested):

activerecord/lib/active_record/core.rb
```diff
@@ -9,6 +9,7 @@ module ActiveRecord
   module Core
     extend ActiveSupport::Concern
     include ActiveModel::Access
+    include ActiveSupport::Inspection
```
seanpdoyle: Based on https://github.com/rails/rails/pull/56521#discussion_r2662164330, this could be replaced with `ActiveModel::Inspection`:

```suggestion
    include ActiveModel::Inspection
```

activesupport/lib/active_support/inspection.rb
viniciusalmeida: > In an effort to reduce the surface area of this changeset, are you open to inlining the ActiveSupport::Inspection implementation back into ActiveModel::Inspection?

Do you mean keeping the PR simple as the https://github.com/rails/rails/pull/56521/commits/47efff7b660ac9e23ca73ab27543085639eebee9 commit, right? Yes, I'm open to that! I'll update it soon

eileencodes:
Hi @viniciusalmeida thanks for opening a PR.

We rejected this feature a year and a half ago https://github.com/rails/rails/pull/52904 because we don't want ActiveModel to be a non-database version of ActiveRecord. Can you elaborate on your use case?

viniciusalmeida:
Hi @eileencodes. Thank you for pointing out this PR! I wasn't aware of it - didn't reach it in my previous research.

I think the concern about ActiveModel becoming "a non-database version of ActiveRecord" misses how ActiveModel is commonly used in practice. I (and IMO many RoR developers) use it as representation models for things that don't map directly to database rows but still benefit from Rails' validation and attribute casting features - form objects, command objects, API request representations, etc.

The problem is that once we use ActiveModel::Attributes, the console experience becomes a little "hostile". Instead of seeing the logical model - the carefully defined attributes - we see implementation details like `@attributes=#<ActiveModel::AttributeSet:0x...>`. To me, this is a suboptimal developer experience, especially when debugging forms or validations in the console and this was what motivated me to propose this change.

From my perspective, this isn't about reducing the functionality gap between ActiveRecord and ActiveModel - it's about completing an existing ActiveModel feature. `ActiveModel::Attributes` already exists; inspection just improves its developer experience.

Do you have different thoughts about it? Or any other frame I can use to reach the goal I described?

eileencodes:
I asked in our core channel to double check on whether we changed our opinion on this. Ultimately the problem is this is a straight copy of Active Record's `inspect` functionality over to Active Model. For us to consider this feature it would need to be designed and implemented for Active Model, taking AM specifics into account.

viniciusalmeida:
Hi @eileencodes, thanks for the feedback! I understand the concern about this being too much of a straight copy.

But the similarity comes from solving the same core problem: both `ActiveModel::Attributes` and `ActiveRecord::Base` have an attributes hash that needs readable formatting in the console.

Would a more focused implementation (just `inspect` + `pretty_print`) be acceptable? I'm happy to redesign if there are ActiveModel-specific concerns I'm missing.

rafaelfranca:
What we are trying to say is that while the feature is here, it is not properlly integrated on Active Model. For example, should `filter_attributes` be automatically set like we do with the Active Record version. If not, why not?

viniciusalmeida:
I understand your point, @rafaelfranca. Your example made me see the change from a perspective I didn't have before. I kinda overlooked the automatic configuration matter due to the information that `ActiveModel` is designed to work independently of Rails. So, using a Rails config was counter-intuitive until I understood how AR does that.

That said, I've already implemented a working version that automatically sets `ActiveModel::Inspection.filter_attributes` from `Rails.application.config.filter_parameters` via the railtie:
```ruby
# activemodel/lib/active_model/railtie.rb
initializer "active_model.set_filter_attributes" do
  config.after_initialize do
    ActiveModel::Inspection.filter_attributes += Rails.application.config.filter_parameters
  end
end
```
If this approach aligns with what the core team has in mind, I'd be happy to push it.

You also mentioned that this specific issue was an example - I'd be glad to hear if there are other blind spots. I'm open to discussing the underlying implications more deeply.

seanpdoyle:
Since this PR is a re-submission of a previously rejected one (https://github.com/rails/rails/pull/52904), it would benefit from _improving_ upon the proposal that preceded it.

With regard to https://github.com/rails/rails/pull/56521#issuecomment-3729746657, #52904 also handled automatic assignment in a Railtie by way of an `ActiveSupport.on_load(…)` block:

https://github.com/rails/rails/pull/52904/files#diff-dee19ffe3a90ab0d9a33232bd29b0c53243a975a5b7f44404fadea7afe1e2997

```ruby
initializer "active_model.set_filter_attributes" do
  ActiveSupport.on_load(:active_model_inspect) do
    ActiveModel::Inspect.filter_attributes += Rails.application.config.filter_parameters
  end
end
```

Since the code you've proposed deviates from that pattern by utilizing a `config.after_initialize { … }` block, it would be useful to communicate the rationale behind that decision.

Naming differences aside (`ActiveModel::Inspect` in #52904, `ActiveModel::Inspection` here), there are a variety of other differences. Are you able to share more information about how and why those differences exist?

viniciusalmeida:
You are absolutely right, @seanpdoyle. To be fair, I was not aware of this PR before Eileen mentioned it. And I agree that my changes could be made based on the progress made over there. I actually [checked the changes](https://patch-diff.githubusercontent.com/raw/rails/rails/pull/52904.patch) this afternoon and, after some tweaks, I was able to get the exact behaviour that solves my use case:
```ruby
class Person
  include ActiveModel::Model

  attribute :name, :string
  attribute :password, :string
end

person = Person.new(name: "Alice", password: "secret")
person # => #<Person name: "Alice", password: "secret">

Person.filter_attributes = [:password] # => [:password]
person # => #<Person name: "Alice", password: [FILTERED]>
```
_PS: railties doesn't set filter_attributes automatically because I ran the irb with `active_model` only (`irb -I activesupport/lib -I activemodel/lib -r active_model`)_

I don't have strong preferences about the name/load technique. We can use Petrick's implementation as a starting point if it makes more sense. In my approach, the main architectural difference is that `Inspect`/`Inspection` is coupled to `Attributes` — since the former doesn't make sense without the latter in the use case I mentioned a few comments ago. This is why I include it automatically when `Attributes` is included, rather than as a standalone module.

I would like to hear from @rafaelfranca or @eileencodes first if they (on behalf of the core team) see potential in this idea or any other blind spots.

PR #56522, opened 2026-01-05
Don't set Vary when forgery protection is skipped

djmb:
When request forgery protection is skipped, we should not add "Sec-Fetch-Site" to the Vary header.

See: https://github.com/djmb/rails/commit/f5a1915f1db22533f3ec5d401f41fb0f06338a79

cc @rosa


rosa (approved):
Ah, good catch! Thanks @djmb 🙏🏻

PR #56523, opened 2026-01-05
CI: Add `libvips` package to generated `ci.yml`

stevepolitodesign:
### Motivation / Background

Prior to this commit, the generated `ci.yml` file was missing the `libvips` package.

### Detail

This commit uses prior art to add `libvips` so long as the `--skip-active-storage` flag was not passed during app generation.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -632,6 +632,9 @@ module Rails
           packages << "python-is-python3"
         end
 
+        # ActiveStorage preview support
+        packages << "libvips" unless skip_active_storage?
+
         packages.compact.sort
       end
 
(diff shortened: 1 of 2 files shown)


railties/lib/rails/generators/app_base.rb
```diff
@@ -632,6 +632,9 @@ def dockerfile_build_packages
           packages << "python-is-python3"
         end
 
+        # ActiveStorage preview support
+        packages << "libvips" unless skip_active_storage?
```
stevepolitodesign, line 636: `dockerfile_build_packages` is called via `ci_packages`, which is called in here:

https://github.com/rails/rails/blob/cf6d46fb947103c7217fc13c6bd33e47de5b6c39/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt#L118-L120

Note that this is lifted from: 

https://github.com/rails/rails/blob/cf6d46fb947103c7217fc13c6bd33e47de5b6c39/railties/lib/rails/generators/app_base.rb#L611

PR #56524, opened 2026-01-06
Fix Rails.app.revision to check git repo in app root

Saidbek:
### Detail

In [#56490](https://github.com/rails/rails/pull/56490), `Rails.app.revision` would check for `.git` in the current working directory instead of the application root. This could cause issues when the application is run from a different directory.

### Checklist

### DIFF
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -406,9 +406,12 @@ module Rails
         @revision = begin
           root.join("REVISION").read.strip.presence
         rescue SystemCallError
-          if Dir.exist?(".git")
-            rev = `git rev-parse HEAD 2> /dev/null`.strip.presence
-            rev if $?.success?
+          r, w = IO.pipe
+          if system("git", "-C", root.to_s, "rev-parse", "HEAD", in: File::NULL, err: File::NULL, out: w)
+            r.read.strip
+          else
+            r.close
+            nil
           end
         end
         @revision_initialized = true


fxn:
Better!

The root directory could have spaces, for example, you cannot just interpolate.

byroot:
Thinking again about this. I think we shouldn't even check for `.git` because it's totally legitimate to have a Rails app in the subdirectory of the repo etc. So we should just try to invoke git and move on.

railties/lib/rails/application.rb
```diff
@@ -406,9 +406,12 @@ def revision
         @revision = begin
           root.join("REVISION").read.strip.presence
         rescue SystemCallError
-          if Dir.exist?(".git")
-            rev = `git rev-parse HEAD 2> /dev/null`.strip.presence
-            rev if $?.success?
+          r, w = IO.pipe
+          if system("git", "-C", root.to_s, "rev-parse", "HEAD", in: File::NULL, err: File::NULL, out: w)
+            r.read.strip
```
bquorning, line 411: For some reason, my process hangs on the `read` call, but e.g. `readline`, `read_nonblock(40)` etc. works fine.

I’m seeing this behavior both on MacOS (ARM) and CI (Ubuntu on ARM).

I think `read` _should_ work, but probably `read_nonblock` would be a better choice here?

bquorning: Ah, if you add a `w.close` before `r.read`, it works fine.

byroot: It's my fault I had the `w.close` but moved the code around and lost it. I'll fix it directly on master.

bquorning: Thank you @byroot 🙏🏼

byroot: Try f7008177810741cf992b0791b9dcfd2bf17abcc3

bquorning: Yep, that works. Thank you for the quick fix.

PR #56527, opened 2026-01-06
Remove Ruby < 3.3 compatibility code

yahonda:
### Motivation / Background

This commit removes Ruby < 3.3 compatibility code since Rails main branch requires Ruby 3.3.

### Detail

- No need to check `SecureRandom.method(:alphanumeric).parameters.size` in Ruby 3.3+ because it is always 2 https://github.com/ruby/securerandom/pull/36

- Ruby 3.3 introduces `Range#overlap?` https://bugs.ruby-lang.org/issues/19839
- Ruby 3.3 introcuces `Fiber#kill` https://bugs.ruby-lang.org/issues/595

### Additional information

Refer to https://github.com/rails/rails/pull/56511

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/range/overlap.rb
+++ b/activesupport/lib/active_support/core_ext/range/overlap.rb
@@ -1,40 +1,5 @@
 # frozen_string_literal: true
 
 class Range
-  unless Range.method_defined?(:overlap?) # Ruby 3.3+
-    # Compare two ranges and see if they overlap each other
-    #  (1..5).overlap?(4..6) # => true
-    #  (1..5).overlap?(7..9) # => false
-    def overlap?(other)
-      raise TypeError unless other.is_a? Range
-
-      self_begin = self.begin
-      other_end = other.end
-      other_excl = other.exclude_end?
-
-      return false if _empty_range?(self_begin, other_end, other_excl)
-
-      other_begin = other.begin
-      self_end = self.end
-      self_excl = self.exclude_end?
-
-      return false if _empty_range?(other_begin, self_end, self_excl)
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/core_ext/range/overlap.rb
```diff
@@ -1,40 +1,8 @@
...
-    private
-    def _empty_range?(b, e, excl)
-      return false if b.nil? || e.nil?
-
-      comp = b <=> e
-      comp.nil? || comp > 0 || (comp == 0 && excl)
-    end
-  end
-
+  # Compare two ranges and see if they overlap each other
```
yahonda: I am not 100% sure if we should keep these comments or not.

byroot:
Ah good catch, not sure how I missed it.

activesupport/lib/active_support/core_ext/range/overlap.rb
byroot: Do they show up in RDOC?

rafaelfranca (approved):

activesupport/lib/active_support/core_ext/range/overlap.rb
yahonda: > Do they show up in RDOC?

No.

https://yahonda-remove-ruby32-compat.rails-docs-preview.pages.dev/api/classes/Range.html

byroot: We can just remove them then.

yahonda: Removed.

PR #56529, opened 2026-01-06
Clarify that column default values don't apply to existing records [ci skip]

### LINKED ISSUE #56520: Potential Documentation Improvement: Migration default values behavior with existing records
Hello Rails team,

First, thank you for the excellent guides! They've been incredibly helpful for learning Rails.

I'm writing about a small issue I encountered while following the product inventory management section. When adding the inventory_count field to existing products, the migration with default: 0 didn't update my existing records - they remained as nil.

Here's what I did:

Created Products without inventory_count

Added migration: add_column :products, :inventory_count, :integer, default: 0

Existing products kept inventory_count: nil while new products correctly got 0

The example code product.inventory_count.positive? failed with NoMethodError for existing products

The code I used (from guide):

erb
<% if product.inventory_count.positive? %>
I temporarily fixed it with:

erb
<% if product.inventory_count.to_i.positive? %>
I might be missing something - perhaps there's a recommended way to handle this that I overlooked? If so, I'd appreciate guidance.

Suggestion: Could the documentation either:

Mention that existing records need manual updates, or

Use nil-safe approaches like &.positive? or .to_i.positive?

This might help other beginners who encounter the same situation.

Thank you for your time and for maintaining such valuable resources!

Best regards,

Berkay Aksu

yujiteshima:
### Summary

  Clarify in the migration guide that setting a `default` value on a column only applies to newly created records. Existing records will retain their current values, including `NULL`.

  ### Motivation

  This addresses confusion reported in #56520, where a developer expected `add_column :products, :inventory_count, :integer, default: 0` to update existing records. The current documentation doesn't explicitly mention this behavior, which can lead to unexpected `nil` values and errors like `NoMethodError` when calling methods on the column.

  ### Changes

  Added a clarifying sentence to the `default` column modifier documentation in the Active Record Migrations guide.

  Fixes #56520

### DIFF
--- a/guides/source/active_record_migrations.md
+++ b/guides/source/active_record_migrations.md
@@ -705,7 +705,8 @@ Column modifiers can be applied when creating or changing a column:
 * `default`      Allows to set a default value on the column. Note that if you
   are using a dynamic value (such as a date), the default will only be
   calculated the first time (i.e. on the date the migration is applied). Use
-  `nil` for `NULL`.
+  `nil` for `NULL`. The default value will only apply to new records;
+  existing records will retain their current values (including `NULL`).
 * `limit`        Sets the maximum number of characters for a `string` column and
   the maximum number of bytes for `text/binary/integer` columns.
 * `null`         Allows or disallows `NULL` values in the column.
--- a/guides/source/active_record_migrations.md
+++ b/guides/source/active_record_migrations.md
@@ -705,8 +705,8 @@ Column modifiers can be applied when creating or changing a column:
 * `default`      Allows to set a default value on the column. Note that if you
   are using a dynamic value (such as a date), the default will only be
   calculated the first time (i.e. on the date the migration is applied). Use
-  `nil` for `NULL`. The default value will only apply to new records;
-  existing records will retain their current values (including `NULL`).
+  `nil` for `NULL`. Depending on your database, existing records may not
+  receive the default value.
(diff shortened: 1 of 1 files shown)


harsh183:
I like seeing the more documentation around migration default behavior

maxfelsher:
I'm not sure the added language is always true. My understanding is that the `default` option for an added column is mostly equivalent to adding `DEFAULT <some value>` to the SQL column definition, which means that this is really a question about how database engines treat this situation. I don't have all the various engines set up to test, but some online searching suggests to me that some engines do apply a new column's default to existing rows in at least some situations.

It might be good to find out more information about the original developer's environment to determine how widespread this issue with existing rows might be.

harsh183:
What if we put a warning like

> WARNING: Adding a `default` type may not default empty values in pre-existing rows. Check your particular database's implementation of DEFAULT or consider writing an explicit backfill for pre-existing rows

yujiteshima:
Thank you both for the feedback!

@maxfelsher Good point about database-specific behavior. I wasn't aware that some databases might handle this differently.

@harsh183 Thanks for the suggestion! I've updated the wording to be less absolute.

I've revised the documentation to:

> Depending on your database, existing records may not receive the default value.

This acknowledges the database-dependent behavior while still alerting developers to the potential issue.

Let me know if this works better!

PR #56532, opened 2026-01-06
[ActiveSupport ContinuousIntegration] Add group filtering to CI with -g/--group flags and nesting support

harsh183:
Note: see newer attempt at https://github.com/rails/rails/pull/57947

### Motivation / Background

At my company, our `bin/ci` takes quite long to run so we internally implemented a group filter to only run parts of the CI relevant to the given change. This has also been helpful for us for distributing the CI workload across multiple cloud workers for our cloud CI. The goal of this PR is to upstream this feature to ActiveSupport since it can be helpful for other rails projects as well.

### Detail

Creates a new `group` block that checks if the CLI argument group name is applicable for the current block and runs the steps accordingly. I was thinking the syntax can look something like:

```rb
# Simple groups
group "lint" do
  step "RuboCop", "bin/rubocop"
  step "ESLint", "yarn", "lint"
end

# Nested groups
group "backend" do
  group "unit" do
    step "Models", "bin/rails test test/models"
  end

  group "integration" do
    step "API", "bin/rails test test/integration"
  end
end
```

I've also moved option parsing to ruby's built in `OptionParser` so that the code for the group flag and pre-existing fail fast option is handled nicely. This can make adding future options nicer as well.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Add group based filtering using `bin/ci -g/--group` to `ActiveSupport::ContinuousIntegration`.
+
+    Example:
+    ```ruby
+    group "lint" do
+      step "Rubocop", "bin/rubocop"
+      step "Biome", "bin/biome"
+    end
+    ```
+
+    *Harsh Deep*
+
 *   Add `ActiveSupport::CombinedConfiguration` to offer interchangeable access to configuration provided by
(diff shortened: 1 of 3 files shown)


harsh183:
https://github.com/rails/rails/pull/56774 had a similar idea for parallelism groups that actually implemented a superset of this syntax, though it doesn't have calling by groups like `bin/ci -g group_name` so I could try a fresh approach on top of that.

harsh183:
newer attempt here: https://github.com/rails/rails/pull/57947

PR #56538, opened 2026-01-07
Abort rails new when bundle install fails

khasinski:
### Summary

When `bundle install --quiet` fails during rails new (e.g., due to missing dependencies), the error is silently ignored and the generator continues, leading to confusing subsequent errors.
  
Fixes: #56313 

### Checklist

UPDATE: I haven't noticed that there already is https://github.com/rails/rails/pull/56314, feel free to close this one if that other PR is preferred

### DIFF
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -677,7 +677,7 @@ module Rails
       end
 
       def bundle_install?
-        !(options[:skip_bundle] || options[:pretend])
+        !(options[:skip_bundle] || options[:pretend] || ENV["RAILS_SKIP_BUNDLE"])
       end
 
       def depends_on_system_test?
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -711,7 +711,11 @@ module Rails
       end
 
       def run_bundle
-        bundle_command("install --quiet", "BUNDLE_IGNORE_MESSAGES" => "1") if bundle_install?
+        return unless bundle_install?
+
+        unless bundle_command("install --quiet", "BUNDLE_IGNORE_MESSAGES" => "1")
+          abort "Bundle install failed. Please run `bundle install` manually to see the error."
+        end
       end
 
       def run_javascript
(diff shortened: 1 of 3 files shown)


rafaelfranca:
I don't think this is the right fix. We don't want to raise an error when other bundler commands fail, only when the bundler install fails when the application is being generated. We should move this logic there.

khasinski:
Makes sense, let me rework it then

rafaelfranca:
CI is not broken, but at least we know the feature is working.

railties/lib/rails/generators/app_base.rb
```diff
@@ -711,7 +708,14 @@ def target_rails_prerelease(self_command = "new")
       end
 
       def run_bundle
-        bundle_command("install --quiet", "BUNDLE_IGNORE_MESSAGES" => "1") if bundle_install?
+        return unless bundle_install?
+
+        success = bundle_command("install --quiet", "BUNDLE_IGNORE_MESSAGES" => "1")
+
+        # Only abort on failure during new app generation, not template application
+        if !success && !options[:template]
```
rafaelfranca: Why is the template application relevant here? I think the tests that were failing was because of template application but it was more an problem in the test setup than in the logic. `bundle install` for template application should also not fail.

khasinski: Ah, sorry, I've misinterpreted your comment and thought it's actually my change and I'm supposed to fix it. Let me quickly roll it back!

rafaelfranca: I mean, you change caused the test to fail, but don't think it is the code that broke it. It is that your change uncovered a problem in the test setup that for some reason is telling to run `bundler install` when we should not do that in those tests. So I think you need to change the tests to make sure we can merge your code change.

khasinski: I think we fix the failing tests by stubbing `bundle_command` for them though 🤔

rafaelfranca: Yes. I think so as well. There are other tests that do it already, so let's try to do the same.

rafaelfranca: Ah, I think we use the `run_generator_instance` method that stub the `bundle_command`

khasinski: I think `run_generator_instance` stubs on a generator object, but this test forks a subprocess via `rails()` so we might need to monkey-patch the module before the fork happens. I've pushed it as a separate commit, but I'll be happy to do it in another way if you want

khasinski: It's green, but I'm not sure if all those changes are actually the way to go

PR #56542, opened 2026-01-07
Enable Arel::TreeManagers to use Table's engine

hmcguire-shopify:
Currently, calling `to_sql` on a `TreeManager` subclass will compile the Arel tree using `Table.engine` (always `ActiveRecord::Base`). This can cause issues when multiple database adapters are used, requiring an explicit engine to be passed to `to_sql`:

```
Arel::InsertManager.new(Model.arel_table).to_sql(Model)
```

This commit refactors the `TreeManager` to store the `Arel::Table` passed in the initializer so that it's engine can be used by default before falling back to `ActiveRecord::Base`.

Now,

```
Arel::InsertManager.new(Model.arel_table).to_sql
```

will use `Model` instead of `ActiveRecord::Base`.

### DIFF
--- a/activerecord/lib/arel/delete_manager.rb
+++ b/activerecord/lib/arel/delete_manager.rb
@@ -5,6 +5,8 @@ module Arel # :nodoc: all
     include TreeManager::StatementMethods
 
     def initialize(table = nil)
+      super
+
       @ast = Nodes::DeleteStatement.new(table)
     end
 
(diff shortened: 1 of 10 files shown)


eileencodes:
Can you write a test for this?

hmcguire-shopify:
> Can you write a test for this?

Done!

PR #56543, opened 2026-01-07
Fix has_one preload with limit in association scope

khasinski:
When using `preload()` on a `has_one` association that includes `limit()` in its scope, the LIMIT clause was incorrectly applied globally across all parent records rather than per-parent-record.

For example, with:

```ruby
has_one :last_comment, -> { order(id: :desc).limit(1) }
```

Preloading multiple posts would generate:

```sql
SELECT * FROM comments WHERE post_id IN (1, 2, 3) LIMIT 1
```

This returned only ONE comment total instead of one per post, causing some associations to incorrectly return nil.

The fix strips `limit` and `offset` clauses from the reflection scope when preloading singular associations (has_one, belongs_to), since these clauses don't make sense when applied globally to a batch query for multiple parent records.

Fixes #56276

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -298,7 +298,19 @@ module ActiveRecord
               scope.where!(reflection.type => model.polymorphic_name)
             end
 
-            scope.merge!(reflection_scope) unless reflection_scope.empty_scope?
+            unless reflection_scope.empty_scope?
+              # For singular associations (has_one, belongs_to), we must not apply
+              # LIMIT/OFFSET from the scope when preloading multiple records, as these
+              # clauses would be applied globally rather than per-parent-record.
+              # For example, `has_one :last_comment, -> { order(id: :desc).limit(1) }`
+              # would result in `WHERE post_id IN (1,2,3) LIMIT 1` returning only
+              # one record total instead of one per post.
+              if reflection.collection?
+                scope.merge!(reflection_scope)
+              else
+                scope.merge!(reflection_scope.except(:limit, :offset))
+              end
+            end
 
             if preload_scope && !preload_scope.empty_scope?
               scope.merge!(preload_scope)
--- a/activerecord/test/cases/associations/eager_test.rb
+++ b/activerecord/test/cases/associations/eager_test.rb
@@ -1751,6 +1751,74 @@ class EagerAssociationTest < ActiveRecord::TestCase
     assert_equal book, Cpk::Order.eager_load(:book).find_by(id: order.id).book
   end
 
+  test "preloading has_one association with scope that has limit" do
+    # Define a temporary model with has_one association using limit in scope
+    post_model = Class.new(ActiveRecord::Base) do
+      def self.name; "PostWithLimitedComment"; end
+      self.table_name = "posts"
+      self.inheritance_column = :disabled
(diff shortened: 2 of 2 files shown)


tylerwood-io:
This and #56319 are competing fixes for #56276 — same root cause, different approaches. I just reviewed #56319 as well, so sharing observations across both.

### Approach comparison

This PR restructures the `reflection_scope` merge in `build_scope` to branch on `reflection.collection?`, stripping `:limit` and `:offset` via `except` for singular associations. #56319 takes a post-merge approach with `scope.limit!(nil)`.

**Stripping `offset` is the right call.** #56319 only strips `limit`, but `offset` has the same global-vs-per-parent problem. If someone had `has_one :foo, -> { order(id: :desc).offset(5).limit(1) }`, the offset would also be applied globally in the batch query. Unusual, but correct to handle.

**The `except` approach avoids mutating the merged scope in-place.** `reflection_scope.except(:limit, :offset)` returns a new relation, leaving the cached `@reflection_scope` untouched. #56319's `scope.limit!(nil)` mutates the already-merged result, which is also safe since `build_scope` creates a fresh `scope` each invocation — but the `except` approach makes the intent more explicit.

### Correctness

The key insight both PRs share is that `LIMIT` on a `has_one` scope is redundant during preloading. The per-parent "take one" filtering already happens in Ruby — `Preloader::Association#load_records` (line ~208 on main):

```ruby
if reflection.collection? || entries.empty?
  entries << record
  assignments = true
end
```

For non-collection associations, only the first record per owner key is kept. The `ORDER BY` from the scope is preserved (neither PR strips it), so the correct record (e.g., "last by id desc") is first in the result set for each parent. Stripping `LIMIT` just removes the global constraint that was incorrectly limiting the total result count.

### Edge cases

1. **`has_one` with `limit(0)`**: Semantically odd, but stripping it is correct — `has_one` already returns at most one record, and a global `LIMIT 0` in the batch query would return nothing for all parents.

2. **`belongs_to` with limit/offset**: `reflection.collection?` is false for `belongs_to`, so limits would also be stripped here. This is correct — `belongs_to` scopes with limits would have the same global batch problem. Though in practice, scoped limits on `belongs_to` are extremely rare.

3. **Interaction with `preload_scope`**: The stripping happens during `reflection_scope` merge, before the separate `preload_scope` merge (lines 303–305 on main). A limit on a manually-passed preload scope would still be preserved. This is the right behavior.

4. **`has_many` is untouched**: The `reflection.collection?` guard ensures collection associations keep their full scope. Correct.

### Test coverage

Two tests (preload and includes) using anonymous `Class.new` models. The self-contained approach is nice for readability. A few additional cases that would strengthen coverage:

- **Multiple parents where one has no matching records**: Confirms the limit stripping still correctly returns nil for parents with no children, rather than "leaking" another parent's record.
- **`eager_load` control test**: Confirms that `eager_load` (which uses JOINs) remains unaffected — it was already correct per #56276 discussion.

For comparison, #56319 includes six tests covering combined associations and nil-record scenarios.

### CHANGELOG

Neither this PR nor #56319 includes a CHANGELOG entry. (Correction to my earlier comment on #56319 where I incorrectly stated this one did — apologies for the confusion.) Rails convention calls for CHANGELOG entries on behavior changes, even bug fixes. The preload SQL generation changes here, so an entry seems warranted.

### Overall

Both PRs are correct fixes. This one is slightly more defensive (handling `offset`) and makes the intent clearer by restructuring the merge flow. #56319 is more minimal (one line). The tradeoff is readability vs. surgical minimalism — likely a maintainer style preference.

PR #56549, opened 2026-01-08
refactor(postgresql): make lookup_cast_type private

seuros:
@rafaelfranca, lookup_cast_type is private now, but we still using internally in the schema_creation.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb
@@ -146,7 +146,8 @@ module ActiveRecord
         if value.is_a?(Proc)
           value.call
         else
-          value = column.cast_type.serialize(value)
+          cast_type = column.cast_type || lookup_cast_type(column.sql_type)
+          value = cast_type.serialize(value)
           quote(value)
         end
       end
@@ -218,9 +219,10 @@ module ActiveRecord
         end
       end
 
-      def lookup_cast_type(sql_type) # :nodoc:
-        type_map.lookup(sql_type)
-      end
+      private
+        def lookup_cast_type(sql_type)
+          type_map.lookup(sql_type)
+        end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


rafaelfranca (commented):
This isn't enough to solve that TODO. If we are calling that method still using `send` the TODO isn't solved. Most of this logic need to go away. See original commit.

https://github.com/rails/rails/commit/5179851f7f67e7625712f1cd264e0d4edc25bbd1 and subsequent http://github.com/rails/rails/commit/20f6edf931c83e4f68b691c4861ef47a3774f409.

seuros:
Thank you , i will do a full update on what is needed. 🙏🏼 

I'm in quest to update few activerecord record gem i maintain, so i try to clean up todos while i'm on it.

activerecord/lib/active_record/connection_adapters/abstract/quoting.rb
```diff
@@ -146,7 +146,8 @@ def quote_default_expression(value, column) # :nodoc:
         if value.is_a?(Proc)
           value.call
         else
-          value = column.cast_type.serialize(value)
+          cast_type = column.cast_type || lookup_cast_type(column.sql_type)
```
skipkayhil, line 149: I think the way to make it private again is to move it back into `quote_default_expression` here. When `column` is a `ColumnDefinition`, `cast_type` seems to always be `nil` so it falls back to `lookup_cast_type`.

---

Maybe even better, `cast_type` could be populated when creating a `ColumnDefinition`, which would make this method simpler. However, that has a downside of an additional query on PG for _every_ column even if the `lookup_cast_type` value is only used for a few `ColumnDefinition`s with default values.

rafaelfranca: Yeah, `cast_type` should never be `nil` when it get here. If `lookup_cast_type` needs to be public so `cast_type` is always populated I'm fine with it. The main problem was removed in https://github.com/rails/rails/pull/56484

PR #56553, opened 2026-01-08
Add read_and_delete method to ActiveSupport::Cache::Store

pekopekopekopayo:
### Motivation / Background

This Pull Request has been created because currently, ActiveSupport::Cache lacks a built-in atomic "read-and-delete" operation. Developers often need to retrieve a value and immediately invalidate it (e.g., for one-time tokens or flash-like data), but doing this with separate read and delete calls can lead to race conditions and requires multiple network round-trips. This PR introduces read_and_delete to provide a thread-safe, atomic (where supported), and performant way to consume cache entries.

### Detail

This Pull Request changes the ActiveSupport::Cache::Store by implementing a new read_and_delete method.

Key implementation details include:

Redis Optimization: In RedisCacheStore, it utilizes the native GETDEL command (available in Redis 6.2.0+) to ensure the operation is atomic and requires only a single network round-trip.

Universal Support: For other cache adapters (File, Memory, MemCache), it provides a fallback implementation that mimics the same behavior to ensure a consistent API across all stores.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/cache/redis_cache_store.rb
+++ b/activesupport/lib/active_support/cache/redis_cache_store.rb
@@ -192,6 +192,75 @@ module ActiveSupport
         end
       end
 
+      # Reads and deletes data from the cache, using the given key.
+      # If there is data in the cache with the given key, then that data is
+      # returned and the entry is atomically deleted. Otherwise, +nil+ is returned.
+      #
+      # This operation is atomic when using Redis 6.2.0 or later via the GETDEL
+      # command. For older Redis versions, it falls back to a non-atomic GET + UNLINK
+      # which may have race conditions under high concurrency.
+      #
+      # Note, if data was written with the <tt>:expires_in</tt> or
+      # <tt>:version</tt> options, both of these conditions are applied before
+      # the data is returned.
+      #
+      # ==== Options
+      #
+      # * +:namespace+ - Replace the store namespace for this call.
+      # * +:version+ - Specifies a version for the cache entry. If the cached
+      #   version does not match the requested version, the read will be treated
+      #   as a cache miss. The entry will be deleted and then restored if the
+      #   version doesn't match.
+      # * +:raw+ - Reads the raw value from the cache, bypassing deserialization.
+      #
+      # Other options will be handled by the specific cache store implementation.
+      #
+      # ==== Examples
+      #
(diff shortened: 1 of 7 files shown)


activesupport/lib/active_support/cache/redis_cache_store.rb
```diff
@@ -487,6 +556,13 @@ def supports_expire_nx?
           @supports_expire_nx = redis_versions.all? { |v| Gem::Version.new(v) >= Gem::Version.new("7.0.0") }
         end
 
+        def supports_getdel?
+          return @supports_getdel if defined?(@supports_getdel)
+
+          redis_versions = redis.then { |c| Array.wrap(c.info("server")).pluck("redis_version") }
+          @supports_getdel = redis_versions.all? { |v| Gem::Version.new(v) >= Gem::Version.new("6.2.0") }
+        end
```
byroot, line 564: I don't think we should bother checking for this. First because the feature is now over 5 years old, but also because it's quite a bit of overhead.

byroot (commented):
I'm a bit on the fence here because only redis really support it. I guess you could do it with FileStore and SolidCache.

But also I'm not sure it's really worth adding a new method. I wonder if it's not better as an option on the `delete` method.

pekopekopekopayo:
@byroot 

Thank you for the feedback and the great idea. I’ll go ahead and implement it as an option for the delete method as you suggested. Just to be sure, do you have any concerns or reservations about adding this functionality in general?

Sorry for the late reply

byroot:
> do you have any concerns or reservations about adding this functionality in general?

Yeah a bit. As I see it, `ActiveSupport::Cache` is meant to abstract the functionality of Memcached. Like it should stay a relatively simple cache. But nowadays many people use Redis as a cache and it has lots of extra functionality that isn't easily replicable in other backends, and that go beyond the use of a cache.

So I worry we're complexifying `AS::Cache` because some users do want the Redis API instead of just a simple cache.

Like your stated use case is:

> (e.g., for one-time tokens or flash-like data)

This doesn't sound like cache duty to me, so I'm thinking it's a bit out of scope for AS cache and you might as well use Redis directly.

pekopekopekopayo:
I understand. If this direction doesn't align with the project's goals, I completely respect that. Thank you so much for your feedback and for taking the time to explain

PR #56556, opened 2026-01-08
Fix postgresql/schema_dumper relation_name appending the wrong schema…

chiperific:
### Detail

Fixes https://github.com/rails/rails/issues/56555

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
@@ -154,6 +154,8 @@ module ActiveRecord
           def relation_name(name)
             if @dump_schemas.size == 1
               name
+            elsif name.include?(".")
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
```diff
@@ -154,6 +154,8 @@ def within_each_schema
           def relation_name(name)
             if @dump_schemas.size == 1
               name
+            elsif name.include?(".")
```
chiperific, line 157: NOTE: Postgres' `oid::regclass::text` returns:
* a schema-qualified name ONLY IF the table's schema is not in the current search path
* a plain name IF the table's schema is in the current search path

Since `self.schema_name` has been set via `within_each_schema`, any table names passed to `relation_name` should already be plain (meaning they're within the current search path) or have their schema already prepended.

PR #56560, opened 2026-01-08
Fix postgresql/schema_dumper to emit all foreign_keys after all tables

chiperific:
### Detail

Fixes https://github.com/rails/rails/issues/56557

#### Fix PostgreSQL schema dumping to emit all foreign keys after all tables.
It appears that as of Rails 8, the schema is sorted/organized by postgres schemas in the order they appear in the `schema_search_path`. 

Tables and foreign keys were grouped together within each schema, which lead to the risk of emitting a foreign_key statement before the reference table was emitted. This issue causes an `ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:` exception to be raised.

The fix ensures we emit all tables first and then emit all foreign keys later.

### Additional information

I preserved the existing `within_each_schema` loop pattern, but I think it could be simpler to revert and just ensure that schema names are appended to table names as available.

We had a patch that fixed the pre-8 behavior, the logic of which could be used in place of the `within_each_schema` block pattern:

        module SchemaDumperPatch
          def self.prepended(base)
            base.class_eval do
              private

              def tables(stream)
                # pull schema names and table names, not just table names
                table_query = <<-SQL
                  SELECT schemaname, tablename
                  FROM pg_tables
                  WHERE schemaname = ANY(current_schemas(false))
                SQL

                # prepend the schema name to the table name
                sorted_tables = @connection.exec_query(table_query, "SCHEMA").map do |table|
                  "#{table["schemaname"]}.#{table["tablename"]}" unless ignored?(table["tablename"])
                end.compact.sort

                sorted_tables.each do |table_name|
                  table(table_name, stream)

                  # Rails 7.2.0.beta1 stops adding a new line character after the table definition
                  # https://github.com/rails/rails/commit/89addc7a6eb8d5a33ede18baf08260d15a3194c0
                  stream.print "\n"
                end

                if @connection.supports_foreign_keys?
                  sorted_tables.each do |tbl|
                    tbl_wo_schema = tbl.split(".").last
                    foreign_keys(tbl, stream) unless ignored?(tbl_wo_schema)
                  end
                end
              end
            end
          end
        end

        ActiveRecord::SchemaDumper.prepend SchemaDumperPatch

### Checklist


chiperific:
Depends on https://github.com/rails/rails/pull/56556

chiperific:
@fatkodima, would appreciate your review as it's based on your changes in https://github.com/rails/rails/commit/e841859e7b9f781c6dc5e29c73c18c6dd229c8d5

rafaelfranca:
Would not this PR change the order of foreign key definitions even for database that don't have different schemas like postgresql? I'd like to avoid that. I understand there are technical reasons why we can't have foreign keys defined after the table definition for postgresql but we don't have that requirement for sqlite and it is nice to find all information related to a table close together than having to go to a different part of the file to find the foreign keys

chiperific:
> Would not this PR change the order of foreign key definitions even for database that don't have different schemas like postgresql? I'd like to avoid that. I understand there are technical reasons why we can't have foreign keys defined after the table definition for postgresql but we don't have that requirement for sqlite and it is nice to find all information related to a table close together than having to go to a different part of the file to find the foreign keys

Sure, I can isolate this more specifically to the postgres adapter.

activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb
```diff
@@ -157,7 +163,7 @@ def relation_name(name)
             elsif name.include?(".")
               name  # Already schema-qualified, don't add another prefix
             else
-              "#{schema_name}.#{name}"
+              [schema_name, name].compact.join(".")
```
chiperific, line 166: Handles the case where `schema_name` is nil

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -132,26 +141,21 @@ def virtual_tables(stream)
       end
 
       def tables(stream)
-        sorted_tables = @connection.tables.sort
-
-        not_ignored_tables = sorted_tables.reject { |table_name| ignored?(table_name) }
```
chiperific, line 137: Created `not_ignored_tables` method to hold this logic.

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -132,26 +141,21 @@ def virtual_tables(stream)
...
           stream.puts if index < not_ignored_tables.count - 1
         end
+      end
 
-        # dump foreign keys at the end to make sure all dependent tables exist.
-        if @connection.supports_foreign_keys?
-          foreign_keys_stream = StringIO.new
-          not_ignored_tables.each do |tbl|
-            foreign_keys(tbl, foreign_keys_stream)
-          end
```
chiperific, line 149: Moved to `foreign_keys` method.

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -132,26 +141,21 @@ def virtual_tables(stream)
...
+      def tables_and_foreign_keys(stream)
+        not_ignored_tables.each_with_index do |table_name, index|
+          table(table_name, stream)
 
-          foreign_keys_string = foreign_keys_stream.string
-          stream.puts if foreign_keys_string.length > 0
+          if @connection.supports_foreign_keys?
+            foreign_keys(stream, [table_name])
+          end
```
chiperific, line 156: Moved inside the `not_ignored_tables.each_with_index` to avoid having to iterate over it twice

activerecord/lib/active_record/schema_dumper.rb
```diff
@@ -314,7 +322,19 @@ def check_parts(check)
         check_parts
       end
 
-      def foreign_keys(table, stream)
+      def foreign_keys(stream, tables = not_ignored_tables)
```
chiperific, line 325: The only reason to add the `tables` attr is to avoid having to iterate over `not_ignored_tables` twice.

I don't love it because it makes `foreign_keys()` feel too similar to `foreign_keys_for_table()`, but the main difference is that `foreign_keys()` can be the place where we create the stream, and I do like that part.

activerecord/test/cases/adapters/postgresql/dbconsole_test.rb
```diff
@@ -10,10 +10,14 @@ class PostgresqlDbConsoleTest < ActiveRecord::PostgreSQLTestCase
 
       ENV_VARS = %w(PGUSER PGHOST PGPORT PGPASSWORD PGSSLMODE PGSSLCERT PGSSLKEY PGSSLROOTCERT PGOPTIONS)
 
-      def run(*)
-        preserve_pg_env do
-          super
-        end
```
chiperific, line 16: This approach didn't properly clean up environment variables in time for the leak checker

activerecord/test/cases/adapters/postgresql/schema_test.rb
```diff
@@ -85,7 +85,7 @@ class Album < ActiveRecord::Base
     def self.default_scope; where(deleted: false); end
   end
 
-  def setup
+  setup do
```
chiperific, line 88: bring `setup` in line with `teardown`

activerecord/test/cases/schema_dumper_test.rb
```diff
@@ -483,7 +483,7 @@ def test_schema_dump_keeps_id_false_when_id_is_false_and_unique_not_null_column_
     assert_match %r{create_table "string_key_objects", id: false}, output
   end
 
-  if ActiveRecord::Base.lease_connection.supports_foreign_keys?
+  if ActiveRecord::Base.lease_connection.requires_referential_integrity_at_definition?
```
chiperific, line 486: As @rafaelfranca pointed out, SQLite supports foreign keys but doesn't need its foreign keys dumped at the bottom.

chiperific:
Would like a review please so this can hopefully be included in the next bug fix release.

chiperific:
@rafaelfranca, can you make some time to review this? Thanks!

chiperific:
Pinging ppl who have recently merged ActiveRecord-tagged PRs:
@fatkodima @byroot @skipkayhil @kamipo @rafaelfranca @eileencodes @yahonda

eileencodes:
Please don't ping anyone who merged an Active Record PR in the last few weeks, we have a lot on our plates.

chiperific:
> Please don't ping anyone who merged an Active Record PR in the last few weeks, we have a lot on our plates.

Sorry, was just hoping the squeaky wheel could get some grease.

chiperific:
@rafaelfranca, trying not be a pest, but would love to see this get included in the next release.

chiperific:
Seems to be no interest.

PR #56561, opened 2026-01-08
Deprecate protect_from_forgery without explicit strategy

### LINKED ISSUE #56537: Unify the default strategy for CSRF protection
### Steps to reproduce

When `protect_from_forgery` is called with no argument, [`:null_session`](https://github.com/rails/rails/blob/f7008177810741cf992b0791b9dcfd2bf17abcc3/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L268) is used, while by default every controller gets [`:exception`](https://github.com/rails/rails/blob/f7008177810741cf992b0791b9dcfd2bf17abcc3/actionpack/lib/action_controller/railtie.rb#L106-L108).

### Expected behavior

I'd expect both defaults to be the same.

### Actual behavior

Defaults are different, depending which part of the source code you are reading and might cause trouble when relying on one or the other default.

### System configuration

**Rails version**: edge

**Ruby version**: 3.4

### Notes

I don't know what is the better default, but I think it's worth unifying them. It's also worth mentioning the latest work done on https://github.com/rails/rails/pull/56350.

Saidbek:
### Motivation / Background

When `protect_from_forgery` is called without arguments, it defaults to `:null_session`. However, when `config.action_controller.default_protect_from_forgery` is enabled, it calls `protect_from_forgery with: :exception`. This inconsistency can cause confusion when developers rely on one behavior or the other.

Resolves https://github.com/rails/rails/issues/56537

### Checklist

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -60,6 +60,7 @@ Below are the default values associated with each target version. In cases of co
 
 #### Default Values for Target Version 8.2
 
+- [`config.action_controller.default_protect_from_forgery_with`](#config-action-controller-default-protect-from-forgery-with): `:exception`
 - [`config.action_controller.forgery_protection_verification_strategy`](#config-action-controller-forgery-protection-verification-strategy): `:header_only`
 - [`config.active_job.enqueue_after_transaction_commit`](#config-active-job-enqueue-after-transaction-commit): `true`
 - [`config.active_record.postgresql_adapter_decode_bytea`](#config-active-record-postgresql-adapter-decode-bytea): `true`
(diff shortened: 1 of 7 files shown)


guides/source/configuring.md
```diff
@@ -2031,6 +2031,17 @@ The default value depends on the `config.load_defaults` target version:
...
 
+#### `config.action_controller.default_protect_from_forgery_with`
+
+Configures the default strategy used when calling `protect_from_forgery` without the `:with` option.
+Defaults to `:null_session`, but will change to `:exception` in a future version of Rails.
+
+Applications can opt into the new behavior early by setting:
+
+```ruby
+config.action_controller.default_protect_from_forgery_with = :exception
```
rafaelfranca, line 2043: Let's add `default_protect_from_forgery_with = :exception` to the `new_framework_default` file as well so new applications get the new behavior already

PR #56565, opened 2026-01-09
Revert "Sort schema cache columns and indexes per table when dumping"

byroot:
Reverts rails/rails#54960

I'm tempted to revert this because while I can get behind sorting `schema.rb` to reduce git conflicts, the schema cache is more of an issue.

By sorting the schema cache we're changing the behavior of some queries such as `pluck("*")` etc, which can lead to bugs (it's not an hypothetical, I just ran into that situation).

For `schema.rb`, there is an escape hatch to preserve the order if you care about it, however for the schema cache, there's none.

cc @flavorjones I wonder if you have a strong opinion here.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/schema_cache.rb
+++ b/activerecord/lib/active_record/connection_adapters/schema_cache.rb
@@ -267,10 +267,10 @@ module ActiveRecord
       end
 
       def encode_with(coder) # :nodoc:
-        coder["columns"]          = @columns.sort.to_h.transform_values { _1.sort_by(&:name) }
+        coder["columns"]          = @columns.sort.to_h
         coder["primary_keys"]     = @primary_keys.sort.to_h
         coder["data_sources"]     = @data_sources.sort.to_h
-        coder["indexes"]          = @indexes.sort.to_h.transform_values { _1.sort_by(&:name) }
+        coder["indexes"]          = @indexes.sort.to_h
         coder["version"]          = @version
       end
 
--- a/activerecord/test/cases/connection_adapters/schema_cache_test.rb
+++ b/activerecord/test/cases/connection_adapters/schema_cache_test.rb
@@ -319,15 +319,11 @@ module ActiveRecord
         values = [["z", nil], ["y", nil], ["x", nil]]
         expected = values.sort.to_h
 
-        named = Struct.new(:name)
-        named_values = [["z", [named.new("c"), named.new("b")]], ["y", [named.new("c"), named.new("b")]], ["x", [named.new("c"), named.new("b")]]]
-        named_expected = named_values.sort.to_h.transform_values { _1.sort_by(&:name) }
-
         coder = {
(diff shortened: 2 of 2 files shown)


flavorjones:
@byroot I would be very interested to learn what sort of application code is depending on the schema cache sort order! Clearly a failure of imagination on my part, so I do want to understand the use case more deeply.

#54960 was opened because the schema cache in our app is (was) being committed to revision control to avoid hitting the database during boot. And the original feature request #42717 wanted it so that a file digest could be used to detect db changes. So I think there are at least a couple of reasons to have the sort order be stable in this file.

If you'd like this behavior to be configurable, I would be happy to do the work today to add that. If you're also OK with that, do you think applications should have to opt-in or opt-out of the sort behavior? (I'm proceeding with the assumption that the default should be "database order" and sorting should be opt-in.)

flavorjones:
@byroot See alternative https://github.com/rails/rails/pull/56566

byroot:
> I would be very interested to learn what sort of application code is depending on the schema cache sort order!

I haven't got to all of it just yet, as I monkey patched the old behavior to unlock the Rails 8.2 upgrade for now.

But from what I could see, the order change did change JSON serialization of the models, so the JSON document is "equivalent", but no longer strictly equal, which bust some cache keys and breaks a bunch of tests.

I'm not gonna merge this revert just yet, still figuring things out, but I'll keep you updated. Thanks for the reactivity.

byroot:
> the schema cache in our app is (was) being committed to revision control to avoid hitting the database during boot.

Right, I think I saw a few people doing that, it's not really the expected use of the schema cache though. (but who am I to judge 😛 ).

flavorjones:
> > the schema cache in our app is (was) being committed to revision control to avoid hitting the database during boot.
>
> Right, I think I saw a few people doing that, it's not really the expected use of the schema cache though

@byroot Hmm, I thought that "don't connect on boot if the schema cache is present" is the intent of some of @rafaelfranca's work e.g. 5179851f and c7328608

rafaelfranca:
> @byroot Hmm, I thought that "don't connect on boot if the schema cache is present" is the intent of some of @rafaelfranca's work e.g. [5179851](https://github.com/rails/rails/commit/5179851f7f67e7625712f1cd264e0d4edc25bbd1) and [c732860](https://github.com/rails/rails/commit/c7328608e55c1c18765f840548c13ac646b71988)

Yes, but it was never the intention to have the cache committed to the repository.

The way I see this cache should be generated when building the container for the application, or after a migration has run in production, but never committed in the repository since the committed cache would be equivalent to the development database, not the production database.

I don't think the schema cache should be sorted. We should merge this PR.

byroot:
Ok, I'm rebasing and I'll merge.

On a side note, I think for both these and schema.rb, we should make sure it easy to swap the implementation, if you need to. I do have some plan to optimize the schema cache dump, so might be the occasion whenever I get to it.

PR #56566, opened 2026-01-09
Add `config.active_record.schema_cache_dump_order`

flavorjones:
### Motivation / Background

In Rails 8.1 ([#54960](https://github.com/rails/rails/pull/54960) the schema cache dump file was changed to sort the columns and indices. Previously they had been emitted in "database order", which could vary across development environments depending on the order in which migrations were applied (leading to git commit churn and an inability to rely on a file digest to indicate changes).

In https://github.com/rails/rails/pull/54960#discussion_r2675996426 and https://github.com/rails/rails/pull/56565, @byroot pointed out that the change in sort order of the columns and indices in the schema cache dump introduced by #54960 could affect the results of some queries (e.g. `pluck("*")`). This changed first appeared in Rails 8.1.0.

### Detail

Introduce `config.active_record.schema_cache_dump_order` to control the order of columns and indexes in schema cache dump files.

When set to `:database` (the default), columns and indexes are written in the order returned by the database. When set to `:sorted`, they are sorted alphabetically for deterministic output.

This configuration option now requires developers to opt-in to sort the schema cache dump.

### Additional information

Original feature request and discussion was at https://github.com/rails/rails/issues/42717.

The pull request https://github.com/rails/rails/pull/48824 partially solved this problem by sorting the tables by name.

This is an alternative to #56565 which simply reverts the original change.

### Checklist


flavorjones:
Test failures seems unrelated.

byroot:
As mentioned on the other PR, I'm still figuring things out. But at this point I think this makes sense at least for people who use `structure.sql`. 

But again, no rush, we'll see.

jasonperrone:
@byroot well I wouldn't say "no rush". The schema order change didn't take into account that some people use Rails to dump database data from one DB, dump the schema, create a new DB from the schema and then try to import that data only to find out that the columns in the dump data don't match the columns of the tables in the database.

flavorjones:
@jasonperrone Can you say more? This PR is about the order in the schema cache file (e.g., `db/schema_cache.yml`), which to my knowledge isn't used for schema dump or restore operations (which more closely concern `db/schema.rb` or `db/structure.sql`).

The schema cache file is a runtime optimization that caches column and index metadata. It's not used to create tables or define database structure. But if I'm misunderstanding your comment, I'd be grateful if you helped me better understand what was broken by #54960.

jasonperrone:
@flavorjones Interesting. It's possible this is the wrong PR, but in Rails 8.1.2, rails db:schema:dump absolutely creates a schema.rb where the columns of each table are in alphabetical order, not in the database's order. I didn't find another PR about sorting column names alphabetically.

jasonperrone:
@flavorjones Ok, this pr https://github.com/rails/rails/pull/55414 is the actual problem and as you can see others complained and an upcoming release will have this functionality as opt-in. So, this was the wrong PR.

PR #56567, opened 2026-01-09
Disable file watcher for console

ngan:
### Motivation / Background

I noticed that when I start Rails development console, it would start watching files for updates. Dev console is an environment that supports reloading but doesn't get automatically reloaded upon change. The goal of this change is to provide a way for things that boot the application to disable the file watcher. I asked on the Rails forum and @fxn replied: https://discuss.rubyonrails.org/t/why-does-rails-console-need-to-watch-for-file-changes/90083

Note: I did not update any tests because I wanted to get feedback regarding the approach.

### Detail

This adds a `DisabledFileUpdateChecker` class that is simply a no-op file watcher class. It also adds a configuration on the Application called `disable_file_watcher` (similar to `disable_sandbox`). When set to `true`, an initializer in the `Finisher` will overwrite the `config.file_watcher` and set it to the `DisabledFileUpdateChecker`.

I took this approach because:
1. I didn't want to mess with any near the concept of "reloading." As in, I didn't want to add more logic to `enable_reloading?` since Console does, in fact, support reloading and changing that seems weird and could have other side effects later.
2. I'm trying to separate out the idea of "watch for changed files and automatically reload" vs "reloadability".
3. I found a handful of places in the framework that will always initialize the file watcher class regardless of `enable_reloading?`.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/disabled_file_update_checker.rb
+++ b/activesupport/lib/active_support/disabled_file_update_checker.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module ActiveSupport
+  # = Null File Update Checker
+  #
+  # This class is used by \Rails when file watching is disabled in certain environments
+  # like console.
+  class DisabledFileUpdateChecker
+    def initialize(files, dirs = {}, &block)
+      unless block
+        raise ArgumentError, "A block is required to initialize a DisabledFileUpdateChecker"
+      end
+
+      @block = block
+    end
+
+    def updated?
+      false
+    end
+
+    def execute
+      @block.call
+    end
+
+    def execute_if_updated
+      false
+    end
+  end
+end
--- a/railties/lib/rails/application/finisher.rb
+++ b/railties/lib/rails/application/finisher.rb
@@ -4,11 +4,18 @@ require "active_support/core_ext/string/inflections"
 require "active_support/core_ext/array/conversions"
 require "active_support/descendants_tracker"
 require "active_support/dependencies"
+require "active_support/disabled_file_update_checker"
 
 module Rails
   class Application
     module Finisher
       include Initializable
+
+      initializer :disable_file_watcher do |app|
+        if app.config.disable_file_watcher
+          app.config.file_watcher = ActiveSupport::DisabledFileUpdateChecker
+        end
+      end
 
       initializer :add_generator_templates do
         ensure_generator_templates_added
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -17,7 +17,7 @@ module Rails
                     :force_ssl, :helpers_paths, :hosts, :host_authorization, :logger, :log_formatter,
                     :log_tags, :silence_healthcheck_path, :railties_order, :relative_url_root,
                     :ssl_options, :public_file_server,
-                    :session_options, :time_zone, :reload_classes_only_on_change,
+                    :session_options, :time_zone, :reload_classes_only_on_change, :disable_file_watcher,
                     :beginning_of_week, :filter_redirect, :x,
                     :content_security_policy_report_only,
                     :content_security_policy_nonce_generator, :content_security_policy_nonce_directives,
@@ -60,6 +60,7 @@ module Rails
         @relative_url_root                       = ENV["RAILS_RELATIVE_URL_ROOT"]
         @reload_classes_only_on_change           = true
         @file_watcher                            = ActiveSupport::FileUpdateChecker
+        @disable_file_watcher                    = true
         @exceptions_app                          = nil
         @autoflush_log                           = true
         @log_formatter                           = ActiveSupport::Logger::SimpleFormatter.new
--- a/railties/lib/rails/commands/console/console_command.rb
+++ b/railties/lib/rails/commands/console/console_command.rb
@@ -86,6 +86,9 @@ module Rails
 
       desc "console", "Start the Rails console"
       def perform
+        require_application!
+        Rails.application.config.disable_file_watcher = true
+
         boot_application!
 
         wrap_with_executor = !options[:skip_executor]


activesupport/lib/active_support/disabled_file_update_checker.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module ActiveSupport
+  # = Null File Update Checker
+  #
+  # This class is used by \Rails when file watching is disabled in certain environments
+  # like console.
+  class DisabledFileUpdateChecker
```
byroot, line 8:
```suggestion
  class NullFileUpdateChecker
```

`Null<Impl>` is somewhat the common pattern in rails, (e.g. we got a `NullLock`, etc).

railties/lib/rails/application/finisher.rb
```diff
@@ -4,12 +4,19 @@
...
 require "active_support/dependencies"
+require "active_support/disabled_file_update_checker"
 
 module Rails
   class Application
     module Finisher
       include Initializable
 
+      initializer :disable_file_watcher do |app|
+        if app.config.disable_file_watcher
```
byroot, line 15: I think I'd just piggy back on the `config.file_watcher`.

e.g. 

```ruby
  if config.file_watcher.nil?
```

Or even just directly set `NullWatcher` from `railties/lib/rails/commands/console/console_command.rb`.

byroot (commented):
The feature make sense, but the implementation can be simplified further.

railties/lib/rails/application/finisher.rb
skipkayhil: Inside `Engine` we also have the `console` hook: https://github.com/rails/rails/blob/4a3bf7637129821ecb04ddb25870cdc0fd88c7dd/activerecord/lib/active_record/railtie.rb#L66-L67

ngan: @skipkayhil The console block runs _after_ initialization. As in, the `file_watcher` class is already [initialized in the Finisher](https://github.com/rails/rails/blob/main/railties/lib/rails/application/finisher.rb#L206).

@byroot I can't do it in the ConsoleCommand either because here's the sequence of events:
1. ConsoleCommand requires the application (via `require_application!`)
2. I get a chance to do whatever (I set `Rails.application.config.disable_file_watcher = true`)
3. ConsoleCommand initializes the application (via `boot_application!`)

In step 3 is where `file_watcher` is set during the initialization process, typically by the application itself in `config/environments/development.rb`. This is especially true for folks using the evented file watcher. They're instructed to do this in their environment file:

```ruby
  # ...
  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
  # ...
end
```

So any setting I make during step `2` on `config.file_watcher` will be overwritten during the initialization process. And I can't do anything _after_ (like in the `console` hook) is too late. This is why I had to make a whole new (private-ish?) configuration attribute to house the disabling option.

ngan (commented):
@byroot I think the ideal solution would be refining the [FileUpdateChecker api](https://api.rubyonrails.org/classes/ActiveSupport/FileUpdateChecker.html). Right now, the actual listening happens at initialization of the evented `file_watcher` class:
- https://github.com/rails/rails/blob/main/activesupport/lib/active_support/evented_file_update_checker.rb#L94

This also means the numerous places that instantiate their own listener they start "working" immediately and somehwhat out of our control.

It would be great if the watcher had a `#start` method that gets called by the framework when file watching should actually start, or in our case, never start. This means you're still able to configure your watcher and we don't replace it (like we do in this PR) unexpectedly, and we control whether or not they start doing the actual work of watching for changes.

Here are the places watchers are created:
- https://github.com/rails/rails/blob/main/railties/lib/rails/application/routes_reloader.rb#L52
- https://github.com/rails/rails/blob/main/actionview/lib/action_view/railtie.rb#L127
- https://github.com/rails/rails/blob/main/railties/lib/rails/application/finisher.rb#L206
- https://github.com/rails/rails/blob/main/activerecord/lib/active_record/migration.rb#L679

fxn:
Hey! I need to sit down and think about this patch.

I'll look at the current state of the code, but intuitively, I believe we should separate "reloading enabled" from "needing a file watcher". And intuitively I think of them as flags, rather than null objects. If there is nothing to watch, I would not set a null watcher, I would set nothing.

That, modulus I have not looked into it yet.

fxn:
I have not looked into it, but indeed I suspect needing a watcher is **the exception**, you do not need a watcher unless you run a service, think `runner` and other commands, they do not need a file watcher.

byroot:
> It would be great if the watcher had a `#start` method that gets called by the framework when file watching should actually start,

I like that.

fxn:
Shouldn't all this be happening already depending on `config.reload_classes_only_on_change`?

byroot:
> Dev console is an environment that supports reloading but doesn't get automatically reloaded upon change.

Actually, thinking a bit more about this. Isn't `rails server`, (and perhaps Spring?) the only case where we want reloading? I'm trying to think of other situations where it's worth it.

e.g. `rails runner` like don't need it, probably not most "rake" tasks, etc.

fxn:
@byroot agreed.

Same reason reloading is disabled by default in Zeitwerk: it is the exception, rather than the norm.

fxn:
I believe this is also telling us that the configuration should probably be designed to be per environment **and** command/task first-class. Reminds me of Rake tasks overriding eager loading, it feels like a hack right now.

byroot:
> the configuration should probably be designed to be per environment **and** command/task first-class.

To me it tells me more that there is two things here.

First configure the type of watcher to use if needed. This is per environment, and the current config is fine.

The second is when to start watching, which is a per command thing. That's why I think @ngan suggestion of having watchers stay inactive until we call `#start` on them makes sense to me.

fxn:
> First configure the type of watcher to use if needed. This is per environment, and the current config is fine.

But the matrix you need, in that case, is to know if you have to call `start`.

BTW I do not mean literally configuration necessarily, it may be a first-class, well-defined way to override.

fxn:
However, I am still on the idea that if reloading is enabled but Rails is not responsible for automatic reloads, then Rails does not need file watchers, ergo just do not create any. That would be my first option.

The fact that the combination reload + automatic is rare, may be a hint that we need to think in terms of the exceptions. Allow list, override to enable automatic but manual is default, whatever.

fxn:
Hey, I have an idea for a way to address this and similar issues. Would also be a proper solution for [#56212](https://github.com/rails/rails/pull/56212#issuecomment-3595182604), which is a consequence of that initializer that overrides a user setting, something I have always considered to be a suspicious hack.

I'll work on it and will write back.

PR #56569, opened 2026-01-10
Make lookup_cast_type private in PostgreSQL adapter

Saidbek:
Remove TODO comment and make lookup_cast_type private as planned after Rails 8.1 release.


guilleiguaran:
The TODO isn't really "move this method to the private section" but "make all the changes needed to be able to move this method to the private section."

See the comment in https://github.com/rails/rails/pull/56549

Saidbek:
Closing as there's already a draft pr to handle it

PR #56580, opened 2026-01-12
Make CSRF header-only protection compatible with local installs using HTTP

rosa:
### Motivation / Background

This came up in https://github.com/basecamp/fizzy/discussions/2186 and was addressed in https://github.com/basecamp/fizzy/pull/2291. This PR is the upstreamed version of that. 

In local installations that don't use HTTPS and where the app is accessed within a local network, requests won't be performed from [a secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Secure_Contexts#potentially_trustworthy_origins). In this case, the browser won't send the `Sec-Fetch-Site` header. This means non-GET requests will be rejected because CSRF protection will fail when using the header-only approach.

### Detail

With this change, we allow these requests with missing `Sec-Fetch-Site` headers if:

- They happen over HTTP
- The app is not configured to force SSL

The `Origin` check always happens in any case.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/request_forgery_protection.rb
+++ b/actionpack/lib/action_controller/metal/request_forgery_protection.rb
@@ -636,7 +636,8 @@ module ActionController # :nodoc:
 
       def verified_via_header_only?
         SAFE_FETCH_SITES.include?(sec_fetch_site_value) ||
-          (sec_fetch_site_value == "cross-site" && origin_trusted?)
+          (sec_fetch_site_value == "cross-site" && origin_trusted?) ||
+          (sec_fetch_site_value.nil? && !request.ssl? && !ActionDispatch::Http::URL.secure_protocol)
       end
 
       def verified_with_legacy_token?
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,21 @@
+*   Make CSRF header-only protection compatible with local installs using HTTP
+
+    In local installations that don't use HTTPS and where the app is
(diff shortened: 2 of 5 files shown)


actionpack/lib/action_controller/metal/request_forgery_protection.rb
```diff
@@ -636,7 +636,8 @@ def using_header_only_for_forgery_protection?
 
       def verified_via_header_only?
         SAFE_FETCH_SITES.include?(sec_fetch_site_value) ||
-          (sec_fetch_site_value == "cross-site" && origin_trusted?)
+          (sec_fetch_site_value == "cross-site" && origin_trusted?) ||
+          (sec_fetch_site_value.nil? && !request.ssl? && !ActionDispatch::Http::URL.secure_protocol)
```
byroot, line 640: Nitpick but I think it would be more readable as:

```ruby
case sec_fetch_site_value
when "same-origin", "same-site"
  true
when "cross-site"
  origin_trusted?
when nil
  request.ssl? && !ActionDispatch::Http::URL.secure_protocol
else
  false
end
```

It also saves on calling `sec_fetch_site_value` many times.

rosa: Good call! I'll submit a separate PR for that.

rosa: Ah, you already did it! https://github.com/rosa/rails/commit/4a3bf7637129821ecb04ddb25870cdc0fd88c7dd

Thank you! 🙏🏻

PR #56582, opened 2026-01-12
Fix PostgreSQL COMMENT ON INDEX for schema-qualified tables

eglitobias:
Fixes #56581

When `add_index` is called on a schema-qualified table (e.g. `"other_schema.widgets"`) with a `comment:` option, Active Record correctly creates the index in the target schema, but generates an unqualified `COMMENT ON INDEX` statement that fails.

PostgreSQL requires index comments to reference the index using its **schema-qualified name**. As a result, Active Record currently emits:

```sql
COMMENT ON INDEX "index_widgets_on_id" IS '...'
```

instead of:
```sql
COMMENT ON INDEX "other_schema"."index_widgets_on_id" IS '...'
```

This causes `PG::UndefinedTable` errors when commenting on indexes created with a specific schema name.

This change ensures that index comments use a schema-qualified index name when `add_index` is invoked on schema-qualified tables, preventing PostgreSQL errors and aligning comment generation with index creation.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -320,7 +320,7 @@ module ActiveRecord
             statements << schema_creation.accept(index_definition)
 
             if supports_comments? && !supports_comments_in_create?
-              statements << change_index_comment_sql(index_definition.index) if index_definition.index.comment.present?
+              statements << change_index_comment_sql(index_definition.index, table_name) if index_definition.index.comment.present?
             end
           end
         end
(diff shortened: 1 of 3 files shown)


eglitobias:
I’ve opened a follow-up PR (#56599) that includes this fix and extends it with a public API, tests, and documentation. If preferred, merging the follow-up PR would make this one redundant.

eglitobias:
Hi! Just a gentle ping in case this slipped through.
Happy to adjust the implementation or tests if needed.
@matthewd would you mind taking a quick look when you have a chance?
Thanks a lot!

Prana111025 (commented):
COMMENT ON INDEX "other_schema"."index_widgets_on_id" IS '...'

PR #56585, opened 2026-01-12
Improve the ScreenshotHelper docs

codergeek121:
This adds a small improvement to the `ActionDispatch::SystemTesting::TestHelpers` docs.

Without looking into the source it's not obvious on what to set the ENV variable `RAILS_SYSTEM_TESTING_SCREENSHOT_HTML` to activate HTML screenshots to  (only `1` works).

An alternative would be to use `ENV.key?("RAILS_SYSTEM_TESTING_SCREENSHOT_HTML")` instead, but this could be considered a breaking change.

### DIFF
--- a/actionpack/lib/action_dispatch/system_testing/test_helpers/screenshot_helper.rb
+++ b/actionpack/lib/action_dispatch/system_testing/test_helpers/screenshot_helper.rb
@@ -19,7 +19,7 @@ module ActionDispatch
         # different one with `Capybara.save_path`
         #
         # You can use the `html` argument or set the
-        # `RAILS_SYSTEM_TESTING_SCREENSHOT_HTML` environment variable to save the HTML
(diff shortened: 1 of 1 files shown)


p8:
We seem to use [`1`](https://github.com/rails/rails/blob/3a4961048ad251b50991ae83135d760a8a9e8ae3/activejob/test/integration/queuing_test.rb#L9), [`true`](https://github.com/rails/rails/blob/3a4961048ad251b50991ae83135d760a8a9e8ae3/railties/lib/rails/generators/testing/behavior.rb#L68) and [`nil?`](https://github.com/rails/rails/blob/3a4961048ad251b50991ae83135d760a8a9e8ae3/activejob/test/helper.rb#L12), so that isn't consistent.

p8:
Thanks @codergeek121 !

codergeek121:
Thank you @p8 :)

Issue #55773, opened 2025-09-26
Authentication generator does not remove old session records


easydatawarehousing:
The Rails authentication generator creates a session record when a user signs in and deletes the session record when the user signs out. If a user removes the session cookie the associated record becomes orphaned. When the user signs in again a new session record is created. Over time a lot of orphaned session records may accumulate.

One solution could be to remove old session records when the user signs in.

``` ruby
def start_new_session_for(user)
  user.sessions.create!(user_agent: request.user_agent, ip_address: request.remote_ip).tap do |session|
    Current.session = session
    cookies.signed.permanent[:session_id] = { value: session.id, httponly: true, same_site: :lax }

    # Cleanup old sessions for the current user
    # Uncomment next line to remove all sessions older than 1 month
    # user.sessions.where.not(id: session.id).where("created_at < ?", 1.month.ago).delete_all
    # Or: uncomment next line to remove all other sessions
    # user.sessions.where.not(id: session.id).delete_all
  end
end
```

Lines are all commented out so the developer can decide on which removal strategy to use.

MatheusRich:
If I understand correctly, if a user logs in in their desktop this would clear their session in their mobile phone for example. If that's the case, I don't think it's a good idea to have it by default.

> If a user removes the session cookie the associated record becomes orphaned.

How would that happen? Manually clearing cookies in the browser?

easydatawarehousing:
Yes this would log out the user from other sessions. That is why I propose to add code as comments, so the developer can decide which strategy to use.

> How would that happen? Manually clearing cookies in the browser?

For instance when the webbrowser is set up to remove all cookies when closing the browser. Moving to another computer.

MatheusRich:
Anyway, I encourage you to open the PR and let's see what the core team thinks.

jeromedalbert:
> If a user removes the session cookie the associated record becomes orphaned. [...] Over time a lot of orphaned session records may accumulate.

Having orphaned sessions seems to me like relatively normal behavior. Some websites like Facebook or Hey allow you to see all your open sessions. It's not uncommon to see stale (orphaned) sessions that happened for one reason or another (clearing cookies, using a difference browser or device, etc):

<img width="495" height="593" alt="Image" src="https://github.com/user-attachments/assets/6e43bd78-cd50-4910-8f02-9a88ba18d333" />

[...]

PR #56594, opened 2026-01-13
[ci skip] Update security guide: clean up orphaned sessions

easydatawarehousing:
Added a section to security guide describing why and how orphaned session records (used by authentication generator) may be cleaned up. See: https://github.com/rails/rails/issues/55773

### DIFF
--- a/guides/source/security.md
+++ b/guides/source/security.md
@@ -243,29 +243,12 @@ With the authentication generator configured as above, your application is ready
 for a more secure user authentication and password recovery process in just a
 few steps.
 
-NOTE: You may need to think about cleaning up orphaned session records.
-If a user deletes the cookie used for authentication (or the webbrowser is set
-up to remove cookies on exit) the user appears to be logged out but the session
-record still exists in the database.
-Depending on your requirements there are several solutions, like:
-
-- Display all other active sessions in the UI to the logged in user and let the
-  user decide which sessions to remove. You would need to implement the
-  necessary views, routes, and controller actions yourself
-- When a user starts a new session remove old sessions. You could add a line to
-  the end of the `start_new_session_for(user)` method. For instance:
-  `user.sessions.where("created_at < ?", 1.month.ago).delete_all`
-  or even `user.sessions.where.not(id: session.id).delete_all`
-- Regularly remove old session records using a rake task, like so:
-
-  ```
-  namespace :authentication do
-    desc "Purge historical sessions (might drop active sessions requiring users to log in again)"
-    task purge_historical: :environment do
-      Session.where(updated_at: ..1.month.ago).delete_all
-    end
-  end
-  ```
+NOTE: You may need to think about cleaning up stale session records. If a user
+deletes the cookie used for authentication (or the webbrowser is set up to remove
+cookies on exit) the user appears to be logged out but the session record still
+exists in the database. Depending on your requirements there are several
+solutions. A simple solution is to add a line to the 'start_new_session_for(user)'
+method: 'user.sessions.where("created_at < ?", 1.month.ago).delete_all'.
 
 Sessions
 --------
--- a/guides/source/security.md
+++ b/guides/source/security.md
@@ -244,9 +244,9 @@ for a more secure user authentication and password recovery process in just a
 few steps.
 
 NOTE: You may need to think about cleaning up stale session records. If a user
-deletes the cookie used for authentication (or the webbrowser is set up to remove
-cookies on exit) the user appears to be logged out but the session record still
-exists in the database. Depending on your requirements there are several
+deletes their cookies, or their browser removes them on exit, or the user never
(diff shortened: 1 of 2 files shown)


eric1234:
This explains everything very well but I'm wondering since it's such an edge case issue if it's adding too much documentation. [Here is the preview of the docs](https://easydatawarehousing-feature.rails-docs-preview.pages.dev/guides/security#session-management).

Maybe we are giving people too many options? I wonder if we could give the user the pointer in the right direction but more succinctly. Maybe just state they need to think about how to clean-up old sessions and give a the middle example since that shows both time based cleanup and sessions related to the user?

Also I find it a bit odd that it's in some sort of "Note" box for the opening paragraph but then the bullet list is outside the note:

image

BTW, I have no real authority here so take or leave my feedback as you see fit. Just offering two cents to help get the ticket closed out.

jeromedalbert:
> Also I find it a bit odd that it's in some sort of "Note" box for the opening paragraph but then the bullet list is outside the note

Yeah, I don't think `NOTE:` highlighted paragraphs support bullet points or code blocks. From memory, I haven't seen them in any Rails guides I've read.

> I'm wondering since it's such an edge case issue if it's adding too much documentation. [...] Maybe we are giving people too many options?

I don't know about "edge" but yeah, it's sort of a secondary case to possibly think about down the line. This could be explained in full-blown paragraphs and code blocks, but if this case is considered secondary, which I think it is, I agree with @eric1234 that this is a bit wordy for a `NOTE:` highlighted paragraph.

@easydatawarehousing For comparison in the linked [issue](https://github.com/rails/rails/issues/55773) you mentioned this current note:

https://github.com/rails/rails/blob/01fb87c25e541ea889263400012bb85e822496f5/guides/source/security.md#L105-L109

So I think it would be reasonable to say that your new note should be similarly sized? Your initial issue description suggested code comments that were similarly sized too. So maybe 3-4 higher-level sentences would be enough, using some terse yet expressive wordsmithing.

guides/source/security.md
```diff
@@ -243,6 +243,30 @@ With the authentication generator configured as above, your application is ready
 for a more secure user authentication and password recovery process in just a
 few steps.
 
+NOTE: You may need to think about cleaning up orphaned session records.
```
jeromedalbert: I don't know if everyone will understand the word "orphaned" for sessions. I think "old sessions" or "stale sessions" (my personal preference) would be a better umbrella term, covering not only the case of clearing cookies, but also the general case of never having explicitly logged out (from different browsers or devices where the user once logged in but then lost interest, or the device was abandoned/lost/reset/etc).

easydatawarehousing:
Thanks for all the feedback!

A 'note' paragraph is according to the [source](https://github.com/rails/rails/blob/main/guides/rails_guides/markdown/renderer.rb#L140) everything between the NOTE keyword and the first line-end. I agree that in the current commit the last sentence 'Depending ... like:' does not belong inside the note.

It would be best if either everything is inside the note, but text would have to be much shorter. Or alternatively make this a separate topic, but this might be a bit overkill for an "edge" case.

In my opinion the root cause of needing to expand the docs is the fact that the authentication generator is very barebones. It is a very nice solution and I use it all the time for internal tools with few users. But if you want to scale to lots of users, use cases and maybe reach a 'devise' like status much more code is needed. So a (rails-engine) gem with full support for sign up, password resets, 2FA, session management, etcetera. A lot of work, will probably never happen.

I guess, unless there are other suggestions, I will go for the short version inside a note block. And change the word orphaned for stale.

easydatawarehousing:
Changed to short (single note) version.

eric1234:
New wording LGTM!

guides/source/security.md
```diff
@@ -243,6 +243,13 @@ With the authentication generator configured as above, your application is ready
 for a more secure user authentication and password recovery process in just a
 few steps.
 
+NOTE: You may need to think about cleaning up stale session records. If a user
+deletes the cookie used for authentication (or the webbrowser is set up to remove
+cookies on exit) the user appears to be logged out but the session record still
+exists in the database. Depending on your requirements there are several
```
jeromedalbert: Optional suggestion:

```suggestion
deletes their cookies, or their browser removes them on exit, or the user never
explicitly logs out on their different devices, stale sessions will accumulate in
the database. Depending on your requirements there are several
```

easydatawarehousing: Thanks. Updated.

jeromedalbert:
You seem to have a rebase issue on your PR, it contains unrelated commits instead of a single commit that only contains your changes.

jeromedalbert:
@easydatawarehousing Are you still interested in moving this PR forward? It currently contains unrelated changes.

easydatawarehousing:
Sorry, my git skills are insufficient to fix this. I guess closing this PR is best thing to do.

PR #56598, opened 2026-01-13
For empty REVISION file, Rails.app.REVISION fallbacks to git fixes #56597

### LINKED ISSUE #56597: If REVISION file exists but is empty/whitespace, Rails.app.revision doesn't fall back to git.
When the REVISION file exists (even if empty), Rails reads from it and doesn't fall back to git. Rails should treat an empty REVISION file as "not present" and fall back to git. The presence of the REVISION file (even empty) prevents the git fallback. Only when the file doesn't exist, it fall back to git.

When the REVISION file exists but is empty, `.strip.presence` returns 'nil', which causes the `begin` block to return 'nil' (not fall through to git). The git command is only executed if there's a `SystemCallError` (file doesn't exist). Git fallback only happens when the REVISION file doesn't exist (SystemCallError).

### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "minitest"
end

require "minitest/autorun"
require "fileutils"
require "tmpdir"

  def setup
    @app_path = Dir.mktmpdir("rails_app_test")
    @original_dir = Dir.pwd

    FileUtils.mkdir_p("#{@app_path}/config")

    File.write("#{@app_path}/config/application.rb", <<~RUBY)
      require "rails"

      module TestApp
          config.eager_load = false
        end
      end
    RUBY

    File.write("#{@app_path}/config/environment.rb", <<~RUBY)
      require_relative "application"
      Rails.application.initialize!
    RUBY
  end

  def teardown
    Dir.chdir(@original_dir)
    FileUtils.rm_rf(@app_path)
  end

  def test_git_commit_sha_is_used_when_revision_file_is_empty
    Dir.chdir(@app_path) do
      File.write("#{@app_path}/REVISION", "")

      `git init`
      `git config user.email "test@example.com"`
      `git config user.name "Test User"`
      File.write("dummy.txt", "content")
      `git add .`
      `git commit -m "Initial commit"`

      commit_sha = `git rev-parse HEAD`.strip

      require "#{@app_path}/config/environment"
      assert_equal commit_sha, Rails.application.revision
    end
  end
end
```

### Expected behavior

Git fallback happens both when the file doesn't exist (SystemCallError) and when it exists but is empty/whitespace.

### Actual behavior

If REVISION file exists but is empty/whitespace, `.strip.presence` returns nil. No fallback to git in that case.

### System configuration

**Rails version**: main branch

**Ruby version**: 4.0.0

tahsin352:
### Motivation / Background

This Pull Request has been created to fix #56597. 

### Detail

When the REVISION file exists (even if empty), Rails reads from it and doesn't fall back to git. Rails should treat an empty REVISION file as "not present" and fall back to git.

This Pull Request fixes it. Now Git fallback happens both when the file doesn't exist (SystemCallError) and when it exists but is empty/whitespace.

Below test cases passes now.

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "minitest"
end

require "minitest/autorun"
require "fileutils"
require "tmpdir"

  def setup
    @app_path = Dir.mktmpdir("rails_app_test")
    @original_dir = Dir.pwd

    FileUtils.mkdir_p("#{@app_path}/config")

    File.write("#{@app_path}/config/application.rb", <<~RUBY)
      require "rails"

      module TestApp
          config.eager_load = false
        end
      end
    RUBY

    File.write("#{@app_path}/config/environment.rb", <<~RUBY)
      require_relative "application"
      Rails.application.initialize!
    RUBY
  end

  def teardown
    Dir.chdir(@original_dir)
    FileUtils.rm_rf(@app_path)
  end

  def test_git_commit_sha_is_used_when_revision_file_is_empty
    Dir.chdir(@app_path) do
      File.write("#{@app_path}/REVISION", "")

      `git init`
      `git config user.email "test@example.com"`
      `git config user.name "Test User"`
      File.write("dummy.txt", "content")
      `git add .`
      `git commit -m "Initial commit"`

      commit_sha = `git rev-parse HEAD`.strip

      require "#{@app_path}/config/environment"
      assert_equal commit_sha, Rails.application.revision
    end
  end
end
```

This PR further adds additional test cases:
1. "revision reads from git commit SHA when REVISION file not present".
2. "REVISION file takes precedence over git commit SHA".
3. "config.revision takes precedence over git commit SHA".

### Checklist

### DIFF
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -406,12 +406,15 @@ module Rails
         @revision = begin
           root.join("REVISION").read.strip.presence
         rescue SystemCallError
+          nil
+        end
+        if @revision.nil?
           r, w = IO.pipe
           success = system("git", "-C", root.to_s, "rev-parse", "HEAD", in: File::NULL, err: File::NULL, out: w)
           w.close
           rev = r.read.strip
           r.close
-          rev if success
+          @revision = rev if success
         end
         @revision_initialized = true
       end
--- a/railties/test/application/app_revision_test.rb
+++ b/railties/test/application/app_revision_test.rb
@@ -47,6 +47,74 @@ module ApplicationTests
       assert_equal "config-wins", Rails.app.revision
     end
 
+    test "revision reads from git commit SHA when REVISION file not present" do
+      Dir.chdir(app_path) do
+        `git init`
+        `git config user.email "test@example.com"`
+        `git config user.name "Test User"`
+        File.write("dummy.txt", "content")
+        `git add .`
+        `git commit -m "Initial commit"`
+
+        commit_sha = `git rev-parse HEAD`.strip
+
+        require "#{app_path}/config/environment"
+        assert_equal commit_sha, Rails.app.revision
+      end
+    end
+
+    test "git commit SHA is used when REVISION file is empty" do
+      Dir.chdir(app_path) do
+        File.write("#{app_path}/REVISION", "")
+
+        `git init`
+        `git config user.email "test@example.com"`
+        `git config user.name "Test User"`
+        File.write("dummy.txt", "content")
+        `git add .`
+        `git commit -m "Initial commit"`
+
+        commit_sha = `git rev-parse HEAD`.strip
+
+        require "#{app_path}/config/environment"
+        assert_equal commit_sha, Rails.app.revision
+      end
+    end
+
+    test "REVISION file takes precedence over git commit SHA" do
+      Dir.chdir(app_path) do
+        File.write("#{app_path}/REVISION", "file-revision-123")
+
+        `git init`
+        `git config user.email "test@example.com"`
+        `git config user.name "Test User"`
+        File.write("dummy.txt", "content")
+        `git add .`
+        `git commit -m "Initial commit"`
+
+        require "#{app_path}/config/environment"
+        assert_equal "file-revision-123", Rails.app.revision
+      end
+    end
+
+    test "config.revision takes precedence over git commit SHA" do
+      Dir.chdir(app_path) do
+        `git init`
+        `git config user.email "test@example.com"`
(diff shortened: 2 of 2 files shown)


byroot:
> Rails should treat an empty REVISION file as "not present" and fall back to git.

Could you explain your reasoning? I don't think I agree with that.

In which scenario and empty REVISION file would be anything but a mistake?

tahsin352:
@byroot Thanks for your feedback.

An empty REVISION file almost certainly represents an error state rather than intentional configuration. When deployment processes fail silently—perhaps due to a command that didn't execute, a race condition, or a permissions issue—they may create the file but write nothing to it. Treating this as a valid revision would preserve the error state, while falling back to git recovers gracefully from what is likely a mistake. 

Throughout Ruby on Rails, empty strings and nil are often treated similarly when they represent "no value." An empty REVISION file literally contains no revision information—it's the file equivalent of an empty string. Rails already treats many empty configurations as equivalent to absent ones (empty arrays, blank strings in validations, etc.). This would maintain that philosophical consistency.

Production systems should be resilient to minor configuration issues. An empty file conveys no information—it's semantically equivalent to the file not existing at all. An empty REVISION file shouldn't break version reporting or deployment tracking. This approach prioritizes practical resilience and user experience over strict file-presence semantics.

byroot:
> Treating this as a valid revision would preserve the error state, while falling back to git recovers gracefully from what is likely a mistake.

I disagree here. I don't think it's a good behavior for a production system to silently ignore this type of error and move on.

> Production systems should be resilient to minor configuration issues. 

100% disagree here. Configuration issues should fail loud and clear as to be corrected immediately. There is nothing worse than an incorrect configuration being ignored, it makes for awful debug session where you are convinced you did something, just for the app or framework to silently ignore you.

PR #56599, opened 2026-01-13
Active Record: Add support for changing PostgreSQL index comments

eglitobias:
### Motivation / Background

Active Record supports comments on tables and columns, but index comments are currently only partially supported and lack a dedicated API for modification after creation.

PostgreSQL supports index comments via `COMMENT ON INDEX`, however Active Record does not currently expose a method to change or remove index comments in migrations.

This Pull Request completes index comment support by introducing a dedicated, reversible API that aligns index comments with existing comment functionality in Active Record.

### Detail

This Pull Request adds a new `change_index_comment` migration API and improves index comment handling for PostgreSQL.

Specifically, it:

* Introduces `change_index_comment`, allowing index comments to be added, changed, or removed in migrations.
* Makes `change_index_comment` reversible when `:from` and `:to` options are provided, consistent with other comment-related APIs.
* Implements `change_index_comment` for PostgreSQL using `COMMENT ON INDEX`, including correct handling of schema-qualified index names.
* Refactors PostgreSQL `add_index` to reuse `change_index_comment` when a comment is specified.
* Adds `supports_index_comments?` to explicitly model adapter capabilities, since index comments are not universally supported across databases.
* Adds test coverage and documentation updates to reflect the new API.

### Additional information

This change builds on a bugfix introduced in https://github.com/rails/rails/pull/56582, and extends it by providing a complete public API for working with index comments.

### Checklist

* [x] Commit message clearly explains the change and motivation.
* [x] Tests are added for the new behavior.
* [x] CHANGELOG updated to reflect the new feature.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add `change_index_comment` and improve PostgreSQL index comment support.
+
+    This change introduces a dedicated API for adding, updating, and removing
+    index comments, making index comments reversible in migrations and consistent
+    with existing table and column comment APIs.
+
+    *Tobias Egli*
+
 *   Fix PostgreSQL schema dumping to handle schema-qualified table names in foreign_key references that span different schemas.
 
         # before
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -874,6 +874,12 @@ module ActiveRecord
       #
       # Note: only supported by PostgreSQL.
       #
+      # ====== Creating an index with a comment
+      #
+      #   add_index(:accounts, :branch_id, comment: "Used by reports")
+      #
+      # Note: only supported by PostgreSQL.
+      #
       # ====== Creating an index where NULLs are treated equally
       #
       #   add_index(:people, :last_name, nulls_not_distinct: true)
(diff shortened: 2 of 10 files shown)


fatkodima:
Commenting is used rarely on databases and if used then mostly on only tables and columns, which makes the most sense and rails already has support for it. 
If we support comments for indexes, then comments for other objects like foreign keys, other constraints, schema, views etc should be supported too? 🤔

I don't think this feature is used enough and worth the added change into the rails framework, but maybe someone from the rails team has other opinions.

eglitobias:
I get the concern, but this is not really a new feature. Rails already supports index comments today. You can add a comment when creating an index, and Rails already returns it via connection.indexes.

My previous PR fixes a bug in this existing behavior and clearly shows that the feature already exists and works: https://github.com/rails/rails/pull/56582

Because of that, I think we should either properly support it or clearly decide to remove it. Keeping a half-working, undocumented feature seems worse than fixing it.

Index comments are probably more common and more useful in practice than e.g. foreign keys. An index often exists for a non-obvious performance reason, and a comment is a good way to explain why it exists to DBAs or anyone inspecting the database later.

fatkodima:
> but this is not really a new feature

I meant this
> Add `change_index_comment`

is a new feature.

eglitobias:
Thanks for the feedback. I really appreciate the discussion.

Yes, `change_index_comment` is a new feature. The motivation is consistency, not introducing a brand-new feature.

Rails already supports index comments via `add_index(comment:)` and exposes them through `connection.indexes`, but unlike other comments there was no dedicated, reversible way to modify or remove them afterward. For tables and columns, this already exists with `change_table_comment` and `change_column_comment`.

This PR completes the existing support by adding the missing “change” operation and reusing it internally, instead of keeping index comments as a partially supported edge case.

If index comments aren’t something Rails wants to support at all, that’s fair but then the existing support should probably be removed rather than left half-implemented.

eglitobias:
@fatkodima I don’t want this PR to block anything. If the preferred outcome is to only merge the [bugfix PR](https://github.com/rails/rails/pull/56582) that's also fine for me.

PR #56601, opened 2026-01-13
Allow retry_on wait procs to accept error as a second argument

jpcamara:
### Motivation / Background

Currently, `retry_on` `wait` procs can only accept the execution count as an argument, which limits the ability to customize retry delays based on the error itself. Some exceptions naturally include retry timing information (e.g., rate limit errors with `Retry-After` headers), but there's no way to access this information in the wait proc.

### Detail

This Pull Request adds support for wait procs to optionally receive the error as a second argument:

```ruby
class RemoteServiceJob < ActiveJob::Base
  retry_on CustomError, wait: ->(executions, error) { error.retry_after || executions * 2 }

  def perform
    # ...
  end
end
```

The implementation uses arity checking to maintain backwards compatibility:
- Procs with arity 0 or 1 receive only the execution count (existing behavior)
- Procs with arity 2+ receive both the execution count and error

This allows developers to implement dynamic retry strategies based on error properties while maintaining full backwards compatibility with existing code.

### Additional information

The change is implemented in `ActiveJob::Exceptions#call_retry_delay_algorithm`, which checks the proc's arity and calls it with the appropriate number of arguments. All existing tests pass, and a new test demonstrates the 2-argument form working correctly.

### Checklist

### DIFF
--- a/activejob/lib/active_job/exceptions.rb
+++ b/activejob/lib/active_job/exceptions.rb
@@ -26,7 +26,8 @@ module ActiveJob
       #
       # ==== Options
       # * <tt>:wait</tt> - Re-enqueues the job with a delay specified either in seconds (default: 3 seconds),
-      #   as a computing proc that takes the number of executions so far as an argument, or as a symbol reference of
+      #   as a computing proc that takes the number of executions so far as an argument (and optionally the error),
+      #   or as a symbol reference of
       #   <tt>:polynomially_longer</tt>, which applies the wait algorithm of <tt>((executions**4) + (Kernel.rand * (executions**4) * jitter)) + 2</tt>
       #   (first wait ~3s, then ~18s, then ~83s, etc)
       # * <tt>:attempts</tt> - Enqueues the job the specified number of times (default: 5 attempts) or a symbol reference of <tt>:unlimited</tt>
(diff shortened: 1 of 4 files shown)


activejob/lib/active_job/exceptions.rb
```diff
@@ -183,12 +185,19 @@ def determine_delay(seconds_or_duration_or_algorithm:, executions:, jitter: JITT
...
-          algorithm.call(executions)
+          call_retry_delay_algorithm(algorithm, executions, error)
         else
           raise "Couldn't determine a delay based on #{seconds_or_duration_or_algorithm.inspect}"
         end
       end
 
+      def call_retry_delay_algorithm(algorithm, executions, error)
+        case algorithm.arity
+        when 0, 1 then algorithm.call(executions)
```
byroot: If arity is `0` this would fail no?

Also not sure it's really worth extracting that method.

byroot (commented):
I think this is a valuable feature, I just have one small nitpick.

I'm not sure what this means for the jitter functionality though.

activejob/lib/active_job/exceptions.rb
jpcamara: Yea, fair. Probably just should have checked 1 and if 0 was given, it's gonna blow up either way and isn't the documented interface.

This work as a simplifcation?

```rb
when Proc
  algorithm = seconds_or_duration_or_algorithm
  algorithm.arity == 1 ? algorithm.call(executions) : algorithm.call(executions, error)
```

byroot: Yep

jpcamara: Much simpler, thanks! Updated.

PR #32228, opened 2018-03-12
fix casting zero-float numbers to boolean

mikekytyzov:
I was a little bit wondering... why?

```ruby
> Boolean(0.0)
true
```

but if we compare, it will be okay because of type casting, its ok 

```ruby
0.0 == 0
true
```

but what is 0.0 ? is that zero ? 

```ruby
0.0.zero?
true
```
lets see a C code

```C
    if (RFLOAT_VALUE(num) == 0.0) {
        return Qtrue;
    }
```

so 0.0 is zero, but 

```ruby
> Boolean(0)
false
```

so if 0.0 is zero it should be casted as 0 and has to be false, right ?

now 0.0, 0.0000000, etc will be false

PR #56603, opened 2026-01-13
Fix unexpected result for Boolean casting

### LINKED ISSUE #56602: ActiveModel::Type::Boolean Cast Unexpected Result for Float Zero
### Steps to reproduce
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end

    t.integer :post_id
  end
end

  has_many :comments
end

  belongs_to :post
end

  def test_integer_and_float_zero_cast_to_boolean
    @integer_cast = ActiveModel::Type::Boolean.new.cast(0)
    @float_cast = ActiveModel::Type::Boolean.new.cast(0.0)

    assert_equal @integer_cast, @float_cast
  end

  def test_integer_zero_cast_to_boolean
    @integer_cast = ActiveModel::Type::Boolean.new.cast(0)
    assert_equal false, @integer_cast
  end

  def test_float_zero_cast_to_boolean
    @float_cast = ActiveModel::Type::Boolean.new.cast(0.0)
    assert_equal false, @float_cast
  end
end
```

### Expected behavior
I would expect float zero (0.0) and integer zero (0) to both cast to the same boolean `false`

### Actual behavior
Float zero (0.0) returns `true`

### System configuration

**Rails version**: 8.1.2

**Ruby version**: 3.2.2

crespire:
### Motivation / Background
We ran into a surprising Boolean casting outcome where float zero and integer zero did not both cast to the same boolean result.

This Pull Request has been created because integer and float zero should return the same result when being cast to a boolean.

Fixes #56602

### Detail

This Pull Request changes `ActiveModel::Type::Boolean` to add an additional `FALSE_VALUE` of `0.0`.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/type/boolean.rb
+++ b/activemodel/lib/active_model/type/boolean.rb
@@ -14,7 +14,7 @@ module ActiveModel
     class Boolean < Value
       include Helpers::Immutable
       FALSE_VALUES = [
-        false, 0,
+        false, 0, 0.0,
         "0", :"0",
         "f", :f,
         "F", :F,
--- a/activemodel/test/cases/type/boolean_test.rb
+++ b/activemodel/test/cases/type/boolean_test.rb
@@ -34,6 +34,7 @@ module ActiveModel
         # explicitly check for false vs nil
         assert_equal false, type.cast(false)
         assert_equal false, type.cast(0)
+        assert_equal false, type.cast(0.0)
         assert_equal false, type.cast("0")
         assert_equal false, type.cast("f")
         assert_equal false, type.cast("F")


skipkayhil:
I don't think we can change this without considering backwards compatibility: https://github.com/rails/rails/pull/32228

crespire:
> I don't think we can change this without considering backwards compatibility: #32228

Yeah, I did see this on a more close search. Agreed, but this result is so surprising to me, I think it's worth a conversation but happy to defer to the team on it.

It does seem like it will require a depreciation warning with a lot of lead time if the team decides that this change should go through.

I'm also unsure how to fix the failing tests, because there seems to be some database element I haven't figured out yet. Will take a look at all the docs and reference material and see if I can get those fixed. I'll embark on this rabbit hole if the team thinks this is worthwhile to pursue!

multiplegeorges:
This inconsistency (bug?) caused a minor production issue for us, which we've worked around. The current behaviour is quite surprising. I'd like to see this PR get merged once it is polished up.

@skipkayhil I completely agree that this change should be done with care. I like @crespire's suggestion to perhaps log a warning/deprecation and change the behaviour when incrementing major versions.

crespire:
Any update on this? It would be good to have some next step here.

PR #56605, opened 2026-01-13
Fix example in docs for partial index (add_index)

stefan131:
Update the partial index example to use a complete where clause condition instead of just the value

The relevant documentation was added in https://github.com/rails/rails/pull/4956. 
Included tests (`activerecord/test/cases/adapters/postgresql/active_schema_test.rb`) validate the required change in documentation:

`activerecord/test/cases/adapters/postgresql/active_schema_test.rb:30-31`:
```ruby
    expected = %(CREATE UNIQUE INDEX "index_people_on_last_name" ON "people" ("last_name") WHERE state = 'active')
    assert_equal expected, add_index(:people, :last_name, :unique => true, :where => "state = 'active'")
```

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -856,11 +856,11 @@ module ActiveRecord
       #
       # ====== Creating a partial index
       #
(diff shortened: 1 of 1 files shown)


fatkodima:
The example was meant to use a boolean `accounts.active` column as a condition, so it correct.

PR #56609, opened 2026-01-14
Add input validation to Array#in_groups

### LINKED ISSUE #56608: Array#in_groups missing input validation
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
end

require "active_support"
require "active_support/core_ext/array/grouping"
require "minitest/autorun"

class InGroupsValidationBugTest < ActiveSupport::TestCase
  def test_in_groups_with_zero_should_raise_argument_error
    error = assert_raises(ArgumentError) { [1, 2, 3].in_groups(0) }
    assert_match(/positive integer/, error.message)
  end

  def test_in_groups_with_negative_should_raise_argument_error
    error = assert_raises(ArgumentError) { [1, 2, 3].in_groups(-1) }
    assert_match(/positive integer/, error.message)
  end

  def test_in_groups_with_nil_should_raise_argument_error
    error = assert_raises(ArgumentError) { [1, 2, 3].in_groups(nil) }
    assert_match(/positive integer/, error.message)
  end

  def test_in_groups_with_float_should_convert_to_integer
    result = [1, 2, 3, 4, 5].in_groups(2.9) # 2.9.to_i => 2 groups
    assert_equal 2, result.size
  end
end
```

### Expected behavior

`Array#in_groups` should validate its input the same way `Array#in_groups_of` does.

### Actual behavior

```ruby
in_groups(0)       # => ZeroDivisionError: divided by 0
in_groups(-1)      # => Returns [] silently (no error)
in_groups(nil)     # => TypeError: nil can't be coerced into Integer
in_groups(2.9)    # => NoMethodError: undefined method 'times' for an instance of Float
```

### System configuration

**Rails version**: 8.1.2

**Ruby version**: 3.3.4

Saidbek:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/56608

`Array#in_groups` was missing input validation that `Array#in_groups_of` has. This caused confusing errors for invalid arguments.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/array/grouping.rb
+++ b/activesupport/lib/active_support/core_ext/array/grouping.rb
@@ -60,6 +60,12 @@ class Array
   #   ["4", "5"]
   #   ["6", "7"]
   def in_groups(number, fill_with = nil, &block)
+    if number.to_i <= 0
+      raise ArgumentError,
+        "Number of groups must be a positive integer, was #{number.inspect}"
+    end
+
+    number = number.to_i
     # size.div number gives minor group size;
     # size % number gives how many objects need extra accommodation;
     # each group hold either division or division + 1 items.
--- a/activesupport/test/core_ext/array/grouping_test.rb
+++ b/activesupport/test/core_ext/array/grouping_test.rb
@@ -85,10 +85,27 @@ class GroupingTest < ActiveSupport::TestCase
       (1..7).to_a.in_groups(3, false)
   end
 
-  def test_in_groups_invalid_argument
+  def test_in_groups_of_invalid_argument
     assert_raises(ArgumentError) { [].in_groups_of(0) }
     assert_raises(ArgumentError) { [].in_groups_of(-1) }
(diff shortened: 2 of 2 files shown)


activesupport/lib/active_support/core_ext/array/grouping.rb
```diff
@@ -60,6 +60,12 @@ def in_groups_of(number, fill_with = nil, &block)
   #   ["4", "5"]
   #   ["6", "7"]
   def in_groups(number, fill_with = nil, &block)
+    if number.to_i <= 0
+      raise ArgumentError,
+        "Number of groups must be a positive integer, was #{number.inspect}"
+    end
+
+    number = number.to_i
```
Saidbek, line 68: Added explicitly after validation because `in_groups` calls `.times` (no float coercion), while `in_groups_of` uses `.each_slice`, which converts floats to integers internally.

Issue #50048, opened 2023-11-13
`alias_attribute` no longer overrides reader method defined in concern


tf:
I realize this is an edge case. Mainly posting the issue to document the change for others. Took me an hour today to understand a failure in our app caused by this change.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "activerecord", "~> 7.1.0"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
    t.string :title
  end
end

module Displayable
  # This is the a default implementation for classes that do not wish to implement display_name themselves
  def display_name
    "Default name"
  end
end

class Post < ActiveRecord::Base
  include Displayable

  alias_attribute :display_name, :title
end

class BugTest < Minitest::Test
  def test_association_stuff
    post = Post.create!(title: "Some title")

    assert_equal "Some title", post.display_name
  end
end
```

### Expected behavior

`alias_attribute` in Post overrides `display_name` reader method defined in `Displayable` module. Test passes.

### Actual behavior

I assume since https://github.com/rails/rails/pull/48533, `alias_attribute` no longer defines the `display_name` reader method on `Post` itself, but on `Post::GeneratedAttributeMethods` which comes after `Displayable` in the ancestor chain:

```
Post.ancestors
# => [Post(id: integer, title: string), Displayable, Post::GeneratedAssociationMethods, Post::GeneratedAttributeMethods, ActiveRecord::Base, ...]
```

Test fails:

```
1) Failure:
BugTest#test_association_stuff [alias_attribute_bug.rb:46]:
Expected: "Some title"
  Actual: "Default name"
```

Changing `activerecord` version to 7.0.0 fixes the test.

### System configuration
**Rails version**: 7.1.2

**Ruby version**: 3.2.2

nvasilevski:
So using the module instead of defining method directly on the class wasn't a necessity but it was a convenient time for us to do that. And technically we should be able to get back to generating methods on the class without breaking any expectations

[...]

PR #42452, opened 2021-06-11
Respect the `dependent: :destroy_async` when replacing the association

smt116:
See https://github.com/rails/rails/issues/42430

There are two ways for deleting an associated record:

  1. by deleting it explicitly (i.e. `parent.has_many.first.destroy!`),
  2. by replacing the association (i.e. `parent.has_many = []`).

Previously, Active Record respected the asynchronous deletion only with
the first approach. When replacing the association, it just detached the
record(s) from the parent without removing them from the database (via
enqueuing relevant background job).

Make the dependent option consistent and cleanup records from the
database even when removing them via replacing the association.

Fixes #42430

Issue #42430, opened 2021-06-09
Inconsistent behavior of `dependent: :destroy_async` for  the`has_many` association


smt116:
### Steps to reproduce

1. Generate a new application:

    ```
    rails new AsyncDestroyTestApp --skip-action-mailer --skip-action-mailbox --skip-action-text --skip-active-storage --skip-action-cable --skip-sprockets --skip-spring --skip-listen --skip-javascript --skip-turbolinks --skip-jbuilder --skip-test --skip-system-test --skip-bootsnap
    ```

2. Apply the following diff (add models):

    ```diff
    diff --git a/app/models/book.rb b/app/models/book.rb
    new file mode 100644
    index 0000000..f1e2a65
    --- /dev/null
    +++ b/app/models/book.rb
    @@ -0,0 +1,3 @@
    +class Book < ApplicationRecord
    +  has_many :tags, dependent: :destroy_async
    +end
    diff --git a/app/models/tag.rb b/app/models/tag.rb
    new file mode 100644
    index 0000000..7b23605
    --- /dev/null
    +++ b/app/models/tag.rb
    @@ -0,0 +1,2 @@
    +class Tag < ApplicationRecord
    +end
    diff --git a/db/migrate/20210609112034_add_books_and_tags.rb b/db/migrate/20210609112034_add_books_and_tags.rb
    new file mode 100644
    index 0000000..baed2c0
    --- /dev/null
    +++ b/db/migrate/20210609112034_add_books_and_tags.rb
    @@ -0,0 +1,8 @@
    +class AddBooksAndTags < ActiveRecord::Migration[6.1]
    +  def change
    +    create_table :books
    +    create_table :tags do |t|
    +      t.references :book, foreign_key: true, null: true
    +    end
    +  end
    +end
    diff --git a/db/schema.rb b/db/schema.rb
    new file mode 100644
    index 0000000..fca4c69
    --- /dev/null
    +++ b/db/schema.rb
    @@ -0,0 +1,24 @@
    +# This file is auto-generated from the current state of the database. Instead
    +# of editing this file, please use the migrations feature of Active Record to
    +# incrementally modify your database, and then regenerate this schema definition.
    +#
    +# This file is the source Rails uses to define your schema when running `bin/rails
    +# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
    +# be faster and is potentially less error prone than running all of your
    +# migrations from scratch. Old migrations may fail to apply correctly if those
[...]

PR #56612, opened 2026-01-14
Fix duplicate after_commit callback registration for destroy_async in class inheritance

yujiteshima:
When a parent class and its subclass both declared associations with `dependent: :destroy_async`, the `after_commit` callback was registered multiple times because `mixin.method_defined?` only checked the class's own `GeneratedAssociationMethods` module, not the inherited one.

This caused `DestroyAssociationAsyncJob` to be enqueued multiple times for the same destruction.

The fix changes `mixin.method_defined?(:_after_commit_jobs)` to `model.method_defined?(:_after_commit_jobs)`, which checks the entire inheritance chain.

Fixes #56610.

### Motivation / Background

This Pull Request has been created because when using class inheritance with `dependent: :destroy_async` associations in both parent and child classes, the async destroy job was being enqueued multiple times for a single record destruction.

```ruby
  has_many :comments, dependent: :destroy_async
end

class VisualPost < Post
  has_many :images, dependent: :destroy_async
end


visual_post = VisualPost.create!
visual_post.destroy  # => DestroyAssociationAsyncJob enqueued multiple times
```

### Detail

This Pull Request changes the method check in add_after_commit_jobs_callback from mixin.method_defined?(:_after_commit_jobs) to model.method_defined?(:_after_commit_jobs). The original code only checked the class's own GeneratedAssociationMethods module. Since each class has its own module, the parent's _after_commit_jobs method was not detected when processing the child class, causing the callback to be registered again. Using model.method_defined? instead checks the entire inheritance chain, including the parent's GeneratedAssociationMethods module.

### Additional information

Related issues:
- #42430 - Another destroy_async bug (different root cause, has existing PR #42452)
- #50048 - Related GeneratedAssociationMethods inheritance issue

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Fix duplicate `after_commit` callback registration for `dependent: :destroy_async` in class inheritance.
+
+    When a parent class and its subclass both declared associations with
+    `dependent: :destroy_async`, the `after_commit` callback that processes
+    async destroy jobs was being registered multiple times, causing
+    `DestroyAssociationAsyncJob` to be enqueued multiple times for the same
+    destruction.
+
+    This fix ensures the callback is only registered once in the parent class
+    and properly inherited by subclasses.
+
+    Fixes #56610.
+
+    *Yuji Teshima*
+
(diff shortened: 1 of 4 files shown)


Aethelflaed:
Any news on that front?

PR #56630, opened 2026-01-17
Isolate Update Side Effects

eric1234:
Fixes #14698.

### Motivation / Background

Nested attribute assignment can persist changes to associated records before the parent is validated. When update runs inside a joinable outer transaction, the rollback raised by update is swallowed and those side effects are committed.

### Detail

This Pull Request changes update/update! to request a savepoint when joining an existing transaction (if savepoints are supported), so the internal rollback restores association state without forcing callers to raise.

It also adds a regression test for nested attributes inside an explicit transaction.

### Additional information

The correctness of this fix depends on support for savepoints. If the database does not support savepoints the historical behavior will still be present.

Given that this is a very edge case and given that all the major adapters support savepoints this probably corrects it for 99% of apps and for those that don't they are no worse off than they have for the 11 years this bug has been present.

The alternative would be to delay SQL changes on nested objects until after all validation but that would be a much bigger change for a very small edge case.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
@@ -450,19 +450,34 @@ module ActiveRecord
 
       def rollback
         unless @state.invalidated?
-          connection.rollback_to_savepoint(savepoint_name) if materialized? && connection.active?
+          if materialized? && connection.active?
+            with_savepoint_error_handling do
+              connection.rollback_to_savepoint(savepoint_name)
+            end
+          end
         end
         @state.rollback!
         @instrumenter.finish(:rollback) if materialized?
       end
 
       def commit
-        connection.release_savepoint(savepoint_name) if materialized?
+        if materialized?
+          with_savepoint_error_handling do
+            connection.release_savepoint(savepoint_name)
+          end
+        end
         @state.commit!
         @instrumenter.finish(:commit) if materialized?
       end
 
       def full_rollback?; false; end
+
+      private
+        def with_savepoint_error_handling
+          yield
+        rescue ActiveRecord::StatementInvalid => error
+          raise unless connection.ignore_savepoint_error?(error)
+        end
     end
 
     # = Active Record Real \Transaction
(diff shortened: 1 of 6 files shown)


eric1234:
Looks like some bug with MySQL deadlocks. Looking into ways to resolve...

activerecord/test/cases/nested_attributes_test.rb
```diff
@@ -1049,6 +1049,35 @@ def test_attr_accessor_of_child_should_be_value_provided_during_update
...
 
+class TestNestedAttributesTransactionRollback < ActiveRecord::TestCase
+  self.use_transactional_tests = false
+
+  def teardown
+    Ship.connection.disable_referential_integrity do
+      Ship.delete_all
+      Pirate.delete_all
+    end
+  end
```
eric1234, line 1060: I wasn't sure about this.

I needed to disable transaction tests since we are testing transaction behavior which meant manual cleanup. Since I create ships and pirates that is what I delete. But without the disabling of referential integrity I would randomly get foreign key failures when running a fuller test suite.

I'm guessing other tests aren't cleaning up after themselves and my `delete_all` is then an issue. Could alt switch to `destroy_all`? Or maybe I shouldn't bother cleaning up after myself?

activerecord/lib/active_record/connection_adapters/abstract/transaction.rb
```diff
@@ -450,19 +450,34 @@ def restart
...
       end
 
       def full_rollback?; false; end
+
+      private
+        def with_savepoint_error_handling
+          yield
+        rescue ActiveRecord::StatementInvalid => error
+          raise unless connection.ignore_savepoint_error?(error)
+        end
```
eric1234, line 480: This change was to get [this test](https://github.com/rails/rails/blob/main/activerecord/test/cases/adapters/abstract_mysql_adapter/nested_deadlock_test.rb) to pass. Evidently when a deadlock occurs in MySQL it rolls back the entire transaction (including all savepoints implicitly). This means when our new implicit savepoint is cleaned up it tries to release that savepoint but it no longer exists.

To resolve I made it so interactions with savepoint during the savepoint book keeping, if the adapter indicates it's just complaining about the savepoint not existing then we ignore that error. Not cleanup needed if the database did implicit cleanup.

eric1234:
I resolved the transaction failures but now getting a failure in actioncable. Seems entirely unrelated so I'm thinking just a flake but I don't know how to re-run a specific step on BuildKite to get it all green.

PR #56632, opened 2026-01-17
Prioritize markdown when accepted

coorasse:
### Motivation / Background

Since we have markdown support, I noticed that this was anyway not served to llms. The reason is that, for example Claude Code, send the following header: `Accept: text/markdown, text/html, */*`. 
For historical reasons (?) Rails serves html because the Accept header includes `*/*`.

This Pull Request changes the behavior for markdown, so that markdown is prioritized.

What I am not sure is about having this change **only** for markdown, but I guess it would otherwise be a too big breaking change...

What we did in our app to support this is the followin:

```
def prioritize_markdown_format
  return unless request.accepts.first&.to_s == "text/markdown"
  request.formats = [:md, :html]
end
```

### Checklist

Related to #55511

### DIFF
--- a/actionpack/lib/action_dispatch/http/mime_negotiation.rb
+++ b/actionpack/lib/action_dispatch/http/mime_negotiation.rb
@@ -220,6 +220,9 @@ module ActionDispatch
         # which case we assume you're a browser and send HTML.
         BROWSER_LIKE_ACCEPTS = /,\s*\*\/\*|\*\/\*\s*,/
 
+        # Match Accept headers where markdown is explicitly preferred before other types
+        MARKDOWN_PREFERRED_ACCEPTS = /\A\s*text\/markdown\s*,/
+
         def params_readable?
           parameters[:format]
         rescue *RESCUABLE_MIME_FORMAT_ERRORS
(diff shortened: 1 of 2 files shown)


actionpack/lib/action_dispatch/http/mime_negotiation.rb
```diff
@@ -220,6 +220,9 @@ def should_apply_vary_header?
         # which case we assume you're a browser and send HTML.
         BROWSER_LIKE_ACCEPTS = /,\s*\*\/\*|\*\/\*\s*,/
 
+        # Match Accept headers where markdown is explicitly preferred before other types
+        MARKDOWN_PREFERRED_ACCEPTS = /\A\s*text\/markdown\s*,/
```
schmijos, line 224: Shouldn't this rathe be `AI_PREFERRED_ACCEPTS` or `LLM_PREFERRED_ACCEPTS` so? We assume a tool.

skipkayhil:
Thank you for the pull request. I do think this should be solved, but not by adding _more_ special handling to the Accept header. I think #57579 is the better direction as it simplifies things instead.

Happy to continue the discussion if you think that PR won't fix this, but for now I'm going to close this.

PR #56634, opened 2026-01-17
Minimize AWS RDS Proxy connection pinning caused by SET for PostgreSQL

i7an:
### Motivation / Background

Rails applications using **PostgreSQL** cannot effectively use [AWS RDS Proxy](https://aws.amazon.com/rds/proxy/). Active Record’s connection setup issues `SET` commands (timezone, encoding, etc.), and those `SET` commands [cause RDS Proxy connection pinning](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html#rds-proxy-pinning.postgres).

Given how widely used AWS and PostgreSQL are, it makes sense to address this.

This issue can also be addressed by using the `EXCLUDE_VARIABLE_SETS` option in [RDS Proxy's SessionPinningFilters](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ConnectionPoolConfigurationInfo.html), but that feature is currently only supported for MySQL.

### Detail

Only issue `SET` command if current value (checked via `SHOW`) differs from expected for:
* `timezone`
* `client_encoding`
* `schema_search_path`
* `client_min_messages`
* `intervalstyle`
* `standard_conforming_strings`
* Custom variables

It is expected that you set these variables in the RDS Proxy **initialization query**. When all the parameters match the expected values, there is no need for `SET` commands, thus pinning is avoided.

**Performance.** For a typical application, this change adds 2-3 `SHOW` commands per connection, resulting in sub-millisecond overhead (verified in both local and cloud environments).

### Additional information

`DatabaseConnectionsCurrentlySessionPinned` before and after the fix:

image

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -15,7 +15,8 @@ module ActiveRecord
           intent = internal_build_intent(sql, name, allow_retry:, materialize_transactions:)
           intent.execute!
           result = intent.raw_result
-          result.map_types!(@type_map_for_results).values
+          result.map_types!(@type_map_for_results) if @type_map_for_results
+          result.values
         end
 
         READ_QUERY = ActiveRecord::ConnectionAdapters::AbstractAdapter.build_read_query_regexp(
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -7,6 +7,7 @@ require "active_support/core_ext/object/try"
 require "active_record/connection_adapters/abstract_adapter"
 require "active_record/connection_adapters/statement_pool"
 require "active_record/connection_adapters/postgresql/column"
+require "active_record/connection_adapters/postgresql/configuration_parameters"
 require "active_record/connection_adapters/postgresql/database_statements"
 require "active_record/connection_adapters/postgresql/explain_pretty_printer"
 require "active_record/connection_adapters/postgresql/oid"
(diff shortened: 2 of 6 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
```diff
@@ -15,7 +15,8 @@ def query_rows(sql, name = "SCHEMA", allow_retry: true, materialize_transactions
           intent = internal_build_intent(sql, name, allow_retry:, materialize_transactions:)
           intent.execute!
           result = intent.raw_result
-          result.map_types!(@type_map_for_results).values
+          result.map_types!(@type_map_for_results) if @type_map_for_results
+          result.values
```
i7an, line 19: To allow for `query_value` before `add_pg_decoders` is called.

activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -1082,13 +1085,29 @@ def reconfigure_connection_timezone
           # If using Active Record's time zone support configure the connection
           # to return TIMESTAMP WITH ZONE types in UTC.
           if default_timezone == :utc
-            intent = QueryIntent.new(adapter: self, processed_sql: "SET SESSION timezone TO 'UTC'", name: "SCHEMA")
```
i7an, line 1085: I think the direct use of `QueryIntent` here is just a leftover that wasn’t removed after the refactoring. All other parameters are already being set through `query_command`. The `timezone` can be passed through variables, and in that case it will be applied via `query_command` as well.

The key distinction is `raw_sql` vs `processed_sql`.

matthewd:
@i7an sorry I missed this PR! 

We've just landed similar work in #57070, with a focus on ensuring the application can manually opt out of setting the value. Spending a `SHOW` to avoid the pin from the `SET` is interesting. 🤔

Issue #54943, opened 2025-04-18
Slow query generation


skipkayhil:
### Steps to reproduce

We have a query in our app where we have to resort to string templates because Active Record generates it too slowly. I've attached a benchmark that reproduces the two examples and the difference in performance between them.

<details>
<summary>Benchmark Script</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "rails"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"

  gem "benchmark-ips"
  gem "sqlite3"
end

require "active_record/railtie"
require "benchmark/ips"

# This connection will do for database-independent bug reports.
ENV["DATABASE_URL"] = "sqlite3::memory:"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.eager_load = false
  config.logger = Logger.new($stdout)
  config.secret_key_base = "secret_key_base"
end
Rails.application.initialize!

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
    t.string :column_one
    t.string :column_two
    t.string :column_three
    t.string :column_four
    t.string :a
    t.bigint :poly_id
    t.string :poly_type
  end

  create_table :comments, force: true do |t|
    t.string :column_one
    t.string :column_two
    t.string :column_three
    t.string :column_four
  end
end

class Post < ActiveRecord::Base
end

class Comment < ActiveRecord::Base
end

BASE_TABLE = "comments"

POLY_TYPES = ["foo", "bar", "baz"]

# not inlined because this is one of a few possible query fragments
A_CONDITION = "posts.a IN ('a', 'b')"

def with_strings
  poly_condition = "posts.poly_type IN (#{POLY_TYPES.map { "'#{_1}'" }.join(",")})"

  poly_condition << " OR (posts.poly_type = 'foo' AND posts.poly_id = '1')"

  <<~SQL
    EXISTS (
      SELECT 1
      FROM posts
      WHERE #{BASE_TABLE}.column_one = posts.column_one
        AND #{BASE_TABLE}.column_two = posts.column_two
        AND #{BASE_TABLE}.column_three = posts.column_three
        AND #{BASE_TABLE}.column_four = posts.column_four
        AND #{A_CONDITION}
        AND (#{poly_condition})
    )
  SQL
end

def with_ar
[...]

PR #56635, opened 2026-01-17
Reduce object allocations when building WHERE clauses

dfl:
### Motivation / Background                                                                                                                                                             
                                                                                                                                                                                          
  Active Record allocates many short-lived objects when building WHERE clauses. In high-throughput applications, this creates GC pressure that affects tail latencies. Multi-tenant apps  
  that scope every query with `where(tenant_id: id)` are a common example where this compounds.                                                                                           
                                                                                                                                                                                          
  Partially addresses #54943                                                                                                                                                              
                                                                                                                                                                                          
  ### Detail                                                                                                                                                                              
                                                                                                                                                                                          
  Add a fast path for simple column equality queries that skips the full predicate expansion machinery. Also avoid intermediate `WhereClause` allocations by using `concat()` instead of  
  `+()`.                                                                                                                                                                                  
                                                                                                                                                                                          
  - Add `simple_attributes?` check in `PredicateBuilder#build_from_hash`                                                                                                                  
  - Add `WhereClause#concat` to append predicates without intermediate object                                                                                                             
  - Extract `build_where_clause_parts` to return array instead of WhereClause                                                                                                             
                                                                                                                                                                                          
  ### Additional information                                                                                                                                                              
                                                                                                                                                                                          
  ~25% fewer allocations for typical queries:                                                                                                                                             
                                                                                                                                                                                          
  - Simple where: 29 → 21 objects (-28%)                                                                                                                                                  
  - Chained wheres: 54 → 38 objects (-30%)                                                                                                                                                
  - Chain of 4: 71 → 55 objects (-23%)                                                                                                                                                    
  
```                                                                                                                                                                                        
  ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin23]                                                                                                               
  ActiveRecord: 8.1.0.alpha                                                                                                                                                               
                                                                                                                                                                                          
  ALLOCATIONS PER QUERY                                                                                                                                                                   
                                                                                                                                                                                          
    Simple column queries (fast path):                                                                                                                                                    
      Simple where (1 col)                 21.0 objects                                                                                                                                   
      Simple where (2 cols)                28.0 objects                                                                                                                                   
      Chained simple wheres                38.0 objects                                                                                                                                   
                                                                                                                                                                                          
    Chained queries:                                                                                                                                                                      
      Chain of 4                           55.0 objects                                                                                                                                   
      Chain of 6                           64.0 objects                                                                                                                                   
  ```
                                                                                                                                                                                          
  <details>                                                                                                                                                                               
  <summary>Benchmark script</summary>                                                                                                                                                     
                                                                                                                                                                                          
  ```ruby                                                                                                                                                                                 
  require "bundler/setup"                                                                                                                                                                 
  require "active_record"                                                                                                                                                                 
                                                                                                                                                                                          
      t.string :status                                                                                                                                                                    
      t.integer :author_id                                                                                                                                                                
      t.timestamps                                                                                                                                                                        
    end                                                                                                                                                                                   
  end                                                                                                                                                                                     
                                                                                                                                                                                          
  def count_allocations(iterations = 1000)                                                                                                                                                
    GC.start                                                                                                                                                                              
    GC.disable                                                                                                                                                                            
    before = GC.stat[:total_allocated_objects]                                                                                                                                            
    iterations.times { yield }                                                                                                                                                            
    after = GC.stat[:total_allocated_objects]                                                                                                                                             
    GC.enable                                                                                                                                                                             
    (after - before).to_f / iterations                                                                                                                                                    
  end                                                                                                                                                                                     
                                                                                                                                                                                          
  {                                                                                                                                                                                       
    "Simple where (1 col)" => -> { Post.where(status: "published") },                                                                                                                     
    "Simple where (2 cols)" => -> { Post.where(status: "published", author_id: 1) },                                                                                                      
    "Chained simple wheres" => -> { Post.where(status: "published").where(author_id: 1) },                                                                                                
    "Chain of 4" => -> { Post.where(status: "x").where(author_id: 1).order(:created_at).limit(10) },                                                                                      
  }.each do |name, block|                                                                                                                                                                 
    puts "#{name}: #{count_allocations(&block).round(1)} objects"                                                                                                                         
  end
  ```
  
  It's unclear if this merits a CHANGELOG entry, since it is a minor internal thing?

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1041,7 +1041,7 @@ module ActiveRecord
     end
 
     def where!(opts, *rest) # :nodoc:
-      self.where_clause += build_where_clause(opts, rest)
+      self.where_clause = where_clause.add_predicates(build_where_clause_parts(opts, rest))
       self
     end
 
@@ -1612,6 +1612,11 @@ module ActiveRecord
       end
 
       def build_where_clause(opts, rest = []) # :nodoc:
+        Relation::WhereClause.new(build_where_clause_parts(opts, rest))
+      end
+      alias :build_having_clause :build_where_clause
+
+      def build_where_clause_parts(opts, rest = []) # :nodoc:
         opts = sanitize_forbidden_attributes(opts)
 
         if opts.is_a?(Array)
@@ -1650,9 +1655,8 @@ module ActiveRecord
           raise ArgumentError, "Unsupported argument type: #{opts} (#{opts.class})"
         end
 
-        Relation::WhereClause.new(parts)
+        parts
       end
-      alias :build_having_clause :build_where_clause
 
       def async! # :nodoc:
         @async = true
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1041,7 +1041,7 @@ module ActiveRecord
     end
 
     def where!(opts, *rest) # :nodoc:
-      self.where_clause = where_clause.add_predicates(build_where_clause_parts(opts, rest))
+      self.where_clause = where_clause.concat(build_where_clause_parts(opts, rest))
       self
     end
 
--- a/activerecord/lib/active_record/relation/where_clause.rb
+++ b/activerecord/lib/active_record/relation/where_clause.rb
@@ -15,8 +15,13 @@ module ActiveRecord
         WhereClause.new(predicates + other.predicates)
       end
 
-      def add_predicates(other_predicates)
-        WhereClause.new(predicates + other_predicates)
+      def concat(other_predicates)
+        if frozen?
+          WhereClause.new(predicates + other_predicates)
+        else
+          @predicates.concat(other_predicates)
+          self
+        end
       end
 
       def -(other)
--- a/activerecord/lib/active_record/relation/where_clause.rb
+++ b/activerecord/lib/active_record/relation/where_clause.rb
@@ -15,6 +15,10 @@ module ActiveRecord
         WhereClause.new(predicates + other.predicates)
       end
 
+      def add_predicates(other_predicates)
+        WhereClause.new(predicates + other_predicates)
+      end
+
       def -(other)
         WhereClause.new(predicates - other.predicates)
       end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Reduce object allocations when building WHERE clauses.
+
+    Fewer short-lived objects means less GC pressure in high-throughput applications.
+
+    *David Lowenfels*
+
 *   Fix PostgreSQL schema dumping to handle schema-qualified table names in foreign_key references that span different schemas.
 
         # before
--- a/activerecord/lib/active_record/relation/predicate_builder.rb
+++ b/activerecord/lib/active_record/relation/predicate_builder.rb
@@ -21,8 +21,27 @@ module ActiveRecord
     end
 
     def build_from_hash(attributes, &block)
+      # Fast path: simple column equality without associations or dot notation
+      if simple_attributes?(attributes)
+        return attributes.map { |key, value| self[key.to_s, value] }
+      end
+
       attributes = convert_dot_notation_to_hash(attributes)
       expand_from_hash(attributes, &block)
+    end
+
+    # Check if attributes can use the fast path (simple column equality).
+    # Returns false if any attribute requires special handling.
+    def simple_attributes?(attributes)
+      attributes.each do |key, value|
+        return false if value.is_a?(Hash)
+        key_s = key.to_s
+        return false if key_s.include?(".")
+        return false if key.is_a?(Array)
+        return false if table.associated_with(key_s)
+        return false if table.aggregated_with?(key_s)
+      end
+      true
     end
 
     def self.references(attributes)

### CODE REVIEW

File: activerecord/lib/active_record/relation/query_methods.rb [hunk 1]
```diff
@@ -1041,7 +1041,7 @@ def where(*args)
     end
 
     def where!(opts, *rest) # :nodoc:
-      self.where_clause += build_where_clause(opts, rest)
+      self.where_clause = where_clause.concat(build_where_clause_parts(opts, rest))
```
byroot, line 1044: I don't think `concat` is the right name for this. Because on array `concat` is a mutating method, so the name is confusing.

dfl: good point! How about extending `WhereClause#+` to accept both types, then no new method is needed?                                                                                        
```ruby                                                                                                                                                                                          
  def +(other)                                                                                                                                                                            
    other_predicates = case other                                                                                                                                                         
      when WhereClause then other.predicates                                                                                                                                              
      when Array then other                                                                                                                                                               
      else raise ArgumentError, "expected WhereClause or Array, got #{other.class}"                                                                                                       
    end                                                                                                                                                                                   
    WhereClause.new(predicates + other_predicates)                                                                                                                                        
  end
  ```

Comment (byroot) [reply activerecord/lib/active_record/relation/query_methods.rb:1044]: I'd rather not add an extra type check if the goal is optimizing. Maybe just `.add_predicates`.

Comment (dfl) [reply activerecord/lib/active_record/relation/query_methods.rb:1044]: done

**Comment by hmcguire-shopify** [comment]
From the linked issue's benchmark, I'm not seeing much change:

```
             strings     1.476k i/100ms
             with_ar   672.000 i/100ms
             strings     14.431k (± 1.9%) i/s   (69.30 μs/i) -     72.324k in   5.013596s
             with_ar      6.570k (± 1.2%) i/s  (152.20 μs/i) -     32.928k in   5.012319s

Comparison:
      strings before:    14697.1 i/s
             strings:    14431.0 i/s - same-ish: difference falls within error
             with_ar:     6570.4 i/s - 2.24x  slower
      with_ar before:     6492.4 i/s - 2.26x  slower
```

The `simple_attributes?` method in particular feels very complex (and can easily fall out of sync with the `PredicateBuilder`) for what appears to be not much gained.

File: activerecord/lib/active_record/relation/query_methods.rb [hunk 2]
```diff
@@ -1041,7 +1041,7 @@ def where(*args)
     end
 
     def where!(opts, *rest) # :nodoc:
-      self.where_clause += build_where_clause(opts, rest)
+      self.where_clause = where_clause.add_predicates(build_where_clause_parts(opts, rest))
```
hmcguire-shopify: Since we're inside `where!`, can the predicate addition be a mutation instead of returning a new `WhereClaude`? Basically, the original `.concat` idea but change the implementation to match the name instead of the name to match the implementation

dfl:
> From the linked issue's benchmark, I'm not seeing much change:

@hmcguire-shopify the aim was to reduce GC pressure rather than enhance speed. The benchmark in this PR description shows allocation reduction, not throughput improvement.

>  | Query | Main | concat only | Full PR |                                                                                                                                                
> |-------|------|-------------|---------|                                                                                                                                                
>   | Simple where (1 col) | 33 | 31 (-6%) | 28 (-15%) |                                                                                                                                    
>   | Simple where (2 cols) | 41 | 40 (-2%) | 34 (-17%) |                                                                                                                                   
>   | Chained simple wheres | 56 | 52 (-7%) | 44 (-21%) |                                                                                                                                   
>   | Chain of 4 | 73 | 69 (-5%) | 61 (-16%) |                                                                                                             
>   Without simple_attributes?: ~2-7% savings (just from concat mutating)                                                                                                                   
>                                                                                                                                                                                           
>   With simple_attributes?: ~15-21% savings                                                                                                                                                
>                                                                                                                                                                                           
>   So simple_attributes? accounts for about 2/3 of the gains. Without it, we'd save 2-4 objects per query instead of 5-12.                                                                 
>                                                                                                                                                                                           
>   The concat mutation alone is probably not worth a PR. The simple_attributes? fast path is where the real savings are - but that's also the controversial part.                          
>

activerecord/lib/active_record/relation/query_methods.rb
dfl: Great idea, this saves even more on allocation:

```ruby
def concat(other_predicates)                                                                                                                                                            
  if frozen?                                                                                                                                                                            
    WhereClause.new(predicates + other_predicates)                                                                                                                                      
  else                                                                                                                                                                                  
    @predicates.concat(other_predicates)                                                                                                                                                
    self                                                                                                                                                                                
  end                                                                                                                                                                                   
end                                                                                                                                                                                  
```                                                                                                                                                                                        
  
The `frozen?` check is needed because `WhereClause.empty` is a frozen singleton.
The overhead is minimal, and now the mutation path makes O(1) allocations instead of O(n) for chained `where`s (10→4 for 2 chained, 18→4 for 4 chained).

activerecord/lib/active_record/relation/where_clause.rb
```diff
@@ -15,6 +15,15 @@ def +(other)
         WhereClause.new(predicates + other.predicates)
       end
 
+      def concat(other_predicates)
+        if frozen?
+          WhereClause.new(predicates + other_predicates)
+        else
+          @predicates.concat(other_predicates)
```
byroot, line 22: I'm not convinced this is correct. `spawn` is pretty much an alias of `clone` so `@predicates` is shared with another `WhereClause` instance and we shouldn't be mutating it.

byroot:
> the aim was to reduce GC pressure rather than enhance speed.

Well, the point of reducing GC pressure, is to spend less time in GC, so to improve speed. But the Ruby GC improved a lot in the last few years so minor reductions in allocation no longer make a noticeable difference.

dfl:
maybe the juice isn't worth the squeeze here then 🤷‍♂️

byroot:
> maybe the juice isn't worth the squeeze here then 🤷‍♂️

Yeah, it's a balance between code complexity and perf gains. As this PR goes right now, the gain isn't measurable, and the code noticeably harder, so I wouldn't merge in this state.

Now if you do find some small allocation reduction that don't make the code substantially more complex, I'm alright with that.

PR #40792, opened 2020-12-10
Ignore strict loading violations on instances loaded through fixtures

ghiculescu:
Fixes an issue reported here: https://discuss.rubyonrails.org/t/how-to-eager-load-fixture-associations-in-rails-6-1/76645

PR #56648, opened 2026-01-20
Fix strict loading fixture association macro

### LINKED ISSUE #56014: Can't disable strict loading if strict_loading option is specified in the association macro

yujiteshima:
### Motivation / Background

This Pull Request has been created because `strict_loading!(false)` does not override the `strict_loading: true` option when it is set via association macros.

Fixes #56373
Fixes #56014

When `strict_loading: true` is specified in an association macro (e.g., `has_many :comments, strict_loading: true`), calling `strict_loading!(false)` on the owner record has no effect - the association still raises `StrictLoadingViolationError` when lazily loaded.

This also affects fixture records, which explicitly set `@strict_loading = false` (added in #40792), but still raise `StrictLoadingViolationError` when accessing associations with `strict_loading: true` in the macro.

### Detail

This Pull Request changes `@strict_loading` to use a tri-state approach:

- **undefined**: use the association macro's `strict_loading` setting (backward compatible)
- **true**: explicitly enabled via `strict_loading!`, but association macro's `strict_loading: false` is still respected (for ActiveStorage compatibility)
- **false**: explicitly disabled via `strict_loading!(false)`, takes precedence over macro setting

Changes:
1. `ActiveRecord::Core#strict_loading?` - Returns `@strict_loading` if defined, otherwise falls back to `strict_loading_by_default`
2. `ActiveRecord::Core#init_internals` - Removes the initialization of `@strict_loading` to allow undefined state
3. `ActiveRecord::Associations::Association#violates_strict_loading?` - Uses tri-state logic:
   - If `@strict_loading` is explicitly `false`, returns `false` (no violation)
   - If `@strict_loading` is explicitly `true`, respects association macro's `strict_loading: false` setting (for ActiveStorage)
   - If `@strict_loading` is undefined, uses association macro setting or falls back to `strict_loading_by_default`

### Additional information

**Relationship to #56015:**

This PR implements the same tri-state approach as #56015 (authored by @otrfrn). PR #56015 has been open since October 2025 with CI failures that appear to be environment-related (activestorage tests). After rebasing on the current main branch, all tests pass successfully.

Since #56015 has been inactive and the CI issues were not addressed, I've created this PR with the same approach to move the fix forward. Credit to the original approach goes to @otrfrn.

Related discussion: https://github.com/rails/rails/issues/56373#issuecomment-3753967150

### Checklist

* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,33 @@
+*   Fix `strict_loading!(false)` to override the `strict_loading: true` option
+    in association macros.
+
+    Previously, when `strict_loading: true` was set on an association (e.g.,
+    `has_many :comments, strict_loading: true`), calling `strict_loading!(false)`
+    on the owner record had no effect - the association would still raise
+    `StrictLoadingViolationError` when lazily loaded.
+
+    Now, explicitly setting `strict_loading!(false)` on the owner takes precedence
+    over the association macro's `strict_loading: true` setting. This also fixes
+    strict loading violations on fixture records, which have `@strict_loading = false`
+    set explicitly.
+
+    ```ruby
+    class Post < ActiveRecord::Base
+      has_many :comments, strict_loading: true
+    end
+
+    post = Post.first
+    post.comments.to_a # => StrictLoadingViolationError (association macro setting)
+
+    post.strict_loading!(false)
+    post.comments.to_a # => OK (explicit setting takes precedence)
+    ```
+
+    Fixes #56373.
+    Fixes #56014.
+
+    *Yuji Teshima*
+
 *   Fix PostgreSQL schema dumping to handle schema-qualified table names in foreign_key references that span different schemas.
 
         # before
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -286,8 +286,20 @@ module ActiveRecord
 
           return unless owner.validation_context.nil?
 
+          # Check owner's explicit strict_loading setting (tri-state)
+          # If explicitly set to false, it takes precedence over association macro setting
+          strict_loading = owner.instance_variable_get(:@strict_loading)
+          unless strict_loading.nil?
+            return false if strict_loading == false
+            # strict_loading is explicitly true, but association macro can override it
+            return reflection.strict_loading? if reflection.options.key?(:strict_loading)
+            return !owner.strict_loading_n_plus_one_only?
+          end
+
+          # @strict_loading is undefined, use association macro setting if present
           return reflection.strict_loading? if reflection.options.key?(:strict_loading)
 
+          # Fall back to strict_loading_by_default
           owner.strict_loading? && !owner.strict_loading_n_plus_one_only?
         end
 
--- a/activerecord/lib/active_record/core.rb
+++ b/activerecord/lib/active_record/core.rb
@@ -686,7 +686,9 @@ module ActiveRecord
 
     # Returns +true+ if the record is in strict_loading mode.
     def strict_loading?
-      @strict_loading
+      return @strict_loading if defined?(@strict_loading)
+
+      self.class.strict_loading_by_default
     end
 
     # Sets the record to strict_loading mode. This will raise an error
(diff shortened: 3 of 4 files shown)


callmesangio:
Bumping this up as the original reporter of #56373.
Is there something actionable that could be done to get this reviewed and moved forward?

eileencodes (changes requested):
This isn't the right way to fix this problem - reaching into `instance_variable_get` in the framework indicates functionality is missing, it shouldn't be used here. 

I'm not sure this is a bug that should be fixed on this level if the problem only exists in fixtures. Fixing at the fixture level to skip strict loading makes more sense to me.

That said, I don't really have the bandwidth to dig into this or come up with a better solution. We're getting a lot of AI generated changes that fix the problem and look plausible but aren't inline with the framework's sensibilities. It now takes someone 30s to generate a PR but it takes us the same amount of time to review. Except now there are many more PRs and many more hackerone reports and many more issues, but the Core team is the same size.

callmesangio:
> This isn't the right way to fix this problem - reaching into `instance_variable_get` in the framework indicates functionality is missing, it shouldn't be used here.

Yes, I agree with you on this. Explicitly exposing the relevant bits of state (or building an abstraction around them) would be the right thing to do here.

> I'm not sure this is a bug that should be fixed on this level if the problem only exists in fixtures. Fixing at the fixture level to skip strict loading makes more sense to me.

Unfortunately this is not limited to fixtures, it extends to `strict_loading!(false)` having no effect when an association is set to be strictly loaded via macro.

> That said, I don't really have the bandwidth to dig into this or come up with a better solution. We're getting a lot of AI generated changes that fix the problem and look plausible but aren't inline with the framework's sensibilities. It now takes someone 30s to generate a PR but it takes us the same amount of time to review. Except now there are many more PRs and many more hackerone reports and many more issues, but the Core team is the same size.

Totally understandable. I can't really tell about this specific PR, but I definitely get the general frustration of OSS maintainers about the current trends.
I may take a stab at this issue personally if I manage to find the time and this PR doesn't move forward. Thanks!

eileencodes:
> I may take a stab at this issue personally if I manage to find the time and this PR doesn't move forward. 

Please do. And I'm not against AI as a starting point for patches but what is submitted must be in line with how the framework is designed.

yujiteshima:
Thank you for the review, @eileencodes.

You're right that using instance_variable_get directly is not the right approach.
I'll rework this to add a proper accessor method (e.g., strict_loading_explicitly_disabled?)
to ActiveRecord::Core that exposes the tri-state logic cleanly, removing the
instance_variable_get call from association.rb.

Also, I'd like to clarify that while the fixture case is one symptom,
the underlying issue (strict_loading!(false) having no effect when an association
macro sets strict_loading: true) affects non-fixture records as well —
as noted by @callmesangio in the thread.

I'll revise the PR with a cleaner approach.

PR #49771, opened 2023-10-24
Reimplement TaggedLogging:

Edouard-chin:
- ### Context

  The Tagged Logging functionality has been a source of a few issues over the years, especially when combined with the broadcasting feature. Initializating a Tagged Logger was also not very intuitive:

  ```ruby 
  logger = Logger.new(STDOUT)
  tagged_logger = ActiveSupport::TaggedLogging.new(logger)
  # Would expect the `tagged_logger` to be an instance of `AS::TaggedLogging` 
  # but it's a clone of the `logger`.

  tagged_logger.formatter = ->(_, _, _, message) do
    { message: message }
  end
  # Modifies the formatter to output JSON formatted logs.
  # This breaks tagged logging.
  ```

  I believe the main reason of those issues is because tagged logging is implemented at the wrong level.

  ### Solution

  I made a proposal on the Ruby logger upstream in ruby/logger#90 to help solve this but it hasn't been reviewed yet. So I thought about adding it here for now. The TL;DR is to decouple formatting and adding extra information to logs (which is what tagged logging does).

  ### Deprecation

  Since TaggedLogging will no longer access the formatter, there is a few things I'd like to deprecate such as setting a default formatter https://github.com/rails/rails/blob/d68e43922bc11829c52ad9f736ad5549fc97631b/activesupport/lib/active_support/tagged_logging.rb#L124 but doing so in this PR would increase the size of the diff significantly and would maybe distract for PR reviews.

  Another thing that I believe should be deprecated is `ActiveSupport::TaggedLogging.new`. Adding tagging functionality to a logger should be done using a more ruby approach such as `logger.extend(AS::TaggedLogging)`.

-------------------

Fix #49757 
Fix #49745 
Fix #46084 
Fix #44668
Fix #52876

PR #56650, opened 2026-01-20
Fix BroadcastLogger to execute blocks only once with multiple destinations

QuirkQ:
### Motivation / Background

  When using `BroadcastLogger` with multiple destinations, calling methods like `tagged` (which go through `method_missing`) would execute the provided block once for each destination logger. This caused problems in real-world usage - for
  example, when using `BroadcastLogger` with `ActiveJob`, callbacks would fire multiple times instead of once.

  The `dispatch` method already had logic to prevent this, but `method_missing` didn't apply the same safeguard.

  ### Detail

  This PR extracts the block-wrapping logic from `dispatch` into a new private method `wrap_block_for_broadcast` and applies it consistently in both `dispatch` and `method_missing`. The wrapper ensures blocks execute only once, with the
  result cached and returned to subsequent loggers.

  Changes:
  - Added `wrap_block_for_broadcast` helper method
  - Updated `method_missing` to wrap blocks before broadcasting
  - Added tests verifying blocks execute once regardless of destination count
  - Updated CHANGELOG

  ### Checklist


QuirkQ:
This PR addresses the same core issue as rails/rails#53093 (which was merged) - ensuring blocks execute only once when broadcasting to multiple loggers.

  The approach here extracts the block-wrapping logic into a dedicated `wrap_block_for_broadcast` helper method and applies it consistently in both `dispatch` and `method_missing`. This prevents duplicate callback executions when using
  features like `tagged` with multiple broadcast destinations.

  **Related work:**
  - rails/rails#53105 - Adds proper TaggedLogging support for BroadcastLogger (currently open, would complement this fix nicely)
  - rails/rails#49771 - Broader TaggedLogging rewrite that also addresses these issues
  - appsignal/appsignal-ruby#1357 - AppSignal had to implement their own broadcasting solution to work around these Rails issues

  The original merged fix (rails/rails#53093) had one subtle breaking change noted by @toncid - it changed `dispatch` to return whatever the first logger returns instead of always returning `true`. Worth keeping in mind for compatibility.

  This fix is particularly important for ActiveJob users, where duplicate callback execution can cause real problems in production.

QuirkQ:
LLM written replication script:
```ruby
# frozen_string_literal: true

# BroadcastLogger bug demo: blocks execute N times for N destinations

require "bundler/inline"

RAILS_VERSION = "~> 8.1"

gemfile do
  source "https://rubygems.org"
  gem "activesupport", RAILS_VERSION
  gem "activejob", RAILS_VERSION
end

require "stringio"
require "active_support/all"
require "active_job"

puts "\nBroadcastLogger block duplication demo"
puts "=" * 50

def make_tagged_logger
  ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(StringIO.new))
end

def report(label, count)
  puts "#{label}: block ran #{count} time(s) #{count == 1 ? "✅" : "❌"}"
end

# --- Basic reproduction ------------------------------------------------------

[2, 4].each do |n|
  broadcast = ActiveSupport::BroadcastLogger.new(*Array.new(n) { make_tagged_logger })
  count = 0
  broadcast.tagged("TEST") { count += 1 }
  report("#{n} destinations", count)
end

# --- Real-world impact: ActiveJob -------------------------------------------

puts "\nActiveJob impact"
puts "-" * 50

ActiveJob::Base.queue_adapter = :test

class PaymentJob < ActiveJob::Base
  def perform(order_id); end
end

[2, 4].each do |n|
  broadcast = ActiveSupport::BroadcastLogger.new(*Array.new(n) { make_tagged_logger })
  broadcast.formatter = broadcast.broadcasts.first.formatter

  ActiveJob::Base.logger = broadcast
  ActiveJob::Base.queue_adapter.enqueued_jobs.clear
  PaymentJob.perform_later("ORDER-#{n}")

  jobs = ActiveJob::Base.queue_adapter.enqueued_jobs.size
  puts "#{n} destinations: enqueued #{jobs} #{jobs == 1 ? "✅" : "❌"}"
end

# --- Workaround: wrap BroadcastLogger in TaggedLogging ----------------------

puts "\nWorkaround (TaggedLogging wraps BroadcastLogger)"
puts "-" * 50

[2, 4].each do |n|
  io1 = StringIO.new
  io2 = StringIO.new
  logger1 = ActiveSupport::Logger.new(io1)
  logger2 = ActiveSupport::Logger.new(io2)

  broadcast = ActiveSupport::BroadcastLogger.new(logger1, logger2)
  wrapped = ActiveSupport::TaggedLogging.new(broadcast)

  count = 0
  wrapped.tagged("TEST") { count += 1 }

  puts "#{n} destinations: block ran #{count} time(s) #{count == 1 ? "✅" : "❌"}"
  wrapped.tagged("ORDER") { wrapped.info "Processing payment" }
  puts "  Logger 1 received logs: #{io1.string.include?("Processing payment") ? "✅" : "❌"}"
  puts "  Logger 2 received logs: #{io2.string.include?("Processing payment") ? "✅" : "❌"}"
  puts "  Tags applied correctly: #{io1.string.include?("[ORDER]") ? "✅" : "❌"}"
end

puts "\nWhy it works:"
puts "  - TaggedLogging#tagged runs the block before BroadcastLogger sees it"
puts "  - Only one block execution, then normal broadcasted logging inside"

# --- Minimal fix shown in the PR --------------------------------------------

puts "\nFix (method_missing):"
puts "-" * 50
puts <<~RUBY

  def method_missing(name, *args, **kwargs, &block)
    loggers = @broadcasts.select { |logger| logger.respond_to?(name) }

    if loggers.one?
      loggers.first.send(name, *args, **kwargs, &block)
    else
      block = wrap_block_for_broadcast(&block) if block_given?
      loggers.map { |logger| logger.send(name, *args, **kwargs, &block) }
    end
  end

  def wrap_block_for_broadcast(&block)
    called, result = false, nil
    proc { |*args, **kwargs|
      called ? result : (called = true; result = yield(*args, **kwargs))
    }
  end

RUBY

```

PR #56652, opened 2026-01-21
Explicitly use Marshal serializer for MemCacheStore Dalli client

zzak:
Fixes #56588

We can assume the default Marshal serializer will be used because the user can never change it, we remove it from the options before init.

Therefore, we can prevent Dalli from warning users about auto-selecting the default Marshal serializer.

Since we previously dropped the serializer option before building the client, users were unable to override the default.

Users are able to pass silence_marshal_warning option to work around this issue as well.

Here is the bug report:

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "dalli", "4.0.0"
end

require "active_support"
require "active_support/cache"
require "active_support/cache/mem_cache_store"
require "minitest/autorun"

class MemCacheTest < ActiveSupport::TestCase
  def setup
    @cache = ActiveSupport::Cache::MemCacheStore.new("memcached:11211")
  end

  def test_memcached_write_and_read
    key = "my_test_key"
    value = "Hello, Dalli!"

    write_success = @cache.write(key, value)
    assert write_success, "Failed to write to cache"

    cached_value = @cache.read(key)

    assert_equal value, cached_value
  end
end
```

Dalli 4.0.0 changelog: https://github.com/petergoldstein/dalli/commit/48290a73

### DIFF
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -90,6 +90,9 @@ module ActiveSupport
         # The value "compress: false" prevents duplicate compression within Dalli.
         @mem_cache_options[:compress] = false
         (OVERRIDDEN_OPTIONS - %i(compress)).each { |name| @mem_cache_options.delete(name) }
+        # Set the default serializer for Dalli to prevent warning about
+        # inheriting the default serializer.
+        @mem_cache_options[:serializer] = Marshal
         @data = self.class.build_mem_cache(*(addresses + [@mem_cache_options]))
       end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -181,7 +181,8 @@ GEM
     crass (1.0.6)
     cssbundling-rails (1.4.1)
       railties (>= 6.0.0)
-    dalli (3.2.8)
+    dalli (4.0.0)
+      logger
     dante (0.2.0)
     dartsass-rails (0.5.1)
       railties (>= 6.0.0)
--- a/activesupport/test/cache/cache_store_setting_test.rb
+++ b/activesupport/test/cache/cache_store_setting_test.rb
@@ -28,7 +28,7 @@ class CacheStoreSettingTest < ActiveSupport::TestCase
   end
 
   def test_mem_cache_fragment_cache_store
-    assert_called_with(Dalli::Client, :new, [%w[localhost], { compress: false }]) do
+    assert_called_with(Dalli::Client, :new, [%w[localhost], { compress: false, serializer: Marshal }]) do
       store = ActiveSupport::Cache.lookup_store :mem_cache_store, "localhost", pool: false
       assert_kind_of(ActiveSupport::Cache::MemCacheStore, store)
     end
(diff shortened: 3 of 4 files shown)


activesupport/lib/active_support/cache/mem_cache_store.rb
```diff
@@ -90,6 +90,9 @@ def initialize(*addresses)
         # The value "compress: false" prevents duplicate compression within Dalli.
         @mem_cache_options[:compress] = false
         (OVERRIDDEN_OPTIONS - %i(compress)).each { |name| @mem_cache_options.delete(name) }
+        # Set the default serializer explicitly unless otherwise specified,
+        # to prevent Dalli from warning about inheriting the default serializer.
+        @mem_cache_options[:serializer] = Marshal
```
byroot, line 95: Do we even need `Marshal` ? I need to double check but we have our own encoders, so I think we only ever get and send strings to Dalli.

zzak: My understanding is that we pass whatever to Dalli, and it serializes that value with Marshal currently.
https://github.com/petergoldstein/dalli/blob/main/lib/dalli/protocol/value_serializer.rb

Since we explicitly remove the serializer option that users might pass in with:

```
(OVERRIDDEN_OPTIONS - %i(compress)).each { |name| @mem_cache_options.delete(name) }
```

`OVERRIDEN_OPTIONS` ([defined here](https://github.com/rails/rails/blob/5faaa59f5b45d4c1e1b966dbfb0dded0e1996130/activesupport/lib/active_support/cache/mem_cache_store.rb#L33-L35)) is just `UNIVERSION_OPTIONS` which includes `:serializer`:
https://github.com/rails/rails/blob/5faaa59f5b45d4c1e1b966dbfb0dded0e1996130/activesupport/lib/active_support/cache.rb#L36

So we're basically saying, don't let the user override the Dalli serializer and pass in the default to avoid this warning:
https://github.com/petergoldstein/dalli/blob/3aae91bf2430c9e43f41120db60e7f51abd102c8/lib/dalli/protocol/value_serializer.rb#L88-L89

If that's really the case, there is what, double serialization happening? Our serialzation format in a string -> Dalli Marshal?

byroot: Ok, so we may not be able to just remove the serialization layer from Dalli as it may cause issue on the next Rails upgrade. Maybe it's possible but we need to study the consequences:

  - What happens to new version reading old format
  - What happens to old version reading new format

If in both case it result in a miss, we could consider just moving forward, but either way it's probably best to have a flag to allow a seemless upgrade.

Also I think we should submit a PR to dalli to get a simple raw client (or an option) as we keep being bit by Dalli's serialization / compression even though we don't want that feature.

byroot: > there is what, double serialization happening? Our serialzation format in a string -> Dalli Marshal?

Depends on the dalli version. Recent ones won't marshal strings: https://github.com/petergoldstein/dalli/pull/1041

byroot: Ah shit, it's not enabled by default :/

byroot:
Ok, merging this to fix the warning and as a safe change to backport.

But moving forward we should entirely bypass Dalli serialization.

PR #56653, opened 2026-01-21
Change config.cache_store note to correct default

AlexMooney:
### Motivation / Background

The environments/production template incorrectly says "Replace the _default in-process memory cache store_ with a durable alternative."  The default is `file_store`: https://github.com/rails/rails/blob/428e3d4f1e7218e4965f35cfe75f08ec0808883a/railties/lib/rails/application/configuration.rb#L58

### Detail

Rewrite the comment and the commented out default cache_store value to reflect `file_store`.

#### Open questions

Should we change https://github.com/rails/rails/blob/bef5e41cbfbba750706b05d21e10955f7c8beb45/railties/lib/rails/generators/rails/app/templates/config/environments/development.rb.tt#L31 to also use `file_store`?

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
+++ b/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
@@ -58,8 +58,8 @@ Rails.application.configure do
   # Don't log any deprecations.
   config.active_support.report_deprecations = false
 
-  # Replace the default in-process memory cache store with a durable alternative.
-  # config.cache_store = :mem_cache_store
+  # Replace the default file system cache store with a durable alternative.
+  # config.cache_store = :file_store
 
   <%- unless options[:skip_active_job] -%>
   # Replace the default in-process and non-durable queuing backend for Active Job.


railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt
```diff
@@ -58,8 +58,8 @@ Rails.application.configure do
   # Don't log any deprecations.
   config.active_support.report_deprecations = false
 
-  # Replace the default in-process memory cache store with a durable alternative.
```
zzak, line 61: I think the comment is meant to encourage memcache as the durable alternative so the comment is correct.

Just that the default is not in-memory, but file system...?

zzak:
> Should we change `development` to also use `file_store`?

This was just changed in 9c8716a during Rails 8.0 for rate-limiting, so no I don't think we should change it already.

byroot:
I think you misunderstood something because the existing comment is correct and your proposed modification isn't.

PR #56654, opened 2026-01-21
Add support for scoped preloading with select in associations

ItaloCobains:
### Motivation / Background

At the company I currently work for, we frequently use preload and includes on tables with many columns. This creates performance issues because ActiveRecord loads all columns even when we only need two or three fields in many cases.

Being able to limit the selected columns for preloaded or included associations would significantly improve query performance and memory usage.

Currently, developers can control column selection for the main query using `select`, but there's no way to do the same for preloaded associations:

```ruby
# Can select columns for main query
User.select(:id, :name).where(active: true)

# But cannot control columns for associations - loads ALL columns
User.preload(:books).where(active: true).select(:id, :name)
```

This is particularly problematic for:
- API endpoints returning JSON with limited fields
- Large tables with text/blob columns that are rarely needed
- High-traffic applications where every KB of data transfer matters

### Detail

This Pull Request adds support for scoped preloading with column selection, allowing developers to specify which columns to load for preloaded associations using lambda expressions or `ActiveRecord::Relation` objects.

**New API:**

```ruby
# Lambda with select
User.preload(books: -> { select(:id, :title) })

# Relation with select
User.preload(books: Book.select(:id, :title))

# Lambda with where and select
User.preload(books: -> { where(published: true).select(:id, :title) })

# Multiple associations with different scopes
User.preload(
  books: -> { select(:id, :title) },
  comments: -> { where(visible: true).select(:id, :body) }
)

# Nested associations with scopes
User.preload(
  books: {
    scope: -> { select(:id, :title) },
    chapters: -> { select(:id, :name) }
  }
)
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Add support for scoped preloading with `preload`.
+
+    You can now apply scopes (including `select`) to preloaded associations
+    using a lambda or an `ActiveRecord::Relation`:
+
+    ```ruby
+    User.preload(posts: -> { select(:id, :title) })
+    # SELECT "users".* FROM "users"
+    # SELECT "posts"."id", "posts"."title", "posts"."user_id" FROM "posts" WHERE ...
+    ```
+
+    Required columns (primary key, foreign key, type columns for STI/polymorphic)
+    are automatically included even if not specified in `select`.
+
+    For nested associations with scopes, use the `:scope` key:
+
+    ```ruby
+    User.preload(posts: { scope: -> { where(published: true) }, comments: :author })
+    ```
+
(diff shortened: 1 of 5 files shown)


felixefelip:
This looks a nice API to me, but alternatively creating an association with scope is feasible to solve this problem?

```ruby
  has_many :books
  has_many :books_with_select_columns, -> { select(:id, :name }, class_name: "Book"
end

# preload will only select id and name
User.preload(:books_with_select_columns)
```

ItaloCobains:
@felixefelip I got your solution, but in large codebase is unfeasible to write a new relation ever that I need a new select on preload. With it, I'll so many line of relation on my models and this not sounds a good pattern to me.

PR #56636, opened 2026-01-17
Detect JS package manager from lockfiles in generators

dfl:
### Motivation / Background                                                                                                                                                             
                                                                                                                                                                                        
Rails generators currently hardcode `yarn` for JavaScript package management, but many teams use npm or pnpm. This forces developers to manually update generated files after running   
generators like `actiontext:install`.                                                                                                                                                   
                                                                                                                                                                                        
Fixes #47416                                                                                                                                                                            
                                                                                                                                                                                        

                                                                                                                                                                                        
### Detail                                                                                                                                                                              
                                                                                                                                                                                        
Generators now detect the package manager from lockfiles:                                                                                                                               
                                                                                                                                                                                        
- `bun.lockb` → bun                                                                                                                                                                     
- `pnpm-lock.yaml` → pnpm                                                                                                                                                               
- `package-lock.json` → npm                                                                                                                                                             
- `yarn.lock` or default → yarn                                                                                                                                                         
                                                                                                                                                                                        
This fixes ActionText, ActionCable, and other generators that install JS dependencies.                                                                                                  
                                                                                                                                                                                        
### Additional information                                                                                                                                                              
                                                                                                                                                                                        
The implementation introduces `Rails::Generators::JsPackageManager`, a small module (~80 lines) that centralizes:                                                                       
                                                                                                                                                                                        
- Detection logic                                                                                                                                                                       
- Command strings (`add`, `install`, `audit`)                                                                                                                                           
- Lockfile names

PR #56655, opened 2026-01-22
Detect bun JS package manager when using text-based lockfile

bai:
### Motivation / Background

https://github.com/rails/rails/pull/56636 streamlined JS package manager detection to be based on presence of lockfiles, however for `bun` it only looks for `bun.lockb`, which is an older binary format bun no longer uses by default. In this PR I add  a check to detect bun based on presence of `bun.lock` too.

### Detail

This Pull Request changes JS lockfile detection to consider a newer lockfile format for bun.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/js_package_manager.rb
+++ b/railties/lib/rails/generators/js_package_manager.rb
@@ -33,7 +33,7 @@ module Rails
       }.freeze
 
       def self.detect(root)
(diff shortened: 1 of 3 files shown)


bai:
@guilleiguaran - submitted this bun.lock fix per your comment https://github.com/rails/rails/pull/56636#issuecomment-3782501451

PR #56659, opened 2026-01-22
Fix CONTENT_TYPE header not cleared between ActionController::TestCase requests

### LINKED ISSUE #55759: CONTENT_TYPE header is not cleared between ActionController::TestCase requests
### Steps to reproduce

- Create an `ActionController::TestCase` test
- Perform a `get` request test after a `post` request test
- Notice the `get` requests has the wrong content type

Sample application https://github.com/arielj/content_type_is_not_cleared

Check https://github.com/arielj/content_type_is_not_cleared/commit/99eed7080c886d8cd421ba663362178bb60041f3 for the details.

- Clone the repo
- Run `rails test`
- Notice this is printed to the console:

```
GET request. Content type: 
POST request. Content type: application/x-www-form-urlencoded
GET request. Content type: application/x-www-form-urlencoded
```

### Expected behavior

The content type should be cleared for the get request

### Actual behavior

The content type is whatever was set by the previous non-get request.

This was not always creating issues for us, but with an update of the Rack gem and some changes in Rails 7.2 (producing `content_length == nil` instead of `content_length == 0`), if a POST request is done with multipart data and then, after that, a GET request is done, the incorrect content_type header makes Rack attempt to [parse a non-existent multipart content](https://github.com/rack/rack/blob/14c8731436785d7e79a4db0f3304769a26083182/lib/rack/request.rb#L513) during the GET request because `form_data?` is true, incorrectly.

My understanding is that the [`scrub_env!` method](https://github.com/rails/rails/blob/2ef3103c8d79eae9ccb4ab32691fd072ea932537/actionpack/lib/action_controller/test_case.rb#L671-L672) should be changed to also delete `CONTENT_TYPE` from the env.

If anyone can confirm this I can make a PR changing this.

### System configuration

**Rails version**: 8.0.3 for the sample app, but we identified it at least since rails 6.1

**Ruby version**: 3.4.5 for the sample app, but we identified it at least since Ruby 3.1

Saidbek:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/55759

The `CONTENT_TYPE` header was persisting between test requests in `ActionController::TestCase`, causing Rack to incorrectly attempt to parse multipart content during GET requests that followed POST requests.

This PR clears the `CONTENT_TYPE` header in the `process_controller_response` method's ensure block, alongside other request state cleanup (XHR headers and query_string).

### Additional information

Updated tests to check headers during requests (using `test_headers` action) rather than after requests complete, as headers are not guaranteed to remain unmodified after the request finishes.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/test_case.rb
+++ b/actionpack/lib/action_controller/test_case.rb
@@ -661,6 +661,7 @@ module ActionController
               @request.delete_header "HTTP_ACCEPT"
             end
             @request.query_string = ""
+            @request.delete_header "CONTENT_TYPE"
 
             @response.sent!
           end
--- a/actionpack/test/controller/test_case_test.rb
+++ b/actionpack/test/controller/test_case_test.rb
@@ -635,9 +635,12 @@ class TestCaseTest < ActionController::TestCase
   end
 
   def test_using_as_json_sets_request_content_type_to_json
-    post :render_body, params: { bool_value: true, str_value: "string", num_value: 2 }, as: :json
+    post :test_headers, params: { bool_value: true, str_value: "string", num_value: 2 }, as: :json
 
-    assert_equal "application/json", @request.headers["CONTENT_TYPE"]
+    env = ::JSON.parse(@response.body)
+    assert_equal "application/json", env["CONTENT_TYPE"]
+
+    post :render_body, params: { bool_value: true, str_value: "string", num_value: 2 }, as: :json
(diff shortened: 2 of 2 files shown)


Saidbek:
hey @guilleiguaran, just following up here, could you pls check once you have some time? Thanks

Saidbek:
hey @guilleiguaran, soft ping, would you mind reviewing it? i've reproduced and raised a fix here, thanks!

PR #56664, opened 2026-01-22
Fix inverse_of on composite keys in ActiveRecord::Associations::Association#inversable? 

masom:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/56662

The issue posted a simple replication script, including a monkeypatch that would address the issue. This PR refactors that patch into an integral part of the `ActiveRecord::Associations::Association#matches_foreign_key?` method.

### Detail

Composite keys would fail on the `record.read_attribute` and `record._has_attribute?` checks as this would pass in an array to a method expecting a string.

Each key component needs to be read one-by-one.

### Checklist


activerecord/lib/active_record/associations/association.rb
```diff
@@ -409,11 +409,20 @@ def inversable?(record)
...
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
+          reflection_fk = reflection.foreign_key
+
+          if reflection_fk.is_a?(Array)
+            reflection_pk = Array(reflection.association_primary_key)
+
+            owner_fk_values = reflection_fk.map { |key| owner.read_attribute(key) }
+            record_pk_values = reflection_pk.map { |key| record.read_attribute(key) }
+
+            owner_fk_values == record_pk_values
```
masom: Claude w/ Opus flagged there is potential for the second branch `(foreign_key_for?(owner) && owner.read_attribute(reflection_fk) == record.id` but I've not been able to replicate/trigger this.

After letting it churn on multiple approaches, the only trigger was to grab the association:
```ruby
  def test_matches_foreign_key_for_has_many_with_composite_key
    # This test directly exercises the second branch in matches_foreign_key?
    # which checks if record.FK == owner.PK (for has_many associations)
    order = cpk_orders(:cpk_groceries_order_1)
    store_id, _order_id = order.id
    book = order.books.create!(id: [store_id, 7], title: "Book")

    # Get the has_many association and test matches_foreign_key? directly
    has_many_assoc = order.association(:books)

    # For has_many, owner is the order, and we're checking if book's FK matches order's PK
    # This exercises the second branch because:
    # - First branch checks owner.FK == record.PK (order doesn't have FK columns)
    # - Second branch checks record.FK == owner.id (book's FK should match order's PK)
    assert has_many_assoc.send(:inversable?, book), "has_many should recognize book as inversable via composite FK"
  end
```

I wasn't able to find a proper use case for this, and left the second branch out of this PR.

activerecord/lib/active_record/associations/association.rb
```diff
@@ -409,11 +409,20 @@ def inversable?(record)
         end
 
         def matches_foreign_key?(record)
-          if foreign_key_for?(record)
-            record.read_attribute(reflection.foreign_key) == owner.id ||
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
+          reflection_fk = reflection.foreign_key
+
+          if reflection_fk.is_a?(Array)
+            reflection_pk = Array(reflection.association_primary_key)
```
adrianna-chang-shopify: Should we verify that the attributes exist, like `foreign_key_for?` is currently checking?

```ruby
  return false unless reflection_fk.all? { |key| owner._has_attribute?(key) }                                                                                                         
  return false unless reflection_pk.all? { |key| record._has_attribute?(key) }     
```

masom: Potentially? I'll give this a try although I'm not sure I understand the value the check provides.

The code would end up looking like this:

```ruby
      owner_fk_values = reflection_fk.map do |key|
        if owner._has_attribute?(key)
          owner.read_attribute(key)
        else
          return false
        end
      end

      record_pk_values = reflection_pk.map do |key|
        if record._has_attribute?(key)
          record.read_attribute(key)
        else
          return false
        end
      end

      owner_fk_values == record_pk_values
```

With an early return if an attribute isn't found, and removing the `compact` calls.

adrianna-chang-shopify: Ah, I'm looking at this with fresh eyes and you're totally right 🤦‍♀️ 

The check in `foreign_key_for?` isn't a guard for ensuring the attrs exist, it's to check whether the FK is on the `record` (we stick with the `if` branch), or whether the FK is on the `owner` side (in which case we drop into the `else` branch).

I also couldn't find a way to replicate the `else` branch for a CPK model either. We'd need to call `matches_foreign_key?` with `record` as `Cpk::Book`, `owner` as `Cpk::Order`, and the reflection being the `HasMany` reflection, but I can't find a way to simulate this in a test 😅 

So I think we're okay to keep your original `if` branch without those `has_attribute?` checks.

activerecord/lib/active_record/associations/association.rb
```diff
@@ -409,11 +409,23 @@ def inversable?(record)
         end
 
         def matches_foreign_key?(record)
-          if foreign_key_for?(record)
-            record.read_attribute(reflection.foreign_key) == owner.id ||
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
+          reflection_fk = reflection.foreign_key
+
+          if reflection_fk.is_a?(Array)
```
mbobin: Why do we need another branch here? To me it looks like the `else` branch, but with array values. What do you think of this refactoring? It updates both branches to use composite keys.

```ruby
  def matches_foreign_key?(record)
    if foreign_key_for?(record)
      record_foreign_key_matches_owner?(record) ||
        (foreign_key_for?(owner) && owner_foreign_key_matches_record?(record))
    else
      owner_foreign_key_matches_record?(record)
    end
  end

  def record_foreign_key_matches_owner?(record)
    foreign_key_values(record) == primary_key_values(owner)
  end

  def owner_foreign_key_matches_record?(record)
    foreign_key_values(owner) == primary_key_values(record)
  end

  def foreign_key_values(record)
    Array(reflection.foreign_key).map { |key| record.read_attribute(key) }
  end

  def primary_key_values(record)
    Array(reflection.association_primary_key).map { |key| record.read_attribute(key) }
  end
```

mbobin: Staring at it a bit more and the condition in `matches_foreign_key?(record)` doesn't make a lot of sense because the `else` branch is not guarded as the primary branch and if we add the guard it looks exactly as the second part of the primary branch. I think the whole thing should look like this:

```ruby
  def matches_foreign_key?(record)
    (foreign_key_for?(record) && record_foreign_key_matches_owner?(record)) ||
      (foreign_key_for?(owner) && owner_foreign_key_matches_record?(record))
  end
  ```

Comment (mbobin) [reply activerecord/lib/active_record/associations/association.rb]: `reflection.association_primary_key` blows up with polymorphic associations:

```
ArgumentError:
       Polymorphic associations do not support computing the class.
```

This made it work, but I'm not sure if it's 100% accurate.

```ruby
          def primary_key_values(record)
            Array(reflection.association_primary_key(record.class)).map { |key| record.read_attribute(key) }
          end
```

masom: Part of the reason for the branch is to keep non-composite primary keys and foreign keys lookup to be as fast as possible, and to only introduce extra allocations to arrays for models that use composite keys.

In a way this changeset is closer to the patch linked in the issue, vs completely changing lookups.

If we're good with introducing the extra allocation for all paths then the proposed refactor would work. Removing the branching would likely make tests cover more cases as we didn't hit an issue with polymorphic relationships in our codebase with this patch applied.

masom: This works in our application, the PR is now updated to this.

tylerwood-io:
For transparency: I submitted #57084 for the same issue (#56829) before finding this PR. The approaches converge on the same solution — \`Array(reflection.foreign_key).map\` for FK values and \`Array(reflection.association_primary_key(record.class)).map\` for PK values.

The key insight that makes this work for both has_many and belongs_to is using \`association_primary_key(record.class)\` rather than \`owner.id\` or \`record.id\` — the association-declared primary key columns can differ from the model's actual primary key when composite keys are specified on the association.

Your test using existing CPK fixtures is more integrated than my inline approach. Given you filed first, happy to close #57084 in favor of this one if core prefers that.

nvasilevski (approved):

mbobin (commented):
Thanks, @masom! I think it looks good!

eileencodes:
Thanks @masom - backported to 8-1-stable and 8-0-stable as well.

mbobin:
> Thanks @masom - backported to 8-1-stable and 8-0-stable as well.

@eileencodes would it be possible to do a backport to 7.2 too? 🙏

eileencodes:
We no longer support 7.2 for bug fixes as of 7 months ago, it is still receiving security patches until August 9 2026. We recommend upgrading because you only have 4 months until 7.2 is EOL.

PR #56665, opened 2026-01-22
Add &block params to `ActionController` methods using yield

dduugg:
### Motivation / Background

There are several methods in `ActionController` that use `block_given?` or `yield` without explicitly declaring a `&block` parameter. While this works at runtime, it creates issues for static analysis tools like Sorbet that check for blocks being passed to methods that don't declare them as parameters (see [sorbet/sorbet#9791](https://github.com/sorbet/sorbet/pull/9791)).

Adding explicit `&block` parameters also improves consistency, generated documentation, and user-friendliness, as many similar methods in the same files already declare `&block` (e.g., `each_pair(&block)`, `each_value(&block)`, `transform_keys(&block)`).

### Detail

This Pull Request changes method signatures to add explicit `&block` parameters to methods that use `block_given?` or `yield`:

**strong_parameters.rb:**
- `fetch(key, *args)` → `fetch(key, *args, &block)`
- `transform_values` → `transform_values(&block)`
- `transform_values!` → `transform_values!(&block)`
- `each_nested_attribute` → `each_nested_attribute(&block)`

**live.rb:**
- `send_stream(filename:, disposition:, type:)` → `send_stream(filename:, disposition:, type:, &block)`
- `new_controller_thread` → `new_controller_thread(&block)`

**instrumentation.rb:**
- `cleanup_view_runtime` → `cleanup_view_runtime(&block)`

**conditional_get.rb:**
- `http_cache_forever(public:)` → `http_cache_forever(public:, &block)`

**mime_responds.rb:**
- `respond_to(*mimes)` → `respond_to(*mimes, &block)`

### Additional information

This is a backward-compatible change. Adding an explicit `&block` parameter to a method that already uses `yield` or `block_given?` does not change runtime behavior—it simply makes the block parameter explicit in the method signature.

### Checklist

Note: The tests and CHANGELOG checkboxes are unchecked because this is a no-op change at runtime (no new behavior to test, no user-facing change to document). Let me know if you'd like me to adjust anything!

### DIFF
--- a/actionpack/lib/action_controller/metal/conditional_get.rb
+++ b/actionpack/lib/action_controller/metal/conditional_get.rb
@@ -318,7 +318,7 @@ module ActionController
     # *   `public`: By default, HTTP responses are private, cached only on the
     #     user's web browser. To allow proxies to cache the response, set `true` to
     #     indicate that they can serve the cached response to all users.
-    def http_cache_forever(public: false)
+    def http_cache_forever(public: false, &block)
       expires_in 100.years, public: public, immutable: true
 
       yield if stale?(etag: request.fullpath,
--- a/actionpack/lib/action_controller/metal/instrumentation.rb
+++ b/actionpack/lib/action_controller/metal/instrumentation.rb
@@ -96,7 +96,7 @@ module ActionController
       #     def cleanup_view_runtime
       #       super - time_taken_in_something_expensive
       #     end
-      def cleanup_view_runtime # :doc:
+      def cleanup_view_runtime(&block) # :doc:
         yield
       end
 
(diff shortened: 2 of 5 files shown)


bensheldon:
Can the analysis tools use anonymous block forwarding style instead? `def method(&)`

That would help preserve the signal that the argument is not explicitly used. As of Ruby 3.2 I don't think there is a performance difference, so purely stylistic.

dduugg:
> Can the analysis tools use anonymous block forwarding style instead? `def method(&)`
> 
> That would help preserve the signal that the argument is not explicitly used. As of Ruby 3.2 I don't think there is a performance difference, so purely stylistic.

This PR tried to carefully follow the existing style. For instance, `ActionController::Parameters` already has `def each_pair(&block)` and `def each_value(&block)`, although `block` is not directly referenced in the implementation of either. 

I'm new in this repo, but it's probably better to open a separate discussion / issue to propose a style rule rather than target individual PRs.

PR #56666, opened 2026-01-23
Skip unique index lookup for insert_all! since it doesn't need conflict target

kirs:
`insert_all!` uses `on_duplicate: :raise` which performs a plain INSERT without any `ON CONFLICT` clause. Previously, `find_unique_index_for` was called unconditionally, requiring a unique index even when one wasn't needed.

This caused unnecessary failures for tables with composite primary keys
where `model.primary_key` differs from the schema's primary key columns, such as when a model sets `self.primary_key = :id` but the table has `primary_key: [:shop_id, :id]`.

`insert_all` (`on_duplicate: :skip`) and `upsert_all` (`on_duplicate: :update`) still require the unique index for generating the conflict target clause.

@eileencodes @matthewd

### DIFF
--- a/activerecord/lib/active_record/insert_all.rb
+++ b/activerecord/lib/active_record/insert_all.rb
@@ -38,7 +38,7 @@ module ActiveRecord
       @returning = (connection.supports_insert_returning? ? primary_keys : false) if @returning.nil?
       @returning = false if @returning == []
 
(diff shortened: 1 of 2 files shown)


eileencodes:
Backported to 8-1-stable and 8-0-stable because this is a bug with composite keys.

PR #56670, opened 2026-01-23
[ci skip] Fix guides for db:prepare

schmijos:
We clearly migrate and dump the schema: https://github.com/rails/rails/blob/660fee087d4c92f0c7c31f11bc29b5f40e399caa/activerecord/lib/active_record/tasks/database_tasks.rb#L212-L213

### DIFF
--- a/guides/source/active_record_migrations.md
+++ b/guides/source/active_record_migrations.md
@@ -1290,13 +1290,12 @@ The `bin/rails db:prepare` command is similar to `bin/rails db:setup`, but it
 operates idempotently, so it can safely be called several times, but it will
 only perform the necessary tasks once.
 
-* If the database has not been created yet, the command will run as the
-  `bin/rails db:setup` does.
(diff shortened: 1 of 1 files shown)


guides/source/active_record_migrations.md
```diff
@@ -1296,7 +1296,8 @@ only perform the necessary tasks once.
   load the schema, run any pending migrations, dump the updated schema, and
   finally load the seed data. See the [Seeding Data
   documentation](#migrations-and-seed-data) for more details.
-* If the database and tables exist, the command will do nothing.
+* If the database and tables exist, the command will run pending migrations,
+  dump the updated schema, but not seed.
```
byroot: I think it would be clearer to modify the previous point to just mention that if the tables haven't been created the scehma is loaded. 

Then have this bullet point mention that then migrations are ran and schema dumped.

PR #56671, opened 2026-01-24
Make DatabaseTasks#migration_class public

zzak:
This behavior is not likely to change in the near future, so this change makes it explicit which class is being used to connect for rake tasks.

See: #56479


eileencodes:
I guess this is fine but I wish I understood the why behind not using the rake tasks.

PR #56673, opened 2026-01-24
Add save_transaction_options to customize transaction options for persistence methods

kirs:
Currently, if you need to use a specific transaction isolation level (e.g., `:read_committed`) for save/destroy operations on a model, you must either:

1. Wrap every call in an explicit transaction block
2. Monkeypatch `with_transaction_returning_status` entirely, which is fragile and breaks when the method changes upstream

Rails already provides `ActiveRecord.with_transaction_isolation_level` for setting isolation levels globally within a block, but there's no per-model way to configure the implicit transactions that persistence methods create.

## Solution

This PR extracts the transaction options into an overridable private method:

```ruby
  private
    def save_transaction_options(connection)
        {} # respect existing transaction
      else
        { isolation: :read_committed }
      end
    end
end
```

The hook receives the connection, allowing conditional logic based on whether a transaction is already open.

## Alternative considered

A class-level configuration like `default_save_transaction_isolation` is another option here:

```
  self.default_save_transaction_isolation = :read_committed
end
```

This would be simpler for the common case but less flexible. I'm happy to implement that instead of the team feels that's preferable.

@eileencodes @matthewd

### DIFF
--- a/activerecord/lib/active_record/transactions.rb
+++ b/activerecord/lib/active_record/transactions.rb
@@ -430,7 +430,7 @@ module ActiveRecord
           status = nil
           ensure_finalize = !connection.transaction_open?
 
-          connection.transaction do
+          connection.transaction(**save_transaction_options(connection)) do
             add_to_transaction(ensure_finalize || has_transactional_callbacks?)
             remember_transaction_record_state
 
@@ -457,6 +457,32 @@ module ActiveRecord
         @_last_transaction_return_status = nil
         @_committed_already_called = nil
         @_new_record_before_last_commit = nil
+      end
+
+      # Returns the transaction options to use when +save+, +destroy+, +touch+,
+      # and other persistence methods open a transaction. Override this method
+      # to customize transaction behavior, for example to set a specific
+      # isolation level.
+      #
+      # The +connection+ parameter provides access to the current database
+      # connection, allowing conditional logic based on connection state
+      # (e.g., whether a transaction is already open).
+      #
+      # Returns an empty hash by default, which uses standard transaction defaults.
+      #
+      #   class Account < ApplicationRecord
+      #     private
+      #       def save_transaction_options(connection)
+      #         if connection.transaction_open?
+      #           {}
+      #         else
+      #           { isolation: :read_committed }
+      #         end
+      #       end
+      #   end
+      #
+      def save_transaction_options(connection)
+        {}
       end
 
       # Save the new record state and id of a record so it can be restored later if a transaction fails.


juniper-shopify:
It would indeed be nice to be able to set things like `requires_new: true` on the save transaction, in addition to its isolation level, so I like the approach proposed here better than the alternative discussed.

---

Another option to consider is to introduce an overridable method like this example that is actually used by `with_transaction_returning_status` to open the implicit transaction:
```
def in_implicit_transaction_for_model_operations(connection)
 connection.transaction { yield }
end
```
A developer could override that method to do any kind of option-setting they wanted to on the transaction that would be opened (and they could even just `yield` and skip opening a transaction entirely in some special cases, if they wanted to).

It's a sharper-edged tool than the approach proposed here, but as we're talking about a case where a developer wants to change something about the way a model handles implicit transactions internally, we're in sharp-edged territory for sure.

eileencodes (changes requested):
Can you write some tests for this and add a changelog?

zzak:
I feel like I've seen this pattern before in Rails where we have a private method for child classes to override, and the method appears under "private instance methods" in the docs, but I can't find it now maybe it was Ruby docs.

As this stands though, the method does not appear in the documentation, so it's questionable whether this can be considered public API for applications to consume:
https://kirs-transaction-options.rails-docs-preview.pages.dev/api/classes/ActiveRecord/Transactions/ClassMethods.html

I wasn't sure if that was already considered, just making a note in case :writing_hand:

afurm:
I like the motivation here (avoiding monkeypatching `with_transaction_returning_status` / wrapping every call), but I’m a bit unsure about the API shape.

Right now the hook is “return a hash of options that gets splatted into `connection.transaction(**opts)`”. That’s flexible, but it’s also a little indirect, and it pushes a lot of subtlety onto app code (e.g. nested transactions, adapters ignoring/raising on isolation when already in a transaction, etc.).

Have you considered an alternative override point that wraps the implicit transaction itself, e.g. something like:

```ruby
private
  def in_implicit_transaction_for_persistence(connection, &block)
    connection.transaction(&block)
  end
```

(or similar naming), where apps can override and decide *how* to run the block (set isolation, requires_new, or even skip opening a transaction in special cases)?

That feels like a more “Rails-y” extension point: it keeps the core responsible for calling the block, and customizations can be implemented without relying on kwargs semantics.

Also +1 to adding tests/changelog so we can be confident this is a supported hook.

ipvalverde:
I have created this PR as another implementation option as per @juniper-shopify and @afurm suggestions. That would give developers more control on how the logic executed: https://github.com/rails/rails/pull/56732

kirs:
https://github.com/rails/rails/pull/56732 was merged and that's a lot more elegant solution to this.

PR #56674, opened 2026-01-24
refactor: Rename keys_including_timestamps to sorted_keys

Saidbek:
### Motivation / Background

Addresses a TODO comment refactor. The `keys_including_timestamps` method name is misleading because it only conditionally includes timestamp attributes (when `record_timestamps?` is true), not always.

### Detail

This Pull Request renames `keys_including_timestamps` to `sorted_keys` to accurately reflect its behaviour: it always returns sorted keys and conditionally extends them with timestamp attributes when needed.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/insert_all.rb
+++ b/activerecord/lib/active_record/insert_all.rb
@@ -81,7 +81,7 @@ module ActiveRecord
 
         verify_attributes(attributes)
 
-        keys_including_timestamps.map do |key|
+        sorted_keys.map do |key|
           yield key, attributes[key]
         end
       end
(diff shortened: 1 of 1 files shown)


afurm:
Nice small refactor — `sorted_keys` reads much closer to what the method actually does (always sorts, and only conditionally adds timestamps).

One thought: is `keys_including_timestamps` used anywhere outside Rails internals? If there’s any chance of external usage (gems/apps reaching into `InsertAll`), it might be worth keeping a backwards-compatible alias (optionally with a deprecation) for a release or two to avoid a surprise break.

Otherwise the rename looks consistent across call sites.

skipkayhil:
Thank you for the pull request, and I recognize there is a TODO, but this is really a cosmetic change so I'm going to close this. https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#write-your-code

PR #56678, opened 2026-01-25
Disable console query cache by default

camallen:
This Pull Request has been created because to fix the rails console using the query cache by default since https://github.com/rails/rails/pull/56297 was merged. 

Resolves #56473

This Pull Request disables the query cache for the session. This behaviour is opt-out and can be disabled by the the `--query-cache` option, e.g. `bin/rails console -q`. Note the -q has no effect when `--skip-executor` options is used.

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -5,6 +5,12 @@
 
     *David Lowenfels*
 
+*   Disable the Active Record query cache in the console by default when using the executor.
+
+    The query cache is now off by default in the console. Pass `--query-cache` to enable it for the session.
+
+    *Cam Allen*
+
 *   Console `reload!` will reset the console's executor, when present.
 
     *Ben Sheldon*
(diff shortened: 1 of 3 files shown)


railties/test/application/console_test.rb
```diff
@@ -191,6 +191,18 @@ def test_console_with_conditional_executor_can_be_disabled
...
   end
 
+  def test_console_query_cache_disabled_by_default
+    spawn_console
+
+    write_prompt "ActiveRecord::Base.connection_pool.query_cache_enabled", "=> false"
+  end
+
+  def test_console_query_cache_option
+    spawn_console "--query-cache"
```
zzak, line 201: Could you add a test for -q as well?

byroot: I don' think it's really worth it.

railties/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 
     *David Lowenfels*
 
+*   Disable the Active Record query cache in the console by default when using the executor.
+
+    The query cache is now off by default in the console. Pass `--query-cache` to enable it for the session.
```
byroot, line 10: I wasn't even really expecting a flag to switch it, but I guess it doesn't hurt.

PR #56679, opened 2026-01-25
Fix overlapping acronyms order dependency in underscore

Saidbek:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/56675

`underscore` currently behaves inconsistently when overlapping acronyms are defined (e.g. USD and USDC). The result depends on the order in which the acronyms are registered, so the same input may produce different outputs.

This occurs because the generated regex relies on alternation (`|`), which is evaluated left-to-right. If a shorter acronym appears before a longer, overlapping one, it will be matched first even when the longer acronym is the correct match.

### Detail

This change removes the order dependency by sorting acronyms by length (longest first) before constructing the regex. Doing so ensures that longer, overlapping acronyms are always matched before shorter ones, regardless of registration order.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/inflector/inflections.rb
+++ b/activesupport/lib/active_support/inflector/inflections.rb
@@ -265,7 +265,8 @@ module ActiveSupport
 
       private
         def define_acronym_regex_patterns
-          @acronym_regex             = @acronyms.empty? ? /(?=a)b/ : /#{@acronyms.values.join("|")}/
+          sorted_acronyms = @acronyms.empty? ? [] : @acronyms.values.sort_by { |a| -a.length }
+          @acronym_regex             = sorted_acronyms.empty? ? /(?=a)b/ : /#{sorted_acronyms.join("|")}/
           @acronyms_camelize_regex   = /^(?:#{@acronym_regex}(?=\b|[A-Z_])|\w)/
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/inflector/inflections.rb
```diff
@@ -265,7 +265,8 @@ def clear(scope = :all)
 
       private
         def define_acronym_regex_patterns
-          @acronym_regex             = @acronyms.empty? ? /(?=a)b/ : /#{@acronyms.values.join("|")}/
+          sorted_acronyms = @acronyms.empty? ? [] : @acronyms.values.sort_by { |a| -a.length }
```
adbatista, line 268: I know I already merged, but I have a question: 
Is the ternary operator really necessary? 
`@acronyms.empty? ? [] : @acronyms.values.sort_by { |a| -a.length }`
if `@acronyms` is empty, values should return `[]` already so the sort_by returns the an array

bogdan: It can be faster because it produces empty array more efficiently. No one probably had ever measured it though.

byroot: The difference is likely too small to measure. But yes avoiding the conditional would have been slightly nicer, but not really worth changing at this point.

adbatista: Yeah, does not need to change. I just asked in case there was something that I was missing. Thank you folks

PR #56680, opened 2026-01-25
Don't parallelize databases that Rails doesn't manage

seanhogge:
### Motivation / Background

This Pull Request has been created because external databases, such as those with `database_tasks` set to `false`, are being renamed in parallelized testing. Obviously, this breaks any connection to that database, and raises errors when classes that use that database are loaded (even if not actually used in tests).

### Detail

This Pull Request changes the behavior of `activerecord/lib/active_record/test_databases.rb` - it skips the renaming and schema reconstruction if `database_tasks` is `false`

### Additional information

Perhaps a more specific setting like `parallelize` should be added to database configs? I'm unaware of a situation in which one would want Rails not to manage a database, but to have it parallelized in tests somehow, but if that situation does exist, then this approach may be naive.

This also may indicate that `database_tasks?` is doing too much - it's really answering "does this database need to be managed, and is it also not a replica."

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Fix database renaming for external databases during parallelized tests.
+
+    When using external databases (i.e., those not managed by Rails), parallelized tests
+    would attempt to rename these databases. This change skips the renaming behavior for
+    databases with `database_tasks` set to `false`.
+
+    *Sean Hogge*
+
 *   Fix PostgreSQL schema dumping to handle schema-qualified table names in foreign_key references that span different schemas.
 
         # before
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -20,9 +20,14 @@ module ActiveRecord
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
       ActiveRecord::Base.configurations.configs_for(env_name: env_name, include_hidden: true).each do |db_config|
+        # Skip databases with database_tasks: false (e.g., external databases)
+        # NOTE: replicas return false when calling `database_tasks?`, so we need to check the configuration hash directly.
+        next if db_config.configuration_hash[:database_tasks] == false
+
         db_config._database = "#{db_config.database}_#{i}"
 
-        if db_config.database_tasks?
+        # Replicas don't need schema reconstruction - they share the primary's database
+        unless db_config.replica?
           ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
         end
       end
--- a/activerecord/test/cases/test_databases_test.rb
+++ b/activerecord/test/cases/test_databases_test.rb
@@ -131,5 +131,39 @@ class TestDatabasesTest < ActiveRecord::TestCase
       ActiveRecord::Base.establish_connection(:arunit)
       ENV["RAILS_ENV"] = previous_env
     end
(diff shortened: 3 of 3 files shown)


eileencodes:
Why not skip the database creation hook instead? 8.1.2 has an API for this https:eapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize. If you're not on 8.1.2 you can clear the hooks with `ActiveSupport::Testing::Parallelization.before_fork_hooks.clear`

seanhogge:
> Why not skip the database creation hook instead? 8.1.2 has an API for this https:eapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize. If you're not on 8.1.2 you can clear the hooks with `ActiveSupport::Testing::Parallelization.before_fork_hooks.clear`

Since it was trivial to try, I tried it first despite this note on that method:

> Note that your test suite may deadlock if you attempt to use only one database with multiple processes.

Indeed, I did get deadlocks.

I'm not familiar with this process in Rails, so I may not fully understand your suggestion. Do you mean to implement my own database creation hook that parallelizes the primary database but does not parallelize the others that I specify (probably in `test_helper.rb`)?

(hopefully helpful context: I am on Rails 8.1.2, recently from 8.0 which is how I hit this behavior - the change at https://github.com/rails/rails/pull/55769/changes appears to be the reason this Just Worked™ before the upgrade)

eileencodes:
I'm not really following your app setup but yes there will be deadlocks, even for the databases outside Rails so you're going to have to handle that in a custom hook.

> Do you mean to implement my own database creation hook that parallelizes the primary database but does not parallelize the others that I specify (probably in test_helper.rb)?

Yea if you're doing something special you should write the custom hooks in your app. Rails can't and shouldn't handle every possible use case for setting up test databases.

seanhogge:
Ok, thanks for taking a look! I got spoiled by Rails 8.0 handling this, so I'll look into writing the custom hooks.

seanhogge:
For posterity/others with this same issue:

I have a few databases that are external to the app: an old Rails 6 app that this one is replacing, a MySQL database for a public facing site, and a MSSQL app from a vendor. I set them up in `config/database.yml` and added

```yaml
readonly: true
database_tasks: false
```

to all of them. On Rails 8.0.2, running parallelized tests worked, and Rails didn't rename the external, unmanaged databases with a suffix.

After this change: https://github.com/rails/rails/pull/55769/changes Rails now renames _all_ databases in parallelized tests so my databases such as `ifs` and `old_site` were being referenced as `ifs_8` or `old_site_5` during tests. Obviously, those databases don't and can't exist because Rails doesn't manage them. This change was added for Rails 8.1. Before this, the `database_tasks: false` appears to have skipped that suffixing which must have been unintended.

The solution for Rails 8.1, as Eileen pointed out, is to disable this behavior in Rails entirely by using `parallelize(workers: :number_of_processors, parallelize_databases: false)` in `test_helper.rb` and writing a custom hook that will parallelize conditionally instead.

Once I have that puzzled out, I'll add an example solution here in hopes that it helps.

seanhogge:
After attempting to replicate this issue on a new rails project, I'm unable to. So it's almost certainly an embarrassing oversight on something in my project that is indeed special, where I thought I was doing "vanilla" external databases.

🤦🏻 

So it's unlikely anyone else will ever encounter this problem, but I guess I'll leave this here as a reminder to myself to be more thorough before opening a PR against a battle-tested project.

PR #56683, opened 2026-01-27
Pass sql query to query log tags

fatkodima:
We had a need to debug why the sql query is not hitting the database replica in production. We already attach a single backtrace line for each query as an sql tag, but this is not enough for this case. So depending on the sql string we want to add more backtrace lines.
 
This problem can be also solved via `ActiveSupport::Notifications.subscribe("sql.active_record")` and some custom logic, but it is easier with the proposed solution and I think there can be other cases where passing sql to query log tags can be useful.

Usage example
```ruby
config.active_record.query_log_tags = [
  sql_length: ->(context) { context[:sql].length }
]
```

### DIFF
--- a/activerecord/lib/active_record/query_logs.rb
+++ b/activerecord/lib/active_record/query_logs.rb
@@ -47,6 +47,13 @@ module ActiveRecord
   # ActiveSupport::CurrentAttributes can be used to store application values. Tags with +nil+ values are
   # omitted from the query comment.
   #
+  # +context+ includes the following keys:
+  #
+  # * +controller+ - current Action Controller instance, if available
+  # * +job+ - current Active Job instance, if available
+  # * +connection+ - current database connection
+  # * +sql+ - current SQL query
+  #
   # Escaping is performed on the string returned, however untrusted user input should not be used.
   #
   # Example:
@@ -62,6 +69,17 @@ module ActiveRecord
   #         static: "value",
   #       },
   #     ]
+  #
+  # WARNING: SQL query can contain sensitive data and using +:sql+ directly as a query log tag
+  # is unsafe because it will log the query unfiltered.
+  #
+  # Example:
+  #
+  #     # unsafe
+  #     config.active_record.query_log_tags = [:sql]
+  #
+  #     # safe
+  #     config.active_record.query_log_tags = [sql_length: ->(context) { context[:sql].length } ]
   #
   # By default the name of the application, the name and action of the controller, or the name of the job are logged
   # using the {SQLCommenter}[https://open-telemetry.github.io/opentelemetry-sqlcommenter/] format. This can be changed
@@ -141,7 +159,7 @@ module ActiveRecord
       end
 
       def call(sql, connection) # :nodoc:
-        comment = self.comment(connection)
+        comment = self.comment(sql: sql, connection: connection)
 
         if comment.blank?
           sql
@@ -194,16 +212,16 @@ module ActiveRecord
 
         # Returns an SQL comment +String+ containing the query log tags.
         # Sets and returns a cached comment if <tt>cache_query_log_tags</tt> is +true+.
-        def comment(connection)
+        def comment(extra_context)
           if cache_query_log_tags
-            self.cached_comment ||= uncached_comment(connection)
+            self.cached_comment ||= uncached_comment(extra_context)
           else
-            uncached_comment(connection)
+            uncached_comment(extra_context)
           end
         end
 
-        def uncached_comment(connection)
-          content = tag_content(connection)
+        def uncached_comment(extra_context)
+          content = tag_content(extra_context)
 
           if content.present?
             "/*#{escape_sql_comment(content)}*/"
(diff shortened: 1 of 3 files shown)


zzak:
Do we document the other keys in the context somewhere?

fatkodima:
I haven't found.

zzak:
I think here might be good, but we can add it after:
https://github.com/fatkodima/rails/blob/aa059ff30eb78b44e634a2988b24202f0fa046a0/activerecord/lib/active_record/query_logs.rb#L65

activerecord/lib/active_record/query_logs.rb
```diff
@@ -194,16 +194,16 @@ def build_handler(name, handler = nil)
 
         # Returns an SQL comment +String+ containing the query log tags.
         # Sets and returns a cached comment if <tt>cache_query_log_tags</tt> is +true+.
-        def comment(connection)
+        def comment(sql, connection)
```
byroot: Maybe just pass as hash? I feel like there might be more we wanna pass in the future.

fatkodima:
Thanks for reviews! I converted to use a hash and added some docs.

activerecord/lib/active_record/query_logs.rb
```diff
@@ -223,9 +230,10 @@ def escape_sql_comment(content)
           comment
         end
 
-        def tag_content(connection)
+        def tag_content(**options)
           context = ActiveSupport::ExecutionContext.to_h
-          context[:connection] ||= connection
+          context[:sql] = options[:sql]
+          context[:connection] ||= options[:connection]
```
byroot:
```suggestion
          context.merge!(options)
```

byroot: Also I think the signature should just be `tag_content(extra_context)` (no kwargs, it's just useless overhead). These aren't kwargs semantically, just a context hash,

fatkodima: Used `reverse_merge!` instead, because otherwise this test fails https://github.com/rails/rails/blob/90ccb4501a2eecb85ee8978994034941d94ebf31/activerecord/test/cases/query_logs_test.rb#L169

byroot (approved):

p8:
Not sure if this is already done, but should the SQL be filtered for sensitive data like we do for the LogSubscriber?
https://github.com/rails/rails/blob/90ccb4501a2eecb85ee8978994034941d94ebf31/activerecord/lib/active_record/log_subscriber.rb#L124-L126

activerecord/lib/active_record/query_logs.rb
```diff
@@ -47,6 +47,13 @@ module ActiveRecord
...
   # omitted from the query comment.
   #
+  # +context+ includes the following keys:
+  #
+  # * +controller+ - current Action Controller instance, if available
+  # * +job+ - current Active Job instance, if available
+  # * +connection+ - current database connection
+  # * +sql+ - current SQL query
+  #
   # Escaping is performed on the string returned, however untrusted user input should not be used.
```
p8, line 57: Is this still valid if the SQL is passed?

fatkodima:
> Not sure if this is already done, but should the SQL be filtered for sensitive data

I think people will not add the SQL again as tags to the original SQL. I assume it will be used only to print some metadata about it (like query size or number of INs in it etc) or base some logic on it (like if some specific query, print more backtrace lines, if not - less).

zzak:
To @p8's point, it might be worth documenting safe usage, if we filter it by default that will throw off metrics Dima is trying to use it for.

e.g. `config.active_record.query_log_tags = [ :sql ]` will log every query unfiltered, bad
but `config.active_record.query_log_tags = [ sql: ->(ctx) { ... } ]` can be shown how use that safely

This can be a follow up, I also wonder if we raise if tags includes `:sql` but no callable :thinking:

fatkodima:
Added a note about unsafety. Appreciate all the feedback 🙏

rohanshiva:
We're hitting `Regexp::TimeoutError`, which includes this change. The error occurs in `ActiveRecord::QueryLogs.escape_sql_comment` at [query_logs.rb:238](https://github.com/rails/rails/blob/f5255a4089cff69af0ce3b63fda751bdc6dd93cb/activerecord/lib/active_record/query_logs.rb#L238):

```ruby
comment.gsub!(%r{\A\s*/\*\+?\s?|\s?\*/\s*\Z}, "")
```

With this PR, the full SQL string is now passed into the context. Even though our query log tags don't use `:sql` directly, for large bulk operations (e.g., `insert_all` with many rows), the regex in `escape_sql_comment` exceeds the 1-second `Regexp.timeout` (set by `config.load_defaults 8.0`) and raises `Regexp::TimeoutError`.


**Stack trace:**
```
Regexp::TimeoutError: regexp match timeout
  from activerecord-8.1.2/lib/active_record/query_logs.rb:220:in 'String#gsub!'
  from activerecord-8.1.2/lib/active_record/query_logs.rb:220:in 'ActiveRecord::QueryLogs.escape_sql_comment'
  from activerecord-8.1.2/lib/active_record/query_logs.rb:209:in 'ActiveRecord::QueryLogs.uncached_comment'
  from activerecord-8.1.2/lib/active_record/query_logs.rb:201:in 'ActiveRecord::QueryLogs.comment'
  from activerecord-8.1.2/lib/active_record/query_logs.rb:144:in 'ActiveRecord::QueryLogs.call'
```

Is there a recommended workaround? Should the SQL string only be passed into the context when a tag actually references `:sql`, to avoid unnecessary overhead?

PR #56685, opened 2026-01-27
Fix `extend_queries` with `support_unencrypted_data` when combined with `normalizes`

marfoldi:
### Motivation / Background

Fixes #56684

When using `encrypts` with `normalizes` on the same attribute, `extend_queries` with `support_unencrypted_data` fails to include plaintext values in queries. This prevents finding records with unencrypted data during encryption migrations.

### Detail

This Pull Request changes `activerecord/lib/active_record/encryption/extended_deterministic_queries.rb`:

1. Prepend `ExtendedEncryptableType` to `NormalizedValueType` to handle `AdditionalValue` in serialize
2. Use `type.cast(value)` in `additional_values_for` to normalize plaintext before creating `AdditionalValue`

**Root cause:** `NormalizedValueType#serialize` was calling `normalize(value)` on `AdditionalValue` objects, producing garbage strings like `"#<ActiveRecord::Encryption::ExtendedDeterministicQueries::AdditionalValue:0x...>"` that got encrypted.

### Additional information

This fix prepends to an ActiveModel class (`NormalizedValueType`) from ActiveRecord, which crosses module boundaries. There may be a cleaner approach that keeps the fix entirely within ActiveRecord or ActiveModel.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Fix `extend_queries` with `support_unencrypted_data` when combined with `normalizes`.
+
+    When using `encrypts` with `normalizes` on the same attribute, queries now correctly
+    include the plaintext value, allowing records with unencrypted data to be found.
+
+    *Péter Bence Márföldi*
+
 *   Fix PostgreSQL schema dumping to handle schema-qualified table names in foreign_key references that span different schemas.
 
         # before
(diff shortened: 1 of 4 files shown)


marfoldi:
The problem with this approach is that it breaks the `SerializeCastValue` optimization & it breaks non-idempotent normalizers.

Issue #49385, opened 2023-09-26
nested_attributes create with pre-filled id


armellarcier:
### Steps to reproduce

```ruby
order = Order.new
order.order_lines_attributes = [{ id: 'uuid1', name: 'first line', _create: true }]
order.save
```

### Expected behavior

creates an Order with an OrderLine, giving the OrderLine 'uuid1' as an id

### Actual behavior

`raise_nested_attributes_record_not_found!`

### Suggestion

Add an option so that nested_attributes looks like

```
          if attributes["id"].blank? || attributes["_create"]
            unless reject_new_record?(association_name, attributes)
              association.reader.build(attributes.except(*UNASSIGNABLE_KEYS))
            end
...
```

### System configuration
**Rails version**: Rails 7.0.3.1

**Ruby version**: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-musl]

nvasilevski:
I just wanted to add that since technically Rails doesn't expect the `_create` option in the attributes it's not a bug but rather a feature request which may be more suitable for https://discuss.rubyonrails.org/ as the issue tracker reserved for bugs. But I'll let someone else to evaluate

PR #56689, opened 2026-01-28
Enhance assign_nested_attributes_for_collection_association to support composite primary key attributes as means for determining create/update/destroy behavior

aseroff:
### Motivation / Background

This Pull Request has been created because the create/update/delete behavior of `assign_nested_attributes_for_collection_association` is unnecessarily coupled to the presence of a field named `id`. With these changes, an associated has_many record can be found based on the presence of compound primary_key attributes, allowing the parent object to update nested objects using predetermined values rather than relying on knowing a primary key that may not be known at time of execution due to an auto incrementing field. The handling also includes using knowledge of the parent object to include the value of the parent in the child's attribute hash if necessary.

For example, a form that generates fields for unpersisted child objects can create them and then update or destroy them without needing a page reload or javascript submission handling. 

### Detail

This Pull Request changes the nested_attribute module's `assign_nested_attributes_for_collection_association` to no longer require explicit `id` values to determine create/update/destroy behavior, or the value of parent object's id if present in the compound primary key.

### Additional information

The method continues to support explicitly providing the primary key as an `id` attribute, but provides the means to infer it using a primary_key lookup of the associated class and the parent object's id.

### Checklist


aseroff:
I think the added tests are the most convincing argument for the inclusion of these changes. Happy to resubmit with better commit messages. I hope you'll consider merging! Thanks!

activerecord/lib/active_record/nested_attributes.rb
```diff
@@ -522,16 +537,29 @@ def assign_nested_attributes_for_collection_association(association_name, attrib
...
           attributes = attributes.with_indifferent_access
 
-          if attributes["id"].blank?
+          id = if explicit_id = attributes["id"] || attributes[:id]
+            Array(explicit_id)
+          else
+            Array(primary_key).map do |pk|
+              if association.reflection.foreign_key == pk
+                association.owner.id
+              else
```
afurm, line 553: This line assumes the association uses the owner’s primary key `owner.id` when a PK component matches the foreign key, but `has_many/has_one can` be configured with `primary_key:` pointing to a different owner column (e.g., primary_key: :email, foreign_key: :user_email). In that case, nested attributes that omit the FK will derive the wrong ID component, miss existing records, and potentially raise RecordNotFound or create duplicates. Should use the reflection’s configured owner key instead of id when filling that FK component.

aseroff: Thanks for the review @afurm, good catch. Something like `association.owner.send(reflection.active_record_primary_key)` instead, no?

afurm: Yep — that’s the idea 👍

A couple tweaks though:
- you’ll want to use the existing `reflection` (or `association.reflection`) and call the *owner key* from the reflection rather than assuming `id`.
- `active_record_primary_key` can be composite (array) in some cases, so it’s safest to handle both scalar + array.
- I’d also prefer `public_send` (or `_read_attribute`) over `send` here, since we’re reading a model attribute.

Something along these lines:

```ruby
owner_pk = association.reflection.active_record_primary_key
owner_value = owner_pk.is_a?(Array) ? owner_pk.map { |k| association.owner.public_send(k) } : association.owner.public_send(owner_pk)
```

…and then use `owner_value` when filling the FK component.

aseroff: The tests, including ones utilizing custom foreign_keys combined with the composite primary_keys, are all passing now. However, HABTM threw a curveball with raising on calling active_record_primary_key. Any suggestions on a better replacement for the rescue on UnknownPrimaryKey?

aseroff:
Looks like this feature has been asked for before: https://github.com/rails/rails/issues/49385

PR #56690, opened 2026-01-28
Use Rails logo as placeholder favicon

sislr:
I'd find it way nicer to have a Rails logo as the favicon by default instead of a red circle. I've tried to lookup why the current favicon is in place, but couldn't find any discussion. I assume the current icon should create an incentive to change it when creating a new project.

I've used https://github.com/rails/website/blob/main/assets/images/favicon.png for now. Dimensions are not correct yet (512x512 is required) and a replacement for `icon.svg` is still missing.


dhh:
Appreciate the suggestion, but I prefer the red dot, as it more clearly indicates "replace this". And I wouldn't want someone to accidentally ship Rails branding in production.

PR #56691, opened 2026-01-28
Respect `urls_expire_in` in Active Storage proxy controller's cache headers

### LINKED ISSUE #54637: ActiveStorage ProxyController should reflect `urls_expire_in` setting in the Cache-Control header
### Steps to reproduce

Set the following two configuration settings and then request a file served by ActiveStorage:
```
config.active_storage.urls_expire_in = 10.minutes  
config.active_storage.resolve_model_to_route = :rails_storage_proxy
```

### Expected behavior

`Cache-Control` expiration time should reflect the ActiveStorage URL expiration time.

### Actual behavior

`Cache-Control` is set to "infinity" (using `http_cache_forever`).

This means that despite Rails not accepting requests for expired URLs (they contain an expired signed id), these requests might not even reach Rails - because browser caches, as well as intermediate caches (e.g. Cloudflare), will happily respond with cached assets due to the long cache expiration times.
The `Proxycontroller`s should respect the expiration setting and set the `Cache-Control` header accordingly.

See
https://github.com/rails/rails/blob/master/activestorage/app/controllers/active_storage/representations/proxy_controller.rb#L14
https://github.com/rails/rails/blob/master/activestorage/app/controllers/active_storage/blobs/proxy_controller.rb#L18

(Fwiw, redirect controllers have a separate setting `config.active_storage.urls_expire_in` that can be set to limit the expiration time of service URLs and which is also used to set the cache expiration time of the redirect response: https://github.com/rails/rails/blob/main/activestorage/app/controllers/active_storage/representations/redirect_controller.rb#L11)

### System configuration

**Rails version**: 8.0.1

**Ruby version**: 3.4.2

alexandergitter:
### Motivation / Background

This Pull Request has been created because the Active Storage proxy controllers currently don't limit cache expiration times based on `urls_expire_in`.

Fixes #54637

### Detail

The `urls_expire_in` setting determines the expiration of URLs generated by Active Storage. The default is nil, making those URLs never expire. When `urls_expire_in` is set, the routes generated by Rails expire after the given time.

Before this change, Active Storage proxy controllers would always send Cache-Control headers that allowed private and public caches to cache responses "forever". This could be problematic especially with caching reverse proxies / CDNs, where the cache can respond with the stored file, long after the URL on the Rails side already expired.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Proxy controllers take the `urls_expire_in` setting into account for
+    Cache-Control headers
+
+    Proxy controllers for files and their representations now limit cache
+    max-age to the value set in `urls_expire_in`. Like before this change, the
+    cache validity is still "forever" when `urls_expire_in` is not set (which,
+    by default, it isn't).
+
+    *Alexander Gitter*
+
 *   Fix `MirrorService#mirror` losing blob metadata when copying to mirrors.
 
     Mirrored copies on S3, Azure, and GCS were served as `application/octet-stream`
--- a/activestorage/app/controllers/active_storage/blobs/proxy_controller.rb
+++ b/activestorage/app/controllers/active_storage/blobs/proxy_controller.rb
@@ -29,7 +29,9 @@ class ActiveStorage::Blobs::ProxyController < ActiveStorage::BaseController
     if request.headers["Range"].present?
       send_blob_byte_range_data @blob, request.headers["Range"]
     else
-      http_cache_forever public: true, last_modified: @blob.created_at do
+      expires_in ActiveStorage.urls_expire_in || 100.years, public: true, immutable: true
+
+      if stale?(etag: request.fullpath, last_modified: @blob.created_at, public: true)
         response.headers["Accept-Ranges"] = "bytes"
         response.headers["Content-Length"] = @blob.byte_size.to_s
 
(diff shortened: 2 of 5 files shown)


afurm:
This direction makes sense to me (capping cache TTL to `urls_expire_in` avoids proxies/CDNs serving content long after the signed URL would be considered expired).

A couple questions on the implementation details:

1) The controllers switch from `http_cache_forever public: true` to `expires_in …` + an explicit `stale?` check. Is there a reason we can’t keep using `http_cache_forever` and only adjust the `max-age`? (Just wondering if we’re reimplementing behavior that helper already gets right.)

2) Related: what’s the rationale for `etag: request.fullpath` and `last_modified: Time.new(2011, 1, 1).utc` here? Is this meant to mirror `http_cache_forever`’s semantics, and does it preserve conditional request behavior exactly?

3) Any concerns around adding `immutable: true` when `urls_expire_in` is set? (Seems OK for content-addressed blobs/variants, but wanted to sanity-check.)

alexandergitter:
Thank you for the feedback, @afurm!

`http_cache_forever` unfortunately doesn't provide any options beyond `public:`. In this case I basically just replaced the call to `http_cache_forever` with its implementation, adjusted for the possibility to configure the max-age.

See https://github.com/rails/rails/blob/d7c8ae65b7045490965218a994c300aea8dbb079/actionpack/lib/action_controller/metal/conditional_get.rb#L321-L327

I imagine the `etag` and `immutable: true` were chosen precisely because - as you mentioned - `http_cache_forever` is mostly meant for permanent, content-addressed resources. Which in the case of Active Storage makes a lot of sense since blobs and their keys are intended to be immutable.

PR #56693, opened 2026-01-29
Combine duplicate `if respond_to?(:action_view)`

r7kamura:
### Motivation / Background

This is a simple refactoring.

### Detail

While looking through the nearby code to change load_defaults from 8.0 to 8.1 in my project, I noticed some unnatural code and attempted to improve it.

FYI: It looks like the code I changed was added here.

- https://github.com/rails/rails/pull/55336/changes#diff-9f26e965e7e2a6842bd6ea755924a621038bc9dd8d63cfa6b34092aa32ede02cR371-R374

### DIFF
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -363,9 +363,6 @@ module Rails
 
           if respond_to?(:action_view)
             action_view.render_tracker = :ruby
-          end
-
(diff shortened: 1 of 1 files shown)


yahonda:
Backported to 8-1-stable https://github.com/rails/rails/commit/6b0ebde4a1157867923096752862743fc05d6aba

PR #56695, opened 2026-01-29
Use a single temporary connection pool when checking test schema

ilianah:
### Motivation / Background

This Pull Request has been created because `load_schema_if_pending!` sets up 2 temporary connection pools per database config - 1 when checking if any schemas need updating and another one when checking pending migrations which introduces unnecessary overhead 

### Detail

This Pull Request changes `load_schema_if_pending!` to set up 1 temporary pool to reuse for both checks. 

### Checklist

### DIFF
--- a/activerecord/lib/active_record/migration.rb
+++ b/activerecord/lib/active_record/migration.rb
@@ -707,11 +707,31 @@ module ActiveRecord
       end
 
       def load_schema_if_pending!
-        if any_schema_needs_update?
-          load_schema!
+        any_schema_needs_update = false
+        pending_migrations = []
+
+        db_configs_in_current_env.each do |db_config|
+          ActiveRecord::PendingMigrationConnection.with_temporary_pool(db_config) do |pool|
+            any_schema_needs_update ||= schema_needs_update?(db_config, pool)
+
+            pending = pool.migration_context.open.pending_migrations
+            pending_migrations.concat(pending)
+          end
         end
 
-        check_pending_migrations
+        if any_schema_needs_update
+          load_schema!
+
+          pending_migrations = []
+          db_configs_in_current_env.each do |db_config|
+            ActiveRecord::PendingMigrationConnection.with_temporary_pool(db_config) do |pool|
+              pending = pool.migration_context.open.pending_migrations
+              pending_migrations.concat(pending)
+            end
+          end
+        end
+
+        check_pending_migrations(pending_migrations)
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/migration.rb
```diff
@@ -707,11 +707,24 @@ def check_all_pending!
...
+            end
+          end
         end
 
-        check_pending_migrations
+        if any_schema_needs_update
+          load_schema!
+        else
+          check_pending_migrations(pending_migrations.flatten)
+        end
```
byroot, line 732:
```suggestion
        if any_schema_needs_update
          load_schema!
        end
        check_pending_migrations(pending_migrations.flatten)
```

You have a logic difference here. Before your patch `check_pending_migrations` was always called.

activerecord/lib/active_record/migration.rb
```diff
@@ -707,11 +707,24 @@ def check_all_pending!
...
+              pending_migrations << pending
+            end
+          end
         end
 
-        check_pending_migrations
+        if any_schema_needs_update
+          load_schema!
+        else
+          check_pending_migrations(pending_migrations.flatten)
```
byroot: Another issue that is causing the tests to fail is that if we called `load_schema!` the list of pending migrations is no longer valid.

activerecord/lib/active_record/migration.rb
ilianah: Hmm, I was thinking that because `load_schema!` already runs pending migrations we don't need to check them after.

ilianah: Nevermind, `load_schema!` runs `db:test:prepare` which only loads the schema

afurm:
Nice cleanup — reusing the pool here should help multi-DB apps avoid extra connection churn during test boot.

A couple small nits:

- In `load_schema_if_pending!`, the `if pending = pool.migration_context.open.pending_migrations` branches will always run (empty arrays are truthy). Might be a little clearer as just `pending_migrations.concat(pool.migration_context.open.pending_migrations)` (or guard with `.any?`).

- In `schema_needs_update?`, the call site `schema_up_to_date?(db_config, nil, nil, pool: pool)` reads a bit “positional-nil heavy”. Not blocking, but if there’s a way to make that intent clearer (even a short comment), it would help future readers.

Otherwise LGTM from my side.

PR #46134, opened 2022-09-26
Facilitate use of any regular ERB in database.yml

eikes:
Commit 37d1429ab1d (#35497) introduced the `DummyERB` to avoid loading the environment when running `rake -T`.

The `DummyCompiler` simply replaced all output from `<%=` with a fixed string and removed everything else. This worked okay when it was used for YAML values. When using `<%=` within a YAML key, it caused an error in the YAML parser, making it impossible to use ERB as you would expect. For example `a database.yml` file containing the following should be possible:

```yaml
  development:
    <% 5.times do |i| %>
    shard_<%= i %>:
      database: db/development_shard_<%= i %>.sqlite3
      adapter: sqlite3
    <% end %>
```

Instead of using a broken ERB compiler we can temporarily use a `Rails.application.config` that does not raise an error when configurations are accessed which have not been set as described in #35468.

This change removes the `DummyCompiler` and uses the standard `ERB::Compiler`. It introduces the `DummyConfig` which delegates all known configurations to the real `Rails::Application::Configuration` instance and returns a dummy string for everything else. This restores the full ERB capabilities without compromising on speed when generating the rake tasks for multiple databases.

Deprecates `config.active_record.suppress_multiple_database_warning`.

PR #56697, opened 2026-01-29
Fix credentials being cached with nil key during `load_database_yaml`

arianf:
### Motivation / Background

Commit 5ba8aa5854 ("Facilitate use of any regular ERB in database.yml") introduced `DummyConfig` to wrap `Rails.application.config` during early `database.yml` parsing. However, it did not wrap `Rails.application.credentials`, causing credentials to be cached with a nil key when accessed in database.yml.

This breaks applications that:
1. Set a custom `key_path` in their environment files (e.g., `config.credentials.key_path = '/path/to/production.key'`)
2. Reference credentials in `database.yml` (e.g., `url: <%= Rails.application.credentials.dig(:database_url) %>`)

When `load_database_yaml` is called during early rake task loading (via `setup_initial_database_yaml` in `databases.rake`), it happens before environment files are loaded, so `config.credentials.key_path` isn't set yet. Credentials get cached with a nil key, causing `credentials.dig(:database_url)` to return nil and database connection failures.

This has been an issue since Rails 7.1.0.

### Detail

This Pull Request wraps `@credentials` with `DummyConfig` during `load_database_yaml`, similar to how `Rails.application.config` is already wrapped. This prevents credentials from being cached during early database.yml parsing.

When the environment loads later and sets `config.credentials.key_path`, credentials will be properly initialized with the correct key.

**Files changed:**
- `railties/lib/rails/application/configuration.rb` - Wrap `@credentials` with `DummyConfig` during `load_database_yaml`
- `railties/test/application/configuration_test.rb` - Add test to verify credentials aren't cached during database.yml parsing

### Additional information

The original commit https://github.com/rails/rails/pull/46134  replaced `DummyERB` (which replaced all `<%= %>` with empty strings) with `DummyConfig` + real ERB evaluation. This enabled full ERB capabilities in database.yml but inadvertently caused credentials to be accessed and cached before the environment was fully loaded.

### Checklist

### DIFF
--- a/railties/lib/rails/application/configuration.rb
+++ b/railties/lib/rails/application/configuration.rb
@@ -468,12 +468,16 @@ module Rails
         if path = paths["config/database"].existent.first
           require "rails/application/dummy_config"
           original_rails_config = Rails.application.config
+          # Use instance_variable_get to avoid triggering lazy initialization which would cache credentials
+          original_credentials = Rails.application.instance_variable_get(:@credentials)
 
           begin
             Rails.application.config = DummyConfig.new(original_rails_config)
+            Rails.application.credentials = DummyConfig.new(nil)
             ActiveSupport::ConfigurationFile.parse(Pathname.new(path))
           ensure
             Rails.application.config = original_rails_config
+            Rails.application.credentials = original_credentials
           end
         else
           {}
--- a/railties/test/application/configuration_test.rb
+++ b/railties/test/application/configuration_test.rb
@@ -2431,6 +2431,25 @@ module ApplicationTests
       end
     end
 
+    test "load_database_yaml does not cache credentials with nil key" do
+      app_file "config/database.yml", <<-YAML
+        development:
+          adapter: sqlite3
+          database: db/development.sqlite3
+        production:
+          adapter: postgresql
+          url: <%= Rails.application.credentials.dig(:database_url) %>
+      YAML
+
+      app "production"
+
+      Rails.application.with(credentials: nil) do
+        Rails.application.config.load_database_yaml
+        assert_nil Rails.application.instance_variable_get(:@credentials),
+          "Credentials should not be cached during load_database_yaml"
+      end
+    end
+
     test "raises with proper error message if no database configuration found" do
       FileUtils.rm("#{app_path}/config/database.yml")
       err = assert_raises RuntimeError do
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Fix credentials being cached with nil key during `load_database_yaml`.
+
+    When `database.yml` references `Rails.application.credentials`, the credentials
+    were incorrectly cached during early rake task loading before environment files
+    set `config.credentials.key_path`. This caused credential lookups to return nil,
+    breaking database connections for apps using custom credential key paths.
+
+    This has been an issue since Rails 7.1.0 when `DummyConfig` was introduced.
+
+    *Arian Faurtosh*
+
 *   Enable Ruby `frozen_string_literal` by default.
 
     New Rails apps now include a `config/bootsnap.rb` file that enables frozen string


railties/lib/rails/application/configuration.rb
```diff
@@ -465,12 +465,15 @@ def load_database_yaml # :nodoc:
         if path = paths["config/database"].existent.first
           require "rails/application/dummy_config"
           original_rails_config = Rails.application.config
+          original_credentials = Rails.application.instance_variable_get(:@credentials)
```
zzak, line 472: Why not use `original = Rails.application.credentials`?

railties/lib/rails/application/configuration.rb
```diff
@@ -465,12 +465,15 @@ def load_database_yaml # :nodoc:
...
 
           begin
             Rails.application.config = DummyConfig.new(original_rails_config)
+            Rails.application.instance_variable_set(:@credentials, DummyConfig.new(nil))
```
zzak: Here too: `Rails.application.credentials = DummyConfig.new(nil)`

railties/lib/rails/application/configuration.rb
```diff
@@ -465,12 +465,15 @@ def load_database_yaml # :nodoc:
...
             ActiveSupport::ConfigurationFile.parse(Pathname.new(path))
           ensure
             Rails.application.config = original_rails_config
+            Rails.application.instance_variable_set(:@credentials, original_credentials)
```
zzak:
```suggestion
            Rails.application.credentials = original_credentials
```

railties/test/application/configuration_test.rb
```diff
@@ -2424,6 +2424,32 @@ def index
...
+
+      # Verify credentials are not cached
+      assert_nil Rails.application.instance_variable_get(:@credentials)
+
+      # This should NOT cache credentials - they should remain nil
+      Rails.application.config.load_database_yaml
+
+      # After load_database_yaml, credentials should still be nil (not cached during parsing)
+      assert_nil Rails.application.instance_variable_get(:@credentials),
+        "Credentials should not be cached during load_database_yaml"
```
zzak:
```suggestion
      Rails.application.with(credentials: nil) do
        Rails.application.config.load_database_yaml
        assert_nil Rails.application.credentials,
        "Credentials should not be cached during load_database_yaml"
      end
```

railties/lib/rails/application/configuration.rb
arianf: @zzak thanks for the feedback, this is actually the only place that shouldn't, because it could cause them to be lazily initialized (cached with the wrong key_path)

```ruby
  def credentials
    @credentials ||= encrypted(config.credentials.content_path, key_path: config.credentials.key_path)
  end
```

I'll make the other changes you suggested tho

zzak: In that case isn't the value just nil anyways?

arianf: Looks like it would be an instance of `ActiveSupport::EncryptedConfiguration` instead of `nil`

https://github.com/rails/rails/blob/e50f232402239391c9680d861a6ba9aa543503f8/railties/lib/rails/application.rb#L625-L632

arianf: @zzak Does that clarify things? Let me know if you have any other suggestions

afurm:
Thanks for digging into this — the issue/cause makes sense (database.yml ERB touching `Rails.application.credentials` before env files set `config.credentials.key_path`, leading to memoization with the wrong key).

LGTM overall. One small thought: when swapping/restoring credentials around `load_database_yaml`, it might be slightly safer to avoid the `credentials=` setter and instead set/restore the ivar directly (e.g. `instance_variable_set(:@credentials, ...)`) to minimize any current/future side-effects in the setter. But the current approach looks fine and the regression test is great.

zzak:
@afurm Please stop spamming issues and PRs with LLM generated responses.

arianf:
@zzak would it be helpful for me to create a rails app showing the issue that this is fixing?

arianf:
@zzak Just rebased onto latest main. Mind taking another look when you get a chance? Or let me know if there is anything I should change

arianf:
Looks like this won't be merged.
If anyone else hits this issue, we work around it by placing this in our `application.rb` file:

```ruby
# application.rb
key_paths = {
  'production' => '/secrets/production.key',
  'staging' => '/secrets/staging.key',
  'demo' => '/secrets/demo.key',
}
key_path = key_paths[Rails.env]
config.credentials.key_path = key_path if key_path.present? && File.exist?(key_path)
```

PR #56701, opened 2026-01-30
The fallback WeakThreadKeyMap (used on Ruby 3.3.0–3.3.4) calls select! on every []= to remove dead thread entries, making each write O(n)

nicolasva:
The fallback WeakThreadKeyMap (used on Ruby 3.3.0–3.3.4) calls select! on every []= to remove dead thread entries, making each write O(n).

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -133,14 +133,19 @@ module ActiveRecord
         WeakThreadKeyMap = ObjectSpace::WeakKeyMap
       else
         class WeakThreadKeyMap # :nodoc:
-          # FIXME: On 3.3 we could use ObjectSpace::WeakKeyMap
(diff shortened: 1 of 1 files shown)


byroot:
it is known yes, that's why it's a fallback.

If you are impacted by this, is there a reason you don't just upgrade Ruby?

nicolasva:
> Oui, c'est connu, c'est pourquoi c'est une solution de repli.
> 
> Si vous êtes concerné par ce problème, y a-t-il une raison pour laquelle vous ne mettez pas simplement à jour Ruby ?

What I wrote addresses the issue for Ruby versions prior to this one. I also considered resolving the problem by restricting it to versions prior to Ruby 3.3.0.

I disagree with leaving things as they are, because it's not obvious that people are unaware that it's incompatible with older Ruby versions.

Furthermore, many people may have business constraints, such as using a gem or other component on an older Ruby version and not being able to easily update it.

That's why I implemented this fix.

byroot:
Yeah, I don't agree with worth the trouble. Thanks for the PR though.

PR #56703, opened 2026-01-30
Fix ThroughReflection#association_primary_key with composite keys

felix-d:
This Pull Request updates `ThroughReflection#association_primary_key` to check if `primary_key` is an array and map each element to a string, matching the existing behavior in `BelongsToReflection#association_primary_key`.

### Motivation / Background

`ThroughReflection#association_primary_key` doesn't handle composite primary keys correctly. When a source association has `primary_key: [:col1, :col2]`, the method calls `.to_s` on the array, producing the string `"[:col1, :col2]"` instead of an array `["col1", "col2"]`.

This causes the auto-generated `_ids` reader method to fail with:

```
ActiveRecord::UnknownAttributeReference: Dangerous query method called with non-attribute argument(s): "[:col1, :col2]"
```

@nvasilevski

### DIFF
--- a/activerecord/lib/active_record/reflection.rb
+++ b/activerecord/lib/active_record/reflection.rb
@@ -1102,7 +1102,11 @@ module ActiveRecord
         # Get the "actual" source reflection if the immediate source reflection has a
         # source reflection itself
         if primary_key = actual_source_reflection.options[:primary_key]
-          @association_primary_key ||= -primary_key.to_s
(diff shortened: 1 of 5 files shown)


felix-d:
Sorry just noticed I shouldn't be creating draft PRs until this is ready. Working on it.

nvasilevski (approved):

felix-d:
The failing test looks unrelated.

PR #56705, opened 2026-01-30
Fix AM::Serialization#read_attribute_for_serialization doc

p8:
Alias methods don't get documented by default.

Define an actual method instead to make sure `read_attribute_for_serialization` shows up in the documentation. This also encodes the method signature to only accept a single key as argument, where `send` accepts any arguments.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/serialization.rb
+++ b/activemodel/lib/active_model/serialization.rb
@@ -164,7 +164,9 @@ module ActiveModel
     #       @data[key]
     #     end
     #   end
-    alias :read_attribute_for_serialization :send
+    def read_attribute_for_serialization(key)
+      send(key)
+    end
 
     private
       def attribute_names_for_serialization


zzak:
@p8 Thanks for your PR! I checked the build and it looks good, one question though:

> and add a :nodoc: comment above the alias (otherwise the :method: directive is ignored.

Is this true? Or is it just that bug where you need a comment separating directives otherwise RDoc will clobber them?

p8:
> Is this true? Or is it just that bug where you need a comment separating directives otherwise RDoc will clobber them?

I just thought it was clearer to mark it as `:nodoc:` than adding some random comment.
It tries to fix the following problem:
```ruby
# :method: a
# This will not document the a method,
# as it is seen as the comment for the `a` alias which RDoc won't document.

alias a b
```

```ruby
# :method: a
# This will document the `a` method,
# as it seen as not belonging to the documentation of the following alias.

# This comment gets ignored but is required. It could also be :nodoc:.
alias a b
```

I also have an PR to make this clearer in RDoc:
https://github.com/ruby/rdoc/pull/1588/changes

p8:
A better solution would be to make it a proper method instead.
This would also make the interface more strict (it should only accept a single argument).

p8:
@zzak I've changed the `alias` to an actual method.

afurm:
Thanks! This makes sense.

Switching from `alias :read_attribute_for_serialization :send` to a real method should make RDoc/YARD pick it up more reliably, and I like that it tightens the hook’s contract to a single `key`.

One small nit: consider using `__send__` instead of `send` to avoid edge cases if a class overrides `send`:

```ruby
def read_attribute_for_serialization(key)
  __send__(key)
end
```

Also: the commit message includes `[ci-skip]`. Even though this is small, it is a runtime code change (not only docs), so it might be better to let CI run unless there’s a specific reason to skip.

p8:
> One small nit: consider using `__send__` instead of `send` to avoid edge cases if a class overrides `send`:

I'd rather not change this, as it might break code that depends on it being `send`.
 
> Also: the commit message includes `[ci-skip]`. Even though this is small, it is a runtime code change (not only docs), so it might be better to let CI run unless there’s a specific reason to skip.

Good point! I've removed the `[ci-skip]`.

PR #56709, opened 2026-01-31
Add param_to_key and key_to_param methods to ActiveModel::Conversion

amkisko:
### Motivation / Background

Hi! While working with partitioned tables and composite ids I found these two methods very much useful especially in situations when there are complex database views with custom columns that represent parts of composite id.

Instance method `to_param` is good and nice, but it requires sophisticated additional operations in order to get param for custom records with custom columns which do not follow Rails convention. Also I could not target core method for parsing result of `to_param` which might be essential sometimes.

Both methods became useful for preparing ids before transition to other parts of system (graphql, rpc, job queues, etc).

### Detail

This Pull Request introduces two new class-level ActiveModel::Conversion methods `param_to_key` and `key_to_param`. The idea is to get shorthand methods which respect `param_delimiter` class attribute (it can be customized per model, yes).

The change follows ActiveModel and ActiveRecord terminology, suitable for both.

Another place where it could go is `lib/active_record/integration.rb` (e.g. initial names were `param_to_id` and `id_to_param`), but I think ActiveModel is better as it gives more coverage and flexibility.

### Additional information

I will continue updating this Pull Request to fulfill requirements, but it is already open for discussion if this change is good for Rails core.

### Checklist


afurm:
Thanks for the PR — the idea of having helpers for composite ids makes sense, and `id_to_param` reads nicely.

One concern: with the current implementation, `param_to_id` always returns an Array (or `nil`), even for a simple `"1"` param (`["1"]`). That’s totally fine if the intent is “inverse of `to_key`” (which is array-ish), but the name `param_to_id` suggests it might sometimes return a scalar. Would it make sense to name this `param_to_key` / `key_to_param` instead, to make the contract explicit?

Also, parsing by splitting on `param_delimiter` can be ambiguous if a single id can contain the delimiter (default `"-"`). Might be worth a short note in docs (or at least tests that show the intended behavior).

Finally, could we add tests for:
- nil param/id
- simple `"1"` roundtrip
- composite `["1", "2"]` roundtrip
- custom `param_delimiter`

---

amkisko:
@afurm Hey! Thank you for feedback, I agree with the points you mentioned. `key` terminology also fits quite well and it seems to work better as the change is for ActiveModel. `id` comes from ActiveRecord — initial place where the patch appeared.

`param_to_id` confusion actually came after my brief refactoring attempt to play with code simplicity. First implementation was `def self.param_to_id(param) = param&.include?(param_delimiter) ? param.split(param_delimiter) : param` — so my initial intention was to get either array (composite id) or string (simple).

I appreciate and like the idea to rename these methods as then it covers and fits ActiveModel better and still both methods will be available for ActiveRecord classes too.

`key` in this terminology quite well associates with composite primary key (Rails guidelines article name).

PR #56710, opened 2026-01-31
[ci skip] Improve clarity in Action Mailer guide example

Tresor11:
### Motivation / Background

The Action Mailer guide's scaffold example doesn't make it clear that `login` and `email_address` should be unique fields. Without uniqueness constraints, the example can be confusing; for instance, two users could share the same name, which is why using login (a unique identifier) is the correct approach in the mailer view. This PR updates the scaffold generator command to explicitly declare login and email_address as unique fields, making the example more consistent and self-explanatory.

### Detail

Updates the scaffold generator command in the Action Mailer guide to:

`$ bin/rails generate scaffold user name email_address:uniq login:uniq`

This makes it explicit that login and email_address are unique identifiers, which explains why they are used in the mailer context.

### Checklist

### DIFF
--- a/guides/source/action_mailer_basics.md
+++ b/guides/source/action_mailer_basics.md
@@ -153,7 +153,7 @@ Here is a sample HTML template that can be used for the welcome email:
 <h1>Welcome to example.com, <%= @user.name %></h1>
 <p>
   You have successfully signed up to example.com,
-  your username is: <%= @user.login %>.<br>
+  your username is: <%= @user.name %>.<br>
 </p>
 <p>
   To log in to the site, just follow this link: <%= link_to 'login', login_url %>.
@@ -176,7 +176,7 @@ Welcome to example.com, <%= @user.name %>
 ===============================================
 
 You have successfully signed up to example.com,
-your username is: <%= @user.login %>.
+your username is: <%= @user.name %>.
 
 To log in to the site, just follow this link: <%= @url %>.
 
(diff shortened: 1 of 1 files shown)


p8:
Thanks for the PR @Tresor11 !
In this case `user.name` isn't uniq, there could be two users named Trésor for example.
Both `user.login` and `user.email_address` are supposed to be uniq: the login should identify the user, and we can only send the email to a single user's email_address.

Maybe we need to change the scaffold generator for the user to:
```
$ bin/rails generate scaffold user name email_address:uniq login:uniq
```

afurm:
Thanks for the PR! I agree the example currently mixes `@user.name` and `@user.login` without context.

That said, changing the *username* line to `@user.name` feels semantically off: `name` is typically a display name and not necessarily unique, whereas `login` (or email) is the unique identifier.

Maybe the smallest improvement here is to keep `@user.login`, but adjust the copy so it’s clear what it is, e.g.:

- “your login is: <%= @user.login %>” (HTML + text)

If the guide wants to avoid introducing `login` at all, then I think the label should also change to “your name is …” (instead of “username”).

Tresor11:
Thank you for the review @p8 ,
Considering that the scaffold generator doesn't add any uniqueness constraints by default, adding the `:uniq` option makes sense in this case and also addresses the current issue.
Happy to update this PR's changes and description to include that change instead, or close this one if you'd prefer to handle it on a different one

p8:
@Tresor11 Yes, could you update the PR to add the unique constraint?

Tresor11:
Thanks @p8 I've updated the PR description and the scaffold generator

fatkodima:
The mailers guide is more focused on how to use mailers, not giving too much details to some best practices related to database schemas design and there are even no model definitions there. So let's keep it as is.

PR #56714, opened 2026-02-02
Fix explicit nil ignored for nullable columns with default functions

maxromaniv:
### Motivation / Background

This Pull Request fixes a bug where explicit `nil` values are ignored for nullable columns with default functions when `partial_inserts = false`.

When creating a record with an explicit `nil` value for a nullable column that has a default function (e.g. `DEFAULT CURRENT_TIMESTAMP`), ActiveRecord incorrectly omits the column from the INSERT statement. This causes the database to use its default function instead of respecting the explicit `nil`.

**The bug manifests differently across database adapters:**
- **PostgreSQL**: Returns the default value immediately (not `nil`) because PostgreSQL uses the `RETURNING` clause to fetch the inserted values, which includes the database-generated default
- **MySQL**: Returns `nil` initially but changes to the default value after `.reload`, making the inconsistency less obvious but still present

This inconsistent behavior makes it difficult to write portable code across different database adapters.

### Detail

This Pull Request changes `attribute_names_for_partial_inserts` in `activerecord/lib/active_record/attribute_methods/dirty.rb` to include explicitly-set `nil` values for nullable auto-populated columns in INSERT statements.

**Changes made:**
- Added logic to detect when a auto-populated column is explicitly set to `nil` by the user
- Added helper method `user_provided_nil?` to check if an attribute is nil and came from the user
- Updated the `else` branch of `attribute_names_for_partial_inserts` to prevent filtering out these explicit `nil` values

**Example:**
```ruby
# Schema: t.timestamp :perform_at, default: -> { "CURRENT_TIMESTAMP" }
class Model < ApplicationRecord
  self.partial_inserts = false
end

# Before: used CURRENT_TIMESTAMP 
Model.create!(perform_at: nil).perform_at # => CURRENT_TIMESTAMP
# After: correctly stores NULL
Model.create!(perform_at: nil).perform_at # => nil

# Omitting the attribute still uses the default (unchanged behavior)
Model.create!.perform_at # => <current timestamp>
```

### Additional information

**Tests added:**
- `test "partial insert off with explicit nil for nullable default function attribute"` - Verifies the bug fix
- `test "partial insert off with unset nullable default function attribute"` - Regression test to ensure omitted attributes still use database defaults

**Test coverage:**
- All tests pass on SQLite (in-memory) and MySQL2
- RuboCop checks pass with no offenses

**CHANGELOG:**
- Updated `activerecord/CHANGELOG.md` with fix description

### Checklist

### DIFF
--- a/activerecord/lib/active_record/attribute_methods/dirty.rb
+++ b/activerecord/lib/active_record/attribute_methods/dirty.rb
@@ -251,11 +251,17 @@ module ActiveRecord
             changed_attribute_names_to_save
           else
             attribute_names.reject do |attr_name|
-              if column_for_attribute(attr_name).auto_populated?
-                !attribute_changed?(attr_name)
-              end
+              next false unless column_for_attribute(attr_name).auto_populated?
+              next false if user_provided_nil?(attr_name)
+
+              !attribute_changed?(attr_name)
             end
           end
+        end
+
+        def user_provided_nil?(attr_name)
+          attribute_came_from_user?(attr_name) &&
+            _read_attribute(attr_name).nil?
         end
     end
   end
--- a/activerecord/test/cases/dirty_test.rb
+++ b/activerecord/test/cases/dirty_test.rb
@@ -969,6 +969,30 @@ class DirtyTest < ActiveRecord::TestCase
     end
   end
 
+  test "partial insert off with explicit nil for nullable default function attribute" do
+    if current_adapter?(:Mysql2Adapter, :TrilogyAdapter) && ActiveRecord::Base.lease_connection.database_version < "8.0"
+      skip "MySQL < 8.0 does not preserve explicit NULL for timestamp defaults"
+    end
+
+    with_partial_writes Aircraft, false do
+      aircraft = Aircraft.create!(name: "Boeing7", manufactured_at: nil)
+      aircraft.reload
+
+      assert_equal "Boeing7", aircraft.name
+      assert_nil aircraft.manufactured_at, "Explicitly set nil should be preserved in database"
+    end
+  end
+
+  test "partial insert off with unset nullable default function attribute" do
+    with_partial_writes Aircraft, false do
+      aircraft = Aircraft.create!(name: "Boeing9")
+      aircraft.reload
+
+      assert_equal "Boeing9", aircraft.name
+      assert_not_nil aircraft.manufactured_at, "Unset attribute should use database default"
+    end
+  end
+
   if current_adapter?(:PostgreSQLAdapter) && supports_identity_columns?
     test "partial insert off with changed composite identity primary key attribute" do
       klass = Class.new(ActiveRecord::Base) do
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,24 @@
+*   Fix explicit `nil` values being ignored for nullable columns with default functions when `partial_inserts = false`.
+
+    Previously, explicitly setting a nullable column to `nil` would be ignored and the
+    database default function would be used instead.
+
+    ```ruby
+    # Schema: nullable column with default function
+    t.timestamp :perform_at, default: -> { "CURRENT_TIMESTAMP" }
+
+    class Model < ApplicationRecord
+      self.partial_inserts = false
+    end
+
+    # Before
+    Model.create!(perform_at: nil).perform_at # => CURRENT_TIMESTAMP
+    # After
+    Model.create!(perform_at: nil).perform_at # => nil
+    ```
+
+    *Max Romaniv*
+
 *   Deprecate the `schema_order` option in PostgreSQL database configurations.
 
     Use `schema_search_path` instead. The `schema_order` alias will be


ipvalverde (commented):
Thanks for pushing this fix @maxromaniv !

activerecord/test/cases/defaults_test.rb
```diff
@@ -318,3 +318,40 @@ class Sqlite3DefaultExpressionTest < ActiveRecord::TestCase
     end
   end
 end
+
+class ExplicitNilWithDefaultFunctionTest < ActiveRecord::TestCase
```
adrianna-chang-shopify: I don't think we need the additional test coverage in `defaults_test.rb` -- the `defaults` test suites are mainly around how default values from particular DBs are applied, schema dumping for different adapters, etc. The test coverage here is identical to what was added in `activerecord/test/cases/dirty_test.rb` so IMHO we could just stick with that.

activerecord/lib/active_record/attribute_methods/dirty.rb
```diff
@@ -251,12 +251,19 @@ def attribute_names_for_partial_inserts
...
+              next false unless column_for_attribute(attr_name).auto_populated?
+              next false if user_provided_nil_for_null_column?(attr_name)
+
+              !attribute_changed?(attr_name)
             end
           end
         end
+
+        def user_provided_nil_for_null_column?(attr_name)
+          column_for_attribute(attr_name).null &&
```
adrianna-chang-shopify: Do we actually care if the column is nullable or not, could this just be a `user_provided_nil?` check?

activerecord/test/cases/defaults_test.rb
maxromaniv: I decided to add them to show that the PostgreSQL adapter behaves differently after the fix, as it now returns a `timestamp` instead of `nil`. For the MySQL adapter, the behavior remains unchanged. At the same time, I realized that it may be enough to explain this in the CHANGELOG.
Removed tests in this file.

activerecord/lib/active_record/attribute_methods/dirty.rb
maxromaniv: My goal was to provide behavior that does not break existing logic after migration and avoids running SQL when it is unnecessary.

I investigated this further and tested the behavior for NOT NULL columns with defaults (e.g., Aircraft.wheels_count). Explicitly setting nil on a NOT NULL column already raises a database constraint error, regardless of whether partial_inserts is true or false, so this change does not meaningfully affect that case.

Therefore, I removed this check and renamed the function to `user_provided_nil`
Thank you!

activerecord/test/cases/dirty_test.rb
```diff
@@ -969,6 +969,30 @@ def check_around
     end
   end
 
+  test "partial insert off with explicit nil for nullable default function attribute" do
+    if current_adapter?(:Mysql2Adapter, :TrilogyAdapter) && ActiveRecord::Base.lease_connection.database_version < "8.0"
+      skip "MySQL < 8.0 does not preserve explicit NULL for timestamp defaults"
+    end
```
maxromaniv, line 975: `explicit_defaults_for_timestamp` default changed from OFF to on for MySQL 8.0. If explicit_defaults_for_timestamp is off, db will return TIMESTAMP instead of NULL. 

Details: https://dev.mysql.com/blog-archive/new-defaults-in-mysql-8-0/ and here https://dev.mysql.com/doc/refman/8.4/en/timestamp-initialization.html

adrianna-chang-shopify: We could have a column with a default function that's not a timestamp column though, right? Do we need test coverage for non-timestamp columns for MySQL < 8?

maxromaniv: I reviewed the documentation and tried a few examples locally (using RAND(), UUID(), etc.). After testing, I confirmed from the official documentation that MySQL versions prior to 8.0.13 do not support default functions for columns other than TIMESTAMP or DATETIME types.

Only CURRENT_TIMESTAMP or NOW() can be used as a default function. Any other function (e.g., RAND(), UUID(), or expressions) is not allowed and will be rejected by the database.
This is documented in the official MySQL manual: [Data Type Default Values — MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#data-type-defaults-explicit-old)

> With one exception, the default value specified in a DEFAULT clause must be a literal constant; it cannot be a function or an expression. This means, for example, that you cannot set the default for a date column to be the value of a function such as [NOW()](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_now) or [CURRENT_DATE](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_current-date). The exception is that, for [TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) and [DATETIME](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) columns, you can specify [CURRENT_TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_current-timestamp) as the default.

adrianna-chang-shopify (approved):
Can you rebase please?

maxromaniv:
Thank you! Rebased this branch! cc: @adrianna-chang-shopify

PR #56720, opened 2026-02-02
Mark error page styles as dynamic for Turbo cleanup

### LINKED ISSUE #56719: Development error page styles persist after back navigation with Turbo
## Description

When using Turbo Drive, navigating to a development error page (404, 500, etc.) and then pressing the browser back button causes the error page's inline styles to persist on the application page. This results in visual corruption (red backgrounds, wrong fonts, etc.).

## Steps to reproduce

1. Create a Rails app with Turbo Drive enabled (default in Rails 7+)
2. Visit any application page
3. Click a link that triggers a routing error (e.g., a non-existent path)
4. The development error page displays correctly
5. Press the browser back button
6. **Bug**: The application page now has the error page's red styling (`background: #C00`, helvetica font, etc.)

## Expected behavior

After pressing back, the application page should be restored cleanly without any styles from the error page.

## Root cause

Turbo's `ErrorRenderer` replaces the entire `<head>` when rendering error pages. When navigating back, `PageRenderer` merges heads but only removes stylesheets marked with `data-turbo-track="dynamic"`. Since the error page's `<style>` tag lacks this attribute, it's never cleaned up.

## Proposed fix

Add `data-turbo-track="dynamic"` to the `<style>` tag in `rescues/layout.erb`:

```diff
--- a/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
+++ b/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
@@ -5,7 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="turbo-visit-control" content="reload">
   <title>Action Controller: Exception caught</title>
-  <style>
+  <style data-turbo-track="dynamic">
     body {
       background-color: #FAFAFA;
```

This allows Turbo's existing `removeUnusedDynamicStylesheetElements()` to clean up the error page styles when navigating away.

## Why this works

Turbo already has logic to remove "dynamic" stylesheets during head merges:

```javascript
// From @hotwired/turbo
get unusedDynamicStylesheetElements() {
  return this.oldHeadStylesheetElements.filter((element) => {
    return element.getAttribute("data-turbo-track") === "dynamic"
  })
}
```

The error page styles are temporary by nature - they should not persist across navigations. Marking them as `dynamic` is semantically correct and leverages existing Turbo functionality.

## Note on `turbo-visit-control: reload`

The template already includes `<meta name="turbo-visit-control" content="reload">`, but this is ignored for error responses because Turbo's `ErrorRenderer` doesn't check it (only `PageRenderer` does). Adding `data-turbo-track="dynamic"` solves the problem from the other direction - cleaning up on back navigation rather than preventing rendering on arrival.

## Environment

- Rails version: 8.1
- Turbo version: 8.x
- Browsers: All (Chrome, Firefox, Safari)

## Workaround

Until fixed, applications can override the template by:

1. Adding an initializer:
```ruby
# config/initializers/turbo_error_pages.rb
unless Rails.env.production?
  Rails.application.config.after_initialize do
    ActionDispatch::DebugView::RESCUES_TEMPLATE_PATHS.unshift(
      Rails.root.join("app/views").to_s
    )
  end
end
```

2. Copying `rescues/layout.erb` to `app/views/rescues/layout.erb` with the fix applied.

Thrizian:
## Summary

When using Turbo Drive, navigating to a development error page and pressing the browser back button causes the error page's inline styles to persist on the application page, resulting in visual corruption (red backgrounds, wrong fonts, etc.).

## Root cause

Turbo's `ErrorRenderer` replaces the entire `<head>` when rendering error pages. When navigating back, `PageRenderer` merges heads but only removes stylesheets marked with `data-turbo-track="dynamic"`. Since the error page's `<style>` tag lacks this attribute, it's never cleaned up.

## Fix

Add `data-turbo-track="dynamic"` to the `<style>` tag in `rescues/layout.erb`. This allows Turbo's existing `removeUnusedDynamicStylesheetElements()` to clean up the error page styles automatically when navigating away.

Fixes #56719

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
+++ b/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb
@@ -5,7 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="turbo-visit-control" content="reload">
   <title>Action Controller: Exception caught</title>
-  <style>
+  <style data-turbo-track="dynamic">
     body {
       background-color: #FAFAFA;
       color: #333;


Thrizian:
After further investigation, this only reproduces when `data-turbo-track: "reload"` is missing from asset tags. Since Rails generates these by default, this was a
  configuration issue on my end. Closing.

PR #56721, opened 2026-02-03
Fix MemCacheStore Dalli protocol deprecation warnings

Saidbek:
### Motivation / Background

Resolves https://github.com/rails/rails/issues/56704

Fixes deprecation warnings in `MemCacheStoreTest` when running with Dalli 4.1.0+ by making the meta protocol the default for Rails 8.2.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/railtie.rb
+++ b/activesupport/lib/active_support/railtie.rb
@@ -176,5 +176,13 @@ module ActiveSupport
           app.config.active_support.use_message_serializer_for_metadata
       end
     end
+
+    initializer "active_support.set_default_memcache_options", before: :initialize_cache do |app|
+      ActiveSupport.on_load(:active_support_cache_mem_cache_store) do
+        if default_options = app.config.active_support.default_memcache_options
+          ActiveSupport::Cache::MemCacheStore.default_memcache_options = default_options
+        end
+      end
+    end
   end
 end
--- a/activesupport/lib/active_support/cache/mem_cache_store.rb
+++ b/activesupport/lib/active_support/cache/mem_cache_store.rb
@@ -33,6 +33,10 @@ module ActiveSupport
       # not be allowed to get down to the Dalli client
       OVERRIDDEN_OPTIONS = UNIVERSAL_OPTIONS
 
+      # Default options for MemCacheStore that are merged with user-provided options.
+      # These defaults are applied when creating a new MemCacheStore instance.
+      cattr_accessor :default_memcache_options, default: {}
+
       # Advertise cache versioning support.
       def self.supports_cache_versioning?
         true
@@ -85,7 +89,7 @@ module ActiveSupport
           raise ArgumentError, "First argument must be an empty array, address, or array of addresses."
         end
 
-        @mem_cache_options = options.dup
+        @mem_cache_options = self.class.default_memcache_options.merge(options.dup)
         # The value "compress: false" prevents duplicate compression within Dalli.
         @mem_cache_options[:compress] = false
         (OVERRIDDEN_OPTIONS - %i(compress)).each { |name| @mem_cache_options.delete(name) }
@@ -288,3 +292,5 @@ module ActiveSupport
     end
   end
 end
+
+ActiveSupport.run_load_hooks(:active_support_cache_mem_cache_store, ActiveSupport::Cache::MemCacheStore)
--- a/activesupport/test/cache/cache_store_setting_test.rb
+++ b/activesupport/test/cache/cache_store_setting_test.rb
@@ -5,6 +5,14 @@ require "active_support/cache"
 require "dalli"
 
 class CacheStoreSettingTest < ActiveSupport::TestCase
+  def setup
+    ActiveSupport::Cache::MemCacheStore.default_memcache_options = { protocol: :meta }
+  end
+
+  def teardown
+    ActiveSupport::Cache::MemCacheStore.default_memcache_options = {}
+  end
+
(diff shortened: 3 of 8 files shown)


activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+*   Support meta protocol option in `ActiveSupport::Cache::MemCacheStore`.
```
skipkayhil: (feedback for future PRs) Is anything user facing being changed in this PR? If not, why a CHANGELOG?

skipkayhil (commented):
We should go the framework default route and make `protocol: :meta` the default for 8.2. 

Since the old protocol is deprecated, that means we should also just enable that new default in the test suite as well so that we don't have to specify `protocol: ` everywhere.

Finally, by going the framework default route we can document the change in `configuring.md`. I'm not sure its worth discussing deprecated configuration in the Caching With Rails guide.

byroot:
> We should go the framework default route

Yep, this also map with what @zzak is working on (disable double serialization). We should have a `default_memcache_options = { ... }` config that we set in `load_defaults`.

activesupport/lib/active_support/railtie.rb
```diff
@@ -176,5 +176,15 @@ class Railtie < Rails::Railtie # :nodoc:
           app.config.active_support.use_message_serializer_for_metadata
       end
     end
+
+    initializer "active_support.set_default_memcache_options" do |app|
+      config.after_initialize do
```
zzak: I think the problem is that cache is initialized well before this, which basically `MemCacheStore.new` with whatever used `config.cache_store = :mem_cache_store` or the default.

* https://github.com/rails/rails/blob/6e7e4a16ea3558d582a0bda1145c2484c94f978d/railties/lib/rails/application/bootstrap.rb#L88
* https://edgeguides.rubyonrails.org/configuring.html#config-cache-store

But something like this is what I started using on my branch:

```ruby
initializer "active_support.dalli_string_fastpath", before: :initialize_cache do |app|
  ActiveSupport.on_load(:active_support_cache_mem_cache_store) do
    AS::Cache::MemCacheStore.whatever = app.config.active_support.whatever
  end
end
```

Saidbek: Thanks for pointing out the issue. I used `before: :initialize_cache` to register the hook before cache initialization, so `Rails.cache` gets the defaults when it's created.
I also added `app.config.after_initialize` to update the defaults after user initializers run. Without it, user overrides in `config/initializers` aren't picked up because they run after `:initialize_cache`. 

```ruby
initializer "active_support.dalli_string_fastpath", before: :initialize_cache do |app|
  ActiveSupport.on_load(:active_support_cache_mem_cache_store) do
    AS::Cache::MemCacheStore.whatever = app.config.active_support.whatever

    app.config.after_initialize do
      AS::Cache::MemCacheStore.whatever = app.config.active_support.whatever
    end
  end
end
```

Please let me know if that makes sense or if there's a better way.

zzak: The cache is already initialized, so changing the constants will have no impact on the memcache client, so I'm not sure the point of the after_initialize. There are other parts of Rails which you cannot change in app initializers that way, basically anything in that bootstrap is an example.

Saidbek: You're right that once the cache is initialized, changing the defaults later won't affect it.
I added `after_initialize` because the test below sets the config in a config/initializers file and expects the class attribute to reflect that value. Without it, the test fails because the hook executes during `:initialize_cache`, so it sets the class attribute to the framework default `{ protocol: :meta }`. When the user initializer runs later and sets the config to `{ protocol: :binary }`, the class attribute isn't updated, causing the test to fail.

https://github.com/Saidbek/rails/blob/0d3b5877ebed19cb6fe24c7a42770b5990867a98/railties/test/application/configuration_test.rb#L3341-L3350

zzak: We shouldn't need an initializer purely for our tests, which as no impact on user code.

byroot: I wonder if there is a good reason for the cache being initialized so early.

Saidbek: @zzak gotcha. I removed it and updated the test to use add_to_config instead, so the config is set before the init.

zzak: @byroot AFAIK it's always been that way:
https://github.com/rails/rails/blame/main/railties/lib/rails/application/bootstrap.rb#L87

> Initialize cache early in the stack so railties can make use of it.

skipkayhil:
This is a bit more complicated now because Dalli 5 was released, and it warns if you _do_ pass `protocol: :meta`. Now I'm thinking there's not a great way we can handle this across versions (especially with `load_defaults`, since Rails 8.2 isn't releasing anytime soon), and we should just update the test suite to fix `emulating_unavailability`.

For users, I think they'll just need to pass `protocol: :meta` to their cache store configuration when upgrading to 4.2, and then remove it when upgrading to 5.0 🤷 

wdyt @byroot

yahonda:
In the meantime https://github.com/rails/rails/pull/56751 was merged to pin the dalli version to 4.

byroot:
> wdyt @byroot

Urk, this is tiring. I don't get why upstream decided to warn when passing `protocol: :meta` on 5.x, this makes it a PITA for the migration.

I'm afraid this is indeed out of our hands, and to fix warnings in Rails itself we should just upgrade to Dalli 5.

PR #56722, opened 2026-02-03
[ci-skip] Document protocol option for MemCacheStore

Saidbek:
### Motivation / Background

Documents the `protocol: :meta` option for `MemCacheStore` in the caching guide.


skipkayhil:
I'm not sure this is worth mentioning here, see my comment: https://github.com/rails/rails/pull/56721#pullrequestreview-3743229626

PR #56723, opened 2026-02-03
Avoid Dalli alive! deprecation in tests

afurm:
### Motivation / Background

This Pull Request has been created because Dalli 4.0.3+ deprecates `Dalli::Client#alive!`, and the MemCacheStore tests can trigger that warning. This change avoids the deprecation noise in the test harness without altering runtime behavior.

### Detail

This Pull Request changes the MemCacheStore tests to use a `SafeDalliClient` that forwards `alive!` to `alive?` when available, and scopes the client swap inside the unavailability helper to keep the behavior test-only.

### Additional information

N/A

### Checklist


p8:
Thanks for the PR @afurm.
It looks this is being handled and discussed in https://github.com/rails/rails/pull/56721

PR #56728, opened 2026-02-04
Add config.active_job.immutable_arguments so perform-argument mutations no longer leak into retries

### LINKED ISSUE #55775: ActiveJob retries do not preserve the original job arguments across retries
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 8"
end

require "active_job/railtie"
require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

class BuggyJob < ActiveJob::Base
  class FooError < StandardError; end

  retry_on FooError, attempts: 3

  def perform(messages)
    messages.each do |message|
      some_data = message.delete(:id)
      message[:bar_id] = some_data
      # call(message)
    end

    raise FooError, "Something went wrong!"
  end
end

class BuggyJobTest < ActiveJob::TestCase
  def test_retry_on_with_mutable_arguments
    messages = [
      { id: 1, foo: "bar" },
      { id: 2, foo: "bar" },
    ]

    perform_enqueued_jobs(only: BuggyJob) do
      assert_raises(BuggyJob::FooError) do
        BuggyJob.perform_later(messages)
      end
    end

    assert_equal 3, performed_jobs.size

    all_job_arguments = performed_jobs.map { |job| deserialize_args_for_assertion(job).fetch(:args).first }

    assert_equal messages, all_job_arguments.uniq
  end
end
```

### Expected behavior

ActiveJob should preserve the original arguments for every retry.

### Actual behavior

For methods that can mutate the parameters, these modifications are carried over to subsequent job executions.

```
Finished in 0.070236s, 14.2377 runs/s, 56.9509 assertions/s.

  1) Failure:
BuggyJobTest#test_retry_on_with_mutable_arguments [example.rb:59]:
--- expected
+++ actual
@@ -1 +1 @@
-[{id: 1, foo: "bar"}, {id: 2, foo: "bar"}]
+[[{id: 1, foo: "bar"}, {id: 2, foo: "bar"}], [{foo: "bar", bar_id: 1}, {foo: "bar", bar_id: 2}], [{foo: "bar", bar_id: nil}, {foo: "bar", bar_id: nil}]]
```

### System configuration

**Rails version**: 8.0.3

**Ruby version**: Ruby 3.4.5

Saidbek:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/55775

When a job mutates its `perform` arguments and then retries, those mutations were reused on the next execution. Retries could then run against modified data.

### Solution

Adds a single global `config.active_job.immutable_arguments`. When enabled, Active Job deep-dups arguments before `perform` so mutations of the `perform` parameters do not leak into retries. Mutating `job.arguments` still updates the stored arguments used on retry.

New applications (`config.load_defaults 8.2`) enable this by default. Existing applications keep the previous behavior until they opt in via `config/initializers/new_framework_defaults_8_2.rb`.

### Checklist

### DIFF
--- a/activejob/lib/active_job/execution.rb
+++ b/activejob/lib/active_job/execution.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/rescuable"
+require "active_support/core_ext/object/deep_dup"
 
 module ActiveJob
   # = Active Job \Execution
@@ -12,6 +13,10 @@ module ActiveJob
   module Execution
     extend ActiveSupport::Concern
     include ActiveSupport::Rescuable
+
+    included do
+      class_attribute :protect_retry_arguments, instance_accessor: false, default: false
+    end
 
     # Includes methods for executing and performing jobs instantly.
     module ClassMethods
@@ -65,7 +70,12 @@ module ActiveJob
       def _perform_job
         ActiveSupport::ExecutionContext[:job] = self
         run_callbacks :perform do
-          perform(*arguments)
+          if self.class.protect_retry_arguments
+            protected_arguments = arguments.map(&:deep_dup)
+            perform(*protected_arguments)
+          else
+            perform(*arguments)
+          end
         end
       end
   end
--- a/activejob/lib/active_job/execution.rb
+++ b/activejob/lib/active_job/execution.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/rescuable"
+require "active_support/core_ext/object/deep_dup"
 
 module ActiveJob
   # = Active Job \Execution
@@ -65,7 +66,8 @@ module ActiveJob
       def _perform_job
         ActiveSupport::ExecutionContext[:job] = self
         run_callbacks :perform do
-          perform(*arguments)
+          args = ActiveJob.immutable_arguments ? arguments.map(&:deep_dup) : arguments
+          perform(*args)
         end
       end
   end
--- a/activejob/lib/active_job/railtie.rb
+++ b/activejob/lib/active_job/railtie.rb
@@ -10,6 +10,7 @@ module ActiveJob
     config.active_job = ActiveSupport::OrderedOptions.new
     config.active_job.custom_serializers = []
     config.active_job.log_query_tags_around_perform = true
+    config.active_job.protect_retry_arguments = false
 
     initializer "active_job.deprecator", before: :load_environment_config do |app|
       app.deprecators[:active_job] = ActiveJob.deprecator
@@ -76,6 +77,10 @@ module ActiveJob
           elsif respond_to? k
             send(k, v)
           end
+        end
+
+        if app.config.active_job.key?(:protect_retry_arguments)
+          ActiveJob::Base.protect_retry_arguments = app.config.active_job.protect_retry_arguments
         end
       end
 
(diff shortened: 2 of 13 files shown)


activejob/lib/active_job/execution.rb
```diff
@@ -13,6 +14,10 @@ module Execution
     extend ActiveSupport::Concern
     include ActiveSupport::Rescuable
 
+    included do
+      class_attribute :protect_retry_arguments, instance_accessor: false, default: false
```
JoeSouthan: Maybe `immutable_job_arguments`?

activejob/lib/active_job/railtie.rb
```diff
@@ -10,6 +10,7 @@ class Railtie < Rails::Railtie # :nodoc:
     config.active_job = ActiveSupport::OrderedOptions.new
     config.active_job.custom_serializers = []
     config.active_job.log_query_tags_around_perform = true
+    config.active_job.protect_retry_arguments = false
```
JoeSouthan: Is there scope to make this the default in a future release of Rails? It's a very easy rough edge to catch yourself on as a developer.

bensheldon: You should always propose the strongest PR possible. If you believe immutable job arguments is the correct behavior, I think this PR should go directly at that outcome and make it a single global configuration with a deprecation warning (e.g. and remove the per-job-class configuration)

activejob/test/cases/protect_retry_arguments_test.rb
```diff
@@ -0,0 +1,73 @@
...
+      self.protect_retry_arguments = true
+
+      def perform(data)
+        data[:mutated] = true
+      end
+    end
+
+    data = { original: "value" }
+    job = job_class.new(data)
+    job.send(:_perform_job)
```
bensheldon: Your tests should go through the public API e.g.

```ruby
job_class = Class.new(ActiveJob::Base) do
  self.queue_adapter = :inline
  self.protect_retry_arguments = true
  
  def perform(data)
    data[:foo] = "baz" # immutable, should do nothing
    arguments[0][:executions] += 1 # mutable when accessed through self.arguments
    
    retry_job if arguments[0][:executions] == 1
  end
end

data = { foo: "bar", executions: 0 }
job_class.perform_later(data)

assert_equal "bar", data[:foo]
assert_equal 2, data[:executions]
```

activejob/lib/active_job/execution.rb
Saidbek: Went with `config.active_job.immutable_arguments` since it's already namespaced under Active Job.

activejob/lib/active_job/railtie.rb
Saidbek: Yes, it's now the default for `load_defaults 8.2`.

Saidbek: Done, it's a single global config with no per-job setting, enabled by default for 8.2.

activejob/test/cases/protect_retry_arguments_test.rb
Saidbek: Updated the tests to go through `perform_later` / `retry_job` instead of calling `_perform_job`.

PR #56729, opened 2026-02-04
Better handle cache deserialization errors

byroot:
While rare, it can happen that some cache backend return truncated responses (memcached...).

But generally speaking, the cache should treat any sort of error during deserialization as a cache miss.


HeyNonster:
Did this recently come up? We started hitting cache deserialization errors back in ~December. Not very frequent ones but it led to us adding some similar fallback logic. We thought there was something going on with our infra though since we had never seen those errors before 🤔

byroot:
> Did this recently come up?

Not recently. Found an old monkey patch at my new job. From instrumentation, it was mostly catching some weird bug in Zlib resulting in errors.

PR #56732, opened 2026-02-04
Add extension point to customize transaction for persistence methods

ipvalverde:
Currently, if you need to use a specific transaction isolation level (e.g., `:read_committed`) for save/destroy operations on a model, you must either:

1. Wrap every call in an explicit transaction block
2. Monkeypatch `with_transaction_returning_status` entirely, which is fragile and breaks when the method changes upstream

Rails already provides `ActiveRecord.with_transaction_isolation_level` for setting isolation levels globally within a block, but there's no per-model way to configure the implicit transactions that persistence methods create.

## Solution

This PR extracts the transaction creation into an overridable private method:

```ruby
  private
    def implicit_persistence_transaction(connection, &block)
        yield # avoid opening a transaction if one is already open
      else
        super
      end
    end
end
```

The method receives the connection and the block, allowing conditional logic based on whether a transaction is already open. This gives caller full control over the logic.

## Alternative considered

[PR implementation](https://github.com/rails/rails/pull/56673)

Although this would allow tweaking the transaction create, it would not give as much control as the current implementation. For example, it would not allow skipping transaction creation or adding custom handling for `ActiveRecord::Rollback` exceptions or `throw` that might happen as part of the callbacks.

@eileencodes @matthewd @kirs 

## Checklist

### DIFF
--- a/activerecord/lib/active_record/transactions.rb
+++ b/activerecord/lib/active_record/transactions.rb
@@ -430,7 +430,7 @@ module ActiveRecord
           status = nil
           ensure_finalize = !connection.transaction_open?
 
-          connection.transaction do
+          implicit_persistence_transaction(connection) do
             add_to_transaction(ensure_finalize || has_transactional_callbacks?)
             remember_transaction_record_state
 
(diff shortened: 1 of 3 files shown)


activerecord/lib/active_record/transactions.rb
```diff
@@ -541,5 +541,33 @@ def add_to_transaction(ensure_finalize = true)
...
+      #       def in_implicit_transaction_for_persistence(connection, &block)
+      #         if connection.transaction_open?
+      #           yield
+      #         else
+      #           super
+      #         end
+      #       end
+      #   end
+      #
+      def in_implicit_transaction_for_persistence(connection, &block)
```
eileencodes: I don't love this API name, it's kind of a mouthful. How about `implicit_persistence_transaction`?

ipvalverde: Thanks, updated it!

PR #56737, opened 2026-02-04
Fix key generator hash digest class caching issue

Saidbek:
### Motivation / Background

[Fix #56736]

When `Rails.app.key_generator` was accessed in an initializer before `config.active_support.key_generator_hash_digest_class` was applied, it would cache a key generator with the wrong hash digest class.

### Detail

The cache key for `Rails.app.key_generator` now includes both `secret_key_base` and `hash_digest_class` as `[secret_key_base, hash_digest_class]`. This ensures that different hash digest classes get different cached generators, preventing stale generators from being reused when the hash digest class changes.

### Checklist

### DIFF
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -176,11 +176,17 @@ module Rails
     # specified +secret_key_base+. The return value is memoized, so additional
     # calls with the same +secret_key_base+ will return the same key generator
     # instance.
+    #
+    # The cache key includes both the +secret_key_base+ and the current
+    # hash_digest_class to ensure that key generators are properly invalidated
+    # when the hash digest class changes.
     def key_generator(secret_key_base = self.secret_key_base)
       # number of iterations selected based on consultation with the google security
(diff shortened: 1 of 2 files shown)


6temes:
I opened an alternative fix in #56746 that takes a different approach — raising a `RuntimeError` instead of changing the memoization key.

The concern with the cache-key approach is that it doesn't protect code that captures the early reference. For example:

```ruby
# config/initializers/repro.rb
VERIFIER = ActiveSupport::MessageVerifier.new(
  Rails.application.key_generator.generate_key("replies"),
  url_safe: true,
)
```

Here `key_generator` is called once during initialization, before `key_generator_hash_digest_class` is applied. The returned instance uses SHA1, and `VERIFIER` holds keys derived from it for the lifetime of the process. A subsequent call to `key_generator` after initialization would return a correct SHA256 instance, but `VERIFIER` is already poisoned.

Raising early makes the misconfiguration impossible to miss, and the error message guides the developer toward a lazily-initialized method instead of a constant.

PR #56742, opened 2026-02-05
Read from `ENV['REVISION']` if present

cocoahero:
### Motivation / Background

A common pattern on many deployment providers is setting the revision via ENV. This is arguably preferred over writing a REVISION file or bundling the git history. This PR updates the `Rails.app.revision` behavior to check `ENV["REVISION"]` before falling back to the `REVISION` file or `git rev-parse`.

### Checklist


rafaelfranca:
I'd like to avoid complicating this code more. We already support using the setter.

PR #56746, opened 2026-02-06
Raise when key_generator is accessed before after_initialize

### LINKED ISSUE #51750: ActiveSupport::KeyGenerator hash digest class differs during and after application initialization
Since rails 7.0, [default hash digest class](https://guides.rubyonrails.org/configuring.html#config-active-support-key-generator-hash-digest-class) for `ActiveSupport::KeyGenerator` is SHA256. 
This SHA256 hash digest class is set through configuration in an `after_initialize` [as found in ActiveSupport railtie](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/railtie.rb).

The problem is that the source code for `ActiveSupport::KeyGenerator` defines a default SHA1 hash digest class if the configuration hash digest class have not been set up. 
It means that using this `KeyGenerator` during application initialization (with SHA1) differs from using it after initialization (with SHA256).

### Steps to reproduce
In any initializer file, print the hash digest class for `ActiveSupport::KeyGenerator`, let's say in `app/config/initializers/digest_class.rb`

```ruby
p ActiveSupport::KeyGenerator.hash_digest_class
```

Then open a rails console in your local environment with `config.eager_load = true`
=> The initializer will print **SHA1**. 
But if you print the same code in the console, it will print **SHA256**.

### Expected behavior
I would expect the `KeyGenerator` to have the same behaviour during and after initialization. 
Because for instance, if we generate a key in an initializer using `ActiveSupport::KeyGenerator#generate_key`, it will be a key generated with SHA1 hash digest class whereas **we expect it to be generated with a SHA256** hash digest class.

Is there a reason why the hash digest class is set in an `after_initialize` ?

### System configuration
**Rails version**:
7.0.7

**Ruby version**:
3.0.6

6temes:
Fixes #51750
Fixes #56736

### Motivation

`Rails.application.key_generator` silently returns a misconfigured
instance when accessed during eager loading.

`config.load_defaults 7.0`+ sets `key_generator_hash_digest_class =
SHA256`, but this is only applied in an `after_initialize` callback.
Eager loading runs **before** `after_initialize`, so any class constant
that calls `key_generator` during eager load gets a `KeyGenerator`
locked to SHA1. This instance is then memoized, poisoning all
subsequent key derivations — including unrelated `MessageVerifier`
instances.

The failure mode is entirely silent: no warning, no error, just wrong
HMAC signatures at runtime (`InvalidSignature: mismatched digest`).

Previously reported in #51750 and #56736.

### Example

```ruby
class Message < ApplicationRecord
  # Evaluated during eager loading, before after_initialize
  VERIFIER = ActiveSupport::MessageVerifier.new(
    Rails.application.key_generator.generate_key("replies"),
    url_safe: true,
  )
end
```

This single constant poisons the memoized `key_generator` for the
entire process, breaking every `MessageVerifier` in the application.

### Change

`key_generator` now raises `RuntimeError` if called before the
`key_generator_hash_digest_class` configuration has been applied and a
custom digest class is configured. The error message explains the issue
and suggests using a lazily-initialized method instead of a constant.

The guard uses a dedicated `@key_generator_hash_digest_class_applied`
flag rather than `initialized?` because Rails internals (ActiveStorage,
ActiveRecord) legitimately call `key_generator` during
`after_initialize` callbacks — when `initialized?` is still `false`.
The flag is set by the `set_key_generator_hash_digest_class` callback
immediately after applying the digest class, so subsequent
`after_initialize` callbacks can safely access `key_generator`.

Apps that do not configure `key_generator_hash_digest_class` (pre-7.0
defaults or no explicit config) are unaffected — the guard only
activates when there is a known misconfiguration risk.

### DIFF
--- a/activesupport/lib/active_support/railtie.rb
+++ b/activesupport/lib/active_support/railtie.rb
@@ -160,6 +160,7 @@ module ActiveSupport
       config.after_initialize do
         if klass = app.config.active_support.key_generator_hash_digest_class
           ActiveSupport::KeyGenerator.hash_digest_class = klass
+          app.send(:key_generator_hash_digest_class_applied!)
         end
       end
     end
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -23,6 +23,19 @@
 
     *Lewis Buckley*
 
+*   Raise when `key_generator` is accessed before `key_generator_hash_digest_class` is applied.
+
+    Calling `Rails.application.key_generator` during eager loading silently
+    returned an instance configured with the wrong digest class (SHA1 instead
+    of SHA256), because `key_generator_hash_digest_class` is only applied in
+    an `after_initialize` callback. The memoized instance then poisoned all
+    subsequent key derivations for the process.
+
+    The key generator now raises if accessed before the digest class has been
+    applied and a custom digest class is configured.
+
+    *Daniel López Prat*
+
 *   Avoid adding `Rack::Sendfile` to the middleware stack if `config.action_dispatch.x_sendfile_header` is `nil`.
 
     The middleware behave as a noop in such case so it's pointless to have it in the stack.
--- a/railties/lib/rails/application.rb
+++ b/railties/lib/rails/application.rb
@@ -121,6 +121,7 @@ module Rails
       @ran_load_hooks    = false
       @revision          = nil
       @revision_initialized = false
+      @key_generator_hash_digest_class_applied = false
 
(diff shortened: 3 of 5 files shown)


6temes:
Worth noting: this guard only protects `key_generator`, not `message_verifier` or `secret_key_base`, which also depend on the key generator transitively. It might be worth adding the same guard to `message_verifier` as a follow-up.

PR #56747, opened 2026-02-06
Narrow rescue in pk_and_sequence_for

nicolasva:
...Narrow rescue clause in pk_and_sequence_for to ActiveRecord::StatementInvalid

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -571,6 +571,8 @@ module ActiveRecord
             SQL
           end
 
+          return nil if result.nil? || result.empty?
+
           pk = result.shift
           if result.last
             [pk, PostgreSQL::Name.new(*result)]
--- a/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb
@@ -578,7 +578,7 @@ module ActiveRecord
             [pk, nil]
           end
         rescue ActiveRecord::StatementInvalid => e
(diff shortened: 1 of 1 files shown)


bhushangahire:
Good change. Worth noting this is actually two improvements bundled:

1. **Explicit `nil`/`empty` short-circuit after the second query** (line 575 of the original would have raised `NoMethodError` on `result.shift` when the second query found no rows — silently swallowed by the bare rescue).
2. **Rescue narrowed** from `StandardError` to `ActiveRecord::StatementInvalid`, with a warn log.

Both are improvements over the silent catch-all. A few things to consider:

**1. Test coverage**

No test added for either path. At minimum: a table that exists but has no PK/sequence should return `nil` cleanly (the new explicit branch). Hard to write without a real Postgres connection, but the PG test suite already does that — a fixture without a sequence would exercise the change.

**2. `PG::Error` is no longer caught**

`ActiveRecord::StatementInvalid` wraps most adapter errors, but not always — e.g., a connection-level `PG::ConnectionBad` can surface directly in some code paths. Previously the bare rescue absorbed those into `nil`. Is propagation the intended behavior? If yes, an inline comment explaining the choice would help future readers.

**3. PR scope vs. title**

The title says "narrow rescue" but the diff also adds a `nil`/`empty` guard *and* a warn log side effect. Either expand the title/description to cover all three, or split into a follow-up PR for the logging change. The logging in particular is a behavior change apps will notice.

**4. Log level**

`warn` on a routine schema introspection failure could get noisy if the calling code is in a retry loop. `debug` might be more appropriate, since the caller already gets a `nil` signal.

Core change is good — these are nits.

rafaelfranca:
@bhushangahire can you stop posting AI generated reviews here? Those aren't helpful.

rafaelfranca (changes requested):
Can you explain why this change, and add tests for it?

fatkodima:
Closing. Please explain the reasoning and we will reopen.

Issue #56740, opened 2026-02-05
ActionController::UnknownHttpMethod returns 500 instead of 405 in Rails 7.2


yayamochi:
### Steps to reproduce
When a request is sent with an invalid HTTP method (e.g., INVALID_METHOD), Rails should ideally catch this as an ActionController::UnknownHttpMethod and return a 405 Method Not Allowed status.

However, in Rails 7.2.3, this leads to a 500 Internal Server Error because the error is raised early in the middleware stack.

```ruby
require 'rails_helper'

describe 'invalid http method should 405' do
  context 'http method is INVALID_METHOD' do
    it do
      process :INVALID_METHOD, '/'
      expect(response).to have_http_status :method_not_allowed
    end
  end
 ## got :internal_server_error

  context 'http method is GET' do
    it do
      process :GET, '/'
      expect(response).to have_http_status :ok
    end
  end
end
```

### Expected behavior
The application should return 405 Method Not Allowed when an invalid HTTP method is provided, maintaining consistency with how Rails handled this in previous versions (e.g., 7.1.x).

### Actual behavior

The application raises ActionController::UnknownHttpMethod, resulting in a 500 Internal Server Error.

Based on my investigation, the error seems to be triggered within the middleware stack before the request reaches the point where it can be gracefully handled as a 405 error. 
This behavior appears to be a regression as it was not present in Rails 7.1.5.2.

This issue seems closely related to:
- https://github.com/rails/rails/issues/38998
- https://github.com/rails/rails/issues/46003

### System configuration
**Rails version**: 
7.2.3

**Ruby version**: 
3.4.6

skipkayhil:
Whoops, I see now that https://github.com/rails/rails/pull/55163 was in `7.2.3`. I backported the fix to 8.1, 8.0, and 7.2 but there may not be another release of 7.2 so you may have to point your Gemfile at the `7-2-stable` branch.

PR #56750, opened 2026-02-07
ActionController::UnknownHttpMethod returns 500 instead of 405 in Rails 7.2 #56740

nicolasva:
fix for issue #56740

**CONTEXT**
The problem lies in DebugExceptions#render_exception (debug_exceptions.rb:68). When a request with an invalid HTTP method arrives:

The router calls request.request_method → ​​throws UnknownHttpMethod
DebugExceptions catches the exception on line 39
In render_exception, request.head? (line 68) calls request.method again on a new request object → throws UnknownHttpMethod a SECOND time
This second exception escapes the rescue exception handler (because it's already caught)
It's sent to ShowExceptions, which should catch it, but depending on the show_exceptions configuration (:none in some test environments), it can go all the way to the server → 500
The same problem potentially exists in DebugExceptions#render_exception on line 68 (request.head?).

**FIX**
The solution is to protect the call to request.head? in DebugExceptions against UnknownHttpMethod, as is already done for InvalidType on lines 63-66:

**TEST**
The fix works correctly. With your modification in debug_exceptions.rb:68-72, request.head no longer throws an uncaught exception when the HTTP method is invalid. The middleware correctly returns 405 Method Not Allowed instead of 500 Internal Server Error in all configuration modes.

_**4 tests done with success :**_
Test | Scenario | Résult
-- | -- | --
1 | INVALID_METHOD + show_exceptions: :all + dev mode | 405
2 | GET valide | 200
3 | INVALID_METHOD + production mode | 405
4 | INVALID_METHOD + show_exceptions: :rescuable | 405

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb
+++ b/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb
@@ -65,7 +65,7 @@ module ActionDispatch
             content_type = Mime[:text]
           end
 
-          if request.head?
+          if request.raw_request_method == "HEAD"
             render(wrapper.status_code, "", content_type)
           elsif api_request?(content_type)
             render_for_api_request(content_type, wrapper)
--- a/actionpack/test/dispatch/debug_exceptions_test.rb
+++ b/actionpack/test/dispatch/debug_exceptions_test.rb
@@ -226,7 +226,7 @@ class DebugExceptionsTest < ActionDispatch::IntegrationTest
     assert_match(/<body>/, body)
     assert_match(/ActionController::MethodNotAllowed/, body)
 
-    get "/unknown_http_method", headers: { "action_dispatch.show_exceptions" => :all }
+    process :unknown, "/unknown_http_method", headers: { "action_dispatch.show_exceptions" => :all }
     assert_response 405
     assert_match(/<body>/, body)
     assert_match(/ActionController::UnknownHttpMethod/, body)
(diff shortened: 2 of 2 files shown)


actionpack/test/dispatch/debug_exceptions_test.rb
```diff
@@ -128,6 +128,9 @@ def call(env)
         rescue Exception
           raise ActionView::Template::Error.new(template)
         end
+      when "/valid_request"
+        req.request_method # raises UnknownHttpMethod for invalid methods
```
skipkayhil: Do we need a new method here vs. reusing something like `/framework_raises`?

actionpack/lib/action_dispatch/middleware/debug_exceptions.rb
```diff
@@ -65,7 +65,13 @@ def render_exception(request, exception, wrapper)
...
           end
 
-          if request.head?
+          begin
+            head_request = request.head?
+          rescue ActionController::UnknownHttpMethod
+            head_request = false
+          end
+
+          if head_request
```
skipkayhil: I think we should just do `request.raw_request_method == "HEAD"`

skipkayhil:
Also please squash your commits into one after addressing comments

actionpack/lib/action_dispatch/middleware/debug_exceptions.rb
nicolasva: it is done @skipkayhil

actionpack/test/dispatch/debug_exceptions_test.rb
nicolasva: it is done @skipkayhil

nicolasva: it is done @skipkayhil

Issue #49648, opened 2023-10-16
Active Support test cases fail using Ubuntu 23.10


yahonda:
### Steps to reproduce
1. Install or upgrade to Ubuntu 23.10
2. Install Ruby 3.2 or whatever Ruby version supported by Rails main branch

```ruby
git clone https://github.com/rails/rails
cd rails/activesupport
bundle install
[...]

PR #56754, opened 2026-02-08
Always include `tzinfo-data` gem

aliismayilov:
### Motivation / Background

Recently, when upgrading from `ruby:3.4-slim` to `ruby:4.0-slim` we got tripped over by the fact that on production TZ data got changed. This was because the base OS changed from Debian Bookworm to Trixie, specifically, Europe/Kiev timezone disappeared (it got renamed to Europe/Kyiv).

### Detail

This Pull Request includes `tzinfo-data` in the standard set of gems regardless of the platform. Locking `tzinfo-data` makes the timezone data more consistent across development, test and production environments.

### Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -154,7 +154,7 @@ platforms :ruby, :windows do
   end
 end
 
-gem "tzinfo-data", platforms: [:windows, :jruby]
+gem "tzinfo-data"
 gem "wdm", ">= 0.1.0", platforms: [:windows]
 
 gem "launchy"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -634,6 +634,8 @@ GEM
       railties (>= 6.0.0)
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
+    tzinfo-data (1.2025.3)
+      tzinfo (>= 1.0.0)
     uber (0.1.0)
     unicode-display_width (3.1.4)
       unicode-emoji (~> 4.0, >= 4.0.4)
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Always include `tzinfo-data` gem in the Gemfile
+
+    *Ali Ismayilov*
+
 *   `Rails.app.revision` now checks `ENV["REVISION"]` before falling back to the `REVISION` file or git.
 
     *Jonathan Baker*
--- a/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
@@ -9,8 +9,7 @@ source "https://rubygems.org"
 # gem "bcrypt", "~> 3.1.7"
 <% end -%>
 
-# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem "tzinfo-data", platforms: %i[ windows jruby ]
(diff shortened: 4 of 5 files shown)


yahonda:
Thanks for bringing this up. I understand the issue of encountering timezone problems after a Docker image upgrade.
However, I'm hesitant to include `tzinfo-data` for all users by default. Let me explain the background.

**Root Cause**

This issue originates from a package restructuring in Ubuntu 23.10+ and Debian 13 (trixie). These distributions have split the `tzdata` package, moving timezone identifiers that Debian/Ubuntu categorizes as "legacy" (like `Europe/Kiev`) into a separate `tzdata-legacy` package.

References:
- [Ubuntu 24.04 Release Notes - tzdata package split](https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#p-99950-tzdata-package-split-29)
- [Debian tzdata 2023c-8](https://launchpad.net/debian/+source/tzdata/2023c-8)

**How Rails CI addressed this**

We've already encountered this issue in Rails' own test suite and resolved it by explicitly installing the `tzdata-legacy` apt package on affected distributions:
- https://github.com/rails/rails/issues/49648
- https://github.com/rails/rails/pull/56066
- https://github.com/rails/buildkite-config/pull/112

**Why I'm hesitant about this approach**

1. Currently, only Debian/Ubuntu-based distributions have implemented this split. It's not a universal issue across all platforms.
2. Not all users need these timezone identifiers that Debian/Ubuntu categorizes as "legacy".
3. Some users may prefer to continue using OS-provided timezone information rather than the `tzinfo-data` gem.
4. Users who explicitly need these timezones (like `Europe/Kiev`) can resolve the issue by either installing the `tzdata-legacy` apt package or adding the `tzinfo-data` gem to their Gemfile.

**Suggested alternative**

That said, I do think we should help users who encounter this issue. How about adding documentation that explains this situation and the available solutions? Alternatively, it might be sufficient to just let the error message lead users to this issue through search, where they can find the context and solutions.

aliismayilov:
Thanks for the detailed context, @yahonda!

I want to clarify — this PR isn't really about the Europe/Kiev => Europe/Kyiv rename specifically. That was just what bit us. The point is that timezone data can silently change under your app when the base OS image gets updated, and there's no good way to notice until something breaks in production. Including `tzinfo-data` unconditionally means timezone behavior is pinned in `Gemfile.lock`, not tied to the OS updates.

A couple of things I wanted to point out:

- The previous default was `gem "tzinfo-data", platforms: %i[ windows jruby ]`, so we're already acknowledging that some platforms need bundled TZ data. But Rails also generates Debian-based Dockerfiles by default, and Debian has now joined that club with the tzdata shenanigans. Rather than documenting another platform-specific workaround, it seems cleaner to just include the gem everywhere.
- One line in the Gemfile template instead of a footnote users have to discover after something breaks. I think this also aligns well with the Rails Doctrine on programmer happiness, timezones changing between deploys because your base image bumped version is exactly the kind of surprise that shouldn't happen by default.
- And for anyone who prefers OS-provided timezone data, opting out is just removing one line from the Gemfile.

yahonda:
I understand that this issue isn't limited to a single timezone, but affects multiple timezones that have been labeled as "legacy." 

However, I'm not convinced that having all users install tzinfo-data is the right approach. I'd like to wait for broader feedback before moving forward.

PR #56758, opened 2026-02-09
Raise an ArgumentError on Date and Time #advance with invalid options

jaydorsey:
### Motivation / Background

This Pull Request has been created to improve the behavior of the `advance` method used on `Date`, `DateTime`, `Time`, and `TimeWithZone` per the suggestion I made in the [ruby forum](https://discuss.rubyonrails.org/t/feature-proposal-activesupport-timewithzone-advance-options-validation/90136)

When dealing with time/date classes:

1. The individual segments are often referred to in the singular (year, month, day)
2. Some methods even use the singular, such as `Date.now.change(year: 2025)`
3. The `advance` method accepts plural options, singular options are a no-op

It’s very easy to make the singular/plural mistake and write something like `Time.zone.now.advance(month: 1)` and not realize it is a no-op because we don’t validate the keys. I've seen this in private repos, but have also found examples in public repos for usage of [year](https://github.com/search?q=%22advance%28year%3A%22+language%3ARuby+&type=code), [month](https://github.com/search?q=%22advance%28month%3A%22+language%3ARuby+&type=code), [day](https://github.com/search?q=%22advance%28day%3A%22+language%3ARuby+&type=code) and [hour](https://github.com/search?q=%22advance%28hour%3A%22+language%3ARuby+&type=code)

### Detail

This Pull Request changes

1. Add the appropriate `assert_valid_keys` check to `Time`, `Date`, `DateTime`, and `TimeWithZone`
3. Add relevant specs for each

### Additional information

1. We could also make it accept both, and simply pluralize any options that come in
    1. This would seem to expand the interface options, making it more confusing (but backwards compatible)
3. We could optionally warn/deprecate on invalid options being passed in instead of raising an ArgumentError
    1. This behavior is an actual bug imho, so I think the ArgumentError is more helpful. The usage of advance with a singular indicates intent to advance by a unit of time, but no advance occurs
5. I provided some examples of where this happened in real public repos in the description. This PR came about because I discovered the issue in two separate private PRs that I worked on.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/date/calculations.rb
+++ b/activesupport/lib/active_support/core_ext/date/calculations.rb
@@ -126,7 +126,7 @@ class Date
   #   # => Mon, 01 Nov 2004
   #
   def advance(options)
-    options.assert_valid_keys(:years, :months, :weeks, :days)
+    options.assert_valid_keys(:years, :months, :weeks, :days, :hours, :minutes, :seconds)
     d = self
 
     d = d >> options[:years] * 12 if options[:years]
--- a/activesupport/lib/active_support/core_ext/date/calculations.rb
+++ b/activesupport/lib/active_support/core_ext/date/calculations.rb
@@ -2,6 +2,7 @@
 
 require "date"
 require "active_support/duration"
+require "active_support/core_ext/hash/keys"
 require "active_support/core_ext/object/acts_like"
 require "active_support/core_ext/date/zones"
 require "active_support/core_ext/time/zones"
@@ -125,6 +126,7 @@ class Date
   #   # => Mon, 01 Nov 2004
   #
   def advance(options)
+    options.assert_valid_keys(:years, :months, :weeks, :days)
     d = self
 
     d = d >> options[:years] * 12 if options[:years]
--- a/activesupport/test/core_ext/date_ext_test.rb
+++ b/activesupport/test/core_ext/date_ext_test.rb
@@ -134,6 +134,20 @@ class DateExtCalculationsTest < ActiveSupport::TestCase
     assert_equal Date.new(2005, 2, 28), Date.new(2004, 2, 29).advance(years: 1) # leap day plus one year
   end
 
+  def test_advance_with_invalid_keys
+    d = Date.new(2005, 2, 28)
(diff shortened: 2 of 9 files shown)


activesupport/test/core_ext/date_ext_test.rb
```diff
@@ -134,6 +134,20 @@ def test_advance
...
+  def test_advance_with_invalid_keys
+    d = Date.new(2005, 2, 28)
+    assert_raise(ArgumentError) { d.advance(year: 1) }
+    assert_raise(ArgumentError) { d.advance(month: 1) }
+    assert_raise(ArgumentError) { d.advance(week: 1) }
+    assert_raise(ArgumentError) { d.advance(day: 1) }
+    assert_raise(ArgumentError) { d.advance(hour: 1) }
+    assert_raise(ArgumentError) { d.advance(minute: 1) }
+    assert_raise(ArgumentError) { d.advance(second: 1) }
+    assert_raise(ArgumentError) { d.advance(hours: 1) }
```
jaydorsey: Just wanted to call these 3 out as intentional because they're no-ops/unused in `Date#advance` (slightly different than the others)

Rails 8.x screenshot:

Image

activesupport/lib/active_support/core_ext/date/calculations.rb
```diff
@@ -125,6 +126,7 @@ def minus_with_duration(other) # :nodoc:
   #   # => Mon, 01 Nov 2004
   #
   def advance(options)
+    options.assert_valid_keys(:years, :months, :weeks, :days, :hours, :minutes, :seconds)
```
jaydorsey, line 129: After observing some failures in the full test suite, I realized that we had some plural options being passed in that raised a lot of errors.

I'm new to minitest so I'm trying to wrap my head around them and understand if there's a better way to solve this than just adding the additional unused keys that also no-op

Image

I'm not sure that a deprecator warning + removing these later (e.g. fixing all specs) is appropriate here but if someone has thoughts I'd appreciate it.

I didn't catch this initially because I wasn't running the full suite locally due to some unrelated errors that I was getting locally (since fixed)

PR #56761, opened 2026-02-10
EventReporter: filter events before building the payload

byroot:
Following https://github.com/rails/rails/pull/55900, we now always pay the efty price of building the event payload and filtering it before we check subscribers to know if they even need it.

I realize this is technically a breaking change, but the filtering logic seem to have been ported over from AS::Notifications but in AS::Notifications we'd save building the payload, here registering a filter doesn't allow to save any substantial overhead.

So in my opinion the feature doesn't make sense as is.

cc @adrianna-chang-shopify @zzak @gmcgibbon any objections?

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Changed `ActiveSupport::EventReporter#subscribe` to only provide the event name during filtering.
+
+    Otherwise the event reporter would need to always build the expensive payload even when there is
+    no active subscriber, which is very wasteful.
+
+    *Jean Boussier*
+
 *   Fix inflections to better handle overlapping acronyms.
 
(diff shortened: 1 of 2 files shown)


adrianna-chang-shopify (commented):
I think this makes sense. When building this originally, I had in mind that certain subscribers might only care about a subset of events, but I didn't think about the fact that certain events might not go to _any_ subscriber, if which case building the payload is unnecessary overhead.

Is it worth still passing `event` as a hash, instead of just the `name`? I suppose that is changing that would be "more" of a breaking change...

byroot:
> Is it worth still passing `event` as a hash, instead of just the `name`? I suppose that is changing that would be "more" of a breaking change...

Yes, I did this so that filters that only look at the name don't break.

PR #56763, opened 2026-02-10
Fix rdoc code formatting

tjschuck:
The simple `+` style doesn't work with more complicated snippets, and `tt` is necessary instead.

Before this is merged, you can see the error here: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-highlight

---

iiokc3ercl 2026-02-10 12-20-02


akhilgkrishnan (approved):
Thanks for the change

PR #56764, opened 2026-02-10
Fix PostgreSQL disable_referential_integrity with duplicate table names

### LINKED ISSUE #30846: Referential integrity not disabled when a table name is shared across schemas (Postgresql)
### Steps to reproduce
1. Create a model with the same table name under two different database schemas.
   1. e.g. `User::Identity` (schema: user, table_name: identities) and `Account::Identity` (schema: account, table_name: identities)
       * These would live `under app/models/user/identity.rb` and `app/models/account/identity.rb`
2. Create fixtures:
   1. `test/fixtures/user/identities.yml`
   2. `test/fixtures/account/identities.yml`

3. run `rails test`

### Expected behavior
Referential integrity via `DISABLE TRIGGER ALL` should be triggered for `user.identities` and `account.identities`

Expected output in logs:
```
ALTER TABLE "user.identities" DISABLE TRIGGER ALL;ALTER TABLE "account.identities" DISABLE TRIGGER ALL;
```

### Actual behavior
Referential integrity is not disabled on _both tables_ because the schema names are not specified in the SQL being executed. E.g. The table "identities" is being referenced instead of "schema_name.identities".

The output will look like:
```
ALTER TABLE "identities" DISABLE TRIGGER ALL;ALTER TABLE "identities" DISABLE TRIGGER ALL;
```
  
### System configuration
**Rails version**:
5.1.4

**Ruby version**:
2.3.4

eglitobias:
### Motivation / Background

Fixes #56757
Fixes #30846

disable_referential_integrity disables/enables triggers using unqualified table names returned by tables. When multiple schemas in search_path contain tables with the same name, PostgreSQL resolves the unqualified name to the first match, leaving triggers enabled in the other schema(s). This can break fixture loading (and similar workflows) when foreign keys exist in those schemas.

Since #57378 this only affects PostgreSQL below 18.4 - the NOT ENFORCED path builds its table names from pg_catalog and is already qualified. The trigger fallback still covers 10.0 to 18.3.

### Detail

This change schema-qualifies table names when disabling/enabling triggers, so all tables in the current search_path are toggled even when duplicate table names exist across schemas. The list is built once and used for both passes; it was queried twice before. A regression test reproduces the failure with duplicate table names and a foreign key in the later schema.

### Additional information

The change is scoped to disable_referential_integrity to avoid altering the behavior of tables/data_source_sql and other call sites. #57378 does the same thing - enforced_foreign_keys uses its own pg_catalog query.

### Checklist

* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. (not needed, minor bug fix)

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
@@ -6,10 +6,11 @@ module ActiveRecord
       module ReferentialIntegrity # :nodoc:
         def disable_referential_integrity # :nodoc:
           original_exception = nil
+          table_names = schema_qualified_tables
 
           begin
             transaction(requires_new: true) do
-              execute(tables.collect { |name| "ALTER TABLE #{quote_table_name(name)} DISABLE TRIGGER ALL" }.join(";"))
+              execute(table_names.collect { |name| "ALTER TABLE #{quote_table_name(name)} DISABLE TRIGGER ALL" }.join(";"))
             end
           rescue ActiveRecord::ActiveRecordError => e
             original_exception = e
@@ -32,7 +33,7 @@ Rails needs superuser privileges to disable referential integrity.
 
           begin
             transaction(requires_new: true) do
-              execute(tables.collect { |name| "ALTER TABLE #{quote_table_name(name)} ENABLE TRIGGER ALL" }.join(";"))
+              execute(table_names.collect { |name| "ALTER TABLE #{quote_table_name(name)} ENABLE TRIGGER ALL" }.join(";"))
             end
           rescue ActiveRecord::ActiveRecordError
           end
@@ -63,6 +64,18 @@ Rails needs superuser privileges to disable referential integrity.
             execute(sql)
           end
         end
+
+        private
+          # Like `data_source_sql`, but returns schema-qualified names so that tables with the same name in multiple
+          # schemas in the search_path are handled correctly.
+          def schema_qualified_tables
+            scope = quoted_scope(nil, type: "BASE TABLE")
+            sql = +"SELECT n.nspname || '.' || c.relname FROM pg_class c "
+            sql << "LEFT JOIN pg_namespace n ON n.oid = c.relnamespace"
+            sql << " WHERE n.nspname = #{scope[:schema]}"
+            sql << " AND c.relkind IN (#{scope[:type] || "'r','p'"})"
+            query_values(sql)
+          end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
```diff
@@ -63,6 +64,18 @@ def check_all_foreign_keys_valid! # :nodoc:
             execute(sql)
           end
         end
+
+        private
+          # Like `data_source_sql`, but returns schema-qualified names so that tables with the same name in multiple
```
eglitobias: Alternative approach: we could add an optional `schema_qualified:` flag to [`data_source_sql`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L1305) (default false) and use it from `disable_referential_integrity` to avoid duplicating the pg_class query.

Something like:

```ruby
def data_source_sql(name = nil, type: nil, schema_qualified: false)
  scope = quoted_scope(name, type: type)
  scope[:type] ||= "'r','v','m','p','f'"

  select = schema_qualified ? "n.nspname || '.' || c.relname" : "c.relname"

  sql = +"SELECT #{select} FROM pg_class c LEFT JOIN pg_namespace n ON n.oid = c.relnamespace"
  sql << " WHERE n.nspname = #{scope[:schema]}"
  sql << " AND c.relname = #{scope[:name]}" if scope[:name]
  sql << " AND c.relkind IN (#{scope[:type]})"
  sql
end

tsvallender: This approach feels a little cleaner to me, but I’m unsure if one or the other is more consistent with the general codebase style.

tsvallender (commented):
This looks eminently sensible and pretty much exactly how I had thought about fixing it 👍

activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
eglitobias: Sorry for the long silence, and thanks for the review.

Rebased on main. #57378 added the PG 18.4+ NOT ENFORCED path in the meantime, which isn't affected. It builds its table names from pg_catalog and is already schema qualified. This now only covers the trigger fallback.

On data_source_sql: the same signature exists in four adapters, and a schema_qualified: flag would only mean something in PostgreSQL. It also changes what the method returns, and that return value is what tables and table_exists? hand out, which have to stay unqualified. #57378 does the same thing - enforced_foreign_keys uses its own pg_catalog query rather than going through data_source_sql.

Also dropped a dead || "'r','p'" fallback.

PR #56765, opened 2026-02-10
Add `transaction_changes` methods for tracking transaction-level attribute changes

lcampanari:
### Motivation / Background

The existing `saved_change_to_*` methods only reflect what changed in the most recent call to `save`. In `after_commit` callbacks this is a problem: when a transaction includes multiple saves, `saved_changes` only shows the last save's diff, not the cumulative change the transaction actually made to the database.

#### Scenario 1
If `name` changes from "Alice" to "Bob" in one save, then from "Bob" to "Carol" in a second save within the same transaction:

```
  saved_change_to_name       #=> ["Bob", "Carol"]    (last save only)
  transaction_change_to_name #=> ["Alice", "Carol"]   (full transaction)
```

#### Scenario 2
If `name` changes from "Alice" to "Bob" in one save, then `age` changes in a second save within the same transaction:

```
  saved_change_to_name       #=> nil                  (not part of last save)
  transaction_change_to_name #=> ["Alice", "Bob"]     (full transaction)
```

### Changes

This PR introduces a new family of methods that track cumulative attribute changes across the entire transaction. They work in any callback phase (`before_save`, `after_save`, `after_commit`), as well as for cross-model access within the same transaction:

  - `transaction_changes` — hash of all cumulative changes across the transaction
  - `transaction_changes?` — whether the transaction changed any attributes
  - `transaction_change_to_attribute(attr)` — the [old, new] pair
  - `transaction_change_to_attribute?(attr, from:, to:)` — query with options
  - `attribute_before_transaction(attr)` — pre-transaction value of any attribute, including ones that didn't change

During `before_save` / `before_update`, pending unsaved changes are included. If an attribute is changed back to its original value within the transaction, it is correctly omitted.

### Details

The implementation snapshots `@attributes` at the start of the transaction (in `remember_transaction_record_state`) and compares against the current attributes at commit time (in `_compute_committed_changes`). For `before_save` / `before_update`, `_compute_transaction_changes` overlays pending unsaved changes on top of the committed-so-far diff. If an attribute is changed back to its original value within the transaction, it is correctly omitted from `transaction_changes`.

### Why not integrate into `saved_changes`?

**1. Silent breaking change for every existing `after_commit` callback**

Every app that uses `saved_change_to_*` in `after_commit` today has established behavior — even if that behavior is sometimes wrong for multi-save transactions, it's *predictable*. Changing what `saved_change_to_name` returns depending on callback context would silently alter return values across every Rails app on upgrade.

**2. They answer fundamentally different questions**

Consider: create a record (name goes `nil -> "Alice"`), then update it (name goes `"Alice" -> "Bob"`). The last `saved_change_to_name` is `["Alice", "Bob"]`. The `transaction_change_to_name` is `[nil, "Bob"]`. These aren't two views of the same data — they're different data. Collapsing them into one method hides the fact that there are two distinct questions people need to ask.

**3. `after_save` fires inside transactions too**

Inside a transaction with multiple saves, `after_save` fires after each one. `saved_change_to_*` must return per-save changes there, that's its contract. So you can't globally swap the semantics; you'd need the method to introspect whether it's being called from `after_save` or `after_commit`. That introspection mechanism itself would be fragile (what about code called from both contexts? helper methods shared between callbacks?).

### Checklist

### DIFF
--- a/activerecord/lib/active_record/attribute_methods/dirty.rb
+++ b/activerecord/lib/active_record/attribute_methods/dirty.rb
@@ -53,6 +53,9 @@ module ActiveRecord
 
       include ActiveModel::Dirty
 
+      EMPTY_HASH = {}.with_indifferent_access.freeze
+      private_constant :EMPTY_HASH
+
       included do
         if self < ::ActiveRecord::Timestamp
           raise "You cannot include Dirty after Timestamp"
@@ -210,7 +213,7 @@ module ActiveRecord
       # values are arrays containing the original value at the start of the
       # transaction and the final committed value.
       def committed_changes
-        @_committed_changes || {}.with_indifferent_access
+        @_committed_changes || EMPTY_HASH
       end
 
       # Will this attribute change the next time we save?
--- a/activerecord/lib/active_record/attribute_methods/dirty.rb
+++ b/activerecord/lib/active_record/attribute_methods/dirty.rb
@@ -33,6 +33,18 @@ module ActiveRecord
     #   person.saved_change_to_name    # => ["Allison", "Alice"]
     #   person.name_before_last_save   # => "Allison"
     #
+    # After the transaction is committed:
+    #
+    #   person.committed_change_to_name?   # => true
+    #   person.committed_change_to_name    # => ["Allison", "Alice"]
+    #   person.name_before_last_commit     # => "Allison"
+    #   person.committed_changes           # => {"name"=>["Allison", "Alice"]}
+    #
+    # The +committed_change_to_*+ methods track cumulative changes across all
+    # saves within a transaction, making them ideal for use in +after_commit+
+    # callbacks. The +saved_change_to_*+ methods, by contrast, only reflect
+    # the most recent save.
+    #
     # Similar to ActiveModel::Dirty, methods can be invoked as
     # +saved_change_to_name?+ or by passing an argument to the generic method
     # <tt>saved_change_to_attribute?("name")</tt>.
@@ -54,6 +66,11 @@ module ActiveRecord
         attribute_method_prefix("saved_change_to_", parameters: false)
         attribute_method_suffix("_before_last_save", parameters: false)
 
+        # Attribute methods for "changed during the last committed transaction?"
+        attribute_method_affix(prefix: "committed_change_to_", suffix: "?", parameters: "**options")
+        attribute_method_prefix("committed_change_to_", parameters: false)
+        attribute_method_suffix("_before_last_commit", parameters: false)
+
         # Attribute methods for "will change if I call save?"
         attribute_method_affix(prefix: "will_save_change_to_", suffix: "?", parameters: "**options")
         attribute_method_suffix("_change_to_be_saved", "_in_database", parameters: false)
@@ -64,6 +81,7 @@ module ActiveRecord
         super.tap do
           @mutations_before_last_save = nil
           @mutations_from_database = nil
+          @_committed_changes = nil
         end
       end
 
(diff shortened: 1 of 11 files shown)


activerecord/lib/active_record/attribute_methods/dirty.rb
```diff
@@ -119,6 +137,82 @@ def saved_changes
...
+      #   value at the start of the transaction is equal to the given value.
+      #
+      # [+to+]
+      #   When specified, this method will return false unless the value
+      #   committed to the database is equal to the given value.
+      def committed_change_to_attribute?(attr_name, **options)
+        attr_name = attr_name.to_s
+        return false unless @_committed_changes&.key?(attr_name)
+        change = @_committed_changes[attr_name]
+        (options.key?(:from) ? change[0] == _type_cast_for_committed(attr_name, options[:from]) : true) &&
```
ericproulx: Maybe declaring `key: nil, to: nil` instead of `**options`. I don't know no if it's a standard in Rails, but it would simplify your code. Maybe a `key: nil;, from: nil, **options` for future proof.

activerecord/lib/active_record/transactions.rb
```diff
@@ -479,6 +480,40 @@ def remember_transaction_record_state
...
+      # database representation differs from the deserialized Ruby object
+      # (e.g. datetime columns). Instead we compare at the serialized
+      # (database) level and deserialize through the type directly.
+      def _compute_committed_changes
+        return unless @_start_transaction_state
+
+        snapshot_attributes = @_start_transaction_state[:attributes]
+        changes = {}.with_indifferent_access
+
+        @attributes.keys.each do |attr_name|
```
ericproulx: Assuming @attributes is a Hash, maybe using `each_key` instead of `keys.each`.

activerecord/lib/active_record/attribute_methods/dirty.rb
nvasilevski: Could we try to rewrite it in a cleaner way? It's a bit hard to follow with basically duplicated checks and the only difference is the index of `change[x]`

For example it looks like we can return `true` early if `options` don't have both `:from` and `:to` keys

activerecord/lib/active_record/attribute_methods/dirty.rb
```diff
@@ -119,6 +137,82 @@ def saved_changes
...
+      # transaction. If the attribute was changed, the result will be an array
+      # containing the original value at the start of the transaction and the
+      # final committed value.
+      #
+      # This method is useful in +after_commit+ callbacks, to see the full
+      # change in an attribute across the entire transaction. It can be
+      # invoked as +committed_change_to_name+ instead of
+      # <tt>committed_change_to_attribute("name")</tt>.
+      def committed_change_to_attribute(attr_name)
+        @_committed_changes&.[](attr_name.to_s)
```
nvasilevski: Do we have an option to always initialize `@_committed_changes` as an empty hash? So we can get rid of  safe navigation `&.` sprinkled across the code

activerecord/lib/active_record/attribute_methods/dirty.rb
lcampanari: The other `_changed?` methods in this same file all use `**options` too, so this is consistent with its siblings. It's actually the cleanest way to solve the "nil is a valid value" problem without introducing or referencing sentinel constants.

activerecord/lib/active_record/attribute_methods/dirty.rb
lcampanari: The `nil` state carries meaning here. It distinguishes "no transaction has been committed" from "a transaction was committed with no changes". If you look at `attribute_before_last_commit`, there are three branches: 

1. Attribute was changed in the transaction -> return the old value from the change pair
2. Transaction happened but attribute wasn't changed (`@_committed_changes` is an empty hash) -> return current value (which equals the pre-transaction value)
3. No transaction committed yet (`@_committed_changes` is `nil`) -> return `nil`

If we always initialize `@_committed_changes` as an empty hash, case 3 collapses into case 2.

activerecord/lib/active_record/transactions.rb
lcampanari: `@attributes` is an `AttributeSet`, which doesn't have the `each_key` method

activerecord/lib/active_record/attribute_methods/dirty.rb
lcampanari: Yes, that can be improved!

lcampanari: Fixed in this commit [Refactor committed_change_to_attribute? for better redability](https://github.com/rails/rails/pull/56765/commits/e838c972b71665b16193cc873f3789c9d94256e0)

activerecord/lib/active_record/attribute_methods/dirty.rb
lcampanari: This method has been removed in the latest refactor: https://github.com/rails/rails/pull/56765/commits/440875ed8927396c342982178f380f1036005221

PR #56767, opened 2026-02-11
Don't eagerly cache JSON Encoder in ActiveRecord::Type::Json

xathien:
If we cache the encoder when the class is first referenced, it prevents us from being able to proper override the encoder later. Instead, just make use of ActiveSupport::JSON's cached encoder and prebuilt `without_escaping` option.

[Fix #56766]

### Checklist

Welcoming guidance on how to properly test this in excess of current tests. 🙂

### DIFF
--- a/activerecord/lib/active_record/type/json.rb
+++ b/activerecord/lib/active_record/type/json.rb
@@ -25,10 +25,8 @@ module ActiveRecord
         end
       end
 
-      JSON_ENCODER = ActiveSupport::JSON::Encoding.json_encoder.new(escape: false)
-
       def serialize(value)
-        JSON_ENCODER.encode(value) unless value.nil?
+        ActiveSupport::JSON::Encoding.encode_without_escape(value) unless value.nil?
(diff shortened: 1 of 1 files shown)


activerecord/lib/active_record/type/json.rb
```diff
@@ -25,10 +25,8 @@ def deserialize(value)
         end
       end
 
-      JSON_ENCODER = ActiveSupport::JSON::Encoding.json_encoder.new(escape: false)
-
       def serialize(value)
-        JSON_ENCODER.encode(value) unless value.nil?
+        ActiveSupport::JSON::Encoding.encode_without_escape(value) unless value.nil?
```
xathien, line 29: I think this is technically more optimal but maybe less idiomatic than the (basically) equivalent:

```suggestion
        ActiveSupport::JSON.encode(value, escape: false) unless value.nil?
```

but I'm more than happy to change it if people deem the other version prettier.

PR #56775, opened 2026-02-11
Update Action Cable origin check to respect X-Forwarded-Host

jordan-brough:
### Motivation / Background

ActionCable was giving me errors like this:
```
Request origin not allowed: https://...
```
with a reverse proxy where `X-Forwarded-Host` should be used instead of `Host`.

### Detail

The `allow_same_origin_as_host` check in `allow_request_origin?` compared the browser's Origin
header against the raw `HTTP_HOST`, which fails behind a reverse proxy where the internal host
differs from the public host.

This updates to use `request.host_with_port` (which respects `X-Forwarded-Host`) and `request.ssl?`
instead, **consistent with how the rest of Rails resolves the host**.

### Checklist

### DIFF
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix Action Cable origin check to respect `X-Forwarded-Host` behind reverse proxies.
+
+    The `allow_same_origin_as_host` check previously compared against the raw
+    `HTTP_HOST` header, which fails when a proxy forwards requests with a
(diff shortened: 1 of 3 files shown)


jordan-brough:
cc @matthewd, since you committed the original [Permit same-origin connections by default](https://github.com/rails/rails/commit/dae404473409fcab0e07976aec626df670e52282), back in 2016?

jordan-brough:
Would love to get a review on this. cc @byroot @gmcgibbon since I see you've both made actioncable commits recently.

PR #21254, opened 2015-08-16
Add Rails commands for existing rake tasks

ccallebs:
Aims to implement #18878.

This is a proof of concept branch. The code is a bit messy and I've left comments where I plan to clean it up. I'm also not sure what dependencies will end up being shared among tasks, so a lot of functionality is in the Base class at the moment.

Placing each set of rake tasks inside of a wrapper class allows some flexibility for the future in terms of providing contextual documentation. For example, `rails test --help` could provide a custom help message for only test commands (and encapsulate those messages inside of each class).

Currently working:
- Original commands
- `rails test:db`

Before I go any further moving more rake tasks to this architecture, I wanted to get feedback on the implementation. I think this approach is sensible and I can knock out the rest of them quickly.

PR #56779, opened 2026-02-11
Optimize `ActiveRecord::Base.new` to conditionally remove STI checks

flavorjones:
### Motivation / Background

`ActiveRecord::Base` overrides the `.new` class method to handle runtime checks for STI models, checking the "type" column and invoking the appropriate subclass's constructor. However, for non-STI models, this method brings disadvantages:

- The STI check is unnecessary overhead for non-STI classes
- In Ruby 4.0, the overridden method prevents a "fast path allocation" optimization from providing additional speedups

This pull request optimizes `ActiveRecord::Base.new` so that non-STI classes can avoid the additional overhead of this method whenever possible, and also take advantage of the Ruby 4.0 "fast path allocation" optimization.

(Note that older versions of Ruby without the "fast allocation path" optimization still show significant speedups due to the removal of runtime checks from the allocation path.)

### Detail

The control flow for STI models is unchanged, and the benchmarks show no change on object allocation performance for STI models.

For non-STI models that do not override `.new`, the method is overridden to point at `Class.new`:

``` ruby
define_singleton_method :new, Class.instance_method(:new)
```

This essentially resets the constructor method back to Ruby's default, eliminates the STI check on every instantiation, and enables Ruby 4.0's fast path optimization for `Class#new`.

Some noteworthy edge cases are taken care of:

- Do not optimize the constructor if the model has already redefined `.new`.
- Tests assert that "deoptimization" happens correctly if a class dynamically redefines `.new`.

The test coverage will also check `opt_new_hit` and `opt_new_miss` debug counters if run with a debug build of Ruby, which should make it a bit easier to determine whether the fast path allocation optimization is working properly

### Benchmark results

Summary: Active Record models not using STI and not redefining `.new` were 15-17% faster on Ruby 4, and 8-17% faster on Ruby 3.4 and 3.3.

When the model redefines `.new` or uses STI, the benchmarks were generally close to the error bars or "same-ish" (my benchmark machine has some jitter I could not eliminate).

```
ruby 4.0.1 (2026-01-13 revision e04267a14b) +PRISM [x86_64-linux]
                              YJIT disabled             YJIT enabled
                              ------------------------  ------------------------
  non-STI, default .new       1.15x faster (±0.01)      1.17x faster (±0.04)
                              68.7k vs 59.6k i/s        197.2k vs 168.4k i/s
  non-STI, redefined .new     same-ish                  same-ish
                              60.0k vs 59.9k i/s        164.5k vs 161.3k i/s
  STI                         same-ish                  same-ish
                              39.2k vs 39.3k i/s        120.2k vs 119.7k i/s

ruby 3.4.8 (2025-12-17 revision 995b59f666) +PRISM [x86_64-linux]
                              YJIT disabled             YJIT enabled
                              ------------------------  ------------------------
  non-STI, default .new       1.13x faster (±0.01)      1.06x faster (±0.04)
                              77.5k vs 68.8k i/s        218.3k vs 206.7k i/s
  non-STI, redefined .new     same-ish                  same-ish
                              68.4k vs 68.2k i/s        191.1k vs 194.4k i/s
  STI                         same-ish                  same-ish
                              42.6k vs 42.6k i/s        141.8k vs 140.3k i/s

ruby 3.3.10 (2025-10-23 revision 343ea05002) [x86_64-linux]
                              YJIT disabled             YJIT enabled
                              ------------------------  ------------------------
  non-STI, default .new       1.10x faster (±0.01)      1.09x faster (±0.04)
                              68.8k vs 62.4k i/s        173.6k vs 159.1k i/s
  non-STI, redefined .new     same-ish                  same-ish
                              61.3k vs 62.1k i/s        145.9k vs 149.6k i/s
  STI                         same-ish                  same-ish
                              39.3k vs 39.3k i/s        108.6k vs 107.6k i/s
```

Full benchmark output here: [benchmark.txt](https://github.com/user-attachments/files/25499368/benchmark.txt)

Full benchmark scripts here: [benchmark.zip](https://github.com/user-attachments/files/25241475/benchmark.zip)

#### Benchmark setup

Benchmarks were run on a benchmark rig:

- frequency scaling turned off
- pinned to a non-hyperthreaded core
- that core was kernel-isolated from the kernel scheduler
- disabled the timer tick, and the chosen CPU did not handle interrupts
- intel turbo boost disabled
- scaling governor set to "performance"

The benchmark script also disables Ruby GC and runs a full mark-and-sweep after warmup before the measured run.

### Context on Ruby 4.0's "Fast path allocation"

More information about Ruby 4.0's "fast allocation path" optimization:

- blog post: [Fast Allocations in Ruby 3.5 | Rails at Scale](https://railsatscale.com/2025-05-21-fast-allocations-in-ruby-3-5/)
- redmine ticket: [Feature #21254: Inlining Class#new - Ruby - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/21254)
- pull request: [Inline Class#new. by tenderlove � Pull Request #13080 � ruby/ruby](https://github.com/ruby/ruby/pull/13080)
- Ruby commit: ruby/ruby@8ac8225c

Because the fast path optimization implemented at ruby/ruby@8ac8225c simply checks if the method is the same as `Class.new` (`rb_class_new_instance_pass_kw`), redefining `.new` to point at `Class.new` allows the "fast path allocation" optimization to work.

Here's the technique demonstrated using Ruby 4.0 compiled with `-DUSE_DEBUG_COUNTER=1`.

``` ruby
#!/usr/bin/env ruby

class A
  class << self
    def new
      # imagine code that is unnecessary for B
    end
  end
end

class B < A
end

RubyVM.reset_debug_counters
10.times { B.new }
RubyVM.show_debug_counters

B.define_singleton_method :new, Class.instance_method(:new)

RubyVM.reset_debug_counters
10.times { B.new }
RubyVM.show_debug_counters
```

Running this, the first set of debug counters shows:

```
[RUBY_DEBUG_COUNTER]    opt_new_hit                                  0
[RUBY_DEBUG_COUNTER]    opt_new_miss                                10
```

But after defining `B.new`:

```
[RUBY_DEBUG_COUNTER]    opt_new_hit                                 10
[RUBY_DEBUG_COUNTER]    opt_new_miss                                 0
```

This implementation takes care to check whether `.new` has already been redefined by the class, and if so does not modify the class.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/inheritance.rb
+++ b/activerecord/lib/active_record/inheritance.rb
@@ -67,10 +67,12 @@ module ActiveRecord
           if subclass.nil? && base_class?
             subclass = subclass_from_attributes(column_defaults)
           end
-        end
 
-        if subclass && subclass != self
-          subclass.new(attributes, &block)
+          if subclass && subclass != self
+            subclass.new(attributes, &block)
+          else
+            super
+          end
         else
           super
         end
@@ -236,6 +238,19 @@ module ActiveRecord
       end
 
       protected
+        def load_schema! # :nodoc:
+          super
+          optimize_new_allocation
+        end
+
+        def reload_schema_from_cache(*) # :nodoc:
+          if @_new_optimized
+            singleton_class.remove_method(:new)
+            @_new_optimized = false
+          end
+          super
+        end
+
         # Returns the class type of the record using the current module as a prefix. So descendants of
         # MyApp::Business::Account would appear as MyApp::Business::AccountSubclass.
         def compute_type(type_name)
@@ -267,6 +282,7 @@ module ActiveRecord
         end
 
         def set_base_class # :nodoc:
+          @_new_optimized = false
           @base_class = if self == Base
             self
           else
@@ -283,6 +299,14 @@ module ActiveRecord
         end
 
       private
+        def optimize_new_allocation
+          return if _has_attribute?(inheritance_column)
+          return if method(:new).unbind != ActiveRecord::Base.method(:new).unbind
+
+          @_new_optimized = true
+          define_singleton_method(:new, Class.instance_method(:new))
+        end
+
         def inherited(subclass)
           super
           subclass.set_base_class
--- a/activerecord/test/cases/fast_path_allocation_test.rb
+++ b/activerecord/test/cases/fast_path_allocation_test.rb
@@ -1,0 +1,234 @@
+# frozen_string_literal: true
+
+require "cases/helper"
+
+class FastPathAllocationTest < ActiveRecord::TestCase
+  test "non-STI class is optimized" do
+    klass = non_sti_class
+
+    assert_not new_is_native?(klass)
+
+    klass.load_schema
+
+    assert_new_is_optimized(klass)
+  end
+
+  test "non-STI class with abstract superclass is optimized" do
+    abstract = Class.new(ActiveRecord::Base) { self.abstract_class = true }
+    klass = non_sti_class(abstract)
+
+    assert_not new_is_native?(klass)
+
+    klass.load_schema
+
+    assert_new_is_optimized(klass)
+  end
+
+  test "STI class and its subclasses are not optimized" do
+    base = sti_class
+    sub = Class.new(base)
+
+    assert_not new_is_native?(base)
+    assert_not new_is_native?(sub)
+
+    base.load_schema
+    sub.load_schema
+
+    assert_new_is_not_optimized(base)
+    assert_new_is_not_optimized(sub)
+  end
+
+  test "class with redefined .new is not optimized" do
+    klass = non_sti_class
+    klass.singleton_class.define_method(:new) { |attributes = nil, &block| super(attributes, &block) }
+
+    assert_not new_is_native?(klass)
+
+    klass.load_schema
+
+    assert_new_is_not_optimized(klass)
+  end
+
+  test "class with .new redefined in superclass is not optimized" do
+    abstract = Class.new(ActiveRecord::Base) { self.abstract_class = true }
+    abstract.singleton_class.define_method(:new) { |attributes = nil, &block| super(attributes, &block) }
+    klass = non_sti_class(abstract)
+
+    assert_not new_is_native?(klass)
+
+    klass.load_schema
+
+    assert_new_is_not_optimized(klass)
+  end
+
+  test "class with prepended .new is not optimized" do
+    mod = Module.new do
+      def new(attributes = nil, &block)
+        super
+      end
+    end
+    klass = non_sti_class
+    klass.singleton_class.prepend(mod)
+
+    assert_not new_is_native?(klass)
+
+    klass.load_schema
+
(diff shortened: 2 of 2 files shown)


activerecord/lib/active_record/inheritance.rb
```diff
@@ -67,12 +68,26 @@ def new(attributes = nil, &block)
...
+            super
+          elsif method(:new).unbind != ActiveRecord::Base.method(:new).unbind
+            # The class redefines `new`, so we can't optimize.
+            @_redefined_new = true
+            super
+          else
+            # Redefine `new` to be the C method `Class.new`, and dispatch to it.
+            # This eliminates the overhead of the checks in this method, and on Ruby 4.0 takes
+            # advantage of the fast path optimization.
+            define_singleton_method(:new, Class.instance_method(:new))
```
byroot: Doing this on the first new isn't ideal.

Have you considered doing this when the schema is loaded?

flavorjones: @byroot Can you say a bit more about what your concern with this approach is? I'm happy to explore other approaches but I'd also like to understand why you'd prefer not to do it this way.

byroot: Right. My concern is that typically you don't allocate a model during boot, only during the first request or job that make use of that model, so post fork, and during a request cycle.

Defining a method like this one does cause some VM cache invalidations etc, and write into the class.

It is always preferable to perform this sort of optimization during boot (aka pre-fork) when possible, and in this case I think it's possible.

flavorjones: I'll explore it.

flavorjones:
@byroot Thanks again for the feedback. I've pushed a new commit that moves fast-path allocation optimization to schema load time as suggested, and performance is essentially unchanged from the earlier benchmarks.

The failing CI jobs look like unrelated flakes.

byroot:
Love it. Congrats on becoming a JIT developer 😂

gmcgibbon:
I believe this might cause an issue for some apps:

```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails", branch: "main", ref: "3295ff950b6c4d9c779d942ac73d9d2d498af898"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.string :body
  end
end

  def initialize(attributes)
    super
  end
end

Post.columns # force schema load

  def test_init
    Post.new
    assert true
  end
end
```

This used to pass and now fails (try f5255a4089cff69af0ce3b63fda751bdc6dd93cb). It isn't clear to me if this is a regression that's worth reverting. The previous behaviour seems wrong.

byroot:
Yeah, we ran into the same issue when we did our upgrade, but the fix was trivial and as you said the old behavior didn't seem particularly correct nor desirable.

So I'm fine with that minor incompatibility.

PR #56782, opened 2026-02-11
 Use instance_variables_to_inspect instead of custom inspect methods …

mbast100:
### Motivation / Background

Ruby 4 added [a new hook `instance_variables_to_inspect`](https://docs.ruby-lang.org/en/4.0/Object.html#method-i-inspect) that lets classes control which instance variables are shown in `inspect` output without overriding `inspect` entirely.

  ### Detail                                                              
   
  There are a total of 56 instances where there is a custom `def inspect` method.                                                   

  **Hide all instance variables (8 changes)**
  These classes override `inspect` to show only `#<ClassName:0xADDRESS>` with no instance variables, likely to avoid leaking secrets or because the object is too large to inspect usefully.

  | Class | File | Reason |
  |-------|------|--------|
  | `ActiveSupport::MessageVerifier` |
  `activesupport/lib/active_support/message_verifier.rb` | Hides secret
  keys |
  | `ActiveSupport::MessageEncryptor` |
  `activesupport/lib/active_support/message_encryptor.rb` | Hides secret
  keys |
  | `ActiveSupport::KeyGenerator` |
  `activesupport/lib/active_support/key_generator.rb` | Hides secret key |
  | `AR::Encryption::Cipher::Aes256Gcm` |
  `activerecord/lib/active_record/encryption/cipher/aes256_gcm.rb` | Hides
   encryption key |
  | `AbstractController::Base` |
  `actionpack/lib/abstract_controller/base.rb` | Too large to inspect |
  | `ActionCable::Connection::Base` |
  `actioncable/lib/action_cable/connection/base.rb` | Too large to inspect|
  | `Rails::Application::Configuration` |`railties/lib/rails/application/configuration.rb` | Too large to inspect|
  | `Rails::Railtie` | `railties/lib/rails/railtie.rb` | Hide all internals |

  **Show specific instance variables (7 changes)**

  | Class | File | Ivars shown |
  |-------|------|-------------|
  | `ActiveModel::Errors` | `activemodel/lib/active_model/errors.rb` |
  `@errors` |
  | `ActiveModel::Error` | `activemodel/lib/active_model/error.rb` |
  `@attribute`, `@type`, `@options` |
  | `ActionController::Parameters` |
  `actionpack/lib/action_controller/metal/strong_parameters.rb` |
  `@parameters`, `@permitted` |
  | `ActionText::Attachment` | `actiontext/lib/action_text/attachment.rb`
  | `@attachable` |
  | `ActiveSupport::Cache::RedisCacheStore` |
  `activesupport/lib/active_support/cache/redis_cache_store.rb` |
  `@options`, `@redis` |
  | `ActiveSupport::Cache::NullStore` |
  `activesupport/lib/active_support/cache/null_store.rb` | `@options` |
  | `ActiveSupport::Cache::FileStore` |
  `activesupport/lib/active_support/cache/file_store.rb` | `@cache_path`,
  `@options` |

  <details>
  <summary><strong>Not migrated (41 instances)</strong></summary>

  These `def inspect` methods cannot be replaced with
  `instance_variables_to_inspect` because they use computed values,
  conditional logic, method calls, non-standard formatting, or delegate to
   other objects.

  | Class | File | Reason |
  |-------|------|--------|
  | `ActionView::Base` | `actionview/lib/action_view/base.rb` | Defined on
   anonymous subclass via `Class.new`; hardcodes class name |
  | `ActiveRecord::Core` (instance) |
  `activerecord/lib/active_record/core.rb` | Complex filtering of AR
  attributes via method calls |
  | `ActiveRecord::Core` (class) |
  `activerecord/lib/active_record/core.rb` | Conditional logic based on
  `schema_loaded?`, `connected?`, `table_exists?` |
  | `ActiveRecord::Relation` |
  `activerecord/lib/active_record/relation.rb` | Shows first 11 records
  with truncation; calls `loaded?`, `annotate` |
  | `ActiveRecord::Relation` (Explain) |
  `activerecord/lib/active_record/relation.rb` | Inner class; calls
  `exec_explain` to execute queries |
  | `ActiveRecord::Promise` | `activerecord/lib/active_record/promise.rb`
  | Calls computed `status` method |
  | `ActiveRecord::Associations::CollectionProxy` |
  `activerecord/lib/active_record/associations/collection_proxy.rb` |
  Calls `load_target` conditionally then delegates to `super` |
  | `AR::ConnectionAdapters::AbstractAdapter` |
  `activerecord/lib/active_record/connection_adapters/abstract_adapter.rb`
   | Method chains, conditional fields based on shard/name |
  | `AR::ConnectionAdapters::ConnectionPool` | `activerecord/lib/active_re
  cord/connection_adapters/abstract/connection_pool.rb` | Calls
  `name_inspect`, `shard_inspect`, `db_config.env_name`, `role` methods |
  | `AR::ConnectionAdapters::QueryIntent` |
  `activerecord/lib/active_record/connection_adapters/query_intent.rb` |
  Shows attr_readers but format uses bare values without `@` prefix |
  | `AR::DatabaseConfigurations::DatabaseConfig` | `activerecord/lib/activ
  e_record/database_configurations/database_config.rb` | `adapter_class`
  is a memoized method — may show `nil` before resolution |
  | `Arel::Nodes::BoundSqlLiteral` |
  `activerecord/lib/arel/nodes/bound_sql_literal.rb` |
  `sql_with_placeholders` is a method; conditionally shows named or
  positional binds |
  | `ActionDispatch::Request` |
  `actionpack/lib/action_dispatch/http/request.rb` | Calls `method`,
  `original_url`, `remote_ip` (lazy-loaded) |
  | `ActionDispatch::Request::Session` |
  `actionpack/lib/action_dispatch/request/session.rb` | Conditional on
  `loaded?` — shows `super` when loaded, custom string when not |
  | `ActionDispatch::MiddlewareStack::Middleware` |
  `actionpack/lib/action_dispatch/middleware/stack.rb` | Conditional on
  `klass.is_a?(Module)`; returns class name string |
  | `ActionDispatch::Routing::Redirect` |
  `actionpack/lib/action_dispatch/routing/redirection.rb` | Non-standard
  format `"redirect(301)"`; no `#<>` wrapping |
  | `ActionDispatch::Routing::PathRedirect` |
  `actionpack/lib/action_dispatch/routing/redirection.rb` | Non-standard
  format `"redirect(301, path)"` |
  | `ActionDispatch::Routing::OptionRedirect` |
  `actionpack/lib/action_dispatch/routing/redirection.rb` | Non-standard
  format with computed options hash |
  | `ActionText::Content` | `actiontext/lib/action_text/content.rb` |
  Calls `to_html` (computed from `fragment`) and `truncate` |
  | `ActionText::AttachmentGallery` |
  `actiontext/lib/action_text/attachment_gallery.rb` | `size` is computed
  from `attachments.size` |
  | `ActionText::Editor::Configurator` |
  `actiontext/lib/action_text/editor/configurator.rb` |
  `configurations.keys` is computed; conditional display |
  | `ActionText::Editor::Registry` |
  `actiontext/lib/action_text/editor/registry.rb` | `configurations.keys`
  is computed; conditional display |
  | `ActionView::Template` | `actionview/lib/action_view/template.rb` |
  Calls `short_identifier` (computed with conditional `Rails.root` logic)|
  | `ActiveSupport::Duration` |
  `activesupport/lib/active_support/duration.rb` | Complex sorting,
  conditional pluralization, `to_sentence` |
  | `ActiveSupport::TimeWithZone` |
  `activesupport/lib/active_support/time_with_zone.rb` | Calls `time`
  (lazy-loads), `zone` (computed), `formatted_offset` |
  | `ActiveSupport::EncryptedConfiguration` |
  `activesupport/lib/active_support/encrypted_configuration.rb` | `keys`
  lazily decrypts YAML |
  | `ActiveSupport::CombinedConfiguration` |
  `activesupport/lib/active_support/combined_configuration.rb` | `keys` is
   a computed method |
  | `ActiveSupport::EnvConfiguration` |
  `activesupport/lib/active_support/env_configuration.rb` | `keys` is a
  computed method |
  | `ActiveSupport::Notifications::Fanout` |
  `activesupport/lib/active_support/notifications/fanout.rb` | Computes a
  summary count from subscriber hashes |
  | `ActiveSupport::Cache::MemoryStore` |
  `activesupport/lib/active_support/cache/memory_store.rb` | Shows
  `@data.size` (method call on ivar) |
  | `ActiveSupport::Cache::MemCacheStore` |
  `activesupport/lib/active_support/cache/mem_cache_store.rb` |
  Conditional — shows `@data` or `@mem_cache_options` |
  | `ActiveSupport::EventedFileUpdateChecker` |
  `activesupport/lib/active_support/evented_file_update_checker.rb` |
  Shows `@core.files.to_a` (method chain on ivar) |
  | `ActiveSupport::OrderedOptions` |
  `activesupport/lib/active_support/ordered_options.rb` | Hash subclass;
  calls `super` (Hash#inspect) |
  | `ActiveSupport::InheritableOptions` |
  `activesupport/lib/active_support/ordered_options.rb` | Hash subclass;
  calls `to_h.inspect` |
  | `ActiveSupport::TestCase` |
  `activesupport/lib/active_support/test_case.rb` | Uses
  `Object.instance_method(:to_s).bind_call(self)` |
  | `ActiveSupport::Deprecation::DeprecationProxy` |
  `activesupport/lib/active_support/deprecation/proxy_wrappers.rb` | Proxy
   object; delegates to `target.inspect` |
  | `ActiveSupport::Deprecation::DeprecatedClassProxy` |
  `activesupport/lib/active_support/deprecation/proxy_wrappers.rb` | Proxy
   object; delegates to `target.inspect` |
  | `ActiveSupport::Testing::EventReporterAssertions::Event` |
  `activesupport/lib/active_support/testing/event_reporter_assertions.rb`
  | Test support inner class; non-standard format |
  | `ActiveStorage::Service` |
  `activestorage/lib/active_storage/service.rb` | Conditional on
  `name.present?` |
  | `ActiveStorage::Service::Configurator` |
  `activestorage/lib/active_storage/service/configurator.rb` |
  `configurations.keys` is computed; conditional display |
  | `ActiveStorage::Service::Registry` |
  `activestorage/lib/active_storage/service/registry.rb` |
  `configurations.keys` is computed; conditional display |

  </details>

### Checklist

### DIFF
--- a/actioncable/lib/action_cable/connection/base.rb
+++ b/actioncable/lib/action_cable/connection/base.rb
@@ -3,6 +3,7 @@
 # :markup: markdown
 
 require "action_dispatch"
+require "active_support/inspect_backport"
 require "active_support/rescuable"
 
 module ActionCable
@@ -165,11 +166,13 @@ module ActionCable
         send_async :handle_close
       end
 
-      def inspect # :nodoc:
-        "#<#{self.class.name}:#{'%#016x' % (object_id << 1)}>"
-      end
+      include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
 
       private
+        def instance_variables_to_inspect
+          [].freeze
+        end
+
         attr_reader :websocket
         attr_reader :message_buffer
 
--- a/activemodel/lib/active_model/error.rb
+++ b/activemodel/lib/active_model/error.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "active_support/inspect_backport"
 
 module ActiveModel
   # = Active \Model \Error
@@ -195,9 +196,12 @@ module ActiveModel
       attributes_for_hash.hash
     end
 
-    def inspect # :nodoc:
-      "#<#{self.class.name} attribute=#{@attribute}, type=#{@type}, options=#{@options.inspect}>"
-    end
+    include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
+
+    private
+      def instance_variables_to_inspect
+        [:@attribute, :@type, :@options].freeze
+      end
 
     protected
       def attributes_for_hash
--- a/activemodel/test/cases/error_test.rb
+++ b/activemodel/test/cases/error_test.rb
@@ -252,4 +252,11 @@ class ErrorTest < ActiveModel::TestCase
 
     assert_equal({ error: :invalid, foo: :bar }, error.details)
   end
+
+  test "inspect" do
+    person = Person.new
+    error = ActiveModel::Error.new(person, :name, :too_short, count: 5)
+
+    assert_match(/\A#<ActiveModel::Error:0x[0-9a-f]+ @attribute=:name, @type=:too_short, @options=#{Regexp.escape({ count: 5 }.inspect)}>\z/, error.inspect)
+  end
 end
--- a/activesupport/lib/active_support/inspect_backport.rb
+++ b/activesupport/lib/active_support/inspect_backport.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module ActiveSupport
+  # Provides a Ruby 4.0-compatible +inspect+ method for Ruby < 4.0.
+  #
+  # Ruby 4.0 introduced +instance_variables_to_inspect+, which lets classes
+  # control which instance variables appear in +inspect+ output without
+  # overriding +inspect+ entirely. This module backports that behavior so
+  # classes can define +instance_variables_to_inspect+ on any Ruby version.
+  #
+  #   class MyClass
+  #     include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
+  #
+  #     private
+  #       def instance_variables_to_inspect
+  #         [:@name, :@status].freeze
+  #       end
+  #   end
+  module InspectBackport # :nodoc:
+    def inspect
+      ivars = instance_variables_to_inspect
+      klass = self.class.name || self.class.inspect
+      addr = "0x%x" % object_id
+
+      if ivars.empty?
+        "#<#{klass}:#{addr}>"
+      else
+        pairs = ivars.map { |ivar| "#{ivar}=#{instance_variable_get(ivar).inspect}" }
+        "#<#{klass}:#{addr} #{pairs.join(", ")}>"
+      end
+    end
+  end
+end
--- a/activesupport/test/inspect_backport_test.rb
+++ b/activesupport/test/inspect_backport_test.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require_relative "abstract_unit"
+require "active_support/inspect_backport"
+
+class InspectBackportTest < ActiveSupport::TestCase
+  test "inspect with no instance variables shows only class and address" do
+    klass = Class.new do
+      include ActiveSupport::InspectBackport
+
+      private
+        def instance_variables_to_inspect
+          [].freeze
+        end
+    end
+
+    obj = klass.new
+    assert_match(/\A#<#{Regexp.escape(klass.inspect)}:0x[0-9a-f]+>\z/, obj.inspect)
+  end
+
+  test "inspect with instance variables shows them" do
+    klass = Class.new do
+      include ActiveSupport::InspectBackport
+      def initialize
+        @name = "test"
+        @count = 42
+      end
+
+      private
+        def instance_variables_to_inspect
+          [:@name, :@count].freeze
+        end
+    end
+
+    obj = klass.new
+    assert_match(/@name="test"/, obj.inspect)
+    assert_match(/@count=42/, obj.inspect)
+  end
+
+  test "inspect uses class name when available" do
+    obj = NamedExample.new("hello")
+    assert_match(/\A#<InspectBackportTest::NamedExample:0x[0-9a-f]+ @value="hello">\z/, obj.inspect)
+  end
+
+  test "inspect with unset instance variable shows nil" do
+    klass = Class.new do
+      include ActiveSupport::InspectBackport
+
+      private
+        def instance_variables_to_inspect
+          [:@missing].freeze
+        end
+    end
+
+    obj = klass.new
+    assert_match(/@missing=nil/, obj.inspect)
+  end
+
+  class NamedExample
+    include ActiveSupport::InspectBackport
+
+    def initialize(value)
+      @value = value
+    end
+
+    private
+      def instance_variables_to_inspect
+        [:@value].freeze
+      end
+  end
+end
--- a/actioncable/test/connection/base_test.rb
+++ b/actioncable/test/connection/base_test.rb
@@ -133,6 +133,13 @@ class ActionCable::Connection::BaseTest < ActionCable::TestCase
     end
   end
 
+  test "inspect does not show internals" do
+    run_in_eventmachine do
+      connection = open_connection
+      assert_match(/\A#<ActionCable::Connection::BaseTest::Connection:0x[0-9a-f]+>\z/, connection.inspect)
+    end
+  end
+
   private
     def open_connection
       env = Rack::MockRequest.env_for "/test", "HTTP_CONNECTION" => "upgrade", "HTTP_UPGRADE" => "websocket",
--- a/actionpack/lib/abstract_controller/base.rb
+++ b/actionpack/lib/abstract_controller/base.rb
@@ -4,6 +4,7 @@
 
 require "abstract_controller/error"
 require "active_support/descendants_tracker"
+require "active_support/inspect_backport"
 require "active_support/core_ext/module/anonymous"
 require "active_support/core_ext/module/attr_internal"
 
@@ -196,11 +197,13 @@ module AbstractController
       @_config ||= self.class.config.inheritable_copy
     end
 
-    def inspect # :nodoc:
-      "#<#{self.class.name}:#{'%#016x' % (object_id << 1)}>"
-    end
+    include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
 
     private
+      def instance_variables_to_inspect
+        [].freeze
+      end
+
       # Returns true if the name can be considered an action because it has a method
       # defined in the controller.
       #
--- a/actiontext/lib/action_text/attachment.rb
+++ b/actiontext/lib/action_text/attachment.rb
@@ -3,6 +3,7 @@
 # :markup: markdown
 
 require "active_support/core_ext/object/try"
+require "active_support/inspect_backport"
 
 module ActionText
   # # Action Text Attachment
@@ -128,11 +129,13 @@ module ActionText
       to_html
     end
 
-    def inspect
-      "#<#{self.class.name} attachable=#{attachable.inspect}>"
-    end
+    include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
 
     private
+      def instance_variables_to_inspect
+        [:@attachable].freeze
+      end
+
       def node_attributes
         @node_attributes ||= ATTRIBUTES.to_h { |name| [ name.underscore, node[name] ] }.compact
       end
--- a/actiontext/test/unit/attachment_test.rb
+++ b/actiontext/test/unit/attachment_test.rb
@@ -95,6 +95,12 @@ class ActionText::AttachmentTest < ActiveSupport::TestCase
     assert_equal "pages/page", attachable.to_editor_content_attachment_partial_path
   end
 
+  test "inspect shows attachable" do
+    attachment = ActionText::Attachment.from_attachable(attachable)
+
+    assert_match(/\A#<ActionText::Attachment:0x[0-9a-f]+ @attachable=/, attachment.inspect)
+  end
+
   private
     def attachment_from_html(html)
       ActionText::Content.new(html).attachments.first
--- a/activemodel/lib/active_model/errors.rb
+++ b/activemodel/lib/active_model/errors.rb
@@ -3,6 +3,7 @@
 require "active_support/core_ext/array/conversions"
 require "active_support/core_ext/string/inflections"
 require "active_support/core_ext/object/deep_dup"
+require "active_support/inspect_backport"
 require "active_model/error"
 require "active_model/nested_error"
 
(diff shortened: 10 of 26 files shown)


activemodel/test/cases/error_test.rb
```diff
@@ -252,4 +252,15 @@ def test_initialize
...
   end
+
+  test "inspect" do
+    person = Person.new
+    error = ActiveModel::Error.new(person, :name, :too_short, count: 5)
+
+    if RUBY_VERSION >= "4.0"
+      assert_match(/\A#<ActiveModel::Error:0x[0-9a-f]+ @attribute=:name, @type=:too_short, @options=\{count: 5\}>\z/, error.inspect)
+    else
+      assert_equal "#<ActiveModel::Error attribute=name, type=too_short, options={count: 5}>", error.inspect
```
skipkayhil:
```suggestion
      assert_equal "#<ActiveModel::Error attribute=name, type=too_short, options=#{{count: 5}.inspect}>", error.inspect
```

Ruby 3.4+ changed the output of `inspect` for `Hash`, but if you use `inspect` in the expected value it'll work on all versions

skipkayhil (approved):
This looks really great, thanks!

One thought I'm having now: it may be better if `inspect` output only changes when Rails is upgraded instead of when Ruby is upgraded 🤔 we may want all the tests to always assert that the "new" output will happen instead of keeping the old output only until an app upgrades Ruby.

cc @byroot would also appreciate your review

actioncable/lib/action_cable/connection/base.rb
```diff
@@ -165,8 +165,14 @@ def on_close(reason, code) # :nodoc:
         send_async :handle_close
       end
 
-      def inspect # :nodoc:
-        "#<#{self.class.name}:#{'%#016x' % (object_id << 1)}>"
+      if RUBY_VERSION >= "4.0"
+        def instance_variables_to_inspect # :nodoc:
```
byroot:
```suggestion
        private def instance_variables_to_inspect
```

Not a huge deal, but these methods are meant to be private. Hence I think we don't need the `:nodoc:`

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -1056,8 +1056,14 @@ def stringify_keys # :nodoc:
       dup
     end
 
-    def inspect
-      "#<#{self.class} #{@parameters} permitted: #{@permitted}>"
+    if RUBY_VERSION >= "4.0"
+      def instance_variables_to_inspect # :nodoc:
+        [:@parameters, :@permitted]
```
byroot:
```suggestion
        [:@parameters, :@permitted].freeze
```

In 4.0, literal arrays containing only frozen elements save an allocation if they are frozen themselves. Not that it matters much here but.

byroot:
So I meant to do something like that for a while, and @skipkayhil did explore it too: https://github.com/rails/rails/compare/main...skipkayhil:rails:hm-qwysxtxtowwkxzqr

I told @skipkayhil I didn't like the idea of patching `Object#inspect`, and still think we shouldn't. However to avoid the numerous duplicated logic, and different output, we could have a module that you can include to get a Ruby 4.0 like `#inspect` method.

This it would look like:

```ruby
  include ActiveSupport::ObjectInspectBackport if RUBY_VERSION < "4"
  private
    def instance_variables_to_inspect
     # ...
    end
```

So less duplication.

mbast100:
> So I meant to do something like that for a while, and @skipkayhil did explore it too: [main...skipkayhil:rails:hm-qwysxtxtowwkxzqr](https://github.com/rails/rails/compare/main...skipkayhil:rails:hm-qwysxtxtowwkxzqr)
> 
> I told @skipkayhil I didn't like the idea of patching `Object#inspect`, and still think we shouldn't. However to avoid the numerous duplicated logic, and different output, we could have a module that you can include to get a Ruby 4.0 like `#inspect` method.
> 
> This it would look like:
> 
> ```ruby
>   include ActiveSupport::ObjectInspectBackport if RUBY_VERSION < "4"
>   private
>     def instance_variables_to_inspect
>      # ...
>     end
> ```
> 
> So less duplication.

@byroot Just addressed this in my last push, lmk if it does not align with what you were expecting.

activesupport/lib/active_support/inspect_backport.rb
```diff
@@ -0,0 +1,33 @@
...
+  #
+  #   class MyClass
+  #     include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
+  #
+  #     private
+  #       def instance_variables_to_inspect
+  #         [:@name, :@status].freeze
+  #       end
+  #   end
+  module InspectBackport
```
byroot:
```suggestion
  module InspectBackport # :nodoc:
```

I don't think we want to expose this as a public API.

activemodel/lib/active_model/error.rb
```diff
@@ -195,8 +196,10 @@ def hash # :nodoc:
       attributes_for_hash.hash
     end
 
-    def inspect # :nodoc:
-      "#<#{self.class.name} attribute=#{@attribute}, type=#{@type}, options=#{@options.inspect}>"
+    include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
+
+    private def instance_variables_to_inspect
```
byroot: we don't use `private def` in Rails (I think?) best to move it under the `private` section.

byroot (commented):
Just a few nitpicks, but I think it's mostly good to go at this point.

activemodel/lib/active_model/error.rb
mbast100: Oops yes typo.

actionpack/test/controller/parameters/accessors_test.rb
skipkayhil, line 402: Test failure is real, this is unused now

skipkayhil:
Rubocop is also failing for real

mbast100:
@skipkayhil yea just addressed the CI failures, will keep an eye on the run and address further if needed.

actionpack/lib/action_controller/metal/strong_parameters.rb
```diff
@@ -1056,9 +1057,7 @@ def stringify_keys # :nodoc:
       dup
     end
 
-    def inspect
-      "#<#{self.class} #{@parameters} permitted: #{@permitted}>"
```
matthewd, line 1060: This feels like quite an aesthetic downgrade on a frequently-inspected class

activesupport/lib/active_support/inspect_backport.rb
```diff
@@ -0,0 +1,33 @@
...
+  #
+  #   class MyClass
+  #     include ActiveSupport::InspectBackport if RUBY_VERSION < "4"
+  #
+  #     private
+  #       def instance_variables_to_inspect
+  #         [:@name, :@status].freeze
+  #       end
+  #   end
+  module InspectBackport # :nodoc:
```
matthewd, line 19:
```suggestion
  module InspectBackport # :nodoc:
    if RUBY_VERSION >= "4"
      def self.append_features(_)
      end
    end

```

... _maybe_? 🤔

activesupport/lib/active_support/inspect_backport.rb
```diff
@@ -0,0 +1,33 @@
...
+    def inspect
+      ivars = instance_variables_to_inspect
+      klass = self.class.name || self.class.inspect
+      addr = "%#018x" % (object_id << 1)
```
matthewd: Is this even sensible in current versions? At minimum it feels like the `18` is ugly and misleading.

activesupport/lib/active_support/inspect_backport.rb
byroot: What I had in mind was to not include the module at all, but just transfer the method with `define_method`. But maybe I'm thinking too much.

actionpack/lib/action_controller/metal/strong_parameters.rb
byroot: Yeah good catch, for this specific one the inspect wasn't just to hide something, we could definitely keep that one.

activesupport/lib/active_support/inspect_backport.rb
byroot: Well, the `object_id` is a lie in the sense that it's not the object address, but it's not like there's much you can do with the object address other than map it into `ObjectSpace.dump_all`.

What's important is that it help spot objects that are otherwise identical. That's the sort of thing I wasted hours debugging in the past.

actionpack/lib/action_controller/metal/strong_parameters.rb
mbast100: Not sure i follow here the behaviour is;
```
Old:  #<ActionController::Parameters {"name"=>"Francesco"} permitted: false>
New:  #<ActionController::Parameters:0x... @parameters={"name"=>"Francesco"},
  @permitted=false>
```
There still a `instance_variables_to_inspect` defined here;
https://github.com/rails/rails/pull/56782/changes#diff-99dab0dfb4d0cfa044997480d5aa1b100dc60347a10cedd6d8f7a0395f6a6efdR1153

This test should capture this change:
https://github.com/rails/rails/pull/56782/changes#diff-bcc7a55b0512ec147d80fc3fe0c69c203ac77fd5e32e73de1f644e89dd96dbb3

mbast100: I miss understood the comment here, sorry 🤦‍♂️

mbast100:
👋  @matthewd @byroot
- Changed from `"%#018x" % (object_id << 1)` to `"0x%x" % object_id` — keeps the `0x` prefix but no zero-padding, and uses `object_id` directly instead of the shifted value.
- Kept the custom inspect for the `ActionController::Parameters`

byroot:
Mostly look good, but the CI failure is legit:

```

Failure:
ApplicationTests::ConfigurationTest#test_config.action_controller.wrap_parameters_is_set_in_ActionController::Base [test/application/configuration_test.rb:1660]:
Expected /@parameters=.*"title" => "foo"/ to match # encoding: ASCII-8BIT
#    valid: true
"#<ActionController::Parameters {\"title\" => \"foo\"} permitted: false>".
```

activesupport/test/inspect_backport_test.rb
```diff
@@ -0,0 +1,71 @@
...
+      include ActiveSupport::InspectBackport
+
+      private
+        def instance_variables_to_inspect
+          [:@missing].freeze
+        end
+    end
+
+    obj = klass.new
+    assert_match(/@missing=nil/, obj.inspect)
```
byroot, line 56: Actually this isn't correct. On 4.0, missing variables aren't displayed at all.

byroot:
Cleaned up and merged as https://github.com/rails/rails/pull/56865

PR #56783, opened 2026-02-12
Fix ActiveStorage::Blob content type methods to handle nil

kudoas:
### Motivation / Background

This Pull Request fixes a `NoMethodError` that occurs when purging unattached Active Storage blobs.

When running `purge_later` to delete unattached records, the process fails if a record has `content_type: nil`. The specific error is:
`undefined method 'start_with?' for nil (NoMethodError)` inside the `image?` method.

This issue was caused by records with `content_type: nil` being inserted into the database. Specifically, this data was created by a vulnerability scanning tool directly sending requests to `/rails/active_storage/direct_uploads` without a content type.

Since it is possible for `content_type` to be nil, the code should handle this gracefully to ensure the cleanup process completes successfully.

### Detail

This Pull Request changes blob type methods to safely handle `nil` values in `content_type`.

It prevents the crash by checking for presence or using safe navigation before calling `start_with?`.

### Checklist

### DIFF
--- a/activestorage/app/models/active_storage/blob.rb
+++ b/activestorage/app/models/active_storage/blob.rb
@@ -195,22 +195,22 @@ class ActiveStorage::Blob < ActiveStorage::Record
 
   # Returns true if the content_type of this blob is in the image range, like image/png.
   def image?
-    content_type.start_with?("image")
+    content_type&.start_with?("image")
   end
 
   # Returns true if the content_type of this blob is in the audio range, like audio/mpeg.
   def audio?
-    content_type.start_with?("audio")
+    content_type&.start_with?("audio")
   end
 
   # Returns true if the content_type of this blob is in the video range, like video/mp4.
   def video?
-    content_type.start_with?("video")
+    content_type&.start_with?("video")
   end
 
   # Returns true if the content_type of this blob is in the text range, like text/plain.
   def text?
-    content_type.start_with?("text")
+    content_type&.start_with?("text")
   end
 
   # Returns the URL of the blob on the service. This returns a permanent URL for public files, and returns a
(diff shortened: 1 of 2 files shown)


activestorage/app/models/active_storage/blob.rb
```diff
@@ -195,22 +195,22 @@ def custom_metadata=(metadata)
 
   # Returns true if the content_type of this blob is in the image range, like image/png.
   def image?
-    content_type.start_with?("image")
+    content_type&.start_with?("image") || false
```
p8: There is no need to explictly return false. Falsey behaviour is pretty common in the Rails codebase.
```suggestion
    content_type&.start_with?("image")
```

activestorage/test/models/blob_test.rb
```diff
@@ -161,6 +161,21 @@ class ActiveStorage::BlobTest < ActiveSupport::TestCase
...
+  test "blob type methods return false for nil content type" do
+    blob = create_blob_before_direct_upload(
+      filename: "unknown_file",
+      byte_size: 100,
+      checksum: "test_checksum",
+      content_type: nil
+    )
+
+    assert_nil blob.content_type
+    assert_equal false, blob.image?
```
p8: You can use the assert_not_predicate instead:
```suggestion
    assert_not_predicate blob, :image?
```

activestorage/app/models/active_storage/blob.rb
kudoas: @p8 

Thanks! I removed the explicit return false.
Fixed in 846182dd9f22b7040d2e559d9697b48b1c5b56f9

activestorage/test/models/blob_test.rb
kudoas: @p8 Fixed this too. 846182dd9f22b7040d2e559d9697b48b1c5b56f9

PR #56791, opened 2026-02-13
Materialize transactions before instrumentation

djmb:
### Motivation / Background

I noticed that when running queries against a database in another datacenter, the timing for the first query in a transaction always looked like it required two round trips. 

Rather than there being two round trips there's a bug where BEGIN/SAVEPOINT queries were double counted when using lazy transaction materialization.

I think this behaviour dates back to https://github.com/rails/rails/commit/542f0951dddac49bf06f7da35d990db4f3829307 (cc @matthewd).

### Detail

When lazy transactions are enabled, the first query inside a transaction triggers BEGIN. This happened inside the log() instrumentation wrapper, so the first query's sql.active_record notification included both the BEGIN round trip and the actual query.

The BEGIN also fires its own notification, so its cost was reported twice.

Move materialization in execute_intent to before log(), so the BEGIN completes with its own notification before the query's instrumentation starts.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -605,14 +605,18 @@ module ActiveRecord
       # Final wrapper around the subclass-specific +perform_query+. Populates the calling
       # intent's raw_result.
       def execute_intent(intent) # :nodoc:
+        materialize_transactions if intent.materialize_transactions
+        should_dirty = intent.materialize_transactions
         log(intent) do |notification_payload|
           intent.notification_payload = notification_payload
-          with_raw_connection(allow_retry: intent.allow_retry, materialize_transactions: intent.materialize_transactions) do |conn|
+          with_raw_connection(allow_retry: intent.allow_retry, materialize_transactions: false) do |conn|
             result = perform_query(conn, intent)
             intent.raw_result = result
             handle_warnings(result, intent.processed_sql)
           end
         end
+      ensure
+        dirty_current_transaction if should_dirty
       end
 
       # Executes SQL statements in the context of this connection without
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -605,11 +605,19 @@ module ActiveRecord
       # Final wrapper around the subclass-specific +perform_query+. Populates the calling
       # intent's raw_result.
       def execute_intent(intent) # :nodoc:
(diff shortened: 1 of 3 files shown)


jeremy:
Found a behavior regression in the error path and pushed a fix.

**The issue**

`execute_intent` now calls `materialize_transactions` before `log()`, but `log()` eagerly evaluates `intent.processed_sql` and `intent.type_casted_binds` as keyword arguments to `instrumenter.instrument` before yielding ([abstract_adapter.rb:1218-1223](https://github.com/rails/rails/blob/ada7eafd138847b809ee1115dcf6423055431a91/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1217-L1232)). `processed_sql` calls `preprocess_query`, which can raise locally — e.g. `ReadOnlyError` from `while_preventing_writes` ([query_intent.rb:270-271](https://github.com/rails/rails/blob/ada7eafd138847b809ee1115dcf6423055431a91/activerecord/lib/active_record/connection_adapters/query_intent.rb#L269-L272)).

With the original PR's `should_dirty = intent.materialize_transactions` assigned before `log()`, a locally rejected write inside a lazy transaction would:
1. Send `BEGIN` to the database
2. Raise `ReadOnlyError` (no DB round trip for the query itself)
3. Execute the `ensure` with `should_dirty = true`, dirtying the transaction

This changes lazy transaction semantics: a write rejected entirely in Ruby can now force a `BEGIN` and mark the transaction non-restorable.

**The fix** ([121740d6](https://github.com/djmb/rails/blob/121740d6a86b972ac8cba58d21ad8ef6ad230a31/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L607-L628))

Two changes to `execute_intent`:

1. Evaluate `intent.processed_sql` and `intent.type_casted_binds` before calling `materialize_transactions`, so local rejections bail out before `BEGIN` is sent.
2. Move `should_dirty = true` inside the `with_raw_connection` block (right before `perform_query`), so dirtying only happens when a query actually reaches the connection — restoring the pre-PR semantics.

**Regression test** ([dc6b35bc](https://github.com/djmb/rails/blob/dc6b35bcd0e52f148cf025a75120e1501be7c953/activerecord/test/cases/transactions_test.rb#L1624-L1633))

```ruby
def test_locally_rejected_query_does_not_materialize_or_dirty_transaction
    connection = ActiveRecord::Base.lease_connection
    ActiveRecord::Base.while_preventing_writes do
      assert_raises(ActiveRecord::ReadOnlyError) { Topic.create!(title: "test") }
    end
    assert_not connection.current_transaction.materialized?
    assert_not connection.current_transaction.dirty?
  end
end
```

PR #56797, opened 2026-02-13
Include virtual? in Column#== and Column#hash for SQLite3 and PostgreSQL

uberjay:
### Summary

The `Deduplicable` module uses `==` and `hash` to maintain a registry of column instances. When a GENERATED (virtual) column was registered first, a regular column on a different table with the same name and type could be deduplicated to the generated instance. This caused the regular column to be silently excluded from INSERT and UPDATE statements, resulting in NULL values in the database.

This PR includes `virtual?` in both `==` and `hash` for `SQLite3::Column` and `PostgreSQL::Column` so columns with different generated_type values are never treated as identical.

### Changes

- **`SQLite3::Column`**: Add `virtual?` to `==` and `hash`
- **`PostgreSQL::Column`**: Add `virtual?` to `==` and `hash`
- Add tests for both adapters verifying that virtual columns are not deduplicated with regular columns

Fixes #56795

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/column.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/column.rb
@@ -69,7 +69,8 @@ module ActiveRecord
           other.is_a?(Column) &&
             super &&
             identity? == other.identity? &&
-            serial? == other.serial?
+            serial? == other.serial? &&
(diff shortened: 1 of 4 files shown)


byroot:
Thanks. Backported to 8-1 and 8-0 stable.

PR #56803, opened 2026-02-14
Fix in_batches use_ranges with limit producing negative LIMIT

daffo:
Fixes #56819.

### Motivation / Background

When `in_batches` is called with `use_ranges: true` on a relation that has a `.limit()`, a `PG::InvalidRowCountInLimitClause: ERROR: LIMIT must not be negative` (or equivalent) error is raised when the limit is not a multiple of the batch size.

### Detail

In the `use_ranges` code path, `values_size` is set to `batch_limit` (the original batch size) rather than the actual number of records. When `remaining < batch_limit`, the relation's limit is updated via `relation = relation.limit(remaining)`, but `batch_limit` stays at the original value.

On the next iteration, `remaining -= values_size` subtracts the stale `batch_limit` from `remaining`, causing it to go negative. This results in `relation.limit(-1)` which raises a database error.

### Fix

Update `batch_limit` alongside the relation's limit when `remaining < batch_limit`.

### Reproduction

```ruby
# With 11 Post records in the database:
Post.limit(5).in_batches(of: 3, use_ranges: true) { |batch| batch.count }
# => raises "LIMIT must not be negative"
```

### Checklist

- [x] Added corresponding tests
- [x] All existing tests pass (`116 runs, 836 assertions, 0 failures, 0 errors`)

### DIFF
--- a/activerecord/lib/active_record/relation/batches.rb
+++ b/activerecord/lib/active_record/relation/batches.rb
@@ -441,8 +441,8 @@ module ActiveRecord
             yielded_relation.load_records(records)
           elsif (empty_scope && use_ranges != false) || use_ranges
             # Efficiently peak at the last value for the next batch using offset and limit.
-            values_size = batch_limit
-            values_last = batch_relation.offset(batch_limit - 1).pick(*cursor)
+            values_size = remaining ? [batch_limit, remaining].min : batch_limit
+            values_last = batch_relation.offset(values_size - 1).pick(*cursor)
 
             # If the last value is not found using offset, there is at most one more batch of size < batch_limit.
             # Retry by getting the whole list of remaining values so that we have the exact size and last value.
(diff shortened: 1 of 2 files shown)


fatkodima:
Can you remove vertical alignment in the tests and ai from the coauthors?

daffo:
> Can you remove vertical alignment in the tests and ai from the coauthors?

should I remove it in all the limit tests or only in my new one?

edit: also I'm actually noticing now, is it worth testing `use_ranges` `[true,false]` on top of `load` `[true,false]` ?
like:
```
[true, false].each do |load|
  [true, false].each do |use_ranges|
```

fatkodima:
Remove only in the added tests. I think only one of `use_ranges` or `load` is enough, no need for different combinations.

daffo:
> Remove only in the added tests. I think only one of `use_ranges` or `load` is enough, no need for different combinations.

kk we good then.

daffo:
I can see CI is failing, but this `ConnectionPoolThreadTest#test_idle_through_keepalive` is unrelated to my change

PR #56805, opened 2026-02-14
Optimize calculating remote IP address

fatkodima:
From production wall/cpu profiling, I noticed that in our case just getting a client's IP uses a couple % of the whole request runtime. We use `rack-attack`, which access the calculated IP for its logic. 

And it was also at the top few lines of the allocations profiling.

Benchmark (old vs new):
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem 'rails', github: "rails/rails"
  gem "benchmark-ips"
end

require 'benchmark/ips'

class OptimizedGetIp < ActionDispatch::RemoteIp::GetIp
  # same method with the last line changed
  def calculate_ip
    # Set by the Rack web server, this is a single value.
    remote_addr = sanitize_ips(ips_from(@req.remote_addr)).last

    # Could be a CSV list and/or repeated headers that were concatenated.
    client_ips    = sanitize_ips(ips_from(@req.client_ip)).reverse!
    forwarded_ips = sanitize_ips(@req.forwarded_for || []).reverse!

    # `Client-Ip` and `X-Forwarded-For` should not, generally, both be set. If they
    # are both set, it means that either:
    #
    # 1) This request passed through two proxies with incompatible IP header
    #     conventions.
    #
    # 2) The client passed one of `Client-Ip` or `X-Forwarded-For`
    #     (whichever the proxy servers weren't using) themselves.
    #
    # Either way, there is no way for us to determine which header is the right one
    # after the fact. Since we have no idea, if we are concerned about IP spoofing
    # we need to give up and explode. (If you're not concerned about IP spoofing you
    # can turn the `ip_spoofing_check` option off.)
    should_check_ip = @check_ip && client_ips.last && forwarded_ips.last
    if should_check_ip && !forwarded_ips.include?(client_ips.last)
      # We don't know which came from the proxy, and which from the user
      raise IpSpoofAttackError, "IP spoofing attack?! " \
        "HTTP_CLIENT_IP=#{@req.client_ip.inspect} " \
        "HTTP_X_FORWARDED_FOR=#{@req.x_forwarded_for.inspect}" \
        " HTTP_FORWARDED=" + @req.forwarded_for.map { "for=#{_1}" }.join(", ").inspect if @req.forwarded_for.any?
    end

    # We assume these things about the IP headers:
    #
    #     - X-Forwarded-For will be a list of IPs, one per proxy, or blank
    #     - Client-Ip is propagated from the outermost proxy, or is blank
    #     - REMOTE_ADDR will be the IP that made the request to Rack
    ips = forwarded_ips + client_ips
    ips.compact!

    # If every single IP option is in the trusted list, return the IP that's
    # furthest away
    first_non_proxy(ips + [remote_addr]) || ips.last || remote_addr
  end

  def first_non_proxy(ips)
    ips.find do |ip|
      return unless ip

      ip = IPAddr.new(ip) if @proxies.all?(IPAddr)
      @proxies.none? { |proxy| proxy === ip }
    end
  end
end

env = Rack::MockRequest.env_for("/").merge!(
  "REMOTE_ADDR" => "1.1.1.1",
  "HTTP_X_FORWARDED_FOR" => "4.2.42.42"
)

req = ActionDispatch::Request.new(env)

Benchmark.ips do |x|
  x.report("original") do
    ActionDispatch::RemoteIp::GetIp.new(req, true, ActionDispatch::RemoteIp::TRUSTED_PROXIES).calculate_ip
  end

  x.report("optimized") do
    OptimizedGetIp.new(req, true, ActionDispatch::RemoteIp::TRUSTED_PROXIES).calculate_ip
  end

  x.compare!
end
```

Results
```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [x86_64-darwin21]
            original     1.033k i/100ms
           optimized     3.224k i/100ms
            original     10.638k (± 7.9%) i/s   (94.00 μs/i) -     53.716k in   5.082355s
           optimized     33.102k (± 6.1%) i/s   (30.21 μs/i) -    167.648k in   5.083553s

Comparison:
           optimized:    33102.0 i/s
            original:    10638.0 i/s - 3.11x  slower
```

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/remote_ip.rb
+++ b/actionpack/lib/action_dispatch/middleware/remote_ip.rb
@@ -163,10 +163,11 @@ module ActionDispatch
         #     - REMOTE_ADDR will be the IP that made the request to Rack
         ips = forwarded_ips + client_ips
         ips.compact!
+        ips << remote_addr
 
         # If every single IP option is in the trusted list, return the IP that's
         # furthest away
-        filter_proxies(ips + [remote_addr]).first || ips.last || remote_addr
+        first_non_proxy(ips) || ips[-2] || ips.last
       end
 
       # Memoizes the value returned by #calculate_ip and returns it for
@@ -194,9 +195,18 @@ module ActionDispatch
         ips
       end
 
-      def filter_proxies(ips) # :doc:
-        ips.reject do |ip|
-          @proxies.any? { |proxy| proxy === ip }
+      def first_non_proxy(ips) # :doc:
+        ips.find do |raw_ip|
+          return unless raw_ip
+
+          ip = IPAddr.new(raw_ip)
+          @proxies.none? do |proxy|
+            if proxy.is_a?(IPAddr)
+              proxy.include?(ip)
+            else
+              proxy === raw_ip
+            end
+          end
         end
       end
     end


actionpack/lib/action_dispatch/middleware/remote_ip.rb
```diff
@@ -194,9 +194,12 @@ def sanitize_ips(ips) # :doc:
         ips
       end
 
-      def filter_proxies(ips) # :doc:
-        ips.reject do |ip|
-          @proxies.any? { |proxy| proxy === ip }
+      def first_non_proxy(ips) # :doc:
+        ips.find do |ip|
```
skipkayhil: Would it make sense to lift the iteration out of this method? I think we could avoid two or three array allocations:

```
forwarded_ips.find { ip_unless_proxy(it) } || client_ips.find { ip_unless_proxy(it) } || ip_unless_proxy(remote_addr) || ips.last || remote_addr
```

fatkodima: Got a slight speedup, but a bit more uglier code
```
Comparison:
           optimized:    33350.7 i/s
            original:    10549.1 i/s - 3.16x  slower
```

PR #56807, opened 2026-02-14
Rails.cache.read: add support to new option delete: true

glaucocustodio:
### Motivation / Background

Add support for atomically reading and deleting a cache entry in `RedisCacheStore` using the Redis [GETDEL](https://redis.io/docs/latest/commands/getdel/) command. This is useful for one-time-read patterns where you want to guarantee the value can only be consumed once, without a race condition between a separate read and delete.

### Detail

This PR changes `RedisCacheStore#read_serialized_entry` to accept a new `delete: true` option. When passed, the underlying Redis command switches from `GET` to `GETDEL`, which atomically returns the value and removes the key in a single round trip.
Three areas were modified:

- `activesupport/lib/active_support/cache/redis_cache_store.rb`: `read_serialized_entry` now accepts a `delete` keyword argument. When true, it calls `c.getdel(key)` instead of `c.get(key)`. Documentation was added to the class-level RDoc.
- `activesupport/lib/active_support/cache/strategy/local_cache.rb`: When `delete: true`, the local cache layer bypasses its in-memory cache (to ensure the `GETDEL` always reaches Redis) and clears the local cache entry afterward to keep both layers consistent.
- `activesupport/test/cache/stores/redis_cache_store_test.rb`: new tests covering: basic read-and-delete, missing keys, raw values, expired entries, Redis key removal verification, and local cache interaction (clearing, bypassing stale entries).

This is a Redis-specific option. Other cache stores silently ignore it (the option falls through `**options` splat).

### Additional information

- Requires Redis 6.2+ (when `GETDEL` was introduced). Should we care about previous versions of Redis?

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,17 @@
+*   Add `delete: true` option to `Rails.cache.read` for atomic read-and-delete (only supported by Redis cache store).
+
+    Uses the Redis [GETDEL](https://redis.io/docs/latest/commands/getdel/) command to atomically return a cached value and remove
+    it in a single operation. Useful for single-use values like OTP codes or
+    one-time tokens.
+
+    ```ruby
+    Rails.cache.write("otp", "123456")
+    Rails.cache.read("otp", delete: true)  # => "123456"
+    Rails.cache.read("otp")                # => nil
+    ```
+
+    *Glauco Custodio*
+
 *   Introduce `ActiveSupport::TestCase.around`
 
     Add a callback, which runs between `TestCase#setup` and `TestCase#teardown`.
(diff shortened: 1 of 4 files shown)


skipkayhil:
Some previous discussion here #56553

Is there a reason to not use Redis directly in this case?

glaucocustodio:
Oh, I was not aware of https://github.com/rails/rails/pull/56553. That said, I believe my proposed approach may integrate more naturally, as it avoids extending the Cache API and instead builds on the existing `read` method, which is widely used - pushing up a big tent.

>Having a big tent doesn’t mean trying to be all things to all people, though. It just means you welcome all people to your party, and allow them to bring their own drinks. We need to lose none of our soul or values by offering others to join us, and we may well learn how to mix a new delicious drink or two

https://rubyonrails.org/doctrine#big-tent

For most use cases it won't make a difference having this new option, but for a few cases it really counts since one can fully leverage  `Rails.cache` instead of depending on a bespoke, lower lever implementation using the Redis API directly.

rafaelfranca (approved):

PR #56818, opened 2026-02-16
Check for `blank?` instead of `nil?` in CombinedConfiguration

jordan-brough:
### Motivation / Background

This Pull Request is extracted from [a discussion in #56430](https://github.com/rails/rails/pull/56430#discussion_r2640480995) with @dhh:

> I don't think nil should be respected, no. We should search through all the configurations for a
> not-nil value. If we can't find one, we raise MissingKey.

> When you do #require, we should blow up **if nil or blank**

### Detail

This Pull Request updates CombinedConfiguration to treat all `blank?` values as missing in configuration lookups, instead of just `nil?` values.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/encrypted_configuration.rb
+++ b/activesupport/lib/active_support/encrypted_configuration.rb
@@ -60,7 +60,7 @@ module ActiveSupport
     end
 
     # Find singular or nested keys.
-    # Raises +KeyError+ if not found or blank.
+    # Raises +KeyError+ if not found or +blank?+ (except +false+).
     #
     # Given configuration:
     #   db_port: ""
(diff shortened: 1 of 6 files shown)


activesupport/lib/active_support/encrypted_configuration.rb
```diff
@@ -59,52 +60,42 @@ def initialize(config_path:, key_path:, env_key:, raise_if_missing_key:)
...
-      elsif default.respond_to?(:call)
-        default.call
-      else
-        default
-      end
+      dig(*key).presence || if default.respond_to?(:call)
+                              default.call
+                            else
+                              default
+                            end
```
MatheusRich: I haven't tested it, but would this work?
```suggestion
      dig(*key).presence || default.try(:call) || default
```

jordan-brough: If `default` is a callable object, and `call` returns a falsy value then that will end up returning the callable object.

activesupport/test/combined_configuration_test.rb
```diff
@@ -115,14 +115,6 @@ class CombinedConfigurationTest < ActiveSupport::TestCase
     assert_equal "dotenv", @combined.require(:nested, :available_in_dotenv_and_cred)
   end
 
-  test "require key with a false value" do
-    assert_equal false, @combined.require(:false)
-  end
-
-  test "option key with a false value" do
-    assert_equal false, @combined.option(:false)
-  end
```
dhh, line 124: Can't lose the ability to set values to false.

jordan-brough: @dhh I've updated to allow `false`.

Anything else `blank?` will be considered ,"not set".
E.g.:
- blank strings
- nil
- empty arrays
- empty hashes

Lmk if we want to change that. Maybe we should only skip nil and blank strings.

PR #56821, opened 2026-02-16
Support html-safe error messages in Active Model

MatheusRich:
### Motivation / Background

I needed to write a helper that would create a HTML element for error messages for a specific attribute in Rails. Something like

```rb
def error_messages_for_attribute(model, attribute)
  if model.errors.include?(attribute)
    messages = model.errors.where(attribute).map(&:full_message)

    tag.span(to_sentence(messages), class: "error-message")
  end
end
```

The problem is that currently error messages are never marked as HTML-safe, even when using the `_html`s suffix [like Action View supports][av], so raw HTML is outputted instead.

[av]: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

### Detail

Validation error message keys ending in `_html` now have their interpolation values HTML-escaped and the resulting message marked as `html_safe`, consistent with the [existing convention][convention] in Action View and Action Controller translation helpers.

```yaml
en:
  errors:
    messages:
      taken_html: "has <em>already</em> been taken"
```

```ruby
model.errors.add(:name, :taken_html)
model.errors.messages.first.html_safe? # => true
```

[convention]: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

### Checklist

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,24 @@
+*   Support html-safe error messages in Active Model.
+
+    Validation error message keys ending in `_html` now have their
+    interpolation values HTML-escaped and the resulting message marked as
+    `html_safe`, consistent with the existing convention in Action View
+    and Action Controller translation helpers.
+
+    ```yaml
+    en:
+      errors:
+        messages:
+          taken_html: "has <em>already</em> been taken"
+    ```
+
+    ```ruby
+    model.errors.add(:name, :taken_html)
+    model.errors.messages.first.html_safe? # => true
+    ```
+
+    *Matheus Richard*
+
 *   Add `has_json` and `has_delegated_json` to provide schema-enforced access to JSON attributes.
 
     ```ruby
--- a/activemodel/lib/active_model/error.rb
+++ b/activemodel/lib/active_model/error.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "active_support/core_ext/string/output_safety"
+require "active_support/html_safe_translation"
 
 module ActiveModel
   # = Active \Model \Error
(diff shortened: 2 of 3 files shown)


seanpdoyle:
Most of the built-in `validates_*_of` style validation declarations have at least one default key-to-message mapping (see [4.5.2 Error Message Interpolation](https://guides.rubyonrails.org/i18n.html#error-message-interpolation) in the I18n Guides).

If treating `_html`-suffixed keys as HTML safe string is deemed to be viable, I wonder if there is a way for Active Model to automatically infer HTML safety based on the presence of a user land-defined I18n key that ends with `_html`.

To expand upon the example shared in the commit message, consider that a [validates_uniqueness_of](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of) declaration will look for a `errors.messages.taken` key (along with a variety of other cascading lookups ending in `.taken`). Could the presence of an `errors.messages.taken_html` translation result in an HTML safe message *without* the need for an explicit call to `model.errors.add(:name, :taken_html)`?

This inference runs the risk of introducing a cross-site scripting injection vulnerability for validations that inject the invalid value through I18n interpolation. Revisiting the table mentioned above by [4.5.2 Error Message Interpolation](https://guides.rubyonrails.org/i18n.html#error-message-interpolation), all of the built-in validations either *do not* interpolate a value, or *do* interpolate either `%{attribute}` (the Ruby-defined attribute name) or `%{count}` (a Ruby-determined integer). 

However, automatic `_html` inference could pose a true risk for any custom validation that decide to interpolate any other kind of user-provided value. Maybe an allow-list limited automatic inference could be a balanced trade-off?

rafaelfranca:
It doesn't seems right to active model to know about HTML since it is just one of the ways to present an error. I think something like this would need to be implemented in the Action View layer, not active model. Maybe Active Model can provide callbacks that allow Action View to implement this, but it should not have ERB or HTML mentions on it. Closing since this isn't directionally correct.

Edgarruiz8585:
>  @Edgarruiz8585
🚀

PR #56822, opened 2026-02-16
More railties test balancing

skipkayhil:
I don't see many ~5 min times for the db tests in CI. We've had
issues where a single test will hang and blow up test time, so maybe
that's what was being seen before?

More importantly, this adjusts the time for configuration_test and
console_test way up so that they aren't packed with so many other tests,
and adds sprockets_assets_test/query_logs_test since they consistently
take > 30s and are currently estimated at the default, 1.


railties/Rakefile
```diff
@@ -17,18 +17,18 @@ task test: "test:isolated"
...
+      "test/application/configuration_test.rb" => 200,
       "test/generators/app_generator_test.rb" => 150,
-      "test/application/rake/dbs_test.rb" => 300,
-      "test/application/rake/dbs_schema_test.rb" => 500,
-      "test/application/rake/dbs_setup_test.rb" => 350,
-      "test/application/rake/dbs_postgresql_test.rb" => 150,
+      "test/application/rake/dbs_test.rb" => 20,
+      "test/application/rake/dbs_schema_test.rb" => 45,
+      "test/application/rake/dbs_setup_test.rb" => 45,
+      "test/application/rake/dbs_postgresql_test.rb" => 15,
```
skipkayhil, line 31: cc @jeremy because you set these numbers initially

PR #56827, opened 2026-02-17
Optimize `RemoteIp#filter_proxies`

byroot:
In the common case, `@proxies` is a list of `IPAddr` instances and `ips` a list of `String`.

`IPAddr#===` does cast the rhs into IPAddr which means we end up reparsing the same IP addreses again and again.

It's preferable to pre-cast them once.

```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +YJIT +PRISM [arm64-darwin25]
            original     1.869k i/100ms
                 opt     7.945k i/100ms
            original     18.556k (± 1.5%) i/s   (53.89 μs/i) -     93.450k in   5.037425s
                 opt     78.641k (± 0.9%) i/s   (12.72 μs/i) -    397.250k in   5.051837s

Comparison:
            original:    18555.6 i/s
                 opt:    78640.8 i/s - 4.24x  faster
```

```ruby
# frozen_string_literal: true

require "bundler/inline"
require "ipaddr"

gemfile do
  gem "benchmark-ips"
end

TRUSTED_PROXIES = [
  "127.0.0.0/8",    # localhost IPv4 range, per RFC-3330
  "::1",            # localhost IPv6
  "fc00::/7",       # private IPv6 range fc00::/7
  "10.0.0.0/8",     # private IPv4 range 10.x.x.x
  "172.16.0.0/12",  # private IPv4 range 172.16.0.0 .. 172.31.255.255
  "192.168.0.0/16", # private IPv4 range 192.168.x.x
  "169.254.0.0/16", # link-local IPv4 range 169.254.x.x
  "fe80::/10",      # link-local IPv6 range fe80::/10
].map { |proxy| IPAddr.new(proxy) }

def original(ips)
  ips.reject do |ip|
    TRUSTED_PROXIES.any? { |proxy| proxy === ip }
  end
end

def opt(ips)
  ips.reject do |raw_ip|
    next true if raw_ip.nil?
    ip = IPAddr.new(raw_ip)
    TRUSTED_PROXIES.any? do |proxy|
      if proxy.is_a?(IPAddr)
        proxy.include?(ip)
      else
        proxy === raw_ip
      end
    end
  end
end

ips = [
  "10.0.0.1",
  "10.0.0.1",
  "3.4.5.6",
  nil,
  "fc01::",
  "fc00::",
  "::1",
  "2001:0db8:85a3:0000:0000:8a2e:0370:7335",
  "2001:0db8:85a3:0000:0000:8a2e:0370:7334",
].freeze

Benchmark.ips do |x|
  x.report("original") do
    original(ips)
  end

  x.report("opt") do
    opt(ips)
  end

  x.compare!(order: :baseline)
end
```

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/remote_ip.rb
+++ b/actionpack/lib/action_dispatch/middleware/remote_ip.rb
@@ -70,7 +70,7 @@ module ActionDispatch
       @proxies = if custom_proxies.blank?
         TRUSTED_PROXIES
       elsif custom_proxies.respond_to?(:any?)
-        custom_proxies
+        custom_proxies.map { |p| p.is_a?(String) ? IPAddr.new(p) : p }
       else
         raise(ArgumentError, <<~EOM)
           Setting config.action_dispatch.trusted_proxies to a single value isn't
@@ -195,8 +195,16 @@ module ActionDispatch
       end
 
       def filter_proxies(ips) # :doc:
-        ips.reject do |ip|
-          @proxies.any? { |proxy| proxy === ip }
+        ips.reject do |raw_ip|
+          next true if raw_ip.nil?
+          ip = IPAddr.new(raw_ip)
+          @proxies.any? do |proxy|
+            if proxy.is_a?(IPAddr)
+              proxy.include?(ip)
+            else
+              proxy === raw_ip
+            end
+          end
         end
       end
     end


skipkayhil:
Looks pretty similar to https://github.com/rails/rails/pull/56805

Another thought I've had about this class is we could try replacing some/all of the `TRUSTED_PROXY` list with `IPAddr` methods, `loopback?`, `private?`, and `link_local?`, and then we'd only ever have to iterate for an application's configured trusted proxy list.

byroot:
> Looks pretty similar to #56805

Lol, indeed. cc @fatkodima

fatkodima:
Haha, really! 😄 

I tested the changes in this PR against my benchmark and got these results
```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [x86_64-darwin21]
            original   995.000 i/100ms
      optimized dima     3.239k i/100ms
      optimized jean     2.575k i/100ms
            original     10.321k (± 8.4%) i/s   (96.89 μs/i) -     51.740k in   5.050026s
      optimized dima     32.651k (± 7.0%) i/s   (30.63 μs/i) -    165.189k in   5.084311s
      optimized jean     24.459k (±14.9%) i/s   (40.89 μs/i) -    121.025k in   5.077950s

Comparison:
      optimized dima:    32651.5 i/s
      optimized jean:    24458.8 i/s - 1.33x  slower
            original:    10320.6 i/s - 3.16x  slower
```

byroot:
I think our patches can be combined. Using `find` helps a lot, but I think your call to `@proxies.all?(IPAddr)` is worse for perf than what I'm doing here.

byroot:
Closing in favor of https://github.com/rails/rails/pull/56827 (I merged some small changes in it)

PR #56828, opened 2026-02-17
fix: breaking support of hyphen in table names 

djezzzl:
### Motivation / Background

Hi, during our upgrade from Rails 7.0 to Rails 8.0 (and even Rails 8.1 and greater, though with different behavior), we noticed a **breaking change** in ActiveRecord.

The following code should reproduce the broken flow:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 8.0.0"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  create_table "table-with-hyphen", force: true do |t|
  end
end

  def test_insert_exec
    assert_nothing_raised do
      ActiveRecord::Base.connection.exec_insert("INSERT INTO \"table-with-hyphen\" DEFAULT VALUES")
    end
  end
end
```

### Detail

In this Pull Request, we add support for hyphen `-` in table names if quoted properly. 

### Additional information

Adding the following code to the example script makes it green.
Related commit: https://github.com/rails/rails/commit/9bbbf4b8b9bf3d7cb31c0425a6646cec49e4bc6c

Previously, `exec_insert` didn't try to derive private keys from the schema via parsing the table name from the SQL query. 

```ruby
# Rails 8.0 throws an error as primary_key for unknown table_ref raises an error
pk = primary_key(table_ref) if table_ref
```

I see that in main, we try to hit the schema cache instead, and return an empty private key, which is okay, though the returning statement is missing from the final SQL when it is expected to be there with a proper table name.   

```ruby
# in main branch, we access the cache, which is not throwning error when the table name is missing
pk = schema_cache.primary_keys(table_ref) if table_ref
```

```ruby
module ActiveRecord
  module ConnectionAdapters
    module DatabaseStatements
        if sql =~ /into\s("[-A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
          $1.delete('"').strip
        end
      end
    end
  end
end
```

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -773,7 +773,7 @@ module ActiveRecord
         end
 
         def extract_table_ref_from_insert_sql(sql)
-          if sql =~ /into\s("[A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
+          if sql =~ /into\s("[-A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
             $1.delete('"').strip
           end
         end
--- a/activerecord/test/cases/database_statements_test.rb
+++ b/activerecord/test/cases/database_statements_test.rb
@@ -22,6 +22,11 @@ class DatabaseStatementsTest < ActiveRecord::TestCase
     assert_not_nil return_the_inserted_id(method: :create)
   end
 
+  def test_extract_table_ref_from_insert_sql_with_hyphen_in_table_name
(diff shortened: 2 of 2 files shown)


djezzzl:
Hi @fractaledmind @adrianna-chang-shopify @nvasilevski @casperisfine,

As someone involved in the mentioned changes, may I have your assistance with the fix, please?

djezzzl:
Hi @byroot, thank you for releasing this and for the backport! 

I was a bit shy, so I didn't touch a changelog with this 😄

byroot:
It's better not to add the changelog yourself for fixes that need to be backported, as we only put a changelog for stable branches.

djezzzl:
Hi there, I have opened a new PR to the same problem with more support: https://github.com/rails/rails/pull/57012

PR #56829, opened 2026-02-17
Fix `has_many_inversing` for associations with composite foreign keys

kirs:
`Association#matches_foreign_key?` passes `reflection.foreign_key` directly to `read_attribute`. When the foreign key is an array, `read_attribute` returns nil, so `inversable?` returns false and identity preservation is broken for composite foreign keys.

Wrap the foreign key with `Array()` and map over each component, matching the pattern already used by `foreign_key_for?`.

A small repro script:

```ruby
  create_table :blog_posts, force: true do |t|
    t.integer :blog_id, null: false
  end

    t.integer :blog_id, null: false
    t.integer :blog_post_id, null: false
  end
end

ActiveRecord::Base.has_many_inversing = true

class BlogPost < ActiveRecord::Base
  has_many :comments,
    primary_key: [:blog_id, :id],
    foreign_key: [:blog_id, :blog_post_id],
    inverse_of: :blog_post
end

  belongs_to :blog_post,
    foreign_key: [:blog_id, :blog_post_id],
    primary_key: [:blog_id, :id],
    inverse_of: :comments
end

blog_post = BlogPost.create!(blog_id: 1)
Comment.create!(blog_id: 1, blog_post_id: blog_post.id)

comment = Comment.first
loaded_post = comment.blog_post
found = loaded_post.comments.detect { |c| c.id == comment.id }

if comment.equal?(found)
  puts "PASS: same Ruby object (inverse_of working)"
else
  puts "FAIL: different Ruby objects (inverse_of broken)"
end
```

@eileencodes @matthewd

### DIFF
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -409,11 +409,12 @@ module ActiveRecord
         end
 
         def matches_foreign_key?(record)
+          fk = Array(reflection.foreign_key)
           if foreign_key_for?(record)
-            record.read_attribute(reflection.foreign_key) == owner.id ||
-              (foreign_key_for?(owner) && owner.read_attribute(reflection.foreign_key) == record.id)
+            fk.map { |k| record.read_attribute(k) } == Array(owner.id) ||
+              (foreign_key_for?(owner) && fk.map { |k| owner.read_attribute(k) } == Array(record.id))
           else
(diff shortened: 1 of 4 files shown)


nvasilevski (approved):

masom:
Related: https://github.com/rails/rails/pull/56664

PR #56837, opened 2026-02-18
Don't filter StructuredEventSubscriber payloads

djmb:
The EventReporter applies ParameterFilter to hash payloads before emitting them to subscribers. When framework log subscribers were updated to consume events through the EventReporter (54264bde), this caused SQL query names and other internal metadata to be filtered.

For example, `config.filter_parameters += [:name]` turns SQL log lines into:

```
[FILTERED] (0.4ms)  SELECT "people".* FROM "people" WHERE ...
```

instead of:

```
Person Load (0.4ms)  SELECT "people".* FROM "people" WHERE ...
```

Add a `filter:` option to `EventReporter#notify` and `#debug`, defaulting to true for backwards compatibility. `StructuredEventSubscriber` passes `filter: false` from `emit_event` and `emit_debug_event`, so all framework subscribers are covered without per-subscriber changes.

### DIFF
--- a/actionview/lib/action_view/structured_event_subscriber.rb
+++ b/actionview/lib/action_view/structured_event_subscriber.rb
@@ -75,6 +75,7 @@ module ActionView
 
       def start(name, id, payload)
         ActiveSupport.event_reporter.debug("action_view.render_start",
+          filter_payload: false,
           is_layout: name == "render_layout.action_view",
           identifier: from_rails_root(payload[:identifier]),
           layout: from_rails_root(payload[:layout]),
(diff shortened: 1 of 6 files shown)


jeremy (changes requested):
Approach looks right: filtering was added (#55489) before framework subscribers moved onto EventReporter (#55900), so they were never meant to be filtered; this restores that boundary.

But I'm not crystal clear on intent, either. /cc @gmcgibbon @zzak 

Note: `ActionView::StructuredEventSubscriber::Start#start` calls `ActiveSupport.event_reporter.debug(...)` directly so it needs the same patch.

djmb:
@jeremy - I've pushed a change to avoid filtering in `ActionView::StructuredEventSubscriber::Start#start` too

PR #56841, opened 2026-02-18
Remove dash for rails new in devcontainers guide [ci skip]

Set27:
`rails-new` command does not exist, and we can find it only in this guide.
I guess this is a typo

### DIFF
--- a/guides/source/getting_started_with_devcontainer.md
+++ b/guides/source/getting_started_with_devcontainer.md
@@ -5,7 +5,7 @@ Getting Started with Dev Containers
 
 After reading this guide, you will know:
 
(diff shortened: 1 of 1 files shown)


yahonda:
Hi, `rails-new` is the correct command name — it's a standalone executable at https://github.com/rails/rails-new. 

Closing this pull request.

Set27:
> Hi, `rails-new` is the correct command name — it's a standalone executable at https://github.com/rails/rails-new.
> 
> Closing this pull request.

Do you think it's worth mentioning in the docs that such a program exists? I mean, it is not obvious from the beginning.

PR #56842, opened 2026-02-18
Revert: Sort table columns by name when dumping schema

bert-mccutchen:
### Motivation / Background

This Pull Request has been created for the Rails core team to review reverting sorted schema columns.

This is based on the findings outlined in discussion on https://github.com/rails/rails/pull/55414#issuecomment-3901961707.

Based on my findings, the original approach of sorting schema columns is not the appropriate solution to the problem the original developer was facing.

### Detail

- Reverted all code changes made in #53281.
- Updated the ActiveRecord CHANGELOG.

### Additional information

N/A

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Revert alphabetical sorting of table columns inside `schema.rb`.
+
+    Alphabetical sorting of table columns inside the schema creates improper production tables when using `db:prepare`.
+
+    *Bert McCutchen*
+
 *   Deprecated `ActiveRecord::ConnectionAdapters::Column#auto_populated?` in favor of
     `auto_populated_on_insert?`
 
--- a/activerecord/lib/active_record/schema_dumper.rb
+++ b/activerecord/lib/active_record/schema_dumper.rb
@@ -192,7 +192,7 @@ module ActiveRecord
           tbl.puts ", force: :cascade do |t|"
 
           # then dump all non-primary key columns
-          columns.sort_by(&:name).each do |column|
+          columns.each do |column|
             raise StandardError, "Unknown type '#{column.sql_type}' for column '#{column.name}'" unless @connection.valid_type?(column.type)
             next if column.name == pk
 
--- a/activerecord/test/cases/schema_dumper_test.rb
+++ b/activerecord/test/cases/schema_dumper_test.rb
@@ -167,14 +167,6 @@ class SchemaDumperTest < ActiveRecord::TestCase
     assert_no_match %r{create_table "ar_internal_metadata"}, output
   end
 
-  def test_table_columns_sorted
-    column_names = column_definition_lines(dump_table_schema("companies")).flatten.filter_map do |line|
-      $1 if line !~ /t\.index/ && line.match(/t\..*"(\w+)"/)
-    end
-
-    assert_equal %w[account_id client_of description firm_id firm_name name rating status type], column_names
-  end
-
   def test_schema_dumps_index_columns_in_right_order
     index_definition = dump_table_schema("companies").split(/\n/).grep(/t\.index.*company_index/).first.strip
     if current_adapter?(:Mysql2Adapter, :TrilogyAdapter)


pixeltrix:
I agree with reverting the change to sorting the columns when dumping for the following reasons:

1. What was an occasional issue with a developer's local database not matching the order of the production database is now a issue of every developer's local database not matching each other developer's local database as well as production. To avoid drift between developers' databases they'll need to recreate and reseed them, making migrations effectively a production only feature. This may make sense for AI agentic workflows where you are creating and seeding databases constantly when spinning up agents but it's not great for humans who may be maintaining local data for a feature they're working on and need to merge in the work of other developers.

2. It breaks workflows where column order is significant such as loading bulk data from CSVs. For example we use `PG::Connection#put_copy_data` to load data updates from CSVs for both development and production environments and now have to put workarounds in to ensure correct ordering.

zzeligg:
I see this has been reverted.

Many thanks, this saves me a LOT of hassle. :+1:

mackuba:
I don't think it was merged, there was just some rebase in the PR or something?

zzeligg:
> I don't think it was merged, there was just some rebase in the PR or something?

You are absolutely right. My mistake.

I didn't look carefully (misread the PR Title change).

Then I am still awfully affected :-/

bert-mccutchen:
> I don't think it was merged, there was just some rebase in the PR or something?

Yeah that was just me following these instructions:
https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#update-your-branch

mackuba:
Yeah the GitHub notification email confused me too tbh

flanger001:
Bumping to keep this open.

jagthedrummer:
I'd love to see this get merged. In the Bullet Train project we use the `columns_hash` to try to figure out the most likely candidate for a `label_string` for a given record by [looking for the first `string` column defined in `columns_hash`](https://github.com/bullet-train-co/bullet_train-core/blob/a455eb734887ca5e6efefbb208e635be196b6b50/bullet_train/app/models/concerns/records/base.rb#L42-L45).

The change in #53281 changed the behavior of `columns_hash` and now we don't have a way to determine which `string` column was defined first.

jagthedrummer:
~Another reason to revert is that this causes problems related to database indexes which rely heavily on "leftmost column" optimizations. https://github.com/rails/rails/pull/53281#issuecomment-4269378088~ - Scratch that. My bad. Column order in the _index_ is important, but not in the _table_.

skipkayhil:
> Another reason to revert is that this causes problems related to database indexes which rely heavily on "leftmost column" optimizations. [#53281 (comment)](https://github.com/rails/rails/pull/53281#issuecomment-4269378088)

I don't think the order of columns within the table has any effect on indexes. If the raw rows are being filtered on then you're already doing a table scan and aren't using an index.

> I'd love to see this get merged. In the Bullet Train project we use the columns_hash to try to figure out the most likely candidate for a label_string for a given record by [looking for the first string column defined in columns_hash](https://github.com/bullet-train-co/bullet_train-core/blob/a455eb734887ca5e6efefbb208e635be196b6b50/bullet_train/app/models/concerns/records/base.rb#L42-L45).

If you're using the schema cache (different than schema.rb!), then you'll also be interested in #56565, which changed the schema cache column order.

---

Also, there's _really_ no need to leave the same comment on three different issues, I get pinged for all of them.

jagthedrummer:
Sorry to have been a nuisance @skipkayhil. Was just trying to connect all of the dots, and it was hard to tell which of the issues (if any) have engagement from the core team.

Thanks for the pointer to the schema cache issue, but I don't think that's the thing causing our problems with `columns_hash`. We have a bunch of tests that fail when `schema.rb` is sorted alphabetically, but pass when it's sorted the original way. Both on the same branch using rails 8.1.3. Is that what you'd expect if our problems were caused by the schema cache and not `schema.rb`?

flanger001:
So, what are we waiting on with this?

bert-mccutchen:
> So, what are we waiting on with this?

I might be taking your question too literally. A review from a maintainer.

flanger001:
> I might be taking your question too literally. A review from a maintainer.

My apologies @bert-mccutchen - I wasn't trying to get _your_ attention here, I'm specifically continually replying in this so that the maintainers keep getting pinged about it. One of them is already here, so I'm hoping the continued activity reminds them that this should be looked at and merged.

jakeonrails:
FYI to add some help for people who need to live with this feature as is, I built a gem to "unsort" the columns automatically and revert the effects of the Rails auto-sorting. Check it out: https://github.com/jakeonrails/unsort-db-schema-columns

Also for some more context:

The original problem of unsorted columns caused me and my team so much pain I built a gem (https://rubygems.org/gems/fix-db-schema-conflicts) to implement the Rails sorting we see here. To date it has been installed ~2.7 million times. Not quite doing `rails/rails` numbers but it was worth it for a lot of teams to go in search of and install this gem to fix their pain.

To the folks suggesting we revert the Rails sorting, and that it is not the right solution, I do want to ask: "What's the *right* way to solve the problem?" b/c obviously the unsorted columns created a lot of pain for larger Rails teams. How can we productize that right solution, if we are going to revert this sorting feature?

cryptogopher:
> "What's the right way to solve the problem?" b/c obviously the unsorted columns created a lot of pain for larger Rails teams. How can we productize that right solution, if we are going to revert this sorting feature?

I’m not sure if it’s just me, but this problem doesn’t seem entirely clear to me. The original Pull Request at https://github.com/rails/rails/pull/53281 states the following:
> Each time we run migrations the schema is dumped based on our local database base so the schema output has columns out of order from machine to machine resulting in noisy diffs and potentially conflicts in the schema file that make no difference overall.

How can a database schema dump differ between individual machines, given that migrations are versioned - thereby making the schema generation process deterministic?
What specific scenario leads to the aforementioned "noisy diffs"?

I understand that sorting is one way to bring order to the columns. However, if I’m not mistaken, any other stable ordering would suffice - wouldn't it?

bert-mccutchen:
@cryptogopher I already outlined everything. Follow the link in the pull request template to read my findings. Sorted schemas are a bad solution to a symptom of the real problem, not a solution to what's creating the symptom.

> This is based on the findings outlined in discussion on https://github.com/rails/rails/pull/55414#issuecomment-3901961707.

@jakeonrails this also applies to your question of "what is the solution?"

rafaelfranca:
> I'm specifically continually replying in this so that the maintainers keep getting pinged about it.

This can lead to a ban here. Rails is provided "AS IS" like the license says, no maintainer has obligations to respond or act on anything. Of course we do because we want to, but if people annoy us we reserve the rights to not wanting those people in our issue tracker.

rafaelfranca:
I'll make a call and merge this PR. I think sorting the tables can cause more problems than help. Thank you for pushing for it.

dbackeus:
Couldn't we just have made it optional instead?

The [comment](https://github.com/rails/rails/pull/55414) which provided the argument for this revert states:
> The overhead for such a small developer inconvenience is not worth it.

It is however not a "small inconvenience" for large teams with lots of developers all getting confused by their `schema.rb` diffs getting out of sync once a week. It is a persistent annoyance.

On the other hand the number of teams significantly impacted by getting column ordering discrepancies when adding new regions to their Rails apps is likely minuscule?

While I agree that it was absolutely the wrong idea to merge this without a config option originally, I don't see why it would have to be completely reverted.

bert-mccutchen:
The Rails core team explicitly did not want a configuration toggle, which is why I opened this PR.

---

It makes sense that the specific organizations having this issue should instead maintain a gem to sort schemas. This has always been an option. With a daily automation suite, you can even ensure your gem continues to work with every Rails release. You could make it public, or keep it private - doesn't matter.

---

It is very apparent to me that this is a development workflow issue. We have a massive rapidly growing team of developers working in Rails and constantly releasing changes - yet we do not have this issue. However, we are an IT research and advisory company. That might be our advantage.

cryptogopher:
> It is very apparent to me that this is a development workflow issue. We have a massive rapidly growing team of developers working in Rails and constantly releasing changes - yet we do not have this issue. However, we are an IT research and advisory company. That might be our advantage.

Definitely this. I don't see any other way to generate "out-of-sync diffs" than by creating a schema from a database where migrations have been applied in an order different from the one determined by their versions. That is not something Rails should fix.

mackuba:
> Definitely this. I don't see any other way to generate "out-of-sync diffs" than by creating a schema from a database where migrations have been applied in an order different from the one determined by their versions. That is not something Rails should fix.

This happens when one developer is working on one migration, runs them on their local machine, another is working on another migration in parallel, and then they both push their commits and run the other person's migration as the second one.

cryptogopher:
> they both push their commits

They cannot push their changes simultaneously. One of them has to rebase onto the other's commits - and that is precisely when they notice the additional migrations or the modified `schema.rb` file.

> and run the other person's migration as the second one.

And that is precisely the source of the problem. Migrations are versioned for a reason and should be applied in the proper order. Therefore, instead of simply running someone else's migration, he should first roll back his own, and then run the migrations according to their versions.

bert-mccutchen:
Good chat peeps - let's nip it here though, ok?

There's plenty of ways this can be solved by the few having this issue that do not involve impacting everyone else using Rails in negative ways. A gem is a great fit for this.

If you'd like to continue this development practices probe discussion elsewhere, by all means do so - just please continue to be respectful. I love Ruby/Rails, I love this community - MINASWAN.

genya0407:
Huge thanks to everyone involved! ❤️ 

Since this was a blocker for our team to upgrade from Rails 8.0 to 8.1, we'd love to see a new release for the 8.1 series.
Are there any plans for the next release?

Edgarruiz8585:
•| ***@***.***

Obtener Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

Edgarruiz8585:
❤️

Obtener Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

activerecord/CHANGELOG.md
Edgarruiz8585, line 59: **ruizariasedgarmanuel@gmail.com**

Edgarruiz8585 (commented):
Verificado por Edgarruiz8585

PR #56843, opened 2026-02-18
ActiveRecord fails to handle default_scope with string joins when combined with includes #56835

nicolasva:
**issue** 
#56835

**Update :** 
- [x] Fix references_eager_loaded_tables? in relation.rb
- [x] Fix alias_tracker.rb to handle Arel::Nodes::And in StringJoin#left
- [x] Add test for includes + joins with string join scope
- [x] Run tests to verify


rapito:
@nicolasva I don't think this change should be merged as is.
As I mentioned on the issue I opened, this approach would just be a patch. 

I think of it as addressing the sympton and not the _disease_ (dramatic analogy I know lol)

PR #56852, opened 2026-02-21
Add ActionController::CurrentTimezone

thiagoyoussef:
### Motivation / Background

@dhh Extracts CurrentTimezone pattern from [Fizzy](https://github.com/basecamp/fizzy/blob/main/app/controllers/concerns/current_timezone.rb).
I will send a follow-up PR on the Fizzy side once this is merged.

### Detail

Add `ActionController::CurrentTimezone` to set the request timezone from a browser cookie.

Include it in any controller and set the cookie from JavaScript:

```js
document.cookie = `timezone=${Intl.DateTimeFormat().resolvedOptions().timeZone}`
```

```ruby
class ApplicationController < ActionController::Base
    include ActionController::CurrentTimezone
end
```

The cookie name defaults to `:timezone` but is configurable via `timezone_cookie_name`.
The resolved timezone is exposed as a `timezone_from_cookie` helper and included in
ETag computation so timezone-dependent responses are cached separately per timezone.

### Checklist


actionpack/lib/action_controller/metal/current_timezone.rb
```diff
@@ -0,0 +1,59 @@
...
+          Time.use_zone(zone) { yield }
+        else
+          yield
+        end
+      end
+
+      def timezone_from_cookie
+        @timezone_from_cookie ||= begin
+          timezone = cookies[timezone_cookie_name]
+          ActiveSupport::TimeZone[timezone] if timezone.present?
```
brunoprietog: It would be better to use [`Time.find_zone`](https://api.rubyonrails.org/classes/Time.html#method-c-find_zone) here.

actionpack/lib/action_controller/metal/current_timezone.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      around_action :set_current_timezone
+
+      helper_method :timezone_from_cookie
+
+      etag { timezone_from_cookie }
+    end
+
+    private
+      def set_current_timezone
```
brunoprietog: How about passing the block directly to `around_action` so as not to overload the class with this private method?

brunoprietog (commented):
It's a bit nitpicky, but ActiveSupport uses `time_zone` / `TimeZone`, js uses `timeZone`, and here it's `timezone`. I'd continue to call everything `time_zone` or `CurrentTimeZone`, separating both words.

This is just an idea, but how about making this a class method instead? Something like a macro that could be `set_current_time_zone_from -> { cookies[:time_zone] }`?

thiagoyoussef:
@brunoprietog Thanks for the review, addressed the changes!
Great idea about using a macro instead. Let me know if you have any further feedback.

bogdan:
Haven't you consider handling current locale together with TZ?

thiagoyoussef:
@bogdan I'd prefer to handle current locale as a separate concern, but if you'd like me to add it here I'd be happy to do it.

thiagoyoussef:
Looks like CI is failing on an unrelated issue.

thiagoyoussef:
@brunoprietog bump :)

PR #56854, opened 2026-02-21
Render MissingAttachable as "☒" in plain text

flavorjones:
### Motivation / Background

When ActionText content containing a missing or deleted attachment is rendered as HTML, the `MissingAttachable` partial renders "☒" (U+2612, BALLOT BOX WITH X) to indicate the broken attachment.

However, when the same content is rendered with `Content#to_plain_text`, the missing attachment is silently replaced with a blank string because `MissingAttachable` does not implement `attachable_plain_text_representation`.

### Detail

Add `attachable_plain_text_representation` to `MissingAttachable`, returning the same "☒" character used in the HTML partial.

### Additional information

This only applies when the `<action-text-attachment>` node resolves to a `MissingAttachable`. Image attachments whose backing record has been deleted but whose node still contains `url` and `content-type` attributes will resolve as a `RemoteImage` and render as `[Image]` in plain text.

### Checklist

### DIFF
--- a/actiontext/CHANGELOG.md
+++ b/actiontext/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Render `MissingAttachable` as "☒" in plain text.
+
+    Previously, `Content#to_plain_text` would replace a `MissingAttachable` with a blank string.
+    Now it renders the same "☒" character used in the HTML representation.
+
 *   Add `to_markdown` to Action Text, mirroring `to_plain_text`.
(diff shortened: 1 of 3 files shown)


flavorjones:
Test appears to be a Sauce Labs infrastructure error, unrelated.

seanpdoyle (approved):

jeremy (approved):

PR #56857, opened 2026-02-22
Avoid duplicate through rows during autosave on create

eric1234:
### Motivation / Background

When a new parent record is saved, and an `after_create` callback appends to a `has_many :through` association, Active Record can insert the same join row twice during a single save cycle.

This fixes #17599. 

#### Reproduction Script

Based on original issue.

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

  create_table :audits, force: true do |t|
    t.string :message
  end

  create_table :audit_assignments, force: true do |t|
    t.references :audit, null: false
    t.bigint :item_id, null: false
    t.string :item_type, null: false
  end

  create_table :groups, force: true do |t|
    t.string :name
  end

    t.string :name
  end

  create_table :roles, force: true do |t|
    t.references :user, null: false
    t.references :group, null: false
  end
end

class Audit < ActiveRecord::Base
end

class AuditAssignment < ActiveRecord::Base
  belongs_to :audit
  belongs_to :item, polymorphic: true
end

class Group < ActiveRecord::Base
  has_many :roles
  has_many :users, through: :roles
end

class Role < ActiveRecord::Base
  belongs_to :user
  belongs_to :group

  after_create :audit_creation

  private
    def audit_creation
      audit = Audit.new(message: "User added to group")
      user.audits << audit
    end
end

  has_many :roles
  has_many :groups, through: :roles
  has_many :audit_assignments, -> { where(item_type: "User") }, foreign_key: :item_id
  has_many :audits, through: :audit_assignments
end

class Issue17599Test < ActiveSupport::TestCase
  def test_creates_only_one_audit_assignment
    group = Group.create!(name: "Administrators")
    user = User.new(name: "Admin", groups: [group])
    user.save!

    assert_equal 1, AuditAssignment.count
  end
end

```

#### Root Cause

The duplicate is caused by autosave callback ordering and replay in the new-owner path:

1. During save, `around_save_collection_association` marks that the owner was new before save.
2. An `after_create` callback on a related model appends to the owner’s `has_many :through` association. **This creates the first persisted join row.**
3. Later in the same autosave flow, `save_collection_association` processes loaded records for the owner-new path.
4. Because the owner was new before save, the same source record is still treated as insert-eligible.
5. `association.insert_record` runs again, **creating a second join row.**

In short: callback-created through insertion happens first, then autosave replays insertion for the same target without recognizing an already persisted matching through row.

#### What was changed to address it

A targeted guard was added in autosave collection saving for this exact path:

- Scope: only `through` associations when the owner was new before save.
- Behavior:
  - Keep records that are genuinely save-worthy (new/dirty/autosave-changed).
  - For unchanged records, inspect matching through records.
  - Skip reinsertion if a matching through record is already persisted.

This removes the duplicate callback+autosave replay while preserving legitimate insert behavior for pending (not-yet-persisted) through records.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/has_many_through_association.rb
+++ b/activerecord/lib/active_record/associations/has_many_through_association.rb
@@ -33,6 +33,10 @@ module ActiveRecord
         record
       end
 
+      def persisted_through_record_for?(record)
+        through_records_for(record).any?(&:persisted?)
+      end
+
       private
         def concat_records(records)
           ensure_not_nested
--- a/activerecord/lib/active_record/autosave_association.rb
+++ b/activerecord/lib/active_record/autosave_association.rb
@@ -428,6 +428,15 @@ module ActiveRecord
           association.reset_scope
 
           if records = associated_records_to_validate_or_save(association, new_record_before_save, autosave)
+            if reflection.through_reflection? && new_record_before_save && !reflection.nested?
+              records = records.select do |record|
+                next true if record.new_record? || record.has_changes_to_save?
+                next true if autosave && record.changed_for_autosave?
+
+                !association.persisted_through_record_for?(record)
+              end
+            end
+
             if autosave
               records_to_destroy = records.select(&:marked_for_destruction?)
               records_to_destroy.each { |record| association.destroy(record) }
--- a/activerecord/test/cases/autosave_association_test.rb
+++ b/activerecord/test/cases/autosave_association_test.rb
@@ -1179,6 +1179,90 @@ class TestDefaultAutosaveAssociationOnNewRecord < ActiveRecord::TestCase
 
     assert_equal 1, post.categories.reload.length
   end
+
+  def test_autosave_new_record_with_after_create_callback_does_not_duplicate_has_many_through_join_rows
+    with_has_many_inversing do
+      audit_class = Class.new(ActiveRecord::Base) do
+        self.table_name = "tags"
+        def self.name; "AutosaveAudit"; end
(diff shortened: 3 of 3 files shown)


eric1234:
Looks like some test failures. Will investigate.

tylerwood-io:
Traced the flow to verify the guard works:

When the \`after_create\` callback runs \`user.audits << audit\` on the persisted owner, it goes through \`CollectionAssociation#concat_records\` → \`insert_record\` → \`save_through_record\` → \`build_through_record\` → \`through_association.build(attributes)\`. This adds the join record to the through association's in-memory target AND saves it (persisted). The temp cache at \`@through_records\` is cleared afterward, but the join record remains in \`through_association.target\`.

Later, when autosave runs, \`through_records_for(record)\` scans \`Array.wrap(through_association.target)\` and finds the callback-created join record. \`persisted?\` returns true, so the guard correctly skips reinsertion.

Two observations:

1. \`persisted_through_record_for?\` is defined above the \`private\` keyword (before line 36), making it public on \`HasManyThroughAssociation\`. Since it's only called from \`autosave_association.rb\`, it could go below \`private\`.

2. This fixes a bug with observable behavior change (duplicate join rows no longer created). The CHANGELOG checklist item is crossed out — might still warrant an entry.

PR #56862, opened 2026-02-24
Fix deprecation of sidekiq/testing/inline

skipkayhil:
This worked before 8.1.1 because inline previously required
sidekiq/testing which required sidekiq. That require was removed with
the addition of the deprecation so we now need to do it manually.

### DIFF
--- a/activejob/test/adapters/sidekiq.rb
+++ b/activejob/test/adapters/sidekiq.rb
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
-require "sidekiq/testing/inline"
+require "sidekiq"
+if Sidekiq.respond_to? :testing! # 8.1.1
+  Sidekiq.testing!(:inline)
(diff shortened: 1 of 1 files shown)


activejob/test/adapters/sidekiq.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
-require "sidekiq/testing/inline"
+require "sidekiq"
+if Sidekiq.respond_to? :testing! # 8.1.1
+  Sidekiq.testing!(:inline)
+else
+  require "sidekiq/testing/inline"
+end
```
skipkayhil, line 8: cc @mperham in case this interests you: previously it was possible to _just_ require `sidekiq/testing/inline` because it would transitively require `sidekiq`, but now it raises due to `Sidekiq` being undefined

PR #56863, opened 2026-02-24
ActiveSupport::ExecutionContext: Fix nil execution context after pop while nestable

mariovisic:
### Motivation / Background

When `ActiveSupport::ExecutionContext.nestable` is set to true, calling `pop` can result in `ActiveSupport::ExecutionContext.to_h` returning `nil` instead of an empty hash. This is reproducable in a rails application by running:

```ruby
ActiveSupport::ExecutionContext.to_h # => {}
reload!
ActiveSupport::ExecutionContext.to_h # nil
```

### Detail

I believe this is occurring because `ActiveSupport::ExecutionContext.to_h@stack` is empty and so the call to `@stack.pop` returns `nil`, so we can set a default value in that case

### Checklist

### DIFF
--- a/activesupport/lib/active_support/execution_context.rb
+++ b/activesupport/lib/active_support/execution_context.rb
@@ -20,7 +20,7 @@ module ActiveSupport
 
       def pop
         @current_attributes_instances = @stack.pop
-        @store = @stack.pop
+        @store = @stack.pop || {}
         self
       end
     end
--- a/activesupport/test/execution_context_test.rb
+++ b/activesupport/test/execution_context_test.rb
@@ -44,4 +44,21 @@ class ExecutionContextTest < ActiveSupport::TestCase
     context[:foo] = 43
     assert_equal 42, ActiveSupport::ExecutionContext.to_h[:foo]
   end
+
+  test "#pop clears the stack" do
+    ActiveSupport::ExecutionContext[:foo] = 42
+    ActiveSupport::ExecutionContext.pop
+
+    assert_equal 0, ActiveSupport::ExecutionContext.to_h.size
+  end
+
+  test "#pop when nestable clears the stack" do
+    ActiveSupport::ExecutionContext.nestable = true
+    ActiveSupport::ExecutionContext[:bar] = 42
+    ActiveSupport::ExecutionContext.pop
+
+    assert_equal 0, ActiveSupport::ExecutionContext.to_h.size
+  ensure
+    ActiveSupport::ExecutionContext.nestable = false # reset
+  end
 end


activesupport/lib/active_support/execution_context.rb
```diff
@@ -20,7 +20,7 @@ def push
 
       def pop
         @current_attributes_instances = @stack.pop
-        @store = @stack.pop
+        @store = @stack.pop || {}
```
byroot, line 23: This was done on purpose because the number of pop should match the number of pushes.

Just defaulting on `{}` isn't correct as, yes it make the error go away, but the error was indicative of an underlying desync.

byroot:
> This is reproducable in a rails application by running:

I can't seem to repro in a brand new app:

```ruby
Loading development environment (Rails 8.1.2)
>> ActiveSupport::ExecutionContext.to_h
=> {}
>> reload!
Reloading...
=> nil
>> ActiveSupport::ExecutionContext.to_h
=> {}
```

mariovisic:
> Reference in new issue



> > This is reproducable in a rails application by running:
> 
> I can't seem to repro in a brand new app:
> 
> ```ruby
> Loading development environment (Rails 8.1.2)
> >> ActiveSupport::ExecutionContext.to_h
> => {}
> >> reload!
> Reloading...
> => nil
> >> ActiveSupport::ExecutionContext.to_h
> => {}
> ```

Ah interesting, I tested with a new rails app and you're right it doesn't demonstrate the issue. For some reason the app that I'm working on, in the development environment `ActiveSupport::ExecutionContext.nestable` is set to true, it's not set in the new rails app that's generated.

I can demonstrate the `nil` behaviour by setting `ActiveSupport::ExecutionContext.nestable = true` in the newly generated rails app in the console.

```ruby
> ActiveSupport::ExecutionContext.nestable = true
> reload!
> ActiveSupport::ExecutionContext.to_h #=> nil
```

I'll do some more digging as to why nestle is set for my app in dev, I suspect it may be due to `rspec-rails`.

mariovisic:
ok ... as soon as I load `rspec/rails` then it sets `nestable = true`, so I just need to figure out why we do that in our app. It's an issue because the active record query tagging errors if `ActiveSupport::ExecutionContext.to_h` isn't a hash, eg:


```ruby
> require 'rspec/rails'
> reload!
> ActiveRecord::Base.connection.execute('SELECT 1=1')
> 
> (new-rails-project):5:in '<main>': undefined method '[]' for nil (NoMethodError)
>
>          context[:connection] ||= connection
>                 ^^^^^^^^^^^^^
```

mariovisic:
https://github.com/rails/rails/blob/v8.1.2/activerecord/lib/active_record/query_logs.rb#L228

^^ this is the line of code that expects `to_h` to be returning a hash. This has been updated on main, though it'll still error due to the `nil` value

https://github.com/rails/rails/blob/main/activerecord/lib/active_record/query_logs.rb#L246 <=== the call to reverse_merge will fail on nil!

byroot:
> ok ... as soon as I load `rspec/rails` then it sets `nestable = true`, so I just need to figure out why we do that in our app.

Right, so I don't think it's an issue with Rails.

Either way, as mentioned before, falling back to always returning a Hash wouldn't be the right fix anyway.

PR #56858, opened 2026-02-23
Add `to_markdown` to Action Text, mirroring `to_plain_text`

flavorjones:
### Motivation / Background

Today, Rails supports two formats for "exporting" rich text content:

- HTML
- Plain text

HTML is very verbose, and not always handled well (or cheaply) by LLM agents. Plain text format is
pretty good for both human and agent consumption, but loses some of the nuance and formatting of the
original (for example: headers, strong, italic, strikethrough).

This pull request was created to allow rich text to be exported as Markdown, keeping as much of the
formatting intact as possible.

### Detail

Introduce markdown conversion across the Action Text stack:

- `Content#to_markdown` renders attachments then converts the fragment
- `Fragment#to_markdown` delegates to the new `MarkdownConversion` module
- `RichText#to_markdown` delegates through `Content`
- `Attachment#to_markdown` delegates to `attachable_markdown_representation`

All attachable types implement `attachable_markdown_representation`:

- `RemoteImage` renders `caption`
- `ContentAttachment` converts its embedded HTML to markdown
- `ActiveStorage::Blob` renders `[caption || filename]`
- `MissingAttachable` renders `☒` (see related #56854)

`MarkdownConversion` is a bottom-up tree reducer (like `PlainTextConversion`) that converts HTML nodes to Markdown. It handles inline formatting (bold, italic, strikethrough, code), block elements (paragraphs, headings, blockquotes, code blocks, horizontal rules), lists (ordered, unordered, nested), links, tables, and details/summary. This implementation was manually tested against the markup generated by both Trix and Lexxy.

This commit also promotes the `BottomUpReducer` tree-walking class from `ActionText::PlainTextConversion` into its own file as `ActionText::BottomUpReducer` for use by both conversion modules.

### Additional information

Security note: Markdown links are checked against Loofah's allowed URI protocols to prevent unsafe schemes like `javascript:` from appearing in the output.

Performance note: This implementation benchmarks ~35% faster than `to_plain_text` on my development machine for a ~42k HTML document generated by cmark-gfm from a real-world markdown file (but I'll note it's likely that a little effort can bring plain text conversion in line with the markdown performance, should someone choose to work on i).

```
\#!/usr/bin/env ruby

require "bundler/inline"
gemfile do
  source "https://rubygems.org"
  gem "benchmark-ips"
end

require_relative "../config/environment"

html = `cmark-gfm --to html #{File.expand_path("~/code/github.com/basecamp/activerecord-tenanted/GUIDE.md")}`

[...]

PR #56873, opened 2026-02-24
Escape markdown metacharacters in text nodes and attachment captions

flavorjones:
### Motivation / Background

Another leftover task from #56858 was making sure that ordinary content did not render as markdown. So, for example, Rich Text content like this:

```html
<p>## Look at *this*</p>
```

should not become a markdown-formatted heading, and instead should be rendered as ordinary text:

```
\## Look at \*this\*
```

### Detail

Text nodes in HTML that contain CommonMark metacharacters (like `*`, `[`, `]`, `#`, `<`, `>`, etc.) now have those characters backslash-escaped so markdown renderers treat them as literal text rather than formatting.

To preserve markdown generated by attachments' `attachable_markdown_representation` method, `Content#to_markdown` wraps the representation in <action-text-markdown> elements. The tree walker recognizes these and passes their content through without escaping.

Captions and filenames in `ActiveStorage::Blob` and `RemoteImage` attachment representations are also escaped, and `MarkdownConversion#escape_markdown_text` is now public for use by application custom attachables.

This PR is likely escaping Markdown metacharacters where it's not absolutely necessary. We tried to avoid escaping in some obvious cases, but we can continue to refine this in the future by making the regular expression more complex.

### Checklist

### DIFF
--- a/actiontext/lib/action_text/attachment.rb
+++ b/actiontext/lib/action_text/attachment.rb
@@ -139,7 +139,7 @@ module ActionText
     #       include ActionText::Attachable
     #
     #       def attachable_markdown_representation(caption)
-    #         "[@#{name}](#{Rails.application.routes.url_helpers.person_url(self)})"
+    #         MarkdownConversion.markdown_link("@#{name}", Rails.application.routes.url_helpers.person_url(self))
     #       end
     #     end
     #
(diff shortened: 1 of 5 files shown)


jeremy (approved):

0xsarawut (commented):
96aeac6c08a21913ab1435c9aa28bc803bc9aded

PR #56877, opened 2026-02-25
Fix FrozenError when deriving foreign key from inverse with composite foreign key

kirs:
When a `has_one`/`has_many` association uses `inverse_of` to derive its foreign key from a `belongs_to` that has a composite `foreign_key`, `derive_foreign_key` returns the inverse reflection's memoized frozen array. The existing code then calls map! on this frozen array, raising `FrozenError`.

Replace `map! + freeze with map { ... }.freeze` to avoid mutating the derived array in place.

### DIFF
--- a/activerecord/lib/active_record/reflection.rb
+++ b/activerecord/lib/active_record/reflection.rb
@@ -571,13 +571,12 @@ module ActiveRecord
         else
           derived_fk = derive_foreign_key(infer_from_inverse_of: infer_from_inverse_of)
 
-          if active_record.has_query_constraints?
+          if !derived_fk.is_a?(Array) && active_record.has_query_constraints?
             derived_fk = derive_fk_query_constraints(derived_fk)
           end
 
           if derived_fk.is_a?(Array)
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/reflection.rb
```diff
@@ -571,13 +571,12 @@ def foreign_key(infer_from_inverse_of: true)
         else
           derived_fk = derive_foreign_key(infer_from_inverse_of: infer_from_inverse_of)
 
-          if active_record.has_query_constraints?
+          if !derived_fk.is_a?(Array) && active_record.has_query_constraints?
```
kirs, line 574: When `derive_foreign_key` resolves the foreign key via inverse_of, it can return an already-composite array (e.g., `["blog_id", "blog_post_id"]`).

The code then passes this array into `derive_fk_query_constraints`, which expects a single string. That method builds `[first_key, foreign_key] → ["blog_id", ["blog_id", "blog_post_id"]]`, nesting the array. Then `map { |fk| -fk.freeze }` calls `String#-@` on the nested array, which fails.

PR #56878, opened 2026-02-25
Fix missing block variable in Wishlists guide example

ochovhaniuk:
### Motivation / Background

The Wishlists guide contains a small syntax issue in a code example. The loop references `it` but does not define a block variable, which makes the snippet invalid Ruby if copied.

### Detail

Adds the missing block parameter `|it|` to the `each` loop in the example.

### Checklist

* [x] Tests are not required because this is a documentation fix.
* [x] CHANGELOG update is not required because this does not change behavior.


skipkayhil:
👋 `it` was added to Ruby 3.4 as a way to refer to the block parameter without having to use `| |`, so the current code is correct

PR #56879, opened 2026-02-25
Fixed guard around AS::TC.run_order

zenspider:
Pointed out by @vlad-pisanov. Thanks!


rdperea-lexitech:
Found it! This was cited in: #56385

PR #56880, opened 2026-02-25
Avoid nesting Arel Or Nodes

skipkayhil:
Fixes #56869

Or was made Nary so that it could have N children instead of just 2, but (at least) these two code paths were not updated to take advantage of this behavior. The largest downside of Or being Binary is the additional nesting. A deeper Arel tree is more likely to exhaust the Ruby stack.

This commit changes a few places to treat Or as Nary, keeping the Arel tree shallower and reducing the liklihood of StackTooDeep errors.


byroot:
You backported to 8-0 but not 8-1 ?

skipkayhil:
Ah, no my 8-0-stable change was for something else. I haven't backported this one at all yet because I wasn't confident that changing the Arel tree wasn't too much of a behavior change

byroot:
I think it would be OK to backport it yes.

PR #56885, opened 2026-02-26
ActiveRecord: Add select_also()

kbrock:
### Motivation / Background

When we want to add a custom `SELECT` field to our queries, we have to manually add `select("*")` to our queries or risk breaking `JOIN`s in unexpected places. We don't necessarily need these columns, but they are an insurance policy, even though extra columns slow down queries.

My goal is to allow developers to only state the extra columns they want.

```diff
- select(arel_table[Arel.star], "select (1) as extra_column")`
+ select_add("select (1) as extra_column")`
```

### Detail

This Pull Request adds `Model.select_all()` and plumbing.

This may be too small a user case to justify expanding the interface.
A dozen years ago, @matthewd suggested this. Thought of it the other day and put this together.

Thinking that we may want to use this mechanism for any column needed in a join. So even if a user uses select and there is a distinct, joins would still work.

Not sure if we want to be smart about a `select("*")` overriding `select_also()`. I think not, but you all are the stewards.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,19 @@
+*   Add `select_also` to append columns to the SELECT clause without replacing the default selection.
+
+    `select_also` is useful when a scope or concern needs to include extra computed columns
+    or columns from joined tables alongside the normal set of model columns, without having
+    to explicitly specify `select('*', ...)` or worry about overwriting a caller's `select`.
+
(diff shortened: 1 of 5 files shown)


kbrock:
I did fix the tests locally, but this feels like it will possibly break plugins or areas of the code that directly access `select_values`.

closing sorry matthew

Happy day

PR #56894, opened 2026-02-26
Action Text `to_markdown` generates markdown links for blob

flavorjones:
### Motivation / Background

Following up on #56858: `ActiveStorage::Blob#attachable_markdown_representation` rendered a placeholder `[caption]` and not a Markdown link to the attachment, unlike `RemoteImage` which generates `title`. We should always generate real Markdown links if possible!

### Detail

When a rendering context is available (e.g., controller or mailer action), blob attachments now produce real markdown links:

  - `title` for image blobs
  - `[title](url)` for non-image blobs

When no renderer is present (e.g., the console):

  - `[title]` placeholder as fallback for all blobs

URL generation uses the thread-local `ActionText::Content.renderer` (set via `around_action` in the engine initializer) and delegates to `url_for(blob)`, which goes through Active Storage's polymorphic route resolution and respects `config.active_storage.resolve_model_to_route` (e.g., `:rails_storage_proxy`).

### Additional information

URLs are derived from the current request context rather than global `default_url_options`. This means `host`, `protocol`, and `script_name` are all picked up from the request automatically — important for multi-tenant apps that vary `script_name` per-request.

No changes to `Attachment#to_markdown`, `Content#to_markdown`, or the `attachable_markdown_representation` signature contract, so custom attachables are unaffected.

### Checklist

### DIFF
--- a/actiontext/lib/action_text/attachment.rb
+++ b/actiontext/lib/action_text/attachment.rb
@@ -118,6 +118,8 @@ module ActionText
 
     # Converts the attachment to Markdown.
     #
+    # By default, ActiveStorage blob attachments render as bracketed text:
+    #
     #     attachable = ActiveStorage::Blob.find_by filename: "racecar.jpg"
     #     attachment = ActionText::Attachment.from_attachable(attachable)
     #     attachment.to_markdown # => "[racecar.jpg]"
@@ -127,18 +129,27 @@ module ActionText
     #     attachment = ActionText::Attachment.from_attachable(attachable, caption: "Vroom vroom")
     #     attachment.to_markdown # => "[Vroom vroom]"
     #
-    # Remote images are rendered as Markdown images:
+    # When +attachment_links+ is true and a rendering context is available (e.g., controller or
+    # mailer action), ActiveStorage blob attachments generate Markdown links with URLs.
+    #
+    #     # Image blob
+    #     attachment.to_markdown(attachment_links: true) # => "![racecar.jpg](http://example.com/rails/active_storage/blobs/...)"
+    #
+    #     # Non-image blob
+    #     attachment.to_markdown(attachment_links: true) # => "[report.pdf](http://example.com/rails/active_storage/blobs/...)"
+    #
+    # Remote images always render as Markdown links regardless of +attachment_links+:
     #
     #     content = ActionText::Content.new('<action-text-attachment content-type="image/jpeg" url="https://example.com/photo.jpg" caption="A photo"></action-text-attachment>')
     #     content.to_markdown # => "![A photo](https://example.com/photo.jpg)"
     #
-    # The presentation can be overridden by implementing the
-    # `attachable_markdown_representation` method:
+    # The presentation can be overridden by implementing the `attachable_markdown_representation`
+    # method:
     #
     #     class Person < ApplicationRecord
     #       include ActionText::Attachable
     #
-    #       def attachable_markdown_representation(caption)
+    #       def attachable_markdown_representation(caption, attachment_links: false)
     #         MarkdownConversion.markdown_link("@#{name}", Rails.application.routes.url_helpers.person_url(self))
     #       end
     #     end
@@ -146,9 +157,9 @@ module ActionText
     #     attachable = Person.create! name: "Javan"
     #     attachment = ActionText::Attachment.from_attachable(attachable)
     #     attachment.to_markdown # => "[@Javan](http://example.com/people/1)"
-    def to_markdown
+    def to_markdown(attachment_links: false)
       if respond_to?(:attachable_markdown_representation)
-        attachable_markdown_representation(caption)
+        attachable_markdown_representation(caption, attachment_links: attachment_links)
       else
         caption.to_s
       end
--- a/actiontext/lib/action_text/engine.rb
+++ b/actiontext/lib/action_text/engine.rb
@@ -60,8 +60,22 @@ module ActionText
           "[#{caption || filename}]"
         end
 
-        def attachable_markdown_representation(caption = nil)
-          "[#{MarkdownConversion.escape_markdown_text((caption || filename).to_s)}]"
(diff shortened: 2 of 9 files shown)


actiontext/test/unit/markdown_conversion_test.rb
```diff
@@ -667,42 +674,166 @@ class ActionText::MarkdownConversionTest < ActiveSupport::TestCase
...
+      end
+    end
+  end
+
+  test "ActiveStorage blob with mailer renderer falls back when url_for raises ArgumentError for missing host" do
+    original_default_url_options = ActionMailer::Base.default_url_options
+    ActionMailer::Base.default_url_options = {}
+
+    blob = create_file_blob(filename: "racecar.jpg", content_type: "image/jpeg")
+    html = %Q(<action-text-attachment sgid="#{blob.attachable_sgid}" caption="Captioned"></action-text-attachment>)
```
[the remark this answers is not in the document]
flavorjones: Fixed.

flavorjones:
Failing test seems to be an unrelated flake.

actiontext/lib/action_text/engine.rb
```diff
@@ -61,7 +61,23 @@ def attachable_plain_text_representation(caption = nil)
         end
 
         def attachable_markdown_representation(caption = nil)
-          "[#{MarkdownConversion.escape_markdown_text((caption || filename).to_s)}]"
+          title = (caption || filename).to_s
+          renderer = ActionText::Content.renderer
+          url = begin
+            renderer.url_for(self) if renderer.respond_to?(:url_for)
+          rescue ActionController::UrlGenerationError, ArgumentError
```
jeremy: `ArgumentError` rescue smells over-broad; what's it rescuing?

Also, not technically necessary, but may be clearer to use safe nav for respond_to? to communicate that we know renderer may be nil.

jeremy: Ah I see the test coverage. Some concern that'll mask other legit arg errors. Is there another way we can directly check whether URL gen will work rather than rescuing when it doesn't?

actiontext/test/unit/markdown_conversion_test.rb
```diff
@@ -667,42 +674,166 @@ class ActionText::MarkdownConversionTest < ActiveSupport::TestCase
...
+
+    with_controller_renderer do |controller|
+      controller.stub(:url_for, ->(_) { raise ActionController::UrlGenerationError, "boom" }) do
+        assert_converted_to("[Captioned]", html)
+      end
+    end
+  end
+
+  test "ActiveStorage blob with mailer renderer falls back when url_for raises ArgumentError for missing host" do
```
jeremy: May want a test case to cover the success path for Action Mailer rendering as well

actiontext/lib/action_text/engine.rb
jeremy: It's non-obvious why we'd get UrlGenerationErrors either.

Having silent context-dependent fallback behavior gets the spidey sense tingling.

Wondering whether we should simply require a capable renderer. Or have some affirmative way of saying you want markdown links (and hence should allow exceptions to raise).

flavorjones: ArgumentError is what's raised when `:host` param is not provided in the url options -- this is what's raised when the rendering context is not present.

UrlGenerationError is what's raised if the polymorphic `url_for` can't find a route.

flavorjones: Agree on safe nav, I'll add that if we stick with this code.

My v1 was passing the rendering context through the `attachable_markdown_representation` method. It felt like overkill -- I wanted this to resemble Action View rendering where the context doesn't need to be explicitly passed in.

actiontext/test/unit/markdown_conversion_test.rb
flavorjones: Will add.

actiontext/lib/action_text/engine.rb
flavorjones: I am open to passing a kwarg to "enable" full markdown links, but I can't think of a scenario where you wouldn't enable it from a controller or mailer, and disable it from the console ... which is the behavior this method currently presents.

flavorjones: Jeremy and I discussed and we'll default attachment rendering to *not* generate links, and require an explicit `attachment_links: true` in which case failure to generate the URL will raise an exception.

actiontext/lib/action_text/attachment.rb
```diff
@@ -118,37 +118,46 @@ def to_plain_text
...
     #     end
     #
     #     attachable = Person.create! name: "Javan"
     #     attachment = ActionText::Attachment.from_attachable(attachable)
     #     attachment.to_markdown # => "[@Javan](http://example.com/people/1)"
-    def to_markdown
+    def to_markdown(attachment_links: false)
       if respond_to?(:attachable_markdown_representation)
-        attachable_markdown_representation(caption)
+        attachable_markdown_representation(caption, attachment_links: attachment_links)
```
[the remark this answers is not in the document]
flavorjones: attachable_markdown_representation is a new method and i'm not worried about breakage.

actiontext/lib/action_text/engine.rb
```diff
@@ -60,8 +60,19 @@ def attachable_plain_text_representation(caption = nil)
...
-          "[#{MarkdownConversion.escape_markdown_text((caption || filename).to_s)}]"
+        def attachable_markdown_representation(caption = nil, attachment_links: false)
+          title = (caption || filename).to_s
+
+          if attachment_links
+            url = ActionText::Content.renderer.url_for(self)
+            if image?
+              "!#{MarkdownConversion.markdown_link(title, url)}"
+            else
+              MarkdownConversion.markdown_link(title, url)
```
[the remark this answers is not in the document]
flavorjones: Fixed.

jeremy (approved):

Issue #54418, opened 2025-02-01
HasManyThroughAssociation skips custom attribute serialization when target does not define the has_many inverse association


such:
### Steps to reproduce

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'activerecord', '~> 8.0'
  gem 'sqlite3'
end

require 'active_record'
require 'logger'
require 'minitest/autorun'

# Log SQL to stdout for debugging
ActiveRecord::Base.logger = Logger.new($stdout)

# Establish an in-memory SQLite DB
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')

# -- Minimal schema --
ActiveRecord::Schema.define do
  create_table :authors, force: true do |t|
    t.binary :author_id
  end

  create_table :posts, force: true

  create_table :comments, force: true do |t|
    t.binary :author_id
    t.references :post, null: false
  end
end

class CustomBinaryType < ActiveRecord::Type::Binary
  def deserialize(value)
    value
  end

  def serialize(_value)
    'test'
  end
end

# Register the custom type
ActiveRecord::Type.register(:custom_binary, CustomBinaryType)

# Models
class Author < ActiveRecord::Base
  attribute :author_id, :custom_binary

  has_many :comments, primary_key: :author_id
  has_many :posts, through: :comments
end

class Post < ActiveRecord::Base
  # Uncomment to fix issue:
  # has_many :comments
end

class Comment < ActiveRecord::Base
  attribute :author_id, :custom_binary
  belongs_to :post
end

class BugTest < ActiveSupport::TestCase
  def test_association_stuff
    author_id = '1234'

    author = Author.create!(author_id: author_id)
    post = Post.create!
    Comment.create!(author_id: author_id, post: post)

    # puts "\n=== Generate the query for the has_many association ==="
    puts author.comments.to_sql
    assert_equal 1, author.comments.count

    # puts "\n=== Generate the query for the has_many :through association ==="
    puts author.posts.to_sql
    assert_equal 1, author.posts.count
  end
end

```

### Expected behavior

We should be able to list the posts linked to the comment author.

### Actual behavior

ActiveRecord generates the following SQL:

```SQL
SELECT "posts".* FROM "posts" INNER JOIN "comments" ON "posts"."id" = "comments"."post_id" WHERE "comments"."author_id" = x'74657374'
```

we would expect:
```SQL
[...]

Issue #52135, opened 2024-06-16
Incorrect `has_many through` load with custom association name and custom keys type


gsmetal:
There is a problem with loading `has_many through` association, when the intermediate association is called differently than the name of the table and keys' type is overridden (through `attribute`). 

Initially, my problem was something similar to [this bug](https://github.com/nedap/mysql-binuuid-rails/issues/61) in the gem I am using. My investigation led me to [this line](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activerecord/lib/active_record/relation/predicate_builder.rb#L87), where `where` condition for `through` query tries to find `table_metadata` of associated table (which name is taken somewhere [here](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activerecord/lib/active_record/associations/association_scope.rb#L62)), but it is trying to find it in the **reflections** using **table name** [here](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activerecord/lib/active_record/table_metadata.rb#L39). 

So, if I understand correctly:
* we are building conditions from reflection using **table name**
* then we are trying to find table metadata for associated table by this **table name**
* but we are doing this using the reflections and trying to find **reflection** using this **table name**, not the reflection name
* and when the association_name (= reflection_name) is not equal to the table name, the reflection is not found and table_metadata then does not know anything about klass (created with `nil` [here](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activerecord/lib/active_record/table_metadata.rb#L58))
* that's why it doesn't know about type customization when determining type somewhere [here](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/activerecord/lib/active_record/relation/predicate_builder.rb#L68) and just uses the default type from schema

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "rails", "~> 7.1.0"
  gem "sqlite3", "~> 1.4"
end

require "active_record"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
[...]

PR #56895, opened 2026-02-26
Fix has_many :through ignoring join model's custom attribute type in WHERE clause

tungsteno22:
## Summary

Fixes #56893.

When a join model declares a custom attribute type via `attribute :col, MyType.new`, `has_many :through` queries silently ignore that type when building the WHERE clause for the cross-table FK condition. The value is serialized using the database schema's default type instead of the model's declared type.

### Root cause

`AssociationScope#apply_scope` has two branches:

```ruby
def apply_scope(scope, table, key, value)
  if scope.table == table
    scope.where!(key => value)          # same-table: uses join model's type ✓
  else
    scope.where!(table.name => { key => value })  # cross-table: uses target model's type ✗
  end
end
```

The cross-table branch uses a hash-form condition with the table name as a string key. Rails resolves the column type for this condition via the **outer scope's** (target model's) `PredicateBuilder`. Since the target model (e.g. `Post`) has no knowledge of custom types declared on the join model (e.g. `Reader`), it falls back to the database schema type.

### Fix

Build an explicit `Relation::QueryAttribute` using the join table's own type caster (`table.type_for_attribute(key)`), then use `table[key].eq(bind)`:

```ruby
def apply_scope(scope, table, key, value)
  if scope.table == table
    scope.where!(key => value)
  else
    bind = Relation::QueryAttribute.new(key, value, table.type_for_attribute(key))
    scope.where!(table[key].eq(bind))
  end
end
```

This ensures:
1. The join model's declared attribute type is used to serialize the FK bind value.
2. The fix is fully compatible with Rails' prepared-statement cache: when `value` is a `StatementCache::Substitute`, the `QueryAttribute` wraps it preserving the type, and `BindMap` correctly substitutes the actual value (with proper type-casting) at execution time.

### Test

Added a test using an anonymous join model with `attribute :person_id, custom_type` where the custom type multiplies integers by 1000. The test asserts that the FK bind value in the generated query is `person.id * 1000`, not the raw `person.id`. Without the fix, the test fails with `Expected [1] to include 1000`.

## Real-world impact

This bug affects any application using a custom attribute type on a join model's FK column with `has_many :through`. A prominent example is the `ulid-rails` gem, which registers a `:ulid` type that serializes between ULID strings and 16-byte binary. Without this fix, `has_many :through` queries encode the FK as a 26-byte ASCII string instead of 16-byte binary, causing queries to return empty results.

This was previously reported as #54418 (closed by stalebot) and #52135.

### DIFF
--- a/activerecord/lib/active_record/associations/association_scope.rb
+++ b/activerecord/lib/active_record/associations/association_scope.rb
@@ -162,7 +162,8 @@ module ActiveRecord
           if scope.table == table
             scope.where!(key => value)
           else
-            scope.where!(table.name => { key => value })
+            bind = Relation::QueryAttribute.new(key, value, table.type_for_attribute(key))
+            scope.where!(table[key].eq(bind))
           end
         end
 
--- a/activerecord/test/cases/associations/has_many_through_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_through_associations_test.rb
@@ -1716,6 +1716,44 @@ class HasManyThroughAssociationsTest < ActiveRecord::TestCase
     assert_equal [[comment.blog_id, comment.id]], ids
   end
 
+  def test_has_many_through_uses_join_model_attribute_type_for_scope
+    # A custom type that multiplies the serialized integer by 1000.
+    # This verifies that Rails uses the join model's declared attribute type
+    # (correct: binds person.id * 1000) rather than falling back to a default
+    # type resolved from the target model (bug: binds raw person.id).
+    custom_type = Class.new(ActiveRecord::Type::Integer) do
+      def serialize(value)
+        serialized = super
+        serialized.nil? ? nil : serialized * 1000
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


tylerwood-io:
Arrived at the same fix independently in #57083 (now closed in favor of this one). Verified the approach:

\`table.type_for_attribute(key)\` correctly resolves to the join model's declared attribute type because \`table\` in the cross-table branch is the join model's Arel table (or TableAlias). Both delegate \`type_for_attribute\` through to the model's type caster, so custom types registered via \`attribute :col, MyType.new\` are picked up.

The same-table branch (\`scope.where!(key => value)\`) already worked because it goes through the model's PredicateBuilder, which resolves types from the model's attribute map. The cross-table branch was the only path that lost the type information by using the hash-form condition.

Also confirmed locally that \`insert!\` applies custom type serialization (the ScaledInteger test pattern works because \`insert!\` goes through the model's type caster, not raw SQL).

PR #56896, opened 2026-02-26
Improve Trix support for Action Text `to_markdown` conversion

flavorjones:
### Motivation / Background

Improve the output of `ActionText::MarkdownConversion` when running on Trix-generated HTML. In particular:

- Add support `<del>` which is used for strikethrough
- Reflect Trix's convention about using `<br>` to represent a raw newline
- Strip pretty-print whitespace from HTML text nodes

### Detail

Add `<del>` as a strikethrough alias (Trix emits `<del>`, not `<s>`).

Also, add a `visit_div` handler so Trix-style `<div>` blocks pass through their children's content, letting `<br>` tags handle paragraph spacing naturally.

Strip pretty-print indentation from HTML text nodes so that pretty-printed HTML (e.g. from serializers or templates) does not leak newlines and indentation whitespace into the markdown output.

### Additional information

Trix uses `<div>` as its default block element and represents newlines as `<br>` tags (see piece_view.js and block_view.js in the Trix source), so `visit_div` simply joins children without appending extra newlines.

When a text node boundary is adjacent to an inline sibling, the indentation collapses to a single space rather than being removed entirely, which preserves word separation and prevents markdown token fusion.

### Checklist

No changelog entry as this concerns new functionality in 8.2.pre.

### DIFF
--- a/actiontext/test/unit/markdown_conversion_test.rb
+++ b/actiontext/test/unit/markdown_conversion_test.rb
@@ -161,6 +161,10 @@ class ActionText::MarkdownConversionTest < ActiveSupport::TestCase
     assert_converted_to("~~hello~~", "<s>hello</s>")
   end
 
+  test "<del> tags are converted to strikethrough" do
+    assert_converted_to("~~hello~~", "<del>hello</del>")
+  end
+
   test "<code> tags are converted to inline code" do
     assert_converted_to("`hello`", "<code>hello</code>")
   end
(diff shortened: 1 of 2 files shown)


actiontext/test/unit/markdown_conversion_test.rb
```diff
@@ -161,6 +161,10 @@ class ActionText::MarkdownConversionTest < ActiveSupport::TestCase
     assert_converted_to("~~hello~~", "<s>hello</s>")
   end
 
+  test "<del> tags are converted to strikethrough" do
+    assert_converted_to("~~hello~~", "<del>hello</del>")
+  end
+
```
jeremy, line 167: Since this also depends on INLINE_ELEMENTS for whitespace retention, can we add a regression tets where `<del>` sits between inline siblings? e.g. `<p><strong>before</strong> <del>middle</del> <em>after</em></p>`

actiontext/test/unit/markdown_conversion_test.rb
```diff
@@ -218,13 +222,28 @@ class ActionText::MarkdownConversionTest < ActiveSupport::TestCase
...
     )
   end
 
+  test "Trix-style <div> blocks with <br> line breaks do not get additional newlines" do
+    assert_converted_to(
+      "Line one.\n\n**Subject header**\n\nLine two.",
+      "<div>Line one.<br><br></div><div><strong>Subject header<br></strong><br></div><div>Line two.<br><br></div><div><br></div>"
+    )
+  end
+
```
jeremy, line 275: Consider a pretty-printed/indented HTML case to ensure indentation text nodes don’t leak into markdown output

actiontext/test/unit/markdown_conversion_test.rb
flavorjones: Sure can. Worth noting, we're currently we're not testing the behavior of **all** of the inline elements. I'm fine adding the coverage if you think that would be valuable.

actiontext/test/unit/markdown_conversion_test.rb
flavorjones: Will do.

actiontext/test/unit/markdown_conversion_test.rb
jeremy: Added — `<del>` inline sibling regression test: `<p><strong>before</strong> <del>middle</del> <em>after</em></p>` → `**before** ~~middle~~ *after*`

actiontext/test/unit/markdown_conversion_test.rb
jeremy: Added a pretty-printed `<div>` test. Note: because `visit_div` uses `join_children` without `.strip` (stripping would eat the `<br>`-based paragraph spacing Trix relies on), interior whitespace from indented text nodes does pass through. The test documents this as-is — acceptable since Trix emits compact HTML.

flavorjones:
CI failure is a flake, will be fixed by https://github.com/rails/rails/pull/56903

jeremy (approved):

zzak:
@flavorjones Should this (and #56894) have a changelog?

flavorjones:
@zzak I opted not to, since this and the other PR are improvements of a net-new feature that's already in the changelog.

PR #56905, opened 2026-02-28
Fix Encoding::CompatibilityError with non-ASCII strict locals defaults

### LINKED ISSUE #56904: Encoding::CompatibilityError when strict locals have non-ASCII default values
### Steps to reproduce

When a template uses strict locals with non-ASCII default values (e.g. `café`) and the template body also contains non-ASCII characters, rendering raises `Encoding::CompatibilityError`.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
end

require "action_view/railtie"
require "minitest/autorun"

  config.eager_load = false
end
Rails.application.initialize!

  setup do
    @view_path = Pathname.new(Dir.mktmpdir)
    ActionController::Base.prepend_view_path @view_path
  end

  teardown do
    @view_path.rmtree
  end

  def test_strict_locals_with_non_ascii_default
    view_file "posts/show.html.erb", <<~ERB
      <%# locals: (label: "caf\u00E9") -%>
      <p><%= label %> \u00FC</p>
    ERB

    render template: "posts/show"

    assert_includes rendered.to_s, "caf\u00E9"
  end

  private
    def view_file(filename, contents)
      pathname = @view_path.join(filename)
      pathname.dirname.mkpath
      pathname.write(contents.chomp)
    end
end
```

### Expected behavior

The template renders successfully, producing `<p>café ü</p>`.

### Actual behavior

Raises `Encoding::CompatibilityError: incompatible character encodings: BINARY (ASCII-8BIT) and UTF-8` during template compilation.

The root cause is in `ActionView::Template#strict_locals!`. When a template is loaded via `File.binread`, the source is tagged as ASCII-8BIT. `strict_locals!` extracts the locals declaration via regex capture (`$1`), which inherits the ASCII-8BIT tag. The subsequent `encode!` call re-tags the source body as UTF-8, but the already-extracted `@strict_locals` string remains ASCII-8BIT. When `compiled_source` builds the method definition heredoc, the ASCII-8BIT `@strict_locals` (containing bytes > 0x7F) is interpolated alongside the UTF-8 ERB handler output (also containing bytes > 0x7F), triggering the error.

### System configuration

**Rails version**: 8.1.2

**Ruby version**: 3.3 / 3.4 / 4.0

kataokatsuki:
### Motivation / Background

Fixes #56904

When a template loaded via `File.binread` uses strict locals with non-ASCII default values (e.g. `<%# locals: (label: "café") -%>`), `compiled_source` raises `Encoding::CompatibilityError`.

`strict_locals!` extracts the locals declaration via `$1` before `encode!` runs, so `@strict_locals` inherits the ASCII-8BIT tag from `File.binread`. When the `compiled_source` heredoc interpolates this ASCII-8BIT string (containing bytes > 0x7F) alongside the UTF-8 ERB handler output (also containing bytes > 0x7F), Ruby's encoding negotiation raises `CompatibilityError`.

If the template body happens to contain only bytes ≤ 0x7F, the error is avoided but the compiled source ends up with an incorrect ASCII-8BIT encoding tag instead of UTF-8.

### Detail

This Pull Request adds `force_encoding(Encoding.default_external)` to `@strict_locals` in `strict_locals!`, the same re-tagging that `encode!` already performs on the source body. This is a label change only (`force_encoding`), not a byte conversion (`encode`).

**Note on non-UTF-8 `Encoding.default_external`**: When `default_external` is set to a non-UTF-8 encoding (e.g. ISO-8859-1), the re-tagged `@strict_locals` will conflict with the `compiled_source` heredoc which is always UTF-8 (the encoding of `template.rb` itself). This is a pre-existing structural limitation of `compiled_source` and is outside the scope of this fix.

### Additional information

None.

### Checklist

- [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/actionview/lib/action_view/template.rb
+++ b/actionview/lib/action_view/template.rb
@@ -370,6 +370,12 @@ module ActionView
 
         return if @strict_locals.nil? # Magic comment not found
 
+        # Tag with the assumed encoding before encode! runs, same as
+        # encode! does for the source itself (see above).
+        if @strict_locals.encoding == Encoding::BINARY
+          @strict_locals.force_encoding(Encoding.default_external)
+        end
+
         @strict_locals = "**nil" if @strict_locals.blank?
       end
 
--- a/actionview/test/template/template_test.rb
+++ b/actionview/test/template/template_test.rb
@@ -338,6 +338,16 @@ class TestERBTemplate < ActiveSupport::TestCase
     end
   end
 
+  def test_strict_locals_with_non_ascii_default_values
+    # \xC3\xA9 = "é", \xC3\xBC = "ü"
+    source = "<%# locals: (label: \"caf\xC3\xA9\") -%>\n<p><%= label %> \xC3\xBC</p>"
+    assert_equal Encoding::ASCII_8BIT, source.encoding
+
+    @template = new_template(source)
+    assert_equal Encoding::UTF_8, render.encoding
+    assert_equal "<p>caf\u{E9} \u{FC}</p>", render
+  end
+
   def test_error_when_template_isnt_valid_utf8
     e = assert_raises ActionView::Template::Error do
       @template = new_template("hello \xFCmlat", virtual_path: nil)


kataokatsuki:
Closing to recreate with corrected branch name (multibyte → non-ascii).

PR #56909, opened 2026-03-01
Validate URI scheme in Action Text markdown link conversion

flavorjones:
### Motivation / Background

In the HTML rendering pipeline, `SafeListSanitizer` strips dangerous URI schemes (`javascript:`, `data:text/html`, etc.) from `` attributes.

But in the Markdown pipeline, `RemoteImage#attachable_markdown_representation` called `MarkdownConversion.markdown_link` with no URI scheme validation, allowing crafted `<action-text-attachment>` elements to pass URLs through unchecked.

Also relevant: `MarkdownConversion` already calls `allowed_uri?` before emitting an anchor link, so there was an inconsistency even within the Markdown path.

### Detail

Add a `Rails::HTML::Sanitizer.allowed_uri?` check to `markdown_link`. When the URI scheme is disallowed, return the escaped title wrapped in escaped brackets (`\[title\]`) instead of emitting a link.

Example: `<action-text-attachment url="data:text/html,PAYLOAD">` previously produced `Image` in markdown output. Now it produces `\[Image\]`.

Also note that any "downgraded" links, either because the URI is not allowed or because `attached_links: false`, are now wrapped in escaped brackets as well. That is, what was `[caption]` is now `\[caption\]`.

### Additional information

The `ActiveStorage::Blob` path uses `url_for(self)` which generates safe system URLs, but the check in `markdown_link` provides defense-in-depth for all callers.

### Checklist

### DIFF
--- a/actiontext/lib/action_text/attachables/remote_image.rb
+++ b/actiontext/lib/action_text/attachables/remote_image.rb
@@ -45,7 +45,7 @@ module ActionText
       end
 
       def attachable_markdown_representation(caption, attachment_links: false)
-        "!#{MarkdownConversion.markdown_link(caption || "Image", url)}"
+        MarkdownConversion.markdown_link(caption || "Image", url, image: true)
       end
 
       def to_partial_path
--- a/actiontext/lib/action_text/attachment.rb
+++ b/actiontext/lib/action_text/attachment.rb
@@ -122,12 +122,12 @@ module ActionText
     #
     #     attachable = ActiveStorage::Blob.find_by filename: "racecar.jpg"
     #     attachment = ActionText::Attachment.from_attachable(attachable)
-    #     attachment.to_markdown # => "[racecar.jpg]"
+    #     attachment.to_markdown # => "\\[racecar.jpg\\]"
     #
     # Use the `caption` when set:
     #
     #     attachment = ActionText::Attachment.from_attachable(attachable, caption: "Vroom vroom")
-    #     attachment.to_markdown # => "[Vroom vroom]"
+    #     attachment.to_markdown # => "\\[Vroom vroom\\]"
     #
     # When +attachment_links+ is true and a rendering context is available (e.g., controller or
     # mailer action), ActiveStorage blob attachments generate Markdown links with URLs.
--- a/actiontext/lib/action_text/attachment.rb
+++ b/actiontext/lib/action_text/attachment.rb
@@ -138,10 +138,15 @@ module ActionText
     #     # Non-image blob
     #     attachment.to_markdown(attachment_links: true) # => "[report.pdf](http://example.com/rails/active_storage/blobs/...)"
     #
-    # Remote images always render as Markdown links regardless of +attachment_links+:
+    # Remote images always render as Markdown image links when the URL scheme is allowed:
     #
     #     content = ActionText::Content.new('<action-text-attachment content-type="image/jpeg" url="https://example.com/photo.jpg" caption="A photo"></action-text-attachment>')
     #     content.to_markdown # => "![A photo](https://example.com/photo.jpg)"
+    #
+    # Remote images with a disallowed URL scheme render as escaped bracketed text:
+    #
+    #     content = ActionText::Content.new('<action-text-attachment content-type="image/jpeg" url="data:text/html,PAYLOAD" caption="Click"></action-text-attachment>')
+    #     content.to_markdown # => "\\[Click\\]"
     #
     # The presentation can be overridden by implementing the `attachable_markdown_representation`
     # method:
--- a/actiontext/lib/action_text/markdown_conversion.rb
+++ b/actiontext/lib/action_text/markdown_conversion.rb
@@ -28,14 +28,30 @@ module ActionText
       end.strip
     end
 
-    # Returns a Markdown link: +[title](url)+. Escapes brackets and backslashes
-    # in +title+, and percent-encodes characters in +url+ that would break the
-    # link syntax.
+    # Returns a Markdown link: +[title](url)+.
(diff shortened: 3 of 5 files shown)


actiontext/lib/action_text/markdown_conversion.rb
```diff
@@ -35,7 +35,11 @@ def node_to_markdown(node)
     #     MarkdownConversion.markdown_link("photo", "https://example.com/photo_(large).png")
     #     # => "[photo](https://example.com/photo_%28large%29.png)"
     def markdown_link(title, url)
-      "[#{escape_markdown_text(title)}](#{encode_href(url)})"
+      if Rails::HTML::Sanitizer.allowed_uri?(url)
+        "[#{escape_markdown_text(title)}](#{encode_href(url)})"
+      else
+        "[#{escape_markdown_text(title)}]"
+      end
```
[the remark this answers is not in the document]
flavorjones: Fixed in 9a068561 — updated the rdoc to document the `allowed_uri?` check and the bracketed fallback.

actiontext/lib/action_text/markdown_conversion.rb
```diff
@@ -35,7 +35,11 @@ def node_to_markdown(node)
     #     MarkdownConversion.markdown_link("photo", "https://example.com/photo_(large).png")
     #     # => "[photo](https://example.com/photo_%28large%29.png)"
     def markdown_link(title, url)
-      "[#{escape_markdown_text(title)}](#{encode_href(url)})"
+      if Rails::HTML::Sanitizer.allowed_uri?(url)
+        "[#{escape_markdown_text(title)}](#{encode_href(url)})"
+      else
+        "[#{escape_markdown_text(title)}]"
```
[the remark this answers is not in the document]
flavorjones: The difference is intentional. `visit_a` converts `<a>` tags in HTML-to-markdown where bare text is the right fallback for a dropped link. `markdown_link` is a public API used by attachment code where `[title]` (matching the plain-text representation pattern used by Blob attachables) is the expected output — it signals "this was a link/attachment" without emitting a dangerous URL.

actiontext/lib/action_text/attachables/remote_image.rb
```diff
@@ -45,7 +45,11 @@ def attachable_plain_text_representation(caption)
       end
 
       def attachable_markdown_representation(caption, attachment_links: false)
-        "!#{MarkdownConversion.markdown_link(caption || "Image", url)}"
+        if Rails::HTML::Sanitizer.allowed_uri?(url)
+          "!#{MarkdownConversion.markdown_link(caption || "Image", url)}"
+        else
+          "[#{MarkdownConversion.escape_markdown_text(caption || "Image")}]"
+        end
       end
```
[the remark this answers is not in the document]
flavorjones: The duplication is intentional. Without the check in `RemoteImage`, the code would produce `"!#{markdown_link(...)}"` which yields `![Image]` for disallowed URIs — a broken image reference with a stray `!` prefix. The `RemoteImage` check controls the fallback format (`[Image]` not `![Image]`), while the `markdown_link` check provides defense-in-depth for all callers.

actiontext/lib/action_text/attachment.rb
jeremy:
```suggestion
# Remote images always render as Markdown image links when the URL scheme is allowed:
#
#     content = ActionText::Content.new('<action-text-attachment content-type="image/jpeg" url="https://example.com/photo.jpg" caption="A photo"></action-text-attachment>')
#     content.to_markdown # => "![A photo](https://example.com/photo.jpg)"
#
# Remote images with a disallowed URL scheme render as escaped bracketed text:
#
#     content = ActionText::Content.new('<action-text-attachment content-type="image/jpeg" url="data:text/html,PAYLOAD" caption="Click"></action-text-attachment>')
#     content.to_markdown # => "\\[Click\\]"
```

actiontext/test/unit/markdown_conversion_test.rb
```diff
@@ -3,6 +3,26 @@
...
+  end
+
+  test "markdown_link with image: true returns image syntax" do
+    assert_equal "![photo](https://example.com/photo.png)", ActionText::MarkdownConversion.markdown_link("photo", "https://example.com/photo.png", image: true)
+  end
+
+  test "markdown_link with image: true and disallowed URI scheme returns escaped title" do
+    assert_equal "\\[Image\\]", ActionText::MarkdownConversion.markdown_link("Image", "data:text/html,PAYLOAD", image: true)
+  end
+
```
jeremy, line 25: Suggest `data:image/png` coverage:

```suggestion
  test "markdown_link with allowed data:image URI produces image link" do
    assert_equal "![photo](data:image/png;base64,abc)", ActionText::MarkdownConversion.markdown_link("photo", "data:image/png;base64,abc", image: true)
  end

  test "RemoteImage attachment with allowed data:image URI renders as image link" do
    assert_converted_to "![Image](data:image/png;base64,abc)", '<action-text-attachment content-type="image/png" url="data:image/png;base64,abc"></action-text-attachment>'
  end
```

jeremy (approved):

actiontext/lib/action_text/attachment.rb
flavorjones: Fixed in 82e80bbc.

actiontext/test/unit/markdown_conversion_test.rb
flavorjones: Added in 524d02f0. Both a `markdown_link` unit test and a `RemoteImage` integration test for `data:image/png;base64,...` URIs.

jeremy (approved):

PR #56916, opened 2026-03-02
Only allow known safe ASCII characters in HTML5 tag names in ActionView::Helpers::TagHelper

afurm:
### Motivation / Background

This Pull Request has been created because `TagHelper` currently accepts tag names containing `<` (for example, `tag("asdf-<")`), which produces invalid HTML instead of raising an error.

### Detail

This Pull Request changes tag-name validation in `ActionView::Helpers::TagHelper` to reject `<` in `ensure_valid_html5_tag_name`.

It also adds regression tests to cover both APIs:

- `tag("asdf-<")` now raises `ArgumentError`
- `tag.public_send(:"asdf-<")` now raises `ArgumentError`

### Additional information

Tests run:

- `bin/test test/template/tag_helper_test.rb`
- `bin/test test/template/output_safety_helper_test.rb`

### Checklist

* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. (Not updated here because this is a minor bug fix.)

### DIFF
--- a/actionview/lib/action_view/helpers/tag_helper.rb
+++ b/actionview/lib/action_view/helpers/tag_helper.rb
@@ -574,7 +574,7 @@ module ActionView
 
       private
         def ensure_valid_html5_tag_name(name)
-          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][^\s\/><]*\z/.match?(name)
+          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][a-zA-Z0-9:_-]*\z/.match?(name)
         end
         module_function :ensure_valid_html5_tag_name
 
--- a/actionview/lib/action_view/helpers/tag_helper.rb
+++ b/actionview/lib/action_view/helpers/tag_helper.rb
@@ -574,7 +574,7 @@ module ActionView
 
       private
         def ensure_valid_html5_tag_name(name)
-          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][^\s\/>]*\z/.match?(name)
+          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][^\s\/><]*\z/.match?(name)
         end
         module_function :ensure_valid_html5_tag_name
 
--- a/actionview/test/template/tag_helper_test.rb
+++ b/actionview/test/template/tag_helper_test.rb
@@ -151,12 +151,24 @@ class TagHelperTest < ActionView::TestCase
     end
   end
 
+  def test_tag_with_dangerous_name_containing_less_than
+    assert_raise(ArgumentError, 'expected "asdf-<" to be invalid') do
+      tag("asdf-<")
+    end
+  end
+
   def test_tag_builder_with_dangerous_name
     INVALID_TAG_CHARS.each_char do |char|
       tag_name = "asdf-#{char}".to_sym
       assert_raise(ArgumentError, "expected #{tag_name.inspect} to be invalid") do
         tag.public_send(tag_name)
       end
+    end
+  end
+
+  def test_tag_builder_with_dangerous_name_containing_less_than
+    assert_raise(ArgumentError, 'expected :"asdf-<" to be invalid') do
+      tag.public_send(:"asdf-<")
     end
   end
 


actionview/lib/action_view/helpers/tag_helper.rb
```diff
@@ -574,7 +574,7 @@ def escape_once(html)
 
       private
         def ensure_valid_html5_tag_name(name)
-          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][^\s\/>]*\z/.match?(name)
+          raise ArgumentError, "Invalid HTML5 tag name: #{name.inspect}" unless /\A[a-zA-Z][^\s\/><]*\z/.match?(name)
```
byroot: Hum, using an exclusion list makes me a bit uncomfortable. Reading a bit on this, it seems that at some point the spec only accepted ASCII characters, but apparently browsers accept UTF-8 ones now?

I'm not sure if there's any value in supporting anything outside the ASCII range, if you need that, use your own helper I guess?

paging @flavorjones for opinion.

afurm: I’ve updated this to use a positive allowlist instead of an exclusion list:
`/\A[a-zA-Z][a-zA-Z0-9:_-]*\z/`

So we now explicitly allow ASCII tag-name characters only (letter first, then alnum/`:`/`_`/`-`), and still reject the regression case with `<`. The existing regression tests for `asdf-<` still pass.

flavorjones:
Hmm, can you say more about "invalid HTML"? The WHATWG parser spec does not exclude the `<` character from the tag name state:

https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state

And the HTML5-compliant parser used by Nokogiri (gumbo) parses tag names like this fine. For example:

```ruby
#!/usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "nokogiri"
end

html = "<html><head></head><body><div></div><asdf-<></asdf-<></body></html>"

doc = Nokogiri::HTML5.parse(html)

doc
# => #(Document:0x168 {
#      name = "document",
#      children = [
#        #(Element:0x170 {
#          name = "html",
#          children = [
#            #(Element:0x178 { name = "head" }),
#            #(Element:0x180 {
#              name = "body",
#              children = [
#                #(Element:0x188 { name = "div" }),
#                #(Element:0x190 { name = "asdf-<" })]
#              })]
#          })]
#      })

doc.to_html
# => "<html><head></head><body><div></div><asdf-<></asdf-<></body></html>"
```

What I mean is: this is legal, well-formed HTML. Can you say more about the problem you're encountering that's leading you to open a PR?

afurm:
> Hmm, can you say more about "invalid HTML"? The WHATWG parser spec does not exclude the `<` character from the tag name state:
> 
> https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
> 
> And the HTML5-compliant parser used by Nokogiri (gumbo) parses tag names like this fine. For example:
> 
> ```ruby
> #!/usr/bin/env ruby
> 
> require "bundler/inline"
> 
> gemfile do
>   source "https://rubygems.org"
>   gem "nokogiri"
> end
> 
> html = "<html><head></head><body><div></div><asdf-<></asdf-<></body></html>"
> 
> doc = Nokogiri::HTML5.parse(html)
> 
> doc
> # => #(Document:0x168 {
> #      name = "document",
> #      children = [
> #        #(Element:0x170 {
> #          name = "html",
> #          children = [
> #            #(Element:0x178 { name = "head" }),
> #            #(Element:0x180 {
> #              name = "body",
> #              children = [
> #                #(Element:0x188 { name = "div" }),
> #                #(Element:0x190 { name = "asdf-<" })]
> #              })]
> #          })]
> #      })
> 
> doc.to_html
> # => "<html><head></head><body><div></div><asdf-<></asdf-<></body></html>"
> ```
> 
> What I mean is: this is legal, well-formed HTML. Can you say more about the problem you're encountering that's leading you to open a PR?

You're right, and I think my original premise was wrong.

I had assumed that allowing `<` in the tag name was generating invalid HTML, but after re-reading the spec I don't think that's accurate. The parser section is permissive here, as you pointed out, and the DOM Standard's element-name validation is aligned with that: for names starting with an ASCII letter, it only rejects ASCII whitespace, NUL, `/`, and `>`; `<` is allowed. It even says the intent is to allow names that are constructible by the HTML parser. See:
- [HTML parser, tag name state](https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state)
- [DOM Standard, valid element local name](https://dom.spec.whatwg.org/#valid-element-local-name)

I don't have a concrete interoperability or security issue beyond my mistaken assumption that this was malformed markup, so I don't think this PR is justified as-is. I'll close it unless there's a stronger reason to keep pursuing it.

PR #56918, opened 2026-03-03
[RF-Docs] [ci-skip] Update the Rails on Rack guide

ayushn21:
This PR updates the Rails on Rack guide.

Overview of changes:

* Several copy improvements.
* Link to initialization guide for description on how the `Rails::Server` starts up.
* Update information about the default middleware stack.
* Reformat structure of the "Internal Middleware Stack" section to match other guides like the "Configuration" guide.
* Add a section on adding custom middleware to the stack.
* Add a section on advanced usage describing the use of the Rack API inside Rails controllers.
* Removed the resources section and added links as part of the prose.

From audit:

- [x] The `bin/rails server` section could probably link to the [Initialization guide](https://guides.rubyonrails.org/initialization.html#rails-server-start) so we only need to keep that in-sync.
- [x] The "Internal Middleware Stack" section should be formatted similar to the configuration guide and/or the "Instrumentation" guide.
- [x] Some of the middleware have more extensive document documentation in the API docs.  (Linked each middleware component to its API docs page)
- [x] Sprockets::Rails::QuietAssets is now Propshaft::QuietAsset

### DIFF
--- a/guides/source/rails_on_rack.md
+++ b/guides/source/rails_on_rack.md
@@ -3,41 +3,149 @@
 Rails on Rack
 =============
 
-This guide covers Rails integration with Rack and interfacing with other Rack components.
+This guide covers Rails' integration with [Rack](https://en.wikipedia.org/wiki/Rack_(web_server_interface)). After reading this guide, you will know:
 
-After reading this guide, you will know:
-
-* How to use Rack Middlewares in your Rails applications.
-* Action Pack's internal Middleware stack.
-* How to define a custom Middleware stack.
+* What Rack is and why Rails uses it.
+* How Rails uses Rack middleware to build an application stack.
+* Action Pack's internal middleware stack.
+* How to configure and change the middleware stack.
+* The underlying Rack API exposed by Rails controllers.
 
 --------------------------------------------------------------------------------
-
-WARNING: This guide assumes a working knowledge of Rack protocol and Rack concepts such as middlewares, URL maps, and `Rack::Builder`.
 
 Introduction to Rack
 --------------------
 
-Rack provides a minimal, modular, and adaptable interface for developing web applications in Ruby. By wrapping HTTP requests and responses in the simplest way possible, it unifies and distills the API for web servers, web frameworks, and software in between (the so-called middleware) into a single method call.
+Rack provides a modular interface for developing web applications in Ruby. By wrapping HTTP requests and responses using a conventional structure, it unifies the API for web servers, web frameworks, and software in between (known as middleware) into a single method call.
 
-Explaining how Rack works is not really in the scope of this guide. In case you
-are not familiar with Rack's basics, you should check out the [Resources](#resources)
-section below.
+This allows Rack compliant web servers like [Puma](https://puma.io) or [Falcon](https://socketry.github.io/falcon/) to be interchangeably used with any Rack based web framework such as Rails.
+
+Before diving into how Rails integrates with Rack, let's look at Rack itself.
+
+### A Basic Rack Application
+
+A Rack app is an object which implements a `call` method. It is passed an [`env`](https://github.com/rack/rack/blob/main/SPEC.rdoc#the-request-environment) hash, known as the Rack environment.
+
+Here's an example of a barebones Rack app:
+
+```ruby
+class App
+  def call(env)
+    [200, { "content-type" => "text/plain" }, ["Hello World"]]
+  end
+end
+
+run App.new
+```
+
+When an HTTP request is made, the Rack-compliant web server parses it to create the `env` hash, and calls the application with `env`. The `call` method must return an array with exactly three elements, representing the HTTP response:
+
+1. The HTTP response code (`200` in the above example).
+2. A hash containing any HTTP response headers we wish to send.
+3. An enumerable object that yields strings, representing the response body.
+
+Rack applications are generally run using the web server's command line program, with the entry point for the application being stored in a `config.ru` file:
+
+```bash
+$ cat > config.ru << APP
+rack_app = lambda do |env|
+  [200, { "content-type" => "text/plain" }, ["Hello World"]]
+end
+run rack_app
+APP
+$ gem install puma
+$ puma
+```
+
+Your app should be available at <http://localhost:9292>.
+
+```bash
+$ curl localhost:9292
+Hello World
+```
+
+### Rack Middleware
+
+Rack applications can be wrapped using _middleware_ which may operate upon a request before it reaches the main application, and again after the application has returned a response to the request. Middleware is usually used for tasks like logging, caching, authentication, and measuring performance.
+
+A Rack middleware must have a `new` method that accepts the Rack app and any arguments used to configure the middleware. The `new` method must return a Rack application that responds to `call`. Typically, Rack middleware are classes, and each instance of the middleware wraps access to the related application:
+
+```ruby
+class MyMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    # Operations before the request hits the main application
+    # -------------------------------------------------------
+
+    # Propgate the request down the middleware stack
+    status, headers, body = @app.call(env)
+
+    # ---------------------------------------
+    # Operations after the request comes back
+
+    # Propogate the response up the middleware stack
+    [status, headers, body]
+  end
+end
+```
+
+Middleware can short-circuit the stack by skipping `@app.call` completely and returning a reponse by itself. This means the request never hits the main application or the remaining middleware in the stack. A middleware to authenticate a request might use this technique.
+
+```ruby
+class AuthenticateRequest
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if authenticated?(env["HTTP_AUTHORIZATION"])
+      @app.call(env)
+    else
+      [401, { "content-type" => "text/plain" }, ["Authentication failed"]]
+    end
+  end
+
+  def authenticated?(token)
+    # ...
+  end
+end
+```
+
+Middleware is added to a Rack app with `use`:
+
+```ruby
+class AuthenticateRequest
+  # ...
+end
+
+class App
+  def call(env)
+    [200, { "content-type" => "text/plain" }, ["Hello World"]]
+  end
+end
+
+use AuthenticateRequest
+run App.new
+```
+
+This DSL to construct Rack applications is provided by [`Rack::Builder`][]. For further information about Rack, consult the [Rack specification](https://rack.github.io/rack/main/SPEC_rdoc.html) and [Rack Website](https://rack.github.io/rack/).
+
+[`Rack::Builder`]: https://rack.github.io/rack/3.2/Rack/Builder.html
 
 Rails on Rack
 -------------
 
-### Rails Application's Rack Object
+### The Primary Rack Object
 
-`Rails.application` is the primary Rack application object of a Rails
-application. Any Rack compliant web server should be using
-`Rails.application` object to serve a Rails application.
+`Rails.application` is the *primary Rack application object* of a Rails
+application. A Rack compliant web server should use the `Rails.application` object to serve a Rails application.
 
-### `bin/rails server`
+### Starting the Rails Server
 
-`bin/rails server` does the basic job of creating a `Rack::Server` object and starting the web server.
-
-Here's how `bin/rails server` creates an instance of `Rack::Server`
+Rails subclasses `Rackup::Server` to create `Rails::Server`. `bin/rails server` instantiates a `Rails::Server` object and starts the web server.
 
 ```ruby
 Rails::Server.new.tap do |server|
@@ -47,43 +155,31 @@ Rails::Server.new.tap do |server|
 end
 ```
 
-The `Rails::Server` inherits from `Rack::Server` and calls the `Rack::Server#start` method this way:
+See the [initialization guide](initialization.html#rails-server-start) for further information on how the server starts up.
 
-```ruby
-class Server < ::Rack::Server
-  def start
-    # ...
-    super
-  end
-end
-```
 
-### Development and Auto-reloading
+Action Dispatch Middleware Stack
+--------------------------------
 
-Middlewares are loaded once and are not monitored for changes. You will have to restart the server for changes to be reflected in the running application.
+`ActionDispatch::MiddlewareStack` is Rails' equivalent of [`Rack::Builder`][]. It's built with more flexibility and features to meet Rails' requirements.
 
-Action Dispatcher Middleware Stack
-----------------------------------
+`Rails::Application` uses `ActionDispatch::MiddlewareStack` to combine internal and external middleware to build the stack which forms a complete Rack application using Rails.
 
-Many of Action Dispatcher's internal components are implemented as Rack middlewares. `Rails::Application` uses `ActionDispatch::MiddlewareStack` to combine various internal and external middlewares to form a complete Rails Rack application.
+### Inspecting the Middleware Stack
 
-NOTE: `ActionDispatch::MiddlewareStack` is Rails' equivalent of `Rack::Builder`,
-but is built for better flexibility and more features to meet Rails' requirements.
-
-### Inspecting Middleware Stack
-
-Rails has a handy command for inspecting the middleware stack in use:
+View the middleware stack by running:
 
 ```bash
 $ bin/rails middleware
 ```
 
-For a freshly generated Rails application, this might produce something like:
+Here's an example from a freshly generated Rails app:
 
 ```ruby
 use ActionDispatch::HostAuthorization
 use Rack::Sendfile
 use ActionDispatch::Static
+use Propshaft::Server
 use ActionDispatch::Executor
 use ActionDispatch::ServerTiming
 use ActiveSupport::Cache::Strategy::LocalCache::Middleware
@@ -91,7 +187,7 @@ use Rack::Runtime
 use Rack::MethodOverride
 use ActionDispatch::RequestId
 use ActionDispatch::RemoteIp
-use Sprockets::Rails::QuietAssets
+use Propshaft::QuietAssets
 use Rails::Rack::Logger
 use ActionDispatch::ShowExceptions
 use WebConsole::Middleware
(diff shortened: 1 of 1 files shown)


guides/source/rails_on_rack.md
```diff
@@ -3,41 +3,36 @@
...
+* How Rails uses Rack middleware to build an application stack.
+* Action Pack's internal middleware stack.
+* How to configure and change the middleware stack.
+* How to insert your own middleware into the stack.
+* The underlying Rack API exposed by Rails controllers.
 
 --------------------------------------------------------------------------------
 
-WARNING: This guide assumes a working knowledge of Rack protocol and Rack concepts such as middlewares, URL maps, and `Rack::Builder`.
+NOTE: This guide assumes an understanding of the Rack specification and associated concepts such as a Rack application, middleware stack, and `Rack::Builder`. See the [Resources](#resources) section below for further information.
```
bhumi1102: How about linking to the resource here directly so the reader doesn't need to jump all the way to the bottom? "See link1 or link2 for more."

guides/source/rails_on_rack.md
```diff
@@ -47,51 +42,39 @@ Rails::Server.new.tap do |server|
...
 
-NOTE: `ActionDispatch::MiddlewareStack` is Rails' equivalent of `Rack::Builder`,
-but is built for better flexibility and more features to meet Rails' requirements.
+`Rails::Application` uses `ActionDispatch::MiddlewareStack` to combine internal and external middleware to build the stack which forms a complete Rack application using Rails.
 
-### Inspecting Middleware Stack
+### Inspecting the Middleware Stack
 
-Rails has a handy command for inspecting the middleware stack in use:
+View the middleware stack in use by running:
```
bhumi1102: nit
```suggestion
View the middleware stack by running:
```

guides/source/rails_on_rack.md
```diff
@@ -47,51 +42,39 @@ Rails::Server.new.tap do |server|
...
 
 ```bash
 $ bin/rails middleware
 ```
 
-For a freshly generated Rails application, this might produce something like:
+Here's an example from a freshly generated Rails 8 app:
```
bhumi1102: No need to mention the version. The guides are meant to match "main".

```suggestion
Here's an example from a freshly generated Rails app:
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +154,315 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+```ruby
+class HomeController
+  def index
+    self.response = [200, {}, ["I'm Home!"]]
+  end
+end
+```
+
+Resources
+---------
```
bhumi1102: Is this enough for a new top level section? probably can roll the links into sentence at the top.

bhumi1102 (commented):
Internal review

guides/source/rails_on_rack.md
ayushn21: Hmm, yeah I like this idea in general, but I'd still like to have the 3 resources as a list for clarity. This text is in a `NOTE` block which as far as I can tell, can't contain multiple block elements, so we can't have a list in it?

If that's the case then I'm inclined to leave the resources where they are, but if there's a way we can show a list within the `NOTE` block, then happy to have it all in one place.

guides/source/rails_on_rack.md
ayushn21: Nice catch :)

guides/source/rails_on_rack.md
ayushn21: Done :)

guides/source/rails_on_rack.md
```diff
@@ -169,178 +154,315 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+        env["action_dispatch.request.parameters"].delete("locale")
+      end
+
+      @app.call(env)
+    end
+  end
+end
+```
+
+### Adding custom middleware to the stack
```
p8: [For headings](https://guides.rubyonrails.org/ruby_on_rails_guides_guidelines.html#headings) we capitalize all words except for prepositions, conjunctions, internal articles, and forms of the verb "to be" 
```suggestion
### Adding Custom Middleware to the stack
```

ayushn21: Ah damn missed this one, fixed now!

guides/source/rails_on_rack.md
ayushn21: Changed my mind about this and rewrote it as a paragraph with inline links, and removed the `Resources` section. Thanks!

guides/source/rails_on_rack.md
ayushn21: Removed now and added to the start of the guide!

guides/source/rails_on_rack.md
```diff
@@ -3,95 +3,192 @@
...
 
 --------------------------------------------------------------------------------
 
-WARNING: This guide assumes a working knowledge of Rack protocol and Rack concepts such as middlewares, URL maps, and `Rack::Builder`.
-
 Introduction to Rack
 --------------------
 
-Rack provides a minimal, modular, and adaptable interface for developing web applications in Ruby. By wrapping HTTP requests and responses in the simplest way possible, it unifies and distills the API for web servers, web frameworks, and software in between (the so-called middleware) into a single method call.
+Rack provides a minimal, modular, and adaptable interface for developing web applications in Ruby. By wrapping HTTP requests and responses using a conventional structure, it unifies the API for web servers, web frameworks, and software in between (known as middleware) into a single method call.
```
p8: Nitpick. I know this is copied from the Rack README but can we maybe remove some adjectives here to make it less sales-y? Is it still `minimal`? And I'm not sure what `adaptable` means in the context of the guide:
```suggestion
Rack provides a modular interface for developing web applications in Ruby. By wrapping HTTP requests and responses using a conventional structure, it unifies the API for web servers, web frameworks, and software in between (known as middleware) into a single method call.
```

guides/source/rails_on_rack.md
```diff
@@ -3,95 +3,192 @@
...
+
+[`Rack::Builder`]: https://rack.github.io/rack/3.2/Rack/Builder.html
+
+Rails on Rack
+-------------
+
+### The Primary Rack Object
+
+`Rails.application` is the *primary Rack application object* of a Rails
+application. A Rack compliant web server should use `Rails.application` object to serve a Rails application.
```
p8:
```suggestion
application. A Rack compliant web server should use the `Rails.application` object to serve a Rails application.
```

guides/source/rails_on_rack.md
```diff
@@ -3,95 +3,192 @@
...
+
+```ruby
+Rails::Server.new.tap do |server|
+  require APP_PATH
+  Dir.chdir(Rails.application.root)
+  server.start
+end
+```
+
+See the [initilization guide](initialization.html#rails-server-start) for further information on how the server starts up.
```
p8:
```suggestion
See the [initialization guide](initialization.html#rails-server-start) for further information on how the server starts up.
```

guides/source/rails_on_rack.md
```diff
@@ -111,49 +208,51 @@ use Rack::TempfileReaper
 run MyApp::Application.routes
 ```
 
-The default middlewares shown here (and some others) are each summarized in the [Internal Middlewares](#internal-middleware-stack) section, below.
+The [Internal Middleware Stack](#internal-middleware-stack) section below summarizes the default middleware components depicated above.
```
p8: Should this be `depicted` instead?
```suggestion
The [Internal Middleware Stack](#internal-middleware-stack) section below summarizes the default middleware components depicted above.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
 
 ### Internal Middleware Stack
 
-Much of Action Controller's functionality is implemented as Middlewares. The following list explains the purpose of each of them:
+Much of Action Controller's functionality is implemented as middleware. The following list explains the purpose of each of them:
+
+#### [`ActionDispatch::HostAuthorization`][]
 
-**`ActionDispatch::HostAuthorization`**
+Prevents DNS rebinding attacks by restricting the hosts to which a request can be sent. See the [configuration guide](configuring.html#actiondispatch-hostauthorization) for configuration instructions.
```
p8: I think it's great that we link to the middlewares.
All headers are currently converted to anchor tags automatically.
Linking to somewhere else in the header breaks this convention.
What do you think about linking in the text instead?
Although I find starting the paragraph with the title a bit meh, I think it improves the opening sentence a bit because it can be read/copied without the context of the title.

```suggestion
#### `ActionDispatch::HostAuthorization`

[`ActionDispatch::HostAuthorization`][] prevents DNS rebinding attacks by restricting the hosts to which a request can be sent. See the [configuration guide](configuring.html#actiondispatch-hostauthorization) for configuration instructions.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
-```ruby
-# config/application.rb
-config.middleware.delete! ActionDispatch::Executor
-```
+The middleware stack is loaded once and isn't monitored for changes. Restart your server after making changes to your middleware stack.
 
 ### Internal Middleware Stack
 
-Much of Action Controller's functionality is implemented as Middlewares. The following list explains the purpose of each of them:
+Much of Action Controller's functionality is implemented as middleware. The following list explains the purpose of each of them:
```
p8, line 293: This list is now sorted by how middleware is loaded in a new Rails application, which seems somewhat arbitrary and is suspected to change. In the "Configuring" guide and the "Instrumentation" guide we [sort alphabetically](https://guides.rubyonrails.org/active_support_instrumentation.html#rails-framework-hooks)
I think we should sort alphabetically here as well.

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+Prevents DNS rebinding attacks by restricting the hosts to which a request can be sent. See the [configuration guide](configuring.html#actiondispatch-hostauthorization) for configuration instructions.
 
-* Guards from DNS rebinding attacks by explicitly permitting the hosts a request can be sent to. See the [configuration guide](configuring.html#actiondispatch-hostauthorization) for configuration instructions.
+[`ActionDispatch::HostAuthorization`]: https://api.rubyonrails.org/classes/ActionDispatch/HostAuthorization.html
 
-**`Rack::Sendfile`**
+#### [`Rack::Sendfile`][]
 
-* Sets server specific X-Sendfile header. Configure this via [`config.action_dispatch.x_sendfile_header`][] option.
+Sets server specific `X-Sendfile header`. Configure this via [`config.action_dispatch.x_sendfile_header`][] option.
```
p8: I think we can expand a little here:
```suggestion
Sets server specific `X-Sendfile header`. This is useful for accelerated file sending if you use a reverse proxy server like Apache or Nginx. For example it can be set to 'X-Sendfile' for Apache. Configure this via [`config.action_dispatch.x_sendfile_header`][] option.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
-* Used to serve static files from the public directory. Disabled if [`config.public_file_server.enabled`][] is `false`.
+Serves static files from the `public` folder. Disabled when [`config.public_file_server.enabled`][] is `false`.
 
+[`ActionDispatch::Static`]: https://api.rubyonrails.org/classes/ActionDispatch/Static.html
 [`config.public_file_server.enabled`]: configuring.html#config-public-file-server-enabled
 
-**`Rack::Lock`**
+#### [`Rack::Lock`][]
+
+Sets `env["rack.multithread"]` to `false` and wraps the application within a mutex.
```
p8: I think the Rack::Lock documentation has a better description that explains the use-case and less the implementation:
```suggestion
Locks every request inside a mutex, so that every request will effectively be executed synchronously.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+Sets a [`Server-Timing`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) header containing performance metrics for the request.
 
-* Sets a [`Server-Timing`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) header containing performance metrics for the request.
+[`ActionDispatch::ServerTiming`]: https://api.rubyonrails.org/classes/ActionDispatch/ServerTiming.html
 
-**`ActiveSupport::Cache::Strategy::LocalCache::Middleware`**
+#### [`ActiveSupport::Cache::Strategy::LocalCache::Middleware`][]
 
-* Used for memory caching. This cache is not thread safe.
+An in-memory cache. This cache is not thread safe.
```
p8: The middleware isn't the cache, but a middleware wrapper for the cache (also copied more information from the API docs):
```suggestion
Middleware for the in-memory local cache. This cache is not thread safe and is intended only for serving as a temporary memory cache for a single thread.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+Allows the method to be overridden if `params[:_method]` is set. This is how Rails supports `PUT`, `PATCH`, and `DELETE` HTTP methods since they are not browser native.
 
-**`ActionDispatch::RemoteIp`**
+[`Rack::MethodOverride`]: https://rack.github.io/rack/3.2/Rack/MethodOverride.html
 
-* Checks for IP spoofing attacks.
+#### [`ActionDispatch::RequestId`][]
 
-**`Sprockets::Rails::QuietAssets`**
+Makes a unique `X-Request-Id` header available to the response and enables the `ActionDispatch::Request#request_id` method.
```
p8: I think the header is set on the request.
I've also added some copy from the API docs to describe what it's used for:
```suggestion
Makes a unique `X-Request-Id` header available to the request and enables the `ActionDispatch::Request#request_id` method.

The unique request id can be used to trace a request end-to-end and would typically end up being part of log files from multiple pieces of the stack.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+Responsible for logging exceptions and showing a debugging page if the request is local.
 
-**`ActionDispatch::Cookies`**
+[`ActionDispatch::DebugExceptions`]: https://api.rubyonrails.org/classes/ActionDispatch/DebugExceptions.html
 
-* Sets cookies for the request.
+#### [`ActionDispatch::ActionableExceptions`][]
 
-**`ActionDispatch::Session::CookieStore`**
+Provides a way to dispatch actions from Rails' error pages.
```
p8: Similar to `DebugExceptions` we should probably mention this only applies to local requests:
```suggestion
Provides a way to dispatch actions from Rails' error pages if the request is local.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+
+#### [`ActionDispatch::Callbacks`][]
+
+Provides callbacks to be executed before and after dispatching the request.
+
+[`ActionDispatch::Callbacks`]: https://api.rubyonrails.org/classes/ActionDispatch/Callbacks.html
+
+#### [`ActiveRecord::Migration::CheckPending`][]
+
+Checks pending migrations and raises `ActiveRecord::PendingMigrationError` if any migrations are pending.
```
p8: I think  we should mention how this can be configured (the link to the configuration doesn't work as it seems this setting is not properlty document in the configuration guide)
```suggestion
Checks pending migrations and raises `ActiveRecord::PendingMigrationError` if any migrations are pending if [`config.action_dispatch.x_sendfile_header`][] is set to `:page_load`.

[`config.action_dispatch.x_sendfile_header`]: configuring.html#config-action-record-migration-error
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+
+[`ActiveRecord::Migration::CheckPending`]: https://api.rubyonrails.org/classes/ActiveRecord/Migration/CheckPending.html
+
+#### [`ActionDispatch::Cookies`][]
+
+Sets cookies for the request.
```
p8: This middleware also writes cookie data on the response:
```suggestion
Reads cookie data from the request and writes cookie data on the response.
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +268,308 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+
+#### [`Rack::Head`][]
+
+Returns an empty body for all `HEAD` requests. It leaves all other requests unchanged.
+
+[`Rack::Head`]: https://rack.github.io/rack/3.2/Rack/Head.html
+
+#### [`Rack::ConditionalGet`][]
+
+Adds support for "Conditional `GET`" so that server responds with nothing if the page wasn't changed.
```
p8: We can expand on this a bit as well (mostly copied from the Rack docs):
```suggestion
Enables "Conditional `GET`" requests using if-none-match and if-modified-since. If the requested page wasn't changed returns a 304 Not Modified and an empty body.
```

p8 (commented):
@ayushn21 Some comments...

guides/source/rails_on_rack.md
ayushn21: Yeah good shout

guides/source/rails_on_rack.md
ayushn21: Yeah fair enough ... I can't think of a better way than starting the sentence with the name of the middleware either so will do that

ayushn21:
All done @p8  .... let me know if I missed anything ... thanks :)

guides/source/rails_on_rack.md
```diff
@@ -169,178 +267,312 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
 
-* Sets up the flash keys. Only available if [`config.session_store`][] is set to a value.
+TIP: You can use any of the above middleware in a custom Rack stack.
 
-[`config.session_store`]: configuring.html#config-session-store
+Custom Middleware
+-----------------
 
-**`ActionDispatch::ContentSecurityPolicy::Middleware`**
+Your Rails app can implement and include custom middleware as per your requirements.
```
p8: This sentence is pretty formal. 
Maybe something like:
```suggestion
You can create your own middleware and include it in your Rails app. 
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +267,312 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+[`Rails::Rack::Logger`][] notifies the logs that the request has begun. After the request is complete, flushes all the logs.
 
-**`ActionDispatch::Flash`**
+[`Rails::Rack::Logger`]: https://api.rubyonrails.org/classes/Rails/Rack/Logger.html
 
-* Sets up the flash keys. Only available if [`config.session_store`][] is set to a value.
+TIP: You can use any of the above middleware in a custom Rack stack.
 
-[`config.session_store`]: configuring.html#config-session-store
+Custom Middleware
```
p8, line 466: This section comes after adding middleware. i think the adding middleware section makes more sense if we add our own custom middleware. Maybe those sections should be combined? First create the middleware, then show how it could bee added?

guides/source/rails_on_rack.md
```diff
@@ -169,178 +267,312 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+Your Rails app can implement and include custom middleware as per your requirements.
 
-* Provides a DSL to configure a Content-Security-Policy header.
+Custom middleware files should be placed in the `lib/` folder and `require`d manually since middleware is not auto-reloaded.
 
-**`Rack::Head`**
+### Creating Middleware
 
-* Returns an empty body for all HEAD requests. It leaves all other requests unchanged.
+Creating a new `lib/middleware/` folder and excluding it from the autoload path is recommended.
```
p8: This sentence and example seem to belong with the previous paragrph which explains why the folder shouldn’t be reloaded.

Also the order seems somewhat backwards to me. To me it seems nicer to start with the actual middleware, then follow with the “boring” autoload parts.

guides/source/rails_on_rack.md
```diff
@@ -169,178 +267,312 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
+```ruby
+# config/initializers/extract_locale.rb
 
-* [Railscast on Rack Middlewares](http://railscasts.com/episodes/151-rack-middleware)
+require "#{Rails.root.join("lib", "middleware", "extract_locale")}"
+
+Rails.application.config.middleware.use RackMiddleware::ExtractLocale
+```
+
+Advanced Usage
```
p8: Maybe we can use a more descriptive title here?
For example:
```suggestion
Calling Rack from your Application
```

Or:
```suggestion
Accessing Rack in Rails
```

It’s also possible to use Rack inside your routes. There is a section on this in the Routing guide. Maybe add that as well?

p8 (commented):
Hi @ayushn21  Some suggestions for the final 2 sections.

guides/source/rails_on_rack.md
ayushn21: It's a tricky one but I think we should leave it as is. Anecdotally, folks are far more likely to add pre-built middleware from Rack or other libraries than write their own. 

In my mind, custom middleware is an advanced topic so I prefer to leave it where it is, further down in the guide.

guides/source/rails_on_rack.md
ayushn21: Yeah good call, amended it

guides/source/rails_on_rack.md
ayushn21: Changed to "Accessing Rack Internals in Rails". Hmmm, let me have a think about the routing stuff ... I'm not keen to duplicate content so might just link to it.

ayushn21:
Thanks @p8  ... all done :)

guides/source/rails_on_rack.md
```diff
@@ -3,95 +3,191 @@
...
+  server.start
+end
+```
+
+See the [initialization guide](initialization.html#rails-server-start) for further information on how the server starts up.
 
-Middlewares are loaded once and are not monitored for changes. You will have to restart the server for changes to be reflected in the running application.
 
 Action Dispatcher Middleware Stack
 ----------------------------------
```
p8:
```suggestion
Action Dispatch Middleware Stack
--------------------------------
```

guides/source/rails_on_rack.md
```diff
@@ -169,178 +267,316 @@ config.middleware.move_before Lifo::ShowExceptions, ActionDispatch::ShowExceptio
...
-If you want an error to be raised when you try to delete a non-existent item, use `delete!` instead.
+### Reloading the Middleware Stack
 
-```ruby
-# config/application.rb
-config.middleware.delete! ActionDispatch::Executor
-```
+The middleware stack is loaded once and isn't monitored for changes. Restart your server after making changes to your middleware stack.
 
 ### Internal Middleware Stack
```
p8, line 291: I find this a bit hard to read now with all the red:
Image

What do you tink about making "Interal Middleware Stack" a top section (so the headers become black) instead?
We do this for the Caching guide as well and I think it's a bit easier on the eyes:

Image

https://edgeguides.rubyonrails.org/caching_with_rails.html#other-cache-stores

p8 (commented):
Sorry @ayushn21 I didn't submit my final 2 comments.

guides/source/rails_on_rack.md
ayushn21: Ahhh I was reading the guides in dark mode so I never saw this issue. I really don't want to make it a top-level section for UI reasons. In the logical document structure, it should be a sub-section under "Action Dispatch Middleware Stack".

In dark mode, all the headings are white. I wonder if there's a case to replicate that in light mode and have all headings black? 

If the red headings are causing issues in multiple places, then I would say that's the problem, and we shouldn't restructure our documents to fit this design decision.

p8: Hmm, yeah I agree we shouldn't restructure the guide to fit design decisions.
Let's keep this as it was.

PR #56919, opened 2026-03-03
Add `Gemfile` specifically for the Rails Guides

ayushn21:
### Motivation / Background

The rake guides:generate command failed for me because I don't have `mysql` installed locally. The gem is part of the bundle and `bundle exec` is used under the hood to run `dartsass` to generate the stylesheets for the guides which loads the bundle for all of Rails. I shouldn't need a DB gem setup just to generate the guides.

I feel the Rails Guides should be logically isolated from the rest of the framework.

### Detail

Add a Gemfile in the guides directory which only defines the gems used to generate the Rails Guides. This allows the use of `bundle exec` without the baggage of the entire framework's requirements.

### Checklist

### DIFF
--- a/guides/Gemfile
+++ b/guides/Gemfile
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "dartsass-rails"
+gem "importmap-rails", ">= 1.2.3"
+
+gem "sdoc", "~> 2.6.4"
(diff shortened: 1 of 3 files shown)


zzak:
As we discussed you can use the env var `BUNDLE_WITHOUT=db` or even `BUNDLE_WITHOUT=db:job:cable:storage` to exclude even more stuff.

PR #56921, opened 2026-03-03
Remove sentence about deleting output dir from guides readme

ayushn21:
### Motivation / Background

The `output` directory for the Rails Guides needs to be deleted before the HTML guides are generated. This PR implements a Rake task to do this so it's quicker for the developer.

EDIT: The `RailsGuides::Generator` cleans the directory automatically before generating the guides, so this is now a documentation change only.

### Detail

Update the Readme to remove the sentence about deleting the `output` directory before generating the Rails Guides.

### Checklist

### DIFF
--- a/guides/README.md
+++ b/guides/README.md
@@ -10,7 +10,7 @@ The editing files for the Guides rebuild reside in `stylesrc` and use SCSS to im
 
 ## Building the Guides in Development
 
-To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. If you make changes to the HTML or ERB, you'll need to remove the "output" directory before running this command. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.
+To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.
 
 ## FAQ
 


guides/Rakefile
```diff
@@ -36,6 +36,12 @@ namespace :guides do
   desc "Lint guides, using `mdl` and check for broken links"
   task lint: ["lint:check_links", "lint:mdl"]
 
+  desc "Deletes the `output` directory"
+  task :clean do
+    output_dir = File.join(File.expand_path(__dir__), "output")
+    FileUtils.remove_dir(output_dir)
```
p8: Can you move this logic to `RailsGuides::Generator`, add a method to access the generator, and call clean on that? This makes sure the only the generator know the `output_dir`.

```suggestion
    rails_generator.clean
```

p8 (changes requested):

guides/Rakefile
ayushn21: Ahh so to action this, I've just noticed that the `Generator` cleans up the assets automatically before generating the guides. It's in the `cleanup_assets` method. Hence it doesn't need to be done manually. I've updated the PR now to be a documentation change only which removes the sentence about manually deleting the `output` folder. Hope that's ok?

ayushn21:
Hey @p8, could you take a peek at this again when you get a moment please?

guides/README.md
```diff
@@ -10,7 +10,7 @@ The editing files for the Guides rebuild reside in `stylesrc` and use SCSS to im
 
 ## Building the Guides in Development
 
-To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. If you make changes to the HTML or ERB, you'll need to remove the "output" directory before running this command. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.
+To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.
```
p8, line 16: Maybe show an example here, similar to some of the other README's?
```suggestion
To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.

  $ bundle exec rake guides:generate
  
```

We also document the rake commands in more detail in the guidelines, but those are at the bottom so easily overlooked?

ayushn21: Good shout, so I grabbed the command from the "Contributing" guide: https://github.com/rails/rails/blob/2e11aac5950afcf929aabecb6120e95bc26f3609/guides/source/contributing_to_ruby_on_rails.md?plain=1#L169.

The `BUNDLE_ONLY=default:doc` would've saved me some pain when I first started on this project and couldn't build the msql gem because I didn't have it installed.

Also added links to the Rails Guides guidelines, and added the same `BUNDLE_ONLY` prefix there too.

What do you think?

PR #56922, opened 2026-03-03
Add a development server to view the Rails Guides

ayushn21:
### Motivation / Background

The HTML Rails Guides can only be viewed by opening the HTML file. While this is fine, serving them over HTTP has better ergonomics and better simulates how they will be access in production.

### Detail

This Pull Request adds a basic `config.ru` to serve the Rails Guides using a `Rack::Static` server. This allows the developer to access the guides at `localhost:9292` during development.

### Checklist

### DIFF
--- a/guides/README.md
+++ b/guides/README.md
@@ -12,6 +12,12 @@ The editing files for the Guides rebuild reside in `stylesrc` and use SCSS to im
 
 To generate new guides into static files, type `rake guides:generate` from inside the `guides` folder. If you make changes to the HTML or ERB, you'll need to remove the "output" directory before running this command. The master SCSS files (style.scss, highlight.scss) will compile as part of this process.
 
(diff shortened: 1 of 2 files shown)


p8:
If we're not going to have a separate Gemfile for the guides, I don't think we want these dependencies to end up in the main Gemfile.

ayushn21:
There are no extra dependencies required for this. Rails already includes `rack` and that's all we need.

p8:
A

> There are no extra dependencies required for this. Rails already includes `rack` and that's all we need.

Ah sorry, I hadn't looked at the actual diff 😞

ayushn21:
Hey @p8 ... I've been using this workflow in my work with the Rails guides for the last few months and I really like the ergonomics. 

Is this something we can merge? Otherwise happy to close and use it just locally.

PR #56928, opened 2026-03-04
Update comment in default application.css for new Rails apps

ayushn21:
### Motivation / Background

This PR follows on from https://github.com/rails/rails/pull/53021. New Rails 8 apps use Propshaft by default which doesn't (as far as I can tell) treat `application.css` as a manifest. It simply fingerprints the file and doesn't do any compilation, hence the existing comment is misleading.

### Detail

The PR updates the comment to state that stylesheets are linked individually, and signposts `cssbundling-rails` if the developer would like to transpile/bundle their CSS.

If my understanding is incorrect, or the comment could be improved, please let me know!

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/app/assets/stylesheets/application.css.tt
+++ b/railties/lib/rails/generators/rails/app/templates/app/assets/stylesheets/application.css.tt
@@ -1,10 +1,8 @@
 /*
- * This is a manifest file that'll be compiled into application.css.
+ * Propshaft serves assets efficiently without transpiling or bundling. By default, each
+ * stylesheet is linked individually hence this file should be used for global
+ * styles only. Consider organizing styles into separate files for maintainability.
  *
- * With Propshaft, assets are served efficiently without preprocessing steps. You can still include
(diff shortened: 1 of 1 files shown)


railties/lib/rails/generators/rails/app/templates/app/assets/stylesheets/application.css.tt
```diff
@@ -1,10 +1,8 @@
 /*
- * This is a manifest file that'll be compiled into application.css.
+ * Propshaft serves assets efficiently without preprocessing. By default, each
```
p8: As we mention transpiling and bundling later on, I think it's probably clearer if use the same terms here.
It also make it a bit more explicit what we mean by preprocessing:
```suggestion
 * Propshaft serves assets efficiently without transpiling or bundling. By default, each
```

ayushn21: Yeah I prefer your wording, updated! And also rebased on `main` so this should be good to go.

PR #56929, opened 2026-03-04
ActionView::Base.precompile_templates compiles templates with strict locals at boot

joelhawksley:
### Motivation / Background

_This PR shamelessly builds on extensive work by @jhawthorn and others :heart:_

Currently, Rails templates are loaded from disk and compiled into a Ruby method the first time they are rendered. This process happens on every worker on forking web servers, every time a worker is spawned (such as on a new deploy).

This optimization aims to improve cold render times and to allow more memory to be shared via CoW on forking web servers by precompiling templates that use strict locals at boot time.

Assuming we want to move forward with this optimization, I plan to work to incorporate https://github.com/jhawthorn/actionview_precompiler into core as well.

For now, I’ve put this behind a configuration flag that defaults to false. I’d recommend that we default it to true for production environments starting in the next major release.

### Benchmark

1000 ERB templates (~100 lines each) with strict locals, Puma with 5 workers (clustered mode), Docker (Linux), Ruby 4.0. PSS (Proportional Set Size) accounts for shared memory pages between forked workers:

| | `precompile_templates=false` | `precompile_templates=true` |
|---|---|---|
| Templates compiled before fork | 0 | 1000 |
| Templates compiled after fork (per worker) | 1000 | 0 |
| Total PSS before rendering (5 workers) | 102 MB | 127 MB |
| Total PSS after rendering (5 workers) | 361 MB | 283 MB |
| **PSS increase from rendering** | **259 MB** | **156 MB** |

With `precompile_templates` enabled, compiled template bytecode is shared across all workers via copy-on-write, reducing post-render memory growth by **40%** (259 MB → 156 MB across 5 workers) and total memory after rendering by **22%** (361 MB → 283 MB).

### Checklist

### DIFF
--- a/actionview/lib/action_view/template.rb
+++ b/actionview/lib/action_view/template.rb
@@ -189,6 +189,38 @@ module ActionView
           const_set(:Types, implementation)
         end
       end
+
+      def precompile!
+        return unless ActionView::Base.precompile_templates
+
+        details = {
+          locale: I18n.available_locales,
+          handlers: ActionView::Template::Handlers.extensions,
+          formats: ActionView::Template::Types.symbols,
+          variants: :any
+        }
+        key = ActionView::LookupContext::DetailsKey.details_cache_key(details)
+
+        ActionController::Base.view_paths.each do |resolver|
+          next unless resolver.is_a?(ActionView::FileSystemResolver)
+
+          resolver.all_template_paths.each do |template_path|
+            templates = resolver.find_all(
+              template_path.name,
+              template_path.prefix,
+              template_path.partial?,
+              details,
+              key,
+              []
+            )
+
+            templates.each do |template|
+              next unless template.strict_locals?
+              template.send(:compile!, ActionController::Base.view_context_class)
+            end
+          end
+        end
+      end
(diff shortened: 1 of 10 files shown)


actionview/lib/action_view/template.rb
```diff
@@ -189,6 +189,38 @@ def mime_types_implementation=(implementation)
...
+              template_path.name,
+              template_path.prefix,
+              template_path.partial?,
+              details,
+              key,
+              []
+            )
+
+            templates.each do |template|
+              next unless template.strict_locals?
```
joelhawksley, line 218: I'm wondering if it might be faster to glob the templates that have the strict locals regex vs. iterating over every templates through the resolver here. Thoughts?

moberegger:
Ah bummer! I was looking forward to this one!

joelhawksley:
@moberegger I'll likely bring it back in a different form, as a backup to a more automatic approach that doesn't require strict locals.

moberegger:
@joelhawksley Oh awesome! I'll keep an eye out!

PR #56932, opened 2026-03-04
Fix dbconsole NotImplementedError message

eglitobias:
### Motivation / Background

This Pull Request has been created because PR #54719 changed AbstractAdapter.dbconsole to interpolate self.class in a class method, which results in the misleading message “Class should define…”. Using self (the adapter class) restores the correct class name in the error.

### Detail
This Pull Request changes the NotImplementedError message in ActiveRecord::ConnectionAdapters::AbstractAdapter.dbconsole to use #{self} instead of #{self.class} (since self is the adapter class in def self.dbconsole).

### Before
```
ActiveRecord::ConnectionAdapters::AbstractAdapter.dbconsole(nil)
NotImplementedError: Class should define `dbconsole` ...
```

### After
```
ActiveRecord::ConnectionAdapters::AbstractAdapter.dbconsole(nil)
NotImplementedError: ActiveRecord::ConnectionAdapters::AbstractAdapter should define `dbconsole` ...
```

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
@@ -139,7 +139,7 @@ module ActiveRecord
 
       # Opens a database console session.
       def self.dbconsole(config, options = {})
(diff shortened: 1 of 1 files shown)


skipkayhil:
Nice find, thanks!

PR #56936, opened 2026-03-04
Fix nil params serialization in ActionController::TestCase

### LINKED ISSUE #56934: `nil` values passed to params not converted to empty strings in ActionController::TestCase
Before Rails 8.1, params with `nil` values passed to the request helper methods (`get`, `post`, etc) were converted to empty strings values. In Rails 8.1 they remain `nil`:

```ruby
get :index, params: { foo: nil }
# the controller would receive:
#    { foo: "" }
#  in Rails 8.1 these return `nil` instead.
#    { foo: nil }
```

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails", branch: "8-1-stable"
end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

end
Rails.application.initialize!

  get "/", to: "test#index"
end

  def index
    render plain: ::JSON.dump(params.to_unsafe_h)
  end
end

class BugTest < ActionController::TestCase
  def setup
    super
    @controller = TestController.new
    @routes = ActionDispatch::Routing::RouteSet.new.tap do |r|
      r.draw do
        ActionDispatch.deprecator.silence do
          get ":controller(/:action(/:id))"
        end
      end
    end
  end

  def test_nil_params
    get :index, params: { foo: nil }
    parsed_params = JSON.parse(@response.body)
    # this would not fail in Rails 8.0
    assert_equal "", parsed_params["foo"]
  end
end
```

Saidbek:
### Motivation / Background

This PR has been created because `ActionController::TestCase` request helpers no longer convert `nil` param values to empty strings after `NilClass#to_query` changed in 8.1, causing a behavioral regression from 8.0. https://github.com/rails/rails/commit/2952e9a85d0a8a7f5057ef1f8aded9b7c0905327

Fixes #56934.

### Detail

`ActionController::TestCase::TestRequest` now normalizes non-path params by converting `nil` values to `""` before query/form encoding. This restores previous functional test behavior for `get`, `post`, and other helper methods using url-encoded params.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/test_case.rb
+++ b/actionpack/lib/action_controller/test_case.rb
@@ -105,12 +105,14 @@ module ActionController
 
       if get?
         if query_string.blank?
-          self.query_string = non_path_parameters.to_query
+          self.query_string = normalize_non_path_parameters(non_path_parameters).to_query
         end
       else
-        if ENCODER.should_multipart?(non_path_parameters)
+        normalized_non_path_parameters = normalize_non_path_parameters(non_path_parameters)
+
+        if ENCODER.should_multipart?(normalized_non_path_parameters)
           self.content_type = ENCODER.content_type
-          data = ENCODER.build_multipart non_path_parameters
+          data = ENCODER.build_multipart normalized_non_path_parameters
         else
           fetch_header("CONTENT_TYPE") do |k|
             set_header k, "application/x-www-form-urlencoded"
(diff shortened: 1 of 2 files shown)


p8:
Thanks for the PR @Saidbek. 
It seems the original change is correct behaviour, as this fixes roundtripping: nil remains nil after serializing/deserializing to a query parameter

PR #56940, opened 2026-03-06
Match parent status codes for exception subclasses

thehale:
### Motivation / Background

Rails maps certain [predefined exceptions](https://guides.rubyonrails.org/configuring.html#config-action-dispatch-rescue-responses) to HTTP status codes. For example, raising `ActiveRecord::RecordNotFound` from a controller results in a 404 Not Found, while unregistered exceptions result in a 500 Server Error.

What feels surprising to me is that subclasses of a predefined exception do not have the same behavior as their parent classes. 

```ruby
class MyCustomNotFound < ActiveRecord::RecordNotFound
end

# inside a controller...
raise ActiveRecord::RecordNotFound # -> 404 Not Found
raise MyCustomNotFound  # -> 500 Server Error 🤔
```

Since subclasses have an "is-a" relationship with their parents, it seems sensible that `MyCustomNotFound` should have the same default status code as `ActiveRecord::RecordNotFound`. Currently, to get the same behavior, each custom subclass has to be explicitly configured.

```ruby
# config/application.rb
config.action_dispatch.rescue_responses["MyCustomNotFound"] = :not_found
```

### Detail

This change results in custom subclasses returning their nearest registered parent class's status code. If no parent is registered, the existing default behavior still applies (map to a 500 Server Error).

The implementation is a simple recursive algorithm that walks up the class hierarchy to see if any parent classes are already registered. The results of each walk are cached so future lookups occur in O(1).

```ruby
class MyCustomNotFound < ActiveRecord::RecordNotFound
end

# inside a controller...
raise ActiveRecord::RecordNotFound # -> 404 Not Found
raise MyCustomNotFound  # -> 404 Not Found 😀
```

### Additional information

There's space to debate whether this is a bugfix or a breaking change.

With this change, existing subclasses of [predefined keys](https://guides.rubyonrails.org/configuring.html#config-action-dispatch-rescue-responses) in `ActionDispatch::ExceptionWrapper.rescue_responses` will start returning their parents' HTTP status codes instead of a 500.

### Checklist

### DIFF
--- a/actionpack/test/dispatch/exception_wrapper_test.rb
+++ b/actionpack/test/dispatch/exception_wrapper_test.rb
@@ -50,6 +50,11 @@ module ActionDispatch
       @cleaner = ActiveSupport::BacktraceCleaner.new
       @cleaner.remove_filters!
       @cleaner.add_silencer { |line| !line.start_with?("lib") }
+      @old_rescue_responses = ExceptionWrapper.rescue_responses.dup
+    end
+
+    teardown do
+      ExceptionWrapper.rescue_responses.replace(@old_rescue_responses)
     end
 
     class_eval "def index; raise TestError; end", "lib/file.rb", 42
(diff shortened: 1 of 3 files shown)


actionpack/test/dispatch/exception_wrapper_test.rb
```diff
@@ -208,6 +208,14 @@ def translate_location(backtrace_location, spot)
       assert_equal 400, wrapper.status_code
     end
 
+    test "#status_code walks ancestors to find registered parent exception" do
```
flavorjones, line 216: I'd like to see a few more tests to completely cover the desired behavior:

- when superclass and subclass are both registered with different status codes, the subclass should be handled correctly
- and exercise the same inheritance scenarios for `rescue_response?`

thehale: Added! Specifically, a registered subclass takes precedence over any registered parent classes.

PR #56942, opened 2026-03-06
Avoid AC::Live deadlocks while awaiting commit

masterful:
### Motivation / Background
This Pull Request has been created because `ActionController::Live` requests can deadlock with the development reloader while the parent request thread is blocked in `await_commit`; Fixes #56888.

### Detail
This Pull Request changes `ActionController::Live` to yield its running share while the parent thread is only waiting for the live child to commit, and adds an unload-specific interlock helper for that handoff.

It also adds a regression test that pauses the live child before it enters the interlock and verifies that unloading can proceed and the request completes afterward. (Not sure if this is super helpful, but my agent gave it to me for free, so ... let me know if you want it tossed instead 😅)

### Additional information
GPT-5.4 was used to validate this approach -- but I would not consider myself to be personally clever enough to actually decide whether this is a reasonable approach or not.

<details><summary>It had a reasonable answer when I asked it if we can unload during the `await_commit`</summary>

> Why I think it is acceptable:
> - While the parent live request thread is in await_commit, it is not executing app code. It is only waiting for the child thread to either commit headers or fail.
> - The thread that actually does the controller work is the child thread inside interlock.running at live.rb.
> - Historically, Rails already relied on this handoff. Commit [3e4a69e52d8](https://github.com/rails/rails/commit/3e4a69e52d8) added exactly this pattern, and because unload used to be compatible with :load, the old permit_concurrent_loads path effectively allowed unloads through too. So this is closer to restoring prior behavior than inventing a new one.
> 
> The real concern is the gap before the child acquires running:
> - If the parent yields and unload wins immediately, the child may start after constants were unloaded.
> - That sounds scary, but it is also the case that the child already has references to the controller instance/class objects it will run on. Ruby objects do not disappear just because their constants were removed.
> - So the main risk is stale-code semantics, not memory corruption. In development, stale-code semantics are usually acceptable if the alternative is deadlock.
</details>

I also updated the [example repo](https://github.com/GetJobber/reproduce-rails-deadlock-behaviour/pull/1) to validate that we no longer have locking with the new approach.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/live.rb
+++ b/actionpack/lib/action_controller/metal/live.rb
@@ -347,7 +347,9 @@ module ActionController
         end
       end
 
-      @_response.await_commit
+      ActiveSupport::Dependencies.interlock.permit_concurrent_unloads do
+        @_response.await_commit
+      end
 
       raise error if error
     end
--- a/actionpack/test/controller/live_stream_test.rb
+++ b/actionpack/test/controller/live_stream_test.rb
@@ -121,6 +121,10 @@ module ActionController
     class TestController < ActionController::Base
       include ActionController::Live
 
+      class << self
+        attr_accessor :before_live_thread_yield_latch, :release_live_thread_latch
+      end
+
       attr_accessor :latch, :tc, :error_latch
 
       def self.controller_path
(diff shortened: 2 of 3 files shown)


masterful:
@matthewd sorry to ping you directly about a full decade after you were in here 🙈, but I'm curious if you have opinions on this 😅 ❤️

crpahl:
cc @rmosolgo as well, but just for context (no rush on our end).

We were rolling out [@defer support](https://graphql-ruby.org/defer/overview), but we had to remove it because it was causing this locking issue in local development for us.

masterful:
I think I have a smidge more confidence in [the new PR](https://github.com/rails/rails/pull/58486) vs this one, though this one does have a slightly smaller footprint - I defer to the maintainers for judgement on which is preferable ❤️ 

(if we want this one, I can try and port over the test, or a version of it, but I'll close this one for now under the assumption that we probably want the new approach)

PR #56945, opened 2026-03-07
ActiveRecord: Support Postgres `RESET` on readonly queries

frodsan:
### Motivation / Background

This Pull Request has been created because the PostgreSQL adapter currently raises an ActiveRecord::ReadOnlyError when attempting to execute a [RESET](https://www.postgresql.org/docs/current/sql-reset.html) command within a read-only context (e.g., prevent_writes: true).

`RESET` acts as a syntactic shortcut for `SET configuration_parameter TO DEFAULT`.

### Checklist

### DIFF
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_prevent_writes_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_prevent_writes_test.rb
@@ -68,6 +68,16 @@ module ActiveRecord
         ActiveRecord::Base.while_preventing_writes do
           assert_equal [], @connection.execute("SET standard_conforming_strings = on").entries
         end
+      ensure
+        @connection.close
+      end
+
+      def test_doesnt_error_when_a_reset_query_is_called_while_preventing_writes
+        ActiveRecord::Base.while_preventing_writes do
+          assert_equal [], @connection.execute("RESET standard_conforming_strings").entries
+        end
+      ensure
+        @connection.close
       end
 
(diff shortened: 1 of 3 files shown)


activerecord/test/cases/adapters/postgresql/postgresql_adapter_prevent_writes_test.rb
```diff
@@ -70,6 +70,12 @@ def test_doesnt_error_when_a_set_query_is_called_while_preventing_writes
         end
       end
 
+      def test_doesnt_error_when_a_reset_query_is_called_while_preventing_writes
+        ActiveRecord::Base.while_preventing_writes do
+          assert_equal [], @connection.execute("RESET standard_conforming_strings").entries
```
byroot, line 77: This test is leaking state, as it mutate the connection config (may currently be harmless but might not in the future and be a source of CI flakiness).

I think we need to close that connection in an `ensure` block.

frodsan: Thanks for having a look! Any reason why don't we close the connection for [the other tests](https://github.com/frodsan/rails/blob/59d2628643672e488f01020004885c911d357389/activerecord/test/cases/adapters/postgresql/postgresql_adapter_prevent_writes_test.rb#L17-L65) as well 😄

byroot: We did, you just reproduced the mistake. It's fine.

PR #56955, opened 2026-03-10
Fix `has_many` replace race when stale loaded associations miss concurrent inserts

faraquet:
This PR introduces an opt-in fix for a race in `ActiveRecord` `has_many` collection replacement.

When `user.posts = [Post.new]` or `user.posts.replace([Post.new])` runs on an already loaded association, `CollectionAssociation#replace_records` currently diffs against the in-memory `target`.

That is unsafe if another session adds a new associated row after the association was loaded but before the replacement finishes. In that case, `replace` deletes only the rows it already knows about and leaves the later insert behind.

Because this changes existing behavior, the new logic is opt-in and remains disabled by default.

It can be enabled either:

```ruby
config.active_record.has_many_strict_replace = true
```

or per association:

```ruby
has_many :posts, strict_replace: true
```

## Example

```ruby
user.posts = [Post.new]
```

At a high level this means:

1. Remove existing associated posts.
2. Insert the new post.
3. End with exactly one associated post.

Today that can fail if the loaded association is stale.

## Current SQL shape

If the loaded target contains only post `101`, Rails may issue SQL like:

```sql
BEGIN;

DELETE FROM "posts"
WHERE "posts"."id" = 101;

INSERT INTO "posts" ("user_id", ...)
VALUES (1, ...);

COMMIT;
```

That works only if the database still matches the loaded target.

## Race Example

Session A loads the association:

```sql
SELECT "posts".*
FROM "posts"
WHERE "posts"."user_id" = 1;
-- => [id = 101]
```

Before Session A replaces the collection, Session B inserts another row:

```sql
INSERT INTO "posts" ("user_id", ...)
VALUES (1, ...);
-- => id = 102
```

Then Session A runs:

```ruby
user.posts = [Post.new]
```

and issues:

```sql
BEGIN;

DELETE FROM "posts"
WHERE "posts"."id" = 101;

INSERT INTO "posts" ("user_id", ...)
VALUES (1, ...);
-- => id = 103

COMMIT;
```

Final database state:

```sql
SELECT "posts"."id"
FROM "posts"
WHERE "posts"."user_id" = 1
ORDER BY "posts"."id";
-- => [102, 103]
```

That is the existing behavior when the association target is stale.

## Why This Happens

`CollectionAssociation#replace_records` currently computes removals from the in-memory `target`, not from the current persisted association rows inside the transaction.

So any rows inserted after the association was loaded are invisible to the delete phase.

## Opt-in behavior

When `config.active_record.has_many_strict_replace = true` or a plain `has_many` association is declared with `strict_replace: true`, replacement computes the current target from fresh persisted association rows instead of relying only on the loaded target.

Conceptually, for a replacement like `user.posts = [Post.new]`, the delete step then behaves like it is based on the current association scope rather than only on previously loaded record ids.

This opt-in behavior applies only to plain `has_many` associations.

It does not apply to `has_many :through` associations, and `strict_replace:` is not supported together with `through:`. Through associations compute their target via the join model chain — refreshing would require reloading the join and source associations, not just the target scope, which can introduce less predictable query and callback behavior.

## Change

This PR adds an opt-in `HasManyAssociation` path that reloads the current persisted association rows from the database and diffs against that state for plain `has_many` associations.

The new path:

- is enabled when `config.active_record.has_many_strict_replace` is `true`
- can also be enabled per association with `strict_replace: true`
- remains disabled by default to preserve existing behavior
- applies only to plain `has_many` associations
- does not apply to `has_many :through` associations — through associations compute their target via the join model chain, so refreshing would require reloading the join and source associations, not just the target scope
- adds an extra SQL query for the internal refresh
- skips strict loading validation for that internal refresh
- uses `uncached` to bypass the query cache for the fresh read, so it does not return the stale result from the initial `load`
- merges fresh persisted rows with in-memory target records via `merge_target_lists` so unsaved built records still participate correctly in `replace`

## Isolation level note

The fresh read runs `scope.to_a` inside whatever transaction context is active. Under PostgreSQL's default READ COMMITTED isolation this correctly sees concurrent inserts — which is the whole point. Under REPEATABLE READ (less common but used by some apps), the snapshot is taken at transaction start, so concurrently inserted rows may still be invisible and the race can persist even with `strict_replace`.

## Tests

Adds coverage for all relevant behaviors:

1. Default behavior remains unchanged when the project-level flag is disabled.
2. With `has_many_strict_replace` enabled:
   1. Load a plain `has_many` association.
   2. Insert another associated row afterward.
   3. Replace the collection with a new record.
   4. Assert that only the replacement record remains associated.
3. With `strict_replace: true` on a single plain `has_many` association:
   1. Load the association.
   2. Insert another associated row afterward.
   3. Replace the collection.
   4. Assert that only the replacement record remains associated.
4. Assert that `strict_replace:` is rejected on `has_many :through`.


tylerwood-io:
A few observations after reading through the diff:

1. **Isolation level interaction.** The fresh read in \`replace_records_current_target\` runs \`scope.to_a\` inside whatever transaction context is active. With PostgreSQL's default READ COMMITTED, this correctly sees concurrent inserts — which is the whole point. Worth noting that under REPEATABLE READ (less common but used by some apps), the fresh read would still return the snapshot from transaction start, so the race would persist even with \`strict_replace\`. Not a blocker, but might be worth a note in the docs.

2. **\`uncached\` wrapper** — good detail. Without it, query cache could return the stale result from the initial \`load\`.

3. **Through exclusion reasoning** — the comment says "different semantics and can introduce less predictable query and callback behavior." The concrete reason is that through associations compute their target via the join model chain, so refreshing would need to reload the join + source associations, not just the target scope. Might be worth making that explicit.

4. **\`merge_target_lists\`** — using this to combine fresh DB rows with unsaved in-memory records is the right call. Without it, \`user.posts.build; user.posts = [Post.new]\` would lose the built record from the diff comparison.

5. **Framework defaults** — since this defaults to \`false\`, it probably doesn't need a \`new_framework_defaults\` template entry yet. But if the intent is to eventually make it the default, that migration path should be planned.

Test coverage looks solid — both global config and per-association option, plus the through rejection.

faraquet:
Thanks for the feedback @tylerwood-io 👍 

All great observations!

I've updated the code comments and CHANGELOG based on your feedback:

1. **Isolation level interaction** - Added a note both in the code comment on `replace_records_current_target` and in the CHANGELOG explaining that the fresh read works as intended under READ COMMITTED but under REPEATABLE READ the snapshot is from transaction start, so the race can persist.

2. **Through exclusion reasoning** - Made the comment explicit: through associations compute their target via the join model chain, so refreshing would require reloading the join and source associations, not just the target scope.

3. **`uncached` wrapper** - Already in place, added a note in the PR description calling out that without it the query cache could return the stale result from the initial `load`.

4. **`merge_target_lists`** - No changes needed, already working as you described.

5. **Framework defaults** - Agreed, no `new_framework_defaults` entry needed for now since it defaults to `false`. When/if Rails team decide to make it the default, I'll plan the migration path with a `new_framework_defaults` template entry at that point.

I also updated the PR description to reflect all of the above.

PR #54126, opened 2025-01-06
Update `NotificationAssertions`'s  `assert_notifcation` to match against payload subsets and return matched notification

larouxn:
### Motivation / Background

This Pull Request has been created because:
1. One currently needs to exhaustively specify entire notification payloads (all key vals) in order to use `NotificationAssertions`'s `assert_notification`, which is not always practical or even possible. Thus, let's match against subsets by default. If one needs exhaustive matching they can do so manually via the following (point 2) or manual notification capture.
2. One currently needs to go lower level and `capture_notifications` even in cases where `assert_notification` would cover most of the desired assertions in the situation that one wants to do further, custom assertions against a matched notification. Thus, let's return the matched notification! Such is similar to how [Minitest's assert_raises](https://docs.seattlerb.org/minitest/Minitest/Assertions.html#method-i-assert_raises).

Regarding point 1. one example is we cannot currently use `assert_notification` in the following test because jobs' notifications' payloads contain various key vals, while we only want to assert against one.

https://github.com/rails/rails/blob/1b327ad4ed0fe5e8c2f1f795e2cf63f90a9833b5/activerecord/test/activejob/job_runtime_test.rb#L15-L21

Here's said assertion failing, exposing the list of payload key vals, only one of which we care about.

```ruby
vscode ➜ /workspaces/rails/activerecord (main) $ bin/test test/activejob/job_runtime_test.rb 
Using sqlite3_mem
Run options: --seed 23503

# Running:

.{:job=>#<JobRuntimeTest::TestJob:0x00007f80d3efa000 @arguments=[], @job_id="59c22946-0730-4db4-a63f-b71b72804a51", @queue_name=#<Proc:0x00007f80d40a4978 /workspaces/rails/activejob/lib/active_job/queue_name.rb:55 (lambda)>, @scheduled_at=nil, @priority=nil, @executions=1, @exception_executions={}, @timezone=nil, @_halted_callback_hook_called=nil>, :adapter=>#<ActiveJob::QueueAdapters::TestAdapter:0x00007f80d3e42388>, :db_runtime=>42.0, :aborted=>nil}
.

Finished in 0.002990s, 668.9259 runs/s, 1003.3888 assertions/s.
2 runs, 3 assertions, 0 failures, 0 errors, 0 skips
```

If we could assert against a subset of the said payload this test would be cleaner and simpler.

Follow up to
- https://github.com/rails/rails/pull/53065
- https://github.com/rails/rails/pull/53700 (merged as per gem PRs)
[...]

PR #53065, opened 2024-09-27
Add `ActiveSupport::Testing::NotificationAssertions` test helper module

larouxn:
### Motivation / Background

This Pull Request has been created because it's currently cumbersome to test that a certain code block/action triggers an `ActiveSupport::Notifications::Event` to be emitted. It would be ideal to have some helpers to `assert` against such event emission. Have heard such helpers could be helpful internal to Rails as well.

Notes
- Planning to clean up various parts of `rails/rails` with these helpers as a follow-up. ([WIP branch](https://github.com/larouxn/rails/tree/migrate_everywhere_to_event_assertions_helpers))
- Already have ideas for ways we can make this module even more useful: [comment from below](https://github.com/rails/rails/pull/53065#issuecomment-2452017030).

Co-authored by @yishus and @seanpdoyle :pray: 

### Detail

This Pull Request adds a new test helpers module called `ActiveSupport::Testing::NotificationAssertions`.

### Additional information

- An earlier form of these test helpers has been in use in production CI for our Rails application for over a month.
- Attempted to include method level documentation with a similar style to nearby code.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #56956, opened 2026-03-10
Simplify applicable `actionpack` tests with `NotificationAssertions` helpers

larouxn:
### Motivation / Background

This Pull Request has been created because I found some more `ActiveSupport::Notifications` centric tests in `actionpack` that can be simplified using the `ActiveSupport::Testing::NotificationAssertions` test helper module which was added in https://github.com/rails/rails/pull/53065 and https://github.com/rails/rails/pull/54126.

Follow up to
- https://github.com/rails/rails/pull/53819

Related recent PRs
- https://github.com/rails/rails/pull/56974
- https://github.com/rails/rails/pull/56975

### Detail

This Pull Request changes five tests in `actionpack/test/controller/redirect_test.rb` to use `NotificationAssertions` test helpers. For example, when using the provided helpers we no longer need to manually clean up subscribers via `ensure` blocks as the helpers do that for us automatically. Additionally, in cases where it makes sense we can directly make assertions using the helpers and in cases where it doesn't we can fall back to simply capturing notifications and manually asserting.

### Additional information

https://api.rubyonrails.org/classes/ActiveSupport/Testing/NotificationAssertions.html

### Checklist

* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. - not necessary

### DIFF
--- a/actionpack/test/controller/redirect_test.rb
+++ b/actionpack/test/controller/redirect_test.rb
@@ -675,12 +675,9 @@ class RedirectTest < ActionController::TestCase
 
   def test_redirect_to_path_relative_url_starting_with_an_at_with_notify
     with_path_relative_redirect(:notify) do
-      events = []
-      subscriber = ActiveSupport::Notifications.subscribe("unsafe_redirect.action_controller") do |*args|
-        events << ActiveSupport::Notifications::Event.new(*args)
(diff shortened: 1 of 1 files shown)


skipkayhil (approved):
Awesome simplification, thanks!

PR #56963, opened 2026-03-11
Restore previous instrumenter after `execute_or_skip`

rosa:
### Motivation / Background

We started using async queries (via `async_load`  and `async_pick`) in one of our apps, with [`async_query_executor`](https://guides.rubyonrails.org/configuring.html#config-active-record-async-query-executor) set to `:global_thread_pool`. We noticed that some requests wouldn't get most of their synchronous queries instrumented and logged. The number of requests with this problem went to zero after a deploy, then increased steadily until we deployed again, when it went back to zero before starting to grow again. 

After some troubleshooting, we realised this was due to the `global_thread_pool` saturating and running some of the async queries in the main thread. This happens because the `fallback_policy` for the global thread pool is `:caller_runs`:
https://github.com/rails/rails/blob/0209a70c57ebef4fb1e271444ec7e62794932db3/activerecord/lib/active_record.rb#L304-L313

The number of tasks that can be running or waiting at most in the global thread pool is 20 (4 threads with the default value for `concurrency` + 16 for the max queue size). We didn't run as many async queries per request with Puma set to a single thread, but Active Record's connection pool reaper also runs its asynchronous maintenance tasks in the global thread pool, so we were seeing the saturation in the end, increasing as the time since the last deploy passed. 

Async queries are executed via  [`QueryIntent#execute_or_skip`](https://github.com/rails/rails/blob/0209a70c57ebef4fb1e271444ec7e62794932db3/activerecord/lib/active_record/connection_adapters/query_intent.rb#L95-L119). This method sets `ActiveSupport::IsolatedExecutionState[:active_record_instrumenter]` to an `EventBuffer` but never restores the previous value. This is safe when the method runs on a background thread (as designed), but when the task executes on the calling thread, because the global thread pool is saturated and because of the `caller_runs` fallback policy, the stale `EventBuffer` permanently replaces the real instrumenter in the thread's `IsolatedExecutionState`. All subsequent `sql.active_record` notifications on that thread are captured by the `EventBuffer` but never published (`flush` is only called once by `ensure_result`, for the async query), silently breaking notification subscribers for `sql.active_record` events. Moreover, I don't think this is ever restored or cleared between requests, so the thread stays contaminated forever, which would also explain the increase in requests with this issue as time passed after the deploy, as more threads would get a broken instrumenter. 

### Detail

This PR implements a fix by saving and restoring the previous instrumenter value around the `EventBuffer'`s lifetime. On background threads, this is a no-op (saves `nil`, restores `nil`). On the caller thread via `caller_runs`, it restores the real instrumenter and prevents contamination.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/query_intent.rb
+++ b/activerecord/lib/active_record/connection_adapters/query_intent.rb
@@ -100,6 +100,7 @@ module ActiveRecord
 
           @pool.with_connection do |connection|
             return unless @mutex.try_lock
+            previous_instrumenter = ActiveSupport::IsolatedExecutionState[:active_record_instrumenter]
             begin
               if pending?
                 @event_buffer = EventBuffer.new(self, ActiveSupport::Notifications.instrumenter)
@@ -112,6 +113,7 @@ module ActiveRecord
             rescue => error
               @error = error
             ensure
+              ActiveSupport::IsolatedExecutionState[:active_record_instrumenter] = previous_instrumenter
               @mutex.unlock
             end
           end
--- a/activerecord/test/cases/relation/load_async_test.rb
+++ b/activerecord/test/cases/relation/load_async_test.rb
@@ -194,6 +194,39 @@ module ActiveRecord
       end
     end
 
+    def test_execute_or_skip_does_not_contaminate_caller_thread_instrumenter
+      skip unless ActiveRecord::Base.connection.async_enabled?
+
+      begin
+        # Intercept schedule_query to run execute_or_skip on the current thread,
+        # simulating what happens when the async executor's caller_runs fallback
+        # policy triggers due to a saturated thread pool.
+        pool = Post.connection_pool
+        old_schedule_query = pool.method(:schedule_query)
+        pool.singleton_class.undef_method(:schedule_query)
+        pool.singleton_class.define_method(:schedule_query) do |future_result|
+          future_result.execute_or_skip
+        end
+
+        # This async query runs execute_or_skip on the current thread
+        Post.async_count
+
+        # After the async query completes, synchronous queries must still
+        # publish sql.active_record notifications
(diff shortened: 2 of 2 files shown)


jeremy (approved):
Great catch. Changelog, please! Let's backport as appropriate, too (not sure how far).

rosa:
Thanks @jeremy! I've added the `CHANGELOG` entry, I'll look into backporting this.

rosa:
> Let's backport as appropriate, too (not sure how far)

Looks like we need to backport to 8.0 and 8.1, 7.2 doesn't have this bug because it uses `connection.with_instrumenter`, that takes care of swapping and then restoring the previous instrumenter: 
https://github.com/rails/rails/blob/87155d23013276a9f3289d648a9a744204d5d212/activerecord/lib/active_record/future_result.rb#L108-L110

I'll prepare the fixes for 8.0 and 8.1.

rosa:
Backports (although TBH I'm not sure if this is how you do it 😅 I've never done it before! Let me know if this should be different): 
- 8.0: https://github.com/rails/rails/pull/56964
- 8.1: https://github.com/rails/rails/pull/56965

byroot:
> This happens because the `fallback_policy` for the global thread pool is `:caller_runs`:

Hum, I can't remember why I chose this, but it doesn't make a whole lot of sense. We kinda already have this behavior provided by the future when you try iterating on it, so It's OK to queue more jobs.

Either way, this patch looks good, I'll see about the executor config later.

byroot:
> Hum, I can't remember why I chose this,

Ah, it's because the only other options are `:abort, or `:discard`. That's said, the queue size default to `max_threads * 4` so that should only happen with very heavy usage, in which case it's not that bad of an idea to get the producer thread to slow down (backpressure).

So perhaps it's best to keep it this way.

PR #56967, opened 2026-03-12
Optimize generated Dockerfile build performance

damln:
### Summary

Two Docker build optimizations for the generated Dockerfile template and the test fixture:

**1. Merge `rm -rf node_modules` into asset precompile layer** (`Dockerfile.tt`)

The separate `RUN rm -rf node_modules` creates an extra layer that forces BuildKit to compute a full filesystem diff for the deleted files. Merging it into the `assets:precompile` RUN step eliminates this overhead (~13s saved in practice).

**2. Replace `chown -R` with `COPY --chown`** (`Dockerfile.test`)

The test fixture Dockerfile uses the slower `chown -R rails:rails db log storage tmp` pattern. The main template already uses `COPY --chown=rails:rails --from=build`, but the test fixture was not updated to match. Using `--chown` on COPY
sets ownership at copy time without creating an extra diff layer (~50s saved in practice).

### Motivation

Profiling a production build showed these two patterns account for ~60-70s of unnecessary build time. The main Dockerfile template already got the `COPY --chown` optimization right — this PR brings the test fixture in line and merges the
node_modules cleanup into a single layer.

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt
@@ -97,13 +97,14 @@ RUN bundle exec bootsnap precompile -j 1 app/ lib/
 <% end -%>
 <% unless options.api? || skip_asset_pipeline? -%>
 # Precompiling assets for production without requiring secret RAILS_MASTER_KEY
+<% if using_node? || using_bun? -%>
+RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile && \
+    rm -rf node_modules
+<% else -%>
 RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile
-
 <% end -%>
 
-<% if using_node? || using_bun? -%>
-RUN rm -rf node_modules
-<% end %>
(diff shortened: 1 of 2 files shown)


damln:
thx @Intrepidd  for the optim in:
```
rm -rf node_modules
```

samuelabadie:
👍

leoamigood:
@damln with the latest changes in Dockerfile template I noticed that files ownership in produced docker image under `/rails` changed from `0:0`  (`root`) to `1000:1000` (`rails`) Is that intentional?

tcwalther:
I saw this change in the Rails weekly digest, but I'm a bit confused - my Dockerfile already said:

```Dockerfile
...
# Run and own only the runtime files as a non-root user for security
RUN groupadd --system --gid 1000 rails && \
    useradd rails --uid 1000 --gid 1000 --create-home --shell /bin/bash
USER 1000:1000

# Copy built artifacts: gems, application
COPY --chown=rails:rails --from=build "${BUNDLE_PATH}" "${BUNDLE_PATH}"
COPY --chown=rails:rails --from=build /rails /rails
...
```

Does this PR address a regression then?

Issue #51630, opened 2024-04-22
`ActiveSupport::Inflector` isolation/conflicts: recommended approach


krystof-k:
I'm trying to utilize `ActiveSupport::Inflector`:

```ruby
# config/initializers/inflections.rb

ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym "API"
  inflect.acronym "ID"
  inflect.acronym "JSON"
  inflect.acronym "PDF"
  inflect.acronym "URL"
end
```

and it breaks all the time.

One thing is autoloading, which can be sort of handled via custom `Zeitwerk::Inflector`. I created one:

<details>
<summary><em>Custom Zeitwerk inflector to override conflicts</em></summary>

```ruby
# config/initializers/zeitwerk.rb

# List of file glob patterns and their corresponding camelized class names
OVERRIDES = {
  "**/gems/actionmailbox-*/app/models/action_mailbox/inbound_email/message_id.rb" => "MessageId",
  "**/gems/avo-*/app/components/avo/fields/id_field" => "IdField",
  "**/gems/avo-*/app/helpers/avo/url_helpers.rb" => "UrlHelpers"
}.freeze

# Override the default Zeitwerk inflector to use the overrides
Rails.autoloaders.each do |autoloader|
  autoloader.inflector = Class.new(Zeitwerk::Inflector) do
    define_method(:camelize) do |basename, abspath|
      # Check if the file path matches any of the overrides
      OVERRIDES.each do |glob_pattern, camelized|
        if File.fnmatch?(glob_pattern, abspath)
          # Return the hard-coded camelized class name
          return camelized
        end
      end
      # For all other files, use the default ActiveSupport::Inflector.camelize
      basename.camelize
    end
  end.new
end
```

</details>

 and while it still breaks all the time, it can be fixed.

But then there are cases, when the conflict is within the code of a 3rd party gem or engine, using the inflector methods like `.classify`, `.underscore`, `.camelize` etc.

For example, I recently bumped into the `JSON` acronym issue in the [active_model_serializers](https://github.com/rails-api/active_model_serializers) gem: 

> It converts `json_api` using `"json_api".classify` to `JsonApi`.
>
> Some controller:
> ```ruby
> render json: data, adapter: :json_api, status: 200
> ```
>
[...]

PR #56968, opened 2026-03-12
Add inflector global side effects warnings throughout the docs [ci skip]

krystof-k:
### Motivation / Background

Warn about custom inflection rules global side effects, which cause confusing and hard-to-trace issues.

Based on the 2-year-old issue #51630, where @rafaelfranca noted "we don't recommend changing the inflector because it has global effect on the Ruby process." Nothing changed since then (and won't in the future I assume).

### Detail

Add warnings to the API docs, guides and `inflections.rb` initializer template.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/inflector/inflections.rb
+++ b/activesupport/lib/active_support/inflector/inflections.rb
@@ -28,6 +28,19 @@ module ActiveSupport
     # rule for cactus will now be the first of the pluralization and
     # singularization rules that is runs. This guarantees that your rules run
     # before any of the rules that may already have been loaded.
+    #
+    # <b>WARNING:</b> Custom inflection rules modify a process-wide singleton
+    # and affect every inflection method call (+camelize+, +classify+,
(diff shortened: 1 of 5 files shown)


krystof-k:
While working on this, I discovered two more issues with Zeitwerk's inflector (now documented in the last two commits):

1. **Zeitwerk's inflector doesn't affect Rails constant resolution.** Routing, STI, parameter wrapping, and many other code paths resolve constants via `camelize` → `constantize`, independently of Zeitwerk. For example, `namespace :api` always looks for `Api::` regardless of Zeitwerk config.

2. **Zeitwerk's overrides affect engine code**, since engines share the application's autoloader.

---

Since Rails 7 is fully on Zeitwerk, these code paths could use Zeitwerk's inflector instead of guessing via `camelize`. This would make scoped inflection a true alternative to global rules and fix the long-standing `Api` vs `API` problem without the need for any workarounds. There is a related issue (zeitwerk/issues/106) in the Zeitwerk repository.

Is this something that could be refactored? I'm happy to open another PR.

PR #56971, opened 2026-03-12
Fix crashing schema dump for virtual tables with SQLite and Spatialite

hschne:
### Motivation / Background

Fixes #56969

All details (reproduction, etc) are found in the issue. 

### Detail

 Changes `(.+)` to `(.*)` in `VIRTUAL_TABLE_REGEX`. With zero-or-more
 matching, empty parentheses produce empty string instead of nil, and
 `"".split(", ")` returns [] without error. No changes to the schema
 dumper are needed.

### Additional information

Added a unit test that asserts VIRTUAL_TABLE_REGEX correctly parses
 a USING SomeModule() SQL string. This doesn't test the issue *per se*, but a full integration test isn't feasible without installing a third-party SQLite extension.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Fix schema dump crash when SQLite3 virtual tables use empty parentheses.
+
+    Virtual table modules such as SpatiaLite's `VirtualSpatialIndex()` declare no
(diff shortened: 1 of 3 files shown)


fatkodima:
Merged #56970, since it was created first.

PR #56983, opened 2026-03-15
[ci skip] Update CHANGELOG with transform_keys! overwrite precedence note

diaphragm:
### Motivation / Background

This Pull Request has been created because the previous bug fix for `ActiveSupport::HashWithIndifferentAccess#transform_keys!` in Rails 8.0.3 introduced a change in overwrite precedence that was not explicitly documented. 

When distinct keys are transformed into the exact same key, the later key now correctly overwrites the earlier one (matching standard Ruby `Hash` behavior). Previously, the earlier key's value was preserved. Documenting this is important to prevent unexpected data loss or silent bugs for users upgrading existing applications.

Fixes #56960

### Detail

This Pull Request changes the `activesupport/CHANGELOG.md` entry for Rails 8.0.3 to include a brief note explaining this overwrite precedence change. It adds a clear `Before` and `After` code example showing the behavior when a camelCase key and a snake_case key collide.

### Additional information

Opening this PR targeting the `8-0-stable` branch as requested in the issue.
cc @byroot

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -95,6 +95,26 @@
 
     *Jason T Johnson*, *Jean Boussier*
 
+    **Note on overwrite precedence:**
+
+    This fix also changed the overwrite precedence when distinct keys are transformed into the same key.
+    The later key now overwrites the earlier one (matching standard Ruby `Hash` behavior).
+    Previously, the earlier key's value was preserved.
+
+    Before:
+
+    ```ruby
+    >> {XY: 1, xy: 2}.with_indifferent_access.transform_keys!(&:downcase)
+    => {"xy"=>1}
+    ```
+
+    After:
+
+    ```ruby
+    >> {XY: 1, xy: 2}.with_indifferent_access.transform_keys!(&:downcase)
+    => {"xy"=>2}
+    ```
+
 *   Fix `ActiveSupport::Cache::MemCacheStore#read_multi` to handle network errors.
 
     This method specifically wasn't handling network errors like other codepaths.


byroot:
This fails the changelog linter.

activesupport/CHANGELOG.md
```diff
@@ -95,6 +95,26 @@
...
+  
+    This fix also changed the overwrite precedence when distinct keys are transformed into the same key.
+    The later key now overwrites the earlier one (matching standard Ruby `Hash` behavior).
+    Previously, the earlier key's value was preserved.
+
+    Before:
+
+    ```ruby
+    >> {xxYy: { a: 1 }, xx_yy: {}}.with_indifferent_access.transform_keys!(&:underscore)
+    => {"xx_yy"=>{"a"=>1}}
```
byroot: You should use a more stripped down, and simple to grasp example:

```suggestion
    >> {xY: 1, Xy: 2}.with_indifferent_access.transform_keys!(&:downcase)
    => {"xy" => 1}
```

diaphragm:
Hi byroot,

Thank you so much for the review and the suggestion to simplify the example! It is definitely a much better approach.

While testing the snippet locally to verify the output, I realized a subtle nuance about the previous behavior: `{xY: 1, Xy: 2}` actually returns `{"xy" => 2}` in both Rails 7.2.0 and 8.0.3.

It seems the specific case where Rails 7.2 preserved the earlier key's value is when the earlier key is transformed and collides with a later key, such as `{ XY: 1, xy: 2 }`.

```ruby
# frozen_string_literal: true

# rails_version = "7.2.0"
rails_version = "8.0.3"

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", rails_version
end

require "minitest/autorun"
require "action_controller"

class ParametersTransformKeysTest < Minitest::Test
  # 7.2.0 => pass
  # 8.0.3 => fail
  def test_behavior_when_transforming_earlier_key
    params = {XY: 1, xy: 2}.with_indifferent_access.transform_keys!(&:downcase)

    assert_equal({"xy"=>1}, params.to_h)
  end

  # 7.2.0 => pass
  # 8.0.3 => pass
  def test_behavior_when_transforming_later_key
    params = {xy: 1, XY: 2}.with_indifferent_access.transform_keys!(&:downcase)

    assert_equal({"xy"=>2}, params.to_h)
  end

  # 7.2.0 => pass
  # 8.0.3 => pass
  def test_behavior_when_transforming_both_keys
    params = {xY: 1, Xy: 2}.with_indifferent_access.transform_keys!(&:downcase)

    assert_equal({"xy"=>2}, params.to_h)
  end
end
```

I have updated the PR to use `{ XY: 1, xy: 2 }` as the simplified example so it accurately demonstrates the difference between the versions. I also removed the trailing whitespace that was causing the linter to fail.

Could you please take another look when you have a moment? Thank you again for your guidance!

PR #56985, opened 2026-03-15
has_one :through ignores preloaded associations, fires N+1 query #56978

nicolasva:
issue #56978

Area | File | Change
-- | -- | --
Bug | — | has_one :through ignores preloaded associations, fires N+1 queries even when the entire object graph is already loaded via includes()
Root Cause | — | HasOneThroughAssociation inherits find_target from SingularAssociation, which always builds a SQL scope and queries the database without checking if the through chain is already in memory
Fix | has_one_through_association.rb:10-28 | Override find_target to traverse already-loaded associations before falling back to a SQL query
Test 1 | has_one_through_associations_test.rb:481 | Preloaded has_one :through triggers zero queries
Test 2 | has_one_through_associations_test.rb:488 | Inverse-direction preloaded chain marks associations as loaded?
Test 3 | has_one_through_associations_test.rb:498 | Returns nil without query when through target is nil
Test 4 | has_one_through_associations_test.rb:510 | Falls back to SQL query when nothing is preloaded
Test 5 | has_one_through_associations_test.rb:518 | Nested has_one :through also uses preloaded chain
Results | — | 57 tests, 187 assertions, 0 failures, 0 errors

### DIFF
--- a/activerecord/lib/active_record/associations/has_one_through_association.rb
+++ b/activerecord/lib/active_record/associations/has_one_through_association.rb
@@ -7,6 +7,26 @@ module ActiveRecord
       include ThroughAssociation
 
       private
+        def find_target(async: false)
+          # If the through association is already loaded and its target also has
+          # the source association loaded, traverse the loaded chain instead of
+          # firing a database query. This prevents N+1 queries when the entire
+          # object graph has been eager-loaded via includes().
+          if through_association.loaded?
+            through_target = through_association.target
+            if through_target
+              source_assoc = through_target.association(source_reflection.name)
+              if source_assoc.loaded?
+                return source_assoc.target
+              end
+            else
+              return nil
+            end
+          end
+
+          super
+        end
+
         def replace(record, save = true)
           create_through_record(record, save)
           self.target = record
--- a/activerecord/test/cases/associations/has_one_through_associations_test.rb
+++ b/activerecord/test/cases/associations/has_one_through_associations_test.rb
@@ -477,6 +477,56 @@ class HasOneThroughAssociationsTest < ActiveRecord::TestCase
 
     assert_predicate(book.association(:order_agreement), :stale_target?)
   end
+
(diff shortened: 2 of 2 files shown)


tylerwood-io:
One edge case worth considering with the `find_target` override: since `find_target` is called from within the `reload` path (`reader` → `reload` → `reset` → `load_target` → `find_target`), an explicit `reload_club` or `association(:club).reload` will also resolve from the in-memory chain instead of querying the database. `reset` clears `self` (the `has_one :through` association), but the through chain lives on `owner.association(through_reflection.name)` — so after `reset`, `through_association.loaded?` is still true and `find_target` returns the cached value.

This means:
```ruby
member = Member.includes(current_membership: :club).first

# Someone updates the club in another process...

member.reload_club  # Still returns the stale in-memory value
```

An alternative is overriding `reader` instead, which only intercepts the read accessor path. `force_reload_reader` and explicit `reload` calls bypass it entirely, preserving the expectation that explicit reloads always hit the database. I took that approach in #57079.

tylerwood-io:
Hey — I have an alternative PR for the same issue at #57079 that takes a different approach worth comparing.

This PR overrides `find_target`, which works for the basic preloaded case. However, `find_target` is called from within the `reload` path (`reader` → `reload` → `reset` → `load_target` → `find_target`). This means `reload_<association>` and explicit `reload` calls would also short-circuit from memory instead of hitting the database, because `reset` only clears `self` — not the through chain on the owner. That could mask stale data after mutations.

#57079 overrides `reader` instead, which avoids this — `force_reload_reader` and explicit reload calls bypass the override entirely and always query the database.

Additionally, #57079 now includes guards that bail out of the fast path for scoped associations (`-> { where(...) }`) and `source_type` polymorphic associations, where walking the preloaded chain could return a record the scoped/typed query would have excluded. These correctness gaps apply to any chain-walking approach, including this one.

PR #56988, opened 2026-03-15
Fix insert_all raising "No unique index found for id" without unique_by

hammadxcm:
### Motivation / Background

Fixes #56953

`insert_all` and `upsert_all` raise `ArgumentError: No unique index found for id` on PostgreSQL when `unique_by` is not provided, even on tables with a standard `id` primary key.

This affects libraries like SolidCable where `Model.insert(...)` (which uses `insert_all` with `on_duplicate: :skip`) fails on PostgreSQL.

**Root cause**: PostgreSQL's `indexes()` method explicitly excludes primary key indexes (`indisprimary = 'f'` in the SQL query). When `find_unique_index_for` can't find the PK in `unique_indexes` and `unique_by` is nil, it falls through to compare `model.primary_key` against `schema_cache.primary_keys(table_name)`. These two values are derived from different code paths and can diverge (composite PK where model overrides PK, multi-database setups with stale cache, etc.).

### Detail

When `unique_by` is nil and no matching unique index is found, `find_unique_index_for` now returns `nil` early. This is safe because:

- **`insert_all` (skip)**: generates `ON CONFLICT DO NOTHING` — valid SQL without a conflict target
- **`upsert_all` (update)**: `conflict_target` already falls back to `primary_keys` directly when `unique_by` is nil (line 275-276 of insert_all.rb)

The `match == primary_keys` branch is preserved for when `unique_by` IS explicitly provided and matches the PK columns (creates a synthetic IndexDefinition).

Tests updated:
- Added 7 new regression tests covering insert_all/upsert_all without unique_by on standard and composite PKs
- Updated existing composite PK test to run on all adapters (removed `!supports_insert_conflict_target?` skip)
- Updated Speedometer test to expect `StatementInvalid` (no real DB PK constraint) instead of `ArgumentError`

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -3,6 +3,20 @@
 
     *Joshua Young*
 
+*   Fix `insert_all` raising `ArgumentError: No unique index found for id`
+    when `unique_by` is not provided and the adapter's `indexes()` method
+    excludes primary key indexes (e.g. PostgreSQL).
+
+    When `unique_by` is nil, `find_unique_index_for` now returns `nil` early
+    instead of trying to match the primary key against `unique_indexes`.
+    This allows `insert_all` to generate `ON CONFLICT DO NOTHING` without a
+    conflict target, and `upsert_all` to fall back to `primary_keys` for
+    the conflict target.
+
+    Fixes #56953.
+
+    *Hammad Khan*
+
 *   Accept encryption credentials as ENV
 
     Taking advantage of Rails.apps.creds (#56455), the `primary_key`, `deterministic_key` and
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -4,14 +4,9 @@
     *Joshua Young*
 
 *   Fix `insert_all` raising `ArgumentError: No unique index found for id`
-    when `unique_by` is not provided and the adapter's `indexes()` method
-    excludes primary key indexes (e.g. PostgreSQL).
-
-    When `unique_by` is nil, `find_unique_index_for` now returns `nil` early
-    instead of trying to match the primary key against `unique_indexes`.
-    This allows `insert_all` to generate `ON CONFLICT DO NOTHING` without a
-    conflict target, and `upsert_all` to fall back to `primary_keys` for
-    the conflict target.
+    when `unique_by` is not provided. `insert_all` uses `ON CONFLICT DO NOTHING`
+    which does not require a conflict target, so the unique index lookup is
+    skipped when `unique_by` is not specified.
 
     Fixes #56953.
 
--- a/activerecord/test/cases/insert_all_test.rb
+++ b/activerecord/test/cases/insert_all_test.rb
@@ -317,21 +317,18 @@ class InsertAllTest < ActiveRecord::TestCase
     end
   end
 
-  def test_insert_all_and_upsert_all_works_with_composite_primary_keys_when_unique_by_is_not_provided
+  def test_insert_all_works_with_composite_primary_keys_when_unique_by_is_not_provided
     skip unless supports_insert_on_duplicate_skip?
 
-    assert_difference "Cart.count", 3 do
+    assert_difference "Cart.count", 2 do
       Cart.insert_all [{ id: 1, shop_id: 1, title: "My cart" }]
 
       Cart.insert_all! [{ id: 2, shop_id: 1, title: "My cart 2" }]
-
-      Cart.upsert_all [{ id: 3, shop_id: 2, title: "My other cart" }]
     end
   end
 
   # Regression tests for https://github.com/rails/rails/issues/56953
-  # insert_all without unique_by should not raise "No unique index found for id"
-  # even when the adapter's indexes() method excludes primary key indexes.
+  # insert_all without unique_by should not raise "No unique index found for id".
 
   def test_insert_all_without_unique_by_does_not_raise_on_standard_primary_key
     skip unless supports_insert_on_duplicate_skip?
@@ -566,7 +563,7 @@ class InsertAllTest < ActiveRecord::TestCase
   def test_upsert_all_fails_for_configured_primary_key_with_no_database_constraint
     skip unless supports_insert_on_duplicate_update? && supports_insert_conflict_target?
 
-    assert_raises ActiveRecord::StatementInvalid do
+    assert_raises ArgumentError do
       Speedometer.upsert_all [{ speedometer_id: "s1", name: "New Speedometer" }]
     end
   end
--- a/activerecord/test/cases/insert_all_test.rb
+++ b/activerecord/test/cases/insert_all_test.rb
@@ -317,22 +317,8 @@ class InsertAllTest < ActiveRecord::TestCase
     end
   end
 
-  def test_insert_all_and_upsert_all_raises_when_no_unique_index_found_for_composite_primary_key
-    skip unless supports_insert_conflict_target?
-
-    error = assert_raises ArgumentError do
-      Cart.insert_all [{ id: 2, shop_id: 1, title: "My cart" }]
-    end
-    assert_match "No unique index found for id", error.message
-
-    error = assert_raises ArgumentError do
-      Cart.upsert_all [{ id: 2, shop_id: 1, title: "My cart" }]
-    end
-    assert_match "No unique index found for id", error.message
-  end
-
   def test_insert_all_and_upsert_all_works_with_composite_primary_keys_when_unique_by_is_not_provided
-    skip unless supports_insert_on_duplicate_skip? && !supports_insert_conflict_target?
+    skip unless supports_insert_on_duplicate_skip?
 
     assert_difference "Cart.count", 3 do
       Cart.insert_all [{ id: 1, shop_id: 1, title: "My cart" }]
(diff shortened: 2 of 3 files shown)


skipkayhil:
Did you look into why the primary key index is excluded from indexes for PostgreSQL? That seems like the thing to fix since it'll make all of the adapters consistent

hammadxcm:
Great question — I did dig into this before settling on the current approach.

It turns out the PK exclusion from `indexes()` is intentional and consistent across all three adapters:

- **PostgreSQL**: filters with `d.indisprimary = 'f'` in SQL (dating back to [11411effde](https://github.com/rails/rails/commit/11411effde), 2005)
- **MySQL**: skips with `next if row["Key_name"] == "PRIMARY"` in Ruby
- **SQLite**: PK autoindexes are named `sqlite_autoindex_*` and get filtered by `next if row["name"].start_with?("sqlite_")`

The main reason is that `indexes()` was designed to serve the schema dumper — it generates `t.index` / `add_index` statements, and including PKs would produce redundant index definitions since the `PRIMARY KEY` constraint already creates them.

Changing `indexes()` to include PKs would require updating all three adapters, their tests, and ensuring the schema dumper filters them out to avoid generating duplicate index definitions. It felt like the wrong lever to pull for this issue.

The real problem was that `find_unique_index_for` was trying to locate an index even when the user never asked for one (`unique_by: nil`). In that case, we don't actually need a specific index:
- `insert_all` just needs `ON CONFLICT DO NOTHING` (no target required)
- `upsert_all` already falls back to `primary_keys` directly for the conflict target

So this fix just short-circuits that unnecessary lookup. Happy to discuss further if you think the adapter-level approach would be better though!

pisarz77:
Tested this patch against Rails 8.1.2 (the version in #56953 reproducer).

**Reproducer** (PostgreSQL 17, Rails 8.1.2, single DB):

```ruby
klass = Class.new(SolidCable::Record) do
  self.table_name = "solid_cable_messages"
  self.primary_key = ["channel", "channel_hash"]
end
klass.insert_all([{ channel: "x", payload: "y", channel_hash: 1, created_at: Time.current }])
```

**Before** (`gem "rails", "~> 8.1.2"`):
```
ArgumentError: No unique index found for ["channel", "channel_hash"]
  activerecord-8.1.2/lib/active_record/insert_all.rb:165:in 'find_unique_index_for'
```

**After** (`gem "rails", github: "hammadxcm/rails", branch: "fix-insert-all-no-unique-index-for-id"`):
```
OK — divergent PK insert_all succeeded
```

Also verified:
- Standard `SolidCable::Message.broadcast` still works
- End-to-end `Comment.create!` with `broadcasts_to :post` succeeds, `solid_cable_messages` row inserted
- Explicit invalid `unique_by:` still raises `ArgumentError: No unique index found for nonexistent_column` (regression behavior preserved)

@hammadxcm — heads up the branch has merge conflicts on `activerecord/CHANGELOG.md` (top-of-file entries diverged) and likely on `activerecord/test/cases/insert_all_test.rb` (commit ccfe1ae1fb added an anonymous-class test around line 352, overlapping your edits). buildkite/rails currently FAILURE — likely just stale base. A rebase on `main` should clear both.

@skipkayhil — your earlier question got a thorough answer (PK exclusion is intentional in all three adapters since 2005, fix at `find_unique_index_for` level avoids touching schema dumper invariants). Could you take another look?

hammadxcm:
Thanks for the heads up @pisarz77 — and for the thorough verification against 8.1.2!

Rebased onto current `main` (commit `26a3751`):
- `activerecord/CHANGELOG.md` conflict resolved (entry placed at top of unreleased section).
- `activerecord/lib/active_record/insert_all.rb` and `activerecord/test/cases/insert_all_test.rb` auto-merged cleanly. The upstream anonymous-class log-message test from `ccfe1ae1fb` sits alongside the regression tests added here (lines 336–399 vs. 411–424) — no semantic overlap.
- Local `ARCONN=sqlite3` run of `insert_all_test.rb`: 98 runs, 224 assertions, 0 failures, 0 errors. Buildkite should now have a clean base.

@skipkayhil — gentle ping in case my earlier reply got buried. Short version: PK exclusion from `indexes()` is intentional and consistent across PG/MySQL/SQLite (PG since 2005), and `indexes()` feeds the schema dumper, so changing it would require coordinated changes to all three adapters plus the dumper. Short-circuiting `find_unique_index_for` when `unique_by` is nil felt like the narrower fix. Happy to take the adapter-level path instead if you'd prefer — just let me know.

skipkayhil:
👋 I commented on the linked issue, is this not already fixed?

Edit: and yes, thank you for responding to my previous review, I agree with your assessment that's not the right change to make

hammadxcm:
One more clarification — I want to be transparent about a scope question I should have flagged earlier.

This PR changes a test added by #56666 (`test_insert_all_and_upsert_all_raises_when_no_unique_index_found_for_composite_primary_key`). That test explicitly asserted `Cart.insert_all` and `Cart.upsert_all` (without `unique_by`) raise `ArgumentError` for divergent-PK models, and #56666's description stated *"insert_all (on_duplicate: :skip) and upsert_all (on_duplicate: :update) still require the unique index for generating the conflict target clause."* I've replaced that with a test asserting success.

My reasoning for the change:

**`:skip` (`insert_all`)** doesn't actually need a conflict target. The adapter SQL builders interpolate `conflict_target` directly:

```ruby
# activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:693
sql << " ON CONFLICT #{insert.conflict_target} DO NOTHING"
```

When `conflict_target` is `nil`, this produces `ON CONFLICT DO NOTHING`, which is valid PostgreSQL/SQLite. So skipping the unique-index lookup for `:skip` is safe SQL-wise.

**`:update` (`upsert_all`)** already has a fallback in `conflict_target` (`activerecord/lib/active_record/insert_all.rb:275–276`):

```ruby
def conflict_target
  if index = insert_all.unique_by
    sql = +"(#{format_columns(index.columns)})"
    sql << " WHERE #{index.where}" if index.where
    sql
  elsif update_duplicates?
    "(#{format_columns(insert_all.primary_keys)})"  # <- falls back to schema_cache primary_keys
  end
end
```

Where `primary_keys` comes from `schema_cache.primary_keys(table_name)` — the actual table's PK, not `model.primary_key`. So when `unique_by` is nil and `find_unique_index_for` returns nil, `:update` produces `ON CONFLICT (shop_id, id) DO UPDATE SET ...` for Cart — using the table's real PK columns. That's the correct behavior on the DB side.

So the strict `find_unique_index_for` lookup was guarding against an SQL invalidity that doesn't actually exist for these two paths. That's why I felt comfortable expanding the scope #56666 set.

That said — if the intent in #56666 was to preserve the strict requirement as user-facing API (e.g., to force users to think about which constraint they're conflicting on), I'd defer to that. Happy to narrow the PR to only handle the standard-PK case (the actual #56953 reproducer) and leave Cart-style divergent-PK behavior unchanged. Just let me know which direction you'd prefer.

cc @kirs since this touches your PR's scope.

skipkayhil:
Ah, something I was missing was `ON CONFLICT DO NOTHING` is allowed without a conflict target, I thought _all_ `ON CONFLICT`s required an index:

> Specifies which conflicts ON CONFLICT takes the alternative action on by choosing arbiter indexes. Either performs unique index inference, or names a constraint explicitly. For ON CONFLICT DO NOTHING, it is optional to specify a conflict_target; when omitted, conflicts with all usable constraints (and unique indexes) are handled. For ON CONFLICT DO UPDATE, a conflict_target must be provided.

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+*   Fix `insert_all` raising `ArgumentError: No unique index found for id`
+    when `unique_by` is not provided and the adapter's `indexes()` method
+    excludes primary key indexes (e.g. PostgreSQL).
+
+    When `unique_by` is nil, `find_unique_index_for` now returns `nil` early
+    instead of trying to match the primary key against `unique_indexes`.
+    This allows `insert_all` to generate `ON CONFLICT DO NOTHING` without a
+    conflict target, and `upsert_all` to fall back to `primary_keys` for
+    the conflict target.
```
skipkayhil: Too many internals leaking into this changelog: we shouldn't mention `find_unique_index_for` at all for example, and I don't think "`indexes()` excluding the primary key" is correct based on what we discussed before.

activerecord/test/cases/insert_all_test.rb
```diff
@@ -507,13 +563,12 @@ def test_upsert_all_does_notupdates_existing_record_by_when_there_is_no_key
     assert_equal "Very fast", Speedometer.find("s3").name
   end
 
-  def test_upsert_all_updates_existing_record_by_configured_primary_key_fails_when_database_supports_insert_conflict_target
+  def test_upsert_all_fails_for_configured_primary_key_with_no_database_constraint
     skip unless supports_insert_on_duplicate_update? && supports_insert_conflict_target?
 
-    error = assert_raises ArgumentError do
+    assert_raises ActiveRecord::StatementInvalid do
```
skipkayhil: This change seems wrong to me: isn't the whole point of that `ArgumentError` being raised to protect from this scenario?

activerecord/test/cases/insert_all_test.rb
```diff
@@ -317,29 +317,85 @@ def test_insert_all_bang_does_not_require_unique_index
...
+
+      Cart.insert_all! [{ id: 2, shop_id: 1, title: "My cart 2" }]
+
+      Cart.upsert_all [{ id: 3, shop_id: 2, title: "My other cart" }]
+    end
+  end
+
+  # Regression tests for https://github.com/rails/rails/issues/56953
+  # insert_all without unique_by should not raise "No unique index found for id"
+  # even when the adapter's indexes() method excludes primary key indexes.
```
skipkayhil:
```suggestion
```

skipkayhil (changes requested):

hammadxcm:
Thanks for the detailed review @skipkayhil — pushed [d9cac63](https://github.com/hammadxcm/rails/commit/d9cac63ac0) addressing all three points:

**CHANGELOG**: Removed the internal method name and the `indexes()` framing. New entry:

> Fix `insert_all` raising `ArgumentError: No unique index found for id` when `unique_by` is not specified. `insert_all` uses `ON CONFLICT DO NOTHING` which does not require a conflict target, so the unique index lookup is skipped when `unique_by` is not specified.

**Narrowed the code fix**: The new branch is `unique_by.nil? && skip_duplicates?` — only `:skip` mode (`insert_all`) skips the index lookup. The `:update` path (`upsert_all`) is unchanged, so `Speedometer.upsert_all` without `unique_by` still raises `ArgumentError` as before.

**Speedometer test**: Reverted back to `assert_raises ArgumentError`.

**Cart test**: Removed `Cart.upsert_all` from the composite-PK test since `upsert_all` still raises in the narrowed fix; test now only covers `insert_all` and `insert_all!`.

PR #56995, opened 2026-03-16
Incomplete fix for CVE-2025-24293: ActiveStorage Vips transformer to RCE #56948

nicolasva:
issue #56948

Summary of changes for CVE-2025-24293 (Issue #56948)
Problem
The CVE-2025-24293 fix only protected the ImageMagick transformer, but not the Vips transformer (default since Rails 7.0). An attacker could achieve RCE via blob.variant(instance_eval: "malicious code") because image_processing gem dispatches operations through public_send.

**Fix**
File | Change
-- | --
image_processing_transformer.rb | Moved method whitelist validation + recursive argument validation from ImageMagick to the base class, protecting all transformers
image_magick.rb | Removed duplicated validation code (now inherited from parent)

**Tests added**
File | Type | Tests
-- | -- | --
image_processing_transformer_test.rb | Unit (no DB) | 17 tests
variant_test.rb | Integration (with DB) | 7 tests

**Test coverage**
Test | Scenario | Expected | Status
-- | -- | -- | --
instance_eval (RCE vector) | variant(instance_eval: "id > /tmp/pwned") | UnsupportedImageProcessingMethod | PASS
system | variant(system: "touch /tmp/dangerous") | UnsupportedImageProcessingMethod | PASS
__send__ / send / public_send | Reflection method bypass | UnsupportedImageProcessingMethod | PASS
Dangerous -write string | variant(resize: "-write /tmp/file") | UnsupportedImageProcessingArgument | PASS
Dangerous -path string (case insensitive) | variant(resize: "-PaTh /tmp/file") | UnsupportedImageProcessingArgument | PASS
Dangerous argument in array | variant(resize: [123, "-write", "/tmp/file"]) | UnsupportedImageProcessingArgument | PASS
Dangerous argument in nested array | variant(resize: [123, ["-write", "/tmp/file"]]) | UnsupportedImageProcessingArgument | PASS
Dangerous argument in hash key | variant(resize: { "-write": "/tmp/file" }) | UnsupportedImageProcessingArgument | PASS
Dangerous argument in hash value | variant(resize: { something: "-write /tmp/file" }) | UnsupportedImageProcessingArgument | PASS
Dangerous argument in nested hash | variant(resize: { something: { "-write": "/tmp/file" } }) | UnsupportedImageProcessingArgument | PASS
Dangerous argument in array inside hash | variant(resize: { something: ["-write", "/tmp/file"] }) | UnsupportedImageProcessingArgument | PASS
Valid transformation | variant(resize_to_limit: [100, 100]) | Passes normally | PASS


activestorage/lib/active_storage/transformers/image_processing_transformer.rb
```diff
@@ -42,6 +42,62 @@ def validate_transformation(name, argument)
               as it always generates a single command.
             ERROR
           end
+
+          method_name = name.to_s.tr("-", "_")
+
+          unless ActiveStorage.supported_image_processing_methods.include?(method_name)
```
flavorjones: The `Vips` transformer needs its own allowlist derived from libvips operation names. The ImageMagick list cannot be shared, because the two backends do not use the same names.

flavorjones (changes requested):
There are (at least) two failing tests with this approach:

- [`optimized variation of GIF blob`](https://github.com/rails/rails/blob/04aeb87acdf7559d2e82b692ec7e4e775ffc845e/activestorage/test/models/variant_test.rb#L133) calls `blob.variant(saver: { optimize_gif_frames: true, optimize_gif_transparency: true })`.
- [`thumbnail variation of JPEG blob processed with VIPS`](https://github.com/rails/rails/blob/04aeb87acdf7559d2e82b692ec7e4e775ffc845e/activestorage/test/models/variant_test.rb#L177) calls `blob.variant(thumbnail_image: 100)`.

Both fail because this change gates the `Vips` transformer with `ActiveStorage.supported_image_processing_methods`, which is a list of ImageMagick option names.

libvips uses different operation names, so `saver` and `thumbnail_image` are examples of safe libvips operations that are absent from that list and now get rejected.

I think this approach is directionally correct, but it needs:

- a canonical Vips-specific allowlist (which may require some effort to generate)
- each backend needs its own configurable allowlist

That second bullet point is a bit vague; I'm not sure whether to keep `supported_image_processing_methods` as the ImageMagick list and add a Vips counterpart, or try to do something more complex that overrides `supported_image_processing_methods`.

nicolasva:
activestorage/lib/active_storage/transformers/vips.rb
Lines: +5
Nature of change: Implements the hook → returns
ActiveStorage.supported_vips_image_processing_methods.
This is the core of the fix: previously, the Vips transformer performed no validation at all (RCE
possible via instance_eval, system…)
────────────────────────────────────────
activestorage/lib/active_storage/transformers/image_magick.rb
Lines: +4
Nature of change: Implements the hook → returns the existing ImageMagick allowlist
(supported_image_processing_methods) — behavior unchanged
────────────────────────────────────────
activestorage/lib/active_storage/engine.rb
Lines: +1
Nature of change: New config option: config.active_storage.supported_vips_image_processing_methods to

extend the Vips allowlist
────────────────────────────────────────
activestorage/test/transformers/image_processing_transformer_test.rb
Lines: +77 / −97
Nature of change: Tests rewritten and parameterized over both backends (Vips + MiniMagick): rejection

of Ruby reflection methods (instance_eval, send, system…), pipeline methods (apply, custom),
combine_options, and dangerous arguments (-write, -path) including nested ones
────────────────────────────────────────
activestorage/CHANGELOG.md
Lines: +12
Nature of change: Entry describing the fix, the new allowlist, and the config option

Summary: the original CVE-2025-24293 fix only covered the ImageMagick transformer; this branch
extends transformation method validation to the Vips transformer via a dedicated allowlist matching
the libvips vocabulary, extensible through configuration.

activestorage/lib/active_storage/transformers/image_processing_transformer.rb
nicolasva: It is fixed @flavorjones

janko:
I believe this RCE vulnerability was patched in ImageProcessing 2.x versions.

flavorjones:
@janko The 2.x guard closes one vector, but there are others. 

`ImageProcessing.unsafe_method?` rejects names whose owner is `BasicObject`, `Kernel`, `Object` or `Module`, which stops `send`, `instance_eval` and friends. It doesn't catch libvips's own methods, since those are owned by `Vips::Image`:

```ruby
img.method("send").owner        # => Kernel        blocked
img.method("write_to_file").owner  # => Vips::Image   allowed
```

libvips exposes a number of file-writing operations, and an attacker-supplied transformation name still reaches them on 2.0.3. The gem closed the reflection route; an allowlist is still needed.

nicolasva:
## Update

This PR has now been rebased onto the latest `rails/rails:main` and the merge conflict has been resolved.

### Implementation

- Moved transformation method and argument validation from the `ImageMagick` transformer into the shared `ImageProcessingTransformer` base class, so all image-processing backends follow the same security path.
- Added a backend-specific `supported_methods` hook. The base class fails closed with `NotImplementedError` when a custom backend does not provide an allowlist.
- Kept the existing ImageMagick allowlist for `MiniMagick`.
- Added `ActiveStorage.supported_vips_image_processing_methods`, a Vips-specific allowlist derived from libvips operations (229 supported entries).
- Added `config.active_storage.supported_vips_image_processing_methods` so applications can extend the Vips allowlist without changing the ImageMagick list.
- Removed unsafe or invalid Vips entries from the allowlist, including filesystem loaders/writers, ICC file operations, `text` file loading, image constructors that cannot be used as instance operations, and metadata mutators.
- Preserved the existing `saver` and `thumbnail_image` operations required by the current Vips GIF and thumbnail workflows.
- Kept recursive validation of strings, arrays, hashes, and nested combinations against `unsupported_image_processing_arguments`.
- Continued rejecting `combine_options`, and added validation coverage for ImageProcessing pipeline methods that could bypass normal transformation validation (`apply`, `operation`, `branch`, `source`, `call`, `custom`, and `instrumenter`).

### Tests and documentation

- Added Vips security tests covering:
  - Ruby reflection and shell methods such as `instance_eval`, `send`, `public_send`, `system`, and `exec`.
  - Arbitrary file-writing methods such as `write_to_file` and `write_to_target`.
  - Vips filesystem operations including loaders, savers, ICC operations, and profile/text loading.
  - Dangerous arguments in strings, arrays, hashes, and nested structures.
  - Separation between Vips and ImageMagick allowlists.
  - Safe existing transformations such as `thumbnail_image`, `saver`, `resize_to_*`, and `colourspace`.
- Reduced the integration coverage in `variant_test.rb` to focused Vips regression cases while keeping the existing behavior checks.
- Added a Railties configuration test for `config.active_storage.supported_vips_image_processing_methods`.
- Updated the Active Storage changelog.

Local verification completed successfully:

- 74 targeted security and regression assertions passed.
- RuboCop reported no offenses.
- Ruby syntax and whitespace checks passed.

The PR is now mergeable. The Buildkite checks are currently running on the rebased commit.

PR #57004, opened 2026-03-18
Reimplement RedisCache store using `redis-client`

byroot:
I've been meaning to do this for several years, but `Redis::Distributed` was a blocker. So I implemented `RedisClient.ring` to offer the same feature.

`redis-client` is configured to use RESP2 protocol, so there's no new server requirement compared to the previous version (that being said the first RESP3 release was 6 years ago).

To keep supporting the `redis:` configuration option, I decided to rename the old store as `DeprecatedRedisCacheStore`, to allow for a smoother transition. It does bloat the diff size though.

### Todo

- [x] ~~For now I've removed the `:redis` constructor argument, but I suspect some people do rely on it to pass a Redis cluster client, so I don't think I can actually entirely remove this option.~~

### DIFF
--- a/activesupport/lib/active_support/cache/redis_cache_store.rb
+++ b/activesupport/lib/active_support/cache/redis_cache_store.rb
@@ -1,15 +1,17 @@
 # frozen_string_literal: true
 
+redis_client_min_version = "0.28.0"
 begin
-  gem "redis", ">= 4.0.1"
-  require "redis"
-  require "redis/distributed"
+  gem "redis-client", ">= #{redis_client_min_version}"
 rescue LoadError
-  warn "The Redis cache store requires the redis gem, version 4.0.1 or later. Please add it to your Gemfile: `gem \"redis\", \">= 4.0.1\"`"
-  raise
+  raise LoadError, <<~MSG
+    The Redis cache store requires the redis-client gem version #{redis_client_min_version} or later.
+    Please add it to your Gemfile:
+      gem "redis-client", ">= #{redis_client_min_version}"
+  MSG
 end
 
-require "connection_pool"
+require "redis-client"
 require "active_support/core_ext/array/wrap"
 require "active_support/core_ext/hash/slice"
 require "active_support/core_ext/numeric/time"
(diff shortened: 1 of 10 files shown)


activesupport/lib/active_support/cache/redis_cache_store.rb
```diff
@@ -60,82 +62,48 @@ class RedisCacheStore < Store
...
       def self.supports_cache_versioning?
         true
       end
 
-      prepend Strategy::LocalCache
+      def self.new(**options)
+        if options[:redis]
+          ActiveSupport.deprecator.warn(<<~MSG.squish)
+            Passing a Redis or ConnectionPool instance via the `:redis` configuration to ActiveSupport::Cache::RedisCacheStore
+            is deprecated and will be removed in Rails 8.2.
```
rafaelfranca:
```suggestion
            is deprecated and will be removed in Rails 8.3.
```

activesupport/test/cache/behaviors/cache_instrumentation_behavior.rb
```diff
@@ -107,7 +107,7 @@ def test_delete_multi_instrumentation
 
   def test_increment_instrumentation
     key_1 = SecureRandom.uuid
-    @cache.write(key_1, 0)
+    @cache.write(key_1, 0, raw: true)
```
rafaelfranca, line 110: It isn't clear why this change was needed.

rafaelfranca (approved):

activesupport/test/cache/behaviors/cache_instrumentation_behavior.rb
byroot: `write(k, 0)` will store `Marshal.dump(0)`, and Redis will refuse to increment that:

```
127.0.0.1:6379> SET foo bar
OK
127.0.0.1:6379> INCR foo
(error) ERR value is not an integer or out of range
```

When you want to initialize a counter you have to use `raw` that is true of all adapters.

But it's a good question, I'm actually unsure how that even worked before. I'll have a look.

byroot: Ok, so it was failing before, but the error was handled by the `failsafe` handler, because the old store rescue all redis errors, which include `CommandError`.

The `failsafe` of the new store only rescue connection errors.

I believe it's a better behavior, but I can rescue all errors if you feel strongly about it.

rafaelfranca: I think this is a better behavior as well, it was just not clear.

rafaelfranca (approved):

PR #57012, opened 2026-03-19
fix: breaking support of different table names

djezzzl:
### Motivation / Background

Hi, during our upgrade from Rails 7.0 to Rails 8.0 (and even Rails 8.1 and greater, though with different behavior), we noticed a **breaking change** in ActiveRecord.

Similar fix was already merged: https://github.com/rails/rails/pull/56828

Copying the reproducing step below:

The following code should reproduce the broken flow:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 8.0.0"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  create_table "table-with-hyphen", force: true do |t|
  end
end

  def test_insert_exec
    assert_nothing_raised do
      ActiveRecord::Base.connection.exec_insert("INSERT INTO [select] DEFAULT VALUES")
    end
  end
end
```

### Detail

In this Pull Request, we add support for many missing table names across different adapters (e.g., MSSQL) when quoted properly. 

### Additional information

Adding the following code to the example script makes it green.
Related commit: https://github.com/rails/rails/commit/9bbbf4b8b9bf3d7cb31c0425a6646cec49e4bc6c

Previously, `exec_insert` didn't try to derive private keys from the schema via parsing the table name from the SQL query. 

```ruby
# Rails 8.0 throws an error as primary_key for unknown table_ref raises an error
pk = primary_key(table_ref) if table_ref
```

I see that in main, we try to hit the schema cache instead, and return an empty private key, which is okay, though the returning statement is missing from the final SQL when it is expected to be there with a proper table name.   

```ruby
# in main branch, we access the cache, which is not throwning error when the table name is missing
pk = schema_cache.primary_keys(table_ref) if table_ref
```

```ruby
module ActiveRecord
  module ConnectionAdapters
    module DatabaseStatements
        if sql =~ /into\s((?:"(?:[^"]|"")+")|(?:\[(?:[^\]]|\]\])*\](?:\.\[(?:[^\]]|\]\])*\])*)|(?:[-A-Za-z0-9_.\[\]]+))\s*/im
          $1.gsub(/\A"|"\z/, "").gsub('""', '"').strip
        end
      end
    end
  end
end
```

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
@@ -785,8 +785,29 @@ module ActiveRecord
         end
 
         def extract_table_ref_from_insert_sql(sql)
-          if sql =~ /into\s("[-A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
-            $1.delete('"').strip
+          quoted_ident  = /"(?:[^"]|"")*"/
+          bracket_ident = /\[(?:[^\]]|\]\])*\]/
+          bare_ident    = /[-A-Za-z0-9_]+/
+          ident_part    = /(?:#{quoted_ident.source}|#{bracket_ident.source}|#{bare_ident.source})/
+
+          into_target = /
+            \binto\s+
+            (?:
+              (?<qualified>#{ident_part.source}(?:\.#{ident_part.source})+) |
+              (?<single_quoted>#{quoted_ident.source}) |
+              (?<single_plain>#{bracket_ident.source}|#{bare_ident.source})
+            )
+            (?=\s|$|[,(;])
+          /imx
+
+          if (m = sql.match(into_target))
+            if m[:qualified]
+              m[:qualified].strip
+            elsif m[:single_quoted]
+              m[:single_quoted][1..-2].gsub('""', '"').strip
+            else
+              m[:single_plain].strip
+            end
           end
         end
     end
--- a/activerecord/test/cases/database_statements_test.rb
+++ b/activerecord/test/cases/database_statements_test.rb
@@ -22,9 +22,47 @@ class DatabaseStatementsTest < ActiveRecord::TestCase
     assert_not_nil return_the_inserted_id(method: :create)
   end
 
-  def test_extract_table_ref_from_insert_sql_with_hyphen_in_table_name
-    sql = "INSERT INTO \"table-with-hyphen\" (column1, column2) VALUES (value1, value2)"
-    assert_equal "table-with-hyphen", @connection.send(:extract_table_ref_from_insert_sql, sql)
+  def test_extract_table_ref_from_insert_sql_with_quoting
+    [
+      "table-name",
+      "table with space",
+      "a" * 120, # long name
+      "таблица",
+      "日本語テーブル",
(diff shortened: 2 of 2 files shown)


djezzzl:
Hi @byroot, could you please help with this similar fix to the original issue fixed here: https://github.com/rails/rails/pull/56828?

djezzzl:
From quick glancy, `trilogy` test doesn't seem to be related:

```

Error:
--
ActiveRecord::TransactionTest#test_reconnect_preserves_isolation_level:
ActiveRecord::StatementInvalid: Trilogy::QueryError: trilogy_read_row: TRILOGY_TRUNCATED_PACKET
lib/active_record/connection_adapters/trilogy/database_statements.rb:51:in `query'
lib/active_record/connection_adapters/trilogy/database_statements.rb:51:in `perform_query'
lib/active_record/connection_adapters/abstract/database_statements.rb:621:in `block (2 levels) in execute_intent'
lib/active_record/connection_adapters/abstract_adapter.rb:1095:in `block in with_raw_connection'
/rails/activesupport/lib/active_support/concurrency/null_lock.rb:9:in `synchronize'
```

Could it be on the main branch?

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -785,8 +785,29 @@ def arel_from_relation(relation)
         end
 
         def extract_table_ref_from_insert_sql(sql)
-          if sql =~ /into\s("[-A-Za-z0-9_."\[\]\s]+"|[A-Za-z0-9_."\[\]]+)\s*/im
-            $1.delete('"').strip
+          quoted_ident  = /"(?:[^"]|"")*"/
+          bracket_ident = /\[(?:[^\]]|\]\])*\]/
+          bare_ident    = /[-A-Za-z0-9_]+/
+          ident_part    = /(?:#{quoted_ident.source}|#{bracket_ident.source}|#{bare_ident.source})/
```
byroot, line 791: Compiling regexp is a costly operation. It shouldn't be done inside a method like that unless there's no other way. All these should be constants.

activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb
```diff
@@ -785,8 +785,29 @@ def arel_from_relation(relation)
...
+
+          into_target = /
+            \binto\s+
+            (?:
+              (?<qualified>#{ident_part.source}(?:\.#{ident_part.source})+) |
+              (?<single_quoted>#{quoted_ident.source}) |
+              (?<single_plain>#{bracket_ident.source}|#{bare_ident.source})
+            )
+            (?=\s|$|[,(;])
+          /imx
```
byroot, line 801: Overall this is way too complicated of a regexp, no way we merge that.

Parsing SQL with a regexp was already not a good idea (there is a comment to that effect) but using such a complicated regexp is a big NO NO.

We should find another solution, so that we pass the table name directly. What might help is the recent `Intent` refactoring, now we should have the Arel AST most of the time, we probably can extract the table name from it without murking with regexps.

There might be other solutions.

PR #57013, opened 2026-03-19
Predefine the well-known Postgres type OIDs

matthewd:
We currently query `pg_type` every time we connect to a Postgres server, to get the mapping from type names to its internal numeric OID values.

For user-defined types, and extensions (including e.g. `hstore`, which we support out of the box), that's necessary: OIDs are unique to the server instance, and don't even survive a DB dump/load.

However, all the core built-in types actually have statically allocated OIDs that are shipped in the Postgres source.

Instead of querying immediately, we can ship our own copy of the well-known OID mappings, and load that upon connect for free.

The first time we encounter an unknown type, we'll _then_ run the same "load all the type definitions we're likely to need" semi-heuristic query that we used to run eagerly... as well as continuing to follow-up with one-shot queries when even that fails us.


activerecord/lib/active_record/connection_adapters/postgresql/oid.rb
```diff
@@ -27,6 +27,7 @@
 require "active_record/connection_adapters/postgresql/oid/xml"
 
 require "active_record/connection_adapters/postgresql/oid/type_map_initializer"
+require "active_record/connection_adapters/postgresql/oid/well_known"
```
eileencodes, line 30: I'm a little hung up on the well known naming since what defines well known? Would builtin_oids be better?

activerecord/lib/active_record/connection_adapters/postgresql/oid/well_known_generator.rb
```diff
@@ -0,0 +1,449 @@
...
+                if cache
+                  cached = read_cached_catalog(url)
+                  return cached if cached
+                end
+
+                uri = URI.parse(url)
+                raise ArgumentError, "catalog URL must use HTTPS: #{url}" unless uri.is_a?(URI::HTTPS)
+
+                request = Net::HTTP::Get.new(uri)
+                response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(request) }
```
byroot:
```suggestion
                response = Net::HTTP.get_response(uri)
```

activerecord/lib/active_record/connection_adapters/postgresql/oid/well_known_values.rb
```diff
@@ -0,0 +1,218 @@
...
+              3802 => "jsonb",
+              3904 => "int4range",
+              3906 => "numrange",
+              3908 => "tsrange",
+              3910 => "tstzrange",
+              3912 => "daterange",
+              3926 => "int8range",
+              4089 => "regnamespace",
+              4096 => "regrole",
+            }.dup
```
byroot: That `dup` seems useless.

matthewd: Ah, yeah, at one point this was pulling from a constant, but I decided that wasn't helpful

activerecord/lib/active_record/connection_adapters/postgresql/oid.rb
matthewd: I'm not sure. 

I'm using it as a term of art, not a value judgement, a la https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports / https://en.wikipedia.org/wiki/Well-known_URI

Postgres docs refer to them as 'pre-loaded catalog rows' and 'preassigned OIDs' ([68.2.2. OID Assignment](https://www.postgresql.org/docs/18/system-catalog-initial-data.html#SYSTEM-CATALOG-OID-ASSIGNMENT)), but IMO those would here feel like they're describing the fact _we've_ built them in, rather than focusing on why it's safe (because they're defined to be universal).

The closest they get to using "well-known" is ["the well-known built-in types"](https://www.postgresql.org/docs/18/infoschema-columns.html#:~:text=the%20well%2Dknown%20built%2Din%20types) [closely related, but not strictly talking about OIDs], though pgJDBC does match our usage: https://github.com/pgjdbc/pgjdbc/blob/1c586bb0da65389803e0b4254b60bc381b5ab4a6/pgjdbc/src/main/java/org/postgresql/core/Oid.java#L18

eileencodes: ok that's good enough for me. They're nodoc'd too so not really an issue with a public interface at least.

eileencodes (approved):

PR #57015, opened 2026-03-19
add a sanitize_url_string helper

gregmolnar:
### Motivation / Background

The `sanitize` helper removes unsafe protocols from the src/href attribute of HTML elements, but if you just want to remove those from a raw URL string, `sanitize` wouldn't help. This actual misled me(and others) in the past and I was under the impression that I mitigated an XSS issue(https://github.com/rubyevents/rubyevents/pull/1558), but I didn't. 
This pull request introduces a `sanitize_url` helper that can be used in the quite common situation of using `link_to` with a block. In case of an unsafe protocol in the input, it returns a fallback that defaults to "/".

### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Add `sanitize_url_string` helper to prevent unsafe protocols in URLs. Example
+    usage:
+    ```
+        sanitize_url_string("javascript:alert(1)", fallback: root_url)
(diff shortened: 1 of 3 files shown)


gregmolnar:
@morgoth pointed out that `sanitize_url` might conflict with a route helper, so I renamed the method to `sanitize_url_string`. If someone has a better idea for a name, I am all ears.

PR #57021, opened 2026-03-20
De-anonymize generated attribute module for Store

hmcguire-shopify:
(based on #57020)

```
❯ bin/rails c
Loading development environment (Rails 8.1.2)
Lobsters(dev):001> User.ancestors
=>
[User (call 'User.load_schema' to load schema informations),
 #<Module:0x0000000125bfcf40>,
```

Previously, ActiveRecord::Store would include an anonymous module into
models which holds the Store's generated attributes.

This commit names the module to make it obvious what the module is in
the list of ancestors, and then uses the name to reference it instead of
storing a reference to the module in an instance variable on the
singleton class.

### DIFF
--- a/activerecord/lib/active_record/store.rb
+++ b/activerecord/lib/active_record/store.rb
@@ -103,7 +103,7 @@ module ActiveRecord
     end
 
     module ClassMethods
-      def inherited(subclass) # :nodoc
+      def inherited(subclass) # :nodoc:
         super
         subclass.instance_variable_set(:@local_stored_attributes, nil)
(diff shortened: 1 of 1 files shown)


eileencodes (approved):

activerecord/lib/active_record/store.rb
```diff
@@ -103,6 +103,11 @@ class << self
     end
 
     module ClassMethods
+      def inherited(subclass) # :nodoc
```
eileencodes: Missed this when I approved.

```suggestion
      def inherited(subclass) # :nodoc:
```

PR #57023, opened 2026-03-21
Reuse loaded associations for has_one :through

Mordorreal:
## Motivation

Fixes #56978

`has_one :through` fires a SQL query even when both the intermediate and target associations are already loaded in memory. This commonly occurs when traversing an object graph loaded via `includes` from the other side of the relationship.

```ruby
accounts = Account.includes(reports: :issues).to_a
issue = accounts.first.reports.first.issues.first

issue.report                # => no query (loaded via inverse)
issue.report.account        # => no query (loaded via inverse)
issue.account               # => fires SELECT! (has_one :through ignores loaded chain)
```

The workaround is `delegate :account, to: :report` or manual chaining, but then you lose composability in scopes, `where` clauses, and `joins`.

## Solution

Override `find_target` in `HasOneThroughAssociation` to check if the through chain is already loaded in memory before hitting the database.

When `owner.association(through_reflection.name)` is loaded:
- If the through record exists and its source association is loaded → return the in-memory target
- If the through record is nil → return nil
- Otherwise → fall through to the existing query path

This mirrors what you'd get from manually chaining `issue.report.account`, making `has_one :through` consistent with manual traversal and `delegate`.

## Benchmark Results

Ruby 4.0.1, SQLite3 in-memory database.

### Speed (benchmark-ips)

**Before (N+1 on every access):**
```
has_one :through (preloaded chain):     1,092 i/s  (915.73 μs/i)
has_one :through (not preloaded):       2,722 i/s  (367.39 μs/i)
manual chain (preloaded, baseline):     1,768 i/s  (565.66 μs/i)
```

**After (in-memory traversal):**
```
has_one :through (preloaded chain):     1,609 i/s  (621.59 μs/i)  ~47% faster
has_one :through (not preloaded):       2,497 i/s  (400.55 μs/i)  no regression
manual chain (preloaded, baseline):     1,681 i/s  (594.78 μs/i)  unchanged
```

### Memory (memory_profiler, 100 iterations)

| Scenario | Before | After |
|---|---|---|
| Preloaded chain | 1,680,200 bytes / 20,302 obj | **8,160 bytes / 201 obj** (99.5% reduction) |
| Not preloaded | 2,492,000 bytes / 27,000 obj | 2,524,000 bytes / 27,200 obj (unchanged) |
| Manual chain | 0 bytes / 0 obj | 0 bytes / 0 obj |

### Query Count

| Access pattern | Before | After |
|---|---|---|
| `issue.account` (preloaded chain) | 1 query (N+1) | **0 queries** |
| `issue.account` (not preloaded) | 1 query | 1 query (unchanged) |
| `issue.report.account` (manual) | 0 queries | 0 queries |

## Test Plan

- [x] Test: preloaded intermediate chain → zero queries
- [x] Test: through record loaded as nil → returns nil, zero queries
- [x] Test: through loaded but source not loaded → falls back to query (existing behavior)
- [x] Test: nothing loaded → existing query behavior unchanged
- [x] All existing `has_one_through_associations_test.rb` tests pass (56 tests)
- [x] Full `activerecord:sqlite3:test` suite passes (9,381 tests, 0 failures)
- [x] RuboCop passes on changed files

### DIFF
--- a/activerecord/lib/active_record/associations/has_one_through_association.rb
+++ b/activerecord/lib/active_record/associations/has_one_through_association.rb
@@ -6,7 +6,31 @@ module ActiveRecord
     class HasOneThroughAssociation < HasOneAssociation # :nodoc:
       include ThroughAssociation
 
+      def reader
+        if !loaded? && !stale_target? && through_chain_loaded?
+          self.target = resolve_target_from_through_chain
+        end
+
+        super
+      end
+
       private
+        def through_chain_loaded?
+          through_assoc = through_association
+          return false unless through_assoc.loaded?
+
+          through_target = through_assoc.target
+          return true unless through_target
+
+          through_target.association(source_reflection.name).loaded?
+        end
+
+        def resolve_target_from_through_chain
+          through_target = through_association.target
+          return nil unless through_target
+          through_target.association(source_reflection.name).target
+        end
+
         def replace(record, save = true)
           create_through_record(record, save)
           self.target = record
--- a/activerecord/test/cases/associations/has_one_through_associations_test.rb
+++ b/activerecord/test/cases/associations/has_one_through_associations_test.rb
@@ -477,6 +477,70 @@ class HasOneThroughAssociationsTest < ActiveRecord::TestCase
 
     assert_predicate(book.association(:order_agreement), :stale_target?)
   end
+
+  def test_has_one_through_reuses_loaded_intermediate_association
+    club = Club.all.merge!(includes: { membership: :member }).first
+    member = club.membership.member
+
+    assert_predicate member.association(:current_membership), :loaded?,
+      "intermediate association (current_membership) should be loaded"
+
+    assert_no_queries do
+      member.club
+    end
+  end
+
+  def test_has_one_through_returns_nil_when_through_is_loaded_but_nil
+    member = Member.create!(name: "No Membership")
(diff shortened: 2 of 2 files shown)


Mordorreal:
The Trilogy::SynchronizationError failure in CI is a pre-existing flaky test caused by the connection pool reaper thread accessing Trilogy connections from its background thread during test setup/teardown. It is unrelated to this PR's changes.

A fix is proposed in #57024.

tylerwood-io:
One edge case worth considering with the `find_target` override: since `find_target` is called from within the `reload` path (`reader` → `reload` → `reset` → `load_target` → `find_target`), an explicit `reload_club` or `association(:club).reload` will also resolve from the in-memory chain instead of querying the database. `reset` clears `self` (the `has_one :through` association), but the through chain lives on `owner.association(through_reflection.name)` — so after `reset`, `through_association.loaded?` is still true and `find_target` returns the cached value.

This means:
```ruby
member = Member.includes(current_membership: :club).first

# Someone updates the club in another process...

member.reload_club  # Still returns the stale in-memory value
```

An alternative is overriding `reader` instead, which only intercepts the read accessor path. `force_reload_reader` and explicit `reload` calls bypass it entirely, preserving the expectation that explicit reloads always hit the database. I took that approach in #57079.

Mordorreal:
Valid point about `find_target` and reload. Closing in favor of #57079 which already handles this correctly. Thanks!

PR #57026, opened 2026-03-22
Fix incompatibility in ActionView with ruby 3.3.0

hieuk09:
### Motivation / Background

This Pull Request has been created because in Ruby 3.3.0, anonymous rest parameter cannot be also used within block. Using latest Rails version (8.1.2) with ruby 3.3.0 raises the below error:

```
anonymous rest parameter is also used within block (SyntaxError)
```

### Detail

This Pull Request changes to use explicit parameter instead of rest parameter.

### Additional information

If Rails 8.1.2 doesn't support ruby 3.3.0, maybe we should update the restriction there.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/capture_helper.rb
+++ b/actionview/lib/action_view/helpers/capture_helper.rb
@@ -44,10 +44,10 @@ module ActionView
       #
       #   @greeting # => "Welcome to my shiny new web page! The date and time is 2018-09-06 11:09:16 -0500"
       #
-      def capture(*, **, &block)
+      def capture(*args, **kwargs, &block)
         value = nil
         @output_buffer ||= ActionView::OutputBuffer.new
-        buffer = @output_buffer.capture { value = yield(*, **) }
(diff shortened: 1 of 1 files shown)


byroot:
Patch looks fine, but I wonder why this wasn't caught on CI.

byroot:
Ah, you mean ruby 3.3.0 specifically? As in it was fix in 3.3.1?

hieuk09:
I have only tested with 3.3.0 (this is fixed in 3.3.9). Let me try other versions.

hieuk09:
Ah yes, this is fixed in 3.3.1.

byroot:
Right I reached the same conclusion.

I think I will make 3.3.1 the required version, because we can't really afford to test 3.3.0 specifically on CI, and so without CI coverage it isn't reasonably possible to ensure the rest of the framework does work with this limitation.

Sorry.

hieuk09:
Thanks, updating the ruby requirement is fine for me, too.

PR #57030, opened 2026-03-22
Add dependent: :detach for Active Storage and updated_at for blobs

Mordorreal:
## Motivation

Fixes #56950

When using `dependent: :detach` (or manually detaching), blobs become unattached but should remain available for a grace period before purging. The existing purge advice uses `created_at`, which means blobs attached for years and then detached are immediately purged — they appear "old" because `created_at` reflects when the blob was uploaded, not when it became unattached.

### 1. Add `dependent: :detach` option

`has_one_attached` and `has_many_attached` now accept `dependent: :detach`. When the record is destroyed, the attachment record is removed but the blob and its file on the service are preserved.

```ruby
  has_one_attached :avatar, dependent: :detach
end
```

Valid `dependent` options are now validated: `:purge_later`, `:detach`, or `false`. Invalid values raise `ArgumentError`.

### 2. Add `updated_at` to `active_storage_blobs`

A nullable `updated_at` column is added to the blobs table. The blob is touched when:
- An attachment is explicitly detached (`user.avatar.detach`)
- An attachment is destroyed via record destroy with `dependent: :detach`

The touch only fires for `dependent: :detach` — not for `dependent: false` where the user explicitly opted out of cleanup behavior.

`DetachMany` uses `touch_all` for a single batch `UPDATE ... WHERE id IN(...)` instead of N individual touches.

### 3. Updated purge guide

The recommended purge task now uses `COALESCE(updated_at, created_at)` instead of `created_at`:

```ruby
ActiveStorage::Blob.unattached.where(
  "COALESCE(active_storage_blobs.updated_at, active_storage_blobs.created_at) <= ?",
  2.days.ago
).find_each(&:purge_later)
```

### Migration Impact
```
Adding nullable updated_at column is a metadata-only operation
on PostgreSQL (instant) and MySQL (instant for nullable without default).
No table rewrite needed. Zero downtime.
```

### Touch Callback Overhead (single blob)
```
blob.touch (UPDATE updated_at):     4,998 i/s  (200 μs/i)
blob.reload (SELECT, baseline):     7,282 i/s  (137 μs/i)
Touch is 1.46x slower than a SELECT — but only runs once per detach.
```

### Batch touch_all vs N individual touches (DetachMany)
```
 1 blob:  touch_all 1.60x faster
 5 blobs: touch_all 6.62x faster
10 blobs: touch_all 12.54x faster
50 blobs: touch_all 43.45x faster
```

### Memory: 50 blobs, 100 iterations
```
N individual touches: 123,272,440 bytes (1,009,952 objects)
Single touch_all:       2,283,100 bytes (24,200 objects)    — 54x less memory
```

### Purge Query Comparison
```
WHERE created_at <= ? (old):                          3,481 i/s
WHERE COALESCE(updated_at, created_at) <= ? (new):    3,883 i/s
Same query plan. COALESCE adds no overhead.
```

## Test Plan

- [x] Test: `dependent: :detach` keeps blob on has_one record destroy
- [x] Test: `dependent: :detach` keeps blobs on has_many record destroy
- [x] Test: blob `updated_at` is touched when attachment is detached
- [x] Test: blob `updated_at` is NOT touched for `dependent: false`
- [x] Test: DetachMany touches blobs in a single batch query
- [x] Test: `ArgumentError` raised for invalid dependent options
- [x] Test: reflection shows `:detach` dependent for has_one and has_many
- [x] Full Active Storage test suite: 0 regressions
- [x] RuboCop passes on all changed files
- [x] Benchmarked: migration, touch overhead, batch touch_all, purge query, memory

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Add `dependent: :detach` option for `has_one_attached` and `has_many_attached`.
+
+    When a record is destroyed with `dependent: :detach`, the attachment record
+    is removed but the blob and its file on the service are preserved. This is
+    useful when you want to keep files around for restoration or auditing.
+
+    Also adds `updated_at` column to `active_storage_blobs` table. The blob is
+    touched when a detached attachment is destroyed, so detached blobs can be
+    safely purged based on when they became unattached rather than when they
+    were originally uploaded.
+
+    Invalid `dependent` options now raise `ArgumentError`.
+
+    *Denis Savchuk*
+
 *   Configurable maximum streaming chunk size
 
     Makes sure that byte ranges for blobs don't exceed 100mb by default.
@@ -27,7 +42,6 @@
     do not respect these metacharacters).
 
     *Mike Dalessio*
-
 
 *   Restore ADC when signing URLs with IAM for GCS
 
--- a/activestorage/db/migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
+++ b/activestorage/db/migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUpdatedAtToActiveStorageBlobs < ActiveRecord::Migration[8.1]
+  def change
+    add_column :active_storage_blobs, :updated_at, :datetime, precision: 6
+  end
+end
--- a/activestorage/db/update_migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
+++ b/activestorage/db/update_migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUpdatedAtToActiveStorageBlobs < ActiveRecord::Migration[8.1]
+  def change
+    add_column :active_storage_blobs, :updated_at, :datetime, precision: 6, if_not_exists: true
+  end
+end
--- a/activestorage/lib/active_storage/attached/model.rb
+++ b/activestorage/lib/active_storage/attached/model.rb
@@ -119,6 +119,7 @@ module ActiveStorage
       # <tt>active_storage_attachments.record_type</tt> polymorphic type column of
       # the corresponding rows.
       def has_one_attached(name, dependent: :purge_later, service: nil, strict_loading: false, analyze: nil)
+        validate_dependent_option(dependent)
         Attached::Model.validate_service_configuration(service, self, name) unless service.is_a?(Proc)
 
         generated_association_methods.class_eval <<-CODE, __FILE__, __LINE__ + 1
@@ -231,6 +232,7 @@ module ActiveStorage
       # <tt>active_storage_attachments.record_type</tt> polymorphic type column of
       # the corresponding rows.
       def has_many_attached(name, dependent: :purge_later, service: nil, strict_loading: false, analyze: nil)
+        validate_dependent_option(dependent)
         Attached::Model.validate_service_configuration(service, self, name) unless service.is_a?(Proc)
 
         generated_association_methods.class_eval <<-CODE, __FILE__, __LINE__ + 1
@@ -282,6 +284,13 @@ module ActiveStorage
         yield reflection if block_given?
         ActiveRecord::Reflection.add_attachment_reflection(self, name, reflection)
       end
+
+      private
+        def validate_dependent_option(dependent)
+          unless dependent.in?([ :purge_later, :detach, false ])
+            raise ArgumentError, "Invalid dependent option: #{dependent.inspect}. Valid options are :purge_later, :detach, or false"
+          end
+        end
     end
 
     class << self
--- a/activestorage/test/models/attached/many_test.rb
+++ b/activestorage/test/models/attached/many_test.rb
@@ -710,6 +710,41 @@ class ActiveStorage::ManyAttachedTest < ActiveSupport::TestCase
     end
   end
 
+  test "detaching dependent attachments on destroy keeps blobs" do
+    [ create_blob(filename: "funky.jpg"), create_blob(filename: "town.jpg") ].tap do |blobs|
+      @user.documents.attach blobs
+
+      assert_no_enqueued_jobs do
+        @user.destroy!
+      end
+
+      assert ActiveStorage::Blob.exists?(blobs.first.id)
+      assert ActiveStorage::Blob.exists?(blobs.second.id)
+      assert ActiveStorage::Blob.service.exist?(blobs.first.key)
+      assert ActiveStorage::Blob.service.exist?(blobs.second.key)
+    end
+  end
+
+  test "detaching many attachments touches blobs in a single query" do
+    blobs = 3.times.map { |i| create_blob(filename: "doc_#{i}.jpg") }
+    @user.documents.attach blobs
+
+    touch_queries = []
+    counter = ->(*args) {
+      payload = args.last
+      if payload[:sql] =~ /UPDATE.*active_storage_blobs.*updated_at/i
+        touch_queries << payload[:sql]
+      end
+    }
+
+    ActiveSupport::Notifications.subscribed(counter, "sql.active_record") do
+      @user.documents.detach
+    end
+
+    assert_equal 1, touch_queries.size,
+      "Expected 1 batch UPDATE, got #{touch_queries.size}: #{touch_queries.inspect}"
+  end
+
   test "duped record does not share attachments" do
     @user.highlights.attach [ create_blob(filename: "funky.jpg") ]
 
--- a/guides/source/active_storage_overview.md
+++ b/guides/source/active_storage_overview.md
@@ -462,6 +462,24 @@ class User < ApplicationRecord
 end
 ```
 
+By default, the `:dependent` option is set to `:purge_later`, which enqueues a
+background job to purge the blob when the record is destroyed. To keep blobs when
+the record is destroyed, use `dependent: :detach`:
+
+```ruby
+class User < ApplicationRecord
+  has_one_attached :avatar, dependent: :detach
+end
+```
+
+When the record is destroyed, the attachment record is removed but the blob and
+its file on the service are preserved. This is useful when you want to keep files
+around for restoration, auditing, or when using storage services without object
+versioning. See [Purging Unattached Uploads](#purging-unattached-uploads) for
+cleaning up detached blobs after a safe period.
+
+To prevent any automatic cleanup, use `dependent: false`.
+
 NOTE: Since Active Storage relies on polymorphic associations, and [polymorphic associations](./association_basics.html#polymorphic-associations) rely on storing class names in the database, that data must remain synchronized with the class name used by the Ruby code. When renaming classes that use `has_one_attached`, make sure to also update the class names in the `active_storage_attachments.record_type` polymorphic type column of the corresponding rows.
 
 [`has_one_attached`]: https://api.rubyonrails.org/classes/ActiveStorage/Attached/Model.html#method-i-has_one_attached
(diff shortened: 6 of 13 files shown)


Mordorreal:
The Trilogy::SynchronizationError failure in CI is a pre-existing flaky test caused by the connection pool reaper thread accessing Trilogy connections from its background thread during test setup/teardown. It is unrelated to this PR's changes.

A fix is proposed in #57024.

activestorage/db/migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUpdatedAtToActiveStorageBlobs < ActiveRecord::Migration[8.1]
+  def change
+    add_column :active_storage_blobs, :updated_at, :datetime, precision: 6
```
brendon, line 5: It might pay to guard this change in case people have already added this column themselves.

guides/source/active_storage_overview.md
```diff
@@ -1644,15 +1662,22 @@ by implementing the methods necessary to upload and download files to the cloud.
...
+There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads) or when using `dependent: :detach` and later destroying the parent record. You can query for unattached records using the [unattached scope](https://github.com/rails/rails/blob/8ef5bd9ced351162b673904a0b77c7034ca2bc20/activestorage/app/models/active_storage/blob.rb#L49). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
 
 ```ruby
 namespace :active_storage do
   desc "Purges unattached Active Storage blobs. Run regularly."
   task purge_unattached: :environment do
-    ActiveStorage::Blob.unattached.where(created_at: ..2.days.ago).find_each(&:purge_later)
+    ActiveStorage::Blob.unattached.where(
+      "COALESCE(active_storage_blobs.updated_at, active_storage_blobs.created_at) <= ?",
```
brendon: This is useful for upgrading, but going forward, the updated_at column will be set by rails on creation (matching the created_at column) so this won't be necessary. In my case I populated existing records with today's date as the updated_at value. Not sure that's a feasible option for everyone though.

brendon:
Thanks @Mordorreal, great to see this moving forward :)

activestorage/db/migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
Mordorreal: Thanks for catching this\! Looking at the existing pattern (e.g. `add_service_name_to_active_storage_blobs`), the migration should live in `db/update_migrate/` with a `column_exists?` guard — that's what `rails app:update` copies for upgrading apps. The current `db/migrate/` placement only works for fresh installs via `active_storage:install`. I'll move it to `db/update_migrate/` with the guard and keep the `db/migrate/` copy for fresh installs.

guides/source/active_storage_overview.md
Mordorreal: You're right that going forward new blobs will have `updated_at` set automatically. The `COALESCE` is mainly needed for the transition period — existing blobs will have `NULL` `updated_at` after running the migration, and not everyone will backfill. Since it adds no performance overhead (benchmarked), it seemed like the safest default for the guide. Happy to add a note clarifying this is primarily for backwards compatibility with pre-existing blobs.

brendon: Sounds good. Perhaps the guide should show the method without coalesce and just mention the possibility of using it in the upgrade notes?

activestorage/db/update_migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddUpdatedAtToActiveStorageBlobs < ActiveRecord::Migration[8.1]
+  def up
+    return unless table_exists?(:active_storage_blobs)
+
+    unless column_exists?(:active_storage_blobs, :updated_at)
```
brendon: You can use `, if_not_exists: true` instead of checking that the column exists first.

guides/source/active_storage_overview.md
Mordorreal: done. thank you.

activestorage/db/update_migrate/20260322000000_add_updated_at_to_active_storage_blobs.rb
Mordorreal: good suggestion. done.

guides/source/active_storage_overview.md
```diff
@@ -1646,13 +1664,16 @@ by implementing the methods necessary to upload and download files to the cloud.
...
+There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads) or when using `dependent: :detach` and later destroying the parent record. You can query for unattached records using the [unattached scope](https://github.com/rails/rails/blob/8ef5bd9ced351162b673904a0b77c7034ca2bc20/activestorage/app/models/active_storage/blob.rb#L49). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
 
 ```ruby
 namespace :active_storage do
   desc "Purges unattached Active Storage blobs. Run regularly."
   task purge_unattached: :environment do
-    ActiveStorage::Blob.unattached.where(created_at: ..2.days.ago).find_each(&:purge_later)
+    ActiveStorage::Blob.unattached.where(
+      "active_storage_blobs.created_at <= ?",
```
brendon: This is a strange regression in style. I think change it back to the way it was originally but use updated_at in the example instead.

guides/source/active_storage_overview.md
```diff
@@ -1646,13 +1664,16 @@ by implementing the methods necessary to upload and download files to the cloud.
 Purging Unattached Uploads
 --------------------------
 
-There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads). You can query for unattached records using the [unattached scope](https://github.com/rails/rails/blob/8ef5bd9ced351162b673904a0b77c7034ca2bc20/activestorage/app/models/active_storage/blob.rb#L49). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
+There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads) or when using `dependent: :detach` and later destroying the parent record. You can query for unattached records using the [unattached scope](https://github.com/rails/rails/blob/8ef5bd9ced351162b673904a0b77c7034ca2bc20/activestorage/app/models/active_storage/blob.rb#L49). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
```
brendon: The URL here should probably be tidied up. See other examples in this file.

brendon (commented):
Thanks for the updates @Mordorreal, I assume you're using AI to help? Overall things still look ok but just be careful with how it can regress.

Mordorreal:
> I assume you're using AI to help? Overall things still look ok but just be careful with how it can regress.


Yes, I do use AI assistance from time to time. English is not my native language, so it helps a lot with writing documentation and commit messages. I agree it requires extra attention — things like that style regression can slip through. But for me, the productivity and psychological wins outweigh the downsides. I'll be more careful reviewing the output before pushing. Thanks for flagging it.

guides/source/active_storage_overview.md
Mordorreal: fixed. thanks

guides/source/active_storage_overview.md
Mordorreal: fixed

brendon:
> > I assume you're using AI to help? Overall things still look ok but just be careful with how it can regress.
> 
> Yes, I do use AI assistance from time to time. English is not my native language, so it helps a lot with writing documentation and commit messages. I agree it requires extra attention — things like that style regression can slip through. But for me, the productivity and psychological wins outweigh the downsides. I'll be more careful reviewing the output before pushing. Thanks for flagging it.

Fair enough. I use it too. It's great for documentation, but when it comes to code, I've learned to pick everything over very carefully :)

I think this is ready for review by a core member now.

brendon:
@joshuay03?

activestorage/lib/active_storage/attached/model.rb
```diff
@@ -282,6 +284,13 @@ def #{name}=(attachables)
         yield reflection if block_given?
         ActiveRecord::Reflection.add_attachment_reflection(self, name, reflection)
       end
+
+      private
+        def validate_dependent_option(dependent)
+          unless dependent.in?([ :purge_later, :detach, false ])
```
joshuay03, line 290: Should `:purge` also be included here? bae458450a landed on main after this PR was opened and promoted it to a first class option, and `test_helper.rb` on main now declares fixtures with it. A rebase would also pick up the `purge_dependent_blob_later` rename in `attachment.rb`.

guides/source/active_storage_overview.md
```diff
@@ -1646,13 +1664,13 @@ by implementing the methods necessary to upload and download files to the cloud.
...
 
-There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads). You can query for unattached records using the [unattached scope](https://github.com/rails/rails/blob/8ef5bd9ced351162b673904a0b77c7034ca2bc20/activestorage/app/models/active_storage/blob.rb#L49). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
+There are cases where a file is uploaded but never attached to a record. This can happen when using [Direct Uploads](#direct-uploads) or when using `dependent: :detach` and later destroying the parent record. You can query for unattached records using the [unattached scope](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html). Below is an example using a [custom rake task](command_line.html#custom-rake-tasks).
 
 ```ruby
 namespace :active_storage do
   desc "Purges unattached Active Storage blobs. Run regularly."
   task purge_unattached: :environment do
-    ActiveStorage::Blob.unattached.where(created_at: ..2.days.ago).find_each(&:purge_later)
+    ActiveStorage::Blob.unattached.where(updated_at: ..2.days.ago).find_each(&:purge_later)
```
joshuay03, line 1673: After running the migration, existing rows will have `updated_at IS NULL`, and this predicate silently drops them since `NULL <= 2.days.ago` isn't truthy. Should the migration backfill `updated_at` from `created_at`, or should this query use `COALESCE` like the upgrade guide does?

activestorage/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+*   Add `dependent: :detach` option for `has_one_attached` and `has_many_attached`.
+
+    When a record is destroyed with `dependent: :detach`, the attachment record
+    is removed but the blob and its file on the service are preserved. This is
+    useful when you want to keep files around for restoration or auditing.
+
+    Also adds `updated_at` column to `active_storage_blobs` table. The blob is
+    touched when a detached attachment is destroyed, so detached blobs can be
+    safely purged based on when they became unattached rather than when they
```
joshuay03, line 9: Is "when they became unattached" quite right here? Would "last modification" be more accurate?

activestorage/test/models/attached/many_test.rb
```diff
@@ -710,6 +710,41 @@ class ActiveStorage::ManyAttachedTest < ActiveSupport::TestCase
...
+        touch_queries << payload[:sql]
+      end
+    }
+
+    ActiveSupport::Notifications.subscribed(counter, "sql.active_record") do
+      @user.documents.detach
+    end
+
+    assert_equal 1, touch_queries.size,
+      "Expected 1 batch UPDATE, got #{touch_queries.size}: #{touch_queries.inspect}"
```
joshuay03, line 745: I think we can use:

```suggestion
    assert_queries_match(/UPDATE.*active_storage_blobs.*updated_at/i, count: 1) { @user.documents.detach }
```

PR #57034, opened 2026-03-22
Skip unsupported --ssl-mode in mysqldump/mysql CLI

### LINKED ISSUE #56876: db:schema:dump does not work when using trilogy, sql schema format and ssl mode
I'm migrating an application from `mysql2` to `trilogy`. It's rails 7.2, but AFAICT the state remains the same in the main branch.

This application dumps the schema using `:sql` format, and the database enforces `caching_sha2_password`, which requires explicit ssl_mode for `trilogy` (it didn't for `mysql2` because it enables it by default, [as I found out that trilogy will start doing the same in an upcoming release](https://github.com/trilogy-libraries/trilogy/issues/199)). 

This explicit database configuration interferes with how the schema dump works though; this is being done by shelling out to `mysqldump`, and a pre-process step turns database configs into cli args. This will include `ssl_mode`, which is turned into  `--ssl-mode`. This cli option exists, but requires an explicit flag on compilation, and mysql/mariadb available distro packages aren't usually compiled with extra flags. This is the case of our base container image, which uses debian bookworm. 

My immediate problem should be fixable once the next version of `trilogy` is published (whenever that happens), but fwiw this will always be a problem if an explicit ssl-mode is set in the configuration even after that. 

### Steps to reproduce

create the rails app in a debian base container (`ruby-3.4.8:bookworm` should be enough, and this is probably reproducible in trixie as well) and install mysql packages from apt-get:

```
apt-get update
  apt-get install -y --no-install-recommends \
    build-essential \
    default-mysql-client \
    default-libmysqlclient-dev
```

you can confirm whether `mysqldump` supports ssl-mode by doing `mysqldump --help` and not see it listed.

in your rails application, use `trilogy`, set:

```ruby
# application.rb
config.active_record.schema_format = :sql
```

in your database.yml:

```
development:
  adapter: trilogy
  # ...
  ssl_mode: required # or preferred_noverify, it doesn't matter which, even disabled would fail, this option just needs to be listed here
```

then run `rails db:schema:dump`.

### Expected behavior

the `structure.sql` file should be correctly created.

### Actual behavior

```
mysqldump: unknown variable 'ssl-mode=required'
bin/rails aborted!
failed to execute: `mysqldump`
Please check the output above for any errors and make sure that `mysqldump` is installed in your PATH and has proper permissions.
```

### System configuration

**Rails version**:  rails 7.1.6 (although this should be the current state in master)

**Ruby version**: ruby 3.4.8 (but the version is irrelevant here)
`

Mordorreal:
## Motivation

Fixes #56876

When `ssl_mode` is set in `database.yml` (common with the Trilogy adapter for `caching_sha2_password`), `db:schema:dump` fails because `mysqldump` on Debian/Ubuntu is not compiled with `--ssl-mode` support:

```
mysqldump: unknown variable 'ssl-mode=required'
```

## Solution

Detect at runtime whether `mysqldump`/`mysql` supports `--ssl-mode` by checking its `--help` output. If supported, the flag is included (backward compatible). If unsupported, it is skipped and a warning is logged.

The detection is memoized per command via a class-level method `MySQLDatabaseTasks.ssl_mode_supported?`.

### Backward compatibility

- **mysqldump with --ssl-mode support** (MySQL 8.0+): flag included as before, zero behavior change
- **mysqldump without --ssl-mode support** (Debian/Ubuntu): flag skipped, warning logged, dump succeeds
- **No ssl_mode in config**: no detection runs, zero overhead

## Benchmark Results

Tested against real MySQL 8.0 (Docker) with both ssl-mode-supporting and simulated Debian mysqldump.

### Detection overhead
```
Cold detection:  ~3ms (shell out to mysqldump --help)
Memoized:        ~0ns (hash lookup, 35,000x faster)
Runs once per task invocation.
```

### Full flow verification
```
Debian mysqldump WITH --ssl-mode:    exit=1 FAIL (the bug)
Debian mysqldump WITHOUT --ssl-mode: exit=0 SUCCESS (our fix)

Detection for Debian mysqldump:  false (correctly skips)
Detection for MySQL 8.0:         true  (correctly includes)
```

### Memory
```
Detection (100 calls): 2,492,200 bytes (600 objects)
Memoized (100 calls):  0 bytes (0 objects)
```

## Other adapters

- **PostgreSQL**: Not affected — uses `PGSSLMODE` environment variable, not CLI flags
- **SQLite**: Not affected — no SSL

## Test Plan

- [x] Test: ssl_mode included when mysqldump supports it (trilogy)
- [x] Test: ssl_mode skipped when mysqldump doesn't support it (trilogy)
- [x] Test: ssl_mode included when mysqldump supports it (mysql2)
- [x] Test: ssl_mode skipped when mysqldump doesn't support it (mysql2)
- [x] RuboCop passes on all changed files
- [x] Full-flow benchmark against real MySQL 8.0
- [x] Verified Debian-like mysqldump fails without fix, succeeds with fix

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Skip `--ssl-mode` flag in `mysqldump`/`mysql` CLI when unsupported.
+
+    When `ssl_mode` is set in `database.yml`, `db:schema:dump` and
+    `db:schema:load` now detect at runtime whether `mysqldump`/`mysql`
+    supports `--ssl-mode`. If unsupported (common on Debian/Ubuntu
+    packages), the flag is skipped and a warning is logged.
+
+    *Denis Savchuk*
+
 *   Deprecate the `schema_order` option in PostgreSQL database configurations.
 
     Use `schema_search_path` instead. The `schema_order` alias will be
--- a/activerecord/lib/active_record/tasks/mysql_database_tasks.rb
+++ b/activerecord/lib/active_record/tasks/mysql_database_tasks.rb
@@ -24,6 +24,16 @@ module ActiveRecord
         connection.charset
       end
 
+      def self.ssl_mode_supported?(cmd = "mysqldump") # :nodoc:
+        @ssl_mode_support ||= {}
+        if @ssl_mode_support[cmd].nil?
+          @ssl_mode_support[cmd] = `#{cmd} --help 2>&1`.include?("ssl-mode")
+        end
+        @ssl_mode_support[cmd]
+      rescue Errno::ENOENT
+        false
+      end
+
       def structure_dump(filename, extra_flags)
         args = prepare_command_options
         args.concat(["--result-file", "#{filename}"])
@@ -73,10 +83,20 @@ module ActiveRecord
             sslcapath: "--ssl-capath",
             sslcipher: "--ssl-cipher",
             sslkey:    "--ssl-key",
-            ssl_mode:  "--ssl-mode"
-          }.filter_map { |opt, arg| "#{arg}=#{configuration_hash[opt]}" if configuration_hash[opt] }
+          }
 
-          args
+          if configuration_hash[:ssl_mode]
+            if self.class.ssl_mode_supported?
+              args[:ssl_mode] = "--ssl-mode"
+            else
+              ActiveRecord::Base.logger&.warn(
+                "mysqldump does not support --ssl-mode, skipping. " \
+                "Ensure your mysqldump version supports this option or remove ssl_mode from database configuration."
+              )
+            end
+          end
+
+          args.filter_map { |opt, arg| "#{arg}=#{configuration_hash[opt]}" if configuration_hash[opt] }
         end
     end
   end
--- a/activerecord/test/cases/adapters/mysql2/mysql2_rake_test.rb
+++ b/activerecord/test/cases/adapters/mysql2/mysql2_rake_test.rb
@@ -370,6 +370,38 @@ module ActiveRecord
         end
     end
 
+    def test_structure_dump_with_ssl_mode_when_supported
+      filename = "awesome-file.sql"
+      ActiveRecord::Tasks::MySQLDatabaseTasks.stub(:ssl_mode_supported?, true) do
+        assert_called_with(
+          Kernel,
+          :system,
+          ["mysqldump", "--ssl-mode=required", "--result-file", filename, "--no-data", "--routines", "--skip-comments", "test-db", {}],
+          returns: true
+        ) do
+          ActiveRecord::Tasks::DatabaseTasks.structure_dump(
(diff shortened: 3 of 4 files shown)


yahonda:
Thanks for picking up #56876. I have two concerns with this approach, though, and I don't think we should merge it as is.

**1. It silently disables TLS**
trilogy connects in plaintext when `ssl_mode` is omitted (still true in 2.13.0). There is a proposal to make TLS the default (trilogy-libraries/trilogy#306, trilogy-libraries/trilogy#307), but it is still under discussion and not merged; the maintainers are leaning toward shipping it in Trilogy 3, and Active Record pins `gem "trilogy", "~> 2.7"`, so it won't reach Rails apps soon. So if database.yml says `ssl_mode: required`, the user is asking for TLS. This PR drops `--ssl-mode` when it detects a MariaDB-flavored `mysqldump` and only logs a warning. MariaDB clients before 11.4 (the issue's Debian bookworm ships 10.11) default to plaintext, so the app talks TLS while `structure_dump` goes over plaintext, guarded by a single `logger.warn` that nobody reads in CI. That contradicts the user's stated intent and is not secure. Clients 11.4+ enable TLS by default anyway, so dropping the flag isn't needed there either.

**2. Runtime detection isn't worth the complexity**
Whether `mysqldump` is MariaDB or MySQL can't be known without running it, and shelling out to `--help` per task adds complexity and cost. The user, on the other hand, already knows they are using the MariaDB client, so an explicit app-side setting is the better fit.

**Where #56876 stands, and what's available today**
- The reporter of #56876 set `ssl_mode` because trilogy needed TLS to authenticate with `caching_sha2_password`. That is no longer the case since trilogy 2.11.0 (trilogy-libraries/trilogy#262), which was merged but unreleased when the issue was filed. If TLS isn't a hard requirement, removing `ssl_mode` resolves the situation in #56876.
- If TLS is required, on main since #58725 an initializer such as `config/initializers/mariadb_cli.rb` keeps TLS for the app and for dump/load while using MariaDB's `mysqldump` / `mysql` (verified against MariaDB 13.0.2). MariaDB's `--ssl` enforces TLS since 5.5.44, so it is the equivalent of MySQL's `--ssl-mode=REQUIRED`.

```ruby
# config/initializers/mariadb_cli.rb
  require "active_record/connection_adapters/trilogy_adapter"
  ActiveRecord::ConnectionAdapters::TrilogyAdapter.cli_arg_map =
    ActiveRecord::ConnectionAdapters::TrilogyAdapter.cli_arg_map.except(:ssl_mode)
end
ActiveRecord::Tasks::DatabaseTasks.structure_dump_flags = { trilogy: ["--ssl"] }
ActiveRecord::Tasks::DatabaseTasks.structure_load_flags = { trilogy: ["--ssl"] }
```

`structure_dump_flags` / `structure_load_flags` alone are not enough, because `--ssl-mode=required` would still be emitted and rejected; the `cli_arg_map` part is required.

PR #57036, opened 2026-03-23
Add previous_attachment_changes for after_commit callbacks

### LINKED ISSUE #56861: ActiveStorage clears `attachment_changes` too early after Rails 7.1.0
## Description

With the introduction of Rails 7.1.0, [a patch](https://github.com/rails/rails/pull/46992) was added to fix the order in which the `after_commit` callbacks are executed. Also, the `prepend` option was turned kind of into a noop. 

The issue we are finding with this is the following use case:

- Picture a `User` model with a `profile` attachment. 
- When the profile is successfully attached and persisted on the DB, we want to execute a certain job.

```ruby
class User 
  has_one_attached :profile
 # has_one_attached injects an after_commit in here: 
 # after_commit(on: %i[ create update ]) { attachment_changes.delete(name.to_s).try(:upload) }

  after_commit :queue_profile_job, if: ->(u) { u.attachment_changes[:profile].present? }
end 
```

The above used to work because the commits would run on inverse order, having `queue_profile_job` running first, and then the activestorage's callback would run last. 
Now, what happens is that they run in ascending order and `attachment_changes` is no longer available unless you declare your model like this:

```ruby
class User 
  after_commit :queue_profile_job, if: ->(u) { u.attachment_changes[:profile].present? }

  has_one_attached :profile
end 
``` 
Honestly, this just reads very weird. Normally one would declare the attachment/association first and then declare the callbacks. It makes more sense to the reader IMO. 

Another possible solution would have been to do this:
```ruby
class User 
  has_one_attached :profile
 
 after_commit :queue_profile_job, prepend: true,  if: ->(u) { u.attachment_changes[:profile].present? }
end 
```
This would have solved the issue as well and keep readability intact, however, prepend was [turned into a noop as well or doesn't work as expected anymore](https://discuss.rubyonrails.org/t/after-commit-with-prepend-and-new-rails-7-1-defaults/84222). So this is also not a viable solution right now.

Regardless of the above, I think the ability to execute this use case shouldn't depend on whether you added your `after_commit` first or after the `has_one/many_attached` method, you should just be able to run something after the attachment finished.

Which brings me to [this solution](https://github.com/rails/rails/pull/51814) that would solve this particular use case for me. 
I think the PR addresses the problem directly and gives user the control it needs to allow it.
What would need to happen to bump that one a little bit to be merged in?

Or maybe it would be good to consider merging [this gem that adds support for it as well](https://github.com/obie/after_attached) if that's possible.

This does not solve the issue reported by [Jonathan on this other thread though](https://github.com/rails/rails/issues/50118). But I think it is still worth to mention how these changes affected other use cases as well.

### Steps to reproduce

<details>
<summary>Here's a test that fails with any version of rails after `7.0.10`, starting from `7.1.0beta1`</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", '7.2.3' # you can also just use latest rails

  gem "sqlite3", "~> 1.7.x"
  gem 'concurrent-ruby', '1.3.4'
end

require "minitest/autorun"

  config.root = __dir__
  config.hosts << "example.org"
  config.eager_load = false

  Rails.logger  = config.logger

  
  # config.active_record.run_after_transaction_callbacks_in_order_defined = false

    local: {
      root: Dir.tmpdir,
      service: "Disk"
    }
  }
end


Rails.application.initialize!

end

  has_one_attached :profile
  # after_commit(on: %i[ create update ]) { attachment_changes.delete(name.to_s).try(:upload) }

  attr_accessor :did_have_attachment_changes

  after_commit prepend: true do
    puts "after_commit callback: attachment changes for profile: #{attachment_changes['profile'].inspect}"
    self.did_have_attachment_changes = attachment_changes['profile'].present?
  end

  after_initialize do
    self.did_have_attachment_changes = nil
  end


end

  def test_upload_and_download
    user = User.create!

    User.find(user.id)
    assert !user.did_have_attachment_changes, "Initial status of attachment changes should be nil"
    puts "Attachment changes: #{user.did_have_attachment_changes}"

    user.profile.attach(
      content_type: "text/plain",
      filename: "dummy.txt",
      io: ::StringIO.new("dummy"),
      )

    assert user.did_have_attachment_changes, "Attachment changes should be present in after_commit callback"

    puts "Attachment changes: #{user.did_have_attachment_changes}"

  end
end
```
</details>

### Expected behavior

`attachment_changes` should be available on `after_commits` callbacks regardless of the order they execute on.

### Actual behavior

`attachment_changes` entry for the attachment is deleted depending on where the has_one_attached is placed on the code.

### System configuration

**Rails version**:  Any rails version starting from `7.1.0.beta1`

**Ruby version**:  Ruby 3.3.10

Mordorreal:
### Motivation / Background

Fixes #56861

Since Rails 7.1 runs `after_commit` callbacks in definition order, user callbacks defined after `has_one_attached`/`has_many_attached` can't see `attachment_changes` — Active Storage's callback deletes the entry before user callbacks run.

```ruby
  has_one_attached :avatar

  # runs AFTER Active Storage's callback — attachment_changes["avatar"] is nil
  after_commit :process_avatar, if: -> { attachment_changes["avatar"].present? }
end
```

### Detail

Preserve attachment changes in `previous_attachment_changes` before clearing them, mirroring Active Record's `previous_changes`:

```ruby
  has_one_attached :avatar

  after_commit :process_avatar, on: %i[create update],
    if: -> { previous_attachment_changes["avatar"].present? }
end
```

`previous_attachment_changes` is cleared on `reload` and `initialize_dup`, matching the lifecycle of `attachment_changes`.

Also replaces `.try(:upload)` with `respond_to?(:upload)` + direct call, removing `.try`'s overhead.

### Additional information

| Path | Before | After |
|---|---|---|
| Create attachment (common) | 3,092k i/s | 4,190k i/s (+1.36×) |
| Delete attachment (edge) | 4,174k i/s | 4,175k i/s (same) |

Memory: +16 bytes/commit (one extra hash entry per transaction).

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/attached/model.rb
+++ b/activestorage/lib/active_storage/attached/model.rb
@@ -156,7 +156,13 @@ module ActiveStorage
 
         after_save { attachment_changes[name.to_s]&.save }
 
-        after_commit(on: %i[ create update ]) { attachment_changes.delete(name.to_s).try(:upload) }
+        after_commit(on: %i[ create update ]) do
+          change = attachment_changes.delete(name.to_s)
+          if change
+            (@previous_attachment_changes ||= {})[name.to_s] = change
+            change.upload if change.respond_to?(:upload)
+          end
+        end
 
         reflection = ActiveRecord::Reflection.create(
           :has_one_attached,
@@ -270,7 +276,13 @@ module ActiveStorage
 
         after_save { attachment_changes[name.to_s]&.save }
 
-        after_commit(on: %i[ create update ]) { attachment_changes.delete(name.to_s).try(:upload) }
+        after_commit(on: %i[ create update ]) do
+          change = attachment_changes.delete(name.to_s)
+          if change
+            (@previous_attachment_changes ||= {})[name.to_s] = change
+            change.upload if change.respond_to?(:upload)
+          end
+        end
 
         reflection = ActiveRecord::Reflection.create(
           :has_many_attached,
@@ -307,6 +319,14 @@ module ActiveStorage
       @attachment_changes ||= {}
     end
 
+    # Returns a hash of attachment changes that were committed in the last
+    # transaction. Similar to ActiveModel's +previous_changes+, this allows
+    # +after_commit+ callbacks to check which attachments were modified,
+    # regardless of callback definition order.
+    def previous_attachment_changes
+      @previous_attachment_changes ||= {}
+    end
+
     def changed_for_autosave? # :nodoc:
       super || attachment_changes.any?
     end
@@ -315,10 +335,14 @@ module ActiveStorage
       super
       @active_storage_attached = nil
       @attachment_changes = nil
+      @previous_attachment_changes = nil
     end
 
     def reload(*) # :nodoc:
-      super.tap { @attachment_changes = nil }
+      super.tap do
+        @attachment_changes = nil
+        @previous_attachment_changes = nil
+      end
     end
   end
 end
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,20 @@
+*   Add `previous_attachment_changes` for `after_commit` callbacks.
+
+    Active Storage now preserves attachment changes in
+    `previous_attachment_changes` before clearing them during the
+    upload `after_commit` callback. User-defined `after_commit`
+    callbacks can read `previous_attachment_changes` to determine
+    which attachments were modified, regardless of callback definition
+    order.
+
+    Previously, `attachment_changes` was destructively cleared during
+    upload, making it unavailable to callbacks defined after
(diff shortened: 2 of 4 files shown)


activestorage/lib/active_storage/attached/model.rb
claudiob, line 318: @Mordorreal I am very in favor of this behavior. Something I noticed, though, is that `attachment_changes` itself is `# :nodoc:`, meaning that technically this is not public API.

I think that if we decide to add `previous_attachment_changes` then we should also make `attachment_changes` a public API (by removing :nodoc: and adding documentation).

Just my two cents. Let's see what the Rails core team says.

PR #57038, opened 2026-03-23
Strip whitespace from URL strings in redirect_to

Mordorreal:
## Summary

URLs with leading or trailing whitespace (e.g. `" https://example.com"`) passed to `redirect_to` were incorrectly classified as path-relative URLs, triggering `PathRelativeRedirectError` or warnings depending on `action_on_path_relative_redirect` config.

This happened because the leading space prevented the scheme regex (`/\A([a-z]...)/`) from matching, causing the URL to fall through to the `String` branch where it was treated as a relative path.

Fixes #57035

## Solution

Handle whitespace inside the existing `case/when` branches of `_compute_redirect_to_location` using `getbyte` guards:

- **Scheme-regex branch**: checks last byte for trailing whitespace (`"https://example.com/ "`)
- **String branch**: checks first byte for leading whitespace (`" https://example.com"`), strips, and re-checks the scheme regex

### Why `getbyte` instead of `String#strip`?

`String#strip` always allocates a new string even when there's nothing to strip. At high RPS this adds unnecessary GC pressure. `getbyte` is a zero-allocation integer comparison that lets us skip `.strip` entirely for the common case (clean URLs with no whitespace).

## Benchmark

```
=== Clean absolute URL (common case) ===
            original:  2.470M i/s  (405ns)
                 new:  2.266M i/s  (441ns) - 1.09x slower

=== Clean path ===
            original:  3.256M i/s  (307ns)
                 new:  3.031M i/s  (330ns) - 1.07x slower

=== Memory: 10k clean URLs ===
Original:  2480000 bytes, 20000 objects
New:       2480000 bytes, 20000 objects  (identical)
```

Overhead is ~20-36ns per call with zero extra memory allocations for clean URLs. At 20k RPS: ~0.72ms/sec total CPU overhead.

## Test Plan

- [x] 5 new tests covering leading whitespace on absolute URLs, trailing whitespace, leading whitespace on paths, and `:raise` mode for both URL types
- [x] Updated `test_filter_redirect_bad_uri` (old test relied on leading-space behavior)
- [x] Full actionpack suite: 4165 runs, 0 failures, 0 errors
- [x] RuboCop: 0 offenses

### DIFF
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -211,8 +211,30 @@ module ActionController
       # https://tools.ietf.org/html/rfc3986#section-3.1 The protocol relative scheme
       # starts with a double slash "//".
       when /\A([a-z][a-z\d\-+.]*:|\/\/).*/i
-        options.to_str
+        if options.is_a?(String)
+          # Strip trailing whitespace from URLs like "https://example.com/ "
+          # to avoid invalid Location headers. We use getbyte as a guard
+          # because String#strip always allocates a new string even when
+          # there's nothing to strip. Byte <= 32 covers all ASCII whitespace
+          # (space, tab, newline, carriage return, form feed).
+          last_byte = options.getbyte(-1)
+          (last_byte && last_byte <= 32) ? options.strip : options
+        else
+          options.to_str
+        end
       when String
+        # Strip leading whitespace from URLs like " https://example.com"
+        # that fail to match the scheme regex due to the leading space,
+        # then re-check if the stripped result is actually an absolute URL.
+        # Same getbyte guard as above to avoid allocating on clean strings.
+        first_byte = options.getbyte(0)
+        if first_byte && first_byte <= 32
+          options = options.strip
+          if options.match?(/\A([a-z][a-z\d\-+.]*:|\/\/).*/i)
+            return options.delete("\0\r\n")
+          end
+        end
+
         if !options.start_with?("/", "?") && !options.empty?
           _handle_path_relative_redirect(options)
         end
--- a/actionpack/test/controller/log_subscriber_test.rb
+++ b/actionpack/test/controller/log_subscriber_test.rb
@@ -38,7 +38,7 @@ class ACLogSubscriberTest < ActionController::TestCase
       end
 
       def filterable_redirector_bad_uri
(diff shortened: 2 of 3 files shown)


p8:
If there's whitespace surrounding a URL Rails shouldn't clean it up.
The source of the whitespace should be fixed instead.

Mordorreal:
> If there's whitespace surrounding a URL Rails shouldn't clean it up. The source of the whitespace should be fixed instead.

Hi @p8, thank you for your comment.
I’m not sure I fully understand what changes you would like to see in this PR. Could you please clarify what exactly should be modified or what approach you would suggest instead?
I’d really appreciate a bit more detail so I can update the PR accordingly.

actionpack/lib/action_controller/metal/redirecting.rb
```diff
@@ -211,8 +211,30 @@ def _compute_redirect_to_location(request, options) # :nodoc:
...
       # starts with a double slash "//".
       when /\A([a-z][a-z\d\-+.]*:|\/\/).*/i
-        options.to_str
+        if options.is_a?(String)
+          # Strip trailing whitespace from URLs like "https://example.com/ "
+          # to avoid invalid Location headers. We use getbyte as a guard
+          # because String#strip always allocates a new string even when
+          # there's nothing to strip. Byte <= 32 covers all ASCII whitespace
+          # (space, tab, newline, carriage return, form feed).
+          last_byte = options.getbyte(-1)
```
byroot, line 220: This sort of micro optimization is not necessary in a codepath that is called at most once per request.

Just do it the simple and readable way, even if it means a string allocation.

byroot:
> I’m not sure I fully understand what changes you would like to see in this PR.

@p8 is saying no change is needed, and that `redirect_to` shouldn't try to handle that case.

actionpack/lib/action_controller/metal/redirecting.rb
Mordorreal: I agree that optimizing for the sake of optimization isn’t necessary. However, please correct me if I’m mistaken — my understanding was that this part of the Rails logic could be relatively heavily used. I was under the impression that, in general, Rails code paths tend to operate under significant load and might benefit from being mindful of CPU and memory usage by default.

Mordorreal:
@byroot in this case I'm closing this PR

PR #57041, opened 2026-03-23
Fix preloader to respect case-insensitive column collations

Mordorreal:
### Motivation / Background

Fixes #56780

When preloading associations where the key column has a case-insensitive collation (SQLite `NOCASE`, MySQL `_ci`, PostgreSQL `citext`), the database returns matching records but the preloader fails to associate them back to their owners. `owners_by_key` is a plain Ruby Hash with case-sensitive key lookup — `"Home"` and `"home"` are different keys in Ruby even though they match in the database.

### Detail

Add `Column#case_sensitive?` with adapter-specific overrides (SQLite: `NOCASE` collation; PostgreSQL: `citext` type and non-deterministic collations detected via the existing `column_definitions` query — zero extra DB round-trips). In the preloader, downcase string keys only when the flag is set — zero overhead on the common case-sensitive path.

### Additional information

1,000 owners, 1,000 records, Ruby 4.0.1:

| Path | Before | After |
|---|---|---|
| CS single key (common case) | 2,826 i/s | 2,985 i/s (+1.06×) |
| CS composite key | 1,517 i/s | 1,442 i/s (−1.05×) |
| CI single key (bug fix, was nil) | — | 2,115 i/s |

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix preloader to respect case-insensitive column collations.
+
+    Preloading associations keyed on case-insensitive columns (SQLite
+    NOCASE, MySQL `_ci`, PostgreSQL citext/non-deterministic ICU) now
+    correctly matches records. Previously, Ruby's case-sensitive Hash
+    lookup silently returned nil.
+
+    *Denis Savchuk*
+
 *   Deprecate the `schema_order` option in PostgreSQL database configurations.
 
     Use `schema_search_path` instead. The `schema_order` alias will be
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -265,7 +265,14 @@ module ActiveRecord
 
           def derive_key(owner, key)
             if key.is_a?(Array)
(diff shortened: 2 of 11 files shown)


tylerwood-io:
The approach of adding \`case_sensitive?\` to the column abstraction and then checking it in the preloader's key derivation is clean. Good that MySQL already had \`case_sensitive?\` — this PR fills in the gap for PostgreSQL (citext + non-deterministic ICU collations via \`collisdeterministic\`) and SQLite3 (NOCASE).

A couple of observations:

1. \`convert_key\` uses Ruby's \`String#downcase\` to normalize case-insensitive keys. This works for ASCII and most Latin scripts, but could diverge from the database's collation rules for some Unicode edge cases (e.g., Turkish \`I\` → \`ı\` vs Ruby's \`I\` → \`i\`, or German \`ß\` equivalences). Probably fine for the vast majority of real-world usage, but worth noting as a known limitation.

2. The composite key handling in \`ci_key_indices\` / \`derive_key\` only downcases the string components that have case-insensitive collations, leaving integer components untouched. Good attention to detail there.

3. This changes observable behavior (preloading now works correctly with case-insensitive keys), so it likely warrants a CHANGELOG entry.

Mordorreal:
Added CHANGELOG entry. Thanks for the review!

PR #57043, opened 2026-03-23
Add test coverage for ActiveModel::Error#strict_match?

hammadxcm:
### Motivation / Background

The `strict_match?` method on `ActiveModel::Error` is a public API for checking if an error matches a given attribute, type, and options exactly, but had no dedicated test coverage.

The existing `match?` method has thorough tests, but `strict_match?` — which differs by requiring all options to match exactly (excluding callback and message options) — was untested.

### Detail

This adds tests for `ActiveModel::Error#strict_match?` verifying:
- Returns true when attribute, type, and options all match exactly
- Returns false when attribute does not match
- Returns false when type does not match
- Returns false when option values differ
- Returns false when options are missing from the check
- Correctly ignores callback options (`:if`, `:unless`, `:on`, `:allow_nil`, `:allow_blank`, `:strict`) and message options

### Additional information

N/A

### Checklist

- [x] Tests added for the previously untested public method.
- [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. _(N/A — test-only change)_

### DIFF
--- a/activemodel/test/cases/error_test.rb
+++ b/activemodel/test/cases/error_test.rb
@@ -85,6 +85,50 @@ class ErrorTest < ActiveModel::TestCase
     assert subject.match?(:mineral, :not_enough, count: 2)
   end
 
+  # strict_match?
+
+  test "strict_match? returns true when attribute, type, and options all match" do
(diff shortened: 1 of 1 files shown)


hammadxcm:
@byroot  Please review

PR #57050, opened 2026-03-23
Combine per-validator and top-level :if/:unless/:on in validates

### LINKED ISSUE #55761: `validates` per-validator options replace top-level options
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.string :title
    t.boolean :local
    t.boolean :global
  end
end

  validates :title, presence: { if: :local? }, if: :global?
end

  def test_association_stuff
    assert_not Post.new(local: true, global: true).valid?

    assert Post.new(local: false, global: true).valid?
    assert Post.new(local: false, global: false).valid?

    # Presence's :if (local) replaces validates's :if (global), so whether global? is truthy or falsey is irrelevant
    assert Post.new(local: true, global: false).valid?
  end
end
```

### Expected behavior

I expected the validation to only be enforced when both `local` and `global` were truthy.

### Actual behavior

The validation is enforced when `local` is truthy, regardless of the value of `global`.

### System configuration

**Rails version**: main

**Ruby version**: 3.4.5

### More Info

That's an over-simplified reproduction.

More specifically, I have a number of validations I want to run only when a certain condition passes, then within those validations, a subset have an additional criteria.

I tried something like:
```ruby
with_options(if: :global?) do |o|
  o.validates :title, presence: true, if: :local?
  o.validates :author, presence: true
  # and so on
end
```

hoping it'd become something like (not literally expecting Rails to define a method combining the conditions):
```ruby
validates :title, presence: true, if: :local_and_global?
validates :author, presence: true, if: :global?
```

But `with_options` documents that the inner options take precedence over the options given to `with_options`, so that wouldn't work the way I wanted, as that becomes:
```ruby
validates :title, presence: true, if: :local?
validates :author, presence: true, if: :global?
```

Then I considered moving the condition inside the specific validator (`presence`), rather than at the `validates` level, along the lines of:
```ruby
with_options(if: :global?) do |o|
  o.validates :title, presence: { if: :local? }
  o.validates :author, presence: true
  # and so on
end
```
hoping it would effectively combine the `:if`s, checking both, yielding something equivalent to:
```ruby
validates :title, presence: { if: :local? }, if: :global?
validates :author, presence: true, if: :global?
```

It seems the options given to individual validators take precedence over the options given at the `validates` level, so that really just evaluates to:
```ruby
validates_presence_of :title, if: :local?
```

Best I can tell, having a validator-specific `:if` and `validates`-level `:if` is not explicitly [documented](https://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates), nor is it explicitly [tested](https://github.com/rails/rails/blob/a821b31a40a03404412a1aafeb9c3190f138ac0b/activemodel/test/cases/validations/validates_test.rb), leading me to believe this combination is simply undefined as to what the expected behavior should be.

I could see three directions to take this in:
1. document that when providing the same option at both levels, the inner one wins
2. check both conditions to enforce validation (though this could be breaking for applications reliant on the current behavior). Perhaps this could be accomplished by adding array support to `:if` where all conditions need to pass to enforce the validation, then Rails could combine both levels of `:if`s into an array automatically (and the end-user can use an array where needed, separately from this issue).
3. disallow/error when the same option is provided at both levels, as the intended effects could be ambiguous (replace or combine?)

There's likely a broader question here in that whatever is decided for `:if` would likely apply to `:unless`, but would it also apply to the other options (ie. `:on`, `:except_on`, `:allow_nil`, `:allow_blank`, or `:strict`)?

Mordorreal:
### Motivation / Background

Fixes #55761.

When using `validates` with both top-level and per-validator `:if`/`:unless`/`:on` options, the per-validator options silently replace the top-level ones instead of combining them:

```ruby
validates :title, presence: { if: :local? }, if: :global?
# Expected: validation runs when BOTH local? AND global? are true
# Actual:   validation runs when local? is true (global? is ignored)
```

This happens because `validates` uses `Hash#merge` which replaces duplicate keys.

### Detail

~~This PR contains two commits:~~

~~**1. Optimize callback condition evaluation (`activesupport`)**~~

~~Compiles callback conditions into a single callable instead of iterating an array with `.all?` on every invocation:~~
~~- 0 conditions → `nil` (no-op check)~~
~~- 1 condition → the lambda itself (direct call, no array iteration)~~
~~- 2 conditions → `first.call && second.call` (direct `&&`, no `all?`)~~
~~- 3+ conditions → `conditions.all? { ... }` (existing behavior)~~

~~This speeds up **all** callbacks in Rails (validations, lifecycle hooks, controller filters).~~

**2. Combine per-validator and top-level options (`activemodel`)**

Adds `_merge_validates_options` that combines `:if`, `:unless`, and `:on` into arrays when specified at both levels. Scalar options (`:message`, `:allow_nil`, `:allow_blank`, `:strict`) keep the existing override behavior.

```ruby
validates :title, presence: { if: :local? }, if: :global?
# Now equivalent to:
validates_presence_of :title, if: [:global?, :local?]
```

### Benchmark Results

Validation execution (Ruby 4.0.1):

| Scenario | Before (main) | After | Delta |
|---|---|---|---|
| top-level `:if` only | 424,777 i/s | **434,393 i/s** | **+2.3% faster** |
| per-validator `:if` only | 422,107 i/s | **427,205 i/s** | **+1.2% faster** |
| combined `:if` (both levels) | 424,330 i/s | 410,180 i/s | -3.3% |
| multi-validator combined | 206,672 i/s | 203,175 i/s | -1.7% |
| class definition | 5,368 i/s | 5,430 i/s | same-ish |

The single-condition paths (the common case) are **faster** than before thanks to the callbacks optimization. The combined case is ~3% slower, which is the irreducible cost of evaluating 2 conditions instead of 1.

### Checklist

- [x] Tests added for `:if`, `:unless`, `:on` combining and scalar override preservation
- [x] Full ActiveSupport test suite passes (5,849 runs, 0 failures)
- [x] Full ActiveModel test suite passes (1,146 runs, 0 failures)
- [x] RuboCop clean
- [x] RDoc updated on `validates` method
- [x] CHANGELOG entry added
- [x] Benchmarked before and after

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Combine `:if`, `:unless`, and `:on` options when specified at both the
+    `validates` level and the per-validator level, instead of the per-validator
+    options silently replacing the top-level ones.
+
+    Before, `validates :title, presence: { if: :local? }, if: :global?` would
+    only check `local?`, ignoring `global?` entirely. Now both conditions must
+    pass for the validation to run.
+
+    Fixes #55761.
+
+    *Denis Savchuk*
+
 *   Add `has_json` and `has_delegated_json` to provide schema-enforced access to JSON attributes.
 
     ```ruby
--- a/activemodel/lib/active_model/validations/validates.rb
+++ b/activemodel/lib/active_model/validations/validates.rb
@@ -108,6 +108,17 @@ module ActiveModel
       # and +:message+ can be given to one specific validator, as a hash:
       #
       #   validates :password, presence: { if: :password_required?, message: 'is forgotten.' }, confirmation: true
+      #
+      # When +:if+, +:unless+, or +:on+ appear at both the +validates+ level and
+      # inside a specific validator's options, they are combined rather than the
+      # inner option replacing the outer one. All +:if+ conditions must pass,
+      # any +:unless+ condition will skip validation, and +:on+ contexts are merged:
+      #
+      #   validates :password, presence: { if: :local_check? }, if: :global_check?
+      #   # Equivalent to: validates_presence_of :password, if: [:global_check?, :local_check?]
+      #
+      # Other per-validator options (+:allow_blank+, +:allow_nil+, +:strict+,
+      # +:message+) override the same option given at the +validates+ level.
       def validates(*attributes)
         defaults = attributes.extract_options!.dup
         validations = defaults.slice!(*_validates_default_keys)
@@ -128,7 +139,7 @@ module ActiveModel
 
           next unless options
 
-          validates_with(validator, defaults.merge(_parse_validates_options(options)))
+          validates_with(validator, _merge_validates_options(defaults, _parse_validates_options(options)))
         end
       end
 
@@ -161,6 +172,16 @@ module ActiveModel
       # additional default keys. This can be done by overwriting this method.
       def _validates_default_keys
         [:if, :unless, :on, :allow_blank, :allow_nil, :strict, :except_on]
+      end
+
+      def _merge_validates_options(defaults, validator_options)
+        defaults.merge(validator_options) do |key, default_val, validator_val|
+          if key == :if || key == :unless || key == :on
+            Array(default_val) + Array(validator_val)
+          else
+            validator_val
+          end
+        end
       end
 
       def _parse_validates_options(options)
--- a/activemodel/test/cases/validations/validates_test.rb
+++ b/activemodel/test/cases/validations/validates_test.rb
@@ -166,4 +166,55 @@ class ValidatesTest < ActiveModel::TestCase
     assert_not_predicate topic, :valid?
     assert_equal ["Y U NO CONFIRM"], topic.errors[:title_confirmation]
   end
+
+  def test_validates_combines_if_from_both_levels
+    Person.validates :title, presence: { if: :condition_is_true }, if: :condition_is_true
+    person = Person.new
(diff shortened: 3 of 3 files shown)


byroot:
That optimization commit is entirely unnecessary.

Mordorreal:
@byroot  Thanks for taking a look!

Could you please share a bit more detail on why you think this optimization isn’t necessary? I might be missing something here.

My thinking was that this part of the system is likely to be executed quite frequently, and even a small improvement (e.g., ~2%) could have a noticeable impact at scale—especially in high-load environments like Shopify.

That said, I’m very open to reconsidering if there are trade-offs or context I haven’t accounted for.

byroot:
This isn't a hotpath and 2% is nothing.

In general, if you are fixing a bug, keep your PR as minimal and focused as possible, and that include the PR description. 

And about the PR description, I don't care whether contributors are using AI or not, but I can't stand how verbose the PR they produce is, it's a huge waste of my time to sip though AI verbal diarrhea. Short and to the point is the way to go.

cgunther:
Thanks for fixing this @Mordorreal ! (original issue author here)

I didn't even realize arrays were already supported for `:if`/`:unless` on validations, so opened #57082 to document that.

PR #57051, opened 2026-03-23
Fix association build to set inverse before assigning attributes

hammadxcm:
### Motivation / Background

Fixes #57042

When using `parent.children.build(attributes)` with a model that has an ActiveStorage attachment with a dynamic service (e.g., `has_one_attached :file, service: ->(record) { record.parent.region_service }`), the service proc fails because the parent association is `nil` at the time the attachment attribute setter runs.

This happens because `ActiveRecord::Core#initialize` calls `super` (which triggers `assign_attributes`) **before** yielding the block that sets the inverse instance via `initialize_attributes`.

### Detail

The fix has two parts:

1. **`ActiveRecord::Core#initialize`** — yield the block **before** calling `super` (which triggers `assign_attributes`). This ensures that `initialize_attributes` (which sets the inverse instance and foreign key) runs before user-provided attributes are assigned.

2. **`Association#build_record`** — move the user-provided block (`yield`) outside of `build_association` so it runs **after** both initialization and attribute assignment. This preserves the existing behavior where blocks passed to `build` can override attribute values.

### Additional information

Verified with 2,289 tests across all association types, STI, nested attributes, autosave, callbacks, persistence, scoping, eager loading, and ActiveStorage — zero new failures introduced.

### Checklist

- [x] Commit message has a detailed description of what changed and why, with issue reference.
- [x] Tests added to verify the inverse is set before attributes are assigned during `build`.
- [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. _(N/A — bug fix)_

### DIFF
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -381,10 +381,11 @@ module ActiveRecord
         end
 
         def build_record(attributes)
-          reflection.build_association(attributes) do |record|
+          record = reflection.build_association(attributes) do |record|
             initialize_attributes(record, attributes)
-            yield(record) if block_given?
           end
+          yield(record) if block_given?
+          record
         end
 
         # Returns true if statement cache should be skipped on the association reader.
--- a/activerecord/lib/active_record/core.rb
+++ b/activerecord/lib/active_record/core.rb
@@ -486,9 +486,10 @@ module ActiveRecord
       init_internals
       initialize_internals_callback
 
+      yield self if block_given?
+
       super
 
-      yield self if block_given?
       _run_initialize_callbacks
     end
 
--- a/activerecord/test/cases/associations/has_many_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_associations_test.rb
@@ -311,6 +311,23 @@ class HasManyAssociationsTest < ActiveRecord::TestCase
     assert_equal car, bulb.car
   end
 
+  def test_build_from_association_sets_inverse_before_assigning_attributes
+    the_car = Car.new(name: "honda")
+    inverse_during_init = nil
+
+    Bulb.define_method(:name=) do |value|
+      inverse_during_init = self.car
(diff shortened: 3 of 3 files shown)


hammadxcm:
@byroot Let me know if you require anything from my side.

tylerwood-io:
Moving \`yield\` before \`super\` in \`Core#initialize\` changes the initialization order for all \`Model.new(attrs) { |r| ... }\` calls, not just association builds. Any code passing a block to \`new\` would see that block run before attribute assignment, which is a behavioral change beyond the scope of the fix.

For example, this pattern would break:

\`\`\`ruby
Post.new(title: "Hello") do |post|
  puts post.title  # => nil (attributes not yet assigned)
end
\`\`\`

#57055 takes a more targeted approach — modifying only \`build_record\` to set the inverse before assigning attributes, without touching \`Core#initialize\`. That keeps the fix scoped to association builds where the problem actually lives.

hammadxcm:
@tylerwood-io You're absolutely right — moving `yield` before `super` in `Core#initialize` changes the initialization order for all `Model.new(attrs) { |r| ... }` calls, which is a behavioral change beyond the scope of this fix.

I've updated the approach to be more targeted:

1. **Reverted `Core#initialize`** — `yield` stays after `super`, preserving the public API contract.
2. **Modified only `build_record`** — passes only STI type attributes to `klass.new` for proper subclass dispatch, then in the block:
   - `set_inverse_instance(record)` — inverse available immediately
   - `record.assign_attributes(attributes)` — attribute setters can now access the parent
   - `initialize_attributes(record, attributes)` — scope/FK protection runs last
3. **Guarded nil attributes in `through_association.rb`** — handles `parent.children.build` (no args) via through associations.

All association test suites pass (has_many, has_one, has_many_through, inverse, STI, nested_attributes) with zero failures.

Thanks for flagging this and for the pointer to #57055 — that more targeted approach is the right call.

PR #57057, opened 2026-03-24
Report Postgres FATAL exceptions more clearly

matthewd:
`FATAL` means the server is choosing to terminate the connection, so it's useful for us to clearly distinguish that type of error and correctly translate it to ConnectionFailed instead of general StatementInvalid (which would leave us with a broken connection, to only be noticed later).

To more reliably get hold of that more "real" exception: 
* use `send_query_params` and friends, followed by manually pumping `raw_connection.get_result` ourselves, instead of relying on `exec_params`'s built-in `get_last_result`
* listen for fatal errors reported via the notice receiver (less common, but depending on its internal state, libpq can deliver it here instead of returning it as a direct query result)

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -164,6 +164,10 @@ module ActiveRecord
           end
 
           def perform_query(raw_connection, intent)
+            if fatal = consume_notice_receiver_fatal_error
+              raise fatal
+            end
+
             raw_connection.discard_results
 
             if intent.prepare
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -164,11 +164,15 @@ module ActiveRecord
           end
 
           def perform_query(raw_connection, intent)
-            result = if intent.prepare
+            raw_connection.discard_results
+
+            if intent.prepare
               begin
                 stmt_key = prepare_statement(intent.processed_sql, intent.binds, raw_connection)
                 intent.notification_payload[:statement_name] = stmt_key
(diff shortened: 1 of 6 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
```diff
@@ -164,11 +164,15 @@ def cancel_any_running_query
           end
 
           def perform_query(raw_connection, intent)
-            result = if intent.prepare
+            raw_connection.discard_results
```
matthewd, line 171: Arguably if we're confident in our state management, we shouldn't need to do this (at least not unless `@raw_connection_dirty`)... but it is what `exec_params` & friends do 🤷🏻‍♂️

eileencodes (approved):

rafaelfranca (approved):

larskanis:
@matthewd Can you please give me a note, if you're unsatisfied with pg's behavior, like the above? I think it makes sense to fix it upstream in `get_last_result`, but I only look from time to time into rails+postgres updates - I cannot follow all of them. A simple ping is enough.

matthewd:
Sorry @larskanis -- yeah, I definitely should've reported this to you upstream. I greatly appreciate your long-term work on `pg`, and side-stepping a behaviour instead of discussing it at the source is not a good way to show that. :pensive::heart:

I wasn't thinking that way at the time, because switching to a manual `get_result` loop moves this code closer to what I need to do for (still WIP) #57089, so it felt like a _positive_ coincidence.

I agree `get_last_result` can likely be fixed to maintain the richer exception. (I doubt it makes sense to change it to report a SYNC without requiring a subsequent nil, though, so probably still not suited for pipeline reads.)

larskanis:
Thank you @matthewd  for your kind words and explanation! For pipelining it's best to use `get_result`. The pipelining support is something I dreamed of, but the activerecord code is rather complex, so I wasn't able to get something working. A pipelined model `save!` would be my use case (as an alternative to explicit bulk inserts).

PR #57060, opened 2026-03-25
Offload ActiveStorage::Blob#metadata sync to background

### LINKED ISSUE #54919: Data race when updating active storage blob metadata
### Steps to reproduce

Because the problem is a data race, the error won't always happen. However, using GCS for local and direct uploading files often reproduces the problem.

### Expected behavior

It won't raise an error.

### Actual behavior

Sometimes, the "409 conflict" error is returned from GCS, and either http server or background worker raises the following error.

```
Google::Cloud::AlreadyExistsError: conflict: The metadata for object "whatever" was edited during the operation. Please try again. (Google::Cloud::AlreadyExistsError)
```

When a blob is attached, http server calls `ActiveStorage::Blob.update_service_metadata`. At the same time, background worker performs `ActiveStorage::AnalyzeJob` and uploads metadata. This data race ends up in the above error.

I think the first step is to move `ActiveStorage::Blob.update_service_metadata` to a background job. I don't want to return the 500 error to users just because metadata was not updated. Then we can add `retry_on Google::Cloud::AlreadyExistsError`. The data race still exists, but this should mostly work. If we want to solve the problem, one way is to use optimistic locking for blobs. But that may be too much work just for metadata. If the direction above looks OK, I'm going to open a PR.

### System configuration

**Rails version**: Rails 8.0.2

### See also

- https://github.com/rails/rails/issues/48909

shouichi:
### Motivation / Background

A race condition can occur when a server and a worker both try to update blob metadata at the same time, leading to errors on services like GCS.

### Detail

This Pull Request offloads metadata updates to a background worker to avoid the data race.

### Additional information

Close https://github.com/rails/rails/issues/54919.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,24 @@
+*   Offload ActiveStorage::Blob#metadata sync to background
+
+    Problem:
+
+    A data race can occur when both a http server and a background worker (
+    `ActiveStorage::AnalyzeJob`) attempt to update a blob's metadata
+    simultaneously. This leads to a 409 conflict on GCS.
+
+    Solution:
+
+    Offload the metadata upload to a background job so the 409 no longer breaks
+    the request with a 500. Applications can add `retry_on` to
+    `ActiveStorage::SyncMetadataJob` to retry the conflict for their service.
+
+    `ActiveStorage::Service#update_metadata` now instruments the work and
+    delegates it to a new `update_metadata_for` method. Custom services that
+    overrode `update_metadata` should override `update_metadata_for` instead, so
+    the work runs inside the instrumentation.
+
+    *Shouichi Kamiya*
+
 *   Update Active Storage for ImageProcessing 2.0.
 
     ImageProcessing 2.0 now requires adding `ruby-vips` or `mini_magick` gems explicitly to the Gemfile.
--- a/activestorage/lib/active_storage/service.rb
+++ b/activestorage/lib/active_storage/service.rb
@@ -72,9 +72,10 @@ module ActiveStorage
     end
 
     # Update metadata for the file identified by +key+ in the service.
-    # Override in subclasses only if the service needs to store specific
-    # metadata that has to be updated upon identification.
     def update_metadata(key, **metadata)
+      instrument :update_metadata, key: key, **metadata do
+        update_metadata_for key, **metadata
+      end
     end
 
     # Return the content of the file at the +key+.
@@ -187,6 +188,11 @@ module ActiveStorage
 
       def custom_metadata_headers(metadata)
         raise NotImplementedError
+      end
+
+      # Override in subclasses only if the service needs to store specific
+      # metadata that has to be updated upon identification.
+      def update_metadata_for(key, **metadata)
       end
 
       def instrument(operation, payload = {}, &block)
--- a/activestorage/test/jobs/sync_metadata_job_test.rb
+++ b/activestorage/test/jobs/sync_metadata_job_test.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "database/setup"
+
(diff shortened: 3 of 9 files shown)


activestorage/test/jobs/sync_metadata_job_test.rb
```diff
@@ -0,0 +1,27 @@
...
+require "database/setup"
+require "active_support/testing/method_call_assertions"
+
+class ActiveStorage::SyncMetadataJobTest < ActiveJob::TestCase
+  include ActiveSupport::Testing::MethodCallAssertions
+
+  setup { @blob = create_blob }
+
+  test "syncs metadata" do
+    assert_called_with(@blob, :sync_metadata, []) do
```
rafaelfranca: Can we avoid mocks and stubs in those tests? I don't think the tests should be coupled to the implementation. Assert the metadata was synced instead.

rafaelfranca (changes requested):

activestorage/test/jobs/sync_metadata_job_test.rb
shouichi: Thank you for the feedback. I changed to use instrumentation.

activestorage/lib/active_storage/service.rb
```diff
@@ -72,9 +72,10 @@ def upload(key, io, checksum: nil, **options)
     end
 
     # Update metadata for the file identified by +key+ in the service.
-    # Override in subclasses only if the service needs to store specific
```
rafaelfranca, line 75: This is now changing the interface of services? Why it is necessary?

shouichi: It's additive — overriding update_metadata still works as before. The hook moves instrumentation into the base class so every service `emits service_update_metadata.active_storage` (previously GCS only). This is not absolutely necessary, happy to revert if you'd rather keep it untouched.

rafaelfranca (approved):

activestorage/CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+
+    A data race can occur when both a http server and a background worker (
+    `ActiveStorage::AnalyzeJob`) attempt to update a blob's metadata
+    simultaneously. This leads to a 409 conflict on GCS.
+
+    Solution:
+
+    Offload the metadata upload to a background job so the 409 no longer breaks
+    the request with a 500. Applications can add `retry_on` to
+    `ActiveStorage::SyncMetadataJob` to retry the conflict for their service.
```
chaadow, line 13: > Applications can add `retry_on` to
    `ActiveStorage::SyncMetadataJob` to retry the conflict for their service.
    
Hi, I would like to know, how applications can add a `retry_on` on a job that is owned by ActiveStorage. maybe through an initializer with a module that you extend or something?? @shouichi @rafaelfranca  ?

Maybe there is another simpler way that i'm missing? Thanks!

rafaelfranca: `ActiveStorage::SyncMetadataJob.retry_on SomeException`?

chaadow: yeah i do have something like this in my app :
```ruby
ActiveSupport.on_load(:active_storage_blob) do
  next unless defined?(Vips::Error)

  ActiveStorage::CreateVariantsJob.discard_on(Vips::Error)
end
```

@rafaelfranca  do you think it's worth it to add the active job classes to the "Load hook" system just like some other ActiveStorage classes do?
this way we can have : 
```ruby
 ActiveSupport.on_load(:active_storage_sync_metadata_job) do
    ActiveStorage::SyncMetadataJob.retry_on SomeException
  end
```

rafaelfranca: Yes. That would be good

chaadow: @rafaelfranca Thanks, I opened https://github.com/rails/rails/pull/57661

PR #57062, opened 2026-03-25
Fix enum default: preventing scoped association build from applying scope value

silenceu:
### Motivation / Background

Fixes #57061

The `default:` option on `enum` silently breaks scoped association `.build` calls. When building a record through a scoped `has_many` (e.g., `has_many :write_posts, -> { write }`), the scope's value is ignored and the record gets the enum default instead.

This is a regression from the old `_default:` behavior. Since Rails 7.2 stable rejects `_default:` with `ArgumentError`, upgrading forces migration to `default:` which silently introduces this bug.

### Detail

**Root cause:** `enum :action, { write: 0, read: 1 }, default: :read` calls `attribute :action, default: :read`, storing the raw symbol `:read`. Dirty tracking sees this as "changed" compared to the DB column state. In `Association#initialize_attributes`, "changed" attributes are excluded from scope application, so the scope's value is silently skipped.

**Two changes fix this:**

1. **`activerecord/lib/active_record/enum.rb`** — Resolve the `default:` label (e.g., `:read`) to its mapped DB value (e.g., `1`) before passing to `attribute()`. This makes `value_before_type_cast` store the integer, not a raw symbol.

2. **`activerecord/lib/active_record/associations/association.rb`** — In `initialize_attributes`, only treat attributes as "already assigned" if they were explicitly set by the user (differing from the class-level default), not merely defaulted via `attribute default:`. This ensures scoped associations can override enum defaults during `.build`.

### Additional information

Reproduction script from the issue:

```ruby
  belongs_to :user
  enum :action, { write: 0, read: 1 }, default: :read
end

  has_many :write_posts, -> { write }, class_name: "Post"
end

user = User.create!
post = user.write_posts.build
post.action # => "read" (BUG: should be "write")
```

Affected versions: Rails 7.0+ through main.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,25 @@
+*   Fix `enum` with `default:` preventing scoped association `.build` from applying scope values.
+
+    When an enum declared a `default:`, building a record through a scoped `has_many`
+    (e.g., `has_many :write_posts, -> { write }`) would silently ignore the scope's
+    value and use the enum default instead.
+
+    ```ruby
+    class Post < ActiveRecord::Base
+      enum :action, { write: 0, read: 1 }, default: :read
+    end
+
+    class User < ActiveRecord::Base
+      has_many :write_posts, -> { write }, class_name: "Post"
+    end
+
+    user = User.create!
+    post = user.write_posts.build
+    post.action # => "write" (was incorrectly "read")
+    ```
+
+    *Song Huang*
+
 *   Deprecate the `schema_order` option in PostgreSQL database configurations.
 
     Use `schema_search_path` instead. The `schema_order` alias will be
--- a/activerecord/lib/active_record/associations/association.rb
+++ b/activerecord/lib/active_record/associations/association.rb
@@ -217,7 +217,15 @@ module ActiveRecord
       def initialize_attributes(record, except_from_scope_attributes = nil) # :nodoc:
         except_from_scope_attributes ||= {}
         skip_assign = [reflection.foreign_key, reflection.type].compact
-        assigned_keys = record.changed_attribute_names_to_save
+        default_attributes = record.class._default_attributes
+        assigned_keys = record.changed_attribute_names_to_save.select { |attr_name|
+          # Only treat attributes as "already assigned" if they were explicitly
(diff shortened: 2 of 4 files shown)


tylerwood-io:
Good root cause analysis. The two-part fix makes sense:

**Part 1 (enum.rb):** Resolving the label to its DB value before passing to \`attribute()\` is a standalone improvement — storing \`:read\` (symbol) as \`value_before_type_cast\` instead of \`1\` (integer) could cause other subtle issues beyond this bug.

**Part 2 (association.rb):** This is what actually fixes the scoped build issue. It distinguishes "attribute changed because user explicitly passed a value" from "attribute changed because the class declared a default." The comparison \`@attributes[attr_name] != default_attributes[attr_name]\` works because deep-duped \`UserProvidedDefault\` objects compare equal to the originals, while \`FromUser\` objects (created by explicit \`assign_attributes\`) do not.

One edge case I checked: if a user explicitly passes the default value (\`user.write_posts.build(action: :read)\`), \`assign_attributes\` creates a \`FromUser\` attribute, which is \`!=\` the \`UserProvidedDefault\` default — so it's correctly treated as user-assigned and the scope doesn't overwrite it.

Two notes:

1. This is a bug fix with observable behavior change, so it likely warrants a CHANGELOG entry (the checklist currently has it unchecked).

2. The \`instance_variable_get(:@attributes)\` in Part 2 is reaching into record internals. This is fine given the code is already deep in AR's association layer, but if there's a public accessor for the attribute set that returns the same objects, that might be marginally cleaner.

afurm:
I reproduced this issue locally as well and opened [#57095](https://github.com/rails/rails/pull/57095) while investigating.

That PR overlaps with this one, but the `association.rb` change takes a slightly different approach: it avoids `record.instance_variable_get(:@attributes)` and instead uses `attributes_with_values(...)` to filter `ActiveModel::Attribute::UserProvidedDefault` attributes from `changed_attribute_names_to_save`.

I verified that approach against the enum regression and the nearby association tests:

- `test_build_from_association_should_respect_scope`
- `test_build_and_create_from_association_should_respect_passed_attributes_over_default_scope`

If this PR is the preferred direction, feel free to ignore #57095 - I just wanted to share the alternative implementation in case it is useful.

PR #17654, opened 2014-11-17
If specify `strict: :default` explicitly, do not set sql_mode.

kamipo:
Related with https://github.com/rails/rails/issues/17370.

PR #57070, opened 2026-03-26
Consolidate configure_connection and allow skipping individual settings

eileencodes:
Refactor PostgreSQL `configure_connection` to build a hash of all session settings, then set them in a single loop through `internal_set_config`. This method checks PostgreSQL's `parameter_status` before issuing SET commands, skipping queries when the server already has the desired value.

Individual settings can be skipped by setting them to `false` in config, which is useful when connecting through a load balancer or proxy:

```yaml
production:
  adapter: postgresql
  standard_conforming_strings: false
  intervalstyle: false
  min_messages: false
  schema_search_path: false
```

For MySQL, `wait_timeout` can be skipped with `wait_timeout: false`, and `sql_mode` can be skipped via `variables: { sql_mode: false }` (consistent with the existing `:default` behavior).

Also deprecate `set_standard_conforming_strings` which is now handled through the consolidated settings hash, and add a `parameter_status` check to `schema_search_path=` to skip redundant SET queries.

Related https://github.com/rails/rails/pull/57013

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
@@ -1032,17 +1032,24 @@ module ActiveRecord
           variables = @config.fetch(:variables, {}).stringify_keys
 
           # Increase timeout so the server doesn't disconnect us.
-          wait_timeout = self.class.type_cast_config_to_integer(@config[:wait_timeout])
-          wait_timeout = 2147483 unless wait_timeout.is_a?(Integer)
-          variables["wait_timeout"] = wait_timeout
+          # Set to false in config to skip.
+          unless @config[:wait_timeout] == false
+            wait_timeout = self.class.type_cast_config_to_integer(@config[:wait_timeout])
+            wait_timeout = 2147483 unless wait_timeout.is_a?(Integer)
+            variables["wait_timeout"] = wait_timeout
+          end
 
           defaults = [":default", :default].to_set
 
           # Make MySQL reject illegal values rather than truncating or blanking them, see
           # https://dev.mysql.com/doc/refman/en/sql-mode.html#sqlmode_strict_all_tables
           # If the user has provided another value for sql_mode, don't replace it.
-          if sql_mode = variables.delete("sql_mode")
-            sql_mode = quote(sql_mode)
+          # Set sql_mode to false or :default in variables to skip.
+          if variables.key?("sql_mode")
+            sql_mode_value = variables.delete("sql_mode")
+            unless sql_mode_value == false || defaults.include?(sql_mode_value)
+              sql_mode = quote(sql_mode_value)
+            end
           elsif !defaults.include?(strict_mode?)
             if strict_mode?
               sql_mode = "CONCAT(@@sql_mode, ',STRICT_ALL_TABLES')"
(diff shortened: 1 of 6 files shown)


activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
```diff
@@ -1092,6 +1107,16 @@ def reconfigure_connection_timezone
           end
         end
 
+        # Sets a PostgreSQL session configuration variable. Uses parameter_status
+        # to skip redundant SET commands when the server already has the desired
+        # value. Pass nil as value to SET TO DEFAULT.
+        def internal_set_config(setting, value)
+          return if value && @raw_connection.parameter_status(setting) == value.to_s
```
matthewd: Looks like this needs `with_raw` for `set_standard_conforming_strings` callers, even though it's [basically?] guaranteed to be safe when called from `configure_connection`

activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
kamipo, line 1060: By adding `false` to the `defaults`, the else branch of `strict_mode?` now becomes dead code. Should this be deprecated?

Actually, the behavior of forcibly switching to non-strict mode with `strict: false` was introduced in Rails 4.2 (https://github.com/rails/rails/pull/16065).

(Prior to that, `strict: false` was a no-op.)

The option `strict: :default` was later introduced to preserve the previous behavior (https://github.com/rails/rails/pull/17654).

Originally, these behaviors could already be achieved using `sql_mode: ''` or `sql_mode: :default`.

As of today (2026), there is neither a need nor a recommendation to run MySQL in non-strict-mode, so I'd say this feature can be deprecated. What do you think?

eileencodes: There's a couple things we want to deprecate which I'll do in followup PRs. Thanks for flagging this one.

jjb:
> Individual settings can be skipped by setting them to false in config, which is useful when connecting through a load balancer or proxy:

This is so awesome!! This solves the [AWS RDS Proxy pinning problem](https://www.smily.com/engineering/rds-database-migration-series---integrating-ruby-on-rails-applications-with-rds-proxy).

One concern is the need to explicitly specify `false` - there is the risk that a future version of rails will introduce a new type of config that has a default, tests pass and performance is fine in staging, but then when it hits prod all the connections get pinned. Maybe could be solved with a test in the suite which boots rails and observes all sql coming over the wire and makes sure none are `SET`. would be nice to have an option to never issue `SET`, something like...

```yaml
production:
  adapter: postgresql
  skip_all_set_commands: true
```

```ruby
def internal_set_config(setting, value)
  return if settings['skip_all_set_commands']
```

matthewd:
@jjb yeah, that's a design trade-off: the "just don't SET anything" alternative has the danger that the unanticipated new variable, which our adapter felt important to set, is not -- with no user action. To me that feels like a bigger _correctness_ risk, without available mitigations (configuring a proxy to refuse queries that would pin, or monitoring and alerting on pinned connections, for example).

jjb:
something to maybe consider is that some of these settings are "operational" (timeouts) and some of them are "semantic/behavioral" (timezone, intervalstyle).

we just got hit with a rare but consistent bug because, without the SET commands live, intervalstyle is set to `postgres`, which makes Rails' handling of intervals break.

it would be cool if there were a mode where all these settings could be specified in database.yml, but then it would only do `SHOW` on them, and if they don't match expectation, rails boot would fail

[maybe not the ideal place to discuss this but i know a lot of people disabling SET in order to work with RDS Proxy are likely to deal with this and read this ticket]

PR #57071, opened 2026-03-26
Fix titleize to capitalize unicode lowercase letters

EldinGuzin:
Inflector#titleize uses [a-z] in its regex to match word-initial lowercase letters for capitalization. This only matches ASCII characters, so Unicode lowercase letters like đ, é, ü, ñ, ć are silently skipped and left lowercase.

- Before
`titleize("ćasim đipa") # => "ćasim đipa"`

- After
`titleize("ćasim đipa") # => "Ćasim Đipa"`

Fix: Replace [a-z] with \p{Ll} in the titleize regex

### DIFF
--- a/activesupport/lib/active_support/inflector/methods.rb
+++ b/activesupport/lib/active_support/inflector/methods.rb
@@ -190,7 +190,7 @@ module ActiveSupport
     #   titleize('raiders_of_the_lost_ark')                      # => "Raiders Of The Lost Ark"
     #   titleize('string_ending_with_id', keep_id_suffix: true)  # => "String Ending With Id"
     def titleize(word, keep_id_suffix: false)
-      humanize(underscore(word), keep_id_suffix: keep_id_suffix).gsub(/\b(?<!\w['’`()])[a-z]/) do |match|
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/inflector/methods.rb
```diff
@@ -190,7 +190,7 @@ def downcase_first(string)
     #   titleize('raiders_of_the_lost_ark')                      # => "Raiders Of The Lost Ark"
     #   titleize('string_ending_with_id', keep_id_suffix: true)  # => "String Ending With Id"
     def titleize(word, keep_id_suffix: false)
-      humanize(underscore(word), keep_id_suffix: keep_id_suffix).gsub(/\b(?<!\w['’`()])[a-z]/) do |match|
+      humanize(underscore(word), keep_id_suffix: keep_id_suffix).gsub(/\b(?<!\w['’`()])\p{Ll}/) do |match|
```
byroot:
```suggestion
      humanize(underscore(word), keep_id_suffix: keep_id_suffix).gsub(/\b(?<!\w['’`()])\p{Lower}/) do |match|
```
`\p{Lower}` would be more readable.

PR #57077, opened 2026-03-27
Deprecate the `strict` option in MySQL database configurations

eileencodes:
The `strict` config option for MySQL was introduced in Rails 4.2 to control sql_mode, but the same behavior can be achieved directly with `variables: { sql_mode: "..." }`.

As of 2026, there is no recommendation to run MySQL in non-strict mode, so this option is no longer needed. Users should migrate to:

- `strict: true`    → remove it (this is the default)
- `strict: false`   → `variables: { sql_mode: "" }`
- `strict: :default` → `variables: { sql_mode: :default }`

cc/ @kamipo

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Deprecate the `strict` option in MySQL database configurations.
+
+    The `strict` option for MySQL will be removed in Rails 8.3 because it is the default behavior.
+
+    To change the default behavior of `strict`, use `variables: { sql_mode: "..." }` to configure `sql_mode` directly.
+
+
+    `strict: false` can be replaced with `variables: { sql_mode: "" }`, and `strict: :default` can be replaced with `variables: { sql_mode: :default }`.
+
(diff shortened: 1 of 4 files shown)


joshuapinter:
@eileencodes Just to clarify, we have the following in our `database.yml` config: 

```yml
  variables:
    sql_mode: TRADITIONAL
```

Can this be removed as "strict" is the new default or should we keep this as it's doing more than just what "strict" does? 

Thanks!

eileencodes:
`TRADITIONAL` combines `STRICT_ALL_TABLES` with `STRICT_TRANS_TABLES`, `ERROR_FOR_DIVISION_BY_ZERO`, `NO_ENGINE_SUBSTITUTION, NO_ZERO_DATE`,  and `NO_ZERO_IN_DATE` where removing it would apply only `STRICT_ALL_TABLES`. So if you don't want to change your config, leave it as is.

joshuapinter:
@eileencodes Beauty, thanks for clarifying!

PR #57083, opened 2026-03-27
Fix has_many :through to use join model's custom attribute type

tylerwood-io:
### Motivation / Background

Fixes #56893.

When a join model declares a custom attribute type via `attribute :col, MyType.new`, `has_many :through` queries silently ignore that type when building the WHERE clause for the cross-table FK condition. The value is serialized using the database schema's default type instead of the model's declared type.

This affects any application using custom attribute types on join model FK columns with `has_many :through`. A real-world example is the `ulid-rails` gem, which registers a `:ulid` type that serializes between ULID strings and 16-byte binary — without this fix, through queries encode the FK as ASCII instead of binary, returning empty results.

Previously reported as #54418 and #52135.

### Detail

`AssociationScope#apply_scope` has two branches:

```ruby
def apply_scope(scope, table, key, value)
  if scope.table == table
    scope.where!(key => value)          # same-table: uses join model's type ✓
  else
    scope.where!(table.name => { key => value })  # cross-table: uses target model's type ✗
  end
end
```

The cross-table branch uses a hash-form condition that resolves the column type via the outer scope's `PredicateBuilder`. Since the target model has no knowledge of custom types declared on the join model, it falls back to the database schema type.

The fix builds an explicit `QueryAttribute` using the join table's own type caster:

```ruby
bind = Relation::QueryAttribute.new(key, value, table.type_for_attribute(key))
scope.where!(table[key].eq(bind))
```

### Benchmark Results

Ruby 3.3.10, SQLite3 in-memory, 5 records, `benchmark-ips`:

```
                         i/s      per-iter
direct has_many:         7,170.0  139.47 μs
through association:     7,076.4  141.32 μs  — same-ish (within error)
```

No measurable regression from the `QueryAttribute` construction.

### Test Plan

- [x] Test with inline models using a custom `ScaledInteger` type that multiplies FK values by 1000 — verifies the through query uses the custom type (fails without fix: returns `[]`)
- [x] Full `has_many_through_associations_test.rb` suite passes (173 runs, 475 assertions, 0 failures)
- [x] Full `activerecord:sqlite3:test` suite passes (9,379 runs, 32,097 assertions, 0 failures)
- [x] CHANGELOG entry added

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Fix `has_many :through` to use join model's custom attribute type in queries.
+
+    When a join model declares a custom attribute type via
+    `attribute :col, MyType.new`, the through association's cross-table
+    WHERE clause silently ignored it, falling back to the database
+    schema's default type.
+
+    Fixes #56893.
+
+    *Tyler Wood*
+
 *   Deprecate the `strict` option in MySQL database configurations.
 
     The `strict` option for MySQL will be removed in Rails 8.3 because it is the default behavior.
--- a/activerecord/lib/active_record/associations/association_scope.rb
+++ b/activerecord/lib/active_record/associations/association_scope.rb
@@ -162,7 +162,8 @@ module ActiveRecord
           if scope.table == table
             scope.where!(key => value)
           else
-            scope.where!(table.name => { key => value })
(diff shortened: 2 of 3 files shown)


tylerwood-io:
Closing in favor of #56895 which addresses the same issue (#56893) with the same fix — explicit QueryAttribute in AssociationScope#apply_scope's cross-table branch. They filed a month before us and have been passing CI since submission.

PR #57086, opened 2026-03-27
Fix parallel test shutdown hang when workers are stuck alive

### LINKED ISSUE #57085: Parallelization#shutdown hangs when workers are stuck alive (not zombies)
### Steps to reproduce

#57053 (fixing #57052) added `reap_dead_workers` inside `Server#shutdown`'s `wait_for_active_workers` loop, which detects workers that have **died** (become zombies) via `Process.waitpid(pid, WNOHANG)`.

However, `waitpid(WNOHANG)` returns `nil` for processes that are still **alive**. If a worker hangs without exiting — e.g. stuck in a finalizer, DRb thread, `at_exit` hook, or `parallelize_teardown` callback — the shutdown hangs forever at one of two points:

1. **`Server#shutdown`** — `active_workers?` is true (worker never called `stop_worker`), `reap_dead_workers` finds nothing (worker is alive), loop repeats indefinitely.
2. **`Parallelization#shutdown`** — worker called `stop_worker` successfully, so `Server#shutdown` completes, but the final blocking `Process.waitpid(pid)` hangs forever because the worker process is still running.

We're hitting this intermittently in CI with system tests (Capybara/Cuprite). Workers get stuck in Chrome cleanup or Ruby shutdown finalizers, the parent blocks on waitpid, and a watchdog has to kill everything ~90s later.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # Test against edge Rails (includes #57053 fix)
end

require "active_support"
require "minitest/autorun"

class ParallelShutdownStuckAliveTest < ActiveSupport::TestCase
  # Worker hangs alive without deregistering.
  # Server#shutdown's reap_dead_workers can't help (waitpid returns nil).
  def test_shutdown_hangs_when_worker_is_stuck_alive
    server = parallelization.instance_variable_get(:@queue_server)
    url = parallelization.instance_variable_get(:@url)

    blocking_distributor = ActiveSupport::Testing::Parallelization::SharedQueueDistributor.new
    server.instance_variable_set(:@distributor, blocking_distributor)

    worker_pid = fork do
      DRb.stop_service
      queue = DRbObject.new_with_uri(url)
      queue.start_worker(0, Process.pid)
      sleep 999 # stuck alive — never exits, never calls stop_worker
    end
    parallelization.instance_variable_set(:@worker_pool, [worker_pid])

    sleep 0.25
    assert server.active_workers?

    # This hangs forever — reap_dead_workers finds nothing (worker is alive)
    assert_raises(Timeout::Error) do
      Timeout.timeout(3) { parallelization.shutdown }
    end
  ensure
    Process.kill("KILL", worker_pid) rescue nil
    Process.waitpid(worker_pid) rescue nil
  end

  # Worker deregisters but hangs during exit.
  # Server#shutdown completes, but final waitpid blocks forever.
  def test_shutdown_hangs_on_final_waitpid
    server = parallelization.instance_variable_get(:@queue_server)
    url = parallelization.instance_variable_get(:@url)

    blocking_distributor = ActiveSupport::Testing::Parallelization::SharedQueueDistributor.new
    server.instance_variable_set(:@distributor, blocking_distributor)

    worker_pid = fork do
      DRb.stop_service
      queue = DRbObject.new_with_uri(url)
      queue.start_worker(0, Process.pid)
      sleep 0.5
      queue.stop_worker(0, Process.pid)
      sleep 999 # stuck alive during exit
    end
    parallelization.instance_variable_set(:@worker_pool, [worker_pid])

    sleep 1

    # Server#shutdown completes (worker deregistered), but
    # Parallelization#shutdown's Process.waitpid(pid) blocks forever
    assert_raises(Timeout::Error) do
      Timeout.timeout(3) { parallelization.shutdown }
    end
  ensure
    Process.kill("KILL", worker_pid) rescue nil
    Process.waitpid(worker_pid) rescue nil
  end
end
```

### Expected behavior

`Parallelization#shutdown` should time out and force-kill workers that don't exit within a reasonable period, rather than hanging indefinitely.

### Actual behavior

`Parallelization#shutdown` hangs forever — either in `Server#shutdown`'s `wait_for_active_workers` loop (scenario 1) or in the final blocking `Process.waitpid(pid)` call (scenario 2).

### System configuration

**Rails version**: main (includes #57053)

**Ruby version**: 3.3.x+

markedmondson:
### Motivation / Background

#57053 (fixing #57052) added `reap_dead_workers` inside `Server#shutdown`'s `wait_for_active_workers` loop to detect workers that have **died** (zombies) via `Process.waitpid(pid, WNOHANG)`. This works when workers are dead, but `waitpid(WNOHANG)` returns `nil` for processes that are still alive.

If a worker hangs without exiting — stuck in a finalizer, DRb thread, `at_exit` hook, or `parallelize_teardown` callback — shutdown hangs forever at one of two points:

1. `Server#shutdown` — worker is alive but never called `stop_worker`, `reap_dead_workers` finds nothing, loop repeats indefinitely.
2. `Parallelization#shutdown` — worker called `stop_worker`, so `Server#shutdown` completes, but the final blocking `Process.waitpid(pid)` hangs forever.

We're hitting this intermittently in CI with system tests (Capybara/Cuprite). Workers get stuck during Chrome cleanup or Ruby shutdown, and a watchdog has to kill everything ~90s later.

Fixes #57085

### Detail

Adds a 30-second timeout to `Parallelization#shutdown`'s worker wait. Instead of a blocking `Process.waitpid(pid)`, it polls with `waitpid(WNOHANG)` against a monotonic clock deadline. If workers don't exit within the timeout, they are force-killed with `SIGKILL` and reaped.

The timeout is defined as `SHUTDOWN_TIMEOUT = 30` on `Parallelization`, which is long enough for normal cleanup (teardown hooks, DRb deregistration) but short enough to avoid stalling CI pipelines.

### Additional information

Two regression tests are included:
1. Worker stuck alive without deregistering — `Server#shutdown` would hang in `wait_for_active_workers`
2. Worker deregistered but stuck during exit — final `Process.waitpid` would block forever

Both tests override `SHUTDOWN_TIMEOUT` to 1 second so they complete quickly. Without the fix, both hang; with it, shutdown completes cleanly via force-kill.

### DIFF
--- a/activesupport/lib/active_support/testing/parallelization.rb
+++ b/activesupport/lib/active_support/testing/parallelization.rb
@@ -65,6 +65,12 @@ module ActiveSupport
         @worker_count
       end
 
+      # How long to wait for workers to exit during shutdown before
+      # force-killing them. Must be long enough for normal cleanup
+      # (parallelize_teardown hooks, DRb deregistration) but short
+      # enough to avoid stalling CI pipelines indefinitely.
+      SHUTDOWN_TIMEOUT = 30 # seconds
+
       def shutdown
         dead_worker_pids = @worker_pool.filter_map do |pid|
           Process.waitpid(pid, Process::WNOHANG)
@@ -73,13 +79,64 @@ module ActiveSupport
         end
         @queue_server.remove_dead_workers(dead_worker_pids)
 
-        @queue_server.shutdown
-        @worker_pool.each do |pid|
-          Process.waitpid(pid)
+        @queue_server.shutdown(timeout: SHUTDOWN_TIMEOUT)
+
+        if @queue_server.active_workers?
+          force_kill_workers
+          @queue_server.remove_dead_workers(@worker_pool)
+        else
+          wait_for_workers
+        end
+      end
+
+      private
+        def wait_for_workers
+          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + SHUTDOWN_TIMEOUT
+
+          @worker_pool.each do |pid|
+            remaining = deadline - Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+            if remaining <= 0
+              force_kill_workers
+              @queue_server.remove_dead_workers(@worker_pool)
+              return
+            end
+
+            wait_for_worker(pid, remaining)
+          end
+        end
+
+        def wait_for_worker(pid, timeout)
+          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+
+          loop do
+            return if Process.waitpid(pid, Process::WNOHANG)
+
+            if Process.clock_gettime(Process::CLOCK_MONOTONIC) > deadline
+              force_kill_workers
+              @queue_server.remove_dead_workers(@worker_pool)
+              return
+            end
+
+            sleep 0.1
+          end
         rescue Errno::ECHILD
           nil
         end
-      end
+
+        def force_kill_workers
+          @worker_pool.each do |pid|
+            Process.kill("KILL", pid)
+          rescue Errno::ESRCH
+            nil
+          end
+
+          @worker_pool.each do |pid|
+            Process.waitpid(pid)
+          rescue Errno::ECHILD
+            nil
+          end
+        end
     end
   end
 end
--- a/activesupport/lib/active_support/testing/parallelization/server.rb
+++ b/activesupport/lib/active_support/testing/parallelization/server.rb
@@ -68,7 +68,7 @@ module ActiveSupport
           @distributor.interrupt
         end
 
-        def shutdown
+        def shutdown(timeout: nil)
           # Wait for initial queue to drain
           while @distributor.pending?
             sleep 0.1
@@ -76,7 +76,7 @@ module ActiveSupport
 
           @distributor.close
 
-          wait_for_active_workers
+          wait_for_active_workers(timeout: timeout)
 
           @in_flight.values.each do |(klass, name, reporter)|
             result = Minitest::Result.from(klass.new(name))
@@ -92,13 +92,18 @@ module ActiveSupport
 
           @distributor.close
 
-          wait_for_active_workers
+          wait_for_active_workers(timeout: timeout)
         end
 
         private
-          def wait_for_active_workers
+          def wait_for_active_workers(timeout: nil)
+            deadline = if timeout
+              Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+            end
+
             while active_workers?
               reap_dead_workers
+              break if deadline && Process.clock_gettime(Process::CLOCK_MONOTONIC) > deadline
               sleep 0.1
             end
           end
--- a/activesupport/test/parallelization_test.rb
+++ b/activesupport/test/parallelization_test.rb
@@ -46,15 +46,16 @@ class ParallelizationTest < ActiveSupport::TestCase
 
     parallelization.start
 
-    sleep 0.5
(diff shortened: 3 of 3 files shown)


byroot:
I haven't looked at the patch yet, but the CI failure is legit.

markedmondson:
Sorry about that, got it now I think.

activesupport/lib/active_support/testing/parallelization.rb
```diff
@@ -73,13 +79,61 @@ def shutdown
         end
         @queue_server.remove_dead_workers(dead_worker_pids)
 
-        @queue_server.shutdown
-        @worker_pool.each do |pid|
-          Process.waitpid(pid)
+        Timeout.timeout(SHUTDOWN_TIMEOUT) { @queue_server.shutdown }
```
byroot: We shouldn't need to use `Timeout.timeout` here. `@queue_server.shutdown` uses a sleep loop, we can pass the timeout as argument.

markedmondson: Yeah good point. I'm thinking we add a `timeout:` keyword to `Server#shutdown` / `wait_for_active_workers`, and break out of the sleep loop when the deadline is exceeded — then check `active_workers?` after it returns to decide whether to force-kill. Something like:

```ruby
def wait_for_active_workers(timeout: nil)
  deadline = if timeout
    Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
  end

  while active_workers?
    reap_dead_workers
    break if deadline && Process.clock_gettime(Process::CLOCK_MONOTONIC) > deadline
    sleep 0.1
  end
end
```

Is that what you had in mind?

byroot: Yes.

byroot:
I made various improvements to your tests: https://github.com/byroot/rails/commit/ffaee516b3c96c7562db26decde1e58397d83dd9, and squashed your commits.

I'll merge on green.

markedmondson:
Thanks for the improvements @byroot! Just wanted to share that we can confirm the `Timeout.timeout` wrapper approach causes a real regression in production.

We shipped a version that wrapped `@queue_server.shutdown` with `Timeout.timeout(30)` and it killed workers mid-run — the 30s timeout fires during the queue drain phase (`sleep 0.1 until @queue.empty?`), which takes minutes for a large test suite.

The `timeout:` parameter approach in this PR is exactly the right fix — deadline on `wait_for_active_workers` only, queue drain stays untimed. We've adopted the same pattern in our local monkeypatch while we wait for this to land.

markedmondson:
We've been running this patch in CI for ~1 day and hit an issue with the 30s `SHUTDOWN_TIMEOUT`.

Phase 2 (`wait_for_active_workers`) starts when the queue drains — i.e. when all tests have been **popped**, not when they've **finished**. So the 30s timeout has to cover the slowest in-progress test plus cleanup time, not just cleanup.

We're seeing system tests that take 30-50s get force-killed mid-run, reported as "result not reported". The comment says the timeout is for "normal cleanup (parallelize_teardown hooks, DRb deregistration)" but it actually covers remaining test execution time too.

We bumped to 60s locally. Might be worth either increasing the default or making it configurable via an env var / class attribute so projects with slow system tests can adjust it.

Also: `report_in_flight_tests` creates a `Minitest::Result` with `time = nil`, which crashes `minitest-reporters` JunitReporter (`nil can't be coerced into Float`). Setting `result.time = 0` fixes it — might be worth adding upstream too.

PR #57087, opened 2026-03-28
Only set Postgres timezone when needed

matthewd:
Follow-up for #57070: `TimeZone` needed a bit of extra help to work with the parameter_status check, including needing to treat `variables` keys case-insensitively.

### DIFF
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -15,6 +15,11 @@ module ActiveRecord
       include DdlHelper
       include ConnectionHelper
 
+      # Force known GUC defaults at connect time so query-count
+      # assertions are deterministic regardless of server configuration.
+      # TimeZone must be non-UTC so the :utc/:local distinction is visible.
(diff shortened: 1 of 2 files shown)


activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -348,23 +358,24 @@ def test_queries_executed_on_fresh_connection_through_first_select
...
-              connection.select_value("SELECT 1+2")
+              capture_sql(include_schema: true) do
+                connection.select_value("SELECT 1+2")
+              end
             end
           end
         end.map(&:squish)
 
         expected_queries = [
-          "SET SESSION timezone TO 'UTC'",
```
matthewd, line 365: This is the main result here: under appropriate configuration, we now skip this query, meaning that (with a PG18+ server) it is possible to complete `configure_connection` without running any queries at all.

eileencodes (approved):

activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
jjb: why is 18+ required?

matthewd: That's when `search_path` became `GUC_REPORT` -> is obtained for free via `parameter_status` without needing the `SHOW`

PR #57089, opened 2026-03-28
PostgreSQL Query Pipelining

matthewd:
PostgreSQL's pipeline mode allows us to have multiple queries in flight, then collect them all afterwards, saving a network round-trip per query. The gain is proportional to latency, so it matters most for cloud-hosted or distributed databases where even a "nearby" server is a few milliseconds away.

This builds on the QueryIntent setup (#55897, #56129, #56267), which has already separated query submission from result consumption.

Currently:
* Explicit use: `pipeline: true` on query methods, paralleling `async:`. Uses the "current" connection, so it's usable in situations where async would not work.
* Automatic: async-eligible queries that can't actually go async (inside a transaction, or no executor thread available) are transparently pipelined instead, so they can still be non-blocking.

Future:
* More explicit APIs: making a model `save!` able to be fully pipelined within a surrounding transaction (?)
* More automatic behaviour: transparently pipelining Relation preload queries [for peer associations, where there's no data dependency]

The bulk of the work here is in robustness around connection failures. A dead connection mid-pipeline is trickier than the single-query case, because all the in-flight queries need to be considered for retry/replay, and that can get complicated.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -521,7 +521,16 @@ module ActiveRecord
               @connections.each do |conn|
                 if conn.in_use?
                   conn.steal!
-                  checkin conn
+                  begin
+                    checkin conn
+                  rescue
+                    # Checkin may fail if the connection is dead (e.g.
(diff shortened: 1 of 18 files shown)


kbrock:
Ping if you need a second set of eyes.

PR #57094, opened 2026-03-30
Delay engine route building

byroot:
Followup: https://github.com/rails/rails/pull/52353

Engines that call `routes do` directly in their definition end up allocating the route object before `:make_routes_lazy` is invoked, entirely defeating the optimization.

That's the case of the `graphql` gem for instance: https://github.com/rmosolgo/graphql-ruby/blob/fa35e798c61683376b4b1bc050ffea29ece7d08c/lib/graphql/dashboard.rb#L47

Instead we can defer evaluating the route blocks to after the `route_set_class` has been configured.

FYI: @gmcgibbon

### DIFF
--- a/railties/lib/rails/engine.rb
+++ b/railties/lib/rails/engine.rb
@@ -542,8 +542,20 @@ module Rails
     # Defines the routes for this engine. If a block is given to
     # routes, it is appended to the engine.
     def routes(&block)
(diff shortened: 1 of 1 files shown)


gmcgibbon:
Wasn't aware of this usage. Thank you!

PR #57100, opened 2026-03-30
Passkeys

monorkin:
### Motivation / Background

Passkeys are a modern authentication method built on the WebAuthn standard that uses public-key cryptography instead of shared secrets — this makes them resistant to phishing and leaks. The standard has wide browser and platform support, and allows users to log in with their face, fingerprint, password manager, or authenticator.

https://github.com/user-attachments/assets/3b7352a6-09f5-4a93-b164-393a73fc1fef
 
This PR adds WebAuthn support to Action Pack along with a Passkey model and helpers for adding passkey-based authentication to any Rails app. It also updates the authentication generator to use magic links and passkeys as a more secure default to passwords. In case someone prefers passwords, they can still invoke the old generator by passing it `--password-based` as an option.

https://github.com/user-attachments/assets/05736520-8872-4d57-a5c0-f045bbaef98e

### Detail

Any record can add passkeys by calling `has_passkeys` and passing it `name` and `display_name` as arguments:

```ruby
  has_passkeys name: :email_address, display_name: :full_name
end
```

The `name` functions as the account identifier for the passkey — this is the name of the passkey. Typically this would be an email address or username. And the `display_name` is a friendly label that's shown by the browser during the passkey registration ceremony. These are the only required attributes. But any holder-specific part of the authentication and registration ceremony can be configured here:

```ruby
  has_passkeys do |config|
    config.registration_options do
      {
        name: email_address,
        display_name: name.split(" ").first
      }
    end

    config.authentication_options do
      { user_verification: "required" }
    end
  end
end
```

`has_passkeys` adds a `passkeys` association to the model calling it. Through that association new passkeys can be registered or existing passkeys authenticated:

```ruby
# Registration
class My::PasskeysController < ApplicationController
  include ActionPack::Passkeys::Request

  def new
    @registration_options = passkey_registration_options(holder: Current.user)
  end

  def create
    ActionPack::Passkey.register(passkey_registration_params, holder: Current.user)
    redirect_to my_passkeys_path
  end
end
```

```ruby
# Authentication
class Sessions::PasskeysController < ApplicationController
  include ActionPack::Passkeys::Request
  
  def new
    @authentication_options = passkey_authentication_options
  end

  def create
    if passkey = ActionPack::Passkey.authenticate(passkey_authentication_params)
      start_new_session_for passkey.holder
      redirect_to root_path
    else
      redirect_to new_session_path, alert: "Authentication failed"
    end
  end
end
```
The `ActionPack::Passkeys::Request` concern adds a few methods that makes it easier to work with passkey requests:
- `passkey_authentication_params` — strong parameters for the authentication ceremony
- `passkey_registration_params` — strong parameters for the registration ceremony
- `passkey_authentication_options` — generates `RequestOptions` for `navigator.credentials.get()`
- `passkey_registration_options(holder:)` — generates `CreationOptions` for `navigator.credentials.create()`

It also adds a before action that sets the current host and origin on `ActionPack::WebAuthn::Current`. These are needed in both ceremonies and this makes it easier to pass them around.

For the frontend side there are two new view helpers:
- `passkey_registration_button` — renders a web component with a form and button for registering a new passkey
- `passkey_sign_in_button` — renders a web component with a form and button for authenticating with a passkey

These have very similar semantics to `button_to` and can be used to render buttons for signing in with a passkey or for registering one:
```erb
<%= passkey_registration_button "Register a new passkey", my_passkeys_path,
      options: @registration_options %>

<%= passkey_sign_in_button "Sign in with a passkey", session_passkey_path,
      options: @authentication_options %>
```

The buttons are backed by web components that handle the browser's WebAuthn API. To use them, the Action Pack JavaScript must be imported:

```ruby
# config/importmap.rb
pin "@rails/actionpack", to: "actionpack.esm.js"
```

```js
// app/javascript/application.js
import "@rails/actionpack"
```

The authentication generator sets this up automatically.

#### ActionPack - WebAuthn & Passkeys

There are two additions to ActionPack: Passkeys and WebAuthn

WebAuthn is the protocol backing the passkeys implementation. This library closely follows the browsers' naming of ceremonies and objects to make it easier to reason about what the code does.

The `CborDecoder` and `CoseKey` are the main part of the implementation. They decode the payload that the browser passes to the server.

CBOR is a binary representation of a JSON object (with some extra features). The authenticator passes a CBOR blob to the browser, and the browser passes it on to the server. This binary format is used because WebAuthn supports different attestation schemes, some of which require cryptographic operations over the authenticator's response.

Attestation, in the context of WebAuthn, is the process of verifying that the authentication device (or program) used is genuine. There are a handful of different attestation schemes out there, but in this PR I added only the simplest — "none", which doesn't perform any verification.

But, as attestation is common in environments with high security requirements, the interface of the library allows users to plug-in their own attestation verifier:

```ruby
class PackedAttestationVerifier
  def verify!(attestation, client_data_json:)
    # Verify the attestation statement signature against the
    # authenticator's certificate chain
  end
end

ActionPack::WebAuthn.register_attestation_verifier("packed", PackedAttestationVerifier.new)
```

All authenticators should work with just "none". We tested a range of different devices from Apple Passwords (Face ID, Touch ID), Samsung Pass, Google Passwords, 1Password, Windows Hello, Yubikeys 4 & 5, and all worked.

The CBOR decoder is implemented as a recursive descent parser. It takes a CBOR blob and returns a Ruby Hash. It supports the full spec including indefinite types, various different float types, and tags (which are ignored) to provide broad device support. To test it I generated various CBOR values using Python's CBOR2 package and added them to the test suite. I also copied any examples from the spec into the test suite.

COSE is a format for encoding encryption keys (you can think of it as a DER file, but it's easier to parse). The decoder here is a simple parser that takes a CBOR blob and returns an OpenSSL PKey.

Passkey is an ActiveRecord record and a wrapper around WebAuthn that provides an easy-to-use interface around the various passkey ceremonies. It handles persistence, registration and authentication which includes controller and view helpers as well as some JavaScript.

To make this work I added a Passkeys engine. This makes it easy to include JavaScript and migrations. The rest of this namespace are concerns that get mixed into ActiveRecord, ActionView and ActionController.

There is also a controller that's used for generating passkey challenges. In all ceremonies, the authenticator uses its private key to sign a challenge generated by the application. The best practice is to have expiring challenges to prevent replay attacks, and that's how they are implemented in this PR (using a message verifier). But that causes a new problem where a user could experience an error authenticating or registering a passkey if they wait too long before triggering the ceremony. So this controller is used to fetch a new challenge for the ceremony when it begins.
#### Railtie

The authentication generator now has two modes:

- Default (no flags)
  Generates magic link + passkey authentication. Creates a `MagicLink` model, a `Sessions::MagicLinksController`, and the magic link mailer. 
  Also adds `letter_opener` for development.
- `--password-based`
  The previous behavior. Creates `has_secure_password`, a `PasswordsController`, the password reset mailer, and adds `bcrypt`.

Both modes generate passkey support (the `Sessions::PasskeysController`, `My::PasskeysController`, routes, JavaScript, and the passkey migration).
#### ActiveSupport

Added `ActiveSupport::Base32` for generating and normalizing Crockford's Base32 strings. 

This encoding is designed to be unambiguous for humans (it excludes I, L, O, and U), and the normalizer is forgiving of common transcription mistakes (e.g., `O` → `0`, `I` → `1`). It's used for the magic link codes.
#### Guides

- Security guide
  Rewrote the Authentication section to cover passkeys, magic links, and password-based authentication as three distinct paths. Added a `has_passkeys` section explaining the model setup and block configuration.
- Getting Started guide
  Updated to use `--password-based` and added a note pointing to the Security guide for the default (passkey + magic link) mode.
- Configuring guide
  Added all `config.action_pack.passkey.*` options.
- Sign Up and Settings guide
  Small update to mention the `--password-based` flag.

##### Configurable parent class

Passkey allows you to configure it's parent class.
This allows for multi-database setups, but I don't know if that's a good way to go about this.
I saw other PRs that add multi-database support, to e.g. ActiveStorage, that take a more meta-programming approach to generating the models.

We have this in Fizzy, but this isn't necessary for passkeys to work, so if there are any issues with this approach I'd prefer to remove it.

##### `Passkey` vs `Passkeys`

`ActionPack::Passkey` and `ActionPack::Passkeys` are separate because Passkey is an ActiveRecord record which crashes if ActiveRecord isn't loaded. In Fizzy we worked around this by using a shim, but in Rails because I have an Engine it has to load before ActiveRecord is loaded and crashes. So I created a new namespace to make things easier.

The engine lives in a `passkeys` directory. To me this is the most logical place to put it, but it looks a bit off compared to the other directories there — any suggestions are welcome.

### Additional information

One of the goals of this passkey implementation was to have no external dependencies.

I've worked with @dhh on this. 

The bulk of the code was extracted from Fizzy:
- https://github.com/basecamp/fizzy/pull/2623
- https://github.com/basecamp/fizzy/pull/2748
- https://github.com/basecamp/fizzy/pull/2751
- https://github.com/basecamp/fizzy/pull/2758
- https://github.com/basecamp/fizzy/pull/1304

Official specifications:
- [Web Authentication (WebAuthn) Level 2](https://www.w3.org/TR/webauthn-2/)
- [RFC 8949 — Concise Binary Object Representation (CBOR)](https://tools.ietf.org/html/rfc8949)
- [RFC 9053 — CBOR Object Signing and Encryption (COSE)](https://datatracker.ietf.org/doc/html/rfc9053)

### Checklist

### DIFF
--- a/actionpack/lib/action_pack/passkeys/engine.rb
+++ b/actionpack/lib/action_pack/passkeys/engine.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+require "active_record/railtie"
+require_relative "../passkeys"
+require_relative "../web_authn"
+require "active_support/core_ext/numeric/bytes"
+
+module ActionPack
+  class Passkeys::Engine < Rails::Engine # :nodoc:
+    def self.find_root(from)
+      Pathname.new(File.expand_path("../../passkeys", __dir__))
+    end
+
+    isolate_namespace ActionPack::Passkeys
+
+    config.action_pack = ActiveSupport::OrderedOptions.new unless config.respond_to?(:action_pack)
+    config.action_pack.passkey = ActiveSupport::OrderedOptions.new
+    config.action_pack.passkey.parent_class_name = "ApplicationRecord"
+    config.action_pack.passkey.routes_prefix = "/rails/action_pack/passkey"
+    config.action_pack.passkey.draw_routes = true
+    config.action_pack.passkey.challenge_url = nil
+    config.action_pack.passkey.default_registration_options = {}
+    config.action_pack.passkey.default_authentication_options = {}
+    config.action_pack.passkey.registration_challenge_expiration = 10.minutes
+    config.action_pack.passkey.authentication_challenge_expiration = 5.minutes
+    config.action_pack.passkey.cbor_max_depth = 16
+    config.action_pack.passkey.cbor_max_size = 10.megabytes
+
+    initializer "action_pack.passkey.config" do
+      passkey_config = config.action_pack.passkey
+
+      ActionPack::Passkeys.parent_class_name = passkey_config.parent_class_name
+      ActionPack::Passkeys.default_registration_options = passkey_config.default_registration_options
+      ActionPack::Passkeys.default_authentication_options = passkey_config.default_authentication_options
+      ActionPack::Passkeys.registration_challenge_expiration = passkey_config.registration_challenge_expiration
+      ActionPack::Passkeys.authentication_challenge_expiration = passkey_config.authentication_challenge_expiration
+      ActionPack::Passkeys.challenge_url = passkey_config.challenge_url
+    end
+
+    initializer "action_pack.passkey.verifier" do
+      config.after_initialize do |app|
+        ActionPack::WebAuthn.challenge_verifier = app.message_verifier("action_pack.webauthn.challenge")
+        ActionPack::WebAuthn.application_name = app.name
+      end
+    end
+
+    initializer "action_pack.passkey.routes" do |app|
+      passkey_config = config.action_pack.passkey
+
+      app.routes.prepend do
+        if passkey_config.draw_routes
+          scope passkey_config.routes_prefix, as: :passkey do
+            post "/challenge" => "action_pack/passkeys/challenges#create", as: :challenge
+          end
+        end
+      end
+    end
+
+    initializer "action_pack.passkey.cbor" do
+      require "action_pack/web_authn/cbor_decoder"
+      ActionPack::WebAuthn::CborDecoder.max_depth = config.action_pack.passkey.cbor_max_depth
+      ActionPack::WebAuthn::CborDecoder.max_size = config.action_pack.passkey.cbor_max_size
+    end
+
+    initializer "action_pack.passkey.holder" do
+      ActiveSupport.on_load(:active_record) do
+        include ActionPack::Passkeys::Holder
+      end
+    end
+
+    initializer "action_pack.passkey.form_helper" do
+      ActiveSupport.on_load(:action_view) do
+        include ActionPack::Passkeys::FormHelper
+      end
+    end
+  end
+end
--- a/actionpack/lib/action_pack/passkeys/engine.rb
+++ b/actionpack/lib/action_pack/passkeys/engine.rb
@@ -23,8 +23,8 @@ module ActionPack
     config.action_pack.passkey.related_origins = []
     config.action_pack.passkey.default_registration_options = {}
     config.action_pack.passkey.default_authentication_options = {}
-    config.action_pack.passkey.registration_challenge_expiration = 10.minutes
-    config.action_pack.passkey.authentication_challenge_expiration = 5.minutes
+    config.action_pack.passkey.registration_timeout = 10.minutes
+    config.action_pack.passkey.authentication_timeout = 5.minutes
     config.action_pack.passkey.cbor_max_depth = 16
     config.action_pack.passkey.cbor_max_size = 10.megabytes
 
@@ -34,8 +34,8 @@ module ActionPack
       ActionPack::Passkeys.parent_class_name = passkey_config.parent_class_name
       ActionPack::Passkeys.default_registration_options = passkey_config.default_registration_options
       ActionPack::Passkeys.default_authentication_options = passkey_config.default_authentication_options
-      ActionPack::Passkeys.registration_challenge_expiration = passkey_config.registration_challenge_expiration
-      ActionPack::Passkeys.authentication_challenge_expiration = passkey_config.authentication_challenge_expiration
+      ActionPack::Passkeys.registration_timeout = passkey_config.registration_timeout
+      ActionPack::Passkeys.authentication_timeout = passkey_config.authentication_timeout
       ActionPack::Passkeys.challenge_url = passkey_config.challenge_url
       ActionPack::Passkeys.related_origins = passkey_config.related_origins
     end
--- a/actionpack/lib/action_pack/passkeys/engine.rb
+++ b/actionpack/lib/action_pack/passkeys/engine.rb
@@ -16,9 +16,11 @@ module ActionPack
     config.action_pack = ActiveSupport::OrderedOptions.new unless config.respond_to?(:action_pack)
     config.action_pack.passkey = ActiveSupport::OrderedOptions.new
     config.action_pack.passkey.parent_class_name = "ApplicationRecord"
+    config.action_pack.passkey.relying_party_id = nil
     config.action_pack.passkey.routes_prefix = "/rails/action_pack/passkey"
     config.action_pack.passkey.draw_routes = true
     config.action_pack.passkey.challenge_url = nil
+    config.action_pack.passkey.related_origins = []
     config.action_pack.passkey.default_registration_options = {}
     config.action_pack.passkey.default_authentication_options = {}
     config.action_pack.passkey.registration_challenge_expiration = 10.minutes
@@ -35,12 +37,14 @@ module ActionPack
       ActionPack::Passkeys.registration_challenge_expiration = passkey_config.registration_challenge_expiration
       ActionPack::Passkeys.authentication_challenge_expiration = passkey_config.authentication_challenge_expiration
       ActionPack::Passkeys.challenge_url = passkey_config.challenge_url
+      ActionPack::Passkeys.related_origins = passkey_config.related_origins
     end
 
     initializer "action_pack.passkey.verifier" do
       config.after_initialize do |app|
         ActionPack::WebAuthn.challenge_verifier = app.message_verifier("action_pack.webauthn.challenge")
         ActionPack::WebAuthn.application_name = app.name
+        ActionPack::WebAuthn.relying_party_id = config.action_pack.passkey.relying_party_id
       end
     end
 
@@ -51,6 +55,10 @@ module ActionPack
         if passkey_config.draw_routes
           scope passkey_config.routes_prefix, as: :passkey do
             post "/challenge" => "action_pack/passkeys/challenges#create", as: :challenge
+          end
+
+          if passkey_config.related_origins.any?
+            get "/.well-known/webauthn" => "action_pack/well_known/webauthn#show"
           end
         end
       end
--- a/actionpack/lib/action_pack/passkeys/engine.rb
+++ b/actionpack/lib/action_pack/passkeys/engine.rb
@@ -23,8 +23,6 @@ module ActionPack
     config.action_pack.passkey.related_origins = []
     config.action_pack.passkey.default_registration_options = {}
     config.action_pack.passkey.default_authentication_options = {}
-    config.action_pack.passkey.registration_timeout = 10.minutes
-    config.action_pack.passkey.authentication_timeout = 5.minutes
     config.action_pack.passkey.cbor_max_depth = 16
     config.action_pack.passkey.cbor_max_size = 10.megabytes
 
@@ -34,8 +32,6 @@ module ActionPack
       ActionPack::Passkeys.parent_class_name = passkey_config.parent_class_name
       ActionPack::Passkeys.default_registration_options = passkey_config.default_registration_options
       ActionPack::Passkeys.default_authentication_options = passkey_config.default_authentication_options
-      ActionPack::Passkeys.registration_timeout = passkey_config.registration_timeout
-      ActionPack::Passkeys.authentication_timeout = passkey_config.authentication_timeout
       ActionPack::Passkeys.challenge_url = passkey_config.challenge_url
       ActionPack::Passkeys.related_origins = passkey_config.related_origins
     end
--- a/actionpack/lib/action_pack/passkeys/form_helper.rb
+++ b/actionpack/lib/action_pack/passkeys/form_helper.rb
@@ -1,0 +1,128 @@
+# frozen_string_literal: true
+
+module ActionPack
+  # = Action Pack Passkey Form Helper
+  #
+  # View helpers for rendering passkey registration and sign-in buttons.
+  module Passkeys::FormHelper
+    # Renders a button for registering a new passkey. Accepts a +label+ string or a block
+    # for button content.
+    #
+    # Options:
+    # - +options+ - WebAuthn creation options (JSON-serializable hash).
+    # - +challenge_url+ - Endpoint to refresh the challenge nonce.
+    # - +wrapper+ - HTML attributes for the outer web component element.
+    # - +form+ - Additional HTML attributes for the +<form>+ tag. Supports a +:param+ key
+    #   to set the form parameter namespace (default: +:passkey+).
+    # - +error+ - HTML attributes for the error message +<div>+. Supports a +:message+ key
+    #   to override the default error text.
+    # - +cancellation+ - HTML attributes for the cancellation message +<div>+. Supports a
+    #   +:message+ key to override the default cancellation text.
+    # - All other options are passed to the +<button>+ tag.
+    def passkey_registration_button(name = nil, url = nil, **options, &block)
+      url, name = name, block ? capture(&block) : nil if block_given?
+      component_options, form_options, button_options, error_options = partition_passkey_options(url, options)
+      error_options[:error][:message] ||= I18n.t("helpers.passkeys.registration.error", default: "Something went wrong while registering your passkey.")
+      error_options[:cancellation][:message] ||= I18n.t("helpers.passkeys.registration.cancelled", default: "Passkey registration was cancelled. Try again when you are ready.")
+      error_options[:duplicate][:message] ||= I18n.t("helpers.passkeys.registration.duplicate", default: "You already have a passkey registered on this device. Remove the existing one first and try again.")
+      param = form_options.delete(:param)
+
+      content_tag("rails-passkey-registration-button", **component_options.transform_keys { |key| key.to_s.dasherize }) do
+        tag.form(**form_options) do
+          hidden_field_tag(:authenticity_token, form_authenticity_token) +
+            hidden_field_tag("#{param}[client_data_json]", nil, id: nil, data: { passkey_field: "client_data_json" }) +
+            hidden_field_tag("#{param}[attestation_object]", nil, id: nil, data: { passkey_field: "attestation_object" }) +
+            hidden_field_tag("#{param}[transports][]", nil, id: nil, data: { passkey_field: "transports" }) +
+            tag.button(name, type: :button, data: { passkey: "register" }, **button_options)
+        end + passkey_error_messages(**error_options)
+      end
+    end
+
+    # Renders a button for signing in with a passkey. Accepts a +label+ string or a block
+    # for button content.
+    #
+    # Options:
+    # - +options+ - WebAuthn request options (JSON-serializable hash).
+    # - +challenge_url+ - Endpoint to refresh the challenge nonce.
+    # - +mediation+ - WebAuthn mediation hint (e.g. +"conditional"+ for autofill-assisted sign in).
+    # - +wrapper+ - HTML attributes for the outer web component element.
+    # - +form+ - Additional HTML attributes for the +<form>+ tag. Supports a +:param+ key
+    #   to set the form parameter namespace (default: +:passkey+).
+    # - +error+ - HTML attributes for the error message +<div>+. Supports a +:message+ key
+    #   to override the default error text.
+    # - +cancellation+ - HTML attributes for the cancellation message +<div>+. Supports a
+    #   +:message+ key to override the default cancellation text.
+    # - All other options are passed to the +<button>+ tag.
+    def passkey_sign_in_button(name = nil, url = nil, **options, &block)
+      url, name = name, block ? capture(&block) : nil if block_given?
+      component_options, form_options, button_options, error_options = partition_passkey_options(url, options)
+      error_options[:error][:message] ||= I18n.t("helpers.passkeys.sign_in.error", default: "Something went wrong while signing in with your passkey.")
+      error_options[:cancellation][:message] ||= I18n.t("helpers.passkeys.sign_in.cancelled", default: "Passkey sign in was cancelled. Try again when you are ready.")
+      param = form_options.delete(:param)
+
+      content_tag("rails-passkey-sign-in-button", **component_options.transform_keys { |key| key.to_s.dasherize }) do
+        tag.form(**form_options) do
+          hidden_field_tag(:authenticity_token, form_authenticity_token) +
+            hidden_field_tag("#{param}[id]", nil, id: nil, data: { passkey_field: "id" }) +
+            hidden_field_tag("#{param}[client_data_json]", nil, id: nil, data: { passkey_field: "client_data_json" }) +
+            hidden_field_tag("#{param}[authenticator_data]", nil, id: nil, data: { passkey_field: "authenticator_data" }) +
+            hidden_field_tag("#{param}[signature]", nil, id: nil, data: { passkey_field: "signature" }) +
+            tag.button(name, type: :button, data: { passkey: "sign_in" }, **button_options)
+        end + passkey_error_messages(**error_options)
+      end
+    end
+
+    private
+      def partition_passkey_options(url, options)
+        passkey_options = options.fetch(:options, {})
+        wrapper_options = options.fetch(:wrapper, {})
+
+        component_options = options
+          .slice(:challenge_url, :mediation)
+          .reverse_merge(challenge_url: default_passkey_challenge_url, options: passkey_options.to_json(except: :challenge))
+
+        form_options = options
+          .fetch(:form, {})
+          .reverse_merge(method: :post, action: url, class: "button_to", param: :passkey)
+
+        error_options = options.slice(:error, :cancellation, :duplicate).reverse_merge(error: {}, cancellation: {}, duplicate: {})
+
+        button_options = options.except(:options, :form, :wrapper, *component_options.keys, *error_options.keys)
+
+        [ wrapper_options.merge(component_options), form_options, button_options, error_options ]
+      end
+
+      def default_passkey_challenge_url
+        if challenge_url = ActionPack::Passkeys.challenge_url
+          instance_exec(&challenge_url)
+        else
+          passkey_challenge_path
+        end
+      end
+
+      def passkey_error_messages(error: {}, cancellation: {}, duplicate: {})
+        error_message, error_attributes = build_passkey_error_options("error", error)
+        cancellation_message, cancellation_attributes = build_passkey_error_options("cancelled", cancellation)
+
+        messages = tag.div(error_message, hidden: true, **error_attributes) +
+          tag.div(cancellation_message, hidden: true, **cancellation_attributes)
+
+        if duplicate[:message]
+          duplicate_message, duplicate_attributes = build_passkey_error_options("duplicate", duplicate)
+          messages += tag.div(duplicate_message, hidden: true, **duplicate_attributes)
+        end
+
+        messages
+      end
+
+      def build_passkey_error_options(type, options)
+        message = options[:message]
+
+        attributes = options.except(:message)
+        attributes[:data] ||= {}
+        attributes[:data][:passkey_error] = type
+
+        [ message, attributes ]
+      end
+  end
+end
--- a/actionpack/lib/action_pack/passkeys/holder.rb
+++ b/actionpack/lib/action_pack/passkeys/holder.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+
+module ActionPack
+  # = Action Pack Passkey Holder
+  #
+  # Adds passkey support to an Active Record model (the "holder" of passkeys).
+  #
+  # == Usage
+  #
+  #   class User < ApplicationRecord
+  #     has_passkeys name: :email_address, display_name: :name
+  #   end
+  #
+  # This sets up a polymorphic +has_many :passkeys+ association and defines two methods on the
+  # model that supply holder-specific options for the WebAuthn ceremonies:
+  #
+  # * +passkey_registration_options+ — merged into ActionPack::Passkeys::Passkey.registration_options
+  # * +passkey_authentication_options+ — merged into ActionPack::Passkeys::Passkey.authentication_options
+  #
+  # == Options
+  #
+  # +has_passkeys+ accepts keyword arguments that map to WebAuthn creation or request option
+  # fields. Values can be symbols (sent to the record), procs (evaluated in the record's context),
+  # or plain values:
+  #
+  # [+name+]
+  #   A human-readable account identifier (typically an email or username) shown by the
+  #   authenticator when the user selects a passkey. Maps to the WebAuthn +user.name+ field.
+  #
+  # [+display_name+]
+  #   A friendly label for the user (typically their full name) shown by the authenticator
+  #   during passkey registration. Maps to the WebAuthn +user.displayName+ field.
+  #
+  #   has_passkeys name: :email, display_name: :name
+  #
+  # For more complex configuration, pass a block that receives an ActionPack::Passkeys::Passkeys::Holder::Config:
+  #
+  #   has_passkeys do |config|
+  #     config.registration_options { { name: email, display_name: name } }
+  #     config.authentication_options  { { user_verification: "required" } }
+  #   end
+  module Passkeys::Holder
+    extend ActiveSupport::Concern
+
+    class_methods do
+      # Declares that this model can hold passkeys. Sets up a polymorphic +has_many+ association
+      # and defines +passkey_registration_options+ and +passkey_authentication_options+ instance methods used
+      # by ActionPack::Passkeys::Passkey to build ceremony options.
+      #
+      # Keyword arguments matching CreationOptions or RequestOptions fields are extracted and
+      # turned into holder-scoped option procs automatically. An optional block yields a Config
+      # for more complex setup.
+      def has_passkeys(**options, &block)
+        config = Config.new(**options)
+        block&.call(config)
+
+        has_many config.association_name,
+          as: :holder,
+          dependent: config.dependent,
+          class_name: "ActionPack::Passkeys::Passkey"
+
+        define_method(:passkey_registration_options) do
+          {
+            id: id,
+            exclude_credentials: public_send(config.association_name)
+          }.merge(config.evaluate_registration_options(self))
+        end
+
+        define_method(:passkey_authentication_options) do
+          { credentials: public_send(config.association_name) }.merge(config.evaluate_authentication_options(self))
+        end
+      end
+    end
+
+    # Configuration object yielded by +has_passkeys+ when a block is given. Allows setting
+    # custom association options and ceremony option blocks.
+    class Config
+      attr_accessor :association_name, :dependent
+
+      def initialize(**options) # :nodoc:
+        @association_name = options.delete(:association_name) || :passkeys
+        @dependent = options.delete(:dependent) || :destroy
+
+        if creation_opts = extract_options_for(ActionPack::WebAuthn::PublicKeyCredential::CreationOptions, options)
+          @registration_options = options_to_proc(creation_opts)
+        end
+
+        if request_opts = extract_options_for(ActionPack::WebAuthn::PublicKeyCredential::RequestOptions, options)
+          @authentication_options = options_to_proc(request_opts)
+        end
+      end
+
+      # Sets a block to evaluate in the holder's context to produce additional authentication options.
+      #
+      #   config.authentication_options { { user_verification: "required" } }
+      def authentication_options(&block)
+        @authentication_options = block
+      end
+
+      # Sets a block to evaluate in the holder's context to produce additional registration options.
+      #
+      #   config.registration_options { { name: email, display_name: name } }
+      def registration_options(&block)
+        @registration_options = block
+      end
+
+      def evaluate_authentication_options(record) # :nodoc:
+        if @authentication_options
+          record.instance_exec(&@authentication_options)
+        else
+          {}
+        end
+      end
+
+      def evaluate_registration_options(record) # :nodoc:
+        if @registration_options
+          record.instance_exec(&@registration_options)
+        else
+          {}
+        end
+      end
+
+      private
+        def extract_options_for(klass, options)
+          keys = klass.attribute_names.map(&:to_sym)
+
+          extracted = options.slice(*keys)
+          options.except!(*keys)
+          extracted if extracted.any?
+        end
+
+        def options_to_proc(options)
+          proc do
+            options.transform_values do |value|
+              case value
+              when Symbol then send(value)
+              when Proc then instance_exec(&value)
+              else value
+              end
+            end
+          end
+        end
+    end
+  end
+end
--- a/actionpack/lib/action_pack/web_authn/authenticator/attestation_response.rb
+++ b/actionpack/lib/action_pack/web_authn/authenticator/attestation_response.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+module ActionPack
+  module WebAuthn
+    module Authenticator
+      # = Action Pack WebAuthn Attestation Response
+      #
+      # Handles the authenticator response from a WebAuthn registration ceremony.
+      # When a user registers a new credential, the authenticator returns an
+      # attestation response containing the new public key and credential ID.
+      #
+      # == Usage
+      #
+      #   response = ActionPack::WebAuthn::Authenticator::AttestationResponse.new(
+      #     client_data_json: params[:response][:clientDataJSON],
+      #     attestation_object: params[:response][:attestationObject],
+      #     origin: "https://example.com"
+      #   )
+      #
+      #   response.validate!
+      #
+      #   # Store the credential
+      #   credential_id = response.attestation.credential_id
+      #   public_key = response.attestation.public_key
+      #
+      # == Validation
+      #
+      # In addition to the base Response validations, this class verifies:
+      #
+      # * The client data type is "webauthn.create"
+      # * The attestation format has a registered verifier
+      # * The attestation statement passes format-specific verification
+      #
+      class AttestationResponse < Response
+        attr_reader :attestation_object
+
+        validate :client_data_type_must_be_create
+        validate :attestation_must_be_valid
+
+        def initialize(attestation_object:, **attributes) # :nodoc:
+          super(**attributes)
+          @attestation_object = attestation_object
+        end
+
+        # Returns the decoded Attestation object, lazily parsed from the raw
+        # attestation object bytes.
+        def attestation
+          @attestation ||= ActionPack::WebAuthn::Authenticator::Attestation.wrap(attestation_object)
+        end
+
+        # Returns the authenticator data extracted from the attestation object.
+        def authenticator_data
+          attestation.authenticator_data
+        end
+
+        private
+          def challenge_purpose
+            "registration"
+          end
+
+          def client_data_type_must_be_create
+            unless client_data["type"] == "webauthn.create"
+              errors.add(:base, "Client data type is not webauthn.create")
+            end
+          end
+
+          def attestation_must_be_valid
+            verifier = ActionPack::WebAuthn.attestation_verifiers[attestation.format]
+
+            if verifier
+              verifier.verify!(attestation, client_data_json: client_data_json)
+            else
+              errors.add(:base, "Unsupported attestation format: #{attestation.format}")
+            end
+          end
+      end
+    end
+  end
+end
--- a/actionpack/lib/action_pack/web_authn/authenticator/response.rb
+++ b/actionpack/lib/action_pack/web_authn/authenticator/response.rb
@@ -1,0 +1,149 @@
+# frozen_string_literal: true
+
+module ActionPack
+  module WebAuthn
+    module Authenticator
+      # = Action Pack WebAuthn Authenticator Response
+      #
+      # Abstract base class for WebAuthn authenticator responses. Provides common
+      # validation logic for both registration (attestation) and authentication
+      # (assertion) ceremonies.
+      #
+      # This class should not be instantiated directly. Use AttestationResponse for
+      # registration or AssertionResponse for authentication.
+      #
+      # == Validation
+      #
+      # The +validate!+ method performs security checks required by the WebAuthn
+      # specification:
+      #
+      # * Challenge verification - ensures the response contains a valid, signed challenge
+      # * Origin verification - ensures the response comes from the expected origin
+      # * User verification - optionally requires biometric or PIN verification
+      #
+      # == Example
+      #
+      #   response = ActionPack::WebAuthn::Authenticator::AssertionResponse.new(
+      #     client_data_json: client_data_json,
+      #     authenticator_data: authenticator_data,
+      #     signature: signature,
+      #     credential: credential,
+      #     origin: "https://example.com",
+      #     user_verification: :required
+      #   )
+      #
+      #   response.validate!
+      #
+      class Response
+        include ActiveModel::Validations
+
+        attr_reader :client_data_json
+        attr_accessor :origin, :user_verification
+
+        validate :challenge_must_be_present
+        validate :challenge_must_not_be_expired
+        validate :origin_must_match
+        validate :must_not_be_cross_origin
+        validate :must_not_have_token_binding
+        validate :relying_party_id_must_match
+        validate :user_must_be_present
+        validate :user_must_be_verified_when_required
+
+        def initialize(client_data_json:, origin: nil, user_verification: :preferred) # :nodoc:
+          @client_data_json = client_data_json
+          @origin = origin
+          @user_verification = user_verification.to_sym
+        end
+
+        def validate!
+          super
+        rescue ActiveModel::ValidationError
+          raise ActionPack::WebAuthn::InvalidResponseError, errors.full_messages.join(", ")
+        end
+
+        # Returns the RelyingParty used for RP ID validation.
+        def relying_party
+          ActionPack::WebAuthn.relying_party
+        end
+
+        # Parses the client data JSON string into a Hash. Raises
+        # +InvalidResponseError+ if the JSON is malformed.
+        def client_data
+          @client_data ||= JSON.parse(client_data_json)
+        rescue JSON::ParserError
+          raise ActionPack::WebAuthn::InvalidResponseError, "Client data is not valid JSON"
+        end
+
+        def authenticator_data # :nodoc:
+          nil
+        end
+
+        private
+          def challenge_must_be_present
+            if client_data["challenge"].blank?
+              errors.add(:base, "Challenge missing")
+            end
+          end
+
+          def challenge_must_not_be_expired
+            return if errors.any?
+
+            signed_message = Base64.urlsafe_decode64(client_data["challenge"])
+
+            unless ActionPack::WebAuthn.challenge_verifier.verified(signed_message, purpose: challenge_purpose)
+              errors.add(:base, "Challenge has expired")
+            end
+          rescue ArgumentError
+            errors.add(:base, "Challenge is invalid")
+          end
+
+          def challenge_purpose
+            nil
+          end
+
+          def origin_must_match
+            if origin.blank?
+              errors.add(:base, "Origin missing")
+            elsif client_data["origin"].blank?
+              errors.add(:base, "Origin missing in client data")
+            elsif !ActiveSupport::SecurityUtils.secure_compare(origin.to_s, client_data["origin"].to_s)
+              errors.add(:base, "Origin does not match")
+            end
+          end
+
+          def must_not_be_cross_origin
+            if client_data["crossOrigin"] == true
+              errors.add(:base, "Cross-origin requests are not supported")
+            end
+          end
+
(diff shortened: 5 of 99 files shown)


MatheusRich:
cc @dhh I think you've implemented this at some point for Fizzy or one of the ONCE apps.

frederikspang:
Quick 2 cents, I'm torn between naming. `has_passkeys` like has_secure_password - Or making it match active storage `has_many_attached`, as in `has_many_passkeys`. 

Not necessarily anything to do anything about- just a little observation.

Massive +1 for having this as first class support though.

actionpack/lib/action_pack/passkeys/form_helper.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+module ActionPack
+  # = Action Pack Passkey Form Helper
+  #
+  # View helpers for rendering passkey registration and sign-in buttons.
+  module Passkeys::FormHelper
+    REGISTRATION_ERROR_MESSAGE = "Something went wrong while registering your passkey."
```
djmb: Should these be specified via i18n locale files?

actionpack/lib/passkeys/app/models/action_pack/passkey.rb
```diff
@@ -0,0 +1,111 @@
...
+  #   options = ActionPack::Passkey.authentication_options
+  #   # Pass options to the browser
+  #
+  #   passkey = ActionPack::Passkey.authenticate(params[:passkey])
+  #
+  # == Holder Integration
+  #
+  # Call +has_passkeys+ in the model to set up the association and configure ceremony options
+  # per-holder. See ActionPack::Passkeys::Holder for details.
+  class Passkey < ActionPack::Passkeys.parent_class_name.constantize
```
djmb: Is there a specific reason to make `ActionPack::Passkey`s superclass configurable when other built-in Rails models are not? 

It is useful to be able to define the superclass, but maybe we shouldn't be providing a piecemeal solution just here and instead separately look at making them all configurable?

actionpack/lib/action_pack/passkey.rb
```diff
@@ -0,0 +1,111 @@
...
+  #   # Pass options to the browser
+  #
+  #   passkey = ActionPack::Passkey.authenticate(params[:passkey])
+  #
+  # == Holder Integration
+  #
+  # Call +has_passkeys+ in the model to set up the association and configure ceremony options
+  # per-holder. See ActionPack::Passkeys::Holder for details.
+  class Passkey < ActionPack::Passkeys.parent_class_name.constantize
+    self.table_name = "action_pack_passkeys"
```
djmb: Is this needed?

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
```diff
@@ -0,0 +1,20 @@
...
+class CreateActionPackPasskeys < ActiveRecord::Migration[8.2]
+  def change
+    create_table :action_pack_passkeys do |t|
+      t.belongs_to :holder, polymorphic: true, null: false
+      t.string :credential_id, null: false
+      t.binary :public_key, null: false
+      t.integer :sign_count, null: false, default: 0
+      t.string :name
+      t.text :transports
+      t.string :aaguid
```
djmb: As this is a guid, we could limit it to 36 characters

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class CreateActionPackPasskeys < ActiveRecord::Migration[8.2]
+  def change
+    create_table :action_pack_passkeys do |t|
```
djmb: I wonder if we should store the `rp_id`? There's no immediate use for it here but with any kind of domain migration for an app or for apps that run under multiple domains it could be very useful to know.

activesupport/lib/active_support/base32.rb
```diff
@@ -0,0 +1,55 @@
...
+  # Generates and normalizes Crockford's Base32 strings. This encoding is designed to be
+  # unambiguous for humans, excluding the characters I, L, O, and U from its alphabet.
+  #
+  # When normalizing user input, visually ambiguous characters are substituted
+  # (O → 0, I → 1, L → 1) and invalid characters are stripped, so that codes
+  # are forgiving to transcription errors.
+  #
+  #   ActiveSupport::Base32.generate(6)    # => "PAK1NG"
+  #   ActiveSupport::Base32.normalize("OIL-123") # => "011123"
+  module Base32
```
djmb, line 18: Is Crockford's the canonical Base32 implementation? RFC 4648 defines two other schemes it calls base32 and base32hex.

I think we maybe need separate modules for each format, or maybe provide the scheme as an argument? Different modules might be better since normalizing seems to be a concept that only applies to the Crockford's format.

actionpack/lib/passkeys/app/models/action_pack/passkey.rb
monorkin: No specific reason. I saw that there are [a few other PRs](https://github.com/rails/rails/pull/54811) that add the ability to configure the parent class of other built-in models so I added this preemptively.

I can change it to inherit from ActiveRecord::Base instead.

activesupport/lib/active_support/base32.rb
monorkin: Moved it to a separate module in [a73db61](https://github.com/rails/rails/pull/57100/commits/a73db61845652f5fbb6c32edf80a3538b61d1edd)
Initially I put it at the root since SecureRandom.base32 is a Crockford base32 generator, it might be worth revisiting that one in a separate PR.

actionpack/lib/action_pack/passkey.rb
monorkin: No, removed it in [239ed61](https://github.com/rails/rails/pull/57100/commits/239ed61c3b8305e6040a8fe1c3afaa96d2ae68c5)

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
monorkin: Added the limit in [bc4c607](https://github.com/rails/rails/pull/57100/commits/bc4c60741b9403bf329f8742600d108980922525)

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
monorkin: Could you expand on what you have in mind?

The way I see it, for domain changes, we have to use [Related Origins Requests](https://passkeys.dev/docs/advanced/related-origins/) (ROR) and allows Passkeys from both domains to work.

What I think I should do is add a config for a list of allowed RP IDs, that way ROR would be trivial to implement.

(Context: A Passkey is bound in the authenticator to an Relying Party ID, which this PR sets to the request's origin, so changing the domain would break all existing passkeys, this is intentional and part of what makes passkeys good alternative to passwords)

actionpack/lib/action_pack/passkeys/form_helper.rb
monorkin: They should, moved them in [ac3888b](https://github.com/rails/rails/pull/57100/commits/ac3888b98e2c7b9f3269df2db42f9e467105bb2c). Followed the same pattern for the key as the other form helpers - e.g. `helpers.passkeys.registration.error`

actionpack/lib/action_pack/passkey.rb
monorkin: It was needed, I forgot the reason, without it the table name resolves to just `passkeys`, and there is no module to define the table name prefix so this looked like the cleaner solution - explicitly set the table name instead of adding an empty action pack module that just sets the prefix.

If `Passkey` was within the engine's namespace this wouldn't be needed. But moving it there results in an awkward module name - `ActionPack::Passkeys::Passkey`

Do you know if there is a convention?

monorkin:
@djmb and I discussed a few things via pings, here's a summary of all the changes based on his feedback:
- Moved Crockford's base 32 to separate module
    Extracts Crockford's Base32 encoding into ActiveSupport::Base32::Crockford because there are three different base 32 implementations with different properties
- Limit AAGUIDs to 36 characters
    Validates that the AAGUID is a valid UUID
- Move error messages for form helpers to I18n
    This is in line with all other form helpers
- Move Passkey into the engine
    Moves the Passkey model from lib/action_pack/passkey.rb into the engine's app/models/ directory which is the expected location for an engine, this also allows us to drop an explicit table name or prefix, but changes the name of the model.
- Isolate the namespace of the engine
    Aligns the engine with the other Rails engines
- Rename JS to actionpack-passkeys
    Separates ActionPack in general from passkeys

The test failure seems to be unrelated

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
```diff
@@ -0,0 +1,115 @@
...
+            ES256,
+            EDDSA,
+            RS256
+          ],
+          authenticatorSelection: {
+            residentKey: resident_key.to_s,
+            requireResidentKey: resident_key == :required,
+            userVerification: user_verification.to_s
+          }
+        }
```
santiagorodriguez96, line 94: The [`authenticatorattachment`](https://www.w3.org/TR/webauthn-2/#dom-authenticatorselectioncriteria-authenticatorattachment) and [`timeout`](https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialcreationoptions-timeout) attributes are missing – is that intentional?

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
```diff
@@ -0,0 +1,115 @@
...
+
+        if exclude_credentials.any?
+          json[:excludeCredentials] = exclude_credentials.map { |credential| exclude_credential_json(credential) }
+        end
+
+        if attestation != :none
+          json[:attestation] = attestation.to_s
+        end
```
santiagorodriguez96, line 102: I'm getting an error when running the conformance specs against the implementation given that we are not setting `attestation` to `none`:

```
AssertionError: Client requestsed that server would set attestation to "none". Server has not returned attestation set to "none"!: expected undefined to equal 'none'
    at eval (eval at compileCode (js/sandbox.js:25:26), <anonymous>:4574:24)
```

The [specification states that](https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialcreationoptions-attestation):

> Its value SHOULD be a member of [AttestationConveyancePreference](https://www.w3.org/TR/webauthn-2/#enumdef-attestationconveyancepreference).

To be honest, I don't really know which implications this might have but I guess you have everything you need to comply with the spec on this one

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
santiagorodriguez96, line 94: Should the implementation support users to set the [`extensions` attribute](https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialcreationoptions-extensions)?

santiagorodriguez96, line 94: Heads up, there're two additional attributes that are defined in level 3 of the specification: [`hints`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-hints) and [`attestationFormats`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-attestationformats).

The first one is already supported by Chrome but doesn't do anything for me in Firefox and Safari (don't know about other browsers though) and the latter doesn't seems to do anything for me in Chrome (don't know about other browsers).

santiagorodriguez96: Worth noting that one of the conformance specs is failing because of the `extension` attribute is not currently supported:

```
AssertionError: Response.extensions MUST be set to the requested extensions! Expected "undefined" to equal "{"example.extension.bool":true}": expected undefined to deeply equal { 'example.extension.bool': true }
    at n.eval (eval at compileCode (js/sandbox.js:25:26), <anonymous>:4562:16)
```

Image

actionpack/lib/action_pack/web_authn/public_key_credential.rb
```diff
@@ -0,0 +1,165 @@
...
+          )
+
+          response.validate!
+
+          new(
+            id: response.attestation.credential_id,
+            public_key: response.attestation.public_key,
+            sign_count: response.attestation.sign_count,
+            aaguid: response.attestation.aaguid,
+            backed_up: response.attestation.backed_up?,
```
santiagorodriguez96, line 106: Should it also store the [backup eligibility](https://www.w3.org/TR/webauthn-3/#authdata-flags-be) of the credential?

railties/lib/rails/generators/rails/authentication/templates/app/controllers/my/passkeys_controller.rb.tt
```diff
@@ -0,0 +1,34 @@
...
+  def index
+    @passkeys = Current.user.passkeys.order(name: :asc, created_at: :desc)
+    @registration_options = passkey_registration_options(holder: Current.user)
+  end
+
+  def create
+    passkey = Current.user.passkeys.register(passkey_registration_params)
+
+    redirect_to edit_my_passkey_path(passkey, created: true)
+  end
```
santiagorodriguez96, line 15: `ActionPack::Passkey.register` will raise an error if something fails:

https://github.com/monorkin/rails/blob/b638493a9af85f91ac5817f8343cec7bba2d1793/actionpack/lib/passkeys/app/models/action_pack/passkeys/passkey.rb#L54-L65

Is it okay for the request to return a server error in that case?

actionpack/lib/action_pack/passkeys/engine.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require "active_record/railtie"
```
santiagorodriguez96, line 3: Small gotcha: this line requires `active_record/railtie` unconditionally. So if you created your Rails app with `--minimal`, `activerecord` will be commented out in `config/application.rb` but if you somehow require `action_pack/passkeys/engine`, then `activerecord` railtie will load anyway which could be kinda unexpected (and will even fail for `--minimal` apps that don't have a `database.yml`, for instance).

I think this is expected because the `passkeys` engine depends on `activerecord`, but just wanted to comment it out 🙂

railties/lib/rails/all.rb
```diff
@@ -12,6 +12,7 @@
   action_cable/engine
   action_mailbox/engine
   action_text/engine
+  action_pack/passkeys/engine
```
santiagorodriguez96, line 15: The engine is only auto-required from `rails/all`, so apps generated with `--minimal` won't pick it up. Obviously this is expected, but the engine won't even appear commented out in the generated `config/application.rb`.

A user with a minimal app (or simply that doesn't require Rails' modules using `rails/all`) currently has no signal that they need to require it – docs don't mention that  requiring `action_pack/passkeys/engine` is needed for this whole feature to work. Perhaps is it worth mention this somewhere?

actionpack/lib/action_pack/web_authn/public_key_credential.rb
```diff
@@ -0,0 +1,165 @@
...
+        #
+        # Raises +InvalidResponseError+ if the attestation is invalid.
+        def register(params, origin: ActionPack::WebAuthn::Current.origin)
+          response = ActionPack::WebAuthn::Authenticator::AttestationResponse.new(
+            client_data_json: params[:client_data_json],
+            attestation_object: params[:attestation_object],
+            origin: origin
+          )
+
+          response.validate!
```
santiagorodriguez96, line 99: There's no way currently for users to indicate if they want to verify that `user_verification` flag was set to `true`.

`ActionPack::WebAuthn::Authenticator::Response` supports a `:user_verification` kwarg and only verifies that the UV flag is present when the `user_verification` param was set to `:required`:

https://github.com/monorkin/rails/blob/b638493a9af85f91ac5817f8343cec7bba2d1793/actionpack/lib/action_pack/web_authn/authenticator/response.rb#L141-L145

However, as the `register` method is not passing the `user_verification` when instantiating the attestation response, so the new instance will have the default `preferred` as the value of `user_verification`:

https://github.com/monorkin/rails/blob/b638493a9af85f91ac5817f8343cec7bba2d1793/actionpack/lib/action_pack/web_authn/authenticator/response.rb#L52

so the `UV` flag will never be verified.

The same happens with the `authenticate` method 👀

santiagorodriguez96 (commented):
Took this for a spin against the [`FIDO Alliance Conformance Test Tool`](https://fidoalliance.org/certification/functional-certification/conformance/) – spun up a [minimal Rails API app](https://github.com/cedarcode/rails-passkeys-conformance-test) that wraps the native WebAuthn implementation with the 4 conformance endpoints (`/attestation/options`, `/attestation/result`, `/assertion/options`, `/assertion/result`) and ran it against the test suite.

Results: most tests failed because the implementation only supports `none` attestation – the conformance suite exercises `packed`, `tpm` and `fido-u2f` none of which are currently wired up. Left a screenshot below with the summary of the run:

<details>
  <summary>See screenshot</summary>
  
  Image

</details>

Beyond the attestation coverage gap, two other tests failed for unrelated reasons – I've left inline comments on both.

I haven't gotten through all the files yet, but have left a few comments on what I've reviewed so far. Will continue going through the rest.

Happy to help however is useful.

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
djmb: The `rp_id` is known at creation time and won't change. It's going to be valuable information for debugging login issues and also useful if you ever decide to switch the app domain or if you support multiple domains - e.g. maybe you only want to display the passkeys for the current domain in the UI.

Its going to be much harder to add later, and I don't think there's any downside to having it now.

monorkin: Agreed, I'll add it in.

monorkin:
@santiagorodriguez96 thank you for the detailed review! This is immensely helpful, I'll start implementing changes next week.

Re attestations:
I intentionally left out everything but "none" and added a way to plug-in other attestation modes if someone needs them. From my understanding and testing, most authenticators should work with "none" they just wont provide some information (e.g. Yubikeys won't return their AAGUID).

```ruby
class PackedAttestationVerifier
  def verify!(attestation, client_data_json:)
    # ...
  end
end

ActionPack::WebAuthn.register_attestation_verifier("packed", PackedAttestationVerifier.new)
```

kardiojack:
Any updates on when this will be merged or areas someone can pitch in?

railties/lib/rails/generators/rails/authentication/authentication_generator.rb
```diff
@@ -38,22 +58,72 @@ def configure_application_controller
...
+        if options.password_based?
+          if File.read(File.expand_path("Gemfile", destination_root)).include?('gem "bcrypt"')
+            uncomment_lines "Gemfile", /gem "bcrypt"/
+            bundle_command("install --quiet")
+          else
+            bundle_command("add bcrypt", {}, quiet: true)
+          end
         else
-          bundle_command("add bcrypt", {}, quiet: true)
+          bundle_command("add letter_opener --group development", {}, quiet: true) if defined?(ActionMailer::Railtie)
```
brunoprietog: I don't think it's a good idea to add `letter_opener` here. It can be quite intrusive, and besides, it doesn't work by default in WSL, you have to make a few changes to get it to work. In that case, if this is really necessary, I would use [`mailbin`](https://github.com/excid3/mailbin) instead, which works out of the box on any platform and doesn't disrupt any workflow.

pglombardo:
As a quiet observer of this PR, I'm hoping this PR progresses.  It would be great to offer a modern auth option.

Edgarruiz8585:
*
***@***.***

Obtener Outlook para Android<https://aka.ms/AAb9ysg>

________________________________
De: Peter Giacomo Lombardo ***@***.***>
Enviado: miércoles, junio 24, 2026 4:45:29 a. m.
Para: rails/rails ***@***.***>
Cc: Subscribed ***@***.***>
Asunto: Re: [rails/rails] Passkeys (PR #57100)

[https://avatars.githubusercontent.com/u/395132?s=20&v=4]pglombardo left a comment (rails/rails#57100)<https://github.com/rails/rails/pull/57100#issuecomment-4788389857>

As a quiet observer of this PR, I'm hoping this PR progresses. It would be great to offer a modern auth option.

  *   ***@***.*** by Edgarruiz8585

Edgarruiz8585:
*
***@***.***

Obtener Outlook para Android<https://aka.ms/AAb9ysg>

________________________________
De: Peter Giacomo Lombardo ***@***.***>
Enviado: miércoles, junio 24, 2026 4:45:29 a. m.
Para: rails/rails ***@***.***>
Cc: Subscribed ***@***.***>
Asunto: Re: [rails/rails] Passkeys (PR #57100)

[https://avatars.githubusercontent.com/u/395132?s=20&v=4]pglombardo left a comment (rails/rails#57100)<https://github.com/rails/rails/pull/57100#issuecomment-4788389857>

As a quiet observer of this PR, I'm hoping this PR progresses. It would be great to offer a modern auth option.

  *   ***@***.***

monorkin:
Just pushed a few changes that should address most of the comments here.
I still have to replace letter opener with mailbin, and add user verification validation.

Edgarruiz8585:
Obtener Outlook para Android<https://aka.ms/AAb9ysg>

________________________________
De: Stanko Krtalić ***@***.***>
Enviado: martes, junio 30, 2026 11:45:21 a. m.
Para: rails/rails ***@***.***>
Cc: Edgar Manuel Ruiz Arias 💜🇲🇽 e IO ***@***.***>; Comment ***@***.***>
Asunto: Re: [rails/rails] Passkeys (PR #57100)

[https://avatars.githubusercontent.com/u/1655218?s=20&v=4]monorkin left a comment (rails/rails#57100)<https://github.com/rails/rails/pull/57100#issuecomment-4846353525>

Just pushed a few changes that should address most of the comments here.
I still have to replace letter opener with mailbin, and add user verification validation.

  *   ***@***.*** ok activada

Edgarruiz8585:
*
***@***.***

________________________________

railties/lib/rails/generators/rails/authentication/authentication_generator.rb
monorkin: Hey @brunoprietog I didn't know about mailbin until you mentioned it, it looks great! I swapped out the gems in https://github.com/rails/rails/pull/57100/commits/6f91b1ca517f83f239654611d2bedefa01af8d5f

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
monorkin: It isn't. I added those in https://github.com/rails/rails/pull/57100/commits/d90efdb718344d95560d93a842f8d34f0eca52eb

Now the challenge expiry is governed by the timeout attribute.

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
monorkin: Changed it in https://github.com/rails/rails/pull/57100/commits/c7c1b483c62c2670d5ee9773023ece193a19fdf8

railties/lib/rails/generators/rails/authentication/authentication_generator.rb
brunoprietog: Nice! ❤️

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
monorkin: Added support for extensions in https://github.com/rails/rails/pull/57100/commits/e477fa19ff18fb9e7e1ce0b925ce116bdfe2a2ea

monorkin: Added them in https://github.com/rails/rails/pull/57100/commits/1784383ef7d813ee5f489efd34fdd96a4b86a53c

actionpack/lib/action_pack/web_authn/public_key_credential.rb
monorkin: Added it in https://github.com/rails/rails/pull/57100/commits/acb79b8e65a25b01d9d34bc56303e341f38bf65d

railties/lib/rails/generators/rails/authentication/templates/app/controllers/my/passkeys_controller.rb.tt
monorkin: I switched to returning `nil` if the registration fails for now - commit https://github.com/rails/rails/pull/57100/commits/8a5eea38ccf5a2601976068309421f7259a9f9ae

In my opinion, if someone feeds in nonsense data the expected result is a client error like a 422. So maybe the better approach is to add an error handler that automatically turns a registration error into a 422.

The biggest advantage of the error is that it propagates the reason for the failure, so if someone wants to, they can process the error and understand what exactly went wrong. While, with the new approach this is lost and there is only a generic "it didn't work" signal.

I'll think about this some more. I'm now leaning towards adding an error handler and letting the code fail. There is no good reason for the request to fail unless someone misconfiguration it or tempered with it.

monorkin: Added `authenticate!` and `register!` in https://github.com/rails/rails/pull/57100/commits/73f1bd85c676d95c26e159801fea33c386ea892d
These raise and can be used when someone wants to inspect the exact failure reason.

actionpack/lib/action_pack/web_authn/public_key_credential/creation_options.rb
monorkin: Improved the interface in https://github.com/rails/rails/pull/57100/commits/4c6107152741d2009644971f7c791f44e9e69a09

actionpack/lib/action_pack/web_authn/public_key_credential.rb
monorkin: Fixed it in https://github.com/rails/rails/pull/57100/commits/7193e026149d053016ba3588dabacde02d1d6446

actionpack/lib/action_pack/passkeys/engine.rb
monorkin: Yeah, I had to draw the line somewhere and this looked to be the precedent set with ActionText and ActiveStorage.

Those seem to have the same quirk - they crash if you explicitly require them in a minimal app.

I don't think there is a good solution here :/

railties/lib/rails/all.rb
monorkin: I updated the generator to manually require the engine in minimal apps https://github.com/rails/rails/pull/57100/commits/ef4ac396ffc42b4fc09f0e2b9a0cad6de2473380, I think that should cover this?

santiagorodriguez96:
Hi again, @monorkin! I don’t think I’ve said it before, but this is really impressive work.

I’m still reviewing the PR, but I wanted to leave a comment because I can’t help wondering: was pulling the lower-level WebAuthn implementation – the CBOR decoder, the COSE key parsing, the ceremony verification, and so on – into a separate gem something you’d considered? I ask because it’s effectively a big part of three independent specs (WebAuthn, CBOR, and COSE), which are constantly evolving, implemented from scratch – feels to me like a lot of surface to add to Rails core. I’m aware of the “no external dependencies” goal – but what if the gem lives under Rails’ own ownership?

What’s more, I think it could be a soft dependency, lazily required only when has_passkeys is used – much like `has_secure_password` leans on `bcrypt` only when it’s used (the generator even runs `bundle add bcrypt`), with a comment noting that’s specifically to avoid the framework depending on a binary library. To me it mirrors how the Postgres/MySQL adapters live in Active Record while the actual wire-protocol drivers (`pg`, `mysql2`) sit in external gems that are required lazily – the Rails-flavored part stays in core, the protocol machinery doesn’t.

Ultimately, I think it comes down to one question: where should the boundary between Rails core and a dedicated gem sit?

monorkin:
Hey @santiagorodriguez96 thank you! And thank you for taking the time to review this 🙏

Excellent questions!

I thought about gem extraction but for a different reason: I have mixed feelings about the engine-within-an-engine design, which a separate gem might solve, but it could introduce some other awkwardness, so I decided to fold everything in and see how it goes.

Personally, I think that the CBOR decoder and COSE parser aren't that complex. There is some inherent complexity when working with binary formats, but besides that each is just one relatively small file that's easy to follow.

WebAuthn is the bulk of the code in this PR. In my mind WebAuthn is akin to HTTP - it's part of the "web standard" - so it makes sense for ActionPack to include this logic. This isn't a hard stance on my part, I'm fine with it being in Rails as well as an external gem.

When I did research for this I saw that the CBOR spec is basically set in stone, while the COSE spec only gets additions in terms of new algorithms. Both protocols are used by hardware authenticators so I don't expect any breaking changes, and any additions will likely have long implementation period anyway.

I saw that WebAuthn does see some changes, but I expect this to slow down now that the spec is widely adopted and fairly mature. I might be mistaken though - curious what you think?

As for the boundary: I don't have a hard stance. I like the current implementation because I don't see the utility of extracting the low-level stuff into a gem. You think that having a separate gem would allow for quicker iteration when the spec changes?

Just a comment on the soft-dependency gems that you mentioned: to me having sqlite3/pg/mysql be a separate gem makes sense because you most often need just one of those for ActiveRecord. BCrypt is optional because of its C extension, [I believe](https://github.com/rails/rails/commit/9b02f3f41f9be96bcc61222a8dcd197c1a2edf79) 

ActionCable actually has a similar case, where the low-level WebSocket protocol is implemented in the websocket gem, but everything around it that drives the socket, channels, and subscriptions is in Rails. To me the current split looks similar to that, but COSE and CBOR are so small that it's easier to include them.

dhh:
I'm happy with where this is heading in its current shape. We're going to battle test it by implementing it for Basecamp. Then we'll have a clear indication of whether anything is missing.

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class CreateActionPackPasskeysPasskeys < ActiveRecord::Migration[8.2]
```
santiagorodriguez96, line 3: I'm getting an error when running the migration:

```
rdm
bin/rails aborted!
NameError: uninitialized constant CreateActionPackPasskeys (NameError)

      Object.const_get(camel_cased_word)
            ^^^^^^^^^^
Did you mean?  CreateActionPackPasskeysPasskeys


Caused by:
NameError: uninitialized constant CreateActionPackPasskeys (NameError)

      Object.const_get(camel_cased_word)
            ^^^^^^^^^^
Did you mean?  CreateActionPackPasskeysPasskeys

Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```


The issue is that the name of the file doesn't match the name of the class

actionpack/lib/action_pack/web_authn/public_key_credential.rb
```diff
@@ -0,0 +1,180 @@
...
+
+        # Verifies an attestation response from the browser and returns a new
+        # PublicKeyCredential with the registered credential data.
+        #
+        # Raises +InvalidResponseError+ if the attestation is invalid.
+        def register(params, **options)
+          response = ActionPack::WebAuthn::Authenticator::AttestationResponse.new(
+            client_data_json: params[:client_data_json],
+            attestation_object: params[:attestation_object],
+            origin: options.fetch(:origin, ActionPack::WebAuthn::Current.origin),
```
santiagorodriguez96, line 103: Should the expected origin be configurable rather than taken from the request?

The [specification recommends Relying Parties not to allow a subdomain origin when verifying the origin of the assertion](https://www.w3.org/TR/webauthn-3/#sctn-code-injection):

> Malicious code could, by the credential [scope](https://www.w3.org/TR/webauthn-3/#scope) rules, be hosted on a subdomain of the [RP ID](https://www.w3.org/TR/webauthn-3/#rp-id). For example, user-submitted code hosted on usercontent.example.org could exercise any [credentials](https://www.w3.org/TR/credential-management-1/#concept-credential) [scoped](https://www.w3.org/TR/webauthn-3/#scope) to the [RP ID](https://www.w3.org/TR/webauthn-3/#rp-id) example.org. If the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) allows a subdomain [origin](https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-origin) when [verifying the assertion](https://www.w3.org/TR/webauthn-3/#rp-op-verifying-assertion-step-origin), malicious users could use this to launch a [man-in-the-middle attack](https://tools.ietf.org/html/rfc4949#page-186) to obtain valid [authentication assertions](https://www.w3.org/TR/webauthn-3/#authentication-assertion) and impersonate the victims of the attack.
> 
> Therefore, the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) by default SHOULD NOT allow a subdomain [origin](https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-origin) when [verifying the assertion](https://www.w3.org/TR/webauthn-3/#rp-op-verifying-assertion-step-origin). If the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) needs to allow a subdomain [origin](https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-origin), then the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) MUST NOT serve untrusted code on any allowed subdomain of [origins](https://www.w3.org/TR/webauthn-3/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised) within the [scope](https://www.w3.org/TR/webauthn-3/#scope) of its [public key credentials](https://www.w3.org/TR/webauthn-3/#public-key-credential).

Deriving it from the origin makes the expected origin submitter-controlled and wouldn't protect users from that issue as subdomains would be accepted.

railties/lib/rails/generators/rails/authentication/templates/app/controllers/sessions/magic_links_controller.rb.tt
```diff
@@ -0,0 +1,23 @@
+class Sessions::MagicLinksController < ApplicationController
+  allow_unauthenticated_access
+  rate_limit to: 10, within: 3.minutes, only: :create, with: -> { redirect_to session_magic_link_path, alert: "Try again later." }
+  before_action :ensure_pending_authentication
+
+  def show
+  end
+
+  def create
+    if magic_link = MagicLink.consume(params[:code])
```
jhubert, line 10: `MagicLink.consume` does a global `active.find_by(code:)`, so a submitted code is checked against **every user's** active codes not just the account being signed in to. With 6-character codes a guessed code that's active for *anyone* signs the attacker in as that person, and the odds scale with the number of concurrently active links. The per-IP rate limit slows this but doesn't bound it across distributed sources.

The controller already knows who's authenticating so scoping the lookup is nearly free:

```ruby
user = User.find_by(email_address: email_address_pending_authentication)
if magic_link = user&.consume_magic_link(params[:code])
```

(or pass the user/email into `MagicLink.consume`). That makes a code worthless without also controlling the pending-authentication session for that email, and brute-force odds become per-account instead of global.

We run passkey + magic-link auth in production for a board-governance product and scope code lookup per-user for exactly this reason (we also use 8-character codes, which may be worth considering as defense in depth. The UX cost of two extra characters is small).

actionpack/lib/passkeys/app/javascript/action_pack/passkey.js
```diff
@@ -0,0 +1,226 @@
...
+    hideErrors(this)
+    this.button.dispatchEvent(new CustomEvent("passkey:start", { bubbles: true }))
+
+    try {
+      const options = this.options
+
+      if (!passkeysAvailable()) throw new Error("Passkeys are not supported by this browser")
+      if (!options) throw new Error("Missing passkey options")
+
+      await refreshChallenge(options, this.challengeUrl, this.purpose)
```
jhubert, line 68: Flagging a failure mode we hit repeatedly in production on iOS/Safari: awaiting a `fetch` between the click and `navigator.credentials.create()/get()` consumes Safari's transient activation window when the network is slow, and the ceremony then throws `NotAllowedError`. Because `NotAllowedError` is mapped to `cancelled` below, the failure is indistinguishable from the user dismissing the sheet. It took us client-side failure telemetry to discover how often "cancellations" were actually this.

It reproduces reliably on real cellular connections (not on fast Wi-Fi), which is exactly where board-member-demographic users live. We shipped three rounds of fixes before converging on: fetch the challenge on `connectedCallback` so the WebAuthn call runs synchronously inside the gesture, and renew it in the background before it expires.

Since the options here already contain a render-time challenge, a middle ground that keeps the just-in-time design: refresh on `connectedCallback` (and on a timer near expiry) instead of inside the click handler, falling back to the last-fetched challenge if a refresh hasn't completed. That preserves expiring challenges without putting network latency inside the activation window.

Related: bucketing `NotAllowedError` as `cancelled` might deserve a note in the docs. It also covers timeouts and this activation expiry, so apps monitoring adoption will undercount real failures.

actionpack/lib/passkeys/app/models/action_pack/passkeys/passkey.rb
```diff
@@ -0,0 +1,151 @@
...
+      #
+      # The user_verification requirement enforced during verification is the same one that
+      # would be used to build this holder's #authentication_options (global defaults overridden
+      # by the holder's own +passkey_authentication_options+).
+      def authenticate!(passkey)
+        credential = to_public_key_credential
+        authentication_options = self.class.authentication_options(holder: holder)
+
+        credential.authenticate(passkey, user_verification: authentication_options.user_verification)
+        update!(sign_count: credential.sign_count, backed_up: credential.backed_up)
```
jhubert, line 134: The read → validate → `update!` here isn't serialized, so two concurrent assertions can both load the same stored `sign_count`, both pass `sign_count_must_increase`, and both persist which defeats the replay protection that validation exists to provide (a cloned credential replaying a captured assertion in parallel with a legitimate one).

I think wrapping the verify-and-update in a row lock closes it:

```ruby
def authenticate!(passkey)
  with_lock do
    credential = to_public_key_credential
    authentication_options = self.class.authentication_options(holder: holder)

    credential.authenticate(passkey, user_verification: authentication_options.user_verification)
    update!(sign_count: credential.sign_count, backed_up: credential.backed_up)
  end

  self
end
```

Reloading inside the lock (or `lock!` + re-reading `sign_count`) matters. The check has to run against the row's current value, not the one loaded before the lock was acquired. We've run this pattern in production without issue; the lock is held only for the signature verification, which is microseconds.

jhubert (commented):
Very excited for this one. We did a custom implementation of Passkeys and MagicLinks in our app and it works really well, but I'd love to standardize on a Rails implementation of it. I tried [extracting ours last year](https://github.com/jhubert/passkey-auth) but couldn't quite get the right level of abstraction. Nice work on finding it!

I've added a few comments based on our experience in production. The most important one being the iOS bug that we spent a log time trying to figure out. Hope they're somewhat helpful. 😃

actionpack/lib/passkeys/app/assets/javascripts/actionpack-passkeys.esm.js
```diff
@@ -0,0 +1,278 @@
...
+      const type = errorType(error);
+      this.button.dispatchEvent(new CustomEvent("passkey:error", {
+        bubbles: true,
+        detail: {
+          error: error,
+          type: type
+        }
+      }));
+    } finally {
+      this._conditionalMediationController = null;
```
jhubert, line 194: Once conditional mediation is aborted (button click) or fails, it's never restarted. `_performCeremony` awaits `abortConditionalMediation()` and the `finally` here nulls the controller, so autofill is dead for the remainder of the page's life.

The sequence that makes this matter: user taps the sign-in button, the sheet offers the wrong credential (or they misfire), they cancel then tap the email field expecting the passkey autofill suggestion. Nothing appears, and there's no visible way to get it back short of reloading. We found this a common recovery path on iOS in particular and added an explicit re-arm after a cancelled/failed modal ceremony.

Suggestion: after a non-AbortError failure in `_performCeremony` (and after the conditional attempt itself fails with `NotAllowedError`), call `_attemptConditionalMediation()` again with a fresh challenge. The abort-before-modal handling here is already correct, it's only the restart that's missing.

railties/lib/rails/generators/rails/authentication/templates/app/controllers/my/passkeys_controller.rb.tt
```diff
@@ -0,0 +1,36 @@
+class My::PasskeysController < ApplicationController
+  include ActionPack::Passkeys::Request
+
+  before_action :set_passkey, only: %i[ edit update destroy ]
+
+  def index
+    @passkeys = Current.user.passkeys.order(name: :asc, created_at: :desc)
+    @registration_options = passkey_registration_options(holder: Current.user)
```
santiagorodriguez96, line 8: I think we should require `user_verification` here in order to turn passkeys login into a MFA login 👀

railties/lib/rails/generators/rails/authentication/templates/app/controllers/sessions_controller.rb.tt
```diff
@@ -1,17 +1,32 @@
 class SessionsController < ApplicationController
+  include ActionPack::Passkeys::Request
+
   allow_unauthenticated_access only: %i[ new create ]
   rate_limit to: 10, within: 3.minutes, only: :create, with: -> { redirect_to new_session_path, alert: "Try again later." }
 
   def new
+    @authentication_options = passkey_authentication_options
```
santiagorodriguez96, line 8: Same here for the authentication flow 🙂

actionpack/lib/action_pack/passkeys/holder.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+
+      private
+        def extract_options_for(klass, options)
+          keys = klass.attribute_names.map(&:to_sym)
+
+          extracted = options.slice(*keys)
+          options.except!(*keys)
+          extracted if extracted.any?
+        end
```
santiagorodriguez96, line 130: nit: to be honest, the first time I read it, I found a little bit off for the `options` param to be mutated.

Perhaps adding a `!` at the end of the method name or maybe changing it to `extract_and_delete_options_for` could help?

santiagorodriguez96: Also, I think I'm not following why the options are being removed after being read altogether. For instance, doing:

```
has_passkeys user_verification: :required
```

would likely apply only to registration, not authentication, right? 

Shouldn't shared options apply to both ceremonies instead?

Maybe I'm missing something? 🤔

actionpack/lib/action_pack/web_authn/authenticator/response.rb
```diff
@@ -0,0 +1,156 @@
...
+            return if errors.any?
+
+            signed_message = Base64.urlsafe_decode64(client_data["challenge"])
+
+            unless ActionPack::WebAuthn.challenge_verifier.verified(signed_message, purpose: challenge_purpose)
+              errors.add(:base, "Challenge has expired")
+            end
+          rescue ArgumentError
+            errors.add(:base, "Challenge is invalid")
+          end
```
santiagorodriguez96, line 99: Isn't this vulnerable to reply attacks within the challenge window given that it is never compared to a stored per-ceremony challenge and also never marked as used?

The [specification specifically recommends this](https://www.w3.org/TR/2026/CR-webauthn-3-20260526/#sctn-cryptographic-challenges:~:text=the%20Relying%20Party%20SHOULD%20store%20the%20challenge%20temporarily%20until%20the%20operation%20is%20complete):

> Therefore, the values of both [PublicKeyCredentialCreationOptions](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialcreationoptions).[challenge](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-challenge) and [PublicKeyCredentialRequestOptions](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialrequestoptions).[challenge](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialrequestoptions-challenge) MUST be randomly generated by [Relying Parties](https://www.w3.org/TR/webauthn-3/#relying-party) in an environment they trust (e.g., on the server-side), and the returned [challenge](https://www.w3.org/TR/webauthn-3/#dom-collectedclientdata-challenge) value in the client’s response MUST match what was generated. This SHOULD be done in a fashion that does not rely upon a client’s behavior, e.g., the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party) SHOULD store the challenge temporarily until the operation is complete.

The only guard left would be the signature count but even then some authenticators always set the sign count to `0` which is accepted.

actionpack/lib/action_pack/web_authn/public_key_credential/options.rb
```diff
@@ -0,0 +1,109 @@
...
+      include ActiveModel::API
+      include ActiveModel::Attributes
+
+      CHALLENGE_LENGTH = 32
+      USER_VERIFICATION_OPTIONS = %i[ required preferred discouraged ].freeze
+
+      attribute :user_verification, default: :preferred
+      attribute :relying_party, default: -> { ActionPack::WebAuthn.relying_party }
+      attribute :timeout
+      attribute :extensions
```
santiagorodriguez96, line 54: Extensions are now being sent as part of the options but there's no way of accessing the extensions results in the response, right?

In fact, the JS components are not requesting the extensions results nor sending them back to the RP.

actionpack/lib/action_pack/web_authn/authenticator/attestation_response.rb
```diff
@@ -0,0 +1,79 @@
...
+      #
+      # * The client data type is "webauthn.create"
+      # * The attestation format has a registered verifier
+      # * The attestation statement passes format-specific verification
+      #
+      class AttestationResponse < Response
+        attr_reader :attestation_object
+
+        validate :client_data_type_must_be_create
+        validate :attestation_must_be_valid
```
santiagorodriguez96, line 38: Should this class also validate that if the `backup_eligible` flag is not set, then the `backed_up` flag is not set?

This is [step `17` of the steps for verifying a registration ceremony in the specification](https://www.w3.org/TR/2026/CR-webauthn-3-20260526/#sctn-registering-a-new-credential:~:text=authData%20is%20set.-,If%20the%20BE%20bit%20of%20the%20flags%20in%20authData%20is%20not%20set%2C%20verify%20that%20the%20BS%20bit%20is%20not%20set,-.):

> 17. If the [BE](https://www.w3.org/TR/2026/CR-webauthn-3-20260526/#authdata-flags-be) bit of the [flags](https://www.w3.org/TR/2026/CR-webauthn-3-20260526/#authdata-flags) in authData is not set, verify that the [BS](https://www.w3.org/TR/2026/CR-webauthn-3-20260526/#authdata-flags-bs) bit is not set.

santiagorodriguez96 (commented):
Thanks for such a thoughtful and warm reply, @monorkin. I think your points are fair.

> WebAuthn is the bulk of the code in this PR. In my mind WebAuthn is akin to HTTP - it's part of the "web standard" - so it makes sense for ActionPack to include this logic.

That's a really interesting way of seeing it! 🙂 
 
> When I did research for this I saw that the CBOR spec is basically set in stone, while the COSE spec only gets additions in terms of new algorithms. Both protocols are used by hardware authenticators so I don't expect any breaking changes, and any additions will likely have long implementation period anyway.
> 
> I saw that WebAuthn does see some changes, but I expect this to slow down now that the spec is widely adopted and fairly mature. I might be mistaken though - curious what you think?

You're right that CBOR and COSE are essentially stable: CBOR is frozen, and COSE really only grows by adding new algorithms, so there's very little maintenance burden on that side.

On WebAuthn though, I do think it's maturing but more in the sense of being hardened by real-world use than settling down – as adoption grows, real-world gaps surface and get addressed (related origins, the signal API, hints, the backup-state flags, conditional UI). What's more, as L3 is being wrapped up, Level 4 is already underway with a good few proposals in flight: `requestUserInfo`/`challengeUrl` for smoother account creation, immediate mode, algorithm migration, hybrid fallback URLs, credential-manager trust groups, and so on. As adoption is expected to keep growing, I wouldn't bet on the pace slowing just yet.
 
> As for the boundary: I don't have a hard stance. I like the current implementation because I don't see the utility of extracting the low-level stuff into a gem. You think that having a separate gem would allow for quicker iteration when the spec changes?

I think that could be a good argument for extracting it into a gem – since the implementation is brand new and yet to be battle tested, being able to ship fixes and updates on its own schedule, rather than waiting on Rails' release timeline, would allow for faster iteration. However, if bug and security fixes can be released as needed as Rails patch versions and features as part of minor versions ([expected to be released every six months](https://guides.rubyonrails.org/maintenance_policy.html#release-schedule:~:text=We%20aim%20to%20release%20a%20version%20containing%20new%20features%20every%20six%20months)), that might be enough for this case.

Mostly I was curious about your thought process, and you've more than covered it. Great work on this! 🙌

railties/lib/rails/generators/rails/authentication/authentication_generator.rb
```diff
@@ -40,24 +66,88 @@ def configure_application_controller
...
+          unless @user_model_exists
+            generate "migration", "CreateUsers", "email_address:string!:uniq", "--force"
+          end
+
+          generate "migration", "CreateMagicLinks", "user:references code:string!:uniq expires_at:datetime!", "--force"
         end
+
         generate "migration", "CreateSessions", "user:references ip_address:string user_agent:string", "--force"
+
+        rails_command "railties:install:migrations FROM=action_pack_passkey", inline: true
```
dixpac, line 150:
```suggestion
        rails_command "railties:install:migrations FROM=action_pack_passkeys", inline: true
```

engine name is plural

actionpack/lib/passkeys/db/migrate/20260213154740_create_action_pack_passkeys.rb
dixpac, line 3:
```suggestion
class CreateActionPackPasskeys < ActiveRecord::Migration[8.2]
```

i think object name should be like this in order for `const_get` to work

PR #53541, opened 2024-11-05
Update ActiveJob docs to show how to halt the callback chain

camallen:
### Motivation / Background

This Pull Request has been created because: I was reviewing a PR that used the callback chain halt technique but couldn't see any details in the ActiveJob docs on how these worked. I figure this change will be useful for someone else and avoid them having to read the source / other docs to find this info.

### Detail

This Pull Request changes the ActiveJob docs, to explain how job callback chain halting works. I coped the details from Active Record and provided a link to https://api.rubyonrails.org/classes/ActiveSupport/Callbacks.html for more details.

### Additional information

Seems the main place rails docs the [ActiveSupport callbacks](https://github.com/rails/rails/blob/a6dee8c5b1baffa0ac9e57016b90d8c88a8db1f6/activesupport/lib/active_support/callbacks.rb) are in https://guides.rubyonrails.org/active_record_callbacks.html. There is some information in the [ActiveSupport docs](https://api.rubyonrails.org/classes/ActiveSupport/Callbacks.html) 

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #57101, opened 2026-03-30
[RF-Docs] [ci-skip] Active Job Basics guide

bhumi1102:
The purpose of this PR is to update the Active Job Basics guide as part of the Rails Foundation documentation project. 

Here's a high level of the changes:

- Restructured the entire guide top level sections and subsections so that it's more cohesive. Removed tiny sections that were not adding any value (as noted in the audit)
- Rewrite the big Solid Queue section and restructure it's subsections with the objective of giving the reader a more of a conceptual overview of how Solid Queue is designed, what's unique about it, how it compares to other queuing backends and highlighting it's features and how to use it in practice. I removed the existing content that was mirroring what is already in the Solid Queue README on Github, so we are not duplicating that.
- Addressed all of the comments from the initial audit

- [X] Nitpick: "Create and Enqueue Jobs" should probably be changed to "Creating and Enqueuing Jobs" which is more common in the guides.
- [X] Maybe the bulk enqueuing top section should be inlined under "2.3. Enqueue Jobs in Bulk" instead of linking to it? It's not a big section

Solid Queue
- [X] should “asynchronous in-process queuing system” be renamed to “asynchronous in-process queuing  backend”
- [X] The the queue in the database.yml examples for development and production could be made clearer by highlighting the queue lines. -> good call
- [X] The order of the defined queues having precedence over job priority might be made clearer if we show an example of the defined queues (production and background) -> Rewrote this
- [X] ⚠️ is used for warning. A WARNING block should probably be used instead. Although this would require some rewriting as we probably don’t want to wrap the whole section in a WARNING block.

Queues
- [X] Maybe this section should be renamed to something like "Enqueuing Jobs" to better describe what it is about? It's not really about the queues itself. -> Restructured the sections
- [X] Maybe add titles to split this up in sections: "Naming Queues" and "Dynamic Enqueuing"? -> Restructured
- [X] Maybe we should recommend a naming scheme for queues? [Latency based naming](https://engineering.gusto.com/scaling-sidekiq-at-gusto-3f9e3279e63) is a popular option. Jean Boussier: "You can mention it as a popular option without making it a recommendation." -> good one, added the recommendation.

Priority
- [X] I think this section can be a sub section of the "Queues/Enqueuing Job" section.

Callbacks
- [X] The callbacks examples could be made clearer by using line highlighting for just the callback code.
- [X] None of the callbacks are described unlike other guides where we describe the callbacks. -> added code snippets for all
- [X] There is an open PR to add a section about halting callbacks, which makes sense to mention: https://github.com/rails/rails/pull/53541 -> thanks for the heads up, added the halting section!

Bulk Enqueuing
- [X] "perform_all_later is a top-level API on Active Job." doesn't seem to add anything and could be removed?
- [X] If we have a "Enqueuing Jobs" section this could be a sub section.

ActionMailer
- [X] This section can probably be removed as this is already described in the ActionMailer guide. Other frameworks also use ActiveJob but aren't mentioned, for example ActionMailbox `incinerate_later` and ActiveStorage `purge_later`. -> +1 removed!

Internationalization
- [X] This section also doesn't add much and seems more related to ActionMailer.

Supported Types for Arguments
- [X] This section seems relevant to the "Enqueuing Jobs" jobs section and should probably be a sub section of that? -> moved to a new section about defining a job

Exceptions
- [X] I think renaming this title to something like "Handling Failed Jobs" better describes what this section is about.

Job Testing
- [X] Could probably be removed? Do we have sections like this for other frameworks? -> removed, doesn't add value agree.

Debugging
- [X] Maybe this should be a part of the "Exceptions" top section?
- [X] Verbose enqueuning logging is enabled by default in development so I don't think it deserves it's own section.

Alternate Queuing Backends
- [X] Maybe it makes sense to move this section after the Solid Queue section? Or we can keep it at the end if we really want to de-emphasize alternative backends. -> restructured such that those sections are next to each other.


seuros:
Thanks for the effort here, the Solid Queue concurrency controls section and the transactional integrity rewrite are improvements worth keeping.

However, the bulk of this PR is restructuring and rewording existing content, which we've generally tried to avoid in the guides. Cosmetic changes create noise in git history and make future diffs harder to read, without improving what a developer actually learns.

The preamble NOTE (Active Job is an interface, not an execution engine) is redundant, any reader who continues past the intro will see the backend configuration and understand this naturally.

Moving sections around and converting inline comments to prose doesn't add information.

The guides have a deliberate flow: teach the concept, then point to the reference docs at the end. 
Now reference links are scattered mid-section. At this rate they will end up in the intro. 
Keep references where they belong at the bottom, after the concept is fully explained.

I'd suggest to keep  only the new things in 2 separate PR
Also, Rails has examples in the code and the tests. The guide is not meant to be a masterclass, any refactoring will require updating both the examples and the guide.

bhumi1102:
> However, the bulk of this PR is restructuring and rewording existing content, which we've generally tried to avoid in the guides.

That is intentional, as this PR is part of the Rails Foundation documentation project (as is anything else tagged "rails foundation"). CC: @AmandaPerino

seuros:
> That is intentional

Could you point me to the new style guidelines ? I have been contributing to this codebase for over a decade and I can't find anything that marks a break from the approach we have always used.

I'm not saying the content is wrong,  I actually tried to contribute something similar myself at one point, because the guide did feel incomplete at first.  @rafaelfranca was clear that diff to be focused. 

But over time I noticed that developers use this guide very differently: some grep it (i do now), some have indexers, and now we have LLMs and RAG pipelines. 

When a RAG system see **Lines changed: 688 additions & 694 deletions** , it doesnt say "oh, that is just reformatting, skip it".

It invalidates the entre cache and treat the same information as if it never existed.  

When i see a big diff, i read it, becaue something must have changed (a breaking change).

I raised this because only 2 of the changes here are genuinely new, the rest is noise.

This particular guide doesn't affect me personally, I know Active Job well. But my younger self would have read it, and I think I would have gone to PHP or Haskell. 

As an engineer, the new format reads like alternating prose and technical content, my brain has to context-switch constantly between the two modes. When I did read it to review it, I really glitched.

Just for info: I did read both versions (not with AI), the new version is feeling off , like something was stripped for good.

AmandaPerino:
> Could you point me to the new style guidelines ? I have been contributing to this codebase for over a decade and I can't find anything that marks a break from the approach we have always used.

@seuros It’s true that edits to the Rails docs have generally been kept minimal, but PRs from the Rails Foundation documentation work do sometimes result in bigger diffs. (Work started [2 years ago](https://rubyonrails.org/2024/2/6/documentation-update-work-has-begun) and we're nearly finished.)

The goal of this project has been to make the guides accurate, up-to-date, and consistent, but also friendly for beginners, so the team approaches each guide differently than a normal docs PR that you have seen in the past. They start with an audit looking at the guide as a whole before the edits begin. This does sometimes result in moving a section or two for a better flow of information. 

The ‘beginner-friendly’ part isn’t always a science, and depends a lot on the writer and the reader.  That said, you've got a great point about RAG systems. We’ll definitely note that down for the last few guides of this project (there aren't many left). 

You called out a few points that bothered you above (thanks for being specific). However, you also mentioned some sections caused you to context-switch - could you point to a few of those specific examples in the file? What would have driven your younger self to PHP or Haskell 😬? (Drastic, but funny. Thank you for sticking around.) That would help us figure out where the balance might be off.

Thanks again for your thoughts.

guides/source/active_job_basics.md
```diff
@@ -18,24 +18,32 @@ After reading this guide, you will know:
...
+
+```ruby
+# If you want to send the email now use #deliver_now
+UserMailer.welcome(@user).deliver_now
+
+# If you want to send the email asynchronously use #deliver_later
+UserMailer.welcome(@user).deliver_later
+```
+
+This section provides a step-by-step guide for defining a job Ruby class and then using the `perform_*` method to enqueue work to be executed in the background.
```
ttilberg: This is the first time the `perform_now` and `perform_later` interfaces are introduced, and it happens immediately after showing an example of `deliver_now` and `deliver_later`, and without prepping them for the difference. I think the mailer example should come after first introducing the main interface methods.

bhumi1102: Good point, will update!

guides/source/active_job_basics.md
```diff
@@ -18,24 +18,32 @@ After reading this guide, you will know:
...
+
+One of the most common jobs in a modern web application is sending emails to users. Active Job can do this outside of the request-response cycle, so the user doesn't have to wait on it. Active Job is integrated with Action Mailer so you can easily send emails asynchronously:
+
+```ruby
+# If you want to send the email now use #deliver_now
+UserMailer.welcome(@user).deliver_now
+
+# If you want to send the email asynchronously use #deliver_later
+UserMailer.welcome(@user).deliver_later
+```
```
ahmad19: Minor nit: the comments are missing a comma after the introductory clause .

```suggestion
# If you want to send the email now, use #deliver_now
UserMailer.welcome(@user).deliver_now

# If you want to send the email asynchronously, use #deliver_later
UserMailer.welcome(@user).deliver_later
```
```

File: guides/source/active_job_basics.md (hunk 1 above)
Comment (bhumi1102) [reply guides/source/active_job_basics.md]: Done. Moved the mailer example and added an explanation of the difference.

File: guides/source/active_job_basics.md (hunk 2 above)
Comment (bhumi1102) [reply guides/source/active_job_basics.md]: Added commas

File: guides/source/active_job_basics.md [hunk 3]
```diff
@@ -65,434 +69,222 @@ class GuestsCleanupJob < ApplicationJob
...
-# free.
-GuestsCleanupJob.perform_later guest
-```
+You can generate a new job that inherits from it:
 
-```ruby
-# Enqueue a job to be performed tomorrow at noon.
-GuestsCleanupJob.set(wait_until: Date.tomorrow.noon).perform_later(guest)
+```bash
+bin/rails generate job process_payment --parent=payment_job
```
p8:
```suggestion
$ bin/rails generate job process_payment --parent=payment_job
```

guides/source/active_job_basics.md
```diff
@@ -65,434 +69,222 @@ class GuestsCleanupJob < ApplicationJob
...
 end
 ```
 
-In this example, only two `MyJob` instances for the same account will run
-concurrently. After that, other jobs will be blocked until one completes.
-
-The `group` parameter can be used to control concurrency across different job
-types. For instance, two different job classes that use the same group will have
-their concurrency limited together:
+Using GlobedId, you can simply do:
```
p8:
```suggestion
Using GlobalID, you can simply do:
```

guides/source/active_job_basics.md
```diff
@@ -65,434 +69,222 @@ class GuestsCleanupJob < ApplicationJob
...
+[`perform_later`]:
+ https://api.rubyonrails.org/classes/ActiveJob/Enqueuing/ClassMethods.html#method-i-perform_later
+[`set`]:
+    https://api.rubyonrails.org/classes/ActiveJob/Core/ClassMethods.html#method-i-set
 
-In the above example, workers will fetch jobs from queues starting with
-"active_storage", like  the `active_storage_analyse` queue and
-`active_storage_transform` queue. Only when no jobs remain in the
-`active_storage`-prefixed queues will workers move on to the `mailers` queue.
+### Supported Parameter Types for `perform`
```
p8: It seems this guide mostly uses "argument" instead of "parameter":
```suggestion
### Supported Argument Types for `perform`
```

guides/source/active_job_basics.md
```diff
@@ -65,434 +69,222 @@ class GuestsCleanupJob < ApplicationJob
...
+  - `ActiveSupport::HashWithIndifferentAccess`
+  - `Array`
+  - `Range`
+  - `Module`
+  - `Class`
 
-### Concurrency Controls
+Active Job supports
+[GlobalID](https://github.com/rails/globalid/blob/main/README.md) for
+parameters. This makes it possible to pass live Active Record objects to your
```
p8:
```suggestion
arguments. This makes it possible to pass live Active Record objects to your
```

guides/source/active_job_basics.md
```diff
@@ -587,8 +381,9 @@ class ProcessVideoJob < ApplicationJob
 end
 ```
 
-```ruby
-ProcessVideoJob.perform_later(Video.last)
+```irb
+last_video = Video.last
+ProcessVideoJob.perform_later(last_video)
```
p8, line 379: If this is an `irb` example, we should use the prefix:
```suggestion
demo(dev)> last_video = Video.last
demo(dev)> ProcessVideoJob.perform_later(last_video)
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
-[Other backends](#alternate-queuing-backends) like Sidekiq have a `push_bulk`
-method, which can push a large number of jobs to Redis and prevent the round
-trip network latency. GoodJob also supports bulk enqueuing with the
-`GoodJob::Bulk.enqueue` method.
+Finally, to start the queue and start processing jobs you can run:
 
-If the queue backend does *not* support bulk enqueuing, `perform_all_later` will
-enqueue jobs one by one.
+```bash
+bin/jobs start
```
p8:
```suggestion
$ bin/jobs start
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
 
-Each job uses the `I18n.locale` set when the job was created. This is useful if
-you send emails asynchronously:
+Solid Queue uses three types of processes to handle job queueing and execution.
+*Workers* poll queues for jobs that are ready to run and execute them.
+*Dispatchers* handle scheduled jobs — they check for jobs due to run in the
+future and move them into the ready queue for workers to pick up. Both workers
+and dispatchers are managed by a *Supervisor*, which forks and monitors them.
+
+When you run `bin/rails jobs`, you're starting the supervisor process, which in
```
p8: I think Solid Queue is started by running `bin/jobs start`:
```suggestion
When you run `bin/jobs start`, you're starting the supervisor process, which in
```
https://github.com/rails/solid_queue/blob/main/lib/generators/solid_queue/install/templates/bin/jobs

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
 ```
 
-NOTE: Using the asynchronous queue from a Rake task (for example, to send an
-email using `.deliver_later`) will generally not work because Rake will likely
-end, causing the in-process thread pool to be deleted, before any/all of the
-`.deliver_later` emails are processed. To avoid this problem, use `.deliver_now`
-or run a persistent queue in development.
+Add `queue` to the development database configuration:
 
+```yaml#6
```
p8: We can probably highlight keys under `queue` as well:
````suggestion
```yaml#6-9
````

File: guides/source/active_job_basics.md [hunk 11]
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
---------------------
+### Workers, Dispatchers, Supervisors
 
-Each job uses the `I18n.locale` set when the job was created. This is useful if
-you send emails asynchronously:
+Solid Queue uses three types of processes to handle job queueing and execution.
+*Workers* poll queues for jobs that are ready to run and execute them.
+*Dispatchers* handle scheduled jobs — they check for jobs due to run in the
+future and move them into the ready queue for workers to pick up. Both workers
+and dispatchers are managed by a *Supervisor*, which forks and monitors them.
```
p8: Would it be clearer if we describe the types in a list?
```suggestion
Solid Queue uses three types of processes to handle job queueing and execution.
1. *Workers* poll queues for jobs that are ready to run and execute them.
2. *Dispatchers* handle scheduled jobs — they check for jobs due to run in the
future and move them into the ready queue for workers to pick up.
3. *Supervisor* manages both workers and dispatchers — it forks and monitors them.
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+In the above example, both job classes share the same concurrency limit per the
+"account_exports" group, which means only one export of either type will run at
+a time for a given account.
+
+NOTE: Concurrency controls do carry overhead since blocked executions must be
+tracked and locks created and updated. So they should be used sparingly. For
+simple throughput limiting, constraining the number of worker threads per queue
+is more efficient.
+
+WARN: Concurrency controls are not compatible with bulk enqueuing via
```
p8: This is currently not getting picked up as a warning:

Image

```suggestion
WARNING: Concurrency controls are not compatible with bulk enqueuing via
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+designed to work seamlessly with different queuing backends. Switching to an
+alternative backend, such as [Sidekiq](https://github.com/sidekiq/sidekiq),
+[GoodJob](https://github.com/bensheldon/good_job), or
+[Resque](https://github.com/resque/resque), requires only a configuration change
+(typically with no modifications to your job code), along with adding the
+queuing backend's adapter to your Gemfile.
+
+You can get an up-to-date list of the adapters in the the API Documentation for
+[`ActiveJob::QueueAdapters`](https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters.html).
+
```
p8: All the build-in adapters for external gems have been [deprecated](https://github.com/search?q=repo%3Arails%2Frails+activejob+deprecated&type=pullrequests), so we should probably remove this as the list won't list external adapters in the future:
```suggestion
```

Another option would be showing the list of alternative backends here, instead of at the end of this section.

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+long-running jobs can make progress after interruptions. When using
+continuations, the job automatically resumes from the last completed step,
+instead of restarting from the beginning.
 
-ActiveJob.perform_all_later(cleanup_jobs)
-```
+To use continuations, include the `ActiveJob::Continuable` module in your Job
+class. You can then define each step inside the `perform` method using
+[`step`](https://api.rubyonrails.org/classes/ActiveJob/Continuation/Step.html)
+inside.
```
p8: The first example shows a lot of new concepts.
Maybe show a simple example here (without the cursor)?

```ruby
class ProcessImportJob < ApplicationJob

  def perform(import_id)
    # Always runs on job start, even when resuming from an interrupted step.
    @import = Import.find(import_id)

    # Step defined using a block
    step :initialize do
      @import.initialize
    end

    step :process do
      @import.records.find_each { |record| record.process }
    end

    # Step defined by referencing a method
    step :finalize
  end

  private
    def finalize
      @import.finalize
    end
end
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+to track progress *within* the step. The code in the step is responsible for
+using the cursor to continue from the appropriate location after an
+interruption.
 
-The return value of `perform_all_later` is `nil`. Note that this is different
-from `perform_later`, which returns the instance of the queued job class.
+Steps are executed as soon as they are encountered. Code that is not part of a
+step will be executed on each job run. If a job is interrupted, previously
+completed steps will be skipped. If a step is in progress, it will be restarted
+or resumed with the last recorded cursor if using cursors.  For example:
```
p8: Before introducing cursors maybe explain how regular steps are executed first?
For example moving "Steps are executed .. steps will be skipped" sentences before "Steps can use an optional cursor...":
```suggestion
Steps are executed as soon as they are encountered. Code that is not part of a
step will be executed on each job run. If a job is interrupted, previously
completed steps will be skipped.

Steps can use an optional
[cursor](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html#class-ActiveJob::Continuation-label-Cursors)
to track progress *within* the step. The code in the step is responsible for
using the cursor to continue from the appropriate location after an
interruption.

If a step is in progress, it will be restarted
or resumed with the last recorded cursor if using cursors.  For example:
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
 
-### Enqueue Multiple Active Job Classes
+```ruby
```
p8, line 692: If we show a simpler example first, we could remove some of the code from this more complex example, or only show the step with cursor.

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+Default Backend: Solid Queue
+------------------------------
+
+Solid Queue is a database-backed queue backend for Active Job and the default
+queue backend for Rails version 8.0 onwards. Rather than requiring a separate
+infrastructure dependency like Redis, Solid Queue uses your existing database to
+persist and process jobs. It supports delayed jobs, job priorities, concurrency
+controls, recurring jobs, and bulk enqueuing.
+
+### Set Up and Default Configuration
```
p8: In other guides we use "Setup":
```suggestion
### Setup and Default Configuration
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+Rails provides an asynchronous in-process queuing backend, which keeps the jobs
+in memory. With the default `async` adapter, if the process crashes or the
+machine is reset, then all outstanding jobs are lost. This can be acceptable for
+non-critical jobs in development.
 
-```ruby
-# If you want to send the email now use #deliver_now
-UserMailer.welcome(@user).deliver_now
+Alternatively, you can use Solid Queue in development. It can be configured it
+in the same way as in the production environment:
```
p8:
```suggestion
Alternatively, you can use Solid Queue in development. It can be configured
in the same way as in the production environment:
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+strict ordering (unlike other queuing backend which may allow relative weights
+so that lower priority queues still receive a proportional share of processing
+time). It is possible for lower queues to be starved if higher queues are
+consistently busy.
+
+NOTE: One way to name queues is based on latency. So instead of "critical",
+"default", or "low", queues could be named "within_30_seconds",
+"within_5_minutes", and "within_1_hour". This can be enforced like a contract by
+configuring your queuing backend to notify your engineering team if a job sits
+in a given queue longer than the corresponding time.
```
p8: This NOTE is probably interesting with other backend as well.
What do you think about moving it to the "Naming Queues" section instead?

Also, naming queues isn't required for understanding queue order. So should it be a TIP instead of a NOTE?

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+
+  def perform(account_id)
+    # ...
   end
 end
 ```
 
-This works with any class that mixes in `GlobalID::Identification`, which by
-default has been mixed into Active Record classes.
+In the above example:
```
p8, line 1077: I think this needs a linebreak:

Image

```suggestion
In the above example:

```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+- The `to` option sets the maximum number of jobs that can run concurrently.
+- The `key` lambda computes a concurrency key from the job's arguments. In the
+  example above, the limit of 1 applies per account rather than globally.
+- The `duration` option acts as a failsafe. So if a worker dies mid-job and
+  fails to release its lock, any blocked jobs become candidates for release once
+  duration has elapsed.
```
p8: In the next paragraph options are prepended with a colon (mostly):
```suggestion
- The `:to` option sets the maximum number of jobs that can run concurrently.
- The `:key` lambda computes a concurrency key from the job's arguments. In the
  example above, the limit of 1 applies per account rather than globally.
- The `:duration` option acts as a failsafe. So if a worker dies mid-job and
  fails to release its lock, any blocked jobs become candidates for release once
  duration has elapsed.
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+When a job with concurrency controls is enqueued, Solid Queue checks a
+database-backed lock for the computed key. If the lock is available, the job is
+marked ready for execution. If not, the behavior depends on the `on_conflict:`
+option. If `on_conflict` is set to `:block` (the default), the job is held in a
+blocked state and marked ready only when a running job finishes. The other
+option is `:discard`, in which case the job is dropped entirely.
 
-You can extend the list of supported argument types. You just need to define
-your own serializer:
+You can also scope limits across *different* job classes using the `group:`
```
p8:
```suggestion
You can also scope limits across *different* job classes using the `:group`
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+- The `key` lambda computes a concurrency key from the job's arguments. In the
+  example above, the limit of 1 applies per account rather than globally.
+- The `duration` option acts as a failsafe. So if a worker dies mid-job and
+  fails to release its lock, any blocked jobs become candidates for release once
+  duration has elapsed.
 
-### Serializers
+When a job with concurrency controls is enqueued, Solid Queue checks a
+database-backed lock for the computed key. If the lock is available, the job is
+marked ready for execution. If not, the behavior depends on the `on_conflict:`
```
p8:
```suggestion
marked ready for execution. If not, the behavior depends on the `:on_conflict`
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+  a_cleanup_task:
+    command: "DeletedStuff.clear_all"
+    schedule: every day at 9am
+```
+
+Each task specifies a `class` or `command` and a `schedule` (parsed using
+[Fugit](https://github.com/floraison/fugit)). You can also pass arguments to
+jobs, such as in the example for `MyJob` where `args` are passed. This can be
+passed as a single argument, a hash, or an array of arguments that can also
+include key word arguments as the last element in the array.
```
p8:
```suggestion
include keyword arguments as the last element in the array.
```

guides/source/active_job_basics.md
```diff
@@ -747,279 +580,707 @@ class GuestsCleanupJob < ApplicationJob
...
+- [Delayed Job](https://github.com/collectiveidea/delayed_job#active-job)
+- [Que](https://github.com/que-rb/que#additional-rails-specific-setup)
+- [Good Job](https://github.com/bensheldon/good_job#readme)
+
+Monitoring and Handling Failed Jobs
+-----------------------------------
+
+### Monitoring With Mission Control
+
+The [Mission Control](https://github.com/rails/mission_control-jobs) tool is a
```
p8: Maybe "gem" or "engine" is more descriptive?
```suggestion
The [Mission Control](https://github.com/rails/mission_control-jobs) engine is a
```

guides/source/active_job_basics.md
```diff
@@ -1065,81 +1326,12 @@ end
 [`retry_on`]:
     https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
 
-### Deserialization
+### Missing Records
 
 GlobalID allows serializing full Active Record objects passed to `#perform`.
```
p8: This seems to imply the AR objects are serialized with all their attributes.
Maybe something like:
```suggestion
GlobalID will use the unique identifier to locate the full Active Record object when calling `#perform`.
```

guides/source/active_job_basics.md
```diff
@@ -1065,81 +1326,12 @@ end
 [`retry_on`]:
     https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
 
-### Deserialization
+### Missing Records
```
p8, line 1395: The previous section already mentions `ActiveJob::DeserializationError` in the `discard_on` example. Not sure if we should maybe use a different error there, or maybe we can merge these two sections?

guides/source/active_job_basics.md
```diff
@@ -65,434 +69,222 @@ class GuestsCleanupJob < ApplicationJob
...
-$ bin/jobs start
+```ruby
+GuestsCleanupJob.perform_later(guest)
 ```
 
-#### Production
-
-Solid Queue is already configured for the production environment. If you open
-`config/environments/production.rb`, you will see the following:
+To enqueue a job to be performed tomorrow at noon:
```
p8: What do you think about moving the `:wait` and `:wait_until` examples to the "Enqueuing Jobs" section, maybe in a "Enqueuing Later" (or something) section? The "Enqueuing Jobs" section already has other examples of using `set` (to set the queue name or priority).

p8 (commented):
@bhumi1102 This is looking great! I've added some final suggestions.

guides/source/active_job_basics.md
bhumi1102: I think I'll change it to Ruby as it doesn't show any irb output, so we don't need the extra prompt.

guides/source/active_job_basics.md
bhumi1102: Yes, Done (manually)!

guides/source/active_job_basics.md
bhumi1102: Thanks for catching this!

guides/source/active_job_basics.md
bhumi1102: moved and changed to TIP!

guides/source/active_job_basics.md
```diff
@@ -18,24 +19,33 @@ After reading this guide, you will know:
...
-tasks like sending emails, processing data, or handling regular maintenance
-activities, such as clean-ups and billing charges. By offloading these tasks
-from the main application thread to a queuing backend like the default Solid
-Queue, Active Job ensures that time-consuming operations do not block the
-request-response cycle. This can improve the performance and responsiveness of
-the application, allowing it to handle tasks in parallel.
+The Active Job Rails framework allows you to declare background jobs and execute
+them on a queuing backend. It provides a consistent, high-level interface for
+common asynchronous tasks such as sending emails, processing data, generating
+reports, and running periodic maintenance work like data clean-up or billing.
```
ahmad19: Opinion: How about if we merge the NOTE section into the introduction as the NOTE sections seems repeated ?


```suggestion
The Active Job Rails framework provides an interface to declare background jobs and execute them on a queuing backend (such as Solid Queue). It provides a consistent, high-level interface for common asynchronous tasks such as sending emails, processing data, generating reports, and running periodic maintenance work like data clean-up or billing.
```

guides/source/active_job_basics.md
bhumi1102: Good call! Moving the list of backends up here.

guides/source/active_job_basics.md
bhumi1102: Changed the first one to Net::OpenTimeout, as that's a better example anyway.

guides/source/active_job_basics.md
bhumi1102: Good call. Started with the simpler example without cursor. Reworded the explanation paragraphs. And added a follow-on shorter example with cursor.

guides/source/active_job_basics.md
bhumi1102: So looking at the other `set` example in the Enqueuing Jobs secation vs. the. These are about _when_ to call `perform` vs. the other one is about which queue/priority. So I like the idea of keeping these two examples here. However, I added a link to the `set` method here for reference as it's the first mention (and it has 4 different use cases actually). Let me know if this is better or if you feel there is a reason to move these still :)

guides/source/active_job_basics.md
bhumi1102: Sure, updated.

guides/source/active_job_basics.md
p8: I found the 5 single line explanations with single line examples somewhat off putting.
Explaining the `set` method breaks this and makes it easier on the eyes for me. ❤️ 
It also makes it clearer what how these examples differ from the default.

guides/source/active_job_basics.md
```diff
@@ -790,279 +629,726 @@ class GuestsCleanupJob < ApplicationJob
...
 ### Available Callbacks
 
-* [`before_enqueue`][]
-* [`around_enqueue`][]
-* [`after_enqueue`][]
-* [`before_perform`][]
-* [`around_perform`][]
-* [`after_perform`][]
-* [`after_discard`][]
+There are several other callbacks that Active Job supports.
```
p8: We just mentioned `before_enqueue` above, so maybe "other" is a bit confusing here when we start with `before_enqueue`:
```suggestion
There are several callbacks that Active Job supports.
```

guides/source/active_job_basics.md
p8: This is a big example. The reader could use some help with what to look for.
Maybe clarify what to look for in the describing paragraph or add a comment in the code:
```suggestion
    # Use the serialized payment_token and billing_profile_id
    # when resuming at the following step.
    step :submit_enrollment do
```

guides/source/active_job_basics.md
```diff
@@ -665,71 +464,111 @@ MyJob.set(priority: 50).perform_later(record)
...
+
+[Other backends](#alternate-queuing-backends) like Sidekiq have a `push_bulk`
+method, which the Sidekiq adapter users. internally to push a large number of
+jobs to Redis and prevent the round trip network latency. GoodJob also supports
+bulk enqueuing with the `GoodJob::Bulk.enqueue` method.
+
+If the queue backend does *not* support bulk enqueuing, `perform_all_later` will
+enqueue jobs one by one.
 
 ### Job Attributes
```
p8: Job attributes are mostly (only?) [relevant for continuable jobs.](https://github.com/rails/rails/pull/57208). The example also uses the `ActiveJob::Continuable` API which hasn't been mentioned here yet and is explained later in the guide.
So I think this section would fit better in the "Job Continuations" section.

bhumi1102: Good point. I moved this section to the Job Continuables. Also added a couple lines explaining what the big example is doing.

guides/source/active_job_basics.md
bhumi1102: Good to know 😊 I reformatted the code examples to put those lines inside the code block as comments. Let me know if you like it that way or want me to put it back. 

I think the `set` examples are relevant here so useful to have for the reader but totally open to how we format, no strong preference! So let me know if you want me change anything else in this section so it flows and is not off putting :)

guides/source/active_job_basics.md
p8: That section is much clearer now!

PR #57103, opened 2026-03-31
Update ActiveRecord::Associations::Preloader::Association.owners_by_key to handle composite keys

rosew:
### Motivation / Background

This Pull Request has been created because code that directly calls `ActiveRecord::Associations::Preloader` will have an extra unnecessary failed query when there is a composite key that is partially or all nil.

### Detail

This Pull Request changes `ActiveRecord::Associations::Preloader::Association.owners_by_key`

`owners_by_key` does not set a result when key is nil. However when key is composite it is an Array and passes the nil check even when the values are nil. This addes a check for Array and skips if any of the values in the array are nil. 

I am considering this a bug fix and not a change on the assumption that a query with a nil value for a composite key will always fail based on the pattern already set in ActiveRecord::Associations::BelongsToAssociation which uses: `Array(reflection.foreign_key).all?`

### Additional information

I chose to use the key.is_a?(Array) instead of casting all to an Array to preserve as close to the performance for the common path as possible.

Running a Benchmark to compare options gave the following results for 1000 iterations:
- No change:   0.177381   0.007517   0.184898 (  0.185270)
- This option (not composite):   0.177714   0.007319   0.185033 (  0.189884)
- This option (composite):   0.167168   0.009131   0.176299 (  0.181202)
- Array(key).all?:   0.262781   0.012571   0.275352 (  0.279990)

### Checklist

### DIFF
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -169,7 +169,7 @@ module ActiveRecord
         def owners_by_key
           @owners_by_key ||= owners.each_with_object({}) do |owner, result|
             key = derive_key(owner, owner_key_name)
-            (result[key] ||= []) << owner if key
+            (result[key] ||= []) << owner if key.is_a?(Array) ? key.all? : key
           end
         end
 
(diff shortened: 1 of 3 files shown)


eileencodes:
Thanks @rosew! Backported to 8-1-stable and 8-0-stable as well.

PR #44654, opened 2022-03-10
Fixed MariaDB default function

kaspernj:
### Summary

Schema doesn't dump default functions correctly when using MariaDB, making them read as strings instead of functions. This can be a problem if you have a column that should get a UUID set automatically using a default function.

PR #57110, opened 2026-03-31
Reduce number of queries during MariaDB column introspection

ilianah:
### Motivation / Background
#44654 fixed MariaDB function default detection by adding a check in `new_column_from_field` which was running `SHOW CREATE TABLE` once per column, resulting in a lot of unnecessary queries for applications with large schemas even if these applications were not running MariaDB.

This PR moves the MariaDB function default detection out of `new_column_from_field` and up into `column_definitions`. This avoids the extra queries for non-MariaDB connections, and for MariaDB issues at most one `SHOW CREATE TABLE` per `columns` call only in cases when a field's default looks like a function. 

### Detail
After issuing `SHOW FULL FIELDS`, if we're on a MariaDB connection, we check whether any field's Default looks like a function call. If at least one does, we issue a single `SHOW CREATE TABLE` to confirm, then mark those fields as `DEFAULT_GENERATED` in the metadata. This routes the fields through the existing `new_column_from_field` branch that MySQL already uses.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/mysql/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/mysql/schema_statements.rb
@@ -194,21 +194,7 @@ module ActiveRecord
             MySQL::TableDefinition.new(self, name, **options)
           end
 
-          def default_type(table_name, field_name)
-            match = create_table_info(table_name)&.match(/`#{field_name}` (.+) DEFAULT ('|\d+|[A-z]+)/)
-            default_pre = match[2] if match
-
-            if default_pre == "'"
-              :string
-            elsif default_pre&.match?(/^\d+$/)
-              :integer
-            elsif default_pre&.match?(/^[A-z]+$/)
-              :function
-            end
-          end
-
           def new_column_from_field(table_name, field, _definitions)
-            field_name = field.fetch("Field")
             type_metadata = fetch_type_metadata(field["Type"], field["Extra"])
             default, default_function = field["Default"], nil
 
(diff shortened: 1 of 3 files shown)


byroot:
The CI failure seem to be legit:

```
Failure:
--
BulkAlterTableMigrationsTest#test_changing_column_null_with_default [test/cases/migration_test.rb:1466]:
7 instead of 6 queries were executed. Queries:
 
SHOW FULL FIELDS FROM `delete_me`
 
SHOW FULL FIELDS FROM `delete_me`
 
ALTER TABLE `delete_me` CHANGE `name` `name` varchar(255) DEFAULT 'NONAME', CHANGE `birthdate` `birthdate` datetime(6) DEFAULT NULL
 
UPDATE `delete_me` SET `age`=0 WHERE `age` IS NULL
 
SHOW FULL FIELDS FROM `delete_me`
 
SHOW CREATE TABLE `delete_me`
 
ALTER TABLE `delete_me` CHANGE `age` `age` int(11) NOT NULL.
Expected: 6
Actual: 7
```

ilianah:
yep, there's a different number of queries now for `mariadb` vs `mysql` - I'll fix it now

activerecord/lib/active_record/connection_adapters/mysql/schema_statements.rb
```diff
@@ -195,7 +201,8 @@ def create_table_definition(name, **options)
           end
 
           def default_type(table_name, field_name)
-            match = create_table_info(table_name)&.match(/`#{field_name}` (.+) DEFAULT ('|\d+|[A-z]+)/)
+            table_info = @cached_create_table_info ||= create_table_info(table_name)
```
byroot: I'm afraid this is way too brittle. This code expect `columns` to be called for the cache to be cleared, that may be true today, but could easily change and cause a nasty bug.

I don't have a better suggestion out of the top of my head, but I'll try to dig deeper into this.

byroot: Ok, so here's what I think this patch should look like.

I think we should improve `AbstractMysqlAdapter#column_definitions`, so that if we're talking to MariaDB, we got over all defaults and if one of them could be a function, we issue a `SHOW CREATE TABLE`, and we update the initial `SHOW FULL FIELDS`.

PR #57112, opened 2026-03-31
Bumping Bundler Version to 4.0.9

adityapandit17:
Bumped bundler version since 4.0.9 is available to use 

https://github.com/ruby/rubygems/releases/tag/bundler-v4.0.9


fatkodima:
New bundler versions are released constantly - no need to bump it each time.

PR #57114, opened 2026-04-01
Document that days_in_month and days_in_year assume proleptic Gregorian calendar

hammadxcm:
### Motivation / Background

Fixes #57109

`Time.days_in_month` and `Time.days_in_year` use a simple lookup table and do not account for the historical Gregorian Calendar Reform, during which countries skipped several days when transitioning from the Julian calendar (e.g., Italy skipped 10 days in October 1582, England skipped 11 days in September 1752).

Rather than adding region-specific calendar reform support — which would be disproportionately complex for an extremely niche historical use case — this PR documents the limitation and points users to Ruby's `Date` class for historical date calculations.

### Detail

Added RDoc documentation to both `Time.days_in_month` and `Time.days_in_year` noting that:
- These methods assume the proleptic Gregorian calendar
- They do not account for the historical Gregorian Calendar Reform
- Users should use Ruby's `Date` class directly for historical dates

### Additional information

```ruby
# Current behavior (unchanged):
Time.days_in_month(10, 1582) # => 31

# Ruby core Date accounts for the reform:
(Date.new(1582, 10, 1)..Date.new(1582, 10, -1)).count # => 21
```

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/time/calculations.rb
+++ b/activesupport/lib/active_support/core_ext/time/calculations.rb
@@ -21,6 +21,14 @@ class Time
 
     # Returns the number of days in the given month.
     # If no year is specified, it will use the current year.
+    #
+    # This method assumes the proleptic Gregorian calendar and does not
+    # account for the historical Gregorian Calendar Reform. During the
+    # reform, countries skipped several days (e.g., Italy skipped 10 days
+    # in October 1582, England skipped 11 days in September 1752). For
+    # such historical dates, use Ruby's +Date+ class directly instead:
+    #
+    #   (Date.new(1582, 10, 1)..Date.new(1582, 10, -1)).count  # => 21
     def days_in_month(month, year = current.year)
       if month == 2 && ::Date.gregorian_leap?(year)
         29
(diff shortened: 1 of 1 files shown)


activesupport/lib/active_support/core_ext/time/calculations.rb
```diff
@@ -21,6 +21,14 @@ def ===(other)
 
     # Returns the number of days in the given month.
     # If no year is specified, it will use the current year.
+    #
+    # This method assumes the proleptic Gregorian calendar and does not
+    # account for the historical Gregorian Calendar Reform. During the
+    # reform, countries skipped several days (e.g., Italy skipped 10 days
+    # in October 1582, England skipped 11 days in September 1752). For
+    # such historical dates, use Ruby's +Date+ class directly instead:
```
broekhuis00: Explaining _what_ the reform was might be unnecessary.
Only '_This method assumes the proleptic Gregorian calendar and does not account for the historical Gregorian Calendar Reform_', a link to an external resource, and the example would suffice IMO.

hammadxcm: Good point — trimmed the inline explanation and replaced it with a link to the Wikipedia article. Thanks for the feedback!

broekhuis00 (commented):
Looks good to me.
I cannot approve it however.

PR #57118, opened 2026-04-01
Enable models to connect/query different adapters.

hmcguire-shopify:
Previously, models cached many adapter specific values which would prevent making queries against different database types. Now, those caches are per `schema_context`, which can be configured in `database.yml`:

```yaml
primary:
  adapter: trilogy
  database: myapp_development
primary_pg:
  adapter: postgresql
  database: myapp_development
  schema_context: pg
```

This enables models to connect to both MySQL and PostgreSQL databases in the same application, and query them successfully:

```ruby
  connects_to shards: {
    mysql: { writing: :primary, reading: :primary },
    pg: { writing: :primary_pg, reading: :primary_pg },
  }
end

User.connected_to(shard: :mysql) { User.first } # queries MySQL
User.connected_to(shard: :pg) { User.first } # queries PostgreSQL
```

### DIFF
--- a/activemodel/lib/active_model/attribute_registration.rb
+++ b/activemodel/lib/active_model/attribute_registration.rb
@@ -55,6 +55,16 @@ module ActiveModel
         end
       end
 
+      def apply_pending_attribute_modifications(attribute_set) # :nodoc:
+        if superclass.respond_to?(:apply_pending_attribute_modifications, true)
+          superclass.send(:apply_pending_attribute_modifications, attribute_set)
+        end
+
+        pending_attribute_modifications.each do |modification|
+          modification.apply_to(attribute_set)
+        end
+      end
+
       private
         PendingType = Struct.new(:name, :type) do # :nodoc:
           def apply_to(attribute_set)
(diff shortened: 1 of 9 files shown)


eileencodes:
I'm still stuck on `schema_context`. I understand that each different adapter needs a different schema, but as a user of a Rails app, it isn't clear what I would set this to.

Is the reason we can't derive this from the adapter name because you might have 2 connections with the same adapter but different schema shapes?

If that's the case can we default to auto setting this value and have the config option as an escape hatch? It feels off I would need to set this for mysql vs postgresql schema contexts when Rails can derive it.

We also need to document the behavior of this config option somewhere.

andrewn617:
I agree with @eileencodes, I think we should have the db adapter name be the default for the schema context key. We know different adapters cannot share a schema context (thus the feature), so it seems to make sense. If I have an app with a mysql shard and pg shard, what do I make the key other than `:mysql` or `:pg`? Of course we can still allow it to be set explicitly for more advanced use cases.

matthewd:
My theory behind specifying a static default for the context key was 1) maximum backwards compatibility [because we match the current behaviour], and 2) the possibility that people might be using "variant" adapters that are [partially?] inter-compatible -- I don't know if it might currently be possible to mix `mysql2` and `trilogy` connections for some limited set of operations, for example, or `postgresql` and `postgis`.

If we were designing from scratch, I might make it default to the connection name or something, such that it was _always_ unique and you had to override to activate any sharing at all... but that would be troublesome for simple sharding cases that work well today.

I agree it seems fine to default to the adapter type, and let my above very-theoretical currently-adapter-mixing user override that default to get back their current behaviour.

... although this might need consideration: https://github.com/rails/rails/pull/57118/changes#diff-eee8446909a6221877323c7b605b430eaf515d8e66aa27ae62de2e7614014c95R206

eileencodes:
What if instead of defaulting to adapter type we default to database type? So then mysql2 and trilogy are `mysql` and postgres and postgis are `postgres`. That maintains backwards compatibility if someone is doing that but doesn't mean having to understand what "default" means.

hmcguire-shopify:
Addressed by #58437

PR #57120, opened 2026-04-01
railties: report missing model on destroy

### LINKED ISSUE #52860: [Bug] [ActiveSupport] "rails destroy model Adress" [sic] does not check first if the model exists
### Steps to reproduce

When I run the script `rails destroy model Adress` [sic] in the CLI, it prints that the model and associated files were removed, even though no such files existed. This behavior could cause confusion when a user makes a typo, as it may be misleading. While there is a robust `check_class_collision` callback when generating objects, a similar approach could be used when reverting the generate command.

Perhaps the `rails destroy ...` command is designed intentionally to assume the developer's responsibility and avoid complex checks, but I wanted to raise it as an issue nonetheless.

```ruby
my-code/my-repo (main*) $ rails destroy model Adress      
      invoke  active_record
      remove    db/migrate/20240910091915_create_adresses.rb
      remove    app/models/adress.rb
      invoke    rspec
      remove      spec/models/adress_spec.rb
      invoke      factory_bot
      remove        spec/factories/adresses.rb
my-code/my-repo (main*) $ 
```

### Expected behavior
If there is a typo in the model name, I should be alerted that no files were removed with that name, or an error should be returned, such as `"Did you mean model name Address?"`

### Actual behavior
When I run the script `rails destroy model Adress` [sic] in the CLI, it prints that the model and associated files were removed, even though no files existed.

### System configuration
**Rails version**:
7.2.1
**Ruby version**:
3.1.2p20

wwenrr:
This addresses misleading output from `rails destroy model` when the model name is a typo.

## Problem

Today, running:

```bash
bin/rails destroy model Adress
```

prints `remove` lines even if `app/models/adress.rb` does not exist. That suggests files were removed when nothing actually happened.

## Change

In `Rails::Command::DestroyCommand`, add a guard for `destroy model`:

- check whether `app/models/<name>.rb` exists
- if missing, print a clear error and exit with status `1`
- if present, keep existing revoke flow unchanged

## Tests Added

`railties/test/commands/destroy_test.rb`

1. `test_destroy_model_reports_missing_model_file`
   - verifies typo case (`Adress`) now reports missing model and exits `1`
2. `test_destroy_model_still_removes_existing_model`
   - verifies normal case (`Book`) still removes generated model file

Closes #52860.

### Environment
- Docker image: `ruby:3.3`

### Command

```bash
docker run --rm -v "$PWD":/rails -w /rails ruby:3.3 bash -lc \
  'bundle config set without "db job cable storage ujs" >/dev/null 2>&1 || true; \
   bundle install -j4 >/tmp/bundle.log && \
   bundle exec ruby -Irailties/test -Itest railties/test/commands/destroy_test.rb'
```

### Result

- `2 runs, 7 assertions, 0 failures, 0 errors, 0 skips`
- behavior verified:
  - typo model name now returns a clear message + non-zero exit (`1`)
  - existing model still follows normal destroy path

### DIFF
--- a/railties/lib/rails/commands/destroy/destroy_command.rb
+++ b/railties/lib/rails/commands/destroy/destroy_command.rb
@@ -22,6 +22,17 @@ module Rails
         boot_application!
         load_generators
 
+        if generator == "model" && args.first
+          model_name = args.first
+          model_path = model_name.underscore
+          model_file = File.join(Rails::Command.root, "app/models/#{model_path}.rb")
+
+          unless File.exist?(model_file)
+            puts "Could not find model '#{model_name}' (expected #{model_file}). Nothing was removed."
+            exit(1)
+          end
+        end
+
         Rails::Generators.invoke generator, args, behavior: :revoke, destination_root: Rails::Command.root
       end
     end
--- a/railties/test/commands/destroy_test.rb
+++ b/railties/test/commands/destroy_test.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require "isolation/abstract_unit"
+require "rails/command"
+
+class Rails::DestroyTest < ActiveSupport::TestCase
+  include ActiveSupport::Testing::Isolation
+
+  setup :build_app
+  teardown :teardown_app
+
+  def test_destroy_model_reports_missing_model_file
+    command_output = rails("destroy", "model", "Adress", allow_failure: true)
+
+    assert_match(/Could not find model 'Adress'/, command_output)
+    assert_equal 1, $?.exitstatus
+  end
+
+  def test_destroy_model_still_removes_existing_model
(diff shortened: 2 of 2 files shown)


wwenrr:
Re-ran the exact verification command from the PR description and confirmed the result:

```bash
docker run --rm -v "$PWD":/rails -w /rails ruby:3.3 bash -lc \
 'bundle config set without "db job cable storage ujs" >/dev/null 2>&1 || true; \
 bundle install -j4 >/tmp/bundle.log && \
 bundle exec ruby -Irailties/test -Itest railties/test/commands/destroy_test.rb'
```

Output:

```text
Run options: --seed 53710

# Running:

..

Finished in 1.236545s, 1.6174 runs/s, 5.6609 assertions/s.

2 runs, 7 assertions, 0 failures, 0 errors, 0 skips
```

PR #57123, opened 2026-04-01
Add durable named default scopes

andrewn617:
### Motivation / Background

`unscoped` is a sharp knife. Sometimes you have a default scope that is important for performance and you'd like to explicitly opt out of it rather than inadvertently remove it with an `unscoped` call. This PR introduces durable named default scopes that need to explicitly be opted out of when unscoping.

### Detail

Default scopes can now be given a name, making them durable. They will not be removed by `unscoped` unless explicitly referenced by name.

```ruby
class Article < ActiveRecord::Base
  default_scope :published, -> { where(published: true) }
  default_scope :user, -> { where(user: User.current) }
end

Article.unscoped.all # Named scopes are preserved
# SELECT * FROM articles WHERE published = true AND user_id = 1

Article.unscoped(:published).all # Only :published is removed
# SELECT * FROM articles WHERE user_id = 1

Article.unscoped(:published, :user).all # Both removed
# SELECT * FROM articles

Article.unscoped(:published).unscoped(:user).all # Both removed
# SELECT * FROM articles
```

**Update**: I made quite a few improvements to this branch based on a thorough review by @matthewd.

I added some guards to raise if we try to unscope non-symbol/string names, or names that don't correspond to default scopes. Also renamed the `relation` method to `raw_relation`. Since it is now non-private (but nodoc) this makes the intention of the method clearer.

Mathew identified a defect where `reload` picked up non-all queries scopes from the current scope. I opened a separate PR https://github.com/rails/rails/pull/57594 though this branch is based on it since it needs that fix to work.

We fixed reloading behaviour to account for named scopes. `reload(unscoped: true)` will not unscope named scopes. Now you can do `reload(unscoped: :foo)` to unscope the `:foo` named scope, or `reload(unscoped: [:foo, true]`) to unscope both `foo` and the unnamed scopes.

We also introduced the `all_queries_scopes` method that returns a relation with all named and unnamed all queries default scopes, as this was needed for the reloading implementation. We also audited places that call `unscope` but should be respecting all-queries scopes. We added the `all_queries_scopes` method to uniqueness validations, fixture loading and updating/reseting counter caches.

### Additional information

I noticed the internal method `build_defualt_scope` can return `nil` which callers have to guard against by doing `build_default_scope || relation`. But `build_default_scope` knows about `relation` so it can just return it and simplify the callers. I put this refactoring on a separate commit tho, I can remove it if people have objections.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -376,6 +376,38 @@
 
     *Ryosuke Okazuka*
 
+*   Add named default scopes.
+
+    Default scopes can now be given a name, making them durable. They
+    will not be removed by `unscoped` unless explicitly referenced by name.
+
+    Calling `unscoped` with no arguments removes only unnamed default
+    scopes. Calling `unscoped(:name, ...)` removes only the listed named
+    scopes and leaves unnamed default scopes intact. You can chain multiple
+    calls together.
+
+    ```ruby
+    class Article < ActiveRecord::Base
+      default_scope { where(visible: true) }
+      default_scope :published, -> { where(published: true) }
+      default_scope :user, -> { where(user: User.current) }
+    end
+
+    Article.unscoped.all # Only the unnamed default scope is removed
+    # SELECT * FROM articles WHERE published = true AND user_id = 1
+
+    Article.unscoped(:published).all # Only :published is removed
+    # SELECT * FROM articles WHERE visible = true AND user_id = 1
+
+    Article.unscoped(:published, :user).all # Both named scopes removed
+    # SELECT * FROM articles WHERE visible = true
+
+    Article.unscoped(:published).unscoped.all # :published and unnamed removed
+    # SELECT * FROM articles WHERE user_id = 1
+    ```
+
+    *Andrew Novoselac*
+
 *   Report PostgreSQL default timestamp and time precision as 6.
 
     Bare PostgreSQL `timestamp` and `time` columns now use their effective
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,44 @@
+*   Add named default scopes.
+
+    Default scopes can now be given a name, making them durable. They
+    will not be removed by `unscoped` unless explicitly referenced by name.
+
+    Calling `unscoped` with no arguments removes only unnamed default
+    scopes. Calling `unscoped(:name, ...)` removes only the listed named
+    scopes and leaves unnamed default scopes intact. You can chain multiple
+    calls together.
+
+    ```ruby
+    class Article < ActiveRecord::Base
+      default_scope { where(visible: true) }
+      default_scope :published, -> { where(published: true) }
+      default_scope :user, -> { where(user: User.current) }
+    end
+
+    Article.unscoped.all # Only the unnamed default scope is removed
+    # SELECT * FROM articles WHERE published = true AND user_id = 1
+
+    Article.unscoped(:published).all # Only :published is removed
+    # SELECT * FROM articles WHERE visible = true AND user_id = 1
+
+    Article.unscoped(:published, :user).all # Both named scopes removed
+    # SELECT * FROM articles WHERE visible = true
+
+    Article.unscoped(:published).unscoped.all # :published and unnamed removed
+    # SELECT * FROM articles WHERE user_id = 1
+    ```
+
+    Reload only considers default scopes marked with `all_queries: true`. Named
+    all-query default scopes survive `reload(unscoped: true)`; pass a name to
+    remove one, or include literal `true` in an array of names to also remove
+    unnamed all-query default scopes.
+
+    Internal operations that bypass ordinary visibility scopes, including
+    uniqueness validation, fixture lookup, and counter maintenance, continue
+    to apply all-query default scopes.
+
+    *Andrew Novoselac* and *Matthew Draper*
+
 *   Add `config.active_record.schema_ignored_tables` to exclude tables from both the
     schema cache and the schema file.
 
@@ -375,38 +416,6 @@
     other spatial types (`GEOMETRY`, `POLYGON`, `LINESTRING`, ...).
 
     *Ryosuke Okazuka*
-
-*   Add named default scopes.
-
-    Default scopes can now be given a name, making them durable. They
-    will not be removed by `unscoped` unless explicitly referenced by name.
-
-    Calling `unscoped` with no arguments removes only unnamed default
-    scopes. Calling `unscoped(:name, ...)` removes only the listed named
-    scopes and leaves unnamed default scopes intact. You can chain multiple
-    calls together.
-
-    ```ruby
-    class Article < ActiveRecord::Base
-      default_scope { where(visible: true) }
-      default_scope :published, -> { where(published: true) }
-      default_scope :user, -> { where(user: User.current) }
-    end
-
-    Article.unscoped.all # Only the unnamed default scope is removed
-    # SELECT * FROM articles WHERE published = true AND user_id = 1
-
-    Article.unscoped(:published).all # Only :published is removed
-    # SELECT * FROM articles WHERE visible = true AND user_id = 1
-
-    Article.unscoped(:published, :user).all # Both named scopes removed
-    # SELECT * FROM articles WHERE visible = true
-
-    Article.unscoped(:published).unscoped.all # :published and unnamed removed
-    # SELECT * FROM articles WHERE user_id = 1
-    ```
-
-    *Andrew Novoselac*
 
 *   Report PostgreSQL default timestamp and time precision as 6.
 
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -261,6 +261,38 @@
 
     *Ryosuke Okazuka*
 
+*   Add named default scopes.
+
+    Default scopes can now be given a name, making them durable. They
+    will not be removed by `unscoped` unless explicitly referenced by name.
+
+    Calling `unscoped` with no arguments removes only unnamed default
+    scopes. Calling `unscoped(:name, ...)` removes only the listed named
+    scopes and leaves unnamed default scopes intact. You can chain multiple
+    calls together.
+
+    ```ruby
+    class Article < ActiveRecord::Base
+      default_scope { where(visible: true) }
+      default_scope :published, -> { where(published: true) }
+      default_scope :user, -> { where(user: User.current) }
+    end
+
+    Article.unscoped.all # Only the unnamed default scope is removed
+    # SELECT * FROM articles WHERE published = true AND user_id = 1
+
+    Article.unscoped(:published).all # Only :published is removed
+    # SELECT * FROM articles WHERE visible = true AND user_id = 1
+
+    Article.unscoped(:published, :user).all # Both named scopes removed
+    # SELECT * FROM articles WHERE visible = true
+
+    Article.unscoped(:published).unscoped.all # :published and unnamed removed
+    # SELECT * FROM articles WHERE user_id = 1
+    ```
+
+    *Andrew Novoselac*
+
 *   Report PostgreSQL default timestamp and time precision as 6.
 
     Bare PostgreSQL `timestamp` and `time` columns now use their effective
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,43 @@
+*   Add named default scopes.
+
+    Default scopes can now be given a name, making them durable. They
+    will not be removed by `unscoped` unless explicitly referenced by name.
+
+    Calling `unscoped` with no arguments removes only unnamed default
+    scopes. Calling `unscoped(:name, ...)` removes only the listed named
+    scopes and leaves unnamed default scopes intact. You can chain multiple
+    calls together.
+
+    ```ruby
+    class Article < ActiveRecord::Base
+      default_scope { where(visible: true) }
+      default_scope :published, -> { where(published: true) }
+      default_scope :user, -> { where(user: User.current) }
+    end
+
+    Article.unscoped.all # Only the unnamed default scope is removed
+    # SELECT * FROM articles WHERE published = true AND user_id = 1
+
+    Article.unscoped(:published).all # Only :published is removed
+    # SELECT * FROM articles WHERE visible = true AND user_id = 1
+
+    Article.unscoped(:published, :user).all # Both named scopes removed
+    # SELECT * FROM articles WHERE visible = true
+
+    Article.unscoped(:published).unscoped.all # :published and unnamed removed
+    # SELECT * FROM articles WHERE user_id = 1
+    ```
+
+    Reload only considers default scopes marked with `all_queries: true`. Named
+    all-query default scopes survive `reload(unscoped: true)`; pass a name to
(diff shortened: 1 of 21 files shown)


activerecord/lib/active_record/scoping/default.rb
```diff
@@ -47,8 +55,33 @@ module ClassMethods
...
+        #   Article.unscoped(:published).all # Only :published is removed
+        #   # SELECT * FROM articles WHERE rating = 'G'
+        #
+        #   Article.unscoped(:published, :rating).all # Both removed
+        #   # SELECT * FROM articles
+        def unscoped(*names, &block)
+          existing = current_scope&.excluded_default_scope_names || []
+          names = (existing + names.map(&:to_sym)).uniq
+          relation_with_named_defaults = build_default_scope(unscoped_names: names)
+          relation_with_named_defaults.excluded_default_scope_names = names
```
matthewd: Is this always a new instance that we're allowed to mutate?

matthewd (commented):
This looks good thanks! It sounds like it matches what I had in mind.

One thing we should make a decision on, and then clearly reflect in the documentation, is whether `unscoped(:foo)` also removes standard/anonymous default scopes.[^1]

[^1]: My quick read of the code had me thinking the current state was "no", but from the tests it sounds like it's actually "yes"... tbh I'm not sure which I prefer -- both have good arguments. 🤔 

I'm not sure off the top of my head, but please check whether we need to consider this state in `structurally_incompatible_values_for`. (Maybe its effects are already applied to the "real" values so it doesn't matter?)

Likewise `Relation::Merger`.

... which potentially circles around a more general question of "should this be in `MULTI_VALUE_METHODS`?"

activerecord/lib/active_record/scoping/default.rb
andrewn617: Yes, unless we pass a relation to this method.

andrewn617:
@matthewd After thinking this over and pairing with @adrianna-chang-shopify I landed on wanting `unscoped(:foo)` to NOT clear unnamed default scopes. I think it is preferable because if I have some existing load bearing unnamed default, like a soft delete scope, I don't want to have to unapply it just because I want to unapply my new named default scope. I think its workable both ways but I think this is slightly better behaviour.

I accomplished that by refactoring the `excluded_default_scope_names` to be a list of scope objects (`excluded_default_scopes`). That way we can track the unnamed scopes there and not introduce yet another piece of state.

> I'm not sure off the top of my head, but please check whether we need to consider this state in structurally_incompatible_values_for. (Maybe its effects are already applied to the "real" values so it doesn't matter?)
> 
> Likewise Relation::Merger.

I don't think anything is needed here. These methods just care about `where_values_hash`, `joins_hash` etc. and there is no `unscoped_values` since those are not values in themselves. So doing this still works as expect:

```
class Foo
  default_scope :bar, -> { joins(:bars) }
end

Foo.unscoped(:bar).structurally_compatible?(Foo.all)
```

Let me know if I missing the thrust of your question tho, but from reading the code and playing with console I think this is all fine.

activerecord/lib/active_record/scoping/default.rb
```diff
@@ -47,8 +56,45 @@ module ClassMethods
...
+        def unscoped(*names, &block)
+          unscoped_relation = if default_scope_override?
+            relation
+          else
+            scopes_to_exclude = if names.empty?
+              default_scopes.reject(&:named?)
+            else
+              names = names.map(&:to_sym)
+              default_scopes.select { |s| names.include?(s.name) }
+            end
```
nvasilevski, line 95: `default_scopes.reject(&:named?)` and `default_scopes.select { |s| names.include?(s.name) }` may show up as wasteful in a microbenchmark, but I’d expect them to be fine in real-world usage given the usual number of default scopes per model.  If this ever does become measurable, we could keep unnamed default scopes in a separate array and named default scopes in a hash keyed by name, so bare `unscoped` and `unscoped(:name)` don’t need to scan the full list

nvasilevski (approved):
Non-blocking: with this change, I think we no longer have a single API that means “remove all default scopes, named or unnamed, no questions asked” unless callers explicitly enumerate the named scopes.
That seems fine for the intended durable-scope semantics, and I don’t think it should block this PR. But someone may reasonably want a development/debugging escape hatch along the lines of `bare_scope` / `without_default_scopes` that removes every default scope regardless of durability. It doesn't necessarily belong to this PR, but worth being conscious that `unscoped` no longer fills that role once named default scopes exist

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,35 @@
...
+    Article.unscoped.all # Only the unnamed default scope is removed
+    # SELECT * FROM articles WHERE published = true AND user_id = 1
+
+    Article.unscoped(:published).all # Only :published is removed
+    # SELECT * FROM articles WHERE visible = true AND user_id = 1
+
+    Article.unscoped(:published, :user).all # Both named scopes removed
+    # SELECT * FROM articles WHERE visible = true
+
+    Article.unscoped(:published).unscoped.all # Named and unnamed removed
```
adrianna-chang-shopify:
```suggestion
    Article.unscoped(:published).unscoped.all # :published and unnamed removed
```

activerecord/lib/active_record/scoping/default.rb
```diff
@@ -137,26 +211,26 @@ def default_scope(scope = nil, all_queries: nil, &block) # :doc:
...
-            if default_scope_override.nil?
-              self.default_scope_override = !Base.is_a?(method(:default_scope).owner)
-            end
-
-            if default_scope_override
-              # The user has defined their own default scope method, so call that
+            if default_scope_override?
               evaluate_default_scope do
                 relation.scoping { default_scope }
               end
```
adrianna-chang-shopify, line 267: Do we need a fallback here too of `|| relation`? Before we were doing `build_default_scope(...) || scope` inside `AR::Scoping::Named.default_scoped`, which would handle a scenario where a user-defined `self.default_scope` method returns `nil`...

activerecord/lib/active_record/scoping/named.rb
```diff
@@ -43,15 +43,11 @@ def scope_for_association(scope = relation) # :nodoc:
 
         # Returns a scope for the model with default scopes.
         def default_scoped(scope = relation, all_queries: nil)
-          build_default_scope(scope, all_queries: all_queries) || scope
+          build_default_scope(scope, all_queries: all_queries)
```
adrianna-chang-shopify, line 56: Do we need to pass the current relation's exclusions here?
```suggestion
          build_default_scope(scope, all_queries: all_queries, excluded: Array(current_scope&.excluded_default_scopes))
```

We go through this code path for `#scope_for_association`, e.g. consider this example:
```ruby
  belongs_to :user
  default_scope(:tenant) { where(tenant_id: Current.tenant_id) }
end

  has_many :posts
end
```

I assume we'd want the removal of the named scope to apply to the association? e.g.
```ruby
Post.unscoped(:tenant) do
  user.posts.to_sql
end
```
=> `SELECT * FROM posts WHERE user_id = 1` (tenant scoping should be removed)

adrianna-chang-shopify: (It feels like we may also need further test coverage around named default scopes + associations / joins / preloading ?)

adrianna-chang-shopify (commented):
I think there may be some edge cases around associations that are missing here. I called out one [here](https://github.com/rails/rails/pull/57123/changes#r3306599243), but I think this would also be an issue:

```ruby
  belongs_to :author
  default_scope -> { where(published: true) }, name: :published
end

  has_many :posts
end
```

And then `Post.unscoped { author.posts.to_a }` => we'd expect the named default scope (`:published`) to be kept, but I don't think it will be:
- Association loading calls  `activerecord/lib/active_record/associations/association.rb:315` → `klass.scope_for_association`.
 - `activerecord/lib/active_record/scoping/named.rb:36-43` checks `current_scope&.empty_scope?`.
- `empty_scope?` at  activerecord/lib/active_record/relation.rb:1338-1340 is
`@values == model.unscoped.values` (both carry the named default, so `empty_scope?` returns `true`)
- `scope_for_association` returns the raw `relation` WITHOUT the `:published` scope applied  

(The preloader code in `activerecord/lib/active_record/associations/preloader/association.rb` does a similar check on `empty_scope?`)

I think the fix at least is simple, basically we need to not classify durable named defaults as empty scopes.

activerecord/lib/active_record/scoping/default.rb
andrewn617: Yes! I think I can simplify it by dropping the `else` condition and just putting `|| relation` after the if statement to cover all cases.

activerecord/lib/active_record/scoping/named.rb
andrewn617: Nice catch thank you. I see what you are saying. I believe I can fix it by making `build_default_scope` responsible for getting the `current_scope.excluded_default_scopes` rather than each caller needing to know to do it and potentially missing it.

However, I am not following your other comment:

> And then Post.unscoped { author.posts.to_a } => we'd expect the named default scope (:published) to be kept, but I don't think it will be:

I don't see that behaviour, it looks like it carries the named scopes through fine. The name `empty_scope` is a bit misleading now since it just means "do we match what we do when we call `unscoped`" 🤔 . I will think about how to improve that. But I do believe it is working, I wrote a test to show it.

adrianna-chang-shopify: > But I do believe it is working, I wrote a test to show it.

Ah, yep you're totally right -- `#scope_for_association` _does_ return the bare `klass.relation` (without the named default scope applied), but I missed the fact that we later merge `association_scope` back in, which starts from `klass.unscoped` and thus does have the named default scope:
https://github.com/rails/rails/blob/f9c1cee7ed4505d72d4482cbb9e16a82e509ccbb/activerecord/lib/active_record/associations/association.rb?plain=1#L115

The preloading case appears to still be an issue though:

```ruby
  def test_unscoped_block_does_not_remove_names_when_preloading_association
    mentor = MentorWithNamedScopeDeveloper.create!(id: 1, name: "David")
    developer = DeveloperWithNamedDefaultScopes.create!(name: "David", mentor_id: mentor.id, firm_id: 1)
    DeveloperWithNamedDefaultScopes.create!(name: "Jamis", mentor_id: mentor.id, firm_id: 2)

    preloaded_mentor = DeveloperWithNamedDefaultScopes.unscoped do
      MentorWithNamedScopeDeveloper.where(id: mentor.id).preload(:developers).first
    end

    assert_equal [developer], preloaded_mentor.developers
  end
```

```
# Running:

F

Failure:
DefaultScopingTest#test_unscoped_block_does_not_remove_names_when_preloading_association [test/cases/scoping/default_scoping_test.rb:794]:
--- expected
+++ actual
@@ -1 +1 @@
-[#<DeveloperWithNamedDefaultScopes id: 12, name: "David", first_name: nil, salary: 70000, firm_id: 1, mentor_id: 1, legacy_created_at: nil, legacy_updated_at: nil, legacy_created_on: nil, legacy_updated_on: nil>]
+#<ActiveRecord::Associations::CollectionProxy [#<DeveloperWithNamedDefaultScopes id: 12, name: "David", first_name: nil, salary: 70000, firm_id: 1, mentor_id: 1, legacy_created_at: nil, legacy_updated_at: nil, legacy_created_on: nil, legacy_updated_on: nil>, #<DeveloperWithNamedDefaultScopes id: 13, name: "Jamis", first_name: nil, salary: 70000, firm_id: 2, mentor_id: 1, legacy_created_at: nil, legacy_updated_at: nil, legacy_created_on: nil, legacy_updated_on: nil>]>



bin/test test/cases/scoping/default_scoping_test.rb:785
```

AFAICT because we explicitly skip merging in the reflection scope in the preloader if we think it's empty:
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/preloader/association.rb#L301

Should `empty_scope` bypass `unscoped` and use the base relation instead, ie. `@values == model.send(:relation).values`?

activerecord/test/cases/scoping/default_scoping_test.rb
```diff
@@ -708,6 +708,79 @@ def test_with_abstract_class_scope_should_be_executed_in_correct_context
...
+  end
+
+  def test_unscoped_block_preserves_named_default_scopes
+    wheres = DeveloperWithNamedDefaultScopes.unscoped { DeveloperWithNamedDefaultScopes.all }.where_values_hash
+    assert_not_includes wheres, "salary"
+    assert_includes wheres, "mentor_id"
+    assert_includes wheres, "firm_id"
+  end
+
+  def test_unscoped_block_with_names_removes_specified_named_default_scopesc
```
adrianna-chang-shopify:
```suggestion
  def test_unscoped_block_with_names_removes_specified_named_default_scopes
```

activerecord/lib/active_record/scoping/named.rb
andrewn617: Ah yeah that's what didn't click from your other comment, the preload case _is_ different. Your proposed fix looks good.

This ties into @nvasilevski's question about needing a "really unscope everything" method, but I don't know if we should introduce it, I will just use `relation` for this use case.

activerecord/lib/active_record/scoping/default.rb
```diff
@@ -126,8 +172,36 @@ def default_scopes?(all_queries: false)
...
+          # See +unscoped+ for more details on removing named default scopes.
+          def default_scope(name_or_scope = nil, scope = nil, all_queries: nil, &block) # :doc:
+            if name_or_scope.is_a?(Symbol) || name_or_scope.is_a?(String)
+              name = name_or_scope.to_sym
+              scope = block if block_given?
+            else
+              name = nil
+              scope = name_or_scope
+              scope = block if block_given?
+            end
```
adrianna-chang-shopify, line 243: Nit, worth dedup'ing this?
```suggestion
            if name_or_scope.is_a?(Symbol) || name_or_scope.is_a?(String)
              name = name_or_scope.to_sym
            else
              name = nil
              scope = name_or_scope
            end
            
            scope = block if block_given?
```

adrianna-chang-shopify (approved):

activerecord/lib/active_record/scoping/default.rb
```diff
@@ -3,11 +3,16 @@
...
   module Scoping
     class DefaultScope # :nodoc:
-      attr_reader :scope, :all_queries
+      attr_reader :scope, :all_queries, :name
 
-      def initialize(scope, all_queries = nil)
+      def initialize(scope, all_queries = nil, name = nil)
         @scope = scope
         @all_queries = all_queries
+        @name = name&.to_sym
```
rafaelfranca: Should we call to_sym? I'd prefer if we raise when it isn't a symbol or do nothing

rafaelfranca (approved):

activerecord/test/cases/scoping/named_scoping_test.rb
```diff
@@ -356,7 +356,7 @@ def pro; end
 
     conflicts = [
       :create,        # public class method on AR::Base
-      :relation,      # private class method on AR::Base
+      :raw_relation,  # private class method on AR::Base
```
matthewd: Maybe `:type_condition`? `raw_relation` isn't private any more, so a simple rename doesn't maintain the test coverage this wants.

PR #57124, opened 2026-04-01
Add request.safe_method? and request.unsafe_method?

thehale:
### Motivation / Background

Controllers and middleware often branch on whether a request might modify server state. Naming these predicates after RFC 9110 §9.2.1 gives that check a standard vocabulary instead of ad-hoc `get? || head?` scattered across call sites.

### Detail

Nothing beyond the title/motivation.

### Additional information

Spec Reference: https://httpwg.org/specs/rfc9110.html#safe.methods

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/http/request.rb
+++ b/actionpack/lib/action_dispatch/http/request.rb
@@ -474,6 +474,17 @@ module ActionDispatch
       authorization.to_s[/\ABearer (.+)\z/, 1]
     end
 
+    # True if the request method is safe per RFC 9110 §9.2.1
+    # (GET, HEAD, OPTIONS, or TRACE).
+    def safe_method?
+      get? || head? || options? || trace?
+    end
+
+    # True if the request method may modify resources. Inverse of #safe_method?.
+    def unsafe_method?
+      !safe_method?
+    end
+
     # True if the request came from localhost, 127.0.0.1, or ::1.
     def local?
       LOCALHOST.match?(remote_addr) && LOCALHOST.match?(remote_ip)
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Add `ActionDispatch::Request#safe_method?` and `#unsafe_method?`.
+
+    `safe_method?` returns true for HTTP methods that are defined as safe per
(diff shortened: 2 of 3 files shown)


byroot:
I like the general idea, but IMO:

  - `request.safe?` isn't super clear at first sight. `request.safe_method?` is likely a better name.
  - I can't see references to `mutative` in that RFC, so I think `request.unsafe_method?` is probably better for the inverse.

thehale:
No objections to those names. Updated accordingly!

flavorjones:
It may be worth linking to https://github.com/rails/rails/pull/54891 for a previous similar PR that attempted to do too much and was not merged. I think this is a nice small first step to encapsulate this concept from the RFC.

actionpack/lib/action_dispatch/http/request.rb
```diff
@@ -474,6 +474,17 @@ def bearer_token
       authorization.to_s[/\ABearer (.+)\z/, 1]
     end
 
+    # True if the request method is safe per RFC 9110 §9.2.1
+    # (GET, HEAD, OPTIONS, or TRACE).
+    def safe_method?
+      get? || head? || options? || trace?
```
byroot, line 480: We might want to add `QUERY` to the list at some point, but I don't think I've ever seen it in the while yet, so no rush.

thehale: <sub><em>Making some notes for future me.</em></sub>

The linked spec for this PR doesn't include QUERY as an explicitly listed safe method.

> Of the request methods defined by this specification, the [GET](https://httpwg.org/specs/rfc9110.html#GET), [HEAD](https://httpwg.org/specs/rfc9110.html#HEAD), [OPTIONS](https://httpwg.org/specs/rfc9110.html#OPTIONS), and [TRACE](https://httpwg.org/specs/rfc9110.html#TRACE) methods are defined to be safe. [[source](https://httpwg.org/specs/rfc9110.html#safe.methods:~:text=Of%20the%20request%20methods%20defined%20by%20this%20specification%2C%20the%20GET%2C%20HEAD%2C%20OPTIONS%2C%20and%20TRACE%20methods%20are%20defined%20to%20be%20safe.)]


But the [draft spec](https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html) for QUERY indicates that it will be a new safe method once fully standardized.

PR #57125, opened 2026-04-02
Avoid renaming database_tasks: false configs for parallel tests

### LINKED ISSUE #57068: Parallel tests rename `database_tasks: false` databases, causing NoDatabaseError
**I discovered this while attempting to migrate from 8.0 to 8.1**

EDIT: This issue seems to be **strongly related** but not identical, I'm getting way out of my comfort zone, so forgive me if I'm wasting somebody's time: https://github.com/rails/rails/issues/56187

### Steps to reproduce

With a multi-database setup where one database has `database_tasks: false` (an external database Rails doesn't manage), parallel tests fail. `ActiveRecord::TestDatabases.create_and_load_schema` renames every database config with the worker suffix, even ones it shouldn't touch.

The rename at [`test_databases.rb:23`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/test_databases.rb#L23) runs unconditionally, while `database_tasks?` only gates the schema load on line 25:

```ruby
db_config._database = "#{db_config.database}_#{i}"      # always runs

if db_config.database_tasks?                              # only this is conditional
  ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
end
```

Setup:

```yaml
# config/database.yml
test:
  primary:
    adapter: postgresql
    database: myapp_test
  external:
    adapter: postgresql
    database: external_db
    database_tasks: false
```

```ruby
# app/models/application_record.rb
  self.abstract_class = true
  connects_to database: { writing: :primary, reading: :primary }
end

# app/models/external/base.rb
module External
  class Base < ActiveRecord::Base
    self.abstract_class = true
    connects_to database: { writing: :external, reading: :external }
  end
end
```

```ruby
# test/test_helper.rb
class ActiveSupport::TestCase
  parallelize(workers: 2)
end
```

Run `bin/rails test`.

### Expected behavior

Databases with `database_tasks: false` should keep their original name across all parallel workers. Rails can't create the suffixed versions (`external_db_0`, `external_db_1`) because it doesn't manage that database, at least that's what the flag means from my understanding.

### Actual behavior

Every parallel worker fails with:

```
ActiveRecord::NoDatabaseError: Database not found: external_db_0
```

PR #55769 added `include_hidden: true` to `configs_for` so replicas get properly renamed and that's correct for replicas that mirror the primary. But it also catches external databases with `database_tasks: false` that aren't replicas and shouldn't be renamed.

A fix would be to skip configs where `database_tasks?` is false:

```ruby
def self.create_and_load_schema(i, env_name:)
  old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"

  ActiveRecord::Base.configurations.configs_for(env_name: env_name, include_hidden: true).each do |db_config|
    next unless db_config.database_tasks?

    db_config._database = "#{db_config.database}_#{i}"
    ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
  end
ensure
  ActiveRecord::Base.establish_connection
  ENV["VERBOSE"] = old
end
```

This preserves PR #55769's behavior for replicas (which have `database_tasks: true` by default) while leaving external databases alone.

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 3.4.2

vuth-dogo:
Fixes #57068.

This keeps `ActiveRecord::TestDatabases.create_and_load_schema` from suffixing configs that explicitly set `database_tasks: false`, while preserving the hidden-replica renaming behavior added by #55769.

A plain `database_tasks?` guard is too broad here because replicas return false from `database_tasks?` but still need the worker suffix. This patch only skips configs with an explicit `database_tasks: false` setting, and adds a regression test for that case.

### Tests
- `cd activerecord && ARCONN=sqlite3 bin/test test/cases/test_databases_test.rb`
- `cd activerecord && ARCONN=sqlite3 bin/test test/cases/database_configurations/hash_config_test.rb -i /database_tasks/`

### DIFF
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -20,10 +20,11 @@ module ActiveRecord
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
       ActiveRecord::Base.configurations.configs_for(env_name: env_name, include_hidden: true).each do |db_config|
-        db_config._database = "#{db_config.database}_#{i}"
-
         if db_config.database_tasks?
+          db_config._database = "#{db_config.database}_#{i}"
           ActiveRecord::Tasks::DatabaseTasks.reconstruct_from_schema(db_config, nil)
+        elsif db_config.replica?
+          db_config._database = "#{db_config.database}_#{i}"
         end
       end
     ensure
--- a/activerecord/test/cases/test_databases_test.rb
+++ b/activerecord/test/cases/test_databases_test.rb
@@ -131,5 +131,37 @@ class TestDatabasesTest < ActiveRecord::TestCase
       ActiveRecord::Base.establish_connection(:arunit)
       ENV["RAILS_ENV"] = previous_env
     end
+
+    def test_create_databases_after_fork_skips_configs_with_database_tasks_false
+      previous_env, ENV["RAILS_ENV"] = ENV["RAILS_ENV"], "arunit"
+      prev_configs, ActiveRecord::Base.configurations = ActiveRecord::Base.configurations, {
+        "arunit" => {
+          "primary" => { "adapter" => "sqlite3", "database" => "test/db/primary.sqlite3" },
+          "external" => { "adapter" => "sqlite3", "database" => "test/db/external.sqlite3", "database_tasks" => false }
+        }
+      }
+
+      idx = 42
+      primary_db_config = ActiveRecord::Base.configurations.configs_for(env_name: "arunit", name: "primary")
+      expected_primary_database = "#{primary_db_config.database}_#{idx}"
+      external_db_config = ActiveRecord::Base.configurations.configs_for(env_name: "arunit", name: "external", include_hidden: true)
+      expected_external_database = external_db_config.database
+      reconstruct_calls = 0
+
+      ActiveRecord::Tasks::DatabaseTasks.stub(:reconstruct_from_schema, ->(db_config, _) {
+        reconstruct_calls += 1
+        assert_equal expected_primary_database, db_config.database
+      }) do
+        ActiveSupport::Testing::Parallelization.after_fork_hooks.each { |cb| cb.call(idx) }
+      end
+
+      assert_equal 1, reconstruct_calls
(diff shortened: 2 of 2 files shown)


activerecord/lib/active_record/test_databases.rb
```diff
@@ -20,6 +20,8 @@ def self.create_and_load_schema(i, env_name:)
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
       ActiveRecord::Base.configurations.configs_for(env_name: env_name, include_hidden: true).each do |db_config|
+        next if db_config.configuration_hash[:database_tasks] == false
```
eileencodes: I don't like that this blocks is now checking database tasks twice. There has to be a better way than this.

eileencodes (changes requested):

activerecord/lib/active_record/test_databases.rb
vuth-dogo: > I don't like that this blocks is now checking database tasks twice. There has to be a better way than this.

Thanks for the review @eileencodes, you right , the double-check was unnecessary.
I thought why not just separate by config type so I reworked it to use `if/elsif`, primary DBs get renamed + schema reconstructed, replicas just get renamed, and external `database_tasks: false` configs are skipped entirely.

mattmcf:
@eileencodes and @vuth-dogo, Is there anything I can do to help push this fix forward? 

We run a static sqlite db that houses a large repository of financial institution look ups. When the parallel test runner suffixes the db with `_{number}`, any tests that rely on active record lookups to that database fail, because the database isn't found in the app build. 

I think this fix is right, and I've brought the spirit of it into our app while we wait for an upstream fix.

vuth-dogo:
> @eileencodes and @vuth-dogo, Is there anything I can do to help push this fix forward?
> 
> We run a static sqlite db that houses a large repository of financial institution look ups. When the parallel test runner suffixes the db with `_{number}`, any tests that rely on active record lookups to that database fail, because the database isn't found in the app build.
> 
> I think this fix is right, and I've brought the spirit of it into our app while we wait for an upstream fix.

Thanks for your reply.

I’m not sure why the PR hasn’t been followed up on yet. The maintainers may just not have had time to revisit it.

PR #57126, opened 2026-04-02
Removed unnecessary code since minimum ruby required is 3.3.1

adityapandit17:
Ruby 3.2.0 had a bug where `Time.new(..., in: "UTC") ` could return an invalid Time object. To work around this, `ActiveModel::Type::Helpers::TimeValue defined fast_string_to_time` via a runtime probe (`Time.new(2000, 1, 1, 0, 0, 0, "-00:00").yday != 1)` and only used the “workaround” implementation for affected Ruby versions.

Since the project’s minimum supported Ruby is `3.3.1`, that runtime probe is now effectively dead (the condition is always false) and the code in the Ruby-3.2 workaround branch can be removed.

### DIFF
--- a/activemodel/lib/active_model/type/helpers/time_value.rb
+++ b/activemodel/lib/active_model/type/helpers/time_value.rb
@@ -61,33 +61,16 @@ module ActiveModel
             end
           end
 
-          if Time.new(2000, 1, 1, 0, 0, 0, "-00:00").yday != 1 # Early 3.2.x had a bug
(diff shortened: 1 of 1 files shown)


fatkodima:
Delete the added test, squash commits to one and will be good to go.

adityapandit17:
@fatkodima Sure, I added test to make sure we should not have such issues in the future.

adityapandit17:
@fatkodima Done

adityapandit17:
CI is failing for some other reason, it was working before I forced pushed, similar thing happens with my other PR

PR #57129, opened 2026-04-02
Use Hash#each_key instead of Hash#keys.each

ashwin47:
### Motivation / Background

  `Hash#keys.each` allocates an intermediate array of all keys before iterating. `Hash#each_key` iterates directly over the
   hash's internal structure without the extra allocation.

  This follows the same cleanup done in #17099 

  ### Detail

  Replace `keys.each` with `each_key` in two locations where the hash is not mutated during iteration:

  - `activemodel/lib/active_model/schematized_json.rb`
  - `activerecord/lib/active_record/attribute_methods/dirty.rb`

  The remaining 3 `keys.each` calls in the codebase (`hash/keys.rb`, `strong_parameters.rb`, `routing/mapper.rb`) use
  `delete` inside the loop, so they need `keys` to snapshot before mutating.

  ### Additional information

  N/A

  ### Checklist

  - [ ] Tests are added or updated if you fix a bug or add a feature.
  - [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor
  bug fixes and documentation changes should not be included.

### DIFF
--- a/activemodel/lib/active_model/schematized_json.rb
+++ b/activemodel/lib/active_model/schematized_json.rb
@@ -57,7 +57,7 @@ module ActiveModel
       def has_delegated_json(attr, **schema)
         has_json attr, **schema
 
-        schema.keys.each do |schema_key|
+        schema.each_key do |schema_key|
           define_method(schema_key)       { public_send(attr).public_send(schema_key) }
           define_method("#{schema_key}?") { public_send(attr).public_send("#{schema_key}?") }
           define_method("#{schema_key}=") { |value| send(attr).public_send("#{schema_key}=", value) }
--- a/activerecord/test/cases/connection_adapters/connection_handler_test.rb
+++ b/activerecord/test/cases/connection_adapters/connection_handler_test.rb
@@ -247,7 +247,7 @@ module ActiveRecord
           assert_instance_of String, db_config.env_name
           assert_instance_of String, db_config.name
 
-          db_config.configuration_hash.keys.each do |key|
+          db_config.configuration_hash.each_key do |key|
             assert_instance_of Symbol, key
           end
         end
--- a/activesupport/test/env_configuration_test.rb
+++ b/activesupport/test/env_configuration_test.rb
@@ -94,7 +94,7 @@ class EnvConfigurationTest < ActiveSupport::TestCase
       @config.reload
       yield
     ensure
-      attributes.keys.each do |key|
+      attributes.each_key do |key|
         ENV.delete(key)
       end
     end


skipkayhil:
If we're going to accept this change, we need to get them all in one go:

```
❯ rg '[^@_]keys\.each' | wc -l
19
```

I found 19, which is more than the 5 you mention.

ashwin47:
Thanks for reviewing @skipkayhil 

I ran `rg '[^@_]keys\.each'` and went through all 19 matches. Only 4 are actually Hash#keys.each calls that can be converted — I have included these two new which are in tests.

  - activesupport/test/env_configuration_test.rb:97
  - activerecord/test/cases/connection_adapters/connection_handler_test.rb:250

The remaining 15 are not convertible:

  - 6 mutate the hash during iteration (delete inside the loop), so keys is needed as a snapshot — memory_store.rb (×3), mapper.rb, hash/keys.rb, strong_parameters.rb
  - 7 are local Array variables, not Hash#keys calls — length.rb, store.rb, parameter_filter_test.rb, timestamp_test.rb, builder.rb, except.rb, slice.rb
  - 2 use different method chains — keys.each.with_index in cipher.rb, keys.each_with_object in preloader/association.rb

seanpdoyle:
> 2 use different method chains — keys.each.with_index in cipher.rb, keys.each_with_object in preloader/association.rb

These might be viable for replacement, since [Hash#each_key](https://ruby-doc.org/3.4.1/Hash.html#method-i-each_key) returns an enumerator when the block is omitted:

```ruby
irb(main):001> hash = { "a" => 1, "b" => 2, "c" => 3 }
irb(main):002> hash.each_key.with_index(1) { |key, index| puts "#{key} at index #{index}" }
a at index 1
b at index 2
c at index 3
=> {"a" => 1, "b" => 2, "c" => 3}
irb(main):003> hash.each_key.with_object([]) { |key, keys| keys << key }
=> ["a", "b", "c"]
```

guilleiguaran:
CI is broken since some of those aren't instances of hashes:

```
NoMethodError: undefined method 'each_key' for an instance of ActiveModel::AttributeSet
```

guilleiguaran (changes requested):
CI failures are legit.

ashwin47:
Thanks @seanpdoyle — you're right that Hash#each_key returns an enumerator, so the chain works for Hashes. I checked both sites and neither receiver is a Hash:              
                                                                                                                                                                               
  - ` activerecord/lib/active_record/encryption/cipher.rb:41` — keys is the kwarg from try_to_decrypt_with_each(..., keys: Array(key)) (line 26), so it's an Array.
  - `activerecord/lib/active_record/associations/preloader/association.rb:47 `— keys is keys_to_load, initialized as Set.new at line 64.                                         
                                                                                                                                                                               
  Neither Array nor Set defines #each_key, so they stay as-is.                                                                                                                 
                                                                                                                                                                               
  ---                                                                                                                                                                          
The CI failure was in `activerecord/lib/active_record/attribute_methods/dirty.rb: @attributes` is an `ActiveModel::AttributeSet`, not a Hash, and it doesn't define #each_key. I've reverted that line and amended the commit. The other three changes operate on real Hashes. cc @guilleiguaran

PR #42686, opened 2021-07-02
Make ActiveStorage default variant transformations explicit and customizable

brenogazzola:
### Summary
While not immediately obvious, ActiveStorage automatically applies certain transformations to all requested variants:

1. Any image file that is not a web image will be converted to PNG;
2. If no quality setting is given, ImageMagick and libvips will automatically apply their quality defaults, which vary depending on the image format;

This PR adds to active storage a feature called `optimizers`, which makes this behavior explicit, fine tunes it for the web, and allows developers to define their own defaults. 

It does that by changing the `default_variant_transformations` in the `representable` concern to have it call upon the optimizers instead of simply returning a hash with the fallback format that should be used for image blobs that are not web images.

Before:
```
def default_variant_transformations
  { format: default_variant_format }
end
```

Now:
```
def default_variant_transformations
  optimizer_class.new(self).transformations
end

def optimizer_class
  ActiveStorage.optimizers.detect { |klass| klass.accept?(format) }
end
```

### Other Information
Should the developer want to display the original image, they can still do that by using `image_tag` without any extra parameters. And if they want to have different defaults, they can prepend their own optimizers to the array, just like they would to with analyzers.

### Further improvements
If we allow the representations controller to detect the accept header and pass it forward to the representation we can potentially add an optimizer for webp (or JPEG XL) and automatically serve modern file formats to browsers that can handle it.

PR #57139, opened 2026-04-03
Skip image processing for identity format conversions in ActiveStorage

g-pavlik:
### Motivation / Background

This Pull Request has been created because it has potential to save few cycles when variant transformation is not necessary.

### Detail

This Pull Request changes the behavior of #variant method. When asking for variant and only transformation that is requested is a format the same as original's format - the processing can be skipped (the variant blog is still downloaded, though)

### Additional information

For more aggresive optimization one can reach out to `#variant_or_self` method proposed in #57138 

### Checklist

* N/A CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

### DIFF
--- a/activestorage/app/models/active_storage/variant.rb
+++ b/activestorage/app/models/active_storage/variant.rb
@@ -114,7 +114,7 @@ class ActiveStorage::Variant
   def process_from_io(io) # :nodoc:
     return if processed?
 
-    variation.transform(io) do |output|
+    variation.transform(io, content_type: blob.content_type) do |output|
       service.upload(key, output, content_type: content_type)
     end
   end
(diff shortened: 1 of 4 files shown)


g-pavlik:
@guilleiguaran thanks for triggering the review, I addressed the comments.

g-pavlik:
Closing this.

The skip is a silent behavior change in a place where the current behavior is load-bearing. Today every variant is re-encoded at the processor's default quality (see #42686), so apps get a smaller file even when the format is unchanged. Skipping the transform silently stops that: those apps would start serving the original bytes — larger, with original metadata intact — under a variant URL.

Making that safe needs a config plus a framework default, which is a bigger change than this PR, and it overlaps with work already in flight (#57731, #58615).

Pursuing the explicit opt-in shape in #57138 instead.

PR #57140, opened 2026-04-04
Fix GET+JSON+params in integration tests for API-only apps

### LINKED ISSUE #57131: GET request with params keyword in IntegrationTest incorrectly routes as POST in Rails 8.1.3
### Steps to reproduce

Using Rails 8.1.3, create a minimal application with a single GET route and a corresponding controller action. Write an integration test that makes a GET request using the `params:` keyword argument with the `as: :json` option.

**Reproduction script (save as bug_report.rb):**

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "8.1.3"
end

require "minitest/autorun"

  config.root = __dir__
  config.secret_key_base = "secret"
  config.eager_load = false
  Rails.logger = config.logger

  routes.draw do
    get "/items", to: "items#index"
    post "/items", to: "items#create"
  end
end

class ItemsController < ActionController::Base
  def index
    render plain: "GET successful with params: #{params[:filter]}"
  end

  def create
    render plain: "POST successful"
  end
end

require "minitest/autorun"

class BugTest < ActionDispatch::IntegrationTest
  def test_get_with_params_keyword_fails
    get "/items", params: { filter: "active" }, as: :json
    assert_response :success
  end

  def test_get_with_manual_query_string_succeeds
    get "/items?filter=active", as: :json
    assert_response :success
  end
end
```

**Run the script:**
```bash
ruby bug_report.rb
```

### Expected behavior

Both tests should pass. The `get` request with the `params:` keyword argument should correctly route to the `GET /items` route and return a success response.

### Actual behavior

The test `test_get_with_params_keyword_fails` fails with a routing error:

```
Expected response to be a <2XX: success>, but was a <404: Not Found>
Response body: {"status":404,"error":"Not Found","exception":"#<ActionController::RoutingError: No route matches [POST] \"/items\">"}
```

The router incorrectly interprets the GET request as a POST request when the `params:` keyword is used with `as: :json`.

The second test using a manually constructed query string passes as expected.

### Additional context

This issue affects Rails 8.1.3. It appears to be related to how the test helper encodes the `params:` hash into the request body for GET requests. When a GET request includes a body (non-standard but technically allowed), the Rails router may misinterpret the request method or route it incorrectly.

The problem manifests in real-world projects where multiple controller tests fail with the same pattern. The workaround is to manually construct query strings instead of using the `params:` keyword argument.

### Workaround

For affected tests, replace:
```ruby
get some_url, params: { id: 123 }, as: :json
```

With:
```ruby
get "#{some_url}?id=123", as: :json
```

### Environment

- Rails version: 8.1.3
- Ruby version: 4.0.2
- Operating system: Ubuntu 25.10

### Related issues

This may be related to or a regression of issues #45409 and #26033.

Mordorreal:
### Summary

`params:` in integration test helpers was ambiguous — for GET requests with `as: :json` it was unclear whether params should go in the query string or request body. The original workaround converted GET to POST with `X-Http-Method-Override`, which broke API-only apps that exclude `Rack::MethodOverride`.

### Fix

Add `query:` and `body:` kwargs for explicit control. Remove the POST conversion hack.

- `query:` — always URL query string, any HTTP method
- `body:` — always encoded request body, any HTTP method
- `params:` — unchanged: GET → query string, other → body

```ruby
get  "/search", query: { q: "rails" }, as: :json
post "/search", query: { page: 1 }, body: { filters: {} }, as: :json
```

Fixes #57131

### Benchmark

Ruby 4.0.1, Rails 8.2.0.alpha

| Scenario | main (before) | PR (after) | Delta |
|---|---|---|---|
| `GET params` (no as:) | 27,428 i/s | 27,178 i/s | same-ish |
| `GET params + as: :json` | 25,946 i/s | 27,667 i/s | **+6.6% faster** |
| `POST params + as: :json` | 28,274 i/s | 27,916 i/s | same-ish |

New kwargs (baseline):

| Scenario | i/s |
|---|---|
| `GET query: + as: :json` | 26,239 |
| `POST body: + as: :json` | 28,126 |
| `POST query:+body: + as: :json` | 23,881 |

### DIFF
--- a/actionpack/lib/action_dispatch/testing/integration.rb
+++ b/actionpack/lib/action_dispatch/testing/integration.rb
@@ -198,6 +198,9 @@ module ActionDispatch
       # *   `params`: The HTTP parameters that you want to pass. This may be `nil`, a
       #     Hash, or a String that is appropriately encoded
       #     (`application/x-www-form-urlencoded` or `multipart/form-data`).
+      #     For GET requests, params are sent as query string. For other methods,
+      #     params are sent as the request body. Use `query:` or `body:` for
+      #     explicit control.
       # *   `headers`: Additional headers to pass, as a Hash. The headers will be
       #     merged into the Rack env hash.
       # *   `env`: Additional env to pass, as a Hash. The headers will be merged into
@@ -208,6 +211,10 @@ module ActionDispatch
       # *   `as`: Used for encoding the request with different content type. Supports
       #     `:json` by default and will set the appropriate request headers. The
       #     headers will be merged into the Rack env hash.
+      # *   `query`: Parameters to always send as URL query string, regardless of
+      #     HTTP method. Accepts a Hash or pre-encoded String.
+      # *   `body`: Parameters to always send as the request body, encoded per `as:`.
+      #     Can be combined with `query:` to send both query string and body params.
       #
       #
       # This method is rarely used directly. Use RequestHelpers#get,
@@ -223,13 +230,15 @@ module ActionDispatch
       # Example:
       #
       #     process :get, '/author', params: { since: 201501011400 }
-      def process(method, path, params: nil, headers: nil, env: nil, xhr: false, as: nil)
+      #     process :get, '/author', query: { since: 201501011400 }, as: :json
+      #     process :post, '/search', query: { page: 1 }, body: { q: "rails" }, as: :json
+      def process(method, path, params: nil, headers: nil, env: nil, xhr: false, as: nil, query: nil, body: nil)
         request_encoder = RequestEncoder.encoder(as)
         headers ||= {}
 
-        if method == :get && as == :json && params
-          headers["X-Http-Method-Override"] = "GET"
-          method = :post
+        if query
+          query_string = query.is_a?(String) ? query : Rack::Utils.build_nested_query(query)
+          path = path.include?("?") ? "#{path}&#{query_string}" : "#{path}?#{query_string}"
         end
 
         if path.include?("://")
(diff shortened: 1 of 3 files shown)


actionpack/lib/action_dispatch/testing/integration.rb
```diff
@@ -227,7 +227,7 @@ def process(method, path, params: nil, headers: nil, env: nil, xhr: false, as: n
         request_encoder = RequestEncoder.encoder(as)
         headers ||= {}
 
-        if method == :get && as == :json && params
+        if method == :get && as == :json && params && app_has_method_override_middleware?
```
byroot: Checking for the existence of a middleware seems a bit too ad hoc and fragile to me, so I'm hesitant to merge this fix as is.

This `headers["X-Http-Method-Override"] = "GET"` seemed weird, but according to git, was introduced in https://github.com/Mordorreal/rails/commit/af1680f51c127c7e40019e68c01b2a693ba2640f

If we take step back, it seems the main issue is that `params:` is ambiguous, and in some case it's not clear if it's meant as query params, or as request body. 

Perhaps a better fix would be to accept both `query:` and `body:` (or something like that), so that ambiguity doesn't exist in the first place.

Mordorreal: Good point. Reworked: removed the middleware check and the GET→POST conversion entirely. Added explicit `query:` and `body:` kwargs instead — `params:` retains its existing behavior (GET → query string, other methods → body).

`get '/path', params: { q: 'rails' }, as: :json` now stays GET with params in the query string, which also makes it work in API-only apps without any middleware detection.

Benchmark shows the common `GET+params+as: :json` case is ~7% faster since we no longer run the encoder or mutate headers.

PR #57143, opened 2026-04-05
Optimize `Relation#blank?` to use `#empty?` instead of loading all records

oggy:
Previously, calling `blank?` on a relation would load all records via `records.blank?`. Now it delegates to `empty?`, which can use a COUNT query or check the loaded state without forcing a full load.

An innocent looking `if user.comments.blank?` in a view can be a performance footgun if it loads all the records.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Optimize `Relation#blank?` to use `#empty?` instead of loading all records.
+
+    Previously, calling `blank?` on a relation would load all records via
+    `records.blank?`. Now it delegates to `empty?`, which can use a COUNT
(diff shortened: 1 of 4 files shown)


kbrock:
Good stuff.

FYI: all databases are failing. I think because it was running an unexpected query. Myabe there is something behind `query.blank?` elsewhere?
```
Failure:
--
RelationTest#test_presence [test/cases/relations_test.rb:2041]:
1 instead of 0 queries were executed. Queries:
 
SELECT 1 AS one FROM "topics" LIMIT ?.
Expected: 0
Actual: 1
```

oggy:
Thanks for the nudge, @kbrock -- test updated.

The current failures (2 specific adapters) are related to connections/auth, which don't seem related to this change.

PR #57152, opened 2026-04-06
[RF-DOCS] Rewrite the Layouts and Rendering guide

ayushn21:
### Motivation / Background

A complete rewrite of the Layouts and Rendering guide. It covered a lot of topics already covered in other guides, and the narrative flow went off on numerous tangents. I've moved content to other guides and removed information as I saw fit.

### Detail

I've rewritten the guide to focus heavily on the _relationship_ between the controller and the view. 

A lot of detail about rendering partials was already covered in the Action View Overview guide, so I've removed it. I moved the information about structuring layouts and similar view related detail into the Action View Overview with it fits a lot better with the narrative.

I've moved the section on request variants out of the Action Controller Overview and added it to this guide where I feel it fits a lot better. I also took the opportunity to add a brief intro about rendering to the Action Controller Overview and signpost this guide since they are very closely related.

### Internal Audit

Section 1: "though it normally hands off any heavy code to the Model" - 'heavy code' is possibly a bit unhelpful for beginners. How about 'it normally hands off more complex logic to methods on the model' or something like that?
Section 2: Creating responses. I almost want to go one step even further back and be sure to cover what we mean by HTTP responses, for those just starting out. Not in massive depth, because it isn't the right place for that, but just a little bit more clarification. Something like: "there are three ways to create an HTTP response in order to display information to the user on the page".

"Call [render](https://api.rubyonrails.org/v7.1.3.4/classes/ActionController/Rendering.html#method-i-render) to create a full response to send back to the browser". Perhaps we could just add one more sentence here to break down what a 'full response' really is in terms of what a new developer/user would experience.

"Call [head](https://api.rubyonrails.org/v7.1.3.4/classes/ActionController/Head.html#method-i-head) to create a response consisting solely of HTTP headers to send back to the browser" - maybe we could add examples of why we might do this?
[New boost](https://3.basecamp.com/3076981/buckets/35498807/boosts/new?boost%5Bboostable_gid%5D=Z2lkOi8vYmMzL1JlY29yZGluZy83Njk3MzY5MTA4)Section 1: "though it normally hands off any heavy code to the Model" - 'heavy code' is possibly a bit unhelpful for beginners. How about 'it normally hands off more complex logic to methods on the model' or something like that?

Section 2: Creating responses. I almost want to go one step even further back and be sure to cover what we mean by HTTP responses, for those just starting out. Not in massive depth, because it isn't the right place for that, but just a little bit more clarification. Something like: "there are three ways to create an HTTP response in order to display information to the user on the page".

"Call [render](https://api.rubyonrails.org/v7.1.3.4/classes/ActionController/Rendering.html#method-i-render) to create a full response to send back to the browser". Perhaps we could just add one more sentence here to break down what a 'full response' really is in terms of what a new developer/user would experience.

"Call [head](https://api.rubyonrails.org/v7.1.3.4/classes/ActionController/Head.html#method-i-head) to create a response consisting solely of HTTP headers to send back to the browser" - maybe we could add examples of why we might do this?
[New boost](https://3.basecamp.com/3076981/buckets/35498807/boosts/new?boost%5Bboostable_gid%5D=Z2lkOi8vYmMzL1JlY29yZGluZy83Njk3MzY5MTA4)

End of section 2.1.1 we mention `status: :unprocessable_entity`. Worth expanding out this slightly to point out that we can render a template with a specific status code etc., and briefly why we might want to do this. We've got more info on this in section 2.2.13.4, so maybe just worth linking to this.

2.2.4 - I feel like this section is slightly confusing. We are told that it's valid to use `render inline:`, but the warning box tells us that there is seldom any reason to use this. Maybe it would be better to find a reason why we would need this, but if not, make it short and clear - more like 'this is here but probably won't be needed'.

2.2.7 and 2.2.8 feel repetitive. Can we combine these into one section and say that it's applicable to both XML and JSON?
2.2.13.1 MIME content type - worth a link to the MDN web docs or similar for explanation.

2.2.14 Finding Layouts - this is mostly a repeat of information at the beginning of the guide - the only difference is the addition of the builder information. We can probably re-organise this to make both sections more DRY.

At the start of Section 3, it mentions 'the rules for finding layouts that were mentioned earlier in this guide'. It would be more user-friendly to also link to them at this point.

3.1.1 - let's provide links to more information about RSS/Atom if possible?
Likewise, would be good to provide a link to the Sprockets gem repo when we mention it, for context.

- I think the "Using `render`" section could be it's own top section. This probably applies to all sections ("Rendering by Default", "Using `redirect_to`" and "Using `head`"). All these sections have interesting sub section that currently don't show up in the sidebar and this guide doesn't have that much sub sections as of yet.

- "Rendering an Action's View" shows an example of explicitly rendering a view (the "edit" view in the "update" action). Maybe rename this section to: "Explicitly Rendering an Action's View"? Maybe the example could be simpler as well with a "show" action that renders for example a "details" view?
- "Ajax libraries" is probably an outdated term? Maybe replace with something about an API end-point?

- The "Finding Layouts" section should probably be a top section in a "Layout and Rendering" guide?
- Some examples here could benefit from having the filename as a comment in the example code.

- The whole "Asset Tag Helpers" section can probably be removed as these are already described the in the "Action View Helpers" guide. I'm not sure it adds anything here.
- Maybe move the "Understanding yield" section to the "Action View Overview" guide under "Layouts" which also handles `yield`?
- The above also applies to "Using the content_for method".
- The "Using Partials" section can probably be removed as this is already described in the "Action View Overview" guide. Partials don't really know anything about Controllers, so it makes more sense to have it in a Action View guide.
- I'm not sure if the "Nested Layouts" adds anything or if that is a commonly used pattern.

### DIFF
--- a/guides/source/action_controller_overview.md
+++ b/guides/source/action_controller_overview.md
@@ -3,1018 +3,362 @@
 Action Controller Overview
 ==========================
 
-In this guide, you will learn how controllers work and how they fit into the request cycle in your application.
+In this guide, you will learn how controllers work and how they fit into the
+request cycle in your application.
 
 After reading this guide, you will know how to:
 
-* Follow the flow of a request through a controller.
-* Access parameters passed to your controller.
-* Use Strong Parameters and permit values.
-* Store data in the cookie, the session, and the flash.
-* Work with action callbacks to execute code during request processing.
-* Use the Request and Response Objects.
+- Follow the flow of a request through a controller.
+- Render HTTP responses as well as redirect responses.
+- Work with action callbacks to execute code during request processing.
+- Access and securely filter parameters passed to your controller.
+- Store data in the cookie, the session, and the flash.
+- Use the Request and Response Objects.
 
 --------------------------------------------------------------------------------
 
-Introduction
-------------
+Action Controller Basics
+------------------------
 
-Action Controller is the C in the Model View Controller
+Action Controller is the **C** in the Model View Controller
 ([MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller))
-pattern. After the [router](routing.html) has matched a controller to an
-incoming request, the controller is responsible for processing the request and
-generating the appropriate output.
+pattern. The [router](routing.html) matches a controller to an incoming request,
+which is responsible for processing the request and generating the response.
 
 For most conventional
 [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer)
 applications, the controller will receive the request, fetch or save data from a
-model, and use a view to create HTML output.
+model, and render a view to create HTML output.
 
-You can imagine that a controller sits between models and views. The controller
-makes model data available to the view, so that the view can display that data
-to the user. The controller also receives user input from the view and saves or
-updates model data accordingly.
+A controller sits between models and views. The controller makes model data
+available to the view, so that the view can display that data to the user. The
+controller also receives user input from the view and saves or updates model
+data accordingly.
 
-Creating a Controller
----------------------
+### Structure of a Controller
 
 A controller is a Ruby class which inherits from `ApplicationController` and has
-methods just like any other class. Once an incoming request is matched to a
-controller by the router, Rails creates an instance of that controller class and
-calls the method with the same name as the action.
+methods just like any other class. Public methods in a controller are also known
+as _actions_, as they are responsible for rendering responses.
 
 ```ruby
-class ClientsController < ApplicationController
-  def new
+# app/controllers/products_controller.rb
+
+class ProductsController < ApplicationController
+  def index
   end
 end
 ```
 
-Given the above `ClientsController`, if a user goes to `/clients/new` in your
-application to add a new client, Rails will create an instance of
-`ClientsController` and call its `new` method. If the `new` method is empty, Rails
-will automatically render the `new.html.erb` view by default.
-
-NOTE: The `new` method is an instance method here, called on an instance of `ClientsController`. This should not be confused with the `new` class method (i.e., `ClientsController.new`).
-
-In the `new` method, the controller would typically create an instance of the
-`Client` model, and make it available as an instance variable called `@client`
-in the view:
+As per Rails conventions, controllers should define up to 7 conventional CRUD
+actions as demonstrated below. This convention is used by the
+[Router DSL](routing.html#crud-verbs-and-actions) to configure resourceful
+routes to the controller. Other actions may be defined and routed to, but these
+are outside Rails conventions. Further details are available in the
+[Routing guide](routing.html).
 
 ```ruby
-def new
-  @client = Client.new
+# config/routes.rb
+
+Rails.application.routes.draw do
+  # A resourceful route to the `ProductsController`
+  resources :products
+end
+```
+
+```ruby
+# app/controllers/products_controller.rb
+
+class ProductsController < ApplicationController
+  # GET /products
+  def index
+    # Display all products
+  end
+
+  # GET /products/new
+  def new
+    # Render a form to create a new product
+  end
+
+  # POST /products
+  def create
+    # Handle the submission form rendered in `new`
+    # and create the product in the database
+  end
+
+  # GET /products/:id
+  def show
+    # Show the product
+  end
+
+  # GET /products/:id/edit
+  def edit
+    # Render a form to edit the product
+  end
+
+  # PATCH /products/:id
+  def update
+    # Handle the submission form rendered in `edit`
+    # and update the product in the database
+  end
+
+  # DELETE /product/:id
+  def destroy
+    # Delete the product
+  end
+end
+```
+
+Once the [router matches](routing.html) an incoming request to a controller and
+action, Rails creates an instance of that controller class and calls the method
+with the same name as the action.
+
+NOTE: The `new` method in the above controller is an instance method, called on
+an instance of `ProductsController`. This should not be confused with the `new`
+class method used to instantiate objects (`ProductsController.new`).
+
+### Naming Conventions
+
+Rails favors pluralizing the resource in the controller's name. For example,
+`ProductsController` is preferred over `ProductController` and
+`SiteAdminsController` over `SiteAdminController` or `SitesAdminsController`.
+However, the plural names are not strictly required — for example,
+`ApplicationController`.
+
+Following this naming convention will allow you to use
+[resourceful routes](routing.html#crud-verbs-and-actions) without needing to
+qualify each with
+[additional options](routing.html#specifying-a-controller-to-use). The
+convention also makes named route helpers consistent throughout your
+application.
+
+The controller naming convention is different from models. While plural names
+are preferred for controller names, the singular form is preferred for
+[model names](active_record_basics.html#naming-conventions) (e.g. `Account` vs.
+`Accounts`).
+
+Controller actions should be _public_, as only _public_ methods are callable as
+actions. Helper methods within controllers which are _not_ intended to be
+actions should be declared `private` or `protected`.
+
+WARNING: Ensure that you do not override methods defined by
+`ActionController::Base` when creating actions. Accidentally redefining them
+could result in `SystemStackError`. If you limit your controllers to the 7 CRUD
+actions, this won't be an issue.
+
+NOTE: If you must use a reserved method as an action name, one workaround is to
+use a custom route to map the reserved method name to your non-reserved action
+method.
+
+### Rendering Responses
+
+Consider the below route and controller.
+
+```ruby
+# config/routes.rb
+
+Rails.application.routes.draw do
+  # A resourceful route to the `ProductsController`
+  resources :products
+end
+```
+
+```ruby
+class ProductsController < ApplicationController
+  def index
+  end
+end
+```
+
+When a user navigates to `/products`, Rails will create an instance of
+`ProductsController` and call its `index` method. If the `index` method is
+empty, Rails will automatically render `app/views/products/index.html.erb`.
+
+You can explicity define the template to render using the `render` method:
+
+```ruby
+class ProductsController < ApplicationController
+  def index
+    # Renders `app/views/products/feed.html.erb`
+    render "feed"
+  end
+end
+```
+
+WARNING: Calling `render` does not `return` from the current scope. Statements
+after the method call will still be executed. Ensure you do not call `render`
+more than once in any given action, as it will raise an
+`AbstractController::DoubleRenderError`.
+
+To enable rendering of additional formats, not just HTML, use a `respond_to`
+block. This will choose the correct template based on the `Accept` header in the
+HTTP request:
+
+```ruby
+class ProductsController < ApplicationController
+  def index
+    respond_to do |format|
+      format.html   # renders `app/views/products/index.html.erb`
+      format.xml    # renders `app/views/products/index.xml.erb`
+    end
+  end
+end
+```
+
+See the
+[Layouts and Rendering guide](layouts_and_rendering.html#rendering-responses)
+for futher details on rendering responses.
+
+In the `index` method, the controller would typically create an array of the
+`Product` model instances, and make it available as an instance variable called
+`@products` in the view:
+
+```ruby
+def index
+  @products = Product.all
 end
 ```
 
 NOTE: All controllers inherit from `ApplicationController`, which in turn
 inherits from
 [`ActionController::Base`](https://api.rubyonrails.org/classes/ActionController/Base.html).
-For [API only](https://guides.rubyonrails.org/api_app.html) applications `ApplicationController` inherits from
+For [API only](https://guides.rubyonrails.org/api_app.html) applications
+`ApplicationController` inherits from
 [`ActionController::API`](https://edgeapi.rubyonrails.org/classes/ActionController/API.html).
 
-### Controller Naming Convention
+### Redirecting Requests
 
-Rails favors making the resource in the controller's name plural. For example,
-`ClientsController` is preferred over `ClientController` and
-`SiteAdminsController` over `SiteAdminController` or `SitesAdminsController`.
-However, the plural names are not strictly required (e.g.
-`ApplicationController`).
-
-Following this naming convention will allow you to use the [default route
-generators](routing.html#crud-verbs-and-actions) (e.g. `resources`) without
-needing to qualify each with options such as
-[`:controller`](routing.html#specifying-a-controller-to-use). The convention
-also makes named route helpers consistent throughout your application.
-
-The controller naming convention is different from models. While plural names
-are preferred for controller names, the singular form is preferred for [model
-names](active_record_basics.html#naming-conventions) (e.g. `Account` vs.
-`Accounts`).
-
-Controller actions should be `public`, as only `public` methods are callable as
-actions. It is also best practice to lower the visibility of helper methods
-(with `private` or `protected`) which are _not_ intended to be actions.
-
-WARNING: Some method names are reserved by Action Controller. Accidentally
-redefining them could result in `SystemStackError`. If you limit your
-controllers to only RESTful [Resource Routing][] actions you should not need to
-worry about this.
-
-NOTE: If you must use a reserved method as an action name, one workaround is to
-use a custom route to map the reserved method name to your non-reserved action
-method.
-
-[Resource Routing]: routing.html#resource-routing-the-rails-default
-
-Parameters
-----------
-
-Data sent by the incoming request is available in your controller in the
-[`params`][] hash. There are two types of parameter data:
-
-- Query string parameters which are sent as part of the URL (for example, after
-  the `?` in `http://example.com/accounts?filter=free`).
-- POST parameters which are submitted from an HTML form.
-
-Rails does not make a distinction between query string parameters and POST
-parameters; both are available in the `params` hash in your controller. For
-example:
+Instead of rendering a fully-formed response, such as an HTML document, you
+might want to redirect the user to a different path. An
+[HTTP redirection status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#redirection_messages)
+can be used for this. Rails' [`redirect_to`][] method uses `302 Found` by
+default.
 
 ```ruby
-class ClientsController < ApplicationController
-  # This action receives query string parameters from an HTTP GET request
-  # at the URL "/clients?status=activated"
-  def index
-    if params[:status] == "activated"
-      @clients = Client.activated
-    else
-      @clients = Client.inactivated
-    end
-  end
-
-  # This action receives parameters from a POST request to "/clients" URL with  form data in the request body.
-  def create
-    @client = Client.new(params[:client])
-    if @client.save
-      redirect_to @client
-    else
-      render "new"
-    end
-  end
-end
+redirect_to photos_url
 ```
 
-NOTE: The `params` hash is not a plain old Ruby Hash; instead, it is an
-[`ActionController::Parameters`][] object. It does not inherit from Hash, but it behaves mostly like Hash.
-It provides methods for filtering `params` and, unlike a Hash, keys `:foo` and `"foo"` are considered to be the same.
+This will send a `302` HTTP response to the browser with `photos_url` in the
+`Location` header. The browser will then make a new `GET` request to the
+`photos_url`.
 
-[`params`]:
-    https://api.rubyonrails.org/classes/ActionController/StrongParameters.html#method-i-params
-[`ActionController::Parameters`]:
-    https://api.rubyonrails.org/classes/ActionController/Parameters.html
-
-### Hash and Array Parameters
-
-The `params` hash is not limited to one-dimensional keys and values. It can
-contain nested arrays and hashes. To send an array of values, append an empty
-pair of square brackets `[]` to the key name:
-
-```
-GET /users?ids[]=1&ids[]=2&ids[]=3
+```http
+HTTP/1.1 302 Found
+referrer-policy: strict-origin-when-cross-origin
+location: http://localhost:3000/photos
+content-type: text/html; charset=utf-8
+cache-control: no-cache
+connection: close
+content-length: 0
 ```
 
-NOTE: The actual URL in this example will be encoded as
-`/users?ids%5b%5d=1&ids%5b%5d=2&ids%5b%5d=3` as the `[` and `]` characters are
-not allowed in URLs. Most of the time you don't have to worry about this because
-the browser will encode it for you, and Rails will decode it automatically, but
-if you ever find yourself having to send those requests to the server manually
-you should keep this in mind.
+NOTE: It's worth being aware that `redirect_to` doesn't move execution to a
+different method within the same request. It sends an HTTP response and then the
+browser makes a new request to the redirected location which won't have any
+context from the previous request.
 
-The value of `params[:ids]` will be the array `["1", "2", "3"]`. Note that
-parameter values are always strings; Rails does not attempt to guess or cast the
-type.
-
-NOTE: Values such as `[nil]` or `[nil, nil, ...]` in `params` are replaced with
-`[]` for security reasons by default. See [Security
-Guide](security.html#unsafe-query-generation) for more information.
-
-To send a hash, you include the key name inside the brackets:
-
-```html
-<form accept-charset="UTF-8" action="/users" method="post">
-  <input type="text" name="user[name]" value="Acme" />
-  <input type="text" name="user[phone]" value="12345" />
-  <input type="text" name="user[address][postcode]" value="12345" />
-  <input type="text" name="user[address][city]" value="Carrot City" />
-</form>
-```
-
-When this form is submitted, the value of `params[:user]` will be `{ "name" =>
-"Acme", "phone" => "12345", "address" => { "postcode" => "12345", "city" =>
-"Carrot City" } }`. Note the nested hash in `params[:user][:address]`.
-
-The `params` object acts like a Hash, but lets you use symbols and strings
-interchangeably as keys.
-
-### Composite Key Parameters
-
-[Composite key parameters](active_record_composite_primary_keys.html) contain
-multiple values in one parameter separated by a delimiter (e.g., an underscore).
-Therefore, you will need to extract each value so that you can pass them to
-Active Record. You can use the `extract_value` method to do that.
-
-For example, given the following controller:
+Alternatively, you can use [`redirect_back`][] to return the user to the page
+they just came from. The location is pulled from the `HTTP_REFERER` header which
+is not guaranteed to be set by the browser, so you must provide a
+`fallback_location`.
 
 ```ruby
-class BooksController < ApplicationController
-  def show
-    # Extract the composite ID value from URL parameters.
-    id = params.extract_value(:id)
-    @book = Book.find(id)
-  end
-end
+redirect_back(fallback_location: root_path)
+# or
+redirect_back_or_to root_path
 ```
 
-And this route:
+WARNING: Similar to `render`, `redirect_to` and `redirect_back` do not
+automatically `return` from the current scope, instead they simply set the HTTP
+response. Statements occurring after them will still be executed.
+
+[`redirect_to`]:
+  https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to
+[`redirect_back`]:
+  https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back
+
+Use the [`status:`](layouts_and_rendering.html#status) option to use a different
+HTTP response code. Both numeric and symbolic values are valid.
 
 ```ruby
-get "/books/:id", to: "books#show"
+redirect_to photos_path, status: :see_other
 ```
 
-When a user requests the URL `/books/4_2`, the controller will extract the
-composite key value `["4", "2"]` and pass it to `Book.find`. The `extract_value`
-method may be used to extract arrays out of any delimited parameters.
-
-### JSON Parameters
-
-If your application exposes an API, you will likely accept parameters in JSON
-format. If the `content-type` header of your request is set to
-`application/json`, Rails will automatically load your parameters into the
-`params` hash, which you can access as you would normally.
-
-So for example, if you are sending this JSON content:
-
-```json
-{ "user": { "name": "acme", "address": "123 Carrot Street" } }
-```
-
-Your controller will receive:
-
-```ruby
-{ "user" => { "name" => "acme", "address" => "123 Carrot Street" } }
-```
-
-#### Configuring Wrap Parameters
-
-You can use [Wrap Parameters][], which automatically add the controller name to
-JSON parameters. For example, you can send the below JSON without a root `:user`
-key prefix:
-
-```json
-{ "name": "acme", "address": "123 Carrot Street" }
-```
-
-Assuming that you're sending the above data to the `UsersController`, the JSON
-will be wrapped within the `:user` key like this:
-
-```ruby
-{ name: "acme", address: "123 Carrot Street", user: { name: "acme", address: "123 Carrot Street" } }
-```
-
-NOTE: Wrap Parameters adds a clone of the parameters to the hash within a key
-that is the same as the controller name. As a result, both the original version
-of the parameters and the "wrapped" version of the parameters will exist in the
-params hash.
-
-This feature clones and wraps parameters with a key chosen based on your
-controller's name. It is configured to `true` by default. If you do not want to
-wrap parameters you can
-[configure](configuring.html#config-action-controller-wrap-parameters-by-default)
-it to `false`.
-
-```ruby
-config.action_controller.wrap_parameters_by_default = false
-```
-
-You can also customize the name of the key or specific parameters you want to
-wrap, see the [API
-documentation](https://api.rubyonrails.org/classes/ActionController/ParamsWrapper.html)
-for more.
-
-[Wrap Parameters]:
-    https://api.rubyonrails.org/classes/ActionController/ParamsWrapper/Options/ClassMethods.html#method-i-wrap_parameters
-
-### Routing Parameters
-
-Parameters specified as part of a route declaration in the `routes.rb` file are
-also made available in the `params` hash. For example, we can add a route that
-captures the `:status` parameter for a client:
-
-```ruby
-get "/clients/:status", to: "clients#index", foo: "bar"
-```
-
-When a user navigates to `/clients/active` URL, `params[:status]` will be set to
-"active". When this route is used, `params[:foo]` will also be set to "bar", as
-if it were passed in the query string.
-
-Any other parameters defined by the route declaration, such as `:id`, will also
-be available.
-
-NOTE: In the above example, your controller will also receive `params[:action]`
-as "index" and `params[:controller]` as "clients". The `params` hash will always
-contain the `:controller` and `:action` keys, but it's recommended to use the
-methods [`controller_name`][] and [`action_name`][] instead to access these
-values.
-
-[`controller_name`]:
-    https://api.rubyonrails.org/classes/ActionController/Metal.html#method-i-controller_name
-[`action_name`]:
-    https://api.rubyonrails.org/classes/AbstractController/Base.html#method-i-action_name
-
-### The `default_url_options` Method
-
-You can set global default parameters for [`url_for`](
-https://api.rubyonrails.org/classes/ActionView/RoutingUrlFor.html#method-i-url_for)
-by defining a method called `default_url_options` in your controller. For
-example:
-
-```ruby
-class ApplicationController < ActionController::Base
-  def default_url_options
-    { locale: I18n.locale }
-  end
-end
-```
-
-The specified defaults will be used as a starting point when generating URLs.
-They can be overridden by the options passed to `url_for` or any path helper
-such as `posts_path`. For example, by setting `locale: I18n.locale`, Rails will
-automatically add the locale to every URL:
-
-```ruby
-posts_path # => "/posts?locale=en"
-```
-
-You can still override this default if needed:
-
-```ruby
-posts_path(locale: :fr) # => "/posts?locale=fr"
-```
-
-NOTE: Under the hood, `posts_path` is a shorthand for calling `url_for` with the
-appropriate parameters.
-
-If you define `default_url_options` in `ApplicationController`, as in the
-example above, these defaults will be used for all URL generation. The method
-can also be defined in a specific controller, in which case it only applies to
-URLs generated for that controller.
-
-In a given request, the method is not actually called for every single generated
-URL. For performance reasons the returned hash is cached per request.
-
-Strong Parameters
------------------
-
-With Action Controller [Strong
-Parameters](https://api.rubyonrails.org/classes/ActionController/StrongParameters.html),
-parameters cannot be used in Active Model mass assignments until they have been
-explicitly permitted. This means you will need to decide which attributes to
-permit for mass update and declare them in the controller. This is a security
-practice to prevent users from accidentally updating sensitive model attributes.
-
-In addition, parameters can be marked as required and the request will result in
-a 400 Bad Request being returned if not all required parameters are passed in.
-
-```ruby
-class PeopleController < ActionController::Base
-  # This will raise an ActiveModel::ForbiddenAttributesError
-  # because it's using mass assignment without an explicit permit.
-  def create
-    Person.create(params[:person])
-  end
-
-  # This will work as we are using `person_params` helper method, which has the
-  # call to `expect` to allow mass assignment.
-  def update
-    person = Person.find(params[:id])
-    person.update!(person_params)
-    redirect_to person
-  end
-
-  private
-    # Using a private method to encapsulate the permitted parameters is a good
-    # pattern. You can use the same list for both create and update.
-    def person_params
-      params.expect(person: [:name, :age])
-    end
-end
-```
-
-### Permitting Values
-
-#### `expect`
-
-The [`expect`][] method provides a concise and safe way to require and permit
-parameters.
-
-```ruby
-id = params.expect(:id)
-```
-
-The above `expect` will always return a scalar value and not an array or hash.
-Another example is form params, you can use `expect` to ensure that the root key
-is present and the attributes are permitted.
-
-```ruby
-user_params = params.expect(user: [:username, :password])
-user_params.has_key?(:username) # => true
-```
-
-In the above example, if the `:user` key is not a nested hash with the specified
-keys, `expect` will raise an error and return a 400 Bad Request response.
-
-To require and permit an entire hash of parameters, [`expect`][] can be used in
-this way.
-
-```ruby
-params.expect(log_entry: {})
-```
-
-This marks the `:log_entry` parameters hash and any sub-hash of it as permitted
-and does not check for permitted scalars, anything is accepted.
-
-WARNING: Extreme care should be taken when calling `expect` with an empty
-hash, as it will allow all current and future model attributes to be
-mass-assigned.
-
-#### `permit`
-
-Calling [`permit`][] allows the specified key in `params` (`:id` or `:admin`
-below) for inclusion in mass assignment (e.g. via `create` or `update`):
-
-```irb
-params = ActionController::Parameters.new(id: 1, admin: "true")
-=> #<ActionController::Parameters {"id"=>1, "admin"=>"true"} permitted: false>
-params.permit(:id)
-=> #<ActionController::Parameters {"id"=>1} permitted: true>
-params.permit(:id, :admin)
-=> #<ActionController::Parameters {"id"=>1, "admin"=>"true"} permitted: true>
-```
-
-For the permitted key `:id`, its value also needs to be one of these permitted
-scalar values: `String`, `Symbol`, `NilClass`, `Numeric`, `TrueClass`,
-`FalseClass`, `Date`, `Time`, `DateTime`, `StringIO`, `IO`,
-`ActionDispatch::Http::UploadedFile`, and `Rack::Test::UploadedFile`.
-
-If you have not called `permit` on the key, it will be filtered out. Arrays,
-hashes, or any other objects are not injected by default.
-
-To include a value in `params` that's an array of one of the permitted scalar
-values, you can map the key to an empty array like this:
-
-```irb
-params = ActionController::Parameters.new(tags: ["rails", "parameters"])
-=> #<ActionController::Parameters {"tags"=>["rails", "parameters"]} permitted: false>
-params.permit(tags: [])
-=> #<ActionController::Parameters {"tags"=>["rails", "parameters"]} permitted: true>
-```
-
-To include hash values, you can map to an empty hash:
-
-```irb
-params = ActionController::Parameters.new(options: { darkmode: true })
-=> #<ActionController::Parameters {"options"=>{"darkmode"=>true}} permitted: false>
-params.permit(options: {})
-=> #<ActionController::Parameters {"options"=>#<ActionController::Parameters {"darkmode"=>true} permitted: true>} permitted: true>
-```
-
-The above `permit` call ensures that values in `options` are permitted scalars
-and filters out anything else.
-
-WARNING: The `permit` with an empty hash is convenient since sometimes it is not
-possible or convenient to declare each valid key of a hash parameter or its
-internal structure. But note that the above `permit` with an empty hash opens
-the door to arbitrary input.
-
-#### `permit!`
-
-There is also [`permit!`][] (with an `!`) which permits an entire hash of
-parameters without checking the values.
-
-```irb
-params = ActionController::Parameters.new(id: 1, admin: "true")
-=> #<ActionController::Parameters {"id"=>1, "admin"=>"true"} permitted: false>
-params.permit!
-=> #<ActionController::Parameters {"id"=>1, "admin"=>"true"} permitted: true>
-```
-
-WARNING: Extreme care should be taken when using `permit!`, as it will allow all
-current and *future* model attributes to be mass-assigned.
-
-[`permit`]:
-    https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit
-[`permit!`]:
-    https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit-21
-[`expect`]:
-    https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect
-
-### Nested Parameters
-
-You can also use `expect` (or `permit`) on nested parameters, like:
-
-```ruby
-# Given the example expected params:
-params = ActionController::Parameters.new(
-  name: "Martin",
-  emails: ["me@example.com"],
-  friends: [
-    { name: "André", family: { name: "RubyGems" }, hobbies: ["keyboards", "card games"] },
-    { name: "Kewe", family: { name: "Baroness" }, hobbies: ["video games"] },
-  ]
-)
-# the following expect will ensure the params are permitted
-name, emails, friends = params.expect(
-  :name,                 # permitted scalar
-  emails: [],            # array of permitted scalars
-  friends: [[            # array of permitted Parameter hashes
-    :name,               # permitted scalar
-    family: [:name],     # family: { name: "permitted scalar" }
-    hobbies: []          # array of permitted scalars
-  ]]
-)
-```
-
-This declaration permits the `name`, `emails`, and `friends` attributes and
-returns them each. It is expected that `emails` will be an array of permitted
-scalar values, and that `friends` will be an array of resources (note the new
-double array syntax to explicitly require an array) with specific attributes.
-These attributes should have a `name` attribute (any permitted scalar values allowed), a
-`hobbies` attribute as an array of permitted scalar values, and a `family`
-attribute which is restricted to a hash with only a `name` key and any permitted
-scalar value.
-
-### Examples
-
-Here are some examples of how to use `permit` for different use cases.
-
-**Example 1**: You may want to use the permitted attributes in your `new` action.
-This raises the problem that you can't use [`require`][] on the root key
-because, normally, it does not exist when calling `new`:
-
-```ruby
-# using `fetch` you can supply a default and use
-# the Strong Parameters API from there.
-params.fetch(:blog, {}).permit(:title, :author)
-```
-
-**Example 2**: The model class method `accepts_nested_attributes_for` allows you to
-update and destroy associated records. This is based on the `id` and `_destroy`
-parameters:
-
-```ruby
-# permit :id and :_destroy
-params.expect(author: [ :name, books_attributes: [[ :title, :id, :_destroy ]] ])
-```
-
-**Example 3**: Hashes with integer keys are treated differently, and you can declare
-the attributes as if they were direct children. You get these kinds of
-parameters when you use `accepts_nested_attributes_for` in combination with a
-`has_many` association:
-
-```ruby
-# To permit the following data:
-# {"book" => {"title" => "Some Book",
-#             "chapters_attributes" => { "1" => {"title" => "First Chapter"},
-#                                        "2" => {"title" => "Second Chapter"}}}}
-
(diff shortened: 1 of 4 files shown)


guides/source/layouts_and_rendering.md
```diff
@@ -3,61 +3,67 @@
...
 
-From the controller's point of view, there are three ways to create an HTTP response:
-
-* Call [`render`][controller.render] to create a full response to send back to the browser
-* Call [`redirect_to`][] to send an HTTP redirect status code to the browser
-* Call [`head`][] to create a response consisting solely of HTTP headers to send back to the browser
+* [`render`][controller.render] to create a full response with a body, usually with a `2xx` status code.
+* [`redirect_to`][] to redirect the user to another path using an HTTP redirect status code.
+* [`respond_to`][] to enable rendering of multiple formats based on the HTTP request's `Accept` header.
+* [`head`][] to create a response consisting solely of HTTP headers without a body.
```
p8: We also have:
- send_data/send_file: https://edgeapi.rubyonrails.org/classes/ActionController/DataStreaming.html
- send_stream: https://edgeapi.rubyonrails.org/classes/ActionController/Live.html#method-i-send_stream

But I'm not sure we should mention these here as well.

Also respond_to doesn't create a HTTP response itself, it always uses one of the others.

guides/source/layouts_and_rendering.md
```diff
@@ -3,61 +3,67 @@
...
 
 [controller.render]: https://api.rubyonrails.org/classes/ActionController/Rendering.html#method-i-render
 [`redirect_to`]: https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to
 [`head`]: https://api.rubyonrails.org/classes/ActionController/Head.html#method-i-head
+[`respond_to`]: https://api.rubyonrails.org/classes/ActionController/MimeResponds.html#method-i-respond_to
+
+`render`ing Responses
```
p8: This looks a bit off in the h2:
Image

guides/source/layouts_and_rendering.md
```diff
@@ -3,61 +3,67 @@
...
 class BooksController < ApplicationController
 end
 ```
 
-And the following in your routes file:
-
 ```ruby
+# config/routes.rb
+
 resources :books
```
p8: The other examples for the controller and view work as-is, so I think we should show all of the required code for the route as well:
```suggestion
Rails.application.routes.draw do
  resources :books
end
```

guides/source/layouts_and_rendering.md
```diff
@@ -423,1017 +439,593 @@ Rails understands both numeric status codes and the corresponding symbols shown
...
-#### Linking to Audio Files with the `audio_tag`
+Even though we're using `render plain:`, the `content-type` HTTP header in the response will be set to `text/vnd.my-mime-type` since it's a format-specific handler.
 
-The [`audio_tag`][] helper builds an HTML5 `<audio>` tag to the specified file. By default, files are loaded from `public/audios`.
+NOTE: This is exactly how [Rails' integration](https://github.com/hotwired/turbo-rails) with [Turbo](https://turbo.hotwired.dev) creates Turbo Stream responses. It registers the `text/vnd.turbo-stream.html` MIME type which allows us to create `.turbo_stream.html` templates and use `format.turbo_stream` in `respond_to` blocks.
 
-```erb
-<%= audio_tag "music.mp3" %>
-```
+Redirecting Requests
```
p8: Redirecting and head responses don't use ActionView.
So I would expect to see these sections in the "ActionController Overview" guide instead of "Views > Layouts and Rendering".

This would make the "ActionController Overview" a bit bigger, but an overview should at least mention all relevant subjects.

guides/source/layouts_and_rendering.md
```diff
@@ -3,61 +3,67 @@
...
 [controller.render]: https://api.rubyonrails.org/classes/ActionController/Rendering.html#method-i-render
 [`redirect_to`]: https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to
 [`head`]: https://api.rubyonrails.org/classes/ActionController/Head.html#method-i-head
+[`respond_to`]: https://api.rubyonrails.org/classes/ActionController/MimeResponds.html#method-i-respond_to
+
+`render`ing Responses
+---------------------
 
-### Rendering by Default: Convention Over Configuration in Action
+The [`render`][controller.render] method is the primary workhorse to create HTTP responses in Rails. It's used to create a fully-formed response with a body which usually contains an HTML document.
```
p8: We introduce the `render` method here, but don't use it explicitly in the following examples.
Instead of highlighting "convention over configuration" I think newcomers would benefit more from starting with examples of explicitly using the `render` method.
After the `render` method has been shown we could mention the automagic conventions.

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
-the `:plain` option to `render`:
+The response will be rendered without a layout by default. Use the `layout:` option to render within a layout template:
 
 ```ruby
-render plain: "OK"
+# Renders within the default layout for the controller, usually: `app/views/layouts/application.html.erb`.
+render html: helpers.tag.strong("Not Found"), layout: true
+
+# Explitly define rendering within the `app/views/layouts/admin.html.erb` layout template.
+render html: helpers.tag.strong("Not Found"), layout: "admin"
```
p8: The comments require scrolling to read:

Image

```suggestion
# Renders within the default layout for the controller,
# usually: `app/views/layouts/application.html.erb`.
render html: helpers.tag.strong("Not Found"), layout: true

# Explitly define rendering within the `app/views/layouts/admin.html.erb`
# layout template.
render html: helpers.tag.strong("Not Found"), layout: "admin"
```

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
 
-TIP: Rendering pure text is most useful when you're responding to Ajax or web
-service requests that are expecting something other than proper HTML.
+NOTE: There's rarely a good reason to do this in practice. Use a template file to render HTML responses.
 
-NOTE: By default, if you use the `:plain` option, the text is rendered without
-using the current layout. If you want Rails to put the text into the current
-layout, you need to add the `layout: true` option and use the `.text.erb`
-extension for the layout file.
+WARNING: When using `html:` option, HTML entities will be escaped if the string is not composed with `html_safe`-aware APIs.
```
p8:
```suggestion
WARNING: When using the `html:` option, HTML entities will be escaped if the string is not composed with `html_safe`-aware APIs.
```

guides/source/action_controller_overview.md
```diff
@@ -53,10 +53,36 @@ end
 Given the above `ClientsController`, if a user goes to `/clients/new` in your
 application to add a new client, Rails will create an instance of
 `ClientsController` and call its `new` method. If the `new` method is empty, Rails
-will automatically render the `new.html.erb` view by default.
+will automatically render `app/views/clients/new.html.erb` by default.
```
p8: Somewhat out of scope, but I think this paragraph would be clearer as a numbered list:
```suggestion
Given the above `ClientsController`, if a user goes to `/clients/new` in your
application to add a new client, Rails will:

* create an instance of `ClientsController`
* call its `new` method
* if the `new` method is empty, automatically render `app/views/clients/new.html.erb` by default
```

guides/source/layouts_and_rendering.md
p8: Similar to the "Active Record Basics" guide I would expect the "ActionController Overview" guide to be the starting point and mention all important features of ActionController. More specific guides can than expand on those features (I always had trouble finding what was where in the ActionController guides).

The "ActionController Overview" currently doesn't explain `redirect_to` or `head`.
I would expect the "ActionController Overview" to at least mention them as these are some of the most important methods of a controller.
So I think the description of the responsibility of a Controller and the methods to create a response, should at least be copied to the "ActionController Overview" guide.

guides/source/action_controller_overview.md
p8, line 79: I'm wondering if we should start a bit simpler here for new comers that have never seen a controller.
Just render "Hello, world" as the first example, before diving into all the details of rendering. This would also avoid the confusion with `new` method till further down.
```suggestion
  def new
     render_to_string "Hello World!"
```

Or maybe just start with a simpler (non-crud) controller that renders plain (this would also require a routing code example, but that applies to the ClientsController as well):
````
```ruby
class WelcomeController < ApplicationController
  def greet
    render plain: "Hello World!"
  end
end
```
````

It would be nice if we could just call `#greet`, but that sadly doesn't work.
`WelcomeController.new.greet`.

Apparently we can call `.action`, which returns a Rack handler (I'm, not recommending we should use this 😄 ):
```
WelcomeController.action(:greet).call({}).last.body
=> "Hello World!"
```

File: guides/source/action_controller_overview.md [hunk 9]
```diff
@@ -53,10 +53,36 @@ end
 Given the above `ClientsController`, if a user goes to `/clients/new` in your
```
p8: Somewhat out of scope, but I think we should add a header here.
Rendering responses is probably the most important function of a controller, it should at least be visible in the sidebar.
(this header should probably be moved up a bit depending on changes above).
```suggestion
### Rendering responses

Given the above `ClientsController`, if a user goes to `/clients/new` in your
```

guides/source/action_controller_overview.md
p8: I'm wondering if we should add a NOTE here to explaining the concept of an action?

guides/source/action_view_overview.md
```diff
@@ -476,19 +476,40 @@ variable named after the partial. In this case, since the partial is
...
+the corresponding partial for each one.
+
+```html+erb
+<%# customers/_customer.html.erb -%>
+<p>Customer: <%= customer.name %></p>
+
+<%# employees/_employee.html.erb -%>
+<p>Employee: <%= employee.name %></p>
+
+<%# index.html.erb -%>
```
p8, line 514: Maybe use separate examples here to avoid confusion, and allow easier copy/pasting?
````suggestion
```html+erb
<%# customers/_customer.html.erb -%>
<p>Customer: <%= customer.name %></p>
```

```html+erb
<%# employees/_employee.html.erb -%>
<p>Employee: <%= employee.name %></p>
```

```html+erb
<%# index.html.erb -%>
````

File: guides/source/action_view_overview.md [hunk 11]
```diff
@@ -643,7 +664,7 @@ values, and more with a `locals:` signature, using the same syntax as Ruby metho
 Here are some examples of the `locals:` signature:
 
 ```html+erb
-<%# app/views/messages/_message.html.erb %>
+<!-- app/views/messages/_message.html.erb -->
```
p8: I think we want to keep these as Ruby comments, as Rails already adds HTML comments when [annotate_rendered_view_with_filenames](https://guides.rubyonrails.org/configuring.html#config-action-view-annotate-rendered-view-with-filenames) is set to true (default in development).
Also applies to the following examples.

guides/source/action_view_overview.md
```diff
@@ -779,9 +801,156 @@ Here is an example of a simple layout in `application.html.erb` file:
...
+  </head>
+  <body>
+    <%= yield %>
+  </body>
+</html>
+```
+
+The main body of the view will always render into the unnamed `yield`. To render content into a named `yield`, call `content_for` with the same argument as the named `yield`.
+
+#### Slotting content with `content_for`
```
p8:
```suggestion
#### Slotting Content with `content_for`
```

guides/source/action_view_overview.md
```diff
@@ -779,9 +801,156 @@ Here is an example of a simple layout in `application.html.erb` file:
...
+
+The [`content_for`][] method allows you to insert content into a named `yield` block in your layout. For example, consider following layout and view:
```
p8:
```suggestion
The [`content_for`][] method allows you to insert content into a named `yield` block in your layout. For example, consider the following layout and view:
```

guides/source/action_view_overview.md
```diff
@@ -247,17 +240,23 @@ method within the view:
 ```
 
 This will look for a file named `_product.html.erb` in the same folder to render
-within that view. Partial file names start with leading underscore character by
+within that view. If no such file exists, Rails will look for it under the
+`app/view/application/` folder. This makes `app/views/application/` a great
```
p8, line 248: This only applies in the context of ActionController and if the controller inherits from ApplicationController, as the resolver will look up the inheritance chain to see if parent controllers have defined the view.
Maybe link to "layouts_and_rendering#template-lookup-hierarcy" for the explanation?

guides/source/action_view_overview.md
```diff
@@ -502,31 +523,31 @@ main partial by using the `:spacer_template` option:
 Rails will render the `_product_ruler.html.erb` partial (with no data passed to
 it) between each pair of `_product.html.erb` partials.
 
-### Counter Variables
+### Metadata Variables
```
p8: The class that defines these is called PartialIteration.
So maybe we can make it more specific?
```suggestion
### Iteration Variables
```

guides/source/layouts_and_rendering.md
```diff
@@ -423,1017 +439,593 @@ Rails understands both numeric status codes and the corresponding symbols shown
...
-    ```
+[`layout`]: https://api.rubyonrails.org/classes/ActionView/Layouts/ClassMethods.html#method-i-layout
 
-* `employees/_employee.html.erb`
+### Setting Layouts Dynamically
 
-    ```html+erb
-    <p>Employee: <%= employee.name %></p>
-    ```
+Invoke the `layout` method with a symbol denoting a method name to select a layout dynamically at runtime.
```
p8: As a non-native speaker I'm not sure all readers know the meaning of `invoke`.
Maybe use the more common `call` instead?
```suggestion
Call the `layout` method with a symbol denoting a method name to select a layout dynamically at runtime.
```

guides/source/layouts_and_rendering.md
```diff
@@ -423,1017 +439,593 @@ Rails understands both numeric status codes and the corresponding symbols shown
...
+      Current.user.admin? ? "admin" : "products"
+    end
+end
 ```
 
-#### Local Variables
+Now, if the current user is a administrator, they'll get the admin-specific layout when viewing a product.
 
-To use a custom local variable name within the partial, specify the `:as` option in the call to the partial:
+You can also invoke `layout` with a [`Proc`](https://docs.ruby-lang.org/en/master/Proc.html) or [`lambda`](https://docs.ruby-lang.org/en/master/Kernel.html#method-i-lambda). The controller instance will be passed into it.
```
p8:
```suggestion
You can also call `layout` with a [`Proc`](https://docs.ruby-lang.org/en/master/Proc.html) or [`lambda`](https://docs.ruby-lang.org/en/master/Kernel.html#method-i-lambda). The controller instance will be passed into it.
```

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
-  end
-end
+render action: :edit, status: :unprocessable_content
 ```
 
-#### Rendering an Action's Template from Another Controller
+NOTE: We render with the HTTP status [`303 Unprocessable Content`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/303) in the above example because it's the idiomatically correct HTTP response code for a form submission error. It's also required by the [Turbo](https://turbo.hotwired.dev) JavaScript library which Rails includes by default.
 
-What if you want to render a template from an entirely different controller from the one that contains the action code? You can also do that with `render`, which accepts the full path (relative to `app/views`) of the template to render. For example, if you're running code in an `AdminProductsController` that lives in `app/controllers/admin`, you can render the results of an action to a template in `app/views/products` this way:
+To render a template which belongs to a different controller, you can use its relative path from the `app/views/` directory. For example, to render `app/views/products/show.html.erb` from the `BooksController`, you'd invoke:
```
p8:
```suggestion
To render a template which belongs to a different controller, you can use its relative path from the `app/views/` directory. For example, to render `app/views/products/show.html.erb` from the `BooksController`, you'd call:
```

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
 
-If the call to `update` fails, calling the `update` action in this controller will render the `edit.html.erb` template belonging to the same controller.
+You can also use a symbol instead of a string to specify the name of the template:
+
+```ruby
+render :edit, status: :unprocessable_content
+```
 
-If you prefer, you can use a symbol instead of a string to specify the action to render:
+You can use the `action:` option to be more explicit:
```
p8, line 186: I don't think we should mention it in the guide but...
`render :edit`
gets [normalized](https://github.com/rails/rails/blob/e2c7935ab2e217c88cf2acfb8c5c03719361e311/actionview/lib/action_view/rendering.rb#L160-L162) to
`render action: "edit"`
which gets [normalized](https://github.com/rails/rails/blob/e2c7935ab2e217c88cf2acfb8c5c03719361e311/actionview/lib/action_view/rendering.rb#L187) to:
`render template: "edit"` 
So `:action` and `:template` basically work the same as they all finally get turned into `:template`.

Ah ok, that was mentioned in the original guide. I think that still adds value as it's confusing to me as well.

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
 ```ruby
 render template: "products/show"
 ```
 
-#### Wrapping it up
-
-The above two ways of rendering (rendering the template of another action in the same controller, and rendering the template of another action in a different controller) are actually variants of the same operation.
-
-In fact, in the `BooksController` class, inside of the update action where we want to render the edit template if the book does not update successfully, all of the following render calls would all render the `edit.html.erb` template in the `views/books` directory:
+Note that rendering a template using this technique does not call the other controller action. Consider the below example:
```
p8: To me this is more confusing for the `:action` argument: `render action: "products/show"`.
As a reader I would wonder why would I use `:action` instead of `:template`? Does it render more of the action?

guides/source/layouts_and_rendering.md
```diff
@@ -101,258 +105,269 @@ If we want to display the properties of all the books in our view, we can do so
...
 render file: "#{Rails.root}/public/404.html", layout: false
 ```
 
-This renders the raw file (it doesn't support ERB or other handlers). By
-default it is rendered within the current layout.
+NOTE: The layout that a static file is inserted into can written using a templating language like ERB. However, the file supplied to the `file:` option must be static. It will not be processed through a templating engine.
 
-WARNING: Using the `:file` option in combination with users input can lead to security problems
-since an attacker could use this action to access security sensitive files in your file system.
+WARNING: Using the `:file` option in combination with users input can lead to security problems since an attacker could use this action to access security sensitive files in your file system.
```
p8: Nitpick that I'm not sure about, but I would use "on" here:
```suggestion
WARNING: Using the `:file` option in combination with users input can lead to security problems since an attacker could use this action to access security sensitive files on your file system.
```

guides/source/layouts_and_rendering.md
```diff
@@ -423,1017 +439,593 @@ Rails understands both numeric status codes and the corresponding symbols shown
...
-With this set of variants Rails will look for the following set of templates and use the first that exists.
+`render_to_string` doesn't return any headers, just the response body. Hence, options such as `status:` are ignored.
 
-- `app/views/home/index.html+mobile.erb`
-- `app/views/home/index.html+desktop.erb`
-- `app/views/home/index.html.erb`
+### Avoiding Double Render Errors
 
-If a template with the specified format does not exist an `ActionView::MissingTemplate` error is raised.
+The `render` method **does not** `return` from the current scope. Lines after a `render` call will still be executed. Calling `render` multiple times in the same controller action is not allowed and will raise a `AbstractController::DoubleRenderError`.
```
p8: This also applies to `redirect_to`, `head` etc, should we mention these as well?

guides/source/layouts_and_rendering.md
```diff
@@ -423,1017 +439,593 @@ Rails understands both numeric status codes and the corresponding symbols shown
...
-To render a partial as part of a view, you use the [`render`][view.render] method within the view:
-
-```html+erb
-<%= render "menu" %>
+```ruby
+head :bad_request
 ```
 
-This will render a file named `_menu.html.erb` at that point within the view being rendered. Note the leading underscore character: partials are named with a leading underscore to distinguish them from regular views, even though they are referred to without the underscore. This holds true even when you're pulling in a partial from another folder:
+This would produce the following headers:
```
p8:
```suggestion
This would produce the following response headers:
```
Or, as it's more than headers:
```suggestion
This would produce the following HTTP response:
```

guides/source/action_controller_overview.md
```diff
@@ -139,7 +165,7 @@ class ClientsController < ApplicationController
     if @client.save
       redirect_to @client
     else
-      render "new"
+      render "new", status: :unprocessable_content
     end
   end
 end
```
p8: This describes specific interaction between the controller and the view.
Maybe move this to the "Layouts and Rendering" guide, which we just linked to?

guides/source/action_controller_overview.md
ayushn21: I don't think a `NOTE` is the right solution, but I agree we should explain a bit more. I've reworded the paragraph to say the request is matched to a "controller and action", and also linked to the routing guide so the reader can click through for further context.

I've also added the route to the below example which should provide further context. The definition and explanation of action should live in the routing guide I believe, so I don't want to add too much detail here.

What do you think?

ayushn21: I'm not keen on this approach, I think we'd end up confusing the reader more than simplifying the concept. Non-crud actions is not something Rails recommends, so I don't think we should start a tutorial by teaching stuff that isn't recommended.

A little bit of confusion when trying to grasp a new concept, I think, is fine. 

That said, the `new` method confusion isn't ideal so I've reworked the example to use `index` instead which should be easier to understand.

guides/source/action_controller_overview.md
ayushn21: I agree with your sentiment here but I can't find an appropriate break in the prose to put in this heading. I think it would only disrupt the flow rather than add to it.

I also don't think "Creating a controller" is a very good section heading. We're not teaching how to generate a controller, it focuses more on structuring controllers, so maybe we should change the section heading to "Structuring Controllers and Rendering Responses"? 

Also I should add, this PR was focused on the Layouts and Rendering guide ... I tried to keep the changes to any other guides as minimal as possible, especially since they have already been updated. 

If you'd like I can spend some more time to tweak the action controller guide as well and maybe move stuff around.

guides/source/action_controller_overview.md
ayushn21: I honestly think it reads better as prose. I changed it to a list and it feels clunky to me. I've reworded it slightly anyway so let me know what you think

guides/source/action_controller_overview.md
ayushn21: This section is about processing parameters which I think belongs in the controller guide. I only added the `status` to this line as it's best practice (and needed for Turbo), so we should consistently use it across the guides. It is not, strictly speaking, relevant to the other changes.

guides/source/action_view_overview.md
ayushn21: Good call, in this guide I think it's ok to assume the controller would inherit from `ApplicationController`, but definitely good to signpost for advanced users. Added a note with a link.

guides/source/action_view_overview.md
ayushn21: I think these are all very closely related as they essentially build the same demo. I can't see why people would want to copy / paste these as they're demo snippets, not usable examples.

Breaking them up into different code blocks would just break up the readers's flow to my mind.

guides/source/action_view_overview.md
ayushn21: Yes, I much prefer `Iteration Variables`, nice one.

guides/source/action_view_overview.md
ayushn21: Fixed, not sure how I managed to do that!

guides/source/layouts_and_rendering.md
ayushn21: I don't think we need to mention these here as it's not quite "rendering a response", but more just sending data on disk. We cover it in the controller guide I believe.

> Also respond_to doesn't create a HTTP response itself, it always uses one of the others.

I know but I don't want to get too much into the weeds in the intro. I think the current line explains what it does at a glance and we can dig into it later on. I'll take a look at what I've written for `respond_do` in the main explanation and add this line if I feel it's needed!

ayushn21: Ok yeah I don't disagree here, but like I said in an earlier comment, I wanted to keep my changes to other guides as minimal as possible. 

If you'd like me to spend some time on the controller guide to improve it and make these improvements while ensuring the flow and relationship with this guide works better, let me know. It will be a slightly bigger job hence I was restrained in the changes I made while working on this guide.

guides/source/layouts_and_rendering.md
ayushn21: Yeah not great, changed to `Rendering Responses`.

guides/source/layouts_and_rendering.md
ayushn21: Yep good shout. I've reworked the section to demo `render` immediately and then demonstrate conventional rendering right after.

guides/source/layouts_and_rendering.md
ayushn21: Added a note explaining this.

guides/source/layouts_and_rendering.md
ayushn21: Reworded slightly differently

guides/source/layouts_and_rendering.md
ayushn21: Reworded to:

> Rendering a template for a different controller action does not call the method corresponding to that controller action. 

This along with the added note explaining that `action:` and `template:` are functionally equivalent should hopefully prevent any confusion.

guides/source/layouts_and_rendering.md
ayushn21: Yep, "on" is the better word, thanks!

guides/source/layouts_and_rendering.md
ayushn21: Yes, but not here I think. We haven't covered those as yet so we shouldn't drop them in here. I will add mentions of this issue further on the guide when we cover those methods.

guides/source/layouts_and_rendering.md
ayushn21: Yeah I think you're right here. Let me have a think about it a bit more though.

guides/source/layouts_and_rendering.md
ayushn21: I see what you mean, but I see this as an opportunity to teach the reader a new word. It's easy enough to look up words these days, especially on Mac when you can just long click the word. Also I think there's enough context here to infer the meaning of the word.

I believe this is how people become better readers and writers and I don't think we should avoid using certain words just because they might not be known to everyone.

That said, I've changed it in this case as I feel I've overused the word across this guide. I'm keen to keep it in some other places though.

ayushn21:
As per feedback, I've given the Action Controller guide a full once-over. There were several aspects I felt needed improvement. As such, in addition to moving the "Redirecting Requests" and "Header-only responses" sections over from the Layouts and Rendering guide, I've made the following improvements:

* Added description of CRUD methods to intro.
* Restructured intro headings.
* Corrected minor technical errors.
* Moved redirection and head responses from layouts and rendering guide.
* Fixed technical errors in wrapped parameter section and examples.
* Enhanced the section on the strong paramteres `expect` method, since it is the recommended approach.
* Added `require` demo to strong parameters.
* Moved info on permitted scalars further up so it applies to parameters in general.
* Renamed "strong parameters" section and moved under "Request Parameters".
* Restructured section so examples flows with the prose and removed the examples and nested parameters sections. Also added link to API docs for advanced usage.
* Simplified cookies demo to remove all extraneous code.
* Fixed the section on Sessions which says they're saved on the server-side.
* Move session stores section above the flash to enhance narrative flow.
* Moved controller callbacks above request parameters as they are a key structural topic for controllers.

guides/source/action_controller_overview.md
p8: I was looking at the "Getting Started" guide and it mentions "actions" but doesn't explain them.
The routing guide mentions "controller actions" a bunch but only has a bare explanation: "... it asks the router to match it to a controller action (aka method)".
Since these are named "controller actions" I would expect the "ActionController Overview" guide to describe them. I'll add this to the basecamp discussion about this guide.

guides/source/action_view_overview.md
```diff
@@ -247,17 +240,25 @@ method within the view:
...
 
 This will look for a file named `_product.html.erb` in the same folder to render
-within that view. Partial file names start with leading underscore character by
+within that view. If no such file exists, Rails will look for it under the
+`app/view/application/` folder. This makes `app/views/application/` a great
+place for your shared partials.
+
+NOTE: Rails doesn't automatically create `app/views/application/`. You'll need to create this folder yourself.
+
+NOTE: Refer to the (Layouts and Rendering guide)[layouts_and_rendering.html#template_lookup_hierarchy] for further information on the lookup hierarchy of partials and template.
```
p8:
```suggestion
NOTE: Refer to the [Layouts and Rendering guide](layouts_and_rendering.html#template_lookup_hierarchy) for further information on the lookup hierarchy of partials and template.
```

guides/source/layouts_and_rendering.md
p8: Yeah, that makes sense.

guides/source/layouts_and_rendering.md
p8: That looks great!

guides/source/layouts_and_rendering.md
p8: Nice!

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+If you limit your controllers to the 7 CRUD actions, this won't be an issue.
 
 NOTE: If you must use a reserved method as an action name, one workaround is to
 use a custom route to map the reserved method name to your non-reserved action
 method.
 
-[Resource Routing]: routing.html#resource-routing-the-rails-default
+### Rendering Responses
+
+Consider the below controller and route.
```
p8: Nitpick, we first show the route, then the controller:
```suggestion
Consider the below route and controller.
```

guides/source/action_controller_overview.md
p8: Nevermind, I really like your explanation in the "ActionController Overview" guide!

guides/source/action_controller_overview.md
p8: I like the reworded version.

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+```
+
+NOTE: When redirecting using `302 Found`, the client may not always make a `GET` request to the `Location`. JavaScript's `fetch` method will make a `GET` request after redirection from a `GET` or `POST` request. But, if the initial request is another method, for example `PUT`, it will use the same method when requesting the redirected location. This is not a problem in Rails because all [form submissions use `POST`](https://guides.rubyonrails.org/form_helpers.html#forms-with-patch-put-or-delete-methods), even when using [Turbo](http://turbo.hotwired.dev). <br><br> When making requests in custom JavaScript, redirect using [`303 See Other`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/303) to ensure a `GET` request to the redirected location, or [`307 Temporary Redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/307) to preserve the original HTTP verb.
+
+NOTE: It's worth being aware that `redirect_to` doesn't move execution to a different method within the same request. It sends an HTTP response and then the browser makes a new request to the redirected location which won't have any context from the previous request.
```
p8: This NOTE is easily missed with the Javascript one above it, and I think a lot of new comers run into the described issue.  (I'm also not a fan of how sequential NOTEs look).

Maybe we can move this up a bit higher, just above the `redirect_back` explanation?

guides/source/action_controller_overview.md
p8: The reworked section looks great.

guides/source/action_view_overview.md
p8: I guess the confusing part to me were the last two lines. 
The first three examples are a comment with the file location and a single erb line.
The last two lines also have a comment but it doesn't describe the file location.
So I had to take a second look to see where those lines belong.

Maybe remove the linebreak after `<h1>Contacts</h1>` to make it clear these belong together?
Or use a separate example for the `index.html.erb` only and maybe move the last comment above outside the example?

p8:
I'm not sure about moving "Callbacks" before "Requests Params" in the ActionController guide.
You can get pretty far creating an application without using any callbacks, but request params are pretty essential.

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
 
 The controller naming convention is different from models. While plural names
 are preferred for controller names, the singular form is preferred for [model
 names](active_record_basics.html#naming-conventions) (e.g. `Account` vs.
 `Accounts`).
 
-Controller actions should be `public`, as only `public` methods are callable as
-actions. It is also best practice to lower the visibility of helper methods
-(with `private` or `protected`) which are _not_ intended to be actions.
+Controller actions should be _public_, as only _public_ methods are callable as
```
p8, line 140: Unrelated to this PR, but I think it would be useful to define in the guidelines when to use italic or bold for words.

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+  def index
+    respond_to do |format|
+      format.html   # renders `app/views/posts/index.html.erb`
+      format.xml    # renders `app/views/posts/index.xml.erb`
+    end
+  end
+end
+```
+
+See the [Layouts and Rendering guide](layouts_and_rendering.html) for futher details.
```
p8: I think we can be more specific here and link to the section about rendering additional formats:
```suggestion
See the [Layouts and Rendering guide](layouts_and_rendering.html#multi-format-responses) for further details on rendering additional formats.
```

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+# or
+head 400
+```
+
+This would produce the following HTTP response:
+
+```http
+HTTP/1.1 400 Bad Request
+Connection: close
+Date: Sun, 24 Jan 2010 12:15:53 GMT
```
p8: We can probably update this date to make Rails seem a bit younger ;):
```suggestion
Date: Fri, 10 Juli 2026 12:15:53 GMT
```

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+      render "new", status: :unprocessable_content
     end
   end
 end
 ```
 
-NOTE: The `params` hash is not a plain old Ruby Hash; instead, it is an
-[`ActionController::Parameters`][] object. It does not inherit from Hash, but it behaves mostly like Hash.
-It provides methods for filtering `params` and, unlike a Hash, keys `:foo` and `"foo"` are considered to be the same.
+NOTE: [`ActionController::Parameters`][] does not inherit from Hash, but it is similar in behavior. Although, unlike a Hash, symbolic and string keys, such as `:foo` and `"foo"`, are considered to be the same.
```
p8: I kind of liked the original sentence using "mostly", as it is more corrrect. It also allows simplifying the second sentence.
Maybe:
```suggestion
NOTE: [`ActionController::Parameters`][] does not inherit from Hash, but it is mostly similar in behavior. Unlike a Hash, symbolic and string keys, such as `:foo` and `"foo"`, are considered to be the same.
```

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
-contain nested arrays and hashes. To send an array of values, append an empty
-pair of square brackets `[]` to the key name:
+Values in an [`ActionController::Parameters`][] object must be a permitted scalar value defined in `ActionController::Parameters::PERMITTED_SCALAR_TYPES`.
+
+```ruby
+ActionController::Parameters::PERMITTED_SCALAR_TYPES
+# => [String, Symbol, NilClass, Numeric, TrueClass, FalseClass, Date, Time, StringIO, IO, ActionDispatch::Http::UploadedFile, Rack::Test::UploadedFile]
+```
+
+The object can contain nested hashes and array, but values within those must also contain a permitted scalar.
```
p8:
```suggestion
The object can contain nested hashes and arrays, but values within those must also contain a permitted scalar.
```

guides/source/action_controller_overview.md
```diff
@@ -190,23 +544,34 @@ To send a hash, you include the key name inside the brackets:
 </form>
 ```
 
-When this form is submitted, the value of `params[:user]` will be `{ "name" =>
-"Acme", "phone" => "12345", "address" => { "postcode" => "12345", "city" =>
-"Carrot City" } }`. Note the nested hash in `params[:user][:address]`.
+When this form is submitted, the value of `params[:user]` will be:
+
+```
```
p8:
```suggestion
```ruby
```

guides/source/action_controller_overview.md
```diff
@@ -190,23 +544,34 @@ To send a hash, you include the key name inside the brackets:
...
 
 [Composite key parameters](active_record_composite_primary_keys.html) contain
-multiple values in one parameter separated by a delimiter (e.g., an underscore).
+multiple values in one parameter separated by a delimiter (such as an underscore).
 Therefore, you will need to extract each value so that you can pass them to
-Active Record. You can use the `extract_value` method to do that.
+Active Record. You can use the [`extract_value`][] method to do that.
 
-For example, given the following controller:
+Consider the below controller and route:
```
p8, line 641: I think mentioning both reads a bit easier than splitting it up 👍

guides/source/action_controller_overview.md
```diff
@@ -245,80 +610,77 @@ Your controller will receive:
...
+
+NOTE: While the action and controller names are available in the params object,
+it is recommended to use the methods [`controller_name`][] and [`action_name`][] instead to access these values.
 
-This feature clones and wraps parameters with a key chosen based on your
-controller's name. It is configured to `true` by default. If you do not want to
-wrap parameters you can
-[configure](configuring.html#config-action-controller-wrap-parameters-by-default)
-it to `false`.
+Parameter wrapping is enabled by default, can be disabled using a configuration option:
```
p8: This sentence seems to miss some words.
Maybe:
```suggestion
Parameter wrapping is enabled by default, but can be disabled using a configuration option:
```

guides/source/action_controller_overview.md
```diff
@@ -245,80 +610,77 @@ Your controller will receive:
...
 
 ```ruby
+# config/application.rb
+
 config.action_controller.wrap_parameters_by_default = false
 ```
```
p8, line 726: Maybe just link to the configuration guide only. That guide should document how configuration can be set.

guides/source/action_controller_overview.md
```diff
@@ -245,80 +610,77 @@ Your controller will receive:
...
+    https://api.rubyonrails.org/classes/ActionController/Metal.html#method-i-controller_name
+[`action_name`]:
+    https://api.rubyonrails.org/classes/AbstractController/Base.html#method-i-action_name
 
 ### Routing Parameters
 
 Parameters specified as part of a route declaration in the `routes.rb` file are
 also made available in the `params` hash. For example, we can add a route that
-captures the `:status` parameter for a client:
+captures the `:status` parameter for a posts:
```
p8:
```suggestion
captures the `:status` parameter for a post:
```

p8: As a side note, in this context `post` could also be interpreted as a POST request.

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+  def index
+    # Display all posts
+  end
+
+  # GET /posts/new
+  def new
+    # Render a form to create a new post
+  end
+
+  # POST /posts
```
p8: Having post refer to the request method and resource here, might confuse some new comers.
I also find `posts` somewhat abstract, not everyone know this refers to BlogPosts or NewsPosts.
If we are going to rename ClientsController, maybe use the more relatable ProductsController or BooksController?

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+
+### Redirecting Requests
+
+Instead of rendering a fully-formed response, such as an HTML document, you might want to redirect the user to a different path. An [HTTP redirection status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#redirection_messages) can be used for this. Rails' [`redirect_to`][] method uses `302 Found` by default.
+
+```ruby
+redirect_to photos_url
+```
+
+This will send a `302` HTTP response to the browser with `photos_url` in the `Location` header. The browser will then make a new `GET` request to the `photos_url`.
```
p8: Similar to the `head` method, maybe we can show an example redirect response here?

guides/source/action_controller_overview.md
```diff
@@ -761,208 +1033,64 @@ not accessing them will do the job.
...
-[`flash.now`]:
-    https://api.rubyonrails.org/classes/ActionDispatch/Flash/FlashHash.html#method-i-now
-
 ### Session Stores
 
-All sessions have a unique ID that represents the session object; these session
-IDs are stored in a cookie. The actual session objects use one of the following
+The storage mechanism for session data can be configured. By default, data is stored in an encrypted cookie, but other stores are available.
+
+All sessions have a unique ID representing the session object. Regardless of the chosen store, this ID is always stored in a cookie. The data can be stored using one of the following
```
p8: Introducing the terms "session ID" and "session data" here makes this paragraph a bit clearer to me. We also use these terms exclusively for the rest of this section.
Maybe:
```suggestion
All sessions have a unique ID representing the session object. Regardless of the chosen store, this session ID is always stored in a cookie. The session data can be stored using one of the following
```

guides/source/action_controller_overview.md
```diff
@@ -1001,191 +1114,155 @@ to rotate existing sessions.
...
 
-The `after_action` callbacks are similar to `before_action` callbacks, but
-because the controller action has already been run they have access to the
-response data that's about to be sent to the client.
+[`ActionDispatch::Flash::FlashHash`]:
+    https://api.rubyonrails.org/classes/ActionDispatch/Flash/FlashHash.html
 
-NOTE: `after_action` callbacks are executed only after a successful controller
-action, and not if an exception is raised in the request cycle.
+#### Displaying flash messages
```
p8:
```suggestion
#### Displaying Flash Messages
```

guides/source/action_controller_overview.md
```diff
@@ -1001,191 +1114,155 @@ to rotate existing sessions.
...
-`after_action` code.)
+[`flash.keep`][] is used to carry over the flash value through to an additional
+request. This is useful when there are multiple redirects.
 
-[`after_action`]: https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-after_action
-[`around_action`]: https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-around_action
+For example, assume that the `root_url` routes to the `index` action in the controller below, and all requests here are redirected to `UsersController#index`.
 
-### Other Ways to Use Callbacks
+If an action sets the flash and redirects to `MainController#index`, those flash values will be lost when during the next redirect.
```
p8:
```suggestion
If an action sets the flash and redirects to `MainController#index`, those flash values will be lost during the next redirect.
```

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+```ruby
+head :created, location: photo_path(@photo)
+```
+
+Which would produce:
+
+```http
+HTTP/1.1 201 Created
+Connection: close
+Date: Sun, 24 Jan 2010 12:16:44 GMT
```
p8:
```suggestion
Date: Fri, 10 Juli 2026 12:16:44 GMT
```

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,512 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+
+Data sent by the incoming request is available in your controller using the
+[`params`][] method which returns an [`ActionController::Parameters`][] object.
+
+There are two types of parameter data:
 
 - Query string parameters which are sent as part of the URL (for example, after
   the `?` in `http://example.com/accounts?filter=free`).
-- POST parameters which are submitted from an HTML form.
+- Parameters defined in the request body in non-`GET` requests,
```
p8: To me this list would be more consistent if we list the two types in a similar way.
This is also how these parameters are referred to later on:
```suggestion
- Request body parameters in non-`GET` requests,
```

guides/source/action_controller_overview.md
```diff
@@ -761,208 +1033,64 @@ not accessing them will do the job.
...
 
-All sessions have a unique ID that represents the session object; these session
-IDs are stored in a cookie. The actual session objects use one of the following
+The storage mechanism for session data can be configured. By default, data is stored in an encrypted cookie, but other stores are available.
+
+All sessions have a unique ID representing the session object. Regardless of the chosen store, this ID is always stored in a cookie. The data can be stored using one of the following
 storage mechanisms:
 
-* [`ActionDispatch::Session::CookieStore`][] - Stores everything on the client.
+* [`ActionDispatch::Session::CookieStore`][] - Stores data in an encrypted cookie.
```
p8: The other items use "Stores the data":
```suggestion
* [`ActionDispatch::Session::CookieStore`][] - Stores the data in an encrypted cookie.
```

guides/source/action_view_overview.md
ayushn21: Ahhhh ok I see where you're coming from now. I've split this into two code blocks, first one is the two partials, and second one is `index.erb` only.

ayushn21:
> I'm not sure about moving "Callbacks" before "Requests Params" in the ActionController guide. You can get pretty far creating an application without using any callbacks, but request params are pretty essential.

It's a bit of a judgement call. My view is that you can't get very far without callbacks either for any non-trivial app. You'd use them either to load records into memory, or for authentication at a minimum. I'd say they're as essential as request params, and since we start the guide talking about "controller structure", I wanted to continue in that theme to talk about callbacks, before diving deeper into the request params and other internal details.

Do you still think we should put request params before callbacks?

guides/source/action_view_overview.md
ayushn21: Bugger 😅. Fixed!

ayushn21 (commented):
Nice catches @p8  ... I've address all your comments. Let me know if there's anything else, otherwise will clean up the commit history and rebase on `main` :)

Thanks!

EDIT: Missed a few comments as they were hidden, addressing now.

guides/source/action_controller_overview.md
ayushn21: I used `Posts` to be keep with the "building a blog" theme in Rails tutorials, but I take your point here. I think `ProductsController` is a better example, will update.

guides/source/action_controller_overview.md
ayushn21: Yep, I can write a proposal in Basecamp of you wish?

guides/source/action_controller_overview.md
ayushn21: This is actually a callout to that guide for rendering in general, not just rendering multi format responses. I'll make it clearer to convey that.

guides/source/action_controller_overview.md
ayushn21: Good shout, done!

guides/source/action_controller_overview.md
ayushn21: Done!

guides/source/action_controller_overview.md
ayushn21: Updated as suggested! I didn't like the phrasing of the original sentence, and usually don't like vague terms like "mostly". In this case we point out a key difference in the second sentence so I think it works.

guides/source/action_controller_overview.md
ayushn21: Done ... dyslexia wreaking havoc here!

guides/source/action_controller_overview.md
ayushn21: That guide is a beast though, I think linking to it is fine when a detailed explanation is required, as that's the correct home for that kind of detail. If it's just a setting, I think keeping it in this guide is more reader friendly as they don't need to go elsewhere just for the syntax of a single setting.

guides/source/action_controller_overview.md
ayushn21: Yeah I'll update this example to `product` along with the rest of the guide.

ayushn21 (commented):
Think I've addressed everything now, thanks!

guides/source/action_controller_overview.md
p8: Ah you removed it. Nice!

ayushn21: Yeah, it wasn't sent as a header on my test Rails 8 app, so figured there's no point having it there!

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,517 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+
+Instead of rendering a fully-formed response, such as an HTML document, you might want to redirect the user to a different path. An [HTTP redirection status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#redirection_messages) can be used for this. Rails' [`redirect_to`][] method uses `302 Found` by default.
+
+```ruby
+redirect_to photos_url
+```
+
+This will send a `302` HTTP response to the browser with `photos_url` in the `Location` header. The browser will then make a new `GET` request to the `photos_url`.
+
+```
```
p8: We can use a http code block here to format it:
```suggestion
```http
```

guides/source/action_controller_overview.md
p8: The new "Getting Started" guide now builds a store and creates a ProductsController as well.

p8:
> It's a bit of a judgement call. My view is that you can't get very far without callbacks either for any non-trivial app. You'd use them either to load records into memory, or for authentication at a minimum. I'd say they're as essential as request params, and since we start the guide talking about "controller structure", I wanted to continue in that theme to talk about callbacks, before diving deeper into the request params and other internal details.
> 
> Do you still think we should put request params before callbacks?

I guess callbacks are more prominent in controllers than in jobs or mailers (where we mention them later).
For ActiveRecord we mostly mention them after validations and migrations.

I had a look at how we mention controller callbacks in the "Getting Started" guide.
We mention a `before_action` callback to load a model, but this is after we mention params and strong params.
And here it's basically a refactoring to remove some duplication, not a required feature.
For authentication we just use the generator which adds a concern with callbacks defined.

Also, if we are going to do authentication in a callback, we also need the session or cookies to explain `Current.user`.
I think we should probably avoid mentioning CurrentAttributes in the "ActionController Overview" guide.

ayushn21:
> I guess callbacks are more prominent in controllers than in jobs or mailers (where we mention them later). For ActiveRecord we mostly mention them after validations and migrations.

Yeah exactly, I feel they're more important in controllers.

> I had a look at how we mention controller callbacks in the "Getting Started" guide.
> We mention a before_action callback to load a model, but this is after we mention params and strong params.
> And here it's basically a refactoring to remove some duplication, not a required feature.

I don't think every guide needs to approach the problem in the exact same way. I haven't seen that guide but that approach probably made more sense there. In this guide, I feel this flow works because we're discussing structure to start with (CRUD methods etc) and callbacks are in keeping with that, before moving further on to internal details.

> For authentication we just use the generator which adds a concern with callbacks defined.

Even so, it's a concept that the developer needs to be familiar with. Also, I find that the narrative approach for a "follow along" guide where the reader is expected to be writing code while reading, is a bit different from a generic guide that looks to explain concepts. 

> Also, if we are going to do authentication in a callback, we also need the session or cookies to explain Current.user.
> I think we should probably avoid mentioning CurrentAttributes in the "ActionController Overview" guide.

We're not really getting into the weeds of it here. As a line of code, this line is easy enough to understand without additional context: `unless Current.user.present?`. Sure, the reader may or may not be familiar with `ActiveSupport::CurrentAttributes`, but that's not need to illustrate the point of this snippet. The point of this snippet is just to show the user that a callback is where you'd do authentication. The actual authentication technique is not important here.

I don't think there's anything wrong this demonstration as it shows the correct Rails practice. We can't explain every concept in detail, and I like to keep the examples as practical as possible.

I'm reluctant to change it to be honest but if you feel strongly about it, I'll have another think.

guides/source/action_controller_overview.md
ayushn21: Ahhh nice, somehow missed the fact that we specify HTTP as a format!

guides/source/action_controller_overview.md
ayushn21: Ah cool

guides/source/action_controller_overview.md
```diff
@@ -8,157 +8,517 @@ In this guide, you will learn how controllers work and how they fit into the req
...
+  end
 end
 ```
 
-NOTE: All controllers inherit from `ApplicationController`, which in turn
-inherits from
-[`ActionController::Base`](https://api.rubyonrails.org/classes/ActionController/Base.html).
-For [API only](https://guides.rubyonrails.org/api_app.html) applications `ApplicationController` inherits from
-[`ActionController::API`](https://edgeapi.rubyonrails.org/classes/ActionController/API.html).
+Once the [router matches](routing.html) an incoming request a controller and
```
p8: Small typo I noticed:
```suggestion
Once the [router matches](routing.html) an incoming request to a controller and
```

p8:
> Yep, I can write a proposal in Basecamp of you wish?

Sure!

p8:
> I'm reluctant to change it to be honest but if you feel strongly about it, I'll have another think.

I don't want to bikeshed too much on moving the callbacks up and don't have feel _that_ strongly about it.
This is probably too trivial to bother Rails Core with, but we could ask the other writers in Basecamp?

p8:
@ayushn21 I'm good with merging this PR. Looking at the order of callback seems good to me now, looking at it with fresh eyes.
There currently is a conflict (sorry about that), mostly related to the HTTP QUERY method. This new section seems to be randomly appended to the guide. It probably also should be moved to the Routing guide instead.
Can you resolve the conflicts?

ayushn21:
@p8 all done!

I made a small tweak to the wording for the `QUERY` method in the action controller advanced topics guide. I removed the new section on the `QUERY` method in the action controller overview as I don't feel it's necessary. It has already been described in the routing guide, and the section was more about the HTTP method, than how Rails handles it. I don't think we need to have that in the guide as we're not teaching HTTP.

p8:
Pfff, those additions for the QUERY PR mention the RFC 10008 each time QUERY is mentioned 😭

p8:
@ayushn21 The "Layouts and Rendering" mostly describes rendering response methods and layout methods in ActionController. It much less related to ActionView.
What do you think about moving this guide under "Routing and Controllers" (after "Action Controller Overview") instead of under "Views"?

ayushn21:
@p8 since we've changed the order and renamed the sections in the index so it flows better sequentially, I'm inclined to leave this guide where it is.

While your point is valid, this guide requires knowledge about the templating, the different templating options supported, partials etc. Those are all covered in the Action View Overview guide. As such, I think we should leave it where it is.

PR #57156, opened 2026-04-07
Query command for read-only database queries

lewispb:
### Motivation / Background

Agents and automated tooling increasingly need to query production databases to answer questions, investigate issues, and gather context. Today the closest option is `rails runner`, but it has real problems for this use case:

- **Unstructured output.** `rails runner "puts Account.count"` produces plain text. There's no consistent format an agent can parse reliably — you're at the mercy of whatever `inspect` or `puts` produces.
- **No read-only enforcement.** Nothing prevents `rails runner "Account.delete_all"` from executing.
- **No pagination.** Large result sets are dumped in full or require the caller to craft their own LIMIT/OFFSET logic.

With Kamal, running commands against production is straightforward (`kamal app exec`), but you need a command that returns structured, parseable output and enforces read-only access. That's what `rails query` provides.

### Detail

Adds `rails query` — a read-only database query command with structured JSON output.

**ActiveRecord expressions:**

```bash
$ rails query "Account.where(plan: 'premium').limit(2)"
```

```json
{
  "columns": ["id", "name", "plan", "created_at"],
  "rows": [
    [1, "Acme", "premium", "2025-01-15T10:30:00Z"],
    [2, "Widgets Co", "premium", "2025-03-22T14:00:00Z"]
  ],
  "meta": {
    "row_count": 2,
    "query_time_ms": 4.2,
    "page": 1,
    "per_page": 100,
    "has_more": false,
    "sql": "SELECT \"accounts\".* FROM \"accounts\" WHERE \"accounts\".\"plan\" = 'premium' LIMIT 2"
  }
}
```

**Raw SQL:**

```bash
$ rails query --sql "SELECT id, name FROM accounts LIMIT 2"
```

```json
{
  "columns": ["id", "name"],
  "rows": [[1, "Acme"], [2, "Widgets Co"]],
  "meta": {
    "row_count": 2,
    "query_time_ms": 1.8,
    "page": 1,
    "per_page": 100,
    "has_more": false,
    "sql": "SELECT id, name FROM accounts LIMIT 2"
  }
}
```

**Scalar results:**

```bash
$ rails query "Account.count"
```

```json
{
  "columns": ["result"],
  "rows": [[42]],
  "meta": {
    "row_count": 1,
    "query_time_ms": 2.1,
    "page": 1,
    "per_page": 100,
    "has_more": false,
    "sql": "Account.count"
  }
}
```

**Schema listing:**

```bash
$ rails query schema
```

```json
{
  "columns": ["table_name", "row_count"],
  "rows": [["accounts", "42"], ["posts", "1337"], ["users", "108"]],
  "meta": {
    "row_count": 3,
    "query_time_ms": 0,
    "page": 1,
    "per_page": 3,
    "has_more": false,
    "sql": ""
  }
}
```

**Table detail (with associations):**

```bash
$ rails query schema accounts
```

```json
{
  "table": "accounts",
  "columns": [
    {"name": "id", "type": "bigint", "null": false, "default": null},
    {"name": "name", "type": "varchar(255)", "null": false, "default": null},
    {"name": "plan", "type": "integer", "null": false, "default": "0"}
  ],
  "indexes": [
    {"name": "index_accounts_on_name", "columns": ["name"], "unique": true}
  ],
  "enums": {
    "plan": {"free": 0, "premium": 1, "enterprise": 2}
  },
  "associations": [
    {"type": "has_many", "name": "users", "class_name": "User", "foreign_key": "account_id"},
    {"type": "has_many", "name": "posts", "class_name": "Post", "foreign_key": "account_id"}
  ]
}
```

**Models (with associations):**

```bash
$ rails query models
```

```json
[
  {
    "model": "Account",
    "table_name": "accounts",
    "associations": [
      {"type": "has_many", "name": "users", "class_name": "User", "foreign_key": "account_id"}
    ]
  },
  {
    "model": "User",
    "table_name": "users",
    "associations": [
      {"type": "belongs_to", "name": "account", "class_name": "Account", "foreign_key": "account_id"}
    ]
  }
]
```

**Errors are always structured JSON:**

```bash
$ rails query --sql "INSERT INTO accounts (name) VALUES ('test')"
```

```json
{
  "error": "Write query attempted while in readonly mode",
  "meta": {"query_time_ms": 0}
}
```

**Subcommands:**
- `rails query schema [TABLE]` — list tables or show table detail with columns, indexes, enums, and associations
- `rails query models` — list ActiveRecord models with table names and associations
- `rails query explain EXPRESSION` — show EXPLAIN output

**Safety:** Connects to the reading replica role by default. Falls back to `while_preventing_writes` for single-database setups.

**Pagination:** `--page` and `--per` options with automatic LIMIT/OFFSET injection for SQL queries.

**Database selection:** `--database` / `--db` option for multi-database apps.

**Instrumentation:** Emits an `"query.rails"` ActiveSupport notification wrapping every query execution. This is a hook point for audit logging — applications that need to track who ran what against production (e.g. via [Console1984](https://github.com/basecamp/console1984)) can subscribe to this event without patching the command. The payload includes the expression that was evaluated.

### Checklist

### DIFF
--- a/railties/lib/rails/commands/query/query_command.rb
+++ b/railties/lib/rails/commands/query/query_command.rb
@@ -1,0 +1,271 @@
+# frozen_string_literal: true
+
+require "json"
+require "rails/command/environment_argument"
+
+module Rails
+  module Command
+    class QueryCommand < Base # :nodoc:
+      include EnvironmentArgument
+
+      class_option :sql, type: :boolean,
+        desc: "Treat input as raw SQL instead of an ActiveRecord expression"
+
+      class_option :database, aliases: "--db", type: :string,
+        desc: "Database configuration to use (e.g. primary_replica)"
+
+      class_option :page, type: :numeric, default: 1,
+        desc: "Page number for paginated results"
+
+      class_option :per, type: :numeric, default: 100,
+        desc: "Results per page (max 10000)"
+
+      desc "query [EXPRESSION]", "Run a read-only query against the database"
+      def perform(expression = nil, *args)
+        boot_application!
+        Rails.application.load_runner
+
+        ActiveSupport::Notifications.instrument("query.rails", expression: expression) do
+          case expression
+          when "schema"
+            run_schema(args.first)
+          when "models"
+            run_models
+          when "explain"
+            run_explain(args.first)
+          else
+            run_query(expression)
+          end
+        end
+      rescue StandardError, SyntaxError, NotImplementedError => e
+        output_error(e.message)
+        exit 1
+      end
+
+      private
+        def run_query(expression)
+          expression = resolve_expression(expression)
+          page = [ options[:page], 1 ].max
+          per = [ [ options[:per], 1 ].max, 10_000 ].min
+
+          start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          result = if options[:sql]
+            with_readonly_connection do |connection|
+              execute_sql(connection: connection, sql: expression, page: page, per: per)
+            end
+          else
+            execute_ar(expression: expression, page: page, per: per)
+          end
+          elapsed_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time) * 1000).round(1)
+
+          say format_result(**result, elapsed_ms: elapsed_ms, page: page, per: per)
+        end
+
+        def run_schema(table = nil)
+          with_readonly_connection do |connection|
+            if table
+              say format_table_detail(connection, table)
+            else
+              say format_table_list(connection)
+            end
+          end
+        end
+
+        def run_models
+          Rails.application.eager_load!
+
+          models = ActiveRecord::Base.descendants
+            .reject(&:abstract_class?)
+            .select { |model| model.table_name.present? }
+            .sort_by(&:name)
+
+          data = models.map do |model|
+            {
+              model: model.name,
+              table_name: model.table_name,
+              associations: format_associations(model)
+            }
+          end
+
+          say JSON.generate(data)
+        end
+
+        def run_explain(expression = nil)
+          expression = resolve_expression(expression)
+
+          if options[:sql]
+            with_readonly_connection do |connection|
+              result = connection.select_all("EXPLAIN #{expression}")
+              say format_result(columns: result.columns, rows: result.rows, sql: "EXPLAIN #{expression}")
+            end
+          else
+            relation = eval(expression, TOPLEVEL_BINDING, "(query)", 1)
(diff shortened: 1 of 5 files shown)


railties/lib/rails/commands/query/query_command.rb
```diff
@@ -0,0 +1,233 @@
...
+            { columns: [ "result" ], rows: [ [ result ] ], sql: expression, truncated: false }
+          end
+        end
+
+        def execute_sql(connection, sql, page, per)
+          unless sql.gsub(/--.*$|\/\*.*?\*\//m, "").match?(/\bLIMIT\b/i)
+            offset = (page - 1) * per
+            sql = "#{sql.rstrip.chomp(';')} LIMIT #{per + 1}"
+            sql += " OFFSET #{offset}" if offset > 0
+          end
```
lewispb, line 181: Slow for large datasets, but fine to start with. Could add cursor pagination if desired.

railties/lib/rails/commands/query/query_command.rb
```diff
@@ -0,0 +1,233 @@
...
+
+          JSON.generate({
+            table: table,
+            columns: columns.map do |col|
+              { name: col.name, type: col.sql_type, null: col.null, default: col.default }
+            end,
+            indexes: indexes.map do |idx|
+              { name: idx.name, columns: idx.columns, unique: idx.unique }
+            end,
+            enums: enums.presence
```
lewispb: Could add associations although agents can figure it out already.

lewispb: Added

guilleiguaran:
Just retried the broken build on Buildkite to get this green

andyjeffries:
I don't know if this is an intentional gap, but doing this with a single record doesn't return a useful output?

```
$ rails query "User.find(1)"
{"columns":["result"],"rows":[["#<User:0x00007fb9de746518>"]],"meta":{"row_count":1,"query_time_ms":38.5,"page":1,"per_page":100,"has_more":false,"sql":"User.find(1)"}}
```

The fix appears to be in `execute_ar` add another case of:

```
when ActiveRecord::Base
    attrs = result.attributes
    { columns: attrs.keys, rows: [attrs.values], sql: expression, truncated: false }
```

Or something similar. Feels like LLMs will often try to get a singular record, if it's being talked about?

andyjeffries:
And apparently (so my LLM just told me) the `Array` branch is open to the same issue (although I haven't needed that yet).

```
$ rails query "User.first(2)"
```

Will return those short inspect values.

Prana111025 (commented):
Thanks

PR #54083, opened 2024-12-30
Remove deprecated rails-ujs included in the output of assets:precompile

taketo1113:
### Motivation / Background

When deploying my Rails app (v8.0.1), I noticed that the Propshaft manifest file and the `public/assets/` directory contained `rails-ujs`, which was supposed to have been removed in Rails 7.2, even though my app does not use `rails-ujs`.

- https://github.com/rails/rails/pull/50535

This `rails-ujs` script was output from within the actionview gem during the execution of `rails assets:precompile`.

- https://github.com/rails/rails/blob/v8.0.1/actionview/app/assets/javascripts/rails-ujs.js

Since `rails-ujs` is still present, there is a risk of accidentally referencing it, which could cause conflicts with Turbo or other scripts that provide equivalent functionality. Therefore, it would be better to remove the lingering `rails-ujs`.

### Steps to reproduce

After running `rails new`, executing `rails assets:precompile` as shown below outputs the `rails-ujs` script in the `public/assets/` directory (`public/assets/rails-ujs-20eaf715.js` and `rails-ujs.esm-e925103b.js`).
The Propshaft manifest file (`public/assets/.manifest.json`) also includes entries related to `rails-ujs`.

**(1) Check Rails and Ruby versions**
```
$ rails -v
Rails 8.0.1
$ ruby -v
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin24]
```

**(2) Create a new Rails app and precompile assets**
```
$ rails new rails-app
$ cd rails-app
$ bin/rails assets:precompile
```

**(3) Observe the output files**
```
$ ls -1 public/assets/
action_cable-5212cfee.js
...
application-8b441ae0.css

rails-ujs-20eaf715.js
rails-ujs.esm-e925103b.js

stimulus-autoloader-9d447422.js
...
turbo.min-fae85750.js
```

In this sample Rails app, the `Rails.application.config.assets.paths` output includes actionview's `app/assets/javascripts` directory:
```ruby
$ bin/rails console
Loading development environment (Rails 8.0.1)
rails-app(dev)> Rails.application.config.assets.paths
=> 
["/path/rails-app/app/assets/builds",
 "/path/rails-app/app/assets/images",
 "/path/rails-app/vendor/bundle/ruby/3.3.0/gems/actioncable-8.0.1/app/assets/javascripts",
 "/path/rails-app/vendor/bundle/ruby/3.3.0/gems/actiontext-8.0.1/app/assets/javascripts",
[...]

PR #57157, opened 2026-04-07
Get rid of rails-ujs.js

byroot:
This should have been deleted a few versions ago.


taketo1113:
@byroot The following files also appear to be unnecessary and could be removed:
- `actionview/app/assets/javascripts/.gitattributes`

Additionally, since it is no longer needed, you can remove `"app/assets/javascripts/*.js"` from the files entry in `actionview/actionview.gemspec`.

For more details on these changes, please refer to the following Pull Request:
- #54083

PR #57161, opened 2026-04-08
Raise ArgumentError when content_security_policy_nonce_auto is misconfigured

drsharp:
### Motivation / Background

Fixes #56480

When `config.content_security_policy_nonce_auto` is set to `true` without also configuring `content_security_policy_nonce_directives`, Rails crashes during initialization with `undefined method 'intersect?' for nil (NoMethodError)`. The default value of `content_security_policy_nonce_directives` is `nil`, and the auto-include assignments in ActionView's `after_initialize` block call `.intersect?` on it unconditionally.

The same problem exists if `content_security_policy_nonce_generator` is omitted — `nonce_auto` requires both to be configured, but neither absence is caught cleanly.

### Detail

This Pull Request changes `actionview/lib/action_view/railtie.rb` to add a validation block at the start of the `after_initialize` hook. When `content_security_policy_nonce_auto` is `true`, it now raises `ArgumentError` with a clear message if either `content_security_policy_nonce_directives` or `content_security_policy_nonce_generator` is `nil`, pointing the developer to `config/initializers/content_security_policy.rb`.

Two tests are added to `railties/test/application/content_security_policy_test.rb` covering each missing-config case.

### Additional information

Both new tests are intentionally minimal — no controller or route setup is needed since the error fires during `after_initialize` before any request is made.

Verified by running the full CSP integration test suite:
Screenshot 2026-04-08 at 12 48 55 PM

"confirm failure" test by running the CSP tests with the changes temporarily commented out:
Screenshot 2026-04-08 at 12 50 10 PM

### Checklist

☝️ intentionally skipping CHANGELOG as I don't believe it's necessary here (although I'm happy to add it if needed

### DIFF
--- a/actionview/lib/action_view/railtie.rb
+++ b/actionview/lib/action_view/railtie.rb
@@ -74,6 +74,20 @@ module ActionView
     end
 
     config.after_initialize do |app|
+      if app.config.content_security_policy_nonce_auto
+        if app.config.content_security_policy_nonce_directives.nil?
+          raise ArgumentError, "Set config.content_security_policy_nonce_directives when " \
+            "using config.content_security_policy_nonce_auto. " \
+            "See config/initializers/content_security_policy.rb for more information."
+        end
+
+        if app.config.content_security_policy_nonce_generator.nil?
+          raise ArgumentError, "Set config.content_security_policy_nonce_generator when " \
+            "using config.content_security_policy_nonce_auto. " \
+            "See config/initializers/content_security_policy.rb for more information."
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


drsharp:
Screenshot 2026-04-08 at 1 31 12 PM

I believe these are unrelated failures. Most likely CI flakiness. 😦

Saidbek:
Isn't it a duplicate of https://github.com/rails/rails/pull/56483 which was already tagged in the original issue?

drsharp:
> Isn't it a duplicate of #56483 which was already tagged in the original issue?

🤦  Huh. I'm not sure how I missed that. I saw that there was a PR attached but it was getting old and I thought it was missing some parts needed to resolve this. We both had the same ideas around what should be done to solve it, with slightly different messaging.

I'm fine to have this PR closed. I didn't mean to duplicate work.

Thanks for pointing it out to me.

PR #57163, opened 2026-04-08
Document that `value_method` and `text_method` arguments of `options_from_collection_for_select` accept Proc [ci skip]

cgunther:
These arguments, and their corresponding versions on `option_groups_from_collection_for_select` along with `group_method` and `group_label_method`, accept a `Proc` in addition to a string/symbol method name to determine what value to return, however this support wasn't documented.


cgunther:
Thanks for the merge!

PR #57166, opened 2026-04-09
Use hash lookup for exact-match regexp filters in ParameterFilter

alexcwatt:
### Motivation / Background

I was looking at one of our applications at Shopify where a number of the ParameterFilter regexes are just exact matches -- e.g., `/^email$/`.

We can optimize ParameterFilter for cases like this and get a meaningful speedup in ParameterFilter. And because we call ParameterFilter so often, it can be meaningful.

### Detail

When ParameterFilter is initialized with anchored regexp filters like `/^code$/` or `/\Atoken\z/`, extract the literal string and store it in a Hash for O(1) lookup instead of iterating all regexps with `.any?`.

This is a common pattern in Rails apps that want to filter specific parameter names without matching substrings. I've included a benchmark script in the commit; if all filters are exact matches, it's 4.5x faster to use the hash to check for which parameters to exclude.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/parameter_filter.rb
+++ b/activesupport/lib/active_support/parameter_filter.rb
@@ -96,6 +96,18 @@ module ActiveSupport
     end
 
   private
+    # If the regexp is an anchored exact match like /^token$/ or /\Atoken\z/,
+    # returns the literal string.
+    def extract_exact_key(regexp) # :nodoc:
+      return if regexp.casefold?
+      source = regexp.source
+      return unless source.start_with?("^", "\\A") && source.end_with?("$", "\\z")
+
+      literal = source.delete_prefix("^").delete_prefix("\\A")
+                      .delete_suffix("$").delete_suffix("\\z")
+      literal if literal.match?(/\A[a-zA-Z0-9_]+\z/)
+    end
+
     def compile_filters!(filters)
       @no_filters = filters.empty?
       return if @no_filters
@@ -103,6 +115,7 @@ module ActiveSupport
       @regexps, strings = [], []
       @deep_regexps, deep_strings = nil, nil
       @blocks = nil
+      @exact_keys = nil
 
       filters.each do |item|
         case item
@@ -111,6 +124,8 @@ module ActiveSupport
         when Regexp
           if item.to_s.include?("\\.")
             (@deep_regexps ||= []) << item
+          elsif (literal = extract_exact_key(item))
+            (@exact_keys ||= {})[literal] = true
           else
             @regexps << item
           end
@@ -139,11 +154,15 @@ module ActiveSupport
     end
 
     def value_for_key(key, value, full_parent_key = nil, original_params = nil)
+      key_s = key.to_s
+
       if @deep_regexps
-        full_key = full_parent_key ? "#{full_parent_key}.#{key}" : key.to_s
+        full_key = full_parent_key ? "#{full_parent_key}.#{key_s}" : key_s
       end
 
-      if @regexps.any? { |r| r.match?(key.to_s) }
+      if @exact_keys && @exact_keys[key_s]
+        value = @mask
+      elsif @regexps.any? { |r| r.match?(key_s) }
         value = @mask
       elsif @deep_regexps&.any? { |r| r.match?(full_key) }
         value = @mask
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/parameter_filter.rb
```diff
@@ -96,13 +96,27 @@ def filter_param(key, value)
...
+    # returns the literal string. Returns nil for case-insensitive patterns,
+    # partial matches, or patterns with metacharacters.
+    def extract_exact_key(regexp) # :nodoc:
+      return if regexp.casefold?
+      source = regexp.source
+      return unless source.start_with?("^", "\\A") && source.end_with?("$", "\\z")
+
+      literal = source.delete_prefix("^").delete_prefix("\\A")
+                      .delete_suffix("$").delete_suffix("\\z")
+      literal unless literal.match?(/[.\[\]\\()|*+?{}]/)
```
byroot: Parsing regexps with regexps makes me uncorfortable. It's very hard to be sure we got this perfectly right.

At the very least, this should error on the side of caution, and only return the literal if we match a safe regexp, not skip returning it unless we match one.

So something like:

```ruby
literal if literal.match?(/\A[a-zA-Z0-9_]\z/)
```

alexcwatt: Good feedback. I made that exact change.

eregon:
What about the performance for apps which don't only do exact matches?
The default seems:
https://github.com/rails/rails/blob/74bb87c5888b3489fc519f048e6a0cd9cc615a5d/guides/source/configuring.md#configfilter_parameters
```ruby
Rails.application.config.filter_parameters += [
  :passw, :email, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn, :cvv, :cvc
]
```
which is not exact matches.

I would also expect this optimization to be counter-productive if there is a mix of exact and substring matches, as `precompile_filter_parameters` would fold everything to a single Regexp, which should match the same or faster than a Hash lookup + a Regexp with the rest

My concern specifically is the added `extract_exact_string_key` and `extract_exact_line_key` will run for every request.
Context is I was looking at this code recently: https://github.com/rails/rails/pull/58680#issuecomment-5574751670

eregon:
I let Claude (Fable) check my concerns and this is its summary:

---

I benchmarked this to quantify my concern above. Setup: CRuby 4.0.4, benchmark-ips, a realistic 17-key nested params hash, comparing the parent of this PR's merge vs current main. Both sides include dd1728eb4e (`to_s`→`source`), which is an independent optimization, so the only variable is this PR (+ follow-up 6f466a833f).

**1. Per-request cost is real: `ParameterFilter.new(precompiled)` got ~1.5x slower.**
`ActionDispatch` builds a `ParameterFilter` per request from the precompiled filters ([filter_parameters.rb#L70](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/http/filter_parameters.rb#L70)):

| | before | after |
|---|---|---|
| precompiled defaults | 809 ns | 1.25 μs (**1.54x slower**) |
| precompiled defaults + 6 exact regexps | 871 ns | 1.35 μs (**1.67x slower**) |
| raw (non-precompiled) mix | 18.96 μs | 25.96 μs (**1.37x slower**) |

**2. And the optimization never activates on this default path.** `precompile_filters` folds exact-match regexps into the union, whose source starts with `(`, so `extract_exact_*` always fails and the hash is never built. With `precompile_filter_parameters` on (the default since 7.1), request filtering pays the ~0.45 μs/request extraction cost and gets nothing back. The hash only activates where raw filters are compiled directly (`ActiveRecord#inspection_filter`, `EventReporter`, precompile disabled).

**3. Where it does activate, it works — including in the mixed case.** With 12 default substring filters + 6 exact regexps, filtering the params hash:

| | i/s |
|---|---|
| before, no precompile (7-regexp `any?` loop) | 28.5k |
| precompiled single 18-branch union | 41.9k |
| after, no precompile (hash + 12-branch regexp) | **48.6k** |

So contrary to my expectation, hash + smaller regexp beat the single folded union by ~16% here (exact hits skip the regexp entirely; misses scan a smaller union). Defaults-only filtering (no exact matches) is unchanged, within error.

Given 2 and 3, the way to actually get this win on the request path would be for `precompile_filters` to keep exact-match regexps out of the union so the hash survives precompilation — that would replace the current ~0.45 μs/request pure loss with the ~16% filtering gain.

PR #57167, opened 2026-04-09
Fix ActiveStorage::IntegrityError when mirroring variants without a checksum

### LINKED ISSUE #57164: ActiveStorage variant mirroring fails when not using track_variants
### Steps to reproduce

Assuming an image processing lib like [vips](https://www.libvips.org/install.html) is installed:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "8.1.3"

  gem "sqlite3"
  gem "image_processing"
end

require "minitest/autorun"

  config.root = __dir__
  config.hosts << "example.org"
  config.eager_load = false

  Rails.logger  = config.logger

  config.active_storage.service = :mirror
    primary: {
      root: Dir.mktmpdir,
      service: "Disk"
    },
    replica: {
      root: Dir.mktmpdir,
      service: "Disk"
    },
    mirror: {
      service: "Mirror",
      primary: :primary,
      mirrors: [:replica]
    }    
  }

  config.active_storage.track_variants = false
end
Rails.application.initialize!

end

  has_one_attached :profile
end

  def test_upload_and_download
    user = User.create!(
      profile: {
        content_type: "image/gif",
        filename: "dummy.gif",
        io: ::StringIO.new(Base64.decode64("R0lGODlhAQABAIABAP///wAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==")),
      }
    )

    user.profile.variant(resize_to_limit: [1, 1]).processed

    assert ActiveStorage::Blob.service.mirrors.first.exist?(user.profile.variant(resize_to_limit: [1, 1]).blob.key)
  end
end

```

### Expected behavior

When processing a variant in these conditions:
- `config.active_storage.track_variants = false`
- `config.active_storage.service = :mirror`

I would expect ActiveStorage to upload the file to the primary service as well as the secondary one.

### Actual behavior

`ActiveStorage::MirrorJob` fails with `ActiveStorage::IntegrityError`.

Explanation:
1. Calling `ActiveStorage::Variant#processed` calls `ActiveStorage::Service::MirrorService#upload` without a checksum ([here](https://github.com/rails/rails/blob/v8.1.3/activestorage/app/models/active_storage/variant.rb#L114)).
2. `ActiveStorage::Service::MirrorService#upload` does two things ([here](https://github.com/rails/rails/blob/v8.1.3/activestorage/lib/active_storage/service/mirror_service.rb#L45)):
  2.1.  It calls `ActiveStorage::Service::DiskService#upload`, which skips verification since the checkum is `nil`.
  2.2. Then enqueues `ActiveStorage::MirrorJob` with a checksum of `nil`.
3. `ActiveStorage::MirrorJob` calls `ActiveStorage::Service::MirrorService#mirror` with a checksum of `nil`, which triggers the `ActiveStorage::IntegrityError` ([here](https://github.com/rails/rails/blob/v8.1.3/activestorage/lib/active_storage/service/mirror_service.rb#L69)).

This also contradicts the behavior explained in the docs because it's enforcing a match on the mirror service despite a checksum not being provided:

> If a +checksum+ is provided, all services will ensure a match when the upload has completed or raise an ActiveStorage::IntegrityError.

Note: This bug doesn't happen when `track_variants = true`.

A possible solution could be done in [ActiveStorage::Service::MirrorService#mirror](https://github.com/rails/rails/blob/v8.1.3/activestorage/lib/active_storage/service/mirror_service.rb#L69):
```diff
- primary.open(key, checksum: checksum) do |io|
+ primary.open(key, checksum: checksum, verify: checksum.present?) do |io|
```

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 3.3.9

Saidbek:
### Motivation / Background

Fixes #57164

This PR has been created because `ActiveStorage::MirrorJob` raises an `ActiveStorage::IntegrityError` when mirroring variants uploaded without a checksum - which happens whenever `track_variants = false`.

### Details

When `config.active_storage.track_variants = false`, variants are uploaded without a checksum, which means `MirrorJob` ends up calling `MirrorService#mirror` with `checksum: nil`. That `nil` eventually reaches `download_and_verify_tempfile`, which tries to verify integrity regardless - comparing the computed digest against `nil` always fails, so `IntegrityError` is raised every time. The fix is a one-line guard that mirrors what `DiskService#upload` already does.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/service.rb
+++ b/activestorage/lib/active_storage/service.rb
@@ -211,7 +211,7 @@ module ActiveStorage
         tempfile.flush
         tempfile.rewind
 
-        verify_integrity_of(tempfile, checksum: checksum) if verify
+        verify_integrity_of(tempfile, checksum: checksum) if verify && checksum
         if block_given?
           begin
             yield tempfile
--- a/activestorage/test/service/mirror_service_test.rb
+++ b/activestorage/test/service/mirror_service_test.rb
@@ -79,6 +79,44 @@ class ActiveStorage::Service::MirrorServiceTest < ActiveSupport::TestCase
     assert_equal "Surprise!", @service.mirrors.third.download(key)
   end
 
+  test "mirroring a file without a checksum does not raise IntegrityError" do
+    key  = SecureRandom.base58(24)
+    data = "Something else entirely!"
+
+    @service.primary.upload key, StringIO.new(data)
+
+    assert_nothing_raised do
+      @service.mirror key, checksum: nil
+    end
+
+    @service.mirrors.each do |mirror|
+      assert_equal data, mirror.download(key)
+    end
+  ensure
+    @service.delete key
+  end
+
+  test "uploading without a checksum mirrors to all services" do
+    old_service = ActiveStorage::Blob.service
+    ActiveStorage::Blob.service = @service
+
+    key  = SecureRandom.base58(24)
+    data = "Something else entirely!"
+    io   = StringIO.new(data)
+
+    assert_performed_jobs 1, only: ActiveStorage::MirrorJob do
+      @service.upload key, io.tap(&:read)
+    end
+
+    assert_equal data, @service.primary.download(key)
+    @service.mirrors.each do |mirror|
+      assert_equal data, mirror.download(key)
+    end
+  ensure
+    @service.delete key
+    ActiveStorage::Blob.service = old_service
+  end
+
   test "URL generation in primary service" do
     filename = ActiveStorage::Filename.new("test.txt")
 


activestorage/lib/active_storage/service.rb
```diff
@@ -211,7 +211,7 @@ def download_and_verify_tempfile(key, checksum: nil, verify: true, name: "Active
         tempfile.flush
         tempfile.rewind
 
-        verify_integrity_of(tempfile, checksum: checksum) if verify
+        verify_integrity_of(tempfile, checksum: checksum) if verify && checksum
```
mrpasquini, line 214: Shouldn't the mirror service service be setting `verify` param to `false` rather than disabling verification if checksum is nil across all services?
I think I prefer #57170 approach

rafaelfranca:
Duplicate of #57170

I'm going with that PR. Thank you

Issue #53332, opened 2024-10-16
ActiveRecord::Store I can't search in the db


leoluchini:
### Steps to reproduce

```ruby
class CreateExclusions < ActiveRecord::Migration[7.2]
  def change
    create_table :exclusions do |t|
      t.jsonb :device
      
      t.timestamps
    end
  end
end


```

Using `store` from `ActiveRecod::Store`

```ruby
class Exclusion < ApplicationRecord
  store :device, accessors: %i(brand brand_model), coder: JSON
end

Exclusion.create(brand: 'AAA')

Exclusion.where("device ->> 'brand' = 'AAA'")
=> []
```
When I checked the database I found the data is stored in this way

```
 id |         device         |         created_at         |         updated_at         
----+------------------------+----------+------------------+---------------------------
 12 | "{\"brand\":\"AAAS\"}" | 2024-10-16 09:42:41.739296 | 2024-10-16 09:42:41.739296
```
When I remove all the logic from `ActiveRecod::Store`

```ruby
class Exclusion < ApplicationRecord
end

Exclusion.create(device: {brand: 'BBB'})

> Exclusion.where("device ->> 'brand' = 'BBB'")
=> [#<Exclusion:0x0000000151f5f558 id: 13, device: {"brand"=>"BBB"}>]

```
The current status in the database is

```
 id |         device         |         created_at         |         updated_at         
----+------------------------+----------+------------------+----------------------------
 12 | "{\"brand\":\"AAAS\"}" | 2024-10-16 09:42:41.739296 | 2024-10-16 09:42:41.739296
 13 | {"brand": "BBB"}       | 2024-10-16 10:54:01.420978 | 2024-10-16 10:54:01.420978
(2 rows)
```

### Expected behavior
be able to perform searches with the information stored through `ActiveRecord::Store`

### Actual behavior
does not return results

### System configuration
**Rails version**: 7.2.1

**Ruby version**: 3.3.5

cubasepp:
https://api.rubyonrails.org/v7.2/classes/ActiveRecord/Store.html

[...]

Issue #49432, opened 2023-09-29
Support for storing ActiveStorage::Blob metadata in a jsonb column type for PostgreSQL


vpereira:
I've been exploring the use of ActiveStorage with PostgreSQL and noticed that the metadata column for ActiveStorage::Blob is stored as text. Given the benefits of the jsonb data type in PostgreSQL (like indexing capabilities, and advanced querying), I'm wondering  if would worth to migrate this column to use jsonb.

So based on that I would like to:

* Understand if there are any specific reasons or considerations why the default type for metadata is text (I can imagine because of other database technologies, but just to be sure).

* Explore the possibility of Rails supporting or recommending jsonb for this purpose, especially since PostgreSQL is a popular choice among Rails developers.

* Suggest that if there are benefits to using jsonb for metadata, perhaps an option could be added to the ActiveStorage generators to allow developers to choose the column type when setting up ActiveStorage.

rafaelfranca:
Thank you so much for the issue but we reserve the issues tracker for issues only.. We recommend you to try to implement the feature and send us a pull request or if you are in doubt if the feature will be accepts ask in the [Rails Core discussion board](https://discuss.rubyonrails.org/c/rubyonrails-core/5) for feedback.

vpereira:
Hi @rafaelfranca,

Thank you for your response. I'm eager to submit a PR addressing this. When adjusting the ActiveStorage migration to use jsonb, should it automatically detect the PostgreSQL driver, or would you prefer this to be controlled via a configuration or rake task flag?

rafaelfranca:
I think it should be automatic.

Issue #48461, opened 2023-06-13
ActiveRecord::Store 7 unexpacted behaviour


dixent:
### Description

After updating Rails 6 to Rails 7 I found failed specs that include store defining in models. I have made research on this problem and looks like the roots came from sources. Looks like `stores` in the 7 version behave unstably. Please have a look:

Test repo with existing specs: https://github.com/dixent/activerecord_7_store

### Steps to reproduce
1. Create a model with field :jsonb, null: false, default: {}
```ruby
create_table :applicants do |t|
      t.jsonb :settings, null: false, default: {}
      t.string :name
      t.timestamps
    end
```
2. Define store for `settings`
```ruby
# app/models/applicant.rb
class Applicant < ApplicationRecord
  store :settings, accessors: [ :age, :phone_number ], coder: JSON
end
```

3. Then two existing problems, all these spec fail, but must pass. If you remove `store :settings, accessors: [ :age, :phone_number ], coder: JSON` from the model all tests pass successfully.
```ruby
# spec/models/applicant_spec.rb
require 'rails_helper'

describe Applicant do
  subject(:new_applicant) { described_class.new }

  it 'must except record without settings' do
    expect(Applicant.create.id).to_not be(nil)
  end

  describe '#order' do
    let!(:low) { Applicant.create(settings: { phone_number: '111' }) }
    let!(:high) { Applicant.create(settings: { phone_number: '222' }) }

    it 'must work order by phone_number DESC' do
      expect(Applicant.order(Arel.sql("settings -> 'phone_number' DESC")).ids).to eq([high.id, low.id])
    end

    it 'must work order by phone_number ASC' do
      expect(Applicant.order(Arel.sql("settings -> 'phone_number' ASC")).ids).to eq([low.id, high.id])
    end
  end
end
```

### System configuration
**Rails version**: 7.0.5

**Ruby version**: 3.2.2

fatkodima:
Can you recreate a test script using this template -  https://github.com/rails/rails/blob/main/guides/bug_report_templates/active_record_main.rb?

dixent:
> Can you recreate a test script using this template - https://github.com/rails/rails/blob/main/guides/bug_report_templates/active_record_main.rb?

I hope it helps.
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

[...]

PR #57168, opened 2026-04-09
Fix `store` on native json/jsonb columns to skip serialization

### LINKED ISSUE #52256: JSON attribute handling inconsistency in ActiveRecord models
There appears to be an inconsistency in how JSON attributes are handled in ActiveRecord models when using the `store` method. 

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"

    t.json :settings, default: {}
  end

  create_table :store_users, force: true do |t|
    t.json :settings, default: {}
  end

  create_table :serialize_users, force: true do |t|
    t.json :settings, default: {}
  end
end

end

class StoreUser < ActiveRecord::Base
  store :settings, accessors: %i[color], coder: JSON
end

class SerializeUser < ActiveRecord::Base
  serialize :settings, coder: JSON
end

  def setup
    @user = User.create!(settings: { color: "red" })
    @store_user = StoreUser.create!(color: "red")
  end

  def test_search_by_json_column
    # This assertion passes
    assert_equal User.find_by("id = ? AND settings ->> 'color' = ?", @user.id, "red"), @user
    # This assertion fails
    assert_equal StoreUser.find_by("id = ? AND settings ->> 'color' = ?", @store_user.id, "red"), @store_user
  end

  def test_value_before_type_cast
    user_value_before_type_cast = @user.read_attribute_before_type_cast("settings")
    # => "{\"color\":\"red\"}"
    store_user_value_before_type_cast = @store_user.read_attribute_before_type_cast("settings")
    # => "\"{\\\"color\\\":\\\"red\\\"}\""

    # This assertion fails
    assert_equal user_value_before_type_cast, store_user_value_before_type_cast
  end

  def test_both_cast_type_and_coder_are_json
    assert_raises(ActiveRecord::AttributeMethods::Serialization::ColumnNotSerializableError) do
      SerializeUser.create!(settings: { color: "red" })
    end
  end
end
```

### Expected behavior

The attributes should be correctly serialized as JSON in the database when using `store` method.

### Actual behavior

The attribute in the `store` method appears to be serialized twice when the database type and the coder type are both set to JSON. This is resulting in an incorrect JSON format in the database.

### System configuration
**Rails version**: main

**Ruby version**: 3.3.0

markedmondson:
## Summary

`store :metadata, coder: JSON` on a native `json` or `jsonb` column currently:
- **Raises on read**: `"no implicit conversion of Hash into String"` — the adapter returns a Hash, but `IndifferentCoder` passes it to `JSON.decode` which expects a String
- **Double-serializes on write**: `IndifferentCoder.dump` produces a JSON string, then `Type::Serialized` passes it to `Type::Json#serialize` which encodes it again

This has been reported multiple times (#52256, #49432, #48461, #53332) and the current guidance is "use `store_accessor` instead" — but `store_accessor` on json columns uses `StringKeyedHashAccessor` which loses symbol key access, a breaking change for any code doing `record.metadata[:key]`.

### Root cause

`store` unconditionally wraps the column in `Type::Serialized` via `IndifferentCoder`. On text columns this is correct — the coder converts between Ruby Hashes and JSON/YAML strings. On json/jsonb columns the adapter already handles serialization, so the extra layer double-serializes writes and breaks reads.

### Fix

`store` now detects native json/jsonb columns (via `type_for_attribute`) and uses a new `Type::IndifferentJson` type instead of `Type::Serialized`. `IndifferentJson` inherits from `Type::Json` and:
- Wraps `deserialize` output in `HashWithIndifferentAccess`
- Returns `IndifferentHashAccessor` from `accessor`

This provides the same symbol/string key access that `store` provides on text columns, without the serialization layer.

The `coder` option is safely ignored on native JSON columns since the adapter handles serialization. On text columns, behavior is completely unchanged.

### ActiveStorage

Also changes the `create_active_storage_tables` migration template from `t.text` to `t.json` for the `metadata` column. `t.json` maps to `jsonb` on PostgreSQL and `json` on SQLite/MySQL. Existing applications are unaffected — `store` works correctly on both text and JSON columns after this change.

### Checklist

- [x] Tests for `store` on a native json column (indifferent access, round-trip, dirty tracking, type check)
- [x] CHANGELOG entries for activerecord and activestorage
- [x] Backward compatible — text column behavior unchanged, coder preserved for existing apps

Fixes #52256, #49432, #48461


activerecord/lib/active_record/store.rb
```diff
@@ -110,8 +110,16 @@ def inherited(subclass) # :nodoc
       end
 
       def store(store_attribute, options = {})
-        coder = build_column_serializer(store_attribute, options[:coder], Object, options[:yaml])
-        serialize store_attribute, coder: IndifferentCoder.new(store_attribute, coder)
+        if native_json_column?(store_attribute)
```
byroot: This isn't possible. Declaration time methods like `store` MUST NOT access the schema, otherwise it creates a hard dependency on the database being up to boot the app, which is a big no no.

If you want to to this, you need to find a way to delay this check to after the schema has been loaded.

markedmondson: Fixed, sorry for the messy force-push history getting there.

Now uses `decorate_attributes` (the same lazy mechanism `serialize` itself uses) so the JSON type check happens inside the decorator block after the schema loads, not at declaration time.

activerecord/lib/active_record/type/indifferent_json.rb
```diff
@@ -0,0 +1,27 @@
...
+    # the double-serialization bug that occurs when Type::Serialized wraps a
+    # json/jsonb column.
+    class IndifferentJson < Json
+      def deserialize(value)
+        result = super
+        if result.is_a?(Hash)
+          result.with_indifferent_access
+        else
+          result
+        end
```
byroot: While I understand the intent, I don't think this is suitable, because it only works if the tope level object is a hash:

```ruby
>> {"foo" => [{"bar" => 1}]}.with_indifferent_access.dig(:foo, 0, :bar)
=> 1
```

But put the same document inside an array, and now you get regular hashes.

byroot: I guess it's OK for `store` specifically because we enforce the object to be a Hash, but then we shouldn't name it like some generic type, it probably should be clearly exclusive to AR::Store.

activerecord/lib/active_record/store.rb
```diff
@@ -110,8 +110,22 @@ def inherited(subclass) # :nodoc
...
+        coder_option = options[:coder]
+        yaml_option  = options[:yaml]
+
+        decorate_attributes([store_attribute]) do |attr_name, cast_type|
+          if cast_type.is_a?(ActiveRecord::Type::Json)
+            # Native json/jsonb columns handle serialization at the adapter level.
+            # Skip the Type::Serialized wrapping (which double-serializes on write
+            # and fails on read) and use IndifferentJson to provide the same
+            # symbol/string key access that IndifferentCoder provides on text columns.
+            ActiveRecord::Type::IndifferentJson.new
```
byroot: There's another concern here, that some people might be running with that setup without realizing, and they'll need to fix the situation and migrate their data before they can upgrade.

We may need to provide a config flag to disable that new behavior.

activerecord/lib/active_record/type/indifferent_json.rb
markedmondson: Moved into `ActiveRecord::Store::IndifferentJsonType` in d3972fcd6 — scoped to the Store module with a comment noting it only handles Hash top-level objects.

activerecord/lib/active_record/store.rb
markedmondson: Added `config.active_record.store_native_json_columns` in d3972fcd6. Defaults to `false`, set to `true` in `load_defaults "8.2"`. Also added a documented entry in the `new_framework_defaults_8_2` template explaining the double-serialization migration concern.

PR #57181, opened 2026-04-13
[RF-DOCS] Rewrite the Asset Pipeline Guide

ayushn21:
### Motivation / Background

A ground up rewrite of the Asset Pipeline guide.

### Detail

I've rewritten the guide with a bigger focus on how a developer would use the asset pipeline and Propshaft, rather than some of the inner workings. 

A lot of detail was repeated and some bits were wrong. I've attempted to make the guide as concise as I could, and flow in the way a user would think about the asset pipeline.

Specifically, I'd like feedback about the diagram at the start of the guide. I'm not sure whether it's any good / helpful?

I've removed all info to migrate from Sprockets to Propshaft, instead linked to the migration guide in the Propshaft repo. I've also removed the section on `tailwindcss-rails` and instead linked to both that and `dartsass-rails`. It seems imbalanced to have a section on the Tailwind gem but not Dart Sass. Also since these are specific to certain libraries, I don't think they belong in the guide, and a link to their repos which has a Readme should be sufficient.

### Internal Audit

- The "After reading this guide, you will know:" intro doesn't mention Javascript or CSS at all. After reading it a newcomer isn't any wiser about what this guide explains. Maybe add "How to serve your Javascript and CSS assets" and "How to use bundlers for Javascript and CSS"
- The item "The main features of Propshaft, and how to set it up" is pretty tied to Propshaft (which I think is only a detail). I think we should make this a bit more libarary agnostic "The main features of the Asset Pipeline, and how to set it up." or maybe this should just be two separate items: "How to setup the Asset Pipeline" and "Using Propshaft for the Asset Pipeline".

#### What is an asset pipeline?

This section is more about Propshaft than the Asset Pipeline.
- It probably makes more sense if we describe the Rails Asset Pipeline instead of a more generic Asset Pipeline in the title (I don't think Asset Pipeline is a term used outside Rails). Maybe the title should be "What is the Asset Pipeline?" or "What is the Rails Asset Pipeline?"
- Maybe we can show a basic Asset Pipeline example where we show how an css file can get linked:

```
<%= stylesheet_link_tag "application", media: "all" %>
```

and gets turned into something like:

```
<link rel="stylesheet" href="assets/application-55f8f81.css" media="all"  /> 
```

- The following sentence can get outdated outdated quickly. "Propshaft is built for an era where transpilation, bundling and compression are less critical for basic applications, thanks to better browser support, faster networks and HTTP/2 capabilities." It can probably just be removed.
- Maybe we should add a heading when it starts describing Propshaft?
- The first command we show in this guide is creating a new application without the asset pipeline. It feels kind of negative. Is that the first thing a reader would want to do? Maybe that section should be added to the Propshaft Setup section instead, after the regular install command.

#### Propshaft Features

- If you just jump to this section there is no description of Propshaft. I think the description of the previous section should be move here instead.
- It feels like this section describes some more advanced gotcha's that would be better described after the introduction to Propshaft in the next section. For example the "Asset Load Order" becomes clearer after the more simple "Setup" section in "Working with Propshaft"
- The numbered list of dependency strategies would probably be easier to scan as regular sections with titles.

#### Working with Propshaft

- "Working" in the title is pretty ambiguous. As this section is mostly about configuring Propshaft, maybe it should be renamed to "Configuring Propshaft"? Or, if we combine it with the previous section: "Using Propshaft for the Asset Pipeline"?
- The steps to setup Propshaft are bit hard to follow in a numbered list format. Maybe it should just be written as a regular section?
- The note ensuring that "the expected compiled filenames end with .js or .css." was a bit unclear to me.
It was changed from the following which had more context about `.sass` and
`.coffee`: https://github.com/rails/rails/blob/c3f2b545f86748c0e792ef0de3211f9d948b92a6/guides/source/asset_pipeline.md?plain=1#L673-L674
As Propshaft doesn't do full precompilation, I'm wondering if this whole section should be moved.
- Maybe the CDN section should be a separate top section? It's a pretty large section but not visible in the sidebar.
- `http://www.example/assets/application-d0e099e021c95eb0de3615fd1d8c4d83.css` isn't a valid URL, it should proabably be: `http://www.example/.com/assets/application- d0e099e021c95eb0de3615fd1d8c4d83.css`

#### Sprockets to Propshaft

- The section is mostly relevant for those that want to migrate to Propshaft. If you start a new Rails would anyone migrate from Propshaft to Sprockets? So I think this section should be called "Migrating from Sprockets to Propshaft".
- As the Bundling and Transpilation sections are also mentioned in the "Advanced Asset Management" top section I think it makes more sense to group them there (or at least have the extensively describe them there)?
- The "Evolution of Asset Management Techniques" and "Sprockets vs. Propshaft" sections basically both describe features from Sprockets that aren't supported in Propshaft. Maybe it should be grouped and renamed to something like "Unsupported features in Propshaft"?
- As this section is mostly about migrating to Propshaft I think it should be moved under the "Advanced Asset Management" top section. I think we should order sections on audience size:
1. Using basic propshaft (most readers)
2. Using advanced asset management (readers that want advanced features)
3. Migrating from Sprockets (readers that haven't yet migrated)

#### Advanced Asset Management
- The "How x Works" sections should at least use regular bullet points instead of numbered lists, as there is no actual order in the points. They should probably even be regular paragraphs with code examples for the different commands.
- The notes for cssbundling-rails and importmap-rails should be an actual NOTE or WARNING
- Maybe we can make the headings more descriptive, describing the use-case instead of the library names:
  - jsbundling-rails -> "Javascript Bundling" or "Using Javascript Bundlers"
  - cssbundling-rails -> "CSS Bundling" or "Using CSS Bundlers"
  - tailwindcss-rails -> "Tailwind CSS integration"
  - importmap-rails -> "Using Import Maps"


guides/source/asset_pipeline.md
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
...
-[Sprockets Asset
-Pipeline](https://guides.rubyonrails.org/v7.2/asset_pipeline.html) in previous
-versions of the Rails Guides. You can also explore the [evolution of asset
-management techniques](#evolution-of-asset-management-techniques) to see how the
-Rails Asset Pipeline has evolved over time.
+NOTE: Before Rails 8, the asset pipeline was powered by [Sprockets](https://github.com/rails/sprockets). You can read about the [Sprockets Asset Pipeline](https://guides.rubyonrails.org/v7.2/asset_pipeline.html) in previous versions of the Rails Guides. A migration guide [is available here](https://github.com/rails/propshaft/blob/main/UPGRADING.md).
 
-Propshaft Features
-------------------
+Propshaft
```
p8: For the Caching and Active Jobs guide, we specifically name the defaults: "Default Store: Solid Cache" and "Default Backend: Solid Queue". This clarifies what "Solid Cache" and "Solid Queue" are for newcomers.

Maybe we could do that here as well, although it's not that easy to switch to another pipeline.
Also is importmap-rails considered part of the default pipeline?
```suggestion
Default Pipeline: Propshaft
```

guides/source/asset_pipeline.md
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
...
+<%= stylesheet_link_tag "application", media: "all" %>
+```
+
+```html
+<link rel="stylesheet" href="assets/application-55f8f81.css" media="all" />
+```
+
+The above example demonstrates a _fingerprint_ or _digest_ in the filename (`55f8f81`). This is calculated by running the file's contents through a [cryptographic hash function](https://en.wikipedia.org/wiki/Cryptographic_hash_function) — hence, when its contents change, the filename also changes. This allows for an aggressive caching strategy as stale assets will never be served.
+
+The core component of the Rails asset pipeline is [**Propshaft**](https://github.com/rails/propshaft). Propshaft is responsible for reading source assets, fingerprinting them, and placing them in the `public/` folder. It **DOES NOT** do any transpiling or bundling. Other gems need to be plugged into Propshaft for this functionality. See [Advanced CSS and JavaScript Management](#advanced-css-and-javascript-management) below.
```
p8: Here we already start introducing Propshaft. Maybe we should start the "Propshaft" section here?

guides/source/asset_pipeline.md
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
 
 After reading this guide, you will know:
 
-* What is an asset pipeline.
-* The main features of Propshaft, and how to set it up.
-* How to migrate from Sprockets to Propshaft.
-* How to use other libraries for more advanced asset management.
+* What the Rails asset pipeline does.
+* What Propshaft is and how it serves CSS, JavaScript, and other assets.
+* What Subresource Integrity is and how Propshaft leverages it.
```
p8: Is Subresource Integrity such an important topic it deserves to be listed here? I think the points we list here should somewhat mirror the top sections.
So I'd expect "Serving Assets" as a list item instead of "Subresource Intergrity".

guides/source/asset_pipeline.md
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
...
+
+The asset pipeline takes a source file, adds a _fingerprint_ (also called a _digest_) to the filename, and places it in the `public/assets/` folder for delivery. It may also preprocess source file before fingerprinting — for example, CSS and JavaScript files may be transpiled and bundled.
+
+<br><br>
+
+![Asset Pipeline Flow Diagram](images/asset_pipeline/asset_pipeline_flow_diagram.png)
+
+<br><br>
+
+Here's an example of how the asset pipeline links a stylesheet in an HTML document:
```
p8: It could be a bit confusing to new comers what Rails does and what the Asset Pipeline computes.
Maybe make it more explicit?
```suggestion
Here's an example of how the asset pipeline computes the path to a stylesheet:
```

guides/source/asset_pipeline.md
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
...
+
+```erb
+<%= stylesheet_link_tag "application", media: "all" %>
+```
+
+```html
+<link rel="stylesheet" href="assets/application-55f8f81.css" media="all" />
+```
```
p8, line 46: It might be a bit confusing for new comers that the `stylesheet_link_tag` generates the resulting HTML.
Maybe explain that it results in the render HTML?
````suggestion
```erb
<%# app/views/layouts/application.html.erb %>
<%= stylesheet_link_tag "application", media: "all" %>
```

This will render as follows:

```html
<link rel="stylesheet" href="assets/application-55f8f81.css" media="all" />
```
````

Or maybe flatten it in a single example?
````suggestion
```ruby
stylesheet_link_tag("application", media: "all")
=> "<link rel=\"stylesheet\" href=\"assets/application-55f8f81.css\" media=\"all\" />"
```
````

File: guides/source/asset_pipeline.md [hunk 6]
```diff
@@ -7,295 +7,227 @@ This guide explains how to handle essential asset management tasks.
...
-exists at `public/assets/rails.png`, it will be served by the web server.
+#### Images and Files
 
-Alternatively, if you are using fingerprinted assets (e.g.,
-`rails-f90d8a84c707a8dc923fca1ca1895ae8ed0a09237f6992015fef1e11be77c023.png`),
-Propshaft will also serve these correctly. The fingerprint is automatically
-applied during the precompilation process.
+Load images using the `image_path` helper:
+
+```erb
```
p8: It would be nice if we could show the resulting HTML of the `image_tag` and `image_path` here as well.

guides/source/asset_pipeline.md
```diff
@@ -315,180 +247,291 @@ export default class extends Controller {
...
+Install `jsbundling-rails` to configure your desired builder:
+
+```
+$ bundle add jsbundling-rails
+$ bin/rails javascript:install:[bun|esbuild|rollup|webpack]
+```
+
+This will configure a `package.json` file and set up a `build` script within it. Your can now use your chosen builder to manage your JavaScript.
+
+NOTE: Use `bin/dev` to run your server in development. This will invoke the `build` script with the `--watch` flag so your JavaScript updates automatically when you edit your files.
```
p8: Maybe use a [TIP ](https://edgeguides.rubyonrails.org/ruby_on_rails_guides_guidelines.html#tip)here instead, as `bin/dev` is not necessarily relevant to the understanding of `jsbundling-rails` 
```suggestion
TIP: Use `bin/dev` to run your server in development. This will invoke the `build` script with the `--watch` flag so your JavaScript updates automatically when you edit your files.
```

guides/source/asset_pipeline.md
```diff
@@ -497,70 +540,47 @@ config.file_watcher = ActiveSupport::EventedFileUpdateChecker
...
-useful when precompiling assets for production as part of a build step that
-otherwise does not need access to the production secrets.
+[environment variable]: https://en.wikipedia.org/wiki/Environment_variable
 
 ```bash
 $ RAILS_ENV=production SECRET_KEY_BASE_DUMMY=1 rails assets:precompile
 ```
 
-By default, assets are served from the `/assets` directory.
+WARNING: When precompiled assets are present in the development environmnemt, the application
```
p8:
```suggestion
WARNING: When precompiled assets are present in the development environment, the application
```

guides/source/asset_pipeline.md
```diff
@@ -586,47 +606,56 @@ location ~ ^/assets/ {
...
+For example, if your Rails application is running at `example.com`
+and your CDN is configured at `mycdnsubdomain.fictional-cdn.com`, then when a
 request is made to `mycdnsubdomain.fictional-cdn.com/assets/smile.png`, the CDN
-will query your server once at `example.com/assets/smile.png` and cache the
-request. The next request to the CDN that comes in to the same URL will hit the
-cached copy. When the CDN can serve an asset directly the request never touches
-your Rails server. Since the assets from a CDN are geographically closer to the
-browser, the request is faster, and since your server doesn't need to spend time
-serving assets, it can focus on serving application code.
+will query your server once at `yourrailsapp.com/assets/smile.png` and cache the
```
p8: If the app runs at `example.com` the CDN will also call `example.com`:
```suggestion
will query your server once at `example.com/assets/smile.png` and cache the
```

guides/source/asset_pipeline.md
```diff
@@ -586,47 +606,56 @@ location ~ ^/assets/ {
...
 
 The CDN you provisioned should give you a custom subdomain for your application
-such as `mycdnsubdomain.fictional-cdn.com` (note fictional-cdn.com is not a
-valid CDN provider at the time of this writing). Now that you have configured
-your CDN server, you need to tell browsers to use your CDN to grab assets
-instead of your Rails server directly. You can do this by configuring Rails to
-set your CDN as the asset host instead of using a relative path. To set your
-asset host in Rails, you need to set [`config.asset_host`][] in
-`config/environments/production.rb`:
+such as `mycdnsubdomain.fictional-cdn.com` (you can also use a CNAME DNS record to point at the CDN from a subdomain of your main application. For example: `assets.yourrailsapp.com` can point at `mycdnsubdomain.fictional-cdn.com` using a CNAME DNS record).
```
p8:
```suggestion
such as `mycdnsubdomain.fictional-cdn.com` (you can also use a CNAME DNS record to point at the CDN from a subdomain of your main application. For example: `assets.example.com` can point at `mycdnsubdomain.fictional-cdn.com` using a CNAME DNS record).
```

p8: If we add dashes to the top-level domain, we can add them to the subdomain as well, for readability (this would need to be changed in more places)
```suggestion
such as `my-cdn-subdomain.fictional-cdn.com` (you can also use a CNAME DNS record to point at the CDN from a subdomain of your main application. For example: `assets.yourrailsapp.com` can point at `my-cdn-subdomain.fictional-cdn.com` using a CNAME DNS record).
```

guides/source/asset_pipeline.md
```diff
@@ -674,85 +697,20 @@ option for your asset helper, which overwrites the value set in
...
-means any cache can store the request. Also we commonly want to set `max-age`
-which is how long the cache will store the object before invalidating the cache.
-The `max-age` value is set to seconds with a maximum possible value of
-`31536000`, which is one year. You can do this in your Rails application by
-setting
+CDN is used, a browser will use this information to cache contents. Generally we want
+our Rails server to tell our CDN (and browser) that the asset is "public". Also we want to
+set `max-age` which is how long the cache will store the object before invalidating the cache.
+The `max-age` value is set to seconds with a maximum possible value of `31536000`, which is one year
+because the figerprinted filename ensure stale assets are never served.
```
p8:
```suggestion
because the fingerprinted filename ensures stale assets are never served.
```

guides/source/asset_pipeline.md
ayushn21: Yep I agree, updated.

guides/source/asset_pipeline.md
ayushn21: I prefer the separate example structure. Updated.

guides/source/asset_pipeline.md
ayushn21: I've moved it a bit higher, just after the note on HTTP/2 below. I'd like to keep it there because I want the reader to have at least a vague idea of what "Propshaft" is before they see the heading.

guides/source/asset_pipeline.md
ayushn21: I don't think Propshaft is comparable to Solid Cache and Solid Queue. Those plug into adapters provided by Rails for other backends.

The concept of an Asset Pipeline is specific to Rails. You can't "plug in" another pipeline, you'd be assembling your own somehow, probably using a set of JS tools, which is not an approach we should be considering for this guide.

I guess in a way `importmap-rails` could be considered part of the default pipeline. But that again goes to show that the pipeline is made up of different things, it's not "pluggable" so I don't think we need to mention that it's the "default".

guides/source/asset_pipeline.md
ayushn21: Good idea, done

guides/source/asset_pipeline.md
ayushn21: Whoops, typo. Changed the earlier mention of `example.com` to `yourrailsapp.com` to make the domain name more descriptive.

p8: We try to use example.com through out the guides. Someone might register `yourrailsapp.com` and host malware or abuse it for SEO purposes.

guides/source/asset_pipeline.md
p8: There still is sprockets, but I agree that new Rails apps should not use it as an alternative.

p8:
To me, the description in the index focuses too much on Propshaft (it's not called the Propshaft Guide).
> The asset pipeline guide explains how to use Propshaft, a framework that handles essential asset management tasks. 

Maybe:
> The asset pipeline guide explains how to transform, cache, and serve static assets such as JavaScript, CSS, and image files.

ayushn21:
> To me, the description in the index focuses too much on Propshaft (it's not called the Propshaft Guide).

Agreed, just updated as suggested

p8:
@ayush I think this one is ready to be merged. Can you wrap to 80 columns? 
https://edgeguides.rubyonrails.org/ruby_on_rails_guides_guidelines.html#column-wrapping

ayushn21:
@p8 ah, wasn't aware of that guideline and my editor automatically wraps lines so hadn't noticed. Is there an automated way to do this? Can't seem to find one in my brief search.

ayushn21:
@p8 I've used Prettify to format it:

`npx prettier --prose-wrap always --print-width 80 --write asset_pipeline.md`

EDIT: Made some manual changes to the automated edit, and squashed commits. Should be good to go now.

p8:
@ayushn21 I was going to merged this but noticed the diagram doesn't have a dark-mode variant:

image

Also the styling doesn't match the one in the "Getting Started" guide:
image

Quickest path to getting this merged is removing the diagram for now and making the diagram in the style of the other ones in a separate PR.

ayushn21:
> I was going to merged this but noticed the diagram doesn't have a dark-mode variant:

Yep I know, I quickly put it together in Excalidraw and had asked for feedback on Basecamp, but since no one replied I left it as-is.

My preference would be to wait until we have a final diagram and then merge it all at once.

guides/source/asset_pipeline.md
```diff
@@ -315,180 +312,359 @@ export default class extends Controller {
...
+--------------------------------------
 
-### Setup
+Since Propshaft doesn't transpile or bundle CSS and JavaScript, Rails provides a
+number of gems should you wish to to do so. These gems augment Propshaft by
+applying transformations to the source files before _handing them over_ to
+Propshaft for fingerprinting and delivery.
 
-Follow these steps for setting up Propshaft in your Rails application:
+### JavaScript Import Map
```
p8, line 377: Now that the new JavaScript guide will describe "JavaScript Import Map" I'm not sure we should describe it here. Import Map doesn't transpile, bundle or apply transformations.

ayushn21: True, but it does form part of the asset pipeline. The asset pipeline is bigger than just Propshaft, even though that's probably 95% of it. 

We need to cover the `jsbundling-rails` and `cssbundling-rails` gems so for completeness I think we should also include import maps, since it's the default.

Also I think the detail here is a bit more advanced than what we have in the JS guide.

guides/source/asset_pipeline.md
```diff
@@ -497,70 +673,56 @@ config.file_watcher = ActiveSupport::EventedFileUpdateChecker
...
-
 #### Precompiling Assets
 
-In production, precompilation is typically run during deployment to ensure that
-the latest versions of the assets are served. Propshaft was explicitly not
-designed to provide full transpiler capabilities. However, it does offer an
-input -> output compiler setup that by default is used to translate `url(asset)`
-function calls in CSS to `url(digested-asset)` instead and source mapping
-comments likewise.
+Procompilation processes all assets through the asset pipeline and writes them
```
p8:
```suggestion
Precompilation processes all assets through the asset pipeline and writes them
```

guides/source/asset_pipeline.md
```diff
@@ -315,180 +312,359 @@ export default class extends Controller {
...
+will automatically pull in the latest version each time you reload the page.
+
+#### Automatic Reloading of Assets
+
+Propshaft automatically checks for updates to assets like JavaScript, CSS, or
+images with every request. This means you can edit these files, reload the
+browser, and instantly see the changes without needing to restart the Rails
+server.
+
+#### Improving Performance With A File Watcher
```
p8: Articles should be lower cased. (I've opened a PR to clarify this in the guidelines).
```suggestion
#### Improving Performance With a File Watcher
```

ayushn21: My bad ... knew about the guideline, just a typo.

guides/source/asset_pipeline.md
```diff
@@ -315,180 +312,359 @@ export default class extends Controller {
...
-    ```
+It supports a number of builders such as [Bun](https://bun.sh),
+[esbuild](https://esbuild.github.io/), [rollup.js](https://rollupjs.org/), and
+[Webpack](https://webpack.js.org/). See the
+[Readme](https://github.com/rails/jsbundling-rails/blob/main/README.md) for
+further details.
 
-5. Set Root Route
+The output from your chosen builder will be written to `app/assets/builds/`
+where Propshaft will pick it up for fingerprinting and delivery.
```
p8: I think we should mention here that `jsbundling-rails` by default configures all the bundlers to build files from `app/javascript`.
The jsbundling-rails README has the following text:
> Whenever the bundler detects changes to any of the JavaScript files in your
project, it'll bundle app/javascript/application.js into app/assets/builds/application.js
(and all other entry points configured).

Also, the documentation for Propshaft, jsbundling-rails and all the bundlers use the term "bundler" instead of "builder".

guides/source/asset_pipeline.md
p8: I can see it makes sense to mention importmap-rails, when we mention jsbundling-rails and cssbundling-rails.
It's a bit confusing with the following title and opening paragraph, that we start with a gem that doesn't transpile, bundle or apply transformations:
> ## Advanced CSS and JavaScript Management
> Since Propshaft doesn't transpile or bundle CSS and JavaScript, Rails provides a number of gems should you wish to to do so. These gems augment Propshaft by applying transformations to the source files before handing them over to Propshaft for fingerprinting and delivery.

"Advanced JavaScript Management" implies there also is "Beginner" or "Easy Javascript Management" (which is `importmap-rails`, I think).

Maybe make clear that transpiling and bundling isn't required for import maps.
Something like:
> To reduce complexity, Propshaft doesn't transpile or bundle CSS and Javascript. Transpiling or bundling is no longer needed when using importmap-rails gem.
>
> If your application requires transpiling or bundling, Rails also provides a number of gems should you wish to to do so. These gems augment Propshaft by applying transformations to the source files before handing them over to Propshaft for fingerprinting and delivery.

Also, the `importmap-rails` section is much more detailed than the "more advanced" `*bundling-rails` sections (we don't mention adding dependencies for js-bundling). 
I don't think adding dependencies is part of the Asset Pipeline and is probably better left to the JS guide.

ayushn21: Ok yeah all fair points. I think the best solution here was to reword the section title, and the intro paragraph:

https://github.com/rails/rails/pull/57181/changes/42c9e768b9a6c51ba06d90ca38e7240044457ee7

> (we don't mention adding dependencies for js-bundling)

We don't do this because adding dependencies in the realm of the bundler itself, the gem just integrates with the bundler. The reader is expected to understand how the bundler works, or refer to its docs for that stuff.

`importmap-rails` is a Rails component, and has its own way to manage dependencies, which is why I've mentioned it. I think we should keep it for complete-ness. It's a common question a ready will have I don't want to break up the narrative by sending them to a whole new guide for it. It's just a couple of lines anyway, we don't go into a lot of details so I think it's worth keeping.

guides/source/asset_pipeline.md
ayushn21: Done!

ayushn21:
@p8 I've updated this with the new diagram, including the dark mode variant. I think it should be ready to merge, is there anything else that needs addressing?

guides/source/asset_pipeline.md
p8: Thanks! I think this looks great now.

PR #57182, opened 2026-04-13
[ci skip] Fix controller name mismatch in Wishlists guide

maxonvim:
The guide uses `resource :wishlist` (singular) which expects a controller named `Products::WishlistController`, but instructs to create `Products::WishlistsController` (plural), causing an "uninitialized constant" error.

### DIFF
--- a/guides/source/wishlists.md
+++ b/guides/source/wishlists.md
@@ -364,10 +364,10 @@ If a user is not logged in, they'll see a link to sign up. Logged in users will
 see a form to select a wishlist and add the product to it.
 
 Next, create the controller to handle this form in
-`app/controllers/products/wishlists_controller.rb` with the following:
+`app/controllers/products/wishlist_controller.rb` with the following:
 
 ```ruby
-class Products::WishlistsController < ApplicationController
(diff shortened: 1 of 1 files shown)


akhilgkrishnan (approved):
@maxonvim Thanks for the change

skipkayhil:
The current text is correct, the only case you should need `WishlistController` is if you use `resources` instead of `resource`:

```ruby
resources :wishlists # WishlistsController
resource :wishlist # WishlistsController

resources :wishlist # WishlistController
```

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.eager_load = false
  config.hosts << "example.org"

end
Rails.application.initialize!

  resource :wishlist, only: :show
end

class WishlistsController < ActionController::Base

  def show
    head :ok
  end
end

  include Rack::Test::Methods

  def test_returns_success
    get "/wishlist"
    assert last_response.ok?
  end

  private
    def app
      Rails.application
    end
end
```

PR #57194, opened 2026-04-16
Clarify TimeZone#parse docs for invalid strings

### LINKED ISSUE #57193: Update documentation for `ActiveSupport::TimeZone#parse` to reflect that it returns `nil` for invalid strings
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activesupport"
end

require "active_support/all"

# Set a time zone
Time.zone = "UTC"

# According to the documentation: 
# "If the string is invalid then an ArgumentError could be raised."
# https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html#method-i-parse

invalid_input = "invalid date string"
result = Time.zone.parse(invalid_input)

puts "Result: #{result.inspect}"

if result.nil?
  puts "Current behavior: returns nil"
else
  puts "Current behavior: returns #{result.class}"
end
```

### Expected behavior

The documentation states that an `ArgumentError` could be raised for invalid strings. Therefore, it should either raise an `ArgumentError` or the documentation should be updated to reflect that it returns `nil` when the string cannot be parsed.

### Actual behavior

The method returns `nil` instead of raising an `ArgumentError`.

```text
Result: nil
Current behavior: returns nil
```

### System configuration

**Rails version**: 7.1, 7.2, 8.0

**Ruby version**: 3.2, 3.3

Saidbek:
### Motivation / Background

Fixes https://github.com/rails/rails/issues/57193

`ActiveSupport::TimeZone#parse` currently returns `nil` for invalid strings, but its API docs said an `ArgumentError` could be raised. This updates the method documentation to match actual behavior.

### Testing

`bundle exec ruby -Itest test/time_zone_test.rb -n test_parse_returns_nil_when_string_without_date_information_is_passed_in`

### Checklist

### DIFF
--- a/activesupport/lib/active_support/values/time_zone.rb
+++ b/activesupport/lib/active_support/values/time_zone.rb
@@ -456,7 +456,7 @@ module ActiveSupport
     #
     #   Time.zone.parse('Mar 2000') # => Wed, 01 Mar 2000 00:00:00 HST -10:00
     #
-    # If the string is invalid then an +ArgumentError+ could be raised.
+    # Returns +nil+ when the string is invalid and cannot be parsed.
     def parse(str, now = now())
       parts_to_time(Date._parse(str, false), now)
     end


maxfelsher:
I think the prior phrasing could have been clearer, but it was correct in saying that an ArgumentError can be raised with invalid strings. See this test: https://github.com/rails/rails/blob/db4bd562d90427c21a4525412c25372dd6994a65/activesupport/test/time_zone_test.rb#L495-L503

The distinction seems to be between strings that don't look like dates at all and ones that are formatted like dates but have invalid values for some of the date components. For the ones that don't look like dates, `Date._parse` returns an empty hash, and `ActiveSupport::TimeZone#parse` returns nil. For the others, `Date._parse` returns a non-empty hash, but the values don't make sense. For example, in the test above, `zone.parse("9000")` calls `Date._parse("9000")`, which returns a hash corresponding to the 0th day of the 90th month. And then an ArgumentError is raised when passing those values to `Time.new`.

I think the new language in this pull request can be interpreted as implying that nil is always returned for invalid strings, which isn't the case. Maybe it should be updated to mention both the nil case and the ArgumentError case?

byroot:
> Maybe it should be updated to mention both the nil case and the ArgumentError case?

I think it would be better to unify the behavior, so that we either always return nil or always raise. Having two behaviors isn't great.

maxfelsher:
> > Maybe it should be updated to mention both the nil case and the ArgumentError case?
> 
> I think it would be better to unify the behavior, so that we either always return nil or always raise. Having two behaviors isn't great.

Yeah, I wasn't sure how much appetite there was for changing the behavior, but I agree that it'd ideally be unified. I'm having trouble thinking of a use case for distinguishing between "doesn't look like a date" and "looks like a date but with invalid values".

It looks like `Time.parse` from the Ruby stdlib (which I'm assuming `ActiveSupport::TimeZone#parse` is modeled on) raises an ArgumentError in both cases, so if compatibility is going to be broken in one direction, it might make sense to align with that. But I don't know the original reasoning behind returning nil in some cases--it looks like it goes back to https://github.com/rails/rails/commit/06a7c2948a8dbf31357b552d468fcf42002736e7, but I don't see an explanation for the change.

byroot:
In the meantime I reverted the change. 

I don't have a very strong opinion on which direction we should go toward. Just like that I'd say ArgumentError is better, as it's less likely to go unnoticed.

However I wonder whether that new behavior should be behind a configuration flag as to allow for easier upgrade? I'd say probably.

Feel free to open a PR if you desire.

maxfelsher:
Thanks! I see @Saidbek has opened a PR to unify the behavior (behind a configuration flag), which is all good by me.

PR #57196, opened 2026-04-16
Add native C extension for Inflector#camelize and #underscore (7.1× / 4.0× speedup)

matrix9180:
## Motivation

`Inflector.camelize` and `Inflector.underscore` are called on every constant
lookup, routing dispatch, and model load. They're among the hottest paths in
any Rails process. Both methods rely on `gsub`/`sub` with block callbacks,
which allocate a `MatchData` object and one or two capture-group strings
(`$1`, `$2`) on every regex match. For a typical call like
`camelize('active_support/core_ext/string')` (7 segments) this amounts to
~18 Ruby objects / ~1 200 bytes of GC work per call.

## Change

Add a C extension (`activesupport/ext/active_support/inflector_core.c`) with
two functions:

- **`_camelize_native(str, acronyms)`** — replaces the `gsub!` hot loop in
  `camelize` for the `uppercase_first_letter: true` (default) path.
- **`_underscore_native(word)`** — replaces the lookbehind `gsub!` in
  `underscore` that inserts `_` at case boundaries.

Both are single-pass byte scanners that write directly into a pre-sized output
buffer. No regex, no `MatchData`, no capture-group strings.

Pure-Ruby fallbacks with the same names are defined first in
`inflector/methods.rb` and remain active when the extension cannot be loaded,
so all environments without a compiled extension continue to work without
behavioural change.

## Numbers

Measured on Ruby 4.0.2 with `benchmark-ips` + `memory_profiler`, no acronyms
defined (the default):

| Method | Objects/call (before → after) | Throughput (before → after) |
|---|---|---|
| `camelize` | 18 → 4 (77% fewer) | 61 k → 438 k i/s (**7.1×**) |
| `underscore` | 5 → 1 (81% fewer) | 90 k → 364 k i/s (**4.0×**) |

The `underscore` result (1 object/call) is the theoretical floor: one
allocation for the returned String itself, nothing else.

## Scope

- The **lowerCamelCase** path of `camelize` (`uppercase_first_letter: false`)
  is uncommon and stays pure Ruby.
- The **acronym pre-processing** step of `underscore` (the
  `acronyms_underscore_regex` gsub) stays pure Ruby; only the
  boundary-detection pass moves to C.
- `humanize`, `pluralize`, `singularize` are unaffected by this PR.

## Tests

```
472 runs, 1670 assertions, 0 failures, 0 errors, 0 skips
```

All existing inflector tests pass with the extension loaded.

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,23 +1,6 @@
-*   Replace the regex-based hot paths in `ActiveSupport::Inflector.camelize`
-    and `ActiveSupport::Inflector.underscore` with native C implementations.
-
-    Both methods previously allocated a `MatchData` object and one or two
-    capture-group strings on every regex match.  For a typical call such as
-    `camelize('active_support/core_ext/string')` this amounted to ~18 Ruby
-    objects (~1 200 bytes) per call.  The C extension replaces those paths
-    with single-pass byte scanners that write directly into a pre-sized output
-    buffer.
-
-    Measured improvement (Ruby 4.0.2, no acronyms defined):
-
-    | Method        | Objects/call | Bytes/call | Throughput              |
-    |---------------|--------------|------------|-------------------------|
-    | `camelize`    | 18 → 4       | 1208 → 224 | 61k → 438k i/s (7.1×)  |
-    | `underscore`  |  5 → 1       |  522 →  96 | 90k → 364k i/s (4.0×)  |
-
-    A pure-Ruby fallback remains active if the extension cannot be loaded
-    (e.g. running from source without compiling), so behaviour is unchanged in
-    all environments.
+*   Speed up `ActiveSupport::Inflector.camelize` and `.underscore` with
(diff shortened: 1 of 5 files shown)


fatkodima:
Rails should not have C code and optimizing these methods will not make a big difference in real applications.
Iirc, there are existing gems for these methods.

matrix9180:
Why thank you for the outright dismissal without any further consideration or discussion. I'll just fork the whole thing and do it anyway.

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

matrix9180:
Optimizing methods that run thousands of times per request and generate tons of throw away objects won't impact real performance.case closed. Thanks for investigating.

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

mohits:
Hi @matrix9180 - given the way that Ruby works, could these be a separate gem that speed-seekers could use since including them after Rails will mean that they will override the Ruby versions?

PR #57201, opened 2026-04-17
Fix TTOU hanging Rails when ffmpeg is spawned for video previews

jdelStrother:
### Motivation / Background

When generating video previews in ActiveStorage, Rails launches ffmpeg as a subprocess. ffmpeg watches for keypresses (eg q to quit), calling tcsetattr to configure the supervising terminal. 

If Rails is run with its own process group (eg using [overman](https://github.com/spinels/overman) that spawns processes with pgroup:true), this causes the Rails process to receive a TTOU signal, which results in Rails hanging indefinitely.

### Detail

We can avoid this by passing /dev/null to ffmpeg's stdin.

### Additional information

I can't see any benefit to providing stdin to any of the ActiveStorage::Previewers, so have just disabled it at the Previewer level rather than just for VideoPreviewer.

Another alternative would be to add `-nostdin` to ActiveStorage.video_preview_arguments

ActiveStorage has some other usage of, eg, `IO.popen("ffprobe"...)`, but from what I can tell ffprobe doesn't do anything funky with stdin, so I've left those as-is.

Tests for this might be tricky, let me know if you think they're really necessary.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/previewer.rb
+++ b/activestorage/lib/active_storage/previewer.rb
@@ -79,7 +79,7 @@ module ActiveStorage
         to.binmode
 
         open_tempfile do |err|
-          IO.popen(argv, err: err) { |out| IO.copy_stream(out, to) }
+          IO.popen(argv, in: IO::NULL, err: err) { |out| IO.copy_stream(out, to) }
           err.rewind
 
           unless $?.success?


guilleiguaran:
Let's add a CHANGELOG entry for this.

jdelStrother:
Sure, see what you think to the updated version

jdelStrother:
@byroot thanks! Though - was it intentional to merge it into main without the changelog entry? And in future would it be better if I do the changelog as a separate commit?

byroot:
There's not a clear policy on this, but yes, for bugfixes I backport, I remove the changelog from the commit I merge on master, and include it in the backport commit.

Does make much sense to include a bugfix in main changelog when it is backported.

PR #57204, opened 2026-04-17
Let add_column raise if :null is true for PKs

fxn:
As the title says, this patch lets `add_column` raise if `:null` is true for PKs. See why in the `CHANGELOG` entry included in the patch.

Some tests uncovered the SQLite adapter could be actually setting `null: true` for PKs in `copy_table`! A code comment in the adapter explains the fine details.

As a curiosity, this started as a documentation patch that I thought was going to take five minutes, I wanted to make the docs of `:null` more clear. Then, I realized that the default is to allow nullables **except for PKs**, where `null: false` is forced unconditionally, and started to write something like "For PKs the option is ignored...". That sentence felt wrong.

This happens when you document: Sometimes, the docs you come with tell you something is actually not right, and you did not realize it until you dug into it and tried to verbalize behavior in detail.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Let `add_column` raise `ArgumentError` if `:null` is set to a true value
+    when defining a primary key.
+
+    Primary keys get a `NOT NULL` constraint unconditionally. In particular,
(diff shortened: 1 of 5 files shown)


fxn:
CI is failing in the initial build step. The suites of the four adapters pass in my machine, and the patch is short, so I'll merge.

PR #57206, opened 2026-04-18
Add enums to query command output

geoffharcourt:
### Motivation / Background

In #57156 a new feature was added to make it easier for external tools such as agents to query the application for the database schema. We've been using a similar tool on our own and we found that adding the enum defintions on a model enabled LLMs to get a better result to some questions with fewer (or sometimes zero) overall queries to the database. Having the enum definitions can remove the need to make a database query just to see the text value of the enums, etc.

### Detail

This change updates the `run_models` and `run_schema` methods invoked by the `rails query models` and `rails query schema` commands to return the enums on the models/tables along with the field definitions and associations.

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -23,6 +23,10 @@
 
     *Lewis Buckley*
 
+*   Update `bin/rails query` to include enum defintions in model and schema output.
+
+    *Geoff Harcourt*
+
 *   Avoid adding `Rack::Sendfile` to the middleware stack if `config.action_dispatch.x_sendfile_header` is `nil`.
 
     The middleware behave as a noop in such case so it's pointless to have it in the stack.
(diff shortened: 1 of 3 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
```
byroot: You probably didn't mean to commit that.

geoffharcourt: Yikes, thought I removed this but hadn't pushed. Apologies for the noise!

lewispb:
Curious if you've used the original implementation yet @geoffharcourt? Is `rails query schema <table>` which already returns enums insufficient?

geoffharcourt:
@lewispb we had a different organization of what was on the model, but I think you're right that this is superfluous. Thanks for reviewing.

PR #57207, opened 2026-04-18
Fix nested `with_connection(prevent_permanent_checkout: true)` incorrectly clearing sticky lease set by `lease_connection`

bensheldon:
When `lease_connection` is called to permanently lease a connection within a `with_connection` block, a subsequent nested `with_connection(prevent_permanent_checkout: true)` would incorrectly clear the sticky flag due to a `&& !sticky_was` guard that prevented restoring `sticky_was` when it was `true`. This caused the outer `with_connection` block to release the connection on exit, even though it had been explicitly leased.

The fix removes the `&& !sticky_was` guard so the sticky state is always restored to its prior value when `prevent_permanent_checkout` is used.

### Example

This is the simple expected preexisting behavior:

```ruby
with_connection do
  lease_connection
end

# 🔒 lease is retained 
```

This PR fixes this behavior:

```ruby
with_connection do
  lease_connection
  with_connection(prevent_permanent_checkout: true) { true }
end

# 🔓 Currently, lease is released ❌
# 🔒 Fix is that lease should be retained ✅ 
```

I believe that I have fix is the correct behavior because I think `with_connection` should always be maintaining/restoring the outer context. 

### Why it matters

I have an advisory lock library that requires reusing the same connection. Not maintaining outer context makes it possible for application code to inadvertently break the library e.g. 

```ruby
library.take_advisory_lock_with_connection_lease do
  # application code
  with_connection(prevent_permanent_checkout: true) do
    lease_connection
    # do stuff
  end

  # library has no leased connection to finalize the block ❌ 
end 
```

### Checklist

* [x] ~~CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.~~ Seems like a minor bufgix.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -455,13 +455,13 @@ module ActiveRecord
           begin
             yield lease.connection
           ensure
-            lease.sticky = sticky_was if prevent_permanent_checkout && !sticky_was
+            lease.sticky = sticky_was if prevent_permanent_checkout
           end
         else
           begin
             yield lease.connection = checkout
           ensure
-            lease.sticky = sticky_was if prevent_permanent_checkout && !sticky_was
+            lease.sticky = sticky_was if prevent_permanent_checkout
             release_connection(lease) unless lease.sticky
           end
         end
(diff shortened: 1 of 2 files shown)


bensheldon:
Test failure looks unrelated.

PR #57211, opened 2026-04-20
Add config to raise on invalid TimeZone#parse strings

Saidbek:
### Motivation / Background

Follow-up to #57194

As discussed, `ActiveSupport::TimeZone#parse` has two different behaviors for invalid strings, which makes it easy to miss bad input:
- Strings with no recognizable date information (e.g. `"foobar"`) return `nil`.
- Strings with out-of-range date components (e.g. `"9000"`, which `Date._parse` interprets as month 90) raise `ArgumentError`.

This PR unifies the behavior so `TimeZone#parse` always raises `ArgumentError` for any invalid string, matching the Ruby standard library's `Time.parse`. The change is gated behind a new `ActiveSupport.raise_on_invalid_time_zone_parse` flag (defaulting to `false`) to preserve backwards compatibility, and is enabled by default in `config.load_defaults "8.2"`.

### Detail

* `ActiveSupport.raise_on_invalid_time_zone_parse`
* Default: `false`; `true` under `load_defaults "8.2"`
* Documented in `configuring.md` and the method docs for `#parse`

### Checklist

### DIFF
--- a/activesupport/lib/active_support/railtie.rb
+++ b/activesupport/lib/active_support/railtie.rb
@@ -31,6 +31,14 @@ module ActiveSupport
       end
     end
 
+    initializer "active_support.raise_on_invalid_parse_string" do |app|
+      config.after_initialize do
+        if app.config.active_support.raise_on_invalid_parse_string
+          ActiveSupport::TimeZone.raise_on_invalid_parse_string = true
+        end
+      end
+    end
+
     initializer "active_support.set_authenticated_message_encryption" do |app|
       config.after_initialize do
         unless app.config.active_support.use_authenticated_message_encryption.nil?
--- a/activesupport/lib/active_support/values/time_zone.rb
+++ b/activesupport/lib/active_support/values/time_zone.rb
@@ -2,6 +2,7 @@
 
 require "tzinfo"
 require "concurrent/map"
+require "active_support/core_ext/class/attribute"
 
 module ActiveSupport
   # = Active Support \Time Zone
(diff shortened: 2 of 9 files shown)


Saidbek:
@byroot following up on your note on #57194 about unifying `TimeZone#parse` invalid-string behavior (prefer raise, behind a config flag for upgrades).

This PR does that:

* `config.active_support.raise_on_invalid_parse_string` (default `false`; `true` under `load_defaults "8.2"`)
* when enabled, unparseable strings like `"foobar"` raise `ArgumentError` instead of returning `nil`, matching out-of-range cases

Would you mind taking a look when you have a moment? Thanks!

activesupport/lib/active_support/values/time_zone.rb
```diff
@@ -189,6 +190,8 @@ class TimeZone
     UTC_OFFSET_WITHOUT_COLON = UTC_OFFSET_WITH_COLON.tr(":", "") # :nodoc:
     private_constant :UTC_OFFSET_WITH_COLON, :UTC_OFFSET_WITHOUT_COLON
 
+    class_attribute :raise_on_invalid_time_zone_parse, instance_accessor: false, default: false
```
byroot: You don't need `class_attribute`.


```suggestion
    @raise_on_invalid_time_zone = false
    singleton_class.attr_accessor :raise_on_invalid_time_zone_parse
```

activesupport/lib/active_support/railtie.rb
```diff
@@ -31,6 +31,14 @@ class Railtie < Rails::Railtie # :nodoc:
       end
     end
 
+    initializer "active_support.raise_on_invalid_time_zone_parse" do |app|
+      config.after_initialize do
+        if app.config.active_support.raise_on_invalid_time_zone_parse
+          ActiveSupport::TimeZone.raise_on_invalid_time_zone_parse = true
+        end
+      end
+    end
```
byroot: I think you might as well make the config be `ActiveSupport.raise_on_invalid_time_zone_parse`, this way you don't need an initializer.

activesupport/lib/active_support/values/time_zone.rb
Saidbek: done https://github.com/rails/rails/pull/57211/changes/82b6b3629a7191651347fb6ac3e74e9cf6cb4c5b

activesupport/lib/active_support/railtie.rb
Saidbek: done https://github.com/rails/rails/pull/57211/changes/cb24e6d80b21d598cd9b0a647abc4bd20f30d270

PR #57214, opened 2026-04-20
Avoid false dirty tracking for equivalent Action Text body HTML

afurm:
### Motivation / Background

This Pull Request has been created because `ActionText::RichText#body` can be considered dirty after load even when the rich text content has not changed.

Fixes #54812.

The reported case involves older `action_text_rich_texts.body` rows that contain a literal non-breaking space character. Current Action Text serialization canonicalizes that value as `&nbsp;`, so Active Record's default serialized dirty check compares the raw stored HTML against the canonicalized serialized HTML and sees a change. Applications using `body_changed?` can then emit false update events for unchanged rich text content.

### Detail

This Pull Request changes `ActionText::RichText#body` serialization to use `comparable: true`.

That makes Active Record compare deserialized `ActionText::Content` values through their existing equality behavior instead of comparing raw serialized HTML strings. As a result, equivalent HTML entity representations no longer mark the rich text body as dirty.

A regression test was added for a legacy-style row containing a literal non-breaking space, bypassing type serialization to reproduce the stored database value from the issue report. The Action Text changelog was also updated.

### Additional information

Validated with:

* `cd actiontext && bin/test test/unit/model_test.rb -n '/content stored with equivalent html entities is not dirty after load/'`
* `cd actiontext && bin/test test/unit/model_test.rb`
* `cd actiontext && bin/test test/unit/content_test.rb`
* `cd actiontext && bin/test $(find test -name '*_test.rb' ! -path 'test/system/*' ! -name 'javascript_package_test.rb')`
* `bundle exec rubocop actiontext/app/models/action_text/rich_text.rb actiontext/test/unit/model_test.rb`

`cd actiontext && bin/test` could not complete in this checkout because `JavascriptPackageTest` requires `rollup`, which is not installed locally. A broader run excluding only that JavaScript packaging test then hit unrelated system-test instability in browser upload tests, so the final broad validation used non-system Action Text tests.

### Checklist

### DIFF
--- a/actiontext/CHANGELOG.md
+++ b/actiontext/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Avoid marking unchanged rich text bodies as dirty when stored HTML contains
+    equivalent entity representations.
+
+    Fixes #54812.
+
+    *Andrii Furmanets*
+
 *   Support block children in editor elements alongside value.
 
     Blocks were introduced in #55827, but only as an alternative to the value
--- a/actiontext/app/models/action_text/rich_text.rb
+++ b/actiontext/app/models/action_text/rich_text.rb
@@ -39,7 +39,7 @@ module ActionText
     cattr_accessor :editors, instance_accessor: false, default: {}.freeze
     cattr_accessor :editor, instance_accessor: false
 
-    serialize :body, coder: ActionText::Content
+    serialize :body, coder: ActionText::Content, comparable: true
     delegate :to_s, :nil?, to: :body
 
     ##
--- a/actiontext/test/unit/model_test.rb
+++ b/actiontext/test/unit/model_test.rb
@@ -81,6 +81,16 @@ class ActionText::ModelTest < ActiveSupport::TestCase
     assert_equal "Hello world", message.content.to_plain_text
   end
 
+  test "content stored with equivalent html entities is not dirty after load" do
+    message = Message.create!(subject: "Greetings", content: "Hello world")
+    html = "<div>Test with non-breaking space: \u00A0</div>"
+
+    # Bypass type serialization to reproduce rich text records written by older Rails versions.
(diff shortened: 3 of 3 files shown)


bhushangahire:
Approach looks right to me. `comparable: true` ([type/serialized.rb:36-47](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/type/serialized.rb#L36-L47)) routes dirty detection through `ActionText::Content#==`, which compares canonicalized `to_html` on both sides — so legacy raw nbsp vs canonicalized `&nbsp;` no longer registers as a change. Surgical fix, single line.

A few questions / nits before this lands:

**1. Performance impact**

Previously `changed_in_place?` was a string compare. Now every check deserializes the raw stored HTML through Nokogiri to build an `ActionText::Content` and calls `to_html` on it. For models with many rich text fields, or callbacks that touch `body_changed?` often, this could measurably bump save overhead. Have you benchmarked? Even a quick `Benchmark.realtime` on a tight loop with a non-trivial document would be reassuring.

**2. Test only covers the legacy case**

The new test verifies an equivalent-but-different-bytes value is *not* dirty. Worth adding a sibling assertion that an actual content change *is* still detected, so the fix can't silently regress in the other direction. Something like:

```ruby
message.content = "Hello there"
assert_predicate message.content, :body_changed?
```

**3. CHANGELOG attribution**

The entry reads "*Andrii Furmanets*" — matches the repo convention from skimming recent entries, just confirming that's the form you want.

**4. `ActionText::Content#==` quirk** ([content.rb:200-206](https://github.com/rails/rails/blob/main/actiontext/lib/action_text/content.rb#L200-L206))

Returns `nil` (not `false`) for unrelated types. Inside `changed_in_place?` the comparison is `old_value != value`, so `nil != value` → `true`, which would mark as changed. Probably fine since the typed attribute guarantees both sides are `Content`, but worth a thought.

Not blocking on any of these — the core change is solid.

PR #57215, opened 2026-04-20
Fix ParameterFilter hash-lookup optimization for line-anchored regexps

alexcwatt:
### Motivation / Background

This is a follow-up fix to https://github.com/rails/rails/pull/57166.

### Detail

That PR treated /^token$/ and /\Atoken\z/ as interchangeable and converted both into exact-string hash lookups. But since ^ and $ are line anchors, it means that /^token$/ should matches multi-line keys (as uncommon as those may be) like "token\nfoo".

This PR fixes the optimization by splitting it into two variants: \A\z is an exact-string match, while ^$ populates a separate @exact_line_keys hash that is consulted for the full key and then for each line of keys containing "\n". Mixed anchors like /\Atoken$/ and /^token\z/ have asymmetric semantics that a hash can't represent and now fall back to regexp matching.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/parameter_filter.rb
+++ b/activesupport/lib/active_support/parameter_filter.rb
@@ -96,18 +96,6 @@ module ActiveSupport
     end
 
   private
-    # If the regexp is an anchored exact match like /^token$/ or /\Atoken\z/,
-    # returns the literal string.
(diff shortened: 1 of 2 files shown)


alexcwatt:
@byroot since you reviewed the previous PR, wanted to tag you on this one also

PR #57221, opened 2026-04-22
Empty SQL query `;` in PostgreSQL is a read query

fatkodima:
Currently, an empty (`;`) SQL query is considered as a write query:

```ruby
User.connection.execute(';')
# => ActiveRecord::ReadOnlyError: Write query attempted while in readonly mode: ;
```

These queries are often used as a faster alternative to `SELECT 1` (or something like that), which is a read query. Was surprised to find out that `;` is considered a write query.


fatkodima:
Initially implemented for all adapters, but sqlite don't like it
```
SQLite3::Exception: cannot use a closed statement (ActiveRecord::StatementInvalid)
```
and also mysql
```
ActiveRecord::StatementInvalid: Mysql2::Error: Query was empty
```
so switched only to postgres.

fatkodima:
Actually a bad idea, `connection.execute("; delete from users")`

PR #53890, opened 2024-12-10
Fix lazy attribute method definitiont to be thread safe

byroot:
Fix: https://github.com/rails/rails/issues/53667
Followup: https://github.com/rails/rails/pull/50594

We can't safely check if attribute methods were generated or not as we may be in a race with another thread. As such we must always consider that maybe the mehtod was just generated.

PR #57226, opened 2026-04-22
Fix method_missing performance regression for virtual SELECT alias attributes

### LINKED ISSUE #57183: Performance regression: method_missing raises NameError for virtual SELECT alias attributes
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # Toggle between these to compare:
  gem "rails", "7.2.3"       # regression
  # gem "rails", "7.2.2.2"   # baseline

  gem "sqlite3"
end

require "minitest/autorun"
require "benchmark"

  config.eager_load = false
  config.logger = Logger.new(File::NULL)
end
Rails.application.initialize!

  create_table :widgets, force: true do |t|
    t.string :name
    t.float :latitude
  end
end

class Widget < ActiveRecord::Base; end

  def test_virtual_attribute_access_performance
    Widget.create!(name: "test", latitude: 37.77)

    # Query with a virtual SELECT alias (common pattern for JOINs)
    widgets = Widget.select("widgets.*, 42 AS virtual_attr").load

    n = 50_000

    real_time = Benchmark.realtime { n.times { widgets.first.latitude } }
    virtual_time = Benchmark.realtime { n.times { widgets.first.virtual_attr } }
    bypass_time = Benchmark.realtime { n.times { widgets.first.read_attribute(:virtual_attr) } }

    puts
    puts "Rails #{Rails::VERSION::STRING} — #{n} attribute accesses"
    puts "  Real column (.latitude):           #{'%.4f' % real_time}s"
    puts "  Virtual attr (.virtual_attr):      #{'%.4f' % virtual_time}s"
    puts "  read_attribute(:virtual_attr):     #{'%.4f' % bypass_time}s"
    puts "  Virtual / Real ratio:              #{'%.1f' % (virtual_time / real_time)}x"
    puts

    # On 7.2.2.2 the ratio is ~12x. On 7.2.3 it jumps to ~30-70x due
    # to NameError raise/rescue in method_missing (PR #53890).
    assert_operator virtual_time / real_time, :<, 20,
      "Virtual attribute access is #{(virtual_time / real_time).round(1)}x slower " \
      "than real column access — expected <20x (7.2.2.2 baseline is ~12x)"
  end
end
```

**Results:**

| | Rails 7.2.2.2 | Rails 7.2.3 | Regression |
|---|---|---|---|
| Real column (`.latitude`) | 0.004s | 0.004s | 1.0x |
| Virtual attr (`.virtual_attr`) | 0.059s | 0.189s | **~3x** |
| `read_attribute` bypass | 0.008s | 0.007s | 1.0x |
| Virtual / Real ratio | 13.6x | 42.1x | |

### Expected behavior

Virtual attribute access via `method_missing` should have similar overhead to 7.2.2.2 — the old implementation checked `attribute_methods_generated?` (an ivar) and fell through to `read_attribute` cheaply.

### Actual behavior

On 7.2.3, each `method_missing` call for a virtual attribute (SELECT alias not backed by a real column) calls `self.class.public_instance_method(name)`, which raises and rescues a `NameError` (~9μs per call due to exception allocation + backtrace capture).

The regression is in #53890's new `method_missing`:

```ruby
method = begin
  self.class.public_instance_method(name)  # raises NameError for virtual attrs
rescue NameError
  nil
end
```

This is correct for thread safety, but the `NameError` raise/rescue path is expensive. For applications that project joined columns as SELECT aliases onto AR objects, the overhead scales with `result_count × virtual_attributes × access_count`.

**Production impact at our scale:** ~300 results × ~3 virtual attrs × 14 pipeline stages = ~12,600 `NameError` exceptions per request = **~113ms of added latency**. Filter stages doing zero I/O (just `results.select { |r| r.some_attr == value }`) became 4-6x slower.

**Workaround:** Define explicit accessors for known virtual attributes to bypass `method_missing` entirely:

```ruby
%w[virtual_attr].each do |attr|
  define_method(attr) { read_attribute(attr) } unless method_defined?(attr)
end
```

### System configuration

**Rails version:** 7.2.3 (regression introduced vs 7.2.2.2)

**Ruby version:** 3.4.9

hammadxcm:
### Motivation / Background

Fixes #57183.

PR #53890 (shipped in 7.2.3) changed `method_missing` in `ActiveRecord::AttributeMethods` to call `self.class.public_instance_method(name)`, which raises and rescues `NameError` for virtual SELECT alias attributes (not backed by real columns). Exception allocation + backtrace capture costs ~9μs per call, causing a **~3x regression** in virtual attribute access.

### Solution

Guard `public_instance_method` with `public_method_defined?` — a zero-allocation boolean check — so the `NameError` path is never hit for virtual attributes. The thread-safety properties introduced in #53890 are fully preserved since `public_method_defined?` is a simple lookup on the method table that `define_attribute_methods` just populated.

### Benchmark

```
50,000 attribute accesses (SQLite, virtual SELECT alias)

                              Before (7.2.3)    After (this PR)
Real column (.latitude):          0.004s             0.010s
Virtual attr (.virtual_attr):     0.189s             0.077s
Virtual / Real ratio:             42.1x              7.9x
```

### Checklist

- [x] CHANGELOG entry added
- [x] Tests added for virtual SELECT alias attribute access
- [x] Existing attribute_methods test suite passes (137 runs, 624 assertions, 0 failures)
- [x] Finder, base, and calculations tests pass

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,14 @@
+*   Fix performance regression in `method_missing` for virtual SELECT alias attributes.
+
+    `method_missing` used `public_instance_method` which raises and rescues
+    `NameError` for attributes not backed by real columns (e.g. `SELECT 42 AS
+    virtual_attr`). Guard with `public_method_defined?` to avoid exception
+    allocation on the common path.
+
+    Fixes #57183.
+
+    *Hammad Khan*
+
 *   Bump the minimum PostgreSQL version to 10.0.
 
     As part of this change, `supports_pgcrypto_uuid?` is deprecated because
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Fix performance regression in `method_missing` for virtual SELECT alias
+    attributes.
+
+    Fixes #57183.
+
+    *Hammad Khan*
+
 *   Deprecate `ActiveRecord::ConnectionAdapters::DatabaseStatements#create`
     in favor of `#insert`.
 
--- a/activerecord/lib/active_record/attribute_methods.rb
+++ b/activerecord/lib/active_record/attribute_methods.rb
@@ -480,10 +480,8 @@ module ActiveRecord
         self.class.define_attribute_methods
 
         # So in all cases we must behave as if the method was just defined.
-        method = begin
+        method = if self.class.public_method_defined?(name)
           self.class.public_instance_method(name)
-        rescue NameError
-          nil
         end
 
         # The method might be explicitly defined in the model, but call a generated
--- a/activerecord/lib/active_record/attribute_methods.rb
+++ b/activerecord/lib/active_record/attribute_methods.rb
@@ -478,10 +478,12 @@ module ActiveRecord
         self.class.define_attribute_methods
 
         # So in all cases we must behave as if the method was just defined.
-        method = begin
-          self.class.public_instance_method(name)
-        rescue NameError
-          nil
+        method = if self.class.public_method_defined?(name)
+          begin
+            self.class.public_instance_method(name)
+          rescue NameError
+            nil
+          end
         end
 
         # The method might be explicitly defined in the model, but call a generated
--- a/activerecord/test/cases/attribute_methods_test.rb
+++ b/activerecord/test/cases/attribute_methods_test.rb
@@ -1311,6 +1311,20 @@ class AttributeMethodsTest < ActiveRecord::TestCase
     end
   end
 
+  test "virtual SELECT alias attributes are accessible via method_missing" do
+    topic = Topic.select("topics.*, 42 AS virtual_counter").first
+    assert_equal 42, topic.virtual_counter
+  end
+
+  test "virtual SELECT alias attributes do not raise NameError internally" do
+    topic = Topic.select("topics.*, 42 AS virtual_counter").first
+    topic.class.define_attribute_methods
+
+    assert_nothing_raised do
+      topic.virtual_counter
+    end
+  end
+
   test "attribute_method?" do
     assert @target.attribute_method?(:title)
     assert @target.attribute_method?(:title=)


hmcguire-shopify:
More thorough benchmark:

```ruby
require "benchmark/ips"
require "active_record"

STAGE = ENV.fetch("STAGE", "after")

ActiveRecord::Base.lease_connection.create_table(:posts) { |t| t.string :title }

row   = Post.create!(title: "x")
extra = Post.select("id, title, 10 as tenderlove").first

CASES = {
  "title"      => -> { row.title },        # attribute method
  "tenderlove" => -> { extra.tenderlove }, # dynamic attribute
}

CASES.each do |label, block|
  puts
  Benchmark.ips do |x|
    x.config(time: 3, warmup: 1)
    x.save!("/tmp/bench_#{label}.hold")
    x.report("#{label} (#{STAGE})") { block.call }
    x.compare!
  end
end
```

```
$ STAGE=after bx ruby --yjit activemodel/bench_attribute_method_patterns.rb

ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +YJIT +PRISM [arm64-darwin23]
       title (after)     1.900M i/100ms
       title (after)     20.222M (± 1.7%) i/s   (49.45 ns/i) -     60.804M in   3.006844s

Comparison:
 title (after): 20221707.5 i/s
title (before): 20074094.1 i/s - same-ish: difference falls within error


ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +YJIT +PRISM [arm64-darwin23]
  tenderlove (after)   128.383k i/100ms
  tenderlove (after)      1.308M (± 1.5%) i/s  (764.56 ns/i) -      3.980M in   3.042859s

Comparison:
 tenderlove (after):  1307938.7 i/s
tenderlove (before):   733007.2 i/s - 1.78x  slower
```

This theoretically could slow down the _first_ caller of a column _if_ attribute methods aren't defined, but once defined this only improves the performance of virtual columns 👍

PR #57227, opened 2026-04-22
Fix duplicate entries accumulating in `aliases_by_attribute_name`

### LINKED ISSUE #57235: Reassigning table_name on a model causes find/first/where.to_a to get progressively slower — O(N²) cost (7.1 - 8.1+)
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "activerecord", "~> 8.1"
  gem "sqlite3"
  gem "benchmark"
end

require "active_record"
require "benchmark"

  create_table(:users) { |t| t.string :name }
  create_table(:posts) { |t| t.string :title }
end

ActiveRecord::Base.connection.execute("INSERT INTO users (name) VALUES ('x')")
ActiveRecord::Base.connection.execute("INSERT INTO posts (title) VALUES ('y')")

klass = Class.new(ActiveRecord::Base)

8.times do |i|
  t = Benchmark.realtime do
    50.times do
      klass.table_name = "users"
      klass.first
      klass.table_name = "posts"
      klass.first
    end
  end
  size = klass.send(:aliases_by_attribute_name)["id"]&.size || 0
  puts format("iteration %d: %7.1fms, aliases_by_attribute_name['id'].size=%d", i, t * 1000, size)
end
```

The loop does only four things per inner iteration: reassign `table_name`, query, reassign `table_name` to the other table, query.

### Expected

Each iteration takes roughly the same time. `aliases_by_attribute_name["id"]` contains a single entry (`"id_value"`) regardless of how many times `table_name` has been reassigned.

### Actual

Per-iteration wall time grows linearly — total cost across N iterations is quadratic. `aliases_by_attribute_name["id"]` grows unbounded by one entry per `table_name=` followed by any record-instantiating query (`find`, `first`, `last`, `where(...).to_a`, etc.).

Output on activerecord 8.1.3 / ruby 4.0.2:

```
iteration 0:   568.8ms, aliases_by_attribute_name['id'].size=100
iteration 1:  1541.9ms, aliases_by_attribute_name['id'].size=200
iteration 2:  2549.8ms, aliases_by_attribute_name['id'].size=300
iteration 3:  3623.2ms, aliases_by_attribute_name['id'].size=400
iteration 4:  4619.1ms, aliases_by_attribute_name['id'].size=500
iteration 5:  5458.5ms, aliases_by_attribute_name['id'].size=600
iteration 6:  6496.9ms, aliases_by_attribute_name['id'].size=700
iteration 7:  7408.3ms, aliases_by_attribute_name['id'].size=800
```

Per-iteration time grew ~13× over 8 iterations of identical work. No stabilization — the slope continues indefinitely.

### Root cause

`ActiveRecord::AttributeMethods::ClassMethods#define_attribute_methods` (introduced in 7.1) registers `alias_attribute :id_value, :id` as part of attribute-method generation. Reassigning `table_name` invalidates the generated methods, so the next record instantiation re-runs `define_attribute_methods`, which re-registers the alias.

`ActiveModel::AttributeMethods::ClassMethods#alias_attribute` appends to `aliases_by_attribute_name[old_name]` without checking for duplicates:

```ruby
aliases_by_attribute_name[old_name] << new_name
```

So every `table_name=` followed by a record-instantiating query appends another `"id_value"` to `aliases_by_attribute_name["id"]`. Subsequent attribute-method generation iterates this array, producing the quadratic growth observed above.

### Scope

The accumulation fires on any query that instantiates an `ActiveRecord::Base` record: `find`, `first`, `last`, `where(...).to_a`, `find_each`, etc. It does **not** fire on aggregate or raw-value queries — `count`, `exists?`, `pluck`, or `where(...)` alone (which returns a lazy relation). So code that only touches these escapes the regression.

### Real-world impact

Encountered while investigating a performance regression in [Stagehand](https://github.com/combinaut/stagehand), a staging-to-production data-sync library. Its `Stagehand::Production::Record` class reassigns `table_name` per synced record to read from and delete from many different tables through a single AR class. After the 7.1 upgrade, per-iteration wall time in a sync of 3000 entries grew from ~22ms (first batch of 1000) to ~210ms (third batch of 1000) — CPU-bound, not I/O-bound. Total runtime: ~3.5 minutes instead of seconds. Any code that swaps `table_name` on a persistent class between queries will exhibit the same shape.

### System configuration

Rails version: 8.1.3 (reproducible on 7.1+)
Ruby version: 4.0.2

### Upstream fix

#57227 addresses this by deduplicating in `alias_attribute`. Filing this issue to document the regression and attach real-world impact, so it's discoverable by anyone hitting the same symptom before the fix lands.

njakobsen:
Fixes #57235

### Motivation / Background

`ActiveModel::AttributeMethods#alias_attribute` appends to `aliases_by_attribute_name[old_name]` without checking for duplicates. When the same alias is registered more than once on the same class, the underlying storage accumulates duplicate entries, and `define_attribute_methods` later iterates every duplicate — generating the same aliased attribute methods once per duplicate.

This is a regression for any code that reassigns `self.table_name = ...` on a reusable AR class. Since Rails 7.1, `ActiveRecord::Base#load_schema!` auto-calls `alias_attribute :id_value, :id` for every `id` column on every schema load ([`activerecord/lib/active_record/model_schema.rb#L628`](https://github.com/rails/rails/blob/d39db5d1891f7509cde2efc425c9d69bbb77e670/activerecord/lib/active_record/model_schema.rb#L628), introduced in [rails/rails@e90b11e77e](https://github.com/rails/rails/commit/e90b11e77e)). Combined with the persistent `aliases_by_attribute_name` structure ([`activemodel/lib/active_model/attribute_methods.rb#L204-L209`](https://github.com/rails/rails/blob/d39db5d1891f7509cde2efc425c9d69bbb77e670/activemodel/lib/active_model/attribute_methods.rb#L204-L209), introduced in [rails/rails@0f5563bd40](https://github.com/rails/rails/commit/0f5563bd40)), repeatedly setting `table_name` appends a new `"id_value"` entry on every schema reload. After N reassignments, `aliases_by_attribute_name["id"]` contains N duplicate `"id_value"` strings, and each `define_attribute_methods` call re-generates aliased accessor methods N times per pattern per column.

Rails 7.0 and earlier did not have either the auto-alias or the persistent storage, so this pattern was effectively free before 7.1.

Compare 7.0 vs 7.1 `alias_attribute`:
- 7.0.10: [`activemodel/lib/active_model/attribute_methods.rb#L204-L235`](https://github.com/rails/rails/blob/f4321f78fe567fd29dd0a0fb5cb17af68efd36bb/activemodel/lib/active_model/attribute_methods.rb#L204-L235) — defines methods inline via `attribute_method_matchers.each`, no `aliases_by_attribute_name`.
- 7.1.0: [`activemodel/lib/active_model/attribute_methods.rb#L204-L209`](https://github.com/rails/rails/blob/d39db5d1891f7509cde2efc425c9d69bbb77e670/activemodel/lib/active_model/attribute_methods.rb#L204-L209) — appends to `aliases_by_attribute_name` (L208), delegates generation to `define_attribute_methods`.

In a representative reproduction, per-query cost grew monotonically from ~12 ms to ~90 ms over ~1000 `table_name=` swaps on a single class, with per-call allocations climbing from ~41 k to ~246 k objects — all attributable to repeated `define_attribute_method_pattern` work firing exponentially more times as duplicates accumulated.

### Detail

Dedupe the append in `alias_attribute`:

```ruby
aliases = aliases_by_attribute_name[old_name]
aliases << new_name unless aliases.include?(new_name)
```

The rest of the method is unchanged. `attribute_aliases.merge(new_name => old_name)` is already idempotent, and `eagerly_generate_alias_attribute_methods` is safe to re-call because `define_attribute_method_pattern` early-returns on `instance_method_already_implemented?`.

### Additional information

Two tests are added:

1. An ActiveModel unit test (`activemodel/test/cases/attribute_methods_test.rb`) calling `alias_attribute :bar, :foo` three times on the same class and asserting `aliases_by_attribute_name["foo"] == ["bar"]`.
2. An ActiveRecord integration test (`activerecord/test/cases/attribute_methods_test.rb`) that mirrors the real-world regression path — an anonymous `ActiveRecord::Base` subclass with `table_name` reassigned between `topics` and `posts`, asserting `aliases_by_attribute_name["id"] == ["id_value"]` after three `.first` queries.

Both tests fail on unpatched `main` (clean diffs: `["bar"] expected, ["bar", "bar", "bar"] actual` and `["id_value"]` expected, `["id_value", "id_value"]` actual) and pass with the patch applied. Full `activemodel` suite (1168 tests) and full `activerecord` sqlite3 suite (9418 tests) pass with zero new failures.

### Checklist

* [x] Tests are added for the bug fix.

### DIFF
--- a/activemodel/lib/active_model/attribute_methods.rb
+++ b/activemodel/lib/active_model/attribute_methods.rb
@@ -204,7 +204,7 @@ module ActiveModel
         old_name = old_name.to_s
         new_name = new_name.to_s
         self.attribute_aliases = attribute_aliases.merge(new_name => old_name)
-        aliases_by_attribute_name[old_name] << new_name
+        aliases_by_attribute_name[old_name] |= [new_name]
         eagerly_generate_alias_attribute_methods(new_name, old_name)
       end
 
--- a/activemodel/test/cases/attribute_methods_test.rb
+++ b/activemodel/test/cases/attribute_methods_test.rb
@@ -206,6 +206,18 @@ class AttributeMethodsTest < ActiveModel::TestCase
     assert_equal({ "bar" => "foo" }, klass.attribute_aliases)
   end
 
+  test "#alias_attribute is idempotent when called multiple times with the same arguments" do
+    klass = Class.new(ModelWithAttributes) do
+      define_attribute_methods :foo
+      alias_attribute :bar, :foo
+      alias_attribute :bar, :foo
+      alias_attribute :bar, :foo
+    end
+
+    assert_equal({ "bar" => "foo" }, klass.attribute_aliases)
+    assert_equal ["bar"], klass.send(:aliases_by_attribute_name)["foo"]
+  end
+
   test "#define_attribute_methods generates attribute methods with spaces in their names" do
     ModelWithAttributesWithSpaces.define_attribute_methods(:'foo bar')
 
--- a/activerecord/test/cases/attribute_methods_test.rb
+++ b/activerecord/test/cases/attribute_methods_test.rb
@@ -55,6 +55,21 @@ class AttributeMethodsTest < ActiveRecord::TestCase
     assert_includes new_topic_model.attribute_aliases, "id_value"
   end
 
+  test "#id_value alias does not accumulate duplicates when table_name is reassigned" do
+    klass = Class.new(ActiveRecord::Base)
+
+    klass.table_name = "topics"
+    klass.first
+
+    klass.table_name = "authors"
+    klass.first
+
+    klass.table_name = "topics"
+    klass.first
+
+    assert_equal ["id_value"], klass.send(:aliases_by_attribute_name)["id"]
+  end
+
   test "#id_value alias is not defined if id_value column exist" do
     new_book_identifier_model = Class.new(ActiveRecord::Base) do
       self.table_name = "book_identifiers"


njakobsen:
Thanks @fatkodima!

PR #57229, opened 2026-04-23
Fix attachment uploads lost when reload is called inside a transaction

### LINKED ISSUE #57222: ActiveStorage uploads don't trigger in a transaction if the model is reloaded
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.root = __dir__
  config.hosts << "example.org"
  config.eager_load = false

  Rails.logger  = config.logger

    local: {
      root: Dir.tmpdir,
      service: "Disk"
    }
  }

end
Rails.application.initialize!

end

  has_one_attached :profile
end

  def test_upload_and_download
    user = ActiveRecord::Base.transaction do
      new_user = User.create!(
        profile: {
          content_type: "text/plain",
          filename: "dummy.txt",
          io: ::StringIO.new("dummy"),
        }
      )
      new_user.reload
    end

    assert_equal "dummy", user.profile.download
  end
end

```

### Expected behavior

The test should pass.

### Actual behavior

The test fails with `ActiveStorage::FileNotFoundError: ActiveStorage::FileNotFoundError` which is triggered by the download method on the attachment caused by the fact that the blob was never uploaded to the storage service.

Calling `reload` makes the owner of the attachment forget the pending attachments, which in turn makes the `after_commit` hook in ActiveStorage skip `upload`.

From `active_storage/attached/model.rb`

```
    after_commit(on: %i[ create update ]) { attachment_changes.delete(name.to_s).try(:upload) }

    def reload(*) # :nodoc:
      super.tap { @attachment_changes = nil }
    end
```

Why would anyone call reload there? It could happen for whatever reason, in our case it was a state transition via the `aasm` gem, which happens to reload the model.

The transaction is a crucial piece of the picture as blob uploads are delayed until the transaction is done.

### System configuration

Rails: 8.1.3, 7.2.3.1

**Ruby version**: 

3.4.6

hammadxcm:
### Motivation / Background

Fixes #57222.

When a model with an attachment is saved inside a transaction and then reloaded
before the transaction commits, the blob is never uploaded to the storage
service. This happens because `reload` clears `@attachment_changes`, so when
the `after_commit` callback fires, it finds nothing to upload.

### Detail

The callback chain for attachments is:

1. `after_save` — persists the attachment and blob records to the database
2. `after_commit` — uploads the file to the storage service via `attachment_changes.delete(name).try(:upload)`

When `reload` is called between steps 1 and 2 (which happens inside a
transaction since `after_commit` is deferred), `@attachment_changes = nil`
wipes the pending upload info.

This PR changes `reload` to preserve `CreateOne`/`CreateMany` changes whose
attachments have already been persisted (i.e. `after_save` already ran).
Unsaved attachment changes are still cleared on reload, matching existing
behavior — the "clearing change on reload" test continues to pass.

### Additional information

Reproduction script from the issue confirms the fix:

```ruby
user = ActiveRecord::Base.transaction do
  new_user = User.create!(
    profile: { content_type: "text/plain", filename: "dummy.txt", io: StringIO.new("dummy") }
  )
  new_user.reload
end

user.profile.download # => "dummy" (was raising ActiveStorage::FileNotFoundError)
```

### Checklist

- [x] Tests are added for both `has_one_attached` and `has_many_attached`.
- [x] CHANGELOG entry added for ActiveStorage.
- [x] RuboCop passes with no offenses.

### DIFF
--- a/activestorage/lib/active_storage/attached/model.rb
+++ b/activestorage/lib/active_storage/attached/model.rb
@@ -318,7 +318,19 @@ module ActiveStorage
     end
 
     def reload(*) # :nodoc:
-      super.tap { @attachment_changes = nil }
+      super.tap do
+        if @attachment_changes
+          @attachment_changes.select! do |_, change|
+            case change
+            when ActiveStorage::Attached::Changes::CreateOne
+              change.attachment.persisted?
+            when ActiveStorage::Attached::Changes::CreateMany
+              change.attachments.any?(&:persisted?)
+            end
+          end
+          @attachment_changes = nil if @attachment_changes.empty?
+        end
+      end
     end
   end
 end
--- a/activestorage/lib/active_storage/attached/model.rb
+++ b/activestorage/lib/active_storage/attached/model.rb
@@ -319,17 +319,17 @@ module ActiveStorage
 
     def reload(*) # :nodoc:
       super.tap do
-        if @attachment_changes
-          @attachment_changes.select! do |_, change|
-            case change
-            when ActiveStorage::Attached::Changes::CreateOne
-              change.attachment.persisted?
-            when ActiveStorage::Attached::Changes::CreateMany
-              change.attachments.any?(&:persisted?)
-            end
+        next unless @attachment_changes
+
+        @attachment_changes.select! do |_, change|
+          case change
+          when ActiveStorage::Attached::Changes::CreateOne
+            change.attachment.persisted?
+          when ActiveStorage::Attached::Changes::CreateMany
+            change.attachments.any?(&:persisted?)
           end
-          @attachment_changes = nil if @attachment_changes.empty?
         end
+        @attachment_changes = nil if @attachment_changes.empty?
       end
     end
   end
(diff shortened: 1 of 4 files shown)


activestorage/lib/active_storage/attached/model.rb
```diff
@@ -318,7 +318,19 @@ def initialize_dup(*) # :nodoc:
     end
 
     def reload(*) # :nodoc:
-      super.tap { @attachment_changes = nil }
+      super.tap do
+        if @attachment_changes
```
tahsin352: We can use 
```next unless @attachment_changes```
instead. It will flatten the code block and exists the block early.

hammadxcm: Good catch, thanks! Pushed b38f0dea8f flattening with `next unless @attachment_changes`.

PR #57234, opened 2026-04-23
Upstream ActionView::Precompiler

joelhawksley:
### Motivation / Background

This PR upstreams the the `actionview_precompiler` gem by @jhawthorn that GitHub has used for over 5 years, saving an estimated ~500MB of memory per container (each with 11 forked workers) for ~7,000 templates.

### Details

View templates can now be eagerly compiled at boot time when `config.action_view.precompile_templates = true` (enabled by default in `load_defaults "8.2"`) and `config.eager_load` is also `true`. This improves cold render times and allows more memory to be shared via copy-on-write on forking web servers. 

The precompiler scans view templates, controllers, and helpers for `render` calls, detects implicit controller action renders, and supports engine view paths. Invalid templates raise errors at boot time rather than failing silently.

Additional directories can be scanned for `render` calls using `config.action_view.precompile_additional_paths`.

Emits a `precompile_templates.action_view` notification with `:count` in the payload.

### Checklist


guides/source/configuring.md
```diff
@@ -2667,6 +2668,21 @@ The default value depends on the `config.load_defaults` target version:
...
+The default value depends on the `config.load_defaults` target version:
+
+| Starting with version | The default value is |
+| --------------------- | -------------------- |
+| (original)            | `false`              |
+| 8.2                   | `true`               |
+
+#### `config.action_view.precompile_additional_paths`
+
+An array of additional directories to scan for `render` calls when precompiling templates. Useful for applications with controllers or helpers outside of the standard engine paths, such as those using Packwerk packs or ViewComponent sidecar templates. Defaults to `[]`.
```
rafaelfranca:
```suggestion
An array of additional directories to scan for `render` calls when precompiling templates. Useful for applications with controllers or helpers outside of the standard engine paths. Defaults to `[]`.
```

actionview/CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+*   Add template precompilation support.
+
+    View templates can now be eagerly compiled at boot time when
+    `config.action_view.precompile_templates = true` (enabled by default in
+    `load_defaults "8.2"`) and `config.eager_load` is also `true`. This
+    improves cold render times and allows more memory to be shared via
+    copy-on-write on forking web servers. With the potential upstreaming
+    of Herb, this optimization becomes even more important due to the
+    likely increased memory and latency overhead of Herb.
```
rafaelfranca:
```suggestion
    copy-on-write on forking web servers.
```

actionview/lib/action_view/railtie.rb
```diff
@@ -82,11 +82,15 @@ class Railtie < Rails::Engine # :nodoc:
     config.after_initialize do |app|
       config.after_initialize do
         ActionView.render_tracker = config.action_view.render_tracker
+        ActionView.precompile_templates = config.action_view.precompile_templates if config.action_view.key?(:precompile_templates)
+        ActionView.precompile_additional_paths = config.action_view.precompile_additional_paths if config.action_view.key?(:precompile_additional_paths)
```
rafaelfranca: This should not be inside a nested `config.after_initialize`

actionview/lib/action_view/precompiler.rb
```diff
@@ -0,0 +1,168 @@
...
+        action = match[:action]
+        prefix = match[:prefix] ? [match[:prefix]] : []
+        partial = !!match[:partial]
+
+        lookup_context.find_all(action, prefix, partial, locals)
+      rescue ActionView::MissingTemplate
+        []
+      end
+
+      class << self
```
rafaelfranca: Class methods should be defined before any instance method. Also it looks like those methods are private but they aren't, so one more reason to follow the convention.

actionview/lib/action_view/precompiler.rb
```diff
@@ -0,0 +1,168 @@
...
+            add_implicit_action_templates(precompiler)
+
+            payload[:count] = precompiler.run
+          end
+        end
+
+        private
+          def each_engine(&block)
+            if defined?(Rails) && Rails.respond_to?(:application) && Rails.application
+              yield Rails.application
```
rafaelfranca: This should be in the Railtie, not here with all those `defined?` checks.

actionview/lib/action_view/precompiler.rb
```diff
@@ -0,0 +1,168 @@
...
+              Rails::Engine.subclasses.each do |engine_class|
+                yield engine_class.instance if engine_class.respond_to?(:instance)
+              end
+            end
+          end
+
+          def add_implicit_action_templates(precompiler)
+            return unless defined?(ActionController::Base)
```
rafaelfranca: This also should not be here. Action view should not know about `ActionController` in its classes. You can do in the railtie.

actionview/lib/action_view/template_scanner.rb
```diff
@@ -0,0 +1,73 @@
...
+    private
+      def each_template
+        Dir["**/*", base: @view_dir].sort.each do |file|
+          fullpath = File.expand_path(file, @view_dir)
+          next if File.directory?(fullpath)
+          yield file, fullpath
+        end
+      end
+
+      class FakeTemplate # :nodoc:
```
rafaelfranca: Same about class definitions being before any instance method. It looks like this class is private but it isn't

rafaelfranca (changes requested):

actionview/lib/action_view/railtie.rb
```diff
@@ -80,11 +80,9 @@ class Railtie < Rails::Engine # :nodoc:
     end
 
     config.after_initialize do |app|
-      config.after_initialize do
-        ActionView.render_tracker = config.action_view.render_tracker
-        ActionView.precompile_templates = config.action_view.precompile_templates if config.action_view.key?(:precompile_templates)
-        ActionView.precompile_additional_paths = config.action_view.precompile_additional_paths if config.action_view.key?(:precompile_additional_paths)
-      end
+      ActionView.render_tracker = app.config.action_view.render_tracker
```
joelhawksley: I went ahead and moved this line out of the redundant `config.initialize`, FWIW

actionview/CHANGELOG.md
```diff
@@ -1,22 +1,26 @@
-*   Render `Hash` and keyword options as dasherized HTML attributes
+*   Add template precompilation support.
 
-    ```ruby
-    tag.button "POST to /clicked", hx: { post: "/clicked", swap: :outerHTML, data: { json: true } }
+    View templates can now be eagerly compiled at boot time when
+    `config.action_view.precompile_templates = true` (enabled by default in
+    `load_defaults "8.2"`) and `config.eager_load` is also `true`. This
+    improves cold render times and allows more memory to be shared via
+    copy-on-write on forking web servers.
```
joelhawksley, line 193: I removed the note about catching template errors at boot as the precompiler does not guarantee to precompile all templates.

joelhawksley (commented):
@rafaelfranca thank you for taking the time to review this proposed change. I've addressed your comments, leaving my changes in separate commits for easier review. Let me know your thoughts, happy to squash to a single commit whenever you prefer ❤️

actionview/lib/action_view/render_parser.rb
```diff
@@ -6,47 +6,73 @@ module ActionView
   class RenderParser # :nodoc:
     ALL_KNOWN_KEYS = [:partial, :template, :layout, :formats, :locals, :object, :collection, :as, :status, :content_type, :location, :spacer_template].freeze
     RENDER_TYPE_KEYS = [:partial, :template, :layout].freeze
+    RENDER_CALL_METHODS = /\A(render|render_to_string)\z/
+    LAYOUT_CALL_METHOD = /\Alayout\z/ main
```
joelhawksley:
```suggestion
    LAYOUT_CALL_METHOD = /\Alayout\z/
```

joelhawksley:
@rafaelfranca just rebased ❤️

joelhawksley:
@rafaelfranca just a friendly nudge here ❤️ - I'd love to land this in 8.2!

joelhawksley:
👋🏻 turns on the CI failures were not flakes like I thought. They've been fixed ❤️

joelhawksley:
Moving to draft while I look at incorporating https://github.com/rails/rails/pull/58532/ here

joelhawksley:
@jhawthorn thanks for the tip on `eager_load_templates`. I was able to close https://github.com/rails/rails/pull/58532/ and limit the scope of changes to be focused on precompilation ❤️

joelhawksley:
@rafaelfranca @jhawthorn @marcoroth I think we'd be wise to consider landing this PR in 8.2 given the increased compilation overhead of Herb. 

In GitHub's benchmarks, we're seeing a ~7x increase in compile time compared to Erubi, which impacts cold render times without this change.

In world where more apps are using Herb, this PR is _more_ useful. What do you think?

marcoroth:
@joelhawksley great progress on this so far! One thing I'm wondering about is that this front-loads the compile work, but it still pays it on every boot. Would moving this kind of work down to Bootsnap make sense, so it's cached across boots? @byroot and I had a previous discussion about this, have you explored anything in that area yet?

joelhawksley:
@marcoroth I have given the Bootsnap idea some thought too, but wanted to stick to an approach that has been in production for a long time. I think we could add a Bootsnap optimization after this approach lands.

PR #57236, opened 2026-04-23
Fix CONTENT_TYPE not cleared between ActionController::TestCase requests

### LINKED ISSUE #55759: CONTENT_TYPE header is not cleared between ActionController::TestCase requests

jackie20:
When a POST request set CONTENT_TYPE (e.g. application/x-www-form-urlencoded), a subsequent GET request in the same test inherited that header because scrub_env! cleaned CONTENT_LENGTH and RAW_POST_DATA but left CONTENT_TYPE intact. Rack uses CONTENT_TYPE to decide whether to parse a body, so a stale value from a prior POST caused spurious form-data parsing on GET requests.

The fix adds `env.delete "CONTENT_TYPE"` to scrub_env!. The existing logic at line 115 already re-sets CONTENT_TYPE via fetch_header only when the next request actually needs a body, so deleting it here is safe for all methods.

Fixes #55759

### Motivation / Background

### Detail

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/test_case.rb
+++ b/actionpack/lib/action_controller/test_case.rb
@@ -683,6 +683,7 @@ module ActionController
           end
           env["rack.input"] = StringIO.new
           env.delete "CONTENT_LENGTH"
+          env.delete "CONTENT_TYPE"
           env.delete "RAW_POST_DATA"
           env
         end
--- a/actionpack/test/controller/test_case_test.rb
+++ b/actionpack/test/controller/test_case_test.rb
@@ -793,6 +793,14 @@ class TestCaseTest < ActionController::TestCase
     assert_equal 0, @request.content_length
   end
 
+  def test_content_type_reset_after_post_request
+    post :no_op, params: { foo: "bar" }
+    assert_equal "application/x-www-form-urlencoded", @request.content_type
+
+    get :no_op
+    assert_nil @request.content_type
+  end
+
   def test_path_is_kept_after_the_request
     get :test_params, params: { id: "foo" }
     assert_equal "/test_case_test/test/test_params/foo", @request.path


rania33840-netizen:
jackie20:fix/test-case-content-type-header-not-cleared-between-requests

rania33840-netizen:
jackie20:fix/test-case-content-type-header-not-cleared-between-requests

PR #57239, opened 2026-04-24
Fix #57231: Clear PostgreSQL schema search path cache on rollback

### LINKED ISSUE #57231: PostgreSQL's schema_search_path cache gets stale after transaction rollback
### Steps to reproduce

Create a local `search_path_bug` PostgreSQL database, then:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "pg"
end

ENV["DATABASE_URL"] = "postgresql://localhost:5432/search_path_bug"

  config.eager_load = false
  config.logger = Logger.new($stderr)

end
Rails.application.initialize!

  execute('DROP SCHEMA IF EXISTS a CASCADE')
  execute('DROP SCHEMA IF EXISTS b CASCADE')
  execute('DROP TABLE IF EXISTS public.posts')

  execute('CREATE SCHEMA a')
  execute('CREATE SCHEMA b')

  execute('CREATE TABLE public.posts (id integer NOT NULL PRIMARY KEY)')
  execute('CREATE TABLE a.posts (id integer NOT NULL PRIMARY KEY)')
  execute('CREATE TABLE b.posts (id integer NOT NULL PRIMARY KEY)')

  execute('INSERT INTO a.posts (id) VALUES (1), (2), (3)')
  execute('INSERT INTO b.posts (id) VALUES (1)')
end

end

def list_counts
  Post.connection.schema_search_path = 'a'
  puts "a.posts: #{Post.count}"
  Post.connection.schema_search_path = 'b'
  puts "b.posts: #{Post.count}"
end

puts '== 1'
list_counts

Post.transaction do
  Post.connection.schema_search_path = 'a'
  raise ActiveRecord::Rollback
end

puts '== 2'
list_counts

puts '== 3'
list_counts
```

### Expected behavior

The above program lists the post count of both schemas 3 times. Between the first and second listings a transaction rolls back after having set the schema_search_path. Nevertheless, all 3 listings should output the same counts.

### Actual behavior

```
$ ruby search_path_bug.rb 2> /dev/null
(...)
== 1
a.posts: 3
b.posts: 1
== 2
a.posts: 1      <== should be: a.posts: 3
b.posts: 1
== 3
a.posts: 3
b.posts: 1
```

Before the transaction starts schema_search_path is set to `b` and the cached value in the connection adapter correctly reflects the actual Postgres schema.

The transaction starts, the program changes the schema_search_path to `a` and the transaction is then rolled back.

At this point the connection adapter's schema_search_path cached value points to `a`, but Postgres has reset it to `b`, i.e the value set when the transaction started.

Next time `Post.connection.schema_search_path = 'a'` is called the connection adapter no-ops it because it's equal to the cached value.

When the query is requested Postgres performs it on schema `b`. Hence the the count 1 rather than 3.

### System configuration

Rails version: 8.1.3 (also reproducible on edge)

Ruby version: ruby 4.0.3 (2026-04-21 revision 85ddef263a) +PRISM [x86_64-linux]

afurm:
### Motivation / Background
This Pull Request has been created because PostgreSQL rolls back `SET search_path` changes when a transaction or savepoint rolls back, but Active Record kept the previously assigned `schema_search_path` cached on the adapter. That could make a later `schema_search_path = ...` call no-op against stale adapter state while PostgreSQL was still using a different search path, causing subsequent unqualified queries to run against the wrong schema.

Fixes #57231.

### Detail

This Pull Request changes the PostgreSQL adapter to clear its cached
`@schema_search_path` after:

* full transaction rollback
* transaction restart via `ROLLBACK AND CHAIN`
* savepoint rollback

With the cache cleared, the next `schema_search_path = ...` assignment compares
against the actual server state and reapplies `SET search_path` when needed.

Regression tests cover both top-level transaction rollback and nested
savepoint rollback.

### Additional information

Validated with:

```bash
ARCONN=postgresql bin/test test/cases/adapters/postgresql/postgresql_adapter_test.rb -i '/schema_search_path_cache_is_cleared_after_(rollback|savepoint_rollback)/'
ARCONN=postgresql bin/test test/cases/adapters/postgresql/postgresql_adapter_test.rb -i '/schema_search_path/'
ARCONN=postgresql bin/test test/cases/adapters/postgresql/postgresql_adapter_test.rb
ARCONN=postgresql bin/test test/cases/adapters/postgresql/transaction_test.rb test/cases/adapters/postgresql/transaction_nested_test.rb
bundle exec rubocop activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
```

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -82,11 +82,21 @@ module ActiveRecord
         def exec_rollback_db_transaction # :nodoc:
           cancel_any_running_query
           query_command("ROLLBACK", "TRANSACTION", allow_retry: false, materialize_transactions: true)
+        ensure
+          @schema_search_path = nil
         end
 
         def exec_restart_db_transaction # :nodoc:
           cancel_any_running_query
           query_command("ROLLBACK AND CHAIN", "TRANSACTION", allow_retry: false, materialize_transactions: true)
+        ensure
+          @schema_search_path = nil
+        end
+
+        def exec_rollback_to_savepoint(name = current_savepoint_name) # :nodoc:
+          query_command("ROLLBACK TO SAVEPOINT #{name}", "TRANSACTION")
+        ensure
+          @schema_search_path = nil
         end
 
         # From https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -320,6 +320,46 @@ module ActiveRecord
         connection&.disconnect!
       end
 
+      def test_schema_search_path_cache_is_cleared_after_rollback
+        old_search_path = @connection.schema_search_path
+        @connection.schema_search_path = "public"
+
+        @connection.transaction do
+          @connection.schema_search_path = "pg_catalog"
+          raise ActiveRecord::Rollback
+        end
+
+        assert_equal "public", @connection.select_value("SHOW search_path")
+
+        @connection.schema_search_path = "pg_catalog"
+        assert_equal "pg_catalog", @connection.select_value("SHOW search_path")
+      ensure
+        @connection.schema_search_path = old_search_path
+      end
+
+      def test_schema_search_path_cache_is_cleared_after_savepoint_rollback
+        old_search_path = @connection.schema_search_path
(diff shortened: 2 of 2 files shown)


yahonda:
Thanks for the PR and the investigation.

Closing this for the reason described in https://github.com/rails/rails/issues/57231#issuecomment-4437740246 — in short, the scenario falls outside the documented usage of schema_search_path=, so the path supported via database.yml is unaffected.

Appreciate the effort here. Please don't hesitate to open further PRs.

PR #57240, opened 2026-04-24
Move generated kamal gem into development group

### LINKED ISSUE #57205: Move generated kamal gem into the development group
### Steps to reproduce

```ruby
  require "fileutils"
  require "tmpdir"

  Dir.mktmpdir do |dir|
    app_path = File.join(dir, "kamal_group_test")

    system("rails new #{app_path} --skip-bundle", exception: true)

    gemfile = File.read(File.join(app_path, "Gemfile"))

    puts gemfile.lines.grep(/kamal/)
    puts
    puts "Inside development group? #{gemfile.match?(/group :development do.*gem \"kamal\", require: false/m)}"
  end


```

### Expected behavior

  The generated Gemfile should place kamal inside group :development.

  kamal is deploy tooling rather than an application runtime dependency, so generating it as a development-only gem
  seems like a better default. That keeps the default bundle smaller and avoids including kamal in production installs
  and container images unless the app explicitly wants that.

  Proposed generated output:
```
  group :development do
    gem "kamal", require: false

    # Use console on exceptions pages [https://github.com/rails/web-console]
    gem "web-console"
  end
```

### Actual behavior

  Rails currently generates kamal at the top level:

 ` gem "kamal", require: false`

  That makes it part of the default bundle, including production bundles, even though the running application does not
  use it.

  I may be missing an intentional reason for this. The only rationale I can think of is preserving bundle exec kamal
  availability in environments where the development group is excluded. But if that is not a requirement,
  group :development seems like the cleaner default.

### System configuration

**Rails version**:  8.1

**Ruby version**: 4.0.1

jaykava:
### Motivation / Background

This Pull Request has been created because `rails new` currently generates `gem "kamal", require: false` as a top-level dependency in the app `Gemfile`, which includes Kamal in production bundles by default even though it is deploy tooling.

Fixes #57205.

### Detail

This Pull Request changes the app Gemfile template so generated `kamal` is placed under `group :development` instead of top-level.

It also updates generator tests to verify the new behavior and guard existing option behavior:

- regular app generation includes Kamal in `group :development`
- API app generation still has expected Gemfile output with Kamal grouped in development
- `--skip-dev-gems` omits Kamal
- `--skip-kamal` still omits Kamal

Changed files:

- `railties/lib/rails/generators/rails/app/templates/Gemfile.tt`
- `railties/test/generators/app_generator_test.rb`
- `railties/test/generators/api_app_generator_test.rb`

### Additional information

Verified with focused generator test runs:

- `bundle exec ruby -Irailties/test railties/test/generators/app_generator_test.rb -n "/kamal|skip_dev_gems/"`
- `bundle exec ruby -Irailties/test railties/test/generators/api_app_generator_test.rb -n "/test_api_modified_files|test_kamal_deploy_yml_excludes_asset_path_for_api_apps/"`

Both passed.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
@@ -29,11 +29,6 @@ gem "solid_cable"
 # Reduces boot times through caching; required in config/boot.rb
 gem "bootsnap", require: false
 <% end -%>
-<% unless options.skip_kamal? -%>
-
-# Deploy this application anywhere as a Docker container [https://kamal-deploy.org]
-gem "kamal", require: false
-<% end -%>
 <% unless options.skip_thruster? -%>
 
 # Add HTTP asset caching/compression and X-Sendfile acceleration to Puma [https://github.com/basecamp/thruster/]
@@ -71,13 +66,22 @@ group :development, :test do
 <%- end -%>
 end
 <% end -%>
-<%- unless options.api? || options.skip_dev_gems? -%>
+<% unless options.skip_dev_gems? -%>
+<% if !options.skip_kamal? || !options.api? -%>
 
 group :development do
+<% unless options.skip_kamal? -%>
+  # Deploy this application anywhere as a Docker container [https://kamal-deploy.org]
+  gem "kamal", require: false
+<% end -%>
+<%- unless options.api? -%>
+
   # Use console on exceptions pages [https://github.com/rails/web-console]
   gem "web-console"
-end
 <%- end -%>
+end
+<% end -%>
+<% end -%>
 <%- if depends_on_system_test? -%>
 
 group :test do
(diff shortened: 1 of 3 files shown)


jaykava:
Hi @fidalgo Could you please retry CI? railties (3.4) timed out pulling ECR image.

railties/lib/rails/generators/rails/app/templates/Gemfile.tt
```diff
@@ -71,6 +66,13 @@ group :development, :test do
 <%- end -%>
 end
 <% end -%>
+<% unless options.skip_kamal? || options.skip_dev_gems? -%>
+group :development do
```
rafaelfranca, line 72: We might end up with two `group :development do` in the way it is written here. We should make sure we don't.

rafaelfranca (changes requested):

railties/lib/rails/generators/rails/app/templates/Gemfile.tt
jaykava: Merge the kamal and web-console gems into a single group :development
block in the Gemfile template to avoid generating duplicate group
declarations.

rafaelfranca (approved):

jaykava:
@rafaelfranca Could you please re-run this failed pipeline.

CI failure in `PostgresqlEnumTest#test_enum_type_cast` is unrelated - PostgreSQL connection pool exhaustion (`FATAL: sorry, too many clients already`). Flaky infra issue.

thanks

PR #57246, opened 2026-04-25
Optimize `ActiveRecord::Relation#extending!` to skip work when called with empty modules

bogdan:
Follow-up to #57199 by @fatkodima, applying the suggestion from the code review.

Instead of guarding at the call site in `merge_multi_values` only, move the early-return guard into `AR::Relation#extending!` itself so all callers benefit universally. The merger retains its own call-site guard since skipping the method call entirely is measurably faster on the hot path (empty extensions).

**Changes:**
- `extending!` returns `self` immediately if no modules and no block are given
- `merge_multi_values` simplified to `relation.extending!(*other.extensions) unless other.extensions.empty?`, removing the manual `a - b` set difference (dedup now handled by `|=` inside `extending!`)

Deduplication is already covered by https://github.com/rails/rails/blob/main/activerecord/test/cases/relation_test.rb#L45-L62

/cc @fatkodima @Hackerbee

### DIFF
--- a/activerecord/lib/active_record/relation/merger.rb
+++ b/activerecord/lib/active_record/relation/merger.rb
@@ -161,10 +161,7 @@ module ActiveRecord
             relation.order!(*other.order_values)
           end
 
-          unless other.extensions.empty?
-            extensions = other.extensions - relation.extensions
-            relation.extending!(*extensions) if extensions.any?
-          end
+          relation.extending!(*other.extensions)
         end
 
         def merge_single_values
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1463,6 +1463,8 @@ module ActiveRecord
     end
 
     def extending!(*modules, &block) # :nodoc:
+      return self if modules.empty? && !block
+
       modules << Module.new(&block) if block
       modules.flatten!
 


fatkodima:
Benchmark for this PR changes
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "benchmark-ips"
end

require "benchmark/ips"

modules = []

def extending(*modules)
end

Benchmark.ips do |x|
  x.report("original") do
    unless modules.empty?
      extending(*modules)
    end
  end

  x.report("PR") do
    extending(*modules)
  end

  x.compare!
end
```

```
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [x86_64-darwin21]
            original     1.479M i/100ms
                  PR   567.630k i/100ms
            original     14.039M (± 7.7%) i/s   (71.23 ns/i) -     71.004M in   5.088547s
                  PR      4.992M (± 8.1%) i/s  (200.32 ns/i) -     24.976M in   5.037796s

Comparison:
            original: 14039021.8 i/s
                  PR:  4991951.2 i/s - 2.81x  slower
```

bogdan:
@fatkodima your benchmark shows that using array spread operator is slow in any context. While it is the case, the requirement to check for `empty?` before using array spread operator in my opinion pushes the repo in the wrong direction. 

There are many places where it is not happening:

https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/association_scope.rb#L152
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/association_scope.rb#L28
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/disable_joins_association_scope.rb#L43

I think such a check only relevant in code that is executed more frequently than 100 times a second.

Do you still want me to do it?

PR #57248, opened 2026-04-25
Fix: prevent duplicate WHERE in create_or_find_by retry

Patfarmurs:
### Summary
Ensured create_or_find_by does not duplicate WHERE conditions when retrying after RecordNotUnique.

### Details
- Adjusted retry logic in ActiveRecord::Relation.
- On retry, reuse the original attributes without appending duplicate conditions.
- Added regression test in relation_test.rb.

### Trade-offs
- Scoped fix to create_or_find_by only.
- Could extend to other retry paths but kept minimal to avoid unintended side effects.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,7 +150,6 @@ platforms :ruby, :windows do
   group :db do
     gem "pg", "~> 1.3"
     gem "mysql2", "~> 0.5", "< 0.5.7"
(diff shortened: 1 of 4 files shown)


fatkodima:
Duplicate of https://github.com/rails/rails/pull/57223.

Issue #50802, opened 2024-01-19
Deadlock occurs in zeitwerk


alpaca-tc:
### Steps to reproduce

There are the following two classes, each of which causes deadlock between Threads when loaded from different Threads.
This occurs in the development environment. That is, `enable_reloading? == true` `allow_concurrency == nil`. 

Sample Application: https://github.com/alpaca-tc/zeitwerk-test

```ruby
# app/models/first.rb
module First
  Second
end

# app/models/second.rb
module Second
  First
end

# app/controllers/constants_controller.rb
class ConstantsController < ActionController::Base
  def first
    First
    head :no_content
  end

  def second
    Second
    head :no_content
  end
end

# config/routes.rb
Rails.application.routes.draw do
  get '/first', to: 'constants#first'
  get '/second', to: 'constants#second'
end
```

```
# Deadlock when requesting in parallel. 
$ curl http://127.0.0.1:3000/first
$ curl http://127.0.0.1:3000/second
```

https://github.com/alpaca-tc/zeitwerk-test

### Expected behavior

Each request can be executed without deadlock.

### Actual behavior

The thread deadlocks and the request is stopped as Processing.

```
bundle exec rails server
...
Started GET "/first" for 127.0.0.1 at 2024-01-19 16:31:35 +0900
Started GET "/second" for 127.0.0.1 at 2024-01-19 16:31:35 +0900
Processing by ConstantsController#first as */*
Processing by ConstantsController#second as */*
```

### Additional information

If I reproduce it minimum, this issue is similar to the following gist.
https://gist.github.com/alpaca-tc/eb9ac65b7761276766c63948c7099cff

When a file with cyclic dependencies is `require/autoload`-ed by multi-thread, a deadlock occurs.
This behavior is intentional for Ruby.

Perhaps to prevent this, an exclusive lock should be acquired by `ActiveSupport::Dependencies.load_interlock` during `require` by zeitwerk. However, since there is no API in zeitwerk to such a hook, I did not know where I should acquire the lock.
The classic autoloader took an exclusive lock, so this problem does not occur.
https://github.com/rails/rails//blob/ac6aa32f7cf66264ba87eabed7c042bb60bcf3a2/activesupport/lib/active_support/dependencies.rb#L358-L392

### System configuration

Rails Version: main
Ruby Version: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]

I guess that it probably occurs in all versions that use zeitwerk.
[...]

PR #57249, opened 2026-04-25
Action Cable: explicitly require sibling deps in Redis adapter

gotrevor-notarize:
### Motivation / Background

The Redis subscription adapter file references two sibling constants at file-load time:

```ruby
class Redis < Base
  prepend ChannelPrefix
```

Action Cable opts the `subscription_adapter/` directory out of eager_load via [`loader.do_not_eager_load(...)`](https://github.com/rails/rails/blob/main/actioncable/lib/action_cable.rb), so `Base` and `ChannelPrefix` are resolved through Zeitwerk's lazy autoload on first reference to the Redis adapter. Under concurrent first-references the autoload-management window is racy and produces:

```
NameError: uninitialized constant ActionCable::SubscriptionAdapter::Base
    class Redis < Base # :nodoc:
                  ^^^^
Did you mean? Base64
```

The race window itself is small, but the *failure state* persists for the lifetime of the affected process: once a process loses the race, the autoload pointer for `Base` has been consumed and `Redis` is never defined. Every subsequent broadcast in that process raises `NameError` until the process is recycled. We observed this in production on a Rails 7.2.3.1 app under Puma cluster mode — a single host accumulated ~500 events across 15 sister Sentry issues over ~80 minutes before being replaced.

### Detail

This patch makes the file's sibling dependencies explicit:

```ruby
require "action_cable/subscription_adapter/base"
require "action_cable/subscription_adapter/channel_prefix"
```

Two reasons this is the right shape:

1. **Matches the file's existing style.** `redis.rb` already explicitly requires `redis` and `active_support/core_ext/hash/except` at the top.
2. **Matches the stated intent in `actioncable.rb`.** The `do_not_eager_load` line is annotated with the comment "Adapters are required and loaded on demand" — meaning the adapter files are expected to load their own dependencies, not rely on Zeitwerk autoload at evaluation time.

### Scope

The same pattern affects `subscription_adapter/test.rb`, `async.rb`, `inline.rb`, and `postgresql.rb` — each references a sibling at file-load time without explicit require. I've kept this PR focused on the file where the bug was actually observed and reproduced. Happy to extend if preferred.

### Tests

The race is non-deterministic — it requires Zeitwerk's internal autoload-management transition to overlap with a constant lookup from another thread. A deterministic regression test would need to either manipulate `$LOADED_FEATURES` + `Module#remove_const` (which pollutes the test process) or run in a subshell. Happy to add one if maintainers prefer a particular shape.

### Related

- rails/rails#50802 (Zeitwerk deadlock)
- fxn/zeitwerk#52 (autoload thread-safety)
- fxn/zeitwerk#198 (sidekiq+bootsnap autoload race)

### Checklist

- [x] Patch is small and focused
- [x] Matches existing file style
- [x] Existing requires demonstrate the file already declares its dependencies explicitly
- [ ] Test added (skipped — see above; happy to add)
- [N/A] CHANGELOG entry (not required for bug fixes per CONTRIBUTING)

### DIFF
--- a/actioncable/lib/action_cable/subscription_adapter/redis.rb
+++ b/actioncable/lib/action_cable/subscription_adapter/redis.rb
@@ -7,6 +7,9 @@ require "redis"
 
 require "active_support/core_ext/hash/except"
 
+require "action_cable/subscription_adapter/base"
+require "action_cable/subscription_adapter/channel_prefix"
+
 module ActionCable
   module SubscriptionAdapter
     class Redis < Base # :nodoc:


gotrevor-notarize:
@byroot now that your #57533 has merged, bump on this one.  That reimplementation didn't add explicit requires, so `Base` and `ChannelPrefix` are still autoloaded and the concurrent-autoload `NameError` this fixes can still occur.  We've seen this happen just twice in production over the past 6 months.  However, each time that it *did* happen was around the time that we'd just upgraded a major component, causing us to roll back and re-evaluate.  So, while there's a slim chance of hitting this race condition, it can be impactful.

PR #57250, opened 2026-04-25
Fix/postgres schema search path cache

Patfarmurs:
### Summary
Ensured PostgreSQL schema_search_path cache refreshes correctly after a transaction rollback.

### Details
- Added override in postgresql_adapter.rb to clear cached schema_search_path after rollback.
- Prevents stale schema_search_path values from persisting across transactions.
- Added regression test in schema_search_path_test.rb to confirm cache refresh.

### Trade-offs
- Scoped fix to rollback only.
- Could extend to other transaction events (commit, savepoint) but kept minimal to avoid unintended side effects.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,7 +150,6 @@ platforms :ruby, :windows do
   group :db do
     gem "pg", "~> 1.3"
     gem "mysql2", "~> 0.5", "< 0.5.7"
-    gem "trilogy", ">= 2.7.0"
   end
 end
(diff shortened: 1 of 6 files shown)


fatkodima:
There is a PR to fix that issue https://github.com/rails/rails/pull/57239.

Patfarmurs:
Thank you fatkodima, I'll work on it again and fix it.

PR #57251, opened 2026-04-26
fix/base-attribute-without-db

Patfarmurs:
### Summary
Fixed regression where ActiveRecord::Base.attribute raised an error if no database connection was defined.

### Details
- Introduced lazy type resolution for attributes defined without a DB connection.
- Added regression test to ensure attributes can be declared safely in models shipped in gems.
- Preserves compatibility with ActiveModel behavior.

### Trade-offs
- Scoped fix to attribute definition only.
- Did not extend to other DB-dependent features.

### Verification
- Ran full test suite successfully.
- New test passes and attributes can be defined without a DB connection.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,7 +150,6 @@ platforms :ruby, :windows do
   group :db do
     gem "pg", "~> 1.3"
     gem "mysql2", "~> 0.5", "< 0.5.7"
-    gem "trilogy", ">= 2.7.0"
   end
 end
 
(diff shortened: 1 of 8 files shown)


fatkodima:
There is a PR to fix that issue https://github.com/rails/rails/pull/57173.

Patfarmurs:
Thanks @fatkodima, I'll work on it later.

PR #57252, opened 2026-04-26
Enable frozen string literal by default

byroot:
Ref: https://github.com/rails/bootsnap/pull/535

This only impact the app own code, and not dependencies. It is also possible to enable it for gems, but some old ones may still not be ready.

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Enable Ruby `frozen_string_literal` by default.
+
+    New Rails apps now include a `config/bootsnap.rb` file that enables frozen string
+    literals. This only impact the application code, not the dependencies.
+
+    It is also possible to enable it for dependencies for reduce allocations, but
+    some older gems may not yet be compatible. If you do attempt this an run into
+    incompatibilites please do report it on the gem bug tracker.
+
+    Additionally, `.rubocop.yml` is configured to assume frozen string literals
+    are enabled, if you decide not to enable frozen string literals for your application,
+    make sureto update the rubocop configuration accordingly.
+
+    *Jean Boussier*
+
 *   Add offline fallback page to the PWA scaffold.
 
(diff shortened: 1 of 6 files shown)


railties/lib/rails/generators/rails/app/templates/rubocop.yml.tt
```diff
@@ -1,6 +1,10 @@
 # Omakase Ruby styling for Rails
 inherit_gem: { rubocop-rails-omakase: rubocop.yml }
 
+AllCops:
+    # This configuration should be kept in sync with `config/bootsnap.rb`
+    StringLiteralsFrozenByDefault: true
```
paracycle, line 7: Shouldn't this be conditional on `depend_on_bootsnap?` as well?

byroot: Yup

railties/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+*   Enable Ruby `frozen_string_literal` by default.
+
+    New Rails apps now include a `config/bootsnap.rb` file that enables frozen string
+    literals. This only impact the application code, not the dependencies.
+
+    It is also possible to enable it for dependencies for reduce allocations, but
+    some older gems may not yet be compatible. If you do attempt this an run into
```
elspethsoup, line 7: Small typo fix

```suggestion
    some older gems may not yet be compatible. If you do attempt this and run into
```

railties/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+    incompatibilites please do report it on the gem bug tracker.
+
+    Additionally, `.rubocop.yml` is configured to assume frozen string literals
+    are enabled, if you decide not to enable frozen string literals for your application,
+    make sureto update the rubocop configuration accordingly.
```
kpumuk, line 12: sureto > sure to

PR #57254, opened 2026-04-26
Fix/postgres schema search path cache

Patfarmurs:
### Summary
This PR fixes a bug where Rails continued to use a stale
`schema_search_path` after a transaction rollback in PostgreSQL.

### Details
- Added `@schema_search_path = nil` in `exec_rollback_db_transaction`
  and `exec_restart_db_transaction`.
- Ensures that the schema search path cache is refreshed after rollback.
- Scoped the fix only to rollback paths to avoid unnecessary overhead
  on commit or other transaction events.

### Trade-offs
- Refreshing the cache on every transaction event was considered, but
  would add performance overhead. Limiting the fix to rollback events
  keeps the change minimal and safe.

### Verification
- Added regression tests to confirm that after rollback, the cached
  schema search path is cleared.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,7 +150,6 @@ platforms :ruby, :windows do
   group :db do
     gem "pg", "~> 1.3"
     gem "mysql2", "~> 0.5", "< 0.5.7"
-    gem "trilogy", ">= 2.7.0"
   end
 end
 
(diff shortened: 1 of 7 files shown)


fatkodima:
Duplicate of https://github.com/rails/rails/pull/57239.

PR #57256, opened 2026-04-27
fix(active_storage): preserve attachment_changes across record reload

### LINKED ISSUE #57222: ActiveStorage uploads don't trigger in a transaction if the model is reloaded

jackie20:
When a record with a pending attachment (set but not yet saved) is reloaded, `@attachment_changes` was unconditionally set to nil. Any attachment change that had already been written to the database via `after_save` but not yet uploaded to the storage service via `after_commit` would be silently dropped, leaving a dangling attachment record with no corresponding file on the service.

The typical trigger is an explicit outer transaction: save writes the attachment record, then the application calls reload (e.g. to refresh other columns), and the outer transaction commits — at which point `after_commit` finds an empty `attachment_changes` hash and skips the upload.

Fix: reset `@active_storage_attached` (the memoised proxy objects) on reload rather than `@attachment_changes` (the pending uploads). Resetting the proxy cache is still correct — it forces the next access to re-read the underlying association from the freshly reloaded record — but the upload queue is preserved so `after_commit` can finish the job.

Adds two regression tests:
- pending attachment_changes survive reload
- an assignment → reload → save round-trip still delivers a downloadable file

Fixes #57222

### Motivation / Background

### Detail

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/attached/model.rb
+++ b/activestorage/lib/active_storage/attached/model.rb
@@ -318,7 +318,7 @@ module ActiveStorage
     end
 
     def reload(*) # :nodoc:
-      super.tap { @attachment_changes = nil }
+      super.tap { @active_storage_attached = nil }
     end
   end
 end
--- a/activestorage/test/models/attached/one_test.rb
+++ b/activestorage/test/models/attached/one_test.rb
@@ -430,6 +430,21 @@ class ActiveStorage::OneAttachedTest < ActiveSupport::TestCase
     assert_predicate @user.avatar, :attached?
   end
 
+  test "reload preserves pending attachment changes" do
+    @user.avatar = fixture_file_upload("racecar.jpg")
+    assert_includes @user.attachment_changes, "avatar"
+    @user.reload
+    assert_includes @user.attachment_changes, "avatar",
+      "reload must not discard unsaved attachment changes"
+  end
+
+  test "attachment assignment survives a reload before save" do
+    @user.avatar = fixture_file_upload("racecar.jpg")
+    @user.reload
+    @user.save!
+    assert_nothing_raised { @user.avatar.download }
+  end
+
   test "attaching an existing blob to a new record" do
     User.new(name: "Jason").tap do |user|
       user.avatar.attach create_blob(filename: "funky.jpg")


tahsin352:
We already have a fix for the issue https://github.com/rails/rails/issues/57222 here: https://github.com/rails/rails/pull/57229. It seems your PR is incomplete, so, we can close this PR. Thanks.

PR #57258, opened 2026-04-27
Fix #57197: Preserve retry_job wait when enqueue is deferred

### LINKED ISSUE #57197: `retry_job` ignores `wait` when called inside transaction
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
end

require "active_job/railtie"
require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

  end
end

end

class LoopingJob < ActiveJob::Base
  self.enqueue_after_transaction_commit = true

  def perform
    Post.transaction do
      retry_job wait: 5.second
    end
  end
end

class LoopingJobTest < ActiveJob::TestCase
  def test_stuff
    LoopingJob.perform_now
    
    scheduled = enqueued_jobs.last["scheduled_at"]

    assert_not_nil scheduled
  end
end
```

### Expected behavior

The job should be scheduled 5 seconds in the future

### Actual behavior

The job has no scheduled_at, meaning its enqueued immediately

### System configuration

**Rails version**: 8.1.3 (https://github.com/rails/rails/pull/55141 broke this)

**Ruby version**: N/A

maryamfayyaz:
**Problem**: When enqueue_after_transaction_commit defers enqueueing, [retry_job(wait: ...)](app://-/index.html?hostId=local#) can lose its delay and enqueue immediately (no scheduled_at / adapter :at).
**Cause**: The actual enqueue happens inside after_all_transactions_commit, after retry_job has restored/cleared transient state, so the deferred super enqueue doesn’t see the original schedule info.
**Fix**: In ActiveJob::EnqueueAfterTransactionCommit#raw_enqueue, capture scheduled_at/queue_name/priority at enqueue time and re-apply them only for the deferred super call.
**Tests**: Add regression test in [enqueue_after_transaction_commit_test.rb](app://-/index.html?hostId=local#) asserting the TestAdapter receives an :at timestamp ~5 seconds in the future.
**Changelog**: Update [CHANGELOG.md](app://-/index.html?hostId=local#).

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Fix `retry_job` to respect `:wait` when enqueue is deferred until after transaction commit.
+
+    *Maryam Fayyaz*
+
 *   Deprecate built-in `queue_classic` Active Job adapter.
 
     *Harun Sabljaković, Wojciech Wnętrzak*
--- a/activejob/lib/active_job/enqueue_after_transaction_commit.rb
+++ b/activejob/lib/active_job/enqueue_after_transaction_commit.rb
@@ -25,9 +25,16 @@ module ActiveJob
       def raw_enqueue
         if self.class.enqueue_after_transaction_commit
           self.successfully_enqueued = true
+          scheduled_at, queue_name, priority = self.scheduled_at, self.queue_name, self.priority
           ActiveRecord.after_all_transactions_commit do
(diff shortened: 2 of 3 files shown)


fatkodima:
Thanks for the PR! Let's move forward with #57198, since it was created earlier.

PR #57260, opened 2026-04-27
Fix authentication generator users migration conflict

Charity-P:
### Motivation / Background

### Detail

### Checklist


Charity-P:
## Summary

This pull request fixes an issue in the authentication generator where the `CreateUsers` migration was generated unconditionally, even when overwrite of `app/models/user.rb` was declined.

## Problem

When an existing `User` model was present and overwrite was declined during generator execution, the generator still created:

`db/migrate/*_create_users.rb`

Running `db:migrate` would then fail because the `users` table already existed.

## Root Cause

The `add_migrations` method always generated both migrations without checking whether the `User` model had actually been created or overwritten successfully.

## Solution

Updated the authentication generator so that:

* `CreateUsers` migration is only generated when the `User` model file is successfully created or overwritten.
* `CreateSessions` migration continues to generate as expected.

## Test Coverage

Added regression test coverage to confirm that when overwrite of `app/models/user.rb` is declined:

* `CreateUsers` migration is skipped
* `CreateSessions` migration is still generated

## Result

This change prevents duplicate `users` table migration conflicts and makes the generator behavior consistent with developer intent.

PR #57267, opened 2026-04-29
Fix bulk job and email enqueueing methods with no arguments

### LINKED ISSUE #57264: undefined method 'deliver_all_later' for module ActionMailer
### Steps to reproduce

[ActionMailer.deliver_all_later](https://github.com/rails/rails/blob/fa8f0812160665bff083a089d2bb2fc1817ea03e/actionmailer/lib/action_mailer/message_delivery.rb#L9) is a documented means of delivering multiple emails at the same time. However, in a brand new Rails v8.1 application...

```ruby
irb> ActionMailer.deliver_all_later
undefined method 'deliver_all_later' for class ApplicationMailer (NoMethodError)
```

The workaround is to `require 'action_mailer/delivery_methods'` in `config/application.rb` below `require 'rails/all'`.

### Expected behavior

It should be defined by default OR it should be documented how this should be included.

### Actual behavior

```ruby
irb> ActionMailer.deliver_all_later
undefined method 'deliver_all_later' for class ApplicationMailer (NoMethodError)
```

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 4.0.3

fatkodima:
Fixes #57264 (see there for the problem discussion).

This PR only moves some code around, without any code changes.

### DIFF
--- a/actionmailer/lib/action_mailer.rb
+++ b/actionmailer/lib/action_mailer.rb
@@ -58,15 +58,47 @@ module ActionMailer
   autoload :QueuedDelivery
   autoload :FormBuilder
 
-  def self.eager_load!
-    super
-
-    require "mail"
-    Mail.eager_autoload!
-
-    Base.descendants.each do |mailer|
-      mailer.eager_load! unless mailer.abstract?
+  class << self
+    # Enqueue many emails at once to be delivered through Active Job.
+    # When the individual job runs, it will send the email using +deliver_now+.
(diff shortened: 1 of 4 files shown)


skipkayhil (approved):
The `self.` vs `class << self` changes are a little odd (it would be nice if the framework was internally consistent), but I think the change itself makes sense

byroot (approved):

Issue #57265, opened 2026-04-28
`load_defaults 8.1` registers ERBTracker instead of RubyTracker when `eager_load = true`


CowboyPatrick:
### Steps to reproduce

The following script (based on `guides/bug_report_templates/action_view.rb`) boots a stock Rails 8.1 application with `load_defaults 8.1` and `eager_load = true`, then asserts that `ActionView::DependencyTracker` registered the expected tracker for `.erb` templates:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"
end

require "action_controller/railtie"
require "action_view/railtie"
require "minitest/autorun"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.eager_load = true # the trigger
  config.logger = Logger.new($stdout)
  config.secret_key_base = "secret_key_base"
end
Rails.application.initialize!

class RenderTrackerEagerLoadTest < Minitest::Test
  def test_load_defaults_8_1_registers_ruby_tracker_under_eager_load
    assert_equal :ruby, Rails.application.config.action_view.render_tracker
    assert_equal :ruby, ActionView.render_tracker

    trackers = ActionView::DependencyTracker.instance_variable_get(:@trackers)
    erb_handler = ActionView::Template.handler_for_extension("erb")

    assert_equal ActionView::DependencyTracker::RubyTracker, trackers[erb_handler]
  end
end
```

### Expected behavior

The test passes — `RubyTracker` is registered for `.erb` templates because `load_defaults 8.1` sets `config.action_view.render_tracker = :ruby` (per #55194).

### Actual behavior

```
RenderTrackerEagerLoadTest#test_load_defaults_8_1_registers_ruby_tracker_under_eager_load:
--- expected
+++ actual
@@ -1 +1 @@
-ActionView::DependencyTracker::RubyTracker
+ActionView::DependencyTracker::ERBTracker
```

The first two assertions pass — both `config.action_view.render_tracker` and the `ActionView.render_tracker` module accessor end up `:ruby` after `initialize!`. But `DependencyTracker` has `ERBTracker` registered, meaning the value the consumer sees during template rendering is the wrong one.

### Root cause

Two pieces interact:

[...]

PR #57268, opened 2026-04-29
Fix: registers ERBTracker instead of RubyTracker when eager_load = true

joaoGabriel55:
### Motivation / Background

Fixes issue #57265 

Fix a bug where `ActionView.render_tracker` was being assigned inside a nested `config.after_initialize` block, which caused it to run too late (after the outer `after_initialize` callbacks had already completed). As a result, when `eager_load: true` was set, the ERB handler was never properly mapped to `RubyTracker` in `ActionView::DependencyTracker`, breaking dependency tracking for ERB templates.

### Detail

This Pull Request changes the assignment of `ActionView.render_tracker` from a nested `config.after_initialize` block to a dedicated `initializer "action_view.set_render_tracker"`, ensuring it runs at the correct point in the boot sequence, before `after_initialize` callbacks fire.

**Before:**
```ruby
  config.after_initialize do
  end
# ...
end
```

**After:**
```ruby
initializer "action_view.set_render_tracker" do |app|
  ActionView.render_tracker = app.config.action_view.render_tracker
end
```

A regression test was added to `railties/test/application/configuration_test.rb` verifying that with `load_defaults 8.1`, `eager_load: true`, and `enable_reloading: false`, the ERB handler is correctly mapped to `ActionView::DependencyTracker::RubyTracker`.

### Additional information

N/A

### Checklist

### DIFF
--- a/actionview/lib/action_view.rb
+++ b/actionview/lib/action_view.rb
@@ -37,6 +37,7 @@ module ActionView
   eager_autoload do
     autoload :Base
     autoload :Context
+    autoload :DependencyTracker
     autoload :Digestor
     autoload :Helpers
     autoload :LookupContext
@@ -91,7 +92,19 @@ module ActionView
   autoload :CacheExpiry
   autoload :TestCase
 
-  singleton_class.attr_accessor :render_tracker
+  singleton_class.attr_reader :render_tracker
+
+  def self.render_tracker=(value)
+    @render_tracker = value
+    case value
+    when :ruby
+      DependencyTracker.register_tracker :erb, DependencyTracker::RubyTracker
+    else
+      DependencyTracker.register_tracker :erb, DependencyTracker::ERBTracker
+    end
+    value
+  end
+
   self.render_tracker = :regex
 
   def self.eager_load!
--- a/actionview/lib/action_view/railtie.rb
+++ b/actionview/lib/action_view/railtie.rb
@@ -80,10 +80,6 @@ module ActionView
     end
 
     config.after_initialize do |app|
-      config.after_initialize do
-        ActionView.render_tracker = config.action_view.render_tracker
-      end
-
       ActiveSupport.on_load(:action_view) do
         app.config.action_view.each do |k, v|
           next if k == :render_tracker
@@ -106,6 +102,10 @@ module ActionView
           ActionView::Resolver.caching = !app.config.reloading_enabled?
         end
       end
+    end
+
+    initializer "action_view.set_render_tracker" do |app|
+      ActionView.render_tracker = app.config.action_view.render_tracker
     end
 
     initializer "action_view.setup_action_pack" do |app|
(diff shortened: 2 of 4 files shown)


actionview/lib/action_view/railtie.rb
```diff
@@ -108,6 +104,10 @@ class Railtie < Rails::Engine # :nodoc:
       end
     end
 
+    initializer "action_view.set_render_tracker" do |app|
+      ActionView.render_tracker = app.config.action_view.render_tracker
```
skipkayhil, line 108: Running inside an initializer is too early, I believe this prevents the value from being changed inside `new_framework_defaults.rb`

Edit: I'm thinking we should probably have `ActionView#render_tracker=` (re-)register the correct tracker?

joaoGabriel55: Hello! Using Claude Code to help me with this question. It appears that the current solution is correct.

`load_config_initializers` runs **before** `set_render_tracker`, not after. Result: `ActionView.render_tracker == :ruby` after boot, the config initializer's value is honored.

Why: `Rails::Application#railties_initializers` iterates `ordered_railties.reverse.flatten` [railties/lib/rails/application.rb:728](railties/lib/rails/application.rb:728). `ordered_railties` ends with `self` (the app), so `.reverse` puts the app **first**. The app's engine initializers (which include `:load_config_initializers`, registered on `Rails::Engine` at engine.rb:662) therefore land *before* any framework railtie's initializers (like `action_view.set_render_tracker`). And `:eager_load!` lives in `Rails::Application::Finisher`, which is appended last, so it runs after `set_render_tracker`.

Does that make sense?

rafaelfranca: Instead of using claude code to answer, which sometimes it get wrong, can you write a test to make sure that problem isn't really a problem?

skipkayhil: Yeah, I think you're right: application initializers can run before this so it would probably fix the issue. However, I think we should still change `render_tracker=` so that the value can be changed at any time instead of only before a particular class has been loaded.

rafaelfranca (approved):

actionview/lib/action_view.rb
```diff
@@ -91,7 +91,21 @@ module ActionView
   autoload :CacheExpiry
   autoload :TestCase
 
-  singleton_class.attr_accessor :render_tracker
+  singleton_class.attr_reader :render_tracker
+
+  def self.render_tracker=(value)
+    @render_tracker = value
+    if defined?(DependencyTracker)
```
rafaelfranca: Why are we using this `defined?`

joaoGabriel55: The `defined?(DependencyTracker)` guard is there because `DependencyTracker` is **not autoloaded by `ActionView`**, it's a standalone class under `action_view/dependency_tracker.rb` that gets required separately (typically by Railties during boot)

If we using `DependencyTracker` directly without the `defined?` check, Ruby would raise a `NameError` since the constant hasn't been loaded yet.

I don't have a lot of experience with the Rails codebase. Should I try a different approach?

rafaelfranca: Yes, let's define an autoload for `DependencyTracker` so we don't need to check if it is defined.

rafaelfranca (changes requested):

rafaelfranca (approved):

PR #57269, opened 2026-04-29
Defer ViewReloader build when no view paths are registered

Korri:
### Summary

Follow-up to #51308 ("Do not build View watcher until the first `updated?` check").

### Root cause

#51308's lazy guard relies on `@watcher` staying `nil` until view paths are registered. But it breaks when `updated?` is called while `dirs_to_watch` is still an empty array. [`#build_watcher`](https://github.com/rails/rails/blob/83423052fc899315fad91c20b5197a186fc8f0fd/actionview/lib/action_view/cache_expiry.rb#L37-L48) still constructs an empty `FileUpdateChecker` and assigns it to `@watcher`. From then on the [`return unless @watcher`](https://github.com/rails/rails/blob/83423052fc899315fad91c20b5197a186fc8f0fd/actionview/lib/action_view/cache_expiry.rb#L55-L58) guard in `#rebuild_watcher` is open, and every engine's subsequent `prepend_view_path` triggers a full `Dir[*globs] + File.mtime` rescan on the cumulative dir set.

Spring hits this when running a test with a warm cache, the preloader walks `Rails.application.reloaders` directly to decide whether to `reload!` before forking ([`spring/lib/spring/application.rb:197`](https://github.com/rails/spring/blob/main/lib/spring/application.rb#L197)) — bypassing `Reloader.check!`. Most engines' `prepend_view_path` calls live behind `ActiveSupport.on_load(:action_mailer)` / `:action_controller_base`, which haven't fired yet in Spring's preload phase, so the probe materializes an empty watcher and pins `@watcher` non-nil for every fork.

### Example

Assuming the (extreme) case where we would have 10 `prepend_view_path` calls:

**Pre-fix**:

 | Step | Real work? |
 |---|---|
 | Early-boot `updated?`  | no (cheap) |
 | `prepend_view_path` ×10 |  **10 real rebuilds** |
 | First request `updated?` |  no rebuild |

 **Post-fix**:

 | Step | Real work? |
 |---|---|
 | Early-boot `updated?` | no |
 | `prepend_view_path` ×10  | **0 rebuilds** |
 | First request `updated?`  | **1 build** |

### Impact

Measured on a Spring-warm test loop in Shopify's Rails monorepo, wall-time on a small workload drops from 5.16 s → 3.66 s (-29 %). Behavior is unchanged for production (`cache_classes = true` skips ViewReloader entirely) and for any app where view paths are registered before the first reloader probe.

### DIFF
--- a/actionview/lib/action_view/cache_expiry.rb
+++ b/actionview/lib/action_view/cache_expiry.rb
@@ -9,13 +9,11 @@ module ActionView
         @watched_dirs = nil
         @watcher = nil
         @previous_change = false
-
-        ActionView::PathRegistry.file_system_resolver_hooks << method(:rebuild_watcher)
       end
 
       def updated?
         build_watcher unless @watcher
-        @previous_change || @watcher.updated?
+        @previous_change || @watcher&.updated?
       end
 
       def execute
@@ -29,6 +27,11 @@ module ActionView
         watcher.execute
       end
 
+      def rebuild_watcher
+        return unless @watcher
+        build_watcher
+      end
+
       private
         def reload!
           ActionView::LookupContext::DetailsKey.clear
@@ -36,10 +39,15 @@ module ActionView
 
         def build_watcher
           @mutex.synchronize do
+            new_dirs = dirs_to_watch
+
+            # Skip the build entirely if there are no view paths to watch and we have not built a watcher yet.
+            return if new_dirs.empty? && @watcher.nil?
+
             old_watcher = @watcher
 
-            if @watched_dirs != dirs_to_watch
-              @watched_dirs = dirs_to_watch
+            if @watched_dirs != new_dirs
+              @watched_dirs = new_dirs
               new_watcher = @watcher_class.new([], @watched_dirs) do
                 reload!
               end
@@ -50,11 +58,6 @@ module ActionView
               @previous_change ||= old_watcher&.updated?
             end
           end
-        end
-
-        def rebuild_watcher
-          return unless @watcher
-          build_watcher
         end
 
         def dirs_to_watch
--- a/actionview/test/template/cache_expiry_test.rb
+++ b/actionview/test/template/cache_expiry_test.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+require "abstract_unit"
+
+class CacheExpiryViewReloaderTest < ActiveSupport::TestCase
+  # A no-op watcher class with the FileUpdateChecker API. Used instead of the
+  # real ActiveSupport::FileUpdateChecker so we can assert exactly when the
+  # watcher is materialized — its #initialize bumps a counter we can read.
+  class CountingWatcher
+    @@built = 0
+
+    class << self
+      def reset!
+        @@built = 0
+      end
+
+      def built
+        @@built
+      end
+    end
+
+    def initialize(_files, _dirs = {}, &block)
+      @block = block
+      @@built += 1
+    end
+
+    def updated?
+      false
+    end
+
+    def execute
+      @block&.call
+    end
+
+    def execute_if_updated
+      false
+    end
+  end
+
+  def setup
+    CountingWatcher.reset!
+    @reloader = ActionView::CacheExpiry::ViewReloader.new(watcher: CountingWatcher)
+  end
+
+  test "#updated? does not build a watcher when there are no view paths to watch" do
+    ActionView::PathRegistry.stub(:all_file_system_resolvers, []) do
+      assert_not @reloader.updated?
+    end
+
+    assert_equal 0, CountingWatcher.built
+  end
+
+  test "#rebuild_watcher is a no-op while @watcher has not been built" do
+    @reloader.rebuild_watcher
+
+    assert_equal 0, CountingWatcher.built
+  end
+
+  test "updated? builds a watcher once view paths are registered" do
+    paths = []
+    ActionView::PathRegistry.stub(:all_file_system_resolvers, paths) do
+      @reloader.updated?
+    end
+    assert_equal 0, CountingWatcher.built
+
+    paths << resolver_for_path("fixtures")
+    ActionView::PathRegistry.stub(:all_file_system_resolvers, paths) do
+      @reloader.updated?
+    end
+
+    assert_equal 1, CountingWatcher.built
+  end
+
+  test "#rebuild_watcher rebuilds the watcher after the first build" do
+    paths = [resolver_for_path("fixtures")]
+    ActionView::PathRegistry.stub(:all_file_system_resolvers, paths) do
+      @reloader.updated?
+    end
+
+    assert_equal 1, CountingWatcher.built
+
+    paths = paths << resolver_for_path("../fixtures/test")
+    ActionView::PathRegistry.stub(:all_file_system_resolvers, paths) do
+      @reloader.rebuild_watcher
+    end
+
+    assert_equal 2, CountingWatcher.built
+  end
+
+  private
+    def resolver_for_path(path)
+      ActionView::PathRegistry.cast_file_system_resolvers(path).first
+    end
+end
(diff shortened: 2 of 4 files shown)


nvasilevski:
>Rails.application.reloader invokes [app.reloaders.map(&:updated?).any?](https://github.com/rails/rails/blob/83423052fc899315fad91c20b5197a186fc8f0fd/railties/lib/rails/application/finisher.rb#L195) early in boot

I'll provide a better review tomorrow but for know I'd like to question this bit - is this expected? `app.reloaders.map(&:updated?).any?` wrapped in a lambda so something has to call this updater check. I wonder why do we happen to check for updates during boot. I think I want to claim that this is the root cause and the `reloader.check`  lambda shouldn't be called on boot in the first place

Korri:
> > Rails.application.reloader invokes [app.reloaders.map(&:updated?).any?](https://github.com/rails/rails/blob/83423052fc899315fad91c20b5197a186fc8f0fd/railties/lib/rails/application/finisher.rb#L195) early in boot
> 
> I'll provide a better review tomorrow but for know I'd like to question this bit - is this expected? `app.reloaders.map(&:updated?).any?` wrapped in a lambda so something has to call this updater check. I wonder why do we happen to check for updates during boot. I think I want to claim that this is the root cause and the `reloader.check` lambda shouldn't be called on boot in the first place

I think you might be right, I updated the description to clarify what is happening here, this is a problem _because_ Spring bypasses the `Reloader.check!`. I'm too out of my area of understanding to understand where the _right_ place is to fix this. It does sound to me like the patch here is a good improvement anyways, but I might be wrong.

Korri:
Here is a quick and dirty 100% AI attempt at fixing this on the Spring side: https://github.com/rails/spring/pull/752

actionview/lib/action_view/cache_expiry.rb
```diff
@@ -15,7 +15,7 @@ def initialize(watcher:, &block)
 
       def updated?
         build_watcher unless @watcher
-        @previous_change || @watcher.updated?
+        @previous_change || @watcher&.updated? || false
```
nvasilevski: we don't have typing so returning `nil` as a falsey value should be fine
```suggestion
        @previous_change || @watcher&.updated?
```

actionview/lib/action_view/cache_expiry.rb
```diff
@@ -36,10 +36,15 @@ def reload!
 
         def build_watcher
           @mutex.synchronize do
+            new_dirs = dirs_to_watch
+
+            # Skip the build entirely if there are no view paths to watch and we have not built a watcher yet.
+            return if new_dirs.empty? && @watcher.nil?
```
nvasilevski, line 45: `dirs_to_watch` has some cost to build, should we split checks into two and only build `dirs_to_watch` if we know we have a `@watcher`
```suggestion
            new_dirs = dirs_to_watch

            # Skip the build entirely if there are no view paths to watch and we have not built a watcher yet.
            return if @watcher.nil? || (new_dirs = dirs_to_watch).empty?
```

nvasilevski (approved):
While I still would like to find some time to understand the whole flow better this change seems to be cheap and never harmful, it sounds reasonable to avoid building a watcher when there is no dirs to watch

actionview/lib/action_view/cache_expiry.rb
Korri: A bunch of tests started failing because they expect `false` so I went with that fallback

actionview/lib/action_view/cache_expiry.rb
Korri: It's `&&` not `||` so we have to run both either way.

actionview/lib/action_view/cache_expiry.rb
Korri: Actually it was my own test 🤦🏻 I'll make it falsey!

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def setup
+    CountingWatcher.reset!
+    @reloader = ActionView::CacheExpiry::ViewReloader.new(watcher: CountingWatcher)
+  end
+
+  def teardown
+    # Drop the rebuild_watcher hook this reloader registered in the global
+    # PathRegistry so it can't fire during later tests.
+    ActionView::PathRegistry.file_system_resolver_hooks.delete_if do |hook|
```
rafaelfranca: This and the other patch that I just saw make me think that `ViewReloader` intializer should not be the one adding the callback to the  `file_system_resolver_hooks`. Let's move it to the railtie and simplify this code.

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+      hook.respond_to?(:receiver) && hook.receiver.equal?(@reloader)
+    end
+  end
+
+  test "updated? does not build a watcher when there are no view paths to watch" do
+    @reloader.define_singleton_method(:dirs_to_watch) { [] }
```
rafaelfranca: Can't we use a regular stub?

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    assert_nil @reloader.instance_variable_get(:@watcher)
```
rafaelfranca:
```suggestion
```

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    assert_not @reloader.updated?
+    assert_nil @reloader.instance_variable_get(:@watcher),
+      "ViewReloader should not materialize a watcher when there are no view paths"
```
rafaelfranca:
```suggestion
```

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    assert_equal 0, CountingWatcher.built
```
rafaelfranca, line 50: This assertion should be enough and doesn't reach to the internals

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  test "rebuild_watcher is a no-op while @watcher has not been built" do
+    @reloader.define_singleton_method(:dirs_to_watch) { [] }
+
+    @reloader.updated?
+
+    # Simulate the `cast_file_system_resolvers` callback firing while no view
+    # paths are registered — common during early Rails boot when an Engine
+    # has not yet run its on_load(:action_mailer) hook.
+    @reloader.send(:rebuild_watcher)
```
rafaelfranca: Given my suggestion of moving the callback registering to outside of the object, this method would be made public, so no more send.

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    assert_nil @reloader.instance_variable_get(:@watcher)
```
rafaelfranca: No need to assert this

actionview/test/template/cache_expiry_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    @reloader.updated?
+
+    assert_not_nil @reloader.instance_variable_get(:@watcher),
+      "ViewReloader should materialize a watcher once there are view paths"
+    assert_equal 1, CountingWatcher.built
+  end
+
+  test "rebuild_watcher rebuilds the watcher after the first build" do
+    paths = [File.expand_path("fixtures", __dir__)]
+    @reloader.define_singleton_method(:dirs_to_watch) { paths.dup }
```
rafaelfranca: Why do we need the `dup` here but not above?

rafaelfranca (changes requested):

Korri:
@rafaelfranca sorry those tests where a mess! They should be good now! The suggestion about moving the registering out works great too, I updated the Changelog to mention it.

rafaelfranca (approved):

PR #57271, opened 2026-04-30
Gracefully handle `nil` during multi-parameter assignment

seanpdoyle:
### Motivation / Background

When assigning to multi-parameter attributes (like `Date` columns), Active Record currently handles `""` values as empty values, resulting in the following:

```ruby
topic = Topic.where.not(last_read: nil).first
topic.attributes = { "last_read(1i)" => "", "last_read(2i)" => "", "last_read(3i)" => "" }
topic.last_read # => nil
```

However, if the `""` values are replaced with `nil`, the assignment raises a `NoMethodError`:

```
NoMethodError: undefined method 'empty?' for nil
```

### Detail

In this scenario, `""` and `nil` are semantically equivalent: the value is absent. However, the implementation invokes [String#empty?][], which is *not* implemented by [NilClass][].

This commit proposes that the implementation check for [String#blank?][] instead (which relies on [String#empty?][] under the hood), since there is also a [NilClass#blank?][] implementation.

```ruby
topic = Topic.where.not(last_read: nil).first
topic.attributes = { "last_read(1i)" => nil, "last_read(2i)" => nil, "last_read(3i)" => nil }
topic.last_read # => nil
```

[String#empty?]: https://ruby-doc.org/3.4.1/String.html#method-i-empty-3F
[NilClass]: https://edgeapi.rubyonrails.org/classes/NilClass.html
[String#blank?]: https://edgeapi.rubyonrails.org/classes/String.html#method-i-blank-3F
[NilClass#blank?]: https://edgeapi.rubyonrails.org/classes/NilClass.html#method-i-blank-3F

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Treat `nil` values as `""` during multi-parameter attribute assignment
+
+    ```ruby
+    topic = Topic.where.not(last_read: nil).first
+    topic.attributes = { "last_read(1i)" => nil, "last_read(2i)" => nil, "last_read(3i)" => nil }
+    topic.last_read # => nil
+    ```
+
+    *Sean Doyle*
+
 *   Bump the minimum PostgreSQL version to 10.0.
 
     As part of this change, `supports_pgcrypto_uuid?` is deprecated because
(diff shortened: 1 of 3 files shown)


chaadow:
wow @seanpdoyle thanks to your PR i learnt two things:
- that `#attributes=` is an alias for `assign_attributes` ( I've always used the latter )
- that you can edit parts of a DateTime attribute

Do you know if there is an official documentatino about this? by spelunking the code, I only found some documentation of a private method that explains the `i` and `f` casting : https://github.com/rails/rails/blob/2d670320f7b02ae879545d5202f0633841b8f196/activerecord/lib/active_record/attribute_assignment.rb#L30-L40

PR #57274, opened 2026-04-30
Fix preloader batch merging for parallel scoped through

### LINKED ISSUE #52061: `includes` loads wrong data for scoped `has_many :through` Association when eager loading nested self join
### Steps to reproduce

Run `bin/rails db:reset && bin/rails test test/models/work_test.rb` in this minimal Rails application: https://github.com/Bilka2/includes-polymorphic. Relevant code: https://github.com/Bilka2/includes-polymorphic/commit/3016fb21c8e91302ae15051c11ba41425f9f5857

I could not reproduce the issue with the single file test template. However, the code is basically the same as the linked commit, so I am including it anyway:

```ruby
# frozen_string_literal: true

# This does NOT reproduce the issue, despite containing the same code as the Rails app linked above

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "sqlite3", "~> 1.4"
end

require "active_record"
require "minitest/autorun"
require "logger"

# config/application.rb
  config.load_defaults 8.0

  config.active_record.belongs_to_required_by_default = false
end

  create_table "taggings", id: :integer, force: :cascade do |t|
    t.integer "tag_id"
    t.string "tag_type", limit: 100, default: ""
    t.integer "work_id"
  end

  create_table "tags", id: :integer, force: :cascade do |t|
    t.string "type"
    t.integer "merger_id"
  end

  create_table "works", id: :integer, force: :cascade do |t|
  end
end

class Tagging < ActiveRecord::Base
  belongs_to :tag, polymorphic: true, inverse_of: :taggings
  belongs_to :work, inverse_of: :taggings
end

  TYPES = ['Fandom', 'Relationship']

  has_many :mergers, foreign_key: 'merger_id', class_name: 'Tag'
  belongs_to :merger, class_name: "Tag"

  has_many :taggings
end

class Fandom < Tag
end
class Relationship < Tag
end

class Work < ActiveRecord::Base
  has_many :taggings, inverse_of: :work
  has_many :tags, through: :taggings

  Tag::TYPES.each do |type|
    has_many type.underscore.pluralize.to_sym,
             -> { where(tags: { type: type }) },
             through: :taggings,
             source: :tag,
             source_type: "Tag"
  end
end

class WorkTest < Minitest::Test
  def test_includes_scoped_polymorphic
    f1 = Fandom.create!
    rel1 = Relationship.create!
    Work.create!(fandoms: [f1], relationships: [rel1])    
    Work.first.relationships.each { |rel| assert_equal "Relationship", rel.type } # passes

    work_with_includes = Work.includes(fandoms: :merger, relationships: :merger).first
    work_with_includes.relationships.each { |rel| assert_equal "Relationship", rel.type } # fails
  end
end
```

### Expected behavior
`work_with_includes.relationships` should return only Relationship tags.

### Actual behavior

`work_with_includes.relationships` also returns the Work's Fandom tag (as well as the Relationship tag).

```
Failure:
WorkTest#test_includes_scoped_polymorphic [test/models/work_test.rb:11]:
Expected: "Relationship"
  Actual: "Fandom"
```

The SQL query looks correct:
```
  Tag Load (0.1ms)  SELECT "tags".* FROM "tags" WHERE "tags"."type" = ? AND "tags"."id" IN (?, ?)  [["type", "Fandom"], ["id", 1], ["id", 2]]
  Tag Load (0.1ms)  SELECT "tags".* FROM "tags" WHERE "tags"."type" = ? AND "tags"."id" = ?  [["type", "Relationship"], ["id", 2]]
```

The order of the associations in `includes` is relevant. If `fandoms:` and `relationships:` are switched, then `relationships` contains only the Relationship tag and `fandoms` exhibits the issue and contains the Relationship tag as well as the Fandom tag.  
If `:merger` (the self join) is removed from `includes`, the issue does not occur.  
If `includes` is split into two calls `includes(fandoms: :merger).includes(relationships: :merger)` the issue does not occur.

### System configuration

[Initial discovery](https://otwarchive.atlassian.net/browse/AO3-6729) and [workaround](https://github.com/otwcode/otwarchive/pull/4814):
**Rails version**: 7.0.8.1 (Issue discovered after upgrade from Rails 6.1)
**Ruby version**: 3.1.4
MariaDB

Latest reproduction:
**Rails version**: 8.0.0.alpha 8d416d09fc7fc226c6e54dd772e1b8f1969c5d61
**Ruby version**: 3.3.2

Edilbek:
# Fix #52061 — parallel scoped `has_many :through` eager loading

Partition preloader batches for inner through steps by the **outer** `has_many :through` reflection so parallel scoped throughs that share the same SQL do not corrupt inverse / STI rows when using nested `includes`. Also fix `records_by_owner` merging for sibling through loaders.

**Fixes #52061**

## Motivation / background

Two issues showed up for the same class of setup (parallel scoped `has_many :through` on a shared join, e.g. `includes(fandoms: :merger, relationships: :merger)`):

1. **`Preloader::Batch#group_and_load_similar`** grouped non-through loaders only by `[loader_query, klass]`. Parallel branches can produce **identical** `LoaderQuery` (same SQL) for the **same** inner through association (e.g. `:taggings`) while still representing different outer throughs (`:fandoms` vs `:relationships`). Merging those into one batch meant **`LoaderRecords`** ran **`set_inverse`** for every loader in the batch on each row, which mixed polymorphic / STI through rows.

2. **`ThroughAssociation`** combined `records_by_owner` from multiple through/source preloaders with **`Hash#merge`**, which **last-wins** on shared owner keys and could drop rows from the other branch.

Using **`loader.scope.object_id`** (or any blanket “split all same-SQL batches”) was too coarse: it broke query-count expectations and duplicate-preload merging for unrelated cases (e.g. **`PreloaderTest#test_preload_grouped_queries_of_middle_records`**).

## Detail

This pull request:

- **`activerecord/lib/active_record/associations/preloader/through_association.rb`**
  - Concatenate `records_by_owner` from sibling loaders instead of `Hash#merge` (with `uniq!` per owner), with a short comment on why `compare_by_identity` on the accumulator is unsafe.
  - When building inner **`through_preloaders`**, pass **`through_parent_reflection`** into **`Preloader`** only when the outer reflection is a **collection** through (**`has_many :through`**, i.e. **`reflection.collection?`**). That tags inner through batches so parallel **`has_many :through`** paths split correctly, while **`has_one :through`** does **not** tag—so it still merges with sibling **`belongs_to`** preloads that hit the same table (same SQL, one query).

- **`activerecord/lib/active_record/associations/preloader.rb`**, **`branch.rb`**, **`association.rb`**
  - Thread **`through_parent_reflection`** from the inner preloader root into branch loaders (only the **first** level under that inner tree, so nested `includes` like `:merger` are unchanged).

- **`activerecord/lib/active_record/associations/preloader/batch.rb`**
  - Extend the batch grouping key with **`loader.through_parent_reflection`** (typically `nil`; set only for inner steps of **`has_many :through`**).

- **`activerecord/test/cases/associations/includes_scoped_through_batch_test.rb`**
  - Isolated regression (temporary tables + namespaced STI models aligned with the issue).

- **`activerecord/CHANGELOG.md`**
  - Documents the behavior change and credits **Edil Talantbek uulu**.

### Tests run locally

```bash
bin/test activerecord/test/cases/associations/includes_scoped_through_batch_test.rb
bin/test activerecord/test/cases/associations_test.rb
bin/test activerecord
```

### Issue

[includes loads wrong data for scoped `has_many :through` when eager loading nested self join #52061](https://github.com/rails/rails/issues/52061)

## Checklist

- [x] Commit message has a detailed description of what changed and why, and references the issue (e.g. `[Fix #52061]`).
- [x] Tests are added or updated for the bug fix.
- [x] CHANGELOG is updated for the behavior change.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,8 +1,15 @@
-*   Fix preloads where multiple scoped `has_many :through` associations share the same query but must not share one batch loader (regression appeared as wrong STI/type rows when nesting `includes` with the same `:through`).
+*   Fix eager loading nested `includes` when several scoped parallel `has_many :through`
+    associations reuse the same through table/query.
+
+    Sibling loaders now concatenate `records_by_owner` hashes instead of using `Hash#merge`,
+    which last-winned overlapped owners. Inner through preloads additionally carry the outer
+    `has_many :through` reflection into batch partitioning so loaders that share identical SQL but
+    must not share one inverse/STI batch are split; outer reflections that are non-collection (e.g.
+    `has_one :through`) are excluded so unrelated sibling preloads stay merged into one query.
 
     Fixes #52061.
 
-    *Edilbek Talantbek uulu*
+    *Edil Talantbek uulu*
 
 *   Bump the minimum PostgreSQL version to 10.0.
 
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Fix preloads where multiple scoped `has_many :through` associations share the same query but must not share one batch loader (regression appeared as wrong STI/type rows when nesting `includes` with the same `:through`).
+
+    Fixes #52061.
+
+    *Edilbek Talantbek uulu*
+
 *   Bump the minimum PostgreSQL version to 10.0.
 
     As part of this change, `supports_pgcrypto_uuid?` is deprecated because
--- a/activerecord/lib/active_record/associations/preloader.rb
+++ b/activerecord/lib/active_record/associations/preloader.rb
@@ -96,7 +96,8 @@ module ActiveRecord
       # associations before querying the database. This can save database
       # queries by reusing in-memory objects. The optimization is only applied
       # to single associations (i.e. :belongs_to, :has_one) with no scopes.
-      def initialize(records:, associations:, scope: nil, available_records: [], associate_by_default: true)
+      def initialize(records:, associations:, scope: nil, available_records: [], associate_by_default: true,
+        through_parent_reflection: nil)
         @records = records
         @associations = associations
         @scope = scope
@@ -108,7 +109,8 @@ module ActiveRecord
           association: nil,
           children: @associations,
           associate_by_default: @associate_by_default,
-          scope: @scope
+          scope: @scope,
+          through_parent_reflection: through_parent_reflection
         )
         @tree.preloaded_records = @records
       end
--- a/activerecord/lib/active_record/associations/preloader/association.rb
+++ b/activerecord/lib/active_record/associations/preloader/association.rb
@@ -99,9 +99,10 @@ module ActiveRecord
             end
         end
 
-        attr_reader :klass
+        attr_reader :klass, :through_parent_reflection
 
-        def initialize(klass, owners, reflection, preload_scope, reflection_scope, associate_by_default)
+        def initialize(klass, owners, reflection, preload_scope, reflection_scope, associate_by_default,
+          through_parent_reflection: nil)
           @klass         = klass
           @owners        = owners.uniq(&:__id__)
           @reflection    = reflection
@@ -109,6 +110,7 @@ module ActiveRecord
           @reflection_scope = reflection_scope
           @associate     = associate_by_default || !preload_scope || preload_scope.empty_scope?
           @model         = owners.first && owners.first.class
+          @through_parent_reflection = through_parent_reflection
           @run = false
         end
 
--- a/activerecord/lib/active_record/associations/preloader/batch.rb
+++ b/activerecord/lib/active_record/associations/preloader/batch.rb
@@ -40,11 +40,14 @@ module ActiveRecord
           def group_and_load_similar(loaders)
             non_through = loaders.grep_v(ThroughAssociation)
 
+            # #52061: loaders can share the same LoaderQuery (SQL) but different
+            # Relation instances; merging their batches breaks set_inverse for
+            # parallel scoped has_many :through preloads.
             grouped = non_through.group_by do |loader|
-              [loader.loader_query, loader.klass]
+              [loader.loader_query, loader.klass, loader.scope.object_id]
             end
 
-            grouped.each do |(query, _klass), similar_loaders|
+            grouped.each do |(query, _klass, _scope_id), similar_loaders|
               query.load_records_in_batch(similar_loaders)
             end
           end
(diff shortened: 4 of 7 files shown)


Edilbek:
Related to: https://github.com/rails/rails/issues/52061

tylerwood-io:
Nice work. The problem is well-identified — parallel scoped `has_many :through` on a shared join table producing identical `LoaderQuery` keys, causing batch merging to corrupt STI/inverse rows.

A few observations from reviewing the implementation:

**The partitioning strategy is sound.** Threading `through_parent_reflection` into the batch grouping key is the right granularity — it splits parallel has_many :through branches while leaving unrelated preloads (and has_one :through) free to merge. The `reflection.collection?` guard is the key insight: has_one :through should still merge with sibling belongs_to preloads to avoid duplicate queries.

**The `merge_records_by_owner` fix is independently valuable.** `Hash#merge` last-winning on shared owner keys is a standalone bug regardless of the batch partitioning. The `concat + uniq!` approach is correct. Good call on the comment about not using `compare_by_identity` — AR row identity is fragile across loaders.

**One question on the branch threading:** `through_parent_reflection` is passed into the first level of children only (`root? ? @through_parent_reflection : nil` in `build_children`). This means nested includes like `:merger` don't carry the tag. That's the right behavior for the reproduction case, but is there a scenario where a nested through under a parallel scoped through would also need partitioning? I think the answer is no because nested includes resolve against the already-partitioned intermediate records, but wanted to confirm that reasoning.

**Food for thought on `collection?` exclusion:** The `collection?` guard means two parallel *scoped* `has_one :through` associations on the same through table wouldn't get partitioned — same bug shape, just for singular associations. Probably vanishingly rare in practice (parallel scoped `has_one :through` on a shared join), but it's the same class of issue. Not suggesting it needs to be addressed here — just noting the theoretical boundary of the fix.

**Test setup:** The self-contained temporary-table approach with `SecureRandom.hex` tokens is clean for isolation. The test correctly validates that `loaded.fandoms` returns only `Fandom` instances and `loaded.relationships` returns only `Relationship` instances, which is the exact assertion that would fail without the fix.

Edilbek:
@tylerwood-io 
Thanks for the thorough review!

On the branch threading: yes — nested includes like :merger intentionally do not carry through_parent_reflection. Only the first level under the inner through preloader is tagged, so parallel has_many :through branches are partitioned when loading the shared join. Nested associations then resolve against those already-partitioned intermediate records, so they don't need the same tag. Propagating it deeper would over-split batches without fixing additional corruption in the reported case.

On the collection? boundary: agreed — parallel scoped has_one :through on a shared join is the same class of issue in theory, but rare. Happy to leave that out of this PR and revisit if it shows up in practice.

PR #57276, opened 2026-04-30
Preserve metadata when mirroring Active Storage blobs

### LINKED ISSUE #57270: ActiveStorage::Service::MirrorService#mirror drops blob metadata when copying to mirror services
### Steps to reproduce

`ActiveStorage::Service::MirrorService#upload` correctly forwards metadata (`content_type`, `filename`, `disposition`, `custom_metadata`) to the primary service, but the asynchronous mirror copy via `MirrorService#mirror` (called from `ActiveStorage::MirrorJob#perform`) does not. Concretely, [line 75 on main](https://github.com/rails/rails/blob/main/activestorage/lib/active_storage/service/mirror_service.rb#L75) calls `service.upload key, io, checksum: checksum` with no kwargs:

```ruby
primary.open(key, checksum: checksum) do |io|
  mirrors_in_need_of_mirroring.each do |service|
    io.rewind
    service.upload key, io, checksum: checksum   # <- no metadata
  end
end
```

That reaches `S3Service#upload` (or the equivalent for Azure/GCS) with a `nil` `content_type`, so the mirrored object on S3/R2/Azure ends up with the bucket default — `application/octet-stream` on AWS S3 and Cloudflare R2.

This is invisible when blobs are served via `rails_blob_url` (the signed URL includes a `response-content-type` override) but breaks any flow that hits the bucket directly: a public CDN custom domain, an `` tag pointed at the canonical bucket URL, range requests for a video, etc. In our setup (Cloudflare R2 + custom domain CDN) every image mirrored from a Cloudinary primary lands as `application/octet-stream` and the browser refuses to render it inline.

### Expected behavior

`MirrorService#mirror` forwards the same blob metadata to each mirror's `#upload` that the primary received at original-upload time, so the mirror copy is byte-and-header equivalent.

### Actual behavior

The mirror copy gets only `checksum:`. `content_type`, `filename`, `disposition`, `custom_metadata` are dropped.

### Reproduction (bundler/inline + Minitest, Rails from main, real MirrorJob via :inline adapter)

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
end

require "active_support"
require "active_support/core_ext"
require "active_job"
require "active_storage"
require "active_storage/service"
require "active_storage/service/mirror_service"

# MirrorJob lives under app/jobs in the gem; load it manually together with
# its base class.
$LOAD_PATH.unshift File.expand_path("app/jobs", Gem.loaded_specs["activestorage"].full_gem_path)
ActiveStorage.queues = { mirror: :default }
require "active_storage/base_job"
require "active_storage/mirror_job"

# Stub ActiveStorage::Blob just enough for `MirrorJob#perform` to find a
# service via `ActiveStorage::Blob.service.try(:mirror, key, checksum:)`.
# A full Blob would require an Active Record schema; we don't need one here.
module ActiveStorage
  class Blob
    class << self
      attr_accessor :service
    end
  end
end

require "minitest/autorun"

# In-memory test double for ActiveStorage::Service that records every upload
# call so we can inspect which kwargs were forwarded.
class RecordingService < ActiveStorage::Service
  attr_reader :uploads

  def initialize
    @uploads = []
    @bytes_by_key = {}
  end

  def upload(key, io, **options)
    @uploads << { key: key, options: options }
    @bytes_by_key[key] = io.read
  end

  def open(key, **)
    yield StringIO.new(@bytes_by_key.fetch(key, ""))
  end

  def exist?(key)
    @bytes_by_key.key?(key)
  end

  def delete(key)
    @bytes_by_key.delete(key)
  end
end

class MirrorServiceMetadataTest < Minitest::Test
  PAYLOAD_METADATA = {
    content_type: "image/jpeg",
    filename:     "photo.jpg",
    disposition:  :attachment
  }.freeze

  def setup
    @primary = RecordingService.new
    @mirror  = RecordingService.new
    @service = ActiveStorage::Service::MirrorService.new(primary: @primary, mirrors: [@mirror])

    # Run the async mirror copy synchronously inside `upload`, exercising
    # ActiveStorage::MirrorJob#perform end-to-end.
    ActiveJob::Base.queue_adapter = :inline
    ActiveStorage::Blob.service = @service
  end

  def test_primary_receives_full_metadata
    @service.upload("key", StringIO.new("bytes"), checksum: "abc123", **PAYLOAD_METADATA)

    received = @primary.uploads.last[:options]
    PAYLOAD_METADATA.each do |key, value|
      assert_equal value, received[key], "expected primary to receive #{key.inspect}"
    end
  end

  def test_mirror_receives_full_metadata_via_real_mirror_job
    # primary.upload runs synchronously; MirrorJob.perform_later runs inline,
    # which calls ActiveStorage::Blob.service.mirror — i.e. our MirrorService.
    @service.upload("key", StringIO.new("bytes"), checksum: "abc123", **PAYLOAD_METADATA)

    received = @mirror.uploads.last[:options]
    PAYLOAD_METADATA.each do |key, value|
      assert_equal value, received[key],
        "expected mirror to receive #{key.inspect} (got #{received.inspect}). " \
        "BUG: MirrorService#mirror drops the metadata."
    end
  end
end
```

Output (Rails main as of today):

```
[ActiveJob] [ActiveStorage::MirrorJob] [<id>] Performing ActiveStorage::MirrorJob ...
.F

  1) Failure:
MirrorServiceMetadataTest#test_mirror_receives_full_metadata_via_real_mirror_job
expected mirror to receive :content_type (got {:checksum=>"abc123"}).
Expected: "image/jpeg"
  Actual: nil

2 runs, 4 assertions, 1 failures, 0 errors, 0 skips
```

`test_primary_receives_full_metadata` passes; `test_mirror_receives_full_metadata_via_real_mirror_job` fails. The real `ActiveStorage::MirrorJob` runs via the `:inline` ActiveJob adapter and reaches `MirrorService#mirror`, where the metadata is lost.

### Suggested fix

`MirrorService#mirror` can look up the blob by `key` (it's the canonical identifier) and forward `Blob#service_metadata` — the same hash `Blob#upload_without_unfurling` already passes to the primary. Sketch:

```ruby
def mirror(key, checksum:)
  metadata = ActiveStorage::Blob.find_by(key: key)&.send(:service_metadata) || {}

  instrument :mirror, key: key, checksum: checksum do
    mirrors_in_need_of_mirroring = mirrors.reject { |service| service.exist?(key) }
    return if mirrors_in_need_of_mirroring.empty?

    primary.open(key, checksum: checksum) do |io|
      mirrors_in_need_of_mirroring.each do |service|
        io.rewind
        service.upload(key, io, checksum: checksum, **metadata)
      end
    end
  end
end
```

Falling back to a metadata-less mirror call when no blob row matches the key preserves correctness for orphan paths.

I'm happy to send a PR with this fix + tests if the approach is acceptable.

### System configuration

- Rails: `main` (verified via the bundler/inline reproducer)
- Also confirmed on Rails 7.1.5.1 in production (same code path, identical bug)
- Ruby 3.3.6
- Affects all storage backends that respect `Content-Type` from upload (S3, R2, GCS, Azure). Disk service is unaffected because Disk doesn't store headers.

afurm:
### Motivation / Background

This Pull Request has been created because `ActiveStorage::Service::MirrorService#mirror` copies blobs from the primary service to mirror services with only the checksum. That drops upload metadata such as `content_type`, `disposition`, `filename`, and `custom_metadata`.

For services that persist those values as object headers, mirrored objects can fall back to backend defaults such as `application/octet-stream`. That breaks flows that serve mirrored objects directly from a bucket or CDN instead of through Rails-generated blob URLs.

Fixes #57270.

### Detail

This Pull Request changes `MirrorService#mirror` to look up the blob by key and forward the blob's existing `service_metadata` when uploading to mirror services. If no blob row exists for the key, mirroring still falls back to the previous metadata-less upload behavior.

The regression test uses a recording service to assert that mirror uploads receive the same service metadata shape that Active Storage passes to the primary upload path, including content type, disposition, filename, and custom
metadata.

### Additional information

Validated with:

```bash
bin/test test/service/mirror_service_test.rb
bin/test test/models/attachment_test.rb -n "/mirroring a directly-uploaded blob after attaching it/"
bin/test test/log_subscriber_test.rb -n "/service_mirror/"
bin/test test/structured_event_subscriber_test.rb -n "/service_mirror/"
bundle exec rubocop activestorage/lib/active_storage/service/mirror_service.rb activestorage/test/service/mirror_service_test.rb
git diff --check
```

No changelog entry is included because this is a minor bug fix.

### Checklist

### DIFF
--- a/activestorage/lib/active_storage/service/mirror_service.rb
+++ b/activestorage/lib/active_storage/service/mirror_service.rb
@@ -66,10 +66,12 @@ module ActiveStorage
     def mirror(key, checksum:)
       instrument :mirror, key: key, checksum: checksum do
         if (mirrors_in_need_of_mirroring = mirrors.select { |service| !service.exist?(key) }).any?
+          metadata = ActiveStorage::Blob.find_by(key: key)&.send(:service_metadata) || {}
+
           primary.open(key, checksum: checksum) do |io|
             mirrors_in_need_of_mirroring.each do |service|
               io.rewind
-              service.upload key, io, checksum: checksum
+              service.upload key, io, checksum: checksum, **metadata
             end
           end
         end
--- a/activestorage/test/service/mirror_service_test.rb
+++ b/activestorage/test/service/mirror_service_test.rb
@@ -1,8 +1,35 @@
 # frozen_string_literal: true
 
 require "service/shared_service_tests"
+require "database/setup"
 
 class ActiveStorage::Service::MirrorServiceTest < ActiveSupport::TestCase
+  class RecordingService < ActiveStorage::Service
+    attr_reader :uploads
+
+    def initialize
+      @uploads = []
+      @bytes_by_key = {}
+    end
+
+    def upload(key, io, **options)
+      @uploads << { key: key, options: options }
+      @bytes_by_key[key] = io.read
+    end
+
+    def download(key)
+      @bytes_by_key.fetch(key)
+    end
+
+    def open(key, **)
+      yield StringIO.new(download(key))
+    end
+
+    def exist?(key)
+      @bytes_by_key.key?(key)
+    end
+  end
+
   mirror_config = (1..3).to_h do |i|
     [ "mirror_#{i}",
       service: "Disk",
@@ -77,6 +104,36 @@ class ActiveStorage::Service::MirrorServiceTest < ActiveSupport::TestCase
     assert_equal data, @service.mirrors.first.download(key)
     assert_equal data, @service.mirrors.second.download(key)
     assert_equal "Surprise!", @service.mirrors.third.download(key)
+  end
+
+  test "mirroring forwards blob service metadata to secondary services" do
+    primary = RecordingService.new
+    mirror = RecordingService.new
+    service = ActiveStorage::Service::MirrorService.new(primary: primary, mirrors: [ mirror ])
+
+    data = "<!doctype html>"
+    checksum = OpenSSL::Digest::MD5.base64digest(data)
+    blob = ActiveStorage::Blob.create_before_direct_upload!(
+      key: SecureRandom.base58(24),
+      filename: "page.html",
+      byte_size: data.bytesize,
+      checksum: checksum,
+      content_type: "text/html",
+      metadata: { custom: { "tenant" => "rails" } }
+    )
+
+    primary.upload blob.key, StringIO.new(data), checksum: checksum
+
+    service.mirror blob.key, checksum: checksum
+
+    assert_equal data, mirror.download(blob.key)
+    assert_equal({
+      checksum: checksum,
+      content_type: ActiveStorage.binary_content_type,
+      disposition: :attachment,
+      filename: blob.filename,
+      custom_metadata: { "tenant" => "rails" }
+    }, mirror.uploads.last[:options])
   end
 
   test "URL generation in primary service" do


rafaelfranca:
Merged in #57595

PR #57277, opened 2026-04-30
[ci skip] Fix typo in Getting Started guide)

cnantais:
settings_profile_path typo is referenced before being defined
(passwords_controller.rb:7): On success, the redirect pointed to settings_profile_path which doesn't exist — would have raised a routing error  even if the form worked.


fatkodima:
That is correct - when password updated, redirect to profile page.

PR #57278, opened 2026-04-30
[ci skip] Update sign_up_and_settings.md – fix User validation conditions

cnantais:
The existing user was created before add_names_to_users ran, so their name columns are NULL. Without on: :create, the validation fires on every update call regardless of which attributes are changing. With this fix, names are still required at sign-up, and updating the password on an existing account won't fail due to unrelated blank columns.


fatkodima:
I think the tutorial is not expected to be in a working state after each example - it only shows some steps that need to be added one by one to have a fully complete app.

PR #57283, opened 2026-05-01
Avoid assert_nil for Mime::NullType request format

mugitti9:
### Motivation / Background
This Pull Request has been created because `request.format` returns a `Mime::NullType` when the requested format is unknown, but the test used `assert_nil` to verify that behavior.

This became incompatible with minitest/minitest@c471347, which changed `assert_nil` to compare with `nil == obj` instead of calling `obj.nil?`.

### Detail

This Pull Request changes the request format test to assert that the returned format is a `Mime::NullType` and explicitly verify its `nil?` behavior.

The existing predicate checks for `html?`, `xml?`, and `json?` are kept.

### Additional information

This is a test-only change.

### Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -344,7 +344,8 @@ GEM
       logger
     mini_mime (1.1.5)
     mini_portile2 (2.8.9)
-    minitest (6.0.0)
+    minitest (6.0.6)
+      drb (~> 2.0)
       prism (~> 1.5)
     minitest-mock (5.27.0)
     minitest-retry (0.2.3)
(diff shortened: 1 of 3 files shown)


mugitti9:
@eugeneius 
Thank you for the review and for taking care of this change.

I updated the assertions in [the commit](https://github.com/rails/rails/pull/57283/changes/bbebe6bf613b5cdcd408742a1d510d576b5d6f5a) because `assert_predicate object, :nil?` is reported by RuboCop. This keeps the test intent the same while avoiding the offense.
I confirmed that all CI checks passed with this change.

I also really appreciate your ongoing work maintaining Rails.

eugeneius:
Thank you!

PR #55799, opened 2025-09-29
[ci skip] Document Optimistic Locking vs. ActiveStorage edge case

g-pavlik:
### Motivation / Background

This Pull Request has been created because I don't believe there's a framework-level solution to #55764, but this edge case should be documented. 

### Detail

This Pull Request add a note in `ActiveRecord::Locking::Optimistic` describing an possible interference with ActionText/ActiveStorage

### Additional info

Here's a relevant screenshot of the preview: 
image
 

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* n/a Tests are added or updated if you fix a bug or add a feature.
* n/a CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

Issue #45567, opened 2022-07-11
Record not touched when ActiveStorage::Blob is analyzed


natematykiewicz:
### Steps to reproduce
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "rails", "~> 7.0.0"
  gem "sqlite3"
end

require "active_record/railtie"
require "active_storage/engine"
require "tmpdir"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  config.eager_load = false
  config.session_store :cookie_store, key: "cookie_store_key"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  config.active_storage.service = :local
  config.active_storage.service_configurations = {
    local: {
      root: Dir.tmpdir,
      service: "Disk"
    }
  }
end

ENV["DATABASE_URL"] = "sqlite3::memory:"

Rails.application.initialize!

require ActiveStorage::Engine.root.join("db/migrate/20170806125915_create_active_storage_tables.rb").to_s

ActiveRecord::Schema.define do
  CreateActiveStorageTables.new.change

  create_table :users, force: true do |t|
    t.timestamps
  end
end

class User < ActiveRecord::Base
  has_one_attached :profile
end

require "minitest/autorun"

class BugTest < Minitest::Test
  def test_upload_and_download
    user = User.create!(
      profile: {
        content_type: "text/plain",
        filename: "dummy.txt",
        io: ::StringIO.new("dummy"),
      }
    )

    old_updated_at = user.updated_at

    # Sanity test, since some database adapters truncate microseconds
    assert_equal user.reload.updated_at, old_updated_at

    # Analyze the blob
    user.profile.blob.analyze

    # User should have been touched
    refute_equal user.reload.updated_at, old_updated_at
  end
end
```

### Expected behavior
Analyzing an ActiveStorage Blob should touch all of the models that the blob is attached to.

### Actual behavior
The model is untouched.

### System configuration
**Rails version**: 6.1.6

**Ruby version**: 2.6

### Additional Context

ActiveStorage Blobs are considered immutable. The table has no `updated_at` column, and the documentation states:

[...]

PR #57284, opened 2026-05-01
Don't bump lock_version on records during blob analysis

### LINKED ISSUE #55764: Models with optimistic locking and has_rich_text are susseptible to race condition when ActiveStorage::Blob is attached and analyzed in a background job
### Steps to reproduce

1. Have a model with `has_rich_text`, optimistic locking, and background jobs (example repo: https://github.com/g-pavlik/rails-lock_version-actiontext-conflict). 
2. Run your server, but stop background jobs (this extends the window for race conditions to happen). 
3. Edit your model (its `lock_version = N`), attach an image, and save (`lock_version = N+1`)
4. Edit this model again (form loads with `lock_version = N+1`)
5. Run background jobs (`bin/jobs`) and wait for ActiveStorage::AnalyzeJob (for the image attached in the content) to finish (this eventually bumps `lock_version = N+2`)
6. Try saving the opened form - the ActiveRecord::StaleObjectError is raised (in the example repo it results in the previous version being loaded in the form)

### Expected behavior

There's no ActiveRecord::StaleObjectError issue

Though I understand why Rails sets it up this way: analyzing an attachment might cause content to present itself differently, so we need to touch the models, which will invalidate the cache. 

### Actual behavior

There's a race condition that can sometimes lead to false-positive StaleObjectErrors. 

### System configuration

**Rails version**: 8.0.2.1

**Ruby version**: ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [x86_64-linux]

### Other notes

This issue was not recognized before because in well-orchestrated systems the opportunity window for this race condition is minimal, so I understand this would probably not be high on the community's priority list to fix. 
In our case it has a big impact, because we also implemented an auto-save feature for that model (the changes are saved as the user types), and this feature is like saving and opening the form very quickly—resulting in conflicts (with oneself) every time. 

I mostly wish to communicate the issue and seek guidance. Solutions I considered: 

#### 1. `has_rich_text: ..., analyze_blobs_async: false` 

For the price of performance, avoid race conditions completely

#### 2. Disable bubbling up the touch events somewhere along the chain.

This risk cache bugs, so no-go

#### 3. Once Removed Content Model and Main Model: 

Instead of `Note  has_rich_text:` make `Note has_one: :content` and `Content belongs_to: :note; has_rich_text: ` 
This has the same issue as 2 of breaking cache

#### 4. Implement custom ActionText::RichText and ActiveStorage::Blob implementations, and force sync blob analyze.

Essentially 1, but not on the framework level 

#### 5. Force save if the lock_version indicates a conflict, but the conflict is with oneself

Maybe possible for our specific setup.

g-pavlik:
### Motivation / Background

  Fixes #55764.

  ActiveStorage::AnalyzeJob writes only to Blob#metadata, but it cascades a touch to attached records and their parents. And on records using optimistic locking, that touch bumps lock_version. A user editing a form during the brief window between attach and analyze submits with the now-stale lock_version and gets ActiveRecord::StaleObjectError, even though no real conflict exists.

  The window is small in well-orchestrated systems but reliably hit by apps that auto-save while jobs are still in flight.

  The cascade itself is intentional: it was added in 294c271062 / #45567 to invalidate caches on parent records after async analysis (without it, a request that built a cache between attach and analyze would keep serving unanalyzed views indefinitely). The lock_version bump, by contrast, is incidental: it just falls out of record.touch going through Locking::Optimistic#_update_row, and no prior PR has discussed it.

  The existing config.active_storage.touch_attachment_records = false knob (#49723) is too coarse to use as a workaround: it disables the entire cascade, reintroducing the cache bug from #45567.

###  Detail

  Adds ActiveRecord::Locking::Optimistic.preserve_lock_version_on_touch, a block helper that suppresses the lock_version bump (and the WHERE lock_version = X constraint) on touches fired inside it. `updated_at` still updates, so cache keys keep invalidating.

  Implementation:

  - _touch_row skips adding locking_column to @_touch_attr_names when the flag is set (so dirty tracking doesn't claim lock_version was touched).
  - _update_row returns super early for attempted_action == "touch" when the flag is set, falling through to Persistence#_update_row (no bump, no constraint). The attempted_action == "touch" guard ensures regular saves still go through the locking path even if the
  flag happens to be set.
  - A prepended DeferredTouch module captures intent at touch_later time onto the record instance, since belongs_to ..., touch: true defers the parent touch to before_committed! — by which point the surrounding block has already exited. The instance flag is consumed
  when the deferred touch fires.

  ActiveStorage::Blob#touch_attachments wraps its cascade with the helper, since blob analysis is exactly the case where bumping the parents' lock_version is unjustified – analysis doesn't modify any parent field.

 ### Additional information

  Related:

  - #55764 issue this PR fixes.
  - #55799  documents the edge case as-is. Not a replacement; if both merge, the doc note can be dropped.
  - #45567 / 294c271062 original motivation for the cascade (cache invalidation).
  - #49723 touch_attachment_records config; insufficient as a workaround.

  Alternatives considered:

  1. analyze_blobs_async: false per attachment. Sidesteps the race at the cost of synchronous analysis on every upload. Discussed in #55764.
  2. Disabling the cascade entirely (touch_attachment_records = false). Reintroduces #45567.
  3. Doc-only (#55799). Leaves the bug in place.

###  Checklist

### DIFF
--- a/activerecord/lib/active_record/locking/optimistic.rb
+++ b/activerecord/lib/active_record/locking/optimistic.rb
@@ -52,8 +52,34 @@ module ActiveRecord
     module Optimistic
       extend ActiveSupport::Concern
 
+      # Captures the +preserve_lock_version_on_touch+ intent on the record at
+      # +touch_later+ time, since the actual touch is deferred until the
+      # transaction commits — by which point the surrounding block has already
+      # exited.
+      module DeferredTouch # :nodoc:
+        def touch_later(*)
+          if Optimistic.preserving_lock_version_on_touch?
+            @_skip_locking_column_on_touch = true
+          end
+          super
+        end
+      end
+
       included do
         class_attribute :lock_optimistically, instance_writer: false, default: true
+        prepend DeferredTouch
+      end
+
+      def self.preserve_lock_version_on_touch # :nodoc:
+        prior = ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch]
+        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch] = true
+        yield
+      ensure
+        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch] = prior
+      end
+
+      def self.preserving_lock_version_on_touch? # :nodoc:
+        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch]
       end
 
       def locking_enabled? # :nodoc:
@@ -85,12 +111,17 @@ module ActiveRecord
         end
 
         def _touch_row(attribute_names, time)
-          @_touch_attr_names << self.class.locking_column if locking_enabled?
+          if locking_enabled? && !_skip_locking_column_on_touch?
+            @_touch_attr_names << self.class.locking_column
+          end
           super
+        ensure
+          @_skip_locking_column_on_touch = nil
         end
 
         def _update_row(attribute_names, attempted_action = "update")
           return super unless locking_enabled?
+          return super if attempted_action == "touch" && _skip_locking_column_on_touch?
 
           begin
             locking_column = self.class.locking_column
@@ -126,6 +157,10 @@ module ActiveRecord
             @attributes[locking_column] = lock_attribute_was
             raise
           end
+        end
+
+        def _skip_locking_column_on_touch?
+          @_skip_locking_column_on_touch || Optimistic.preserving_lock_version_on_touch?
         end
 
         def destroy_row
--- a/activestorage/app/models/active_storage/blob.rb
+++ b/activestorage/app/models/active_storage/blob.rb
@@ -422,14 +422,19 @@ class ActiveStorage::Blob < ActiveStorage::Record
     end
 
     def touch_attachments
-      attachments.then do |relation|
-        if ActiveStorage.touch_attachment_records
-          relation.includes(:record)
-        else
-          relation
+      # The cascade exists to invalidate cache keys; it must not bump
+      # +lock_version+ on parents, since blob analysis does not modify any
+      # field of theirs and would otherwise race with concurrent edits.
+      ActiveRecord::Locking::Optimistic.preserve_lock_version_on_touch do
+        attachments.then do |relation|
+          if ActiveStorage.touch_attachment_records
+            relation.includes(:record)
+          else
+            relation
+          end
+        end.each do |attachment|
+          attachment.touch
         end
-      end.each do |attachment|
-        attachment.touch
       end
     end
 
--- a/activerecord/test/cases/locking_test.rb
+++ b/activerecord/test/cases/locking_test.rb
@@ -199,6 +199,41 @@ class OptimisticLockingTest < ActiveRecord::TestCase
     assert_not_predicate stale_person, :saved_changes?
   end
 
+  def test_preserve_lock_version_on_touch_skips_lock_version_bump
+    p1 = Person.find(1)
+    original_lock_version = p1.lock_version
+
+    travel 1.second do
+      ActiveRecord::Locking::Optimistic.preserve_lock_version_on_touch do
+        p1.touch
+      end
+    end
+
+    assert_equal original_lock_version, p1.lock_version
+    assert_equal original_lock_version, Person.find(1).lock_version
+    assert_equal ["updated_at"], p1.saved_changes.keys
+  end
+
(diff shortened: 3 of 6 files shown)


activerecord/lib/active_record/locking/optimistic.rb
```diff
@@ -52,8 +52,46 @@ module Locking
...
+      def self.preserve_lock_version_on_touch
+        prior = ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch]
+        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch] = true
+        yield
+      ensure
+        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch] = prior
+      end
+
+      def self.preserving_lock_version_on_touch? # :nodoc:
+        !!ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch]
```
rafaelfranca:
```suggestion
        ActiveSupport::IsolatedExecutionState[:active_record_preserve_lock_version_on_touch]
```

activerecord/lib/active_record/locking/optimistic.rb
```diff
@@ -52,8 +52,46 @@ module Locking
...
+      # held by another process do not become stale.
+      #
+      # This is useful when a touch cascades from a side-effect that does not
+      # actually modify the touched record's user-facing fields — for example,
+      # Active Storage analyzing a blob attached to a record.
+      #
+      #   ActiveRecord::Locking::Optimistic.preserve_lock_version_on_touch do
+      #     post.touch # updated_at is updated, lock_version is unchanged
+      #   end
+      def self.preserve_lock_version_on_touch
```
rafaelfranca: I'd prefer to not make this public API. It is needed only because of a framework feature. I don't want application to start calling this method.

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+*   Add `ActiveRecord::Locking::Optimistic.preserve_lock_version_on_touch`.
```
rafaelfranca: Since this should be private, this changelog entry ins't neeed

activestorage/app/models/active_storage/blob.rb
```diff
@@ -396,15 +396,21 @@ def service_metadata
       end
     end
 
+    # The cascade exists to invalidate cache keys; it must not bump
+    # +lock_version+ on attachment records' parents, since blob analysis does
+    # not modify any field of theirs and would otherwise race with concurrent
+    # edits.
```
rafaelfranca: This is going to become the documentation of this method, and we don't want that. Move the comment to inside the method.

rafaelfranca (changes requested):

activerecord/lib/active_record/locking/optimistic.rb
g-pavlik: Got it! AFAIU
```
# :nodoc: 
```
is the proper way to do it, LMK if I'm off here

activestorage/app/models/active_storage/blob.rb
g-pavlik: moved

natematykiewicz:
Thank you for this PR! As the author of #45567, I've been tagged in a handful of issues over the past 4 years related to that change.

I stand by the need to `touch` the attached records for cache invalidation, but the `lock_version` incrementing has always been an unfortunate unintended side effect.

I wonder if this would resolve the Papertrail complaints as well. (The `touch` was inserting audit logs for Papertrail users, which means every record with an attachment had been "edited" within a few seconds of being inserted.)

g-pavlik:
> I stand by the need to `touch` the attached records for cache invalidation, but the `lock_version` incrementing has always been an unfortunate unintended side effect.

I agree, and must admit that LLM was very helpful with uncovering the intent behind the current behavior. Hopefully the acrobatics of "touch created_at without touching the lock_version" would be accepted by the community. 

> 
> I wonder if this would resolve the Papertrail complaints as well. (The `touch` was inserting audit logs for Papertrail users, which means every record with an attachment had been "edited" within a few seconds of being inserted.)

I'll do some searching for issues and PRs and will mention this PR if I find anything. Thanks for the pointer!
Update: the PaperTrail issue likely will still persist (because we still bump `updated_at`), more details: https://gist.github.com/g-pavlik/d87de494c59a8a8d374c6d557d41c243

rafaelfranca (approved):

PR #57286, opened 2026-05-02
Generate AGENTS.md for new Rails applications

andyw8:
### Motivation / Background

AI coding agents benefit from a brief description of an application's tech stack. Without it, agents have to infer the setup by reading multiple files, consuming unnecessary tokens.

The `AGENTS.md` convention is supported by a growing number of AI coding tools as a tool-agnostic way to provide context to agents.

### Detail

Generates an `AGENTS.md` file by default when running `rails new`, containing:

- Rails version (with API mode noted if applicable)
- JavaScript approach (with Hotwire status)
- CSS framework
- Asset pipeline
- Database (omitted if `--skip-active-record`)
- Test framework

Pass `--skip-agents-file` to suppress generation.

### Additional information

**On opt-out by default**: this is consistent with Rails' omakase philosophy. The file is small and easily deleted, similar to how `.github/ci.yml` is generated by default.

**On staleness**: the file reflects the stack at generation time and is intended to be a living document, updated as the application evolves — not just for framework changes, but to capture architectural decisions and conventions that help agents understand the application more deeply.

### Checklist

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Generate an `AGENTS.md` file for new Rails applications.
+
+    `AGENTS.md` provides context for AI coding agents about the application's
+    stack: Rails version, JavaScript approach, CSS framework, asset pipeline,
+    database, and test framework. Pass `--skip-agent-file` to suppress generation.
+
+    *Andy Waite*
+
 *   Enable Ruby `frozen_string_literal` by default.
 
     New Rails apps now include a `config/bootsnap.rb` file that enables frozen string
--- a/railties/lib/rails/generators/app_base.rb
+++ b/railties/lib/rails/generators/app_base.rb
@@ -111,6 +111,9 @@ module Rails
         class_option :skip_ci,             type: :boolean, default: nil,
                                            desc: "Skip GitHub CI files"
 
+        class_option :skip_agent_file,     type: :boolean, default: nil,
+                                           desc: "Skip AGENTS.md file"
+
         class_option :skip_kamal,          type: :boolean, default: nil,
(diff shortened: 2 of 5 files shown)


rafaelfranca:
Thank you for the pull request but we will not add this file to applications at the moment. 

We had this discussion before between ourselves. There is no value to provide this file for rails applications because most models were already trained with information how to write applications. In fact there are a few research showing that writing an agents.md with trivial information is more detrimental to the performance of the models than not having it because the information tends to get stale.

andyw8:
Hey Rafael,

Thanks for the rapid response 😁 

I think there are two aspects this:

I absolutely agree that agents file shouldn't contain things that the model is already trained on. That's why I focused on the details that are specific to the generated app, such as the chosen CSS framework, JS setup, and database.

But I do see the other perspective - the AI model can determine these facts about the app configuration, even if not as efficiently, and there is indeed a risk of staleness.

hron84:
I would like to add one more viewpoint: a template AGENTS.md could help people figuring out how to write it.

For AI agents, redundancy is not an issue, but rather just slightly rises the importance of the already known guidelines/rules. 

But for human developers - especially for those are not expert in working with AI assistance - could help a lot seeing a good starting point, a good structure for these kind of files. Rails' own AGENTS.md is not fitting because it contains a lot of Rails-specific stuff, a typical Rails application does not look like how the `rails` project is organized.

Rails itself is based on conventions. Having a convention to have an AGENTS.md in the project root could bring value for the community.

andyw8:
In case this is ever revived, here's an example `AGENTS.md` from Fizzy by 37signals/Basecamp:

https://github.com/basecamp/fizzy/blob/main/AGENTS.md

PR #42005, opened 2021-04-16
Add support for vips in the image analyzer

brenogazzola:
### Summary
Fixes: #41940 
Fixes: #41941

Adds support for using vips as an image analyzer in active storage.

### Other Information
This is pretty much my first non trivial PR, so advice on how to improve this is welcome. Just trying to get the ball running on this one, since I'd like to use this too.

Issue #41940, opened 2021-04-13
ActiveStorage image analyzer does not handle exif rotation for mirrored images


wilg:
https://github.com/rails/rails/blob/ec5c9a8fcd39335bf79540366370c3dbeb297d13/activestorage/lib/active_storage/analyzer/image_analyzer.rb#L51-L53

should include TopRight and BottomLeft as well

<img width="741" alt="Screen Shot 2021-04-12 at 7 03 50 PM" src="https://user-images.githubusercontent.com/54353/114487596-b6574500-9bc4-11eb-9d7e-98739424057c.png">

I switched to use a Vips analyzer which works correctly https://github.com/rails/rails/issues/41940

PR #57290, opened 2026-05-02
Fix image analyzer reporting wrong dimensions for mirrored EXIF orientations

bogdan:
### Motivation / Background

`rotated_image?` in both the Vips and ImageMagick image analyzers determines whether to swap width/height when storing blob metadata. The matched orientation set is wrong in two ways — it incorrectly swaps dimensions for EXIF orientations 2 and 4 (flip-only, no rotation), and fails to swap for orientations 5 and 7 (flip + 90° rotation, which do change portrait/landscape).

This means images shot with a mirrored EXIF orientation have their width and height reported incorrectly in `blob.metadata` — a landscape image is recorded as portrait and vice versa.

Related: #41940. The bug was introduced in #42005, which added orientations `TopRight`/`BottomLeft` (2 and 4) to fix the originally missing mirrored cases, but those two are flip-only and should not trigger a swap. Orientations `LeftTop`/`RightBottom` (5 and 7), which do involve a 90° rotation and do require a swap, were never added.

### Detail

This Pull Request changes `rotated_image?` in both analyzers to match only the four EXIF orientations that involve a 90° or 270° rotation (5, 6, 7, 8), which are the only ones that require swapping width and height:

| EXIF | Transform | Should swap? | Before | After |
|------|-----------|:---:|:---:|:---:|
| 2 `Top-right` | Horizontal flip | No | swapped ✗ | unchanged ✓ |
| 4 `Bottom-left` | Vertical flip | No | swapped ✗ | unchanged ✓ |
| 5 `Left-top` | Flip + rotate 270° CW | Yes | unchanged ✗ | swapped ✓ |
| 6 `Right-top` | Rotate 90° CW | Yes | swapped ✓ | swapped ✓ |
| 7 `Right-bottom` | Flip + rotate 90° CW | Yes | unchanged ✗ | swapped ✓ |
| 8 `Left-bottom` | Rotate 270° CW | Yes | swapped ✓ | swapped ✓ |

Two fixture images are added (EXIF orientation 2 and 5) with tests covering both analyzers, which fail on the unpatched code.

### Additional information

Note that thumbnail/variant generation is unaffected — vips `auto_rot` correctly handles all eight EXIF orientations when producing output pixels. The bug is isolated to the metadata stored in `blob.metadata[:width]` and `blob.metadata[:height]`.

### Checklist


bogdan:
cc @wilg @brenogazzola, please provide your opinion.

Edgarruiz8585:
>  @Edgarruuz8585

Edgarruiz8585:
> ***@***.***

PR #57292, opened 2026-05-03
fix: ActiveModel::Attributes method override precedence regression

### LINKED ISSUE #57281: Rails 8.1 regression: deferred ActiveModel::Attributes method generation changes override precedence
### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "8.1.0"
  # This passes on Rails 8.0.3:
  # gem "rails", "8.0.3"
  gem "minitest"
end

require "active_model/railtie"
require "minitest/autorun"

  config.eager_load = false
end

Rails.application.initialize!

module Overrides
  extend ActiveSupport::Concern

  def test
    "overridden"
  end
end

class MyClassOverrideBeforeAttribute
  include ActiveModel::Model
  include Overrides

  attribute :test, :string
end

class MyClassOverrideAfterAttribute
  include ActiveModel::Model

  attribute :test, :string

  include Overrides
end

  def test_override_before_attribute_definition
    assert_equal "overridden", MyClassOverrideBeforeAttribute.new.test
  end

  def test_override_after_attribute_definition
    assert_equal "overridden", MyClassOverrideAfterAttribute.new.test
  end
end
```

### Expected behavior

Both tests should pass.

On Rails 8.0.3, both classes return `"overridden"` from `#test`, regardless of whether the `Overrides` concern is included before or after the `attribute :test, :string` declaration.

### Actual behavior

On Rails 8.1.0, `MyClassOverrideBeforeAttribute.new.test` returns `nil` instead of `"overridden"`, while `MyClassOverrideAfterAttribute.new.test` still returns `"overridden"`.

From tracing the change, this appears to be caused by `ActiveModel::Attributes` no longer being fully loaded when the module is included in the class. Instead, the relevant behavior is loaded later when the `attribute` class method is encountered. That changes the effective module load order, so in the `MyClassOverrideBeforeAttribute` case the generated attribute methods are defined after the override method and take precedence over it.

This appears to be a behavior change in `ActiveModel::Attributes` method precedence between Rails 8.0 and 8.1. I could not find this called out in the Rails 8.1 release notes as a breaking change.

This may relate to https://github.com/rails/rails/pull/54431, which appears to be the change that introduced the new loading behavior.

### System configuration

**Rails version**: 8.1.0 (failing), 8.0.3 (passing)

**Ruby version**: ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin24]

notorious94:
Fixes #57281.

  In Rails 8.1, modules included between `include ActiveModel::Attributes`
  and the first `attribute` declaration no longer override generated
  attribute methods, even though they did in 8.0.

  ```ruby
  module Overrides
    extend ActiveSupport::Concern
    def test = "overridden"
  end

  class Klass
    include Overrides
    attribute :test, :string
  end

  Klass.new.test # 8.0: "overridden", 8.1: nil
  ```

### Cause

  #54431 deferred creation of the anonymous generated_attribute_methods module so it isn't built when nothing will be put in it. That removed an implicit side effect that was load-bearing for ancestor-chain ordering: ActiveModel::Attributes's included do calls attribute_method_suffix, which calls undefine_attribute_methods, which previously called generated_attribute_methods and thus eagerly inserted the module into the class's ancestor chain at include time. After #54431, the module is first inserted later, when the first attribute declaration runs, so any modules included in between sit below the generated module and get shadowed by the generated reader/writer.

### Fix

Explicitly reserve the slot in ActiveModel::Attributes's included do block, so its position is fixed at include time, regardless of when the attribute is first called:

  ```ruby
  included do
    attribute_method_suffix "=", parameters: "value"
    send(:generated_attribute_methods)
  end
  ```

This is scoped to ActiveModel::Attributes only. Active Record uses `ActiveModel::Attributes::Normalization` (not ActiveModel::Attributes) and has its own named-module mechanism via initialize_generated_modules, so the test_no_anonymous_modules invariant from #54431 is preserved. Users of bare ActiveModel::AttributeMethods retain #54431's deferred behavior.

### Tests

Added a regression test in `activemodel/test/cases/attributes_test.rb` covering both override-before and override-after ordering.


fatkodima:
@notorious94 How would you explain malware in this PR?

PR #57293, opened 2026-05-03
[RF-Docs] Accessibility guide

brunoprietog:
### Motivation / Background

Rails has not had an accessibility guide. This adds one: a Rails-flavored take on [WCAG 2.2](https://www.w3.org/TR/WCAG22/).

### Detail

This adds a new top-level guide at `guides/source/accessibility.md` and registers it under the *Views* group in `documents.yaml`.

It covers:

* Why accessibility matters.
* What assistive technologies are and how people use them to navigate the web.
* Why semantic HTML is the foundation of accessibility.
* When and how to use ARIA to fill the gaps that native HTML leaves.
* How to structure pages with language, titles, landmarks, skip links, and headings.
* How to build accessible links, buttons, multimedia, tables, and forms.
* How to use native HTML for common patterns like disclosures, dialogs, popovers, and tooltips.
* How to write CSS that keeps focus, target size, contrast, zoom, and user preferences in check.
* How to manage focus, announce updates, signal in-flight submissions, and confirm destructive actions in apps driven by Turbo.
* How to test for accessibility with automated tools and with the same input methods real users rely on.

### Additional information

Built locally with `bundle exec rake guides:generate:html ONLY=accessibility` and validated with `bundle exec rake guides:validate`.

### Checklist

### DIFF
--- a/guides/source/accessibility.md
+++ b/guides/source/accessibility.md
@@ -854,21 +854,9 @@ Two related attributes control what gets announced:
   Override it only when a removal is genuinely meaningful on its own.
 
 For example, a button that changes its visible text to reflect the current state
-needs a live region so screen readers announce the new label. A Stimulus
-controller manages the state with a value, and CSS shows the appropriate text:
-
-```js
-// app/javascript/controllers/swap_controller.js
-import { Controller } from "@hotwired/stimulus"
-
-export default class extends Controller {
-  static values = { pressed: Boolean }
-
-  toggle() {
-    this.pressedValue = !this.pressedValue
-  }
-}
-```
+needs a live region so screen readers announce the new label. The markup holds
+both labels, CSS shows the one for the current state, and a Stimulus controller
+toggles that state:
 
 ```html+erb
 <%= tag.button type: "button",
@@ -887,6 +875,19 @@ export default class extends Controller {
 [data-swap-pressed-value="false"] [data-swap-target="secondary"],
 [data-swap-pressed-value="true"] [data-swap-target="primary"] {
   display: none;
+}
+```
+
+```js
+// app/javascript/controllers/swap_controller.js
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { pressed: Boolean }
+
+  toggle() {
+    this.pressedValue = !this.pressedValue
+  }
 }
 ```
 
@@ -943,6 +944,11 @@ space-separated list of tokens, like `aria-labelledby` and `aria-describedby`,
 passing an array joins the entries with a single space and drops `nil` or
 `false` entries. This keeps references to other elements declarative even when
 the list depends on the state of the record.
+
+```ruby
+tag.span aria: { labelledby: [ "title", false, "subtitle" ] }
+# => <span aria-labelledby="title subtitle"></span>
+```
 
 ### The First Rule of ARIA
 
--- a/guides/source/accessibility.md
+++ b/guides/source/accessibility.md
@@ -128,9 +128,8 @@ the web:
 * **Screen magnifiers:** Programs that enlarge portions of the screen for people
   with low vision. Users see only a small area of the page at a time and
   navigate by panning around. Good structure, consistent layouts, and clear
-  focus indicators help magnifier users orient themselves. Windows includes
-  Magnifier, macOS and iOS/iPadOS include Zoom, and Android includes
-  Magnification.
+  focus indicators help magnifier users orient themselves. macOS and iOS/iPadOS
+  include Zoom, Windows includes Magnifier, and Android includes Magnification.
 
 * **Voice control:** Software that allows users to speak commands instead of
   using a keyboard or mouse. A user might say "click Submit" or "click Email" to
@@ -148,7 +147,8 @@ the web:
 
 * **Hands-free pointing:** Head and eye trackers let people with significant
   motor limitations control a cursor without a mouse. Activation usually relies
-  on dwell-to-click, so adequate target size and pointer-friendly interactions
+  on dwell-to-click, where holding the pointer still over a target for a moment
+  triggers a click, so adequate target size and pointer-friendly interactions
   matter most for these users. macOS includes Head Pointer, iOS and iPadOS
   include Eye Tracking, and Windows offers Eye Control with compatible hardware.
 
@@ -241,11 +241,11 @@ between, then stop at whatever they want to read in full.
 The categories readers jump through are consistent everywhere: headings (and
 specific heading levels), landmark regions, links, buttons, form fields, tables,
 graphics, and lists. How each category is triggered differs per reader, though.
-NVDA and Orca assign one letter of the alphabet to each category: pressing that
-letter jumps to the next element of that type. VoiceOver on macOS calls the
-feature the **Rotor**, a category picker the user opens to choose what to step
-through, and also offers a single-letter mode called **Quick Nav** that behaves
-like NVDA and Orca.
+Screen readers NVDA and Orca assign one letter of the alphabet to each category:
+pressing that letter jumps to the next element of that type. VoiceOver on macOS
+calls the feature the **Rotor**, a category picker the user opens to choose what
+to step through, and also offers a single-letter mode called **Quick Nav** that
+behaves like NVDA and Orca.
 
 This is why HTML structure matters so much. Quick navigation only works when the
 page has the structure it jumps through. If a page has no headings, the heading
@@ -331,12 +331,16 @@ behavior that the `<button>` provides. A native `<button>`:
 * Submits its parent form when pressed.
 * Shows a native focus indicator.
 
-Some of this can be added manually: `tabindex="0"` for focus, `role="button"`
-for screen reader announcement, JavaScript handlers for Enter, Space, and click.
-But the browser provides many subtle behaviors for free, and it is easy to miss
-one: form submission, `disabled` state management, focus handling, interaction
-with screen reader modes, and more. The native element is what guarantees the
-keyboard and name/role/value plumbing that WCAG requires ([WCAG 2.1.1
+Some of this can be added manually:
+
+* `tabindex="0"` for focus.
+* JavaScript handlers for Enter, Space, and click.
+* `role="button"` for screen reader announcement.
+
+But the browser provides many subtle behaviors for free: form submission,
+`disabled` state management, focus handling, interaction with screen reader
+modes, and more. It is easy to miss one. The native element is what guarantees
+the keyboard and name/role/value plumbing that WCAG requires ([WCAG 2.1.1
 Keyboard][wcag-keyboard], [WCAG 4.1.2 Name, Role, Value][wcag-name-role-value]).
 
 [wcag-keyboard]: https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html
@@ -392,6 +396,15 @@ directly as a string:
 The screen reader now announces "Delete article, button" instead of just
 "button."
 
+NOTE: The HTML [`title`][title-attr] attribute is sometimes used for the same
+purpose, but it is unreliable as the sole source of an accessible name. Browsers
+only reveal it as a tooltip on mouse hover, so touch and keyboard users cannot
+see it, and screen reader support varies. It can be useful as an additional hint
+on an element that already has a visible label or an `aria-label`, but do not
+rely on it to name a control.
+
+[title-attr]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/title
+
 Sometimes an element needs a name that is already visible on the page, for
 example when a page has multiple `<nav>` elements and screen reader users need
 to tell them apart. [`aria-labelledby`][aria-labelledby] solves this by
@@ -420,9 +433,11 @@ voice control users would say ([WCAG 2.5.3 Label in Name][wcag-label-in-name]).
 
 [wcag-label-in-name]: https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html
 
-The `aria-labelledby` attribute accepts more than one ID separated by spaces.
+The `aria-labelledby` attribute accepts more than one ID, separated by spaces.
 This lets the accessible name compose from several existing pieces of page
 content without duplicating strings.
+
+#### Naming Best Practices
 
 When an element does have visible text, any `aria-label` or `aria-labelledby`
 applied to it must contain that visible text verbatim, ideally as a prefix.
@@ -446,15 +461,6 @@ conflict with how the user has configured their reader. Name the control, trust
 the native announcement, and rewrite the name itself if it causes friction in
 pronunciation.
 
-NOTE: The HTML [`title`][title-attr] attribute is sometimes used for the same
-purpose, but it is unreliable as the sole source of an accessible name. Browsers
-only reveal it as a tooltip on mouse hover, so touch and keyboard users cannot
-see it, and screen reader support varies. It can be useful as an additional hint
-on an element that already has a visible label or an `aria-label`, but do not
-rely on it to name a control.
-
-[title-attr]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/title
-
 #### Naming Limitations and Visually Hidden Text
 
 `aria-label` and `aria-labelledby` only work on elements whose **role** supports
@@ -462,16 +468,27 @@ naming: interactive elements, landmarks, and other identifiable roles like
 `<img>`, `<table>`, `<dialog>`, `group`, or `list`. They are **prohibited** on
 elements with a role that does not support naming. A plain `<div>`, `<span>`, or
 `<p>` has a role (`generic` or `paragraph`) that prohibits naming, so
-`aria-label` on these elements is invalid and screen readers will ignore it.
-However, the restriction is on the *role*, not the *element*: `<div role="group"
-aria-label="...">` is valid because the `group` role supports naming, even
-though the underlying element is a `<div>`. The WAI-ARIA specification lists all
-[roles supporting name from
-author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor).
+`aria-label` on these elements is invalid. However, the restriction is on the
+*role*, not the *element*: `<div role="group" aria-label="...">` is valid
+because the `group` role supports naming, even though the underlying element is
+a `<div>`. The WAI-ARIA specification lists all [roles supporting
+naming](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor).
 
-This means that on a generic element like a paragraph or a `<div>`, `aria-label`
-is ignored entirely. And on elements whose role does accept it, the attribute
-*replaces* the accessible name rather than adding to it.
+A notification badge is a common case where this goes wrong. Sighted users see a
+number beside a bell icon, but the number alone means nothing to a screen
+reader, and the instinct is to reach for an `aria-label` on the `<span>`.
+
+```html+erb
+<span aria-hidden="true">🔔</span>
+<span aria-label="unread notifications">3</span>
+```
+
+A plain `<span>` has the `generic` role, which is not meant to carry a label,
+making this `aria-label` invalid. Screen readers handle it inconsistently, and
+the "unread notifications" label reaches some users while others get only "3"
+with no hint of what it counts, so the context cannot be relied on. On an
+element whose role *does* support naming the attribute fails the other way,
+replacing the accessible name instead of adding to it.
 
 The solution is **visually hidden text**, a `<span>` that is visible to screen
 readers but hidden from sighted users with CSS. This is useful for adding
@@ -490,8 +507,6 @@ does not support naming, or when the context is part of a larger text flow:
 ```
 
 ```html+erb
-<%# A notification badge: sighted users see "3" next to a bell icon,
-    but a screen reader needs to know what "3" means %>
 <span aria-hidden="true">🔔</span>
 <span>3<span class="visually-hidden"> unread notifications</span></span>
 ```
@@ -559,8 +574,8 @@ visible on the page, such as a hint below a form field:
 The screen reader announces the label, the field type, and the description, for
 example: "Password, secure text field, Must be at least 12 characters."
 
-Like `aria-labelledby`, `aria-describedby` accepts more than one ID separated by
-spaces, so a description can compose from several existing pieces of page
+Like `aria-labelledby`, `aria-describedby` accepts more than one ID, separated
+by spaces, so a description can compose from several existing pieces of page
 content without duplicating strings.
 
 NOTE: Like `aria-label`, descriptions only work reliably on elements whose role
@@ -777,7 +792,9 @@ export default class extends Controller {
 When using `aria-pressed`, do not change the button's text to reflect the state.
 If the button says "Like" when unpressed and "Unlike" when pressed, the screen
 reader announces "Unlike, pressed", which contradicts itself. Keep the text
-stable and let `aria-pressed` carry the state.
+stable and let `aria-pressed` carry the state. If the design instead needs the
+label itself to change, that is also a valid approach, covered in [Announcing
+Dynamic Changes with `aria-live`](#announcing-dynamic-changes-with-aria-live).
 
 Also avoid toggle verbs like "Toggle sidebar" in the label: saying "toggle"
 duplicates what `aria-pressed` already communicates. Prefer action verbs like
@@ -822,8 +839,8 @@ Two related attributes control what gets announced:
 
 * [`aria-atomic="true"`][aria-atomic]: Announces the **entire** live region
   content when any part changes, not just the changed part. Useful when the
-  change only makes sense in context, for example a clock showing "12:35 PM"
-  where announcing just "35" would be meaningless.
+  change only makes sense in context. For example, a clock changing to "12:35
+  PM" should announce "12:35 PM", not just "35".
 * [`aria-relevant`][aria-relevant]: Controls which types of changes trigger
   announcements. Values can be combined (space-separated):
   * `additions`: New elements added to the region.
@@ -942,8 +959,10 @@ handle all other button behaviors by hand. And as the [Semantic
 HTML](#semantic-html) section explained, the result rarely matches the quality
 of the native element.
 
-This is why the most important rule of ARIA is: **if a native HTML element
+This is why [the W3C's first rule][using-aria] is: **if a native HTML element
 provides the semantics and behavior required, use it instead of adding ARIA.**
+
+[using-aria]: https://www.w3.org/TR/using-aria/#rule1
 
 NOTE: Incorrect ARIA is worse than no ARIA at all. The [WebAIM Million
 report](https://webaim.org/projects/million/) finds year after year that home
--- a/guides/source/accessibility.md
+++ b/guides/source/accessibility.md
@@ -1,0 +1,5028 @@
+**DO NOT READ THIS FILE ON GITHUB, GUIDES ARE PUBLISHED ON
+<https://guides.rubyonrails.org>.**
+
+Accessibility
+=============
+
+This guide covers how to build accessible web applications with Rails.
+Accessibility (often abbreviated as "a11y") means making applications usable by
+everyone, regardless of their abilities or how they interact with a computer.
+
+After reading this guide, you will know:
+
+* Why accessibility matters.
+* What assistive technologies are and how people use them to navigate the web.
+* Why semantic HTML is the foundation of accessibility.
+* When and how to use ARIA to fill the gaps that native HTML leaves.
+* How to structure pages with language, titles, landmarks, skip links, and
+  headings.
+* How to build accessible links, buttons, multimedia, tables, and forms.
+* How to use native HTML for common patterns like disclosures, dialogs,
+  popovers, and tooltips.
+* How to write CSS that keeps focus, target size, contrast, zoom, and user
+  preferences in check.
+* How to manage focus, announce updates, signal in-flight submissions, and
+  confirm destructive actions in apps driven by Turbo.
+* How to test for accessibility with automated tools and with the same input
+  methods real users rely on.
+
+--------------------------------------------------------------------------------
+
+Why Accessibility Matters
+-------------------------
+
+Accessibility means building applications that work for people regardless of how
+they use a computer: keyboard only, screen reader, magnifier, voice control,
+switch device, or any combination.
+
+Most of the web does not. The [WebAIM Million][webaim-million] report audits the
+top one million home pages every year and consistently finds detectable WCAG
+failures on around 95% of them. The [World Health Organization
+estimates][who-disability] that roughly one in six people lives with a
+significant disability, so the gap between that audience and the inaccessible
+state of the web is large.
+
+This guide is a practical reference for closing that gap in a Rails application.
+It covers the specific mistakes that lock out part of an audience, and it shows
+where Rails helpers and modern HTML already do the work. Building accessibility
+in from the start costs little; retrofitting it later is the expensive option.
+
+[webaim-million]: https://webaim.org/projects/million/
+[who-disability]: https://www.who.int/news-room/fact-sheets/detail/disability-and-health
+
+### The Web Content Accessibility Guidelines (WCAG)
+
+The [Web Content Accessibility Guidelines
+(WCAG)](https://www.w3.org/WAI/standards-guidelines/wcag/) are the
+internationally recognized standard for web accessibility. They are organized
+around [four
+principles](https://www.w3.org/WAI/fundamentals/accessibility-principles/),
+often summarized with the mnemonic **POUR**:
+
+* **Perceivable**: Information and user interface components must be presentable
+  to users in ways they can perceive (text alternatives for images, captions for
+  video, sufficient color contrast).
+* **Operable**: User interface components and navigation must be operable
+  (keyboard accessible, enough time to interact, no content that causes
+  seizures).
+* **Understandable**: Information and the operation of user interface must be
+  understandable (readable text, predictable behavior, help with errors).
+* **Robust**: Content must be robust enough that it can be interpreted reliably
+  by a wide variety of user agents, including assistive technologies.
+
+The widely accepted baseline for web applications is conformance with [WCAG 2.2
+Level AA](https://www.w3.org/TR/WCAG22/). It was also adopted as an ISO standard
+([ISO/IEC 40500:2025](https://www.iso.org/standard/91029.html)) in October 2025,
+consolidating its role as the de facto reference.
+
+This guide is not an exhaustive walkthrough of every WCAG success criterion.
+Many of them are covered directly, some are cited in passing where they come up,
+and a few fall outside the scope of a typical Rails application. Treat the
+criteria referenced throughout as pointers into the specification rather than a
+checklist, and consult [WCAG 2.2](https://www.w3.org/TR/WCAG22/) when a project
+needs formal conformance.
+
+### Legal Requirements
+
+Web accessibility is also a legal requirement in many jurisdictions, typically
+through regional standards that incorporate WCAG. In the **European Union**, the
+[European Accessibility Act][eaa] applies to a defined range of consumer
+products and services, and compliance is demonstrated against the harmonized [EN
+301 549][en-301] standard. In the **United States**, the [Americans with
+Disabilities Act][ada] applies to government and private websites, and [Section
+508][s508] of the Rehabilitation Act sets WCAG-based requirements for federal
+agencies.
+
+Many other countries maintain similar frameworks. The W3C Web Accessibility
+Initiative publishes a [comprehensive list of accessibility policies by
+country](https://www.w3.org/WAI/policies/), which is the right starting point
+for any specific jurisdiction.
+
+[eaa]: https://commission.europa.eu/strategy-and-policy/policies/justice-and-fundamental-rights/disability/european-accessibility-act-eaa_en
+[en-301]: https://digital-strategy.ec.europa.eu/en/policies/web-accessibility-directive-standards-and-harmonisation
+[ada]: https://www.ada.gov/
+[s508]: https://www.access-board.gov/ict/
+
+How People Use the Web
+----------------------
+
+Before writing any code, it helps to understand how different people interact
+with web applications. Not everyone uses a mouse, a monitor, or a touchscreen in
+the same way, and some do not use them at all.
+
+### Assistive Technologies
+
+An **assistive technology** is any tool that helps a person interact with a
+computer. The categories below are the ones that matter most when building for
+the web:
+
+* **Screen readers:** Programs that convert screen content into speech or
+  braille output. A blind or visually impaired user listens to a synthetic voice
+  that announces text, headings, links, buttons, and form controls as they
+  navigate. Screen readers do not "see" the page; they read the underlying HTML
+  structure. Popular screen readers include [NVDA](https://www.nvaccess.org/)
+  (Windows), [VoiceOver][vo-mac-guide] (built into macOS),
+  [Orca](https://orca.gnome.org/) (Linux), [VoiceOver][vo-ios-guide] (built into
+  iOS and iPadOS), and [TalkBack][talkback-guide] (built into Android).
+
+* **Screen magnifiers:** Programs that enlarge portions of the screen for people
+  with low vision. Users see only a small area of the page at a time and
+  navigate by panning around. Good structure, consistent layouts, and clear
+  focus indicators help magnifier users orient themselves. Windows includes
+  Magnifier, macOS and iOS/iPadOS include Zoom, and Android includes
+  Magnification.
+
+* **Voice control:** Software that allows users to speak commands instead of
+  using a keyboard or mouse. A user might say "click Submit" or "click Email" to
+  interact with buttons and form fields. For this to work, every interactive
+  element needs a visible text label: if a button has no text, voice control
+  users cannot refer to it by name. Voice control is built into all major
+  platforms: Voice Control on macOS/iOS, Voice Access on Windows and Android.
+
+* **Switch devices:** Physical devices used by people with significant motor
+  limitations. A switch device might be a single button or a sip-and-puff
+  device. The OS scans the interface and highlights items in turn, and the user
+  presses the switch to activate the current target. Keyboard accessibility and
+  a clean focus order are essential for switch users. Switch Control is built
+  into macOS, iOS, and iPadOS, and Switch Access is built into Android.
+
+* **Hands-free pointing:** Head and eye trackers let people with significant
+  motor limitations control a cursor without a mouse. Activation usually relies
+  on dwell-to-click, so adequate target size and pointer-friendly interactions
+  matter most for these users. macOS includes Head Pointer, iOS and iPadOS
+  include Eye Tracking, and Windows offers Eye Control with compatible hardware.
+
+[vo-mac-guide]: https://support.apple.com/guide/voiceover/welcome/mac
+[vo-ios-guide]: https://support.apple.com/guide/iphone/turn-on-and-practice-voiceover-iph3e2e415f/ios
+[talkback-guide]: https://support.google.com/accessibility/android/answer/6283677
+
+Everything in an application (every link, button, form field, dialog, and menu)
+must be reachable and operable with the **keyboard alone**. This is the
+foundation that all of these technologies depend on. Screen readers, voice
+control, and switch devices all ultimately rely on the same underlying keyboard
+and focus model.
+
+### How Screen Readers Work
+
+Screen readers are the assistive technology most affected by how HTML is
+written, so they deserve a closer look.
+
+A screen reader does **not** see the page the way a sighted user does. Instead,
+it reads the [**accessibility tree**][accessibility-tree], a structured
+representation of the page that the browser builds from the HTML. That tree is
+to screen readers what the visual layout is to sighted users.
+
+[accessibility-tree]: https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree
+
+Content is read in the **order it appears in the HTML source**, from top to
+bottom. If CSS visually reorders elements (with `order`, `position`, or `grid`
+placement), screen reader users will hear the content in a different order than
+sighted users see it. This disconnect can make pages confusing or unusable
+([WCAG 1.3.2 Meaningful Sequence][wcag-meaningful-sequence]). The HTML source
+order is the truth for assistive technologies.
+
+[wcag-meaningful-sequence]: https://www.w3.org/WAI/WCAG22/Understanding/meaningful-sequence.html
+
+When the HTML is **semantic**, meaning the correct elements are used for their
+intended purpose, the browser creates a rich accessibility tree. It knows that a
+`<button>` is a button, a `<nav>` is navigation, an `<h2>` is a second-level
+heading. The screen reader can then announce these elements accurately and let
+the user navigate between them efficiently.
+
+Without that semantic structure, when the page is built from `<div>` and
+`<span>` elements alone, the accessibility tree is flat and meaningless, like a
+wall of unstyled text. The screen reader cannot distinguish a heading from a
+paragraph, or a button from plain text.
+
+Everything above applies to every screen reader, on every platform: they all
+traverse the same accessibility tree. What changes between devices is how the
+user drives the screen reader. On desktop, navigation is driven by the keyboard.
+On mobile, by touch gestures.
+
+#### On Desktop
+
+Screen reader users **do not** navigate pages by pressing Tab from top to
+bottom. The Tab key only moves between things a user can click or type into,
+known as **interactive elements**. On the web these are links, buttons, and form
+controls. Everything else (headings, paragraphs, images, lists, tables, and
+ordinary text) is not a Tab stop and is skipped entirely. Pressing it repeatedly
+on a typical article page jumps between navigation links and form fields while
+the article body itself stays out of reach. A screen reader user who only had
+Tab available would never read the article.
+
+To make the rest of the page reachable, the screen reader **intercepts keyboard
+input** and reassigns keys to serve as navigation commands. When one is running,
+pressing `H` does not type the letter "h"; it jumps to the next heading.
+Pressing `D` jumps to the next landmark region. The reader takes over the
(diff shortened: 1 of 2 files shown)


guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+When the design changes the button text on toggle (for example, "Show completed"
+and "Hide completed") instead of carrying `aria-pressed`, the label itself is
+what indicates the state. [Announcing Dynamic Changes with
+`aria-live`](#announcing-dynamic-changes-with-aria-live) covers how to make that
+change reach the screen reader.
+
+TIP: For settings or preferences that toggle between on and off, consider using
+a native `<input type="checkbox" switch>` instead. It renders as a toggle switch
+and the screen reader announces its on/off state automatically. In browsers that
+do not yet support the `switch` attribute, it falls back to a regular checkbox.
```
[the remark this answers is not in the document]
brunoprietog: See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#switch

brunoprietog:
cc @seanpdoyle

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+* Is announced as "button" by screen readers.
+* Can be activated by voice control users saying "click Save."
+* Submits its parent form when pressed.
+* Shows a native focus indicator.
+
+Some of this can be added manually: `tabindex="0"` for focus, `role="button"`
+for screen reader announcement, JavaScript handlers for Enter, Space, and click.
+But the browser provides many subtle behaviors for free, and it is easy to miss
+one: form submission, `disabled` state management, focus handling, interaction
+with screen reader modes, and more. The native element is what guarantees the
```
p8: Nitpick. I first read this as "form submission" being the easy one to miss: "But the browser provides many subtle behaviors for free, and it is easy to miss one: form submission."

Maybe first list the behaviors and move the "easy to miss" part to a new sentence?
```suggestion
But the browser provides many subtle behaviors for free: form submission, `disabled` state management, focus handling, interaction
with screen reader modes, and more. It is easy to miss one. The native element is what guarantees the
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+
+* Is focusable with the keyboard (Tab key).
+* Can be activated with Enter or Space.
+* Is announced as "button" by screen readers.
+* Can be activated by voice control users saying "click Save."
+* Submits its parent form when pressed.
+* Shows a native focus indicator.
+
+Some of this can be added manually: `tabindex="0"` for focus, `role="button"`
+for screen reader announcement, JavaScript handlers for Enter, Space, and click.
```
p8: Nitpick. This sentence was a bit difficult to read for me.

Maybe make it a list as well? This could also make it easier to compare the native behavior with the manual behavior:
```suggestion
Some of this can be added manually:

* `tabindex="0"` for focus
* JavaScript handlers for Enter, Space, and click.
* `role="button"` for screen reader announcement

```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+naming text is already visible on the page. It reuses existing content and keeps
+a single source of truth, so if someone updates the heading text, the accessible
+name updates automatically. With `aria-label`, it is common for someone to
+change the visible text and forget to update the attribute, leaving the
+accessible name out of sync and the visible label no longer included in the name
+voice control users would say ([WCAG 2.5.3 Label in Name][wcag-label-in-name]).
+
+[wcag-label-in-name]: https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html
+
+The `aria-labelledby` attribute accepts more than one ID separated by spaces.
```
p8: Nitpick. Adding a comma would make it clearer "separated by spaces" applies to "more than one ID" instead of "one ID seperated by spaces":
```suggestion
The `aria-labelledby` attribute accepts more than one ID, separated by spaces.
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  (Windows), [VoiceOver][vo-mac-guide] (built into macOS),
+  [Orca](https://orca.gnome.org/) (Linux), [VoiceOver][vo-ios-guide] (built into
+  iOS and iPadOS), and [TalkBack][talkback-guide] (built into Android).
+
+* **Screen magnifiers:** Programs that enlarge portions of the screen for people
+  with low vision. Users see only a small area of the page at a time and
+  navigate by panning around. Good structure, consistent layouts, and clear
+  focus indicators help magnifier users orient themselves. Windows includes
+  Magnifier, macOS and iOS/iPadOS include Zoom, and Android includes
+  Magnification.
```
p8: Maybe start with MacOS first, to clarify that we are talking about Windows the OS, not a browser window?
```suggestion
  focus indicators help magnifier users orient themselves. MacOS and iOS/iPadOS include
  Zoom, Windows includes Magnifier,  and Android includes
  Magnification.
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+* **Switch devices:** Physical devices used by people with significant motor
+  limitations. A switch device might be a single button or a sip-and-puff
+  device. The OS scans the interface and highlights items in turn, and the user
+  presses the switch to activate the current target. Keyboard accessibility and
+  a clean focus order are essential for switch users. Switch Control is built
+  into macOS, iOS, and iPadOS, and Switch Access is built into Android.
+
+* **Hands-free pointing:** Head and eye trackers let people with significant
+  motor limitations control a cursor without a mouse. Activation usually relies
+  on dwell-to-click, so adequate target size and pointer-friendly interactions
```
p8: I'm not familiar with dwell-to-click. Maybe that should be explained here?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+reader user the same ability in audio: a single keystroke jumps to the next
+heading on the page and announces it; another keystroke jumps to the next one;
+another to the next landmark region; another to the next link. The user can
+survey the page structure this way without listening to the paragraphs in
+between, then stop at whatever they want to read in full.
+
+The categories readers jump through are consistent everywhere: headings (and
+specific heading levels), landmark regions, links, buttons, form fields, tables,
+graphics, and lists. How each category is triggered differs per reader, though.
+NVDA and Orca assign one letter of the alphabet to each category: pressing that
```
p8: Maybe remind readers here that NVDA and Orca are screen readers?
```suggestion
Screen readers NVDA and Orca assign one letter of the alphabet to each category: pressing that
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+This is why HTML structure matters so much. Quick navigation only works when the
+page has the structure it jumps through. If a page has no headings, the heading
+shortcut does nothing. If a page has no landmark regions, the landmark shortcut
+does nothing. The user is left to arrow through every single line of content to
+find what they need, like reading an entire book that has no table of contents.
+
+##### Focus Mode
+
+Focus mode is the opposite. Every keystroke is passed through to the browser as
+if the screen reader were not running, so pressing `H` types the letter "h",
```
p8, line 267: When I read "pressing `H`" I interpret that as pressing the Shift and h key.
Should this be "pressing `h`" (with a lower case) instead?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+naming: interactive elements, landmarks, and other identifiable roles like
+`<img>`, `<table>`, `<dialog>`, `group`, or `list`. They are **prohibited** on
+elements with a role that does not support naming. A plain `<div>`, `<span>`, or
+`<p>` has a role (`generic` or `paragraph`) that prohibits naming, so
+`aria-label` on these elements is invalid and screen readers will ignore it.
+However, the restriction is on the *role*, not the *element*: `<div role="group"
+aria-label="...">` is valid because the `group` role supports naming, even
+though the underlying element is a `<div>`. The WAI-ARIA specification lists all
+[roles supporting name from
+author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor).
```
p8: The link "roles supporting name from author" is the first time "author" is mentioned.

In context of the paragraph, would it be clearer if we replace "roles supporting name from
author" with "roles supporting naming" for the link?

```suggestion
though the underlying element is a `<div>`. The WAI-ARIA specification lists all
[roles supporting naming](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor).
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+change the visible text and forget to update the attribute, leaving the
+accessible name out of sync and the visible label no longer included in the name
+voice control users would say ([WCAG 2.5.3 Label in Name][wcag-label-in-name]).
+
+[wcag-label-in-name]: https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html
+
+The `aria-labelledby` attribute accepts more than one ID separated by spaces.
+This lets the accessible name compose from several existing pieces of page
+content without duplicating strings.
+
```
p8, line 447: There are several paragraphs here, discussing different topics, without any headers.

Maybe a header could be added to the paragraphs describing some naming best practices?
For example:
```suggestion

#### Naming Best Practices

```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+the screen reader says natively. The instruction often does not even apply,
+since the user may be on touch or voice control rather than the input method the
+hint assumes. The same principle covers pronunciation. Phonetic spellings or
+splitting a word across `<span>` letters to make the reader say a name
+differently break translation, find-in-page, and braille output, and they
+conflict with how the user has configured their reader. Name the control, trust
+the native announcement, and rewrite the name itself if it causes friction in
+pronunciation.
+
+NOTE: The HTML [`title`][title-attr] attribute is sometimes used for the same
```
p8: I think this note about the title attribute could be mentioned a bit earlier.
Maybe right after the first example of adding the label to the icon-only button?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+aria-label="...">` is valid because the `group` role supports naming, even
+though the underlying element is a `<div>`. The WAI-ARIA specification lists all
+[roles supporting name from
+author](https://www.w3.org/TR/wai-aria-1.2/#namefromauthor).
+
+This means that on a generic element like a paragraph or a `<div>`, `aria-label`
+is ignored entirely. And on elements whose role does accept it, the attribute
+*replaces* the accessible name rather than adding to it.
+
+The solution is **visually hidden text**, a `<span>` that is visible to screen
```
p8, line 501: It's a bit unclear from reading the text which problem the following solution solves.
What would be a reason to still keep the notifications in spans?
Also, would it help to show an example with problematic code?

For example (not sure if this is correct):
````suggestion
```html+erb
<%# the aria-label is ignored here %>
<span>🔔</span>
<span aria-label="unread notifications">3</span>
```

The solution is **visually hidden text**, a `<span>` that is visible to screen
````

File: guides/source/accessibility.md [hunk 13]
```diff
@@ -0,0 +1,5028 @@
...
+    <%= image_tag slide.image, alt: slide.alt %>
+    <h3><%= slide.title %></h3>
+    <%= link_to "Read more", slide.url %>
+  <% end %>
+<% end %>
+```
+
+Rails omits the attribute when the value is `false`, so the active slide renders
+without `inert` and every other slide carries it. The single attribute covers
+every descendant, including ones added to the template later.
```
p8, line 687: Great example of how this applies to Rails!

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+      data: {
+        controller: "toggle-button",
+        toggle_button_pressed_value: @article.liked_by?(current_user),
+        action: "toggle-button#press" } %>
+```
+
+When using `aria-pressed`, do not change the button's text to reflect the state.
+If the button says "Like" when unpressed and "Unlike" when pressed, the screen
+reader announces "Unlike, pressed", which contradicts itself. Keep the text
+stable and let `aria-pressed` carry the state.
```
p8: After reading this paragraph I wondered what would be the solution if I do want to change the button's text. It's only 2 paragraphs later that it is explained.
I wonder if that paragraph should follow this one, or maybe the paragraphs can be combined?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+* **`aria-live="assertive"`** (or [`role="alert"`][role-alert]): The screen
+  reader interrupts immediately. Use sparingly, for errors and critical warnings
+  only.
+
+Two related attributes control what gets announced:
+
+* [`aria-atomic="true"`][aria-atomic]: Announces the **entire** live region
+  content when any part changes, not just the changed part. Useful when the
+  change only makes sense in context, for example a clock showing "12:35 PM"
+  where announcing just "35" would be meaningless.
```
p8: Maybe split this sentence and make it explicit what will be announced?
```suggestion
  change only makes sense in context. For example, a clock changing to "12:35 PM"
  should announce "12:35 PM" not just "35".
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  </span>
+<% end %>
+```
+
+```css
+[data-swap-pressed-value="false"] [data-swap-target="secondary"],
+[data-swap-pressed-value="true"] [data-swap-target="primary"] {
+  display: none;
+}
+```
```
p8, line 879: While looking at the example, I couldn't understand how this example worked because the css was outside my browser window. Maybe move the css code example between the js and html code examples?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+value is `nil`, but `false` renders as the literal string `"false"` instead, so
+reach for `nil` when the attribute should not appear at all. The screen reader
+announces "current step" on the active item, so users know where they are in the
+flow even when the visual styling alone conveys it.
+
+Helpers also normalize array values. For ARIA attributes that take a
+space-separated list of tokens, like `aria-labelledby` and `aria-describedby`,
+passing an array joins the entries with a single space and drops `nil` or
+`false` entries. This keeps references to other elements declarative even when
+the list depends on the state of the record.
```
p8, line 946: I think passing an array to any ARIA attribute joins the entry as an array:
https://github.com/rails/rails/blob/ad36942b04e9c31ccee5ea4b0071944d30fcf43e/actionview/lib/action_view/helpers/tag_helper.rb#L1532

Maybe we could also show examples here?
````suggestion
Helpers also normalize array values. For ARIA attributes that take a
space-separated list of tokens, like `aria-labelledby` and `aria-describedby`,
passing an array joins the entries with a single space and drops `nil` or
`false` entries. This keeps references to other elements declarative even when
the list depends on the state of the record.

```ruby
tag.span aria: { labelledby: ["foo", false, "bar"] }
=> "<span aria-labelledby=\"foo bar\"></span>"
```
````

File: guides/source/accessibility.md [hunk 18]
```diff
@@ -0,0 +1,5028 @@
...
+
+### The First Rule of ARIA
```
p8, line 953: If this is the first rule, shouldn't we mention it earlier? 😉 
Maybe move this section to the top?

p8 (commented):
This is looking great @brunoprietog ❤️  . I've picked up a lot of new information. I've added some comments to what I've read so far.

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  with low vision. Users see only a small area of the page at a time and
+  navigate by panning around. Good structure, consistent layouts, and clear
+  focus indicators help magnifier users orient themselves. Windows includes
+  Magnifier, macOS and iOS/iPadOS include Zoom, and Android includes
+  Magnification.
+
+* **Voice control:** Software that allows users to speak commands instead of
+  using a keyboard or mouse. A user might say "click Submit" or "click Email" to
+  interact with buttons and form fields. For this to work, every interactive
+  element needs a visible text label: if a button has no text, voice control
```
silva96: This feels a bit too absolute. Later the guide says icon-only buttons with an `aria-label` can be reached by voice control, so "every interactive element needs a visible text label" contradicts that.

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+example: "Password, secure text field, Must be at least 12 characters."
+
+Like `aria-labelledby`, `aria-describedby` accepts more than one ID separated by
+spaces, so a description can compose from several existing pieces of page
+content without duplicating strings.
+
+NOTE: Like `aria-label`, descriptions only work reliably on elements whose role
+supports them: interactive elements, landmarks, and other identifiable roles.
+They will not work on a plain `<div>`, `<span>`, or `<p>`, unless the element
+has an explicit role that supports naming.
```
silva96: Small wording issue: this says descriptions work on roles that support "naming", but this section is about descriptions. Should this be "roles that support descriptions" instead?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+`<table>` whenever possible.
+
+### Sortable Tables
+
+When a table can be sorted by its column headers, sighted users see a visual
+indicator (an arrow, bold text) showing which column is sorted and in which
+direction. A screen reader user has no way to perceive these visual cues.
+
+The [`aria-sort`][aria-sort] attribute on the `<th>` communicates the current
+sort state programmatically. It accepts `ascending`, `descending`, or `none`.
```
NakajimaTakuya:
```md
It accepts `ascending`, `descending`, or `none`.
```

I think `other` could also be a valid value—is there a specific reason why it’s been excluded?

NakajimaTakuya (commented):
This is a great document. I can't wait for it to be released.

guides/source/accessibility.md
nertzy: Maybe a good fit for the `<kbd>` tag

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/kbd

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+Accessibility means building applications that work for people regardless of how
+they use a computer: keyboard only, screen reader, magnifier, voice control,
+switch device, or any combination.
+
+Most of the web does not. The [WebAIM Million][webaim-million] report audits the
+top one million home pages every year and consistently finds detectable WCAG
+failures on around 95% of them. The [World Health Organization
+estimates][who-disability] that roughly one in six people lives with a
+significant disability, so the gap between that audience and the inaccessible
+state of the web is large.
```
seanpdoyle, line 43: This section does a great job of describing the data and statistics related to disability.

> significant disability

In addition to what's described as "significant" disability (which I interpret to also mean "permanent"), could it be worth mentioning more temporary disabilities (having a broken arm or broken fingers, or having dilated eyes and impaired vision after an optometrist appointment, for example) or situational/environmental disabilities (being a parent holding a child and only being limited to browsing with one hand, browser in a dark environment, for example)?

guides/source/accessibility.md
seanpdoyle: The [token_list](https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-token_list) view helper is also useful in situations that require a list of identifiers for `aria-` attributes:

```erb
<span aria-labelledby="<%= token_list("foo", false, "bar") %>"></span>
<%# => <span aria-labelledby="foo bar"></span> %>
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+[aria-current="page"] {
+  font-weight: bold;
+  border-bottom: 2px solid currentColor;
+}
+```
+
+`aria-current` accepts other values for contexts where "page" does not fit:
+`step` for a multi-step flow, `location` for the active item in a map or tree,
+`date` for a calendar's selected day, `time` for a timeline, or `true` as a
+generic fallback.
```
seanpdoyle, line 735: There is currently an open PR to change `link_to` helper calls to behave like the example in the code snippet:

https://github.com/rails/rails/pull/57350

Insights like this paragraph may be valuable in reviewing that proposed code change.

If that PR is merged (either before or after this change to the Guides), would it make sense to mention the new behavior in the code sample?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+}
+```
+
+```html+erb
+<%= tag.button "Like", type: "button",
+      data: {
+        controller: "toggle-button",
+        toggle_button_pressed_value: @article.liked_by?(current_user),
+        action: "toggle-button#press" } %>
+```
```
seanpdoyle, line 792: Maybe I'm misunderstanding the guidance being provided here, but could the sample be simplified by replacing the [Stimulus Value](https://stimulus.hotwired.dev/reference/values#change-callbacks) usage with directly declaring `aria: { pressed: … }` and reading from and writing to [ariaPressed](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaPressed)?


```js
// app/javascript/controllers/toggle_button_controller.js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  press() {
    this.element.ariaPressed = !this.element.ariaPressed
  }
}
```

```html+erb
<%= tag.button "Like", type: "button",
      aria: { pressed: @article.liked_by?(current_user) },
      data: { controller: "toggle-button", action: "toggle-button#press" } %>
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+
+When using `aria-pressed`, do not change the button's text to reflect the state.
+If the button says "Like" when unpressed and "Unlike" when pressed, the screen
+reader announces "Unlike, pressed", which contradicts itself. Keep the text
+stable and let `aria-pressed` carry the state.
+
+Also avoid toggle verbs like "Toggle sidebar" in the label: saying "toggle"
+duplicates what `aria-pressed` already communicates. Prefer action verbs like
+"Show sidebar": the screen reader announces "Show sidebar, button, pressed" or
+"Show sidebar, button, not pressed", which is clear and not redundant.
```
seanpdoyle, line 804: This is valuable guidance!

guides/source/accessibility.md
seanpdoyle: Would it make sense to declare *both* in the `html+erb` block by nesting the CSS in a `<style>` element?

```html+erb
<style>
  [data-swap-pressed-value="false"] [data-swap-target="secondary"],
  [data-swap-pressed-value="true"] [data-swap-target="primary"] {
    display: none;
  }
</style>

<%= tag.button type: "button",
      data: {
        controller: "swap",
        swap_pressed_value: @showing_completed,
        action: "swap#toggle" } do %>
  <span aria-live="polite" aria-atomic="true" aria-relevant="all">
    <span data-swap-target="primary">Show completed</span>
    <span data-swap-target="secondary">Hide completed</span>
  </span>
<% end %>
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  also pushes an empty entry into browser history, so the back button stops
+  working as expected. If the control performs an action, use a `<button>` (or
+  `button_to`) from the start.
+
+  ```html+erb
+  <%# Avoid: an anchor with no destination, wired up with JavaScript %>
+  <a href="#" data-action="notification#markAsRead:prevent">Mark as read</a>
+
+  <%# Prefer: a button %>
+  <%= button_to "Mark as read", notification_readings_path(@notification) %>
```
seanpdoyle, line 1346: Based on guidance from MDN, a `<button>` element with `[type="button"]`:

> has no default behavior, and does nothing when pressed by default. It can have client-side scripts listen to the element's events, which are triggered when the events occur.

Would it make sense for this guide to mention that behavior-only `<button>` elements should be declared with [type="button"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type) so that they do not submit any `<form>` element ancestors when clicked?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  ```
+
+* **A link for an action that changes server state**, for example `link_to` with
+  `data-turbo-method`. Even when the `href` is valid, the semantics are wrong. A
+  user hearing "Publish article, link" reasonably expects to navigate to a page
+  where they can review the article before submitting. Triggering the action
+  directly on activation removes the review they were counting on. Prefer
+  `button_to`:
```
seanpdoyle, line 1354: While it might not apply directly to the scenario outlined below, is it also worth mentioning that there is a difference between `<button>` and `<a>` elements when driving with a keyboard? 

An `<a>` element with focus will not respond to a <kbd>space</kbd> in the same way that it responds to a <kbd>enter</kbd> (i.e. triggering a "click"), but a `<button>` element will respond to both in the same way (i.e. triggering a "click").

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+
+### Buttons inside Other Forms
+
+A common reason developers use `link_to` with `data-turbo-method` is to avoid
+nesting forms, for example a delete button inside a form that edits a record.
+HTML does not allow nested `<form>` elements, and `button_to` generates its own
+`<form>`.
+
+The solution is the HTML [`form`][form-attr] attribute, which associates a
+button with a form **anywhere** in the DOM by referencing its `id`:
```
seanpdoyle, line 1389: Browsers also support `<button>` and `<input type="button">` elements with [formmethod](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#formmethod) and [formaction](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#formaction) attributes.

In this scenario, browsers would ignore a `[formmethod="delete"] attribute on a submit button, but `@hotwired/turbo-rails` [maps a `[formmethod]` attribute to Rails' `_method` attribute prior to submission](https://github.com/hotwired/turbo-rails/pull/239).

Similarly, Action View supports [encoding `:formmethod` options to the `_method` parameter for HTTP verbs other than `GET` and `POST`](https://github.com/rails/rails/pull/41045).

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+A link with no destination is not really a link. HTML does not have a `disabled`
+attribute for `<a>`: applying it has no effect in browsers, and screen readers
+ignore it.
+
+When an action that is normally a link should be unavailable, the simplest
+answer is to not render a link at all. The classic example is pagination, where
+the next control disappears at the end of the list:
+
+```html+erb
+<%= link_to "Next", url_for(page: @page.next_param) unless @page.last? %>
```
seanpdoyle: While this example fits neatly on a single line, I think that my Ruby-parsing brain misinterpreted to be equivalent to calling `link_to "Next", nil`:

```html+erb
<%= link_to "Next", @page.last? ? url_for(page: @page.next_param) : nil %>
```

What do you think about changing the structure of the example so that it is more self-evident that the conditional skips rendering the element entirely?




```suggestion
<% unless @page.last? %>
  <%= link_to "Next", url_for(page: @page.next_param) %>
<% end %>
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+readers announce when the user navigates to the table. `<th>` with
+[`scope="col"`][scope-attr] marks column headers and `scope="row"` marks row
+headers. These are what the screen reader reads aloud as the user moves between
+cells. The navigation method varies by platform: keyboard shortcuts on desktop
+(for example, `Ctrl + Alt + Arrow` keys in NVDA), or gestures on mobile (for
+example, in VoiceOver on iOS, swipe left/right moves between columns and the
+Rotor set to rows lets users swipe up/down to move between rows).
+
+[caption-el]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/caption
+[scope-attr]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/th#scope
```
seanpdoyle, line 2015: From a testing perspective, there's benefit to using elements like `<table>` that have built in semantics. For example, [Capybara](https://github.com/teamcapybara/capybara) has built-in support for a [`:table` selector](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) (as well as a [within_table](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Session#within_table-instance_method) helper method) that knows how to resolve a table based on its rows, columns, and caption:

```ruby
page.find :table, "Monthly revenue by region"

assert_table "Monthly revenue by region", with_rows: [
  { "Region" => "North", "January" => "$10,000", "February" => "$12,000" }
]

within_table "Monthly revenue by region" do
  # …
end
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+
+export default class extends Controller {
+  static targets = [ "selectAll", "item" ]
+
+  connect() {
+    this.refresh()
+  }
+
+  toggleAll() {
+    this.itemTargets.forEach(item => item.checked = this.selectAllTarget.checked)
```
seanpdoyle, line 2179: In an effort to simplify this Stimulus controller example code, could the `selectAll` target be replaced with an event listener that accesses the `<input type="checkbox">` through the Event's [target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) property?

```suggestion
  static targets = [ "item" ]

  connect() {
    this.refresh()
  }

  toggleAll(event) {
    this.itemTargets.forEach(item => item.checked = event.target.checked)
```

Then, the `selectAll` checkbox can drop the target:

```diff
- <%= form.checkbox :select_all,
-       data: { selection_target: "selectAll", action: "selection#toggleAll" } %>
+ <%= form.checkbox :select_all, data: { action: "selection#toggleAll" } %>
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+`<input>`'s `id`:
+
+```html
+<label for="user_name">Name</label>
+<input type="text" name="user[name]" id="user_name">
+```
+
+This explicit association has multiple benefits: screen readers announce "Name"
+when the input is focused, voice control users can say "click Name" to focus the
+field, and clicking the label focuses the input (enlarging the clickable area).
```
seanpdoyle, line 2243: Again, System tests (and other HTML-specific tests) that utilize [Capybara](https://github.com/teamcapybara/capybara) can access a variety of [built-in selectors](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) like `:field` (or `:fillable_field` used with calls to [fill_in](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Actions#fill_in-instance_method)), `:checkbox` and `:radio_button` (used with calls to [check](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Actions#check-instance_method) / [uncheck](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Actions#uncheck-instance_method) and [choose](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Actions#choose-instance_method), respectively), and `:fieldset` (used with calls to [within_fieldset](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Session#within_fieldset-instance_method)).

All of these selectors (and the assertions and actions that rely on them) are capable of resolving elements by their accessible names (`<label>` for form controls) and descriptions (`<fieldset>`-`<legend>` pairings), rather than CSS selectors. 

Strong Capybara conventions can drive more accessible user interfaces. Is that worth mentioning briefly here, with a link to the `#writing-accessible-system-tests` heading down below?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+  ]
+end
+```
+
+#### Writing Accessible System Tests
+
+Write system tests that interact with the application the way assistive
+technology users do. [Rails system tests](testing.html#system-testing) use
+[Capybara][capybara] under the hood, and its finders already work in terms of
+the accessible name. Prefer finding elements by their **accessible names**
```
seanpdoyle: What do you think about replacing "finders" with "selectors", since a [selector](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) is a first-party term used by Capybara?

Would it be worth linking to the list of [first-party Selectors](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) that Capybara provides?

guides/source/accessibility.md
```diff
@@ -0,0 +1,5028 @@
...
+
+# When several <nav> elements exist, disambiguate by their accessible name:
+within :element, "nav", "aria-label" => "Primary" do
+  assert_link "Home", href: root_path
+end
+
+within :element, "nav", "aria-label" => "Footer" do
+  assert_link "Privacy", href: privacy_path
+end
+```
```
seanpdoyle, line 4511: The [capybara_accessible_selectors](https://github.com/citizensadvice/capybara_accessible_selectors) gem is a third-party extension to Capybara's built-in selectors.

It provides selectors for a variety of accessibility concepts, including landmarks:

```ruby
# https://github.com/citizensadvice/capybara_accessible_selectors#main
within :main, "Welcome" do # "Welcome" here is the locator argument, which resolves to an <h1>
 # …
end

https://github.com/citizensadvice/capybara_accessible_selectors#navigation
# When several <nav> elements exist, disambiguate by their accessible name:
within :navigation, "Primary" do # "Primary" is the locator argument, which resolves [aria-label] or [aria-labelledby]
  assert_link "Home", href: root_path
end

within :navigation, "Footer" do
  assert_link "Privacy", href: privacy_path
end
```

seanpdoyle (commented):
*Tremendous* work @brunoprietog!

This is really valuable documentation to provide.

In addition to review comments, I have two additional thoughts:

1. Since this document is lengthy, I wonder if the markdown links risk colliding with one another (for example, an occurrence of `[name][]` toward the beginning of the document being overridden by an occurrence of `[name][]` later in the document, where the the multiple occurrences of `[name]: …` are different values. Could that be a problem? Does `rails/rails` have some kind of static tooling to prevent that? Would inlining markdown links to `[name](…)`-like versions solve that?

2. Several of the code samples shared throughout either don't include "Rails specific" code (`.html.erb` templates, Action View helpers, etc) or are verbose. I think this guide could serve as a loose specification/inspiration for future improvements to Action View and Hotwire, as well as an opportunity to establish more ergonomic conventions!

Thank you for dedicating your time and attention toward this!

guides/source/accessibility.md
brunoprietog: Added the colon that introduces the behaviors, and "It is easy to miss one" is its own sentence.

guides/source/accessibility.md
brunoprietog: Made it a list, parallel to the native-behavior list above.

guides/source/accessibility.md
brunoprietog: Added it, and the same in the parallel `aria-describedby` sentence.

guides/source/accessibility.md
brunoprietog: I had phrased it that way to maintain consistency with the guide, which always mentions Windows first, but if it makes things clearer, I guess that’s better. Reordered to lead with macOS.

guides/source/accessibility.md
brunoprietog: Right, defined it inline.

guides/source/accessibility.md
brunoprietog: Done, reworded to lead with "Screen readers".

guides/source/accessibility.md
brunoprietog: On the casing, I went back and forth. [NVDA's docs](https://download.nvaccess.org/documentation/userGuide.html) do write these lowercase, but they're lowercase across the board (`control`, `escape` too), which would clash with the capitalized modifiers used elsewhere in the guide. [Deque](https://dequeuniversity.com/screenreaders/nvda-keyboard-shortcuts), [Apple's VoiceOver docs](https://support.apple.com/guide/voiceover/general-commands-cpvokys01/mac), and the [Google](https://developers.google.com/style/ui-elements) and [Microsoft](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/term-collections/keys-keyboard-shortcuts) style guides all go uppercase letter with capitalized modifier. The `H`/`h` split is intentional too, the key is `H`, the letter it types is `h`. That said, happy to lowercase the single nav keys, I don't have a strong opinion here.

On `<kbd>`, agreed it's the proper element for keys. Doing it would mean wrapping every key across the command tables, and since the guides theme doesn't style `<kbd>` today (just a normalize reset), it would also want a little CSS added to the theme so the keys don't render unstyled. I'm open to doing it here. Do you think it belongs in this PR alongside the guide content, or is touching the theme styling cleaner as its own change?

guides/source/accessibility.md
brunoprietog: Changed the link text to "roles supporting naming", same target.

guides/source/accessibility.md
brunoprietog: Added the "Naming Best Practices" heading.

guides/source/accessibility.md
brunoprietog: Moved it up next to that example.

guides/source/accessibility.md
brunoprietog: True, added a "before" example and made the failure explicit: `aria-label` on a generic element is invalid and announced inconsistently.

guides/source/accessibility.md
brunoprietog: Added a pointer to the `aria-live` version right where the question comes up.

guides/source/accessibility.md
brunoprietog: Good point, changed to: should announce "12:35 PM", not just "35".

guides/source/accessibility.md
brunoprietog: Reordered to HTML, CSS, JS so the CSS sits next to its markup. I kept them as separate blocks rather than an inline `<style>` to match how CSS is shown elsewhere in the guide. I'm not sure if we should display it as a style element, given CSP and what's generally recommended in Rails, but I'm open to it if you prefer that.

guides/source/accessibility.md
brunoprietog: Added an example of the `aria:` option joining an array and dropping falsy entries, scoped to token-list attributes where an array is meaningful. I used the `aria:` form rather than `token_list`, which reads cleanest in the helper context, but `token_list` is a good tool for building the value directly. Want both shown? Since Rails already handles this automatically with an array, I thought it would be cleaner and more elegant to leave it that way.

guides/source/accessibility.md
brunoprietog: I thought about it, but it seemed odd to start talking about the first ARIA rule without first explaining what it does or why it’s there. I figured the rule would make more sense once you have a better understanding of what ARIA offers and how you might use it to bypass that rule. "The First Rule of ARIA" is the W3C's own term (the first of their [Rules of ARIA Use](https://www.w3.org/TR/using-aria/#firstrule)), so "first" is their numbering, not necessarily its position here. I added a link to that rule anyway, what do you think?

guides/source/accessibility.md
brunoprietog: You're right, reworded so the icon-only button with an `aria-label` is still reachable, the real failure is a control with no accessible name at all.

guides/source/accessibility.md
brunoprietog: Good catch! Since it works similarly to a label in terms of consistency in screen readers, I assumed the same rule applied, but neither was right. `aria-describedby` and `aria-description` are global, not role-restricted like `aria-label`. Reworded so descriptions are valid on any element but announced reliably only on an interactive control.

guides/source/accessibility.md
brunoprietog: Yes, you're right. I don't remember ever needing it in practice, but there's no reason to leave it out, I've added it.

guides/source/accessibility.md
brunoprietog: Yes, makes sense! Added a paragraph on temporary and situational disability.

guides/source/accessibility.md
brunoprietog: It was interesting, and it was great to have inspired that PR with the guide, but #57350 was closed.

guides/source/accessibility.md
brunoprietog: I did it this way because of Stimulus's automatic type casting of values, ARIA properties return a string even though they are actually booleans, but it's true that it's not really necessary, and it's simpler this other way. As `element.ariaPressed` is a string, so `!this.element.ariaPressed` is always false (a non-empty string is truthy) and the button never toggles back, I went with `this.element.ariaPressed = this.element.ariaPressed !== "true"`.

guides/source/accessibility.md
brunoprietog: Yes I think so, added a note on it.

guides/source/accessibility.md
brunoprietog: That's right! It's an important distinction, I forgot to mention it. Wrote that in.

guides/source/accessibility.md
brunoprietog: Didn't notice turbo-rails mapped `formmethod` to `_method`, nice, I remember having some issues with it in the past, but I never properly debugged it and just switched to the form element. Do you think it's worth adding this here to expand on the concept? In that case, we would accept that certain parameters are shared, like parameters that might not necessarily need to be sent, but it’s true that this approach is more elegant in some scenarios.

guides/source/accessibility.md
brunoprietog: Wrapped it in an explicit block.

guides/source/accessibility.md
brunoprietog: Nice extra reason to reach for a real `<table>`. Added a short note in the Tables section making that point, linking down to the system-tests section with the `within_table` example.

guides/source/accessibility.md
brunoprietog: I think the target has to stay here. `refresh()` needs a handle on the select-all checkbox to set its `checked`/`indeterminate` when items change (and on `connect`), and there the event target is an item, not the select-all. So `event.target` in `toggleAll` alone wouldn't let me drop it. Am I missing a way around that?

guides/source/accessibility.md
brunoprietog: Added a brief note in the Labels section that Capybara finds controls by their label (and fieldsets by their legend), so good labels make for simpler system tests, with a link down to Writing Accessible System Tests.

guides/source/accessibility.md
brunoprietog: Yes, switched to "selectors" and linked the list.

guides/source/accessibility.md
brunoprietog: Added a tip for it (`within :navigation, "Primary"`), keeping the plain `within :element` example for anyone not pulling in the gem. Do you think it's worth going into more detail here?

brunoprietog:
Thank you all for your thorough reviews! ❤️ Went through every comment and applied most of them across a few commits, replies inline.

brunoprietog:
> 1. Since this document is lengthy, I wonder if the markdown links risk colliding with one another (for example, an occurrence of `[name][]` toward the beginning of the document being overridden by an occurrence of `[name][]` later in the document, where the the multiple occurrences of `[name]: …` are different values. Could that be a problem? Does `rails/rails` have some kind of static tooling to prevent that? Would inlining markdown links to `[name](…)`-like versions solve that?

I did a thorough review of this, and as far as I could tell, I didn't find any duplicates. I'm not sure if there's a tool for that, to be honest. I did it this way because the links were a bit long, and it was easier to avoid reaching column 80, although in some cases I couldn't avoid it anyway (or in the tables).

> 2. Several of the code samples shared throughout either don't include "Rails specific" code (`.html.erb` templates, Action View helpers, etc) or are verbose. I think this guide could serve as a loose specification/inspiration for future improvements to Action View and Hotwire, as well as an opportunity to establish more ergonomic conventions!

Yes, that would be great. An official guide covering Hotwire integration with Rails that follows the patterns as they were intended. Right now, all the information is pretty scattered and not very beginner-friendly. That would also be a big help to AI agents in the future, because right now they tend to reinvent the wheel quite a bit or follow patterns that are considered contrary to the Stimulus philosophy, for example. Even once the guide is published, anyone could ask an agent to review the app based on the guidelines in the guide.

Once again, thank you for taking the time to review it so thoroughly!

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+  overrides the default while still signaling the state:
+
+  ```css
+  :disabled {
+    opacity: 0.6;
+    cursor: default;
+  }
+  ```
+
+* **Explain why nearby**, in visible text or in the surrounding layout. A
```
p8, line 2897: "Explain why nearby" was a bit confusing upon first read.
Maybe expand it a bit for more context?
```suggestion
* **Explain why the control is disabled nearby**, in visible text or in the surrounding layout. A
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+different breakpoints have no perfect solution: there is no way to ship one HTML
+source that matches every breakpoint's reading order. Building mobile-first
+reduces the risk in practice, since a source order that works on a single column
+tends to translate cleanly to wider layouts. When a breakpoint truly cannot
+avoid a reorder, test the result at that width with a keyboard and a screen
+reader to confirm the experience is still understandable.
+
+### Horizontal Scrolling
+
+Wide tables, image rails, and "card carousels" are often forced into a
```
p8, line 3658: I wasn't familiar with the term "image rails", it should probably be quoted like "card carousels"?
```suggestion
Wide tables, "image rails", and "card carousels" are often forced into a
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+semantics:
+
+```css
+p a {
+  color: #0044cc;
+  text-decoration: underline;
+}
+```
+
+TIP: The [axe browser extension](https://www.deque.com/axe/devtools/extension/),
```
p8, line 3748: This TIP feels like it should be moved before or after the section about declaring `color` and `background-color` together, not after using `text-decoration` for links.

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+font-size: 62.5% }`, force every reader to bump their preferred size by 60
+percent just to land back on what they originally chose. Form controls escape
+that inheritance by default and ship with browser-fixed typography that ignores
+the body cascade, so add one rule to bring them back in line:
+
+```css
+select, textarea, input, button {
+  font: inherit;
+}
+```
```
p8, line 3791: While it's a good CSS trick that I wasn't aware of, I'm wondering if we should just remove this as we just recommended not using `html { font-size: 62.5% }`
```suggestion
percent just to land back on what they originally chose. 
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+
+[Turbo Frames][turbo-frames] replace content within a frame. Lazy-loaded
+sections, such as a comments thread that only loads when the user requests it,
+are a common use case. These replacements do not move focus automatically, so
+activating the link to show the content leaves the user with no signal that
+anything happened.
+
+Placing `autofocus` inside the replacement markup directs focus to the revealed
+content once the frame updates. The [`turbo_frame_tag`][turbo_frame_tag] helper
+renders the frame:
```
p8, line 4109: The description of `turbo_frame_tag` put me off a bit here. It had me wonder what I should pay attention to.
Maybe explain the examples and describe `turbo_frame_tag` as a side note?
```suggestion
Placing `autofocus` inside the replacement markup directs focus to the revealed
content once the frame updates. For example, showing the comments for an article by using a Turbo Frame (the [`turbo_frame_tag`][turbo_frame_tag] helper
renders the frame):
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+  driven_by :selenium, using: :headless_chrome, screen_size: [1400, 1400]
+
+  def assert_accessible(**options)
+    options
+      .inject(Axe::Matchers::BeAxeClean.new) do |matcher, (method, value)|
+        matcher.public_send(method, *Array.wrap(value))
+      end
+      .audit(page).then do |audit|
+        assert audit.passed?, audit.failure_message
+      end
```
p8, line 4389: We don't typically use leading dots in guides examples and chain methods calls after `end` on the same line:
```suggestion
    options.inject(Axe::Matchers::BeAxeClean.new) do |matcher, (method, value)|
      matcher.public_send(method, *Array.wrap(value))
    end.audit(page).then do |audit|
      assert audit.passed?, audit.failure_message
    end
```

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+  end
+end
+```
+
+Calling `assert_accessible` without options runs the rule set axe enables by
+default: WCAG 2.0, 2.1, and 2.2 Level A and AA, plus best-practice rules.
+AAA-level rules and experimental rules are excluded by default and must be opted
+into explicitly.
+
+Each keyword argument maps to a chainable method on the matcher:
```
p8, line 4411: We call the "keyword argument" "option" above and below, maybe use it here as well?
```suggestion
Each option maps to a chainable method on the matcher:
```

Or maybe keep "argument":
```suggestion
Each option argument maps to a chainable method on the matcher:
```

guides/source/accessibility.md
p8: We try to avoid linking to gems aren't part of the default Gemfile or under the rails namespace.

guides/source/accessibility.md
```diff
@@ -0,0 +1,5103 @@
...
+page.send_keys :tab
+
+# Activate the focused element
+page.send_keys :enter
+```
+
+Use keyboard driving for the flows that depend on it, and the semantic finders
+everywhere else.
+
+### Manual Testing
```
p8, line 4578: This section goes into detail about using third party tools (how to install, which commands to use, menu items etc). I think that is a bit out of scope for this guide?
While I think it is useful having a central place to describe these tools, it can easily get outdated. Tools change and new tools get released.

Maybe just briefly describe the tools and link to their homepages for details?

p8 (commented):
Thanks for writing this @brunoprietog. This was a wonderful read. I've learned a ton and started fixing things in some of my apps.

This currently is the longest guide but I feel like most of the content is very relevant. 
It also wouldn't make sense to split it or move content to other guides.
Maybe some of the details (installing and short cut keys) for third party tooling can be cut a bit (as I've mentioned in a comment).

I've added some nitpick comments but I think this guide is mostly good to merge.

PR #57298, opened 2026-05-04
Fix `eager_load` for CPK models with explicit `select`

### LINKED ISSUE #57296: invalid fields added to SELECT when eager loading is used on a model with composite primary key
### Steps to reproduce

For a model with a composite primary key, the fields are incorrectly added to the `SELECT` list when eager loading is used with the `select` method.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.integer :other_id, null: false
    t.string :title
  end

    t.integer :post_id
  end
end

  self.primary_key = [:id, :other_id]
  has_many :comments
end

  belongs_to :post
end

  def test_association_stuff
    bad_select_fields = /"posts"\."\[""id"", ""other_id""\]"/

    # pass
    refute Post.select("post.title").to_sql.match(bad_select_fields)
    refute Post.eager_load(:comments).to_sql.match(bad_select_fields)

    # fail
    refute Post.eager_load(:comments).select("post.title").to_sql.match(bad_select_fields)
  end
end

```

### Expected behavior

The `SELECT` list should only include valid fields.

### Actual behavior

The `SELECT` list includes a stringified array: 
```sql
"posts"."[""id"", ""other_id""]" AS t0_r0,
```

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 3.4.9

fatkodima:
Fixes #57296.

### DIFF
--- a/activerecord/lib/active_record/associations/join_dependency.rb
+++ b/activerecord/lib/active_record/associations/join_dependency.rb
@@ -168,8 +168,7 @@ module ActiveRecord
         def aliases
           @aliases ||= Aliases.new join_root.each_with_index.map { |join_part, i|
             column_names = if join_part == join_root && !join_root_alias
-              primary_key = join_root.primary_key
-              primary_key ? [primary_key] : []
+              Array(join_root.primary_key)
             else
               join_part.column_names
             end
--- a/activerecord/test/cases/associations/eager_test.rb
+++ b/activerecord/test/cases/associations/eager_test.rb
@@ -1754,6 +1754,17 @@ class EagerAssociationTest < ActiveRecord::TestCase
     assert_equal [agreement_3, agreement_4], results.last.order_agreements.sort_by(&:signature)
   end
 
+  test "preloading has_many with cpk and explicit select" do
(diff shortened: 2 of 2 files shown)


nvasilevski (approved):

activerecord/lib/active_record/associations/join_dependency.rb
```diff
@@ -169,7 +169,7 @@ def aliases
           @aliases ||= Aliases.new join_root.each_with_index.map { |join_part, i|
             column_names = if join_part == join_root && !join_root_alias
               primary_key = join_root.primary_key
-              primary_key ? [primary_key] : []
+              primary_key ? Array(primary_key) : []
```
nvasilevski: Could it now be one-liner `Array(join_root.primary_key)`? Since `Array(nil) # => []`

fatkodima: Thanks, good idea! 👍

PR #57300, opened 2026-05-05
Generate constants for enum values

pierry01:
### Motivation / Background

When you declare an `enum` on an Active Record model today, Rails generates predicate, bang, and scope methods, but it does not expose the values as constants. That means application code either hardcodes the underlying strings (`User.where(status: "active")`) or every project ends up reimplementing the same boilerplate (`ACTIVE = "active"`) on each model.

This PR makes Rails generate those constants automatically, using the same `prefix`/`suffix` rules already applied to method names — so the constants stay in sync with the methods you already get.

### Detail

For each enum value, a class-level constant is defined on the model. The constant name is `"#{prefix}#{label}#{suffix}".upcase` (matching the generated method name), and its value is the frozen label string.

```ruby
class Conversation < ApplicationRecord
  enum :status, [ :active, :archived ]
end

Conversation::ACTIVE   # => "active"
Conversation::ARCHIVED # => "archived"

Conversation.where(status: Conversation::ACTIVE)
```

`prefix` and `suffix` are honored:

```ruby
class Order < ApplicationRecord
  enum :status, [ :placed, :shipped ], prefix: true
end

Order::STATUS_PLACED  # => "placed"
Order::STATUS_SHIPPED # => "shipped"
```

```ruby
  enum :font_size, [ :small, :medium, :large ], prefix: :with, suffix: true
end

Book::WITH_SMALL_FONT_SIZE  # => "small"
Book::WITH_MEDIUM_FONT_SIZE # => "medium"
Book::WITH_LARGE_FONT_SIZE  # => "large"
```

A few design choices worth flagging for review:

- **Pre-existing constants are not overwritten.** The implementation uses `const_defined?(name, false)` to skip when the class already defines a constant of the same name, so any manual declarations on the model take precedence and existing apps don't break on upgrade.
- **Invalid identifiers are skipped silently.** If the resulting name does not match `\A[A-Z][A-Z0-9_]*\z` (e.g. the label is `"404"` or contains non-ASCII characters), no constant is created, but the method is still generated as it is today. Consistent with the current behavior for similar labels.
- **No new option.** Rather than adding a `constants: false` opt-out, the silent skip on collision covers the only meaningful conflict case. Happy to add an explicit option if reviewers prefer.
- **Frozen string deduped via `-label`** instead of `label.freeze`, to take advantage of the frozen string pool.

### Additional information

- Documentation is updated in the module-level docstring for `ActiveRecord::Enum` (renders to api.rubyonrails.org).
- No dedicated guide for enums exists under `guides/source/`, so no guide change is required.
- `bin/rubocop activerecord/lib/active_record/enum.rb activerecord/test/cases/enum_test.rb` passes clean.
- `bundle exec rake test:sqlite3 TEST=test/cases/enum_test.rb` passes (111 runs, 503 assertions, 0 failures).
- Full `bundle exec rake test:sqlite3` shows 0 regressions versus `main`.

### Checklist

* [x] Tests are added.
* [x] CHANGELOG is updated.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,30 @@
+*   Define class-level constants for enum values, mirroring the
+    `prefix`/`suffix` naming scheme used for the generated methods. The
+    constant value is the frozen label string.
+
+        class Conversation < ApplicationRecord
+          enum :status, [ :active, :archived ]
+        end
+
+        Conversation::ACTIVE   # => "active"
+        Conversation::ARCHIVED # => "archived"
+
+        Conversation.where(status: Conversation::ACTIVE)
+
+    With `prefix:` or `suffix:`, the constant name follows the same shape
+    as the generated method:
+
+        class Order < ApplicationRecord
+          enum :status, [ :placed, :shipped ], prefix: true
+        end
+
+        Order::STATUS_PLACED  # => "placed"
+        Order::STATUS_SHIPPED # => "shipped"
+
+    Constants already defined on the class are not overwritten.
+
+    *Jean Mendonça*
+
 *   Bump the minimum PostgreSQL version to 10.0.
 
     As part of this change, `supports_pgcrypto_uuid?` is deprecated because
--- a/activerecord/lib/active_record/enum.rb
+++ b/activerecord/lib/active_record/enum.rb
@@ -118,6 +118,22 @@ module ActiveRecord
   #     enum :status, [ :active, :archived ], instance_methods: false
   #   end
   #
+  # A class-level constant is also defined for each value of the enum, named
+  # after the corresponding generated method (so +:prefix+ and +:suffix+ are
+  # honored as well). The constant's value is the frozen label string:
+  #
+  #   class Conversation < ActiveRecord::Base
+  #     enum :status, [ :active, :archived ]
+  #   end
+  #
+  #   Conversation::ACTIVE   # => "active"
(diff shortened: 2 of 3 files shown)


fatkodima:
Thanks for the PR, but I think existing `Conversation.statuses[:active]` syntax covers this.

pierry01:
> Thanks for the PR, but I think existing `Conversation.statuses[:active]` syntax covers this.

@fatkodima thanks for taking a look! I understand `Conversation.statuses[:active]` returns the same value, but I'd argue it doesn't fully cover the same use case:

1. Typo safety: `Conversation.statuses[:activ]` silently returns `nil` and propagates downstream, often showing up as a confusing bug far from the call site. A method-based accessor would raise `NoMethodError` immediately, at the exact line where the mistake was made.
2. Editor/LSP support: Symbol keys inside a hash lookup are opaque to Solargraph, RubyMine, Sorbet, and similar tools — there's no autocomplete and no go-to-definition. A real method, on the other hand, is discoverable: developers can find available values without grepping the model, and the editor flags removed/renamed values right away.
3. Refactor safety: Renaming an enum value with the hash syntax requires a project-wide string/symbol search and is easy to miss. Method calls are picked up by static analysis and IDE rename refactorings.

These are small ergonomic wins individually, but together they meaningfully reduce a class of silent bugs that the current syntax can't catch. Happy to iterate on the API shape if the concern is surface area rather than the feature itself.

PR #53512, opened 2024-10-31
Allow hidden_field(_tag) to accept a custom autocomplete value

brendon:
https://github.com/rails/rails/pull/43280 introduced an enforced `autocomplete="off"` to all hidden inputs generated by Rails to fix a [firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=520561).

Unfortunately it's also a legitimate use-case to specify an `autocomplete` with a value such as `username` and a value on a hidden input. This hints to the browser that (in this example) the username of a password reset form is what we've provided as the value and the password manager can store it as such.

This commit only sets `autocomplete="off"` if another `autocomplete` value isn't provided.

Supersedes https://github.com/rails/rails/pull/47798 which didn't seem to be as elegant a solution and didn't have proper tests.

PR #57304, opened 2026-05-06
[ci skip] Fix getting started guide: remove autocomplete="off" from authenticity_token example

ryu:
The getting started guide shows an HTML form example with `autocomplete="off"` on the `authenticity_token` hidden input:

```html
    <input type="hidden" name="authenticity_token" value="..." autocomplete="off">
```
Since Rails 8.1.0, `autocomplete="off"` is no longer added to hidden inputs generated by `token_tag` (and others). This was removed in #53512 because the `autocomplete` attribute is not applicable to `hidden` input elements per the HTML spec.

This PR updates the example to reflect the current output.

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -1237,7 +1237,7 @@ will look like this:
 
 ```html
 <form action="/products" accept-charset="UTF-8" method="post">
-  <input type="hidden" name="authenticity_token" value="UHQSKXCaFqy_aoK760zpSMUPy6TMnsLNgbPMABwN1zpW-Jx6k-2mISiF0ulZOINmfxPdg5xMyZqdxSW1UK-H-Q" autocomplete="off">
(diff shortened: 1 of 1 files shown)


fatkodima:
That PR did not remove `autocomplete="off"`, it made it configurable, but still adds the original value if none is provided.

ryu:
To clarify: `config.load_defaults 8.1` sets `action_view.remove_hidden_field_autocomplete = true` (see [configuration.rb L368](https://github.com/rails/rails/blob/main/railties/lib/rails/application/configuration.rb#L368)), so `autocomplete="off"` is not output in a standard Rails 8.1 app.

The default of `ActionView::Base.remove_hidden_field_autocomplete` is `false`, but `load_defaults 8.1` overrides it to `true`. The guide example therefore does not reflect the actual output for apps using Rails 8.1 defaults.

fatkodima:
Correct PR link https://github.com/rails/rails/pull/55989

PR #57305, opened 2026-05-06
Add alias expected for ActionController::Parameters

schmijos:
[`expect`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect) looks like [`except`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-except).

After this PR I'm able to write it like this:

```rb
params = ActionController::Parameters.new(a: 1, b: 2, c: 3)

params.expect(:a, :b)
params.expected(:a, :b) # <-- new

params.except(:a, :b)
params.without(:a, :b)
```

The idea orginates from [my Rails Forum discussion](https://discuss.rubyonrails.org/t/expect-looks-like-except/90144). Since we already have the [alias `required` for `require`](https://github.com/rails/rails/blob/2d670320f7b02ae879545d5202f0633841b8f196/actionpack/lib/action_controller/metal/strong_parameters.rb#L533), I think this doesn't fall under bike-shedding as of: [PR#51674 issue comment](https://github.com/rails/rails/pull/51674#issuecomment-2341196891)

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Add `ActionController::Parameters#expected` and `#expected!` as
+    aliases of `#expect` and `#expect!`.
+
+    `expect` is one letter away from `except` and easy to misread at a glance.
+    `expected` reads as the same intent without the visual collision.
+
+        params.expected(person: [:name, :age])
+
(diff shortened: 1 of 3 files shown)


rafaelfranca:
I'm ok with this one. @dhh?

PR #57306, opened 2026-05-06
Add `strip_null_bytes` option to the PostgreSQL adapter

sislr:
### Motivation / Background

PostgreSQL does not support null bytes in text columns. Any string containing a null byte raises an `ArgumentError` from the `pg` gem. In practice, null bytes can appear in user-submitted form data (e.g. search fields), causing unhandled 500 errors.

I started a discussion in https://discuss.rubyonrails.org/t/handling-null-bytes-in-postgres/89924 and explored multiple approaches to this problem. For now, I've decided to implement a `PostgreSQLAdapter` option `strip_null_bytes`.  The option defaults to false to preserve existing behavior. When `strip_null_bytes` is enabled, null bytes are being stripped from strings during quoting and type casting.

### Additional information

I spoke with @tenderlove about this topic at last year's Rails World. We still see this error coming up from time to time, so I thought I'd give it a try :)

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
@@ -125,6 +125,7 @@ module ActiveRecord
 
         # Quotes strings for use in SQL input.
         def quote_string(s) # :nodoc:
+          s = s.delete("\0") if self.class.strip_null_bytes
           with_raw_connection(allow_retry: true, materialize_transactions: false) do |connection|
             connection.escape(s)
           end
@@ -181,6 +182,8 @@ module ActiveRecord
             encode_range(value)
           when Rational
             value.to_f
+          when String
+            self.class.strip_null_bytes ? value.delete("\0") : super
           else
             super
           end
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,23 @@
+*   Add `strip_null_bytes` option to the PostgreSQL adapter.
+
+    PostgreSQL does not support null bytes (`\0`) in text columns. When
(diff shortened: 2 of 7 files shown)


activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb
```diff
@@ -181,6 +182,8 @@ def type_cast(value) # :nodoc:
             encode_range(value)
           when Rational
             value.to_f
+          when String
+            self.class.strip_null_bytes ? value.delete("\0") : super
```
sislr, line 186:
```suggestion
            self.class.strip_null_bytes ? super(value.delete("\0")) : super
```
not 100% sure whether we'd need another super call here

sislr: I also thought about extracting the type casting logic into a separate `Postgres::String` type. I've inlined it for the moment to keep it simple

chaadow:
Really cool! I created my own helper that I was using in a reactive way ( in reaction to Sentry errors ) but having this option enabled would allow me to proactively sanitize user input and get rid of the helper!

Hopefully this will get reviewed and merged soon!

sislr:
hey @rafaelfranca, would appreciate your opinion on this :)

flavorjones:
Hi @sislr, I replied at length in the discussion thread you started at https://discuss.rubyonrails.org/t/handling-null-bytes-in-postgres/89924/2?u=flavorjones

But to summarize here: I'm not convinced this is the right solution for this problem because it's overly broad and introduces behaviors that violate some core behaviors of Active Record.

However, a potential alternative might be to wrap this ArgumentError with an Active Record exception class that the application can then choose how to handle. I hope we can continue the conversation on the discussion thread!

PR #57309, opened 2026-05-06
Prevent test_checkout_fairness_by_group from hanging indefinitely

bogdan:
### Motivation / Background

`test_checkout_fairness_by_group` in `activerecord/test/cases/connection_pool_test.rb` can hang CI indefinitely. The test polls for thread completion with an unbounded `loop { sleep 0.1; break if ... }`; if a waiter never reaches the success or error path, the loop spins forever until the agent's job timeout kills the build.

Observed hanging build: https://buildkite.com/rails/rails/builds/128041

The same family of tests has been flaky before — see #53591.

### Detail

Wraps the poll loop in `Timeout.timeout(5)` so the failure surfaces as a normal test failure instead of a stuck CI job.

### Additional information

Test-only change. A follow-up could replace the poll with a `Concurrent::CountDownLatch` (already used elsewhere in this file) to drop the `sleep` entirely.

### Checklist

* [x] Commit message describes what changed and why.
* [x] Tests are updated.
* [ ] CHANGELOG — n/a, test-only fix.

### DIFF
--- a/activerecord/test/cases/associations/belongs_to_associations_test.rb
+++ b/activerecord/test/cases/associations/belongs_to_associations_test.rb
@@ -1915,7 +1915,7 @@ class BelongsToWithForeignKeyTest < ActiveRecord::TestCase
 end
 
 class AsyncBelongsToAssociationsTest < ActiveRecord::TestCase
-  include WaitForAsyncTestHelper
+  include WaitForTestHelper
 
   self.use_transactional_tests = false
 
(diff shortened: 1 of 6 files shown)


rafaelfranca:
I really dislike `Timeout.timeout` because it can raise the exception in random places of the execution leaving the process sometimes in a unrecoverable state, so my fear with this fix is that the timeout will just make all the tests after that fail. Can we go with the `CountDownLatch` implementation?

bogdan:
@rafaelfranca good note, I didn't think it is so evil. `CountDownLatch` usage cause too much restructure of the test, I think there is better approach.

I searched the repo for existing patterns on avoiding this problem and found the one I think makes the most sense.
Please take a look.

PR #57310, opened 2026-05-06
fix: has_json accessors after requiring active_model

tellodaniel:
### Motivation / Background

`ActiveModel::SchematizedJson` calls `String#remove`, which is defined by `active_support/core_ext/string/filters`.

That dependency is currently implicit. When `ActiveModel::SchematizedJson` is loaded without another file having already loaded that string extension, calling a `has_json` accessor can raise a `NoMethodError`.

### Detail

This Pull Request makes the dependency explicit by requiring `active_support/core_ext/string/filters` from `active_model/schematized_json`.

### Additional information

I verified the fix with a standalone Active Model script and the Active Model test suite.

No CHANGELOG entry was added because this is a minor bug fix.

### Reproduce steps

This reproduces on `main` by running:

```bash
bundle exec ruby -Iactivesupport/lib -Iactivemodel/lib -e 'require "active_model"; account = Class.new { include ActiveModel::Attributes; include ActiveModel::SchematizedJson; attribute :settings; has_json :settings, restricts_access: true }.new; p account.settings.restricts_access'
```

### Checklist

### DIFF
--- a/activemodel/lib/active_model/schematized_json.rb
+++ b/activemodel/lib/active_model/schematized_json.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/core_ext/hash/reverse_merge"
+require "active_support/core_ext/string/filters"
 require "active_support/core_ext/symbol/starts_ends_with"
 
 module ActiveModel


rafaelfranca:
How can we reproduce the error to know it is fixed by this PR?

tellodaniel:
@rafaelfranca 

You can reproduce with:

```bash
bundle exec ruby -Iactivesupport/lib -Iactivemodel/lib -e 'require "active_model"; account = Class.new { include ActiveModel::Attributes; include ActiveModel::SchematizedJson; attribute :settings; has_json :settings, restricts_access: true }.new; p account.settings.restricts_access'
```

I updated the description to include this.

PR #57311, opened 2026-05-07
Accept Active Record Encryption credentials as ENV

claudiob:
Taking advantage of `Rails.apps.creds` (#56455), the `primary_key`, `deterministic_key` and `key_derivation_salt` required by ActiveRecord::Encryption can now also be provided through environment variables named:

- `ACTIVE_RECORD_ENCRYPTION__PRIMARY_KEY`
- `ACTIVE_RECORD_ENCRYPTION__DETERMINISTIC_KEY`
- `ACTIVE_RECORD_ENCRYPTION__KEY_DERIVATION_SALT`

### DIFF
--- a/railties/test/application/creds_test.rb
+++ b/railties/test/application/creds_test.rb
@@ -16,6 +16,7 @@ class Rails::CredsTest < ActiveSupport::TestCase
     app("production")
 
     ENV["MYSTERY"] = "hidden"
+    Rails.app.creds.reload
     assert_equal "hidden", Rails.app.creds.require(:mystery)
 
     ENV.delete("MYSTERY")
@@ -32,6 +33,7 @@ class Rails::CredsTest < ActiveSupport::TestCase
 
     ENV["MYSTERY"] = "hidden"
     Rails.app.creds = ActiveSupport::CombinedConfiguration.new(Rails.app.envs)
+    Rails.app.creds.reload
     assert_equal "hidden", Rails.app.creds.require(:mystery)
 
     ENV.delete("MYSTERY")
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Accept encryption credentials as ENV
+
+    Taking advantage of Rails.apps.creds (#56455), the `primary_key`, `deterministic_key` and
+    `key_derivation_salt` required by ActiveRecord::Encryption can now also be provided through
+    environment variables named `ACTIVE_RECORD_ENCRYPTION__PRIMARY_KEY`,
+    `ACTIVE_RECORD_ENCRYPTION__DETERMINISTIC_KEY`, `ACTIVE_RECORD_ENCRYPTION__KEY_DERIVATION_SALT`.
+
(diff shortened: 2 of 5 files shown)


claudiob:
Sorry about the failing tests… fixing right now 🛠️

railties/test/application/creds_test.rb
```diff
@@ -16,6 +16,7 @@ class Rails::CredsTest < ActiveSupport::TestCase
     app("production")
 
     ENV["MYSTERY"] = "hidden"
+    Rails.app.creds.reload
```
claudiob, line 19: The reason why I had to change this test is because, after this commit, `Rails.app.creds` is invoked loaded during `ActiveSupport.on_load(:active_record_encryption)` and this method uses memoization to store `Rails.app.envs`:

https://github.com/rails/rails/blob/ce3fb7af108d5ee31c06e3765a6694eace0677c4/railties/lib/rails/application.rb#L562-L564

I _believe_ this is something unrelated to this PR, meaning: when you declare `ENV[key] = value` in the code, then you should not expect `key` to be inside `Rails.app.env` or `Rails.app.creds` unless you explicitly reload them.

Until now the test was passing because it assumed nobody else had called the method beforehand.

Please let me know if I've done anything wrong here. Thanks.

claudiob:
Additional context for the value of this PR comes from [this article](https://andycroll.com/ruby/use-rails-combined-credentials/) by @andycroll 

- I have a Rails app which uses Active Record encrypts
- The values for `primary_key`, `deterministic_key`, `key_derivation_salt` are stored inside the encrypted file config/environments/production.yml.enc   
- The app is deployed to two different stages: `sandbox` (used for test) and `production` (real prod)
- Both stages use `RAILS_ENV=production` as they are meant to be as similar as possible
- Many developers have access to the `sandbox` stage (including the ENV), while only superadmins have access to `production` since production has real production credentials (Cloudflare, Twilio, …)
- For this reason, we don't use on the `sandbox` stage the same `RAILS_MASTER_KEY` as in `production`
- Therefore `sandbox` is not able to decrypt config/environments/production.yml.enc (for security reason)
- Therefore we redefine the values using ENV on sandbox

This PR enables the above: simply declare `ACTIVE_RECORD_ENCRYPTION__PRIMARY_KEY=...` on sandbox and still have the power of Active Record encryption without sharing key with `production` and without having to add a workaround to config/application.rb with [something like this](https://guides.rubyonrails.org/active_record_encryption.html#generate-encryption-key):

```ruby
config.active_record.encryption.primary_key = credentials.dig(:active_record_encryption, :primary_key) || ENV["ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY"]
config.active_record.encryption.deterministic_key = credentials.dig(:active_record_encryption, :deterministic_key) || ENV["ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY"]
config.active_record.encryption.key_derivation_salt = credentials.dig(:active_record_encryption, :key_derivation_salt) || ENV["ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT"]
```

PR #57316, opened 2026-05-07
Reject malformed Mailgun ingress timestamps

### LINKED ISSUE #57315: Action Mailbox Mailgun ingress raises on malformed timestamp
### Steps to reproduce

Send a Mailgun Action Mailbox ingress request with a configured `MAILGUN_INGRESS_SIGNING_KEY`, but pass a malformed `timestamp` value:

```ruby
post rails_mailgun_inbound_emails_url, params: {
  timestamp: "not-a-number",
  token: "7VwW7k6Ak7zcTwoSoNm7aTtbk1g67MKAnsYLfUB7PdszbgR5Xi",
  signature: "ef24c5225322217bb065b80bb54eb4f9206d764e3e16abab07f0a64d1cf477cc",
  "body-mime" => file_fixture("../files/welcome.eml").read
}
```

The same behavior can also be observed by instantiating the authenticator directly:

```ruby
ActionMailbox::Ingresses::Mailgun::InboundEmailsController::Authenticator.new(
  key: "k",
  timestamp: "not-a-number",
  token: "t",
  signature: "s"
).authenticated?
```

### Expected behavior

The malformed timestamp should be treated as a failed authentication attempt and the ingress should reject the request, matching the documented behavior for Mailgun requests whose signature or timestamp cannot be validated.

### Actual behavior

`Integer(timestamp)` raises `ArgumentError`, so a malformed timestamp can escape the authentication check as a server error instead of returning `401 Unauthorized`.

### System configuration

**Rails version**: main branch (`ad36942b04e9`)

**Ruby version**: 3.3.1

afurm:
### Motivation / Background

Fixes #57315.

The Mailgun Action Mailbox ingress parsed the request `timestamp` with `Integer(timestamp)` while constructing the authenticator. A malformed timestamp could raise `ArgumentError` before authentication returned `false`, so the ingress produced a server error instead of rejecting the request as unauthorized.

### Detail

This keeps the raw timestamp string for the Mailgun HMAC input, parses a separate timestamp value for the recency check with `exception: false`, and treats invalid timestamps as failed authentication.

### Additional information

Validated with:

```bash
cd actionmailbox && bin/test test/controllers/ingresses/mailgun/inbound_emails_controller_test.rb
cd actionmailbox && bin/test
bundle exec rubocop actionmailbox/app/controllers/action_mailbox/ingresses/mailgun/inbound_emails_controller.rb actionmailbox/test/controllers/ingresses/mailgun/inbound_emails_controller_test.rb
git diff --check
```

### DIFF
--- a/actionmailbox/app/controllers/action_mailbox/ingresses/mailgun/inbound_emails_controller.rb
+++ b/actionmailbox/app/controllers/action_mailbox/ingresses/mailgun/inbound_emails_controller.rb
@@ -85,7 +85,8 @@ module ActionMailbox
         attr_reader :key, :timestamp, :token, :signature
 
         def initialize(key:, timestamp:, token:, signature:)
-          @key, @timestamp, @token, @signature = key, Integer(timestamp), token, signature
+          @key, @timestamp, @token, @signature = key, timestamp.to_s, token, signature
+          @parsed_timestamp = Integer(timestamp, exception: false)
         end
 
         def authenticated?
@@ -99,7 +100,7 @@ module ActionMailbox
 
           # Allow for 2 minutes of drift between Mailgun time and local server time.
           def recent?
-            Time.at(timestamp) >= 2.minutes.ago
+            @parsed_timestamp && Time.at(@parsed_timestamp) >= 2.minutes.ago
           end
 
           def expected_signature
--- a/actionmailbox/test/controllers/ingresses/mailgun/inbound_emails_controller_test.rb
+++ b/actionmailbox/test/controllers/ingresses/mailgun/inbound_emails_controller_test.rb
@@ -76,6 +76,24 @@ class ActionMailbox::Ingresses::Mailgun::InboundEmailsControllerTest < ActionDis
     assert_response :unauthorized
   end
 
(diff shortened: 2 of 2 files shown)


actionmailbox/app/controllers/action_mailbox/ingresses/mailgun/inbound_emails_controller.rb
```diff
@@ -99,7 +100,9 @@ def signed?
 
           # Allow for 2 minutes of drift between Mailgun time and local server time.
           def recent?
-            Time.at(timestamp) >= 2.minutes.ago
+            @parsed_timestamp && Time.at(@parsed_timestamp) >= 2.minutes.ago
+          rescue ArgumentError, RangeError
```
rafaelfranca: Why is this here? I don't see tests for those rescue. Can you add it please?

afurm: Good point. I removed the extra `rescue` instead of adding coverage for it, since the fix only needs `Integer(timestamp, exception: false)` to handle malformed timestamps.

PR #57317, opened 2026-05-07
Ractor-safe class attribute implementation

jhawthorn:
This modifies the class_attribute implementation to be Ractor-safe, with nearly the same performance (a tiny bit slower, but within 10%).

This is done by splitting the "namespaced_name" method, which was the block-method used with define_method to get the inheritance-based definition into two.

Previously we had a method named something like `__class_attr__name`, which would return the value, and we'd modify the value via closure and redefine it when the class changed. Under this commit instead we have `__class_attr__name_owner`, which returns just the class, rather than the value, and is redefined when the class changes. The value is read from an `attr_reader` named `__class_attr__name`, which does not need redefinition.

Because classes are Ractor-shareable, that allows us to make `__class_attr__name_owner` a shareable proc. And then we get Ractor ivar semantics from reading the value from the `attr_reader` (if the value is shareable all Ractors can read it, otherwise only the main Ractor can read it).

This also fixes a bug where instance_reader was defined unconditionally when setting on a singleton class.

**Benchmark**

``` ruby
require "bundler/setup"
require "active_support/core_ext/class/attribute"

class Base
  class_attribute :setting, default: :hello
  class_attribute :nums, default: [1, 2, 3].freeze
end

class Sub < Base
end

class SubSub < Sub
end

Sub.setting = :world

obj_base = Base.new
obj_sub = Sub.new

IPS.ips do |x|
  x.report("Base.setting (default)") { Base.setting }
  x.report("Sub.setting (overridden)") { Sub.setting }
  x.report("SubSub.setting (inherited 2 levels)") { SubSub.setting }
  x.report("Base.new.setting (instance)") { obj_base.setting }
  x.report("Sub.new.setting (instance, overridden class)") { obj_sub.setting }
end
```

**Before**

```
❯ ips --yjit bench_class_attr.rb
ruby 4.0.1 (2026-01-13 revision e04267a14b) +YJIT +PRISM [x86_64-linux]
                      Base.setting (default):    50.330M i/s (± 0.0%)
                    Sub.setting (overridden):    50.304M i/s (± 0.0%)
         SubSub.setting (inherited 2 levels):    50.130M i/s (± 0.0%)
                 Base.new.setting (instance):    42.409M i/s (± 0.3%)
Sub.new.setting (instance, overridden class):    42.130M i/s (± 0.1%)

Summary
  Base.setting (default) ran
    1.00 ± 0.00 times faster than Sub.setting (overridden)
    1.00 ± 0.00 times faster than SubSub.setting (inherited 2 levels)
    1.19 ± 0.00 times faster than Base.new.setting (instance)
    1.19 ± 0.00 times faster than Sub.new.setting (instance, overridden class)
```

**After**

```
❯ ips --yjit bench_class_attr.rb
ruby 4.0.1 (2026-01-13 revision e04267a14b) +YJIT +PRISM [x86_64-linux]
                      Base.setting (default):    49.119M i/s (± 0.1%)
                    Sub.setting (overridden):    48.812M i/s (± 0.0%)
         SubSub.setting (inherited 2 levels):    49.157M i/s (± 0.0%)
                 Base.new.setting (instance):    42.835M i/s (± 0.1%)
Sub.new.setting (instance, overridden class):    42.865M i/s (± 0.0%)

Summary
  SubSub.setting (inherited 2 levels) ran
    1.00 ± 0.00 times faster than Base.setting (default)
    1.01 ± 0.00 times faster than Sub.setting (overridden)
    1.15 ± 0.00 times faster than Sub.new.setting (instance, overridden class)
    1.15 ± 0.00 times faster than Base.new.setting (instance)
```

### DIFF
--- a/activesupport/lib/active_support/callbacks.rb
+++ b/activesupport/lib/active_support/callbacks.rb
@@ -944,7 +944,7 @@ module ActiveSupport
             # HACK: We're making assumption on how `class_attribute` is implemented
             # to save constantly duping the callback hash. If this desync with class_attribute
             # we'll lose the optimization, but won't cause an actual behavior bug.
-            unless singleton_class.private_method_defined?(:__class_attr__callbacks, false)
+            unless singleton_class.private_method_defined?(:__class_attr__callbacks_owner, false)
               self.__callbacks = __callbacks.dup
             end
             name = name.to_sym
--- a/activesupport/lib/active_support/class_attribute.rb
+++ b/activesupport/lib/active_support/class_attribute.rb
@@ -3,24 +3,24 @@
 module ActiveSupport
   module ClassAttribute # :nodoc:
     class << self
-      def redefine(owner, name, namespaced_name, value)
-        if owner.singleton_class?
-          if owner.attached_object.is_a?(Module)
-            redefine_method(owner, namespaced_name, private: true) { value }
+      def redefine(owner, name, owner_method, reader_method, value, instance_reader)
+        ivar_name = :"@#{reader_method}"
+        owner.instance_variable_set(ivar_name, value)
+
+        owner_proc =
+          if defined?(Ractor.shareable_proc)
+            Ractor.shareable_proc { owner }
           else
-            redefine_method(owner, name) { value }
+            -> { owner }
           end
+
+        # If redefining on a singleton class, and including instance_reader, we
+        # need to update it to use self.singleton_class instead of self.class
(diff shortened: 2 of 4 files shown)


rafaelfranca (approved):

byroot:
Ack

byroot:
By "ack" I meant I'd review it soon 😅

byroot:
Nice patch. I guess my only review comment would have been to add a test for Ractor support.

PR #57323, opened 2026-05-08
Enable `Style/MutableConstant` cop with `literals` style

paracycle:
### Motivation / Background

This Pull Request has been created as part of work towards making the Rails framework more Ractor-safe. Ractor safety requires that for any class/module that can be shared between Ractors all of their constants are shareable as well. For simple constants that is trivially possible by freezing the constants at definition time, unless the constant is being used for a mutable context.

### Detail

This Pull Request enables the `Style/MutableConstant` cop for literal constant definitions, which should mostly be static values for the lifetime of the framework. For the rare cases where constants are being used in order to be mutated (e.g. hashes used for caching, or constants defined for extension by framework consumers) have been marked with a `rubocop:disable` comment, until we have better fixes for them.

### Checklist


actiontext/lib/action_text/trix_attachment.rb
```diff
@@ -8,15 +8,15 @@ class TrixAttachment
     TAG_NAME = "figure"
     SELECTOR = "[data-trix-attachment]"
 
-    COMPOSED_ATTRIBUTES = %w( caption presentation )
+    COMPOSED_ATTRIBUTES = %w( caption presentation ).freeze
     ATTRIBUTES = %w( sgid contentType url href filename filesize width height previewable content ) + COMPOSED_ATTRIBUTES
```
rafaelfranca, line 12: Should this be frozen as well?

paracycle: For the scope of this PR, I am only changing the locations flagged by the cop. Since this constant isn't a "literal", it is not flagged by the cop.

We could freeze this constant, but I'd rather not do it as part of this PR.

rafaelfranca: 👍🏽

tools/rdoc-to-md
```diff
@@ -8,7 +8,7 @@ require "strscan"
 require "rdoc"
 require "prism"
 
-OPTIONS = {}
+OPTIONS = Hash.new
```
byroot: Might be better just not to make it a constant?

railties/lib/rails/code_statistics.rb
```diff
@@ -28,7 +28,7 @@ class CodeStatistics
       %w(Channel\ tests     test/channels),
       %w(Integration\ tests test/integration),
       %w(System\ tests      test/system),
```
byroot, line 30: I'm surprised the cop doesn't care about this. We're freezing the outter array, but it's still mutable.

railties/lib/rails/code_statistics.rb
```diff
@@ -38,18 +38,18 @@ class CodeStatistics
...
-    EXTENSIONS = %w[rb js ts css scss coffee rake erb]
+    EXTENSIONS = %w[rb js ts css scss coffee rake erb].freeze
 
     PATTERN = /^(?!\.).*?\.(#{EXTENSIONS.join("|")})$/
 
-    class_attribute :directories, default: DIRECTORIES
-    class_attribute :test_types, default: TEST_TYPES
-    class_attribute :extensions, default: EXTENSIONS
-    class_attribute :pattern, default: PATTERN
+    class_attribute :directories, default: DIRECTORIES.dup
```
byroot: This is strictly worse code IMO, we're freezing a constant just to dup it?, really seems pointless working around a cop that doesn't make sense in the case in question.

activesupport/lib/active_support/xml_mini.rb
```diff
@@ -60,16 +60,18 @@ def content_type
...
-    } unless defined?(FORMATTING)
+    }.freeze unless defined?(FORMATTING)
 
     unless defined?(PARSING)
       PARSING = {
         "symbol"       => Proc.new { |symbol|  symbol.to_s.to_sym },
         "date"         => Proc.new { |date|    ::Date.strptime(date, "%Y-%m-%d") },
-        "datetime"     => Proc.new { |time|    Time.xmlschema(time).utc rescue ::DateTime.parse(time).utc },
+        "datetime"     => datetime_value = Proc.new { |time|    Time.xmlschema(time).utc rescue ::DateTime.parse(time).utc },
+        "dateTime"     => datetime_value,
```
byroot: What's this? An LLM halulu?

activesupport/lib/active_support/messages/metadata.rb
```diff
@@ -9,14 +9,14 @@ module Messages # :nodoc:
...
-      ENVELOPE_SERIALIZERS = [
+      ENVELOPE_SERIALIZERS = Array.new [
         *SerializerWithFallback::SERIALIZERS.values,
         ActiveSupport::JSON,
         ::JSON,
         Marshal,
       ]
 
-      TIMESTAMP_SERIALIZERS = [
+      TIMESTAMP_SERIALIZERS = Array.new [
```
byroot: I think we might as well use a `rubocop:disable` directly rather than to use a non-idiomatic construct to confuse the cop.

activesupport/lib/active_support/core_ext/time/conversions.rb
```diff
@@ -5,7 +5,7 @@
 require "active_support/values/time_zone"
 
 class Time
-  DATE_FORMATS = {
+  DATE_FORMATS = Hash.new.merge({
```
byroot: Same here `CONST = Hash.new.merge(literal)` is disgusting code that make 0 sense except to not get caught by the police, we shouldn't jump through such hoops.

byroot:
I haven't reviewed everything as the PR is a bit too big.

In short, everything that just add a `.freeze` to constants I'm fine with, and ideally would be in a standalone PR with just that.

For all the rest, most of the code seem quite suspect, and should be in smaller chunks so it can be reviewed properly.

activesupport/lib/active_support/xml_mini.rb
paracycle: Nope, I wrote it myself. I found it to be a compact way to share hash values between different keys without having to modify the hash after definition. Reminded me of the YAML syntax for anchors. 

Open to alternatives.

paracycle:
> In short, everything that just add a `.freeze` to constants I'm fine with, and ideally would be in a standalone PR with just that.
> 
> For all the rest, most of the code seem quite suspect, and should be in smaller chunks so it can be reviewed properly.

Happy to start with rubocop disable comments for all the violations in this PR and submit other changes individually. Does that work?

activesupport/lib/active_support/xml_mini.rb
byroot: The problem is that this emits a warning:

```ruby
p({
  "foo" => foo = 1, # warning: key "foo" is duplicated and overwritten on line 3
  "foo" => foo + 1,
})
```

byroot:
> Happy to start with rubocop disable comments for all the violations in this PR and submit other changes individually. Does that work?

Yeah, works for me.

byroot:
> other changes individually.

Also doesn't have to be exactly 1:1, as long as it's small enough chunks that it's reasonably reviewable.

railties/lib/rails/code_statistics.rb
paracycle: The weird thing is that the previous code was silently mutating the constants that were being passed in as the default value. Anyway, I've now changed these to be `rubocop:disable` instead.

railties/lib/rails/code_statistics.rb
paracycle: Agreed. I can look at improving the cop separately.

activesupport/lib/active_support/xml_mini.rb
paracycle: I think you might be missing the fact that the keys are not the same here. One is `datetime` and the other is `dateTime`. So it is more like:
```ruby
p({
  "foobar" => foo = 1,
  "fooBar" => foo,
})
```

activesupport/lib/active_support/messages/metadata.rb
paracycle: Done

activesupport/lib/active_support/core_ext/time/conversions.rb
paracycle: Reverted to `rubocop:disable`

activesupport/lib/active_support/xml_mini.rb
byroot: Oh wow, that was subtle.

guides/source/action_controller_advanced_topics.md
```diff
@@ -225,7 +225,7 @@ Implementing token authentication with Rails can be done using the
 
 ```ruby
 class PostsController < ApplicationController
-  TOKEN = "secret"
+  TOKEN = "secret".freeze
```
byroot: This is unfortunate :/

.rubocop.yml
```diff
@@ -172,6 +172,10 @@ Style/FrozenStringLiteralComment:
     - 'actiontext/db/migrate/**/*.rb'
     - '**/*.md'
 
+Style/MutableConstant:
+  Enabled: true
+  EnforcedStyle: literals
```
byroot, line 177:
```suggestion
  EnforcedStyle: literals
  Exclude:
    - '**/*.md'
```

So we don't impact docs?

byroot (approved):
Just one small thing, but otherwise 👍

guides/source/action_controller_advanced_topics.md
paracycle: Good catch, removed!

.rubocop.yml
paracycle: Added. Honestly, I wasn't expecting Markdown files to be checked by RuboCop. I see now that `rubocop-md` is activated.

byroot: Yep, that's why I wanted a smaller PR :)

railties/lib/rails/code_statistics.rb
paracycle: I've created a PR to fix this in RuboCop: https://github.com/rubocop/rubocop/pull/15166

PR #57326, opened 2026-05-09
Use megatest for Rails own testing

byroot:
Migrated Active Model and Active Record to iron things out.

Remaining blockers:

  - [x] ~~`rake test:*:isolated` is missing. Might need to add support for that specific behavior in megatest. But unsure how useful it really is. Might be the occasion rethink it?~~ After a chat with Rafael, we concluded they weren't that useful.
  - [ ] Fixture insertion need to be optimized, because megatest does randomize test order fully, not just on a per suite basis, meaning the tests that disable transactional fixtures get interwined much more with the others, causing fixtures cache to be invalidated way more often, this causes a significantly slowdown https://share.firefox.dev/4uD9jHN . At the very least we shouldn't reparse fixture files every time (~25% of runtime), and we can probably optimize the row insertion to (another ~25% of runtime).
  - [ ] Another issue with ordering and fixtures, is that the `@@already_loaded_fixtures` cache is keyed on a perfect match of the list of "fixtures set". Very few test suites have the same list of sets, so with all tests intertwined, the hit rate is extremely low.

A short term bandaid could be to add an option to order tests Minitest style, that would solve most of the perf difference. But longer term, improving fixtures is without a doubt the way to go.

TODO / nice to have:

  - [ ] Only `sqlite3_mem` support parallelism for now (but does yield a 3x speedup). Unlocking parallelism for other adapters would be nice, but not strictly a blocker.
  - [x] The Arel suite use `Minitest::Spec` syntax, so it has to be run standalone. Would be nice to either convert it (quite a big yak-shave) or to have some sort of adapter.


activemodel/test/cases/forbidden_attributes_protection_test.rb
```diff
@@ -25,7 +25,7 @@ def to_h
   end
 end
 
-class ActiveModelMassUpdateProtectionTest < ActiveSupport::TestCase
+class ActiveModelMassUpdateProtectionTest < ActiveModel::TestCase
```
byroot, line 28: We probably need a cop or something, because this is a bit too easy of a mistake to make without realizing.

rafaelfranca:
arel test suite was converted

byroot:
> arel test suite was converted

❤️

byroot:
Found a test leak: `bin/test --fail-fast -a mysql2 --seed 27831`

byroot:
> Found a test leak: 

Fixed in https://github.com/rails/rails/pull/57352

PR #57339, opened 2026-05-11
Add ActionController::Parameters#deep_transform_values

Edilbek:
## Motivation / Background

This PR has been created because I'd like a safer way of running a deep value transformation on `ActionController::Parameters`. Today, the only option is `params.to_unsafe_h.deep_transform_values { ... }`, which discards the permitted/unpermitted distinction and returns a plain `Hash`. This forces application code to bypass the strong-parameters guardrails for a routine normalization task (e.g., stripping whitespace from every nested string before validation).

`ActionController::Parameters` already exposes `deep_transform_keys` and `deep_transform_keys!`, and Active Support already extends `Hash` with `deep_transform_values`. This PR completes the symmetry by adding `deep_transform_values` and `deep_transform_values!` to `Parameters`, matching `Hash#deep_transform_values` while keeping the result inside `ActionController::Parameters` with `permitted?` preserved.

## Details

```ruby
params = ActionController::Parameters.new(
  user: {
    email: "  ALICE@EXAMPLE.COM  ",
    profile: { bio: "  Hello world  " }
  }
)

params.deep_transform_values { |v| v.is_a?(String) ? v.strip.downcase : v }
# => #<ActionController::Parameters {"user"=>#<ActionController::Parameters {"email"=>"alice@example.com", "profile"=>#<ActionController::Parameters {"bio"=>"hello world"} permitted: false>} permitted: false>} permitted: false>

# The bang variant mutates in place
params.deep_transform_values! { |v| v.is_a?(String) ? v.strip : v }

# permitted? status is carried over
params.permit!
params.deep_transform_values { |v| v }.permitted? # => true


Edilbek:
The failing activerecord trilogy (4.0) [mariadb] job doesn't appear to be related to this PR.

PR #57332, opened 2026-05-10
fix: unfreeze Date::DATE_FORMATS

luizkowalski:
### Motivation / Background

reverts the unintended `freeze` introduced by 34bce3f693b354f2172d8cf48648fe068605b2d5, following what was done in `activesupport/lib/active_support/core_ext/time/conversions.rb`

### Detail

Freezing DATE_FORMATS breaks adding new formats, as shown by the docs in the very same constant

### Checklist

Before submitting the PR make sure the following are checked:

* [X] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [X] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [X] Tests are added or updated if you fix a bug or add a feature.
* [ ] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #57341, opened 2026-05-11
Add tests for constants that are expected to be mutated

paracycle:
### Motivation / Background

Since [the PR freezing mutable constants](https://github.com/rails/rails/pull/57323) landed, there have been 2 PRs (#57332 and #57333) fixing cases where the constants were expected to be mutated by framework users, but there were no tests encoding those expectation.

### Detail

This Pull Request adds 2 tests for those missing mutable constant expectations

### Checklist

### DIFF
--- a/activerecord/test/cases/adapters/abstract_mysql_adapter/sql_types_test.rb
+++ b/activerecord/test/cases/adapters/abstract_mysql_adapter/sql_types_test.rb
@@ -10,6 +10,14 @@ class SqlTypesTest < ActiveRecord::AbstractMysqlTestCase
     assert_equal "blob", type_to_sql(:binary)
   end
 
+  def test_native_database_types_can_be_extended
+    ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter::NATIVE_DATABASE_TYPES[:custom_mysql_type] = { name: "mediumtext" }
+
+    assert_equal "mediumtext", type_to_sql(:custom_mysql_type)
+  ensure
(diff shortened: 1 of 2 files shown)


activerecord/test/cases/adapters/abstract_mysql_adapter/sql_types_test.rb
```diff
@@ -10,6 +10,14 @@ def test_binary_types
     assert_equal "blob", type_to_sql(:binary)
   end
 
+  def test_native_database_types_can_be_extended
+    ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter::NATIVE_DATABASE_TYPES[:custom_mysql_type] = { name: "mediumtext" }
```
kamipo, line 14: How about making all `NATIVE_DATABASE_TYPES` mutable?
Currently, only the `NATIVE_DATABASE_TYPES` for the sqlite3 adapter is frozen.

paracycle: Honestly, I think we should deprecate the mutation of constants as public API and provide better APIs that don't require any constant mutations.

byroot: Yes these should be class instance variables with a public class method to register new entries

byroot:
Merging as is just for coverage, but as mentioned we should deprecate these constants in favor of proper registration API

PR #57345, opened 2026-05-11
Introduce `ActiveSupport::TimeFormats` and `ActiveSupport::DateFormats` for registering custom date formats

paracycle:
### Motivation / Background

This Pull Request has been created in response to the comments by @byroot [here](https://github.com/rails/rails/pull/57332#issuecomment-4415097265) and [here](https://github.com/rails/rails/pull/57341#discussion_r3219396920).

### Detail

This Pull Request changes allows adding custom date formats to `to_fs` without modifying the global `Time::DATE_FORMATS` or `Date::DATE_FORMATS` hashes. Custom formats are added via `ActiveSupport::TimeFormats.register` or `ActiveSupport::DateFormats.register`, and will only be available to `Time#to_fs`, and `Date#to_fs` respectively.

At the same time, the existing `Time::DATE_FORMATS` and `Date::DATE_FORMATS` constants are still supported for backward compatibility, but they are now deprecated and will be removed in the next version of Rails. This encourages users to migrate to the new approach for better encapsulation and to avoid potential conflicts with other libraries that may also modify the global date formats.

```ruby
ActiveSupport::TimeFormats.register(:month_and_year, '%B %Y')
ActiveSupport::DateFormats.register(:short_ordinal, ->(date) { date.strftime("%B #{date.day.ordinalize}") })

Time.now.to_fs(:month_and_year) # => "February 2024"
Date.today.to_fs(:short_ordinal)  # => "February 21st"
```

### Additional information

The `ActiveSupport::TimeFormats.list` and `ActiveSupport::DateFormats.list` are initialized to `Time::DATE_FORMATS` and `Date::DATE_FORMATS` respectively, and **not** frozen, so that appending to the constants can still modify the lists until the constants are removed completely. This is important for backward compatibility.

However, as soon as a call to the respective `register` method is made, the `list` is modified and frozen considering that the user has opted in to the new API.

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,3 +1,32 @@
+*   Introduce `ActiveSupport::TimeFormats` and `ActiveSupport::DateFormats`
+    for registering custom date formats.
+
+    This allows adding custom date formats to `to_fs` without modifying the
+    global `Time::DATE_FORMATS` or `Date::DATE_FORMATS` hashes. Custom formats
+    are added via `ActiveSupport::TimeFormats.register` or
+    `ActiveSupport::DateFormats.register`, and will only be available to
+    `Time#to_fs`, and `Date#to_fs` respectively.
+
+    At the same time, the existing `Time::DATE_FORMATS` and `Date::DATE_FORMATS`
+    constants are still supported for backward compatibility, but they are now
+    deprecated and will be removed in the next version of Rails. This encourages
+    users to migrate to the new approach for better encapsulation and to avoid
+    potential conflicts with other libraries that may also modify the global
+    date formats.
+
+    ```ruby
+    ActiveSupport::TimeFormats.register(:month_and_year, '%B %Y')
+    ActiveSupport::DateFormats.register(
+      :short_ordinal,
+      ->(date) { date.strftime("%B #{date.day.ordinalize}") }
+    )
+
+    Time.now.to_fs(:month_and_year) # => "February 2024"
+    Date.today.to_fs(:short_ordinal)  # => "February 21st"
+    ```
+
+    *Ufuk Kayserilioglu*
+
 *   Add `start_day` argument to `this_week?` for consistency with `all_week`
 
     `this_week?` now accepts an optional `start_day` argument, matching the
--- a/activesupport/lib/active_support/core_ext/time/conversions.rb
+++ b/activesupport/lib/active_support/core_ext/time/conversions.rb
@@ -45,15 +45,15 @@ class Time
   #   time.to_fs(:iso8601)      # => "2007-01-18T06:10:17-06:00"
   #
   # == Adding your own time formats to +to_fs+
-  # You can add your own formats to the +Time::DATE_FORMATS+ hash.
-  # Use the format name as the hash key and either a strftime string
+  # You can add your own formats using the +ActiveSupport::TimeFormats.register+ method.
+  # Use the format name as the name and either a strftime string
   # or Proc instance that takes a time argument as the value.
   #
   #   # config/initializers/time_formats.rb
-  #   Time::DATE_FORMATS[:month_and_year] = '%B %Y'
-  #   Time::DATE_FORMATS[:short_ordinal]  = ->(time) { time.strftime("%B #{time.day.ordinalize}") }
+  #   ActiveSupport::TimeFormats.register(:month_and_year, '%B %Y')
+  #   ActiveSupport::TimeFormats.register(:short_ordinal, ->(time) { time.strftime("%B #{time.day.ordinalize}") })
   def to_fs(format = :default)
-    if formatter = DATE_FORMATS[format]
+    if formatter = ::ActiveSupport::TimeFormats.lookup(format)
       formatter.respond_to?(:call) ? formatter.call(self).to_s : strftime(formatter)
     else
       to_s
--- a/activesupport/lib/active_support/date_formats.rb
+++ b/activesupport/lib/active_support/date_formats.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "active_support/core_ext/date/conversions"
+
+module ActiveSupport
+  module DateFormats
+    @list = Date::DATE_FORMATS.dup.freeze
+    @deprecated_list = Date::DATE_FORMATS
+    Date.deprecate_constant :DATE_FORMATS
+
+    # :nodoc:
+    def self.lookup(format)
+      @list[format] || @deprecated_list[format]
+    end
+
+    # Registers a new date format for formatting `Date` instances.
+    # See +Date::DATE_FORMATS+ for built-in formats.
+    # Use the format name as the name and either a strftime string or
+    # Proc instance that takes a date argument as the value.
+    def self.register(name, format)
+      @list = @list.merge(name => format).freeze
+    end
+  end
+end
--- a/activesupport/lib/active_support/time_with_zone.rb
+++ b/activesupport/lib/active_support/time_with_zone.rb
@@ -223,11 +223,11 @@ module ActiveSupport
     # Accepts an optional <tt>format</tt>:
     # * <tt>:default</tt> - default value, mimics Ruby Time#to_s format.
     # * <tt>:db</tt> - format outputs time in UTC :db time. See Time#to_fs(:db).
-    # * Any key in +Time::DATE_FORMATS+ can be used. See active_support/core_ext/time/conversions.rb.
+    # * Any key in +ActiveSupport::TimeFormats+ can be used. See active_support/core_ext/time/conversions.rb.
     def to_fs(format = :default)
       if format == :db
         utc.to_fs(format)
-      elsif formatter = ::Time::DATE_FORMATS[format]
+      elsif formatter = ::ActiveSupport::TimeFormats.lookup(format)
         formatter.respond_to?(:call) ? formatter.call(self).to_s : strftime(formatter)
       else
         to_s
--- a/activerecord/lib/active_record/integration.rb
+++ b/activerecord/lib/active_record/integration.rb
@@ -10,7 +10,7 @@ module ActiveRecord
       ##
       # :singleton-method:
       # Indicates the format used to generate the timestamp in the cache key, if
-      # versioning is off. Accepts any of the symbols in +Time::DATE_FORMATS+.
+      # versioning is off. Accepts any of the symbols in +ActiveSupport::TimeFormats+.
       #
       # This is +:usec+, by default.
       class_attribute :cache_timestamp_format, instance_writer: false, default: :usec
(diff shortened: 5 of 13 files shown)


byroot:
So how I was picturing this new API, was more akin to:

```ruby
module ActiveSupport
  module DateFormats
    @list = {
      # defaults
    }.freeze

   class << self
     attr_reader :list # :nodoc: 
     def register(name, value)
       @list = @list.merge(name => value).freeze
     end
   end
  end
end
```

This way:

  - We remove some constants / methods from core types.
  - The configuration is always frozen, hence always available from secondary ractors (which is your goal AFAIK)

paracycle:
@byroot Yeap, that sounds good to me. However, I think we still need to keep the current constant around until we can remove it, right? Which reminds me that modifying that constant should still result in the old behaviour until that time as well, I guess. This PR isn't quite doing that either yet.

byroot:
> think we still need to keep the current constant around until we can remove it, right? Which reminds me that modifying that constant should still result in the old behaviour until that time as well, I guess.

Yes on both counts.

paracycle:
Ok, I think I have the API that you had in mind implemented and added a note in the PR description about why I have not frozen the initial value of the class ivar.

rafaelfranca:
Why do we need this constant to me mutated? If people want formatted date and time objects they should probably using i18n? Maybe we should just deprecate the constant and tell people to use `I18n.localize`?

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+*   Introduce `Time::Formats` and `Date::Formats` for registering custom date formats.
+
+    This allows adding custom date formats to `to_fs` without modifying the global `Time::DATE_FORMATS` or `Date::DATE_FORMATS` hashes. Custom formats are added via `Time::Formats.register` or `Date::Formats.register`, and will only be available to `Time#to_fs`, and `Date#to_fs` respectively.
+
+    At the same time, the existing `Time::DATE_FORMATS` and `Date::DATE_FORMATS` constants are still supported for backward compatibility, but they are now deprecated and will be removed in the next version of Rails. This encourages users to migrate to the new approach for better encapsulation and to avoid potential conflicts with other libraries that may also modify the global date formats.
```
p8: Hi @ufuk. Could you wrap the changelog entry to around 80 characters?
See: https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#updating-the-changelog

byroot:
> If people want formatted date and time objects they should probably using i18n?

I'll have a look tomorrow, but I'm pretty sure Intercom has such custom format, IIRC Shopify too?

I think this may not exactly map `1:1` with I18n, as it's not for a particular locale, but for a particular context (e.g. `time.to_fs(:some_provider_api)`)

rafaelfranca:
I don't see any case were `time.to_fs(:some_provider_api)` can't be `I18n.locale(time, format: :some_provider_api)`

matthewd:
I feel like I18n is more suited to user-facing strings, rather than programmatic formatting. 

IMO we need a pretty high bar to justify dropping an entire API feature, especially here where we've already fairly-recently bumped anyone-using-this through a deprecation (`to_s` -> `to_fs`) :confused:

activesupport/CHANGELOG.md
paracycle: Done, thank you!

activesupport/lib/active_support/core_ext/time/conversions.rb
```diff
@@ -26,6 +26,24 @@ class Time
     iso8601: lambda { |time| time.iso8601 }
   }
 
+  module Formats
```
byroot: So my thinking was that if we're going to deprecate the old constant, we might as well not put the new one into core classes like `Time`. That's why my snippet used a module under `ActiveSupport`. Could also just as well be a config on `ActiveSupport` itself.

paracycle: Right, I see. However, we _do_ need two separate lists, one for `Time` formatting and the other for `Date`s, so we can do `ActiveSupport::DateFormats` and `ActiveSupport::TimeFormats`. Alternatively, as you suggest, we can add `ActiveSupport.register_date_format` and `ActiveSupport.register_time_format` with corresponding `ActiveSupport.date_formats` and `ActiveSupport.time_formats` lists.

I think I personally prefer the namespaces over the top-level config. WDYT?

byroot: > we can do ActiveSupport::DateFormats and ActiveSupport::TimeFormats.

Yep, perfectly fine by me.

luizkowalski:
I was working on a similar PR during the weekend, but this one looks better 😄 

one thing I thought was that it would be nice if one could register the time formats on `application.rb` as well, something along these lines

```
initializer "active_support.register_date_and_time_formats" do |app|
  config.after_initialize do
    app.config.active_support.date_formats&.each do |name, format|
      Time::Formats.register(name, format)
    end
    app.config.active_support.time_formats&.each do |name, format|
      Time::Formats.register(name, format)
    end
  end
end
```

that would avoid one more file on the initializers folders

activesupport/lib/active_support/time_with_zone.rb
```diff
@@ -223,11 +223,11 @@ def to_s
     # Accepts an optional <tt>format</tt>:
     # * <tt>:default</tt> - default value, mimics Ruby Time#to_s format.
     # * <tt>:db</tt> - format outputs time in UTC :db time. See Time#to_fs(:db).
-    # * Any key in +Time::DATE_FORMATS+ can be used. See active_support/core_ext/time/conversions.rb.
+    # * Any key in +ActiveSupport::TimeFormats.list+ can be used. See active_support/core_ext/time/conversions.rb.
     def to_fs(format = :default)
       if format == :db
         utc.to_fs(format)
-      elsif formatter = ::Time::DATE_FORMATS[format]
+      elsif formatter = ::ActiveSupport::TimeFormats.list[format]
```
byroot: I think this is insufficient. The constant is deprecated, but it should still work. Whereas right now it no longer does. I think we can handle that with something like: 

```suggestion
      elsif formatter = (::ActiveSupport::TimeFormats.list[format] || ::Time::DATE_FORMATS[format])
```

But we need a way to read the constant without emitting a deprecation, which we can probably do by keeping a non-deprecated reference to the hash in the constant.

e.g. 

```ruby
    @list = Time::DATE_FORMATS.dup.freeze
    @deprecated_list = Time::DATE_FORMATS
    Time.deprecate_constant :DATE_FORMATS
```

and then:

```suggestion
      elsif formatter = (::ActiveSupport::TimeFormats.list[format] || ::ActiveSupport::TimeFormats.deprecated_list[format])
```

byroot (changes requested):
Sorry for the late review, I was AFK for a while.

activesupport/lib/active_support/time_with_zone.rb
paracycle: Yeap, this makes sense. The deprecated constant was working until someone called `register` after which it stopped working in this implementation. I figured that someone calling `register` meant that they opted into the new API, but it could be a gem or something similar that makes the call before the application has a chance to register their own formats via the constant, so this makes sense.

I have removed the `list` method and encapsulated the format lookup with fallback inside a `lookup` method on each format module. Hope that makes sense.

byroot:
LGTM, but can't merge, and can't fix conflicts myself: 

Capture d’écran 2026-05-19 à 20 20 03

Edgarruiz8585:
+ > by: Edgar Manuel Ruiz Arias @Edgarruiz8585
***@***.*** ***@***.***
***@***.*** ***@***.***

paracycle:
@byroot I resolved the conflict. Thanks!

activesupport/lib/active_support/date_formats.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module ActiveSupport
+  module DateFormats
+    @list = Date::DATE_FORMATS.dup.freeze
+    @deprecated_list = Date::DATE_FORMATS
+    Date.deprecate_constant :DATE_FORMATS
+
+    # :nodoc:
+    def self.lookup(format)
+      @list[format] || @deprecated_list[format]
```
rafaelfranca, line 13: What is this logic here? Since `@deprecated_list` is the same as `@list` why do we need `deprecated_list`?

activesupport/lib/active_support/date_formats.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require "active_support/core_ext/date/conversions"
+
+module ActiveSupport
+  module DateFormats
+    @list = Date::DATE_FORMATS.dup.freeze
+    @deprecated_list = Date::DATE_FORMATS
+    Date.deprecate_constant :DATE_FORMATS
+
+    # :nodoc:
```
rafaelfranca, line 11: This made all method nodoc, no? Even if not, `# :nodoc:` goes on the side of the method name.

activesupport/lib/active_support/date_formats.rb
rafaelfranca: Also, there is no test for this logic

byroot: > Since @deprecated_list is the same as @list why do we need deprecated_list?

They're initially the same, but if someone mutates the old constant `Date::TIME_FORMAT[:foo] = blah"`, they'll diverge and that format will only be in `@deprecated_list`.

activesupport/lib/active_support/date_formats.rb
byroot: Good catch.

activesupport/lib/active_support/date_formats.rb
paracycle: `@list` is the frozen copy of the format list that we replace as new formats are registered using `register`. `@deprecated_list` is the list that references the old constant, so that we still respect additions to the constant if any are made, until we remove the constant completely. The latter exists so that we don't reference the `TIME_FORMATS` constant in the fallback and trigger the deprecation warning internally.

activesupport/lib/active_support/date_formats.rb
paracycle: Sorry, my bad. I'll fix it in a minor PR.

activesupport/lib/active_support/date_formats.rb
byroot: > lso, there is no test for this logic

Hum, while working on adding a test for the deprecation, I realized this is using Ruby's `deprecate_constant`, I'll see about using the Active Support one so it can be sent to the proper `Deprecator` instance.

byroot: It's getting late here, so I'll finish tomorrow, but this should do: https://github.com/rails/rails/pull/57405

PR #57348, opened 2026-05-11
[RF-Docs] [ci-skip] Update Active Record Composite Primary Keys Guide 

bhumi1102:
This PR is part of the Rails Foundation documentation project and updates the Active Record Composite Primary Keys Guide. One of the main updates include adding more code examples.


guides/source/active_record_composite_primary_keys.md
```diff
@@ -19,69 +20,137 @@ After reading this guide you will be able to:
...
+end
+```
+
+This keeps `id` as the primary key at the database level while instructing Active Record to always include `company_id` in queries, updates, and deletes.
+
+For example, for `Developer.find(1)`, this is what the query looks like *with* the above `query_constraints`:
+
+```sql
+SELECT * FROM developers WHERE company_id = 5 AND id = 1
+```
```
fatkodima: This example is wrong. Where `company_id = 5` came from? I think an example of updating an existing record should be used.

guides/source/active_record_composite_primary_keys.md
```diff
@@ -143,79 +204,55 @@ class Book < ApplicationRecord
 end
 ```
 
-In this setup, `Order` has a composite primary key consisting of `[:shop_id,
-:id]`, and `Book` belongs to `Order`. Rails will assume that the `:id` column
-should be used as the primary key for the association between an order and its
-books. It will infer that the foreign key column on the books table is
-`:order_id`.
+For the association above, `Order` has a composite primary key of `[:store_id, :id]`, and `Book` belongs to `Order`. Rails infers that the foreign key column on the books table is `order_id`, and will use only the `id` portion of the composite key when querying the association.
```
fatkodima:
```suggestion
For the association above, `Order` has a composite primary key of `[:store_id, :id]`, and `Book` belongs to `Order`. Rails infers that the foreign key column on the `books` table is `order_id`, and will use only the `id` portion of the composite key when querying the association.
```

guides/source/active_record_composite_primary_keys.md
```diff
@@ -19,69 +20,137 @@ After reading this guide you will be able to:
...
+```
+
+This tells Active Record that records are uniquely identified by the combination of both columns, not a single `id`.
+
+In most cases, you don't need to declare a composite primary key with
+`self.primary_key` in your model at all. If you define the composite primary key
+in your migration and use the default `schema.rb` format, Rails will read the
+primary key from the schema automatically at boot time and your model will work.
+
+On the other hand, if your application uses `structure.sql` instead of `schema.rb` or if you're connecting to a legacy or external database that was not created by your migrations, you can use `self.primary_key` to explicitly declare a composite primary key.
```
fatkodima: Mentioning and the distinction between `schema.rb` and `structure.sql` files is wrong. This files are used only for developer convenience - so a new developer can locally create a new database with the correct schema (avoiding running migrations from the beginning) or other developers can recreate the database.

Rails does not use any of these to get the tables, columns etc schema info. It infers the schema by reading internal catalogs directly from the database.

guides/source/active_record_composite_primary_keys.md
bhumi1102: updated

guides/source/active_record_composite_primary_keys.md
bhumi1102: updated

guides/source/active_record_composite_primary_keys.md
```diff
@@ -354,3 +393,20 @@ alices_adventure_in_wonderland_in_books:
...
+A composite primary key creates a database index on its columns in the order they are declared. A CPK of `[:store_id, :sku]` means the database can efficiently use that index for queries filtering on `store_id` alone, or `store_id` and `sku` together, but not `sku` alone. The leading column should be the one you filter on most frequently. In a multi-tenant application, placing the tenant identifier first (e.g. `store_id`) makes sense, since almost every query will be scoped to a store.
+
+### Index Foreign Key Columns Manually
+
+When another table references a composite primary key, the foreign key columns on that table need their own index. Unlike single column foreign keys, Rails does not add these automatically. Without an explicit index, any join or association query back to the parent table will result in a full table scan.
+
+You can add the index manually in your migration:
+
+```ruby
+add_index :books, [:order_store_id, :order_number]
```
p8, line 411: Should this be `order_sku` to match the `[:store_id, :sku]` example above?
```suggestion
add_index :books, [:order_store_id, :order_sku]
```

guides/source/active_record_composite_primary_keys.md
```diff
@@ -112,58 +190,37 @@ The SQL equivalent of the above is:
...
 
-Take caution when using `find_by(id:)` on models where `:id` is not the primary
-key, such as composite primary key models. See the [Active Record Querying][]
-guide to learn more.
-
-[Active Record Querying]: active_record_querying.html#conditions-with-id
+WARNING: When using  `where` or `find_by`, the key `id` matches against an `:id` attribute on the model only (if the model has one). It does not resolve to the full composite primary key the way `find` does. Use `find` when you want to look up a record by its full composite primary key. See the [Active Record Querying](active_record_querying.html#conditions-with-id) guide for more detail.
 
 Associations between Models with Composite Primary Keys
 -------------------------------------------------------
```
p8, line 215: Since this is the "Composite Primary Keys" guide and associations are always between models, maybe we can remove some redundancy and shorten this long title?
```suggestion
Associations
------------
```

guides/source/active_record_composite_primary_keys.md
```diff
@@ -19,69 +20,146 @@ After reading this guide you will be able to:
...
+or in applications where the schema has been designed to partition data across
+[multiple databases (sharding)](active_record_multiple_databases.html) or
+isolate data per customer or tenant (multitenancy). Composite primary keys is an
+Active Record feature that spans migrations, query methods, and associations.
+
+NOTE: Composite primary keys do increase complexity and can be slower than a
+single primary key column. Ensure your use case requires a composite primary key
 before using one.
 
+### Using `query_constraints` as an Alternative
```
p8, line 36: Ah nice, I didn't know about `query_constraints`.

This section goes into detail about `query_constraints` before describing the details of composite primary keys. As a reader it is good to know about alternatives, but I probably first want to read about composite primary keys.
Could we maybe mention them briefly here first (without any code examples) and link to a section describing the details further on, after composite primary keys have been described in detail? This makes it easier to compare them (I don't know the details yet here to compare them) and feels less like we try to sidetrack the reader.

guides/source/active_record_composite_primary_keys.md
```diff
@@ -19,69 +20,146 @@ After reading this guide you will be able to:
...
+The above Active Record method results in the following SQL:
 
 ```sql
 SELECT * FROM products WHERE store_id = 3 AND sku = "XYZ12345"
 ```
 
-To find multiple records with composite IDs, pass an array of arrays to `#find`:
+NOTE: The `find` method expects the values in the same order as the columns were declared in `primary_key` when querying with composite primary key. Also, if only one value is provided, such as `Product.find(3)`, then Rails raises an `ActiveRecord::RecordNotFound` error.
+
+To find multiple records with composite IDs, you can pass an array of arrays to `find`. For example, to find the products with primary keys [1, "ABC98765"] and [7, "ZZZ11111"]
```
p8, line 149: Maybe wrap these arrays with code fences?
```suggestion
To find multiple records with composite IDs, you can pass an array of arrays to `find`. For example, to find the products with primary keys `[1, "ABC98765"]` and `[7, "ZZZ11111"]`
```

guides/source/active_record_composite_primary_keys.md
bhumi1102: > feels less like we try to sidetrack the reader.

Fair enough :) I added this additional paragraph with an example for cpk to make it more concreate before going into the subsection about `query_constraints`. 

For example, an inventory tracking application may have a table where each row
is identified by a `store_id` and a product's `sku` (stock keeping unit)
together but neither column alone identifies a row on its own. Active Record
supports this by allowing you to declare `[:store_id, :sku]` as the primary key
as detailed in a
[section](#declaring-composite-primary-keys-and-creating-migrations) below.

bhumi1102:
HI @p8, I've addressed the last few comments (that I had missed earlier), wrapped column at 80, and also merged `main`.

guides/source/active_record_composite_primary_keys.md
```diff
@@ -19,69 +21,177 @@ After reading this guide you will be able to:
...
+effective identity when querying.
+
+
+Declaring Composite Primary Keys and Creating Migrations
+--------------------------------------------------------
 
-You can create a table with a composite primary key by passing the
-`:primary_key` option to `create_table` with an array value:
+To create a table with a composite primary key, you can pass an array to the
+`primary_key:` option in the database migration:
```
p8: As I understand it (reading this guide), composite primary keys are typically declared by the migration for the new table. Calling `self.primary_key=` is for a legacy table.

So maybe we can shorten the heading here, as the migration typically declares the primary key?
And maybe mention they are detected by Rails automatically at the beginning?
This makes it a bit clearer the migration is all you need in most cases.
I copied this sentence from a parapgraph further down.

```suggestion
Declaring Composite Primary Keys
--------------------------------

Rails detects the primary key automatically from the database at runtime.
To create a table with a composite primary key, you can pass an array to the
`primary_key:` option in the database migration:
```

guides/source/active_record_composite_primary_keys.md
p8: Thanks!

guides/source/active_record_composite_primary_keys.md
bhumi1102: Updating! - made the title shorter. Updated wording in that section a bit. It still makes sense to keep the "automatically detected" part _after_ the migration since that's the primary message to the reader is: write a migration --> schema.rb gets the cpk. Let me know if this addresses what you're thinking!

PR #50623, opened 2024-01-06
Pass render options and block to calls to `#render_in`

seanpdoyle:
### Motivation / Background

Closes [#45432][]

Support for objects that respond to `#render_in` was introduced in [#36388][] and [#37919][]. Those implementations assume that the instance will all the context it needs to render itself. That assumption doesn't account for call-site arguments like `locals: { ... }` or a block.

### Detail

This commit expands support for rendering with a `:renderable` option to incorporate locals and blocks. For example:

```ruby
class Greeting
  def render_in(view_context, **options, &block)
    if block
      view_context.render html: block.call
    else
      view_context.render inline: <<~ERB.strip, **options
        Hello, <%= name %>
      ERB
    end
  end
end

render(Greeting.new)                    # => "Hello, World"
render(Greeting.new, name: "Local")     # => "Hello, Local"
render(Greeting.new) { "Hello, Block" } # => "Hello, Block"
```

Since existing tools depend on the `#render_in(view_context)` signature
without options, this commit deprecates that signature in favor of one
that accepts options and a block.

[#45432]: https://github.com/rails/rails/pull/45432
[#36388]: https://github.com/rails/rails/pull/36388
[#37919]: https://github.com/rails/rails/pull/37919

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #46202, opened 2022-10-06
PoC: AbstractController::Base#renderer_for

casperisfine:
### Context 

There has been some demands from view systems not based on paths, to offer some extension API that are better suited for them: https://discuss.rubyonrails.org/t/replace-to-partial-path-calls-with-a-more-abstract-to-renderable/81457

I'm in favor of the general idea, but not at all with the proposed solutions so far.

### Existing API

Currently you can customize the behavior of `render some_object` in two ways:

- The object can respond to `to_partial_path` and return a view path.
- The object can respond to `render_in` and return a string representation.

What I don't like with both of these APIs is that they ask the object how to render itself. Applications often have several different ways to represent the same object based on the context, e.g. frontend vs admin panel, or HTML vs JSON.

### Proposal

It would be way more powerful if the view or controller was in charge of selecting the proper renderer for a given object.

e.g. (pseudo code)

```ruby
module Admin
  class BaseController < ApplicationController
    def renderer_for(object)
      { partial: "admin/#{object.class.underscore}" }
    end
  end
```

On paper, you'd probably want to define this on thew view, but in practive Rails users very rarely interact with the view, hence why I do this on the controller. But I'm maybe missing something.

Another benefit of such API would be to allow view component systems to better integrate with Rails by implementing their own conventions, e.g. (pseudo code)

```ruby
module MyViewComponentExtension
  def renderer_for(object)
    component_name = "#{object.class.name}Component"
    "#{self.class.module_parent_name}#{component_name}".safe_constantize ||
      component_name.constantize
  end
end
```

It's somewhat similar to what Active Model Serializer used to do for its default serializer system:
https://github.com/rails-api/active_model_serializers/blob/0-9-stable/lib/action_controller/serialization.rb

It used to work quite well.

### Note

his is purely a proof of concept to demonstrate the general idea, what capabilities it brings to the table.

I'm absolutely not happy with that implementation and I'm not suggesting to merge it. If we chose to go with the idea, we
should carefully study what the API would look like.

PR #57349, opened 2026-05-12
Add default `#render_in` implementation to `ActiveModel::Conversion`

seanpdoyle:
### Motivation / Background

Follow-up to [#50623][]
Follow-up to [#46202][]

The `#render_in` method was introduced as an integration touch-point for views to intercept object rendering conventionally handled by Action View. 

This PR aims to provide Active Model-level (and therefore Active Record-level) flexibility to control how Action View transform models into HTML and JSON strings.

For example, users of the [view_component](https://github.com/ViewComponent/view_component) gem can "intercept" model rendering to directly an instance to a `ViewComponent::Base` instance.

### Detail

With the following view partial:

```erb
<%# app/views/people/_person.html.erb %>

<%= local_assigns[:shout] ? person.name.upcase : person.name %>
```

Callers can render an instance of `Person` as a positional argument or a `:renderable` option:

```ruby
person = Person.new(name: "Ralph")

render person                                       # => "Ralph"
render person, shout: true                          # => "RALPH"
render renderable: person                           # => "Ralph"
render renderable: person, locals: { shout: true }  # => "RALPH"
```

This change aims to preserve backward compatibility by providing a default `#render_in` method to relies on render the instance with a view partial inferred from combining `partial: to_partial_path` and `object: self`.

[#50623]: https://github.com/rails/rails/pull/50623
[#46202]: https://github.com/rails/rails/pull/46202#issuecomment-1281993332

### Checklist

### DIFF
--- a/actionview/test/actionpack/controller/render_test.rb
+++ b/actionview/test/actionpack/controller/render_test.rb
@@ -592,6 +592,10 @@ class TestController < ActionController::Base
     render partial: "hash_greeting", collection: [ { first_name: "Pratik" }, { first_name: "Amy" } ], locals: { greeting: "Hola" }
   end
 
+  def renderable_hash
+    render renderable: Quiz::Question.new(params[:name])
+  end
+
   def partial_with_implicit_local_assignment
     @customer = Customer.new("Marcel")
     render partial: "customer"
@@ -725,6 +729,7 @@ class RenderTest < ActionController::TestCase
     get :partial_with_nested_object_shorthand, to: "test#partial_with_nested_object_shorthand"
     get :partial_with_hashlike_locals, to: "test#partial_with_hashlike_locals"
     get :partials_list, to: "test#partials_list"
+    get :renderable_hash, to: "test#renderable_hash"
     get :render_action_hello_world, to: "test#render_action_hello_world"
     get :render_action_hello_world_as_string, to: "test#render_action_hello_world_as_string"
     get :render_action_hello_world_with_symbol, to: "test#render_action_hello_world_with_symbol"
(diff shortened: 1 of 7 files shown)


claudiob:
@seanpdoyle Thanks for this PR! Unfortunately this broke my Rails app. 

I'm going to try to create a trimmed down version of the app to better identify the issue. 
Meanwhile I'll describe it here, maybe this can help already.

---

I have a standard route/controller for an `update` action. The `update` action is defined like this:

```ruby
# app/controllers/api/bookings_controller.rb
class API::BookingsController < ApplicationController
  def create
    @booking = Booking.find params.expect(:id)
    @booking.update update_params
    response.status = :ok
  end
end
```

Since I'm not calling `render`, Rails automatically find the matching view to render, a JBuilder file:

```ruby
# app/views/api/bookings/update.json.jbuilder
json.partial! @booking
```

Since I'm not specifying which partial to render, Rails would automatically find this matching one:

```ruby
# app/views/api/bookings/_booking.json.jbuilder
json.extract! booking, :id, :status
```

This last part is what's broken after this PR. `json.partial! @booking` doesn't look for the `_booking.json.builder` partial in the same folder:

```
Error:
API::BookingsControllerTest#test_PUT_/api/bookings_responds_with_200_with_valid_authentication_and_params:
RuntimeError: Neutered Exception ActionView::Template::Error: Missing partial bookings/_booking with {locale: [:en], formats: [:json], variants: [], handlers: [:raw, :erb, :html, :builder, :ruby, :jbuilder]}.

Searched in:
  * "/Users/claudiob/code/fountain/app/views"
  * "/Users/claudiob/source/rails/actiontext/app/views"
  * "/Users/claudiob/source/rails/actionmailbox/app/views"

    app/views/api/bookings/update.json.jbuilder:1
    test/controllers/api/bookings_controller_test.rb:33:in 'block in <class:BookingsControllerTest>'
    test/controllers/api/base_controller_test.rb:6:in 'block in <class:BaseControllerTest>'
```

---

Let me know if any of this rings a bell. Happy to share a more complete example if that's needed.
Thank you 🙇

seanpdoyle:
@claudiob thank you for sharing that context. 

@guilleiguaran I've opened https://github.com/rails/rails/pull/57374 to revert this PR. I'll open a follow-up to incorporate tests that exercise the use-case outlined by @claudiob.

PR #57350, opened 2026-05-12
Auto-add aria-current="page" to link_to on the current page

NotGrm:
### Motivation / Background

  This Pull Request has been created because there is no built-in way to mark the link to the current page in link_to. Apps that build navigation menus today have two options, both unsatisfying:

  - Compute the active state per-link and toggle a .active class (boilerplate, easy to forget).
  - Use link_to_unless_current, which removes the link entirely on the current page — a regression for screen-reader users and a styling hassle (the markup changes shape).

  Adding aria-current="page" automatically when the link's target matches current_page? solves both problems at once:

  - Accessibility: screen readers announce the current page without any extra work from the developer.
  - Styling: apps can highlight the active nav item from CSS alone (nav a[aria-current="page"] { … }), with no class toggling and no conditional helpers.

  This is a small, focused addition; the per-call :current option keeps callers in full control when they need different semantics (step, location, …) or want to opt out.

Duplicate of #57343 that I closed by mistake

 ### Detail

  This Pull Request changes ActionView::Helpers::UrlHelper#link_to:

  - When the link's URL matches the current request (current_page?), link_to now emits aria-current="page" automatically.
  - A new :current option in html_options overrides the auto behavior:
    - current: false — suppress the attribute on the current page.
    - current: true — force aria-current="page".
    - current: "step" / current: :location — set a custom ARIA value (per WAI-ARIA: page, step, location, date, time).
  - An aria-current value supplied by the caller — either directly ("aria-current" => …) or nested (aria: { current: … }) — is never overwritten.
  - When no request is available (e.g. mailer views), auto-detection is silently skipped; link_to never raises.

  Implementation mirrors the existing convert_options_to_data_attributes pattern: a private convert_options_to_aria_attributes(url, html_options) helper that returns the updated hash, plus a small
  link_targets_current_page? guard that rescues no-request and URL-parsing failures.

  Example:
```
  <%# On /articles %>
  <nav>
    <%= link_to "Home", root_path %>
    <%= link_to "Articles", articles_path %>
  </nav>

  renders:

  <nav>
    <a href="/">Home</a>
    <a href="/articles" aria-current="page">Articles</a>
  </nav>

  nav a[aria-current="page"] { font-weight: bold; }
```

  ### Additional information

  - This is a behavior change: any link_to that points at the current page will now carry a new attribute. The change is opt-out (current: false) rather than opt-in to keep the API ergonomic; the accessibility improvement
  is a strict win for the common case. Five existing PolymorphicControllerTest cases were updated to reflect the new attribute — there were no other regressions in the Action View suite (1657 tests, 0 failures).
  - aria-current is a [WAI-ARIA 1.1 attribute](https://www.w3.org/TR/wai-aria-1.1/#aria-current); page is one of its defined values.
  - Related existing helper: link_to_unless_current. This change does not deprecate it, but for most navigation use cases the new behavior should be preferred since it keeps the link semantically present.

  ### Checklist

### DIFF
--- a/actionview/CHANGELOG.md
+++ b/actionview/CHANGELOG.md
@@ -1,3 +1,26 @@
+*   Automatically add `aria-current="page"` to `link_to` when the link's URL
+    matches the current request.
+
+    Beyond the accessibility win, this lets you style the active link in a
+    navigation directly from CSS — no `.active` class to toggle, no
+    `link_to_unless_current` wrapper:
+
+        # On /articles
+        <%= link_to "Articles", articles_path %>
+        # => <a href="/articles" aria-current="page">Articles</a>
+
+        /* stylesheet */
+        nav a[aria-current="page"] { font-weight: bold; }
+
+    The new `:current` option overrides the auto behavior — pass `false` to
+    suppress the attribute, `true` to force it, or a String/Symbol (e.g.
+    `"step"`) for a custom ARIA value:
+
+        link_to "Step 2", step_path(2), current: "step"
+        # => <a href="/steps/2" aria-current="step">Step 2</a>
+
+    *Jeremy Bertrand*
+
 *   Pass render options and block to calls to `#render_in`
 
     ```ruby
--- a/actionview/lib/action_view/helpers/url_helper.rb
+++ b/actionview/lib/action_view/helpers/url_helper.rb
@@ -92,6 +92,12 @@ module ActionView
       #
       # ==== Options
       # * <tt>:data</tt> - This option can be used to add custom data attributes.
+      # * <tt>:current</tt> - Controls the <tt>aria-current</tt> attribute. Pass
+      #   <tt>true</tt> to force <tt>aria-current="page"</tt>, <tt>false</tt> to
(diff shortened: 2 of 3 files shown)


p8:
Thanks for the PR @NotGrm 
I don’t think we use any aria attributes for any other tag helper in Rails. I guess it’s up to the developer to add them?

rafaelfranca:
Yeah, I don't like this is automatic. I think `link_to_unless_current` should be more than enough. Thank you for the pull request.

NotGrm:
Sorry I wasn't able to respond earlier 🙏 

First of all I want to thank you @p8 @rafaelfranca to have taken sometimes reviewing my proposal. It means a lot to me as it was my first time trying to contribute to Rails.

For the story, the idea of the PR came to me after reading #57293 and the importance of `aria-current` for screen-reader users and how replacing a link by simple text was not "efficient". So I thought that automatically adding the attribute was aligned with the doctrine.

Accessibility matters to me, so I will thought of a way to make this change available to others as a Rails extension 💪

PR #57352, opened 2026-05-12
Properly reset fixtures cache after non-transactional tests

byroot:
Extracted from: https://github.com/rails/rails/pull/57326

### DIFF
--- a/activerecord/lib/active_record/test_fixtures.rb
+++ b/activerecord/lib/active_record/test_fixtures.rb
@@ -140,16 +140,14 @@ module ActiveRecord
 
         if run_in_transaction?
           # Load fixtures once and begin transaction.
-          @loaded_fixtures = @@already_loaded_fixtures[@fixture_cache_key]
-          unless @loaded_fixtures
-            @@already_loaded_fixtures.clear
+          unless (@loaded_fixtures = @@already_loaded_fixtures[@fixture_cache_key])
+            invalidate_already_loaded_fixtures
             @loaded_fixtures = @@already_loaded_fixtures[@fixture_cache_key] = load_fixtures(config)
           end
 
           setup_transactional_fixtures
         else
           # Load fixtures for every test.
-          ActiveRecord::FixtureSet.reset_cache
           invalidate_already_loaded_fixtures
           @loaded_fixtures = load_fixtures(config)
         end
@@ -182,6 +180,7 @@ module ActiveRecord
       end
 
       def invalidate_already_loaded_fixtures
+        ActiveRecord::FixtureSet.reset_cache
         @@already_loaded_fixtures.clear
       end
 
--- a/activerecord/test/cases/fixtures_test.rb
+++ b/activerecord/test/cases/fixtures_test.rb
@@ -605,7 +605,9 @@ class FixturesTest < ActiveRecord::TestCase
     db_url_tmp = ENV["DATABASE_URL"]
     ENV["DATABASE_URL"] = "sqlite3::memory:"
     ActiveRecord::Base.stub(:configurations, {}) do
-      test_case = Class.new(ActiveRecord::TestCase) do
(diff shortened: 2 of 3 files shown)


activerecord/lib/active_record/test_fixtures.rb
```diff
@@ -140,16 +140,14 @@ def setup_fixtures(config = ActiveRecord::Base)
 
         if run_in_transaction?
           # Load fixtures once and begin transaction.
-          @loaded_fixtures = @@already_loaded_fixtures[@fixture_cache_key]
-          unless @loaded_fixtures
-            @@already_loaded_fixtures.clear
+          unless (@loaded_fixtures = @@already_loaded_fixtures[@fixture_cache_key])
```
rafaelfranca, line 143: Any reason for this change? Code is more confusing now. It seems like the instance variable is only set when the cache don't have an entry

byroot: You mean the `unless (var = ...)` ? No strong reason. I can revert that one part.

byroot: 0f1617eb24c1b1f83450b1179be281c060ef837d

Edgarruiz8585:
🚀

Édgar Manuel reaccionó a través de Gmail
<https://www.google.com/gmail/about/?utm_source=gmail-in-product&utm_medium=et&utm_campaign=emojireactionemail#app>

Edgarruiz8585:
>  by: Edgar Manuel Ruiz Arias. @Edgarruiz8585
***@***.*** ***@***.***
***@***.*** ***@***.***

rafaelsales:
@byroot @rafaelfranca this one made a project's test suite about 2x slower, so flagging it here.

Bumping our pin from 80e2db7 to ef79c57 took `bin/rails test` from 213s to 908s on the same 5892 tests. It bisects to this PR #57352 / commit https://github.com/rails/rails/commit/17cb852d72a501ce3eefd7f9d647bebf73a1f550.

The non-transactional branch is unchanged, but the transactional one used to call bare `@@already_loaded_fixtures.clear` and now goes through `invalidate_already_loaded_fixtures`, which brings `FixtureSet.reset_cache` along with it. So a cache key miss no longer just drops a memo, it re-reads and re-inserts every fixture table.

We miss constantly because `rails/test_help.rb` gives `ActionDispatch::IntegrationTest` the fixtures path twice: it inherits it from `ActiveSupport::TestCase`, then the load hook appends that same array again. Integration tests end up with a different `@fixture_cache_key` than model and job tests, so every alternation between the two reloads all 109 fixture sets. That is the second unchecked blocker in #57326, except it does not need randomized ordering to show up.

On a 490 test slice, same app commit and only the Rails ref changing: 1 fixture reload before this commit, 29 after, so `load_fixtures` went from 1.3s to 63.4s.

How much each of those 29 reloads costs depends on the Postgres version:

- **Postgres 16**: 0.53s per reload. `disable_referential_integrity` still issues one batched `DISABLE TRIGGER ALL`.
- **Postgres 18.4**: 2.19s per reload, which is what my CI runs. 7051ca8 and 2c392fc switch `disable_referential_integrity` and `check_all_foreign_keys_valid!` to the per constraint `NOT ENFORCED` path, so each reload issues ~378 `ALTER TABLE ... ALTER CONSTRAINT` statements for the schema's 189 foreign keys, plus revalidation.

So Postgres 18.4 is not the trigger (the 213s baseline ran on it too), it just makes each reload about 4x more expensive.

Working around it on our side with:

```ruby
ActiveSupport.on_load(:action_dispatch_integration_test) do
  self.fixture_paths = fixture_paths.uniq
end
```

byroot:
https://github.com/rails/rails/pull/58454

But yeah, I meant to re-do the fixture code entirely, it's stupidly wasteful, but I just can't find the time/energy to dig into it.

PR #57360, opened 2026-05-13
Fix renderable arity checks when method is overridden

chaadow:
Use `Kernel#method` explicitly when checking a renderable object's render_in arity, so objects that define their own method reader do not break rendering. Along with a regression test.

More context: https://github.com/rails/rails/pull/50623#discussion_r3230073756

cc @rafaelfranca / @seanpdoyle 

### Checklist

### DIFF
--- a/actionview/lib/action_view/template/renderable.rb
+++ b/actionview/lib/action_view/template/renderable.rb
@@ -14,7 +14,9 @@ module ActionView
       end
 
       def render(context, locals)
-        if @renderable.method(:render_in).arity == 1
+        render_in_method = Kernel.instance_method(:method).bind_call(@renderable, :render_in)
+
+        if render_in_method.arity == 1
           ActionView.deprecator.warn <<~WARN
             Action View support for #render_in without options is deprecated.
 
--- a/actionview/test/template/render_test.rb
+++ b/actionview/test/template/render_test.rb
@@ -429,6 +429,22 @@ module RenderTestCases
     end
   end
 
+  def test_render_renderable_object_with_method_reader
+    renderable = Class.new do
+      attr_reader :method
+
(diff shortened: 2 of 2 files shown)


actionview/lib/action_view/template/renderable.rb
```diff
@@ -14,7 +14,7 @@ def identifier
       end
 
       def render(context, locals)
-        if @renderable.method(:render_in).arity == 1
+        if render_in_method.arity == 1
```
seanpdoyle, line 19: Would it be acceptable to forego the private method and instead inline the call to a local variable?

```suggestion
        render_in_method = Kernel.instance_method(:method).bind_call(@renderable, :render_in)
        
        if render_in_method.arity == 1
```

It's a bit more verbose, but this guard clause wraps an `ActionView.deprecator` call, so it'd co-locate the logic for however long the temporary conditional remains in the codebase. Whoever removes the deprecation code doesn't risk forgetting about the private `render_in_method` method.

actionview/lib/action_view/template/renderable.rb
```diff
@@ -36,6 +36,11 @@ def render(context, locals)
       def format
         @renderable.try(:format)
       end
+
+      private
+        def render_in_method
+          Kernel.instance_method(:method).bind_call(@renderable, :render_in)
+        end
```
seanpdoyle:
```suggestion
```

actionview/lib/action_view/template/renderable.rb
chaadow: Sure!

chaadow: Done @seanpdoyle

PR #57361, opened 2026-05-13
Fix create_or_find_by duplicating WHERE conditions on retry within transaction

### LINKED ISSUE #57192: create_or_find_by applies WHERE conditions twice when retrying after RecordNotUnique in transaction
create_or_find_by applies WHERE conditions twice when retrying after RecordNotUnique 

### Steps to reproduce

1. Call  `where(field: "example_value")` on an ActiveRecord model. This will make CollectionProxy keep conditions in next call
2. Call `create_or_find_by` with a different `field: "another_value"`. Retry query will combine both conditions, building an impossible to meet conditions.

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activerecord", "8.1.3"
  gem "sqlite3"
end

require "active_record"
require "logger"

  adapter: "sqlite3",
  database: ":memory:"
)

  create_table :blog_posts, force: true do |t|
    t.string :title
    t.timestamps
  end

  create_table :slug_redirects, force: true do |t|
    t.string :from_slug, null: false
    t.references :redirectable, polymorphic: true, null: false
    t.timestamps
  end

  add_index :slug_redirects, [:from_slug, :redirectable_type, :redirectable_id],
            unique: true, name: "index_slug_redirects_uniqueness"
end

class BlogPost < ActiveRecord::Base
  has_many :slug_redirects, as: :redirectable, dependent: :destroy
end

class SlugRedirect < ActiveRecord::Base
  belongs_to :redirectable, polymorphic: true
end

puts "\n" + "=" * 70
puts "TEST: RecordNotFound with DIFFERENT values (polluted relation)"
puts "=" * 70

SlugRedirect.delete_all
blog_post = BlogPost.create!(title: "Test")

# Pre-create with "old-slug"
SlugRedirect.create!(from_slug: "old-slug", redirectable: blog_post)
puts "\nCreated redirect with from_slug: 'old-slug'"

# Now simulate what happens when a relation is pre-scoped with a different value
# This is like having a CollectionProxy that already has where(from_slug: 'new-slug')
new_slug = "new-slug"
old_slug = "old-slug"

puts "\nSimulating polluted relation scenario:"
puts "- Base relation has: where(from_slug: '#{new_slug}')"
puts "- Calling create_or_find_by(from_slug: '#{old_slug}')"
puts "- Retry query will have: from_slug='#{new_slug}' AND from_slug='#{old_slug}'"
puts "- This matches NOTHING → RecordNotFound\n\n"

begin
  BlogPost.transaction do
    # Create a "polluted" relation - this simulates what happens when
    # CollectionProxy retains previous where conditions
    polluted_relation = SlugRedirect.where(
      from_slug: new_slug,  # The "pollution" - a DIFFERENT value
      redirectable: blog_post
    )

    # Now call create_or_find_by with a DIFFERENT from_slug
    # The retry query will combine both conditions
    result = polluted_relation.create_or_find_by(from_slug: old_slug)
    puts "Result: #{result.inspect}"
  end
rescue ActiveRecord::RecordNotFound => e
  puts "RecordNotFound raised!"
  puts " #{e.message}"
rescue => e
  puts "Error: #{e.class} - #{e.message}"
end

puts "\nDone!"
```

### Expected behavior

When `create_or_find_by catches` `RecordNotUnique` inside a transaction, the retry query should have each condition once:

```sql
SELECT * FROM "slug_redirects" 
WHERE "from_slug" = $1 AND "redirectable_type" = $2 AND "redirectable_id" = $3 
LIMIT 1 FOR UPDATE
```

### Actual behavior

The retry query has duplicate WHERE conditions — each attribute appears twice:

```sql
SELECT * FROM "slug_redirects" 
WHERE "from_slug" = $1 AND "redirectable_type" = $2 AND "redirectable_id" = $3 
  AND "from_slug" = $4 AND "redirectable_type" = $5 AND "redirectable_id" = $6 
LIMIT 1 FOR UPDATE
```

### Impact

When this retry occurs on a scoped relation (e.g., a `CollectionProxy` or chained `where`), the duplicate conditions can have **different** values, causing `RecordNotFound`

```ruby
# Relation already scoped with from_slug: "new-slug"
scoped_relation = SlugRedirect.where(from_slug: "new-slug", redirectable: blog_post)

# Calling with different value
scoped_relation.create_or_find_by(from_slug: "old-slug")

# Retry query becomes:
# WHERE from_slug = 'new-slug' AND from_slug = 'old-slug' → matches nothing → RecordNotFound
```

### Root cause:

In [activerecord/lib/active_record/relation.rb](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L283), both `create_or_find_by` and `create_or_find_by!` have:

```
rescue ActiveRecord::RecordNotUnique
    where(attributes).lock.find_by!(attributes)  # <-- BUG: attributes applied twice
  else
    find_by!(attributes)  # <-- This branch is correct
  end
end
```

The where(attributes).find_by!(attributes) applies attributes twice.

### System configuration

**Rails version**: 8.1.3
**Ruby version**:  4.0.1

irvanmhndra:
When `create_or_find_by` rescues a `RecordNotUnique` inside a transaction, the previous `where(attributes).lock.find_by!(attributes)` applied the attributes twice — and if the relation already had conflicting scope conditions for the same key, the generated SQL would be contradictory (e.g. `WHERE nick = 'irvan' AND nick = 'bob'`), raising `RecordNotFound`.

Fix by using `rewhere(attributes).lock.take!` instead: `rewhere` replaces existing conditions for those attribute keys rather than stacking them, and `take!` fetches the record without adding further WHERE clauses.

Fixes #57192

### Motivation / Background

`create_or_find_by` raises `RecordNotFound` when called on a scoped relation
inside a transaction. When rescuing `RecordNotUnique`, the previous code used
`where(attributes).lock.find_by!(attributes)`, which applied the lookup
attributes twice on top of the existing scope. If the scope already had a
condition for the same key with a different value, the SQL became contradictory:

```sql
-- Subscriber.where(nick: "irvan").create_or_find_by(nick: "bob") inside a transaction
WHERE nick = 'irvan' AND nick = 'bob' AND nick = 'bob'
```

This always returns zero rows, raising `RecordNotFound` instead of the existing record.

### Detail

Replaced `where(attributes).lock.find_by!(attributes)` with
`rewhere(attributes).lock.take!` in the transactional rescue branch of both
`create_or_find_by` and `create_or_find_by!`.

`rewhere` removes any existing scope conditions for the specified attribute
keys before applying the new ones, preventing conflicting duplicates.
`take!` then fetches the first matching record without appending further
WHERE clauses.

### Checklist

* [x] Commit message has a detailed description of what changed and why. Fixes #57192
* [x] CHANGELOG updated in activerecord.


rafaelfranca:
Duplicate of #57223

PR #57365, opened 2026-05-14
Support dynamic `store:` option in `rate_limit`

caws:
## Summary

Also support dynamic `store:` option in `rate_limit` like [we already do for some of the other params.](https://github.com/rails/rails/blob/18f0be3dda62235ea71ab8627322815d66dd6367/actionpack/lib/action_controller/metal/rate_limiting.rb#L91)

Proposed in https://discuss.rubyonrails.org/t/proposal-support-dynamic-store-option-in-rate-limit/91088.

## Motivation / Background

The `rate_limit` API supports method names and callables for some of the parameters. 
Extending this to `:store` allows applications to resolve the cache store at runtime (per-request, per-tenant, gradual cache migrations, etc.) without boilerplate conditionals in controllers. Static usages untouched.

## Detail

Changes `ActionController::RateLimiting` so that `store:` accepts:

- Static values (unchanged behavior), e.g. `store: RedisCacheStore.new(...)`.
- Instance method names (symbols), evaluated in the controller context via `send` (like we already do for [to](https://github.com/rails/rails/blob/18f0be3dda62235ea71ab8627322815d66dd6367/actionpack/lib/action_controller/metal/rate_limiting.rb#L92) and [within](https://github.com/rails/rails/blob/18f0be3dda62235ea71ab8627322815d66dd6367/actionpack/lib/action_controller/metal/rate_limiting.rb#L93)).
- Callables (lambdas/procs), evaluated in the controller context via `instance_exec` following the existing pattern.

Resolution logic:

- If value is a Symbol → `send(value)`
- Else if it responds to `call` → `instance_exec(&value)`
- Else → use as-is

## Additional information

- Aligns with earlier changes to allow method names for `:to`, `:within`, `:by`, and `:with`.
- Negligible overhead (a couple of conditional checks) on the rate-limited path.

## Checklist

### DIFF
--- a/actionpack/test/controller/rate_limiting_test.rb
+++ b/actionpack/test/controller/rate_limiting_test.rb
@@ -86,7 +86,7 @@ class RateLimitedSharedFourController < RateLimitedSharedController
   end
 end
 
-class RateLimitedWithDynamicStoreController < ActionController::Base
+class RateLimitedWithStoreFromMethodController < ActionController::Base
   self.cache_store = ActiveSupport::Cache::MemoryStore.new
   rate_limit to: 2, within: 2.seconds, store: :store_for_request, only: :limited
 
@@ -102,7 +102,7 @@ class RateLimitedWithDynamicStoreController < ActionController::Base
     end
 end
 
-class RateLimitedWithCallableStoreController < ActionController::Base
+class RateLimitedWithStoreFromCallableController < ActionController::Base
   self.cache_store = ActiveSupport::Cache::MemoryStore.new
   rate_limit to: 2,
              within: 2.seconds,
@@ -304,8 +304,8 @@ class RateLimitingTest < ActionController::TestCase
   end
 
   test "dynamic store via method" do
-    @controller = RateLimitedWithDynamicStoreController.new
-    RateLimitedWithDynamicStoreController.cache_store.clear
+    @controller = RateLimitedWithStoreFromMethodController.new
+    RateLimitedWithStoreFromMethodController.cache_store.clear
 
     get :limited
     assert_response :ok
@@ -324,8 +324,8 @@ class RateLimitingTest < ActionController::TestCase
   end
 
   test "dynamic store via callable" do
-    @controller = RateLimitedWithCallableStoreController.new
-    RateLimitedWithCallableStoreController.cache_store.clear
+    @controller = RateLimitedWithStoreFromCallableController.new
+    RateLimitedWithStoreFromCallableController.cache_store.clear
 
     get :limited
     assert_response :ok
--- a/actionpack/test/controller/rate_limiting_test.rb
+++ b/actionpack/test/controller/rate_limiting_test.rb
@@ -95,12 +95,11 @@ class RateLimitedWithDynamicStoreController < ActionController::Base
   end
 
   private
-
-  def store_for_request
-    params[:use_alternate_store] ?
-      ActiveSupport::Cache::MemoryStore.new :
-      self.class.cache_store
-  end
+    def store_for_request
+      params[:use_alternate_store] ?
+        ActiveSupport::Cache::MemoryStore.new :
+        self.class.cache_store
+    end
 end
 
 class RateLimitedWithCallableStoreController < ActionController::Base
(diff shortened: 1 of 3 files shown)


actionpack/test/controller/rate_limiting_test.rb
```diff
@@ -86,6 +86,38 @@ def limited_shared_four
   end
 end
 
+class RateLimitedWithDynamicStoreController < ActionController::Base
```
joshuay03: Nit. I don't think `DynamicStore` is descriptive enough. Callables give us a dynamic store too, maybe something like this for consistency?


```suggestion
class RateLimitedWithStoreFromMethodController < ActionController::Base
```

actionpack/test/controller/rate_limiting_test.rb
```diff
@@ -86,6 +86,38 @@ def limited_shared_four
...
+
+  private
+    def store_for_request
+      params[:use_alternate_store] ?
+        ActiveSupport::Cache::MemoryStore.new :
+        self.class.cache_store
+    end
+end
+
+class RateLimitedWithCallableStoreController < ActionController::Base
```
joshuay03: To keep things consistent with the proposed naming above:


```suggestion
class RateLimitedWithStoreFromCallableController < ActionController::Base
```

actionpack/test/controller/rate_limiting_test.rb
caws: Sounds good to me. Addressed the comment.

actionpack/test/controller/rate_limiting_test.rb
caws: Sounds good to me. Addressed the comment.

joshuay03 (approved):

matthewd:
I'm having a bit of trouble articulating the use case to myself here.

I can see changing the effective limit for premium users, and/or bucketing differently (premium users by individual id; others as a single "everyone-else").

For a data store migration, I feel like you'd more likely want a custom store class that provides fall-through reads or similar.

And for user-based sharding... that again feels more suited to a custom store setup: I don't think we generally provide this sort of callable-indirection for _other_ store uses?

---

It's a tiny code change, and is especially well-precedented on this method where we already callable-resolve nearly every other option... so I'm not _against_ doing this. But I'd still prefer to have a reasonably clear story of why it's useful, beyond the open-ended idea that runtime-dynamic behaviour gives more flexibility: that's abstractly true of every option to every class-level macro, but applying it as a universal design rule would make many APIs impossibly contortionate. 😕

caws:
> I'm having a bit of trouble articulating the use case to myself here.
> 
> I can see changing the effective limit for premium users, and/or bucketing differently (premium users by individual id; others as a single "everyone-else").
> 
> For a data store migration, I feel like you'd more likely want a custom store class that provides fall-through reads or similar.
> 
> And for user-based sharding... that again feels more suited to a custom store setup: I don't think we generally provide this sort of callable-indirection for _other_ store uses?
> 
> It's a tiny code change, and is especially well-precedented on this method where we already callable-resolve nearly every other option... so I'm not _against_ doing this. But I'd still prefer to have a reasonably clear story of why it's useful, beyond the open-ended idea that runtime-dynamic behaviour gives more flexibility: that's abstractly true of every option to every class-level macro, but applying it as a universal design rule would make many APIs impossibly contortionate. 😕

That is a very fair point, @matthewd 🤔

I don't think there is a unique case that cannot be implemented via a custom store wrapper. Most of the examples I had in mind can indeed be modeled that way, as you pointed out, but could be expressed more directly with this change.

I also don't think Rails has many examples of dynamically selecting infrastructure objects specifically, so I wouldn't claim there's a broader precedent there.

My rationale was more local to the `rate_limit` API itself: it already evaluates most of its configuration in request context (`to:`, `within:`, `by:`, `with:`), and `store:` seemed to be the one option that couldn't participate in that pattern.

Allowing `store:` to be resolved in the same way as the other options felt like a natural extension of the existing API (at a low cost), and a more direct expression of the intent when the backing store itself depends on request context.

joshuay03:
Circling back on this one. @matthewd, do you still have reservations after @caws's reply, or are you happy to let it through? No push from me either way, happy to leave it if you're not convinced.

PR #57367, opened 2026-05-14
Prevent development welcome route from duplicating on route reload

curi:
### Motivation / Background

Rails registers its internal development welcome route from `routes_reloader.run_after_load_paths`.

That hook currently calls `app.routes.append` on each route reload. `RouteSet#append` stores append blocks persistently, while route reloads clear and finalize the realized routes. As a result, repeated `reload_routes!` calls in development add another internal `GET /` welcome route each time.

### Detail

Register the internal welcome route append block only once per application boot, while preserving the existing `append` behavior and route ordering. This keeps app-appended root routes taking precedence over the internal welcome route.

### Additional information

No existing issue found for this specific duplicate welcome-route behavior.

### Tests

- `ruby -c railties/lib/rails/application/finisher.rb`
- `ruby -c railties/test/application/routing_test.rb`
- `bundle exec rubocop railties/lib/rails/application/finisher.rb railties/test/application/routing_test.rb`
- `cd railties && bin/test test/application/routing_test.rb -i '/not duplicated|appended root/'`
- `cd railties && bin/test test/application/routing_test.rb`

### DIFF
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Prevent the internal development welcome route from being duplicated on route reloads.
+
+    *Elliot Temple*
+
 *   Remove `new_framework_defaults` during `app:update` when `config.load_defaults`
     already targets the current Rails version.
 
--- a/railties/lib/rails/application/finisher.rb
+++ b/railties/lib/rails/application/finisher.rb
@@ -147,9 +147,13 @@ module Rails
             get ".well-known/appspecific/com.chrome.devtools.json" => "rails/devtools#show",      internal: true
           end
 
+          welcome_route_appended = false
           routes_reloader.run_after_load_paths = -> do
-            app.routes.append do
-              get "/" => "rails/welcome#index", internal: true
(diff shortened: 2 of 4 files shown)


curi:
The buildkite failure is a connection leak in activerecord postgresql (4.0). I think that test is worth rerunning because my patch only affects Railties routing and the test passed on my Mac.

rafaelfranca (changes requested):
I think the right way to fix this is by clearing append blocks when routes are reloaded.

curi:
> I think the right way to fix this is by clearing append blocks when routes are reloaded.

Clearing all append blocks when reloading routes causes two test failures (ApplicationTests::RoutingTest "routes appending blocks" and "routes appending blocks after reload"). I'm unsure if the existing behavior is important to preserve or those tests should be changed. Another option is adding an append_transiently method so append stays persistent, then adding the welcome route with append_transiently.

rafaelfranca (approved):
ok. I found a different way to solve the problem and pushed to your branch. Thank you for the pull request

PR #57368, opened 2026-05-14
Limit the size of strings we call `to_i` on in ActiveRecord

tenderlove:
Calling `to_i` on very long strings can take a very long time and could be used as a DoS vector.  Users should validate input lengths before calling `to_i`, but we can do it somewhat automatically in ActiveModel.

This commit limits auto-integer coercion to the first `_limit * 4` bytes of a string, where `_limit` is the column's storage size in bytes.  For a default 4-byte integer that is 16 bytes; for an 8-byte bigint it is 32 bytes.  That is comfortably larger than the maximum number of digits any value in range can have (10 digits for a 4-byte int, 19 for an 8-byte bigint), with room left over for a sign or a short slug suffix.

For example, if we have a model like this:

```ruby
ActiveRecord::Base.connection.create_table :users do |t|
  t.integer :age, limit: 1
  t.timestamps
end

class User < ActiveRecord::Base; end

User.create(age: '9' * 5_000_000)
```

Before this commit, casting `age` would walk the entire 5 MB string. After this commit, only the first `_limit * 4` bytes are considered, bounding the work `to_i` has to do.

At first, I based this change off the width of the column from the database, but that doesn't work because ActiveRecord supports "auto integerization" of slugs.  For example you might have a controller that looks like this:

```ruby
Post.where(id: params[:id]).first
```

We expect slugs to come in with values like "123-hello-world-etc", so limiting strictly to the column width would break that.  Multiplying by 4 leaves room for typical slug suffixes while still bounding how much work `to_i` does.

### DIFF
--- a/activemodel/lib/active_model/type/integer.rb
+++ b/activemodel/lib/active_model/type/integer.rb
@@ -110,7 +110,15 @@ module ActiveModel
         end
 
         def cast_value(value)
-          value.to_i rescue nil
+          case value
+          when ::Integer
+            value
+          when ::String
+            str = value.bytesize > _limit * 4 ? value.byteslice(0, _limit * 4) : value
+            str.to_i rescue nil
+          else
+            value.to_i rescue nil
+          end
         end
 
         def max_value
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Limit the size of strings `ActiveModel::Type::Integer` will coerce with `to_i`.
+
+    Calling `to_i` on very long strings can take a long time and could be used as
+    a DoS vector. Integer casting now only considers the first `_limit * 4` bytes
+    of a string (16 bytes for a default 4-byte integer, 32 bytes for an 8-byte
+    bigint), which is enough to hold the maximum representable value plus a sign
+    or a short slug suffix.
+
+    *Aaron Patterson*, *Jean Boussier*
+
 *   Support proc and symbol for `NumericalityValidator`s `:in` option
 
     ```ruby
--- a/activemodel/test/cases/type/integer_test.rb
+++ b/activemodel/test/cases/type/integer_test.rb
@@ -165,6 +165,20 @@ module ActiveModel
         end
       end
 
+      test "long strings are truncated before being cast" do
+        type = Type::Integer.new # default limit: 4, byte cap: 16
+        # Only the first _limit * 4 == 16 bytes should be considered
+        assert_equal ("1" * 16).to_i, type.cast("1" * 17)
+      end
+
+      test "strings within the byte cap are cast normally" do
+        type = Type::Integer.new # default limit: 4, byte cap: 16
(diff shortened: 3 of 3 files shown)


rafaelfranca (approved):
Makes sense to me. I'd just write tests and add a changelog entry so people know to expect large strings will become nil

byroot (approved):

activemodel/lib/active_model/type/integer.rb
```diff
@@ -110,7 +110,15 @@ def out_of_range?(value)
         end
 
         def cast_value(value)
-          value.to_i rescue nil
+          case value
+          when ::Integer
+            value
+          when ::String
+            return nil if value.bytesize > 1024
```
byroot: What I had in mind the other day was to use `@limit * 2` (or `* 4` or something), but also to parse the result of that, e.g.:

```ruby
str = value.bytesize > @limit * 2 ? value.byteslice(0, @limit * 2) : value
str.to_i rescue nil
```

byroot: Also unrelated, but perhaps it'd be better to use `Integer(str, 10, exception: false)` and get rid of the blanket rescue.

tenderlove:
> Makes sense to me. I'd just write tests and add a changelog entry so people know to expect large strings will become nil

Yep, just wanted to make sure this seems like an OK solution before doing that.  I'll do it before we land this 😄

activemodel/lib/active_model/type/integer.rb
tenderlove: > What I had in mind the other day was to use `@limit * 2` (or `* 4` or something), but also to parse the result of that, e.g.:

Actually I like this much better.  I'll implement it.

byroot: > perhaps it'd be better to use `Integer(str, 10, exception: false)` and get rid of the blanket rescue.

Seems like we can't because:

```ruby
>> "1ignore".to_i
=> 1
>> Integer("1ignore", 10, exception: false)
=> nil
```

Too bad.

byroot:
I implemented the suggested changes. As for the changelog, I suppose we'll backport this?

If that's the case then I prefer to only add the changelog in the backport commits.

activemodel/lib/active_model/type/integer.rb
tenderlove: > Too bad.

Still, doing the limit + byteslice is better IMO.

tenderlove:
> I implemented the suggested changes. As for the changelog, I suppose we'll backport this?
> 
> If that's the case then I prefer to only add the changelog in the backport commits.

Ugh, I'm really sorry.  I didn't see this and updated the PR.  @byroot if you have the branch around can you force push again?

> I implemented the suggested changes. As for the changelog, I suppose we'll backport this?

I think we should

byroot:
> can you force push again?

No need your pach is just as good as mine, I just used `*2` instead of `*4` that's basically all.

PR #57370, opened 2026-05-15
Prevent false leaked-connection reports during reaper maintenance

curi:
### Motivation / Background

`ActiveRecord::TestCase#check_connection_leaks` scans each pool during teardown to catch connections left owned by background threads.

`ConnectionPool::Reaper` can legitimately hold a connection under `reaper_lock` while it performs maintenance. The leak check currently inspects `pool.connections` without taking that lock, so teardown can race the reaper, report a leaked connection owned by `AR Pool Reaper`, and kill the maintenance thread even though no test leaked anything.

### Detail

Synchronize the leak scan with `pool.reaper_lock` so teardown waits for any in-flight maintenance to finish before classifying in-use connections as leaks.

Add a regression test that reproduces the false positive by holding a connection for maintenance from a thread named `AR Pool Reaper`. The new test fails against `origin/main` and passes with this patch.

### Additional information

This surfaced as an unrelated CI failure while validating #57367, where the reported leaked owner was the pool reaper thread.

### Tests

- `cd activerecord && ARCONN=postgresql bundle exec ruby -Itest test/cases/test_case_test.rb -v`
- `cd activerecord && ARCONN=postgresql bundle exec ruby -Itest test/cases/reaper_test.rb`
- `cd activerecord && ARCONN=postgresql bundle exec ruby -Itest test/cases/associations/has_and_belongs_to_many_associations_test.rb`
- `bundle exec rubocop activerecord/test/cases/test_case.rb activerecord/test/cases/test_case_test.rb`

### DIFF
--- a/activerecord/test/cases/test_case.rb
+++ b/activerecord/test/cases/test_case.rb
@@ -60,15 +60,18 @@ module ActiveRecord
           end
         end
 
-        pool.reap
-        pool.connections.each do |conn|
-          if conn.in_use?
-            if conn.owner != Fiber.current && conn.owner != Thread.current
-              leaked_conn << [conn.owner, conn.owner.backtrace]
-              conn.owner&.kill
+        # Avoid racing the pool reaper while it is performing maintenance.
+        pool.reaper_lock do
+          pool.reap
+          pool.connections.each do |conn|
+            if conn.in_use?
+              if conn.owner != Fiber.current && conn.owner != Thread.current
+                leaked_conn << [conn.owner, conn.owner.backtrace]
+                conn.owner&.kill
+              end
+              conn.steal!
+              pool.checkin(conn)
             end
-            conn.steal!
-            pool.checkin(conn)
           end
         end
       end
--- a/activerecord/test/cases/test_case_test.rb
+++ b/activerecord/test/cases/test_case_test.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require "cases/helper"
+
+module ActiveRecord
+  class TestCaseLeakCheckTest < ActiveRecord::TestCase
+    include WaitForTestHelper
+
+    self.use_transactional_tests = false
+
+    def test_check_connection_leaks_does_not_report_connections_held_for_reaper_maintenance
+      pool = ActiveRecord::Base.connection_pool
+
+      conn = ActiveRecord::Base.lease_connection
+      conn.select_value("SELECT 1")
+      ActiveRecord::Base.connection_handler.clear_active_connections!
+
+      assert_not_nil conn
+
+      release_maintenance_latch = Concurrent::CountDownLatch.new(1)
+      leak_check_attempted_latch = Concurrent::CountDownLatch.new(1)
+
+      maintenance_thread = Thread.new do
+        Thread.current.name = "AR Pool Reaper"
+
+        pool.reaper_lock do
+          pool.send(:checkout_for_maintenance, conn)
+          release_maintenance_latch.wait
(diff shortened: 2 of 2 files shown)


activerecord/test/cases/test_case_test.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      maintenance_started = Concurrent::Event.new
+
+      maintenance_thread = Thread.new do
+        Thread.current.name = "AR Pool Reaper"
+
+        pool.reaper_lock do
+          pool.send(:checkout_for_maintenance, conn)
+          maintenance_started.set
+          sleep 0.2
```
rafaelfranca: Do we need to sleep here? Can't we also use an event of countdown latch?

curi: Thanks. I removed the sleep using latches and verified the test still fails on the unpatched code.

activerecord/test/cases/test_case.rb
Edgarruiz8585, line 1:
```suggestion
# frozen_string_literal: true @Edgarruuz8585
```

headius:
I'm getting confusing errors after this merge running the activerecord tests on JRuby.

Nearly all tests now fail with
```
NoMethodError: undefined method 'reaper_lock' for an instance of ActiveRecord::ConnectionAdapters::ConnectionPool
```

The only place I see it defined is via the stub in the test added for this PR, but the global test_case.rb was modified to use it. Am I missing something? Why is this method not defined globally somewhere?

byroot:
It's defined there: https://github.com/rails/rails/blob/3469de6d1cb4e2e1a329f23ff66b4e60518d55ee/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L576-L578

headius:
It does not appear to be defined on 8-0-STABLE even though the `test_case.rb` uses it.

* test_case.rb: https://github.com/rails/rails/blob/8-0-stable/activerecord/test/cases/test_case.rb#L63
* connection_pool.rb with no reaper_lock: https://github.com/rails/rails/blob/8-0-stable/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L498-L522

Perhaps this was merged to 8-0-STABLE without appropriate support code.

headius:
FYI I am testing JRuby support for 8.0 on the way to 8.1+ so we can get our adapter support released for all 8.x versions.

byroot:
Right, you should have mentioned 8-0 from the start, I assumed you meant `main`.

Yes, that commit couldn't be backported this far I reverted it.

headius:
@byroot Sorry about that, I didn't realize it was relevant until you pointed out where it was defined, and didn't expect a breaking change to get merged to 8-0-STABLE. Thanks for looking into it.

headius:
@byroot The revert breaks the test suite due to a remaining include of `WaitForTestHelper` here:

https://github.com/rails/rails/blob/8-0-stable/activerecord/test/cases/test_case_test.rb#L7

headius:
Side question: are verification CI tests not run against 8-0-STABLE anymore? I would have expected this breakage to be caught by Rails CI.

byroot:
Yes it does: https://buildkite.com/rails/rails/builds?branch=8-0-stable, but the backporting process is to cherry-pick what is worth backporting directly in the `-stable` branch without going though a PR, we stabilize the stable branches CI asynchronously.

headius:
@byroot Is there a branch I can use that I know will always be green or should I just test against most recent tag going forward?

byroot:
> I just test against most recent tag going forward?

tags should be green yes.

PR #57371, opened 2026-05-15
[RF-Docs] [ci-skip] Update the Rails Generators guide

bhumi1102:
This PR is part of the Rails Foundation documentation project and updates the Rails Generators guide.

Note: Section "Applications Templates" and "Rails Generator API" do not require a review. They have already gone through the full review process already as they came from merging another guide into this guide.


guides/source/generators.md
```diff
@@ -446,6 +488,46 @@ $ bin/rails generate scaffold Comment body:text
...
+
+When resolving generator template files, Rails looks first in the application's
+`lib/templates/` directory, then falls back to the generator's own `source_root`
+directory. This means we can override the templates used by Rails' built-in
+generators by placing our own versions in `lib/templates/`. For example, we
+could override the [scaffold controller template][] or the [scaffold view
+templates][].
+
+To see this in action, let's create a
+`lib/templates/erb/scaffold/index.html.erb.tt` file with the content shown below. The extra `.tt` extension tells Rails that the file is a generator template needs to be processed by Thor first (The `.tt` stands for "thor template").
```
p8: Ah, I never known `.tt` stands for "thor template". 🤔 

```suggestion
`lib/templates/erb/scaffold/index.html.erb.tt` file with the content shown below. The extra `.tt` extension tells Rails that the file is a generator template that needs to be processed by Thor first (The `.tt` stands for "thor template").
```

guides/source/generators.md
p8, line 300: Maybe we can use the scope here when copying the file?
For example as a sub dir:
```suggestion
  scope = options["scope"]
  copy_file "initializer.rb", "config/initializers/#{scope}/#{file_name}.rb"
```

p8 (commented):
Look great @bhumi1102 !

guides/source/generators.md
```diff
@@ -3,99 +3,167 @@
 Creating and Customizing Rails Generators & Templates
```
p8: This title is one of the longest.
It also spans two lines in the menu navigation.
What do you think about shortening it to:
```suggestion
Rails Generators & Templates
```

guides/source/generators.md
bhumi1102: Same here, was curious so looked it up :)

guides/source/generators.md
bhumi1102: Good idea, updated!

PR #57377, opened 2026-05-16
Add `enforced:` option for foreign keys on PostgreSQL 18.4+

yahonda:
This pull request adds `enforced:` option for foreign keys on PostgreSQL 18.4+

### Motivation / Background

On PostgreSQL 17 or earlier, bypassing foreign key checks requires `DISABLE TRIGGER ALL`, which needs superuser privileges. Re-enabling triggers does not re-validate existing rows — so invalid data can silently slip in.

PostgreSQL 18 introduces the NOT ENFORCED option for foreign keys, which fixes both: the table owner can toggle it (no superuser needed), and switching back to ENFORCED re-validates all rows. A remaining blocker — that toggling lost the DEFERRABLE / INITIALLY DEFERRED attributes Rails relies on — was fixed in PostgreSQL 18.4, so Active Record can finally support this.

A follow-up pull request will `update disable_referential_integrity` to use NOT ENFORCED on PostgreSQL 18.4+.

### Detail

- `NOT ENFORCED` foreign keys are available since PostgreSQL 18.0, but `DEFERRABLE` was lost on them until 18.4 ("Fix loss of deferrability of foreign-key triggers", https://www.postgresql.org/docs/release/18.4/). Rails therefore requires PostgreSQL 18.4 or later for this feature.

- When `enforced: false` is passed to `add_foreign_key`, the constraint is created as `NOT ENFORCED`, meaning PostgreSQL skips referential integrity checks during DML. PostgreSQL marks `NOT ENFORCED` constraints as `NOT VALID` internally (and `VALIDATE CONSTRAINT` does not apply to them), so the schema dumper outputs both `enforced: false` and `validate: false`.

- `change_foreign_key` toggles the enforced status of an existing foreign key. Without this method, users would need to issue raw `ALTER TABLE ... ALTER CONSTRAINT` SQL to disable enforcement temporarily (e.g., for bulk DML that loads the referenced and referencing tables in arbitrary order) and then re-enable it.

### Additional information

- [PostgreSQL 18.0 Release Notes](https://www.postgresql.org/docs/release/18.0/)
> Allow [CHECK](https://www.postgresql.org/docs/release/18.0/sql-createtable.html#SQL-CREATETABLE-PARMS-CHECK) and [foreign key](https://www.postgresql.org/docs/release/18.0/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES) constraints to be specified as NOT ENFORCED (Amul Sul) [§](https://postgr.es/c/ca87c415e) [§](https://postgr.es/c/eec0040c4)

- [PostgreSQL 18.4 Release Notes](https://www.postgresql.org/docs/release/18.4/)
> Fix loss of deferrability of foreign-key triggers (Yasuo Honda) [§](https://postgr.es/c/5db5e3396)

- A previous attempt to avoid the superuser requirement via SET CONSTRAINTS ALL DEFERRED was [reverted](https://github.com/rails/rails/commit/2812694aa65e4d0591e0ea6c5f7d8c97927be7dc) because referential actions other than NO ACTION cannot be deferred, even when the constraint is declared DEFERRABLE.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,34 @@
+*   Add `enforced:` option to `add_foreign_key` and `change_foreign_key` for PostgreSQL 18.4+.
+
+    `NOT ENFORCED` foreign keys are available since PostgreSQL 18.0, but `DEFERRABLE`
+    was lost on them until 18.4 ("Fix loss of deferrability of foreign-key triggers",
+    https://www.postgresql.org/docs/release/18.4/). Rails therefore requires PostgreSQL
+    18.4 or later for this feature.
+
+    When `enforced: false` is passed to `add_foreign_key`, the constraint is created as `NOT ENFORCED`,
+    meaning PostgreSQL skips referential integrity checks during DML.
+    PostgreSQL marks `NOT ENFORCED` constraints as `NOT VALID` internally (and `VALIDATE CONSTRAINT`
+    does not apply to them), so the schema dumper outputs both
+    `enforced: false` and `validate: false`.
+
+    `change_foreign_key` toggles the enforced status of an existing foreign key. Without
+    this method, users would need to issue raw `ALTER TABLE ... ALTER CONSTRAINT` SQL
+    to disable enforcement temporarily (e.g., for bulk DML that loads the referenced
+    and referencing tables in arbitrary order) and then re-enable it.
+
+    ```ruby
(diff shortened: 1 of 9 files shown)


yahonda:
> A follow-up pull request will update disable_referential_integrity to use NOT ENFORCED on PostgreSQL 18.4+.

https://github.com/rails/rails/pull/57378 has been opened as a draft one.

Issue #57224, opened 2026-04-22
`guard_load_hooks` does not allow engines to extend application models


flavorjones:
### Summary

Related to #56201, which introduced `guard_load_hooks` (cc @gmcgibbon).

On Rails 8.2.dev, an engine that extends a host-app model with `config.to_prepare` emits `guard_load_hooks` warnings on boot (when not eager-loading). This is a long-standing pattern (concrete example: the [audits1984](https://github.com/basecamp/audits1984) engine extends the application's auditor class with `has_one :auditor_token`) we'd expect to be allowed.

This is essentially what the gem does:

```ruby
class MyEngine::Engine < ::Rails::Engine
  config.to_prepare do
    MyEngine.auditor_class.constantize.has_one :auditor_token,
      class_name: "MyEngine::AuditorToken", foreign_key: :auditor_id, dependent: :delete
  end
end
```

Five warnings fire when you boot a host that uses this engine without eager loading (`bin/rails assets:precompile`, `runner`, test suite) — one per guarded component that `AR::Base` transitively loads:

```
:active_model_translation was loaded before application initialization.
:active_record_encryption was loaded before application initialization.
:active_model_secure_password was loaded before application initialization.
:active_record was loaded before application initialization.
:active_record_sqlite3adapter was loaded before application initialization.
```

### Steps to reproduce

Self-contained script (approximates the engine pattern by referencing `ActiveRecord::Base` directly inside `config.to_prepare`):

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", github: "rails/rails", branch: "main"
  gem "sqlite3"
end

require "rails"
require "active_record/railtie"
require "logger"

ENV["DATABASE_URL"] = "sqlite3::memory:"

# A Rails engine that extends a host-app model via `config.to_prepare`.
# A real engine would constantize a configurable host class and install a
# `has_one` on it; we approximate by just referencing ActiveRecord::Base,
# which autoloads it.
class MyEngine < Rails::Engine
  config.to_prepare do
    ActiveRecord::Base
  end
end

# Minimal Rails application.
class TestApp < Rails::Application
  config.root = __dir__
  config.eager_load = false
  config.secret_key_base = "x" * 30
  config.logger = Logger.new($stdout)

  # Make early load-hook firings fatal so this script terminates with a
  # clear exception rather than a log line. Comment out to see all warnings.
  config.action_on_early_load_hook = :raise
end

[...]

PR #57379, opened 2026-05-17
Module load hooks

gmcgibbon:
### Motivation / Background

This Pull Request has been created because Rails is missing a way to insert arbitrary load hooks for specific autoloadable classes. Zeitwerk has support for this with `on_load("MyClassName")`, but Rails doesn't provide a public-api-friendly way of accessing the Zeitwerk loader in Rails (the only way is `Rails.autoloaders.main`, which is private API). Even if you were to use this API, the semantics here are slightly different to `ActiveSupport.on_load`. 

### Detail

This Pull Request adds `ActiveSupport.on_module_load` to fill this gap. Instead of using `to_prepare` to eagerly load a class:

```rb
Rails.configuration.to_prepare do
  MyClassName.config = true
end
```

We can now instead do something like this:

```rb
ActiveSupport.on_module_load(:MyClassName) do
  self.config = true
end
```

### Additional information

More information here: https://github.com/rails/rails/issues/57224

### Checklist


activesupport/lib/active_support/module_load_hooks.rb
```diff
@@ -0,0 +1,43 @@
...
+  # outside of \Rails, then ActiveSupport.autoloader must be set manually.
+  module ModuleLoadHooks
+    self.attr_accessor :autoloader
+
+    # Declares a block that will be executed when an autoloadable module is fully
+    # loaded. If the module has already loaded, the block is executed
+    # immediately.
+    def on_module_load(name, &block)
+      raise NotImplementedError, "#{self}.autoloader must be set." unless autoloader
+      autoloader.on_load(name.to_s) do |mod|
```
flavorjones: I believe this needs to remove any leading `::` part of the string, which Zeitwerk does not recognize and so will not call the `on_load` callback.

Specific example: "Identity" would work, but "::Identity" will not.

In my original comment in https://github.com/rails/rails/issues/57224#issuecomment-4299722522 I used this approach to do this:

```ruby
      cpath = class_name.delete_prefix("::")
```

Maybe this is behavior belongs in Zeitwerk?

flavorjones: Or maybe this PR just needs to document that the root scope operator is not supported.

gmcgibbon: I'll open a patch upstream in Zeitwerk, and hopefully the silent failure can be addressed there somehow. If not, we can apply it here.

gmcgibbon: https://github.com/fxn/zeitwerk/pull/335

rafaelfranca (commented):
Why is this API in active support and not in the Rails::Application class? Is there any use case that requires it to be used outside of a Rails application?

Adding the concept of an autoloader to Active Support is a smell that this doesn't belong here.

gmcgibbon:
I placed it on Active Support because I wanted module load hooks to look similar to component ones. It might be confusing to some developers to have to look in other places to set module load hooks. I'll push another commit with the method hanging off `Rails::Appliction`, and I'll squash if we like it better.

fxn:
Hey, let me add some context here.

I saw clearly at the beginning that chasing parity with Zeitwerk's API would be a lost game. Rails provides autoload paths, eager load paths, ..., not a lot more.

Zeitwerk additionally provides:

* Custom logging (does have a bridge interface)
* Custom inflections
* Ignoring parts of the project
* Eager loading directories
* Eager loading namespaces
* Different kinds of callbacks
* Collapsing directories
* Custom root namespaces
* ....

It is deliberate that Rails does not try to play catchup, it stays at what you had in Rails 5 in terms of what is configurable, basically. Anything else, grab the loader.

And that is why `Rails.autoloaders` is **public API**. If you look at the [guide ](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html), there is [a section saying so](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#rails-autoloaders), and several examples addressing different needs in the guide explicitly using these accessors.

fxn:
The comment above is about the general mindset: The fact that an API does not have a counterpart in Rails is not by itself a justification, that is deliberate, a conscious design choice. The difference in API surface is substantial, and loaders can be accessed precisely so that users have it all at their disposal (a moving target).

But if there was another good reason to do it, I would not recommend the name `on_module_load`, because autoloading is about _constants_, not classes or modules. The type of the value hold by the constant is arbitrary.

gmcgibbon:
> It is deliberate that Rails does not try to play catchup

Understood. I'm fine with changing the name of the API to `on_constant_load` or something along those lines, but

> Rails.autoloaders is public API

I think we may not all agree on this. The [`Rails::Autoloaders`](https://github.com/rails/rails/blob/bf0dec9546476548d5dc3b01684f0008ea85b4e3/railties/lib/rails/autoloaders.rb#L4) class is private API, but [`Rails.autoloaders`](https://api.rubyonrails.org/classes/Rails.html#method-c-autoloaders) is in the API docs, but undocumented. Mentioning it in the guide is pretty clearly stating we meant for it to be public, or at lest we condone using it in applications. @rafaelfranca let me know if you have opinions here.

fxn:
> I think we may not all agree on this. 

That is not up to discussion, I am the one who wrote all this and I know the interface is public.

I suppose I wanted to hide `Rails::Autoloaders` because I wanted `Rails.autoloaders` to be the public interface, not the exact class implementing that API. But with that, the methods are not in the API docs.

What needs to be fixed is the API docs.

fxn:
> I'm fine with changing the name of the API to on_constant_load or something along those lines

I think this is not a good idea. Why special-case this callback? What about `on_unload`? What about `ignore`? What about `collapse`? What about future APIs?

That is why these interfaces exist in Rails via `Rails.autoloaders` (again, as explained in the guide).

rafaelfranca:
> What needs to be fixed is the API docs.

Correct, if all those methods are inside a class with nodoc, unfortunately, at this point, they are all private API, the guide is just documenting private API.

What I mean is that `Rails.autoloaders` is pubic, `Rails.autoloaders.main` isn't at the moment.

Let's fix the API doc then.

rafaelfranca:
Answering `Why special-case this callback?`, mostly because we are defining an use case in Rails that is to allow constant to be loaded only when needed to extend behavior. For this new use case we really don't care about the other callbacks.

The API being defined by this PR is:

How do you extend an existing constant in the application code in libraries without loading that application code on boot?

Pretty much a API to solve the problem that exists on:

```ruby
Rails.configuration.to_prepare do
  MyClassName.config = true
end
```

We want something that would say:

```ruby
execute_this_code_when_this_constant_is_loaded_so_my_rails_plugin_can_extend_its_behavior("MyClassName") do
  MyClassName.config = true
end
```

fxn:
Yep, I think this is mostly vocabulary. To me a "public interface" is a public contract, and the API is not the only one saying what is the public contract. We have two conflicting docs in this regard, and what I am trying to say is that `Rails.autoloaders.main` is actually public, so the premise that originates this PR does not hold.

Regarding the use case, that is literally a use case [documented](https://github.com/fxn/zeitwerk#the-on_load-callback) in the docs of Zeitwerk (only not saying "Rails", because those are agnostic). And being able to do that in Rails was the request from @byroot that motivated the creation of the `on_load` callback. Wonder if Shopify has some occurrence in `config/environments` or something, that was the motivation.

So, that use case is currently supported, and the interface is

```ruby
Rails.autoloaders.main.on_load(...) { ... }
```

just like the rest of APIs that also have use cases in Rails, you `collapse` for certain STI patterns, you `eager_load_dir` for them too, you `push_dir` to define custom root namespaces, ..., there are many use cases in Rails covered by this flexible API.

You cannot special-case one of these without breaking this symmetry, really.

fxn:
Additionally, you may need to do that for a constant managed by the `once` autoloader, you'd miss a symmetry also in that sense.

fxn:
> only not saying "Rails", because those are agnostic

Just realized it does say "Rails" in this case, and there is environment config, probably because of the use case shared by @byroot at the time.

rafaelfranca:
> the API is not the only one saying what is the public contract

Actually I think that the API is the only one that defines the public contract, at least in the way we have being operating for the past 15 years.

If something is nodoc we can change without depreciation, exactly because the contract isn't public.

Of course, if something was documented in a guide and nodoc in the api, we can consider that a API bug and fix the API. 

I'd prefer if we don't extend the definition of what is public contract of the framework to different documents and instead just correct the intention of making those methods part of the public contract in the API. 

We can still make sure people use the intended API and still document in the methods and modules marking them as public API of the framework by the definition we were being using for so long.

fxn:
OK, let's say it this way: This PR is like "since Rails.autoloaders.main is private interface, let's add this method".

My feedback is that interface should be in the API documentation, it is a mistake on my side it does not show up. In practice, the presence of the interface in official documentation is my way of showing that was my intention since the start. And people reading the guide have been using this since Rails 6.

If you think about it, what is the point of making `Rails.autoloaders` be listed in the API docs, if you cannot do anything with it?

I would have documented `main` and friends probably in the docs of the method `Rails.autoloaders`, hiding the type of the returned object, leaving the nodoc in place. Because `Rails::Autoloaders` is not something the user should know about. That is something we could change anytime, `Rails.autoloaders.*` is the public contract you cannot change. I believe that was my intention but never did it.

fxn:
And, the commit message would not be "make this private interface public", it would be "add missing docs to the API".

Because users following the multiple examples in the guide were not abusing private interface (a capital sin), they were using public interface documented in official documentation. It is a bug in the API docs, not a change in the intention of what could be used.

See what I mean?

fxn:
I have added documentation to `Rails.autoloaders` in https://github.com/rails/rails/pull/57435, and have cherry-picked to `8-1-stable` (this is something I normally do with doc enhancements).

And have added internal docs to `Rails::Autoloaders` in https://github.com/rails/rails/commit/399cc4734a4ba906e21bd54ed144dd735f28ae8a.

byroot:
For the record: https://github.com/fxn/zeitwerk/issues/142

We did discuss the possibility of a proper Rails integration at the time, first to not have to consider which autoloader is responsible for that constant, and second for handling the footgun that the callback isn't called if the constant was already loaded.

No strong opinion, but I kinda still feel that Zeitwerk's `on_load` is a very powerful feature, but not quite easy enough to use for most Rails users.

fxn:
> to not have to consider which autoloader is responsible for that constant

Well, in principle you should not define `on_load` for a class an autoloader does not manage. It could even be an error condition. It is not because at config time I don't know (a directory could be pushed later), and in part because I don't want to traverse the project tree on `setup` to validate, for performance. So, the documented contract for now is: If you make a mistake, it is up to you.

Users do need to understand what is reloaded and what is not in their applications.

fxn:
Another mismatch is that not all application code is managed by autoloaders. What if MyGateway is in `lib`?

Why I should not get called? The use case is the same regardless, no?

I see fundamental cons to this idea implemented as a mere proxy to the autoloaders:

1. Goes against the intentional symmetry of the design.
2. Defines callbacks in autoloaders that do not manage the constant. While technically it does not raise, I don't think it is right, and makes autoloaders process blocks on load that will never fire.
3. Does not cover application code as such, only a subset. An important one, but a partial feature. That partiality is already covered and its scope clear by reading the code today.

byroot:
> Well, in principle you should not define `on_load` for a class an autoloader does not manage. I

Yes, and I agree it's a perfectly fine decision for Zeitwerk given its scope, but that's also why a higher level Rails aware equivalent may make sense.

The Zeitwerk API is too error prone for most users IMO. e.g. consider cases like:

  - `MyLib` is handled by zeitwerk
  - There is a `Rails.autoloaders.main.on_load('MyLib')` in an initializer.
  - Later on for whatever reason someone move that constant to now be eager required.
  - The `on_load` now silently fail.

That is why I believe there is a case for a higher level, more use friendly and error proof API at the Rails level.

Something that can check if the constant already exist, in which case it just run the callback immediately, but also perhaps in integrated with Rails `eager_load = true` and will raise an error or emit a warning for callbacks defined on non-existing constants, etc.

NB: I haven't had the time to look at this specific PR implementation, I'm not saying this one should be merged or whatever, just sharing a use case I had when working on Shopify's monolith.

fxn:
@byroot That would be a different PR description and a different patch, I would certainly be open to explore that in Rails. Such `on_load` would fire even for `"String"`.

fxn:
There is a gotcha though in your use case.

If you later load the file with `require` in an initializer, `on_load` is not fired, true. That file is no longer managed by the autoloader. Let's assume the test suite does not catch that, possible.

But before, you were guaranteed reloadable constants were available in the block, and now you don't have that guarantee and your code breaks. This is all very tricky!

I am open to making such enhanced wrapper, but what is possible, the use cases and implications have to be carefully considered.

byroot:
> That would be a different PR description and a different patch,

Yes definitely, I don't see much value in just delegating to Zeitwerk.

I'd rather look into a more "bullet proof" / easier to handle API that tries to account for corner cases (already loaded constant, constant not managed by zeitwerk, constant not existing at all).

gmcgibbon:
With https://github.com/rails/rails/pull/57435, it looks like this is solved, so I'll close this. 

Though, regarding 

> a more "bullet proof" / easier to handle API that tries to account for corner cases

Perhaps we can add additional validations in the Zeitwerk check?

https://github.com/rails/rails/blob/84d3c32d026f0807e866907b93be2a3c5786b794/railties/lib/rails/zeitwerk_checker.rb#L4-L9

I can open a separate patch for that if there's interest.

Thanks for the discussion folks!

PR #51840, opened 2024-05-15
Update link to `i18n` gem repository [ci-skip]

p8:
The repository got moved to the `ruby-i18n` organization.

PR #57381, opened 2026-05-17
[RF Docs] Rails Internationalization

Ridhwana:
### Motivation / Background

This Pull Request has been created to improve the Rails Internationalization docs.

### Detail

A big change is that I've reordered the guide to start off with beginner friendly content, then move into managing locales in requests (as its own section for easier navigation), then API features and finally an advaced section.

CleanShot 2026-03-31 at 21 10 47@2x

Other parts tackled from [here](https://github.com/rails/rails/pull/51840#discussion_r1603318416):

- [x] Remove the Contributing section which references a inactive forum.
- [ ] Remove the authors section, as we don't mention the authors in other guides.
- [x] Unlike other guides the introduction does a quick summation of steps to take to implement I18n. The whole introduction should probably be rewritten with a "After reading this guide, you will know..."
- [x] "This means, that in the :en locale, the key hello will map to the Hello world string. " This could probably be replaced with a code example:
 I18n.t('hello')
 => "Hello world"
- [x] Code examples would also improve the mentioned ActiveModel validation messages.
- [ ] "Managing the Locale across Requests" should probably be a top section.
- [x] - The first example of setting the locale through a param is missing a header. Something like:  "Setting the Locale from a request parameter"
- [x] - In the "Passing Variables to Translations" section the product_price switches from `$` to `€` which is probably not typical. Maybe a better example would be using "dutch" and "spanish" where dutch uses "€ 100" and spanish "100 €".
- [x] - The note about `number_to_currency` should probably be a NOTE instead.
- [x] - The line "(Future Rails versions may well bring this automagic localization to assets in public, etc.)" can probably be removed, as I don't think this has been implemented or is planned.
- [ ] 
- [x] - The returned values for the examples in "Basic Lookup, Scopes, and Nested Keys" would make things clearer.
- [x] - The resulting HTML for "app/views/home/index.html.erb " in "Using Safe HTML Translations" is missing. This would help in seeing what gets escaped and what doesn't.
- [x] - The example usage in "Translations for Active Record Models" like `User.model_name.human` are probably clearer in a code block.
- [x] - The methods mentioned in "Overview of Other Built-In Methods that Provide I18n Support" would be clearer with code examples.
- [x] - The "Custom Translations" examples are more clear if we also show the output:
- [x] I18n.t "date.formats.short" # => "%b %d"
- [ ] 
- [x] - The "Using Different Backends" section shows an example how to add a different backend, but doesn't go into detail on how to use that backend.
- [x] - The "Using Different Exception Handlers" also mentions how to configure Rails to handle MissingTranslationData exceptions. That section should probably be renamed "Handling Missing Translations" or "Handling I18n Exceptions". This section could probably use an example as well:
- [x] > I18n.t('missing')
- [x] Translation missing: en.missing (I18n::MissingTranslationData)
- [ ] and maybe examples can be added for the other exceptions as well?
- [x] - The "Translating Model Content" section should not link to 3rd party gems. This section should probably just be a note in the introduction instead.
- [ ] 

### Checklist


guides/source/i18n.md
```diff
@@ -3,350 +3,453 @@
 Rails Internationalization (I18n) API
```
Ridhwana: @p8 I'm thinking of renaming this to Rails Internationalization (I18n) or Internationalization (I18n), what do you think?

p8: Yes, I think that would be an improvement. Everything else also has an API, so it seems redundant.
"Internationalization (I18n)" would have my slight preference.

Also maybe we can move this guide a bit further down in the "Digging Deeper" menu?
Maybe move it under "Working with Javascript"? Readers would probably introduce JavaScript before thinking about I18n?

Ridhwana: Done ✅ Thanks @p8

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
 * How to use I18n to translate Active Record errors or Action Mailer E-mail subjects
-* Some other tools to go further with the translation process of your application
+* How to customize backends and handle translation-related exceptions
 
 --------------------------------------------------------------------------------
+Internationalization and Localization
+-------------------------------------
 
-NOTE: The Ruby I18n framework provides you with all necessary means for internationalization/localization of your Rails application. You may also use various gems available to add additional functionality or features. See the [rails-i18n gem](https://github.com/svenfuchs/rails-i18n) for more information.
+Internationalization (`I18n`) is the process of preparing your application to
```
rafaelfranca:
```suggestion
Internationalization (I18n) is the process of preparing your application to
```

We aren't talking about a constant so it should not be using monospacing font. There are a lot of places where this is happening in this guide

Ridhwana: I'll fix this @rafaelfranca, thanks for pointing it out

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
-* Tell Rails how to set, preserve, and switch locales.
-
-In the process of _localizing_ your application you'll probably want to do the following three things:
-
-* Replace or supplement Rails' default locale - e.g. date and time formats, month names, Active Record model names, etc.
-* Abstract strings in your application into keyed dictionaries - e.g. flash messages, static text in your views, etc.
-* Store the resulting dictionaries somewhere.
-
-This guide will walk you through the I18n API and contains a tutorial on how to internationalize a Rails application from the start.
+The Ruby `i18n` gem, which ships with Ruby on Rails, provides an easy-to-use and extensible framework for translating your application into a single custom language or for providing multi-language support.
```
MatheusRich: Could we link the gem?
```suggestion
The Ruby [`i18n` gem](https://github.com/ruby-i18n/i18n), which ships with Ruby on Rails, provides an easy-to-use and extensible framework for translating your application into a single custom language or for providing multi-language support.
```

MatheusRich: Nevermind, I see we link below. Maybe this is unecessary. Or maybe we should move the link here?

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
 
-```ruby
-# config/initializers/locale.rb
+### Organization of Locale Files
 
-# Where the I18n library should search for translation files
-I18n.load_path += Dir[Rails.root.join("lib", "locale", "*.{rb,yml}")]
+When using the default `Simple` backend shipped with the i18n library,
+dictionaries are stored in plain-text files on disk. Putting translations for
+all parts of your application in one file per locale can become hard to manage,
```
MatheusRich, line 142: TIL you could do that. I've never seen this in the wild. Should we put more emphasis on doing this when things get too big, or that this is not a prescriptive hierarchy. I liked how the previous text mention that users should pick a hierarchy that makes sense for them.

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
+  def index
+    flash[:notice] = t("hello_flash")
+  end
+end
+```
 
-One option you have is to set the locale from the domain name where your application runs. For example, we want `www.example.com` to load the English (or default) locale, and `www.example.es` to load the Spanish locale. Thus the _top-level domain name_ is used for locale setting. This has several advantages:
+```html+erb
+<!-- app/views/home/index.html.erb -->
+<h1><%= t "hello_world" %></h1>
```
MatheusRich: Any particular reason to skip parenthesis here? I noticed we used it in the controller and a few other view samples.

```suggestion
<h1><%= t("hello_world") %></h1>
```

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
-# in your /etc/hosts file to try this out locally
-def extract_locale_from_tld
-  parsed_locale = request.host.split(".").last
-  I18n.available_locales.map(&:to_s).include?(parsed_locale) ? parsed_locale : nil
-end
+```yaml
+# config/locales/es.yml
+es:
+  hello_world: Hola mundo!
+  hello_flash: Hola flash!
```
MatheusRich: Not a Spanish speaker, but pretty sure we need ¡ on the translated versions
```suggestion
  hello_world: ¡Hola mundo!
  hello_flash: ¡Hola flash!
```

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
 
-Rails contains infrastructure for "centralizing dynamic decisions about the URLs" in its [`ApplicationController#default_url_options`](https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Base.html#method-i-default_url_options), which is useful precisely in this scenario: it enables us to set "defaults" for [`url_for`](https://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for) and helper methods dependent on it (by implementing/overriding `default_url_options`).
+```yaml
+# config/locales/es.yml
+es:
+  currency: "€ "
+```
 
-We can include something like this in our `ApplicationController` then:
+If the product's price is 100 then Dutch would want `"€ 100"` while Spanish
```
MatheusRich: Not to be pedantic, but this isn't true for _all_ Spanish. Just from Spain, I think.

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
 
-You would probably need to map URLs like these:
+Rails also provides [`relative_time_in_words`](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-relative_time_in_words) for a localized
+past/future phrase rather than a formatted timestamp:
 
-```ruby
-# config/routes.rb
-get "/:locale" => "dashboard#index"
+```erb
+<p><%= relative_time_in_words(3.minutes.from_now) %></p>
```
MatheusRich, line 381: Should we put examples of what that renders? Even if as a comment

guides/source/i18n.md
```diff
@@ -1,352 +1,455 @@
...
-  def extract_locale_from_accept_language_header
-    request.env["HTTP_ACCEPT_LANGUAGE"].scan(/^[a-z]{2}/).first
-  end
+You can add custom pluralization rules for Spanish, like below:
+
+```ruby
+# config/initializers/inflections.rb
+ActiveSupport::Inflector.inflections(:es) do |inflect|
+  inflect.irregular "el pais", "los paises"
+end
```
MatheusRich: I wonder if it is confusing to add inflections to two-word strings. I can't give an example in Spanish, so here's a single-word one in Portuguese
```suggestion
ActiveSupport::Inflector.inflections(:pt_br) do |inflect|
  inflect.irregular "aluguel", "aluguéis"
end
```

guides/source/i18n.md
```diff
@@ -355,306 +458,340 @@ class ApplicationController < ActionController::Base
...
 
-![rails i18n demo translated to English](images/i18n/demo_translated_en.png)
+NOTE: Controller-based locale extraction works well for simple applications. For
+production applications that derive the locale from host or path information,
+Rack middleware can extract the locale before routing and store it in the Rack
+environment. See [Rails on Rack](rails_on_rack.html) for more about Rack
+middleware in Rails applications.
 
-If the locale is set via the URL to the pirate locale (`http://localhost:3000?locale=pirate`), the response renders the pirate strings:
+### Setting the Locale from URL Params
```
MatheusRich, line 528: I wonder if this should be a subsection the te Request section, or at least show up right after it, since they're tightly linked.

guides/source/i18n.md
```diff
@@ -355,306 +458,340 @@ class ApplicationController < ActionController::Base
...
+# app/controllers/concerns/locale.rb
+around_action :switch_locale
 
-```erb
-<!-- app/views/home/index.html.erb -->
-<h1><%= t :hello_world %></h1>
-<p><%= flash[:notice] %></p>
-<p><%= l Time.now, format: :short %></p>
+def switch_locale(&action)
+  locale = current_user&.locale
```
MatheusRich: Nit: should we use CurrentAttributes here? 
```suggestion
  locale = Current.user&.locale
```

guides/source/i18n.md
Ridhwana: Thanks, I changed this to "might" so that we're not speaking in absolutes.

Ridhwana:
@MatheusRich and @rafaelfranca thank you so much for the review, I've addressed all your feedback.

guides/source/i18n.md
```diff
@@ -1,352 +1,465 @@
...
+* provide translations for application text
+* add or customize locale-specific formats, such as dates, times, and currencies
+* organize and maintain translation data
 
-### The Overall Architecture of the Library
+How I18n in Ruby on Rails Works
+-------------------------------
 
-Thus, the Ruby I18n gem is split into two parts:
+Rails ships with the Ruby i18n gem, which provides the underlying internationalization framework. In this guide, the I18n API refers to the interface your Rails application uses to work with that framework, such as `I18n.t`, `I18n.l`, and Rails translation helpers. As part of this system, static strings in Rails itself, such as Active Record validation messages and time and date formats, are already internationalized.
```
p8: If the link is removed from the first sentence, the link would be added here.
```suggestion
Rails ships with the Ruby [i18n gem](https://github.com/ruby-i18n/i18n), which provides the underlying internationalization framework. In this guide, the I18n API refers to the interface your Rails application uses to work with that framework, such as `I18n.t`, `I18n.l`, and Rails translation helpers. As part of this system, static strings in Rails itself, such as Active Record validation messages and time and date formats, are already internationalized.
```

guides/source/i18n.md
```diff
@@ -1,352 +1,465 @@
...
 
-`I18n.locale` can leak into subsequent requests served by the same thread/process if it is not consistently set in every controller. For example executing `I18n.locale = :es` in one POST requests will have effects for all later requests to controllers that don't set the locale, but only in that particular thread/process. For that reason, instead of `I18n.locale =` you can use `I18n.with_locale` which does not have this leak issue.
+Now, when this view is rendered, it will show an error message indicating that
+the translations for the keys `"hello_world"` and `"hello_flash"` are missing.
 
-The locale can be set in an `around_action` in the `ApplicationController`:
+```
+en, hello_world
+translation missing: en, hello_flash
+```
```
p8, line 235: I found this somewhat confusing.
The default en.yml contains `hello: "Hello world"` which is very similar to `hello_world: Hello world!`.
I had to double check `hello_world` was missing.
Also I would expect to see a translation missing for hello_world as well (the warning is present in the HTML though).
Maybe use different examples and show the rendered HTML instead? This would also show where the translation missing warnings can be seen:

```suggestion
Now, when this view is rendered, it will show an error message indicating that
the translations for the keys `"home_title"` and `"home_flash"` are missing.

```html+erb
<h1><span class="translation_missing" title="translation missing: en.home_title">Home Title</span></h1>
<p>Translation missing: en.home_flash</p>
```
```

File: guides/source/i18n.md [hunk 14]
```diff
@@ -1,352 +1,465 @@
...
+translations from other locale files. Rails does not switch locales from a query
+parameter automatically; your application must set the locale for the request,
+for example with an `around_action` as shown in
+[Managing the Locale across Requests](#managing-the-locale-across-requests).
 
-* The locale is an _obvious_ part of the URL.
-* People intuitively grasp in which language the content will be displayed.
-* It is very trivial to implement in Rails.
-* Search engines seem to like that content in different languages lives at different, inter-linked domains.
+If the locale is set via the URL to the Spanish locale (`http://localhost:3000?locale=es`), the response renders the Spanish strings:
```
p8: Another option would be to skip the complexity of around filters for now and change the default locale config instead.
````suggestion

If the default locale is set to the Spanish locale

```ruby
# config/application.rb
config.i18n.default_locale = :es
```

 the response renders the Spanish strings:
````

File: guides/source/i18n.md [hunk 15]
```diff
@@ -355,306 +468,340 @@ class ApplicationController < ActionController::Base
...
+next.
+
+For example, you may want URLs such as `www.example.com/books?locale=ja` or
+`www.example.com/ja/books`.
 
-![rails i18n demo untranslated](images/i18n/demo_untranslated.png)
+This approach has many of the same advantages as using the domain name for the
+locale; the locale is visible in the URL, the URL can be shared, and search
+engines can distinguish between language-specific pages. It does require a
+little more setup, though.
```
p8: For the "Locale from Domain Name" section we show the advantages as a list.
I think it would be nice to show the advantages here as a list as well, for easier comparison.
Also this should probably mention it has the same advantages as "using a parameter" (the domain name hasn't been mentioned yet)?
```suggestion
This approach has many of the same advantages as using a parameter for the
locale:
* the locale is visible in the URL
* the URL can be shared
* search engines can distinguish between language-specific pages
 
It does require a little more setup, though.
```

guides/source/i18n.md
```diff
@@ -355,306 +468,340 @@ class ApplicationController < ActionController::Base
...
-* localizing dates, numbers, currency, etc.
 
 ### Looking up Translations
 
 #### Basic Lookup, Scopes, and Nested Keys
 
-Translations are looked up by keys which can be both Symbols or Strings, so these calls are equivalent:
+Translations are looked up by keys. String keys are a good default because they
+work naturally with full keys such as `"books.index.title"` and relative keys
+such as `".title"`:
```
p8: The term "relative keys" isn't used anywhere else in the guide ("lazy lookup" is the term used later on).
It could be confusing to use a term that is explained later on.
Also, I think "books.index.title" is a good example by itself, ".title" doesn't add that much.

Maybe remove  the mention of relative keys here? 
```suggestion
work naturally with keys such as `"books.index.title"`:
```

guides/source/i18n.md
```diff
@@ -355,306 +468,340 @@ class ApplicationController < ActionController::Base
...
 
 ```ruby
-I18n.t :message
 I18n.t "message"
```
p8: Maybe use the "books.index.title" example here as well?
```suggestion
I18n.t "books.index.title"
```

guides/source/i18n.md
```diff
@@ -355,306 +468,340 @@ class ApplicationController < ActionController::Base
...
-I18n.t :missing, default: "Not here"
+I18n.t "missing", default: "Not here"
 # => 'Not here'
 ```
 
-If the `:default` value is a Symbol, it will be used as a key and translated. One can provide multiple values as default. The first one that results in a value will be returned.
+If the `:default` value is a string key, it will be used as a key and translated. One can provide multiple values as default. The first one that results in a value will be returned.
 
-E.g., the following first tries to translate the key `:missing` and then the key `:also_missing.` As both do not yield a result, the string "Not here" will be returned:
+The example below first tries to translate the key `"missing"` and then the key `"also_missing"`. Since both do not yield a result, the string "Not here" will be returned:
```
p8: Nitpick. To me it would be more logical to group the single and array values as separate paragraphs.
```suggestion
If the `:default` value is a string key, it will be used as a key and translated.

One can also provide multiple values as default. The first one that results in a value will be returned.
The example below first tries to translate the key `"missing"` and then the key `"also_missing"`. Since both do not yield a result, the string "Not here" will be returned:
```

guides/source/i18n.md
```diff
@@ -734,29 +881,33 @@ end
 
 ### Pluralization
 
-In many languages — including English — there are only two forms, a singular and a plural, for
-a given string, e.g. "1 message" and "2 messages". Other languages ([Arabic](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ar), [Japanese](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ja), [Russian](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#ru) and many more) have different grammars that have additional or fewer [plural forms](http://cldr.unicode.org/index/cldr-spec/plural-rules). Thus, the I18n API provides a flexible pluralization feature.
+In many languages,including English, there are only two forms, a singular and a plural, for
```
p8:
```suggestion
In many languages, including English, there are only two forms, a singular and a plural, for
```

guides/source/i18n.md
```diff
@@ -770,15 +921,16 @@ If the lookup for the key does not return a Hash suitable for pluralization, an
...
-The I18n gem provides a Pluralization backend that can be used to enable locale-specific rules. Include it
-to the Simple backend, then add the localized pluralization algorithms to translation store, as `i18n.plural.rule`.
+The I18n gem provides a `Pluralization` backend that can be used to enable
+locale-specific rules. Include it to the `Simple` backend, then add the
+localized pluralization algorithms to translation store, as `i18n.plural.rule`.
 
 ```ruby
 I18n::Backend::Simple.include(I18n::Backend::Pluralization)
 I18n.backend.store_translations :pt, i18n: { plural: { rule: lambda { |n| [0, 1].include?(n) ? :one : :other } } }
 I18n.backend.store_translations :pt, apples: { one: "one or none", other: "more than one" }
```
p8: Using some linebreaks here makes sure the example doesn't require scrolling, similar to the inbox example above.
```suggestion
I18n.backend.store_translations :pt, i18n: {
  plural: { rule: lambda { |n| [0, 1].include?(n) ? :one : :other } } 
}
I18n.backend.store_translations :pt, apples: {
  one: "one or none",
  other: "more than one"
}
```

guides/source/i18n.md
```diff
@@ -788,22 +940,24 @@ Alternatively, the separate gem [rails-i18n](https://github.com/svenfuchs/rails-
...
 
+WARNING: The i18n gem stores `I18n.locale` using `Thread.current`. Applications serving multiple requests from the same thread should set the locale explicitly around each request, for example with `I18n.with_locale`.
+
 If no locale is passed, `I18n.locale` is used:
 
 ```ruby
 I18n.locale = :de
-I18n.t :foo
+I18n.t "foo"
 I18n.l Time.now
```
p8: Maybe show what key is used for lookup here?
```suggestion
I18n.t "foo"    # Looks up de.foo
I18n.l Time.now # Looks up de.time.formats.default
```

guides/source/i18n.md
```diff
@@ -788,22 +940,24 @@ Alternatively, the separate gem [rails-i18n](https://github.com/svenfuchs/rails-
...
 ```
 
 Explicitly passing a locale:
 
 ```ruby
-I18n.t :foo, locale: :de
+I18n.t "foo", locale: :de
 I18n.l Time.now, locale: :de
```
p8: Maybe use different locales here to make the examples clearer?
```suggestion
I18n.t "foo", locale: :br    # Looks up br.foo
I18n.l Time.now, locale: :es # Looks up es.time.formats.default
```

guides/source/i18n.md
```diff
@@ -830,6 +984,15 @@ en:
 <div><%= t('title.html') %></div>
```
p8: The red HTML tags really stand out:
Image

What do you think about removing the extra divs here to make it clearer what HTML tags are rendered?
This would also mean we should probably delete the screenshot.
```suggestion
<%= t('welcome') %>
<%= raw t('welcome') %>
<%= t('hello_html') %>
<%= t('title.html') %>
```

guides/source/i18n.md
```diff
@@ -830,6 +984,15 @@ en:
...
 ```
 
+This renders:
+
+```html
+<div>&lt;b&gt;welcome!&lt;/b&gt;</div>
+<div><b>welcome!</b></div>
+<div><b>hello!</b></div>
+<div><b>title!</b></div>
```
p8:
```suggestion
&lt;b&gt;welcome!&lt;/b&gt;
<b>welcome!</b>
<b>hello!</b>
<b>title!</b>
```

guides/source/i18n.md
```diff
@@ -1039,36 +1220,230 @@ en:
...
+    ```ruby
+    distance_of_time_in_words(Time.current, Time.current + 3.minutes)
+    # => "3 minutes"
+    ```
+
+    It uses translations from the `datetime.distance_in_words` scope.
+
+* [`relative_time_in_words`](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-relative_time_in_words)
+  builds on `distance_of_time_in_words` and adds a localized past/future
+  wrapper.
```
p8: I found the API documentation description a bit more clear.
Maybe:
```suggestion
  builds on `distance_of_time_in_words` and adds a localized prefix/suffix
  depending on whether the time is in the past or future.
```

guides/source/i18n.md
```diff
@@ -1039,36 +1220,230 @@ en:
...
+
+    ```ruby
+    relative_time_in_words(3.minutes.from_now)
+    # => "in 3 minutes"
```
p8, line 1402: Might as well add the past case here as well:
```suggestion
    # => "in 3 minutes"
    relative_time_in_words(3.minutes.ago)
    # => "3 minutes ago"
```

guides/source/i18n.md
```diff
@@ -1106,51 +1481,86 @@ en:
...
+That does not mean you're stuck with these limitations, though. The Ruby I18n gem makes it very easy to exchange the `Simple` backend implementation with something else that fits better for your needs, by passing a backend instance to the `I18n.backend=` setter.
 
-For example, you can replace the Simple backend with the Chain backend to chain multiple backends together. This is useful when you want to use standard translations with a Simple backend but store custom application translations in a database or other backends.
+For example, you can replace the `Simple` backend with the `Chain` backend to chain multiple backends together. This is useful when you want to use standard translations with a `Simple` backend but store custom application translations in a database or other backends.
 
-With the Chain backend, you could use the Active Record backend and fall back to the (default) Simple backend:
+With the `Chain` backend, you could use the Active Record backend and fall back to the (default) `Simple` backend:
 
 ```ruby
 I18n.backend = I18n::Backend::Chain.new(I18n::Backend::ActiveRecord.new, I18n.backend)
```
p8: This example currently requires scrolling:

Image

```suggestion
I18n.backend = I18n::Backend::Chain.new(
  I18n::Backend::ActiveRecord.new,
  I18n.backend
)
```

guides/source/i18n.md
```diff
@@ -1106,51 +1481,86 @@ en:
...
 ```
 
-### Using Different Exception Handlers
+In this configuration, lookups first check the Active Record backend and then fall back to the locale files loaded by the `Simple` backend:
+
+```ruby
+I18n.t("welcome")
+# Looks for "welcome" in the Active Record backend first, then in YAML/Ruby locale files.
```
p8: This one isn't fully readable because of the copy button:

Image

```suggestion
# Looks for "welcome" in the Active Record backend first,
# then in YAML/Ruby locale files.
```

guides/source/i18n.md
```diff
@@ -1106,51 +1481,86 @@ en:
...
+| `I18n::UnknownFileType`              | the backend does not know how to handle a file type that was added to `I18n.load_path`            |
+
+#### Handling Missing Translations
+
+If `config.i18n.raise_on_missing_translations` is `true`, missing translations
+raise errors instead of returning the usual `"Translation missing: ..."`
+message. If the value is `:strict`, model translations also raise. It is a good
+idea to enable this in your test environment so you can catch missing
+translations early.
+
```
p8, line 1710: Maybe add a raising example here as well?
````suggestion

For example:

```ruby
I18n.t("missing")
=> Translation missing: en.missing (I18n::MissingTranslationData)
```

````

File: guides/source/i18n.md [hunk 30]
```diff
@@ -1106,51 +1481,86 @@ en:
...
-| `I18n::UnknownFileType`              | the backend does not know how to handle a file type that was added to `I18n.load_path` |
+| `I18n::UnknownFileType`              | the backend does not know how to handle a file type that was added to `I18n.load_path`            |
+
+#### Handling Missing Translations
+
+If `config.i18n.raise_on_missing_translations` is `true`, missing translations
+raise errors instead of returning the usual `"Translation missing: ..."`
+message. If the value is `:strict`, model translations also raise. It is a good
+idea to enable this in your test environment so you can catch missing
+translations early.
```
p8, line 1709: Maybe swap this with the "default `config.i18n.raise_on_missing_translations` is `false`" paragraph?
So we first show the Rails default, and then we expand with the example of setting it to `true`.

guides/source/i18n.md
p8: It feels like all the "Translations for *" sections deserve their own top section.
These are less about the I18n API, and more about build-in Rails functionality.
Optionally flatten the "Other I18n Helpers" section, to make them visible in the sidebar as well.

What do you think about something like:
- Overview of the I18n API Features
  - Looking up Translations
  - Pluralization
  - Setting and Passing a Locale
  - Using Safe HTML Translations
- Build-in Rails Translations (or maybe "Rails Translations")
  - Active Record Models
  - Action Mailer E-Mail Subjects
  - Form Helpers
  - Action View Helper Methods
  - Active Model Methods
  - Active Support Methods

guides/source/i18n.md
```diff
@@ -1,352 +1,465 @@
...
-* Tell Rails how to set, preserve, and switch locales.
-
-In the process of _localizing_ your application you'll probably want to do the following three things:
-
-* Replace or supplement Rails' default locale - e.g. date and time formats, month names, Active Record model names, etc.
-* Abstract strings in your application into keyed dictionaries - e.g. flash messages, static text in your views, etc.
-* Store the resulting dictionaries somewhere.
-
-This guide will walk you through the I18n API and contains a tutorial on how to internationalize a Rails application from the start.
+The Ruby [i18n gem](https://github.com/ruby-i18n/i18n), which ships with Ruby on Rails, provides an easy-to-use and extensible framework for translating your application into a single custom language or for providing multi-language support.
```
p8: Starting with a non-Rails gem in the first sentence is not typical for the guides.
Maybe make the introduction more higher-level and move the link to the opening section?
```suggestion
In this guide you will learn how to translate your application into a single non-English language or provide multi-language support.
```

guides/source/i18n.md
```diff
@@ -355,306 +468,340 @@ class ApplicationController < ActionController::Base
...
-I18n.t "success.true"  # => 'True!'
-I18n.t "success.on"    # => 'On!'
-I18n.t "success.false" # => 'False!'
-I18n.t "failure.false" # => Translation Missing
-I18n.t "failure.off"   # => Translation Missing
-I18n.t "failure.true"  # => Translation Missing
-```
+* The locale is an obvious part of the URL.
+* People intuitively grasp in which language the content will be displayed.
+* It is trivial to implement in Rails.
```
p8: This probably applies to other solutions as well.
Also you need to setup your application and server to handle multiple domains, which isn't that trivial?
```suggestion
```

guides/source/i18n.md
p8: If the `divs` are removed this should be deleted (the image as well):
```suggestion
```

guides/source/i18n.md
```diff
@@ -1,352 +1,465 @@
...
+Then inflector-based helpers can use those rules for that locale:
 
-In practice, more robust code is necessary to do this reliably. Iain Hecker's [http_accept_language](https://github.com/iain/http_accept_language/tree/master) library or Ryan Tomayko's [locale](https://github.com/rack/rack-contrib/blob/main/lib/rack/contrib/locale.rb) Rack middleware provide solutions to this problem.
+```ruby
+"aluguel".pluralize(:pt_br)
+# => "aluguéis"
+```
 
-##### Inferring the Locale from IP Geolocation
+When regional locales are used together with `I18n.fallbacks`, inflection rules can also come from a fallback locale. For example, `en-GB` can reuse `:en` inflections when that fallback is mapped explicitly.
```
p8: Maybe show an example here for a fallback as well (and link to the configuration guide)?
````suggestion
When regional locales are used together with [`I18n.fallbacks`][], inflection rules can also come from a fallback locale. For example, `en-GB` can reuse `:en` inflections when that fallback is mapped explicitly.

```ruby
config.i18n.fallbacks = [:"en-GB": :en]
```

[`I18n.fallbacks`]:
    configuring.html#config-i18n-fallbacks
````

File: guides/source/i18n.md [hunk 34]
```diff
@@ -1039,36 +1220,230 @@ en:
...
+or the date and time formats in
+[`activesupport/lib/active_support/locale/en.yml`](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/locale/en.yml).
+With the default `Simple` backend, you can store translations in YAML files or in
+plain Ruby hashes.
+
+NOTE: The i18n library takes a pragmatic approach to locale keys, including only the _locale_ ("language") part, like `:en`, `:pl`, not the _region_ part, like `:"en-US"` or `:"en-GB"`, which are traditionally used for separating "languages" and "regional setting" or "dialects". Many international applications use only the "language" element of a locale such as `:cs`, `:th`, or `:es` (for Czech, Thai, and Spanish). However, there are also regional differences within different language groups that may be important. For instance, in the `:"en-US"` locale you would have $ as a currency symbol, while in `:"en-GB"`, you would have £. Nothing stops you from separating regional and other settings in this way: you just have to provide full "English - United Kingdom" locale in a `:"en-GB"` dictionary.
```
p8: I'm not sure I understand this NOTE.
`rails`, `i18n` and `i18n-rails` all support the region part in locales.
Is the pragmatic approach that excluding the region is supported as well?

Maybe it becomes clearer if we remove the first sentence?
Also, this NOTE seems more at home after we mention "The I18n library uses English as the default locale", as it describes why we use `:en` and not `:"en-US"`.

```suggestion
NOTE: Many international applications use only the "language" element of a locale such as `:cs`, `:th`, or `:es` (for Czech, Thai, and Spanish). However, there are also regional differences within different language groups that may be important. For instance, in the `:"en-US"` locale you would have $ as a currency symbol, while in `:"en-GB"`, you would have £. Nothing stops you from separating regional and other settings in this way: you just have to provide full "English - United Kingdom" locale in a `:"en-GB"` dictionary.
```

guides/source/i18n.md
```diff
@@ -1039,36 +1220,230 @@ en:
...
+```ruby
+I18n.t "success.true"  # => 'True!'
+I18n.t "success.on"    # => 'On!'
+I18n.t "success.false" # => 'False!'
+I18n.t "failure.false" # => Translation Missing
+I18n.t "failure.off"   # => Translation Missing
+I18n.t "failure.true"  # => Translation Missing
+```
+
+### Store Custom Translations
```
p8: Nitpick, most other headings use the "present participle" verb form:
```suggestion
### Storing Custom Translations
```

guides/source/i18n.md
```diff
@@ -1039,36 +1220,230 @@ en:
...
--------------------------------------
+    ```ruby
+    ["apples", "oranges", "pears"].to_sentence
+    # => "apples, oranges, and pears"
+    ```
+
+Advanced Configuration and Setup
+---------------------------------
+
+### Configure the load paths and default locale
```
p8: Nitpick, most other headings use the "present participle" verb form (setting, adding, passing):
```suggestion
### Configuring the load paths and default locale
```

p8 (commented):
Hi @Ridhwana this is looking great.
Some comments, mostly nitpicks.

guides/source/i18n.md
Ridhwana: ✅

hachi8833:
@Ridhwana I opened the following PR to remove the redundant section↓. Hope you'd check this when you have time 🙏

- [[doc][ci-skip] Remove redundant "Setting the Locale from URL Params" section in I18n edge guides by hachi8833 · Pull Request #58169 · rails/rails](https://github.com/rails/rails/pull/58169)

PR #57383, opened 2026-05-18
[ci skip] Simplify Association Callbacks examples in Active Record Guides

flavio-b:
### Motivation / Background

There are a couple of examples in the Guides of how to use association callbacks (`before_add` etc). In one example, the callback loads an association (`books`) and performs validation based on the count of books. Personally, I avoid calling associations in callbacks because it can lead to N+1s and makes the code difficult to test. Plus, the validation in that kind of callback doesn't cover the scenario where one calls `Book.create(author: author)` directly.

I understand that many like to call associations in callbacks, so you'll find arguments on both sides. In any case, I think the example should be more agnostic and simpler to keep the focus on the callback mechanics and less on side-effects and usages. Readers may decide how best to use use the callback according to their preferences.

There was also an example with two callbacks, but I couldn't understand the goal of the second callback. It didn't appear to change any state and had a return value with no clear purpose.

### Detail

This Pull Request simplifies the first association callback example to use simple logging. The second example is shortened. I think it conveys the idea clearly enough. The PR also moves a couple of sentences to organize the flow of the text a bit. I kept the warning about the effects of `throw :abort` in there as reminder for people who might want to do validation in these.

### DIFF
--- a/guides/source/active_record_callbacks.md
+++ b/guides/source/active_record_callbacks.md
@@ -972,54 +972,35 @@ association callbacks:
 
 You can define association callbacks by adding options to the association.
 
-Suppose you have an example where an author can have many books. However, before
-adding a book to the authors collection, you want to ensure that the author has
-not reached their book limit. You can do this by adding a `before_add` callback
-to check the limit.
+Suppose you have an example where an author has many books. Before
+adding a book to the authors collection, you want to log the new addition.
+You can do this by adding a `before_add` callback. Rails passes the object 
+being added or removed to the callback for you to use.
 
 ```ruby
 class Author < ApplicationRecord
-  has_many :books, before_add: :check_limit
+  has_many :books, before_add: :log_book_addition
 
   private
-    def check_limit(_book)
-      if books.count >= 5
-        errors.add(:base, "Cannot add more than 5 books for this author")
-        throw(:abort)
-      end
+    def log_book_addition(book)
+      puts "Adding '#{book.title}' to #{name}'s collection"
     end
+end
+```
+
+At times you may want to perform multiple actions on the associated object. In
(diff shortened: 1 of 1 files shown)


rafaelfranca:
I don't like the new examples since they don't explain the abort behavior, but I also don't like the old example for multiple callbacks since the second callback doesn't make any sense. 

@p8 did we already review this guide? If yes, can the author revisit those examples?

p8:
@rafaelfranca Yes, that guide was already reviewed. I'll notify the author.

flavio-b:
@rafaelfranca, I tried to keep `throw :abort`, but couldn't come up with an example where I'd be super comfortable using it. That was the motivation for proposing a change.

The only case I can see for `throw :abort` in this kind of callback is validation. If validation uses direct fields of `Author` or `Book`, it might be fine, but the moment you need to read associations, it runs into limitations fast:
```ruby
class Author < ApplicationRecord
  has_many :books, before_add: :validate_book_addition
  
  def validate_book_addition(book)
    throw :abort if self.books.count >= 5 # SQL query
    throw :abort unless self.reviewers.all? { it.approve?(book) } # SQL query
    throw :abort unless self.publisher.can_publish?(book) # SQL query
  end
end
```
Even if we preload those associations in `validate_book_addition `, it's still an extra query (one for the `author` instance and one to preload associations) per book added. Now imagine a mass import with several books and authors, things will slow down. So, IMHO, this approach often requires a full refactoring sooner or later, so I'm hesitant to recommend it via an example in the guide.

I'm curious to see if others are finding a good use case for this. If people like the original example with `abort :throw` and loading the association in the callback, I think we could at least replace that other `calculate_shipping_charges` method, and it should be good to go.

PR #57388, opened 2026-05-19
`http_cache_forever` now accept an optional `last_modified:` keyword parameter.

byroot:
It still defaults to January 1st 2011, but you now can substitute it for a relevant time if there is one.

Then use that new option in Active Storage ProxyController.

### DIFF
--- a/actionpack/test/controller/render_test.rb
+++ b/actionpack/test/controller/render_test.rb
@@ -1025,6 +1025,13 @@ end
 class HttpCacheForeverTest < ActionController::TestCase
   class HttpCacheForeverController < ActionController::Base
     def cache_me_forever
+      last_modified = params[:last_modified] && Time.iso8601(params[:last_modified])
+      http_cache_forever(public: params[:public], last_modified: last_modified) do
+        render plain: "hello"
+      end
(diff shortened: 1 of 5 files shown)


actionpack/test/controller/render_test.rb
```diff
@@ -1025,6 +1025,13 @@ def test_live_head_ok
 class HttpCacheForeverTest < ActionController::TestCase
   class HttpCacheForeverController < ActionController::Base
     def cache_me_forever
+      last_modified = params[:last_modified] && Time.iso8601(params[:last_modified])
+      http_cache_forever(public: params[:public], last_modified: last_modified) do
+        render plain: "hello"
+      end
+    end
+
+    def cache_me_forever_last_modified
```
chaadow, line 1034: @byroot Shouldn't there be a test that does `get :cache_me_forever_last_modified` or maybe reverse the content of `def cache_me_forever` in here? ( or both )

byroot: `cache_me_forever_last_modified` is just a leftover from a previous version of this pR. I'll remove it.

PR #57393, opened 2026-05-19
Load image processing backend upfront

janko:
### Motivation / Background

Processing the first Active Storage variant after deploy has extra overhead today.

### Detail

The image_processing gem has autoloading that will load the referenced backend on-demand. However, that means the first processed variant after deploy is paying the penalty for loading the backend code.

We avoid this overhead by loading the backend upfront. This also improves copy-on-write for preforking web servers, decreasing memory usage of forked subprocesses.

### Additional information

Loading `image_processing/vips` has 19MB extra on top of `image_processing`, so that memory would now be shared across forked subprocesses.

I measured the require overhead to be 50-100ms on my M1 MacBook Air, but I haven't tested on a production Linux server with Bootsnap activated.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Require image processing backend upfront when Active Storage is being loaded.
+
+    This removes extra overhead when processing first variant after deploy and improves copy-on-write for preforking web servers.
+
+    *Janko Marohnić*
+
 *   ActiveStorage ProxyController now set relevant `Last-Modified`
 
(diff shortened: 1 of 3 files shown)


activestorage/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+*   Require image processing backend upfront when Active Storage is being loaded.
+
+    This removes extra overhead when processing first variant after deploy and improves copy-on-write for preforking web servers.
+
```
guilleiguaran, line 4: Please add your name or username below the changelog entry.

janko: Added ✅

PR #57395, opened 2026-05-20
Add query predicate hooks for Active Record types

kirs:
This adds internal predicate-builder hooks that let Active Record attribute types customize how attributes and query values are represented in hash `where` predicates.

Types can define:

- `query_attribute(attribute)` to transform the left-hand side of predicates
- `query_value(attribute, value, predicate_builder:)` to transform bound values while preserving prepared binds

The predicate builder uses these hooks for equality predicates, array predicates, and ranges.

## Why this is needed

Some database-backed types need a different SQL representation for reads than for writes.

For example, a MySQL UUID type stored in `binary(16)` should let callers pass readable UUID strings while rendering predicates as:

```sql
id = UUID_TO_BIN(?)
```

Another case is a text column that stores the original user-provided value, but should compare using a normalized expression:

```sql
lower(custom_immutable_unaccent(name)) = lower(custom_immutable_unaccent(?))
```

These transformations cannot live in `cast` or `serialize`, because they are only needed when building predicates.

### DIFF
--- a/activerecord/lib/active_record/relation/predicate_builder.rb
+++ b/activerecord/lib/active_record/relation/predicate_builder.rb
@@ -47,7 +47,7 @@ module ActiveRecord
     #     end
     #     ActiveRecord::PredicateBuilder.new("users").register_handler(MyCustomDateRange, handler)
     def register_handler(klass, handler)
-      @handlers.unshift([klass, handler])
+      @handlers.unshift([klass, handler_accepts_type?(handler) ? handler : handler_without_type(handler)])
     end
 
     def [](attr_name, value, operator = nil)
@@ -58,34 +58,27 @@ module ActiveRecord
       value = value.id if value.respond_to?(:id)
       type = table.type(attribute.name)
 
+      predicate_for(attribute, value, operator, type)
+    end
+
+    def predicate_for(attribute, value, operator, type)
       if operator ||= type.force_equality?(value) && :eq
         if type.transforms_query_predicates?
-          build_predicate(attribute, value, operator, type, true)
+          right = query_value(attribute, value, type)
+          left = right.nil? ? attribute : type.query_attribute(attribute)
         else
-          bind = build_bind_attribute(attribute.name, value, type)
-          attribute.public_send(operator, bind)
+          right = build_bind_attribute(attribute.name, value, type)
+          left = attribute
         end
+
+        left.public_send(operator, right)
       else
-        handler_for(value).call(attribute, value)
+        handler_for(value).call(attribute, value, type)
       end
     end
 
-    def build_predicate(attribute, value, operator = :eq, type = table.type(attribute.name), transformable = type.transforms_query_predicates?)
+    def array_predicate_for(attribute, values, type, transformable)
       if transformable
-        right = query_value(attribute, value, type)
-        left = right.nil? ? attribute : type.query_attribute(attribute)
-      else
-        right = build_bind_attribute(attribute.name, value, type)
-        left = attribute
-      end
-
-      left.public_send(operator, right)
-    end
-
-    def build_array_predicate(attribute, values)
-      type = table.type(attribute.name)
-
-      if type.transforms_query_predicates?
         type.query_attribute(attribute).in(
           values.map { |value| query_value(attribute, value, type) }
         )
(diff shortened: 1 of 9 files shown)


activerecord/lib/active_record/relation/predicate_builder.rb
```diff
@@ -57,17 +57,51 @@ def [](attr_name, value, operator = nil)
     def build(attribute, value, operator = nil)
       value = value.id if value.respond_to?(:id)
       if operator ||= table.type(attribute.name).force_equality?(value) && :eq
-        bind = build_bind_attribute(attribute.name, value)
-        attribute.public_send(operator, bind)
+        build_predicate(attribute, value, operator)
```
rafaelfranca: This is probably the most possible branch that is now slowed down because of the possibility of an override. Can we make sure the common path is short circuited?

activerecord/lib/active_record/relation/predicate_builder.rb
```diff
@@ -57,17 +57,51 @@ def [](attr_name, value, operator = nil)
...
+
+      left.public_send(operator, right)
+    end
+
     def build_bind_attribute(column_name, value)
       Relation::QueryAttribute.new(column_name, value, table.type(column_name))
     end
 
+    def query_attribute(attribute)
+      type = table.type(attribute.name)
```
rafaelfranca: All those table.type lookups in those methods flows are annoying me. It feels like we should be passing the type and pushing the lookup up a level

activerecord/lib/active_record/relation/predicate_builder/array_handler.rb
```diff
@@ -20,7 +20,14 @@ def call(attribute, value)
           case values.length
           when 0 then NullPredicate
           when 1 then predicate_builder.build(attribute, values.first)
-          else Arel::Nodes::HomogeneousIn.new(values, attribute, :in)
+          else
+            if predicate_builder.query_transformable?(attribute)
+              predicate_builder.query_attribute(attribute).in(
+                values.map { |value| predicate_builder.query_value(attribute, value) }
```
rafaelfranca: Here for example we are doing type lookup many times for the values when the attribute is always the same

activerecord/lib/active_record/relation/predicate_builder/array_handler.rb
```diff
@@ -20,7 +20,14 @@ def call(attribute, value)
           case values.length
           when 0 then NullPredicate
           when 1 then predicate_builder.build(attribute, values.first)
-          else Arel::Nodes::HomogeneousIn.new(values, attribute, :in)
+          else
```
rafaelfranca: This branch feels like should be pushed down to the predicate builder

activerecord/lib/active_record/relation/predicate_builder/basic_object_handler.rb
```diff
@@ -8,8 +8,7 @@ def initialize(predicate_builder)
       end
 
       def call(attribute, value)
-        bind = predicate_builder.build_bind_attribute(attribute.name, value)
-        attribute.eq(bind)
+        predicate_builder.build_predicate(attribute, value)
```
rafaelfranca: Let's just pass all arguments and avoid the default arguments in the method.

activerecord/lib/active_record/relation/predicate_builder.rb
```diff
@@ -56,16 +56,65 @@ def [](attr_name, value, operator = nil)
...
+          attribute.public_send(operator, bind)
+        end
       else
         handler_for(value).call(attribute, value)
       end
     end
 
-    def build_bind_attribute(column_name, value)
-      Relation::QueryAttribute.new(column_name, value, table.type(column_name))
+    def build_predicate(attribute, value, operator = :eq, type = table.type(attribute.name), transformable = type.transforms_query_predicates?)
```
rafaelfranca: I think this method should be renamed to `predicate_for`. `build` should be the only entry point of the PredicateBuilder. All other methods are just helpers, they aren't building it.

We should also avoid the default value and always pass them.

activerecord/lib/active_record/relation/predicate_builder.rb
```diff
@@ -56,16 +56,65 @@ def [](attr_name, value, operator = nil)
...
+          range.exclude_end?
+        )
+      )
+    end
+
+    def build_bind_attribute(column_name, value, type = table.type(column_name))
+      Relation::QueryAttribute.new(column_name, value, type)
+    end
+
+    def query_value(attribute, value, type = table.type(attribute.name))
```
rafaelfranca: This one should be private

PR #57403, opened 2026-05-20
Update Active Storage for ImageProcessing 2.0

janko:
### Motivation / Background

[ImageProcessing 2.0](https://github.com/janko/image_processing/blob/master/CHANGELOG.md#200-2026-05-20) has been released, which brings some breaking changes.

### Detail

ImageProcessing 2.0 now requires adding `ruby-vips` and/or `mini_magick` gems explicitly to the Gemfile, so I updated instructions and the generator. I also added `LoadError` branches for those gems.

The new version now also [blocks untrusted formats](https://www.libvips.org/2022/05/28/What's-new-in-8.13.html#blocking-of-unfuzzed-loaders) by default on libvips 8.13+. These include BPM, PSD and ICO, which are currently accepted as "variable" content types by Active Storage. Not sure what's the best way to handle this.

### Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,7 +118,9 @@ group :storage do
   gem "aws-sdk-s3", require: false
   gem "google-cloud-storage", "~> 1.11", require: false
 
-  gem "image_processing", "~> 1.2"
+  gem "image_processing", "~> 2.0"
+  gem "ruby-vips", "~> 2.3"
+  gem "mini_magick", "~> 5.0"
 end
 
 # Action Mailbox
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -278,9 +278,7 @@ GEM
       mutex_m
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
-    image_processing (1.14.0)
-      mini_magick (>= 4.9.5, < 6)
-      ruby-vips (>= 2.0.17, < 3)
+    image_processing (2.0.1)
     importmap-rails (2.1.0)
       actionpack (>= 6.0.0)
       activesupport (>= 6.0.0)
@@ -517,7 +515,7 @@ GEM
       rubocop-performance
       rubocop-rails
     ruby-progressbar (1.13.0)
-    ruby-vips (2.2.2)
+    ruby-vips (2.3.0)
       ffi (~> 1.12)
       logger
     ruby2_keywords (0.0.5)
@@ -693,7 +691,7 @@ DEPENDENCIES
   dartsass-rails
   debug (>= 1.1.0)
   google-cloud-storage (~> 1.11)
-  image_processing (~> 1.2)
+  image_processing (~> 2.0)
   importmap-rails (>= 1.2.3)
   jbuilder
   jsbundling-rails
@@ -703,6 +701,7 @@ DEPENDENCIES
   libxml-ruby
   listen (~> 3.3)
   mdl (!= 0.13.0)
+  mini_magick (~> 5.0)
   minitest (~> 6.0)
   minitest-mock
   minitest-retry
@@ -733,6 +732,7 @@ DEPENDENCIES
   rubocop-performance
   rubocop-rails
   rubocop-rails-omakase
+  ruby-vips (~> 2.3)
   rubyzip (~> 2.0)
   sdoc (~> 2.6.4)
   selenium-webdriver (>= 4.20.0)
@@ -846,7 +846,7 @@ CHECKSUMS
   hashdiff (1.1.2) sha256=2c30eeded6ed3dce8401d2b5b99e6963fe5f14ed85e60dd9e33c545a44b71a77
   httpclient (2.9.0) sha256=4b645958e494b2f86c2f8a2f304c959baa273a310e77a2931ddb986d83e498c8
   i18n (1.14.8) sha256=285778639134865c5e0f6269e0b818256017e8cde89993fdfcbfb64d088824a5
-  image_processing (1.14.0) sha256=754cc169c9c262980889bec6bfd325ed1dafad34f85242b5a07b60af004742fb
+  image_processing (2.0.1) sha256=69c93216dafe1179343a6c3ae8bfa554e3c1f5eece4043f255595e76df2cbf3f
   importmap-rails (2.1.0) sha256=9f10c67d60651a547579f448100d033df311c5d5db578301374aeb774faae741
   io-console (0.8.1) sha256=1e15440a6b2f67b6ea496df7c474ed62c860ad11237f29b3bd187f054b925fcb
   irb (1.15.2) sha256=222f32952e278da34b58ffe45e8634bf4afc2dc7aa9da23fed67e581aa50fdba
@@ -955,7 +955,7 @@ CHECKSUMS
   rubocop-rails (2.30.3) sha256=fc5a6506daa916d15e282cc806943afa64a020bf592b93a94025d89a2a78a715
   rubocop-rails-omakase (1.0.0) sha256=26635821283d6a03dc8b68172123716a62db96123eb035f736ebb91283f96418
   ruby-progressbar (1.13.0) sha256=80fc9c47a9b640d6834e0dc7b3c94c9df37f08cb072b7761e4a71e22cff29b33
-  ruby-vips (2.2.2) sha256=af5be8e9f2857a98424d7606ecfc2c21c5e107d0802915190dcc093f56e4dd5b
+  ruby-vips (2.3.0) sha256=e685ec02c13969912debbd98019e50492e12989282da5f37d05f5471442f5374
   ruby2_keywords (0.0.5) sha256=ffd13740c573b7301cf7a2e61fc857b2a8e3d3aff32545d6f8300d8bae10e3ef
   rubyzip (2.4.1) sha256=8577c88edc1fde8935eb91064c5cb1aef9ad5494b940cf19c775ee833e075615
   rufus-scheduler (3.9.2) sha256=55fa9e4db0ff69d7f38c804f17baba0c9bce5cba39984ae3c5cf6c039d1323b9
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Update Active Storage for ImageProcessing 2.0.
+
+    ImageProcessing 2.0 now requires adding `ruby-vips` or `mini_magick` gems explicitly to the Gemfile.
+
+    *Janko Marohnić*
+
 *   Prevent `ActiveStorage.touch_attachment_records = false` from crashing the attachment of a Blob.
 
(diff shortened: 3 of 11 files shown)


janko:
I see that the tests for SVG, PSD, BPM and ICO are failing, because they're now an "unknown" format to libvips (in reality they're blocked). I see two options:

1. Fail with a more explicit user-friendly error that makes it clear the format is supported by libvips but blocked. We could set `Vips.block_untrusted(false)` temporarily around the failing tests.
2. Have Active Storage call `Vips.block_untrusted(false)` and allow all formats to avoid a breaking change. That can appear as defeating the purpose of a safe default, but Active Storage already has an allowlist of image formats, so it's not like it would accept *anything*.

janko:
As it's not immediately clear to me how to better communicate blocked formats, I went with the option 2 to retain backwards compatibility.

rafaelfranca (approved):

jeromedalbert:
This PR breaks `rails new` on any system that doesn't have vips (see issue https://github.com/rails/rails/issues/57612). Here is a tentative fix PR: https://github.com/rails/rails/pull/57613.

Edgarruiz8585:
> ### Motivation / Background
> 
> [ImageProcessing 2.0](https://github.com/janko/image_processing/blob/master/CHANGELOG.md#200-2026-05-20) has been released, which brings some breaking changes.
> 
> ### Detail
> 
> ImageProcessing 2.0 now requires adding `ruby-vips` and/or `mini_magick` gems explicitly to the Gemfile, so I updated instructions and the generator. I also added `LoadError` branches for those gems.
> 
> The new version now also [blocks untrusted formats](https://www.libvips.org/2022/05/28/What's-new-in-8.13.html#blocking-of-unfuzzed-loaders) by default on libvips 8.13+. These include BPM, PSD and ICO, which are currently accepted as "variable" content types by Active Storage. Not sure what's the best way to handle this.
> 
> ### Checklist
> 
> Before submitting the PR make sure the following are checked:
> 
> * [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
> * [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
> * [x] Tests are added or updated if you fix a bug or add a feature.
> * [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.
>  by: Edgarruiz8585

flavorjones:
@janko Would an alternative to this be to accept that untrusted loaders are blocked, and then re-enable only the untrusted loaders that are absolutely needed for Rails?

I **think** this would only include `magickload` for BMP/ICO/PSD support.

I'm asking because in most Rails apps there's no good reason to support the other untrusted loaders by default (pdfload, matload, fitsload, csvload, etc.). It might be better for applications to explicitly enable an untrusted loader as needed (which they can do in an app initializer).

As it pertains to backwards-compatibility, we could consider doing it only when image_processing 2.0 is being used?

janko:
@flavorjones That sounds like a reasonable compromise to me. It does seem to me like libvips would then inherit any unsafe loaders from ImageMagick, though I might be wrong. Do you know how to selectively enable `magickload` operation with `ruby-vips`?

flavorjones:
> It does seem to me like libvips would then inherit any unsafe loaders from ImageMagick

@janko This is correct, and I'm not totally sure whether it's better or worse! I need to do a bit more exploration of the exposed API surface area, but I think this is what the matrix looks like (note this will vary across distros and platforms):

| Format    | libvips loader     | Parser behind libvips            | IM coder | Parser behind IM            |
| --------- | ------------------ | -------------------------------- | -------- | --------------------------- |
| SVG       | `svgload`          | librsvg (`librsvg-2.so.2`)       | `SVG`    | librsvg 2.62.3, same object |
| OpenEXR   | `openexrload`      | OpenEXR (`libOpenEXR-3_4.so.33`) | `EXR`    | OpenEXR 3.4.13, same object |
| JPEG 2000 | `jp2kload`         | OpenJPEG (`libopenjp2.so.7`)     | `JP2`    | OpenJPEG 2.5.4, same object |
| JPEG-XL   | `jxlload` (module) | libjxl (`libjxl.so.0.12`)        | `JXL`    | libjxl 0.12.0, same object  |
| MATLAB          | `matload`      | libmatio (absent in this build) | `MAT`    | IM's own `coders/mat.c`, level 5 only |
| FITS            | `fitsload`     | cfitsio (`libcfitsio.so.10`)    | `FITS`   | IM's own FITS coder                   |
| Radiance HDR    | `radload`      | libvips internal C              | `HDR`    | IM's own RGBE coder                   |
| PPM/PNM/PGM/PBM | `ppmload`      | libvips internal C              | `PNM`    | IM's own PNM coder                    |
| `.v` / `.vips`  | `vipsload`     | libvips internal C              | `VIPS`   | IM's own VIPS coder                   |
| PDF    | `pdfload`      | poppler (`libpoppler.so.162`, module) | `PDF`    | MagickCore PDF coder shelling out to **Ghostscript** (`gs -dSAFER …`) |
| CSV             | `csvload`                  | libvips internal C                      | No ImageMagick image coder; also no `is_a`, so it is suffix-selected rather than content-sniffed |
| Analyze / NIfTI | `analyzeload`, `niftiload` | libvips internal C, niftiio             | No ImageMagick coder                                                                             |
| Headerless raw  | `rawload`                  | libvips internal C                      | No `is_a`, so `vips_foreign_find_load` can never select it by content                            |
| OpenSlide       | `openslideload`            | OpenSlide (`libopenslide.so.1`, module) | Not tested; `.svs` is TIFF-based and `tiffload` is trusted, so it likely never reaches magick    |
| BMP                   | none                  | `BMP`    | IM internal      |
| TGA                   | none                  | `TGA`    | IM internal      |
| PCX                   | none                  | `PCX`    | IM internal      |
| DNG / CR2 / NEF / ARW | none in this build    | `DNG`    | libraw 0.22.2    |

The one that jumps out to me as a downgrade is `pdfload` which would move off of libpoppler and instead shell out to ghostscript. We might be able to block that by introducing an IM `policy.xml` but ... I haven't explored that at all.

> Do you know how to selectively enable magickload/

Yup, looks like we'd update the code added to activestorage/lib/active_storage/transformers/vips.rb in this PR to something like:

```ruby
Vips.block_untrusted(true)

# have to unblock both variations to be effective across all platforms
Vips.block("VipsForeignLoadMagick", false)
Vips.block("VipsForeignLoadMagick7", false)
```

janko:
@flavorjones If we enabled `magickload`, won't that mean we enable all the security vulnerabilities ImageMagick brings? Can't users protect themselves by simply not installing ImageMagick in their Docker images (in which case `magickload` shouldn't be activated)? If we could include `magickload` but limit it *only* to BMP/ICO/PSD, that would make sense to me.

flavorjones:
@janko Great questions. If we look at the default Rails container `ruby:4.0-slim` which today is based on Debian Trixie, `libvips` declares `libmagickcore7` and many other libraries as direct dependencies:

```
Installing:
  libvips42t64

Installing dependencies:
  adduser                  libgdk-pixbuf-2.0-0       libopenexr-3-1-30
  dbus                     libgdk-pixbuf2.0-bin      libopenjp2-7
  dbus-bin                 libgdk-pixbuf2.0-common   libopenslide0
  dbus-daemon              libglib2.0-0t64           libp11-kit0
  dbus-session-bus-common  libglib2.0-data           libpango-1.0-0
  dbus-system-bus-common   libgnutls30t64            libpangocairo-1.0-0
  dirmngr                  libgomp1                  libpangoft2-1.0-0
  fontconfig               libgpg-error-l10n         libpaper-utils
  fontconfig-config        libgpg-error0             libpaper2
  fonts-dejavu-core        libgpgme11t64             libpixman-1-0
  fonts-dejavu-mono        libgpgmepp6t64            libpng16-16t64
  fonts-droid-fallback     libgpm2                   libpoppler-glib8t64
  fonts-noto-mono          libgraphite2-3            libpoppler147
  fonts-urw-base35         libgs-common              libpsl5t64
  ghostscript              libgs10                   libraw23t64
  gnupg                    libgs10-common            libreadline8t64
  gnupg-l10n               libgssapi-krb5-2          librsvg2-2
  gnupg-utils              libharfbuzz0b             librsvg2-common
  gpg                      libhdf5-310               librtmp1
  gpg-agent                libheif-plugin-aomenc     libsasl2-2
  gpg-wks-client           libheif-plugin-dav1d      libsasl2-modules
  gpgconf                  libheif-plugin-libde265   libsasl2-modules-db
  gpgsm                    libheif-plugin-x265       libsharpyuv0
  gpgv                     libheif1                  libsm6
  imagemagick-7-common     libhwy1t64                libspng0
  krb5-locales             libice6                   libssh2-1t64
  libaec0                  libidn12                  libsz2
  libaom3                  libidn2-0                 libtasn1-6
  libapparmor1             libijs-0.35               libthai-data
  libarchive13t64          libimagequant0            libthai0
  libassuan9               libimath-3-1-29t64        libtiff6
  libatomic1               libjbig0                  libunistring5
  libavahi-client3         libjbig2dec0              libwebp7
  libavahi-common-data     libjpeg62-turbo           libwebpdemux2
  libavahi-common3         libjxl0.11                libwebpmux3
  libbrotli1               libk5crypto3              libx11-6
  libcairo-gobject2        libkeyutils1              libx11-data
  libcairo2                libkrb5-3                 libx265-215
  libcfitsio10t64          libkrb5support0           libxau6
  libcgif0                 libksba8                  libxcb-render0
  libcom-err2              liblcms2-2                libxcb-shm0
  libcups2t64              libldap-common            libxcb1
  libcurl3t64-gnutls       libldap2                  libxdmcp6
  libcurl4t64              liblerc4                  libxext6
  libdatrie1               liblqr-1-0                libxml2
  libdav1d7                libltdl7                  libxrender1
  libdbus-1-3              libmagickcore-7.q16-10    libxt6t64
  libde265-0               libmatio13                pinentry-curses
  libdeflate0              libncursesw6              poppler-data
  libexif12                libnghttp2-14             publicsuffix
  libexpat1                libnghttp3-9              readline-common
  libfftw3-double3         libngtcp2-16              sensible-utils
  libfontconfig1           libngtcp2-crypto-gnutls8  shared-mime-info
  libfontenc1              libnpth0t64               x11-common
  libfreetype6             libnspr4                  xdg-user-dirs
  libfribidi0              libnss3                   xfonts-encodings
  libgcrypt20              libnuma1                  xfonts-utils
```

So that's really inflexible and doesn't allow simple uninstallation of libmagick or other untrusted libraries. An ImageMagick policy file to limit to only those file types is definitely another option but that feels tricky and not necessarily easily portable to every target system (at least from a Rails configuration point of view).

So I think at some point we probably just want to flat block unfuzzed parsers, like image_processing does by default in 2.0 (and the libvips maintainers recommend).

PR #57406, opened 2026-05-20
[#57402] Fix Active Record Pool Reaper thread leak after `Parallelization#shutdown`

### LINKED ISSUE #57402: ActiveRecord connection pool reaper thread leaks after parallel testing
### Steps to reproduce

`ActiveRecord::ConnectionAdapters::ConnectionPool::Reaper` spawns a class-level singleton thread (named `"AR Pool Reaper"`) per reaping `frequency`. It only terminates when every pool registered for that frequency has been `discard!`ed or garbage-collected (see `reaper.rb#spawn_thread`).

`ActiveSupport::Testing::Parallelization#shutdown` never tells AR to discard the parent process's connection pool. `ActiveRecord::TestDatabases` only registers an `after_fork_hook` (to create per-worker databases) and no equivalent end-of-tests cleanup. Result: in a process that uses parallel testing and then expects to exit naturally, the AR Pool Reaper thread keeps sleeping forever and prevents process termination.

Bug report template — fails on `rails:main` (verified at `rails-c3c65d10b0bb`):

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
end

require "active_support"
require "active_record"
require "minitest/autorun"

  adapter: "sqlite3",
  database: ":memory:",
  reaping_frequency: 5
)
ActiveRecord::Base.connection_pool.with_connection { |c| c.execute("SELECT 1") }
sleep 0.1

class ARPoolReaperLeakTest < Minitest::Test
  def test_ar_pool_reaper_is_cleaned_up_after_parallelization_shutdown
    assert Thread.list.find { |t| t.name == "AR Pool Reaper" },
      "Setup precondition: AR Pool Reaper thread should exist"

    parallelization.start
    parallelization.shutdown

    reaper = Thread.list.select { |t| t.alive? && t.name == "AR Pool Reaper" }

    assert_empty reaper,
      "Expected parallel-test teardown to discard the parent's connection pool. " \
      "Leaked:\n" + reaper.map { |t| Array(t.backtrace).first(3).join("\n  ") }.join("\n")
  end
end
```

### Expected behavior

After parallel testing finishes, the parent process's AR connection pool is discarded (or otherwise unregistered from the reaper), so the `"AR Pool Reaper"` background thread terminates and the process can exit cleanly.

### Actual behavior

The reaper thread is still alive after `Parallelization#shutdown`:

```
Failure: ARPoolReaperLeakTest#test_ar_pool_reaper_is_cleaned_up_after_parallelization_shutdown
Expected parallel-test teardown to discard the parent's connection pool. Leaked:
  rails-.../activerecord/lib/active_record/connection_adapters/abstract/connection_pool/reaper.rb:74:in `sleep'
  rails-.../activerecord/lib/active_record/connection_adapters/abstract/connection_pool/reaper.rb:74:in `block in spawn_thread'
Expected [#<Thread:0x...@AR Pool Reaper sleep>] to be empty.
```

We hit this in a Rails app's CI alongside #57027 — minitest summary prints `0 failures`, then the process hangs at exit and the CI container is killed by a timeout. The DRb leak in `Parallelization#shutdown` (filed separately) compounds the problem.

Possible fixes:

1. `ActiveRecord` registers an `ActiveSupport::Testing::Parallelization.run_cleanup_hook` (or new end-of-tests hook) that calls `ActiveRecord::Base.connection_handler.each_connection_pool(&:discard!)` on the parent.
2. Alternatively, have the Reaper itself opt-out of preventing process exit (e.g., mark the thread as a daemon thread via `Thread.current.abort_on_exception` / its modern equivalent).

(1) is more surgical and contained in the test-framework code path; (2) changes Reaper's behavior in production processes too, which is more invasive. I'd lean toward (1) but happy to defer to maintainer preference. Will open a PR with a regression test once direction is settled.

### System config

* Rails version: `main` (verified at `rails-c3c65d10b0bb`); also reproduced on `8-0-stable` (8.0.4.1) and `8-1-stable` (8.1.3).
* Ruby version: 3.3.10 (arm64-darwin25), also Linux glibc x86_64 in CI.

Affects both `8-0-stable` and `8-1-stable` (and presumably older branches). Backport candidates.

ruyrocha:
### Motivation / Background

After parallelized test runs, the parent process leaks the Active Record Pool Reaper thread, holding open connection pools and file descriptors indefinitely. In long-lived CI processes running many suites sequentially this accumulates threads and open DB connections.

Fixes #57402 

### Detail

Three layered bugs caused the leak:

1. `Parallelization#shutdown` never called `run_cleanup_hooks`, so no parent-side cleanup ran at all.
2. AR never registered a `run_cleanup_hook` to discard its pools.
3. `clear_all_connections!` calls `disconnect!` not `discard!`, so even if a hook had been registered the reaper thread would not have been stopped.

This PR fixes all three: `Parallelization#shutdown` now calls `run_cleanup_hooks` after workers finish; Active Record registers a cleanup hook that calls `discard!` on all connection pools; and `ConnectionPool#discard!` now calls `Reaper.discard_pool`, which removes the pool from the reaper's registry and immediately kills and joins the reaper thread if no pools remain at that frequency.

### Additional information

Reproduction script:

```ruby
require "bundler/inline"
gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
end
require "active_support"
require "active_record"
require "active_record/test_databases"
require "minitest/autorun"

  adapter: "sqlite3",
  database: ":memory:",
  reaping_frequency: 5
)
ActiveRecord::Base.connection_pool.with_connection { |c| c.execute("SELECT 1") }
sleep 0.1

class ARPoolReaperLeakTest < Minitest::Test
  def test_ar_pool_reaper_is_cleaned_up_after_parallelization_shutdown
    assert Thread.list.find { |t| t.name == "Active Record Pool Reaper" }

    parallelization.start
    parallelization.shutdown

    assert_empty Thread.list.select { |t| t.alive? && t.name == "Active Record Pool Reaper" }
  end
end
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,16 @@
+*   Fix Active Record Pool Reaper thread leak after `Parallelization#shutdown`.
+
+    After parallelized test runs, the parent process leaked the Active Record Pool
+    Reaper thread, holding open connection pools and file descriptors for up to
+    `reaping_frequency` seconds (or indefinitely if never reaped).
+
+    Three fixes: `Parallelization#shutdown` now calls `run_cleanup_hooks`;
+    Active Record registers a cleanup hook that discards all connection pools; and
+    `ConnectionPool#discard!` now calls `Reaper.discard_pool`, which immediately
+    kills and joins the reaper thread when no pools remain at that frequency.
+
+    *Ruy Rocha*
+
 *   Reset `lock_version` after a nested savepoint rollback.
 
     When a record was saved inside a `transaction(requires_new: true)` block
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -555,6 +555,7 @@ module ActiveRecord
         @reaper_lock.synchronize do
           synchronize do
             return if self.discarded?
+            Reaper.discard_pool(self)
             @connections.each do |conn|
               conn.discard!
             end
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool/reaper.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool/reaper.rb
@@ -53,6 +53,23 @@ module ActiveRecord
             end
           end
 
+          def discard_pool(pool) # :nodoc:
+            @mutex.synchronize do
+              @pools.each do |frequency, refs|
+                refs.reject! do |ref|
+                  ref.__getobj__ == pool
+                rescue WeakRef::RefError
+                  true
+                end
+
+                if refs.empty?
+                  @pools.delete(frequency)
+                  @threads.delete(frequency)&.tap { |t| t.kill; t.join }
+                end
+              end
+            end
+          end
+
           def pools(refs = nil) # :nodoc:
             refs ||= @mutex.synchronize { @pools.values.flatten(1) }
 
--- a/activerecord/lib/active_record/test_databases.rb
+++ b/activerecord/lib/active_record/test_databases.rb
@@ -16,6 +16,10 @@ module ActiveRecord
       end
     end
 
+    ActiveSupport::Testing::Parallelization.run_cleanup_hook do
+      ActiveRecord::Base.connection_handler.each_connection_pool.each(&:discard!)
+    end
+
     def self.create_and_load_schema(i, env_name:)
       old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"
 
--- a/activerecord/test/cases/reaper_test.rb
+++ b/activerecord/test/cases/reaper_test.rb
@@ -200,6 +200,48 @@ module ActiveRecord
         pool.discard!
       end
 
+      def test_discard_pool_removes_pool_from_registry
+        pool_config = duplicated_pool_config(reaping_frequency: "0.1")
+        pool = ConnectionPool.new(pool_config)
+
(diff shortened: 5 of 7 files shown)


rafaelfranca:
Can you please write tests for all this new code being added?

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+*   Fix AR Pool Reaper thread leak after `Parallelization#shutdown`.
```
rafaelfranca:
```suggestion
*   Fix Active Record Pool Reaper thread leak after `Parallelization#shutdown`.
```

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+
+    After parallelized test runs, the parent process leaked the AR Pool Reaper
```
rafaelfranca:
```suggestion
    After parallelized test runs, the parent process leaked the Active Record Pool Reaper
```

activerecord/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+    thread, holding open connection pools and file descriptors for up to
+    `reaping_frequency` seconds (or indefinitely if never reaped).
+
+    Three fixes: `Parallelization#shutdown` now calls `run_cleanup_hooks`;
+    AR registers a cleanup hook that discards all connection pools; and
```
rafaelfranca:
```suggestion
    Active Record registers a cleanup hook that discards all connection pools; and
```

rafaelfranca (approved):

PR #57407, opened 2026-05-21
[#57387] Fix STI subclass records leaking into preloaded has_many :through...

### LINKED ISSUE #57387: Preloading an STI-subclass `has_many :through` together with the parent STI class's `has_many :through` leaks sibling-subclass records
### Steps to reproduce

A `has_many :through` association with `class_name:` pointing to an STI subclass and `source:` pointing to the parent STI class returns records of sibling STI types when it is preloaded **alongside** the parent class's `has_many :through` association. Loading the association directly, or preloading it alone, works correctly.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails", "7.2.3"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

  create_table :posts

  create_table :tags do |t|
    t.string :name
    t.string :type, null: false
  end

  create_table :taggings do |t|
    t.belongs_to :post
    t.belongs_to :tag
  end
end

class Tagging < ActiveRecord::Base
  belongs_to :post
  belongs_to :tag
end

  has_many :taggings
  has_many :posts, through: :taggings
end

class Category < Tag; end
class Keyword  < Tag; end

  has_many :taggings
  has_many :tags,       through: :taggings
  has_many :categories, class_name: "Category", through: :taggings, source: :tag
end

  def setup
    @post = Post.create!
    @post.taggings.create!(tag: Category.create!(name: "news"))
    @post.taggings.create!(tag: Keyword.create!(name: "breaking"))
  end

  # passes: standalone association load applies the STI filter
  def test_without_preload_categories_is_filtered_by_sti
    p = Post.find(@post.id)
    assert_equal ["Category"], p.categories.map { |t| t.class.name }.uniq
  end

  # passes: preloading just the subclass association applies the STI filter
  def test_preloading_only_categories_is_filtered_by_sti
    p = Post.preload(:categories).find(@post.id)
    assert_equal ["Category"], p.categories.map { |t| t.class.name }.uniq
  end

  # FAILS: preloading the parent STI association together with the subclass association
  #        causes `post.categories` to contain Keyword records.
  def test_preloading_tags_and_categories_should_filter_categories_by_sti
    p = Post.preload(:tags, :categories).find(@post.id)

    assert_equal ["Category"],
                 p.categories.map { |t| t.class.name }.uniq,
                 "expected only Category in post.categories, got: " \
                   "#{p.categories.map { |t| [t.class.name, t.name] }.inspect}"
  end
end
```

Running this prints (final lines):

```
1) Failure:
BugTest#test_preloading_tags_and_categories_should_filter_categories_by_sti:
expected only Category in post.categories, got: [["Category", "news"], ["Keyword", "breaking"]].
Expected: ["Category"]
  Actual: ["Category", "Keyword"]

3 runs, 3 assertions, 1 failures, 0 errors, 0 skips
```

The SQL emitted by the failing case is a single un-typed lookup that loads every related row of the parent STI table:

```sql
SELECT "taggings".* FROM "taggings" WHERE "taggings"."post_id" = 1
SELECT "tags".*     FROM "tags"     WHERE "tags"."id" IN (1, 2)
```

…and the resulting `Tag` rows (instantiated as their STI subclasses) are assigned to **both** `post.tags` **and** `post.categories`.

For comparison, `preload(:categories)` alone emits the expected, filtered query:

```sql
SELECT "tags".* FROM "tags"
INNER JOIN "taggings" ON "tags"."id" = "taggings"."tag_id"
WHERE "tags"."type" = 'Category' AND "taggings"."post_id" = 1
```

### Expected behavior

`post.categories` should return only `Category` records (the STI subclass declared by `class_name: "Category"`), regardless of which sibling associations on the same `:through` table are preloaded alongside it.

### Actual behavior

When `:categories` is preloaded in the same call as the parent STI class's `has_many :through` association (`:tags`), `post.categories` containing `Keyword` (or any other sibling STI subclass) records.

This behavior does **not** occur for `preload(:keywords, :categories)` (two STI **subclass** siblings, neither of them the base class).

### System configuration

Reproduced on:
  - Rails 7.2.3.1
  - Rails 8.0.5
  - Rails 8.1.3

**Ruby version**: ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [x86_64-linux]

ruyrocha:
when a base-class association is preloaded alongside a subclass association.

Fixes #57387

### Motivation / Background

When preloading a base-class `has_many :through` association (e.g. `:categories`)
alongside an STI subclass `has_many :through` association (e.g. `:special_categories`)
on the same through table, the subclass association's target would be incorrectly
populated with records of sibling STI types.

### Detail

`ThroughAssociation#records_by_owner` builds its result by flat-mapping
`source_records_by_owner` over the through records. Because `source_records_by_owner`
is seeded from the through join model's `belongs_to :source` reflection — which points
to the base class — the SQL query it issues has no STI `WHERE type = ?` filter.
When the base-class association has already been preloaded (and thus the through
records are already cached), all source records of all STI sibling types end up in
`source_records_by_owner`, and the subclass association gets all of them.

The fix filters `records` by `reflection.klass` after the flat-map when the
association targets an STI subclass (`reflection.klass != reflection.klass.base_class`).
`reflection.klass` is `SpecialCategory` (from `class_name: "SpecialCategory"`), so
`is_a?(reflection.klass)` correctly rejects sibling types without affecting
non-STI associations.

### Additional information

The bug is reproducible with any preload order that includes both a base-class
through association and an STI subclass through association sharing the same
join model. Standalone loads and single-association preloads are unaffected because
they hit the correct filtered SQL path directly.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix `has_many :through` preload leaking sibling STI records into subclass associations.
+
+    Preloading a base-class through association (e.g. `:categories`) alongside an STI
+    subclass through association (e.g. `:special_categories`) would cause the subclass
+    association to contain records of sibling STI types. `ThroughAssociation#records_by_owner`
+    now filters records by the reflection's class when the association targets an STI subclass.
+
+    *Ruy Rocha*
+
 *   Fix `ActiveRecord::MessagePack` serialization raising `NoMethodError`
     for any record with a populated `time` column, which made such records
     uncacheable through the MessagePack cache serializer.
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Fix `has_many :through` preload leaking sibling STI records into subclass associations.
+
+    Preloading a base-class through association (e.g. `:categories`) alongside an STI
+    subclass through association (e.g. `:special_categories`) would cause the subclass
+    association to contain records of sibling STI types. `ThroughAssociation#records_by_owner`
+    now filters records by the reflection's class when the association targets an STI subclass.
+
+    *Ruy Rocha*
+
 *   Apply `all_queries: true` default scopes consistently across counter caches,
     uniqueness validations, fixture lookups, and record reloads, while those
     operations continue to bypass ordinary default scopes.
--- a/activerecord/lib/active_record/associations/preloader/through_association.rb
+++ b/activerecord/lib/active_record/associations/preloader/through_association.rb
@@ -29,6 +29,10 @@ module ActiveRecord
               source_records_by_owner[record]
             end
 
+            if reflection.klass != reflection.klass.base_class
+              records = records.select { |r| r.is_a?(reflection.klass) }
+            end
+
             records.compact!
             records.sort_by! { |rhs| preload_index[rhs] } if scope.order_values.any?
             records.uniq! if scope.distinct_value
--- a/activerecord/test/cases/associations/has_many_through_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_through_associations_test.rb
@@ -1738,6 +1738,20 @@ class HasManyThroughAssociationsTest < ActiveRecord::TestCase
     assert_equal [[comment.blog_id, comment.id]], ids
   end
 
+  def test_preload_sti_subclass_association_alongside_base_class_association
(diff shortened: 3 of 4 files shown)


ruyrocha:
@rafaelfranca do you have any thoughts on this? Thank you!

PR #57409, opened 2026-05-21
Use `connection.reset!` instead of `reset_connection` to prevent connection leak in PG test teardowns

yahonda:
### Motivation / Background

The Rails nightly suite `activerecord postgresql` job fails with `FATAL: sorry, too many clients already`, e.g.

Each affected test's teardown ends with `reset_connection`, defined in `activerecord/test/support/connection_helper.rb` as:

```ruby
def reset_connection
  original_connection = ActiveRecord::Base.remove_connection
  ActiveRecord::Base.establish_connection(original_connection)
end
```

That is, it swaps the whole pool. With transactional fixtures the fixture transaction still holds the existing connection, so `ConnectionPool#disconnect!` silently times out trying to acquire it exclusively and the PG socket is leaked. Across the suite (especially `PostgresqlRangeTest`, which has 46 tests) this is enough to exhaust `max_connections=100` on the PG server and cause the `too many clients already` failure.

### Detail

Replacing `reset_connection` with `@connection.reset!` keeps the existing connection but issues `DISCARD ALL` on PostgreSQL, which clears the same cached plans / session state without disconnecting, so no pool swap and no leak.

The `reset_connection` helper itself is intentionally left untouched because `PostgreSQLReferentialIntegrityTest` relies on it for a different purpose — getting a fresh adapter instance after monkey-patching `@connection` with `MissingSuperuserPrivileges`. That use case genuinely needs the pool swap; `reset!` would not unextend the module. So `reset_connection` has two distinct usages today:

| Use | Required action |
| --- | --- |
| query plan cache clear (enum, range, composite, domain) | `reset!` is enough |
| fresh adapter instance (referential_integrity) | pool swap is required |

Renaming `reset_connection` to reflect the "fresh adapter" semantics it provides today is a worthwhile follow-up but out of scope here.

### Per-test-class measurements

Captured with a diagnostic probe (see the branch [`yahonda/pg-leak-diagnostic-probe`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-diagnostic-probe) for reproduction) running each test file in isolation against `postgres:alpine` (`max_connections=100`), TCP, YJIT off:

| Test file       | runs | BEFORE: pg peak / in-test +1 events | AFTER: pg peak / in-test +1 events |
| --------------- | ---- | ----------------------------------- | ---------------------------------- |
| enum_test       |   20 |                              42 / 20 |                              2 / 0 |
| range_test      |   46 |                               48 / 0 |                              2 / 0 |
| composite_test  |    3 |                                5 / 0 |                              2 / 0 |
| domain_test     |    2 |                                3 / 0 |                              2 / 0 |

`pg peak` is the maximum number of PG client backends held by the test process at any point during the file's run (queried via `pg_stat_activity`, baseline ≈ 2 fixture-loaded sessions). `+1 events` are individual tests whose teardown was credited a direct +1 PG-side delta; `range_test`'s leak is entirely between-tests (every per-test teardown's `reset_connection` adds a connection to the next test's baseline).

Across the full PG suite, this fix drops total PG peak from ~94/100 to ~46/100 (combined with separate fixes for LoadAsync and standalone adapter tests, which are not in this PR).

### Additional information

### Checklist

- [x] Tests are updated.

### DIFF
--- a/activerecord/test/cases/adapters/postgresql/composite_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/composite_test.rb
@@ -33,7 +33,7 @@ module PostgresqlCompositeBehavior
 
     @connection.drop_table "postgresql_composites", if_exists: true
     @connection.execute "DROP TYPE IF EXISTS full_address"
-    reset_connection
+    @connection.reset!
     PostgresqlComposite.reset_column_information
   end
 end
--- a/activerecord/test/cases/adapters/postgresql/domain_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/domain_test.rb
@@ -23,7 +23,7 @@ class PostgresqlDomainTest < ActiveRecord::PostgreSQLTestCase
   teardown do
     @connection.drop_table "postgresql_domains", if_exists: true
     @connection.execute "DROP DOMAIN IF EXISTS custom_money"
-    reset_connection
+    @connection.reset!
   end
 
   def test_column
--- a/activerecord/test/cases/adapters/postgresql/enum_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/enum_test.rb
@@ -30,10 +30,10 @@ class PostgresqlEnumTest < ActiveRecord::PostgreSQLTestCase
   end
 
   teardown do
-    reset_connection
+    @connection.reset!
     @connection.drop_table "postgresql_enums", if_exists: true
     @connection.drop_enum "mood", if_exists: true
-    reset_connection
+    @connection.reset!
   end
 
   def test_column
@@ -191,7 +191,7 @@ class PostgresqlEnumTest < ActiveRecord::PostgreSQLTestCase
     $stdout = original_stdout
     @connection.drop_table "pg_enum_array_schema_dump_accounts", if_exists: true
     @connection.drop_enum "pg_enum_array_schema_dump_account_type", if_exists: true
-    reset_connection
+    @connection.reset!
   end
 
   def test_drop_enum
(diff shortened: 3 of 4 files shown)


yahonda:
Closing this PR.

The BEFORE/AFTER numbers in the description were inflated by a binding leak in the diagnostic probe on [`yahonda/pg-leak-diagnostic-probe`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-diagnostic-probe). The finalizer lambda was defined inline inside `PgConnectionAllocTrace.record(conn, stack)`:

```ruby
def record(conn, stack)
  oid = conn.object_id
  ObjectSpace.define_finalizer(conn, ->(_id) { ALLOCATIONS.delete(oid) })
end
```

In MRI, the lambda's closure binding includes every local of the enclosing method — `conn`, `stack`, `addr`, `oid` — even though the body only references `oid`. The finalizer table is a GC root, so each registered `PG::Connection` was pinned via the binding. The C-side socket therefore never closed and the `pg_stat_activity` session for it stayed alive, inflating the `pg_server peak` numbers this PR cited as motivation.

Re-measured against a corrected probe (lambda built in a separate method so its binding does not capture `conn`), against `postgresql@18` locally with `max_connections=100`, each file run in isolation:

| Test file | reported BEFORE | corrected BEFORE | corrected AFTER |
| --- | --- | --- | --- |
| range_test    | 48 / +1×0  | 2 / +1×0  | 2 / +1×0 |
| enum_test     | 42 / +1×20 | 4 / +1×19 | 2 / +1×0 |
| composite_test | 5 / +1×0  | 2 / +1×0  | 2 / +1×0 |
| domain_test   | 3 / +1×0   | 2 / +1×0  | 2 / +1×0 |

Only `enum_test` shows a measurable PG-server-side effect from the `reset_connection` → `connection.reset!` change. `range_test`, `composite_test`, and `domain_test` show no peak difference before/after — their "leaks" were 100% probe artifact.

Closing rather than rewriting: the original motivation (reducing the per-file PG client backend count that contributes to the nightly `too many clients already` failure) is not supported by the corrected data, and the nightly failure itself needs to be re-investigated against a probe that does not pin its own subjects. If `connection.reset!` is desirable on style grounds (lighter than a full pool swap when only DISCARD ALL is required), it could be re-pitched separately without the leak framing.

PR #57410, opened 2026-05-21
Disconnect test-local PostgreSQL adapters to prevent connection leak

yahonda:
### Motivation / Background

This addresses the same `FATAL: sorry, too many clients already` failure in the Rails nightly suite as #57409:

Many tests in `PostgreSQLAdapterTest` construct a one-off adapter via `PostgreSQLAdapter.new(...)` (directly, or via the `connection_without_insert_returning` test helper) and let the local `connection` go out of scope at the end of the test body. The underlying PG socket is then only closed by the pg gem's C-level finalizer, which runs at indeterminate GC time. Across the suite, those sockets stack up against the `max_connections=100` ceiling on the server.

### Detail

Introduce a `with_postgresql_adapter(config_hash) { |connection| ... }` private helper that builds the adapter, yields it, and disconnects it in an `ensure` block. Convert the affected tests to use the helper so the PG socket is closed deterministically as soon as the test exits the block. `connection_without_insert_returning` is refactored to take a block and delegate to the new helper.

This change is independent of #57409 (which converts `reset_connection` to `@connection.reset!` in 4 different PG test files) and can be merged in either order.

### Per-test-file measurements

Captured with a diagnostic probe (see the topic branch [`yahonda/pg-leak-diagnostic-probe`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-diagnostic-probe) for reproduction) running `postgresql_adapter_test.rb` in isolation against `postgres:alpine` (`max_connections=100`), TCP, YJIT off:

| Metric                       | BEFORE | AFTER |
| ---------------------------- | ------ | ----- |
| runs                         |    107 |   107 |
| pg_server peak               |     23 |     7 |
| per-test +1 PG-side events   |     29 |    14 |

`pg_server peak` is the maximum number of PG client backends held by the test process at any point during the file's run (queried via `pg_stat_activity`, baseline ≈ 2 fixture-loaded sessions). `+1 events` are individual tests whose teardown was credited a direct +1 PG-side delta.

The fix removes the 15 leak events attributable to local `PostgreSQLAdapter.new(...)` adapters that the new helper closes deterministically. The remaining 14 events come from tests that touch `@connection` (the pool-managed connection) in a way that triggers `verify!`/`reconnect!`, which is out of scope for this PR.

### Additional information

### Checklist

- [x] Tests are updated.

### DIFF
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -99,42 +99,44 @@ module ActiveRecord
 
       def test_reconnect_after_bad_connection_on_check_version_with_0_return
         db_config = ActiveRecord::Base.configurations.configs_for(env_name: "arunit", name: "primary")
-        connection = ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.new(db_config.configuration_hash.merge(connection_retries: 0))
-        connection.connect!
+        with_postgresql_adapter(db_config.configuration_hash.merge(connection_retries: 0)) do |connection|
+          connection.connect!
 
-        # mimic a connection that hasn't checked and cached the server version yet i.e. without a raw_connection
-        connection.pool.instance_variable_set(:@server_version, nil)
-        connection.raw_connection.stub(:server_version, 0) do
-          error = assert_raises ActiveRecord::ConnectionNotEstablished do
+          # mimic a connection that hasn't checked and cached the server version yet i.e. without a raw_connection
+          connection.pool.instance_variable_set(:@server_version, nil)
+          connection.raw_connection.stub(:server_version, 0) do
+            error = assert_raises ActiveRecord::ConnectionNotEstablished do
+              connection.reconnect!
+            end
+            assert_equal "Could not determine PostgreSQL version", error.message
+          end
+
+          # can reconnect after a bad connection
+          assert_nothing_raised do
             connection.reconnect!
           end
-          assert_equal "Could not determine PostgreSQL version", error.message
-        end
-
-        # can reconnect after a bad connection
-        assert_nothing_raised do
-          connection.reconnect!
         end
       end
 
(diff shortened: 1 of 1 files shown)


yahonda:
Hiding whitespace makes diff clear. https://github.com/rails/rails/pull/57410/changes?w=1

yahonda:
Closing this PR.

The BEFORE/AFTER numbers in the description were inflated by a binding leak in the diagnostic probe on [`yahonda/pg-leak-diagnostic-probe`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-diagnostic-probe) — see #57409 for the mechanism (a finalizer lambda's enclosing-method binding pinned each `PG::Connection` via the finalizer table, so sockets never closed and `pg_stat_activity` sessions never went away).

Re-measured against a corrected probe (lambda built in a separate method so its binding does not capture `conn`), against `postgresql@18` locally with `max_connections=100`, `postgresql_adapter_test.rb` in isolation:

|  | reported BEFORE | corrected BEFORE | corrected AFTER |
| --- | --- | --- | --- |
| pg_server peak       | 23 | 3  | 3  |
| per-test +1 events   | 29 | 16 | 15 |

The peak is unchanged before/after the fix, and the per-test `+1` event count drops by 1 — essentially within noise. The `with_postgresql_adapter { … }` helper is still arguably a cleaner pattern for tests that build a one-off adapter (explicit `disconnect!` on `ensure`), but the leak-reduction motivation does not survive the re-measurement. Closing rather than re-pitching as a style-only change; a successor PR can be framed appropriately if there is appetite.

The "Floating PG::Connection cycle" analysis in the description — `set_notice_receiver` block at [postgresql_adapter.rb:1062](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L1062) capturing `self` and creating an adapter ↔ raw_connection ↔ Proc cycle — does describe a real Ruby-side retention pattern, but with the probe fixed those cycles are broken cleanly at GC time and do not survive into `pg_stat_activity` as the BEFORE numbers claimed.

PR #57422, opened 2026-05-21
Fixes to query command

### LINKED ISSUE #57419: New rails query command returns useless responses outside of recordsets
https://github.com/rails/rails/pull/57156 creates an amazing new `rails query` command. But I don't know if this is an intentional gap, but doing this with a single record doesn't return a useful output?

```
$ rails query "User.find(1)"
{"columns":["result"],"rows":[["#<User:0x00007fb9de746518>"]],"meta":{"row_count":1,"query_time_ms":38.5,"page":1,"per_page":100,"has_more":false,"sql":"User.find(1)"}}
```

The fix appears to be in `execute_ar` add another case of:

```
when ActiveRecord::Base
    attrs = result.attributes
    { columns: attrs.keys, rows: [attrs.values], sql: expression, truncated: false }
```

Or something similar. Feels like LLMs will often try to get a singular record, if it's being talked about?

And apparently (so my LLM just told me) the `Array` branch is open to the same issue (although I haven't needed that yet).

```
$ rails query "User.first(2)"
```

Will return those short inspect values.

### LINKED ISSUE #57420: Deletion bypassing read-only in new rails query command
There is a bug on line 159 in `execute_ar` in the new `rails query` (https://github.com/rails/rails/pull/57156) command that will allow LLMs to bypass the readonly part and execute potentially dangerous/destructive queries:

```
  def execute_ar(expression:, page:, per:)
    result = eval(expression, TOPLEVEL_BINDING, "(query)", 1)
    # ...
```
For an AR expression (no `--sql`), the eval runs before any read-only guard is set up. The `with_readonly_connection_for` block only wraps:

  - Raw SQL paths (line 61, `with_readonly_connection` around `execute_sql`)
  - The subsequent `connection.select_all(sql)` call when the `eval` happens to return an `ActiveRecord::Relation` (lines 165-169)
  - `EXPLAIN` paths (lines 105, 112)

My LLM decided to run `User.delete_all` directly against production:

  1. `eval("User.delete_all")` executed directly against the default connection — no guard
  2. AR fired the DELETE FROM users statement (which then hit the FK constraint and aborted)
  3. The integer return value would have flowed to the else branch on line 179

The same hole exists for any AR expression that itself performs writes - `User.update_all(...)`, `user.save!`, `user.destroy`, `User.create!(...)`, etc. The guard is only applied to relations that are converted to SQL and run via `connection.select_all`, not to whatever side effects the eval'd Ruby triggers.

The proposed fix from my Claude would be to wrap the eval itself, e.g.:

```
  def execute_ar(expression:, page:, per:)
    with_readonly_connection do |_|
      result = eval(expression, TOPLEVEL_BINDING, "(query)", 1)
      # ... existing case statement
    end
  end
```

Caveats with that fix:
 
  - `with_readonly_connection` currently keys off `ActiveRecord::Base`, so multi-DB models on a different connection class wouldn't be protected. For full coverage you'd need a two-pass approach (`eval` inside `ActiveRecord::Base.while_preventing_writes`, then re-resolve onto the right connection class for relation execution) - or just blanket every connection class.
  - The reading-role path (`connected_to(role: :reading)`) would force writes to fail because the replica is read-only at the MySQL level too, which is the desired behavior.

andyjeffries:
### Motivation / Background

This Pull Request has been created because I want to be able to use the new `rails query` with my LLM, but as it stands there is one real deal breaker of an aspect of usage that is broken #57419, and one DANGEROUS bypassing of the database connection being readonly #57420.

Fixes #57420.
Fixes #57419.

### Detail

This Pull Request implements the fixes I proposed in those two issues.

### Checklist

* [ ] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs. - unknown if these are really considered separate issues, if so, feel free to tell me and I'll split the two commits out to separate PRs.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included. There's no behaviour change or additional feature - just bug fixes, even though not minor. I don't know if it should be included or not.

### DIFF
--- a/railties/lib/rails/commands/query/query_command.rb
+++ b/railties/lib/rails/commands/query/query_command.rb
@@ -99,7 +99,7 @@ module Rails
               say format_result(columns: result.columns, rows: result.rows, sql: "EXPLAIN #{expression}")
             end
           else
-            relation = eval(expression, TOPLEVEL_BINDING, "(query)", 1)
+            relation = with_readonly_connection { eval(expression, TOPLEVEL_BINDING, "(query)", 1) }
             sql = relation.to_sql
             with_readonly_connection_for(relation.model.connection_class_for_self) do |connection|
               result = connection.select_all("EXPLAIN #{sql}")
@@ -148,7 +148,7 @@ module Rails
         end
 
         def execute_ar(expression:, page:, per:)
-          result = eval(expression, TOPLEVEL_BINDING, "(query)", 1)
+          result = with_readonly_connection { eval(expression, TOPLEVEL_BINDING, "(query)", 1) }
 
           case result
           when ActiveRecord::Relation
@@ -161,13 +161,22 @@ module Rails
             end
           when ActiveRecord::Result
             { columns: result.columns, rows: result.rows, sql: expression, truncated: false }
+          when ActiveRecord::Base
+            attrs = result.attributes
+            { columns: attrs.keys, rows: [ attrs.values ], sql: expression, truncated: false }
           when Hash
             rows = result.map { |key, val| [ key, val ] }
             { columns: [ "key", "value" ], rows: rows, sql: expression, truncated: false }
           when Array
-            rows = result.map { |val| Array(val) }
-            cols = Array.new(rows.first&.length.to_i) { |i| "column_#{i}" }
-            { columns: cols, rows: rows, sql: expression, truncated: false }
+            if result.first.is_a?(ActiveRecord::Base)
+              columns = result.first.attributes.keys
+              rows = result.map { |record| record.attributes.values }
+              { columns: columns, rows: rows, sql: expression, truncated: false }
+            else
+              rows = result.map { |val| Array(val) }
+              cols = Array.new(rows.first&.length.to_i) { |i| "column_#{i}" }
+              { columns: cols, rows: rows, sql: expression, truncated: false }
+            end
           else
             { columns: [ "result" ], rows: [ [ result ] ], sql: expression, truncated: false }
           end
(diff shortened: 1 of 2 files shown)


lewispb:
Thanks @andyjeffries - looks good.

railties/lib/rails/commands/query/query_command.rb
```diff
@@ -161,13 +161,22 @@ def execute_ar(expression:, page:, per:)
...
+            { columns: attrs.keys, rows: [ attrs.values ], sql: expression, truncated: false }
           when Hash
             rows = result.map { |key, val| [ key, val ] }
             { columns: [ "key", "value" ], rows: rows, sql: expression, truncated: false }
           when Array
-            rows = result.map { |val| Array(val) }
-            cols = Array.new(rows.first&.length.to_i) { |i| "column_#{i}" }
-            { columns: cols, rows: rows, sql: expression, truncated: false }
+            if result.first.is_a?(ActiveRecord::Base)
+              columns = result.first.attributes.keys
```
fatkodima:
```suggestion
              columns = result.first.attribute_names
```

railties/lib/rails/commands/query/query_command.rb
```diff
@@ -161,13 +161,22 @@ def execute_ar(expression:, page:, per:)
...
+              rows = result.map { |record| record.attributes.values_at(*columns) }
```
fatkodima:
```suggestion
              rows = result.map { |record| record.attributes }
```

fatkodima (commented):
Please squash commits into one.

andyjeffries:
Amended in line with the feedback above @fatkodima and squashed to a single commit.

PR #57427, opened 2026-05-21
Fix #57401: Stop DRb service when shutting down parallel test workers

### LINKED ISSUE #57401: ActiveSupport::Testing::Parallelization#shutdown doesn't call DRb.stop_service — DRb thread leaks in parent
### Steps to reproduce

`ActiveSupport::Testing::Parallelization#initialize` starts a DRb service:

```ruby
def initialize(worker_count, work_stealing: false)
  ...
  @url = DRb.start_service("drbunix:", @queue_server).uri
end
```

…but `#shutdown` never calls `DRb.stop_service`:

```ruby
def shutdown
  ...
  @queue_server.shutdown        # shuts the queue Server object
  @worker_pool.each { Process.waitpid(_1) ... }   # waits for workers
end
```

The DRb `accept` loop thread (owned by the `DRb` module) keeps running. In a process that uses `parallelize` for a single test run and then expects to exit naturally, the leaked thread prevents process termination.

Bug report template — fails on `rails:main` (verified at `rails-c3c65d10b0bb`):

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
end

require "active_support"
require "minitest/autorun"
require "drb"
require "drb/unix"

class ParallelizationDRbLeakTest < Minitest::Test
  def test_drb_thread_is_cleaned_up_after_parallelization_shutdown
    parallelization.start
    parallelization.shutdown

    leaked = Thread.list
                   .reject { |t| t == Thread.main }
                   .select(&:alive?)
                   .select do |t|
      bt = Array(t.backtrace).join("\n")
      bt.include?("drb/drb.rb") && bt.match?(/accept_or_shutdown|main_loop/)
    end

    assert_empty leaked,
      "Expected Parallelization#shutdown to call DRb.stop_service. Leaked:\n" +
      leaked.map { |t| Array(t.backtrace).first(5).join("\n  ") }.join("\n---\n")
  end
end
```

### Expected behavior

After `Parallelization#shutdown` returns, the DRb service it started in `#initialize` is stopped and its `accept` loop thread has terminated.

### Actual behavior

The DRb accept loop thread is still alive after `#shutdown`:

```
Failure: ParallelizationDRbLeakTest#test_drb_thread_is_cleaned_up_after_parallelization_shutdown
Expected Parallelization#shutdown to call DRb.stop_service. Leaked:
  drb-2.2.3/lib/drb/drb.rb:1015:in `select'
  drb-2.2.3/lib/drb/drb.rb:1015:in `accept_or_shutdown'
  drb-2.2.3/lib/drb/unix.rb:106:in `accept'
  drb-2.2.3/lib/drb/drb.rb:1765:in `main_loop'
  drb-2.2.3/lib/drb/drb.rb:1586:in `block in run'
Expected [#<Thread:0x... sleep>] to be empty.
```

Because this is a non-daemon thread blocked in `select` on a Unix socket, processes that run `parallelize` and then expect to exit naturally hang at exit. We hit this in a Rails app's CI: minitest's summary prints `0 failures`, then the process never terminates and the CI container is killed by a timeout.

Workers don't have this leak because `Parallelization::Worker#start` explicitly calls `DRb.stop_service` after forking. Only the parent is affected.

Proposed fix (small): call `DRb.stop_service` at the end of `Parallelization#shutdown` (after `Process.waitpid` so we don't tear down DRb while it's still in use). I'm happy to open a PR with a regression test once we agree on the approach — wanted to file an issue first to confirm this is the right framing.

### System config

* Rails version: `main` (verified at `rails-c3c65d10b0bb`); also reproduced on `8-0-stable` (8.0.4.1) and `8-1-stable` (8.1.3).
* Ruby version: 3.3.10 (arm64-darwin25), also Linux glibc x86_64 in CI.
* drb gem version: 2.2.3.

Affects both `8-0-stable` and `8-1-stable` (and presumably older branches). Backport candidates.

mugitti9:
### Motivation / Background

Fixes #57401

When running tests in parallel with processes (`parallelize` / `PARALLEL_WORKERS`), the parent process starts a DRb service in `Parallelization#initialize` via `DRb.start_service`. However, `#shutdown` only shut down the queue server and waited for workers; it never called `DRb.stop_service`.

As a result, the parent kept a DRb accept-loop thread alive after workers exited. In some environments, this prevented the test process from exiting, for example CI timing out after printing `0 failures`. Worker processes already call `DRb.stop_service` after fork, so only the parent was missing cleanup.

### Detail

- Call `DRb.stop_service` at the end of `ActiveSupport::Testing::Parallelization#shutdown`, after all worker processes have been reaped with `Process.waitpid`
- Add a regression test that asserts no DRb accept threads remain after `shutdown`

### Additional information

I verified the following locally:

- With the fix: `bin/test test/parallelization_test.rb -i "/shutdown stops the DRb/"` passes
- Without `DRb.stop_service`, the same test fails with a leaked DRb accept thread, verified locally

### Checklist

### DIFF
--- a/activesupport/lib/active_support/testing/parallelization.rb
+++ b/activesupport/lib/active_support/testing/parallelization.rb
@@ -80,6 +80,8 @@ module ActiveSupport
           nil
         end
 
+        DRb.stop_service
+
         Parallelization.run_cleanup_hooks.each(&:call)
       end
     end
--- a/activesupport/test/parallelization_test.rb
+++ b/activesupport/test/parallelization_test.rb
@@ -94,6 +94,26 @@ class ParallelizationTest < ActiveSupport::TestCase
     assert_not server.active_workers?
   end
 
+  test "shutdown stops the DRb service started in initialize" do
+    skip "Process-based parallelization requires fork" unless Process.respond_to?(:fork)
+
+    parallelization = ActiveSupport::Testing::Parallelization.new(1)
+    parallelization.start
+    parallelization.shutdown
+
+    leaked = Thread.list
+      .reject { |t| t == Thread.main }
+      .select(&:alive?)
+      .select do |t|
+        bt = Array(t.backtrace).join("\n")
+        bt.include?("drb/drb.rb") && bt.match?(/accept_or_shutdown|main_loop/)
+      end
+
+    assert_empty leaked,
+      "Expected Parallelization#shutdown to call DRb.stop_service. Leaked:\n" +
+      leaked.map { |t| Array(t.backtrace).first(5).join("\n  ") }.join("\n---\n")
+  end
+
   test "seeded distribution assigns tests to workers round-robin" do
     worker_count = 2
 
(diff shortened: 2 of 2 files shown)


rafaelfranca:
Tests are still broken. Isn't because they run their org Drb server so you will still see a alive Drb server?

PR #57428, opened 2026-05-21
Fix check_all_foreign_keys_valid! corrupts NOT VALID state on same-named FKs + same table name across schemas

tahsin352:
### Motivation / Background
Follow up to https://github.com/rails/rails/pull/53636

This Pull Request has been created to fix `regclass` cast to be schema-qualified.

### Detail

The `UPDATE pg_constraint SET convalidated=false` inside `check_all_foreign_keys_valid!` matches by `conrelid::regclass = '%3$I'::regclass`, but `'orders'::regclass` resolves using `search_path` — so when two schemas each contain a table named `orders`, both their FK constraints get reset to `NOT VALID`, but only the one being processed gets re-validated. The other schema's FK is permanently left `NOT VALID`.

When two schemas each contain a table with the same name and a FK constraint with the same name, the unqualified `conrelid::regclass` cast in the `UPDATE` resolved via `search_path` and could reset `convalidated=false` on the *wrong* schema's constraint, leaving it permanently `NOT VALID`. After `check_all_foreign_keys_valid!` both must be `convalidated = true`. Before the fix, the `search_path`-ambiguous `UPDATE reset convalidated=false` on whichever schema came *second* in the loop, leaving it permanently invalid.

One token changes in the FORMAT string: replace the bare `'%3$I'::regclass` cast with a schema-qualified `'%2$I.%3$I'::regclass` cast. We have added a test case also.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb
@@ -90,7 +90,7 @@ module ActiveRecord
             BEGIN
             FOR r IN (
               SELECT FORMAT(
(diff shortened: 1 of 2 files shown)


fatkodima:
CI is red.

> so when two schemas each contain a table named `orders`

But we scope the `UPDATE` query with a schema too. What am I missing?

PR #57431, opened 2026-05-22
fix: validate subcommand in `rails plugin` command

### LINKED ISSUE #57430: "rails plugin" ignores invalid subcommands
### Steps to reproduce

Run:

```sh
rails plugin foo bar
```

### Expected behavior

"foo" is not a valid subcommand, so Rails should print an error message or a usage message.

### Actual behavior

Rails creates a plugin named "bar" (as if I ran `rails plugin new bar`).

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 4.0.2

ruyrocha:
`rails plugin foo bar` silently ignored the invalid subcommand "foo" and proceeded to create a plugin named "bar" as if `rails plugin new bar` had been run. Now it prints an error and exits with status 1.

Fixes #57430

### Motivation / Background

Running `rails plugin` with an invalid subcommand (anything other than `new`) silently discarded the invalid subcommand and proceeded as if `rails plugin new` had been run. This is confusing and error-prone, as
typos or misunderstandings of the available subcommands go unnoticed.

### Detail

`PluginCommand#perform` now validates the subcommand argument before proceeding. If the subcommand is anything other than `"new"`, it prints an error message and exits with status 1.

### Checklist

### DIFF
--- a/railties/lib/rails/commands/plugin/plugin_command.rb
+++ b/railties/lib/rails/commands/plugin/plugin_command.rb
@@ -21,7 +21,16 @@ module Rails
       class_option :no_rc, desc: "Skip evaluating .railsrc."
 
       def perform(type = nil, *plugin_args)
-        plugin_args << "--help" unless type == "new"
+        if type
+          unless type == "new"
+            error "'#{type}' is not a valid plugin subcommand. Valid subcommand: new."
+            error "Run '#{self.class.executable} -h' for help."
+            exit 1
+          end
+        else
+          plugin_args << "--help"
+        end
+
 
         unless options.key?("no_rc") # Thor's not so indifferent access hash.
           railsrc = File.expand_path(options[:rc])
--- a/railties/CHANGELOG.md
+++ b/railties/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Validate subcommand in `rails plugin` command.
+
+    `rails plugin foo bar` silently ignored the invalid subcommand "foo"
+    and proceeded to create a plugin named "bar". Now it prints an error
+    and exits with status 1.
+
+    Fixes #57430.
+
+    *Ruy Rocha*
+
 *   Prevent the internal development welcome route from being duplicated on route reloads.
 
     *Elliot Temple*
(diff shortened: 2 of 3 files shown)


railties/lib/rails/commands/plugin/plugin_command.rb
```diff
@@ -21,6 +21,12 @@ def self.banner(*) # :nodoc:
       class_option :no_rc, desc: "Skip evaluating .railsrc."
 
       def perform(type = nil, *plugin_args)
+        if type && type != "new"
```
rafaelfranca: I'm pretty sure Thor is able to register only the command we want to exist. Did you try to use the features in Thor?

railties/lib/rails/commands/plugin/plugin_command.rb
```diff
@@ -21,6 +21,12 @@ def self.banner(*) # :nodoc:
...
+          error "'#{type}' is not a valid plugin subcommand. Valid subcommand: new."
+          error "Run '#{self.class.executable} -h' for help."
+          exit 1
+        end
+
         plugin_args << "--help" unless type == "new"
```
rafaelfranca: Also, isn't the goal of this line to show help when the type isn't "new"?

railties/lib/rails/commands/plugin/plugin_command.rb
ruyrocha: Sure, sure, @rafaelfranca! Now it rejects anything else other than `new`. I appreciate your inputs here :)

railties/lib/rails/commands/plugin/plugin_command.rb
ruyrocha: Absolutely! It displays the help message when `type` is `nil`

railties/lib/rails/commands/plugin/plugin_command.rb
ruyrocha: CI wasn't happy at all - https://buildkite.com/rails/rails/builds/128686/table?jid=019e4eb3-638b-435d-b822-daf40a9e2b29&tab=output

So I just reverted the changes

rafaelfranca (approved):

PR #57432, opened 2026-05-22
Rename `reset_connection` to `reset_pool`; have `reset_connection` reset just the existing connection

yahonda:
### Motivation

The nightly `activerecord postgresql` job fails with `FATAL: sorry, too many clients already` (e.g. [build #128640](https://buildkite.com/rails/rails/builds/128640#019e4c87-ef26-4267-8987-f7440998fd9c)). This PR addresses one of three test-side leak sources identified during the investigation.

Earlier attempt: #57409 (closed; the BEFORE/AFTER numbers there were inflated by a binding leak in the diagnostic probe — see [`yahonda/pg-leak-repro`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-repro) for the corrected methodology).

### Detail

`PostgresqlRangeTest`, `PostgresqlEnumTest`, `PostgresqlCompositeTest`, and `PostgresqlDomainTest` call `reset_connection` in their teardown to drop cached query plans. The helper was implemented as `remove_connection` + `establish_connection`, swapping the whole pool. With transactional fixtures the fixture transaction still holds the existing connection, so `ConnectionPool#disconnect!` silently times out and the PG socket is leaked.

These four files only need to clear cached plans / session state — not swap the pool. The fix changes `reset_connection`'s implementation to `@connection.reset!` (`DISCARD ALL` on the existing connection), and renames the previous pool-swap implementation to `reset_pool` for the callers that genuinely need a fresh adapter (e.g. `PostgreSQLReferentialIntegrityTest`, which extends a module onto the adapter and needs that module unextended).

Note: the four test files mentioned above are the origin of the leak but **do not appear in this PR's diff** — they keep calling `reset_connection` unchanged (vs `main`), and silently benefit from the new helper implementation. The diff only shows the helper itself plus the callers that migrate to `reset_pool`.

### Measurement

`postgres:alpine` (PG 18, `max_connections=100`), running only the four files that originally leaked, `--seed=22021` (worst case for the baseline), N=5, container restart per run:

| | peak (5 runs) |
| --- | --- |
| BEFORE | 9, 12, 7, 7, 8 |
| AFTER | 2, 2, 2, 2, 2 (= fixture baseline) |

Methodology: [`yahonda/pg-leak-repro`](https://github.com/yahonda/rails/tree/yahonda/pg-leak-repro).

### Checklist

- [x] Tests are updated.

### DIFF
--- a/activerecord/test/cases/adapters/abstract_mysql_adapter/active_schema_test.rb
+++ b/activerecord/test/cases/adapters/abstract_mysql_adapter/active_schema_test.rb
@@ -36,7 +36,7 @@ class ActiveSchemaTest < ActiveRecord::AbstractMysqlTestCase
   end
 
   teardown do
-    reset_connection
+    reset_pool
   end
 
   def test_add_index
--- a/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/activerecord/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -323,7 +323,7 @@ module ActiveRecord
       end
 
       def test_queries_executed_on_fresh_connection_through_first_select
-        reset_connection
+        reset_pool
 
         queries = PostgreSQLAdapter.with(decode_dates: false, decode_money: false, decode_bytea: false) do
           with_timezone_config(default: :utc) do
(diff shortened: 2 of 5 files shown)


rafaelfranca (commented):
What do you think on swapping the implementation of `reset_connection` to be what you are doing now and create a new method `reset_pool` for the current implementation? What you are doing now is what I expect ```
reset_connection` to be
```

yahonda:
Agreed. I have updated `reset_connection` method just calls `@connection.reset!` and created a new `reset_pool` method for the current reset_connection method implementation.

yahonda:
Let's see how 2nd commit is going, Once the CI passes, I'll squash commits.

yahonda:
Commits are squashed.

rafaelfranca (approved):

PR #57435, opened 2026-05-22
Add missing docs for Rails.autoloaders

fxn:
The interface available from `Rails.autoloaders` was documented in the [autoloading guide](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html). This patch adds missing documentation for the API docs.


rafaelfranca:
Thank you!

PR #57452, opened 2026-05-23
Handle malformed Mandrill events payloads in Action Mailbox

Messier81:
### Motivation / Background

The Mandrill ingress returns `422 Unprocessable Content` for invalid JSON in `mandrill_events`, but valid JSON of the wrong shape (`null`, a scalar, a hash, or an array containing non-objects) makes `.select` raise `NoMethodError` (or `TypeError` for non-object elements). That's a 500 instead of a 422.

### Detail

Validate that the parsed payload is an Array of Hashes before iterating. A malformed payload now raises `MalformedEventsError`, which the existing `rescue` handles alongside `JSON::ParserError`, returning `422 Unprocessable Content` without creating inbound emails.

### Additional information

Before the fix, a malformed-shape payload raised `NoMethodError: private method 'select' called for nil:NilClass` from `raw_emails`.

Validation:

```bash
cd actionmailbox && bin/test test/controllers/ingresses/mandrill/inbound_emails_controller_test.rb
cd actionmailbox && bin/test
bundle exec rubocop actionmailbox/app/controllers/action_mailbox/ingresses/mandrill/inbound_emails_controller.rb actionmailbox/test/controllers/ingresses/mandrill/inbound_emails_controller_test.rb
git diff --check

### DIFF
--- a/actionmailbox/CHANGELOG.md
+++ b/actionmailbox/CHANGELOG.md
@@ -1,3 +1,10 @@
+*   Return `422 Unprocessable Content` for Mandrill events payloads that don't parse to a JSON array of objects.
+
+    Previously, valid JSON of the wrong shape (e.g. `null`, a scalar, an object, or an array containing non-objects)
+    raised an unhandled `NoMethodError` and resulted in a 500. This now matches the existing behavior for invalid JSON.
+
+    *Nesan Vettivel*
+
 *   Deprecate `Mail::Address.wrap` because it isn't used.
 
     *Gannon McGibbon*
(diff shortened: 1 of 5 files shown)


rafaelfranca:
How an a malformed Joan get to the endpoint?

rafaelfranca (approved):

Messier81:
> How an a malformed Joan get to the endpoint?

Anyone with the signing key sending a bad payload, or a webhook set up wrong. Now the endpoint returns 422 instead of crashing.

PR #57466, opened 2026-05-25
Dump enum types with fully qualified names for PostgreSQL

fatkodima:
This is a rebase and rework of #56278.
Closes #56278.


agrobbin:
Thanks for resurrecting this @fatkodima! Great to see it make it into `main`.

PR #57467, opened 2026-05-25
Add shims for Ractor shareability methods

andrewn617:
### Motivation / Background

This Pull Request has been created because we are going to start adding Ractor support to the framework. This is going to lead us to call `Ractor.make_shareable`, `Ractor.shareable_proc` etc. in many places which will necessitate checking the ruby version. Let's add a shim so we don't have to litter the code base with those checks.

### Detail

I added new core extension to Kernel which defines `ractor_shareable?`, `ractor_make_shareable`, `ractor_shareable_proc` and `ractor_shareable_lamda`. In environments where those are available, it forwards them to the Ractor method, otherwise they are a no-op.

I realize that adding new core extensions especially to Kernel can be a hot button topic, tho IMO its the right way to go so I started here. But if there is push back, we could do something like `RactorShareability.shareable?` instead. I am open to feedback or suggestions.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/core_ext/kernel.rb
+++ b/activesupport/lib/active_support/core_ext/kernel.rb
@@ -2,4 +2,5 @@
 
 require_relative "kernel/concern"
 require_relative "kernel/reporting"
+require_relative "kernel/ractor_shareability"
 require_relative "kernel/singleton_class"
--- a/activesupport/lib/active_support/core_ext/kernel/ractor_shareability.rb
+++ b/activesupport/lib/active_support/core_ext/kernel/ractor_shareability.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+# Shims for +Ractor+ shareability methods so framework code can call them
+# unconditionally regardless of the Ruby version.
+#
+# When +Ractor+ is not defined, or the underlying method is not available, the
+# shim is a no-op that simply returns its argument (or the given block).
+# Otherwise the call is forwarded to the matching +Ractor+ class method.
+#
+#   ractor_make_shareable(obj)        # => Ractor.make_shareable(obj)        or obj
+#   ractor_shareable?(obj)            # => Ractor.shareable?(obj)            or obj
+#   ractor_shareable_proc   { ... }   # => Ractor.shareable_proc   { ... }   or the block
+#   ractor_shareable_lambda { ... }   # => Ractor.shareable_lambda { ... }   or the block
+module Kernel
+  private
+    if defined?(Ractor) && Ractor.respond_to?(:make_shareable)
+      # Makes +obj+ Ractor-shareable by delegating to +Ractor.make_shareable+.
+      #
+      # The +copy:+ option is forwarded unchanged. On Ruby versions without
+      # +Ractor.make_shareable+, this shim returns +obj+ unchanged.
+      def ractor_make_shareable(obj, copy: false)
+        Ractor.make_shareable(obj, copy: copy)
+      end
+    else
+      def ractor_make_shareable(obj, copy: false)
+        obj
+      end
+    end
+
+    if defined?(Ractor) && Ractor.respond_to?(:shareable?)
+      # Returns whether +obj+ is Ractor-shareable by delegating to
+      # +Ractor.shareable?+.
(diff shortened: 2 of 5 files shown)


activesupport/lib/active_support/core_ext/kernel.rb
```diff
@@ -2,4 +2,5 @@
 
 require_relative "kernel/concern"
 require_relative "kernel/reporting"
+require_relative "kernel/shareability"
```
gmcgibbon:
```suggestion
require_relative "kernel/ractor_shareability"
```
We should qualify this module's name and its methods to state that it is to do with Ractors. `ractor_shareable?`, `ractor_shareable_lambda`, etc.

rafaelfranca: Agree. And that has less change of clashing with existing method names in other people classes.

rafaelfranca (changes requested):
Those methods can clash too much with existing code.

activesupport/lib/active_support/core_ext/kernel.rb
andrewn617: Good point. It's done @rafaelfranca @gmcgibbon

matthewd: I agree this seems safer against user-code conflicts, but I'm still a little nervous about potential conflicts with future upstream methods as this general requirement spreads through the ecosystem. 🤔

andrewn617: Do you mean upstream in Ruby? It seems unlikely to me, but I take the general point.

What about `Rails.ractor_shareable_proc`

rafaelfranca: In that case we just undefined ours. This isn't different from all other core extensions we have here and sometimes downstream to Ruby

activesupport/lib/active_support/core_ext/kernel/ractor_shareability.rb
```diff
@@ -0,0 +1,56 @@
...
+    def ractor_shareable_lambda(self: nil, &block)
+      Ractor.shareable_lambda(self: { self: }[:self], &block)
+    end
+  else
+    def ractor_shareable_lambda(&block)
+      block
+    end
+  end
+
+  private :ractor_make_shareable, :ractor_shareable?, :ractor_shareable_proc, :ractor_shareable_lambda
```
rafaelfranca: Why not use a regular private as directive in the top of the module?

activesupport/lib/active_support/core_ext/kernel/ractor_shareability.rb
```diff
@@ -0,0 +1,56 @@
...
+# shim is a no-op that simply returns its argument (or the given block).
+# Otherwise the call is forwarded to the matching +Ractor+ class method.
+#
+#   ractor_make_shareable(obj)        # => Ractor.make_shareable(obj)        or obj
+#   ractor_shareable?(obj)            # => Ractor.shareable?(obj)            or obj
+#   ractor_shareable_proc   { ... }   # => Ractor.shareable_proc   { ... }   or the block
+#   ractor_shareable_lambda { ... }   # => Ractor.shareable_lambda { ... }   or the block
+module Kernel
+  if defined?(Ractor) && Ractor.respond_to?(:make_shareable)
+    def ractor_make_shareable(obj, copy: false)
```
rafaelfranca: I think we should document all methods

activesupport/lib/active_support/core_ext/kernel.rb
```diff
@@ -2,4 +2,5 @@
 
 require_relative "kernel/concern"
 require_relative "kernel/reporting"
```
rafaelfranca, line 4: This new extension should be documented in the guides as well

rafaelfranca (changes requested):

rafaelfranca (approved):

gmcgibbon (approved):

byroot:
I just saw this. Do we really need / want to add this in `Kernel`? Could perfectly be singleton methods on `ActiveSupport`.

PR #57468, opened 2026-05-26
Allow showing server URL under Puma via config.server_host

ddbrendan:
### Motivation / Background

This Pull Request has been created because Rails currently suppresses the server URL in the boot banner when running under Puma (introduced in #29902 to avoid showing an incorrect URL when `config/puma.rb` overrides the host or port).

### Detail

This Pull Request adds a new application config option, `config.server_host`. When set, the boot banner displays a URL using that host, even under Puma. When unset, the existing behavior is preserved. The URL line is shown for non-Puma handlers and suppressed under Puma.

### Additional information

The `display_host` helper falls back to `options[:Host]` (the bind address) when `config.server_host` is unset, so non-Puma handlers continue to display the bound address as before.

### Checklist

- [x] Tests are added covering: opt-in under Puma, override for non-Puma, and the existing suppress-under-Puma default.
- [x] CHANGELOG update

### DIFF
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -569,6 +569,17 @@ in `config/credentials.yml.enc`. See the [`secret_key_base` API documentation](
 https://api.rubyonrails.org/classes/Rails/Application.html#method-i-secret_key_base)
 for more information and alternative configuration methods.
 
+#### `config.server_host`
+
+Sets a host name to display in the `bin/rails server` boot banner. When set, the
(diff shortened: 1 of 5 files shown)


rafaelfranca:
Thank you for the pull request but showing a host in a specific command doesn't warrant an application level configuration. I'd rather not show the host at all than introduce a new config. Let's see if there is another approach we can use, maybe a change in puma that would allow it to work as other server.

PR #57473, opened 2026-05-26
Use registry to track jobs that change their queue adapter

### LINKED ISSUE #57440: ActiveJob::TestHelper scans all job descendants twice per test to reset queue adapters
### Steps to reproduce

`ActiveJob::TestHelper` recomputes `queue_adapter_changed_jobs` in both `before_setup` and `after_teardown`:

```ruby
def before_setup
  queue_adapter_specific_to_this_test_class = queue_adapter_for_test
  queue_adapter_changed_jobs.each do |klass|
    # ...
  end
  # ...
end

def after_teardown
  super
  queue_adapter_changed_jobs.each { |klass| klass.disable_test_adapter }
end

def queue_adapter_changed_jobs
  (ActiveJob::Base.descendants << ActiveJob::Base).select do |klass|
    # only override explicitly set adapters, a quirk of `class_attribute`
    klass.singleton_class.public_instance_methods(false).include?(:_queue_adapter)
  end
end
```

That makes every test that includes `ActiveJob::TestHelper` pay an `O(number_of_job_classes)` scan twice per test, even when the test does not enqueue jobs and even when no job-specific queue adapter changed.

Here is a small benchmark that mirrors a large Rails app with many job classes:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activejob", "8.1.3"
  gem "benchmark-ips"
end

require "active_job"
require "active_job/test_helper"
require "benchmark/ips"

class ApplicationJob < ActiveJob::Base; end

2_500.times do |i|
  Object.const_set("GeneratedJob#{i}", Class.new(ApplicationJob))
end

helper = Object.new.extend(ActiveJob::TestHelper)

puts "ActiveJob #{ActiveJob::VERSION::STRING}"
puts "Job descendants: #{ActiveJob::Base.descendants.size}"
puts "Changed jobs: #{helper.send(:queue_adapter_changed_jobs).size}"

Benchmark.ips do |x|
  x.config(time: 5, warmup: 2)

  x.report("queue_adapter_changed_jobs") do
    helper.send(:queue_adapter_changed_jobs)
  end

  x.report("[ActiveJob::Base]") do
    [ActiveJob::Base]
  end

  x.compare!
end
```

On my machine this produces:

```text
ActiveJob 8.1.3
Job descendants: 2501
Changed jobs: 1
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin24]

queue_adapter_changed_jobs    762.293 i/s    (1.31 ms/i)
[ActiveJob::Base]              21.772M i/s   (45.93 ns/i)
```

Because `queue_adapter_changed_jobs` runs in both setup and teardown, this benchmark translates to about 2.6 ms of reflection overhead per test at this scale. In a large Rails app, this method shows up prominently in test-suite profiles; the hot operation is `Module#public_instance_methods(false)`, which allocates an array for each job class and then linearly scans it.

### Expected behavior

Changing / resetting Active Job's test queue adapter should not require a full reflection scan of every `ActiveJob::Base` descendant before and after every test.

Ideally the work should be proportional to the number of jobs that actually have an explicitly changed adapter, or avoided entirely for tests that do not need Active Job test adapter setup.

### Actual behavior

Every test that includes `ActiveJob::TestHelper` pays the descendant scan twice:

1. once in `before_setup`, to enable the test adapter; and
2. once in `after_teardown`, to disable it.

This is especially expensive in large applications where:

* there may be thousands of `ActiveJob::Base` descendants;
* `ActiveJob::TestHelper` can be included by broad test base classes or indirectly through mailer test helpers;
* most tests do not enqueue jobs; and
* the set of classes with explicit queue adapters is usually tiny and stable after boot.

### Possible directions

A few possible ways to avoid the per-test full scan:

* Track classes when `queue_adapter=` / `_queue_adapter=` is explicitly assigned, and have `queue_adapter_changed_jobs` read that tracked set instead of reflecting over all descendants.
* Cache the result when classes are eager-loaded, with an invalidation path for lazy loading / reloading / newly defined test jobs.
* Lazily install the test adapter only around helper operations that need it instead of in every test setup.
* If current `class_attribute` inheritance semantics mean setting the test adapter on `ActiveJob::Base` is sufficient, consider avoiding the descendant scan and operating on `ActiveJob::Base` directly.

I am happy to work on a PR, but wanted to confirm the preferred direction first because this code has some subtle compatibility history around per-job queue adapters in tests.

### System configuration

**Rails version**: Active Job / Active Support 8.1.3; the same implementation exists on `rails/rails` main / 8.2.0.alpha

**Ruby version**: ruby 4.0.5

### LINKED ISSUE #57441: ActiveJob::TestHelper no longer detects subclasses with explicit queue adapters
### Steps to reproduce

`ActiveJob::TestHelper#queue_adapter_changed_jobs` appears to rely on a `class_attribute` implementation detail to detect job classes that explicitly set their own queue adapter:

```ruby
def queue_adapter_changed_jobs
  (ActiveJob::Base.descendants << ActiveJob::Base).select do |klass|
    # only override explicitly set adapters, a quirk of `class_attribute`
    klass.singleton_class.public_instance_methods(false).include?(:_queue_adapter)
  end
end
```

That predicate no longer detects subclass assignments in Active Job / Active Support 8.1.3.

Reproduction:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activejob", "8.1.3"
end

require "active_job"
require "active_job/test_helper"
require "minitest/autorun"

class InlineAdapter
  def enqueue(job); end
  def enqueue_at(job, timestamp); end
end

class ApplicationJob < ActiveJob::Base; end

class CustomAdapterJob < ApplicationJob
  self.queue_adapter = InlineAdapter.new
end

class QueueAdapterChangedJobsTest < Minitest::Test
  include ActiveJob::TestHelper

  def test_jobs_with_explicit_queue_adapters_are_detected
    changed_jobs = send(:queue_adapter_changed_jobs)

    assert_equal InlineAdapter, CustomAdapterJob._queue_adapter.class
    assert_includes changed_jobs, CustomAdapterJob
  end
end
```

On 8.1.3 this fails with:

```text
Expected [ActiveJob::Base] to include CustomAdapterJob.
```

Changing the gem version to 8.0.0 makes the test pass.

A smaller Active Support-only reproduction shows the underlying behavior change: https://gist.github.com/bmansoob/cd460b18647907f65d941e8171b12381

Local output from that gist:

```text
# ActiveSupport 8.0.0
HedwigJob:
  _queue_adapter value: "HedwigAdapter"
  public own:
[:_queue_adapter, :_queue_adapter=]

Rails queue_adapter_changed_jobs-style result:
  [FakeJobBase, HedwigJob]

RESULT: OK - subclass explicit assignment is detected
```

```text
# ActiveSupport 8.1.3
HedwigJob:
  _queue_adapter value: "HedwigAdapter"
  public own:
[]
  private own:
[:__class_attr__queue_adapter, :__class_attr__queue_adapter=]

Rails queue_adapter_changed_jobs-style result:
  [FakeJobBase]

RESULT: BROKEN - subclass explicit assignment is NOT detected
```

### Expected behavior

`queue_adapter_changed_jobs` should include job classes that explicitly set their own queue adapter.

In the reproduction above, `CustomAdapterJob` explicitly assigns:

```ruby
self.queue_adapter = InlineAdapter.new
```

so it should be included in `queue_adapter_changed_jobs`, as it was in Active Job / Active Support 8.0.0.

### Actual behavior

`CustomAdapterJob._queue_adapter` is set correctly, but `queue_adapter_changed_jobs` returns only `[ActiveJob::Base]`.

The current predicate:

```ruby
klass.singleton_class.public_instance_methods(false).include?(:_queue_adapter)
```

no longer works as a proxy for "this subclass explicitly assigned the class attribute".

In 8.0.0, assigning the class attribute on a subclass defined public `_queue_adapter` / `_queue_adapter=` singleton methods directly on that subclass, so the predicate worked.

In 8.1.3, assigning the class attribute on a subclass stores the value behind internal generated methods instead. The subclass has the correct `_queue_adapter` value, but no public `_queue_adapter` method directly defined on its singleton class, so the predicate misses it.

### Notes

This may be a correctness issue in `ActiveJob::TestHelper`, or it may mean the descendant scan is now vestigial under the newer `class_attribute` implementation. Either way, the current code and comment still say Rails is detecting jobs with explicitly set adapters, but that is no longer what the predicate does.

Possible directions:

* update `queue_adapter_changed_jobs` to use a predicate compatible with the current `class_attribute` implementation;
* expose/use a supported `class_attribute` ownership predicate instead of reflecting on generated methods; or
* if setting the test adapter on `ActiveJob::Base` is now sufficient for all descendants, simplify `queue_adapter_changed_jobs` accordingly and remove the stale subclass-detection predicate.

Related performance issue for this same scan: #57440

### System configuration

**Rails version**: Active Job / Active Support 8.1.3; also reproduced on `rails/rails` main / 8.2.0.alpha

**Ruby version**: ruby 4.0.5

rafaelfranca:
This will avoid rescanning all Active Job descendants in `ActiveJob::TestHelper` before and after each test by tracking classes that explicitly change their queue adapter.

Fixes #57440.
Fixes #57441.

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Avoid rescanning all Active Job descendants in `ActiveJob::TestHelper`
+    before and after each test by tracking classes that explicitly change
+    their queue adapter.
+
+    *Rafael Mendonça França*
+
 *   Add `ActiveJob::Attributes` for declaring typed attributes that persist across
     job serialization and deserialization. It is included by `ActiveJob::Continuable`
     but can also be used standalone.
--- a/activejob/lib/active_job/test_helper.rb
+++ b/activejob/lib/active_job/test_helper.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/core_ext/class/subclasses"
+require "active_support/descendants_tracker"
 require "active_support/testing/assertions"
 
 module ActiveJob
@@ -18,8 +19,25 @@ module ActiveJob
       included do
         class_attribute :_test_adapter, instance_accessor: false, instance_predicate: false
 
-        @_queue_adapter_changed_jobs = Set.new << self
+        @_queue_adapter_changed_jobs = ActiveSupport::DescendantsTracker::WeakSet.new
+        @_queue_adapter_changed_jobs << self
+        @_queue_adapter_changed_jobs_backfilled = false
         singleton_class.attr_reader :_queue_adapter_changed_jobs # :nodoc:
+        singleton_class.attr_accessor :_queue_adapter_changed_jobs_backfilled # :nodoc:
+
+        def self.backfill_queue_adapter_changed_jobs # :nodoc:
+          return if _queue_adapter_changed_jobs_backfilled
+
+          jobs_classes = ActiveJob::Base.descendants + [ActiveJob::Base]
+
+          jobs_classes.each do |klass|
+            if klass.singleton_class.private_method_defined?(:__class_attr__queue_adapter_owner, false)
+              _queue_adapter_changed_jobs << klass
+            end
+          end
+
+          self._queue_adapter_changed_jobs_backfilled = true
+        end
       end
 
       module ClassMethods
@@ -48,8 +66,10 @@ module ActiveJob
     end
 
     def before_setup # :nodoc:
+      ActiveJob::Base.backfill_queue_adapter_changed_jobs
+
       queue_adapter_specific_to_this_test_class = queue_adapter_for_test
-      ActiveJob::Base._queue_adapter_changed_jobs.each do |klass|
+      queue_adapter_changed_jobs.each do |klass|
         if queue_adapter_specific_to_this_test_class
           klass.enable_test_adapter(queue_adapter_specific_to_this_test_class)
         elsif klass._queue_adapter.nil?
@@ -65,7 +85,7 @@ module ActiveJob
     def after_teardown # :nodoc:
       super
 
-      ActiveJob::Base._queue_adapter_changed_jobs.each { |klass| klass.disable_test_adapter }
+      queue_adapter_changed_jobs.each { |klass| klass.disable_test_adapter }
     end
 
     # Returns a queue adapter instance to use with all Active Job test helpers.
@@ -763,6 +783,10 @@ module ActiveJob
         job.scheduled_at = Time.at(payload[:at]) if payload.key?(:at)
         job.send(:deserialize_arguments_if_needed) unless skip_deserialize_arguments
         job
+      end
+
+      def queue_adapter_changed_jobs
+        ActiveJob::Base._queue_adapter_changed_jobs.to_a
       end
 
       def validate_option(only: nil, except: nil)
--- a/activejob/lib/active_job/test_helper.rb
+++ b/activejob/lib/active_job/test_helper.rb
@@ -17,11 +17,20 @@ module ActiveJob
 
       included do
         class_attribute :_test_adapter, instance_accessor: false, instance_predicate: false
+
+        @_queue_adapter_changed_jobs = Set.new << self
+        singleton_class.attr_reader :_queue_adapter_changed_jobs # :nodoc:
       end
 
       module ClassMethods
         def queue_adapter
           self._test_adapter.nil? ? super : self._test_adapter
+        end
+
+        def queue_adapter=(name_or_adapter) # :nodoc:
+          super.tap do
+            ActiveJob::Base._queue_adapter_changed_jobs << self
+          end
         end
 
         def disable_test_adapter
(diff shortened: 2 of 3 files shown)


byroot:
Hum, can we take a step back here?

My understanding is that this test helper want all jobs to use the `TestAdapter`, to do that it does replace `ActiveJob::Base.queue_adapter`, but since this is a `class_attribute`, it has to find all child classes on which `queue_adapter` has been overridden.

Wouldn't it be simpler if we had some `ActiveJob.forced_queue_adapter` or something, that all jobs would use over their own `queue_adapter` if it's set? This way we'd be able to achieve the same in `O(1)`.

Or am I totally off?

PR #57475, opened 2026-05-26
Active Record Ractor-safe freezing

gmcgibbon:
### Motivation / Background

This Pull Request has been created to continue to make Active Record more Ractor safe.

### Detail

This Pull Request changes various Active Record modules to freeze and constantize as much as possible so that they can be used in Ractorized context.

### Additional information

Similar to https://github.com/rails/rails/pull/57323

### Checklist

### DIFF
--- a/activerecord/lib/active_record/readonly_attributes.rb
+++ b/activerecord/lib/active_record/readonly_attributes.rb
@@ -8,7 +8,7 @@ module ActiveRecord
     extend ActiveSupport::Concern
 
     included do
-      class_attribute :_attr_readonly, instance_accessor: false, default: []
+      class_attribute :_attr_readonly, instance_accessor: false, default: [].freeze
     end
 
     module ClassMethods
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/relation/query_attribute.rb
```diff
@@ -26,7 +26,7 @@ def type_cast(value)
       end
 
       def value_for_database
-        @value_for_database = _value_for_database unless defined?(@value_for_database)
+        @value_for_database = _value_for_database.freeze unless defined?(@value_for_database)
```
andrewn617: I think we need to also not memoize this but set it in the initializer. Otherwise it will raise a ractor isolation error the first time we lazily evaluate this.

rafaelfranca (approved):

activerecord/lib/active_record/relation/query_attribute.rb
gmcgibbon: Ended up reverting this. We can't eagerly memorize this value.

PR #57479, opened 2026-05-27
Move migration compatibility behavior to per-adapter modules

yahonda:
### Motivation / Background

Adapter-specific migration compatibility logic was scattered across a single monolithic `Compatibility` class with `adapter_name` string comparisons. This made it hard to maintain and impossible for third-party adapters to extend without monkey-patching (e.g. [cockroachdb adapter](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/1c5d94bb654320d4bdaf4e8b5fd9fa010fc91a41), [oracle-enhanced #2596](https://github.com/rsim/oracle-enhanced/pull/2596), [oracle-enhanced #2710](https://github.com/rsim/oracle-enhanced/pull/2710)).

### Detail

Each built-in adapter now defines its own `MigrationCompatibility` module and returns it from the new `migration_compatibility_module_for` hook. Third-party adapters can override this hook to ship per-version legacy behavior without patching Rails internals.

The hook is registered on `Mysql2Adapter` and `TrilogyAdapter` (not `AbstractMysqlAdapter`) so third-party adapters subclassing the abstract class are not implicitly opted in.

### Additional information

Verified that the CockroachDB adapter's [monkey patch](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/1c5d94bb654320d4bdaf4e8b5fd9fa010fc91a41) is no longer needed — `migration_compatibility_module_for` is inherited from `PostgreSQLAdapter`, and `:datetime` → `:timestamp` conversion works correctly for `Migration[6.1]` on CockroachDB without any adapter-side patching.

### Checklist


activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
```diff
@@ -669,6 +669,10 @@ def create_virtual_table(*) # :nodoc:
       def drop_virtual_table(*) # :nodoc:
       end
 
+      def migration_compatibility_module_for(migration_class)
+        nil
```
rafaelfranca, line 673: Just remove this

rafaelfranca (commented):
I like the idea but this removed all the simplicity of the superclass of the migration containing the version, now, if I get this correctly the code of that class changes the first time a migration is executed so it is hard to predict which behavior that class would have, it depends on modules included at runtime, that depends on which adapter is used. That is a lot of complexity to me. 

A common design pattern to solve this problem is the strategy pattern, and we can avoid the inheritance complexity (module being included at runtime)

yahonda:
Agreed. I've started reworking it around the strategy pattern on a separate branch, and I'll open it as a separate pull request once it's ready.

yahonda:
https://github.com/rails/rails/pull/58162 has been opened that should not require  https://github.com/rails/rails/pull/57480 and something like that.

yahonda:
Superseded by #58162

PR #57480, opened 2026-05-27
Detect multi-adapter migration compatibility conflicts on shared base classes

yahonda:
### Motivation / Background

Follow-up to #57479. Seeking feedback on whether this conflict detection is worth adding.

In a multi-database app, sharing a single migration base class across different adapter types (e.g. PostgreSQL primary + MySQL secondary) silently accumulates conflicting compatibility modules. This can cause one adapter's defaults to leak into another's migration runs.

### Detail

Add `ActiveRecord::Migration::Compatibility::ConflictError`, raised when a migration base class is associated with compatibility behavior from more than one adapter type. The error message recommends using a distinct migration base class per adapter type. Single-adapter applications are unaffected.

### Additional information

N/A

### Checklist


rafaelfranca:
> In a multi-database app, sharing a single migration base class across different adapter types (e.g. PostgreSQL primary + MySQL secondary) silently accumulates conflicting compatibility modules. This can cause one adapter's defaults to leak into another's migration runs.

Yeah, that is another reason why I think that approach is the wrong one.

yahonda:
Superseded by #58162

PR #57483, opened 2026-05-27
Make exception wrapper ractor safe

andrewn617:
### Motivation / Background

`ActionDispatch::ExceptionWrapper` state needs to be made Ractor safe.

### Detail

`rescue_responses` and `rescue_templates` already have a config on the engine. So I just freeze the underlying hash and create a new frozen one when adding the configured values.

`wrapper_exceptions` and `silent_exceptions` have no public config so I added them.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
+++ b/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
@@ -26,7 +26,7 @@ module ActionDispatch
       "ActionController::TooManyRequests"                  => :too_many_requests,
       "Rack::QueryParser::ParameterTypeError"              => :bad_request,
       "Rack::QueryParser::InvalidParameterError"           => :bad_request
-    )
+    ).freeze
 
     cattr_accessor :rescue_templates, default: Hash.new("diagnostics").merge!(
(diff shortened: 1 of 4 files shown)


actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
```diff
@@ -35,16 +35,16 @@ class ExceptionWrapper
...
       "ActionController::MissingExactTemplate" => "missing_exact_template",
-    )
+    ).freeze
 
     cattr_accessor :wrapper_exceptions, default: [
       "ActionView::Template::Error"
-    ]
+    ].freeze
 
     cattr_accessor :silent_exceptions, default: [
```
rafaelfranca, line 44: This needs to be a config as well. It is here not for application, but for rails plugins to configure

rafaelfranca (changes requested):

actionpack/lib/action_dispatch/middleware/exception_wrapper.rb
andrewn617: Ah my bad. Done

rafaelfranca (approved):

PR #57488, opened 2026-05-27
Allow `reorder` to be called without arguments

pierry01:
### Summary

`reorder` currently requires at least one argument and raises `ArgumentError: The method .reorder() must contain arguments.` when called with none. As a result, the only way to strip an existing order off a relation is to pass an explicit `nil`:

```ruby
Post.order(:title).reorder(nil) # removes the ORDER BY
Post.order(:title).reorder      # before: ArgumentError
```

This makes the no-argument call a shorthand for `reorder(nil)`, so it simply removes any existing order:

```ruby
Post.order(:title).reorder # => no ORDER BY clause
```

The behavior for all existing call styles is unchanged — `reorder(:id)`, `reorder("id desc")`, `reorder(nil)`, `reorder([])` all behave exactly as before. The argument-sanitizing pipeline (`sanitize_order_arguments` + flatten + `compact_blank!`) is preserved for the non-empty case; only the "must contain arguments" guard is dropped for `reorder`.

### Other Information

- `reorder` is removed from the shared `test_no_arguments_to_*_raise_errors` list, and a new test (`test_order_with_reorder_without_arguments_removes_the_order`) covers the empty call clearing the order while still flipping `reordering_value`.
- CHANGELOG entry added.

All `activerecord` relation tests pass on SQLite, and RuboCop reports no offenses on the changed files.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Allow `reorder` to be called without arguments to remove an existing order.
+
+    Previously `reorder` required at least one argument and raised an
+    `ArgumentError` when called with none, so removing an order meant calling
+    `reorder(nil)`. It can now be called without arguments as a shorthand:
+
+    ```ruby
+    Post.order(:title).reorder # => no ORDER BY clause, same as reorder(nil)
+    ```
+
+    *Jean Mendonça*
+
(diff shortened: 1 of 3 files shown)


fatkodima:
Imo, `reorder(nil)` makes some sense and reads like "reorder with nil order, aka remove the order". 
Just `reorder` does not make sense and reads simply like "reorder" 🤷. But in what order etc?
The clearest method to remove the order is `unscope(:order)`.

PR #20868, opened 2015-07-13
Make AC::Parameters not inherited from Hash

sikachu:
This is another take at #14384 as we decided to wait until `master` is targeting Rails 5.0. This commit is implementation-complete, as it guarantees that all the public methods on the hash-inherited Parameters
are still working (based on test case). We can decide to follow-up later if we want to remove some methods out from Parameters.

PR #57490, opened 2026-05-28
fix: ParameterFilter not filtering nested ActionController::Parameters

### LINKED ISSUE #57474: ActiveSupport::ParameterFilter does not filter nested ActionController::Parameters
### Steps to reproduce

```ruby
# I want to filter every parameter called 'secret'
filter = ActiveSupport::ParameterFilter.new ['secret']
```

```ruby
# If I pass a Ruby hash with a root 'secret' and a nested 'secret' key, both keys are filtered ✅
params = { "secret" => 1234, "nested" => { "secret" => 5678 } }
filter.filter params
# => {"secret" => "[FILTERED]", "nested" => {"secret" => "[FILTERED]"}}
```

```ruby
# If I pass an instance of ActionController::Parameters, the nested key is not filtered 🚫
params = ActionController::Parameters.new "secret" => 1234, "nested" => { "secret" => 5678 }
filter.filter params
# => #<ActionController::Parameters {"secret" => "[FILTERED]", "nested" => #<ActionController::Parameters {"secret" => 5678} permitted: false>} permitted: false>
```

### Root cause

`ActiveSupport::ParameterFilter` is able to encrypt keys nested in a Hash because of these lines of code:

https://github.com/rails/rails/blob/e852dfeeeafc4424735b441c06a45a74c87ff067/activesupport/lib/active_support/parameter_filter.rb#L190-L191

`ActiveSupport::ParameterFilter` does not inherit from Hash since #20868 therefore the nested behavior is not triggered.

### Expected behavior

I would like to know the opinion of the Core team on this issue. 

I personally think it's a bug (or at least a very unexpected behavior) and I'm happy to write a Pull Request to fix it.

But it could also be considered a lack of documentation: in this case I'm happy to improve the [documentation for filter](https://api.rubyonrails.org/classes/ActiveSupport/ParameterFilter.html#method-i-filter) to explain that the `params` argument should not be an instance of `ActionController::Parameters`.

Thanks!

ruyrocha:
### Motivation / Background

This Pull Request has been created because `ActiveSupport::ParameterFilter` silently skips filtering sensitive keys nested inside `ActionController::Parameters` objects. Since `ActionController::Parameters` stopped inheriting from `Hash` in #20868, the `value.is_a?(Hash)` guard in `value_for_key` is never entered for nested parameter objects. This is a security issue: sensitive keys such as `password`, `token`, and `secret` in nested params pass through unfiltered to logs and error trackers. Fixes #57474.

### Detail

This Pull Request changes `value_for_key` in `ActiveSupport::ParameterFilter` to add an `elsif value.respond_to?(:to_unsafe_h)` branch that converts the nested `ActionController::Parameters` to a plain Hash via `to_unsafe_h` before recursing. This keeps the `is_a?(Hash)` fast path untouched and avoids introducing an `activesupport` -> `actionpack` dependency.

### Additional information

The fix converts to a plain Hash before recursing rather than recursing into the `ActionController::Parameters` directly. This means the return type of a filtered nested value is a plain Hash, which is acceptable since filtered parameters are only used for logging and never returned to the application as params.

### Checklist

### DIFF
--- a/activesupport/test/parameter_filter_test.rb
+++ b/activesupport/test/parameter_filter_test.rb
@@ -3,6 +3,7 @@
 require_relative "abstract_unit"
 require "active_support/core_ext/hash"
 require "active_support/parameter_filter"
+require "action_controller"
 
 class ParameterFilterTest < ActiveSupport::TestCase
   test "process parameter filter" do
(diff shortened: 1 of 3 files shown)


activesupport/test/parameter_filter_test.rb
```diff
@@ -3,6 +3,7 @@
 require_relative "abstract_unit"
 require "active_support/core_ext/hash"
 require "active_support/parameter_filter"
+require "action_controller"
```
rafaelfranca, line 6: Active support should not depend on action controller

rafaelfranca (changes requested):

rafaelfranca:
Thank you for the pull request but the premise of the issue is wrong. The framework never pass a `ActiveControlller::Parameters` to the ParameterFilter. It only pass hashes and arrays.

PR #57493, opened 2026-05-28
[ci skip] Add step to create path for redirection before launching the application

### LINKED ISSUE #57471: Non-existent path helper in guides/source/sign_up_and_settings.md
### Steps to reproduce

In the Sign Up and Settings tutorial, the path helper `settings_profile_path` is used on line 369, but this path helper is created only later (l. 435) in the tutorial.

[link to location](https://github.com/rails/rails/blob/b5270673a754808984f7cc1afb36c0b72cb1ed3f/guides/source/sign_up_and_settings.md?plain=1#L369)

### Expected behavior

Path helpers in the tutorial are used only after they are created

### Actual behavior

Path helpers in the tutorial are used before they are created

maurobusso:
Fixes #57471

### Motivation / Background

This PR adds a step in the Sign Up and Settings tutorial docs which create the route for redirecting the user once the password is updated.
 
This Pull Request has been created because currently the path helper `settings_profile_path` is used before is actually created. This might be confusing for someone that follows this tutorial.

### Detail

This PR adds a step in the Sign Up and Settings tutorial docs which create the route for redirecting the user once the password is updated.

### DIFF
--- a/guides/source/sign_up_and_settings.md
+++ b/guides/source/sign_up_and_settings.md
@@ -390,6 +390,15 @@ browser to bypass this validation. To prevent this and ensure the validation
 always runs, we use `.with_defaults(password_challenge: "")` to set a default
 value even if the `password_challenge` parameter is missing.
 
+Now let's add the route to redirect the user once the password is updated.
+
+```ruby#3
+namespace :settings do
+  resource :password, only: [ :show, :update ]
+  resource :profile, only: [ :show ]
+end
+```
+
 You can now visit http://localhost:3000/settings/password to update your
 password.
 
@@ -425,7 +434,7 @@ first and last name.
 
 ### Profile Routes & Controller
 
-In `config/routes.rb`, add a profile resource under the settings namespace. We
+In `config/routes.rb`, add an `update` action on the profile resource under the settings namespace. We
 can also add a root to the namespace to handle any visits to `/settings` and
 redirect them to profile settings.
 


paul-louyot:
@rafaelfranca @maurobusso this PR does not fix #57471 : 

we can on the edge guides that the helper is still used before the route is declared.

https://edgeguides.rubyonrails.org/sign_up_and_settings.html#safely-updating-passwords

screenshot:

image

<hr>

I see two solutions to fix the issue:
- either keep using `settings_profile_path` inside `Settings::PasswordsController` and move the creation of the route, controller and view before the use of this helper,
- or revert this PR and replace `settings_profile_path` inside `Settings::PasswordsController` by `settings_password_path`

maurobusso:
Hi @paul-louyot, thank you for your comment. I am not sure I understand, why this does not fix #57471 ?

paul-louyot:
I opened the issue because a path_helper was used before it existed in the tutorial, and that is still the case now that this PR is merged.

I'm not sure if the issue should actually be fixed, because it from an app behavior level, it makes sense to redirect the user to `settings_profile_path`, and the tutorial might be easier to follow if each step is small. But using a non-existent path_helper might also be confusing for a new Rails user.

Anyway, I wanted to make explicit that this PR does not change the situation of using a path_helper before it exists

maurobusso:
I mean yes the path helper is still technically used before is created but is created right after the controller and before the user updates the password by going to `http://localhost:3000/settings/password`. So I think for the purpose of this tutorial if, you follow it step by step, it is still clear and easy to understand for a new Rails user. Unless there are some convention I was not aware of.

I guess the step for creating the path could be more clearly divided maybe 🤔

PR #57499, opened 2026-05-28
Allow Active Storage URL expiry to be configured with a callable

victorlcampos:
### Motivation / Background

`config.active_storage.service_urls_expire_in` and `config.active_storage.urls_expire_in` are stored as module-level state (`mattr_accessor` on `ActiveStorage`) and read directly wherever URLs are generated (`Blob#url`, `Variant#url`, the redirect controllers' `Cache-Control` header, and the `rails_storage_*` route helpers for signed-id expiry).

In a multi-tenant application that needs a different URL expiry per tenant, there is no hook to vary these values at request time without monkey-patching `Blob#signed_id` and the redirect controller.

### Detail

This makes both readers resolve a callable on read: if the configured value responds to `call`, it is invoked and the result is used. Because every consumption site already reads `ActiveStorage.service_urls_expire_in` / `ActiveStorage.urls_expire_in`, the resolution propagates without touching `Blob`, `Variant`, the routes, or the controllers.

```ruby
config.active_storage.service_urls_expire_in = -> { Current.tenant&.url_expiry || 5.minutes }
```

Fully backward compatible: `ActiveSupport::Duration`, `Integer` and `nil` do not respond to `call`, so they pass through unchanged.

### Additional information

Includes a unit test for both readers, integration tests proving the callable is resolved end-to-end (signed-id expiry and the `Cache-Control` header), a CHANGELOG entry, and guide docs.


victorlcampos:
@rafaelfranca

PR #43871, opened 2021-12-15
Allow any allowed host with port

rafaelfranca:
Allow any allowed host with port
    
All allowed host should be allowed if they are being accessed using a port.
    
Fixes #43870.

PR #57503, opened 2026-05-29
Reject malformed hosts with extra ports

### LINKED ISSUE #37956: Malformed host header causes server errors on redirect
### Steps to reproduce

* Create a minimal Rails application with one endpoint (endpoint R) redirecting to another endpoint.
* Call endpoint R with a malformed `Host` header (eg. `www.example.com:80:80`).  This will cause the server to generate an internal server error (return code 500).

The issue seems to be caused by `ActionDispatch` which tries to create the redirect URI based on the incorrect host header from the request.

Executable test case

```
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", github: "rails/rails"
end

require "minitest/autorun"
require "rack/test"

  config.root = __dir__
  config.hosts << "www.example.com" << "www.example.com:80"

  Rails.logger = config.logger

  routes.draw do
    get '/home', to: 'test#home'
    get '/home2', to: redirect('/home')
  end
end

  def home
    render plain: 'Home sweet home'
  end
end


class BugTest < ActionDispatch::IntegrationTest

  def test_home_endpoint_works
    get '/home'
    assert response.ok?
    assert "Home sweet home", response.body
  end

  def test_redirection_works
    get '/home2'
    assert response.redirect?

    get response.location
    assert response.ok?
    assert "Home sweet home", response.body
  end

  def test_redirection_does_not_error_on_malformed_host_header
    get '/home2', headers: {'Host': 'www.example.com:80:80'}
    assert_not_equal 500, response.status, "Malformed Host header caused a server error"
  end

  private

  def app
    Rails.application
  end

end
```

### Expected behavior
The rails server should not generate internal server errors on malformed requests. While these requests are not correct by themselves, a malicious user could use this to deliberately generate errors on the server.
Alternatively, there should be a way to catch these errors.

### Actual behavior
The rails server generates internal server errors on malformed requests.

### System configuration
**Rails version**: 6.1.0.alpha from https://github.com/rails/rails.git (at master@58df9a4)

**Ruby version**: 2.5.5

afurm:
### Motivation / Background

Fixes #37956.

When an allowed host is configured with an explicit port, HostAuthorization currently appends another optional port segment to the generated regexp. That lets a malformed Host value such as `www.example.com:80:80` pass the allowlist and reach downstream redirect code that expects a valid host.

### Detail

This Pull Request keeps the existing optional-port behavior for hosts configured without a port, while avoiding the extra optional port when the configured host already includes one. It also adds coverage for exact explicit-port hosts and malformed extra-port hosts.

### Additional information

Validation:

- `cd actionpack && bin/test test/dispatch/host_authorization_test.rb -i "/extra ports/"`
- `cd actionpack && bin/test test/dispatch/host_authorization_test.rb`
- `cd actionpack && bundle exec rubocop lib/action_dispatch/middleware/host_authorization.rb test/dispatch/host_authorization_test.rb`
- `git diff --check`

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/middleware/host_authorization.rb
+++ b/actionpack/lib/action_dispatch/middleware/host_authorization.rb
@@ -74,14 +74,26 @@ module ActionDispatch
 
         def sanitize_string(host)
           if host.start_with?(".")
-            /\A#{SUBDOMAIN_REGEX}?#{Regexp.escape(host[1..-1])}#{PORT_REGEX}?\z/i
+            /\A#{SUBDOMAIN_REGEX}?#{Regexp.escape(host[1..-1])}#{port_regexp(host)}\z/i
           else
-            /\A#{Regexp.escape host}#{PORT_REGEX}?\z/i
+            /\A#{Regexp.escape host}#{port_regexp(host)}\z/i
           end
         end
 
         def extract_hostname(host)
           host.slice(VALID_IP_HOSTNAME, "host") || host
+        end
+
+        def port_regexp(host)
+          host_with_port?(host) ? "" : "#{PORT_REGEX}?"
+        end
+
+        def host_with_port?(host)
+          if host.start_with?("[")
+            host.match?(/\]:\d+\z/)
+          else
+            host.count(":") == 1 && host.match?(/:\d+\z/)
+          end
         end
     end
 
--- a/actionpack/test/dispatch/host_authorization_test.rb
+++ b/actionpack/test/dispatch/host_authorization_test.rb
@@ -274,6 +274,39 @@ class HostAuthorizationTest < ActionDispatch::IntegrationTest
     assert_match "Success", response.body
   end
 
+  test "hosts configured with explicit port work" do
+    @app = build_app(["host.test:3000"])
+
+    get "/", env: {
+      "HOST" => "host.test:3000",
+      "action_dispatch.show_detailed_exceptions" => true
+    }
+
+    assert_response :ok
+    assert_match "Success", response.body
+  end
+
+  test "blocks malformed hosts with extra ports" do
+    redirect_app = -> env do
+      request = ActionDispatch::Request.new(env)
+      uri = URI.parse("/login")
+      uri.scheme = request.scheme
+      uri.host = request.host
+
+      [301, { Rack::LOCATION => uri.to_s }, []]
+    end
+
+    @app = build_app(["www.example.com:80"], lint: false, app: redirect_app)
+
+    get "/", env: {
+      "HOST" => "www.example.com:80:80",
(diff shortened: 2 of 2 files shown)


p8:
Thanks for the PR @afurm 
Why would anyone add ports to the allowed hosts?
Host authorization protection is about DNS. A domain name will point to the same host no matter what the port is.

afurm:
You are right that users should not need to add ports to `config.hosts`; `config.hosts << "www.example.com"` is enough, and Rails has intentionally allowed that entry to match `www.example.com:3000` since #43871.

The reason this came up is that #37956 includes an explicit-port entry (`www.example.com:80`). With the current sanitizer, that becomes a regexp that also accepts another optional port, so `www.example.com:80:80` is treated as authorized and can reach downstream code that expects a valid host.

So the PR is not trying to recommend port-specific host allowlists. It is just making an existing explicit-port entry match only that exact host header instead of also matching malformed extra-port variants. I can drop the positive explicit-port test if that makes the intent clearer.

actionpack/lib/action_dispatch/middleware/host_authorization.rb
```diff
@@ -74,15 +74,27 @@ def sanitize_regexp(host)
...
           host.slice(VALID_IP_HOSTNAME, "host") || host
         end
+
+        def port_regexp(host)
+          host_with_port?(host) ? "" : "#{PORT_REGEX}?"
+        end
+
+        def host_with_port?(host)
+          if host.start_with?("[")
+            host.match?(/\]:\d+\z/)
```
p8, line 93: @byroot If I comment out this branch, the tests are still green.
This check also doesn't make any sense. Which host starts with a "["?

p8: Ah ok, this is for IPV6 "[::]".
Still, commenting out this branch doesn't break the tests.

Also, this commit doesn't check for "[::]:80:80"

p8: It seems we are reinventing IP/host parsing here.

byroot: I'll add a test with an IPv6.

But note that the goal here isn't to be fully bullet proof or adversarial, just to detect when one of the authorized host contain a port number, which does make sense to me.

byroot: b96756bc69f397b41753be34d66cbe55566e07ef

p8: Ah yes that makes sense.

PR #50979, opened 2024-02-06
Action Cable server adapterization

palkan:
### Motivation / Background

This PR aims to make Action Cable more flexible and extensible as a library/framework by separating low-level, implementation specific details (e.g., handling WebSockets, dealing with concurrency) from abstractions (Connection, Channel classes). That would make it possible to use user-level Action Cable code with alternative web servers or execution models (e.g., Fibers) without rewriting it or monkey-patching internals.

> [!TIP]
> You can find a more detailed description of the problem, motivation and potential benefits of such refactoring in [this document](https://docs.google.com/document/d/1oGf86NAzQL9sDw6rTfUNSe3mzzg7wX64sc34D-_3EDI/edit#heading=h.d3db1fd03ano).

> [!NOTE] 
> This PR does not introduce any significant new features or public API breaking changes. It doesn't affect the existing functionality of built-in Action Cable server.

> [!IMPORTANT]
> This functionality from this PR is already available for Rails 7+ application through the [actioncable-next gem](https://github.com/anycable/actioncable-next). Please, give it a try and share your feedback here, this would help to get this PR merged into Rails.

### Detail

The changes are grouped into commits to reflect the following major changes.

1) **Reorganizing threaded executors**. 

Action Cable relies on two threaded executors right now: one to perform user-space code (Worker) and another one to perform internal async operations around pub/sub. The latter was for some (historical?) reason hidden within the `StreamEventLoop` class that implements and IO loop to perform read/write operations on sockets. I've extracted the executor and promoted it into a server attribute independent of IO event loop. The main motivation of this extraction is not only in enforcing SoC but to separate WebSocket-specific code from more generic async executor used throughout the codebase.

I've also wrapped the Concurrent Ruby threaded executor into a custom class with a minimal interface (`#post`, `#timer` and `#shutdown` methods), so it will be possible to configure a custom executor for a server in the future (one use-case we have in mind is using a Fiber-based executor instead of a Thread-based one).

Note that this PR doesn't include any configuration-related code (e.g., to support smth like `config.action_cable.async_executor = :fiber`) but I'm considering adding something like that in the follow-up PRs.

2) **Limiting the scope of async operations**.

This step is the follow-up to the async executor extraction. We have a few places in the Channel internals which directly refer to the `event_loop` but they're all related to pub/sub operations. So, to avoid unnecessary async operations at the abstraction level, I moved all it to the pub/sub layer by and standardized it across adapters by using a `SubscriberMap::Async` subclass.

[...]

Issue #48292, opened 2023-05-24
ActionCable sporadically fails to confirm subscribe because the subscription "already exists"


shoffing:
### Steps to reproduce

- Create a very basic channel that simply streams for the current user:

```rb
class PerfTestChannel < ApplicationCable::Channel
  def subscribed
    stream_for current_user
  end
end
```

- On page load, subscribe to the channel.

```js
App.cable.subscriptions.create('PerfTestChannel');
```

### Expected behavior

The ActionCable client should be able to reliably subscribe to a channel.

### Actual behavior

- On a development instance with browser network latency simulation, or any staging/production instance, the `subscribe` command will sporadically (varying from 5%-70% depending on load) fail to be confirmed. The `SubscriptionGuarantor` will repeatedly attempt to resubscribe every 500ms, but no confirmation will ever be received.

![image](https://github.com/rails/rails/assets/2903742/8dc591b4-ef3a-4d7b-867d-b92ee6f5b906)

- I wrote a JS script to assist in reproducing this error. It repeatedly disconnects & subscribes to the channel until the subscription fails to leave the pendingSubscriptions state in under 600ms. Running this in browser JS console reliably reproduces the issue on a staging/production instance:

```js
(async function() {
  let i = 0;
  while (true) {
    i++;
    App.cable.subscriptions.subscriptions.forEach(sub => App.cable.subscriptions.remove(sub));
    App.cable.disconnect();
    App.cable.subscriptions.create('PerfTestChannel');
    await new Promise(resolve => setTimeout(resolve, 600));
    const pendingSubs = App.cable.subscriptions.guarantor.pendingSubscriptions;
    if (pendingSubs.length > 0) {
      break;
    } else { console.log('...' + i); }
  }
})()
```

### Investigation

This bug was tricky to investigate as there was nothing useful in our logs. The `/cable` request was successfully being upgraded to websockets as expected. We saw some errors that might have been relevant, but none of them occurred reliably with the issue being observed. After digging through the ActionCable source code, we have determined the issue is caused by this line in `subscriptions.rb`: https://github.com/rails/rails/blob/0f5c8c5bc61b7e382e64cad4846406021bc8cd35/actioncable/lib/action_cable/connection/subscriptions.rb#L34

It returns silently if the subscription "already exists". We confirmed this is the source of the error by monkey-patching the class in a `config/initializers` to add some logging with the `return`:

```rb
# config/initializers/monkeypatch_actioncable.rb
# frozen_string_literal: true

[...]

PR #44653, opened 2022-03-10
ActionCable: Subscribe uniquely

sj26:
If a subscription for the same identifier already exists then it has already been or is being subscribed via the connection, so no need to re-subscribe. Without this change, a subscription message is sent to the server and the guarantor waits for a subscription confirmation which never comes, and so keeps sending subscription messages over and over every second. Fixes #44652.

I'm not sure this is a comprehensive solve, and I'm not sure yet how to test it. I'd love some feedback on the approach. 🙏

I don't *think* short-circuiting a subscription like this will be a problem. I can't find a place where subscription confirmations are shared back to subscriptions for handling.

Issue #44652, opened 2022-03-10
ActionCable: Repeated subscription attempts


sj26:
### Steps to reproduce

1. Create an actioncable subscription.
2. Await subscription confirmation.
3. Create another subscription with the same identifier.
4. Observe websocket showing subscription attempts every second.

I tried creating a small reproduction script, but involving views and action cable got too complicated. Instead, here's an app which replicates it:

https://github.com/sj26/action_cable_test

It creates two subscriptions with the same identifier, with a timeout to be sure that the first subscription has already been confirmed:

```erb
<div id="count"></div>
<%= javascript_tag do %>
  window.App.consumer.subscriptions.create({channel: "TestChannel"}, {
    received({ count }) {
      document.getElementById("count").innerText = count;
    }
  });
<% end %>

<div id="contents"></div>
<%= javascript_tag do %>
  setTimeout(function() {
    window.App.consumer.subscriptions.create({channel: "TestChannel"}, {
      received({ contents }) {
        document.getElementById("contents").innerText = contents;
      }
    });
  }, 1000)
<% end %>
```

Then a stream of subscription messages can be send in the websocket traffic:

<img width="1193" alt="image" src="https://user-images.githubusercontent.com/14028/157585817-14b2546d-0601-4faf-9334-e2615127be77.png">

because the guarantor considers the second subscription pending:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/14028/157586067-5ce7d08d-263a-4ea7-bfe4-9ab581f589b1.png">

because actioncable short circuits existing subscriptions, and so doesn't submit multiple confirmations:

https://github.com/rails/rails/blob/v7.0.2/actioncable/lib/action_cable/connection/subscriptions.rb#L32

### Use case

We have a react frontend, and we want to subscribe to channels in each of the leaf components which require dynamic updates directly. Sometimes this means multiple components will be interested in the same updates, and subscribe to the same identifiers. Trying to deduplicate these subscriptions somehow requires a lot more state management which seems more fragile and uneccessary. Previous versions of actioncable before the guarantor seem to work great for this use case, and the actioncable implemenation seem to consider subscription duplication. It's the addition of the subscription guarantor in #41581 to fix #38668 which seems to have created this issue. I think it would continue to work great with some gentle adjustment.

### Expected behavior

Subscribing to the same channel identifier multiple times should not constantly send subscription messages.

### Actual behavior

[...]

PR #57504, opened 2026-05-29
Action Cable: make duplicate subscribe idempotent

samuel-williams-shopify:
Make Action Cable's handling of duplicate `subscribe` commands idempotent: when a client subscribes to a channel it is already subscribed to, the server re-transmits `confirm_subscription` and returns. The `subscribed` callback is not invoked again, no new subscription is registered, and the client's subscription guarantor receives a confirmation and resolves.

This replaces the current behaviour of raising `ActionCable::Connection::Subscriptions::AlreadySubscribedError` (introduced in the unreleased adapterization refactor in #50979 and not yet present in any tagged Rails release).

`Subscriptions::AlreadySubscribedError` is removed; the other error classes (`MalformedCommandError`, `ChannelNotFound`, `UnknownCommandError`, `UnknownSubscription`) are unaffected.

## Motivation

The right behaviour for duplicate `subscribe` has been an open question in Action Cable for nearly a decade. Multiple long-running issues describe the same underlying problem:

- **#24875 (2016)** "Cable: Should error on double-subscribe" — original debate; never resolved. @jeremy commented: *"Don't try to take action in an uncertain state. Kick an error back to the client."* Still open.
- **#44652 (2022)** "ActionCable: Repeated subscription attempts" — 25+ comments. Symptom: with the silent no-op, the client's subscription guarantor never receives `confirm_subscription` and retries every second forever. The reporter (@sj26) explicitly proposed *"send a subscription confirmation from the actioncable server when the client asks for a subscription which is already in place, instead of swallowing it silently."* That proposal is what this PR implements.
- **#48292 (2023)** "ActionCable sporadically fails to confirm subscribe because the subscription 'already exists'" — Turbo-specific reproduction; closed as stale.
- **#44653 (2022)** companion client-side workaround PR (still open).

A common theme across these reports is users either carrying a custom client-side patch or switching to AnyCable to escape the bug. The relevant code paths are now shared via the recent upstream sync, so the symptom now exists in stock Rails as well.

PR **#50979 ("Action Cable server adapterization", merged 2026-05-28)** changed the long-standing silent `return` to `raise AlreadySubscribedError`. That addresses #24875's request for an explicit signal, but it makes #44652 strictly worse: clients still don't receive the confirmation they need, and now there's also an unhandled exception propagating up the connection.

## Why idempotent re-transmit is the right answer

Realistic clients re-issue `subscribe` commands as a matter of normal operation, not as a programming error:

- Turbo's `<turbo-cable-stream-source>` is re-inserted into the DOM on morph/refresh cycles, which triggers a re-subscribe.
- React / Vue / Svelte components subscribe in lifecycle hooks; remounting re-subscribes.
- After a transient network issue, a client may resubscribe before the server has noticed the previous connection went away.
- The original race condition addressed by @palkan's #26547 in 2016 (which introduced the silent return in the first place): a deferred confirmation hadn't yet been sent when the client issued a follow-up command, so the client retried.

From the client's perspective, the meaningful response to *"please subscribe me to X"* when it is already subscribed is *"you are subscribed to X"* — exactly the `confirm_subscription` frame the client is waiting for. That's what this PR sends. It preserves the at-most-once semantics of `#subscribed` callbacks (no side-effect duplication) while resolving the client's state machine.

## Why removing `AlreadySubscribedError` is safe

The class was introduced by commit `7a8f26dcff` ("Improve exceptions in Action Cable subscriptions") inside PR #50979, which merged to `main` on 2026-05-28. The most recent tagged release is **v8.1.3** (2026-03-24), which predates the merge. Therefore `AlreadySubscribedError` has never appeared in a released Rails. No user code outside `rails/rails` main can be depending on its name today.

## Behavioural diff

|                                 | Before #50979 | After #50979 (current `main`) | This PR |
| ------------------------------- | :-----------: | :---------------------------: | :-----: |
| `subscriptions` Hash unchanged  | ✅            | ✅                            | ✅      |
| `#subscribed` runs only once    | ✅            | ✅                            | ✅      |
| Client receives a confirmation  | ❌ (#44652)   | ❌ + exception                 | ✅      |
| Exception propagates up         | ❌            | ✅ (`AlreadySubscribedError`)  | ❌      |

## Test

`actioncable/test/connection/subscriptions_test.rb` now asserts that resubscribing:

1. Does **not** invoke `#subscribed` a second time.
2. Transmits a fresh `confirm_subscription` to the client.
3. Leaves `subscriptions.identifiers.size` at 1.

Full ActionCable test suite passes (238 runs, 0 failures, 0 errors).

## Notes for reviewers

- The `Subscriptions::Error` base class and the remaining four error subclasses are unaffected.
- `MalformedCommandError` continues to be raised when `identifier` is missing (genuine protocol violation).
- The fix is purely server-side; clients (`@rails/actioncable`, ports, third parties) need no change. Existing `subscription.received` callbacks fire on the re-confirmation if registered, matching their behaviour for the initial confirmation.

### DIFF
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,3 +1,21 @@
+*   Make duplicate `subscribe` commands idempotent.
+
+    When a client sends a `subscribe` command for an identifier it is already
+    subscribed to, re-transmit `confirm_subscription` instead of raising
+    `Subscriptions::AlreadySubscribedError`. The `subscribed` callback is not
+    invoked a second time and no new subscription is registered. This
+    addresses long-standing reports (#24875, #44652, #48292) where clients
+    that legitimately re-issue subscribe commands (e.g. Turbo
+    `<turbo-cable-stream-source>` across morph/refresh cycles, React/Vue
+    components, reconnect loops) would either spin retrying because no
+    confirmation arrived, or, more recently, crash the connection on the
+    raised exception.
+
+    The `Subscriptions::AlreadySubscribedError` class is removed; it was
+    introduced in the unreleased adapterization refactor and never shipped.
+
+    *Samuel Williams*
+
 *   Respect calls to `#reject` in `before_subscribe` callbacks.
 
     It doesn't call `#subscribed` if a `before_subscribe` callback calls `#reject`.
--- a/actioncable/lib/action_cable/connection/subscriptions.rb
+++ b/actioncable/lib/action_cable/connection/subscriptions.rb
@@ -14,12 +14,6 @@ module ActionCable
     class Subscriptions # :nodoc:
       class Error < StandardError; end
 
-      class AlreadySubscribedError < Error
-        def initialize(identifier)
-          super "Already subscribed to #{identifier}"
-        end
-      end
-
       class ChannelNotFound < Error
         def initialize(channel_id)
           super "Channel not found: #{channel_id}"
@@ -64,7 +58,17 @@ module ActionCable
 
         raise MalformedCommandError, data unless id_key.present?
 
-        raise AlreadySubscribedError, id_key if subscriptions.key?(id_key)
+        # If the client re-subscribes to a channel it's already subscribed to,
+        # re-transmit the confirmation so the client's subscription guarantor
+        # can resolve. This is common with libraries that re-insert
+        # subscription tags into the DOM (e.g. Turbo's <turbo-cable-stream-source>
+        # across morph/refresh) and was the long-standing behaviour of
+        # ActionCable prior to its server adapterization refactor (see #44652,
+        # #48292, #24875).
+        if subscriptions.key?(id_key)
+          connection.transmit identifier: id_key, type: ActionCable::INTERNAL[:message_types][:confirmation]
+          return
+        end
 
         subscription = subscription_from_identifier(id_key)
 
--- a/actioncable/test/connection/subscriptions_test.rb
+++ b/actioncable/test/connection/subscriptions_test.rb
@@ -6,7 +6,7 @@ class ActionCable::Connection::SubscriptionsTest < ActionCable::TestCase
   class ChatChannelError < Exception; end
 
   class Connection < ActionCable::Connection::Base
-    attr_reader :exceptions
+    attr_reader :exceptions, :socket
 
     rescue_from ChatChannelError, with: :error_handler
 
@@ -72,13 +72,21 @@ class ActionCable::Connection::SubscriptionsTest < ActionCable::TestCase
     assert_empty @subscriptions.identifiers
   end
 
-  test "double subscribe command" do
+  test "double subscribe command re-transmits confirmation and is idempotent" do
     setup_connection
+    channel = subscribe_to_chat_channel
 
-    subscribe_to_chat_channel
+    # Resubscribing must not raise, must not invoke `#subscribed` a second
+    # time (which would re-run user code and double up side-effects like
+    # `stream_from`), must leave the subscription set unchanged, and must
+    # transmit another `confirm_subscription` so the client's subscription
+    # guarantor stops retrying.
+    confirmation = { identifier: @chat_identifier, type: ActionCable::INTERNAL[:message_types][:confirmation] }
 
-    assert_raises ActionCable::Connection::Subscriptions::AlreadySubscribedError do
-      subscribe_to_chat_channel
+    assert_not_called(channel, :subscribed) do
+      assert_called_with(@connection.socket, :transmit, [confirmation]) do
+        @subscriptions.execute_command "command" => "subscribe", "identifier" => @chat_identifier
+      end
     end
 
     assert_equal 1, @subscriptions.identifiers.size


samuel-williams-shopify:
@palkan can you please review, thanks!

samuel-williams-shopify:
The only thing I'm not really sure about is whether subscribe and unsubscribe are supposed to be matched for a given connection... or whether multiple subscribes followed by a single unsubscribe is sufficient to actually unsubscribe - e.g. multiple components subscribing, a single component unsubscribes, then no updates are received, etc.

palkan:
> This replaces the current behaviour of raising ActionCable::Connection::Subscriptions::AlreadySubscribedError

Ha, my next suggested move was to introduce a proper `"type":"error"` message to communicate the problem to the client 😁

IMO, a client sending double `subscribe` commands, or sending `subscribe` after receiving a confirmation, is a **misbehaving client**. We should fix it, not the server (that's what we did in [anycable-client](https://github.com/anycable/anycable-client)). And that's why I think introducing an error response is the right move.

You've already raised one concern regarding re-sending confirmations:

> whether multiple subscribes followed by a single unsubscribe is sufficient to actually unsubscribe

It should be sufficient; both the client and the server are assumed to have only one subscription for a given identifier (but there could be multiple Subscription objects (=event handlers) attached to a given subscription). However, think about race conditions (typical in Hotwire apps): `subscribe`-s and `unsubscribe`-s could be fired in any order, how to handle that? (I'm linearizing them per identifier in anycable-client).

Here is an example:

- A client has a confirmed subscription
- The client sends `unsubscribe` and `subscribe` one immediately after each other
- The server process them in separate threads
- The `#subscribed` callback finishes first—it's fast, it's just re-sending the confirmation
- The `#unsubscribed` callback took a bit longer (some state cleanup) and it finishes last—efficiently removing the server-side subscription
- The client receives the confirmation—all looks good! But in fact it's unsubscribed.

Without the re-transmission it would attempt to re-subscribe again and restore the state (though I think there could still be a room for race condition here, too).

> It preserves the at-most-once semantics of #subscribed callbacks (no side-effect duplication) while resolving the client's state machine.

What about the client-side semantics for `connected()` callbacks? Receiving multiple confirmations would result in double-execution of the connected callbacks (currently). **They should also be at-most-once, too**. There is a simple workaround: to add a hint to the confirmation message that this is a re-transmission (and handle it in the client). But I'd hesitate to extend the protocol in such a way. Or we can track the subscription state on the client side (I don't see us doing that) and ignore subsequent confirmations.

----

I chose raising an exception in the first place (in #50979) to highlight that such things should never happen if the client behaves correctly. This is an exception, not an error (semantically).

The real problem is the client, not the server.

----

> Turbo's <turbo-cable-stream-source> is re-inserted into the DOM on morph/refresh cycles, which triggers a re-subscribe.
  React / Vue / Svelte components subscribe in lifecycle hooks; remounting re-subscribes.
  After a transient network issue, a client may resubscribe before the server has noticed the previous connection went away.
  
To handle all of these, the client must be careful about managing state transitions and keeping track of the _intents_. We can't fix that server-side; we can just hack around some issues (though likely at the cost of introducing new ones).

ioquatix:
@palkan, that makes sense to me. If it's a client side issue, raising makes total sense.

In consideration of that, is something like https://github.com/rails/rails/pull/44653 the right way forward?

samuel-williams-shopify:
(Also I'll close this PR since we don't want to change this behaviour).

chaadow:
@palkan Hi, sorry if my question is naive or irrelevant, but I understand you've put all these client-side protections in [anycable-client](https://github.com/anycable/anycable-client), but if Rails ships with a javascript client through `action_cable` and that client currently is unsafe + you refactored ActionCable and decided to raise an error ( `Subscriptions::AlreadySubscribedError`) because that's the client's fault, why don't you port the safety measures that you've implemented in [anycable-client](https://github.com/anycable/anycable-client) on `rails`? 

Since this behavior (raise) didn't exist before the refactoring, I suspect old javascript clients will start encountering new issues once they upgrade to the next released rails version

palkan:
>  I suspect old javascript clients will start encountering new issues once they upgrade to the next released rails version

Not really; old clients aren't affected: no matter if we swallow double subscribes or raise, nothing changes from the client perspective—it just receives nothing. The exception is for developer to know about the problem and do something about it (you can always add `rescue_from Subscriptions::AlreadySubscribedError` and ignore it to restore the previous behaviour).

> why don't you port the safety measures that you've implemented in [anycable-client](https://github.com/anycable/anycable-client) on rails?

There is no straightforward way to port these measures; the most easiest solution would be to just replace `@rails/actioncable` with `@anycable/web` (maybe, a slim version of it to cover only Rails features and nothing more). I'd be happy to do that if I get a green light from the team (not sure who's responsible for Action Cable JS side right now?).  /cc @rafaelfranca @ioquatix

ioquatix:
> There is no straightforward way to port these measures

Why is that?

palkan:
> Why is that?

`anycable-client` has a very [different architecture](https://github.com/anycable/anycable-client/blob/master/docs/architecture.md) (multiple components with state machines etc.), heavily relies on async/await for the control flow, comes with tons of tests to cover all the known edge cases regarding subscriptions—porting that would mean rewriting most of the Action Cable JS code. Here is, for example, how subscribing works (to prevent various race conditions): https://github.com/anycable/anycable-client/blob/599c02216687b1152a1513e0db8b00050c1310ea/packages/core/cable/index.js#L354-L486

It hasn't started like that. This complexity emerged from (poor) user experience we had to fix. Porting just some bits would be like putting yet another band aid (like the current guarantor). Porting everything (the logic behind it) is possible, for sure. But it's not a low effort task, and it carries the risks of missing something. That's what I meant by "not straightforward". Rewriting from scratch based on the `anycable-client` foundation looks more straightforward (to me).

palkan:
@samuel-williams-shopify FYI I've started working on the client improvements (better guarantor/locking) in the background; will open a PR in ~2 weeks (I'm OOO and traveling the next one and this one short due to 🇺🇸, so...). Stay tuned!

PR #57507, opened 2026-05-29
Fix has_delegated_json not applying normalizations on assignment.

### LINKED ISSUE #57464: has_delegated_json does not work with normalization
### Steps to reproduce

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # gem "rails"

  gem "sqlite3"
end

require "minitest/autorun"

  config.eager_load = false

end
Rails.application.initialize!

    t.string :content
    t.json :stuff
  end
end

  has_delegated_json :stuff, name: :string

  normalizes :name, with: -> attr { attr.strip }
  normalizes :content, with: -> attr { attr.strip }
end

  def test_stuff
    post = Post.create! name: " clean me please ", content: " this one cleans up "

    assert_equal "this one cleans up", post.content
    assert_equal "clean me please", post.name
  end
end
```

Normalization does not work with has_delegated_json.
From developer point of view the expectation would be to have this combination of features work with each other.

### System configuration

**Rails version**: main

**Ruby version**: 4.0.5

ruyrocha:
Fix #57464 

### Motivation / Background

This Pull Request has been created because #57464

### Detail

Apply normalization and write back attribute on has_delegated_json assignment.

This Pull Request applies normalization and write back attribute on `has_delegated_json` assignment.

### Additional information

The existing `SchematizedJsonTest` uses a plain `ActiveModel` object without `Normalization` included; the `respond_to?` guard ensures those models continue to work unchanged.

### Checklist

### DIFF
--- a/activemodel/CHANGELOG.md
+++ b/activemodel/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   `has_delegated_json` now applies `normalizes` declarations on assignment.
+
+    Fix #57464
+
+    *Ruy Rocha*
+
 *   Limit the size of strings `ActiveModel::Type::Integer` will coerce with `to_i`.
 
     Calling `to_i` on very long strings can take a long time and could be used as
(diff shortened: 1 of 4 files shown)


rafaelfranca (changes requested):
`has_delegated_json` should just be changed to use ActiveModel::Attribute APIs

PR #57508, opened 2026-05-29
Make Ractor shareability methods only available on 4.0 and above.

andrewn617:
### Motivation / Background

Follow up to https://github.com/pulls?q=is%3Apr+author%3Aandrewn617+archived%3Afalse+is%3Aclosed

I realized that this is problematic for Ruby 3.x since `Ractor.make_shareable` is available but not `Ractor.shareable_proc`. 

Since we will define a shim for `make_shareable` but `shareable_proc` will be a no-op in that environment, it will lead us to calling `make_shareable` on objects containing unshareable procs and raising.

```
proc = ractor_shareable_proc { foo }

ractor_make_shareable(Bar.new(proc)) # => raises on 3.4 but works on 4.0
```

### Detail

Ruby 4.0 will be a hard requirement for using Ractors with Rails, so I think it is okay if we just define all these methods if we are on 4.0 and above, and have them all be no-ops on Ruby 3.x

### Checklist

### DIFF
--- a/activesupport/CHANGELOG.md
+++ b/activesupport/CHANGELOG.md
@@ -1,9 +1,8 @@
 *   Add shims for `Ractor` shareability methods so framework code can call them
     unconditionally regardless of the Ruby version.
 
-    When `Ractor` is not defined, or the underlying method is not available, the
-    shim is a no-op that simply returns its argument (or the given block).
-    Otherwise the call is forwarded to the matching `Ractor` class method.
+    On Ruby 4.0 and above, these methods delegate to the corresponding +Ractor+ methods. On older
+    Ruby versions, these methods are no-ops that return their argument.
 
     ```ruby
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/key_generator.rb
```diff
@@ -62,9 +62,18 @@ def initialize(key_generator)
       @cache_keys = Concurrent::Map.new
     end
 
+    def freeze
```
gmcgibbon: This seems unrelated. Maybe we can do this in another PR?

andrewn617: Urgh sorry about that, yes something from another branch I have on the go my bad. Removed.

byroot:
FYI: https://github.com/rails/rails/pull/57526

PR #46519, opened 2022-11-18
AbstractAdapter: only synchronize when necessary

casperisfine:
Ref: https://github.com/rails/rails/pull/28083

Right now all database operations are synchronized with a monitor even though it's useless in the vast majority of cases.

The synchronization is only useful when transactional fixtures are enabled.

In production, connections are never shared across threads, as such this extra synchronization is wasteful.

### Motivation

Well, there is obviously the performance overhead of that synchronization.

But beyond this, what prompted me to work on this is that I think it may help with https://github.com/rails/rails/issues/45994. If the lock is only used in test, it's much less of a problem to use a slower lock implementation, which open the door to use a pure ruby monitor that is owned by `Thread` rather than `Fiber`, allowing us to bring back the original behavior.

cc @matthewd @eileencodes, I'd love your thoughts on this.
FYI: @ngan @technicalpickles

PR #57517, opened 2026-05-30
Fixes #51780: prevent shared PostgreSQL queries from racing with reconnects

### LINKED ISSUE #51780: Possible race condition in postgresql type_map initialization

oneearedrabbit:
### Motivation / Background

abstract_adapter.rb with default settings has a race condition. Rails 7.1 introduced a regression in https://github.com/rails/rails/pull/46519 by assuming that connections are never shared across threads. In practice, that assumption doesn't always hold, especially in applications that use a connection pool together with RDS Proxy. PostgreSQLAdapter requires a lock since type_map is initialized lazily, a change made in https://github.com/rails/rails/pull/44570.

Cross-referencing https://github.com/rails/rails/issues/51780#issuecomment-4583353372.

### Detail

Add a `connection_lock` database config for applications that share a connection across threads or fibers. When enabled, ActiveRecord locks the adapter during reconnect and PostgreSQL result processing so a shared connection cannot be used while it is being reconnected.

I noticed that some type_map readers were not guarded, I added `@lock.synchronize` in a few additional places. I am leaving this to your best judgement since that work ballooned beyond connection_pool, abstract_adapter, and postgresql_adapter classes, and I am less familiar with database_statements class.

### Additional information

Repro of the original issue:
```ruby
$ cat Gemfile
ruby "3.3.4"
gem "activerecord", "=8.1.3"

$ cat database.yml
development:
  adapter: postgresql
  url: postgres://localhost:5432/giza_development
  database: giza_development
  username: postgres
  gssencmode: disable
  pool: 5

$ cat repro.rb
require "active_record"
require "yaml"
config = YAML.load(File.read("database.yml"))["development"]
ActiveRecord::Base.establish_connection(config)
Thread.abort_on_exception = true
ActiveRecord::Base.connection_pool.with_connection do |connection|
  Thread.new(connection) do |pg|
    pg.reconnect!
  end
  p connection.select_values("SELECT 1")
end

$ irb -r ./repro.rb
/opt/homebrew/lib/ruby/gems/3.3.0/gems/activerecord-8.1.3/lib/active_record/connection_adapters/postgresql_adapter.rb:876:in `get_oid_type': undefined method `key?' for nil (NoMethodError)

          if !type_map.key?(oid)
                      ^^^^^
```

After the following patch is applied, the error is no longer reproducible:
```ruby
module ActiveRecord::ConnectionAdapters::PostgreSQLAdapterPatch
  def initialize(
    config_or_deprecated_connection,
    deprecated_logger = nil,
    deprecated_connection_options = nil,
    deprecated_config = nil
  )
    super
    @lock = ActiveSupport::Concurrency::ThreadMonitor.new
  end
end

class ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
  prepend ActiveRecord::ConnectionAdapters::PostgreSQLAdapterPatch
end
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Add a `connection_lock` database configuration for applications that share
+    a connection across threads or fibers. When enabled, it prevents
+    PostgreSQL queries from racing with reconnects on that shared connection.
+
+    *oneearedrabbit*
+
 *   Fix deadlock when pool-less connection materializes while fetching database
     server version.
 
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -373,7 +373,7 @@ module ActiveRecord
           @connections << @pinned_connection
         end
 
-        @pinned_connection.lock_thread = ActiveSupport::IsolatedExecutionState.context if lock_thread
+        @pinned_connection.install_execution_context_lock(ActiveSupport::IsolatedExecutionState.context) if lock_thread
         @pinned_connection.pinned = true
         @pinned_connection.begin_transaction joinable: false, _lazy: false
       end
@@ -398,7 +398,7 @@ module ActiveRecord
           if @pinned_connection.nil?
             connection.pinned = false
             connection.steal!
-            connection.lock_thread = nil
+            connection.restore_configured_connection_lock
             checkin(connection)
           end
         end
--- a/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb
@@ -178,7 +178,8 @@ module ActiveRecord
         @allow_preconnect = false
         @visitor = arel_visitor
         @statements = build_statement_pool
-        self.lock_thread = nil
+        @configured_connection_lock = @config[:connection_lock]
+        @lock = build_adapter_lock(effective_connection_lock)
 
         @prepared_statements = !ActiveRecord.disable_prepared_statements && self.class.type_cast_config_to_boolean(
           @config.fetch(:prepared_statements) { default_prepared_statements }
@@ -204,12 +205,50 @@ module ActiveRecord
         "#<#{self.class.name}:#{'%#016x' % (object_id << 1)} env_name=#{pool.db_config.env_name.inspect}#{name_field} role=#{role.inspect}#{shard_field}>"
       end
 
-      def lock_thread=(lock_thread) # :nodoc:
-        @lock =
-        case lock_thread
-        when Thread
+      def install_execution_context_lock(lock_context) # :nodoc:
+        @lock = build_adapter_lock(effective_connection_lock(lock_context))
+      end
+
+      def restore_configured_connection_lock # :nodoc:
+        @lock = build_adapter_lock(effective_connection_lock)
+      end
+
+      private def effective_connection_lock(lock_context = nil)
+        configured_lock =
+          case self.class.type_cast_config_to_boolean(@configured_connection_lock)
+          when nil, false
+            nil
+          when :thread, "thread"
+            :thread
+          when :monitor, "monitor"
+            :monitor
+          else
+            raise ArgumentError, "connection_lock must be one of nil, false, :thread, or :monitor"
+          end
+
+        return configured_lock unless lock_context
+
+        requested_lock =
+          case lock_context
+          when Thread
+            :thread
+          when Fiber
+            :monitor
+          else
+            raise ArgumentError, "lock_context must be a Thread or Fiber"
+          end
+
+        return :monitor if configured_lock == :monitor || requested_lock == :monitor
+        return :thread if configured_lock == :thread || requested_lock == :thread
+
+        nil
+      end
+
+      private def build_adapter_lock(lock_mode)
+        case lock_mode
+        when :thread
           ActiveSupport::Concurrency::ThreadMonitor.new
-        when Fiber
+        when :monitor
           ::Monitor.new
         else
           ActiveSupport::Concurrency::NullLock
--- a/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
@@ -15,7 +15,10 @@ module ActiveRecord
           intent = internal_build_intent(sql, name, allow_retry:, materialize_transactions:)
           intent.execute!
           result = intent.raw_result
-          result.map_types!(@type_map_for_results).values
+
+          @lock.synchronize do
+            result.map_types!(@type_map_for_results).values
+          end
         end
 
         READ_QUERY = ActiveRecord::ConnectionAdapters::AbstractAdapter.build_read_query_regexp(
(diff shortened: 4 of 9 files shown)


jondavidschober:
+1 for this. I’ve been hesitant to upgrade rails because this was such an impactful regression for us

skipkayhil:
> by assuming that connections are never shared across threads

This is a valid assumption though, you _should not_ be sharing connections between threads.

> I’ve been hesitant to upgrade rails because this was such an impactful regression for us

Have you followed the progress in the linked issue? There were many fixes applied/released to address the `type_map` issue.

oneearedrabbit:
@skipkayhil I don't disagree :-) My point is that there is a known race condition in ActiveRecord/PostgreSQLAdapter which `lock_thread` was originally meant to guard against. Apparently, there were quite a few reports showing that the shared connections contract isn't always followed, and with `NullLock` that became a weak spot which led to #51780.

Conceptually, there are two possible directions:
1. If connections are truly never shared, then we should brainstorm on enforcing the contract and making violations easier to detect.
2. If connections are not supposed to be shared, but in practice they are, whether intentionally or not, then `lock_thread` should be configurable from application code.

Right now, the implementation is at midpoint: the mechanism exists, but users can neither configure it nor easily diagnose underlying issues.

Option 1 is essentially "not a bug; fix the vendor libraries", which probably also means adding better diagnostics so users can find culprits more easily, while Option 2 is a more practical and user-friendly compromise.

At the very least, I do hope we can agree that a race condition itself is undesirable regardless of any external factors.

byroot:
> you _should not_ be sharing connections between threads.

Yep.

> making violations easier to detect.

So you mention RDS proxy being a cause, which seems weird to me because I don't see what a proxy would change in term of in process concurrent accesses.

Have you yourself ran a modified version of the adapter that raises on concurrent access to try to find the root cause?

oneearedrabbit:
Writing this here down for archival purposes and for any future spelunkers, since I haven't yet been able to isolate the issue cleanly in our environment:
1. Rails 8.0.3 to Rails 8.0.5 upgrade -> no significant changes observed.
2. Rails 8.0.5 to Rails 8.1.3 upgrade -> escalated both `TypeMap` and `get_oid_type` errors. I'm providing stack traces below, but these are side-effects and don't seem by themselves explain much.
3. Reduced max_age -> escalated both errors again.
4. pg gem 1.5.8 to 1.6.3 upgrade -> no changes.

Some additional application specific notes:
1. Long-running Temporal activities
2. We applied two patches to the PostgreSQLAdapter to avoid connection pinning in RDS Proxy, https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html:
    - deleted DISCARD ALL in `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter#reset!`
    - replaced SET statements in `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter#configure_connection`

I was unable to identify any cross-thread shared database connections through static analysis beyond the standard Rails mechanism.

After we changed the `max_age` setting, my working hypothesis was the reaper thread marks a connection as stale and ultimately calls "connect", which instantiates a new `@raw_connection` instance and therefore `@type_map`, creating a small window in which the PG connection is in an incomplete state. That was a stretching theory, and I am unable to prove it by using standard interfaces. The earlier repro mechanically called `reconnect!`, and I am not confident I fully understand the conditions under which the issue manifests beyond what's been said already.

We continue to investigate the issue.

Stack traces:

<details>
<summary>Stack trace for the TypeMap error</summary>
<pre>
lib/pg/result.rb:17:in `type_map=': wrong argument type nil (expected PG::TypeMap) (TypeError)
	from lib/pg/result.rb:17:in `map_types!'
	from postgresql/database_statements.rb:16:in `query'
	from connection_adapters/postgresql_adapter.rb:1058:in `column_definitions'
	from abstract/schema_statements.rb:110:in `columns'
	from connection_adapters/schema_cache.rb:345:in `block (2 levels) in columns'
	from abstract/connection_pool.rb:457:in `with_connection'
	from connection_adapters/schema_cache.rb:344:in `block in columns'
	from connection_adapters/schema_cache.rb:343:in `fetch'
	from connection_adapters/schema_cache.rb:343:in `columns'
	from connection_adapters/schema_cache.rb:354:in `block in columns_hash'
	from connection_adapters/schema_cache.rb:353:in `fetch'
	from connection_adapters/schema_cache.rb:353:in `columns_hash'
	from connection_adapters/schema_cache.rb:54:in `columns_hash'
	from connection_adapters/schema_cache.rb:198:in `columns_hash'
	from active_record/model_schema.rb:609:in `load_schema!'
	from active_record/counter_cache.rb:212:in `load_schema!'
	from encryption/encryptable_record.rb:127:in `load_schema!'
	from active_record/model_schema.rb:555:in `block in load_schema'
	from active_record/model_schema.rb:552:in `synchronize'
	from active_record/model_schema.rb:552:in `load_schema'
	from active_record/model_schema.rb:444:in `columns_hash'
	from active_record/attributes.rb:257:in `block in _default_attributes'
	from abstract/connection_pool.rb:463:in `with_connection'
	from active_record/connection_handling.rb:313:in `with_connection'
	from active_record/attributes.rb:256:in `_default_attributes'
	from active_model/attribute_registration.rb:38:in `attribute_types'
	from active_record/attribute_methods.rb:261:in `_has_attribute?'
	from active_record/inheritance.rb:61:in `new'
</pre>
</details>

<details>
<summary>Stack trace for the get_oid_map error</summary>
<pre>
connection_adapters/postgresql_adapter.rb:876:in `get_oid_type': undefined method `key?' for nil (NoMethodError)
	from postgresql/database_statements.rb:188:in `block in cast_result'	
from <internal:numeric>:237:in `times'
	from postgresql/database_statements.rb:185:in `cast_result'
	from abstract/database_statements.rb:555:in `internal_exec_query'
	from abstract/database_statements.rb:708:in `select'
	from abstract/database_statements.rb:76:in `select_all'
	from abstract/query_cache.rb:275:in `block in select_all'
	from abstract/query_cache.rb:313:in `block (2 levels) in cache_sql'
	from abstract/query_cache.rb:90:in `compute_if_absent'
	from abstract/query_cache.rb:311:in `block in cache_sql'
	from concurrency/null_lock.rb:9:in `synchronize'
	from abstract/query_cache.rb:310:in `cache_sql'
	from abstract/query_cache.rb:275:in `select_all'
	from active_record/querying.rb:68:in `_query_by_sql'
	from active_record/querying.rb:53:in `block in find_by_sql'
	from abstract/connection_pool.rb:457:in `with_connection'
	from active_record/connection_handling.rb:313:in `with_connection'
	from active_record/querying.rb:52:in `find_by_sql'
	from active_record/statement_cache.rb:156:in `execute'
	from active_record/core.rb:455:in `block in cached_find_by'
	from abstract/connection_pool.rb:463:in `with_connection'
	from active_record/connection_handling.rb:313:in `with_connection'
	from active_record/core.rb:443:in `cached_find_by'
	from active_record/core.rb:278:in `find'
</pre>
</details>

PR #41397, opened 2021-02-10
Enable scoping to apply to all queries

eileencodes:
Similar to https://github.com/rails/rails/pull/40720
and https://github.com/rails/rails/pull/40805 this change allows for the
`scoping` method to apply to all queries in the block. Previously this
would only apply to queries on the class and not the instance. Ie
`Post.create`, `Post.all`, but not `post.update`, or `post.delete`.

The change here will create a global scope that is applied to all
queries for a relation for the duration of the block.

Benefits:

This change allows applications to add a scope to any query for the
duration of a block. This is useful for applications using sharding to
be able to control the query without requiring a `default_scope`. This
is useful if you want to have more control over when a `scoping` is used
on a relation. This also brings `scoping` in parity with the behavior of
`default_scope` so there are less surprises between the behavior of
these two methods.

There are a caveats to this behavior:

1) The `scoping` only applies to objects of the same type. IE you cannot
scope `Post.where(blog_id: 1).scoping` and then expect `post.comments`
will apply `blog_id = 1` to the `Comment` query. This is not possible
because the scope is `posts.blog_id = 1` and we can't apply the `posts`
scope to a `comments` query. To solve this, scopes must be nested.
2) If a block is scoped to `all_queries` it cannot be unscoped without
exiting the block. I couldn't find a way around this but ActiveRecord
scoping is a bit complex and turning off `all_queries` when it's already
on in nested scoping blocks had interesting behavior that I decided was
best left out.

PR #57518, opened 2026-05-30
Expose `previously_affected_rows` on Active Record instances after create/update/destroy events and `scoping(all_queries: true)` to allow granular alternative to `Locking::Optimistic`

bensheldon:
### Motivation / Background

This PR extracts the `affected_rows` checking from `Locking::Optimistic` into a generally useful reader `previously_affected_rows`.

Being able to access `previously_affected_rows` at the instance level unlocks the ability to write granular optimistic locking behavior using the more-recently added `scoping(all_queries)` (#41397) behavior e.g.

```ruby
post = Post.last

old_body = post.body
Post.where(body: old_body).scoping(all_queries: true) do
  record.update(body: “new_body”)
end

oops_the_body_already_changed! if previously_affected_rows.zero? 
```

This is different from `Locking::Optimistic` which only allows for a single change counter for the entire record.

While it’s possible to do something similar using `update_all` I really wanted to be able to use this behavior at the AR instance level.

### Detail

This change extracts the row-counting behavior that exists in `Locking::Optimistic` into AR `Persistence` and refactors/simplifies `Locking::Optimistic` to use the new behavior.

### Additional information

- I exposed `previously_affected_rows` to the public interface of an AR record. I could also imagine it _only_ being populatd within the context of an `around_save`. That would add a little more work to express usable behavior, but no objection if we wanted to limit it
- I think `previously_affected_rows` is a _consistent_ name with analogous functionality (`previously new_record?`), but it does slightly leak the abstraction of “affected_rows” . Naming is hard.

### Checklist


activerecord/test/cases/scoping/relation_scoping_test.rb
```diff
@@ -427,6 +427,31 @@ def test_nested_scoping_applies_with_all_queries_set
     end
   end
 
+  def test_previously_affected_rows_detects_scoped_update_match
+    post = Post.first
+    old_body = post.body
+
+    Post.where(body: old_body).scoping(all_queries: true) do
```
byroot, line 434:
```suggestion
    Post.where(id: post.id, body: old_body).scoping(all_queries: true) do
```

I know it's just a test so doesn't matter for correctness, but just so I'm sure I understand the intent, you'd normally have the `id` in `where` right?

activerecord/test/cases/scoping/relation_scoping_test.rb
```diff
@@ -427,6 +427,31 @@ def test_nested_scoping_applies_with_all_queries_set
...
+
+  def test_previously_affected_rows_detects_scoped_update_miss
+    post = Post.first
+    old_body = post.body
+
+    # Another process changes the body out from under us.
+    Post.where(id: post.id).update_all(body: "changed elsewhere")
+
+    Post.where(body: old_body).scoping(all_queries: true) do
+      assert post.update(body: "a brand new body"), "update still reports success"
```
byroot, line 449: > update still reports success

Should it though?

I think this is an interesting feature request, as that's something I smei-recently do, just with `Post.where(...).update_all(body: "new body) == 1`, but I see how it would be convenient to go though the model layer.

But it also feel likes a needlessly complicated interface to return an integer when in reality what you want to know is whether the update changed something, so basically just a boolean.

Changing the behavior of `#update` might be tricky though, but if we have to I'd prefer to introduce another method (or keyword param) than to persist `previously_affected_rows`.

activerecord/test/cases/scoping/relation_scoping_test.rb
bensheldon: No. Even in practice I wouldn't because I'd expect the `where(id: ...)` to come from the `update` method and be merged with the outer `scoping` query.

byroot: Right, my bad.

activerecord/test/cases/scoping/relation_scoping_test.rb
bensheldon: >> update still reports success
>
> Should it though?

That's a much spicier change 😆 The return value `save`/`update` seems to be "return true if valid and not aborted" regardless of what actually happens with the database.

There's 3 reasons why an update might not affect any rows, but still return true:

1. There are no value changes to the record
1. The row was previously deleted, which seems to be a [longstanding unexpectedness](https://stackoverflow.com/questions/73662052/no-error-on-updating-deleted-object-in-rails) with the interface.
2. The (relatively newish, Rails 7.0) interaction with `scoping(all_queries: true)`

The bulk of that seems to live in `#_update_record`:

https://github.com/rails/rails/blob/e2f7b51035a60a4955509145f15ca65d7c0f7af5/activerecord/lib/active_record/persistence.rb#L976-L982

...and from that, you can imagine that it becomes hard to distinguish which situation is being dealt with.

Anyways, I'll do a revision with a boolean instead of the integer 👍🏻

byroot: > ou can imagine that it becomes hard to distinguish which situation is being dealt with.

Indeed, but I don't think it's much of a problem.

`#update` specifically we certainly shouldn't change, because its return value is frequently checked expecting validation errors:

```ruby
if @post.update(params...)
  render :ok 
else
  render @post.errors
end
```

But I could be convinced to add another method with that behavior, or a keyword argument. My suggestion would be `#update?`, but might not be descriptive enough.

Perhaps `optimistic_update` or something like that? Could even take care of adding the modified attributes to the query.

bensheldon: I'd like the same predicate to cover `save`, `update`, `increment/decrement`, `touch`, and `destroy`.

I don't think `scoping(all_queries: true)` works analogously on record creation (I think it adds the where queries as VALUEs rather than adding WHERE)

bensheldon:
@byroot I changed the method to a predicate: `#previously_affected_row?`

For example, this is how I'd imagine writing a platinum-plated controller with this feature, to enforce that no unexpected changes were made to a Post between it being edited/reviewed and action being taken. I'll admit that it's a little twisty in that it exposes the sharp edges of the current lifecycle methods, but I'm happy that it makes it possible to handle the edges:

```ruby
class PostsController < ApplicationController
  def update
    post = Post.find(params[:id].to_i)
    post_params = params.expect post: [:previous_body, :body] # previous_body => post.body in hidden form field at time of form render
    previous_body = post_params.delete(:previous_body)

    valid_save = Post.where(body: previous_body).scoping(all_queries: true) do
      post.update(post_params)
    end

    if valid_save
      if !post.saved_changes?
        redirect_to post_path(post), notice: "No changes made", status: :see_other
      elsif post.previously_affected_row? # <-- new feature
        redirect_to post_path(post), notice: "Post successfully updated", status: :see_other
      else
        edited_body = post.body
        post.reload
        post.errors.add(:body, "Body already modified. Your rejected change: #{edited_body}")
        render :edit, status: :unprocessable_entity
      end
    else
      render :edit, status: :unprocessable_entity
    end
  end

  def destroy
    post = Post.find(params[:id].to_i)
    previous_body = params.dig(:post, :previous_body) # post.body in hidden form field at time of form/button render

    valid_destroy = Post.where(body: previous_body).scoping(all_queries: true) do
      post.destroy
    end

    if valid_destroy
      if post.previously_affected_row? # <-- new feature
        redirect_to posts_path, notice: "Post successfully deleted", status: :see_other
      else
        redirect_back fallback_location: post_path(post), notice: "Post modified; review changes and perform delete again", status: :see_other
      end
    else
      redirect_back fallback_location: post_path(post), notice: "Destroy failed: #{post.errors.full_messages.join(', ')}", status: :see_other
    end
  end
end
```

byroot:
I understand where you are coming from, but this feel too low-level and tacked on for me.
I think your use case should be supported, but I don't think `#previously_affected_row?` is the answer to "What would Rails do?"

PR #57526, opened 2026-06-01
Revert adding ractor helpers in Kernel

byroot:
Followup: https://github.com/rails/rails/pull/57467

I don't think adding these helpers in Kernel and making them Rails public API is a good idea. It means we'll have to keep them essentially forever even though the given intent is to simplify Rails own implementation.

Realistically once Rails drop support for older Ruby version we'll no longer need these helpers, hence we should expose them in Kernel and in our public API.

Instead we can use an internal module. We can debate whether it should be purely singleton methods, or `module_method` so you can include it.


rafaelfranca (approved):
I mean, I'm the first one to be against new additions to core extensions, but I do think those methods should be in kernel or have special syntax is Ruby itself already. 

But, I'm ok with those being private right now until it is clear they need to be used in applications and we have no good alternatives.

byroot:
> I'm ok with those being private right now until it is clear they need to be used in applications and we have no good alternatives.

Yes, If there is a good case for exposing those then I'm fine with it. Here I was mostly weirded out by these being added before there was any use.

andrewn617:
Alright fair. I have some PRs in the works using these, instead I will use a module.

PR #57527, opened 2026-06-01
Fix syntax for Rack response in Rack guide

ayushn21:
### Motivation / Background

There's a small mistake in the syntax to define a Rack response in the Rack guide.

### Detail

This PR corrects the syntax to use a `Rack::Response` object as required by Rails, rather than an array.

### Checklist


p8:
Thanks @ayushn21 . I've tested that this indeed results in a correct response.

PR #57532, opened 2026-06-01
Fix grouped calculations by a `belongs_to` association with a composite primary key model

55728:
### Summary

Grouping a calculation by a `belongs_to` association that points to a composite primary key model raises `ArgumentError` instead of returning the grouped result:

```ruby
class Order < ApplicationRecord
end

  belongs_to :order, foreign_key: [:shop_id, :order_id]
end

Book.group(:order).count
# => ArgumentError: Expected corresponding value for ["shop_id", "id"] to be an Array
```

The single-column equivalent (`Comment.group(:post).count`, keyed by the associated records) has always worked; only the composite path is broken — it raises outright, so the feature is entirely unavailable for composite-key associations.

### Cause

When a grouped calculation is grouped by a `belongs_to` association, `execute_grouped_calculation` loads the associated records to key the result by them:

```ruby
key_ids     = calculated_data.collect { |row| row[group_aliases.first] }
key_records = association.klass.base_class.where(association.klass.base_class.primary_key => key_ids)
```

For a composite key, the grouped columns are the multiple foreign key components (`group_fields = Array(association.foreign_key)`), but `key_ids` only collects `group_aliases.first` — a single column. `primary_key` is then an array of column names, so `where(["shop_id", "id"] => [single, single, …])` hands the predicate builder a flat list of scalars where it expects an array of tuples, and it raises `ArgumentError`.

### Fix

When the associated model has a composite primary key, collect the full tuple of grouped columns per row, matching the value shape the predicate builder (and `index_by(&:id)`, which already returns the composite id tuple) expects. The single-column branch is unchanged.

### Tests

Added a regression test (`calculations_test.rb`) that groups `count` by a composite-key `belongs_to` and asserts the result is keyed by the associated records — **verified red on `main` (the `ArgumentError`), green with the fix** on both **sqlite3** and **postgresql**. Full `calculations_test.rb` passes on both adapters.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,13 @@
+*   Fix grouped calculations (e.g. `count`) grouped by a `belongs_to` association
+    that points to a composite primary key model.
+
+    `Book.group(:order).count`, where `Book belongs_to :order` and `Order` has a
+    composite primary key, raised `ArgumentError: Expected corresponding value
+    for ["shop_id", "id"] to be an Array`. The grouped result is now keyed by the
+    associated records as it already is for single-column keys.
+
+    *Kenta Ishizaki*
+
 *   Fix `has_many`/`has_one :through` associations with `disable_joins: true`
     silently returning an empty result when the source points to a composite
     primary key model.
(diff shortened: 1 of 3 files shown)


rafaelfranca (approved):

55728:
Thank you for the review 😊 Resolved the conflicts.

PR #57535, opened 2026-06-02
Fix flaky ActiveRecord concurrency tests on Trilogy.

ruyrocha:
### Motivation / Background

This PR fixes flaky ActiveRecord concurrency tests that fail under MySQL/Trilogy with `ActiveRecord::StatementInvalid: Record has changed since last read` (MySQL ER_CHECKREAD, error 1020). The failures occur because:

1. `CreateOrFindByWithinTransactions#duel` wraps `find_or_create_by` in an outer transaction, which on MySQL REPEATABLE READ creates a SAVEPOINT scenario that `find_or_create_by` was never designed to handle under concurrent load.
2. `ConcurrentTransactionTest` writer threads hit transient ER_CHECKREAD errors when multiple transactions read and write the same row concurrently, but the tests had no retry logic.
3. Threads used `.lease_connection.close` to release connections, which is indirect and can leak connections on exception paths.

### Detail

**`CreateOrFindByWithinTransactions#duel` (`relations_test.rb`)**
- Remove the outer `Subscriber.transaction` wrapper from both threads so `find_or_create_by` manages its own top-level transaction. This fixes the "SAVEPOINT does not exist" error under MySQL REPEATABLE READ.
- Add `ensure` blocks with `Subscriber.connection_pool.release_connection` to prevent connection leaks if `yield` raises.
- Remove the misleading `# premature read for REPEATABLE READ` comment — the `find_by` is now outside any transaction and serves only as a timing delay.

**`ConcurrentTransactionTest` (`transactions_test.rb`)**
- Add retry logic (up to 3 attempts) for transient MySQL ER_CHECKREAD (1020) errors in both writer and observer threads.
- Introduce `mysql_checkread_error?(exception)` helper that matches on the underlying adapter error number (`error_number`, `errno`, or `error_code`) instead of fragile string matching on the error message.
- Add `rescue ActiveRecord::SerializationFailure` as a forward-compatible path for when PR #57568 lands (which will classify ER_CHECKREAD as `SerializationFailure`). The `StatementInvalid` fallback with `mysql_checkread_error?` remains for backward compatibility.
- Replace `.lease_connection.close` with `.connection_pool.release_connection` in `ensure` blocks so connections are released even when retries are exhausted.
- Writer threads **raise** after exhausting retries — they must complete their work.
- Observer thread **swallows** exhausted retries and continues its sampling loop, then asserts `successful_reads > 0` to prevent a vacuous pass where the test passes without ever verifying the salary invariant.

### Additional information

```shell
Finished in 105.048636s, 91.5100 runs/s, 310.8846 assertions/s.
9613 runs, 32658 assertions, 0 failures, 0 errors, 46 skips

You have skipped tests. Run with --verbose for details.
Using trilogy
Run options: --seed 41115

# Running:

..............................

Finished in 0.444074s, 67.5563 runs/s, 229.6914 assertions/s.
30 runs, 102 assertions, 0 failures, 0 errors, 0 skips
```

### Checklist

### DIFF
--- a/activerecord/test/cases/relations_test.rb
+++ b/activerecord/test/cases/relations_test.rb
@@ -2574,36 +2574,34 @@ class CreateOrFindByWithinTransactions < ActiveRecord::TestCase
     end
 
     private
-      def duel
-        assert_nil Subscriber.find_by(nick: "bob")
+    def duel
+      assert_nil Subscriber.find_by(nick: "bob")
 
-        a_wakeup = Concurrent::Event.new
-        b_wakeup = Concurrent::Event.new
+      a_wakeup = Concurrent::Event.new
+      b_wakeup = Concurrent::Event.new
 
-        a = Thread.new do
-          Subscriber.transaction do
-            a_wakeup.wait
-            yield
-            b_wakeup.set
-          end
-        end
-
-        b = Thread.new do
-          Subscriber.transaction do
-            # Read the record prematurely for MySQL REPEATABLE READ to kick in
-            Subscriber.find_by(nick: "bob")
-
-            a_wakeup.set
-            b_wakeup.wait
-            yield
-          end
-        end
-
-        a.join
-        b.join
-
-        assert_equal 1, Subscriber.where(nick: "bob").count
+      a = Thread.new do
+        a_wakeup.wait
+        yield  # find_or_create_by runs its own top-level transaction
+        b_wakeup.set
+      ensure
+        Subscriber.connection_pool.release_connection
       end
+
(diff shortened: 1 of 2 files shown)


ruyrocha:
ping @byroot @matthewd @rafaelfranca

yahonda:
Could you share a link to a rails CI (Buildkite) job where these tests actually
failed, plus the exact command that reproduces it
(`ARCONN=... bin/test ... --seed ...`), not just the seed? I don't recall seeing
this fail on CI, though I may have missed it.

PR #57536, opened 2026-06-02
Fix handling of `precision` in timestamp columns in Postgres migrations

mackuba:
### Motivation / Background

While migrating my app from SQLite to Postgres some time ago, I've noticed some inconsistencies in how the precision is set in timestamp columns depending on how they're declared in the migration DSL. I made an example repo here that demonstrates the problem: https://github.com/mackuba/rails-timestamp-precision-demo.

It creates a table with 3 sets of 4 timestamp fields each: `datetime`, `timestamp`, and `timestamptz`, with `precision` set to 0, 6, `nil` and not specified. These columns all end up as either `timestamp` or `timestamptz` in the database, since `datetime` maps to one of those depending on the `datetime_type` setting, but where they differ is in the precision flags. I would have expected the 4 precision options to end up with the same result in each of the 3 sets, but they don't:

- the `timestamptz` set of columns have no precision at all, none of the 4 fields
- the `timestamp` and `timestamptz` with no precision set default to a precision `nil`, instead of 6 as `datetime` does

Note, this is all Postgres specific, since in MySQL `datetime` and `timestamp` are two very different column types.

### Detail

I've made a number of fixes in:

- `ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaStatements#type_to_sql`
- `ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnMethods::TableDefinition#new_column_definition`
- `ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaDumper#schema_precision`

which make the precision behave the same across Postgres columns declared as `datetime`, `timestamp` and `timestamptz`, both with `datetime_type` set to `:timestamptz` and with the default setting, in how the `schema.rb` is written and in what actually ends up in the database structure. Plus I've added some tests that I think should cover all of this.

### Additional information

I made all the changes in the `PostgreSQL` subclasses of the components, since these are all Postgres-specific things. Alternatively, some changes could be made in the abstract base class instead, e.g. the first commit could just be replaced with adding `:timestamptz` here to the base `SchemaStatements`:

```rb
elsif [:datetime, :timestamp, :time, :interval].include?(type) && precision ||= native[:precision]
```

since the `:interval`, which I think is also Postgres-specific, is already there. But it felt more appropriate to put it in the Postgres subclasses.

We could also move the `:interval` from there to the Postgres subclass now for consistency.

One thing to note is that this will create diffs in some people's `schema.rb` on update and after regenerating the schema, because of the changed defaults:

- if someone had a `t.timestamptz` field (or a `t.timestamp` field in `datetime_type = :timestamptz` mode), it would have no precision and be printed to `schema.rb` with no precision option, but now it will be dumped as `..., precision: nil`, since the default will now be 6
- in `datetime_type = :timestamptz` mode, if someone had a `t.timestamp` field with `precision: 6`, it will be dumped to `schema.rb` with the option omitted since that's now the default as in `datetime`

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,29 @@
+*   Fix PostgreSQL migrations ignoring `precision` option for `timestamptz` columns.
+
+    Previously, columns with an explicit `timestamptz` type were always created
+    without a precision specifier, regardless if the `precision` option was passed
+    and what it was set to.
+
+    *Kuba Suder*
+
+*   Change default precision for `timestamp` and `timestamptz` columns in
+    PostgreSQL migrations to be 6 like in `datetime` columns.
+
+    `datetime` column type maps to either `timestamp` or `timestamptz` in
+    PostgreSQL databases, depending on the `datetime_type` setting. However,
+    while `datetime` has a default precision of 6 since Rails 7.0, migrations
+    that added columns with type passed as explicit `timestamp` or `timestamptz`
+    used a default precision of `nil` if no `precision` option was added.
+
+    This change makes the `timestamp` and `timestamptz` columns always use a
+    default precision of 6, regardless if the type is specified as `datetime`
+    or an explicit `timestamp(tz)` in the migration. This also affects how
+    columns with these types (if they aren't rendered as `datetime`) are dumped
+    into the `schema.rb`: `precision: nil` is added if the column doesn't have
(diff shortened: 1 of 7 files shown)


deepakmahakale:
Thanks. Right now, this is a major clutter in our project

When we dump and restore, this precision is lost in dev environments, causing an inconsistent DB on other devs' machines

PR #57547, opened 2026-06-02
Test Mime::Type#=== and nil matching

hammadxcm:
`Mime::Type#===` matches when the type is included in a given array (used for `case`/`when` against a list of mime types), and `#=~`/`#match?` guard against a `nil` argument by returning false. Neither the array form of `#===` nor the `nil` guard had a test.

This adds coverage for both. No production code changes — test-only.


fatkodima:
I think maybe open a single PR for missing tests (if they are simple) per Rails component in the future 🤔.

hammadxcm:
> I think maybe open a single PR for missing tests (if they are simple) per Rails component in the future 🤔.

Sure. I am not familiar with the practices we follow. I will create single PRs now.

PR #57551, opened 2026-06-02
Rewrite the Action Cable overview guide

ayushn21:
### Motivation / Background

A restructure and rewrite of the Action Cable Overview guide.

### Detail

Key changes:

* Restructure the flow of the guide to demonstrate end to end usage before drilling down into individual concepts.
* Add an intro section describing WebSockets and Pub/Sub.
* Add details about the architecture of Action Cable (server and pub/sub).
* Mention the recent adapterization of the server.
* Remove section on full-stack examples, instead integrating the info with the narrative flow.
* Describe the adapters in its own section.
* Add details on the `stream_from` and `stream_for` methods.

#### Internal Audit

- as app/channels are not generated in a new Rails app then we need some setup instructions and general introduction/overview at the start of this guide.
- Missing context in the first paragraph about why you might need to integrate websockets with your application (i.e. chat functionality seems like the most likely/common use-case).
- Section 3.2.3 is a repeat of section 3.1.2 - error handling. 
- Likewise 3.2.4 is a repeat of 3.1.3 - callbacks.
- The full-stack examples section - it's a good idea to include something like this, but there is so much code here that might be completely foreign to the reader. Maybe a separate tutorial/blog-style guide around implementing ActionCable would be better for this content, and keep the guide for reference/explanation only.
- Reading further, I can see that there's a link to more examples in section 6.3. I would maybe suggest that we remove the examples here and just move this link to further up the guide for readers to choose a walk through of how to implement the functionality upon following the link.

This was a bit of a funny one to think about with regards to beginners, because I think you wouldn't be trying to implement and understand lots of these concepts if you were a complete beginner - I think it's more intermediate-level functionality, once you have the basics down.

That said, I think the structure of this guide is the biggest thing we can think about where beginners are concerned. So, to imagine a beginner reading maybe the first few paragraphs with some broad concepts and explanations of terminology, and then perhaps having the option to follow a link to step through an implementation guide, that they may not understand all of at first, but can follow instructions. Then the more advanced reader can engage with the following sections, that include higher level concepts.

Maybe mention at the top that "Now that Hotwire is the default, the majority of apps won't need custom channels. And those that do can get the files back via the generator."

New Rails apps use [SolidCabl](https://github.com/rails/solid_cable)e by default instead of the async adapter (as mentioned in the Dependencies section).
I agree with Harriet that the structure of the guide can be improved for beginners. I'm wondering if it would be possible to integrate the fullstack examples in the other sections, instead of doing it separately. (this would be a lot of work though).
Should this guide explain broadcasting with Hotwire?

This guide starts with a lot of new concepts at once without showing any examples. I think it would be nice if we could start introducing concepts on-by-one with working code examples.

For example by first starting with just the ServerSide Components and using the browser console/debugger we can send messages to the server and see what responses we get back. After that we can introduce the client side.

Showing how this works will also help debugging ActionCable issues.

For example, if I start a new Rails app and add a channel:

```
# app/channels/appearance_channel.rb
class AppearanceChannel < ActionCable::Channel::Base
  def subscribed
    logger.debug "Subscribed!"
    logger.debug params.inspect
  end
end
```

I can then talk to the ActionCable server with plain Javascript:

```
websocket = new WebSocket("ws://localhost:3000/cable");
websocket.addEventListener("open", () => {
  console.log("CONNECTED");
  payload = JSON.stringify({
    command: "subscribe",
    identifier: JSON.stringify({
      channel: "AppearanceChannel",
      foo: "bar"
    })
  })
  console.log(payload)
  websocket.send(payload)
});

// Receiving messages:
websocket.addEventListener("message", (e) => {
  console.log(`RECEIVED: ${e.data}`);
});
```

This shows how subscribing works and what messages are received.

We could also show an example that broadcast a message to a chat room:

```
class ChatChannel < ActionCable::Channel::Base
  def subscribed
    room = "chat_#{params[:room]}"
    user = params[:user]
    stream_from room

    # With the default async Cable adapter we
    # can't send this message from the console.
    ActionCable.server.broadcast(
      room,
      { body: "#{user} joined the chat" }
    )
  end
end
```

```
websocket = new WebSocket("ws://localhost:3000/cable");
websocket.addEventListener("open", () => {
  console.log("CONNECTED");
  payload = JSON.stringify({
    command: "subscribe",
    identifier: JSON.stringify({
      channel: "ChatChannel",
      room: "general",
      user: "alice"
    })
  })
  console.log(payload)
  websocket.send(payload)
});

// Receiving messages:
websocket.addEventListener("message", (e) => {
  console.log(`RECEIVED: ${e.data}`);
});
```

Using the browser console could also be used for examples like authorization.

Server-Side Components

I think it makes more sense to start with the Channels instead of the Connections. Channels can work without any custom Connections.
So this section could start with examples without authorization.
Authorization could be added later.

It also is a bit confusing to me how ApplicationCable::Connection gets configured. It turns out, this is hardcoded. If you add `ApplicationCable::Connection` specifically it gets used for the connection:
https://github.com/rails/rails/blob/89a0c5d5f6c903810dd945ee803e563f689ac23a/actioncable/lib/action_cable/engine.rb#L59-L60

Client Server Interactions
- This section could probably be merged in the "Server-side Components" section and "Client-side Components" section.
- The Streams and Broadcast section are related. Broadcasts are also mentionedin the Stream section. Maybe it's better to combine them to a "Streams and Broadcasts" section?

Fullstack examples
I think the fullstack examples can be removed. If we show all functionality in the other sections they don't add much.

Configuration
- Should we maybe explain the adapters earlier on in a "Getting started"/"Solid Cable setup" section? Guides like, the ActiveJob do this as well. Also, if a reader uses the async adapter things work a bit differently, for example broadcasts from the console won't work. Maybe we should also show the examples with Solid Cable configured?
- The Rails configurations should probably be moved to the "Configuring" guide, as that is where we document most of the configurations.

Running Standalone Cable Servers
The Notes section can probably be removed. We already mention how to access the session cookies in another section.

Dependencies
This section doesn't add much. We also don't mention depedencies for other components.

Deployment
This section could probably be integrated in the "Running Standalone Cable Servers". Maybe that should be renamed to "Running Cable Servers in production"?

### Checklist


ayushn21:
Some additional notes with some internal audit responses:

* Some parts are repeated in the `Server Components` and `JavaScript Client Library` sections. I think this is ok as I would like each section to be as standalone as possible. Given the full-stack nature of Action Cable, it's often a catch-22 situation when trying to explain both ends of a feature. I would rather repeat the information from the two different points of view.

> Maybe mention at the top that "Now that Hotwire is the default, the majority of apps won't need custom channels. And those that do can get the files back via the generator."

I'd rather not make assumptions about what majority of apps may or may not need. I would rather the guide focus on what Action Cable does rather than what a user _might_ do. Armed with the information of the guide I would trust the user to make their own judgement on what they need.

> Should this guide explain broadcasting with Hotwire?

Nope, we've got the greenlight for a separate Hotwire guide where we can cover that.

> I can then talk to the ActionCable server with plain Javascript:

I've stuck to using the Action Cable JS library. Since it's an Action Cable guide, I want to keep the focus tight. My hope is that the concepts are explained well enough that if a user would like to go deeper and experiment directly with the underlying WebSocket API, they should have the knowledge to do so after reading the guide.

> I think it makes more sense to start with the Channels instead of the Connections. Channels can work without any custom Connections.

The connection is still the entry point so I've started there. I hope my narrative flow makes it the logical choice.

> The Rails configurations should probably be moved to the "Configuring" guide, as that is where we document most of the configurations.

The configuration options are fairly key to running Action Cable so I've decided to keep them. I feel it goes well with the flow.

> Dependencies. This section doesn't add much. We also don't mention depedencies for other components.

This is true, but Action Cable's server is unique in its scope. I've removed this section but kept the mention of the dependencies. I think it's good for the reader to understand the underlying architecture at a basic conceptual level.

PR #57555, opened 2026-06-03
Test Type::Integer#cast of a Symbol returns nil

hammadxcm:
`ActiveModel::Type::Integer`'s documentation uses a `Symbol` as the example of a value that casts to `nil` (because `Symbol` does not define `#to_i`):

```ruby
person.age = :not_an_integer
person.age # => nil
```

The closest existing test cast a bare `Object`, so the documented `Symbol` example was never exercised. This covers it. No production code changes — test-only.


fatkodima:
> The closest existing test cast a bare `Object`

Thanks, but I think that test case (even from its description) covers this too.

Issue #57552, opened 2026-06-02
Action Text: concurrent saves of a `has_rich_text` attribute raise `ActiveRecord::RecordNotUnique`


danielmorrison:
Based on real exceptions noticed in a medium-traffic production app.

## Steps to reproduce

When two requests save the same record's rich text at nearly the same moment (a double-tapped button, a retried request, two open editor tabs), neither has an `action_text_rich_texts` row loaded, so each builds a brand-new one in memory
and both `INSERT` it. The loser violates the `index_action_text_rich_texts_uniqueness` unique index on `(record_type, record_id, name)` and raises `ActiveRecord::RecordNotUnique`.

The script below is the deterministic, single-process equivalent of that race (self-contained, based on the official `active_storage` bug report template):

```ruby
# Run with:  ruby action_text_concurrency.rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  # gem "rails", github: "rails/rails", branch: "main"
  gem "sqlite3"
end

require "active_record/railtie"
require "active_storage/engine"
require "action_text/engine"
require "minitest/autorun"

ENV["DATABASE_URL"] = "sqlite3::memory:"

class TestApp < Rails::Application
  config.load_defaults Rails::VERSION::STRING.to_f
  config.root = __dir__
  config.hosts << "example.org"
  config.eager_load = false
  config.session_store :cookie_store, key: "cookie_store_key"
  config.secret_key_base = "secret_key_base"
  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger
  config.active_storage.service = :local
  config.active_storage.service_configurations = {
    local: { root: Dir.tmpdir, service: "Disk" }
  }
  config.active_job.queue_adapter = :inline
end
Rails.application.initialize!

require ActiveStorage::Engine.root.join("db/migrate/20170806125915_create_active_storage_tables.rb").to_s
require ActionText::Engine.root.join("db/migrate/20180528164100_create_action_text_tables.rb").to_s

ActiveRecord::Schema.define do
  CreateActiveStorageTables.new.change
  CreateActionTextTables.new.change
  create_table :messages, force: true
end

class Message < ActiveRecord::Base
  has_rich_text :content
end

class ConcurrentRichTextSaveTest < ActiveSupport::TestCase
  def test_second_concurrent_save_should_not_raise
    message = Message.create! # no content yet -> no rich-text row exists

    request_a = Message.find(message.id)
[...]

PR #57564, opened 2026-06-03
[#57552] Fix concurrent saves of `has_rich_text` raising `ActiveRecord::RecordNotUnique`.

ruyrocha:
### Motivation / Background

This Pull Request has been created because #57552

### Detail

When two requests simultaneously save rich text content for the same record and attribute, both build a new `ActionText::RichText` in memory and attempt to `INSERT` it on the parent save. The database unique index on `(record_type, record_id, name)` guarantees that only one succeeds; the loser raises `ActiveRecord::RecordNotUnique`. 

### Checklist

### DIFF
--- a/actiontext/CHANGELOG.md
+++ b/actiontext/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Fix concurrent saves of `has_rich_text` raising `ActiveRecord::RecordNotUnique`.
+
+    *Ruy Rocha*
+
 *   Dispatch all Active Storage `direct-upload:`-prefixed events
 
     Add support for `direct-upload:before-blob-request` and `direct-upload:before-storage-request`.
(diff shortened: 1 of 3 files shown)


fabriazza:
Thanks for working on this. I think this needs a different recovery boundary before it can safely handle the race.

`RichText#save` is called inside the parent record’s autosave transaction. On PostgreSQL, once `super` raises `RecordNotUnique`, that transaction is aborted, so the subsequent `find_by`/`update!` can raise `PG::InFailedSqlTransaction`. `create_or_find_by!` avoids this by attempting its insert inside `transaction(requires_new: true)`.

There are also risks in rescuing every `RecordNotUnique` from `RichText#save`, since the violation could originate from another unique constraint or callback, and directly setting `@new_record`, `@previously_new_record`, and `@transaction_state` bypasses Active Record persistence/dirty-tracking state. Updating only `body` may also omit custom RichText subclass attributes or attachment state.

Could we first add PostgreSQL coverage and tests for the returned association’s persisted/dirty state? Maintainer guidance may be needed on whether this should use a targeted savepoint/retry or an atomic upsert.

PR #57565, opened 2026-06-03
Preserve index attributes when renaming an index without native support

55728:
## Summary

The fallback `rename_index` silently drops a partial index's `WHERE` clause, column sort order, and every other index attribute except `columns`/`unique`:

```ruby
add_index :posts, :views, where: "title IS NOT NULL", name: "idx"
rename_index :posts, "idx", "idx2"
# idx2 is now a plain index — WHERE is gone
```

The most common trigger is `rename_column` on SQLite (the default dev/test DB), which reindexes default-named indexes through `rename_index`. For a unique partial index this also changes which rows the uniqueness constraint covers.

## Which adapters are affected

| Adapter | `rename_index` path | Affected? |
|---|---|---|
| PostgreSQL | `ALTER INDEX … RENAME` (native) | No |
| MySQL ≥ 5.7.6 / MariaDB ≥ 10.5.2 | `ALTER TABLE … RENAME INDEX` (native) | No |
| SQLite | fallback (`super`) | **Yes** |
| MySQL < 5.7.6 / MariaDB < 10.5.2 | fallback (`super`) | **Yes** |

Notably, SQLite's own `copy_table_indexes` (used during table rebuild) already preserves `where` and `order` — `rename_index` was the outlier.

## Fix

Carry the old `IndexDefinition`'s `where`, `orders`, `lengths`, `opclasses`, `using`, `type`, `include`, `nulls_not_distinct`, and `comment` into the `add_index` call, each guarded by presence.

(`comment` only matters on the MySQL/MariaDB fallback path — SQLite has no index comments — but it's carried for completeness so the recreated index mirrors every introspected attribute.)

## Tests

- `test_rename_index_preserves_where_clause` (guarded by `supports_partial_index?`)
- `test_rename_index_preserves_order` (guarded by `supports_index_sort_order?`)

Both snapshot the attribute before rename and assert it survives (robust to per-adapter introspection differences).

Verified red→green on sqlite3 (`where` → `nil`, `orders` → `{}`  on `main`; both preserved with the fix). Green on postgresql (native rename, unaffected). Full `index_test.rb` (33), `columns_test.rb` (36), `sqlite3_adapter_test.rb` (97) stay green.

## Note

This was reported in #16619 (2014) and closed with "the abstract implementation is for adapters that don't have `where`." That rationale no longer holds — SQLite supports partial indexes (`supports_partial_index?`) yet uses this abstract fallback because it has no native `rename_index`, so an adapter that *does* support `where` now silently loses it.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Fix `rename_index` to preserve a partial index's `WHERE` for SQLite and older MySQL/MariaDB versions.
+
+    *Kenta Ishizaki*
+
 *   Fix PostgreSQL `foreign_keys` returning a corrupted `to_table` for a foreign key
     that references a table in a quoted schema.
 
--- a/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb
@@ -1084,7 +1084,7 @@ module ActiveRecord
         # this is a naive implementation; some DBs may support this more efficiently (PostgreSQL, for instance)
         old_index_def = indexes(table_name).detect { |i| i.name == old_name }
         return unless old_index_def
-        add_index(table_name, old_index_def.columns, name: new_name, unique: old_index_def.unique)
+        add_index(table_name, old_index_def.columns, name: new_name, unique: old_index_def.unique, where: old_index_def.where)
         remove_index(table_name, name: old_name)
       end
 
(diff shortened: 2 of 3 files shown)


fatkodima:
I think worth fixing only for the missing `where` attribute. Others are adapter specific (like postgres, which already supports renaming an index) or not supported by very old versions (mysql, mariadb), which will be definitely dropped some time soon.

55728:
Thanks for the trim and merge! Noted — will keep the scope minimal next time 👀✨

PR #57568, opened 2026-06-03
Classify MySQL check-read errors as serialization failures

matthewd:
MariaDB reports 1020 by default under repeatable-read in recent versions; this doesn't _prevent_ that, but means we at least correctly understand the full transaction has been invalidated and avoid further failure on a bad savepoint rollback.

Related #53727, #57535

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb
@@ -916,6 +916,7 @@ module ActiveRecord
         ER_NO_REFERENCED_ROW_2  = 1452
         ER_DATA_TOO_LONG        = 1406
         ER_OUT_OF_RANGE         = 1264
+        ER_CHECKREAD            = 1020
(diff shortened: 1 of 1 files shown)


rafaelfranca (approved):

matthewd:
My only question here is 1) does MySQL agree `1020` is a transaction-rollback error, 2) if not, is it reasonable/safe for us to "mistakenly" treat it as such?

My intuition is that it would probably be safe, and the worst case is that we'd believe we're in a less recoverable state than we actually are... which seems fine. But I haven't traced to verify.

matthewd:
> My intuition is that it would probably be safe, and the worst case is that we'd believe we're in a less recoverable state than we actually are... which seems fine. But I haven't traced to verify.

No: if we believe the transaction has been `invalidated?`, we won't rollback, so getting it wrong is Bad. My understanding is that MySQL will only emit 1020 in rare circumstances, but it still doesn't seem good to risk getting desynced; this might need a bit more thought.

PR #57569, opened 2026-06-03
Handle invalid filenames when inferring send file content type

afurm:
Summary:
- Avoid raising when send_data infers a content type from a filename that File.extname rejects.
- Use the same fallback for send_stream while keeping Content-Disposition filename escaping unchanged.
- Add coverage for null-byte filenames in send_data and send_stream.

Reproduction:
- The new send_data regression failed before the fix with ArgumentError: path name contains null byte at actionpack/lib/action_controller/metal/data_streaming.rb:141.

Tests:
- cd actionpack && bin/test test/controller/send_file_test.rb
- cd actionpack && bin/test test/controller/live_stream_test.rb
- cd actionpack && bin/test test/dispatch/content_disposition_test.rb
- RUBOCOP_CACHE_ROOT=/private/tmp/rubocop_cache bundle exec rubocop actionpack/lib/action_dispatch/http/content_disposition.rb actionpack/lib/action_controller/metal/data_streaming.rb actionpack/lib/action_controller/metal/live.rb actionpack/test/controller/send_file_test.rb actionpack/test/controller/live_stream_test.rb

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Fall back to the default send file content type when MIME inference cannot
+    read the provided filename extension.
+
+    *Andrii Furmanets*
+
 *   Rate limiting calls `cache_key` on `by:` if the object responds to it.
 
     ```ruby
--- a/actionpack/lib/action_controller/metal/data_streaming.rb
+++ b/actionpack/lib/action_controller/metal/data_streaming.rb
@@ -138,7 +138,9 @@ module ActionController # :nodoc:
         else
           if !type_provided && options[:filename]
             # If type wasn't provided, try guessing from file extension.
-            content_type = Mime::Type.lookup_by_extension(File.extname(options[:filename]).downcase.delete(".")) || content_type
+            content_type = Mime::Type.lookup_by_extension(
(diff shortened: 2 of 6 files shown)


p8:
What is the use-case for this?
And why should "sample\0.csv" result in "sample"?

afurm:
The use case is `send_data` / `send_stream` with a user-provided suggested filename and no explicit `type`. The filename is only used for headers, but MIME inference currently calls `File.extname`, so a malformed filename can raise before Rails falls back to the default content type.

I updated the stream test so the body is unrelated to the filename; `sample\0.csv` is only the filename under test.

p8:
But what is your use-case for wanting to fix this?
Are you allowing params in a controller to define the filename used in send_data (not recommended)?
Are you using the unsanitized filename of an ActiveStorage file (also not recommended)?

Raising the ArgumentError is a lot more helpful in finding an issue than having to look into the Rails source code why some filenames don't have extensions.

afurm:
The change mainly helps apps that pass a malformed suggested filename while relying on implicit type inference. That is a weak use case, and Rails raising here is more useful because it points back to the invalid filename instead of silently falling back to `application/octet-stream`.

I’m fine closing this.

PR #57570, opened 2026-06-03
Enable per-pool query log tags formatter

hmcguire-shopify:
A `query_log_tags_format` key in a `database.yml` entry overrides the global `config.active_record.query_log_tags_format` for connections in that pool, so different databases can emit `:legacy` or `:sqlcommenter` formatted comments.

```yaml
production:
  primary:
    database: primary
  analytics:
    database: analytics
    query_log_tags_format: sqlcommenter
```

### Checklist

### DIFF
--- a/activerecord/lib/active_record/database_configurations/hash_config.rb
+++ b/activerecord/lib/active_record/database_configurations/hash_config.rb
@@ -105,6 +105,12 @@ module ActiveRecord
         configuration_hash[:query_cache]
       end
 
+      def query_log_tags_format # :nodoc:
+        return @query_log_tags_format if defined? @query_log_tags_format
+
+        @query_log_tags_format = configuration_hash[:query_log_tags_format]&.to_sym
+      end
+
       def max_queue
         max_threads * 4
       end
--- a/activerecord/lib/active_record/query_logs.rb
+++ b/activerecord/lib/active_record/query_logs.rb
@@ -85,6 +85,17 @@ module ActiveRecord
   # using the {SQLCommenter}[https://open-telemetry.github.io/opentelemetry-sqlcommenter/] format. This can be changed
   # via {config.active_record.query_log_tags_format}[https://guides.rubyonrails.org/configuring.html#config-active-record-query-log-tags-format]
   #
+  # The format can also be overridden per connection pool by setting +query_log_tags_format+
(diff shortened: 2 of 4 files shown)


activerecord/lib/active_record/query_logs.rb
```diff
@@ -181,6 +185,25 @@ def query_source_location # :nodoc:
...
+            LegacyFormatter
+          when :sqlcommenter
+            SQLCommenter
+          else
+            raise ArgumentError, "Formatter is unsupported: #{format}"
+          end
+        end
+
+        def resolve_formatter(connection)
+          if (format = connection.pool.db_config.query_log_tags_format)
```
nvasilevski, line 200: As with other per-adapter stuff we learned that reaching out to the pool on a hot path may be noticeable so I wonder if at least we could avoid "punishing" applications that don't use multiple databases or if none of the databases have explicit `query_log_tags_format` configured.

Would this be an overkill to do a check like:
```ruby
if explicit_query_log_tags_format_configured?
  def resolve_formatter(connection)
      format = connection.pool.db_config.query_log_tags_format
      format ? formatter_for(format) : @formatter
  end
else
  def def resolve_formatter(_connection)
      @formatter
  end
```

Not a strong point, perhaps best to wait for a regression report before optimizing prematurely

hmcguire-shopify: For the other stuff, the hot path was Model -> Pool because the "current pool" has to be looked up from scratch (whats the current role, whats the current shard, etc.). Going from a known connection to its pool is still an instance variable lookup in all cases.

nvasilevski: Right, thanks for clarifying!

nvasilevski (approved):

activerecord/lib/active_record/database_configurations/hash_config.rb
```diff
@@ -105,6 +105,10 @@ def query_cache
         configuration_hash[:query_cache]
       end
 
+      def query_log_tags_format # :nodoc:
+        configuration_hash[:query_log_tags_format]&.to_sym
```
byroot: Not the end of the world, but can we avoid calling `to_sym` on every access? Maybe just a memoize pattern?

rafaelfranca (approved):

PR #57573, opened 2026-06-04
[RF-DOCS] Update the guide for using Rails as an API

ayushn21:
### Motivation / Background

Update and simplify the guide for using Rails as an API.

### Detail

- Changed the structure of the guide to be more general and less like a step-by-step tutorial. This is in keeping with the rest of the guides as they may not always be read or followed all the way from beginning to end.
- Simplified the section on why we would use Rails as an API. It read like a feature-list for Rails. I've removed all that detail and replaced it with a simple pitch stating the majority of Rails still applies in API applications. I've also removed language which makes assumptions about the reader's opinions.
- Added sections on JBuilder and Builder, as well as XML responses which, while not as popular as JSON, is still a valid API data-interchange format.
- Updated the sections on Rack middleware and Action Controller modules. Some of the information was out of date — for example, the `ActionController::Caching` module is now included in `ActionController::API`.

I considered adding a new section on authenticating HTTP requests for an API, but it's not suitable for a general Rails guide. It would make too many assumption about the internal authentication model of the application. Instead, I've linked the modules for HTTP authentication included in Rails which the developer can use to build their authentication system.

### Internal Audit

This guide sometimes uses lowercase "api". I think it should always be "API".
This guide is about a JSON API, where we don't mention GraphQL which is another popular API implementation.

The Basic Configuration
- Should this section be renamed to something like "Setting Up an Application as an  API" or "Using Rails as an API".

Choosing Middleware
- Rack::Cache is described but it's not in the intro list of middlewares.

### DIFF
--- a/guides/source/api_app.md
+++ b/guides/source/api_app.md
@@ -5,174 +5,104 @@ Using Rails for API-only Applications
 
 In this guide you will learn:
 
-* What Rails provides for API-only applications
-* How to configure Rails to start without any browser features
-* How to decide which middleware you will want to include
-* How to decide which modules to use in your controller
+- Why you should use Rails to build an API.
+- How to configure Rails without any browser features.
+- Rendering JSON and XML responses.
+- The Rack middleware stack in API applications.
+- The Action Controller modules excluded from API applications.
 
 --------------------------------------------------------------------------------
 
 What is an API Application?
 ---------------------------
 
-Traditionally, when people said that they used Rails as an "API", they meant
-providing a programmatically accessible API alongside their web application.
-For example, GitHub provides [an API](https://developer.github.com) that you
-can use from your own custom clients.
+An API application speaks a data-interchange format such as JSON or XML, rather
+than HTML which is used to paint a user interface. Its purpose is to integrate
+with another piece of software, as opposed to being used by a human.
 
-With the advent of client-side frameworks, more developers are using Rails to
-build a back-end that is shared between their web application and other native
-applications.
+While Rails
+[provides a mechanism to build an API](action_view_overview.html#jbuilder)
+alongside your full-stack application, a Rails application that only speaks JSON
+(or another similar format) can be a useful configuration. The same Rails API
+could service the JavaScript frontend in the browser, as well as the native
+mobile apps and other integrations.
 
-For example, X uses its [public API](https://developer.x.com/) in its web
-application, which is built as a static site that consumes JSON resources.
+[Basecamp](https://github.com/basecamp/bc3-api) is a good example of the former
+approach where HTML is used for web browsers, but a JSON API is also provided
+for external software integrations. [X](https://x.com) uses the latter approach
+in its web application which is built using a frontend that consumes JSON
+resources from an internal API.
 
-Instead of using Rails to generate HTML that communicates with the server
-through forms and links, many developers are treating their web application as
-just an API client delivered as HTML with JavaScript that consumes a JSON API.
+This guide covers Rails' support for building an API application without any
+HTML rendering or browser-related features.
 
-This guide covers building a Rails application that serves JSON resources to an
-API client, including client-side frameworks.
-
-Why Use Rails for JSON APIs?
-----------------------------
-
-The first question a lot of people have when thinking about building a JSON API
-using Rails is: "isn't using Rails to spit out some JSON overkill? Shouldn't I
-just use something like Sinatra?".
-
-For very simple APIs, this may be true. However, even in very HTML-heavy
-applications, most of an application's logic lives outside of the view
-layer.
-
-The reason most people use Rails is that it provides a set of defaults that
-allows developers to get up and running quickly, without having to make a lot of trivial
-decisions.
-
-Let's take a look at some of the things that Rails provides out of the box that are
-still applicable to API applications.
-
-Handled at the middleware layer:
-
-- Reloading: Rails applications support transparent reloading. This works even if
-  your application gets big and restarting the server for every request becomes
-  non-viable.
-- Development Mode: Rails applications come with smart defaults for development,
-  making development pleasant without compromising production-time performance.
-- Test Mode: Ditto development mode.
-- Logging: Rails applications log every request, with a level of verbosity
-  appropriate for the current mode. Rails logs in development include information
-  about the request environment, database queries, and basic performance
-  information.
-- Security: Rails detects and thwarts [IP spoofing
-  attacks](https://en.wikipedia.org/wiki/IP_address_spoofing) and handles
-  cryptographic signatures in a [timing
-  attack](https://en.wikipedia.org/wiki/Timing_attack) aware way. Don't know what
-  an IP spoofing attack or a timing attack is? Exactly.
-- Parameter Parsing: Want to specify your parameters as JSON instead of as a
-  URL-encoded String? No problem. Rails will decode the JSON for you and make
-  it available in `params`. Want to use nested URL-encoded parameters? That
-  works too.
-- Conditional GETs: Rails handles conditional `GET` (`ETag` and `Last-Modified`)
-  processing request headers and returning the correct response headers and status
-  code. All you need to do is use the
-  [`stale?`](https://api.rubyonrails.org/classes/ActionController/ConditionalGet.html#method-i-stale-3F)
-  check in your controller, and Rails will handle all of the HTTP details for you.
-- HEAD requests: Rails will transparently convert `HEAD` requests into `GET` ones,
-  and return just the headers on the way out. This makes `HEAD` work reliably in
-  all Rails APIs.
-
-While you could obviously build these up in terms of existing Rack middleware,
-this list demonstrates that the default Rails middleware stack provides a lot
-of value, even if you're "just generating JSON".
-
-Handled at the Action Pack layer:
-
-- Resourceful Routing: If you're building a RESTful JSON API, you want to be
-  using the Rails router. Clean and conventional mapping from HTTP to controllers
-  means not having to spend time thinking about how to model your API in terms
-  of HTTP.
-- URL Generation: The flip side of routing is URL generation. A good API based
-  on HTTP includes URLs (see [the GitHub Gist API](https://docs.github.com/en/rest/reference/gists)
-  for an example).
-- Header and Redirection Responses: `head :no_content` and
-  `redirect_to user_url(current_user)` come in handy. Sure, you could manually
-  add the response headers, but why?
-- Caching: Rails provides page, action, and fragment caching. Fragment caching
-  is especially helpful when building up a nested JSON object.
-- Basic, Digest, and Token Authentication: Rails comes with out-of-the-box support
-  for three kinds of HTTP authentication.
-- Instrumentation: Rails has an instrumentation API that triggers registered
-  handlers for a variety of events, such as action processing, sending a file or
-  data, redirection, and database queries. The payload of each event comes with
-  relevant information (for the action processing event, the payload includes
-  the controller, action, parameters, request format, request method, and the
-  request's full path).
-- Generators: It is often handy to generate a resource and get your model,
-  controller, test stubs, and routes created for you in a single command for
-  further tweaking. Same for migrations and others.
-- Plugins: Many third-party libraries come with support for Rails that reduce
-  or eliminate the cost of setting up and gluing together the library and the
-  web framework. This includes things like overriding default generators, adding
-  Rake tasks, and honoring Rails choices (like the logger and cache back-end).
-
-Of course, the Rails boot process also glues together all registered components.
-For example, the Rails boot process is what uses your `config/database.yml` file
-when configuring Active Record.
-
-**The short version is**: you may not have thought about which parts of Rails
-are still applicable even if you remove the view layer, but the answer turns out
-to be most of it.
-
-The Basic Configuration
+Why Use Rails for APIs?
 -----------------------
 
-If you're building a Rails application that will be an API server first and
-foremost, you can start with a more limited subset of Rails and add in features
-as needed.
+When using Rails as an API, we replace a view layer that generates HTML with one
+that generates JSON, XML, or a similar data-interchage format. Nothing else
+changes.
 
-### Creating a New Application
+Rails still provides a feature-rich environment to develop server applications
+with its security features, MVC architecture, and resourceful routing. Tools
+such as [Active Record][], [Active Job][], [Active Storage][], [Action
+Mailer][], and [native caching][] are all still available in API applications.
 
-You can generate a new api Rails app:
+Most of Rails' features are still relevant even when building an API-only
+application.
+
+[Active Record]: active_record_basics.html
+[Active Job]: active_job_basics.html
+[Active Storage]: active_storage_overview.html
+[Action Mailer]: action_mailer_basics.html
+[native caching]: caching_with_rails.html
+
+Using Rails for API-only Applications
+-------------------------------------
+
+Rails can be configured to be used only as an API, which will exclude the HTML
+rendering and browser-related features. Some features can be added back in as
+required.
+
+### Creating a New API Application
+
+You can generate a new API-only Rails app using:
 
 ```bash
-$ rails new my_api --api
+$ rails new my_api_app --api
 ```
 
-This will do three main things for you:
+This will:
 
-- Configure your application to start with a more limited set of middleware
-  than normal. Specifically, it will not include any middleware primarily useful
-  for browser applications (like cookies support) by default.
-- Make `ApplicationController` inherit from `ActionController::API` instead of
-  `ActionController::Base`. As with middleware, this will leave out any Action
-  Controller modules that provide functionalities primarily used by browser
-  applications.
+- Set `ApplicationController` to inherit from [`ActionController::API`][]
+  instead of [`ActionController::Base`][], which
+  [excludes the modules](#api-controller-modules) for browser-related
+  functionality.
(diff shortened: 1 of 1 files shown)


andynu:
Great guide. I have one suggestion. The guide correctly notes that the ActionController::API only provides a non-browser subset of behavior. There is one big one that I wonder how many people will hit and be confused by: What happens if you try to call out to a template?

The doc currently has this line:
> Make ApplicationController inherit from ActionController::API instead of ActionController::Base. As with middleware, this will leave out any Action Controller modules that provide functionalities primarily used by browser applications.

But I wonder if a specific gotcha callout, or note about template rendering might be worthwhile.

If a controller in an API-only application calls `render :some_template` (or relies on implicit rendering), the response will not raise — it will return 204 No Content for the implicit case, or 200 OK with an empty text/plain body for the explicit case. There is no exception or log line indicating the template was ignored. If you need template rendering in a specific controller, include ActionView::Rendering (and ActionView::Layouts if you also need layouts) to restore it.

ayushn21:
Thanks for that @andynu! Super helpful, because I agreed with your point and went looking for a good place to mention it in the guide with some background detail. In that process I found that the `builder` gem requires `ActionView::Rendering` to be manually included, whereas `jbuilder` includes it automatically.

In my test app, I always had `jbuilder` in my bundle meaning I didn't stumble upon this gotcha. I've now added step in the `builder` section to add the `ActionView::Rendering` module, and added a note mentioning the `204` response when it's missing.

https://github.com/rails/rails/pull/57573/commits/e7de3b8eb5db31c27c94d85c14132959f18cafba

✌️

guides/source/api_app.md
```diff
@@ -5,174 +5,89 @@ Using Rails for API-only Applications
...
 
 ```bash
-$ bin/rails g scaffold Group name:string
+$ rails new my_api_app --api
 ```
 
-Before we can use our scaffolded code, we need to update our database scheme.
+This will:
+
+- Set `ApplicationController` to inherit from [`ActionController::API`][] instead of [`ActionController::Base`][], which [excludes the modules](#action-controller-modules) for browser-related functionality.
```
alexkuebo: Am I wrong or is there no anchor on this page for `#action-controller-modules`? I only see `api-controller-modules`.

ayushn21: Nice catch, I renamed the heading but forgot to rename the anchor. Fixed now :)

deepan-g2 (commented):
A few minor consistency notes on the code examples:

guides/source/api_app.md
```diff
@@ -204,350 +119,328 @@ class GroupsController < ApplicationController
...
     # Use callbacks to share common setup or constraints between actions.
     def set_group
-      @group = Group.find(params[:id])
+      @group = Group.find(params.expect(:id))
     end
 
     # Only allow a list of trusted parameters through.
     def group_params
-      params.expect(group: [:name])
+      params.expect(group: [ :name ])
```
deepan-g2: The spacing in `[ :name ]` doesn't match Ruby style guide conventions for compact arrays. Rails codebase typically uses `[:name]` without interior spaces. Not a blocker, just noting for consistency.

guides/source/api_app.md
```diff
@@ -204,350 +119,328 @@ class GroupsController < ApplicationController
...
-```irb
-irb> Group.create(name: "Rails Founders")
-irb> Group.create(name: "Rails Contributors")
-```
-
-With some data in the app, we can boot up the server and visit <http://localhost:3000/groups.json> to see our JSON data.
+Assuming we have two `Group`s in our database named _Rails Founders_ and _Rails Contributors_, the `/groups` endpoint will generate the following JSON response.
 
-```json
+```bash
```
deepan-g2, line 165: The curl output below shows all attributes (id, created_at, updated_at), but the JBuilder example only returns the name. This might confuse readers about what the actual output is. Could either match both examples, or add a note explaining that JBuilder lets you selectively control which fields to include.

guides/source/api_app.md
```diff
@@ -204,350 +119,328 @@ class GroupsController < ApplicationController
...
+$ curl -s localhost:3000/groups | jq
```
deepan-g2, line 166: The jq command isn't included by default on most systems. Readers without it installed will see unparsed JSON and might think something went wrong. Maybe show the raw output too, or note that jq requires separate installation (brew install jq)?

guides/source/api_app.md
ayushn21: Fair point, in this case on a second look, I actually think removing spaces looks better so will revert to how it was.

guides/source/api_app.md
ayushn21: I'd hope the snippet demonstrating the JBuilder code should explain why only the name is output in that example. 

```ruby
json.array! @groups do |group|
  json.name(group.name)
end
```

I'm reluctant to spell every bit out for the reader, and I feel the code example should be sufficient in this case to explain why it's just the name and not all the details.

guides/source/api_app.md
ayushn21: Excellent point, let me have a think and will maybe add a note to that effect. I don't want to remove it as the output is really hard to read without the `jq` utility, but it's not great to drop it in there without additional explanation since it's not default. Will improve it.

p8:
@ayushn21  It seems the docs-preview link to an incorrect version: https://buildkite.com/rails/docs-preview/builds/21647/annotations
Could you rebase this to main so a new preview gets generated?

ayushn21:
@p8 done!

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
-Rails ships with a number of other middleware that you might want to use in an
-API application, especially if one of your API clients is the browser:
+API Controller Modules
+----------------------
 
-- `Rack::MethodOverride`
-- `ActionDispatch::Cookies`
-- `ActionDispatch::Flash`
+The base class for API-only controllers, [`ActionController::API`][], includes a
+subset of the modules included [`ActionController::Base`][] which is used in
```
p8, line 420:
```suggestion
subset of the modules included by [`ActionController::Base`][] which is used in
```

guides/source/api_app.md
```diff
@@ -5,174 +5,104 @@ Using Rails for API-only Applications
...
+While Rails
+[provides a mechanism to build an API](action_view_overview.html#jbuilder)
+alongside your full-stack application, a Rails application that only speaks JSON
+(or another similar format) can be a useful configuration. The same Rails API
+could service the JavaScript frontend in the browser, as well as the native
+mobile apps and other integrations.
+
+[Basecamp](https://github.com/basecamp/bc3-api) is a good example of the former
+approach where HTML is used for web browsers, but a JSON API is also provided
+for external software integrations. [X](https://x.com) uses the latter approach
```
p8, line 32: Maybe use an example here that is still a Rails application?
```suggestion
for external software integrations. [Shopify](https://shopify.com) uses the latter approach
```

guides/source/api_app.md
```diff
@@ -5,174 +5,104 @@ Using Rails for API-only Applications
...
+
+Why Use Rails for APIs?
 -----------------------
 
-If you're building a Rails application that will be an API server first and
-foremost, you can start with a more limited subset of Rails and add in features
-as needed.
+When using Rails as an API, we replace a view layer that generates HTML with one
+that generates JSON, XML, or a similar data-interchage format. Nothing else
+changes.
```
p8, line 44: `replace` isn't correct here as a regular Rails app can still render JSON.
And an API only app also removes other browser-related features:
```suggestion
When using Rails as an API, we remove a view layer that generates HTML and other browser-related
features. Nothing else changes.
```

guides/source/api_app.md
```diff
@@ -5,174 +5,104 @@ Using Rails for API-only Applications
...
+Rails can be configured to be used only as an API, which will exclude the HTML
+rendering and browser-related features. Some features can be added back in as
+required.
 
-### Generating a New Resource
+### Creating a New API Application
 
-To see how our newly created API handles generating a new resource, let's create
-a new Group resource. Each group will have a name.
+You can generate a new API-only Rails app using:
```
p8, line 69: I think we can be more explicit here about the `--api` option:
```suggestion
You can generate a new API-only Rails app using the `--api` option:
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
 ]
 ```
 
-### Changing an Existing Application
-
-If you want to take an existing application and make it an API one, read the
-following steps.
+NOTE: The above example, and more examples further down use the `jq` command
+line utility to format JSON output. If it isn't installed on your system,
+consult the [download page](https://jqlang.org/download/) to for instructions.
```
p8, line 185:
```suggestion
NOTE: The above example, and more examples further down, use the `jq` command
line utility to format JSON output. If it isn't installed on your system,
consult the [download page](https://jqlang.org/download/) for instructions.
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
-config.debug_exception_response_format = :default
-```
+# app/views/groups/index.json.jbuilder
 
-To render debugging information preserving the response format, use the value `:api`.
-
-```ruby
-config.debug_exception_response_format = :api
+json.array! @groups do |group|
+  json.name(group.name)
```
p8, line 201: Maybe we can show how we can override attributes here as well?
This would also clarify how the `json#title` method works.
(This also change the output)
```suggestion
  json.title(group.name)
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
-$ bin/rails middleware
+$ curl -s localhost:3000/groups | jq
+[
+  {
+    "name": "Rails Founders"
+  },
+  {
+    "name": "Rails Contributors"
+  }
+]
```
p8, line 232: If we change the attribute to title:
```suggestion
[
  {
    "title": "Rails Founders"
  },
  {
    "title": "Rails Contributors"
  }
]
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
-def show
-  @post = Post.find(params[:id])
+# app/views/groups/index.xml.builder
 
-  if stale?(last_modified: @post.updated_at)
-    render json: @post
+xml.groups do
+  @groups.each do |group|
+    xml.group do
+      xml.name group.name
```
p8, line 250: Similar to the comment about JSON:
```suggestion
      xml.title group.name
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+```bash
+$ curl -s localhost:3000/groups
+<groups>
+  <group>
+    <name>Rails Founders</name>
+  </group>
+  <group>
+    <name>Rails Contributors</name>
+  </group>
+</groups>
```
p8, line 294: If we change the attribute in the builder to title:
```suggestion
<groups>
  <group>
    <title>Rails Founders</title>
  </group>
  <group>
    <title>Rails Contributors</title>
  </group>
</groups>
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
-Make sure to configure your server to support these options following the
-instructions in the `Rack::Sendfile` documentation.
-
-[`config.action_dispatch.x_sendfile_header`]: configuring.html#config-action-dispatch-x-sendfile-header
+This will also change [`config.debug_exception_response_format`][] to `:api`. To
+preserve rendering errors as HTML pages, set it to `:default` in your
+`config/environments/development.rb`.
 
-### Using ActionDispatch::Request
+Change your `ApplicationController`'s super-class:
```
p8, line 338: This feels like it is missing a transitional adverb:
```suggestion
Lastly, change your `ApplicationController`'s super-class:
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+| [`ActionController::Caching`][]          | Enables controller-level caching techniques such as Russian Doll caching                                                                                                                         |
+| [`ActionController::DataStreaming`][]    | Support for `send_file` and `send_data`.                                                                                                                                                         |
+| [`ActionController::Instrumentation`][]  | Support for the instrumentation hooks defined by Action Controller (see [the instrumentation guide](active_support_instrumentation.html#action-controller) for more information regarding this). |
+| [`ActionController::ParamsWrapper`][]    | Wraps the parameters hash into a nested hash, so that you don't have to specify root elements sending POST requests for instance.                                                                |
```
p8, line 438:
```suggestion
| [`ActionController::ParamsWrapper`][]    | Wraps the parameters hash into a nested hash, so that you don't have to specify root elements when sending POST requests for instance.                                                                |
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+| [`ActionController::ApiRendering`][]     | Basic support for rendering.                                                                                                                                                                     |
+| [`ActionController::Renderers::All`][]   | Support for `:json`, `:xml` and similar renderers.                                                                                                                                               |
+| [`ActionController::ConditionalGet`][]   | Support for `stale?`.                                                                                                                                                                            |
+| [`ActionController::StrongParameters`][] | Support for parameters filtering in combination with Active Model mass assignment.                                                                                                               |
```
p8, line 434: Some of these sentences start with a verb, and other not
```suggestion
| [`ActionController::Redirecting`][]      | Adds support for `redirect_to`.                                                                                                                                                                       |
| [`ActionController::ApiRendering`][]     | Adds basic support for API rendering.                                                                                                                                                                     |
| [`ActionController::Renderers::All`][]   | Adds support for `:json`, `:xml` and similar renderers.                                                                                                                                               |
| [`ActionController::ConditionalGet`][]   | Adds support for `stale?`.                                                                                                                                                                            |
| [`ActionController::StrongParameters`][] | Adds support for parameters filtering in combination with Active Model mass assignment.                                                                                                               |
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+| [`ActionController::Caching`][]          | Enables controller-level caching techniques such as Russian Doll caching                                                                                                                         |
+| [`ActionController::DataStreaming`][]    | Support for `send_file` and `send_data`.                                                                                                                                                         |
+| [`ActionController::Instrumentation`][]  | Support for the instrumentation hooks defined by Action Controller (see [the instrumentation guide](active_support_instrumentation.html#action-controller) for more information regarding this). |
```
p8, line 437:
```suggestion
| [`ActionController::DataStreaming`][]    | Adds support for `send_file` and `send_data`.                                                                                                                                                         |
| [`ActionController::Instrumentation`][]  | Adds support for the instrumentation hooks defined by Action Controller (see [the instrumentation guide](active_support_instrumentation.html#action-controller) for more information regarding this). |
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+| [`ActionController::ParamsWrapper`][]    | Wraps the parameters hash into a nested hash, so that you don't have to specify root elements sending POST requests for instance.                                                                |
+| [`ActionController::Head`][]             | Support for returning a response with no content, only headers.                                                                                                                                  |
```
p8, line 439:
```suggestion
| [`ActionController::Head`][]             | Adds support for returning a response with no content, only headers.                                                                                                                                  |
```

guides/source/api_app.md
```diff
@@ -226,328 +156,360 @@ class GroupsController < ApplicationController
...
+class ApplicationController < ActionController::API
+  include AbstractController::Translation
+end
+```
 
-The best place to add a module is in your `ApplicationController`, but you can
-also add modules to individual controllers.
+Some useful modules you may wish to include are:
+
+|                                                                       |                                                                                                                                            |
```
p8, line 492: This table is pretty wide and doesn't look that great on smaller screens, requiring horizontal scrolling:

Image

I don't have a good solution here, other than not using a table or maybe removing the namespacing for the Auth modules.

guides/source/api_app.md
ayushn21: Does Shopify work this way though? Thought their core app was a Rails monolith with SSR?

guides/source/api_app.md
ayushn21: Ahhh great catch .... tweaked the wording a bit from your suggestion though :)

guides/source/api_app.md
ayushn21: I've changed it to `group_name` rather than `title` just to be a bit clearer, but nice idea for this demo :)

guides/source/api_app.md
ayushn21: Hmmmm, let me have a think

p8:
@ayushn21 The build is currently failing because of a broken link:
https://buildkite.com/rails/rails/builds/133272/list?sid=01a07bef-133f-46bb-ac76-0148d69c9495&tab=output#01a07bef-13a9-4418-8f29-f0bda8fd7408/L1601

guides/source/api_app.md
ayushn21: There's no reason for this to be a table actually. I'm going to reformat as a list, it should be easier to read that way.

ayushn21:
@p8 strange build failure as it doesn't seem to be from the changes in this guide. I've rebased on `main` which should hopefully sort it.

I've also reformatted the table as a list as mentioned in the earlier comment.

PR #57574, opened 2026-06-04
Make the backtrace cleaner ractor shareable by default

andrewn617:
### Motivation / Background

We are trying to make `Rails.application` ractor shareable. It contains a backtrace cleaner with some default filters and silencers which are not shareable. 

### Detail

Make the backtrace cleaner's default silencers and filters shareable with `Ractor.shareable_proc`.

The developer can add filters and we can't be sure if they are expecting to call methods on `self`. So we can't safely make these shareable by default since self is changed to `nil` inside of a shareable proc. If we did so for a proc calling a method on self it would only raise at call time which is a big footgun.

Perhaps we could update these apis to offer a way to opt in to making them shareable by default, or use prism to check if the proc calls self, but I don't have fully formed idea to propose yet.

On the other hand, for class macro type things where we `instance_exec` the block, we can make those shareable by default and we have a PR to do that on the way.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/testing/ractors_assertions.rb
+++ b/activesupport/lib/active_support/testing/ractors_assertions.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ActiveSupport
+  module Testing
+    module RactorsAssertions # :nodoc: all
+      private
+        if RUBY_VERSION >= "4.0"
+          def assert_ractor_shareable(obj)
+            assert_nothing_raised { Ractor.make_shareable(obj) }
+          end
+        else
+          def assert_ractor_shareable(obj)
+            assert true
+          end
+        end
+    end
+  end
+end
--- a/activesupport/test/backtrace_cleaner_test.rb
+++ b/activesupport/test/backtrace_cleaner_test.rb
@@ -2,6 +2,7 @@
 
 require_relative "abstract_unit"
 require "rails/test_unit/line_filtering"
+require "active_support/testing/ractors_assertions"
 
 class BacktraceCleanerFilterTest < ActiveSupport::TestCase
   def setup
@@ -93,6 +94,8 @@ class BacktraceCleanerFilterAndSilencerTest < ActiveSupport::TestCase
 end
 
 class BacktraceCleanerDefaultFilterAndSilencerTest < ActiveSupport::TestCase
+  include ActiveSupport::Testing::RactorsAssertions
+
   def setup
     @bc = ActiveSupport::BacktraceCleaner.new
   end
@@ -135,6 +138,10 @@ class BacktraceCleanerDefaultFilterAndSilencerTest < ActiveSupport::TestCase
     backtrace = [Gem.default_dir, *Gem.path].map { |path| "/parent#{path}/gems/nosuchgem-1.2.3/lib/foo.rb" }
 
     assert_equal backtrace, @bc.clean(backtrace)
+  end
+
+  test "backtrace cleaner is Ractor shareable" do
+    assert_ractor_shareable @bc
   end
 end
 
--- a/railties/lib/rails/backtrace_cleaner.rb
+++ b/railties/lib/rails/backtrace_cleaner.rb
@@ -7,25 +7,35 @@ module Rails
   class BacktraceCleaner < ActiveSupport::BacktraceCleaner # :nodoc:
     RENDER_TEMPLATE_PATTERN = /:in [`'].*_\w+_{2,3}\d+_\d+'/
 
+    class << self
+      def root
+        @root ||= Rails.root && "#{Rails.root}/"
+      end
+    end
+
     def initialize
       super
-      add_filter do |line|
+      add_filter(&ActiveSupport::Ractors.shareable_proc do |line|
         # We may be called before Rails.root is assigned.
         # When that happens we fallback to not truncating.
-        @root ||= Rails.root && "#{Rails.root}/"
-        @root && line.start_with?(@root) ? line.from(@root.size) : line
-      end
-      add_filter do |line|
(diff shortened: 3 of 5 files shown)


railties/lib/rails/backtrace_cleaner.rb
```diff
@@ -9,23 +9,23 @@ class BacktraceCleaner < ActiveSupport::BacktraceCleaner # :nodoc:
...
       super
-      add_filter do |line|
+      add_filter(&ActiveSupport::Ractors.shareable_proc do |line|
         # We may be called before Rails.root is assigned.
         # When that happens we fallback to not truncating.
-        @root ||= Rails.root && "#{Rails.root}/"
-        @root && line.start_with?(@root) ? line.from(@root.size) : line
-      end
-      add_filter do |line|
+        root = Rails.root && "#{Rails.root}/"
```
rafaelfranca: Can we avoid calculating this string for each line we are trying to filter? That seems expensive.

andrewn617: I think it's tricky because even if we eagerly evaluate it (we could do it when we `freeze` this since we know `Rails.root` will be set by then), the backtrace instance is not the `self` of this proc, and we can't make it the self since we are still in the middle of initializing it, so it can't be frozen. I think we could put it on in a class instance var, and make sure that is evaluated when we freeze the instance.

But let me sleep on it to see if I think of anything better.

activesupport/test/backtrace_cleaner_test.rb
```diff
@@ -280,4 +280,12 @@ def indirect_caller_locations
 
     assert_equal [], @bc.clean_locations(@locations)
   end
+
+  if RUBY_VERSION >= "4.0"
+    test "backtrace cleaner is Ractor shareable" do
+      assert_nothing_raised do
+        ActiveSupport::Ractors.make_shareable(@bc)
```
byroot: You should add some sort of `assert_shareable` method, that would include the Ruby version check, etc.

Otherwise we'll end up with thousands of Ruby version checks across the test suite.

andrewn617: Yeah was thinking about this was just gonna wait till I had some more cases, but I'll add it. I assume we want to no-doc that for now like the helpers?

byroot: > was just gonna wait till I had some more cases

You can modify the helper as you add more cases if its API no longer work well.

> no-doc that for now like the helpers?

Yes, a private test helper, we already have a bunch of those.

railties/lib/rails/backtrace_cleaner.rb
andrewn617: I put on class ivar. The other option is to put in a method on the instance, but then we need to call with like `Rails.backtrace_cleaner.root` in the block, but seemed weird to make this class know where the instance lives on Rails 🤷

railties/lib/rails/backtrace_cleaner.rb
```diff
@@ -7,25 +7,35 @@ module Rails
   class BacktraceCleaner < ActiveSupport::BacktraceCleaner # :nodoc:
     RENDER_TEMPLATE_PATTERN = /:in [`'].*_\w+_{2,3}\d+_\d+'/
 
+    class << self
+      def root
+        @root ||= Rails.root && "#{Rails.root}/"
```
byroot, line 12: That will fail if a Ractor calls this while `@root` if still nil.

You might need something like:

```suggestion
        return @root if defined?(@root) 
        return unless Ractor.main?
        @root ||= Rails.root && "#{Rails.root}/"
```

andrewn617: I see what you're saying, but I don't think there would be a valid case when a non-main Ractor is calling this before the application has been frozen (and thus `Rails.root` assigned). But I can still add this.

andrewn617: And sorry, I know I haven't made the big picture clear in these PRs. Soon I will open a PR proposing how we would how we freeze the application, I am just getting some pieces in place first - tho maybe it would be better to have it in one PR.

But the mental model I am working with is that we will boot the app, then freeze `Rails.application`, then create ractors to serve requests. So I am thinking calling this from a non-main Ractor before the app is fully booted shouldn't happen.

Still we could put the guard, but we do memoizations like that in a lot of places on the boot path. I was kind of thinking it would be a matter of convention that they are not ever called on non-main ractor until we fully booted the app and they are all evaluated, so we would not need to put this guard everywhere.

gmcgibbon (approved):

activesupport/lib/active_support/testing/ractors_assertions.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "minitest/mock"
```
gmcgibbon:
```suggestion
```

activesupport/lib/active_support/testing/ractors_assertions.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Testing
+    module RactorsAssertions # :nodoc: all
+      private
+        def assert_ractor_shareable(obj)
+          if RUBY_VERSION >= "4.0"
+            assert_nothing_raised { Ractor.make_shareable(obj) }
+          else
+            assert true
+          end
+        end
```
byroot:
```suggestion
        if RUBY_VERSION >= "4.0"
          def assert_ractor_shareable(obj)
            assert_nothing_raised { Ractor.make_shareable(obj) }
          end
        else
          def assert_ractor_shareable(obj)
            assert true
          end
        end
```

PR #57579, opened 2026-06-04
Add RFC 9110 compliant Accept header content negotiation opt-in

### LINKED ISSUE #57353: Accept header intentionally ignored if it includes */*
### Steps to reproduce

Have a controller action like

```ruby
def index
  respond_to do |format|
    format.html
    format.json
  end
end
```

Make a request including the following Accept header (e.g. via curl):

```
Accept: application/json, */*
```

### Expected behavior

According to RFC 9110, the priorization should be from most specific to least specific:

> Media ranges can be overridden by more specific media ranges or specific media types. If more than one media range applies to a given type, the most specific reference has precedence. For example,
>
> `Accept: text/*, text/plain, text/plain;format=flowed, */*`
>
> have the following precedence:
>
> 1. `text/plain;format=flowed`
> 2. `text/plain`
> 3. `text/*`
> 4. `*/*`

That rails follows this RFC could be assumed from reading the docs for [MimeResponds](https://api.rubyonrails.org/classes/ActionController/MimeResponds.html):

> Rails determines the desired response format from the HTTP Accept header submitted by the client.

### Actual behavior

Rails will render the HTML format, even though a more specific format requested by the client would have been available. This seems to be done on purpose, when checking the [source code of `MimeNegotation`](https://github.com/rails/rails/blob/f26f56ed4e19dea9119fcfe067eff36b7f0b5dc5/actionpack/lib/action_dispatch/http/mime_negotiation.rb#L229-L232). Quoting the code comment:

> We use normal content negotiation unless you include `**/**` (sic) in your list, in which case we assume you're a browser and send HTML.

I am aware that this was originally reported as https://github.com/rails/rails/issues/8987 and back then it was rejected, because the workaround has been added intentionally for "browsers [with] broken accept headers".

Since this was... _checking notes_... **13 years ago**, I wanted to ask whether this work around is still necessary. I am not sure which other browsers have problems following the RFC, but IE7 shouldn't be an issue anymore...

In contrast, this came up for me, because we were removing non-standard headers from our code that we thought were not needed anymore. I was very surprised to see that the content negotation was influenced by the non-standard `X-Requested-With` HTTP header. It also didn't help that the docs (probably in an attempt of keeping things simple) didn't mention that other headers than `Accept` were involved in choosing which format to deliver. So the only way to figure this out is through checking the Rails source code or through finding old bug reports about it. It can't be figured out by looking at internet standards or the Rails documentation.

### System configuration

**Rails version**: 8.1.3 (code is still present in `main` branch)

**Ruby version**: 4.0.2

williantenfen:
### Motivation / Background

This Pull Request adds RFC 9110 compliant Accept header content negotiation while maintaining backward compatibility. Rails has used a workaround since 2010 that assumes any Accept header containing `*/*` is from a browser and defaults to HTML. This was necessary for IE7, but modern browsers properly follow RFC 9110 and no longer need this special handling.

  The feature is opt-in via `config.action_dispatch.respect_accept_header_rfc9110` and defaults to `false`, ensuring no breaking changes for existing applications.

Added as opt-in flag, future versions can deprecate and major version can remove workaround for IE7 I think... 

  Fixes #57353

### Detail

This Pull Request adds:

  1. **New Configuration Flag** (`config.action_dispatch.respect_accept_header_rfc9110`)
     - Defaults to `false` for backward compatibility
     - When enabled, respects RFC 9110 media type specificity and quality values
     - When disabled, uses the existing IE7 workaround (`BROWSER_LIKE_ACCEPTS` filter)

  2. **Configuration Conflict Warning**
     - If both `ignore_accept_header` and `respect_accept_header_rfc9110` are enabled, a warning is logged
     - Explains that `ignore_accept_header` takes precedence

  3. **Upgrade Path**
     - Documentation in `new_framework_defaults_8_2.rb.tt` with commented-out opt-in
     - Clear explanation of the change and its impact
     - CHANGELOG entry for users

  4. **Comprehensive Tests**
     - `test_rfc9110_accept_header_respects_specificity`: Verifies specific types are prioritized over wildcards
     - `test_rfc9110_with_xhr_request`: Ensures XHR requests work correctly in both modes

### Additional information

  **Example Use Case:**
  ```ruby
  # Before (current behavior):
  # Accept: application/json, */*
  # Returns: HTML (due to IE7 workaround)

  # After (with respect_accept_header_rfc9110 = true):
  # Accept: application/json, */*
  # Returns: JSON (RFC 9110 compliant)
```

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/http/mime_negotiation.rb
+++ b/actionpack/lib/action_dispatch/http/mime_negotiation.rb
@@ -18,6 +18,7 @@ module ActionDispatch
 
       included do
         mattr_accessor :ignore_accept_header, default: false
+        mattr_accessor :strict_accept_header, default: false
       end
 
       # The MIME type of the HTTP request, such as [Mime](:xml).
@@ -227,8 +228,11 @@ module ActionDispatch
         end
 
         def valid_accept_header
-          (xhr? && (accept.present? || content_mime_type)) ||
-            (accept.present? && !accept.match?(BROWSER_LIKE_ACCEPTS))
+          if xhr?
+            accept.present? || content_mime_type
+          elsif accept.present?
+            self.class.strict_accept_header || !accept.match?(BROWSER_LIKE_ACCEPTS)
+          end
         end
 
         def use_accept_header
--- a/actionpack/lib/action_dispatch/railtie.rb
+++ b/actionpack/lib/action_dispatch/railtie.rb
@@ -16,6 +16,7 @@ module ActionDispatch
     config.action_dispatch.show_exceptions = :all
     config.action_dispatch.tld_length = 1
     config.action_dispatch.ignore_accept_header = false
+    config.action_dispatch.strict_accept_header = false
     config.action_dispatch.rescue_templates = {}
     config.action_dispatch.rescue_responses = {}
     config.action_dispatch.wrapper_exceptions = []
@@ -80,6 +81,7 @@ module ActionDispatch
 
       ActiveSupport.on_load(:action_dispatch_request) do
         self.ignore_accept_header = app.config.action_dispatch.ignore_accept_header
+        self.strict_accept_header = app.config.action_dispatch.strict_accept_header
         ActionDispatch::Request::Utils.perform_deep_munge = app.config.action_dispatch.perform_deep_munge
       end
 
--- a/actionpack/test/controller/mime/respond_to_test.rb
+++ b/actionpack/test/controller/mime/respond_to_test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "abstract_unit"
+require "active_support/core_ext/object/with"
 require "active_support/log_subscriber/test_helper"
 
 class RespondToController < ActionController::Base
@@ -606,6 +607,15 @@ class RespondToControllerTest < ActionController::TestCase
     @request.accept = "application/json, application/xml, */*"
     get :json_xml_or_html
     assert_equal "HTML", @response.body
+
+    ActionDispatch::Request.with(strict_accept_header: true) do
+      get :json_xml_or_html
+      assert_equal "JSON", @response.body
+
+      @request.accept = "application/xml, */*"
+      get :json_xml_or_html
+      assert_equal "XML", @response.body
+    end
   end
 
   def test_handle_any_with_template
--- a/guides/source/configuring.md
+++ b/guides/source/configuring.md
@@ -64,6 +64,7 @@ Below are the default values associated with each target version. In cases of co
 - [`config.action_controller.forgery_protection_verification_strategy`](#config-action-controller-forgery-protection-verification-strategy): `:header_only`
 - [`config.action_controller.rescue_from_event_backtrace`](#config-action-controller-rescue-from-event-backtrace): `:array`
 - [`config.action_dispatch.default_headers`](#config-action-dispatch-default-headers): `{ "X-Frame-Options" => "SAMEORIGIN", "X-Content-Type-Options" => "nosniff", "X-Permitted-Cross-Domain-Policies" => "none", "Referrer-Policy" => "strict-origin-when-cross-origin" }`
+- [`config.action_dispatch.strict_accept_header`](#config-action-dispatch-strict-accept-header): `true`
 - [`config.active_job.enqueue_after_transaction_commit`](#config-active-job-enqueue-after-transaction-commit): `true`
 - [`config.active_record.postgresql_adapter_decode_bytea`](#config-active-record-postgresql-adapter-decode-bytea): `true`
 - [`config.active_record.postgresql_adapter_decode_money`](#config-active-record-postgresql-adapter-decode-money): `true`
@@ -2309,6 +2310,19 @@ config.action_dispatch.domain_extractor = CustomDomainExtractor
 #### `config.action_dispatch.ignore_accept_header`
 
 Is used to determine whether to ignore accept headers from a request. Defaults to `false`.
+
+#### `config.action_dispatch.strict_accept_header`
+
+Controls whether an `Accept` header containing `*/*` forces an HTML response.
+When enabled, Rails honors more specific types instead — e.g. `Accept:
+application/json, */*` returns JSON instead of HTML.
+
+The default value depends on the `config.load_defaults` target version:
+
+| Starting with version | The default value is |
+| --------------------- | -------------------- |
+| (original)            | `false`              |
+| 8.2                   | `true`               |
 
 #### `config.action_dispatch.x_sendfile_header`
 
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by williantenfen** [comment]
@rafaelfranca 🇧🇷

File: actionpack/lib/action_dispatch/http/mime_negotiation.rb [hunk 1]
```diff
@@ -227,8 +228,16 @@ def params_readable?
...
+          if self.class.respect_accept_header_rfc9110
+            # RFC 9110 mode: Accept header is valid if present, respecting specificity.
+            # XHR requests get special handling to work without an Accept header.
+            (xhr? && (accept.present? || content_mime_type)) || accept.present?
+          else
+            # Legacy mode: Assume any Accept header with */* is from a browser,
+            # defaulting to HTML. This workaround was added for IE7 (2010).
+            (xhr? && (accept.present? || content_mime_type)) ||
+              (accept.present? && !accept.match?(BROWSER_LIKE_ACCEPTS))
+          end
```
skipkayhil: I think breaking this up with `if` makes sense, but we shouldn't duplicate the rest of the conditions. Maybe something like:
```suggestion
          if xhr?
            accept.present? || content_mime_type
          elsif accept.present?
            self.class.respect_accept_header_rfc9110 || !accept.match?(BROWSER_LIKE_ACCEPTS))
          end
```

railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
```diff
@@ -101,3 +101,24 @@
...
+#
+# Modern browsers send properly formatted Accept headers where HTML has appropriate
+# priority without relying on */* detection. API clients that include */* as a fallback
+# will now get their requested format instead of HTML.
+#
+# Note: This setting only applies when `config.action_dispatch.ignore_accept_header`
+# is false (the default). If you've set `ignore_accept_header` to true, the Accept
+# header is completely ignored regardless of this setting.
+#++
+# Rails.application.config.action_dispatch.respect_accept_header_rfc9110 = true
```
skipkayhil: If we want new apps to get this value then we should add to `load_defaults` as well

actionpack/lib/action_dispatch/http/mime_negotiation.rb
williantenfen: updated, thanks!

railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_8_2.rb.tt
williantenfen: got it, updated - thank you!

actionpack/lib/action_dispatch/railtie.rb
```diff
@@ -76,10 +77,22 @@ class Railtie < Rails::Railtie # :nodoc:
...
+      if app.config.action_dispatch.ignore_accept_header &&
+         app.config.action_dispatch.respect_accept_header_rfc9110
+        (Rails.logger || ActiveSupport::Logger.new($stdout)).warn(
+          "WARNING: Both 'ignore_accept_header' and 'respect_accept_header_rfc9110' are enabled. " \
+          "'ignore_accept_header' takes precedence, disabling Accept header processing entirely. " \
+          "The 'respect_accept_header_rfc9110' setting will have no effect. " \
+          "Remove one of these configurations."
+        )
+      end
+
```
skipkayhil: I'm not sure its worth warning (and I think it probably makes things more complicated).

If an app updates to `load_defaults 8.2` and has `ignore_accept_header = true`, they would then have to also set `respect_accept_header_rfc9110 = false`, which seems unnecessary.

guides/source/configuring.md
```diff
@@ -2310,6 +2311,14 @@ config.action_dispatch.domain_extractor = CustomDomainExtractor
...
 Is used to determine whether to ignore accept headers from a request. Defaults to `false`.
 
+#### `config.action_dispatch.respect_accept_header_rfc9110`
+
+When enabled, respects the Accept header according to RFC 9110, prioritizing more specific media types over wildcards. For example, `Accept: application/json, */*` will return JSON instead of HTML.
+
+Previously, Rails assumed any Accept header containing `*/*` was from a browser and would default to HTML. This workaround was necessary for IE7 but is no longer required for modern browsers.
+
+Defaults to `true` for new Rails 8.2 applications. For applications upgrading from earlier versions, this defaults to `false` to maintain backward compatibility. Set to `true` to enable RFC 9110 compliant behavior.
+
```
skipkayhil, line 2326: We should add the `(original)`/`8.2` table to this description like other framework defaults.

skipkayhil:
One last thing: I think we may be overwhelming users with discussion of an RFC. I think we could simplify the framing of this change as "opt out of forcing HTML when Accept contains the wildcard format", which is much easier to understand in my opinion. I'd be open to alternatives here as well, but I do think the current amount of text may be too much.

williantenfen:
> One last thing: I think we may be overwhelming users with discussion of an RFC. I think we could simplify the framing of this change as "opt out of forcing HTML when Accept contains the wildcard format", which is much easier to understand in my opinion. I'd be open to alternatives here as well, but I do think the current amount of text may be too much.

Makes sense, should we then change flag name? some options:
 - skip_forcing_html_response_on_wildcard_presence
 - skip_forcing_html_on_wildcard_accept
 - opt_out_forcing_html_response_on_wildcard_presence
 - opt_out_wildcard_html_fallback

Thanks, meanwhile I will do the other changes...

actionpack/test/controller/mime/respond_to_test.rb
```diff
@@ -608,6 +608,41 @@ def test_browser_check_with_any_any
     assert_equal "HTML", @response.body
   end
 
+  def test_rfc9110_accept_header_respects_specificity
```
williantenfen: @skipkayhil should there be new tests cases for this new flag or should we somehow update existing ones to assert over depending on the flag value?

williantenfen: sorry, just looking into conventions - will have a commit with proposed change for this

skipkayhil: I think what you did makes sense: keep existing assertions but add some new ones for the new behavior 👍

skipkayhil:
I took the liberty of force pushing with a new config name: `strict_accept_header` (and expanded the commit message with more context, I love going through the history of things like this 😅).

Let me know if it looks good to you and I'll merge!

williantenfen:
> I took the liberty of force pushing with a new config name: `strict_accept_header` (and expanded the commit message with more context, I love going through the history of things like this 😅).
> 
> Let me know if it looks good to you and I'll merge!

awesome! thank you!

hardest part: naming variables/things 😂

Edgarruiz8585:
Qk yes Edgarruiz8585
+ > ***@***.***
.

skipkayhil (approved):

skipkayhil:
Thanks again for the PR and contributing to Rails! 🎉

PR #57583, opened 2026-06-04
Fix plain English explanation of nested query in Active Record Querying guide [ci-skip]

VladNegara:
### Motivation / Background

This Pull Request has been created because [the current explanation for nested association joining](https://guides.rubyonrails.org/active_record_querying.html#joining-nested-associations-multiple-level) in the Active Record Querying guide is plain wrong.

### Detail

This Pull Request changes the explanation in the aforementioned subsection to reflect the actual structure of the query.

To explain how the plain English query deviates from the Active Record and SQL forms, consider a scenario in which an author's books have never been ordered, and have received reviews only from customers who haven't bought it (arguably weird, but not explicitly disallowed). In that case, the author would still be a part of the result set of the Active Record and SQL queries, while the plain English version excludes it.

Furthermore, the plain English explanation seems to suggest that details about suppliers are returned by the query. In fact, the join only acts as a filter to consider only books which have at least one supplier.

### Additional information

It seem like PR #56341 (not yet live on the deployed guides) updated the phrasing of the explanation, while leaving the content unchanged.

### Checklist

### DIFF
--- a/guides/source/active_record_querying.md
+++ b/guides/source/active_record_querying.md
@@ -2518,8 +2518,8 @@ SELECT authors.* FROM authors
   INNER JOIN suppliers ON suppliers.id = books.supplier_id
 ```
 
-The SQL query will return all authors that have books with reviews _and_ have
-been ordered by a customer, _and_ the suppliers for those books.
+The SQL query will return all authors that have a book which has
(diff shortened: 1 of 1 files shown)


p8:
Thanks @VladNegara !

PR #57590, opened 2026-06-05
Fix order-dependent test failures with `sqlite3_mem`

### LINKED ISSUE #57589: `clear_cache!` corrupts SchemaCache for subsequent tests, causing "no such table" errors.
### Steps to reproduce

Run the full Active Record test suite against SQLite3 in-memory:

```bash
ARCONN=sqlite3_mem bundle exec rake sqlite3_mem:test
```

The failures are **order-dependent**. `DirtyTest#test_field_named_field` must run before the affected tests. Its `ensure` block calls `ActiveRecord::Base.clear_cache!`, which leaves the `SchemaCache` in a broken state for all subsequent tests in the same process:

```ruby
# test/cases/dirty_test.rb
def test_field_named_field
  ActiveRecord::Base.lease_connection.create_table :testings do |t|
    t.string :field
  end
ensure
  ActiveRecord::Base.lease_connection.drop_table :testings rescue nil
  ActiveRecord::Base.clear_cache!  # everything breaks after this
end
```

### Expected behavior

`clear_cache!` should not leave `SchemaCache` in a state where subsequent tests cannot resolve existing tables. After a cache clear, column and table lookups should re-query the database and repopulate the cache transparently.

### Actual behavior

Every test running after `test_field_named_field` in the same process fails. Two distinct failure modes are observed, indicating the cache is not simply cleared but left in a poisoned state:

**Hard raise** — `SchemaCache#columns` does a strict `Hash#fetch` with no fallback, so any table lookup raises instead of re-querying:
```
ActiveRecord::StatementInvalid: Could not find table 'posts'
ActiveRecord::StatementInvalid: Could not find table 'humans'
# ... hundreds more
```

**Silent empty columns** — some code paths get an empty columns hash without raising, so models load with no attributes at all:
```
ActiveModel::UnknownAttributeError: unknown attribute 'name' for Bulb.
NoMethodError: undefined method 'foo=' for an instance of #<Class:...>(id: integer, foo: )
```

The second failure mode is more insidious: the cache returns stale/empty column data rather than raising, causing models to be defined with no columns. This suggests `clear_cache!` sets the cache to an empty frozen structure rather than `nil`/sentinel that would trigger re-introspection.

Total failures across the full rake run: ~3000 errors spanning virtually every test file.

### System configuration

**Rails version**: main

**Ruby version**: 4.0.5

ruyrocha:
### Motivation / Background

This Pull Request has been created because running `ARCONN=sqlite3_mem bundle exec rake sqlite3_mem:test` produced ~3000 order-dependent test failures, all tracing back to two tests that are incompatible with SQLite in-memory databases. Individual test files passed in isolation, but the full suite cascaded into failures because these tests destroyed shared state that subsequent tests depended on.

Fixes #57589

### Detail

This Pull Request fixes two independent root causes:

**1. `DirtyTest#test_field_named_field` — SchemaCache corruption**

The test called `ActiveRecord::Base.clear_cache!` in its `ensure` block after `create_table`/`drop_table`. This replaced the global SchemaCache with an empty instance, causing "Could not find table" errors in all subsequent tests. Rewritten to use an anonymous class with the `attribute` API on the existing `pirates` table, following the patterns already established in the same file (e.g. `test "attribute_will_change! doesn't try to save non-persistable attributes"`). The top-level `Testings` class definition is also removed.

**2. `AdapterConnectionTest` sql_notifications tests — Database destruction**

Two tests (`test_suppresses_notifications_when_sql_notifications=false` and `test_sql_notifications_are_enabled_by_default`) were outside the `unless in_memory_db?` guard that protects the rest of `AdapterConnectionTest`. They use `run_without_connection`, which calls `remove_connection` followed by `establish_connection`. On SQLite `:memory:` databases, `remove_connection` destroys the in-memory database, and the subsequent `establish_connection` creates a fresh empty one with no schema. Wrapped both tests with `unless in_memory_db?`, consistent with the rest of the class.

### Additional information

The `unconnected_test.rb` already demonstrates the correct pattern for tests that must `remove_connection` on in-memory databases — it calls `load_schema if in_memory_db?` in its teardown to restore the schema after reconnecting.

### Checklist

* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #57589]`

### DIFF
--- a/activerecord/test/cases/adapter_test.rb
+++ b/activerecord/test/cases/adapter_test.rb
@@ -977,24 +977,26 @@ module ActiveRecord
       end
     end
 
-    test "suppresses notifications when sql_notifications=false" do
-      run_without_connection do |orig_connection|
-        ActiveRecord::Base.establish_connection(orig_connection.merge(sql_notifications: false))
+    unless in_memory_db?
+      test "suppresses notifications when sql_notifications=false" do
+        run_without_connection do |orig_connection|
+          ActiveRecord::Base.establish_connection(orig_connection.merge(sql_notifications: false))
 
+          notifications = capture_notifications("sql.active_record") do
+            Post.first
+          end
+
+          assert_empty notifications
+        end
+      end
+
+      test "sql_notifications are enabled by default" do
         notifications = capture_notifications("sql.active_record") do
           Post.first
         end
 
-        assert_empty notifications
+        assert_not_empty notifications
       end
-    end
-
-    test "sql_notifications are enabled by default" do
-      notifications = capture_notifications("sql.active_record") do
-        Post.first
-      end
-
-      assert_not_empty notifications
     end
   end
 
--- a/activerecord/test/cases/dirty_test.rb
+++ b/activerecord/test/cases/dirty_test.rb
@@ -607,17 +607,15 @@ class DirtyTest < ActiveRecord::TestCase
     assert_not pirate.previous_changes.key?("created_on")
   end
 
-  class Testings < ActiveRecord::Base; end
   def test_field_named_field
-    ActiveRecord::Base.lease_connection.create_table :testings do |t|
-      t.string :field
+    klass = Class.new(ActiveRecord::Base) do
+      self.table_name = "pirates"
+      attribute :field, :string
     end
+
     assert_nothing_raised do
-      Testings.new.attributes
+      klass.new.attributes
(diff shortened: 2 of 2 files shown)


ruyrocha:
ping @byroot

PR #57594, opened 2026-06-05
Fix bug where reload leaks ordinary scopes into all_queries lookups.

andrewn617:
### Motivation / Background

While working on https://github.com/rails/rails/pull/57123 @matthewd caught an existing bug on main. When we reload a model with `all_queries` default scopes, we call `all(all_queries)`. But `all` includes queries from the `current_scope` which means we can leak non `all_queries` scopes into what should be an all queries only query.

### Detail

Instead we should call `default_scoped(all_queries: true)` directly. Then we can check the `current_global_scope` to include any current `all_queries` scopes.

### Checklist


gmcgibbon (commented):
Looks like this is missing a changelog. Please add one.

PR #57597, opened 2026-06-06
Reflect a variant's format in its representation URL filename

### LINKED ISSUE #57596: Active Storage: wrong file extension for variant when you use the format option
### Steps to reproduce

```ruby
variant = item.image.variant(
   resize_to_fill: [600, 300],
   format: 'webp',
   saver: { quality: 80, strip: true }
)

image_tag variant
```

The variant is in webp format (correct mime type), but the image url displayed in the view ends with `.jpg` extension (like the original image).

### Expected behavior

In order to avoid issues with CDN, tools, browser downloads, etc. the file name should end with `.webp` (i.e. the variant format).

### Actual behavior

The original file extension is used.

### System configuration

**Rails version**: 8.1

**Ruby version**: 4.0

dbaratta:
### Motivation / Background

Fixes #57596.

When you create a variant that changes the format, the URL still comes out with the original blob's extension:

```ruby
variant = item.image.variant(resize_to_fill: [600, 300], format: "webp")
image_tag variant
```

The image itself is a correct `webp`, but the URL ends in `.jpg`. That trips up CDNs, download prompts, and anything that keys off the extension.

### Detail

`rails_storage_redirect` and `rails_storage_proxy` build the URL's trailing filename from `model.blob.filename` (the original blob) instead of `model.filename` (the representation), so the variant's format never made it into the URL. The representation's `#filename` already encodes the format, so I pointed both helpers at it.

Nothing changes for a variant that keeps its format: `blob.variant` defaults the variation format to the blob's own format, so `model.filename` still ends in `.jpg` there.

Previews needed some care. `Preview#filename` used to delegate to its (processed) `presentation`, so pointing the routes at `model.filename` would make URL generation either raise on an unprocessed preview or, worse, process the variant inline (a `service.exist?` check, or a full transform and upload when the preview image already exists). Building a URL shouldn't touch the service. A preview's filename is just `base + format`, so `Preview#filename` now builds that directly without forcing processing, and falls back to the blob's filename until the preview has run. Unprocessed preview URLs are unchanged; once a preview is processed its URL picks up the preview image's real format (`.png` for a PDF, `.jpg` for a video), which is what gets served anyway.

### Additional information

Added URL helper tests on both the redirect and proxy sides (variant changing and keeping its format, a tracked `VariantWithRecord`, and processed/unprocessed previews), plus `Preview#filename` tests for the fallback and for not processing the variant when it's read.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Reflect a variant's format in its representation URL filename.
+
+    When a variant changed the format (e.g. `variant(format: :webp)`), the generated
+    URL still used the original blob's filename, so it ended with the original
+    extension (e.g. `.jpg`) instead of the variant's (e.g. `.webp`). This could
+    confuse CDNs, download tools, and browsers. The URL now uses the variant's
+    filename, which reflects the variant format.
+
+    Processed preview URLs likewise reflect the preview image's format (e.g. `.png`
+    for a PDF, `.jpg` for a video) rather than the source file's extension, matching
+    what is served. Unprocessed previews still fall back to the blob's filename, so
+    generating a URL never triggers preview processing.
+
+    *Dominic Baratta*
+
 *   Fix `MirrorService#mirror` losing blob metadata when copying to mirrors.
 
     Mirrored copies on S3, Azure, and GCS were served as `application/octet-stream`
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,18 @@
+*   Reflect a variant's format in its representation URL filename.
+
+    When a variant changed the format (e.g. `variant(format: :webp)`), the generated
+    URL still used the original blob's filename, so it ended with the original
+    extension (e.g. `.jpg`) instead of the variant's (e.g. `.webp`). This could
+    confuse CDNs, download tools, and browsers. The URL now uses the variant's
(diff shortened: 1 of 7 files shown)


dbaratta:
Rebased on current main — would appreciate a review @janko.

janko:
@dbaratta I don't actually know Active Storage that well, I only maintain the `image_processing` dependency, which is used by other file attachment libraries as well 😉

dbaratta:
> @dbaratta I don't actually know Active Storage that well, I only maintain the `image_processing` dependency, which is used by other file attachment libraries as well 😉

Ahh, I'm a bit new to trying to get stuff into Rails :) ... I mostly pinged you since well you were the merge conflict I had to resolve on the change log. Do you know if I need to ping someone on this or just sorta hang out and eventually someone will come along and pick it up? 

I suspect this one might spark some discussion about part of the bug fix being over in the model due to having to work around some weird things that happen before a file is processed.

dbaratta:
Gentle bump on this one. It fixes #57596 (a variant's format not reflected in its representation URL filename) and has tests on both the redirect and proxy helpers.

@guilleiguaran since you touched the redirect/proxy mode warnings recently, would you be able to take a look? One part I'd want eyes on: the preview fix changes `Preview#filename` so building a URL doesn't force processing, which brushes up against the analyze-before-validation behavior. Happy to adjust the approach there.

PR #57598, opened 2026-06-06
Fix `find` with an explicit `order` + `offset` past the ids raising a spurious `RecordNotFound`

55728:
### Summary

`find` with an array of ids, an explicit `order`, and an `offset` greater than the number of ids raises a spurious `RecordNotFound` with a nonsensical negative count:

```ruby
Model.order(:id).offset(10).find([1, 2, 3])
# => ActiveRecord::RecordNotFound: Couldn't find all Models with 'id':
#    (1, 2, 3) (found 0 results, but was looking for -7).
```

The same query **without** an `order` correctly returns `[]`.

### Cause

An explicit `order` routes `find` through `find_some` (no order goes through `find_some_ordered`). `find_some` computes the expected number of results and adjusts it for the offset:

```ruby
# 11 ids with limit 3, offset 9 should give 2 results.
if offset_value && (ids.size - offset_value < expected_size)
  expected_size = ids.size - offset_value
end
```

When `offset_value` is greater than `ids.size`, `ids.size - offset_value` is negative, so `expected_size` becomes negative. `result.size` (`0`) then never equals the negative `expected_size`, and `find` raises `RecordNotFound`.

The semantically-equivalent no-order path slices the ids by offset/limit and returns `[]` for the same input, so the ordered path is an inconsistent outlier that crashes on ordinary pagination-past-the-end input. The boundary case `offset == ids.size` already returns `[]` on both paths.

### Fix

Clamp the offset-adjusted expected size at zero:

```ruby
expected_size = [ids.size - offset_value, 0].max
```

With this, `result.size` (`0`) `== 0` and `find` returns `[]`, matching the no-order path. The existing offset edge case (11 ids, limit 3, offset 9 → 2 results: `max(2, 0) == 2`) is unaffected.

### Test

Added two regression tests to `finder_test.rb`:

- `test_find_with_order_and_offset_past_the_ids_returns_empty` asserts that an offset past the number of ids returns `[]` on both the ordered (`find_some`) and unordered (`find_some_ordered`) paths, and covers the `offset == ids.size` boundary.
- `test_find_with_order_limit_and_offset_matches_unordered_path` exercises seven `limit`/`offset` combinations against all 11 Developer fixture ids, verifying that the ordered path (`find_some`) returns the same records as the unordered path (`find_some_ordered`) — including offsets within range, limits past the end, the boundary `offset == ids.size`, and offsets past the end (which used to raise).

Verified fail → pass on **sqlite3**, **postgresql**, and **mysql2**; the full `finder_test` suite is green on all three adapters.

### Notes

* No upstream issue/PR found.
* Behaviour change is limited to a previously-raising edge case (offset strictly greater than the number of ids passed to `find`, with an explicit order); it now returns `[]` instead of raising.

### DIFF
--- a/activerecord/lib/active_record/relation/finder_methods.rb
+++ b/activerecord/lib/active_record/relation/finder_methods.rb
@@ -542,10 +542,6 @@ module ActiveRecord
       def find_some(ids)
         return find_some_ordered(ids) unless order_values.present?
 
-        relation = where(primary_key => ids)
-        relation = relation.select(table[primary_key]) unless select_values.empty?
-        result = relation.to_a
-
         expected_size =
           if limit_value && ids.size > limit_value
             limit_value
@@ -554,9 +550,17 @@ module ActiveRecord
           end
 
         # 11 ids with limit 3, offset 9 should give 2 results.
-        if offset_value && (ids.size - offset_value < expected_size)
-          expected_size = ids.size - offset_value
+        if offset_value
+          if ids.size <= offset_value
+            return []
+          elsif ids.size - offset_value < expected_size
+            expected_size = ids.size - offset_value
+          end
         end
+
+        relation = where(primary_key => ids)
+        relation = relation.select(table[primary_key]) unless select_values.empty?
+        result = relation.to_a
 
         if result.size == expected_size
           result
@@ -567,6 +571,7 @@ module ActiveRecord
 
       def find_some_ordered(ids)
         ids = ids.slice(offset_value || 0, limit_value || ids.size) || []
+        return [] if ids.empty?
 
         relation = except(:limit, :offset)
         relation = relation.where(primary_key => ids)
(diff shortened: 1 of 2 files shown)


fatkodima:
I changed the code a bit to make it more explicit about the offset boundaries. I also moved a few lines around to avoid a useless SQL query if it is known that it will not have any results (when the offset is >= than the number of ids).

Also note, that we tend to not add changelog entries for [simple] bug fixes, only for new features, new deprecations or something very likely to be interesting for the user. The reason is that when releasing a new major/minor version, the changelog files can become too large (especially for activerecord) and noisy and when reading important stuff can be missed by developers. I merged a few PRs with changelogs, but they will probably be removed before the final release.

Btw, great work, keep going! ❤️

55728:
Thank you @fatkodima — the early return to skip the useless query is a great touch, much cleaner.

Got it on the CHANGELOG policy, that makes total sense. I'll keep entries for features/deprecations only going forward.

Your encouragement truly means a lot. I'm loving every minute of contributing to Rails and I'll definitely keep going! 🔥

PR #57599, opened 2026-06-06
Strip whitespace from string URLs passed to redirect_to

dbaratta:
### Motivation / Background

Fixes #57035.

`redirect_to " https://www.example.com"` (note the leading space) raises `ActionController::Redirecting::PathRelativeRedirectError`, or with the default `:log` setting produces a broken `Location` such as `http://yourhost https://www.example.com`. The scheme check in `_compute_redirect_to_location` is anchored with `\A`, so a single leading space stops `" https://..."` from being recognized as an absolute URL and it gets classified as a path relative redirect instead.

Browsers ignore leading and trailing whitespace when they navigate (per the WHATWG URL spec), so a stray space pasted in front of an otherwise valid URL shouldn't break the redirect.

### Detail

`_compute_redirect_to_location` now strips leading and trailing whitespace from string targets before classifying them:

```ruby
options = options.strip if options.is_a?(String)
```

The strip runs before the existing redirect protections, so behavior for genuinely unsafe input is unchanged:

* `"  example.com  "` still strips to `"example.com"` and is flagged as a path relative redirect (the warning/raise message shows the trimmed value).
* A whitespace padded cross-host URL is now recognized as cross-host and goes through the same open redirect protection it would without the whitespace, instead of becoming a broken same-host `Location`.

One existing test action in `log_subscriber_test.rb` relied on a leading space making `" s:/invalid-string0uri"` unparseable after host prefixing. Since that input is no longer host prefixed, I changed it to `"s:/invalid string0uri"`, which is still unparseable by `URI.parse` regardless of trimming, so it keeps exercising the `[FILTERED]` rescue path it was written for.

### Additional information

Added coverage in `redirect_test.rb` for a whitespace padded absolute URL redirecting to the trimmed URL, that same URL not raising under `action_on_path_relative_redirect = :raise`, and a whitespace padded path relative URL still being host prefixed and still raising. RuboCop and the Action Pack controller redirect and log subscriber tests pass.

### Checklist

### DIFF
--- a/actionpack/CHANGELOG.md
+++ b/actionpack/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Strip leading and trailing whitespace from string URLs passed to `redirect_to`.
+
+    Browsers ignore surrounding whitespace when navigating, so `redirect_to " https://example.com"`
+    now redirects to `https://example.com` instead of being misclassified as a path relative
+    redirect and producing a broken `Location`. Trimming happens before the redirect protections
+    run, so genuinely path relative and cross-host URLs are still flagged as before.
+
+    *Dominic Baratta*
+
 *   Rate limiting calls `cache_key` on `by:` if the object responds to it.
 
     ```ruby
--- a/actionpack/lib/action_controller/metal/redirecting.rb
+++ b/actionpack/lib/action_controller/metal/redirecting.rb
@@ -204,6 +204,12 @@ module ActionController
     end
 
     def _compute_redirect_to_location(request, options) # :nodoc:
+      # Browsers strip leading and trailing whitespace from URLs before
+      # navigating, so do the same here. Otherwise an otherwise valid URL like
+      # " https://example.com" would be misclassified as a path relative
+      # redirect because of the leading space.
+      options = options.strip if options.is_a?(String)
+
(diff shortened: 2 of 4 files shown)


dbaratta:
@byroot @p8, looping you in since you both looked at #57038.

This is the plain `String#strip` version from that review, and it also covers the leading-whitespace case in the original report (the reporter's example was a space in front of `https://...`, which the trailing-only approach wouldn't have caught).

If the call was that `redirect_to` shouldn't touch surrounding whitespace at all, I'm glad to close. But since the current behavior emits a broken `Location` rather than rejecting cleanly, and browsers strip this anyway per WHATWG, it felt worth a second look now that the implementation is simpler. No worries either way.

p8:
Why would you want to redirect to an invalid URL?
Is this coming from user input? Than you should clean it up before passing it to redirect_to. 
Stripping in redirect_to would hide an issue with unclean URLs getting passed.

p8:
@dbaratta I’ve closed that issue. 
The URLs from user input should be cleaned with something like normalizes. 
https://github.com/rails/rails/issues/57035#issuecomment-4637751618

PR #57602, opened 2026-06-06
Make Active Record Railtie Ractor safe

gmcgibbon:
### Motivation / Background

This Pull Request has been created to make Active Record's railtie Ractor safe.

### Detail

This Pull Request changes railtie message verifier procs to be shareable.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/railtie.rb
+++ b/activerecord/lib/active_record/railtie.rb
@@ -351,12 +351,12 @@ To keep using the current cache store, you can turn off cache versioning entirel
       ActiveRecord.message_verifiers = app.message_verifiers
 
       use_legacy_signed_id_verifier = app.config.active_record.use_legacy_signed_id_verifier
-      legacy_options = { digest: "SHA256", serializer: JSON, url_safe: true }
+      legacy_options = { digest: "SHA256", serializer: JSON, url_safe: true }.freeze
 
       if use_legacy_signed_id_verifier == :generate_and_verify
-        app.message_verifiers.prepend { |salt| legacy_options if salt == "active_record/signed_id" }
+        app.message_verifiers.prepend(&ActiveSupport::Ractors.shareable_proc { |salt| legacy_options if salt == "active_record/signed_id" })
       elsif use_legacy_signed_id_verifier == :verify
-        app.message_verifiers.rotate { |salt| legacy_options if salt == "active_record/signed_id" }
+        app.message_verifiers.rotate(&ActiveSupport::Ractors.shareable_proc { |salt| legacy_options if salt == "active_record/signed_id" })
       elsif use_legacy_signed_id_verifier
         raise ArgumentError, "Unrecognized value for config.active_record.use_legacy_signed_id_verifier: #{use_legacy_signed_id_verifier.inspect}"
       end


gmcgibbon:
Seems like `ActiveRecord.database_cli` is being mutated in tests. Freezing config values doesn't necessarily solve the problem in the right way as applications may be do the same. 🤔

activerecord/lib/active_record/railtie.rb
```diff
@@ -399,6 +399,22 @@ class Railtie < Rails::Railtie # :nodoc:
...
+          :protected_environments,
+          :schema_cache_ignored_tables,
+          :database_cli,
+          :db_warnings_ignore,
+          :queues,
+          :query_transformers,
+          :yaml_column_permitted_classes,
+        ].each do |configuration|
+          ActiveRecord.public_send(configuration).freeze
+        end
```
byroot:
```suggestion
        ActiveRecord.protected_environments.freeze
        ActiveRecord.schema_cache_ignored_tables.freeze
        ActiveRecord.database_cli.freeze
        ActiveRecord.db_warnings_ignore.freeze
        ActiveRecord.queues.freeze
        ActiveRecord.query_transformers.freeze
        ActiveRecord.yaml_column_permitted_classes.freeze
```

But even that isn't quite correct, several of these are arrays or hashes, so you need recursive freezing.

It might also be better just to modify the setters to freeze what they receive.

gmcgibbon: We can't easily do it on setters because you can mutate objects in place. I've opted to use initializers instead so that configs are frozen during the boot process. Either that, or we need to look at custom data structures that freeze on mutation (`[]=`, `<<`, etc.) I believe?

activerecord/lib/active_record/railtie.rb
```diff
@@ -454,5 +454,17 @@ class Railtie < Rails::Railtie # :nodoc:
...
+
+    initializer "active_record.freeze_configurations" do
+      config.after_initialize do
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.protected_environments)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.schema_cache_ignored_tables)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.database_cli)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.db_warnings_ignore)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.queues)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.query_transformers)
+        ActiveSupport::Ractors.make_shareable(ActiveRecord.yaml_column_permitted_classes)
```
byroot: cc @jhawthorn @tenderlove when I looked at this yesterday, I though the logical way to do this would have been:

```ruby
ActiveRecord.query_transformers = Ractor.make_shareable(ActiveRecord.query_transformers, copy: true)
```

The problem however is procs:

```ruby
>> Ractor.make_shareable([->{}])
(irb):7:in 'Ractor.make_shareable': Proc's self is not shareable: #<Proc:0x0000000121673c30 (irb):7 >> Ractor.make_shareable([->{}], copy: true)
(irb):8:in 'Ractor.make_shareable': allocator undefined for Proc (TypeError)
```

Which makes me think, `Ractor.make_sharable(stuff, copy: true, shareable_proc: true)` should be a thing? Thoughts?

gmcgibbon: Agree. I am also starting to think that initialization isn't the right place for this freezing code as it prevents the test suite and developers from mutating these values later. We should call it in a special hook when we actually do the ractor preparation. I'll mark this as blocked for now.

jhawthorn: @byroot I'd support the proposal but given how much discussion was required to add `shareable_proc` (and a big part of it being not automatic in `make_sharable`) I'm not sure how that would go

byroot: I feel you. But it's probably worth at least asking Koichi about it.

gmcgibbon:
After thinking about the freezing of configuration a little more, I'm going to pitch a configuration variable that freezes configs after initialization of each Railtie (eg. the Active Record Railtie will freeze Active Record config values). I'll open a patch to introduce that config so we can use it here.

gmcgibbon:
https://github.com/rails/rails/pull/57852. I'm going to revert the config freezing here and just merge the initializer change. We can move the discussion to the other PR.

PR #57606, opened 2026-06-06
Resolve attribute aliases for the locking column in `update_all`

55728:
### Summary

`update_all` corrupts the optimistic-locking column when it is set through an `alias_attribute`: the value is silently dropped on SQLite/MySQL, and on PostgreSQL the statement raises.

```ruby
class Widget < ActiveRecord::Base
  alias_attribute :v, :lock_version
end

Widget.where(id: id).update_all(v: 10)
# SQLite/MySQL: lock_version ends up 1, not 10  (value silently lost)
# PostgreSQL:   PG::SyntaxError: multiple assignments to same column "lock_version"
```

The canonical name works correctly:

```ruby
Widget.where(id: id).update_all(lock_version: 10) # => lock_version is 10
```

### Cause

`update_all` auto-increments the locking column unless the caller already sets it, but the guard recognizes the column only by its canonical name:

```ruby
if model.locking_enabled? &&
    !updates.key?(model.locking_column) &&
    !updates.key?(model.locking_column.to_sym)
  attr = table[model.locking_column]
  updates[attr.name] = _increment_attribute(attr)
end
values = _substitute_values(updates)
```

An aliased key (`v`) is not equal to `"lock_version"` / `:lock_version`, so the guard fires and appends the auto-increment. `_substitute_values` then resolves **both** `v` and the appended key to the same physical column, producing two assignments to `lock_version`:

```sql
UPDATE "widgets" SET "lock_version" = 10, "lock_version" = COALESCE("widgets"."lock_version", 0) + 1 ...
```

SQLite/MySQL take the last assignment (the auto-increment), dropping the user's value; PostgreSQL rejects the duplicate column assignment.

### Fix

Resolve each update key through `model.attribute_alias` before comparing it to the locking column, so an aliased lock column is recognized just like the canonical name — `insert_all` already resolves aliases this way:

```ruby
if model.locking_enabled? &&
    updates.keys.none? { |key| (model.attribute_alias(key) || key.to_s) == model.locking_column }
  ...
end
```

The auto-increment still fires when no update key maps to the locking column, and the canonical-name path is unchanged.

### Test

Added `test_update_all_with_lock_column_set_through_an_alias` to `OptimisticLockingTest` (a model on the `people` table with `alias_attribute :aliased_lock_version, :lock_version`). It asserts `update_all(aliased_lock_version: 10)` sets the column to `10` and affects one row.

Verified fail → pass on **sqlite3** (`10` → `1` lost write on `main`), **postgresql** (crash on `main`), and **mysql2**. `locking_test` and `relation/update_all_test` are green on all three adapters.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Fix `update_all` corrupting the optimistic locking column when it is set
+    through an `alias_attribute`.
+
+    *Kenta Ishizaki*
+
 *   Fix `serialize` with `coder: ActiveRecord::Coders::JSON` silently double-encoding
     a native `json`/`jsonb` column.
 
--- a/activerecord/lib/active_record/relation.rb
+++ b/activerecord/lib/active_record/relation.rb
@@ -619,8 +619,7 @@ module ActiveRecord
 
       if updates.is_a?(Hash)
         if model.locking_enabled? &&
-            !updates.key?(model.locking_column) &&
-            !updates.key?(model.locking_column.to_sym)
+            updates.keys.none? { |key| (model.attribute_alias(key) || key.to_s) == model.locking_column }
           attr = table[model.locking_column]
           updates[attr.name] = _increment_attribute(attr)
         end
--- a/activerecord/test/cases/locking_test.rb
+++ b/activerecord/test/cases/locking_test.rb
@@ -28,6 +28,11 @@ class ReadonlyNameShip < Ship
   attr_readonly :name
 end
 
+class PersonWithAliasedLockVersion < ActiveRecord::Base
+  self.table_name = :people
+  alias_attribute :aliased_lock_version, :lock_version
+end
+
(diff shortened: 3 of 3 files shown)


fatkodima:
Can you provide a real world use case when someone will want to create an alias for the locking column? I am thinking if this is a problem worth fixing.

55728:
Good question 👀✨ — a renamed/legacy lock column aliased to a friendlier name is the most natural one, but I think the case rests more on **what happens** than on how common the alias is: the user's explicit value is silently dropped on SQLite/MySQL (no error), and PostgreSQL raises. `update_all` already resolves aliases for every other column — the locking guard is the only spot that doesn't, and `insert_all`/`upsert_all` already resolve attribute aliases for this reason, so this just brings `update_all` in line.

That said, happy to close it if you feel it's too niche — just wanted to flag the silent-drop angle. 🙏

fatkodima:
I will let other reviewers to also do their reviews.

PR #54172, opened 2025-01-09
JSON serialized attributes can return symbolized keys

matthaigh27:
- Allow `json: { symbolize_names: true }` to be passed to `serialize`
- Opted for `symbolize_names` for parity with `ruby/json`

### Motivation / Background

This Pull Request has been created because I make heavy use of symbols and on JSON serialized attributes, I found myself always either using `.deep_symbolize_keys` or, when it wasn't causing problems, creating a `SymbolizedJSON` class just to pass `symbolize_names` into `JSON.parse`.

### Detail

This Pull Request changes the options available to the `serialize` attribute in active record. Similar to the `yaml:` option, you can now specify `json:` with the only option right now being `symbolize_names`

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #57609, opened 2026-06-06
Fix `serialize coder: ActiveRecord::Coders::JSON` double-encoding a JSON column

55728:
### Summary

Declaring `serialize` with `coder: JSON` on a column whose type is already a native `json`/`jsonb` raises `ColumnNotSerializableError`, but the equivalent `coder: ActiveRecord::Coders::JSON` slips past the same guard and silently double-encodes the value:

```ruby
  # `data` is a native json/jsonb column
  serialize :data, coder: JSON                          # => raises ColumnNotSerializableError (good)
end

  serialize :data, coder: ActiveRecord::Coders::JSON    # => silently accepted (bug)
end

Post.create!(data: { "a" => 1 })
# raw DB value: "\"{\\\"a\\\":1}\""   (a JSON string of a JSON string)
# a model reading the column without the serializer sees a String, not a Hash
```

The two spellings are meant to be equivalent, so one protecting the user while the other corrupts data is surprising.

### Cause

`build_column_serializer` already treats both spellings as the same JSON coder:

```ruby
if coder == ::JSON || coder == Coders::JSON
  coder = Coders::JSON.new
end
```

But the compatibility guard only checks the bare `::JSON` constant:

```ruby
def type_incompatible_with_serialize?(cast_type, coder, type)
  cast_type.is_a?(ActiveRecord::Type::Json) && coder == ::JSON ||
    cast_type.respond_to?(:type_cast_array, true) && type == ::Array
end
```

So for a `Type::Json` column, `coder: JSON` is correctly rejected, while `coder: ActiveRecord::Coders::JSON` is not — it wraps the JSON type in a `Type::Serialized`, and the value is JSON-encoded twice (once by the serializer, once by the JSON type underneath).

### Fix

Broaden the guard to recognize both spellings, mirroring `build_column_serializer`:

```ruby
def type_incompatible_with_serialize?(cast_type, coder, type)
  cast_type.is_a?(ActiveRecord::Type::Json) && (coder == ::JSON || coder == Coders::JSON) ||
    cast_type.respond_to?(:type_cast_array, true) && type == ::Array
end
```

`ColumnNotSerializableError` is now raised consistently for both forms.

### Test

Added `test_not_compatible_with_serialize_active_record_coders_json` to the shared `JSONSharedTestCases`, alongside the existing `coder: JSON` test, asserting that declaring `coder: ActiveRecord::Coders::JSON` on a native JSON column raises `ColumnNotSerializableError`.

Verified fail → pass on **sqlite3** ("nothing was raised" → raises), **postgresql**, and **mysql2**. The adapter JSON suites and `json_attribute_test` / `serialized_attribute_test` are green on all three adapters.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Fix `serialize` with `coder: ActiveRecord::Coders::JSON` silently double-encoding
+    a native `json`/`jsonb` column.
+
+    *Kenta Ishizaki*
+
 *   Fix `update_all` / `delete_all` ignoring `group` and `having`, updating or
     deleting every row in the table instead of only the rows that satisfy the
     `HAVING` clause.
--- a/activerecord/lib/active_record/attribute_methods/serialization.rb
+++ b/activerecord/lib/active_record/attribute_methods/serialization.rb
@@ -236,7 +236,7 @@ module ActiveRecord
           end
 
           def type_incompatible_with_serialize?(cast_type, coder, type)
-            cast_type.is_a?(ActiveRecord::Type::Json) && coder == ::JSON ||
+            cast_type.is_a?(ActiveRecord::Type::Json) && (coder == ::JSON || coder == Coders::JSON) ||
               cast_type.respond_to?(:type_cast_array, true) && type == ::Array
           end
       end
--- a/activerecord/test/cases/json_shared_test_cases.rb
+++ b/activerecord/test/cases/json_shared_test_cases.rb
@@ -227,6 +227,15 @@ module JSONSharedTestCases
     end
   end
 
+  def test_not_compatible_with_serialize_active_record_coders_json
+    new_klass = Class.new(klass) do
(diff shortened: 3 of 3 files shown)


skipkayhil:
`Coders::JSON` is undocumented (internal to the framework), so no one should be passing it here

55728:
Thank you for the context. Good to know for future contributions.

skipkayhil:
Hmm, I looked into the history more and I think the intention is actually for `ActiveRecord::Coders::JSON` to be public 🤔#54172 cc @byroot 

In that case, I think we _do_ need a fix here, but I think we can make it better: maybe we can pass `column_serializer` in instead of `coder` and then we only have to check `Coders::JSON`?

byroot:
> I think the intention is actually for `ActiveRecord::Coders::JSON` to be public

Eh, not really, at least not right away, but yes we should make it work. Also not that `Coders::JSON` can be instantiated, so you want to also check `Coders::JSON === coder`

55728:
Thank you @skipkayhil for digging into the history and landing the fix! 😊

PR #57619, opened 2026-06-07
[ActiveModel] Add a `self` feature to `normalizes`

yaroslav:
### Motivation / Background

This Pull Request has been created because:

1. Issue: sometimes `normalizes` runs several times on every `valid?` call.
2. Feature: `normalizes` would be much more useful if there is access to other record fields.

#### Commit 1. Re-run `normalizes` only when the value changed in place

`normalize_changed_in_place_attributes` ran the normalizer again on every `valid?` call for a record that is not yet saved.
It used `attribute_changed_in_place?` to decide. That check compares the current value with the saved one, but an unsaved record has no saved value (it is `nil`), so any attribute that was just read looked changed. The normalizer then ran again on every validation.

For normalizers that return the same result when run twice, this only wasted work. For the others, the result kept growing: for example, a `succ` normalizer turned "b" into "c", then "d", then "e" on each validation.

Now the normalizer runs again only when the value was really changed in place.
We check this by comparing the current value with a fresh normalization of the original value. If the value was only read (not changed), we reset it to the original value, so later validations skip it and a later read normalizes the original value again instead of an already-normalized one.

##### Benchmarks

Create, update, and save run the normalizer the same number of times as before. Validating an already-normalized value again is now faster and makes no extra allocations.
Benchmark, three string attributes normalized with strip/downcase, read once then validated many times on an unsaved record (Ruby 4.0, Apple M-series):

    YJIT off:  480k i/s  -> 533k i/s   (~1.1x)
    YJIT on:   1.10M i/s -> 1.13M i/s  (about the same)
    allocations per valid?:  6 -> 6

#### Commit 2. Add a `self:` option to `normalizes`

With `self: true`, the `with:` normalizer runs with the record as `self`, so it can read other attributes. This is useful when a value depends on another attribute of the same record.

      normalizes :nickname, self: true,
        with: -> nickname { nickname.presence || name&.split&.first }
    end

    user = User.new(name: "David Heinemeier Hansson", nickname: "")
    user.nickname # => "David"

    user = User.new(name: "David Heinemeier Hansson", nickname: "DHH")
    user.nickname # => "DHH"

Such a normalizer needs a record, so it only runs on assignment and during validation. In contexts without a record (queries and `normalize_value_for`), the value is left unchanged.

### Additional information

The first patch needs to be applied so that the second works (otherwise, it triggers a mentioned bug). Therefore, a PR with two commits.

Assisted-by: Opus 4.8

### Checklist


yaroslav:
Extracted the bugfix into https://github.com/rails/rails/pull/57639.

yaroslav:
@rafaelfranca Hi! I think I've fixed this one. The bugfix is on main and the feature addition is in this PR. I'd like to request a review, if possible

PR #51159, opened 2024-02-22
Fix beginless and endless range comparisons

AlexMooney:
### Motivation / Background

This Pull Request has been created because some endless and beginless range comparisons raise `RangeError` due to `ActiveSupport::CompareWithRange`

### Detail

This Pull Request changes the `CompareWithRange#===` and `CompareWithRange#include?` methods to support begin and endless ranges on the left and right hand side of the comparison.

### Additional information

POR:
```ruby
> (1..) === (2..)
=> false
> (..1) === (..2)
=> false
```
Rails 7.1.3:
```ruby
> (1..) === (2..)
RangeError: cannot get the last element of endless range
from …/activesupport-7.1.3/lib/active_support/core_ext/range/compare_range.rb:23:in `last'
> (..1) === (..2)
RangeError: cannot get the first element of beginless range
from …/activesupport-7.1.3/lib/active_support/core_ext/range/compare_range.rb:24:in `first'
```
This PR:
```ruby
> (1..) === (2..)
=> true
> (2..) === (1..)
=> false
> (..1) === (..2)
=> false
> (..2) === (..1)
=> true
```

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Changes that are unrelated should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #57622, opened 2026-06-08
Fix `Range#include?` / `Range#===` raising on exclusive non-integer sub-ranges

55728:
### Summary

ActiveSupport extends `Range#include?` and `Range#===` to support comparing a range against another range. When the argument is a range, the override computed the other range's maximum element like this (`activesupport/lib/active_support/core_ext/range/compare_range.rb`):

```ruby
value_max = !exclude_end? && value.exclude_end? ? value.max : value.last
```

`Range#max` enumerates the range, so for an **exclusive-end range whose endpoint is not an `Integer`** it raises instead of returning a boolean:

```ruby
(1.0..10.0).include?(2.0...5.0)
# => TypeError: cannot exclude non Integer end value

t = Time.now
(t..(t + 1.hour)).include?(t...(t + 10.minutes))
# => TypeError: can't iterate from Time
```

This affects exclusive `Float`, `Time`, `DateTime`, `TimeWithZone`, and `BigDecimal` sub-ranges — exactly the kinds of ranges used for time windows and tolerance bands — turning an ordinary membership check into a hard crash. The `value.max` call exists only so that an integer case such as `(1..10).include?(1...11)` resolves to `true`; for every non-discrete element type it simply blows up.

### Why this fix

The hand-rolled range-vs-range comparison predates `Range#cover?` accepting a range argument (added in Ruby 2.6). Since Rails requires a newer Ruby, the range branch can delegate straight to `Range#cover?`, which is the canonical "is this range fully contained" check and returns the correct boolean for every combination of inclusive/exclusive ends and element type — including the backwards and empty-range edge cases the override handled specially:

```ruby
def ===(value)
  if value.is_a?(::Range)
    cover?(value)
  else
    super
  end
end

def include?(value)
  if value.is_a?(::Range)
    cover?(value)
  else
    super
  end
end
```

The scalar (`else super`) branch is left untouched, so native `Range#include?` / `Range#===` behavior for non-range arguments (e.g. `('a'..'f').include?('c')`) is unchanged.

`Range#cover?` was verified to produce the **same result as every existing assertion** in `range_ext_test.rb` — identical inclusive/exclusive ranges, exclusive-end integer ranges (`1..10` vs `1...11`), backwards ranges (`5..3` → false), empty exclusive ranges (`3...3` → false), endless/beginless ranges, and the overlap cases — while also returning the correct boolean for the previously-crashing Float/Time/BigDecimal cases.

### Behavior after the fix

```ruby
(1.0..10.0).include?(2.0...5.0)    # => true  (was: TypeError)
(1.0..10.0).include?(2.0...10.5)   # => false (was: TypeError)

t = Time.utc(2005, 12, 10, 15, 30)
(t..(t + 2.hours)).include?((t + 30.minutes)...(t + 90.minutes))  # => true (was: TypeError)
```

### Testing

Added to `activesupport/test/core_ext/range_ext_test.rb`:

- `test_should_include_exclusive_end_float_range`
- `test_should_not_include_exclusive_end_float_range_past_end`
- `test_should_include_exclusive_end_time_range`

Without the fix these error with `TypeError: cannot exclude non Integer end value` / `can't iterate from Time`. With the fix:

- `range_ext_test.rb`: 51 runs, 164 assertions, 0 failures, 0 errors
- `time_ext_test.rb`, `date_ext_test.rb`, `date_time_ext_test.rb`, `time_zone_test.rb`, `time_with_zone_test.rb`: all green (no regressions)

No CHANGELOG entry — small bug fix, not a feature or deprecation.

### DIFF
--- a/activesupport/lib/active_support/core_ext/range/compare_range.rb
+++ b/activesupport/lib/active_support/core_ext/range/compare_range.rb
@@ -15,13 +15,7 @@ module ActiveSupport
     # The given range must be fully bounded, with both start and end.
     def ===(value)
       if value.is_a?(::Range)
-        is_backwards_op = value.exclude_end? ? :>= : :>
-        return false if value.begin && value.end && value.begin.public_send(is_backwards_op, value.end)
-        # 1...10 includes 1..9 but it does not include 1..10.
-        # 1..10 includes 1...11 but it does not include 1...12.
-        operator = exclude_end? && !value.exclude_end? ? :< : :<=
-        value_max = !exclude_end? && value.exclude_end? ? value.max : value.last
-        super(value.first) && (self.end.nil? || value_max.public_send(operator, last))
+        cover?(value)
       else
         super
       end
@@ -40,13 +34,7 @@ module ActiveSupport
     # The given range must be fully bounded, with both start and end.
     def include?(value)
       if value.is_a?(::Range)
-        is_backwards_op = value.exclude_end? ? :>= : :>
-        return false if value.begin && value.end && value.begin.public_send(is_backwards_op, value.end)
-        # 1...10 includes 1..9 but it does not include 1..10.
-        # 1..10 includes 1...11 but it does not include 1...12.
-        operator = exclude_end? && !value.exclude_end? ? :< : :<=
-        value_max = !exclude_end? && value.exclude_end? ? value.max : value.last
-        super(value.first) && (self.end.nil? || value_max.public_send(operator, last))
+        cover?(value)
       else
         super
       end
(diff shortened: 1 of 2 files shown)


skipkayhil:
Does this also fix some of the cases described in #51159? If so, maybe we should adopt some of those tests and add that PR author as Co-author

55728:
Yes — verified that delegating to `cover?` resolves all the endless/beginless cases from #51159 (they still crash on main today). Adopted the relevant test assertions and credited @AlexMooney as co-author. Force-pushed.

PR #57626, opened 2026-06-08
Introduce a new mechanism for applications to prepare for ractor safety

Edouard-chin:
### Motivation / Background

We'd like to make Rails applications be Ractor safe. One of the issue is about procs passed to some Rails API. We'd like those application procs to be made shareable, however, specific (and admittedly edge case) application code may prevent us from making the proc shareable.

For example, this snippet will raise a Ractor::IsolationError.

```ruby
    to_resolve = [:basket, anchor: "items"]

    resolve("Cart") { to_resolve }
  end
```

This is one example in the rails codebase where we'd hit this problem but there are few others.

### Detail

This PR introduces a way for applications to know whether all procs passed to Rails API are ractors safe. It's opt-in by default and include 3 modes:

* `:raise` -> Rails attempt to make the proc shareable. It it's not, a loud `Ractor::IsolationError` will be raised.
* `:warn` -> We attempt to make the proc shareable. If it's not, we rescue `Ractor::IsolationError` and output a warning.
* `nil` -> We *do not* attempt to make the proc shareable.

This PR introduces this new feature and one codepath (the RouteSet) uses it. We'd be using this mechanism in other part of Rails in future PRs.

### Checklist

cc/ @matthewd @etiennebarrie @gmcgibbon @andrewn617

### DIFF
--- a/actionpack/test/dispatch/routing/route_set_test.rb
+++ b/actionpack/test/dispatch/routing/route_set_test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "abstract_unit"
+require "active_support/core_ext/object/with"
 
 module ActionDispatch
   module Routing
@@ -185,6 +186,44 @@ module ActionDispatch
         route = @set.from_requirements(controller: "baz", action: "index")
 
         assert_nil route
+      end
+
+      if RUBY_VERSION >= "4.0"
+        test "#resolve raises an error when a proc is not shareable and unshareable_proc_action is :raise" do
+          assert_raise(Ractor::IsolationError) do
+            ActiveSupport::Ractors.with(unshareable_proc_action: :raise) do
+              draw do
+                to_resolve = [:basket, anchor: "items"]
+
+                resolve("Cart") { to_resolve }
+              end
+            end
+          end
+        end
+
+        test "#resolve trigger a deprecation when a proc is not shareable and unshareable_proc_action is :warn" do
+          assert_deprecated(/Rails attempted to make a Proc .* Ractor shareable/, ActiveSupport.deprecator) do
+            ActiveSupport::Ractors.with(unshareable_proc_action: :warn) do
+              draw do
+                to_resolve = [:basket, anchor: "items"]
+
+                resolve("Cart") { to_resolve }
+              end
+            end
+          end
+        end
+
+        test "#resolve does not attempt to make a proc shareable when unshareable_proc_action is nil" do
+          assert_nothing_raised do
+            ActiveSupport::Ractors.with(unshareable_proc_action: nil) do
+              draw do
+                to_resolve = [:basket, anchor: "items"]
+
+                resolve("Cart") { to_resolve }
+              end
+            end
+          end
+        end
       end
 
       private
--- a/activesupport/lib/active_support/ractors.rb
+++ b/activesupport/lib/active_support/ractors.rb
@@ -5,6 +5,36 @@ module ActiveSupport
   # unconditionally regardless of the Ruby version.
   module Ractors # :nodoc:
     class << self
+      attr_accessor :unshareable_proc_action
+
+      # Attempt to make a proc shareable. If successful, a shareable proc is returned.
+      # If a Ractor::IsolationError is raised, the outcome will depend on how
+      # the user's application configuration:
+      #
+      # :raise - The error is raised
+      # :warn  - A deprecation warning is triggered and the original unshareable proc is returned.
+      def try_shareable_proc(proc = nil, &block)
+        proc ||= block
+        return proc unless unshareable_proc_action
+
+        shareable_proc(&proc)
+      rescue Ractor::IsolationError
+        case unshareable_proc_action
+        when :raise
+          raise
+        when :warn
+          ActiveSupport.deprecator.warn(<<~MSG)
+            Rails attempted to make a Proc from your application Ractor shareable but a Ractor
+            Isolation error was raised. The proc being returned is not Ractor safe and a runtime
+            error may occur anytime during the request lifecycle.
+
+            #{proc.inspect}
+          MSG
+
+          proc
+        end
+      end
+
       if defined?(Ractor) && RUBY_VERSION >= "4.0"
         # Makes +obj+ Ractor-shareable by delegating to +Ractor.make_shareable+.
         #
--- a/actionpack/lib/action_dispatch/routing/route_set.rb
+++ b/actionpack/lib/action_dispatch/routing/route_set.rb
@@ -685,7 +685,7 @@ module ActionDispatch
         def initialize(name, defaults, &block)
           @name = name
           @defaults = defaults
-          @block = block
+          @block = ActiveSupport::Ractors.try_shareable_proc(block)
         end
 
         def call(t, args, only_path = false)
--- a/activesupport/test/ractors_test.rb
+++ b/activesupport/test/ractors_test.rb
@@ -44,5 +44,59 @@ class KernelRactorShareabilityTest < ActiveSupport::TestCase
       assert ActiveSupport::Ractors.shareable?(lambda)
       assert_equal "2", lambda.call
     end
+
+    def test_try_shareable_proc_when_action_is_nil
+      old = ActiveSupport::Ractors.unshareable_proc_action
+      ActiveSupport::Ractors.unshareable_proc_action = nil
+
+      proc = -> { }
+      attempted = ActiveSupport::Ractors.try_shareable_proc(proc)
+
+      assert_same(proc, attempted)
+    ensure
+      ActiveSupport::Ractors.unshareable_proc_action = old
+    end
+
+    def test_try_shareable_proc_creates_a_shareable_proc
+      old = ActiveSupport::Ractors.unshareable_proc_action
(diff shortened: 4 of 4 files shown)


actionpack/test/dispatch/routing/route_set_test.rb
```diff
@@ -187,6 +187,53 @@ def call(env)
...
+
+          assert_raise(Ractor::IsolationError) do
+            draw do
+              to_resolve = [:basket, anchor: "items"]
+
+              resolve("Cart") { to_resolve }
+            end
+          end
+        ensure
+          ActiveSupport::Ractors.unshareable_proc_action = old
```
gmcgibbon: You can use `with` to cut down on boilerplate here: https://github.com/rails/rails/blob/ddf8c2f393e43178a0f523413c3ad681d264ee0d/activesupport/lib/active_support/core_ext/object/with.rb#L26

activesupport/lib/active_support/ractors.rb
```diff
@@ -5,6 +5,33 @@ module ActiveSupport
...
     class << self
+      attr_accessor :unshareable_proc_action
+
+      # Attempt to make a proc shareable. If successful, a shareable proc is returned.
+      # If a Ractor::IsolationError is raised, the outcome will depend based on how
+      # the user's application configuration:
+      #
+      # :raise - The error is raised
+      # :warn  - A deprecation warning is triggered and the original proc (unshareable) is returned.
+      def shareable_proc_attempt(proc)
```
gmcgibbon:
```suggestion
      def try_shareable_proc(...)
```
We need to accept a block here too.

activesupport/lib/active_support/ractors.rb
```diff
@@ -5,6 +5,33 @@ module ActiveSupport
   # unconditionally regardless of the Ruby version.
   module Ractors # :nodoc:
     class << self
+      attr_accessor :unshareable_proc_action
+
+      # Attempt to make a proc shareable. If successful, a shareable proc is returned.
+      # If a Ractor::IsolationError is raised, the outcome will depend based on how
```
gmcgibbon:
```suggestion
      # If a Ractor::IsolationError is raised, the outcome will depend on how
```

activesupport/lib/active_support/ractors.rb
```diff
@@ -5,6 +5,33 @@ module ActiveSupport
...
+      # the user's application configuration:
+      #
+      # :raise - The error is raised
+      # :warn  - A deprecation warning is triggered and the original proc (unshareable) is returned.
```
gmcgibbon:
```suggestion
      # :warn  - A deprecation warning is triggered and the original unshareable proc is returned.
```

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+*   Added a new `config.active_support.unshareable_proc_action`
+
+    Running Rails on Ractors is coming fast but not there yet. In order
+    for applications to be prepared to be compatible with Ractors, we are introducing
+    this configuration to control the behaviour when an application
+    code passes a non shareable proc to a Rails API.
+
+    There are 3 modes:
+
+    * :raise -> Rails attempt to make the proc shareable. It it can not, a loud `Ractor::IsolationError` will be raised.
```
gmcgibbon:
```suggestion
    * :raise -> Rails attempts to make the proc shareable. If it can't, a `Ractor::IsolationError` will be raised.
```

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+    * :warn ->  Rails attempt to make the proc shareable. If it's not, we rescue `Ractor::IsolationError` and output a warning.
```
gmcgibbon:
```suggestion
    * :warn ->  Rails attempts to make the proc shareable. If it can't, we rescue `Ractor::IsolationError` and output a warning.
```

activesupport/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+    * `nil` ->  Rails *do not* attempt to make the proc shareable.
```
gmcgibbon:
```suggestion
    * `nil` ->  Rails *does not* attempt to make the proc shareable.
```

guides/source/configuring.md
```diff
@@ -3163,6 +3163,14 @@ The default value depends on the `config.load_defaults` target version:
 | (original)            | `false`              |
 | 7.0                   | `true`               |
 
+#### `config.active_support.unshareable_proc_action`
+
+Controls whether procs passed to class macros (e.g. routes) are made Ractor-shareable before being instance_exec'd, and what happens when one can't be shared:
+
+* :raise — try to make it shareable; let Ractor::IsolationError propagate.
```
gmcgibbon:
```suggestion
* `:raise` — try to make it shareable; let `Ractor::IsolationError` propagate.
```

guides/source/configuring.md
```diff
@@ -3163,6 +3163,14 @@ The default value depends on the `config.load_defaults` target version:
...
+* :warn — try to make it shareable; rescue Ractor::IsolationError and emit a deprecation warning.
```
gmcgibbon:
```suggestion
* `:warn` — try to make it shareable; rescue `Ractor::IsolationError` and emit a deprecation warning.
```

guides/source/configuring.md
```diff
@@ -3163,6 +3163,14 @@ The default value depends on the `config.load_defaults` target version:
...
+* nil (default) — don't try to make the proc shareable.
```
gmcgibbon:
```suggestion
* `nil` (default) — don't try to make the proc shareable.
```

railties/test/application/configuration_test.rb
```diff
@@ -4717,6 +4717,13 @@ def new(app); self; end
       assert_equal :thread, ActiveSupport::IsolatedExecutionState.isolation_level
     end
 
+    test "#unshareable_proc_action can be set" do
```
gmcgibbon: What is this testing? Routes aren't loaded eagerly in development.

activesupport/lib/active_support/ractors.rb
```diff
@@ -5,6 +5,33 @@ module ActiveSupport
...
+        return proc unless unshareable_proc_action
+
+        shareable_proc(&proc)
+      rescue Ractor::IsolationError
+        case unshareable_proc_action
+        when :raise
+          raise
+        when :warn
+          ActiveSupport.deprecator.warn(<<~MSG)
+            Rails attempted to make a Proc from your application Ractor shareable but a Ractor
```
byroot, line 27: You might as well include `proc.inspect` as it will include the `source_location` making it much easier to locate.

byroot (commented):
IMO this is way way too soon, and putting the cart before the horse. At this stage there's basically nothing you can do from a Ractor.

Why should we commit to public APIs, documentations, etc?

Let's discuss this when at least something works. I want us to be able to reap it out if the experiment doesn't pan out.

activesupport/lib/active_support/railtie.rb
```diff
@@ -23,6 +23,14 @@ class Railtie < Rails::Railtie # :nodoc:
       end
     end
 
+    initializer "active_support.unshareable_proc_action" do
+      config.after_initialize do |app|
+        if action = app.config.active_support.unshareable_proc_action
```
gmcgibbon:
```suggestion
        if action = app.config.active_support.ractor_unshareable_proc_action
```

activesupport/CHANGELOG.md
Edouard-chin: Removed the changelog altogether based on Jean's feedback

guides/source/configuring.md
Edouard-chin: Ditto

activesupport/lib/active_support/ractors.rb
Edouard-chin: I need to access the argument (in the event that the `action` is nil and the proc has to be early returned as is, or when the `action` is warn and the proc has to be also returned as is), so I can't do `...`

activesupport/lib/active_support/railtie.rb
Edouard-chin: Removed the active_support configuration based on Jeans'comment. Opted to keep `unshareable_proc_action` as `ActiveSupport::Ractors.ractor_unshareable_proc_action` sounded weird.

Edouard-chin:
> Why should we commit to public APIs, documentations, etc?

@byroot Thanks for your feedback. I hope I understood your comment correctly, but I was under the impression that you were ok with introducing this change in Rails, however without documenting them yet or adding railtie configuration. Am I correct ?

byroot:
> Am I correct ?

Yes, I don't mind the feature existing, I just want us to be able to entirely change it without having to go through a deprecation cycle if it happens that we need it to be different (which is likely given how early all this work is).

activesupport/lib/active_support/ractors.rb
gmcgibbon: Ok makes sense, but we still don't accept `&block` here, which we need to. Please make that change. I'm using this in https://github.com/rails/rails/pull/57629 and had to make changes to support this.

gmcgibbon (approved):

PR #57629, opened 2026-06-09
Make callbacks ractor-safe

gmcgibbon:
### Motivation / Background

This Pull Request makes `ActiveSupport::Callbacks` ractor-safe.

### Detail

This Pull Request changes `ActiveSupport::Callbacks` to do the following:
- Makes user-supplied procs Ractor-shareable when configured
- Makes `__callbacks` / `CallbackChain` shareable
- Makes `set_callbacks` build a new chain instead of mutating the existing one

Ractor shareable behaviour switch taken from https://github.com/rails/rails/pull/57626, which will merge before this.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/callbacks.rb
+++ b/activesupport/lib/active_support/callbacks.rb
@@ -67,7 +67,7 @@ module ActiveSupport
 
     included do
       extend ActiveSupport::DescendantsTracker
-      class_attribute :__callbacks, instance_writer: false, instance_predicate: false, default: {}
+      class_attribute :__callbacks, instance_writer: false, instance_predicate: false, default: {}.freeze
     end
 
     CALLBACK_FILTER_TYPES = [:before, :after, :around].freeze
@@ -153,7 +153,8 @@ module ActiveSupport
       module Conditionals # :nodoc: all
         class Value
           def initialize(&block)
-            @block = block
+            @block = Ractors.shareable_proc(&block)
+            freeze
           end
           def call(target, value); @block.call(value); end
         end
@@ -245,11 +246,12 @@ module ActiveSupport
 
         def initialize(name, filter, kind, options, chain_config)
           @chain_config = chain_config
-          @name    = name
-          @kind    = kind
-          @filter  = filter
-          @if      = check_conditionals(options[:if])
-          @unless  = check_conditionals(options[:unless])
+          @name            = name
+          @kind            = kind
+          @original_filter = filter
+          @filter          = try_shareable_proc(filter)
+          @if              = check_conditionals(options[:if])
+          @unless          = check_conditionals(options[:unless])
 
           compiled
         end
@@ -267,7 +269,7 @@ module ActiveSupport
         end
 
         def matches?(_kind, _filter)
-          @kind == _kind && filter == _filter
+          @kind == _kind && (filter == _filter || (@original_filter  == _filter.object_id))
         end
 
         def duplicates?(other)
@@ -296,6 +298,19 @@ module ActiveSupport
             end
         end
 
+        def freeze # :nodoc:
+          return self if frozen?
+
+          @filter = Ractors.make_shareable(@filter)
+          @if = make_conditionals_ractor_shareable(@if)
+          @unless = make_conditionals_ractor_shareable(@unless)
+          @compiled = nil
+
+          compiled
+          super
+          self
+        end
+
         # Wraps code with filter
         def apply(callback_sequence)
           compiled.apply(callback_sequence)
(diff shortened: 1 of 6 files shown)


activesupport/lib/active_support/callbacks.rb
```diff
@@ -247,7 +248,8 @@ def initialize(name, filter, kind, options, chain_config)
           @chain_config = chain_config
           @name    = name
           @kind    = kind
-          @filter  = filter
+          @filter_id = filter.object_id if filter.is_a?(::Proc)
```
byroot: Is this because `try_shareable_proc` might return a different Proc instance, and you still need to be able to compare it in `matches?`.

I think it's preferable to store `@original_filter` and compare that. Accessing `object_id` has side effect, it modify the object shape and might even resize it, we want to avoid that as much as possible.

activesupport/lib/active_support/callbacks.rb
```diff
@@ -321,14 +336,25 @@ def check_conditionals(conditionals)
...
+          end
+
+          def try_shareable_proc(object)
+            object.is_a?(Proc) ? Ractors.shareable_proc_attempt(object) : object
+          end
+
+          def make_conditionals_ractor_shareable(conditionals)
+            return conditionals if conditionals.empty?
+
+            conditionals.map { |conditional| Ractors.make_shareable(conditional) }.freeze
```
byroot:
```suggestion
            Ractors.make_shareable(conditionals)
```

Make shareable is recursive and freeze things, no need to map. If you don't want to freeze the original object, you can use `make_shareable(conditions, copy: true)`.

activesupport/lib/active_support/callbacks.rb
```diff
@@ -940,20 +1003,27 @@ def get_callbacks(name) # :nodoc:
             __callbacks[name.to_sym]
           end
 
-          def set_callbacks(name, callbacks) # :nodoc:
-            # HACK: We're making assumption on how `class_attribute` is implemented
-            # to save constantly duping the callback hash. If this desync with class_attribute
-            # we'll lose the optimization, but won't cause an actual behavior bug.
```
byroot, line 979: It's a bit annoying to lose that optimization, as it will impact boot performance (on large apps, there's lots of callbacks to register).

Should this be a case of keeping the simple, mutable initialization code, and then making the structure Ractor safe in one go when calling `freeze` on it?

byroot:
I know I'm gonna sound like the usual party pooper, but this is a lot of extra complexity on what is already a very complex to work with and performance sensitive piece of code for no tangible benefit at this stage.

I have a vague idea of why you are doing this based on backchanels, but I think it would really help if there was public visibility on the goal and blockers.

Because if this was the last piece of the puzzle, I'd understand taking extra complexity here, but if it's still unclear whether the whole thing is even possible, then it's a much harder sell.

gmcgibbon:
@byroot No problem. I can park this PR until we have ractor-safe features built that need this. I think what's important right now is getting feedback on the approach. I'll review your comments.

activesupport/lib/active_support/callbacks.rb
```diff
@@ -358,14 +383,16 @@ def expand(target, value, block)
           end
 
           def make_lambda
-            lambda do |target, value, &block|
-              target.send(@method_name, &block)
+            method_name = @method_name
+            Ractors.shareable_proc_attempt do |target, value, &block|
```
matthewd: We shouldn't need to 'attempt' for built-in procs like this, I don't think -- only user-supplied ones?

Edouard-chin:
@byroot 👋 We'd like to proceed with this PR as it's one of the few remaining code paths to let us create a new rails app, scaffold a resource and serve its request from a Ractor. 

Regarding your concerns on performance, I wrote this benchmark https://gist.github.com/Edouard-chin/16662be264d62d88dfa10f4d50fd328f which shows no significant difference (both when defining callbacks and running them).

I also tackled complexity by making the diff a bit smaller, but ultimately yes there is some added complexity, although not by much. Happy to address your feedback if you have others, but just wanted to let you know we'd like to proceed with this patch.

activesupport/lib/active_support/callbacks.rb
```diff
@@ -276,7 +278,7 @@ def merge_conditional_options(chain, if_option:, unless_option:)
         end
 
         def matches?(_kind, _filter)
-          @kind == _kind && filter == _filter
+          @kind == _kind && (filter == _filter || (@original_filter == _filter.object_id))
```
byroot, line 281: `filter` is a `Proc` here, right?

byroot:
Go ahead.

PR #57633, opened 2026-06-09
Preserve falsy default value when building `HashWithIndifferentAccess`

55728:
### Summary

`HashWithIndifferentAccess.new` copies the source hash's default value so that reads for missing keys behave consistently after conversion (introduced in 6e574e8a11 to make `.new` match `.new_from_hash_copying_default`). The copy is guarded by `if hash.default`:

```ruby
self.default = hash.default if hash.default
```

Because `false` (and `nil`) are falsy in Ruby, a source hash built with a deliberately-falsy default — e.g. `Hash.new(false)`, a common pattern for "unknown keys are off" flag maps — silently loses its default. Reads for missing keys then return `nil` instead of `false`.

```ruby
source = Hash.new(false)
source["a"] = 1
hwia = ActiveSupport::HashWithIndifferentAccess.new(source)

hwia.default     # => nil   (expected: false)
hwia["missing"]  # => nil   (expected: false)
```

### Fix

Guard on `unless hash.default.nil?` so a falsy-but-present default is preserved. A plain hash's default is `nil`, so the common case is a no-op.

```ruby
self.default = hash.default unless hash.default.nil?
```

### Notes

- One-line change in `activesupport/lib/active_support/hash_with_indifferent_access.rb`.
- Adds a regression test; full HWIA suite (96 runs) is green.
- No CHANGELOG entry (simple bug fix, not a feature/deprecation).

### DIFF
--- a/activesupport/lib/active_support/hash_with_indifferent_access.rb
+++ b/activesupport/lib/active_support/hash_with_indifferent_access.rb
@@ -75,7 +75,7 @@ module ActiveSupport
         update(constructor)
 
         hash = constructor.is_a?(Hash) ? constructor : constructor.to_hash
-        self.default = hash.default if hash.default
+        self.default = hash.default unless hash.default.nil?
         self.default_proc = hash.default_proc if hash.default_proc
       else
         super(constructor)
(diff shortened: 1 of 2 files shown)


55728:
Thanks for the lightning-fast merge @fatkodima ⚡😊

PR #57641, opened 2026-06-09
Enable per-pool query log tags prepend config

hmcguire-shopify:
[Refactor query_logs test helper to be generic](https://github.com/rails/rails/pull/57641/commits/a3a0668f60177041d80db20b22651f80683b9adf) 

So that it can be used with other keys in the following commits

---

[Allow disabling query_log_tags per pool](https://github.com/rails/rails/pull/57641/commits/5d813755af2a289701d82a21318d6413da7b8f5d) 

A `query_log_tags` key in a `database.yml` entry configures query log
tags for connections checked out from that pool. Set it to `false` to
opt the pool out of tagging even when tags are enabled globally, or to
a hash of options such as `format` to override the global
`config.active_record.query_log_tags_format`. Pools without an explicit
configuration fall back to the global settings.

This supersedes the top-level `query_log_tags_format` key, folding the
per-pool formatter into the same hash and adding the ability to disable
tagging per pool.

```yaml
production:
 primary:
   database: primary
 analytics:
   database: analytics
   query_log_tags:
     format: sqlcommenter
 replica:
   database: replica
   replica: true
   query_log_tags: false
```

---

[Enable per-pool query log tags prepend config](https://github.com/rails/rails/pull/57641/commits/0551469a644b3b51d5d785aad1e57f5177e06e20) 

A `prepend_comment` key under `query_log_tags` in a `database.yml` entry
overrides the global `config.active_record.query_log_tags_prepend_comment`
for connections in that pool, so different databases can prepend or append
the query comment.

```yaml
production:
 primary:
   database: primary
 analytics:
   database: analytics
   query_log_tags:
     prepend_comment: true
```

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -69,12 +69,13 @@
 
 *   Allow query log tags to be configured per connection pool.
 
-    A `query_log_tags` key in a `database.yml` entry configures tagging for
-    connections checked out from that pool. Set it to `false` to opt a pool out
-    of tagging even when tags are enabled globally, or to a hash of options:
-    `format` will override the global
-    `config.active_record.query_log_tags_format` so different databases can emit
-    `:legacy` or `:sqlcommenter` formatted comments.
+    Several query log tag settings can be overridden per pool with a `query_log_tags`
+    key in a `database.yml` entry. Its `format` and `prepend_comment` options override
+    the global `config.active_record.query_log_tags_format` and
+    `config.active_record.query_log_tags_prepend_comment`, so different databases can
+    emit `:legacy` or `:sqlcommenter` comments and prepend or append them. Setting
+    `query_log_tags` to `false` instead opts a pool out of tagging entirely, even when
+    tags are enabled globally.
 
     ```yaml
     production:
@@ -84,6 +85,7 @@
         database: analytics
         query_log_tags:
           format: sqlcommenter
+          prepend_comment: false
       replica:
         database: replica
         replica: true
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -67,11 +67,14 @@
 
     *Rosa Gutierrez*
 
-*   Allow the query log tags format to be configured per connection pool.
+*   Allow query log tags to be configured per connection pool.
 
-    A `query_log_tags_format` key in a `database.yml` entry overrides the global
-    `config.active_record.query_log_tags_format` for connections in that pool, so
-    different databases can emit `:legacy` or `:sqlcommenter` formatted comments.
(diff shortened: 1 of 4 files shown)


matthewd:
I don't think it's useful to configure `query_log_tags` per connection (or more accurately, I don't think it's useful to do so from within `database.yml`, where you can't use custom procs etc). And without that, varying `cache_query_log_tags` also seems unhelpful.

But it does seem like `query_log_tags_enabled` might desirable?

(Honestly I feel like this could probably all be collapsed into a single option with 5 values -- off, prepend colons, prepend equals, append colons, append equals -- but that means introducing a mismatch with how the existing config works, or churning the existing config, and neither of those seem high-value.)

hmcguire-shopify:
> I don't think it's useful to configure `query_log_tags` per connection (or more accurately, I don't think it's useful to do so from within `database.yml`, where you can't use custom procs etc). And without that, varying `cache_query_log_tags` also seems unhelpful.

👍 Yeah I had similar thoughts.

> But it does seem like query_log_tags_enabled might desirable?

🤔 yeah, potentially

> (Honestly I feel like this could probably all be collapsed into a single option with 5 values -- off, prepend colons, prepend equals, append colons, append equals -- but that means introducing a mismatch with how the existing config works, or churning the existing config, and neither of those seem high-value.)

Agree. I don't particularly like `query_log_tags_prepend_comment` as a config parameter either. Maybe we could nest the 2 (or 3 with `enabled`) that seem reasonable to do?

```
production:
  primary:
    database: primary
  analytics:
    database: analytics
    query_log_tags:
      format: sqlcommenter
      prepend_comment: false # maybe `action: prepend` would be "ideal", but churn
  another:
    database: three
    query_log_tags:
      enabled: false
```

nvasilevski (approved):

adrianna-chang-shopify (approved):

matthewd:
@hmcguire-shopify yeah, I think I like that. It does break symmetry with the standard config values, but still seems like it's probably worth it to avoid spreading out e.g. in HashConfig, relative to how niche these options are. 

Simple `query_log_tags: false` for the disabled case?

hmcguire-shopify:
> Simple `query_log_tags: false` for the disabled case?

Something I thought of while iterating: if we support `query_log_tags: false` it probably makes sense to also have `query_log_tags: true` for "I want global config for everything but only these dbs enabled".

Edit: hmm, that's actually more complicated, because `ActiveRecord.query_transformers` is only added when globally `true`. We could maybe leave `false` as the opt-out without mentioning `true` as an opt-in (because it wouldn't work without a larger change), but the inconsistency is unfortunate

nvasilevski (approved):

PR #57642, opened 2026-06-09
Remove some memoization in Active Record and Active Model

gmcgibbon:
### Motivation / Background

This Pull Request has been created because Active Records have a lot of memoized values. This makes it difficult to use them in non-main Ractors. We can remove some of the unnecessary memorization with minimal performance impact. Here's a benchmark:

<details>
  <summary>Benchmark Source</summary>

  ```ruby
# frozen_string_literal: true

require "bundler/inline"

RAILS_ROOT = File.expand_path("../..", __dir__)
LABEL = ENV["BRANCH"] || "main"

gemfile(true) do
  source "https://rubygems.org"

  gem "activesupport", path: RAILS_ROOT
  gem "activemodel", path: RAILS_ROOT
  gem "activerecord", path: RAILS_ROOT

  gem "sqlite3"
  gem "benchmark-ips"
end

require "active_record"
require "benchmark/ips"
require "tmpdir"

ActiveRecord::Base.establish_connection(ENV["DATABASE_URL"])
ActiveRecord::Base.logger = nil

ActiveRecord::Schema.verbose = false
    t.string :title
  end
end

end

class DefaultScopedPost < ActiveRecord::Base
  self.table_name = "posts"
  default_scope { where.not(title: nil) }
end

Post.create!(title: "hello")

post = Post.first
title = post.title

SCENARIOS = {
  "Post#title"                  => -> { post.title },
  "Post#method_missing(:title)" => -> { post.public_send(:title) },
  "Post.arel_table"             => -> { Post.arel_table },
  "Post.column_names"           => -> { Post.column_names },
  "Post#to_partial_path"        => -> { post.to_partial_path },
  "Post.new"                    => -> { Post.new },
  "Post#respond_to?(:title)"    => -> { post.respond_to?(:title) },
  "DefaultScopedPost.all"       => -> { DefaultScopedPost.all.to_a },
  "Post.where(title:).load"     => -> { Post.where(title: title).load },
}

SCENARIOS.each do |name, callable|
  # A stable, filesystem-safe hold file per scenario so both runs line up.
  hold_file = File.join(Dir.tmpdir, "ar_bench_#{name.gsub(/\W+/, '_')}.json")

  puts
  puts " #{name} ".center(80, "=")
  puts

  Benchmark.ips do |x|
    x.config(time: 3, warmup: 1)
    x.report("#{name} (#{LABEL})", &callable)
    x.save!(hold_file)
    x.compare!
  end
end
  ```
</details>

Main
```
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin23]
                 Post#title   760.612k i/100ms
Post#method_missing(:title)   262.269k i/100ms
            Post.arel_table     2.823M i/100ms
          Post.column_names     3.391M i/100ms
       Post#to_partial_path     2.166M i/100ms
                   Post.new    30.420k i/100ms
   Post#respond_to?(:title)   487.990k i/100ms
         Post.where(title:)    18.058k i/100ms
      DefaultScopedPost.all    12.463k i/100ms
    Post.where(title:).load     1.257k i/100ms
                 Post#title      7.494M (± 3.5%) i/s  (133.43 ns/i) -     38.031M in   5.074489s
Post#method_missing(:title)      2.667M (± 8.2%) i/s  (374.93 ns/i) -     13.376M in   5.014897s
            Post.arel_table     26.707M (± 2.5%) i/s   (37.44 ns/i) -    135.494M in   5.073355s
          Post.column_names     30.449M (± 5.1%) i/s   (32.84 ns/i) -    152.590M in   5.011309s
       Post#to_partial_path     19.918M (± 4.7%) i/s   (50.21 ns/i) -     99.643M in   5.002704s
                   Post.new    293.933k (± 6.9%) i/s    (3.40 μs/i) -      1.491M in   5.071158s
   Post#respond_to?(:title)      4.749M (± 2.0%) i/s  (210.57 ns/i) -     23.912M in   5.035120s
         Post.where(title:)    175.377k (± 9.2%) i/s    (5.70 μs/i) -    884.842k in   5.045363s
      DefaultScopedPost.all    124.386k (± 7.3%) i/s    (8.04 μs/i) -    623.150k in   5.009799s
    Post.where(title:).load     12.434k (± 7.5%) i/s   (80.42 μs/i) -     62.850k in   5.054511s
```
This branch (ed35c9e8126b1618b5f6a7d88e8b66534a13ff5c)
```
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin23]
                 Post#title   741.100k i/100ms
Post#method_missing(:title)   260.044k i/100ms
            Post.arel_table   507.494k i/100ms
          Post.column_names   383.205k i/100ms
       Post#to_partial_path   646.323k i/100ms
                   Post.new    25.564k i/100ms
   Post#respond_to?(:title)   483.966k i/100ms
         Post.where(title:)    14.240k i/100ms
      DefaultScopedPost.all    10.621k i/100ms
    Post.where(title:).load     1.105k i/100ms
                 Post#title      7.369M (± 2.8%) i/s  (135.71 ns/i) -     37.055M in   5.028737s
Post#method_missing(:title)      2.628M (± 5.4%) i/s  (380.55 ns/i) -     13.262M in   5.046969s
            Post.arel_table      5.215M (± 5.9%) i/s  (191.76 ns/i) -     26.390M in   5.060544s
          Post.column_names      4.029M (± 5.1%) i/s  (248.21 ns/i) -     20.310M in   5.041067s
       Post#to_partial_path      6.164M (± 6.6%) i/s  (162.24 ns/i) -     31.024M in   5.033123s
                   Post.new    255.921k (± 4.2%) i/s    (3.91 μs/i) -      1.304M in   5.094409s
   Post#respond_to?(:title)      4.798M (± 2.0%) i/s  (208.40 ns/i) -     24.198M in   5.043027s
         Post.where(title:)    143.167k (± 6.9%) i/s    (6.98 μs/i) -    726.240k in   5.072694s
      DefaultScopedPost.all    108.206k (± 5.7%) i/s    (9.24 μs/i) -    541.671k in   5.005915s
    Post.where(title:).load     11.534k (± 6.1%) i/s   (86.70 μs/i) -     58.565k in   5.077673s
```

### Detail

This Pull Request changes Active Record and Active Model by removing the following memoization:
- `arel_table`
- `attribute_method_patterns_cache`
- `attributes_builder`
- `columns_names`
- `default_scope_override`

We're doing this to make Active Record classes more Ractor-friendly. It is possible some of these ivars need to be restored or usage needs to be refactored if the performance impact is too costly. 

### Checklist

### DIFF
--- a/activemodel/lib/active_model/attribute_methods.rb
+++ b/activemodel/lib/active_model/attribute_methods.rb
@@ -68,6 +68,7 @@ module ActiveModel
     CALL_COMPILABLE_REGEXP = /\A[a-zA-Z_]\w*[!?]?\z/
 
     included do
+      @attribute_method_patterns_cache = Concurrent::Map.new(initial_capacity: 4)
       class_attribute :attribute_aliases, instance_writer: false, default: {}
       class_attribute :attribute_method_patterns, instance_writer: false, default: [ ClassMethods::AttributeMethodPattern.new ]
     end
@@ -219,7 +220,7 @@ module ActiveModel
           attribute_method_patterns.each do |pattern|
             alias_attribute_method_definition(code_generator, pattern, new_name, old_name)
           end
-          attribute_method_patterns_cache.clear
+          @attribute_method_patterns_cache.clear
         end
       end
 
@@ -313,7 +314,7 @@ module ActiveModel
           attribute_method_patterns.each do |pattern|
             define_attribute_method_pattern(pattern, attr_name, owner: owner, as: as)
           end
-          attribute_method_patterns_cache.clear
+          @attribute_method_patterns_cache.clear
         end
       end
 
@@ -376,7 +377,7 @@ module ActiveModel
         @generated_attribute_methods&.module_eval do
           undef_method(*instance_methods)
         end
-        attribute_method_patterns_cache.clear
+        @attribute_method_patterns_cache.clear
       end
 
       def aliases_by_attribute_name # :nodoc:
@@ -387,7 +388,7 @@ module ActiveModel
         def inherited(base) # :nodoc:
           super
           base.class_eval do
-            @attribute_method_patterns_cache = nil
+            @attribute_method_patterns_cache = Concurrent::Map.new(initial_capacity: 4)
             @aliases_by_attribute_name = nil
             @generated_attribute_methods = nil
           end
@@ -405,21 +406,8 @@ module ActiveModel
           @generated_attribute_methods&.method_defined?(method_name)
         end
 
-        # The methods +method_missing+ and +respond_to?+ of this module are
-        # invoked often in a typical rails, both of which invoke the method
-        # +matched_attribute_method+. The latter method iterates through an
-        # array doing regular expression matches, which results in a lot of
-        # object creations. Most of the time it returns a +nil+ match. As the
-        # match result is always the same given a +method_name+, this cache is
-        # used to alleviate the GC, which ultimately also speeds up the app
-        # significantly (in our case our test suite finishes 10% faster with
-        # this cache).
-        def attribute_method_patterns_cache
-          @attribute_method_patterns_cache ||= Concurrent::Map.new(initial_capacity: 4)
-        end
-
         def attribute_method_patterns_matching(method_name)
-          attribute_method_patterns_cache.compute_if_absent(method_name) do
+          @attribute_method_patterns_cache.compute_if_absent(method_name) do
             attribute_method_patterns.filter_map { |pattern| pattern.match(method_name) }
           end
         end
--- a/activerecord/lib/active_record/core.rb
+++ b/activerecord/lib/active_record/core.rb
@@ -11,6 +11,8 @@ module ActiveRecord
     include ActiveModel::Access
 
     included do
+      @arel_table = Arel::Table.new(klass: self)
+
       ##
       # :singleton-method:
       #
@@ -400,7 +402,7 @@ module ActiveRecord
 
       # Returns an instance of +Arel::Table+ loaded with the current table name.
       def arel_table # :nodoc:
-        @arel_table ||= Arel::Table.new(table_name, klass: self)
+        @arel_table
       end
 
       def predicate_builder # :nodoc:
@@ -431,7 +433,7 @@ module ActiveRecord
           end
 
           subclass.class_eval do
-            @arel_table = nil
+            @arel_table = Arel::Table.new(klass: self)
             @predicate_builder = nil
             @inspection_filter = nil
             @filter_attributes ||= nil
--- a/activerecord/lib/active_record/model_schema.rb
+++ b/activerecord/lib/active_record/model_schema.rb
@@ -277,7 +277,7 @@ module ActiveRecord
         end
 
         @table_name        = value
-        @arel_table        = nil
+        @arel_table        = Arel::Table.new(klass: self)
         @sequence_name     = nil unless @explicit_sequence_name
         @predicate_builder = nil
       end
@@ -448,10 +448,8 @@ module ActiveRecord
       end
 
       def attributes_builder # :nodoc:
-        @attributes_builder ||= begin
-          defaults = _default_attributes.except(*(column_names - Array(primary_key)))
-          ActiveModel::AttributeSet::Builder.new(attribute_types, defaults)
-        end
+        defaults = _default_attributes.except(*(column_names - Array(primary_key)))
+        ActiveModel::AttributeSet::Builder.new(attribute_types, defaults)
       end
 
       def columns_hash # :nodoc:
@@ -508,7 +506,7 @@ module ActiveRecord
 
       # Returns an array of column names as strings.
       def column_names
-        @column_names ||= columns.map(&:name).freeze
+        columns.map(&:name).freeze
       end
 
       def symbol_column_to_string(name_symbol) # :nodoc:
@@ -585,12 +583,10 @@ module ActiveRecord
         def reload_schema_from_cache(recursive = true)
           @_returning_columns_for_insert = nil
           @_returning_columns_for_update = nil
-          @arel_table = nil
-          @column_names = nil
+          @arel_table = Arel::Table.new(klass: self)
           @symbol_column_to_string_name_hash = nil
           @content_columns = nil
           @column_defaults = nil
-          @attributes_builder = nil
           @columns = nil
           @columns_hash = nil
           @schema_loaded = false
--- a/activerecord/lib/active_record/attribute_methods.rb
+++ b/activerecord/lib/active_record/attribute_methods.rb
@@ -81,7 +81,7 @@ module ActiveRecord
         attribute_method_patterns.each do |pattern|
           alias_attribute_method_definition(code_generator, pattern, new_name, old_name)
         end
-        attribute_method_patterns_cache.clear
+        @attribute_method_patterns_cache.clear
       end
 
       def alias_attribute_method_definition(code_generator, pattern, new_name, old_name) # :nodoc:
--- a/activerecord/lib/active_record/attribute_methods/primary_key.rb
+++ b/activerecord/lib/active_record/attribute_methods/primary_key.rb
@@ -127,16 +127,11 @@ module ActiveRecord
             self._primary_key_definition = ActiveRecord::Key.for(value)
 
             include CompositePrimaryKey if primary_key_definition.composite?
-
-            @attributes_builder = nil
           end
 
           private
             def inherited(base)
               super
-              base.class_eval do
-                @attributes_builder = nil
-              end
             end
         end
     end
--- a/activerecord/lib/active_record/scoping/default.rb
+++ b/activerecord/lib/active_record/scoping/default.rb
@@ -17,7 +17,6 @@ module ActiveRecord
       included do
         # Stores the default scope for the class.
         class_attribute :default_scopes, instance_writer: false, instance_predicate: false, default: []
-        class_attribute :default_scope_override, instance_writer: false, instance_predicate: false, default: nil
       end
 
       module ClassMethods
@@ -145,11 +144,7 @@ module ActiveRecord
           def build_default_scope(relation = relation(), all_queries: nil)
             return if abstract_class?
 
-            if default_scope_override.nil?
-              self.default_scope_override = !Base.is_a?(method(:default_scope).owner)
-            end
-
-            if default_scope_override
+            if !Base.is_a?(method(:default_scope).owner)
               # The user has defined their own default scope method, so call that
               evaluate_default_scope do
                 relation.scoping { default_scope }
--- a/activerecord/lib/arel/table.rb
+++ b/activerecord/lib/arel/table.rb
@@ -8,10 +8,11 @@ module Arel # :nodoc: all
     @engine = nil
     class << self; attr_accessor :engine; end
 
-    attr_accessor :name
+    attr_writer :name
     attr_reader :table_alias, :klass
 
-    def initialize(name, as: nil, klass: nil, type_caster: klass&.type_caster)
+    def initialize(name = nil, as: nil, klass: nil, type_caster: klass&.type_caster)
+      raise ArgumentError, "A name or klass is required." unless klass || name
       name = name.name if name.is_a?(Symbol)
 
       @name = name
@@ -25,6 +26,10 @@ module Arel # :nodoc: all
         as = nil
       end
       @table_alias = as
+    end
+
+    def name
+      @name || @klass&.table_name
     end
 
     def alias(name = "#{self.name}_2")
@@ -89,7 +94,7 @@ module Arel # :nodoc: all
       # Perf note: aliases and table alias is excluded from the hash
       #  aliases can have a loop back to this table breaking hashes in parent
       #  relations, for the vast majority of cases @name is unique to a query
-      @name.hash
+      name.hash
     end
 
     def eql?(other)
--- a/activerecord/test/cases/arel/support/fake_record.rb
+++ b/activerecord/test/cases/arel/support/fake_record.rb
@@ -124,6 +124,13 @@ module FakeRecord
 
   class Base
     attr_accessor :connection_pool
+    singleton_class.attr_accessor :table_name
+
+    class << self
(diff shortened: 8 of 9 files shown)


gmcgibbon:
`respond_to?` calls seem to be considerably slower. We may need to restore some memoization there. I think I need to restore `predicate_builder` and `symbol_column_to_string_name_hash`.

tenderlove:
> `respond_to?` calls seem to be considerably slower.

Can we refactor that codepath such that we don't need to do the "respond_to" call?

gmcgibbon:
The code path is the benchmark. Happy to remove it from the script if you think calling respond_to? isn't important. Maybe it isn't. I do `x.report("Post#respond_to?(:title)") { post.respond_to?(:title) }`

tenderlove:
> The code path is the benchmark. Happy to remove it from the script if you think calling respond_to? isn't important. Maybe it isn't. I do `x.report("Post#respond_to?(:title)") { post.respond_to?(:title) }`

Oh. It looks faster on this branch?  I think benchmark/ips has a feature where you can save a run and then compare it with a new one. Maybe that would make reading the benchmarks easier?

gmcgibbon:
Yes, with `symbol_column_to_string_name_hash` restored. I'll post a separate benchmark and commit with it removed so it is easier to understand!

gmcgibbon:
Here's what the latest commit looks like (0d48dbd37e9e58de4aa4679cf48d68555936e309):
```
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin23]
                 Post#title   802.439k i/100ms
Post#method_missing(:title)   261.806k i/100ms
            Post.arel_table   517.361k i/100ms
          Post.column_names   411.083k i/100ms
       Post#to_partial_path   637.101k i/100ms
                   Post.new    27.396k i/100ms
   Post#respond_to?(:title)   104.599k i/100ms
         Post.where(title:)    15.133k i/100ms
      DefaultScopedPost.all    11.210k i/100ms
    Post.where(title:).load     1.170k i/100ms
                 Post#title      7.584M (± 2.4%) i/s  (131.85 ns/i) -     38.517M in   5.078591s
Post#method_missing(:title)      2.645M (± 3.9%) i/s  (378.11 ns/i) -     13.352M in   5.048582s
            Post.arel_table      5.326M (± 2.9%) i/s  (187.76 ns/i) -     26.903M in   5.051254s
          Post.column_names      4.036M (± 3.6%) i/s  (247.76 ns/i) -     20.554M in   5.092425s
       Post#to_partial_path      6.449M (± 5.2%) i/s  (155.06 ns/i) -     32.492M in   5.038230s
                   Post.new    262.361k (± 2.8%) i/s    (3.81 μs/i) -      1.315M in   5.012209s
   Post#respond_to?(:title)      1.034M (± 3.5%) i/s  (967.31 ns/i) -      5.230M in   5.058958s
         Post.where(title:)    148.677k (± 4.6%) i/s    (6.73 μs/i) -    756.650k in   5.089232s
      DefaultScopedPost.all    112.257k (± 4.9%) i/s    (8.91 μs/i) -    571.710k in   5.092886s
    Post.where(title:).load     12.164k (± 5.8%) i/s   (82.21 μs/i) -     60.840k in   5.001685s
```

gmcgibbon:
We should also validate the `_to_partial_path` slow down isn't an issue. I assume this may affect view performance, but it may not matter.

activerecord/lib/active_record/model_schema.rb
```diff
@@ -510,8 +510,7 @@ def column_names
       end
 
       def symbol_column_to_string(name_symbol) # :nodoc:
-        @symbol_column_to_string_name_hash ||= column_names.index_by(&:to_sym)
-        @symbol_column_to_string_name_hash[name_symbol]
+        column_names.index_by(&:to_sym)[name_symbol]
       end
```
hmcguire-shopify, line 515: This method is only called in a single place to cache `sym -> str` lookup. If we aren't caching the hash anymore, we probably shouldn't create it in the first place (vs. just doing the single `str -> sym` conversion we actually want)

hmcguire-shopify:
> I think benchmark/ips has a feature where you can save a run and then compare it with a new one. Maybe that would make reading the benchmarks easier?

Highly recommend this as well, some of these don't look like "minimal performance impact"

gmcgibbon:
```
Comparison:
         Post#title (main):  7412490.5 i/s
Post#title (ar_less_ivars):  7391397.1 i/s - same-ish: difference falls within error

Comparison:
         Post#method_missing(:title) (main):  2552861.4 i/s
Post#method_missing(:title) (ar_less_ivars):  2482636.0 i/s - same-ish: difference falls within error

Comparison:
         Post.arel_table (main): 30164025.5 i/s
Post.arel_table (ar_less_ivars):  5048875.4 i/s - 5.97x  slower

Comparison:
         Post.column_names (main): 29241229.6 i/s
Post.column_names (ar_less_ivars):  3823151.8 i/s - 7.65x  slower

Comparison:
         Post#to_partial_path (main): 22422575.2 i/s
Post#to_partial_path (ar_less_ivars):  6241020.4 i/s - 3.59x  slower

Comparison:
         Post.new (main):   300572.0 i/s
Post.new (ar_less_ivars):   261801.2 i/s - 1.15x  slower

Comparison:
         Post#respond_to?(:title) (main):  4725043.5 i/s
Post#respond_to?(:title) (ar_less_ivars):  1026186.3 i/s - 4.60x  slower

Comparison:
         DefaultScopedPost.all (main):   126147.1 i/s
DefaultScopedPost.all (ar_less_ivars):   107502.3 i/s - 1.17x  slower

Comparison:
         Post.where(title:).load (main):    11984.5 i/s
Post.where(title:).load (ar_less_ivars):    11463.1 i/s - same-ish: difference falls within error
```

That's a lot of data. What jumps out to me is `Post#respond_to?(:title)` and `Post#to_partial_path`. We need those ivars I think. The results for `DefaultScopedPost.all` and `Post.new` may also not be trivial. I'll reduce the surface area of this PR to not affect these areas. I think `Post.arel_table` being in the ballpark of 5x slower is expected because it isn't testing a real use-case. Might need a different public API benchmark for that one.

activerecord/lib/active_record/model_schema.rb
tenderlove: 💯 

I'm not sure if we sort the columns, but if we did we could bsearch and have "nearly hash lookup" performance.  Also we could probably add a "name as symbol" (`name_as_sym` or `sym_name` or something) method on the column and quit doing the to_sym.

activerecord/lib/active_record/core.rb
```diff
@@ -400,7 +400,7 @@ def inspect # :nodoc:
 
       # Returns an instance of +Arel::Table+ loaded with the current table name.
       def arel_table # :nodoc:
-        @arel_table ||= Arel::Table.new(table_name, klass: self)
+        Arel::Table.new(table_name, klass: self)
```
tenderlove: Could we eagerly allocate this?  I see the micro benchmarks are slower (which makes sense) but it feels like we could eagerly allocate, and then fall back to allocating an object for cases where we can't (anonymous AR models for example, I think?)

gmcgibbon: We cannot because it needs the table_name which is configurable by the class or computed. It has to be lazy at some level I think as a result.

activerecord/lib/active_record/model_schema.rb
gmcgibbon: I'll take a closer look at this. If it's too hard to get similar perf here let's defer it to another PR.

gmcgibbon: ~Columns are not ordered by default which causes failures in the test suite. I could try to order the columns hash when loading from the schema / db instead, but I don't know if that breaks any public API. I assume not.~ I'm wrong. I can sort the memoized column names before bsearching.

gmcgibbon:
```
Comparison:
         Post#title (main):  7345950.4 i/s
Post#title (ar_less_ivars):  7009138.2 i/s - same-ish: difference falls within error

Comparison:
Post#method_missing(:title) (ar_less_ivars):  2602367.7 i/s
         Post#method_missing(:title) (main):  2602186.7 i/s - same-ish: difference falls within error

Comparison:
         Post.arel_table (main): 31670629.7 i/s
Post.arel_table (ar_less_ivars):  4816670.0 i/s - 6.58x  slower

Comparison:
         Post.column_names (main): 31459768.3 i/s
Post.column_names (ar_less_ivars):  3868909.1 i/s - 8.13x  slower

Comparison:
Post#to_partial_path (ar_less_ivars): 22442000.8 i/s
         Post#to_partial_path (main): 22060996.4 i/s - same-ish: difference falls within error

Comparison:
         Post.new (main):   293542.5 i/s
Post.new (ar_less_ivars):   231166.3 i/s - 1.27x  slower

Comparison:
         Post#respond_to?(:title) (main):  4807099.4 i/s
Post#respond_to?(:title) (ar_less_ivars):  3364502.4 i/s - 1.43x  slower

Comparison:
         DefaultScopedPost.all (main):   123151.0 i/s
DefaultScopedPost.all (ar_less_ivars):   104388.9 i/s - 1.18x  slower

Comparison:
         Post.where(title:).load (main):    11351.4 i/s
Post.where(title:).load (ar_less_ivars):    10777.3 i/s - same-ish: difference falls within error
```

Looks like default scope, new record creation, and respond_to? is still slow. 🤔  We may need `symbol_column_to_string_name_hash` after all.

activerecord/lib/active_record/core.rb
matthewd: It has the `klass`, so _possible_ it could ask that for the table name on demand? (Arguably just creates a different problem, and maybe a worse one... but maybe?)

gmcgibbon:
With that put back, we just have some extra slowness on initialization calls left to worry about:
```
Comparison:
         Post#title (main):  7578546.2 i/s
Post#title (ar_less_ivars):  7522791.6 i/s - same-ish: difference falls within error

Comparison:
         Post#method_missing(:title) (main):  2648693.5 i/s
Post#method_missing(:title) (ar_less_ivars):  2630078.8 i/s - same-ish: difference falls within error

Comparison:
         Post.arel_table (main): 30384379.8 i/s
Post.arel_table (ar_less_ivars):  5250713.5 i/s - 5.79x  slower

Comparison:
         Post.column_names (main): 29748030.7 i/s
Post.column_names (ar_less_ivars):  4028621.2 i/s - 7.38x  slower

Comparison:
         Post#to_partial_path (main): 21499198.6 i/s
Post#to_partial_path (ar_less_ivars): 20830830.3 i/s - same-ish: difference falls within error

Comparison:
         Post.new (main):   306622.2 i/s
Post.new (ar_less_ivars):   259771.3 i/s - 1.18x  slower

Comparison:
         Post#respond_to?(:title) (main):  4865446.5 i/s
Post#respond_to?(:title) (ar_less_ivars):  4828766.5 i/s - same-ish: difference falls within error

Comparison:
         DefaultScopedPost.all (main):   125433.6 i/s
DefaultScopedPost.all (ar_less_ivars):   110214.6 i/s - same-ish: difference falls within error

Comparison:
         Post.where(title:).load (main):    11894.2 i/s
Post.where(title:).load (ar_less_ivars):    11502.8 i/s - same-ish: difference falls within error
```
We need to restore `@finder_needs_type_condition` to fix that. I think we can calculate that early, but we would sacrifice the ability to detect type columns when we reload data from schema cache (I imagine in development). I'll push a commit restoring `finder_needs_type_condition?` too for now so that every measure besides calling column_names and arel_table directly are comparable with main:

```
Comparison:
Post.new (ar_less_ivars):   560736.5 i/s
         Post.new (main):   554267.0 i/s - same-ish: difference falls within error
```

gmcgibbon:
This LGTM now. I've attached the benchmark to the git commit.

activemodel/lib/active_model/attribute_methods.rb
```diff
@@ -415,7 +416,7 @@ def instance_method_already_implemented?(method_name)
         # significantly (in our case our test suite finishes 10% faster with
         # this cache).
         def attribute_method_patterns_cache
-          @attribute_method_patterns_cache ||= Concurrent::Map.new(initial_capacity: 4)
+          @attribute_method_patterns_cache
```
matthewd: This should probably now be an `attr_reader` at minimum, though I think I'd personally lean toward changing callers to just use the ivar directly. That would mean the comment needs a new home, but it feels a bit odd floating on this hollowed method anyway IMO... maybe put it on the initial assignment so it stays with the `C::Map.new`? (Though I'd _also_ argue that the comment should be drastically, if not entirely dropped... it seems to effectively be "this is accessed often enough to make a cache desirable; this is cacheable; caches are desirable" -- all good justification for adding it in the first place, but to me all implied by the cache's very existence now[^1].)

[^1]: Admittedly part of your current workstream is actively reviewing and questioning whether some caches still justify themselves 😅 -- but a comment that explains it was a good idea when it was added doesn't actually better-inform that process.

activerecord/lib/active_record/core.rb
matthewd: If you haven't yet, I'd like to consider the "eager `Arel::Table` lazily asks its `klass` for `table_name`" option -- there are known user-code patterns that call this very frequently (the `User.arel_table[:username]` pattern).

Alternatively and maybe even better, eagerly allocate it in `reset_table_name`?

activemodel/lib/active_model/attribute_methods.rb
gmcgibbon: - Method removed
- Comment dropped

Looks better to me 👍

activerecord/lib/active_record/core.rb
gmcgibbon: I brought back the ivar and used eager assignment. This changes `Post.arel_table` to be 1.11x slower instead of 5.70x  slower without memoization.

tenderlove (approved):

yahonda:
Looks like the CI error is relevant https://github.com/rails/rails/issues/57824

gmcgibbon:
I'll take a look. Thanks @yahonda

Issue #42577, opened 2021-06-23
`number_to_currency` printing negative zero when precision > 0


pjpires:
In https://github.com/rails/rails/pull/37865 `number_to_currency` was fixed to properly print values that would result in a negative zero when passing `precision: 0`. However, it doesn't work when precision is > 0 as seen below.

### Steps to reproduce

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
end

require "active_support"
require "action_view"
require "active_support/core_ext/object/blank"
require "minitest/autorun"

class BugTest < Minitest::Test
  include ActionView::Helpers::NumberHelper

  def test_stuff
    assert_equal "$0", number_to_currency(-0.456789, precision: 0) # This works fine
    assert_equal "$0.0", number_to_currency(-0.0456789, precision: 1) # This does not (prints "-$0.0")
  end
end
```

**Note**: calling `number_to_currency` without specifying `precision` also causes issues, since the default value for precision is > 0.

### Expected behavior

Given the above test "$0.0" should be printed.

### Actual behavior

"-$0.0" is printed instead.

### System configuration
**Rails version**:
main

**Ruby version**:
2.7.2

flavorjones:
@pjpires Thanks for asking this question. I believe this is functioning as designed. The tests for this method are:

```ruby
      def test_number_to_currency
        [@instance_with_helpers, TestClassWithClassNumberHelpers, ActiveSupport::NumberHelper].each do |number_helper|
          assert_equal("$1,234,567,890.50", number_helper.number_to_currency(1234567890.50))
          assert_equal("$1,234,567,890.51", number_helper.number_to_currency(1234567890.506))
          assert_equal("-$1,234,567,890.50", number_helper.number_to_currency(-1234567890.50))
          assert_equal("-$ 1,234,567,890.50", number_helper.number_to_currency(-1234567890.50, format: "%u %n"))
          assert_equal("($1,234,567,890.50)", number_helper.number_to_currency(-1234567890.50, negative_format: "(%u%n)"))
          assert_equal("$1,234,567,892", number_helper.number_to_currency(1234567891.50, precision: 0))
          assert_equal("$1,234,567,891", number_helper.number_to_currency(1234567891.50, precision: 0, round_mode: :down))
[...]

PR #57648, opened 2026-06-10
Fix `number_to_currency` ignoring `:round_mode` when choosing the negative format

55728:
### Summary

`number_to_currency` can silently drop the minus sign on a negative amount, or render `"-$0.00"` that #42577 was meant to eliminate, depending on the `:round_mode`:

```ruby
number_to_currency(-0.001, round_mode: :up)        # => "$0.01"   (negative amount shown as positive)
number_to_currency(-0.009, round_mode: :down)      # => "-$0.00"
number_to_currency(-0.005, round_mode: :half_even) # => "-$0.00"
```

The default round mode (half-up) is unaffected — the bug only surfaces when a non-default `:round_mode` is passed.

### Cause

Since 9a05443d1e (#42577), the converter only applies `:negative_format` when the rounded amount remains nonzero. But the guard decides negativity with a hardcoded round-half-up threshold:

```ruby
format = options[:negative_format] if options[:precision].nil? || (number_d * 10**options[:precision]) >= 0.5
```

while the value actually rendered is produced by `NumberToRoundedConverter`, which honors the documented `:round_mode` option. Whenever the hardcoded assumption disagrees with the requested round mode, the chosen format contradicts the rendered number.

### Fix

Decide the format from the actually-rounded value instead, by rounding through the same `RoundingHelper` machinery the converter uses:

```ruby
format = options[:negative_format] unless RoundingHelper.new(options).round(number_d).zero?
```

The default round mode, `precision: nil`, `significant: true`, and `precision: 0` behaviors are all unchanged.


55728:

PR #57651, opened 2026-06-10
Fix associations on a new record when the owner has a composite primary key

jsaubry:
### Motivation / Background

This Pull Request has been created because a `has_many` or `has_one` association on a new (unsaved) record returns an empty result when the owner has a composite primary key, even when every key column is populated and matching rows exist. The association is never queried — it just resolves to `[]/nil` without raising.

A single-key owner in the same situation queries correctly, so this is a composite-key parity break. Using the existing `Cpk::Book model (primary key [:author_id, :id], has_many :chapters, foreign_key: [:author_id, :book_id])`:

 ```ruby
   book = Cpk::Book.new(id: [1, 10])   # new_record?, but author_id + id are populated
   book.chapters.to_a
   # Before: []                     ← never queried
   # After:  [#<Cpk::Chapter ...>]  ← queries on (author_id, id)
 ```

 ### Detail

The cause is in `ActiveRecord::Associations::ForeignAssociation#foreign_key_present?`. It passed the composite `active_record_primary_key array` (e.g. `["author_id", "id"]`) to `attribute_present?`, which stringified the array, found no column by that name, and always returned false. That made `null_scope?` (`new_record? && !foreign_key_present?`) always true for an unsaved owner, short-circuiting the query.

 The fix checks each key column individually; single-column keys are unchanged:

 ```ruby
   Array(reflection.active_record_primary_key).all? do |key|
     owner.attribute_present?(key)
   end
 ```

`foreign_key_present?` is shared by `has_many` and `has_one` via `ForeignAssociation` (feeding `CollectionAssociation#null_scope?` and `Association#find_target?` respectively), so both are covered.

 ### Additional information

Same family as other recent composite-key fixes (#57534, #57158), but a distinct mechanism: the key array is misused as an attribute name when checking for presence. Tests were added to both `has_many_associations_test.rb` (`Cpk::Book → chapters`) and `has_one_associations_test.rb` (`Cpk::Order → book`), each covering the all-keys-present and a-key-missing cases — red on main, green with the fix.

 ### Checklist

 - [x] Tests are added or updated.
 - [x] CHANGELOG files are updated.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,9 @@
+*   Fix `has_many` and `has_one` associations on a new record returning an empty
+    result when the owner has a composite primary key, even when every primary
+    key column is populated.
+
+    *Jean-Samuel Aubry-Guzzi*
+
 *   Fix `increment!` / `decrement!` on models with query constraints to include
     every query constraint column in the counter update.
 
--- a/activerecord/lib/active_record/associations/foreign_association.rb
+++ b/activerecord/lib/active_record/associations/foreign_association.rb
@@ -4,7 +4,9 @@ module ActiveRecord::Associations
   module ForeignAssociation # :nodoc:
     def foreign_key_present?
       if reflection.klass.primary_key
-        owner.attribute_present?(reflection.active_record_primary_key)
+        Array(reflection.active_record_primary_key).all? do |key|
+          owner.attribute_present?(key)
+        end
       else
         false
       end
--- a/activerecord/test/cases/associations/has_many_associations_test.rb
+++ b/activerecord/test/cases/associations/has_many_associations_test.rb
@@ -2073,6 +2073,27 @@ class HasManyAssociationsTest < ActiveRecord::TestCase
     assert great_author.books.include?(book)
   end
 
+  def test_collection_for_new_record_owner_with_composite_primary_key_present
+    book = Cpk::Book.create!(id: [1, 10], title: "Some book")
+    chapter = book.chapters.create!(id: [1, 100], title: "Some chapter")
+
+    new_book = Cpk::Book.new(id: [1, 10])
+    assert_predicate new_book, :new_record?
+
+    assert_equal [chapter], new_book.chapters.to_a
+  end
+
+  def test_collection_for_new_record_owner_with_composite_primary_key_missing
+    book = Cpk::Book.create!(id: [1, 10], title: "Some book")
+    book.chapters.create!(id: [1, 100], title: "Some chapter")
+
+    new_book = Cpk::Book.new(author_id: 1)
+    assert_predicate new_book, :new_record?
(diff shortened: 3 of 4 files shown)


tylerwood-io:
Clean fix for a clear composite primary key parity gap. The root cause analysis in the description is accurate — `attribute_present?` receives the composite key array (e.g., `["author_id", "id"]`), stringifies it via the default `to_s`, finds no column by that name, and silently returns `false`. This makes `null_scope?` always true for unsaved CPK owners, preventing the association query from ever firing.

### The fix

```ruby
Array(reflection.active_record_primary_key).all? do |key|
  owner.attribute_present?(key)
end
```

`Array()` is idiomatic for handling both scalar and array inputs, and `.all?` correctly requires every key column to be present before allowing the query. For single-key associations, `Array("id")` produces `["id"]`, so behavior is unchanged — no branching needed.

### Edge case analysis

1. **Partial key population** (e.g., `Cpk::Book.new(author_id: 1)` with `id` unset): `.all?` returns false → `null_scope?` returns true → association returns empty/nil. Correct — a partial composite key can't form a valid WHERE clause. The tests cover this explicitly.

2. **Zero-valued key columns**: `attribute_present?` returns true for `0` (integer), so `Cpk::Book.new(id: [0, 1])` would correctly fire the query. Consistent with single-key behavior where `Post.new(id: 0).comments` also queries.

3. **Empty string in key columns**: `attribute_present?` returns false for `""`, so a string-typed PK column set to empty string would treat the key as absent. Also consistent with single-key behavior.

4. **`has_many` and `has_one` coverage**: `foreign_key_present?` lives in `ForeignAssociation`, which is included by both `HasManyAssociation` (via `CollectionAssociation#null_scope?`) and `HasOneAssociation` (via `Association#find_target?`). The tests cover both association types — good.

5. **Through associations are unaffected**: `has_many :through` and `has_one :through` don't include `ForeignAssociation` — they resolve via the intermediate join model. So this fix correctly doesn't touch them.

6. **Existing `Array()` usage in the same module**: `nullified_owner_attributes` (line 14) and `set_owner_attributes` (line 25) already use `Array(reflection.foreign_key)` / `Array(reflection.join_primary_key)` for CPK support in the same file. The fix follows the established pattern.

7. **Interaction with `inverse_of`**: When an unsaved owner with populated CPK columns queries its association, returned records will have their inverse set to the unsaved owner. This is existing behavior for single-key associations (e.g., `Post.new(id: existing_id).comments` sets `comment.post` to the new record). Not new behavior — just parity.

### Test quality

Four tests covering the present/missing matrix across `has_many` and `has_one`, using existing CPK fixtures (`Cpk::Book`, `Cpk::Order`). The pattern of creating a persisted record then instantiating a new record with the same PK values clearly demonstrates the fix. CHANGELOG is present and well-written.

Looks good.

nvasilevski (approved):

PR #57652, opened 2026-06-10
Fix `distinct.count` for models with a composite primary key

augustocbx:
### Motivation / Background

`relation.distinct.count` raises `ActiveRecord::StatementInvalid` for any model
with a composite primary key. Runnable reproduction (from the bug report
templates):

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"

  create_table :orders, primary_key: ["shop_id", "id"], force: true do |t|
    t.integer "shop_id", null: false
    t.integer "id", null: false
  end
end

class Order < ActiveRecord::Base
end

  def test_distinct_count
    Order.create!(shop_id: 1, id: 1)
    assert_equal 1, Order.distinct.count
    # ActiveRecord::StatementInvalid: no such column: "shop_id", "id":
    #   SELECT COUNT(DISTINCT ["shop_id", "id"]) FROM "orders"
  end
end
```

Fixes #55401.

### Detail

**Cause:** when counting with `distinct` and `column_name == :all`,
`perform_calculation` swaps `:all` for the primary key to emit the cheaper
`COUNT(DISTINCT pk)` instead of a subquery
(`activerecord/lib/active_record/relation/calculations.rb:451`). That shortcut
predates composite primary key support: an Array primary key falls into
`aggregate_column`'s `else` branch and is stringified into a single
non-existent column name.

**Fix:** skip the single-column shortcut when the model has a composite
primary key. `column_name` then stays `:all`, and `build_count_subquery?`
routes the relation into the existing `build_count_subquery` path — whose own
comment notes that multi-column `COUNT DISTINCT` must go through a subquery.
This is the exact query shape CPK models already produce today when an
ORDER BY is present (`Order.order(:x).distinct.count` works on main). The
`model.composite_primary_key?` guard mirrors its existing uses in
`finder_methods.rb`. This also fixes `distinct.limit(n).count`, which raised
the same way through `build_count_subquery`.

Single-PK models take exactly the same code path as before.

### Additional information

- Supersedes #55432 (thanks @robzolkos for the earlier patch) — that approach
  added Array handling in `aggregate_column`/`operation_over_aggregate_column`
  plus a hand-rolled count-over-subquery; routing into the existing
  `build_count_subquery` needs only this one-line guard.
- `distinct.group(:x).count` on a CPK model also raises on main; it raises
  before and after this change and is deliberately out of scope (it needs
  separate handling in `execute_grouped_calculation`, and grouped distinct
  full-row counting is semantically `COUNT(*)` per group since primary keys
  are unique).
- Tests after rebasing onto current `main`: `test/cases/calculations_test.rb`
  on SQLite in-memory (239 runs, 664 assertions, 0 failures, 0 errors); full
  Active Record suite on SQLite (9,849 runs, 35,110 assertions, 0 failures,
  0 errors); RuboCop on both changed Ruby files (no offenses). MySQL adapters
  were not available locally; CI covers those adapters.

### Checklist


activerecord/test/cases/calculations_test.rb
```diff
@@ -433,6 +433,17 @@ def test_count_for_a_composite_primary_key_model
     assert_equal(1, Cpk::Book.where(author_id: book.author_id, id: book.id).count)
   end
 
+  def test_distinct_count_for_a_composite_primary_key_model
+    assert_equal(Cpk::Book.count, Cpk::Book.distinct.count)
+
+    book = cpk_books(:cpk_great_author_first_book)
+    assert_equal(1, Cpk::Book.where(author_id: book.author_id, id: book.id).distinct.count)
```
akostadinov, line 440: How about if operation is not `count` but something else? Would it matter?

augustocbx: This change only runs inside `if operation == "count"` in `perform_calculation`, so `sum`, `average`, `minimum`, and `maximum` keep their existing paths. The composite-primary-key guard only changes how `distinct.count` chooses its count column. I rebased onto current `main` and reran the calculations tests and full Active Record SQLite suite; both pass.

PR #57656, opened 2026-06-10
Support composite primary keys in `excluding`

55728:
### Summary

`Relation#excluding` (and its alias `without`) raises `ActiveRecord::StatementInvalid` on a model with a composite primary key, whereas it works for a single primary key:

```ruby
# Single primary key — works
Post.excluding(post)

# Composite primary key — raises
Cpk::Book.excluding(book)
# => ActiveRecord::StatementInvalid: no such column: cpk_books.["author_id", "id"]
```

### Cause

`excluding!` builds the exclusion predicate with `predicate_builder[primary_key, records]`. For a composite primary key `primary_key` is an array (e.g. `["author_id", "id"]`), which is coerced into a single, bogus column name, producing SQL that references a non-existent column like `"cpk_books"."[""author_id"", ""id""]"`.

### Fix

For composite primary keys, build the predicate through `build_where_clause(primary_key => ids)` and invert it, mirroring how the finders already match composite keys with `where(primary_key => ids)`. The single primary key path is left byte-for-byte unchanged, so there is no behavior change for the common case.

Both the `excluding(record)` and `excluding(relation)` forms (the latter goes through `relation.ids`) are covered.

### DIFF
--- a/activerecord/lib/active_record/relation/query_methods.rb
+++ b/activerecord/lib/active_record/relation/query_methods.rb
@@ -1656,8 +1656,8 @@ module ActiveRecord
     alias :without :excluding
 
     def excluding!(records) # :nodoc:
-      predicates = [ predicate_builder[primary_key, records].invert ]
-      self.where_clause += Relation::WhereClause.new(predicates)
+      ids = records.map { |record| record.is_a?(model) ? record.id : record }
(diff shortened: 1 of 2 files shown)


fatkodima:
Refactored the implementation to make it simpler.

55728:
Thank you for the review and refactoring, as always! 😊

PR #57660, opened 2026-06-10
Fix `normalizes` casting raw database values on in-place change detection

chaadow:
cc @rafaelfranca  @yaroslav 

### Motivation / Background

This Pull Request has been created because #57639 introduced a regression for
attributes built from the database when a `normalizes` attribute is changed in
place.

`normalized_attribute_changed_in_place?` compares the current value against
`attribute.type.cast(attribute.value_before_type_cast)`. For attributes built
from the database, `value_before_type_cast` holds the raw serialized value,
which must be deserialized rather than cast. For example, a json attribute
holds a raw JSON string, and `Type::Json#cast` does not parse strings, so the
raw string was passed straight into the user-supplied normalizer block:

```ruby
  # settings is a json column
  normalizes :settings, with: -> settings { settings.transform_keys(&:downcase) }
end

user = User.find(User.create!(settings: { "foo" => "bar" }).id)
user.settings["BAZ"] = "qux" # change the attribute in place
user.valid?
# => NoMethodError: undefined method 'transform_keys' for an instance of String
```

The same error occurs for new records when the column has a database default
(e.g. a jsonb column with a default of `{}`), since default attributes are
also built from the raw database value.

### Detail

This Pull Request changes `normalized_attribute_changed_in_place?` to use
`Attribute#type_cast` instead of `attribute.type.cast`. `Attribute#type_cast`
dispatches to `type.deserialize` for attributes coming from the database and to
`type.cast` for attributes coming from the user, leaving the existing behavior
for user-assigned values unchanged.

### Checklist

### DIFF
--- a/activemodel/lib/active_model/attributes/normalization.rb
+++ b/activemodel/lib/active_model/attributes/normalization.rb
@@ -145,7 +145,7 @@ module ActiveModel
         end
 
         def normalized_attribute_changed_in_place?(attribute)
-          attribute.changed_in_place? && attribute.value != attribute.type.cast(attribute.value_before_type_cast)
+          attribute.changed_in_place? && attribute.value != attribute.type_cast(attribute.value_before_type_cast)
         end
 
         class NormalizedValueType < ActiveSupport::Delegation::DelegateClass(ActiveModel::Type::Value) # :nodoc:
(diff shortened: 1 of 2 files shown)


yaroslav:
@chaadow thank you for finding this

PR #57662, opened 2026-06-10
Add ActiveSupport::Ractors.on_main to proxy work to the main Ractor

gmcgibbon:
### Motivation / Background

This Pull Request has been created because non-main Ractors need a way to delegate assignment of variables to the main Ractor. API usage will refactor something like:
```rb
class Something
  def self.something
     @ivar ||= :a
  end
end
```
To this:
```rb
class Something
  def self.something
     @ivar || ActiveSupport::Ractors.on_main { @ivar ||= :a }
  end
end
```
So that non-main Ractors may be lazily load values when it is necessary. I'm trying to remove some class level memoization in Active Record in https://github.com/rails/rails/pull/57642, but sometimes memos are needed, and we can memoize safely with this Ractor proxying API.

Alternatively, we discussed implementing a autoloader hook that would call a method like:

```rb
def ractorize!
  some_memo_method
  other_memo_method
end
```

which would use eager-load semantics to prepare an object for non-main Ractor use. This was dismissed as we didn't want an approach that coupled us to the autoloader.

### Detail

This Pull Request adds `.on_main` to `ActiveSupport::Ractors` which uses https://github.com/jhawthorn/ractor-dispatch to do the proxying. We may want to look at upstreaming this feature in Active Support in the future, if this patch is accepted.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/ractors.rb
+++ b/activesupport/lib/active_support/ractors.rb
@@ -68,6 +68,20 @@ module ActiveSupport
       end
 
       if defined?(Ractor) && RUBY_VERSION >= "4.0"
+        def on_main(obj = nil, &block)
+          if Ractor.main?
+            obj.instance_eval(&block)
+          else
+            require "ractor/dispatch"
+            shareable_block = Ractor.shareable_proc(&block)
+            Ractor::Dispatch.main.run { obj.instance_eval(&shareable_block) }
+          end
+        end
+
+        def main?
+          Ractor.main?
+        end
+
         # Makes +obj+ Ractor-shareable by delegating to +Ractor.make_shareable+.
         #
         # The +copy:+ option is forwarded unchanged. On Ruby versions without
@@ -104,6 +118,14 @@ module ActiveSupport
           Ractor.shareable_lambda(...)
         end
       else
+        def main?
+          Ractor.current == Ractor.main
+        end
+
+        def on_main(obj = nil, &block)
+          obj.instance_eval(&block)
+        end
+
         def make_shareable(obj, copy: false)
           obj
         end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -71,6 +71,7 @@ PATH
       logger (>= 1.4.2)
       minitest (>= 5.1)
       psych (>= 4)
+      ractor-dispatch (>= 0.2.0)
       securerandom (>= 0.3)
       tzinfo (~> 2.0, >= 2.0.5)
       uri (>= 0.13.1)
@@ -431,6 +432,7 @@ GEM
       rack (>= 1.3)
     rackup (2.2.1)
       rack (>= 3)
+    ractor-dispatch (0.2.0)
     rails-dom-testing (2.3.0)
       activesupport (>= 5.0.0)
       minitest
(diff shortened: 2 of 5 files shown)


gmcgibbon:
Looks like adding the dependency is tricker than I thought. I'll place this into draft for now.

activesupport/lib/active_support/ractors.rb
```diff
@@ -6,6 +6,18 @@ module ActiveSupport
   module Ractors # :nodoc:
     class << self
       if defined?(Ractor) && RUBY_VERSION >= "4.0"
+        gem "ractor-dispatch"
+        require "ractor/dispatch"
```
byroot: This would break all users already on 4.0. You need an extra runtime flag to enable this, or try to require the gem and handle it being missing.

byroot:
So I've been told about this by Ufuk a few weeks ago, so I'll share my opinion here.

I understand that this is intended as an escape hatch for the few pieces that can't be made Ractor safe (yet?).
But it really makes me uncomfortable to have cross ractor communication like this sprinkled around the codebase, as you really don't know what the main ractor is doing, hence how long it might take for it to execute your snippet, etc.

That's why in my opinion it would have been more pertinent to start from the opposite side, as in progressively make lower level APIs like `Logger`, `ActiveSupport::Cache`, `ActiveRecord::ConnectionAdapters`, etc ractor safe. So that even if the whole request isn't in a Ractor, you're still getting benefits from what has been made ractor-safe, without having to rely on hacks like this one to close the gaps.

matthewd:
> you really don't know what the main ractor is doing, hence how long it might take for it to execute your snippet

While this is true, there's also no getting around the fact that certain things can only be done on the main ractor (for various reasons, ranging from maybe-temporary lib compat issues, to fundamental properties of the runtime). Without an ability to send work back to the main ractor, you just can't shift work off main to begin with... so you're always in a worse position because you're still reliant on the main ractor to do this work... you're just _also_ relying on it doing the thing that would otherwise use this method, as well.

We're obviously not going to spray direct calls to this all over the place, but if we want to have an indirection that implements a cross-ractor assignable variable as in the OP example, we need this primitive to build that `crattr_accessor`-or-whatever on top of.

byroot:
> so you're always in a worse position

That really depend on what your goal is. To me, having this sort of unexpected/unpredicatble bottlenecking on the main ractor makes for a very wonky architecture with hard to predict performance, so is a worse position.

> We're obviously not going to spray direct calls to this all over the place

It's really not that obvious to me.

activesupport/lib/active_support/ractors.rb
gmcgibbon: We lowered the version constraint for ractor-dispatch, and now it should work properly. I've added it as a dependency to the Active Support gemspec instead of a peer dependency. 

We will likely vendor this code eventually.

gmcgibbon:
Any objections with the new API itself? I added a usage commit to see how it looks.

byroot:
Yeah, I still think it's absolutely awful, but if everyone else is convinced 🤷

gmcgibbon:
We could make a core extension that might help readability a bit, or use the autoloading approach I pitched above (or something else?). I think this is a pretty big decision for the framework, so I'll hold off merging for more comments.

byroot:
> We could make a core extension that might help readability a bit

My bad, my objection is with the architecture, not the API.

activesupport/lib/active_support/ractors.rb
```diff
@@ -36,6 +36,21 @@ def try_shareable_proc(proc = nil, &block)
...
+            obj.instance_eval(&block)
+          else
+            shareable_block = Ractor.shareable_proc(&block)
+            Ractor::Dispatch.main.run { obj.instance_eval(&shareable_block) }
+          end
+        end
+
+        def main?
+          Ractor.main?
+        end
```
Edouard-chin, line 83: ~I'm missing the point of adding this ?~

~`Ractor.main?` works on all supported Ruby that Rails supports and is faster to type than `Activesupport::Ractors.main?`~

Edit: Ignore me, I forgot that Ruby 3.3.x is still a version Rails supports

activesupport/lib/active_support/ractors.rb
```diff
@@ -104,6 +119,14 @@ def shareable_lambda(...)
           Ractor.shareable_lambda(...)
         end
       else
+        def main?
+          true
```
Edouard-chin: Maybe `Ractor.main == Ractor.current`  ?

gmcgibbon: Good idea, folks can still uses Ractor on 3.3, so why not.

tenderlove (approved):
I'm going to merge this. If it turns out to be the wrong direction, then I will do the work to revert it.

PR #57677, opened 2026-06-11
Fix `Array#to_query` to skip empty Array elements, mirroring the empty-Hash case

55728:
## Problem

`Array#to_query` was recently updated (68160c4417) to drop nested elements whose serialization is empty, so that an empty Hash element is omitted from the query string. That filtering relies on the resulting string being empty (`q unless q.empty?`). An empty *Array* element, however, does not produce an empty string: `[].to_query(prefix)` falls through to `nil.to_query(prefix)`, which returns the bare escaped key (non-empty). As a result an empty Array element leaks a stray valueless `a[][]` segment (no `=`), inconsistent with how an empty Hash element is handled. Anyone serializing params that contain a nested empty array (e.g. `{ a: [1, [], 2] }.to_query` from form/URL helpers) hits this.

## Reproduction

```ruby
require "active_support/core_ext/object/to_query"

[1, [], 2].to_query("a")
# actual:   "a%5B%5D=1&a%5B%5D%5B%5D&a%5B%5D=2"   (stray "a%5B%5D%5B%5D")
# expected: "a%5B%5D=1&a%5B%5D=2"

{ a: [1, {}, 2] }.to_query  # => "a%5B%5D=1&a%5B%5D=2"  (empty Hash already dropped)
{ a: [1, [], 2] }.to_query  # => stray segment, before the fix
```

The empty-Hash element is already elided; the empty-Array element should behave the same way.

## Fix

`Array#to_query` now applies the same guard that `Hash#to_query` already uses for nested empty collections: skip an element when it is an empty Hash or Array before delegating to its `to_query`. This is a 3-line change that makes the Array path consistent with the Hash path. The top-level empty-array case is unchanged: `[].to_query("a")` still returns `"a%5B%5D"`, because that case is handled by the separate `if empty?` branch.

## Test

Added `test_array_with_empty_array` in `activesupport/test/core_ext/object/to_query_test.rb`, paralleling the existing `test_array_with_empty_hash`: it asserts `{ a: [1, [], 2] }.to_query == "a%5B%5D=1&a%5B%5D=2"`. The test fails before the change (the stray `a%5B%5D%5B%5D` segment appears) and passes after.

### DIFF
--- a/activesupport/lib/active_support/core_ext/object/to_query.rb
+++ b/activesupport/lib/active_support/core_ext/object/to_query.rb
@@ -60,8 +60,8 @@ class Array
       nil.to_query(prefix)
     else
       filter_map { |value|
-        q = value.to_query(prefix)
-        q unless q.empty?
+        next if (value.is_a?(Hash) || value.is_a?(Array)) && value.empty?
+        value.to_query(prefix)
       }.join "&"
     end
   end
@@ -83,9 +83,8 @@ class Hash
   # are sorted lexicographically in ascending order.
   def to_query(namespace = nil)
     query = filter_map do |key, value|
-      unless (value.is_a?(Hash) || value.is_a?(Array)) && value.empty?
-        value.to_query(namespace ? "#{namespace}[#{key}]" : key)
-      end
+      next if (value.is_a?(Hash) || value.is_a?(Array)) && value.empty?
+      value.to_query(namespace ? "#{namespace}[#{key}]" : key)
     end
 
     query.sort! unless namespace.to_s.include?("[]")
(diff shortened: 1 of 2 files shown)


activesupport/lib/active_support/core_ext/object/to_query.rb
```diff
@@ -60,8 +60,9 @@ def to_query(key)
       nil.to_query(prefix)
     else
       filter_map { |value|
-        q = value.to_query(prefix)
-        q unless q.empty?
+        unless (value.is_a?(Hash) || value.is_a?(Array)) && value.empty?
```
byroot: `unless` with multiple condition is always very hard to understand. This would be better rewritten as an `if` condition or using a guard clause.

55728: Thanks for the review! 👀✨ You're right that the negated `unless` with a compound condition is hard to read. I've rewritten it as an early-return guard clause (`next if ...`) so the skip condition reads positively, and applied the same change to `Hash#to_query` so both stay consistent. Behavior is unchanged and the existing tests still pass.

```ruby
filter_map { |value|
  next if (value.is_a?(Hash) || value.is_a?(Array)) && value.empty?
  value.to_query(prefix)
}.join "&"
```

PR #57680, opened 2026-06-11
Ensure `assert_no_changes` isn't given a static value

amomchilov:
### Motivation / Background

There's a sharp edge to `assert_no_changes` (and friends): if you accidentally give them a "static" value (like an Integer), they will always pass, even if a change was made. E.g.

```ruby
#!/usr/bin/ruby

require "active_support"
require "active_support/test_case"
require "minitest/autorun"

class AssertNoChangesTest < ActiveSupport::TestCase
  def test_value_changed_but_test_passes
    a = []
    assert_no_changes(a.size) do
      a << "something"
    end
  end
end
```

I should have passed `-> { a.size }` or `"a.size"` (which will be `eval`-ed), but the test silently passes anyway.

Rails should help catch these scenarios.

### Detail

This Pull Request changes `assert_no_changes` to raise `ArgumentError` unless the expression their given is a callable (typically a Proc), String, or Symbol.

#### Correctness

This change isn't totally fool-proof. If the expression you're observing is a `String` (or Symbol), then it'll be treated as Ruby code and evaled. There's no way for us to tell apart a static string expression that should have been wrapped in a proc, from a string that's intended to be evaluated as Ruby source.

That's less likely to fail silently though (your string is mostly likely not semantically valid Ruby), so it's fine. e.g.

```ruby
oncall = "Alice"
assert_no_changes(oncall) do # NameError: uninitialized constant AssertNoChangesTest::Alice
  oncall = "Bob"
end
```

#### Backwards compatibility

This will now raise `ArgumentError` in the rare cases where people were passing a non-String, non-Symbol value that responds to `to_s`, that also happened to eval'ed correctly. This is quite unlikely:

```ruby
class Hmm
  def to_s = "x" # evals correctly in the context below
end

def test_hmm
  x = 1
  assert_no_changes(Hmm.new) do # passes correctly
    # no change
  end
end
```

If it's intended, there's a trivial fix of calling `to_s` on the argument to `assert_no_changes`.

### Checklist

### DIFF
--- a/activesupport/test/test_case_test.rb
+++ b/activesupport/test/test_case_test.rb
@@ -82,6 +82,15 @@ class AssertionsTest < ActiveSupport::TestCase
     end
   end
 
+  def test_assert_no_difference_with_non_callable
+    error = assert_raises ArgumentError do
+      assert_no_difference @object.num do # Should have been `-> { @object.num }`!
+        @object.increment
+      end
+    end
+    assert_match("The expression must be a callable object like a Proc, or a String of Ruby code. Got 0", error.message)
+  end
+
   def test_assert_difference
     assert_difference "@object.num", +1 do
       @object.increment
@@ -178,6 +187,15 @@ class AssertionsTest < ActiveSupport::TestCase
       end
     end
     assert_equal "Object Changed.\n`@object.num` didn't change by 1, but by 0.\nExpected: 1\n  Actual: 0", error.message
+  end
+
+  def test_assert_difference_with_non_callable
+    error = assert_raises ArgumentError do
+      assert_difference @object.num, +1 do # Should have been `-> { @object.num }`!
+        @object.increment
+      end
+    end
+    assert_match("The expression must be a callable object like a Proc, or a String of Ruby code. Got 0", error.message)
   end
 
   def test_hash_of_lambda_expressions
@@ -315,6 +333,15 @@ class AssertionsTest < ActiveSupport::TestCase
     assert_changes -> { token }, from: /\w{32}/, to: /\w{32}/ do
       token = SecureRandom.hex
     end
+  end
+
+  def test_assert_changes_with_non_callable
+    error = assert_raises ArgumentError do
+      assert_changes @object.num, from: 0, to: 1 do # Should have been `-> { @object.num }`
+        @object.increment
+      end
+    end
+    assert_match("The expression must be a callable object like a Proc, or a String of Ruby code. Got 0", error.message)
   end
 
   def test_assert_changes_with_message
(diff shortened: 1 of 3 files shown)


activesupport/test/test_case_test.rb
```diff
@@ -82,6 +82,15 @@ def test_assert_no_difference_with_multiple_expressions_fail
     end
   end
 
+  def test_assert_no_difference_with_non_callable
```
amomchilov, line 85: Prior to my changes, this `assert_no_difference` call would already fail, because it would be trying to `eval` an Integer (because unlike the others, this code path didn't call `to_s`

```
AssertionsTest#test_assert_no_difference_with_non_callable [test/test_case_test.rb:86]:
[ArgumentError] exception expected, not
Class: <TypeError>
Message: <"no implicit conversion of Integer into String">
```

activesupport/test/test_case_test.rb
```diff
@@ -180,6 +189,15 @@ def test_assert_difference_message_with_lambda
     assert_equal "Object Changed.\n`@object.num` didn't change by 1, but by 0.\nExpected: 1\n  Actual: 0", error.message
   end
 
+  def test_assert_difference_with_non_callable
```
amomchilov, line 192: Ditto, this would fail with:

```
AssertionsTest#test_assert_difference_with_non_callable [test/test_case_test.rb:193]:
[ArgumentError] exception expected, not
Class: <TypeError>
Message: <"no implicit conversion of Integer into String">
```

activesupport/test/test_case_test.rb
```diff
@@ -317,6 +335,15 @@ def test_assert_changes_with_to_and_from_and_case_operator
     end
   end
 
+  def test_assert_changes_with_non_callable
```
amomchilov, line 338: `assert_changes` is inert to this problem, because the static value couldn't change, so it would already fail anyway:

```
Failure:
AssertionsTest#test_assert_changes_with_non_callable [test/test_case_test.rb:339]:
`0` didn't change
```

This PR still improves the message, to help find the issue faster.

activesupport/test/test_case_test.rb
```diff
@@ -458,6 +485,15 @@ def check.call
     assert_match(/#<Object:0x.*changed/, error.message)
   end
 
+  def test_assert_no_changes_message_with_non_callable
```
amomchilov, line 488: Prior to my changes, this `assert_no_changes` would silently pass

```
Failure:
AssertionsTest#test_assert_no_changes_message_with_non_callable [test/test_case_test.rb:489]:
ArgumentError expected but nothing was raised.
```

amomchilov (commented):
Here's what the behaviour used to be like before my change

PR #57682, opened 2026-06-11
Ractor safe readonly attributes

andrewn617:
### Motivation / Background

We are working on making Rails ractor safe. A lot of hashes and arrays are stored in class attributes as configuration, they're mutable so we need to find a way for them to be ractor safe.

This PR focuses on `ActiveRecord::Base._attr_readonly`.

### Detail

Instead of doing the book keeping of freezing all these when we freeze the application, we are going to make them frozen by default. We already did this with exception wrapper configs in https://github.com/rails/rails/pull/57483. The default array is frozen, and then when adding new elements we dup the array, add the new element and freeze it again.

On this PR I also improved the assertion I added in https://github.com/rails/rails/pull/57574. We really need two assertions. `assert_ractor_shareable` for things like this that we expect to be shareable without us needing to make them shareable. And then `assert_ractor_make_shareable` for things like the backtrace cleaner, which we expect to be able to make shareable when freezing the application.

### Checklist

### DIFF
--- a/activerecord/lib/active_record/readonly_attributes.rb
+++ b/activerecord/lib/active_record/readonly_attributes.rb
@@ -28,7 +28,7 @@ module ActiveRecord
       #   post.title = "a different title" # raises ActiveRecord::ReadonlyAttributeError
       #   post.update(title: "a different title") # raises ActiveRecord::ReadonlyAttributeError
       def attr_readonly(*attributes)
-        self._attr_readonly |= attributes.map(&:to_s)
+        self._attr_readonly = ActiveSupport::Ractors.make_shareable(_attr_readonly | attributes.map(&:to_s))
 
         if ActiveRecord.raise_on_assign_to_attr_readonly
           include(HasReadonlyAttributes)
--- a/activesupport/test/backtrace_cleaner_test.rb
+++ b/activesupport/test/backtrace_cleaner_test.rb
@@ -140,8 +140,8 @@ class BacktraceCleanerDefaultFilterAndSilencerTest < ActiveSupport::TestCase
     assert_equal backtrace, @bc.clean(backtrace)
   end
 
-  test "backtrace cleaner is Ractor shareable" do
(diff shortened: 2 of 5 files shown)


activerecord/lib/active_record/readonly_attributes.rb
```diff
@@ -28,7 +28,7 @@ module ClassMethods
       #   post.title = "a different title" # raises ActiveRecord::ReadonlyAttributeError
       #   post.update(title: "a different title") # raises ActiveRecord::ReadonlyAttributeError
       def attr_readonly(*attributes)
-        self._attr_readonly |= attributes.map(&:to_s)
+        self._attr_readonly = (_attr_readonly | attributes.map { |a| a.to_s.freeze }).freeze
```
gmcgibbon: This seems a bit obtuse. Can we do `self._attr_readonly = ActiveSupport::Ractors.make_shareable(_attr_readonly |  attributes.map(&:to_s))` instead?

This is essentially a deep freeze.

activesupport/test/backtrace_cleaner_test.rb
```diff
@@ -140,8 +140,8 @@ def setup
     assert_equal backtrace, @bc.clean(backtrace)
   end
 
-  test "backtrace cleaner is Ractor shareable" do
-    assert_ractor_shareable @bc
```
gmcgibbon, line 144: Should we remove this helper if it isn't needed anymore?

activerecord/lib/active_record/readonly_attributes.rb
andrewn617: Sure

activesupport/test/backtrace_cleaner_test.rb
andrewn617: I am using it in the new tests for read only attr

gmcgibbon (approved):

PR #57684, opened 2026-06-11
Add a Ractor-shareable ActiveSupport::TaggedLogging.shareable_logger

etiennebarrie:
This mirrors `ActiveSupport::TaggedLogging.logger`, but returns a logger that can be made shareable and used from multiple Ractors.

The logger is an ActiveSupport::ShareableLogger (a ::Logger subclass whose device proxies writes to a background ShareableLogger::Writer), wrapped with tagged logging.

It reproduces the logger's behavior if there's a failure while writing: writes to stderr, and doesn't cause an error in the caller. (I looked into using the error reporter, but since it itself falls back to the logger, it was a bit too circular).

There's some risk about unbounded queue growth from the Ractor port, but it's mostly mitigated by enforcing the contract that at the end of the request, the logger should be flushed. An endpoint logging a lot could end up using more memory than without the ractor logger still. Still, I have a [follow-up branch](https://github.com/Shopify/rails/compare/activesupport-shareable-logger...shareable-logger-backpressure?expand=1) with an optional threshold that, once reached, causes the logger to become synchronous, but it requires the `ractor_safe` gem, so I feel like it's better to keep it optional.

For now, it's undocumented, but it should be usable from normal threaded Rails applications, albeit with unclear performance gains, if any. But once we have this, we should be able to disable the mutex in `Logger::LogDevice` for example, since only the Writer thread writes to the logdev.

### DIFF
--- a/activesupport/lib/active_support/tagged_logging.rb
+++ b/activesupport/lib/active_support/tagged_logging.rb
@@ -123,6 +123,15 @@ module ActiveSupport
       new ActiveSupport::Logger.new(*args, **kwargs)
     end
 
+    # Returns a logger that can be used from Ractors. Accepts the same arguments
+    # as Logger.new. The returned logger is an ActiveSupport::Ractors::Logger
+    # (a ::Logger subclass whose device proxies writes to a background Writer)
+    # wrapped with tagged logging, so it can be made shareable and used from any
+    # Ractor.
+    def self.ractor_logger(*args, **kwargs) # :nodoc:
+      new(ActiveSupport::Ractors::Logger.new(*args, **kwargs))
+    end
+
     def self.new(logger)
       logger = logger.clone
 
--- a/activesupport/test/ractor_logger_test.rb
+++ b/activesupport/test/ractor_logger_test.rb
@@ -1,0 +1,261 @@
+# frozen_string_literal: true
+
+require_relative "abstract_unit"
+require "fileutils"
+require "timeout"
+
+class RactorLoggerTest < ActiveSupport::TestCase
+  if RUBY_VERSION >= "4.0"
+    include ActiveSupport::Testing::Isolation
+
+    test ".ractor_logger writes through a writer" do
+      path = log_path("basic.log")
+      logger = ActiveSupport::TaggedLogging.ractor_logger(path)
+
+      logger.info("hello")
+      logger.flush
+
+      assert_includes File.read(path), "hello"
+    ensure
+      logger&.close
+    end
+
+    test ".ractor_logger returns a ::Logger subclass wrapped with tagged logging" do
+      path = log_path("subclass.log")
+      logger = ActiveSupport::TaggedLogging.ractor_logger(path)
+
+      assert_kind_of ActiveSupport::Ractors::Logger, logger
+      assert_kind_of ::Logger, logger
+    ensure
+      logger&.close
+    end
+
+    test "consumer preserves Logger::LogDevice rotation" do
+      path = log_path("rotation.log")
+      # Rotate after a tiny size so a few writes trigger a roll-over.
+      logger = ActiveSupport::Ractors::Logger.new(path, 5, 64)
+
+      20.times { |i| logger.info("rotation-message-#{i}") }
+      logger.flush
+
+      assert_not_empty Dir["#{path}.*"], "expected Logger::LogDevice to rotate the log file"
+    ensure
+      logger&.close
+    end
+
+    test "reopen refreshes the device after the file is rotated externally" do
+      path = log_path("reopen.log")
+      logger = ActiveSupport::TaggedLogging.ractor_logger(path)
+
+      logger.info("before")
+      logger.flush
+
+      rotated = "#{path}.1"
+      File.rename(path, rotated) # an external tool moves the file, like logrotate
+
+      logger.reopen # reopen the same path -> a fresh file
+      logger.info("after")
+      logger.flush
+
+      assert_includes File.read(rotated), "before"
+      assert_includes File.read(path), "after"
+      assert_not_includes File.read(path), "before"
+    ensure
+      logger&.close
+    end
+
+    test "uses Active Support tagged logging formatter for tags" do
+      path = log_path("tags.log")
+      logger = ActiveSupport::TaggedLogging.ractor_logger(path)
+
+      logger.tagged("request-id") { logger.info("hello") }
+      logger.flush
+
+      assert_includes File.read(path), "[request-id] hello"
+    ensure
+      logger&.close
+    end
+
+    test "uses LocalTagStorage for tagged logger without block" do
+      path = log_path("local_tag_storage.log")
+      logger = ActiveSupport::TaggedLogging.ractor_logger(path)
+
+      tagged_logger = logger.tagged("job")
+      tagged_logger.info("performed")
+      logger.info("plain")
+      logger.flush
+
+      contents = File.read(path)
+      assert_includes contents, "[job] performed"
+      assert_includes contents, "plain"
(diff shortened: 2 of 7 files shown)


activesupport/lib/active_support/shareable_logger/writer.rb
```diff
@@ -0,0 +1,141 @@
...
+
+module ActiveSupport
+  class ShareableLogger # :nodoc:
+    class Writer # :nodoc:
+      def self.spawn(...)
+        new.spawn(...)
+      end
+
+      def initialize
+        unless defined?(::Ractor::Port)
```
gmcgibbon: Do we expect this to run in Ruby 3.x? This is private API, so we may not need this.

railties/test/application/configuration_test.rb
```diff
@@ -2043,6 +2043,27 @@ def index
       assert_includes(Rails.logger.broadcasts, logger)
     end
 
+    test "config.logger can be a tagged shareable logger" do
+      skip "Ractor::Port is unavailable" unless defined?(::Ractor::Port)
```
gmcgibbon: We're been conditionalizing with `RUBY_VERSION` instead of doing this. You can conditionally define the test based on the version of Ruby instead.

activesupport/lib/active_support/shareable_logger/device_proxy.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "active_support/shareable_logger/writer"
+
+module ActiveSupport
+  class ShareableLogger # :nodoc:
```
gmcgibbon:
```suggestion
  class RactorShareableLogger # :nodoc:
```
I think it makes sense to qualify that this is related to Ractors. Unless you want to embed it in `ActiveSupport::Ractors` instead.

activesupport/lib/active_support/tagged_logging.rb
```diff
@@ -123,6 +123,17 @@ def self.logger(*args, **kwargs)
       new ActiveSupport::Logger.new(*args, **kwargs)
     end
 
+    # Returns a logger that can be used from Ractors. Accepts the same arguments
+    # as Logger.new. The returned logger is an ActiveSupport::ShareableLogger
+    # (a ::Logger subclass whose device proxies writes to a background Writer)
+    # wrapped with tagged logging, so it can be made shareable and used from any
+    # Ractor.
+    def self.shareable_logger(*args, **kwargs) # :nodoc:
+      require "active_support/shareable_logger"
```
gmcgibbon: We can use autoload instead. I believe regular loggers aren't autoloaded because we expect them to be used.

railties/test/application/configuration_test.rb
```diff
@@ -2043,6 +2043,27 @@ def index
...
+
+      add_to_config <<~RUBY
+        require "active_support/tagged_logging"
```
gmcgibbon:
```suggestion
```
Required or autoloaded in the shareable_logger call

activesupport/lib/active_support/shareable_logger/writer.rb
gmcgibbon: Ah, I see we have to opt into the logger right now. I still feel like we may need a configuration option at some point to turn on Ractor features, in which case we can assume Ractor logging instead of having to explicitly set `config.logger`. Perhaps up for debate. This is fine for now though. 👍

activesupport/lib/active_support/shareable_logger/device_proxy.rb
etiennebarrie: Open to different names but I found "ractor shareable" to be a mouthful, and "shareable" isn't overloaded so I felt like this could work.
`ActiveSupport::Ractors::ShareableLogger` or even `ActiveSupport::Ractors::Logger` would be fine too, but I felt like it was distinct enough in the sense that it can be used without ractorizing the whole application.

etiennebarrie:
It crashes with Ruby 4.0.5 but the fix has been backported to the 4.0.x branch so it should be fixed in 4.0.6.

railties/test/application/configuration_test.rb
etiennebarrie: Set the condition to the next release of Ruby, which means CI should fail because it skips tests for now, until 4.0.6 ships with the fixes for the crash.
For Ruby <= 4.0, the tests are not defined, that way [they don't fail CI](https://github.com/rails/rails/blob/11b1d3543eed44635b2c87d86ffc9ae0d950c26b/tools/test_common.rb#L13).

activesupport/lib/active_support/shareable_logger/device_proxy.rb
gmcgibbon: All of our naming so far has been prefixed to make it more easily associated with Ractors. `ActiveSupport::Ractors.make_shareable`, `assert_ractor_shareable`, etc. I'm ok with `ActiveSupport::Ractors::Logger`. Let's go with that one.

activesupport/lib/active_support/shareable_logger/device_proxy.rb
```diff
@@ -0,0 +1,40 @@
...
+    class DeviceProxy # :nodoc:
+      def initialize(*args, **logdev_options)
+        @writer = Writer.spawn(*args, **logdev_options)
+        @closed = false
+      end
+
+      # Logger::LogDevice#write contract: returns the number of bytes written.
```
gmcgibbon:
```suggestion
```
This seems implied by the code

activesupport/test/ractor_logger_test.rb
```diff
@@ -0,0 +1,272 @@
...
+require "fileutils"
+require "timeout"
+
+class ShareableLoggerTest < ActiveSupport::TestCase
+  if RUBY_VERSION >= "4.0"
+    include ActiveSupport::Testing::Isolation
+
+    setup do
+      @original_experimental_warning = Warning[:experimental]
+      Warning[:experimental] = false
```
gmcgibbon: This shouldn't be needed. We ignore Ractor warnings on CI

activesupport/lib/active_support/tagged_logging.rb
```diff
@@ -123,6 +123,15 @@ def self.logger(*args, **kwargs)
       new ActiveSupport::Logger.new(*args, **kwargs)
     end
 
+    # Returns a logger that can be used from Ractors. Accepts the same arguments
+    # as Logger.new. The returned logger is an ActiveSupport::ShareableLogger
+    # (a ::Logger subclass whose device proxies writes to a background Writer)
+    # wrapped with tagged logging, so it can be made shareable and used from any
+    # Ractor.
+    def self.shareable_logger(*args, **kwargs) # :nodoc:
```
gmcgibbon:
```suggestion
    def self.ractor_logger(*args, **kwargs) # :nodoc:
```
If we rename the class, this should be renamed too.

gmcgibbon (approved):
👍 We will make this public API at a later date. Also, I don't think can be merged until a new Ruby is released that fixes the crash.

activesupport/test/ractor_logger_test.rb
etiennebarrie: tools/strict_warnings.rb raises on warnings, and I can't find code to ignore them.

This is the first PR using `Ractor.new`, other Ractor APIs don't raise a warning.

etiennebarrie:
> Also, I don't think can be merged until a new Ruby is released that fixes the crash.

I've used `ActiveSupport::Testing::Isolation` and there's no crash anymore.

gmcgibbon:
Looks like there's a Ractor warning on CI that should be fixed if you rebase.

gmcgibbon (approved):

PR #57687, opened 2026-06-12
Assortment of CI/test fixes

matthewd:
One `lib` change, to `report`-and-continue for exceptions that occur while disconnecting an AR pool -- which _is_ supported framework behaviour/API, but I think our own test suite is likely one of the biggest callers.

### DIFF
--- a/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
+++ b/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
@@ -521,9 +521,15 @@ module ActiveRecord
               @connections.each do |conn|
                 if conn.in_use?
                   conn.steal!
-                  checkin conn
+
+                  ActiveSupport.error_reporter.handle(source: "active_record.connection_pool") do
+                    checkin conn
+                  end
                 end
-                conn.disconnect!
+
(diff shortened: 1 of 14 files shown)


activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb
```diff
@@ -521,9 +521,18 @@ def disconnect(raise_on_acquisition_timeout = true)
               @connections.each do |conn|
                 if conn.in_use?
                   conn.steal!
-                  checkin conn
+                  begin
+                    checkin conn
+                  rescue => error
+                    ActiveSupport.error_reporter.report(error, handled: true, source: "active_record.connection_pool")
+                  end
```
byroot: nit:
```suggestion
                  ActiveSupport.error_reporter.handle(source: "active_record.connection_pool") do
                    checkin conn
                  end
```

But perhaps you want to avoid the extra frames?

matthewd: I do sometimes default to avoiding frames in places where it really doesn't matter; this is a good example where I should definitely just do the shorter/cleaner thing using the available API. Changed, thanks!

PR #57690, opened 2026-06-12
Keep the Redis listener alive when the last channel is unsubscribed

55728:
### Problem

The redis-client rewrite (`ef812c2652`) drives the listen loop to `break` (and nils `@subscribed_client`) whenever an unsubscribe brings the Redis subscription count to zero — a silent production outage.

The previous redis-gem implementation held a permanent subscription to `_action_cable_internal`, so the count only reached zero on an explicit shutdown. The rewrite dropped that sentinel, so removing the last user channel now kills the listener thread — on any process where every channel is transiently unsubscribed (low traffic, single-channel apps, nodes draining during deploys).

`ensure_listener_running` guards with `@thread ||= Thread.new ...`, so the dead thread is never replaced. Every later `subscribe` queues into `@when_connected` forever, `@subscribed_client` stays nil, and all broadcasts are silently dropped. Reconnecting clients are accepted but never wired to Redis.

### Fix

Restore the sentinel: subscribe to `_action_cable_internal` in `listen`, right after `@subscribed_client` is set. Normal channel removal leaves the count at 1; only an explicit shutdown (which unsubscribes from everything) drives it to 0 and ends the loop. The sentinel has no subscribers in the `SubscriberMap`, so messages on it are a no-op.

Added a regression test with two sequential subscriptions on the same adapter — the first unsubscription drives the count to 0, and the second verifies the listener is still alive. Full adapter suite passes (27 runs, 0 failures).

### DIFF
--- a/actioncable/lib/action_cable/subscription_adapter/redis.rb
+++ b/actioncable/lib/action_cable/subscription_adapter/redis.rb
@@ -76,6 +76,13 @@ module ActionCable
         class Listener < SubscriberMap::Async
           delegate :logger, to: :@adapter
 
+          # A permanent internal subscription keeps the Redis subscription count
+          # above zero while any user channel comes and goes, so removing the
+          # last user channel doesn't end the listen loop. Only an explicit
+          # shutdown (which unsubscribes from everything) drives the count to
+          # zero and stops the listener.
+          INTERNAL_CHANNEL = "_action_cable_internal"
+
           def initialize(adapter, config_options, executor)
             super(executor)
 
@@ -101,6 +108,8 @@ module ActionCable
 
             @reconnect_attempt = 0
             @subscribed_client = pubsub_client
+
+            pubsub_client.call("subscribe", INTERNAL_CHANNEL)
 
             until @when_connected.empty?
               @when_connected.shift.call
--- a/actioncable/test/subscription_adapter/redis_test.rb
+++ b/actioncable/test/subscription_adapter/redis_test.rb
@@ -16,6 +16,23 @@ class RedisAdapterTest < ActionCable::TestCase
     end
   end
 
+  def test_resubscribe_after_removing_the_last_channel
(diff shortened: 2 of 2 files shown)


chaadow:
@55728 Hi! you beat me to it! I was about to submit  a report + pretty much submit the same PR.  

The rewrite broke our chat application! and initially I thought it was the recent ActionCable rewrite.

you did mention this:
> ensure_listener_running guards with @thread ||= Thread.new ..., so the dead thread is never replaced. Every later subscribe queues into @when_connected forever, @subscribed_client stays nil, and all broadcasts are silently dropped. Reconnecting clients are accepted but never wired to Redis.

in my app I also prepend a module that rewrote `ensure_listener_running`: 
```ruby
 def ensure_listener_running
    if @thread && !@thread.alive?
      @thread = nil
      @reconnect_attempt = 0
    end

    super
  end
```
to properly reset the dead thread, this way next time `@thread ||=` gets called, it recreates a fresh thread

Do you think it's worth adding to this Pull request?

chaadow:
cc @byroot

55728:
Thanks for sharing your experience, @chadow 😊 — great to see independent confirmation from a production app!

I considered the thread-reset approach too, but went with the sentinel for two reasons:

1. **No broadcast gap** — resetting the thread on the next `subscribe` call leaves a window where broadcasts are silently dropped. The sentinel keeps the listener alive continuously, so there's no gap.
2. **Restoring the original design** — the previous redis-gem implementation used exactly this `_action_cable_internal` sentinel. The rewrite inadvertently dropped it, so this restores a deliberate design choice rather than adding a new mechanism.

Your root-cause analysis is spot-on though — `@thread ||=` never replacing a dead thread is the key insight 👀✨

byroot:
@chaadow feel free to open a PR and tag me for the thread restarting.

PR #57693, opened 2026-06-12
Make `ActiveRecord::Base.with` call `Object#with` when passed a block

janko:
### Motivation / Background

I have custom class accessors on my Active Record models, which I wanted to temporarily set to a value for the duration of a block. I naturally reached for the `Object#with` method, but was surprised it didn't work.

### Detail

`ActiveRecord::Base.with` currently delegates to
`ActiveRecord::Relation#with` for building CTEs, which rejects blocks. We use that to call `Object#with` when a block was passed, provided that the core extension was loaded. This is consistent with how `Enumerable#find` and `Enumerable#select` are called instead of query methods when a block is passed in.

A common use case would be temporarily enabling Active Record logging in tests:

```rb
ActiveRecord::Base.with(logger: Logger.new($stdout)) do
  # ...
end
```

### Checklist

### DIFF
--- a/activerecord/lib/active_record/querying.rb
+++ b/activerecord/lib/active_record/querying.rb
@@ -17,11 +17,23 @@ module ActiveRecord
       :and, :or, :annotate, :optimizer_hints, :extending,
       :having, :create_with, :distinct, :references, :none, :unscope, :merge, :except, :only,
       :count, :average, :minimum, :maximum, :sum, :calculate,
-      :pluck, :pick, :ids, :async_ids, :strict_loading, :excluding, :without, :with, :with_recursive,
+      :pluck, :pick, :ids, :async_ids, :strict_loading, :excluding, :without, :with_recursive,
       :async_count, :async_average, :async_minimum, :async_maximum, :async_sum, :async_pluck, :async_pick,
       :insert, :insert_all, :insert!, :insert_all!, :upsert, :upsert_all
     ].freeze # :nodoc:
     delegate(*QUERYING_METHODS, to: :all)
+
+    # Add a Common Table Expression (CTE) that you can then reference within another SELECT statement.
+    #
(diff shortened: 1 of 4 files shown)


activerecord/lib/active_record/querying.rb
```diff
@@ -17,12 +17,24 @@ module Querying
...
 
+    # Add a Common Table Expression (CTE) that you can then reference within another SELECT statement.
+    #
+    # See ActiveRecord::QueryMethods#with for more information.
+    #
+    # When given a block, and the Object#with core extension is loaded, this
+    # delegates to it instead, temporarily setting the given attributes on the
+    # class for the duration of the block and restoring them afterwards.
+    def with(...)
+      return super if block_given? && defined?(super)
```
byroot, line 34: I don't think we need the `defined?(super)`, it's not possible to load Active Record without Active Support, or am I missing something?

janko: Active Record doesn't load *all* of Active Support:

```sh
$ bundle exec ruby -r active_support/all -e "p Object.method_defined?(:with)"
true

$ bundle exec ruby -r active_record -e "p Object.method_defined?(:with)"
false
```

In this case, I thought it's better to delegate to `Relation#with` than to raise a `NoMethodError`.

byroot: Ah my bad.

In that case that's fine by me, but `Relation#with` should raise if given a block, so that it remains consistent no matter whether all of Active Support or not.

We can use the same error as Ruby 4.1 `&nil` feature so it's easily convertable in the future:

```ruby
def foo(&nil)
end
foo { 1 } #  no block accepted (ArgumentError) 
```

byroot:
Nevermind, it's already there.

byroot:
Thanks. Backported to 8.1 and 8.0

PR #57694, opened 2026-06-12
Add Arel.array_bind for IN queries as a single array bind

### LINKED ISSUE #57678: AWS RDS Proxy Connection Pinning: prepared_statements: false inlines all bind values via SubstituteBinds
### Steps to reproduce

We use AWS RDS Proxy for our AWS RDS Postgres database, and we're experiencing connection pinning from queries larger than 16kb in size. Under `prepared_statements: false`, every bind value gets inlined into the SQL text by `Arel::Collectors::SubstituteBinds` before the statement leaves Rails. For `where(id: array)`, that means every element of the array lands in the SQL string and the statement grows linearly with the array. 

I can rewrite queries so that they use `find_by_sql` with a manually-built `QueryAttribute`, but that is bypassing `ActiveRecord::Relation` and I have to find each call site and edit them. I would like to have the kind of solution that fixes it everywhere. I considered applying a monkeypatch (and I have one that I've cooked up with Claude & Codex), but my team members who are better at Ruby/Rails than I am have gently suggested that I should take this issue here.

Claude made me this script to demonstrate the problem (but it's easy to reproduce in any scenario where the size of the `IN()` is large). The script below runs with `ruby script.rb`. It needs a reachable Postgres server; libpq's standard environment variables (`PGHOST`, `PGUSER`, `PGDATABASE`, etc.) handle the rest of the connection.

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rails"
  gem "pg"
end

require "active_record"

  adapter:             "postgresql",
  database:            ENV.fetch("PGDATABASE", "postgres"),
  prepared_statements: false,
)

ActiveRecord::Schema.define { create_table(:posts, force: true) }

ids = (1..5_000).to_a
sql = Post.where(id: ids).to_sql

puts "sql bytes:                  #{sql.bytesize}"
puts "last id (#{ids.last}) inlined:  #{sql.include?(", #{ids.last})")}"
# sql bytes:                  ~31000
# last id (5000) inlined:     true
#
# 5,000 ids is already past the 16 KB statement-size threshold that RDS Proxy
# uses to decide whether to pin a backend connection.
```

### Expected behaviour

The hope is that we no longer generate mega-sized SQL statements that cause RDS connection pinning. I can imagine there are two paths forward here: extension and default behaviour change.

1. **Add something to the API to support this.** A way to mark a specific `where` bind for wire-protocol delivery — something like `Model.where(id: Arel.array_bind(ids))` — that produces `WHERE id = ANY($1::bigint[])` regardless of the `prepared_statements` setting.

2. **Default behaviour change on PostgreSQL.** Under `prepared_statements: false`, `where(id: array)` could emit `ANY($1::bigint[])` via `exec_params` instead of an inlined `IN (...)`.

I confess to not being an expert on rails (neither as a user nor a contributor) so I do not presume to say which is the right one.

### Actual behaviour

`Arel::Collectors::SubstituteBinds` inlines every bind into the SQL text. `Post.where(id: ids).to_sql` produces:

```sql
SELECT "posts".* FROM "posts" WHERE "posts"."id" IN (1, 2, ..., N)
```
This produces a very long SQL string which results in connection pinning. 

### Why this matters

RDS Proxy multiplexes a small pool of backend Postgres connections across many app processes. It [pins the backend connection](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.html#rds-proxy-pinning) for the lifetime of the session whenever any single statement's SQL text exceeds 16 KB. A pinned connection is dedicated to one app process and stops multiplexing from happening. If this happens enough, we can run out of connections.

### System configuration

**Rails version**: 8.0.5

**Ruby version**: 3.4.7

Saidbek:
### Motivation / Background

Fixes #57678

When PostgreSQL is configured with `prepared_statements: false`, Active Record inlines every bind value into the SQL text. For an array query such as `where(id: array)`, this means the statement size grows linearly with the array length: `WHERE "posts"."id" IN (1, 2, ..., N)`. With 5,000 ids this is already ~31 KB of SQL.

This is a problem behind AWS RDS Proxy, which "pins" a backend connection whenever a single statement's SQL text exceeds 16 KB, defeating connection multiplexing and risking pool exhaustion. Large `IN (...)` clauses cross that threshold easily. The same lists also push toward PostgreSQL's 65,535 bind-parameter limit and churn the plan cache (a different N produces different SQL).

### Detail

This Pull Request adds `Arel.array_bind`, an opt-in way to send an `IN`/`NOT IN` list as a single array bind parameter instead of expanding every value:

```ruby
Post.where(id: Arel.array_bind(ids))
# => SELECT "posts".* FROM "posts" WHERE "posts"."id" = ANY($1)

Post.where.not(id: Arel.array_bind(ids))
# => SELECT "posts".* FROM "posts" WHERE "posts"."id" <> ALL($1)
```

On PostgreSQL the entire list is sent as one bind (`$1`), so the SQL text is a constant size no matter how many values are matched (~58 bytes whether the list has 3 or 50,000 elements). `= ANY(array)` is exactly equivalent to `IN (...)`, and `<> ALL(array)` to `NOT IN (...)`.

Ordinary `where(id: array)` queries are completely unchanged.

### Checklist


jzarnett:
This does look like it solves the use case I reported in #57678 (though I emphasize again, I am not an expert in this!). Thanks for looking into this so quickly!

activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
```diff
@@ -82,6 +82,32 @@ def high_precision_current_timestamp
...
+
+          subtype = attribute.type_caster
+          return if subtype.nil? || subtype.is_a?(OID::Array)
+
+          # Mirror Arel::Nodes::HomogeneousIn#casted_values: values the element
+          # type can't serialize (e.g. an out-of-range integer) are dropped so
+          # they don't match. Falling back to the default expansion when nothing
+          # remains preserves the existing `IN (NULL)` behavior.
+          values = values.select { |value| subtype.serializable?(value) }
+          return if values.empty?
```
jeffreyolio, line 108: `serializable?` does not guarantee that serialization produces a non-null value. Rails considers `"bogus"` serializable for an integer column but serializes it to `nil`, so `where.not(id: Arel.array_bind([1, "bogus"]))` sends `[1, NULL]` to `<> ALL` and filters out every other row. Can we reuse the post-serialization filtering from `HomogeneousIn#casted_values` here and add a negated regression test?

activerecord/lib/active_record/statement_cache.rb
```diff
@@ -83,6 +83,7 @@ def add_bind(obj, &)
         @parts << Substitute.new
         self
       end
+      alias_method :add_bind_param, :add_bind
```
jeffreyolio: This alias loses the wire-bind behavior. `PartialQuery#sql_for` later quotes every `Substitute` into the SQL and shifts it out of the bind list, so cached array binds grow with the array again under `prepared_statements: false`. This needs a distinct retained-bind representation and a statement-cache regression test.

activerecord/lib/arel/visitors/postgresql.rb
```diff
@@ -140,6 +140,42 @@ def visit_Arel_Nodes_IsDistinctFrom(o, collector)
...
+        # never carries +nil+ values or fewer than two elements
+        # (PredicateBuilder::ArrayBindHandler splits those off beforehand), so
+        # the +ANY+/+ALL+ NULL-handling caveats do not apply here.
+        def visit_Arel_Nodes_HomogeneousIn(o, collector)
+          return super unless o.wire?
+
+          bind = @connection.build_homogeneous_in_bind(o.attribute, o.values)
+          return super if bind.nil?
+
+          collector.preparable = false
```
jeffreyolio: This query always has one placeholder, so it does not need the `HomogeneousIn` safeguard for variable bind counts. Marking it unpreparable prevents Rails from reusing the prepared statement and works against the plan-cache benefit described in the PR. Can we leave it preparable and cover that in a test?

activerecord/lib/active_record/relation/predicate_builder/array_bind_handler.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        values = value.values.map { |x| x.is_a?(Base) ? x.id : x }
+        nils = values.compact!
+        ranges = values.extract! { |v| v.is_a?(Range) }
+
+        values_predicate =
+          case values.length
+          when 0 then NullPredicate
+          when 1 then predicate_builder.build(attribute, values.first)
+          else Arel::Nodes::HomogeneousIn.new(values, attribute, :in, wire: true)
```
jeffreyolio: This handler duplicates `ArrayHandler` almost line for line, while `wire:` adds a transport mode to the generic `HomogeneousIn` node. A dedicated `HomogeneousArrayBind < HomogeneousIn` node plus a small overridable factory in `ArrayHandler` would reuse the nil and range handling and let the PostgreSQL visitor dispatch by node type instead of carrying a boolean through hashing and inversion.

activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb
Saidbek: Done, now filters with serialize + non-nil (mirroring casted_values) and adds a negated regression test.

activerecord/lib/active_record/statement_cache.rb
Saidbek: Done, `PartialQueryCollector` now has a distinct RetainedBind path so array binds stay as real parameters under `prepared_statements: false`, with a statement-cache regression test.

activerecord/lib/arel/visitors/postgresql.rb
Saidbek: Done, left the single-placeholder form preparable and covered that in a test.

activerecord/lib/active_record/relation/predicate_builder/array_bind_handler.rb
Saidbek: Done, introduced `HomogeneousArrayBind` with an `ArrayHandler` factory and dropped the wire: flag.

jzarnett:
@Saidbek Is there anything I can contribute that would help progress this? I have continued to chip away at large `IN` lists in our codebase but it is a little bit whack-a-mole and we'd love it if we can have the solution in activerecord for real :)

Saidbek:
@jzarnett thanks for following up, nothing else needed on the impl side right now. Looking for a review, and I'd appreciate it if you could give it a try and provide any feedback.

PR #57697, opened 2026-06-13
Don't fire remove_channel for an unknown channel

55728:
### Problem

`SubscriberMap`'s `@subscribers` uses a key-inserting default proc (`Hash.new { |h, k| h[k] = [] }`). `#remove_subscriber` read `@subscribers[channel]` unguarded, so removing a subscriber for a channel that was never subscribed (or was already removed) autovivified an empty array, found it empty, and called `remove_channel` — issuing a real `UNLISTEN` / Redis `unsubscribe` for a channel the server isn't even tracking.

This is the odd one out among the three accessors: the sibling `broadcast` already bails with `return if !@subscribers.key?(channel)`, and `add_subscriber` keys off the same `key?` check to decide whether a channel is new. Only `remove_subscriber` autovivifies.

While the channel-level `stop_stream_from` guards its `pubsub.unsubscribe` call behind `streams.delete(broadcasting)`, the server-wide `SubscriberMap` is still reachable with an unknown channel through the async adapters: `SubscriberMap::Async` posts `add_subscriber` and `remove_subscriber` onto the executor as separate tasks, so a subscribe immediately followed by an unsubscribe can run the remove before the add has registered the channel — autovivifying it and emitting a spurious `UNLISTEN` for a channel that was never `LISTEN`ed.

### Fix

Bail out early when the channel is absent, matching the `key?` guard the sibling `broadcast` already uses:

```diff
 def remove_subscriber(channel, subscriber)
   @sync.synchronize do
+    return if !@subscribers.key?(channel)
+
     @subscribers[channel].delete(subscriber)

     if @subscribers[channel].empty?
       @subscribers.delete channel
       remove_channel channel
     end
   end
 end
```

The early `return` value is unused — every caller (`redis`, `postgresql`, `inline`) just invokes it for effect.

### Testing

- New `SubscriberMapTest` test "remove_subscriber from an unknown channel does not trigger remove_channel", next to the existing "broadcast should not change subscribers": **red** = `remove_channel` called 1 time; **green** = 0 times.
- Full `actioncable/test/subscription_adapter/subscriber_map_test.rb`: **2 runs, 0 failures**.

No CHANGELOG entry (bug fix).

### DIFF
--- a/actioncable/lib/action_cable/subscription_adapter/subscriber_map.rb
+++ b/actioncable/lib/action_cable/subscription_adapter/subscriber_map.rb
@@ -26,7 +26,8 @@ module ActionCable
 
       def remove_subscriber(channel, subscriber)
         @sync.synchronize do
-          @subscribers[channel].delete(subscriber)
+          return if !@subscribers.key?(channel)
+          return unless @subscribers[channel].delete(subscriber)
 
           if @subscribers[channel].empty?
             @subscribers.delete channel
--- a/actioncable/test/subscription_adapter/subscriber_map_test.rb
+++ b/actioncable/test/subscription_adapter/subscriber_map_test.rb
@@ -12,6 +12,29 @@ class SubscriberMapTest < ActionCable::TestCase
     assert_equal origin, @subscription_map.instance_variable_get(:@subscribers)
   end
 
+  test "remove_subscriber from an unknown channel does not trigger remove_channel" do
+    setup_subscription_map
+    origin = @subscription_map.instance_variable_get(:@subscribers).dup
+
+    assert_not_called(@subscription_map, :remove_channel) do
+      @subscription_map.remove_subscriber("not_exist_channel", "subscriber")
+    end
+
+    assert_equal origin, @subscription_map.instance_variable_get(:@subscribers)
+  end
+
+  test "remove_subscriber for an absent subscriber on a known channel does not trigger remove_channel" do
+    setup_subscription_map
+    @subscription_map.add_subscriber("channel", "subscriber", nil)
+    origin = @subscription_map.instance_variable_get(:@subscribers).dup
+
+    assert_not_called(@subscription_map, :remove_channel) do
+      @subscription_map.remove_subscriber("channel", "other_subscriber")
+    end
+
+    assert_equal origin, @subscription_map.instance_variable_get(:@subscribers)
+  end
+
   private
     def setup_subscription_map
       @subscription_map = ActionCable::SubscriptionAdapter::SubscriberMap.new


actioncable/lib/action_cable/subscription_adapter/subscriber_map.rb
```diff
@@ -26,6 +26,8 @@ def add_subscriber(channel, subscriber, on_success)
 
       def remove_subscriber(channel, subscriber)
         @sync.synchronize do
+          return if !@subscribers.key?(channel)
+
           @subscribers[channel].delete(subscriber)
```
byroot:
```suggestion
          return unless @subscribers[channel].delete(subscriber)
```

55728: Thanks @byroot! 😊 Using the return value of `delete` to gate the cleanup is nicer — I've adopted that part.

I kept the `return if !@subscribers.key?(channel)` line, though, rather than collapsing all of 29–31 into the single `return unless @subscribers[channel].delete(subscriber)`. The reason is that `@subscribers` has a key-inserting default proc (`Hash.new { |h, k| h[k] = [] }`), so reading `@subscribers[channel]` for an unknown channel autovivifies `channel => []`. That doesn't fire `remove_channel` (good — `delete` returns `nil` and we bail), but it leaves a stray empty array behind, which then trips up a subsequent `add_subscriber`:

```ruby
new_channel = !@subscribers.key?(channel)  # now false, because remove autovivified the key
...
add_channel channel, on_success if new_channel  # skipped → no LISTEN/subscribe ever issued
```

This is exactly the `SubscriberMap::Async` race the PR targets (remove posted before the add has registered the channel). I verified it against the real `SubscriberMap` (overriding only `remove_subscriber` with the suggested one-liner, everything else untouched), instrumenting `add_channel`/`remove_channel`:

```ruby
m.remove_subscriber("ch", "sub")    # remove runs first
m.add_subscriber("ch", "sub", nil)  # add runs after

# WITHOUT the key? guard (suggested one-liner):
#   add_channel calls => []                      → add_channel never fires
#   @subscribers      => {"ch" => ["sub"]}        → "sub" registered, but no LISTEN/subscribe issued
#
# WITH the key? guard (kept):
#   add_channel calls => ["add_channel(ch)"]      → LISTEN/subscribe issued as expected
#   @subscribers      => {"ch" => ["sub"]}
```

So keeping the `key?` guard avoids the autovivification, and your `return unless …delete(…)` then handles the "channel known but subscriber not present" case cleanly:

```ruby
def remove_subscriber(channel, subscriber)
  @sync.synchronize do
    return if !@subscribers.key?(channel)
    return unless @subscribers[channel].delete(subscriber)

    if @subscribers[channel].empty?
      @subscribers.delete channel
      remove_channel channel
    end
  end
end
```

Happy to drop the guard if you think the remove-before-add ordering can't actually happen (e.g. the executor serializes them) — let me know what you'd prefer.

PR #57701, opened 2026-06-13
Add 2-day cooldown to default Gemfile templates

vivekmiyani:
### Motivation / Background

This Pull Request has been created to protect new Rails applications from supply-chain attacks. Bundler 4.0.13 introduced a `cooldown` feature.

Reference: https://blog.rubygems.org/2026/06/03/cooldown-let-new-gems-be-vetted.html

### Detail

This Pull Request changes the default `Gemfile` template in `railties`. It adds `cooldown: 7` to the default `https://rubygems.org` source.

### Checklist

### DIFF
--- a/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
@@ -1,4 +1,4 @@
-source "https://rubygems.org"
+source "https://rubygems.org", cooldown: 7
 
 <% gemfile_entries.each do |gemfile_entry| -%>
 <%= gemfile_entry %>
--- a/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
+++ b/railties/lib/rails/generators/rails/app/templates/Gemfile.tt
@@ -1,4 +1,4 @@
-source "https://rubygems.org", cooldown: 7
+source "https://rubygems.org", cooldown: 2
 
 <% gemfile_entries.each do |gemfile_entry| -%>
 <%= gemfile_entry %>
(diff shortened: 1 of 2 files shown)


vivekmiyani:
Note: This feature was introduced in Bundler 4.0.13. I tested it against older versions (like 4.0.12), and Bundler ignores the `cooldown` keyword rather than breaking. 

However, if the core team feels this is too strict, and it might cause confusion for older setups, I can wrap this in a Bundler version check.

byroot:
I'm generally in favor of this, however:

  - I think 7 days is too much, the goal of cooldown is to leave time for automated scanner. The recent attacks were found, and the package yanked in under an hour. One or two days should be plenty. 
  - I'm don't remember out of the top of my head what `rails new` does about bundler, ideally we'd update it if you aren't yet on a supported versions. So there's a bit of testing / research to do it.

justinko:
7 days seems to be the sweet spot: https://nesbitt.io/2026/03/04/package-managers-need-to-cool-down.html

I vote for 3 days.

vivekmiyani:
I've updated the cooldown to 2. I can update to 3 based on feedback if needed.

Regarding Bundler versions, `rails new` delegates to the system's installed Bundler.

https://github.com/rails/rails/blob/43a1f3b574cceafcd78d288a1339e37711cf79c7/railties/lib/rails/generators/app_base.rb#L709-L711

If we want to encourage users to upgrade, we can just print a warning saying "Your Bundler version does not support the cooldown feature".

jeromedalbert:
Additional data point: GitHub [announced](https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/) a default cooldown of 3 days for Dependabot updates.

eliotsykes:
Would this cooldown prevent dependabot from opening PRs for security updates?

The reason I ask is dependabot ignores its own default 3-day cooldown for insecure gems, to patch them immediately.

PR #57710, opened 2026-06-13
Fix `has_many_attached` `attach` deleting files attached concurrently

### LINKED ISSUE #42941: `ActiveStorage::Attached::Many#attach` is not "thread safe"
### Steps to reproduce

First setup the sample application:

1. `gem install rails -v6.1.4`
2. `rails new ConcurrentAttachTestApp --skip-action-mailer --skip-action-mailbox --skip-action-text --skip-action-cable --skip-sprockets --skip-spring --skip-listen --skip-javascript --skip-turbolinks --skip-jbuilder --skip-test --skip-system-test --skip-bootsnap`
3. apply the following diff:

    ```diff
    diff --git a/Gemfile b/Gemfile
    index 01a705b..25b3e19 100644
    --- a/Gemfile
    +++ b/Gemfile
    @@ -15,6 +15,8 @@ gem 'puma', '~> 5.0'
     # Use Active Storage variant
     # gem 'image_processing', '~> 1.2'
    
    +gem 'pg'
    +
     group :development, :test do
       # Call 'byebug' anywhere in the code to stop execution and get a debugger console
       gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
    @@ -26,6 +28,8 @@ group :development do
       # Display performance information such as SQL time and flame graphs for each request in your browser.
       # Can be configured to work on production as well see: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/README.md
       gem 'rack-mini-profiler', '~> 2.0'
    +
    +  gem 'parallel'
     end
    
     # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
    diff --git a/app/models/user.rb b/app/models/user.rb
    new file mode 100644
    index 0000000..2db75b6
    --- /dev/null
    +++ b/app/models/user.rb
    @@ -0,0 +1,3 @@
    +class User < ApplicationRecord
    +  has_many_attached :files
    +end
    diff --git a/config/database.yml b/config/database.yml
    index 4a8a1b2..78ac63a 100644
    --- a/config/database.yml
    +++ b/config/database.yml
    @@ -10,8 +10,12 @@ default: &default
       timeout: 5000
    
     development:
    -  <<: *default
    -  database: db/development.sqlite3
    +  adapter: postgresql
    +  database: concurrent_attach_development
    +  encoding: unicode
    +  host: localhost
    +  pool: 5
    +  username: postgres
    
     # Warning: The database defined as "test" will be erased and
     # re-generated from your development database when you run "rake".
    diff --git a/db/migrate/20210804080852_create_active_storage_tables.active_storage.rb b/db/migrate/20210804080852_create_active_storage_tables.active_storage.rb
    new file mode 100644
    index 0000000..8779826
    --- /dev/null
    +++ b/db/migrate/20210804080852_create_active_storage_tables.active_storage.rb
    @@ -0,0 +1,36 @@
    +# This migration comes from active_storage (originally 20170806125915)
    +class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
    +  def change
    +    create_table :active_storage_blobs do |t|
    +      t.string   :key,          null: false
    +      t.string   :filename,     null: false
    +      t.string   :content_type
    +      t.text     :metadata
    +      t.string   :service_name, null: false
    +      t.bigint   :byte_size,    null: false
    +      t.string   :checksum,     null: false
    +      t.datetime :created_at,   null: false
    +
    +      t.index [ :key ], unique: true
    +    end
    +
    +    create_table :active_storage_attachments do |t|
    +      t.string     :name,     null: false
    +      t.references :record,   null: false, polymorphic: true, index: false
    +      t.references :blob,     null: false
    +
    +      t.datetime :created_at, null: false
    +
    +      t.index [ :record_type, :record_id, :name, :blob_id ], name: "index_active_storage_attachments_uniqueness", unique: true
    +      t.foreign_key :active_storage_blobs, column: :blob_id
    +    end
    +
    +    create_table :active_storage_variant_records do |t|
    +      t.belongs_to :blob, null: false, index: false
    +      t.string :variation_digest, null: false
    +
    +      t.index %i[ blob_id variation_digest ], name: "index_active_storage_variant_records_uniqueness", unique: true
    +      t.foreign_key :active_storage_blobs, column: :blob_id
    +    end
    +  end
    +end
    diff --git a/db/schema.rb b/db/schema.rb
    new file mode 100644
    index 0000000..dbb533c
    --- /dev/null
    +++ b/db/schema.rb
    @@ -0,0 +1,51 @@
    +# This file is auto-generated from the current state of the database. Instead
    +# of editing this file, please use the migrations feature of Active Record to
    +# incrementally modify your database, and then regenerate this schema definition.
    +#
    +# This file is the source Rails uses to define your schema when running `bin/rails
    +# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
    +# be faster and is potentially less error prone than running all of your
    +# migrations from scratch. Old migrations may fail to apply correctly if those
    +# migrations use external dependencies or application code.
    +#
    +# It's strongly recommended that you check this file into your version control system.
    +
    +ActiveRecord::Schema.define(version: 2021_08_04_080852) do
    +
    +  # These are extensions that must be enabled in order to support this database
    +  enable_extension "plpgsql"
    +
    +  create_table "active_storage_attachments", force: :cascade do |t|
    +    t.string "name", null: false
    +    t.string "record_type", null: false
    +    t.bigint "record_id", null: false
    +    t.bigint "blob_id", null: false
    +    t.datetime "created_at", null: false
    +    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
    +    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
    +  end
    +
    +  create_table "active_storage_blobs", force: :cascade do |t|
    +    t.string "key", null: false
    +    t.string "filename", null: false
    +    t.string "content_type"
    +    t.text "metadata"
    +    t.string "service_name", null: false
    +    t.bigint "byte_size", null: false
    +    t.string "checksum", null: false
    +    t.datetime "created_at", null: false
    +    t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
    +  end
    +
    +  create_table "active_storage_variant_records", force: :cascade do |t|
    +    t.bigint "blob_id", null: false
    +    t.string "variation_digest", null: false
    +    t.index ["blob_id", "variation_digest"], name: "index_active_storage_variant_records_uniqueness", unique: true
    +  end
    +
    +  create_table "users", force: :cascade do |t|
    +  end
    +
    +  add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
    +  add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
    +end
    ```

4. `bundle install`
5. `bin/rails active_storage:install`
6. `bin/rails db:setup`

Now try the following code in the `bin/rails console`:

```ruby
User.create

Parallel.each(1..10) do
  ActiveRecord::Base.connection_pool.with_connection do
    User.first.files.attach(io: File.open("README.md"), filename: "README.md")
  end
end

User.first.files.count
```

Note that `User.first.files.count` may return a smaller number than 10.

### Expected behavior
Each `Object#attach` appends the new file and does not touch existing attachments.

### Actual behavior
`Object#attach` may purge files that are uploaded concurrently.

### System configuration
**Rails version**: 6.1.4

**Ruby version**: 3.0.1

williantenfen:
### Motivation / Background

Fixes #42941

On a persisted record, `attach` read the current set of blobs and reassigned the whole collection, so two concurrent `attach` calls working from the same snapshot would each replace the collection and delete the other's file.

The Guides document that `attach` always appends and never replaces, so `attach` now inserts only the new attachments instead of reassigning the collection. The staged change tracks whether it should append or replace, which lets `attach` fold new files into a pending change while preserving assignment (`record.files = [...]`) semantics.

### Detail

`ActiveStorage::Attached::Many#attach` was implemented as a read-modify-**replace** of the entire collection:
```ruby
def attach(*attachables)
  record.public_send("#{name}=", blobs + attachables.flatten)
  # ...
end
```

It read blobs (the current set), appended the new attachables, and reassigned #{name}=. On save, CreateMany#assign_associated_attachments does record.#{name}_attachments = ..., and a has_many collection assignment destroys members not in the assigned set. So when two requests both load [A] and one attaches B while the other attaches C, the second save reassigns the collection from its stale [A] snapshot and deletes the first request's B. The added file (B or C) is silently lost.

This conflicts with the documented contract in the Active Storage guide:

 <b>Calling .attach always appends new files</b>. It never replaces existing attachments

This PR makes attach:

- ActiveStorage::Attached::Changes::CreateMany gains an append flag (default false, preserving the existing replace behavior used by record.files = [...]). When set, save inserts only the newly built attachments via concat instead of reassigning the whole collection, leaving existing rows — including ones added concurrently — untouched.
- ActiveStorage::Attached::Many#attach now stages an append-mode change. It folds the new files into any change that's already pending and preserves that change's mode, so:
 -- a bare attach (nothing pending) appends;
 -- record.files = [...] then attach stays a replace (assignment wins);
 -- record.files = [] then attach replaces;
 -- repeated attach calls (including on a dirty/unsaved record, before a single save) stay an append.

Because the existing-vs-new determination is already by blob_id (plus the index_active_storage_attachments_uniqueness unique index), the insert-only path is order-independent and behavior-preserving for the non-concurrent cases: attaching an already-attached blob remains a no-op, and attaching the same file bytes twice still produces two blobs/attachments as before.

### Additional information

The new regression tests use the Group model rather than User. User's table has a lock_version column, and its optimistic locking masks the bug by raising ActiveRecord::StaleObjectError instead of silently losing data — most application models that hit this have no lock_version, so they get the silent data loss. Group has no optimistic locking, which lets two stale instances reproduce the race deterministically.

Tests added (they fail on main, pass with this change):
- attaching does not delete attachments added concurrently
- attaching after changing another attribute does not delete attachments added concurrently (covers the deferred/dirty-record save path)
- assigning then attaching on an existing record replaces and then appends (guards that = replace semantics are preserved)

The full ActiveStorage::ManyAttachedTest and ActiveStorage::OneAttachedTest suites pass with no new failures.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,15 @@
+*   Fix `has_many_attached` `attach` deleting files attached concurrently.
+
+    On a persisted record, `attach` read the current set of blobs and reassigned
+    the whole collection, so two concurrent `attach` calls working from the same
+    snapshot would each replace the collection and delete the other's file. The
+    Guides document that `attach` always appends and never replaces, so `attach`
+    now inserts only the new attachments instead of reassigning the collection.
+
+    Fixes #42941.
+
+    *Willian Tenfen Wazilewski*
+
 *   Fix `MirrorService#mirror` losing blob metadata when copying to mirrors.
 
     Mirrored copies on S3, Azure, and GCS were served as `application/octet-stream`
--- a/activestorage/lib/active_storage/attached/changes/create_many.rb
+++ b/activestorage/lib/active_storage/attached/changes/create_many.rb
@@ -4,9 +4,11 @@ module ActiveStorage
   class Attached::Changes::CreateMany # :nodoc:
     attr_reader :name, :attachables, :pending_uploads
     attr_accessor :record
+    attr_writer :append
 
     def initialize(name, record, attachables, pending_uploads: [])
       @name, @record, @attachables = name, record, Array(attachables)
+      @append = false
       blobs.each(&:identify_without_saving)
       @pending_uploads = Array(pending_uploads) + subchanges_without_blobs
       attachments
@@ -28,6 +30,10 @@ module ActiveStorage
       pending_uploads.each(&:upload)
     end
 
+    def append?
+      @append
+    end
+
     def save
       assign_associated_attachments
       reset_associated_blobs
@@ -47,11 +53,19 @@ module ActiveStorage
       end
 
       def assign_associated_attachments
-        record.public_send("#{name}_attachments=", persisted_or_new_attachments)
+        if append?
+          record.public_send("#{name}_attachments").concat(new_attachments)
+        else
+          record.public_send("#{name}_attachments=", persisted_or_new_attachments)
+        end
       end
 
       def reset_associated_blobs
         record.public_send("#{name}_blobs").reset
+      end
+
+      def new_attachments
+        attachments.select(&:new_record?)
       end
 
       def persisted_or_new_attachments
--- a/activestorage/lib/active_storage/attached/many.rb
+++ b/activestorage/lib/active_storage/attached/many.rb
@@ -49,10 +49,22 @@ module ActiveStorage
     #   document.images.attach(io: File.open("/path/to/racecar.jpg"), filename: "racecar.jpg", content_type: "image/jpeg")
     #   document.images.attach([ first_blob, second_blob ])
     def attach(*attachables)
+      append =
+        case change = record.attachment_changes[name]
+        when ActiveStorage::Attached::Changes::CreateMany then change.append?
+        when ActiveStorage::Attached::Changes::DeleteMany then false
+        else true
+        end
+
       record.public_send("#{name}=", blobs + attachables.flatten)
+
+      change = record.attachment_changes[name]
+      change.append = append if change.is_a?(ActiveStorage::Attached::Changes::CreateMany)
+
       if record.persisted? && !record.changed?
         return if !record.save
       end
+
       record.public_send("#{name}")
     end
 
--- a/activestorage/test/database/create_groups_migration.rb
+++ b/activestorage/test/database/create_groups_migration.rb
@@ -3,6 +3,7 @@
 class ActiveStorageCreateGroups < ActiveRecord::Migration[6.0]
   def change
     create_table :groups do |t|
+      t.string :name
     end
   end
 end
--- a/activestorage/test/models/attached/many_test.rb
+++ b/activestorage/test/models/attached/many_test.rb
@@ -989,4 +989,47 @@ class ActiveStorage::ManyAttachedTest < ActiveSupport::TestCase
     special_user = user.becomes(SpecialUser)
     assert_predicate special_user.highlights, :attached?
   end
+
+  test "attaching does not delete attachments added concurrently" do
+    group = Group.create!
+    group.files.attach create_blob(filename: "a.txt")
+
+    one = Group.find(group.id)
+    two = Group.find(group.id)
+    one.files_blobs.to_a
+    two.files_blobs.to_a
+
+    one.files.attach create_blob(filename: "b.txt")
+    two.files.attach create_blob(filename: "c.txt")
+
+    assert_equal %w[ a.txt b.txt c.txt ], group.reload.files_blobs.map { |blob| blob.filename.to_s }.sort
+  end
+
+  test "attaching after changing another attribute does not delete attachments added concurrently" do
+    group = Group.create!
+    group.files.attach create_blob(filename: "a.txt")
+
+    one = Group.find(group.id)
+    two = Group.find(group.id)
+    one.files_blobs.to_a
+    two.files_blobs.to_a
+
+    one.files.attach create_blob(filename: "b.txt")
+
+    two.name = "Renamed"
+    two.files.attach create_blob(filename: "c.txt")
+    two.save!
+
+    assert_equal %w[ a.txt b.txt c.txt ], group.reload.files_blobs.map { |blob| blob.filename.to_s }.sort
+  end
+
+  test "assigning then attaching on an existing record replaces and then appends" do
+    group = Group.create!
+    group.files.attach create_blob(filename: "old.txt")
+
+    group.files = [ create_blob(filename: "new.txt") ]
+    group.files.attach create_blob(filename: "added.txt")
+
+    assert_equal %w[ added.txt new.txt ], group.reload.files_blobs.map { |blob| blob.filename.to_s }.sort
+  end
 end
--- a/activestorage/test/test_helper.rb
+++ b/activestorage/test/test_helper.rb
@@ -214,6 +214,7 @@ class SpecialUser < User; end
 
 class Group < ActiveRecord::Base
   has_one_attached :avatar
+  has_many_attached :files
   has_many :users, autosave: true
 
   accepts_nested_attributes_for :users


williantenfen:
Running this against master fails:

```ruby
# frozen_string_literal: true

# Reproduction for https://github.com/rails/rails/issues/42941
# `ActiveStorage::Attached::Many#attach` may erase files uploaded concurrently.
#
# Run with:
#   cd activestorage && bin/test test/models/tmp_concurrent_attach_test.rb

require "test_helper"
require "database/setup"

# A model WITHOUT optimistic locking (no lock_version), like the typical app in
# the issue. The dummy `User` table has a `lock_version` column, whose optimistic
# locking masks this bug by raising StaleObjectError instead of silently losing data.
class Gallery < ActiveRecord::Base
  has_many_attached :photos
end

class ActiveStorage::ConcurrentAttachTest < ActiveSupport::TestCase
  include ActiveJob::TestHelper

  setup do
    ActiveRecord::Base.lease_connection.create_table(:galleries, force: true) do |t|
      t.string :name
      t.timestamps
    end
    @gallery = Gallery.create!
  end

  teardown do
    ActiveStorage::Blob.all.each(&:delete)
    ActiveRecord::Base.lease_connection.drop_table(:galleries, if_exists: true)
  end

  # Deterministic version of the race: two separate instances of the same
  # record each read the current set of blobs ([A]) *before* either of them
  # writes. This is exactly what happens when two web requests load the record
  # at the same time and each calls `attach`.
  test "deterministic: concurrent attach erases the other request's file" do
    blob_a = create_blob(filename: "a.txt")
    @gallery.photos.attach(blob_a)
    assert_equal 1, @gallery.reload.photos_attachments.count

    r1 = Gallery.find(@gallery.id)
    r2 = Gallery.find(@gallery.id)

    # Both requests load the current attachments/blobs (= [A]) before saving.
    r1.photos_blobs.to_a
    r2.photos_blobs.to_a

    blob_b = create_blob(filename: "b.txt")
    blob_c = create_blob(filename: "c.txt")

    r1.photos.attach(blob_b) # request 1 commits [A, B]
    r2.photos.attach(blob_c) # request 2 commits [A, C] from its stale [A] snapshot

    filenames = @gallery.reload.photos_blobs.map { |b| b.filename.to_s }.sort
    assert_equal %w[a.txt b.txt c.txt], filenames,
      "expected all three files to survive, but got: #{filenames.inspect}"
  end

  # SAME interleaving, but implemented as an append-only INSERT (what the
  # underlying has_many association does) instead of reassigning the whole set.
  # No locking involved. If this passes, the data loss above is an artifact of
  # attach's replace-the-collection implementation, not an inherent race that
  # requires a locking strategy.
  test "append-only insert survives the identical stale-snapshot interleaving" do
    blob_a = create_blob(filename: "a.txt")
    @gallery.photos.attach(blob_a)

    r1 = Gallery.find(@gallery.id)
    r2 = Gallery.find(@gallery.id)
    r1.photos_blobs.to_a
    r2.photos_blobs.to_a

    blob_b = create_blob(filename: "b.txt")
    blob_c = create_blob(filename: "c.txt")

    r1.photos_attachments.create!(blob: blob_b)
    r2.photos_attachments.create!(blob: blob_c)

    filenames = @gallery.reload.photos_blobs.map { |b| b.filename.to_s }.sort
    assert_equal %w[a.txt b.txt c.txt], filenames,
      "append-only should keep all three, got: #{filenames.inspect}"
  end

  # What happens if a regular attribute is dirtied before attach? The fix's
  # append path only triggers for a persisted & unchanged record with nothing
  # staged. Dirtying an attribute makes `appendable?` false, so attach falls
  # back to the legacy buffered "replace the whole collection" path, which is
  # applied on the next save -- re-exposing the lost update.
  test "dirtying an attribute before attach falls back to replace and loses data" do
    blob_a = create_blob(filename: "a.txt")
    @gallery.photos.attach(blob_a)

    r1 = Gallery.find(@gallery.id)
    r2 = Gallery.find(@gallery.id)
    r1.photos_blobs.to_a
    r2.photos_blobs.to_a

    blob_b = create_blob(filename: "b.txt")
    blob_c = create_blob(filename: "c.txt")

    r1.photos.attach(blob_b) # r1 clean -> append path -> [A, B]

    r2.name = "Test"         # dirty a normal column
    r2.photos.attach(blob_c) # not appendable -> buffered replace over stale [A]
    survives_before_save = @gallery.reload.photos_blobs.map { |b| b.filename.to_s }.sort
    r2.save!                 # buffered CreateMany replaces collection with [A, C]

    filenames = @gallery.reload.photos_blobs.map { |b| b.filename.to_s }.sort
    puts "  [before r2.save!] #{survives_before_save.inspect}"
    puts "  [after  r2.save!] #{filenames.inspect}"
    assert_equal %w[a.txt b.txt c.txt], filenames,
      "expected all three to survive, got: #{filenames.inspect}"
  end
end
```

PR #57716, opened 2026-06-14
Clean up the subscription slot when subscribe_to_channel raises

chaadow:
### Motivation / Background

This Pull Request has been created because a failed `subscribe_to_channel` leaves a half-initialized subscription stranded in the connection's `subscriptions` hash, permanently poisoning that identifier's slot.

`Connection::Subscriptions#add` inserts the subscription into the hash before calling `subscribe_to_channel`:

```ruby
subscriptions[id_key] = subscription
subscription.subscribe_to_channel
```

If `subscribe_to_channel` raises (e.g. an exception in the channel's `subscribed` or a `before_subscribe` callback), the entry is never removed. Since `add` raises `AlreadySubscribedError` when the key is already present, every later subscribe for that identifier is rejected for the lifetime of the connection, and the client's `SubscriptionGuarantor` retries it forever.

### Detail

This Pull Request changes `add` to remove the entry when `subscribe_to_channel` does not complete successfully.

Reordering (subscribe first, insert after) is **not** an option: the insert-before-subscribe order is intentional, because `subscribe_to_channel` may call `reject_subscription`, which removes the entry via `connection.subscriptions.remove_subscription`. So the fix wraps `subscribe_to_channel` and deletes the entry only on a non-successful subscribe, leaving the rejection and success paths untouched:

```ruby
subscriptions[id_key] = subscription

successfully_subscribed = false
begin
  subscription.subscribe_to_channel
  successfully_subscribed = true
ensure
  subscriptions.delete(id_key) unless successfully_subscribed
end
```

A test subscribes a channel whose `subscribed` raises, then asserts the identifier is not left in `identifiers` and that the same identifier can be subscribed again (reaching the channel instead of being rejected with `AlreadySubscribedError`).

### Checklist

### DIFF
--- a/actioncable/lib/action_cable/connection/subscriptions.rb
+++ b/actioncable/lib/action_cable/connection/subscriptions.rb
@@ -76,7 +76,18 @@ module ActionCable
 
         if subscription
           subscriptions[id_key] = subscription
-          subscription.subscribe_to_channel
+
+          successfully_subscribed = false
+          begin
+            subscription.subscribe_to_channel
+            successfully_subscribed = true
+          ensure
+            # Don't leave a half-initialized subscription occupying the
+            # identifier's slot if subscribe_to_channel raised: it would reject
+            # every later subscribe for that identifier with AlreadySubscribedError
+            # until the connection is torn down.
+            subscriptions.delete(id_key) unless successfully_subscribed
+          end
         else
           id_options = ActiveSupport::JSON.decode(id_key).with_indifferent_access
           raise ChannelNotFound, id_options[:channel]
--- a/actioncable/test/connection/subscriptions_test.rb
+++ b/actioncable/test/connection/subscriptions_test.rb
@@ -37,6 +37,12 @@ class ActionCable::Connection::SubscriptionsTest < ActionCable::TestCase
     end
   end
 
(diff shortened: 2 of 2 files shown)


chaadow:
@byroot This doesn't completely fix https://github.com/rails/rails/issues/44652, but i think it may get rid of one of the conditions under which `AlreadySubscribedError` is raised.

palkan:
Hey @chaadow,

Thanks for the PR!

> This doesn't completely fix https://github.com/rails/rails/issues/44652

This can cause repeated subscribe commands but that's not the common cause of #44652.
---

I'm hesitant to approve this change (if my voice counts 🙂) because it would lead to a malformed subscribe request to stuck in a loop and raise the same exception over and over again; and that won't be a _cheap_ AlreadySubscribedError (that doesn't touch the application code).

I'd better reject failed subscription with the "error" reason but... the protocol is not rich enough to express this (and updating the protocol is a much bigger scope).

So, for now I'd recommend application developers to keep `#subscribed` callback as robust as possible and rescue from potential exceptions at their side.

The only problem I see here that the "already subscribed" error message in the subsequent attempt is misleading 🤔.

byroot:
We've discussed this bug recently, I believe @tenderlove has a fix for it.

PR #57727, opened 2026-06-14
Accept a Float wait in `retry_on`

55728:
### Summary

`retry_on` documents `:wait` as "a delay specified either in seconds (default: 3 seconds)". An Integer wait works, and the public `set(wait: 2.5)` API already accepts a bare Float, but `retry_on ..., wait: 2.5` crashes:

```ruby
class FloatWaitJob < ActiveJob::Base
  retry_on SomeError, wait: 2.5, attempts: 3
  def perform; raise SomeError; end
end

FloatWaitJob.perform_now
# => RuntimeError: Couldn't determine a delay based on 2.5
```

`determine_delay` matches `ActiveSupport::Duration, Integer` but not a bare `Float`, so a Float falls through to the `else` branch and raises. Because this happens inside the `rescue_from` handler, it **replaces the original application exception** with a confusing internal error and the job is never re-enqueued.

```ruby
when ActiveSupport::Duration, Integer
  delay = seconds_or_duration_or_algorithm.to_i
  ...
else
  raise "Couldn't determine a delay based on #{seconds_or_duration_or_algorithm.inspect}"
```

### Fix

Match `Numeric` so a Float is handled like an Integer:

```ruby
when ActiveSupport::Duration, Numeric
  delay = seconds_or_duration_or_algorithm.to_i
```

`.to_i` is kept, so a Float wait truncates to whole seconds exactly as an `ActiveSupport::Duration` wait already does (`2.5.seconds` → 2). This keeps the change minimal and avoids altering the long-standing delay granularity; enabling genuine sub-second delays would be a separate decision (it depends on adapter support).

### Tests

Added `FloatWaitError` to the retry test job (`wait: 2.5`) and a test asserting the retry is scheduled at the expected delay, mirroring the existing "long wait job" test.

### DIFF
--- a/activejob/lib/active_job/exceptions.rb
+++ b/activejob/lib/active_job/exceptions.rb
@@ -179,7 +179,7 @@ module ActiveJob
           delay = executions**4
           delay_jitter = determine_jitter_for_delay(delay, jitter)
           delay + delay_jitter + 2
-        when ActiveSupport::Duration, Integer
+        when ActiveSupport::Duration, Integer, Float
           delay = seconds_or_duration_or_algorithm.to_i
           delay_jitter = determine_jitter_for_delay(delay, jitter)
           delay + delay_jitter
--- a/activejob/test/cases/exceptions_test.rb
+++ b/activejob/test/cases/exceptions_test.rb
@@ -119,6 +119,21 @@ class ExceptionsTest < ActiveSupport::TestCase
       end
     end
 
+    test "float wait job" do
+      travel_to Time.now
+      random_amount = 1
+      delay_for_jitter = random_amount * 2 * ActiveJob::Base.retry_jitter
+
+      Kernel.stub(:rand, random_amount) do
+        RetryJob.perform_later "FloatWaitError", 2, :log_scheduled_at
+        assert_equal [
+          "Raised FloatWaitError for the 1st time",
+          "Next execution scheduled at #{(Time.now + 2.seconds + delay_for_jitter).to_f}",
+          "Successfully completed job"
+        ], JobBuffer.values
+      end
+    end
+
     test "polynomially retrying job includes jitter" do
       travel_to Time.now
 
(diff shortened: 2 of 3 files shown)


fatkodima:
While a simple change, I wonder if it makes sense to accept a float given that it will be converted to an integer in any case.
@rafaelfranca wdyt?

activejob/lib/active_job/exceptions.rb
```diff
@@ -179,7 +179,7 @@ def determine_delay(seconds_or_duration_or_algorithm:, executions:, error: nil,
           delay = executions**4
           delay_jitter = determine_jitter_for_delay(delay, jitter)
           delay + delay_jitter + 2
-        when ActiveSupport::Duration, Integer
+        when ActiveSupport::Duration, Numeric
```
byroot: `Numeric` is too wide:

```
>> Numeric.subclasses
=> [Date::Infinity, Complex, Rational, Float, Integer]
```
Don't make sense to allow Complex or Rational here.

55728: Good catch — narrowed to `Integer, Float`.

p8: But float also doesn’t make sense, as to_i turns it into an Integer again on the next line, as Fatko mentioned.

p8: Maybe it should call to_f instead?

55728: The `to_i` truncation is intentional — it matches what Duration already does. Enabling sub-second delays would depend on adapter support, so I kept it out of scope here.

p8: Yes, but durations are always in whole seconds. Calling to_i on a duration doesn’t round it, it returns the value. Allowing a float implies a granularity that isn’t there.

p8: This also still raises on job retry when passing other unsupported types, replacing the original error. Maybe retry_on should raise an ArgumentError for unsupported types for wait? So an error is raised when the class is loaded, instead of raising after the job failed.

55728: Good point — validating the `wait:` type eagerly in `retry_on` (at class load time) would catch unsupported types much earlier. I'll open a follow-up PR for that.

Issue #54300, opened 2025-01-20
ProxyController's digest is different if format specified as symbol for preprocessed variants


fenelon:
I've encountered an issue with proxy mode and ActiveStorage::Representations::ProxyController.

When using `preprocessed: true` on a variant, specifying format as a symbol will prevent the ActiveStorage::Representations::ProxyController to find a preproceed variant. Changing the format to a string works. When accessing a variant using `variant(:variant_name).proceesed`, both symbols and stings are handled correctly, the problem seems to be only with the ProxyController's algorithm for calculating the digest differently. 

Example:

```ruby
class Image < ApplicationRecord
  has_one_attached :original, dependent: :destroy do |attachable|
    saver = { quality: 80, smart_subsample: true, lossless: false, strip: true }
    
    # Will NOT use the preprocessed variant in proxy mode
    attachable.variant :tiny, resize_to_fill: [120, 120, { crop: :low }], 
                              preprocessed: true, 
                              format: :webp, 
                              saver: saver 

    # Will use the preprocessed variant in proxy mode
    attachable.variant :thumb, resize_to_fill: [200, 200, { crop: :low }], 
                               preprocessed: true, 
                               format: 'webp', 
                               saver: saver
  end
end
```

Then reproduce like this

```ruby
image = Image.last
image.original.variant(:tiny).url # will be processed on request
image.original.variant(:thumb).url # will serve the preprocessed image
```

zzak:
Can you add a bit more detail about how to reproduce the issue, either provide a repo or [bug script](https://github.com/rails/rails/blob/main/guides/bug_report_templates/active_storage.rb).

I tried to reproduce the issue but both variants were processed on create:
https://github.com/zzak/Issue_Check_54300

```
[ActiveJob] Enqueued ActiveStorage::AnalyzeJob (Job ID: 1e3995d6-909f-4c39-8eb1-62660d3cb9c3) to Async(default) with arguments: #<GlobalID:0x00007
f2e0eb51da8 @uri=#<URI::GID gid://issue-check54300/ActiveStorage::Blob/7>>
[ActiveJob] ↳ app/controllers/images_controller.rb:27:in `block in create'
[ActiveJob] Enqueued ActiveStorage::TransformJob (Job ID: a7a8cd23-e90d-4047-986e-5410eb316d4f) to Async(default) with arguments: #<GlobalID:0x000
[...]

PR #57731, opened 2026-06-14
Fix Active Storage variants re-processed when format is a Symbol

### LINKED ISSUE #57302: Active Storage variant format: :webp (Symbol) causes cache misses — variant re-processed on every request
### Steps to reproduce

```ruby
# In your model:
has_one_attached :image do |attachable|
  attachable.variant :medium, format: :webp, resize_to_limit: [800, 800]
end

# Requesting the variant repeatedly:
photo = Photo.first
photo.image.variant(:medium).processed.url  # processes the image
photo.image.variant(:medium).processed.url  # processes the image AGAIN
photo.image.variant(:medium).processed.url  # processes the image AGAIN 2
``` 

Changing `format: :webp` to `format: "webp"` (String instead of Symbol) fixes the issue.

### Expected behavior

The variant is processed once and the cached version is served on subsequent requests.

### Actual behavior

The variant is re-processed on every request. `ActiveStorage::Variation` computes a digest to identify a variant for cache lookup. A Symbol and a String serialize differently, so `format: :webp` and `format: "webp"` produce different digests. The variant is stored under one digest but looked up under a different one, causing a silent cache miss every time. There is no warning or error — images are just processed repeatedly, causing severe performance degradation.

### System configuration

**Rails version**: 8.1.3

**Ruby version**: 3.3.4

hanscanonico:
### Motivation / Background

Fixes #57302.

When an Active Storage variant is defined with a Symbol `format` (e.g. `format: :webp`), the variant is re-processed on every request instead of being served from the cache, causing severe, silent performance degradation. Changing `format: :webp` to `format: "webp"` works around it.

```ruby
has_one_attached :image do |attachable|
  attachable.variant :medium, format: :webp, resize_to_limit: [800, 800]
end

photo.image.variant(:medium).processed.url # re-processes every time
```

### Detail

A variation's `digest` (used as the lookup key for `ActiveStorage::VariantRecord`) is derived from its transformations via `Marshal.dump`. `Marshal` encodes the Symbol `:webp` differently from the String `"webp"` — and even `:webp.to_s` (US-ASCII) differs from a UTF-8 `"webp"`.

Named variants keep `format: :webp` as a **Symbol** in memory, so the variant is stored under the Symbol-based digest. But when the variant is later requested through its signed `variation_key`, the message verifier decodes it with the configured serializer — JSON since `load_defaults 7.1` (`config.active_support.message_serializer = :json_allow_marshal`) — which turns the format into a **String**. The lookup digest no longer matches the stored one, the cache silently misses, and the image is re-processed.

This is the same underlying inconsistency discussed in #54300.

**Fix:** normalize the `:format` transformation to a UTF-8 String in `ActiveStorage::Variation#initialize`, so the Symbol form and the String form produce the same `digest`. Variants already stored with a String format keep the same digest and are not invalidated; only the previously-broken Symbol case changes (it processes once more, then caches correctly).

### Additional information

- New unit tests in `variation_test.rb`: the `:format` is normalized to a UTF-8 String, Symbol vs String formats produce the same digest, and a variation decoded from its own signed `key` has the same digest as the in-memory one (this last test reproduces the actual bug and fails on `main`).
- New end-to-end test in `variant_with_record_test.rb`: a variant stored via the in-memory Symbol form is reused (not re-processed) when looked up via the decoded String form. Confirmed it fails before the fix (a second `VariantRecord` is created) and passes after.
- Applications still on the Marshal message serializer (`load_defaults` below 7.1) were internally consistent on the Symbol digest, so they never hit the bug; after this change they reprocess variants with an explicit `format:` once. Noted in the CHANGELOG.
- Only `format` is normalized. Other Symbol-valued transformations (e.g. `saver: { subsample_mode: :on }`) also change type through the JSON serializer, but a general normalization would change the digest of every existing variant that uses one, a much wider blast radius. `format` is the common case since Active Storage's own `default_variant_transformations` injects it. Complementary to #58473, which keeps transformation key order stable at enqueue time.

### Checklist

### DIFF
--- a/activestorage/CHANGELOG.md
+++ b/activestorage/CHANGELOG.md
@@ -1,3 +1,17 @@
+*   Fix variants being re-processed on every request when a `format` is given as a Symbol.
+
+    A variation's digest is derived from its transformations via `Marshal.dump`,
+    which encodes a Symbol (e.g. `format: :webp`, as kept in memory for named
+    variants) differently from the String the message serializer produces when
+    decoding a signed variation key. The variant was therefore stored under one
+    digest but looked up under another, silently missing the cache and
+    re-processing the image on every request. The `format` is now normalized to a
+    UTF-8 String so both forms resolve to the same variant.
+
+    Fixes #57302.
+
+    *Hans Canonico*
+
 *   Fix `MirrorService#mirror` losing blob metadata when copying to mirrors.
 
     Mirrored copies on S3, Azure, and GCS were served as `application/octet-stream`
--- a/activestorage/app/models/active_storage/variation.rb
+++ b/activestorage/app/models/active_storage/variation.rb
@@ -45,6 +45,16 @@ class ActiveStorage::Variation
 
   def initialize(transformations)
     @transformations = transformations.deep_symbolize_keys
+
+    if @transformations.key?(:format)
+      # Normalize the format to a UTF-8 String so that a variation referenced
+      # with a Symbol (e.g. +format: :webp+, as kept in memory for named
+      # variants) and the same variation decoded from a signed key (where the
+      # JSON message serializer turns it into a String) produce the same
+      # #digest. Otherwise the variant is stored under one digest but looked up
+      # under another, causing it to be re-processed on every request.
+      @transformations[:format] = @transformations[:format].to_s.encode(Encoding::UTF_8)
+    end
   end
 
   def default_to(defaults)
--- a/activestorage/test/models/variant_with_record_test.rb
+++ b/activestorage/test/models/variant_with_record_test.rb
@@ -54,6 +54,24 @@ class ActiveStorage::VariantWithRecordTest < ActiveSupport::TestCase
     assert_equal 67, image.height
   end
 
+  test "serving a variant processed with a Symbol format when looked up with a String format" do
+    blob = create_file_blob(filename: "racecar.jpg")
+
+    # Variants defined in the model keep their format as a Symbol in memory.
+    assert_difference -> { blob.variant_records.count }, +1 do
+      blob.variant(resize_to_limit: [100, 100], format: :webp).processed
+    end
+
+    # When a variant is requested from a signed key, the message serializer can
+    # turn the format into a String (e.g. the JSON serializer). The String form
+    # must resolve to the same record instead of being re-processed.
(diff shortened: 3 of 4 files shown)


marcosdemelo:
Confirming this in production on Rails 8.1.3.1 (`load_defaults 8.1`, `variant_processor = :vips`, `track_variants` on), with some numbers on the impact.

**Where the type is lost.** Named variants keep `format: :webp` as a Symbol, and Active Job preserves it, so `TransformJob` and `blob.variant(:name).processed` both write the Symbol digest. The signed `variation_key` in the variant URL goes through `ActiveStorage.verifier`, which uses the JSON message serializer since 7.1, so `Variation.decode` hands back `format: "webp"`. `Representations::BaseController#set_representation` then looks up a digest nothing has ever written and calls `process`. Same for the seven `preprocessed: true` variants we declare:

```ruby
declared = ActiveStorage::Variation.new(resize_to_limit: [150, 150], format: :webp).default_to(format: :webp)
declared.digest                                          # => "FVZpDw4C2OgQx2/or1Yr6DHCdxE="
ActiveStorage::Variation.decode(declared.key).digest     # => "seUQ4LAcjFnxn7I733lE3OYL76A="
```

**Impact.** Every image opened in a browser is transformed twice: once by the job, once inside a web process serving the first redirect. Counting `active_storage_variant_records` for one attachment with 13,268 images: 13,267 have the Symbol digest written by the job, and 7,701 of those also have the String digest written by the redirect controller. Across all our variants, about 36% of the table (13k rows, and as many duplicate files on S3) is the second copy. Per-request memory profiling showed those redirects at 1.3 to 1.4 s each, adding 25 to 29 MB of RSS to the web process every time. That native libvips memory is never returned, so this alone explained 40 to 50% of the daily RSS growth of our web processes.

**On the fix.** Normalizing to a UTF-8 String is the right direction for us. We verified that declaring `format: "webp"` in the model today produces the same digest this PR produces after decoding, so apps that switch to strings as a workaround keep their records valid once this is merged. Normalizing to a Symbol instead would break the String records the redirect controller has been writing for every app on 7.1+.

One note on #58473: key order also matters to the digest, but it did not bite us here because `default_to` reorders both paths the same way.

PR #57742, opened 2026-06-15
Introduce `ActiveJob::DeserializationError::RecordNotFound`

rosa:
### Motivation / Background

We've run into a problem with one of our email delivery jobs for our HEY email service a couple of times, where the job is discarded due to a transient networking issue when connecting to our DB.
```
Discarded Entry::DelayedDeliveryJob (Job ID: c690f56a-1bb9-412f-be1a-f1dbc37b801c) 
due to a ActiveJob::DeserializationError (Error while trying to deserialize arguments: 
Trilogy::QueryError: trilogy_read_row: TRILOGY_TRUNCATED_PACKET).
```

We have alerts for pending deliveries, so we've located those, and re-enqueued a new job for the record that failed to be deserialised due to the `TRILOGY_TRUNCATED_PACKET` error. 

For this job in particular, we use 
```ruby
  discard_on ActiveJob::DeserializationError
```
because the underlying record is often deleted before the job is executed, as this powers the _Undo send_ feature in HEY, so that's a legit case. We don't want to discard the job when a transient networking error happens, though; we just want to retry normally. However, `ActiveJob::DeserializationError` wraps _any_ error that happens during deserialization, not just missing records, even though the `ApplicationJob` template suggests this is the only reason and encourages you to discard on that exception with this comment: 

```ruby
# Most jobs are safe to ignore if the underlying records are no longer available
# discard_on ActiveJob::DeserializationError
```

Searching online about previous discussions on this, I came across this post by @gravitystorm in the Ruby on Rails forums: [Is `discard_on ActiveJob::DeserializationError` a footgun?](https://discuss.rubyonrails.org/t/is-discard-on-activejob-deserializationerror-a-footgun/88442), where he describes the situation perfectly:

> So if you enable `discard_on ActiveJob::DeserializationError` and you have a short network outage, all your legitimate pending jobs could be instantly and silently discarded. Which is probably not what you want?

This PR is a possible solution to this, backwards-compatible and without breaking changes. 

### Detail

This PR introduces a new  `ActiveJob::DeserializationError::RecordNotFound` exception. This new exception extends `ActiveJob::DeserializationError`  but is raised only when deserialization fails because a record is missing. It uses the new `GlobalID::Locator.fetch` introduced in https://github.com/rails/globalid/pull/206. 

In this way, existing handlers for `ActiveJob::DeserializationError` will continue to work in the same way, but jobs can now instead handle `ActiveJob::DeserializationError::RecordNotFound` to ensure any other kind of issue while deserialising jobs doesn't result in the job being discarded:

```ruby
discard_on ActiveJob::DeserializationError::RecordNotFound
```

**Note**: jobs that set `discard_on ActiveRecord::RecordNotFound`, which could be raised both during deserialization or within `perform`, would now get  `ActiveJob::DeserializationError::RecordNotFound` for missing records during deserialization, so they would need to discard both errors if that's what they actually need: 
```ruby
discard_on ActiveRecord::RecordNotFound, ActiveJob::DeserializationError::RecordNotFound
```

### Checklist

### DIFF
--- a/activejob/lib/active_job/arguments.rb
+++ b/activejob/lib/active_job/arguments.rb
@@ -12,6 +12,24 @@ module ActiveJob
       super("Error while trying to deserialize arguments: #{$!.message}")
       set_backtrace $!.backtrace
     end
+
+    # Raised when arguments couldn't be deserialized because a record they
+    # reference through a Global ID no longer exists, most likely because it
+    # was deleted after the job was enqueued.
+    #
+    # Jobs whose arguments may legitimately reference deleted records can
+    # discard this error:
+    #
+    #   class SearchIndexingJob < ApplicationJob
+    #     discard_on ActiveJob::DeserializationError::RecordNotFound
+    #   end
+    #
+    # Unlike its parent class, it isn't raised for other errors that may occur
+    # while deserializing arguments, such as transient database connectivity
+    # failures, where the referenced record may still exist and a retry could
+    # succeed.
+    class RecordNotFound < DeserializationError
+    end
   end
 
   # Raised when an unsupported argument type is set as a job argument. We
@@ -81,6 +99,8 @@ module ActiveJob
     # GlobalID.
     def deserialize(arguments)
       arguments.map { |argument| deserialize_argument(argument) }
+    rescue GlobalID::Locator::RecordNotFound
+      raise DeserializationError::RecordNotFound
     rescue
       raise DeserializationError
     end
@@ -130,7 +150,7 @@ module ActiveJob
       end
 
       def deserialize_global_id(hash)
-        GlobalID::Locator.locate hash[GLOBALID_KEY]
+        GlobalID::Locator.fetch hash[GLOBALID_KEY]
       end
 
       def custom_serialized?(hash)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,7 +47,7 @@ PATH
       rails-html-sanitizer (~> 1.7)
     activejob (8.2.0.alpha)
       activesupport (= 8.2.0.alpha)
-      globalid (>= 0.3.6)
+      globalid (>= 1.4.0)
     activemodel (8.2.0.alpha)
       activesupport (= 8.2.0.alpha)
     activerecord (8.2.0.alpha)
@@ -221,7 +221,7 @@ GEM
     fugit (1.11.1)
       et-orbi (~> 1, >= 1.2.11)
       raabro (~> 1.4)
-    globalid (1.2.1)
+    globalid (1.4.0)
       activesupport (>= 6.1)
     google-apis-core (0.15.1)
       addressable (~> 2.5, >= 2.5.1)
@@ -830,7 +830,7 @@ CHECKSUMS
   ffi (1.17.1-x86_64-linux-gnu) sha256=8c0ade2a5d19f3672bccfe3b58e016ae5f159e3e2e741c856db87fcf07c903d0
   ffi-compiler (1.3.2) sha256=a94f3d81d12caf5c5d4ecf13980a70d0aeaa72268f3b9cc13358bcc6509184a0
   fugit (1.11.1) sha256=e89485e7be22226d8e9c6da411664d0660284b4b1c08cacb540f505907869868
-  globalid (1.2.1) sha256=70bf76711871f843dbba72beb8613229a49429d1866828476f9c9d6ccc327ce9
+  globalid (1.4.0) sha256=037f12fbf1d9d7a014d501c2d5c77356fd4ddd96d7a7991d6700bba96706f427
   google-apis-core (0.15.1) sha256=91484122791af5b2f3d3f4297912748febe2b5d704d04ad54cbf5df87339a71a
   google-apis-iamcredentials_v1 (0.22.0) sha256=38c1de913d541802398841a124bb5592f1bae71119b5aede5369647eee7a3dad
   google-apis-storage_v1 (0.49.0) sha256=48b45e448fbfb6dda538be4a77ec955009ab7aaaf77aa660459e4691a7187ca2
--- a/actionmailbox/app/jobs/action_mailbox/incineration_job.rb
+++ b/actionmailbox/app/jobs/action_mailbox/incineration_job.rb
@@ -12,7 +12,7 @@ module ActionMailbox
   class IncinerationJob < ActiveJob::Base
     queue_as { ActionMailbox.queues[:incineration] }
 
-    discard_on ActiveRecord::RecordNotFound
+    discard_on ActiveRecord::RecordNotFound, ActiveJob::DeserializationError::RecordNotFound
 
     def self.schedule(inbound_email)
       set(wait: ActionMailbox.incinerate_after).perform_later(inbound_email)
--- a/actionmailer/test/message_delivery_test.rb
+++ b/actionmailer/test/message_delivery_test.rb
@@ -238,6 +238,12 @@ class MessageDeliveryTest < ActiveSupport::TestCase
       raise "boom, missing find"
     end
 
+    # GlobalID::Locator.fetch, used during argument deserialization, looks
+    # records up through a non-raising finder, so blow up here too.
+    def self.where(id:)
+      raise "boom, missing find"
+    end
+
     attr_reader :id
     def initialize(id = 1)
       @id = id
@@ -256,7 +262,8 @@ class MessageDeliveryTest < ActiveSupport::TestCase
     # on the mailer class.
     assert_nothing_raised { message.deliver_later }
     assert_equal DelayedMailer, DelayedMailer.last_rescue_from_instance
-    assert_equal "Error while trying to deserialize arguments: boom, missing find", DelayedMailer.last_error.message
+    assert_match "Error while trying to deserialize arguments:", DelayedMailer.last_error.message
+    assert_match "boom, missing find", DelayedMailer.last_error.message
   end
 
   test "allows for keyword arguments" do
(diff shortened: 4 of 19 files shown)


activejob/lib/active_job/arguments.rb
```diff
@@ -12,6 +13,23 @@ def initialize # :nodoc:
...
+    # discard this error:
+    #
+    #   class SearchIndexingJob < ApplicationJob
+    #     discard_on ActiveJob::DeserializationError::RecordNotFound
+    #   end
+    #
+    # Unlike its parent class, it isn't raised for other errors that may
+    # occur while deserializing arguments, such as transient database
+    # connectivity failures, where the referenced record may still exist,
+    class RecordNotFound < DeserializationError
```
byroot, line 31: While this make sense to preserve backward compatibility, that doesn't solve the footgun.

byroot (commented):
I think taking a block might be easier. Unfortunately `discard_on` already takes a block and doesn't care about the return value.

Perhaps:

```ruby
discard_on ActiveJob::DeserializationError, unless: (_job, error) -> { !error.cause.is_a?(ActiveRecord::RecordNotFound) }
```

activejob/lib/active_job/arguments.rb
rosa: Right, well, it does for people who move their `discard_on` declarations to use this new exception, and for people starting out new. 

My first take on this was different and changed the behaviour of `DeserializationError` itself, making it be only about missing records. This solves the footgun for everyone, but it's also misleading and arguably wrong, because a _deserialization_ error is an error deserializing, it doesn't need to be a missing record. The current behaviour is correct, despite being a footgun.

rosa:
> think taking a block might be easier. Unfortunately `discard_on` already takes a block and doesn't care about the return value.
> 
> Perhaps:
> 
> ```ruby
> discard_on ActiveJob::DeserializationError, unless: (_job, error) -> { !error.cause.is_a?(ActiveRecord::RecordNotFound) }
> ```

Do you suggest adding `unless` and `if` arguments, callback-style, to `discard_on`? And then propose using this in the template instead? I'm good with that, though it has the same problem as the solution proposed here, no? Keeps backwards compatibility but doesn't solve the footgun. 

I like this approach because it doesn't require that Active Record registers any exceptions, and I'm ok with just solving the problem going forward or for people who actively switch to this.

byroot:
Right sorry, I now realize I slightly misinterpreted the premises. I now get that the end goal is:

```ruby
discard_on ActiveJob::DeserializationError::RecordNotFound
```

That said, I still do think we should make it easy to match on the cause

```ruby
discard_on cause: ActiveRecord::RecordNotFound
# or
discard_on ActiveRecord::RecordNotFound, cause: true
```

Or even just match the cause by default?

```ruby
discard_on ActiveRecord::RecordNotFound
```

But perhaps you specifically want to also ensure it's  a deserialization error too?

```ruby
discard_on ActiveJob::DeserializationError, cause: ActiveRecord::RecordNotFound
```

rosa:
You need to make sure it's a `ActiveJob::DeserializationError` because an `ActiveRecord::RecordNotFound` exception could be raised within the regular job (in `perform`), not just on argument deserialization 🤔 So it needs to be a specific exception for that. I think having a `cause` option would make sense as well!

gravitystorm:
I like the original suggestion of something like `ActiveJob::DeserializationError::RecordNotFound`, since it would be the same code in the job, no matter which backend ActiveJob is configured with.

If it needs an explicit, backend-specific, exception specified (e.g. `ActiveRecord::RecordNotFound` in the `cause`) then the documentation becomes more complex too - since the job code would need to become tied to the backend chosen.

byroot:
> I like the original suggestion

Dunno. I understand where it's coming from, but to me it's extra global configuration because `discard_on` isn't quite flexible enough.

I tend to prefer a little callback / local configuration over global configuration you need to know about.

That being said, GlobalID arguments pointing to now deleted records is quite a common issue and it definitely would be worth having an happy path for it, I'm just not entirely convinced that is it.

rosa:
Honestly, I'm ok with any solution as long as we can fix this. The `cause` argument seemed fine (though a bit obscure perhaps? it's not something used in other exception handlers in Rails I don't think 🤔), and so did the `if`/`unless` options. The original proposed solution includes new global configuration but Rails users don't need to worry about it, only alternative ORM authors would need to, to register their own exception there, so it's something that almost nobody needs to know about.

byroot:
I wonder if we even need a configuration thought. I need to check the code, but I think we know when we're looking up a global ID. We could just base it off that.

byroot:
Yep: https://github.com/rails/rails/blob/2987d46f6d6c5bd621d9f2d0a591b756ca404760/activejob/lib/active_job/arguments.rb#L85

Pretty sure we can achieve the same thing with no configuration and just a single line change.

rosa:
Hmm... Do you mean perhaps calling `GlobalID::Locator.locate_many` instead of `GlobalId::Locator#locate`, to avoid the `ActiveRecord::RecordNotFound` raised by `model_class#find`, and then handling the case that no results are returned? Here: 
https://github.com/rails/rails/blob/2987d46f6d6c5bd621d9f2d0a591b756ca404760/activejob/lib/active_job/arguments.rb#L132-L134

That could work, combined with the new `ActiveJob::DeserializationError::RecordNotFound` exception, though it'd be more than one line change, but yes.

rosa:
To be clear, we can't just raise that exception when we're locating a global ID, because that has the same problem as the current approach: locating a global ID might raise because the record doesn't exist (`ActiveRecord::RecordNotFound`) or because there's any other problem with the DB, usually transient (eg. the `Error while trying to deserialize arguments: 
Trilogy::QueryError: trilogy_read_row: TRILOGY_TRUNCATED_PACKET` that prompted this PR).

byroot:
Indeed, but I'd prefer if we had GlobalID figure out the difference between the record not being found and transient errors.

It may end up being not super different from your current patch, I think it'd be cleaner at that layer.

rosa:
> Indeed, but I'd prefer if we had GlobalID figure out the difference between the record not being found and transient errors.

I think, from the GlobalID perspective, the difference is clear. `GlobalId::Locator#locate` already raises different exceptions when the record is not found vs. transient errors. It's Active Job the one that wraps all of them under the same `DeserializationError` exception. So, GlobalID does the right thing already. I'm not sure how we could change GlobalID to fix this. It seems we need to change how we use GlobalID, rather, thus the change to `locate_many`?

byroot:
> So, GlobalID does the right thing already.

I wouldn't say that. GlobalID calls an ActiveRecord-Like API and let whichever underlying exception bubble up, which in my book is a bad thing, because now you have no clue what `GlobalID.locate` may raise, it's a leaky abstraction.

It would be better if `GlobalID.locate` would raise one of two errors:

  - That ID is no longer valid, won't ever work
  - Something happened, try again later.

Which is essentially what you are doing here in AJ, but I think it would be better done at the GlobalID layer.

rosa:
Got it, so, you propose that `GlobalID.locate` raises two own errors, such as 
```
GlobalID::RecordNotFound
GlobalID::RecordFindException # or GlobalID::RecordLocationError or something else like this
```
The problem is that this is a big breaking change, as I assume people everywhere are already handling `ActiveRecord::RecordNotFound` after calling `locate`, all over the place 🤔

byroot:
> so, you propose that `GlobalID.locate` raises two own errors

Yes, I think it would be much better design.

> The problem is that this is a big breaking change,

Indeed, but I don't think it's that much of a problem. We can either bump the version to a new major, and there isn't hat many globalid consumers: https://rubygems.org/gems/globalid/reverse_dependencies, or just make that a new, purely additive API.

rosa:
Cool! I'll do that, will probably go for the new, purely additive API, if I can think of a nice name 🤔 Thanks!

rosa:
Finally opened the PR in GlobalID: https://github.com/rails/globalid/pull/206

williantenfen:
@rosa Hi, I was following this for learning more about Rails internals. If you dont mind I had a PR https://github.com/rosa/rails/pull/7 for your branch with changes to use new globalID API. Not sure if is right... 
feel free to close it if not.. 🙈 
sorry If I'm getting on the way..
thanks 👋

rosa:
Hey @williantenfen thanks! That looks mostly good 😊

williantenfen:
> Hey @williantenfen thanks! That looks mostly good 😊

great, appreciate any comments I could work on to improve 😊 - if not thats fine as well, I will just keep following to see how this ends... thanks...

rosa:
@byroot, this is ready I think. There's a small wrinkle I hadn't realised until switching to the new version of `globalid` and the new method: some jobs might have been using `discard_on ActiveRecord::RecordNotFound` (like jobs in Rails's own frameworks, like Active Storage or Action Mailbox), which would cover both deserialization and `perform`. Since we wouldn't be using Active Record's `find` under the hood anymore, `ActiveRecord::RecordNotFound` won't be raised for records not found during deserialization, so jobs would start to fail with `ActiveJob::DeserializationError::RecordNotFound` instead of being discarded. This is a breaking change, or at least a change in behaviour. I think it's a good change, though, because discarding on `ActiveRecord::RecordNotFound` is, in many cases, not what you want, but it's something to point out in any case. Maybe in the upgrade guide, besides the CHANGELOG?

byroot:
> in many cases, not what you want, but it's something to point out in any case.

Yeah, I think it's a positive change, as you very well may want to distinguish both.

I think the breakage is limited as essentially it will just create some noise until you update the `discard_on`, not actually breaking a functionality. So I'm fine with just pointing it out in the changelog.

rosa:
Perfect! Then this should be ready.

PR #57743, opened 2026-06-15
Fix `ActiveSupport::InheritableOptions#to_h` for nested inheritable options

andrewn617:
### Motivation / Background

If you have an `InheritableOptions` containing another `InheritableOptions` as its parent (like you get from the `inheritable_copy` method), then calling `to_h` returns the options object, not a hash.

### Detail

Call `.to_h` on the parent before we merge it.

### Checklist


fatkodima:
We don't usually add changelogs for minor bug fixes, so please remove it and will be good to merge.

PR #57752, opened 2026-06-16
More shareable constants

andrewn617:
### Motivation / Background

Working on making the framework Ractor-safe. We need to freeze the values in constants. These are just a few more that I encountered while working with a test app.

### Detail

These are hashes with have mutable values / default procs that we need to freeze with `freeze` / `Ractor.shareable_`.

### Checklist

### DIFF
--- a/actionview/lib/action_view/helpers/tag_helper.rb
+++ b/actionview/lib/action_view/helpers/tag_helper.rb
@@ -41,7 +41,7 @@ module ActionView
       TAG_TYPES.merge! CLASS_PREFIXES.index_with(:class)
       TAG_TYPES.freeze
 
-      PRE_CONTENT_STRINGS             = Hash.new { "" }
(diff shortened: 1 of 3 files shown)


actionview/lib/action_view/helpers/tag_helper.rb
```diff
@@ -41,7 +41,7 @@ module TagHelper
       TAG_TYPES.merge! CLASS_PREFIXES.index_with(:class)
       TAG_TYPES.freeze
 
-      PRE_CONTENT_STRINGS             = Hash.new { "" }
+      PRE_CONTENT_STRINGS             = Hash.new(&ActiveSupport::Ractors.shareable_proc { "" })
```
etiennebarrie, line 44: Since `""` is frozen and always returned on missing keys, and the default_proc doesn't mutate the Hash, we should be able to use `default` instead of `default_proc`. https://github.com/rails/rails/pull/57764

PR #57753, opened 2026-06-16
Make ExecutionContext.after_change ractor safe

andrewn617:
### Motivation / Background

Working on making the framework Ractor safe. `ActiveSupport::ExecutionContext.after_change_callbacks` are just an array of procs and thus not Ractor safe.

### Detail

This is not widely used (it is no-doced). The procs are just `called` not `instanced_execed`, so the caller that creates the proc needs to be responsible for making it Ractor shareable. Then we just dup, add, freeze and reassign the array when adding a new one.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/execution_context.rb
+++ b/activesupport/lib/active_support/execution_context.rb
@@ -32,7 +32,7 @@ module ActiveSupport
       end
     end
 
-    @after_change_callbacks = []
+    @after_change_callbacks = [].freeze
 
     # Execution context nesting should only legitimately happen during test
     # because the test case itself is wrapped in an executor, and it might call
@@ -42,10 +42,10 @@ module ActiveSupport
     @nestable = false
 
     class << self
-      attr_accessor :nestable
+      attr_accessor :nestable, :after_change_callbacks
 
       def after_change(&block)
-        @after_change_callbacks << block
+        @after_change_callbacks = [*@after_change_callbacks, block].freeze
       end
 
       # Updates the execution context. If a block is given, it resets the provided keys to their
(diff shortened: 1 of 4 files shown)


activesupport/lib/active_support/execution_context.rb
```diff
@@ -42,10 +42,10 @@ def flush
     @nestable = false
 
     class << self
-      attr_accessor :nestable
+      attr_accessor :nestable, :after_change_callbacks
 
       def after_change(&block)
-        @after_change_callbacks << block
+        @after_change_callbacks = (@after_change_callbacks | [block]).freeze
```
gmcgibbon: Using `|` is probably fine, but is technically a behaviour change. We can likely just do `@after_change_callbacks = (@after_change_callbacks + [block]).freeze` right?

gmcgibbon: Actually, after looking at https://github.com/rails/rails/pull/57754, I think we should try casting the proc at the method that stores the proc too. Even though this is private API and only used once, I think `Ractors.shareable_proc` is better here (no need to try to cast it).

byroot: > Using | is probably fine, but is technically a behaviour change. 

It's not fine though, it's a set operation so way more costly, but more importantly it's impossible `block` is already present in the array, so it's wasteful.

I'm also worried about boot time regression with all these change from mutation to concatenation.

gmcgibbon: Good point. Ok, let's revert use of `|` then. 👍

gmcgibbon: @andrewn617 I merged https://github.com/rails/rails/pull/57756, so please also revert the `|` there too in this PR. They're symbols, but the added perf considerations here apply.

andrewn617: Sorry about that. Thanks for fixing this and the others @gmcgibbon

PR #57754, opened 2026-06-16
Make AbstractController::Caching.view_cache_dependency ractor safe

andrewn617:
### Motivation / Background

Make `AbstraceController::Caching.view_cache_dependency` ractor safe

### Detail

When adding a new dependency, we dup, add, freeze and reassign the underlying `_view_cache_dependencies` array. Following pattern in https://github.com/rails/rails/pull/57743

These a procs, but they are instance execed, so we can call `ActiveSupport::Ractors.try_shareable_proc` on them to make them Ractor shareable for apps opting in to Ractors.

### Checklist

### DIFF
--- a/actionpack/lib/abstract_controller/caching.rb
+++ b/actionpack/lib/abstract_controller/caching.rb
@@ -44,13 +44,15 @@ module AbstractController
       delegate :enable_fragment_cache_logging, :enable_fragment_cache_logging=, to: :config
       self.enable_fragment_cache_logging = false
 
-      class_attribute :_view_cache_dependencies, default: []
+      class_attribute :_view_cache_dependencies, default: [].freeze
       helper_method :view_cache_dependencies if respond_to?(:helper_method)
     end
 
     module ClassMethods
       def view_cache_dependency(&dependency)
-        self._view_cache_dependencies += [dependency]
+        self._view_cache_dependencies = [
+          *_view_cache_dependencies, ActiveSupport::Ractors.try_shareable_proc(dependency)
+        ].freeze
       end
     end
 
--- a/actionpack/test/controller/caching_test.rb
+++ b/actionpack/test/controller/caching_test.rb
@@ -3,6 +3,8 @@
 require "fileutils"
 require "abstract_unit"
 require "lib/controller/fake_models"
+require "active_support/testing/ractors_assertions"
+require "active_support/core_ext/object/with"
 
 CACHE_DIR = "test_cache"
 # Don't change '/../temp/' cavalierly or you might hose something you don't want hosed
@@ -343,6 +345,8 @@ class CacheHelperOutputBufferTest < ActionController::TestCase
 end
 
 class ViewCacheDependencyTest < ActionController::TestCase
+  include ActiveSupport::Testing::RactorsAssertions
+
   class NoDependenciesController < ActionController::Base
   end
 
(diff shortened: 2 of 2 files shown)


actionpack/lib/abstract_controller/caching.rb
```diff
@@ -44,13 +44,13 @@ def cache_configured?
...
 
-      class_attribute :_view_cache_dependencies, default: []
+      class_attribute :_view_cache_dependencies, default: [].freeze
       helper_method :view_cache_dependencies if respond_to?(:helper_method)
     end
 
     module ClassMethods
       def view_cache_dependency(&dependency)
-        self._view_cache_dependencies += [dependency]
+        self._view_cache_dependencies = (_view_cache_dependencies | [ActiveSupport::Ractors.try_shareable_proc(dependency)]).freeze
```
byroot: Same here, why do we need a set operation when it wasn't needed before?

byroot: I'm not convinced we can make these proc shareable without breaking at least some users.

Just because `Ractor.sharable_proc` doesn't immediately raise doesn't mean it returns a working proc, e.g.:

```ruby
class Foo
  def foo
    p :ok
  end

  def callback
    -> { foo }
  end
end

proc = Foo.new.callback
p proc.call
p Ractor.shareable_proc(&proc).call
```

Given this is public API, this change may break some apps.

byroot: Also cc @jhawthorn @tenderlove on this one.

I'm not sure we can unconditionally make procs shareables across Rails. We might need to introduce yet another config of some sort.

tenderlove: Ya, I agree it could break apps. We should add a config, but we just need to read it in `ActiveSupport::Ractors.try_shareable_proc` right?  Eventually we could default to true for new apps

byroot: > we just need to read it in `ActiveSupport::Ractors.try_shareable_proc` right?

Yes.

> Eventually we could default to true for new apps

Perhaps, if things turn out well, but the UX of shareable procs is still not ideal. From my example above:

```ruby
/tmp/rac.rb:7:in 'block in Foo#callback': undefined local variable or method 'foo' for nil (NameError)

    -> { foo }
```

This might make you feel like you're taking crazy pills if you don't know the details of Ractor shareable procs etc.

matthewd: For the general discussion:

I feel like I would like a more-strict version of `shareable_proc` that rejected any call (or mention?) on `self` 🤔 (but maybe that already came up and was dismissed upstream... or is more impossible than I'm picturing?)

That aside, we could at least do better than that example by injecting a poison object instead of nil. Not perfect, but we'd have control of the exception _most_ of the time.

---

But for this instance, and a great number of others:

No, my understanding is that if you get past `shareable_proc`, you're fine. Because: https://github.com/rails/rails/blob/fa8f0812160665bff083a089d2bb2fc1817ea03e/actionpack/lib/abstract_controller/caching.rb#L58

byroot: > But for this instance, and a great number of others:

Yeah, perhaps not the best example, but I just wanted to point out that making an user provided Proc shareable may very well be a breaking change.

gmcgibbon: try_shareable_proc has unshareable_proc_action which can be set to warn or raise. We won't be raising by default, at least initially. I've revert the set. Thanks!

byroot: I think you missed my argument / demo script.

Just because `Ractor.shareable_proc` doesn't raise doesn't mean it "worked". It may very well leave the proc is in broken state, and cause it to raise latter when called.

andrewn617: Sorry for the slow response, I've been away for a few days.

Yeah, we certainly cannot make user supplied procs shareable if we plan to `call` them. For example for the backtrace cleaner I did not change the `add_filter` api to make the proc shareable, I just made the default filters we add in the framework shareable, and user will need to do the same for custom filters. Personally I don't think another config will help, imo raising only when the block eventually gets called is a footgun. I think user will have to be responsible for making those procs shareable themselves, at least for now, perhaps we can do something more sophisticated as the feature evolves..

But for things like this that we call with `instance_exec` we can make them shareable, which is why we introduced `try_shareable_proc` and the config to raise, warn or opt out.

PR #57757, opened 2026-06-16
Make InheritableOptions ractor safe

andrewn617:
### Motivation / Background

I am working on making it possible to make a Rails application Ractor safe, but some engines contain inheritable options which currently can't be made Ractor safe. This PR fixes that.

### Detail

Inheritable Options have a default proc which accesses the parent hash in case of a key miss. This proc is hard to make ractor safe because it is stored on the options, but needs to reference the options' parent ivar. So we can't make the proc shareable without making the options shareable, and we can't make the options shareable without making the proc shareable. I don't see a good way around that at the moment. One idea I had was to make a new default proc with `self: @parent` but that requires `Ractor.make_shareable`ing the parent, which seems wrong to do in the `freeze` method as it may have unsafe values. We can't raise an isolation error when calling `freeze` directly.

To get around this, we can drop the default proc when freezing and reverse merge the parent hash into the inheritable options. That way it still quacks like an inheritable options but it no longer needs to fallback to the parent.

This also required me to change the way we implement `overriden?` since after freezing the options will have all the keys the parent has, thus making it appear that all keys are overriden. Instead, I decided to keep track of the keys as they're added.

### Checklist

### DIFF
--- a/activesupport/lib/active_support/ordered_options.rb
+++ b/activesupport/lib/active_support/ordered_options.rb
@@ -127,6 +127,10 @@ module ActiveSupport
       super || @parent.key?(key)
     end
 
+    def keys
+      @parent.keys | super
+    end
+
     def overridden?(key)
       !!(@parent && @parent.key?(key) && own_key?(key.to_sym))
     end
--- a/activesupport/lib/active_support/ordered_options.rb
+++ b/activesupport/lib/active_support/ordered_options.rb
@@ -100,6 +100,16 @@ module ActiveSupport
       end
     end
 
+    def freeze
+      return self if frozen?
+
+      @own_keys = own_keys.dup.freeze
+      self.default_proc = nil
+      @parent.freeze
+      replace(to_h)
+      super
+    end
+
     def to_h
       @parent.to_h.merge(self)
     end
@@ -120,12 +130,12 @@ module ActiveSupport
       pp.pp_hash(to_h)
     end
 
-    alias_method :own_key?, :key?
-    private :own_key?
-
     def key?(key)
       super || @parent.key?(key)
     end
+
+    alias_method :_own_keys, :keys
+    private :_own_keys
 
     def keys
       @parent.keys | super
@@ -147,5 +157,14 @@ module ActiveSupport
       to_h.each(&block)
       self
     end
+
+    private
+      def own_key?(key)
+        own_keys.include?(key.to_sym)
+      end
+
+      def own_keys
+        @own_keys || _own_keys
+      end
   end
 end
--- a/activesupport/test/ordered_options_test.rb
+++ b/activesupport/test/ordered_options_test.rb
@@ -3,8 +3,11 @@
 require "pp"
 require_relative "abstract_unit"
 require "active_support/ordered_options"
+require "active_support/testing/ractors_assertions"
 
 class OrderedOptionsTest < ActiveSupport::TestCase
+  include ActiveSupport::Testing::RactorsAssertions
+
   def test_usage
     a = ActiveSupport::OrderedOptions.new
 
@@ -354,5 +357,29 @@ class OrderedOptionsTest < ActiveSupport::TestCase
     io = StringIO.new
     PP.pp(object, io)
     assert_equal({ one: "first value", two: "second value", three: "third value" }.inspect, io.string.strip)
+  end
+
+  def test_inheritable_options_ractor_shareable
+    object = ActiveSupport::InheritableOptions.new(one: "first value")
+    object[:two] = "second value"
+    object["three"] = "third value"
+
+    assert_ractor_make_shareable(object)
+
+    assert_equal "first value", object[:one]
+    assert_equal "second value", object[:two]
+    assert_equal "third value", object[:three]
+  end
+
+  def test_overridden_works_when_frozen
+    object = ActiveSupport::InheritableOptions.new(one: "first value", two: "second value")
+    object[:two] = "second value"
+    object["three"] = "third value"
+
+    object.freeze
+
+    assert object.overridden?(:two)
+    assert_not object.overridden?(:one)
+    assert_not object.overridden?(:three)
   end
 end
--- a/activesupport/test/ordered_options_test.rb
+++ b/activesupport/test/ordered_options_test.rb
@@ -250,6 +250,14 @@ class OrderedOptionsTest < ActiveSupport::TestCase
     assert_not object.key?(:four)
   end
 
+  def test_inheritable_options_keys
+    object = ActiveSupport::InheritableOptions.new(one: "first value")
+    object[:two] = "second value"
+    object["three"] = "third value"
+
+    assert_equal [:one, :two, :three], object.keys
+  end
+
   def test_inheritable_options_overridden
     object = ActiveSupport::InheritableOptions.new(one: "first value", two: "second value", three: "third value")
     object["one"] = "first value override"


activesupport/lib/active_support/ordered_options.rb
```diff
@@ -88,6 +89,7 @@ def inspect
   #   h.boy  # => 'John'
   class InheritableOptions < OrderedOptions
     def initialize(parent = nil)
+      @own_keys = []
```
byroot: Probably should be a set.

activesupport/lib/active_support/ordered_options.rb
```diff
@@ -100,8 +102,21 @@ def initialize(parent = nil)
...
+    def freeze
+      self.with_defaults!(@parent.to_h)
+      self.default_proc = nil
+      @parent.freeze
+      @own_keys.freeze
+      super
+    end
+
+    def []=(k, v) # :nodoc:
+      @own_keys |= [k.to_sym]
```
byroot: Once it's made a set:
```suggestion
      @own_keys  << k.to_sym
```

byroot (commented):
I don't like that we're increasing baseline memory usage (keeping `@own_keys`) to allow a feature the vast majority won't use, at least not before a while.

We could probably delay the building of `@own_key` to when `freeze` is called, with some minimal extra complexity.

But then, there's the `default_proc` thing, which really make me think `.freeze` doesn't always mean we need to be shareable, it's quite unfortunate that we can't know if we're freezing just to avoid muation or to be Ractor compatible.

activesupport/lib/active_support/ordered_options.rb
```diff
@@ -100,8 +102,21 @@ def initialize(parent = nil)
...
 
+    def freeze
+      self.with_defaults!(@parent.to_h)
+      self.default_proc = nil
+      @parent.freeze
+      @own_keys.freeze
+      super
+    end
+
+    def []=(k, v) # :nodoc:
```
etiennebarrie: Overriding `[]=` is not enough to track overrides:

```ruby
options = ActiveSupport::InheritableOptions.new({foo: 42, bar: 21})
options.merge!(foo: 21)
options.overridden? :foo
# main: => true
# this branch: => false
```

I think we should track the keys when we're freezing, it's the only point we know the hash's not going to be mutated anymore.

activesupport/lib/active_support/ordered_options.rb
```diff
@@ -100,8 +102,21 @@ def initialize(parent = nil)
       end
     end
 
+    def freeze
+      self.with_defaults!(@parent.to_h)
```
etiennebarrie: `with_defaults!` requires the `core_ext` (as you correctly added), but I feel like we don't need this.

`with_defaults!(other)` is `replace(reverse_merge(other))`, which is `replace(other.merge(self))` but we already have `to_h` as `@parent.merge(self)`, so we could just use `replace(to_h)`.

There's a difference  if `@parent` is not a Hash depending on what it returns for `to_h` or `merge`, but that should be fixed in `to_h` if we think it's worth it.

activesupport/lib/active_support/ordered_options.rb
```diff
@@ -120,8 +135,9 @@ def pretty_print(pp)
       pp.pp_hash(to_h)
     end
 
-    alias_method :own_key?, :key?
-    private :own_key?
+    def own_key?(key)
```
etiennebarrie:
```suggestion
    private def own_key?(key)
```

activesupport/lib/active_support/ordered_options.rb
andrewn617: True. I already fixed that in https://github.com/rails/rails/pull/57743 so yeah let's do that

andrewn617:
@byroot Yeah I agree... It feels bad to leak Ractor concerns into the `freeze` method, I wish there was a way to hook into make_shareable but I think that was rejected in Ruby? If you think that or some other idea could be approved upstream I think we should push for it...

Or we could put a hook in the framework for things to make themselves shareable before we freeze the application, but I got (warranted IMO) pushback on that idea. There don't seem to be places that genuinely need that, and I can get requests running in ractors without it. But the flip side is in places like this we are adding ractor stuff to the freeze method.

activesupport/lib/active_support/ordered_options.rb
andrewn617: This works, but now I realize there is a bug on main:

```
options = InheritableOptions.new(foo: :bar)
options.key?(:foo) # => true
options.keys # => []
```

So I added a fix for that as well

byroot:
> f you think that or some other idea could be approved upstream I think we should push for it...

In the mean time, there might be something we could do on our side, something like:

```ruby
module AS::Ractors
  def make_shareable(obj, copy: false)
    if obj.respond_to?(:make_shareable)
      obj.make_shareable(copy: copy)
   else
      Ractor.make_shareable(...)
  end
end
```

This would allow classes to split their regular freeze from their `make_shareable`.

andrewn617:
That could work, but the question is where / when do we call that method? It would need to be before `Ractor.make_shareable(Rails.application)`, which is why I think we would need a hook.

I actually have the code to make the application shareable, this is the last open PR blocking that, so I will open that today and we could have a better discussion, since there are a few other cases like this to look at.

byroot:
> It would need to be before `Ractor.make_shareable(Rails.application)`, which is why I think we would need a hook.

Similarly, `Ractor.make_shareable(Rails.application)` would become `Rails.application.make_shareable`. Meaning you can hook all you want.

andrewn617:
@byroot I opened https://github.com/rails/rails/pull/57825 so you can look at the bigger picture of what's left to do to freeze the application, it isn't too much, those there is one similar case when freezing key generators.

We definitely could go with a hook, but I got push back on that from some folks when discussing which way to go. It's not strictly necessary, but I do share your concern about polluting objects freeze methods. Tho there aren't too many cases so far.

On the flip side, adding the hook will lead to needing to quite a bit of book keeping vs. just relying on recursively freezing.

byroot:
> I opened #57825 so you can look at the bigger picture of what's left to do to freeze the application

Thanks. Much appreciated.

> it isn't too much, those there is one similar case when freezing key generators.

I believe you, but I doubt this will be over once `Application` can be made Ractor shareable. There's a lot more global state a Rails app reference that would also need to be made shareable if you want to do anything useful from a Ractor.

andrewn617:
What if we did something like:

```
class Rails::Application
  def ractorize!
     @ractorizing = true
     ...
  end
end

class Object
  def freeze
     if Rails.application.ractorizing?
       object.make_shareable if object.respond_to? :make_shareable
     end

    super
  end
end
```

Then we don't have to keep track of all the objects that need special freezing called and can rely on recurisve freezing, but we still get a separate make shareable method on the object.

jhawthorn:
> But then, there's the `default_proc` thing, which really make me think `.freeze` doesn't always mean we need to be shareable, it's quite unfortunate that we can't know if we're freezing just to avoid muation or to be Ractor compatible.

I talked about this with @andrewn617 and then @ko1. I also initially felt that it would be nice to have a `on_make_shareable` callback or something like that to do the "Ractor-specifics" instead of general immutability.

The biggest problem (of several) we found with that is that it makes it very awkward if you `freeze` first, because then the object is immutable and you can't modify it to make it shareable.

```ruby
class InheritableOptions < Hash
  def freeze
    @parent.freeze
    super() # actually freezes
  end

  def on_make_shareable # proposed callback for make_shareable
    self.default_proc = nil
    freeze
  end
end
```

With this `Ractor.make_shareable(x)` works as we want it to, we're able to remove the default proc. But now `Ractor.make_shareable(x.freeze)` will raise `FrozenError: can't modify frozen Hash` because we can't remove the default proc from a frozen hash 😞.

This makes me think that the purpose of `freeze` is to put the object in a "final immutable state" that is prepared for any future operations that may need to be done. `Ractor.make_shareable` and sending to another Ractor _is_ one of those operations, so I do think it makes sense to do the necessary Ractor preparations in `freeze`.

PR #57764, opened 2026-06-17
Use Hash default instead of default proc

etiennebarrie:
Since the string is frozen and always returned on missing keys, and since the default_proc doesn't mutate the Hash, we can use a value default instead of a default proc.

Follow-up to #57752


skipkayhil:
Fixed in #57760

PR #57765, opened 2026-06-17
Validate `retry_on` `wait:` argument type at class definition

55728:
Follow-up to #57727, based on @p8's [suggestion](https://github.com/rails/rails/pull/57727#discussion_r3420741025).

### Summary

An unsupported `wait:` type (e.g. `retry_on SomeError, wait: "soon"`) currently raises a `RuntimeError` from `determine_delay` — inside the `rescue_from` handler, after a job has already failed. This replaces the original application exception with a confusing internal error.

This PR moves the type check up to `retry_on` itself, so the `ArgumentError` is raised when the job class is loaded rather than when the retry handler fires.

### Changes

The supported types mirror `determine_delay`'s `case` branches:

- `Integer`, `Float` — seconds
- `ActiveSupport::Duration` — e.g. `1.hour`
- `:polynomially_longer`
- `Proc` — e.g. `->(executions) { executions * 2 }`

The `else` branch in `determine_delay` is kept as a defensive fallback.

### DIFF
--- a/activejob/lib/active_job/exceptions.rb
+++ b/activejob/lib/active_job/exceptions.rb
@@ -64,6 +64,14 @@ module ActiveJob
       #    end
       #  end
       def retry_on(*exceptions, wait: 3.seconds, attempts: 5, queue: nil, priority: nil, jitter: JITTER_DEFAULT, report: false)
+        case wait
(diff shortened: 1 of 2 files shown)


fatkodima:
Makes sense, thanks.

PR #57638, opened 2026-06-09
Refactor Composite Primary Key implementation by using polymorphism

paracycle:
### Motivation / Background

This Pull Request has been created because the original Composite Primary Key work introduced a lot of brittle conditional logic scattered around the codebase.

### Detail

This Pull Request changes almost all of the trivial conditional CPK logic into polymorphic `ActiveRecord::Key` (and subclasses) to simplify the codebase and to make it more flexible.

### Checklist

Before submitting the PR make sure the following are checked:

* [x] This Pull Request is related to one change. Unrelated changes should be opened in separate PRs.
* [x] Commit message has a detailed description of what changed and why. If this PR fixes a related issue include it in the commit message. Ex: `[Fix #issue-number]`
* [x] Tests are added or updated if you fix a bug or add a feature.
* [x] CHANGELOG files are updated for the changed libraries if there is a behavior change or additional feature. Minor bug fixes and documentation changes should not be included.

PR #57773, opened 2026-06-18
Remove outdated documentation for overriding primary_key

ilianah:
After https://github.com/rails/rails/pull/57638 we can no longer simply override primary_key value as it needs to be in sync with primary_key_definition.


byroot (approved):
cc @rafaelfranca @paracycle any objections?

Technically that's a breaking change (that we experienced in production...), but I don't see how we could continue to support this pattern with the recent primary key refactoring.

PR #57784, opened 2026-06-19
Fix inverse matching for association composite primary keys

Korri:
[rails/rails#56664](https://github.com/rails/rails/pull/56664) fixed array-vs-array inverse matching, but this still missed associations that declare a composite `primary_key` while the owner class itself keeps scalar `id`. In that shape, the owner-side comparison in [`Association#matches_foreign_key?`](https://github.com/Korri/rails/blob/51c5ec91f978e4322fc9ff0604e6b6c123fc1ab1/activerecord/lib/active_record/associations/association.rb#L413-L431) needs the association's `active_record_primary_key`.

The regression is modeled in [`AssociationsTest#test_belongs_to_a_model_with_composite_association_primary_key_sets_inverse_of`](https://github.com/Korri/rails/blob/51c5ec91f978e4322fc9ff0604e6b6c123fc1ab1/activerecord/test/cases/associations_test.rb#L166-L175), mirroring the failure shape: child FK values like `[shop_id, order_id]` should match owner values `[shop_id, id]`.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,8 @@
+*   Fix inverse association matching for `has_many` and `has_one` associations
+    with association-specific composite primary keys.
+
+    *Hugo Vacher*
+
 *   Support polymorphic associations with custom primary keys through `:inverse_of`.
 
     When using polymorphic associations with `:inverse_of`, ActiveRecord now respects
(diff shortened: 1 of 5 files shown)


nvasilevski (approved):

nvasilevski:
Could you please squash commits into one? So it's mergeable right away when someone from the core/committers team reviews

PR #57801, opened 2026-06-21
Make Action Cable executor configurable

samuel-williams-shopify:
### Motivation / Background

Action Cable currently memoizes `ActionCable::Server::ThreadedExecutor` directly inside `ActionCable::Server::Base#executor`. That makes it difficult for servers or adapters with a different concurrency model to provide an executor that still satisfies Action Cable's small executor contract.

This adds a public configuration hook so applications can provide an executor factory instead of patching `@executor` directly.

### Detail

- Add `config.action_cable.executor`, defaulting to `ActionCable::Server::ThreadedExecutor`.
- Add `ActionCable::Server::ThreadedExecutor.call(server)` so the default executor can be used as a factory.
- Instantiate the memoized server executor via `config.executor.call(self)`.
- Cover the default executor and a custom executor factory in `ActionCable::Server::Base` tests.

The expected executor surface remains the existing small interface Action Cable already uses: `#post`, `#timer`, and `#shutdown`.

### Checklist

- [x] Tests pass: `ruby -Iactioncable/test actioncable/test/server/base_test.rb`
- [x] RuboCop passes for touched Ruby files

### DIFF
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Add `config.action_cable.executor` for configuring the Action Cable executor.
+
+    The configured object is called with the server instance and must return an
+    executor responding to `#post`, `#timer`, and `#shutdown`. This allows
+    applications and adapters to provide executor implementations that match
+    their concurrency model.
+
+    *Samuel Williams*
+
 *   Respect calls to `#reject` in `before_subscribe` callbacks.
 
     It doesn't call `#subscribed` if a `before_subscribe` callback calls `#reject`.
--- a/actioncable/lib/action_cable/server/base.rb
+++ b/actioncable/lib/action_cable/server/base.rb
@@ -8,6 +8,10 @@ module ActionCable
   module Server
     # A wrapper over ConcurrentRuby::ThreadPoolExecutor and Concurrent::TimerTask
     class ThreadedExecutor # :nodoc:
+      def self.call(server)
+        new(max_size: server.config.executor_pool_size, name: "streamer")
+      end
+
       def initialize(max_size: 10, name: "server")
         @executor = Concurrent::ThreadPoolExecutor.new(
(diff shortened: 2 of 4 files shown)


samuel-williams-shopify:
After reviewing the original Action Cable modularity discussion in #50979 again, I think this PR is probably the wrong long-term direction, so I'm going to close it.

The important conclusion from that discussion is that `action_cable/server/*` is the Rails default Rack hijack server implementation, while the reusable Action Cable boundary is lower down: connection, channel, subscription/broadcasting, protocol, etc. Alternative servers such as Falcon should be able to replace or bypass the default server implementation rather than configuring more of its internals.

This PR makes the executor configurable on `ActionCable::Server::Base`, which is useful as an incremental compatibility hook, but it also risks turning the default server's internal execution model into a public configuration target. That could bake in the current server implementation shape instead of encouraging a cleaner server boundary.

So I think the better follow-up is not "make the default server executor configurable", but rather to define the natural extension point for replacing the Action Cable server implementation itself. Once that boundary is clear, an async/Falcon integration can own scheduling, timers, socket handling, and lifecycle directly instead of fitting into the default threaded server's executor contract.

Relevant discussion from #50979:

- Executor adapterization was explicitly deferred as a follow-up: https://github.com/rails/rails/pull/50979#discussion_r1480459654
- The `action_cable/server` folder was described as the Rails default Rack hijack server, with other implementations encouraged to use the lower-level Action Cable pieces: https://github.com/rails/rails/pull/50979#discussion_r1480469738
- A later unresolved point was whether the executor interface would also need a public `schedule` operation for listener-style subscription adapters: https://github.com/rails/rails/pull/50979#issuecomment-2380320587

Thanks, and apologies for the churn here. I think closing this keeps us from cementing the wrong abstraction.

samuel-williams-shopify:
cc @palkan @matthewd @rafaelfranca

PR #57803, opened 2026-06-21
Allow Action Cable server implementation to be configured

samuel-williams-shopify:
### Motivation / Background

[Action Cable modular server refactoring](https://github.com/rails/rails/pull/50979) made a useful distinction between the default Rails Rack hijack server implementation and the lower-level Action Cable runtime pieces used by connections, channels, broadcasting, and pub/sub.

Alternative server integrations, such as Falcon/Async, should be able to replace the server implementation itself rather than configuring internals of `ActionCable::Server::Base` such as its executor. Configuring those internals risks making the default server implementation shape the public extension target.

This PR adds a small server replacement hook instead: `config.action_cable.server_class`.

Follow up to <https://github.com/rails/rails/pull/57801> and builds on <https://github.com/rails/rails/pull/50979> and <https://github.com/rails/rails/pull/57847>.

### Detail

- Expose `ActionCable.config` as the canonical top-level configuration accessor.
- Add `ActionCable.config.server_class`, defaulting to `ActionCable::Server::Base`.
- Build the `ActionCable.server` singleton with `config.server_class.new(config: config)`.
- Keep the existing `ActionCable::Server::Base#initialize(config: ...)` keyword API intact.
- Keep the existing server contract: the configured server is still the `ActionCable.server` singleton and Rack endpoint mounted by Rails, so it must respond to `#call(env)`.

A configured server class can be set with:

```ruby
config.action_cable.server_class = MyCableServer
```

The server class is initialized with the Action Cable configuration object using the existing keyword shape:

```ruby
def initialize(config:)
  @config = config
end
```

This intentionally starts with the existing `Server#call(env)` Rack endpoint shape. A future refinement could split the Rack integration point from the runtime object with something like `ActionCable.server.middleware`, where the default implementation returns `self`; this PR leaves that extra API out until there is a concrete need for a separate mounted object.

### Checklist

- [x] Tests pass: `ruby -Iactioncable/test actioncable/test/server/base_test.rb`
- [x] Tests pass: `ruby -Iactioncable/test actioncable/test/server/health_check_test.rb`
- [x] RuboCop passes for touched Ruby files

### DIFF
--- a/actioncable/lib/action_cable.rb
+++ b/actioncable/lib/action_cable.rb
@@ -78,6 +78,22 @@ module ActionCable
 
   # Singleton instance of the server
   module_function def server
-    @server ||= ActionCable::Server::Base.new
+    @server ||= build_server
+  end
+
+  module_function def server=(server)
+    @server = server
+  end
+
+  module_function def config
+    @config ||= ActionCable::Configuration.new
+  end
+
+  module_function def build_server
+    if config.server.respond_to?(:new)
+      config.server.new
+    else
+      config.server.call
+    end
   end
 end
--- a/actioncable/lib/action_cable.rb
+++ b/actioncable/lib/action_cable.rb
@@ -78,22 +78,10 @@ module ActionCable
 
   # Singleton instance of the server
   module_function def server
-    @server ||= build_server
-  end
-
-  module_function def server=(server)
-    @server = server
+    @server ||= config.server_class.new(config: config)
   end
 
   module_function def config
     @config ||= ActionCable::Configuration.new
-  end
-
-  module_function def build_server
-    if config.server.respond_to?(:new)
-      config.server.new
-    else
-      config.server.call
-    end
   end
 end
--- a/actioncable/lib/action_cable/configuration.rb
+++ b/actioncable/lib/action_cable/configuration.rb
@@ -8,10 +8,11 @@ module ActionCable
   # # Action Cable Configuration
   #
   # An instance of this configuration object is available via
-  # ActionCable.server.config, which allows you to tweak Action Cable
-  # configuration in a Rails config initializer.
+  # ActionCable.config and ActionCable.server.config, which allows you to tweak
+  # Action Cable configuration in a Rails config initializer.
   class Configuration
     attr_accessor :logger, :log_tags
+    attr_accessor :server
     attr_accessor :connection_class, :worker_pool_size, :executor_pool_size
     attr_accessor :disable_request_forgery_protection, :allowed_request_origins, :allow_same_origin_as_host, :filter_parameters
     attr_accessor :cable, :url, :mount_path
@@ -22,6 +23,7 @@ module ActionCable
     def initialize
       @log_tags = []
 
+      @server = -> { ActionCable::Server::Base.new(config: self) }
       @connection_class = -> { ActionCable::Connection::Base }
       @worker_pool_size = 4
       @executor_pool_size = 10
--- a/actioncable/lib/action_cable/configuration.rb
+++ b/actioncable/lib/action_cable/configuration.rb
@@ -7,9 +7,8 @@ require "rack"
 module ActionCable
   # # Action Cable Configuration
   #
-  # An instance of this configuration object is available via
-  # ActionCable.config and ActionCable.server.config, which allows you to tweak
-  # Action Cable configuration in a Rails config initializer.
+  # An instance of this configuration object is available via ActionCable.config,
+  # which allows you to tweak Action Cable configuration in a Rails config initializer.
   class Configuration
     attr_accessor :logger, :log_tags
     attr_accessor :server_class
--- a/actioncable/lib/action_cable/configuration.rb
+++ b/actioncable/lib/action_cable/configuration.rb
@@ -12,7 +12,7 @@ module ActionCable
   # Action Cable configuration in a Rails config initializer.
   class Configuration
     attr_accessor :logger, :log_tags
-    attr_accessor :server
+    attr_accessor :server_class
     attr_accessor :connection_class, :worker_pool_size, :executor_pool_size
     attr_accessor :disable_request_forgery_protection, :allowed_request_origins, :allow_same_origin_as_host, :filter_parameters
     attr_accessor :cable, :url, :mount_path
@@ -23,7 +23,7 @@ module ActionCable
     def initialize
       @log_tags = []
 
-      @server = -> { ActionCable::Server::Base.new(config: self) }
+      @server_class = ActionCable::Server::Base
       @connection_class = -> { ActionCable::Connection::Base }
       @worker_pool_size = 4
       @executor_pool_size = 10
--- a/actioncable/lib/action_cable/server/base.rb
+++ b/actioncable/lib/action_cable/server/base.rb
@@ -42,7 +42,7 @@ module ActionCable
       include ActionCable::Server::Broadcasting
       include ActionCable::Server::Connections
 
-      cattr_accessor :config, instance_accessor: false, default: ActionCable::Server::Configuration.new
+      cattr_accessor :config, instance_accessor: false, default: ActionCable.config
 
       attr_reader :config
 
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,7 +1,7 @@
 *   Allow configuring the Action Cable server implementation.
 
-    `config.action_cable.server` can now be set to a server class or factory.
-    The configured server remains the `ActionCable.server` singleton and Rack
+    `config.action_cable.server_class` can now be set to a server class. The
+    configured server remains the `ActionCable.server` singleton and Rack
     endpoint, allowing alternative server implementations to provide the Action
     Cable runtime boundary without configuring internals of the default server.
 
--- a/actioncable/CHANGELOG.md
+++ b/actioncable/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Allow configuring the Action Cable server implementation.
+
+    `config.action_cable.server` can now be set to a server class or factory.
+    The configured server remains the `ActionCable.server` singleton and Rack
+    endpoint, allowing alternative server implementations to provide the Action
+    Cable runtime boundary without configuring internals of the default server.
+
+    *Samuel Williams*
+
 *   Move `ActionCable::Server::Configuration` to `ActionCable::Configuration`.
 
     The old constant remains available as an alias.
--- a/actioncable/lib/action_cable/engine.rb
+++ b/actioncable/lib/action_cable/engine.rb
@@ -11,7 +11,7 @@ module ActionCable
     config.action_cable = ActiveSupport::OrderedOptions.new
     config.action_cable.mount_path = ActionCable::INTERNAL[:default_mount_path]
     config.action_cable.precompile_assets = true
-    config.action_cable.server = ActionCable::Server::Base
+    config.action_cable.server_class = ActionCable::Server::Base
 
     guard_load_hooks(
       :action_cable_channel, :action_cable_connection,
--- a/actioncable/lib/action_cable/engine.rb
+++ b/actioncable/lib/action_cable/engine.rb
@@ -11,6 +11,7 @@ module ActionCable
     config.action_cable = ActiveSupport::OrderedOptions.new
     config.action_cable.mount_path = ActionCable::INTERNAL[:default_mount_path]
     config.action_cable.precompile_assets = true
+    config.action_cable.server = ActionCable::Server::Base
 
     guard_load_hooks(
       :action_cable_channel, :action_cable_connection,
--- a/actioncable/test/server/base_test.rb
+++ b/actioncable/test/server/base_test.rb
@@ -88,4 +88,46 @@ class BaseTest < ActionCable::TestCase
     assert_same ActionCable::Configuration, ActionCable::Server::Configuration
     assert_instance_of ActionCable::Configuration, ActionCable::Server::Base.config
   end
+
+  test "ActionCable.server uses configured server factory class" do
+    server_class = Class.new do
+      attr_reader :config
+
+      def initialize
+        @config = ActionCable::Server::Base.config
+      end
+    end
+
+    with_server_factory(server_class) do
+      assert_instance_of server_class, ActionCable.server
+      assert_same ActionCable::Server::Base.config, ActionCable.server.config
+    end
+  end
+
+  test "ActionCable.server uses configured server factory" do
+    server = Object.new
+
+    with_server_factory(-> { server }) do
+      assert_same server, ActionCable.server
+    end
+  end
+
+  private
+    def with_server_factory(factory)
+      previous_factory = ActionCable.config.server
+      previous_server = ActionCable.instance_variable_get(:@server) if ActionCable.instance_variable_defined?(:@server)
+
+      ActionCable.remove_instance_variable(:@server) if ActionCable.instance_variable_defined?(:@server)
+      ActionCable.config.server = factory
+
+      yield
+    ensure
+      ActionCable.config.server = previous_factory
+
+      if defined?(previous_server)
+        ActionCable.instance_variable_set(:@server, previous_server)
+      elsif ActionCable.instance_variable_defined?(:@server)
+        ActionCable.remove_instance_variable(:@server)
+      end
+    end
 end
--- a/actioncable/test/server/base_test.rb
+++ b/actioncable/test/server/base_test.rb
@@ -89,45 +89,37 @@ class BaseTest < ActionCable::TestCase
     assert_instance_of ActionCable::Configuration, ActionCable::Server::Base.config
   end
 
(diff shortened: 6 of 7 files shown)


samuel-williams-shopify:
See <https://github.com/socketry/async-cable/pull/8> for the Executor implementation.

samuel-williams-shopify:
cc @palkan wdyt?

actioncable/lib/action_cable/configuration.rb
```diff
@@ -0,0 +1,75 @@
...
+    attr_accessor :disable_request_forgery_protection, :allowed_request_origins, :allow_same_origin_as_host, :filter_parameters
+    attr_accessor :cable, :url, :mount_path
+    attr_accessor :precompile_assets
+    attr_accessor :health_check_path, :health_check_application
+    attr_writer :pubsub_adapter
+
+    def initialize
+      @log_tags = []
+
+      @server = -> { ActionCable::Server::Base.new(config: self) }
```
palkan: I wonder if we should go with the `server_class` pattern instead and introduce the contract for server implementations (`.new(config:)`?

`server_class` would be clearer here, for example: https://github.com/socketry/async-cable/pull/9/changes#diff-13f8d1f8b4dce835d5e33365b7e9a7d71e38a666119c2ce5d2b3050024a9a7f5R13

(we're not providing a server instance).

And, probably, we can simplify a bit and just require a Proc as a configuration (similarly to `connection_class`).

So, the `.server` would become:

```ruby
module_function def server
  @server ||= config.server_class.new(config:)
end
```

This way we won't need the `.build_server` helper.

actioncable/lib/action_cable.rb
```diff
@@ -78,6 +78,22 @@ module ActionCable
 
   # Singleton instance of the server
   module_function def server
-    @server ||= ActionCable::Server::Base.new
+    @server ||= build_server
+  end
+
+  module_function def server=(server)
```
palkan: Why do we need this shortcut?

samuel-williams-shopify: It was for testing, I've removed it in favor of a small test helper.

actioncable/lib/action_cable/configuration.rb
samuel-williams-shopify: Done.

actioncable/lib/action_cable/server/base.rb
```diff
@@ -51,7 +51,7 @@ def self.logger; config.logger; end
 
       attr_reader :mutex
 
-      def initialize(config: self.class.config)
+      def initialize(config = self.class.config)
```
tenderlove: Is changing this necessary?  I don't mind too much, but if it's public API I don't see an upside to changing it.

actioncable/lib/action_cable/server/configuration.rb
```diff
@@ -2,74 +2,10 @@
...
+require "action_cable/configuration"
 
 module ActionCable
   module Server
-    # # Action Cable Server Configuration
-    #
-    # An instance of this configuration object is available via
-    # ActionCable.server.config, which allows you to tweak Action Cable
-    # configuration in a Rails config initializer.
-    class Configuration
```
tenderlove: Moving this class to its own file seems good. Can you open a PR that does just that? Then we can have a small diff here.

tenderlove (commented):
This seems fine to me, in general.  The only thing I worry about is that it basically makes the interface totally public, which I guess is fine since it's a public class anyway?

actioncable/lib/action_cable/server/configuration.rb
samuel-williams-shopify: Opened #57847 with just the configuration class move: https://github.com/rails/rails/pull/57847

actioncable/lib/action_cable/server/base.rb
samuel-williams-shopify: I agree, I don't think we should change this.

samuel-williams-shopify: Agreed. I restored `ActionCable::Server::Base#initialize(config: ...)` and the server class hook now instantiates with `config.server_class.new(config: config)`, so this public extension point remains source-compatible.

samuel-williams-shopify:
> The only thing I worry about is that it basically makes the interface totally public, which I guess is fine since it's a public class anyway?

I don’t think this materially changes the compatibility concern / surface area, but I do think that before this is finally released we should have a clear public interface, which is what rails/rails#57804 is trying to move toward.

Ideally, I think the server interface is flatter than the current implementation shape. For example, external Action Cable code should be able to say:

```ruby
server.broadcast(...)
server.remote_connections.where(...).disconnect
server.perform_work(connection, receiver, :method_name, *args)
server.post { ... }
server.timer(interval) { ... }
server.logger
server.config
```

rather than reaching through:

```ruby
server.worker_pool.async_invoke(...)
server.worker_pool.halt
server.worker_pool.post(...)
server.worker_pool.timer(...)
server.executor.post(...)
server.executor.timer(...)
```

In other words, I’m not trying to make every method on `ActionCable::Server::Base` public. I’m trying to make `ActionCable.server` the public runtime boundary, and keep things like `executor` / `worker_pool` as implementation details of the default server.

That seems better for compatibility too: alternative server implementations can provide the small server-level contract Action Cable needs, without having to mimic the internals of `ActionCable::Server::Base`.

palkan (commented):
LGTM 🚀

samuel-williams-shopify:
I had a useful side discussion with @matthewd about the shape of this extension point, and I think it’s worth capturing the tradeoff here.

One concern is that `ActionCable.config` currently includes settings that are specific to the default server implementation. For example, `connection_class`, `worker_pool_size`, and `executor_pool_size` all assume particular implementation details of `ActionCable::Server::Base`. If `server_class` allows a different server implementation to be selected, those settings may no longer be meaningful, or may need different semantics for different server implementations.

A related concern is compatibility: if `ActionCable.server` directly becomes the custom server object, then the methods used by Action Cable and by applications form a public runtime contract. That may make future interface changes harder, because alternative server implementations and downstream libraries could need to change in lockstep with Rails.

The counterpoint is that `ActionCable.server` is already the global runtime object used by Action Cable components for broadcasting, remote connections, connection handling, pub/sub, and the mounted Rack endpoint. This PR does not introduce that boundary, but it does make the boundary configurable and replaceable. That is a real design commitment: it means the methods Action Cable expects from its server become an interface that can be externally provided, rather than purely an internal Rails implementation detail that can evolve freely.

There is also an alternative design direction where Action Cable keeps users talking only to Action Cable provided objects, and those objects delegate to server-supplied components behind a smaller compatibility layer. That could preserve more room for future compatibility translation.

I think the key decision is whether making `ActionCable.server` replaceable is a real compatibility problem, and if so, whether the default server should instead become a proxy over some internal adapter, such as an executor-like object, that provides the relevant primitives. That proxy would still need to cover the same practical surface area: broadcasting, connection handling, the mounted `/cable` endpoint, subscription adapters, timers, and work scheduling all have execution-model assumptions. Replacing only the executor, as it is currently designed, has proven too narrow for the async server integration.

Even if we introduced that internal adapter boundary, it would not solve the configuration mismatch by itself. `ActionCable.config` would still contain default-server-specific settings such as `worker_pool_size` and `executor_pool_size`, and we would still need internal changes to make those concepts less central, for example folding `worker_pool` behavior into the executor or otherwise removing `worker_pool` from the effective interface Action Cable components depend on.

So the tradeoff is between:

- adding a small configuration hook for the existing `ActionCable.server` boundary, which is direct and works with the current architecture; and
- designing a deeper internal adapter layer behind the server, which may preserve more compatibility flexibility long term but requires a larger redesign of the default server, executor, worker pool, and configuration shape.

This PR intentionally takes the smaller step: keep the existing `ActionCable.server` runtime boundary, expose `ActionCable.config` as the pre-instantiation configuration object, and allow the server class to be selected there. If we think being able to replace `ActionCable.server` with a user provided object cements too much of the server interface, then the alternative is probably not just a different configuration location; it is a broader internal refactor to define the adapter boundary that the server proxies to.

palkan:
> whether the default server should instead become a proxy over some internal adapter, such as an executor-like object, that provides the relevant primitives. 

Design-wise, that looks like the best option to me; `ActionCable.server` is more like an _application instance_ right now, not just a _server_. 


> Replacing only the executor, as it is currently designed, has proven too narrow for the async server integration.

Could you please share the integration points missing? I wonder, how big is the refactoring surface 🤔. 

From [this PR](https://github.com/socketry/async-cable/pull/9/changes#diff-8c0d8ff44e85a9a2f629a3bd9c38048a5ba09146a5f2b193453651d081a577a2R31), it seems that adapterizing the Rack handler part and making the executor class / instance configurable should do the trick. 

So, it seems that for now , the "designing a deeper internal adapter layer" proposal could be minimized down to:

- Making the executor class configurable (`config.action_cable.server.executor_class = Async::Cable::Executor`)
- Making the Rack handler configurable (`config.action_cable.server.rack_handler = Async::Cable::Handler`); the handler encapsulates the Socket interface, so no need to configure it separately; I think, it's safe to assume that we stick to Rack and don't need additional flexibility of supporting anything beyond `#call(env)`.
- Extracting the current Rack handler from `action_cable/server/*` 

The primary `server.rb` change would be smth like this:

```diff
       def call(env)
         return config.health_check_application.call(env) if env["PATH_INFO"] == config.health_check_path
         setup_heartbeat_timer
-        Socket.new(self, env).process
+        rack_handler.call(env)
       end
```

No need to duplicate health checking logic, heartbeats (they're already go through the executor).

Wait, what about `worker_pool_size` and `executor_pool_size? Here is what we can do:

- Keep `worker_pool_size` and the threaded worker pool as is; it's only initialized and used by the server implementation if it needs it (alternative non-async servers may benefit from it since it gives a robust way of executing userspace code via a Rails executor). We can also rename the option to smth less specific, say, `work_concurrency`, so the configuration could be reused even by the Async impl as a barrier configuration (to keep the number of concurrent Action Cable commands under the control).

- Refactor `executor_pool_size` to `executor_args` / ` executor_options` as ActiveSupport::OrderedOptions or as arbitrary Hash/Object and allow custom server implementations to configure it to their liking. The default configuration would change from `config.action_cable.executor_pool_size = 10` to `config.action_cable.executor_args = {pool_size: 10}`. I doubt we can find a production app having this parameter non-default, so the change is breaking only in theory (and we can add a deprecated fallback just in case).

Thoughts?

---

P.S. Speaking of the desired flexibility of Action Cable interface, I think, we should think of it more as "designed for a few" than "designed for anyone". I don't expect people building many custom server implementations, so we can introduce minimal integration points gradually (and carefully) keeping particular use cases in mind. 

We're not rethinking the final design of Action Cable for the next 10 years (again); we will continue improving it continuously from now on (at least that's how I see it).

samuel-williams-shopify:
@palkan thanks for your reply.

I tried replacing executor and the surface area was insufficient for a proper async server implementation. The most obvious one was the rack `/cable` handling. Yes, you could have configuration points for the various internal parts (middleware, executor, worker pool), but in what reasonable scenario is someone going to want to use async-cable for only part of the server implementation. I think that's just inviting unnecessary complexity.

Regarding your `server.rb` example:

```
       def call(env)
         return config.health_check_application.call(env) if env["PATH_INFO"] == config.health_check_path
         setup_heartbeat_timer
-        Socket.new(self, env).process
+        rack_handler.call(env)
       end
```

To me, this is kind of painful, since I kind of want to replace everything in that method, I think, or at least, I don't want to be limited by the existing design and thinking. That applies to all the other server entry points. I don't think server should expose the worker pool, or executor, or rack middleware, it should have a flat public interface IMHO.

Also, I think if we are concerned about the future server interface needing changes, that's not really a problem for Rails, we can even have an interface validation check after `server_class.new` -> check that it responds to the right interface etc.

palkan:
Thanks @samuel-williams-shopify!

 > To me, this is kind of painful, since I kind of want to replace everything in that method, I think, or at least, I don't want to be limited by the existing design and thinking. That applies to all the other server entry points. I don't think server should expose the worker pool, or executor, or rack middleware, it should have a flat public interface IMHO.
 
Yeah, I agree.

> we can even have an interface validation check after server_class.new -> check that it responds to the right interface etc.

I'd better have another _layer_; we (Rails) can stay in control of the `ActionCable.server` interface and let users and library developers not to worry about compatibility. 

---

**tl;dr** Extract `ActionCable.server.websocket_server` (w/ event loop, worker pool, etc.), keep `ActionCable.server` API intact, keep `ActionCable.server.executor`, keep all the configuration options intact (even for the default WebSocket server implementation).

For now, the public/internal surface of the `server` API is as follows:

```
# Properties
server.logger
server.config #== ActionCable.config
server.executor # used by subscription adapters; not sure if we need to move this down to the server 
implementation level yet
server.pubsub # pub/sub adapter
server.remote_connections

# Methods
server.broadcast # delegates to pub/sub adapter
server.disconnect # delegates to remote_connections -> delegates to #broadcast 
server.restart

# Other
mount ActionCable.server => mount_path # in routes
```

Then, `worker_pool`, `event_loop`, `heartbeat_timer`, `connections_map` and health checking could be extracted into a lower level abstraction, say, `server.websocket_server`. All the properties and methods above stay the same (`#restart` calls `websocket_server.try(:restart)`). We can change the mount part to `mount ActionCable.server.websocket_server => ... if ActionCable.server.websocket_server.try(:mountable?)`—additional optional predicate.

We can keep `worker_pool_size`, `event_pool_size`, etc in the ActionCable::Configuration object—it's fine to have default implementation options there. Alternative websocket_server implementation may or may not use them. 

Applying to the `async-cable`, you will need to configure two things in the railtie:

```ruby
# WebSocket server implementation
config.action_cable.server.websocket_server = -> { Async::Cable::Server.new(config: ::ActionCable.config) }
config.action_cable.server.executor = -> { Async::Cable::Executor.new(config: ::ActionCable.config) }
```

The usage of Procs allows us not to enforce any construction interface (you can use your own config or whataver). A bit more flexible, I think.

This is just a sketch, of course, but if the overall direction looks good to everyone, I'd be happy to bring it to life. /cc @samuel-williams-shopify @matthewd  @tenderlove

samuel-williams-shopify:
I personally prefer this PR. I don't like over complicating the design for hypothetical issues. I'd like to flatten the interface, not make it deeper. Law of Demeter and all that.

matthewd:
Yeah, I like @palkan's design here.

We are a framework, not a well-known constant that directly connects callers to an implementation-provided object.

While I don't foresee a sudden flood of additional alternative ACa servers, I still feel that Async's particular need to re-colour ~everything is not representative. 

> I kind of want to replace everything in that method, I think, or at least, I don't want to be limited by the existing design and thinking

> Also, I think if we are concerned about the future server interface needing changes, that's not really a problem for Rails, we can even have an interface validation check after server_class.new -> check that it responds to the right interface etc.

This feels like a much more Rack-shaped approach; that's not what Rails is. Active Job doesn't just point you over to `Sidekiq::Client` or whatever.

samuel-williams-shopify:
Okay, that's fair, so the alternative to this PR I'd like to see is:

1. A separate PR, adjacent to this one, exposing the hooks/configuration in ActionCable.
2. A PR to `Async::Cable` to adopt that interface, and works efficiently.
3. A PR to `AnyCable` which also adopts that interface, if applicable.
4. Rails' own server should not need any internal wiring beyond that interface to work.

In addition to that, I think that:

1. It should **NOT** be possible to configure ActionCable in an invalid state, e.g. if `websocket_server` comes from `AnyCable` and `executor` comes from `Async::Cable`, it should work. Otherwise I don't know why those can be configured separately.
2. The implementation is completely efficient for a given web application server and actioncable server, e.g. the default server should be efficient within it's design parameters for any server, `Async::Cable` should be event driven on both puma and falcon where possible, etc. In other words, the server should not expose a thread backed worker pool if running on an event loop.
3. Ideally the server can be used independently of Rails, e.g. this might be a useful general purpose messaging interface (e.g. could Sinatra or Roda use it? Maybe not a bad idea to think about how it can be totally generic).

@palkan will you implement this?

> The usage of Procs allows us not to enforce any construction interface (you can use your own config or whataver). A bit more flexible, I think.

It's easy enough to make a class respond to `call` -> `new`, e.g. `class Async::Cable::Server; def self.call(...); self.new(...); end; end` - I'm not sure using a Proc has much advantage, because the opposite is also true to implement `.new` to do whatever you want. One of the nice features of Ruby IMHO.

In addition, if you go down this path, it would make more sense:

```rails
config.action_cable.server.websocket_server = -> {|config| Async::Cable::Server.new(config)}
```

A keyword argument should only be used for truly optional arguments, config isn't optional IMHO. However, the existing interface for `ActionCable::Server` requires a keyword argument. If we can change it to just be a required positional argument, that would be my preference, unless there is precedence for this elsewhere.

palkan:
> @palkan will you implement this?

Yeah, I'd be happy to. Will need ~2 weeks given my current schedule (and parallel AC stuff).

palkan:
@samuel-williams-shopify @matthewd Here you go! https://github.com/rails/rails/pull/58121

PR #57804, opened 2026-06-21
Route Action Cable work scheduling through server

samuel-williams-shopify:
### Motivation / Background

This builds on #57803 and includes that change squashed as the first commit so the PR can be reviewed directly against `main`.

If Action Cable can be backed by a custom server implementation, the server itself should be the public runtime boundary. Routing scheduling through `ActionCable.server` keeps `executor` and `worker_pool` as implementation details of the default server rather than configuration targets that external integrations need to depend on.

### Detail

* Add server-level scheduling APIs to `ActionCable::Server::Base`:
  * `perform_work(connection, receiver, method, *args, &block)`
  * `post(task = nil, &block)`
  * `timer(interval, &block)`
  * `schedule(&block)`
* Route connection, channel timer, socket, and subscription adapter work scheduling through the server.
* Keep `executor` and `worker_pool` private to `ActionCable::Server::Base`.
* Avoid forcing executor initialization from `pubsub`.

### Notes

The first commit contains the #57803 changes squashed into one commit. The second commit contains the scheduling changes.

### Verification

Passed locally:

```
ruby -Iactioncable/test actioncable/test/server/base_test.rb
ruby -Iactioncable/test actioncable/test/server/health_check_test.rb
ruby -Iactioncable/test actioncable/test/server/socket_test.rb
ruby -Iactioncable/test actioncable/test/channel/periodic_timers_test.rb
ruby -Iactioncable/test actioncable/test/subscription_adapter/async_test.rb
ruby -Iactioncable/test actioncable/test/subscription_adapter/inline_test.rb
ruby -Iactioncable/test actioncable/test/subscription_adapter/test_adapter_test.rb
ruby -Iactioncable/test actioncable/test/connection/test_case_test.rb
ruby -Iactioncable/test actioncable/test/channel/stream_test.rb
bundle exec rubocop actioncable/lib/action_cable.rb actioncable/lib/action_cable/configuration.rb actioncable/lib/action_cable/engine.rb actioncable/lib/action_cable/server/base.rb actioncable/lib/action_cable/server/socket.rb actioncable/lib/action_cable/connection/base.rb actioncable/lib/action_cable/channel/periodic_timers.rb actioncable/lib/action_cable/subscription_adapter/base.rb actioncable/lib/action_cable/subscription_adapter/async.rb actioncable/lib/action_cable/subscription_adapter/postgresql.rb actioncable/lib/action_cable/subscription_adapter/redis.rb actioncable/lib/action_cable/subscription_adapter/subscriber_map.rb actioncable/test/server/base_test.rb actioncable/test/server/health_check_test.rb actioncable/test/channel/periodic_timers_test.rb actioncable/test/stubs/test_socket.rb actioncable/test/subscription_adapter/common.rb actioncable/test/subscription_adapter/redis_test.rb actioncable/test/subscription_adapter/channel_prefix.rb
```

Redis adapter tests timed out locally waiting for subscription confirmation, which appears environment related. PostgreSQL adapter tests were not run locally due missing local database setup.

### DIFF
--- a/actioncable/lib/action_cable/server/base.rb
+++ b/actioncable/lib/action_cable/server/base.rb
@@ -107,34 +107,28 @@ module ActionCable
         @event_loop || @mutex.synchronize { @event_loop ||= StreamEventLoop.new }
       end
 
-      # The worker pool is where we run connection callbacks and channel actions. We
-      # do as little as possible on the server's main thread. The worker pool is an
-      # executor service that's backed by a pool of threads working from a task queue.
-      # The thread pool size maxes out at 4 worker threads by default. Tune the size
-      # yourself with `config.action_cable.worker_pool_size`.
-      #
-      # Using Active Record, Redis, etc within your channel actions means you'll get a
-      # separate connection from each thread in the worker pool. Plan your deployment
-      # accordingly: 5 servers each running 5 Puma workers each running an 8-thread
-      # worker pool means at least 200 database connections.
-      #
-      # Also, ensure that your database connection pool size is as least as large as
-      # your worker pool size. Otherwise, workers may oversubscribe the database
-      # connection pool and block while they wait for other workers to release their
-      # connections. Use a smaller worker pool or a larger database connection pool
-      # instead.
-      def worker_pool
-        @worker_pool || @mutex.synchronize { @worker_pool ||= ActionCable::Server::Worker.new(max_size: config.worker_pool_size) }
+      def perform_work(connection, receiver, method, *args, &block)
+        worker_pool.async_invoke(receiver, method, *args, connection: connection, &block)
       end
 
-      # Executor is used by various actions within Action Cable (e.g., pub/sub operations) to run code asynchronously.
-      def executor
-        @executor || @mutex.synchronize { @executor ||= ThreadedExecutor.new(max_size: config.executor_pool_size, name: "streamer") }
+      def post(task = nil, &block)
+        executor.post(task, &block)
+      end
+
+      def timer(interval, &block)
+        executor.timer(interval, &block)
+      end
+
+      def schedule(&block)
+        Thread.new do
+          Thread.current.abort_on_exception = true
+          block.call
+        end
       end
 
       # Adapter used for all streams/broadcasting.
       def pubsub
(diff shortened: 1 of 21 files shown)


actioncable/lib/action_cable/server/base.rb
```diff
@@ -107,34 +107,28 @@ def event_loop
...
+      def timer(interval, &block)
+        executor.timer(interval, &block)
+      end
+
+      def schedule(&block)
+        Thread.new do
+          Thread.current.abort_on_exception = true
+          block.call
+        end
       end
```
palkan, line 127: I'm not sure about the `#schedule` interface. As I understand, we need it only in some subscription adapters to spawn background listeners—yes, that's something still coupled with the threaded nature of the original Action Cable. However, extracting just the `Thread.new ...` call seems not enough: for example, Redis subscriber _knows_ that the returned value is a Thread (and not something else) and relies on that fact: https://github.com/rails/rails/blob/fe66563e16ca49b1789b3bfcdbeb3353e3899939/actioncable/lib/action_cable/subscription_adapter/redis.rb#L150

How will the async version look like?

If we can't get rid of hard dependency on threads in subscribers, I don't see any value in extracting the `#schedule` helper. I think, introduce some abstract _Task_ interface (`schedule(&block) -> Task`) and refactor subscribers to use it should do the trick.

samuel-williams-shopify: I basically agree, and I think defining a more minimal interface is probably a good idea, but in any case, spawning threads isn't inherently incompatible witih Async, it's just a little less efficient.

palkan: Then, maybe, let's avoid extracting `#schedule` for now and keep explicit Thread usage in the corresponding adapters?

> threads isn't inherently incompatible witih Async, it's just a little less efficient

I guess, for the most efficient usage of async environment, we will need dedicated adapters anyway, right?

samuel-williams-shopify: > Then, maybe, let's avoid extracting #schedule for now and keep explicit Thread usage in the corresponding adapters?

Yes, that's probably a good idea, the adapters minimum contract should be thread safety.

> I guess, for the most efficient usage of async environment, we will need dedicated adapters anyway, right?

Correct, and we have PRs ready to implement this, e.g. using `async-redis`.

PR #57814, opened 2026-06-22
Make ActiveSupport::JSON encoders Ractor-shareable

etiennebarrie:
If we freeze early, we can avoid having to call `Ractor.make_shareable` and just use the ractor-shareable JSON encoders from Ractors.

On 3.4 and below the test is not defined at all to avoid failing skips. Updating JSON was also necessary for the skips not to fail in Ruby 4.0.

### DIFF
--- a/activesupport/lib/active_support/json/encoding.rb
+++ b/activesupport/lib/active_support/json/encoding.rb
@@ -61,22 +61,22 @@ module ActiveSupport
       U2029 = -"\u2029".b
 
       ESCAPED_CHARS = {
-        U2028 => '\u2028'.b,
-        U2029 => '\u2029'.b,
-        ">".b => '\u003e'.b,
-        "<".b => '\u003c'.b,
-        "&".b => '\u0026'.b,
+        U2028 => -'\u2028'.b,
+        U2029 => -'\u2029'.b,
+        ">".b => -'\u003e'.b,
+        "<".b => -'\u003c'.b,
+        "&".b => -'\u0026'.b,
       }.freeze
 
-      HTML_ENTITIES_REGEX = Regexp.union(*(ESCAPED_CHARS.keys - [U2028, U2029]))
-      FULL_ESCAPE_REGEX = Regexp.union(*ESCAPED_CHARS.keys)
-      JS_SEPARATORS_REGEX = Regexp.union(U2028, U2029)
+      HTML_ENTITIES_REGEX = Regexp.union(*(ESCAPED_CHARS.keys - [U2028, U2029])).freeze
+      FULL_ESCAPE_REGEX = Regexp.union(*ESCAPED_CHARS.keys).freeze
+      JS_SEPARATORS_REGEX = Regexp.union(U2028, U2029).freeze
 
       class JSONGemEncoder # :nodoc:
         attr_reader :options
 
         def initialize(options = nil)
-          @options = options || {}
+          @options = options.dup.freeze || {}.freeze
         end
 
         # Encode the given object into a JSON string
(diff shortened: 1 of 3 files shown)


activesupport/lib/active_support/json/encoding.rb
```diff
@@ -61,22 +61,22 @@ module Encoding # :nodoc:
...
+      HTML_ENTITIES_REGEX = Regexp.union(*(ESCAPED_CHARS.keys - [U2028, U2029])).freeze
+      FULL_ESCAPE_REGEX = Regexp.union(*ESCAPED_CHARS.keys).freeze
+      JS_SEPARATORS_REGEX = Regexp.union(U2028, U2029).freeze
 
       class JSONGemEncoder # :nodoc:
         attr_reader :options
 
         def initialize(options = nil)
-          @options = options || {}
+          @options = options.freeze || {}.freeze
```
byroot:
```suggestion
          @options = options.dup.freeze || {}.freeze
```
nitpick: We probably shouldn't mutate the argument.

etiennebarrie: Yeah I wasn't sure if there was a convention here. My assumption was that any place where literal options are passed, we don't need to dup, any place where a reusable Hash is reused via a variable could just `dup` at the call site if they really care about it not being frozen. In other words we are passed ownership of the options Hash and can mutate it freely, avoiding multiple allocations in the process. But I can see how that could be problematic, fixed.

byroot: Honestly I highly doubt anyone would ever complain, but it's kinda expected that you don't mutate your arguments, so...

byroot:
CI ran into an interesting VM crash: https://buildkite.com/rails/rails/builds/130083#019ef009-aaee-40c9-8e13-319df16f6aed/L8326

```
/rails/activesupport/test/core_ext/module/attribute_accessor_test.rb:14: [BUG] should have cvar cache entry
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [x86_64-linux]

-- C level backtrace information -------------------------------------------
/usr/local/lib/libruby.so.4.0(rb_print_backtrace+0x14) [0x7fdf7c0cf28f] /usr/src/ruby/vm_dump.c:1105
/usr/local/lib/libruby.so.4.0(rb_vm_bugreport) /usr/src/ruby/vm_dump.c:1450
/usr/local/lib/libruby.so.4.0(rb_bug_without_die_internal+0x71) [0x7fdf7bd93805] /usr/src/ruby/error.c:1098
/usr/local/lib/libruby.so.4.0(rb_bug) /usr/src/ruby/error.c:1116
/usr/local/lib/libruby.so.4.0(update_classvariable_cache+0xe) [0x7fdf7bd96602] /usr/src/ruby/vm_insnhelper.c:1614
/usr/local/lib/libruby.so.4.0(vm_exec_core+0x31ef) [0x7fdf7c0b1c7f] /usr/src/ruby/vm_insnhelper.c:1647
/usr/local/lib/libruby.so.4.0(vm_exec_loop+0xa) [0x7fdf7c0b524a] /usr/src/ruby/vm.c:2825
/usr/local/lib/libruby.so.4.0(rb_vm_exec) /usr/src/ruby/vm.c:2801
```

It rings a bell, it think it may already have been fixed a couple weeks ago.

etiennebarrie:
Ah my bad, I removed a skip because I couldn't repro that crash on this branch, but if it has failed, it will continue to do so and cause CI flakiness so we should probably revert until 4.0.6 ships. 😞

PR #57816, opened 2026-06-22
Set browser binary when preloading system test driver

titusfortner:
### Motivation / Background

When system tests preload the Selenium driver, `Browser#preload` sets the global `Chrome/Firefox::Service.driver_path` and relied on `DriverFinder` also setting `options.binary`. That assignment was an unintended side effect of `DriverFinder` mutating the options object, and Selenium 4.45 removed it. Now that `Service.driver_path` is set, Selenium skips Selenium Manager when each worker builds its driver, so it resolves only the driver path and never the browser path — leaving `options.binary` unset and failing with "cannot find Chrome binary."

See the upstream report: https://github.com/SeleniumHQ/selenium/issues/17698

### Detail

This Pull Request resolves the driver and browser paths from the same `DriverFinder` and assigns `options.binary` explicitly, rather than depending on the removed side effect. This works across the selenium-webdriver versions covered by the existing `>= 4.20.0` branch.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch/system_testing/browser.rb
+++ b/actionpack/lib/action_dispatch/system_testing/browser.rb
@@ -65,7 +65,9 @@ module ActionDispatch
         def resolve_driver_path(namespace)
           # The path method has been deprecated in 4.20.0
           if Gem::Version.new(::Selenium::WebDriver::VERSION) >= Gem::Version.new("4.20.0")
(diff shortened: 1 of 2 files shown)


titusfortner:
@byroot / @gmcgibbon sorry for the direct pings. I'm not certain who reviews these typically, but this is a small fix for a regression from Selenium 4.45. Would appreciate a look when you have a moment, thanks!

PR #57817, opened 2026-06-22
Use the float value for `:wait` if passed to `ActiveJob.retry_on`

p8:
If we are going to allow passing a Float, we should use the float value instead of converting it to an Integer.
This is similar to `ActiveJob.set(wait: 4.5)` where we [use the float value](https://github.com/rails/rails/blob/b8990a6e917570d9c4e3f6a694fcc5cfa755ae53/activejob/lib/active_job/core.rb#L176-L177) as well.

By calling `to_f` instead of `to_i` durations with floats, lile `1.5.seconds`, work as well.

This also changes the test to use a float of `4.5` to distinguish it from the 2 for the amount of retries.

https://github.com/rails/rails/pull/57727 was also missing a Changelog entry.

### Checklist

### DIFF
--- a/activejob/CHANGELOG.md
+++ b/activejob/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Allow `retry_on` to accept a Float for `:wait`.
+
+    Similar to `ActiveJob.set(wait: 1.5)`, allow a Float to be passed.
+    Also, passing an unsupported `:wait` type will now raise when loading the
+    job class.
+
+    *Kenta Ishizaki, Petrik de Heus*
+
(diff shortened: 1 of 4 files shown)


fatkodima:
`ActiveSupport::Duration` can also have float values, like `1.5.minutes`, but it is converted to integer currently.

p8:
1.5.minutes is stored as 90 integer seconds. 
1.5.seconds is stored as a float:
```ruby
1.5.seconds.to_f
=> 1.5
```
So I've changed the implementation to call `to_f` instead of `to_i`.

55728:
Thank you for adding the CHANGELOG entry and including my name, @p8! 👀✨🎉

PR #57821, opened 2026-06-23
Prevent ActiveSupport::CurrentAttributes#set from leaking unassigned keys

antonzaharia:
### Summary

This PR makes `ActiveSupport::CurrentAttributes` treat `nil` as unset: assigning `nil` to an attribute removes its key from the internal `@attributes` hash instead of storing `nil`.

### Motivation

A `CurrentAttributes` reader returns `nil` for any attribute that hasn't been set, so a stored `nil` and an absent key are indistinguishable to callers. `CurrentAttributes` also deliberately omits attributes without a default from `@attributes` to keep serialized payloads (Sidekiq jobs, request context, etc.) small — and stray `nil` keys quietly defeat that.

This also surfaced as a leak in `#set`, which delegates to `Object#with`. `with` restores the previous values through the writers when the block ends, so a defaultless attribute — absent before the block — was written back as `nil` and left behind for the rest of the execution loop.

### Example

```ruby
  attribute :request_id
end

Current.attributes # => {}

Current.set(request_id: "abc") do
  # ... processing
end

# Before: key leaks as nil
Current.attributes # => { request_id: nil }

# After: stays empty
Current.attributes # => {}


p8:
Does it make sense to store nil values on Current? Or should setting an attribute to nil always delete the key?
That would probably be a simpler implementation.

antonzaharia:
> Does it make sense to store nil values on Current? Or should setting an attribute to nil always delete the key? That would probably be a simpler implementation.

@p8  Good call - makes sense. A reader returns nil for any unset attribute anyway, so a stored nil and a missing key are indistinguishable. Writers now delete the key on nil (and nil defaults are omitted too), which fixes the #set leak at the root and keeps #set a plain with delegation. Updated the title/description and added tests.

antonzaharia:
> Does it make sense to store nil values on Current? Or should setting an attribute to nil always delete the key? That would probably be a simpler implementation.

Hey @p8 ! Anything else left to address on this one?

antonzaharia:
> Does it make sense to store nil values on Current? Or should setting an attribute to nil always delete the key? That would probably be a simpler implementation.

Hey @p8 ! Mind having another look please?

PR #57823, opened 2026-06-23
Initialize ActionView::LookupContext.view_context_class early

etiennebarrie:
This avoids interacting with the mutex when possible, i.e. when enable_reloading is false.

This lets us use the view context class from Ractors without any additional setup.

### DIFF
--- a/actionview/lib/action_view/lookup_context.rb
+++ b/actionview/lib/action_view/lookup_context.rb
@@ -58,7 +58,6 @@ module ActionView
 
       @details_keys = Concurrent::Map.new
       @digest_cache = Concurrent::Map.new
-      @view_context_mutex = Mutex.new
 
       def self.digest_cache(details)
         @digest_cache[details_cache_key(details)] ||= Concurrent::Map.new
@@ -80,7 +79,7 @@ module ActionView
         ActionView::PathRegistry.all_resolvers.each do |resolver|
           resolver.clear_cache
         end
-        @view_context_class = nil
+        ActionView::LookupContext.reset_view_context_class
         @details_keys.clear
         @digest_cache.clear
       end
@@ -88,13 +87,21 @@ module ActionView
       def self.digest_caches
         @digest_cache.values
       end
+    end
 
-      def self.view_context_class
-        @view_context_mutex.synchronize do
-          @view_context_class ||= ActionView::Base.with_empty_template_cache
-        end
+    def self.reset_view_context_class
+      @view_context_mutex.synchronize { @view_context_class = nil }
+    end
+
+    def self.view_context_class
+      return @view_context_class if @view_context_class
+      base = ActionView::Base # prevent recursive locking
+      @view_context_mutex.synchronize do
+        @view_context_class = base.with_empty_template_cache
       end
     end
+    @view_context_mutex = Mutex.new
+    ActiveSupport.on_load(:action_view) { ActionView::LookupContext.view_context_class }
 
     # Add caching behavior on top of Details.
     module DetailsCache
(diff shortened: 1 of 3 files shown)


actionview/test/template/lookup_context_test.rb
```diff
@@ -197,6 +197,19 @@ def teardown
...
     assert_equal ["foo"], @lookup_context.prefixes
   end
+
+  if RUBY_VERSION >= "4.0"
+    test "view_context_class is Ractor-shareable" do
+      @original_experimental_warning = Warning[:experimental]
+      Warning[:experimental] = false
+      ActionView::LookupContext.view_context_class # needs to be eager-loaded to be ractor-shareable
+      assert_nothing_raised do
+        Ractor.new { ActionView::LookupContext.view_context_class }.value
```
etiennebarrie, line 211: I wasn't able to make it crash when running with `bin/test actionview` but I just tried and it crashes with `bundle exec rake actionview:test` 😞 
This will need a skip once again, and won't be mergeable yet 🫤

etiennebarrie: Fixed by using `ActiveSupport::Testing::Isolation`.

actionview/lib/action_view/lookup_context.rb
```diff
@@ -80,21 +79,29 @@ def self.clear
...
+
+    def self.view_context_class
+      return @view_context_class if @view_context_class
+      base = ActionView::Base # prevent recursive locking
+      @view_context_mutex.synchronize do
+        @view_context_class = base.with_empty_template_cache
       end
     end
+    @view_context_mutex = Mutex.new
+    ActiveSupport.on_load(:action_view) { ActionView::LookupContext.view_context_class }
```
gmcgibbon, line 104: This would load Active Record when Action View loads, wouldn't it? That could cause a speed regression.

etiennebarrie: As far as I can tell it doesn't? There's no mention of ActiveRecord here, where would it come from?

Ran this little script against main and this branch and I don't see a difference:
```ruby
root = __dir__
$LOAD_PATH.unshift("#{root}/actioncable/lib", "#{root}/actionmailbox/lib", "#{root}/actionmailer/lib",
  "#{root}/actionpack/lib", "#{root}/actiontext/lib", "#{root}/actionview/lib", "#{root}/activejob/lib",
  "#{root}/activemodel/lib", "#{root}/activerecord/lib", "#{root}/activestorage/lib", "#{root}/activesupport/lib",
  "#{root}/railties/lib")

require "action_view"
require "action_cable"
require "action_mailbox"
require "action_mailer"
require "action_controller"
require "action_text"
require "action_view"
require "active_job"
require "active_model"
require "active_record"
require "active_storage"
require "active_support"
require "rails"

p ActiveSupport.instance_variable_get(:@loaded).fetch(:active_record).any?
_ = ActionView::Base
p ActiveSupport.instance_variable_get(:@loaded).fetch(:active_record).any?
```

actionview/test/template/lookup_context_test.rb
```diff
@@ -199,6 +199,23 @@ def teardown
...
+  class LookupContextRactorTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    test "view_context_class is Ractor-shareable" do
+      @original_experimental_warning = Warning[:experimental]
+      Warning[:experimental] = false
+      ActionView::LookupContext.view_context_class # needs to be eager-loaded to be ractor-shareable
+      assert_nothing_raised do
+        Ractor.new { ActionView::LookupContext.view_context_class }.value
+      end
```
byroot:
```suggestion
      assert_same ActionView::LookupContext.view_context_class, Ractor.new { ActionView::LookupContext.view_context_class }.value
```

?

etiennebarrie: 👍

PR #57825, opened 2026-06-23
Introduce ractorize

andrewn617:
### Motivation / Background

We are working on making the framework Ractor safe. This PR enables a basic Rails application to be made ractor shareable.

### Detail

We introduce the `Rails::Application#ractorize!` method which prepares the application to be shared and then calls `Ractor.make_shareable` on it. For now, there are some restrictions on the setup of the application:

- Application code and routes must be eager loaded
- Action cable is not supported
- App must use a null cache store

We hope to remove these restrictions as the feature becomes more mature, this is simply a starting point.

For the last few weeks we have been setting the groundwork, but this PR includes a few final pieces to get this working:

- Making inheritable options ractor safe (there is a separate PR for this with some debate around approach https://github.com/rails/rails/pull/57757, I thought it'd be helpful to open this PR to continue that discussion)
- Removing default procs that mutate the hash when freezing some hashes
- We make the application's key generators store Ractor local, since concurrent map is not Ractor-shareable. If we have a production ready ractor-safe map in the future we can use that, but this is a pragmatic compromise for now.
- Eagerly building each engine's app and throwing away the mutex, which is not ractor shareable
- Making a few procs ractor shareable

### Checklist

### DIFF
--- a/activesupport/lib/active_support/key_generator.rb
+++ b/activesupport/lib/active_support/key_generator.rb
@@ -60,11 +60,24 @@ module ActiveSupport
     def initialize(key_generator)
       @key_generator = key_generator
       @cache_keys = Concurrent::Map.new
+      @ractor_key = nil
+    end
+
+    def freeze
+      @ractor_key = "_caching_key_generator_#{object_id}".to_sym
+      Ractor[@ractor_key] = @cache_keys
+      @cache_keys = nil
+      super
     end
 
     # Returns a derived key suitable for use.
     def generate_key(*args)
-      @cache_keys[args.join("|")] ||= @key_generator.generate_key(*args)
+      cache_keys[args.join("|")] ||= @key_generator.generate_key(*args)
     end
+
+    private
+      def cache_keys
+        @cache_keys || (Ractor[@ractor_key] ||= Concurrent::Map.new)
+      end
   end
 end
--- a/railties/test/application/ractors_test.rb
+++ b/railties/test/application/ractors_test.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require "isolation/abstract_unit"
+require "active_support/testing/ractors_assertions"
+
+if RUBY_VERSION >= "4.0"
+  module ApplicationTests
+    class RactorsTest < ActiveSupport::TestCase
+      include ActiveSupport::Testing::Isolation
+      include ActiveSupport::Testing::RactorsAssertions
+
+      def setup
+        build_app
+
+        add_to_env_config "production", "ActiveSupport::Ractors.unshareable_proc_action = :raise"
+
+        # Remove defaults that are not compatible
+        add_to_env_config "production", "config.logger = ActiveSupport::Ractors::Logger.new"
+        add_to_env_config "production", "config.public_file_server.enabled = false" # Requires release of https://github.com/rack/rack/pull/2469
+        add_to_env_config "production", "config.cache_store = :null_store"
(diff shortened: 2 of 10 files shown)


railties/test/application/ractors_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      include ActiveSupport::Testing::RactorsAssertions
+
+      def setup
+        build_app
+
+        add_to_env_config "production", "ActiveSupport::Ractors.unshareable_proc_action = :raise"
+
+        # Remove some defaults that are not compatible
+        add_to_env_config "production", "config.logger = ActiveSupport::Logger.new(nil)"
+        add_to_env_config "production", "config.public_file_server.enabled = false"
```
andrewn617: This is fixed in https://github.com/rack/rack/pull/2469 so we can remove this when a version ships with that commit. But there are some other ractor related things to push to Rack first

railties/test/application/ractors_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      include ActiveSupport::Testing::Isolation
+      include ActiveSupport::Testing::RactorsAssertions
+
+      def setup
+        build_app
+
+        add_to_env_config "production", "ActiveSupport::Ractors.unshareable_proc_action = :raise"
+
+        # Remove some defaults that are not compatible
+        add_to_env_config "production", "config.logger = ActiveSupport::Logger.new(nil)"
```
andrewn617: This can be switched to the ractor logger when that ships https://github.com/rails/rails/pull/57684

gmcgibbon (approved):

activesupport/lib/active_support/key_generator.rb
```diff
@@ -60,11 +60,24 @@ class CachingKeyGenerator
     def initialize(key_generator)
       @key_generator = key_generator
       @cache_keys = Concurrent::Map.new
+      @ractor_key = nil
+    end
+
+    def freeze
+      @ractor_key = "_caching_key_generator_#{object_id}".to_sym
+      Ractor[@ractor_key] = @cache_keys
+      @cache_keys = nil
```
Edouard-chin, line 69: I would find this easier to understand if we used a `if/else` condition in the `cache_keys` method rather than setting this to `nil`.

```ruby
if Ractor.main
  @cache_keys
else
  Ractors[@ractor_key]
end
```

andrewn617: Yeah I agree it would but unfortunately we can't do it because the cache keys are not freezable (since they are a concurrent map).

We could make it ractor local from the start. But then apps not using ractors would take a (slight) perf hit when accessing this.

andrewn617: Another option I played with is introducing a "ractor cache key generator" that puts the cache key generator in a ractor local. That way the cache key generator itself doesn't have to know about freezing.

But it felt like a bit of rigamarole to set it up at freezing time. But on second thought maybe it's worth it to do that.

gmcgibbon: I feel like we will hit this problem again on different ivars of a Rails application. It would likely make sense to put Ractor local storage there instead. I was also thinking that because Ractors are/can be reused, we might want to use a weak map, but in practice, applications are global so maybe a non-issue.

gmcgibbon: Actually, `"action_dispatch.key_generator" => key_generator` in the env config makes this difficult. I don't think the config should be made local, so the current approach is ideal for now.

PR #57826, opened 2026-06-23
Deprecate Mime::SET, Mime::LOOKUP and Mime::EXTENSION_LOOKUP

etiennebarrie:
These constants expose the MIME type registries directly, leaking mutable internal state. Everything they're used for is already covered by the public API, so deprecate direct access in favor of:

    Mime::SET              -> Mime.symbols / Mime[...] / Mime::Type.lookup
    Mime::LOOKUP           -> Mime::Type.lookup
    Mime::EXTENSION_LOOKUP -> Mime.extensions / Mime::Type.lookup_by_extension / Mime[...]

Enumerating every registered extension (including synonyms) was previously only possible through Mime::EXTENSION_LOOKUP, so Mime.extensions is added as the public counterpart to Mime.symbols to cover that case.

The constants become ActiveSupport::Deprecation::DeprecatedObjectProxy instances that still delegate to the registries but warn on use.

### DIFF
--- a/actionpack/lib/action_dispatch/http/mime_type.rb
+++ b/actionpack/lib/action_dispatch/http/mime_type.rb
@@ -3,6 +3,8 @@
 # :markup: markdown
 
 require "singleton"
+require "active_support/deprecation"
+require "action_dispatch/deprecator"
 
 module Mime
   class Mimes
@@ -43,27 +45,49 @@ module Mime
     end
   end
 
-  SET              = Mimes.new
-  EXTENSION_LOOKUP = {} # rubocop:disable Style/MutableConstant
-  LOOKUP           = {} # rubocop:disable Style/MutableConstant
+  @registry            = Mimes.new
+  @lookup_by_string    = {}
+  @lookup_by_extension = {}
+
+  SET = ActiveSupport::Deprecation::DeprecatedObjectProxy.new( # :nodoc:
+    @registry,
+    "Mime::SET is deprecated. Use Mime.symbols to enumerate registered types, or Mime[...] / Mime::Type.lookup to look one up.",
+    ActionDispatch.deprecator,
+  )
+  LOOKUP = ActiveSupport::Deprecation::DeprecatedObjectProxy.new( # :nodoc:
+    @lookup_by_string,
+    "Mime::LOOKUP is deprecated. Use Mime::Type.lookup instead.",
+    ActionDispatch.deprecator,
+  )
+  EXTENSION_LOOKUP = ActiveSupport::Deprecation::DeprecatedObjectProxy.new( # :nodoc:
+    @lookup_by_extension,
+    "Mime::EXTENSION_LOOKUP is deprecated. Use Mime.extensions to enumerate registered extensions, or Mime::Type.lookup_by_extension / Mime[...] to look one up.",
+    ActionDispatch.deprecator,
+  )
 
   class << self
+    attr_reader :registry, :lookup_by_string, :lookup_by_extension # :nodoc:
+
     def [](type)
       return type if type.is_a?(Type)
       Type.lookup_by_extension(type)
     end
 
     def symbols
-      SET.symbols
+      @registry.symbols
+    end
+
+    def extensions
+      @lookup_by_extension.keys
     end
 
     def valid_symbols?(symbols) # :nodoc:
-      SET.valid_symbols?(symbols)
+      @registry.valid_symbols?(symbols)
     end
 
     def fetch(type, &block)
       return type if type.is_a?(Type)
-      EXTENSION_LOOKUP.fetch(type.to_s, &block)
+      @lookup_by_extension.fetch(type.to_s, &block)
     end
   end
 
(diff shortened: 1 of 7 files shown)


actionpack/lib/action_dispatch/http/mime_type.rb
```diff
@@ -43,9 +45,26 @@ def valid_symbols?(symbols) # :nodoc
     end
   end
 
-  SET              = Mimes.new
-  EXTENSION_LOOKUP = {} # rubocop:disable Style/MutableConstant
-  LOOKUP           = {} # rubocop:disable Style/MutableConstant
+  REGISTRY            = Mimes.new
```
gmcgibbon: If we're defining new api for this, and the constants get mutated, wouldn't it make sense to use ivars instead?

etiennebarrie:
Yeah I had ivars instead of constants at first but I switched back because I was worried about performance degradation. The difference is not huge but it's still there.

One additional advantage of constants is lexical scoping which allows us to skip defining accessors, it's a bit nicer to access the constants directly. But yes, I have an upcoming PR which will freeze and change those constants on mutation, which will involve `remove_const`/`const_set` and at that point it does become a bit nasty.

But I think we should also deprecate mutating after configuration, so maybe constants could make more sense in the end? 🙃 

```ruby
require "benchmark/ips"

# ---- CONSTANT style: lexical constant, inline-cached by the iseq ----
module ConstStyle
  LOOKUP = { "text/html" => :html, "application/json" => :json }.freeze
  class Type
    def self.lookup(k) = LOOKUP[k]            # const resolved via lexical nesting
  end
end

# ---- MODULE IVAR style: ivar on the module + reader, called from nested scope ----
module IvarStyle
  @lookup = { "text/html" => :html, "application/json" => :json }.freeze
  class << self
    attr_reader :lookup
  end
  class Type
    def self.lookup(k) = IvarStyle.lookup[k]  # const(IvarStyle) + attr_reader + index
  end
end

# ---- FLOOR: same-scope ivar (no module hop), just for reference ----
module SameScopeIvar
  @lookup = { "text/html" => :html, "application/json" => :json }.freeze
  class << self
    attr_reader :lookup
    def lookup_key(k) = @lookup[k]            # direct @ivar in the defining scope
  end
end

KEY = "text/html"

Benchmark.ips do |x|
  x.report("lexical constant")     { ConstStyle::Type.lookup(KEY) }
  x.report("module ivar (reader)") { IvarStyle::Type.lookup(KEY) }
  x.report("same-scope ivar")      { SameScopeIvar.lookup_key(KEY) }
  x.compare!
end
```

</details>

<details><summary>Ruby 3.4.9</summary>

```
ruby 3.4.9 (2026-03-11 revision 76cca827ab) +YJIT +PRISM [arm64-darwin25]
    lexical constant     3.001M i/100ms
module ivar (reader)     2.516M i/100ms
     same-scope ivar     2.559M i/100ms
    lexical constant     35.538M (± 0.3%) i/s   (28.14 ns/i) -    180.045M in   5.066223s
module ivar (reader)     28.981M (± 0.3%) i/s   (34.50 ns/i) -    145.951M in   5.036035s
     same-scope ivar     29.435M (± 0.3%) i/s   (33.97 ns/i) -    148.436M in   5.042880s

Comparison:
    lexical constant: 35538356.7 i/s
     same-scope ivar: 29434682.6 i/s - 1.21x  slower
module ivar (reader): 28981405.8 i/s - 1.23x  slower
```

</details>

<details><summary>Ruby 4.0.5</summary>

```
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +YJIT +PRISM [arm64-darwin25]
    lexical constant     3.073M i/100ms
module ivar (reader)     2.897M i/100ms
     same-scope ivar     2.918M i/100ms
    lexical constant     34.884M (± 0.4%) i/s   (28.67 ns/i) -    175.137M in   5.020620s
module ivar (reader)     32.445M (± 1.8%) i/s   (30.82 ns/i) -    162.230M in   5.000206s
     same-scope ivar     32.656M (± 0.7%) i/s   (30.62 ns/i) -    163.419M in   5.004177s

Comparison:
    lexical constant: 34883563.6 i/s
     same-scope ivar: 32656469.2 i/s - 1.07x  slower
module ivar (reader): 32444704.9 i/s - 1.08x  slower
```

</details>

<details><summary>Ruby head</summary>

```
ruby 4.1.0dev (2026-06-26T12:53:14Z upstream/master cf93b84bb9) +YJIT +PRISM [arm64-darwin25]
    lexical constant     3.061M i/100ms
module ivar (reader)     2.847M i/100ms
     same-scope ivar     2.927M i/100ms
    lexical constant     33.634M (± 0.3%) i/s   (29.73 ns/i) -    168.337M in   5.004989s
module ivar (reader)     30.504M (± 4.7%) i/s   (32.78 ns/i) -    153.740M in   5.039933s
     same-scope ivar     31.373M (± 0.6%) i/s   (31.87 ns/i) -    158.044M in   5.037610s

Comparison:
    lexical constant: 33633865.1 i/s
     same-scope ivar: 31372859.4 i/s - 1.07x  slower
module ivar (reader): 30504416.2 i/s - 1.10x  slower
```

</details>

gmcgibbon:
This is tricky. Can we try benchmarking a request or something where mime types are used to determine if the impact is significant? I think ivars make more sense, and I'm surprised we'd have to use constants for performance sake.

etiennebarrie:
I don't think the performance difference will make a difference (still doing benchmarking/profiling to confirm). And ivars make more sense than constants for registries or caches that get mutated, but in that case, since we ultimately want the constants to hold data that's only mutated during initialization, but then becomes truly constant for the rest of the application lifecycle, I feel like constants could work better than ivars.

Edit: here's a benchmark with real requests: https://github.com/etiennebarrie/rails-app/tree/mime-storage-benchmark it's not really visible at the request level.

gmcgibbon:
Let's proceed with ivars then. I think it makes more sense.

etiennebarrie:
I moved to ivars but testing this I also realized that propshaft needs an additional API (`Mime.extensions`) to list the extensions.

gmcgibbon (approved):

PR #57827, opened 2026-06-23
Warn when a database config provides both a `url` and conflicting keys

cgmoore120:
### Summary

In a multi-database `config/database.yml`, a secondary database commonly inherits the primary's `url:` via a YAML anchor and then sets its own `database:`:

```yaml
production:
  primary: &primary
    url: <%= ENV["DATABASE_URL"] %>   # e.g. postgres://user:pass@host/myapp_production
  queue:
    <<: *primary
    database: myapp_production_queue   # silently ignored
```

Because the value parsed from the URL takes precedence over sibling keys, `queue` silently connects to `myapp_production`, not `myapp_production_queue`. A confusing downstream symptom: `db:prepare` then appears to "skip" the secondary — it resolves to the primary database, whose `schema_migrations` table already exists, so nothing is created.

### What this changes

This does **not** change precedence — the URL still wins, which is intentional. It only adds a warning when a single configuration entry provides both a `url` and a conflicting key (`database:`, `host:`, …), so the otherwise-silent conflict becomes visible:

```
Database configuration for "queue" (production) provides both a `url` and the
conflicting key(s): database. The value(s) from the URL take precedence, so the
explicitly configured value(s) are ignored.
```

The check lives in `build_db_config_from_hash`, so it applies only to a literal `url:` key written alongside other keys in the same entry. URLs supplied through the `DATABASE_URL` environment variable go through a separate path (`environment_url_config`) and are intentionally **not** warned about — overriding `database.yml` via `DATABASE_URL` is a supported, common pattern.

The warning uses `Kernel#warn` (writes to `$stderr`, silenceable), fires only on a genuine conflict (explicit value present and different), and never interferes with config loading (the URL parsing in the check is guarded).

### Notes

- A CHANGELOG entry and tests are included (conflict warns; url-only doesn't; matching value doesn't; `DATABASE_URL` override doesn't).
- If a warning in this code path is undesirable, I'm happy to convert this into a documentation clarification in the multiple-databases guide instead.

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,12 @@
+*   Warn when a database configuration provides both a `url` and conflicting
+    keys such as `database:` or `host:` in the same entry (for example a
+    `database:` alongside a `url:` inherited via a YAML anchor). The value from
+    the URL still takes precedence; the warning just surfaces the otherwise
+    silent conflict. URLs provided through the `DATABASE_URL` environment
+    variable are unaffected.
+
+    *Chris Moore*
+
 *   Support polymorphic associations with custom primary keys through `:inverse_of`.
 
     When using polymorphic associations with `:inverse_of`, ActiveRecord now respects
--- a/activerecord/lib/active_record/database_configurations.rb
+++ b/activerecord/lib/active_record/database_configurations.rb
@@ -277,12 +277,46 @@ module ActiveRecord
         config_without_url = config.dup
         config_without_url.delete :url
 
+        warn_about_keys_ignored_by_url(env_name, name, url, config_without_url) if url
+
         DatabaseConfigurations.db_config_handlers.reverse_each do |handler|
           config = handler.call(env_name, name, url, config_without_url)
           return config if config
         end
 
         nil
+      end
+
+      # When a configuration provides both a +url+ and keys that the URL also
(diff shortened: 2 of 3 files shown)


cgmoore120:
Pushed a fix for the failing build.

The first revision did the check inside `UrlConfig#initialize`, which also fired for the legitimate case where `DATABASE_URL` overrides a `database.yml` that uses discrete keys (e.g. `database:`) — tripping the existing `assert_silent` guard around `setup_initial_database_yaml`.

I've moved the detection to `build_db_config_from_hash`, so it now applies only to a literal `url:` key written alongside conflicting keys in the same configuration entry. URLs coming from `DATABASE_URL` (handled separately in `environment_url_config`) are no longer warned about, and I added a regression test covering that case.

yahonda:
Thank you for digging into this and for the clear write-up.

I'd rather not add a check to the connection configuration path that most
users won't need. The precedence this would warn about — that when a `url` and
other connection keys conflict in the same entry, the value from the URL takes
precedence — is already documented in the "Connection Preference" section of
the configuring guide:

https://github.com/rails/rails/blob/2db1126cf881eb05fd99727c3b7e11cee3e8b4c3/guides/source/configuring.md?plain=1#L3865

So I think this behavior is already covered there, and I'm going to close this
pull request.

If you think the documentation could be improved here, please feel free to open
a separate documentation pull request. I'm looking forward to your future
contributions.

PR #57829, opened 2026-06-23
Refactor Arel::Table to use keyword name: argument

gmcgibbon:
### Motivation / Background

This Pull Request has been created because as of https://github.com/rails/rails/pull/57642, `Arel::Table` now takes an optional optional argument. We can make this a keyword instead. The class is nodoc so it is private API.

### Detail

This Pull Request changes `Arel::Table` names to be keywords instead of needing to pass nil for the klass use-case.

### Checklist

### DIFF
--- a/activerecord/lib/arel/table.rb
+++ b/activerecord/lib/arel/table.rb
@@ -11,7 +11,7 @@ module Arel # :nodoc: all
     attr_writer :name
     attr_reader :table_alias, :klass
 
-    def initialize(name = nil, as: nil, klass: nil, type_caster: klass&.type_caster)
+    def initialize(name: nil, as: nil, klass: nil, type_caster: klass&.type_caster)
       raise ArgumentError, "A name or klass is required." unless klass || name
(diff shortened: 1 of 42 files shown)


activerecord/lib/arel/table.rb
```diff
@@ -11,7 +11,7 @@ class << self; attr_accessor :engine; end
     attr_writer :name
```
kamipo, line 11: What do you think about removing this `attr_writer :name`?
It doesn't seem necessary to be able to change the name of an `Arel::Table` after it has been instantiated.

gmcgibbon: Yeah I don't see any references at a glance. Makes sense to remove it to me!

byroot:
For the record this change broke Intercom.

I know `Arel` isn't really public API but in practice a ton of people and gems do use it. I'll update our app, but IMO we should probably support the old positional argument with a deprecation, otherwise I doubt I'll be the only one to run into this.

PR #57830, opened 2026-06-23
Make ActionController Renderers ractor safe

Edouard-chin:
### Motivation / Background

I'd like to make Action Controller Renderers Ractor safe. It currently is not because an object that holds the name of the renderers (`js`, `xml`...) isn't frozen. I can't just freeze it because we mutate it at boot time.

### Detail

This PR includes 3 commits

1. Deprecate a constant (It was documented but I think this should be for internal usage only).
2. Add a way for deprecated constant to change their target after initialization.
3. Stop mutating a Set when adding or removing renderer and freeze the set.

I added more explanation in each commit.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal/renderers.rb
+++ b/actionpack/lib/action_controller/metal/renderers.rb
@@ -67,8 +67,8 @@ module ActionController
       extend ActiveSupport::Concern
       include Renderers
 
-      included do
-        self._renderers = RENDERERS
+      def _all_renderers # :nodoc:
+        _renderers + Renderers.all
       end
     end
 
@@ -102,7 +102,9 @@ module ActionController
     #     end
     def self.add(key, &block)
       define_method(_render_with_renderer_method_name(key), &block)
-      @all << key.to_sym
+      (@all += [key.to_sym]).freeze
+
+      RENDERERS.target = @all
     end
 
     # This method is the opposite of add method.
@@ -111,9 +113,11 @@ module ActionController
     #
     #     ActionController::Renderers.remove(:csv)
     def self.remove(key)
-      @all.delete(key.to_sym)
+      (@all -= [key.to_sym]).freeze
       method_name = _render_with_renderer_method_name(key)
       remove_possible_method(method_name)
+
+      RENDERERS.target = @all
     end
 
     def self._render_with_renderer_method_name(key) # :nodoc:
@@ -171,7 +175,7 @@ module ActionController
     end
 
     def _render_to_body_with_renderer(options) # :nodoc:
-      _renderers.each do |name|
+      _all_renderers.each do |name|
         if options.key?(name)
           _process_options(options)
           method_name = Renderers._render_with_renderer_method_name(name)
@@ -179,6 +183,10 @@ module ActionController
         end
       end
       nil
+    end
+
+    def _all_renderers # :nodoc:
+      _renderers
     end
 
     add :json do |json, options|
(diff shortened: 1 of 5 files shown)


actionpack/lib/action_controller/metal/renderers.rb
```diff
@@ -23,9 +25,17 @@ def initialize(format)
   module Renderers
     extend ActiveSupport::Concern
 
-    # A Set containing renderer names that correspond to available renderer procs.
-    # Default values are `:json`, `:js`, `:xml`.
-    RENDERERS = Set.new
+    class << self
+      attr_reader :all # :nodoc:
```
matthewd: Assuming we _don't_ switch to immediately freezing the internal value, should we document the reader and have it return a frozen copy? ISTM the most likely existing usage of `RENDERERS` would be reading what's already registered.

actionpack/lib/action_controller/metal/renderers.rb
```diff
@@ -87,7 +97,9 @@ module All
     #     end
     def self.add(key, &block)
       define_method(_render_with_renderer_method_name(key), &block)
-      RENDERERS << key.to_sym
+      @all += [key.to_sym]
```
matthewd: If we care about maintaining existing observable behaviour during deprecation, this should mutate in place, to ensure an existing reference-via-constant sees the changes.

matthewd (commented):
I'm not sure we even need a stepped deprecation here... while the constant was documented, it's never been the obvious way to register a renderer -- you would've had to insert into this set _and_ monkey-patch an internally-named renderer method into the same place `add` would've put it.

Unless we know of at least anecdotal usage in that way, I'm inclined to assume we can safely freeze this immediately.

activesupport/lib/active_support/deprecation/proxy_wrappers.rb
```diff
@@ -42,6 +42,10 @@ def initialize(object, message, deprecator)
         @deprecator = deprecator
       end
 
+      def target=(object)
```
etiennebarrie, line 45: Oh I'm going to need something like this too for #57826

Edouard-chin:
> you would've had to insert into this set and monkey-patch an internally-named renderer method into the same place add would've put it.

Yep that's right. I'm assuming people wouldn't mutate that constants as well.
I don't have any evidence of this, but that constant may have been used to check what renderers exists (couldn't find any obvious public API to know this). Your suggestion about making the new reader public makes sense.

It's unlikely that application code was mutating that constant, but I'm less sure about the "existing reference-via-constant" in which case we have to continue to mutate in place and can't freeze now.

I'm inclined to freeze now as you suggested.

etiennebarrie:
This is limited evidence, but I could only find tests/specs checking the contents, no mutation: https://github.com/search?q=ActionController%3A%3ARenderers%3A%3ARENDERERS+NOT+path%3A%2F.rbi%24%2F+NOT+language%3AHTML+&type=code&p=1

etiennebarrie:
I found this one too: https://github.com/heartcombo/responders/blob/v3.2.0/lib/action_controller/responder.rb#L284

We should add a method to check inclusion for these cases.

actionpack/lib/action_controller/metal/renderers.rb
etiennebarrie: Remove the reference to `Renderers::RENDERERS` here.

gmcgibbon (approved):

PR #57831, opened 2026-06-23
Guides: Include `min` value for `form.number_field` [ci skip]

duggiefresh:
### Motivation / Background

This adds a `min` value to the `form.number_field` for a Product's `inventory_count`. This helps prevent the reader from setting and saving the inventory count as negative number within the demo application.

### Detail

This PR updates the example within the [Rails Guide: Action Mailer and Email Notifications](https://guides.rubyonrails.org/getting_started.html#action-mailer-and-email-notifications)

### Additional information

N/A

### Checklist

### DIFF
--- a/guides/source/getting_started.md
+++ b/guides/source/getting_started.md
@@ -2094,7 +2094,7 @@ We'll need to add the inventory count to the product form in
 
   <div>
     <%= form.label :inventory_count, style: "display: block" %>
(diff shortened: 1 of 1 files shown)


p8:
Thanks for the PR @duggiefresh 
I can see the benefit of this, but we also add a validation to the Product that can no longer be tested in the form if we add a `min`. Some readers might think the validation in the model also affects the form.

PR #57834, opened 2026-06-24
Make EventReporter::LogSubscriber::LEVEL_CHECKS ractor safe

55728:
Apply the same Ractor-safety treatment used in `ActiveSupport::LogSubscriber::LEVEL_CHECKS` to its `EventReporter` counterpart: wrap the level-check lambdas with `ActiveSupport::Ractors.shareable_lambda` so the constant is Ractor-shareable.


55728:
Consolidated into #57839 (grouped by gem).

PR #57835, opened 2026-06-24
Make Arel BIND_BLOCK constants ractor safe

55728:
Wrap the bind-substitution procs in `Arel::Visitors::ToSql` and `Arel::Visitors::PostgreSQL` with `ActiveSupport::Ractors.shareable_proc` so the `BIND_BLOCK` constants are Ractor-shareable.

Neither proc captures any external state.


55728:
Consolidated into #57840 (grouped by gem).

PR #57836, opened 2026-06-24
Make StructuredEventSubscriber::DEBUG_CHECK ractor safe

55728:
Wrap the debug-mode check proc with `ActiveSupport::Ractors.shareable_proc` so `StructuredEventSubscriber::DEBUG_CHECK` is Ractor-shareable.

The proc only calls `ActiveSupport.event_reporter.debug_mode?` at the module level and captures no instance state.


55728:
Consolidated into #57839 (grouped by gem).

PR #57838, opened 2026-06-24
Make NestedAttributes::REJECT_ALL_BLANK_PROC ractor safe

55728:
Wrap the `:all_blank` reject proc with `ActiveSupport::Ractors.shareable_proc` so `ActiveRecord::NestedAttributes::ClassMethods::REJECT_ALL_BLANK_PROC` is Ractor-shareable.

The proc receives the attributes hash as an argument and captures no external state.


55728:
Consolidated into #57840 (grouped by gem).

PR #57839, opened 2026-06-24
Make some ActiveSupport constants ractor safe

55728:
Wrap the level-check lambdas in `EventReporter::LogSubscriber::LEVEL_CHECKS` and the debug-mode check proc in `StructuredEventSubscriber::DEBUG_CHECK` with `ActiveSupport::Ractors.shareable_lambda` / `.shareable_proc` so both constants are Ractor-shareable.

Neither captures any instance state — the lambdas receive the logger as an argument, and the proc only makes a module-level call to `ActiveSupport.event_reporter.debug_mode?`.


55728:
Thank you for the review @byroot 😊

PR #57832, opened 2026-06-24
Bump json to 2.20.0 and pass `allow_comments: true`

yahonda:
### Motivation / Background

This Pull Request bumps the bundled json gem to 2.20.0.

json 2.20.0 (ruby/json#1001, commit 138b9a2c) deprecates the parser's default
support for JavaScript comments in favor of an `allow_comments: true` option.
Parsing JSON that contains `//` or `/* */` comments without that option now
emits a deprecation warning ("Encountered comment in JSON. This will raise an
error in json 3.0 unless enabled via `allow_comments: true`").

### Detail

This Pull Request bumps json `2.15.2` → `2.20.0` in `Gemfile.lock` and passes
`allow_comments: true` at the parse sites that intentionally read comment-bearing
JSON:

- Active Support's JSON message serializer fallback `_load`
  (`active_support/messages/serializer_with_fallback.rb`), which loads irregular
  JSON through `ActiveSupport::JSON.decode`.
- Railties' `devcontainer.json`, which is JSONC by spec: the `db:system:change`
  generator that re-reads it (`change_generator.rb`) and the generator test
  helper that asserts on the generated file (`generators_test_helper.rb`).

The comments are intentional in both subsystems — the Active Support test
exercises comment tolerance, and `devcontainer.json` is JSONC by spec — so the
change opts in to `allow_comments: true` rather than removing the comments.

This fixes the failures observed in the nightly build
([rails-nightly #4483](https://buildkite.com/rails/rails-nightly/builds/4483)):

- `MessagesSerializerWithFallbackTest#test_:json_serializer_can_load_irregular_JSON` ([log](https://buildkite.com/rails/rails-nightly/builds/4483#019ef648-9b6f-463e-9828-03260b201c87/L7951))
- `AppGeneratorTest#test_devcontainer_skip_kamal` ([log](https://buildkite.com/rails/rails-nightly/builds/4483#019ef648-9b73-49c0-8b0c-91f869c53beb/L1340))
- `Rails::Generators::Db::System::ChangeGeneratorTest#test_change_from_db_with_devcontainer_service_to_one_without` ([log](https://buildkite.com/rails/rails-nightly/builds/4483#019ef648-9b74-4ea5-a182-f3a5ed24f19f/L1490))

<details>
<summary>Failures fixed (nightly #4483)</summary>

```
MessagesSerializerWithFallbackTest#test_:json_serializer_can_load_irregular_JSON:
[...]

PR #57842, opened 2026-06-24
Make ActiveSupport::JSON encoders Ractor-shareable

etiennebarrie:
Take 2 on #57814 

Using `ActiveSupport::Testing::Isolation`, we can get around the Ruby crashes.

### DIFF
--- a/activesupport/test/json/encoding_test.rb
+++ b/activesupport/test/json/encoding_test.rb
@@ -603,6 +603,21 @@ EXPECTED
     end
 end
 
+if RUBY_VERSION >= "4.0"
+  class JSONRactorShareabilityTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    def test_encoders_are_ractor_shareable
+      assert_equal '{"a":1}', Ractor.new { ActiveSupport::JSON.encode({ a: 1 }, escape: false) }.value
+      assert_equal '{"a":1}', Ractor.new { ActiveSupport::JSON.encode({ a: 1 }) }.value
+      ActiveSupport::JSON::Encoding.with(escape_js_separators_in_json: false) do
+        assert_equal '{"a":1}', Ractor.new { ActiveSupport::JSON.encode({ a: 1 }) }.value
+      end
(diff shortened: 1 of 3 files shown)


activesupport/test/json/encoding_test.rb
```diff
@@ -603,6 +603,27 @@ def with_time_precision(value)
     end
 end
 
+if RUBY_VERSION >= "4.0"
+  class JSONRactorShareabilityTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    def test_encoders_are_ractor_shareable
+      skip "Missing Ractor-shareable JSON::Coder" unless Gem::Version.new(::JSON::VERSION) >= Gem::Version.new("2.17.0")
```
gmcgibbon: Why do we need this?

activesupport/test/json/encoding_test.rb
```diff
@@ -603,6 +603,27 @@ def with_time_precision(value)
...
+      begin
+        original_experimental, Warning[:experimental] = Warning[:experimental], false
```
gmcgibbon: We do this a lot. We should either create a helper for this and revert my patch to strict warnings that ignores Ractor warnings, or just let the warnings happen and expect them to be ignored on CI. I think it makes sense to go the helper route.

etiennebarrie: I went with suppressing it completely.

activesupport/test/json/encoding_test.rb
etiennebarrie: I had this before json was upgraded, but I guess we don't need it since CI now runs 2.20.0 since #57832.

activesupport/test/json/encoding_test.rb
gmcgibbon: Yeah I thought I made this change already, but it hasn't merged yet: https://github.com/Shopify/rails/blob/5802e5ce6c363713f3cf8686efa42211f8ad0773/tools/strict_warnings.rb#L29 😅 

LGTM

PR #57845, opened 2026-06-24
Report PostgreSQL default timestamp/time precision as 6

adrianna-chang-shopify:
### Motivation / Background

PostgreSQL omits the typmod for bare timestamp/time columns, but those types still behave with microsecond precision by default. Reflect that effective precision in Active Record metadata so type casting normalizes timestamps consistently with persisted database values.

### Detail

  This change now uses the PostgreSQL typmod value passed through the type map
  instead of parsing the formatted SQL type string. When `fmod == -1`, Active
  Record reports precision `6`; otherwise it uses the explicit typmod precision.
  That keeps column metadata and type casting aligned with PostgreSQL's effective
  behavior for bare and explicit `timestamp`, `timestamptz`, `time`, and `timetz`
  columns.

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,11 @@
+*   Report PostgreSQL default timestamp and time precision as 6.
+
+    Bare PostgreSQL `timestamp` and `time` columns now use their effective
+    microsecond precision in Active Record type metadata, matching the
+    database's persisted precision.
+
+    *Adrianna Chang*
+
 *   Fix inverse association matching for `has_many` and `has_one` associations
     with association-specific composite primary keys.
 
--- a/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+++ b/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
@@ -810,6 +810,14 @@ module ActiveRecord
           end
         end
 
+        def register_class_with_precision(mapping, key, klass, **kwargs) # :nodoc:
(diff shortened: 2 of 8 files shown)


yahonda:
Hi,

Rather than parsing the precision out of the type string with a regexp, I think we can let PostgreSQL tell us the precision directly.

We could use `atttypmod` from `pg_attribute` — passed to the type map block as `fmod` — and treat `fmod == -1` (no typmod) as the default precision of 6. The PostgreSQL documentation describes the `-1` value:

> The value will generally be -1 for types that do not need `atttypmod`.
>
> — [pg_attribute, PostgreSQL 10](https://www.postgresql.org/docs/10/catalog-pg-attribute.html)

For `timestamp`/`time`, `atttypmod` is the precision itself, and bare columns report `-1`. This has been stable from PostgreSQL 10 through current.

This means overriding `register_class_with_precision` rather than just `extract_precision` (the latter only receives the `sql_type` string, not `fmod`), but I think it reads more clearly than maintaining the regexp — and `numeric`/`interval` already use `fmod` the same way.

Example implementation: https://github.com/yahonda/rails/blob/357d601b852551f02f5066df1766c3724d3d8e9e/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L832-L837

adrianna-chang-shopify:
Hey @yahonda , thank you kindly for the review! ❤️ Completely agree that this is a cleaner approach, I've made the suggested edits!

gmcgibbon (approved):

PR #57852, opened 2026-06-25
Freeze configuration

gmcgibbon:
### Motivation / Background

This Pull Request has been created to freeze active record and action pack configurations after initialize. This allows them to be referenced in non-main ractors. The `make_shareable` call will only apply in Rubies where this is relevant 4.0+.

### Detail

This Pull Request changes the railtie of Active Record and Action Pack to freeze configuration values that are mutable. That would be  `ActionDispatch::ExceptionWrapper.rescue_responses`, `rescue_templates`, `wrapper_exceptions`, and `silent_exceptions`. Also `ActiveRecord.query_transformers`, but we may need to add more later.

### Checklist

### DIFF
--- a/actionmailer/lib/action_mailer/railtie.rb
+++ b/actionmailer/lib/action_mailer/railtie.rb
@@ -75,6 +75,15 @@ module ActionMailer
       app.config.paths["test/mailers/previews"].concat(options.preview_paths)
     end
 
+    initializer "action_mailer.freeze_configuration" do
+      config.after_initialize do
+        ActiveSupport.on_load(:action_mailer) do
+          ActiveSupport::Ractors.try_make_shareable(ActionMailer::Base.preview_paths)
+          ActiveSupport::Ractors.try_make_shareable(ActionMailer::Base.smtp_settings)
+        end
+      end
+    end
+
     config.after_initialize do |app|
       options = app.config.action_mailer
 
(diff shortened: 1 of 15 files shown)


gmcgibbon:
- Configuration values can be anything the user wants, so we need to introduce a try_make_shareable helper.
- If we're using try_make_shareable we probably don't need to introduce a new config option, we can hinge on the try ractorizing config behaviour we already have
- This patch doesn't work with 3.3, but I'm less bothered by that if we're not introducing a config option to explicitly freeze

actionmailer/lib/action_mailer/railtie.rb
```diff
@@ -75,6 +75,15 @@ class Railtie < Rails::Railtie # :nodoc:
       app.config.paths["test/mailers/previews"].concat(options.preview_paths)
     end
 
+    initializer "action_mailer.freeze_configuration" do
+      config.after_initialize do
+        ActiveSupport.on_load(:action_mailer) do
+          ActiveSupport::Ractors.try_make_shareable(ActionMailer::Base.preview_paths)
```
Edouard-chin: I'm not fully sure when those configs would ever raise a Ractor shareable error? They just hold strings and will get frozen without issues.

IIRC the previous config this PR introduced was to opt-out from freezing configuration *in case* those configs were mutated by the application at runtime. 
I think the `try_make_shareable` adds no value in this case

Edouard-chin (commented):
Made a comment on the diff hunk but have the same comment for all others `try_make_shareable` call.

gmcgibbon:
I decided to narrow the scope of this patch to just the configs we need for now. We can add other ones later.

Edouard-chin (approved):

PR #57854, opened 2026-06-25
Make Controller middlewares ractor safe

Edouard-chin:
### Motivation / Background

The middleware for a controller (not to be confused with the rack middleware stack) isn't Ractor safe. It has to be made shareable (and all objects accessible from it).

### Detail

This commit implements a copy on write approach. One of the complication is that a controller can currently either call `use` (which is a wrapper around the MiddlewareStack object), or directly access the Middleware Stack. In the latter case, we need to create a proxy so that anytime a method on the stack is called, we can refreeze the stack.

### Checklist

### DIFF
--- a/actionpack/lib/action_controller/metal.rb
+++ b/actionpack/lib/action_controller/metal.rb
@@ -2,6 +2,8 @@
 
 # :markup: markdown
 
+require "active_support/ractors"
+require "active_support/core_ext/module/delegation"
 require "action_dispatch/middleware/stack"
 
 module ActionController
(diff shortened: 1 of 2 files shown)


gmcgibbon (commented):
This can theoretically done in https://github.com/rails/rails/pull/57825, can't it? If so, we should hold off until that PR is considered.

Edouard-chin:
We can wait for https://github.com/rails/rails/pull/57825, but IMHO this change isn't only about being ractor safe and could be a good addition to prevent people from shooting themselves in the foot if they mutate the stack at runtime.

I'd like if possible to get the stack frozen, even if an application doesn't use Ractors.

actionpack/lib/action_controller/metal.rb
```diff
@@ -216,6 +217,12 @@ def initialize
       super
     end
 
+    def self.eager_load! # :nodoc:
+      ActiveSupport::Ractors.make_shareable(middleware)
+
+      super
+    end
```
andrewn617: I would rather not tie this behaviour to eager loading, since we would like to support lazy loading with ractors in the future 🤔 Plus, I think all middleware should be loaded whether app is eager or lazy loaded right? So can we put it in an after initialize

Edouard-chin: >  Plus, I think all middleware should be loaded whether app is eager or lazy loaded right? 

No, this isn't the Rack middleware stack, it's the Controller middleware stack. Each controllers has its own middleware stack, so in order to make this shareable the controller has to be loaded because it will do something like:

```
class ArticlesController < ApplicationController
  use PublisherMiddleware
end

class UserController < ApplicationController
  use SomeOtherMiddleware
end
```

We have to iterate over each controllers in order to freeze their middleware stack (and therefore it's coupled with eager loading).

---

The only way to make this shareable without eager loading is a copy on write (as described in the PR description but it's much heavier)

andrewn617: Sorry 😮‍💨 

But, that only makes coupling to eager loading more problematic. It's "ok for now" but something we would need to fix to allow lazy loading.

PR #57856, opened 2026-06-25
Load ActionDispatch::Request::Session and Utils at boot time

andrewn617:
### Motivation / Background

While working on ractor stuff I noticed `ActionDispatch::Request::Session` is not loaded until the first request. It is autoloaded since rails/rails@8ce6b0c to avoid circularly requiring `ActionDispatch::Request`.

### Detail

This class (and `Utils`) are no doced, so I don't think we need to keep it at `Request::Session`. We can move it to `Http::Session` like the other things `ActionDispatch::Request` relies on and require it when we load that file. That way it will be loaded at boot time with everything else.

### Checklist

### DIFF
--- a/actionpack/lib/action_dispatch.rb
+++ b/actionpack/lib/action_dispatch.rb
@@ -101,6 +101,11 @@ module ActionDispatch
     autoload :Parameters
     autoload :UploadedFile, "action_dispatch/http/upload"
     autoload :URL
+
+    eager_autoload do
+      autoload :Session
+      autoload :Utils
+    end
   end
 
   module Session
@@ -146,9 +151,10 @@ module ActionDispatch
   singleton_class.attr_accessor :verbose_redirect_logs
   self.verbose_redirect_logs = false
 
-  def eager_load!
+  def self.eager_load!
     super
     Routing.eager_load!
+    Http.eager_load!
   end
 end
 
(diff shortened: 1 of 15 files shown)


gmcgibbon (commented):
Would it make more sense to load the constants during boot using eager_autoload? I don't want to introduce a boot time regression in development.

andrewn617:
Sure 👍

actionpack/lib/action_dispatch/middleware/session/abstract_store.rb
```diff
@@ -5,6 +5,7 @@
 require "rack/utils"
 require "rack/request"
 require "rack/session/abstract/id"
+require "action_dispatch/http/session"
```
gmcgibbon:
```suggestion
```
Needs to be removed before merge.

gmcgibbon (approved):

actionpack/lib/action_dispatch.rb
```diff
@@ -146,9 +151,10 @@ def self.resolve_store(session_store) # :nodoc:
   singleton_class.attr_accessor :verbose_redirect_logs
   self.verbose_redirect_logs = false
 
-  def eager_load!
+  def self.eager_load!
```
andrewn617, line 154: When testing this I noticed it was still not working. The issue is this needs to be a class method, so `Routing` also was not eager loading until now unfortunately.

I think we should find a way to make setting these up a bit more automated, since it seems a bit error prone. I think we can add to the eager load namespaces automatically when we call `eager_autoload`, we just have to make sure we can handle recursive cases when eager loading a constant loads another constant that defines eager loads. Then we can get rid of all these `self.eager_load!` methods. I will try to get that working and open a PR.

PR #57858, opened 2026-06-25
Only document the public ActionView::Template API [ci skip]

etiennebarrie:
register_template_handler is the supported way to teach Action View about a new template type, and gems such as [haml](https://github.com/haml/haml/blob/v7.2.0/lib/haml/rails_template.rb#L55) and [markaby](https://github.com/markaby/markaby/blob/v0.9.4/lib/markaby/rails.rb#L9) already call it as ActionView::Template.register_template_handler. It was undocumented because it is defined on the internal Template::Handlers module that Template extends, which is marked :nodoc:.

Document it as a singleton method on ActionView::Template, where it actually lives once the module is extended, mirroring how local_assigns is documented.

Everything else under ActionView::Template is internal: the registry helper methods, the built-in handler classes (ERB, Raw, Html, Builder), and Template's own class and instance methods are not meant to be called by applications. Mark the handler classes :nodoc: and use :stopdoc: to stop documenting Template past its public entry points, so the API page only lists register_template_handler and local_assigns.

This also drops the :nodoc: on Template's `class << self`, which never had any effect: RDoc does not honor :nodoc: on a singleton class, so mime_types_implementation=, validate_formats and normalized_formats had been showing up as public API.

### DIFF
--- a/actionview/lib/action_view/template.rb
+++ b/actionview/lib/action_view/template.rb
@@ -7,8 +7,6 @@ module ActionView
   class Template
     extend ActiveSupport::Autoload
 
-    STRICT_LOCALS_REGEX = /\#\s+locals:\s+\((.*?)\)(?=\s*-?%>|\s*$)/m
-
     # === Encodings in ActionView::Template
     #
     # ActionView::Template is one of a few sources of potential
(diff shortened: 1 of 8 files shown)


etiennebarrie:
https://shopify-actionview-template-n9kd.rails-docs-preview.pages.dev/api/classes/ActionView/Template.html
vs.
https://edgeapi.rubyonrails.org/classes/ActionView/Template.html (this will change in the future but is mostly like 
https://api.rubyonrails.org/v8.1.3/classes/ActionView/Template.html with the two methods from #57780 added).

gmcgibbon (approved):

PR #57863, opened 2026-06-25
Fix alias_attribute precedence over included modules

### LINKED ISSUE #50048: `alias_attribute` no longer overrides reader method defined in concern

gramos:
### Motivation / Background

Fixes #50048.

`alias_attribute` should allow an attribute alias to override a method with the same name that comes from an included module. For example, if a model includes a module that defines `display_name`, and the model then declares `alias_attribute :display_name, :title`, calling `display_name` should read the `title` attribute.

Before this change, the method generated for the alias was defined in `GeneratedAttributeMethods`. Included modules can appear before `GeneratedAttributeMethods` in the ancestor chain, so Ruby resolved the method from the included module first and never reached the generated alias method.

### Detail

This Pull Request detects when an alias attribute method would be generated below an existing method with higher precedence in the ancestor chain. In that case, it generates the alias methods in a separate `GeneratedAliasAttributeMethods` module that is included with higher precedence.

Aliases without that conflict continue to use `GeneratedAttributeMethods`, so the normal alias generation path remains unchanged.

The change also clears generated alias methods when attribute methods are undefined, matching the existing regeneration lifecycle for generated attribute methods.

### Additional information

A regression test covers an included module defining the alias reader before `alias_attribute` is declared.

Validation run locally with Ruby 3.4.9:

```bash
cd activerecord
bin/test test/cases/attribute_methods_test.rb -i "/alias_attribute override methods defined in included modules/"
bin/test test/cases/attribute_methods_test.rb -i "/alias_attribute/"
bin/test test/cases/attribute_methods_test.rb
```

RuboCop:

```bash
bundle exec rubocop activerecord/lib/active_record/attribute_methods.rb activerecord/test/cases/attribute_methods_test.rb
```

### Checklist

### DIFF
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Fix `alias_attribute` overriding methods defined in included modules.
+
+    *Gaston Ramos*
+
 *   Report PostgreSQL default timestamp and time precision as 6.
 
     Bare PostgreSQL `timestamp` and `time` columns now use their effective
--- a/activerecord/CHANGELOG.md
+++ b/activerecord/CHANGELOG.md
@@ -1,3 +1,7 @@
+*   Fix `alias_attribute` overriding methods defined in included modules.
+
+    *Gaston Ramos*
+
 *   Avoid unnecessary association preloader queries for nil foreign keys when
     the foreign key and association primary key have different types.
 
--- a/activerecord/lib/active_record/attribute_methods.rb
+++ b/activerecord/lib/active_record/attribute_methods.rb
@@ -44,6 +44,7 @@ module ActiveRecord
         private_constant :GeneratedAttributeMethods
         @attribute_methods_generated = false
         @alias_attributes_mass_generated = false
+        @alias_attribute_method_definition_owners = {}
         include @generated_attribute_methods
 
         super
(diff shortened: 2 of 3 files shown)


gramos:
I’ve rebased this PR onto the latest `main` and resolved the conflict in `activerecord/CHANGELOG.md`.
The relevant Active Record test passes:

136 runs, 613 assertions, 0 failures, 0 errors, 0 skips.

Since this PR has been open since June 25, I’d really appreciate a review when someone has time. Thanks!

